PR #293, opened 2011-06-13
fixed #516

liswei:
fixed #516 Undefined Behaviour in ccCArray.h (signed / unsigned conversion)


cocos2dx/actions/CCActionManager.cpp
```diff
@@ -286,7 +286,7 @@ void CCActionManager::removeAction(cocos2d::CCAction *pAction)
 	if (pElement)
 	{
 		unsigned int i = ccArrayGetIndexOfObject(pElement->actions, pAction);
-		if ((int)i != -1)
+		if (UINT_MAX == i)
```
haqu, line 289: should be `if (UINT_MAX != i)`, right?

minggo:
Yeap, you are right. I am sorry for my careless to find it.

PR #302, opened 2011-06-16
Parameters In CCMutableDictionary methods should be passed by const reference instead copy

moadib:
http://codeidol.com/cpp/effective-cpp/Designs-and-Declarations/Item-20-Prefer-pass-by-reference-to-const-to-pass-by-value/


walzer:
You're right. Thanks!
I believe there're still many structure->reference parameter can be optimized in the engine. I may re-check them in the 1.0.0 upgrading work.

PR #305, opened 2011-06-20
Removed unnecessary copy of ccBMFontDef structure in CCLabelBMFont::createFontChars()

moadib:
Removed unnecessary copy of ccBMFontDef structure in CCLabelBMFont::createFontChars()

### DIFF
--- a/cocos2dx/label_nodes/CCLabelBMFont.cpp
+++ b/cocos2dx/label_nodes/CCLabelBMFont.cpp
@@ -476,7 +476,7 @@ namespace cocos2d{
             
 			kerningAmount = this->kerningAmountForFirst(prev, c);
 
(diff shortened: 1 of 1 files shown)


moadib:
GitHub's feature "Edit this file" doesn't work as expected :( Instead change of 1 line it suggest a lot of removes and adds for lines, that i've not changed... 

May be reason in non windows Line Endings?

walzer:
Merged. Thanks. 
We are used to met this problem, while some guys work on mac, and others work on windows for win32 port & android port.
But I review your diff, it shows only 1 line is modified. It's right :)

moadib:
Diff contains 1 line modified because I've used console tools, not github "edit this file" feature :)

PR #306, opened 2011-06-21
changed postfix increment of iterators to prefix increment

moadib:
May be it is not reasonable, but better to use prefix increment operator :)


walzer:
Thanks. It's said that prefix increment is faster than postfix increment in STL.

PR #307, opened 2011-06-21
fixed #528

minggo:
refactor lua support


walzer:
Merged manually. Thanks for your work

PR #318, opened 2011-06-23
Issue539

liswei:
in CCFileUtils_wophone.cpp,call unzOpen, then may not call unzClose(WVGA) .


walzer:
recommit please, don't delete HelloLua

PR #332, opened 2011-06-29
fix bug, add plist when resourcePath is not empty.

zhaixudong:
http://www.cocos2d-x.org/boards/6/topics/2080


zhaixudong:
fix bug, add plist when resourcePath is not empty.
http://www.cocos2d-x.org/boards/6/topics/2080

PR #344, opened 2011-07-02
After refactoring CCAtlasNode author forgot to look corresponding files

moadib:
After refactoring CCAtlasNode forgot to rename corresponding variables in CCLabelAtlas and CCTileMapAtlas.


minggo:
Oh, Bin has refactored it. Thank you for you pull request.
BTW, we are upgrading the code to 1.0.0-r3, so the code is not stable.

PR #361, opened 2011-07-07
To resolve Bug #569: ChipmunkTest Sensors crashes in a few seconds

flyingpacer:
To resolve Bug #569: ChipmunkTest Sensors crashes in a few seconds


minggo:
Oh, we should not change the source of chipmunk.
It works ok on other platforms, only crashes on win32.
So, it is not the problem of chipmunk.

PR #373, opened 2011-07-08
commit some pkg files

flyingpacer:
commit some pkg files


flyingpacer:
to commit lua support

PR #385, opened 2011-07-12
fixed #602 and update lua support

flyingpacer:
fixed #602 and update lua support


flyingpacer:
reset the gl perspctive project

PR #426, opened 2011-07-27
fixed #630 To realize the buffer-saving function in CCRenderTexture

flyingpacer:
fixed #630 To realize the buffer-saving function in CCRenderTexture


flyingpacer:
to remove @ in ccassert

PR #429, opened 2011-07-28
add CCMenu::init() and CCMenu::menu() to create empty menus

leon-li:


minggo:
Why to add these functions to CCMenu.

leon-li:
Sometimes it is more logical to create the menu a one place and add items to it at another place.  It's more convenient to have these functions.

minggo:
I think you can use menuWithItem(NULL) and addChild() to achieve the logic.
All right?

leon-li:
Yes.  However, it is more convenient to have menu() and init().  And more importantly, the purpose is more clear.

minggo:
If so, I think the name can be node, not menu.
In fact, it hase node in cocos2d-iphone, because c++ should implement it in every class.

minggo:
And the node call init() in it, not add another method.

leon-li:
ok.  I will rename menu() to node()

leon-li:
Done.  Please check.

minggo:
Oh, I think you can add the logic of init() in node(), and remove the init().
I think less function is great.

leon-li:
The init() serves another purpose: to avoid putting the object into the auto release pool.

minggo:
I don't understand. Could you give me a example?

leon-li:
If you call CCMenu::node(), the created object is "autorelease"ed.  If you call" CCMenu \* menu = new CCMenu(); menu->init();", it is not "autorelease"ed. There is a slight performance gain by not putting it into the pool. Plus, our auto release pool is not thread safe at the moment.

minggo:
Yeap, you are right.
Thank you.

PR #432, opened 2011-07-28
Changes to support additional extensions + bugfix

reflog:
To support CCScrollLayer - made TargetedTouchHandler visible.
To support RadioMenu - itemForTouch made public
Added two useful macros for defining scene() and node() methods with params
Added support for optional default values in CCUserDefault getters
Corrected bug and leak in getBool

### DIFF
--- a/cocos2dx/support/CCUserDefault.cpp
+++ b/cocos2dx/support/CCUserDefault.cpp
@@ -155,10 +155,8 @@ bool CCUserDefault::getBoolForKey(const char* pKey, bool defaultValue)
 
 	if (value)
 	{
-		if (! strcmp(value, "true"))
-		{
-			ret = true;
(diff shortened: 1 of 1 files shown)


minggo:
+#define SCENE_FUNC_PARAM(**TYPE**,**PARAMTYPE**,**PARAM**) 
-  static cocos2d::CCScene\* scene(**PARAMTYPE** **PARAM**) \

I think it is better to rename scene() to node(). What do you think?

reflog:
I don't mind actually (I'm not big on naming conventions :) )
I just went with the naming that was used in the HelloWorld.cpp sample (from
the cocos2d-x templates).

minggo:
So, modify it plz.
Then,  I can merge it.

PR #451, opened 2011-08-04
Small fixes to the Android project templates

alinradut:
- fix for the 'create-android-project.sh throws a warning on OSX because of uname -o parameter' issue http://www.cocos2d-x.org/boards/10/topics/2808
- fix for the 'build_native is not executable by default' issue http://www.cocos2d-x.org/boards/10/topics/2807


alinradut:
Oops, I committed the create-android-project.sh file with the NDK_ROOT and ANDROID_NDK_ROOT defines commented out. I define them in the .bash_profile file so I didn't need them here.

PR #452, opened 2011-08-04
Small fixes to the Android project templates

alinradut:
Uncommented the path definitions.

### DIFF
--- a/create-android-project.sh
+++ b/create-android-project.sh
@@ -5,8 +5,8 @@
 # Don't modify the script until you know what you do.
 
 # set environment paramters
-#NDK_ROOT="/workspace/android-dev/android-ndk-r5"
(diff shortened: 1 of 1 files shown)


walzer:
Thanks! I will test it on cygwin and merge it tomorrow :)

minggo:
uname -s print the kernel name, in my cygwin, the output is:
CYGWIN_NT-5.1
So, can not just use -s to resolve it.

minggo:
I have tested the codes:

<pre>
KERNEL_NAME=`uname -s | grep "CYGWIN*"`
echo $KERNEL_NAME
if [ $KERNEL_NAME"hi" != "hi" ]; then
    echo "Error!!!"
    echo "Don't run in cygwin. You should run corresponding bat."
    exit 
fi
</pre>


It works well.
So I will merge your pull request and make a change.

PR #455, opened 2011-08-05
Lua AddSearchPath issue #647

giom:
As per the forum thread on http://www.cocos2d-x.org/boards/11/topics/2709
Here's the patch for AddSearchPath for Lua


walzer:
Maybe we should add a little sample code, or write the doxygen comments to CCScriptSupport methods. I will create issues for this

PR #462, opened 2011-08-10
issue #540, add RootViewController into xcode3 templates; rename _XXX_Pre

walzer:
issue #540, add RootViewController into xcode3 templates; rename _XXX_Prefix.pch to Prefix.pch; remove MIT license in user code.
build succeed in xcode4, but untested in xcode3


walzer:
I also add CCTexturePVR.h/.cpp, include/CCTouchHandler.h in xcode3 4 tempaltes

PR #468, opened 2011-08-15
Methods for AffineTransform now passed by reference

moadib:
Also, guys, please refactor your code to pass arguments by reference. It is very important for performance, not syntax sugar.


minggo:
Yeap, we should refactor it. We will do it asap.

moadib:
Applied this changes to https://github.com/cocos2d/cocos2d-x/pull/474

PR #480, opened 2011-08-19
CCMutableArray optimization

moadib:
Please look at this topic http://www.cocos2d-x.org/boards/6/topics/3164


minggo:
You can not use 
std::sort(pArray->begin(), pArray->end(), less);
in 
CCTouchDispatcher::rearrangeHandlers()
or it may crash. You can refer bug #680

minggo:
Oh, you use reserve(), so I think you are right. I will check the usage of insertObjectAtIndex(), and merge your pull request. Thank you.

PR #491, opened 2011-08-28
Feature #687 - add preloadBackgroundMusic and unloadBackgroundMusic to Android

augustorsouza:
These 3 commits are an attempt to add preloadBackgroundMusic and unloadBackgroundMusic to Android. It is a first version implementation. I think a good approach is to have an ArrayList of MediaPlayer objects in Cocos2dxMusic class, instead of just one MediaPlayer. But, we need to discuss which approach is best before implementing, for now I think this is a start.

It aims to implement feature #687: http://www.cocos2d-x.org/issues/687

### DIFF
--- a/CocosDenshion/android/SimpleAudioEngine.cpp
+++ b/CocosDenshion/android/SimpleAudioEngine.cpp
@@ -64,11 +64,6 @@ namespace CocosDenshion
         preloadBackgroundMusicJNI(pszFilePath);
 	}
 
-    void SimpleAudioEngine::unloadBackgroundMusic(const char* pszFilePath)
-	{
-        unloadBackgroundMusicJNI(pszFilePath);
-	}
-
 	void SimpleAudioEngine::playBackgroundMusic(const char* pszFilePath, bool bLoop)
 	{
         playBackgroundMusicJNI(pszFilePath, bLoop);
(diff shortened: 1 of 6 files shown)


minggo:
Is unloadBackgroundMusic needed? Because we only allow playing one background music at a time.

augustorsouza:
In the feature #687 description, it is requested. The way I implemented it is useless, but I think it is a good idea to let the code there and improve it in the next releases. If the ideia of having an ArrayList of MediaPlayer objects is accepted by you guys, the unloadBackgroundMusic is going to be really usefull.

minggo:
Cocos2d-iphone also supports playing one background music at a time, and I am not sure we need playing more than one background music, so I think unloadBackgroundMusic is not needed now.

But thank you again. Could you modify the code and repull the code?

augustorsouza:
I agree with you. Let's follow the cocos2d-iphone behavior! =)

unloadBackgroundMusic removed successfully! I will try help the project more from now on. It is a really interesting project. I will keep looking in the website to check for open issues or features, but send me an email every time you guys need help, ok?

Best regards...

minggo:
You are welcome to join us. What's your id in the forum? I will ask walzer to add you to the developer list, then you can modify the issue.
And what's your email? Which field are you familiar with?

minggo:
My email is ming@cocos2d-x.org. I will merge the pull request, and it will be closed. 
Connect me, man.

PR #511, opened 2011-09-18
Fix tolua bindings

ciaranj:
I struggled to add some new lua bindings for the particle system (not included in this patch.)  This patch adds an ant script that applies the changes suggested in the readme file and automatically overwrites the committed file.  You still need windows to use this, but may look at getting it running on posix tolua++ installations as well.

Interestingly the generated file was _very_ different to the one that was checked in, has that file been hand-crafted at all ?

Also the current .pkg files appeared out-of-sync with the current cocos2d-x source code, particularly around the CCAccelerometerDelegate class, I've also added a file that I believe to be correct. 

I'm hopefull that this will make it considerably easier for people to submit future lua bindings to the project :) 

Thanks!

### DIFF
--- a/tools/tolua++/README
+++ b/tools/tolua++/README
@@ -1,48 +1,19 @@
-1. The usage of tolua++
-    You can use the command to generate the LuaCocos2d.cpp:
-tolua++.exe -tCocos2d -o LuaCocos2d.cpp Cocos2d.pkg
+1. Generating the lua<-->C bindings with tolua++
 
-2. Modify the generated file
-After you run the command, the generated file under the
-directory the same as tolua++.exe. cocos2dx/lua_support/LuaCocos2d.cpp
-is generate by this way. But here are some differents between them, because
-tolua++.exe generates some error codes about the template. The errors are:
+    An ant script has been provided to generate the relevant files, to do this after
+    modifying the .pkg files you should use the following command in this directory:
 
--------------------------------------------------------------------------------------------
-               genereated code               |            correct code                    |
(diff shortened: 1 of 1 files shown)


minggo:
Thank you for your contribution, it is great.
I have 2 questions:
1. Could you add the usage of build.xml into ReadMe?
2. If the people use the build.xml to generate LuaCocos2d.cpp, does he need to modify the cpp file or he should do nothing?

ciaranj:
1) yup will do : 'ant' ;)
2) nope nothing to do to the file, it should just work!
I'll provide a new pull request shortly.

minggo:
I have read the README, thank you. I have a question, if people don't have ant programe, he may use tolua.exe to generate LuaCocos2d.cpp. So I think using ant is just one appoach, README should describe tow approach.

What do you think?

ciaranj:
Given how difficult manually maintaining that file is I would suggest discouraging manually editing it, ant forms part of the android development kit doesn't it? (I'm not an android developer so this is a queTion more than a statement :) )

I dont have a strong view on it though. Perhaps you could combine the readmes as you see fit if you merge?

minggo:
Yes, ant is part of android development kit, but many people aren't familiar with it. Then they will ask many questions, and we should do lots of work.

ciaranj:
Hmm, interesting. My experience is most people are familiar with ant.  Is there a tool you would prefer scripts in? I avoided bash specifically to avoid having to have all the other utilities present?

PR #515, opened 2011-09-19
Initial example working version of a lua project generator

ciaranj:
Hiyah.  This pull requests provides an example ant build script and associated template file for generating adhoc lua projects, it doesn't currently use 'HelloLua' (for purely historic/organic reasons) .. it would perhaps be better if it did..but it doesn't.. although I suspect it would be easy enough to re-point it at the HelloLua project.

What do you think?

### DIFF
--- a/tools/lua_project_generator/README
+++ b/tools/lua_project_generator/README
@@ -1,0 +1,14 @@
+This tool can be used to generate a 'jumping off point' for creating cross platform mobile games using the LUA bindings
+with Cocos2d-x. 
+
+Usage is simple: 
+
+  'ant' 
+
+You will be asked 3 questions.
+
+1) The path to your NDK (e.g. /cygdrive/c/my_first_ndk or /usr/bin/my_first_ndk )
+2) The 'Name' of your project (e.g. AwesomeMegaDeath)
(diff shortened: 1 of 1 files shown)


minggo:
Could you please write a README to describe the purpose of it, and the usage?

ciaranj:
Sure.  'ant' again :).  the file exposes its own description but I'll add a readme too :)

minggo:
Thank you.
Could you also write a document in http://www.cocos2d-x.org/projects/cocos2d-x/boards/11 to describe it and the lua binding. Then more people will notify it and share it.

ciaranj:
Okey-dokey :) will do

ciaranj:
Ok. This is done :)  And as requested, posted a link here: http://www.cocos2d-x.org/boards/11/topics/3957

minggo:
Does it copy all the files into tools/lua_project_generator/template?
If the cocos2dx modify it's code, can this template work ok?

ciaranj:
No, sadly at the minute it doesn't.  It is separate from HelloLua (HelloLua had 'too much').   I think that that would be a sensible move forward at some point though ?

minggo:
Yes, thank you.

PR #522, opened 2011-09-22
Linux port

laschweinski:
1. delete lib folder, now no any moving is applied.
2. cocos2dx_loc is not required to be assigned now
3. modify mycreateAndroid.sh to create new android and linux integrated project correctly
4. move mycopyfile.sh from template/android to template/linux and do some modifications


minggo:
Too mane files changed. It is not needed.

minggo:
Too mane files changed. It is not needed.

minggo:
Too mane files changed. It is not needed.

PR #523, opened 2011-09-22
OpenGLES2.0 support

stoulouse:
Hello, here is a first attempt to support OGLES2.0 using backward compatibility shaders.
It works on iOS with some bugs because sometime GL direct mode is used, but it globally works well.
I think Android and other platform should be easy to do because he code is cross platform.
I'll continue to work on this to make Android port available when i'll have time.


minggo:
Actually, it is awksome. It is an amazing work.

cocos2d-iphone is developing on 2.0, which will support GLES2.0. It is a new project, not based on 1.0.
We will develop it ASAP. So, I am welcome you to contribute for the 2.0 project in the future.

I don't know how to deal with your pull request. It is a good job, but I can not merged it into master. Because we
are compatible with cocos2d-iphone. There are tow ways to do it:
1. I create a new branch for it, and you pull the request to the new branch, may be named gles2.0.
2. You pull request to the 2.0 project after we doing it.
What do you think?

stoulouse:
Hi minggo, i think you can create a branch named gles2.0 where i can pull the request.
I've already did that job for cocos2d-iphone at the beginning of this year but they didn't add it to their repository because they had other things to do. Now i don't use cocos2d-iphone anymore because i need cross platform and there is no point for me to use an objc only code.
My point is I want this code to be supported because I use cocos inside my own 3d engine which support both ogles1 and 2. I think you can add this to master branch because it does not break gles1 code and stay close to cocos2d-iphone base code, but if you prefer not to do that, a branch for everyone is not a bad idea, i can also maintain my own repository but i'd like to have more people working on it because i can't work on other platform like airplay, windows or wophone.
I also don't want my code lost like it was the case for cocos2d-iphone.
So do what you think is best for your project. Cheers, ++samuel

minggo:
Because we are releasing a stable version these days. So I decide to new a branch named gles2.0, and merge the codes into it. Then we merge it  into master in next version. What do you think?

Thank you again for your awksome work and your sharing.

stoulouse:
i think it's a great idea. How will i be able to update the branch code ? fork+pull again ?

minggo:
You can repull it to the new branch. I create it just now. Then I will notify you.

minggo:
Ok. I have created gles2.0. You can close this pull request and repull it to gles2.0.

stoulouse:
thank you very much

PR #525, opened 2011-09-23
Gles2.0

stoulouse:
here it is.


laschweinski:
shader shader shader I love shader....thanks...

PR #526, opened 2011-09-23
SolutionDir removal 

ciaranj:
This patch changes around the build-deps of the visual studio projects so that rather than the libraries _pushing_ their outputs into the calling solutions (if a solution is used), the solution references them instead,  what do you think ?

(also see Issue 399 where someone else has run into this problem)

### DIFF
--- a/Box2D/proj.win32/Box2D.win32.vcxproj
+++ b/Box2D/proj.win32/Box2D.win32.vcxproj
@@ -37,10 +37,10 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup>
     <_ProjectFileVersion>10.0.30319.1</_ProjectFileVersion>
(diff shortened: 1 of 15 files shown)


minggo:
We pull all libs into the solutions, because
1. The app can be published easily, because all libs are there.
2. The template of vs need the prebuilt libs, and they find the libs in the solutions directory Debug.win32 or Release.win32.

ciaranj:
Sure of course, if you look this behaviour (1) is maintained in the pull request, previously it relied on a mixture of copy and the knowledge that dependent projects would write into the solution folder. Now only 1 approach is used, the copy for release & publish.

I'm not sure what you mean about the templates but I'll take a look.

minggo:
Sorry.
We are pulishing a new stable version. So, I will try your pull request in the next version.
Thank you.

minggo:
What's the benefit of this pull request? Could you explain more detail?

PR #537, opened 2011-10-05
Added a helper method to draw filled polygons

reflog:


moadib:
It is not suitable for non-closed polygon, since it may be simple segmented line.

PR #541, opened 2011-10-10
Gles2.0

stoulouse:
gles2.0 working on iOS and android

### DIFF
--- a/cocos2dx/Android.mk
+++ b/cocos2dx/Android.mk
@@ -144,6 +144,7 @@ LOCAL_LDLIBS := -L$(call host-path, $(LOCAL_PATH)/platform/third_party/android/l
                  -ljpeg
 
 # define the macro to compile through support/zip_support/ioapi.c                
-LOCAL_CFLAGS := -DUSE_FILE32API -DDEBUG -DCOCOS2D_DEBUG=1
+LOCAL_CFLAGS := -DUSE_FILE32API
+#LOCAL_CFLAGS := -DUSE_FILE32API -DDEBUG -DCOCOS2D_DEBUG=1
                                  
 include $(BUILD_SHARED_LIBRARY)


minggo:
Thank you, I have some questions:
1. You change target=android-8 to target=android-10, does it only work with andorid2.3? 
Cocos2d-x should support Android from 2.1.
1. -DCOCOS2D_DEBUG=1
   COCOS2D_DEBUG, what does the micro used for?
2. /\* Setup the context factory for 2.0 rendering.
        \* See ContextFactory class definition below  
        */
       setEGLContextFactory(new ContextFactory());

If I want to use gles1.1, does these codes work ok? I mean, is it compatible with gles1.1?

stoulouse:
hello,
target and COCOS2D_DEBUG are bad merge from an other project.
i'm new on android and i don't know how to select the best configuration and check if there is gles2.0 available or not. the code is compatible with gles1.1 but i only need to check if there is gles2.0 and if it's not the case tell it to native code.
also there is one more question which i don't how to solve: on gles2.0 there is shaders and i don't know how to copy then to assets on android with your templates. thank you for your help. ++samuel

minggo:
So I think you should not modify target and remove -DDEBUG and -DCOCOS2D_DEBUG.
The template will generate build_native.sh at $PROJECT_ROOT/android. The script will copy all files under Resources into assets.

stoulouse:
I know about the Resources copy rule, but I think shaders should not be mixed with the other assets so that users don't mix up with it. That's why I think there should be a rule that copy shaders from cocos2dx/support/gl_support/OpenGLES/OpenGLES20/shaders in assets folder silently. What do you think?

stoulouse:
here is your request commit

minggo:
Thank you, I think you should also remove -DDEBUG -DCOCOS2D_DEBUG=1 in cocos2dx/Andorid.mk.
I don't have good idea about how to copy shaders now. 

One more question. You said:
check if there is gles2.0 and if it's not the case tell it to native code.
But I don't find where to tell native code.

stoulouse:
you don't have to tell native but set accordingly from java in this function :
        public EGLContext createContext(EGL10 egl, EGLDisplay display, EGLConfig eglConfig) {
            Log.w(TAG, "creating OpenGL ES context");
            checkEglError("Before eglCreateContext", egl);
            int[] attrib_list = {EGL_CONTEXT_CLIENT_VERSION, 2, EGL10.EGL_NONE };
            EGLContext context = egl.eglCreateContext(display, eglConfig, EGL10.EGL_NO_CONTEXT, attrib_list);
            checkEglError("After eglCreateContext", egl);
            return context;
        }
you should replace int[] attrib_list = {EGL_CONTEXT_CLIENT_VERSION, 2, EGL10.EGL_NONE }; by int[] attrib_list = {EGL_CONTEXT_CLIENT_VERSION, 1, EGL10.EGL_NONE }; for gles1.1.
Maybe the correct way to do it is by first trying to find a config using int[] attrib_list = {EGL_CONTEXT_CLIENT_VERSION, 2, EGL10.EGL_NONE }; and if it return null try with int[] attrib_list = {EGL_CONTEXT_CLIENT_VERSION, 1, EGL10.EGL_NONE };

is it the right way to create the best gles context???

minggo:
Yes, I think you should consider the error of gles context initialization. Try 2.0 first, if error, than try 1.1.

I have a question. If java can not create gles2.0 context, but c++ use gles2.0, then what will happen?

minggo:
How about the progress?

stoulouse:
Hi, sorry for my late reply, how is the project going? I'm currently using gles-2.0 branch in 2/3 iPhone/android projects now so i guess it's pretty stable (for my cases).

PR #545, opened 2011-10-12
More simple and correct way to detect language for win32 application, als

moadib:
More simple and correct way to detect language for win32 application, also added detection of all languages contained in ccLanguageType enumeration .

Added support for detect russian language.

Take a look :)


walzer:
Smart! Why I haven't found this way.

PR #556, opened 2011-10-19
Fix compile error on 64 bit Linux.  Fix some warnings.

leon-li:


minggo:
#STATICLIBS = $(STATICLIBS_DIR)/libfreetype.a \

# $(STATICLIBS_DIR)/libcurl.a \

# $(STATICLIBS_DIR)/libxml2.a \

# $(STATICLIBS_DIR)/libpng.a \

# $(STATICLIBS_DIR)/libjpeg.a

Why comment these lines? Doesn't it cause error?

leon-li:
Those static libs were compiled on 32 bit system.  Cause linking errors on 64 bit system.

minggo:
Oh, it is the problem. Could you offer 64 bit version?
And should modify the makefile to link correct libs.

leon-li:
Do we really need to bundle those libraries for Linux?  For any Linux distribution, it is very trivial to install those libs.

minggo:
You comment the link, then it can not work on 32 bit and 64 bit.

leon-li:
I will address this a bit later.

tangzero:
I have the static libs for 64 bits systems: http://db.tt/0TFvzjLb

minggo:
tangzero, have you tested on 64-bits linux?

leon-li:
Hi tangzero,
Do you have 64 bit libmodex for CocosDenshion?

leon-li:
I have bundled the 64 bit libs provided by tangzero.  Please help me do a test?  CocosDenshion does not work at the moment because of the lack of 64 bit libfmodex.

tangzero:
I tested on Ubuntu 11.10 64bit.

minggo:
Thank you.

PR #562, opened 2011-10-20
fixed #756 CCMenu can't be initialized by empty menu

flyingpacer:
fixed #756 CCMenu can't be initialized by empty menu


minggo:
The code may not be compiled on  Android.

minggo:
The mangling of 'va_list' has changed in GCC 4.4. It is a struct, not a void*.
And NDK use gcc4.4. So can not fixed it like this.

minggo:
So I have to close the pull request, the error built by NDK toolchain is:
error: cannot convert 'int' to 'va_list' in initialization

PR #564, opened 2011-10-22
Purge KeypadDispatcher in destructor incorrect, must be called in purgeDi

moadib:
Purge KeypadDispatcher in destructor incorrect, must be called in purgeDirector() method.

I'm faced with situation, when CCKeypadDispatcher holds last reference to Layer, which holds reference to Sprite when Texture already destroyed.

This fix it. Take a look at situation please.


moadib:
Read this:

http://www.cocos2d-x.org/boards/12/topics/4738

minggo:
I will check it.
Thank you.

moadib:
No, it's not correct :\

PR #565, opened 2011-10-23
Makefiles cleanup

Lucretia:
Hi, can you take a look at these changes and possibly implement them in the master?

I modified the makefile's because I'm on 64-bit and using the native libs was better for me.

### DIFF
--- a/Box2D/proj.linux/Makefile
+++ b/Box2D/proj.linux/Makefile
@@ -1,62 +1,58 @@
-CC      = gcc
-CXX     = g++
-AR      = ar
-TARGET	= libbox2d.a
-CCFLAGS = -Wall -g -O2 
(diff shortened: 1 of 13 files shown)


minggo:
Thank you.
I think int is big enough, because it is used as the index, not used for address.

Lucretia:
Int is not big enough as it fails the build on a 64-bit machine.

minggo:
Linux port is used for developing, so I don't think it is valuable to modify logic code of engine to make it work on 64-bit.

Lucretia:
I do, I'm using it to develop my game and then I test on Android (and later iOS). So, I really want this to work properly.

I also find it strange that the OGLES libs aren't used on Linux like they are on the Win32 version.

PR #567, opened 2011-11-01
Fix compile error under ubuntu 11.10

leon-li:
Changed all NULL to 0

### DIFF
--- a/doxygen/doxygen.cocos2d-x
+++ b/doxygen/doxygen.cocos2d-x
@@ -1457,7 +1457,9 @@ INCLUDE_FILE_PATTERNS  =
 # instead of the = operator.
 
 PREDEFINED             = "CC_PROPERTY_READONLY(varType, varName, funName)=protected: varType varName;\n/** get##funName*/\n public: virtual varType get##funName(void);" \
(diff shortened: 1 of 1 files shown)


minggo:
NULL cause compile error on ubuntu 11.10?

leon-li:
Yes.  Can also be fixed by "include <cstddef>"

moadib:
Or even
`#ifndef NULL
#define NULL 0 
#endif`
No need so many changes as requested.

tangzero:
"include <cstddef>" works fine.

minggo:
I agree with tangzero. And bug 780 had been created for it some days ago.
http://www.cocos2d-x.org/issues/780

leon-li:
I knew we can solve this by defining NULL or including the proper headers.

But what are the drawbacks of using "0" directly?  You will never have this problem if "0" is used directly.   And how "NULL" should be defined is an debatable topic.  The meaning of a "0" pointer should be very clear to every C/C++ programmer.  Until we have "nullptr" from C++ standard, I think using "0" is actually simpler and easier to understand.  Changing "NULL" to "0" was just minutes of search and replace.

minggo:
Ok. You win. It is not a big issue.

PR #570, opened 2011-11-14
Newest

dumganhar:
add bada port


minggo:
github don't show the diff of many files, so I can not review the codes.
BTW, the pull request cannot be automatically merged.

PR #571, opened 2011-11-15
fix memory-leak for lua engine

jbyu:
add destructor for luaEngine
add void\* type to fix the problem while calling CCCallFuncND in lua script


minggo:
getLastValidIterator() may return a invalid iterator  m_array.end().

PR #607, opened 2011-12-06
* fix Xcode 4 template, AppController.mm not call AppDelegate:applicationDidEnterBackground() & AppDelegate:applicationWillEnterForeground()

dualface:


walzer:
dualface, I love your gravatar, so cool it is :)

PR #608, opened 2011-12-07
Update CCTouch to include the ID/PID of the touch

GMTDev:
As discussed here by Gavt: http://www.cocos2d-x.org/boards/6/topics/4043?r=5751#message-5751

"it is needed for a case like this: Say you are moving three sprites around the screen using three fingers, I receive a CCSet with three CCTouch items in it (one CCTouch for each finger). Then I take a finger off the screen so I have two left, I now receive a CCSet with two CCTouch items (two pairs of coordinates), I can't tell which finger disappeared! I could "guess" which finger disappeared by checking old coordinates but I might get it wrong if for example if all the fingers were close to each other when one was removed.

At the moment I am using the address of the "CCTouch*" to determine which finger disappeared. This works due to the code in Cocos2d's Eagleview.mm but I think this could break in a future version of Cocos2D-X. Ideally the real UITouch's ID should be passed onto the CCTouch class."

### DIFF
--- a/cocos2dx/include/CCTouch.h
+++ b/cocos2dx/include/CCTouch.h
@@ -41,7 +41,7 @@ public:
     int view() { return m_nViewId; }
     int id(){ return m_iID; }
 
-    void SetTouchInfo(int nViewId, float x, float y, int iID)
+    void SetTouchInfo(int nViewId, float x, float y, int iID=0)
     {
         m_nViewId   = nViewId;
         m_prevPoint = m_point;


minggo:
Could you please modify android too?

GMTDev:
OK, I think that is all that is needed for the Android version, compiles fine but I can't test it (having trouble getting Android cocos2d-x apps to run on my laptop!).

minggo:
I think void
SetTouchInfo(int nViewId, float x, float y, int iID) 
should change to
void SetTouchInfo(int nViewId, float x, float y, int iID = 0)

Then it can keep compatible with old version, and some platforms such as win32 don't support multitouch.
What about your opinion?

GMTDev:
Yeah that's a good idea.

minggo:
Could you please modify the code?
Then I can merge it.

Thank you.

GMTDev:
OK, done.

PR #619, opened 2011-12-14
Removing statically hard-coded Resource Path for QNX on CocosDenshion and some fix on QNX branch

lynxluna:
Hello,

I remove the statically hard-coded Resource Path for CocosDenshion for QNX. It's hard-coded to "app/native/Resources", therefore we can't play sound outside that path (eg. from shared directories). It's better to use CCFileUtils to get full path of sound files outside CocosDenshion Module. I also fix the cocos2d-x's proj.qnx for accidentally includes marmalade.

### DIFF
--- a/tests/tests/CocosDenshionTest/CocosDenshionTest.cpp
+++ b/tests/tests/CocosDenshionTest/CocosDenshionTest.cpp
@@ -5,6 +5,8 @@
 // android effect only support ogg
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
 	#define EFFECT_FILE		"effect2.ogg"
+#elif( CC_TARGET_PLATFORM == CC_PLATFORM_MARMALADE)
+	#define EFFECT_FILE		"effect1.raw"
 #else
 	#define EFFECT_FILE		"effect1.wav"
 #endif // CC_PLATFORM_ANDROID
@@ -54,7 +56,7 @@ m_nSoundId(0)
 
 	for (int i = 0; i < m_nTestCount; ++i)
 	{
-#if (CC_TARGET_PLATFORM == CC_PLATFORM_AIRPLAY)
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_MARMALADE)
 		CCLabelBMFont* label = CCLabelBMFont::labelWithString(testItems[i].c_str(),  "fonts/arial16.fnt");
 #else
         CCLabelTTF* label = CCLabelTTF::labelWithString(testItems[i].c_str(), "Arial", 24);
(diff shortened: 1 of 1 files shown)


dumganhar:
Thanks for your pull request, I have tested it , but the CocosDenshionTest will not work according to your code. Could you fix it?

lynxluna:
Where are you testing the code? at device or sim? In the device, it runs perfectly.

dumganhar:
I tested it on playbook device.Any idea?

lynxluna:
I think I found out the culprit, in the file CCFileUtils_qnx.cpp, it simply returns the relative path instead of current executable path.

Therefore CCFileUtils::fullPathFromRelativePath returns the same relative path (pszRelativePath)

```
const char* CCFileUtils::fullPathFromRelativePath(const char *pszRelativePath)
{
    return pszRelativePath;
}
```

The hard-coded "./app/native/Resource path on CocosDenshion on QNX is to overcome this. I'll try to put the implementation for getting full path using s_strResourcePath

dumganhar:
Okay, waiting for your good news. Thanks!

lynxluna:
I put the path for the sounds on the test since I tried to implement 'correct' way by returning absolute path on the CCFileUtils::fullPathFromRelativePath but it breaks the CCTextureAtlas. I think this is the only way rather than hard-coding the path on the CocosDenshion. Programmers can specify path freely on the BAR file.

I think I will search for better way later.

dumganhar:
Oh, my god, I forgot to add .cproject file by 'git add -f', Thanks for reminding. 
Your modification in CocosDenshionTest will be merged tomorrow.
And I hope you can find the way to get full path in 'CCFileUtils::fullPathFromRelativePath' function.

lynxluna:
Okay, looks like the path issue on QNX has been resolved, I restore previous CocosDenshion Test to the Original One. It runs very well and can run all tests on my Playbook smoothly. If you want to merge, merge this one instead.

CocosDenshion/qnx/SimpleAudioEngine.cpp
```diff
@@ -71,7 +71,6 @@ of this software and associated documentation files (the "Software"), to deal
 	static strm_dict_t 		  *s_volumeDictionary = 0;
 
 	static SimpleAudioEngine  *s_engine = 0;
-	static string s_strResourcePath = "./app/native/Resource/";
```
lynxluna, line 74: Cannot use this, since we have other directories besides ./app [as described here](https://bdsc.webapps.blackberry.com/native/documentation/com.qnx.doc.native_sdk.devguide/com.qnx.doc.native_sdk.devguide/topic/accessible_folders.html). Remove all references to this hard-coded directory.

cocos2dx/platform/qnx/CCFileUtils_qnx.cpp
```diff
@@ -101,7 +117,7 @@ void CCFileUtils::setResourcePath(const char *pszResourcePath)
 	{
 		std::string title = "Notification";
 		std::string msg = "Get data from file(";
-		msg.append(pszFileName);
+		msg.append(full_path);
```
lynxluna, line 120: Showing full path rather than abbreviated file name

dumganhar:
Hi, lynxluna
I notice your commit 'Fixing Warning on CLang 3.0 on XCode 4.2', and you modify some global functions to static global inline functions, but the 'get...' is added with inline, and 'set...'not be added? Why not add it all ?

lynxluna:
Actually that's my preference of not inlining long function. I know the compiler is smart enough on deciding which one needed for inlining.

So, what do you think? Should we inline all those?

dumganhar:
Good habit, Okay, we will keep your modification.

dumganhar:
I have tested it , CocosDenshionTest can work now, but when I click Mainmenu button, The application will crash! Would you get a crash ?

lynxluna:
It is not consistent, sometimes it hangs, sometimes it is not. I suspect it is behaviour of the [QNX Multimedia Renderer](https://bdsc.webapps.blackberry.com/native/reference/com.qnx.doc.mme.mmrenderer/topic/arch.html) I'll investigate tonight.

dumganhar:
Okay, thanks a lot.

PR #646, opened 2011-12-28
Change win32 from GLES to OpenGL

honghui:
Using the GLEW OpenGL Extension and GLFW toolkit, like Linux version do.


honghui:
wait a minute,I forgot the vs2010 solution.

moadib:
Thanks :) Nice addition.

PR #656, opened 2012-01-04
support pvr cache in VolatileTexture Class and TextureCache

sapsaldog:
In android, cocos2dx has some problem pvr texture,

It works okay, before push home button,

I added some code in CCTextureCache.cpp&h 

It may work well, I tested on three android phones, Motorola droid, galaxy S1, S2

you should ignore other files except CCTextureCache.cpp&h  (sorry for this)

enjoy your work, thanks


minggo:
I don't know how to ignore other files when merging your pull request.
BTW: This pull request cannot be automatically merged, you should synchronize codes.

walzer:
This bug was fixed in http://www.cocos2d-x.org/issues/967 at 5 months ago.

PR #667, opened 2012-01-11
Modify some files for qnx platform

dumganhar:
Modification:
1.rename create_**__project.sh to create-**_-project.sh
2.modify helloworld project setting and use new icon for qnx project.


jonyrock:
Hi guys. 
Looks like you guys forget to fix it in the ios template https://github.com/cocos2d/cocos2d-x/blob/v3/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm#L99

And I think that it is a bad idea to call it twice. You can check that 

``` C++
- (void)applicationWillResignActive:(UIApplication *)application {
    /*
     Sent when the application is about to move from active to inactive state. This can occur for certain types of temporary interruptions (such as an incoming phone call or SMS message) or when the user quits the application and it begins the transition to the background state.
     Use this method to pause ongoing tasks, disable timers, and throttle down OpenGL ES frame rates. Games should use this method to pause the game.
     */
     //We don't need to call this method any more. It will interupt user defined game pause&resume logic
    cocos2d::Application::getInstance()->applicationDidEnterBackground();
}

- (void)applicationDidBecomeActive:(UIApplication *)application {
    /*
     Restart any tasks that were paused (or not yet started) while the application was inactive. If the application was previously in the background, optionally refresh the user interface.
     */
     //We don't need to call this method any more. It will interupt user defined game pause&resume logic
    /* cocos2d::Director::getInstance()->resume(); */
    cocos2d::Application::getInstance()->applicationWillEnterForeground();
}
```

I think that it's right to call `applicationDidEnterBackground` ONLY in `applicationWillResignActive`
and `applicationWillEnterForeground` only in `applicationDidBecomeActive`

PR #668, opened 2012-01-11
make CCTouchDelegate::touchDelegateRetain() and touchDelegateRelease as pure virtual function

minggo:


minggo:
add icons for wophone

PR #682, opened 2012-01-15
Added Marmalade support to TextInputTest and HelloLua samples

gzito:


dumganhar:
Hi,gzito
I review your code, and found that tests can't be run correctly. Nothing was shown except a close button and a black background. I tested it on windows platform! But the HelloLua and HelloWorld worked fine. By the way, could you please delete the comment such as 
//#if (CC_TARGET_PLATFORM != CC_PLATFORM_MARMALADE)
#include "CCUserDefault.h"
//#endif
Thanks!

Issue #626, opened 2011-12-19
Compiling with ndk-r7 still not working


Bahamut:
Under windows, i can finally compile using ndk-r7 withouth problems (just some warnings) but the APK generated is not OK... at least i always get black screens with android 2.1 and 2.2 emulators (works with android 4 emulators).

This doens't happens with ndk-r6b.
I dont know about mac or unix, i just use cygwing, last cocos2dx stable and latest ndk r7.

sapsaldog:
When I compiled using by ndk-r7 in mac, It works perfect.

but actually I didnt tested on emulators, just tested on devices

Bahamut:
It should work with emulators, i don't know why it shoudlnt...

PR #691, opened 2012-01-30
Fix bug #626

winbinar:


minggo:
Bug #626 is created in 2011-07-21, cocos2d-x didn't integrate pthread  that time.

PR #693, opened 2012-01-30
Use modules in the Android NDK build

folecr:
Usage of Android modules to define shared and static libraries
- Simplifies dependency management
- Cleans up makefiles for NDK
- Possible to easily declare static and shared library builds and allow projects that use this to use either one of the two.

Note : please discuss these changes. Do not pull yet! 
This will break android test and demo code (such as HelloWorld) until the test/demo project build files are re-written to use the NDK modules.

### DIFF
--- a/cocos2dx/Android.mk
+++ b/cocos2dx/Android.mk
@@ -4,7 +4,7 @@ include $(CLEAR_VARS)
 
 LOCAL_MODULE := cocos2dx_shared
 
-LOCAL_MODULE_FILENAME := cocos2d
+LOCAL_MODULE_FILENAME := libcocos2d
 
 LOCAL_SRC_FILES := \
 CCConfiguration.cpp \


minggo:
Yes, it will simplify dependency management, but it require developers to know more feature about grammar of Android.mk.

folecr:
Developers of Cocos2D-x : 
- can easily compile external libraries from source (like png, jpeg) with optimization/modification
- will have simple dependency management

Users of Cocos2D-x :
- Will have an Android compliant build file
- Can easily include Cocos2D-x in their projects

I think it will have a positive effect for Cocos2D-x on Android.

minggo:
Yes, you are right.
I think we can merge your pull request.
Could you please modify tests and other demo?

folecr:
OK. I will resubmit pull request after modifications.

minggo:
Thank you.

minggo:
Where is NDK_MODULE_PATH defined?
When I compile HelloWorld with the pull request, I met the problem:

Android NDK: /cygdrive/e/cocos2d-x/tests/test.android/jni/../../../cocos2dx/Andr
oid.mk: Cannot find module with tag 'cocos2dx/platform/third_party/android/modul
es/libpng' in import path
Android NDK: Are you sure your NDK_MODULE_PATH variable is properly defined ?

folecr:
I run using a script from the command line. The script looks like this :

```
GAME_ENGINE_DIR=$1
echo "Game Engine Dir = " $GAME_ENGINE_DIR

/opt/android/android-ndk/ndk-build \
    NDK_PROJECT_PATH=${GAME_ENGINE_DIR}/testbed-android \
    NDK_MODULE_PATH=/Users/folecr/ws/open/Cocos/cocos2d-x \
    NDK_DEBUG=1 \
    V=1 \
&& \
ant debug
```

folecr:
Changes in progress... I have been trying to get a pure C++ demo of Cocos2D working. (ie no JNI code.) I think this is possible with the exception of the TextInput portions.

BTW, the test, HelloWorld and implementation of Cocos2D is all mixed up - for example Android Cocos2D depends on code in HelloWorld without which Android Cocos2D will not run. This needs to be cleaned up too. I don't know if I will do that now because we need to discuss a little bit about how to organize source code before making these changes.

minggo:
Yeap, cocos2d is a lib, it can not run by itself.
What is "NDK_MODULE_PATH"?
Is it an key word of Android.mk, like LOCAL_SRC_FILES?

minggo:
I have done the other work in https://github.com/cocos2d/cocos2d-x/pull/712.
Now, HelloWorld and other demos work correctly.
Thank you.

folecr:
Thanks for the merge!

minggo:
Because RTTI(dynamic_cast) may cause error on android, refer http://gcc.gnu.org/faq.html#dso for more detail information.
So I have to build cocos2dx as a static library.
Then I can not use LOCAL_STATIC_LIBRARIES in cocos2dx/Android.mk.
And I have to add include path like previous in cocos2dx/Android.mk, and use module that uses libcocos2dx.a.

Is there any good mothed to use module when building a static library?

Thank you.

folecr:
I think we found the same bug :) I was trying to track this down yesterday.

For cocos2d-x we you can change line 132 in cocos2d-x / cocos2dx / Android.mk

> include $(BUILD_SHARED_LIBRARY)

to 

> include $(BUILD_STATIC_LIBRARY)

folecr:
I will submit a new pull request. Let's discuss there.

folecr:
https://github.com/cocos2d/cocos2d-x/pull/715

PR #695, opened 2012-01-31
Implementation to detect when the application goes and return from background in QNX (Black Berry Playbook)

GermanMtz:
applicationWillEnterForeground will be called when the device returns from idle mode and when the application brings the application to fullscreen after being in the background.

And applicationDidEnterBackground will be called when the device goes to idle mode and when the OS send the application to the carousel of application displaying them with thumbnail size.

### DIFF
--- a/cocos2dx/platform/qnx/CCEGLView_qnx.cpp
+++ b/cocos2dx/platform/qnx/CCEGLView_qnx.cpp
@@ -85,7 +85,8 @@ static CCEGLView* s_pInstance = NULL;
 CCEGLView::CCEGLView()
 : m_pDelegate(NULL),
   m_fScreenScaleFactor(1.0),
-  m_bNotHVGA(false)
+  m_bNotHVGA(false),
+  m_isWindowActive(false)
 {
 	s_pInstance = this;
 	m_eglDisplay = EGL_NO_DISPLAY;
@@ -844,7 +845,6 @@ bool CCEGLView::HandleEvents()
 
 		if (domain == navigator_get_domain())
 		{
-			const bool isPaused = CCDirector::sharedDirector()->isPaused();
 			switch (bps_event_get_code(event))
 			{
 				case NAVIGATOR_SWIPE_DOWN:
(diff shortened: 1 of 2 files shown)


dumganhar:
Hi, GermanMtz,
dducharme who works in rim company has done a 'pull request' to us, and we merge his code, I am sorry I can not merge your code automatically by github. So could you please pull the latest code and do a pull request again?
BTW, I notice the code you modify as follow:

<pre>
                case NAVIGATOR_WINDOW_INACTIVE:
                    CCApplication::sharedApplication().applicationDidEnterBackground();
                    break;

                case NAVIGATOR_WINDOW_ACTIVE:
                    CCApplication::sharedApplication().applicationWillEnterForeground();
                    break;
                case NAVIGATOR_WINDOW_STATE:
                {
                    switch(navigator_event_get_window_state(event))
                    {
                    case NAVIGATOR_WINDOW_FULLSCREEN:
                        CCApplication::sharedApplication().applicationWillEnterForeground();
                        break;
                    case NAVIGATOR_WINDOW_THUMBNAIL:
                        CCApplication::sharedApplication().applicationDidEnterBackground();
                        break;
                    }
                    break;
                }
</pre>


applicationWillEnterForeground will be invoked twice when application is launched, one is from  NAVIGATOR_WINDOW_ACTIVE and another is from NAVIGATOR_WINDOW_FULLSCREEN? Any good way to avoid this?

Thanks very much!

GermanMtz:
Hi:

I have already fetched the new changes and updated the fix to invoke the functions 'applicationWillEnterForeground' and 'applicationDidEnterBackground' to be called only once per time.

Any other comment please let me know 
And have an excellent day

dumganhar:
I don't think using the isPaused method of CCDirector is a good way.
Because it depends on whether user invokes 'CCDirector::sharedDirector()->pause();' in the 'void AppDelegate::applicationDidEnterBackground()' . I think it's a good way by just using a member variable of CCEGLView to control this logic. Do you think so ?

GermanMtz:
Yes, of course, in fact in the previous commit I was in doubt of which approach to use.

And just to confirm before to commit the change, which which should be the correct behavior of cocos2dx the method 'applicationWillEnterForeground' should be called right after the application starts or it should not be called when launching?

dumganhar:
Hi, You should invoke applicationWillEnterForeground if application receives NAVIGATOR_WINDOW_ACTIVE message when launching.

GermanMtz:
Done, the changes have been committed, and sorry for the inconvenience

dumganhar:
Thanks for your contribution!

PR #697, opened 2012-01-31
Updating the following:

dducharme:
CocosDenshion - now using OpenAL exclusively for audio, added ogg/vorbis support there
Accelerometer - fixing the accelerometer for different initial orientations of the device
Application - paying attention to the locale of the device and setting the current language accordingly
CCEGLView - paying attention to navigator events (menu, application go to background, application enters foreground), fixed up orientation code, fixed up screen event pointer code


minggo:
There is an ogg file in tests/Res, could you use it and not add background.ogg?
I think we it it better to keep cocos2d-x as small size as possible.

dducharme:
The cocosdenshion test would then not play any sound for the background.  I guess we could change it to some other wav file that could act as a background?

dumganhar:
Hi, dducharme,
Is mp3 format not supported in OpenAL? Does it just support wav format? I notice ogg file need to be decoded to pcm buffer.

dducharme:
It doesn't.  OpenAL only supports pcm audio. 

----- Original Message -----

PR #698, opened 2012-02-01
* add CCNode::getChildrenCount()

dualface:
- add CCNode::getChildrenCount()


minggo:
There is a function CCNode::getChildren(), why should add this function?

dualface:
CCNode::m_pChildren default is NULL, after first time call CCNode::addChild(), m_pChildren is valid.

CCNode::getChildrenCount() always check m_pChildren.

minggo:
You should check the result of CCNode::getChildren() to compute count every time you want to get children count.
We should keep engine as simple as possible.

dualface:
cancel

PR #699, opened 2012-02-01
* add CCNode::getPositionX/Y, CCNode::setPositionX/Y

dualface:
1. code less :)
2. In Lua, pass CCPoint object from/to C++ very slow. A number faster than xx time.


dualface:
cancel

PR #700, opened 2012-02-01
a solution about bug #942

leopoldchen:
#942 Invoke CCScheduler:sharedScheduler():scheduleScriptFunc in lua may cause crashion.

this solution is quite similar to the original unscheduleSelector. 
use a flag to mark if we want to delete the scriptfunc entry.
and in the unscheduleScriptFunc function, we only set this flag (currentTimerSalvaged) as true, then in the tick function, we delete it after timer update.

file list:
CCScheduler.h/cpp
&test code is in hello.lua


minggo:
Thank you.
But please not add these files:
cocos2d-win32.vc2010.sdf
ipch/cocos2d-win32-5f8e8667/libcocos2d-f1ffc4f1.ipch
ipch/cocos2d-win32-5f8e8667/libcocos2d-f1ffc4f1.ipch

leopoldchen:
I delete these files now

minggo:
Thank you.
Because pull request: https://github.com/cocos2d/cocos2d-x/pull/706 refactor the lua supporting, and fix this bug.
So, I think I should close the pull request.
Sorry.

leopoldchen:
I see, is that a cool job done by YAOGUAI?
waiting for it!!!

minggo:
Yes, it is.
He has developed with lua for a few games.
He is an expert of lua.

PR #705, opened 2012-02-03
Alignment support for CCLabelBMFont

irwinb:
I ported over [pingpongboss's](https://github.com/pingpongboss/CCLabelBMFontMultiline) awesome addition to CCLabelBMFont to cocos2d-x.

No tests yet..

### DIFF
--- a/cocos2dx/include/CCLabelBMFont.h
+++ b/cocos2dx/include/CCLabelBMFont.h
@@ -157,11 +157,13 @@ namespace cocos2d{
 		CCBMFontConfiguration *m_pConfiguration;
 		CCTextAlignment m_pAlignment;
 		float m_fWidth;
+		bool m_bLineBreakWithoutSpaces;
 	public:
 		CCLabelBMFont()
 			: m_cOpacity(0)           
 			, m_bIsOpacityModifyRGB(false)
-             , m_pConfiguration(NULL)
+            , m_pConfiguration(NULL)
+			, m_bLineBreakWithoutSpaces(false)
 		{}
 		virtual ~CCLabelBMFont();
 		/** Purges the cached data.
@@ -180,14 +182,15 @@ namespace cocos2d{
 		void createFontChars();
 		// super method
 		virtual void setString(const char *label);
-		virtual void setString(const char *label, bool from_update);
-		virtual void updateString(bool from_update);
+		virtual void setString(const char *label, bool fromUpdate);
+		virtual void updateString(bool fromUpdate);
 		virtual const char* getString(void);
         virtual void setCString(const char *label);
 		virtual void setAnchorPoint(const CCPoint& var);
 		virtual void updateLabel();
 		virtual void setAlignment(CCTextAlignment alignment);
 		virtual void setWidth(float width);
+		virtual void setLineBreakWithoutSpace(bool breakWithoutSpace);
 
 #if CC_LABELBMFONT_DEBUG_DRAW
 		virtual void draw();
--- a/cocos2dx/label_nodes/CCLabelBMFont.cpp
+++ b/cocos2dx/label_nodes/CCLabelBMFont.cpp
@@ -207,6 +207,10 @@ namespace cocos2d{
 	{
 		using namespace std;
 		int len = str->size();
+
+		if ( len <= 0 )
+			return;
+
 		int last_index = len - 1;
 
 		// Only start trimming if the last character is whitespace..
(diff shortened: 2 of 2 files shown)


dumganhar:
Hi, irwinb, thanks for your contribution first.
I have found some issues in your code.
My test code as follow:

<pre>
    CCLabelBMFont* pFont = CCLabelBMFont::labelWithString("My LabelMy LabelMy Label", "bitmapFontChinese.fnt", CCTextAlignmentCenter, 200);
    pFont->setPosition( ccp(size.width/2, size.height/2) );
    addChild(pFont);
</pre>

this works as expected.But nothing will be shown if I change the string to "MyLabelMyLabelMyLabel".
And another issue is that you haven't consider other language like Chinese, Korean or Japanese. Since this is supported in the latest code as a new feature.So you should know how this works by reading the code of 'CCLabelBMFont::createFontChars'. Do not deal with string as ANSI string, but UTF-8 string, therefore, the way of counting string length need to be changed to 'cc_utf8_strlen' and the character buffer need to be changed from UTF-8 to Unicode since 'char id' is using unicode in '.fnt' files. 

Regards,
James Chen

dumganhar:
Yes, it's the same as '\0' to terminate string.

> I will fix all of this, but before I do, I was wondering the reasoning behind these two lines
> 
> `unsigned short* pUniStr = new unsigned short[utf8len+1];
>         pUniStr[utf8len] = 0;`
> 
> What is the purpose of setting pUniStr[utf8len] to 0?  Is it the same as '\0' to terminate the string?

dumganhar:
Hi, irwinb,
Thanks for your works, but it still can't work perfectly. For example, when I use "M yLabelMyLabelMyLabel" as a string, nothing will be shown. If I use "MyLabelMyLabelMyLabel", labels are just shown in one line rather than multi-line. My test code is as follows:

<pre>
LabelBMFontChinese::LabelBMFontChinese()
{
    CCSize size = CCDirector::sharedDirector()->getWinSize();
    CCLabelBMFont* pLable = CCLabelBMFont::labelWithString("M yLabelMyLabelMyLabel", "fonts/bitmapFontChinese.fnt", CCTextAlignmentCenter, 200);
    pLable->setPosition(ccp(size.width / 2, size.height /2));
    this->addChild(pLable);
}
</pre>

dumganhar:
If I use the code which use Chinese letters below, my application will crash.

<pre>
LabelBMFontChinese::LabelBMFontChinese()
{
    CCSize size = CCDirector::sharedDirector()->getWinSize();
    CCLabelBMFont* pLable = CCLabelBMFont::labelWithString("中国", "fonts/bitmapFontChinese.fnt", CCTextAlignmentCenter, 200);
    pLable->setPosition(ccp(size.width / 2, size.height /2));
    this->addChild(pLable);
}
</pre>

Could you please fix this issue?

dumganhar:
Hi, irwinb,
I have reviewed your code, you have done a perfect work, but I found that when there are not space and '\n 'characters, the bitmap label will not be set a newline. I don't know whether this is necessary? What's your opinion?

irwinb:
You are too kind..

Hmm, it should probably be that new line characters be inserted if the string contains 0 white space characters.  The reason for that is because if a user gives a label an explicit width, they would expect that the label never expands past the defined bounds.  I was thinking about that but I totally forgot :).  I'll quickly fix that.

dumganhar:
Ok, thank you so much.

irwinb:
So I was using the label today; I found that I preferred that it never break lines at places other than spaces.  I'm not sure what is preferred for other languages, but for English at least, I would prefer that it not break up words.

It would be nice to get some input about other languages.

dumganhar:
Yes, English uses space to distinguish words, but for other languages, they may not use space to achieve that, for example, for Chinese, we use symbol like comma to distinguish sentence, however, there is not space between words. So I think 'break line' is necessary.

dumganhar:
Hello, irwinb,
  Do you still focus on this issue?

irwinb:
Hey dumhanhar, sorry to keep you guys waiting and not following up.  It is crunch time for us right now preparing for GDC..

After GDC passes and it quites down a bit, I will finish this.  I see the next release is in 24 days.  That is more than reachable.

dumganhar:
Okay, thanks. :)

irwinb:
Not sure if anyone has looked at this yet but I implemented both line breaking strategies.  The user can set which strategy they prefer.  Lines are broken on white space by default but can be set to break on any character.  

I chose this because when creating English labels, it is often desirable to have automatic breaking on white space characters.  On the other hand, as dumganhar mentioned, some languages don't require the use of spaces in their text so breaking on any character is required.

dumganhar:
Hi, irwinb, I am testing this pull request. If there are no bugs, I will merge it to our master branch.
Thanks a lot. :)

dumganhar:
Great work, irwinb, I have merged your commit to the master, Thanks very much.

PR #716, opened 2012-02-09
build cocos2dx as static lib and remove some unneeded codes

minggo:


minggo:
Yes, you are right.
Any suggestions to simplify the usage of Android.mk?

minggo:
If no one have suggestions, I will modify HelloWorld, HelloLua and android template.
Because other work based on this.

Issue #720, opened 2012-02-12
Android, Build : Build dependencies using official tools and from source


folecr:
Cocos2D-X on Android depends on the following external BINARY libraries :
- libxml2
- libjpeg
- libpng
  
  ... also due to the current build system it automatically incorporates
- some Crystax libraries
- libiconv
- unneeded portions of libjpeg in the prebuilt binaries

This needs to be fixed.
## Explanation

Currently external depenencies are built using Crystax's toolchain and distributed as prebuilt binaries.
Reasons :
1. There are no distributable makefiles in the Cocos sources for these dependencies
2. For a long time Android did not support C++ fully
3. It is assumed that the build time will be very long (possibly because of Android's poor build sytem)
### The fix is now possible because Android NDK has
- Better makefiles
- Better C++ support
## Suggested fix
1. Modularize the Android NDK makefiles
2. Add Cocos2dx/Android specific makefiles to external dependencies
3. Fork a correct version of external depenencies to Cocos2D's github
4. Build external dependencies from source
5. Use standard toolchain
6. Remove depenency on 3rd-party toolchain
7. Prove that compile time is reasonable (so that it is possible to distribute as source)

folecr:
I'll submit a pull request to fix these issues. I currently have the fix and successfully building HelloWorld.
I am testing HelloLua.

folecr:
Here are the source repositories for the dependencies

#### libxml2
- https://github.com/folecr/platform_external_libxml2
- Note : uses an old version of libxml2 that does not depend on iconv or icu4c

#### libpng
- https://github.com/folecr/platform_external_libpng
- Version used by Android Gingerbread with modified makefiles

#### libjpeg
- https://github.com/folecr/jpeg8d
- Newer version than that used by Android Gingerbread
- Cocos2D-x uses functions that are available in latest version so I used this version instead of hacking the old

minggo:
Great.
I think we should use prebuilt lib as default to reduce build time.
We also provide resource code for those who want to build from source code.

What's your opinion?

folecr:
I will commit changes and let you decide : I did not see significant
slowdown in build time.

On Feb 12, 2012, at 6:01 PM, minggo
reply@reply.github.com
wrote:

folecr:
FYI, changes are here : https://github.com/folecr/cocos2d-x/tree/externals

I am tracking down a crash in libjpeg when running HelloLua. When I am satisfied that it is stable I will submit a pull request.

[...]

PR #729, opened 2012-02-17
Fix for issue 720. Android : Build dependencies from official repos and from sources

folecr:
https://github.com/cocos2d/cocos2d-x/issues/720
# Summary
- These changes will build libpng, libjpeg and libxml2 from source
- Note that the above repositories are currently not in cocos source. They are in my github 
- https://github.com/folecr/jpeg8d
- https://github.com/folecr/platform_external_libpng
- https://github.com/folecr/platform_external_libxml2
- We need to decide how to organize the cocos repository so that it includes the dependencies' sources
# Organization
1. We can use git submodules in the cocos2d-x to point to sources
2. We can include binaries in the cocos2d-x repo and keep sources separate
3. We can keep refer to external repos from within the cocos2d-x repo

Currently my build files are setup as no. 3

Please discuss these changes before pulling on how to organize the sources : 1, 2 or 3.


folecr:
Do the android tests run for everyone? They are currently failing for me (even before my above changes)
I am trying to find out why they are failing ...

minggo:
I think we can use them like box2d and chipmunk.

folecr:
Two issues
- It is more convenient to track bugs down by keeping the projects separate (cd _external repo_ , run unit tests on _external repo_ = find bug!)
- libpng, libjpeg and libxml2 are fairly small. However this version of libxml2 does not support all character sets. So if you have to include another library such as icu4c it is a huge codebase.

Separating the external dependencies is my preferrred solution... We could keep the current organization for this version and if you are preparing for a new version of cocos, that would be the right time to organize externals.

minggo:
How developers use cocos2d-x?
We release cocos2d-x as a zip file, developers can only download the zip file to develop games.
And many developers fork from github, and use the source code.

folecr:
If you fork from github :

```
> git clone cocos2d.git ##### will get all cocos2d sources
> cd cocos2d
> git submodule init && git submodule update ##### will get all dependencies into current directory structure
```

If you use downloaded zip, cocos will package all external sources in one zip file

```
cocos2d.zip
    +--- cocos2d
             .
             .
             .
        +--- libjpeg
        +--- libpng
        +--- libxml2
        +--- Box2D
        +--- chipmunk
```

I think it will be easier for cocos2d developers to maintain this. What do you think?

folecr:
Also, we currently use the above technique (git submodules) to work with both XCode and Android projects... so I think it might work. Please let me know your opinion.

moadib:
If you trying prepare cocos2d-x to use git submodules, may be cocos2d-x itself must be splitted to separate repositories? Because we no need in tests, helloworld, etc in working projects near cocos2d-x :)

folecr:
The difference is that cocos code is dependent on the cocos team.

External modules are dependent on fixes from upstream. Example, if there is a critical bugfix in libxml2 we need to easily bring that into cocos. If we maintain our own copy of the source managing patches and testing on external repos is hard. Maintaining separate repos with a thin patch for cocos makes importing fixes easy. It also makes contributing our changes to upstream easy.

So, splitting cocos, tests, helloworld etc. is _nice to have_. Splitting external dependencies is essential.

moadib:
Agreed.

minggo:
But we want to use released version of external modules, they are stable and not change frequently.

walzer:
Another problem is, e.g., we can use libxml2 from ios sdk / android sdk but don't need to compile it by ourselves. This way will be more stable than compile it from upstream.

walzer:
And config + maintain all 3rd libraries on all platforms: ios, android, blackberry playbook, bada, windows, linux, marmalade ... would be huge of works to do. A bit terrible.

folecr:
@minggo 

> But we want to use released version of external modules, they are stable and not change frequently.

Would you prefer option (2) that I suggested?
"We can include binaries in the cocos2d-x repo and keep sources separate"

folecr:
@walzer 

> Another problem is, e.g., we can use libxml2 from ios sdk / android sdk but don't need to compile it by ourselves. This way will be more stable than compile it from upstream.

Unfortunately on Android there are no official binaries for the libraries that cocos2d-x uses such as libjpeg, libpng, libxml2. The official recommendation is to build from source. Refer to the below statements from the Android team :
- https://groups.google.com/d/msg/android-ndk/EzbqwrS3tz8/AkJxPX66vlgJ
- https://groups.google.com/d/msg/android-ndk/EzbqwrS3tz8/AkJxPX66vlgJ

folecr:
To repeat, 
- The current situation is that Android libs are of unknown provenance - nobody knows where the libraries are coming from.
- There is no good way to optimize (at least for Android) without using a completely new version of the libraries.
- For example, Strawberry Milkshake's very well written guide on optimization http://www.cocos2d-x.org/boards/10/topics/1615 - is hard to do on Android because the sources of the libraries used are not available.

@walzer

> ... would be huge of works to do. A bit terrible.

Yes, it is work. It is not too much work if we keep it separate.
Here is my suggestion :
- Keep separate repos for libjpeg, libxml2, libpng, Box2D, Chipmunk (basically all dependencies)
- Build binaries once from these lib's sources for each platform
- Binaries are included in the official Cocos2d-x distribution and in the sources
- Allow _optional_ building from sources (most developers will never build these binaries from source but those who want to optimize will want to)

Do you have a simpler solution? This suggestion is very similar to what exists today with the only difference being that developers know exactly what sources were used for building libraries.

I can implement the above for Android... please let me know.

justinkb:
I think this is a great idea and should definitely be implemented. Including binaries when the source is readily available (and not that big at all) should never be done IMO.

justinkb:
@folecr I had to add some additional things in various Android.mk files to get this to build, by the way. Off the top of my head, include for libjpeg in the cocos2dx Android.mk file, and include for CocosDenshion to my project's Classes/Android.mk.

minggo:
Now, some platform use prebuilt lib, some use source codes.
I think we should use the same way.

It may make a big change to cocos2d-x, because we should change it for many platforms.
So I think we should do this change in next released version, because we will release a new version
in these days.

folecr:
That might work. Do you have a planned date for the next version?
Even better, if you could clone a repo for the next version I can maintain a patchset for the proposed changes.

walzer:
As the matter of fact, I wish next version would merge Rolando's js binding, but still waiting for legal permission from xxxga

folecr:
I will discuss with Rolando.

justinkb:
I just did the same for the win32 platform, i.e. replaced all prebuilt libraries for that platform by visual studio compiled subprojects (except the PVRFrame libs/dlls of course). It wasn't that hard, yet still nontrivial. Is there any interest in this?

minggo:
Yes, it is not very hard.
We want to put all the resources into another folder named external.

And we want to achieve the goals:
1. You can also use build_native.sh to build android project.
2. It will not increase building time, so we want to use lib as default.
3. Developer can easily use source code. So we should offer Android.mk for android, vs project file and so on.

minggo:
@folecr 
Many platforms may use the same source code of libpng, libjpeg and so on.
So what about put these source codes under external/? external is under the root of cocos2dx.

folecr:
That was my original thought too. Let us commit these changes first and we can move them later?

folecr:
Also, please wait for one more change from me : I need to make the same changes as above for the tests app too. In the meanwhile please review the code!

minggo:
Did you mean merge it first and then move them in another pull request after you making changes for the tests app?

folecr:
Yes.

I will commit few more changes.
Lets review and merge them.

Then let's move the sources to cocos2d-x/externals after this merge. I think that it might require some changes to build systems of other platforms too and this merge has a lot of changes already.

minggo:
Please let me know if it is ready to be merged.
I will do test after you finishing the job and merge it.

folecr:
@minggo I am done with my changes for this pull request. Note that the dependencies are in the same location as before (cocos2d-x/cocos2dx/platform/third_party...) and not in cocos2d-x/externals. Let us move to the new location after these have been integrated - the move should be simple and there are already a lot of changes to test.

Please review and merge. Thanks!

minggo:
@folecr 
Thank you. I will do test and merge it ASAP.

minggo:
@folecr 

We can not build cocos2dx as a dynamic library, or it will cause error because of dynamic_cast.
More info you can refer  http://gcc.gnu.org/faq.html#dso.

And I have some questions:
1.  Can only use ndk-r7? When I use ndk-r6b, I met the error:

Android NDK: Invalid APP_STL value: gnustl_shared
Android NDK: Please use one of the following instead:  system stlport_static stl
port_shared gnustl_static

When I use ndk-r7, the error disappear.
1.  I use ndk-r7, I met the error when building libcocos2dx.so

arm-linux-androideabi-g++.exe: /cygdrive/d/programe/android/ndk/android-ndk-r7/s
ources/cxx-stl/gnu-libstdc++/libs/armeabi/libsupc++.a: No such file or directory

folecr:
@minggo 

> We can not build cocos2dx as a dynamic library, or it will cause error because of dynamic_cast.

I was able to successfully run HelloWorld, HelloLua and the Cocos2D-X tests. Could you add a test for the failure so I can reproduce it?

I am using the latest stable NDK - Revision 7b. I do not see the other errors you mention. Could you give me more details if you get the same errors with NDK r7b?

minggo:
@folecr 
You can try TouchesTest in tests to reproduce it.

I will try NDK r7b and let you know the result.

minggo:
@folecr 
When I use NDK r7b, it builds ok.
As I said, it will crash when running TouchesTest because we use cocos2dx as a dynamic library.

folecr:
I'll test this tomorrow. If I can fix the shared library, I will do so. If not, I'll convert the makefiles to build all code into one library to be loaded from Java/Dalvik.

minggo:
@folecr 
Have you reproduced the crash caused by dynamic_cast?

folecr:
Sorry I have been busy with GDC and other things. I am going to look
at it this weekend.

On Mar 8, 2012, at 6:37 PM, minggo
reply@reply.github.com
wrote:

minggo:
Ok, have a good time in GDC.

minggo:
@folecr 
What's the state of this pull request?
I find it use cocos2dx as shared library as default.

folecr:
I checked this today - dlopen() on Android does not seem to respect the RTLD_GLOBAL flag. This might be the reason dynamic_cast is broken.

I will modify it to use static linkage only.

PS - I needed to make a change to the test code to make it compile successfully. Will submit other pull request.

folecr:
I updated all my changes! :D

Requires changes to CocosDenshion. If you want to keep the tree clean, I suggest you merge https://github.com/cocos2d/cocos2d-x/pull/773 before this. (The changes on this branch have the changes on the cocosdenshion branch too.)

Tested all four flavors of HelloWorld (sources+release, sources+debug, prebuilt+release, prebuilt+debug)
Tested the HelloLua and Tests projects too (one flavor each) and did not see any failures.

Please pull and merge!

minggo:
@folecr 
Have you tested this case, build for armeabi and x86 together.
You can modify Application.mk and add the line:
APP_ABI := armeabi x86

Try to use prebuilt library and source code.
Thank you.

folecr:
Could you describe what the issue is?

folecr:
Updated commits :

Fixed tests build to remove a make warning about circular dependencies
Added

```
APP_ABI := all
```

to Application.mk of HelloWorld, HelloLua and test.android.

minggo:
Awesome work.

I have some suggestions:
1. Don't build for all platforms as default, it is slow. Developers had complained it.
2. How about removing tests/Android.mk and merge it into tests/test.android/jni/Android.mk, just as HelloWorld did.

folecr:
> 1. Don't build for all platforms as default, it is slow. Developers had complained it.

I agree. I'll remove this change.

> 1. How about removing tests/Android.mk and merge it into tests/test.android/jni/Android.mk, just as HelloWorld did.

The tests code shared between platforms is a large body of code. I think it is more manageable as it's own library. It is also easy to build when you update the tests because a developer of tests will not be updating the android specific code often when writing tests. This will make the build simpler/faster.

minggo:
Ok, I will merge the pull request after you removing

```
APP_ABI := all
```

folecr:
Updated. Please merge!

PR #730, opened 2012-02-22
Added CCEGLView::EventHandler class for qnx-platform

moadib:
We've released first version of our game to Blackberry Appworld and now going to integrate inapp purchases and scoreloop. This require some work with event handling which occurs inside CCEGLView.

I've added CCEGLView::EventHandler class and for qnx-version. Now i can inherit Application from and have possibility catch nessesary events and dispatch in appropriate managers. If event handled HandleBSPEvent must return true to indicate that this event must be skipped in main event handling loop.

### DIFF
--- a/cocos2dx/platform/qnx/CCEGLView_qnx.cpp
+++ b/cocos2dx/platform/qnx/CCEGLView_qnx.cpp
@@ -96,6 +96,8 @@ CCEGLView::CCEGLView()
     m_screenEvent = 0;
     m_screenWindow = 0;
 
+    strcpy(m_window_group_id, "");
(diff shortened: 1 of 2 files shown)


moadib:
Also added screen_create_window_group call to window initialization and getter for group id. This is necessary for work with payment service and may be something else.

Example of use:

```
paymentservice_purchase_request(digital_good_id, digital_good_sku, digital_good_name,
            metadata, purchase_app_name, purchase_app_icon, CCEGLView::sharedOpenGLView().getWindowGroupId(),
            &request_id)
```

dumganhar:
Thanks, :)

PR #731, opened 2012-02-22
Stop unnecessary Spam in output log regarding failure to find target action.

Wartortle:
Added call guard in CCNode::stopAllActions() to prevent unnecessary call
to CCActionManager::removeAllActionsFromTarget() so that in situations
where there are no actions, we do not get any Spam in the output log about
failing to find action/target.
Ref: http://www.cocos2d-x.org/boards/6/topics/2406?r=2465#message-2465

### DIFF
--- a/cocos2dx/platform/ios/CCImage_ios.mm
+++ b/cocos2dx/platform/ios/CCImage_ios.mm
@@ -578,21 +578,25 @@ bool CCImage::_initWithRawData(void *pData, int nDatalen, int nWidth, int nHeigh
 bool CCImage::_initWithJpgData(void *pData, int nDatalen)
 {
     assert(0);
+    return false;
 }
 
 bool CCImage::_initWithPngData(void *pData, int nDatalen)
 {
     assert(0);
+    return false;
 }
 
 bool CCImage::_saveImageToPNG(const char *pszFilePath, bool bIsToRGB)
 {
     assert(0);
+    return false;
 }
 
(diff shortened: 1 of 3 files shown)


minggo:
Wartortle
Thank you for your work.
But there are some issues:
1. You do much thing in one pull request, could you please divide them into more?
2. The pull request can not been automatically merged.

Wartortle:
Apologies but this is partly due to my inexperience with Git and partly due to GitHub.  I had initially committed and raised a pull request for the Accelerometer Simulation.
Then, I added another one or two commits for the other fixes I did and tried to raise another pull request.  It didn't allow me to do so and, from what I gather, merged the two requests into one.
Sorry if it's caused problems but I tried to follow the instructions and it really wasn't obvious to me what to do.
If you can provide any clarification on how I should have handled this I would be most grateful.

minggo:
You should do one thing in one branch, don't do all the thing in the same branch.
Then you can switch to different branch to pull request.

Wartortle:
Okay, understood.

minggo:
So, could you divide the pull request into more pull requests?
And don't forget to synchronize your codes with cocos2d-x.

Wartortle:
I'll have a go, sure.

minggo:
I will close the pull request, what's your opinion?

Wartortle:
Yes, it's in the main code already - I checked.

PR #733, opened 2012-02-25
add an android template for creating projects with box2d support

walzer:
as the title


minggo:
I don't think we need this bash file, because now it is easy to include box2d, you should just add "box2d_shared" in Android.mk. For example, you add "box2d_shared" in HelloWorld/Classes/Android.mk after "cocosdenshion_shared", then, it will include box2d module.

walzer:
OK. Waiting for you function.

PR #741, opened 2012-03-02
remove libiconv.a

minggo:
libiconv is LGPL based, so can not use it as a static lib.


walzer:
But could we offer iconv as a dynamic library? I'm not sure if someone need it.

minggo:
It is just the first step.

minggo:
We should offer a .so or source code in next version after merge folecer's pull request.

PR #760, opened 2012-03-08
Added support files for Accelerometer Simulation for iPhone Simulator an...

Wartortle:
...d

Unimotion
Ref: http://code.google.com/p/accelerometer-simulator/wiki/Home
Ref: https://github.com/gmarik/unimotion

### DIFF
--- a/cocos2dx/platform/ios/Simulation/AccelerometerSimulation.h
+++ b/cocos2dx/platform/ios/Simulation/AccelerometerSimulation.h
@@ -1,0 +1,59 @@
+/*
+ *  AccelerometerSimulation.h
+ *  AccelerometerGraph
+ *
+ *  Created by Otto Chrons on 9/26/08.
+ *  Copyright 2008 Seastringo Oy. All rights reserved.
+ *
+ */
+#import <TargetConditionals.h>
+
+// when compiling to ARM (iPhone device), hide everything and use system defaults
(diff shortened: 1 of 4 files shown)


minggo:
Thank you, may be you can also write a document to describe how to use it in http://www.cocos2d-x.org.

Wartortle:
Hi Minggo,
There isn't very much to do.  I've logged in to the Cocos2D-x site but can't see how I can change any of the Wiki pages.  Perhaps I don't have the correct access rights.

To use this under the iOS simulator, you don't need to do anything - the code is automatically compiled in just for the simulator and, if active, it overrides the default accelerometer control to interface with Unimotion.

Basically, to use Unimotion, follow the instructions for building it as indicated here:
https://github.com/gmarik/unimotion

Once built, run Unimotion as given in the above docs.
Build your iOS app and run on the iOS Simulator.
Pick up your macbook and wiggle it around and the accelerometer in the macbook will drive the accelerometer interface in the Cocos2D-x game.

Hope this helps.
Roc

minggo:
I think you can write a doc in forum, then I will migrate it to wiki.

Wartortle:
http://www.cocos2d-x.org/boards/7/topics/9278

Is that okay?

minggo:
Thank you.
I have added a link in wiki.

PR #761, opened 2012-03-08
Fixed and improved building on Linux

peoro:
Added linux libraries to .gitignore: *.a (static libraries) and *.so (shared ones)

Fixed a typo in two file names in cocos2dx/platform/linux/: CCAplication_linux.h and CCAplication_linux.cpp, which were missing a "p".
Also fixed two broken paths in cocos2dx/proj.linux/Makefile, which were using "../platform/Linux/" instead of "../platform/linux/" (with lower case "l").

Fixed parameters to a SetTouchInfo call in cocos2dx/platform/linux/CCEGLView_linux.cpp: three parameters were used, the first one, 0, appears to be no longer accepted.

### DIFF
--- a/cocos2dx/platform/linux/CCEGLView_linux.cpp
+++ b/cocos2dx/platform/linux/CCEGLView_linux.cpp
@@ -156,7 +156,7 @@ void mousePosEventHandle(int iPosX,int iPosY) {
 	if (iButtonState == GLFW_PRESS) {
 		if (iPosX!=(int)s_pMainWindow->m_mousePoint.x||iPosY!=(int)s_pMainWindow->m_mousePoint.y) {
 			//it movies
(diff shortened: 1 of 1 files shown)


folecr:
Regarding

> d64918f ".a (archives, static libraries) and .so (shared libraries) files added to .gitignore"

Some of the other platforms like Android want to ship with prebuilt libraries for external dependencies to save build time. If you explicitly specify the full path to the files instead of globally adding *.so and *.a it would serve the linux build system and  be compatible with other platforms too.

minggo:
Yeap, you can not just ignore .so and .a, should add the path.

peoro:
Right, I didn't think of that: I never add binary files to my git repos.
Maybe in the next days I'll submit a new pull request to ignore only libraries built by build-linux.sh.

Are the other commits OK?
What should I do now? Open a new pull request with the other two commits only, or can you just pull them with `git-cherry-pick`?

minggo:
You can modify .gitignore and pull the request again.

peoro:
Updated.

Issue #732, opened 2012-02-22
CCImage_Linux has bugs


justinkb:
initWithString gives, in the best case, badly rendered glyphs and in the worst case segfault due to memory deallocation problem...

Easy to reproduce. Just make a CCLabelTTF on linux.

minggo:
Label in linux looks not very good, but it is used for debugging, and we have not enough resource to do it better.
May be you can help the community to do it.

PR #764, opened 2012-03-09
Linux implementation of CCImage::initWithString now uses fontconfig

peoro:
Added fontconfig support for Linux.

`CCImage::initWithString` now uses fontconfig to find a font matching its `pFontName` parameter.

By disabling hte `USE_FONTCONFIG` macro it's possible to use the old function, where `pFontName` was used a font file path.

This should fix issue #732

### DIFF
--- a/cocos2dx/platform/linux/CCImage_linux.cpp
+++ b/cocos2dx/platform/linux/CCImage_linux.cpp
@@ -11,6 +11,9 @@
 #include "ft2build.h"
 #include "CCStdC.h"
 #include FT_FREETYPE_H
+#ifdef USE_FONTCONFIG
+	#include <fontconfig/fontconfig.h>
+#endif
 
 #define szFont_kenning 2
 
@@ -32,6 +35,10 @@ public:
 		iInterval = szFont_kenning;
 		m_pData = NULL;
 		reset();
+
+#ifdef USE_FONTCONFIG
+		libError |= ! FcInit();
+#endif
 	}
 
 	~BitmapDC() {
@@ -40,6 +47,9 @@ public:
 //		if (m_pData) {
 //			delete m_pData;
 //		}
+#ifdef USE_FONTCONFIG
+		FcFini();
+#endif
 
 	}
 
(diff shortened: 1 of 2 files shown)


minggo:
Is fontconfig a common library?
If not, the engine should offer one?

peoro:
Fontconfig is a very common library: kind of any graphical application and library that needs to display some text uses fontconfig, including Qt and GTK. Cocos2dx users may require to install development files for fontconfig, but they are available for any distribution.

About libraries needed by cocos2dx I've got some doubts though:
- FreeType2 is shipped with cocos2dx sources, although it's really common as well (fontconfig depends on FreeType2): you'll find it on any Linux system with any kind of graphical interface installed.
- GLFW is not shipped with cocos2dx sources, and it's far less common: GLFW is available on any Linux distribution as well, but never installed by default.
  I think an issue could be opened about this, but I don't really mind.

Last thing about this pull request: the semantics of my `CCImage::initWithString` is that of using the font matching `pFontName`. If `pFontName` is a path to a TTF file it won't load it, will rather load the font matching such path (which will probably be the default system font).
The function should be changed a little if its semantics should be something like "load a font file or a font matching the given name".

minggo:
Thank you.
I am not familiar with linux, the port of it is finished by a developer.

I compile the code, and find my system don't install fontconfig. There is an error to include <fontconfig/fontconfig.h>, my system is ubuntu11.04. So fontconfig is not so common that it is installed by default.

So I think you can not use fontconfig as default, or many developers may meet the same problem like me.
How about remove -DUSE_FONTCONFIG as default.

minggo:
Thank you.
I am not familiar with linux, the port of it is finished by a developer.

I compile the code, and find my system don't install fontconfig. There is an error to include <fontconfig/fontconfig.h>, my system is ubuntu11.04. So fontconfig is not so common that it is installed by default.

So I think you can not use fontconfig as default, or many developers may meet the same problem like me.
How about remove -DUSE_FONTCONFIG as default.

peoro:
I believe that fontconfig is installed, try typing `fc-match Arial` in a terminal: that command should use fontconfig to look for a font matching "Arial". Also the fontconfig library should be present: its path would be something like `/usr/lib/libfontconfig.so.1`.

The problem is that development files (ie: header files) are not installed by default, you'll need to install `libfontconfig-dev` for that. This is the same for any software, ubuntu doesn't install any -dev package by default.

Anyway ok, should I remove USE_FONTCONFIG macro and pull the request again?

minggo:
Yes, please remove USE_FONTCONFIG macro.
Thank you.

PR #765, opened 2012-03-09
Linux / Ubuntu 64 bit fixes #2

eliam:
Hi,

I've now included all the original changes.  Not familiar with git :-/


minggo:
I think the pull request is similar to https://github.com/cocos2d/cocos2d-x/pull/761.

PR #773, opened 2012-03-15
Cocosdenshion JNI use

folecr:
Use JNI Helper in CocosDenshion.
Until now CocosDenshion relies on JNI_OnLoad to initialize.

This change allows CocosDenshion to be built either as a static or shared library.

It is required to use C++ correctly in the Android build.

Note : this adds a depedency to CocosDenshion so it can no longer be built independently. If necessary this can be refactored by adding a parameter to the CocosDenshion constructor that would take an initializer.

### DIFF
--- a/CocosDenshion/android/Android.mk
+++ b/CocosDenshion/android/Android.mk
@@ -10,8 +10,11 @@ jni/SimpleAudioEngineJni.cpp
 
 LOCAL_EXPORT_C_INCLUDES := $(LOCAL_PATH)/../include
 
-LOCAL_C_INCLUDES := $(LOCAL_PATH)/../include
-                   
+LOCAL_C_INCLUDES := $(LOCAL_PATH)/../include \
+                    $(LOCAL_PATH)/../../cocos2dx/include \
+                    $(LOCAL_PATH)/../../cocos2dx/platform \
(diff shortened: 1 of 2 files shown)


rolandoam:
I think it's ok. Probably no-one will use this version of CocosDenshion alone, so the dependency is not an issue.

minggo:
Why to build cocosdenshion as a static library?
Is it any benefit.
This makes cocosdenshion dependent of cocos2dx.

folecr:
Denshion is C++
Cocos is C++ but needs to have all dependencies linked statically due to the crash with dynamic_cast.

If they are separate shared libraries then we have to deal with managing two separate instances of C++ runtimes. With the issues we have had on Android with C++ libs this is not good (especially dynamic_cast.)

The other solution might be to write Denshion in C. (Since it is only a few classes.) however, this will require modifying a lot of code in both Cocos and Denshion.

I think that for now this is the solution to use. Maybe we can change the interface for initialization of Denshion in the coming weeks? Then we can initialize it differently for the shared and static cases and make the dependency optional. However, please remember that even in this case... It is still not recommended to use two separate c++ shared libs unless we build the c++ runtimes as shared.

folecr:
PS Going by all the reports in the android-ndk groups it might be sensible to link everything statically on Android just like on iOS

PR #790, opened 2012-03-21
Added getCurrentLanguageCode fucntion to CCApplication.

Zakay:
Returns standard 2 char language codes

### DIFF
--- a/cocos2dx/platform/android/CCApplication_android.h
+++ b/cocos2dx/platform/android/CCApplication_android.h
@@ -90,7 +90,7 @@ public:
     @brief Get current language code
     @return Current language code
     */
-	static std::string getCurrentLanguageCode();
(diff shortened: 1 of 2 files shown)


Zakay:
Sorry about my editor, doing a little bit to much whitespace fixes

minggo:
Different platforms may return different code, then the game can not use the same code in different platforms.
That's the reason we implement getCurrentLangua().

Zakay:
Thats why there are standards.
http://sv.wikipedia.org/wiki/ISO_639

Which both android/ios follow.
For the other platforms, where a direct use of the locale can't be used, you could build support for a bunch of languages just like you did with getCurrentLanguage. 

A reverse of getCurrentLanguage is needed anyway. (Its comes free on ios/android because they use the standard)

PR #823, opened 2012-04-02
Fix some issues in OpenAL implementation of CocosDenshion for qnx

moadib:
Only paused sources need to be returned to PLAYING state, otherwise resumeBackgroundMusic and resumeAllEffects causes start of playing source even it is not startet previously.
Also fix error in "resume" and "pause" all sounds - there is no iterating over array of corresponding sources.


dumganhar:
Hi, moadib, thanks for your efforts on qnx platform. :)

PR #825, opened 2012-04-05
Enabling point particle systems.

dducharme:
Enabling the point particle systems as they are now working in 2.0 and now using bps events for the screen.

### DIFF
--- a/cocos2dx/include/CCParticleSystemPoint.h
+++ b/cocos2dx/include/CCParticleSystemPoint.h
@@ -28,7 +28,7 @@ THE SOFTWARE.
 
 /*#include "Availability.h"*/
 #include "CCParticleSystem.h"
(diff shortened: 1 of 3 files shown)


dumganhar:
Hi, dducharme, should we discard 1.0 sdk support?
And how about the progress of custom ttf support?
There are two issues on our site, http://www.cocos2d-x.org/issues/1024 and http://www.cocos2d-x.org/issues/1023.
Will you try to fix them?

dducharme:
Hi James,

Yes I think its probably good to get rid of 1.0 support at this time and just go with 2.0.  

As for the font issue, I think we'll have to do that through skia and I'm not quite sure what the support is like.  Will try and get a fix if that is possible. 
- Dale

----- Original Message -----

dumganhar:
Ok, thanks.

PR #827, opened 2012-04-06
use <> to include pthread.h and semaphore.h

minggo:


minggo:
wait gles20 to merge the commit

PR #828, opened 2012-04-06
Modify CCRenderTexture::end() default behaviour

folecr:
- Android port may cache textures.
- This is set to true by default.
- Should be false so that behaviour is the same between Android and other platforms that don't have this feature.

NOTE : This feature has performance problems on some devices so the default behaviour is even more critical. Needs to be investigated in future.

### DIFF
--- a/cocos2dx/include/CCRenderTexture.h
+++ b/cocos2dx/include/CCRenderTexture.h
@@ -80,7 +80,7 @@ public:
 
 	/** ends grabbing*/
 	// para bIsTOCacheTexture       the parameter is only used for android to cache the texture
-	void end(bool bIsTOCacheTexture = true);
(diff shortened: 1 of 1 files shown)


folecr:
# Test
- RenderTextureSave calls end()
- RenderTextureTest calls end(false)

## On Android devices...

### Before these changes
- RenderTextureSave is much slower than RenderTextureTest

### After these changes
- RenderTextureSave and RenderTextureTest are the same speed

minggo:
Thank you.
I will test corresponding test.

folecr:
Looks OK?

minggo:
Sorry, I am busy for developing cocos2d-x gles20 these days.
So I haven't tested.

I will test is ASAP.
Thank you again.

PR #848, opened 2012-04-17
Fix win32

dualface:
增加两个热键：
CTRL+X 退出
CTRL+R 重新启动

增加控制台窗口输出，方便脱离 Visual Studio 环境进行调试。

例如美工调试素材，不需要为美工安装 VS，只需要把执行文件、DLL 和资源文件拷贝到美工机器上。美工直接修改资源文件后重启游戏就能看到效果。

对于策划和脚本开发人员，这个修改也可以显著提高工作效率。

### DIFF
--- a/HelloLua/win32/main.cpp
+++ b/HelloLua/win32/main.cpp
@@ -2,16 +2,32 @@
 
 #include "AppDelegate.h"
 
+// uncomment below line, open debug console
+// #define USE_WIN32_CONSOLE
+
 int APIENTRY _tWinMain(HINSTANCE hInstance,
-					   HINSTANCE hPrevInstance,
-					   LPTSTR    lpCmdLine,
-					   int       nCmdShow)
+                       HINSTANCE hPrevInstance,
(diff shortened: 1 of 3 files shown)


minggo:
Please use English.

dualface:
sorry :)
1. Add two hotkeys for Win32 application:

CTRL+X = Quit
CTRL+R = Restart
1. Add console window for Win32 application, all debug message output to this console window.

---

These two changes can significantly improve the work efficiency.

For the artist:
Without Visual Studio environment, you can always test the effect of the picture displayed in the application.

For game designers:
Can immediately test the modified script file.

dumganhar:
I think we should disable the macro USE_WIN32_CONSOLE by default.

minggo:
And it is better to write a document describes how to use it.
Then other developers will notice the feature and use it.

dualface:
change win32 main.cpp, debug console set default disabled.

how to use: http://cn.cocos2d-x.org/bbs/forum.php?mod=viewthread&tid=434 (CHINESE ONLY)

PR #854, opened 2012-04-18
Android: Minor fix for template/android/build_native.sh

NatWeiss:
A potential problem happens when the filename of a resource being copied by the template/android/build_native.sh script has a space character in it.

Example: having "My Resource.png" in your resource folder causes the bash warning "binary operator expected" and fails to copy the file to the assets folder in the copy resources phase.


minggo:
Thank you, man.

PR #860, opened 2012-04-22
CCLabelBMFont updateLabel() optimizations and fixes

irwinb:
After my initial additions to CCLabelBMFont, some bugs were found while using the label at Cardboard Box Games.  Here are those fixes.  It includes warning removals, improvements in multi-line text alignment and performance.


dumganhar:
Hi, at line 1278 and 1279 in CCLabelBMFont.cpp, why should use 

int j = 0;  //this j will not take effect.
for (unsigned j = 0; j < line_length; j++)

instead of 

for (int j = 0; j < line_length; j++)

??

PR #863, opened 2012-04-23
fixed bugs and issues

unitpoint:


minggo:
Could you add some comment here for shared finalizer?

unitpoint:
I had a lot of troubles finding memory leaks because of many static instances used in cocos2d-x. The shared finalizer allows programer check application memory leaks like that (example for marmalade port):

```
int main()
{
    IwMemBucketInit();
    // IwMemBucketDebugSetBreakpoint(445);
    IW_CALLSTACK("App::runGame");

    int32 checkpoint = IwMemBucketCheckpoint();

    App * app = new App;
    int ret = app->Run();
    delete app;

    IwMemBucketDebugCheck(0, checkpoint, "check-app-leak.txt");

    IwMemBucketTerminate();
    return ret;
}
```

application destructor frees all instances now so file "check-app-leak.txt' contains empty data if OK and some debug information on leaks found.

minggo:
Static objects are deleted at the end of app.
Are tools that detecting memory leak treat these static objects as memory leak?

unitpoint:
Yes.

minggo:
Ok, thank you.

I think there is a but in CCSharedFinnalizer::finalizeAll(), invoke

```
delete sharedFinalizer;
```

twice

unitpoint:
It makes sense because of someone can create objects in the finalizer process

minggo:
I am sorry, could you explain it?

unitpoint:
Programer can touch shared objects to check some thing when the finalizer process in progress. It produces create the shared objects that already freed, for example CCPoolManager and so on. My investigation gave me information that all shared objects are deleted after delete operator called twice. The delete operator checks if pointer is NULL. So deleting the NULL pointer is absolute safely and makes sense to free all shared objects properly.

minggo:
```
All shared objects are deleted after delete operator called twice.
```

I think it is not a good way. Why should do it like this?
Why can not delete all shared objects if delete operator called one time?

unitpoint:
You can make the delete operator called once. In most cases that will be enough.

unitpoint:
I'm not sure, is this request processed?

minggo:
I am sorry. I am very busy these days.
So have not time to test and merge it.

I will handle it ASAP.

minggo:
Sorry for late merging.

I have two questions:
1. struct **SharedFinalizer** in CCObject.cpp is not used.
2. When i merge your your codes and do test, i meet the compiling error in CCSharedFinalizer::atexit().
The error message is 
"No member named 'find' in namespace 'std'"

My environments:
Xcode 4.3.2
Mac OS X 10.7.3

I don't know why 'find' is not in 'std'.

unitpoint:
struct **SharedFinalizer** is used to call CCSharedFinalizer::finalizeAll at application exit inside of destructor of **SharedFinalizer**.

I have removed std::find, change please CCSharedFinalizer::atexit function to:

```
void CCSharedFinalizer::atexit(void(*func)())
{
    CCSharedFinalizer * finalizer = getInstance();
    for(int i = 0; i < finalizer->funcs.size(); i++){
        if(finalizer->funcs[i] == func){
            return;
        }
    }
    finalizer->funcs.push_back(func);
}
```

minggo:
Can not be automatically merged.
Please fix the conflict, then i will test and merge it.

Thank you.

PR #883, opened 2012-05-03
fix JNI leaks

ehooi:
release local references earlier.
more stable when running big loop with call JNI methods.

FYI. from http://java.sun.com/docs/books/jni/html/objtypes.html

Do not forget to call ReleaseStringChars when you no longer need access to the string elements returned from GetStringChars. The ReleaseString-Chars call is necessary whether GetStringChars has set *isCopy to JNI_TRUE or JNI_FALSE. ReleaseStringChars either frees the copy or unpins the instance, depending upon whether GetStringChars has returned a copy or not.


minggo:
Thank you.

PR #885, opened 2012-05-03
add CocosBuilder support for gles20 branch

walzer:
It can work on iOS now. Android and windows is coming soon.


dumganhar:
There are Chinese comments in CCBReader.h.

walzer:
Eventually I finished the CCBReader v1 support.
I think the implementation is elegant and easy to read.

PR #889, opened 2012-05-04
Killa support

ex:
I did some changes to the Lua bindings to support the language I'm creating (Killa).
Killa is based on Lua 5.2 and adds some helpful features like mandatory variable declarations and JavaScript-like syntaxis. However it's still in a very early phase of development (but it's totally functional).
Some minimal updates to tolua++ were done and my changes only work if KILLA is predefined.

I'm not sure if the team is interested on this, just saying :D


minggo:
So crazy.
I think few people know Killa, so I think it is not suitable to merge it.

ex:
hahaha, I knew it. I'll keep it synchronized anyway. Some bits could be useful when you decide to use Lua 5.2 instead of 5.1

ghost:
crazy work！

PR #891, opened 2012-05-04
Made CCNode anchorPoint Lua bindings work the same way as position

LaunchingPadTim:
CCNode anchorPoint still required creation of an intermediate CCPoint
in Lua. Made it take two floats directly instead (like position does).


minggo:
LaunchingPadTim, what do you think about what dualface said?

PR #896, opened 2012-05-07
Virtual Keyboard on BB10/PlayBook

Ian-MacDonald:
There was already support for bringing up and pushing down the virtual keyboard, but this will now handle the OS events signalling that the keyboard has been pushed to or removed from the screen by the user (by swiping to center from bottom-left, for example).

### DIFF
--- a/cocos2dx/platform/qnx/CCEGLView_qnx.cpp
+++ b/cocos2dx/platform/qnx/CCEGLView_qnx.cpp
@@ -60,14 +60,6 @@ PFNGLFRAMEBUFFERTEXTURE2DOESPROC   CCEGLView::glFramebufferTexture2DOES = 0;
 PFNGLDELETEFRAMEBUFFERSOESPROC     CCEGLView::glDeleteFramebuffersOES = 0;
 PFNGLCHECKFRAMEBUFFERSTATUSOESPROC CCEGLView::glCheckFramebufferStatusOES = 0;
 
-enum Orientation
-{
-	PORTRAIT,
-	LANDSCAPE,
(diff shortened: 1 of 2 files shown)


Ian-MacDonald:
Accidentally requested a pull of the master.

cocos2dx/platform/qnx/CCEGLView_qnx.cpp
```diff
@@ -779,6 +772,86 @@ CCSize CCEGLView::getSize()
...
+  case -270:
+  case -90:
+  case 90:
+  case 270:
+    temp = size[0];
+    size[0] = size[1];
+    size[1] = temp;
+    skip = 0;
+    break;
+  }
```
Ian-MacDonald, line 814: dispatchKeyboardWillHide was never actually called before - only the Show methods were called.

cocos2dx/platform/qnx/CCEGLView_qnx.cpp
```diff
@@ -60,14 +60,6 @@ of this software and associated documentation files (the "Software"), to deal
...
 PFNGLCHECKFRAMEBUFFERSTATUSOESPROC CCEGLView::glCheckFramebufferStatusOES = 0;
 
-enum Orientation
-{
-	PORTRAIT,
-	LANDSCAPE,
-	AUTO
-};
-
-static Orientation orientation = LANDSCAPE;
```
Ian-MacDonald, line 63: This will handle the native gesture that pulls the keyboard up (swipe to center from bottom left).

PR #897, opened 2012-05-08
Native Keyboard support BB10/PlayBook

Ian-MacDonald:
There was already support for bringing up and pushing down the virtual keyboard, but this will now handle the OS events signalling that the keyboard has been pushed to or removed from the screen by the user (by swiping to center from bottom-left, for example).


dumganhar:
Thanks. :)

PR #902, opened 2012-05-16
CCPoint extension

Klaim:
Some additional basic functions to ease manipulation of CCPoint.

### DIFF
--- a/cocos2dx/include/CCGeometry.h
+++ b/cocos2dx/include/CCGeometry.h
@@ -76,6 +76,7 @@ inline CCPoint operator*( const CCPoint a, const float factor )
 
 inline CCPoint operator/( const CCPoint a, const float factor )
 {
+    assert( factor != 0.f );
     return CCPoint( a.x / factor, a.y / factor );
 }
     
@@ -99,6 +100,7 @@ inline CCPoint& operator*=( CCPoint& a, const float factor )
 
 inline CCPoint& operator/=( CCPoint& a, const float factor )
 {
+    assert( factor != 0.f );
     a = a / factor;
     return a;
 }


cocos2dx/include/CCGeometry.h
```diff
@@ -46,6 +48,93 @@ class CC_DLL CCPoint
...
+}
+    
+inline CCPoint operator*( const CCPoint a, const float factor )
+{
+    return CCPoint( a.x * factor, a.y * factor );
+}
+
+inline CCPoint operator/( const CCPoint a, const float factor )
+{
+    return CCPoint( a.x / factor, a.y / factor );
```
minggo, line 80: How about add 

```
assert(factor != 0)
```

before return?

cocos2dx/include/CCGeometry.h
```diff
@@ -46,6 +48,93 @@ class CC_DLL CCPoint
...
+
+inline CCPoint& operator*=( CCPoint& a, const float factor )
+{
+    a = a * factor;
+    return a;
+}
+
+inline CCPoint& operator/=( CCPoint& a, const float factor )
+{
+    a = a / factor;
```
minggo, line 104: How about add 

```
assert(factor != 0)
```

before doing dividing?

Klaim, line 104: Right, I'll add it.

cocos2dx/include/CCGeometry.h
Klaim, line 80: Right.

minggo:
Thank you.
And i will add you into AUTHORS.

PR #904, opened 2012-05-21
Remove(d) whitespaces in VertexShader/FragmentShader defines.

nicolasgramlich:
Removed lots of (most likely) unnecessary whitespaces. This applies to all other shader files as well. My limited C++ knowledge doesn't allow me to judge if this affects compilation though. =P


minggo:
It is a shader file, and it is compiled by opengl es compiler.
I think it doesn't affect compilation, but thank you all the same.

nicolasgramlich:
Yes, it only affect compilation speed (less whitespaces to parse) and saves a few bytes. Although that difference will be hard to measure and readability might be preferred here.

Feel free to close this pull request.

PR #905, opened 2012-05-22
Fixed CCDirector using CCLabelBMFont instead of CCLabelTTF.

nicolasgramlich:
The usage of CCLabelTTF for a frequently updating labels is discouraged, since it creates a significant amount of very short lived objects/textures. Especially on Android this caused the Garbage Collector to run every 2 seconds, disposing about 200kB/second worth of Objects (Bitmap, Canvas, Paint, String, Pattern, ...) + costly JNI calls.

Things to change along with this pull request are: 
- Add a fps_images-hd.png
- Verify that the iOS and MSVC templates still work correctly.

### DIFF
--- a/template/xcode4/base_ios.xctemplate/TemplateInfo.plist
+++ b/template/xcode4/base_ios.xctemplate/TemplateInfo.plist
@@ -90,6 +90,26 @@
 			<key>Path</key>
 			<string>Resources/iTunesArtwork</string>
 		</dict>
+		<key>Resources/fonts/fps_images.fnt</key>
+		<dict>
+			<key>Group</key>
+			<array>
+				<string>Resources</string>
+				<string>fonts</string>
+			</array>
+			<key>Path</key>
+			<string>Resources/fonts/fps_images.fnt</string>
+		</dict>
+		<key>Resources/fonts/fps_images.png</key>
+		<dict>
+			<key>Group</key>
+			<array>
+				<string>Resources</string>
+				<string>fonts</string>
+			</array>
+			<key>Path</key>
+			<string>Resources/fonts/fps_images.png</string>
+		</dict>
 	</dict>
 	<key>Identifier</key>
 	<string>org.cocos2d-x.iosbase</string>
@@ -109,6 +129,8 @@
 		<string>Resources/Icon@2x.png</string>
 		<string>Resources/Info.plist</string>
 		<string>Resources/iTunesArtwork</string>
+		<string>Resources/fonts/fps_images.fnt</string>
+		<string>Resources/fonts/fps_images.png</string>
 	</array>
 	<key>Platforms</key>
 	<array>


cocos2dx/sprite_nodes/CCSpriteBatchNode.cpp
```diff
@@ -129,8 +129,6 @@ void CCSpriteBatchNode::visit(void)
 {
     CC_PROFILER_START_CATEGORY(kCCProfilerCategoryBatchSprite, "CCSpriteBatchNode - visit");
 
-    CCAssert(m_pParent != NULL, "CCSpriteBatchNode should NOT be root node");
```
minggo, line 132: Why delete this line?

minggo:
Why i use CCLabelTTF instead of CCLabelBMFont is hat, the fps is used in debug version.
It will not effect release version.
When using CCLabelBMFont, the release version should also include the fnt and picture files for fps.

cocos2dx/sprite_nodes/CCSpriteBatchNode.cpp
nicolasgramlich, line 132: This change is necessary since the CCLabelBMFont objects in CCDirector don't have parent's to them and so this assert fails.
@ricardoquesada said that this line was once added to prevent users from using a CCSpriteBatchNode as the root node and then complaining in the forums. We agree that removing that line would be the proper fix.

nicolasgramlich:
@minggo true, but on the other hand properly measuring the performance in debug mode with the previous code is difficult. (At least on Android, due to the frequent(!) Garbage Collection.)

minggo:
Yeap, I agree with you.
Could you please modify to other two projects: HelloLua and tests.
Thank you.

nicolasgramlich:
I will add the two resources to those projects tomorrow too. (Heading out of office as I'm writing this =P )

Note: after merging these, it should be tested that all the template (especially iOS) still work (...since I guess that the two new files somehow have to be registered in the template file?).

minggo:
Thank you.
We will test it, and we should do some modification for android and win32 template.

nicolasgramlich:
Just verified, that the status quo works of the Android template works.

nicolasgramlich:
Test also work, at least on Android.

minggo:
Thank you.
I will test ASAP.

minggo:
Oh, you didn't modify corresponding xcode project.
No matter, i will do it.

minggo:
And there is a bug in xcode template.
When create a project by template, Resources/fonts is added into project by "Create groups for any added folders", 
but we use the font in CCDirector.cpp like this

```
m_pFPSLabel = CCLabelBMFont::labelWithString("00.0", "fonts/fps_images.fnt");
m_pSPFLabel = CCLabelBMFont::labelWithString("0.000", "fonts/fps_images.fnt");
m_pDrawsLabel = CCLabelBMFont::labelWithString("000", "fonts/fps_images.fnt");
```

So can not find resources.

minggo:
I think there are two ways to resolve it.
1. Move fps_images.fnt/png into Resources, not under Resources/fonts, and modify codes in CCDirector.cpp.
2. Modify TemplateInfo.plist to create folder for fonts, but i don't know how to do it.

nicolasgramlich:
1. doesn't seem to be supported for Xcode templates, that's what I wanted to state with my previous commit message =)

minggo:
So i modify it with 1?
Then i will pull request and merge it.

PR #906, opened 2012-05-23
cocos2d-x for Android : Use a library to organize code common across all projects

folecr:
This pull request
- Creates an Android library of Java sources and Android layout resources
- HelloWorld, HelloLua and tests.android have been updated to use this library
- Code used by cocos2d-x projects can now be tested easily in tests.android (earlier it was very painful to copy code to tests.android, re-generate a project that then used this new code. Now, it is possible to modify, test and use this code from one set of source files.)

I will update the create-android-project.sh/.bat scripts to use this structure.


minggo:
Sounds good, thank you.
I think i will merge it after you correct the shell script.
Would you modify it in the pull request or in another one?

folecr:
I will update the shell script in my next commit tomorrow. Could you
merge it after?

On May 22, 2012, at 7:22 PM, minggo
reply@reply.github.com
wrote:

minggo:
That's ok.
Thank you.

folecr:
Updated!

Please review the .bat file before merge : I haven't been able to test it on Windows.

minggo:
Thank you.
I will test ASAP.

minggo:
When i create android project by eclipse, the library of java codes, the java files under src/org/cocos2dx/lib are not included automatically.
How to include them?

folecr:
I used the documentation at
http://developer.android.com/guide/developing/projects/projects-eclipse.html#ReferencingLibraryProject
However, the method to use library projects in eclipse is not clearly described by Google. These are the steps that worked for me :
1. Open the library project in eclipse
2. Open the application project also
3. Open _Properties->Android_ in the application project, the library projected should be auto detected as a dependency
4. Set the build order correctly (yes, eclipse did not set this correctly!) _Properties->Java Build Path->Order and Export_. Move **Android Dependencies** to above the application project
5. Build. Your build will fail with errors. :-) Then rebuild (no changes necessary.) Your build will succeed. (This seems to be a bug in eclipse.)

Success!

Note : You also have to build the native library using build_native.sh to use cocos2dx.

minggo:
I think it is complex to developers, what do you think?

folecr:
hey... no I don't think it is complicated.
- Google has been fixing bugs in their Eclipse plugin - I updated to the latest plugin and the Eclipse bugs I mentioned above seem to have been fixed. So, now, only steps 1, 2, and 5 are needed.
- If you are an Eclipse user it is one more step
- If you build using Ant, there is no extra step, dependencies are resolved automatically.

Also, Google has said that they will be adding support for exporting library projects as a jar but haven't given a date (http://android-developers.blogspot.com/2011/10/changes-to-library-projects-in-android.html, http://tools.android.com/recent.)

As soon as Google officially supports this, we can use it to export a prebuilt jar for the cocos2dx Android library code making the entire build process much easier. (I don't think it is worth it to export a jar of the library code without official support from the Android build system.)

minggo:
As you know, it is complicated for developers to set up NDK environment. 
If we add these steps, it will become more complicated for those who want to use cocos2d-x to develop android program.
Use Ant to build project is ok, but we also need eclipse to debug java codes.

So is there any method to simplify it? How about use reference to these source codes, and add .class .project into source control?

folecr:
Unfortunately Eclipse has a bug where project files do not support relative paths.

folecr:
Please see my latest changes
- Removes use of resources - XML files and R.java - for layout. Uses Java code instead.
- Builds Android library as a jar
- Android project generation script and template updated to use this jar
- I've tested with Ant build and Eclipse.

Note : I added a "prebuilt" directory that stores a prebuilt jar so that the Android project generation script works out of the box on a fresh clone.

folecr:
With these changes, both types of developers can use the Android project from Eclipse and Ant
- Developers not interested in modifying Cocos can generate a project using the Android template and the Eclipse/ANT project needs no further modification!
- Developers who will modify Cocos can build automatically using ANT. These developers will need to manually add a reference to the Android library if using Eclipse. Since this class of developer is knowledgeable in Cocos2D-X I think this is a good compromise.

dumganhar:
Hi,Surith
Eclipse does support relative paths, just drag the folder which is common to your project in 'Package Explorer', eclipse will ask you to 'Copy files and folders', 'Link to files and folders' or 'link to files and recreate folder structure with virtual folders', select the second option(Link to files and folders). Then the folder will be added into your project with a arrow at the bottomright corner. 
But that 's not the end. You should make it can be compiled. PLZ go the the 'Properties' of your project, and select 'Java Build Path'-->'Source'-->'Add Folder'-->select the relative folder--> click ok. Now eclipse treat the relative folder as a source folder. Enjoy. :)

minggo:
I think it is better to use relative path in HelloWorld, HelloLua and tests.
All the projects refer to the same java codes under cocos2dx/platform/third_party/...

When creating a new project, we copy the java codes into the new project.
So the modification will not affect sample projects offered by cocos2d-x.

What's your opinion, Surith?

folecr:
@dumganhar @minggo 
I think we are targeting two kinds of users :
1. Developer of cocos2d-x
2. Developer of games using cocos2d-x

(Some users will be both)

For User 1 it is important to build directly from source and debug sources. This user knows how to use Eclipse and NDK and is willing to configure project files as @dumganhar suggested in his comment above.

For User 2, everything must work out of the box. I think that it is simpler for User 2 to receive a binary distribution of cocos2d-x. This will simplify build and make the build process very fast.

My commit is for java code to build a library and it will make both User 1's and User 2's work simpler, IMHO.
- User 1 can fix bugs easily because there is only one place to change code for Android.
- User 2 will receive a binary library and can start developing games by creating one Eclipse project.

(In future I think User 2 should receive binary library of C/C++ code also. I can work on this after two weeks.)

Is my thinking correct?

minggo:
Yes, you are right, there are two kinds of users.
But i think User 2 also want to see the resource codes. If we don't offer source codes,
we should have perfect document like Android or JDK.

And developer don't have to configure Eclipse and NDK, because we do it. What developer have to do
is just import the project, and all configuration are there.

folecr:
I've updated the scripts so that they copy the sources to the generated project's directory instead of relying on the Android library mechanism. (Since libraries require extra setup steps on Eclipse.) This should satisfy the user who wants a one-step process of generating and building a Cocos2D-X Android project.

HelloWorld, HelloLua and test.android continue to use the common code as a library so that there is no extra duplication of source files among these projects. Advanced users may use these as a template if they are interested in manually setting up an Android project that will refer to a library of Java code.

Please review and merge.

minggo:
Thank you.
I think i will modify it to include java codes automatically for HelloWorld, HelloLua and tests.

PR #911, opened 2012-05-23
Missing std:: namespace

edbartley:
The isnan functions are missing the std namespace scoping. Without the std:: namespace scope the code does not compile on xcode.


minggo:
In which situation did you meet the problem.
It is used in tests, and don't find this bug.

Thank you.

edbartley:
I'm using xcode 4.3.2 and the newest version of Cocos2d-x from the MASTER
branch, not GLES20. The ios HelloWorld project fails to compile with the
following error:

"Semantic Issue: Use of undeclared identifier 'isnan'; did you mean
'std::isnan'?"

This issue seems to be affecting the MASTER branch only. The GLES20 branch
is not affected.

On Wed, May 23, 2012 at 11:34 PM, minggo <
reply@reply.github.com

PR #917, opened 2012-05-27
Added the ability to disable denshion when using lua.  Fixed async texture cache not working on IOS

jreitman:
This is useful for people using FMOD or other audio libraries.  Just a define in CC config.


minggo:
Thank you.
But LuaCocos2d.cpp is generated automatically, so you can not modify this file directly.
I think you should modify corresponding .pkg file.

jreitman:
Ok.  You should take the semaphore change however as that is implement broken on iOS bug 1241 on your roadmap. Not only does it not work it causes a tread to consume 100% of CPU if accidentally used 

-- Jeff Reitman

minggo:
Yeap, i agree with you.
So could you please revert the modification of LuaCocos2d.cpp?

jreitman:
Will do tomorrow am and shoot you an email

-- Jeff Reitman

minggo:
Thank you.

jreitman:
Updated!

minggo:
Thank you.

PR #952, opened 2012-05-31
Fix for Xcode error "Reference to 'less' is ambiguous" in CCTouchDispatcher.cpp

NatWeiss:
This little pull request fixes the Xcode error "Reference to 'less' is ambiguous" in CCTouchDispatcher.cpp. The problem is that CCTouchDispatcher.cpp's static less() function is declared outside the cocos2d namespace, so it is ambiguous versus the std::less() function. Moving it into the cocos2d namespace and explicitly referencing it fixes the error.


minggo:
Do you mind to pull request to gles20 branch?
We move our focus on it forward.

NatWeiss:
Sure.

PR #958, opened 2012-06-01
Fixes the declaration of tgaLoadRLEImageData()

NatWeiss:
This fixes the declaration of tgaLoadRLEImageData() in cocos2dx/support/image_support/TGAlib.cpp to match it's implementation.

This causes the android build to be successful.


minggo:
Thank you.

minggo:
And i add you into AUTHORS
https://github.com/cocos2d/cocos2d-x/blob/gles20/AUTHORS#L89

NatWeiss:
Cool. Thank you!

PR #961, opened 2012-06-01
CCControlButton update. Added zoomOnTouchDown property and setPreferredS...

icewind:
Ported some functions from original CCControlExtensions repository. Added feature to set the preferred size of the background sprite and toggle zooming on TouchDown event


dumganhar:
Thanks for your contribution.

PR #965, opened 2012-06-03
Enhances CCTransitionScene to work even if there is no running scene.

NatWeiss:
When you first start up a Cocos2D iPhone app, you can fade in the first scene with a CCTransitionFade. This little enhancement allows Cocos2D-X to fade in the first scene as well.

The commit enhances CCTransitionScene to work even if there is no running scene. It creates a blank autorelease CCScene to transition from if necessary.


minggo:
Thank you.

PR #979, opened 2012-06-07
Override setOpacity to affect a background sprites

icewind:
Calling (CCControlButton*)button->setOpacity was affecting only label. Now opacity
value applies also to a background sprites


minggo:
Thank you.
I will merge it ASAP.

PR #980, opened 2012-06-08
Allows Cocos2D-X to be built for iOS including all cpp files

NatWeiss:
If one is upgrading an older Xcode project to the latest Cocos2D-X, they may find it easy to drag and drop the cocos2d-x/cocos2dx and cocos2d-x/CocosDenshion folders into their Xcode project, then remove any references to platforms besides iOS and remove any references to proj.\* folders.

One problem is that cocos2dx/platform/CCThread.cpp is included even though there is a cocos2dx/platform/ios/CCThread.mm. This commit wraps CCThread.cpp in "#if (CC_TARGET_PLATFORM != CC_PLATFORM_IOS)" to prevent the compiler from giving an error about CCThread's functions already being implemented.

The other problem is that both cocos2dx/extensions/CCBReader/CCBReader_v1.cpp and cocos2dx/extensions/CCBReader/CCBReader_v2.cpp are included. This causes the compiler to error about CCBReader's functions being already implemented. This commit fixes the problem by adding a "#define CCB_READER_VERSION 2" to CCBReader.h and then using "#if CCB_READER_VERSION == 1" and "#if CCB_READER_VERSION == 2" in the respective cpp files.

Now Cocos2D-X can be drag and dropped into a project and built without errors (providing the extraneous platform and project folders are removed).

### DIFF
--- a/cocos2dx/extensions/CCBReader/CCBReader.h
+++ b/cocos2dx/extensions/CCBReader/CCBReader.h
@@ -30,6 +30,8 @@
 #include "cocos2d.h"
 #include "CCBCustomClass.h"
 
+#define CCB_READER_VERSION 2
+
 NS_CC_EXT_BEGIN
 
 /**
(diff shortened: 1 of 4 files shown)


NatWeiss:
Sorry. This broke the android compile. Will fix and re-submit.

PR #989, opened 2012-06-08
Fixed a few release-mode Xcode warnings about unused parameters.

NatWeiss:
When building in release mode with Xcode, the assert calls get turned to void, so a few parameters go unused. This pull request fixes those unused parameter warnings and a few other such warnings.


minggo:
Thank you.

PR #1013, opened 2012-06-12
Updating to work with BB10

dducharme:
Updating the source code to work with BB10.  Involves updates to the screen calls as well as changing the accelerometer over to use the sensor api.  Also renamed all references from QNX to BlackBerry.


dumganhar:
Thanks. :)

PR #1033, opened 2012-06-15
Added automatic TMX scaling: SD to HD (or vice versa).

NatWeiss:
With Cocos2D iPhone, it was easy to implement your own CCTMXXMLParser because you could simply write a category of CCTMXMapInfo to override all the XML parsing. This allowed a programmer to write their own XML parser to scale up or down the tile sizes and positions in the code, rather than relying on an external tool to create HD TMX files.

This pull request adds complete, automatic SD to HD (or vice versa) functionality to the Cocos2D-X's TMX parsing code.

A person using Cocos2D-X as normal will not be affected by this change. A person can continue using the old method of manual HD TMX files.

If one wants to enable the automatic HD TMX files, they simply call a few methods to setup the TMX parsing to automatically scale up/down the tile sizes and object positions. Here's an example that works for iPad:

CCTMXTiledMap::setPositionScaling(2.0f); // double positions
CCTMXTiledMap::setTileSizeScaling(2.0f); // double tile sizes (set this to 4.0f if retina enabled)
CCTMXTiledMap::setWhetherPositionsAreInPixels(true); // make the positionAt function return positions in pixels (depends on your implementation)

Note that the get/setWhetherPositionsAreInPixels() function is a workaround solution for now. It is necessary because all of the tile sizes are in pixels, but CCTMXLayer's positionAt function returns positions in points. Calling CCTMXTiledMap::setWhetherPositionsAreInPixels(true) causes the positionAt function to return a position in pixels, and therefore match the tile sizes. A better solution would be to redo the all the TMX code to work either with pixels or points, like giving the CCTMXLayer a positionAt function (points) and a positionInPixelsAt function.

Note that this pull request also fixes a bug in the iOS CCFileUtils::fullPathFromRelativeFile(const char *pszFilename, const char *pszRelativeFile) function. The problem was that this function automatically added the HD suffixes to the pszRelativeFile, but neglected to add them to the pszFilename (which was actually more important in the case of TMX tileset parsing). This fix correctly loads the HD versions of tileset images.

Now a person can have Cocos2D-X automatically load HD versions of tileset images and automatically have all the tile sizes and positions scaled correctly. :)

### DIFF
--- a/cocos2dx/platform/android/CCEGLView.h
+++ b/cocos2dx/platform/android/CCEGLView.h
@@ -27,7 +27,6 @@ THE SOFTWARE.
 
 #include "cocoa/CCGeometry.h"
 #include "platform/CCEGLViewProtocol.h"
-#include "platform/CCPlatFormMacros.h"
 
 NS_CC_BEGIN
 


NatWeiss:
Hey, since no one has commented on this pull request and it accidentally got some additional commits, I'm closing it.

Can you give me some hints why it wasn't merged?

dumganhar:
Hi, NatWeiss,
Sorry for our delay on this pull request! We are busy on gles20 branch recently, and we will release a version tomorrow.
After that, We will focus on this PR. Thanks very much for improving cocos2d-x with us. :)

dumganhar:
Since this pull request can't be auto-merged, could you pull our latest code and send PR to us again?

PR #1034, opened 2012-06-15
CCBIReader fixes + building and running flawless on iOS and Android.

nicolasgramlich:
Added some fixes in the CCBIReader.
The CCBIReader and the CocosBuilderTest are now building and running flawless on iOS and Android.
Added some #define macros to reduce boilerplate code for CCBMemberVariableAssigner and CCBSelectorResolver.

### DIFF
--- a/tests/proj.ios/test.xcodeproj/project.pbxproj
+++ b/tests/proj.ios/test.xcodeproj/project.pbxproj
@@ -724,6 +724,8 @@
 		4ACB592F158AB71200E0D552 /* MenuTestLayer.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = MenuTestLayer.cpp; sourceTree = "<group>"; };
 		4ACB5930158AB71200E0D552 /* MenuTestLayer.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = MenuTestLayer.h; sourceTree = "<group>"; };
 		4ACB5931158AB71200E0D552 /* MenuTestLayerLoader.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = MenuTestLayerLoader.h; sourceTree = "<group>"; };
+		4ACB5938158BE3F300E0D552 /* ParticleSystemTestLayer.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ParticleSystemTestLayer.h; sourceTree = "<group>"; };
+		4ACB5939158BE3F300E0D552 /* ParticleSystemTestLayerLoader.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ParticleSystemTestLayerLoader.h; sourceTree = "<group>"; };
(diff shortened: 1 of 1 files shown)


cocos2dx/extensions/CCBReader/CCBMemberVariableAssigner.h
```diff
@@ -5,9 +5,16 @@
 
 NS_CC_EXT_BEGIN
 
+#define CCB_MEMBERVARIABLEASSIGNER_GLUE(TARGET, MEMBERVARIABLENAME, MEMBERVARIABLETYPE, MEMBERVARIABLE) if(pTarget == TARGET && pMemberVariableName->compare(MEMBERVARIABLENAME) == 0) { \
```
minggo, line 8: pTarget == TARGET?

For example, 

```
bool HelloCocosBuilderLayer::onAssignCCBMemberVariable(CCObject * pTarget, CCString * pMemberVariableName, CCNode * pNode) {
    CCB_MEMBERVARIABLEASSIGNER_GLUE(this, "mBurstSprite", CCSprite *, this->mBurstSprite);
    CCB_MEMBERVARIABLEASSIGNER_GLUE(this, "mTestTitleLabelTTF", CCLabelTTF *, this->mTestTitleLabelTTF);

    return false;
}
```

pTarget is "this", TARGET's type is CCObject*.
Through they pointer the same object, but their value is different.
When it is passed in as a pointer off CCObject, its value is adjusted.

minggo:
This pull request can not be merged automatically.
Could you please resolve it and repull it?

cocos2dx/extensions/CCBReader/CCBMemberVariableAssigner.h
nicolasgramlich, line 8: Argh, this language!!! o_O

Anyway, isn't there an upcast to CCObject \* happening, since in this case "this" is a subclass of CCObject*?

Or should I change to something like:

```
dynamic_cast<void*>(pTarget) == dynamic_cast<void*>(TARGET)
```

just to be sure?

nicolasgramlich:
@minggo I already did the merge locally on my machine. The problem is that there is some weird "one release too much" of a CCString happening. I tried tracking it down, but didn't have success =( It's hard to figure out where the problem was, since you guys are so productive =D

If you want, I can push the code and you can try tracking it down.

minggo:
About CCString problem, you can refer Jame Chen's pull request:
https://github.com/cocos2d/cocos2d-x/pull/1035

cocos2dx/extensions/CCBReader/CCBMemberVariableAssigner.h
minggo, line 8: Is it all passing parameters are type of CCObject*?
If so, i think you can change to 

```
dynamic_cast<CCObject*>(pTarget) == TARGET
```

dumganhar, line 8: I think this line has no problem, compiler will auto adjust the pointer and compare them.
I have made it run on mac-mini successfully, the reason is what I mention in my PR(https://github.com/cocos2d/cocos2d-x/pull/1035). We will test it on other computers. If no problem, we will merge it.

dumganhar:
Hello, nicolasgramlich,
We have added CCScrollView to extension. Also added CCScrollViewLoader, but it's not implemented since I am not familiar with cocosbuilder and ccbi's structure. Could you help us to implement it ? Thanks.

nicolasgramlich:
Sure, I will today at work. =)

Btw, what was the problem implementing the loader? Given the other loaders I hoped it would be quite easy to add new loaders. =/

---

Letters were harmed in this message!

dumganhar:
When I saw your implementation of CCScrollViewLoader, I have known that it's easy, :) 
Besides, I want to know whether CocosBuilder will support CCTableView?

PR #1035, opened 2012-06-16
issue #1330: merge PR#1034 and made it works on win32.

dumganhar:
Refers to https://github.com/cocos2d/cocos2d-x/pull/1034.

### DIFF
--- a/tests/tests/ExtensionsTest/CocosBuilderTest/HelloCocosBuilder/HelloCocosBuilderLayer.h
+++ b/tests/tests/ExtensionsTest/CocosBuilderTest/HelloCocosBuilder/HelloCocosBuilderLayer.h
@@ -33,12 +33,12 @@ class HelloCocosBuilderLayer
         virtual bool onAssignCCBMemberVariable(cocos2d::CCObject * pTarget, cocos2d::CCString * pMemberVariableName, cocos2d::CCNode * pNode);
         virtual void onNodeLoaded(cocos2d::CCNode * pNode, cocos2d::extension::CCNodeLoader * pNodeLoader);
 
-        virtual void onMenuTestClicked(cocos2d::CCObject * pSender, cocos2d::extension::CCControlEvent pCCControlEvent);
-        virtual void onSpriteTestClicked(cocos2d::CCObject * pSender, cocos2d::extension::CCControlEvent pCCControlEvent);
-        virtual void onButtonTestClicked(cocos2d::CCObject * pSender, cocos2d::extension::CCControlEvent pCCControlEvent);
-        virtual void onLabelTestClicked(cocos2d::CCObject * pSender, cocos2d::extension::CCControlEvent pCCControlEvent);
-        virtual void onParticleSystemTestClicked(cocos2d::CCObject * pSender, cocos2d::extension::CCControlEvent pCCControlEvent);
-        virtual void onScrollViewTestClicked(cocos2d::CCObject * pSender, cocos2d::extension::CCControlEvent pCCControlEvent);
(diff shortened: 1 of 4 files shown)


dumganhar:
Hi, nicolasgramlich,
I have modified some codes, and made it works both on android and win32. Also, fixed some memory issues.
I have no mac for testing at home(It's weekend now. :) ). So could you test ios port again based on my pull request. If no bugs, I will merge it.Thanks.

minggo:
The app will crash when i touch a menu button of CocosBuilder test case.

minggo:
Except the menu "scroll view", because it does nothing now.

dumganhar:
I changed the order of inheritance for some classes, for example, HelloCocosBuilderLayer.
CCLayer must be the first base class, if not , it will crash on win32. I don't understand why it will crash on ios.

dumganhar:
Maybe it is relevant to compiler?

tests/tests/ExtensionsTest/CocosBuilderTest/HelloCocosBuilder/HelloCocosBuilderLayer.h
```diff
@@ -17,23 +17,28 @@
...
 
-        virtual cocos2d::SEL_MenuHandler onResolveCCBCCMenuSelector(cocos2d::CCObject * pTarget, const char * pSelectorName);
-        virtual cocos2d::extension::SEL_CCControlHandler onResolveCCBCCControlSelector(cocos2d::CCObject * pTarget, const char * pSelectorName);
-        virtual bool onAssignCCBMemberVariable(cocos2d::CCObject * pTarget, const char * pMemberVariableName, cocos2d::CCNode * pNode);
+        virtual cocos2d::SEL_MenuHandler onResolveCCBCCMenuItemSelector(cocos2d::CCObject * pTarget, cocos2d::CCString * pSelectorName);
+        virtual cocos2d::extension::SEL_CCControlHandler onResolveCCBCCControlSelector(cocos2d::CCObject * pTarget, cocos2d::CCString * pSelectorName);
+        virtual bool onAssignCCBMemberVariable(cocos2d::CCObject * pTarget, cocos2d::CCString * pMemberVariableName, cocos2d::CCNode * pNode);
         virtual void onNodeLoaded(cocos2d::CCNode * pNode, cocos2d::extension::CCNodeLoader * pNodeLoader);
 
-        virtual void onMenuTestClicked(cocos2d::CCObject * pSender, cocos2d::extension::CCControlEvent pCCControlEvent);
```
dumganhar, line 31: Callback function needn't  to be virtual function.

nicolasgramlich:
@dumganhar Btw, it works on iOS now!

dumganhar:
It's the reason that  'Callback function needn't to be virtual function'.

dumganhar:
For example, onMenuTestClicked, onSpriteTestClicked,onButtonTestClicked and so on.

nicolasgramlich:
Thanks a lot btw! I'm lacking substantial C++ experience and probably would have taken me days to find this issue!

dumganhar:
If you face other ugly bugs, just contact us. We are willing to help you.
Btw, you used std::string rather than CCString previously, and it would cause lots of crash. I found that it was the reason you used the 'const char\* pstr = (some functions return std::string).c_str();', after this line was executed, pstr would point to bad address since destruction of std::string appeared.

After that, you changed std::string to CCString, if CCString is an autorrelease object and haven't be retained, it will be destructed at the end of this game loop. So, it works fine.

nicolasgramlich:
Thanks. In the meantime I had a version that returned the std::strings and the const char \* returned by c_str was only used within the same method and it still crashed somewhere sometimes. With CCString everything is fine, even though it's a tiny overhead. =)

PR #1085, opened 2012-07-06
Remove wrong @Override to fix compilation error on newest Eclipse

ngocdaothanh:
Newest Eclipse reports compilation error with @Override


minggo:
I think it can not be removed for compatibility.
You can fix the error by doing this:
Project -> Properties -> Java Compiler -> Compiler compatibility level
select 1.6

minggo:
What's your Eclipse version?

minggo:
My Eclipse version is 3.7.2.
You can fix the problem by clean the project.

minggo:
Any progress?

ngocdaothanh:
I'm using Eclipse 3.7.2 on Mac.
It uses Java 1.5 by default.
You're right, setting "Compiler compatibility level" to 1.6 solves the @Override compilation error.

PR #1086, opened 2012-07-06
Remove wrong @Override to fix compilation error on newest Eclipse

ngocdaothanh:
Newest Eclipse reports compilation error with @Override


minggo:
I think this pull request is the same as https://github.com/cocos2d/cocos2d-x/pull/1085.

PR #1087, opened 2012-07-06
Fix compilation error: platform/android/CCEGLView.h:30

ngocdaothanh:
Fix compilation error:
platform/android/CCEGLView.h:30:39: error: platform/CCPlatFormMacros.h: No such file or directory

### DIFF
--- a/cocos2dx/platform/android/CCEGLView.h
+++ b/cocos2dx/platform/android/CCEGLView.h
@@ -27,7 +27,6 @@ THE SOFTWARE.
 
 #include "cocoa/CCGeometry.h"
 #include "platform/CCEGLViewProtocol.h"
-#include "platform/CCPlatFormMacros.h"
 
 NS_CC_BEGIN
 


minggo:
It is a typo error.
http://www.cocos2d-x.org/issues/1372 is created for it.

I don't know if it works just remove the include line.
I will try and merge it.

ngocdaothanh:
This affects the latest release:
http://cocos2d-x.org/news/62

For example, there will be error compiling the HelloWorld sample.
I guess people with less experience with Cocos2d-x will not know how to fix it themselves after downloading the release.
A new release should be made as soon as possible.

minggo:
It works on Windows and Mac OS X.
It can not work on linux, which is less user used.
So we will not release new version right now just for this issue.

But thank you all the way.

dumganhar:
Hi, all,
Since CCPlatformMacros.h has been included in CCGeometry.h, I think this line can be removed.

PR #1088, opened 2012-07-09
Gles20

mustime:
fixed #1379: fixed typing error in CCNode.cpp line 74


minggo:
Please create a separate branch for an issue, and don't commit unneeded codes, such as codes for test.

PR #1089, opened 2012-07-09
Fixes a few Xcode 4.5-specific error messages.

NatWeiss:
Using the newest Xcode 4.5, I get several errors (not warnings). For example, Xcode complains that it cannot convert -1 to an unsigned value. Simply casting the results of several expressions fixes the errors.


minggo:
Thank you.
I think i should update my Xcode to 4.5.

PR #1091, opened 2012-07-09
Fixes Win32 error "The application failed to initalize properly"

NatWeiss:
On Win32 XP SP3 with Visual Studio 2010, linking with the "incremental linking" option causes the runtime error message "The application failed to initalize properly 0xc000000d" in both Debug and Release configurations. The HelloWorld and test applications both fail to run because of this error.

See this forum post:
http://cocos2d-x.org/boards/6/topics/12188?r=12431

After reading Microsoft's documentation on the incremental linker. I see no reason why this is enabled for Cocos2D-X projects by default. Here's a link to their documentation:
http://msdn.microsoft.com/en-us/library/4khtbfyf(v=vs.80).aspx

This pull request turns off incremental linking and fixes the runtime error.


dumganhar:
Hi, Nat Weiss,
I have tested all projects on my win7. Both 'debug' and 'release' mode work fine.
Does this issue only appear on win xp ? 
And could you also fix this for other projects, such as HelloWolrd, HelloLua, tests ?

NatWeiss:
I only have XP to test with so there's no way for me to verify if it only appears on XP.

The other projects do not need this fix. Only the cocos2dx and CocosDenshion projects need to not be incrementally linked. Setting them to not be incrementally linked fixes all other projects like HelloWorld, tests, etc.

dumganhar:
OK, Thanks very much.

PR #1098, opened 2012-07-14
consistent the variable name in different functions.

walzer:


walzer:
Thanks, I fixed it in commit https://github.com/cocos2d/cocos2d-x/commit/56a3535dc1b35411bf938e712e894a8e66365aa8

PR #1103, opened 2012-07-19
Made all Android projects use the java files as a Android Library Project.

nicolasgramlich:
Made all Android projects use the java files as a Android Library Project, instead of linking to the src_common folder. Adapted the create-android-project.sh script (copy_files.sh) to inject a Android Library Project reference into the project.properties instead of copying/duplicating the src_common files.

Note: this is not a drop in PR, but the create-android-project.sh was tested a bit at least though ;)


template/android/copy_files.sh
```diff
@@ -70,10 +66,19 @@ modify_applicationdemo(){
...
+    sed "s/ApplicationDemo/$APP_NAME/;s/helloworld/game/;s/org\.cocos2dx\.application/$PACKAGE_PATH/" $APP_DIR/proj.android/src/org/cocos2dx/application/ApplicationDemo.java > $APP_DIR/proj.android/src/$PACKAGE_PATH_DIR/$APP_NAME.java
     rm -fr $APP_DIR/proj.android/src/org/cocos2dx/application
 }
 
+# modify project.properties
+modify_projectproperties(){
+    # fix library path
+	echo "" >> $APP_DIR/proj.android/project.properties
+
+	# TODO Eventually replace absolute with relative path?
```
nicolasgramlich, line 78: In my quick test the absolute path didn't work too well, so a relative path might have to be used.

nicolasgramlich:
Also some README might need to be updated to reflect the fact that the Android Library Projects are mandatory now?

dumganhar:
Hi, is this PR similar with #906 ?

nicolasgramlich:
It kind of makes complete what @folecr started in #906, but when using create_android_project the new project also has a library reference over copying the files.

dumganhar:
Thanks, I will test this PR and discuss with @minggo since I am not very familiar with java development.

folecr:
FYI, please see discussion in my previous pull request https://github.com/cocos2d/cocos2d-x/pull/906. @minggo made valuable points about usability of the generated project for users new to Android.

minggo:
@nicolasgramlich 

After your modification, what should i do to run HelloWorld, tests and so on?

nicolasgramlich:
After running build_native.sh, adding HelloWorld and the library project to Eclipse and running the HelloWorld should be sufficient.

minggo:
I am sorry, we have adjusted the directory structure.
So the pull request can not be merged automatically.

Now the developer can import all samples at one time, just import it.
So it is friendly to developers.
And i don't find any benefit to use the java files as a Android Library Project.

I may be wrong, please correct me.

nicolasgramlich:
The benefit of having it as a Library Project is simply that this is the way things are done correctly in the Android world. Imo copying all source files to the target is just plain wrong.

minggo:
It doesn't copy all source files to the target. It just reference to the source codes.
Because we put all java files to cocos2dx/platform/android/java. 
Suppose a developer is going to run a sample, it is hard to find the path.
May be we should put the java source files to a more suitable directory.

When we generate a android project, we copy the java source files to its project, because 
the developer may change the source code. And we don't want the modification affects engine's
samples.

What's your opinion?

minggo:
I am sorry, the java source files can be imported automatically.
But the pull request can not be merged automatically, could you modify it and repull a request.

And how do you think when a developer creates a new android project.
Should the new project use its own java files or just shares with samples?

nicolasgramlich:
Given my limited knowledge of cocos2d-x I think:

Since the java files coming with cocos2d-x are tightly bound to the native code, no game-developer should be touching them anyway. Which means they can sit in one central location within the cocos2d-x repo and are used by tests, games, etc...

minggo:
Yeap, i agree with you.
Could you modify the pull request?
Now it can not be merged automatically.

Thank you.

nicolasgramlich:
I will, tomorrow. :-)
On Jul 26, 2012 6:32 PM, "minggo" <
reply@reply.github.com>
wrote:

minggo:
Thank you.

minggo:
Should i close this pull request?

PR #1109, opened 2012-07-21
Adding Mac support!

NatWeiss:
Got Mac support to work with the latest branch of Cocos2d-X! :)

Resizing the Cocos2d-X Mac window works.

Full screen support works.

Audio works.

Touch input works.

Here's a video of it working:
http://www.youtube.com/watch?v=I5nqC3-GeLM

This pull request mainly features the cocos2dx/platform/mac and CocosDenshion/mac folders.

The cocos2dx/platform/mac folder started with the zip Walzer posted here: 
http://cocos2d-x.org/boards/6/topics/10352

I referenced a bit of ImperialPenguin's code, a lot of Cocos2d iPhone 2.0's and got it all working with the latest idioms of Cocos2d-X 2.0. The most difficult part was realizing that CCDirectorMac is not necessary. After incorporating it's main features into EAGLView and CCEGLView, everything started working amazingly. CCWindow is from Cocos2d iPhone and helps the EAGLView setFullScreen function. Input works too, thanks to CCEventDispatcher and setting it in EAGLView's initWithFrame.

There's one function in CCImage that I couldn't get to compile so I commented it out. It's potImageData function. I'm not sure where it belongs in the Cocos2d-X 2.0 paradigm, but the example game that I tried out did not need it to function 100%.

Most of the files in CocosDenshion/mac are a copy of CocosDenshion/ios except for CDXMacOSXSupport which comes from Walzer's zip. Copying them into the CocosDenshion/mac folder allows a person to compile cleanly for Mac without the ios folder, much like how the cocos2dx/platform folder works today.

This pull request was a lot of effort, so I will leave it up to somebody else to create the HelloWorld, tests, and proj.mac folders.
## 

Note that this is the second time I've submitted this pull request. This new request features a much improved CCFileUtils based on the platform/ios/CCFileUtils.m. It can detect high resolutions and load the appropriate -ipad graphics.

Also much improved is the mouse event handling in EAGLView. The previous pull request had a bug where the mouse up event was sometimes missed because it wasn't transmitting the correct id to Cocos2d-X's touch handler. This pull request uses the eventNumber as the id to accurately track the matching mouse up/dragged/down events.

Also, EAGLView's keyDown, keyUp, etc. methods which aren't actually doing anything call the super method, which effectively passes the event on to the next responder. In your typical Mac app, this will be the NSWindow. So you can catch keyUp and keyDown events in a custom NSWindow in your game's proj.mac folder.
## 

This is now the third pull request and the isIPad function of CCFileUtils now uses CCApplication's isIPad so a person can get that piece of information in their own Cocos2d-X app. Also, the isIPad boolean is determined statically in CCApplication so that if the window is resized, any new graphics loaded stay consistent with the old graphics which have already been loaded.

### DIFF
--- a/cocos2dx/platform/mac/CCApplication.mm
+++ b/cocos2dx/platform/mac/CCApplication.mm
@@ -27,6 +27,7 @@
 #import <Cocoa/Cocoa.h>
 
 #import "CCGeometry.h"
+#include "CCDirector.h"
 #import "CCDirectorCaller.h"
 
 NS_CC_BEGIN;
@@ -73,7 +74,16 @@ void CCApplication::statusBarFrame(cocos2d::CCRect * rect)
 
 bool CCApplication::isIpad()
 {
-	return false;
+	// save whether to load iPad resources so if the window is resized, we are consistent
+	static bool isSet = false;
+	static bool isIPad = false;
+	if( !isSet )
+	{
+		CCSize winSize = CCDirector::sharedDirector()->getWinSize();
+		isIPad = (winSize.width >= 1024 && winSize.height >= 768);
+		isSet = true;
+	}
+	return isIPad;
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


minggo:
It is an awesome work.
It is better if you can make tests work on mac.

PR #1130, opened 2012-07-26
issue #1284: using OpenSL for playing effects on I9100

mustime:


CocosDenshion/android/opensl/OpenSLEngine.cpp
```diff
@@ -0,0 +1,518 @@
...
+
+	return fd;
+}
+
+
+/**********************************************************************************
+ *   engine
+ **********************************************************************************/
+static SLObjectItf engineObject = NULL;
+static SLEngineItf engineEngine;
```
minggo: Why not initialize engineEngine?

Is it more readable if rename "engineObject"  to "g_pEngineObject"?

CocosDenshion/android/opensl/SimpleAudioEngineOpenSL.cpp
```diff
@@ -0,0 +1,122 @@
+#include "SimpleAudioEngineOpenSL.h"
+
+#define  PLAYSTATE_STOPPED 1
+#define  PLAYSTATE_PAUSED  2
+#define  PLAYSTATE_PLAYING 3
+#define  FILE_NOT_FOUND	   -1
+
+OpenSLEngine * s_pOpenSL = 0;
```
minggo: Should use static here?

minggo:
Please don't use OpenSL as default.
The implementation has some bugs.
The app will crash when add effect volume.

PR #1133, opened 2012-07-27
Fixed CCTouch IDs so they have unique values.

NatWeiss:
When using CCTouch objects and the ccTouchesBegan() / ccTouchesEnded() functions, it is very important to have a unique ID so that if multiple touches are enabled, an end touch correctly corresponds with a begin touch.

The way CCEGLProtocol was written, it calls CCTouch's setTouchInfo() with the index into it's own private static touches container as the ID of the CCTouch. Since the touches container is frequently cleaned up, this causes most of CCTouch IDs to be zero.

This pull request fixes that by using the provided id variable in the call to setTouchInfo(). Now, ccTouchesBegan() gets CCTouch objects with unique ID values that correctly correspond with ID values sent to ccTouchesEnded().

### DIFF
--- a/cocos2dx/platform/CCEGLViewProtocol.cpp
+++ b/cocos2dx/platform/CCEGLViewProtocol.cpp
@@ -214,12 +214,12 @@ void CCEGLViewProtocol::handleTouchesBegin(int num, int ids[], float xs[], float
             CCTouch* pTouch = s_pTouches[nUnusedIndex] = new CCTouch();
             if (m_bNeedScale)
             {
-                pTouch->setTouchInfo(nUnusedIndex, (x - m_rcViewPort.origin.x) / m_fScreenScaleFactor, 
+                pTouch->setTouchInfo(id, (x - m_rcViewPort.origin.x) / m_fScreenScaleFactor, 
                     (y - m_rcViewPort.origin.y) / m_fScreenScaleFactor);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


dumganhar:
Hi, NatWeiss,
I think you tested this on android since android platform will pass you collect id(0,1,2,3), 
but some other platform may not. On ios, we have codes bellow:

<pre>
- (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event
{
    int ids[CC_MAX_TOUCHES] = {0};
    float xs[CC_MAX_TOUCHES] = {0.0f};
    float ys[CC_MAX_TOUCHES] = {0.0f};
    
    int i = 0;
    for (UITouch *touch in touches) {
        ids[i] = (int)touch;
        xs[i] = [touch locationInView: [touch view]].x;
        ys[i] = [touch locationInView: [touch view]].y;
        ++i;
    }
    cocos2d::CCEGLView::sharedOpenGLView().handleTouchesBegin(i, ids, xs, ys);
}
</pre>


The ids parameter on ios means unique UITouch pointer. So we use getUnUsedIndex in CCEGLViewProtocol.cpp to create an ordinal id(0, 1, 2, 3 ...). 

This design is to encapsulate all platform touch logic. 

We haven't found this issue in our "MutiTouchTest", so could you please send us a demo to reproduce the issue mentioned in this PR?  Thanks very much.

minggo:
Can i close this pull request?

PR #1137, opened 2012-07-27
Cleaner build script

folecr:
I noticed you were cleaning up the directory structure. Now is a good time to clean up the build scripts too :)
I've modified the build_native.sh
- Avoid duplicate path variables
- Use environment variables after testing
- If not, use sensible defaults

References
- http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in
- http://stackoverflow.com/questions/228544/how-to-tell-if-a-string-is-not-defined-in-a-bash-shell-script

### DIFF
--- a/samples/HelloCpp/proj.android/build_native.sh
+++ b/samples/HelloCpp/proj.android/build_native.sh
@@ -1,6 +1,6 @@
-# set params
-NDK_ROOT_LOCAL=/home/james/Software/android/android-ndk-r8b
-COCOS2DX_ROOT_LOCAL=/home/james/Project/cocos2d-x
+APPNAME="HelloCpp"
+
+# options
 
 buildexternalsfromsource=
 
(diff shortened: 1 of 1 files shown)


minggo:
The script is more readable now, thank you.

I have a suggestion: not use "debug" and "verbose", and just pass parameter from console?
Because ndk-build accepts many parameters, like "clear", "debug=1" and so on.
I think we don't have to do the same job as ndk-build does.

folecr:
Yes, it's a personal preference if those options should be at the script level or whether the call to ndk-build should use the script's command line arguments ...

I'll remove that change (it's a separate commit) and then you can merge the pull request.

folecr:
Done. Please merge.

minggo:
Thank you.

Would you mind also changing other samples and android template.

folecr:
I could do that next week. I'll open a separate pull request.

minggo:
ok, thank you.

PR #1142, opened 2012-07-31
Api adjustment according to cocos2d JS API doc

walzer:


walzer:
Too much modifications to APIs.
Forget it.

PR #1147, opened 2012-08-01
Further clean up of android build scripts and template

folecr:
These changes match https://github.com/cocos2d/cocos2d-x/pull/1137

Note :
- Separate to my changes here, there are compilation errors in generated android projects due to an incorrect icon file.
- There are compilation errors in HelloLua - I think due to the Lua bindings not being updated correctly.

Since these issues are not related to my changes I am submitting them for merge. The above bugs need to be fixed separately.

### DIFF
--- a/template/android/build_native.sh
+++ b/template/android/build_native.sh
@@ -1,9 +1,6 @@
-# set params
-NDK_ROOT=__ndkroot__
-COCOS2DX_ROOT=__cocos2dxroot__
-GAME_ROOT=$COCOS2DX_ROOT/__projectname__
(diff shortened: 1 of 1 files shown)


minggo:
@folecr 

> > Separate to my changes here, there are compilation errors in generated android projects due to an incorrect icon file.

I didn't meet this error. I could create an android project after applying your pull request, and did't meet any compilation errors. Am i missing something?

folecr:
You are right. I tested now and the icons are created correctly. However in one case of previous testing "android create project" generated an icon named "ic_launcher.png" instead of "icon.png".

I'll post if I am able to reproduce this case.

PR #1155, opened 2012-08-04
Gles20 add asynchronous http request and its tests

greathqy:
HI Walzer, i had finished the CCHttpRequest and its tests file

I thought we may add tcp/udp facility to cocos2d-x in the future, so i had not remove the Http prefix from HttpRequestPacket&HttpResponsePacket for distinguishing reason.

And per our discussion at weibo, the addPostTask now have 3 overload for the user to choose, my experience is request via http, most of user will simply choose json format for data interchange, so the std::string & one will works fine. The others who want post binary data contains NULL byte can use the const char \* or vector<char> one.

For the response i think use std::string for responseData have no problem
User can use responseData->data() and responseData.length() on rare cases, at most time responseData.c_str() will suitable

Please review, Thanks


walzer:
Refactored and merged. Thanks for you contribution.

PR #1175, opened 2012-08-15
rewrite CCFileUtils, separate platform related source code

dayongxie:
在论坛上我提到过，关于底层平台相关类代码隔离的问题

 http://www.cocos2d-x.org/boards/6/topics/14281

CCFileUtils原先的实现中各个平台的代码没有实现充分的隔离，个人感觉采用include cpp代码的方式看着不太舒服，所以我把CCFileUtils部分代码重写了一下，实现部分改的地方不多，主要是代码结构做了一些调整，另外增加了windows的CCFileUtils的实现CCFileUtilsWin32，添加了平台相关接口。代码已在android, ios, mac, win32平台上测试通过。

望采纳，谢谢


cocos2dx/platform/win32/CCFileUtilsWin32.h
```diff
@@ -0,0 +1,25 @@
+#ifndef __CC_FILEUTILS_WIN32_H__
+#define __CC_FILEUTILS_WIN32_H__
+
+#include "platform/CCFileUtils.h"
+#include <windows.h>
+
+NS_CC_BEGIN
+
+class CC_DLL CCFileUtilsWin32 :
```
minggo, line 9: Why it calls CCFileUtilsWin32 on win32? Then can not use CCFileUtils in the codes, should use different class names on different platforms?

cocos2dx/platform/win32/CCFileUtilsWin32.h
```diff
@@ -0,0 +1,25 @@
...
+	public CCFileUtils
+{
+public:
+	CCFileUtilsWin32();
+
+	virtual const char* fullPathFromRelativePath(const char *pszRelativePath);
+
+	void setResourcePath(const char *pszDirectoryName);
```
minggo, line 17: setResourceDirectory()?

cocos2dx/platform/win32/CCFileUtilsWin32.h
```diff
@@ -0,0 +1,25 @@
...
+	const char* getResourcePath();
```
minggo, line 18: getResourceDirectory?

cocos2dx/platform/win32/CCFileUtilsWin32.h
dayongxie, line 9: No, the CCFileUtils::sharedFileUtils method will create an object of class CCFileUtilsWin32, but it returns CCFileUtils*, so it won't affect for user how to use it.

cocos2dx/platform/win32/CCFileUtilsWin32.h
dayongxie, line 17: The purpose of setResourcePath is different with setResourceDirectory

The method setResourceDirectory can specify the relative  path for the resource root, It's enough under IOS，android;
However under windows I want set the absolute path too.
So that's why I add class CCFileUtilsWin32 to provide the extention interface to do the thing;

dayongxie, line 17: On Windows, the default resource path is the directory of current directory, If I want to specify another directory, I should do this:
#include "CCFileUtilsWin32.h"

CCFileUtilsWin32\* fileutils = (CCFileUtilsWin32*)CCFileUtils::sharedUtils();
fileutils->setResourcePath("X:\Resource");

minggo:
Got it.
Because some other pull requests, this pull request can not been automatically merged.
Could you please resolve the conflict, then repull it?

Thank you.

dayongxie:
OK, I have resolved the conflict, please review.

minggo:
Thank you.
We will review it ASAP.

cocos2dx/platform/ios/CCFileUtils.mm
```diff
@@ -182,6 +182,11 @@ static void static_addValueToCCDict(id key, id value, CCDictionary* pDict)
     return m_obDirectory.c_str();
 }
 
+const char* CCFileUtils::getResourceDirectory()
```
minggo: This function is implemented here https://github.com/cocos2d/cocos2d-x/blob/gles20/cocos2dx/platform/ios/CCFileUtils.mm#L180

dayongxie:
OK, I have removed the function

cocos2dx/Android.mk
```diff
@@ -75,12 +75,13 @@ particle_nodes/CCParticleSystemQuad.cpp \
...
 platform/CCThread.cpp \
 platform/platform.cpp \
+platform/CCFileUtils.cpp \
 platform/CCEGLViewProtocol.cpp \
 platform/android/CCEGLView.cpp \
 platform/android/CCAccelerometer.cpp \
 platform/android/CCApplication.cpp \
 platform/android/CCCommon.cpp \
-platform/android/CCFileUtils.cpp \
+platform/android/CCFileUtilsAndroid.cpp \
```
minggo: Why should rename to CCFileUtilsAndroid.cpp?
I think the name is not consistent with other files, such as CCCommon.cpp.

cocos2dx/platform/win32/CCFileUtilsWin32.h
minggo, line 9: Linux port set resource root path in CCApplication.
Because this interface is platform dependent, so i think it is better to set resource root path at CCApplication.
And the application should set resource path at platform dependent codes, a good place is  proj.win32/main.cpp.

What's your opinion?

dayongxie:
I use to create a class CCFileUtilsAndroid like CCFileUtilsWin32, so renamed the file;
Now I renamed it to orignal.

minggo:
Can't CCFileUtilsWin32 be renamed to CCFileUtils?
Is there any reason?

dayongxie:
The file named CCFileUtilsWin32.cpp because I created a class CCFileUtilsWin32 which is a implementation for CCFileUtils on Windows, the class declared in CCFileUtilsWin32.h, but can't named with CCFileUtils.h, if I do that, the header file with the same filename would cause conflicts. 
That's my reason, although the file name will not be consistent with other files, I can't find any better way.

minggo:
Can declare CCFileUtilsWin32 in CCFileUtils.cpp?
Is class CCFileUtilsWin32 used in other places?

dayongxie:
CCFileUtilsWin32 provide some function only valuable for Windows, so It can't declare in CCFileUtils.cpp

minggo:
I mean platform/CCFileUtils.cpp.
1. rename platform/CCFileUtilsWin32.cpp to platform/CCFileUtils.cpp
2. declare CCFileUtilsWin32 in platform/CCFileUtils.cpp

dayongxie:
to 1. CCFileUtilsWin32.cpp is in folder platform/win32/, I can rename it to platform/win32/CCFileUtils.cpp, but I'm not sure if it is a good idear.

to 2. Class CCFileUtilsWin32 maybe used in other places, It can't be declared in a cpp file.

minggo:
Ok, got it.
Could you add some comment in CCFileUtilsWin32.h|.cpp to explain the reason?
Because it is a special case, and we will forget the reason.

Thank you.

minggo:
Linux port set resource root path in CCApplication.
Because this interface is platform dependent, so i think it is better to set resource root path at CCApplication.
And the application should set resource path at platform dependent codes, a good place is proj.win32/main.cpp.

What's your opinion?

PR #1181, opened 2012-08-15
Fix XCode 4.4 compiler warnings about string formatting.

psineur:
Subj.


walzer:
Aha, I knew your swear on twitter is about using 2dx :p
But I am in bed now. Can I check this issue tomorrow?

psineur:
Yeah, forced to use it to port a game.

Hah, of course ;)

BTW, if you hate C++ too, why not add Android support to Cocos2D-iPhone and stick with Objective-C?
Stella and some guy on cocos2d-iphone forum already did it and trying to sell, but i don't think this will work.
It must be Open-Source...

15.08.2012, â 18:57, Walzer íàïèñàë(à):

PR #1206, opened 2012-08-20
Update cocos2dx/platform/mac/CCGL.h

aninggo:
add VAO support

### DIFF
--- a/cocos2dx/platform/mac/CCGL.h
+++ b/cocos2dx/platform/mac/CCGL.h
@@ -30,4 +30,10 @@ THE SOFTWARE.
 
 #define CC_GL_DEPTH24_STENCIL8      -1
 
+
+#define glDeleteVertexArrays            glDeleteVertexArraysAPPLE
+#define glGenVertexArrays               glGenVertexArraysAPPLE
+#define glBindVertexArray               glBindVertexArrayAPPLE
+
+
 #endif // __CCGL_H__


cocos2dx/platform/mac/CCGL.h
```diff
@@ -30,4 +30,10 @@ of this software and associated documentation files (the "Software"), to deal
 
 #define CC_GL_DEPTH24_STENCIL8      -1
 
+
+#define glDeleteVertexArrays            glDeleteVertexArraysAPPLE
+#define glGenVertexArrays               glGenVertexArraysAPPLE
+#define glBindVertexArray               glBindVertexArrayAPPLE
+
+
```
minggo, line 38: Any benefit from this change?
Now mac port works well.

aninggo, line 38: enable CC_TEXTURE_ATLAS_USE_VAO  !  

#if CC_TEXTURE_ATLAS_USE_VAO
    //
    // Using VBO and VAO
    //
    glBindVertexArray( m_uVAOname );

minggo, line 38: So these codes should be changed too, right?
https://github.com/cocos2d/cocos2d-x/blob/gles20/cocos2dx/include/ccConfig.h#L175

aninggo, line 38: I think so.

minggo, line 38: So could you modify it too?
I think it is more suitable to modify it in this commit.

PR #1227, opened 2012-08-23
Gles20

donaldalanmorrison:
The changes remove LLVM Static Analyzer Warnings/Errors (only two remain.)

PR #1231, opened 2012-08-24
Gles20

donaldalanmorrison:
Fix for merge request #1227 (MSVC++2010 doesn't support C99-style named-initializers like clang.)


minggo:
Thank you.
A developer found this issue and revert it.
So now there is conflict in this file.
Could you please resolve the conflict and repull it?

donaldalanmorrison:
OK.  There was no conflict other than the commit history.  I've rebuilt and tested even! :-)

cocos2dx/kazmath/src/ray2.c
```diff
@@ -119,7 +119,7 @@ void calculate_line_normal(kmVec2 p1, kmVec2 p2, kmVec2* normal_out) {
 
 kmBool kmRay2IntersectTriangle(const kmRay2* ray, const kmVec2* p1, const kmVec2* p2, const kmVec2* p3, kmVec2* intersection, kmVec2* normal_out) {
     kmVec2 intersect;
-    kmVec2 final_intersect = {.x = 0, .y = 0}, normal = {.x = 0, .y = 0}; // Silencing LLVM SA
+    kmVec2 final_intersect = {0., 0.}, normal = {0., 0.}; // Silencing LLVM SA.
```
dumganhar, line 122: Why not just use kmVec2 final_intersect = {0};
As I know, {0} will initialize all value of struct to zero.

minggo:
Thank you.

donaldalanmorrison:
太棒了！欢迎您。

cocos2dx/kazmath/src/ray2.c
donaldalanmorrison, line 122: Perhaps a better fix to satisfy the LLVM SA would be: to not initialize (as before -- a possible speed-up), but also chain the if statements into if else clauses so that the Static Analyzer can see the values will be set.  最好的问候。

PR #1245, opened 2012-08-27
support auto generated JS bindings for cocos2d-x

folecr:


folecr:
For review. Please do not pull this yet.

minggo:
I have looked through the changes, and have some questions:
1. Java_org_cocos2dx_lib_Cocos2dxActivity.cpp
   This file includes <android/assert_manager.h> and <android/assert_manager_jni.h>.
   These header files only exist on Android v2.3 or later, but cocos2d-x should support Android version since v2.2.
   This file only receive asset manager object, but i didn't find where uses it.
2. Add a create function in CCAnimation
   
   static CCAnimation\* create(CCArray *arrayOfAnimationFrameNames, float delayPerUnit)
   {
       return CCAnimation::create(arrayOfAnimationFrameNames, delayPerUnit);
   }
   I don't know why this function is needed.
3. CocosDenshion/android/Android.mk seems to use older version. Could you please add your modification based on latest codes?
4. Use if(1) in CCMenuItem::activate(). Is it for testing?
5. It is better if the javascript binding can support function that supports default parameter value, then we don't have to add so many functions.

dumganhar:
Hi Surith,
I have successfuly made new bindings works on WINDOWS. There are serval changes I have made.
1. readFileInMemory function in ScriptCore is not necessary, we can use CCString::createWithContentOfFile instead. If in this way, Java_org_cocos2dx_lib_Cocos2dxActivity.cpp is not necessary too. Also, lots of ANDROID defination in ScriptCore can be removed.
2. CCCallFuncND doesn't need to inherit from TypeInfo since its super class CCCallFuncN has inherited it.
3. We should not make cocos2dx project depends on SpiderMonkey. The codes in CCObject::~CCObject() should be wrapped.  I have defined a function (removeJSObjectByCCObject) as follows:
   
   <pre>
   CCObject::~CCObject(void)
   {
   // if the object is managed, we should remove it
   // from pool manager
   if (m_bManaged)
   {
       CCPoolManager::sharedPoolManager()->removeObject(this);
   }
   
   // if the object is referenced by Lua engine, remove it
   if (m_nLuaID)
   {
       CCScriptEngineManager::sharedManager()->getScriptEngine()->removeCCObjectByID(m_nLuaID);
   }
   #ifdef COCOS2D_JAVASCRIPT
   CCScriptEngineManager::sharedManager()->getScriptEngine()->removeJSObjectByCCObject(this); 
   #endif
   }
   </pre>
   
   I don't know whether this name is suitable. If not, please correct me.
4. In cocos2d_specifics.cpp file, there is a calling named __builtin_trap in js_break function, I think this is a platform function, I comment it to make it been compiled ok.
5. Also in cocos2d_specifics.cpp,  I found a function calling named 'js_new_proxy', I can't find its implementation, so i changed it to JS_NEW_PROXY.
6. About 'js_log' function, I think using CCLOG to implement it is more appropriate.
7. All functions in ScriptingCore.h, such as, executeString, executeGlobalFunction... should return a default value, otherwise we can make it works on WIN32.
8. JS_GC invoking in ScriptingCore::forceGC will lead to crash.

For more details about my changes, please visit my branch: ( https://github.com/dumganhar/cocos2d-x/tree/surith_jsbinding ). 

Also, I can't make all js test works on windows. There are some crashes need to be fixed.

-- James Chen

folecr:
This pull request was too large for github to show the diffs. This was mainly because of one commit 
https://github.com/folecr/cocos2d-x/commit/51df4cd63fea48287374691bfb495fd3b80cde61

... that removes the spidermonkey source code. My intention was to eventually replace this with a mozilla submodule reference. 

I have now :
- rebased changes so that the updates to scripting/javascript are in a separate branch
- removed that commit from this set of changes
- squashed the commits that simply update the submodule tools/cxx-generator into one

You can see the diffs on github for this pull request now.

folecr:
@dumganhar 

> Also, I can't make all js test works on windows. There are some crashes need to be fixed.

Which version of spidermonkey did you use on Windows and which tests failed? Thanks.

folecr:
@minggo 

> It is better if the javascript binding can support function that supports default parameter value, then we don't have to add so many functions.

Unfortunately, right now the bindings generation does not support default parameters.

folecr:
@dumganhar 

> We should not make cocos2dx project depends on SpiderMonkey. The codes in CCObject::~CCObject() should be wrapped. I have defined a function (removeJSObjectByCCObject) as follows:

Agreed and I think your change is good.

For the future : I think it is a good idea to have a common scripting interface (Lua, JS, etc.)

folecr:
@minggo 

> Java_org_cocos2dx_lib_Cocos2dxActivity.cpp This file includes and . These header files only exist on Android v2.3 or later, but cocos2d-x should support Android version since v2.2. This file only receive asset manager object, but i didn't find where uses it.

Thanks for catching this. I've removed this change for now. These changes were originally for improving CCFileUtils on Android (redmine issues 1434, 1435, 1436.) I will discuss this separately.

folecr:
@minggo 

> Add a create function in CCAnimation
> 
> static CCAnimation\* create(CCArray *arrayOfAnimationFrameNames, float delayPerUnit)
> {
> return CCAnimation::create(arrayOfAnimationFrameNames, delayPerUnit);
> }
> I don't know why this function is needed.

To handle variable arguments : in this case when create() is called from JS with 2 arguments.

folecr:
@minggo 

> Use if(1) in CCMenuItem::activate(). Is it for testing?

Fixed. JS specific code should be within #ifdef COCOS2D_JAVASCRIPT.

folecr:
@minggo 

> CocosDenshion/android/Android.mk seems to use older version. Could you please add your modification based on latest codes?

Thanks. Fixed.

folecr:
@dumganhar 

Comments on your branch :
1. Instead of overwriting the Resources directory in samples/TestJavascript :
   - Add submodule pointing to cocos2d-js-test
   - Update the build.sh to copy js and Resources from within this folder into the package
   - This ensures that we are using the exact same code on cocos2d-iphone, cocos2d-x. What do you think?
2. What is the reason to remove reinterpret_cast() and replace it with a custom hash function?
3. There are many lines of code with stdint.h commented out. Could you please fix?

rohankuruvilla:
@minggo 

Fixed Garbage Collection crashes. Its stable on iOS and Android, even after running force GC at multiple random places in JS. merged with surith's repo, so part of this pull request.

Pending:
1. xcode template generation for cocos2dx-js is to be merged into this pull request. 
2. chipmunk bindings for android need to be merged into this pull request.

folecr:
@dumganahar, we have taken some of your changes for running JS on iOS. @rohankuruvilla has committed these changes into https://github.com/folecr/cocos2d-x/tree/run_js

This branch may be used for testing the js bindings.

I'll continue with updating the Android project tomorrow...

folecr:
I've updated the branch with commits from @rohankuruvilla, @dumganahar and me.
- Thanks for your comments - integrated changes where necessary.
- Test application samples/TestJavascript runs on iOS
- Test application samples/TestJavascript runs on Android

dumganhar:
Surith, thanks very much.
I will make it works ok on win32 ASAP and then send you a PR.
Since js-bindings is still not very stable, Minggo and I have considered to integrate the new js-binding in next version.
Today Walzer and Minggo have gone to Beijing to attend Developer Conference which held at Sep.1. We have to release a new version before Sep.1. So I am sorry for that.
BTW, Could you tell me what your skype ID is?

dumganhar:
@rohankuruvilla @folecr 
Oh, good news about the gc problem has been resolved. I will try it on win32. Great work.

folecr:
Updated with my latest change to apply the correct flags for all application sources.
js is now easy to build, easy to demo and ready on iOS and Android.

dumganhar:
@folecr 
OK, I found that you also fixed the scope issue, but I didn't find the update of submodule.

dumganhar:
@folecr @rohankuruvilla 
I found the update in @rohankuruvilla 's repo. Thanks.

dumganhar:
@folecr @rohankuruvilla @funkaster 
I have merged your codes, thanks a lot.

PR #1264, opened 2012-08-30
CCRGBAProtocol must be public in order for actions like CCTintTo to have...

edbartley:
... an affect on the CCScale9Sprite.


dumganhar:
Thank you.

PR #1265, opened 2012-08-30
fixed #1451: Merged new js-binding codes.

dumganhar:


pilwon:
In `CCLayer::excuteScriptTouchHandler(...)`,

``` c++
return CCScriptEngineManager::sharedManager()->getScriptEngine()->executeTouchEvent(m_pScriptHandlerEntry->getHandler(), nEventType, pTouch);
```

needs to be changed to

``` c++
#ifdef COCOS2D_JAVASCRIPT
    return CCScriptEngineManager::sharedManager()->getScriptEngine()->executeTouchEvent(1, nEventType, pTouch);
#else
    return CCScriptEngineManager::sharedManager()->getScriptEngine()->executeTouchEvent(m_pScriptHandlerEntry->getHandler(), nEventType, pTouch);
#endif
```

to avoid crash when targeted touch is used.

PR #1266, opened 2012-08-30
Fixed bug where texParams would get lost after reloading textures on and...

MartinSherburn:
Fixes issue described here: http://www.cocos2d-x.org/boards/6/topics/14749

If you make any calls to setAliasTexParameters, setAntiAliasTexParameters or setTexParameters then those changes are lost on android if you go to the home screen and back into the game because the OpenGL context is lot. This change caches those changes to VolatileTexture and re-applies them after the textures are reloaded.


minggo:
Thanks.
But i think this issue is resolved in this PR: https://github.com/cocos2d/cocos2d-x/pull/1198.

MartinSherburn:
That PR only fixes the issue for TMX layers (tiled map). My fix is much more general, user calls to setAliasTexParameters, setAntiAliasTexParameters or setTexParameters will get saved & restored as well which is what I happen to be doing in a few places, also the case described on the forum.

MartinSherburn:
I'm sure you can recreate this bug with the Texture2D tests. Load up the anti-aliased/aliased test, press home, then go back to the test.

minggo:
Thank you.
I think your solution is better.
So i will test and merge it.
And i will remove previous solution.

PR #1277, opened 2012-09-01
Adding VC2012 sln file and VS2012 changes to build-win32.bat & install-templates.msvc.bat

trentmillar:
I changed the ordering in the bat file to process newer VC++ versions before older ones. 

This was an issue for me running multiple versions of VS on my box.

Tested with VS2012 RTM.

### DIFF
--- a/build-win32.bat
+++ b/build-win32.bat
@@ -33,8 +33,9 @@ if %VC_VER%==90 (
     msbuild cocos2d-win32.vc2010.sln /p:Configuration="Debug" 
     msbuild cocos2d-win32.vc2010.sln /p:Configuration="Release"
 ) else if %VC_VER%==110 (
-    msbuild cocos2d-win32.vc2012.sln /p:Configuration="Debug" 
-    msbuild cocos2d-win32.vc2012.sln /p:Configuration="Release"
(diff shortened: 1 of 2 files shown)


minggo:
It doesn't have corresponding 2012 project, does it open corresponding 2010 project?

trentmillar:
Yes, the environment variables are checked so whatever version of Visual Studio the user has will be the solution that is used in the build. MSBuild will then target the correct version when compiling the project in release & debug.

If I understand correctly, the project files are version agnostic to begin with. Only until you actually open the solution file in Visual Studio will the version actually get added to each project and they are re-written by Visual Studio to match the correct version.

This is what is added to the vcxproj files:
 &lt;PlatformToolset&gt;v110&lt;/PlatformToolset&gt;

My changes are leveraging what was already used by cocos2d-x.

I did notice that the vcproj file is referring to, Version="9.00"  but it must have been this way for a long time and doesn't look to cause issues with VS2010 or 2012 so I left that.

trentmillar:
I went through my changes again and changed the version of the sln file, I had the incorrect version.

Also in the build-win3.bat file I added a Clean before targeting the build. Also I parallelized the release & debug builds

dumganhar:
Thanks.

PR #1278, opened 2012-09-01
Remove code signing from MAC TestCpp project.

edbartley:
People without signing certs can't build the project.


dumganhar:
Thanks. :)

PR #1279, opened 2012-09-02
min/max allowed value for CCControlSlider

elipsitz:
This sets the maximum and minimum values that can actually be selected for CCControlSlider, even if more are displayed.

### DIFF
--- a/extensions/GUI/CCControlExtension/CCControlSlider.cpp
+++ b/extensions/GUI/CCControlExtension/CCControlSlider.cpp
@@ -163,16 +163,6 @@ CCControlSlider* CCControlSlider::create(CCSprite * backgroundSprite, CCSprite* 
         m_minimumValue   = m_maximumValue - 1.0f;
      setValue(m_value);
  }
- 
-  void CCControlSlider::setMinimumAllowedValue(float minimumAllowedValue)
- {
-     m_minimumAllowedValue = minimumAllowedValue;
- }
-
-  void CCControlSlider::setMaximumAllowedValue(float maximumAllowedValue)
- {
-     m_maximumAllowedValue = maximumAllowedValue;
- }
 
   //this is the same as CCControl::getTouchLocation, but it returns the position relative to the position of this control
   CCPoint CCControlSlider::getTouchLocationInControl(CCTouch* touch)


minggo:
Why we need "allowed value"?
Because there are minimum and maximum values.

elipsitz:
Maybe you want to temporarily have only a part of the slider selectable. Or
perhaps you have a group of sliders whose allowed values depend on the
other sliders in the group (a maximum total value). In my game, I'm using
it for stat selection.

extensions/GUI/CCControlExtension/CCControlSlider.h
```diff
@@ -55,6 +55,11 @@ class CCControlSlider: public CCControl
     virtual void setMinimumValue(float val);
     CC_SYNTHESIZE_READONLY(float, m_maximumValue, MaximumValue);
     virtual void setMaximumValue(float val);
+    CC_SYNTHESIZE_READONLY(float, m_minimumAllowedValue, MinimumAllowedValue);
+    virtual void setMinimumAllowedValue(float val);
+    CC_SYNTHESIZE_READONLY(float, m_maximumAllowedValue, MaximumAllowedValue);
+    virtual void setMaximumAllowedValue(float val);
+
```
minggo, line 61: I'd suggest you not use macro to declare variable or member function.
Because it is hard to debug.
If you want to use macro, i think you should use CC_SYNTHESIZE here.

elipsitz, line 61: Alright, changed, I just did that because of the other things that were
already there.

extensions/GUI/CCControlExtension/CCControlSlider.h
```diff
@@ -55,6 +55,11 @@ class CCControlSlider: public CCControl
     virtual void setMinimumValue(float val);
     CC_SYNTHESIZE_READONLY(float, m_maximumValue, MaximumValue);
     virtual void setMaximumValue(float val);
+    CC_SYNTHESIZE(float, m_minimumAllowedValue, MinimumAllowedValue);
+    virtual void setMinimumAllowedValue(float val);
+    CC_SYNTHESIZE(float, m_maximumAllowedValue, MaximumAllowedValue);
+    virtual void setMaximumAllowedValue(float val);
+
```
minggo, line 61: Oh, CC_SYNTHESIZE already declares set function.

elipsitz, line 61: Oh, I see what you were saying: all I need is a generic getter/setter.

minggo:
Thank you.
It is better if you can add test case in samples/TestCpp/Classes/ExtensionsTest/ControlExtensionTest/CCControlSliderTest/.

elipsitz:
Added a test case.

minggo:
Could you pull the request for the test case?

PR #1280, opened 2012-09-03
Supported display Unicode chars.

Yodesoft:


cocos2dx/platform/linux/CCImage.cpp
```diff
@@ -195,7 +238,9 @@ class BitmapDC
 			return false;
 		}
 		do {
-			iError = FT_New_Face( library, pFontName, 0, &face );
+			//iError = FT_New_Face( library, pFontName, 0, &face );
+            //TODO: Create font name array???
+            iError = FT_New_Face( library, "/usr/share/fonts/truetype/wqy/wqy-microhei.ttc", 0, &face );
```
dumganhar: Why should use hard code of .ttc file path here ?
Will using 'pFontName' cause some issues?

Yodesoft: In normal usage , pFontName is family name, but Freetype need a full path of font to pass to FT_New_Face. So we need a function to handle font info, e.g.  get full path by family name or style stuff...

cocos2dx/platform/linux/CCImage.cpp
```diff
@@ -22,8 +22,23 @@
...
+};
+
+struct FontTableItem {
+    char* family_name;
+    char* style_name;
+    char* filename;
+};
+
+const int fontTableItems = 4;
+const char* fontPath = "/usr/share/fonts/truetype/";
```
minggo, line 36: What will happen if "usr/share/fonts/truetype/" doesn't exist?

Yodesoft:
if not any font exists, getBitmap will return FALSE.

/usr/share/fonts/truetype/  is a common path in Linux distribution for truetype font. If you want to add more special font, add to fontTable or improve it to support more path.

PR #1284, opened 2012-09-04
Change 2012 solution version number.

trentmillar:
The VS2012 sln should have it's "solution version" number changed from 11 to 12. Without this it will associate to Visual Studio 2010.

Note, this is not the same as the 11.0 main project version number. This is used for the solution file only.


trentmillar:
Actually looks like the sln file is messed up. It is the text duplicated twice. I will fix it tonight at home and create a new pull.

trentmillar:
Actually looks like the sln file is messed up. It is the text duplicated twice. I will fix it tonight at home and create a new pull.

PR #1285, opened 2012-09-05
Update samples/TestCpp/Classes/ExtensionsTest/ControlExtensionTest/CCCon...

elipsitz:
...trolSliderTest/CCControlSliderTest.cpp

Added a test for the restricted slider


minggo:
Thank you.
I think it is better to show lower slider's value, just like upper slider value.
What's your opinion?

elipsitz:
Made the lower slider's value show too.

minggo:
Thank you.

PR #1286, opened 2012-09-05
Fixed crash with CCLabelBMFont

moadib:
Fixed crash with CCLabelBMFont when center- or right-aligned text have line breaks each other

You can reproduce crash with

```
CCLabelBMFont* text = CCLabelBMFont::create("abc\n\n1", "font.fnt");
text->setAlignment(kCCTextAlignmentCenter);
addChild(text);
```

### DIFF
--- a/cocos2dx/label_nodes/CCLabelBMFont.cpp
+++ b/cocos2dx/label_nodes/CCLabelBMFont.cpp
@@ -1260,6 +1260,7 @@ void CCLabelBMFont::updateLabel()
             {
                 float lineWidth = 0.0f;
                 unsigned int line_length = last_line.size();
+				// if last line is empty we must just increase lineNumber and work with next line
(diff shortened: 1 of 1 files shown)


minggo:
I have tested the codes you pasted.
It doesn't crash.
You can test it with my branch "CCLabelBMFont_crash".

moadib:
Hmm :)

Try change 
`CCLabelBMFont* text = CCLabelBMFont::create("abc\n\n1", "font.fnt");`
to
`CCLabelBMFont* text = CCLabelBMFont::create("abc\n\n123", "font.fnt");`

minggo:
You are right.
Thank you.

minggo:
How about add some comment of these codes?
And i have created an issue for it http://www.cocos2d-x.org/issues/1464.

PR #1287, opened 2012-09-05
remove delete this from init methods

ivzave:
should not delete this in init methods because create() does it


minggo:
Thank you.

PR #1289, opened 2012-09-06
Addition to new Multi-resolution adaption scheme

moadib:
I've added ability to use `setContentScaleFactor()` in pair with `setDesignResolutionSize()` on all platforms.

Example from real life, we have "iphone" and "iphonehd" resource folders and want make our android game with this resources.

We won't make additional assets which adopted for 720x480 resolution. But just scale all with `
CCEGLView::sharedOpenGLView()->setDesignResolutionSize(480, 320, kResolutionShowAll);
` isn't as good as it can be. It is still use "iphone" resources

What i suggest?

``` cpp
if (pDirector->getWinSize().height > 320)
{
    CCFileUtils::sharedFileUtils()->setResourceDirectory("hd");
    pDirector->setContentScaleFactor(2.0f);  // attention! must be called before setDesignResolutionSize
}
else
    CCFileUtils::sharedFileUtils()->setResourceDirectory("normal");

CCEGLView::sharedOpenGLView()->setDesignResolutionSize(480, 320, kResolutionShowAll);
```

How it works?
It is also scaled to 720x480, but uses resources for retina. I think that everyone understand that downscaled texture looks better than upscaled :)
And it is work without any changes in game code.

Of course you still can use all as designed by developers.

### DIFF
--- a/cocos2dx/CCDirector.cpp
+++ b/cocos2dx/CCDirector.cpp
@@ -412,7 +412,8 @@ float CCDirector::getZEye(void)
     float zeye = m_obWinSizeInPixels.height / 1.1566f;
     if (m_pobOpenGLView->isRetinaEnabled())
         zeye /= CC_CONTENT_SCALE_FACTOR();
-       return zeye;    
+
+    return zeye;    
 }
 
 void CCDirector::setAlphaBlending(bool bOn)
--- a/cocos2dx/platform/win32/CCEGLView.cpp
+++ b/cocos2dx/platform/win32/CCEGLView.cpp
@@ -214,7 +214,7 @@ LRESULT CCEGLView::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
             if (m_bIsRetinaEnabled)
             {
                 pt.x /= CC_CONTENT_SCALE_FACTOR();
-			    pt.y /= CC_CONTENT_SCALE_FACTOR();
+                pt.y /= CC_CONTENT_SCALE_FACTOR();
             }
             CCPoint tmp = ccp(pt.x, m_obScreenSize.height - pt.y);
             if (m_obViewPortRect.equals(CCRectZero) || m_obViewPortRect.containsPoint(tmp))
--- a/cocos2dx/platform/ios/CCEGLView.mm
+++ b/cocos2dx/platform/ios/CCEGLView.mm
@@ -54,11 +54,14 @@ bool CCEGLView::setContentScaleFactor(float contentScaleFactor)
     if ([[EAGLView sharedEGLView] respondsToSelector:@selector(setContentScaleFactor:)])
     {
         UIView * view = [EAGLView sharedEGLView];
-        view.contentScaleFactor = contentScaleFactor;
-        [view setNeedsLayout];
-        
+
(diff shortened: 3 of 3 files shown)


cocos2dx/CCDirector.cpp
```diff
@@ -406,7 +409,10 @@ void CCDirector::purgeCachedData(void)
 
 float CCDirector::getZEye(void)
 {
-    return (m_obWinSizeInPixels.height / 1.1566f / CC_CONTENT_SCALE_FACTOR());    
+    float zeye = m_obWinSizeInPixels.height / 1.1566f;
+    if (m_pobOpenGLView->isRetinaEnabled())
+        zeye /= CC_CONTENT_SCALE_FACTOR();
+       return zeye;
```
minggo: How about add a line space before "return zeye"?

cocos2dx/platform/win32/CCEGLView.cpp
```diff
@@ -210,7 +210,12 @@ LRESULT CCEGLView::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
         if (m_pDelegate && MK_LBUTTON == wParam)
         {
             POINT point = {(short)LOWORD(lParam), (short)HIWORD(lParam)};
-            CCPoint pt(point.x/CC_CONTENT_SCALE_FACTOR(), point.y/CC_CONTENT_SCALE_FACTOR());
+            CCPoint pt(point.x, point.y);
+            if (m_bIsRetinaEnabled)
+            {
+                pt.x /= CC_CONTENT_SCALE_FACTOR();
+			    pt.y /= CC_CONTENT_SCALE_FACTOR();
```
minggo: How about format this line?

moadib:
Look at last changes, i I have considered your comments and added little changes for ios CCEGLView
Now we have run example code on New Ipad too.

Anyway, if you won't pick this pullrequest i insist that you move assignment of m_bIsRetinaEnabled to enableRetina(). It will be more clear from point of logic :)

minggo:
As you can see, in samples/HelloCpp, we also use hd resources for all android devices.

moadib:
Yes, but you have different "design resolutions" for iphone without retina and other. Also you don't use "normal" resources for android devices with 480x320 resolution.

The idea is to use the same design resolution as possible. As option, of course.

walzer:
moadib, I love this idea :)

moadib:
In our repos we have modified cocos2d so that remove word Retina from library and all classes. Only terms that we use in our logic are "design resolution" and "contet scale factor" :)

minggo:
Could you share your repos?
Is it https://github.com/moadib/cocos2d-x?

moadib:
No, it is in our local repo :) Ok, i think that @BorMor can share it later.

minggo:
Thank you.

minggo:
@moadib 
This pull request do similar work as you. And it does more work.
https://github.com/cocos2d/cocos2d-x/pull/1366

Could you please have a look?
And i think your pull request is better about how to use content scale factor.

moadib:
Take a look at pull-request of @BorMor (we work together) http://github.com/cocos2d/cocos2d-x/pull/1372

PR #1290, opened 2012-09-07
added lua support to linux platform

ivzave:
Overview:
- added liblua project & makefile
- added HelloLua project & makefile
- added TestLua project & makefile
- fixed make-all-linux-project.sh to build added stuff
- fixed build.php script for generating tolua++ bindings

### DIFF
--- a/tools/tolua++/build.php
+++ b/tools/tolua++/build.php
@@ -12,7 +12,14 @@
 defined('DS') or define('DS', DIRECTORY_SEPARATOR);
 define('SRC_DIR', __DIR__ . DS);
 define('OUT_DIR', dirname(__DIR__) . '/../scripting/lua/cocos2dx_support/');
-define('TOLUA_BIN', (DS == '\\') ? 'tolua++.exe' : '/usr/local/bin/tolua++');
+if (PHP_OS == 'Linux')
+{
+	define('TOLUA_BIN', '/usr/bin/tolua++5.1');
+}
+else
+{
+	define('TOLUA_BIN', (DS == '\\') ? 'tolua++.exe' : '/usr/local/bin/tolua++');
+}
 
 class Builder
 {
@@ -131,6 +138,11 @@ $replace = '
  THE SOFTWARE.
  ****************************************************************************/
 
+#ifdef __GNUC__
+#define __VA_COPY(to, from) va_copy((to), (from))
+#else
+#define __VA_COPY(to, from) ((to) = (from))
+#endif
 ';
 $FIX[$find] = $replace;
 
@@ -248,6 +260,15 @@ static int tolua_collect_LUA_STRING (lua_State* tolua_S)
 }
 EOT;
 $replace = '';
+$FIX[$find] = $replace;
+
+$find = <<<EOT
+  va_list args = *((va_list*)  tolua_tousertype(tolua_S,3,0));
+EOT;
+$replace = <<<EOT
+  va_list args;
+  __VA_COPY(args, *((va_list*)  tolua_tousertype(tolua_S,3,0)));
+EOT;
 $FIX[$find] = $replace;
 
 // main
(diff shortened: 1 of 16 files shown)


minggo:
Thank you.
I will test it ASAP.

ivzave:
I wish to update this but I need the new build.xml ant script (the current one in repo generates incorrect bindings). I can't contact YuLei since he doesn't own a profile on github. Can anybody ask him to commit the new ant script he used to generate bindings?

dumganhar:
Hi, ivzave,
Here is the YuLei's github address: https://github.com/dualface

ivzave:
Thanks.

tools/tolua++/build.php
```diff
@@ -250,6 +262,15 @@ private function _fixLuabindingFile()
 $replace = '';
 $FIX[$find] = $replace;
 
+$find = <<<EOT
+  va_list args = *((va_list*)  tolua_tousertype(tolua_S,3,0));
+EOT;
+$replace = <<<EOT
+  va_list args;
+  __VA_COPY(args, *((va_list*)  tolua_tousertype(tolua_S,3,0)));
```
dumganhar, line 270: Why should use _va_copy_? If not using it, it's also compiled ok.
What's more, it seems that **CCMenu::alignItemInColumns** cannot works correctly,
When app run into **tolua_Cocos2d_CCMenu_alignItemsInColumns00**, an error occurs.
Any ideas?

ivzave, line 270: GCC for 64-bit target defines va_list as an array so it can't be assigned directly. _va_copy_ is arch-independent way to copy va_lists in GCC.
How did you test _alignItemsInColumns_? Is lua able to generate C va_lists? I think there's no need in public va_list methods.

dumganhar, line 270: So did you just add _va_list_ to make it compiled ok on <code>Linux 64bit</code> machine?
Yep, i think tolua++ doesn't support _va_lists_ .

ivzave, line 270: Yes, _va_copy_ is for linux 64-bit compatibility. If we remove
`void alignItemsInColumns(unsigned int columns, va_list args);`
`void alignItemsInRows(unsigned int rows, va_list args);`
from CCMenu.pkg then there's no need in this fix.

dumganhar, line 270: OK, i will contact **YuLei** and ask him whether we should delete these two functions.
Thanks.

dumganhar, line 270: Hey, i tested it on my ARCHLINUX 64bit machine, but no errors occured without using _va_copy_.
What's the distribution of linux you are using?

ivzave, line 270: Ubuntu 12.04, gcc 4.6.3. I can test it again later today and report any problems.

dumganhar, line 270: OK, thank you.

ivzave, line 270: Still getting errors with stock LuaCocos2d.cpp:
`../../../scripting/lua/cocos2dx_support/LuaCocos2d.cpp: In function ‘int tolua_Cocos2d_CCMenu_alignItemsInColumns00(lua_State*)’:
../../../scripting/lua/cocos2dx_support/LuaCocos2d.cpp:41885:61: error: array must be initialized with a brace-enclosed initializer
../../../scripting/lua/cocos2dx_support/LuaCocos2d.cpp: In function ‘int tolua_Cocos2d_CCMenu_alignItemsInRows00(lua_State*)’:
../../../scripting/lua/cocos2dx_support/LuaCocos2d.cpp:41920:61: error: array must be initialized with a brace-enclosed initializer`

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/1752.
Thanks.

Issue #1291, opened 2012-09-07
setDesignResolutionSize crash in 2.0.2


moremorefun:
ver 2.0.2
call #setDesignResolutionSize# in main func like 
# 

AppDelegate app;
CCEGLView\* eglView = CCEGLView::sharedOpenGLView();
eglView->setFrameSize(960, 540);
eglView->setDesignResolutionSize(960, 540, kResolutionShowAll);
int ret = CCApplication::sharedApplication()->run();
# 

in #setDesignResolutionSize# will call #CCDirector::sharedDirector()->setGLDefaultValues();#

but at the func #setGLDefaultValues#, in that time, #m_pobOpenGLView#  is not inited.
so it fail in #CCAssert(m_pobOpenGLView, "opengl view should not be null");#

I don't no  #CCDirector::sharedDirector()->setGLDefaultValues();# is necessar,
but i comment it, and It work just fine.

moremorefun:
My fault , it shoud be called in AppDelegate not main.

PR #1292, opened 2012-09-07
fix setDesignResolutionSize crash

moremorefun:
When setDesignResolutionSize call setGLDefaultValues function,
but at that moment m_pobOpenGLView is NULL.
m_pobOpenGLView is initialization when CCApplication::sharedApplication()->run(),
at that time setGLDefaultValues will be called too.

check the issues/1291, https://github.com/cocos2d/cocos2d-x/issues/1291

I am new here, so I am not sure the change is OK.

Thank you.


minggo:
We should call setGLDefaultValues() to make setDesignResolutionSize() take effect.

moremorefun:
In win and android, at this situation, 'CCAssert(m_pobOpenGLView, "opengl view should not be null")' will fail.
How can I fix it or I do something wrong to call the setDesignResolutionSize ?

moremorefun:
I see, I should call the setGLDefaultValues in AppDelegate
Sorry for this.

PR #1301, opened 2012-09-10
Update location of spidermonkey-android makefile

folecr:
A more sensible location for the android makefile

Allows user of spidermonkey to write :

```
$(call import-module,scripting/javascript/spidermonkey-android)
```

instead of 

```
$(call import-module,scripting/javascript/spidermonkey-android/android)
```


dumganhar:
Thanks. :)

minggo:
Oh, can this work?
It is magic.

PR #1306, opened 2012-09-11
Fix for VC 2012 v110 Runtime

martell:
Fix for the issue i created earlier.

Can you guys check that visual studio 2010 and 2008 isn't broken before merging the code base?

I'm on windows 8 and I don't have other visual studio versions on this desktop.

On another note I'm working on a cmake build system for generating all the project files for all platforms.

Working so far is

Mingw64
Visual Studio All versions.
Android building.

Have to test IOS, MAC and linux before I create a pull for it.
Hopefully it will be before 2.04 version.


dumganhar:
I think you are right,  vs2012 will automatically convert vs2010's '.vcxproject' files and add 'v110' target.
So could we close this PR?

martell:
I think you guys are right. You can close it. Thanks johnangel.
Maintaining the different versions seems quite a hassle.
I have implemented a cmake solution for cocos2dx, It works on android and win32 atm.
Would you guys accept a merge that uses cmake?

PR #1312, opened 2012-09-13
OpenGP framebuffer access violation fix + fixing include paths

ivfit:
1) Even if the good opengl version is available, framebuffer support is done as an extension, not in the native opengl. This fix should bind dynamically all the framebuffers methods from the correct ARB/EXT.  Ref: tito@kivy, https://github.com/kivy/kivy/commit/c76fed7440a9fb7470c57a8387021c3afdeaf2bf

2) Fixing include paths to be able to compile cocos2d-x using external Visual Studio solution. This way we can have completely separate VS solution containing our project + referencing original cocos2d-x projects, without need to copy all the files.

### DIFF
--- a/cocos2dx/platform/win32/CCEGLView.cpp
+++ b/cocos2dx/platform/win32/CCEGLView.cpp
@@ -335,7 +335,7 @@ LRESULT CCEGLView::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
         if (m_pDelegate && MK_LBUTTON == wParam)
         {
             POINT point = {(short)LOWORD(lParam), (short)HIWORD(lParam)};
-            CCPoint pt(point.x, point.y);
+            CCPoint pt(point.x/CC_CONTENT_SCALE_FACTOR(), point.y/CC_CONTENT_SCALE_FACTOR());
             CCPoint tmp = ccp(pt.x, m_obScreenSize.height - pt.y);
             if (m_obViewPortRect.equals(CCRectZero) || m_obViewPortRect.containsPoint(tmp))
             {
@@ -356,7 +356,7 @@ LRESULT CCEGLView::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
         if (MK_LBUTTON == wParam && m_bCaptured)
         {
             POINT point = {(short)LOWORD(lParam), (short)HIWORD(lParam)};
-            CCPoint pt(point.x, point.y);
+            CCPoint pt(point.x/CC_CONTENT_SCALE_FACTOR(), point.y/CC_CONTENT_SCALE_FACTOR());
             int id = 0;
             pt.x *= m_windowTouchScaleX;
             pt.y *= m_windowTouchScaleY;
@@ -371,7 +371,7 @@ LRESULT CCEGLView::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
         if (m_bCaptured)
         {
             POINT point = {(short)LOWORD(lParam), (short)HIWORD(lParam)};
-            CCPoint pt(point.x, point.y);
+            CCPoint pt(point.x/CC_CONTENT_SCALE_FACTOR(), point.y/CC_CONTENT_SCALE_FACTOR());
             int id = 0;
             pt.x *= m_windowTouchScaleX;
             pt.y *= m_windowTouchScaleY;
@@ -394,18 +394,22 @@ LRESULT CCEGLView::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
               for (UINT i=0; i < cInputs; i++)
               {
                 TOUCHINPUT ti = pInputs[i];
-                CCPoint pt(TOUCH_COORD_TO_PIXEL(ti.x)/CC_CONTENT_SCALE_FACTOR(), TOUCH_COORD_TO_PIXEL(ti.y)/CC_CONTENT_SCALE_FACTOR());
+                POINT input;
+                input.x = TOUCH_COORD_TO_PIXEL(ti.x);
+                input.y = TOUCH_COORD_TO_PIXEL(ti.y);
+                ScreenToClient(m_hWnd, &input);
+                CCPoint pt(input.x/CC_CONTENT_SCALE_FACTOR(), input.y/CC_CONTENT_SCALE_FACTOR());
                 CCPoint tmp = ccp(pt.x, m_obScreenSize.height - pt.y);
                 if (m_obViewPortRect.equals(CCRectZero) || m_obViewPortRect.containsPoint(tmp))
                 {
                   pt.x *= m_windowTouchScaleX;
                   pt.y *= m_windowTouchScaleY;
 
-                  if((ti.dwFlags & TOUCHEVENTF_DOWN)!=0)
+                  if (ti.dwFlags & TOUCHEVENTF_DOWN)
                     handleTouchesBegin(1, reinterpret_cast<int*>(&ti.dwID), &pt.x, &pt.y);
-                  else if((ti.dwFlags & TOUCHEVENTF_MOVE)!=0)
+                  else if (ti.dwFlags & TOUCHEVENTF_MOVE)
                     handleTouchesMove(1, reinterpret_cast<int*>(&ti.dwID), &pt.x, &pt.y);
-                  else if((ti.dwFlags & TOUCHEVENTF_UP)!=0)
+                  else if (ti.dwFlags & TOUCHEVENTF_UP)
                     handleTouchesEnd(1, reinterpret_cast<int*>(&ti.dwID), &pt.x, &pt.y);
                 }
               }


cocos2dx/platform/CCEGLViewProtocol.h
```diff
@@ -64,7 +64,7 @@ class CC_DLL CCEGLViewProtocol
     /**
      * Set the frame size of EGL view.
      */
-    virtual void setFrameSize(float width, float height);
+    virtual bool setFrameSize(float width, float height);
```
dumganhar: What situation will this method return false?

ivfit: It will return false if it fails to create app window or initialize OpenGL.

See Windows implementation of CCEGLView::setFrameSize:
https://github.com/johnangel/cocos2d-x/blob/d071a5e5693dd7c0b517c61f5dea3043ca2c2cb2/cocos2dx/platform/win32/CCEGLView.cpp

More info in my comment:
https://github.com/cocos2d/cocos2d-x/pull/1312#commitcomment-1877279

cocos2dx/platform/win32/CCFileUtils.cpp
```diff
@@ -168,17 +169,45 @@ unsigned char* CCFileUtils::getFileData(const char* pszFileName, const char* psz
...
+
+	// Get filename of executable only, e.g. MyGame.exe
+	char *base_name = strrchr(full_path, '\\');
+
+	if(base_name)
+	{
+		char app_data_path[_MAX_PATH + 1];
+
+		// Get local app data directory, e.g. C:\Documents and Settings\username\Local Settings\Application Data
+		if (SUCCEEDED(SHGetFolderPathA(NULL, CSIDL_LOCAL_APPDATA, NULL, SHGFP_TYPE_CURRENT, app_data_path)))
```
dumganhar: Why should we place wirteable files to 'App Data Directory' ? Is it because users need publish their app to desktop?

cocos2dx/platform/CCEGLViewProtocol.h
dumganhar: I think we should refactor CCEGLView for win32 port.
When we first invoke 'CCEGLView::sharedEGLView()' , we should create the window and initialize GL environment. If it fails when initializing GL, it just return NULL.

setFrameSize just resize the window size, so it needn't to have a return value.
What about your opinion?

cocos2dx/platform/win32/CCFileUtils.cpp
ivfit: Starting from Windows Vista it is not possible to write C:\Program Files without admin privileges.

So if the game is installed in C:\Program Files (which is usually the case), it will try to do that and end up in wrong folder (Virtual Store).

App Data is proper way to do such things. 

Good explanation what happens is in chapter 6. Virtual Store:
http://blogs.msdn.com/b/chinmay_palei/archive/2011/01/16/windows-7-application-compatibility-issues-fix-centre.aspx

This way app can be installed properly in C:\Program Files and writing will work as Microsoft designed (to C:\Documents and Settings\username\Local Settings\Application Data).

cocos2dx/platform/CCEGLViewProtocol.h
ivfit: If I undersood correctly your idea is to move:

if(!Create((LPCTSTR)m_szViewName, (int)width, (int)height))
{
PostQuitMessage(-1);
return false;
}

from CCEGLView::setFrameSize function to CCEGLView::sharedOpenGLView()?

Then in main.cpp after:

CCEGLView\* eglView = CCEGLView::sharedOpenGLView();
we should add something like:
if(eglView == NULL) return -1;

cocos2dx/platform/win32/CCFileUtils.cpp
dumganhar: But win32 port is mainly for debugging.  Developers usually put cocos2d-x engine to a non-C dist.  And some writeable config files can easily be found at the root directory of Debug.win32 folder.  They may want to modify these files for debugging.
In your case,  developers will need to find their config files at a very long path(C:\Documents and Settings\username\Local Settings\Application Data....). Is there any other ways to achieve that?

cocos2dx/platform/CCEGLViewProtocol.h
dumganhar: Yes, you're right. So could you have a try for this?

cocos2dx/platform/win32/CCFileUtils.cpp
ivfit: We can add if Debug mode (#ifdef _DEBUG) then don't write to AppData, but for Release mode do use AppData?

Or make this configurable by setting some variable?

cocos2dx/platform/CCEGLViewProtocol.h
ivfit: Sure thing, I will try this later today and let you know if it works fine.

cocos2dx/platform/win32/CCFileUtils.cpp
dumganhar: Oh, you got it. I think using debug macro is better.

cocos2dx/platform/CCEGLViewProtocol.h
dumganhar: OK, thanks a lot. :)

cocos2dx/platform/win32/CCFileUtils.cpp
ivfit: I will commit such change later, also.

dumganhar: OK, don't forget to pull our latest codes before you do some changes. :)

cocos2dx/platform/CCEGLViewProtocol.h
ivfit: Fixed and commited.

cocos2dx/platform/win32/CCFileUtils.cpp
ivfit: Fixed and commited.

scripting/javascript/bindings/generated
```diff
@@ -1 +0,0 @@
-Subproject commit bb14056784ea0d3ab3f30ae5a86c1bd3ad483703
```
dumganhar: Why did you delete this submodule?

tools/cxx-generator
```diff
@@ -1 +0,0 @@
-Subproject commit 40ad79373b6c041a3312afb36da6985b91f0a3dc
```
dumganhar, line 1: Why did you delete this submodule?

cocos2dx/platform/win32/CCEGLView.cpp
```diff
@@ -457,13 +531,15 @@ void CCEGLView::resize(int width, int height)
         m_windowTouchScaleX = frameSize.width / width;
         m_windowTouchScaleY = frameSize.height / height;
 
+/*
         TCHAR buff[MAX_PATH + 1];
         memset(buff, 0, sizeof(buff));
         swprintf_s(buff, MAX_PATH, L"%s - %0.0fx%0.0f - %0.2f",
                    kWindowClassName, frameSize.width, frameSize.height, 1.0f / m_windowTouchScaleX);
         SetWindowText(m_hWnd, buff);
+*/
```
dumganhar: Why did you comment these lines?

tools/cxx-generator
ivfit, line 1: This is not intentional and shouldn't be deleted.

Tried:
git merge upstream/gles20
Already up-to-date.

Tried:
git checkout upstream/gles20 generated

And it always generates empty directory, don't know how to fix it... Maybe subproject support is missing in Eclipse EGit. Do you recommend some GIT client for Windows?

scripting/javascript/bindings/generated
ivfit: This is not intentional and shouldn't be deleted. See my other comment.

cocos2dx/platform/win32/CCEGLView.cpp
ivfit: I've just commited a fix to show this info in Debug mode.

cocos2dx/platform/win32/CCEGLView.cpp
```diff
@@ -209,11 +279,10 @@ bool CCEGLView::Create(LPCTSTR pTitle, int w, int h)
 
         CC_BREAK_IF(! m_hWnd);
 
-        resize(w, h);
-
```
dumganhar, line 213: Hey, did you find that after removing 'resize(w,h);' in 'Create' method, the created window's height and width are both zero before invoking 'setFrameSize'?  The touch and display area will be wrong.
I fixed this by make the width and height to 100 by default. But I don't think this is a good solution. Could you check this issue?

ivfit, line 213: setFrameSize must be called to set proper window size. The same thing was necessary before this commit.

tools/cxx-generator
dumganhar, line 1: I use TortoiseGit and SmartGit for Windows. And sometimes I use the bash.

cocos2dx/platform/win32/CCEGLView.cpp
```diff
@@ -271,6 +381,42 @@ LRESULT CCEGLView::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
...
+                {
+                  pt.x *= m_windowTouchScaleX;
+                  pt.y *= m_windowTouchScaleY;
+
+                  if((ti.dwFlags & TOUCHEVENTF_DOWN)!=0)
+                    handleTouchesBegin(1, reinterpret_cast<int*>(&ti.dwID), &pt.x, &pt.y);
+                  else if((ti.dwFlags & TOUCHEVENTF_MOVE)!=0)
+                    handleTouchesMove(1, reinterpret_cast<int*>(&ti.dwID), &pt.x, &pt.y);
+                  else if((ti.dwFlags & TOUCHEVENTF_UP)!=0)
+                    handleTouchesEnd(1, reinterpret_cast<int*>(&ti.dwID), &pt.x, &pt.y);
```
dumganhar, line 413: Hey, by reviewing your codes, I found that win7 or win8 devices support multi-touch.
But i think your codes about dealing with multi-touch is wrong.
You should collect all touch events in one WM_TOUCH event, then invoking 'handleTouchesBegin(numberOfTouches, id array, posX array, posY array);'

dumganhar, line 413: After you make it works, please make sure that 'MultiTouchTest' in 'TestCpp' can also work. Thanks.

cocos2dx/platform/win32/CCEGLView.cpp
ivfit, line 213: This is corrected by making initial window size 480x320.

cocos2dx/platform/win32/CCEGLView.cpp
ivfit, line 413: MultiTouchTest is working properly now on Win8 Ultrabook.

There is no need to collect all touch events (i.e. make 9 different dynamic arrays), because handleTouches\* functions also iterate over every event the same way.

dumganhar, line 413: I don't think so, in 'handleTouchesBegin', we just add all touches to CCSet, at the end of 'handleTouchesBegin', we only invoke 'm_pDelegate->touchesBegan(&set, NULL);' once.

dumganhar:
Hey, I picked some of your changes at PR #1465. Thank you.

PR #1320, opened 2012-09-15
Fixes CCThread.cpp so it is not included for Mac builds.

NatWeiss:
The Mac platform already has it's own CCThread.mm, so it's necessary to add an include guard if one is dragging and dropping the entire cocos2dx folder into an Xcode project (a project update technique).


minggo:
Thank you.

PR #1323, opened 2012-09-15
Added keyboard support

wdevore:
Added keyboard including EGL routing for OSX.

### DIFF
--- a/cocos2dx/CCDirector.cpp
+++ b/cocos2dx/CCDirector.cpp
@@ -323,7 +323,6 @@ void CCDirector::setOpenGLView(CCEGLView *pobOpenGLView)
             updateContentScaleFactor();
         }
 
-         m_pobOpenGLView->setTouchDelegate(m_pTouchDispatcher);
         m_pobOpenGLView->setTouchDelegate(m_pTouchDispatcher);
         m_pTouchDispatcher->setDispatchEvents(true);
 
--- a/cocos2dx/platform/CCEGLViewProtocol.h
+++ b/cocos2dx/platform/CCEGLViewProtocol.h
@@ -78,23 +78,26 @@ public:
     virtual CCPoint getVisibleOrigin() const;
     
     /** 
-     * Set the design resolution size.
-     * Behavior undefined when enableRetina == true.
+     * Set the design resolutin size.
+     * You can't use it with enableRetina together.
      * @param width Design resolution width.
      * @param height Design resolution height.
-     * @param resolutionPolicy The resolution policy desired, you may choose:
-     *                         [1] kCCResolutionExactFit Fill screen by stretch-to-fit: if the design resolution ratio of width to height is different from the screen resolution ratio, your game view will be stretched.
-     *                         [2] kCCResolutionNoBorder Full screen without black border: if the design resolution ratio of width to height is different from the screen resolution ratio, two areas of your game view will be cut.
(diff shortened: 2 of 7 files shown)


walzer:
Thanks wdevore, I think your code is the start point to add keyboard support for all desktop platforms.
But as the game rule of cocos2d-x, you should also add a test case in TestCpp to show how to use your new feature, will you?

wdevore:
I can add a test case to TestCpp. However, the unmodified test case is currently failing during the call:

CCScheduler::resumeTarget

on the line

HASH_FIND_INT(m_pHashForSelectors, &pTarget, pElement);

with a bad-access. It is a macro so I am unable to step-in and debug it. Oddly enough the proj.mac template works fine so I am trying to compare what is different with the test-case harness verse the templates example.

wdevore:
I tried studying the Hash macros but they make no sense to me.

Oddly enough I have only been using the published cocos2d-x sdk and never really tried running the test-cases. I am curious, who coded the OSX test harness and I assumed it worked for them (I am using OSX 10.7.4 at this time.)

If I could chat with them maybe they could help me.

Any thoughts?
Cheers.

wdevore:
Update,

I was able to run the 2.0.2 release TestCpp fine. Something must have changed between what is in GitHub and release 2.0.2. I will try and do a compare to see what the difference is. It could be either XCode setting or code changes.

Hmmm....
Cheers.

wdevore:
Update 2,

Comparing the 2.0.2 release with the Trunk I noticed that CCNode::onEnter does something completely different with the ScriptEngine. When I commented-out the Trunk script engine code the application would finally run without Hashing exception. Any thoughts?:

the 2.0.2 release:

```
if (m_nScriptHandler)
{
    CCScriptEngineManager::sharedManager()->getScriptEngine()->executeFunctionWithIntegerData(m_nScriptHandler, kCCNodeOnEnter);
}
```

the Trunk:

```
if (m_eScriptType != kScriptTypeNone)
{
    CCScriptEngineManager::sharedManager()->getScriptEngine()->executeNodeEvent(this, kCCNodeOnEnter);
}
```

cocos2dx/keyboard_dispatcher/CCKeyboardDispatcher.cpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+bool CCKeyboardDispatcher::init(void)
+{
+    m_bDispatchEvents = true;
+    m_pHandlers = CCArray::createWithCapacity(8);
+    m_pHandlers->retain();
+    m_pHandlersToAdd = CCArray::createWithCapacity(8);
+    m_pHandlersToAdd->retain();
+    m_pHandlersToRemove = ccCArrayNew(8);
```
minggo, line 65: Why not use CCArray like m_pHandlersToAdd?
I think it is better to use the same codes if there is not any especial reason.

wdevore, line 65: I agree, it was a copy-paste from the keypad code.

rtissera:
Should be merged with win32 keyboard support from this thread IMHO in order to have both OSX and Win32 implementations "on par" : http://www.cocos2d-x.org/boards/6/topics/20064

dumganhar:
Since we will rename gles20 branch to master. I will close it firstly.
I will send a new Pull Request which contains this changes. 
Thanks for your contribution.

PR #1331, opened 2012-09-17
Modify template by view controller

jzgenius:
第一次搞，错了别怪我哈，晕晕乎乎的～～


jzgenius:
[template] modify AppController, add viewController & window variant. all Xcode template modified.

minggo:
But no file changed.

PR #1333, opened 2012-09-18
[template] modify AppController, add viewController & window variant. al...

jzgenius:
...l Xcode template modified.


walzer:
Thanks for your contribution :)

PR #1336, opened 2012-09-18
Android Java+JNI refactoring.

nicolasgramlich:
Refactoring of the Android Java code into a library project.
Refactoring of the Android Java code to decoupled from Cocos2dxActivty (->Cocos2dxHelper) so that it can also be used in Fragments, but not only Activities.
Added a #IFDEF in ScriptingCore.cpp to allow loading of script files from the external assets directory instead of the apk-internal assets directory.
Refactoring of the Android JNI glue code into properly named files.

### DIFF
--- a/scripting/javascript/bindings/Android.mk
+++ b/scripting/javascript/bindings/Android.mk
@@ -9,6 +9,7 @@ LOCAL_MODULE_FILENAME := libscriptingcore-spidermonkey
 LOCAL_SRC_FILES := ScriptingCore.cpp \
                    cocos2d_specifics.cpp \
                    CCPhysicsSprite.cpp \
+                   cocos2dx.cpp \
                    js_manual_conversions.cpp \
                    js_bindings_chipmunk_manual.cpp \
                    js_bindings_chipmunk_functions.cpp


nicolasgramlich:
Had a quick discussion about this with @minggo .

(Not perfectly tested though!)

cocos2dx/platform/android/jni/Java_org_cocos2dx_lib_Cocos2dxHelper.cpp
```diff
@@ -0,0 +1,177 @@
+#include <stdlib.h>
+#include <jni.h>
+#include <android/log.h>
+#include <android/asset_manager.h>
+#include <android/asset_manager_jni.h>
```
minggo, line 5: <android/asset_manager.h> and <android/asset_manager_jni.h> exist since Android2.3, and cocos2d-x should support from Android2.2, so can not use these headers.

nicolasgramlich, line 5: I don't think I had added this, but moved it here from another file. This must have been removed from the cocos2d-x in the meantime?

minggo, line 5: I think it is because you didn't synchronize your codes before doing the modification.

minggo:
I fix some bugs based on this pull request.
And i pull a request: https://github.com/cocos2d/cocos2d-x/pull/1358.

Thank you for your work.
Now it makes java codes more clearly.

nicolasgramlich:
Awesome thanks! =)

PR #1288, opened 2012-09-05
CCscrollView Update

edbartley:
CCscrollView screenPos also needs to be adjusted by the scale in order to be used with Retina display. Otherwise the clipping region is not in the correct position.

PR #1337, opened 2012-09-18
MSVS and MAC project updates

edbartley:
Updated MSVS projects to use Project relative paths rather than Solution relative paths. See discussion with dumganhar in this closed pull request: https://github.com/cocos2d/cocos2d-x/pull/1288

Removed the optimization switch from the debug build in MAC project and added explicit header search paths.


minggo:
Thank you.

PR #1341, opened 2012-09-19
fix bug of not checking object before post notification

nichos1983:
I've fixed the bug. The bug description discussion is available here:
http://www.cocos2d-x.org/boards/6/topics/15853?r=15909
Please merge the fix, would you?


minggo:
Thank you.

PR #1342, opened 2012-09-19
[iOS]Add setMute and isMuted function to SimpleAudioEngine

jzgenius:
[iOS] Add setMute and isMuted function to SimpleAudioEngine.

### DIFF
--- a/CocosDenshion/include/SimpleAudioEngine.h
+++ b/CocosDenshion/include/SimpleAudioEngine.h
@@ -26,48 +26,20 @@ THE SOFTWARE.
 #ifndef _SIMPLE_AUDIO_ENGINE_H_
 #define _SIMPLE_AUDIO_ENGINE_H_
 
-#include <stddef.h>
 #include "Export.h"
-#include <typeinfo>
-#include <ctype.h>
-#include <string.h>
+#include <stddef.h>
 
 namespace CocosDenshion {
-
-class TypeInfo
-{
-public:
-    virtual long getClassTypeInfo() = 0;
-};
-
-static unsigned int getHashCodeByString(const char *key)
-{
-	unsigned int len = strlen(key);
-	const char *end=key+len;
-	unsigned int hash;
-
-	for (hash = 0; key < end; key++)
-	{
-		hash *= 16777619;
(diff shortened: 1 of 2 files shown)


CocosDenshion/include/SimpleAudioEngine.h
```diff
@@ -91,19 +63,13 @@ class EXPORT_DLL SimpleAudioEngine : public TypeInfo
     @param pszFilePath The path of the background music file,or the FileName of T_SoundResInfo
     @param bLoop Whether the background music loop or not
     */
-    void playBackgroundMusic(const char* pszFilePath, bool bLoop);
-    void playBackgroundMusic(const char* pszFilePath) {
-    	this->playBackgroundMusic(pszFilePath, false);
-    }
```
minggo, line 97: Can not remove this function, it is needed for js-binding.
Because js-binding can not generate binding codes for the function that has default value.

CocosDenshion/include/SimpleAudioEngine.h
```diff
@@ -91,19 +63,13 @@ class EXPORT_DLL SimpleAudioEngine : public TypeInfo
...
+    void playBackgroundMusic(const char* pszFilePath, bool bLoop = false);
 
     /**
     @brief Stop playing background music
     @param bReleaseData If release the background music data or not.As default value is false
     */
-    void stopBackgroundMusic(bool bReleaseData);
-    void stopBackgroundMusic() {
-    	this->stopBackgroundMusic(false);
-    }
```
minggo, line 106: Can not remove this function, it is needed for js-binding.
Because js-binding can not generate binding codes for the function that has default value.

minggo:
We should check out that the feature can easily been implemented in other platforms.

minggo:
Please don't remove other codes, because they are useful for js-binding.

Thank you.

walzer:
Hi, @jzgenius, I would close this pull request for the reasons below:
1. You haven't update the source code in `SimpleAudioEngine` that deal with js-binding. As the result, your pull request REMOVED js-binding sources.
2. There's no test case in TestCpp for `setMute` interface.
3. `setMute` was only added to ios port, which will lead to compilation error on android+windows+mac. At least you should add empty functions for other platforms.

PR #1346, opened 2012-09-19
HttpClient - writeData callback fix

sburavtsov:
As this callback can be called many times when we are getting larget GET response, we need to append data to response data buffer instead of clear it each time.

Here is fix for that.


sburavtsov:
too much commits, sorry

PR #1347, opened 2012-09-19
Fix for HttpClient writeData callback

sburavtsov:
writeData can be called several times, so we need to provide appending data instead of replacing it with latest call data.


walzer:
I've already fixed this bug 10 days ago, in this commit https://github.com/cocos2d/cocos2d-x/commit/5d6513316a48d6caa1f42e0e3d0a2178ea2e95f2
But thank you all the same :)

PR #1352, opened 2012-09-20
To be able to handle 302 redirects with curl we need to set option CURLOPT_FOLLOWLOCATION

sburavtsov:
This requires to use any service which redirects requests to own CDN urls, like wuala for example.


walzer:
Great! I haven't considered this condition. Thank you for your fix!

Issue #1356, opened 2012-09-21
ISO C++ violation


GavinBIC:
In file included from ../../extensions/CCBReader/CCNodeLoader.h:6:0,
                 from ../../extensions/CCBReader/CCBFileLoader.h:4,
                 from ../../extensions/CCBReader/CCBFileLoader.cpp:1:
../../extensions/CCBReader/CCBReader.h: At global scope:
../../extensions/CCBReader/CCBReader.h:238:30: error: ISO C++ forbids initialization of member ‘hasScriptingOwner’ [-fpermissive]
../../extensions/CCBReader/CCBReader.h:238:30: error: making ‘hasScriptingOwner’ static [-fpermissive]
../../extensions/CCBReader/CCBReader.h:238:30: error: ISO C++ forbids in-class initialization of non-const static member ‘hasScriptingOwner’
make: **\* [../../extensions/CCBReader/CCBFileLoader.o] Error 1

minggo:
I think i fixed this issue.
You can get latest codes and have a try.

Thank you.

GavinBIC:
great! thanks.

PR #1357, opened 2012-09-21
Linux makefile

GavinBIC:
linux build doesn't do both debug & release builds, it only do debug builds. modified some makefile and make-all script to do all builds.
- most recent snapshot 9b8941059ce2fb9d94b328479c4275e42d25a433  has some error while building, so branch is branched from few days back. however this branch won't have a error fix. it's not the purpose of this branch.
- I'll make another pull request that fixes build error that violates ISO C++. issue #1356


minggo:
Thank you.

PR #1359, opened 2012-09-21
Ptr int casting issue

GavinBIC:
casting pointer to int with C style casting causes problem in other platform. using stdint can solve this problem. this issue also breaks the linux build. I fixed it.

animation layer test's source code name changed, and this breaks TestCpp linux build that's fixed in this branch.

### DIFF
--- a/cocos2dx/platform/win32/CCStdC.h
+++ b/cocos2dx/platform/win32/CCStdC.h
@@ -50,6 +50,12 @@ THE SOFTWARE.
 #define MIN     min
 #define MAX     max
 
+#if _MSC_VER >= 1600
+    #include <stdint.h>
+#else
+    #include "./compat/stdint.h"
+#endif
+
 // Structure timeval has define in winsock.h, include windows.h for it.
 #include <Windows.h>
 
--- a/cocos2dx/platform/win32/compat/stdint.h
+++ b/cocos2dx/platform/win32/compat/stdint.h
@@ -1,0 +1,247 @@
+// ISO C9x  compliant stdint.h for Microsoft Visual Studio
+// Based on ISO/IEC 9899:TC2 Committee draft (May 6, 2005) WG14/N1124
+//
+//  Copyright (c) 2006-2008 Alexander Chemeris
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//   1. Redistributions of source code must retain the above copyright notice,
+//      this list of conditions and the following disclaimer.
+//
+//   2. Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//   3. The name of the author may be used to endorse or promote products
+//      derived from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR IMPLIED
+// WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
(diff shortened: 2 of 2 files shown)


dumganhar:
But there isn't a 'stdint.h' on WINDOWS platform. It will cause compile error.

GavinBIC:
I don't have full msvc build system, stdint exists in MSVS 2010. however that's the minimum version that supports stdint. there is a BSD-licensed compatible header at http://code.google.com/p/msinttypes/source/browse/trunk/stdint.h that I attached at the recent commit. please check if it works.

cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
```diff
@@ -385,7 +385,7 @@ CCSprite * CCTMXLayer::insertTileForGID(unsigned int gid, const CCPoint& pos)
     CCRect rect = m_pTileSet->rectForGID(gid);
     rect = CC_RECT_PIXELS_TO_POINTS(rect);
 
-    int z = (int)(pos.x + pos.y * m_tLayerSize.width);
+    intptr_t z = (intptr_t)(pos.x + pos.y * m_tLayerSize.width);
```
minggo, line 388: Why should change int to intptr_t here? It is not a pointer.

cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
```diff
@@ -447,7 +447,7 @@ CCSprite * CCTMXLayer::appendTileForGID(unsigned int gid, const CCPoint& pos)
...
```
minggo, line 450: Why should change int to intptr_t here? It is not a pointer.

cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
GavinBIC, line 388: Oops, didn't look carefully. z gets casted to (void _) at few lines down. ccCArrayInsertValueAtIndex(m_pAtlasIndexArray, (void_)z, indexForZ); 
I thought it was a offset of a pointer, but it looks like a value that gets inserted. anyway it get casted to a flexible-sized type(pointer).
so I thought setting intptr_t was OK 'cause intptr_t is not pointer, it's integer type that can hold pointer address. refering from www.cplusplus.com/reference/clibrary/cstdint/ . actually there's no data size conversion this way.

alternative way is doing below.
int z = ...
ccCArrayInsertValueAtIndex(m_pAtlasIndexArray, (void*)(intptr_t)z, indexForZ);
and there can be a conversion in x86_64 : int32 -> int64 .

we have to do this to get rid of warning that complains about cast between different sized types.
anyway, I think both ways aren't neat. Because 'storing integer inside pointer is a hack.
To become neat, using template container(stl or else) can be a good way to improve the neatness. but I think it's OK for not toucing the working legacy.

We can overload ccCarrayInsertValueAtIndex that hides the casting, or use template function with template specialization. but this commit will just fix the warning that'll keep the little hacking.

cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
GavinBIC, line 450: I wrote about it at above. let me know what you think.
I'll probably try other way that keeps z as int.

cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
minggo, line 388: Ok, i understand it now.
Could you add comment at this line, then we know why we use intptr_t here.
And i think we should not use stl or template if possible, because they are hard to debug, and hard to generating binding codes for lua or javascript.

PR #1360, opened 2012-09-22
Ne wplatform third party ios libraries libcurl

jzgenius:
New libcurl.a, armv7 & armv7s & simulator Supported.


minggo:
What's the version of the curl?
Shouldn't it change corresponding header files?
And could you please update other platforms' curl?

jzgenius:
1. It's curl-7.21.4.
2. include files no change.
3. only iOS

walzer:
Tested, this lib works properly on both ios6 simulator and ios 6 device.

PR #1361, opened 2012-09-22
add the capability to set custom http header to HttpClient

ncrothe:
I added the capability to set custom http headers to http wrapper. Among other things this allows to set content-type for POSTs to something else than the standard "application/x-www-form-urlencoded", e.g. to "application/json".


walzer:
Hi, you should add a test case in TestCpp/Classes/ExtensionsTest/NetworkTest/HttpClient.cpp, and teach how to use your interface in there.

ncrothe:
Hi, thanks for pointing that out. I added a test (to the POST tests), that should explain how to use it. Let me know if you need more.

walzer:
Merged. Thanks for your contribution.

PR #1372, opened 2012-09-24
Remove retina specific methods and logic

BorMor:
We remove iOS specific methods from cocos2d-x. Only setContentScaleFactor and setDesignResolution is enough for supporting all resolutions.

This changes not affects Tests and only change HelloWorld. As you can see it is much simpler now.

Our main idea is:
In initialization code user must set design resolution and content scale factor. For example, in modified HelloWorld we suggest that our design resolution is 480x320, so we set it in initialization code. But on devices with larger screen(if size.height > 320) we may use -hd resources to provide better picture. So we set folder to iphonehd and set contentscalefactor to 2. If you want add support for new ipad just add 4x textures and set contentscale factor to 4.

If you're accept this approach to multiresolution support we also suggest remove all confusing termins like WinSizeInPoints, VisibleSize. Only getWinSize should be used in logic.

### DIFF
--- a/cocos2dx/platform/ios/EAGLView.mm
+++ b/cocos2dx/platform/ios/EAGLView.mm
@@ -151,6 +151,11 @@ static EAGLView *view = 0;
         
         originalRect_ = self.frame;
         self.keyboardShowNotification = nil;
+		
+		if ([view respondsToSelector:@selector(setContentScaleFactor:)])
+		{
+			view.contentScaleFactor = [[UIScreen mainScreen] scale];
+		}
     }
         
     return self;
@@ -200,13 +205,13 @@ static EAGLView *view = 0;
 -(int) getWidth
 {
     CGSize bound = [self bounds].size;
-    return bound.width;
+    return bound.width * self.contentScaleFactor;
 }
 
 -(int) getHeight
 {
     CGSize bound = [self bounds].size;
-    return bound.height;
+    return bound.height * self.contentScaleFactor;
 }
 
 
@@ -401,8 +406,8 @@ static EAGLView *view = 0;
     int i = 0;
     for (UITouch *touch in touches) {
         ids[i] = (int)touch;
-        xs[i] = [touch locationInView: [touch view]].x;
-        ys[i] = [touch locationInView: [touch view]].y;
+        xs[i] = [touch locationInView: [touch view]].x * view.contentScaleFactor;;
+        ys[i] = [touch locationInView: [touch view]].y * view.contentScaleFactor;;
         ++i;
     }
     cocos2d::CCEGLView::sharedOpenGLView()->handleTouchesBegin(i, ids, xs, ys);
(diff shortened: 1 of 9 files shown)


moadib:
So let's discussion begins :) We believe that this is best approach for existing mess with different resolutions :)

minggo:
Hmm, it seems great.
I will checkout out it in detail.

cocos2dx/platform/ios/EAGLView.mm
```diff
@@ -151,6 +151,11 @@ - (id) initWithFrame:(CGRect)frame pixelFormat:(NSString*)format depthFormat:(GL
         
         originalRect_ = self.frame;
         self.keyboardShowNotification = nil;
+		
+		if ([view respondsToSelector:@selector(setContentScaleFactor:)])
+		{
+			view.contentScaleFactor = [[UIScreen mainScreen] scale];
```
minggo, line 157: The developer may don't want to scale content factor by default.

walzer:
@dualface

moadib:
@minggo 
We assume that user should specify design resolution and do not bother that retina enabled or not.

Don't want use -hd resources? No problems, just call `setDesignResolution(480, 320);` and don't set up content scale factor.

minggo:
This approach should be used in android too, so i think there should be two arguments when set content scale factor. One for x coordinate, one for y coordinate.

moadib:
This approach used by us on all platforms. We've tested it on ios, android, blackberry.

Different scale factors for x-axis and y-axis? Didn't think that this is good idea.

On devices with different aspect ratio, if you want use all screen - you just get appropriate design resolution.

For example, instead

`CCEGLView::sharedOpenGLView()->setDesignResolutionSize(480, 320, kResolutionShowAll);`

you can use this code

```
float sf = sz.width / sz.height;
if (sf >= 1.5f)
    CCEGLView::sharedOpenGLView()->setDesignResolutionSize(320 * sf, 320, kResolutionShowAll);
else
    CCEGLView::sharedOpenGLView()->setDesignResolutionSize(480, 480 / sf, kResolutionShowAll);
```

then your game will fill any screen, but you must be ensured that game logic assume that screen may be little wider than 480 or taller than 320

minggo:
I think we can not remove CCDirectory::enableRetinaDisplay() now, it will break compatibility. May be should make it deprecated, and delete it in future.

minggo:
@BorMor 
I agree with you.
We can remove post-fix such as "InPoint" and "InPixel".

minggo:
@BorMor 
I pulled a request based on this pull request: https://github.com/cocos2d/cocos2d-x/pull/1415.
I planned to pull a request to you, but can not find your repo when pulling a request.

That pull request fix conflict, and do some modification.
If you don't mind, i will close this pull request.

BorMor:
Yes, close this request :)

minggo:
@BorMor 
@moadib 
What are your email address? I want to discuss multi resolution with you.
Thank you.

moadib:
Feel free to write to ursinepaw [at] gmail :)

PR #1373, opened 2012-09-24
fixed cocos2dx linux TestCpp build break

GavinBIC:
linux TestCpp build breaks with fontconfig update. fontconfig is being used recently and library linkage doesn't exist at makefile. added a line to link the library.


minggo:
Thank you.

PR #1379, opened 2012-09-25
Fix cc notification observer perform selector

jzgenius:
1. not retain m_target and m_object in CCNotificationObserver function.
2. not perform m_object to performSelector target.

### DIFF
--- a/cocos2dx/support/CCNotificationCenter.cpp
+++ b/cocos2dx/support/CCNotificationCenter.cpp
@@ -165,8 +165,10 @@ CCNotificationObserver::CCNotificationObserver(CCObject *target,
                                                CCObject *obj)
 {
     m_target = target;
+	CC_SAFE_RETAIN(m_target);
     m_selector = selector;
     m_object = obj;
+	CC_SAFE_RETAIN(m_object);
     
     m_name = new char[strlen(name)+1];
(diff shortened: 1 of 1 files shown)


dumganhar:
Do we also need to modify the constructor of CCNotificationObserver ?

<pre>
CCNotificationObserver::CCNotificationObserver(CCObject *target, 
                                               SEL_CallFuncO selector,
                                               const char *name,
                                               CCObject *obj)
{
    m_target = target;
    CC_SAFE_RETAIN( m_target );   // Added
    m_selector = selector;
    m_object = obj;
    CC_SAFE_RETAIN( m_object );   // Added
    
    m_name = new char[strlen(name)+1];
    memset(m_name,0,strlen(name)+1);
    
    string orig (name);
    orig.copy(m_name,strlen(name),0);
}
</pre>

jzgenius:
Yes, your right!
Great!

dumganhar:
Thanks. :)

dumganhar:
The 'object' and 'target' parameters should not be retained. We must use weak reference, otherwise the life-cycle of target object will be delayed until the destruction of CCNotificationCenter.

dumganhar:
I have reverted this file.

PR #1385, opened 2012-09-25
fix font render blur, for ios/mac/windows

dualface:
### REASON

Render fonts, create texture width and height can not be divided by 2, then the display will be blurred (because the origin at the center of the image).

This patch to ensure that the width and height of the texture created font rendering is always a multiple of 2.

.

渲染字体时，如果创建的材质宽度和高度不能被 2 整除，那么显示时将会变得模糊（因为原点在图像的中央）。

这个补丁确保渲染字体时创建的材质宽度和高度总是 2 的倍数。


minggo:
I think CC_FIX_ARTIFACTS_BY_STRECHING_TEXEL is used for this purpose.

PR #1400, opened 2012-09-26
fix compilation error in xcodebuild command line mode, which are caused by ios6 compatibility

walzer:
- fix FontLabel, sync with github.com/zynga/FontLabel.  Please note that FontLabel has been removed in cocos2d-iphone
- fix HelloCpp, HelloLua, SimpleGame, TestCpp, TestLua, TestJavascript
- fix xcod4 templates.

### DIFF
--- a/samples/HelloCpp/proj.ios/RootViewController.mm
+++ b/samples/HelloCpp/proj.ios/RootViewController.mm
@@ -61,7 +61,7 @@
 - (NSUInteger) supportedInterfaceOrientations
 {
 #ifdef __IPHONE_6_0
-    return UIInterfaceOrientationMaskLandscape;
+    return UIInterfaceOrientationMaskAllButUpsideDown;
 #endif
 }
 
--- a/samples/HelloLua/proj.ios/RootViewController.mm
+++ b/samples/HelloLua/proj.ios/RootViewController.mm
@@ -60,7 +60,7 @@
 // For ios6, use supportedInterfaceOrientations & shouldAutorotate instead
 - (NSUInteger) supportedInterfaceOrientations{
 #ifdef __IPHONE_6_0
-    return UIInterfaceOrientationMaskLandscape;
+    return UIInterfaceOrientationMaskAllButUpsideDown;
 #endif
 }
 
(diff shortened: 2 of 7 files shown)


walzer:
ok, good idea. Writing UIInterfaceOrientationMaskAllButUpsideDown makes sense, because developers still need to set the orientation in target->Summary->Supported Interface Orientations

donaldalanmorrison:
@walzer Well, it was a half-complete idea. :(  It breaks the template.

If you also add to the plist file Initial interface orientation : Landscape Right or Left, this helps the template some, but the Portrait orientation is still broken in the template because the background image must be changed upon rotation.  But to do this would really require a transition between the images, so I'm not sure what the fix is yet.

It would really be a new feature: A layout transition between landscape and portrait, (rather than a scene transition).

walzer:
@donaldalanmorrison sorry I didn't catch your meaning. If we design a game from scratch, we should decided that it should be a landscape game or portrait, then artist will draw the whole game in either landscape or portrait. 

Do you mean that, you would create a game which fits both landscape and portrait? How will you transition game contents between landscape and portrait?

donaldalanmorrison:
@walzer That is the challenge.  For some games it could make sense, for others it would not.

For example, a game with a large map (larger than the display) which scrolls, but with small game characters, it could make sense.  Or for a puzzle game where the puzzle does not fit in the whole screen.

By contrast, it would not work for a puzzle game with a fixed board size. 

There are challenges for both the artist and the programmer.

donaldalanmorrison:
@walzer In the example of the template, the launch background image would be static as usual but have both Portrait and Landscape versions; but after launch, the template would switch to a more blank background and add sprites with layout as appropriate for the current orientation.

Note that the Apple GLKit template demonstrates rotation between portrait and landscape (with two cubes visible).  What it does not demonstrate is a menu ("overlay"); and how the aspect ratio change would change the art esthetic / layout.

donaldalanmorrison:
@walzer It's less of a challenge on the iPad because of the large screen size.  Also, the iPad is supposed to support all 4 orientations (not just 3 as for the iPhone/iPod Touch).

walzer:
OK. It's another topic, very interesting. Let's discuss it on the forum. We can have a try but it's not a high priority issue currently.

PR #1405, opened 2012-09-27
Complete delete CCDirector after called CCDirector::purgeDirector()

dualface:
For win/mac platform, change simulator view size whitout restart simulator.

改变模拟器分辨率，而不重新启动模拟器。可以显著提高开发效率，也方便 IDE 进行调试（重启会导致 IDE 调试结束）。

### DIFF
--- a/cocos2dx/CCDirector.cpp
+++ b/cocos2dx/CCDirector.cpp
@@ -73,21 +73,20 @@ NS_CC_BEGIN
 // XXX it should be a Director ivar. Move it there once support for multiple directors is added
 
 // singleton stuff
-static CCDisplayLinkDirector s_SharedDirector;
+static CCDisplayLinkDirector *s_SharedDirector = NULL;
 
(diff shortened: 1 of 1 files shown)


cocos2dx/CCDirector.cpp
```diff
@@ -73,21 +73,20 @@ of this software and associated documentation files (the "Software"), to deal
 // XXX it should be a Director ivar. Move it there once support for multiple directors is added
 
 // singleton stuff
-static CCDisplayLinkDirector s_SharedDirector;
+static CCDisplayLinkDirector *s_SharedDirector = NULL;
```
donaldalanmorrison, line 76: @minggo Declaring s_SharedDirector static here is file-scoped.  Don't you want to put it static in the header file's class definition instead?  Then it would be a Class Variable, not an instance variable, and not file-scoped.

The comment on line 73 refers to the case where it's not static, but an instance variable.

很抱歉，这是更容易观察。

PR #1408, opened 2012-09-27
Fix the compilation error in Linux (ubuntu 12.04)

swcai:
Fix the compilation error in Linux (ubuntu 12.04)

Signed-off-by: Wenwei Cai stanley.w.cai@gmail.com

### DIFF
--- a/cocos2dx/proj.linux/Makefile
+++ b/cocos2dx/proj.linux/Makefile
@@ -2,7 +2,7 @@ CC      = gcc
 CXX     = g++
 TARGET	= libcocos2d.so
 CCFLAGS = -Wall -fPIC
-CXXFLAGS = -Wall -fPIC
+CXXFLAGS = -Wall -fPIC -fpermissive
 VISIBILITY = 
 
 LIBS = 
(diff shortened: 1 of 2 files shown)


cocos2dx/proj.linux/Makefile
```diff
@@ -2,7 +2,7 @@ CC      = gcc
 CXX     = g++
 TARGET	= libcocos2d.so
 CCFLAGS = -Wall -fPIC
-CXXFLAGS = -Wall -fPIC
+CXXFLAGS = -Wall -fPIC -fpermissive
```
dumganhar, line 5: Why we should add '-fpermissive' parameter? It's for template inheritance. But cocos2dx haven't used it.

swcai, line 5: It is not for template inheritance, but downgrade some diagnostics about nonconformant code from errors to warnings. With the gcc in ubuntu 12.04, I got some errors as below:

../../extensions/CCBReader/CCBAnimationManager.cpp: In member function ‘void cocos2d::extension::CCBAnimationManager::runAnimations(int, float)’:
../../extensions/CCBReader/CCBAnimationManager.cpp:473:53: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]

dumganhar:
Have you tried the latest codes? We have modified the return value of 'CCDictionary::getIntKey' from 'int' to 'intptr_r'.
I think it's resolved.

GavinBIC:
On Fri 28 Sep 2012 04:56:51 PM KST, Wenwei Cai wrote:

> I found the bug at a3ab303
> https://github.com/cocos2d/cocos2d-x/commit/a3ab30324cde92e9115af67c1180609b465db9ea
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/cocos2d/cocos2d-x/pull/1408#issuecomment-8967219.

what is it?

cocos2dx/proj.linux/Makefile
GavinBIC, line 5: Why ignore nonconformant code? It can possibly cause a latent bug.

swcai, line 5: Since you fixed it in CCDictionary, please ignore the patch. Thanks.

PR #1415, opened 2012-09-29
Multi resolution

minggo:
Fix conflict of the pull request: https://github.com/cocos2d/cocos2d-x/pull/1372.
Remove CCDirector::m_obWinSizeInPixel and rename m_obWinSizeInPoint to m_obWinSize.


minggo:
@BorMor 
I found it is hard to remove xxxInPixel, because many functions use it.

PR #1416, opened 2012-09-29
Gles20

trentmillar:
-Adding solid circle to CCDrawaingPrimitives   ac90a0f
I simply added the Solid Circle. This is something in Cocos2d but not -x

 -Fixing merge issues   41881db
Looks like the visual studio 2012 template sln file has duplicated content.


minggo:
This pull request can not be automatically merged.
Could you please resolve the conflict?

Thank you.

PR #1417, opened 2012-09-30
add setMute and isMuted function to SimpleAudioEngine. iOS & Mac is comp...

jzgenius:
add setMute and isMuted function to SimpleAudioEngine. iOS & Mac is completed, other platform need to coding.

### DIFF
--- a/CocosDenshion/android/SimpleAudioEngine.cpp
+++ b/CocosDenshion/android/SimpleAudioEngine.cpp
@@ -379,4 +379,15 @@ void SimpleAudioEngine::stopAllEffects()
 	}
 }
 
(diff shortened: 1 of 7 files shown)


minggo:
I find there is an issue about this implementation. I add a menu item in CocosDenshionTest to invoke setMute().
1. run CocosDenshionTest in TestCpp
2. play background music
3. set mute(work as expected, become muted)
4. click HOME button
5. play another app using cocos2d-x and it should play background music
6. click HOME button
7. run TestCpp again
8. click "play background music" again (can hear the background music)
9. set mute (can not set mute, also hear the background music)

PR #1418, opened 2012-09-30
Add Simple removeObserver function into CCNotificationCenter.

jzgenius:
Add Simple removeObserver function into CCNotificationCenter.

### DIFF
--- a/cocos2dx/support/CCNotificationCenter.cpp
+++ b/cocos2dx/support/CCNotificationCenter.cpp
@@ -114,6 +114,25 @@ void CCNotificationCenter::removeObserver(CCObject *target,const char *name)
     }
 }
 
+void CCNotificationCenter::removeObserver(CCObject *target)
+{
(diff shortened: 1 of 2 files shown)


walzer:
I inclined to reject this commit. The reason is that:  a target probably observes many events with different names. e.g.

``` cpp
CCNotificationCenter::sharedNotificationCenter()->addObserver(pMonster, callfuncO_selector(Monster::attacked), "monster-attacked, NULL);
CCNotificationCenter::sharedNotificationCenter()->addObserver(pMonster, callfuncO_selector(Monster::dead), "monster-dead, NULL);
```

This usage could be very very common. Please read the code of `CCNotificationObserver`, an observer != a target.

But in your code, you remove only one observer, even don't know which one is removed, then return.
If you're looking for `revmoeTarget`, please loop the whole `m_observers` and remove all _observers_ of your assigned _target_.

PR #1419, opened 2012-09-30
Run Android debugger with Cocos2D-X projects!

folecr:
Please merge! This is incredibly useful for Android development.
- All created android projects now have an auto-generated ndkgdb.sh that will allow full gdb debugging!
- As a pre-generated example, HelloCpp sample can now be debugged by running ndkgdb.sh

Notes:
- The gdb server on the device/emulator takes a few millis to startup... which means that code run at app startup/bootstrap time cannot yet be debugged
- I will fix this next

### DIFF
--- a/template/android/copy_files.sh
+++ b/template/android/copy_files.sh
@@ -52,6 +52,12 @@ copy_build_native(){
     chmod u+x $APP_DIR/proj.android/build_native.sh
 }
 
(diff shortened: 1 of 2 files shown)


walzer:
Hi, we're in a Nation Days' holiday until next Monday. Would I review the code later?

folecr:
OK!

PR #842, opened 2012-04-15
Gles20

jbyu:
add project for ios
fixed memory leak in CCGLProgram
add pixel format I8 in CCTexture2D
update Director.pkg for LuaCocos2d

PR #1420, opened 2012-10-01
Fix removeTileAt crash corresponding this post: http://www.cocos2d-x.org...

sburavtsov:
.../boards/6/topics/4336

### DIFF
--- a/cocos2dx/sprite_nodes/CCSpriteBatchNode.cpp
+++ b/cocos2dx/sprite_nodes/CCSpriteBatchNode.cpp
@@ -641,9 +641,9 @@ void CCSpriteBatchNode::removeSpriteFromAtlas(CCSprite *pobSprite)
         m_pobDescendants->removeObjectAtIndex(uIndex);
 
         // update all sprites beyond this one
-        unsigned int count = m_pobDescendants->count();
+        unsigned int count = uIndex; // we will update items from the beginning to uIndex as m_pobDescendants is in ascending order
 
-        for(; uIndex < count; ++uIndex)
+        for(uIndex = 0; uIndex < count; ++uIndex)
         {
             CCSprite* s = (CCSprite*)(m_pobDescendants->objectAtIndex(uIndex));
             s->setAtlasIndex( s->getAtlasIndex() - 1 );


sburavtsov:
Btw, it fixes crashes for me, which 100% repeatable if I remove tile in action complete callback.
With this fix it works np.

minggo:
Could you please provide a test case?

sburavtsov:
Hi minggo, actually it seems there is deeper issue than just this change, I need investigate more. This change removes immediate crash and I can send test case, but let me first to find real crash issue and fix it.

minggo:
Hi @sburavtsov 
Have you found the real crash?

sburavtsov:
Well, it stops 100% repeatable crashes, but I not sure if it adds one more - it happens after some actions and long period of inactivity. Crash log always the same and I'd like to talk with someone of core developers as I not sure I know enough to figure out real reason.
I will update this post at this week with details, or you can contact me via email: serobur 'at' gmail.com

sburavtsov:
Hi minggo, sorry I was thinking this crash is related to this change/fix. It is not, so you can use it.
Real reason of crash is in HttpClient.ccp, I'm looking now how to fix it.

minggo:
What should i do with this PR?
Close it?

sburavtsov:
Sorry for some mess with this issue.
This PR fix immediate crash if you are doing removeTile while tile sprite is playing action, so it seems it is right, and also it seems to be right after some looking in code. But I just found one more crash and it seems to be connected with that - game crashes here:

``` C++
#if CC_REBIND_INDICES_BUFFER
    glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, m_pBuffersVBO[1]);
#endif

#if CC_TEXTURE_ATLAS_USE_TRIANGLE_STRIP
    glDrawElements(GL_TRIANGLE_STRIP, (GLsizei) n*6, GL_UNSIGNED_SHORT, (GLvoid*) (start*6*sizeof(m_pIndices[0])) );
#else
-> here    glDrawElements(GL_TRIANGLES, (GLsizei) n*6, GL_UNSIGNED_SHORT, (GLvoid*) (start*6*sizeof(m_pIndices[0])) );
#endif // CC_TEXTURE_ATLAS_USE_TRIANGLE_STRIP

#if CC_REBIND_INDICES_BUFFER
    glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, 0);
#endif
```

after some minutes of app inactivity (this in on simulator).
And stack trace is

``` log
#8  0x000ee993 in cocos2d::CCTextureAtlas::drawNumberOfQuads(unsigned int, unsigned int) at .../libs/cocos2dx/textures/CCTextureAtlas.cpp:624
#9  0x000ee84e in cocos2d::CCTextureAtlas::drawQuads() at .../libs/cocos2dx/textures/CCTextureAtlas.cpp:583
#10 0x000d10c1 in cocos2d::CCSpriteBatchNode::draw() at .../libs/cocos2dx/sprite_nodes/CCSpriteBatchNode.cpp:407
#11 0x000cfbc5 in cocos2d::CCSpriteBatchNode::visit() at .../libs/cocos2dx/sprite_nodes/CCSpriteBatchNode.cpp:160
#12 0x00037373 in cocos2d::CCNode::visit() at .../libs/cocos2dx/base_nodes/CCNode.cpp:735
#13 0x00037373 in cocos2d::CCNode::visit() at .../libs/cocos2dx/base_nodes/CCNode.cpp:735
#14 0x00037373 in cocos2d::CCNode::visit() at .../libs/cocos2dx/base_nodes/CCNode.cpp:735
#15 0x0003b0ec in cocos2d::CCDirector::drawScene() at .../libs/cocos2dx/CCDirector.cpp:229
```

if you have any ideas it can help a lot as it seems like batch node is corrupted, but I don't have clear idea on how to find/fix that, as I don't know much on SpriteBatchNode internals.

it seems I need to look more there to get real fix for this or it is another, not connected issue.

minggo:
Did you mean that even you use this PR, it still crashes?

sburavtsov:
This PR fixes one known issue with immediate crash in specific use case. I can provide test for that I think and this case described on forum (by link in first message).

But there is another issue I posted above. And I don't know if it is connected with this PR or not. I is hard to reproduce it, but it exists. As it is in SpriteBatchNode  I think they can be connected. That is why I think I need to research more on that.
May be best solution will be to make special test project for TMXTileMap and try to find real reasons.

sburavtsov:
Update:
- there is no more known issues for me in CCSpriteBatchNode
- crash described above happened one more time in iPhone Simulator, after about 20 hours app run.
- it seems it happens just when it runs on iPhone Simulator and I working with XCode at the same time
- it seems to be very close to this issue/fix:
  https://github.com/cocos2d/cocos2d-x/pull/842/files - CCPlatformMacros.h and seems to be platform-dependent, it crashes on iMac now
  I will keep it running on iPhone for today to see what will happen.
- also I found very similar crash question here:
  http://stackoverflow.com/questions/12846659/weird-seemingly-random-crashes-when-calling-gldrawelements-in-ios-simulator

On pull request with SpriteBatchNode fix - it can be merged to gles20 as it works fine.

minggo:
Ok.
Thank you.

minggo:
After digging into the codes, i think we can not fix the issue by this modification.
As you commented in codes, uIndex as m_pobDescendants is in ascending order, so we should minus uIndex after the specified sprite. 

Assume the uIndex are:
1, 2, 3, 4, 5
If we remove a sprite that which uIndex is 3 without changing other sprites' uIndex, it looks like
1, 2, 4, 5
If we use the codes in this PR, the result will be
0, 1, 4, 5
It is wrong.

And this will cause crash by running PerfermanceTest/PerformanceNodeChildrenTest.

PR #1421, opened 2012-10-01
fixed scrolling bug when scroll view is not at CCPointZero position in w...

chuckzhang:
if ccscrollview or cctableview is not at CCPointZero postion in world, the scrolling behaviour is not correct.


dumganhar:
Hi, chuckzhang
I have already fixed some bugs of CCScrollView at PR #1428, it contains the change of this PR.
So, do you think we can close this Pull Request?

chuckzhang:
cooool

PR #1426, opened 2012-10-07
Android - Cocos2dxActivity update

edbartley:
Allow Cocos2dxActivity subclasses to override the View creation process for developers that use a custom Layout and specialized views.


minggo:
Thank you.

PR #1439, opened 2012-10-10
Updated script behaviour

folecr:
- Auto generated bindings are always pushed to remote repo.
- If a remote repository is not given, exit with failure.
- Updating the submodule reference in cocos2d-x is not done unless
- ... remote cocos2dx repo is set.


dumganhar:
Thanks.

PR #1443, opened 2012-10-12
Run cocos2d-html5-tests on cocos2d-x

folecr:
- Remove copied tests, use https://github.com/cocos2d/cocos2d-html5-tests as a submodule
- Android packaging scripts so that the APK contains assets in the same location as iOS
- ndkgdb.sh so you can debug on Android
- Android project that can build and run these tests


minggo:
Thank you.

PR #1447, opened 2012-10-12
use SoundPool.OnLoadCompleteListener to play effect if it is not preloaded

minggo:
...s not preloaded


minggo:
1. use SoundPool.OnLoadCompleteListener to play effect if it is not preloaded
2. re-init data when Cocos2dxSound.end() is called
3. resume all effects manually, because SoundPool.autoResume() only resume effects that are paused by SoundPool.autoPause()

PR #1451, opened 2012-10-13
potential container modification while it's been traversing in CCNotificationCenter

micahli:
fix a bug with CCNotificationCenter that observer->performSelector(observer) may call CCNotificationCenter::removeObserver and cause CCARRAY_FOREACH(m_observers, obj) access invalid address


dumganhar:
Thanks, it was merged.

PR #1452, opened 2012-10-14
Fix getLocationInView () method's comment

shinriyo:
getLocationInView () method's comment is mistake.
I get the comment from the previousLocationInView () method.


dumganhar:
Hey, I think the comment of 'previousLocationInView' is wrong rather than 'getPreviousLocationInView'.

shinriyo:
Yes. I mistook so I sent other pull request

PR #1464, opened 2012-10-16
Added TTF support to Blackberry

ImperialPenguin:
This should fix all of the issues that I said I'd fix many months ago and just got around to. If there are any problems let me know. I'm still struggling with Git, so hopefully I did everything correctly.


ImperialPenguin:
The latest changes to CCImage should fix all of the font issues. They work fine for me.

dumganhar:
Thanks, it has been merged. But there are still some wrong display. Could you compare 'FontTest' on blackberry to win32 or mac? You will find that.

PR #1465, opened 2012-10-17
issue #1511: fixes as follows

dumganhar:
[1] OpenGL framebuffer access violation fix
[2] Fixing include paths to be able to compile cocos2d-x using external Visual Studio solution
[3] Application built in release mode will get writable directory in system disk.
[4] Multi-Touch support for win7/8 tablet or ultrabook.


ivfit:
Great, thanks :)

Why you've added #if(WINVER >= 0x0601) for Touch support?

I've created late binding so it can be compiled on Windows XP using #define WINVER _WIN32_WINNT_WINXP and the same app would work on XP and have Touch support if exists on Win7/8.

dumganhar:
Because vs2008's WINVER is 0x600, some compilation errors will appear in vs2008 if we don't check this.

ivfit:
That is probably due to old Microsoft SDK (winuser.h) file being used.

Having #ifdef WINVER >= _WIN32_WINNT_WIN7 (0x0601) makes late binding completely unnecessary.

Maybe we should detect VS2008 if it makes problems during compiling (_MSC_VER)?

Does your VS2010 need #if(WINVER >= 0x0601)? I've checked VS2010 on XP, Win7 and Win8 and it works without it.

dumganhar:
The Macro WINVER in VS2010 is equal to 0x0601. It works.
Yeah, is checking this macro can't work? Or are there any better ways to achieve that?
What's your opinion?

ivfit:
If I understand correctly, if you remove #if(WINVER >= 0x0601) from cocos2dx/platform/win32/CCEGLView.cpp:

1) VS2008 has some compilation errors
2) What about VS2010?

dumganhar:
Yes,  these compilation errors only appear on VS2008 since there isn't a define of HTOUCHINPUT in winuser.h.
VS2010 is ok.

ivfit:
So it is better that inside cocos2dx/platform/win32/CCEGLView.cpp instead of:

#if(WINVER >= 0x0601)

we put:

#if(_MSC_VER >= 1600) // Visual Studio 2010 or better

dumganhar:
Yeah, I agree. Thanks.

dumganhar:
Hey, johnangel
I have changed '#if(WINVER >= 0x0601)' to '#if(_MSC_VER >= 1600)'. Thank you.

ivfit:
Great, many thanks James!

PR #1481, opened 2012-10-19
Marmalade Pull Requests

mrmop:
Hi, 

I have added two commits on behalf of Marmalade. 

Commit 1 - Marmalade Support - This basically adds proj.marmalade and Marmalade specific files and folders  as well as implementation of numerous examples and the Cocos2d-x test bed. These changes do not affect the main Cocos2d-x code base

Commit 2 - Marmalade Cocos2d-x Main Repo Changes - These are direct changes to the main Cocos2d-x code base, basically a couple of improvements and a couple of critical changes that would need to be included into the main Cocos2d-x code base for Marmalade support to work without compiler errors. The critical changes include:

cocos2dx/include/cocos2d.h - We have added  header includes that  pull in various Marmalade Cocos2d-x platform specific headers (in the same way that the other platforms do)

cocos2dx/kazmath/include/kazmath/neon_matrix_impl.h - We have added an additional check for I3D_ARCH_ARM to enable compile on x86 and ARM builds under Marmalade.

Looking forward to your response

Thanks

Mat


dumganhar:
Hi, Mat Hopwood
Thanks for your contribution.
Why did you copy all TestCpp resource to sample/TestCpp/proj.marmalade/data folder?
I think these resource files should not be in git index. Otherwise our repo will contain lots of duplicated resources.
Could it be copied by something similar with 'Pre Build Event' on VS?

walzer:
1. pre-build event
2. alias
3. tweak a bit in makefile

Any is OK.

mrmop:
HI James, 

Its a standard folder structure that Marmalade uses. I can take a look to see if I can redirect the Marmalade data folder to point at the Resources folder

dumganhar:
Why not also remove the dependence of resources in  'HelloCpp, HelloLua, SampleGame'/proj.marmalade/data ?

dumganhar:
In tolua_is.c:

<pre>
/* the equivalent of lua_is* for usertype */
static int lua_isusertype (lua_State* L, int lo, const char* type)
{
    if (!lua_isuserdata(L,lo)) {
        if (!push_table_instance(L, lo)) {
            return 0;
        };
    };
    {
        /* check if it is of the same type */
        // MARMALADE CHANGE: WE MUST DISABLE THIS, AS IT DOES NOT PERMIT HIERARCHIES MORE THAN 1 DEEP!
        return 1;
/*        int r;
        const char *tn;
        if (lua_getmetatable(L,lo))        // if metatable?
        {
            lua_rawget(L,LUA_REGISTRYINDEX);  // get registry[mt]
            tn = lua_tostring(L,-1);
            r = tn && (strcmp(tn,type) == 0);
            lua_pop(L, 1);
            if (r)
                return 1;
            else
            {
                // check if it is a specialized class
                lua_pushstring(L,"tolua_super");
                lua_rawget(L,LUA_REGISTRYINDEX); // get super
                lua_getmetatable(L,lo);
                lua_rawget(L,-2);                // get super[mt]
                if (lua_istable(L,-1))
                {
                    int b;
                    lua_pushstring(L,type);
                    lua_rawget(L,-2);                // get super[mt][type]
                    b = lua_toboolean(L,-1);
                    lua_pop(L,3);
                    if (b)
                        return 1;
                }
            }
        }*/
    }
    return 0;
}
</pre>

Why return 1 directly? Does it affect other platforms?
I think we shouldn't modify the sources of 'tolua++' if it's not an error.

dumganhar:
Hey Mat,
I have made some changes of Marmalade port based on this Pull Request.
But there is a bug of github that I can't send a Pull Request to your repo.
So could you merge these changes to your branch?
This is the URL : https://github.com/dumganhar/cocos2d-x/tree/mrmop-marmalade

mrmop:
I will sort that later today

mrmop:
Hi James, 

I've merged your changes back into mine and added a new update. I have removed the need to host resources in the Marmalade project data folders with exception to fonts because they are not present in the current crop of Cocos2dx samples. I have had to add those into the resources folders to enabled the Marmalade versions of the samples to run.properly. The files are very small however.

Note that there is also a change to CCNode.cpp/.h. A new method called updateTransform() has been added.

Also, scripting/lua/tolua/tolua_is.c has been reverted back to its original state before my first commit, Marmalade have found another way to implement the functionality it needs without affecting tolua_is.c

And finally, a quick note regarding CCSprite:

We believe that the CCSprite::updateTransform() function needs to inherit from a base function on the CCNode.
Some of the updateTransform() functionality is really part of CCNode,and should not be considered specific to CCSprite and its derivatives.

For example, CCSprite.cpp, line 556:

```
    arrayMakeObjectsPerformSelector(m_pChildren, updateTransform, CCSprite*);
```

...is broken. It seems to assume that all children of a CCSprite are of type CCSprite (or a derivative). But what if they are a different derivative of CCNode? This function crashes.

The line should simply pass in CCNode\* instead, and ensure we have CCNode::updateTransform() defined:

```
    arrayMakeObjectsPerformSelector(m_pChildren, updateTransform, CCNode*);
```

All being well with the latest commit, do you think that it would be possible to include Marmalades changes in the next release that's due to go out?

Thanks

Mat

dumganhar:
Hey Mat,
Why is a derivative of CCNode added to CCSpriteBatchNode.
You should look into 'CCSpriteBatchNode::addChild' function. It will make an assert if the child type isn't 'CCSprite' or the derivative of CCSprite.

<pre>
void CCSpriteBatchNode::addChild(CCNode *child, int zOrder, int tag)
{
    CCAssert(child != NULL, "child should not be null");
    CCAssert(dynamic_cast<CCSprite*>(child) != NULL, "CCSpriteBatchNode only supports CCSprites as children");
    CCSprite *pSprite = (CCSprite*)(child);
    // check CCSprite is using the same texture id
    CCAssert(pSprite->getTexture()->getName() == m_pobTextureAtlas->getTexture()->getName(), "CCSprite is not using the same texture id");
    CCNode::addChild(child, zOrder, tag);
    appendChild(pSprite);
}
</pre>

-- All being well with the latest commit, do you think that it would be possible to include Marmalades changes in the next release that's due to go out?

If there are no many issues, we will merge your changes and release it in the next version.
But we should test it first. Could you also help us test marmalade port on all supported platform?

dumganhar:
Hey Mat,
I found a bug of the implementation of 'glViewport' on Marmalade port.

<pre> void  glViewport (GLint x, GLint y, GLsizei width, GLsizei height);</pre>

That is if x < 0, it will take no effects.Also if y < 0, the display is incorrect.
Just 'HelloCpp' can reproduce this issue.
You will find out what the difference is if you run HelloCpp and compare marmalade port with win32 port.

I think this is a fatal bug need to be resolved before being merged.
- James

mrmop:
Hi James, 

The issue we've identified is about adding non-sprite CCNode objects as children of CCSprite objects, not as children of CCSpriteBatchNode objects.

You can see from the line of code we identified in CCSprite::updateTransform() that it expects all children to be of type CCSprite. Surely this is a bug? Why would you not be able to add a different kind of CCNode as a child of CCSprite? Our proposal to make updateTransform() a virtual method of CCNode, and change the line in CCSprite::updateTransform() to case instead to CCNode, would solve the problem. 

Thanks

Mat

mrmop:
HI James, 

Marmalade is not doing anything to modify the behaviour of glViewport, we simply pass through to the underlying driver.
According to the GL spec, "Viewport width and height are silently clamped to a range that depends on the implementation.".
So actually the behaviour of this function is undefined in the case that x <0 or y < 0.

Marmalade supports the use of 3 different OpenGL ES 2.0 drivers on desktop (Windows), and they behave as follows:
- ANGLE driver (default): Does not permit x<0 or y<0. What happens is that x and y are clamped to 0, and the height (but not the width!) reduce accordingly.
- Imagination driver: Seems to permit x<0 and y<0 and behaves "fine"
- Qualcomm driver: Seems to permit x<0 and y<0 and behaves "fine"

The OpenGL ES driver can be changed using the Marmalade Simulator "Configuration > GL" dialog.
So, in the event that a developer really wants to see behaviour for x<0 or y<0, then can use these drivers on Windows.
The behaviour on device will depend on the behaviour of the device driver, in exactly the same way as if developers are using Cocos2d-x "directly" for that platform.

Thanks

Mat

mrmop:
HI James, 

Whilst testing TestCPP I received an assertion while running the particles test at line 335 of CCParticleSystem.cpp:

```
                    bool isOK = image->initWithImageData(deflated, deflatedLen);
                    CCAssert(isOK, "CCParticleSystem: error init image with Data");
```

The format from the plist file is not extracted and passed as a 3rd parameter to initWithImageData()

Is this Cocos2d-x issue or is the platform specific implementation of CCImage::initWithImageData() meant to figure out the format when an invalid format is the eFmt parameter passed?

Thanks

Mat

dumganhar:
The data format in some of plist files is TIFF. So you should add TITF library to 'cocos2dx/platform/third_party'.
For more details, you should refer to 'cocos2dx/platform/CCImageCommon_cpp.h'.

bool CCImage::initWithImageData(void \* pData, 
                                int nDataLen, 
                                EImageFormat eFmt/\* = eSrcFmtPng_/, 
                                int nWidth/_ = 0_/,
                                int nHeight/_ = 0_/,
                                int nBitsPerComponent/_ = 8*/)
{
    bool bRet = false;
    do 
    {
        CC_BREAK_IF(! pData || nDataLen <= 0);

```
    if (kFmtPng == eFmt)
    {
        bRet = _initWithPngData(pData, nDataLen);
        break;
    }
    else if (kFmtJpg == eFmt)
    {
        bRet = _initWithJpgData(pData, nDataLen);
        break;
    }
    else if (kFmtTiff == eFmt)
    {
        bRet = _initWithTiffData(pData, nDataLen);
        break;
    }
    else if (kFmtRawData == eFmt)
    {
        bRet = _initWithRawData(pData, nDataLen, nWidth, nHeight, nBitsPerComponent);
        break;
    }
    else
    {
        // if it is a png file buffer.
        if (nDataLen > 8)
        {
            unsigned char* pHead = (unsigned char*)pData;
            if (   pHead[0] == 0x89
                && pHead[1] == 0x50
                && pHead[2] == 0x4E
                && pHead[3] == 0x47
                && pHead[4] == 0x0D
                && pHead[5] == 0x0A
                && pHead[6] == 0x1A
                && pHead[7] == 0x0A)
            {
                bRet = _initWithPngData(pData, nDataLen);
                break;
            }
        }

        // if it is a tiff file buffer.
        if (nDataLen > 2)
        {
            unsigned char* pHead = (unsigned char*)pData;
            if (  (pHead[0] == 0x49 && pHead[1] == 0x49)
                || (pHead[0] == 0x4d && pHead[1] == 0x4d)
                )
            {
                bRet = _initWithTiffData(pData, nDataLen);
                break;
            }
        }

        // if it is a jpeg file buffer.
        if (nDataLen > 2)
        {
            unsigned char* pHead = (unsigned char*)pData;
            if (   pHead[0] == 0xff
                && pHead[1] == 0xd8)
            {
                bRet = _initWithJpgData(pData, nDataLen);
                break;
            }
        }
    }
} while (0);
return bRet;
```

}

dumganhar:
I will try to switch driver to 'Imaginatio' or 'Qualcomm'. Thanks.

dumganhar:
All right, I think you are right. I will consider to merge these changes. Thanks. :)

<pre>
The issue we've identified is about adding non-sprite CCNode objects as children of CCSprite objects, not as children of CCSpriteBatchNode objects.
You can see from the line of code we identified in CCSprite::updateTransform() that it expects all children to be of type CCSprite. Surely this is a bug? Why would you not be able to add a different kind of CCNode as a child of CCSprite? Our proposal to make updateTransform() a virtual method of CCNode, and change the line in CCSprite::updateTransform() to case instead to CCNode, would solve the problem.
</pre>

mrmop:
Hi James, 

We've found an issue with Box2D assertions on ARM builds. There is a block of assertions at the top of b2Body constructor
(b2Body.cpp, line 25):

b2Assert(bd->position.IsValid());
b2Assert(bd->linearVelocity.IsValid());
b2Assert(b2IsValid(bd->angle));
b2Assert(b2IsValid(bd->angularVelocity));
b2Assert(b2IsValid(bd->angularDamping) && bd->angularDamping >= 0.0f);
b2Assert(b2IsValid(bd->linearDamping) && bd->linearDamping >= 0.0f);

As this post says:
http://www.box2d.org/forum/viewtopic.php?f=3&t=7559

...some of these assertions can fire falsely in ARM builds. "The problem lies with the arm implementation of std::numeric_limits<float32>::infinity(). This returns 0 on ARM builds and causes b2IsValid to always return false (0 < x && x < 0)"

Would it be possible to wrap these within an #ifndef ARM or similar construct?

Thanks

Mat

dumganhar:
Hey Mat,
I don't think it's a good way to modify the source code of BOX2D.
Does Marmalade use its own arm compiler for all supported platforms?
If so, why do other arm compilers(ios, android, blackberry) work fine?
What's the return value of **numeric_limits<int>::has_infinity**?
Maybe the implementation of stl in Marmalade is different? 
- James

mrmop:
HI James, 

Can I ask why the Marmalade pull request didn't make 2.0.4?

Thanks

Mat

dumganhar:
Hi Mat,
Sorry for that we didn't merge Marmalade port to v2.0.4. There are some reasons as follows:
- [1] Marmalade port doesn't implement our multi-resolution feature. The **HelloCpp** example show how to use this feature. 
  So something else need to be done, it's that you should update the CCFileUtils implementation of Marmalade port.
  For details, you should look inside **cocos2dx/platform/win32(or blackberry,mac, etc)/CCFileUtils.cpp** .
  We have supported one design resolution to match multiply resource. Therefore we have added **CCFileUtils::setResourceDirectory** to select which resource to load. 
  For example, if we have invoked **CCFileUtils::sharedFileUtils()->setResourceDirectory("hd")** in **AppDelegate.cpp**, 
  When **CCSprite::create("sprite.png")** is called, we search sprite.png in the **RootResourcePath/hd/sprite.png**,
  If **RootResourcePath/hd/sprite.png** doesn't exist, we will go back to **RootResourcePath/sprite.png** to find it.
  So Marmalade port should implement this logic of file searching.
  Besides, in HelloCpp and TestCpp  project, you should add **HelloCpp/Resources** as the root resource folder, since we have used resource directory to select resource rather than suffix.
- [2] LibTiff needs to be added to Marmalade port. Otherwise, ParticleTest will crash.
- [3] Could Marmalade port use the same implementation of CCImage which in CCImageCommon.cpp.h?

I hope these could be done before next version(v2.0.5). Expecting the back of Marmalade port.

Let me know what you think?

Thanks again for your contribution. :)

walzer:
Hey, guys, don't let this issue sink.

dumganhar:
Hi @mrmop,
I'm sorry for merging this PR a little late. 
Thanks for your contribution.
BTW, could you help us to resolve the remaining issues of Marmalade port?

PR #1482, opened 2012-10-19
fixes ScriptingCore for FF 16 API

rolandoam:
Ok, this should work now on FF 16... will leave this as a PR until someone can give it a review (just to make sure it's not breaking some other platform :panda_face:)


dumganhar:
Thank you. :)

PR #1491, opened 2012-10-23
Fix for correct unscheduling and instance destruction

sburavtsov:


extensions/network/HttpClient.cpp
```diff
@@ -354,9 +354,9 @@ CCHttpClient* CCHttpClient::getInstance()
 
 void CCHttpClient::destroyInstance()
 {
-    CCDirector::sharedDirector()->getScheduler()->unscheduleSelector(schedule_selector(CCHttpClient::dispatchResponseCallbacks), 
-                                                                     CCHttpClient::getInstance());
-    CC_SAFE_RELEASE_NULL(s_pHttpClient);
+    CC_ASSERT(s_pHttpClient);
```
minggo, line 357: Why not use CC_SAFE_RELEASE_NULL?
It will check the validity of s_pHttpClient.

sburavtsov, line 357: Actually due design nature of HttpClient we should always call getInstance and it will create instance in case it was NULL, so my CC_ASSERT doesn't require too, sorry, and in destructor static var will be nulled, so it is np and can be done both ways. I just put NULL to destructor to make it look more logically.
Note - if we will keep HttpClient instance somewhere else except scheduler (which keeps it while we unschedule handler) this method either will work not as expected (don't destroy instance), or will make static var NULL.
I not sure what can be best suggestion for this case.

minggo:
Thank you.

PR #1507, opened 2012-10-27
fixed a couple typos

Xyzyx101:
There are a couple of capital letters where they don't belong.  It is probably harmless on Windows but it breaks on Linux.


dumganhar:
Thanks. Merged.

PR #1515, opened 2012-10-30
fixed #1531: The CCSprite::updateTransform() function needs to inherited from a base function on the CCNode.

dumganhar:


dumganhar:
Thank @mrmop.

PR #1520, opened 2012-10-30
CCControl* should not respond to touches if the control is not visible

edbartley:
This pull requests modifies some CCControl objects to prevent touch responses if the control is not visible.


dumganhar:
Thank you. Merged.

PR #1521, opened 2012-10-31
fix win32 CCLuaLog memory leaks, and invalid Console UTF8 output

dualface:
fix win32 CCLuaLog memory leaks, and invalid Console UTF8 output

### DIFF
--- a/cocos2dx/platform/win32/CCCommon.cpp
+++ b/cocos2dx/platform/win32/CCCommon.cpp
@@ -68,8 +68,8 @@ void CCLuaLog(const char *pszMsg)
 	WideCharToMultiByte(CP_ACP, 0, widebuff, -1, buff, bufflen, NULL, NULL);
 	puts(buff);
 
-	delete widebuff;
-	delete buff;
+	delete[] widebuff;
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/win32/CCCommon.cpp
```diff
@@ -55,16 +55,21 @@ void CCMessageBox(const char * pszMsg, const char * pszTitle)
...
     OutputDebugStringA("\n");
 
-    puts(pszMsg);
+	bufflen = WideCharToMultiByte(CP_ACP, 0, widebuff, -1, NULL, 0, NULL, NULL);
+	char* buff = new char[bufflen + 1];
+	memset(buff, 0, sizeof(char) * (bufflen + 1));
+	WideCharToMultiByte(CP_ACP, 0, widebuff, -1, buff, bufflen, NULL, NULL);
+	puts(buff);
+
+	delete widebuff;
```
dumganhar: You should use 'delete[]' or 'CC_SAFE_DELETE_ARRAY' to destroy array.
Why not puts pszMsg directly?

PR #1544, opened 2012-11-03
Minor bugfix extensions/GUI/CCScrollView/CCTableView.cpp

ncrothe:
numberOfCellsInTableView returns unsigned int so, if it returns 0 and you subtract 1, you get a very big number and idx is never bigger than that - there are some places that did not handle this very well. I propose a few fixes.

### DIFF
--- a/extensions/GUI/CCScrollView/CCTableView.cpp
+++ b/extensions/GUI/CCScrollView/CCTableView.cpp
@@ -149,7 +149,7 @@ void CCTableView::updateCellAtIndex(unsigned int idx)
 
 void CCTableView::insertCellAtIndex(unsigned  int idx)
 {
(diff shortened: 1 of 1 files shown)


dumganhar:
Hey, I think this issue has been fixed. You haven't pull the lastest code.
You could look inside the file(https://github.com/cocos2d/cocos2d-x/blob/gles20/extensions/GUI/CCScrollView/CCTableView.cpp).

ncrothe:
Drat, you're right. I'll check if that version works for us.

dumganhar:
Hi @ncrothe,
Could i close this Pull Request?

ncrothe:
Yes, the changes seem to work for us, thanks.

PR #1562, opened 2012-11-08
Android: Access to files from APK is boosted

mingulov:
Open APK file just once and cache a file list information (name and position inside archive). Also a general ZipFile helper class is created to read multiple files from Zip archives - might be used for another platforms if needed.

Previously it was so for each separate file access operation:
- Open zip archive, initially parse its structure (unzOpen)
- Locate the particular file there by a linear search through full archive (unzLocateFile).
  So if file does not exist - still an archive has to be handled fully, even slower than a file exists.
- Read file.
- Close zip archive

After this commit:
- Once - open zip/apk file and collect file list information
  (average consumed time - the same like a search for 2 different files by unzLocateFile)
- When needed - directly retrieve file position inside the archive, setting it up to the zip reader and reading file

ZipFile class is located in support/zip_support/ZipUtils.h and .cpp to prevent creation of a new files and adding them to multiple project files.

std::map is used to store file list information, if supported by a platform - might be changed to std::unordered_map locally (defined at ZipUtils.cpp).

ZipFIle class declaration uses private implementation to prevent overloading header with a lot of other headers (unzip.h etc).

Speedup is visible even on just a simple TestCpp project on TextureCacheTest. On more complex projects with a lot assets / multiple loading / reloading operations - speedup is much more significant.

### DIFF
--- a/cocos2dx/support/zip_support/ZipUtils.cpp
+++ b/cocos2dx/support/zip_support/ZipUtils.cpp
@@ -300,7 +300,7 @@ public:
     unzFile zipFile;
 
     // std::unordered_map is faster if available on the platform
-    typedef std::map<std::string, ZipEntryInfo> FileListContainer;
+    typedef std::map<std::string, struct ZipEntryInfo> FileListContainer;
     FileListContainer fileList;
 };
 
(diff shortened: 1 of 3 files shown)


cocos2dx/support/zip_support/ZipUtils.cpp
```diff
@@ -280,4 +284,139 @@ int ZipUtils::ccInflateCCZFile(const char *path, unsigned char **out)
...
+        CC_BREAK_IF(it ==  m_data->fileList.end());
+
+        unz_file_pos filePos = it->second;
+
+        int nRet = unzGoToFilePos(m_data->zipFile, &filePos);
+        CC_BREAK_IF(UNZ_OK != nRet);
+
+        char szFilePathA[UNZ_MAXFILENAMEINZIP + 1];
+        unz_file_info FileInfo;
+        nRet = unzGetCurrentFileInfo(m_data->zipFile, &FileInfo,
```
minggo: It seems that szFilePathA is not needed, can we pass NULL to unzGetCurrentFileInfo()?

cocos2dx/support/zip_support/ZipUtils.cpp
```diff
@@ -280,4 +284,139 @@ int ZipUtils::ccInflateCCZFile(const char *path, unsigned char **out)
...
+                szFilePathA, sizeof(szFilePathA) - 1, NULL, 0, NULL, 0);
+        CC_BREAK_IF(UNZ_OK != nRet);
+
+        nRet = unzOpenCurrentFile(m_data->zipFile);
+        CC_BREAK_IF(UNZ_OK != nRet);
+
+        pBuffer = new unsigned char[FileInfo.uncompressed_size];
```
minggo: Can cache uncompressed_size too?

cocos2dx/platform/android/CCFileUtils.cpp
```diff
@@ -43,6 +45,7 @@ CCFileUtils* CCFileUtils::sharedFileUtils()
     {
         s_pFileUtils = new CCFileUtils();
         s_strResourcePath = getApkPath();
+        s_pZipFile = new ZipFile(s_strResourcePath, "assets/");
```
minggo: s_pZipFile is not deleted?

minggo:
It is an awesome work.
Thank you.

cocos2dx/support/zip_support/ZipUtils.cpp
mingulov: Thank you, a call to unzGetCurrentFileInfo at ZipFile::getFileData is removed, uncompressed_size is cached at ZipFile::setFilter together with file pos info.

cocos2dx/platform/android/CCFileUtils.cpp
mingulov: Fixed, will be deleted at CCFileUtils::purgeFileUtils together with s_pFileUtils.

PR #1569, opened 2012-11-14
Fix CCMenu::setEnabled(false) doesn't work after ccTouchBegan

wenbin1989:
If call CCMenu::setEnabled(false) after CCMenu's ccTouchBegan method is
called, menuItems in the menu will still be selected and actived.

### DIFF
--- a/cocos2dx/menu_nodes/CCMenu.cpp
+++ b/cocos2dx/menu_nodes/CCMenu.cpp
@@ -214,14 +214,19 @@ void CCMenu::onExit()
 
 void CCMenu::setHandlerPriority(int newPriority)
 {
+    m_handlerPriority = newPriority;
+    
     CCTouchDispatcher* pDispatcher = CCDirector::sharedDirector()->getTouchDispatcher();
-    pDispatcher->setPriority(newPriority, this);
+    if (pDispatcher->findHandler(this))
+    {
+        pDispatcher->setPriority(newPriority, this);
+    }
 }
 
 void CCMenu::registerWithTouchDispatcher()
 {
     CCDirector* pDirector = CCDirector::sharedDirector();
-    pDirector->getTouchDispatcher()->addTargetedDelegate(this, kCCMenuHandlerPriority, true);
+    pDirector->getTouchDispatcher()->addTargetedDelegate(this, m_handlerPriority, true);
 }
 
 bool CCMenu::ccTouchBegan(CCTouch* touch, CCEvent* event)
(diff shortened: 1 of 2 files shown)


minggo:
Could you give some demo or describe how to reproduce it step by step?
Thank you.

wenbin1989:
Here is the demo code(https://gist.github.com/4070914), please overwrite to `samples/HelloCpp/Classes/HelloWorldScene.h` and `samples/HelloCpp/Classes/HelloWorldScene.cpp`, then run the HelloCpp demo.

In the demo code, I want to disable the close menu when drag the layer, which registers to targetedDelegate with priority higher than CCMenu and swallow false. It is used frequently in scrollview.
Although I call `m_pMenu->setEnabled(false)` in the layer's `ccTouchMoved` method, the close menu will still active when touch end, it is not disabled. This is because CCMenu only handle enable property in ccTouchBegan, after that, it will be useless to call setEnabled(false).
The patch can fix this problem.

Just have a look, thanks.

wenbin1989:
About the second commit, in the origin `samples/HelloCpp` sample, just add

``` C++
pMenu->setHandlerPriority(-1);
```

below

``` C++
// create menu, it's an autorelease object
CCMenu* pMenu = CCMenu::create(pCloseItem, NULL);
pMenu->setPosition(CCPointZero);
this->addChild(pMenu, 1);
```

you'll find the problem.

minggo:
@wenbin1989 
I don't find what problem it is.
Could you please describe it more clearly? Describe the steps you do, and what issue it cause.

minggo:
Did you mean that
1. Touch the close menu
2. Drag the layer
3. Touch ended, the ended position in the content size of close menu
4. Close menu will response the touch ended event

Close menu is disabled when dragging, so it should not response touch ended event.
Right?

wenbin1989:
@minggo 

Yes it it!

The demo maybe is not so proper, sorry about that.

This problem most happens in scrollview with menu in it.  The normal way is to disable the menu when scroll the view. After scrolled, even if the touchended position is in the content size of the menu, the menu still should not be actived.

Current implement of CCMenu::setEnable cannot handle that situation, CCMenu::setHandlerPriority also has some bugs if it's called before onEnter.

dumganhar:
Since we renamed gles20 branch to master.
I closed this Pull Request. And sent it to master branch here(https://github.com/cocos2d/cocos2d-x/pull/1850).
We should discuss in that url. Thank you.

PR #1570, opened 2012-11-14
Tiff support and resource location

mrmop:
- Tiff support added to CCImage
- CCFileUtils updated to support resource folders
- Examples updated to support multiple resolution assets


cocos2dx/platform/marmalade/CCFileUtils.cpp
```diff
@@ -70,21 +70,26 @@ const char* CCFileUtils::fullPathFromRelativePath(const char *pszRelativePath)
...
     {
 		char szDriver[3] = {s_pszResourcePath[0], s_pszResourcePath[1], 0};
         pRet->m_sString = szDriver;
+        pRet->m_sString += resDir;
         pRet->m_sString += pszRelativePath;
     }
     else
     {
         pRet->m_sString = s_pszResourcePath;
+        pRet->m_sString += resDir;
```
dumganhar, line 92: Hi Mat,
Thank you for adding this resource directory feature.
But you might miss one thing, that is we should check whether the file exists. If not, we should search the root directory of resource folder.
There are some implementations of other platform(win32,linux,mac,ios) you could refer to.
The android port's implementation is a little different since this function just return _pszRelativePath_ which is passed into. 
Because android is using just one _apk_ file, we did this file check logic at _getFileData_.

dumganhar:
Hi Mat,
I merged this PR, thanks.
But will you implement that I said above?

PR #1578, opened 2012-11-15
Update tools/tolua++/build.php

thbourlove:
If you run this:

```
tolua++ -o LuaCocos2d.cpp /path_to_cocos2d/cocos2d-2.0.3/tools/tolua++/Cocos2d.pkg
```

tolua++ will regard the 'cocos2d-2' as package name, but which you want is 'Cocos2d'(input file root name).

### DIFF
--- a/tools/tolua++/build.php
+++ b/tools/tolua++/build.php
@@ -37,8 +37,8 @@ class Builder
 
     function build()
     {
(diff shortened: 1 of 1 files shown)


dumganhar:
Thanks. :)

PR #1581, opened 2012-11-16
add libXrandr library

guanqun:
From glfw v2.6, it adds Xrandr code path, so in order to compile Linux
successfully, we need to add this library.

Otherwise, there are following errors:

```
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRSetScreenConfig'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRConfigCurrentConfiguration'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRSelectInput'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRFreeScreenConfigInfo'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRConfigCurrentRate'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRConfigRates'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRQueryExtension'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRGetScreenInfo'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRUpdateConfiguration'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRSetScreenConfigAndRate'
../../../lib/linux/Debug/libcocos2d.so: undefined reference to `XRRConfigSizes'
collect2: ld returned 1 exit status
make: *** [TestCpp] Error 1
```

### DIFF
--- a/cocos2dx/proj.linux/Makefile
+++ b/cocos2dx/proj.linux/Makefile
@@ -208,7 +208,7 @@ STATICLIBS = $(STATICLIBS_DIR)/libfreetype.a \
 				$(STATICLIBS_DIR)/libtiff.a \
 
 SHAREDLIBS = 
-SHAREDLIBS += -lglfw -lGL -lcurl -lfontconfig
+SHAREDLIBS += -lglfw -lGL -lcurl -lfontconfig -lXrandr
 SHAREDLIBS += -Wl,-rpath,../../cocos2dx/platform/third_party/linux/glew-1.7.0/glew-1.7.0/lib
 SHAREDLIBS += -L../../cocos2dx/platform/third_party/linux/glew-1.7.0/glew-1.7.0/lib -lGLEW
 


mdkcore0:
I had to do something like this too, but I just have added -lXrandr on the cocos2d-x linux project, as you can see here: https://github.com/rodrigogolive/cocos2d-x/commit/a3cf5f7a537754f8e2d5f69b8a387ec7551f648b

I've did a pull request with some changes for the linux version, but it wasn't accepted until today yet :p

dumganhar:
Hi, @rodrigogolive 
Sorry for our delay of your PR.
And I am using ArchLinux 64 bit, but it doesn't need to add -lXrandr library.
Can it be compiled through all linux distribution if adding -lXrandr?

mdkcore0:
Hi @dumganhar, no problem ;)

It's expected to run, as it's a library dependency. But as exist a lot of linux distributions, I can't answer you exactly. What I can do is create some virtual machines and test on some linux flavors (debian/ubuntu, fedora, arch, gentoo, slackware).

What do you think?

guanqun:
@dumganhar for some distributions, it might not need such fixes, because as I said in the description, after glfw 2.6, it adds the xrandr dependency. so if the version is before 2.6, i don't think we need it.

dumganhar:
@guanqun 
But i'm using glfw2.7 and I don't need to add xrandr.

<pre>
community/glfw 2.7.7-1 [installed]
    A free, open source, portable framework for OpenGL application development
</pre>

dumganhar:
@rodrigogolive 
It's will be great if no problems at most distributions.
Another issue it that we are using apt-get to get the dependent libraries in the **make-all-linux-project.sh** .
Could we use different commands on different distributions?
For example, on Arch Linux, it will be _sudo pacman -S glfw_.
I know it's a painful thing that there are so many distibutions. :(

guanqun:
@dumganhar weird though, maybe the glfw has the options to enable the xrandr support or not. If more and more people report this issue, we can enable this.

mdkcore0:
@guanqun and @dumganhar: Here I compiled glfw, and there are no way to explicit disable Xrandr on the compilation (it doesn't provide a configure.sh to pass command, just a makefile and had to use a especific argument to compile: make x11).

Maybe who created the package for some specific distro have created a patch to explicit disable the calls for xrandr. Can you check this?

@dumganhar I can work on the make-all-linux-project.sh to try to cover some distros on it. But we need to choose some, as many of the use different package systems and, maybe, different libraries package names. So, on what distros we can focus on?

I'll remove that patch from my pull request and create another one soon with just this issue.

guanqun:
@rodrigogolive thanks for your detailed investigation, I'll check later if there's a patch to disable it. about the distros to support, I would suggest we focus on two types: debian based and red hat based, i.e. `apt-get` and `yum`. That would cover the most common distros, IMHO.

mdkcore0:
@guanqun: sorry for the delay, I'm very busy these days

It'll work on it soon as possible, and will follow your suggestions as well ;)

dumganhar:
@rodrigogolive 

> I can work on the make-all-linux-project.sh to try to cover some distros on it. But we need to choose some, as many of the use different package systems and, maybe, different libraries package names. So, on what distros we can focus on?

Basically, we need to make it works on Ubuntu, OpenSUSE, Fedora, Mandriva, Slackware and ArchLinux at least.

dumganhar:
Since this Pull Request has no replies. I will close it firstly.
We will rename gles20 branch to master.
So could you send the Pull Request to master branch(default branch) later?

PR #1587, opened 2012-11-16
Rssource path root change

mrmop:
fullPathFromRelativePath will now search root if file is not found in
path


dumganhar:
Thanks. :)

PR #1589, opened 2012-11-16
improves the debugger

rolandoam:
Some changes to the debugger. This changes a little bit the way the debugger is executed: There's no way to create new globals (compartments) from js anymore. Also, to enable the debugger, you can call `scriptingCore->enableDebugger()`. That will load the `debugger.js` script, prepare the debugger environment and expose the `startDebugger` function, which you can test for in your js code, and then call with the right arguments. I will update also the `debugger.js` and the example `main.js` (as well as the instructions) in the cxx-generator repo.

### DIFF
--- a/scripting/javascript/bindings/ScriptingCore.cpp
+++ b/scripting/javascript/bindings/ScriptingCore.cpp
@@ -218,7 +218,7 @@ void jsb_register_cocos2d_config( JSContext *_cx, JSObject *cocos2d)
 // #else
     str = JS_InternString(_cx, "mobile");
 // #endif
-    JS_DefineProperty(_cx, ccconfig, "deviceType", STRING_TO_JSVAL(str), NULL, NULL, JSPROP_ENUMERATE | JSPROP_READONLY | JSPROP_PERMANENT);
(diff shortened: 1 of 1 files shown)


scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -219,7 +218,7 @@ void jsb_register_cocos2d_config( JSContext *_cx, JSObject *cocos2d)
 // #else
     str = JS_InternString(_cx, "mobile");
 // #endif
-    JS_DefineProperty(_cx, ccconfig, "platform", STRING_TO_JSVAL(str), NULL, NULL, JSPROP_ENUMERATE | JSPROP_READONLY | JSPROP_PERMANENT);
+    JS_DefineProperty(_cx, ccconfig, "deviceType", STRING_TO_JSVAL(str), NULL, NULL, JSPROP_ENUMERATE | JSPROP_READONLY | JSPROP_PERMANENT);
```
dumganhar: Hi Rolando,
_cc.Config.deviceType_ has been changed to _cc.Config.plaform_ in the latest _cocos2d-js-tests_.
So why you changed it back to _deviceType_?

rolandoam: sorry, my mistake: it was a merge conflict and I picked the wrong version :)

dumganhar:
Thanks.

PR #1591, opened 2012-11-17
The S3E data dir ONLY used by the emulator/simulator.

fape:
I got error while try to deploy with Marmalade to Android or Blackberry.
See also: https://devnet.madewithmarmalade.com/questions/1008/s3e-data-dir-ignored.html

Documentation interesting part: http://docs.madewithmarmalade.com/native/devtools/mkbfilesettings/buildoptions.html#s3eenvironmentoptions

### DIFF
--- a/samples/HelloCpp/proj.marmalade/HelloCPP.mkb
+++ b/samples/HelloCpp/proj.marmalade/HelloCPP.mkb
@@ -16,7 +16,7 @@ subprojects
 
 assets
 {
-	(data)
+	(../Resources)
 	fonts
 	ipad
 	ipadhd
--- a/samples/HelloLua/proj.marmalade/HelloLua.mkb
+++ b/samples/HelloLua/proj.marmalade/HelloLua.mkb
@@ -19,7 +19,7 @@ subprojects
 
 assets
 {
-	(data)
+	(../Resources)
 	background.mp3
 	background.ogg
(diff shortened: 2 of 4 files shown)


dumganhar:
Thanks.
So **s3e-data-dir** only for **emulator/simulator**, **assets** for devices?
BTW, could you give me an article about how to deploy application to android device, since I only make it run on simulator.

fape:
You have to select an other "Solution Configuration" in visual studio to deploy to devices and press "Start". E.g Use "GCC (ARM) Debug/Release" [Imgur](http://i.imgur.com/58zV8) for android and blackberry. After everything compiled you'll get "Marmalade System Deployment Tool". In this "Wizard" you can select the target devices platform [Imgur](http://i.imgur.com/MunpY). In the "Stage Deploying"  pressing Run button'll generate the deploy-able file (apk).

dumganhar:
Yes, i did it as you said, but when i deploy it to android device, _Marmalade System Deployment Tool_ told me there is an error. Something like _can't find the Signature_.
Since it's weekend now, i will tell you the detail error information next monday. 
Thanks again. :)

fape:
Ok, keep in touch. 
What version of java/jdk do you use? I think I got the same when I tried with java 7. Please try with jdk6.

PR #1611, opened 2012-11-21
Add glBlendFuncSeparate to the cache system

pidjay:
This is needed when we use different blending modes for the RGB and alpha channels.

Signed-off-by: Pierre-Jean Quillere pj.quillere@moglue.com

### DIFF
--- a/cocos2dx/shaders/ccGLStateCache.cpp
+++ b/cocos2dx/shaders/ccGLStateCache.cpp
@@ -49,6 +49,8 @@ static GLuint    s_uCurrentShaderProgram = -1;
 static GLuint    s_uCurrentBoundTexture[kCCMaxActiveTexture] =  {(GLuint)-1,(GLuint)-1,(GLuint)-1,(GLuint)-1, (GLuint)-1,(GLuint)-1,(GLuint)-1,(GLuint)-1, (GLuint)-1,(GLuint)-1,(GLuint)-1,(GLuint)-1, (GLuint)-1,(GLuint)-1,(GLuint)-1,(GLuint)-1, };
 static GLenum    s_eBlendingSource = -1;
 static GLenum    s_eBlendingDest = -1;
+static GLenum    s_eBlendingSourceAlpha = -1;
+static GLenum    s_eBlendingDestAlpha = -1;
 static int       s_eGLServerState = 0;
 static GLuint    s_uVAO = 0;
 #endif // CC_ENABLE_GL_STATE_CACHE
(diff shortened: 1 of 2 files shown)


cocos2dx/shaders/ccGLStateCache.cpp
```diff
@@ -101,30 +103,49 @@ void ccGLUseProgram( GLuint program )
...
-    if (sfactor != s_eBlendingSource || dfactor != s_eBlendingDest)
+    if (sfactor != s_eBlendingSource || dfactor != s_eBlendingDest ||
+		sfactor != s_eBlendingSourceAlpha || dfactor != s_eBlendingDestAlpha)
     {
         s_eBlendingSource = sfactor;
         s_eBlendingDest = dfactor;
-        SetBlending(sfactor, dfactor);
+		s_eBlendingSourceAlpha = sfactor;
+        s_eBlendingDestAlpha = dfactor;
+        SetBlending(sfactor, dfactor, sfactor, dfactor);
```
minggo, line 129: glBlendFunc(sfactor, dfactor) == glBlendFuncSeparate(sfactor, dfactor, sfactor, dfactor)?

pidjay, line 129: @minggo Yes, that is correct.
This is explained in the official wiki of OpenGL's website: http://www.opengl.org/wiki/GLAPI/glBlendFuncSeparate#Description

minggo, line 129: glBlendFunc(GL_SRC_ALPA_SATURATE, GL_ZERO) != glBlendFuncSeparate(GL_SRC_ALPA_SATURATE, GL_ZERO, GL_SRC_ALPA_SATURATE, GL_ZERO).

Because GL_SRC_ALPA_SATURATE can only be used for source RGB blend factor.

pidjay, line 129: Is it? I'm not really familiar with OpenGL programming, but what I've read is that GL_SRC_ALPHA_SATURATE can be used for both RGB and alpha blending factors.
Correct me if I'm wrong.
Source: http://www.khronos.org/opengles/sdk/docs/man/xhtml/glBlendFuncSeparate.xml

minggo, line 129: "OpenGL Programming Guide" said that.

And not all SetBlending() usage is changed.

pidjay, line 129: Ah yes... I wonder why the compiler didn't catch that for me :/

PR #1620, opened 2012-11-21
remove physics_nodes from marmalade

fape:
Remove physics_nodes path from marmalade cocos2dx.mkf file.


dumganhar:
Thank you. :)

PR #1644, opened 2012-11-24
Update samples/TestCpp to compatible with marmalade

fape:
- Update samples/TestCpp to compatible with marmalade, 
- filter out some test
- Readd physics_nodes with new, correct path (some kind of revert of #1620)


dumganhar:
Thanks. :+1:

PR #1645, opened 2012-11-25
Update marmalade languages

fape:
- fix Russian language select
- Add Korean language 
- set default language to English


dumganhar:
Thanks, merged.

PR #1646, opened 2012-11-25
Fix missing files from templates

guanqun:
When we create the iphone project from xcode template, it can't even compile successfully by default. The root cause of this issue is that the `.plist` files in the template files are not corrected accordingly. 

This branch fixes this issue.

Please review!


minggo:
I think i fix it in this PR https://github.com/cocos2d/cocos2d-x/pull/1652.

Thank you.

PR #1652, opened 2012-11-26
Gles20

minggo:


minggo:
1. Use macro CC_TARGET_OS_IPHONE and CC_TARGET_OS_MAC instead of TARGET_OS_IPHONE and TARGET_OS_MC.
2. Update xcode template to make it work.

PR #1661, opened 2012-11-27
fix memory leak in uthash

guanqun:


dumganhar:
Hi, could you also send this patch to the UTHASH official?
Thanks.

guanqun:
Hi @dumganhar a good suggestion, and then I downloaded the official UTHASH source code and found it's already been fixed there... :)  So you can just pull from the official UTHASH and I can happily ignore my credit for this patch, as I'm not the first one to spot this issue. It should be given to the right person.

dumganhar:
Thank you, issue 1592 (http://www.cocos2d-x.org/issues/1592) was created.

dumganhar:
Closed. We will update UTHASH before next version release. Thanks. :)

PR #1662, opened 2012-11-27
when user specified 'group' parameter, it should be parsed as an 'int'

guanqun:
Otherwise, when we do slicing, it would report type error.


minggo:
Could you please tell me how to cause type error?
Thank you.

guanqun:
@minggo just pass the parameter "-g 0" to this python script.

minggo:
Thank you.

PR #1668, opened 2012-11-28
issue #1581: JSBinding bug fixes.

dumganhar:
Some fixes of JSBinding codes:
- [1] Check whether the proxy was already added in <code> JS_NEW_PROXY</code>
- [2] In struct <code>schedFunc_proxy_t</code>, <code>JSScheduleWrapper_</code> --> <code>CCArray_</code>
  **Reason**: One js function may correspond to many targets. To debug this, you could refer to JSScheduleWrapper::dump function. It will prove that i'm right. :)
- [3] In <code>ScriptingCore::cleanupSchedulesAndActions</code> function, we must invoke unschedule for all targets and remove the proxy both in <code>_schedFunc_target_ht</code> and <code>_schedTarget_native_ht</code>, otherwise the hash tables will grow bigger and bigger, so I added a new function JSScheduleWrapper::removeAllTargetsForNatiaveNode to make these things easier.
- [4] To easily find out the bugs of binding codes, I added <code>JS_SetGCZeal</code> in <code>ScriptingCore::createGlobalContext</code>, it only works in DEBUG mode.
- [5] In js_cocos2dx_CCNode_getChildren, we should add the generated array to root to avoid gc happen when invoking JS_SetElement.
- [6] The JSCallFuncWrapper isn't needed since an action will be run by a cc.Node and it will be released at the CCNode::cleanup. It's different from the binding of schedule.
- [7] Some improvements of JSScheduleWrapper class.
- [8] Added a new function JSScheduleWrapper::setTarget, it's for <code>js_CCNode_unschedule</code> to find out which target need to be unscheduled.
- [9] Commented <code>JS_SetReservedSlot</code> in <code>js_CCNode_scheduleOnce</code> and <code>js_CCNode_schedule</code>.
  **Reason**:
  For <code>js_CCNode_scheduleOnce</code>: Don't add the callback function to the reserved slot of this js object.Since the class of js object may be inherited from cocos class(e.g. cc.Sprite). The subclass will not contain reserved slots. It will crash if invoking this.
  For <code>js_CCNode_schedule</code>: Don't add js callback function to the reserved slot of scheduler js object. Since the scheduler is an object always rooted.
  So the callback function might not be released when gc comes.I looked inside the implementation of cc.Node.schedule, and it doesn't use JS_SetReservedSlot there.

### DIFF
--- a/scripting/javascript/bindings/cocos2d_specifics.cpp
+++ b/scripting/javascript/bindings/cocos2d_specifics.cpp
@@ -65,7 +65,6 @@ JSBool js_cocos2dx_CCNode_getChildren(JSContext *cx, uint32_t argc, jsval *vp)
 			cocos2d::CCArray *children = node->getChildren();
 
 			JSObject *jsarr = JS_NewArrayObject(cx, children->count(), NULL);
-            JS_AddNamedObjectRoot(cx, &jsarr, "CCNode_getChildren");
 
 			for (int i=0; i < children->count(); i++) {
 				cocos2d::CCNode *child = (cocos2d::CCNode*)children->objectAtIndex(i);
@@ -73,8 +72,6 @@ JSBool js_cocos2dx_CCNode_getChildren(JSContext *cx, uint32_t argc, jsval *vp)
 				jsval childVal = OBJECT_TO_JSVAL(childProxy->obj);
 				JS_SetElement(cx, jsarr, i, &childVal);
 			}
-
-            JS_RemoveObjectRoot(cx, &jsarr);
 
 			JS_SET_RVAL(cx, vp, OBJECT_TO_JSVAL(jsarr));
 		}


scripting/javascript/bindings/cocos2d_specifics.cpp
```diff
@@ -63,13 +63,19 @@ JSBool js_cocos2dx_CCNode_getChildren(JSContext *cx, uint32_t argc, jsval *vp)
 		if (proxy) {
 			cocos2d::CCNode *node = (cocos2d::CCNode *)(proxy->ptr ? proxy->ptr : NULL);
 			cocos2d::CCArray *children = node->getChildren();
+
 			JSObject *jsarr = JS_NewArrayObject(cx, children->count(), NULL);
```
rolandoam, line 67: We should not be building the array later, this is what I would do

``` c++
jsval elts[children->count()];
for (...) {
    // set elts[i] here
}
JSObject *jsarr = JS_NewArrayObject(cx, children->count(), jsval);
```

That way, we won't need to root anything.

dumganhar, line 67: Agree!
Thanks.
Did you some other opinions about my other changes?
I like std::vector more than CCArray too. But I don't want to make lots of changes in this release(Next Tuesday).

dumganhar, line 67: Hi Rolando,
Did you check the offical document about gc? https://developer.mozilla.org/en/docs/SpiderMonkey_Garbage_Collection_Tips

Avoid malloc'ing temporary storage that contains unrooted jsvals:
/\* Bad! */
jsint i, len;
jsval *vec;
JSString *str;
JSObject *myArrayObj;

len = NumberOfNativeStrings();
vec = JS_malloc(cx, len \* sizeof(jsval));
if (!vec) return JS_FALSE;
for (i = 0; i < len; i++) {
    str = JS_NewStringCopyZ(cx, GetNativeString(i));
    if (!str) {
        JS_free(cx, vec);
        return JS_FALSE;
    }
    vec[i] = STRING_TO_JSVAL(str);
}
myArrayObj = JS_NewArrayObject(cx, len, vec);
JS_free(cx, vec);
if (!myArrayObj) return JS_FALSE;
OtherStuffThatMightGC();
*rval = OBJECT_TO_JSVAL(myArrayObj);

/\* Good! */
JSObject *myArrayObj;
jsint i, len;
JSString *str;
jsval val;

myArrayObj = JS_NewArrayObject(cx, 0, NULL);
if (!myArrayObj) return JS_FALSE;
*rval = OBJECT_TO_JSVAL(myArrayObj);
len = NumberOfNativeStrings();
for (i = 0; i < len; i++) {
    str = JS_NewStringCopyZ(cx, GetNativeString(i));
    if (!str) return JS_FALSE;
    val = STRING_TO_JSVAL(str);
    if (!JS_SetElement(cx, myArrayObj, i, &val))
        return JS_FALSE;
}
OtherStuffThatMightGC();
Note that this example also shows tip #3 (root as you go).

dumganhar, line 67: Yep, i was wrong, JS_SetElement will not trigger a GC.
Deleted it.

scripting/javascript/bindings/cocos2d_specifics.cpp
```diff
@@ -900,28 +1004,29 @@ JSBool js_CCNode_scheduleOnce(JSContext *cx, uint32_t argc, jsval *vp)
...
+            sched->scheduleSelector(schedule_selector(JSScheduleWrapper::scheduleFunc), tmpCobj, 0, 0, delay, !node->isRunning());
         }
-        
+        // Don't add the callback function to the reserved slot of this js object.
+        // Since the class of js object may be inherited from cocos class(e.g. cc.Sprite)
+        // The subclass will not contain reserved slots.
+        //JS_SetReservedSlot(proxy->obj, 0, argv[0]);
         JS_SET_RVAL(cx, vp, JSVAL_VOID);
-        
-        JS_SetReservedSlot(proxy->obj, 0, argv[0]);
```
rohankuruvilla, line 923: We need to set ReservedSlot for the case where the user passes in an anonymous function. Which if not rooted, or added as a reserved slot to the node, will be collected. We can do something like the following to overcome the subclass issue (Taken from cocos2d-iphone):

```
JSClass *klass = JS_GetClass(p->obj);
int slots = JSCLASS_RESERVED_SLOTS(klass);

if(slots >= 1 ) {
     JS_SetReservedSlot(proxy->obj, 0, argv[0]);
} 
```

PR #1670, opened 2012-11-29
Fixing Accelerometer crash on Android

rohankuruvilla:
Thanks @folecr for the help

### DIFF
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxAccelerometer.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxAccelerometer.java
@@ -70,13 +70,13 @@ public class Cocos2dxAccelerometer implements SensorEventListener {
 	// ===========================================================
 
 	public void enable() {
-		this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_GAME);
+		this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_NORMAL);
 	}
 
         public void setInterval(float interval) {
 	        // Honeycomb version is 11
 	        if(android.os.Build.VERSION.SDK_INT < 11) {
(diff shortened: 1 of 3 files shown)


cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxAccelerometer.java
```diff
@@ -70,13 +70,13 @@ public Cocos2dxAccelerometer(final Context pContext) {
 	// ===========================================================
 
 	public void enable() {
-		this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_GAME);
+		this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_NORMAL);
```
minggo, line 73: Why not use SENSOR_DELAY_GAME as default? Because our engine is used for game development.

cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxAccelerometer.java
```diff
@@ -70,13 +70,13 @@ public Cocos2dxAccelerometer(final Context pContext) {
...
 	public void enable() {
-		this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_GAME);
+		this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_NORMAL);
 	}
 
         public void setInterval(float interval) {
 	        // Honeycomb version is 11
 	        if(android.os.Build.VERSION.SDK_INT < 11) {
-		    this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_GAME);
+		    this.mSensorManager.registerListener(this, this.mAccelerometer, SensorManager.SENSOR_DELAY_NORMAL);
```
minggo, line 79: The same question as mentioned above.

PR #1677, opened 2012-11-29
Fix for CCRenderTexture on Android sleep/resume

chenium:
When an android app goes into the background, all gl context is lost, including any frame buffer objects that are required for rendering to a texture. When the android app returns to focus, a new FBO is generated for each render texture allowing rendering to the texture to continue.


minggo:
Thank you.

PR #1686, opened 2012-11-30
Quick fix to resolve a crash on some Android devices  

folecr:
- CCParticleSystemQuad::postStep() was updated to use claimed higher performance code from cocos2d-iphone
- The new code causes a crash on some Android devices (Nexus 4, Galaxy SIII) but not on others (Galaxy Nexus, HTC G10)
- This patch uses the older code to attempt to resolve this crash

I haven't fully investigted the reason for the crash but am submitting this patch to stabilize for the next release. Please review before merging!


minggo:
Thank you.

PR #1696, opened 2012-12-02
Update for Lua marmalade

fape:
- Add ignore marmalade generated Visual Studio directory
- Update Lua and HelloLua marmalade project file
- Add TestLua marmalade project

### DIFF
--- a/samples/TestLua/proj.marmalade/TestLua.mkb
+++ b/samples/TestLua/proj.marmalade/TestLua.mkb
@@ -1,6 +1,6 @@
 options
 {
-	module_path="../../../cocos2dx/proj.marmalade/;../../../CocosDenshion/proj.marmalade/;../../../scripting/lua/proj.marmalade/"
+	module_path="../../../cocos2dx/proj.marmalade/;../../../CocosDenshion/proj.marmalade/;../../../scripting/lua/proj.marmalade/;../../../extensions/proj.marmalade/;../../../external/chipmunk/proj.marmalade/;../../../external/Box2D/proj.marmalade/"
 	s3e-data-dir = "../Resources"
 }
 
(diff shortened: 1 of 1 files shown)


dumganhar:
Thanks @fape .
We decide to release a new version on 4.12.2012. So what about the status of marmalade port now?
If you have more improvements to commit, please send it these days.
Thanks again for your contribution. :) I will test this Pull Request tomorrow since it's late at night now.

dumganhar:
BTW, please don't forget to update the _AUTHOR_ file.

fape:
- All samples with Marmalade project (HelloCpp, HelloLua, TestCpp, TestLua, SimpleGame) are compiled without error. But as I see most of the test in TestCpp and TestLua are NOT working right now.
- External: Only Box2D works. (Chipmunk does not) 
- TestLua works only on devices. The program doesn't find some image/sound/etc resources in Marmalade emulator/simulator, because 's3e-data-dir' option does not support multiple directory. I'm going to find a "workaround" for this problem tomorrow.

dumganhar:
Good to know that all projects are compiled successfully.
Yep, there are some issues in TestLua. We will improve it in the future.

_External: Only Box2D works. (Chipmunk does not)_
Why Chipmunk cannot work?

_TestLua works only on devices. The program doesn't find some image/sound/etc resources in Marmalade emulator/simulator, because 's3e-data-dir' option does not support multiple directory. I'm going to find a "workaround" for this problem tomorrow._
Did you look inside win32 project of **TestLua**? We use pre-build event to copy all resources to just one folder since win32 and android also don't support multiple resource directories. Maybe marmalade port can refer to this way? Right?

fape:
Why Chipmunk cannot work?
Because Marmalade's engineers made some changes with Chipmunk 5.4.3 for work together with Marmalade (https://github.com/marmalade/chipmunk) And I didn't have enough time yet to check with Cocos's Chipmunk v6.1.2. what's working and what is not working. But I'd like to do that.

dumganhar:
Thanks, it's merged.

PR #1712, opened 2012-12-04
update CHANGELOG

minggo:


vsaase:
why use metrics.density? this value does not behave as expected, as described in the android api reference:

"This value does not exactly follow the real screen size (as given by xdpi and ydpi, but rather is used to scale the size of the overall UI in steps based on gross changes in the display dpi. For example, a 240x320 screen will have a density of 1 even if its width is 1.8", 1.3", etc. However, if the screen resolution is increased to 320x480 but the screen size remained 1.5"x2" then the density would be increased (probably to 1.5)."

i suggest to change this to metrics.xdpi

minggo:
@dumganhar please have a look at this issue.

PR #1724, opened 2012-12-04
CCScrollView and CCTableView updates

edbartley:
This pull request fixes an issue with CCScrollView scrolling incorrectly if the container items do not fill or exceed the ScrollView bounds. The issue can be observed by modifying TableViewTestLayer::numberOfCellsInTableView(…) in the ExtensionsTest/TableViewTest project from 20 cells to 2 cells.

This pull request also adds additional functionality to the CCTableView delegates:
1) adds CCTableViewDelegate::tableCellHighlight(...) to inform the delegate that a table cell is currently being pressed. Thus allowing the implementer to highlight the cell or perform some animation and provide visual feedback of the press to the user.

2) adds CCTableViewDelegate::tableCellUnhighlight(...) to inform the delegate that a table cell is no longer being pressed. Thus allowing the implementer to unhighlight the cell or perform some animation and provide visual feedback of the release to the user.

### DIFF
--- a/extensions/GUI/CCScrollView/CCScrollView.cpp
+++ b/extensions/GUI/CCScrollView/CCScrollView.cpp
@@ -346,8 +346,8 @@ void CCScrollView::relocateContainer(bool animated)
     newY     = oldPoint.y;
     if (m_eDirection == kCCScrollViewDirectionBoth || m_eDirection == kCCScrollViewDirectionHorizontal)
     {
-        newX     = MIN(newX, max.x);
         newX     = MAX(newX, min.x);
+        newX     = MIN(newX, max.x);
     }
 
     if (m_eDirection == kCCScrollViewDirectionBoth || m_eDirection == kCCScrollViewDirectionVertical)


dumganhar:
Hi @edbartley ,
Your improvement looks great.
But I found an issue, that is cells will not go back to its origin position after i move it. This happens when I return _2_ in _numberOfCellsInTableView_ . I captured some picture as follows:
- Before moving the horizontal cells

Alt text
- After moving the horizontal cells

Alt text

Any idea about that? The cells need to go back

BTW, as i know, there are some issues of running action for CCTableViewCell. Since CCTableView reuse cells that are out of screen, some actions will stop or have wrong behaviors after moving cells. I think CCTableView needs to be refactored. How about your opinion?

edbartley:
Thank you for testing! There was a small regression in my local code-base when I was reverting debug code. I just committed the fix for the horizontal scroll behavior.

CCTableView is not optimal and certainly not bulletproof, but it is functional. The CCTableView still exhibits improper behavior with respect to touches because the touch is not tested against the boundaries of each cell. Instead, an "optimization" is used that can result in a cell-touch even though the cell is not being touched. This occurs when touching outside the cell boundaries but still within the ScrollView region. I'll be refactoring the touch behavior at a later time (I don't have the bandwidth right now).

Regarding animations and object pooling: I think it's reasonable to expect animations to stop when the cell is recycled and it's the developers responsibility to manage the animations when the cell is reused. Any animations that are expected to live beyond the life of a recycled cell should not be attached to the cell, they should be attached to a less volatile part of the scene graph. Something that could be beneficial is to add a delegate method that informs the implementer that the cell is about to be recycled. The implementer may then perform any cell management before the cell is detached from the scene graph.

dumganhar:
> CCTableView is not optimal and certainly not bulletproof, but it is functional. The CCTableView still exhibits improper behavior with respect to touches because the touch is not tested against the boundaries of each cell. Instead, an "optimization" is used that can result in a cell-touch even though the cell is not being touched. This occurs when touching outside the cell boundaries but still within the ScrollView region. I'll be refactoring the touch behavior at a later time (I don't have the bandwidth right now).

Yes, the last cell should not get response if touching outside the cell boundaries.

> Regarding animations and object pooling: I think it's reasonable to expect animations to stop when the cell is recycled and it's the developers responsibility to manage the animations when the cell is reused. Any animations that are expected to live beyond the life of a recycled cell should not be attached to the cell, they should be attached to a less volatile part of the scene graph. Something that could be beneficial is to add a delegate method that informs the implementer that the cell is about to be recycled. The implementer may then perform any cell management before the cell is detached from the scene graph.

Make sense. 

BTW, I noticed that your game _Hide Sheep_ also uses a list( _in the Animal Store_ ) like CCTableView, right?
It looks great, could you make a second demo about how to use CCTableView like that when you have time? I think lots of developers need it. Thanks. :)

dumganhar:
Merged, thanks.

edbartley:
I used a custom ScrollView for the Animal Store in Hide the Sheep but it can easily be switched to the new CCTableView. I can make a demo but can't commit to a completion date. I'll work on it as soon as I can.

dumganhar:
> I used a custom ScrollView for the Animal Store in Hide the Sheep but it can easily be switched to the new CCTableView. I can make a demo but can't commit to a completion date. I'll work on it as soon as I can.

Ok, Thanks. :)

PR #1730, opened 2012-12-06
Some cosmetics

fape:
- merge duplicated line for 'fape' in AUTHORS
-  include all resources from Resource dir to SampleGame marmalade project file


dumganhar:
Thanks. :)

PR #1732, opened 2012-12-06
Add randomObject to CCDictionary

tks2shimizu:
Although CCArray has randomObject() method, CCDictionary does not have it.
So I added randomObject() method to CCDictionary.


dumganhar:
Thanks.

PR #1740, opened 2012-12-09
[android] Cocos2dxHelper/CCFileUtils JNI fixes and improvements

rtissera:
- Added proper JNIEXPORT / JNICALL modifiers to Cocos2dxHelper
- Rewrote CCFileUtils::getWriteabblePath() to get cache directory through Context.getCacheDir() Java method and wrote appropriate JNI bridge. This does fix a critical issue on Nexus 10 (CCUserDefault not working).


minggo:
Thank you.

moadib:
I have note. Cache directory may be cleaned at any moment by system. 

That mean that CCUserDefaults can be lost at any moment(if there is low space, then Android can delete files from apps cache dirs).

PR #1744, opened 2012-12-10
[scripting] add CCNode::scheduleUpdateWithPriorityLua()

dualface:
add CCNode::scheduleUpdate() script support.

example:

```
local function onEnterFrame(dt)
    print(dt)
end

node:scheduleUpdateWithPriorityLua(onEnterFrame, 0)
```

advantage:
-   CCScheduler::scheduleScriptFunc() registered global event, when the exit of the scene must manually clean up. CCNode::scheduleUpdateWithPriorityLua() registered local event, will automatically clean up the node remove from parent.
-  better performance.

### DIFF
--- a/cocos2dx/CCScheduler.cpp
+++ b/cocos2dx/CCScheduler.cpp
@@ -150,7 +150,6 @@ void CCTimer::update(float dt)
     }
     else
     {
-        CCScriptEngineProtocol *pEngine = CCScriptEngineManager::sharedManager()->getScriptEngine();
         if (m_bRunForever && !m_bUseDelay)
         {//standard timer usage
             m_fElapsed += dt;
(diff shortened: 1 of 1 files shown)


cocos2dx/base_nodes/CCNode.h
```diff
@@ -720,7 +722,10 @@ class CC_DLL CCNode : public CCObject
      @since v0.7.1
      */
     CCPoint convertTouchToNodeSpaceAR(CCTouch * touch);
-    
+        
+    /** Schedules for script. */
+    void scheduleUpdateScriptHandlerWithPriority(int nHandler, int priority);
```
minggo: How about rename to scheduleUpdateWithPriorityLUA().

PR #1757, opened 2012-12-13
Fix crash for call runAnimation() in CCBAnimationManagerDelegate::completedAnimationSequenceNamed

tiantian20007:
Fix crash for call runAnimation() in CCBAnimationManagerDelegate::completedAnimationSequenceNamed; Set mRunningSequence to NULL before any runAnimation() call in this delegate method;

### DIFF
--- a/extensions/CCBReader/CCBAnimationManager.cpp
+++ b/extensions/CCBReader/CCBAnimationManager.cpp
@@ -665,22 +665,24 @@ void CCBAnimationManager::setAnimationCompletedCallback(CCObject *target, SEL_Ca
 
 void CCBAnimationManager::sequenceCompleted()
 {
+    const char *runningSequenceName = mRunningSequence->getName();
+    int nextSeqId = mRunningSequence->getChainedSequenceId();
+    mRunningSequence = NULL;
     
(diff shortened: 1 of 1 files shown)


extensions/CCBReader/CCBAnimationManager.cpp
```diff
@@ -665,23 +665,25 @@ void CCBAnimationManager::setAnimationCompletedCallback(CCObject *target, SEL_Ca
...
+    if(lastCompletedSequenceName != runningSequenceName) {
+        lastCompletedSequenceName = runningSequenceName;
     }
     
     if (mDelegate)
     {
-        mDelegate->completedAnimationSequenceNamed(mRunningSequence->getName());
+        // There may be another runAnimation() call in this delegate method
+        // which will assign mRunningSequence
+        mDelegate->completedAnimationSequenceNamed(runningSequenceName);
```
dumganhar, line 680: Yep, it looks good.
But a demo to reproduce this issue is better. Therefore i could make sure this change can work.
Could you email the demo to me?

dumganhar:
Thank you, merged.

PR #1764, opened 2012-12-18
fixed #1617: Some improvements for JS Bindings.

dumganhar:
- **1) Changed <code>cc.REPEAT_FOREVER = - 1</code> to <code>cc.REPEAT_FOREVER = 0xffffffff</code>**
  
  **[Reason]**: If cc.REPEAT_FOREVER = -1, it will be a very big double value after converting it to double by JS_ValueToNumber on android.
  Then cast it to unsigned int, the value will be 0. The schedule will not be able to work.
  I don't know why this occurs only on android.
  
  **[Solution]**: Instead of passing -1 to it, I assign it with max value of unsigned int in c++.
- **2) Added two helper function, cc.ArrayGetIndexOfObject and cc.ArrayContainsObject.**
- **3) Added JSScheduleWrapper::removeTargetForNativeNode to avoid memory leaks.**
- **4) Improvments for JSTouchDelegate. Added four functions as follows:**

<pre>
// Set the touch delegate to map by using the key (pJSObj).
static void setDelegateForJSObject(JSObject* pJSObj, JSTouchDelegate* pDelegate);

// Get the touch delegate by the key (pJSObj).
static JSTouchDelegate* getDelegateForJSObject(JSObject* pJSObj);

// Remove the delegate by the key (pJSObj).
static void removeDelegateForJSObject(JSObject* pJSObj);

void unregisterTouchDelegate();
</pre>

And exported cc.unregisterTouchDelegate(); to js.
Fix a memory leak for JSTouchDelegate by making it as an autorelease object.
- **5) Don't add js callback function to the reserved slot of object.**
  [Reason]: The target object may execute more than one schedule.
  Therefore, previous js callback function will be replaced
  by the current one. For example:
  
  <pre>
  this.scheduleOnce(function() { temporary function 1 }, 0.5);
  this.scheduleOnce(function() { temporary function 2 }, 0.5);
  </pre>
  
  In this case, the temporary function 1 will be removed from reserved slot 0.
  And temporary function 2 will be set to reserved slot 0 of this object.
  If gc is triggered before the JSScheduleWrapper::scheduleFunc is invoked,
  crash will happen. You could simply reproduce it by adding <code>**jsc**.garbageCollect()</code>; after scheduleOnce.
  
  **[Solution]** Because one schedule corresponds to one JSScheduleWrapper, we root
  the js callback function in JSScheduleWrapper::setJSCallbackFunc and unroot it
  at the destructor of JSScheduleWrapper.

### DIFF
--- a/scripting/javascript/bindings/ScriptingCore.cpp
+++ b/scripting/javascript/bindings/ScriptingCore.cpp
@@ -114,28 +114,18 @@ static void unRootObject(JSContext *cx, JSObject *obj) {
 }
 
 static void getJSTouchObject(JSContext *cx, CCTouch *x, jsval &jsret) {
-    js_type_class_t *classType;
-    TypeTest<cocos2d::CCTouch> t;
-    uint32_t typeId = t.s_id();
-    HASH_FIND_INT(_js_global_type_ht, &typeId, classType);
-    assert(classType);
-    JSObject *_tmp = JS_NewObject(cx, classType->jsclass, classType->proto, classType->parentProto);
-    js_proxy_t *proxy, *nproxy;
-    JS_NEW_PROXY(proxy, x, _tmp);
-    void *ptr = x;
-    JS_GET_PROXY(nproxy, ptr);
-    JS_AddNamedObjectRoot(cx, &nproxy->obj, "CCTouch");
-    jsret = OBJECT_TO_JSVAL(_tmp);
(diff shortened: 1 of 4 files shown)


rohankuruvilla:
Hi James, looks good! Thanks a lot!

dumganhar:
So could it be merged?

rohankuruvilla:
Yes, no reason not to. Thanks!

PR #1766, opened 2012-12-19
Enable Windows XP support in Visual Studio 2012 Update 1

ivfit:


dumganhar:
Hi, as i know, vs2012 can only run on Win8.
So did it support xp, win7 now ?

ghost:
Vs2012 can run on Windows 7.
But it don't support metro ui.
I used vs2010 run on xp,
but it can't create a new cocos2d-x project.

ivfit:
Visual Studio 2012 can be installed on Windows 7 and 8:
http://www.microsoft.com/visualstudio/eng/products/compatibility

Visual Studio 2012 Update 1 can be used to target Windows XP, 7 and 8:
http://www.microsoft.com/visualstudio/eng/visual-studio-update#story-update-1
"But Update 1 isn’t just about new Windows platforms. It also enables you to target Windows XP with native C++ applications in Visual Studio 2012."

CocosDenshion/proj.win32/CocosDenshion.vcxproj
```diff
@@ -20,10 +20,12 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <CharacterSet>NotSet</CharacterSet>
+    <PlatformToolset>v110_xp</PlatformToolset>
```
dumganhar: If adding <code>PlatformToolSet</code> to v110_xp, did it work on vs2010?

ivfit: I've added now condition to set PlatformToolSet to v110_xp for Visual Studio 2012 only.

Should work on vs2010 (not tested).

dumganhar:
Looks great. I will test it. Thanks.

CocosDenshion/proj.win32/CocosDenshion.vcxproj
```diff
@@ -20,10 +20,12 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <CharacterSet>NotSet</CharacterSet>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' == '11.0'">v110_xp</PlatformToolset>
```
dumganhar, line 23: I can't make it compiled successfully if using v110_xp. Instead, it works on both vs2010 and vs2012 if i changing v110_xp to v110.
Any ideas ?

ivfit, line 23: We need exactly "v110_xp" for VS2012. What exactly does VS2010 report as an error when compiling?

dumganhar, line 23: It reported that it couldn't find 'v110_xp',
Is "v110_xp" only valid on windows xp?

ivfit, line 23: "v110_xp" is available when you install Update 1 of Visual Studio 2012.

It is pretty strange that VS2010 ignores Condition="'$(VisualStudioVersion)' == '11.0'".

Please try adding another line targeting VS2010 into project file and see if it works this way:
(add brackets at the begin and end)

PlatformToolset Condition="'$(VisualStudioVersion)' == '10.0'">v110</PlatformToolset

PlatformToolset Condition="'$(VisualStudioVersion)' == '11.0'">v110_xp</PlatformToolset

dumganhar, line 23: OK, I will test it. Thanks.

dumganhar:
@johnangel 
Hi, i installed xp vm and try to install vs2012 ultimate. But the problem is that i couldn't install it on windows xp sp3.
So in which way you are using to install vs2012 on xp sp3?
Microsoft will deprecate xp, therefore did we really need to support it?

dumganhar:
I have read the official article here(http://www.microsoft.com/visualstudio/eng/products/compatibility).
I didn't see that it supports _xp_.

ivfit:
Visual Studio 2012 can be installed on Windows 7 and 8 only:
http://www.microsoft.com/visualstudio/eng/products/compatibility

Visual Studio 2012 Update 1 can be used to compile executables that can work on Windows XP, 7 and 8 platforms:
http://www.microsoft.com/visualstudio/eng/visual-studio-update#story-update-1
"It also enables you to target Windows XP with native C++ applications in Visual Studio 2012."

I think we should support XP because it still has 40% market share:
http://thenextweb.com/microsoft/2012/12/01/the-big-fight-begins-windows-xp-drops-below-40-market-share-while-windows-8-passes-1/

I've fixed VS2012 to target Windows XP, 7 and 8 platform using "v110_xp" platform toolset. The only thing left is to fix VS2010 to compile properly.

Please try adding another line targeting VS2010 into project file and see if it works this way:
(add brackets at the begin and end)

PlatformToolset Condition="'$(VisualStudioVersion)' == '10.0'">v110</PlatformToolset

PlatformToolset Condition="'$(VisualStudioVersion)' == '11.0'">v110_xp</PlatformToolset

dumganhar:
Agree, so the toolset _v110_xp_ is to make the application that generated by vs2012 to run on windows xp.
I got it. We need to support it of course. 
I will test it tomorrow by using the two lines you added.
Thanks in advance.

ivfit:
Right. So the idea is to force VS2012 compiler to use v110_xp toolset and to force VS2010 to use it's default vs110 toolset.

ivfit:
Just commited this change, please check now using VS2010. Thanks.

PR #1770, opened 2012-12-21
issue #1536:add permissions and modify module path error

minggo:


deemen:
Hi, I use Cocos2d-x with Marmalade and one of these commits broke our MIPS build. Specifically this https://github.com/dumganhar/cocos2d-x/commit/49ad7cc99101692b5201b3c989e91f5388eda7c9

The compiler complains thats cpu-features.h is missing at #include <cpu-features.h> in dsp/cpu.c

PR #1775, opened 2012-12-21
CocosDenshion Android - missing sound effects with some conditions

mingulov:
Android SoundPool.load returns 0 if sound might not be loaded - for example, it does not exist. Previously - 0 has been ignored and incorrectly written to HashMap, so it was not possible to understand on the application side that something goes wrong (and use some additional workaround).

Use case:
1) On the game loading - preload all possible sound effects (to speed up their playing later, when it will be needed)
2) Game APK does not have any sound file, they are downloaded later in a separate thread.
3) Application tries to play sound effects as usual, when needed.

So due to this bug - as there was no any sound file exist during the game loading - internal HashMap contains incorrect 0 as id for every effect, and there is no any sound effect during
the first application start.

Fixed by adding additional check to Cocos2dxSound.java - to preloadEffect and createSoundIDFromAsset.

### DIFF
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxSound.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxSound.java
@@ -108,7 +108,10 @@ public class Cocos2dxSound {
 
 		if (soundID == null) {
 			soundID = this.createSoundIDFromAsset(pPath);
-			this.mPathSoundIDMap.put(pPath, soundID);
+			// save value just in case if file is really loaded
+			if (soundID != Cocos2dxSound.INVALID_SOUND_ID) {
+				this.mPathSoundIDMap.put(pPath, soundID);
+			}
 		}
 
 		return soundID;
@@ -270,6 +273,11 @@ public class Cocos2dxSound {
 		} catch (final Exception e) {
 			soundID = Cocos2dxSound.INVALID_SOUND_ID;
 			Log.e(Cocos2dxSound.TAG, "error: " + e.getMessage(), e);
+		}
+
+		// mSoundPool.load returns 0 if something goes wrong, for example a file does not exist
+		if (soundID == 0) {
+			soundID = Cocos2dxSound.INVALID_SOUND_ID;
 		}
 
 		return soundID;


cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxSound.java
```diff
@@ -272,6 +275,11 @@ public int createSoundIDFromAsset(final String pPath) {
 			Log.e(Cocos2dxSound.TAG, "error: " + e.getMessage(), e);
 		}
 
+		// mSoundPool.load returns 0 if something goes wrong, for example a file does not exist
```
minggo, line 278: Does "soundID = this.mSoundPool.load(pPath, 0);" return 0 when a file does not exist?
Because when a file does not exist, "soundID = this.mSoundPool.load(this.mContext.getAssets().openFd(pPath), 0);" will trigger an exception.

mingulov, line 278: Yes, SoundPool.load (in both cases - with file name or assets file description) does not throw an exception but returns 0 instead of.

At the line "soundID = this.mSoundPool.load(this.mContext.getAssets().openFd(pPath), 0);" exception is thrown by _openFd_ [call](http://developer.android.com/reference/android/content/res/AssetManager.html#openFd%28java.lang.String%29) :

E/Cocos2dxSound(9355): java.io.FileNotFoundException: zzzz.wav
E/Cocos2dxSound(9355):  at android.content.res.AssetManager.openAssetFd(Native Method)
E/Cocos2dxSound(9355):  at android.content.res.AssetManager.openFd(AssetManager.java:341)
E/Cocos2dxSound(9355):  at org.cocos2dx.lib.Cocos2dxSound.createSoundIDFromAsset(Cocos2dxSound.java:273)
E/Cocos2dxSound(9355):  at org.cocos2dx.lib.Cocos2dxSound.preloadEffect(Cocos2dxSound.java:110)
E/Cocos2dxSound(9355):  at org.cocos2dx.lib.Cocos2dxHelper.preloadEffect(Cocos2dxHelper.java:165)
...

If somehow value returned by openFd a moment later will become unusable due to any reason - SoundPool returns 0 on such value, like for example if add the next check at the end of createSoundIDFromAsset:

```
        try {
            android.content.res.AssetFileDescriptor fd = null;
            soundID = this.mSoundPool.load(fd, 0);
        } catch (final Exception e) {
            soundID = Cocos2dxSound.INVALID_SOUND_ID;
            Log.e(Cocos2dxSound.TAG, "error: " + e.getMessage(), e);
        }
        Log.d(Cocos2dxSound.TAG, "id 2: " + soundID);
```

There will no be any info about an exception as it was not thrown, and just a line that 0 as ID is returned:
D/Cocos2dxSound(9355): id 2: 0

PR #1777, opened 2012-12-22
repair linux build and add 64-bit configurations for all linux eclipse projects

ivzave:
didn't test 32-bit builds but should work fine

### DIFF
--- a/CocosDenshion/proj.linux/.cproject
+++ b/CocosDenshion/proj.linux/.cproject
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<?fileVersion 4.0.0?>
-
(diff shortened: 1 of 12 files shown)


dumganhar:
Makefile can work for 32bit and 64bit, but eclipse projects can't work for both 32 and 64bit.
Any idea about that?
Thanks.

ivzave:
I've added 2 separate configurations for 64-bit - Debug64 and Release64. This should work, am I wrong?

dumganhar:
Got it,i will test it.Thanks.
ÔÚ 2012-12-22 ÏÂÎç9:31£¬"Igor Zavorotkin" notifications@github.comÐ´µÀ£º

dumganhar:
Thank you.
It's merged at https://github.com/cocos2d/cocos2d-x/pull/1779.

PR #1778, opened 2012-12-22
Compilation fix for Android clang 3.1 toolchain

mingulov:
Android NDK r8c supports clang3.1 as a toolchain.
It is impossible to compile CocosDenshion current version by that clang3.1 due to the "error: expected expression".


minggo:
Thank you.

PR #1783, opened 2012-12-24
marmalade acceleromater improvement

fape:
Use CCDirector->getAccelerometer() instead of CCAccelerometer::sharedAccelerometer(), because sharedAccelerometer() does NOT contains m_pAccelDelegate "callback" instance (m_pAccelDelegate allways NULL)


dumganhar:
Merged, thanks.

PR #1792, opened 2012-12-26
[luabinding] make CCEditBox compatible Lua

dualface:
HOW TO USE:

```
local editbox = CCEditBox:create(...)
editbox:registerScriptEditboxHandler(function(event, control)
    if event == "ended" then
        CCLuaLog(string.format("username = %s", control:getText()))
    end
end)
```

### DIFF
--- a/extensions/GUI/CCEditBox/EditBoxImplIOS.mm
+++ b/extensions/GUI/CCEditBox/EditBoxImplIOS.mm
@@ -67,7 +67,7 @@
         self.textField = [[[CustomUITextField alloc] initWithFrame: frameRect] autorelease];
         if (!textField_) break;
         [textField_ setTextColor:[UIColor whiteColor]];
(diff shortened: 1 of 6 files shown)


extensions/GUI/CCEditBox/EditBoxImplIOS.mm
```diff
@@ -66,9 +67,10 @@ -(id) initWithFrame: (CGRect) frameRect editBox: (void*) editBox
         self.textField = [[[CustomUITextField alloc] initWithFrame: frameRect] autorelease];
         if (!textField_) break;
         [textField_ setTextColor:[UIColor whiteColor]];
-        textField_.font = [UIFont systemFontOfSize:frameRect.size.height*2/3]; //TODO need to delete hard code here.
+        textField_.font = [UIFont systemFontOfSize:(frameRect.size.height - 16)]; //TODO need to delete hard code here.
```
dumganhar, line 70: Why to subtract a const value 16? On ipad, the letter 'g' will out of bound.

PR #1799, opened 2012-12-27
Added CCFloat, CCDouble and CCBool

tks2shimizu:
CCArray and CCDictionary are very convenient classes. 
If float, double and bool can also be used to them, it will be still easier-to-use. 
Therefore, I added CCFloat, CCDouble and CCBool classes. 

Usage is the same as CCInteger class.

### DIFF
--- a/cocos2dx/cocoa/CCBool.h
+++ b/cocos2dx/cocoa/CCBool.h
@@ -1,3 +1,27 @@
+/****************************************************************************
+ Copyright (c) 2010-2012 cocos2d-x.org
+ 
+ http://www.cocos2d-x.org
+ 
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+ 
+ The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+ 
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ ****************************************************************************/
+
 #ifndef __CCBOOL_H__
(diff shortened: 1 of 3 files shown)


cocos2dx/cocoa/CCBool.h
```diff
@@ -0,0 +1,35 @@
...
+    bool getValue() const {return m_nValue;}
+
+    static CCBool* create(bool v)
+    {
+        CCBool* pRet = new CCBool(v);
+        pRet->autorelease();
+        return pRet;
+    }
+private:
+    double m_nValue;
```
minggo: Why not use bool?

cocos2dx/cocoa/CCDouble.h
```diff
@@ -0,0 +1,35 @@
...
+    double getValue() const {return m_nValue;}
+
+    static CCDouble* create(double v)
+    {
+        CCDouble* pRet = new CCDouble(v);
+        pRet->autorelease();
+        return pRet;
+    }
+private:
+    double m_nValue;
```
minggo: I think it is better to rename m_nValue to m_dValue.
Because its type is double.

cocos2dx/cocoa/CCDouble.h
```diff
@@ -0,0 +1,35 @@
...
+{
+public:
+    CCDouble(double v)
+        : m_nValue(v) {}
+    double getValue() const {return m_nValue;}
+
+    static CCDouble* create(double v)
+    {
+        CCDouble* pRet = new CCDouble(v);
+        pRet->autorelease();
```
minggo: How about checking pRet before using it?

cocos2dx/cocoa/CCFloat.h
```diff
@@ -0,0 +1,35 @@
...
+    float getValue() const {return m_nValue;}
+
+    static CCFloat* create(float v)
+    {
+        CCFloat* pRet = new CCFloat(v);
+        pRet->autorelease();
+        return pRet;
+    }
+private:
+    float m_nValue;
```
minggo: m_nValue -> m_fValue?

cocos2dx/cocoa/CCFloat.h
```diff
@@ -0,0 +1,35 @@
...
+{
+public:
+    CCFloat(float v)
+        : m_nValue(v) {}
+    float getValue() const {return m_nValue;}
+
+    static CCFloat* create(float v)
+    {
+        CCFloat* pRet = new CCFloat(v);
+        pRet->autorelease();
```
minggo: How about validating pRet?

tks2shimizu:
Thank you for checking.

I changed as follows.
・The variable name was changed.
・The check of pRet was added.

Once again, please check.

minggo:
Thank you, i will test it.

farwayer:
Thx a lot! I used CCBValue from Cocos Builder before, but native support is more right. And it would be great to see CCPointer for void\* also.

minggo:
@tks2shimizu 
Would you mind add licences for the cpp/h files?

tks2shimizu:
sorry...
The license was added.

tks2shimizu:
Aren't CCFloat, CCDouble and CCBool used?

winipcfg:
I would rather create a template class for that so I could create CCFloat/CCDouble/CCBool/etc
class CCFloat : public CCObjectT< float > 

But that's just my coding style anyway

tks2shimizu:
@minggo 
Thanks, merged.

@winipcfg 
Thank you for the advice.

minggo:
@winipcfg 
Thank you for your suggestion.
But i think we should use as simple c++ feature as we can.
Template is hard to debug, and it is hard to be binded to other language.

PR #1800, opened 2012-12-28
CCTableView update, CCMenu bug fix

edbartley:
Added an optional method to the CCTableViewDelegate that informs the implementer when a cell in the CCTableView will be recycled.

Fixed a bug in CCMenu that prevented the CCMenu touch priority from being updated by a call to CCMenu::setTouchPriority(int). The CCMenu touch priority was set to kCCMenuHandlerPriority.

### DIFF
--- a/cocos2dx/menu_nodes/CCMenu.cpp
+++ b/cocos2dx/menu_nodes/CCMenu.cpp
@@ -198,7 +198,7 @@ void CCMenu::setHandlerPriority(int newPriority)
 void CCMenu::registerWithTouchDispatcher()
 {
     CCDirector* pDirector = CCDirector::sharedDirector();
-    pDirector->getTouchDispatcher()->addTargetedDelegate(this, kCCMenuHandlerPriority, true);
+    pDirector->getTouchDispatcher()->addTargetedDelegate(this, CCLayer::getTouchPriority(), true);
 }
 
 bool CCMenu::ccTouchBegan(CCTouch* touch, CCEvent* event)


edbartley:
Reference this conversation with Dumganhar: https://github.com/cocos2d/cocos2d-x/pull/1724#issuecomment-11046801

extensions/GUI/CCScrollView/CCTableView.h
```diff
@@ -119,19 +130,21 @@ class CCTableView : public CCScrollView, public CCScrollViewDelegate
...
      * An initialized table view object
      *
      * @param dataSource data source;
+     * @param viewDelegate view delegate
      * @param size view size
      * @param container parent object for cells
      * @return table view
      */
-    static CCTableView* create(CCTableViewDataSource* dataSource, CCSize size, CCNode *container);
+    static CCTableView* create(CCTableViewDataSource* dataSource, CCTableViewDelegate* viewDelegate, CCSize size, CCNode *container);
```
minggo, line 147: I think it is better to add a new function to keep compatibility.

edbartley, line 147: The delegate is already required by CCTableView in order for CCTableView to function without crashing. Adding the delegate as a required parameter to the existing constructor formalizes the requirement. Additionally, updating ones code to accommodate the new required parameter is very simple. 

I recommend either requiring the delegate parameter in the constructor or refactoring CCTableView so that it does not crash if a delegate is not provided by the implementer. 

What are your thoughts?

AdyFly:
Hello I've noticed a strange behavior on CCTableView class.
When using the bounceable feature, the table can be scrolled too much ( meaning that the cells are no longer visible ).
I think it would be nice if we can somehow restrict the scroll ( similar to settings menu on IOS ).

Thanks for your nice work!
Adrian

extensions/GUI/CCScrollView/CCTableView.h
dumganhar, line 147: > Additionally, updating ones code to accommodate the new required parameter is very simple.

Hi @edbartley ,
I agree with you. But lots of developers blamed us for the changes of `API`.
I don't know whether they will agree this. :(

dumganhar:
Closed it since we renamed gles20 branch to master.
Please discuss in https://github.com/cocos2d/cocos2d-x/pull/1851. Thanks.

PR #1806, opened 2012-12-29
Marmalade third party cleanup

fape:
- Libjpeg, libpng and zlib removed because these provided by Marmalade platform/SDK.
- Move libtiff project file to root of marmalade third party directory (same as libxml2 and freetype)

### DIFF
--- a/cocos2dx/platform/third_party/marmalade/libtiff.mkf
+++ b/cocos2dx/platform/third_party/marmalade/libtiff.mkf
@@ -1,9 +1,12 @@
-includepath headers
+includepath 
+{
+	libtiff/headers
+}
 
 files
 {
-	[Source]
-	(source)
+	[source]
+	(libtiff/source)
 	tif_aux.c
     tif_close.c
     tif_codec.c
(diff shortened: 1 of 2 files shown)


dumganhar:
Thanks, merged.
@fape 
But there are two issues i found in marmalade port. Could you help us to fix them?
- In SampleGame project, the window is Portrait, but it need to be Landscape.
- _TestLua_ can't work.

Any thoughts ?

fape:
Thanks!
- SampleGame: Try change surface settings in simulator [ Configuration/ Surface... / Predefined Resolution to eg. 1024 X 600 (iphone  5 Landscape) ]. As i see there are some .gitignore in wrong directory (HelloCpp and SampleCpp) so development.icf and app.icf are ignored. I'm going to commit the fix in few hours.
- TestLua: The simulator option s3e-data-dir can't work with multiple directroy. There is an undocumented postbuild option where eg an python script can copy resouces (like in eg android makefile). I'm going to try. If it works I'll commit the fix today.

PR #1814, opened 2013-01-03
Marmalade SimpleGame force landscape 

fape:
- move .gitignore
- force landscape mode
- add marmalade config files (with simulator settings)


dumganhar:
Thanks. :)

PR #1815, opened 2013-01-03
Marmalade TestLua postbuild script

fape:
- add marmalade config files
- add postbuild copy resources script


dumganhar:
Thanks.

PR #1817, opened 2013-01-04
Added EditBox implementation for Mac OSX.

jozef-pridavok:
Added the very first version of EditBox for Mac.


dumganhar:
Hi @jpridavok ,thanks for adding editbox for MAC port, but did you notice that the position of system control was wrong.
In _TestCpp/ExtenstionsTest/EditboxTest_, there are three editbox controls.
They are 

<pre>
Name:
PassWord:
Email:
</pre>

But your implementation for mac, the order is reversed.
It becomes

<pre>
Email:
Password:
Name:
</pre>

Any thoughts about that?

minggo:
@jpridavok 
The file names are:

CCEditBoxImplMac.h
CCEditBoxImplMac.mm
EditBoxImplMac.h
EditBoxImplMac.mm

These file names are not so clearly. Why not just add two files CCEditBoxImplMac.h and CCEditBoxImplMac.mm?

dumganhar:
@jpridavok 
I think @minggo is right.
I have moved the implementation of <code>EditBoxImplIOS._</code> to <code>CCEditBoxImplIOS._</code>  here(https://github.com/cocos2d/cocos2d-x/pull/1819).

So could you also do that for MAC port after you fixe the issue i said ?
Thanks a lot.

jozef-pridavok:
Hi,

@dumganhar: i've tested it in my project (with just one control), probably i missed wrong position. I'm going to check it.

@minggo: There is no problem with refactoring of implementation into one pair of file. I tried to keep the same convention as iOS.

minggo:
@jpridavok 
Thank you.
Please let me know if it is ready for testing.

jozef-pridavok:
Hello @dumganhar, @minggo,

I've fixed wrong position of the editbox controls on Mac, and i've refactored source code to use just one implementation file.

Jozef

PS: I am working on Win32 version, but it seems that i have to use popup input box (similar to android solution) because of side effects when placing native win32 control to opengl context (ugly flickering)... Do you think it's okay for the very first implementation?

dumganhar:
@jpridavok 
Good work. 

> I am working on Win32 version, but it seems that i have to use popup input box (similar to android solution) because of side effects when placing native win32 control to opengl context (ugly flickering)... Do you think it's okay for the very first implementation?

I agree with you. Win32 is mainly for debugging.So we could use inputbox to implement it first. Thanks.

dumganhar:
Thank you. Merged.

PR #1824, opened 2013-01-07
Enable Windows XP support in Visual Studio 2012 Update 1, another try

ivfit:
Setting PlatformToolSet to v110_xp instead of default v110 in Visual Studio 2012 for Windows XP support, leaving default v100 in Visual Studio 2010.


dumganhar:
OK, I check it now.

dumganhar:
I have tried your changes on win8 vs2012(without update1).There are lots of errors.

<pre>
1>C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V110\Platforms\Win32\PlatformToolsets\v110_xp\Microsoft.Cpp.Win32.v110_xp.targets(36,5): warning MSB8003: Could not find WindowsSdkDir_71A variable from the registry.  TargetFrameworkVersion or PlatformToolset may be set to an invalid version number.
</pre>

It's ok on vs2010.
I'm sure that if i update my vs2012 to update1, it will work. But i think lots of developers will not make this update.
Is there a way for making it works both on vs2012 with and without update1?

ivfit:
I'll look into it.

Anyway, wondering why would anybody skip compiler update? It's like you don't update your OS with service packs.

ivfit:
Unfortunately I couldn't find property for detecting is Update installed or not:
http://msdn.microsoft.com/en-us/library/bb629394.aspx

dumganhar:
I think we shouldn't assume that all developers will update their compiler. If he didn't update it, then he will got errors when compiling. Then he will complain that cocos2d-x can't run by vs2012. Though, we can make a document to remind developers to update the compiler, but there may be some developers who don't read release note.
Any thought?

> I'll look into it.
> Anyway, wondering why would anybody skip compiler update? It's like you don't update your OS with service packs.

ivfit:
Maybe to check does %ProgramFiles(x86)%\Microsoft SDKs\Windows\7.1A folder exist?

And instead of using WindowsSdkDir_71A variable to hardcode above path?

Ref: http://blogs.msdn.com/b/vcblog/archive/2012/10/08/windows-xp-targeting-with-c-in-visual-studio-2012.aspx

dumganhar:
So the path in VS2012 update 1 and not update are different?
Another thing, how do you check the system is 32bit or 64bit since the path isn't the same too.
Also, the path can be changed to other folders when installing vs.
Maybe you could try that, if it works, please let me know, i will help to test it. Thanks.

ivfit:
Yes, there are several SDKs in c:\Program Files (x86)\Microsoft SDKs\Windows. I don't think this path can be changed during the installation.

Please confirm you don't have "c:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A", but have "c:\Program Files (x86)\Microsoft SDKs\Windows\v8.0".

We can write in readme.txt that if somebody wants XP support, it should install VS2012 Update 1 on 64bit system.

dumganhar:
On my windows xp3, I installed vs2010, and the sdk path is <code>C:\Program Files\Microsoft SDKs\Windows\v7.0A</code>.

> We can write in readme.txt that if somebody wants XP support, it should install VS2012 Update 1 on 64bit system.

So you mean that we just use <code>v110</code> instead of <code>v110_xp</code>?
And write a _README_ file to describe how to make the application built with vs2012 to run on _windows xp_?

ivfit:
You said it compiles fine on VS2010, thought we are talking just about VS2012 now?

Here is the idea for VS2012: if exists("c:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A") then use PlatformToolSet v110_xp else use v110

README: To be able to run compiled games on Windows XP/Vista/7/8, please install VS2012 Update 1 on 64bit OS.

dumganhar:
> You said it compiles fine on VS2010, thought we are talking just about VS2012 now?

Yes, it's ok on VS2010.

> Here is the idea for VS2012: if exists("c:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A") then use PlatformToolSet v110_xp else use v110

It will be a good way. So when could you finish this?

ivfit:
I've commited changed CocosDenshion.vcxproj, please let me know if it works.

dumganhar:
> I've commited changed CocosDenshion.vcxproj, please let me know if it works.

Thank you, i will test it tomorrow.

dumganhar:
> I've commited changed CocosDenshion.vcxproj, please let me know if it works.

Hi @johnangel 
Yes, CocosDension project works on vs2012 without update1 .
But others still can't works. So could you also modify other projects? Thanks.

ivfit:
Other projects are changed, also.

Should work on both 32-bit and 64-bit VS2012 properly.

dumganhar:
@johnangel 
This pull request can't be automatically merged. So could you pull the latest codes and resolve conflicts?

ivfit:
Resolved the conflicts.

dumganhar:
Thanks. Merged.

PR #1827, opened 2013-01-10
Assert fix

folecr:
Fixes for use of assertion macros

All platforms
- report line number correctly

Android
- Assert sent to log instead of a UI popup

### DIFF
--- a/cocos2dx/include/ccMacros.h
+++ b/cocos2dx/include/ccMacros.h
@@ -37,7 +37,11 @@ THE SOFTWARE.
 #ifndef CCAssert
 #if COCOS2D_DEBUG > 0
 extern void CC_DLL cc_assert_script_compatible(bool cond, const char *msg);
-#define CCAssert(cond, msg)         cc_assert_script_compatible(!!(cond), (msg))
+#define CCAssert(cond, msg)                             \
+    {                                                   \
+        cc_assert_script_compatible(!!(cond), (msg));   \
+        CC_ASSERT(cond);                                \
+    }
 #else
 #define CCAssert(cond, msg)
 #endif
--- a/cocos2dx/script_support/CCScriptSupport.cpp
+++ b/cocos2dx/script_support/CCScriptSupport.cpp
@@ -32,8 +32,6 @@ void CC_DLL cc_assert_script_compatible(bool cond, const char *msg)
     {
         return;
     }
-    
-    CC_ASSERT(cond);
 }
 
 NS_CC_BEGIN


cocos2dx/include/ccMacros.h
```diff
@@ -36,8 +36,7 @@ of this software and associated documentation files (the "Software"), to deal
 
 #ifndef CCAssert
 #if COCOS2D_DEBUG > 0
-extern void CC_DLL cc_assert_script_compatible(bool cond, const char *msg);
-#define CCAssert(cond, msg)         cc_assert_script_compatible(!!(cond), (msg))
+#define CCAssert(cond, msg)         CC_ASSERT(cond)
```
dumganhar: Hi @Surith,
Why did you revert the definition of CCAssert to CC_ASSERT?
cc_assert_script_compatible is needed for lua-binding.

cocos2dx/cocoa/CCGeometry.cpp
```diff
@@ -133,9 +133,6 @@ bool CCSize::equals(const CCSize& target) const
 
 void CCRect::setRect(float x, float y, float width, float height)
 {
-    // Only support that, the width and height > 0
-    CCAssert(width >= 0.0f && height >= 0.0f, "width and height of Rect must not less than 0.");
-
```
dumganhar: Why did you delete this assert of checking the width and height?

folecr:
Using cc_assert_script_compatible() will result in a bug in CCAssert() : the file, function name and line number will allways be reported to be that of cc_assert_script_compatible() instead of the true location of the assert.

cocos2dx/include/ccMacros.h
dumganhar: The implementation of cc_assert_script_compatible will invoke CC_ASSERT.

<pre>
void CC_DLL cc_assert_script_compatible(bool cond, const char *msg)
{
    cocos2d::CCScriptEngineProtocol* pEngine = cocos2d::CCScriptEngineManager::sharedManager()->getScriptEngine();
    if (!cond && pEngine && pEngine->executeAssert(cond, msg))
    {
        return;
    }
    
    CC_ASSERT(cond);
}
</pre>

dumganhar:
So we need to make <code> cc_assert_script_compatible</code> as a macro, right?

cocos2dx/cocoa/CCGeometry.cpp
folecr: In the exact same file setRect() is called with explicit parameters (0.0f, 0.0f, 0.0f, 0.0f)

https://github.com/folecr/cocos2d-x/blob/7d3df00d241cf6fca0c1382f1ce080522a70dbf4/cocos2dx/cocoa/CCGeometry.cpp#L115

folecr:
I'd suggest the opposite : when Lua is being used, redefine CCAssert to 
1) Call CC_ASSERT
2) also, pass the assert to the Lua engine.

dumganhar:
> I'd suggest the opposite : when Lua is being used, redefine CCAssert to 
> 1) Call CC_ASSERT
> 2) also, pass the assert to the Lua engine.

Yep, it's great. But it need to use macro to define script support.
Now, in the engine, we are using variable to check whether need script binding is opened.
But we didn't using marco to distinguish script support is because all projects in solution of win32 will depend on libcocos2dx, and the marco defined in TestJavascript will not affect the libcocos2dx. What i mean is that there are no a _APP_FLAG_ like android. Do you have any idea about that?

dumganhar:
Another thing:

<pre>
void CCRect::setRect(float x, float y, float width, float height)
 {
    // Only support that, the width and height > 0
********  why did you delete this line ? *********
    CCAssert(width >= 0.0f && height >= 0.0f, "width and height of Rect must not less than 0.");
</pre>

folecr:
Please see my above reply 

In the exact same file setRect() is called with explicit parameters (0.0f, 0.0f, 0.0f, 0.0f)

https://github.com/folecr/cocos2d-x/blob/7d3df00d241cf6fca0c1382f1ce080522a70dbf4/cocos2dx/cocoa/CCGeometry.cpp#L115

folecr:
What I mean :
- The function is checking if 0.0f is passed in... however in the _same_ file a valid call with 0.0f as a parameter is being made and is working fine.
- So, it is not a problem to use 0.0f as a parameter.

dumganhar:
About <code>width >= 0.0f && height >= 0.0f</code>,
It means that we could pass 0.0f but not < 0.0f.
0.0f is valid.

folecr:
Whoops my bad! I've removed the incorrect commit and updated the branch.

dumganhar:
> Whoops my bad! I've removed the incorrect commit and updated the branch.

Thanks. 

> I'd suggest the opposite : when Lua is being used, redefine CCAssert to 
> 
> > 1) Call CC_ASSERT
> > 2) also, pass the assert to the Lua engine.
> > Yep, it's great. But it need to use macro to define script support.
> > Now, in the engine, we are using variable to check whether need script binding is opened.
> > But we didn't using marco to distinguish script support is because all projects in solution of win32 will depend on libcocos2dx, and the marco defined in TestJavascript will not affect the libcocos2dx. What i mean is that there are no a APP_FLAG like android. Do you have any idea about that?

What about your opinion?

cocos2dx/platform/android/CCPlatformDefine.h
```diff
@@ -1,16 +1,27 @@
...
+
+#define CC_MESSAGE_PSEUDOASSERT(cond, msg)                          \
+    if (!(cond)) {                                                  \
+        __android_log_print(ANDROID_LOG_ERROR,                      \
+                            "cocos2d-x assert",                     \
+                            "file:%s function:%s line:%d, %s",      \
+                            __FILE__, __FUNCTION__, __LINE__, msg); \
+    }
+
+#define CC_ASSERT(cond) CC_NO_MESSAGE_PSEUDOASSERT(cond)
```
dumganhar, line 24: Do you think we should use the msg param of CCAssert?
Now, the msg param isn't used.
So maybe 

<pre>
#define CCAssert(cond, msg) CC_ASSERT_MSG(cond, msg)
</pre>

is more suitable.
And on android, we define CC_ASSERT_MSG to CC_MESSAGE_PSEUDOASSERT.

folecr, line 24: Yes, of course we should switch it which is why I added that macro.

However CCAssert is defined globally not per platform. And it is defined to
use CC_ASSERT(cond)

When other platforms have been fixed we can easily switch to the correct
definition.

On Jan 9, 2013, at 19:21, James Chen notifications@github.com wrote:

In cocos2dx/platform/android/CCPlatformDefine.h:

> -    if (!(cond)) {                                              \
> -        __android_log_print(ANDROID_LOG_ERROR,                  \
> -                            "cocos2d-x assert",                 \
> -                            "%s function:%s line:%d",           \
> -                            **FILE**, **FUNCTION**, **LINE**);  \
> -    }
>   +
>   +#define CC_MESSAGE_PSEUDOASSERT(cond, msg)                          \
> -    if (!(cond)) {                                                  \
> -        __android_log_print(ANDROID_LOG_ERROR,                      \
> -                            "cocos2d-x assert",                     \
> -                            "file:%s function:%s line:%d, %s",      \
> -                            **FILE**, **FUNCTION**, **LINE**, msg); \
> -    }
>   +
>   +#define CC_ASSERT(cond) CC_NO_MESSAGE_PSEUDOASSERT(cond)

Do you think we should use the msg param of CCAssert?
Now, the msg param isn't used.
So maybe

#define CCAssert(cond, msg) CC_ASSERT_MSG(cond, msg)

is more suitable.
And on android, we define CC_ASSERT_MSG to CC_MESSAGE_PSEUDOASSERT.

—
Reply to this email directly or view it on
GitHubhttps://github.com/cocos2d/cocos2d-x/pull/1827/files#r2600743.

folecr:
I'll update the definition of CCAssert to call the script engine's assert
too. That is easiest for now.

Longer term, to use cocos2dx as a library : maybe provide a template based
mechanism to generate CCApplication and supporting classes as per the
application's requirements at compile time. This should be cleaner design
and higher performance than using macros and runtime checks. We should
discuss this separately.

On Jan 9, 2013, at 19:04, James Chen notifications@github.com wrote:

Whoops my bad! I've removed the incorrect commit and updated the branch.

Thanks.

I'd suggest the opposite : when Lua is being used, redefine CCAssert to

1) Call CC_ASSERT
2) also, pass the assert to the Lua engine.
Yep, it's great. But it need to use macro to define script support.
Now, in the engine, we are using variable to check whether need script
binding is opened.
But we didn't using marco to distinguish script support is because all
projects in solution of win32 will depend on libcocos2dx, and the marco
defined in TestJavascript will not affect the libcocos2dx. What i mean is
that there are no a APP_FLAG like android. Do you have any idea about that?

 What about your opinion?

—
Reply to this email directly or view it on
GitHubhttps://github.com/cocos2d/cocos2d-x/pull/1827#issuecomment-12078708.

folecr:
Updated the commits. Ready for merge.

folecr:
No problem : I changed the order. CCAssert works fine. I don't use Lua so I haven't tested that code path.

dumganhar:
@folecr 
I will test the lua part. Thanks.

dumganhar:
Hi Surith,
You used `__android_log_print` to implement CCAssert.
But it isn't a really assert since it just output the log. And application will not crash.
In other platforms, if we invoke `assert`, app will crash immediately.
Do we also need to make it crash on android by adding `assert` at the end of `__android_log_print`?

folecr:
I've tried the android assert function. It is buggy : on my Galaxy Nexus
with Jelly Bean a SIGSEGV is sent from a thread that is not the one where
assert() was called.

There are bug reports that the Android debugger still does not work
correctly even in Jelly Bean especially with multithreaded apps (I haven't
investigated the behavior on various versions of Android)

My suggestion is to leave it as log print for now. When Android's behavior
under gdb is stabilized we can easily switch to using the android assert
function.

On Jan 13, 2013, at 20:00, James Chen notifications@github.com wrote:

Hi Surith,
You used __android_log_print to implement CCAssert.
But it isn't a really assert since it just output the log. And application
will not crash.
In other platforms, if we invoke assert, app will crash immediately.
Do we also need to make it crash on android by adding assert at the end of
__android_log_print?

—
Reply to this email directly or view it on
GitHubhttps://github.com/cocos2d/cocos2d-x/pull/1827#issuecomment-12206168.

dumganhar:
Merged.
Thanks.

PR #1829, opened 2013-01-10
Added the very first implementation of CCEditBox for Win32.

jozef-pridavok:
Here is the very first version of CCEditBox for Windows. Useful for testing...


extensions/GUI/CCEditBox/CCEditBoxImplWin.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+/*
+
+#include "CCEditBoxImplWin.h"
+#include "CCEditBox.h"
+#include "CCEGLView.h"
+
+NS_CC_EXT_BEGIN
+
+CCEditBoxImpl* __createSystemEditBox(CCEditBox* pEditBox)
```
dumganhar, line 252: Why there are the same functionality in this file? The same as Line36.

dumganhar, line 252: Oh, my fault, i just notice that was commented.

jozef-pridavok:
It's an "in-place" version of CCEditBox, but there are problems when adding native controls to the OpenGL window...

dumganhar:
Hi @jpridavok 
Did you notice that some UTF-8 letters can't be shown in CCEditBox?
I use Chinese to test it. I think other languages also have this problem.

dumganhar:
Thank you, merged.
But could you resolve the issue i mentioned?

jozef-pridavok:
Hello,
yes, sure, I am going to resolve it. (there is just ANSI version, that's easy to add Unicode support)... I'll commit PR...

dumganhar:
OK, Thanks.

PR #1831, opened 2013-01-11
fixed #1653: Handling correctly when parsing int value in JSB.

dumganhar:
- Using `jsval_to_xxx` instead of `JS_ValueToIntxx`.
  We need to check the return value of `JSAPI` such as `JS_ValueToNumber`
  in our own jsval_to_xxx function. It will raise an error in the terminal for debugging more easily.
- Reporting an error when the count of parameters is wrong.
- Setting the return value(`JSVAL_VOID`) for the functions that return void.
  Perhaps it's not needed. But i think this could be more precise。
- Using `JSB_PRECONDITION2` macro to check return value.
  This marco is the same as which defined in cocos2d-iphone jsb.

_I have also sent a Pull Request to cxx-generator here(https://github.com/funkaster/cxx-generator/pull/24)._

@rohankuruvilla @folecr @funkaster
Please review it and feel free to give some suggestions. Thanks.

### DIFF
--- a/scripting/javascript/bindings/js_bindings_ccbreader.cpp
+++ b/scripting/javascript/bindings/js_bindings_ccbreader.cpp
@@ -8,6 +8,7 @@
 
 #include "js_bindings_ccbreader.h"
 #include "ScriptingCore.h"
+#include "js_bindings_config.h"
 
 USING_NS_CC;
 USING_NS_CC_EXT;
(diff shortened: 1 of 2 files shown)


rohankuruvilla:
Hi James,

Yes these look good. Should improve performance a bit. Thanks a lot!

dumganhar:
Thank @Rohan for reviewing. :)

PR #1832, opened 2013-01-11
debugger for js!

rolandoam:
Hey guys,

This PR is for the debugger, it has only been tested on iOS so far, Surith will test it on Android. You need a patched version of Spidermonkey: you will need to apply `https://dl.dropbox.com/u/29043245/jsbindings.patch` on top of Ricardo's version of Spidermonkey in order to fix a bug that triggers when debugging.

Right now, I've only tested the debugger with MoonWarriors, but at least it works. If you need to know how to make it work, just send me an email or start a thread in the js list.

Enjoy!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,9 @@ Developers:
     Rolando Abarca
         Javascript Binding and testjs
 
+    Victor K(fnz)
+        fix memory leak of CCTextureCache::addImageAsync() on iOS
+
     tks2shimizu
         add CCFloat, CCDouble and CCBool
 
@@ -125,7 +128,7 @@ Developers:
     Carlos Sessa
 	implement the accelerometer module for Android port
 
-    JianHua Chen (Dumganhar)
+    James Chen(dumganhar)
         author of Bada port
 
     Erawppa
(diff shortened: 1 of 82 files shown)


minggo:
Could we apply the patch on Ricardo's repo of SpiderMonkey?
It is not convenient to apply it every time.
And i think we should use the same SpiderMonkey codes.

dumganhar:
Hi Rolando,
I have downloaded the patch, but it's just a patch for adding debugger support for cocos2d-iphone.
It's not a patch for SpiderMonkey js engine. I am confused about that.

James

rolandoam:
Sorry! wrong patch link :)
Yes, the idea is to apply the patch on Ricardo's repo, the patch was intended for you to apply it and test it in the mean time.

Here's the link to the right patch:

https://bug822383.bugzilla.mozilla.org/attachment.cgi?id=694110

dumganhar:
Hi @Rolando,
I got an error that lack of `js_bindings_dbg.h`. Seems something you forgot to commit. Right?

dumganhar:
Merged at #1852. Thanks.
BTW, we have renamed gles20 branch to master, and set the default branch to `master` too.

osuka:
Hi guys any guides on how I could use this? I'm trying to figure out how/if I can debug the Javascript code in an iOS build. I can enable it (JSB_ENABLE_DEBUGGING 1) and I see JS_SetDebugMode is invoked, but other than that I can't seem to figure out how to actually connect to the debugger. Any help/pointers?

rolandoam:
Just FYI, my fork of Spidermonkey has the patches fixing the bug

https://github.com/funkaster/Spidermonkey-1

PR #1833, opened 2013-01-12
added setStringWithFormat for CCLabelxxx.

tks2shimizu:
When setting a string to CCLabelxxx, it is convenient if you can use a format. 
Therefore, setStringWithFormat was added. 
(The kMaxStringLen is using the same value as CCString class.)

### DIFF
--- a/cocos2dx/include/CCProtocols.h
+++ b/cocos2dx/include/CCProtocols.h
@@ -111,6 +111,9 @@ public:
     // sets a new label using an string
     virtual void setString(const char *label) = 0;
 
+    // sets a new label using a string with format
+    virtual void setStringWithFormat(const char* format, ...) = 0;
+    
     /** returns the string that is rendered */
     virtual const char* getString(void) = 0;
 };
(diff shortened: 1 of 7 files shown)


dumganhar:
Do we really need this method?
It's easy to achieve that by:

<pre>
ttf->setString(CCString::createWithFormat("%d, %s", 123, "Hello")->getCString());
</pre>

tks2shimizu:
One line will become simple if this is used. 

<pre>
ttf->setStringWithFormat("%d, %s", 123, "Hello");
</pre>

dumganhar:
But you can easily define a macro like that:
`#define ccsf(...) CCString::createWithFormat(__VA_ARGS__)->getCString()`
Then it will be:
`ttf->setString(ccsf("%d, %s", 123, "Hello"));`
What's your opinion?

tks2shimizu:
Oh. I'm sorry.
I don't know the ccsf(...).

dumganhar:
`ccsf` is not defined in cocos2d-x
What i mean is that you can define it in your application.

PR #1834, opened 2013-01-12
keep working tree clean after build (linux)

ivzave:


minggo:
@ivzave 
Why not overwrite existing resources?
If a file's content is changed, then it should overwrite old one.
Right?

ivzave:
@minggo problem is the .icf files are overwritten by `cp`. I don't use marmalade so don't know if TestLua needs separate .icf files or not. If not we should just replace .icf in TestLua by those in TestCpp in repo.

`cp` doesn't support `--exclude`. `sh` executed by `make` doesn't support regexps so `!(*.icf)` pattern should not work. `Find`+`grep` could work or `rsync` (it's simplier). Can we use `rsync`?

fape:
Yeap, TestLua needs own *.icf files. 
I'm also interested in overwrite or not, because now TestLua's marmalade postbuild script ( cccopy.py ) does not overwrite existing files.

dumganhar:
Thank you. It was merged.

PR #1839, opened 2013-01-14
Fix addImageAsync iOS memory leak

fnz:
Bug: http://www.cocos2d-x.org/issues/1087
Thread: http://www.cocos2d-x.org/boards/6/topics/17031


minggo:
Thank you.

PR #1851, opened 2013-01-16
CCTableView update, CCMenu bug fix

cocos2dx-pull-request:
Since we renamed gles20 branch to master.
Please discuss in this Pull Request. Thanks.

Added an optional method to the CCTableViewDelegate that informs the implementer when a cell in the CCTableView will be recycled.

Fixed a bug in CCMenu that prevented the CCMenu touch priority from being updated by a call to CCMenu::setTouchPriority(int). The CCMenu touch priority was set to kCCMenuHandlerPriority.

### DIFF
--- a/cocos2dx/menu_nodes/CCMenu.cpp
+++ b/cocos2dx/menu_nodes/CCMenu.cpp
@@ -198,7 +198,7 @@ void CCMenu::setHandlerPriority(int newPriority)
 void CCMenu::registerWithTouchDispatcher()
 {
     CCDirector* pDirector = CCDirector::sharedDirector();
-    pDirector->getTouchDispatcher()->addTargetedDelegate(this, kCCMenuHandlerPriority, true);
+    pDirector->getTouchDispatcher()->addTargetedDelegate(this, CCLayer::getTouchPriority(), true);
 }
 
 bool CCMenu::ccTouchBegan(CCTouch* touch, CCEvent* event)


edbartley:
@dumganhar RE: continued discussion from https://github.com/cocos2d/cocos2d-x/pull/1800#discussion_r2631252

Being blamed for making a good code design change is a good thing!. Worry about making Cocos2dx world class software.

Here is the problem with the current design:
If a developer implements the CCTableView class without calling CCTableView::setDelegate(...) after CCTableView::create() then the code will crash.

Why?
CCTableView will dereference the m_pTableViewDelegate pointer in numerous locations. If CCTableView::setDelegate(...) is not called by the developer then m_pTableViewDelegate will be NULL causing a crash. With the current design, TableViewDelegate is required by the CCTableView class to function.

Two proper design solutions:
1 - An implementation of the TableViewDelegate should be provided in the ctor of the CCTableView class as seen in the pull request. In most cases the developer will pass a this pointer in the new parameter.

2 - Refactor the CCTableView so that TableViewDelegate is not required by the CCTableView. This solution is less efficient because it will require numerous if...then checks to make sure the m_pTableViewDelegate is not dereferenced if it is NULL.

I'll do the refactoring, if needed. I just want the code to be designed properly.

dumganhar:
Hey @edbartley 
I think we should do that in 2).
Why?
If developers don't set `CCTableViewDelegate`, it should be ok since the `CCTableViewDataSource` was set in `create` function, the tableview has cells and it just doesn't want to notify its status.

Also, CCTableView may be supported by some `UI tools`, Like CCScrollView is supported in `CocosBuilder`. `CCScrollViewLoader` will not take care of the delegate. If developer want to set the delegate, it could be done in the custom class, and get CCScrollView pointer, then set the delegate.

But first of all, it needs to work ok if delegate wasn't set.

What's your opinion?

edbartley:
I'll modify the pull request on Monday.

edbartley:
Since the branch name change I haven't been able to merge, commit, and push properly. I have the changes prepared but git is being a royal git!

dumganhar:
Why?
You should merge by `git pull https://github.com/cocos2d/cocos2d-x.git master`.
We have changed `gles20` branch to `master`

edbartley:
I did that but it won't allow me to push the merge to my Cocos2dx fork on Github. I fixed it but I'll have to submit a new pull request. Please feel free to close this one.

dumganhar:
Close this Pull Request since it has been opened here(https://github.com/cocos2d/cocos2d-x/pull/1886).

PR #1863, opened 2013-01-18
Adding Readme for CocosPlayer

rohankuruvilla:
Re-using this open commit. The following part can be ignored

/*

NOTE: CCFileUtils not implemented on iOS, and will probably break the iOS build.

I am swamped with trying to get CocosPlayer for Android working before the release, so would really appreciate any help I could get with:
1. Porting CCFileUtils changes to iOS. They are almost exactly the same as CCFileUtils in cocos2d-iphone.
2. Testing all the projects with it. 

NOTE: resource directories have been modified to "resources-xsmall", "resources-small", "resources-medium", "resources-large", "resources-xlarge". These changes are being implemented in CocosBuilder right now. Should be done by tomorrow. Till then for testing, you can replace -large, with -ipad.

UPDATE: Added commit to fix the issue of search paths for CocosBuilder resources.

*/

Thanks,

Rohan

### DIFF
--- a/samples/Javascript/CocosPlayer/README.md
+++ b/samples/Javascript/CocosPlayer/README.md
@@ -1,0 +1,30 @@
+# CocosPlayer
+CocosPlayer JS is an Android implementation of CocosPlayer for Javascript based CocosBuilder projects. It is a wireless sync tool that works in conjunction with CocosBuilder. Using this, you can publish your CocosBuilder JS project and run them wirelessly. The system uses bonjour/zeroconf protocol for network discovery and connection.
+
+NOTE: Right now for iOS equivalent, use CocosPlayer JS for cocos2d-iphone.
+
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/CCFileUtils.h
```diff
@@ -91,20 +99,38 @@ class CC_DLL CCFileUtils
     @brief   Get the writeable path
     @return  The path that can write/read file
     */
-    std::string getWriteablePath();
+    static std::string getWriteablePath();
```
dumganhar: Why changing this method to `static`?

dumganhar:
@rohankuruvilla
Do we need to close this PR since we have merge the PR(https://github.com/cocos2d/cocos2d-x/pull/1885)?

Issue #1667, opened 2012-11-28
addImageAsync not worked now


dayongxie:
The bug is caused by the CCScheduler::scheduleSelector()

minggo:
It can work.
You can try TextureCacheTest.

PR #1867, opened 2013-01-18
issue #1667: Adding file name lookup support.

dumganhar:


dumganhar:
Close it since it has been done in https://github.com/cocos2d/cocos2d-x/pull/1885.

PR #1869, opened 2013-01-21
fixed #1639:add CCBEaseInstant

minggo:


lp43:
Dear Minggo
I can't use the newest version of Cocos2d-x because I used LevelHelper that need cocos2dx version under 2.0.4 right now.
I want to use CocosBuilder 2.1 and CCBAnimationManager->runAnimations(specific timeline).
But it has some error and will crash on it.
I checked your team have solved this problem,
https://github.com/cocos2d/cocos2d-x/pull/1757
but I can't figure out what version I can use in my situation.
Thanks!

PR #1872, opened 2013-01-21
Added <uses-feature android:glEsVersion="0x00020000"/> to AndroidManifest

moadib:
Since cocos2d-x uses gles20 we should have this for proper Google Play filtering

Here is note about it http://developer.android.com/reference/android/opengl/GLSurfaceView.html#setEGLContextClientVersion(int)


dumganhar:
Could users download and install gles20 application if without adding this to AndroidManifest.xml?

moadib:
Yes they can. But users with devices which doesn't support gles20 can too and then they would  damn you in reviews because application doesn't work.

It is allows Google Play to filter your application to users who can use it.

dumganhar:
Thank you.

PR #1874, opened 2013-01-21
Fix: CCScale9Sprite_create(capInsets, file)

MarcelBloemendaal:
Fixed calling of wrong initializer in CCScale9Sprite, the create(CCRect
capInsets, const char\* file). It used to call initWithFile(file,
capInsets), but this is the initializer for a file and a frame rect.
The initializer with capInsets has the capInsets rect as the first
argument. This caused unexpected behaviour.

### DIFF
--- a/extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
+++ b/extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
@@ -468,7 +468,7 @@ bool CCScale9Sprite::initWithFile(CCRect capInsets, const char* file)
 CCScale9Sprite* CCScale9Sprite::create(CCRect capInsets, const char* file)
 {
     CCScale9Sprite* pReturn = new CCScale9Sprite();
-    if ( pReturn && pReturn->initWithFile(file, capInsets) )
+    if ( pReturn && pReturn->initWithFile(capInsets, file) )
     {
         pReturn->autorelease();
         return pReturn;


extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
```diff
@@ -468,7 +468,7 @@ bool CCScale9Sprite::initWithFile(CCRect capInsets, const char* file)
 CCScale9Sprite* CCScale9Sprite::create(CCRect capInsets, const char* file)
 {
     CCScale9Sprite* pReturn = new CCScale9Sprite();
-    if ( pReturn && pReturn->initWithFile(file, capInsets) )
+    if ( pReturn && pReturn->initWithFile(capInsets, file) )
```
dumganhar, line 471: `CCScale9Sprite::initWithFile(const char* file, CCRect rect);`
The first parameter is file name, and the second parameter is rectangle.
Why should make this reversal?

MarcelBloemendaal, line 471: Hello James,

Thx for reviewing. Be careful: there are 2 functions initWithFile, one
takes a file and a rect, the other takes a rect and a file. The second uses
the rect as capInsets, not as frame rect. In the create function I
corrected, the rect given is not a frame rect, but a cap insets rect. If
you call initWithFile(const char_, CCRect), the rect will however be used
as a frame rect. The correct function is initWithFile(CCRect, const char_).

Kind Regards,

Marcel Bloemendaal.

2013/1/21 James Chen notifications@github.com

dumganhar, line 471: Oops, who make these two `create` and `initWithFile`.
I think we just need one. We should delete one.
What's your opinion?

MarcelBloemendaal, line 471: Well, I agree they are a bit confusing, but they both are useful. It seems
appropriate to construct a Scale9Sprite with a file and capInsets.
Especially since for some reason setting the cap insets later (by calling
setCapInsets()) doesn't seem to work. (I haven't had the time to look into
that..) Maybe we could rename one initWithFile to something like
'initWithCapInsetsAndFile, but throughout cocos2d-x all create functions
are just called 'create', so there wouldn't be much of a point to that. I
would prefer a class that WORKS to a class that is 'less confusing'. ;)

But feel free to do as you see fit. :)

2013/1/21 James Chen notifications@github.com

dumganhar, line 471: Agree. 
IMO, we should rename `create(CCRect capInsets, const char* file)` to `createWithCapInsetsAndFile`,
`initWithFile(CCRect capInsets, const char* file)` to `initWithCapInsetsAndFile(CCRect capInsets, const char* file)`.
What 's your thoughts?

dumganhar, line 471: And deprecate old methods.

dumganhar:
Merged. And another issue(http://www.cocos2d-x.org/issues/1675) was created.

PR #1875, opened 2013-01-21
keep working tree clean after build (linux)

ivzave:
Previous request (https://github.com/cocos2d/cocos2d-x/issues/1834) was merged but lost (https://github.com/cocos2d/cocos2d-x/commit/5a521601ebd7eb1703abf8f37da5f3e780095373). Just repeating it.


dumganhar:
Thank you.

PR #1883, opened 2013-01-23
fixed #1670: fixed retain interaction. Only CCNode should retain CCAnimationManager, CCAnimationManger shouldn't retain CCNode.

dumganhar:


dumganhar:
Merged firstly. Needs to check the status of this PR(https://github.com/ricardoquesada/CCBReader/pull/1).

PR #1885, opened 2013-01-23
Refactoring CCFileUtils. CrystalCraze works both on iOS and Android now.

dumganhar:
- To test this Pull Request, you need to do is as follows:

`git checkout -b new-branch`
`git pull https://github.com/dumganhar/cocos2d-x.git refactor-ccfileutils`
`git submodule update --init`.
`Enter tools/tojs folder, and execute genbindings.sh.`

**TO BE DONE:**
- Adding file cache support.
- Adding win32 support.
- Probably we need to add the default  strategy of resource structure rather than setting it at AppDelegate.cpp.

@rohankuruvilla

### DIFF
--- a/cocos2dx/platform/android/CCFileUtils.cpp
+++ b/cocos2dx/platform/android/CCFileUtils.cpp
@@ -37,6 +37,8 @@ static CCFileUtils* s_pFileUtils = NULL;
 // record the zip on the resource path
 static ZipFile *s_pZipFile = NULL;
 
+static std::map<std::string, std::string> s_fullPathCache;
+
 CCFileUtils* CCFileUtils::sharedFileUtils()
 {
     if (s_pFileUtils == NULL)
@@ -53,7 +55,6 @@ bool CCFileUtils::init()
 {
     m_pSearchPathArray = new CCArray();
     m_pSearchPathArray->addObject(CCString::create("assets/"));
-    m_pSearchPathArray->addObject(CCString::create(""));
     
     m_pSearchResolutionsOrderArray = new CCArray();
     m_pSearchResolutionsOrderArray->addObject(CCString::create(""));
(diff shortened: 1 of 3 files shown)


cocos2dx/platform/android/CCFileUtils.cpp
```diff
@@ -66,50 +82,122 @@ void CCFileUtils::purgeCachedEntries()
...
+    }
+    path += file_path;
+    path += resourceDirectory;
+    
+    if (path.size() > 0 && path[path.length()-1] != '/')
+    {
+        path += "/";
+    }
+    path += file;
+    ret += path;
```
minggo: ret is equal to path, so can remove it.

cocos2dx/platform/android/CCFileUtils.cpp
```diff
@@ -66,50 +82,122 @@ void CCFileUtils::purgeCachedEntries()
 
 const char* CCFileUtils::fullPathFromRelativePath(const char *pszRelativePath)
 {
-    return pszRelativePath;
+    CCString* pRet = CCString::create("");
+    pRet->m_sString = fullPathForFilename(pszRelativePath);
```
minggo: Why not write it like this

CCString\* pRet = CCString::create(fullPathForFilename(pszRelativePath));

Because this function is frequently used, so i think we should try our best to optimize it.

cocos2dx/platform/android/CCFileUtils.cpp
```diff
@@ -42,17 +42,33 @@ CCFileUtils* CCFileUtils::sharedFileUtils()
...
         s_pZipFile = new ZipFile(resourcePath, "assets/");
     }
     return s_pFileUtils;
 }
 
+bool CCFileUtils::init()
+{
+    m_pSearchPathArray = new CCArray();
+    m_pSearchPathArray->addObject(CCString::create("assets/"));
+    m_pSearchPathArray->addObject(CCString::create(""));
```
minggo: It is not needed.

cocos2dx/platform/android/CCFileUtils.cpp
dumganhar: Make sense.

PR #1886, opened 2013-01-23
CCTableView update. CCMenu bug fix.

edbartley:
Fixed a CCMenu bug that prevented the touch priority from being set to the value specified by a call to CCMenu::setTouchPriority(int).

Added a delegate method to CCTableView to inform the implementor that a cell as about to be recycled. CCTableView no longer crashes if a CCTableViewDelegate is not provided.


dumganhar:
Thank you.

PR #1889, opened 2013-01-24
LibXml2 replaced with Tinyxml2 and Mingw Compiler Support

martell:
Cocos now compiles with the Mingw and Mingw64 compiler on windows.
Platform folder only midified and it should act as normal for all other compilers
You must make your own makefile (I use cmake)

Could you guys merge please?

Now I have added tinyxml2 to both the windows 8 branch and the main gles20 branch to start the merge of the code base :)

Please Test


martell:
I'll separate it out into two commits shortly and I'll do CCUserDefaults also

bakwc:
sory, didn't noticed that you use cmake. Could you please share (or better commit) your cmake build files?

PR #1894, opened 2013-01-25
Initial Import of tinyxml2 to replace libxml2

martell:
More work must be done to support all the different platforms.
I'll start by doing CCUserDefault.cpp and then I'll move onto
windows visual studio files in the next commit and then I'll move
across all the platforms

### DIFF
--- a/cocos2dx/support/CCUserDefault.cpp
+++ b/cocos2dx/support/CCUserDefault.cpp
@@ -40,7 +40,7 @@ NS_CC_BEGIN
  * export xmlNodePtr and other types in "CCUserDefault.h"
  */
 
-static xmlNodePtr getXMLNodeForKey(const char* pKey, xmlNodePtr *rootNode)
+static tinyxml2::XMLElement* getXMLNodeForKey(const char* pKey, tinyxml2::XMLElement** rootNode, tinyxml2::XMLDocument **doc)
 {
     tinyxml2::XMLElement* curNode = NULL;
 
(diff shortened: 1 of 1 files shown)


minggo:
Why replace libxml2 with tinyxml2?
Any benefit?

martell:
http://github.com/leethomason/tinyxml2/issues/52
The major benefit I really see would be smaller memory allocations.

tinyxml2 homepage does have this comparison between tinyxml and tinyxml2.

Advantages of TinyXML-2

The focus of all future dev.
Many fewer memory allocation (1/10th to 1/100th), uses less memory (about 40% of TinyXML-1), and faster (~5x on read).
No STL requirement.
More modern C++, including a proper namespace.
Proper and useful handling of whitespace

If you consider the fact that tinyxml-1 is roughly on par with libxml2 for read speeds then this is good news.

tinyxml2 seems to be designed with mobile devices in mind as lee (the creator of tinyxml) stated when I asked him so I say it would be a very good move for cocos plus the windows 8 branch uses tinyxml so the merge would be easier.
(I've submitted a pull request for the upgrade for that project also;) )

plus its only one source and header file so you could drop it into the source like you guys did with kazmath which would remove a dependecy for an xml parser such as libxml2 which would take some time to compile from source as it is a dom parser also. This would mean a smaller program outputted for each game also (depending on compiler optimizations)

minggo:
@martell
It sounds good.
Is it a stable release version?

martell:
https://github.com/leethomason/tinyxml2/commit/374c25e2a6550af209e51ea33938e3b8a089ec7b

I must update to this.
The project seems really stable to be honest.
He increments the version when ever he has added some new functionality and only when it has been properly tested.

minggo:
So it is not a released version?
Should we waif for a released version?

martell:
It's at version 1.0.11.
Just like cocos he changes some small parts of the api from time to time. (Shouldn't affect us)

I thought 1.0.0 was a release version and the increments were just bug fix releases and minor api tidies.

minggo:
Okay, got it.

minggo:
I think it is better to move tinyxml2 into support/.
What's your opinion?

martell:
That's fine by me you guys know more about the structure of the project than I do

cocos2dx/platform/CCSAXParser.cpp
```diff
@@ -21,15 +22,67 @@
...
+
+	std::vector<const char*> attsVector;
+	for( const tinyxml2::XMLAttribute* attrib = firstAttribute; attrib; attrib = attrib->Next() )
+	{
+		//CCLog("%s", attrib->Name());
+		attsVector.push_back(attrib->Name());
+		//CCLog("%s",attrib->Value());
+		attsVector.push_back(attrib->Value());
+	}
+	attsVector.push_back(nullptr);
```
minggo, line 66: What's nullprt?
It is not defined.

martell:
Sorry thats c++11 code that was in cocos2dx-win8 I replaced it with NULL but forgot to push up.
I'll fix it asap.

Basically nullptr is better than using NULL for pointers because the compiler recognizes, it is different than giving a value of 0. I don't know if that's the best explanation but I'm sure google will help :)

minggo:
But the compiler which dones't support c++11 doesn't recognize it.

martell:
Exactly, I think it was put in there just to test if window rt compilers did support c++11.

We can actually comment out the line.

minggo:
Yep, i will comment out it.

minggo:
I merged codes in this PR https://github.com/cocos2d/cocos2d-x/pull/1959

PR #1902, opened 2013-01-27
fixed error "NDK_MODULE_PATH variable is properly defined" when building on windows

lity:
Fixed such error "Cannot find module with tag 'libjpeg' in import path  
Android NDK: Are you sure your NDK_MODULE_PATH variable is properly defined ? " when building on windows if the project is created by "create-android-project.bat".


dumganhar:
I have tested this issue both on win32 and MAC. But it just works fine. 
I don't know why. :(

minggo:
You should define NDK_MODULE_PATH in build_native.sh.

PR #1930, opened 2013-01-29
Added TMX polygon parsing.

NatWeiss:
This parses a TMX file's polygon node into a CCArray of CCDictionary objects containing a x and y keys.


NatWeiss:
I used autorelease for the CCString objects to keep it the same as the above code for parsing an x and y value for an object. However, wouldn't it be better if they were all using release instead? If so, I can submit another pull request.

minggo:
@NatWeiss
Yep, release is better than autorelease here.

minggo:
@NatWeiss
Would you add TMX polyline parsing too?

NatWeiss:
Yep.

PR #1931, opened 2013-01-30
[Linux] fixed loading truetype fonts from resource directory

Weeds:
We should use CCFileUtils to get the full path to font resources.
Additionally the input font is converted to lowercase for checking if it is
a ttf file. So we can also load .TTF or .tTf files.


dumganhar:
Thanks @Weeds.
I will test it with PR #1932 ASAP.

dumganhar:
Thank you.

PR #1932, opened 2013-01-30
[Linux] fixed horizontal font alignment

Weeds:
kAlignTop and kAlignBottom should center text horizontally
kAlignTopRight and kAlignBottomRight should align text to the right edge


dumganhar:
Thank you.

PR #1936, opened 2013-01-30
Added the very first implementation of CCEditBox for Win32.

dumganhar:


dumganhar:
Merged. Thanks @jpridavok.

PR #1940, opened 2013-01-31
Support for opening URLs using the browser

ivfit:
For Windows, iOS and Android.

### DIFF
--- a/cocos2dx/platform/android/CCApplication.cpp
+++ b/cocos2dx/platform/android/CCApplication.cpp
@@ -124,6 +124,22 @@ ccLanguageType CCApplication::getCurrentLanguage()
     return ret;
 }
 
+void CCApplication::openURL(const char* pszUrl)
+{
+    JniMethodInfo minfo;
+    
+    if(JniHelper::getStaticMethodInfo(minfo,
+                                      "org/cocos2dx/lib/Cocos2dxActivity",
+                                      "openURL",
+                                      "(Ljava/lang/String;)V"))
+    {
+        jstring StringArg1 = minfo.env->NewStringUTF(pszUrl);
+        minfo.env->CallStaticVoidMethod(minfo.classID, minfo.methodID, StringArg1);
+        minfo.env->DeleteLocalRef(StringArg1);
+        minfo.env->DeleteLocalRef(minfo.classID);
+    }
+}
+
 TargetPlatform CCApplication::getTargetPlatform()
 {
     return kTargetAndroid;
--- a/cocos2dx/platform/android/CCApplication.h
+++ b/cocos2dx/platform/android/CCApplication.h
@@ -60,6 +60,8 @@ public:
      @brief Get target platform
      */
     virtual TargetPlatform getTargetPlatform();
+    
+    void openURL(const char* pszUrl);
 
 protected:
     static CCApplication * sm_pSharedApplication;
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
@@ -31,6 +31,8 @@ import android.os.Bundle;
 import android.os.Message;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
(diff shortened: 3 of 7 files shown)


dualface:
see https://github.com/dualface/cocos2d-x-extensions/blob/cocos2dx20/extensions/native/CCNative.h

CCNative::openURL()

minggo:
If an game opens a url using the codes of this PR, can it back to the previous game on iOS?

ivfit:
Using Back button on Android it returns to game.

On iOS and Windows there is no Back button, so cannot return to game directly?

dualface:
If an game need open url in game, should create Web control. etc: On iOS, use UIWebView.

OpenURL() use system browser open url.

ivfit:
openURL function didn't exist at all in cocos2d-x, so it would be nice to have integrated this PR first.

Then you can extend it, so using some parameter developer can request use of web control instead of system browser.

dualface:
I think these functions add own project better than merge into cocos. 

YuLei Liao (dualface)
qeeplay.com

minggo:
I am agree with @dualface.

ivfit:
So you don't want to be able to open URLs inside of cocos2d-x games?

There is a feature request for this:
http://www.cocos2d-x.org/boards/6/topics/3769

minggo:
Yep, i think it is better to be a good example, but not be merged into engine.

leszek-s:
What a shame you do not want this feature integrated with cocos2d-x :( As someone said here http://www.cocos2d-x.org/boards/6/topics/11290?r=16150#message-16150 I also think that this feature is super useful for game development especially on iOS where you can open urls like "itms-apps://itunes.com/apps/developer" to quickly easily open App Store app showing your apps or url like "itms-apps://ax.itunes.apple.com/WebObjects/......." to open app store on comments/review page of your game to let user rate it. I implement this features in all my apps/games so in my opinion openURL function should be integrated to cocos2d-x.

ghost:
@minggo @dualface 
I think you did not really understand how this is supposed to work. It is not about integrating a web page into an App, such as with UIWebView in iOS. Its about opening an URL in the native web browser.

For example, think about a game that has a "like us on facebook" button. When it is clicked, it would open the facebook page so the user can like it. Another example would be to open the developers home page.

It would be a nice feature to have. I have been searching for a nice way to implement it, and the only solution I found that does not base on writing native code for each platform, was using an ezibyte. 

In short, I think it really would save some time to quite a few developers.

ricardoquesada:
This is the way I see it:
- Tasks that are common to games should be provided by cocos2d (it could be in the core engine, as an extension, or something else, but users should code tasks that are common to games).
- Those tasks should have a certain "order & place": 
  - Users should be able to do common tasks easily. 
  - Users should be able to find the features easily.
- Common tasks should work on all the supported platform

Taking that into account, lets analyze the "open URL" functionality:
- I don't think it is a task that is widely used. But it is not rare enough to discard it.
- Games might need that functionality: On the About or Credits scenes developers might want to add a link to their support page.
- The patch says that it works on iOS, Android and Windows. 
  - Mac is missing. Should be easy to add support for it.
  - Linux is missing.
- A test case is missing in patch.
- The patch adds the method `openURL` in `CCApplication`. But I'm not sure that's the best place.
  - I think `Application` should be use only for tasks associated with the application.
    - As a side note: We need to define better the purpose of `Application`.
  - Perhaps we can create the `utils` namespace and add this function there.

Example:

``` c++
utils::openURL("www.example.com");
utils::openURI:("irc://freeenode.net");
// probably openURI is better than openURL since it supports any kind of protocol
// but I'm not sure how easy is to implement this on all the platforms.
utils::openURI:("http://www.example.com");

// this is another handy utility: grab an screenshot
utils::takeScreenshot(...");
utils::
```

I think the most difficult part is to think of a good place to put these little utilities.
The `cocos2d::utils` namespace is an alternative. Any other option ?

ghost:
@ricardoquesada I totally agree. I would be very excited to see that features implemented in cocos2d-x.

minggo:
Add `cocos2d::utils` is a good idea. This function can add extensions that most developer need.

PR #1944, opened 2013-02-01
add CCShapeNode

dualface:
CCShapeNode.h includes:
-   CCCircleShape
-   CCRectShape
-   CCPointShape
-   CCPolygonShape

easy use, more flexibility


dumganhar:
Hi, CCDrawingPrimitives will be deprecated.
So could you use CCDrawNode to implement CCxxxShape?

dumganhar:
My meaning is that make CCxxxShape inherit from CCDrawNode.
That will be better.

dualface:
OK

leszek-s:
i'm looking forward to see this feature in cocos2d-x

PR #1945, opened 2013-02-01
add CCLuaObjcBridge

dualface:
Call Objective-C form Lua, and call Lua from Objective-C.
-   Call Objective-C Class Static Method from Lua
-   Pass aruments to Objective-C
-   Pass Lua function to Objective-C
-   Call Lua function from Objective-C
### Lua code:

``` lua
local args = {
    username  = "dualface",
    password  = "123456",
    autologin = true,
    callback  = onLogin
}

local ok, ret = CCLuaObjcBridge.callStaticMethod("MySDK", "login", args)
if ok then
    -- call success
    print("ret = ", ret)
else
    -- call failure
    print("error = ", ret)
end
```
### Objective-C method:

``` objectivec
#include "CCLuaObjcBridge.h"

+ (NSString *)login:(NSDictionary *)dict
{
    NSString *username = [dict objectForKey:@"username"];
    NSString *password = [dict objectForKey:@"password"];
    BOOL autologin = [[dict objectForKey:@"autologin"] boolValue];
    int callback = [[dict objectForKey:@"callback"] intValue];

    if (.....)
    {
        CCLuaObjcBridge::callLuaFunctionById(callback, "ok");
    }
    else
    {
        CCLuaObjcBridge::callLuaFunctionById(callback, "error message");
    }

    CCLuaObjcBridge::releaseLuaFunctionById(callback);

    return @"something";
}
```

### DIFF
--- a/scripting/lua/cocos2dx_support/CCLuaEngine.cpp
+++ b/scripting/lua/cocos2dx_support/CCLuaEngine.cpp
@@ -54,6 +54,16 @@ bool CCLuaEngine::init(void)
     return true;
 }
 
+void CCLuaEngine::addSearchPath(const char* path)
+{
+    m_stack->addSearchPath(path);
+}
+
+void CCLuaEngine::addLuaLoader(lua_CFunction func)
+{
+    m_stack->addLuaLoader(func);
+}
+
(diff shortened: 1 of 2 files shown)


dualface:
When use % char, call CCLog crashed. puts() not.

PR #1952, opened 2013-02-04
Ensure we don't return invalid texture pointers.

thomaslee:
Bug manifested while trying to load a large image via
CCSprite::create(const char *) on a Nexus 7, which only supports
textures up to 2048x2048px.

CCSprite::create(const char *) returned true after failing to load the
large texture even though the underlying CCTexture2D had been released
during CCTexture2D::initWithImage(CCImage *) during the call to
CCTextureCache::addImage(const char *).

Now if CCTexture2D::initWithImage fails, we return NULL from
CCTextureCache::addImage(const char *). This will cause
CCSprite::create(const char *) to return NULL when texture loading
fails, which I believe is the expected/intended behaviour.

### DIFF
--- a/cocos2dx/textures/CCTextureCache.cpp
+++ b/cocos2dx/textures/CCTextureCache.cpp
@@ -460,6 +460,8 @@ CCTexture2D * CCTextureCache::addImage(const char * path)
                 else
                 {
                     CCLOG("cocos2d: Couldn't add image:%s in CCTextureCache", path);
+                    // texture is release()d if CCTexture2D::initWithImage fails
+                    texture = NULL;
                 }
             }
         } while (0);


cocos2dx/textures/CCTextureCache.cpp
```diff
@@ -460,6 +460,8 @@ CCTexture2D * CCTextureCache::addImage(const char * path)
                 else
                 {
                     CCLOG("cocos2d: Couldn't add image:%s in CCTextureCache", path);
+                    // texture is release()d if CCTexture2D::initWithImage fails
+                    texture = NULL;
```
minggo, line 464: I think should release it before assigning NULL.

thomaslee, line 464: Per the comment, the release() call is currently occuring in the CCTexture2D::initWithImage call. If you choose to release here, you'll want to remove the release() calls inside CCTexture2D::initWithImage or you'll be potentially accessing freed memory.

Not sure what implications that would have for other consumers of initWithImage, so I chose the safer option.

dumganhar, line 464: No, In `CCTexture2D::initWithImage`, if it fails, it will delete itself.
I think the delete operation should be after the line of invoking `CCTexture2D::initWithImage`.

thomaslee, line 464: Agreed @dumganhar. I think changing the semantics of CCTexture2D::initWithImage would be more consistent. Again, my concern here was that there was code elsewhere that would rely on the current behavior of initWithImage.

minggo, line 464: @thomaslee
I think you are right. We should not release texture in CCTexture2D::initWithImage.

minggo, line 464: Will you modify CCTexture2D::initWithImage in this PR or create a new PR?

thomaslee, line 464: I'll submit a separate PR -- feel free to close this one. Thanks!

PR #1954, opened 2013-02-04
Fix possible use of an invalid texture pointer.

thomaslee:
If CCTexture2D::initWithString fails inside CCLabelTTF, we would try to
continue with an invalid pointer.

### DIFF
--- a/cocos2dx/label_nodes/CCLabelTTF.cpp
+++ b/cocos2dx/label_nodes/CCLabelTTF.cpp
@@ -269,12 +269,14 @@ bool CCLabelTTF::updateTexture()
 		return false;
     }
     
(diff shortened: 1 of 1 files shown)


minggo:
We may create a CCLabelTTF that doen't have texture, and set the texture later.
The PR will cause crash in LabelTest.

thomaslee:
This is a similar issue to #1952. The texture will delete itself inside the initWithString call (actually via initWithImage) if that call fails. The PR may not be the correct solution -- it may be better/more consistent to change the semantics of initWithImage as I think @dumganhar is suggesting -- but there's definitely a potential invalid memory access attempt here.

thomaslee:
This should be fixed with the resolution of #1957

PR #1957, opened 2013-02-04
Don't call release() if CCTexture2D::initWith* fails

thomaslee:
Per the discussion on #1952 -- don't call this->release() if CCTexture2D::initWithImage or CCTexture2D::initWithPVRTCData fails.

At a glance, I don't think this would introduce any leaks given that we generally prefer to return a blank/empty CCTexture2D than a NULL pointer (i.e. most code seems to assume that initWithImage etc. will leave the CCTexture2D intact).

That said, it could probably use a quick review :)

### DIFF
--- a/cocos2dx/textures/CCTexture2D.cpp
+++ b/cocos2dx/textures/CCTexture2D.cpp
@@ -265,7 +265,7 @@ bool CCTexture2D::initWithImage(CCImage *uiImage)
     if (imageWidth > maxTextureSize || imageHeight > maxTextureSize) 
     {
         CCLOG("cocos2d: WARNING: Image (%u x %u) is bigger than the supported %u x %u", imageWidth, imageHeight, maxTextureSize, maxTextureSize);
-        return NULL;
+        return false;
     }
     
     // always load premultiplied images


cocos2dx/textures/CCTexture2D.cpp
```diff
@@ -266,7 +265,6 @@ bool CCTexture2D::initWithImage(CCImage *uiImage)
     if (imageWidth > maxTextureSize || imageHeight > maxTextureSize) 
     {
         CCLOG("cocos2d: WARNING: Image (%u x %u) is bigger than the supported %u x %u", imageWidth, imageHeight, maxTextureSize, maxTextureSize);
-        this->release();
         return NULL;
```
minggo: Would you mind also changing NULL to false?

thomaslee: Oops, good catch :)

dumganhar:
I think we should also invoke `CC_SAFE_RELEASE_NULL(pTexture)` after if  `initWithImage` return false.

thomaslee:
Not quite sure I follow. Where do you mean?

dumganhar:
I mean that:

<pre>
CCTexture2D* pTexture = new CCTexture2D();
if (pTexture && pTexture->initWithImage(...))
{
    ... do something
}
else
{
// Should invoke `CC_SAFE_RELEASE_NULL(pTexture)` here.
    `CC_SAFE_RELEASE_NULL(pTexture)` 
}
</pre>

minggo:
Yep, but should do it in a separate PR.

minggo:
@thomaslee
Will you fix the memory leak as @dumganhar  said?

thomaslee:
Sure, I'll take a look at it tomorrow evening. It's getting kind of late here. :)

thomaslee:
Actually, just before I go -- I'm not sure we always want to do this. For example, the main consumer of CCTexture2D::initWithImage is CCTextureCache::addImage -- there's a lot of code that relies upon CCTextureCache::addImage always returning a valid texture. I think it would break a lot of stuff if it suddenly started returning NULL when initWithImage fails. I don't think we want to change that behavior do we?

I can't really see it being used anywhere else where we're not doing something similar. So I guess I'd like to know where @dumganhar feels the extra release() is necessary?

minggo:
I am not sure what error will happen if we invoke release. But i think we should have a try.

thomaslee:
Hrm. I don't really feel confident doing this given I haven't had much exposure to the code base. Further, I think releasing & assigning to NULL in every case would break a lot of stuff. I don't really have the time to commit to fixing the breakages. :(

Honestly, I think in most cases we should already be doing the Right Thing from what code I have seen -- though of course there may be exceptions. I don't think just arbitrarily release()ing stuff & returning NULL where callers expect a soft fail over is the right way to go though.

If you guys on the core have a solid grasp of how you want this to work, maybe you should run with this one? I'll just get in the way :)

minggo:
@thomaslee
Yep, we will do more tests before we modify it.

thomaslee:
Great -- thanks!

PR #1959, opened 2013-02-06
use tinyxml2 instead of libxml2

minggo:


martell:
Man that was good thanks minggo :)

minggo:
It is my pleasure.

martell:
Does this get me on the Authors list for my CV? :)

minggo:
I am sorry, just added it.

martell:
Yay :+1:
On my CV :)

RichardGale:
Does tinyxml improve parsing time, or decrease memory use? Any links to resources/results on this change?

PR #1964, opened 2013-02-06
Marmalade webp fixes

fape:
Add some fixes for webp image format in marmalade. 
Tested with simultator and android


dumganhar:
Thank you.

PR #1967, opened 2013-02-06
Fix for unaligned memory access crash in CCBReader::readFloat()

BodbDearg:
A fix for a potential unaligned memory access crash in CCBReader::readFloat(), caused by incorrect compiler optimisations/assumptions surrounding the use of memcpy().

See source comments for more details on the change.

### DIFF
--- a/extensions/CCBReader/CCBReader.cpp
+++ b/extensions/CCBReader/CCBReader.cpp
@@ -515,8 +515,21 @@ float CCBReader::readFloat() {
                  * TODO still applies in C++ ? */
                 float * pF = (float*)(this->mBytes + this->mCurrentByte);
                 float f = 0;
-                memcpy(&f, pF, sizeof(float));
-                this->mCurrentByte += 4;
+                
+                // N.B - in order to avoid an unaligned memory access crash on 'memcpy()' the the (void*) casts of the source and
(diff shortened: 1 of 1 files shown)


BodbDearg:
Checked for the same issue in the Cocos2d/Objective-C version of CCBReader - looks as though the issue was worked around there with a byte by byte copy:

https://github.com/cocos2d/CCBReader/commit/a205146d4664e15f5fcf9d47146b26f016d6a055#commitcomment-2574544

This would be an alternate way of fixing the problem.

dumganhar:
Thank you. Merged.

BodbDearg:
Thanks very much!

sunxiaoyu:
this issue occurred to me on my release version,but codes works well on real-device debug,I have notice that cocos2d for iPhone share the same codes as your pre version,but it still works well,there is no different compiler settings between two testing projects.So I wonder if this bug only happened in some very particular conditions:)

minggo:
Could you please use latest codes? This file is changed a little.
Thank you.

sunxiaoyu:
okay! I have already did

minggo:
Can you reproduce it?

sunxiaoyu:
This fix has already solved my issue,I just got confused why didin't I meet the same issue on my project which is based on cocos2d for iphone

PR #1974, opened 2013-02-11
CCScale9Sprite : Simpler, correct handling of spritesheet sources

folecr:
- bounds of insets are calculated relative to sprite
- transformations are then applied to get the texture coordinates
- handle sprites stored rotated in a spritesheet correctly


dumganhar:
Thanks.

PR #1977, opened 2013-02-12
Added window_group creation to CCEGLView for blackberry

moadib:
This has already been a part of https://github.com/cocos2d/cocos2d-x/pull/730


dumganhar:
Thanks.

Issue #1976, opened 2013-02-12
On Blackberry CocosDenshion doesn't support multiple background musics, used only last loaded.


moadib:
It's awful.

haxpor:
Is that supposed to be the case for other platforms? Mostly 1 music but multiple of sfx played in the game.
Anyway better waiting for other contributors to answer this.

moadib:
On IOS, Android and Windows too we can preload two(for example) tracks and switch them by calling `PlayBackgroundMusic("source1");` or `PlayBackgroundMusic("source2");`

PR #1978, opened 2013-02-12
Added support for multiple background music sources to CocosDenshion for BB

moadib:
As i mentioned here https://github.com/cocos2d/cocos2d-x/issues/1976 Blackberry implementation have different behavior compare to Ios, Android and Windows

If user want to preload multiple background musics and switch them then we need something like proposed changes.


dumganhar:
Thank you.

PR #1981, opened 2013-02-14
Fix for CCScrollView clipping bug

ztoky:
fix : wrong cliping rect when ancestors scale is not 1


dumganhar:
Thank you. It was merged at https://github.com/cocos2d/cocos2d-x/pull/1986.

PR #1982, opened 2013-02-15
CCLens3d can be concave

billy1380:


cocos2dx/actions/CCActionGrid3D.h
```diff
@@ -96,7 +96,10 @@ class CC_DLL CCLens3D : public CCGrid3DAction
     inline float getLensEffect(void) { return m_fLensEffect; }
     /** Set lens center position */
     inline void setLensEffect(float fLensEffect) { m_fLensEffect = fLensEffect; }
-    
+    /** Set whether lens is concave */
+    inline void setConcave(float fLensEffect) { m_bConcave = fLensEffect; }
```
minggo, line 100: Why the type of parameter is float?
Isn't it boolean?

minggo:
What changes were done to cocos2dx/proj.ios/cocos2dx.xcodeproj/project.pbxproj?

cocos2dx/actions/CCActionGrid3D.h
billy1380, line 100: It should be:

 inline void setConcave(bool bConcave) { m_bConcave = bConcave; }

Very sorry... copy paste mistake

billy1380:
My default XCode settings are 2 spaces for tab, so I changed the project setting so that it is 4 spaces per tab to match the rest of the code files...

billy1380:
minggo, just reject it... I will send you another pull request and try not to make so many mistakes

PR #1991, opened 2013-02-18
Make ScriptingCore.cpp compile with C++11 on iOS

SevInf:
In my current project, I need to use both JS Bindings and C++11 features.
When flags -stdlib=libc++ -std=c++11 are set, iOS compilation fails because of write, read, and close functions are part of the unistd.h header file and its not included in ScriptingCore.cpp. Including it won't break compilation with default settings (tested on iOS and Android), but allow to compile project on iOS using C+11.


dumganhar:
Thank you.

PR #2000, opened 2013-02-19
fixed the fileformat of proj.android/jni/list.sh

mutoo:
list.sh doesn't work in linux/mac due to its fileformat


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2004.
Thanks.

But does this file have any usefulness?
I have never used this file before. Is it used for other tools?

mutoo:
@dumganhar I used ${find ../../Classes/. -name "*.cpp" | xargs} to find out all cpp classes, so as I can append those to Android.mk
but later I found list.sh does the same thing.
(I am not good at this aspect)
do you have any suggestion that is better to do so?

dumganhar:
No idea about the better approach. Just use the way that works. :)

PR #1849, opened 2013-01-16
Added keyboard support for MAC

cocos2dx-pull-request:
Refers to #1323.

PR #2002, opened 2013-02-19
Keyboard dispatcher and delegate.

jllust:
Keyboard dispatcher and delegate added. NSEvents bypassed Obj-c dispatching in favor of direct call to c++ method. Delegate methods inherited into CCLayer with enabled flag.


dumganhar:
<code>This pull request cannot be automatically merged.</code>
And is this Pull Request same as https://github.com/cocos2d/cocos2d-x/pull/1849?

jllust:
Yes, I would say I built much the same thing as the open request #1849
I will withdraw my request and check for open tickets next time before I spend time writing something.
Thanks.

PR #2006, opened 2013-02-20
Add error logging for iOS when a font fails to load during text rendering

BodbDearg:
Make CCImage on iOS log an error message when an attempt is made to render text with a font which could not be loaded. Print an error message to the console and the name of the font which failed to load (in addition to current logic of loading a default replacement font) so that developers are alerted to the issue.

### DIFF
--- a/cocos2dx/platform/ios/CCImage.mm
+++ b/cocos2dx/platform/ios/CCImage.mm
@@ -23,6 +23,7 @@ THE SOFTWARE.
 ****************************************************************************/
 #import "CCImage.h"
 #import "CCFileUtils.h"
+#import "CCCommon.h"
 #import <string>
 
 #import <Foundation/Foundation.h>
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/ios/CCImage.mm
```diff
@@ -220,6 +221,7 @@ static bool _initWithString(const char * pText, cocos2d::CCImage::ETextAlign eAl
         }      
         else
         {
+            CCLOGERROR("Error! Failed to load the font '%s'! A default system font will be used instead.", pFontName);
             fntName = _isValidFontName(pFontName) ? fntName : @"MarkerFelt-Wide";
```
minggo, line 225: Can not say use default system font before invoking _isValidFontName().

BodbDearg, line 225: Is it not true that the code would always use a system font when reaching the else {} because if even if _isValidFontName() returns true, the same font load which failed would be attempted again with 'UIFont fontWithName:fntName size:nSize];'? If that call failed once for the same value of 'fntName' then it would almost certainly fail again - causing the system font to be used even if the font name is determined to be ok.

I agree though, this code here is probably not the best. After working a little more with fonts and cocos builder CCB files I think I understand the nature of what is going on here a little better, and have a few additional changes to submit in order to make custom ttf fonts referenced from CocosBuilder CCB files work in a cross platform way. Going to close this pull request for now and submit a separate pull request instead with those changes.

Thanks for taking a look at this!

Issue #1971, opened 2013-02-10
(iOS) Improper Action with CCEditbox


CodeSnooker:
Hi James

@dumganhar 

Cocos2d-x Version: cocos2d-2.1beta3-x-2.1.0

Steps to reproduce issue:
1. Create a Scene.
2. Add a full screen size layer to the scene.
3. Add a CCEditbox to layer and put some text to it.
4. Add a CCMenuItem to layer and set actions to move layer to some point.
5. Click on Menu Item to initiate the move action of layer.

Result:

Only the background of the CCEditbox has moved but the text is there at the same place.

Another test
1. Hide the CCEditbox.
2. Click on the area where the CCEditbox was added.

Result:
1. Text is still visible even after hiding the box.
2. Keyboard will pop up on touch the area where CCEditbox was added.

Note: The issue is reproducible on iOS platform only.

CodeSnooker:
@dumganhar  any update on this?

dumganhar:
Hi, @CodeSnooker 
I sent a Pull Request to resolve some issues of CCEditBox here(https://github.com/cocos2d/cocos2d-x/pull/2045/files).
I think it will work with some actions that changing the postion of CCEditBox or its parent's layer.
Because it called `setPosition` in CCEditBoxImplIOS::visit.
But other actions like rotate and scale will not work.

PR #2012, opened 2013-02-20
Added 'setFont' and 'setAnchorPoint' to CCEditBox

neokim:
Now CCEditBox has an ability to change its font and size, and also it is working fine now with 'setAnchorPoint' on iOS which wasn't working properly.
The changes are applied to the bundle test project and tested on iOS and Android.
ps) Because there was a linkage error on Android build with the test project with its default setting, I needed to add 'APP_ABI = armeabi-v7a' to Application.mk. I'm using ndk-r7b.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ExtensionsTest/EditBoxTest/EditBoxTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/ExtensionsTest/EditBoxTest/EditBoxTest.cpp
@@ -38,7 +38,7 @@ EditBoxTest::EditBoxTest()
     // top
     m_pEditName = CCEditBox::create(editBoxSize, CCScale9Sprite::create("extensions/green_edit.png"));
     m_pEditName->setPosition(ccp(visibleOrigin.x+visibleSize.width/2, visibleOrigin.y+visibleSize.height*3/4));
-#ifdef CC_TARGET_OS_IPHONE
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
 	m_pEditName->setFont("Paint Boy", 25);
 #else
 	m_pEditName->setFont("fonts/Paint Boy.ttf", 25);
@@ -54,7 +54,7 @@ EditBoxTest::EditBoxTest()
     // middle
     m_pEditPassword = CCEditBox::create(editBoxSize, CCScale9Sprite::create("extensions/orange_edit.png"));
     m_pEditPassword->setPosition(ccp(visibleOrigin.x+visibleSize.width/2, visibleOrigin.y+visibleSize.height/2));
-#ifdef CC_TARGET_OS_IPHONE
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
 	m_pEditPassword->setFont("A Damn Mess", 30);
 #else
 	m_pEditPassword->setFont("fonts/A Damn Mess.ttf", 30);


samples/Cpp/TestCpp/Classes/ExtensionsTest/EditBoxTest/EditBoxTest.cpp
```diff
@@ -38,8 +38,14 @@
     // top
     m_pEditName = CCEditBox::create(editBoxSize, CCScale9Sprite::create("extensions/green_edit.png"));
     m_pEditName->setPosition(ccp(visibleOrigin.x+visibleSize.width/2, visibleOrigin.y+visibleSize.height*3/4));
+#ifdef CC_TARGET_OS_IPHONE
```
dumganhar: I think using `#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)` is better.

scripting/javascript/bindings/generated
```diff
@@ -1 +1 @@
-Subproject commit 35c9124a808c8e67b5a9bef68d4ff4433dad8bd4
+Subproject commit 4731abe94375ad4b9bde20cfd8bd48d0e1538fb1
```
dumganhar, line 1: Why did you change the submodule reference?

tools/cxx-generator
```diff
@@ -1 +1 @@
-Subproject commit b92f99a3e53c5d4c19eb166e649664bd7fd80bab
+Subproject commit 39286b435de49cfb745297bc514ed9f7d6d1f7f9
```
dumganhar, line 1: Why did you change the submodule reference?

samples/Cpp/TestCpp/proj.android/jni/Application.mk
```diff
@@ -1,2 +1,3 @@
 APP_STL := gnustl_static
 APP_CPPFLAGS := -frtti -DCC_ENABLE_CHIPMUNK_INTEGRATION=1 -DCOCOS2D_DEBUG=1
+APP_ABI = armeabi-v7a
```
dumganhar, line 3: Which host os you are using? MAC or Windows? I have tested on MAC, and it was ok.

samples/Cpp/TestCpp/Classes/ExtensionsTest/EditBoxTest/EditBoxTest.cpp
neokim: No problem.

scripting/javascript/bindings/generated
neokim, line 1: I was new to git in the morning and still I am. I don't know what they meant, just thought that it's kind of repository metadata. You can just rollback them if they are not necessary.

samples/Cpp/TestCpp/proj.android/jni/Application.mk
neokim, line 3: Mac. It might be caused by ndk version. I'm using ndk-r7b. What about you?

dumganhar, line 3: Yes, please use ndk-r8d and try again.

extensions/GUI/CCEditBox/CCEditBoxImplAndroid.cpp
```diff
@@ -88,6 +90,19 @@ bool CCEditBoxImplAndroid::initWithSize(const CCSize& size)
...
+	if(m_pLabel != NULL) {
+		m_pLabel->setFontName(pFontName);
+		m_pLabel->setFontSize(fontSize);
+	}
+	
+	if(m_pLabelPlaceHolder != NULL) {
+		m_pLabelPlaceHolder->setFontName(pFontName);
+		m_pLabelPlaceHolder->setFontSize(fontSize);
+	}
+}
```
dumganhar, line 104: How about separating `setFont` to methods as follows?
- setFontName
- setPlaceholderFontName
- setFoneSize
- setPlaceholderFontSize

neokim, line 104: That was my first bet. But then, on iOS `[UIFont fontWithName:fntName size:fontSize]` needs to be called twice to change both font and size.

dumganhar, line 104: OK, how about
- setFont
- setPlaceholderFont
  ?

neokim, line 104: I've added `setPlaceholderFont()` API to CCEditBox. It is not implemented on iOS because `UITextField` of iOS doesn't provide any way to customize placeholder. To do that, `UITextField` needs to be subclassed and overridden.

samples/Cpp/TestCpp/Classes/ExtensionsTest/EditBoxTest/EditBoxTest.cpp
dumganhar: Could you also update this?

extensions/GUI/CCEditBox/CCEditBoxImplAndroid.cpp
dumganhar, line 104: > To do that, UITextField needs to be subclassed and overridden.

OK, thank you. So do you have a plan to make it supported on iOS?

neokim, line 104: I have customized the iOS implementation of CCEditBox in my project to solve the issue #1971 and some other problems. It already supports placeholder customization partially.
The customization I made is that UITextField is visible only when the CCEditBox is in edit mode. And in non-edit mode, UITextField is hidden and CCLabelTTF is shown instead with all same attributes at the same place. So the placeholder in non-edit mode can be customized.
I think that this customization is just a work-around for some problems and definitely not a perfect solution.
But it is still useful, so I'm going to issue another pull request for this after current pull request.

dumganhar, line 104: OK, I will merge this Pull Request today. Thanks @neokim .

dumganhar:
Seems win32 and mac platforms also need to be updated.:)

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2044
Thanks.

PR #2017, opened 2013-02-21
Added cleanup() and reset() functionality to ScriptingCore. Also added functionality to set an externalScriptPath.

nicolasgramlich:
This PR is best viewed with whitespace changes ignored: https://github.com/cocos2d/cocos2d-x/pull/2017?w=1

The methods `cleanup()` and `reset()` are necessary i.e. when a game is not just a single screen (i.e. an Android Activity), but a screen is destroyed and recreated, there should be a way to reset the `ScriptingCore`. On Android the GLThread changes when the Activity i.e. is destroyed and recreated. Touching the `ScriptingCore` causes a guaranteed crash. -> When the 'screen' is destroyed the caller has to ensure that the `cleanup()` is called from the same Thread. This change was developed with @funkaster and @folecr.

The `externalScriptPath` is very useful i.e. when debugging on the iOS simulator, one can redirect the path the scripts are loaded from to the host machine (On Android i.e. to the SD-Card), allowing to change the JS files without recompilation and even without restarting the app.

Also removed a good bunch of trailing whitespaces.

### DIFF
--- a/scripting/javascript/bindings/Android.mk
+++ b/scripting/javascript/bindings/Android.mk
@@ -17,7 +17,8 @@ LOCAL_SRC_FILES := ScriptingCore.cpp \
                    js_bindings_system_functions.cpp \
                    js_bindings_system_registration.cpp \
                    js_bindings_ccbreader.cpp \
-                   js_bindings_core.cpp
+                   js_bindings_core.cpp \
+                   generated/cocos2dx.cpp
 
 LOCAL_CFLAGS := -DCOCOS2D_JAVASCRIPT
 


scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -441,13 +445,29 @@ JSBool ScriptingCore::runScript(const char *path, JSObject* global, JSContext* c
...
+                return JS_FALSE;
+            }
+        }
+
+        rpath = tmp;
     } else {
+        cocos2d::CCFileUtils * futil = cocos2d::CCFileUtils::sharedFileUtils();
         rpath = futil->fullPathForFilename(path);
     }
+
```
dumganhar, line 470: Is `externalScriptPath` necessary?
The logic in `CCFileUtils::fullPathForFilename` will check whether the `path` passed in is a relative or absolute path.
Therefore, i think 

<pre>
if (this->externalScriptPath == "") {
        const char * tmp;

        if (path[0] == '/') {
            tmp = path;
        } else {
            cocos2d::CCFileUtils * futil = cocos2d::CCFileUtils::sharedFileUtils();
            tmp = futil->fullPathFromRelativePath(path);

            if (!tmp) {
                CCLOG("!tmp. returning JS_FALSE");
                return JS_FALSE;
            }
        }

        rpath = tmp;
    } else {
        cocos2d::CCFileUtils * futil = cocos2d::CCFileUtils::sharedFileUtils();
        rpath = futil->fullPathForFilename(path);
    }
</pre>

is not needed.
PLZ look at the implementation of `CCFileUtils::fullPathForFilename`

<pre>
std::string CCFileUtils::fullPathForFilename(const char* pszFileName)
{
    CCAssert(pszFileName != NULL, "CCFileUtils: Invalid path");
    
    std::string strFileName = pszFileName;
    if (isAbsolutePath(pszFileName))   // We check whether the path is an absolute path here.
    {
        //CCLOG("Return absolute path( %s ) directly.", pszFileName);
        return pszFileName;
    }
</pre>

nicolasgramlich, line 470: Sorry, I should have explained it better...

The solution provided allows for seamless change of the root path for all
the scripts in the ScriptingCore. Say you could do something like (pseudo
code)

'''
#ifdef SIMULATOR
ScriptingCore->getInstance()->sertExternalScriptPath("/PATH/ON/HOST");
#endif
'''

And all ''runScript'' and ''require'' calls in JS go through to the host.
No changes in JS required.

I hope that explains it better.

dumganhar:
> The solution provided allows for seamless change of the root path for all
> the scripts in the ScriptingCore.

But we could easily achieve that by add a high priority search path as follows:

<pre>
#ifdef SIMULATOR
CCFileUtils* pFileUtils = CCFileUtis::sharedFileUtils();
vector<string> searchPaths = pFileUtils->getSearchPaths();
searchPaths.insert(0, "/PATH/ON/HOST");
pFileUtils->setSearchPaths(searchPaths);
#endif
</pre>

Right?

nicolasgramlich:
Coll, I didn't know that functionality. Let me make a new (and cleaner) PR with only the resetting.

dumganhar:
OK, Thank you.

PR #2013, opened 2013-02-20
add a transform in CCNode

jxhgzs:
...ship but we still need to maintain transformation  parent-child
        Example:
          GameObject_Parent(CCNode; create from CCNode)
                 GameObject_sub1(CCNode; Create from the same CCSpriteBatchNode,the node's parent is CCLayer )
                        RenderComponents1( Create form the same CCSpriteBatchNode)
                        Components2
                        ....
                 GameObject_sub2 (CCNode; Create from the CCNode,the node's parent is GameObject_Parent )
                       RenderComponents1( not used batchNode)
         Now, GameObject_sub1 not GameObject_Parent node, but I want it to simulate the change of the son is GameObject_Parent;
         I Need additional the father's matrix  in child Node ;

Example:(in cocos2d-x HelloWorldScene.cpp)
        //create batchNode
    CCSpriteBatchNode\* batch=  CCSpriteBatchNode::create("Icon.png");
    this->addChild(batch);
    //create sprite
    CCSprite\* parent = CCSprite::createWithTexture(batch->getTexture());
    CCSprite\* sub = CCSprite::create("Icon-72.png");  
    batch->addChild(parent);
    this->addChild(sub);  // it is not parent's sub

```
//position
parent->setPosition(ccp(200, 200));
CCAffineTransform t = parent->nodeToParentTransform();
sub->translateFromOtherNode(t);

//scale
parent->setScale(2);
t = parent->nodeToParentTransform();
sub->translateFromOtherNode(t);

//rotation
parent->setRotation(20);
t = parent->nodeToParentTransform();
sub->translateFromOtherNode(t);
```

PR #2018, opened 2013-02-21
create tiled map from buffer

jxhgzs:
add function for initializes a TMX Tiled Map with a TMX formatted XML string and a texture buffer;
When you need to pack resources will be used

### DIFF
--- a/cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
+++ b/cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
@@ -65,12 +65,6 @@ bool CCTMXLayer::initWithTilesetInfo(CCTMXTilesetInfo *tilesetInfo, CCTMXLayerIn
 		}
         else
 		{
-#if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
-            if (tilesetInfo->m_sSourceImage[0] == '/')
-			{
-                tilesetInfo->m_sSourceImage.erase(tilesetInfo->m_sSourceImage.begin());
-			}
-#endif
             texture = CCTextureCache::sharedTextureCache()->addImage(tilesetInfo->m_sSourceImage.c_str());
 		}
 	}


cocos2dx/base_nodes/CCNode.cpp
```diff
@@ -1158,13 +1159,17 @@ CCAffineTransform CCNode::nodeToParentTransform(void)
...
 
     return m_sTransform;
 }
-
+void CCNode::translateFormOtherNode(CCAffineTransform &transform)
+{
+	m_sOtherTransform  = CCAffineTransformMake(transform.a, transform.b, transform.c, transform.d, transform.tx, transform.ty);
+	m_bTransformDirty = true;
+}
 CCAffineTransform CCNode::parentToNodeTransform(void)
```
dumganhar, line 1167: Why does this PR contain some changes in https://github.com/cocos2d/cocos2d-x/pull/2013 ?

cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
```diff
@@ -57,10 +57,23 @@ bool CCTMXLayer::initWithTilesetInfo(CCTMXTilesetInfo *tilesetInfo, CCTMXLayerIn
...
+			texture = tilesetInfo->m_pTexture;
+		}
+        else
+		{
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
+            if (tilesetInfo->m_sSourceImage[0] == '/')
+			{
+                tilesetInfo->m_sSourceImage.erase(tilesetInfo->m_sSourceImage.begin());
+			}
+#endif
```
dumganhar: Is this only needed for Android?
Why does the image source string contain an absolute path?
In which case this will happen?
I think in `.tmx` file, it should not contain an image source as an absolute path.

cocos2dx/tilemap_parallax_nodes/CCTMXLayer.cpp
```diff
@@ -57,10 +57,17 @@ bool CCTMXLayer::initWithTilesetInfo(CCTMXTilesetInfo *tilesetInfo, CCTMXLayerIn
...
+	{
+        if (tilesetInfo->m_pTexture != NULL)
+		{
+			texture = tilesetInfo->m_pTexture;
+		}
+        else
+		{
+            texture = CCTextureCache::sharedTextureCache()->addImage(tilesetInfo->m_sSourceImage.c_str());
+		}
+	}
```
minggo, line 70: Please format the codes。

jxhgzs, line 70: it is right in xcode; May is a web page display problems

jxhgzs:
Use other method

PR #2033, opened 2013-02-23
Add VS2012 Express template installing support

fispurring:
Add VS2012 Express For Windows Desktop support in install-templiates-msvc.bat


dumganhar:
Thank you.

PR #2050, opened 2013-02-26
fixed #1751: Clicking outside of CCTableViewCell will also get response when container size is smaller than view size.

dumganhar:


jdmunro:
Sorry to dig up an old issue - it seems this issue exists if the bottom-up order is used. On a side note, howcome `kCCTableViewFillBottomUp` is the default fill order? Most iOS users would be more familiar with a top-down fill order by default.

PR #2075, opened 2013-03-01
Tiny Linux Makefile improvements

azmeuk:
Hi,
I have modified linux makefile so they can handle source dependencies. Now if a header file is edited, sources that depends on this header will be recompiled.


azmeuk:
This pull request do it event better : https://github.com/cocos2d/cocos2d-x/pull/2093

gelldur:
Sorry i'm not sure is this commit merged or not?

azmeuk:
It is not merged, and shouldn't be, as #2093 does the same thing, plus other features.

PR #2080, opened 2013-03-02
fix compile error with xcode on case-sensitive file system.

biteforest:
fix compile error with xcode on case-sensitive file system.


minggo:
Thank you.

Issue #1812, opened 2013-01-03
CCScrollView bug


ivyshao:
1.  when add a menu in scrollview, will not scroll if click on the menu
   - need to set touch priority to kCCMenuHandlerPriority-1
2. after fix the first one, after scroll, the button click event is also triggered
   - need claim touch
3. button out of scroll content size, is still clickable
   - need check if touch is in the content size area

dumganhar:
So did you fix these issues? If so, could you send the patch to us?

ivyshao:
no, I meet same issues for cocos2d-iphone, and fixed. But no patch for cocos2d-x.

dumganhar:
Could you send me the patch you fixed for cocos2d-iphone?
I could help to make it works on cocos2d-x.
Thanks.

ivyshao:
diff: 

wrong one: 

fixed one: 

Changes:
1. claimTouch in ccTouchMoved
2.registerWithTouchDispatcher with priority kCCMenuTouchPriority - 1
3. in ccTouchBegan:
    if (!CGRectContainsPoint(frame, [container_ convertToWorldSpace:[container_ convertTouchToNodeSpace:touch]])) {
        return YES;
    }
4. removeTouchInChildren in ccTouchEnded if touch is not handle in CCTouchBegin

BR

ivyshao:
I have fixed this. I override CCScrollView and CCMenu, to make it work. will send you my code

dumganhar:
OK, thank you. :)

ztoky:
Is this bug fixed?

dumganhar:
@ztoky 
You could refer to https://github.com/cocos2d/cocos2d-x/pull/2081 .
It isn't merged now.

PR #2081, opened 2013-03-02
fix https://github.com/cocos2d/cocos2d-x/issues/1812, fix CCScrollView with menu issues

ivyshao:


michaelcontento:
Hi,

I've merged this PR into my local cocos2d-x clone and fixed some stuff (missing include guards and namespace definitions) - the result can be found here https://gist.github.com/michaelcontento/5443151.

Just apply my gist after this PR and everything should work fine :)
Would be nice if this could be merged into `master` soon.

Thanks, Michael

PR #2082, opened 2013-03-03
Adding resumeSchedulerAndAction(bool) and pauseSchedulerAndActions(bool)

gelldur:
request from http://www.cocos2d-x.org/boards/6/topics/6887

### DIFF
--- a/cocos2dx/base_nodes/CCNode.cpp
+++ b/cocos2dx/base_nodes/CCNode.cpp
@@ -1097,7 +1097,7 @@ void CCNode::resumeSchedulerAndActions()
 void CCNode::resumeSchedulerAndActions(bool applyToChildren)
 {
 	resumeSchedulerAndActions();
-	if(applayToChildren == false)
+	if(applyToChildren == false)
 		return;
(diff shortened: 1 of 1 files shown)


cocos2dx/base_nodes/CCNode.cpp
```diff
@@ -1094,12 +1094,41 @@ void CCNode::resumeSchedulerAndActions()
     m_pActionManager->resumeTarget(this);
 }
 
+void CCNode::resumeSchedulerAndActions(bool applayToChildren)
```
dumganhar: `applaytoChildren` ~> `applyToChildren`?

gelldur:
I fixed misspell.

minggo:
Is it needed? I think the sub class can override this method if it needs this feature.

gelldur:
Hmm i use it in CCLayer and in CCScene. Maybe it is'n so much needed but it is short method and nice feature to make pause/resume game

gelldur:
Merge or don't merge :)

PR #2083, opened 2013-03-03
fix CCScale9Sprite bugs

biteforest:
fix CCScale9Sprite bugs. when  reading spriteframe frome plist, spriteframe's ratote may be seted. but CCScale9Sprite can't parse spriteframe right. Another bugs is inset of ccscale9sprite.

### DIFF
--- a/extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
+++ b/extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
@@ -308,7 +308,7 @@ bool CCScale9Sprite::updateWithBatchNode(CCSpriteBatchNode* batchnode, CCRect re
         CCRect rotatedcentertopbounds = centertopbounds;
         
         t = CCAffineTransformTranslate(t, rect.size.height+rect.origin.x, rect.origin.y);
-        t = CCAffineTransformRotate(t, 1.57079633);
+        t = CCAffineTransformRotate(t, 1.57079633f);
         
         centerbounds = CCRectApplyAffineTransform(centerbounds, t);
         rightbottombounds = CCRectApplyAffineTransform(rightbottombounds, t);


folecr:
Hi,

Could you please add a test case for the bugs you fixed?
Current test cases are here:
 https://github.com/cocos2d/cocos2d-js-tests/blob/master/tests/ExtensionsTest/S9SpriteTest/S9SpriteTest.js

Thanks.

dumganhar:
@biteforest 
Also, why did you modify the whole file?
I couldn't see what you changed on github. Was the `RETURN` symbol different?

biteforest:
sorry,its my first time to use github, iam not good at git. next time i will push the bug branch instead of master branch. 

已从手机发送

-------- Original message --------
Subject: Re: [cocos2d-x] fix CCScale9Sprite bugs (#2083)

biteforest:
ok, if im free this weekend, i ll try

已从手机发送

-------- Original message --------
Subject: Re: [cocos2d-x] fix CCScale9Sprite bugs (#2083)

biteforest:
@dumganhar @folecr  
Reject this pull request, i'll send pull request later.

biteforest:
@dumganhar 
i don't why the whole was changed. but i review this change by tortoisgit, the change of file content is ok

folecr:
@biteforest I think your editor is saving the files after changing the whitespace. You should be able to disable that feature.

You can see the diffs on github with whitespace changes ignored by adding a ?w=1 to the URL (from https://github.com/blog/967-github-secrets)

biteforest:
@folecr i copy the two different files and view them in hexeditor, but I never find any diffrent between white spaces. you can try to merge this branch in your test branch. I really don't know what's the problem.

minggo:
May be the line break is different. Did you use VS line break style?

PR #2086, opened 2013-03-03
update build-win32.bat

lgb:
- return to current directory after run the test
- some small changes

### DIFF
--- a/build-win32.bat
+++ b/build-win32.bat
@@ -48,7 +48,7 @@ echo. * Check the cocos2d-win32 application "TestCpp.exe" ...
 echo. */
 echo.
 
-cd ".\Release.win32\"
+pushd ".\Release.win32\"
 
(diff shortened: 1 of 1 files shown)


build-win32.bat
```diff
@@ -44,11 +44,11 @@ if %VC_VER%==100 (
...
 
 echo./*
-echo.* Check the cocos2d-win32 application "TestCpp.exe" ...
-echo.*/
+echo. * Check the cocos2d-win32 application "TestCpp.exe" ...
+echo. */
 echo.
 
-cd ".\Release.win32\"
+pushd ".\Release.win32\"
```
dumganhar, line 51: Why should use `pushd`?
Is `build_win32.bat` used by other script ?

lgb, line 51: No, but I call it from command line and was surprised to being in a different directory.

dumganhar:
Thank you.

Issue #689, opened 2012-01-29
SimpleAudioEngine on Linux port


xal:
While try to play background music on linux port

``` c++
SimpleAudioEngine::sharedEngine()->playBackgroundMusic("background.wav", true);
```

Nothing happens

But, when i play effect

``` c++
 SimpleAudioEngine::sharedEngine()->playEffect("move.wav");
```

It's played twice

I try other sound format's and but nothing has changed.
Also I add this code before play sound 

``` c++
SimpleAudioEngine::sharedEngine()->preloadEffect("move.wav");
SimpleAudioEngine::sharedEngine()->preloadBackgroundMusic("background.wav");
```

Nothing has changed too.

I compile and run same project on win32 it's work perfectly.

MySystem : Linux Mint 11 x32 (based on Ubuntu 11.04)
cocos2d : latest release from github

minggo:
Did you run CocosDenshionTest?

xal:
Of course, I did.
I got same result, as in first post

minggo:
Ok, I will test it.

godyZ:
Please, could you open a bug in redmine in future? We are in process of migrating all the issues to redmine and github issue won't be available after August.

PR #2091, opened 2013-03-03
fix for SimpleAudioEngine::playEffect() and playBackgroundMusic() playing twice on linux

jsambuo:
On linux, calling SimpleAudioEngine::playEffect() would play the sound effect twice.
SimpleAudioEngine::playBackgroundMusic() also had the same issue.

This should resolve issue #689


minggo:
Thank you.

PR #2092, opened 2013-03-03
Fix all compiler warnings produced by gcc.

sbc100:
This allows the linux and NaCl builds (and perhaps others)
to be compiles with -Wall and -Werror, and it makes the build
output much less noisy.

### DIFF
--- a/CocosDenshion/blackberry/SimpleAudioEngine.cpp
+++ b/CocosDenshion/blackberry/SimpleAudioEngine.cpp
@@ -208,7 +208,6 @@ namespace CocosDenshion
 		ALenum 			format;
 		int 			result;
 		int 			section;
-		int				err;
 		unsigned int 	size = 0;
 
 		if (ov_fopen(pszFilePath, &ogg_file) < 0)
--- a/cocos2dx/platform/nacl/CCApplication.cpp
+++ b/cocos2dx/platform/nacl/CCApplication.cpp
@@ -34,6 +34,7 @@ NS_CC_BEGIN
 CCApplication* CCApplication::sm_pSharedApplication = 0;
 bool CCApplication::s_running = false;
 
+/*
 static long getCurrentMillSecond()
 {
     long lLastTime;
@@ -43,6 +44,7 @@ static long getCurrentMillSecond()
     lLastTime = stCurrentTime.tv_sec * 1000 + stCurrentTime.tv_usec * 0.001;
     return lLastTime;
 }
+*/
 
 CCApplication::CCApplication()
 {
--- a/cocos2dx/platform/nacl/CCFileUtilsNaCl.h
+++ b/cocos2dx/platform/nacl/CCFileUtilsNaCl.h
@@ -44,7 +44,7 @@ private:
     CCFileUtilsNaCl() {}
 public:
     /* override funtions */
-    bool init() {}
+    bool init() { return true; }
     virtual std::string getWritablePath();
     virtual bool isFileExist(const std::string& strFilePath);
 };
--- a/cocos2dx/platform/nacl/CCImage.cpp
+++ b/cocos2dx/platform/nacl/CCImage.cpp
@@ -78,7 +78,8 @@ public:
         iMaxLineHeight = 0;
         //Free all text lines
         size_t size = vLines.size();
-        for (int i=0; i<size; ++i) {
+        for (size_t i=0; i<size; ++i)
+        {
             TextLine line = vLines[i];
             free(line.text);
         }
@@ -330,8 +331,6 @@ public:
     bool getBitmap(const char *text, int nWidth, int nHeight, CCImage::ETextAlign eAlignMask, const char * pFontName, float fontSize)
     {
         FT_Error iError;
-
-        bool bRet = false;
         if (libError)
             return false;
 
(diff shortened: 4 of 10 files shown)


gelldur:
This could be nice.

cocos2dx/layers_scenes_transitions_nodes/CCLayer.h
```diff
@@ -163,8 +163,10 @@ class CC_DLL CCLayer : public CCNode, public CCTouchDelegate, public CCAccelerom
     int  excuteScriptTouchHandler(int nEventType, CCSet *pTouches);
 };
 
+#ifdef __apple__
 #pragma mark -
 #pragma mark CCLayerRGBA
+#endif
```
dumganhar, line 170: I think we just comment `#pragma mark CCLayerRGBA` is ok. 
Don't add this `#ifdef #endif`. :)

samples/Cpp/TestCpp/Classes/SpriteTest/SpriteTest.cpp
```diff
@@ -750,7 +750,7 @@ void SpriteBatchNodeZOrder::reorderSprite(float dt)
         }
     }
     
-    int prev = -1;
+    int CC_UNUSED prev = -1;
     CCArray* children = asmtest->getChildren();
```
dumganhar, line 754: Why `prev` is an unused variable? It's used after the line where it's declared.

samples/Cpp/TestCpp/Classes/TileMapTest/TileMapTest.cpp
```diff
@@ -174,7 +174,7 @@ void TMXOrthoTest::onExit()
     CCTMXTiledMap* map = CCTMXTiledMap::create("TileMaps/orthogonal-test1.tmx");
     addChild(map, 0, kTagTileMap);
 
-    CCSize s = map->getContentSize();
+    ////CCSize s = map->getContentSize();
     ////----UXLOG("ContentSize: %f, %f", s.width,s.height);
```
dumganhar: I think we need to uncomment `UXLOG` and change it to CCLOG, also don't comment `CCSize s = map->getContentSize();`.

scripting/lua/tolua/tolua_map.c
```diff
@@ -729,7 +729,7 @@ TOLUA_API void tolua_array (lua_State* L, const char* name, lua_CFunction get, l
 TOLUA_API void tolua_dobuffer(lua_State* L, char* B, unsigned int size, const char* name) {
 
 #ifdef LUA_VERSION_NUM /* lua 5.1 */
-    luaL_loadbuffer(L, B, size, name) || lua_pcall(L, 0, 0, 0);
+    if (!luaL_loadbuffer(L, B, size, name)) lua_pcall(L, 0, 0, 0);
```
dumganhar, line 732: Which `warning` of this line produces?

cocos2dx/layers_scenes_transitions_nodes/CCLayer.h
sbc100, line 170: Ok.  Should I just remove the pragma completely?   I don't imagine it will have any effect if it is commented, although it seems to be the only "mark" in the entire codebase so it seems unlikely that anyone is actually using it I guess.

samples/Cpp/TestCpp/Classes/SpriteTest/SpriteTest.cpp
sbc100, line 754: This variable is only used in CCAssert statements, which are compiled out in release builds.  Even though its assigned to several times, new versions of GCC will (rightly) warn that its value is never used, so all the assignments are pointless.

I think this is good use case for gcc's unused attribute (i,e, CC_UNUSED).

IIUC only other alternatives would be to globally disable this warning (not a good idea IMHO) or wrap all the use of 'prev' in "ifdef DEBUG".

samples/Cpp/TestCpp/Classes/TileMapTest/TileMapTest.cpp
sbc100: OK. I wasn't clear to me what all this logging was supposed to do, or if it was just something someone had left in there for debugging.

I'll update as you suggest.

How do you guys prefer to get updated patches?  Should I use "git --amend" and then "push -f" to update this request?  Or generate another changelist an git push it?  Or a new pull request after "git --amend"?

scripting/lua/tolua/tolua_map.c
sbc100, line 732: CC obj/release/tolua/tolua_map.o
../tolua/tolua_map.c: In function ‘tolua_dobuffer’:
../tolua/tolua_map.c:732:39: error: value computed is not used [-Werror=unused-value]

Since its computing a boolean value which is the logical 'or' of two expressions, but the result is never used.  I've not seen this pattern before in C code, where logical operation are used on statements out side of conditions.  I think the "if" version is much clearer, and does the same time.

samples/Cpp/TestCpp/Classes/TileMapTest/TileMapTest.cpp
minggo: I think you  can modify codes in the same branch and update again. Then the PR will be updated automatically.

cocos2dx/layers_scenes_transitions_nodes/CCLayer.h
minggo, line 170: Yep, i think it is ok to remove them.

sbc100, line 170: Done

samples/Cpp/TestCpp/Classes/TileMapTest/TileMapTest.cpp
sbc100: Done

minggo:
Thank you.

PR #2093, opened 2013-03-03
Linux build system

sbc100:
This pull request relies on the previous pull request which removes all the warnings
from the code.

This change makes several fundamental improvements to the linux
build system:
- Remove duplication: a lot of the common make logic is now in
  a central shared makefile: cocos2dx/proj.linux/cocos2dx.mk
- Quiet by default: unless you specify V=1 on the commandline
  the output is now very clean and quiet.
- Dependencies: gcc's automatic dependancy generation is now
  used (-MMD).  This means that if you change a header file
  then all the object that depend on it now get rebuilt.
- Don't store objects in the source tree: previously object
  files were stored alongside the sources, which meant that
  debug and release object clobbered each other which leads
  to confusion and lots of unnecessary rebuilding.Q
- Library dependencies: each binary now depends on the libraries
  it links aginst.  This means that if the cocos2dx library
  changes then all examples will get re-linked.
- Top level Makefile: there is now a top level Makefile that
  can be shared/used by any platform the uses make as the
  build system (for example NaCl).  Everything can be built
  by simply running 'make' at the top level. e.g:
  
    make DEBUG=1
    make DEBUG=1 clean

### DIFF
--- a/cocos2dx/proj.nacl/cocos2dx.mk
+++ b/cocos2dx/proj.nacl/cocos2dx.mk
@@ -9,9 +9,9 @@ NACL_ARCH ?= x86_64
 NACL_AR ?= $(NACL_ARCH)-nacl-ar
 NACL_CC ?= $(NACL_ARCH)-nacl-gcc
 NACL_CXX ?= $(NACL_ARCH)-nacl-g++
-CCFLAGS ?= -Wall -Werror
-CXXFLAGS ?= -Wall -Werror
-ARFLAGS ?= cr
+CCFLAGS += -Wall -Werror
+CXXFLAGS += -Wall -Werror
+ARFLAGS = cr
 
 THIS_MAKEFILE := $(CURDIR)/$(word $(words $(MAKEFILE_LIST)),$(MAKEFILE_LIST))
 
@@ -77,6 +77,12 @@ LDFLAGS += -L$(NACLPORTS_ROOT)/lib/$(MULTILIB_SUFFIX)
 # Some cococs sources use #pragma mark
 CCFLAGS += -Wno-unknown-pragmas
 CXXFLAGS += -Wno-unknown-pragmas
+
+ifeq ($(NACL_ARCH),arm)
+# Don't warn about mangling of 'va_list' on arm builds
+CCFLAGS += -Wno-psabi
+CXXFLAGS += -Wno-psabi
+endif
 
 ifdef NACL_MOUNTS
 DEFINES += -DOLD_NACL_MOUNTS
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,5 +1,9 @@
 # Ignore thumbnails created by windows
 Thumbs.db
+<<<<<<< HEAD
+=======
+
+>>>>>>> Improved build system from linux.
```
azmeuk: It seems you had a little conflict here :)

cocos2dx/proj.linux/cocos2dx.mk
```diff
@@ -0,0 +1,107 @@
...
+    -I$(COCOS_SRC)/kazmath/include \
+    -I$(COCOS_SRC)/platform/linux \
+    -I$(COCOS_SRC)/platform/third_party/linux/libpng \
+    -I$(COCOS_SRC)/platform/third_party/linux/libjpeg \
+    -I$(COCOS_SRC)/platform/third_party/linux/libtiff/include \
+    -I$(COCOS_SRC)/platform/third_party/linux/libwebp
+
+LBITS := $(shell getconf LONG_BIT)
+ifeq ($(LBITS),64)
+INCLUDES += -I$(COCOS2DX_PATH)/platform/third_party/linux/include64
```
azmeuk: Shouldn't it be $(COCOS_SRC) instead (l36,38) ?

.gitignore
sbc100: fixed

cocos2dx/proj.linux/cocos2dx.mk
sbc100: fixed

sbc100:
Please take another look

azmeuk:
If you want to get rid of some more warnings, you can try compile the project with clang. It shows me some overloaded-virtual and unneeded-internal-declaration warnings.

minggo:
@sbc100 
Because this PR modify many files, so it should be merged ASAP.
Would you mind that i merge this PR, and you can send another PR for more modification?

minggo:
When i run 

```
make
```

in the root of cocos2d-x.
The error messages are:

```
/bin/sh: 1: g++: not found
make[1]: *** [obj/release/Collision/Shapes/b2ChainShape.o] Error 127
make[1]: Leaving directory `/home/minggo/cocos2d-x/external/Box2D/proj.linux'
make: *** [all] Error 2
```

But g++ exists:

```
minggo@minggo-Parallels-Virtual-Platform:~/cocos2d-x$ g++
The program 'g++' can be found in the following packages:
 * g++
 * pentium-builder
```

What's wrong?

minggo:
Sorry, i forget to run ./make-all-linux-project.sh.

But there is another problem. Because "set -e" is put before checking dependence packages. So it will prevent automatically installing dependence packages.

I think it is better to put "set -e" just before compiling, what's your opinion?

make-all-linux-project.sh
```diff
@@ -1,19 +1,12 @@
 #!/bin/bash
 
+# Exit of first error.
+set -e
```
minggo, line 4: This code will prevent automatically installing dependence packages.
I think it is better to be added before compiling.

sbc100:
Re: g++... I think that g++ really is missing and your Linux distro is telling you how to get.

As far as im concerned this is ready to be merged.  The clang build can be a separate change.  

Re: set -e  ... Not sure why its causing you probs.  Presumably we do want to fail if apt-get fails?  I will take a look later when I am back at my machine.

PR #2100, opened 2013-03-04
getStartLocationInView() and getStartLocation() methods in CCTouch

leszek-s:
Access to first/starting point of a CCTouch for example for easier recognition of custom gestures.


minggo:
Thank you.

PR #2106, opened 2013-03-06
Update unload effect function (android normal)

waiter:
when unload effect, sounID maybe null so the app will crash because of NPE.
On some device , i found the app crashes because sounID is null.
So i think it's better to check whether sounID is null or not.


minggo:
Thank you.

PR #2107, opened 2013-03-06
Use shared NodeLoaderLibrary in CCBReader bindings

SevInf:
In my project, I have native C++ node with JS bindings and CocosBuilder plugin. Currently, there is no way to add custom node loader to CCBReader instance, used in JS bindings - default loader library is always created. This patch replaces creation of new loader library with shared instance, so we can now call:

``` cpp
CCNodeLoaderLibrary::sharedCCNodeLoaderLibrary()->registerCCNodeLoader("MyNode", MyLoader::loader());
```

on app startup and MyLoader and custom node it loads will be available in JS.

### DIFF
--- a/scripting/javascript/bindings/js_bindings_ccbreader.cpp
+++ b/scripting/javascript/bindings/js_bindings_ccbreader.cpp
@@ -274,7 +274,7 @@ JSBool js_cocos2dx_CCBReader_createSceneWithNodeGraphFromFile(JSContext *cx, uin
 JSBool js_CocosBuilder_create(JSContext *cx, uint32_t argc, jsval *vp)
 {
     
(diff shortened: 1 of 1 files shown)


dumganhar:
Hello @SevInf ,
Your change looks good.
I have a question:
If using cocos2d-x jsb and CocosBuilder, do we still need to write `XXXLoader` on native codes?

SevInf:
JS Controller for the most use cases is just enough, so we need to write to custom loaders only when we want to use CocosBuilder for custom nodes, written in C++ (like in my case).

dumganhar:
Got it. I will merge this PR.
BTW, i have fixed the issue you created here(https://github.com/funkaster/cxx-generator/issues/28).
Could you try that?

dumganhar:
Thank you.

PR #2108, opened 2013-03-06
fix a bug in CCControlButton

csy1988525:
fixed a bug that users can't set a button highlighted in the upper logic. also modify some coding format.


dumganhar:
Merged. Thanks.

PR #2123, opened 2013-03-08
Add extensions ios proj

dpull:
Add extensions ios proj

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,9 @@ Developers:
     Rolando Abarca
         Javascript Binding and testjs
 
+    Jimmy Sambuo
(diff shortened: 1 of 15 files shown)


0001-cocos2d-x-extensions-ios-staic-library-proj-file.patch
```diff
@@ -0,0 +1,599 @@
```
dumganhar, line 0: Why did you add `599  0001-cocos2d-x-extensions-ios-staic-library-proj-file.patch` to root folder?

dumganhar, line 0: Could you update this Pull Request?

PR #2124, opened 2013-03-08
Fix package installation in make-all-linux-project

sbc100:
There was a bug where missing packages would cause the script
to exit.

This change also makes several improvements:
- script can now be run from anywhere (not only the cocos root)
- apt-get is called only once to install all missing packages
  (this makes output much more informative on first use)
- added comment/docs to the top of the script


dumganhar:
Thank you.

PR #2134, opened 2013-03-10
Added CCNotificationCenter::removeAllObservers(CCObject *target)

AsadR:
Sometimes it's more convenient to use this call in destructors so that all observers associated with a particular target are removed. This prevents crashes in case a developer forgets to remove all observers for their object before it is destroyed.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ExtensionsTest/NotificationCenterTest/NotificationCenterTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/ExtensionsTest/NotificationCenterTest/NotificationCenterTest.cpp
@@ -128,10 +128,19 @@ NotificationCenterTest::NotificationCenterTest()
     }
 
     CCNotificationCenter::sharedNotificationCenter()->postNotification(MSG_SWITCH_STATE, (CCObject*)(intptr_t)item->getSelectedIndex());
+
+    /* for testing removeAllObservers */
+    CCNotificationCenter::sharedNotificationCenter()->addObserver(this, callfuncO_selector(NotificationCenterTest::doNothing), "random-observer1", NULL);
+    CCNotificationCenter::sharedNotificationCenter()->addObserver(this, callfuncO_selector(NotificationCenterTest::doNothing), "random-observer2", NULL);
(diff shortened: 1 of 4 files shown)


samples/Cpp/TestCpp/Classes/ExtensionsTest/NotificationCenterTest/NotificationCenterTest.cpp
```diff
@@ -128,10 +128,19 @@ void Light::updateLightState()
...
+    CCNotificationCenter::sharedNotificationCenter()->addObserver(this, callfuncO_selector(NotificationCenterTest::doNothing), "random-observer1", NULL);
+    CCNotificationCenter::sharedNotificationCenter()->addObserver(this, callfuncO_selector(NotificationCenterTest::doNothing), "random-observer2", NULL);
+    CCNotificationCenter::sharedNotificationCenter()->addObserver(this, callfuncO_selector(NotificationCenterTest::doNothing), "random-observer3", NULL);
 }
 
 void NotificationCenterTest::toExtensionsMainLayer(cocos2d::CCObject* sender)
 {
+    /* for testing removeAllObservers */
+    int numObserversRemoved = CCNotificationCenter::sharedNotificationCenter()->removeAllObservers(this);
+    CCAssert(numObserversRemoved >= 3, "All observers were not removed!");
```
dumganhar, line 142: `"All observers were not removed!"` ~> `"All observers were removed!"`

AsadR, line 142: The assert will trigger when numObserversRemoved < 3, which means there are still observers leftover and not all of them were removed. In that situation the debug message should say why the assert triggered, so the reason we give for the assert trigger is: "All observers were not removed!".

When all the observers are successfully removed the assertion does not trigger.

dumganhar, line 142: Make sense. Thank you.

PR #2142, opened 2013-03-11
Branch fix cc scale9 sprite anchorpint bug

biteforest:
fix cc scale9 sprite anchorpint bug


biteforest:
ignore this request. it will cause other anchor point problem. there should be another way to solve this problem

biteforest:
I don't why the author force the anchor point is (0.5,0.5), but this will cause problem when set anchor point.
for example, if the developer set the anchor point of ccscale9sprite in cocosbuilder, that won't work.

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/2370. Thanks for your feedback.

PR #2143, opened 2013-03-12
Added Method to create a CCMenuItemToggle from a CCArray with a target selector

CodeSnooker:
I have added a Method to create a CCMenuItemToggle from a CCArray with a target selector.

### DIFF
--- a/cocos2dx/menu_nodes/CCMenuItem.cpp
+++ b/cocos2dx/menu_nodes/CCMenuItem.cpp
@@ -714,6 +714,24 @@ CCArray* CCMenuItemToggle::getSubItems()
     return m_pSubItems;
 }
 
+CCMenuItemToggle * CCMenuItemToggle::createWithTarget(CCObject* target, SEL_MenuHandler selector, CCArray* menuItems)
+{
+    CCMenuItemToggle *pRet = new CCMenuItemToggle();
+    pRet->CCMenuItem::initWithTarget(target, selector);
+    pRet->m_pSubItems = CCArray::create();
+    pRet->m_pSubItems->retain();
+    
+    for (int z=0; z < menuItems->count(); z++)
+    {
+        CCMenuItem* menuItem = (CCMenuItem*)menuItems->objectAtIndex(z);
+        pRet->m_pSubItems->addObject(menuItem);
+    }
+    
+    pRet->m_uSelectedIndex = UINT_MAX;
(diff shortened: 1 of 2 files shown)


shinriyo:
I hope the method. cocos2d-x for iPhone have the method

cocos2dx/menu_nodes/CCMenuItem.cpp
```diff
@@ -714,6 +714,24 @@ CCArray* CCMenuItemToggle::getSubItems()
     return m_pSubItems;
 }
 
+CCMenuItemToggle * CCMenuItemToggle::createWithTarget(CCObject* target, SEL_MenuHandler selector, CCArray* menuItems)
+{
+    CCMenuItemToggle *pRet = new CCMenuItemToggle();
+    pRet->CCMenuItem::initWithTarget(target, selector);
+    pRet->m_pSubItems = CCArray::create();
+    pRet->m_pSubItems->retain();
```
minggo, line 722: Should not determine the return value of pRet and initWithTarget?

cocos2dx/menu_nodes/CCMenuItem.cpp
```diff
@@ -714,6 +714,24 @@ CCArray* CCMenuItemToggle::getSubItems()
...
+    
+    for (int z=0; z < menuItems->count(); z++)
```
minggo, line 724: Should not determine the validation of menuItems?

cocos2dx/menu_nodes/CCMenuItem.cpp
CodeSnooker, line 722: Sorry, I didn't get your statement.

cocos2dx/menu_nodes/CCMenuItem.cpp
CodeSnooker, line 724: Validation can be added but are you sure that we have validation for cases as well while creating a CCMenuItemToggle object? 

I didn't find any.

Check the method body of:

```
bool CCMenuItemToggle::initWithTarget(CCObject* target, SEL_MenuHandler selector, CCMenuItem* item, va_list args)
```

and 

```
CCMenuItemToggle * CCMenuItemToggle::createWithTarget(CCObject* target, SEL_MenuHandler selector, CCMenuItem* item, ...)
```

minggo:
You are right.
Thank you.

PR #2144, opened 2013-03-12
Workaround for some problems due to UITextField of CCEditBoxImplIOS is attached as a view above OpenGL view.

neokim:
Workaround for some problems due to UITextField of CCEditBoxImplIOS is attached as a view above OpenGL view. It has a topmost touch priority higher than any other CCNode objects, and also cannot be veiled by any other CCNode objects.
In this change, UITextField is visible only when the CCEditBox is in edit mode. And in non-edit mode, UITextField is hidden and CCLabelTTF is shown instead with all same attributes at the same place.


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2161. Thanks.

PR #2150, opened 2013-03-12
[Linux] add cache for font family -> font path mapping

Weeds:
As FcFontMach is quite an expensive call the results of getFontFile() are
now cached in an std::map if either a ttf file is packed with the app or
if a font file has been found by using fontconfig

Just loading the starting scene of the Test Application requires about 22.000.000 function calls somewhere in the fontconfig library, consuming about 75% of the time.
Calls to FcFontMatch are now reduced to one call per font family.

### DIFF
--- a/cocos2dx/platform/linux/CCImage.cpp
+++ b/cocos2dx/platform/linux/CCImage.cpp
@@ -21,7 +21,7 @@
 using namespace std;
 
 // as FcFontMatch is quite an expensive call, cache the results of getFontFile
-std::map<std::string, std::string> fontCache;
+static std::map<std::string, std::string> fontCache;
 
 struct LineBreakGlyph {
 	FT_UInt glyphIndex;


minggo:
Thank you.

dumganhar:
@Weeds 
I merged your PR https://github.com/cocos2d/cocos2d-x/pull/2167
But now, this PR can't be automatically merged. 
Could you resolve the conflict? Thanks.

Weeds:
Conflict should be resolved now.

cocos2dx/platform/linux/CCImage.cpp
```diff
@@ -20,6 +20,9 @@
 
 using namespace std;
 
+// as FcFontMatch is quite an expensive call, cache the results of getFontFile
+std::map<std::string, std::string> fontCache;
+
```
dumganhar, line 25: How about making `fontCache` as `static` variable?

Weeds:
You are right, we probably don't need to access the variable from outside of the compilation unit, therefore external linkage is not required.

PR #2159, opened 2013-03-13
Add CCSet::create() function

BodbDearg:
Add CCSet::create() to create, autorelease and return a new empty set.


BodbDearg:
Thank you!

PR #2167, opened 2013-03-14
[Linux] fixed font rendering artifacts, added simple line break

Weeds:
This patch fixed the size calculations for true type fonts (especially for
italic fonts)
Additinally a very simple greedy line break algorithm has been added for
true type labels.


dumganhar:
Thanks.

PR #2171, opened 2013-03-15
modify a memory leak when set a user data to CCNode.

ucchen:
modify a memory leak when set a user data to CCNode.


dumganhar:
@ucchen 
This is not an issue since setUserData is a weak reference function.
Developer has responsibility to delete it outside CCNode.

csy1988525:
I think we should check the UserData when ccnode is released.  not every developer will obey this rule.

dumganhar:
@csy1988525 
When UserData is deleted in the destructor of CCNode, and developer also deletes it outside, it will cause more problems. Furthermore, `who allocates it has responsibility to release it` is  the base rule of CPP Programming.

PR #2173, opened 2013-03-15
Quick fix on make-all-linux-project script

rogerzanoni:
After the introduction of the "set -e" option, the build script fails on non debian/ubuntu linux distributions.

This is a quick fix to allow the users of distros other than debian/ubuntu to build the library, not a definitive solution.
The COCOS2DX_USEAPT environment variable can be set in the user shell config or used directly in
the command line: COCOS2DX_USEAPT=false ./make-all-linux-project.sh

### DIFF
--- a/make-all-linux-project.sh
+++ b/make-all-linux-project.sh
@@ -8,32 +8,36 @@
 # Exit of first error.
 set -e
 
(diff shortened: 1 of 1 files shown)


minggo:
Thank you.

PR #2178, opened 2013-03-16
Newlines in cclog

sbc100:
This is two minor changes:
- fix linux library location.
- strip newlines from log messages before presenting them.

Testing on linux and nacl.

### DIFF
--- a/cocos2dx/proj.linux/cocos2dx.mk
+++ b/cocos2dx/proj.linux/cocos2dx.mk
@@ -10,14 +10,12 @@ DEFINES += -DLINUX
 
 THIS_MAKEFILE := $(CURDIR)/$(word $(words $(MAKEFILE_LIST)),$(MAKEFILE_LIST))
 ifndef COCOS_ROOT
-COCOS_ROOT ?= $(realpath $(dir $(THIS_MAKEFILE))/../..)
-else
-RPATH_REL = ../..
+COCOS_ROOT := $(realpath $(dir $(THIS_MAKEFILE))/../..)
 endif
 COCOS_SRC = $(COCOS_ROOT)/cocos2dx
 OBJ_DIR ?= obj
 
-LIB_DIR = $(COCOS_SRC)/lib/linux
+LIB_DIR = $(COCOS_ROOT)/lib/linux
 BIN_DIR = bin
 
 INCLUDES +=  \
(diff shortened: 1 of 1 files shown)


cocos2dx/proj.linux/cocos2dx.mk
```diff
@@ -10,14 +10,12 @@ DEFINES += -DLINUX
...
-COCOS_ROOT ?= $(realpath $(dir $(THIS_MAKEFILE))/../..)
-else
-RPATH_REL = ../..
+COCOS_ROOT := $(realpath $(dir $(THIS_MAKEFILE))/../..)
 endif
 COCOS_SRC = $(COCOS_ROOT)/cocos2dx
 OBJ_DIR ?= obj
 
-LIB_DIR = $(COCOS_SRC)/lib/linux
+LIB_DIR = $(COCOS_ROOT)/lib/linux
```
minggo, line 18: Is the path right?

sbc100, line 18: I think so... the libraries should live in <root>/lib/ right, not <root>/cocos2dx/lib.  I think I mistakenly changed this when I created the common .mk .. and this changes it back.

minggo, line 18: But the libraries in COCOS_ROOT/cocos2dx/platform/third_party/linux/libraries.

sbc100, line 18: This variable stores the location where the libraries get built to.  The third party folder is for prebuilt platform 
libraries.

sbc100:
Do you like this change now?

cocos2dx/platform/linux/CCCommon.cpp
```diff
@@ -34,15 +34,23 @@ void CCLog(const char * pszFormat, ...)
...
-	fprintf(stderr, "cocos2d-x debug info [%s]\n",  szBuf);
+    // Strip any trailing newlines from log message.
+    size_t len = strlen(szBuf);
+    while (len && szBuf[len-1] == '\n')
+    {
+      szBuf[len-1] = '\0';
+      len--;
+    }
+
+    fprintf(stderr, "cocos2d-x debug info [%s]\n",  szBuf);
```
dumganhar, line 48: Why do we need to strip `\n` for `CCLog` ? 
If developer really wants to output a new line in console, this will break it. Right?

sbc100, line 48: The problem I'm trying to solve here is the the console output on linux is made much harder to read when CCLog callsites includes \n a the end of the string.  The newline is implicitly added by the fprintf in this function so is not needed at the callsite. I'm only stripping trailing newlines not all newlines.

If you think we should leave it to the developer to remember they don't need to add newlines (unlike printf where they need them) then we can remove this part of the change and only update the callsites in cocos2dx code where extraneous newlines were includes?  I'm happy to make this change if you like?

What do you think?

dumganhar, line 48: Make sense. I think your changes is good. Thank you.

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2346.
Thanks.

PR #2192, opened 2013-03-18
[iOS] Using system font by `[UIFont systemFontOfSize]` when font name wasn't found.

dumganhar:


dumganhar:
Close it firstly, i will send a new PR in a different branch.

PR #2193, opened 2013-03-18
added kResolutionFixedHeight and kResolutionFixedWidth resolution policy

Weeds:
These new resolution policies will either ignore the width or height of the
specified design resolution size, but scale the ignored dimension, so it
matches the aspect ratio of the device.

Example:
A device with 854x480 pixels and a design resolution size set to 480x320
and the kResolutionFixedHeight policy, will create an internal canvas of
the size of 570x320px
If the device original size is 800x480 its internal size will be
534x320px
The height for both examples stays the same, the width is adjusted to match
the aspect ratio

Benefits:
- no distortions
- full canvas is usable, the visibility origin is 0/0
- I can use getWinSize() to place objects which is more intuitive
- for objects that should be placed at 0/0 I can use CGPointZero or 0/0 instead
  of the VisibilityRect methods, which is more readable
- using this method projects from the 1.x branch are probably much easier to
  port

Disadvantages:
- it is the developers responsibility to create the game code so that it
  supports multiple aspect ratios

NOTE:
I can only test on Linux or Android, so the code has not been tested on other platforms that are supported by cocos2dx

### DIFF
--- a/cocos2dx/platform/CCEGLViewProtocol.cpp
+++ b/cocos2dx/platform/CCEGLViewProtocol.cpp
@@ -79,6 +79,16 @@ void CCEGLViewProtocol::setDesignResolutionSize(float width, float height, Resol
         m_fScaleX = m_fScaleY = MIN(m_fScaleX, m_fScaleY);
     }
 
+    if ( resolutionPolicy == kResolutionFixedHeight) {
+    	m_fScaleX = m_fScaleY;
+    	m_obDesignResolutionSize.width = ceilf(m_obScreenSize.width/m_fScaleX);
+    }
+
+    if ( resolutionPolicy == kResolutionFixedWidth) {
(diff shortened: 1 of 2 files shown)


dumganhar:
Hi @Weeds ,
Could you update the wiki page here http://www.cocos2d-x.org/projects/cocos2d-x/wiki/Multi_resolution_support ?

Weeds:
Sure, I'll check the page tomorrow.

Weeds:
Page has been updated, I've also added a note that the new resolution policies are not yet available in release builds. If this is merged then we should add a note to the policies that they are available since version 2.1.3 or 2.2.0 or whatever the next release will be called.

dumganhar:
Thanks. :)

dumganhar:
Merged. Thanks.

PR #2207, opened 2013-03-19
fixed a bug when button's parent is hided, the button still can responce to clicked event

csy1988525:
fixed a bug when button's parent is hided, the button still can responce to clicked event

### DIFF
--- a/extensions/GUI/CCControlExtension/CCControlButton.cpp
+++ b/extensions/GUI/CCControlExtension/CCControlButton.cpp
@@ -638,6 +638,14 @@ bool CCControlButton::ccTouchBegan(CCTouch *pTouch, CCEvent *pEvent)
         return false;
     }
     
+    for (CCNode *c = this->m_pParent; c != NULL; c = c->getParent())
+    {
+        if (c->isVisible() == false)
+        {
+            return false;
+        }
+    }
+    
     m_isPushed = true;
     this->setHighlighted(true);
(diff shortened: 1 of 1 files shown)


extensions/GUI/CCControlExtension/CCControlButton.cpp
```diff
@@ -638,6 +638,14 @@ bool CCControlButton::ccTouchBegan(CCTouch *pTouch, CCEvent *pEvent)
         return false;
     }
     
+    for (CCNode *c = this->m_pParent; c != NULL; c = c->getParent())
```
harikrishnar, line 641: cant we use the hasVisibleParents() method present in CCControl class?

csy1988525, line 641: yes you can. just do it

extensions/GUI/CCControlExtension/CCControlButton.cpp
```diff
@@ -638,6 +638,14 @@ bool CCControlButton::ccTouchBegan(CCTouch *pTouch, CCEvent *pEvent)
...
+    {
+        if (c->isVisible() == false)
```
dumganhar, line 643: It's better to change it to `if (!c->isVisible())`

harikrishnar, line 643: this is already fixed now, in a previous request

PR #2210, opened 2013-03-19
Add ability to get the number of scenes in the CCDirector stack.

BodbDearg:
Make a change to Cocos2dx to allow us to reliably remove all scenes on the CCDirector stack except for the root scene. Provide an accessor which returns the current number of scenes in the CCDirector scene stack. We need to know whether or not there is only a root scene (1 scene present) in the stack in other to be able to reliably remove all scenes except the root. Performing the popToRootScene() operation while there is only a root scene present will remove the root and cause the CCDirector to be purged- an event which client code probably wants to avoid in most cases. This change is intended to allow switch over reliably to a completely new scene stack (clearing the old) by allowing client code to pop all scenes except the root, and replace the old root with the new root desired.

### DIFF
--- a/cocos2dx/CCDirector.h
+++ b/cocos2dx/CCDirector.h
@@ -239,6 +239,7 @@ public:
     void replaceScene(CCScene *pScene);
     
     /** Give the number of scenes present in the scene stack.
+     *  Note: this count also includes the root scene node.
      */
     unsigned int getSceneCount();
 


minggo:
If so, should we change the behavior of popToRootScene()?

totallyeviljake:
what we did in cocos2d-xna is to add a method "CanPopScene" which returns true when there is a scene on the stack. Otherwise, we would end up with a blank but active game if the user clicked the back button on the gamepad or soft key.

BodbDearg:
"If so, should we change the behavior of popToRootScene()?"

Ideally the behaviour of popToRootScene() should probably change so that it does nothing when there is just a root scene on the dialog stack. The name implies that it removes every other scene with the exception of the root, so it is a bit unexpected when it sometimes removes the root also. I think it would make much more sense if it always left the root scene in-place on the dialog stack after every call.

I'm wondering what would be the overall impact on current Cocos2dx apps of changing the behaviour? Do any apps rely on the current behaviour- perhaps to purge the CCDirector?

totallyeviljake:
@DarraghCoy I think that's off the topic of this PR. This PR is just to add the scene count property to the CCDirector. I think a better method is just to return a Boolean if a scene can be popped of the stack. I can't think of any useful need for the scene stack depth. 

The following is better suited in its own issue:

Performing the popToRootScene() operation while there is only a root scene present will remove the root and cause the CCDirector to be purged-

That's a clear bug. popToRootScene should not remove the root scene.

BodbDearg:
"what we did in cocos2d-xna is to add a method "CanPopScene" which returns true when there is a scene on the stack. Otherwise, we would end up with a blank but active game if the user clicked the back button on the gamepad or soft key."

Makes sense - kinda the same idea as the scene count. Name of the function might be slightly misleading though in some cases. If there is only the root scene on the stack and you return false in 'CanPopScene()' to indicate not to use 'popToRootScene()' then that could be a bit confusing. Technically there is still a scene on the stack which can be popped - so returning false from 'CanPopScene' does not really make sense in this situation. I think perhaps the scene count would be better  because it can be used however the client code wants. It also provides much more information.

"I can't think of any useful need for the scene stack depth."

Perhaps not in most situations - but I don't think there is any harm in providing this information regardless. Maybe client code might want to manually pop scenes of the stack (one by one) do something in between each pop? I don't think there's any compelling reason to throw away the information and truncate it down to a bool when you could just return the count itself.

"That's a clear bug. popToRootScene should not remove the root scene."

Agreed! I'll make a separate pull request to fix up this issue.

totallyeviljake:
I think your arguments are fair. There is no harm in providing the count. Make sure that you comment/note about the count being inclusive of the root scene.

BodbDearg:
Thanks! Just adjusted the comment for the CCDirector::getSceneCount to clarify that it includes the root scene node.

totallyeviljake:
Thank you for contributing! We greatly appreciate your efforts. Expect @minggo to ask for an extension in the test cases to test out your changes.

BodbDearg:
You're very welcome - thanks for taking the time to look at the changes!

BodbDearg:
Just as a side note - checking out whether we should change CCDirector::popToRootScene or not to behave more as expected. In the Cocos2dx header it's very clearly documented that the current behaviour is intentional:

```
/**Pops out all scenes from the queue until the root scene in the queue.
 * This scene will replace the running one.
 * The running scene will be deleted. If there are no more scenes in the stack the execution is terminated.
 * ONLY call it if there is a running scene.
 */
void popToRootScene(void);
```

It's also the same in Cocos2d iPhone:
https://github.com/cocos2d/cocos2d-iphone/blob/develop-v2/cocos2d/CCDirector.m#L421

Not so sure we can change this now - even though the logic seems a bit odd when there is only a root scene in the stack. The scene count should serve as a workaround though if people don't want to lose the root scene in the stack and end up purging the CCDirector.

minggo:
I have asked @ricardoquesada to answer this question.

BodbDearg:
Thanks minggo!

ricardoquesada:
Yes, I think it is a bug. `popToRootScene` should  do nothing if there is only one scene.

Does it make sense to add this API as well ?

``` objective-c
// the name could be changed
-(void) popToSceneStackLevel:(unsigned int)stackLevel;
```

And the implementation of `popToRootScene` would be:

``` objective-c
-(void) popToRootScene
{
     [self popToSceneStackLevel:1];
}
```

totallyeviljake:
Ricardo makes a better case here - we do want to keep with the original iphone interface, so "the method to "popToSceneStackLevel" should be in the cocos2d-x API. In Cocos2d-XNA we'll name it "PopToSceneStackLevel(int stackLevel)" because are using the .NET formatting rules from Microsoft.

ricardoquesada:
Ok. I will add that API in cocos2d-iphone.

ricardoquesada:
Added in cocos2d-iphone:
https://github.com/cocos2d/cocos2d-iphone/commit/59c5272bffc2c9fb4685ff5665af33d0595dabf1

BodbDearg:
Thanks Ricardo! I think it makes sense to close this pull request now; will submit a separate pull request for implementing the Cocos2d-x version of popToSceneStackLevel() instead. Thanks for taking the time to look into this!

BodbDearg:
The Cocos2dx implementation of popToSceneStackLevel(int level) has now been submitted in the following pull request:
https://github.com/cocos2d/cocos2d-x/pull/2500

totallyeviljake:
we added it to cc2d-xna: https://github.com/totallyevil/cocos2d-xna/issues/143#issuecomment-17327550

slawo:
`getSceneCount` would be very useful in scenes which need to indicate with visual cues that other scenes are available underneath.

I personally use it.

dg-no-9:
is this functionality not added in 2.2 version? I couldn't find it.

PR #2211, opened 2013-03-19
IOS, Added CCEditBoxImplIOS kEditBoxInputFlagSensitive settings.

whitegfx:
Added case to hide IOS autocorrection in kEditBoxInputFlagSensitive
mode.


dumganhar:
Thank you.

PR #2231, opened 2013-03-20
Physics fixes

sbc100:
These changes mostly address issues with CCPhysicsSprite.

See git log for details.


sbc100:
Do you like this change now?

scripting/lua/cocos2dx_support/CCLuaEngine.cpp
```diff
@@ -187,7 +187,8 @@ int CCLuaEngine::executeLayerTouchEvent(CCLayer* pLayer, int eventType, CCTouch
     const CCPoint pt = CCDirector::sharedDirector()->convertToGL(pTouch->getLocationInView());
     m_stack->pushFloat(pt.x);
     m_stack->pushFloat(pt.y);
-    return m_stack->executeFunctionByHandler(nHandler, 3);
+    m_stack->pushInt(pTouch->getID());
+    return m_stack->executeFunctionByHandler(nHandler, 4);
```
dumganhar: `CCLuaEngine::executeLayerTouchEvent` is for `TargettedDelegate`, it means that no multi-touch events are sent.
So is `ID` flag needed?
Do you think that `ID` should only be used for `multi-touch`?
Do we just need add `ID` for `CCLuaEngine::executeLayerTouchesEvent(CCLayer* pLayer, int eventType, CCSet *pTouches)`?

dumganhar: @sbc100 
Any thoughts?

sbc100: I think you might be right.  This change should probably have been a separate PR in any case.

I didn't realise that TargettedDelegate didn't do multi-touch.. I though it was just there so that certain
touches could be filtered out and certain touches accepted.  Are you sure you have more than one touch
accepted by a targeted delegate at once?

I think it might be useful to always pass the touchID through the lua in any case (since it is always available
in C++) but I'll remove this from the PR for now.

dumganhar: > I think it might be useful to always pass the touchID through the lua in any case (since it is always available
> in C++) 

Since it will not break anything, i think that's ok to add an extra argument(id).

> but I'll remove this from the PR for now.

Right, it should not be in this PR.

sbc100: Hmm, yes, it looks like targeted delegates can still do multi-touch because they can "accept" more than one touch at a time.   The difference seems that with the targeted delegate you callbacks for each of your accepted touches rather than a since callback containing a set.

I think this means that the touch ID is always useful/needed in lua and I will create a separate PR for this.

sbc100:
Removed LuaEngine change.

Re-based onto current master.

Please take another look.

scripting/lua/cocos2dx_support/CCLuaEngine.cpp
dumganhar: Oh, it seems that I made a mistake of the meaning of targeted delegate.
Yes, you are right.  :)

dumganhar:
> Please take another look.

My pleasure. :)

dumganhar:
Thanks. Merged.

PR #2158, opened 2013-03-13
Eclipse Project import - Windows part

fape:
Based of #2046, Thanks to @rohankuruvilla
- Add some missing dependency
- Update READ.md-s with windows/cygwin specific part

PR #2239, opened 2013-03-21
Eclipse Project import - Windows part 2

fape:
Based on #2158
- Instead of `-o igncr` use SHELLOPTS on windows
- update README files


minggo:
Thank you.

PR #2243, opened 2013-03-21
Fix linux compilation of CCTexturePVR.cpp

sbc100:
On linux (and possibly other platforms)
GL_COMPRESSED_RGB_PVRTC_2BPPV1_IMG is simply not defined since we don't
include GLES/gl2ext.h.

### DIFF
--- a/cocos2dx/textures/CCTexturePVR.cpp
+++ b/cocos2dx/textures/CCTexturePVR.cpp
@@ -64,6 +64,9 @@ static const ccPVRTexturePixelFormatInfo PVRTableFormats[] = {
 	// 8: LA_88
 	{GL_LUMINANCE_ALPHA, GL_LUMINANCE_ALPHA, GL_UNSIGNED_BYTE, 16, false, true, kCCTexture2DPixelFormat_AI88},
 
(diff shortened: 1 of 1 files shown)


billy1380:
I think the constant is an opengles thing. Getting the same issue on mac build.

sbc100:
Yes, this fix should work on all platforms that don't have this (e.g. all desktop GL platforms).

dumganhar:
@sbc100 ,thank you.
But did you notice that these macros are used in `PVRTableFormats` array?
And `PVRTableFormats` array is used in `v2_pixel_formathash` and `v3_pixel_formathash`.
I think these two arrays, i mean `v2_pixel_formathash` and `v3_pixel_formathash` also need to check whether `GL_COMPRESSED_RGB_PVRTC_2BPPV1_IMG` is defined.
Otherwise, it may cause overflow.

minggo:
Yep, i think @dumganhar is right.
This macro should be applied to v3_pixel_formathash and v2_pixel_formathash too.

sbc100:
Done.

dumganhar:
Thanks.

PR #2245, opened 2013-03-22
Obfuscated js sources can run on JSB correctly now

walzer:
1. remove "require" in scripting/javascript/bindings/js/ files, unify to jsb.js
2. add obfuscate.xml into TestJavascript, the usage is:  ant -buildfile obfuscate.xml. This step will regenerate TestJavascript/game.js
3. Add a target called "TestJavascriptObfuscated" into TestJavascript.xcodeproj. "remove game.js && ant -buildfile obfuscate.xml" was written into a build phase in this target. As the result, in each time you build and run TestJavascriptObfuscated, all js source in cocos2d-js-tests will be obfuscated into game.js, which is the only js file in obfuscated target.

I have only modified the iOS project. Other platforms can do the same.

NOTE: to run the target TestJavascriptObfuscated correctly, these jobs should be done:
1. merge my pull request in cocos2d-js-tests repo https://github.com/cocos2d/cocos2d-js-tests/pull/140
2. move the samples/Javascript/cocos2d-js-test submodule's ref to the merged reference in step 1.


scripting/javascript/bindings/js/jsb_chipmunk.js
```diff
@@ -2,8 +2,6 @@
 // Chipmunk defines
 //
 
-require('jsb_chipmunk_constants.js');
```
dumganhar, line 5: Why did you delete this line?

scripting/javascript/bindings/js/jsb_cocos2d.js
```diff
@@ -2,8 +2,6 @@
 // cocos2d constants
 //
 
-require('jsb_cocos2d_constants.js');
```
dumganhar, line 5: Why did you delete this line?

walzer:
There 2 files are in the obfuscate list, they can't have "require" instruction.
All "require"s are move into a new added jsb.js.
So only the raw js source will require "jsb.js" instead of require everywhere.

PR #2247, opened 2013-03-22
release && set nil

hailongiPhone:
First destroy CCDirector instance with  cocos2d::CCDirector::sharedDirector()->end() 

Then create new CCDirector instance with
cocos2d::CCApplication::sharedApplication()->run();

APP will crash at 
CCApplication::setAnimationInterval

[[CCDirectorCaller sharedDirectorCaller] setAnimationInterval: interval ];

### DIFF
--- a/cocos2dx/platform/ios/CCDirectorCaller.mm
+++ b/cocos2dx/platform/ios/CCDirectorCaller.mm
@@ -51,6 +51,7 @@ static id s_sharedDirectorCaller;
 +(void) destroy
 {
     [s_sharedDirectorCaller release];
+    s_sharedDirectorCaller = nil;
 }
 
 -(void) alloc


cocos2dx/platform/ios/CCDirectorCaller.mm
```diff
@@ -51,6 +51,7 @@ +(id) sharedDirectorCaller
 +(void) destroy
 {
     [s_sharedDirectorCaller release];
+    s_sharedDirectorCaller = nil;
```
minggo, line 54: How about adding this code in dealloc()?
I think it is more reasonable to add it there.

hailongiPhone, line 54: It ‘s OK.
But in my opinion,here we just want to destroy the singleton.
to avoiding dangling pointer errors
release & set nil is the two step what we need to do.
So it is maybe  more sensible to add the code here.

also the other purgeXXX function always do release & set nil

e.g
void CCAnimationCache::purgeSharedAnimationCache(void)
{
    CC_SAFE_RELEASE_NULL(s_pSharedAnimationCache);
}

minggo, line 54: Yep, i think you are right.

PR #2251, opened 2013-03-23
Fix bug: CCControl's touch priority is not set correctly; m_nDefaultTouchPriority is never used;

tiantian20007:
Fix bug: CCControl's touch priority is not set correctly; m_nDefaultTouchPriority is never used;

### DIFF
--- a/extensions/GUI/CCControlExtension/CCControl.cpp
+++ b/extensions/GUI/CCControlExtension/CCControl.cpp
@@ -37,7 +37,6 @@ NS_CC_EXT_BEGIN
 
 CCControl::CCControl()
 : m_bIsOpacityModifyRGB(false)
-, m_nDefaultTouchPriority(0)
 , m_eState(CCControlStateNormal)
 , m_hasVisibleParents(false)
 , m_bEnabled(false)
@@ -76,8 +75,7 @@ bool CCControl::init()
         setHighlighted(false);
 
         // Set the touch dispatcher priority by default to 1
-        setDefaultTouchPriority(1);
-        this->setTouchPriority(m_nDefaultTouchPriority);
+        this->setTouchPriority(1);
         // Initialise the tables
         m_pDispatchTable = new CCDictionary(); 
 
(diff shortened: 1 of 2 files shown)


dumganhar:
So do you think we should remove `(g|s)etDefaultTouchPriority` in `CCControl` class?
It seems to be unused now.

tiantian20007:
Yeah, maybe it's better to remove this duplicate property, I was thinking why this property was added;

dumganhar:
I'm also confused. :(
If you have found out why the author add this property, plz let me know. Thanks.

tiantian20007:
See from previous change log of CCControl.cpp:

``` c++
 void CCControl::registerWithTouchDispatcher()
 {
-    CCDirector::sharedDirector()->getTouchDispatcher()->addTargetedDelegate(this, m_nDefaultTouchPriority, true);
+    CCDirector::sharedDirector()->getTouchDispatcher()->addTargetedDelegate(this, getTouchPriority(), true);
 }
```

then m_nDefaultTouchPriority is never used again;

tiantian20007:
Js binding needs to be update.

dumganhar:
OK, don't worry about jsbinding, it'll be auto generated by jenkins.
Thanks.:)

tiantian20007:
Well, that's awesome to me ！So I don't need to run generateJsBinding.sh manually every time the interface of my c++ code changed after using jenkins?

dumganhar:
If you delete or adding some methods, you should run `tools/tojs/genbindings-win32.bat or genbindings.sh` to test whether your changes could take effect. But the modified codes in the submodule of `scripting/javascript/bindings/generated` should not be committed. And it should be reverted before you pull the latest codes which merge your changes.

You could look at the status of cxx-generator project on jenkins here(http://ci.cocos2d-x.org/)
Every time there is a interface changed, jenkins will pull latest codes, and re-generate binding codes.
Then, send a PR by `cocosRobot` like this (https://github.com/cocos2d/cocos2d-x/pull/2278).

dumganhar:
Thanks. Merged. I will update the AUTHOR file later.

Issue #2241, opened 2013-03-21
WebP library is not universal on mac


billy1380:
There is an issue when attempting to build a project on mac while targeting both x86 and x86_64 i.e. universal binaries.

The linker throws a warning that it will ignore libwebp.a then you get the Undefined symbols for architecture x86_64 messages.

dumganhar:
Yes, all the tests for MAC are set to  32bit target. So we didn't build a library of x86_64 version.

billy1380:
Just in case anyone was interested I have used the framework quite extensively on mac x86_64 and everything builds and runs nicely. Most of the frameworks used e.g. libcurl, libxml2 are universal on mac so none of them have caused any issues. Would you like me to have a look at what it involves to build wepb as a universal binary and try to add it?

dumganhar:
If you could help to generate a universal library of webp, that will be great. :)
You could find the source in my repo(https://github.com/dumganhar/cocos2dx-third-party-libs).
Thanks.

billy1380:
So, I forked the repo and trying to figure out how to get it to build a universal version based on http://www.gnu.org/software/autoconf/manual/autoconf.html#Multiple-Architectures

In the mean time I have found that I can get a prebuilt version of the library using 

``` bash
sudo port install webp +universal
```

I have tested the x86_64 build with it and it works (no errors)

My question is this: is there any reason we can't just use that?

dumganhar:
> My question is this: is there any reason we can't just use that?

I don't know the webp version installed by `port`.
We use the same source of webp to keep the behavior to be the same.
Therefore, i think it's better to build it by ourself. :)

dumganhar:
Also, could you add x64 targets for MAC projects?

billy1380:
I will give it a shot...

billy1380:
I have submitted a pull request, not sure whether this is the best way to solve the problem... but it certainly stops the test app from crashing because the addresses are not being truncated. See what you think.

billy1380:
@dumganhar are you happy for me to close this one?

dumganhar:
I closed it. Thanks.

PR #2252, opened 2013-03-23
Fixes for #2241 universal webp

billy1380:
This pull contains the changes required to build cocos2d, the tests, hellocpp and the xp template for 64 bit or universal.

I have also updated the version of webp built from https://github.com/dumganhar/cocos2dx-third-party-libs


dumganhar:
Thanks.

dumganhar:
Hi @billy1380 ,
Did you notice that `TestCpp/ExtensionsTest/CocosBuilderTest` crashes on OSX if using 64bit target?
But it works ok on 32bit target. Any ideas?

billy1380:
Not noticed it... will get the latest version, have a look and see if I can fix it.

PR #2256, opened 2013-03-25
Add CCRemoveSelf Action

waiter:
Add RemoveSelf Action, so we can simple call CCRemoveSelf to remove the
node from parent.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ActionsTest/ActionsTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/ActionsTest/ActionsTest.cpp
@@ -31,6 +31,7 @@ TESTLAYER_CREATE_FUNC(ActionCallFunc);
 TESTLAYER_CREATE_FUNC(ActionCallFuncND);
 TESTLAYER_CREATE_FUNC(ActionReverseSequence);
 TESTLAYER_CREATE_FUNC(ActionReverseSequence2);
+TESTLAYER_CREATE_FUNC(ActionRemoveSelf);
 TESTLAYER_CREATE_FUNC(ActionOrbit);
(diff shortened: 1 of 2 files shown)


cocos2dx/actions/CCActionInstant.cpp
```diff
@@ -189,6 +189,45 @@ CCObject* CCToggleVisibility::copyWithZone(CCZone *pZone)
...
+		pRet->autorelease();
+	}
+
+	return pRet;
+}
+
+void CCRemoveSelf::update(float time) {
+	CC_UNUSED_PARAM(time);
+	m_pTarget->removeFromParentAndCleanup(true);
+}
```
totallyeviljake, line 213: This action is very useful, but it should not require the cleanup. While the default should be to cleanup, there needs to be a way for the cleanup to be optional. imagine a node that is removed from the graph, but needs to be continuously updated on its selector because it will be re-added later to the graph.

dumganhar, line 213: I agree. I think passing an argument to `create` to determine whether to cleanup the node.

waiter, line 213: I‘ve add an argument. @dumganhar @totallyeviljake

totallyeviljake:
thank you!!

dumganhar:
@waiter 
Could you add a testcase in TestCpp/ActionTest to demonstrate how to use it?

waiter:
@dumganhar 
I've add a testcase. 
Sorry for so late to see your message.

dumganhar:
@waiter 
Thank you. I'll check it.

PR #2259, opened 2013-03-25
CocosPlayer Android update

rohankuruvilla:


dumganhar:
@rohankuruvilla 
Thanks. Merged.
BTW, did you notice that when publishing js app from CocosBuilder to CocosPlayer, it will launch twice when download is finished. Could you help to fix it?

PR #2267, opened 2013-03-26
Added some missing function for setting integer shader uniforms in CCGLProgram

nickveri:
I just added a few functions that were missing from CCGLProgram:
setUniformLocationWith1i - setUniformLocationWith4i

and

setUniformLocationWith2iv - setUniformLocationWith4iv.


minggo:
Thank you.

PR #2268, opened 2013-03-26
fixed potential crash in CCScheduler::removeHashElement

Weeds:
If the scheduler is the last object to have retained the pElements target and
the targets desctructor calls CCScheduler::unschedule() invalid memory is
accessed as pElement->timers is already freed, but the pElement is not removed
yet.
Therefore the call to target->release() should be made after pElement has been
removed from the hash table

### DIFF
--- a/samples/Cpp/TestCpp/Classes/SchedulerTest/SchedulerTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/SchedulerTest/SchedulerTest.cpp
@@ -26,6 +26,7 @@ TESTLAYER_CREATE_FUNC(SchedulerUpdateAndCustom)
 TESTLAYER_CREATE_FUNC(SchedulerUpdateFromCustom)
 TESTLAYER_CREATE_FUNC(RescheduleSelector)
 TESTLAYER_CREATE_FUNC(SchedulerDelayAndRepeat)
+TESTLAYER_CREATE_FUNC(SchedulerIssue2268)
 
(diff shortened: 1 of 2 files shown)


dumganhar:
Hi @Weeds 
I think your changes is right. Thank you.
But could you send me a Demo to reproduce this issue since i couldn't  reproduce this crash?

Weeds:
I could add a test case to the SchedulerTest.
Still you may need to use valgrind or a similar tool to verify that something is broken.
If the memory area that was occupied by the timer array has not been reused, you may not see a crash.

dumganhar:
OK, I'll also check that by using `valgrind`.
When could you add the new test for `SchedulerTest`?

Weeds:
I've added the testcase, it seems that the issue only occurs when using custom selectors.

I know that this code looks wrong:
testNode->release();
testNode->unscheduleAllSelectors();

But as I said in the commit notes, it might not be that easy to tell who has ownership of the objet in real-world applications.
The code would also work (i.e. cause the crash) if zhe testNode retain()/release() calls where to be removed.

I'm not sure on the naming of the TesCase I used Issue 2268 as it is the issue number of the pull request.
If you track this by redmine issues, you might want to open a report there and I'll update the code.

dumganhar:
OK, remind issue http://www.cocos2d-x.org/issues/2003 was created.
Thank you.

dumganhar:
Thanks.

PR #2271, opened 2013-03-27
[Linux] fixed font rendering crash for certain fonts

Weeds:
face->size->metrics->ascender seems to be unreliable for some fonts,
additionally the freetype documentation says that it may be used
differently for different fonts.
Therefore it may happen that the ascender of a font face is less then
the glyphs bounding box resulting in accessing invalid memory.

The fix is to use the bbox attribute instead of the ascender


azmeuk:
Thanks, it fixes my issue!

dumganhar:
Thank you.

PR #2272, opened 2013-03-28
fixed potential crash in CCSaxParser

Weeds:
The character array which is given to CCSAXParser::parse() may not be NULL
terminated.
Therefore we must also add the size of the data array to the parameter list
of the tinyxml2::XMLDocument::parse() call

### DIFF
--- a/cocos2dx/platform/CCSAXParser.cpp
+++ b/cocos2dx/platform/CCSAXParser.cpp
@@ -105,7 +105,7 @@ bool CCSAXParser::init(const char *pszEncoding)
 bool CCSAXParser::parse(const char* pXMLData, unsigned int uDataLength)
 {
 	tinyxml2::XMLDocument tinyDoc;
-	tinyDoc.Parse(pXMLData);
+	tinyDoc.Parse(pXMLData, uDataLength);
 	XmlSaxHander printer;
 	printer.setCCSAXParserImp(this);
 	


dumganhar:
Hi @Weeds 
Do you mean that `tinyxml2` could deal with string without `NULL terminated` ?

Weeds:
It seems like it cannot. The default parameter of the data lenght for tinyxml2::XMLDocument::parse() is -1 which lets tinyxml calculate the data size itself using strlen().
Of course this will expect a NULL terminated string, otherwise it will search for the first NULL byte following the string. (Thereby accessing memory behind the actual char pointer)

I've seen the invalid access using valgrind on Linux by performing the actionTest within the TestApp. When the animation plist is loaded the error occurs.
The issues is triggered as CCFileUtils::readFileData() does not return a NULL terminated string, but the data pointer plus its size (which is correct) However the size parameter was ignored when the data was given to tinyxml

Weeds:
Ok let me clarify that a little:

tinyxml2::XMLDocument::parse() will work with a NULL terminated char array. If the size parameter is omitted it will calculate the size using strlen() -> everything is OK
If the character array is not NULL terminated we must pass the correct size of the array as tinyxml cannot figure out the size of the data array by itself.

dumganhar:
> I've seen the invalid access using valgrind on Linux by performing the actionTest within the TestApp. When the animation plist is loaded the error occurs.

Hi @Weeds ,
I also tested it by using `valgrind`, but `ActionTest` works without errors.
Is my usage of `valgrind` wrong?
What i did is entering `cocos2d-x/samples/Cpp/TestCpp/proj.linux/bin/debug` and executing `valgrind ./TestCpp`.

dumganhar:
And could you tell me which test in `ActionTest` crashed? Thanks.

dumganhar:
Hi @Weeds ,
You are right. I debugged in VS2012. It did exist this issue.
But I still can't use `valgrind` to find out this issue.

PR #2276, opened 2013-03-28
Linux OpenAL support

azmeuk:
Hi.
This branch allows Linux users to use OpenAL instead of FMOD by using make with `OPENAL=1`. There is also a way to disable vorbis support with `NOVORBIS=1`. OpenAL simple audio engine is approximately the blackberry simple audio engine.
Ubuntu users may want to install `libvorbis-dev libopenal-dev libalut-dev` to make it work.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/CocosDenshionTest/CocosDenshionTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/CocosDenshionTest/CocosDenshionTest.cpp
@@ -5,7 +5,7 @@
 // android effect only support ogg
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
     #define EFFECT_FILE        "effect2.ogg"
-#elif( CC_TARGET_PLATFORM == CC_PLATFORM_MARMALADE )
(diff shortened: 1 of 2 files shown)


dumganhar:
@azmeuk 
Thanks for your contribution.
I found an issue that the ogg music can't be played in `CocosDenshionTest`.
You could modify CocosDenshionTest like as follows to reproduce this issue.

<pre>
#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
    #define MUSIC_FILE        "music.mid"
#elif (CC_TARGET_PLATFORM == CC_PLATFORM_BLACKBERRY || CC_TARGET_PLATFORM == CC_PLATFORM_LINUX)
    #define MUSIC_FILE        "background.ogg"
#else
    #define MUSIC_FILE        "background.mp3"
#endif // CC_PLATFORM_WIN32
</pre>

azmeuk:
Hi. This should be fixed now.
Sorry for the multiple commits, I had troubles with git amend :)

dumganhar:
OK, Thank you. I'll check it ASAP. :)

PR #2279, opened 2013-03-29
Fix a display bug in CCScrollView

sjchao:
Fix the display bug when a scrollView nested in another scrollView. The parent's scissor rect need to be considered, when setting the scissor rect in the subScrollView.

### DIFF
--- a/cocos2dx/platform/CCEGLViewProtocol.h
+++ b/cocos2dx/platform/CCEGLViewProtocol.h
@@ -98,12 +98,12 @@ public:
      */
     virtual void setScissorInPoints(float x , float y , float w , float h);
 
-    /*
+    /**
      * Get whether GL_SCISSOR_TEST is enable
      */
     virtual bool isScissorEnabled();
 
-    /*
+    /**
      * Get the current scissor rectangle
      */
     virtual CCRect getScissorRect();


cocos2dx/platform/CCEGLViewProtocol.cpp
```diff
@@ -154,17 +154,33 @@ void CCEGLViewProtocol::setScissorInPoints(float x , float y , float w , float h
...
-    if (pszViewName != NULL && strlen(pszViewName) > 0)
-    {
-        strncpy(m_szViewName, pszViewName, sizeof(m_szViewName));
-    }
-}
-
-const char* CCEGLViewProtocol::getViewName()
-{
-    return m_szViewName;
+bool CCEGLViewProtocol::isScissorEnable()
```
dumganhar: `isScissorEnable`  --> `isScissorEnabled`?

sjchao: `isScissorEnabled` is more fine, but I don't know what should I do in the next step.
Should I change this method name, and then send a new pull request?

dumganhar: No, you just need to push your changes to your remote repo. If the branch is the same, this Pull Request will be auto updated.

extensions/GUI/CCScrollView/CCScrollView.cpp
```diff
@@ -492,11 +492,24 @@ void CCScrollView::beforeDraw()
...
-		CCRect frame = getViewRect();
-        
-        glEnable(GL_SCISSOR_TEST);
-        
-        CCEGLView::sharedOpenGLView()->setScissorInPoints(frame.origin.x, frame.origin.y, frame.size.width, frame.size.height);
+		m_bScissorRestored = false;
+        CCRect frame = getViewRect();
+        if (CCEGLView::sharedOpenGLView()->isScissorEnabled()) {
+            m_bScissorRestored = true;
+            m_tParentScissorRect = CCEGLView::sharedOpenGLView()->getScissorRect();
```
dumganhar, line 499: Hi @sjchao 
Your commit is cool, but did you consider another situation?
That's one scrollView A contains some sub-scrollview B, C, D...
When in the CCScrollView:beforeDraw of C, the method of `m_tParentScissorRect = CCEGLView::sharedOpenGLView()->getScissorRect();` may return the scissor rectangle of B rather than A, Right?

sjchao, line 499: Thank you for your discussion.
I have considered this situation. It works as follow:
1. Scrollview A's method `beforeDraw()` is called firstly. In this method, 
`glEnable(GL_SCISSOR_TEST);`
`CCEGLView::sharedOpenGLView()->setScissorInPoints(frame.origin.x, frame.origin.y, frame.size.width, frame.size.height);`
is called. This will set the scissor rectangle of A as the global scissor rectangle.
2. Assume that the zOrder of sub-scrollview B is less than the C's, so that the B's method `visit()` is called firstly: 
    1) In the B's `beforeDraw()` method, the global scissor rectangle(ie, the scissor rectangle of A) is recorded in member variable `m_tParentScissorRect`. Then the intersection of A's scissor rectangle and B's scissor rectangle is set as the global scissor rectangle.
    2) The `visit()` method of B's subview is called, then B's `draw()` method is called.
    3) B's `afterDraw()` method is called. In this method, A's scissor rectangle(recorded in the member variable `m_tParentScissorRect`) is      restored as the global scissor rectangle. This means that the current scissor rectangle is same as the scissor rectangle before B is visited.
3. C's `visit()` method is called as B. In the `beforeDraw()` method of C, the method `CCEGLView::sharedOpenGLView()->getScissorRect();` will return the scissor rectangle of A, because the A's scissor rectangle is restored in B's `afterDraw()` method.
4. A's other subview is visited, and then A is draw.
4. Finally, A's `afterDraw()` method is called, this will set the `GL_SCISSOR_TEST` disabled.
Am I right?
Sorry about my poor English!

dumganhar, line 499: That makes sense. :+1:

dumganhar:
I'll merge it firstly, and could you also add a test demo to `TestCpp/ExtensionsTest/ScrollViewTest`?
If so ,that will be great.

sjchao:
OK, I will add a test demo tonight or tomorrow.

dumganhar:
> OK, I will add a test demo tonight or tomorrow.

Thanks a lot.

waiter:
bug

@sjchao 
@dumganhar 

I think these is still have a bug.  
Look at the image above. ScrollView A contains ScrollView B. And B have a child (a sprite , the sprite size is big). 
Because B is not intersect with A , so the scissor of B is not enabled. It means B's child is still draw without scissor.
In this case, I think the "area C" is visible! 
So it is the bug.
I think you can use a bool value, like m_isNeedDraw. It is default true , and when a ScrollView is not intersect with its parent ScrollView , m_isNeedDraw will be false.
And in the visit() function , use m_isNeedDraw to decide draw or not.

Sorry about may poor English and image!

PR #2281, opened 2013-03-29
NotificationCenter implementing for js

tiantian20007:
Implement a parallel cc.NotificationCenter for js; When native CCNotificationCenter post a notification, observers in it will be notified; Posting notification from it, observers in native CCNotificationCenter will not be notified;


cocos2dx/script_support/CCScriptSupport.h
```diff
@@ -206,7 +206,7 @@ class CC_DLL CCScriptEngineProtocol
     
     virtual int executeMenuItemEvent(CCMenuItem* pMenuItem) = 0;
     /** Execute a notification event function */
-    virtual int executeNotificationEvent(CCNotificationCenter* pNotificationCenter, const char* pszName) = 0;
+    virtual int executeNotificationEvent(CCNotificationCenter* pNotificationCenter, const char* pszName, CCObject *object) = 0;
```
dumganhar, line 209: Adding new argument will break luabinding. You should also update `cocos2d-x/scripting/lua/cocos2dx_support/CCLuaEngine.h/.cpp`

tiantian20007, line 209: already done;

tiantian20007:
Any suggestion?

scripting/javascript/bindings/js/jsb_cocos2d.js
```diff
@@ -370,6 +370,48 @@ cc.LabelAtlas.create = function( a,b,c,d,e ) {
 
 cc.LayerMultiplex.create = cc.LayerMultiplex.createWithArray;
 
+/** 
+ * NotificationCenter
+ * A pure js object, not associate with native CCNotificationCenter, but a parallel one;
+ * When native CCNotificationCenter post a notification, observers in it will be notified too;
+ * However, posting notification from it, observers in native CCNotificationCenter will not be notified;
+ */
```
dumganhar, line 391: Why don't bind the native CCNotificationCenter directly, like normal native object?

scripting/javascript/bindings/js/jsb_cocos2d.js
```diff
@@ -370,6 +370,48 @@ cc.LabelAtlas.create = function( a,b,c,d,e ) {
 
 cc.LayerMultiplex.create = cc.LayerMultiplex.createWithArray;
 
+/** 
+ * NotificationCenter
+ * A pure js object, not associate with native CCNotificationCenter, but a parallel one;
+ * When native CCNotificationCenter post a notification, observers in it will be notified too;
+ * However, posting notification from it, observers in native CCNotificationCenter will not be notified;
```
dumganhar, line 390: Native NotificationCenter will never notify event to js since `m_scriptHandler` is only used for `LUABinding`.
If you want to use `cc.NotificationCenter`, just copy the implementation of that in cocos2d-html5. It's also pure js and it's compatible with `cocos2d-html5`. 

<pre>
void CCNotificationCenter::postNotification(const char *name, CCObject *object)
{
...
    if (m_scriptHandler) // 
    {
        CCScriptEngineProtocol* engine = CCScriptEngineManager::sharedManager()->getScriptEngine();
        engine->executeNotificationEvent(this, name);
    }
}
</pre>

tiantian20007:
I just simply invoke CCNotificationCenter::sharedNotificationCenter()->registerScriptObserver(1) to make it works. Actually, in my project, c++ is handling all the network stuff (for the current version of js binding is weak for network, although websocket will be supported in the future, not all server for games can be fully implemented with websocket), when c++ got a socket response, I call CCNotificationCenter::sharedNotificationCenter()->postNotification(), through the code in this pull request, cc.NotificationCenter on the js side will get that notification, so I can update the UI written in js. I consider this to be a very common case. 
Why I did not bind the native CCNotificationCenter directly? Because I want avoid the complication and the limit for native CCNotificationCenter: we can only use CCObject js binding object. I just want it to be loose couple. 
I'll take a look at the cc.NotificationCenter implementation in cocos2d-html5.

minggo:
@tiantian20007 
cocos2d-html5 and cocos2d-iphone don't have CCNotificationCenter, and cocos2d-html5, cocos2d-iphone, cocos2d-x should share the same api in js. So i think we can not merge this PR.

PR #2282, opened 2013-03-30
Fix CCFileUtils 'createXXXXWithContentsOfFile' path lookup issue

BodbDearg:
When calling 'CCFileUtils::createCCDictionaryWithContentsOfFile' and 'CCFileUtils::createCCArrayWithContentsOfFile' on iOS/OSX these functions call upon 'CCFileUtils::fullPathForFilename' to resolve the path given into a full path which can be used with system file IO functions. This matches the convention found throughout the cocos2dx library and is expected behaviour. However, on Android and other platforms it appears calling 'CCFileUtils::createCCDictionaryWithContentsOfFile' or 'CCFileUtils:: createCCArrayWithContentsOfFile' does not do the same resolution using 'CCFileUtils::fullPathForFilename' - resulting in file paths which are correctly specified (and which worked on iOS/OSX) to fail to load on these platforms.

Fix this issue by performing a lookup/resolve of the file path using 'CCFileUtils::fullPathForFilename' before doing the low level loading work itself. This brings the behaviour of other platforms in line with iOS and OSX.

### DIFF
--- a/cocos2dx/platform/CCFileUtils.cpp
+++ b/cocos2dx/platform/CCFileUtils.cpp
@@ -314,14 +314,16 @@ public:
 
 CCDictionary* CCFileUtils::createCCDictionaryWithContentsOfFile(const std::string& filename)
 {
+    std::string fullPath = fullPathForFilename(filename.c_str());
     CCDictMaker tMaker;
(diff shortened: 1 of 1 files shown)


dumganhar:
Thank you.

BodbDearg:
Thanks!

PR #2285, opened 2013-04-01
Fix CCUserDefault.cpp compiling on Android. 

BodbDearg:
CCUserDefault.cpp shouldn't compile on Android.


dumganhar:
Thank you.

BodbDearg:
You're welcome - thanks!

PR #2288, opened 2013-04-01
Android: fix emulator for OpenGL ES 2.0

mchinen:
Current HEAD crashes in examples when running android emulator, this fixes it.
See the forum http://www.cocos2d-x.org/boards/6/topics/12563?r=25446#message-25446
Code taken from forum post by Romain TISSERAND

I haven't done many github pull requests, so I botched the "Merge branch mchinensedit" commit message and as a result I think you'll see two duplicate commits.


rtissera:
We use this code in production at DotEmu and I can confirm it does detect emulator properly and does not cause any side issue on "real" Android devices (to our best knowledge). Please merge as it's really useful.
Thanks for creating the pull request mchinen, I was too lazy about this one ;)

minggo:
Thank you.

PR #2289, opened 2013-04-01
print() function in LUA should write to android device log (adb logcat)

DenizPiri:
CCLOG uses __android_log_print internally, this allows lua's print() function
to print to android log output and allows developer to see print output from
LUA using "adb logcat".


minggo:
Thank you.

PR #2291, opened 2013-04-02
fixed #1944: The subclass of cc.Sprite,cc.Layer... doesn't need cc.associateWithNative anymore.

dumganhar:


jdmunro:
Looks good - is it a good idea to commit commented out code?

dumganhar:
@jdmunro 
Yes, I think i should remove the unused codes. Just keep an empty implementation.
Thank you.

scripting/javascript/bindings/js_bindings_opengl.cpp
```diff
@@ -60,6 +60,18 @@ JSBool js_cocos2dx_GLNode_constructor(JSContext *cx, uint32_t argc, jsval *vp)
...
 }
 
+static JSBool js_cocos2dx_GLNode_ctor(JSContext *cx, uint32_t argc, jsval *vp)
+{
+	JSObject *obj = JS_THIS_OBJECT(cx, vp);
+    GLNode *nobj = new GLNode();
+    js_proxy_t* p;
+    JS_NEW_PROXY(p, nobj, obj);
+    nobj->autorelease();
+    JS_AddNamedObjectRoot(cx, &p->obj, "cocos2d::extension::CCBReader");
```
ricardoquesada: Probably it should not say `"cocos2d::extension:CCBReader"`... perhaps it should say something like `"GLNode"`

dumganhar: Sure, it's a `copy` `paste` mistake. I'll correct it. Thanks.

dumganhar:
@ricardoquesada 
Updated.

PR #2295, opened 2013-04-02
CCEditBox iOS enchantments/bugfixes

SevInf:
1. Allow to set fontName and fontSize for placeholder separately. This ability have a few use cases, but most notably it allows to load CCEditBox using CocosBuilder plugin. I'm currently working on one and I'm going to open source plugin and loader when it will be ready.
   Not sure if I should send PR with a plugin to CocosBuilder repo: cocos2d-iphone have no edit box control. Will be glad to know your and @vlidholt opinion on this topic.
2. Correctly resize the subnodes and native text field when CCEditBox's contentSize changes.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ExtensionsTest/EditBoxTest/EditBoxTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/ExtensionsTest/EditBoxTest/EditBoxTest.cpp
@@ -39,10 +39,11 @@ EditBoxTest::EditBoxTest()
     m_pEditName = CCEditBox::create(editBoxSize, CCScale9Sprite::create("extensions/green_edit.png"));
     m_pEditName->setPosition(ccp(visibleOrigin.x+visibleSize.width/2, visibleOrigin.y+visibleSize.height*3/4));
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
-	m_pEditName->setFont("Paint Boy", 25);
+    m_pEditName->setFontName("Paint Boy");
 #else
-	m_pEditName->setFont("fonts/Paint Boy.ttf", 25);
+	m_pEditName->setFontName("fonts/Paint Boy.ttf");
 #endif
+    m_pEditName->setFontSize(25);
     m_pEditName->setFontColor(ccRED);
     m_pEditName->setPlaceHolder("Name:");
     m_pEditName->setPlaceholderFontColor(ccWHITE);


dumganhar:
Hi @SevInf , Your improvement of CCEditBox looks great.
I am not familiar with CocosBuilder's plugin system, so i have no idea about that.
Could you provide some informatations about `Cocosbuilder plugin`? Thank you.

SevInf:
CocosBuilder supports Node plugins that allow to edit custom nodes properties the same way as it supports builtin nodes. My plugin will allow to place and set up CCEditBox using CocosBuilder. Here is a work in progress screenshot of an editor:

editor

CocosBuilder docs have a [guidelines](https://github.com/cocos2d/CocosBuilder/blob/master/Documentation/X1.%20Creating%20Node%20Plug-ins.md#adding-a-plug-in-built-into-cocosbuilder) on adding built in plugins, and text editor control was [requested](https://github.com/cocos2d/CocosBuilder/issues/177). But I'm still not sure whether or not it makes sense to ship CocosBuilder with node not supported by all of the cocos2d versions.

dumganhar:
@SevInf 
It looks perfect.
Cocos2d-html5 now also supports CCEditBox too.
@vlidholt, Are you interested in this feature?

dumganhar:
@SevInf 
When the plugin of CCEditBox in CocosBuilder are ready, please let me know. 
I think many developers need it. Thanks. :)

extensions/GUI/CCEditBox/CCEditBox.h
```diff
@@ -232,6 +244,18 @@ class CCEditBox
...
+	 * Set the placeholder's font name.
+	 * @param pFontName The font name.
+	 */
+	void setPlaceholderFontName(const char* pFontName);
+    
+    /**
+	 * Set the placeholder's font size.
+	 * @param fontSize The font size.
+	 */
+	void setPlaceholderFontSize(int fontSize);
```
dumganhar, line 256: @SevInf 
You have added two new methods and only added the implementation for iOS.
This will break other platforms like Android, win32, mac which supports CCEditBox.
Would you like to also make this change for those platforms?

SevInf:
@dumganhar, it shouldn't break anything. Implementation of this methods still calling full `m_pEditBoxImpl->setFont`. I'll check it later today on Android, but I currently have no Windows PC, so it will be good if someone else check it on Win32.

dumganhar:
If you add some codes that invokes `setFontName` in TestCpp/ExtensionTest/EditBoxTest, there will be some linking error on other platforms. Because they are not implemented for those platforms.

SevInf:
@dumganhar, I've added test for separate name/size setters. It compiles and works successfully on iOS and Android.

dumganhar:
@SevInf 
Yes, you are right. I reviewed the codes. Your changes will not break other platforms. I'll merge it. Thank you.

PR #2296, opened 2013-04-02
JNI fixes, added missing JNIEXPORT / JNICALL declarations

rtissera:
This is needed if compiling with custom visibility settings (example :
-fvisibility=hidden)


dumganhar:
Thanks.

PR #2306, opened 2013-04-03
Clang warnings

azmeuk:
Clang 3.2 reveals some more warnings than gcc. This patch fixes most of them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,23 +1,23 @@
 PLATFORM = linux
 
-define MAKE_TARGET
-	+$(MAKE) -C external/chipmunk/proj.$(PLATFORM) $@
-	+$(MAKE) -C external/Box2D/proj.$(PLATFORM) $@
-	+$(MAKE) -C CocosDenshion/proj.$(PLATFORM) $@
-	+$(MAKE) -C extensions/proj.$(PLATFORM) $@
-	+$(MAKE) -C cocos2dx/proj.$(PLATFORM) $@
-	+$(MAKE) -C scripting/lua/proj.$(PLATFORM) $@
-	+$(MAKE) -C samples/Cpp/HelloCpp/proj.$(PLATFORM) $@
-	+$(MAKE) -C samples/Cpp/TestCpp/proj.$(PLATFORM) $@
-	+$(MAKE) -C samples/Cpp/SimpleGame/proj.$(PLATFORM) $@
-	+$(MAKE) -C samples/Lua/HelloLua/proj.$(PLATFORM) $@
-	+$(MAKE) -C samples/Lua/TestLua/proj.$(PLATFORM) $@
-endef
+MAKE_TARGET =	external/chipmunk/proj.$(PLATFORM)\
+external/Box2D/proj.$(PLATFORM)\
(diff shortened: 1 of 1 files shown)


cocos2dx/actions/CCActionGrid3D.h
```diff
@@ -47,6 +47,7 @@ class CC_DLL CCWaves3D : public CCGrid3DAction
     inline void setAmplitudeRate(float fAmplitudeRate) { m_fAmplitudeRate = fAmplitudeRate; }
 
     /** initializes an action with duration, grid size, waves and amplitude */
+    using CCGrid3DAction::initWithDuration;
     bool initWithDuration(float duration, const CCSize& gridSize, unsigned int waves, float amplitude);
```
dumganhar, line 51: Why do we should add one line `using XXX:virtual_method_name` before virtual function in subclass?
What errors will `clang` produce?

azmeuk:
If you override a virtual method with another one that has a different signature (different arguments or return type), the parent method is "hidden" and produces a `-Woverload-virtual` warning. More details [here](http://stackoverflow.com/questions/4146499/why-does-a-virtual-function-get-hidden#4146525).

cocos2dx/support/ccUTF8.cpp
```diff
@@ -77,6 +78,12 @@ else if ((Char & 0xfe) == 0xfc)        \
 ((Char) < 0x200000 ? 4 :            \
 ((Char) < 0x4000000 ? 5 : 6)))))
 
+unsigned short are always < 0x10000. Test cases that return 4 and more are useless
+*/
+
+#define UTF8_LENGTH(Char)            \
+((Char) < 0x80 ? 1 :                \
+((Char) < 0x800 ? 2 : 3))
```
azmeuk, line 86: Here, clang warns about the fact that an unsigned short is always lesser than 0x10000 (that is 65536). This macro is only used with unsigned short. So testcases for 4 and more seem to be useless.

dumganhar:
@azmeuk
How to compile linux port by using clang?
Could i just modify `CC` and `CXX` variables to `clang++` in Makefile?

azmeuk:
You should edit `CC` and `CXX` to `clang` and `clang++` in `cocos2dx/proj.linux/cocos2dx.mk` and comment `CXXFLAGS += -Wno-unused-but-set-variable` in `external/Box2D/proj.linux/Makefile`.

azmeuk:
I edited the main Makefile so if you don't have to implement every subproject if you add a new platform: make checks if a directory exists before it tries to compiles.

azmeuk:
Why this request is solved by [374f534](https://github.com/cocos2d/cocos2d-x/commit/374f534e745cf2f1beda80128f6d23d1b209fb98) ?

dumganhar:
Hey, this is because that cocos2d-x redmine issue system is associated with cocos2d-x repo on github.
When we need to resolve an issue , we do `fixed #111`, so `111` issue on redmine will be auto closed.
But the github issue system also uses this strategy. Therefore, this issue was closed wrongly.Sorry for that.

PR #2307, opened 2013-04-04
Support variable cell sizes in CCTableView

SevInf:
`tableCellSizeForIndex` method was added to `CCTableViewDataSource`. Old one, `cellSizeForTable` left for backward compatibility and still will be working if `tableCellSizeForIndex` is not overridden. The same is true for JS bindings. 

Performance drop on a large amount of cells is theoretically possible, but I've tested on 20000 cells I didn't saw any noticeable difference.


dumganhar:
@SevInf 
Thanks. Could you add a new test demo in `TestCpp/ExtensionsTest/TableViewTest` to demonstrate how to use this cool feature?

PR #2309, opened 2013-04-05
fix property anchorPoint of CCScale9Sprite 

zhanghaojie:
fix property anchorPoint of CCScale9Sprite doesn't work in ccb. Because initSpriteFrame method reset anchor point


zhanghaojie:
oh!  it seems still not good. setSpriteFrame reset inset property. forget my commit. maybe add a new method or add argument is a good idea

zhanghaojie:
my commit it works for now.  setSpriteFrame is called before setInset method.  but it is dangerous if ccb loader exchange the call sequence

dumganhar:
Hi @zhanghaojie ,
Your change seems great. Thanks for your feedback and bug fix.

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2370. Thanks.

PR #2325, opened 2013-04-08
fixed #2034:make CCUserDefault compatible with previous version on Android and iOS

minggo:


ChrisOrz:
It doesn't work with cocosbuilder v3.0 alpha3.Incompatible ccbi file version .

minggo:
Why?

jimmy0807:
Hi mingo
in CCUserDefault.cpp or CCUserDefaultAndroid.cpp
in function static tinyxml2::XMLElement\* getXMLNodeForKey(const char\* pKey, tinyxml2::XMLDocument **doc)
pXmlBuffer hasn't been released

minggo:
@jimmy808126 
Thank you.

PR #2326, opened 2013-04-08
Major problem with initWithObjects method of CCArray

harikrishnar:
A new unused array was created instead of using the instance on which
the method was called


dumganhar:
Thanks.

PR #2212, opened 2013-03-19
Fix linux build

rogerzanoni:
The cocos library uses libpthread and libGL on linux, so we need to link
with them.

PR #2327, opened 2013-04-08
Issue glEnable in linker under Debian Wheezy

Squallium:
I have just install the new version of Debian (Wheezy) in my system with an ATI Graphic Card, and when I launch the make-all-linux-project.sh script I get the following error:

<pre>
/usr/bin/ld: obj/debug/Classes/ClippingNodeTest/ClippingNodeTest.o: undefined reference to symbol 'glEnable'
/usr/bin/ld: note: 'glEnable' is defined in DSO /usr/lib/libGL.so.1 so try adding it to the linker command line
/usr/lib/libGL.so.1: could not read symbols: Invalid operation
make[1]: *** [bin/debug/TestCpp] Error 1
make[1]: se sale del directorio `/home/borja/Programas/cocos2d-2.1rc0-x-2.1.2/samples/Cpp/TestCpp/proj.linux'
make: *** [all] Error 2
</pre>

I fixed adding in the linker options of the cocos2dx.mk the "-lGL" flag.

### DIFF
--- a/cocos2dx/proj.linux/cocos2dx.mk
+++ b/cocos2dx/proj.linux/cocos2dx.mk
@@ -91,7 +91,7 @@ SHAREDLIBS += -lfmodex
 endif
 endif
 
-SHAREDLIBS += -lglfw -lGLEW -lfontconfig
+SHAREDLIBS += -lglfw -lGLEW -lfontconfig -lGL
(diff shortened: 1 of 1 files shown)


azmeuk:
See #2212 :)

Squallium:
azmeuk, thanks for the related issue :-) . I didn't see it when I was investigated.

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/2212.
Thank you.

Issue #2304, opened 2013-04-03
[Linux] CCFileUtils::init() - Buffer size for readlink


rafaelx:
The buffer size that is passed to readlink should be reduced by one for the string terminating zero. Additionally, the result of readlink should be verified.

char fullpath[1024] = {0};
ssize_t length = readlink("/proc/self/exe", fullpath, sizeof(fullpath)-1);
if (length <= 0) {
     // react somehow on the situation:
     // For example, write on the console via CCLog and return
}

// the additional string terminating zero needs also some space within the buffer
fullpath[length] = '\0';

rafaelx:
Closed, since pull request #2328 has been merged in.

PR #2328, opened 2013-04-08
Correction of passed buffer size to readlink ...

rafaelx:
... and verification of returned result by readlink for issue #2304


dumganhar:
Thanks.

PR #2340, opened 2013-04-09
Small fixes to CocosPlayer

rohankuruvilla:
1. Fixing AppDelegate to include OpenGL bindings
2. Adding orientation change JNI stubs
3. Commenting out cache clean on receiving new zip file (was causing consistency issues)
4. Fixing bug where "Connected!" message was displayed on a different thread


dumganhar:
Thanks.

PR #2343, opened 2013-04-10
Linux: Set writable path to "<user's home>/.config/<app name>"

Yodesoft:


dumganhar:
`This pull request cannot be automatically merged.`
Please pull our latest codes.

Yodesoft:
Do i need to open a new pull request?

minggo:
It is ok using the same pull request.
Why should change writable path to <user's home>/.config/<app name>?
It will break compatibility.

Yodesoft:
Write to ~/.config/ is more clearer than write to ~/

dumganhar:
@Yodesoft 
But now the writable path is in application's path.

Yodesoft:
App's path is not writable for all users.

dumganhar:
@Yodesoft 
Make sense, I'll merge it. Thanks.

cocos2dx/platform/linux/CCFileUtilsLinux.cpp
```diff
@@ -33,27 +33,27 @@ CCFileUtils* CCFileUtils::sharedFileUtils()
...
+    stat(path.c_str(), &st);
+    if (!S_ISDIR(st.st_mode)) {
+        mkdir(path.c_str(), 0744);
     }
-
-    fullpath[length] = '\0';
-    std::string resourcePath = fullpath;
-    resourcePath = resourcePath.substr(0, resourcePath.find_last_of("/"));
-    resourcePath += "/../../../Resources/";
-    m_strDefaultResRootPath = resourcePath;
```
dumganhar, line 48: You deleted these lines, it will lead to resource searching failure.

dumganhar, line 48: It's easy to reproduce this issue if you run `TestCpp`.

Yodesoft:
Sorry, it's my failure.
Those codes had restored.

Weeds:
You should not use $_ to access the application name.
$_ Gives the last argument to the previous command. (whatever this means)
This will only work if you start the application from the command line (without a path)

So for example if you start your app (called TestApp) by typing TestApp into the command line the output is OK, if you type /somedir/TestApp $_ will contain the typed path (i.e. /somedir/TestApp) if I click on the binary using KDE $_ will expand to /usr/lib64/kde4/libexec/start_kdeinit_wrapper

You also may want to check if $XDG_CONFIG_HOME does exist.
See http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

Yodesoft:
Got it. Thanks @Weeds

dumganhar:
Thanks, @Yodesoft ,@Weeds .

PR #2349, opened 2013-04-10
Fix linux build - Box2DTestbed

rogerzanoni:
Compiling on linux causes the error "array subscript is above array
bounds [-Werror=array-bounds]". This commit adds boundary checking
using the constant defining the maximum number of items in the "nuke" array.


dumganhar:
Hi @rogerzanoni ,
How to reproduce this issue? I couldn't reproduce it on my UBUNTU 1210 64bit.

belak:
This warning was added in a newer version of gcc. I can confirm this exists on ArchLinux with gcc-4.8.0.

belak:
This might be relevant: http://gcc.gnu.org/bugzilla/show_bug.cgi?id=55085

dumganhar:
Thanks, @belak .
I'll update my gcc compiler and have a try.

dumganhar:
Thanks.

PR #2350, opened 2013-04-10
Add support for cookies to HttpClient, and support for getting headers from HttpResponse

randyv2zhurosoft:
Add support for cookies to HttpClient. This relies on FileUtils being able to access a file to store the cookies themselves between instances of CURL.

This also adds support for getting headers from HttpResponse, in case someone wants to read those in addition to response data (e.g. to handle cookies manually).

### DIFF
--- a/extensions/network/HttpClient.cpp
+++ b/extensions/network/HttpClient.cpp
@@ -30,7 +30,7 @@
 #include <pthread.h>
 #include <semaphore.h>
 #include <errno.h>
-
+#include "CCFileUtils.h"
 #include "curl/curl.h"
 
 NS_CC_EXT_BEGIN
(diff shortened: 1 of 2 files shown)


dumganhar:
Does it need to save cookie? Why do we need to use cookie in games? Could you give me a hint?

extensions/network/HttpClient.cpp
```diff
@@ -68,6 +68,7 @@
 
 typedef size_t (*write_callback)(void *ptr, size_t size, size_t nmemb, void *stream);
 
+static const char *cookieFilename = (CCFileUtils::sharedFileUtils()->getWritablePath() + "cookieFile.txt").c_str();
```
dumganhar, line 71: Should we write hard code of the path of cookie file?
I think providing a method to set the path is more elegant.

hzd822:
great! cookie is need for some 3rd  channel in android.

randyv2zhurosoft:
Cookies are often used in games which have a server login component (e.g. to store a session cookie).

The path of the cookie file can be an option, though typically it seems people don't care what the path is and just want cookies to work.

randyv2zhurosoft:
It looks like the latest version of CCHttpRequest.cpp has changed, so I'll submit a new pull request.

PR #2351, opened 2013-04-10
Adding popSceneBack() function to CCDirector

mdkcore0:
This function deletes the running scene and returns the actual scene
back to the caller. This way, we can simply use it to perform a
transition animation, like this:

```
CCDirector::sharedDirector()->replaceScene(
        CCTransitionSlideInL::create(0.5,
        CCDirector::sharedDirector()->popSceneBack()));
```

### DIFF
--- a/cocos2dx/CCDirector.cpp
+++ b/cocos2dx/CCDirector.cpp
@@ -625,6 +625,13 @@ void Director::popScene(void)
     }
 }
 
+Scene* Director::popSceneBack(void)
+{
+    popScene();
+
+    return _nextScene;
+}
+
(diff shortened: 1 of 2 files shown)


michaelcontento:
:+1:

minggo:
If there is one scene, popSceneBack will cause CCDirector::end() to be invoked.

mdkcore0:
Hi @minggo , it's the same behavior of popScene(); do you have another suggestion of behavior for this situation??

mdkcore0:
Closing this PR in favor of a new one (with tests and over the 'develop' branch; PR #3244)

PR #2352, opened 2013-04-10
Multiprocessor Compilation on Windows

lgb:
I just enable subj for all VC++ projects. Also I set Minimal Rebuid to default (off) for all debug configurations because conflicts with Minimal Rebuild and Multiprocessor Compilation.

It's two changes decrease times of build from 4-5 minutes to 1.5


dumganhar:
> It's two changes decrease times of build from 4-5 minutes to 1.5

Which project ? Did you mean the time of building the whole solution?

lgb:
build-win32.bat compile entire solution, I check with it.

dumganhar:
@lgb 
OK, thank you. I'll also check it by using that script.

rtissera:
Very useful change, VS2012 tends to compile slowly without this option.
Could you merge it please ?

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/2976 .

PR #2356, opened 2013-04-11
Fix variable initialization in SpriteBatchNode and LabelBMFont to allow JS extension

MarkEHenderson:
Initialize a few variables to null when SpriteBatchNode and LabelBMFont are constructed.

Without this, creating then destroying a LabelBMFont without calling its init method will result in a crash (about 90% of time on android, about 10% of time on IOS).

The extend function in the JS bindings will create an instance of each class that is extended, so this checkin is needed to allow JS level LabelBMFont extension.


cocos2dx/label_nodes/CCLabelBMFont.cpp
```diff
@@ -536,6 +536,7 @@ bool CCLabelBMFont::initWithString(const char *theString, const char *fntFile, f
 , m_pAlignment(kCCTextAlignmentCenter)
 , m_fWidth(-1.0f)
 , m_pConfiguration(NULL)
+, m_pReusedChar(NULL)
 , m_bLineBreakWithoutSpaces(false)
 , m_tImageOffset(CCPointZero)
 , m_pReusedChar(NULL)
```
dumganhar, line 541: `m_pReusedChar` is already initialized behind `m_tImageOffset`.

dumganhar:
Thanks.

PR #2361, opened 2013-04-13
Implement CCDictionary::writeToFile();

tiantian20007:
Use NSDictionay to do the job for iOS and Mac, and using tinyxml2 for other platform;

### DIFF
--- a/cocos2dx/platform/CCFileUtils.cpp
+++ b/cocos2dx/platform/CCFileUtils.cpp
@@ -338,15 +338,17 @@ static tinyxml2::XMLElement* generateElementForDict(cocos2d::CCDictionary *dict,
  */
 bool CCFileUtils::writeToFile(cocos2d::CCDictionary *dict, const std::string &fullPath)
 {
-    CCLOG("tinyxml2 CCDictionary %d writeToFile %s", dict->m_uID, fullPath.c_str());
-    bool bRet = false;
+    //CCLOG("tinyxml2 CCDictionary %d writeToFile %s", dict->m_uID, fullPath.c_str());
     tinyxml2::XMLDocument *pDoc = new tinyxml2::XMLDocument();
     if (NULL == pDoc)
         return false;
     
     tinyxml2::XMLDeclaration *pDeclaration = pDoc->NewDeclaration("xml version=\"1.0\" encoding=\"UTF-8\"");
     if (NULL == pDeclaration)
+    {
+        delete pDoc;
         return false;
+    }
     
     pDoc->LinkEndChild(pDeclaration);
     tinyxml2::XMLElement *docType = pDoc->NewElement("!DOCTYPE plist PUBLIC \"-//Apple//DTD PLIST 1.0//EN\" \"http://www.apple.com/DTDs/PropertyList-1.0.dtd\"");
@@ -355,19 +357,23 @@ bool CCFileUtils::writeToFile(cocos2d::CCDictionary *dict, const std::string &fu
     tinyxml2::XMLElement *pRootEle = pDoc->NewElement("plist");
     pRootEle->SetAttribute("version", "1.0");
     if (NULL == pRootEle)
+    {
+        delete pDoc;
         return false;
+    }
     pDoc->LinkEndChild(pRootEle);
     
     tinyxml2::XMLElement *innerDict = generateElementForDict(dict, pDoc);
     if (NULL == innerDict )
+    {
+        delete pDoc;
         return false;
+    }
     pRootEle->LinkEndChild(innerDict);
     
-    bRet = tinyxml2::XML_SUCCESS == pDoc->SaveFile(fullPath.c_str());
+    bool bRet = tinyxml2::XML_SUCCESS == pDoc->SaveFile(fullPath.c_str());
     
-    if(pDoc)
-        delete pDoc;
-    
+    delete pDoc;
     return bRet;
 }
 
@@ -376,34 +382,25 @@ bool CCFileUtils::writeToFile(cocos2d::CCDictionary *dict, const std::string &fu
  */
 static tinyxml2::XMLElement* generateElementForObject(cocos2d::CCObject *object, tinyxml2::XMLDocument *pDoc)
 {
-    tinyxml2::XMLElement* rootNode = NULL;
     // object is CCString
-    CCString *str = dynamic_cast<CCString *>(object);
-    if (str)
+    if (CCString *str = dynamic_cast<CCString *>(object))
     {
-        rootNode = pDoc->NewElement("string");
+        tinyxml2::XMLElement* node = pDoc->NewElement("string");
         tinyxml2::XMLText* content = pDoc->NewText(str->getCString());
-        rootNode->LinkEndChild(content);
-        return rootNode;
+        node->LinkEndChild(content);
+        return node;
     }
     
     // object is CCArray
-    CCArray *array = dynamic_cast<CCArray *>(object);
-    if (array)
-    {
-        rootNode = generateElementForArray(array, pDoc);
-        return rootNode;
-    }
+    if (CCArray *array = dynamic_cast<CCArray *>(object))
+        return generateElementForArray(array, pDoc);
     
     // object is CCDictionary
-    CCDictionary *innerDict = dynamic_cast<CCDictionary *>(object);
-    if (innerDict)
-    {
-        rootNode = generateElementForDict(innerDict, pDoc);
-        return rootNode;
-    }
+    if (CCDictionary *innerDict = dynamic_cast<CCDictionary *>(object))
+        return generateElementForDict(innerDict, pDoc);
     
-    return rootNode;
+    CCLOG("This type cannot appear in property list");
+    return NULL;
 }
 
 /*
--- a/cocos2dx/platform/CCFileUtils.h
+++ b/cocos2dx/platform/CCFileUtils.h
@@ -352,9 +352,6 @@ protected:
      */
     virtual bool writeToFile(CCDictionary *dict, const std::string& fullPath);
     
-    // for test
-    //virtual bool writeToFileTest(CCDictionary *dict, const std::string& fullPath);
-    
     /**
      *  Creates an array by the contents of a file.
      *  @note This method is used internally.
--- a/cocos2dx/platform/ios/CCFileUtilsIOS.mm
+++ b/cocos2dx/platform/ios/CCFileUtilsIOS.mm
@@ -308,7 +308,7 @@ CCDictionary* CCFileUtilsIOS::createCCDictionaryWithContentsOfFile(const std::st
 
 bool CCFileUtilsIOS::writeToFile(CCDictionary *dict, const std::string &fullPath)
 {
-    CCLOG("iOS||Mac CCDictionary %d write to file %s", dict->m_uID, fullPath.c_str());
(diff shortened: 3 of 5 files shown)


cocos2dx/platform/ios/CCFileUtilsIOS.mm
```diff
@@ -234,6 +306,24 @@ static void addValueToCCDict(id key, id value, CCDictionary* pDict)
     return pRet;
 }
 
+bool CCFileUtilsIOS::writeToFile(CCDictionary *dict, const std::string &fullPath)
+{
+    CCLog("iOS||Mac CCDictionary %d write to file %s", dict->m_uID, fullPath.c_str());
```
dumganhar: It's better to use `CCLOG`.

cocos2dx/platform/mac/CCFileUtilsMac.mm
```diff
@@ -231,6 +303,22 @@ static void addValueToCCDict(id key, id value, CCDictionary* pDict)
     return pRet;
 }
 
+bool CCFileUtilsMac::writeToFile(CCDictionary *dict, const std::string &fullPath)
+{
+    CCLog("iOS||Mac CCDictionary %d write to file %s", dict->m_uID, fullPath.c_str());
```
dumganhar: It's better to use `CCLOG`.

dumganhar:
Looks great. :+1:

dumganhar:
Do you think CCArray also need this function? Like `CCArray::writeToFile`?

scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -736,8 +736,55 @@ int ScriptingCore::executeMenuItemEvent(CCMenuItem* pMenuItem)
...
+    jsval cc;
+    JSContext *cx = this->cx_;
+    JSObject *global = this->global_;
+    // get global var cc
+    JS_GetProperty(cx, global, "cc", &cc);
+    if (cc == JSVAL_VOID)
+        return 0;
+    
+    // get cc.NotificationCenter: a pure js object
+    jsval owner;
```
dumganhar: Why this Pull Request contains some commits in https://github.com/cocos2d/cocos2d-x/pull/2281 ?

tiantian20007: Sorry, wrong operation... I'll fix it;

tiantian20007:
Fixed. I merged the wrong branch on my local repository and pushed it ...
I think most case we are handling .plist file which is naturally coupled with CCDictionary.

cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -326,16 +327,89 @@ CCArray* CCFileUtils::createCCArrayWithContentsOfFile(const std::string& filenam
...
+        if (str)
+        {
+            tinyxml2::XMLElement* tmpNode = pDoc->NewElement("string");
+            rootNode->LinkEndChild(tmpNode);
+            tinyxml2::XMLText* content = pDoc->NewText(str->getCString());
+            tmpNode->LinkEndChild(content);
+            continue;
+        }
+        
+        CCDictionary *innerDict = dynamic_cast<CCDictionary *>(object);
```
dumganhar: We should also consider CCDictionary contains CCArray elements.

minggo:
@tiantian20007 
Why implement the logical in CCFileUtils?
I think it is better to implement in CCDictionary, right?
And CCFileUtils just save data to a file, it doesn't care about what the data is.

cocos2dx/platform/CCFileUtils.cpp
tiantian20007: Good point, I'll add that.

tiantian20007:
Because CCDictionary::createWithContentsOfFile is also implemented in CCFileUtils. 
I think it make sense considering we already have CCFileUtilsIOS, CCFileUtilsMac and CCFileUtilsAndroid to do the platform dependent jobs. And we don't have CCDictionaryIOS or CCDictionaryAndroid.

minggo:
I am not sure, because it will make CCFileUtils too complicated.

cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -326,11 +327,132 @@ CCArray* CCFileUtils::createCCArrayWithContentsOfFile(const std::string& filenam
...
+ */
+static tinyxml2::XMLElement* generateElementForArray(cocos2d::CCArray *array, tinyxml2::XMLDocument *pDoc);
+static tinyxml2::XMLElement* generateElementForDict(cocos2d::CCDictionary *dict, tinyxml2::XMLDocument *pDoc);
+
+/*
+ * Use tinyxml2 to write plist files
+ */
+bool CCFileUtils::writeToFile(cocos2d::CCDictionary *dict, const std::string &fullPath)
+{
+    CCLOG("tinyxml2 CCDictionary %d writeToFile %s", dict->m_uID, fullPath.c_str());
```
sergey-shambir: Please, don't spam to logs in upstream version

cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -326,11 +327,132 @@ CCArray* CCFileUtils::createCCArrayWithContentsOfFile(const std::string& filenam
...
+    pDoc->LinkEndChild(pRootEle);
+    
+    tinyxml2::XMLElement *innerDict = generateElementForDict(dict, pDoc);
+    if (NULL == innerDict )
+        return false;
+    pRootEle->LinkEndChild(innerDict);
+    
+    bRet = tinyxml2::XML_SUCCESS == pDoc->SaveFile(fullPath.c_str());
+    
+    if(pDoc)
```
sergey-shambir: pDoc not deleted if any created element is null. Can you create RAII wrapper to ensure that XMLDocument will be deleted? Probably std::auto_ptr is a good choice.

Also there no need to declare 'bool bRet' on top, move it to 'bool bRet = tynyxml::XML_SUCCESS == ...'

cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -326,11 +327,132 @@ CCArray* CCFileUtils::createCCArrayWithContentsOfFile(const std::string& filenam
...
+}
+
+/*
+ * Generate tinyxml2::XMLElement for CCObject through a tinyxml2::XMLDocument
+ */
+static tinyxml2::XMLElement* generateElementForObject(cocos2d::CCObject *object, tinyxml2::XMLDocument *pDoc)
+{
+    tinyxml2::XMLElement* rootNode = NULL;
+    // object is CCString
+    CCString *str = dynamic_cast<CCString *>(object);
```
sergey-shambir: You can use 'if (CCString *str = dynamic_cast<CCString *>(object)) { ... }'. I.e. declaration inside if condition.

cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -326,11 +327,132 @@ CCArray* CCFileUtils::createCCArrayWithContentsOfFile(const std::string& filenam
...
+    
+    // object is CCDictionary
+    CCDictionary *innerDict = dynamic_cast<CCDictionary *>(object);
+    if (innerDict)
+    {
+        rootNode = generateElementForDict(innerDict, pDoc);
+        return rootNode;
+    }
+    
+    return rootNode;
```
sergey-shambir: Please don't just return NULL, add CCLOG to warn "This type cannot appear in property list". Actually, series of dynamic_cast isn't good solution, but it is only way until #2401 will be merged (i'll update code when both requests will be merged).

Style note: no need to declare tinyxml2::XMLElement\* rootNode for whole function.

cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -326,11 +327,132 @@ CCArray* CCFileUtils::createCCArrayWithContentsOfFile(const std::string& filenam
...
+            rootNode->LinkEndChild(element);
+    }
+    return rootNode;
+}
+
+
 #else
 NS_CC_BEGIN
 
 /* The subclass CCFileUtilsIOS and CCFileUtilsMac should override these two method. */
```
sergey-shambir, line 450: these two method -> these three methods, of course if you will not move it outside CCFileUtils

cocos2dx/platform/CCFileUtils.h
```diff
@@ -347,6 +347,15 @@ class CC_DLL CCFileUtils : public TypeInfo
     virtual CCDictionary* createCCDictionaryWithContentsOfFile(const std::string& filename);
     
     /**
+     *  Write a dictionary to a plist file.
+     *  @note This method is used internally.
+     */
+    virtual bool writeToFile(CCDictionary *dict, const std::string& fullPath);
+    
+    // for test
```
sergey-shambir: remove test code

cocos2dx/platform/ios/CCFileUtilsIOS.mm
```diff
@@ -86,6 +87,40 @@ static void addItemToCCArray(id item, CCArray *pArray)
     }
 }
 
+static void addCCObjectToNSArray(CCObject *object, NSMutableArray *array)
+{
+    // add string into array
+    if (CCString *ccString = dynamic_cast<CCString *>(object)) {
+        NSString *strElement = [NSString stringWithCString:ccString->getCString() encoding:NSUTF8StringEncoding];
```
sergey-shambir, line 94: Replace with stringWithUTF8String, since stringWithCString is deprecated.

cocos2dx/platform/ios/CCFileUtilsIOS.mm
```diff
@@ -136,6 +171,43 @@ static void addValueToCCDict(id key, id value, CCDictionary* pDict)
     }
 }
 
+static void addCCObjectToNSDict(const char * key, CCObject* object, NSMutableDictionary *dict)
```
sergey-shambir, line 174: You can create unified method getNSPropertyObjectFromCCObject() to simplify addCCObjectToNSDict and addCCObjectToNSArray

cocos2dx/platform/ios/CCFileUtilsIOS.mm
```diff
@@ -244,6 +316,24 @@ static void addValueToCCDict(id key, id value, CCDictionary* pDict)
     return pRet;
 }
 
+bool CCFileUtilsIOS::writeToFile(CCDictionary *dict, const std::string &fullPath)
+{
+    CCLOG("iOS||Mac CCDictionary %d write to file %s", dict->m_uID, fullPath.c_str());
```
sergey-shambir: Don't spam please

tiantian20007:
Thx for reviewing. I'll modify that.

cocos2dx/platform/ios/CCFileUtilsIOS.mm
tiantian20007, line 94: [NSString stringWithCString:] is deprecated, however, [NSString  stringWithCString: encoding:] is not.
https://developer.apple.com/library/mac/#documentation/Cocoa/Reference/Foundation/Classes/NSString_Class/Reference/NSString.html#//apple_ref/occ/clm/NSString/stringWithCString:encoding:

cocos2dx/platform/CCFileUtils.cpp
tiantian20007: I'll try your visitor class when it is merged; Thx again;

PR #2372, opened 2013-04-15
Remove Swap Alpha Operate in the nativeInitBitmapDC

Yodesoft:
Swap Alpha in the nativeInitBitmapDC is not necessary. 
Remove those codes to get a great performance improvement.

### DIFF
--- a/cocos2dx/platform/CCFileUtils.cpp
+++ b/cocos2dx/platform/CCFileUtils.cpp
@@ -30,6 +30,9 @@ THE SOFTWARE.
 #include "support/tinyxml2/tinyxml2.h"
 #include "support/zip_support/unzip.h"
 #include <stack>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
 
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


dumganhar:
@Yodesoft 
Why isn't the swap alpha operation needed?

Yodesoft:
Each value of ARGB is the same value, so we don't need to swap alpha.

minggo:
Now the value passed into opengl is RGBA. So should swap alpha.

Yodesoft:
ARGB to RGBA is correct way, but each value of the data come from the canvas.drawText is as below:
11011100,11011100,11011100,11011100
10010010,10010010,10010010,10010010
As you saw, swap alpha is not necessary, we dont have to do swap.

minggo:
@Yodesoft 
I will test your commit.
But i still don't understand your meaning.
I can not find clues by the data you pasted.

Yodesoft:
We dump the bitmap data(drawn by canvas.drawText) to binary format, and found that each value of RGBA of one point is same value all the time. 

example:
one point from bitmap data is 2459079314(unsigned int)
to RGBA is 146,146,146,146
to binary is 11011100,11011100,11011100,11011100
If value of R is always equal value of A, so swap alpha is not necessary.

Hope those infomations could help you understand what i meaning.

minggo:
Can you confirm all R equals to A?

Yodesoft:
Yes.
I write a testcase in nativeInit Bitmap DC then found all R is equals A.

This commit was tested on my Desire HD with Android 4.2.1. Of course, more tests needed.

PR #2378, opened 2013-04-16
Add the keypad return button support.

akira-cn:
Add return button support:  

```
      if(this.setKeypadEnabled){    
            this.setKeypadEnabled(true);
            this.registerScriptKeypadHandler(1);
        }
```

......

```
backClicked:function() {
    cc.log("back button pressed");
}
```

### DIFF
--- a/scripting/javascript/bindings/ScriptingCore.cpp
+++ b/scripting/javascript/bindings/ScriptingCore.cpp
@@ -868,6 +868,23 @@ int ScriptingCore::executeAccelerometerEvent(CCLayer *pLayer, CCAcceleration *pA
 
 int ScriptingCore::executeLayerKeypadEvent(CCLayer* pLayer, int eventType)
 {
+    js_proxy_t * p;
+    JS_GET_PROXY(p, pLayer);
+    
+    if(p){
+        switch(eventType){
+        case 1:
+            executeFunctionWithOwner(OBJECT_TO_JSVAL(p->obj), "backClicked", 0, NULL, NULL);
+            break;
+        case 2:
+            executeFunctionWithOwner(OBJECT_TO_JSVAL(p->obj), "menuClicked", 0, NULL, NULL);
+            break;
+        default:
+            break;
+        }
+        return 1;
+    }
+
     return 0;
 }
 


scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -868,6 +868,23 @@ int ScriptingCore::executeAccelerometerEvent(CCLayer *pLayer, CCAcceleration *pA
 
 int ScriptingCore::executeLayerKeypadEvent(CCLayer* pLayer, int eventType)
 {
+    js_proxy_t * p;
+    JS_GET_PROXY(p, pLayer);
+    
+    if(p){
+        switch(eventType){
+        case 1:
```
dumganhar, line 876: It's better to use `kTypeBackClicked` and `kTypeMenuClicked` to distinguish which type of event.

dumganhar:
@akira-cn 
Is `this.registerScriptKeypadHandler(1);` necessary?
I think we should refactor CCLayer::keyBackClicked and CCLayer::keyMenuClicked like this:

<pre>
void CCLayer::keyBackClicked(void)
{
    if (m_pScriptKeypadHandlerEntry || m_eScriptType == kScriptTypeJavascript) // Modify this line, so registerScriptKeypadHandler isn't needed, because it's only for luabinding.
    {
        CCScriptEngineManager::sharedManager()->getScriptEngine()->executeLayerKeypadEvent(this, kTypeBackClicked);
    }
}
</pre>

scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -868,6 +868,23 @@ int ScriptingCore::executeAccelerometerEvent(CCLayer *pLayer, CCAcceleration *pA
...
+            executeFunctionWithOwner(OBJECT_TO_JSVAL(p->obj), "backClicked", 0, NULL, NULL);
```
dumganhar, line 877: If there isn't any arguments you want to pass to js.
You could simply use `executeFunctionWithOwner(OBJECT_TO_JSVAL(p->obj), "backClicked");`.

PR #2381, opened 2013-04-16
fix compilation error in cocos2dx for BlackBerry PlayBook

lgb:
Someone renamed variable and don't change all applying of its?


dumganhar:
Thanks.

PR #2391, opened 2013-04-17
Log asserts

sbc100:
Log assert messages to the console.

Also, rename/cleanup CCScriptSupport::executeAssert.

### DIFF
--- a/cocos2dx/include/ccMacros.h
+++ b/cocos2dx/include/ccMacros.h
@@ -36,13 +36,12 @@ THE SOFTWARE.
 
 #ifndef CCAssert
 #if COCOS2D_DEBUG > 0
-extern void CC_DLL cc_assert_script_compatible(bool cond, const char *msg);
-#define CCAssert(cond, msg) do {                      \
-      cc_assert_script_compatible(!!(cond), (msg));   \
-      if (!(cond)) {                                  \
-        if (strlen(msg))                              \
-          cocos2d::CCLog("Assert failed: %s", msg);   \
-        CC_ASSERT(cond);                              \
+extern bool CC_DLL cc_assert_script_compatible(const char *msg);
+#define CCAssert(cond, msg) do {                              \
+      if (!(cond)) {                                          \
+        if (!cc_assert_script_compatible(msg) && strlen(msg)) \
+          cocos2d::CCLog("Assert failed: %s", msg);           \
+        CC_ASSERT(cond);                                      \
       } \
     } while (0)
 #else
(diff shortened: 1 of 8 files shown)


cocos2dx/script_support/CCScriptSupport.h
```diff
@@ -225,8 +225,11 @@ class CC_DLL CCScriptEngineProtocol
 
     /** function for common event */
     virtual int executeEvent(int nHandler, const char* pEventName, CCObject* pEventSource = NULL, const char* pEventSourceClassName = NULL) = 0;
-    /** function for assert test */
-    virtual bool executeAssert(bool cond, const char *msg = NULL) = 0;
+
+    /** called by CCAssert to allow scripting engine to handle failed assertions
+     * @return true if the assert was handled by the script engine, false otherwise.
+     */
+    virtual bool handleAssert(const char *msg) = 0;
```
dumganhar, line 232: Yes, I like `handleAssert` more than `executeAssert`, but do you think other `executeXXX` also need to be renamed to `handleXXX`?
For instance, `executeEvent` ~> `handleEvent` ?

sbc100, line 232: Maybe.  I'm don't feel strongly about the others.

I changed this one as it isn't really executing a callback, its deciding if the script engine should handle/display
the assert.   At least that is my understanding of the purpose of this code.   It doesn't call use code/handler at all.

sbc100:
Rebased onto master.  Please take another look.

cocos2dx/include/ccMacros.h
```diff
@@ -36,12 +36,14 @@ of this software and associated documentation files (the "Software"), to deal
...
-#define CCAssert(cond, msg)                             \
-    {                                                   \
-        cc_assert_script_compatible(!!(cond), (msg));   \
-        CC_ASSERT(cond);                                \
-    }
+extern bool CC_DLL cc_assert_script_compatible(const char *msg);
+#define CCAssert(cond, msg) do {                              \
+      if (!(cond)) {                                          \
+        if (!cc_assert_script_compatible(msg) && strlen(msg)) \
+          cocos2d::CCLog("Assert failed: %s", msg);           \
```
dumganhar, line 43: Is it better to output filename and line number for cocos2d::CCLog?

sbc100, line 43: The actual call to assert below should include the line number.  It was only the condition which was missing.. At least on linux and nacl.

dumganhar, line 43: That makes sense.

dumganhar:
Thank you.

PR #2394, opened 2013-04-18
Method signature fix

jotel:
Fixed setAccelerometerInterval signature in setAccelerometerIntervalJNI function.. wrong type of argument used.


dumganhar:
Thanks.

PR #2402, opened 2013-04-18
Adding new macro CCB_MEMBERVARIABLEASSIGNER_GLUE_WEAK to CCBMemberVariableAssigner.h

brawsome:
We've found it useful to use CCB_MEMBERVARIABLEASSIGNER_GLUE_WEAK in
situations where we don't want CCObject derived pointers retained. It's
also useful if you want to assign a pointer dynamically cast to an
abstract base class such as CCLabelProtocol.


dumganhar:
Thank you.

PR #2406, opened 2013-04-19
Adding CCDataVisitor and CCPrettyPrinter

dumganhar:
The usage of CCPrettyPrinter is:

<pre>
    CCPrettyPrinter vistor;
    CCDictionary* pDict = CCDictionary::createWithContentsOfFile("animations/animations.plist");
    pDict->acceptVisitor(vistor);
    CCLog("%s", vistor.getResult().c_str());
</pre>

### DIFF
--- a/cocos2dx/textures/CCTexturePVR.cpp
+++ b/cocos2dx/textures/CCTexturePVR.cpp
@@ -443,7 +443,7 @@ bool CCTexturePVR::unpackPVRv3Data(unsigned char* dataPointer, unsigned int data
 	// unsupported / bad pixel format
 	if (! infoValid)
     {
-		CCLOG("cocos2d: WARNING: unsupported pvr pixelformat: %lx", pixelFormat );
+		CCLOG("cocos2d: WARNING: unsupported pvr pixelformat: %lx", (unsigned long)pixelFormat );
 		return false;
 	}
     


sergey-shambir:
Thanks! I think it better to rename it to CCPrettyPrinter and create method getResult() which returns resulting string. Default visitor can be good example of usage, but there are no need in intermediate visitor since CCDataVisitor calls visitObject() for all unhandled classes.

I.e. class that serializes cocos2d-x data structs to plist will override visit() method for CCDictionary, CCArray, CCString, CCFloat/CCBool/CCDouble/CCInteger and  override visitObject() method to print error message in log (maybe with correct type name received from typeid(*obj).name.

dumganhar:
> I think it better to rename it to CCPrettyPrinter and create method getResult() which returns resulting string. 

Make sense. So if i understand clearly, do you mean that we don't need to `CCLog` in `CCPrettyPrinter`, just save all output string in a buffer, then offer a `getResult()` method?

> Default visitor can be good example of usage, but there are no need in intermediate visitor since CCDataVisitor calls visitObject() for all unhandled classes.

Do you mean that `CCPrettyPrinter` doesn't need to override `visitObject` method? But `visitObject` is a pure method. It should be implemented in CCDataVisitor's subclass.

> (maybe with correct type name received from typeid(*obj).name.

Agree, I'll do it like this. Thanks.

sergey-shambir:
> Make sense. So if i understand clearly, do you mean that we don't need to CCLog in CCPrettyPrinter, just save all output string in a buffer, then offer a getResult() method?

Yes.

> Do you mean that CCPrettyPrinter doesn't need to override visitObject method? But visitObject is a pure method. It should be implemented in CCDataVisitor's subclass

No, just mean that word 'default' in CCDefaultVisitor misleads user: s/he thinks that CCDefaultVisitor should be  subclassed instead of subclassing CCDataVisitor directly.

dumganhar:
Got it, Thank you. I will update this Pull Request soon.

sergey-shambir:
Are there any troubles, or your are busy after release day ;) ?

Actually change will show benefits only when extensions developers will start use it. Currently I see 2 candidates: CCDictionary::writeToFile() and 3rd-party extensions for JSON serialization; I will help to update them.

P.S. JSON also can have own class CCNull, which complicates things a bit.

minggo:
I have told to @dumganhar about this PR.
He will handle it quickly.

dumganhar:
@sergey-shambir 
It was merged. :)

walzer:
OMG, you shouldn't add `acceptVisitor(CCDataVisitor &visitor)` into CCObject.
CCObject was designed only for calculate reference count. Maybe I should rename CCObject to CCReference in the future.

dumganhar:
I agree. Maybe we should rename it to `Ref` in cocos2d 3.0.
The member variables of CCObject like 

<pre>
unsigned int        m_uID;
int                 m_nLuaID;
</pre>

need to be moved to `ScriptTarget` class like which in GamePlay.

ricardoquesada:
Putting the Script data/methods in a new object is a good idea.

Having a way to dump objects to string is also useful.

ricardoquesada:
@dumganhar 

Two quick questions:

a)
currently I need to dump the contents of a `CCDictionary` into a string... could you post a little sample that shows how to do it ?

b)
And also, how can we improve the `CCDictionary # valueForKey` to return a string when the value is a CCinteger ? Currently it is using a `dynamic_cast` to check if it is string and it doesn't look that good:
( eg: https://github.com/cocos2d/cocos2d-x/blob/cocos2d-2.1rc0-x-2.1.3/cocos2dx/cocoa/CCDictionary.cpp#L190 )

Thanks!

dumganhar:
@ricardoquesada 

> a)
> currently I need to dump the contents of a CCDictionary into a string... could you post a little sample that shows how to do it ?

What's the format of the string dumped from CCDictionary? If you just want to show the content of CCDictionary when debugging, CCPrettyPrinter could do that job. If that couldn't meet your requirement, you could implement a new subclass of CCDataVisitor, maybe CCJsonVisitor to dump CCDictionary to JSON format.

> b)
> And also, how can we improve the CCDictionary # valueForKey to return a string when the value is a CCinteger ? Currently it is using a dynamic_cast to check if it is string and it doesn't look that good:

Alternatively, base type like CCInteger, CCBool should implement a virtual method like `toString`. I don't find a better way to achieve that.

PR #2407, opened 2013-04-19
Added __attribute__(__format__()) for CCLog and CCString

sergey-shambir:
Uses printf format. Enabled for GCC and clang. Fixed resulting warnings for all platforms.

### DIFF
--- a/cocos2dx/platform/CCPlatformMacros.h
+++ b/cocos2dx/platform/CCPlatformMacros.h
@@ -259,6 +259,12 @@ public: virtual void set##funName(varType var)   \
 #define CC_FORMAT_PRINTF(formatPos, argPos)
 #endif
 
+#if defined(_MSC_VER)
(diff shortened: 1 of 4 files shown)


sergey-shambir:
I will update this request soon in order to add format checks for CCString

cocos2dx/platform/CCPlatformMacros.h
```diff
@@ -244,6 +244,27 @@ CC_DEPRECATED_ATTRIBUTE static __TYPE__* node() \
     #define CC_DEPRECATED_ATTRIBUTE
 #endif 
 
+/*
+ * only certain compiler support __attribute__((format))
+ * formatPos - 1-based position of format string argument
```
minggo, line 249: > 1-based position

What's the meaning?

sergey-shambir, line 249: First argument has index 1, not 0.

PR #2413, opened 2013-04-19
Rename libbox2d.a -> libBox2D on linux and NaCl platforms.

sbc100:
This is the correct name for the library upstream.

Remove legacy code from:
  template/multi-platform-cpp/proj.linux/build.sh

Update comment in tools/tolua++/build.sh

### DIFF
--- a/external/Box2D/Android.mk
+++ b/external/Box2D/Android.mk
@@ -4,7 +4,7 @@ include $(CLEAR_VARS)
 
 LOCAL_MODULE := box2d_static
 
-LOCAL_MODULE_FILENAME := libbox2d
+LOCAL_MODULE_FILENAME := libBox2D
 
 LOCAL_SRC_FILES := \
 Collision/Shapes/b2ChainShape.cpp \
--- a/external/Box2D/proj.linux/Makefile
+++ b/external/Box2D/proj.linux/Makefile
@@ -1,4 +1,4 @@
-TARGET = libbox2d.a
+TARGET = libBox2D.a
 
 SOURCES = ../Collision/Shapes/b2ChainShape.cpp \
 ../Collision/Shapes/b2CircleShape.cpp \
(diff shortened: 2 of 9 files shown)


sbc100:
ping..

minggo:
If should rename libbox2d to libBox2D, how about other libs, such as chipmunk, png and so on?
And this PR can not be merged automatically.

sbc100:
As far as I know all the other library names match the upstream naming convention.  I will rebase and update the PR

sbc100:
Fixed merge conflicts and rebased onto current master.   Also fixed gcc warning that was recently introduced.

sbc100:
Rebased again.   This change should be good to go.

minggo:
Hi @sbc100, i don't think we need to rename libbox2d to libBox2D.
If we change it, we should change all other libs. It is a big work but not much benefit.
What's your opinion?

sbc100:
I was mostly trying to make it so that it would be easy to replace the box2d lib that we build with the upstream box2d build.   The idea being that box2d is basically an unmodified upstream dependency of cocos2dx and building it from source is just one option.  However, its not that hard to swap out anyway... not a big deal.

I don't think it effects any of the other library names BTW.  Box2D just happens to use uppercase letters in its upstream library name, whereas the others libjpeg libpng, etc don't.

By the way, one of the reasons I was trying to do this drop-in replacement is that the box2d version that we ship is now quite old and there is a bug I am hitting that was already fixed back in 2010.  The was going to suggest upgrading our version of box2d but I was hoping we should do it a new official version:

http://box2d.org/forum/viewtopic.php?f=3&t=9263

What do you think?  Should we upgrade box2d?  Maybe I should start a forum discussion about it.

minggo:
Why i am not agree with the replacement is that, it may break developers. Cocos2d-x developers get used to libbox2d. If it is changed to libBox2D, i don't know what will happen:)

I think it is good to discuss it in cocos2d-x forum.

Thank you again.

aberaud:
My 2 cents is that the library should be renamed to match the upstream name.

The real name is "libBox2D" and there is no reason to use a different name. It also makes easier to replace the library with some other version.

"Normal" game developers use the build-in version so nothing breaks for them.

PR #2418, opened 2013-04-21
network: refactoring, fixed unlikely memory leak, added PUT/DELETE methods

sergey-shambir:
Previously curl_slist_free_all(cHeaders) called in loop after
many break statements, so there was no garantue that it will be called
in all cases.

Second commit implements PUT and DELETE methods, adds test cases to TestCpp.


dumganhar:
Cool :+1:

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/2523. Thanks.

PR #2422, opened 2013-04-22
Adding CCLOG before original path is returned in CCFileUtils::fullPathForFileName to aid in debugging

brawsome:
There's currently no way to be sure, from a calling function, that
fullPathForFilename has returned an invalid file path, which makes it
difficult to quickly track down missing, or incorrect file paths. Added
a CCLOG before the original string is returned to make debugging
missing or incorrect file paths easier.


dumganhar:
Thanks.

PR #2428, opened 2013-04-23
add spine parser

yuezang:


minggo:
It is done in this PR:https://github.com/cocos2d/cocos2d-x/pull/2437.

PR #2433, opened 2013-04-23
Fixes bug which caused invalid GL errors

NatWeiss:
There are certain places that CHECK_GL_ERROR_DEBUG() is called which expect that macro to clear any GL errors. When compiled in release mode (or any mode without COCOS2D_DEBUG==1) this currently doesn't happen because the macro becomes a no-op.

This fixes the macro so it at least clears any GL errors.

Consider CCTexturePVR::createGLTexture(), for example. It sets a bunch of GL parameters, calls CHECK_GL_ERROR_DEBUG() to clear any errors, then proceeds to generate textures and then check for errors manually without the macro. If running in release mode, and setting the parameters does generate an error, then it will be accidentally attributed to the generation of textures, hence causing a false error. The function will think that the texture creation failed when in reality it was from setting the parameters.

I was consistently getting these false errors when loading a PVR texture, unloading it, then loading it again on an older Mac. It took nearly 4 hours of programming time to track down the bug. It was in the call glPixelStorei(GL_UNPACK_ALIGNMENT,1). GL was returning GL_INVALID_ENUM, which is ridiculous, as the parameters are always constant. Nevertheless, GL was giving an error anyway. I suspect this is the reason the CHECK_GL_ERROR_DEBUG() line was added right there, to clear any such false errors.

### DIFF
--- a/cocos2dx/include/ccMacros.h
+++ b/cocos2dx/include/ccMacros.h
@@ -228,7 +228,7 @@ It should work same as apples CFSwapInt32LittleToHost(..)
 #endif
 
 #if !defined(COCOS2D_DEBUG) || COCOS2D_DEBUG == 0
-#define CHECK_GL_ERROR_DEBUG()
+#define CHECK_GL_ERROR_DEBUG() do {glGetError();} while (false)
 #else
 #define CHECK_GL_ERROR_DEBUG() \
     do { \


sergey-shambir:
Are there other places where CHECK_GL_ERROR_DEBUG used to clean erroneous state? I think it's better to use loop with CCLOG (not CCLog) in CCTexturePVR::createGLTexture() rather than change macro.

dumganhar:
@sergey-shambir 
I agree with you. We should refactor this macro like this:

<pre>
#define CHECK_GL_ERROR_DEBUG() \
    do { \
        GLenum __error = glGetError(); \
        if(__error) { \
            CCLOG("OpenGL error 0x%04X in %s %s %d\n", __error, __FILE__, __FUNCTION__, __LINE__); \
        } \
    } while (false)
</pre>

dumganhar:
@NatWeiss 
Normally, developer has ability to check the gl errors in DEBUG mode. If any errors are found, they should be fixed   rather than ignore them. Right?

Therefore, _CHECK_GL_ERROR_DEBUG_ will be valid only in debug mode. IMO, release mode shouldn't do anything.

@NatWeiss @sergey-shambir 
What're your opinions?

Weeds:
James is right on this, there should be no generic calls to glGetError in release mode.

You may want to check the presentation from Valve at GDC. I'don't know if the talk is available online, but the slides can be found here: https://developer.nvidia.com/sites/default/files/akamai/gamedev/docs/Porting%20Source%20to%20Linux.pdf
The talk is about issues that where encountered when porting the Source Engine to Linux (and therefore OpenGL)

Take a look at the Driver Serialization Slide.
According to Valve most drivers are multithreaded (not sure if this is true for mobile GPU drivers), so any OpenGL call you make is moved to a seperate thread by the driver to get submitted to the GPU.
Now there are quite a few methods that will require those driver threads to halt and serialize their state, which you may have guessed is very expensive (and probably throws away many of the optimizations in the driver). One of these methods happens to be glGetError().

I agree with @sergey-shambir, we should change the macro call in the texture loader instead of changing hte macro for release mode.
However it would be even more interesting why the glPixelStorei returned an invalid enum. Possible driver bug?

NatWeiss:
Sorry, I can't seem to dig up which line returned the invalid enum. I do have a feeling it was a driver bug though, as when I was looking at the line, it seemed that the call should never return an error.

dumganhar:
I closed this PR.Thanks.

PR #2443, opened 2013-04-24
Fixing broke linux build

mdkcore0:
The spine code building on Linux stops when checking that some variables can be used when not initialized. This commit fixes it.

I've successively compiled and ran the SpineTest on TestCpp. Please, check the code for possible mistakes that I could have made.

### DIFF
--- a/extensions/spine/Atlas.cpp
+++ b/extensions/spine/Atlas.cpp
@@ -55,9 +55,14 @@ void AtlasRegion_dispose (AtlasRegion* self) {
 
 /**/
 
-typedef struct {
+typedef struct _struct_Str {
 	const char* begin;
 	const char* end;
+
+    _struct_Str()
+        : begin(NULL)
+        , end(NULL) {
+    }
 } Str;
 
 static void trim (Str* str) {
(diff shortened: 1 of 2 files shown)


extensions/proj.linux/Makefile
```diff
@@ -75,7 +75,7 @@ SOURCES = ../CCBReader/CCBFileLoader.cpp \
 
 include $(COCOS_ROOT)/cocos2dx/proj.linux/cocos2dx.mk
 
-CXXFLAGS += -Wno-multichar
+CXXFLAGS += -Wno-multichar --std=gnu++11
```
minggo: Use C++11?

mdkcore0: I've added this because the Str struct 'begin' and 'end' members need to be initialized (due to this flag: '-Werror=maybe-uninitialize'), and to do it, this flag is needed (or just '--std=c++11').

sergey-shambir:
C++11 is not necessary, moreover, GCC supports non-static member initializers since 4.7 and earlier GCC versions will not work; MSVC still doesn't support member initializers.

Please create constructor with inline implementation, like here: http://paste.ubuntu.com/5600483/

mdkcore0:
@sergey-shambir, I've updated the PR with your suggestion, thanks!

minggo:
Thank you, I will test it ASAP.

PR #2453, opened 2013-04-25
- Added pause/resume functionality to CCTouchDelegate, checking this in CCTouchDispatcher::touches

janvdk:
- Added pause/resume functionality to CCTouchDelegate, 
- Added check for pause/resume state to CCTouchDispatcher

This is useful in a custom scene hierarchy where a scene is added and touches in a previous scene should not be processed, but the previous scene must still be present and updating in the background.

### DIFF
--- a/cocos2dx/touch_dispatcher/CCTouchDelegateProtocol.h
+++ b/cocos2dx/touch_dispatcher/CCTouchDelegateProtocol.h
@@ -45,7 +45,7 @@ class CC_DLL CCTouchDelegate
 {
 public:
 
-    CCTouchDelegate() {}
+    CCTouchDelegate() : m_bPaused(false) {}
 
     virtual ~CCTouchDelegate()
     {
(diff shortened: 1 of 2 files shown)


minggo:
Seems great, thank you.

cocos2dx/touch_dispatcher/CCTouchDelegateProtocol.h
```diff
@@ -64,6 +64,13 @@ class CC_DLL CCTouchDelegate
      virtual void ccTouchesEnded(CCSet *pTouches, CCEvent *pEvent) {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
      virtual void ccTouchesCancelled(CCSet *pTouches, CCEvent *pEvent) {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
 
+public:
+    virtual bool isTouchPaused() const { return m_bPaused; }
+    virtual void ccTouchPause() { m_bPaused = true; }
+    virtual void ccTouchResume() { m_bPaused = false; }
```
minggo, line 70: Add paused and resume function here is not so good, because all other functions are touch event call back functions.

janvdk:
Ok thanks,

I took another look at this, how about if I add this to CCTargetedTouchHandler, since it is already storing state for swallowing touches? If this sounds good I will make a new pull request.

Cheers,
Jan

minggo:
If a scene wants to disable a delegate, it can remove the delegate from touch dispatcher, and add it again when needed.

What's you opinion?

janvdk:
Ok I will try to implement this way. Thanks!

PR #2459, opened 2013-04-25
Add Travis build support for linux and NaCl

sbc100:
This branch fixes existing build errors on linux and NaCl and adds a travis config
file for continuous integration.

### DIFF
--- a/README.mdown
+++ b/README.mdown
@@ -1,6 +1,8 @@
 cocos2d-x
 =========
 
+[![Build Status](https://travis-ci.org/sbc100/cocos2d-x.png?branch=travis)](https://travis-ci.org/sbc100/cocos2d-x)
+
 [cocos2d-x][1] is a multi-platform 2D game framework in C++, branched on
 [cocos2d-iphone][2] and licensed under MIT.  The master branch on github uses
 OpenGL ES 2.0 rendering, while the old gles11 branch uses OpenGL ES 1.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: cpp
+script:
+ - export NACL_SDK_ROOT=$PWD/nacl_sdk/pepper_canary
+ - export PATH=$PATH:$NACL_SDK_ROOT/toolchain/linux_x86_newlib/bin
+ - make -j4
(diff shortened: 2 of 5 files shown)


sbc100:
Did you see the "Good to merge" additions to the PR above?  Seems like a travis is a great addition.  Would be good to add unit tests as well as just build tests in the future.

sbc100:
Rebased onto master since several of the warnings was already fixed in the mean time..  Please take another look.

README.mdown
```diff
@@ -1,6 +1,8 @@
 cocos2d-x
 =========
 
+[![Build Status](https://travis-ci.org/sbc100/cocos2d-x.png?branch=travis)](https://travis-ci.org/sbc100/cocos2d-x)
```
dumganhar, line 4: Do you think we should create a new branch for `travis` in `cocos2d/cocos2d-x` repo or just using `master` branch directly?

sbc100, line 4: I would put master here yes.

The .travis.yml is setup to build all branches, but only one can be reported in the README (unfortunately, I imagine that will change at some point).

So you will want to merge this change, then create a travis account that is linked to 
cocos2d/cocos2d-x.  Once it is configured you can create a CL to change this image URL to point
to cocos2d/cococs2d-x rather than sbc100/cocos2d-x.

dumganhar, line 4: > then create a travis account that is linked to cocos2d/cocos2d-x

Do you mean I should create a account for travis and fork `cocos2d/cocos2d-x`?

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2517. Thanks.

README.mdown
sbc100, line 4: They way it is supposed to work (I think), is that you log into https://travis-ci.org/ with your github account, then enable building of the existing cocos2dx repo.  From then on all pushes to that repo will be tested automatically.

dumganhar, line 4: Will all Pull Requests be tested or just be tested after it's `merged`?

dumganhar, line 4: Ok, I think i should read Travis doc first. :)

PR #2458, opened 2013-04-25
Add encrypted PVR.CCZ support to ZipUtils + Tests

edbartley:
This pull request adds encrypted PVR.CCZ support based upon http://www.codeandweb.com/texturepacker/contentprotection

Add sample to TestCPP to demonstrate usage.

PR #2463, opened 2013-04-26
Add decryption for pvr.ccz file encrypted by texturePacker

tiantian20007:
Referring to http://www.codeandweb.com/texturepacker/contentprotection
Just call ZipUtils::caw_setkey_part to set your encryption code;


tiantian20007:
It seems we've done the same thing ^^ 
https://github.com/cocos2d/cocos2d-x/pull/2458

PR #2466, opened 2013-04-26
don't use named semaphore for syncing image loading thread

Weeds:
Named semaphores are unique across process bounds, therefore a semaphore
left in an invalid state may cause locks on other processes.
The code has been exchanged by using pthread mutex and condition variables

See http://www.cocos2d-x.org/boards/6/topics/27188

Code has been tested on Android, iOS and Linux. Have not tested on Windows.
For testing I would recommend adding a log output at the following lines:
Line 123: This method call should block until the next signal will wake the thread, on windows we need to check that the call really blocks and the code does not enter a busy loop.

Optional:
Line 203: waking up the loading thread, so it can be closed
Line 290: waking up the thread for loading a new image

### DIFF
--- a/cocos2dx/textures/CCTextureCache.cpp
+++ b/cocos2dx/textures/CCTextureCache.cpp
@@ -170,6 +170,11 @@ static void* loadImage(void* data)
         s_pAsyncStructQueue = NULL;
         delete s_pImageQueue;
         s_pImageQueue = NULL;
+
+        pthread_mutex_destroy(&s_asyncStructQueueMutex);
+        pthread_mutex_destroy(&s_ImageInfoMutex);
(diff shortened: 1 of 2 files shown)


tiantian20007:
Will this effect CCHttpClient for it's also using named semaphores for iOS?

Weeds:
Thanks for the hint, no it is not included in the above patch. I'll amend some changes.

minggo:
Awesome work, Weeds.
Thank you.

Weeds:
Added Code for HttpClient.

Note that the HttpClient crashes on 64Bit machines (or at least on linux). I will open a seperate pull request for that issue, as it has nothing to do with the semaphore/pthread stuff

PR #2469, opened 2013-04-27
Introduce HttpResponse callback selector type. to resolve #2365

### LINKED ISSUE #2365: CCHttpRequest callback selector type
In CCHttpRequest member, the selector's type is **SEL_CallFuncND** which means CCNode \* object as the first argument. 

> SEL_CallFuncND     _pSelector;      /// callback function, e.g. MyLayer::onHttpResponse(CCObject *sender, void *data)

And this selector is used at CCHttpClient:: dispatchResponseCallbacks as this way.

> ```
>    if (pTarget && pSelector) 
>    {
>        (pTarget->*pSelector)((CCNode *)this, response);
>    }
> ```

Here, the 'this' is a **CCHttpClient** instance which derived from **CCObject**, but by the type of selector, there exists oddly casting. Why this invalid casting is needed? Is the reason there is no selector type which takes **(CCObject *, void *)** as its parameters? If so, how about adding new type for supporting those cases?

coolengineer:
Hello, I made a cheap patch for http reponse callback as discussed in #2365.

Review please ;-)

### DIFF
--- a/extensions/network/HttpClient.cpp
+++ b/extensions/network/HttpClient.cpp
@@ -510,11 +510,11 @@ void CCHttpClient::dispatchResponseCallbacks(float delta)
         
         CCHttpRequest *request = response->getHttpRequest();
         CCObject *pTarget = request->getTarget();
-        SEL_CallFuncND pSelector = request->getSelector();
+        SEL_HttpResponse pSelector = request->getSelector();
 
         if (pTarget && pSelector) 
         {
-            (pTarget->*pSelector)((CCNode *)this, response);
+            (pTarget->*pSelector)(this, response);
         }
         
         response->release();
--- a/extensions/network/HttpRequest.h
+++ b/extensions/network/HttpRequest.h
@@ -30,11 +30,17 @@
 
 NS_CC_EXT_BEGIN
 
+class CCHttpClient;
+class CCHttpResponse;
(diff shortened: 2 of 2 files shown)


sergey-shambir:
Did you fixed warnings about deprecated function in Samples directory?

P.S. any warning will break build on linux.

dumganhar:
We can’t automatically merge this pull request.

samples/Javascript/Shared
```diff
@@ -1 +1 @@
-Subproject commit 721c3adc5bc3a7d01de39f3d0f50a9cea2c82916
+Subproject commit 53c37689dba36041befbe3bee953383f08505486
```
dumganhar, line 1: Why did you modify the submodule reference?

coolengineer, line 1: Oh.. it's mistake.. can you drop that commit? Or should I cancel and make new request?

sergey-shambir, line 1: Actually, you can update change with some git manipulations: create another branch with "git checkout -b pull_request_fix origin/network-httpclient", cherry-pick 2nd commit by "git cherry-pick c766aed", and that use "git push -f origin".

coolengineer:
Request changed to #2490

PR #2476, opened 2013-04-27
Adds weak reference support

TimothyZhang:
Most memory leaks are caused by circular references.  CCWeakReference is created to solve this problem.
## Usage:

```
CCObject *obj = new CCObject()
CCWeakReference ref1 = CCWeakRefenrece(obj);  
CCWeakReference ref2;
ref2 = ref1;
ref2.getObject(); // obj
CC_SAFE_RELEASE_NULL(obj);
ref2.getObject(); // NULL
```
## Efficiency:

The only overhead is `releaseWeakReferenceFor(this);` called in CCObject::~CCObject(). But the time it takes is really negligible. It runs at O(lgN) time, in which N is the amount of CCObjects been weakly referenced.

### DIFF
--- a/cocos2dx/cocoa/CCObject.cpp
+++ b/cocos2dx/cocoa/CCObject.cpp
@@ -30,6 +30,7 @@ THE SOFTWARE.
 
 NS_CC_BEGIN
 
+
 CCObject* CCCopying::copyWithZone(CCZone *pZone)
 {
     CC_UNUSED_PARAM(pZone);
@@ -47,8 +48,12 @@ CCObject::CCObject(void)
     m_uID = ++uObjectCount;
 }
 
+static void releaseWeakReferenceFor(CCObject *obj);
 CCObject::~CCObject(void)
 {
+    // release weak reference to this object
+    releaseWeakReferenceFor(this);
+
     // if the object is managed, we should remove it
     // from pool manager
     if (m_uAutoReleaseCount > 0)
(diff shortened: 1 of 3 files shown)


cocos2dx/cocoa/CCObject.h
```diff
@@ -27,6 +27,8 @@ of this software and associated documentation files (the "Software"), to deal
 
 #include "platform/CCPlatformMacros.h"
 
+#include <map>
```
sergey-shambir, line 30: Move this include to CCObject.cpp

cocos2dx/cocoa/CCObject.h
```diff
@@ -93,6 +95,42 @@ class CC_DLL CCObject : public CCCopying
 #define event_selector(_SELECTOR) (SEL_EventHandler)(&_SELECTOR)
 #define compare_selector(_SELECTOR) (SEL_Compare)(&_SELECTOR)
 
+
+// weak reference to CCObject
+class CC_DLL CCWeakReference {
+public:    // interfaces
```
sergey-shambir, line 101: No need in annotations like // interfaces or // private methods.

sergey-shambir:
I think it's better to split weak reference on two classes: template class CCWeakPointer and private class CCWeakPointerData, with all methods private and friends 'CCObject' and 'CCWeakPointer'. Only CCWeakPointer will be accessible for users.

In such case you will have weak pointer with strong typing, so there will be no need to use getObjectEx() or dynamic_cast: CCWeakPointer will use static_cast instead and accept type _T to it's constructor. Like this: https://github.com/ivzave/cocos2dx-ext/blob/master/CCPointer.h

You can add underscore to private class name or put it to nested namespace - to exclude it from code completion in any IDE.

TimothyZhang:
Thanks for your suggestion.

What's the purpose of CCWeakPointerData ? Did you mean putting the `_referent_map` and `releaseWeakReferenceFor()` together into it?

cocos2dx/cocoa/CCObject.cpp
```diff
@@ -115,4 +120,133 @@ bool CCObject::isEqual(const CCObject *pObject)
...
+// the key is CCObject::m_uID
+typedef std::map<unsigned int, Referent> ReferentMap;
+
+cocos2d::ReferentMap *_referent_map = NULL;
+
+// the order of initialization of global variables is undeterminable, using lazyload
+// can avoid potential crash.
+static inline ReferentMap* GetReferentMap()
+{
+  if( !_referent_map )
```
sergey-shambir, line 137: You can hide _referent_map inside this function using static variable, which will be initialized once (it's standard C/C++ feature):

static _referent_map = new ReferentMap();

Of course such code isn't thread-safe too.

sergey-shambir:
Currently CCWeakReference works with CCObject, but it can be template class and work with concrete type. In the other hand, template class implementation should be in header. So you can define class CCWeakReferenceData that will do actual work using CCObject, and put it's implementation to CCObject.cpp.

Than, make CCWeakReferenceData member of CCWeakReference and call it's methods, converting object with static_cast when needed and accepting (_T *) into CCWeakReference constuctor.

In this case you will take benefits of strong typing without need to put all CCWeakReference implementation to header.

I also suggest to rename CCWeakReference to CCWeakPointer.

sergey-shambir:
I will make pull request to your repository to show my suggestions in code.

sergey-shambir:
Cannot create pull request to your repo, created #2481 instead

TimothyZhang:
Would the name `CCWeakPointer` make people thinking it could been used as a pointer?
Like `CCPointer p`, we can write `p->xxx();` , but we can't use `CCWeakPointer` like that。

sergey-shambir:
Nor std::weak_ptr neither QWeakPointer have operator ->. In both cases, user should convert weak pointer to strong and than work with it, otherwise pointer can become null at any moment.

In other hand, cocos2d-x have autorelease pull and supports creating/deleting objects only from main thread, so user can just get data from weak pointer, check if it's null and than work. However, I think it's better to not extend weak pointer with -> and \* operators to avoid confusion.

sergey-shambir:
I've played with similar class a bit more and now think that CCWeakPointer should have -> and \* operators. So I'll update my pull request soon and implement your suggestion.

Minggo: how about to create smart pointers system with 3 kinds: CCStrongPointer (similar to [1]), CCWeakPointer and CCRawPointer? CCRawPointer just inits pointer with NULL without any overhead, so user doesn't need to initialize pointer with NULL in constructor.

[1] https://github.com/ivzave/cocos2dx-ext/blob/master/CCPointer.h

minggo:
Please refer to #2481, I think this PR should be close too.

Issue #1738, opened 2012-12-07
Bug: Memory leak with CCBAnimationManager


fgig77:
The CocosBuilderTest example will leak layers when going into and out of the AnimationsTestLayer. This is because AnimationsTestLayer keeps a reference to CCBAnimationManager and CCBAnimationManager keeps a reference to AnimationsTestLayer. It's a circular dependency. Neither class will get it's destructor called because the reference count never drops.

For more info:
http://www.cocos2d-x.org/boards/6/topics/18169

TimothyZhang:
we may use weak reference, if pull request(https://github.com/cocos2d/cocos2d-x/pull/2476) would been applied.

godyZ:
@fgig77 Please, could you open a bug in redmine in future? We are in process of migrating all the issues to redmine and github issue won't be available after August.

PR #2481, opened 2013-04-27
Adds weak reference support

sergey-shambir:
Most memory leaks are caused by circular references. CCWeakPointer is created to solve this problem.
## Usage:

```
CCDictionary *obj = new CCDictionary()
CCWeakPointer<CCDictionary> ref1(obj);  
CCWeakPointer<CCDictionary> ref2;
ref2 = ref1;
ref2.getObject(); // obj
CC_SAFE_RELEASE_NULL(obj);
ref2.getObject(); // NULL
```
## Efficiency:

The only overhead is `releaseWeakReferenceFor(this);` called in `CCObject::~CCObject()`. But the time it takes is really negligible. It runs at O(lgN) time, in which N is the amount of CCObjects been weakly referenced.

### DIFF
--- a/cocos2dx/cocoa/CCObject.h
+++ b/cocos2dx/cocoa/CCObject.h
@@ -139,8 +139,23 @@ public:
         return m_data == weakref.m_data;
     }
 
-    inline _T *getObject() {
+    inline _T *getObject() const {
         return static_cast<_T *>(m_data.getObject());
+    }
+
+    T *operator ->() const
+    {
+        return getObject();
+    }
+
+    T &operator *()
+    {
+        return *getObject();
+    }
+
+    const T &operator *() const
+    {
+        return *getObject();
     }
 
 private:


minggo:
Is it the same as https://github.com/cocos2d/cocos2d-x/pull/2476?

sergey-shambir:
Behaviour created by Timothy didn't changed. I've added template class CCWeakPointer, everithing else are just style changes.

minggo:
I am not sure if this feature is needed.
It will make things complex at first glance.

sergey-shambir:
Yes, it really increases complexity (despite that weak and strong references is classic C++ concept, people come from other languages may not know it, and lack of strong pointer mishleads user too).

It also can hurt perfomance when many CCObjects used, in data structs or another place. Two optimizatons can be done in this area: 1) use std::unordered_map or std::tr::unordered_map on some platforms; 2) keep amount of week references in CCObject and do not release references if m_nWeakRefs == 0.

About pros of weak refs: it can be helpful when working with CCNode, since it lifetime managed by scene and it should be destroyed when scene destroys, or after removeFromParentAndCleanup(). Weak ref also benefits from single-threaded nature of cocos2d-x: user shouldn't cast pointer to shared_ptr before use it.

So I can suggest following:
1) move weak ref stuff from CCObject to CCNode
2) use m_nWearRefs and std::unordered_map (if available) to decrease perfomance drop.
3) add classic strong ref pointer

I'm also curious: Timothy, which use-cases do you have for CCWearReference in addition to safe CCNode tracking?

TimothyZhang:
1) In most case, weak reference is used in callbacks.

  Under delegate/observer pattern, the dispatcher retains the target, which should be released later in a proper place. But programmer might forget to do that.

   For example, in our games, we have a EventManager. When a value changed, say "gold", we post a new event something like "gold_changed".  Many objects are interested in this event. The CCLabel(s) displaying gold amount should be updated, the AchievementManager need to check whether a new achievemnt is acquired, the QuestManager would check whether a quest is finished,...  There're many such kind of callbacks, dereferencing them manualy is a big headache.

2) It can be used to avoid circular reference. eg #1738.
    Some of my CCNodes needs a userobject, which  holds a weak reference to the CCNode itself.
    `CCNode *node = CCNode::create();
    node->setUserObject( MyUserObject::create(node) );`    

3) Sometimes we have to reference to something may or may not exists, a weak refrence can be more convenient than a strong reference.  
4) It can be used to cache objects, although it's rarely used.

In general, weak refenrece is very convenient, like CCPointer, CCStrongPointer...

sergey-shambir:
So, CCWeakPointer may be implemented only for CCNode without negative effects?

walzer:
@sergey-shambir 

Well, CCWeakPointer is useful, but my suggestion is not to merge it into cocos2d v2.x versions. I have been tried to avoid using c++ templates in v1.x & 2.x versions, so cocos2d-x would be easier to learn for objective-c & java programmers.

In cocos2d 3.0, let's consider shared & weak ptr usage. Perhaps we should use your CCWeakPointer, or use C++11's shared_ptr & weak_ptr. v3.0 is planned in 3~4 months.

As the result, if developers don't like c++ templates and c++11, please stay on v2.x, if you want powerful features, move to v3.x.

sergey-shambir:
Thanks for clarification, than I'll close this pull request.

walzer:
Bump up.
Refer to Ricardo's #2542 and http://www.cocos2d-x.org/boards/6/topics/29971?r=30058#message-30058,
we should consider implementing weak ptr in cocos2d v3.0

This implementation of CCWeakPointer uses a std::map to restore all pointer-ref pairs. While a game usually has  1000+ objects, release each object need to find it in ReferentMap, that is too expensive.

Any ideas to improve the performance?

sergey-shambir:
- CCObject can track amount of weak references and do nothing when weak reference count is 0.
- std::unordered_map can be used instead of std::map. It should be available on all platforms, but there are bug in libstdc++ 4.7: http://stackoverflow.com/questions/11614106/is-gcc-stdunordered-map-implementation-slow-if-so-why
- Finally, some classes can be removed from CCObject hierarchy, classes from /cocoa are good candidates.

slawo:
Instead of using a map... a pointer to a weak reference utility object holding a pointer to your CCObject can be used. let's call the struct/class of this utility object CCWeakReferenceImpl<T>.
The CCObject holds a pointer to the CCWeakReferenceImpl<T>
The insatnce of CCWeakReferenceImpl<T> holds a pointer to the CCObject and a referenceCount.
Each CCWeakReference<T> instead of holding a pointer to CCObject hold a pointer to the CCWeakReferenceImpl<T> which gives them access to the CCObject.
When the CCObject is destroyed it simply changes the pointer in his instance of CCWeakReferenceImpl<T> to NULL and all CCWeakReference<T> will immediately point to NULL.
When the last WeakReference<T> is cleared/destroyed it sets the ref count in CCWeakReferenceImpl<T> to 0 and the CCWeakReferenceImpl<T> destroys itself.
This is the most efficient way of implementing weak references in a framework as it enables the creation and destruction of an object without using arrays and repeating an operation that can be done only once.

Adding to this I believe that if CCWeakReference is not templated it should be named differently i.e. CCObjectRef or CCObjectWeakReference

PR #2483, opened 2013-04-27
Fixed crash in HttpClient request on 64Bit

Weeds:
According to the cURL documentation the pointer that should be passed
to curl_easy_getinfo with CURLINFO_RESPONSE_CODE must be a long
pointer,

On platforms where long is 64Bit this will cause a stack corruption
destroying the pointer to the original resonse as the current response
code variable was a 32 Bit int value.

See: http://curl.haxx.se/libcurl/c/curl_easy_getinfo.html


minggo:
Thank you.

PR #2490, opened 2013-04-28
Introduce HttpResponse callback selector type to resolve #2365, fixed version ;-)

### LINKED ISSUE #2365: CCHttpRequest callback selector type

coolengineer:
Dropped the changeset of shared directory from previous pull-request.
And fixed warnings of samples which were related with changing sources.


dumganhar:
Thanks for your contribution. It looks good. I'll test it soon.

dumganhar:
Merged, thanks.

coolengineer:
Thanks!

PR #2493, opened 2013-04-30
Added support for javascript byte code

sunzhuoshi:
A tool named "jsbcc" introduced to compile javascript files to byte code files
Byte code file uses ".bc" suffix. i.e. "test.js" will be compiled to "test.js.bc". 
When running, "test.js.bc" will be checked first, if not found, "test.js" will be used

Tested on iOS and Android, but no test cases included, for different repos.


sunzhuoshi:
Fail to handle multiple files when used via pipeline

PR #2494, opened 2013-04-30
Added support for javascript byte code

sunzhuoshi:
A tool named "jsbcc" introduced to compile javascript files to byte code files
Byte code file uses ".bc" suffix. i.e. "test.js" will be compiled to "test.js.bc". 
When running, "test.js.bc" will be checked first, if not found, "test.js" will be used

Tested on iOS and Android, but no test cases included, for different repos.

### DIFF
--- a/scripting/javascript/bindings/ScriptingCore.cpp
+++ b/scripting/javascript/bindings/ScriptingCore.cpp
@@ -446,7 +446,7 @@ void ScriptingCore::createGlobalContext() {
     }
 }
 
-std::string RemoveFileExt(const std::string& filePath) {
+static std::string RemoveFileExt(const std::string& filePath) {
     size_t pos = filePath.rfind('.');
     if (0 < pos) {
         return filePath.substr(0, pos);
@@ -462,6 +462,7 @@ JSBool ScriptingCore::runScript(const char *path, JSObject* global, JSContext* c
         return false;
     }
     cocos2d::CCFileUtils *futil = cocos2d::CCFileUtils::sharedFileUtils();
+    std::string fullPath = futil->fullPathForFilename(path);
     if (global == NULL) {
         global = global_;
     }
@@ -471,7 +472,7 @@ JSBool ScriptingCore::runScript(const char *path, JSObject* global, JSContext* c
     JSScript *script = NULL;    
     js::RootedObject obj(cx, global);
 	JS::CompileOptions options(cx);
-	options.setUTF8(true).setFileAndLine(path, 1);
+	options.setUTF8(true).setFileAndLine(fullPath.c_str(), 1);
     
     // a) check js file first
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
@@ -482,7 +483,7 @@ JSBool ScriptingCore::runScript(const char *path, JSObject* global, JSContext* c
         script = JS::Compile(cx, obj, options, (char*)content, strlen((char*)content));
     }
 #else
-    script = JS::Compile(cx, obj, options, path);
+    script = JS::Compile(cx, obj, options, fullPath.c_str());
 #endif
     // b) no js file, check jsc file
     if (!script) {


dumganhar:
Hi, do you think renaming the suffix `.js.bc`(javascript byte code) to `.jsc`(javascript compiled) is better since `cocos2d-iphone` is using `.jsc` suffix?
Plz refer to https://github.com/zynga/jsbindings/blob/master/src/manual/jsb_core.mm#L459

sunzhuoshi:
Well, I will change it to ".jsc".  
By the way, I've check cocos2d-iphone's code. It also supports auto-generating byte code files. Maybe it is not necessary, if a javascript byte code "compiler" supplied. Please help to synch with cocos2d-iphone's dev team about this issue.

dumganhar:
> By the way, I've check cocos2d-iphone's code. It also supports auto-generating byte code files. Maybe it is not necessary, if a javascript byte code "compiler" supplied.

I think it's necessary. When developers write their games, they'll change js source code frequently, so every time they change source code, they don't need to use `compile tool` immediately. Therefore, `Compile Tool` is used only when games are finished and packaged.

sunzhuoshi:
Yeah, you are right about code change, but is it necessary for developers to maintain their byte code files when writing their code, or what is the benefit to do this? It will be OK to just use the js source files, and they will be checked before byte code files.

dumganhar:
Cocos2d-iphone now will check file date to execute `js` or `jsc` files. If jsc file is older than js file, jsc file will be replaced with latest jsc file generated by latest js file.

You could refer to https://github.com/zynga/jsbindings/blob/master/src/manual/jsb_core.mm#L487.
The macro `JSB_ENABLE_JSC_AUTOGENERATION` does that job.

For short-term, the PR is enough.
For long-term, I think we should implement a method in `CCFileUtils` to check file date like which in Obj-C:

<pre>
NSDate *jscDate = [attrs fileCreationDate];
</pre>

What's your opinion?

sunzhuoshi:
Yeah, I've check the code. 
It is the solution from Zynga, just use auto-generating without a separate "compiler". To be honest, it it the solution I used first. After consideration, I've decided to create a new simple command line tool to replace the "lazy but work" solution. 
I think the main purpose of introduce byte code is to protect code(maybe faster). and it is not necessary to use byte code files when developing or to distribute js source files in the release version. I'll use byte code files only in the release version. and I'll create a custom script(only byte code files copied, js source files excluded) to build the release version to be released or tested.
By the way, I am a game developer, too. :)

And I've changed the loading order to load js source file first, maybe a little slower for the release version, but more convenient for developers.
If you think Zynga's solution is better, I will be fine. And you have to add the rest code for auto-generation.

dumganhar:
I agree with you. Let's do it in this way now.
Does this tool work on Windows?

scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -446,45 +446,63 @@ void ScriptingCore::createGlobalContext() {
...
+        //JSScript* script = JS_CompileScript(cx, global, (char*)content, contentSize, path, 1);
+        script = JS::Compile(cx, obj, options, (char*)content, strlen((char*)content));
+    }
 #else
-    // Removed in SpiderMonkey 19.0
-    //JSScript* script = JS_CompileUTF8File(cx, global, rpath.c_str());
-	JSScript *script = JS::Compile(cx, obj, options, rpath.c_str());
+    script = JS::Compile(cx, obj, options, rpath.c_str());
 #endif
+    // b) no js file, check jsc file
```
dumganhar, line 488: I think we should check whether jsc file exists first, if it doesn't exist, then check js file.

sunzhuoshi, line 488: Yeah, it is my former solution. Good for product, but inconvenient for developers. i.e. if you found something wrong, after a long time check, you would find the jsb file is an older one... I've suffered from this... Perphaps it is one of the reasons why auto-generating is needed in Zynga's solution.
Maybe a performance test is needed to tell how much performance lost if we load js files first.

sunzhuoshi:
Yeah, you're right.
And I hate the former file coder...

scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -446,45 +446,55 @@ void ScriptingCore::createGlobalContext() {
     }
 }
 
+std::string RemoveFileExt(const std::string& filePath) {
```
dumganhar: It would be better if you mark this function as a `static` function.

sunzhuoshi: Yeah, you're right.

PR #2495, opened 2013-04-30
Support negative node scales of CCScrollView

ptekchand:
Not checking for negative scale values in CCScrollView::getViewRect causes intersectsRect calls in CCScrollView (eg: to check if the touch was within the bounds) to return false. This is because the CCRect being returned has a negative width and height.

One may set a negative scale for something like allowing scrolling top to bottom in terms of window coordinates (i.e. Y positive downwards).
http://stackoverflow.com/a/16194082


dumganhar:
Thanks.

PR #2496, opened 2013-04-30
Fix missing protocol method warning for iOS 6.0 addition

mchinen:
Fixes a warning when compiling for iOS 6.0 (a new method was added to UITextInput Protocol in iOS 6.0, which EAGLView implements).

The method does nothing but CCLog, like the other method implementations for hit detection.


michaelcontento:
It would be nice, if you could indent the method body with 4 instead of 3 spaces :)

mchinen:
Ah, sorry, forgot about that.  Fix with proper indent added.

michaelcontento:
Thanks and I hope it gets merged soon :+1:

minggo:
Thank you.

PR #2501, opened 2013-04-30
Fixed a bug in ScriptingCore::evalString not setting the outval.

nicolasgramlich:
Talked to @ricardoquesada and @rohankuruvilla about this and the fix is apparently the proper way to do this.


minggo:
Thank you @nicolasgramlich .
Because the PR can not be automatically merged, so i did the modification in #2503.

PR #2502, opened 2013-05-01
Overload operators +,-,*,/ in CCPoint

aberaud:
CCPoint are very often used as 2D vectors. Overloading basic operators in CCP would lead to a much cleaner, readable code when manipulating points.
This have already been implemented successfully in other Cocos projects such as cocos2d-xna.

This PR allow to add/subtract CCPoint as vectors:

``` c++
CCPoint a(5, 10), b(6, 10);
c = a+b; // c == (11, 20)
c = a-b; // c == (-1, 0)
```

It also defines multiplication by a scalar:

``` c++
c = a/2; // c == (2.5, 5)
c = b*2; // c == (12, 20)
```

Comparison for a typical use-case:
before:

``` c++
CCPoint res = ccpAdd(ccpMult(ccpAdd(point1, point2), 2), ccpMult(ccpAdd(point2, point3), 2));
```

after:

``` c++
CCPoint res = (point1 + point2) * 2 + (point2 + point3) * 2;
```

### DIFF
--- a/cocos2dx/cocoa/CCGeometry.cpp
+++ b/cocos2dx/cocoa/CCGeometry.cpp
@@ -73,6 +73,14 @@ CCPoint CCPoint::operator*(float a) const
 	return p;
 };
 
+CCPoint CCPoint::operator/(float a) const
+{
+	CCPoint p;
+	p.x = this->x / a;
+	p.y = this->y / a;
+	return p;
+}
+
 void CCPoint::setPoint(float x, float y)
 {
     this->x = x;
--- a/cocos2dx/cocoa/CCGeometry.h
+++ b/cocos2dx/cocoa/CCGeometry.h
@@ -49,6 +49,7 @@ public:
     CCPoint operator+(const CCPoint& right) const;
     CCPoint operator-(const CCPoint& right) const;
     CCPoint operator*(float a) const;
+    CCPoint operator/(float a) const;
     void setPoint(float x, float y);
     bool equals(const CCPoint& target) const;
 };


minggo:
How about use [RVO](http://en.wikipedia.org/wiki/Return_value_optimization) to optimize it?

sergey-shambir:
Did you mean method with "inline", defined in header?

aberaud:
It may be possible to inline, but for simple 3-intructions functions like these, compilers will inline anyway.

cocos2dx/cocoa/CCGeometry.cpp
```diff
@@ -49,6 +49,38 @@ of this software and associated documentation files (the "Software"), to deal
     return *this;
 }
 
+CCPoint CCPoint::operator+(const CCPoint& right) const
```
metalbass, line 52: If you want to optimize it with RVO i'd do this:

<pre lang="cpp"><code>CCPoint CCPoint::operator+(const CCPoint& right) const
{
    return CCPoint(this->x + right.x, this->y + right.y);
}</code></pre>

aberaud, line 52: Not sure since the constructor then calls setPoint :

``` c++
CCPoint::CCPoint(float x, float y)
{
    setPoint(x, y);
}
```

setPoint then set the values in the same way the PR does.

metalbass, line 52: You are right, constructor should be like this:

<pre lang="cpp"><code>CCPoint::CCPoint(float x, float y) : x(x), y(y) { }</code></pre>


There's no need to call a method for that, anyway :P

minggo, line 52: @metalbass 
Right.

aberaud:
PR #2504 implements the changes proposed here

PR #2504, opened 2013-05-02
Move ccpX functions to CCPoint methods and operators

aberaud:
Following the discussion in https://github.com/cocos2d/cocos2d-x/pull/2502 , I implemented the proposed changes in this PR.

CCPoint are very often used as 2D vectors. Overloading basic operators in CCP would lead to a much cleaner, readable code when manipulating points.
This have already been implemented successfully in other Cocos projects such as cocos2d-xna.

This PR allow to add/subtract CCPoint as vectors:

``` c++
CCPoint a(5, 10), b(6, 10);
c = a+b; // c == (11, 20)
c = a-b; // c == (-1, 0)
```

It also defines multiplication by a scalar:

``` c++
c = a/2; // c == (2.5, 5)
c = b*2; // c == (12, 20)
```

Comparison for a typical use-case:
before:

``` c++
CCPoint res = ccpAdd(ccpMult(ccpAdd(point1, point2), 2), ccpMult(ccpAdd(point2, point3), 2));
```

after:

``` c++
CCPoint res = (point1 + point2) * 2 + (point2 + point3) * 2;
```

### DIFF
--- a/cocos2dx/cocoa/CCGeometry.cpp
+++ b/cocos2dx/cocoa/CCGeometry.cpp
@@ -78,6 +78,7 @@ CCPoint CCPoint::operator*(float a) const
 
 CCPoint CCPoint::operator/(float a) const
 {
+	CCAssert(a, "CCPoint division by 0.");
     return CCPoint(this->x / a, this->y / a);
 }
 
@@ -162,6 +163,7 @@ CCSize CCSize::operator*(float a) const
 
 CCSize CCSize::operator/(float a) const
 {
+	CCAssert(a, "CCSize division by 0.");
     return CCSize(this->width / a, this->height / a);
 }
 


aberaud:
If it make sense we could also make the same changes for CCSize and allow conversion between CCPoint and CCSize by overloading operator=(const CCSize&) in CCPoint, as proposed in http://www.cocos2d-x.org/boards/6/topics/27272?r=27580#message-27580 .

This would simplify the code further by allowing stuff like:

``` c++
CCSize winSize = CCDirector::sharedDirector()->getWinSize();
CCPoint screenCenter = winSize / 2;
```

Another cool addition would be to have length() and lengthSq() methods in CCPoint. These functions would compute the length or squared length of the vector. It would be possible to write code like:

``` c++
CCPoint vecta(0, 3);
CCPoint vectb(4, 0);
float length = (vecta - vectb).length(); // points dist = 5.0
```

If people agree I will add these changes to this PR.

metalbass:
I would get the CCSize operators and cross-assignments (CCPoint.operator=(const CCSize&) & CCSize.operator=(const CCPoint&))
Can't wait for this to get into next version of the framework :)

aberaud:
Do the + and - operators make sense with CCSize ? Or shall we just implement \* and / ?

aberaud:
BTW cross-assignments between CCPoint and CCSize will not only require operator= overloading but also conversion constructors to work smoothly:

``` c++
CCPoint::CCPoint(const CCSize& size)
{
    setPoint(size.width, size.height);
}

CCPoint& CCPoint::operator= (const CCSize& size)
{
    setPoint(size.width, size.height);
    return *this;
}
```

Otherwise

``` c++
CCPoint a = size;
```

would not work but

``` c++
CCPoint a;
a = size;
```

would, which would be weird.

michaelcontento:
:+1: for this PR and both (`*`, `/` and `+`, `-`) additions to `CCSize`. 

Also the cross-assignments between `CCPoint` and `CCSize` is something I could immediately take advantage of. The current API regarding `CCSize` and `CCPoint` is not that developer friendly :/

@wagaf-d : As you've digged into this topic: I'm quite new to cocos2d-x, but is there a reason why `CCPoint` is rather empty and we have a ballpark of loose functions like `ccpAdd`, `ccpNeg`, `ccpDot` and friends? Wouldn't it be more read-/writable if those were member functions of `CCPoint`? At least they would be easier to discover for newbies like me ;)

metalbass:
@wagaf-d You're right about that :P
@michaelcontento Seems to me that cocos2d-x is programmed like it's in C in some parts...

winipcfg:
Functions like ccpAdd, ccpSub comes from Cocos2d library which is written in Objective-C.

aberaud:
@michaelcontento I agree, it would be very relevant to have these functions as CCPoint methods. I suggest to have this PR merged first, since it's already a good step forward.

metalbass:
@winipcfg Well, in either case they are not very object-oriented, that's why i've said that they look like C to me (not object-oriented). All the methods that use only info from the class could/should be put inside the class.

@wagaf-d I don't know if we want to implement only a part or all of those loose methods that use one or more points. My initial guess is that having only a part of them in the class could be confusing (once you find them in one place you won't be searching anymore, even there are more somewhere else...) so we need to implement them all, even if they just return simple calls to the original ones.

I'm not very familiar with github, so i don't know if there's anyway that i can help by adding some of those methods (can i do a pull request on a pull request? xD)

michaelcontento:
@metalbass I share your point of view and we should do this in one "big" change. But it shouldn't be that complicated as it's mostly a re-arrangement of existing code :) 

But - IMHO - we should move the real code inside `CCPoint` and use them from `ccpAdd` and friends. Because this feels more like "The Right Way"© as it encourages a clean API design with maintained BC -- and not vice versa.

> I'm not very familiar with github, so i don't know if there's anyway that i can help by adding some of those methods (can i do a pull request on a pull request? xD)

Just fork [this repo/branch](https://github.com/wagaf-d/cocos2d-x/tree/ccpchanges) and send @wagaf-d a PR :)

metalbass:
Ok, pull request here https://github.com/wagaf-d/cocos2d-x/pull/1

aberaud:
@michaelcontento  Makes total sense. 

How did @metalbass and all his messages disappeared from the discussion ? 

Anyway it's fairly easy so I'll do it.

Do you think functions ccpLength, ccpAngle, ccpAngleSigned etc. should be named length(), angle() etc. or getLength(), getAngle() ?
I have a preference for the shortest, simplest version but getX may be clearer ?

What about ccpSegmentIntersect, ccpIntersectPoint and ccpLineIntersect ? Do you think we should add them as static methods to CCPoint or keep them in a separate file ?

A first list of methods that would be added to CCPoint:

``` c++
    bool fuzzyEquals(const CCPoint& target, float variance) const;
    float length() const;
    float lengthSq() const;
    float angle() const;
    float angle(const CCPoint& other) const;
    float angleSigned() const;
    float angleSigned(const CCPoint& other) const;
    float dot(const CCPoint& other) const;
    float cross(const CCPoint& other) const;
    CCPoint perp() const;
    CCPoint rPerp() const;
    CCPoint project(const CCPoint& other) const;
    CCPoint rotate(const CCPoint& other) const;
    CCPoint unrotate(const CCPoint& other) const;
    CCPoint normalize() const;
    CCPoint lerp(const CCPoint& other, float alpha) const;
    CCPoint rotateByAngle(const CCPoint& pivot, float angle) const;
```

And a static function:

``` c++
    static CCPoint forAngle(const float a);
```

michaelcontento:
Huh? It seems that @metalbass deleted his account entirely :/

> Do you think functions ccpLength, ccpAngle, ccpAngleSigned etc. should be named length(), angle() etc. or getLength(), getAngle() ?

Hmm .. I would try to keep the cocos2d-x convention and it seems, after a quick look in some classes like `CCNode` and `CCSprite`, that the long way (`getLength`, `getAngle` etc.) is the way to go.

> A first list of methods that would be added to CCPoint: [.. snip ..]

Don't forget `ccpDistance` and `ccpDistanceSQ` :)

> What about ccpSegmentIntersect, ccpIntersectPoint and ccpLineIntersect ? Do you think we should add them as static methods to CCPoint or keep them in a separate file ?

IMHO they should be moved, as static methods, too. Yes, static methods are not the hottest things on earth but neither are global `ccpXX` methods. Single methods in a `cocos2d::ccpoint` namespace would fulfil my personal feelings but its not common in this codebase (right?). So I'd simply go with the static methods in `CCPoint`.

aberaud:
Please review the last commits for correctness.

@michaelcontento Ok I will rename if it's the convention in cocos2d. However where to draw the line ? Should it be getDot() and getCross(), getPerp() instead of cross(), dot() and perp() ?

For distance() and distanceSq it would be static functions CCPoint::distance(a, b); right ?

michaelcontento:
> Please review the last commits for correctness.

Looks good to me :+1: 

>  Ok I will rename if it's the convention in cocos2d. However where to draw the line ? Should it be getDot() and getCross(), getPerp() instead of cross(), dot() and perp() ?

Maybe this is something where @minggo can/should help us out. 

> For distance() and distanceSq it would be static functions CCPoint::distance(a, b); right ?

Hm .. why not a normal member function like:

```
CCPoint a(5, 5), b(10, 10);
float distance = a.distance(b);
```

Is, at least to me, more readable than:

```
CCPoint a(5, 5), b(10, 10);
float distance = CCPoint::distance(a, b);
```

And there is no drawback/overhead as you need both `CCPoint` objects in both cases. What do you think?

metalbass:
@wagaf-d @michaelcontento I'm still here, i didn't delete my account nor any message :S
Do you see this? Have you seen my pull request? (operators +=, -=, *= and /=)

All the methods that are not getters, but functions should not include the word get, which implies constant access time (return of a variable) when this methods might not have constant time.

cocos2dx/cocoa/CCGeometry.cpp
```diff
@@ -61,21 +93,46 @@ bool CCPoint::equals(const CCPoint& target) const
         && (fabs(this->y - target.y) < FLT_EPSILON);
 }
 
+bool CCPoint::fuzzyEquals(const CCPoint& b, float var) const
+{
+    if(x - var <= x && b.x <= x + var)
```
minggo: x - var <= b.x?

minggo:
> Ok I will rename if it's the convention in cocos2d. However where to draw the line ? Should it be getDot() and getCross(), getPerp() instead of cross(), dot() and perp() ?
> 
> > Maybe this is something where @minggo can/should help us out.

Usually the function name should be an action, such add, sub, getLength and so on.

minggo:
So i think some function names should be changed, such as
length -> getLength

metalbass:
@minggo the word "get" implies constant time (return of a variable) and those methods might have higher cost (specially those with sqrt calls)

cocos2dx/cocoa/CCGeometry.cpp
aberaud: Yup :+1:

aberaud:
Suggested changes are implemented in the last commit.

I also renamed the PR to better reflect the actual change.

cocos2dx/cocoa/CCGeometry.h
```diff
@@ -45,9 +49,82 @@ class CC_DLL CCPoint
...
     void setPoint(float x, float y);
     bool equals(const CCPoint& target) const;
+    bool fuzzyEquals(const CCPoint& target, float variance) const;
+
+    inline float getLength() const {
+        return sqrtf(x*x + y*y);
+    };
+
+    inline float getLengthSq() const {
+        return dot(*this); //x*x + y*y;
```
minggo, line 81: Does `getLengthSq` reflect the meaning?
What's Sq short for?

cocos2dx/cocoa/CCGeometry.h
```diff
@@ -45,9 +49,82 @@ class CC_DLL CCPoint
...
+
+    inline float getDistanceSq(const CCPoint& other) const {
+        return (*this - other).getLengthSq();
+    };
+
+    inline float getDistance(const CCPoint& other) const {
+        return (*this - other).getLength();
+    };
+
+    float getAngle(const CCPoint& other) const;
```
minggo, line 110: May be the name of `getAngle` and `getDistance` are not so suitable, because it is not a property of a point. It is a relationship between two points.

I am not sure.
What's your opinion?

cocos2dx/cocoa/CCGeometry.h
```diff
@@ -45,9 +49,82 @@ class CC_DLL CCPoint
...
+
+    inline CCPoint getPerp() const {
+        return CCPoint(-y, x);
+    };
+
+    inline CCPoint getRPerp() const {
+        return CCPoint(y, -x);
+    };
+
+    inline CCPoint project(const CCPoint& other) const {
```
minggo, line 148: Other project to this?

cocos2dx/cocoa/CCGeometry.h
aberaud, line 81: ccpLengthSq and ccpDistanceSq already existed.
Sq is for squared. The squared distance is cheaper to compute and often enough to compare distances.

cocos2dx/cocoa/CCGeometry.h
aberaud, line 110: As @metalbass mentioned, p.getDistance(other) is more convenient and natural to use than a static method.

Is is also more consistent with getAngle() with no params (same as angle with ccp(0, 1)) and getLength() (distance to origin).

cocos2dx/cocoa/CCGeometry.h
aberaud, line 148: Do you mean it should ? Currently this projects on other.

cocos2dx/cocoa/CCGeometry.h
```diff
@@ -45,9 +49,82 @@ class CC_DLL CCPoint
...
+    inline CCPoint rotate(const CCPoint& other) const {
+        return CCPoint(x*other.x - y*other.y, x*other.y + y*other.x);
+    };
+
+    inline CCPoint unrotate(const CCPoint& other) const {
+        return CCPoint(x*other.x + y*other.y, y*other.x - x*other.y);
+    };
+
+    inline CCPoint normalize() const {
+        return *this / getLength();
```
minggo, line 178: Should determine if the value of getLength() is 0?

cocos2dx/cocoa/CCGeometry.cpp
```diff
@@ -84,6 +139,32 @@ bool CCPoint::equals(const CCPoint& target) const
...
+}
+
+CCSize CCSize::operator*(float a) const
+{
+    return CCSize(this->width * a, this->height * a);
+}
+
+CCSize CCSize::operator/(float a) const
+{
+    return CCSize(this->width / a, this->height / a);
```
minggo, line 167: Should checkout that a is not equal to 0?

minggo:
@wagaf-d 
Could you add some comment for new functions, such as `lerp`?
Thank you.

cocos2dx/cocoa/CCGeometry.cpp
aberaud, line 167: There's not much we could do, and checking would add some performance overhead, so let the user check for the 0-case himself (the user has more context information so he would be able to do something more appropriate in this case).
The result will also be consistent with a normal float division by 0 (but with two floats), which users may expect.

minggo, line 167: How about add assert?

cocos2dx/cocoa/CCGeometry.h
```diff
@@ -45,9 +49,156 @@ class CC_DLL CCPoint
...
+
+    /** Calculates the projection of this over other.
+     @return CCPoint
+     @since v2.1.4
+     */
+    inline CCPoint project(const CCPoint& other) const {
+        return other * (dot(other)/other.dot(other));
+    };
+
+    /** Rotates two points.
```
minggo: > Rotates two points.

Could you describe it more detail?
Thank you.

aberaud:
To be honest, I'm still not convinced that doing an assert would be of any benefit for the user. 2D vector division will be heavily used for many kinds of computation in game engines, and having a condition check for every single division may prevent the compiler and the CPU to optimize the computation. However since CCAssert are disabled in release mode I pushed the change anyway.

minggo:
Thank you.

cty41:
now we still lack of some operator, like +=, -=, *=, /=, !=,I will add code later

PR #2505, opened 2013-05-02
Emscriptenate

j4m3z0r:
Patch to add Emscripten support. Basic changes:
- Add pre-build freetype, libpng, etc for Emscripten
- Add submodule for Emscripten
- Add platform directory for Emscripten
- Add code to push vertex data to server-side buffers as WebGL does not support client-side buffers.
- Lots of little fixes.

Current status is that it can run the TestCpp sample. Almost all tests run, still some visual glitches, but probably good enough to start building on top of now.

To build, run

```
make -f Makefile.emscripten
```

JS, HTML, data files will be put into the standard proj.emscripten/bin/release directories. Note that these files need to be hosted on a webserver to run, as they depend on XHRs.

Ricardo Quesada has taken a look at the diff, and we agree that this is a great facility to add to Cocos2D-X. He's suggested using object composition for the vertex buffer code. Let's discuss on the mailing list, and if we need to re-work it, I'll submit another pull. I think this follows the structure of the current code though, so would like to start with this.

### DIFF
--- a/README.mdown
+++ b/README.mdown
@@ -24,6 +24,8 @@ Supported Platforms
    * Windows: stable, well tested on WinXP/Vista/Win7. Please upgrde the drive
      of your video card if you meet problems on OpenGL functions
    * Linux: support but not very stable.
+   * Emscripten: Alpha-level. Most features implemented. Needs testing.
+     Contributed by Zynga staff.
 
 You can visit our continuous integration system http://ci.cocos2d-x.org to
 check the stability on edge version.


minggo:
@j4m3z0r 
What's emscriptenate port?
Bring cocos2d-x to web browser?

rtissera:
This is support to write game in pure C++ through Cocos2D-X and publish as HTML5 application thanks to Emscripten.
Emscripten is a highly efficient C++ --> JS compiler based on LLVM and Google Closure Compiler.

j4m3z0r:
@minggo 
Basically, yes. Emscripten is a C++ compiler that emits JavaScript. With this patch you can take an existing game written in C++ using cocos2d-x and deploy it to the web with very few modifications.

minggo:
Awesome.

@j4m3z0r 
Could you resolve conflicts?

j4m3z0r:
Hi Minggo,

I went through and did a rebase against cocos2dx/master, and have tested that I can merge my changes against that tree now and that they build and run.

Will that work for you?

minggo:
It can be merged now.
Thank you.
Could you write a document to describe how to run it on web browser?

j4m3z0r:
Absolutely. Where is the best place for it? I can add a README, or would you rather something on a Wiki? The instructions are actually very short.

j4m3z0r:
Ok, I see you've got a wiki linked, and the README also enumerates targetable platforms. I'll add a line about Emscripten support to the README and register for the wiki so I can add some content there.

minggo:
I have added an item for `EMSCRIPTEN` in [cocos2d-x wiki page](http://www.cocos2d-x.org/projects/cocos2d-x/wiki/Developers_Manual).

Now the code uses macro `EMSCRIPTEN` to do EMSCRIPTEN special operation. It makes codes not so elegant, any other way to do it more elegant?

j4m3z0r:
I thought a lot about that. I'm not happy about them either.

The main use of the #ifdef EMSCRIPTEN blocks is to transfer data to server-side OpenGL buffers, rather than the client-side behavior that was originally in there. There are two ways we could avoid those that I can think of:
1. Switch everything to using server-side arrays (ie, remove the non EMSCRIPTEN blocks)
2. Re-factor the code to use a separate graph traversal pass and rendering pass. We could then have an EmscriptenRenderer that knew to have this behavior.

I asked Ricardo his thoughts on (1), and he felt that it would slow down some calls, so we shouldn't do that -- fair enough.

FWIW, I think (2) is the right answer, but that is a significant refactoring effort that would likely destabilize the code for a while, so I went with the safer (if slightly uglier) option.

minggo:
Got it.
Thank you.

minggo:
I am sorry. I modified AUTHORS file which cause conflict to this PR.
Could you please resolve it, i will merge it.

j4m3z0r:
Working on this now.

j4m3z0r:
Do you want a new rebase, or will a merge suffice?

minggo:
I will merge this PR, and you can refactor it later in other PR.
What's your opinion?

j4m3z0r:
Ok, so that changeset should resolve the conflict with AUTHORS (I re-rebased), and adds a note to README about the Emscripten support.

I'd say go ahead and merge that: whilst not perfect, it will be useful to people looking to deploy to the web.

Regarding the refactor -- let's discuss further. I would rather get some more fixes to improve performance and add features first (for instance, the current code does not do sound -- would like to get that added).

j4m3z0r:
Ok, I've now added basic getting started documentation at

http://www.cocos2d-x.org/projects/cocos2d-x/wiki/Emscriptenate

BTW, this probably belongs under a "Setting up development environments" section called "Emscripten". Did you want to move the page?

Let me know once this is merged.

minggo:
Because AUTHORS is frequently modified, so i resolve it in #2521.
I wanted to send a PR to you, but failed.

Please review it or you resolved conflict by yourself?

j4m3z0r:
The AUTHORS file in your Emscriptenate branch looks fine to me -- can you just merge into master from there?

I'm new to git, so if you need me to do something, send clear instructions. Happy to do whatever you need, but keep it simple :)

minggo:
I wanted to send PR to you, but failed. I can not select you when sending pull request.
How about send you the merged AUTHORS to you?

minggo:
I can not find your email address.
How about send me a letter: xiaoming.zhang@cocos2d-x.org?

minggo:
I merged code, please synchronize your codes then have a test.

j4m3z0r:
I just pulled down master and built and it seems fine to me. Thanks!

PR #2507, opened 2013-05-02
Simplify CCPhysicsNode

sbc100:
No need to override nodeToParentTransform completely only ensure
that position and rotation are up-to-date before performing normal
CCNode::nodeToParentTransform.

This change reduces the ammount of cocos2dx and box2d specific code
significantly.

Tested with TestCpp in box2d and chipmunk configuration.

### DIFF
--- a/extensions/physics_nodes/CCPhysicsSprite.cpp
+++ b/extensions/physics_nodes/CCPhysicsSprite.cpp
@@ -167,26 +167,40 @@ void CCPhysicsSprite::setIgnoreBodyRotation(bool bIgnoreBodyRotation)
 // Override the setters and getters to always reflect the body's properties.
 const CCPoint& CCPhysicsSprite::getPosition()
 {
-    updatePosFromPhysics();
+    updatePositionFromPhysics();
     return CCNode::getPosition();
 }
 
 void CCPhysicsSprite::getPosition(float* x, float* y)
 {
-    updatePosFromPhysics();
+    updatePositionFromPhysics();
     return CCNode::getPosition(x, y);
 }
 
 float CCPhysicsSprite::getPositionX()
 {
-    updatePosFromPhysics();
+    updatePositionFromPhysics();
     return m_obPosition.x;
 }
 
 float CCPhysicsSprite::getPositionY()
 {
-    updatePosFromPhysics();
+    updatePositionFromPhysics();
     return m_obPosition.y;
+}
+
+float CCPhysicsSprite::getRotation()
+{
+    updateRotationFromPhysics();
+    return CCNode::getRotation();
+}
+
+// returns the transform matrix according the Box2D Body values
+CCAffineTransform CCPhysicsSprite::nodeToParentTransform()
+{
+    updatePositionFromPhysics();
+    updateRotationFromPhysics();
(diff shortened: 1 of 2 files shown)


dumganhar:
<pre>Failed — The Travis build failed (Details)</pre>

@sbc100 ,What's Travis?
It seems to be a command line building environment which could check whether your PR was built successfully, right?

sbc100:
Travis is a free continuous integration solution for open source projects hosted on github.

I created a PR to add travis support to cocos but it hasn't landed yet:

https://github.com/cocos2d/cocos2d-x/pull/2459

Once the travis.yml file does land then future PR's (like this one) will me marks as Green if the pass all the build/test commands.

The only real limitation is that the build and test has to all run on linux.  Right now I'm just testing the linux and NaCl builds.

Feel free to merge the other PR first and then I will rebase this and it should go green :)

sbc100:
The reason the travis build is failing for this branch is the the .travis.yml file is not present so travis is assuming default (ruby) project rather than a Make-base linux project which cocos2dx is.

dumganhar:
OK, it seems a cool feature. :) 
I will also check #2459 soon. Thanks.

sbc100:
Rebased.

extensions/physics_nodes/CCPhysicsSprite.cpp
```diff
@@ -201,56 +215,38 @@ void CCPhysicsSprite::setCPBody(cpBody *pBody)
...
 }
 
-// returns the transform matrix according the Chipmunk Body values
-CCAffineTransform CCPhysicsSprite::nodeToParentTransform()
-{
-    // Although scale is not used by physics engines, it is calculated just in case
-	// the sprite is animated (scaled up/down) using actions.
-	// For more info see: http://www.cocos2d-iphone.org/forum/topic/68990
-	cpVect rot = (m_bIgnoreBodyRotation ? cpvforangle(-CC_DEGREES_TO_RADIANS(m_fRotationX)) : m_pCPBody->rot);
-	float x = m_pCPBody->p.x + rot.x * -m_obAnchorPointInPoints.x * m_fScaleX - rot.y * -m_obAnchorPointInPoints.y * m_fScaleY;
```
dumganhar, line 240: These lines were deleted, do we need to consider scale and skew property like `setPostion` and `setRotation`?

dumganhar, line 240: Do you still focus this PR?

sbc100, line 240: Sorry for the late reply.  Yes, I think this change should be merged.  I'm pretty sure that since we no longer have a custom nodeToParentTransform, but instead use the base CCNode one that all the scale and scew properties will be correctly honored.
Indeed if it had been implemented like this to begin with than I don't think the issue being refereed to would have occurred at all.

I will verify by running the test case in the forrum referenced forum topic and get back to you to confirm.

sbc100, line 240: I verified that the issue refereed to in http://www.cocos2d-iphone.org/forum/topic/68990 does not occur after this change.

Hopefully this PR is good to go now.

PR #2510, opened 2013-05-03
Let CCProgressTimer truly supports setOpacity()

liukun:
Currently, CCProgressTimer does not change opacity because its customized draw(). So I modified updateColor() and let setOpacity() call it.

Similar to cocos2d-iphone commit https://github.com/cocos2d/cocos2d-iphone/commit/8866f46c4169169dce89997463a75f58a8bcb305

### DIFF
--- a/cocos2dx/misc_nodes/CCProgressTimer.cpp
+++ b/cocos2dx/misc_nodes/CCProgressTimer.cpp
@@ -149,6 +149,30 @@ void CCProgressTimer::setReverseProgress(bool reverse)
     }
 }
 
+void CCProgressTimer::setOpacity(GLubyte opacity)
+{
+    CCNodeRGBA::setOpacity(opacity);
+    updateColor();
+}
+
+void CCProgressTimer::setColor(const ccColor3B& color3)
+{
+    CCNodeRGBA::setColor(color3);
+    updateColor();
+}
+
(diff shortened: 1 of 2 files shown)


cocos2dx/misc_nodes/CCProgressTimer.cpp
```diff
@@ -203,6 +227,13 @@ void CCProgressTimer::updateColor(void)
     if (m_pVertexData)
     {
         ccColor4B sc = m_pSprite->getQuad().tl.colors;
+        float op = _displayedOpacity / 255.f;
+        sc.a *= op;
```
minggo, line 231: *= -> = ?

liukun, line 231: @minggo hmm, how about changing to

```
    ccColor4B sc = m_pSprite->getQuad().tl.colors;
    sc.a = _displayedOpacity;
    if (m_pSprite->getTexture()->hasPremultipliedAlpha()) {
        float op = _displayedOpacity / 255.f;
        sc.r *= op;
        sc.g *= op;
        sc.b *= op;
    }
```

minggo, line 231: sc = (m_pSprite->getColor, getOpacity())

getOpaicity returns real opacity, not displayed opacity, and color from getColor().

PR #2526, opened 2013-05-07
Fix for CCDrawNode displaying incorrect colors.

danielhaaser:
CCDrawNode doesn't display colors correctly when not using VAO.  The normalization parameter of the glVertexAttribPointer call for the color attribute was set incorrectly.


dumganhar:
Thanks.

PR #2527, opened 2013-05-07
Fixing CCScale9Sprites when loaded from CCB files

Clarinexus:
CCScale9Sprites that were loaded from CCB files could not be set to a color or opacity other than white and full opacity. This fix allows CCB-loaded sprites to have dynamic color and opacity again.

### DIFF
--- a/extensions/GUI/CCControlExtension/CCScale9Sprite.h
+++ b/extensions/GUI/CCControlExtension/CCScale9Sprite.h
@@ -308,7 +308,9 @@ public:
      */
     virtual bool isOpacityModifyRGB(void);
     virtual void setOpacity(GLubyte opacity);
+	virtual GLubyte CCScale9Sprite::getOpacity();
     virtual void setColor(const ccColor3B& color);
+	virtual const ccColor3B& CCScale9Sprite::getColor();
 
     virtual bool updateWithBatchNode(CCSpriteBatchNode* batchnode, CCRect rect, bool rotated, CCRect capInsets);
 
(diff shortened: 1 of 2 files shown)


dumganhar:
Yep, your fix is right.

> CCScale9Sprites that were loaded from CCB files could not be set to a color or opacity other than white and full opacity. 

But the _setColor_ and _setOpacity_ did work. Why did you say that `could not be set to a color or opacity other than white and full opacity.`?

Clarinexus:
The functions setColor and setOpacity worked only for programmatically-added CCScale9Sprites, but ones that were loaded in through CCB files couldn't be changed from white or fully opaque, and weren't taking the CCB file's color/opacity settings on creation.

Clarinexus:
The real problem was that updateWithBatchNode was calling getColor and getOpacity and then setting that color/opacity at the end, causing CCB-loaded sprites to go white/opaque.

extensions/GUI/CCControlExtension/CCScale9Sprite.h
```diff
@@ -308,7 +308,9 @@ class CCScale9Sprite : public CCNodeRGBA
      */
     virtual bool isOpacityModifyRGB(void);
     virtual void setOpacity(GLubyte opacity);
+	virtual GLubyte CCScale9Sprite::getOpacity();
```
dumganhar, line 311: `CCScale9Sprite::` should be removed.

extensions/GUI/CCControlExtension/CCScale9Sprite.h
```diff
@@ -308,7 +308,9 @@ class CCScale9Sprite : public CCNodeRGBA
...
     virtual void setColor(const ccColor3B& color);
+	virtual const ccColor3B& CCScale9Sprite::getColor();
```
dumganhar, line 313: CCScale9Sprite:: should be removed.

PR #2536, opened 2013-05-08
Added win32 support

sunzhuoshi:


dumganhar:
Merged, thanks.
Next step: we should also update the jsbcc.sh script by using just one command `find <dir> -name '*.js' | jsbcc`.

PR #2538, opened 2013-05-08
Fix ccCArray

edbartley:
Please see my comments on this commit for an explanation of the issue: https://github.com/cocos2d/cocos2d-x/commit/076f38c5dcc7b0d5195145dac42392eb1bcc7939


brawsome:
Hey Ed,

If this works then go for it. I hadn't realized it had actually been pulled already!

Thanks,
Andrew

PR #2539, opened 2013-05-08
Added const qualification to some CCNode's getters

ghost:
As we want to be able to copy construct CCNode's subclasses we want to be able to do this:

```
class SubNode : public cocos2d::CCNode
{
public:
    SubNode(const SubNode& other)
    {
        this->setContentSize(other.getContentSize());
    }
};
```

So i've added constness to the following getters:

<ul>
<li>getTag()</li>
<li>getContentSize()</li>
<li>getChildrenCount()</li>
</ul>

Of course this are not the only methods that need const added, but seem the most obvious ones.
We cannot add const to all of them because implementations of them would need to be const as well.

### DIFF
--- a/cocos2dx/base_nodes/CCNode.cpp
+++ b/cocos2dx/base_nodes/CCNode.cpp
@@ -326,7 +326,7 @@ CCArray* CCNode::getChildren()
     return m_pChildren;
 }
 
-unsigned int CCNode::getChildrenCount(void)
+unsigned int CCNode::getChildrenCount(void) const
 {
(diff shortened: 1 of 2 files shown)


minggo:
I will review it ASAP.
Thank you @metalgit92.

dumganhar:
You should also modify `CCScrollView::getContentSize` since it overrides the parent's method.

gelldur:
Yeah finally const getters :)

ghost:
@dumganhar I'll make another pull request to add more constness and begin adding methods to it.
@gelldur I've been missing them a lot too :)

PR #2540, opened 2013-05-08
Fix for broken ccArrayGetIndexOfObject

jotel:
The optimisation introduced in 076f38c has broken ccArrayGetIndexOfObject which fails  now (i.e. returns CC_INVALID_INDEX) for object inside array but with indexes > 0. And it has nasty effect for all  CCTouchHandlers not unregistered and still handling events even if no longer existing... 

Here is a fix for that.

Regards
Jaro


dumganhar:
Thanks. It was also fixed at https://github.com/cocos2d/cocos2d-x/pull/2538.
I'll merge one of them. :)

edbartley:
Pick me! Pick me! I was first! :)

dumganhar:
@edbartley, yes, you are first. :+1: 
I also update the author for you guys. https://github.com/cocos2d/cocos2d-x/commit/9820021d0fdc92962b5595a14f2bfe880265963f

PR #2542, opened 2013-05-09
Initial shared_ptr for CCObject

ricardoquesada:
a Shared Pointer that lets you use CCObject objects with STL containers
like std::map, std::vector, etc.

### DIFF
--- a/cocos2dx/shared_ptr.h
+++ b/cocos2dx/shared_ptr.h
@@ -1,0 +1,114 @@
+/****************************************************************************
+ Copyright (c) 2013      Zynga Inc.
+
+ http://www.cocos2d-x.org
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ ****************************************************************************/
+
+#ifndef __cocos2dx__shared_ptr__
+#define __cocos2dx__shared_ptr__
+
+#include "cocoa/CCObject.h"
+#include "platform/CCPlatformMacros.h"
+
+NS_CC_BEGIN
+
+//
+// code based on the following articles:
+// http://techieworldz.blogspot.com/2012/04/simple-sharedptr-implementation.html
+// http://en.wikipedia.org/wiki/Auto_ptr
+//
+
+/** A simple shared pointer to be used only with CCObject and subclasses.
+ It allows CCObject objects to be used in STL containers like std::map and std::vector.
+ Instead of implementing its own reference counting, it uses the referece counting implemented by the CCObject.
+ */
+class shared_ptr
+{
+
+public:
+	//
+	// 20.4.5.1 construct/copy/destroy:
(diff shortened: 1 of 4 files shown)


cocos2dx/shared_ptr.h
```diff
@@ -0,0 +1,114 @@
...
+		if(this->m_ptr != rhs.m_ptr)
+		{
+			if(this->m_ptr != NULL)
+			{
+				//decrement this counter
+				m_ptr->release();
+			}
+			//assing rhs and increment counter
+			this->m_ptr = rhs.m_ptr;
+			this->m_ptr->retain();
```
minggo, line 82: Should checkout validation of `this->m_ptr`?

cocos2dx/shared_ptr.h
```diff
@@ -0,0 +1,114 @@
...
+	shared_ptr& operator=(shared_ptr& rhs)
+	{
+		if(this->m_ptr != rhs.m_ptr)
+		{
+			if(this->m_ptr != NULL)
+			{
+				//decrement this counter
+				m_ptr->release();
+			}
+			//assing rhs and increment counter
```
minggo, line 80: assing -> assign?

cocos2dx/shared_ptr.h
```diff
@@ -0,0 +1,114 @@
...
+		this->m_ptr = rhs.m_ptr;
+
+		//increment the counter as we have one more object now
+		this->m_ptr->retain();
+	}
+
+	/** Assignment */
+	shared_ptr& operator=(shared_ptr& rhs)
+	{
+		if(this->m_ptr != rhs.m_ptr)
```
sergey-shambir, line 73: Instead of check with if, you can retain rhs object first and then release lhs, copied to local variable. Situation when pointer assigned to itself is rare, so it better to optimize code for case when self-assign does not happen.

cocos2dx/shared_ptr.h
```diff
@@ -0,0 +1,114 @@
...
+			this->m_ptr = rhs.m_ptr;
+			this->m_ptr->retain();
+		}
+		return *this;
+	}
+
+	/** destructor. It will call "release" on the shared object. */
+	~shared_ptr()
+	{
+		m_ptr->release();
```
sergey-shambir, line 90: No check for null

cocos2dx/shared_ptr.h
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	/** returns a reference to the shared object */
+	CCObject& operator*() const {
+		return *m_ptr;
```
sergey-shambir, line 95: Add assertion that object is not null. On android, debugging is hard since by default Eclipse doesn't show stack trace, so assertion is move verbose than SIGSEGV.

cocos2dx/shared_ptr.h
```diff
@@ -0,0 +1,114 @@
...
+	};
+
+	/** returns the shared object (the pointer) */
+	CCObject* operator->() const {
+		return m_ptr;
+	}
+
+	/** returns the shared object (the pointer) */
+	CCObject* get() const  {
```
sergey-shambir, line 104: Maybe it's more handy to overload operator T \* (operator CCObject \* in this case)?

sergey-shambir:
I don't like non-template class. Also shared_ptr is unexpected name, something like CCSharedPointer or just CCPointer is more understandable.

Moreover, user can type std::shared_ptr instead of cocos2d::shared_ptr and get wrong behavior.

sergey-shambir:
Also please remove any coments except doxygen and license.

SevInf:
Agree with @sergey-shambir, while this class really helps to host CCObject inside STL containers, without argument  it will produce not very readable code with a lot of typecasts. In C++11 we can add a static assert to ensure that template argument is a subclass of CCObject:

``` c++
template<typename T> class CCSharedPtr {
    static_assert(std::is_base_of<CCObject, T>::value, "CCSharedPtr must be used only with CCObject subclass");
    T* operator->() const {...};
};
```

In C++03 we can use boost for static asserts and `is_base_of`. 
Alternatively, we can create generic shared_ptr template with embedded reference counting and specialize it for CCObjects to use retain/release. This way we can have a smart pointer for all kind of pointers, no matter if they inherit CCObject or not.

ricardoquesada:
and here is another implementation of a similar object
http://www.codeproject.com/Articles/64111/Building-a-Quick-and-Handy-Reference-Counting-Clas

ricardoquesada:
@sergey-shambir 
yep. Probably `cocos2d::shared_ptr` is confusing.

@sergey-shambir @SevInf 
newbie question: why should it support templates if it only supports `CCObjects` ? thanks.

walzer:
cc::SharedPtr
cc::WeakPtr

sergey-shambir:
Without templates users will face with two problems:
- Pointer should be explicitly casted to target class before using.
- cc::SharedPtr that should contain CCNode can be easily initialized with cc::SharedPtr that contains CCDictionary. This is common mistake for Objective-C programmers, but in Objective-C ducky typing often hides effect of such mistake. For example, programmer can cast NSArray to NSDictionary and call message `count`; the same trick causes segfault in C++.

ricardoquesada:
@sergey-shambir 
Thanks. There will be a problem with non-virtual functions as well. Yes, the template is needed.
I will rewrite this class.

ricardoquesada:
closing it. This should not be merged.

PR #2544, opened 2013-05-09
Fix custom font loading on Android for resource mapped font files

BodbDearg:
1: Use the CCFileUtils::fullPathForFilename function to lookup the resolved path for the font in the Android specific function, CCImage::getBitmapFromJava.

2: Remove a warning previously added in pull request 2422 (https://github.com/cocos2d/cocos2d-x/pull/2422) in CCFileUtils::fullPathForFilename in order to support change (1) above. If we pass just a system font name to CCFileUtils::fullPathForFilename such as 'Arial' (instead of a path) then it should just return 'Arial' and not generate a warning. We can't assume that when a path (or handle) given to CCFileUtils::fullPathForFilename isn't found within Cocos2dx that it doesn't exist or isn't valid. Depending on the platform there might be files or folders or handles that could be accessible to the app which might exist outside of the knowledge of Cocos2dx. The correct place to warn about missing resource files would be at the point where the resource loading itself fails; e.g if a texture or sound fails to load for example.

### DIFF
--- a/cocos2dx/platform/CCFileUtils.cpp
+++ b/cocos2dx/platform/CCFileUtils.cpp
@@ -636,8 +636,6 @@ std::string CCFileUtils::fullPathForFilename(const char* pszFileName)
         }
     }
     
-    CCLOG("cocos2d: fullPathForFilename: No file found at %s. Possible missing file.", pszFileName);
-    
     // The file wasn't found, return the file name passed in.
     return pszFileName;
 }
--- a/cocos2dx/platform/android/CCImage.cpp
+++ b/cocos2dx/platform/android/CCImage.cpp
@@ -28,6 +28,7 @@ THE SOFTWARE.
 #include "platform/CCImageCommon_cpp.h"
 #include "platform/CCPlatformMacros.h"
 #include "platform/CCImage.h"
+#include "platform/CCFileUtils.h"
 #include "jni/JniHelper.h"
 
 #include <android/log.h>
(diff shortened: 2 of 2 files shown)


cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -636,8 +636,6 @@ unsigned char* CCFileUtils::getFileDataFromZip(const char* pszZipFilePath, const
         }
     }
     
-    CCLOG("cocos2d: fullPathForFilename: No file found at %s. Possible missing file.", pszFileName);
```
dumganhar, line 639: Any better solution when loading a missing file?

cocos2dx/platform/android/CCImage.cpp
```diff
@@ -65,6 +66,17 @@ class BitmapDC
             return false;
         }
 
+        // Do a full lookup for the font path using CCFileUtils in case the given font name is a relative path to a font file asset,
+        // or the path has been mapped to a different location in the app package:
+        std::string fullPathOrFontName = CCFileUtils::sharedFileUtils()->fullPathForFilename(pFontName);
+
+		// If the path name returned includes the 'assets' dir then that needs to be removed, because the android.content.Context
+		// requires this portion of the path to be omitted for assets inside the app package.
+		if (fullPathOrFontName.find_first_of("assets/") == 0)
```
dumganhar, line 75: This fix looks great. Thanks. I'll test it soon.

dumganhar, line 75: We should use `find` instead of find_first_of, they are different.
Please refer to http://www.cplusplus.com/reference/string/string/find_first_of/

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/2557.

cocos2dx/platform/CCFileUtils.cpp
BodbDearg, line 639: I think the best thing would probably be to issue the warning wherever resource loading code fails. For example, in texture or sound loading code when a texture file or audio file is not found... That way we could also add extra context/information to the error because we know more about how the file is being used in that situation.

cocos2dx/platform/android/CCImage.cpp
BodbDearg, line 75: Thanks, makes sense!

BodbDearg:
Thanks for the merging this in!

PR #2546, opened 2013-05-09
CCLabelTTF supports for shadow and stroke

carlomorgantinizynga:
CCLabelTTF now supports text with shadow and stroke

### DIFF
--- a/cocos2dx/label_nodes/CCLabelTTF.h
+++ b/cocos2dx/label_nodes/CCLabelTTF.h
@@ -156,12 +156,15 @@ protected:
     std::string * m_pFontName;
     /** Font size of the label */
     float m_fFontSize;
+    /** label's string */
+    std::string m_string;
     
     /** font shadow */
     bool    m_shadowEnabled;
     CCSize  m_shadowOffset;
     float   m_shadowOpacity;
     float   m_shadowBlur;
+    
     
     /** font stroke */
     bool        m_strokeEnabled;
@@ -171,7 +174,7 @@ protected:
     /** font tint */
     ccColor3B   m_textFillColor;
 
-    std::string m_string;
+    
 };
 
 


cocos2dx/label_nodes/CCLabelTTF.cpp
```diff
@@ -255,30 +281,251 @@ void CCLabelTTF::setFontName(const char *fontName)
 bool CCLabelTTF::updateTexture()
 {
     CCTexture2D *tex;
-    
-    // let system compute label's width or height when its value is 0
-    // refer to cocos2d-x issue #1430
     tex = new CCTexture2D();
+    
+    assert(tex);
```
dumganhar: Why to make an assert here? If tex is NULL, it will return false before.

cocos2dx/platform/android/java/project.properties
```diff
@@ -12,4 +12,4 @@
 
 android.library=true
 # Project target.
-target=android-8
+target=android-17
```
dumganhar, line 15: android-8 --> android-17 ?? Why?

minggo:
> > I'm not convinced with the initWithStringShadowStroke name. In fact, I'm not convinced with the initWithString   name either.
> > initWithString was miss-ported from Objective-C's:
> > - (id) initWithString:(NSString_)string fontName:(NSString_)name fontSize:(CGFloat)size dimensions:(CGSize)dimensions hAlignment:(CCTextAlignment)hAlignment vAlignment:(CCVerticalTextAlignment)vAlignment lineBreakMode:(CCLineBreakMode)lineBreakMode;
> >   I would use another name of this method. Perhaps initWithString()... and just overload it.
> >   A better name would be init()... but I guess we can't change that

I think we should use init() as possible.

minggo:
Oh, init is used for other purpose.
So i think initWithString is a good idea.

cocos2dx/include/ccTypes.h
```diff
@@ -49,6 +51,14 @@ of this software and associated documentation files (the "Software"), to deal
     ccColor3B c = {r, g, b};
     return c;
 }
+
+/** returns YES if both ccColor3B are equal. Otherwise it returns NO.
```
minggo: YES -> true
NO -> false

zcgit:
how about supporting BOLD text?

dumganhar:
Do you see that <pre> Failed — The Travis CI build failed (Details)</pre>?
Please refers to https://travis-ci.org/dumganhar/cocos2d-x/jobs/7047552

cocos2dx/include/ccTypes.h
carlomorgantinizynga: this should now be fixed

minggo:
@carlomorgantinizynga 
How about renaming `initWithStringShadowStroke` to `initWithString`? Because all other functions are named `initWithString`.

cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxBitmap.java
```diff
@@ -116,19 +125,115 @@ public static void createTextBitmap(String pString, final String pFontName, fina
 		Cocos2dxBitmap.initNativeObject(bitmap);
 	}
 
-	private static Paint newPaint(final String pFontName, final int pFontSize, final int pHorizontalAlignment) {
+	public static void createTextBitmapShadowStroke(String pString,  final String pFontName, final int pFontSize,
+													final float fontTintR, final float fontTintG, final float fontTintB,
+													final int pAlignment, final int pWidth, final int pHeight, final boolean shadow,
+													final float shadowDX, final float shadowDY, final float shadowBlur, final boolean stroke,
+													final float strokeR, final float strokeG, final float strokeB, final float strokeSize) {
```
dumganhar, line 108: The implementation of `createTextBitmap` is a subset of  `createTextBitmapShadowStroke`. There are some duplicated code in these two functions.
So it's better to keep only one implementation, I mean, just one function `createTextBitmap` to support shadow and stroke.

carlomorgantinizynga, line 108: createTextBitmapShadowStroke and createTextBitmap now share the same code

cocos2dx/platform/android/java/project.properties
carlomorgantinizynga, line 15: I didn't want to change that file..and if I did, I didn't do it on purpose....what should be the target? 8? Not really familiar with Android stuff.

cocos2dx/label_nodes/CCLabelTTF.cpp
carlomorgantinizynga: assert has now been removed

carlomorgantinizynga:
Hello everyone.
I did fix most of these issues reported already. Please let me know if there is more that needs to be addressed before proceeding with the merge.
Thank you.

cocos2dx/platform/android/java/project.properties
dumganhar, line 15: Yes, it should be `android-8` which means android 2.2 OS. Cocos2d-x now supports android os  from android 2.2.

dumganhar:
Since this PR can't be automatically merged, I have sent you a PR(https://github.com/carlomorgantinizynga/cocos2d-x/pull/1).

cocos2dx/platform/android/CCImage.cpp
```diff
@@ -137,8 +164,78 @@ bool CCImage::initWithString(
...
+{
+	for(int c = 0; c < numPixels; ++c, ++pImageMemory)
+	{
+		// copy the current pixel
+		unsigned int currenPixel =  (*pImageMemory);
+		// swap channels and store back
+		char *pSource = (char *) 	&currenPixel;
+		*pImageMemory = (pSource[0] << 24) | (pSource[3]<<16) | (pSource[2]<<8) | pSource[1];
+	}
+}
```
dumganhar, line 251: How about invoking BitmapDC:: swapAlpha to implement this function?

carlomorgantinizynga:
I've just rebased this with the latest cocos2d-x. This should fix all the merging conflicts you were having.

cocos2dx/platform/android/CCImage.cpp
carlomorgantinizynga, line 251: I'll take a look to this. Also, I've rebased my pull request on the latest cocos2d-x. This should fix all the merge problem you were having. Please let me know.

evan-liu:
I tried this by enable stroke on an white label, whole label turns to black.  

enableStroke(ccc(0, 0, 0), 2, true)

carlomorgantinizynga:
is it C++ or JS? Can you send me the entire init code for the label that doesn't work so I can test it? Thanks!

evan-liu:
It's c++. CCLabelTTF loaded from ccbi. I will test create and init one instead when I get to work.

evan-liu:
Here is my test code  :

```
CCLabelTTF *label = CCLabelTTF::create("Test Stroke", "Verdana", 24);
label->setPosition(ccp(240, 400));
addChild(label);
label->enableStroke(ccc3(0, 0, 0), 3, true);
```

All black. Did I miss any thing before enableStroke?

cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxBitmap.java
```diff
@@ -88,47 +89,124 @@ public static void setContext(final Context pContext) {
...
+		
+		// padding needed when using shadows (not used otherwise)
+		float bitmapPaddingX   = 0.0f;
+		float bitmapPaddingY   = 0.0f;
+		float renderTextDeltaX = 0.0f;
+		float renderTextDeltaY = 0.0f;
+		
+		if ( shadow ) {
+
+			int shadowColor = 0xff7d7d7d;
```
zeyangl, line 131: shadowColor parameter should be exposed as a parameter in CCLabelTTF, otherwise it's impossible to set fully black shadow.

wsgfz:
same with @eidiot.It all goes black or the color I set to stroke.

how is the issue going?
@carlomorgantinizynga

PR #2548, opened 2013-05-10
Update spine extension to upstream version from git://github.com/Esoteri...

lumendes:
Sync spine extension with upstream version.
Upstream at: git://github.com/EsotericSoftware/spine-runtimes.git
Synced with revision ab54939e9be79236425be8ad97d406c839bcf953
Only tested in win32 & android platforms.

### DIFF
--- a/extensions/spine/Bone.h
+++ b/extensions/spine/Bone.h
@@ -28,10 +28,7 @@
 
 #include <spine/BoneData.h>
 
-#ifdef __cplusplus
 namespace cocos2d { namespace extension {
-extern "C" {
-#endif
 
 typedef struct Bone Bone;
 struct Bone {
@@ -57,9 +54,6 @@ void Bone_setToSetupPose (Bone* self);
 
 void Bone_updateWorldTransform (Bone* self, int/*bool*/flipX, int/*bool*/flipY);
 
-#ifdef __cplusplus
-}
 } }
-#endif
 
 #endif /* SPINE_BONE_H_ */
--- a/extensions/spine/Animation.h
+++ b/extensions/spine/Animation.h
@@ -26,10 +26,7 @@
 #ifndef SPINE_ANIMATION_H_
 #define SPINE_ANIMATION_H_
 
-#ifdef __cplusplus
 namespace cocos2d { namespace extension {
-extern "C" {
-#endif
 
 typedef struct Timeline Timeline;
 struct Skeleton;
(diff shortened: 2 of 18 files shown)


lumendes:
Had a look to other pull requests, and apparently there was no need to create a branch for this. Sorry if it's inconvenient.

dumganhar:
Hi, spine parser should be in a namespace(we are now put it in cocos2d::extension).
`c` language has no namespace, so some codes in spine runtime may be conflicted with other modules, for instance, `jsoncpp`.
Therefore, we should change the source code's suffix from `.c` to `.cpp`.

lumendes:
Hi, by spine parser you mean full spine extension or just json files ?
I think all projects should have option to compile C as C++, at least win32 does have.
All .c files contain ifdef for including namespace if compiled as C++.

```
+#ifdef __cplusplus
 namespace cocos2d { namespace extension {
+#endif
```

Does actually renaming .c files to .cpp changes something a compilation level ?

dumganhar:
> I think all projects should have option to compile C as C++, at least win32 does have.

So why not use c++ directly?

lumendes:
I meant all cocos2dx sub-projects, ios, android, etc, in tone of question. I'm not sure if its out-of-the-box, at least in win32 you need to change project settings.
spine upstream dropped the c++ implementation in favor of c only, not sure where to start :)
What is your suggestion ?

dumganhar:
Firstly, I don't know how to use c++ compiler to compile c source in other platforms except win32.
Secondly, namespace is really needed to avoid conflicts.

So IMO, we should change the suffix from `.c` to `.cpp` and change the namespace from `spine` to `cocos2d::extension`.

lumendes:
But that is how it is in this pull request, spine source `.c` and namespace is `cocos2d::extension`
In first comment you said to change `.c` to `.cpp` and now you're saying to change `.cpp` to `.c`.
I'm confused, am I missing something ? :)

dumganhar:
Sorry, my mistake. Update last post.

lumendes:
James, did as you suggested. Compiled fine in win32 and android platforms.
Please have a look into it.

dumganhar:
@lumendes 
Thanks. I'll test it.

extensions/spine/Bone.h
```diff
@@ -28,7 +28,10 @@
 
 #include <spine/BoneData.h>
 
+#ifdef __cplusplus
 namespace cocos2d { namespace extension {
+extern "C" {
+#endif
```
dumganhar: Why all methods in Bone.h should be `extern "C"` while in `Bone.cpp` it doesn't?

lumendes:
@dumganhar yes I believe those were still leftovers from c implementation to allow to compile C as C++.

lumendes:
Latest Travis CI build is passed even if link above still states "Determining merge status"

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/2745.
Thanks. I have added you to AUTHORS.

PR #2560, opened 2013-05-13
Fixed iOS CCMenu bug in Multitouch

kernys:
When below enviornments,
- CCMenu has multiples items (count > 1)
- EAGLView.multipleTouchEnabled = YES;

1) Touch one finger on CCMenuItem.
2) Touch another finger on other CCMenuItem on the CCMenu.
3) Detach first one.
4) m_eState is always kCCMenuStateTrackingTouch
5) This results in disabled CCMenu forever.  Very Critical.

This patchs check when you touch one finger and another finger is about
to touch, cancel first one. and never gone into problem.


dumganhar:
I tested it in TestCpp on iPod touch 4. But I could not reproduce this issue.
When the first finger was released, it'll enter the menu's callback and go to next scene.

dumganhar:
@kernys 
Any comments?

PR #2563, opened 2013-05-14
Fix a potential crash SimpleAudioEngineOpenSL::playEffect

BodbDearg:
Fix a crash in SimpleAudioEngineOpenSL::playEffect which would occur when OpenSLEngine::preloadEffect returned FILE_NOT_FOUND. In the case where the sound fails to preload, we should not attempt to call OpenSLEngine::setEffectLooping on that sound.

### DIFF
--- a/CocosDenshion/android/opensl/SimpleAudioEngineOpenSL.cpp
+++ b/CocosDenshion/android/opensl/SimpleAudioEngineOpenSL.cpp
@@ -88,24 +88,25 @@ void SimpleAudioEngineOpenSL::setEffectsVolume(float volume)
 
 unsigned int SimpleAudioEngineOpenSL::playEffect(const char* pszFilePath, bool bLoop)
 {
-	unsigned int soundID;
-	do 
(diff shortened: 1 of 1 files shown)


BodbDearg:
Additional info:

We ran into the crashes on a Samsung Galaxy S2 (Samsung i9100) - which seems to be caused by or related to the audio driver bug mentioned here: http://www.cocos2d-x.org/issues/1284 

It looks as though when too many sounds are played, OpenSLEngine::preloadEffect will return 'FILE_NOT_FOUND' and result in the crash scenario this change is trying to fix. The code would crash on the call to OpenSLEngine::setEffectLooping in that situation...

This fix won't help issue of the audio not playing on the S2 when too many sounds are played at once, but it will at least avert a crash and allow the user to continue playing the game.

rtissera:
Looks good to me.

minggo:
Thank you.

BodbDearg:
Thank you!

PR #2565, opened 2013-05-14
Fixed CCNotificationCenter documentation and behavior 

gkosciolek:
I observed that CCNotificationCenter uses object parameter in addObserver and postNotification methods. The documentation said that his is custom object passed to notification. However the actual behavior was different. The object parameter was used like  notification sender in Apple's NSNotificationSender - it is not just object passed to notification. It was actually used to determine if the notification will be sent. 

This is good idea as this approach offers more flexibility that just passing the dummy object parameter. However there was bug in that approach. You couldn't addObserver for two notifications that only differs by the object parameter. The observerExisted method never checked the object parameter even if it was used later in postNotification. 

Now the fix. I did two things:
1. Fixed addObserver and observerExisted because it didn't allowed to add observers to notifications that only differs by sender
2. Updatet the API documentation and parameter name from object to sender to reflect it's true purpose.

### DIFF
--- a/cocos2dx/support/CCNotificationCenter.cpp
+++ b/cocos2dx/support/CCNotificationCenter.cpp
@@ -63,7 +63,7 @@ void CCNotificationCenter::purgeNotificationCenter(void)
 //
 // internal functions
 //
-bool CCNotificationCenter::observerExisted(CCObject *target,const char *name)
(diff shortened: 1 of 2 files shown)


gkosciolek:
Closed as it used old branch.

PR #2566, opened 2013-05-15
Submit the change for memcpy done by ThunderSoft

liuyq:
updated according to the review comment from minggo
change to update directly with reference for memcpy problem instead of using function
CCTextureAtlas::updateQuad

Signed-off-by: Yongqin Liuliuyq0307@thundersoft.com


cocos2dx/label_nodes/CCLabelAtlas.cpp
```diff
@@ -144,34 +143,33 @@ void CCLabelAtlas::updateAtlasValues()
...
-        quad.tl.colors = c;
-        quad.tr.colors = c;
-        quad.bl.colors = c;
-        quad.br.colors = c;
-        m_pTextureAtlas->updateQuad(&quad, i);
+        quads[i].tl.colors = c;
+        quads[i].tr.colors = c;
+        quads[i].bl.colors = c;
+        quads[i].br.colors = c;
+        m_pTextureAtlas->updateQuad(NULL, i);
```
minggo: It is strange to invoke `m_pTextureAtlas->updateQuad(NULL, i)`. 
Just invoke `m_bDirty = true`.

cocos2dx/sprite_nodes/CCSprite.h
```diff
@@ -412,6 +412,21 @@ class CC_DLL CCSprite : public CCNodeRGBA, public CCTextureProtocol
...
+
+    /** whether or not the verticles attribute of Quad has been updated */
+    inline virtual bool isQuadVerticlesDirty(void) { return m_bQuadVerticlesDirty; }
+    /** specify if the verticles attribute of Quad has been updated */
+    inline virtual void setQuadVerticlesDirty(bool bDirty) { m_bQuadVerticlesDirty = bDirty; }
+
+    /** whether or not the texCoords attribute of Quad has been updated */
+    inline virtual bool isQuadTexCoordsDirty(void) { return m_bQuadTexCoordsDirty; }
+    /** specify if the texCoords attribute of Quad has been updated */
+    inline virtual void setQuadTexCoordsDirty(bool bDirty) { m_bQuadTexCoordsDirty = bDirty; }
```
minggo: It is better to declare these interfaces to private, and remove virtual.

cocos2dx/textures/CCTextureAtlas.cpp
```diff
@@ -309,9 +309,9 @@ void CCTextureAtlas::updateQuad(ccV3F_C4B_T2F_Quad *quad, unsigned int index)
 
     m_uTotalQuads = MAX( index+1, m_uTotalQuads);
 
-    m_pQuads[index] = *quad;    
-
-
+    if (quad != NULL){
+        m_pQuads[index] = *quad;
+    }
```
minggo: I think we don't have to modify the logic of this function.

cocos2dx/tilemap_parallax_nodes/CCTileMapAtlas.cpp
```diff
@@ -198,35 +198,35 @@ void CCTileMapAtlas::updateAtlasValueAt(const CCPoint& pos, const ccColor3B& val
...
-    quad.tl.colors = color;
-    quad.br.colors = color;
-    quad.bl.colors = color;
+    quad->tr.colors = color;
+    quad->tl.colors = color;
+    quad->br.colors = color;
+    quad->bl.colors = color;
 
-    m_pTextureAtlas->updateQuad(&quad, index);
+    m_pTextureAtlas->updateQuad(NULL, index);
```
minggo: The same suggestion as above one.

cocos2dx/sprite_nodes/CCSprite.h
go3k: inline and virtual can't incompatible at all in C++, this must be modified.

liuyq: I will delete the virtual keyword from the definition,
but there is an existing method definition including inline and virtual
both:

inline virtual bool isDirty(void) { return m_bDirty; }

So I don't think there is any problem with the definition syntax.

Thanks,
Yongqin Liu

2013/5/24 go3k notifications@github.com

liuyq:
close this request

PR #2570, opened 2013-05-16
cpSpacePointQueryFirst returns Shape object when it should return null

smshuja:
When calling space.pointQueryFirst() through JSB, it always returns a shape even if no shape is found and calling any function on this value results in Bad Access error. I have added a check to return null if no shape is found.

### DIFF
--- a/scripting/javascript/bindings/js_bindings_chipmunk_functions.cpp
+++ b/scripting/javascript/bindings/js_bindings_chipmunk_functions.cpp
@@ -4072,9 +4072,9 @@ JSBool JSB_cpSpacePointQueryFirst(JSContext *cx, uint32_t argc, jsval *vp) {
 	if(ret_val) {
 		ret_jsval = opaque_to_jsval( cx, ret_val );
 		JS_SET_RVAL(cx, vp, ret_jsval);
-    } else {
+	} else {
 		JS_SET_RVAL(cx, vp, JSVAL_NULL);
-    }
+	}
 	return JS_TRUE;
 }
 


smshuja:
Any idea why Travis CI build failed? I checked the build log but doesn't seem to be anything related to my commits.

samples/Javascript/Shared
```diff
@@ -1 +1 @@
-Subproject commit bfa9252643fe4463dafdec5cf491cc1c9c406a12
+Subproject commit 999fabb306cddffd0c043ee1cff6e8627ac0df24
```
dumganhar, line 1: Why to change the submodule reference?

scripting/javascript/bindings/generated
```diff
@@ -1 +1 @@
-Subproject commit 07c06b1537159e3f9828039a14ea0a80df05ec7c
+Subproject commit 2f59a8dbc553ea2ef966dc595684f027a6d56931
```
dumganhar, line 1: Why to change the submodule reference?

dumganhar:
Failed — The Travis CI build failed (Details), you could find what the compilation error here(https://travis-ci.org/cocos2d/cocos2d-x/jobs/7213436)

dumganhar:
Your changes seem great, but please revert the modification of the submodule reference.

samples/Javascript/Shared
smshuja, line 1: Sorry for that, I am new to git and github, it happened when I pulled the changes from the cocos2d-x master

smshuja:
I guess I will have to create a branch and send a pull request.

PR #2574, opened 2013-05-17
Fix compiling error of AssetsManager project on windows

minggo:


slobodin:
How about merge?

slobodin:
Thank you!

PR #2575, opened 2013-05-17
Add CURLOPT_SSL_VERIFYPEER option setter to HttpClient

hokuts:
Added CURLOPT_SSL_VERIFYPEER option setter to HttpClient, so that developer specify to allow Self-signed certificate on SSL testing environment.


extensions/network/HttpClient.cpp
```diff
@@ -227,6 +227,10 @@ bool configureCURL(CURL *handle)
     if (code != CURLE_OK) {
         return false;
     }
+    code = curl_easy_setopt(handle, CURLOPT_SSL_VERIFYPEER, CCHttpClient::getInstance()->isSSLVerifyPeer());
```
dumganhar: Does `CURLOPT_SSL_VERIFYHOST` also need to be set like `CURLOPT_SSL_VERIFYPEER`?
And could you pull the latest codes since this PR can't be automatically merged.

PR #2577, opened 2013-05-17
Added CCBFile Skew support and fixed Marmalade compile errors

deemen:
CCBFiles could be skewed in CocosBuilder but this was not showing up in the game. Added two lines to make this happen.

Marmalade wouldn't compile unless a physics integration define was set. Disabled this requirement by only including the two cpp files conditionally.

Added support for other CCBNodeWrappers (like CCBFile) so that a user could potentially add a tinting CCBFile or a CCBFile with other special properties.

### DIFF
--- a/extensions/proj.marmalade/cocos2dx-ext.mkf
+++ b/extensions/proj.marmalade/cocos2dx-ext.mkf
@@ -42,11 +42,6 @@ files
     ("../GUI/CCScrollView")
     "*.h"
     "*.cpp"
-
-    ["spine"]
-    ("../spine")
-    "*.h"
-    "*.cpp"
 }
 
 if {{ defined CC_ENABLE_CHIPMUNK_INTEGRATION or defined CC_ENABLE_BOX2D_INTEGRATION }}


dumganhar:
Did you notice that `Failed — The Travis CI build failed (Details)`.
You should also update Makefile or project configuration for other platforms rather than only for Marmalade.

cocos2dx/include/CCProtocols.h
```diff
@@ -125,6 +125,9 @@ class CC_DLL CCRGBAProtocol
     virtual void updateDisplayedOpacity(GLubyte opacity) = 0;
 };
 
+void updateDisplayedColorChildren(CCNode *parent, const ccColor3B &parentColor);
+void updateDisplayedOpacityChildren(CCNode *parent, GLubyte parentOpacity);
```
dumganhar: Is it a good idea to add two global functions?

deemen: It is for behavior that is common between CCNodeRGBA and CCLayerRGBA. The
function needs to be implemented exactly the same in both, and would have
been a static function. I removed the code duplication by making a global.

If you prefer, I can make it a static function of RGBAProtocol, which would
remove the global.

On Monday, May 20, 2013, James Chen wrote:

> In cocos2dx/include/CCProtocols.h:
> 
> > @@ -125,6 +125,9 @@ class CC_DLL CCRGBAProtocol
> >      virtual void updateDisplayedOpacity(GLubyte opacity) = 0;
> >  };
> > 
> > +void updateDisplayedColorChildren(CCNode *parent, const ccColor3B &parentColor);
> > +void updateDisplayedOpacityChildren(CCNode *parent, GLubyte parentOpacity);
> 
> Is it a good idea to add two global functions?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/cocos2d/cocos2d-x/pull/2577/files#r4292670
> .

## 

Sent from Gmail Mobile

dumganhar: > If you prefer, I can make it a static function of RGBAProtocol, which would remove the global.

Yes, I think that will be better.

cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
```diff
@@ -577,15 +577,7 @@ void CCLayerRGBA::updateDisplayedOpacity(GLubyte parentOpacity)
...
-        CCObject *obj = NULL;
-        CCARRAY_FOREACH(m_pChildren, obj)
-        {
-            CCRGBAProtocol *item = dynamic_cast<CCRGBAProtocol*>(obj);
-            if (item)
-            {
-                item->updateDisplayedOpacity(_displayedOpacity);
-            }
-        }
+        updateDisplayedOpacityChildren(this, _displayedOpacity);
```
dumganhar: It's ok to invoke a static method without `CCRGBAProtocol::`, 
But I think adding `CCRGBAProtocol::` as prefix will make the codes clearer.

cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
```diff
@@ -597,15 +589,7 @@ void CCLayerRGBA::updateDisplayedColor(const ccColor3B& parentColor)
...
-        CCObject *obj = NULL;
-        CCARRAY_FOREACH(m_pChildren, obj)
-        {
-            CCRGBAProtocol *item = dynamic_cast<CCRGBAProtocol*>(obj);
-            if (item)
-            {
-                item->updateDisplayedColor(_displayedColor);
-            }
-        }
+        updateDisplayedColorChildren(this, _displayedColor);
```
dumganhar: The same issue as i mentioned.

deemen:
Fixed, thanks for the review!

extensions/CCBReader/CCNodeLoader.cpp
```diff
@@ -64,42 +63,6 @@ void CCNodeLoader::parseProperties(CCNode * pNode, CCNode * pParent, CCBReader *
...
-                    {
-                        bFound = true;
-                        break;
-                    }
-                }
-                setProp &= bFound;
-            }
-        }
-        else if (isExtraProp && pNode == pCCBReader->getAnimationManager()->getRootNode())
-        {
```
dumganhar, line 93: Why did you delete these lines?

deemen, line 93: I deleted them because 1) the code doesn't currently work and 2) it was difficult to refactor it to work with the new CCBNodeWrapper interface I created.

This code, the way I understand it, is meant to forward the custom properties from the CCBFile node to the node it contains. The second block (the else if) loads but the root node of the sub-ccb file with the extraPropsNames array and when the CCBFile is parsed later, the first block will execute.

If you look at like 74, the pNode parameter (which is one of the parameters of the function) is assigned the internal node and then the current property (propertyName) is searched for in the extraPropsNames array. If it is found, it is set below on the internal node (now pNode). This will only work for 1 parameter, because the next time the main loop runs, the pNode is no longer set to the CCBFile, so line 69 will never be true and all other extra properties won't be checked.

I thought it best to disable this functionality for now and fix it in a later commit.

cocos2dx/base_nodes/CCNode.cpp
```diff
@@ -1315,21 +1315,32 @@ void CCNodeRGBA::setOpacity(GLubyte opacity)
...
+        CCNode *node = static_cast<CCNode*>(obj);
+        CCRGBAProtocol *rgba = dynamic_cast<CCRGBAProtocol*>(node);
+        if(rgba)
+        {
+            rgba->updateDisplayedOpacity(parentOpacity);
+        }
+        else
+        {
+            CCRGBAProtocol::updateDisplayedOpacityChildren(node, parentOpacity);
+        }
```
dumganhar, line 1340: It's better to make the `static_cast` line to `else` like this:

```
    CCRGBAProtocol *rgba = dynamic_cast<CCRGBAProtocol*>(obj);
    if(rgba)
    {
        rgba->updateDisplayedOpacity(parentOpacity);
    }
    else
    {
          CCNode *node = static_cast<CCNode*>(obj);
        CCRGBAProtocol::updateDisplayedOpacityChildren(node, parentOpacity);
    }
```

deemen:
Can pull request be merged now?

PR #2578, opened 2013-05-17
Grid setProjection2D fix

ricardoquesada:
It doesn't multiply it by the CC_CONTENT_SCALE_FACTOR().
Not needed since it is CCDirector::getWinSizeInPixels() already has it
calculated


dumganhar:
Thanks.

PR #2579, opened 2013-05-18
Ccconfiguration extension

ricardoquesada:
1st part of big patch:
- The CCConfiguration is stored in a `CCDictionary`
- Initialization of the `CCConfiguration` and loading GL values is splitted. This allows users to obtain values before creating the GL context

Also fixes some issues with the `DataVisitor` pattern.

### DIFF
--- a/cocos2dx/CCConfiguration.cpp
+++ b/cocos2dx/CCConfiguration.cpp
@@ -101,43 +101,43 @@ void CCConfiguration::dumpInfo(void) const
 
 	CCPrettyPrinter visitor(0);
 	m_pDefaults->acceptVisitor(visitor);
-	cout << visitor.getResult();
 
+	CCLOG("%s", visitor.getResult().c_str());
 }
 
 void CCConfiguration::gatherGPUInfo()
 {
-	m_pDefaults->setObject( new CCString( (const char*)glGetString(GL_VENDOR)), "gl.vendor");
-	m_pDefaults->setObject( new CCString( (const char*)glGetString(GL_RENDERER)), "gl.renderer");
-	m_pDefaults->setObject( new CCString( (const char*)glGetString(GL_VERSION)), "gl.version");
+	m_pDefaults->setObject( CCString::create( (const char*)glGetString(GL_VENDOR)), "gl.vendor");
+	m_pDefaults->setObject( CCString::create( (const char*)glGetString(GL_RENDERER)), "gl.renderer");
+	m_pDefaults->setObject( CCString::create( (const char*)glGetString(GL_VERSION)), "gl.version");
 
     m_pGlExtensions = (char *)glGetString(GL_EXTENSIONS);
 
     glGetIntegerv(GL_MAX_TEXTURE_SIZE, &m_nMaxTextureSize);
-	m_pDefaults->setObject( new CCInteger((int)m_nMaxTextureSize), "gl.max_texture_size");
+	m_pDefaults->setObject( CCInteger::create((int)m_nMaxTextureSize), "gl.max_texture_size");
 
     glGetIntegerv(GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS, &m_nMaxTextureUnits);
-	m_pDefaults->setObject( new CCInteger((int)m_nMaxTextureUnits), "gl.max_texture_units");
+	m_pDefaults->setObject( CCInteger::create((int)m_nMaxTextureUnits), "gl.max_texture_units");
 
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
     glGetIntegerv(GL_MAX_SAMPLES_APPLE, &m_nMaxSamplesAllowed);
-	m_pDefaults->setObject( new CCInteger((int)m_nMaxSamplesAllowed), "gl.max_samples_allowed");
+	m_pDefaults->setObject( CCInteger::create((int)m_nMaxSamplesAllowed), "gl.max_samples_allowed");
 #endif
 
     m_bSupportsPVRTC = checkForGLExtension("GL_IMG_texture_compression_pvrtc");
-	m_pDefaults->setObject( new CCBool(m_bSupportsPVRTC), "gl.supports_PVRTC");
+	m_pDefaults->setObject( CCBool::create(m_bSupportsPVRTC), "gl.supports_PVRTC");
 
     m_bSupportsNPOT = true;
-	m_pDefaults->setObject( new CCBool(m_bSupportsNPOT), "gl.supports_NPOT");
+	m_pDefaults->setObject( CCBool::create(m_bSupportsNPOT), "gl.supports_NPOT");
 	
     m_bSupportsBGRA8888 = checkForGLExtension("GL_IMG_texture_format_BGRA888");
-	m_pDefaults->setObject( new CCBool(m_bSupportsBGRA8888), "gl.supports_BGRA8888");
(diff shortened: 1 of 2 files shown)


cocos2dx/CCConfiguration.cpp
```diff
@@ -128,4 +160,87 @@ bool CCConfiguration::checkForGLExtension(const string &searchName)
...
+	CCObject *ret = m_pDefaults->objectForKey(key);
+	if( ret ) {
+		if( CCDouble *obj=dynamic_cast<CCDouble*>(ret) )
+			return obj->getValue();
+
+		if( CCInteger *obj=dynamic_cast<CCInteger*>(ret) )
+			return obj->getValue();
+	}
+
+	return NULL;
```
angeltown: Building error on travis ci https://travis-ci.org/cocos2d/cocos2d-x/jobs/7268093

<pre>
../CCConfiguration.cpp: In member function ‘double cocos2d::CCConfiguration::getNumber(const char*) const’:

../CCConfiguration.cpp:238:9: error: converting to non-pointer type ‘double’ from NULL [-Werror=conversion-null]

cc1plus: all warnings being treated as errors
</pre>


We should return 0.0 instead.

cocos2dx/CCConfiguration.cpp
```diff
@@ -85,17 +81,65 @@ bool CCConfiguration::init(void)
...
+	m_pDefaults->acceptVisitor(visitor);
+	cout << visitor.getResult();
+
+}
+
+void CCConfiguration::gatherGPUInfo()
+{
+	m_pDefaults->setObject( new CCString( (const char*)glGetString(GL_VENDOR)), "gl.vendor");
+	m_pDefaults->setObject( new CCString( (const char*)glGetString(GL_RENDERER)), "gl.renderer");
+	m_pDefaults->setObject( new CCString( (const char*)glGetString(GL_VERSION)), "gl.version");
```
dumganhar: Hi, Riq,
We should use `CCString::create()`, `CCInteger::create()` to create a value for CCDictionary rather than just `new`, otherwise it'll cause memory leak issues. 
`create` methods are always return `autorelease` instances.

cocos2dx/CCConfiguration.cpp
```diff
@@ -85,17 +81,65 @@ bool CCConfiguration::init(void)
...
+//        CCObject *obj = pElement->getObject();
+//		const char *key = pElement->getStrKey();
+//
+//		printf("%s = %s\n", key, obj->getDescription()->getCString() );
+//
+//    }
+
+	CCPrettyPrinter visitor(0);
+	m_pDefaults->acceptVisitor(visitor);
+	cout << visitor.getResult();
```
dumganhar: `cout` can't work on Android. We need to use `CCLOG`. :)

ricardoquesada:
Ok. I will update the patch. thanks

cocos2dx/cocoa/CCDictionary.cpp
```diff
@@ -189,22 +190,24 @@ CCObject* CCDictionary::objectForKey(intptr_t key)
...
-        pStr = CCString::create("");
-    }
-    return pStr;
+	CCObject *value = objectForKey(key);
+	if( value ) {
+		CCPrettyPrinter visitor(0);
+		value->acceptVisitor(visitor);
+		return CCString::create( visitor.getResult() );
+	}
+	return CCString::create("");
```
dumganhar, line 210: I did a test about the performance of visitor pattern or `dynamic` as follows:

```
CCObject* obj = CCString::create("hello");

CC_PROFILER_START("dynamic");
for (int i = 0; i < 10000000; ++i) {
    dynamic_cast<CCString*>(obj);
}

CC_PROFILER_STOP("dynamic");


CC_PROFILER_START("visitor");

for (int i = 0; i < 10000000; ++i) {
    CCPrettyPrinter visitor(0);
    obj->acceptVisitor(visitor);
}

CC_PROFILER_STOP("visitor");

CC_PROFILER_DISPLAY_TIMERS();
```

The result on iPhoneSimulator is:

<pre>
Cocos2d: dynamic: avg time, 210.308472ms
Cocos2d: visitor: avg time, 1205.037476ms
</pre>


I'll do more tests on iOS devices. But it seems that visitor pattern is slower than dynamic_cast.

cocos2dx/CCDirector.cpp
```diff
@@ -322,6 +320,13 @@ void CCDirector::setOpenGLView(CCEGLView *pobOpenGLView)
 
         m_pobOpenGLView->setTouchDelegate(m_pTouchDispatcher);
         m_pTouchDispatcher->setDispatchEvents(true);
+
+		// XXX: This method should not be called explicity.
+		// XXX: instead cocos2d should emitt an "gl view set", and CCConfiguration should listen to that event
+		CCConfiguration::sharedConfiguration()->gatherGPUInfo();
+
+		// dump config data once OpenGL was initialized.
+		CCConfiguration::sharedConfiguration()->dumpInfo();
```
dumganhar, line 329: These lines should be invoked right after `if (m_pobOpenGLView != pobOpenGLView) {`, otherwise, `createStatsLabel` will create FPS LabelTTF, it will use `CCConfiguration::getMaxTextureSize` while the `CCConfiguration` hasn't been initialized and the return value of `getMaxTextureSize` is 0.

cocos2dx/cocoa/CCDictionary.cpp
dumganhar, line 210: The same test on iOS device( iPod touch 4):

<pre>
Cocos2d: dynamic: avg time, 1054.384033ms
Cocos2d: visitor: avg time, 10478.743042ms
</pre>


So IMO, CCPrettryPrinter is only for outputting logs, we should not use it in other places. What's your opinion?

ricardoquesada, line 210: Thanks for doing the perf test. Interesting results.

What I tried to do was to fix the code. `CCDictionary # valueForKey` only worked if the value was string. If it weren't a string, it returned `NULL`, and that is not correct. So the way I see it is that is was broken, so I fixed it. That code is not really necessary for the work that I am doing, so if you want I can revert it back. 

Regarding the performance, the new code is 10x slower... which is very bad, but if you put it in context, it `valueForKey` is not used a lot in the code.

Your call. Let me know what should I do.

cocos2dx/CCDirector.cpp
ricardoquesada, line 329: Ok. good to know. thanks.

cocos2dx/cocoa/CCDictionary.cpp
dumganhar, line 210: > That code is not really necessary for the work that I am doing, so if you want I can revert it back.

Yes, I think we should revert it back first.

> Regarding the performance, the new code is 10x slower... which is very bad, but if you put it in context, it valueForKey is not used a lot in the code.

Is it a good idea to add a virtual method to CCObject named `toString` or `toValue`, it returns empty string by default.
CCBool, CCInteger, CCString, CCDouble, etc overload that method and return string of their value.

ricardoquesada, line 210: Ok. I will revert it back.

Adding `toString` in `CCObject` is good... in fact in my original patch I had something like that, but I removed it because the visitor pattern was added. So, I think we should use either `toString` or the visitor pattern, but not both.

ricardoquesada:
closing this PR. I created a new one:
https://github.com/cocos2d/cocos2d-x/pull/2615

PR #2580, opened 2013-05-18
Add cookie support to HttpRequest and add support for callbacks on CC.ControlButton

randyv2zhurosoft:
I'm not very familiar with git, so I apologize for lumping two changes into one pull request.

Two things are included in this request:
1. The changes to HttpClient.cpp adds support for cookies. They're stored in a hardcoded location, but I don't think this is an issue since cookies should "just work", and this change makes that happen. As an example, you can see how cookies are handled in iOS' code, where the user of the SDK never needs to specify or know where the cookies are stored; it's enough to simply know that they are being stored. Clients of HTTP libraries expect cookies to be received and stored by default, and this functionality is key in maintaining sessions in many apps such as games that use HTTP as a transport for server synchronization.
2. A manual binding is added for CC.ControlButton so a client can add their own callbacks to CC.ControlButton. Additionally, the event enum is imported into JS as well. I don't know if there's a better way to do this than maintaining it by hand, but looking at the existing code base it appears that this is the way it's been done so far.


randyv2zhurosoft:
Apparently I'm really not good with git, as another change snuck in here unexpectedly.

A third change:
1. Change CCControlButton so it zooms based on the scale that a user has previously set, instead of always scaling to 1.1 (when zoomed) and 1.0 (when not-zoomed).

extensions/network/HttpClient.cpp
```diff
@@ -58,6 +58,8 @@
 
 typedef size_t (*write_callback)(void *ptr, size_t size, size_t nmemb, void *stream);
 
+static const char *cookieFilename = (CCFileUtils::sharedFileUtils()->getWritablePath() + "cookieFile.txt").c_str();
```
dumganhar: It's better that developer could set the path of cookieFile.

dumganhar: Also, we should add a method to disable or enable cookie. It's disabled by default.

randyv2zhurosoft: Fair enough. I'll make a change to allow developers to pass in a filename, or null if they want to use the one I've defined here.

dumganhar: OK, before you make that change, please pull the latest codes firstly, there probably are some conflicts. Thanks.

randyv2zhurosoft: I merged the latest code and re-applied my change. I'm not sure it falls into your best practices or standard style, but let me know if it's not quite what works for you guys and I'll adjust it.

dumganhar:
Thank you. I'll merge your code to develop branch.

tiantian20007:
I think the Callbacks on CC.ControlButton part should also be merged;
CCControlButton is widely used. 
I wonder why this feature is missing for so long? 
http://cocos2d-x.org/issues/1690

PR #2581, opened 2013-05-19
Update js_bindings_ccbreader.cpp

zj831007:
method name is error


dumganhar:
It's better to make it compatible with old method. Therefore, you probably need to add

<pre>
cc.AnimationManager.setCompletedAnimationCallback = cc.AnimationManager.setAnimationCompletedCallback;
</pre>

to `scripting/javascript/js/jsb_cocos2d_extension.js`

PR #2583, opened 2013-05-19
Bug in previous modifications made to GL_UNPACK_ALIGNMENT 

silverscania:
I previously fixed a bug where the row byte alignment was set incorrectly for other pixel formats.

This introduced a new bug as discussed here:https://github.com/cocos2d/cocos2d-x/commit/e66e7cfcca2676d8e1b3fa44b27e096729c68f5a#commitcomment-3243089

This was because setting the row byte alignment to anything other than 1 for non power of two textures, even if the alignment was 2, 4 or 8 was incorrect.


silverscania:
Sorry this is wrong. Closing this.
The real reason is this line:

```
case kCCTexture2DPixelFormat_RGB888:
            // It is 32 and not 24, since its internal representation uses 32 bits.
            ret = 32;
            break;
```

bitsPerPixelForFormat function returns wrong number of bits for RGB_888 textures.
Will fix another way.

PR #2584, opened 2013-05-19
pointQueryFirst returns invalid Shape when a shape is not found.

smshuja:
It will now return NULL if a shape is not found.


dumganhar:
Thanks.

PR #2587, opened 2013-05-19
It's better 'retain' returns 'CCObject *' than void

SAM-tak:
In Objective-C, NSObject's 'retain' returns self.
So I think it's better 'retain' returns 'CCObject *' like 'autorelease'.

In Lua:
local x = CCArray:create():retain()
It's more convenient x is set by CCArray instead of nil.

And added armv7s to valid architecture in xcode project file of CocosDenshion

### DIFF
--- a/samples/Lua/HelloLua/Resources/hello.lua
+++ b/samples/Lua/HelloLua/Resources/hello.lua
@@ -1,14 +1,6 @@
-print("hello lua!")
-
 -- cclog
-if _ccLog then
-	cclog = function(...)
-		_ccLog(string.format(...))
-	end
-else
-	cclog = function(...)
-		print(string.format(...))
-	end
+cclog = function(...)
+    print(string.format(...))
 end
 
 -- for CCLuaEngine traceback


Weeds:
You should probably not write code like above.
CCArray:create() (or other create methods) may return NULL if the initialization fails. I'm not sure how lua handles that case but I guess this would also result in a LUA error at best or an application crash when accessing a NULL pointer.
You can write this in Objective-C as nothing happens if you send a message to a nil object. This is not the cse in C/C++ (and possibly LUA)

dumganhar:
@Weeds 

> CCArray:create() (or other create methods) may return NULL if the initialization fails.

I agree.

SAM-tak:
> CCArray:create() (or other create methods) may return NULL if the initialization fails. I'm not sure how lua handles that case but I guess this would also result in a LUA error at best or an application crash when accessing a NULL pointer.

I think this is desirable behavior (to me at least).
If you doubt create() may returns null probably, you can check create()'s return value if 'retain' returns CCObject_.
'void retain' is no helpful for this situation. 'retain' returns CCObject_ or not, you should check create()'s return value if you don't hope app crash.
I think it's convenient **possible** to write like below:

``` Lua
Foo.spriteframes = {
    CCArray:create():retain(),
    CCArray:create():retain(),
    CCArray:create():retain()
}
```

instead of **must** to write like below:

``` Lua
Foo.spriteframes = {
    CCArray:create(),
    CCArray:create(),
    CCArray:create(),
}
Foo.spriteframes[1]:retain()
Foo.spriteframes[2]:retain()
Foo.spriteframes[3]:retain()
```

SAM-tak:
In early days of cocos2d-x, most people translate objective-c cocos2d source to cocos2d-x, and many people crash apps by method call for null, but now, not a few people write app with cocos2d-x at beginning, and objective-c style factory method like 'arrayWithXXX' are deprecated.
I think it's no longer effective people can notice 'possibility of method call for null' at compile time by retain method be defined as void function.

Weeds:
Another issue is, that the code won't work with C/C++
You still won't be able to write:

Foo\* object = Foo::create()->retain();

As you would need to cast the value returned by retain() all the time.

@dumganhar 
Just noticed a small issue in CCArray implementation: Should the CCArray constructor call init()? Because in that case init() may be called multiple times when using the create() methods.

SAM-tak:
> As you would need to cast the value returned by retain() all the time.

That's right. but I focus to Lua. This change dosen't present any improvement to C++ writer.
but it improve writability in Lua and no break semantics in C++.

PR #2590, opened 2013-05-20
New PivotJoint will now accept 3 or 4 parameters like in html5 version

smshuja:
This will make the API consistent with the html5 version


dumganhar:
Thank you.

PR #2607, opened 2013-05-21
Implementing foreground and background status for Tizen port.

pyrasis:
Implementing foreground status and background status.

And modify SimpleGame's AppDelegate code comment.

### DIFF
--- a/cocos2dx/platform/tizen/CCOspApplication.cpp
+++ b/cocos2dx/platform/tizen/CCOspApplication.cpp
@@ -158,9 +158,7 @@ CCOspApplication::OnBackground(void)
         timer->Cancel();
     }
 
-	CCApplication::sharedApplication()->applicationDidEnterBackground();
-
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/tizen/CCOspApplication.cpp
```diff
@@ -152,6 +157,10 @@ of this software and associated documentation files (the "Software"), to deal
     {
         timer->Cancel();
     }
+
+	CCApplication::sharedApplication()->applicationDidEnterBackground();
+
+	CCNotificationCenter::sharedNotificationCenter()->postNotification(EVENT_COME_TO_BACKGROUND, NULL);
```
dumganhar: Does tizen also lose texture when entering background like android?
If not, why to call `postNotification(EVENT_COME_TO_BACKGROUND, NULL);`

pyrasis:
Tizen doesn't need the notification.

dumganhar:
Thanks.

PR #2608, opened 2013-05-21
CCLabelTTF::enableShadow(…) const update

edbartley:
Make shadowOffset parameter of CCLabelTTF::enableShadow(…) a const reference so that code like "enableShadow(CCSize(0,1), 0.5f, 2);" may be written.


yaowei1188:
the enableShadow() how to specify the color of shadow?
i try to invoke the method as following:
enableShadow(CCSize(0,1), 0.5f, 2)
it seems it doesn't work!
any body can help me?
thanks in advance!

edbartley:
Shadow color appears to be hard code but that shouldn't be too hard to
change.

PR #2612, opened 2013-05-22
Add SSL certificate verify option setter to HttpClient

hokuts:
Add CURLOPT_SSL_VERIFYPEER and CURLOPT_SSL_VERIFYHOST option setter to HttpClient, so that developer can specify to allow self-signed certificate on SSL testing environment.
This is second request of same issue, first is below.
https://github.com/cocos2d/cocos2d-x/pull/2575

### DIFF
--- a/extensions/network/HttpClient.h
+++ b/extensions/network/HttpClient.h
@@ -93,7 +93,11 @@ public:
     
     /**
      * Change SSL verify peer flag
-     * @param sslVerifyPeer
+     * @param sslVerifyPeer Curl option value CURLOPT_SSL_VERIFYPEER.
+     * - false: Don't verify the authenticity of the peer's certificate.
+     * - true : Verify the authenticity of the peer's certificate. (Default)
+     * @sa CCHttpClient::isSSLVerifyPeer()
+     * @sa http://curl.haxx.se/libcurl/c/curl_easy_setopt.html#CURLOPTSSLVERIFYPEER
      * @return NULL
      */
     inline void setSSLVerifyPeer(bool sslVerifyPeer){_sslVerifyPeer = sslVerifyPeer;}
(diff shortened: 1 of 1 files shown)


extensions/network/HttpClient.h
```diff
@@ -89,7 +89,37 @@ class CCHttpClient : public CCObject
...
     inline int getTimeoutForRead() {return _timeoutForRead;};
-        
+    
+    
+    /**
+     * Change SSL verify peer flag
+     * @param sslVerifyPeer
+     * @return NULL
+     */
+    inline void setSSLVerifyPeer(bool sslVerifyPeer){_sslVerifyPeer = sslVerifyPeer;}
```
sergey-shambir, line 103: Please move implementation to HttpClient.cpp, don't use inlines.

extensions/network/HttpClient.h
```diff
@@ -89,7 +89,37 @@ class CCHttpClient : public CCObject
...
+     */
+    inline bool isSSLVerifyPeer(){return _sslVerifyPeer;}
+
+    
+    /**
+     * Change SSL verify host flag
+     * @param sslVerifyHost
+     * @return NULL
+     */
+    inline void setSSLVerifyHost(long sslVerifyHost){_sslVerifyHost = sslVerifyHost;}
```
sergey-shambir, line 127: Add more documentation here, please. Cocos2d-x targeted not only to experienced teams, but also to individual developers with various skills.

See PHP curl docs for used curl options as an example.

extensions/network/HttpClient.h
hokuts, line 103: I think inline implementations is suitable in this case, because other setter/getter methods in this class are defined inline, too.

extensions/network/HttpClient.h
hokuts, line 127: Added a commit. Check it, please.

PR #2618, opened 2013-05-23
All android projects support multiple processors now. Usage: ./build_native.sh -p

dumganhar:


dumganhar:
Close it since it'll print lots of logs in terminal when a build is finished. Travis ci only support log with max 10000 lines.

PR #2623, opened 2013-05-23
Fixed setHeader while Android had problems with stringstreams.

casarock:
HI,

Android did not work reliable with string stream for header concatenation. We've changed it to be sure XMLHttpRequest works cross plattform.


dumganhar:
Thanks.

PR #2624, opened 2013-05-23
Fixed linebreaks on Mac OSX

benjamesward:
Previously the CCLabelTTF linebreaks worked on a character by character basis, so you'd have a newline inserted in the middle of a word. Now it breaks as you'd expect (on spaces and special characters).

Before: This is a long sen\ntence.
After: This is a long \nsentence.

### DIFF
--- a/cocos2dx/platform/mac/CCImage.mm
+++ b/cocos2dx/platform/mac/CCImage.mm
@@ -396,12 +396,12 @@ static bool _initWithString(const char * pText, cocos2d::CCImage::ETextAlign eAl
 				NSUInteger length = [string length];
 				NSRange range = NSMakeRange(0, 1);
 				NSUInteger width = 0;
-                NSUInteger lastBreakLocation = 0;
+				NSUInteger lastBreakLocation = 0;
 				for (NSUInteger i = 0; i < length; i++) {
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/mac/CCImage.mm
```diff
@@ -397,12 +396,15 @@ static bool _initWithString(const char * pText, cocos2d::CCImage::ETextAlign eAl
 				NSUInteger length = [string length];
 				NSRange range = NSMakeRange(0, 1);
 				NSUInteger width = 0;
+                NSUInteger lastBreakLocation = 0;
 				for (NSUInteger i = 0; i < length; i++) {
 					range.location = i;
-					[lineBreak appendString:[string substringWithRange:range]];
+                    NSString *character = [string substringWithRange:range];
+					[lineBreak appendString:character];
+                    if ([@"!?.,-= " rangeOfString:character].location != NSNotFound) { lastBreakLocation = i; }
```
minggo: Could you please fix the indent?

benjamesward: Sorry about that. I've fixed and committed, although I'm still new to Github. Did I do it right? :-)

PR #2641, opened 2013-05-24
Linux port: added libX11 to linked libraries.

sergey-shambir:
Otherwise build fails on linux.


dumganhar:
Thanks. @sergey-shambir , which linux distribution are you using?

sergey-shambir:
Ubuntu Raring 13.04 64-bit.

PR #2642, opened 2013-05-24
Implemented CCDictionary::writeToFile with CCDataVisitor.

sergey-shambir:
1. More strict code with wider types set.
2. Visitor is faster than dynamic_cast.
3. Added fully automatic tests.

### DIFF
--- a/cocos2dx/platform/CCFileUtils.cpp
+++ b/cocos2dx/platform/CCFileUtils.cpp
@@ -26,6 +26,10 @@ THE SOFTWARE.
 #include "CCDirector.h"
 #include "cocoa/CCDictionary.h"
 #include "cocoa/CCString.h"
+#include "cocoa/CCBool.h"
+#include "cocoa/CCInteger.h"
+#include "cocoa/CCDouble.h"
+#include "cocoa/CCFloat.h"
 #include "CCSAXParser.h"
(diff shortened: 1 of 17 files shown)


dumganhar:
About 2) Visitor is faster than dynamic_cast.?
Did you make a performance test?

dumganhar:
BTW, ci build fails. Could you please look at it?

sergey-shambir:
I've not updated XCode projects for Mac and iOS (and going to do it tomorrow). Everything else was updated (makefiles, vcproj).

sergey-shambir:
I've not updated some makefiles and used exceptions. I will fix it soon.

sergey-shambir:
Please check if it builds correctly on Mac/iOS, since i've added files and folder manually.

`CCData` support and `CCArray::writeToFile` are missed, I think there are no need to fix it in this pull request.

sergey-shambir:
About dynamic cast: virtual method call faster ~10 times on Visual C++ and ~3 times on g++/clang. Also visitor needs only 2 virtual method calls (`accept` & `visit`), while any other approach requires chain of `if` checks, each check needs `dynamic_cast` or substring lookup in `typeid(*obj).name()`.

http://tinodidriksen.com/2010/04/14/cpp-dynamic-cast-performance/

PR #2658, opened 2013-05-28
[Android] use onWindowFocusChanged(bool) instead of onResume()/onPause()

michaelcontento:
we should use onWindowFocusChanged(bool) to detect if the app is currently
visitible to the user or not. because onPause()/onResume() leads to false
positives on some android versions (e.g. Kindle Fire). if you lock those
devices, while the app running, everything is fine (onPause() is handled
properly). but as soon as you wake up the device, and the lock screen gets
visible, the app is resumed and the music is played (onResume() is
called to early).

this is a) not the best behaviour and b) such games will be rejected on the
samsung app store. onWindowFocusChanged(bool) is, like onPause()/onResume(),
available since API level 1 and it should be safe to simply switch.

### DIFF
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
@@ -77,19 +77,15 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
 	// ===========================================================
 
 	@Override
-	protected void onResume() {
-		super.onResume();
-
-		Cocos2dxHelper.onResume();
-		this.mGLSurfaceView.onResume();
-	}
-
-	@Override
-	protected void onPause() {
-		super.onPause();
-
-		Cocos2dxHelper.onPause();
-		this.mGLSurfaceView.onPause();
+	public void onWindowFocusChanged(final boolean hasWindowFocus) {
+		super.onWindowFocusChanged(hasWindowFocus);
+		if (hasWindowFocus) {
+			Cocos2dxHelper.onResume();
+			this.mGLSurfaceView.onResume();
+		} else {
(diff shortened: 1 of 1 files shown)


rtissera:
Looks good to me, this is a very common issue on Android.
The PR should be safe but I would suggest a lot of testing, this is quite a big change for Android platform implementation.
Have you done extensive testing on 2.3.x, 3.x and ICS/JB devices ?

michaelcontento:
Hi,

I've tested this change on a lot of devices on my own[1] and the game is online in various Android based App Stores[2] without even one bug report regarding problems caused by this change.

So it seems, at least to me, safe to merge :)

Best regards,
Michael

[1] from Android 2.3 to 4.2; tablets, smartphones and even "proprietary Android" on the Kindle Fire first gen
[2] [Amazon App Store](http://www.amazon.com/gp/product/B00CUOSANI?t=coragames-21),  [Google Play Store](https://play.google.com/store/apps/details?id=com.coragames.dtdng) and [Samsung Apps](http://apps.samsung.com/earth/topApps/topAppsDetail.as?productId=000000609905)

attilaz:
Here is what google devs think is the best solution:
http://android-developers.blogspot.hu/2011/11/making-android-games-that-play-nice.html

We are using the second solution, but it failed for us on 4.0.4 devices (Google Nexus and Sony Xperia). On these devices when we press the pause/power button we get:  'onPause' then 'onResume' and ;onPause' again (no onWindowFocusChanged) then after pressing the home button, we get onResume.
After unlocking the screen we get OnWindowFocusChanged(true), OnWindowFocusChanged(false), OnWindowFocusChanged(true),

Hope it helps.

rtissera:
Thanks for the link, very instructive.
Then we need to elaborate a better strategy based on Google recommendations.

2013/7/17 attilaz notifications@github.com

PR #2663, opened 2013-05-28
Fix Tizen CCEditBox bug.

pyrasis:
- Encoding UTF-8 and check empty string.
- Keep a previous text.

### DIFF
--- a/cocos2dx/platform/tizen/CCOspForm.cpp
+++ b/cocos2dx/platform/tizen/CCOspForm.cpp
@@ -61,7 +61,8 @@ CCOspForm::OnTerminating(void)
 {
     result r = E_SUCCESS;
 
-    __pKeypad->Destroy();
+    if (__pKeypad)
+        __pKeypad->Destroy();
 
     return r;
 }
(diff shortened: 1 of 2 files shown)


cocos2dx/platform/tizen/CCOspForm.cpp
```diff
@@ -116,8 +116,11 @@ of this software and associated documentation files (the "Software"), to deal
 void CCOspForm::OnTextValueChanged(const Tizen::Ui::Control& source)
 {
     String text = __pKeypad->GetText();
-    AsciiEncoding ascii;
-    m_pfEditTextCallback((const char *)ascii.GetBytesN(text)->GetPointer(), m_pCtx);
+    Utf8Encoding utf8;
+    const char* pText = "";
+    if (!text.IsEmpty())
+        pText = (const char *)utf8.GetBytesN(text)->GetPointer();
+    m_pfEditTextCallback(pText, m_pCtx);
```
dumganhar, line 126: Does `pText` need be `delete`d after `m_pfEditTextCallback(pText, m_pCtx);` since there is a `N` in GetByteN which means that `new`.

pyrasis:
OK. delete ByteBuffer memory.

dumganhar:
Thanks. :)

PR #2666, opened 2013-05-29
Ability to save/retrieve CCData into/from CCUserDefault

jotel:
I needed to save some binary data into CCUserDefault object... so I created this patch.
It basically moves CCData.\* files from extensions/CCBReader into cocos2dx/cocoa and add some extra methods to CCUserDefault class.

The binary data is stored as base64 encoded string (added encoding function into cocos2dx/support/base64.cpp

I think that other developers might find this useful and you add it to the project... 
I hope that I didn't miss any build script/project... and platforms i'm not familiar will still build  without errors.

Regards
Jaro

### DIFF
--- a/cocos2dx/support/user_default/CCUserDefault.mm
+++ b/cocos2dx/support/user_default/CCUserDefault.mm
@@ -424,7 +424,7 @@ CCData* CCUserDefault::getDataForKey(const char* pKey, CCData* defaultValue)
     }
     else
     {
-        unsigned char *bytes = {};
+        unsigned char *bytes = {0};
         unsigned long size = 0;
         
         if (data.length > 0) {


dumganhar:
Did you see travis-ci build fails ? https://travis-ci.org/cocos2d/cocos2d-x/builds/7583580

jotel:
No :( I'll take a look tonight.

Thanks

cocos2dx/support/user_default/CCUserDefault.mm
```diff
@@ -374,6 +375,70 @@ static void deleteNodeByKey(const char *pKey)
...
+#endif
+    
+    NSData *data = [[NSUserDefaults standardUserDefaults] dataForKey:[NSString stringWithUTF8String:pKey]];
+    if (! data)
+    {
+        return defaultValue;
+    }
+    else
+    {
+        unsigned char *bytes = {};
```
minggo: This line will cause compiling error on iOS.

jotel: It compiles and works for me on iOS ... Can you tell please what the compilation issue you experience ?

Thanks
Jaro

jotel:
Is fixed now... all green;)

Thanks
Jaro

cocos2dx/support/user_default/CCUserDefault.mm
jotel: You're right... I missed that as I'm using a different C++ Language Dialect
in my project and no compilation errors were reported.

Should be fine now.

Regards
Jaro

Jaroslaw Lewandowski

jotel@jotel.net
http://www.linkedin.com/in/jotel

minggo:
Thank you.

dumganhar:
@minggo 
Does this PR also need to be merged to master branch?
I saw it didn't in master branch yet.

minggo:
Master is used for releasing v2.1 final version. After releasing, it can not be modified directly.
This feature is not needed to be merged into master.

PR #2667, opened 2013-05-29
CCConfiguration has format

ricardoquesada:
CCconfiguration has a file format.
It has a data and metadata


minggo:
merged into develop too

PR #2648, opened 2013-05-27
adding norwegian language

frranck:

PR #2669, opened 2013-05-29
adding norwegian language

frranck:


minggo:
Duplicated to #2648?

frranck:
There was too much commits in 2648, so I created with one

minggo:
So please close #2648.

PR #2671, opened 2013-05-29
Implement setIMEKeyboardState function for Tizen

pyrasis:
- Implement setIMEKeyboardState
- Modify space character check.

### DIFF
--- a/cocos2dx/platform/tizen/CCImage.cpp
+++ b/cocos2dx/platform/tizen/CCImage.cpp
@@ -159,7 +159,7 @@ public:
 				return false;
 			}
 
-			if (isspace(unicode)) {
+			if (unicode == ' ') {
 				currentPaintPosition += face->glyph->metrics.horiAdvance >> 6;
 				prevGlyphIndex = glyphIndex;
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/tizen/CCImage.cpp
```diff
@@ -159,7 +159,7 @@ class BitmapDC
 				return false;
 			}
 
-			if (isspace(unicode)) {
+			if (unicode == ' ') {
```
dumganhar, line 162: Why to do this change?

pyrasis:
"안녕하세요" is hello in Korean.

isspace function recognizes some unicode Korean characters as a space character.

안 -> space character
녕 -> ok
하 -> ok
세 -> space character
요 -> space character

isspace(unicode)
isspace

unicode == ' '
equalspace

Do you have any idea?

dumganhar:
No ideas about that.

dumganhar:
Why not use `Landscape` mode for HelloCpp ?

pyrasis:
No reason. I will modify to set landscape mode.

PR #2676, opened 2013-05-30
Memcpy change submit to develop branch

liuyq:
Update the usage of call of CCTextureAtlas.updateQuad in CCLabelAtlas.updateAtlasValues and in CCTileMapAtlas::updateAtlasValueAt.

Update the reference directly instead of calling the updateQuad method.

PR #2678, opened 2013-05-30
Memcpy in CCSprite

liuyq:
CCSprite.cpp Update the usage of calling of CTextureAtlas.updateQuad

change to use the reference directly instead of calling the method.
and change to assign each value in the calling place instead of using
the dirty tag according to the review comment


liuyq:
this one is duplicated to https://github.com/cocos2d/cocos2d-x/pull/2676, so close it here.

PR #2681, opened 2013-05-30
Submit the change in CCSprite.cpp for memcpy to master branch

liuyq:
Submit the change in CCSprite.cpp for memcpy to master branch

Change to use reference to update the value indirectly instead of calling  CCTextureAtlas::updateQuad method


liuyq:
Sorry, seems I did a wrong operation, close this and will resubmit a new pull request.

PR #2682, opened 2013-05-30
CSprite.cpp Update the usage of calling of CTextureAtlas.updateQuad

liuyq:
change to use the reference directly instead of calling the method.
and change to assign each value in the calling place instead of using
the dirty tag according to the review comment

Signed-off-by: Yongqin Liu  liuyq0307@thundersoft.com


cocos2dx/sprite_nodes/CCSprite.cpp
```diff
@@ -511,10 +511,40 @@ void CCSprite::updateTransform(void)
...
+            CCAssert(m_uAtlasIndex >= 0 && m_uAtlasIndex < capacity, "updateTransform: Invalid m_uAtlasIndex");
+
+            unsigned int totalQuads = m_pobTextureAtlas->getTotalQuads();
+            ccV3F_C4B_T2F_Quad* quad = &((m_pobTextureAtlas->getQuads())[m_uAtlasIndex]);
+            ccVertex3F common = m_sQuad.bl.vertices;
+
+            quad->br.vertices = common;
+            quad->tl.vertices = common;
+            quad->tr.vertices = common;
+            quad->bl.vertices = common;
```
minggo: vertices may not the same, can not be assigned to the same value `common`.

PR #2683, opened 2013-05-30
Add CCRect::unionWithRect

BodbDearg:
Add a new CCRect utility function called 'unionWithRect' which computes the minimum rectangle that contains the current rectangle and another given rectangle, similar to to the following .NET function: http://msdn.microsoft.com/en-us/library/ms521839.aspx


BodbDearg:
Whoops... Wrong branch to merge into. Closing.

PR #2686, opened 2013-05-31
Spidermonkey for Android/x86

folecr:
- Allows building and running Cocos2d-X JS apps on Android x86 
- Particularly useful for development on the Android x86 emulator


dumganhar:
Thanks.

PR #2688, opened 2013-05-31
New action: CCCallFunction

ricardoquesada:
receives an std::function<void()>. This action deprecates the previous
ones: CCCallFunc, CCCallFuncO and CCCallFuncND

Enables C++ 11 compilation options

### DIFF
--- a/cocos2dx/actions/CCActionInstant.cpp
+++ b/cocos2dx/actions/CCActionInstant.cpp
@@ -375,11 +375,12 @@ void CCPlace::update(float time) {
 }
 
 //
-// CCCallFunction
+// CallFunc
 //
-CCCallFunction * CCCallFunction::create(std::function<void()> func)
+
+CCCallFunc * CCCallFunc::create(const std::function<void()> &func)
 {
-    CCCallFunction *pRet = new CCCallFunction();
+    CCCallFunc *pRet = new CCCallFunc();
 
     if (pRet && pRet->initWithFunction(func) ) {
         pRet->autorelease();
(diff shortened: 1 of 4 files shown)


minggo:
It is great. I think we can refactor CCCallFuncXXX using this way.

ricardoquesada:
@minggo Yes, I think `CCCallFunction` could replace the old `CCCallFunc`, `CCCallFuncN`, `CCCallFuncND` and `CCCallFuncO`.

Travis failed because C++ 11 is not being used (or enabled) for Android, Linux and Nacl. 
If possible, could you update the build scripts ? thanks.

scripting/javascript/bindings/generated
```diff
@@ -1 +1 @@
-Subproject commit d3baf739ac7c0a35904c7d8adc8d018b562d593f
+Subproject commit f729f121e9952c17b3a7fcbadbd5715b227a45e8
```
minggo: Why modify submodule pointer?

minggo:
Ok, i will send you a pull request.

scripting/javascript/bindings/generated
ricardoquesada: mmm... it shouldn't do that... bug.

minggo: Yep, may be you forgot to update submodule.

sergey-shambir:
Please pass std::function<void()> func by const reference.

ricardoquesada:
@sergey-shambir 
Thanks for the tip.
Regarding the lambda, will it go out of scope if I pass it as a reference ?

ricardoquesada:
@sergey-shambir 
Passing functions as `const &`. Internally I'm copying them to a local variable.

@minggo 
I updated the patch. I would appreciate it If you could merge it, and then fix the compilation issues with Android. Thanks.

sergey-shambir:
How about further refactoring for CCCallFunc and it's subclasses? You can remove union and `m_pSelectorTarget` at all, and use `std::bind` inside old-style `create()` function to remove extra class members. So `CCCallFunc` will keep only one member of type `std::function<void()>`, and subclasses will reuse it using std::bind.

sergey-shambir:
Oh, I've forgot about reference counting. So `std::bind` should be used only with smart pointer, and such refactoring can be out scope for this change.

ricardoquesada:
@sergey-shambir 
yes, good idea. I will remove the `m_pSelectorTarget` ivar

sergey-shambir:
I've added pull request #2697  to enable c++11 on platforms with makefiles, you can merge it when it will be accepted to pass build. C++11 on Visual Studio is always enabled (MSVC just have no way to disable it).

minggo:
@ricardoquesada 
I have sent you a pull request to enable c++11.

minggo:
The pull request is [here](https://github.com/ricardoquesada/cocos2d-x/pull/1).

minggo:
@ricardoquesada 
As you didn't merge my pull request, i sent #2709 to enable c++11.

And it seems that there is a conflict to merge this pull request. Because you modified some project settings as i did. So could you please synchronize the code to resolve it and it will trig a new travis compiling.

minggo:
I pulled your codes and fix conflict in #2720.
And will fix compiling errors in that PR.

ricardoquesada:
Added a new PR: https://github.com/cocos2d/cocos2d-x/pull/2727

@minggo Thanks for sending me a PR.
It is much easier for me if you do this:
- merge my PR in your local repo
- apply necessary patches to your local repo
- and then push it to cocos2d-x.

The PR you sent me had some issues (basically because I was "rebasing" intead of "merging" to prevent adding unnecessary "merge" logs). So today I created a new PR, but both you and I lost "one" day because we have a huge timezone difference.
Thanks.

PR #2689, opened 2013-05-31
fix a small warning project warning setting

davidxifeng:


dumganhar:
fixed at https://github.com/cocos2d/cocos2d-x/pull/2753. thanks.

PR #2693, opened 2013-05-31
Update XMLHTTPRequest.cpp

zj831007:
add UTF-8 or UTF-16 character handler of send param


dumganhar:
-        data = JS_EncodeString(cx, str);
-        JSStringWrapper strWrap(str);
-        data = (char*)strWrap.get().c_str(); 

`data` must be a `std::string`, and otherwise `data` will be destroyed right after strWraper was destroyed.

dumganhar:
Fixed at (master branch)#2710 and (develop branch)#2711. Thanks for your feedback.

PR #2694, opened 2013-05-31
Support for play sound effect with pitch , pan, gain

veena-peruvaje:
Port code from object c to  play sound effect with pitch, pan, gain

### DIFF
--- a/CocosDenshion/include/SimpleAudioEngine.h
+++ b/CocosDenshion/include/SimpleAudioEngine.h
@@ -163,6 +163,15 @@ public:
     }
 
     /**
+    @brief Play sound effect  with a file path, pitch, pan and gain
+    @param pszFilePath The path of the effect file,or the FileName of T_SoundResInfo
+    @bLoop Whether to loop the effect playing, default value is false
+    @pitch Pitch of the sound effect
+    @pan   Stereo effect of the sound
+    @gain  gain of the sound effect
+    */
+    unsigned int playEffect(const char* pszFilePath, bool bLoop, float pitch, float pan, float gain);
+    /**
     @brief Pause playing sound effect
     @param nSoundId The return value of function playEffect
     */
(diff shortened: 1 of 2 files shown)


dumganhar:
Hi, you added a new function(playEffect) for iOS, but other platform will be broken.
See here: https://travis-ci.org/cocos2d/cocos2d-x/builds/7662164.

sergey-shambir:
We faced with same problem and just used this topic to tweak local cocos2d version: http://www.cocos2d-x.org/boards/6/topics/5572?r=10217

Pitch/pan/gain have known implementations for iOS/OSX/android, but not for other platforms... So can we create separate branch which will be merged to `develop` when pitch/pan/gain will be done for all platforms? I can try to implement it for Linux (afaik it have 2 CocosDension backends: FMOD and OpenAL).

minggo:
@sergey-shambir 
I think you can implement it for other platforms and update them in this pull request.

sergey-shambir:
Ok, I'm working on implementations for other platforms. What's done:
- correct iOS/Mac backend
- Linux (FMOD/OpenAL) - it's also tested.
- Nacl, Blackberry (both OpenAL) backends
- Better doxygen and extended CocosDenshion test in TestCpp application.

Emscipten and tizen ports partially done, win32 and marmalade not started yet.

Now I encountered following problems:
- Emscripten port uses SDL, which have no support for pitch. There are workaround implemented Doom port for Linux: directly edit sound buffer data. There are also another solution: rewrite all code to use OpenAL, which now supported by Emscripten: https://github.com/kripken/emscripten/blob/master/src/library_openal.js . Which approach I should use?
- Tizen port uses Tizen::Media::Player which have poor api (I see no way to implement pinch&pan). I also propose rewrite it to OpenAL which is part of Tizen SDK 2.1 (cocos2d-x uses 2.0 at this moment).

P.S. Why win32 port have no license headers and contains non-unicode comments ;) ?

minggo:
@sergey-shambir 
Awesome job.
To be honest, i am not familiar with all platforms.
We focus on iOS and android. Other platforms are not target to release.
So i think you can do as much as you can for Emscriten and Tizen port.

minggo:
And other developers will contribute for it if he/she is familiar with that platform.

cocosQA:
Dear veena-peruvaje:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6498

PR #2697, opened 2013-05-31
Platforms: enabled c++11 flag on Linux, NaCL, Android, Emscripten.

sergey-shambir:


gelldur:
Finally :)

aberaud:
Why not using the standard flag " -std=c++11 " ?

This would force compilation with a compiler that actually support the final version of the standard.

sergey-shambir:
GCC 4.6 is primary platform for cocos2d v.3, because it's default compiler for Android, Blackberry, some Linux distributions. It supports all important features of c++11, but have no flag "-std=c++11" (which was turned on in version 4.7).

ricardoquesada:
yeah!

minggo:
@sergey-shambir 
Is there any difference between `-std=gnu++0x` and `-std=c++11`?
And could you please add comments in makefiles that describe why not use `-std=c++11`?

minggo:
I have sent a pull request #2709 to enable c++11. I used `-std=c++11`, it works ok on Android and linux(ubuntu 12.0).

I also enabled c++11 for all samples.

sergey-shambir:
Thanks, I'll close this request now.

minggo:
Yeap, `-std=c++11` failed on travis, so i use `-std=gnu++0x` instead.
Thank you @sergey-shambir.

PR #2699, opened 2013-06-01
Update XMLHTTPRequest.cpp

zj831007:
Add Multi-language support to metod of send .Modified last branch.


dumganhar:
Fixed at (master branch)https://github.com/cocos2d/cocos2d-x/pull/2710 and (develop branch)https://github.com/cocos2d/cocos2d-x/pull/2711. Thanks for your feedback.

PR #2703, opened 2013-06-01
Linux: added project part for QtCreator.

sergey-shambir:
Allows easily setup cocos2d on Linux in QtCreator IDE. User can just add 2 strings to `*.pro`, example:

``` c++
COCOS2D_MODULES += extensions box2d CocosDenshion
include(/home/sergey/Applications/cocos2d-x/cocos2dx/proj.linux/cocos2dx.pri)
```

It adds defines, include paths, scans COCOS2D_MODULES to include modules and builds (+links with) shared and static libraries on demand.

I can write wiki page about setup steps, if needed.

### DIFF
--- a/cocos2dx/proj.linux/cocos2dx.prf
+++ b/cocos2dx/proj.linux/cocos2dx.prf
@@ -2,9 +2,10 @@
 # Do not include this file in your project: see cocos2dx.pri.
 ################################################################################
 
-# (Set variables in global scope).
-
 linux {
+    # Use OpenAL in default QtCreator project.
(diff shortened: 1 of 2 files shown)


dumganhar:
Yes, it'll be better if there is a wiki page for describing this functionality. Thanks.

sergey-shambir:
There is wiki page: http://www.cocos2d-x.org/projects/cocos2d-x/wiki/QtCreator_project_setup_on_Linux

sergey-shambir:
I'll update it soon to support c++11 and switch to OpenAL by default (since FMOD is free only for non-commercial use).

dumganhar:
Merged. Thanks.

PR #2704, opened 2013-06-01
Hardware keyboard support

reneklacan:
Summary:
- added hw keyboard support for linux and mac + tests for it
- updated makefile for linux
- fixed ignored fwrite return value in FileUtilsTest that was preventing from successfull build on linux
- added #ifndef macro in zip support ioapi.h file because some version of minizip does not contain OF macro but contains _Z_OF
- added vim .swo to the .gitignore

Usage:
- You have to define "KEYBOARD_SUPPORT" directive

### DIFF
--- a/cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
+++ b/cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
@@ -167,7 +167,6 @@ void CCLayer::setTouchEnabled(bool enabled)
         m_bTouchEnabled = enabled;
         if (m_bRunning)
         {
-            printf(" -- running2\n");
             if (enabled)
             {
                 this->registerWithTouchDispatcher();


dumganhar:
Hi, travis-ci build failed, could you check that?

reneklacan:
problem was with build on adroid platform and I have realized that code for hw keyboard support on android is useless so I have used compiler directive so right now keyboard support is not even included for other platforms ... right now it is only compiled for linux but it isn't problem to make it work on windows, just "KEYBOARD_SUPPORT" needs to be added to directives.

cocos2dx/keyboard_dispatcher/CCKeyboardDelegate.h
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * @addtogroup input
+ * @{
+ */
+
+class CC_DLL CCKeyboardDelegate
+{
+public:
+    // The back key clicked
```
sergey-shambir: Wrong comment here and 3 lines below. Also use /// instead of // to mark comment as doxygen brief note.

cocos2dx/proj.linux/Makefile
```diff
@@ -9,6 +9,8 @@ INCLUDES += \
 			-I../../external/chipmunk/include/chipmunk \
 			-I../../extensions/network \
 
+DEFINES += -DKEYBOARD_SUPPORT
```
sergey-shambir: Not needed since cocos2dx.mk includes it.

rtissera:
We use very similar code on Win32, Mac and Linux at work (no KeyboardHandler though, no per layer support).
The implementation proposed here looks very good to me, please merge ;)
I can provide backends for Win32 / Mac if needed.

cocos2dx/proj.linux/Makefile
reneklacan: fixed

cocos2dx/keyboard_dispatcher/CCKeyboardDelegate.h
reneklacan: fixed like you said but also I dont see "///" used anywhere in the code when I inspected other source files, the most common is "/\* ... */"

minggo:
Thank you @simpliplant.
We are using std::function and std::bind to simplify CCLayer. 
The commit is [here](https://github.com/minggo/cocos2d-x/commit/a53936cd33d053f5ab29ce0ce4687f03ea16a8d6).

I think CCKeyboardDelegate can also use similar mechanism.

reneklacan:
You are definitely right but problem the problem is that this pull request is from develop branch where there is not C++11 compiler flag yet so if you will merge this code I promise that I will update the code to use std::bind. Ok?

Also is there any convention that guides when I am supposed to use "develop" or "master" branch for contribution?

minggo:
c++11 is enabled now.

master branch is used for releasing, develop branch is used for developing. develop branch will be merged to master branch when it is ready for releasing.

Because develop branch was created a few days ago. So there is a duration that we sent a pull request to master before releasing v2.1 final version. 

master is locked for releasing now. The new version will be released at about June 12.

reneklacan:
Thanks for explanation, I thought so.

I have also rewritten keyboard dispatcher as you adviced.

minggo:
@simpliplant 
Thank you.
Please let me know if it is ready to be merged.

reneklacan:
I have merged the latest code and it should be ready for merge.

rtissera:
Would be nice to see it merged for upcoming 2.1.4 release

minggo:
Thanks @simpliplant.
It is a pity that this pull request can not be merged, there is some conflict.
Could you please help to resolve it?

@rtissera 
v2.1.4 is in debugging for releasing.
So we will not add any new feature.
I am sorry about it.

reneklacan:
Sure. Pulled and manually resolved conflicts and tested out.

reneklacan:
Merged pull request from @mchinen which provides keyboard support for mac

reneklacan:
@minggo 
If you intend to merge keyboard support then maybe this is the right time... I have merged the latest code from cocos2d-x/develop right now.
Also consider https://github.com/cocos2d/cocos2d-x/pull/2776 :)

cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
```diff
@@ -161,6 +167,7 @@ void CCLayer::setTouchEnabled(bool enabled)
         m_bTouchEnabled = enabled;
         if (m_bRunning)
         {
+            printf(" -- running2\n");
```
minggo: Why not use CCLOG()?
printf can not work on all platforms.

extensions/spine/extension.cpp
```diff
@@ -60,10 +60,10 @@ char* _readFile (const char* path, int* length) {
 	fseek(file, 0, SEEK_SET);
 
 	data = MALLOC(char, *length);
-	fread(data, 1, *length, file);
+	size_t ret = fread(data, 1, *length, file);
```
minggo, line 63: ret is not used, why define it?

minggo, line 63: ret is not used, why define it?

cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
reneklacan: Little mistake, forgot to delete debug printf, I have deleted that line

extensions/spine/extension.cpp
reneklacan, line 63: because I was not able to compile it on linux because of compiler error that fread return value was ignored, that's all

minggo:
@simpliplant 
I will merge this pull request.
Thank you.

PR #2723, opened 2013-06-03
Removing unused variable

mdkcore0:
These variables are just used on CCAssert comparisons, and result on a compilation error on Linux (-Werror=unused-variable).


minggo:
Thank you.

PR #2726, opened 2013-06-03
A problem of CCDrawNode::drawPolygon

fispurring:
In this line :

```
bool outline = (fillColor.a > 0.0 && borderWidth > 0.0);
```

I think here should be borderColor rather than fillColor,right?


dumganhar:
Thanks.

PR #2727, opened 2013-06-03
Adds std::function support in CCCallFunc

ricardoquesada:
And fixes some minor issues on the iOS and OS X projects


ricardoquesada:
@minggo If you could merge this PR, it would be great. I know it is not compiling on Travis, but if you could patch it locally in your repo, and then push it, it would be great. Thanks.

minggo:
Ok.

PR #2728, opened 2013-06-03
Adds setObject to CCConfiguration

ricardoquesada:
also adds the test


samples/Cpp/TestCpp/Classes/ConfigurationTest/ConfigurationTest.cpp
```diff
@@ -222,3 +224,27 @@ void ConfigurationDefault::onEnter()
...
+//------------------------------------------------------------------
+void ConfigurationSet::onEnter()
+{
+    ConfigurationBase::onEnter();
+
+	CCConfiguration *conf = CCConfiguration::sharedConfiguration();
+
+	conf->setObject("this.is.an.int.key", CCInteger::create(10) );
+	conf->setObject("this.is.a.bool.key", CCBool::create(true) );
+	conf->setObject("this.is.a.string.key", CCString::create("hello world") );
```
dumganhar, line 241: `this.is.an.int.key` should be `this.is.an.int.value` ?

ricardoquesada, line 241: Thanks. You are correct.
Was this PR request not merged because of that ? If so, if you could fix tiny "bugs" like those in your repo, it would help me a lot.
eg: In this case, I think it is easier to fix it in your local repo, than add the comments here and then wait for my answer and new PR.
What's frustrating, is that it takes a whole day for this since we are in complete different time zones. Thanks.

dumganhar, line 241: OK, I'll fix it in my repo.

PR #2744, opened 2013-06-04
Testmenu sorted

ricardoquesada:
Menu from TestCpp refactored.
- Items are sorted alphabetically
- Much simpler to add a new test: in just one place.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/controller.cpp
+++ b/samples/Cpp/TestCpp/Classes/controller.cpp
@@ -9,7 +9,11 @@
 #include "testResource.h"
 #include "tests.h"
 
-static std::vector< std::tuple<std::string, std::function<TestScene*()>>> g_aTestNames = {
+
+struct {
+	const char *test_name;
+	std::function<TestScene*()> callback;
+} g_aTestNames[] = {
 
 	{ "Accelerometer", []() { return new AccelerometerTestScene(); } },
 	{ "ActionManagerTest", [](){return new ActionManagerTestScene(); } },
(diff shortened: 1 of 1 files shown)


dumganhar:
Travis-ci build failed. 
The reason is because your usage of initialize a tuple is the extension of clang. That'll break the compilation of GCC build. Please refer to this thread (http://stackoverflow.com/questions/14961809/returning-a-tuple-from-a-function-using-uniform-initialization-syntax).
The solution is to use 

<pre>
static std::vector< std::tuple<std::string, std::function<TestScene*()>>> g_aTestNames = {
    std::make_tuple("Accelerometer", []() { return new AccelerometerTestScene(); } ),
    std::make_tuple("ActionManagerTest", [](){return new ActionManagerTestScene(); } ),
    std::make_tuple("ActionsEaseTest", [](){return new ActionsEaseTestScene();} ),
    std::make_tuple("ActionsProgressTest", [](){return new ProgressActionsTestScene(); } ),
    std::make_tuple("ActionsTest", [](){ return new ActionsTestScene(); } ),
</pre>

samples/Cpp/TestCpp/Classes/controller.cpp
```diff
@@ -150,12 +93,15 @@ static TestScene* CreateTestScene(int nIdx)
...
     // add menu items for tests
     m_pItemMenu = CCMenu::create();
-    for (int i = 0; i < TESTS_COUNT; ++i)
+    for (int i = 0; i < g_testCount; ++i)
     {
 // #if (CC_TARGET_PLATFORM == CC_PLATFORM_MARMALADE)
 //         CCLabelBMFont* label = CCLabelBMFont::create(g_aTestNames[i].c_str(),  "fonts/arial16.fnt");
 // #else
-        CCLabelTTF* label = CCLabelTTF::create(g_aTestNames[i].c_str(), "Arial", 24);
+		std::pair<std::string, std::function<TestScene*()>> e = g_aTestNames[i];
```
dumganhar: This should be `std::tuple` rather than `std::pair`, right?

ricardoquesada:
Thanks @dumganhar 

I updated the PR.

dumganhar:
Thanks.

PR #2756, opened 2013-06-05
Adds C++11 std::function<> support to MenuItem

ricardoquesada:
Adds std::function<> support to `CCMenuItem`


ricardoquesada:
thanks for merging it. I know it was causing some conflicts on Linux with the deprecated functions.

PR #2768, opened 2013-06-05
solve android crash with CCLabelTTF setString function

jxgzxh111:


waiter:
It's really a big bug!!!!! Please merge it.

cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxBitmap.java
```diff
@@ -390,8 +390,11 @@ private static int computeY(final FontMetricsInt pFontMetricsInt,
 				}
 
 				/* Remove spaces at the beginning of a new line. */
-				while (pString.charAt(i) == ' ') {
-					++i;
+				if(i < charLength)
+				{
+					while (pString.charAt(i) == ' ') {
+						++i;
+					}
```
waiter, line 397: while (i < charLength && pString.charAt(i) == ' ') {
    ++i;
}

I think it must be like this.
@jxgzxh111

minggo:
@jxgzxh111 I think @waiter 's suggestion is right.

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2907.
Thanks.

yihuang:
What is this bug, how to reproduce?
My game have some crash reports, i'm not sure if they are due to this bug.

PR #2776, opened 2013-06-05
Support for loading tilemaps with plain XML layer format

reneklacan:
Loading tilemap supports only layers with base64 encoded data (uncompressed/compressed).

Change in this pull request adds possibility to load TMX map with XML layer format.

### DIFF
--- a/cocos2dx/tilemap_parallax_nodes/CCTMXXMLParser.cpp
+++ b/cocos2dx/tilemap_parallax_nodes/CCTMXXMLParser.cpp
@@ -72,7 +72,6 @@ TMXLayerInfo::TMXLayerInfo()
 , _minGID(100000)
 , _maxGID(0)        
 , _offset(PointZero)
-, _currentTileIndex(0)
 {
     _properties= new Dictionary();;
 }
@@ -408,9 +407,16 @@ void TMXMapInfo::startElement(void *ctx, const char *name, const char **atts)
         if (pTMXMapInfo->getParentElement() == TMXPropertyLayer)
         {
             TMXLayerInfo* layer = (TMXLayerInfo*)pTMXMapInfo->getLayers()->lastObject();
+            Size layerSize = layer->_layerSize;
             unsigned int gid = (unsigned int)atoi(valueForKey("gid", attributeDict));
-            layer->_tiles[layer->_currentTileIndex] = gid;
-            layer->_currentTileIndex++;
+            int tilesAmount = layerSize.width*layerSize.height;
+            int currentTileIndex = layer->_tiles[tilesAmount - 1];
+            layer->_tiles[currentTileIndex] = gid;
+
+            if (currentTileIndex != tilesAmount - 1)
+            {
+                layer->_tiles[tilesAmount - 1] = currentTileIndex + 1;
+            }
         }
         else
         {
@@ -503,7 +509,7 @@ void TMXMapInfo::startElement(void *ctx, const char *name, const char **atts)
 
             int *tiles = (int *) malloc(tilesAmount*sizeof(int));
             for (int i = 0; i < tilesAmount; i++)
-                tiles[i] = -1;
+                tiles[i] = 0;
 
             layer->_tiles = (unsigned int*) tiles;
         }
--- a/cocos2dx/tilemap_parallax_nodes/CCTMXXMLParser.h
+++ b/cocos2dx/tilemap_parallax_nodes/CCTMXXMLParser.h
@@ -99,7 +99,6 @@ public:
     unsigned int        _minGID;
     unsigned int        _maxGID;
     Point               _offset;
-    unsigned int        _currentTileIndex;                 
 public:
     TMXLayerInfo();
     virtual ~TMXLayerInfo();


cocos2dx/tilemap_parallax_nodes/CCTMXXMLParser.cpp
```diff
@@ -404,14 +404,32 @@ void TMXMapInfo::startElement(void *ctx, const char *name, const char **atts)
...
-        
-        pTMXMapInfo->setParentElement(TMXPropertyTile);
+        if (pTMXMapInfo->getParentElement() == TMXPropertyLayer)
+        {
+            TMXLayerInfo* layer = (TMXLayerInfo*)pTMXMapInfo->getLayers()->lastObject();
+            Size layerSize = layer->_layerSize;
+            unsigned int gid = (unsigned int)atoi(valueForKey("gid", attributeDict));
+            int tilesAmount = layerSize.width*layerSize.height;
+
+            for (int i = 0; i < tilesAmount; i++)
```
boyu0: This is not very efficiency. It's start at 0 to n-1 to find out the correct tile, it's not necessary, can you store the count and reuse it?

reneklacan: I agree with you but I did not want to make a big impact on codebase and I considered fact that this code is not called during the gameplay but only on the start ... but ok, I will fix that :)

boyu0: You can store the count at tile[tilesAmount-1]
at data parse start:
tiles[tilesAmount-1] = tilesAmount - 2;
........

at tile:
--tiles[tilesAmount];
tiles[tilesAmount - 3 - tiles[tilesAmount-1]] =  gid;

sample
we have 5 elements
t[0]  t[1]  t[2]  t[3]  t[4] 
start t[every] = -1   count=5  t[4] = 3

index 0: t[4]=2  t[5 - 3 - t[4]] -> t[0] = gid0
index 1: t[4]=1  t[5 - 3 - t[4]] -> t[1] = gid1
index 2: t[4]=0  t[5 - 3 - t[4]] -> t[2] = gid2
index 3: t[4]=-1 t[5 - 3 - t[4]] -> t[3] = gid3
index 4: t[4]=-2 t[5 - 3 - t[4]] -> t[4] = gid4

When layer parse end, you can test t[count-2],  if  it's -1, you change the t[count-1] to -1, if it's not -1, the t[count-1] is -1 or the gid number, so you can just leave it.

this won't add an extra count variable to and more efficiency.

reneklacan: Oh really nice idea, but I updated code with extra variable before I could see your comment :-S ... should I rewrite it?

reneklacan: I have rewritten it like you said :)

cocos2dx/tilemap_parallax_nodes/CCTMXXMLParser.cpp
```diff
@@ -404,14 +404,30 @@ void TMXMapInfo::startElement(void *ctx, const char *name, const char **atts)
...
+            Size layerSize = layer->_layerSize;
+            unsigned int gid = (unsigned int)atoi(valueForKey("gid", attributeDict));
+            int tilesAmount = layerSize.width*layerSize.height;
+            int currentTileIndex = layer->_tiles[tilesAmount - 1];
+            layer->_tiles[currentTileIndex] = gid;
+
+            if (currentTileIndex != tilesAmount - 1)
+            {
+                layer->_tiles[tilesAmount - 1] = currentTileIndex + 1;
+            }
```
boyu0, line 419: Sorry, I think you didn't doing right.
First, you didn't detect and set layer->_tiles[tilesAmount - 1] to 0  at Layer parse end, so When I don't have tiles[tilesAmount - 1] data in TMX, we get a wrong data.
Second, if I have tiles[tilesAmount - 2] data and don't have tiles[tilesAmount - 1] data in TMX,  we have no way to find out whether the tiles[tilesAmount - 1] is a real data or just a count.
So if you want do it my way, I think you'd better understand my previous comment and think carefully.

minggo, line 419: Thanks @simpliplant and @boyu0.

@boyu0 
I think it is better if you can write some pseudocode.

boyu0, line 419: OK....
at data parse
for each tiles
    tiles[i] = -1;
tiles[tilesAmount-1] = tilesAmount - 2;

in elementName == "tile"
--tiles[tilesAmount];
tiles[tilesAmount - 3 - tiles[tilesAmount-1]] = gid;

in layer parse end
if(tilesAmount >= 2 && tiles[TilesAmount - 2] == -1)  tiles[TilesAmount - 1] = -1

reneklacan, line 419: I did understand your comment but if the layer had more or less elements then "tileAmount" then tilemaps would be corrupted, exported maps via Tiled has always layers with "height*width" elements in tilemaps. Am I not right?. So I thought that this kind of control would be unnecessary. Decoding of compressed and encoded tilemaps does not do any kind of control either and if you have corrupted map it will just segfault or will get some unexpected result. This code have been tested in our game where we are using tilemaps with plain XML format and everything worked ok ... But if you insist I can rewrite it and test for corrupted maps. I can also add some tests to TestCpp if you want to.

boyu0, line 419: Thank you for your reply.
I write TMX with width and height are both 40, and I only write two tile nodes in `<data>`, I open it with Tiled, it doesn't corrupt, and it shows correctly in my mac and win7, That's why I want that extra control.
I haven't test it in encoded tilemaps, If it corrupted in our engine, I will fix it.
This is my TMX file of `<layer>` node, you can test it in your computer.

```
<layer name="Ground" width="40" height="40">
  <data>
   <tile gid="30"/>
   <tile gid="30"/>
  </data>
 </layer>
```

PR #2778, opened 2013-06-06
on dealloc, log "this" with %p, not %x.

ricardoquesada:
Also adds %p where it was missing.
And make a few fixes on the iOS project


dumganhar:
Thanks.

PR #2794, opened 2013-06-06
Missing config files

ricardoquesada:
...and add functionality on the Director to override the default values
    from the config file.

```
Supported values:

 - cocos2d.x.fps
 - cocos2d.x.display_fps
 - cocos2d.x.gl.projection
 - cocos2d.x.texture.pixel_format_for_png
 - cocos2d.x.texture.pvrv2_has_alpha_premultiplied

If the "config.plist" file is not present, or some keys are not present,
the default values are going to be used

IMPORTANT:
 The folder "samples/Cpp/TestCpp/Resources/configs" needs to be added
 in the Android, Linux, Win32 projects.
 Already added in OS X and iOS Xcode projects.
```


ricardoquesada:
Thanks for merging it.
Would it be possible to apply it on the `develop` branch as well ? thanks!

PR #2795, opened 2013-06-06
Fix VS build (nullptr instead of NULL when assigning to std function)

slobodin:


slobodin:
Just append to hotfix windows build. And also, I do not understand how can we cast C++ reference to NULL as is done, for example, in CCMenuItem:
`_callback(NULL)`

ricardoquesada:
yes, my bad... it should be `nullptr`... and clang should have raised a warning.

PR #2796, opened 2013-06-07
Adds Base class to tests

ricardoquesada:
Base class displays the Menu( <- O -> )
Improved in some tests the way to create entries.
Instead of defining the tests twice: the "create" method, and the entry
   in the array, it is only needed in the array (ActionsTest, SpriteTest
   and Texture2DTest)

**IMPORTANT**:
- Android, Linux, Emsciprten and Win32 projects were not fixed.
  The files `BaseTest.cpp` and `BaseTest.h` needs to be added in their make files.


ricardoquesada:
Thanks for fixing the linux,win32,etc. makefiles

PR #2812, opened 2013-06-08
CPP Tests migrated to the new C++11 API

ricardoquesada:
Many tests were converted to the C++11 API
and some tests were also simplified by using the "CL" macro
instead of the "CF" macro.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/Box2DTest/Box2dTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/Box2DTest/Box2dTest.cpp
@@ -115,23 +115,20 @@ void Box2DTestLayer::initPhysics()
 
 void Box2DTestLayer::createResetButton()
 {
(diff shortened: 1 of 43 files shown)


EduKoala:
menu_selector looks great now std::bind looks so ugly :(

ricardoquesada:
@EduKoala 
You can still use `menu_selector`...  I found `menu_selector` somewhat ugly as well.
If you don't want to use `std::bind`, you can use a lambda function, or a free function... basically any callable object is supported in the new API

ricardoquesada:
An alternative is:
- since `menu_selector` is just a macro, we could create a new macro called `menu_selctor2` that does that, but for the new API. eg:

``` c++
#define menu_selector2( __function__, __instance__)  std::bind( __function__, __instance__, std::placeholder::_1)
```

PR #2824, opened 2013-06-08
Cast variables to their own type, and print warning info if no corresponding lua callback function instead of crash

edwardzhou:
1) cast variables to their own type; 
2) print warning info if no corresponding lua callback function instead of crash;


dumganhar:
Thanks

PR #2850, opened 2013-06-09
Making CCScrollView zoom behaving more like a UIScrollView when a user tries

LiyeZhang:
... to zooming in and out. That is, when zooming out would change its offset to be beyond the max or below the min, to cuts it off at either the max or the min.


dumganhar:
Hi, @LiyeZhang 
Since `master` branch is for released only, could you please send this PR to develop branch?
BTW, could you also add a test case for this bugfix in TestCpp? That'll be better. Thanks.

PR #2869, opened 2013-06-11
Jsb function getCPBody return type is not right

fispurring:
Jsb function getCPBody return a number,not a CPBody pointer that we expect.


dumganhar:
Thanks.

chygmm:
this question still exist!

PR #2871, opened 2013-06-11
CGBitmapContextCreate param type error in iOS 7.

pilwon:
Fixes `CGBitmapContextCreate` param type error visible in `iOS 7` when running test JS projects (tested w/ Developer Preview for iPhone/XCode5):

```
Jun 10 ##:##:## XXX.local YYY[58909] <Error>: CGBitmapContextCreate: unsupported parameter combination: 8 integer bits/component; 32 bits/pixel; 3-component color space; kCGImageAlphaPremultipliedLast; 722 bytes/row.
```
- [Reference] [Image resized error: CGBitmapContextCreate: unsupported parameter](http://stackoverflow.com/questions/2235210/image-resized-error-cgbitmapcontextcreate-unsupported-parameter)


minggo:
Thanks @pilwon, i think this bug was fixed in #2873.

PR #2873, opened 2013-06-11
Fixes for iOS7 and Xcode5

ricardoquesada:
Fixes for iOS 7 and Xcode 5.
The main fix is in CCImage.cpp. If not applied, labels won't render correctly on iOS7.


minggo:
merged into master too.

rafiamerald:
Hey Thanks man, I fixed it.

PR #2874, opened 2013-06-11
LabelTTF vertical alignment fixed on OS X

ricardoquesada:
vertical alignment was broken on OS X.
top and bottom were switched.

IMPORTANT:
This patch should be applied in "master" as well (both on "develop" and "master")

### DIFF
--- a/cocos2dx/platform/mac/CCImage.mm
+++ b/cocos2dx/platform/mac/CCImage.mm
@@ -373,7 +373,7 @@ static bool _initWithString(const char * pText, cocos2d::CCImage::ETextAlign eAl
 		
 		// alignment, linebreak
 		unsigned uHoriFlag = eAlign & 0x0f;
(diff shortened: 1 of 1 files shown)


minggo:
@ricardoquesada 

> IMPORTANT:
> This patch should be applied in "master" as well (both on "develop" and "master")

But master has been released. And developer will be merged into master when releasing a new version.

ricardoquesada:
@minggo 
I don't think it will be a problem to merge this patch in master.
Git won't apply it again in develop if it is already present there.

minggo:
What i mean is that i have uploaded zip file to google code for releasing. If i merge this code into master i have to replace the zip file with these code. Should i?

minggo:
merged into master.

ricardoquesada:
@minggo yes, you will need to upload a new version to google code :)

PR #2875, opened 2013-06-11
Adds CALLBACK_0, CALLBACK_1 and CALLBACK_2

ricardoquesada:
Handy macros for callbacks.
Instead of:
    std::bind(func_ptr, instnace, std::placeholder::_1)
you use:
    CALLBACK_1(func_ptr, instance)

And also removes more warnings from the tests


sergey-shambir:
Please add prefix CC_ before macro.

ricardoquesada:
yes, I think it should have the `CC` prefix, but I didn't add it because `schedule_selector`, `menu_selector`, `compare_selector`, etc... which are macros, don't use the `CC` prefix. And they are in lower case...

ricardoquesada:
Patch with the `CC` prefix:
https://github.com/cocos2d/cocos2d-x/pull/2885

PR #2881, opened 2013-06-12
Improvement in CCControllButton

gelldur:
Litle fix in CCControllUtils in CCRectUnion. Also i changed CCScale9Sprite to CCNode now you can set fixed sprite as a button.

It should be working with old CCScale9Sprite (i have tested it) but now TestCPP is corrupted (without my changes too), but when i tested it before pull from master everything works great.

Thanks to that change you can pass normal CCSprite as a button and it shouldn't stretch.

For example

```
CCControlButton* pButton3 = CCControlButton::create (
                            CCSprite::create ( "test/hw2.png" ) );
pButton3->setPosition ( 200, 200 );
addChild ( pButton3 );
```

### DIFF
--- a/extensions/GUI/CCControlExtension/CCControlButton.cpp
+++ b/extensions/GUI/CCControlExtension/CCControlButton.cpp
@@ -43,18 +43,18 @@ enum
 };
 
 CCControlButton::CCControlButton()
-: m_currentTitle(NULL)
+: m_currentTitle(nullptr)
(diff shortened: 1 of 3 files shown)


gelldur:
i must fix nullptr i thought that cocos2dx uses C++11

dumganhar:
Hi, could you send it to develop branch?

gelldur:
Yeah shure

PR #2884, opened 2013-06-14
Actions with clone() pattern 1/3

ricardoquesada:
This is the 1st patch of 3.

1)
Removes the copyWithZone() patterns.
Adds the clone() pattern instead (which is more C++ friendly)
Adds also "const" to getters.

### DIFF
--- a/cocos2dx/actions/CCAction.cpp
+++ b/cocos2dx/actions/CCAction.cpp
@@ -37,20 +37,9 @@ NS_CC_BEGIN
 //
 
 CCAction::CCAction()
-:CCObject()
-,m_pOriginalTarget(NULL)
+:m_pOriginalTarget(NULL)
 ,m_pTarget(NULL)
 ,m_nTag(kCCActionTagInvalid)
-{
-}
-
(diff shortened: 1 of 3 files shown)


cocos2dx/actions/CCAction.cpp
```diff
@@ -37,12 +37,23 @@ of this software and associated documentation files (the "Software"), to deal
 //
 
 CCAction::CCAction()
-:m_pOriginalTarget(NULL)
+:CCObject()
```
minggo: Why should invoke father's constructor explicitly?

ricardoquesada: yep... my bad, there is no need to call it explicitly since it is the default constructor.

ricardoquesada:
@minggo 
fixed.

dumganhar:
Hi Riq, 
I merged my PR in bindings-generator(https://github.com/cocos2d/bindings-generator/pull/9).
It will not generate jsbinding glue codes for copy constructor.

PR #2886, opened 2013-06-14
Develop

dpull:


cocos2dx/base_nodes/CCNode.cpp
```diff
@@ -841,14 +841,11 @@ void CCNode::visit()
             }
         }        
     }
-    else
+    elses
```
ricardoquesada: `elses` instead of `else` ?
strange.

dpull: sorry

PR #2896, opened 2013-06-15
issue #2129: New naming of member varibles, renaming m_iVar to _var.

dumganhar:


gelldur:
why you now change convention ? m_ to _ prefix ? Now it is unreadable

dumganhar:
Hi @gelldur 
Since cpp is strong type language, without m_[b | f | p | s] will be compact.
Lots of libraries don't use `Hungarian` naming now. It's too verbose.
You could refer to `Box2D`, `GamePlay`, `cocos2d-iphone`, etc.

gelldur:
Hi @dumganhar
ok i understand why not use [b | f | p | s]  but m_ is good i think even now days :). Ok then.

dumganhar:
@gelldur 
I also like `m_` as prefix. There are some choices to name a member variable:
1) `m_XXX` (Used in Box2d)
2) `mXXX`  (Often used in Java)
3) `_XXX`  (Used in GamePlay and cocos2d-iphone)
I think we should discuss which one is more suitable.
@minggo and @ricardoquesada prefer 3).
But IMO, 2) is better since we could have `sXXX` as static variable and `gXXX` as global variable.
What do you think?

gelldur:
Hi @dumganhar 
I prefer 1 (the best)  or 2 . 3 i see a lot in python or MS code. But decision is up to you i must adapt :)

When someone uses 3 i think is this private member or something like "do not use" as python programmers use it (i think)

walzer:
@gelldur Now it's similar to Google Cpp Style http://google-styleguide.googlecode.com/svn/trunk/cppguide.xml#Variable_Names
The current style m_iVar was applied by me since 2010, when cocos2d-x started. To be honest it's a Microsoft style because I worked on WinCE/WinMobile system for 5 years. 
But now Ricardo Quesada & Minggo prefer option (3). I can pros like
(1) it's closer to google cpp style
(2) it's closer to javascript code, so maintaining cocos2d-html5 version will be eaiser

nickveri:
Another idea would be to always access member variables using "this", e.g., this->hasChildren. This approach is closer to JavaScript and it makes the use of a prefix redundant. Prefixes for static or global variables might still be necessary though.

PR #2901, opened 2013-06-16
Get rid of unwanted logs entries

jotel:


minggo:
Thanks @jotel.
And i think it is better to comment these codes.
Because you may want to use these codes when debugging.

What's your opinion?

jotel:
Hey, resolved. I added this log entries when debugging the this part of code I created. Now when they work I don't think they would be useful... If by any chance debugging will be needed in the future this simple log entries can be easily recreated...

Thanks
Jaro

cocosQA:
Dear jotel:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6495

PR #2902, opened 2013-06-17
update "fix a bug that SimpleAudioEngine may cause crash on i9100" 

waiter:
Since the recreatePlayer function return bool, I think it will be better to determine whether the recreatePlayer function is successfully or not before call setEffectLooping function.

And, when I use "fix a bug that SimpleAudioEngine may cause crash on i9100", it still crashes sometime. I don't know why. So I change the code. It may be more safe, and I have not found the crash.

Hope can understand my English.


minggo:
@waiter
Did you mean before you applying the patch, it would crash. After applying this patch, the crash was fixed?

waiter:
@minggo 
Before I applying the patch, it would still crash sometime. And I do not know why.
After applying this patch, the crash is not happen again. But I'm not sure whether the bug is fixed or not.

minggo:
@waiter 
Thank you.

PR #2904, opened 2013-06-17
fix WebSocket url parse error for 'ws://domain.com/websocket' pattern

edwardzhou:
the existing code only works fine for ws://domain.com, but not for  ws://domain.com/websocket. it will be parsed as
_host = "domain.com/websocket"
_port = 80
_path = "/"

the correct parse should be:
_host = "domain.com"
_port = 80
_path = "/websocket"

this patch also works for "ws://domain.com:8080/websocket"


dumganhar:
Thanks.

PR #2911, opened 2013-06-18
Purge CCScriptEngineManager

toughrogrammer:
CCScriptEngineManager did not release


dumganhar:
Why are there so many commits?

toughrogrammer:
I was not accustomed to use git.

sungjuly:
I think @growingdever should rebase it. you can find git command. `git rebase`

PR #2916, opened 2013-06-18
fixed #2124: make CCImage::initWithIamgeFileThreadSafe() thread safe

minggo:


shane-yao:
hi, mingo

This patch fixed the addImageAsync thread issue，but seem it only pull to 3.0alpha-pre, neither in 2.1.4 nor in 2.1.5。

Can you merge it to 2.x.x or should I make a PR for this? Thanks for this fix any way.

minggo:
@future0906 
The codes are changed, because we use Android AssetManager to load file, so it doesn't have the problem caused by ZipFile. Because AssetManager is enabled in Android 2.3, and cocos2d-x 2.x should support Android 2.2. So can not merge the feature into cocos2d-x 2.x.

shane-yao:
Sorry, I don't quite get your meaning.

Did you means cocos2d-x 2.x on Android platform use "Android AssetManager" when it's enabled(Android 2.3 or later)？"ZipFile" is used for Android 2.2 support?

If this is correct, why did this issue happen when I test game on my XiaoMi 2S Android 4.1?And when I patch this merge on 2.1.5，this issue fixed.I checked and logged the code ,found cocos2dx is using zip file to read the file, not AssetManager.

Is this relative to this message when I run ./build_native.sh:
":128: Android NDK: WARNING: APP_PLATFORM android-14 is larger than android:minSdkVersion 8 in ./AndroidManifest.xml" ?

minggo:
What i mean is that we use another method to fix this issue, which used `Android AssetManager`. And `Android AssetManager` is available since 2.3. 

Of course we can merge this pull request into v2.x.

shane-yao:
OK,I think I understand what you mean.I will try to create a pull request to v2.x.

minggo:
Thanks.

PR #2919, opened 2013-06-18
Asynchronous Image loading for Emscripten

j4m3z0r:
Added async image loading for Emscripten builds. Async loaded images are fetched via a normal HTTP request for the image, so when this is used, it allows for a much smaller initial download since images can just be fetched on demand.

This is done by adding a new CCTextureCache subclass rather than a nest of #ifdefs. I think this is a cleaner design. Note also that a .js file was added in order to implement the actual async image loader -- this requires slightly different flags to Emscripten (--js-library).

The async image loader will divide up images into multiple sub-rectangles, and pixel data is alpha pre-multiplied and copied into the texture buffer over multiple frames. This approach helps to minimize UI stalls when textures are loading.

Firefox load times are greatly improved by this change (~3X), but Chrome is actually a little bit slower. I'm yet to look into why, but believe something like this is the right long-term solution, so pushing this up now.


minggo:
Thanks @j4m3z0r.
Because of #2129, this pull request can not be merged automatically.

dumganhar:
@minggo @j4m3z0r 
This Pull Request was merged at https://github.com/cocos2d/cocos2d-x/pull/2924.

@j4m3z0r 
Since we are now developing cocos2d-x3.0 on develop branch, cocos2d-x 3.0 uses some c++11 features.
I noticed that you didn't pull our latest codes on develop branch. 
So could you pull the latest codes before you send another Pull Request next time? Thanks.

j4m3z0r:
Will do.

Will this PR get merged into master, or have I missed the window for that? This is a big win for Emscripten built games, so would like to find a way to integrate it there if possible.

dumganhar:
No, it will not be in master branch before next release. master branch updates only when we release a new version now. Let's work on `develop` branch only. Thanks.

PR #2920, opened 2013-06-18
Clone reverse fixes

ricardoquesada:
2nd part of big patch.
- `copyWithZone()` replaced by `clone()`
  `clone()` already returns an autoreleased object. So, no need to call `autorelease` manually.
- `reverse()` is const, and in some cases is pure virtual function.

examples updated.
Added missing `clone()` and `reverse()`


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/2922

dumganhar:
Thanks.

cocos2dx/actions/CCActionGrid.cpp
```diff
@@ -93,6 +75,11 @@ void CCGridAction::startWithTarget(CCNode *pTarget)
     }
 }
 
+CCGridAction* CCGridAction::reverse() const
+{
+	return (CCGridAction*)CCReverseTime::create( this->clone() );
```
dumganhar, line 80: Is `CCReverseTime` an instance of `CCGridAction` ?

ricardoquesada, line 80: Nope. Actually that could trigger a potential bug.
I should add the XXX there.

PR #2933, opened 2013-06-19
Compiles on OS X

ricardoquesada:
compiles on Mac.

BTW, we need pure virtual functions for these kind of interfaces.

### DIFF
--- a/cocos2dx/platform/mac/CCAccelerometer.h
+++ b/cocos2dx/platform/mac/CCAccelerometer.h
@@ -25,6 +25,7 @@ THE SOFTWARE.
 #ifndef __PLATFORM_MAC_CCACCELEROMETER_H__
 #define __PLATFORM_MAC_CCACCELEROMETER_H__
 
+#include <functional>
 #include "platform/CCAccelerometerDelegate.h"
 
 NS_CC_BEGIN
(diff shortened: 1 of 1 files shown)


scripting/javascript/bindings/generated
```diff
@@ -1 +1 @@
-Subproject commit f83f542e173bb82b281b617c31248f0ad8b01df7
+Subproject commit f85c95751a8a59846776789be30519e45b3911fd
```
dumganhar, line 1: Why was the submodule changed?

dumganhar:
> we need pure virtual functions for these kind of interfaces.

I totally agree. :)

minggo:
CCAccelerometer is not interface, it is a class to register to receive message.

dumganhar:
@minggo @ricardoquesada 

> CCAccelerometer is not interface, it is a class to register to receive message.

Yes, @minggo  is right. `CCAccelerometer` isn't a interface class.
But I think `CCAccelerometerDelegate` should be pure virtual class.

``` cpp
class CC_DLL CCAccelerometerDelegate
{
public:
    virtual void didAccelerate(CCAcceleration* pAccelerationValue)  = 0;  // Changing to pure virtual.
};
```

minggo:
CCAccelerometerDelegate is removed.

minggo:
Thanks. It is resolved in #2936.

PR #2939, opened 2013-06-20
Update create_project.py

toughrogrammer:
'Sample 1' way is not okay. If not input language, script doesn't generate project. (Must input language, but 'Sample 1' way said that you don't need to input language.)


dumganhar:
Thanks.

PR #2944, opened 2013-06-20
issue #2129: Removing CC prefix and renaming member variables for plugin module.

dumganhar:


jdmunro:
Is the CC prefix removal going to have any effect on the JSB API?

dumganhar:
> Is the CC prefix removal going to have any effect on the JSB API?

It'll not affect JSB API.

PR #2947, opened 2013-06-20
[Tizen] Fix compile error

pyrasis:
- Fix compile error.

### DIFF
--- a/extensions/LocalStorage/LocalStorage.cpp
+++ b/extensions/LocalStorage/LocalStorage.cpp
@@ -29,7 +29,7 @@
 
 #include "cocos2d.h"
 
-#if (CC_TARGET_PLATFORM != CC_PLATFORM_ANDROID)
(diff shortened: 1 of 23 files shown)


extensions/LocalStorage/LocalStorage.cpp
```diff
@@ -29,7 +29,7 @@
 
 #include "cocos2d.h"
 
-#if (CC_TARGET_PLATFORM != CC_PLATFORM_ANDROID)
+#if (CC_TARGET_PLATFORM != CC_PLATFORM_ANDROID && CC_TARGET_PLATFORM != CC_PLATFORM_TIZEN)
```
dumganhar, line 32: Tizen port has its own implementation of localStorage?

pyrasis, line 32: It is temporarily.
I will prepare and add sqlite3 library for Tizen x86 and armv7l.

dumganhar, line 32: OK

dumganhar:
Merged. Thanks.

Issue #2948, opened 2013-06-20
a potential bug in Layer.


zcgit:
bool Layer::init()
{
    bool bRet = false;
    do 
    {  
        Director \* pDirector;
        CC_BREAK_IF(!(pDirector = Director::sharedDirector()));
        this->setContentSize(pDirector->getWinSize());
        _touchEnabled = false;
        _accelerometerEnabled = false;
        // success
        bRet = true;
    } while(0);
    return bRet;
}

we should replace "_touchEnabled = false;" to setTouchEnabled(bool);

dumganhar:
Hey, is it needed?
in the constructor:

``` cpp
Layer::Layer()
_touchEnabled(false)
```

in Layer::init():
If invoking setTouchEnabled(false), it'll take no effect. You could look inside the logic of `Layer::setTouchEnabled`

zcgit:
I'm sure it's a potential bug.
for example:
    1.you create a Layer by Layer\* pLayer = Layer::create(),
    2.added it to a Node, then pLayer is running. 
    3.then do pLayer-> setTouchEnabled(true); so it will invoke registerWithTouchDispatcher() to addXXXDelegate.
    4.execute pLayer->init(); at this time _touchEnabled is false; but it never removed XXXDelegate!

dumganhar:
But `init` should not be invoked since you're using `create` pattern.
Or it should only be called right after `Layer* p = new Layer();`

zcgit:
yes , but if you do this in CCLayerColor , you will find if you do these below:
1.you create a Layer by CCLayerColor\* pLayer = CCLayerColor::create(),
2.added it to a Node, then pLayer is running. 
3.then do pLayer-> setTouchEnabled(true); so it will invoke registerWithTouchDispatcher() to addXXXDelegate.
4.execute pLayer->init(); at this time _touchEnabled is false; but it never removed XXXDelegate!

often , we want to execute pLayer->init() to reset CCLayerColor. so if we do this , it will cause add twice XXXDelegate to CCLayerColor if you call setTouchEnabled(true) again. just suppose we forgot invoked setTouchEnabled before.

dumganhar:
It makes sense. So could you please send a Pull Request for this issue?

PR #2950, opened 2013-06-20
a potential bug in Layer.

zcgit:
https://github.com/cocos2d/cocos2d-x/issues/2948

### DIFF
--- a/extensions/GUI/CCScrollView/CCScrollView.cpp
+++ b/extensions/GUI/CCScrollView/CCScrollView.cpp
@@ -183,7 +183,7 @@ void ScrollView::setTouchEnabled(bool e)
     {
         _dragging = false;
         _touchMoved = false;
-        _touches->removeAllObjects();
+        if(_touches)_touches->removeAllObjects();
     }
 }
 


cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
```diff
@@ -77,7 +77,8 @@ bool Layer::init()
         Director * pDirector;
         CC_BREAK_IF(!(pDirector = Director::sharedDirector()));
         this->setContentSize(pDirector->getWinSize());
-        _touchEnabled = false;
+        //_touchEnabled = false;
```
dumganhar: I think delete the comment will be better.

cocos2dx/layers_scenes_transitions_nodes/CCLayer.cpp
```diff
@@ -77,7 +77,8 @@ bool Layer::init()
...
+        setTouchEnabled(false);
         _accelerometerEnabled = false;
```
dumganhar: Does `_accelerometerEnabled = false` also need to be `setAccelerometerEnabled(false)`?

zcgit:
oh , no , I'm sorry , I found a bug in CCScrollView after I modified it , I' thinking how to solve it ...

dumganhar:
Thanks.

PR #2955, opened 2013-06-20
NaCl port fixes

sbc100:
Recent NaCl fixes for the develop branch.

This re-enabled travis for NaCl ARM.  NaCl's x86 compiler is gcc 4.4.3 which only has partial support for C++11 so will need some more work.

I had posted a similar PR against master which I will close.

The travis build is failing because the GEN_JS stage doesn't seem to work on forked repos :(  This is unrelated to this change.


dumganhar:
Merged, Thanks.

PR #2956, opened 2013-06-21
Re-adds 'CC' prefix to objective-c classes.

ricardoquesada:
And also makes it compile on Mac, which was broken


dumganhar:
Merged, thanks.

PR #2970, opened 2013-06-21
a potential bug in ScrollView

zcgit:
check _touches in setTouchEnabled

### DIFF
--- a/cocos2dx/base_nodes/CCNode.h
+++ b/cocos2dx/base_nodes/CCNode.h
@@ -1375,6 +1375,8 @@ protected:
     
     int _tag;                         ///< a tag. Can be any number you assigned just to identify this node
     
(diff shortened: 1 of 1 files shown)


extensions/GUI/CCScrollView/CCScrollView.cpp
```diff
@@ -183,7 +183,7 @@ void ScrollView::setTouchEnabled(bool e)
     {
         _dragging = false;
         _touchMoved = false;
-        _touches->removeAllObjects();
+        if(_touches)_touches->removeAllObjects();
```
dumganhar: Could you format the codes to 

``` cpp
if (_touches)
{
    _touches->removeAllObjects();
}
```

zcgit:
ok,I'll do it right now.

dumganhar:
Fixed in the latest codes on develop branch. Thanks.

PR #2971, opened 2013-06-21
Merge PR https://github.com/cocos2d/cocos2d-x/pull/2969

dumganhar:


minggo:
Compiling errors on native client and emscripten port.
- native client: it is because the compiler can not support c++11 well. So i think we should disable it in travis CI.
- emscripten port: it is because it uses a structure named AsyncStruct, which is put in to CCTextureCache private field. So should declare it with public and modify the usage of it in CCTextureCacheEmscripten.h/cpp.

dumganhar:
Closed it since it couldn't be auto- updated when I had new commits.

quiasmo:
Hey, I'm looking at the emscripten files in 3.0x and they have vanished without a trace. Do you guys know anything about that?

dumganhar:
It's in our roadmap. It will probably be supported in cocos2d-x3.2

quiasmo:
Hey, I would be happy to hack together some pull requests to get this working asap, but I can’t seem to find the legacy code in the repositories.
Where can I find it?

PR #2974, opened 2013-06-22
[Tizen] Fix compile error

pyrasis:
- Fix compile error.
- The range-based for loop is supported by GCC since 4.6. but Tizen SDK has GCC 4.5

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ActionsTest/ActionsTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/ActionsTest/ActionsTest.cpp
@@ -1371,6 +1371,22 @@ void ActionStacked::runActionsInSprite(Sprite *sprite)
 
 void ActionStacked::ccTouchesEnded(Set* touches, Event* event)
 {
+#ifdef CC_PLATFORM_TIZEN
+    SetIterator it;
+    Touch* touch;
+
+    for( it = touches->begin(); it != touches->end(); it++)
+    {
+        touch = (Touch*)(*it);
+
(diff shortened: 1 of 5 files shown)


sergey-shambir:
I've created bug in Tizen tracker: https://bugs.tizen.org/jira/browse/TDIST-238?focusedCommentId=17117&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17117

sergey-shambir:
I'm not sure that bug will be fixed soon in Tizen and maybe we can use another approach to enable clang 3.1 and static STL library with proper c++11 support (including std::thread). So tizen projects will be compiled with " -nostdlibc++ -std=c++0x" and include paths / linking with custom STL.

STLport is common solution for such case, but AFAIK it have bad c++ 2011 support even in development branch. libstdc++ 4.6/4.7 used on android and can be helpful. But at this moment I'm playing with libc++ from LLVM project; I've managed to compile it for i386-based emulator (not ARM device) and still didn't tested.

I hope that tizen port can stay in separate branch until this issue will be resolved. Actually it have various issues like text mirroring on emulator, and AFAIK no accelerator support, so small separation from upstream is not bad.

pyrasis:
https://developer.tizen.org/contests/tizen-app-challenge

Tizen SDK 2.2 seems to will be released.
Let's try again using Tizen SDK 2.2

pyrasis:
Unfortunately Tizen 2.2 SDK doesn't support range-based for loop.

PR #2975, opened 2013-06-22
Refactored SimpleAudioEngine for Win32 using OpenAL

metalpasman:
Setting volume works now for background music and sound effects. 
Added OGG support using vorbis ogg

### DIFF
--- a/samples/Cpp/TestCpp/Classes/PerformanceTest/PerformanceNodeChildrenTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/PerformanceTest/PerformanceNodeChildrenTest.cpp
@@ -2,6 +2,14 @@
 
 #include <algorithm>
 
+#ifndef srandom
+#define srandom srand
+#endif
+
+#ifndef random
+#define random rand
+#endif
+
 // Enable profiles for this file
 #undef CC_PROFILER_DISPLAY_TIMERS
 #define CC_PROFILER_DISPLAY_TIMERS() Profiler::getInstance()->displayTimers()


sergey-shambir:
Does cocos2d-x build as static library on windows? OpenAL soft have LGPL license, so it puts restrictions on static linking with proprietary games. Either OpenAL or cocos2d-x should be distributed as dll in order to avoid problems.

I also have question to cocos2d-x team: how should work `SimpleAudioEngine::rewindBackgroundMusic()': should it shop music or not? It stops music on OpenAL implementations for Linux, Blackberry and Native Client, but does not stop on Linux+FMOD (and probably on Android/iOS, I didn't check yet). If it shouldn't, proper rewind implementation looks like this: http://paste.ubuntu.com/5791989/

Thanks for this change - it helps me to implement pitch/pan/gain support for Windows too, and I'm going to make pull request after you change will be merged. Now only marmalade, emscrippten and android+OpenES have no effects support (in my branch, not in develop).
- emscrippten can be moved to OpenAL later;
- OpenES have 3D effects support, I just didn't coded this yet.

metalpasman:
I have static linked for this only vorbis. Alut and OpenAL32 are dynamically linked so this shouldn't be a problem

dumganhar:
> Either OpenAL or cocos2d-x should be distributed as dll in order to avoid problems.

Hi @sergey-shambir 
I have already fixed a strange bug here (https://github.com/cocos2d/cocos2d-x/pull/2976).
Refactoring `libcocos2d` and `libCocosDenshion` project from dynamic library to static library.
It will be pity if `OpenAL` license has to make `libcocos2d` project use `dynamic library`

rtissera:
I can't see the problem having cocos2d and CocosDenshion as static library, with OpenAL as a separate DLL.
That's the same case than SDL which is LGPL, has in used in many commercial games.
As far as you don't modifiy OpenAL source code and the OpenAL32.dll is the stock one, you're good to go and just need to include a copy of LGPL license. Please correct me if I'm wrong.

dumganhar:
@rtissera 
Thanks for your update.
That's a good news.
After merging this PR, I'll add OpenAL license to `licenses` folder. ;)
@metalpasman 
Thank you for adding this cool feature.

rtissera:
Also please have a look at OpenAL Soft since OpenAL SI (Sample Implementation) is considered as deprecated.
More information here : http://en.wikipedia.org/wiki/OpenAL#Implementations

dumganhar:
There is no sound on `Parallels VM win8`, I used MBP 2013 15inch retina.

rtissera:
Looks like a known issue on Parallels :
http://forum.parallels.com/showthread.php?210337-OpenAL-sound-device-and-driver

Maybe use OpenAL by default and fallback to "old" implementation if device
cannot be opened ?

2013/6/26 James Chen notifications@github.com

dumganhar:
Thanks for provide this information.
There is another Pull Request https://github.com/cocos2d/cocos2d-x/pull/2999
It uses `DirectSound` to play effect and music, and also implements volume control.
How to check whether PC supports `OpenAL`?
What we need to do is:
1) Check whether supports `OpenAL`
2) If not supports, uses the solution at https://github.com/cocos2d/cocos2d-x/pull/2999

What do you think? Do you mind to achieve it?

dumganhar:
BTW, I will send you a PR for we have changed CocosDenshion project from dynamic library to static library.

metalpasman:
Maybe we should have all 3 implementations and either allow injection in SimpleAudioEngine and/or have a standard fall back system

rtissera:
Yes, I agree with Cristian.

We should do :
- Try to create an OpenAL device using alcOpenDevice. If NULL, switch back
  to DirectSound.
- Try to open DirectSound device, if NULL (since DirectSound is deprecated
  in Win8), fallback to current implementation (WaveOut/WINMM).

We could even remove DLL dependency at startup by getting OpenAL function
pointers from DLL dynamically using GetProcAdress().

What do you think guys ?

2013/6/26 Cristian Cristea notifications@github.com

metalpasman:
Ok, I will start implementing

dumganhar:
@rtissera 
Agree. That's great.

dumganhar:
@metalpasman 

I have sent you a PR(https://github.com/metalpasman/cocos2d-x/pull/1), please merge the PR before you start.

dumganhar:
@metalpasman
May I ask the state of this PR?
Do you need to merge this PR for OpenAL support now.

dumganhar:
Could OpenAL be supported for all windows computers except Parallels VM?
If so, we could just use OpenAL, it will be easy for maintaining.

metalpasman:
I've created the generic subsystem with only OpenAL working now (Soft OpenAL). I'm implementing directsound and Mci for fallback.
Once I switched to soft openal apparently the range of supported computers expanded since I also had one configuration on which it didn't work but it works now

dumganhar:
If we use _soft openal_ which can be run on most windows computers, are _direct sound_ and _Mci_ needed?
That will take us more cost of maintaining.
What's your opinion?

metalpasman:
I would go with openal and if there are reported issues we can quickly implement the directsound subsystem. I would have the fallback to mci though just to be sure. There is a branch that adds support for volume on some configurations which is a plus.

rtissera:
+1 for Cristian, soft openal is supported and should not give much issues.
Then we will see and fallback to legacy APIs if some issues arise.

2013/7/1 Cristian Cristea notifications@github.com

metalpasman:
I've committed the fallback to Mci but this has given rise to a new issue. There is no uniform support for file formats.
What I suggest is a method in SimpleAudioEngine that will retrieve a list of supported formats so the game can choose what sound files to load at runtime. 
An automated extension switch might also work, so when trying to load an OGG file on Mci it would try to open an MP3 or WAV

Other than that if it passes your tests it can be merged from my point of view. I've also included support for pitch, gain, pan to match Pitch 7platforms PR

sergey-shambir:
No need to care much about formats, because there also is DirectSound-based backed. So (in future, after this change will be merged) it can be used like OpenAL, with one difference: decoders keep PCM data and it's attirubutes separately in OpenAL, and compose PCM with WAV header into single WAV file in memory in DirectSound.

sergey-shambir:
About compile error: please, don't add new method to SimpleAudioEngine, just make changes under the hood:
`return m_Audio->playEffect(pszFilePath, bLoop, 1.0, 0.0, 1.0);`
So playEffect doesn't take new parameters, but anything _inside_ win32 sound engine does.

metalpasman:
@sergey-shambir fixed that

minggo:
A newbie question: not a method can work on all windows systems?
Why?

metalpasman:
Soft OpenAl should work, but I cannot test on all configurations. If other people can confirm it works I think it should be ok. Other than that you can't go wrong with a fallback system:)

minggo:
Ok, so i think we can just use Soft OpenAL now. And if other developers find it can't work, then we can fix it then.

What's your opinion?

rtissera:
Soft OpenAL should be fine & safe, yes.
We don't really need sound hardware acceleration for Win32 platform anyway.

2013/7/4 minggo notifications@github.com

metalpasman:
I agree

metalpasman:
Since Pitch 7platforms PR have been merged should I update this branch accordingly?

sergey-shambir:
Yes, that will be good (merge conflict should occur in win32 playEffect() delcaration/definition).

rtissera:
Agree, this would be an useful addition.

2013/12/4 lmskater notifications@github.com

PR #2977, opened 2013-06-23
Update Makefile for linux extensions

slobodin:


dumganhar:
Hi @flaming0 
Now, only win32, android and iOS support websocket.
If you want to make it works on Linux, please compile `libwebsockets` firstly.
Otherwise, adding `WebSocket` to makefile will make no sense.

slobodin:
Hi!

Yes, I've complied libwebsockets and installed it. But I got unresolved externals of WebSocket class members. So, I decided to include this rule to Makefile.

slobodin:
Am I understand right, that libwebsockets is not installed on build server?

dumganhar:
How did you install `libwebsockets`? By `apt-get install` ?
If so, you should modify `tools/travis-script/before-install.sh` for linux port.

slobodin:
Unfortunately, there is no libwebsockets package on my Ubuntu 13.04.
I built it from source.

dumganhar:
So I think we should also install `libwebsocket` from source in `tools/travis-script/before-install.sh`.
Could you try to do that?

slobodin:
Actually I'm not master of writing bash scripts :). But in order to install `libwebsockets` we need to create Makefile via cmake and install additional dependencies (on my machine this library required only `libssl`). And then simply run make && make install.

dumganhar:
OK, I'll try to make `WebSocket` works on Linux too.

slobodin:
Thank you!

dumganhar:
Closed this PR since we're using CMake for linux build now.

PR #2982, opened 2013-06-25
TestCpp for iOS uses libcocos2dx, libCocosDenshion, etc...

ricardoquesada:
...as dependencies, and other minor fixes

### DIFF
--- a/samples/Cpp/HelloCpp/proj.ios/HelloCpp.xcodeproj/project.pbxproj
+++ b/samples/Cpp/HelloCpp/proj.ios/HelloCpp.xcodeproj/project.pbxproj
@@ -290,10 +290,6 @@
 				);
 				INFOPLIST_FILE = "HelloCpp-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 5.0;
-				LIBRARY_SEARCH_PATHS = (
-					"$(inherited)",
-					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
-				);
 				ONLY_ACTIVE_ARCH = YES;
 				PRODUCT_NAME = HelloCpp;
 				PROVISIONING_PROFILE = "";
@@ -332,10 +328,6 @@
 				);
 				INFOPLIST_FILE = "HelloCpp-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 5.0;
-				LIBRARY_SEARCH_PATHS = (
-					"$(inherited)",
-					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
-				);
 				ONLY_ACTIVE_ARCH = YES;
 				PRODUCT_NAME = HelloCpp;
 				PROVISIONING_PROFILE = "";
--- a/samples/Cpp/SimpleGame/proj.ios/SimpleGame.xcodeproj/project.pbxproj
+++ b/samples/Cpp/SimpleGame/proj.ios/SimpleGame.xcodeproj/project.pbxproj
@@ -299,10 +299,6 @@
 				);
 				INFOPLIST_FILE = "SimpleGame-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 5.0;
-				LIBRARY_SEARCH_PATHS = (
-					"$(inherited)",
-					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
-				);
 				ONLY_ACTIVE_ARCH = YES;
 				PRODUCT_NAME = SimpleGame;
 				PROVISIONING_PROFILE = "";
@@ -340,10 +336,6 @@
 				);
 				INFOPLIST_FILE = "SimpleGame-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 5.0;
-				LIBRARY_SEARCH_PATHS = (
-					"$(inherited)",
-					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
-				);
 				ONLY_ACTIVE_ARCH = YES;
 				PRODUCT_NAME = SimpleGame;
 				PROVISIONING_PROFILE = "";
--- a/samples/Cpp/TestCpp/proj.ios/TestCpp.xcodeproj/project.pbxproj
+++ b/samples/Cpp/TestCpp/proj.ios/TestCpp.xcodeproj/project.pbxproj
@@ -1774,7 +1774,6 @@
 				LIBRARY_SEARCH_PATHS = (
 					"\"$(SRCROOT)/../../../../cocos2dx/platform/third_party/ios/libraries\"",
 					"\"$(SRCROOT)/../../../../external/libwebsockets/ios/lib\"",
-					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
 				);
 				ONLY_ACTIVE_ARCH = YES;
 				OTHER_CFLAGS = "-O2";
@@ -1820,7 +1819,6 @@
 				LIBRARY_SEARCH_PATHS = (
 					"\"$(SRCROOT)/../../../../cocos2dx/platform/third_party/ios/libraries\"",
 					"\"$(SRCROOT)/../../../../external/libwebsockets/ios/lib\"",
-					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
 				);
 				ONLY_ACTIVE_ARCH = YES;
 				PRODUCT_NAME = TestCpp;


samples/Cpp/AssetsManagerTest/proj.ios/AssetsManagerTest.xcodeproj/project.pbxproj
```diff
@@ -169,13 +169,13 @@
...
+				A0F3A9B5177905F600CF0EB6 /* libCocosDenshion.a */,
+				A0F3A9B6177905F600CF0EB6 /* libextensions.a */,
+				A0F3A9B7177905F600CF0EB6 /* libjsbindings.a */,
 				1A6B05CA17782B8700FFE52B /* libsqlite3.dylib */,
 				1A6B05C817782B8100FFE52B /* libz.dylib */,
-				1A6B0595177828E700FFE52B /* libchipmunk.a */,
-				1A6B0596177828E700FFE52B /* libcocos2dx.a */,
-				1A6B0597177828E700FFE52B /* libCocosDenshion.a */,
-				1A6B0598177828E700FFE52B /* libextensions.a */,
-				1A6B0599177828E700FFE52B /* libjsbindings.a */,
```
dumganhar, line 178: Any different? What errors will it produce?

samples/Cpp/AssetsManagerTest/proj.ios/AssetsManagerTest.xcodeproj/project.pbxproj
```diff
@@ -424,6 +424,7 @@
 					"\"$(SRCROOT)/../../../../CocosDenshion/third_party/blackberry/libraries/armle-v7\"",
 					"\"$(SRCROOT)/../../../../CocosDenshion/third_party/blackberry/libraries/x86\"",
 					"\"$(SRCROOT)/../../../../scripting/javascript/spidermonkey-ios/lib\"",
+					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
```
dumganhar: Why should add this line?
Is `DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht` the same on other Mac?
I tested all projects on iphone-simulator and iPad 3, they worked ok. I used xcode 4.6.2 on Moutain lion 10.8.4.

samples/Cpp/HelloCpp/proj.ios/HelloCpp.xcodeproj/project.pbxproj
```diff
@@ -109,7 +121,11 @@
 		29B97323FDCFA39411CA2CEA /* Frameworks */ = {
 			isa = PBXGroup;
 			children = (
-				1A0C0A6A1777F65600838530 /* libcocos2dx.a */,
+				A0F3A9BF1779069A00CF0EB6 /* libBox2D.a */,
+				A0F3A9C01779069A00CF0EB6 /* libchipmunk.a */,
+				A0F3A9C11779069A00CF0EB6 /* libCocosDenshion.a */,
+				A0F3A9C21779069A00CF0EB6 /* libextensions.a */,
+				A0F3A9BD1779062B00CF0EB6 /* libcocos2dx.a */,
```
dumganhar, line 120: Why should `HelloCpp` project depend on `libBox2D`, `libchipmunk.a`,... etc.

samples/Cpp/AssetsManagerTest/proj.ios/AssetsManagerTest.xcodeproj/project.pbxproj
ricardoquesada, line 178: What I did, was to move those dependencies in the "Framework" folder.

samples/Cpp/AssetsManagerTest/proj.ios/AssetsManagerTest.xcodeproj/project.pbxproj
ricardoquesada: I really don't know :-(
I think Xcode added that line automatically.

samples/Cpp/HelloCpp/proj.ios/HelloCpp.xcodeproj/project.pbxproj
ricardoquesada, line 120: Good point. I can remove it.

ricardoquesada, line 120: fixed.

ricardoquesada:
IMPORTANT:
TestCpp depends on the cocos2dx, chipmunk, ... projects.

So, users should not open the TestCpp project directly, because it will fail, it is not self contained.
Also, the daily builder will fail. Instead of trying to build TestCpp.xcodeproj, it build it from the iOS xcworkspace.

One solution, is to do it like in cocos2d-iphone:
- One big .xcodeproj with all the base libraries: cocos2d, box2d, chipmunk, etc..
- Another big .xcodeproj with all the tests and samples... and this one depends on the previous .xcodeproj.

I didn't want to make such a change without asking, but I think that is the correct way to prevent the above mentioned bug.

dumganhar:
> So, users should not open the TestCpp project directly, because it will fail, it is not self contained.

We could write a README file or WIKI page to describe this changes.

> Also, the daily builder will fail. Instead of trying to build TestCpp.xcodeproj, it build it from the iOS xcworkspace.

Yes, the only entry is `workspace` now. `Daily builder` need to be changed.
But what do you mean `Daily Builder` ? Do you mean our `ci.cocos2d-x.org` or `travis-ci` ?
If yes, I think we just need to change the configuration on `ci.cocos2d-x.org`, `travis-ci` doesn't support mac now.
BTW, now `ci.cocos2d-x.org` is always offline. We only use `Travis-CI` now.

dumganhar:
> One big .xcodeproj with all the base libraries: cocos2d, box2d, chipmunk, etc..

That will be one big .xcodeproj like TestCpp, depends on cocos2d, box2d, chipmunk, etc,
And other sample projects also have this dependence, in this way, we will compile cocos2d, box2d, chipmunk... every time like before. That's really waste compilation time, since workspace will not do that. It just build cocos2d, box2d... one time for all projects in workspace.

I used `workspace` is inspired by `GamePlay` project.

dumganhar:
Could the solution you said achieve this?

0fd3b4b3-1b9e-428f-ac63-0f773e7856de

ricardoquesada:
Yes, I meant `travis-ci`.

I think the current solution is better than the old one. I am going to use it all the time, and I think all iOS / Mac users are going to use it as well.

Yes, the solution that I proposed could generate those files. If you want to try it, just open the workspace that is included in cocos2d-iphone.

But migrating to the solution that I propose will "break" the current "standard", since the "proj.ios" and "proj.mac" directories won't be needed anymore.
I think if we update travis-ci, so that is builds the TestCpp" from the workspace file, we are fine. Could you do that ?  Thanks!

dumganhar:
> I think if we update travis-ci, so that is builds the TestCpp" from the workspace file, we are fine. Could you do that ? 

It's a pity that `travis-ci` could not build a task(several jobs in it) with different jobs that some jobs run on `Linux` while some jobs run on `Mac`. 

Now, we're only active `Linux` travis-ci build. I think `Travis-ci` will support this feature soon.
If not, I also have a idea to make it supports both linux and mac, but that will be ugly. :(
About ci changes, I will let you know if it has any informations. :)

dumganhar:
Why I only active Linux for `travis-ci` is to that, we and you guys are all using `MBP`, so the errors will be known very soon. And we could correct it soon.

samples/Cpp/HelloCpp/proj.ios/HelloCpp.xcodeproj/project.pbxproj
```diff
@@ -282,6 +290,10 @@
 				);
 				INFOPLIST_FILE = "HelloCpp-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 5.0;
+				LIBRARY_SEARCH_PATHS = (
+					"$(inherited)",
+					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
```
dumganhar: We need to remove this line as I make sure that `cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht` is different on different MACs.

samples/Cpp/SimpleGame/proj.ios/SimpleGame.xcodeproj/project.pbxproj
```diff
@@ -296,7 +299,10 @@
 				);
 				INFOPLIST_FILE = "SimpleGame-Info.plist";
 				IPHONEOS_DEPLOYMENT_TARGET = 5.0;
-				LIBRARY_SEARCH_PATHS = "$(inherited)";
+				LIBRARY_SEARCH_PATHS = (
+					"$(inherited)",
+					"\"$(SRCROOT)/../../../../../../Library/Developer/Xcode/DerivedData/cocos2d-ios-efgwnjcjlugbweaakhyrdehqseht/Build/Products/Debug-iphoneos\"",
```
dumganhar: If you don't have time to do that, I will be happy to help you.

samples/Cpp/AssetsManagerTest/proj.ios/AssetsManagerTest.xcodeproj/project.pbxproj
dumganhar, line 178: The dependencies was in `Framework` folder, I don't know why it's wrong on your MBP.
Anyway, I will do a test on my colleague‘s MBP. :)

dumganhar, line 178: I have checked this issue on another MBP, but it was in `Frameworks` folder.
Could you checkout a clean branch that synchronizing the latest develop branch of cocos2d-x repo and try again?
BTW, what's your system version, Xcode version?

samples/Cpp/SimpleGame/proj.ios/SimpleGame.xcodeproj/project.pbxproj
ricardoquesada: argh... f****ing Xcode... it was added automatically by Xcode.

ricardoquesada:
fixed.

dumganhar:
Merged. Thanks. :)

PR #2983, opened 2013-06-25
Improve Android file system read performance

folecr:
Faster file system read performance on Android!
- Raw filesystem read is approximately 40% faster
- Game startup is approximately 12% faster
- Uses Native AssetManager
- ... Instead of custom unzip code
- Also, because the code now does not rely on internal data structures used in `zip_support/ZipUtils.*`, it should be re-entrant.

Tests
- Tested with https://github.com/folecr/SlideShowFSTest
- Test results are average of 3 runs
- Tested on Galaxy Nexus with Android 4.2.2

Test 1 https://github.com/folecr/SlideShowFSTest/tree/master/samples/Cpp/RawFSTest
- Reads image data using CCFileUtils
- Time taken to read data
  - Before : 297.5 millis
  - After : 125 millis
- Approximately 40% faster

Test 2 https://github.com/folecr/SlideShowFSTest/tree/master/samples/Cpp/SlideShowFSTest
- Creates and displays sprites from image data
- Approximately 12% faster to display all sprites

### DIFF
--- a/cocos2dx/platform/android/CCFileUtilsAndroid.cpp
+++ b/cocos2dx/platform/android/CCFileUtilsAndroid.cpp
@@ -162,11 +162,9 @@ unsigned char* FileUtilsAndroid::doGetFileData(const char* pszFileName, const ch
     {
         if (forAsync)
         {
-            // pData = s_pZipFile->getFileData(fullPath.c_str(), pSize, s_pZipFile->_dataThread);
-            abort();
+            // Nothing to do when using the Android assetmanager APIs???
         }
-        else
-        {
+
         string fullPath = fullPathForFilename(pszFileName);
 
         const char* relativepath = fullPath.c_str();
@@ -200,7 +198,6 @@ unsigned char* FileUtilsAndroid::doGetFileData(const char* pszFileName, const ch
         }
 
         AAsset_close(asset);
-        }
     }
     else
     {


dumganhar:
``` sh
/home/travis/bin/android-ndk/toolchains/arm-linux-androideabi-4.7/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.7/../../../../arm-linux-androideabi/bin/ld: error: cannot find -landroid
```

Why only `TestJavascript` can not find -landroid? Anything else need to be modified for `TestJavascript`?

cocos2dx/platform/android/CCFileUtilsAndroid.cpp
```diff
@@ -22,16 +22,37 @@ of this software and associated documentation files (the "Software"), to deal
...
+
+#include <stdlib.h>
+
+#define  LOG_TAG    "CCFileUtilsAndroid.cpp"
+#define  LOGD(...)  __android_log_print(ANDROID_LOG_DEBUG,LOG_TAG,__VA_ARGS__)
 
 using namespace std;
 
-NS_CC_BEGIN
+static AAssetManager* s_assetmanager;
```
minggo, line 37: It is better to initialize it.

cocos2dx/platform/android/CCFileUtilsAndroid.cpp
```diff
@@ -72,16 +90,26 @@ bool FileUtilsAndroid::isFileExist(const std::string& strFilePath)
...
-        if (s_pZipFile->fileExists(strPath))
+        if (strFilePath.find(_defaultResRootPath) != 0)
         {
-            bFound = true;
-        } 
+            // Didn't find "assets/" at the beginning of the path
+        } else {
+            // Found "assets/" at the beginning of the path and we don't want it
+            s += strlen("assets/");
+        }
```
minggo, line 101: Do nothing in `if` sentence. 
Why not change to 

``` c++
if (strFilePath.find(_defaultResRootPath) == 0)
{
      // Found "assets/" at the beginning of the path and we don't want it
      s += strlen("assets/");
}
```

minggo:
Dumdanhar said

> > Why only TestJavascript can not find -landroid? Anything else need to be modified for TestJavascript?

It is because doesn't change AssetsManagerTest android project setting.

minggo:
@folecr 
I fixed compiling error in #2990.
And i will modify codes as you said in email in another pull request.
I will let you know when i send the pull request.

folecr:
Thanks!

minggo:
@folecr 
I send #3008 to modify some codes because of the usage of asset manager.
Would you please review it. 
Thanks.

PR #2985, opened 2013-06-25
Adds mac workspace

ricardoquesada:
for the moment it includes TestCpp.
But Chipmunk, Box2d, cocos2dx, extensions and CocosDenshion are libraries
Like in the iOS project


dumganhar:
I don't know why we ignored `*.xcworkspacedata` files in `.gitignore`. -_-!!!
Therefore, one thing has to be done is either to remove it from `.gitignore` or `git add -f cocos2d-mac.xcworkspacedata/contents.xcworkspacedata`.

ricardoquesada:
fixed.

dumganhar:
Merged. :) So would you add other projects to mac workspace?

ricardoquesada:
yes, I will add more projects to it.

PR #2986, opened 2013-06-25
Cleanup travis build scripts.

sbc100:
before-install.sh and run-script.sh now both given an error if
no PLATFORM was not specified correctly.

Environment variables are now set as part of run-script.sh.  This
makes the travis output simpler and makes it possible to run
run-script.sh outside of travis.

javascript binding generation will not exit early if GH_USER
etc are not set.  This allows the travis scripts to be run in
cloned repos which were otherwise erroring out trying to go
the JS bindings commit.

When generating javascript bindgings, don't username and email
with '--global'.  It should be sufficient to set this only for
the binding repo.

### DIFF
--- a/tools/travis-scripts/generate-jsbindings.sh
+++ b/tools/travis-scripts/generate-jsbindings.sh
@@ -49,14 +49,13 @@ if [ "$GEN_JSB"x != "YES"x ]; then
     exit 0
 fi
 
-#Set git user
-git config --global user.email ${GH_EMAIL}
-git config --global user.name ${GH_USER}
-
 # Update submodule of auto-gen JSBinding repo.
 pushd "$GENERATED_WORKTREE"
 
 git checkout -B master
+#Set git user
+git config user.email ${GH_EMAIL}
+git config user.name ${GH_USER}
 #Set remotes
 git remote add upstream https://${GH_USER}:${GH_PASSWORD}@github.com/folecr/cocos2dx-autogen-bindings.git 2> /dev/null > /dev/null
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
---- 
 language: cpp
-env: 
-  matrix: 
+env:
+  matrix:
   - GEN_JSB=YES
   - PLATFORM=linux DEBUG=1
   # Since switching to C++11 only the ARM version of the nactive client
@@ -10,7 +9,7 @@ env:
   # - PLATFORM=nacl DEBUG=1 NACL_ARCH=arm
   - PLATFORM=android
   - PLATFORM=emscripten DEBUG=1
-  global: 
+  global:
   - secure: |-
       cHXOjAda+t2tmBVuVAiQfvhDN+unToO6hX1rc3dJeGDOk4rlQnylJacjA7o/
       enkWeHagbW2Ja4h7BxQz4h7dada8zEbIbEMK4VEHx6AWwQM+GIEqkJPv5g8k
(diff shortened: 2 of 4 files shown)


sbc100:
As you can see from the travis results, building from a non-official repo now works.

Also, moving the env setting into the scripts makes the scripts themselves useable locally as well as making the output from travis more readable.

dumganhar:
Thanks for these changes, I'll check that soon. :)
BTW, we disable native client on travis-ci build since it doesn't support `std::thread`, `std::mutex`, etc which are c++11 thread APIs supported from gcc 4.7.

tools/travis-scripts/generate-jsbindings.sh
```diff
@@ -49,16 +49,15 @@ if [ "$GEN_JSB"x != "YES"x ]; then
...
 # Update submodule of auto-gen JSBinding repo.
 pushd "$GENERATED_WORKTREE"
 
 git checkout -B master
+#Set git user
+git config user.email ${GH_EMAIL}
+git config user.name ${GH_USER}
 #Set remotes
-git remote add upstream https://${GH_USER}:${GH_PASSWORD}@github.com/folecr/cocos2dx-autogen-bindings.git 2> /dev/null > /dev/null
+git remote add upstream https://${GH_USER}:${GH_PASSWORD}@github.com/folecr/cocos2dx-autogen-bindings.git > /dev/null
```
dumganhar: Why to delete `2> /dev/null` ? If so, the security code will be printed, right?

dumganhar: Except this, everything looks perfect. Thanks.

sbc100: Oh, I didn't realize why it was redirected.

I had the problem that this command was failing, which was resulting in a silent failure (for no apparent reason).  

I just did some investigation and it looks like 'git remote add' doesn't output anythign to the stderr or stdout in normal circumstances.  

In any case my problem was fixed by check for GH_USER and GH_PASSWORD prior to even calling this script.

sbc100: Want me to send new PR with this line reverted?

dumganhar: Just update this PR is enough. Don't send a new PR.

dumganhar: Or by your convenience. :)

sbc100:
Done.

dumganhar:
Thanks. :)

sbc100:
Regarding C++11.  For ARM we ship gcc-4.8 so I can probably make all the C++11 stuff work.

For x86 and x86_64 we are stuck with GCC 4.4.3, and will be for some time I'm afraid.

Has there been any public discussions about which C++11 features to use?  For the sake of compatibility (which I understand is very important for the cocos2dx, right?) I imagine it makes sense to limit the list C++11 features that are essential to building cococs.  To be honest I would have advised against the idea altogether but I guess the decision has been made.   I fear that the more C++11 we use the less likely we can port to random new platforms that lag in such things.

For example our x86 compilers (4.4.6) have std::function in the std library but don't support the lambda syntax.

dumganhar:
@sbc100 

You could join our discussion group here(https://groups.google.com/forum/#!topic/cocos2d-js-devel/vWe_F3LN4Zo)
Just send the request and write down who you are, @ricardoquesada will add you to the group. Thanks.

> I fear that the more C++11 we use the less likely we can port to random new platforms that lag in such things.

You're right, but we should not wait all compilers support most c++11 features. That will delay the development of cocos2d-x. If there is a platform doesn't support the c++11 feature which we're using, we will probably stop the support for that port temporarily. And we will advise user who want to make their games run on these platforms that don't support c++11 well to use cocos2d-x2.x. When these platforms support it again, we will be happy to continue to maintain it.
Anyway, to be honest, the most popular platforms in cocos2d-x are Android, iOS, Win32(for development). Hope what I said didn't hurt anybody. :)

PR #2995, opened 2013-06-25
Fix bug in CCLabelBMFont::m_bIsOpacityModifyRGB when changing font file

adiumcommodore:
When the font a CCLabelBMFont uses is changed via setFntFile the m_bIsOpacityModifyRGB property is not updated to reflect whether the new texture uses premultiplied alpha. This can cause incorrect rendering if the label was created without a font file originally, such as through a call to CCLabelBMFont::create().

The fix updates the property through its mutator function to ensure any glyph sprites already created are updated.


minggo:
@adiumcommodore 
Could you please add a test for it?

PR #2999, opened 2013-06-25
[win32]MciPlayer replaced with DirectSound

ololomax:
Added sound and music volume control for win32


dumganhar:
What do you think of OpenAL support? https://github.com/cocos2d/cocos2d-x/pull/2975

I think OpenAL support will be better, what's your opinion?

sergey-shambir:
Please use utf8 encoding for all files and don't use non-english comments.
- Does it support mp3 and ogg? Seems that only wav is supported.
- SimpleAudioEngine::playEffect() will be extended soon with pitch (playback rate), pan (stereo effect) and gain (single effect volume) parameters. Can you implement these parameters for Direct Sound backend later?

ololomax:
@dumganhar I'm not really sure what's better, I've chosen DirectSound because it is simplier (in my opinion)

@sergey-shambir 
- wav only
- I can try to do this later

sergey-shambir:
Cocos2d-x have two first-class platforms where all features should work OK in release versions: iOS and android. Many games based on cocos2dx released for this platforms, and many games still under development or maintenance.

Win32 platform often used for development (because development directly on android is hurt), and less for deployment. Also windows developers usually have no experience in writing portable code and probably will not use recommended format if it's complicates development.

Cocos2d-x supports ogg/vorbis on most platforms, it is recommended format. So win32 platform needs support ogg, otherwise Visual Studio users discard recommendations. OpenAL backed for windows supports ogg, but doesn't work when windows runs in Parallels VM on Mac OS X.

In other hand, ogg can be decoded to PCM. Maybe DirectSound can play PCM buffers?

ololomax:
@sergey-shambir I think it can, aren't WAV and PCM is the same formats?

dumganhar:
@ololomax 
WAV contains wav-header and PCM content.

cocosQA:
Dear ololomax:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6493

PR #3008, opened 2013-06-26
remove unneeded codes when using assets manager

minggo:
- revert ZipUtils
- fix a potencial bug for multi-threading
- put Image::initWithImageFileThreadSafe into private field because it is used only for addImageAsync


folecr:
Looks good to me. Thanks for the changes!

(Unrelated question : In future do we plan to make autorelease threadsafe?)

minggo:
@folecr 
Ricardo created a thread discussing whether to use share_pointer or retain/release [here](http://www.cocos2d-x.org/boards/6/topics/30253).

PR #3014, opened 2013-06-27
fix sub ccb node resource root path bug

wtyqm:
When load sub ccb type node , create a new ccbreader from parent
ccbreader. Should copy ccbRootPath from parent, otherwise, load
resource will fail if sub ccb node resource use special path


dumganhar:
Thanks.

PR #3019, opened 2013-06-27
removed libxml2 dependence from libcocos2d

Squallium:
Using eclipse in linux I realize that the libxml2 library is no longer exists, so I removed the dependency. I hope it helps!


minggo:
Thanks @Squallium.

PR #3033, opened 2013-06-29
Improvement cccontroll button

gelldur:
Litle fix in CCControllUtils in CCRectUnion. Also i changed CCScale9Sprite to CCNode now you can set fixed sprite as a button.

It should be working with old CCScale9Sprite (i have tested it) but now TestCPP is corrupted (without my changes too), but when i tested it before pull from master everything works great.

Thanks to that change you can pass normal CCSprite as a button and it shouldn't stretch.

For example

CCControlButton\* pButton3 = CCControlButton::create (
                            CCSprite::create ( "test/hw2.png" ) );
pButton3->setPosition ( 200, 200 );
addChild ( pButton3 );

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ExtensionsTest/ControlExtensionTest/CCControlSceneManager.cpp
+++ b/samples/Cpp/TestCpp/Classes/ExtensionsTest/ControlExtensionTest/CCControlSceneManager.cpp
@@ -110,7 +110,7 @@ Scene *ControlSceneManager::currentControlScene()
     case kControlButtonTest_HelloVariableSize:return ControlButtonTest_HelloVariableSize::sceneWithTitle(s_testArray[_currentControlSceneId]);
     case kControlButtonTest_Event:return ControlButtonTest_Event::sceneWithTitle(s_testArray[_currentControlSceneId]);
     case kControlButtonTest_Styling:return ControlButtonTest_Styling::sceneWithTitle(s_testArray[_currentControlSceneId]);
-    case ControlButtonTest_SpriteBackground:return ControlButtonTest_SpriteBackground::sceneWithTitle(s_testArray[_currentControlSceneId]);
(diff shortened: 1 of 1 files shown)


dumganhar:
@gelldur 
Thanks for your contribution.
Do you mind to add a new test in `TestCpp/ExtensionsTest/CCControlButtonTest`?

gelldur:
Ok

gelldur:
OK i added some test but i can;t test it ;/ a lot of compile errors on linux so sorry if it don't work ;/

I could do it better but now i don;t have time to fix compile errors (because it can't find such class like Sprite)

extensions/GUI/CCControlExtension/CCControlUtils.cpp
```diff
@@ -164,7 +164,7 @@ Rect ControlUtils::RectUnion(const Rect& src1, const Rect& src2)
     float x2 = MAX(src1.getMaxX(), src2.getMaxX());
     float y2 = MAX(src1.getMaxY(), src2.getMaxY());
     
-    result.origin=Point(x1,x2);
+    result.origin=Point(x1,y1);
```
gelldur, line 167: I think here should be y1 am i right?

samples/Cpp/TestCpp/Classes/ExtensionsTest/ControlExtensionTest/CCControlButtonTest/CCControlButtonTest.cpp
```diff
@@ -275,3 +275,30 @@ ControlButton *ControlButtonTest_Styling::standardButtonWithTitle(const char *ti
     return button;
 }
 
+//ControlButtonTest_SpriteBackground
```
gelldur, line 278: I think i add it good but can't test :(

PR #3036, opened 2013-07-01
Update XMLHTTPRequest.cpp

zj831007:
When network connect failed, still invoke the javascript callback method: onreadystateCallback


dumganhar:
Why're the indentions different? You probably need to use _4 spaces_ instead of _tab_.

zj831007:
I close this PR  for my mistake , and I don't know how to open it again now. Can it will be auto merged?

PR #3040, opened 2013-07-01
SSL cert verification error was always ignored.

hokuts:
I think settings below must be optional to protect agatinst man-in-the-middle attack.

```
curl_easy_setopt(handle, CURLOPT_SSL_VERIFYPEER, 0L);
curl_easy_setopt(handle, CURLOPT_SSL_VERIFYHOST, 0L);
```

### DIFF
--- a/extensions/network/HttpClient.cpp
+++ b/extensions/network/HttpClient.cpp
@@ -242,12 +242,19 @@ static bool configureCURL(CURL *handle)
     }
     
     if(HttpClient::getInstance()->isIgnoreSslVerification()){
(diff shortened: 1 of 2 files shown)


dumganhar:
So if enabling SSL cert verification, do we need to set `curl_easy_setopt(curl, CURLOPT_CAINFO, "_path_\ca-bundle.crt");` in HttpRequest or developer set it outside HttpRequest?

hokuts:
I think it depends on build config of libcurl (and maybe client environment) whether we NEED or not.
It is a nice idea to enable developers to set certificates path.
I'll add a commit. Thanks.

dumganhar:
OK, thank you.

PR #3041, opened 2013-07-01
Clear error message buffer before request.

hokuts:
In the below case,
1. 1st request. It was finished with error message.
2. 2nd reqeust. It has no error message.

after 2nd request, this buffer sill had 1st error.

### DIFF
--- a/extensions/network/HttpClient.cpp
+++ b/extensions/network/HttpClient.cpp
@@ -135,7 +135,7 @@ static void networkThread(void)
         
         int32_t responseCode = -1;
         int retValue = 0;
-        s_errorBuffer[0] = 0;
+        memset(s_errorBuffer , 0, sizeof(s_errorBuffer));
 
         // Process the request -> get response packet
         switch (request->getRequestType())


extensions/network/HttpClient.cpp
```diff
@@ -135,6 +135,7 @@ static void networkThread(void)
         
         int32_t responseCode = -1;
         int retValue = 0;
+        s_errorBuffer[0] = 0;
```
dumganhar: Why did you just set the content of first index to 0?
Is `memset(s_errorBuffer , 0, sizeof(s_errorBuffer ));` more suitable?

hokuts: I did because I assume a message string in the buffer is zero-terminated.
But you are right, `memset` to fill with zeroes is better.
I will make another commit. Thanks.

cocosQA:
Dear hokuts:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6491

PR #3042, opened 2013-07-01
Update XMLHTTPRequest.cpp

zj831007:
When network connect failed, still invoke the javascript callback method: onreadystateCallback, change some tab to 4 spaces with #3036

### DIFF
--- a/scripting/javascript/bindings/XMLHTTPRequest.cpp
+++ b/scripting/javascript/bindings/XMLHTTPRequest.cpp
@@ -180,46 +180,50 @@ void MinXmlHttpRequest::handle_requestResponse(cocos2d::extension::HttpClient *s
     
     if (!response->isSucceed())
     {
+        status = statusCode;
         CCLOG("response failed");
         CCLOG("error buffer: %s", response->getErrorBuffer());
-        return;
-    }
-    
-    // set header
-    std::vector<char> *headers = response->getResponseHeader();
-    
-    char* concatHeader = (char*) malloc(headers->size() + 1);
-    std::string header(headers->begin(), headers->end());
-    strcpy(concatHeader, header.c_str());
-    
-    std::istringstream stream(concatHeader);
-    std::string line;
-    while(std::getline(stream, line)) {
-        _gotHeader(line);
-    }
-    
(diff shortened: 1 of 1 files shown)


scripting/javascript/bindings/XMLHTTPRequest.cpp
```diff
@@ -180,46 +180,50 @@ void MinXmlHttpRequest::handle_requestResponse(cocos2d::extension::HttpClient *s
     
     if (!response->isSucceed())
     {
+        status = statusCode;
         CCLOG("response failed");
         CCLOG("error buffer: %s", response->getErrorBuffer());
-        return;
```
dumganhar, line 185: In your changes, but when an error is triggered, do we need to invoke `onreadystateCallback` in JS?

zj831007, line 185: Yes, we will pop error message to user when connect is failed or connect is timeout, so change the code like this, for example in game "MT", when the network is very slowly, and then the game pop a message "网络不给力啊"。In the line 232, the method  "onreadystateCallback" had been triggered.

dumganhar, line 185: But why did you use the same callback `onreadystateCallback` for this error?

zj831007, line 185: at line 183,  there has some code to set status value:   status = statusCode; so use the same callback in native code, and js client can show different err message by the status value. Do you have a better way to solve this problem？

dumganhar, line 185: How is the error dealt with on Html5? We should keep the same API as Html5.

zj831007, line 185: On html5,  we dealt as the same as this,   the http status is show there: http://httpstatus.es/
when stauts is 599, the network is connected timeout. may be should change some return value on httpclient to return timeout status,  because current only return -1 when timeout.

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/4807 . Thanks.

PR #3043, opened 2013-07-01
solve android crash with CCLabelTTF setString function

jxgzxh111:


cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxBitmap.java
```diff
@@ -390,8 +390,11 @@ private static int computeY(final FontMetricsInt pFontMetricsInt,
...
 
 				/* Remove spaces at the beginning of a new line. */
-				while (pString.charAt(i) == ' ') {
-					++i;
+				if(i < charLength)
+				{
+					while (pString.charAt(i) == ' ') {
+						++i;
+					}
 				}
```
minggo, line 398: I think it should changed to

``` c++
while (i < charLength && pString.charAt[i] == ' ') {
    ++i;
}
```

PR #3045, opened 2013-07-01
Unified mac project

ricardoquesada:
includes 2 xcodeprojects:
- one for all the libraries
- one for all the samples

It is easier to maintain, and easier for user to include in their projects.
The samples project includes the "libraries".
No need to use a workspace. 

Also, includes CocosDragon, MoonWarriors, CrystalCraze and Watermelon With Me (because it is super easy to add new projects now)


dumganhar:
So do you think that iOS projects also need to be changed like this?

dumganhar:
Found an issue in TestCpp:

```
2013-07-02 10:20:17.646 Test cpp[23417:303] Unable to load nib file: MainMenu, exiting
```

dumganhar:
> 2013-07-02 10:20:17.646 Test cpp[23417:303] Unable to load nib file: MainMenu, exiting

I fixed it.

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3046.

ricardoquesada:
@dumganhar 
Yes, I would do the same changes for iOS.
Game developers could just use the new .xcodeproject for their games. It is just one .xcodeproj for all the cocos2d related libs. And also is super easy to test all the samples, since all the samples are in just one project.
And easier to add samples as well. Just "duplicate" the target.

ricardoquesada:
thanks for fixing the compile issue.

dumganhar:
> Game developers could just use the new .xcodeproject for their games. It is just one .xcodeproj for all the cocos2d related libs. And also is super easy to test all the samples, since all the samples are in just one project.
> And easier to add samples as well. Just "duplicate" the target.

Yep, that seems to be easier for developer to add new projects by what you said "duplicate". The old way uses "workspace" can't achieve that.
Thanks.

.gitignore
```diff
@@ -59,7 +59,6 @@ build_*_vc10/
 *.pbxuser
 *.xcbkptlist
 *.xcscheme
-*.xcworkspacedata
```
dumganhar, line 62: If we will not use "workspace", I think we should ignore .xcworkspacedata file.
Any reason you delete it from .gitignore?

ricardoquesada, line 62: I needed to add that line in order to remove the old workspace.

PR #3052, opened 2013-07-02
Pivotjoint constructor and pointQueryFirst in Space class will return false instead of invalid shape object

smshuja:
Previously I had made the same changes on functions, now the same changes are made to classes. Previous pull requests for reference https://github.com/cocos2d/cocos2d-x/pull/2590 https://github.com/cocos2d/cocos2d-x/pull/2584


dumganhar:
Thanks.

PR #3056, opened 2013-07-02
Add ccDrawSolidCircle

BodbDearg:
Similar to ccDrawCircle() primitives drawing function except the circle is filled.


BodbDearg:
Thanks minggo!

quique123:
Hi, I ned help upgrading the ccDrawSolidCircle method i have that used glVertexPointer to one that now uses, i suppose, glVertexAttribPointer for iOS.  Thx in advance

Issue #2283, opened 2013-03-31
CCSet::removeAllObjects gets wrong logic


cooolinx:
I'm new to cocos2d-x and C++, but I found my code get an unexpected result:

``` cpp
CCSet *set = CCSet::create();
set->addObject(ccs("one"));
CCLOG("Before removeAllObjects: %d", set->count());  // 1
set->removeAllObjects();
CCLOG("After removeAllObjects: %d", set->count());  // Still 1
```

I saw the source code of CCSet::removeAllObjects(), it seems that there's no erase operation on the inner std::set, so I changed the code and it works great for me:

``` cpp
    CCSetIterator it;
    for (it = m_pSet->begin(); it != m_pSet->end(); )  // removed '++it'
    {
        if (! (*it))
        {
            break;
        }

        (*it)->release();
        m_pSet->erase(it++);  // erase iterator on std::set
    }
```

BTW. My version is cocos2d-2.1rc0-x-2.1.2

PR #3057, opened 2013-07-02
Bugfixes to the Set class.

BodbDearg:
1: If adding an object to the container then only retain the object again provided it's not already been added. The underlying std::set does not support duplicate entries. (see: www.cocos2d-x.org/boards/6/topics/10876)
2: When removing an object only CC_SAFE_RELEASE it if it was actually contained in the set. (see: http://www.cocos2d-x.org/boards/6/topics/10876)
3: Fix a bug in 'removeAllObjects' where the underlying std::set container was not being cleared. (see: https://github.com/cocos2d/cocos2d-x/issues/2283)

### DIFF
--- a/cocos2dx/cocoa/CCSet.cpp
+++ b/cocos2dx/cocoa/CCSet.cpp
@@ -92,27 +92,32 @@ int Set::count(void)
 
 void Set::addObject(Object *pObject)
 {
-    CC_SAFE_RETAIN(pObject);
-    _set->insert(pObject);
+    if (_set->count(pObject) == 0)
+    {
+        CC_SAFE_RETAIN(pObject);
+        _set->insert(pObject);
+    }
 }
 
 void Set::removeObject(Object *pObject)
 {
-    _set->erase(pObject);
-    CC_SAFE_RELEASE(pObject);
+    if (_set->erase(pObject) > 0)
+    {
+        CC_SAFE_RELEASE(pObject);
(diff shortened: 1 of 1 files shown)


cocos2dx/cocoa/CCSet.cpp
```diff
@@ -92,27 +92,32 @@ int Set::count(void)
...
     {
-        if (! (*it))
+        if (!(*it))
         {
             break;
         }
-
+        
         (*it)->release();
+        _set->erase(it++);
```
minggo, line 120: Is it safe to remove elements when iterating?

minggo, line 120: It will cause crash on iOS simulator.
I think it can be written like this

``` c++
void Set::removeAllObjects()
{
    SetIterator it = _set->begin();
    SetIterator tmp;

    while (it != _set->end())
    {
        if (!(*it))
        {
            break;
        }

        tmp = it;
        ++tmp;
        _set->erase(it);
        it = tmp;
    }
}
```

minggo, line 120: I fixed it in #3064.

BodbDearg:
Thanks minggo, nice catch! Had copied the fix from https://github.com/cocos2d/cocos2d-x/issues/2283 but didn't realise there were issues with that also. Thanks for fixing it up!

PR #3058, opened 2013-07-02
Add extra safety to selector casting macros

BodbDearg:
Use static_cast<> instead of C-Style casts in the callfunc0_selector() etc. series of macros as this will allow the compiler to perform more stringent error checking and catch out cases where incompatible function signatures are being casted. This helps prevent very subtle and difficult to track down bugs where the signature of the function being casted is not what Cocos2dx expect - which can cause strange things to happen and corrupt the current stack frame.


BodbDearg:
Thank you!

PR #3060, opened 2013-07-03
ios and mac projects unified in one simple xcode project

ricardoquesada:
Not finished yet:
- iOS only includes Test Cpp (Simple Game not working yet on iOS)

It is easier to maintain and add more targets, files, etc.


dumganhar:
I ran with <code>Hello Cpp iOS</code>, but got the error bellow.
cb40ab49-8a68-4d13-944e-7e75f3ef7b14

dumganhar:
Also, we need to add <code> cocos2dx/textures/etc/etc1.h/.cpp</code>, otherwise there will be some compilation errors.

ricardoquesada:
Yes, hello cpp for iOS is not working yet.
Only "Test cpp"

ricardoquesada:
fixed project.
Added Etc. texture files
Added missing iOS projects
Added JS obfuscate for Mac

What missing: 
Lua for iOS (and Mac).

@dumganhar could you add the Lua project here ? I haven't used Lua bindings yet so I don't know what are the needed steps to add it.

Thanks!

dumganhar:
OK, I 'll add them!

James Chen

dumganhar:
Done at https://github.com/cocos2d/cocos2d-x/pull/3074

Added lua projects for both iOS and Mac.

Issue #1145, opened 2012-07-31
CCScale9Sprite does not support rotated spriteframe


chuckzhang:
As described, using rotated sprite frame for CCScale9Sprite does not render correctly.

ghost:
It's not ported from Cocos2d to the Cocos2d-x version. The Cocos2d version does it.

Check

https://github.com/YannickL/CCControlExtension/commit/a920b8e5ba7e2b387d0a97880232a63d004f2d4e

chuckzhang:
I've noticed, but is there any plan to support this feature in the near future?

ghost:
If you update the code? Yes!

walzer:
issue created http://www.cocos2d-x.org/issues/1473 
I will sync your code to c++, thanks!

minggo:
It is fixed in this issue: http://www.cocos2d-x.org/issues/1473

lava-hammer:
I found this bug at release version "cocos2d-2.1rc0-x-2.1.3" May 1st 2013.  And checked my CCScale9Sprite.cpp with latest github version. I didn't see enough changes to fix the bug. Keep watching it.

smilingpoplar:
@minggo, @walzer CCScale9Sprite does not work for rotated spriteframe in cocos2dx v2.1.4 too.

PR #3066, opened 2013-07-03
fix CCScale9Sprite for rotated sprite frame

smilingpoplar:
fix issue #1145

### DIFF
--- a/extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
+++ b/extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
@@ -138,6 +138,7 @@ bool Scale9Sprite::updateWithBatchNode(SpriteBatchNode* batchnode, Rect rect, bo
     _scale9Image->removeAllChildrenWithCleanup(true);
 
     _capInsets = capInsets;
+    m_bSpriteFrameRotated = rotated;
     
     // If there is no given rect
     if ( rect.equals(RectZero) )


extensions/GUI/CCControlExtension/CCScale9Sprite.cpp
```diff
@@ -138,6 +138,7 @@ bool Scale9Sprite::updateWithBatchNode(SpriteBatchNode* batchnode, Rect rect, bo
     _scale9Image->removeAllChildrenWithCleanup(true);
 
     _capInsets = capInsets;
+    m_bSpriteFrameRotated = rotated;
```
dumganhar, line 141: Where was `m_bSpriteFrameRotated` defined?

smilingpoplar, line 141: v2.1.4 has this `m_bSpriteFrameRotated` defined, https://github.com/cocos2d/cocos2d-x/blob/cocos2d-x-2.1.4/extensions/GUI/CCControlExtension/CCScale9Sprite.cpp

I didn't notice you guys have remove this var. Why remove it? 

Should I clone v2.1.4 and push to v2.1.4? I'm not clear about your git workflow. Anyway, adding this line to v2.1.4 works for me.

dumganhar, line 141: It was not removed, but renamed.

James Chen
ÔÚ 2013-7-3 ÏÂÎç11:46£¬"YangLe" notifications@github.comÐ´µÀ£º

smilingpoplar, line 141: alright, let me create a new pull

PR #3069, opened 2013-07-03
Fix win build

slobodin:
std::max and win32 max macro

### DIFF
--- a/cocos2dx/cocoa/CCGeometry.cpp
+++ b/cocos2dx/cocoa/CCGeometry.cpp
@@ -303,10 +303,12 @@ Rect Rect::unionWithRect(const Rect & rect) const
         std::swap(otherTopY, otherBottomY);   // Other rect has negative height
     }
     
-    float combinedLeftX = std::min(thisLeftX, otherLeftX);
-    float combinedRightX = std::max(thisRightX, otherRightX);
-    float combinedTopY = std::max(thisTopY, otherTopY);
-    float combinedBottomY = std::min(thisBottomY, otherBottomY);
+    using namespace std;
+    
+    float combinedLeftX = min(thisLeftX, otherLeftX);
+    float combinedRightX = max(thisRightX, otherRightX);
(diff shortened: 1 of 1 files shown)


cocos2dx/cocoa/CCGeometry.cpp
```diff
@@ -303,10 +303,12 @@ Rect Rect::unionWithRect(const Rect & rect) const
...
-    float combinedLeftX = std::min(thisLeftX, otherLeftX);
-    float combinedRightX = std::max(thisRightX, otherRightX);
-    float combinedTopY = std::max(thisTopY, otherTopY);
-    float combinedBottomY = std::min(thisBottomY, otherBottomY);
+    using namespace std;
+    
+    float combinedLeftX = min(thisLeftX, otherLeftX);
+    float combinedRightX = max(thisRightX, otherRightX);
+    float combinedTopY = max(thisTopY, otherTopY);
+    float combinedBottomY = min(thisBottomY, otherBottomY);
```
dumganhar, line 311: Can't `std::min`, `std::max`  work on windows?

slobodin:
They can't because they hiding by WinAPI macro definition `max` and `min`. Until we specify `NOMINMAX` define before including `windows.h`

winipcfg:
FYI:
http://stackoverflow.com/questions/5004858/stdmin-gives-error

However, adding statement "using namespace std" is not a good practice in general
http://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

slobodin:
> adding statement "using namespace std" is not a good practice in general

In this case this statement takes effect only in scope of `Rect::unionWithRect` function.

So, in windows it uses WinAPI macro defs, in gcc it uses `std::min` or `std::max`.

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/3090

dumganhar:
Thanks.

dumganhar:
Now we're supporting 'std::max' and 'MAX' at the same time.

PR #3070, opened 2013-07-04
Adds const to getters

ricardoquesada:
Many getters in cocos2d are not declared as const.
This patch adds const to many cocos2d properties, specially in CCNode and
subclasses


minggo:
fixed compiling errors in #3072.

PR #3073, opened 2013-07-04
Add namespace prefix to selector macros

xxuejie:
Current selector macros does not have `cocos2d::` namespace prefix, so if we do not use the following line:

```
using namespace cocos2d;
```

The compiler will signal an error.


minggo:
conflict with #3063.

xxuejie:
@minggo Already merged with the latest version. Would you like to take a look again?

minggo:
@xxuejie 
I will merge it if there is not compiling error.

PR #3083, opened 2013-07-05
Support Tizen 2.2 SDK

pyrasis:
- Update project files.
- based on master branch.


pyrasis:
Tizen SDK 2.2 can't compile develop branch, because the SDK doesn't support range-based for loop yet.
so, I work Tizen port on master branch.

dumganhar:
But master branch is for stable release. It means that when develop branch is ready for release, we will merge develop branch to master branch. Thus, we should not modify codes on master branch. Sorry for your inconvenience .

pyrasis:
OK.
I'm going to contact Samsung to this problem.

Thanks.

PR #3084, opened 2013-07-05
Fixed linux compilation bug in eclipse

azmeuk:
Added some guards to prevent Eclipse to compile twice the same class.


dumganhar:
Thanks.

PR #3094, opened 2013-07-08
Linux: added SDL backend as alternative to GLFW.

sergey-shambir:
Summary:
GLFW v.3 released in June. New version breaks API without real need:
http://cocos2d-x.org/boards/6/topics/31277
Meanwhile SDL is very stable. SDL2 used in Steam Runtime and will be
part of Ubuntu SDK, it's supported by other distributions.

Completed:
- Window title now contains app name
- OpenGL initialization and window handling with SDL
- Keyboard, text edit, mouse and touches support with SDL

Notes:
- SDL on X11 have no proper multitouch, but backend handles multitouch
  events to support future Linux graphics.


sergey-shambir:
CI build failed, looks like error in python scripts or some configuration.

NameErrorNameError: : global name 'LLVM_ROOT' is not defined

minggo:
@sergey-shambir 
Could you use 4 spaces instead of tab or 8 spaces.

sergey-shambir:
Done for platform/linux/CCEGLView, header and implementation.

lgb:
sounds good

minggo:
Why should use both SDL and GLFW?
It makes things complex.
I think we should use one for as platforms as possible.

Then it is easy to maintain.

sergey-shambir:
I'm agree. Removed GLFW code.

sergey-shambir:
Oh, I forget that SDL2 is still RC: it does not exists in main Ubuntu 12.04 repos. It was reason why SDL and GLFW backends can co-exists. But there also is another solution: use libsdl1.2 or mixed SDL1.2 / SDL2 backends (that means less macro checks), or put SDL2 copy to sources.

So should I use GLFW2 mixed with SDL2, SDL1.2 mixed with SDL2 or include SDL2 to cocos2dx sources and create make option to use SDL2 provided by OS?

dumganhar:
Could SDL2 be installed by `apt-get` command? 
Could we add a non-official repo like `sudo add-apt-repository xxx` to install `SDL2`?

sergey-shambir:
Yes, there is good candidate: https://launchpad.net/~zoogie/+archive/sdl2-snapshots

dumganhar:
@sergey-shambir 
Seems great, did you test it?
If it works, could you update travis-ci build script for linux port?
It's in `tools/travis-scripts/before-install.sh`

sergey-shambir:
Rebased both commits and updated travis-ci script, now it adds OpenMW team repository: https://launchpad.net/~openmw/+archive/build

Tested on local machine; compiles OK on travis.

dumganhar:
Great, I'll test it.

dumganhar:
Thanks. It was merged.

ricardoquesada:
@sergey-shambir 
What are the benefits of SDL2 over GLFW ? Thanks.

sergey-shambir:
Recently GLFW released new version that have no sugnificant improvements, but changes function names. It added to Fedora repositories, so Fedora has GLFW 3.x when Ubuntu has GLFW 2.x.

SDL 1.x existed for more than 10 years with stable API and ABI, it was added to many distributions due to stability. SDL2 fixes much issues in SDL 1.x. For example, Dota 2 on linux uses SDL2 and can Alt+Tab from fullscreen correctly; multimedia keys work too; this is an obstacle for other fullscreen apps because in X11 window manager concept conflicts with fullscreen concept and window manager hotkeys often do not work. Another thing is support for mobile OSes and their input devices.

Also SDL2 takes more love from distributiion developers: it is part of steam runtime, added to various distros despite that it still in release candidate state, there was work on porting it to Wayland and Mir display servers.

ricardoquesada:
@sergey-shambir 
Thanks.

I haven't used GLFW (although I read its features) and the last time that I used SDL1 was like 10 years ago. So, I'm not very familiar with any those library.

But if I understood correctly, I think GLFW makes more sense (from a feature-wise point of view) than SDL2, since it is more lightweight. SDL is much more than just a simple GL window wrapper, and it includes many features that are already provided by cocos2d.

And since GLFW is light (and also portable), it makes sense to have all our desktops (Windows, Linux and Mac) using it.

My other concern with SDL2 is that it is not available in Ubuntu 13.04 by default.

If GLFW v3.0 is small enough, then we can bundle it with cocos2d-x (we are already doing it with `libpng`, `libfreetype`, etc.).

On the other hand, once the new renderer is complete, we could target "SDL" as a target if needed (although it seems a bit far-fetched). And it seems that SDL2 supports iOS and Android as well.

BTW, apparently libsdl2 v2.0.0 was released: http://www.libsdl.org/tmp/download-2.0.php

My thoughts:
- Adding support for SDL2 could be a good thing for Linux support.
- Removing support for GLFW was a bit premature.

How easy is to re-add GLFW support ? Thanks.

sergey-shambir:
It's not hard to return back GLFW and use preprocessor to select backend.

sergey-shambir:
Updated information: SDL2 dll for windows is about 800KB, but static library can reduce this size since linker can remove functions that never called. With static library user also doesn't need to put dll into game folder to distribute it. Things are same for MacOSX.

On Linux SDL2 is available for developer with PPA, but distribution actually is not easy with both SDL2 and GLFW: only opensource games can be easily added to repositories, only Ubuntu have limited support for commercial games in binary form. Since Ubuntu 13.10 it will be changed: this version will introduce stable platform API (like winapi or OSX APIs), and SDL2 probably will be part of ubuntu platform. So when real distribution will be possible, SDL2 will not add anything to executable size.

Unzipped Windows headers and libs for SDL2 are about 5MB of disk space. 1MB for /include, 200 KB for two SDL2.lib versions (x86 and x64).

sergey-shambir:
On the other hand, Ubuntu platform will also contain Qt5, so maybe SDL2 is not needed at all. Please tell me what to do with that ;)

ricardoquesada:
As far as I know, Ubuntu Phone will use QT + QML as their "default" development toolcahin.

In this ticket we are discussing how to handle this:
http://www.cocos2d-x.org/issues/2460

The idea is to stop having Mac code, Windows code and Linux code, and to merge all the desktops into just one "platform": it could be GLFW or SDL2, or SFML, or QT5 or something else.

PR #3103, opened 2013-07-08
iss3102: make extension test submenu in TestCpp scrollable

himynameschris:
Too many entries in the test list prevent them all from being visible, added logic to scroll menu similar to the logic found int he main test layer to scroll the menu position based on touch


dumganhar:
Thanks, merged at https://github.com/cocos2d/cocos2d-x/pull/3105.

PR #3110, opened 2013-07-09
CocosDenshion: fixed wrong assertion in android OpenSL engine

sergey-shambir:
Playback rate interface cannot be initialized for some players in
android 4.x, it's not cocos2d bug.


minggo:
What will happen if it failed?

sergey-shambir:
Android 4.x have known bug with OpenSL drivers: interface to change playback rate is not available for URI-based player. I've added null pointer check to first pull request, but forgot to remove assertion.

So in debug build TestCpp asserted on Samsung Galaxy S2 with android upgraded to 4.x (actually I have only S3 and it's asserted when I tested). In release build and on android 2.x nothing happened.

minggo:
I think it is better to uncomment assert, and add comments to describe why to do it.
What's your opinion?

PR #3114, opened 2013-07-09
closed #2388: Creating a default texture for CCSprite if setTexture(nullptr) is invoked.

dumganhar:
...


cocos2dx/sprite_nodes/CCSprite.cpp
```diff
@@ -1083,6 +1067,16 @@ void Sprite::updateBlendFunc(void)
     }
 }
 
+static unsigned char cc_2x2_white_image[] = {
```
minggo, line 1080: Please add some comment of this variable.

PR #3117, opened 2013-07-09
Solving 'black screen' on android

jotel:
Hi,

On Andorid there could be a case that an Activity is Destroyed but the JNI code still stays in memory. In such case OpenGL context is recreated and all currently active CCTexture2D objects (in running scene) which have texture names from the not anymore existing  OpenGL context are rendered in black or could render some other textures if the later were created in the mean time. 
The typical scenario is when user changes language configuration in his/her device.
To solve that problem the textures' names need to be rebind(reloaded) based on the initial input parameters and as far as I'm aware the only way is to set CC_ENABLE_CACHE_TEXTURE_DATA to 1 and when OpenGL context is reinitiaded then call CCTextureCache::reloadAllTextures(); That way all you textures are valid again. When you have i18n application it could be a good moment to re-init the FileUtils searchPaths and searchResolutionsOrder to enable load localized resources from new locations (based in current language).

Regards
Jaro

p.s. originally i created that patch for 2.1.4 version... but then realised that some work has been done... unfortunately it didn't solve all the cases (and the one most important for me i.e. changing the device locale).. so I hope that would be useful for you.

### DIFF
--- a/cocos2dx/textures/CCTexture2D.cpp
+++ b/cocos2dx/textures/CCTexture2D.cpp
@@ -560,15 +560,15 @@ bool Texture2D::initWithString(const char *text, const FontDefinition& textDefin
     
     
     #else
-        bool requestUnsupported = textDefinition->_shadow._shadowEnabled || textDefinition->_stroke._strokeEnabled;
+        bool requestUnsupported = textDefinition._shadow._shadowEnabled || textDefinition._stroke._strokeEnabled;
     
         CCAssert(requestUnsupported == false, "Currently shadow and stroke only supported on iOS and Android!");
     
         do
         {
-            CCImage* pImage = new CCImage();
+            Image* pImage = new Image();
             CC_BREAK_IF(NULL == pImage);
-            bRet = pImage->initWithString(text, (int)textDefinition->_dimensions.width, (int)textDefinition->_dimensions.height, eAlign, textDefinition->_fontName.c_str(), (int)textDefinition->_fontSize);
+            bRet = pImage->initWithString(text, (int)textDefinition._dimensions.width, (int)textDefinition._dimensions.height, eAlign, textDefinition._fontName.c_str(), (int)textDefinition._fontSize);
             CC_BREAK_IF(!bRet);
             bRet = initWithImage(pImage);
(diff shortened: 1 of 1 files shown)


minggo:
> > On Andorid there could be a case that an Activity is Destroyed but the JNI code still stays in memory. In such case OpenGL context is recreated and all currently active CCTexture2D objects (in running scene) which have texture names from the not anymore existing OpenGL context are rendered in black or could render some other textures if the later were created in the mean time. 

Good to know, thanks.
Any link to describe it?

minggo:
If the activity is destroyed, how about kill this program?
Reloading all textures and set up all environment will make codes more complex.
If the activity is destroyed, i think it makes sense to kill the program.

tiantian20007:
A game may start a service to get push notification. If you kill the program, the service may be killed too.

jotel:
These are mostly my own findings after some time of debugging and
frustration :)

But here are some links which helps:

Activity lifecycle:
http://developer.android.com/reference/android/app/Activity.html#ActivityLifecycle
List of activity changes causing to destroy it:
http://developer.android.com/guide/topics/manifest/activity-element.html#config
http://developer.android.com/training/basics/activity-lifecycle/starting.html#Destroy

Jaroslaw Lewandowski

jotel@jotel.net
http://www.linkedin.com/in/jotel

minggo:
> > A game may start a service to get push notification. If you kill the program, the service may be killed too.

Ok.
How about not doing rendering in this case?

jotel:
Yes, this is another solution... you could destroy CCDirector ... but I'm
not sure if you ever considered recreating director in the same process (I
don't have you have ability to kill JNI bindings :) ... so I decided to
patch it that way (most of the code was already there) as It could be the
case that clean shutdown of CCDirector might be more complex.. but yes it
seems to be a better approach if works :) I'll stick with my own solution
in my fork for now /I need something working now / :)

Jaro

Jaroslaw Lewandowski

jotel@jotel.net
http://www.linkedin.com/in/jotel

minggo:
Thanks @jotel.
I think we have to use this method before finding better solution.

PR #3118, opened 2013-07-09
Now opacity option for shadow in CCLabelTTF works.

jotel:
Now code uses properly opacity settings for text shadows.


jotel:
not sure why build for emscripten platform failed... I didn't change anything which could affect that...

Jaro

minggo:
It is not your fault.
I have asked Carlo, who implements the feature, to help to review the codes.

PR #3120, opened 2013-07-10
V2x

chengstory:


cocos2dx/platform/CCFileUtils.h
```diff
@@ -252,6 +252,10 @@ class CC_DLL CCFileUtils : public TypeInfo
       * @since v2.1
       */
      void addSearchPath(const char* path);
+
+	 void removeSearchPath(const char *path);
+
+	 void removeAllPath();
```
minggo, line 258: how about change to `removeAllSearchPath()`

cocos2dx/support/component/CCComponent.h
```diff
@@ -43,9 +43,12 @@ class CC_DLL CCComponent : public CCObject
     virtual void serialize(void* r);
     virtual bool isEnabled() const;
     virtual void setEnabled(bool b);
+	virtual void setNode(CCNode *pNode);
+	virtual CCNode* getNode();
```
minggo, line 47: const CCNode\* getNode() const?

PR #3121, opened 2013-07-10
Add append function for String class.

tks2shimizu:
The append function is useful when adding a string.
So, I added it to String class.


tks2shimizu:
Thanks!

PR #3128, opened 2013-07-11
re adds copy() as a deprecated method

ricardoquesada:
...in order to make easier the port to v3.0

### DIFF
--- a/cocos2dx/cocoa/CCObject.h
+++ b/cocos2dx/cocoa/CCObject.h
@@ -49,6 +49,15 @@ public:
 	/** returns a copy of the object */
     virtual Clonable* clone() const = 0;
 	virtual ~Clonable() {};
+
+    /** returns a copy of the object.
+     @deprecated Use clone() instead
+     */
+    CC_DEPRECATED_ATTRIBUTE void copy() const
+    {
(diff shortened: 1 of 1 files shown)


cocos2dx/cocoa/CCObject.h
```diff
@@ -49,6 +49,15 @@ class CC_DLL Clonable
 	/** returns a copy of the object */
     virtual Clonable* clone() const = 0;
 	virtual ~Clonable() {};
+
+    /** returns a copy of the object.
+     @deprecated Use clone() instead
+     */
+    CC_DEPRECATED_ATTRIBUTE void copy() const
```
dumganhar, line 56: Do you think `copy` should return `void*`?
Consider the old usage:

``` cpp
CCShow* action = CCShow::create();
CCShow* myCopy = (CCShow*)action->copy();
```

ricardoquesada, line 56: you are right, it should return `CCObject*`

dumganhar, line 56: Fixed at https://github.com/cocos2d/cocos2d-x/pull/3130.

PR #3130, opened 2013-07-11
Minor fix of PR 3128.

dumganhar:


ricardoquesada:
thanks for fixing it.

PR #3136, opened 2013-07-11
getInstance() / destroyInstance() are used...

ricardoquesada:
... instead of sharedXXX / purgeXXX.
They are more C++ friendly, and also easier to remember.

common files + Mac files + iOS files + tests/samples files were updated.

The old methods are deprecated now.

### DIFF
--- a/samples/Cpp/TestCpp/proj.android/project.properties
+++ b/samples/Cpp/TestCpp/proj.android/project.properties
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-10
+target=android-17
 
 android.library.reference.1=../../../../cocos2dx/platform/android/java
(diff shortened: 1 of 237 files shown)


dumganhar:
Thanks. I'll fix the ci build problem. :)

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3139
Thanks.

samples/Cpp/TestCpp/proj.android/project.properties
```diff
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-10
+target=android-17
```
minggo, line 11: Is there any reason to change to android-17?

Issue #2354, opened 2013-04-11
CCLabelBMFont crash on glDrawElements in CCTextureAtlas::drawNumberOfQuads sometimes


tiantian20007:
It does not happen every time, however It never happen before I upgrade to 2.1.2.

dumganhar:
Which platform?
Does `TestCpp/LabelTest` which uses CCLabelBMFont also crash sometimes?

tiantian20007:
Sorry I forgot to add detail info: xcdoe 4.6.1 crash on simultor 6.1 and device: 6.1.3 & 4.3.3.
TestCpp/LabelTest works fine;

dumganhar:
Could you provide a demo to reproduce this issue and send it to me?

tiantian20007:
OK, I'll do it later.

tiantian20007:
It's hard to reproduce this issue, but it happens sometime.
It crashed at line:631 of CCTextureAtlas.cpp, it's an openGL call "glDrawElements" in CCTextureAtlas::drawNumberOfQuads when a CCLabelBMFont::draw is invoked:

``` c++
glDrawElements(GL_TRIANGLES, (GLsizei) n*6, GL_UNSIGNED_SHORT, (GLvoid*) (start*6*sizeof(m_pIndices[0])) );
```

I looked into m_pIndices, and I found  only one element:  *m_pIndices GLushort 0
It looks like an empty array
May that be the reason for crash?

XCode gave the info: EXC_BAD_ACCESS (code=2, address=0x0)

dumganhar:
Do you mean `m_pIndices` is NULL or the value of `m_pIndices` (i means m_pIndices[0]) is 0?

tiantian20007:
m_pIndices is not NULL. 
But m_pIndices[0] is 0.

tiantian20007:
![tmp4996898b](https://f.cloud.github.com/assets/406098/412468/3f130052-abb7-11e2-8d91-f66d38064954.png)

dumganhar:
So have you found the way to reproduce this issue?

tiantian20007:
No, I just record the stack trace when it happened;

djciego:
I've found the same error. In my case it happens when I do the following:

//Create a label and add it
CCLabelBMFont *label1 = CCLabelBMFont::create("test", "someFont1.fnt");
this->addChild(label1);

// Remove this label
label1->removeFromParentAndCleanup(true);

[...]

PR #3144, opened 2013-07-13
Fix Issue #2354

jllust:
Once a Texture Atlas VAO has deallocated by glDeleteVertexArrays. The GLStateCache may have been recently set to that vaoID. When shortly after a new glGenVertexArrays gets the same vaoID but the glBindVertexArray is bypassed because of the GLStateCache is not aware the VAO was just invalidated, and new vertex are copied to the old freed address.


dumganhar:
Thanks.

PR #3145, opened 2013-07-14
Added JSB support for ccpDistanceSQ

jdmunro:
This function wasn't previously exposed in JSB so I've added the binding.


jdmunro:
I used the old style API from master, not the new one - I will fix that up and then reopen the MR.

dumganhar:
Thanks.

PR #3146, opened 2013-07-14
Add componentsSeparatedByString function for String class.

tks2shimizu:
For String Class and string, there is no function to split the string.
So, I added componentsSeparatedByString function.
The function name is the same as NSString.


dumganhar:
Thanks.

PR #3147, opened 2013-07-14
Changed the order of function calls within CocosDenshion::SimpleAudioEngine::end()

rafaelx:
to first detach the OpenAL source from the buffers so that sound buffers can be successfully deleted afterwards.

Additionally, improved the documentation within the header file a little bit.

### DIFF
--- a/CocosDenshion/openal/SimpleAudioEngineOpenAL.cpp
+++ b/CocosDenshion/openal/SimpleAudioEngineOpenAL.cpp
@@ -111,7 +111,12 @@ namespace CocosDenshion
 
     static void stopBackground(bool bReleaseData)
     {
-		alSourceStop(s_backgroundSource);
+        // The background music might have been already stopped
+        // Stop request can come from 
+        //   - stopBackgroundMusic(..)
+        //   - end(..)
+        if (s_backgroundSource != AL_NONE)
+            alSourceStop(s_backgroundSource);
 
 		if (bReleaseData)
 		{
@@ -130,7 +135,7 @@ namespace CocosDenshion
 			}
 		}
 
-		s_backgroundSource = AL_NONE;
+        s_backgroundSource = AL_NONE;
     }
 
     static void setBackgroundVolume(float volume)
@@ -246,6 +251,7 @@ namespace CocosDenshion
 
 	void SimpleAudioEngine::playBackgroundMusic(const char* pszFilePath, bool bLoop)
 	{
+        // If there is already a background music source we stop it first
 		if (s_backgroundSource != AL_NONE)
 			stopBackgroundMusic(false);
 
@@ -263,6 +269,7 @@ namespace CocosDenshion
 		{
 			s_backgroundSource = it->second->source;
 			alSourcei(s_backgroundSource, AL_LOOPING, bLoop ? AL_TRUE : AL_FALSE);
+            setBackgroundVolume(s_volume);
 			alSourcePlay(s_backgroundSource);
 			checkALError("playBackgroundMusic:alSourcePlay");
 		}
@@ -270,6 +277,10 @@ namespace CocosDenshion
 
 	void SimpleAudioEngine::stopBackgroundMusic(bool bReleaseData)
 	{
+        // If there is no source, then there is nothing that can be stopped
+        if (s_backgroundSource == AL_NONE)
+            return;
+
         ALint state;
         alGetSourcei(s_backgroundSource, AL_SOURCE_STATE, &state);
         if (state == AL_PLAYING)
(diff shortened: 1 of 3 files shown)


rafaelx:
When CocosDenshion::SimpleAudioEngine::end() tries to delete the sound buffers for effects and background music it results at Linux to a failure message on the shell.

AL_INVALID_OPERATION in end:alDeleteBuffers
AL_INVALID_OPERATION in end:alDeleteBuffers

The reason for this message is that the buffers are still attached to the source.
Changing the order of function calls solves the problem, means the buffer needs to be detached from the source and then it can be deleted without problems. In the current implementation it was the other way around.

See also SimpleAudioEngine::unloadEffect, where the cleanup was implemented in the correct order.

dumganhar:
Thanks for your specific explain. I will test it soon.

dumganhar:
@rafaelx 
I tested this issue on Ubuntu 12.10 64bit, I still get some errors after applying your changes.

```
AL_INVALID_NAME in end:alSourceStop
```

rafaelx:
Oh, this might be caused by an API function call that I have not used so far.
I will have a deeper look into this issue and will run additionally the CocosDenshion tests from the samples directory.

rafaelx:
Hi dumganhar!

I switched to the 'develop' branch to see how the original code behaves, so I run CocosDenshionTest.
What I have noticed is after playing the background music for the first time the following...

AL_INVALID_NAME in playBackgroundMusic:alSourcei

..but the music got played.

If the music is stopped and played again than this message from 'playBackgroundMusic' will not show again.
It appears only if 'alSourcei(s_backgroundSource, AL_LOOPING, bLoop ? AL_TRUE : AL_FALSE)' is called for the first time.

According to the documentation that I have found this message could be an indication that the passed source id to alSourcei is not valid, which could be verified with 'alIsSource', but it returned always true.

The source id gets generated within 'SimpleAudioEngine::preloadBackgroundMusic' and stored for later use in a map from which 'SimpleAudioEngine::playBackgroundMusic' gets the source id back and delegates it to 'alSourcei'. The generated and used source id was always the same.

Stopping the music triggers in the original code..

AL_INVALID_NAME in end:alSourceStop

..but I see also..

AL_INVALID_OPERATION in end:alDeleteBuffers

..on which the provided fix is focusing.

I think, there is also another issue in the OpenAL related implementation of CocosDenshion and not any side effect that is caused by my changes.

rafaelx:
I guess, after looking more on the existing code, there is still room for more improvements.
I will update my pull request accordingly and drop an info here.

dumganhar:
OK. Thanks.

rafaelx:
The implementation of SimpleAudioEngineOpenAL uses a global static variable to maintain its internal state.
This variable 's_backgroundSource' is either set to a valid source in the sense of OpenAL or to 'AL_NONE'.

Problem A)
In CocostDenshionTest we have the following scenario.
1. Background music gets preloaded, but the global variable is then still not set to a source, which is ok, because this is only a preparing step.
2. Then, the volume gets set but this fails, because there is still no source available on which the OpenAL function can operate on. Hence, the OpenAL function call to set the volume produces an error which is not checked within 'setBackgroundVolume'.
3. Now, when the user initiates the playback of the background music then within 'playBackgroundMusic' the previously generated OpenAL error will be catched and appears.

Problem B)
Some functions like 'pauseBackgroundMusic', 'resumeBackgroundMusic', 'rewindBackgroundMusic' and 'isBackgroundMusicPlaying' might be used befor the playback was ever started. All those functions are making a call to appropriate OpenAL functions without checking whether a valid source exists. This will also cause error messages to appear later or sooner. 

Problem C)
Stopping a background music is made by a call to 'stopBackgroundMusic(bool)' which calls 'stopBackground(bool)'.
With the boolean one can indicate whether to delete the buffer that is attached to the source or not.
In both cases the static property 's_backgroundSource' will be set to 'AL_NONE', means we forget about the source.
This has the effect that you can not adjust the volume and rewind, but also function calls to resume and pause would fail or at least cause an OpenAL error, due to the missing check for a valid source. (-> Problem B) 

Problem D)
In CocostDenshionTest the destructor of SimpleAudioEngine is never called, hence the function call to alutExit is missing.
-> Issue #3148 

The pull request fixes the mentioned problems and provides some more info within the documentation.

CocosDenshion/openal/SimpleAudioEngineOpenAL.cpp
```diff
@@ -307,13 +340,23 @@ of this software and associated documentation files (the "Software"), to deal
 
 	bool SimpleAudioEngine::willPlayBackgroundMusic()
 	{
-		return true;
```
rafaelx, line 310: I think this method should return only <i>true</i> because if the background music is stopped <i>s_backgroundSource</i> is always set to <i>AL_NONE</i>.
I forgot to remove this change, which I have introduced due to my change on <i>stopBackgroundMusic</i>, where I have set <i>s_backgroundSource</i> to <i>AL_NONE</i> only when deleting the source and buffers, but I removed this back to the original version.

dumganhar:
Thanks, I'll test it soon.

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3203
Thanks.

PR #3155, opened 2013-07-15
override updateDisplayedColor and updateDisplayedOpacity method in Scale9Sprite class

pktangyue:
Since Scale9Sprite extends from NodeRGBA, but it didn't updateDisplayedColor and updateDisplayedOpacity method .

So if a LayerRGBA, which has `setCascadeColorEnabled(true)` and `setCascadeOpacityEnabled(true)`, has a child of Scale9Sprite,  when it `setColor` or `setOpacity`, the color or opacity of the Scale9Sprite child would not change.

This patch override these method and solove the problem.


dumganhar:
Thanks.

PR #3156, opened 2013-07-15
Ignore formatting specifiers in JavaScript log messages.

colesbury:


scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -583,7 +583,7 @@ JSBool ScriptingCore::log(JSContext* cx, uint32_t argc, jsval *vp)
         JS_ConvertArguments(cx, argc, JS_ARGV(cx, vp), "S", &string);
         if (string) {
             JSStringWrapper wrapper(string);
-            js_log((char *)wrapper);
+            js_log("%s", (char *)wrapper);
```
dumganhar, line 586: Hi, what's the error if we don't add `%s`?

colesbury, line 586: The function cc.log() doesn't use printf style formatting, however js_log treats it as if it does.  If you log a message that contains %s or other formatting specifiers, the log message will be incorrect and potentially crash the program.

For example, the code:

cc.log("Foo %s bar");

Should produce a log message like:

JS: Foo %s bar

But before this PR would produce something like:

JS: Foo  [GARBAGE]  bar

Where [GARBAGE] is garbage from the stack.

dumganhar, line 586: Thank you. It makes sense.

PR #3157, opened 2013-07-15
Adds more readability to the class internals.

ricardoquesada:
Changes:
- creator (static) methods, and singleton methods (static) are always at the top of the class
- Constructors, destructros, and init methods comes next
- Then the instance methods for the class
- Then the overriden methods
- and finally the ivars

Also, overriden methos have the "override" context keyword


dumganhar:
Thanks.

PR #3169, opened 2013-07-16
Add JSFunctionWrapper in preparation for bindings-generator changes.

colesbury:
This adds an implementation of JSFunctionWrapper, which is used by this bindings-generator PR:

https://github.com/cocos2d/bindings-generator/pull/16

The JSFunctionWrapper adds roots to a JSObject and jsval for the lifetime of the wrappers.  The intended use is something like this:

std::shared_ptr<JSFunctionWrapper> func(new JSFunctionWrapper(cx, jsthis, jsfunc));
auto lambda = [=]() -> { func.invoke(...) }

The lambda captures the shared_ptr by value, so the lifetime of the wrapper (and JS root) is tied to the lifetime of the lambda.

### DIFF
--- a/scripting/javascript/bindings/ScriptingCore.h
+++ b/scripting/javascript/bindings/ScriptingCore.h
@@ -310,6 +310,34 @@ private:
     JSStringWrapper &operator=(const JSStringWrapper &another);
 };
 
+// wraps a function and "this" object
+class JSFunctionWrapper
+{
+    JSContext *_cx;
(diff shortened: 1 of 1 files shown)


dumganhar:
Cool, I will test it. :+1:

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3180. Also updated bindings-generator reference.

dumganhar:
Thanks a lot.

PR #3170, opened 2013-07-17
Some more changes to make the code more c++ friendly:

ricardoquesada:
- nodeToParentTransform -> getNodeToParentTransform() (the same for its friends)
  - Deprecates the old methods
  - adds 'consts' to those methods
- boundingBox() -> getBoundingBox()
  - the new one is const
  - Deprecates the old method.
- Adds overrides keywords in CCNodeRGBA and CCArmature
- AffineMatrix are mutables

### DIFF
--- a/cocos2dx/base_nodes/CCNode.h
+++ b/cocos2dx/base_nodes/CCNode.h
@@ -464,7 +464,7 @@ public:
      * 
      * @param fRotationX    The X rotation in degrees which performs a horizontal rotational skew.
      */
-    virtual void setRotationX(float fRotaionX);
+    virtual void setRotationX(float rotaionX);
     /**
      * Gets the X rotation (angle) of the node in degrees which performs a horizontal rotation skew.
      *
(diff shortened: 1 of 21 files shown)


cocos2dx/base_nodes/CCNode.h
```diff
@@ -1000,7 +1003,7 @@ class CC_DLL Node : public Object
      *
      * @return The number of actions that are running plus the ones that are schedule to run
      */
-    unsigned int numberOfRunningActions(void);
+    unsigned int numberOfRunningActions();
```
minggo, line 1006: Should also change to `getNumberOfRunningActions`?

dumganhar:
Thanks. Merged at https://github.com/cocos2d/cocos2d-x/pull/3177

cocos2dx/base_nodes/CCNode.h
ricardoquesada, line 1006: yes.
We should use the `get` prefix for all our getters. `numberOfRunningActions` follows the Objective-C pattern, and we should move away from it.

PR #3178, opened 2013-07-17
Refactor CCImage and CCTexture2D

boyu0:
Refactor CCImage and CCTexture2D, make CCImage support for gray image data and add some convert function of convert image data to different texture data.
Build the libjepg, libpng and libtiff to support ios and mac, Let the ios and mac use the these library to parse the image file.

### DIFF
--- a/cocos2dx/platform/CCImageCommon_cpp.h
+++ b/cocos2dx/platform/CCImageCommon_cpp.h
@@ -1085,14 +1085,14 @@ bool Image::initWithPVRv3Data(void *data, int dataLen)
     
     if (!testFormatForPvrTCSupport(pixelFormat))
     {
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/android/java/project.properties
```diff
@@ -12,4 +12,4 @@
 
 android.library=true
 # Project target.
-target=android-10
+target=android-17
```
dumganhar: Why did you change android-10 to android-17?

cocos2dx/platform/CCImage.h
```diff
@@ -49,6 +58,14 @@ class CC_DLL Image : public Object
     
     Image();
     virtual ~Image();
+    
+    /**
+     @brief Determine how many mipmaps can we have. 
+     Its same as define but it respects namespaces
+    */
+    enum {
```
minggo, line 66: use `enum class` instead

PR #3183, opened 2013-07-17
more best practices fixes

ricardoquesada:
- Adds more `const` in getters
- Overriden methods have the `override` keyword
- Reorganizes the structure of the class:
  # creators first, then constructor, destructors, init
  # then overridden methods
  # variables at the end
- removes Hungarian notation from parts of the code

### DIFF
--- a/cocos2dx/CCScheduler.h
+++ b/cocos2dx/CCScheduler.h
@@ -52,36 +52,35 @@ class Set;
 class CC_DLL Timer : public Object
 {
 public:
+    /** Allocates a timer with a target and a selector. */
+    static Timer* createWithTarget(Object *target, SEL_SCHEDULE selector);
+    /** Allocates a timer with a target, a selector and an interval in seconds. */
+    static Timer* createWithTarget(Object *target, SEL_SCHEDULE selector, float seconds);
+    /** Allocates a timer with a script callback function and an interval in seconds. */
+    static Timer* createWithScriptHandler(int nHandler, float seconds);
+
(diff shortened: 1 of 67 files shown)


cocos2dx/CCScheduler.h
```diff
@@ -52,37 +52,36 @@ of this software and associated documentation files (the "Software"), to deal
 class CC_DLL Timer : public Object
 {
 public:
+    /** Allocates a timer with a target and a selector. */
+    static Timer* createWithTarget(Object *target, SEL_SCHEDULE selector);
```
minggo, line 56: How about just named with create()?
Because all other classes just name create().

dumganhar, line 56: @ricardoquesada 
I agree with Minggo, we need to use 'createWithXXX' method only when 'create' cannot distinguish parameters.
For example, in this situation

``` cpp
static Sequence* create(FiniteTimeAction *pAction1, ...);
static Sequence* createWithTwoActions(FiniteTimeAction *pActionOne, FiniteTimeAction *pActionTwo);
```

`createWithTwoActions` is needed since if it's `create` as the first one, complier will trigger compilation errors.

Therefore, IMO, `createWithXXX` is only needed in some cases.
If I'm wrong, please correct me.

Anyway, I merged your PR after I resolved the conflicts. If you agree with my point, maybe you need to correct it in another PR or I'll be glad to do that. :)

ricardoquesada, line 56: I agree with you guys.
The only thing that I did was to rename `timerWithTarget` from `createWithTarget`, because `timerWithTarget` is not C++ friendly.

I will send another PR later.

minggo, line 56: Thanks.

Issue #2465, opened 2013-04-26
Improve CCDirector::purgeCachedData


tiantian20007:
I see when a game running on iOS get a memory warning, cocos2d::CCDirector::sharedDirector()->purgeCachedData() will be called. Inside it:

``` C++
void CCDirector::purgeCachedData()
{
    CCLabelBMFont::purgeCachedData();    
    if (s_SharedDirector->getOpenGLView())
    {
        CCTextureCache::sharedTextureCache()->removeUnusedTextures();
    }
    CCFileUtils::sharedFileUtils()->purgeCachedEntries();
}
```

However, most case, we use CCSpriteFrame which will retain textures.
So I think it is better to do like this:

``` C++
void CCDirector::purgeCachedData()
{
    CCLabelBMFont::purgeCachedData();
    if (s_SharedDirector->getOpenGLView())
    {
        CCSpriteFrameCache::sharedSpriteFrameCache()->removeUnusedSpriteFrames();
        CCTextureCache::sharedTextureCache()->removeUnusedTextures();
    }
    CCFileUtils::sharedFileUtils()->purgeCachedEntries();
}
```

What's your opinion?

dualface:
+1

dumganhar:
+1 too.
I agree. SpriteFrameCache needs to be cleared.

tiantian20007:
https://github.com/cocos2d/cocos2d-x/pull/3189

PR #3189, opened 2013-07-18
Also removeUnusedSpriteFrames when purgeCachedData;

tiantian20007:
When a game running on iOS get a memory warning, cocos2d::CCDirector::sharedDirector()->purgeCachedData() will be called. 
However, most case, we use CCSpriteFrame which will retain textures.
So we also need to removeUnusedSpriteFrames when purgeCachedData.


tiantian20007:
https://github.com/cocos2d/cocos2d-x/issues/2465

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/3209
Thanks.

PR #3192, opened 2013-07-18
More best practices fixes

ricardoquesada:
- class order fixes
  - static methods (creators) at the beginning
  - then constructors / destructors
  - then init
- more `const` fixes
  - adds some const getters
  - some getters have 2 versions: `const` and no-const version
- renamed CocosDenshion::sharedEngine -> getInstance()


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3197
Thanks.

PR #3196, opened 2013-07-19
[Tizen] Fix compile error

pyrasis:
- Tizen SDK doesn't support override keyword yet.
- add class keyword before friend keyword.


dumganhar:
Thanks.

PR #3201, opened 2013-07-19
Adding 'the greatest change.' 

pauljonescodes:
A resolution policy that will scale based on which difference in design resolution and screen size is the bigger one.


dumganhar:
Why do you name it 'greatest'? That's a bit confused.

dumganhar:
And could you describe that in what situation we need to use this strategy?

dumganhar:
@PLJNS 
And could you describe that in what situation we need to use this strategy?

cocosQA:
Dear PLJNS:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6488

PR #3202, opened 2013-07-20
more fixed related to C++ best practices

ricardoquesada:
- Removes some CC_PROPERTY / CC_SYSTHESIZE
- `CCAssert` -> `CCASSERT` since it is a macro
- `unsinged int` -> `int` in some places where it could be difficult to find underflow bugs

and other minor fixes


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3205
Thanks.

PR #3210, opened 2013-07-22
Extensions deprecations

azmeuk:
Hi. I added a file CCDeprecated-ext.h with deprecated class names, typedefs and enums.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -414,6 +414,7 @@ Developers:
         Added some guards to prevent Eclipse to compile twice the same class.
         Linux Eclipse projects updates
         Refactored emscripten-build.sh, it's no longer need to be edited to make emscripten work.
+        Creation of CCDeprecated-ext.h
 
     elmiro
         Correction of passed buffer size to readlink and verification of result return by readlink.


extensions/CCDeprecated-ext.h
```diff
@@ -0,0 +1,349 @@
...
+CC_DEPRECATED_ATTRIBUTE typedef AssetsManager CCAssetsManager ;
+CC_DEPRECATED_ATTRIBUTE typedef AssetsManagerDelegateProtocol CCAssetsManagerDelegateProtocol ;
+CC_DEPRECATED_ATTRIBUTE typedef PhysicsSprite CCPhysicsSprite ;
+CC_DEPRECATED_ATTRIBUTE typedef PhysicsDebugNode CCPhysicsDebugNode ;
+CC_DEPRECATED_ATTRIBUTE typedef ComController CCComController ;
+CC_DEPRECATED_ATTRIBUTE typedef ComAttribute CCComAttribute ;
+CC_DEPRECATED_ATTRIBUTE typedef InputDelegate CCInputDelegate ;
+CC_DEPRECATED_ATTRIBUTE typedef ComAudio CCComAudio ;
+
+#if (CC_TARGET_PLATFORM != CC_PLATFORM_LINUX)
```
azmeuk: I just can tell that Websockets are not enable with Linux, but I am not sure for the other platforms.

dumganhar:
Hi @azmeuk 
For extensions, I think we should just add deprecated typedef in relative header files.
CCDeprecated-ext isn't suitable for extensions.
Reason:
There are lots of modules in extensions. Unlike engine, we deal with engine as one module, so `CCDeprecated.h` is reasonable. But for extensions, it's better to make the `deprecated typedef` in the relative header files.
Thanks.

azmeuk:
You mean for each .h file, set the deprecation after the class definition, or you mean a CCDeprecated.h file for each extension project ?

dumganhar:
@azmeuk 
I think it a bit more, I think your approach is good for now.
I'm trying to merge your PR now.

PR #3212, opened 2013-07-22
Updated emscripten-build.sh

azmeuk:
emscripten-build no longer need to be edited to make emscripten work. User just needs to set $PYTHON and $LLVM if installed in a non standard location.


dumganhar:
Thanks.

PR #3216, opened 2013-07-23
More best practices fixes

ricardoquesada:
- Removes Hungarian notation from the CPP test (there are still some Hungarian to remove)
- Label and MotionStreak: Uses the CL() entry for the tests (lambda functions)

### DIFF
--- a/samples/Cpp/TestCpp/Classes/PerformanceTest/PerformanceSpriteTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/PerformanceTest/PerformanceSpriteTest.cpp
@@ -303,20 +303,20 @@ void SpriteMainScene::initWithSubTest(int asubtest, int nNodes)
     addChild(infoLabel, 1, kTagInfoLayer);
 
     // add menu
-    SpriteMenuLayer* pMenu = new SpriteMenuLayer(true, TEST_COUNT, s_nSpriteCurCase);
-    addChild(pMenu, 1, kTagMenuLayer);
-    pMenu->release();
+    SpriteMenuLayer* menuLayer = new SpriteMenuLayer(true, TEST_COUNT, s_nSpriteCurCase);
+    addChild(menuLayer, 1, kTagMenuLayer);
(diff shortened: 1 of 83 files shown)


samples/Cpp/TestCpp/Classes/PerformanceTest/PerformanceSpriteTest.cpp
```diff
@@ -408,77 +408,77 @@ void SpriteMainScene::onDecrease(Object* pSender)
 // For test functions
 //
 ////////////////////////////////////////////////////////
-void performanceActions(Sprite* pSprite)
+void performanceActions(Sprite* scene)
```
dumganhar: `pSprite` --> `scene` ?

dumganhar: Lots of `pSprite` --> `scene`, wrong replacement?

ricardoquesada: ouch. thanks. I'll fix it.

ricardoquesada:
@dumganhar fixed, and I added other fixes.

As you can see, it won't compile, but this is because the JS bindings are outdated.
I renamed a few methods (that were renamed recently, but the new names weren't C++ friendly), and that's why the JS bindings and Lua bindings are not compiling.

thanks.

dumganhar:
> As you can see, it won't compile, but this is because the JS bindings are outdated.
> I renamed a few methods (that were renamed recently, but the new names weren't C++ friendly), and that's why the JS bindings and Lua bindings are not compiling.

OK, I'll check it. I like the new naming. :)
Thanks.

minggo:
Hi @ricardoquesada 
There is a header file named `CCDeprecated.h` in include.
We will move `CCLog` into it.

ricardoquesada:
Thanks @minggo.
Next time I will move the deprecated functions there.

PR #3231, opened 2013-07-24
Resolved compilation errors in PR https://github.com/cocos2d/cocos2d-x/pull/3216

dumganhar:


dumganhar:
@ricardoquesada 
We have added a new header file for deprecated methods or macros. So the deprecated method 'CCLog' should be placed into 'CCDeprecated.h', I have done that in this commit aaa5d45 .

dumganhar:
@ricardoquesada 
It was merged.

PR #3237, opened 2013-07-24
closed #2428: Renaming namespace from 'cocos2d' to 'cc' and marking 'cocos2d' as the alias of 'cc'.

dumganhar:


pilwon:
@dumganhar Wouldn't it make more sense to set namespace alias in the other direction? (`namespace cc = cocos2d;`)

dumganhar:
I think @walzer will describe the reason about this.

minggo:
@pilwon 
I think there are two reasons
- `cc` is already used in JSB and cocos2d-html5. So it makes more consistent.
- `cc` is short than `cocos2d`, easy to print

Is there any potential problem to change the namespace to `cc`.
Please let us know.
Thanks.

ricardoquesada:
this was previously discussed in the forum, and we agreed that the namespace should be `cocos2d`.
In any case, if someone wants to use `cc`, they could create an alias.

http://www.cocos2d-x.org/boards/6/topics/29971

minggo:
Ok.
Let's close it.

PR #3251, opened 2013-07-25
Added virtual and override keywords to Scene::init() method

metalbass:
Scene::init() method is hidding (and should be overriding) Node::init().
I've also added c++11 override keyword to it.


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/5422.
Thanks.

Issue #3095, opened 2013-07-08
sigslot::_connection_base# (from 0 to 8) don't have virtual destructors


metalbass:
A subclass of connection_base3 is beign deleted through a pointer of this type, which doesn't have a virtual destructor and will cause a memory leak because no other destructors will be called!!!
You can see it in sigslot.h lines 760, 872, 916...

This happens with a lot of "interfaces" of sigslot. Every class that will be inherited MUST have virtual destructors because someone could delete a object of through a pointer to them and have a memory leak.

PR #3253, opened 2013-07-25
Added virtual destructors to some interfaces to fix issue 3095

metalbass:
From https://github.com/cocos2d/cocos2d-x/issues/3095 :

A subclass of connection_base3 is beign deleted through a pointer of this type, which doesn't have a virtual destructor and will cause a memory leak because no other destructors will be called!!!
You can see it in sigslot.h lines 760, 872, 916...

This happens with a lot of "interfaces" of sigslot. Every class that will be inherited MUST have virtual destructors because someone could delete a object of through a pointer to them and have a memory leak.


dumganhar:
Merged, thanks.

PR #3254, opened 2013-07-25
Added virtual keyword to Node::description

metalbass:
One should be able to override this and use it like a toString method for debug.


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/CCNode.h#L150
Thanks.

PR #3257, opened 2013-07-25
move enum into class

minggo:


michaelcontento:
Just a short question: Why don't you use `enum class` instead of `enum`?

This would simple replace the current all-caps prefix with a real scope. As an example
`Director::getInstance()->setProjection(Director::PROJECTION_3D);`
would become:
`Director::getInstance()->setProjection(Director::Projection::3D);`

minggo:
@michaelcontento 
Any benefit?

dumganhar:
@michaelcontento , Do you mean like this?

``` cpp
class Director
{
    class Projection
    {
         enum
         {
               2D,
               3D
          }
    };
};

```

ricardoquesada:
I prefer `Director::PROJECTION_3D` to `Director::Projection::3D`.
I think it is easier to read.

@dumganhar 
In C++11 you can use:

``` c++
class Director {
   enum class Projection {
      2D,
      3D,
   }
}
```

ricardoquesada:
Thinking about this feature again, I think what @michaelcontento suggested is better, because it will make the code more consistent.
so +1 for @michaelcontento suggestion.

ricardoquesada:
I added a PR for this:
https://github.com/cocos2d/cocos2d-x/pull/3261

minggo:
Good to know, thanks.

minggo:
But if an `enum` is already declared in a class, now want to change it to `enum class`, how to keep compatibility?

```
class A
{
public:
    enum Enum1
    {
        ...
    };

    // parameter type is Enum1
    void func1(Enum1 e);
};
```

ricardoquesada:
@minggo 
I didn't understand the problem.

In you example the type will be the same:

``` c++
// With enum
class A {
enum Type {
 HELLO_WORLD,
};
void func(Type type);
}
```

``` c++
// with enum class
class A {
enum class Type {
 HELLO_WORLD,
};
void func(Type type);
}
```

What are the differences:

``` c++
A obj;
obj.func( A::HELLO_WORLD ); // Valid only for the 1st example
obj.func( A::Type::HELLO_WORLD); // Valid for both examples
```

Also in the 1st example, you can have collisions.

More info: 
http://www.cprogramming.com/c++11/c++11-nullptr-strongly-typed-enum-class.html

minggo:
``` c++
class A
{
public:
    enum Enum1
    {
        kEnum1,
    };

    // parameter type is Enum1
    void func1(Enum1 e);
};

// usage
A a;
a.func1(A::kEnum1);
```

 With enum class

``` c++
class A
{
public:
    enum class Enum1
    {
        kEnum1,
    };

    // parameter type is Enum1
    void func1(Enum1 e);
};

// usage
A a;
a.func1(A::kEnum1); // error here, should pass A::Enum1::kEnum1
```

ricardoquesada:
@minggo Yes, that's correct. What is the problem with that ?

minggo:
The problem is how to keep compatibility?

ricardoquesada:
@minggo 
Sorry, I don't understand the problem. Could you put an example ?
cocos2d v2.1.4 uses `kNameVar` for everything. There is no problem in using `enum class` for v3.0 and adding a typedef in `CCDeprecated.h` in order to be backwards compatible.

If I'm missing something, please let me know. thanks.

minggo:
Yep, if an enum is declared outside of a class, it is ok.
But if an enum is already declared in a class as previous example, then i don't know how to keep compatibility.

May be you can refer to `WebSocket.h`, which already declare an enum in it named `ErrorCode`.
If i use `enum class` instead of `enum`.
I don't know how to keep compatibility.

minggo:
Because in v2.1.4, it uses `WebSocket::ERROR_TIME_OUT`.
But how to translate to `WebSocket::ErrorCode::TIME_OUT`(suppose `ERROR_TIME_OUT` is renamed to `TIME_OUT`)?

PR #3261, opened 2013-07-25
Constants fixes

ricardoquesada:
`Director::PROJECTION_XXX` -> `Director::Projection::XXX`
`Label::TEXT_ALIGNMENT_XXX` -> `Label::HAlignment::XXX`
`Label::TEXT_VERTICAL_ALIGNMENT_XXX` -> `Label::VAlignment::XXX`
`Texture2D::PIXEL_FORMAT_XXX` -> `Texture2D::PixelFormat::XXX`
`BlendFunc::BLEND_FUNC_DISABLE` -> `BlendFunc::DISABLE`
`Transition::ORIENTATION_XXX` -> `Transition::Orientation::XXX`
`Image::FORMAT_XXX` -> `Image::Format::XXX`
`Image::_function()` -> `Image::function()` // private methods should not start with '_'
`DrawPrimitives::drawColor()` -> `DrawPrimitives::setDrawColor()`

Also adds more constant for BlendFunc
And fixes Premultiplied Alpha in PVR v3 images
And tests are updated with the new functions

**IMPORTANT**
JS bindings doesn't compile because the `enum class`, although they are using `int`, they don't have implicit conversions.
So, the JS bindings config file, we need to add a rule that says: `Image::Format` is an `int`, etc...

### DIFF
--- a/cocos2dx/layers_scenes_transitions_nodes/CCLayer.h
+++ b/cocos2dx/layers_scenes_transitions_nodes/CCLayer.h
@@ -57,16 +57,7 @@ All features from Node are valid, plus the following new features:
 */
 class CC_DLL Layer : public Node, public TouchDelegate, public KeypadDelegate
 {
-public:
-    /** Different ways to dispatch touches */
-    enum class TouchDispatchMode
-    {
-        /** Dispatches all the events at once, in set of events */
-        ALL_AT_ONCE,
-        /** Dispatches the touches one at the time */
-        ONE_BY_ONE,
-    };
-    
+public:    
     /** creates a fullscreen black layer */
     static Layer *create(void);
     Layer();


dumganhar:
OK, I'll fix the issue of `enum class` in Bindings-generator today. :)

cocos2dx/CCDirector.h
```diff
@@ -85,19 +85,19 @@ class CC_DLL Director : public Object, public TypeInfo
...
-    enum Projection
+    enum class Projection
     {
         /// sets a 2D projection (orthogonal projection)
-        PROJECTION_2D,
+        _2D,
         
         /// sets a 3D projection with a fovy=60, znear=0.5f and zfar=1500.
-        PROJECTION_3D,
+        _3D,
```
dumganhar, line 94: Should it be `2D` rather than `_2D`, right?

ricardoquesada, line 94: Ideally yes, but C++ does not accept enums that start with numbers.
So, I added the `_` as a prefix.
This same "best practice" is being used by the STL. eg:
`std::placeholder::_1`... and not `std::placeholder::1`  because it cannot start with number.

dumganhar, line 94: Got it.

ricardoquesada:
I added 2 more commits.

Bindings won't compile because they need to be regenerated.

dumganhar:
@ricardoquesada 
Updating: Binding-generator supports `strong enum type` now.
I 'll update its reference in another PR. :)

ricardoquesada:
@dumganhar 
good to know!

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3265
Thanks.
BTW, one thing we have to know : 

``` cpp
typedef enum Director::Projection ccDirectorProjection;
```

 can't work on Windows, since `Director::Projection` is already a type, no need to add `enum` keyword.

ricardoquesada:
ok. thanks. good to know.

PR #3266, opened 2013-07-26
Documentation fix: use \p to reference a parameter

timothyqiu:
The doc of `Follow::create()` was broken due to the use of @param tag.

```
/** creates the action with a set boundary,
 It will work with no boundary if @param rect is equal to Rect::ZERO.
 */
```

renders:

```
creates the action with a set boundary, It will work with no boundary if
Parameters
    rect    is equal to Rect::ZERO.
```

The 'work with no boundary if' part was meant to be followed by 'rect is equal to Rect::ZERO' part.

The orginal @param is a way to describe a parameter. To reference a parameter, use `\p` instead.

Ref: http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdp

### DIFF
--- a/cocos2dx/actions/CCAction.h
+++ b/cocos2dx/actions/CCAction.h
@@ -201,8 +201,11 @@ Instead of using Camera as a "follower", use this action instead.
 class CC_DLL Follow : public Action
 {
 public:
-    /** creates the action with a set boundary,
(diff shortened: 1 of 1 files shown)


minggo:
Thanks.

PR #3272, opened 2013-07-26
Make CCBAnimationManager::getSequenceId public; add method to get duration of timeline in ccb;

tiantian20007:
It's common case to check whether a timeline exist by name; So CCBAnimationManager::getSequenceId should be public; 

Adding the CCBAnimationManager::getSequenceDuration method to get duration of exist timeline is pretty usefully when we make our animation all with cocosbulder.

Yes, there already is a method CCBAnimationManager::getSequence(int nSequenceId), then we can call CCBSequence::getDuration to get that info. However, CCBSequence is not binded for JS, we need a more simple interface to get that info for JS^^

### DIFF
--- a/extensions/CCBReader/CCBAnimationManager.cpp
+++ b/extensions/CCBReader/CCBAnimationManager.cpp
@@ -261,6 +261,14 @@ CCBSequence* CCBAnimationManager::getSequence(int nSequenceId)
     return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


dumganhar:
Could you pull the latest codes? Otherwise, this PR could not be automatically merged.

tiantian20007:
done
https://github.com/cocos2d/cocos2d-x/pull/3363

dumganhar:
Thanks.

PR #3284, opened 2013-07-26
Doesn't use deprecated methods from DrawingPrimitives

ricardoquesada:
Also renames DrawPrimitives::drawColorXXX() -> DrawPrimitives::setColorXXX()


ricardoquesada:
ignore.

Issue #2892, opened 2013-06-14
CCNotificationCenter behavior do not match it's documentation


gkosciolek:
I observed that CCNotificationCenter uses "object" parameter in addObserver and postNotification methods. The documentation said that this is: "The extra parameter which will be passed to the callback function.". 

However the actual behavior is different. The "object" parameter is used like notification sender in Apple's NSNotificationCenter - it is not just object passed to notification. It is actually used to determine if the notification will be sent or not. When you pass an "object" when registering to notification this means that you want to receive this notification but only from the "object" you passed. When you pass NULL it means that you want to receive this notification from all senders. This idea allows you for grater flexibility it is also more natural when you have iOS background with NSNotificationCenter.

The problem with current cocos2d-x CCNotificationCenter is that it only implement the idea above partially: during postNotification when the object, if not NULL, is used to determine if the notification should be sent or not. You cannot however register two notifications that only differs by "object" parameter. 

That is why I created a pool request #2565 that did two things:
1.  Fixed addObserver and observerExisted because it didn't allowed to add observers to notifications that only differs by sender
2. Updatet the API documentation and parameter name from "object" to "sender" to reflect it's true purpose.

gkosciolek:
New pool request on develop branch #3285

gkosciolek:
Pool request #3285 has been merged by dumganhar.

PR #3285, opened 2013-07-26
Fix for #2892.

gkosciolek:
1. Updatet the API documentation and parameter name from "object" to "sender" to reflect it's true purpose.
2. Fixed addObserver and observerExisted because it didn't allowed to add observers to notifications that only differs by sender


dumganhar:
Thanks.

PR #3287, opened 2013-07-26
Fix build with -Werror=format-security (default in Android NDK r9)

rudchenkos:
Hi,

I have made a few fixes to make the library compile on the newest (-r9) Android NDK. Would be happy if these gets upstream.

Thanks!


minggo:
Thanks.
Fixed conflict in #3288.

PR #3300, opened 2013-07-27
issue #2430: Moving enum type to class for CCBReader.

dumganhar:


extensions/CCBReader/CCBValue.h
```diff
@@ -12,39 +12,31 @@
...
-    {
-        int nValue;
-        float fValue;
-    } mValue;
+public:
     
-    std::string _value;
-    Array* _arrValue;
-    int mType;
+    enum Type
```
minggo, line 30: Why not use `enum class`?

PR #3306, opened 2013-07-28
fix crash issue in CCArray::initWithObjects()

timothyqiu:
The `initWithObjects()` function always crashes.

```
Array *arr = new Array();
arr->initWithObjects(foo, bar, nullptr);
```

In the implementation of `initWithObjects()`, `data` is set to `nullptr` by `ccArrayFree()`, but never gets allocated by `ccArrayNew()`. So the following calls to `addObject()` will crash due to the null pointer.


timothyqiu:
Oops, I think I altered the behavior when `pObject` is `nullptr`. I'll fix it right now...

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/5423. 
Thanks.

PR #3332, opened 2013-07-30
Clang support for Linux

azmeuk:
Allows to compile with clang instead of gcc by using `CLANG=1` makefile option. Hides warnings from Box2D since it is an external lib.


sergey-shambir:
Nice thing, seems to be OK codewise. It also useful for QtCreator project part to automatically use clang for both game and cocos libs if user selected this compiler; I'll update cocos2dx.pri later.

PR #3341, opened 2013-07-31
QtCreator project part: extended Linux detection.

sergey-shambir:
Should fix potential "Cannot recognize platform" errors.

QtCreator (qmake build system) have 'specs' to select platform and compiler, typical spec name look like `linux-g++-64`, `unsupported/linux-clang`, `win32-msvc-2010` and so on. New platform condition uses more precise detection with `*linux-*` instead of just `linux`, it fixes issue that I previously found on some QtSDK version.


1755:
Need edit box2d.prf too

sergey-shambir:
Thanks, updated commit.

PR #3346, opened 2013-08-01
undefined argument should not be counted as a valid argument

nakosung:


dumganhar:
Thanks.
But when you pass an `undefined` to the second argument, it will go to the logic of 

```
std::vector<std::string> protocols;
....
....
....
cobj->init(*delegate, url, &protocols);
```

protocols will be an empty vector.
But in my opinion, the better way is to modify `init` method in `WebSocket.cpp`

```
    int protocolCount = 0;
    if (protocols && protocols->size() > 0)
    {
        protocolCount = protocols->size();
    }
    else
    {
        protocolCount = 1;
    }

    _wsProtocols = new libwebsocket_protocols[protocolCount+1];
    memset(_wsProtocols, 0, sizeof(libwebsocket_protocols)*(protocolCount+1));

    if (protocols)  // Here should be ------>>>>  if (protocols && protocols->size() > 0)
    {
        int i = 0;
        for (std::vector<std::string>::const_iterator iter = protocols->begin(); iter != protocols->end(); ++iter, ++i)
        {
            char* name = new char[(*iter).length()+1];
            strcpy(name, (*iter).c_str());
            _wsProtocols[i].name = name;
            _wsProtocols[i].callback = WebSocketCallbackWrapper::onSocketCallback;
        }
    }
    else
    {
        char* name = new char[20];
        strcpy(name, "default-protocol");
        _wsProtocols[0].name = name;
        _wsProtocols[0].callback = WebSocketCallbackWrapper::onSocketCallback;
    }
```

nakosung:
:) In either way, I just want to get rid of crashes while toying around with node.js websocket-stream module.

nakosung:
It turns out that your suggestion is the only solution, my 'patch' didn't work.

nakosung:
Ooops. the last commit is not related to this topic. That was my mistake because I'm not familiar with pull-requesting. Sorry! BTW, adding cc.RemoveSelf to js binding doesn't kill anybody. So it is safe to merge with.

dumganhar:
Thank you, I will check it.

nakosung:
Yeah, I will be happy with whatever solves the problem. :)

I learned that relying on argc would be erroneous. 
1. 8. 1., 오후 6:02, James Chen notifications@github.com 작성:

PR #3347, opened 2013-08-01
More documentation fixes

timothyqiu:
- Use C++ code in C++ comments.
  - `YES`/`NO` to `true`/`false` (e.g. Returns YES if blah blah blah...)
  - Change sample code to C++ code in documents. (see cocos2dx/actions/CCActionTween.h)
  - Function reference (e.g. `setTexture:withRect` to `setTextureWithRect`)
- Removed Hungarian notation from parts of the code.
- Display sample code as code in documentation. (Surround code in document with `code`/`endcode` tags.)
- Fixed tag typo: `deprecatd` -> `deprecated`
- Changed some doc due to the previously changed source code.
- Removed some trailing whitespaces.

### DIFF
--- a/cocos2dx/actions/CCActionInterval.h
+++ b/cocos2dx/actions/CCActionInterval.h
@@ -461,7 +461,7 @@ public:
 	virtual JumpTo* reverse(void) const override;
 };
 
-/** @typedef bezier configuration structure
+/** Bezier configuration structure
  */
 typedef struct _ccBezierConfig {
     //! end position of the bezier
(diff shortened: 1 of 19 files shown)


timothyqiu:
It seems that the Travis CI bot failed the first job due to a connection error to github :(

sergey-shambir:
Btw, you can check this change using "-Wdocumentation" flag in clang (supported since clang 3.2 in official build, and probably since 4.0 in clang from XCode distribution).

It should find mismatched parameters, as minimum.

minggo:
Thanks @timothyqiu.

@sergey-shambir said

> > Btw, you can check this change using "-Wdocumentation" flag in clang (supported since clang 3.2 in official 
> > build, and probably since 4.0 in clang from XCode distribution).
> > 
> > It should find mismatched parameters, as minimum.

Good to know.

PR #3349, opened 2013-08-01
A better way to locate project creator

timothyqiu:
The use of `os.getcwd()+'/tools/project_creator/'` to locate project creator is problematic. This assumes that the user is executing the script from the cocos2d-x root directory:

```
$ pwd
/home/timothy/repo/cocos2d-x
$ ./create-multi-platform-projects.py  # Okay
$ cd ..
$ cocos2d-x/create-multi-platform-projects.py  # Failed
```

The last try will produce '': No such file or directory".

I changed the way to locate cocos2d-x root directory by using the path of the script itself.

p.s. Also, `os.path.join()` is safer than connect strings manually.

### DIFF
--- a/create-multi-platform-projects.py
+++ b/create-multi-platform-projects.py
@@ -5,5 +5,6 @@ import os
 from tools.project_creator import create_project
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


minggo:
Thanks.

PR #3354, opened 2013-08-02
s3tc compressed texture support

godyZ:
s3tc compressed texture decode supported in ios android win32 osX


sergey-shambir:
Code that decodes S3TC textures is under LGPL license. So cocos2dx with this change inside should be distributed under LGPL license, that's inconvient for iOS where dynamic linking is hard.

Is there another S3TC decoder under BSD-like license? (cocos2dx itself uses MIT)

minggo:
@thanks @sergey-shambir.
I will checkout it.

PR #3355, opened 2013-08-02
closed #2434: simply Android.mk

minggo:


folecr:
(Cross posting from discussion in #3357)

Using shell commands in a makefile is fine if we are building on a known system. However Cocos is built on multiple platforms (Linux, OSX, Windows) each of which could have different shells (bash, Cygwin bash) and with different versions of find, with different line endings (windows vs. unix line endings), etc. The situation is further complicated by user customizations. For example, some users will change the default shell to 'C' shell which will affect the shell command in GNU Make.

Is it necessary to auto generate file dependencies? Yes it is tedious to maintain lists of source files in Android.mk but it is also simple.It is better than having to debug problems in makefiles caused by unknown platforms.

So, I would recommend keeping makefiles as dumb lists.

If we do need to auto-generate makefiles we should use a platform that is designed to work the same across different environments - such as Python + GyP.

minggo:
@folecr 
I think you are right.
We should use less platform dependent codes.

PR #3357, opened 2013-08-03
Simpler, more modern Android port of cocos2d-x

folecr:
The goals of this port
- Remove Java/JNI calls in the rendering and event loops (performance critical portions of the code)
- Improve the perceived lag in event (touch, accelerometer) processing. When playing games the event processing feels less laggy!
- Improve readability and maintainability by converting to C++/C where possible. Hopefully this will allow more of the cocos community to contribute to the Android port!

Features
- Rendering uses the NDK API
- Removal of JNI interaction with rendering loop allows future rendering optimizations to be easily shared with Android
- Android lifecycle is managed with the NDK API
- Event processing receives events directly from the event sink in native code (no round trip via Dalvik)
- Lifecycle, rendering, event processing all happen on the same Linux thread
- Updated the CocosDenshion implementation (more below)

Major to do items, in order of priority : 
- Lua projects to be updated
- Text input is broken.
- Open GL initialization should be done directly by cocos2d::Director instead of being initialized in the Android platform-specific code (thanks @minggo for review)
- CocosDenshion uses the Java API for sound and music playback on all devices. This needs to be fixed to use the Open SL API on devices where the Java API is known to perform badly.
- CocosDenshion resource loading should be multi-threaded
- A more C++-like design for the Android code : currently it uses almost all C code for the Android port.

Tests
- Cpp and JS projects build successfully
- Games (SimpleGame, CocosDragon, MoonWarriors play successfully)
- Ran through TestCpp and TestJavaScript.


folecr:
Please review before merging if possible.
(@minggo you have already reviewed my earlier changes and I have incorporated some of your feedback. thanks!)

I will work on the to do items in the following weeks.

minggo:
Awesome job, @folecr.
Because i simplified Android.mk in #3355, it cause conflicts.
This pull request can not be merged automatically.

folecr:
I could ... but I have a comment about #3355.

Using shell commands in a makefile is fine if we are building on a known system. However Cocos is built on multiple platforms (Linux, OSX, Windows) each of which could have different shells (bash, Cygwin bash) and with different versions of find, with different line endings (windows vs. unix line endings), etc. The situation is further complicated by user customizations. For example, some users will change the default shell to 'C' shell which will affect the shell command in GNU Make.

Is it necessary to auto generate file dependencies? Yes it is tedious to maintain lists of source files in Android.mk but it is also simple.It is better than having to debug problems in makefiles caused by unknown platforms.

So, I would recommend keeping makefiles as dumb lists.

If we do need to auto-generate makefiles we should use a platform that is designed to work the same across different environments - such as Python + GyP.

(I will post the same comment on #3355 if you would like to discuss there.)

minggo:
@folecr 
I think you are right.
I will use python to auto-generate Android.mk.

Because this pull request is big, and it is easy to cause conflict with other pull requests, so could you please resolve  the conflicts first? I think i have to merge this pull request first. And you can improve it if needed.

What's your opinion?

folecr:
@minggo updated branch so it can be automatically merged. thanks!

folecr:
@minggo regarding #3355 
- we could start by creating a python script that is run manually by developers who change cocos2d-x files or by CocosRobot. This script will generate a makefile that is committed into cocos.
- once we are sure that the script works correctly, then we can decide if we should remove all the committed makefiles and auto-generate the makefiles for all developers.

This will allow us to experiment with the scripts and maybe try a tool like GyP https://code.google.com/p/gyp/ without affecting the developers who need to use cocos2d-x.

minggo:
@folecr
I agree with you.

folecr:
I've updated the branch to merge cleanly. I am investigating some errors I see when using the latest NDK (r9)
I'll send further patches tomorrow.

folecr:
Ah .. nevermind. The errors were not with the latest NDK but simply a merge problem.

folecr:
Rebased! Merge tested and merges cleanly on develop.
@minggo please note the changes to cocos2dx/Android.mk

This pull request should now be good for merge. Thanks.

dumganhar:
@folecr 
It seems to be the problem refer to http://stackoverflow.com/questions/14187970/fatal-error-android-native-app-glue-h-no-such-file-or-directory .

folecr:
@dumganhar I fixed it by rebasing correctly. It should be fine now.

dumganhar:
OK, I will check it. Is it late at night there? You're such a hard working guy. :+1:

folecr:
Just trying to get the Android port in shape :-)

minggo:
@folecr 
I merged it since it is a big commit, and it is an important commit.
You can continue to improve it in other pull request.

Thanks.

folecr:
Perfect! Thanks.

I'll send a PR for the updated templates in a few minutes. After that the commits intended for this change should be complete.

Improvements may continue after...

PR #3359, opened 2013-08-03
Fixed the crash caused by improper deletion of VBOs in ParticleSystemQuad

dotsquid:
Here is a description of the problem and a simple solution I found. http://www.cocos2d-x.org/boards/6/topics/27145?r=32896
Here is a test project which demonstrates the problem (should be placed in $COCOS2DX_HOME/projects/). https://docs.google.com/file/d/0B1F8RGhitNUxQ3RxdVpDYjZqOUU/edit?usp=sharing

### DIFF
--- a/cocos2dx/particle_nodes/CCParticleSystemQuad.cpp
+++ b/cocos2dx/particle_nodes/CCParticleSystemQuad.cpp
@@ -596,6 +596,7 @@ void ParticleSystemQuad::setBatchNode(ParticleBatchNode * batchNode)
             CC_SAFE_FREE(_indices);
 
             glDeleteBuffers(2, &_buffersVBO[0]);
+            memset(_buffersVBO, 0, sizeof(_buffersVBO));
 #if CC_TEXTURE_ATLAS_USE_VAO
             glDeleteVertexArrays(1, &_VAOname);
             GL::bindVAO(0);


cocos2dx/particle_nodes/CCParticleSystemQuad.cpp
```diff
@@ -596,6 +596,7 @@ void ParticleSystemQuad::setBatchNode(ParticleBatchNode * batchNode)
             CC_SAFE_FREE(_indices);
 
             glDeleteBuffers(2, &_buffersVBO[0]);
+            memset(_buffersVBO, 0, sizeof(_buffersVBO));
 #if CC_TEXTURE_ATLAS_USE_VAO
             glDeleteVertexArrays(1, &_VAOname);
             GL::bindVAO(0);
```
dumganhar, line 602: Hey, @dotsquid 
Thanks for the patch.
But if you test your demo project by XCode, you will find that it crash also after this fix.
The reason is that `_VAOname` also needs to be set to zero after `GL::bindVAO(0)`.

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/3369
Thanks.

dotsquid:
It seems that the crash you are talking about is fixed by this patch https://github.com/cocos2d/cocos2d-x/pull/3144, isn't it?

dumganhar:
@dotsquid 
No, that patch #3144 didn't fix this issue. 
We should reset `_VAOname` to `0`.
Please refer to my PR https://github.com/cocos2d/cocos2d-x/pull/3369

dotsquid:
Ah, I see. Yep, you are right.

tmsoft:
There are more places than just the CCParticleSystemQuad that might need this fix.  I've applied it to 3 places in CCParticleSystemQuad and in CCTextureAtlas.  Basically any place that has glDeleteVertexArrays(1, &m_uVAOname);

dotsquid:
Two other places are the destructors and ParticleSystemQuad::setupVBOandVAO, I guess?
There is no reason to reset _VAOname-s there, since:
1) the destructors will ...uhm... destruct those objects, and their _VAOname-s won't be accessible;
2) in ParticleSystemQuad::setupVBOandVAO _VAOname is overwritten by glGenVertexArrays two lines below glDeleteVertexArrays.

dumganhar:
@tmsoft
Yep, I agree with @dotsquid , I have checked that. It's not needed.

tmsoft:
Your right, I'm being a little compulsive over this blowup as it hasn't completely gone away for me.  Although I think the particles might be fixed, I am still seeing a few crashes in sprite batch nodes.  In my code, I'm doing a lot of sprite add/deletes on the same batch node along with particle systems.  None of the stack traces help at all as it's always memory cleanup related.  And it's always on Android and not iOS making it even more difficult to debug.  Have you heard of anything similar happening with the CCSpriteBatchNode?  And is there any patches that I may have missed?  I'm using 2.1.4 zip download.  Cheers

PR #3363, opened 2013-08-05
make CCBAnimationManager::getSequenceId public; add method CCBAnimationManager::getSequenceDuration to get duration of timeline;

tiantian20007:
It's common case to check whether a timeline exist by name; So CCBAnimationManager::getSequenceId should be public;

Adding the CCBAnimationManager::getSequenceDuration method to get duration of exist timeline is pretty usefully when we make our animation all with cocosbulder.

Yes, there already is a method CCBAnimationManager::getSequence(int nSequenceId), then we can call CCBSequence::getDuration to get that info. However, CCBSequence is not binded for JS, we need a more simple interface to get that info for JS.

### DIFF
--- a/extensions/CCBReader/CCBAnimationManager.cpp
+++ b/extensions/CCBReader/CCBAnimationManager.cpp
@@ -262,6 +262,14 @@ CCBSequence* CCBAnimationManager::getSequence(int nSequenceId)
     return NULL;
 }
 
+float CCBAnimationManager::getSequenceDuration(const char *pSequenceName)
+{
+    int id = getSequenceId(pSequenceName);
+    if (id != -1)
+        return getSequence(id)->getDuration();
+    return 0;
+}
+
 
 void CCBAnimationManager::moveAnimationsFromNode(Node* fromNode, Node* toNode) {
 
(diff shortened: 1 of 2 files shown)


dumganhar:
@tiantian20007 
Thanks for your contribution.
There is an issue I don't quite understand. That's you said:

> Adding the CCBAnimationManager::getSequenceDuration method to get duration of exist timeline is pretty usefully when we make our animation all with cocosbulder.

Could you demonstrate it about how to use these two interface in your game?

tiantian20007:
In my game, all animation of characters is made in cocosbuilder. Different action of one character is separated by different timelines; I call CCBAnimationManager::runAnimationsForSequenceNamed to run those animations. However, without check the existence of a timeline, calling runAnimationsForSequenceNamed with non exist timeline name will meet a CCAssert. I don't want my game to stop in that case. So I use getSequenceId to checkout whether a timeline exist before use it;
Then, for half real time Turn-based game, the duration of each timeline is essential to calculate the delay and sequence for concurrent actions;

dumganhar:
Thanks.

PR #3368, opened 2013-08-05
Multi-resolution emscripten html template.

azmeuk:
Default emscripten html template update.
Minor emscripten warnings fix, and minor emscripten Makefile refactoring.


dumganhar:
Thanks.

PR #3371, opened 2013-08-05
issue #2433: Updating bindings-generator, adding lua target. It's not finished.

dumganhar:


dumganhar:
Closed it firstly, I'll send another PR after we finish it.

PR #3376, opened 2013-08-06
Update templates to match newer, mostly-NDK Android port

folecr:


folecr:
@minggo @dumganhar Should be able to enable template test builds in CI now!

minggo:
@folecr 
We have just disable template checking in CI:)

dumganhar:
@folecr 
Do you agree to enable ci-build for templates?
@minggo  thinks that we only need to enable it when we will release a new version.
What's your opinion?

folecr:
I think it is fine to enable template builds. It is good to show users of cocos that template builds are being checked regularly.
But... the Android build already takes a long time in Travis (more than 15 minutes.) If we could move it to another build machine it would be ideal.

dumganhar:
@folecr 
Yep, Android build takes a long time. I have updated the travis script that making a `soft link` to make all android projects use the same `objs`. That means `cocos2dx`, `extensions`, `CocosDenshion` only need to be built once.
That have saved a lot of building time. 

I think moving to another building machine is not a good idea.
The better way maybe enabling `parallels` build for android, like `make -j8}`.
I did try that, but when linking, there will be lots of logs output, it's more than `10000` lines which travis-ci will treat it as errors.

Is it a way to enable parallels build but not output so many logs when linking?
You could test it in TestJavascript by `build_native.sh -p`

template/multi-platform-cpp/proj.android/project.properties
```diff
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-10
+target=android-13
```
minggo, line 11: Why change to `android-13`?

template/multi-platform-cpp/proj.android/src/nojava.txt
```diff
@@ -0,0 +1 @@
+No Java files
```
minggo, line 1: What this file used for?

folecr, line 1: There are no Java files. However, if the src/ directory does not exist then the Android build (via build.xml) will fail. Also, Git does not allow empty directories.

So, adding a text file allows : src/ directory to exist without Java files and build to succeed always.

template/multi-platform-cpp/proj.android/project.properties
folecr, line 11: android-13 is required to build (not to run)
It is required to build with support for the configuration changes screenSize, smallestScreenSize that were added in API level 13.

See http://developer.android.com/guide/topics/manifest/activity-element.html and search for "Added in API level 13"

folecr:
@dumganhar 

> Yep, Android build takes a long time

The correct way to fix this : generate libraries from cocos2d-x shared code and cocos2d-x platform/android code.
Then link with these libraries for each app. This will fix the long build times and make distribution of cocos2d-x easier and make it easy for people to use cocos2d-x.

The problems
- Android does not support source+binary libraries fully in their build system ... they have publicly stated they will support this in the future. (no time frame given)
- Because the Android build system does not support libraries, we have to do dependency management ourselves. This might cause build errors unles libraries are manually re-built by the developer
- ... because it requires manual rebuilds this might cause more confusion than today's build system.
- Also, cocos2d-x has multiple configurations... Chipmunk vs. Box, JS vs Lua that are set at build time. When we can move these configuration options to runtime (via inheritance in C++ or dynamic linking at library load time) it will be easier to use cocos2d-x as a prebuilt library.

Even with these problems I think it is worthwhile to build cocos2d-x as a binary library. We should work on it.

template/multi-platform-cpp/proj.android/project.properties
minggo, line 11: Does it affect the application which will be run on Android 2.3?

minggo:
@folecr 

> >  I think it is worthwhile to build cocos2d-x as a binary library.

Did you mean generate .so?

template/multi-platform-cpp/proj.android/project.properties
minggo, line 11: What will happen if missing android-13 SDK?

template/multi-platform-cpp/proj.android/src/nojava.txt
minggo, line 1: So add this comment into this file?

folecr, line 1: Good idea. Will do. Thanks!

template/multi-platform-cpp/proj.android/project.properties
folecr, line 11: android-13 is required for building (ie, developer should have it installed.) It is not required for running (Android 2.3 is still the minimum.)

folecr:
@minggo 

> > I think it is worthwhile to build cocos2d-x as a binary library.
> 
> Did you mean generate .so?

.a, with headers.

The final .so to be generated by the developer using cocos as a library.
I will create a changeset to clarify ... We can discuss there.

template/multi-platform-cpp/proj.android/project.properties
minggo, line 11: So please modify README.mdown to mention it.

PR #3378, opened 2013-08-06
[Tizen] fix acceleration wrong coordinates.

pyrasis:
- fix acceleration coordinates.
- change between x and y.
- invert the sign of y.


dumganhar:
Thanks

PR #3380, opened 2013-08-06
Run script from string

alenl2:
I use this so i can reload my JS files directly from server(Websocket or xmlhttprequest) i makes quick fixes much quicker :)


nakosung:
You can do that within javascript by calling 'eval(javascriptSourceCode)'. BTW, AFAIK, spider monkey has a performance issue with eval'ed code comparing to compiled code..

alenl2:
i was thinking the same thing but it dos not work. if i use eval() you get errors like 
Cocos2d: JS: ....../jsb_cocosbuilder.js:74:TypeError: _ccbGlobalContext[documentControllerName] is not a constructor

and i'm thinking i dident do the pull request right... right?

nakosung:
Disclaimer: I'm not a collab of cocos2d-x.

'eval' isn't different from executeScript. You can check it within console of your browser. Are you requiring jsb.js in your 'eval'?

alenl2:
yes i do. i use cocosbuilder generated files

Issue #3089, opened 2013-07-05
CCMenu Issue


valentiplunge:
I created a CCMenu object and added some CCMenuItems as children with Z increasing in every CCMenuItem, something like:

addChild(CCMenuItem, 1)
addChild(CCMenuItem, 2)
...

But when I touched them the priority was inverted, when I touched the CCMenuItem with z:2 I activated the CCMenuItem with z:1. For additional information the first CCMenuItem was a background and the second one was a button over it. I looked up in the code of CCMenu and I think it is fault of a CCARRAY_FOREACH that should be CCARRAY_FOREACH_REVERSE.

Thank you.

metalbass:
I can reproduce this. It's easy to notice when the items overlap.

valentiplunge:
Yes, I will try to remember it.

Thank you very much.

godyZ:
:)

godyZ:
The issue is migrated to http://www.cocos2d-x.org/issues/2711

PR #3396, opened 2013-08-08
Fixing CCMenu's Z ordering

metalbass:
Implementing this issue: https://github.com/cocos2d/cocos2d-x/issues/3089


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/CCMenu.cpp#L523
Thanks.

PR #3398, opened 2013-08-08
exposed AssetsManager to javascript and added multiple-assetsManager support

nakosung:
1. You can use AssetsManager within Javascript.
2. Multiple AssetsManagers support.
   - Script package / Resource package / Promotion package / ...

### DIFF
--- a/extensions/network/WebFile.cpp
+++ b/extensions/network/WebFile.cpp
@@ -23,16 +23,39 @@ private:
 private:
 	std::mutex _mutex;
 	std::list<cb_pair_t> _queue;
+    size_t _pending_requests;
+    bool _scheduled;
 
 public:
+    Caller()
+    : _pending_requests(0), _scheduled(false)
+    {}
+    
+    // curl thread
 	void requestCall(fn_cb_t callback, bool result)
 	{
 		_mutex.lock();
 		_queue.push_front(cb_pair_t(callback, result));
-		doScheduled();
 		_mutex.unlock();
 	}
+    
+    // main thread
+    void increment()
+    {
+        if (++_pending_requests) {
+            doScheduled();
+        }
+    }
+    
+    // main thread
+    void decrement()
+    {
+        if (_pending_requests-- == 0) {
+            doUnscheduled();
+        }
+    }
 
+    // main thread communicates with curl thread via _queue
 	void update(float dt)
 	{
 		_mutex.lock();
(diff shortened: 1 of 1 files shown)


nakosung:
1. Cocos2d-x has nice javascript binding and awesome built-in assetsManager support, but there wasn't jsb for assetsManager so I added it!
2. Having two or more assetsManager to maintain assets with several packages wasn't working. The version key was shared among assetsManagers, so I added suffix(simple string hash) on key names.
3. Deleted `#include` . There was no need to include curl on header.
4. Made AssetsManager derived from Node class to handle object life cycle gracefully within Javascript. (Just attach your assetManager to your layer or something else)

minggo:
Thanks @nakosung.
Please use [new cocos2d-x coding style](http://www.cocos2d-x.org/projects/cocos2d-x/wiki/Cocos2d_C++_Coding_Style).
- don't use Hungarian style, such as `pDirector`
- key_of_version -> keyOfVersion

nakosung:
I updated my code applying coding convention. (The code containing hungarian style was not mine. It was just copied-and-pasted from AssetsManagerTest project.)

minggo:
Thanks.

extensions/AssetsManager/AssetsManager.cpp
```diff
@@ -604,4 +631,68 @@ void AssetsManager::Helper::handleUpdateSucceed(Message *msg)
...
+        ErrorCallback errorCallback;
+        ProgressCallback progressCallback;
+        SuccessCallback successCallback;
+    };
+
+    auto* manager = new AssetsManager(packageUrl,versionFileUrl,storagePath);
+    auto* delegate = new DelegateProtocolImpl(errorCallback,progressCallback,successCallback);
+    manager->setDelegate(delegate);
+    manager->_shouldDeleteDelegateWhenExit = true;
+    return manager;
```
minggo, line 658: 1. As the rule in cocos2d-x, static function `create` should return an autorelease object.
2. Why not save std::function as member variable of AssetsManager instead of creating a temple class `DelegateProtocolImpl`?

nakosung, line 658: 1. Okay.
2. To fix like your suggestions there will be (..somewhat..) big change of the interface. Is it all right? I was trying to minimize the change. What is your recommendation?
   - delete `::_delegate`, add several `std::function<>` and save `_delegate::*` into those `std::function<>`. (This will prevent to change api)
   - keep my change (This also keep the api same)

nakosung:
Last commit is related with sample project[AssetsManagerTest].

nakosung:
I failed to separate 'web socket support' from this pull request. The last commit (790ba7f) isn't related with this AssetsManager PR.

nakosung:
Sorry. I forgot to mention 790ba7f contains patch which was made by Steve Longhurst. http://www.cocos2d-x.org/boards/20/topics/32130?r=33880

nakosung:
_bump_ hello!

nakosung:
moved to https://github.com/cocos2d/cocos2d-x/pull/3548

dumganhar:
> moved to #3548 

Thanks.

dumganhar:
Could you also separate the bindings of AssetMangager to another Pull Request?

nakosung:
I'm afraid that adding JSB is related to so many files and commits because it breaks sample project. :( If you really want it, I should re-write new PR for it.

PR #3399, opened 2013-08-08
use cocos2d-x c++ style

minggo:
1. func( xxx ) -> func(xxx)
2. include lib header before other header files
3. func(void) -> func()
4. my_var -> myVar
5. NULL -> nullptr
6. add a space after `if`
7. put `{` in a new line


minggo:
It is a continuous work.

gelldur:
why you dont use autoformatter  like astyle?

PR #3404, opened 2013-08-09
Documentation fix: broken links and wrong encoding in the doc...

timothyqiu:
- Use C++ function names instead of Obj-C names when referencing functions.
- Changed some GB2312 encoded characters to UTF-8. （GB2312 encoded `’` is not valid UTF-8）
- Changed two `XXX class for cocos2d for iphone` description to `'XXX class for cocos2d-x'`.
- For `Object`, `AutoreleasePool`, and `PoolManager`:
  - Documentation added.
  - Applied the cocos2d-x C++ style guide.
  - Fixed some wrong indent.


dumganhar:
Thanks.

PR #3408, opened 2013-08-10
[Tizen] Update project files

pyrasis:
- include stdint.h
- set etc, s3tc header path
- exclude CCImageCommonWebp.cpp, include etc, s3tc directories.
- update project files.


dumganhar:
Thanks

PR #3410, opened 2013-08-11
Fixes full paths on Android

NatWeiss:
If one adds a search resolutions order directory without a trailing slash, CCFileUtils::getFullPathForDirectoryAndFilename will fail on Android. For example it will return:

assets/sdBg1.png

Instead of the expected:

assets/sd/Bg1.png

This small commit fixes the function to add a slash if necessary, fixing the issue on Android.

### DIFF
--- a/cocos2dx/platform/CCFileUtils.cpp
+++ b/cocos2dx/platform/CCFileUtils.cpp
@@ -785,7 +785,11 @@ void FileUtils::loadFilenameLookupDictionaryFromFile(const char* filename)
 
 std::string FileUtils::getFullPathForDirectoryAndFilename(const std::string& strDirectory, const std::string& strFilename)
 {
-    std::string ret = strDirectory+strFilename;
+    std::string ret = strDirectory;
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/CCFileUtils.cpp
```diff
@@ -785,7 +785,11 @@ void FileUtils::loadFilenameLookupDictionaryFromFile(const char* filename)
 
 std::string FileUtils::getFullPathForDirectoryAndFilename(const std::string& strDirectory, const std::string& strFilename)
 {
-    std::string ret = strDirectory+strFilename;
+    std::string ret = strDirectory;
+    if( strDirectory.size() && strDirectory[strDirectory.size()-1] != '/' )
+		ret += '/';
+	ret += strFilename;
+
```
dumganhar, line 792: Thanks, but could you fix the indention? We're using 4 spaces for indention.

NatWeiss:
I fixed the indentation in this pull:
https://github.com/cocos2d/cocos2d-x/pull/3447

Note that the code around it used tabs for indentation, so I used that style and added comments.

PR #3411, opened 2013-08-11
fixed stroke font color with channel color values other than 255

maciekczwa:
When stroke color is set to values like #C261BD it becomes black due to bad casting.


dumganhar:
Thanks.

dumganhar:
Did you try it on android emulator?
With this patch, device can work now but emulator couldn't.
Could you find out the reason?
Thanks.

PR #3416, opened 2013-08-12
Android gdb script for TestCpp

folecr:
Also : build Android's TestCpp for both x86 and ARM.


samples/Cpp/TestCpp/proj.android/project.properties
```diff
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-17
+target=android-14
```
minggo: Does the version should be android 14?
It causes the android build version confused.
What version should be for cocos2d-x?

folecr: You are right that it's confusing. And I have updated the pull request to match other projects i.e. they all use android-13

(The reason I made the change for both minimum-sdk and target-sdk was to avoid compilation errors from the platform headers which were not conforming to C++11 standards. This change was not important.)

PR #3418, opened 2013-08-12
Qt5 update

sergey-shambir:
Updated pull request, whole list of changes:
- Build fixes for recent changes
- Added mouse and keyboard events handling
- Added looping and gain effect to CocosDenshion
- Added support for Qt compiled with '-opengl desktop' configure option; Qt allows to check with which OpenGL platform it was compiled using macro QT_OPENGL_ES_2, so we include GLES headers on Qt compiled with '-opengl es2' and GL headers on Qt compiled with '-opengl desktop'. After this change I was able to compile and run Qt5 port on Ubuntu 13.10 (Ubuntu uses '-opengl desktop' on all non-ARM architectures).

However, Qt5 port still fails with shader compilation error on my Intel HD 3000 videocard (while Linux native port works OK). Qt5 port works on my secondary videocard (NVIDIA GeForce GT 540M).


sergey-shambir:
I'm working on bigger update.

PR #3422, opened 2013-08-13
Major clean up of the (still experimental) new Label system

carlomorgantinizynga:
Major clean up of the (still experimental) new Label system + new tests suite which emulates the "old" tests using the new Label.

### DIFF
--- a/cocos2dx/label_nodes/CCFont.cpp
+++ b/cocos2dx/label_nodes/CCFont.cpp
@@ -88,7 +88,9 @@ void Font::setCurrentGlyphCollection(GlyphCollection glyphs, const char *customG
 const char * Font::getCurrentGlyphCollection()
 {
     if (_customGlyphs)
+    {
         return _customGlyphs;
+    }
     else
     {
         return getGlyphCollection(_usedGlyphs);


cocos2dx/label_nodes/CCFont.cpp
```diff
@@ -25,8 +25,86 @@
...
+}
+
+const char * Font::getCurrentGlyphCollection()
+{
+    if (_customGlyphs)
+        return _customGlyphs;
+    else
+    {
+        return getGlyphCollection(_usedGlyphs);
+    }
```
minggo, line 97: I think it is better to include `{}` for `if` block because `else` has `{}`.

cocos2dx/label_nodes/CCFont.h
```diff
@@ -26,31 +26,54 @@
...
     virtual Rect                  getRectForChar(unsigned short theChar);
     
-    virtual unsigned short int  * getUTF16Text(const char *pText, int &outNumLetters);
     virtual int                   getUTF16TextLenght(unsigned short int *pText);
+    virtual unsigned short int  * getUTF16Text(const char *pText, int &outNumLetters);
     virtual unsigned short int  * trimUTF16Text(unsigned short int *pText, int newBegin, int newEnd);
     
+protected:
+    
+    Font();
```
minggo, line 65: No implementation of `Font`?

cocos2dx/label_nodes/CCFont.h
```diff
@@ -26,31 +26,54 @@
...
+    virtual ~Font() {}
+    void setCurrentGlyphCollection(GlyphCollection glyphs, const char *customGlyphs = 0);
+    const char * getGlyphCollection(GlyphCollection glyphs);
+    
+private:
+    
+    GlyphCollection     _usedGlyphs;
+    char              * _customGlyphs;
```
minggo, line 73: These two member variables are not initialized.

cocos2dx/label_nodes/CCFont.h
carlomorgantinizynga, line 65: Not sure why you don't see the implementation of the constuctor Font() {} in CCFont.cpp (it should be there)

cocos2dx/label_nodes/CCFont.cpp
carlomorgantinizynga, line 97: fixed it

cocos2dx/label_nodes/CCFont.h
carlomorgantinizynga, line 73: they are in the constructor for Font() which is at the very top of CCFont.cpp

minggo:
Compiling error on Android and linux.
I will fix it in another pull request.

minggo:
I tried to fix compiling errors, but i couldn't, because there is a logical error. If you add `CCFontCache.cpp/.h` into Xcode project, you can find it easily.

The error caused by Android is
- Don't remove unneeded files in cocos2dx/Android.mk
- should use `#include "support/ccUTF8.h"` in CCFontFreeType.cpp

cocos2dx/label_nodes/CCFont.h
minggo, line 65: Sorry about it.
I just see the changed codes.

cocos2dx/label_nodes/CCFont.h
minggo, line 73: Yep, ,my fault.

dumganhar:
It seems that `CCFontCache` isn't needed any more. Right? If so, why not to delete the source files.

dumganhar:
BTW, in cocos2d-x 3.0, we don't need `CC` prefix now.

PR #3423, opened 2013-08-14
Explicitly choose debugging data format version for Android in debug mode ...

folecr:
... to work around a bug introduced in the Android NDK toolchain.


folecr:
@dumganhar Any idea why the Travis server hung on this for >50 minutes? Thanks.

dumganhar:
It's limited by travis. The time of every build should not more than a hour.
- James
  ÔÚ 2013-8-16 ÉÏÎç3:56£¬"Surith Thekkiam" notifications@github.comÐ´µÀ£º

folecr:
The Android build used to take about 15 minutes though. Any idea why it is taking more than double the time?
I just rebased the commit in the hope that it will trigger a Travis rebuild. I'll watch to see how long it takes now.

dumganhar:
It seems that you have added `x86` support for `TestJavascript` which depends on `cocos2dx`, `extension`,`chipmunk`, `CocosDenshion`. Therefore, it will build all of this x86 libraries.
In the `tools/travis-scripts/run-script.sh`, it makes a directory called `android_build_objs` at the root folder of `cocos2d-x`. 

<pre>
build_android()
{
    echo "Current dir: `pwd`"
    pushd $1/proj.android
    ln -s $COCOS2DX_ROOT/android_build_objs obj
    ./build_native.sh
    popd
}
</pre>

And the building order is

<pre>
    # Create a directory for temporary objects
    mkdir android_build_objs
    # Build samples
    echo "Building samples ..."
    cd $COCOS2DX_ROOT/samples/Cpp
    build_android HelloCpp
    build_android TestCpp
    build_android AssetsManagerTest
    cd $COCOS2DX_ROOT/samples/Javascript
    build_android TestJavascript  # it will build x86 libraries and maybe replace the built arm libraries, I will check that.
    build_android CocosDragonJS
    build_android CrystalCraze
    build_android MoonWarriors
    build_android WatermelonWithMe
    cd $COCOS2DX_ROOT/samples/Lua
    build_android HelloLua
    build_android TestLua
</pre>

dumganhar:
Does it have to enable `x86` architecture for android? It will double the building time of android.

folecr:
> Does it have to enable x86 architecture for android?

Two reasons to do so 
1. Android x86 emulator is much more usable than ARM emulator (similar to iOS development.) Especially when debugging - debugging is incredibly slow with the ARM emulator and is fast with the x86 emulator.
2. There are many x86 Android devices on the market. It will be good to catch incompatibility with these devices.

My opinion is that yes, x86 builds are necessary. I realize that we may not have enough CI resources for this at this time. I will leave it up to you to decide.

dumganhar:
> 1. Android x86 emulator is much more usable than ARM emulator (similar to iOS development.) Especially when debugging - debugging is incredibly slow with the ARM emulator and is fast with the x86 emulator.

I agree. But is it the reason to enable `x86` architecture by default? Yep, for convenience, we add `x86` APP_ABI. But very time it will build `arm` architecture too. It also waste time for debugging on x86 emulator. Right?

> 1. There are many x86 Android devices on the market. It will be good to catch incompatibility with these devices.

If developer has gotten a `x86` device, he surely know that how to enable x86 architecture. Do you mean that developers who are developing x86 application will don't know how to enable it?

But that's only my opinion, if we have to enable `x86`, I think the `travis-ci` script should disable it before buiding.

minggo:
>  I think the travis-ci script should disable it before building

Yes, we should.

folecr:
Sure thing : I'll send a pull request to remove x86.

folecr:
Closing this pull request until Android fixes bugs preventing debugging on ARM devices https://code.google.com/p/android/issues/detail?id=58373
OR
we enable testing of x86 builds of cocos2d-x.

PR #3428, opened 2013-08-16
issue #2533:add ATITC comressed texture (*.ktx file) support

godyZ:
Soft decode test in Win32, ios, Mac. Device decode test in HTC G14
Adreno 220 GPU.

### DIFF
--- a/cocos2dx/platform/third_party/common/atitc/atitc.cpp
+++ b/cocos2dx/platform/third_party/common/atitc/atitc.cpp
@@ -1,5 +1,5 @@
 /****************************************************************************
- Copyright (c) 2010 cocos2d-x.org
+ Copyright (c) 2013 cocos2d-x.org
  
  http://www.cocos2d-x.org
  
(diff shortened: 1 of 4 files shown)


cocos2dx/Android.mk
```diff
@@ -127,6 +127,7 @@ textures/CCTextureAtlas.cpp \
 textures/CCTextureCache.cpp \
 platform/third_party/common/etc/etc1.cpp\
 platform/third_party/common/s3tc/s3tc.cpp\
+platform/third_party/common/atitc/atitc.cpp\
```
dumganhar, line 130: It's better to add a space before backslash.

cocos2dx/platform/CCImageCommon_cpp.h
```diff
@@ -594,6 +640,9 @@ bool Image::isPvr(const unsigned char * data, int dataLen)
     }else if (isS3TC(data, dataLen))
     {
         return Format::S3TC;
+    }else if (isATITC(data, dataLen))
```
dumganhar, line 643: According cocos2d's code style, it's better to add a space before `else if`.

godyZ, line 643: ok, i will update that right now

PR #3431, opened 2013-08-16
add iterator for Array and make Array can be used in range-based loop

minggo:


samples/Cpp/TestCpp/Classes/SpriteTest/SpriteTest.cpp
```diff
@@ -243,7 +243,7 @@ void Sprite1::addNewSpriteWithCoords(Point p)
 
 void Sprite1::ccTouchesEnded(Set* touches, Event* event)
 {
-    for (auto &item: *touches)
```
ricardoquesada, line 246: I guess it should be `&item`, and not `item`... it is faster.

PR #3435, opened 2013-08-18
language option is invalid print mesage

shinriyo:
If we input invalid mistook language option


tools/project_creator/create_project.py
```diff
@@ -71,6 +71,10 @@ def checkParams():
         context["src_project_name"] = "HelloJavascript"
         context["src_package_name"] = "org.cocos2dx.hellojavascript"
         context["src_project_path"] = os.path.join(template_dir, "multi-platform-js")
+    else:
+        print "Your language parameter doesn\'t exist." \
```
minggo, line 75: Would you please use print(), because this script can run on python2.7 and python3.0.

PR #3437, opened 2013-08-19
Update common.pri

slug404:
in /usr/include/tiffio.h not define tmsize_t


sergey-shambir:
Thanks! Can you tell please, which Linux distribution you've used? Did you installed "libtiff-dev" package or run "install-deps-linux.sh" script?

PR #3438, opened 2013-08-19
Carray with std vector

ricardoquesada:
Implements `cocos2d::Array` with  `std::vector<>` using `RCPtr` plus other minor fixes.

Other minor changes:
- `lastObject`, `objectAtXXX` were deprecated and added the `get` version of it.


ricardoquesada:
I'll send a better one later.

PR #3440, opened 2013-08-20
Better array

ricardoquesada:
Adds support for `std::vector + RCPtr<Object>` in `CCArray`.

Why would we want it ?
Basically it allow us to make performance tests on CCArray.

Last, but not least, the sideeffect is that now `CCArray` has a cleaner API.

ah... additionally all getters with the `get`, were deprecated, and the `get` version was added.

### DIFF
--- a/cocos2dx/base_nodes/CCNode.cpp
+++ b/cocos2dx/base_nodes/CCNode.cpp
@@ -697,7 +697,7 @@ void Node::detachChild(Node *child, bool doCleanup)
 void Node::insertChild(Node* child, int z)
 {
     _reorderChildDirty = true;
-    ccArrayAppendObjectWithResize(_children->data, child);
+    _children->addObject(child);
     child->_setZOrder(z);
 }
 


samples/Cpp/TestCpp/Classes/SchedulerTest/SchedulerTest.cpp
```diff
@@ -695,11 +695,11 @@ void SchedulerUpdate::onEnter()
 void SchedulerUpdate::removeUpdates(float dt)
 {
     auto children = getChildren();
-    Node* node;
 
     for (auto c : *children)
     {
-        node = static_cast<Node*>(c);
+        auto obj = static_cast<Object*>(c);
```
minggo, line 701: Don't need to convert to `Object*`.

cocos2dx/touch_dispatcher/CCTouchDispatcher.cpp
```diff
@@ -290,9 +304,10 @@ TouchHandler* TouchDispatcher::findHandler(Array* pArray, TouchDelegate *pDelega
     return NULL;
 }
 
-void TouchDispatcher::rearrangeHandlers(Array *pArray)
+void TouchDispatcher::rearrangeHandlers(Array *array)
 {
-    std::sort(pArray->data->arr, pArray->data->arr + pArray->data->num, less);
+    std::sort(array->data->arr, array->data->arr + array->data->num, less);
```
minggo, line 309: So can only work when using ccArray in Array?

ricardoquesada, line 309: Apparently the iterator that you added for `Array` doesn't work with

``` c++
std::sort( std::begin(*array), std::end(*array), less);
```

Not sure why :-( ... but that is something that needs to be fixed.

minggo, line 309: Ok, i will fix it after merging.

PR #3442, opened 2013-08-20
Update CCScrollView.cpp

xbruce:
~ScrollView _touches is null when developer using cc.ScrollView.extend in JSB to create a CCScrollView


dumganhar:
Thanks.

PR #3443, opened 2013-08-20
合并原作者修改

HoGarfield:


minggo:
Could you please don't use Chinese.
And could you please divide this pull request into some small ones?

PR #3455, opened 2013-08-22
Dictionary array fixes

ricardoquesada:
In many places `Dictionary` and `Array` are not being initialized. In fact `Dictionary` doesn't  have the `init()` method creating potential leaks.
Also in objects like `Armature` and the new `LabelTTF`, the `Array` object is not being used a pointer. So it doesn't use the 2 phase initialization, creating potential leaks.

This patch fixes all those issues.

### DIFF
--- a/cocos2dx/cocoa/CCArray.h
+++ b/cocos2dx/cocoa/CCArray.h
@@ -50,10 +50,10 @@ class RCPtr
 public:
 	//Construct using a C pointer
 	//e.g. RCPtr< T > x = new T();
-	RCPtr(T* ptr = NULL)
+	RCPtr(T* ptr = nullptr)
     : _ptr(ptr)
 	{
-        if(ptr != NULL) {ptr->retain();}
+        if(ptr != nullptr) {ptr->retain();}
 	}
 
 	//Copy constructor
@@ -69,13 +69,13 @@ public:
     : _ptr(ptr._ptr)
 	{
 //        printf("Array: Move Constructor: %p\n", this);
-        ptr._ptr = NULL;
+        ptr._ptr = nullptr;
 	}
 
 	~RCPtr()
 	{
 //        printf("Array: Destructor: %p\n", this);
-        if(_ptr != NULL) {_ptr->release();}
+        if(_ptr != nullptr) {_ptr->release();}
 	}
 
 	//Assign a pointer
@@ -87,8 +87,8 @@ public:
         //The following grab and release operations have to be performed
         //in that order to handle the case where ptr == _ptr
         //(See comment below by David Garlisch)
-        if(ptr != NULL) {ptr->retain();}
-        if(_ptr != NULL) {_ptr->release();}
+        if(ptr != nullptr) {ptr->retain();}
+        if(_ptr != nullptr) {_ptr->release();}
         _ptr = ptr;
         return (*this);
 	}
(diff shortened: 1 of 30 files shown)


cocos2dx/cocoa/CCArray.h
```diff
@@ -397,53 +397,18 @@ class CC_DLL Array : public Object, public Clonable
...
-    typedef ArrayIterator const_iterator;
-    iterator begin();
-    iterator end();
-
-#endif
-    
-
-
-public:
-#if CC_USE_ARRAY_VECTOR
```
minggo, line 492: Why remove ArrayIterator?

cocos2dx/cocoa/CCArray.h
```diff
@@ -397,53 +397,18 @@ class CC_DLL Array : public Object, public Clonable
...
     std::vector<RCPtr<Object>> data;
+
 #else
+    Object** begin() { return &data->arr[0]; }
+    Object** end() { return &data->arr[data->num]; }
```
minggo, line 411: add begin() and end() can be used in range-based loop, but is not enough for std::sort().

cocos2dx/cocoa/CCArray.h
ricardoquesada, line 492: It is not needed any more. It was replaced by these 2 functions:

``` c++
  Object** begin() { return &data->arr[0]; }
  Object** end() { return &data->arr[data->num]; } 
```

And they work with `std::sort( begin(array), end(array), xxx)`

cocos2dx/cocoa/CCArray.h
ricardoquesada, line 411: why not ? what's the problem with it?

cocos2dx/cocoa/CCDictionary.cpp
```diff
@@ -358,12 +359,18 @@ Object* Dictionary::randomObject()
...
-        pRet->autorelease();
+        ret->autorelease();
     }
-    return pRet;
+    return ret;
+}
+
+bool Dictionary::init()
+{
+    retain();
```
dumganhar: Why to add `retain` in `init` method?  It will increase the reference count . it may cause `memory leak` issues.

ricardoquesada: ouch... it should be `autorelease()`... good catch... sending a new commit

ricardoquesada: Actually it should be empty... because the `autorelease` is on `create`

cocos2dx/cocoa/CCArray.h
minggo, line 411: It's my fault. It can work.
Great.

It makes things simple.

cocos2dx/cocoa/CCArray.h
minggo, line 492: Yep. It is great.

ricardoquesada:
Please don't merge it yet... working on a bug fix with the current PR. thanks.

minggo:
It conflicts with #3454, but your method is better.
So would you please fix the conflict by using your codes in Array.h?

ricardoquesada:
Done!
fixed bug and fixed conflicts.

cocos2dx/cocoa/CCDictionary.cpp
```diff
@@ -383,9 +389,7 @@ void Dictionary::acceptVisitor(DataVisitor &visitor)
 
 Dictionary* Dictionary::createWithContentsOfFile(const char *pFileName)
 {
-    Dictionary* pRet = createWithContentsOfFileThreadSafe(pFileName);
-    pRet->autorelease();
-    return pRet;
+    return createWithContentsOfFileThreadSafe(pFileName);
 }
```
dumganhar, line 393: It will not return an autorelease object in `createWithContentsOfFileThreadSafe`. Therefore, it's `ThreadSafe` in some cases.
But `Dictionary::createWithContentsOfFile` should return an `autorelease` object.

ricardoquesada, line 393: interesting.
So probably `createWithContentsOfFile()` should not call `createWithContentsOfFileThreadSafe()`.
what do you think ?

dumganhar, line 393: The difference between `createWithContentsOfFile` and `createWithContentsOfFileThreadSafe` is just `autorelease`.
So if we don't invoke `createWithContentsOfFileThreadSafe` in `createWithContentsOfFile`, there will be some duplicated codes.

PR #3456, opened 2013-08-22
[NaCL] Add missing include of <sys/mount.h>

sbc100:
This was previously included by nacl_io.h but
upstrean changes in NaCl removed this.


dumganhar:
Thanks.

PR #3460, opened 2013-08-22
更新

Aresky:


minggo:
Thanks.

Could you please
- don't use Chinese in pull request title and codes
- please refer to the [contributing to cocos2d-x](http://www.cocos2d-x.org/projects/cocos2d-x/wiki/Contribution)
- please add meaningful commit message

PR #3471, opened 2013-08-23
Support cocos2d-x 3.0 in plugin

natural-law:
@folecr
This pull request is used to discuss the support of cocos2d-x 3.0 in plugin.


folecr:
@natural-law Thanks! Are there instructions on building the plugin samples? I wasn't able to build the Android project by running build_native.sh.

runOnUiThread should still be available from NativeActivity. Google has an example of integrating with Unity that uses callbacks which will schedule code to be run on the Ui thread (and hence all plugin Views and event handling should work)

The link is [here](https://code.google.com/p/google-mobile-dev/source/browse/Android/plugin-library/src/com/google/unity/AdMobPlugin.java?repo=admob-unity-plugin)

Could you please try to modify the source code for the PluginWrapper by scheduling the callback using Activity's runOnUiThread? I'll try and get the plugin samples building and try out AdMobs.

natural-law:
@folecr 
Thanks for your suggestion. I will try the way you suggested.

To build the plugin samples, you should run the script : plugin/tools/publish.sh at first.
I'm sorry for my mistake. I'll modify the build_native.sh of the plugin samples. Add commands to invoke the publish.sh.

I'll inform you if I have any improvement. Thank you once more.

natural-law:
@folecr 
The way you suggested is worked. I add an interface to init plugin with NativeActivity.
Now the initialization & invoking interfaces of plugins are worked well.

But I encounter another problem. When the native callback methods in plugin are invoked, there is a lot of log like this:

<pre>
08-26 17:02:13.298: E/dalvikvm(16835): JNI ERROR: env->self != thread-self (0x40e99078 vs. 0x4002d010); auto-correcting
</pre>

Whether it's all right or a hidden danger? What should I do for this?

BTW, if I invoke MessageBox() in the native callback method. The demo will be crashed. 
Because the value of variable sCocos2dxHelperListener in class Cocos2dxHelper is null.

folecr:
It looks like JNI is calling into the UI thread with the JNIEnv from the other (native activity) thread. The solution is to attach to the VM from the UI thread.
http://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/invocation.html

Taking a look,it looks like initialization of the app is not working for me and is crashing. I'll look into this some more to see what is broken.
P.S. Here is my branch with your changes and applied on top of cocos2d/cocos2d-x/develop
https://github.com/folecr/cocos2d-x/commits/natural-law_plugin

minggo:
@folecr
If a JNI is calling from other thread but gl thread, how can make the JNI be called in gl thread?

minggo:
The JNI is from java to c++.

folecr:
@minggo The easiest would be to generate a custom APP_CMD that is processed in the thread's loop here https://github.com/cocos2d/cocos2d-x/blob/develop/cocos2dx/platform/android/nativeactivity.cpp#L491

I'll check to see why plugin sample is crashing then I'll add a helper function to do this.

(I hope I understood your question correctly.)

folecr:
@natural-law This is the stack trace I see when running HelloPlugins https://gist.github.com/folecr/6360536

The Dalvik VM rejects Cocos2dxHelper https://gist.github.com/folecr/6360536#file-gistfile1-txt-L9 but it looks like calls using this class succeed in other applications. I confirmed with TestCpp.

It also rejects PluginWrapper https://gist.github.com/folecr/6360536#file-gistfile1-txt-L17 but it looks like a JNI call further down fails.

folecr:
Do you see any similar crashes on your device?

minggo:
> > @minggo The easiest would be to generate a custom APP_CMD that is processed in the thread's loop here https://github.com/cocos2d/cocos2d-x/blob/develop/cocos2dx/platform/android/nativeactivity.cpp#L491
> > I'll check to see why plugin sample is crashing then I'll add a helper function to do this.
> > (I hope I understood your question correctly.)

Yep, it is my question.
And i think it should be easier for developers to generate a custom APP_CMD, for example, don't have to modify engine codes to handle this command.

natural-law:
@folecr 
I just had a test in my environment. The Dalvik VM rejects class Cocos2dxHelper & PluginWrapper too. The log is here : https://gist.github.com/natural-law/6361450

I don't know why the two classes are rejected in HelloPlugins. But it's not the cause of the crash. Because the sample running well on my device.

From the stack trace, I think the reason is Java class PluginWrapper was not found. You can try to run script plugin/tools/publish.sh again. Then rebuild the sample project.

Also, I will take a test on other computer to make sure all files needed had been pushed into my repository.

folecr:
Yes - I just dexdump-ed the classes and PluginWrapper is missing! I'll try building again.

folecr:
Could you add a readme/documentation on running the plugin sample?

natural-law:
@folecr 
I'm sorry there isn't any document for running the plugin sample. But I can show you a simple steps to run the sample:
1. Run plugin/tools/publish.sh to publish plugins (You should input the path of android NDK, android SDK & Apache Ant at the first time you run it)
2. After step 1, you should got a directory plugin/publish generated. (There are java libs, C++ libs and other resources in the directory)
3. Run the build_native.sh in the sample project.
4. Import android project : plugin/publish/plugins/nd91/android/DependProject into Eclipse. (The sample project depend on it). Build the project.
5. Import android project : cocos2dx/platform/android/java into Eclipse. Build the project.
6. Import the sample project into Eclipse. Build it in Eclipse. Then you can debug the sample.

Could you tell me what your problem is?

folecr:
Thanks for the build instructions. One of the libraries was missing
https://github.com/folecr/cocos2d-x/commit/5b11a385668dd0ba44e777f25e218c791c5a66fa

natural-law:
@folecr 
Hi. Can the sample run on your device? Have you encountered the log I mentioned before : 

<pre>
JNI ERROR: env->self != thread-self (0x40e99078 vs. 0x4002d010); auto-correcting
</pre>

I think you have a misunderstanding of my description. The log is shown when C++ methods are invoked from Java.
So I have the same question as @minggo. How can make a C++ JNI method be invoked in GL thread?

Thanks!

folecr:
No - I haven't encountered the log. The plugin sample runs but none of the plugin libraries such ast twitter, etc. are included in the build so i see failures when calls are made.

I'll reproduce the error above separately and fix it.

natural-law:
So, can my pull request be merged? I mean the changes of file nativeactivity.cpp in my pull request.

If this pull request will be merged, please wait for a minute. I should modify the main.cpp in all samples of engine.
Because of a parameter was added in method cocos_android_app_init().

Please let me know if the error is fixed. Thanks very much!

folecr:
Those changes look good! Please go ahead with your changes to add the
function signature change to all samples.

I would suggest separating the pull requests into two : one for the changes
above and a separate one for the plugin specific changes. This way the
change history will be clear.

On Aug 29, 2013, at 20:45, Bin Zhang notifications@github.com wrote:

So, can my pull request be merged? I mean the changes of file
nativeactivity.cpp in my pull request.

If this pull request will be merged, please wait for a minute. I should
modify the main.cpp in all samples of engine.
Because of a parameter was added in method cocos_android_app_init().

Please let me know if the error is fixed. Thanks very much!

—
Reply to this email directly or view it on
GitHubhttps://github.com/cocos2d/cocos2d-x/pull/3471#issuecomment-23538996
.

natural-law:
I closed this pull request. There is a new pull request to modify the method cocos_android_app_init() in engine :
https://github.com/cocos2d/cocos2d-x/pull/3528

PR #3484, opened 2013-08-26
Override LayerMultiplex::init() 

ledyba:
LayerMultiplex::create() method calls LayerMultiplex::init(),
but LayerMultiplex::init() is not overridden,
so LayerMultiplex::create() returns invalid object( _layers is not initialized).

In this pull request, I overrode init method.


minggo:
Thanks.

PR #3496, opened 2013-08-28
closed #2450: Fixing a bug that HttpClientTest crashes on Mac

### LINKED ISSUE #2450: LuaJIT bytecode parsing error on Android device
I found there's a parsing error when I was porting a project using LuaJIT bytecode on Android device:  
////////////////////////////////////////////////////////////////////////
D/cocos2d-x debug info: [LUA ERROR](binary): cannot load malformed bytecode
////////////////////////////////////////////////////////////////////////
But the same LuaJIT bytecode work perfectly on iOS.

I turn back to vanilla Lua bytecode after I got this error(Of couse I've changed the default LuaJIT module back to Lua). Still error appear on Android device:
////////////////////////////////////////////////////////////////////////
D/cocos2d-x debug info: [LUA ERROR] binary string: unexpected end in precompiled chunk
////////////////////////////////////////////////////////////////////////

I'm wondering whether it's relevant to the difference between CCLuaEngine::executeString and CCLuaEngine::executeScriptFile(In some case that work, like quick-cocos2d-x).

dumganhar:
And adding WebsocketTest and SocketIOTest for Mac port.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ExtensionsTest/ExtensionsTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/ExtensionsTest/ExtensionsTest.cpp
@@ -10,7 +10,7 @@
 #include "ComponentsTest/ComponentsTestScene.h"
 #include "ArmatureTest/ArmatureScene.h"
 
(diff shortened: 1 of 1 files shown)


dumganhar:
This PR will cause HttpClientTest crash on IOS.

PR #3503, opened 2013-08-29
Iss2460 glfw refractor

dabingnn:


sergey-shambir:
CI build failed because builtin GLFW distribution was not detected correctly. Try to test build on local machine after removing glfw3 package from system.

dabingnn:
Thank you @sergey-shambir ! I replaced "./install_glfw.sh" with "$DIR/install_glfw.sh". It works.

ricardoquesada:
cool.
I have just tested it on linux and it doesn't link.
It seems that libglfw3.a was not added into the repo.

dabingnn:
Thank you, Riq
Thank you,
I can not find glfw3 library on online linux repo, currently version are glfw2. So I wrote a bash "install_glfw.sh" in directory tools/travis-scripts. The bash will download source and build library on linux.
I forgot to include the bash into "install-deps-linux.sh". I will include it. Currently, you can run the "install_glfw.sh", then the samples should link correctly.
Please tell me if any problem happens.

ricardoquesada:
Thanks. I've just tested it.
A few comments about it:
- The script apparently needs `curl` so the package `curl` needs to be installed (I did it manually).
- And probably the `install_glfw.sh` script should be merged with `~/cocos2d-x/install-linux-deps.sh`. Our users don't need to run 2 scripts to run cocos2d in linux.
- And after installing the new lib, `ldconfig` should be run, otherwise it won't find the new installed lib.

After doing that, it worked Ok for me. Thanks!

ricardoquesada:
BTW, GLFW 3.0.2 was released a few days ago... it would be great to download it instead of v3.0.1.

thanks!

dabingnn:
@ricardoquesada 
Thank you for your review. Here is my pull request!
[https://github.com/cocos2d/cocos2d-x/pull/3606](https://github.com/cocos2d/cocos2d-x/pull/3606)

I fixed some bug for `glfw`, and added glfw installation script.
I am not familiar with linux shell. 
Could you please review the scripts changing for me? The changes for scripts are listed as follows:
1. `install_glfw.sh` are shared by linux and our `travis scripts`, so we cannot merge it into `install-linux-deps.sh` currently.
2. `curl` has been add in the script.
3. `sudo ldconfig` are called after `sudo make install`.
4. `GLFW` version has been changed from `3.0.1` to `3.0.2` .

ricardoquesada:
It looks good! thanks!
I tested it today and it worked as expected.

PR #3505, opened 2013-08-29
add include/openssl *.h files into libcrl/include

pktangyue:
In my cocos2d-x project, I use [sqlcipher](http://sqlcipher.net/) to encrypt my sqlite3 db file.

There was not too much trouble in iOs, but I met big problem in Android.

In its `sqlite3.c`, it has some codes from openssl. For example:

```
#include <openssl/rand.h>
#include <openssl/evp.h>
#include <openssl/hmac.h>
```

While I make, it showed me error message: 'No such file: openssl/rand.h'.

Then I tried to import openssl from  [platform_external_openssl](https://github.com/android/platform_external_openssl), but failed.

And I post two questions about it.

https://github.com/minggo/libcurl-build/issues/2

http://www.cocos2d-x.org/boards/6/topics/34356

After many tries, I have solved it.

Since [libcurl-build](https://github.com/minggo/libcurl-build) built the libcurl with openssl, what I need to do is just add the [`include/openssl`](https://github.com/minggo/libcurl-build/tree/master/openssl-android/include/openssl) into cocos2d-x source code.

And it works! It's simple!


minggo:
No, i don't think cocos2d-x need to add these header files. You should add these files in your own project.

pktangyue:
Ok, I understand. But if anyone want to use `openssl`, I think there should be somewhere to tell him, "you just need to add this header files in your cocos2d-x project."

minggo:
Yep. You should add your own header files if you want to integrate a 3rd lib.

PR #3506, opened 2013-08-29
Label ttf set dimensions

imbahom:
If i set dimension's width or height to a negative number , it will crash in file CCImage.mm .

unsigned char\* data = new unsigned char[(int)(dim.width \* dim.height \* 4)];

### DIFF
--- a/cocos2dx/base_nodes/CCNode.cpp
+++ b/cocos2dx/base_nodes/CCNode.cpp
@@ -72,7 +72,7 @@ CCNode::CCNode(void)
 , m_nZOrder(0)
 , m_pChildren(NULL)
 , m_pParent(NULL)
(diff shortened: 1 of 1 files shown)


cocos2dx/label_nodes/CCLabelTTF.cpp
```diff
@@ -86,6 +86,8 @@ CCLabelTTF* CCLabelTTF::create(const char *string, const char *fontName, float f
                                const CCSize &dimensions, CCTextAlignment hAlignment, 
                                CCVerticalTextAlignment vAlignment)
 {
+    if (dimensions.width > 0 || dimensions.height > 0 )
+        return;
```
imbahom, line 90: sorry , i was sleepy while committing
should be  return nullptr;

PR #3507, opened 2013-08-30
revert back fps interval to 0.1s

ricardoquesada:
Why was it set to 2 seconds?


minggo:
I am not sure. May be some mistaken.

PR #3508, opened 2013-08-30
Fixing up for the 10.2 version of BB10.

dducharme:
- fixes the template file which was not compiling
- Lua support was not compiling
- added in .ogg for the background test in the CocosDenshionTest for Lua/HelloLua/Resources/hello.lua


dumganhar:
Thanks. @dducharme , we will release cocos2d-x v2.2 today, I will check this issue soon.

dumganhar:
Hi @dducharme ,
We have removed all other templates except `multi-platform-xxx`. So the fix for blackberry template is not needed now.
Also, I have fixed the compilation errors yesterday. Please check again by pull the latest codes on master branch.
Thanks for the comment of `CocosDenshionTest` of TestLua. I also added it in another PR(https://github.com/cocos2d/cocos2d-x/pull/3831).

PR #3511, opened 2013-08-31
Ooops, reverting optimization for C

ricardoquesada:
still valid for Neon, but the current optimization is not valid for C.

Please, apply this patch otherwise cocos2d won't work correctly
on simulator, windows, linux, mac, etc...

### DIFF
--- a/cocos2dx/kazmath/src/mat4.c
+++ b/cocos2dx/kazmath/src/mat4.c
@@ -218,8 +218,9 @@ kmMat4* const kmMat4Multiply(kmMat4* pOut, const kmMat4* pM1, const kmMat4* pM2)
 {
 #if defined(__ARM_NEON__)
 
-    float mat[16];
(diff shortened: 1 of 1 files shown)


dumganhar:
It's because in 

```
kmMat4* const kmMat4LookAt(kmMat4* pOut, const kmVec3* pEye,
                     const kmVec3* pCenter, const kmVec3* pUp)
{
...
    kmMat4Translation(&translate, -pEye->x, -pEye->y, -pEye->z);
    kmMat4Multiply(pOut, pOut, &translate);  // Use pOut as input and output value.

    return pOut;
}
```

ricardoquesada:
Thanks. This code was also triggering the bug

``` c
void kmGLMultMatrix(const kmMat4* pIn)
{
    lazyInitialize();
    kmMat4Multiply(current_stack->top, current_stack->top, pIn);
}
```

PR #3513, opened 2013-08-31
AddChild test fix

ricardoquesada:
`sortAllChildren` should not be part of the test.
It is tested separatedly on another test


minggo:
I added missing head in #3516.

ricardoquesada:
ok. thanks.

PR #3515, opened 2013-08-31
Master perftest fixes

ricardoquesada:
Improves performance tests for 'master'.

Now 'master' and 'develop' have the same perf tests.


minggo:
I added missing head in #3519.

PR #3521, opened 2013-09-01
fixed playing >30 samples via Open SL

kompjoefriek:
Fixed playing >30 different samples via OpenSL on Android (which is used
by the Galaxy S2)

Tested by running game with about 48 samples, running over an hour.
(Cocos2d-x 2.1.5 on Nexus 4)
Please test thoroughly

### DIFF
--- a/CocosDenshion/android/opensl/OpenSLEngine.cpp
+++ b/CocosDenshion/android/opensl/OpenSLEngine.cpp
@@ -202,10 +202,12 @@ static EffectList& sharedList()
 	return s_List;
 }
 
+// list to keep track of CallbackContexts that need to be removed.
+// removing them in another thread (like during the callback) will cause problems.
 static vector<CallbackContext*>& removalList()
 {
-	static vector<CallbackContext*> s_List;
-	return s_List;
+	static vector<CallbackContext*> s_ContextList;
+	return s_ContextList;
 }
 
 unsigned int _Hash(const char *key)


minggo:
Thanks.
@folecr
Would you please help to review it?

folecr:
@kompjoefriek What is the original bug? - Is it a limitation of Android Open SL or is it a bug in Samsung firmware?

If it is indeed a limitation specified by Open SL can we query the system via an API to get maximum number of samples allowed and adjust accordingly?
If it is a bug in Samsung's firmware we should test on an S2 before submitting for merge.

kompjoefriek:
When playing >30 different samples via OpenSL, the 31th and up sample could not be created because OpenSL only allows 30 AudioPlayers to exist at one time. The 31th and up sample will just not play because not AudioPlayer could be created.

This change allows OpenSL to Create and Destroy the AudioPlayers when needed, having 30 AudioPlayers at most.
It seems to be an Android OpenSL limitation, and I could not find any way to query the maximum number of allowed players.
I could remove the hard-coded number 30 by just trying to create a new AudioPlayer, free-up space and retry when it fails. But that will result in a lot of memory errors in the LogCat.

This still should be tested on a Galaxy S2 before merging. (sorry, I don't have any available)

folecr:
> It seems to be an Android OpenSL limitation, and I could not find any way to query the maximum number of allowed players.
> I could remove the hard-coded number 30 by just trying to create a new AudioPlayer, free-up space and retry when it fails. 

This seems like the right way to do it.

> But that will result in a lot of memory errors in the LogCat.

Other than looking ugly, does this matter?

> This still should be tested on a Galaxy S2 before merging. (sorry, I don't have any available)

Yes, I think this is critical. Unless we find documentation from Samsung about what works and what does not, this seems necessary.

kompjoefriek:
I tried a solution where i would delete an AudioPlayer each time it failed to create a new one (and then retry to create a new one), but a lot of times it would not play a sound sample when it was so posed to...
Maybe someone else could have go at that, my solution failed. hard.

I've tried to search for any way I could query the available memory or resources OpenSL could use, but no luck (QueryAvailableVoices looks interesting but the engine capabilities interface is not available on android).
So i rewrote the limit of 30 AudioPlayers to a limit of 32 objects, as documented here: http://mobilepearls.com/labs/native-android-api/opensles/
I keep track of the number of objects i create and destroy, this way i know if i need to delete an AudioPlayer first.

I hope this version is good enough for you guys, it's the best i can get it.
We would still need someone with a galaxy s2 to test it properly, although the OpenSL implementation should not be very different from that in my phone (Nexus 4).

Also: the travis build failed, but I can't see if it's because of my changes or something else.

folecr:
@minggo Please review and test on Galaxy S2.

minggo:
Thanks @kompjoefriek and @folecr.

WenhaiLin:
I test this request on Galaxy S2.
It's not crash,but not support play 30 different samples at the same time.The log output "AudioFlinger could not create track, status: -12" at 15th samples.

kompjoefriek:
Thank you for testing @Dhilan007. I do not see any way to fix that in the current implementation. It looks like this requires a rewrite using buffer queues to limit the number of AudioPlayers. (other ideas are welcome)

Thank you @minggo and @folecr for reviewing.
Please reject this pull request. I will start rewriting shortly.

minggo:
Thanks @kompjoefriek.

PR #3528, opened 2013-09-02
Modify the method cocos_android_app_init()

natural-law:
Add parameter in method cocos_android_app_init() to initialize plugin.


folecr:
@natural-law Thanks!
These changes look good to me.

PR #3529, opened 2013-09-02
Update PerformanceNodeChildrenTest.cpp

cristian1980:
Fix compile errors in VS2012


minggo:
I think we should replace `srandom` with `srand`.

cristian1980:
srand() works ok on VS2012.

PR #3531, opened 2013-09-03
add "speed" to CCBAnimationManager, and relative test code

TimothyZhang:
# Purpose

In many situations, we want to change the speed of animations. For example, we may want to have differenct game characters with different walk speeds.

So I added  `_speed` property, which defaults to `1.0f` to CCBAnimationManager. Just call `setSpeed()` before running the animations.
# TestCase

I just reused the testcase of `ExtensionTest/CocosBuilderTest/AnimationsTest`，click any button repeatly would change the speed among [0.25, 0.5, 1.0, 2.0, 4.0]

I've tested with cocos2d-x-2.1.4 on win32,iOS and android
coco2d-x-3.0alphapre is only tested on win32, but it should be platform independent.

### DIFF
--- a/extensions/CCBReader/CCBAnimationManager.cpp
+++ b/extensions/CCBReader/CCBAnimationManager.cpp
@@ -27,7 +27,7 @@ CCBAnimationManager::CCBAnimationManager()
 , _rootContainerSize(Size::ZERO)
 , _delegate(NULL)
 , _runningSequence(NULL)
-
+, _speed(1.0f)
 {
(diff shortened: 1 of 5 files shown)


cocosQA:
Dear TimothyZhang:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6483

PR #3533, opened 2013-09-03
Reimplement EditBox for android

ledyba:
In cocos2d-x in 3, implementation of "EditBox" for Android is broken.

So, I reimplemented that. Here are the points.
### Create Cocos2dxHelperListenerImpl

In version 2, Cocos2dxActivity implements Cocos2dxHelperListener, but that activity does not exist any more.

So I create the classs "Cocos2dxHelperListenerImpl" that implements Cocos2dxHelperListener.

This class is used only in Cocos2dxHelper, so I locate that class in Cocos2dxHelper.java.
### Use std::atomic to send modified text instead of executing function on GLThread.

In version 2, Cocos2dxActivity used GLSurfaceView#queueEvent (in runOnGLThread) to invoke native funtion that update native EditBox object, but in version 3, we do not used GLSurfaceView.

There is no way to invoke any function under cocos2d-x loop thread ( just like Android Handler class), so I use std::atomic to send the modified text to cocos2d-x main loop thread.

thanks.

### DIFF
--- a/extensions/GUI/CCEditBox/CCEditBoxImplAndroid.cpp
+++ b/extensions/GUI/CCEditBox/CCEditBoxImplAndroid.cpp
@@ -236,7 +236,8 @@ void EditBoxImplAndroid::setAnchorPoint(const Point& anchorPoint)
 }
 
 void EditBoxImplAndroid::visit(void)
-{ // don't need to be implemented on android platform.
+{
+	// checking whether user edited the text or not.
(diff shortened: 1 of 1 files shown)


ledyba:
Oops... Travis CI failed...

ledyba:
```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

Compiler has crashed?

PR #3534, opened 2013-09-03
Reimplement EditBox for android

ledyba:
In cocos2d-x in 3, implementation of "EditBox" for Android is broken.

So, I reimplemented that. Here are the points.

( I sent this pull request before, but Travis CI has crashed, so I resend this. )
### Create Cocos2dxHelperListenerImpl

In version 2, Cocos2dxActivity implements Cocos2dxHelperListener, but that activity does not exist any more.

So I create the classs "Cocos2dxHelperListenerImpl" that implements Cocos2dxHelperListener.

This class is used only in Cocos2dxHelper, so I locate that class in Cocos2dxHelper.java.
### Use std::atomic to send modified text instead of executing function on GLThread.

In version 2, Cocos2dxActivity used GLSurfaceView#queueEvent (in runOnGLThread) to invoke native funtion that update native EditBox object, but in version 3, we do not used GLSurfaceView.

There is no way to invoke any function under cocos2d-x loop thread ( just like Android Handler class), so I use std::atomic to send the modified text to cocos2d-x main loop thread.

thanks.

### DIFF
--- a/build/cocos2d_libs.xcodeproj/project.pbxproj
+++ b/build/cocos2d_libs.xcodeproj/project.pbxproj
@@ -46,6 +46,124 @@
 /* Begin PBXBuildFile section */
 		1599607B1811329600C7D42C /* libluajit.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 1599607A1811329600C7D42C /* libluajit.a */; };
 		15996080181132DE00C7D42C /* libluajit.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 1599607F181132DE00C7D42C /* libluajit.a */; };
+		1A2C7872181A11EC004E5527 /* jsb_cocos2dx_auto.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1A2C785E181A11EC004E5527 /* jsb_cocos2dx_auto.cpp */; };
+		1A2C7874181A11EC004E5527 /* jsb_cocos2dx_auto.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1A2C785E181A11EC004E5527 /* jsb_cocos2dx_auto.cpp */; };
+		1A2C7876181A11EC004E5527 /* jsb_cocos2dx_auto.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 1A2C785F181A11EC004E5527 /* jsb_cocos2dx_auto.hpp */; };
+		1A2C7878181A11EC004E5527 /* jsb_cocos2dx_auto.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 1A2C785F181A11EC004E5527 /* jsb_cocos2dx_auto.hpp */; };
+		1A2C787E181A11EC004E5527 /* jsb_cocos2dx_builder_auto.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1A2C7861181A11EC004E5527 /* jsb_cocos2dx_builder_auto.cpp */; };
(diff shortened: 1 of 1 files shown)


folecr:
@ledyba the more intuitive solution might be to pass a message to the cocos thread like discussed briefly here
https://groups.google.com/forum/#!topic/android-ndk/v2OITtaZTes

And also adding a generic API so that multiple parts of the system : text input, plugins, game code, etc. may pass messages to the cocos thread from other threads. If we can use C++ lambdas as the payload of the message the API would be clean.

Would you like to try this for CCEditBox?

ledyba:
Thank you for your comment!

There is no guarantee that std::function can be passed through pipe.
For example, this code always crashes.

``` C++
#include <stdio.h>
#include <unistd.h>
#include <sys/types.h>
#include <iostream>
#include <cstdlib>
#include <functional>

void pushFunction(int fd)
{
    std::string message = "hello";
    std::function<std::string()> f = [message](){
        return message; // message is "copy captured".
    };
    write(fd, &f, sizeof(f));

}

int main()
{
    int     fd[2];
    pipe( fd );

    pushFunction(fd[1]);
    std::function<std::string()>* fptr =
        reinterpret_cast<std::function<std::string()>*>( malloc(sizeof(std::function<std::string()>)) );
    read(fd[0], fptr, sizeof(*fptr));
    std::cout << "Function call: " << (*fptr)() <<std::endl;
    close(fd[0]);
    close(fd[1]);
    free(fptr);
    return 0;
}
```

http://ideone.com/ESrHgm

But, we can make the thread-safe queue with std::deque and std::mutex.
I would like to implement that queue and use for message passing in CCEditBox.
Should that thread safe queue (CCThreadSafeQueue?) be global? Or should that queue be used only in CCEditBox?

folecr:
> But, we can make the thread-safe queue with std::deque and std::mutex.

Yes.

> I would like to implement that queue and use for message passing in CCEditBox.

Great!

> Should that thread safe queue (CCThreadSafeQueue?) be global? Or should that queue be used only in CCEditBox?

Currently there are two use cases that I know of
1. Edit Box
2. Plugins (such as twitter, admobs, etc. that will callback into cocos)

I would suggest adding a `cocos::android::Queue`
with methods
1. for adding a function to this queue to be used by any thread and which will send a message via fd and using the ALooper
2. and to pop a function (cocos thread only) - reading from this queue to be done in nativeactivity.cpp when the fd is notified.

ledyba:
Thank you for your advice!

We cannot use ALooper because ALooper use FileDescriptor, on witch we cannot send std::function.
So, I created "cocos2d::android::Queue" and "cocos2d::android::Executor".

"Queue" is a thread-safe queue. We can send std::string, Object_, Array_, int ,float, std::function<void()> on this queue.CCEditBoxImplAndroid use this.

``` C++
// on remote thread
cocos2d::android::Queue queue;
queue.push( EVENT_ID_SEND_STRING, "string message" ); //you can use any integer as event id.

//on main loop thread.
cocos2d::android::Queue::Payload p;
while( queue.pop(p) ){
    println("message code: %d\n", p.code());
    println("message: %s\n", p.as<std::string>().c_str());
    //-> "message:  string message"
}

```

Executor is a class equavalent to "Handler" in Java. You can execute arbitrary function on cocos2d main loop thread.

Example:

``` C++
// on remote thread.
cocos2d::android::Executor::getInstance()->post([](){
   // this lambda expression runs on main loop thread.
    printf("Hello world on cocos2d thread!");
});
```

And to tell you the truth, both Queue and Executor is available not only android, but also other platforms.

PR #3548, opened 2013-09-06
exposed AssetsManager to javascript and added multiple-assetsManager support

nakosung:
restricted to only files just needed 
1. You can use AssetsManager within Javascript.
2. Multiple AssetsManagers support.
   - Script package / Resource package / Promotion package / ...

### DIFF
--- a/samples/Cpp/AssetsManagerTest/Classes/AppDelegate.cpp
+++ b/samples/Cpp/AssetsManagerTest/Classes/AppDelegate.cpp
@@ -92,7 +92,7 @@ void UpdateLayer::update(cocos2d::Object *pSender)
     pProgressLabel->setString("");
     
     // update resources
-    getAssetsManager()->update();
+    pAssetsManager->update();
     
     isUpdateItemClicked = true;
 }
@@ -114,7 +114,7 @@ void UpdateLayer::reset(cocos2d::Object *pSender)
     system(command.c_str());
 #endif
     // Delete recorded version codes.
-    getAssetsManager()->deleteVersion();
+    pAssetsManager->deleteVersion();
     
     createDownloadedDir();
 }


nakosung:
- copied from earlier closed PR(https://github.com/cocos2d/cocos2d-x/pull/3398)
1. Cocos2d-x has nice javascript binding and awesome built-in assetsManager support, but there wasn't jsb for assetsManager so I added it!
2. Having two or more assetsManager to maintain assets with several packages wasn't working. The version key was shared among assetsManagers, so I added suffix(simple string hash) on key names.
3. Deleted `#include` . There was no need to include curl on header.
4. Made AssetsManager derived from Node class to handle object life cycle gracefully within Javascript. (Just attach your assetManager to your layer or something else)

extensions/AssetsManager/AssetsManager.h
```diff
@@ -41,7 +41,7 @@
  *  The updated package should be a zip file. And there should be a file named
  *  version in the server, which contains version code.
  */
-class AssetsManager
+class AssetsManager : public Node
```
dumganhar, line 44: Why does AssetsMangager inherit from Node?

extensions/AssetsManager/AssetsManager.h
```diff
@@ -185,13 +203,18 @@ class AssetsManager
...
+    void *_curl;
+
     Helper *_schedule;
     unsigned int _connectionTimeout;
     
-    AssetsManagerDelegateProtocol *_delegate; // weak reference
+    AssetsManagerDelegateProtocol *_delegate; 
     
     bool _isDownloading;
+    bool _shouldDeleteDelegateWhenExit;
```
minggo, line 213: Where to set this member variable?

extensions/AssetsManager/AssetsManager.h
nakosung, line 44: In order to manage its life-cycle within Javascript, some ref-cnt scheme should be involved. AFAIK, extending some object from Node and attaching it into scene-graph, life-cycle management comes for free.

extensions/AssetsManager/AssetsManager.h
```diff
@@ -138,6 +146,15 @@ class AssetsManager
...
      */
     friend int assetsManagerProgressFunc(void *, double, double, double, double);
+
+    /** @brief Initializes storage path. 
+     */
+    void createStoragePath();
+
+    /** @brief Destroys storage path. 
+     */
+    void destroyStoragePath();
```
minggo: Should these two functions be public?

extensions/AssetsManager/AssetsManager.h
nakosung, line 44: _shouldDeleteDelegateWhenExit is set by AssetsManager::create function. Is is necessary for deleting anonymous class used by JSB.

extensions/AssetsManager/AssetsManager.cpp
```diff
@@ -604,4 +631,69 @@ void AssetsManager::Helper::handleUpdateSucceed(Message *msg)
...
+
+        virtual void onError(AssetsManager::ErrorCode errorCode) { errorCallback(int(errorCode)); }
+        virtual void onProgress(int percent) { progressCallback(percent); }
+        virtual void onSuccess() { successCallback(); }
+
+    private :
+        ErrorCallback errorCallback;
+        ProgressCallback progressCallback;
+        SuccessCallback successCallback;
+    };
```
minggo, line 651: It is a smart design, so AssetsManagerDelegateProtocol is not needed.
Right?

extensions/AssetsManager/AssetsManager.h
nakosung, line 44: I think 'destroying' storagePath should be public for further management issues. (Not enough disk space…) If some destroying function is to be public, I think that corresponding creating function should be public.

extensions/AssetsManager/AssetsManager.cpp
nakosung, line 651: AssetsManagerDelegateProtocol is needed by previous design. (AssetsManager::setDelegate) It provides just a closure-like thin wrapper.

minggo, line 651: Yep, but we can mark `AssetsManagerDelegateProtocol` as deprecated.

extensions/AssetsManager/AssetsManager.cpp
```diff
@@ -102,6 +109,26 @@ void AssetsManager::checkStoragePath()
     }
 }
 
+// Multiple key names
+static std::string key_with_hash( const char* prefix, const std::string& url )
```
minggo: It is better to rename it to `keyWithHash` to obey cocos2d-x coding style.

extensions/AssetsManager/AssetsManager.h
minggo, line 44: But AssetsManager will delete downloaded package after finished.

samples/Cpp/AssetsManagerTest/Classes/AppDelegate.cpp
```diff
@@ -141,6 +139,14 @@ bool UpdateLayer::init()
 {
     Layer::init();
     
+    /** Creates assets manager */
+    pAssetsManager = new AssetsManager("https://raw.github.com/minggo/AssetsManagerTest/master/package.zip",
+                                       "https://raw.github.com/minggo/AssetsManagerTest/master/version",
+                                       pathToSave.c_str());
```
minggo, line 145: Memory leak, should use corresponding `create` version.

dumganhar:
> In order to manage its life-cycle within Javascript, some ref-cnt scheme should be involved. AFAIK, extending some object from Node and attaching it into scene-graph, life-cycle management comes for free.

Hi, you could use js object's life-cycle to control cpp's.
The way is to generate binding glue codes in another module( I mean using another .ini, probably, tools/tojs/cocos2d_script_control_cpp.ini), then set the `script_control_cpp` section to `yes`.

samples/Cpp/AssetsManagerTest/Classes/AppDelegate.cpp
minggo, line 145: Or should invoke `release` in destructor.

nakosung, line 145: There's no men-leak, because I added assetsManager into scene-graph.

extensions/AssetsManager/AssetsManager.cpp
nakosung: That will be trivial. Okay.

extensions/AssetsManager/AssetsManager.h
nakosung, line 44: I didn't know that, so those functions can be private.

nakosung, line 44: ooopps. I replied to wrong thread. I'll use git-hub client instead of e-mail.

nakosung:
I thought that inventing another life-cycle-management for JSB was over-kill.

extensions/AssetsManager/AssetsManager.cpp
nakosung, line 651: Unlike Javascript's closure, std::function doesn't provide context(this pointer). So C++ programmers will require pass there user_data(void*) also. That makes ::create function much like old C API.

```
void set_callback( std::function<void(void*)> callback, void* user_data );
```

So I think old ProtocolDelegate style API should remain for C++ programmers. new std::function<> create function is only for JSB. If I could extend ProtocolDelegate in Javascript, I didn't write this new function.

samples/Cpp/AssetsManagerTest/Classes/AppDelegate.cpp
```diff
@@ -141,6 +139,14 @@ bool UpdateLayer::init()
...
+    pAssetsManager->setDelegate(this);
+    pAssetsManager->setConnectionTimeout(3);
+    addChild(pAssetsManager);
```
minggo, line 148: Invoke `new` instead of `createxxx` will cause memory leak. AssetsManager will not be released. When invoke `addChild`, the reference count is 2.

nakosung, line 148: Yes, you're right.

minggo, line 148: So, could you please fix the memory leak and conflicts?
Thanks.

nakosung, line 148: fixed https://github.com/nakosung/cocos2d-x/commit/6fc666079a860fa1cebaf593d32a3e2ffaae7203

minggo:
Thanks.
Could you please fix conflicts?

nakosung:
I'm not sure this is the right way to resolve conflicts. :( I resolved conflicts against latest cocos2d-x/develop branch.

minggo:
Yep, i think it is right.

minggo:
There is a compiling error on Android platform, could you please fix it.
Thanks.

minggo:
You can refer to this [url](https://travis-ci.org/cocos2d/cocos2d-x/jobs/11144745)

nakosung:
travis says good to merge :accept:

PR #3549, opened 2013-09-06
Update cocos2d_specifics.cpp

jimmystar:
fix bug about scheduler, if this value is -1, it will work all right in Emulator, but run error in real device, just call one time.


dumganhar:
Thanks.

PR #3553, opened 2013-09-06
Fix a memory leak in Set::removeAllObjects.

BodbDearg:
The set maintains a strong reference to all objects so release() should be called whenever something is removed from the set.

### DIFF
--- a/cocos2dx/cocoa/CCSet.cpp
+++ b/cocos2dx/cocoa/CCSet.cpp
@@ -121,9 +121,9 @@ void Set::removeAllObjects()
         
         tmp = it;
         ++tmp;
-        Object * pObject = *it;
+        Object * obj = *it;
         _set->erase(it);
-        CC_SAFE_RELEASE(pObject);
+        CC_SAFE_RELEASE(obj);
         it = tmp;
     }
 }


cocos2dx/cocoa/CCSet.cpp
```diff
@@ -121,7 +121,9 @@ void Set::removeAllObjects()
         
         tmp = it;
         ++tmp;
+        Object * pObject = *it;
         _set->erase(it);
+        CC_SAFE_RELEASE(pObject);
```
dumganhar: Thanks. But could you don't use `Hungary Naming`.  `Object* obj = *it` will be good.
About the code style for cocos2d-x 3.0, please refer to http://www.cocos2d-x.org/wiki/Cocos2d_c++_coding_style?project_id=cocos2d-x.

BodbDearg: Thanks! I dislike Hungarian notation myself also but was trying to keep the style consistent with some of the other functions in the class which still used it. Nice to see Cocos2dx is moving away from this naming convention!

A fresh commit has been added with the Hungarian notation removed.

dumganhar:
Thanks.

PR #3554, opened 2013-09-06
Fileutils perf improvements

ricardoquesada:
Improves performance when creating `Node` and `Sprite` objects.

--> creates a common platform for apple: `CCFileutils` has only one file
--> prepends '/' to 'key' that are added in `TextureCache`
--> removed unused code
--> optimized `isAbsolutePath`
--> components in Node uses lazy alloc

Overall:
Creating 500 empty sprites went from 4.4s to 0.6s
Creating 500 non empty sprites went from 1.6s to 0.8

### DIFF
--- a/samples/Cpp/TestCpp/Classes/DataVisitorTest/DataVisitorTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/DataVisitorTest/DataVisitorTest.cpp
@@ -71,9 +71,9 @@ void PrettyPrinterDemo::onEnter()
     
     vistor.clear();
     addSprite();
-    dict = TextureCache::getInstance()->snapshotTextures();
-    dict->acceptVisitor(vistor);
-    log("%s", vistor.getResult().c_str());
+//    dict = TextureCache::getInstance()->snapshotTextures();
+//    dict->acceptVisitor(vistor);
+//    log("%s", vistor.getResult().c_str());
 }
 
 void DataVisitorTestScene::runThisTest()


cocos2dx/textures/CCTextureCache.cpp
```diff
@@ -250,28 +239,30 @@ void TextureCache::addImageAsyncCallBack(float dt)
...
        // cache the texture file name
        VolatileTexture::addImageTexture(texture, filename);
 #endif
-        // cache the texture
-        _textures->setObject(texture, filename);
+        // cache the texture. retain it, since it is added in the map
+        _textures.insert( std::make_pair(filename, texture) );
+        texture->retain();
+
         texture->autorelease();
```
minggo, line 266: Invoke `retain` then invoke `autorelease`, i think these two functions can be removed.

ricardoquesada, line 266: almost... if you remove both, it will fail in this scenario:

``` c++
// pseudo code
auto texture = TextureCache::addImage("sprite.png");
if( something ) {

   TextureCache::removeTexture( texture );

  // it will fail, because at this point the texture will be invalid.
  do_something_with_texture( texture );
}
```

ricardoquesada:
and thanks for merging it!

cocos2dx/textures/CCTextureCache.cpp
minggo, line 266: But it is now in `addImageAsyncCallBack`, not `addImage`. `addImageAsyncCallBack` is triggered by `addImageAsync` which returns nothing.

ricardoquesada, line 266: if that's the case, then you are correct.

PR #3560, opened 2013-09-09
issue 2771: Add physical API to engine

boyu0:


minggo:
Fix compiling error in #3665.

PR #3563, opened 2013-09-09
You have to pass long pointer for CURLINFO_RESPONSE_CODE.

ledyba:
You should pass long pointer to curl_easy_getinfo when you want to get CURLINFO_RESPONSE_CODE.

``` C+++
  CURLINFO_RESPONSE_CODE    = CURLINFO_LONG   + 2,
```

In 64bit gcc, sizeof(int) == 4 and sizeof(long) == 8, so irrelevant memory area will be overwritten when compiling on 64bit.

### DIFF
--- a/extensions/network/HttpClient.cpp
+++ b/extensions/network/HttpClient.cpp
@@ -324,7 +324,9 @@ public:
     {
         if (CURLE_OK != curl_easy_perform(_curl))
             return false;
-        CURLcode code = curl_easy_getinfo(_curl, CURLINFO_RESPONSE_CODE, responseCode);
+        long respCode;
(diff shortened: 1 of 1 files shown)


dumganhar:
Thanks. I will check it.

ledyba:
Thanks.
On Linux 64bit, HttpClient always crashes due to this problem.
In android and iOS, both of them are 32bit platform, so HttpClient do not crash at all.

But, iPhone 5s will be 64 bit platform, so it may cause crash.

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/3652 . Thanks. Could you please test the codes on develop branch? Thanks. If no problems, please close this PR.

ledyba:
Thanks！I have checked!

PR #3565, opened 2013-09-09
exposed addImageAsync to jsb

nakosung:
addImageAsync wasn't exposed.


nakosung:
this PR isn't related with lua_pop.

PR #3568, opened 2013-09-10
Master alloc perf test

ricardoquesada:
Tested on:

OS X 10.9 with Xcode 5.0-preview
- iOS project works OK.
- Mac project works OK with Xcode 5, but it didn't finish linking on Xcode 4. I guess it is a bug with OS X 10.9. Unfortunately I don't have a OS X 10.8 to test it.


minggo:
I will checkout it and add files in other project setting.

ricardoquesada:
ok. thanks.

minggo:
> Mac project works OK with Xcode 5, but it didn't finish linking on Xcode 4. I guess it is a bug with OS X 10.9. Unfortunately I don't have a OS X 10.8 to test it.

Change `C++ Language Dialect` and `C++ Standard Library` to `compiler-default` fix the problem. I don't know why. May c++0x and libc++ is not compatible with previous version.

minggo:
Done in #3577.

ricardoquesada:
thanks!

PR #3580, opened 2013-09-11
Fixed compilation on VC12 (VS2013 RC)

seemk:
std::min and std::max require the algorithm header, omitting this breaks the build on Visual Studio 2013 RC.


dumganhar:
Thanks.

PR #3583, opened 2013-09-12
Optimize GUI lib

nutty898:
1. Modify GUI lib;
2. Modify memory mange to "autorelease";


dumganhar:
A lot of `100644 → 100755`.. Why?

extensions/CocoStudio/GUI/UIWidgets/ScrollWidget/UIDragPanel.h
```diff
@@ -307,10 +310,21 @@ class UIDragPanel : public Layout, public UIScrollInterface
     void moveToInit();
     void moveToUpdate(float t);
     virtual void onSizeChanged();
+    /*compatible*/
+    /**
+     * These methods will be removed
+     */
+    virtual void setClippingEnable(bool is){setClippingEnabled(is);};
```
dumganhar, line 317: The parameter should be `bool isEnabled` rather than `bool is`.

nutty898, line 317: Theoe methods will be removed when published

PR #3601, opened 2013-09-14
add scenetest hd version

chengstory:
1、add scenetest hd resoruces.

### DIFF
--- a/samples/Cpp/TestCpp/Resources/hd/scenetest/fishes/Butterflyfish/Butterflyfish.ExportJson
+++ b/samples/Cpp/TestCpp/Resources/hd/scenetest/fishes/Butterflyfish/Butterflyfish.ExportJson
@@ -384,147 +384,147 @@
   ],
   "texture_data": [
     {
-      "name": "Fish11\\normal\\actor_fish_shizi_normal_001",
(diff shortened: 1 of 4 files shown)


samples/Cpp/TestCpp/Resources/hd/scenetest/fishes/Butterflyfish/Butterflyfish.ExportJson
```diff
@@ -0,0 +1,537 @@
...
+    },
+    {
+      "name": "Fish11\\normal\\actor_fish_shizi_normal_008",
+      "width": 0.0,
+      "height": 0.0,
+      "pX": 0.5,
+      "pY": 0.5
+    },
+    {
+      "name": "Fish11\\normal\\actor_fish_shizi_normal_009",
```
dumganhar: We should not use `backslash`(`\\`), instead, we should use `/`. I found some resource files use `/` and some use `\\`. Why?

chengstory: The file is exported by old version.

PR #3605, opened 2013-09-15
update armature

2youyou2:
1.Change rotation range from (-180 -- 180) to (-infinity -- infinity)
2.Fix not init CCContourData
3.Change hd resource from "\" to "/"


minggo:
@2youyouo2 
Please fix compiling errors.

PR #3639, opened 2013-09-17
Fix crash when websocket close by server side. send an error message to UI when libwebsocket_client_connect() return nullptr

u0u0:


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/5496. Thanks.

PR #3679, opened 2013-09-18
fix the CCTouchScriptHandlerEntry will release twice lua reference

dragoncheng:
the bug issue is Bug #2535(http://cocos2d-x.org/issues/2535)

### DIFF
--- a/cocos2dx/script_support/CCScriptSupport.cpp
+++ b/cocos2dx/script_support/CCScriptSupport.cpp
@@ -49,10 +49,10 @@ ScriptHandlerEntry* ScriptHandlerEntry::create(int nHandler)
 
 ScriptHandlerEntry::~ScriptHandlerEntry(void)
 {
-    if (_handler != NULL )
+    if (_handler != 0 )
     {
         ScriptEngineManager::getInstance()->getScriptEngine()->removeScriptHandler(_handler);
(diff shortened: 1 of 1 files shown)


cocos2dx/script_support/CCScriptSupport.cpp
```diff
@@ -97,8 +101,12 @@ TouchScriptHandlerEntry* TouchScriptHandlerEntry::create(int nHandler,
 
 TouchScriptHandlerEntry::~TouchScriptHandlerEntry(void)
 {
-    ScriptEngineManager::getInstance()->getScriptEngine()->removeScriptHandler(_handler);
-    LUALOG("[LUA] Remove touch event handler: %d", _handler);
+	if (_handler != 0 )
+	{
+        ScriptEngineManager::getInstance()->getScriptEngine()->removeScriptHandler(_handler);
```
dumganhar, line 106: `TouchScriptHandlerEntry` is inherited from `ScriptHandlerEntry`. So I think the `removeScriptHandler` is only needed in its parent's destructor.  `TouchScriptHandlerEntry`'s destructor will not need it. Right?

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/5425 . Thanks.

PR #3682, opened 2013-09-18
fix CCTouchScriptHandlerEntry will release twice lua reference and texture initWithString

dragoncheng:
1: fix CCTouchScriptHandlerEntry will release twice lua reference
2: fix the error when texture stroke lable under IOS and Android.


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/3810. Thanks.

PR #3701, opened 2013-09-18
Small fixes for Xcode templates

ricardoquesada:
Also, uses `auto` in the template code to be more c++11 friendly


dumganhar:
Thanks.

PR #3726, opened 2013-09-20
Adds missing icons to templates and samples

ricardoquesada:
Adds missing icons on iOS templates and samples


minggo:
@ricardoquesada 
What are these icons used for?

ricardoquesada:
spotlight, the "search" functionality inside ios

PR #3739, opened 2013-09-22
add judgment for retaining m_actionSpawn

liamcindy:


extensions/CocoStudio/Action/CCActionNode.cpp
```diff
@@ -313,7 +313,10 @@ CCSpawn * ActionNode::refreshActionProperty()
 	}
 
 	m_actionSpawn = CCSpawn::create(cSpawnArray);
-	m_actionSpawn->retain();
+	if (m_actionSpawn != NULL)
+	{
+		m_actionSpawn->retain();
+	}
```
dumganhar, line 322: It's cleaner to use `CC_SAFE_RETAIN` and `CC_SAFE_RELEASE` to check whether the pointer is valid.

liamcindy, line 322: Thanks

PR #3741, opened 2013-09-22
Fixed some compiling-warnings (MSVC2012).

syrokomskyi:
C4996 \* was declared deprecated
C4305 \* truncation from 'double' to 'float'
C4018 \* signed/unsigned mismatch
C4800 \* forcing value to bool 'true' or 'false'


dumganhar:
Thanks.

PR #3742, opened 2013-09-23
Clean up some API documentation

mlepage:
Mostly typo fixes, and in one case changed "queue" to "stack" as I believe the scenes are stored in a stack not a queue.


minggo:
Thanks.

PR #3767, opened 2013-09-23
re-introduce Cocos2dxHelper.runOnGLThread(Runnable)

michaelcontento:
It's crucial to execute various tasks synchonized with the rendering loop /
thread, but this ability (runOnGLThread) got removed with the NativeActivity
transitions a few days ago. This patch adds this method back again.
# Why can't we just use runOnUiThread()? Because it seems not to be synchronized

properly and can cause broken OpenGL textures or even the whole app to crash.

Isn't there a better way to get into the mainloop? I hope so! But I'm not a
experienced Android developer and this solution is quite simple / the only one
that worked without any problems.

### DIFF
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
@@ -23,8 +23,10 @@ THE SOFTWARE.
  ****************************************************************************/
 package org.cocos2dx.lib;
 
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.io.UnsupportedEncodingException;
 import java.util.Locale;
+import java.lang.Runnable;
 
 import android.app.Activity;
 import android.app.AlertDialog;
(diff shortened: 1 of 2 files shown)


minggo:
@michaelcontento 
Awesome job. I think add runOnGLThread in Director is better. Then all platforms can use this API. 
You can refer to [this thread](http://www.cocos2d-x.org/forums/6/topics/34700) for more information.

What's your opinion?

michaelcontento:
I've no real opinion about where this should be added or if it's a good idea to make this available for all platforms. But, IMHO, it's very important to get this **available from the Java side** with plain Runnable objects! 

That's the reason why most of the code is Java and C++ only calls the dispatching at the right time. But if someone is able to handle the Runnable references through C++ and expose `Director::runOnGLThread()` to Java -- awesome! :+1: 

But then we should maybe move `Cocos2dxHelper.runOnUiThread()` to the `Director`/`ThreadHelper` too? 
And, at least to me, it seems that the future of this is quite uncertain, right? 

So what about accepting this patch now and move both `runOnUiThread` and `runOnGLThread` at the same time in the future? Because, IMHO, `runOnGLThread` is quite important for everyone who tries to integrate 3rd-party libs.

Thanks! :)

minggo:
May be you misunderstood my thoughts.
What i mean is that, we implement `Cocos2dxHelper.runOnGLThread()` in `ThreadHelper` class. Then Director can handler it in main loop.

As you said, it is important. And i have to merge this PR right now:)

Thanks again.

PR #3768, opened 2013-09-23
wrong variable used which could crash the whole app

michaelcontento:


minggo:
Good catch.
Why it doesn't cause compiling error? 
Is there a global variable named `env`?

PR #3783, opened 2013-09-24
updates Android project.properties

ricardoquesada:
Instead of using API level 10 or 13, it updates it to 18 since
if you install the latest SDK, by default it will only install level API 18.

### DIFF
--- a/cocos2dx/platform/android/java/project.properties
+++ b/cocos2dx/platform/android/java/project.properties
@@ -12,4 +12,4 @@
 
 android.library=true
 # Project target.
-target=android-10
+target=android-18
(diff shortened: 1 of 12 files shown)


minggo:
Why?
I think we should use 13, because cocos2d-x support Android since 2.3(13). And we should make sure we don't use API that doesn't exist in Android 13.

And if you have downloaded Android SDK 13, it will not change project.properties. The file will be changed because you don't download Android SDK 13.

ricardoquesada:
ah... ok. Thanks. So, i'm closing this PR then.
One project mentions that it uses API level 10 though.

PR #3801, opened 2013-09-25
update armature

2youyou2:
1.fix armature transform calculate
2.add get bone's world state function


extensions/CocoStudio/Armature/CCBone.cpp
```diff
@@ -73,6 +73,7 @@ CCBone *CCBone::create(const char *name)
     m_tWorldTransform = CCAffineTransformMake(1, 0, 0, 1, 0, 0);
     m_bBoneTransformDirty = true;
     m_eBlendType = BLEND_NORMAL;
+	m_tWorldInfo = NULL;
```
dumganhar: Please fix the indention.

PR #3806, opened 2013-09-25
fix a bug that index may out of range

minggo:


minggo:
This bug is fixed in developer

PR #3825, opened 2013-09-25
Updated ReadMe to use master branch. Removed comments with error characters

stammen:
1. Updated ReadMe for How to build for WinRT (Windows 8.0) to use master branch instead of winrt branch. 
2. Removed comments with error characters.


dumganhar:
Thanks

PR #3829, opened 2013-09-25
Winrt audio fix

stammen:
1. adds missing background.wav file for wp8 to samples/cpp/testcpp/resources
2. fixes compiler warnings in CocosDenshion for wp8


dumganhar:
Thanks.

PR #3847, opened 2013-09-26
fix crash by using getFileData to read text files.

langresser:
the buffer got by getFileData is not end by zero, if use strlen and so on then app may crash.(eg. CCUserDefault->getXMLNodeForKey)
I think it is safe to alloc a buffer that end by zero,and return then real size. image data is also ok.


dumganhar:
Added new API for reading data or string from file.

``` c++
FileUtils::getDataFromFile(filename);
FileUtils::getStringFromFile(filename);
```

Closed this PR.

PR #3854, opened 2013-09-27
add *.xccheckout to gitignore

pktangyue:
Reference: 
https://github.com/github/gitignore/pull/774 
http://stackoverflow.com/questions/18340453/should-xccheckout-files-in-xcode5-be-ignored-under-vcs


dumganhar:
Thanks.

PR #3857, opened 2013-09-27
issues #2834:Use Android standalone toolchain on windows to remove the d...

WenhaiLin:
Use Android standalone toolchain on windows to remove the dependence of cygwin


dumganhar:
looks great.

PR #3861, opened 2013-09-28
proper handling of keyboard repeat events after update to glfw3

marrone:
Keyboard onKeyReleased events are being repeatedly fired even before the you physically release the keyboard. 
This appears to have happened with the move from glfw2 to glfw3 which now enabled REPEAT by default, see:
http://www.glfw.org/docs/3.0/moving.html#moving_repeat

To preserve the original behaviour where the onKeyReleased event is fired only when the keyboard key is actually physically released, we need to ignore the GLFW_REPEAT action, see:
http://www.glfw.org/docs/3.0/group__input.html#ga592dd1919f8a1dc7576b13cdd8b7b695

### DIFF
--- a/cocos2dx/platform/mac/CCEGLView.mm
+++ b/cocos2dx/platform/mac/CCEGLView.mm
@@ -238,8 +238,11 @@ void EGLViewEventHandler::OnGLFWMouseMoveCallBack(GLFWwindow* window, double x, 
 
 void EGLViewEventHandler::OnGLFWKeyCallback(GLFWwindow *window, int key, int scancode, int action, int mods)
 {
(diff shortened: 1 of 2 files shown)


minggo:
Thanks @marrone.
It is a nice patch. Could you please modify MAC and windows too?

marrone:
I could make the change, but I don't currently have either a Mac or Windows platform to verify against

marrone:
Ok, I've updated mac and win32 platforms... though like I said I am not currently able to actually verify on those platforms myself

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/5018 . Thanks.

PR #3862, opened 2013-09-28
Crash fix for postNotification()

BodbDearg:
Fix crashes in CCNotificationCenter::postNotification caused by unsafe iteration of the observers list. The code should maintain a strong reference to all observer targets while iterating through the list, in case they are destroyed while we iterate.

### DIFF
--- a/cocos2dx/support/CCNotificationCenter.cpp
+++ b/cocos2dx/support/CCNotificationCenter.cpp
@@ -182,6 +182,30 @@ void NotificationCenter::postNotification(const char *name, Object *sender)
     Array* ObserversCopy = Array::createWithCapacity(_observers->count());
     ObserversCopy->addObjectsFromArray(_observers);
     Object* obj = NULL;
+
+    // CRASH FIX:
+    // Maintain a strong reference to the target of each observer while we iterate through the observers list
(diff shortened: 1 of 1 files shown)


BodbDearg:
Note: another (possibly better?) way of fixing this problem might be to have 'NotificationObserver' observer maintain a strong reference rather than a weak reference to it's target object. I think that would be a very big architectural change however and potentially very dangerous for existing cocos2dx codebases since it would greatly affect the object lifecycle and when things get destroyed. This change is much more limited in scope and far less risky.

minggo:
Thanks @DarraghCoy 
We will integrate `NotificationCenter` into `EventDispatcher`. And `NotificationCenter` will be deprecated or removed.

dumganhar:
@DarraghCoy 
Could you try `EventDispatcher` to dispatch event?
I deprecated `NotificationCenter` in PR https://github.com/cocos2d/cocos2d-x/pull/3864 .

BodbDearg:
Thanks! I had not realised NotificationCenter had been deprecated since we are still using Cocos2dx 2.x (we're planning to migrate to 3.x soon). I will close this pull request.

PR #3867, opened 2013-09-30
fix bug: the alpha value of CCProgressTimer can not be updated correctly sometimes

nichos1983:
Steps of reproducing the bug:
1 Create A of CCNode type. Create B as a CCProgressTimer. Add B as the child of A.
2 call A->setCascadeOpacityEnabled(true).
3 call A->runAction(CCFadeTo::create(1.0f, 0));

Result:
Due to the setting in step 2, the alpha value(opacity) of B is expected to be updated to the same value as A when running the action. And alpha of B is zero when the action is finished. But the result is, the alpha value of A is correct, while B is wrong and the alpha of B is not zero in the end.

Root cause:
CCProgressTimer is actually a wrapper of CCSprite. If we run an action, when updating the alpha value of CCProgressTimer, we need to update the alpha value of its child CCSprite accordingly. So we need to overwrite void ProgressTimer::updateDisplayedOpacity(GLubyte parentOpacity).

Fix:
Add the function and overwrite it. The bug is gone.

### DIFF
--- a/cocos2dx/misc_nodes/CCProgressTimer.cpp
+++ b/cocos2dx/misc_nodes/CCProgressTimer.cpp
@@ -170,6 +170,14 @@ GLubyte ProgressTimer::getOpacity() const
     return _sprite->getOpacity();
 }
 
+void ProgressTimer::updateDisplayedOpacity(GLubyte parentOpacity)
+{
+    _sprite->updateDisplayedOpacity(parentOpacity);
+    
+    NodeRGBA::updateDisplayedOpacity(parentOpacity);
+    updateColor();
+}
+
 // Interval
 
 ///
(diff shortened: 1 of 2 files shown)


dumganhar:
@nichos1983 
Thanks for this fix. Could you please add a test case in `ActionProgressTest`?

sbrednikhin:
Same with color. Cascade color won't work correctly.
Better way is to override `updateDisplayedColor` and `updateDisplayedOpacity`
```
void cocos2d::ProgressTimer::updateDisplayedOpacity(GLubyte parentOpacity)
{
    Node::updateDisplayedOpacity(parentOpacity);

    if (_sprite)
        _sprite->updateDisplayedOpacity(_displayedOpacity);
}
```

```
void cocos2d::ProgressTimer::updateDisplayedColor(const Color3B & parentColor)
{
    Node::updateDisplayedColor(parentColor);

    if (_sprite)
        _sprite->updateDisplayedColor(_displayedColor);
}
```

In that case custom `setOpacity()` and `setColor()` can be made deprecated.

PR #3869, opened 2013-09-30
WinRT LabelTest: Testing CCLabelTTF with Chinese Character fix

stammen:
Fix for LabelTest: Testing CCLabelTTF with Chinese Character

The Marker Felt TTF file does not contain any Chinese characters so FreeType cannot render the characters. WinRT version of the test uses PMingLiU system font. 

WP8 will still not render Chinese Characters until Marker Felt TTF file includes the missing characters.

### DIFF
--- a/samples/Cpp/TestCpp/Classes/LabelTest/LabelTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/LabelTest/LabelTest.cpp
@@ -1124,7 +1124,14 @@ string LabelTTFMultiline::subtitle()
 LabelTTFChinese::LabelTTFChinese()
 {
     CCSize size = CCDirector::sharedDirector()->getWinSize();
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_WINRT)
+    // Marker Felt is missing Chinese characters
+    CCLabelTTF *pLable = CCLabelTTF::create("中国", "PMingLiU", 30);
+#else
     CCLabelTTF *pLable = CCLabelTTF::create("中国", "Marker Felt", 30);
+#endif
+
+    
     pLable->setPosition(ccp(size.width / 2, size.height /2));
     this->addChild(pLable);
 }


minggo:
But it works on other platforms, the reason may be
a. other platforms don't support `Marker Felt TTF`, then it choose another one
b. `Marker Felt TTF` on other platforms include Chinese characters

But i think iOS support `Marker Felt TTF`, why it can render Chinese characters? `Marker Felt TTF` is different on different platforms?

stammen:
Unfortunately WinRT does not work this way. I looke at the iOS code and it uses system calls to render the font. We had to use FreeType to get fonts working on WinRT and Windows Phone 8 (WP8). WP8 does not allow access to system fonts from C++ (they left that out of the SDK) and the SDK does not include any useful font functions. Windows Phone 8 must include a TTF file with Chinese fonts in order to be able to render them. Can you recommend an open source Chinese TTF file?

minggo:
I am sorry. I am not familiar with open source Chinese TTF file.
Do you want to use a Chinese TTF file  for default font?

I think i should merge this pull request and post a news in forum now.
What's your opinion?

Weeds:
I just pulled the Marker Felt font from my iPad and it does not contain any chinese characters as well. And looking at a sample application that displays various unicode tables all fonts have the same glyphs for CJK characters. So I guess iOS (like Android) uses a fallback font when rendering a string that contains glyphs that are not in the current font.

I'm not sure if Arial does provide all the required glyphs (and if it is safe for developers to ship it with their applications)
But a better solution would be to check if FT_Get_Char_Index returns 0 for the current face. In that case the implementation could switch to a fallback font and render the glyph with that font.

Would it be possible to create the font bitmap with C#? In that case we could use system functions like on iOS on Android. However I would prefer to stick with Freetype as it can be used on multiple platforms (Linux, Blackberry,...)

PR #3870, opened 2013-09-30
Added Http Support for WinRT and WP8

martell:
This is an initial import to add support for HttpClient into extensions for WinRT and WP8.
This has been taken from the WP8 and WinRT directX port done by Atom Split Games

It uses 
1. c++11 threading instead of pthreads as pthreads is not supported on winrt.
2. Windows::Networking::Sockets instead of libcurl as winsock is not available on WinRT 

the way Windows networking sockets handles headers is strange. 
I will fix this in an update with a work around should be very a simple fix.

Also note windows networking sockets must have either http:// or https:// in the url.
Curl tends to handle the request if you don't specify the protocol.

Check out this first app using it here
http://apps.microsoft.com/windows/en-ie/app/56c0be05-ff43-4ae3-9330-faa0b84498a6

Martell Out ;)

### DIFF
--- a/extensions/network/HttpClient.cpp
+++ b/extensions/network/HttpClient.cpp
@@ -1,7 +1,7 @@
 /****************************************************************************
  Copyright (c) 2010-2012 cocos2d-x.org
  Copyright (c) 2012 greathqy
- Copyright (c) 2013 Martell Malone
+ Copyright (c) 2013-2014 Martell Malone < martell malone at g mail dot com >
  
  http://www.cocos2d-x.org
  
@@ -44,6 +44,7 @@
 #include <ppltasks.h>
 #include <sstream>
 #include <robuffer.h>
+
 
 using namespace concurrency;
 using namespace Microsoft::WRL;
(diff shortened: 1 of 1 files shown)


unigame:
hi martell, i have  test  it and it seems to be Asynchronous,but  non-blocking.how do i make it blocking when response is not arrived?
Forgive my poor English, this is from Google translation

martell:
Hi unigame,
It should work the same as the libcurl implementation but I haven't made it non blocking. 
We may be better off using this in the long run.
https://winrtsock.codeplex.com/
Kind Regards
Martell

martell:
hey @stammen, @minggo
I've updated this pull request to match up with the cocos2d-x http request headers so now http should work on windows rt and wp8 just like any other platform.

I have begun to port libwebsockets to both wp8 and winrt and curl to winrt so at a later point we can use curl on everything and libwebsockets on everything, but for the time being this adds http support on winrt and wp8

@koowolf Your opinion might be needed here as you are working on the libcurl port for WP8
should we merge this networking sockets version for now until curl and webscokets is ready on all platforms?

Thanks
Martell

stammen:
Have you looked at https://winrtsock.codeplex.com/? It might be a useful starting point for porting the WinSock API.

martell:
I have if you look up to my comments above you will see I referenced it. :)
This is what I plan on using to create libcurl and libwebsockets for winrt.

stammen:
Cool! Thanks.

martell:
In the mean time until all those projects are ready maybe we should merge this?
We can change it to the winrtsock version then once they are done?

martell:
In light of both of our thoughts on this I have ported libcurl to winrt in #5098

squirrelChen:
Good job! Thanks!

PR #3871, opened 2013-10-02
Android add onActivityResult listener

michaelcontento:
For some third-party libs it's required to handle messages send to `onActivityResult` of the main activity. But with the transition to `NativeActivity` this rather difficult to add custom callbacks to `onActivityResult`.

With this change it's possible to register listeners that get called once something triggers `onActivityResult`. All listeners are stored and managed in `Cocos2dxHelper`, because it seems to be the current best place for "helper stuff that android requires here and there". `Cocos2dxNativeActivity` is the new root activity and inherits from `NativeActivity`. The only change here is a little loop in `onActivityResult` to iterate through and call all listeners.
#### Why don't store and manage all listeners in `Cocos2dxNativeActivity` itself?

I'd like to keep the `Cocos2dxNativeActivity` as short, simple and close to the `NativeActivity` as possible. And by moving all helper stuff into `Cocos2dxHelper` it's pretty easy to search/grep through all java files and have a look at such custom integrations (similar to `(int)` vs `static_cast<int>()` in C++).

### DIFF
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
@@ -23,6 +23,8 @@ THE SOFTWARE.
  ****************************************************************************/
 package org.cocos2dx.lib;
 
+import java.util.Set;
+import java.util.LinkedHashSet;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.io.UnsupportedEncodingException;
@@ -41,6 +43,7 @@ import android.os.Build;
 import android.util.DisplayMetrics;
 import android.view.Display;
 import android.view.WindowManager;
+import android.preference.PreferenceManager.OnActivityResultListener;
 
 public class Cocos2dxHelper {
 	// ===========================================================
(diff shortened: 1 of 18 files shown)


minggo:
Should other call back, such as orientation changed, be implemented in Java too?
If so, then what should be implemented in c++ and what should be implemented in java?

sachingarg05:
I ran into an issue where trying to do any cocos/gl operation in onActivityResult fails with with a logcat message saying no valid EGL context. This seems correct because cocos2dx reloads all textures when activity resumes and that happens 'after' onActivityResult.

It seems to be either because Android unloads GL context on activity pause ( http://stackoverflow.com/questions/3579775/no-current-opengl-context-when-returning-from-activity ), or because onActivityResult is called on main UI thread but all GL rendering needs to happen on GL thread which is I think same as native activity's main loop. Or it could be a combination of both.

Did you not run into similar issue with your implementation?

sachingarg05:
I fixed my GL thread problem in onActivityResult result by only saving activity data in onActivityResult (using JNI), and then doing GL/Cocos2dx processing in AppDelegate::applicationWillEnterForeground. So that problem is not related to cocos2dx or this pull request, it is just a known issue with Android's onActivityResult.

@minggo , I think there are many things which different apps will want to do in Java activity. The common and popular cases like orientation change should be supported within cocos2dx, but for other cases it will be difficult to support everything using such listeners etc. And it will be difficult for user to find out what is the cocos2dx way of doing Android things.

It might be less complicated to just let users subclass from Cocos2dxNativeActivity in their code and add the code that they need for onActivityResult etc... (or they can subclass directly from android.app.NativeActivity in their code if we don't need Cocos2dxNativeActivity for anything else). Then they can use standard Android code.

@michaelcontento , am I missing some additional functionality that this PR gives which can't be implemented by subclassing Cocos2dxNativeActivity ?

sachingarg05:
Many Android developers may not know how to use native activity, I am also learning how to deal with it and its not easy. 

To guide them, maybe we should create Cocos2dxNativeActivity in the generated sample project code, and not inside cocos2dx/platform/android/java/src/org/cocos2dx/lib.

Then when user creates a new Android project, they will see a (empty) Android activity class where they can start adding/copying their android code without thinking too much.

PR #3872, opened 2013-10-04
Clang build fix

azmeuk:
This fixes the Linux build and adds a Travis build for clang


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/5013. Thanks.
BTW, we are using `CMake` for building linux projects now.

PR #3875, opened 2013-10-07
fix bug: sys.localStorage.getItem() do not support utf8 in jsb.

rablwupei:


scripting/javascript/bindings/js_bindings_system_functions.cpp
```diff
@@ -30,8 +31,13 @@ JSBool JSB_localStorageGetItem(JSContext *cx, uint32_t argc, jsval *vp) {
...
 
 	ret_val = localStorageGetItem((char*)arg0  );
+    
+    std::string utf8;
+    if (ret_val) {
+        utf8 = ret_val;
+    }
 
-	jsval ret_jsval = charptr_to_jsval( cx, ret_val);
+	jsval ret_jsval = std_string_to_jsval(cx, utf8);
```
dumganhar, line 40: How about use `c_string_to_jsval`? 
And the codes

```
std::string utf8;
    if (ret_val) {
        utf8 = ret_val;
    }
```

will be not needed.

dumganhar:
Hi, @rablwupei 
Thanks for your contribution. 
But this Pull Request contains more than one bug fix. It will be great if  you could separate it to several Pull Request?

PR #3876, opened 2013-10-07
fixed bug of WebSocket that it can not receive over 4096 byte

HoGarfield:


dumganhar:
Hi HoGarfield,
Thanks. Could you please add a new test in `ExtensionsTest/WebsocketTest`?

PR #3879, opened 2013-10-07
 inscreasing method setDesignResolutionSize to CCDirector which is in CCEGLView

HoGarfield:
There isn`t CCEGLView in cocos2d-x jsb.so setDesignResolutionSize is
increased in CCDirector,in order to work with Multiple Resolutions

### DIFF
--- a/cocos2dx/CCDirector.cpp
+++ b/cocos2dx/CCDirector.cpp
@@ -65,6 +65,52 @@ THE SOFTWARE.
 #include "CCConfiguration.h"
 #include "keyboard_dispatcher/CCKeyboardDispatcher.h"
 
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
+#include "platform/ios/CCEGLView.h"
+#endif // CC_TARGET_PLATFORM == CC_PLATFORM_IOS
+
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
+#include "platform/android/CCEGLView.h"
+#endif // CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID
+
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_BLACKBERRY)
+#include "platform/blackberry/CCEGLView.h"
+#endif // CC_TARGET_PLATFORM == CC_PLATFORM_BLACKBERRY
+
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
+#include "platform/win32/CCEGLView.h"
+#endif // CC_TARGET_PLATFORM == CC_PLATFORM_WIN32
+
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
+#include "platform/mac/CCEGLView.h"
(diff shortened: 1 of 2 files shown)


cocos2dx/CCDirector.cpp
```diff
@@ -65,6 +65,52 @@ of this software and associated documentation files (the "Software"), to deal
...
+#endif // CC_TARGET_PLATFORM == CC_PLATFORM_EMSCRIPTEN
+
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_TIZEN)
+#include "platform/tizen/CCEGLView.h"
+#endif // CC_TARGET_PLATFORM == CC_PLATFORM_TIZEN
+
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_QT5)
+#include "platform/qt5/CCEGLView.h"
+#endif // CC_TARGET_PLATFORM == CC_PLATFORM_QT5
+
```
dumganhar, line 113: Only `#include "CCEGLView.h"` is enough since we `platform/XXX/` is in the include path.

cocos2dx/CCDirector.cpp
```diff
@@ -918,6 +964,13 @@ void Director::createStatsLabel()
     _FPSLabel->setPosition(CC_DIRECTOR_STATS_POSITION);
 }
 
+void Director::setDesignResolutionSize(float width, float height, int resolutionPolicy)
+{
+    ResolutionPolicy temp[]={ResolutionPolicy::EXACT_FIT,ResolutionPolicy::NO_BORDER,ResolutionPolicy::SHOW_ALL,ResolutionPolicy::FIXED_HEIGHT,ResolutionPolicy::FIXED_WIDTH,ResolutionPolicy::UNKNOWN};
```
dumganhar, line 969: Why not just use `Director::setDesignResolutionSize(float width, float height, ResolutionPolicy resolutionPolicy)`?

HoGarfield, line 969: Can I use class ResolutionPolicy in JSB?

dumganhar, line 969: Bindings-generator treats `enum` as `int`.
So you just need to define the constant in `jsb_cocos2d.js`.

``` js
cc.RESOLUTION_POLICY = {
    // The entire application is visible in the specified area without trying to preserve the original aspect ratio.
    // Distortion can occur, and the application may appear stretched or compressed.
    EXACTFIT:0,
    // The entire application fills the specified area, without distortion but possibly with some cropping,
    // while maintaining the original aspect ratio of the application.
    NOBORDER:1,
    // The entire application is visible in the specified area without distortion while maintaining the original
    // aspect ratio of the application. Borders can appear on two sides of the application.
    SHOW_ALL:2,
    // The application takes the height of the design resolution size and modifies the width of the internal
    // canvas so that it fits the aspect ratio of the device
    // no distortion will occur however you must make sure your application works on different
    // aspect ratios
    HEIGHT:3,
    // The application takes the width of the design resolution size and modifies the height of the internal
    // canvas so that it fits the aspect ratio of the device
    // no distortion will occur however you must make sure your application works on different
    // aspect ratios
    WIDTH:4,

    UNKNOWN:5
};
```

PR #2883, opened 2013-06-13
fixed strokeColor and fontFillColor of CCLabelTTF

musikov:

PR #3880, opened 2013-10-07
support orientation change in android

ondesly:
Catching APP_CMD_CONFIG_CHANGED in nativeactivity in Android to change screen size and notify app about orientation change.

applicationConfigChanged method in CCApplicationProtocol to notify app about orientation change.

### DIFF
--- a/cocos2dx/platform/android/nativeactivity.cpp
+++ b/cocos2dx/platform/android/nativeactivity.cpp
@@ -534,6 +534,18 @@ static void engine_handle_cmd(struct android_app* app, int32_t cmd) {
                 engine_draw_frame(engine);
             }
             break;
+		case APP_CMD_CONFIG_CHANGED: {
+			// Orientation changed
+			cocos2d::EGLView *view = cocos2d::Director::getInstance()->getOpenGLView();
+
+			cocos2d::Size size = view->getFrameSize();
+
+			view->setFrameSize(size.height, size.width);
+			view->setDesignResolutionSize(size.height, size.width, ResolutionPolicy::SHOW_ALL);
+			// Notify app about changes
+			cocos2d::Application::getInstance()->applicationConfigChanged();
+			break;
+		}
         case APP_CMD_TERM_WINDOW:
             // The window is being hidden or closed, clean it up.
             engine_term_display(engine);
--- a/cocos2dx/platform/CCApplicationProtocol.h
+++ b/cocos2dx/platform/CCApplicationProtocol.h
@@ -57,6 +57,13 @@ public:
     */
     virtual void applicationWillEnterForeground() = 0;
 
+	/**
+	@brief  This function will be called when config like orientation was changed.
+	* @js NA
+	* @lua NA
+	*/
+	virtual void applicationConfigChanged() = 0;
+
     /**
     @brief    Callback by Director for limit FPS.
     @param interval The time, expressed in seconds, between current frame and next.
--- a/template/multi-platform-cpp/Classes/AppDelegate.cpp
+++ b/template/multi-platform-cpp/Classes/AppDelegate.cpp
@@ -48,3 +48,8 @@ void AppDelegate::applicationWillEnterForeground() {
     // if you use SimpleAudioEngine, it must resume here
     // SimpleAudioEngine::sharedEngine()->resumeBackgroundMusic();
 }
+
+// this function will be called when config like orientation was changed
+void AppDelegate::applicationConfigChanged() {
+
+}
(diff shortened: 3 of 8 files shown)


dumganhar:
But `configChanged` contains `orientationChanged`. Is this name suitable for `orientation changed`?
Another, how does developer know the `orientation` has been changed to which orientation?
Does this callback should contain parameters?

cocos2dx/platform/android/nativeactivity.cpp
```diff
@@ -534,6 +534,18 @@ static void engine_handle_cmd(struct android_app* app, int32_t cmd) {
...
             }
             break;
+		case APP_CMD_CONFIG_CHANGED: {
+			// Orientation changed
+			cocos2d::EGLView *view = cocos2d::Director::getInstance()->getOpenGLView();
+
+			cocos2d::Size size = view->getFrameSize();
+
+			view->setFrameSize(size.height, size.width);
+			view->setDesignResolutionSize(size.height, size.width, ResolutionPolicy::SHOW_ALL);
```
minggo, line 544: Why set design resolution to `SHOW_ALL`? The design resolution may be set before.

ondesly, line 544: I just copied it from iOS method. It doesn't work without setDesignResolutionSize.

ondesly:
APP_CMD_CONFIG_CHANGED is used not only for orientation but any changes in configuration. configChanged will be opened because of different causes, not only orientation change. But it's suitable for orientation very well.

I think developer can look at new width and height: width is bigger, so landscape, height is bigger, so portrait. But it's good idea to provide some parameters. For me it isn't useful, because app needs to know only new screen size, not landscape or portrait.

cocos2dx/platform/android/nativeactivity.cpp
minggo, line 544: I think `EGLViewProtocol` should add a method to get resolution policy, suppose `getResolutionPolicy`, then the method should change to

`view->setDesignResolutionSize(size.height, size.width, view->getResolutionPolicy());`

ondesly, line 544: Is it ok if I add changes to this pull request?

minggo, line 544: Yep, i think it is ok.
It is better if you want to add it in another pull request.

ondesly, line 544: It's my first pull request. I don't know how do it right.
Should I add a method to get resolution policy in new pull request and update this with it or do it other way?

sachingarg05:
I proposed a similar change request as a issue (not a pull request, sorry). Please see my proposed code here: http://www.cocos2d-x.org/issues/2883 It also takes care of resizing the scene and includes support for iOS. For Android I used APP_CMD_CONFIG_CHANGED just like in your code.

applicationConfigChanged can take android_app\* as parameter but it wont be a cross platform parameters. In my code I used this as callback: void onOrientationChange(bool isLandscape, int w, int h);

sachingarg05:
App needs to know landscape or portrait on iOS because if landscape, then app needs to manually swap height/width value before passing it to cocos2dx. Check my code for details.

sachingarg05:
Another problem in your code is that in applicationConfigChanged you are just swapping values of height and width. On some devices with on-screen buttons like nexus 4 and nexus 7, height and width are not same values due to new location of on-screen buttons after orientation change. Same happens on iOS if the app has status bar visible. We need to get the new height and width values again from the OS.

cocos2dx/platform/android/nativeactivity.cpp
minggo, line 544: You can add a method in another pull request, suppose pull request 2. After 2 is merged, you can synchronize cocos2d-x latest code, then use the method added in pull request 2.

sachingarg05, line 544: Mingo, Ondesly, please check my comments on this pull request and on the code I posted on Issue #2883. This pull request can be improved to support more android devices and iOS also. Unfortunately I am not familiar with GIT to create another pull request but I have posted all my code in redmine issue #2883.

ondesly:
Rwzpub, you are right. I don't test on such devices. It should be improved.

sachingarg05:
My code at http://www.cocos2d-x.org/issues/2883  takes care of all these issues for both iOS and Android. It will be really great if you can create a pull request based on that. It uses same APP_CMD_CONFIG_CHANGED for android, adds a similar callback for iOS. In callback I do setDesignResolutionSize just like you, and then I reload current scene again so that it is redrawn for new screen size.

ondesly:
Ok. I will check it and test on devices and will try to make new pull request. Thank you!

PR #3882, opened 2013-10-07
fixed Bug that it crashed when calling schedule function

HoGarfield:


dumganhar:
Hi, @HoGarfield
Thanks for your contribution. Could you add a new test to reproduce this issue in TestJavascript?

scripting/javascript/bindings/cocos2d_specifics.cpp
```diff
@@ -1938,7 +1938,7 @@ JSBool js_CCScheduler_schedule(JSContext *cx, uint32_t argc, jsval *vp)
                 tmpCObj->setPureJSTarget(tmpObj);
             }
             
-            JSScheduleWrapper::setTargetForSchedule(argv[1], tmpCObj);
+            JSScheduleWrapper::setTargetForSchedule(argv[0], tmpCObj);
```
dumganhar, line 1941: It should be `argv[0]`, since `cc.Scheduler.schedule`'s parameter is `target, func, xxx, xxx,xxx`, and the `func` is the second parameter, therefore, it should be `argv[1]`.

PR #3883, opened 2013-10-07
fixed bug that it reset content offset when reloadData be called

HoGarfield:


dumganhar:
Could you add a test in `TestCpp/ExtensionsTest/TableViewTest` to reproduce this issue? Thanks.

HoGarfield:
1. add a button to TableViewTest which call function reloadData of CCTableView by press it.

2.run TableViewTest 

3.scroll the table

4.press the button in order to call function reloadData

5.the table will be changed to no scroll

Issue #2677, opened 2013-05-30
jsbinding of plugin-x  for android  compile erro


zj831007:
I create a project as the sample of HelloAnalytics-JS, but there have man erro when built it, such as: jni/../../../../plugin/jsbindings/auto/jsb_pluginx_umeng_auto.cpp:571:74: error: 'CCLOGINFO' was not declared in this scope  and so on, It is same as there are man .h file  that not in the namespace of plugin can not find,  how to setting the plugin project?
My Android.mk file is like this:
LOCAL_PATH := $(call my-dir)

include $(CLEAR_VARS)

LOCAL_MODULE := cocos2djs_shared

LOCAL_MODULE_FILENAME := libcocos2djs

LOCAL_SRC_FILES := wuxia/main.cpp \
                   ../../Classes/AppDelegate.cpp \
                   ../../../../plugin/jsbindings/manual/jsb_pluginx_spidermonkey_specifics.cpp  \
                   ../../../../plugin/jsbindings/manual/pluginxUTF8.cpp \
                   ../../../../plugin/jsbindings/manual/jsb_pluginx_basic_conversions.cpp \
                   ../../../../plugin/jsbindings/manual/jsb_pluginx_manual_iap.cpp \
                   ../../../../plugin/jsbindings/manual/jsb_pluginx_extension_registration.cpp \
                   ../../../../plugin/jsbindings/auto/jsb_pluginx_umeng_auto.cpp \
                   ../../../../plugin/jsbindings/auto/jsb_pluginx_protocols_auto.cpp 

LOCAL_C_INCLUDES := $(LOCAL_PATH)/../../Classes \
                    $(LOCAL_PATH)/../../../../plugin/jsbindings \
                    $(LOCAL_PATH)/../../../../plugin/jsbindings/auto \
                    $(LOCAL_PATH)/../../../../plugin/jsbindings/manual 

LOCAL_WHOLE_STATIC_LIBRARIES := cocos2dx_static 
LOCAL_WHOLE_STATIC_LIBRARIES += cocosdenshion_static
LOCAL_WHOLE_STATIC_LIBRARIES += chipmunk_static
LOCAL_WHOLE_STATIC_LIBRARIES += spidermonkey_static
LOCAL_WHOLE_STATIC_LIBRARIES += scriptingcore-spidermonkey
LOCAL_WHOLE_STATIC_LIBRARIES += PluginProtocolStatic
[...]

PR #3885, opened 2013-10-08
issues #2677:Use CoreMotion's API to replace UIAccelerometer.

WenhaiLin:


minggo:
@Dhilan007 
I thin you should modify template Xcode project settings too.

WenhaiLin:
@minggo 
I modified complete.Thank you.

cocos2dx/platform/ios/CCDevice.mm
```diff
@@ -6,13 +6,13 @@
 #import <UIKit/UIKit.h>
 
 // Accelerometer
-#import <Foundation/Foundation.h>
-#import <UIKit/UIKit.h>
-#import <functional>
+#include<CoreMotion/CoreMotion.h>
```
minggo: I think it is better to use `import`.

PR #3888, opened 2013-10-08
Avoid unnecessary object duplication using passing arguments by reference

ghost:
In `Scale9Sprite` class, member function uses passing by value for Rect objects.
Using passing by value copies objects and it increases some amount of overhead.

In `Scale9Sprite` class, these object duplication is not necessary,
and adding `const &` keyword makes the meaning of the parameter clearer.


dumganhar:
Thanks.

PR #3891, opened 2013-10-08
add feature: cc.Sequence.create() and cc.Spawn.create() can pass Array actions

rablwupei:
eg: cc.Sequence.create([cc.Show.create(), cc.Delay.create(1), cc.Hide.create()]);


dumganhar:
Thanks.

PR #3893, opened 2013-10-08
HttpClient: set CURLOPT_ACCEPT_ENCODING if Accept-Encoding found in request headers

yyuu:
Set `CURLOPT_ACCEPT_ENCODING` if the `Accept-Encoding` header found in the request headers. This is required to enable libcurl's support of `Content-Encoding` (compression of HTTP contents).

### DIFF
--- a/extensions/network/HttpClient.cpp
+++ b/extensions/network/HttpClient.cpp
@@ -311,8 +311,8 @@ public:
             }
         }
 
-        if (request->getAcceptEncoding() != NULL) {
-          setOption(CURLOPT_ACCEPT_ENCODING, request->getAcceptEncoding());
+        if (const char *acceptEncoding = request->getAcceptEncoding()) {
+          setOption(CURLOPT_ACCEPT_ENCODING, acceptEncoding);
         }
 
         return setOption(CURLOPT_URL, request->getUrl())


extensions/network/HttpClient.cpp
```diff
@@ -311,6 +311,10 @@ class CURLRaii
             }
         }
 
+        if (request->getAcceptEncoding() != NULL) {
```
sergey-shambir: Please don't call this method twice, save result to local variable. There's also trick `if (const char *acceptEncoding = request->getAcceptEncoding()) {...}`.

yyuu: Thanks for reviewing :) I will update my patch as you suggested.

extensions/network/HttpRequest.h
```diff
@@ -214,7 +215,21 @@ class HttpRequest : public Object
...
+
+    /** Get the value of Accept-Encoding header **/
+    inline const char* getAcceptEncoding() {
+        for (std::vector<std::string>::iterator it = _headers.begin(); it != _headers.end(); ++it) {
+            int colon = it->find_first_of(":");
+            std::string key = it->substr(0, colon-1);
+            std::string val = it->substr(it->find_first_not_of(" \t", colon+1)); // strip leading whitespaces
+            std::transform(key.begin(), key.end(), key.begin(), ::tolower); // case insensitive
+            if(key.compare("accept-encoding") == 0) {
+                return val.c_str();
```
dumganhar, line 227: It's dangerous to return a pointer point to temporary variable(std::string val).

cocosQA:
Dear yyuu:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6481

PR #3894, opened 2013-10-08
fix memory leak in spine

rablwupei:


extensions/spine/SkeletonJson.cpp
```diff
@@ -240,7 +240,7 @@ SkeletonData* SkeletonJson_readSkeletonDataFile (SkeletonJson* self, const char*
 		return 0;
 	}
 	skeletonData = SkeletonJson_readSkeletonData(self, json);
-	FREE(json);
+    delete [] json;
```
minggo, line 243: But json is created by `NEW`, and the definition of `NEW` is 

```
   #define NEW(TYPE) CALLOC(TYPE,1)
```

So use `FREE` here is correct.

rablwupei, line 243: json create in `const char* json = _Util_readFile(path, &length);`

in this function it use:

```
char* _Util_readFile (const char* path, int* length) {
    unsigned long size;
    char* data = reinterpret_cast<char*>(FileUtils::getInstance()->getFileData(
        FileUtils::getInstance()->fullPathForFilename(path).c_str(), "r", &size));
    *length = size;
    return data;
}
```

`getFileData()` need `delete []`

minggo, line 243: Sorry, i thought `skeletonData`.
You are right, it is a memory leak.

pkxpp:
the new version is still FREE, not delete, why not modify?

PR #3895, opened 2013-10-08
fix bug: EditBox setText sometimes does not work

rablwupei:
when _systemControl.textField.hidden == NO


dumganhar:
Hi, could you tell me how to reproduce this bug? I couldn't reproduce it.

rablwupei:
oh, sorry, it have be fixed in v3.0.0.

PR #3908, opened 2013-10-10
Changed allKeys() to return an empty array instead of NULL

hawkwood:
Changed allKeys() to return an empty array instead of NULL if the dictionary exists but is empty.

http://www.cocos2d-x.org/forums/6/topics/36909?r=36914#message-36914


dumganhar:
Hi @hawkwood ,
We deprecated Dictionary since @3.0beta@, we will not continue to maintain these classes. 
We suggest you use the new @Template Container@ of which method @keys()@ will return an empty std::vectorstd::string. 
I will close this PR. Thanks all the same.

PR #3912, opened 2013-10-10
issue #2823: [JSB] Support remote debugging JSB project by FireFox.

dumganhar:
### NOT FINISHED
- Step into can't work.

### DIFF
--- a/scripting/javascript/bindings/ScriptingCore.cpp
+++ b/scripting/javascript/bindings/ScriptingCore.cpp
@@ -2014,6 +2014,136 @@ JSBool JSBDebug_getEventLoopNestLevel(JSContext* cx, unsigned argc, jsval* vp)
     return JS_TRUE;
 }
 
+//#pragma mark - Debugger
+
+static void _clientSocketWriteAndClearString(std::string& s)
+{
+    ::send(clientSocket, s.c_str(), s.length(), 0);
+    s.clear();
+}
+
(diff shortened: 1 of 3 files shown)


jdmunro:
@dumganhar I'm really keen on using this - how usable is it? Are there any preliminary instructions to try it?

dumganhar:
@jdmunro 
It's not finished yet. There is an issue about using Spidermonkey v24.
This Pull Request is still using Spidermonkey23, it will cause some problems (like the offset of js source is wrong) when debugging.
I will update the README.md after I fix this issue. 

I don't know whether some one else could help me fix it.
Anyway, I will also send a PR of upgrading spidermonkey to v24.
Thanks.

dumganhar:
@jdmunro 

I found that the problem of wrong offset of break point is caused by firefox in my Retina MBP 15inch.
Therefore, I think this PR could work.
I will update the README soon. THanks.

dumganhar:
@jdmunro 

README.md was updated. Please view it at https://github.com/cocos2d/cocos2d-x/blob/develop/scripting/javascript/bindings/js/debugger/README.md 
thanks.

jdmunro:
@dumganhar Great, thanks I'll take a look. At the minute we're working with a cocos2d-x 2.2 based project. Will there be any plans to port this functionality back to the 2.x branch or will it remain part of 3.0 only?

PR #3916, opened 2013-10-10
issue 2771: physical

boyu0:


cocos2dx/layers_scenes_transitions_nodes/CCScene.cpp
```diff
@@ -43,6 +44,7 @@ of this software and associated documentation files (the "Software"), to deal
 
 Scene::~Scene()
 {
+    CC_SAFE_DELETE(_physicsWorld);
```
dumganhar, line 48: need define `#ifdef CC_USE_PHYSICS` here.

PR #3917, opened 2013-10-10
support orientation change in android

ondesly:
Catching onContentRectChanged event in nativeactivity in Android to change screen size and notify app about orientation change.

applicationOrientationChanged method in CCApplicationProtocol to notify app about orientation change.


sachingarg05:
I haven't tested this but it seems to be doing it right. Thanks ondesly, I hope it gets merged soon.

minggo:
@ondesly 
You should implement ApplicationProtocol::applicationOrientationChanged() in all samples.

minggo:
It causes compiling errors: https://travis-ci.org/cocos2d/cocos2d-x/builds/12372533.

minggo:
And there is not implementation for iOS?

ondesly:
I forgot about samples and will make implementation for iOS too.

sachingarg05:
Why did you close this pull request? 

Samples don't have to support orientation change, they only need to compile to not break the build. This can be done by not making applicationOrientationChanged a pure virtual function.

Instead of:
   virtual void applicationOrientationChanged(int newWidth, int newHeight) = 0;
We can do:
   virtual void applicationOrientationChanged(int newWidth, int newHeight) { }

Actual orientation change support in samples can be added later.

Similarly, iOS support can be made as part of a separate pull request. It doesn't has to be all done together (wont break the build and these changes will only help when that is added).

It will be a big step forward if just this pull request got merged.

ondesly:
I think Application protocol should have only pure virtual functions.
I found a better way. I will put applicationOrientationChanged in Application of android and iOS only. And I found bug with current implementation.
I will make new pull request soon with all changes.

sachingarg05:
I don't want to complicate things more, but just want to share one idea.

On Windows and Mac Desktop apps, we can use the 'same' callback to support window resizing. All the 'core' implementation will stay the same, we will only need to add support to call this callback on Windows and Mac when windows size is changed. 

So for now, maybe it will be better if we name the callback function onSizeChanged instead of onOrientationChanged. Then this same code can be reused for window resizing also on Desktop apps.

ondesly:
Yeah. I am aim to do this. ScreenSizeChanged will be better than orientationChnaged

PR #3927, opened 2013-10-12
Master nutty optimizelayout

nutty898:


extensions/CocoStudio/GUI/BaseClasses/UIWidget.cpp
```diff
@@ -59,7 +59,8 @@
 m_nActionTag(0),
 m_size(CCSizeZero),
 m_customSize(CCSizeZero),
-m_pLayoutParameter(NULL),
+//m_pLayoutParameter(NULL),
+m_pLayoutParameterDictionary(NULL),
```
dumganhar, line 62: If `m_pLayoutParameter` is not needed, please remove it.

PR #3937, opened 2013-10-13
Buildscript to install dependencies for linux path fix

ghost:
Running build/install-deps-linux.sh gave the following error:
tools/travis-scripts/install_glfw.sh: No such file or directory

Probably because the install script was moved into the build/ dir and the path to the travis script wasnt updated.


ghost:
Seems a lot more build scripts arent working at the moment.

PR #3938, opened 2013-10-14
Added triangle, quad bezier, cubic bezier drawing to DrawNode

elipsitz:
I'm personally using the triangle method in order to draw triangulated convex polygons. 

The quad and cubic beziers, are, of course, very useful for a variety of applications. I cannot guarantee that the cubic one will work with self intersecting beziers.


dumganhar:
Thanks. I will check it. Could you also add some tests to `TestCpp/DrawingPrimitivesTest`?

elipsitz:
Of course, I've added a few simple tests to DrawPrimitivesTest

dumganhar:
Merged. Thank you.

PR #3963, opened 2013-10-17
Change the class name of textfield to compare, so that you can touch any...

kpkhxlgy0:
...where outside of the virtual keyboard.


dumganhar:
Thank you.

PR #3969, opened 2013-10-18
fix tolua_fix, cleanup peertable after object deleted

dualface:
``` lua
local scene = CCScene:create()
CCDirector:sharedDirector():runWithScene(scene)

local node = CCNode:create()
scene:addChild(node)
node.t = 1
node:removeFromParentAndCleanup(true)

print(node.t)

scene:scheduleUpdate(function()
    print(node.t)
    scene:unscheduleUpdate()
end)

```

without this path, output is:

```
 - 1
 - 1
```

with this path, output is:

```
 - 1
 - nil
```


dualface:
fix Lua 5.2 compatibility

PR #3974, opened 2013-10-18
issue 2771: physical

boyu0:


cocos/physics/CCPhysicsBody.h
```diff
@@ -43,7 +43,7 @@
 class PhysicsBodyInfo;
 
 
-const PhysicsMaterial PHYSICSBODY_MATERIAL_DEFAULT(0.0f, 1.0f, 1.0f);
+const PhysicsMaterial PHYSICSBODY_MATERIAL_DEFAULT = {1.0f, 1.0f, 1.0f};
```
dumganhar, line 46: Windows doesn't support `initialized list`, so please revert it.

boyu0, line 46: PhysicsMaterial is a struct, so it's ok.

PR #3976, opened 2013-10-19
replace free with CC_SAFE_FREE, avoid wild pointer.

joewan:
void CCScheduler::removeHashElement(_hashSelectorEntry *&pElement)
{
     cocos2d::CCObject *target = pElement->target;

```
ccArrayFree(pElement->timers);
HASH_DEL(m_pHashForTimers, pElement);
CC_SAFE_FREE(pElement);

// make sure the target is released after we have removed the hash element
// otherwise we access invalid memory when the release call deletes the target
// and the target calls removeAllSelectors() during its destructor
target->release();
```

}

replace free with CC_SAFE_FREE, avoid wild pointer. when passing a pointer into a free- memory-function, we need to pass the reference of the pointer.

### DIFF
--- a/cocos2dx/actions/CCActionManager.cpp
+++ b/cocos2dx/actions/CCActionManager.cpp
@@ -87,7 +87,7 @@ void CCActionManager::actionAllocWithHashElement(tHashElement *pElement)
 
 }
 
-void CCActionManager::removeActionAtIndex(unsigned int uIndex, tHashElement *pElement)
+void CCActionManager::removeActionAtIndex(unsigned int uIndex, tHashElement *&pElement)
 {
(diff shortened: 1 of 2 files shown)


joewan:
I think we need to refactor these functions, such as removeActionAtIndex && deleteHashElement. it is dirty and dangerous.

dumganhar:
Yep, I agree with your point.
But could you adding test case to `BugTest` in `TestCpp` to reproduce this issue?
Thanks.

joewan:
this is a tiny "synatx" problem, it does not cause crashes or problems now,  but it will need the user know more when to use these functions.

PR #3980, opened 2013-10-19
fixed bug that it reset content offset when reloadData be called

HoGarfield:


extensions/GUI/CCScrollView/CCTableView.cpp
```diff
@@ -103,7 +103,6 @@ void TableView::setVerticalFillOrder(VerticalFillOrder fillOrder)
 
 void TableView::reloadData()
 {
-    _oldDirection = Direction::NONE;
```
dumganhar, line 106: Hi, is this patch correct?
IMO, we need to reset `_oldDirection` since `reloadData` will re-arrange offset of each cells in `_updateContentSize`.

PR #3981, opened 2013-10-19
fixed bug of WebSocket that it can not receive over 4096 byte

HoGarfield:


dualface:
thanks HoGarfield

edwardzhou:
this patch only fix websocket cannot receive 4096+ data. but still has problem on sending 4096+ data.

cocos/network/WebSocket.cpp
```diff
@@ -221,6 +221,9 @@ enum WS_MSG {
 , _delegate(nullptr)
 , _SSLConnection(0)
 , _wsProtocols(nullptr)
+,_pending_frame_data_len(0)
+,_current_data_len(0)
+,_current_data(NULL)
```
minggo, line 226: Please use `_currentData` instead to obey [cocos2d-x coding style].
And please use `nullptr` instead. We use c++11 in develop branch.

Thanks.

edwardzhou:
can please also include the send 4096+ bytes fix as well, ref: https://github.com/cocos2d/cocos2d-x/pull/4014

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/5008. Thanks.

PR #3985, opened 2013-10-20
1.release array for uihelper; 2.fix bug.

joewan:
sometimes we call CCSpriteFrameCache::purgeSharedSpriteFrameCache in other places,
UIHelper still keeps the filenames in "UIHelper::m_textureFiles" and then we can not addSpriteFrame successfully for these files again.


dumganhar:
UIHelper has been refactored . So this PR is invalid now. I closed it.

PR #3987, opened 2013-10-21
Fix keyboard enabled flag

v1ctor:
This function setKeyboardEnabled(bool enabled) don't work when layer shows the second time, but keyboard listener is removed in function onExit().


sergey-shambir:
Nice change, +1 from me.

dumganhar:
We have fixed this issue.
All event listeners would be paused.
You could refer to https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/CCNode.cpp#L1094 .
So this patch is not needed now.
I closed this PR. Thanks.

PR #3997, opened 2013-10-23
android & ios screen size change support

ondesly:
- applicationScreenSizeChanged method only for ios and android
- nativeactivity in android waits for onContentRectChanged and after 30 ms calls applicationScreenSizeChanged
- didRotateFromInterfaceOrientation in ios calls applicationScreenSizeChanged

### DIFF
--- a/cocos/2d/platform/android/CCApplication.cpp
+++ b/cocos/2d/platform/android/CCApplication.cpp
@@ -128,4 +128,8 @@ Application::Platform Application::getTargetPlatform()
     return Platform::OS_ANDROID;
 }
 
+void Application::applicationScreenSizeChanged(int newWidth, int newHeight) {
+
+}
+
 NS_CC_END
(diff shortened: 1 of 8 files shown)


sachingarg05:
@minggo I tested this a bit and it seems to work ok. Can we please have this merged before the next alpha release?

minggo:
@sachingarg05
Sure, i will test it and merge it.
Thanks for notifying.

@ondesly 
Good job.

ondesly:
@minggo 
Thank you! It's my first pull request. I'm very excited.

sachingarg05:
@ondesly @minggo Now that native activity is gone from 3.0, we will need to add a JNI function call in Java onSurfaceChanged, and that JNI function will need to call our new cocos2d::Application::getInstance()->applicationScreenSizeChanged(newWidth, newHeight);

minggo:
@sachingarg05 
We will support it.
Thanks.

PR #4008, opened 2013-10-25
issue #3069: Refactor and Improve EventDispatcher

dumganhar:
Changes contain:
- `EventListenerTouch` has to be separated to `EventListenerTouchOneByOne` and `EventListenerTouchAllAtOnce`, it will be clearer for developer to decide to use which type of touch event listener.
- Remove all controller codes from `cocos2d::Layer`. `Layer` should not be a controller. All event listener registrations should use `EventDispatcher`.
- Don’t insert member variables which are not relevant to `Node`. `Node` should be as **small**, **fast** as possible.
- Improve the **SORT** logic of `EventListener`, we should divide event listeners into two parts and store them in different vectors. One is for `fixed priority event listeners`, another is for `scene graph based priority listeners`. In this way, when a listener’s priority is changed, only one of the vectors should be sorted.
- Use `int` as key instead of `string` in listener map.
- EventDispatcher is managed by Director.

### DIFF
--- a/cocos/2d/platform/android/nativeactivity.cpp
+++ b/cocos/2d/platform/android/nativeactivity.cpp
@@ -422,18 +422,20 @@ static int32_t handle_key_input(AInputEvent *event)
 {
     if (AKeyEvent_getAction(event) == AKEY_EVENT_ACTION_UP)
     {
+        auto dispatcher = cocos2d::Director::getInstance()->getEventDispatcher();
+
         switch (AKeyEvent_getKeyCode(event))
         {
         case AKEYCODE_BACK:	
             {
                 cocos2d::EventKeyboard event(cocos2d::EventKeyboard::KeyCode::KEY_BACKSPACE, false);
-                cocos2d::EventDispatcher::getInstance()->dispatchEvent(&event);
+                dispatcher->dispatchEvent(&event);
             }
             return 1;
         case AKEYCODE_MENU:
             {
                 cocos2d::EventKeyboard event(cocos2d::EventKeyboard::KeyCode::KEY_MENU, false);
-                cocos2d::EventDispatcher::getInstance()->dispatchEvent(&event);
+                dispatcher->dispatchEvent(&event);
             }
             return 1;
         default:
@@ -629,8 +631,8 @@ void android_main(struct android_app* state) {
                             acc.z = event.acceleration.z/10;
                             acc.timestamp = 0;
                             cocos2d::EventAcceleration accEvent(acc);
-
-                            cocos2d::EventDispatcher::getInstance()->dispatchEvent(&accEvent);
+                            auto dispatcher = cocos2d::Director::getInstance()->getEventDispatcher();
+                            dispatcher->dispatchEvent(&accEvent);
                         } else {
                             // ACONFIGURATION_ORIENTATION_LAND
                             // swap x and y parameters
@@ -640,8 +642,8 @@ void android_main(struct android_app* state) {
                             acc.z = event.acceleration.z/10;
                             acc.timestamp = 0;
                             cocos2d::EventAcceleration accEvent(acc);
-
-                            cocos2d::EventDispatcher::getInstance()->dispatchEvent(&accEvent);
+                            auto dispatcher = cocos2d::Director::getInstance()->getEventDispatcher();
+                            dispatcher->dispatchEvent(&accEvent);
                         }
                     }
                 }
--- a/cocos/2d/platform/linux/CCEGLView.cpp
+++ b/cocos/2d/platform/linux/CCEGLView.cpp
@@ -223,7 +223,8 @@ void EGLViewEventHandler::OnGLFWMouseMoveCallBack(GLFWwindow* window, double x, 
 void EGLViewEventHandler::OnGLFWKeyCallback(GLFWwindow *window, int key, int scancode, int action, int mods)
 {
     EventKeyboard event(g_keyCodeMap[key], GLFW_PRESS == action);
-    EventDispatcher::getInstance()->dispatchEvent(&event);
+    auto dispatcher = Director::getInstance()->getEventDispatcher();
+    dispatcher->dispatchEvent(&event);
 }
 
 void EGLViewEventHandler::OnGLFWCharCallback(GLFWwindow *window, unsigned int character)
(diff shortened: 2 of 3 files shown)


dumganhar:
@ricardoquesada @minggo

cocos/2d/CCNode.cpp
```diff
@@ -935,7 +932,8 @@ void Node::onEnter()
     arrayMakeObjectsPerformSelector(_children, onEnter, Node*);
 
     this->resumeSchedulerAndActions();
-
+    EventDispatcher::getInstance()->resumeTarget(this);
```
ricardoquesada: perhaps this function could be merged with `resumeScheduleAndActions` and only have one named `resume`.
The same for pause.

cocos/2d/CCEvent.h
```diff
@@ -40,15 +40,25 @@
  */
 class Event
 {
+public:
+    enum EventType
```
winipcfg: Why not using enum class?

dumganhar: Hi @winipcfg ,
I didn't use `enum class` was because I changed the key of listener map from `string` to `int`.
So the custom event had to use `int` as key. I defined a normal enumeration like:

```
    enum EventType
    {
        TYPE_TOUCH = 1,
        TYPE_KEYBOARD = 3,
        TYPE_ACCELERATION,
        TYPE_CUSTOM
    };
```

The custom event should be larger than  or equal to `TYPE_CUSTOM`.
But the `enum class` is strong type.

```
EventType a = EventType::CUSTOM + 1;
```

will not work.

I thought about this a lot. I think custom event should not use integer for initialization. 
Instead, it should use `string` to initialize, then, convert it to `hash code` by `std::hash`.
In this way, `enum class` will make sense.

Thanks for your advise. I will update this PR soon.

dumganhar: There is still an issue should be resolved before using `enum class`. If it could not be fixed, `enum class` will not make sense. The issue is:

``` cpp
class Event
{
public:
    enum class Type
    {
        TOUCH = 1,
        KEYBOARD = 3,
        ACCELERATION,
        CUSTOM
    };

protected:
    /** Constructor */
    Event(Type type);
public:
    /** Destructor */
    virtual ~Event();

    /** Gets the event type */
    inline Type getType() const { return _type; };
...
Type _type;     ///< Event type
...
};
```

``` cpp
EventCustom::EventCustom(const std::string& eventName)
: Event(std::hash<std::string>()(eventName))
, _userData(nullptr)
{
}
```

``` cpp
class EventListener : public Object
{
public:
    enum class Type
    {
        TOUCH_ONE_BY_ONE = 1,
        TOUCH_ALL_AT_ONCE,
        KEYBOARD,
        ACCELERATION,
        CUSTOM
    };
```

``` cpp
class EventDispatcher
{
    /** Listeners map */
    std::unordered_map<EventListener::Type, EventListenerVector*> _listeners;

    /** The map of dirty flag */
    std::unordered_map<EventListener::Type, DirtyFlag> _priorityDirtyFlagMap;
};
```

`Event::getType` will not get what's defined in enum class `Event::Type` since custom event should convert string to hash code which is the type.
The same problem appears in `EventListener::Type`.
Any suggestion?

dumganhar: `Event::getType` will make no sense.

cocos/2d/CCNode.cpp
dumganhar: Hi, Riq, I Updated this PR, please review again! :)

cocos/2d/CCEvent.h
winipcfg: Don't get the point. Any events other than default types are "Custom" and should return Type::CUSTOM
Are you trying to create event types Custom1, Custom2, etc? Is it necessary to separate event types Custom1, Custom2, etc in std::unordered_map?

EventType a = EventType::CUSTOM + 1;

btw, this will give error but I guess it can be bypassed by using static_cast. Not a good way though

dumganhar: Sample codes of Custom Event:

``` cpp
    _listener = EventListenerCustom::create("game_custom_event1", [=](EventCustom* event){
        log("Custom event 1 received, ");
    });

    _eventDispatcher->addEventListenerWithFixedPriority(_listener, 1);

    auto sendItem = MenuItemFont::create("Send Custom Event 1", [=](Object* sender){
        EventCustom event("game_custom_event1");
        event.setUserData(xxx);
        _eventDispatcher->dispatchEvent(&event);
    });
```

> Are you trying to create event types Custom1, Custom2, etc? Is it necessary to separate event types Custom1, Custom2, etc in std::unordered_map?

Yep, in the current implementation, it uses different type for custom listener in std::unordered_map.
If we use the same type `Event::Type::CUSTOM`, then any kind of custom events will be sent to callback.
For instance,
I created a `game_custom_event1`, `game_custom_event2`... listeners like above.
Then the callback of `game_custom_event1` will also response the `game_custom_event2` since we only have one type for Custom event.

dumganhar: Any suggestions?

dumganhar: > EventType a = EventType::CUSTOM + 1;
> btw, this will give error but I guess it can be bypassed by using static_cast. Not a good way though
> Force conversion seems not to be elegant.

The string for initialize `EventCustom` will be converted to `hash code`.

``` cpp
EventCustom::EventCustom(const std::string& eventName)
: Event(std::hash<std::string>()(eventName))
, _userData(nullptr)
{
}

EventListenerCustom* EventListenerCustom::create(const std::string& eventName, std::function<void(EventCustom*)> callback)
{
    EventListenerCustom* ret = new EventListenerCustom();
    if (ret && ret->init(std::hash<std::string>()(eventName), callback))
    {
        ret->autorelease();
    }
    else
    {
        CC_SAFE_DELETE(ret);
    }
    return ret;
}

```

Do you mean the `Event::getType` needs to return `Event::Type::Custom`, but in the std::unordered_map, it still uses the hash code for key?

winipcfg: Yes. "game_custom_event1", "game_custom_event2", etc are custom events and they should return Event::Type::Custom. Otherwise, Event::Type is meaningless.

But if you only want to return all "game_custom_event1", it is supposed std::unordered_map use the hashcode/name as the key value.

dumganhar: @winipcfg 

I updated this PR, and used enum class for event type and listener type.
Could you check that here https://github.com/dumganhar/cocos2d-x/commit/d3afa5028979e3228b20c53b35b2282b6965b1a5 ?

I also added another variable called `EventListener::ListenerID _listenerID` for the key of `std::unordered_map`.

What do you think this change?

winipcfg: Just checked that unordered_map already use std::hash as hash function of std::string. No further comment other than that.:-)
http://stackoverflow.com/questions/15595804/c-hash-function-for-string-in-unordered-map
http://en.cppreference.com/w/cpp/utility/hash

dumganhar: Thank you. :) Got it.

PR #4010, opened 2013-10-25
CCRects converted with jsb does not create origin/size properties

luisparravicini:
When using javascript through jsb, a call to getBoundingBox() returns an object with properties x/y/height/width and not origin/size properties.
Submitted as issue [3065](http://www.cocos2d-x.org/issues/3065)
cocos2d-js-tests pull request: [231](https://github.com/cocos2d/cocos2d-js-tests/pull/231)

### DIFF
--- a/cocos/scripting/javascript/bindings/ScriptingCore.cpp
+++ b/cocos/scripting/javascript/bindings/ScriptingCore.cpp
@@ -1799,10 +1799,9 @@ jsval ccacceleration_to_jsval(JSContext* cx, const Acceleration& v) {
 jsval ccrect_to_jsval(JSContext* cx, const Rect& v) {
     JSObject *tmp = JS_NewObject(cx, NULL, NULL, NULL);
     if (!tmp) return JSVAL_NULL;
-    JSBool ok = JS_DefineProperty(cx, tmp, "x", DOUBLE_TO_JSVAL(v.origin.x), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
-                JS_DefineProperty(cx, tmp, "y", DOUBLE_TO_JSVAL(v.origin.y), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
-                JS_DefineProperty(cx, tmp, "width", DOUBLE_TO_JSVAL(v.size.width), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
-                JS_DefineProperty(cx, tmp, "height", DOUBLE_TO_JSVAL(v.size.height), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT);
+    JSBool ok = JS_DefineProperty(cx, tmp, "size", ccsize_to_jsval(cx, v.size), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
+                JS_DefineProperty(cx, tmp, "origin", ccpoint_to_jsval(cx, v.origin), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT);
+
     if (ok) {
         return OBJECT_TO_JSVAL(tmp);
     }


cocos/scripting/javascript/bindings/ScriptingCore.cpp
```diff
@@ -1799,10 +1799,9 @@ jsval ccacceleration_to_jsval(JSContext* cx, const Acceleration& v) {
 jsval ccrect_to_jsval(JSContext* cx, const Rect& v) {
     JSObject *tmp = JS_NewObject(cx, NULL, NULL, NULL);
     if (!tmp) return JSVAL_NULL;
-    JSBool ok = JS_DefineProperty(cx, tmp, "x", DOUBLE_TO_JSVAL(v.origin.x), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
-                JS_DefineProperty(cx, tmp, "y", DOUBLE_TO_JSVAL(v.origin.y), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
-                JS_DefineProperty(cx, tmp, "width", DOUBLE_TO_JSVAL(v.size.width), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
-                JS_DefineProperty(cx, tmp, "height", DOUBLE_TO_JSVAL(v.size.height), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT);
+    JSBool ok = JS_DefineProperty(cx, tmp, "size", ccsize_to_jsval(cx, v.size), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT) &&
+                JS_DefineProperty(cx, tmp, "origin", ccpoint_to_jsval(cx, v.origin), NULL, NULL, JSPROP_ENUMERATE | JSPROP_PERMANENT);
+
```
dumganhar, line 1804: Hi @luisparravicini ,
We should not delete `x, y, width, height` directly and use `origin, size` . That will break the compatibility of JSB.
Do you have any ideas about keep the compatibility?

litefeel:
Do not change API of cocos2dx in all language.

cocos/scripting/javascript/bindings/ScriptingCore.cpp
luisparravicini, line 1804: Keeping x/y/height/width and adding size/origin would be exactly like what getBoundingBox() returns in js.

The thing is, (maybe because this is my first attempt at modifying cocos2d and not knowing the inner workings of it) but I don't really know why a ccrect in this case is not using size/origin.

Can you take a look at my last note on http://www.cocos2d-x.org/issues/3065#note-3 ?
Thanks

ricardoquesada, line 1804: From the API point of view, I think it is simpler to use `x, y, width, height` directly.
Although `origin` and `size` are correct, the user has to type more.

So, instead of fixing `CCRect` and I would fix the return value `getBoundingBox()`

I do not know if that can be done automatically with the bindings... if the return value is `CCRect`, then return `x,y,width,height`.

luisparravicini, line 1804: Sorry I'm doing too many questions: if `CCRects` (when used through jsb) have to return `x, y, width, height`, which is what's happening now (not taking into account this pr), why not leave `getBoundingBox()` as is but change in js `cc._RectApplyAffineTransformIn` to just don't return `origin` and `size`? That's what made me create the issue and this patch in the first place (unless such a change in the API is unacceptable).

ricardoquesada, line 1804: Thanks Luis.
Tomorrow I'll review our code... I'll give you a detailed answer tomorrow.

ricardoquesada, line 1804: I added this bug report about regarding this:
http://www.cocos2d-x.org/issues/3260

cocosQA:
Dear luisparravicini:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6480

PR #4012, opened 2013-10-26
include climits for UINT_MAX

ledyba:
UINT_MAX is used in CCPhysBody.cpp, but climits header is not included.


sergey-shambir:
I'm agree, build failed for me too due to UINT_MAX.

dumganhar:
Thanks.

PR #4014, opened 2013-10-27
fix WebSocket cannot send/receive more than 4096 bytes of data

edwardzhou:


cocos/network/WebSocket.cpp
```diff
@@ -511,44 +514,62 @@ int WebSocket::onSocketCallback(struct libwebsocket_context *ctx,
...
-                        
-                        bytesWrite = libwebsocket_write(wsi,  &buf[LWS_SEND_BUFFER_PRE_PADDING], data->len, writeProtocol);
-                        
-                        if (bytesWrite < 0)
+                        // Do we have another fragments to send?
+                        else if (remaining != n)
                         {
-                            CCLOGERROR("%s", "libwebsocket_write error...");
+                            data->issued += n;
+                            break;
```
dumganhar, line 560: @edwardzhou 
Should this line be `continue;` rather than `break;`?
How to test this issue?

edwardzhou, line 560: should break. because we have to waiting for the next LWS_CALLBACK_CLIENT_WRITEABLE message for the next write.

test case: send out a 4k+ data (eg, 10k)

dumganhar, line 560: Yep, I test it. But `LWS_CALLBACK_CLIENT_WRITEABLE` was called only once.

edwardzhou, line 560: I have tested it before sent out pull request. I added many CCLOG for trace, but cleaned for pull request.

dumganhar:
Fixed.

PR #4016, opened 2013-10-28
Update UITextField.h

rongxiong:
corrected word error


minggo:
@iSongWorld 
I think you should send a pull request again to include the modification for UITextField.cpp and UITextField.h.

Thanks.

PR #4019, opened 2013-10-28
Clear NoSuchMethodError Exception when JniHelper fails to find methodID

superrad:
When calling GetMethodID a NoSuchMethodError will be raised if the method cannot be found.
See http://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp16660
This exception should be cleared if JNI execution is to be continued.
Currently JniHelper informs that the method is not found, but there is no way to continue execution without manually clearing the exception. This prevents checking for the existence of a method and recovering from it being missing.

This fix will clear the exception if JniHelper fails to find the method.


minggo:
Thanks.

PR #4021, opened 2013-10-29
update armature

2youyou2:
1.improve performance
2.add gotoAndPlay method
3.fix asynchronous reading error
4.change CCArmature::boundingBox to apply CCNode::boundingBox
5.fix different armatures may play the same movement on some android device

### DIFF
--- a/extensions/CocoStudio/Armature/display/CCDisplayFactory.cpp
+++ b/extensions/CocoStudio/Armature/display/CCDisplayFactory.cpp
@@ -253,6 +253,15 @@ void CCDisplayFactory::createParticleDisplay(CCBone *bone, CCDecorativeDisplay *
 {
     CCParticleDisplayData *displayData = (CCParticleDisplayData *)decoDisplay->getDisplayData();
     CCParticleSystem *system = CCParticleSystemQuad::create(displayData->plist.c_str());
+    
+    system->removeFromParent();
+    
+    CCArmature *armature = bone->getArmature();
+    if (armature)
+    {
+        system->setParent(bone->getArmature());
+    }
+
     decoDisplay->setDisplay(system);
 }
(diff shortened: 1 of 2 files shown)


extensions/CocoStudio/Armature/CCBone.cpp
```diff
@@ -74,6 +74,8 @@ CCBone *CCBone::create(const char *name)
     m_bBoneTransformDirty = true;
     m_eBlendType = BLEND_NORMAL;
     m_tWorldInfo = NULL;
+    m_pArmatureParentBone = NULL;
+    m_fDataVersion = 0;
```
xingsenma, line 78: Suggest dataVersion is used  as a global variables.

extensions/CocoStudio/Armature/utils/CCSpriteFrameCacheHelper.h
```diff
@@ -47,14 +47,11 @@ class  CCSpriteFrameCacheHelper
      */
     void addSpriteFrameFromFile(const char *plistPath, const char *imagePath);
 
-    CCTextureAtlas *getTexureAtlasWithTexture(CCTexture2D *texture);
```
xingsenma, line 50: I think "CCSpriteFrameCacheHelper.h" this file can be delete from armature.

extensions/CocoStudio/Armature/CCBone.cpp
2youyou2, line 78: There is a global dataVersion for current CCArmature's version
This one is the data version of the CocoStudio Editor.

extensions/CocoStudio/Armature/utils/CCSpriteFrameCacheHelper.h
2youyou2, line 50: Yes, this class is useless now.

PR #3048, opened 2013-07-02
Iss2328 refractor etc

dabingnn:
issue 2328 Etc support refractor
1. C++ decode etc
2. add supportsETC() function to configuration
3. change test case, to allow all platform read pkm file
4. fix bug Texture2D::_pixelFormat not initialized in constructor!

PR #4033, opened 2013-10-31
Closed #3048: cc.rect can receiving different type parameters now

dingpinglv:
cc.rect can receiving different type parameters now

### DIFF
--- a/cocos/scripting/javascript/script/jsb_cocos2d.js
+++ b/cocos/scripting/javascript/script/jsb_cocos2d.js
@@ -231,25 +231,26 @@ cc.sizeEqualToSize = function (size1, size2)
 };
 
 /**
- * create a rect object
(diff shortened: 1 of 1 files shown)


cocos/scripting/javascript/script/jsb_cocos2d.js
```diff
@@ -230,12 +230,29 @@ cc.sizeEqualToSize = function (size1, size2)
     return ((size1.width == size2.width) && (size1.height == size2.height));
 };
 
-//
-// Rect
-//
+/**
+ * create a rect object
+ * @param {Number|cc.Point|cc.Rect} [x1] a Number value as x or a cc.Point object as origin or a cc.Rect clone object
```
dumganhar: It should be `cc.point` and `cc.rect`. There are not `cc.Point` in JSB.

dumganhar:
Thanks.

PR #4035, opened 2013-10-31
MouseEvent and MouseListener

darkdukey:
Add MouseEvent and Mouse Listener to handle mouse input in desktop systems.

### DIFF
--- a/cocos/2d/Android.mk
+++ b/cocos/2d/Android.mk
@@ -39,6 +39,8 @@ CCEventAcceleration.cpp \
 CCEventCustom.cpp \
 CCEventDispatcher.cpp \
 CCEventKeyboard.cpp \
+CCEventMouse.cpp \
+CCEventListenerMouse.cpp \
 CCEventListener.cpp \
 CCEventListenerAcceleration.cpp \
 CCEventListenerCustom.cpp \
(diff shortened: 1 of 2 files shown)


dumganhar:
Hi @darkdukey ,
This PR looks great, thanks in advance. :)

cocos/2d/CCEventListenerMouse.h
```diff
@@ -0,0 +1,57 @@
...
+    static EventListenerMouse* create();
+
+    /// Overrides
+    virtual EventListenerMouse* clone() override;
+    virtual bool checkAvailable() override;
+
+    std::function<void(Event* event)> onMouseDown;
+    std::function<void(Event* event)> onMouseUp;
+    std::function<void(Event* event)> onMouseMove;
+    std::function<void(Event* event)> onMouseScroll;
```
dumganhar, line 48: What do you think of defining the callback to `std::function<void(EventMouse* event)>` to avoid cast?

darkdukey, line 48: Sounds good,

Can IDE (xcode, VS) recognize void(EventMouse\* event) and show user hints ? 

-Nite

dumganhar, line 48: No, but if you use a different type, compiler will trigger an error. So I think it isn't relevant to change it `std::function<void(EventMouse* event)>`. What do you think?

dumganhar:
Thank you.

PR #4051, opened 2013-11-01
Fixed dead recursive when CCUserDefault.destoryInstance function called in IOS platform

HoGarfield:


cocos/2d/CCUserDefault.mm
```diff
@@ -503,7 +503,6 @@ static void deleteNodeByKey(const char *pKey)
 
 void UserDefault::destroyInstance()
 {
-    delete _userDefault;
     _userDefault = NULL;
```
dumganhar, line 506: Hi, Thanks.
But I think we should delete `_userDefault` in `UserDefault::destroyInstance`.

``` cpp
UserDefault::~UserDefault()
{
    CC_SAFE_DELETE(_userDefault);   // This line should be deleted.
    _userDefault = NULL;
}
```

dumganhar:
Fixed at current develop branch.

PR #4052, opened 2013-11-01
Update FmodAudioPlayer.cpp

denisacostaq:
If you run pauseBackgroundMusic and then isBackgroundMusicPlaying (), this returns true when it should return false (I think), This can be tested in the demos:
cocos2d-x-2.2.0/samples/Cpp/TestCpp/Classes/CocosDenshionTest

Also if you run playBackgroundMusic (const char \* pszFilePath, bool bloop) with bloop made ​​false returns true call to isBackgroundMusicPlaying () after finishing the music.

This short patch corrects pequenno. :)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,4 @@
-cocos2d-x authors & contributors
+co cos2d-x authors & contributors
 
 (ordered by the join in time)
 


cocosQA:
Dear denisacostaq:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6478

PR #4059, opened 2013-11-01
remove eventlistener in onExit

mosquitodu:
we need remove event listener in onExit, because in onEnter will add event listener

### DIFF
--- a/cocos/2d/CCMenu.cpp
+++ b/cocos/2d/CCMenu.cpp
@@ -58,7 +58,12 @@ enum
 //
 //CCMenu
 //
+Menu::Menu()
+:_selectedItem(NULL)
+,_touchListener(NULL)
+{
 
+}
 Menu::~Menu()
 {
     CCLOGINFO("In the destructor of Menu. %p", this);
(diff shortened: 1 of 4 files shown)


dumganhar:
Thanks. But why did you change file mode to 755?

cocos/2d/CCMenu.cpp
```diff
@@ -184,16 +189,18 @@ void Menu::addChild(Node * child, int zOrder, int tag)
...
+        _touchListener = EventListenerTouchOneByOne::create();
+        _touchListener->setSwallowTouches(true);
+        
+        _touchListener->onTouchBegan = CC_CALLBACK_2(Menu::onTouchBegan, this);
+        _touchListener->onTouchMoved = CC_CALLBACK_2(Menu::onTouchMoved, this);
+        _touchListener->onTouchEnded = CC_CALLBACK_2(Menu::onTouchEnded, this);
+        _touchListener->onTouchCancelled = CC_CALLBACK_2(Menu::onTouchCancelled, this);
+        
+        _eventDispatcher->addEventListenerWithSceneGraphPriority(_touchListener, this);
+    }
```
dumganhar, line 203: I think we should move 'touch event registration' to 'Menu::initWithArray'. In that way, this issue could be fixed.

mosquitodu:
Sorry, I don’t change it, it changed by my compare tool.
在 2013年11月1日，下午5:47，James Chen notifications@github.com 写道：

cocos/2d/CCMenu.cpp
mosquitodu, line 203: yes, I agree with you, we should do add event listener in init function, such as in Menu initWithArray not onEnter, i am not family the new event dispatcher, if we move to init, now anther scene pushed, the old one also can catch the event? but in old way, onExit we may be remove the event register?

dumganhar, line 203: > if we move to init, now anther scene pushed, the old one also can catch the event? but in old way, onExit we may be remove the event register?

We have refactored the `EventDispatcher` and added `EventDispatcher::pauseTarget, resumeTarget` which are invoked in `Node::pause, resume`, so moving the event registration to `init` will make sense. The event listener will be paused to stop receiving events.

dumganhar, line 203: fixed at https://github.com/cocos2d/cocos2d-x/pull/4065 .
Could you test it ?

PR #4069, opened 2013-11-02
EventDispatcher::removeEventListener failed to remove listener somtime.

ledyba:
_toAddedListener list is not searched in EventDispatcher::removeEventListener,
so this function often failed to remove event listener.


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/4072 . Thanks.

PR #4073, opened 2013-11-03
pXmlBuffer is not null terminated

ledyba:
pXmlBuffer is not null-terminated, so you have to pass size parameter to XmlDocument::Parse.


dumganhar:
Thank you. Could you also send a PR to master branch?

ledyba:
here is a new pull request https://github.com/cocos2d/cocos2d-x/pull/4074

dumganhar:
Thanks. Merged into develop branch.

PR #4076, opened 2013-11-04
Master nutty fixwrongcode linux

nutty898:
fixed linux build error


extensions/CocoStudio/GUI/UIWidgets/ScrollWidget/UIListViewEx.cpp
```diff
@@ -294,9 +294,9 @@ void UIListViewEx::removeLastItem()
     removeItem(m_pItems->count() -1);
 }
 
-UIWidget* UIListViewEx::getItem(int index)
+UIWidget* UIListViewEx::getItem(unsigned int index)
```
minggo, line 297: It is better to use `int`. So don't have to change it.

PR #4086, opened 2013-11-05
Project is 64-bit ARM friendly

ricardoquesada:
Still doesn't compile in ARM64 due to missing libraries in ARM64 mode, but at least cocos2d compiles file.

Also this patch enables a very important flag in Xcode:
- 64-to-32-bit conversion warning.

There are more than 200 warnings because of this issue. And if we don't fix those warnings we could have serious bugs.
( I already fixed some of them )

### DIFF
--- a/cocos/2d/CCFont.cpp
+++ b/cocos/2d/CCFont.cpp
@@ -75,7 +75,7 @@ void Font::setCurrentGlyphCollection(GlyphCollection glyphs, const char *customG
         default:
             if (customGlyphs)
             {
-                int lenght = strlen(customGlyphs);
+                size_t lenght = strlen(customGlyphs);
                 _customGlyphs = new char [lenght + 2];
                 memcpy(_customGlyphs, customGlyphs, lenght);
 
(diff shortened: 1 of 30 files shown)


minggo:
So did you mean we will not use int any more?

ricardoquesada:
@minggo 
We should use `int` where it makes sense...

As a rule, we should use `long` for "generic" code.

As an example, the code that was ported from cocos2d-iphone that was using `NSInteger`, in cocos2d-x should be `long`.

A `long` uses 64-bit registers in 64-bit architectures, and 32-bit register in 32-bit architectures.
On the other hand, `int` is 32-bit registers both in 32-bit and 64-bit architectures.

minggo:
@ricardoquesada 
Got it.

dakikifr:
@ricardoquesada 

This is not true ("long uses 64 bits registers in 64 bits architecture").

'long' sizeof is not consistant across platforms and compilers.
For example, using MSVC, a long is always 4 bytes even in 64 bits. Using gcc linux, it's 4 bytes in 32 bits, and 8 bytes in 64 bits.
You should use int64_t instead (or a custom type that is defined to a 8 bytes type in 64 bits and to 4 bytes in 32 bits, or even "long long" which is guaranteed in C99 to be at least 8 bytes).

PR #4094, opened 2013-11-05
fix cocos2d-x and cocosdenshion ios project setting, to avoid symbol not...

stubma:
fix cocos2d-x and cocosdenshion ios project setting, to avoid symbol not found error


CocosDenshion/proj.ios/CocosDenshion.xcodeproj/project.pbxproj
```diff
@@ -212,7 +212,6 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
-				ARCHS = "$(ARCHS_STANDARD_32_BIT)";
```
dumganhar, line 215: Why could deleting this line resolve the issue?
When will the issue appear?

stubma:
test environment: xcode 5.0.1, target device is iphone 5
create a cocos2d-x project, drag cocos2d-x and cocosdenshion project into it, try to run it on device

get link error:


cocos2d-x and cocosdenshion ios project has a non-default architecture setting, I think it is not necessary. delete them will return it to default value, works fine.

dumganhar:
Hi @stubma 
I tested project created by `project_creator` of master branch. And following your step, I could not reproduce this issue. Any comments?

stubma:
don't forget call some cocosdenshion api in test code

dumganhar:
Yep, I invoking `SimpleAudioEngine::end();` method in AppDelegate.cpp.

stubma:
project_creator includes cocosdenshion source directly. I don't use it that way. Try include cocosdenshion project.

dumganhar:
Yep, I tried it same as you. But I could not get the linking error results.

stubma:
because the project generated by project-creator has a different architecture setting. 
image

If you restore test project architecture setting to default value, error should be seen

PR #4096, opened 2013-11-05
Android onResume bug

seagullua:
In Android onResume may be called when the lock screen of the phone is
still locked. Thus for example music may play while the screen is
locked.

Info: http://stackoverflow.com/questions/11731285/onresume-being-called-over-and-over-while-phone-screen-is-locked

I changed the code to make notification `AppDelegate::applicationWillEnterForeground` just when the application
became visible for the user.


seagullua:
This do not work on all devices. Closed

PR #4099, opened 2013-11-06
Adds more 64-bit fixes

ricardoquesada:
Replaces more `int` with `long` where it makes sense.

Also , it repalces some `unsigned long` with `long` for "lenght"
values as described in our c++ guideline


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/4108 .
Thanks.

PR #4100, opened 2013-11-06
Iss2983 multiple gl view

dabingnn:
add Node::getDirector(). replace some call Director::getInstance() by getDirector()


dabingnn:
This patch has been discarded.

PR #4120, opened 2013-11-07
continue replacing `const char*` with `const std::string&`

ricardoquesada:
This patch includes:
- Continue replacing `const char*` with `const std::string&`  where it makes sense.
- It also removes some Hungarian notation from the code.
- It fixes some indentation issues in part of the code
- It replaces `int` or `unsigned int` with `long` where it makes sense.


minggo:
Fixed building errors in #4122.

PR #4123, opened 2013-11-07
fix Accelerometer Test with missing: Device::setAccelerometerEnabled(tru...

xhcnb:
Just add Device::setAccelerometerEnabled(true) when use Acc sensor.
...e);


dumganhar:
Thanks.Could you fix the indention?

xhcnb:
sorry for my editor setting, I have fixed the indention

PR #3859, opened 2013-09-28
Adds Point::ANCHOR_XXX constants

ricardoquesada:
Make the code more explicit, less error prone, and less objects to create
Updates the SpriteTest with these constants as well

PR #4127, opened 2013-11-07
Point: Adds ANCHOR_XXX constants like ANCHOR_MIDDLE, ANCHOR_TOP_RIGHT, e...

WenhaiLin:
...tc.

### DIFF
--- a/cocos/base/CCGeometry.cpp
+++ b/cocos/base/CCGeometry.cpp
@@ -284,6 +284,15 @@ Point Point::getIntersectPoint(const Point& A, const Point& B, const Point& C, c
 }
 
 const Point Point::ZERO = Point(0, 0);
+const Point Point::ANCHOR_MIDDLE = Point(0.5, 0.5);
+const Point Point::ANCHOR_BOTTOM_LEFT = Point(0, 0);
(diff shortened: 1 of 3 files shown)


minggo:
New pull request for #3859.

ricardoquesada:
thanks.

winipcfg:
Isn't it better to use 1.0f, 0.5f, etc instead?

minggo:
@winipcfg 
What did you mean?

ricardoquesada:
@winipcfg 
you can still use 1.0 or 0.5 if you want.

minggo:
@winipcfg 
Yep, you are right.

@Dhilan007 
Could you please fix it?

cocos/base/CCGeometry.cpp
```diff
@@ -284,6 +284,15 @@ Point Point::getIntersectPoint(const Point& A, const Point& B, const Point& C, c
 }
 
 const Point Point::ZERO = Point(0, 0);
+const Point Point::ANCHOR_MIDDLE = Point(0.5, 0.5);
```
minggo, line 287: 0.5 -> 0.5f.

WenhaiLin:
@minggo 
yes,i should fix it

PR #4129, opened 2013-11-07
Colors are now compare- and explicit convertible

michaelcontento:


cocos/2d/ccTypes.h
```diff
@@ -31,29 +31,33 @@ of this software and associated documentation files (the "Software"), to deal
...
-    {}
-    
-    bool equals(const Color3B &other)
+    Color3B();
+    Color3B(GLubyte _r, GLubyte _g, GLubyte _b);
+    explicit Color3B(const Color4B& color);
+    explicit Color3B(const Color4F& color);
+
+    bool operator==(const Color3B& right) const;
+    bool operator==(const Color4B& right) const;
```
minggo, line 50: Is it reasonable to do this color3b == color4b?

michaelcontento, line 50: I admit that it might not the most used feature of all - but it's possible. First I started with all the explicit constructors and comparison only for the same type. And at one point I asked myself, why should the comparison be limited? Yes, the chances are good that the alpha channel is not fully set. But does it hurt to implement the two comparison methods? No. Does it help the user and provide a consistent API? Yes.

Actually I'd love to add `cocos2d::extension::HSV` into the compare- and convertible mix ... ;)

PR #4149, opened 2013-11-09
Return non-zero if ndk build fails.

dumganhar:


minggo:
If we build with `all` or `lua`, we don't want to it exit at once. May be should continue.

minggo:
Because travis also use `all` to build all android samples. So it is reasonable to exit when meeting an error.

dumganhar:
> If we build with all or lua, we don't want to it exit at once. May be should continue.

If cpp codes contain errors, why not to exit directly?
`Return directly` is needed when `cplusplus` source codes have errors.

BTW, the old logic of  `build_native.sh` did it like this.

dumganhar:
In this way, the result of travic-ci will be correct.
You could refer to the latest ci-build of android port.
Although, the cpp codes triggered compilation errors, but it didn't quit. What was worse, it return 0 which means `no errors`.

minggo:
Yep, may be we should add an option to control if should exit at once in future.

dumganhar:
Agree.

PR #4151, opened 2013-11-09
[Win32] fix compile error.

pyrasis:
- fix argument type. FileUtilsWin32::getFileData
- fix C2552 error in VS2012.
- Update files list. libGUI.


dumganhar:
Thanks.

PR #4156, opened 2013-11-10
Update CCImageCommon_cpp.h

bopohaa:
webp don't work for me


dumganhar:
Hi, @bopohaa 
Thanks for this patch, could you tell me which platform are you using to reproduce the issue?

bopohaa:
I used win32 platform and compile on Visual Studio 2013.

dumganhar:
So,  did the `webp test` of `Texture2DTest` crash?

dumganhar:
Thanks.

PR #4159, opened 2013-11-11
Fix typos and other trivial cleanup

mlepage:
Some minor cleanup in CCSprite.h. My text editor cleans up empty lines too, so some changes are just removing whitespace, hope that's OK. If not I'll have to redo it without the whitespace-only changes.


minggo:
Thanks.

PR #3025, opened 2013-06-28
closed #2339 refactor checkParams() in create_project.py, use OptionParser to parse the params

boyu0:

PR #4166, opened 2013-11-11
issue #3025: move deprecated texture cache implementation to cpp file

dabingnn:


cocos/2d/CCTextureCache.cpp
```diff
@@ -79,6 +79,15 @@ void TextureCache::destroyInstance()
...
 }
 
+TextureCache * TextureCache::sharedTextureCache() 
+{
+    return Director::getInstance()->getTextureCache();
+}
+
+void TextureCache::purgeSharedTextureCache() 
+{ 
+}
```
dumganhar, line 89: Empty implementation?

dabingnn, line 89: In previous implementation, Only director called TextureCache::DestroyInstance()
since TextureCache is not a singleton at all, it will be owned by Director, so its life will be managed by Director. Director directly do the release operation.

PR #4174, opened 2013-11-11
Master nutty optimizeguireader

nutty898:


minggo:
@sbc100
Could you please help to checkout why it is broken on native client?

Thanks.

minggo:
I fixed in #4180.

PR #4186, opened 2013-11-12
replaces `delete[]` with `free()` in C-based API

ricardoquesada:
API that returns a newly allocated buffer as an output argument (not return value) are error-prone.
- Users forget to release the newly allocated buffer
- Or the call `delete` instead of `delete[]`

But some of those API need to call `realloc` on the buffer.
But `realloc` is only valid if the buffer was previously allocated with `malloc`.
If a buffer needs to be re-allocated using a C++ API, then `std::vector<char*>` should be used instead...

So, this patch does:
- Migrates the API from  `new []` / `delete[]` to  `malloc()` / `free()`
- Fixes all the memory issues: incorrect deallocs and memory leaks
- Updates the documentation
- And fixes misc issues with the API: removes `cc` from the ZipUtils class.

### DIFF
--- a/cocos/2d/CCFontFreeType.cpp
+++ b/cocos/2d/CCFontFreeType.cpp
@@ -136,7 +136,7 @@ FontFreeType::~FontFreeType()
     }
     if (_ttfData)
     {
-        delete _ttfData;
+        free(_ttfData);
         _ttfData = nullptr;
     }
 }
--- a/cocos/2d/CCParticleSystem.cpp
+++ b/cocos/2d/CCParticleSystem.cpp
@@ -396,7 +396,7 @@ bool ParticleSystem::initWithDictionary(Dictionary *dictionary, const std::strin
                         CCASSERT( buffer != NULL, "CCParticleSystem: error decoding textureImageData");
                         CC_BREAK_IF(!buffer);
                         
-                        int deflatedLen = ZipUtils::ccInflateMemory(buffer, decodeLen, &deflated);
+                        int deflatedLen = ZipUtils::inflateMemory(buffer, decodeLen, &deflated);
                         CCASSERT( deflated != NULL, "CCParticleSystem: error ungzipping textureImageData");
                         CC_BREAK_IF(!deflated);
                         
@@ -420,8 +420,8 @@ bool ParticleSystem::initWithDictionary(Dictionary *dictionary, const std::strin
             bRet = true;
         }
     } while (0);
-    CC_SAFE_DELETE_ARRAY(buffer);
-    CC_SAFE_DELETE_ARRAY(deflated);
+    free(buffer);
+    free(deflated);
     return bRet;
 }
 
--- a/cocos/2d/CCTMXXMLParser.cpp
+++ b/cocos/2d/CCTMXXMLParser.cpp
@@ -66,7 +66,7 @@ TMXLayerInfo::~TMXLayerInfo()
     CC_SAFE_RELEASE(_properties);
     if( _ownTiles && _tiles )
     {
-        delete [] _tiles;
+        free(_tiles);
         _tiles = NULL;
     }
 }
@@ -757,12 +757,12 @@ void TMXMapInfo::endElement(void *ctx, const char *name)
                 // int sizeHint = s.width * s.height * sizeof(uint32_t);
                 int sizeHint = (int)(s.width * s.height * sizeof(unsigned int));
                 
-                int inflatedLen = ZipUtils::ccInflateMemoryWithHint(buffer, len, &deflated, sizeHint);
+                int inflatedLen = ZipUtils::inflateMemoryWithHint(buffer, len, &deflated, sizeHint);
                 CCASSERT(inflatedLen == sizeHint, "");
                 
                 inflatedLen = (size_t)&inflatedLen; // XXX: to avoid warnings in compiler
                 
-                delete [] buffer;
+                free(buffer);
                 buffer = NULL;
                 
                 if( ! deflated )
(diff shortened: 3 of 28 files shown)


cocos/2d/CCFontFreeType.cpp
```diff
@@ -136,7 +136,7 @@ bool FontFreeType::createFontObject(const std::string &fontName, int fontSize)
     }
     if (_ttfData)
     {
-        delete _ttfData;
+        free(_ttfData);
```
dumganhar, line 139: FileUtils::getInstance()->getFileData return C++ Binary Array which is allocated by `new []`, So we should use `delete[]` to release it.

dumganhar, line 139: My fault, I saw your changes, `new[]` has been changed to `malloc`.But it will break the compatibility.

ricardoquesada, line 139: yes... correct. It will break compatibility, but it is because we are calling `realloc` on some of them, and that is forbidden.

The options are:
- You use `malloc`, `realloc`, `free`
- You use `new []`, `delete[]`  (without `realloc`!!)
- You use `std::vector<>` if you need realloc

dumganhar, line 139: Yep, I saw this changes only affect `FileUtils::getInstance()->getFileData()`, right? Another changes are all used internally.
So to keep compatibility, could we do this:
- Adding new method for getting string and binary data from file, maybe `FileUtils::getStringFromFile`, `FileUtils::getDataFromFile`, and they return pointer which is allocated by `malloc`.
- Deprecating `FileUtis::getFileData`, and it still should return pointer which is allocated by `new[]`, but do a copy from the `getStringFromFile` or `getDataFromFile`.

dumganhar, line 139: By the way, the method `FileUtils::getFileData` is not a good design, you could see its arguments, the `mode` is always `rb` or `rt`, it is `readonly` of course. And there is no way to return a string. it always return binary data.

ricardoquesada, line 139: Yes, probably adding a new API that returns a malloced buffer is better.
Take into account that it also affects the `ZipUtils` class. So, there are about 4 or 5 API that need to be deprecated / replaced.

dumganhar, line 139: Yep, we could deprecate `ZipUtils::ccXXX`, and add `ZipUtils::XXX` like you have done.
`ZipUtils::ccXXX` will do a copy operation from `malloc buffer` of `ZipUtils::XXX`.

ricardoquesada, line 139: Sounds good.

Perhaps we should improve the name in`getFileData`... it is somewhat error prone.
Because "get" implies something like "return the already obtained value".
What about something like: `readAndAllocateFileData()` ? Is that too long ?

minggo, line 139: Yep, `readxxx` is more reasonable.

minggo, line 139: Why not just `readFileData`?

dumganhar, line 139: `readFileData` is a little strange. `readFile` will make sense.
Microsoft uses the file API like `ReadFile`, `WriteFile`.
Since `readFile` return data of course. So do we need the long one `readAndAllocateFileData`?

dumganhar, line 139: Add @Wu-Hao to discuss

ricardoquesada, line 139: My main issue is not `get` vs. `read`... but that the API should imply in a way that the returned buffer needs to be released by the user.
Otherwise the memory will be leaked..
that's why one proposal is `readAndAllocate...()`... `getAndAllocate...()` should work as well.
Also `allocBufferWithDataFrom...()` should work too.

dumganhar, line 139: > My main issue is not get vs. read... but that the API should imply in a way that the returned buffer needs to be released by the user.

About implying user to release allocated buffer, how about do it like which in `Tizen` by adding `N` at the end?

`unsigned char* FileUtils::allocBufferWithDataFromFileN()`, when user see api that has a `N` suffix, he/she should take care of the allocated buffer and has responsibility to release it.

minggo, line 139: Add `N` postfix doesn't make sense. It is not clear and we will have a new regular. Have a clear name makes more sense. 

I like `readAndAllocatexxx()`.

ricardoquesada, line 139: I agree with @minggo.
The Tizen convention is not easy to understand.

dumganhar, line 139: OK. But the API imply user to use `delete[]` or `free` to release the allocated buffer? 
It seems that it could only be done by good comments.

minggo, line 139: Yes.
Good comments is needed too.

ricardoquesada, line 139: Yes, we need good documentation.
The API should not imply whether to use `free()` or `delete[]`, but it should imply that the  user is responsible for releasing the returned buffer.
And the documentation should explicitly say: "The user should release the returned buffer by calling `free()` (or `delete[]`)"

dumganhar, line 139: Agree. :)

ricardoquesada, line 139: Good. What are the next steps ?
Who is going to implement those changes and when ?

The 2 issues to solve are:
- improve the API
- fix the current memory leaks.

My patch fixes the memory leaks. But not the API.

One way to do is:
- apply my patch (fixes memory leaks)
- then change the API (which is going to be easier).

minggo:
@dabingnn will continue the work.

ricardoquesada:
@minggo ok. thanks.

minggo:
It is done in #4262.

PR #4188, opened 2013-11-12
bind studio and gui

WenhaiLin:


scripting/javascript/bindings/jsb_cocos2dx_studio_manual.cpp
```diff
@@ -0,0 +1,542 @@
...
+    JSObject *obj = JS_THIS_OBJECT(cx, vp);
+    js_proxy_t *proxy = jsb_get_js_proxy(obj);
+    cocos2d::extension::UIWidget* cobj = (cocos2d::extension::UIWidget *)(proxy ? proxy->ptr : NULL);
+    JSB_PRECONDITION2( cobj, cx, JS_FALSE, "Invalid Native Object");
+
+    if (argc == 2) {
+        jsval *argv = JS_ARGV(cx, vp);
+
+        JSStudioEventListenerWrapper *tmpObj = new JSStudioEventListenerWrapper();
+        cobj->setUserObject(tmpObj);
```
dumganhar, line 85: It will cause memory leak since `setUserObject` retains the wrapper.
So you should `release` tmpObj after adding it to `Node::setUserObject`.

dumganhar, line 85: The same as other `addXXXListener`.

WenhaiLin, line 85: thanks so much

PR #4193, opened 2013-11-12
Adds markdown comments in the doxstrings

ricardoquesada:
Since Doxygen 1.8.0 (we are already using it), it is possible to
add markdown comments in the doxygen strings.

This will make our comments easier to read.

This patch  only adds markdown to CCNode.h as a test.

If you like, I'll keep adding markdown comments as I see them.
Please, encourage the rest of the team to do the same.

### DIFF
--- a/cocos/2d/CCNode.h
+++ b/cocos/2d/CCNode.h
@@ -82,22 +82,22 @@ bool nodeComparisonLess(Object* p1, Object* p2);
 
 class EventListener;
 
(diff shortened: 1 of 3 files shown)


minggo:
Seems great. Any link of the information about how to do it?

ricardoquesada:
@minggo 
Here is the info: http://www.stack.nl/~dimitri/doxygen/manual/markdown.html

minggo:
Thanks.

PR #4205, opened 2013-11-12
fix direct delete touch object without use release crash bug;

bmanGH:
I want use touch->retain() to retain touch information in touch handle.
But it always crash if I do this.
Finally I found there is direct delete touch object in CCEGLViewProtocol:: handleTouchesOfEndOrCancel .
It's works for me with replace delete to release() .
Please check if this is correct.
Thanks.


dumganhar:
Could you pull the latest codes since there are conflicts.

bmanGH:
pull latest code resolved conflicts.

minggo:
@bmanGH 
Please remove cocos2dx/platform/CCEGLViewProtocol.cpp.

PR #4207, opened 2013-11-12
Configuration of VAO in runtime

bmanGH:
use Configuration::getInstance()->supportsShareableVAO() in place of CC_TEXTURE_ATLAS_USE_VAO.
and
fix GL_EXT_discard_framebuffer always disable in EAGLView.mm.

### DIFF
--- a/cocos/2d/CCDrawNode.cpp
+++ b/cocos/2d/CCDrawNode.cpp
@@ -115,7 +115,8 @@ DrawNode::~DrawNode()
     glDeleteBuffers(1, &_vbo);
     _vbo = 0;
     
-    if (Configuration::getInstance()->supportsShareableVAO()) {
+    if (Configuration::getInstance()->supportsShareableVAO())
+    {
         glDeleteVertexArrays(1, &_vao);
         GL::bindVAO(0);
         _vao = 0;
@@ -160,7 +161,8 @@ bool DrawNode::init()
     
     ensureCapacity(512);
     
-    if (Configuration::getInstance()->supportsShareableVAO()) {
+    if (Configuration::getInstance()->supportsShareableVAO())
+    {
         glGenVertexArrays(1, &_vao);
         GL::bindVAO(_vao);
     }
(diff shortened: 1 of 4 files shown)


bmanGH:
pull latest code resolved conflicts.

cocos/2d/CCDrawNode.cpp
```diff
@@ -114,11 +115,11 @@ static inline Tex2F __t(const Vertex2F &v)
     glDeleteBuffers(1, &_vbo);
     _vbo = 0;
     
-#if CC_TEXTURE_ATLAS_USE_VAO      
-    glDeleteVertexArrays(1, &_vao);
-    GL::bindVAO(0);
-    _vao = 0;
-#endif
+    if (Configuration::getInstance()->supportsShareableVAO()) {
```
minggo: I think we should put `{` in a new line to follow cocos2d-x coding style: https://github.com/cocos2d/cocos2d-x/blob/develop/docs/CODING_STYLE.md.

Thanks.

cocos/2d/platform/ios/EAGLView.mm
```diff
@@ -237,8 +237,11 @@ -(BOOL) setupSurfaceWithSharegroup:(EAGLSharegroup*)sharegroup
     
     context_ = [renderer_ context];
     
-
-    //discardFramebufferSupported_ = [[Configuration sharedConfiguration] supportsDiscardFramebuffer];
+    #if GL_EXT_discard_framebuffer == 1
```
minggo, line 240: Where defines `GL_EXT_discard_framebuffer`?
CCConfiguration already test "GL_EXT_discard_framebuffer". So it is better to use the result.

bmanGH, line 240: `GL_EXT_discard_framebuffer` defined in iOS OpenGLES framework "es2/glext.h"

Because EAGLView has initialized before Configuration initialized (gatherGPUInfo called in Director::setOpenGLView function). So I direct use OpenGL extension macro define here.

minggo, line 240: Got it. 
Thanks.

minggo:
Thanks.

PR #4209, opened 2013-11-12
Fix custom properties not being reset on loadNode in CCBReader

FrancoisDupayrat:
When loading a CCB file, the custom properties were not reset between CCNode, causing some nodes to have custom properties they shouldn't have. This is due to m_pCustomProperties being an instance variable of CCNodeLoader.

The patch fix this behavior by removing all objects from m_pCustomProperties before loading a CCNode.

Tested on iOS only, there is no platform-specific code.

### DIFF
--- a/cocos/editor-support/cocosbuilder/CCNodeLoader.cpp
+++ b/cocos/editor-support/cocosbuilder/CCNodeLoader.cpp
@@ -27,7 +27,7 @@ Dictionary* NodeLoader::getCustomProperties()
 }
 
 Node * NodeLoader::loadNode(Node * pParent, CCBReader * ccbReader) {
-    m_pCustomProperties->removeAllObjects();
+    _customProperties->removeAllObjects();
     Node * ccNode = this->createNode(pParent, ccbReader);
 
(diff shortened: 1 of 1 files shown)


cocos/editor-support/cocosbuilder/CCNodeLoader.cpp
```diff
@@ -27,6 +27,7 @@ Dictionary* NodeLoader::getCustomProperties()
 }
 
 Node * NodeLoader::loadNode(Node * pParent, CCBReader * ccbReader) {
+    m_pCustomProperties->removeAllObjects();
```
dumganhar: There isn't `m_pCustomProperties`, it was renamed to `_customProperties`.

FrancoisDupayrat: Indeed, fixed with next commit, thanks.

dumganhar:
It was merged into develop branch. Please refer to https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/editor-support/cocosbuilder/CCNodeLoader.cpp#L31

Thanks for your feedback and contribution. :)

PR #4216, opened 2013-11-13
fixed animation play error for master

liamcindy:


extensions/CocoStudio/Action/CCActionNode.cpp
```diff
@@ -479,4 +474,13 @@ void ActionNode::easingToFrame(float duration,float delayTime,ActionFrame* destF
 	cAction->update(delayTime);
 }
 
+
+bool ActionNode::isActionDoneOnce()
+{
+	if (m_action == nullptr)
```
minggo: Master branch doesn't enable c++ 11.

liamcindy: Thank you, I will fix it

PR #4234, opened 2013-11-14
Fix: UserDefault::createXMLFile with correct XML definition.

xhcnb:
We should use NULL to create a default XML declaration.
If just pass "1.0" to it , will cause XMLDocument::NewDeclaration have an wrong xml header.

### DIFF
--- a/cocos/2d/CCUserDefault.cpp
+++ b/cocos/2d/CCUserDefault.cpp
@@ -479,7 +479,7 @@ bool UserDefault::createXMLFile()
     {  
         return false;  
     }  
-	tinyxml2::XMLDeclaration *pDeclaration = pDoc->NewDeclaration(NULL);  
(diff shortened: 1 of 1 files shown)


cocos/2d/CCUserDefault.cpp
```diff
@@ -479,7 +479,7 @@ bool UserDefault::createXMLFile()
     {  
         return false;  
     }  
-	tinyxml2::XMLDeclaration *pDeclaration = pDoc->NewDeclaration("1.0");  
+	tinyxml2::XMLDeclaration *pDeclaration = pDoc->NewDeclaration(NULL);
```
minggo: NULL -> nullptr.
We use c++11 in develop branch.

Thanks.

xhcnb: sorry, i have fixed

PR #4240, opened 2013-11-14
fix: when use different count custom properties in CocosBuilder will get...

xhcnb:
... wrong custom properties in game

Same as #4168, on master branch


dumganhar:
Thanks.

PR #4244, opened 2013-11-14
fixed #3185

chengstory:
1. add "override" behind virtual function.
2. remove duplicate function in class ComAttribute.


cocos/editor-support/cocostudio/CCComAttribute.cpp
```diff
@@ -28,8 +28,7 @@ of this software and associated documentation files (the "Software"), to deal
 namespace cocostudio {
 
 ComAttribute::ComAttribute(void)
-: _attributes(NULL)
-, _jsonDict(NULL)
+: _jsonDict(NULL)
```
minggo: NULL -> nullptr.

PR #4281, opened 2013-11-16
Update UIWidget.h

zhulangren:
please check the code 
call himself ?????
void removeCCNode(bool cleanup){removeCCNode(cleanup);};


minggo:
Thanks for you feedback.
But please don't use pull request for feedback.
It is better to report it on [cocos2d-x official forum](http://cocos2d-x.org/forums).

Thanks.

PR #4283, opened 2013-11-16
apply cocos2dx coding style

minggo:
1. fix some warnings
2. remove hungarian
3. NULL -> nullptr
4. remove `void` if there is not any parameters of a function


minggo:
It is not finished.
There are so many files to be modified.
I will continue the work in other pull request.

PR #4286, opened 2013-11-16
add resume/pause handler to Scene class

matsuhiro:
I think that a developer can get the pause and resume event of a scene is better.
For example, if a developer create some threads, when the scene go to the back, a developer want to stop all threads. I think.

Would you give me some comments?


sergey-shambir:
Scenes can disappear with transition. Apple uses "willX" and "didX" methods to allow programmer select either begin or end of any timed transformation.

dumganhar:
@matsuhiro , I agree with @sergey-shambir . Why should we have these callbacks?
Application has the callback for `resume from background` and `enter background`.

gelldur:
why you can;t listen for notifications it is better solution

PR #4275, opened 2013-11-15
issue #2771: physical

boyu0:

PR #4287, opened 2013-11-17
Fix physicsbody

xhcnb:
PhyicsNody has missed the function: resetForces, and its setDynamic function is not implemented properly.

### DIFF
--- a/cocos/physics/CCPhysicsBody.cpp
+++ b/cocos/physics/CCPhysicsBody.cpp
@@ -393,6 +393,11 @@ PhysicsShape* PhysicsBody::addShape(PhysicsShape* shape)
     return shape;
 }
 
+void PhysicsBody::resetForces()
+{
+    cpBodyResetForces(_info->getBody());
+}
+
 void PhysicsBody::applyForce(const Vect& force)
 {
     applyForce(force, Point::ZERO);
--- a/cocos/physics/CCPhysicsBody.h
+++ b/cocos/physics/CCPhysicsBody.h
@@ -90,6 +90,10 @@ public:
     virtual PhysicsShape* addShape(PhysicsShape* shape);
     
     /**
+     * @brief Zero both the forces and torques currently applied to the body.
+     */
+    virtual void resetForces();
+    /**
      * @brief Applies a immediate force to body.
      */
     virtual void applyForce(const Vect& force);


cocos/physics/CCPhysicsBody.cpp
```diff
@@ -305,6 +306,7 @@ void PhysicsBody::setDynamic(bool dynamic)
             if (_world != nullptr)
             {
                 cpSpaceRemoveBody(_world->_info->getSpace(), _info->getBody());
+                cpSpaceConvertBodyToStatic(_world->_info->getSpace(), _info->getBody());
```
boyu0, line 309: I don't think this is necessary, because the body is removed from the space. maybe add this line can improve performance, I'm not sure.
The other reason is, if you set a body to static, the cpBodySetPos won't work. this is not I want, So I use cpBodyNew(INFINITY, INFINITY) instead of cpBodyNewStatic() to construct a static body.

cocos/physics/CCPhysicsBody.cpp
```diff
@@ -391,6 +393,11 @@ PhysicsShape* PhysicsBody::addShape(PhysicsShape* shape)
     return shape;
 }
 
+void PhysicsBody::resetForces()
+{
+    cpBodyResetForces(_info->getBody());
+}
```
boyu0, line 399: It has implemented in https://github.com/cocos2d/cocos2d-x/pull/4275, and merged, you can get it from the latest code, thanks.

cocos/physics/CCPhysicsBody.cpp
xhcnb, line 309: Thanks, I have a test scene and without cpSpaceConvertBodyToStatic if I changed one body to static via setDynamic(false), the body will not collide with others
I just keep the two lines in my private repo to make my phyics bodies happy
thanks to point out that " if you set a body to static, the cpBodySetPos won't work"

boyu0, line 309: Yes you are right, I wrote a demo to test it, the body have collider but act weird, I don't know why, I will figure out it and fix this bug in another way, thanks.

xhcnb:
@lmskater , sorry, this patch is outdated, I'll close it right now. Please ignore this patch.

PR #4320, opened 2013-11-18
fixed ui anmation crash when stop animation for master

liamcindy:


minggo:
Compiling errors: https://travis-ci.org/cocos2d/cocos2d-x/builds/14149405

PR #4335, opened 2013-11-19
recover controller on layer.

WenhaiLin:


dumganhar:
should not merge this pull request

cocos/2d/CCLayer.cpp
```diff
@@ -87,6 +95,400 @@ Layer *Layer::create()
...
+        return ScriptEngineManager::getInstance()->getScriptEngine()->sendEvent(&event);
+    }
+
+    return 0;
+}
+
+
+/// Touch and Accelerometer related
+
+void Layer::addTouchListener()
```
dumganhar: This method is not needed.

PR #4337, opened 2013-11-19
CCFontAtlas bug fixed

SiarheiFedartsou:


minggo:
@SiarheiFedartsou 
What's the behaviour of this issue?

SiarheiFedartsou:
@minggo without this fix CCFontAtlas doesn't use all allocated memory

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/CCFontAtlas.cpp#L176
@SiarheiFedartsou ,Thank you.

PR #4343, opened 2013-11-20
Adds virtual destructors on Interfaces

ricardoquesada:
Classes with at least one virtual function needs a virtual destructor.
This patch adds virtual destructors to  Interface classes

And also enables  'warn on  missing virtual destructors'


dumganhar:
Thanks.

PR #4351, opened 2013-11-21
Constructors are protected

ricardoquesada:
This patch basically moves all the Constructors, Destructors and `init` methods to the `protected:` space.

it also adds the `CC_DISALLOW_COPY_AND_ASSIGN` macro, and uses this macro in the `private:`

The patch also fixes other minor issues, like replacing `const char*` with `const std::string&`

### DIFF
--- a/cocos/scripting/javascript/bindings/cocos2d_specifics.cpp
+++ b/cocos/scripting/javascript/bindings/cocos2d_specifics.cpp
@@ -541,39 +541,38 @@ JSBool js_cocos2dx_CCMenuItemFont_create(JSContext *cx, uint32_t argc, jsval *vp
 
 JSBool js_cocos2dx_CCMenuItemToggle_create(JSContext *cx, uint32_t argc, jsval *vp)
 {
-  if (argc >= 1) {
-    jsval *argv = JS_ARGV(cx, vp);
-    cocos2d::MenuItemToggle* ret = cocos2d::MenuItemToggle::create();
+    if (argc >= 1) {
+        jsval *argv = JS_ARGV(cx, vp);
+        cocos2d::MenuItemToggle* ret = cocos2d::MenuItemToggle::create();
 
-    for (uint32_t i=0; i < argc; i++) {
-      js_proxy_t *proxy;
-      JSObject *tmpObj = JSVAL_TO_OBJECT(argv[i]);
-      proxy = jsb_get_js_proxy(tmpObj);
-      cocos2d::MenuItem* item = (cocos2d::MenuItem*)(proxy ? proxy->ptr : NULL);
-      TEST_NATIVE_OBJECT(cx, item)
-	if(i == 0) ret->initWithItem(item);
-	else ret->addSubItem(item);
-    }
+        for (uint32_t i=0; i < argc; i++) {
+            js_proxy_t *proxy;
+            JSObject *tmpObj = JSVAL_TO_OBJECT(argv[i]);
+            proxy = jsb_get_js_proxy(tmpObj);
+            cocos2d::MenuItem* item = (cocos2d::MenuItem*)(proxy ? proxy->ptr : NULL);
+            TEST_NATIVE_OBJECT(cx, item)
+            ret->addSubItem(item);
+        }
         
-    jsval jsret;
-    if (ret) {
-      js_proxy_t *proxy = jsb_get_native_proxy(ret);
-      if (proxy) {
-	jsret = OBJECT_TO_JSVAL(proxy->obj);
-      } else {
-	// create a new js obj of that class
-	proxy = js_get_or_create_proxy<cocos2d::MenuItemToggle>(cx, ret);
-	jsret = OBJECT_TO_JSVAL(proxy->obj);
-      }
-    } else {
-      jsret = JSVAL_NULL;
-    }
+        jsval jsret;
+        if (ret) {
+            js_proxy_t *proxy = jsb_get_native_proxy(ret);
+            if (proxy) {
+                jsret = OBJECT_TO_JSVAL(proxy->obj);
+            } else {
+                // create a new js obj of that class
+                proxy = js_get_or_create_proxy<cocos2d::MenuItemToggle>(cx, ret);
+                jsret = OBJECT_TO_JSVAL(proxy->obj);
+            }
+        } else {
+            jsret = JSVAL_NULL;
+        }
         
-    JS_SET_RVAL(cx, vp, jsret);
-    return JS_TRUE;
-  }
-  JS_ReportError(cx, "wrong number of arguments");
-  return JS_FALSE;
+        JS_SET_RVAL(cx, vp, jsret);
+        return JS_TRUE;
+    }
+    JS_ReportError(cx, "wrong number of arguments");
+    return JS_FALSE;
 }
 
 // "setCallback" in JS


cocos/2d/CCActionTween.h
```diff
@@ -69,9 +69,9 @@ class CC_DLL ActionTween : public ActionInterval
 {
 public:
     /** creates an initializes the action with the property name (key), and the from and to parameters. */
-    static ActionTween* create(float duration, const char* key, float from, float to);
+    static ActionTween* create(float duration, const std::string& key, float from, float to);
     /** initializes the action with the property name (key), and the from and to parameters. */
-    bool initWithDuration(float duration, const char* key, float from, float to);
+    bool initWithDuration(float duration, const std::string& key, float from, float to);
```
minggo, line 74: This `init` is not put into `protected` section.

cocos/2d/CCParticleExamples.cpp
```diff
@@ -86,7 +86,7 @@ ParticleFire* ParticleFire::createWithTotalParticles(unsigned int numberOfPartic
     return pRet;
 }
 
-bool ParticleFire::initWithTotalParticles(unsigned int numberOfParticles)
+bool ParticleFire::initWithTotalParticles(int numberOfParticles)
```
minggo, line 89: Why not use `long` instead?

cocos/2d/CCTransition.h
```diff
@@ -245,16 +235,14 @@ class CC_DLL TransitionMoveInR : public TransitionMoveInL
 {
 public:
     static TransitionMoveInR* create(float t, Scene* scene);
-    /**
-     * @js ctor
-     */
+    virtual void initScenes();
```
minggo: Why not put `initScenes()` into `protected` section?

extensions/GUI/CCControlExtension/CCControlButton.h
```diff
@@ -209,6 +195,21 @@ class ControlButton : public Control
...
+    /**
+     * @js NA
+     * @lua NA
+     */
+    virtual ~ControlButton();
+
+    virtual bool init();
+    virtual bool initWithLabelAndBackgroundSprite(Node* label, Scale9Sprite* backgroundSprite);
+    virtual bool initWithBackgroundSprite(Scale9Sprite* sprite);
+    virtual bool initWithTitleAndFontNameAndFontSize(std::string title, const char * fontName, float fontSize);
```
minggo: Why not use  `std::string` for `fontName`?

extensions/GUI/CCControlExtension/CCControlSwitch.cpp
```diff
@@ -34,32 +34,20 @@
 class ControlSwitchSprite : public Sprite, public ActionTweenDelegate
 {
 public:
+    /** creates an autorelease instance of ControlSwitchSprite */
+    static ControlSwitchSprite* createWithMaskSprite(
```
minggo, line 38: How about rename to `create`?

cocos/2d/CCActionTween.h
ricardoquesada, line 74: Correct.
This patch only covers Nodes, not Actions.

cocos/2d/CCParticleExamples.cpp
ricardoquesada, line 89: We can change everything to `long`, but the purpose of this patch is not to do that.

cocos/2d/CCTransition.h
ricardoquesada: probably it should have in `protected:` since day one.
Why wasn't it put there ? I do not know. 
The purpose of this patch was to move `createXXX()`... and its associated `initXXX()` to `protected:`.

This `initScenes()` was out of the scope of this patch.
But yes, please, feel free to move it to `protected:`

extensions/GUI/CCControlExtension/CCControlSwitch.cpp
ricardoquesada, line 38: sounds good.

extensions/GUI/CCControlExtension/CCControlButton.h
ricardoquesada: You are right. It should be `std::string`.

Just remember that migrating all `const char*` to `std::string&` was outside the scope of this patch.

ricardoquesada:
Any ETA when this patch is going to be merged ?
This is a big patch that changes many nodes. The sooner it is merged, the easier.

Of course, I can merge it myself, but I wanted to know if had any issue with it before merging it.
thanks.

cocos/2d/CCParticleExamples.cpp
ricardoquesada, line 89: Important thing:
My fix was related to this: https://github.com/cocos2d/cocos2d-x/blob/develop/docs/CODING_STYLE.md#on-unsigned-integers

But migrating all `int` to `long` seems reasonable. But before doing that, we should read this: https://github.com/cocos2d/cocos2d-x/blob/develop/docs/CODING_STYLE.md#integer-types 

Probably we need to do more research before migrating everything `int` to `long`.
For the moment we should migrate `int` to `long` when the compiler complains.

minggo:
I don't merge it because it breaks travis down. You can open https://travis-ci.org/ to have a look. I don't know why travis doesn't synchronize to github. We haven't modified the script about travis.

ricardoquesada:
sure. I'll fix those issues on LInux and Android.

ricardoquesada:
I tested everything on Android and Linux, and it works OK (besides the JS Bindings).

The problem is this one:
https://github.com/cocos2d/cocos2d-x/blob/develop/samples/Cpp/TestCpp/Classes/testBasic.cpp#L51

I needed to change `new` to `create()`... and Travis is using gcc 4.6.3 which still has some bugs in c++11.

I think the fix is to use a new compiler in Travis, like 4.8.

sergey-shambir:
1. There is `CC_DISABLE_COPY` macro which does the same thing as `CC_DISALLOW_COPY_AND_ASSIGN` and uses `delete` keyword when compiler supports it (checks only GCC and clang versions, always disabled for MSVC).
2. Maybe it's better to make RenderTexture ctor public. We had issues on PowerVR videocards on android and solved them by de-autoreleasing RenderTexture instances and reducing amount of such instances (i.e. we created one texture and splitted it to several parts instead of using several RenderTextures to create a few different textures).

ricardoquesada:
@sergey-shambir 
thanks. I renamed the `CC_DISABLE_COPY` to `CC_DISALLOW_COPY_AND_ASSIGN`.
and removed the other `CC_DISALLOW_COPY_AND_ASSIGN` macro.

And I made public the `RenderTexture` ctor.

What is the best way to reproduce the bug in `RenderTexture` ? I want to fix it so I can make its ctor protected again.

thanks.

sergey-shambir:
I'm not sure that it's bug, we faced with situation when PowerVR videocards returned OpenGL error "not enough memory" (code 0x502) when many autoreleased CCRenderTextures was created at game logic update while rendering one frame. Reproducing issue on Samsung Galaxy Tab 7.0 was very easy for us, on other devices (Samsung S2, Samsung S3, etc) it was much harder.

CCRenderTexture allocates framebuffer in init method and destroys it in destructor. So graphics memory stays allocated until instance isn't deleted, and it isn't deleted until next frame because CCRenderTexture is autoreleased.

What we did:
- we had human unit which consists of left leg, right leg, and body (with head), each body part was rendered by own CCRenderTexture instance
- CCRenderTexture instances was automatically added to autorelease pool
- some in-game conditions caused instant creation of 2-3 new human units on each frame
  on some moment graphics memory alloc failed on PowerVR and wasn't recovered even after all units was removed from scene (in the same game session).

Ways to solve issue:
- use non-autoreleased RenderTexture (we did this)
- add boolean property which will change CCRenderTexture behavior to allocate and free graphics memory on each begin/end

dumganhar:
Travis-ci build failed is because JSB and Lua projects were built with errors.

ricardoquesada:
@sergey-shambir 
thanks. So, it is not a "bug" in cocos2d... yell... it is a bug, but the bug is with the `autorelease` stuff that temporary it could allocate too much memory.

ricardoquesada:
Great!

So, are the JS bindings working again ?
How did you fix them ?
Thanks.

dumganhar:
@ricardoquesada 
Yes, I fixed the JSB issues.
As you know, only classes used for `extend` have `ctor` method.
So we don't need to add `ctor` binding for all classes.
I have added a section called `classes_need_extend` in `bindings-generator` for configuration of which classes need to be extended.

In `jsb_cocos2d.js`, only some classes have extend method.

``` javascript
cc.Node.extend = cc.Class.extend;
cc.Layer.extend = cc.Class.extend;
cc.LayerGradient.extend = cc.Class.extend;
cc.LayerColor.extend = cc.Class.extend;
cc.Sprite.extend = cc.Class.extend;
cc.MenuItemFont.extend = cc.Class.extend;
cc.Scene.extend = cc.Class.extend;
cc.DrawNode.extend = cc.Class.extend;
```

So I only generate `ctor` for these classes. I added `classes_need_extend` section in  `tools/tojs/cocos2dx.ini`.

```
classes_need_extend = Node Layer.* Sprite MenuItemFont Scene DrawNode
```

About the changes of Bindings-generator, please refer to https://github.com/cocos2d/bindings-generator/pull/50/files .

Another importance thing is in `js_XXX_ctor` binding glue codes, I did what you suggested( `new T` --> `T::create()`).
In this case, if a class needs to be extended, it should have a default `create` function since its `constructor` is **protected**.

ricardoquesada:
Well done James!!!!

dumganhar:
:)

PR #4352, opened 2013-11-21
touch outside of the cell on table view bottom, it will set _touchedCell to first cell

mosquitodu:
when touched on list bottom, not in any cell, in __indexFromOffset,
high will be -1, it always return 0, it should return invalid index. in
scrollViewDidScroll, when startIdx is Inavlid, we need set it to 0,
first of cell, not the last cell.

### DIFF
--- a/extensions/GUI/CCScrollView/CCTableView.cpp
+++ b/extensions/GUI/CCScrollView/CCTableView.cpp
@@ -392,7 +392,7 @@ int TableView::__indexFromOffset(Point offset)
         }
     }
 
-    if (low <= 0) {
+    if (low <= 0 && high >= 0) {
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


dumganhar:
@mosquitodu 
If we touch outside of scrollview area, tableview will not be get response.

``` c++
bool ScrollView::onTouchBegan(Touch* touch, Event* event)
{
    if (!this->isVisible())
    {
        return false;
    }

    Rect frame = getViewRect();

    //dispatcher does not know about clipping. reject touches outside visible bounds.
    if (_touches.size() > 2 ||
        _touchMoved          ||
        !frame.containsPoint(touch->getLocation()))  // Here will fail and return false.
    {
        return false;
    }
```

And that cause `TableView::onTouchBegan` return false too.
So I could not reproduce the issue you mentioned.

Could you describe it more?

PR #4353, opened 2013-11-21
windows version create_project.py when creating new project the the sulotions are allways HelloCpp.vcxproj

meiry:
when creating new project using the create_project.py script for example :
cocos2d-x-3.0alpha0\tools\project_creator>python create_project.py -p gui_test -k com.test.guitest -l cpp
the VS sulotions files allways named "HelloCpp" like this :
HelloCpp.vcxproj.user
HelloCpp.vcxproj.filters
HelloCpp.vcxproj
HelloCpp.sln


dumganhar:
I think you have sent the PR the wrong branch.
And I think this issue was fixed. So i closed this PR.

PR #4355, opened 2013-11-21
Fixes usage of wrong member variable when cloning CallFuncN objects

danielshir:
The clone() function for CallFuncN uses the wrong function member variable.
It uses the one used by CallFunc (`_function`) rather than CallFuncN's (`_functionN`).


dumganhar:
I think we have fixed the problem, you could refer to https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/CCActionInstant.cpp#L475 .

danielshir:
Yep, that fixed it.
Thanks!

PR #4356, opened 2013-11-21
Updated WinRT and WP8 to pass App Certification tests

stammen:
1. updated WinRT and WP8 to pass App Certification tests
2. updated project templates
3. updated to latest version of ANGLE
4. fixed type that was preventing loading of precompiled shaders in Windows 8 WinRT builds
5. Added Trim() support for app Suspend() in Windows 8.1 WinRT
6. Fixed CCEGLView::SetScissorInPoints() on WP8 to fixed libextensions CCScollView bug


saudahmed:
great this stopped from submitting my app to the store

PR #4362, opened 2013-11-22
Fix: on android, Device::setAccelerometerInterval will causes dead loop

xhcnb:
it's a low level mistake  
: )

### DIFF
--- a/cocos/2d/platform/android/CCDevice.cpp
+++ b/cocos/2d/platform/android/CCDevice.cpp
@@ -18,17 +18,17 @@ void Device::setAccelerometerEnabled(bool isEnabled)
 {
     if (isEnabled)
     {
-        enableAccelerometer_native();
+        enableAccelerometerJni();
     }
     else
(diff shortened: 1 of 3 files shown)


cocos/2d/platform/android/CCDevice.cpp
```diff
@@ -28,7 +28,7 @@ void Device::setAccelerometerEnabled(bool isEnabled)
 
 void Device::setAccelerometerInterval(float interval)
 {
-	setAccelerometerInterval(interval);
+	::setAccelerometerInterval(interval);
```
minggo: Thanks.
But it think it is better to change the jni function name.

xhcnb:
@minggo ,U are right, Just renamed the native functions.

xhcnb:
Renamed.
Thanks for your guys hard working, hope for 3.0 release.

minggo:
Thanks.

PR #4363, opened 2013-11-23
Fixed warnings in VS2012: C4018 and C4244.

syrokomskyi:


cocos/gui/UIListView.cpp
```diff
@@ -301,7 +301,8 @@ void UIListView::removeLastItem()
 
 UIWidget* UIListView::getItem(unsigned int index)
 {
-    if ((int)index < 0 || index >= _items->count())
+    // @todo Signed/unsigned mishmash. Index is must be unsigned int always. Yes?
+    if ( ((int)index < 0) || ((int)index >= _items->count()) )
```
minggo: May be we should change the type of parameter.
So could you please revert the modification of this file?

Thanks.

PR #4364, opened 2013-11-23
Visual Studio 2013 support

ivfit:


minggo:
Master branch will only used for critical bug fixed before 3.0 stable version is released. So please don't add new feature in this branch.

Thanks.

ivfit:
Right, Visual Studio 2013 support is a critical bug fix. Without that VS 2013 won't compile Cocos2d-x at all:
https://github.com/finalhit/cocos2d-x/commit/fda3608c89d059430183461aa8ce5cf27903b697

PR #4367, opened 2013-11-24
proj.mac FontTest fix (was not rendering TTFs)

lajos:
The proj.mac version of TestCpp FontTest was not rendering fonts because the system couldn't locate the TTFs.
- added ATSApplicationFontsPath=. to info.plist
- added UIAppFonts array to info.plist
- added MAC target condition to FontTest.cpp (defining same values as IOS)


dumganhar:
Could you also send a PR to develop branch? Thanks.

lajos:
@dumganhar : Here's the develop branch pull request with the same fix : https://github.com/cocos2d/cocos2d-x/pull/4373

dumganhar:
Thanks.

PR #4373, opened 2013-11-25
FontTest was not rendering TTFs

lajos:
The proj.mac version of TestCpp FontTest was not rendering fonts because the system couldn't locate the TTFs.
- added ATSApplicationFontsPath=. to info.plist
- added UIAppFonts array to info.plist
- added MAC target condition to FontTest.cpp (defining same values as IOS)


dumganhar:
Thanks.

PR #4377, opened 2013-11-26
Master

sscholle:
Changes:
Added support for Websockets over SSL (Previous implementation only supported unencrypted transmission)

Test Case:
Connect to wss://echo.websocket.org
to Test Secure Websockets

Platforms: 
Tested on iOS and Android


dumganhar:
Could you send the PR to develop branch? Master branch is only for critical bug fix.

PR #4389, opened 2013-11-27
override addChild in Sprite to fix error in binding generator

minggo:
Binding generator will bind functions that be declared in the class. Sub-class should have to override all functions to generate all needed binding functions.

### DIFF
--- a/cocos/2d/CCSprite.h
+++ b/cocos/2d/CCSprite.h
@@ -409,7 +409,7 @@ public:
     virtual void removeChild(Node* child, bool cleanup) override;
     virtual void removeAllChildrenWithCleanup(bool cleanup) override;
     virtual void reorderChild(Node *child, int zOrder) override;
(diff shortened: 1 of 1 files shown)


cocos/2d/CCSprite.h
```diff
@@ -412,7 +409,10 @@ class CC_DLL Sprite : public NodeRGBA, public TextureProtocol
     virtual void removeChild(Node* child, bool cleanup) override;
     virtual void removeAllChildrenWithCleanup(bool cleanup) override;
     virtual void reorderChild(Node *child, int zOrder) override;
-    using Node::addChild;
+    // Shoul also override addChild(Node*) and addChild(Node*, int), or binding generator will only
```
edwardzhou: typo: Shoul  ->  Should

minggo: Thanks.

PR #4394, opened 2013-11-27
update armature

2youyou2:
change ColliderFilter param type


minggo:
Please fix compiling errors: https://travis-ci.org/cocos2d/cocos2d-x/pull_requests

2youyou2:
The compiler says that "chipmunk.h" cann't find, but i can compile successfully in local .

PR #4399, opened 2013-11-27
Added QtCreator project for Linux and Windows (MinGW)

sergey-shambir:
Tested on
1. Ubuntu 13.10, 64-bit, clang compiler
2. Ubuntu 13.10, 64-bit, GCC compiler
3. Windows XP, 32-bit, MinGW compiler
4. Windows 7, 64-bit, MinGW 32-bit compiler

There're no Qt5 port yet, but I'm going to add it later. Current progress is in [1], TODOs are in that commit message. Probably I should make refactoring on SimpleAudioEngine before publishing Qt5 port, because in Qt5 audio can be decoded only asynchroniously so it's better to move whole SimpleAudioEngine into separate thread for all platforms.

P.S. In game published to GooglePlay we have rare fatal Java exceptions due to non thread-safe SimpleAudioEngine in cocos2dx 2.1.4, so probably refactoring SimpleAudioEngine will be useful for Android too.

[1] https://github.com/sergey-shambir/cocos2d-x/commit/be7beb8d765dc03a1f5aa781e7b6a5f32f9a7e32

### DIFF
--- a/cocos/audio/win32/MciPlayer.cpp
+++ b/cocos/audio/win32/MciPlayer.cpp
@@ -1,5 +1,4 @@
 #include "MciPlayer.h"
-#include <tchar.h>
 
 #define WIN_CLASS_NAME        "CocosDenshionCallbackWnd"
 #define BREAK_IF(cond)      if (cond) break;
@@ -20,7 +19,7 @@ MciPlayer::MciPlayer()
 {
     if (! s_hInstance)
     {
-        s_hInstance = GetModuleHandle( NULL );            // Grab An Instance For Our Window
+        s_hInstance = GetModuleHandleA( NULL );            // Grab An Instance For Our Window
 
         WNDCLASSA  wc;        // Windows Class Structure
 
@@ -57,7 +56,7 @@ MciPlayer::MciPlayer()
         NULL );
     if (_wnd)
     {
-        SetWindowLong(_wnd, GWL_USERDATA, (LONG)this);
+        SetWindowLongA(_wnd, GWL_USERDATA, (LONG)this);
     }
 }
 
@@ -86,7 +85,7 @@ void MciPlayer::Open(const char* pFileName, UINT uId)
         mciOpen.lpstrDeviceType = (LPCSTR)MCI_ALL_DEVICE_ID;
         mciOpen.lpstrElementName = pFileName;
 
-        mciError = mciSendCommand(0,MCI_OPEN, MCI_OPEN_ELEMENT, (DWORD)&mciOpen);
+        mciError = mciSendCommandA(0,MCI_OPEN, MCI_OPEN_ELEMENT, (DWORD)&mciOpen);
         BREAK_IF(mciError);
 
         _dev = mciOpen.wDeviceID;
@@ -103,7 +102,7 @@ void MciPlayer::Play(UINT uTimes /* = 1 */)
     }
     MCI_PLAY_PARMS mciPlay = {0};
     mciPlay.dwCallback = (DWORD_PTR)_wnd;
-    s_mciError = mciSendCommand(_dev,MCI_PLAY, MCI_FROM|MCI_NOTIFY,(DWORD)&mciPlay);
+    s_mciError = mciSendCommandA(_dev,MCI_PLAY, MCI_FROM|MCI_NOTIFY,(DWORD)&mciPlay);
     if (! s_mciError)
     {
         _playing = true;
@@ -147,11 +146,11 @@ void MciPlayer::Rewind()
     {
         return;
     }
-    mciSendCommand(_dev, MCI_SEEK, MCI_SEEK_TO_START, 0);
+    mciSendCommandA(_dev, MCI_SEEK, MCI_SEEK_TO_START, 0);
 
     MCI_PLAY_PARMS mciPlay = {0};
     mciPlay.dwCallback = (DWORD)_wnd;
-    _playing = mciSendCommand(_dev, MCI_PLAY, MCI_NOTIFY,(DWORD)&mciPlay) ? false : true;
+    _playing = mciSendCommandA(_dev, MCI_PLAY, MCI_NOTIFY,(DWORD)&mciPlay) ? false : true;
 }
 
 bool MciPlayer::IsPlaying()
@@ -174,7 +173,7 @@ void MciPlayer::_SendGenericCommand(int nCommand)
     {
         return;
     }
-    mciSendCommand(_dev, nCommand, 0, 0);
+    mciSendCommandA(_dev, nCommand, 0, 0);
 }
 
 //////////////////////////////////////////////////////////////////////////
@@ -186,7 +185,7 @@ LRESULT WINAPI _SoundPlayProc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam)
     MciPlayer * pPlayer = NULL;
     if (MM_MCINOTIFY == Msg 
         && MCI_NOTIFY_SUCCESSFUL == wParam
-        &&(pPlayer = (MciPlayer *)GetWindowLong(hWnd, GWL_USERDATA)))
+        &&(pPlayer = (MciPlayer *)GetWindowLongA(hWnd, GWL_USERDATA)))
     {
         if (pPlayer->_times)
         {
@@ -195,11 +194,11 @@ LRESULT WINAPI _SoundPlayProc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam)
 
         if (pPlayer->_times)
         {
-            mciSendCommand(lParam, MCI_SEEK, MCI_SEEK_TO_START, 0);
+            mciSendCommandA(lParam, MCI_SEEK, MCI_SEEK_TO_START, 0);
 
             MCI_PLAY_PARMS mciPlay = {0};
             mciPlay.dwCallback = (DWORD)hWnd;
-            mciSendCommand(lParam, MCI_PLAY, MCI_NOTIFY,(DWORD)&mciPlay);
+            mciSendCommandA(lParam, MCI_PLAY, MCI_NOTIFY,(DWORD)&mciPlay);
         }
         else
         {
@@ -207,7 +206,7 @@ LRESULT WINAPI _SoundPlayProc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam)
         }
         return 0;
     }
-    return DefWindowProc(hWnd, Msg, wParam, lParam);
+    return DefWindowProcA(hWnd, Msg, wParam, lParam);
 }
 
 } // end of namespace CocosDenshion


cocos/audio/win32/MciPlayer.cpp
```diff
@@ -35,14 +36,14 @@
         wc.lpszMenuName   = NULL;                           // We Don't Want A Menu
         wc.lpszClassName  = WIN_CLASS_NAME;                 // Set The Class Name
 
-        if (! RegisterClass(&wc)
+        if (! RegisterClassA(&wc)
```
dumganhar, line 38: Why to use `Ascii` function?

cocos/audio/win32/MciPlayer.cpp
```diff
@@ -8,7 +9,7 @@
 static HINSTANCE s_hInstance;
 static MCIERROR  s_mciError;
 
-static LRESULT WINAPI _SoundPlayProc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam);
+LRESULT WINAPI _SoundPlayProc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam);
```
dumganhar, line 11: Why to delete `static` keyword?

cocos/network/HttpClient.cpp
```diff
@@ -28,6 +28,10 @@
 #include <queue>
 #include <errno.h>
 
+#ifdef __MINGW32__
+  #include "winsock2.h"
```
dumganhar: Is it better to put it to `win32/CCStdC.h` since `WinSocket2.h` was there?

samples/Cpp/SimpleGame/proj.win32/main.cpp
```diff
@@ -1,16 +1,21 @@
-#include "main.h"
+//#include "main.h"
```
dumganhar: Why to comment this line?

akof1314: Here is a mistake, I fork sergey-shambir: qtcreator, add Qt for MinGW support, there is no main.h file

cocos/audio/win32/MciPlayer.cpp
akof1314, line 11: because in MinGW, it is a error.: 'LRESULT CocosDenshion::_SoundPlayProc(HWND, UINT, WPARAM, LPARAM)' was declared 'extern' and later 'static' [-fpermissive]

delete static keyword does not affect vc project

cocos/audio/win32/MciPlayer.cpp
akof1314, line 38: In fact CocosDenshion.vcxproj CharacterSet is NotSet, means that it is the Ansi version, external parameter is const char_, then do not want to convert to const wchar_, must also be Ansi version, but for Qt Creator MinGW version, identified as a version of Unicode, leading to MciPlayer is not working, so I suggest MciPlayer.cpp to Ansi version

samples/Cpp/SimpleGame/proj.win32/main.cpp
sergey-shambir: Actually there is main file, it just was not added to project.

cocos/network/HttpClient.cpp
sergey-shambir: Ok, I will put it to win32/CCStdC and move <windows.h> below <winsock2.h> to fix MinGW warning.

build/qtcreator/user_configuration.prf
```diff
@@ -0,0 +1,12 @@
+
+# This file can be used to configure cocos2d in your local cocos2dx tweak.
+# Default settings are fine, but some indie developers or teams may prefer to
+# change them.
+
+# By default we use Chipmunk physics engine since Box2D is unmaintained.
```
minggo, line 6: Box2D released 2.3 in Nov.3. 
Why did you say it is unmaintained?
Any link for the information?

Thanks.

cocos/2d/platform/win32/CCFileUtilsWin32.cpp
```diff
@@ -55,12 +55,24 @@ static void _checkPath()
     if (0 == s_resourcePath.length())
     {
         WCHAR utf16Path[CC_MAX_PATH] = {0};
-        GetCurrentDirectoryW(sizeof(utf16Path)-1, utf16Path);
+
+#if CC_QTCREATOR_ON_WINDOWS
+        // Use "%executable_dir%/Resources" as root path.
+        HMODULE hModule = GetModuleHandleW(NULL);
```
minggo, line 61: `NULL` -> `nullptr`
`hModule` -> `module`

We use c++11, and we will remove all hungarian style.

minggo:
@sergey-shambir 
Thanks for the pull request. Awesome job.

I found that you uploaded many static libs for win32 platform. Why should use static libs? If it has to use static libs, then i think we should delete other dynamic libraries on win32.

akof1314:
@minggo because some external libraries is build by vc project, it can not be used on MinGW.

minggo:
Ok, got it, but i think it is confused to put them into `win32` folder. Because when a developer want to link some library, he doesn't know the difference. So i think it is better to create a new folder, such as `mingw`, to hold the libs.

minggo:
And one more question: is MinGW needed for QTCreator on windows?

sergey-shambir:
It's shipped as part of QtSdk for windows, while Visual Studio should be downloaded separately and sometimes glitches (e.g. Visual Studio 2010 doesn't work with QtCreator without SP1 update). So MinGW is more reliable when need to build app with QtCreator.

build/qtcreator/user_configuration.prf
sergey-shambir, line 6: Ok, I'll remove this note from comment.

minggo:
VS2012 is required to run cocos2d-x on windows. It makes things complicated to support MinGW, because we should ship and maintain so many libs.

akof1314:
@minggo MinGW libraries maybe does not need add to github. Some people are interested in MinGW, then can build their own. More people, then consider adding.

minggo:
@akof1314 
Did you agree with me? We don't have to support MinGW.

akof1314:
@minggo recommend add QtCreator project ,except the MinGW static libraries on win32. @sergey-shambir

sergey-shambir:
Ok, I'will remove MinGW static libraries and prepare project for MSVC 2012 as target compiler.

martell:
We would support Mingw64 because it can come with clang which would then work with emscripten.
Besides Visual Studio is very bloated for win32 development

minggo:
@sergey-shambir @akof1314 
Thanks.

@martell 
Windows and Qt are not releasing platform, we don't want to make engine too complex to be maintained. And it will slow down the developing speed, because we have to be care of so many platforms when adding a new feature.

martell:
@minggo

I understand where you are coming from.
Will you have a look at #4411 Hopefully that will make it easier.
I would consider windows mac and linux development platforms.
As you have said though Windows its not a Release platform.

We should have windows, mac, and linux running via cmake to make the development more streamlined.

I will then add in Android and IOS to cmake once you guys are happy with it 
(I already have it done it for my own projects)

minggo:
@martell 
It seems great. We have considered use `CMake` to generate all platform projects, but we wonder if `CMake` can meet our requirement. As you know, we use too many IDE configurations, such as share all targets in one `Xcode` project.

And some one suggests to use [premake](http://industriousone.com/premake). It seems `premake` can generate `Xcode` project on windows. I am not familiar with `premake` and `CMake`. I don't know which one is better. Could you give some suggestion of it?

Thanks.

martell:
@minggo 
CMake is a more proven system. It works on MSVC Xcode Eclipse Sublime etc
CMake would be best for Android as you can use it to do the cpp compiling rather than having a build_native.sh and build_native.cmd Have a look here https://code.google.com/p/android-cmake/ (It's used in opencv)
We could get rid of those annoying .mk files then and build and link cocos as a library rather than from source every time.

In my PR I have simplified generating Visual Studio 2010, 2012 and 2013 to a simple bat file
If you have a windows box you can test out my pull request. (recent jsoncpp changes in master branch may break it though)

My favourite part of CMake is that it generates Sublime Text Projects. :D

minggo:
@martell 
Thanks for your suggestion and sharing. It sounds good to use just `CMake` to generate all project files.

ricardoquesada:
@minggo 
I think it is possible to generate Xcode projects with CMake.

``` sh
$ cmake -G Xcode ...
```

I haven't tried it myself, but apparently it works OK. We need to do some tests, though.

If we can unify everything on CMake, it would be great.

martell:
@minggo
The only issue I have with CMake is that for visual studio you have to generate seperate projects for x86 arm x64

@ricardoquesada
cmake -v   will give you all generators you can use on that platform
Xcode is one of them for both IOS and MAC

I have cmake implemented for project generation in cocos2d-x 2.2 for most platforms

minggo:
@ricardoquesada 
Yep, we should do more research on `CMake`.

@martell 
Thanks for your work.

ricardoquesada:
@martell 
thanks. sounds good!

minggo:
@martell 
The reason i don't want to support MinGW is that it has to add all libs that cocos2d-x used. And we have to build library for MinGW when we add a new lib. Even worse is that we have to test for MinGW when releasing.

martell:
@minggo
I've done that already for you.
https://github.com/martell/MINGW-packages

The packages will be part of msys2 which uses pacman (like arch linux)
Any package that runs on linux will be added here.
Most of arch linux PKGBUILD files work without much change.

They don't even need to be in the cocos2d-x project.

just
pacman -S mingw-w64-i686-libglfw3 (to install)
equivalent to ubuntu's apt-get

It also has python tcl /tk etc so we don't need cygwin to run plugin-x or project-creator.py
Infact i use it for my git bash and it has shell script support so I dropped cygwin completely 
thanks to alexpux :D

We should treat Mingw64 and Msys2 like linux and not windows when used together.
That will fix the Windows specific problems we have

Check it out
http://sourceforge.net/projects/msys2

minggo:
@martell 
Thanks.

My question are
- Is MinGW needed and why?
- Will you maintain MinGW port?
- Will you test MinGW before releasing a new version of cocos2d-x. Now we don't have auto-testing script. We have to do testing manually. It cost us much time to do testing.

martell:
@minggo
1. MinGW is needed if we want to use gcc or clang on windows. (future emscripten target on windows)
   MSYS2 should be the system it runs on because cygwin is horrible
2. I can maintain it.
3. It can be built from Linux with travis.
   http://packages.ubuntu.com/lucid/mingw-w64

The Real work is in cmake.
Once CMake Works adding Mingw is a simple task.

minggo:
@martell 

> > MinGW is needed if we want to use gcc or clang on windows. (and for emscripten target on windows)
> > MSYS2 should be the system it runs on because cygwin is horrible

Why should we use gcc or clang on windows? Most developers are used to using VS for developing on windows. And now we don't need cygwin any more. As you see, we use python to build android projects. As for emscripten, i am not sure if it is valuable to support MinGW to support emscripten target on windows. But it is great to hear that you will maintain it.

> > It can be built from Linux with travis. MinGW is not just for windows as the host hence (Minimal gcc for Windows)

Did you mean we can test MinGW on linux?

martell:
@minggo
Why should we use gcc or clang on windows?
The second answer gives a better insight.

Did you mean we can test MinGW on linux?"

Yes we can build the Windows exe from Linux (in travis ci case ubuntu specifically)

We then can also build for Linux from Windows 
We can even build Mac from both Windows or Linux

EDIT: We can build for IOS from Windows and Linux Desktops also ;)

"This is true cross platform development" 

How do you do quotes on github by the way :P

minggo:
It sounds good to build all all in one platform. But we will add auto-testing, so we have to run samples on different environments.

You can use quotes like this 

``` c++
>> the contests to be quoted

Add comment here.
```

martell:
Initially for MingW we could just use Wine http://www.winehq.org/

minggo:
Yep, i have heard wine. But will it become too much complex?

And we should test all compilers that developer can use. For example, developers will use VS for developing, then we should test VS.

minggo:
So, i think we should remove the static libs now. 

@sergey-shambir 
What's your opinion?

martell:
@minggo 
glew png jpeg tiff webp glfw3 etc are all part of msys2 now in both x86 and x64 forms
http://sourceforge.net/projects/msys2/
I was talking to alexpux and he told me he is currently adding qt5 as a package to msys2

This should be very relevant here and the mingw static libs could be removed from this commit
I strongly advise leaving pacman manage the Mingw libraries instead of constantly updating them within the cocos2d-x repo

minggo:
> This should be very relevant here and the mingw static libs could be removed from this commit
> I strongly advise leaving pacman manage the libraries instead of constantly updating them within the cocos2d-x repo

It would be good. As we used dpkg on ubuntu.

martell:
Just had a look through the code of this pull request.

@sergey-shambir
We don't need to add qt creator ".pro" project files because it will add even more project types.
QT creator supports cmake so #4434 should be a better fit.
I'm using it with sublime and qtcreator right now :)

Also the newer r4 headers with mingw match visual studio headers much better now so half of the mingw specific changes aren't needed any more :)

dumganhar:
Could anyone update this PR since it could not be automatically merged?

sergey-shambir:
Ok, I'll update it in next few days.

dumganhar:
@sergey-shambir 
Thanks.

dumganhar:
@sergey-shambir , could you use `CMake` to organize qt projects? QTCreator supports to import CMakeLists.txt.
To have `.pro` or `.pri` which are QT project files will take us more time to maintain it.

darkdukey:
I personally would love to see this happen, since I've been play with Qt and MinGW these days.

martell:
Hey @dumganhar,

It should be already possible to load the project into QTCreator via Cmake.
You just have to have the cmake commandline with the options you want and drop it into QtCreator

Kind Regards
Martell

cocosQA:
Dear sergey-shambir:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6473

PR #4400, opened 2013-11-27
Fix ccb reader related bug in develop branch

btspoony:
- fix ccb reader’s animation manager using setUserObject conflict with layer’s setTouchNode
- JSB: the second argument of ccb loading can be NULL

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -93,6 +93,7 @@ Developers:
     martell
         use tinyxml2 to replace libxml2
         Added Mingw-crt Support without breaking VS SDK
+        CMake support for windows.
 
     mchinen
         fix emulator issue for OpenGL ES 2.0 on Android
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,9 @@ cocos2d-x-3.0beta0 ?? 2013
      [FIX]          XMLHttpRequest receives wrong binary array.
      [NEW]          Bindings-generator supports to bind 'unsigned long'.
      [FIX]          'Test Frame Event' of TestJavascript/CocoStudioArmatureTest Crashes.
+[Windows]
+     [NEW]          CMake support for windows.
+     
 cocos2d-x-3.0alpha1 Nov.19 2013
 [all platforms]
      [DOC]          Added RELEASE_NOTES and CODING_STYLE.md files
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8)
 project (Cocos2dx)
 
 # The version number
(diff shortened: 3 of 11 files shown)


cocos/editor-support/cocosbuilder/CCBReader.cpp
```diff
@@ -293,7 +293,16 @@ Node* CCBReader::readNodeGraphFromData(Data *pData, Object *pOwner, const Size &
...
-        pNode->setUserObject(manager);
+        
+        // avoid user object conflict with other implementation
+        auto dict = static_cast<Dictionary*>(pNode->getUserObject());
+        if (dict == nullptr)
+        {
+            dict = Dictionary::create();
+            pNode->setUserObject(dict);
+        }
+        dict->setObject(manager, "ccbmanager");
```
dumganhar, line 304: But if user object is not a `Dictionary`, the `static_cast` will make no sense.

btspoony, line 304: Hi, i am refer to here: 
https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/scripting/javascript/bindings/cocos2d_specifics.cpp#L3045

Maybe we should change the implementation of these parts.....

dumganhar, line 304: That's for JSBinding codes. We could manage it and make sure that `userObject` is `dict`.
JS code should never invoke `Node::setUserObject | setUserData` since it's easy for js to add extra properties to an object.

btspoony, line 304: or... just ensure the user object is a `Dictionary`...
... i think i know how to do it...

dumganhar, line 304: Could you add a new test to reproduce this issue in  `CocosBuilderTest`? 
We should find the best approach to resolve it. Thanks.

btspoony, line 304: where is `CocosBuilderTest` ?...
just now i am thinking may be we could use `dynamic_cast` to ensure `getUserObject` returns `Dictionary` ...

dumganhar, line 304: It's in `cocos2d-x/samples/Cpp/TestCpp/Classes/ExtensionsTest/CocosBuilderTest`.

Yep, I think we could use `dynamic_cast` for now.

btspoony, line 304: this issued will be reproduced by using JSB to load ccb file..
it seems CocosBuilderTest only for c++

dumganhar, line 304: No, you could refer to `cocos2d-x/samples/Javascript/Shared/tests/ExtensionsTest/CocosBuilderTest.`

btspoony:
this issue can be avoided, just DO NOT using CCLayer as the rootNode
currently close this issue =-=

PR #4401, opened 2013-11-27
add helper funcion to invoke a function in gl thread

minggo:


minggo:
I will modify `AssetsManager` and other modules to use `ThreadHelper`.

ricardoquesada:
We are synchronized :)

I've just implemented something similar for the console that I'm working on:
https://github.com/ricardoquesada/cocos2d-x/blob/console/cocos/2d/CCScheduler.cpp#L829

Instead of creating another class, I just added the `performFunctionInCocosThread` method in `Scheduler`.

IMO, `Scheduler` is better than `ThreadHelper` for this particular case.

What do you think about it ?
thanks.

minggo:
Ok, would you send a pull request for it?
I will change the codes.

ricardoquesada:
Yes, I'll send a PR once I finish the console code.
It will be next week.
thanks!

minggo:
Thanks. It sounds reasonable to implement it in `Schedule`.

PR #4411, opened 2013-11-30
cmake support for windows

martell:
Quickly added cmake support for windows

Tested on Visual Studio 2013

Instructions for Windows
1. install cmake and include in PATH variable.
2. go to build folder
3. run .cmd file corresponding to your visual studio version
   eg.  win32-msvc-2013-x86 for visual studio 2013
4. Open solution in created folder build/win32-msvc-2013-x86/cocos2dx
5. Set HelloCpp as the active project and try to run it :)
   Have fun

Can I get some feedback please?
I only hacked it together quickly so I want to know does it work for others.

Martell

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ set(Cocos2dxSamples_VERSION_MINOR 0)
 
 include(build/BuildHelpers.CMakeLists.txt)
 
+if(WIN32 OR APPLE)
+
 option(USE_CHIPMUNK "Use chipmunk for physics library" ON)
 option(USE_BOX2D "Use box2d for physics library" OFF)
 option(DEBUG_MODE "Debug or release?" ON)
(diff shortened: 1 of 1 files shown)


dumganhar:
There is an error on travis-ci, could you check that? https://travis-ci.org/cocos2d/cocos2d-x/jobs/14729723

martell:
Which Platform is that for?

EDIT: I realised its linux

I'll fix it soon

martell:
@dumganhar fixed ;)

dumganhar:
Thank you.

CMakeLists.txt
```diff
@@ -10,20 +10,19 @@ include(build/BuildHelpers.CMakeLists.txt)
...
-option(BUILD_EDITOR_SPINE "Build editor support for spine" ON)
-option(BUILD_EDITOR_COCOSTUDIO "Build editor support for cocostudio" ON)
-option(BUILD_EDITOR_COCOSBUILDER "Build editor support for cocosbuilder" ON)
+option(BUILD_LIBS_LUA "Build lua libraries" OFF)
+option(BUILD_GUI "Build GUI library" OFF)
+option(BUILD_NETWORK "Build network library" OFF)
+option(BUILD_EXTENSIONS "Build extension library" OFF)
+option(BUILD_EDITOR_SPINE "Build editor support for spine" OFF)
+option(BUILD_EDITOR_COCOSTUDIO "Build editor support for cocostudio" OFF)
+option(BUILD_EDITOR_COCOSBUILDER "Build editor support for cocosbuilder" OFF)
```
dumganhar, line 21: Why to set these switch to `OFF` by default?

dumganhar:
@martell 
Unfortunately, It still failed. Please check https://travis-ci.org/cocos2d/cocos2d-x/jobs/14792731 .

dumganhar:
If you have a Ubuntu VM, you could test on it.

CMakeLists.txt
martell, line 21: It would be best to get the minimal version working on Win Mac and Linux and then enable them as we go.

I can enable them for Linux by default?

martell:
Sorry I don't atm. 
Travis is quite handy :)

Just fix HelloCPP and that should do it

CMakeLists.txt
dumganhar, line 21: But the top level CMakeLists.txt is designed for building all libraries and projects.
All binaries will be placed in `cocos2d_root/bin` and all libraries will be placed in `cocos2d_root/lib`.
Template will uses the libraries  in `lib`  folder.
Could the template in `cocos2d_root/template/multi-platform-cpp/HelloCpp` use the CMakeLists.txt in the top level folder?

Now we only static link the libraries in `lib` folder. As I know, `add_subdirectory` can't do this, right?
If it could, I think `minimal version` you did will make sense.:)

martell, line 21: No I wouldn't use that one, I would create one from scratch in template/multi-platform-cpp/HelloCpp
Only variable it would need is COCOS_HOME directory
That then could be used by project_creator.py
that way the project could be put anywhere in the filesystem not just in the projects folder and still compile

I've reverted the Options for Linux :)

dumganhar, line 21: Sounds great. :)

martell:
@dumganhar It passed Travis Linux build now

dumganhar:
merged. thx.

dumganhar:
@martell 
TestCpp also works now by using cmake on windows.
Please refer to https://github.com/cocos2d/cocos2d-x/pull/5016 .

PR #4412, opened 2013-11-30
fix printf compile error in Linux

yvanc:
Fixed compile error when performing make under Ubuntu linux


dumganhar:
Which version of Ubuntu are you using? We are using `Travis` for CI. You could refer to 
https://travis-ci.org/cocos2d/cocos2d-x/builds
And there aren't any problems when compiling.

dumganhar:
Travis CI is using `Ubuntu 12.04 64bit`.

yvanc:
Linux mint 14 32-bit

dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/4908/files.
Thanks.

PR #4416, opened 2013-12-03
use size_t instead of long in Array

minggo:
1. use `size_t` instead of `long` in Array API
2. remove unneeded spaces
3. add spaces before and after binary operation, such as `+`

### DIFF
--- a/cocos/base/CCArray.cpp
+++ b/cocos/base/CCArray.cpp
@@ -105,7 +105,7 @@ Array* Array::createWithArray(Array* otherArray)
     return otherArray->clone();
 }
 
-Array* Array::createWithCapacity(long capacity)
+Array* Array::createWithCapacity(size_t capacity)
 {
     CCASSERT(capacity>=0, "Invalid capacity");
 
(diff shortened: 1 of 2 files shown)


minggo:
@ricardoquesada 
Could you please review it?

minggo:
I will add other commit after this one is merged.

minggo:
I will change the API first, then fix warnings later.

ricardoquesada:
Sure. I'll review tomorrow Tuesday.

minggo:
I am not sure if i use too much `size_t`. May be only `count` and `capacity` should use `size_t`.

ricardoquesada:
@minggo 
Looks good to me.

You are using `size_t` for size and index, and that is the expected use:
http://stackoverflow.com/a/1951548

minggo:
Ok.

minggo:
I change the type of `Array::getIndexOfObject()` to `size_t`, then what's the return value when the object is not contained in the Array?

minggo:
Ok, i can return the max value of size_t.

ricardoquesada:
Good point.

2 options:
-  return std::max() or something like that
- or `assert()` if the object doesn't exist

minggo:
Thanks. Since we will return `ssize_t` instead of `size_t`, i think it can return `-1`.

PR #4418, opened 2013-12-03
issue #2771 physical, fix some compile error when CC_USE_ARRAY_VECTOR set to 1

boyu0:
still have some compile error in UI part.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -251,9 +251,9 @@ void Label::alignText()
     int strLen = cc_wcslen(_currentUTF16String);  
     if (_children && _children->count() != 0)
     {
-        for (auto child: *_children)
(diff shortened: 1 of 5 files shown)


cocos/base/CCArray.cpp
```diff
@@ -258,12 +258,12 @@ void Array::addObjectsFromArray(Array* otherArray)
...
 }
 
-void Array::insertObject(Object* object, int index)
+void Array::insertObject(Object* object, long index)
 {
     data.insert( std::begin(data) + index, RCPtr<Object>(object) );
 }
 
-void Array::setObject(Object* object, int index)
+void Array::setObject(Object* object, long index)
```
minggo, line 266: Don't change `int` to `long`. Please refer to this thread: http://www.cocos2d-x.org/forums/6/topics/39953?page=1.

PR #4420, opened 2013-12-03
issue #2790: (ONLY FOR REVIEW NOW)Template Containers for Reference counted classes, Value for Pure Data

dumganhar:
Please don't merge, this PR is only for review. Thanks.
- Next: Fixes the compilation errors for JSB and LUA projects.

### DIFF
--- a/tools/travis-scripts/run-script.sh
+++ b/tools/travis-scripts/run-script.sh
@@ -74,7 +74,7 @@ elif [ "$PLATFORM"x = "linux"x ]; then
     cd $COCOS2DX_ROOT/build
     mkdir -p linux-build
     cd linux-build
-    cmake ../..
+    cmake ../.. -DBUILD_LIBS_LUA=OFF -DBUILD_HelloLua=OFF -DBUILD_TestLua=OFF
     make -j10
     cd ../../template/multi-platform-cpp
     cmake .


cocos/2d/CCAnimation.h
```diff
@@ -225,6 +213,30 @@ class CC_DLL Animation : public Object, public Clonable
 	virtual Animation *clone() const override;
 
 protected:
+    
+    /**
+     * @js ctor
+     */
```
minggo, line 100: Should remove this? Since it is a protected method.

dumganhar, line 100: I don't know whether `doxgen` doc deals with protected methods.

cocos/2d/platform/CCFileUtils.h
```diff
@@ -300,6 +296,24 @@ class CC_DLL FileUtils
     virtual void setPopupNotify(bool notify);
     virtual bool isPopupNotify();
 
+    /**
+     *  Converts the contents of a file to a ValueDict.
+     *  @note This method is used internally.
+     */
+    virtual ValueDict fileToValueDict(const std::string& filename);
```
minggo: Should modify the comment and have a better name.

cocos/base/CCArray.cpp
```diff
@@ -478,7 +478,7 @@ Array* Array::createWithContentsOfFile(const char* fileName)
 
 Array* Array::createWithContentsOfFileThreadSafe(const char* fileName)
 {
-    return FileUtils::getInstance()->createArrayWithContentsOfFile(fileName);
+    return nullptr;//FIXME:XXX  FileUtils::getInstance()->createArrayWithContentsOfFile(fileName);
```
minggo: Should fix it?

cocos/base/CCDictionary.cpp
```diff
@@ -399,7 +463,7 @@ Dictionary* Dictionary::createWithContentsOfFile(const char *pFileName)
 
 bool Dictionary::writeToFile(const char *fullPath)
 {
-    return FileUtils::getInstance()->writeToFile(this, fullPath);
+    return false;//FIXME: XXX FileUtils::getInstance()->writeToFile(this, fullPath);
```
minggo: Should fix it now?

cocos/base/CCMap.h
```diff
@@ -0,0 +1,252 @@
...
+            if (iter->second == object)
+            {
+                keys.push_back(iter->first);
+            }
+        }
+        
+        return keys;
+    }
+
+    V getObjectForKey(const K& key) const
```
minggo: How about rename to `getObject`?

minggo: Or rename it to `find` as std::map did?

kis2009dsh: How about "return std::move(keys);"    to use move ctor instead of copy ctor in std::vector ? 

发送时间: 2013年12月3日(星期二) 下午5:55

主题: Re: [cocos2d-x] issue #2790: (ONLY FOR REVIEW NOW)Template Containersfor Reference counted classes, Value for Pure Data (#4420)

In cocos/base/CCMap.h:

cocos/base/CCMap.h
```diff
@@ -0,0 +1,252 @@
...
+    }
+    
+    void removeObjectsForKeys(const std::vector<K>& keys)
+    {
+        std::for_each(keys.cbegin(), keys.cend(), [this](const K& key){
+            removeObjectForKey(key);
+        });
+    }
+    
+    void removeAllObjects()
```
minggo: Rename it to `clear`?

cocos/base/CCMap.h
```diff
@@ -0,0 +1,252 @@
...
+    V getObjectForKey(const K& key) const
+    {
+        auto iter = _data.find(key);
+        if (iter != _data.end())
+            return iter->second;
+        
+        return nullptr;
+    }
+    
+    void setObject(V object, const K& key)
```
minggo: Rename to `insert` as std::map did?

cocos/base/CCMap.h
```diff
@@ -0,0 +1,252 @@
...
+    {
+        auto iter = _data.find(key);
+        if (iter != _data.end())
+        {
+            iter->second->release();
+            _data.erase(iter);
+        }
+    }
+    
+    void removeObjectsForKeys(const std::vector<K>& keys)
```
kis2009dsh: Should we use "template<typename InIter> void removeObjectsForKeys( InIter first , InIter last )" instead of "std::vector<K>"? 
Because the type of keys could be std::map , or std::list , and we don't want to copy them into a new vector.

cocos/2d/CCAnimation.h
minggo, line 100: No matter doxygen deals with protected methods or not, we should remove it to avoid error in js documents.

cocos/base/CCDictionary.cpp
dumganhar: Yep, I will fix it tomorrow.

cocos/base/CCMap.h
dumganhar: @kis2009dsh 
The compiler which supports `c++11` will do a default optimization for this.

dumganhar: @minggo 
Map<T>::at(key_type) maybe a better name which is the same as std::map.

cocos/base/CCMap.h
dumganhar: OK. I will rename all methods with `stl` function naming.

cocos/base/CCMap.h
dumganhar: @kis2009dsh 
Good suggestion.
But the iterator of std::map is different from std::vector. How to implement this function?
Could you paste some codes for that? Thanks.

cocos/base/CCMap.h
dumganhar: OK.

cocos/2d/platform/CCFileUtils.h
dumganhar: How about `getValueDictFromFile` or `getValueDictWithFile`?

cocos/2d/CCAnimation.h
dumganhar, line 100: All right.

cocos/base/CCMap.h
kis2009dsh: @dumganhar 
Hi , I wrote a sample . 
<code>
template<typename InputIterator> void
removeObjectsForKeys( InputIterator first , InputIterator last ) 
{
      std::for_each( first , last , [this](const K& key) {
             removeObjectForKey( key );
      } );
}

//// Because all of the iterator in STL have same interface "operator++()" / "operator*()" / "operator->()" , called std::input_iterator_catagory 
</code>

dumganhar: OK, thank you. I will implement what you suggested. :)

samples/Cpp/TestCpp/Classes/PerformanceTest/PerformanceNodeChildrenTest.cpp
```diff
@@ -325,11 +325,11 @@ void IterateSpriteSheetCArray::update(float dt)
...
-    CCARRAY_FOREACH(children, object)
-    {
-        auto sprite = static_cast<Sprite*>(object);
-        sprite->setVisible(false);
-    }
+//FIXME: James    CCARRAY_FOREACH(children, object)
+//    {
+//        auto sprite = static_cast<Sprite*>(object);
+//        sprite->setVisible(false);
+//    }
```
minggo, line 332: Should fix it now?

samples/Cpp/TestCpp/Classes/PerformanceTest/PerformanceNodeChildrenTest.cpp
```diff
@@ -437,11 +436,11 @@ const char*  CallFuncsSpriteSheetForEach::testName()
 void CallFuncsSpriteSheetCMacro::update(float dt)
 {
     // iterate using fast enumeration protocol
-    auto children = batchNode->getChildren();
+    auto& children = batchNode->getChildren();
 
     CC_PROFILER_START(this->profilerName());
 
-    arrayMakeObjectsPerformSelector(children, getPosition, Node*);
+//FIXME: James    arrayMakeObjectsPerformSelector(children, getPosition, Node*);
```
minggo, line 443: Should fix it now? Use lambda.

PR #4424, opened 2013-12-03
Fix for broken WinRT and WP8 builds caused by change to rapidjson

stammen:
This pull request fixes broken WinRT and WP8 builds by the change to rapidjson. 

libextensions has been updated for WinRT and WP8.


minggo:
Thanks.

PR #4429, opened 2013-12-04
Change long to size_t or int

minggo:
1. use `size_t` in `ccCArray`
2. use `size_t` for data length, such as used in `Image` and `Texture2D`
3. change`long` to `int` in some places
4. fix some warnings and hungarian

### DIFF
--- a/cocos/2d/platform/android/CCFileUtilsAndroid.cpp
+++ b/cocos/2d/platform/android/CCFileUtilsAndroid.cpp
@@ -130,17 +130,17 @@ bool FileUtilsAndroid::isAbsolutePath(const std::string& strPath) const
 }
 
 
-unsigned char* FileUtilsAndroid::getFileData(const char* filename, const char* mode, long * size)
+unsigned char* FileUtilsAndroid::getFileData(const char* filename, const char* mode, size_t * size)
 {    
     return doGetFileData(filename, mode, size, false);
 }
 
-unsigned char* FileUtilsAndroid::getFileDataForAsync(const char* filename, const char* pszMode, long * pSize)
+unsigned char* FileUtilsAndroid::getFileDataForAsync(const char* filename, const char* pszMode, size_t * pSize)
 {
     return doGetFileData(filename, pszMode, pSize, true);
 }
 
-unsigned char* FileUtilsAndroid::doGetFileData(const char* filename, const char* mode, long * size, bool forAsync)
+unsigned char* FileUtilsAndroid::doGetFileData(const char* filename, const char* mode, size_t * size, bool forAsync)
 {
     unsigned char * data = 0;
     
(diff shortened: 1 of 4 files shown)


cocos/2d/CCDrawNode.cpp
```diff
@@ -142,7 +142,7 @@ DrawNode* DrawNode::create()
     return pRet;
 }
 
-void DrawNode::ensureCapacity(long count)
+void DrawNode::ensureCapacity(int count)
```
ricardoquesada, line 145: since this is capacity `size_t` is valid.
But `int` is OK too in this context I guess.

cocos/2d/CCTextureAtlas.cpp
```diff
@@ -110,7 +110,7 @@ void TextureAtlas::setQuads(V3F_C4B_T2F_Quad* quads)
 
 // TextureAtlas - alloc & init
 
-TextureAtlas * TextureAtlas::create(const char* file, long capacity)
+TextureAtlas * TextureAtlas::create(const char* file, int capacity)
```
ricardoquesada, line 113: same as before, capacity could be `size_t`, but in this context `int` is OK too.

cocos/2d/CCTextureAtlas.h
```diff
@@ -158,26 +158,26 @@ class CC_DLL TextureAtlas : public Object
...
      @since 1.1
     */
-    void fillWithEmptyQuadsFromIndex(long index, long amount);
+    void fillWithEmptyQuadsFromIndex(int index, int amount);
 
     /** draws n quads
     * n can't be greater than the capacity of the Atlas
     */
-    void drawNumberOfQuads(long n);
+    void drawNumberOfQuads(int n);
```
ricardoquesada, line 173: so, capacity, index, size, etc.. are valid cases for `size_t`, but 2 ^ 32 is big enough for any GL array.
So, it is OK to use `int` in this context.

ricardoquesada:
One more thing:

I've just realized that `size_t` is different than `ssize_t`, so I think we should replace `size_t` with `ssize_t` :-)

`ssize_t` means "signed size_t", and we should try to use "signed" types as much as possible.
https://github.com/cocos2d/cocos2d-x/blob/develop/docs/CODING_STYLE.md#on-unsigned-integers

So, this will also fix the issue regarding `return -1` when an object is not find in `getIndexOfObject`

minggo:
I think i have to close this pull request, since @dumganhar is implementing container in #4420. After merging #4420, `Array` and other container will be deprecated, many codes will use new container instead. And i will replace `size_t` with `ssize_t`.

ricardoquesada:
ok.

hulefei:
Will cocos2dx replace Array with Vector?

minggo:
Yep, we suggest to use cocos2d::Vector in 3.0.

minggo:
New pull request for this issue is #4438.

PR #4432, opened 2013-12-04
fix properties elementName;I checked the name used by Tile from 0.8.0 at least

mooosu:
I found cocos2d has same bug in all release versions.
the patch line is about around line 561.
and because I use VIM , it trims spaces automatically.

patch code:
-    } 
-    else if (elementName == "property")
-    }
-    else if (elementName == "properties")


ricardoquesada:
thanks.
what's the best way to test this patch ?
Do you have a .tmx file that reproduces this issue ?

PR #4433, opened 2013-12-05
Console

ricardoquesada:
A helper class that let's the user control the game remotely.
Currently it supports these commands: "fps on", "fps off", "scene graph", "help", "exit".

But it is easy to add new commands without modifying the cocos2d code.

It also adds a helper function in Scheduler that lets you run an `std::function<void()>` in the cocos2d thread.

Use cases:
- It can use for realtime debugging (useful for the new renderer)
- for QA automation

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -68,11 +68,8 @@ static void printSceneGraph(int fd, Node* node, int level)
 
     mydprintf(fd, " %s: z=%d, tag=%d\n", node->description(), node->getZOrder(), node->getTag());
 
-    auto children = node->getChildren();
-    if( children ) {
-        for(const auto& child: *children)
-            printSceneGraph(fd, static_cast<Node*>(child), level+1);
-    }
+    for(const auto& child: node->getChildren())
+        printSceneGraph(fd, child, level+1);
 }
 
 static void printSceneGraphBoot(int fd)


ricardoquesada:
Test on Mac, iOS, Linux and Android.

Haven't tested it on Windows. And most probably it won't work on Windows since Windows has some strange API calls for sockets. 
One option is to disable this feature for Windows until windows support is added.

cocos/2d/CCScheduler.cpp
```diff
@@ -950,8 +947,43 @@ void Scheduler::update(float dt)
...
+        }
+    }
+
+    //
+    // Functions allocated from another thread
+    //
+
+    // Testing size is faster than locking / unlocking.
+    // And almost never there will be functions scheduled to be called.
+    if( _functionsToPerform.size() > 0 ) {
```
dumganhar: I think `!empty()` is faster than `size() > 0`.

dumganhar:
I will test it on Windows. Thanks :).

dumganhar:
Could we use this `console` for `JSDebugger` ?

cocos/2d/CCScheduler.cpp
ricardoquesada: fixed. thanks.

ricardoquesada:
Yes, actually one very useful Use Case is the JS Debugger.

We can replace part of the logic of JS Debugger with the new Console.
We might need to add a few functions in the `Console` in order to support the JS Debugger properly, but I think it is worth doing it.

cocos/base/CCConsole.cpp
```diff
@@ -0,0 +1,399 @@
...
+        sched->performFunctionInCocosThread( std::bind(&Director::setDisplayStats, dir, true));
+    } },
+    { "fps off", [](int anFd) {
+        Director *dir = Director::getInstance();
+        Scheduler *sched = dir->getScheduler();
+        sched->performFunctionInCocosThread( std::bind(&Director::setDisplayStats, dir, false));
+    } },
+    { "scene graph", std::bind(&Console::commandSceneGraph, this, std::placeholders::_1) },
+    { "exit", std::bind(&Console::commandExit, this, std::placeholders::_1) },
+    { "help", std::bind(&Console::commandHelp, this, std::placeholders::_1) }, }
```
dumganhar: VS2012 doesn't support `initializer list`. So this usage will cause compilation errors on windows.

dumganhar:
Great.
Then, lots of `socket connection, thread switch` codes could be removed from `ScriptingCore`.

dumganhar:
Hi @ricardoquesada 
Could you let me know how to test this `console`, so I could make sure it also works on windows?

dumganhar:
Got it. I used `telnet 127.0.0.1 5678` and logon successfully. :)

ricardoquesada:
Thanks for merging it.

Regarding the JS Debugger, one possible way to implement it is by subclassing `Console` and adding all the needed functionality there.

`Console` was designed in a way that it is possible to have multiple instances of it (each one is going to run on its own thread listening to a different TCP port).

PR #4434, opened 2013-12-05
Msys2 and Mingw added fixed x64 cocos/audio on Windows for MSVC and MinGW

martell:
Fixed some bad cast's to long's also.
Arm64 will like these fixes. ;)
Small cmake cleanup

@minggo @dumganhar ;)
windows cocos2d-x x64 ScreenShot

http://i39.tinypic.com/24wh935.jpg

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,12 +195,6 @@ if(WIN32)
   )
   endif()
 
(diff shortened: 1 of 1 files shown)


dumganhar:
Looks great. Thanks. ;)

PR #4438, opened 2013-12-05
replace long with int or ssize_t

minggo:
1. Use `ssize_t` when reading file content or initialize with file content.
2. Use `int` instead of `long` for parameter and member variable in other places, including in `Vector` and `Map`. The reason of using `int` for container is that engine API will use the result after invoking container API, then it will cause more warnings. For example, 

``` cpp
void Node::removeChild(Node* child, bool cleanup /* = true */)
{
    // explicit nil handling
    if (_children.empty())
    {
        return;
    }

    auto index = _children.getIndex(child);
    if( index != CC_INVALID_INDEX )
        this->detachChild( child, index, cleanup );
}
```

If the result of `_children.getIndex(child)` is `ssize_t`, then `this->detachChild( child, index, cleanup )` will cause warnings.

### DIFF
--- a/cocos/physics/CCPhysicsWorld.cpp
+++ b/cocos/physics/CCPhysicsWorld.cpp
@@ -55,6 +55,12 @@ NS_CC_BEGIN
 const float PHYSICS_INFINITY = INFINITY;
 extern const char* PHYSICSCONTACT_EVENT_NAME;
 
+const int PhysicsWorld::DEBUGDRAW_NONE = 0x00;
+const int PhysicsWorld::DEBUGDRAW_SHAPE = 0x01;
+const int PhysicsWorld::DEBUGDRAW_JOINT = 0x02;
+const int PhysicsWorld::DEBUGDRAW_CONTACT = 0x04;
+const int PhysicsWorld::DEBUGDRAW_ALL = DEBUGDRAW_SHAPE | DEBUGDRAW_JOINT | DEBUGDRAW_CONTACT;
+
 namespace
 {
     typedef struct RayCastCallbackInfo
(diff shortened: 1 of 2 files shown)


minggo:
Because this pull request is big, so i merge it first. Please add comments if there is any error or suggestion.

cocos/2d/CCActionCatmullRom.cpp
```diff
@@ -43,7 +43,7 @@
  *  Implementation of PointArray
  */
 
-PointArray* PointArray::create(unsigned int capacity)
```
ricardoquesada, line 46: for this particular case `ssize_t` is better than `int` because it is a capacity.

cocos/2d/CCActionCatmullRom.cpp
```diff
@@ -154,9 +154,9 @@ void PointArray::removeControlPointAtIndex(unsigned int index)
     delete pRemovedPoint;
 }
 
-unsigned int PointArray::count() const
+int PointArray::count() const
 {
-    return _controlPoints->size();
+    return  static_cast<int>(_controlPoints->size());
```
ricardoquesada, line 159: this cast wouldn't be needed if we were using `ssize_t` instead of `int`

cocos/2d/CCActionManager.cpp
```diff
@@ -87,7 +87,7 @@ void ActionManager::actionAllocWithHashElement(tHashElement *element)
 
 }
 
-void ActionManager::removeActionAtIndex(long index, tHashElement *element)
```
ricardoquesada, line 90: for `index` and `capacity` it i OK to use `ssize_t`

cocos/2d/CCActionManager.h
```diff
@@ -126,7 +126,7 @@ class CC_DLL ActionManager : public Object
 protected:
     // declared in ActionManager.m
 
-    void removeActionAtIndex(long index, struct _hashElement *pElement);
```
ricardoquesada, line 129: same here: `ssize_t` is better for index

PR #4440, opened 2013-12-05
CCRenderTexture added constructors without scaling

seagullua:
The width and the height of the texture is scaled by CC_CONTENT_SCALE_FACTOR(). The three new constructors where added to give possibility to also create textures without automatic scale.

### DIFF
--- a/cocos2dx/misc_nodes/CCRenderTexture.cpp
+++ b/cocos2dx/misc_nodes/CCRenderTexture.cpp
@@ -243,12 +243,60 @@ CCRenderTexture * CCRenderTexture::create(int w, int h)
     return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


dumganhar:
Hi @seagullua 
Any test case for non-scaling usage?

seagullua:
The test case is simple. For rendering background I needed power of 2 texture. 
1. When I create the texture `CCRenderTexture::create(512, 512)` this value will be multiplied by Scale Factor and as a result the texture will not always be power of 2.
2. I tried to pass `CCRenderTexture::create(512 / scale_factor, 512 / scale_factor)`. This did not work either because float gets truncated to int and final texture can sometimes have size 511 or 513 which is not power of 2 either.

This non-scale constructors was the only way for me to create power of 2 texture on all devices.

PR #4447, opened 2013-12-06
Bug #3162: Merge Node and NodeRGBA, fix opacity incorrect bug.

boyu0:
Deprecate RGBAProtocol, NodeRGBA and LayerRGBA，add cascade opacity and cascade color test


minggo:
@ricardoquesada 
What's your opinion to merge NodeRGBA?

minggo:
@boyu0 
I think you should deprecated old functions and classes, can not just delete them.

dualface:
+1

Merge NodeRGBA to Node, remove all duplicate codes.

boyu0:
@minggo  ok, I will add them back and duplicate them.

ricardoquesada:
Interesting.

Some context:
`NodeRGBA` is kind of a weird node. Because it has `color` and `opacity` but you can't draw a `NodeRGBA`. So having this node is not very useful. Multiple inheritance would be better in this case... And a component-based model will solve it better way yet.

So, merging `NodeRGBA` into `Node` will simplify our code (+1 for this).
But will transfer the "weird" properties to `Node`. `Node` will have `opacity` and `color` but if you draw it nothing will happen.

What bug are we trying to fix with this PR ?
The inheritance of color and opacity ?
If so, then it seems OK to me.

But we should know that it transfer (or at least it doesn't fix) the "weird" issue of having a Node with `opacity` and `color`, but a Node can't be drawn.

When we move to a Component-base model, we need to fix this properly.

minggo:
This pull request is to fix [this issue](http://www.cocos2d-x.org/issues/3162).

boyu0:
@ricardoquesada 
I was trying to fix this bug: Child's opacity will not be changed when its parent's setCascadeOpacityEnabled was set to true and opacity was changed.
http://www.cocos2d-x.org/issues/2920
To fix this it would add lots of code, I found that classes inherit from RGBAProtocol( NodeRGBA, LayerRGBA, LabelBMFont and Label) have almost same code, it has so much duplicate code, It's hard to maintain. I thought NodeRGBA is a little overdesign, merge it to Node can simplify our engine, so I merge them.

ricardoquesada:
@boyu0 
correct. I think merging it is the right thing to do to fix that particular bug.

minggo:
@boyu0 
Could you please fix conflicts.
Then i will merge it.

cocos/2d/CCNode.h
```diff
@@ -1477,6 +1499,14 @@ class CC_DLL Node : public Object
...
     PhysicsBody* _physicsBody;        ///< the physicsBody the node have
 #endif
+    
+    // opacity controls
+	GLubyte		_displayedOpacity;
+    GLubyte     _realOpacity;
+	Color3B	    _displayedColor;
+    Color3B     _realColor;
+	bool		_cascadeColorEnabled;
+    bool        _cascadeOpacityEnabled;
```
minggo, line 1517: Error indent.

cocos/2d/CCNode.h
```diff
@@ -1485,46 +1515,38 @@ class CC_DLL Node : public Object
...
  All features from Node are valid, plus the following new features:
  - opacity
  - RGB colors
-
+ 
  Opacity/Color propagates into children that conform to the RGBAProtocol if cascadeOpacity/cascadeColor is enabled.
  @since v2.1
  */
-class CC_DLL NodeRGBA : public Node, public RGBAProtocol
+class CC_DLL CC_DEPRECATED_ATTRIBUTE NodeRGBA : public Node, public RGBAProtocol
```
minggo: Please put deprecated classes into Deprecated.h/cpp.

minggo:
And please add test case for this issue.
Thanks.

PR #4449, opened 2013-12-07
issue #2790: Deprecates CCArray in some places, uses Vector<T> instead.

dumganhar:


cocos/physics/CCPhysicsBody.h
```diff
@@ -103,9 +100,9 @@ class PhysicsBody : public Object
     /* remove all shapes */
     void removeAllShapes(bool reduceMassAndMoment = true);
     /* get the body shapes. */
-    inline Array* getShapes() const { return _shapes; }
+    inline const Vector<PhysicsShape*> getShapes() const { return _shapes; }
```
minggo: I think it is better to return a reference.

cocos/physics/CCPhysicsWorld.cpp
```diff
@@ -392,16 +392,16 @@ void PhysicsWorld::queryPoint(PhysicsPointQueryCallbackFunc func, const Point& p
...
     cpSpaceNearestPointQuery(this->_info->getSpace(),
                              PhysicsHelper::point2cpv(point),
                              0,
                              CP_ALL_LAYERS,
                              CP_NO_GROUP,
                              (cpSpaceNearestPointQueryFunc)PhysicsWorldCallback::getShapesAtPointFunc,
-                             arr);
+                             &arr);
     
     return arr;
```
minggo: Return std::move(arr)?

cocos/physics/CCPhysicsBody.cpp
```diff
@@ -368,7 +364,7 @@ PhysicsShape* PhysicsBody::addShape(PhysicsShape* shape, bool addMassAndMoment/*
     if (shape == nullptr) return nullptr;
     
     // add shape to body
-    if (_shapes->getIndexOfObject(shape) == CC_INVALID_INDEX)
+    if (_shapes.getIndex(shape) == -1)
```
minggo, line 367: Use `CC_INVALID_INDEX` instead?

cocos/physics/CCPhysicsBody.h
dumganhar: Ok.

cocos/physics/CCPhysicsWorld.cpp
dumganhar: Make sense.

cocos/physics/CCPhysicsBody.cpp
dumganhar, line 367: CC_INVAILD_INDEX was defined in ccCArray.h. But the new Vector<T> will not include that header file.

extensions/GUI/CCScrollView/CCTableView.cpp
```diff
@@ -104,28 +97,24 @@ void TableView::setVerticalFillOrder(VerticalFillOrder fillOrder)
...
 {
     _oldDirection = Direction::NONE;
-    Object* pObj = NULL;
-    CCARRAY_FOREACH(_cellsUsed, pObj)
-    {
-        TableViewCell* cell = static_cast<TableViewCell*>(pObj);
 
+    _cellsUsed.forEach([this](TableViewCell* cell){
         if(_tableViewDelegate != NULL) {
             _tableViewDelegate->tableCellWillRecycle(this, cell);
```
minggo, line 103: May be should use `this-> _tableViewDelegate`, or it will cause compiling error in some platform.

extensions/GUI/CCScrollView/CCTableView.cpp
```diff
@@ -181,22 +174,20 @@ void TableView::insertCellAtIndex(long idx)
...
 
-    cell = static_cast<TableViewCell*>(_cellsUsed->objectWithObjectID(idx));
+    auto cell = cellAtIndex(idx);
     if (cell)
     {
-        newIdx = _cellsUsed->indexOfSortedObject(cell);
-        for (long i = newIdx; i<_cellsUsed->count(); i++)
+        newIdx = _cellsUsed.getIndex(cell);
+        // Move all cells behind the inserted position
+        for (long i = newIdx; i < _cellsUsed.size(); i++)
```
minggo, line 184: Why use `long`?

extensions/GUI/CCScrollView/CCTableView.cpp
dumganhar, line 103: It only causes problems when using template. As used in Vector<T> or Map<K, V>.

extensions/GUI/CCScrollView/CCTableView.cpp
dumganhar, line 184: It was changed from `int` to `long` at previous version.

minggo, line 184: Ok, i will change them in another pull request.

extensions/GUI/CCScrollView/CCTableView.cpp
minggo, line 103: Ok.

cocos/physics/CCPhysicsBody.cpp
minggo, line 367: But it uses `CC_INVAILD_INDEX` in other places. May be we should keep them consistent.

dumganhar, line 367: Where shall we put these `const` variable?

PR #4471, opened 2013-12-10
fix compile error for vs project.

WenhaiLin:


cocos/2d/ZipUtils.h
```diff
@@ -31,6 +31,8 @@ of this software and associated documentation files (the "Software"), to deal
 
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
 #include "platform/android/CCFileUtilsAndroid.h"
+#elif(CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
+#include "CCStdC.h"
```
dumganhar, line 36: Please add comments for why we have to include this header file only for windows. Thanks.

cocos/editor-support/spine/extension.h
```diff
@@ -54,6 +54,10 @@
 #ifndef SPINE_EXTENSION_H_
 #define SPINE_EXTENSION_H_
 
+#if defined(_MSC_VER)
+#include "CCStdC.h"
```
dumganhar, line 59: The same here.

PR #4473, opened 2013-12-11
issue #2790: Deprecates CCDictionary, uses ValueMap for pure base data. Adds StringUtils class .

dumganhar:
- Adds Value::getDescription() for dumping logs.

### DIFF
--- a/cocos/scripting/lua/bindings/LuaOpengl.cpp
+++ b/cocos/scripting/lua/bindings/LuaOpengl.cpp
@@ -5322,8 +5322,8 @@ static int tolua_Cocos2d_CCGLProgram_vertexShaderLog00(lua_State* tolua_S)
         if (!self) tolua_error(tolua_S,"invalid 'self' in function 'vertexShaderLog'", NULL);
 #endif
         {
-            const char* tolua_ret = (const char*)  self->getVertexShaderLog();
-            tolua_pushstring(tolua_S,(const char*)tolua_ret);
+            std::string tolua_ret = self->getVertexShaderLog();
+            tolua_pushstring(tolua_S, tolua_ret.c_str());
         }
     }
     return 1;
@@ -5354,8 +5354,8 @@ static int tolua_Cocos2d_CCGLProgram_fragmentShaderLog00(lua_State* tolua_S)
         if (!self) tolua_error(tolua_S,"invalid 'self' in function 'fragmentShaderLog'", NULL);
 #endif
         {
-            const char* tolua_ret = (const char*)  self->getFragmentShaderLog();
-            tolua_pushstring(tolua_S,(const char*)tolua_ret);
+            std::string tolua_ret = self->getFragmentShaderLog();
+            tolua_pushstring(tolua_S, tolua_ret.c_str());
         }
     }
     return 1;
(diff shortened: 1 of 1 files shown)


cocos/base/CCString.h
```diff
@@ -192,15 +194,53 @@ class CC_DLL String : public Object, public Clonable
...
 #define StringMake(str) String::create(str)
-#define ccs               StringMake
+#define ccs             StringMake
+
+class StringUtils
+{
+public:
+    
+    template<typename T>
+    static std::string toString(T arg)
```
michaelcontento, line 212: Why don't you just use `std::to_string`? 
Because it's missing on android/gnustl?
Maybe we should stick with `std::to_string` and add the missing functions to the `std` namespace **only** for android/gnustl builds?

I know .. it's not the best option to mess within `std::` **BUT**:
- it will keep the library clean
- we only have to touch a single spot, if gnustl is updated
- we fix the problem for the user too (otherwise the problem continues within the users game code)

cocos/base/CCString.h
```diff
@@ -192,15 +194,53 @@ class CC_DLL String : public Object, public Clonable
...
+        bool operator() (__String * a, __String * b) const {
             return strcmp(a->getCString(), b->getCString()) < 0;
         }
 };
 
 #define StringMake(str) String::create(str)
-#define ccs               StringMake
+#define ccs             StringMake
+
+class StringUtils
```
michaelcontento, line 207: Why is this a class? It just could be a simple namespace with a few functions, right?

cocos/base/CCString.h
```diff
@@ -192,15 +194,53 @@ class CC_DLL String : public Object, public Clonable
...
+    
+    template<typename T>
+    static std::string toString(T arg)
+    {
+        std::stringstream ss;
+        ss << arg;
+        return ss.str();
+    }
+    
+    static std::string toStringWithFormat(const char* format, ...) CC_FORMAT_PRINTF(1, 2)
```
michaelcontento: Maybe `StringUtils::printf` is a better name? It clearly indicates the formatting rules (printf-style) and does not contain the word "String" twice ;)

dumganhar: agree!
- James

cocos/base/CCString.h
dumganhar, line 212: Good point. Thanks, i will check it out!
- James

cocos/base/CCString.h
dumganhar, line 207: You're right!
- James

cocos/base/CCString.h
dumganhar: @michaelcontento 
But the `C` printf is a `void` function, here we return `std::string` rather than output the result to console.
So do you still think `printf` makes sense?

cocos/base/CCString.h
dumganhar, line 212: Yep, there were not `std::to_string` in `GCC 4.8` of android toolchain.
@minggo @ricardoquesada 
Do you think that hacking `std` namespace is a good idea?

cocos/base/CCString.h
michaelcontento: Hm .. true. There is not a single `*printf` function that returns the result buffer/string, so it may be confusing to name it `printf`. What about just `StringUtils::format`? It's easy to learn / recognize that `format` follows the `printf` rules and it's short and does not contains "String" twice ;)

PR #4479, opened 2013-12-12
use ssize_t for index and capacity

minggo:


cocos/2d/CCTMXLayer.cpp
```diff
@@ -558,8 +558,8 @@ void TMXLayer::removeChild(Node* node, bool cleanup)
 
     CCASSERT(_children.contains(sprite), "Tile does not belong to TMXLayer");
 
-    unsigned int atlasIndex = sprite->getAtlasIndex();
-    unsigned int zz = (size_t)_atlasIndexArray->arr[atlasIndex];
+    ssize_t atlasIndex = sprite->getAtlasIndex();
+    ssize_t zz = (size_t)_atlasIndexArray->arr[atlasIndex];
```
ricardoquesada, line 562: no need to do the cast of `(size_t)`.

cocos/network/HttpResponse.h
```diff
@@ -150,7 +150,7 @@ class HttpResponse : public cocos2d::Object
     
     /** Set the http response errorCode
      */
-    inline void setResponseCode(int value)
+    inline void setResponseCode(long value)
```
ricardoquesada, line 153: probable `int` here is correct.

cocos/2d/CCTMXLayer.cpp
dumganhar, line 562: It's needed since `ccCArray` stores element with `Object*` type. But I think we should cast it to `ssize_t` rather than `size_t`.

Do we still need to use `ccCArray` in our engine codes as we have deprecated `CCArray`.
Could we just use `std::vector` for non-reference counted type?

cocos/network/HttpResponse.h
dumganhar, line 153: The place uses this method is at `setResponseCode` in HttpClient.cpp Line 191.

dumganhar, line 153: And the `processGetTask`, `processPostTask` ... methods access `long` as argument. So I think @minggo changed `int` to `long` is to resolve that warning on 64bit system.

cocos/2d/CCTMXLayer.cpp
ricardoquesada, line 562: Yes, correct, cast to `ssize_t` is better.
We should not use `ccArray` anymore.
We can replace it with `std::vector<>` or `cocos2d::Vector<>`

minggo, line 562: Ok, i will modify it.

cocos/network/HttpResponse.h
minggo, line 153: @dumganhar 
You are right. I will try to use int for other places if possible.

PR #4488, opened 2013-12-13
Fixed "Scew" property in animation

AIGRIND-LLC:
This fix adds "skew" property to the list of animated properties in NodeLoader::parsePropTypeFloatXY.
After that CCBAnimationManager class can implement backward "skew" transformation when animation changes in CCBAnimationManager::runAnimationsForSequenceIdTweenDuration.

### DIFF
--- a/cocos/2d/CCGLProgram.cpp
+++ b/cocos/2d/CCGLProgram.cpp
@@ -298,17 +298,17 @@ void GLProgram::use()
     GL::useProgram(_program);
 }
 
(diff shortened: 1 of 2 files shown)


dumganhar:
Thanks. Could you add some test case in TestCpp/TestJavascript?

AIGRIND-LLC:
If shader is invalid, program crashed. …
Shuld view assert + log error.
Platform: Windows
Explain: Functions in glew32.dll have unstandart call conversion. Typecast (GLInfoFunction) is invalid.

PR #4493, opened 2013-12-16
update for ui animation frame ease type

liamcindy:


minggo:
You should modify it for all samples and template.

PR #4496, opened 2013-12-16
issue #3341:fix TextureCache::addImageAsync repeatedly execute for the same image

WenhaiLin:


cocos/2d/CCTextureCache.cpp
```diff
@@ -170,17 +170,40 @@ void TextureCache::loadImage()
...
+        auto it = _textures.find(asyncStruct->filename);
+        if( it == _textures.end() )
         {
-            CC_SAFE_RELEASE(image);
-            CCLOG("can not load %s", filename);
-            continue;
+           _imageInfoMutex.lock();
+           ImageInfo *imageInfo;
+           unsigned int pos = 0;
+           unsigned int infoSize = _imageInfoQueue->size();
```
minggo: Is `size()` return `unsigned int`?

minggo: I think it should be `size_t` here.

WenhaiLin: std：size_t?

minggo: size_t or std::size_t both are ok.

PR #4497, opened 2013-12-16
reduce useless convert in CCScrollView.cpp

zhiqiangxu:
```
_container->convertToWorldSpace(_container->convertTouchToNodeSpace(touch))
```

is the same as 

```
touch->getLocation()
```


dumganhar:
@zhiqiangxu , good finding.

dumganhar:
Thanks. Issue http://www.cocos2d-x.org/issues/3559 was created.

PR #4499, opened 2013-12-17
New renderer

ricardoquesada:
New Renderer for cocos2d-x

### DIFF
--- a/samples/Cpp/TestCpp/Classes/ExtensionsTest/ExtensionsTest.cpp
+++ b/samples/Cpp/TestCpp/Classes/ExtensionsTest/ExtensionsTest.cpp
@@ -124,8 +124,14 @@ void ExtensionsMainLayer::onEnter()
     
     _eventDispatcher->addEventListenerWithSceneGraphPriority(listener, this);
     
+    auto mouseListener = EventListenerMouse::create();
+    mouseListener->onMouseScroll = CC_CALLBACK_1(ExtensionsMainLayer::onMouseScroll, this);
+    
+    _eventDispatcher->addEventListenerWithSceneGraphPriority(mouseListener, this);
+    
     addChild(_itemMenu);
 }
+
 
 void ExtensionsMainLayer::onTouchesBegan(const std::vector<Touch*>& touches, Event  *event)
 {
@@ -158,6 +164,30 @@ void ExtensionsMainLayer::onTouchesMoved(const std::vector<Touch*>& touches, Eve
 
     _itemMenu->setPosition(nextPos);
     _beginPos = touchLocation;
+    s_tCurPos   = nextPos;
+}
+
+void ExtensionsMainLayer::onMouseScroll(Event* event)
+{
+    auto mouseEvent = static_cast<EventMouse*>(event);
+    float nMoveY = mouseEvent->getScrollY() * 6;
+    
+    auto curPos  = _itemMenu->getPosition();
+    auto nextPos = Point(curPos.x, curPos.y + nMoveY);
+    
+    if (nextPos.y < 0.0f)
+    {
+        _itemMenu->setPosition(Point::ZERO);
+        return;
+    }
+    
+    if (nextPos.y > ((g_maxTests + 1)* LINE_SPACE - VisibleRect::getVisibleRect().size.height))
+    {
+        _itemMenu->setPosition(Point(0, ((g_maxTests + 1)* LINE_SPACE - VisibleRect::getVisibleRect().size.height)));
+        return;
+    }
+    
+    _itemMenu->setPosition(nextPos);
     s_tCurPos   = nextPos;
 }
 
--- a/samples/Cpp/TestCpp/Classes/ExtensionsTest/ExtensionsTest.h
+++ b/samples/Cpp/TestCpp/Classes/ExtensionsTest/ExtensionsTest.h
@@ -15,6 +15,9 @@ public:
     Menu* _itemMenu;
 
 	int _testcount;
+    
+protected:
+    void onMouseScroll(Event* event);
 };
 
 class ExtensionsTestScene : public TestScene


cocos/2d/CCRenderTexture.h
```diff
@@ -59,35 +59,35 @@ class CC_DLL RenderTexture : public Node
     /** creates a RenderTexture object with width and height in Points, pixel format is RGBA8888 */
     static RenderTexture * create(int w, int h);
 
-    /** starts grabbing */
+    virtual /** starts grabbing */
     void begin();
```
dumganhar, line 63: Should the `virtual` keyword be at the beginning of `void begin();`?

ricardoquesada, line 63: Yes... i"m not sure why it is before the `/* ... */` . ha!

cocos/2d/renderer/CCNewSpriteBatchNode.cpp
```diff
@@ -0,0 +1,83 @@
...
+void NewSpriteBatchNode::draw()
+{
+    // Optimization: Fast Dispatch
+    if( _textureAtlas->getTotalQuads() == 0 )
+    {
+        return;
+    }
+
+    for(const auto &child: _children)
+        child->updateTransform();
```
dumganhar, line 62: Using `std::for_each` or `_children.forEach()` will be faster.

cocos/2d/renderer/GroupCommand.cpp
```diff
@@ -0,0 +1,106 @@
...
+    for(auto it = _groupMapping.begin(); it != _groupMapping.end(); ++it)
+    {
+        if(!it->second)
+        {
+            return it->first;
+        }
+    }
+
+    //Create new ID
+//    int newID = _groupMapping.size();
```
dumganhar, line 55: This line isn't needed now? If so, we need to delete it.

cocos/2d/renderer/GroupCommand.h
```diff
@@ -0,0 +1,68 @@
...
+    ~GroupCommandManager();
+
+    bool init();
+
+    int getGroupID();
+    void releaseGroupID(int groupID);
+
+protected:
+    GroupCommandManager();
+    map<int, bool> _groupMapping;
```
dumganhar, line 32: Using `std::unordered_map` will be faster.

cocos/2d/renderer/MaterialManager.h
```diff
@@ -0,0 +1,50 @@
...
+
+    bool init();
+
+    int getTextureID(GLuint textureID);
+    int getShaderID(GLuint shaderID);
+    int getBlendFuncID(GLint blendFunc);
+
+    map<GLuint, int> _textureIDMapping;
+    map<GLuint, int> _shaderIDMapping;
+    map<BlendFunc, int> _blendFuncMapping;
```
dumganhar, line 45: The same here. Please use the faster `std::unordered_map`.

cocos/2d/renderer/NewClippingNode.cpp
```diff
@@ -0,0 +1,281 @@
...
+// this will allow nesting up to n ClippingNode,
+// where n is the number of bits of the stencil buffer.
+static GLint layer = -1;
+
+static void setProgram(Node *n, GLProgram *p)
+{
+    n->setShaderProgram(p);
+
+    for(const auto &child : n->getChildren())
+        setProgram(child, p);
```
dumganhar, line 24: Using `std::for_each` or `Vector<T>.forEach` will be faster.

cocos/2d/renderer/RenderCommandPool.h
```diff
@@ -0,0 +1,84 @@
...
+    {
+    }
+    ~RenderCommandPool()
+    {
+//        if( 0 != _usedPool.size())
+//        {
+//            CCLOG("All RenderCommand should not be used when Pool is released!");
+//        }
+        _freePool.clear();
+        for (typename std::list<T*>::iterator iter = _allocatedPoolBlocks.begin(); iter != _allocatedPoolBlocks.end(); ++iter)
```
dumganhar, line 31: Using `auto` keyword for `iter` will make things simpler.

stonejiang, line 31: yes. C++0x is better.

cocos/2d/CCRenderTexture.h
dumganhar, line 63: `end` has this issue too. :)

cocos/editor-support/cocostudio/CCTransformHelp.h
```diff
@@ -47,7 +47,9 @@ class  TransformHelp
     static void transformToParentWithoutScale(BaseData &node, const BaseData &parentNode);
 
     static void nodeToMatrix(const BaseData &_node, cocos2d::AffineTransform &_matrix);
+    static void nodeToMatrix(const BaseData &node, kmMat4 &matrix);
     static void matrixToNode(const cocos2d::AffineTransform &_matrix, BaseData &_node);
+    static void matrixToNode(const kmMat4 &_matrix, BaseData &_node);
```
dumganhar, line 52: We should not use `underline` for parameters. I think you did a `copy & paste` since the line `49` has the same issue.
It's an old issue that hasn't been found.

samples/Cpp/TestCpp/Classes/controller.cpp
```diff
@@ -19,6 +19,7 @@ struct {
     // TESTS MUST BE ORDERED ALPHABETICALLY
     //     violators will be prosecuted
     //
+    { "ANewRenderTest", []() { return new NewRendererTestScene(); } },
```
dumganhar, line 22: `ANewRenderTest`, Typo? Should it be `NewRendererTest`?

cocos/2d/renderer/CCNewSpriteBatchNode.cpp
ricardoquesada, line 62: Why is `forEach()` faster than `for_each`.

I thought that their implementation were the same.

cocos/2d/renderer/GroupCommand.h
ricardoquesada, line 32: agreed

cocos/2d/renderer/GroupCommand.cpp
ricardoquesada, line 55: Not sure... perhaps it was added for debugging, or perhaps that feature is still in development.
Minggo wants to merge the new renderer ASAP, even though it is not 100% finished.

cocos/2d/renderer/CCNewSpriteBatchNode.cpp
dumganhar, line 62: Yep, the implementation of `Vector<T>::forEach`is using `std::for_each`.
I meant that `for_each` was faster than `Range Loop`.

cocos/2d/renderer/GroupCommand.cpp
dumganhar, line 55: I know that. I will make `Linux` and `Android` work today.

cocos/2d/renderer/CCNewSpriteBatchNode.cpp
ricardoquesada, line 62: Interesting. Have you done any performance tests that compares `forEach()` vs. `for_each()` ?
Where can I find the results ?
thanks.

dumganhar, line 62: Hey, actually, it's you did the performance test here (https://docs.google.com/a/cocos2d-x.org/spreadsheet/ccc?key=0AvvkdgVbWvpZdHFudzdDT3NuYTRNTHlZZzRGZWYzMmc&usp=drive_web#gid=0).

ricardoquesada, line 62: Yes, thanks. But I don't think I tested `::forEach`. 

Which tests are you looking at particular ? Thanks!

dumganhar:
Thanks, Merged at https://github.com/cocos2d/cocos2d-x/pull/4502.

cocos/2d/renderer/CCNewSpriteBatchNode.cpp
dumganhar, line 62: OK, I will test `Vector<T>::forEach`, in fact, it's just a wrapper of `std::for_each`.

dumganhar, line 62: I tested it on iphone 4S, iterating 2500 nodes , Vector<T>::forEach uses 700us while `std::for_each` uses 600us.
So I think we should delete the function `Vector::forEach`.
I will do that. Thanks.

ricardoquesada, line 62: probably `std::for_each()` and `Vector::forEach` have the same performance.
If you take a look at the result of my tests,  the results are similar... some of them are a bit faster but we are talking about a few milliseconds of difference.

And since the difference is so small, I still don't know if one method is faster than the other, or is it because we don't have reliable ways to test the performance of these functions.

dumganhar, line 62: Yep, I agree. So do you think `Vector::forEach` is needed?  Is it duplicated with `std::for_each`?
And `Vector::forEach` will travel through all elements in Vector, while `std::for_each` could let you assign the range [first, last) elements needed to be travelled. @minggo also think these methods are duplicated.
If you like it , let's discuss more benefits of it. Thanks.

ricardoquesada, line 62: I do not think they are needed.
We can use `std::for_each`... or even `for(  auto x: vector)` if applicable, so I don't think they are needed.

dumganhar, line 62: Agree. :)
So now `Vector<T>` supports:
- Normal Loop: for (ssize_t i = 0; i < vec.size(); ++i) { auto e = vec.at(i); ... }
- Rangle Loop: for (auto& e : vec)
- STL for each: std::for_each(vec.begin(), vec.end(), function)

PR #4500, opened 2013-12-17
[Win32] fix compile error.

pyrasis:
- fix argument type. TableViewTestLayer::tableCellSizeForIndex, TableViewTestLayer::tableCellAtIndex


dumganhar:
Thanks.

PR #4507, opened 2013-12-17
Removes Vector<T>::forEach, please use std::for_each instead. The same as Vector<T>::sort.

dumganhar:


gelldur:
Sorry why do you use std::for_each instead of for-range from C++11 ? If performance reason can you link any resources? Thanks.

dumganhar:
@gelldur, Thank you for review this PR.
The reason I removed it was:
- It was duplicated with std::for_each.
- Vector::forEach will travel through all elements in Vector, while std::for_each could let you assign the range [first, last) elements needed to be travelled. It's more elegant.
- Vector::forEach is just a simple wrapper of `std::for_each`. In my opinion, using `std::for_each` is enough.

What's your opinion?

You could see more discusses at this PR(https://github.com/cocos2d/cocos2d-x/pull/4499).

PR #4513, opened 2013-12-17
Fixes to asset manager to ensure there is no crash on DNS resolver as well as no hanging if connection breaks during download

gamechefs:
This patch fixes the AssetsManager by ensuring that proper curl settings are done

AssetsManager (and the app using it) crashes if CURLOPT_NOSIGNAL is not set and there is a DNS resolution issue.  This is a known bug in curl on systems using posix threads and the recommended fix is applied here.

The AssetsManager perpetually waits if there is a connection timeout. The speed limit check enables curl to abort the download with a proper error code if there is a network connectivity issue.

### DIFF
--- a/extensions/assets-manager/AssetsManager.cpp
+++ b/extensions/assets-manager/AssetsManager.cpp
@@ -51,6 +51,10 @@ NS_CC_EXT_BEGIN;
 #define BUFFER_SIZE    8192
 #define MAX_FILENAME   512
 
+#define LOW_SPEED_LIMIT 1L
+#define LOW_SPEED_TIME 5L
+
+
 // Message type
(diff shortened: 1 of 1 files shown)


dumganhar:
@gamechefs @omersaeed 
This patch looks great.  Thanks a lot.

> This is a known bug in curl on systems using posix threads and the recommended fix is applied here.

Could you please point me out about where is this recommended fix?

dumganhar:
Merged , thanks.

gamechefs:
https://www.mail-archive.com/xmms2-devel@lists.xmms2.org/msg00039.html
http://curl.haxx.se/libcurl/c/curl_easy_setopt.html

PR #4529, opened 2013-12-18
Refactors Data class, adds FileUtils::getStringFromFile, FileUtils::getDataFromFile and deprecates FileUtils::getFileData.

dumganhar:


cocos/2d/platform/CCFileUtils.h
```diff
@@ -76,15 +77,26 @@ class CC_DLL FileUtils
...
     /**
+     *  Gets string from a file.
+     */
+    virtual std::string getStringFromFile(const std::string& filename);
+    
+    /**
+     *  Creates binary data from a file.
+     *  @return A data object.
+     */
+    virtual Data getDataFromFile(const std::string& filename);
```
dumganhar, line 88: @ricardoquesada , @minggo 
Do you like these APIs ?
We need to deprecate `FileUtils::getFileData` since it will make developer confused with who needs to release the memory.
The new APIs will resolve the problem as the memory will be released  in the destructor of `Data`. Developer doesn't need to take care of the memory.

minggo, line 88: It seems good to me.

PR #4534, opened 2013-12-18
Edit hungarian notation

boyu0:
Change NULL to nullptr, edit hungarian notation.


cocos/2d/CCComponent.cpp
```diff
@@ -61,7 +61,7 @@ void Component::serialize(void *ar)
 Component* Component::create(void)
 {
     Component * pRet = new Component();
```
minggo: Use `pRet` here.

minggo:
Well done.
Please fix conflict, then merge it ASAP.

Thanks.

PR #4543, opened 2013-12-19
Node cleanup

ricardoquesada:
Uses "normal" interation without the lambda to prevent another object in the stack.

removed unused code.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -51,6 +51,7 @@ THE SOFTWARE.
 
 // externals
 #include "kazmath/GL/matrix.h"
+#include "kazmath/vec4.h"
 #include "CCComponent.h"
 #include "CCComponentContainer.h"
 
@@ -171,10 +172,7 @@ Node::~Node()
 
     for (auto& child : _children)
     {
-        if (child)
-        {
-            child->_parent = NULL;
-        }
+        child->_parent = NULL;
     }
 
     removeAllComponents();
@@ -576,9 +574,8 @@ void Node::cleanup()
     }
     
     // timers
-    std::for_each(_children.begin(), _children.end(), [](Node* child){
+    for( const auto &child: _children)
         child->cleanup();
-    });
 }
 
 
@@ -713,33 +710,26 @@ void Node::removeAllChildren()
 void Node::removeAllChildrenWithCleanup(bool cleanup)
 {
     // not using detachChild improves speed here
-    if (!_children.empty())
+    for (auto& child : _children)
     {
-        for (auto& child : _children)
+        // IMPORTANT:
+        //  -1st do onExit
+        //  -2nd cleanup
+        if(_running)
         {
-            if (child)
-            {
-                // IMPORTANT:
-                //  -1st do onExit
-                //  -2nd cleanup
-                if(_running)
-                {
-                    child->onExitTransitionDidStart();
-                    child->onExit();
-                }
-
-                if (cleanup)
-                {
-                    child->cleanup();
-                }
-                // set parent nil at the end
-                child->setParent(nullptr);
-            }
+            child->onExitTransitionDidStart();
+            child->onExit();
         }
-        
-        _children.clear();
+
+        if (cleanup)
+        {
+            child->cleanup();
+        }
+        // set parent nil at the end
+        child->setParent(nullptr);
     }
     
+    _children.clear();
 }
 
 void Node::detachChild(Node *child, ssize_t childIndex, bool doCleanup)
@@ -793,39 +783,10 @@ void Node::reorderChild(Node *child, int zOrder)
 
 void Node::sortAllChildren()
 {
-#if 0
-    if (_reorderChildDirty)
-    {
-        int i,j,length = _children.size();
-
-        // insertion sort
-        for(i=1; i<length; i++)
-        {
-            j = i-1;
-            auto tempI = static_cast<Node*>( _children.at(i) );
-            auto tempJ = static_cast<Node*>( _children.at(j) );
-
-            //continue moving element downwards while zOrder is smaller or when zOrder is the same but mutatedIndex is smaller
-            while(j>=0 && ( tempI->_ZOrder < tempJ->_ZOrder || ( tempI->_ZOrder == tempJ->_ZOrder && tempI->_orderOfArrival < tempJ->_orderOfArrival ) ) )
-            {
-                _children.fastSetObject( tempJ, j+1 );
-                j = j-1;
-                if(j>=0)
-                    tempJ = static_cast<Node*>( _children.at(j) );
-            }
-            _children.fastSetObject(tempI, j+1);
-        }
-
-        //don't need to check children recursively, that's done in visit of each child
-
-        _reorderChildDirty = false;
-    }
-#else
     if( _reorderChildDirty ) {
         std::sort( std::begin(_children), std::end(_children), nodeComparisonLess );
         _reorderChildDirty = false;
     }
-#endif
 }
 
 
@@ -871,10 +832,8 @@ void Node::visit()
         // self draw
         this->draw();
 
-        // Uses std::for_each to improve performance.
-        std::for_each(_children.cbegin()+i, _children.cend(), [](Node* node){
-            node->visit();
-        });
+        for(auto it=_children.cbegin()+i; it<_children.cend(); ++it)
+            (*it)->visit();
     }
     else
     {
@@ -939,9 +898,8 @@ void Node::onEnter()
 {
     _isTransitionFinished = false;
 
-    std::for_each(_children.begin(), _children.end(), [](Node* child){
+    for( const auto &child: _children)
         child->onEnter();
-    });
 
     this->resume();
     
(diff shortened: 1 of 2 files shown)


cocos/2d/CCNode.cpp
```diff
@@ -576,9 +574,8 @@ void Node::cleanup()
     }
     
     // timers
-    std::for_each(_children.begin(), _children.end(), [](Node* child){
+    for( const auto &child: _children)
```
dumganhar, line 577: But the performance you did showed that `std::for_each` is faster than `range loop` although `std::for_each` does allocate a stack object `child` on each iteration.

cocos/2d/CCNode.cpp
```diff
@@ -871,10 +832,8 @@ void Node::visit()
         // self draw
         this->draw();
 
-        // Uses std::for_each to improve performance.
-        std::for_each(_children.cbegin()+i, _children.cend(), [](Node* node){
-            node->visit();
-        });
+        for(auto it=_children.cbegin()+i; it<_children.cend(); ++it)
```
dumganhar, line 835: `it != _children.cend()` will be more reasonable. See the usage here (http://www.cplusplus.com/reference/unordered_map/unordered_map/cend/).

cocos/2d/CCNode.cpp
```diff
@@ -1236,20 +1191,19 @@ AffineTransform Node::getNodeToParentAffineTransform() const
 
         // Build Transform Matrix
         // Adjusted transform calculation for rotational skew
-        kmScalar mat[] = { cy * _scaleX, sy * _scaleX,     0,  0,
-                        -sx * _scaleY, cx * _scaleY,    0,  0,
+        _transform = { cy * _scaleX, sy * _scaleX, 0,  0,
+                      -sx * _scaleY, cx * _scaleY, 0,  0,
```
dumganhar, line 1195: @Dhilan007
This breaks `windows` build. Windows doesn't support `initializer list`.

cocos/editor-support/cocostudio/CCArmature.cpp
```diff
@@ -385,13 +386,11 @@ Dictionary *Armature::getBoneDic() const
...
         // Build Transform Matrix
         // Adjusted transform calculation for rotational skew
-        kmScalar mat[] = { cy * _scaleX, sy * _scaleX,     0,  0,
+        _transform = { cy * _scaleX, sy * _scaleX,     0,  0,
                            -sx * _scaleY, cx * _scaleY,    0,  0,
                            0,  0,  1,  0,
                            x,  y,  0,  1 };
 
-        kmMat4Fill(&_transform, mat);
-
```
dumganhar, line 394: The same here.
If we don't want the stack objects being generated. Probably, we need to define some macros for the assignments.

cocos/2d/CCNode.cpp
ricardoquesada, line 577: My tests ( https://docs.google.com/spreadsheet/ccc?key=0AvvkdgVbWvpZdHFudzdDT3NuYTRNTHlZZzRGZWYzMmc&usp=drive_web#gid=0 ) don't say that `std::for_each` is faster than `for()`.

_iteration_ and _invocation_ are 2 different tests and the results should not be mixed.

In _iteration_ the slowest is: iterator with `begin() / end()` (which is basically what `std::for_each` does. And the fastest is `c++11 for()`
In _invocation_ `std::for_each` and the old macro have more or less the same performance. But if we can avoid using it, that is great because `std::for_each` creates another frame in the stack.

cocos/2d/CCNode.cpp
ricardoquesada, line 835: good catch! I'll fix it.

cocos/2d/CCNode.cpp
ricardoquesada, line 1195: well, yes, but in this particular case I'm using and old C feature that used to work before.
If this breaks the windows build I'll fix it. But I used this code before.

cocos/editor-support/cocostudio/CCArmature.cpp
ricardoquesada, line 394: For sure, we need a better Math library.

cocos/2d/CCNode.cpp
ricardoquesada, line 835: Fixed:
https://github.com/cocos2d/cocos2d-x/pull/4547

And I also converted some `std::for_each()` to `for()`.
Sorry it there was a misunderstanding with the performance tests.

minggo, line 835: We should test `std::for_each()` and `for()` without engine codes. And we should add this to our coding style.

cocos/2d/CCNode.cpp
dumganhar, line 577: Got it. It's my fault of reading the wrong line. :)
So the you didn't test `std::for_each`, right? If so, I will test it to see the performance of iterating through `std::for_each`.

ricardoquesada, line 577: I'll add a test right now.

dumganhar, line 577: OK, thank you. Could you please also paste the result on google doc?  :)

ricardoquesada, line 577: sure!

ricardoquesada, line 577: Results are here:
https://docs.google.com/spreadsheet/ccc?key=0AvvkdgVbWvpZdHFudzdDT3NuYTRNTHlZZzRGZWYzMmc&usp=drive_web#gid=0

Basically C++11 `for` and `std::for_each` have almost the same perf. Perhaps C++11 `for` is a tiny bit faster.

dumganhar, line 577: OK, I will submit a PR to replace `std::for_each` to `for(auto& )`.

dumganhar, line 577: Thanks a lot.

cocos/2d/CCNode.cpp
```diff
@@ -1236,29 +1191,32 @@ const kmMat4& Node::getNodeToParentTransform() const
...
             {
                 // XXX: Argh, kmMat needs a "translate" method
-                _transform.mat[12] += -_anchorPointInPoints.x;
-                _transform.mat[13] += -_anchorPointInPoints.y;
+                kmVec4 v = {-_anchorPointInPoints.x, -_anchorPointInPoints.y, 0, 1};
+                kmVec4Transform(&v, &v, &_transform);
+                _transform.mat[12] = v.x;
+                _transform.mat[13] = v.y;
+                _transform.mat[14] = v.z;
+                _transform.mat[15] = v.w;
```
dumganhar, line 1219: @ricardoquesada , why isn't here just a translation?  Did this modify the skew behavior?

ricardoquesada: could you give me more context? I don't remember that PR very much. thanks.

dumganhar: @ricardoquesada ,Sure, I found that the behavior of skew in 2.x & 3.x is different.
In 2.x, setSkewX(30) + setRotation(30) will show:
2 x

In 3.x, setSkewX(30) + setRotation(30) will show:
3 x

TEST CODE:

2.x:

``` c++
    {
        auto sprite = CCSprite::create("fonts/grossini.png");
        sprite->setPosition(ccp(visibleSize.width/2 + origin.x, visibleSize.height/2 + origin.y));
        addChild(sprite, 8888);
    }

    {
        auto sprite = CCSprite::create("fonts/grossini.png");
        sprite->setPosition(ccp(visibleSize.width/2 + origin.x, visibleSize.height/2 + origin.y));

        sprite->setSkewX(30);// SET SKEWX
        sprite->setRotation(30);// SET ROTATION

        addChild(sprite, 99999);
    }

    {
        CCDrawNode* d = CCDrawNode::create();

        d->drawSegment(ccp(origin.x, origin.y + visibleSize.height/2), ccp(origin.x + visibleSize.width, origin.y + visibleSize.height/2), 0.5f, ccc4f(1.0f, 0, 0, 1.0f));
        d->drawSegment(ccp(origin.x + visibleSize.width/2, origin.y), ccp(origin.x + visibleSize.width/2, origin.y + visibleSize.height), 0.5f, ccc4f(1.0f, 0, 0, 1.0f));

        addChild(d);
    }
```

3.x:

``` c++
    {
        auto sprite = Sprite::create("fonts/grossini.png");
        sprite->setPosition(Vec2(visibleSize / 2) + origin);

        addChild(sprite, 8888);
    }

    {
        auto sprite = Sprite::create("fonts/grossini.png");
        sprite->setPosition(Vec2(visibleSize / 2) + origin);

        sprite->setSkewX(30); // SET SKEWX 
        sprite->setRotation(30); // SET ROTATION

        addChild(sprite, 99999);
    }

    {
        auto d = DrawNode::create();
        d->drawLine(Vec2(origin.x, origin.y + visibleSize.height/2), Vec2(origin.x + visibleSize.width, origin.y + visibleSize.height/2), Color4F::RED);
        d->drawLine(Vec2(origin.x + visibleSize.width/2, origin.y), Vec2(origin.x + visibleSize.width/2, origin.y + visibleSize.height), Color4F::RED);

        addChild(d);
    }
```

I don't know which one should be the correct one. At glance, I figure that the behavior of `setSkewX` + `setRotation` is correct in 2.x but wrong in 3.x.

In addition, cocos2d-html5 seems to use the behavior in 2.x, so some `skew + rotation` animations will be broken in 3.x while they work fine in cocos2d-x 2.x & cocos2d-html5.

dumganhar: @ricardoquesada , I fixed it at https://github.com/cocos2d/cocos2d-x/pull/16736
Please review that PR. :)

ricardoquesada: @dumganhar done. thanks.

PR #4545, opened 2013-12-19
new spine runtime applied

seobyeongky:
hi
This branch has [new spine runtime version](https://github.com/EsotericSoftware/spine-runtimes). there are prefix "sp" in this new version. I tested in my Mac and IOS and Android. But not Windows and Linux system. It supports new skeleton format(bounding property).


dumganhar:
Thanks. I will test it soon.

dumganhar:
Merged. Thanks.

PR #4547, opened 2013-12-19
Node fixes with the iterator (thanks James)

ricardoquesada:
And replaces some `std::for_each()` with `for()` to prevent another object in the stack.


dumganhar:
:) Actually, I like `c++11 range loop` more than `std::for_each` since `std::for_each` will need a lambda which also need to capture variables by `[&], [=], &xx]`. That isn't really good for `loop`. If their performance are nearly the same.
I vote for `c++11 range loop` too. :)

minggo:
Yep.
lambda will consume more resources(it should remember the outer environment), and it is more unreadable.

ricardoquesada:
yep +1 for C+11 `for` is the performance is equal or better for `std::for_each`... (I would be surprised if it is not).

PR #4556, opened 2013-12-19
Fix BlendID Generation, also convert depth to the RenderCommandID

darkdukey:


cocos/2d/renderer/CCRenderCommand.cpp
```diff
@@ -37,13 +37,24 @@
...
+{
+    _viewport = viewport;
+    
+    //convert depth from float to unsigned int
+    //currently we only support 2 digits after decimal point
+    _depth = depth * 100 + CC_RENDERER_DEPTH_HALF;
+    _depth = MIN(_depth, CC_RENDERER_DEPTH_MAX);
+    _depth = MAX(_depth, CC_RENDERER_DEPTH_MIN);
+}
+
```
ricardoquesada, line 72: This document explains how to convert a depth that is expressed in a `float` into a 24-bit integer:
http://www.sjbaker.org/steve/omniv/love_your_z_buffer.html

darkdukey, line 72: Thanks riq, I will change it and calcuate z depth properly.

darkdukey, line 72: I did some test and find out this method only works for perspective camera, with orthogonal we will get divide by zero error((1<<N) \* ( a + b / z )). since the distance to camera could be 0 (the z). Maybe we should separate the different camera type as part of the RenderCommandID, then use different ways to generate depth.

ricardoquesada, line 72: Yes, perhaps we will need to ways to generate the Z.
If the user is using an orthogonal projection, then the math involved should be less  complex (and faster).

PR #4562, opened 2013-12-19
Removed CC_SAFE_DELETE of the static CCUserDefault object in the UserDefault destructor

dmurtagh:
- The existing code caused the UserDefault destructor to call itself
  recursively an infinte number of times causing stack overflow.
- Test case is to call the following from anywhere in your code:
  CCUserDefault::purgeSharedUserDefault();
- Only applies to iOS. Tested on iOS.


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/4571.

dumganhar:
Thanks.

PR #4564, opened 2013-12-19
Fixed non-functioning default value for CCUserDefault::getDoubleForKey()

dmurtagh:
- defaultValue was not being passed into the JNI CallStaticDoubleMethod
  method previously.
-  CCUserDefault::getDoubleForKey() would return undefined garbage when
  the key doesn’t exist.
  Test Case:
  Call the following from anywhere in your code (in release
  configuration). X will be garbage.
  Call the following from your code:
  double x =
  CCUserDefault::sharedUserDefault()->getDoubleForKey("Some-Non-existant-k
  ey-1234", 0);
  CCLOG("x = %4.2f", x);

Fix Tested on Android. Affects Android only.


dumganhar:
Could you also send a PR to master branch? Thanks.

PR #4565, opened 2013-12-19
replaces std::for_each with for()

ricardoquesada:
replaces `std::for_each` with C++11 `for( : )`


dumganhar:
Thanks, Riq.:)

ricardoquesada:
no problem.
If you search for `std::for_each` there are a few places that I didn't change... (two or three more places).
If you could replace them with `for()` it would be great. thanks.

dumganhar:
OK, I will replace it today.
- James

ricardoquesada:
thanks

PR #4572, opened 2013-12-20
spine simple JS binding supported

seobyeongky:
hi
This branch has simple Spine JS binding. I tested on iOS. But no test on other platforms.


dumganhar:
Could you paste the test codes?

minggo:
It is better to add test case for it.

seobyeongky:
OK I wil commit test case soon

dumganhar:
@seobyeongky 
Thanks. Merged at #4611

PR #4573, opened 2013-12-20
http client features

colprog:
support concurrent http request.
support direct download to file.
support std::function style callback.
add progress callback.
update libcurl to support https using darwinssl on iOS.


dumganhar:
Thanks. Could you add test case for `add progress callback` in TestCpp?

colprog:
@dumganhar test added

PR #4576, opened 2013-12-20
math lib,transform 

houguohua:
upload our mathlib,transform

### DIFF
--- a/cocos/base/C3DTransform.cpp
+++ b/cocos/base/C3DTransform.cpp
@@ -1,0 +1,683 @@
+#include "C3DMath.h"
+#include "C3DTransform.h"
+#include "cocos2d.h"
+
+namespace cocos3d
+{
+
+C3DTransform::C3DTransform()
+    : _matrixDirtyBits(0), _listeners(NULL)
+{
+
+    _scale.set(C3DVector3::one());
+}
+
+C3DTransform::C3DTransform(const C3DVector3& scale, const C3DQuaternion& rotation, const C3DVector3& translation)
+    : _matrixDirtyBits(0), _listeners(NULL)
+{
+   
+    set(scale, rotation, translation);
+}
+
+C3DTransform::C3DTransform(const C3DVector3& scale, const C3DMatrix& rotation, const C3DVector3& translation)
+    : _matrixDirtyBits(0), _listeners(NULL)
+{
+   
+    set(scale, rotation, translation);
+}
+
+C3DTransform::C3DTransform(const C3DTransform& copy)
+    : _matrixDirtyBits(0), _listeners(NULL)
+{
+   
+    set(copy);
+}
+
+C3DTransform::~C3DTransform()
+{
+    CC_SAFE_DELETE(_listeners);
+}
+
+const C3DMatrix& C3DTransform::getMatrix() const
+{
+    if (_matrixDirtyBits)
+    {
+        bool hasTranslation = !_position.isZero();
+        bool hasScale = !_scale.isOne();
+        bool hasRotation = !_rotation.isIdentity();
+
+        // Compose the matrix in TRS order since we use column-major matrices with column vectors and
+        // multiply M*v (as opposed to XNA and DirectX that use row-major matrices with row vectors and multiply v*M).
+        if (hasTranslation || (_matrixDirtyBits & DIRTY_TRANSLATION) == DIRTY_TRANSLATION)
+        {
+            C3DMatrix::createTranslation(_position, &_matrix);
+            if (hasRotation || (_matrixDirtyBits & DIRTY_ROTATION) == DIRTY_ROTATION)
+            {
+                _matrix.rotate(_rotation);
+            }
+            if (hasScale || (_matrixDirtyBits & DIRTY_SCALE) == DIRTY_SCALE)
+            {
+                _matrix.scale(_scale);
+            }
+        }
+        else if (hasRotation || (_matrixDirtyBits & DIRTY_ROTATION) == DIRTY_ROTATION)
+        {
+            C3DMatrix::createRotation(_rotation, &_matrix);
+            if (hasScale || (_matrixDirtyBits & DIRTY_SCALE) == DIRTY_SCALE)
+            {
+                _matrix.scale(_scale);
+            }
+        }
+        else if (hasScale || (_matrixDirtyBits & DIRTY_SCALE) == DIRTY_SCALE)
+        {
+            C3DMatrix::createScale(_scale, &_matrix);
+        }
+
+        _matrixDirtyBits = 0;
+    }
+
+    return _matrix;
+}
+
+const C3DVector3& C3DTransform::getScale() const
+{
+    return _scale;
+}
+
+void C3DTransform::getScale(C3DVector3* scale) const
+{
+    scale->set(_scale);
+}
+
+float C3DTransform::getScaleX() const
+{
+    return _scale.x;
+}
+
+float C3DTransform::getScaleY() const
+{
+    return _scale.y;
+}
+
+float C3DTransform::getScaleZ() const
+{
+    return _scale.z;
+}
+
+const C3DQuaternion& C3DTransform::getRotation() const
+{
+    return _rotation;
+}
+
+void C3DTransform::getRotation(C3DQuaternion* rotation) const
+{
+    assert(rotation);
+
+    rotation->set(_rotation);
+}
+
+void C3DTransform::getRotation(C3DMatrix* rotation) const
+{
+    assert(rotation);
+
+    C3DMatrix::createRotation(_rotation, rotation);
+}
+
+float C3DTransform::getRotation(C3DVector3* axis) const
+{
+    assert(axis);
+    return _rotation.toAxisAngle(axis);
+}
+
+const C3DVector3& C3DTransform::getPosition() const
+{
+    return _position;
+}
+
+void C3DTransform::getPosition(C3DVector3* translation) const
+{
+    translation->set(_position);
+}
+
+float C3DTransform::getPositionX() const
+{
+    return _position.x;
+}
+
+float C3DTransform::getPositionY() const
+{
+    return _position.y;
+}
+
+float C3DTransform::getPositionZ() const
+{
+    return _position.z;
+}
+
+C3DVector3 C3DTransform::getForwardVector() const
+{
+    C3DVector3 v;
+    getForwardVector(&v);
+    return v;
+}
+
+void C3DTransform::getForwardVector(C3DVector3* dst) const
+{
+    getMatrix().getForwardVector(dst);
+}
+
+
+C3DVector3 C3DTransform::getUpVector() const
+{
+    C3DVector3 v;
+    getUpVector(&v);
+    return v;
+}
+
+void C3DTransform::getUpVector(C3DVector3* dst) const
+{
+    getMatrix().getUpVector(dst);
+}
+
+
+C3DVector3 C3DTransform::getRightVector() const
+{
+    C3DVector3 v;
+    getRightVector(&v);
+    return v;
+}
+
+void C3DTransform::getRightVector(C3DVector3* dst) const
+{
+    getMatrix().getRightVector(dst);
+}
+
+void C3DTransform::rotate(float qx, float qy, float qz, float qw)
+{
+    C3DQuaternion q(qx, qy, qz, qw);
+    _rotation.multiply(q);
+    dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::rotate(const C3DQuaternion& rotation)
+{
+    _rotation.multiply(rotation);
+    dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::rotate(const C3DVector3& axis, float angle)
+{
+    C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromAxisAngle(axis, angle, &rotationQuat);
+    _rotation.multiply(rotationQuat);
+    _rotation.normalize();
+    dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::rotate(const C3DMatrix& rotation)
+{
+    C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromRotationMatrix(rotation, &rotationQuat);
+    _rotation.multiply(rotationQuat);
+    dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::rotateX(float angle)
+{
+    C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromAxisAngle(C3DVector3::unitX(), angle, &rotationQuat);
+    _rotation.multiply(rotationQuat);
+    dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::rotateY(float angle)
+{
+    C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromAxisAngle(C3DVector3::unitY(), angle, &rotationQuat);
+    _rotation.multiply(rotationQuat);
+    dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::rotateZ(float angle)
+{
+    C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromAxisAngle(C3DVector3::unitZ(), angle, &rotationQuat);
+    _rotation.multiply(rotationQuat);
+    dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::setRotationX(float angle)
+{	
+	C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromAxisAngle(C3DVector3::unitX(), angle, &rotationQuat);
+	_rotation = rotationQuat;
+	dirty(DIRTY_ROTATION);
+
+}
+void C3DTransform::setRotationY(float angle)
+{
+	C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromAxisAngle(C3DVector3::unitY(), angle, &rotationQuat);
+	_rotation = rotationQuat;
+	dirty(DIRTY_ROTATION);
+}
+
+
+void C3DTransform::setRotationZ(float angle)
+{
+	C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromAxisAngle(C3DVector3::unitZ(), angle, &rotationQuat);
+	_rotation = rotationQuat;
+	dirty(DIRTY_ROTATION);
+}
+
+void C3DTransform::scale(float scale)
+{
+    _scale.scale(scale);
+    dirty(DIRTY_SCALE);
+}
+
+void C3DTransform::scale(float sx, float sy, float sz)
+{
+    _scale.x *= sx;
+    _scale.y *= sy;
+    _scale.z *= sz;
+    dirty(DIRTY_SCALE);
+}
+
+void C3DTransform::scale(const C3DVector3& scale)
+{
+    _scale.x *= scale.x;
+    _scale.y *= scale.y;
+    _scale.z *= scale.z;
+    dirty(DIRTY_SCALE);
+}
+
+void C3DTransform::scaleX(float sx)
+{
+    _scale.x *= sx;
+    dirty(DIRTY_SCALE);
+}
+
+void C3DTransform::scaleY(float sy)
+{
+    _scale.y *= sy;
+    dirty(DIRTY_SCALE);
+}
+
+void C3DTransform::scaleZ(float sz)
+{
+    _scale.z *= sz;
+    dirty(DIRTY_SCALE);
+}
+
+void C3DTransform::set(const C3DVector3& scale, const C3DQuaternion& rotation, const C3DVector3& translation)
+{
+    _scale.set(scale);
+    _rotation.set(rotation);
+    _position.set(translation);
+    dirty(DIRTY_TRANSLATION | DIRTY_ROTATION | DIRTY_SCALE);
+}
+
+void C3DTransform::set(const C3DVector3& scale, const C3DMatrix& rotation, const C3DVector3& translation)
+{
+    _scale.set(scale);
+    C3DQuaternion rotationQuat;
+    C3DQuaternion::createFromRotationMatrix(rotation, &rotationQuat);
+    _rotation.set(rotationQuat);
+    _position.set(translation);
+    dirty(DIRTY_TRANSLATION | DIRTY_ROTATION | DIRTY_SCALE);
+}
+
+void C3DTransform::set(const C3DVector3& scale, const C3DVector3& axis, float angle, const C3DVector3& translation)
+{
+    _scale.set(scale);
+    _rotation.set(axis, angle);
+    _position.set(translation);
+    dirty(DIRTY_TRANSLATION | DIRTY_ROTATION | DIRTY_SCALE);
+}
+
+void C3DTransform::set(const C3DTransform& transform)
+{
+    _scale.set(transform._scale);
+    _rotation.set(transform._rotation);
+    _position.set(transform._position);
+    dirty(DIRTY_TRANSLATION | DIRTY_ROTATION | DIRTY_SCALE);
+}
+
+void C3DTransform::setIdentity()
+{
+    _scale.set(1.0f, 1.0f, 1.0f);
+    _rotation.setIdentity();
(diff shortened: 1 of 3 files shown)


cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
+#include "C3DMath.h"
+#include "C3DTransform.h"
+#include "cocos2d.h"
+
+namespace cocos3d
+{
+
+C3DTransform::C3DTransform()
+    : _matrixDirtyBits(0), _listeners(NULL)
```
minggo, line 9: NULL -> nullptr. Now we enable c++11.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
+#include "C3DMath.h"
```
minggo, line 1: We should add license of cocos2d-x. Of course, can modify it later.

cocos/base/C3DTransform.h
```diff
@@ -0,0 +1,728 @@
...
+ * transform's corresponding scale, rotation, or translation
+ * component is updated (it is not as if the scale, rotate, or translate
+ * is applied to the transform's matrix).
+ *
+ * Note: To construct a C3DTransform from a transformation matrix stored as a C3DMatrix,
+ * first decompose the C3DMatrix into its separate translation, scale, and rotation
+ * components using matrix.decompose(C3DVector3, C3DQuaternion, C3DVector3) and then pass
+ * those arguments to the appropriate constructor or set methods of C3DTransform.
+ */
+class  C3DTransform
```
minggo, line 26: I think we should remove `C3D` prefix. Just use `Transform`.
Of course, we can modify it later.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+#include "C3DTransform.h"
+#include "cocos2d.h"
+
+namespace cocos3d
```
minggo, line 5: I am not sure if we should use another namespace other than `cocos2d` in base folder.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+}
+
+const C3DVector3& C3DTransform::getScale() const
+{
+    return _scale;
+}
+
+void C3DTransform::getScale(C3DVector3* scale) const
+{
+    scale->set(_scale);
```
boyu0, line 89: get? set?

cocos/base/C3DTransform.h
```diff
@@ -0,0 +1,728 @@
...
+         * An optional long value that is specified to the Listener's callback.
+         */
+        long cookie;
+    };
+
+    /**
+     * Defines the matrix dirty bits for marking the translation, scale and rotation
+     * components of the C3DTransform.
+     */
+    enum MatrixDirtyBits
```
minggo, line 672: We can use `enum class` now.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+const C3DQuaternion& C3DTransform::getRotation() const
+{
+    return _rotation;
+}
+
+void C3DTransform::getRotation(C3DQuaternion* rotation) const
+{
+    assert(rotation);
+
+    rotation->set(_rotation);
```
boyu0, line 116: get? set?

cocos/base/C3DTransform.h
```diff
@@ -0,0 +1,728 @@
...
+     * Marks this transform as dirty and fires transformChanged().
+     */
+    void dirty(char matrixDirtyBits);
+
+    /**
+     * Called when the transform changes.
+     */
+    virtual void transformChanged();
+
+	virtual void copyFrom(const C3DTransform* other);
```
minggo, line 689: Wrong indent.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+}
+
+const C3DVector3& C3DTransform::getPosition() const
+{
+    return _position;
+}
+
+void C3DTransform::getPosition(C3DVector3* translation) const
+{
+    translation->set(_position);
```
boyu0, line 139: get? set? translation? _position?

cocos/math/c3d/C3DMathUtilty.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+
+C3DMathUtility::C3DMathUtility(): _costab(NULL), _sintab(NULL)
+{
+
+}
+
+// approximate value of sin(rad), precision 1 degree
+float C3DMathUtility::sin(float rad, bool bUseTable)
```
minggo, line 34: Please don't use hungarian style now.

cocos/math/c3d/C3DMathUtilty.cpp
```diff
@@ -0,0 +1,118 @@
...
+    _sintab = new float[count];
+    
+    for (int i = 0; i < count; i++) {
+        float rad = MATH_DEG_TO_RAD(i);
+        _sintab[i] = sinf(rad);
+        _costab[i] = cosf(rad);
+    }
+}
+
+}
```
minggo, line 118: It is better to add a space line at the end of a cpp file.

cocos/math/c3d/C3DMatrix.cpp
```diff
@@ -0,0 +1,1179 @@
+#include "C3DMath.h"
+
+#include "C3Dneon_matrix_impl.h"
```
minggo, line 3: Please don't use `_` to connect words to obey [cocos2d-x coding style](http://www.cocos2d-x.org/wiki/Cocos2d_C++_Coding_Style).

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+    l.listener = listener;
+    l.cookie = cookie;
+    _listeners->push_back(l);
+}
+
+void C3DTransform::removeListener(C3DTransform::Listener* listener)
+{
+    if (_listeners)
+    {
+        for (std::list<TransformListener>::iterator itr = _listeners->begin(); itr != _listeners->end(); itr++)
```
boyu0, line 581: for(auto listener : _listeners) is better.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+    matRotate.m[6] = yaxis.z;
+    matRotate.m[7] = 0;
+
+    matRotate.m[8] = zaxis.x;
+    matRotate.m[9] = zaxis.y;
+    matRotate.m[10] = zaxis.z;
+    matRotate.m[11] = 0;
+
+    setRotation(matRotate);
+	setPosition(position);
```
boyu0, line 641: wrong indent.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+void C3DTransform::copyFrom(const C3DTransform* other)
+{
+	_scale = other->_scale;
+	_position = other->_position;
+	_rotation = other->_rotation;
+	_matrix = other->_matrix;
+	_matrixDirtyBits = other->_matrixDirtyBits;
+
+	transformChanged();
```
boyu0, line 652: wrong indent

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+}
+
+void C3DTransform::setRotationAlong(const C3DVector3& point, const C3DVector3& axis, float angle)
+{
+    // p' = rot * (p - point) + point = rot * p - rot * point + point
+	_rotation.set(axis, angle);
```
boyu0, line 658: wrong indent

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+    _position = rot * (_position - point) + point; 
+    dirty(DIRTY_TRANSLATION | DIRTY_ROTATION);
+}
+
+//void C3DTransform::setRotation(const C3DVector3& dir, const C3DVector3& up)
+//{
+//	C3DMatrix matRotate;
+//	C3DMatrix::createLookAt(C3DVector3(0, 0, 0), -dir, up, &matRotate);
+//	setRotation(matRotate);
+//}
```
boyu0, line 681: if doesn't needed, delete it.

cocos/math/c3d/C3DMatrix.cpp
```diff
@@ -0,0 +1,1179 @@
...
+    float t_b = 1.0f / (top - bottom);
+    float f_n = 1.0f / (zFar - zNear);
+
+    memset(dst, 0, MATRIX_SIZE);
+    dst->m[0] = 2.0f * zNear * r_l;
+    dst->m[5] = 2.0f * zNear * t_b;
+	dst->m[8] = (right + left) * r_l;
+	dst->m[9] = (top + bottom) * t_b;
+	dst->m[10] = (-(zFar + zNear)) * f_n;
+	dst->m[11] = -1.0f;
```
boyu0, line 240: wrong indent.

cocos/math/c3d/C3DMatrix.cpp
```diff
@@ -0,0 +1,1179 @@
...
+
+void C3DMatrix::getForwardVector(C3DVector3* dst) const
+{
+    assert(dst);
+
+    //dst->x = -m[4];
+    //dst->y = -m[5];
+    //dst->z = -m[6];
+
+	dst->x = -m[8];
```
boyu0, line 686: lots of wrong indent, please fix them.

cocos/math/c3d/C3DMatrix.h
```diff
@@ -0,0 +1,988 @@
...
+     * Creates a view matrix based on the specified input parameters.
+     *
+     * @param eyePosition The eye position.
+     * @param targetPosition The target's cener position.
+     * @param up The up vector.
+     * @param dst A matrix to store the result in.
+     */
+    static void createLookAt(const C3DVector3& eyePosition, const C3DVector3& targetPosition, const C3DVector3& up, C3DMatrix* dst);
+
+	// ͨ���ᴴ������ lvlong
```
boyu0, line 143: Garbled text. please fix it.

cocos/base/C3DTransform.cpp
minggo, line 89: `scale` is a pointer, invoke `scale->set()` will change internal content of it.

cocos/base/C3DTransform.cpp
minggo, line 581: Yep, can use range-base loop.

minggo, line 581: And it is better to use as

```
for (auto &listener : _listeners)
```

cocos/base/C3DTransform.cpp
dabingnn, line 116: It seems the "const C3DQuaternion& C3DTransform::getRotation() const" is enough. the second one is likely to cause some misunderstanding, do you think so?

cocos/base/C3DTransform.cpp
dabingnn, line 139: Same as rotation&scale&translation

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+            if ((*itr).listener == listener)
+            {
+                _listeners->erase(itr);
+                break;
+            }
+        }
+    }
+}
+
+void C3DTransform::transformChanged()
```
dabingnn, line 592: It seems better not put those listeners in the basic transform class. Eg, if C3dNode contains a C3DTransform, it is better to add listeners in C3DNode instead of C3DTransform. How do you think about this?

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+void C3DTransform::setRotationAlong(const C3DVector3& point, const C3DVector3& axis, float angle)
+{
+    // p' = rot * (p - point) + point = rot * p - rot * point + point
+	_rotation.set(axis, angle);
+    _position = point - _rotation * point;
+    dirty(DIRTY_TRANSLATION | DIRTY_ROTATION);
+}
+
+void C3DTransform::rotateAlong(const C3DVector3& point, const C3DVector3& axis, float angle)
```
dabingnn, line 663: If the C3DTransform contains _scale, how to explain the effects of this function? It is a bit confused.

cocos/base/C3DTransform.h
```diff
@@ -0,0 +1,728 @@
...
+    std::list<TransformListener>* _listeners;
+
+private:
+
+
+};
+
+}
+
+#endif
```
dabingnn, line 728: Seems the C3DTransform is a bit fat. It contains translation, scale, rotation logic, as well as lookat logic and rotation along line logic. Can we place these different features into individual class?

cocos/math/c3d/C3DMatrix.cpp
```diff
@@ -0,0 +1,1179 @@
...
+    dst->m[2] = crsp * cy + sr * sy;
+    dst->m[6] = crsp * sy - sr * cy;
+    dst->m[10] = cr * cp;
+
+    dst->m[3] = dst->m[7] = dst->m[11] = 0.0;
+    dst->m[15] = 1.0;
+
+}
+
+void C3DMatrix::createLookAt(const C3DVector3& eyePosition, const C3DVector3& targetPosition, const C3DVector3& up, C3DMatrix* dst)
```
dabingnn, line 125: We has C3DMatrix::createLookAt & C3DTransform::lookAt, any differences between them?

cocos/base/C3DTransform.cpp
dumganhar, line 581: We should use `const` keyword when compiler doesn't complain.

```
for(const auto &listener : _listeners)
```

cocos/math/c3d/C3DMatrix.cpp
```diff
@@ -0,0 +1,1179 @@
...
+    inverse.m[15] = m[8] * a3 - m[9] * a1 + m[10] * a0;
+
+    multiply(inverse, 1.0f / det, dst);
+
+    return true;
+}
+
+bool C3DMatrix::isIdentity() const
+{
+    return (memcmp(m, MATRIX_IDENTITY, MATRIX_SIZE) == 0);
```
dabingnn, line 748: Be careful of float precision, it may always return false.

cocos/base/C3DTransform.cpp
dumganhar, line 581: @minggo @boyu0 
The next lines use 

```
_listeners->erase(itr);
```

for (const auto& ) range loop can't support this.

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+            }
+        }
+    }
+}
+
+void C3DTransform::transformChanged()
+{
+    if (_listeners)
+    {
+        for (std::list<TransformListener>::iterator itr = _listeners->begin(); itr != _listeners->end(); itr++)
```
dumganhar, line 596: Here we could use the fastest `Range loop` and use `auto` keyword to simply codes.
`for (const auto &l : _listeners)`

cocos/base/C3DTransform.cpp
```diff
@@ -0,0 +1,683 @@
...
+        {
+            TransformListener& l = *itr;
+            l.listener->transformChanged(this, l.cookie);
```
dumganhar, line 599: It seems `transformChanged`  is a callback function. So adding `on` prefix will make our codes more realizable.
`onTransformChanged`.
If the listener could be `std::function` will be great.

cocos/base/C3DTransform.h
dumganhar, line 672: @minggo 
`enum class` could not use bit operation like `|`, `&`, etc.

cocos/base/C3DTransform.cpp
minggo, line 581: Yep. So ignore this one.

cocos/base/C3DTransform.h
minggo, line 672: You are right. enum class is not convenient if it needs these operation. Or should override these operation.

houguohua, line 672: Thank all for your advice.Let's modify the unreasonable code in next week.

---

发件人: minggo notifications@github.com
发送时间: 2013年12月20日 21:15
收件人: cocos2d/cocos2d-x
抄送: 侯国华
主题: Re: [cocos2d-x] math lib,transform (#4576)

In cocos/base/C3DTransform.h:

> -        /**
> -         \* Listener for C3DTransform events.
> -         */
> -        Listener\* listener;
>   +
> -        /**
> -         \* An optional long value that is specified to the Listener's callback.
> -         */
> -        long cookie;
> -    };
>   +
> -    /**
> -     \* Defines the matrix dirty bits for marking the translation, scale and rotation
> -     \* components of the C3DTransform.
> -     */
> -    enum MatrixDirtyBits

You are right. enum class is not convenient if it needs these operation. Or should override these operation.

ricardoquesada:
Thanks for the PR.

As @minggo mentioned, it is important to understand the in cocos2d v3.0, there will be no difference between 2d objects and 3d objects.
So it is important to remove the C3D prefix from the class names and from the file names.
Also, it is important to keep using the `cocos2d` namespace (not the `cocos3d` namespace).

And the math code should be under `cocos/math/`, not `cocos/math/c3d`.

Anyway, the `Transform` and the math library are great starting points.
After merging this PR, we need to integrate `Transform` into `Node`. I guess it should be difficult (specially since we are already using mat4x4 matrix in node).

Regarding operator overloading in the math library, please read this:
https://github.com/cocos2d/cocos2d-x/blob/develop/docs/CODING_STYLE.md#operator-overloading

So, in our code (cocos2d), I wouldn't use any `operator` operations for matrix / vertex.
In our code we should use the "named" version since it is easier search, easier to understand PR (diffs), and easier to find performance issues.

``` c++
auto r = m1 * m2; // looks cheap... it is just a multiplication

auto r = m1.multiply(m2);  // doesn't look cheap. We shouldn't abuse calling 'multiply'
```

Regarding the math API, I prefer the following:

``` c++
// output is returned
auto result = m1.multiply( m2 );  // <-- I prefer this API  (glm and GLKMatrix use this kind of API)

// This API requires more work for the user.
matrix output;   // user needs to declare the variable on another line
m1.multiply(m2, output);   // kazmath and our API use this kind of API
```

cocos/base/C3DTransform.cpp
houguohua, line 89: void C3DTransform::getScale(C3DVector3\* scale) const
{
    scale->set(_scale);
}
this function is ok.

cocos/base/C3DTransform.cpp
houguohua, line 592: ok,good idea

cocos/base/C3DTransform.cpp
houguohua, line 663: _scale remains unchange in this function

cocos/base/C3DTransform.cpp
houguohua, line 116: void Node::getPosition(float\* x, float\* y) const
{
    *x = _position.x;
    *y = _position.y;
}

This function is in CCNode  file

cocos/base/C3DTransform.cpp
houguohua, line 5: for math and transform,we should use one namespace ,for example "cocos"

cocos/base/C3DTransform.cpp
boyu0, line 89: I think this is make confused to who review or maintain this code, is there a better way to do this?

houguohua, line 89: void Node::getPosition(float\* x, float\* y) const
{
    *x = _position.x;
    *y = _position.y;
}
this function is in CCNode.

It have good performance,also ,It makes confused to you.

---

发件人: boyu0 notifications@github.com
发送时间: 2013年12月23日 14:51
收件人: cocos2d/cocos2d-x
抄送: 侯国华
主题: Re: [cocos2d-x] math lib,transform (#4576)

In cocos/base/C3DTransform.cpp:

> -        }
>   +
> -        _matrixDirtyBits = 0;
> -    }
>   +
> -    return _matrix;
>   +}
>   +
>   +const C3DVector3& C3DTransform::getScale() const
>   +{
> -    return _scale;
>   +}
>   +
>   +void C3DTransform::getScale(C3DVector3\* scale) const
>   +{
> -    scale->set(_scale);

I think this is make confused to who review or maintain this code, is there a better way to do this?

boyu0, line 89: No, I mean you use `set` in function `getXXX`, it's very weird. @minggo @dumganhar What's your opinion?

houguohua, line 89: what wrong?
set the transform's _scale  value into the scale which is a function parameter.
Is it not possible?

---

发件人: boyu0 notifications@github.com
发送时间: 2013年12月23日 15:38
收件人: cocos2d/cocos2d-x
抄送: 侯国华
主题: Re: [cocos2d-x] math lib,transform (#4576)

In cocos/base/C3DTransform.cpp:

> -        }
>   +
> -        _matrixDirtyBits = 0;
> -    }
>   +
> -    return _matrix;
>   +}
>   +
>   +const C3DVector3& C3DTransform::getScale() const
>   +{
> -    return _scale;
>   +}
>   +
>   +void C3DTransform::getScale(C3DVector3\* scale) const
>   +{
> -    scale->set(_scale);

No, I mean you use set in function getXXX, it's very weird. @minggohttps://github.com/minggo @dumganharhttps://github.com/dumganhar What's your opinion?

houguohua, line 89: Do you have a better way?
except that we delete this function?

---

发件人: 侯国华
发送时间: 2013年12月23日 15:48
收件人: cocos2d/cocos2d-x; cocos2d/cocos2d-x
主题: 答复: [cocos2d-x] math lib,transform (#4576)

what wrong?
set the transform's _scale  value into the scale which is a function parameter.
Is it not possible?

---

发件人: boyu0 notifications@github.com
发送时间: 2013年12月23日 15:38
收件人: cocos2d/cocos2d-x
抄送: 侯国华
主题: Re: [cocos2d-x] math lib,transform (#4576)

In cocos/base/C3DTransform.cpp:

> -        }
>   +
> -        _matrixDirtyBits = 0;
> -    }
>   +
> -    return _matrix;
>   +}
>   +
>   +const C3DVector3& C3DTransform::getScale() const
>   +{
> -    return _scale;
>   +}
>   +
>   +void C3DTransform::getScale(C3DVector3\* scale) const
>   +{
> -    scale->set(_scale);

No, I mean you use set in function getXXX, it's very weird. @minggohttps://github.com/minggo @dumganharhttps://github.com/dumganhar What's your opinion?

cocos/base/C3DTransform.cpp
boyu0, line 116: I'm agree with @dabingnn 
`getPosition(float* x, float* y) const` function returns x and y value with a copy, user can't change the value inside Node (Node::_position), but if you pass a pointer, it make no difference with `C3DQuaternion* getRotation()` or `C3DQuaternion& getRotation()`

cocos/base/C3DTransform.cpp
boyu0, line 89: I think if you want remain this function, `scale = _scale` is ok for me, I think this is what `getXXX` mean. And I think `C3DVector3& C3DTransform::getScale()` is better. what's your opinion?

houguohua, line 89: getXXX have two way:
1.function return

2.function parameters

Perhaps you are right.and ,we will dellete this function.

---

发件人: boyu0 notifications@github.com
发送时间: 2013年12月23日 16:23
收件人: cocos2d/cocos2d-x
抄送: 侯国华
主题: Re: [cocos2d-x] math lib,transform (#4576)

In cocos/base/C3DTransform.cpp:

> -        }
>   +
> -        _matrixDirtyBits = 0;
> -    }
>   +
> -    return _matrix;
>   +}
>   +
>   +const C3DVector3& C3DTransform::getScale() const
>   +{
> -    return _scale;
>   +}
>   +
>   +void C3DTransform::getScale(C3DVector3\* scale) const
>   +{
> -    scale->set(_scale);

I think if you want remain this function, scale = _scale is ok for me, I think this is what getXXX mean. And I think C3DVector3& C3DTransform::getScale() is better. what's your opinion?

PR #4585, opened 2013-12-21
Update CCBMemberVariableAssigner.h

jimmystar:
update macro definition, chang name of pMemberVariableName, pTarget and pNode, or this will make compile error in use of macros.

### DIFF
--- a/cocos/editor-support/cocosbuilder/CCBMemberVariableAssigner.h
+++ b/cocos/editor-support/cocosbuilder/CCBMemberVariableAssigner.h
@@ -6,9 +6,9 @@
 namespace cocosbuilder {
 
 #define CCB_MEMBERVARIABLEASSIGNER_GLUE(TARGET, MEMBERVARIABLENAME, MEMBERVARIABLETYPE, MEMBERVARIABLE) \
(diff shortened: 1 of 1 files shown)


dumganhar:
But the testcpp didn't show any compiler error without this patch. Could you describe more about it?

dumganhar:
@jimmystar , any updates?

dumganhar:
I would not apply this patch since it will be incompatible with old codes.
For instance: User has to change their codes :

``` c++
bool ButtonTestLayer::onAssignCCBMemberVariable(Object * pTarget, const char * pMemberVariableName, Node * pNode) {
    CCB_MEMBERVARIABLEASSIGNER_GLUE(this, "mCCControlEventLabel", LabelBMFont *, this->mControlEventLabel);

    return false;
}
```

to

``` c++
bool ButtonTestLayer::onAssignCCBMemberVariable(Object * target, const char * memberVariableName, Node * node) {
    CCB_MEMBERVARIABLEASSIGNER_GLUE(this, "mCCControlEventLabel", LabelBMFont *, this->mControlEventLabel);

    return false;
}
```

dumganhar:
Close this PR. Thanks.

PR #4596, opened 2013-12-23
issue #2050: fix tilemap firstgid bug

boyu0:
change unsigned int to int.
fix bug.


dumganhar:
This Pull Request couldn't be automatically merged.

cocos/2d/CCTMXLayer.cpp
```diff
@@ -178,16 +174,9 @@ void TMXLayer::setupTiles()
             if (gid != 0) 
             {
                 this->appendTileForGID(gid, Point(x, y));
-
-                // Optimization: update min and max GID rendered by the layer
-                _minGID = MIN(gid, _minGID);
-                _maxGID = MAX(gid, _maxGID);
```
dumganhar, line 184: Why to delete these lines? Any issue of it?

boyu0, line 184: Because there is only one place to use  this two variable in our engine:  
`CCASSERT( _maxGID >= _tileSet->_firstGid && _minGID >= _tileSet->_firstGid, "TMX: Only 1 tileset per layer is supported");`
and because of the gid some times do AND operation to some mask value, in that case, this two variable can't predict anything.
and I changed the gid to int, do AND operation always generate a negative number, that cause _minGID and _maxGID are wrong number, and may triger this assert sometimes.

minggo:
Why not merge it?

dumganhar:
@boyu0 
Did you add the test case for the this PR?

boyu0:
@dumganhar 
`autorelease` is called where  `createTileMalayer` function is called.
Maybe move `autorelease` call into `CLN` is better, or add `CREATE_FUNC` to each classes and use `CL` instead.
Which is better? What's your opinion?

dumganhar:
@boyu0 ,
Yep, I think this is better than before.

dumganhar:
Thanks.

PR #4620, opened 2013-12-24
support pass by value for setPosition\setAnchorPoint\setContentSize

WenhaiLin:


scripting/javascript/bindings/cocos2d_specifics.cpp
```diff
@@ -3418,6 +3386,123 @@ static JSBool js_cocos2dx_CCCamera_getEyeXYZ(JSContext *cx, uint32_t argc, jsval
     js_cocos2dx_CCCamera_getXYZ(getEyeXYZ)
 }
 
+#define js_cocos2dx_setAnchorPoint(className) \
+    jsval *argv = JS_ARGV(cx, vp);                                                   \
```
dumganhar: Using `template` will make our codes more elegant.

dumganhar:
Thanks.

PR #4625, opened 2013-12-24
Develop new clipping node

dabingnn:


cocos/2d/CCClippingNode.cpp
```diff
@@ -31,10 +31,17 @@
...
+#include "CCGroupCommand.h"
+#include "CCCustomCommand.h"
 
 NS_CC_BEGIN
 
 static GLint g_sStencilBits = -1;
+// store the current stencil layer (position in the stencil buffer),
+// this will allow nesting up to n ClippingNode,
+// where n is the number of bits of the stencil buffer.
+static GLint layer = -1;
```
dumganhar: Please change the global variable name. Name it to `s_layer` will be better.

cocos/2d/CCClippingNode.cpp
```diff
@@ -179,96 +200,110 @@ void ClippingNode::drawFullScreenQuadClearStencil()
...
     
+    GroupCommand* groupCommand = GroupCommand::getCommandPool().generateCommand();
+    groupCommand->init(0,_vertexZ);
+    renderer->addCommand(groupCommand);
+
+    renderer->pushGroup(groupCommand->getRenderQueueID());
+
+    CustomCommand* beforeVisitCmd = CustomCommand::getCommandPool().generateCommand();
+    beforeVisitCmd->init(0,_vertexZ);
+    beforeVisitCmd->func = CC_CALLBACK_0(ClippingNode::beforeVisit, this);
```
dumganhar: Why not to add prefix `on` like `onBeforeVisit`?

cocos/2d/CCClippingNode.cpp
```diff
@@ -303,20 +338,20 @@ void ClippingNode::visit()
...
+
     // enable alpha test only if the alpha threshold < 1,
     // indeed if alpha threshold == 1, every pixel will be drawn anyways
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC || CC_TARGET_PLATFORM == CC_PLATFORM_WINDOWS || CC_TARGET_PLATFORM == CC_PLATFORM_LINUX)
-    GLboolean currentAlphaTestEnabled = GL_FALSE;
-    GLenum currentAlphaTestFunc = GL_ALWAYS;
-    GLclampf currentAlphaTestRef = 1;
+//    GLboolean currentAlphaTestEnabled = GL_FALSE;
+//    GLenum currentAlphaTestFunc = GL_ALWAYS;
+//    GLclampf currentAlphaTestRef = 1;
```
dumganhar: These lines are not needed now?

dabingnn: The visit function has completely replaced by the NewClippingNode. These variables has become member variables. It is never needed. Maybe we need to remove those commented codes.

cocos/2d/CCClippingNode.cpp
dabingnn: sound better, I will fix it.

cocos/2d/CCClippingNode.cpp
dabingnn: OK

cocos/2d/CCClippingNode.cpp
dumganhar: Yep, if they aren't needed now, please remove them. Thanks.

PR #4631, opened 2013-12-24
Change const char* to const std::string&

boyu0:


minggo:
Please fix conflicts.

dumganhar:
Thanks

PR #4637, opened 2013-12-24
math lib,transform 

houguohua:
1.put math lib into math dieectory
2.remove C3D prefix from class name
3.remove cocos3d namespace

### DIFF
--- a/cocos/base/Transform.cpp
+++ b/cocos/base/Transform.cpp
@@ -80,10 +80,6 @@ const Vector3& Transform::getScale() const
     return _scale;
 }
 
-void Transform::getScale(Vector3* scale) const
-{
-    scale->set(_scale);
-}
 
 float Transform::getScaleX() const
 {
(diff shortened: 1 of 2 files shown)


cocos/base/Transform.cpp
```diff
@@ -0,0 +1,610 @@
...
+	_matrix = other->_matrix;
+	_matrixDirtyBits = other->_matrixDirtyBits;
+
+	transformChanged();
+}
+
+void Transform::setRotationAlong(const Vector3& point, const Vector3& axis, float angle)
+{
+    // p' = rot * (p - point) + point = rot * p - rot * point + point
+	_rotation.set(axis, angle);
```
dabingnn, line 591: a wrong indention here, could you please use space instead of tab for indention? It is more compatible.

cocos/math/Matrix.cpp
```diff
@@ -356,7 +356,15 @@ void C3DMatrix::createRotation(const C3DVector3& axis, float angle, C3DMatrix* d
...
     float c, s;
-    C3DMathUtility::getInstance().sincos(angle, &s, &c, !approximate);
+    if (approximate)
+    {
+        MathUtility::getInstance().sincos(angle, &s, &c);
+    }
+    else
+    {
+        c = cos(angle);
+        s = sin(angle);
```
dabingnn, line 366: We use sin, cos functions here. It seems the parameter are radians in MathUtility. However, the parameter here is angle. It may cause some confusions. Can we make it clearer? which sounds better.

PR #4640, opened 2013-12-25
issue #2920: temporary fix dictation input crash on ios7.

boyu0:


minggo:
Please add comment in codes to describe it.
May be add "XXX: Fix me ..." is better.

boyu0:
No problem.

minggo:
And it is better to add a issue number in comment.

extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
```diff
@@ -414,10 +414,10 @@ - (void) textChanged
             _systemControl.textField.keyboardType = UIKeyboardTypeDecimalPad;
             break;
         case EditBox::InputMode::SINGLE_LINE:
-            _systemControl.textField.keyboardType = UIKeyboardTypeDefault;
+            _systemControl.textField.keyboardType = UIKeyboardTypeEmailAddress;
```
dumganhar: Why does `SINGLE_LINE` mode use `UIKeyboardTypeEmailAddress`. It's weird.

boyu0: Yes, you are right. But this is only way I can find to fix this bug for now.

dumganhar: But how to input chinese if it's for email input?

boyu0: I have tested that, it seems it makes no difference with default type except have more button than default at the bottom of keyboard.

dumganhar: You mean that you can also input `chinese` letters in `email` mode?

PR #4654, opened 2013-12-25
support android single touch mode

pktangyue:
For cocos2dx 2.x version, Android is multi-touch supported by default.

But in some case, multi-touch would cause some unexpected problem, so I just want it only handle one touch at a time.

You can use `Cocos2dxGLSurfaceView.getInstance().setMultiTouchEnabled(false);` to disable multi touch mode.

### DIFF
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
@@ -184,6 +184,10 @@ public class Cocos2dxGLSurfaceView extends GLSurfaceView {
 		this.mIsMultiTouchEnabled = mIsMultiTouchEnabled;
 	}
 	
+	public boolean isMultiTouchEnabled() {
+		return this.mIsMultiTouchEnabled;
+	}
+	
 	@Override
 	public boolean onTouchEvent(final MotionEvent pMotionEvent) {
 		// these data are used in ACTION_MOVE and ACTION_CANCEL


dumganhar:
Is `android` providing an API to disable multi-touch?

pktangyue:
I think there is no simple way to disable multi-touch, just like [__glView setMultipleTouchEnabled:YES]; in iOs.

I find [Disabling multitouch in android in stackoverflow](http://stackoverflow.com/questions/12111862/disabling-multitouch-in-android), someone said 'you should turn off [splitMotionEvents](http://developer.android.com/reference/android/R.attr.html#splitMotionEvents) or [windowEnableSplitTouch](http://developer.android.com/reference/android/R.attr.html#windowEnableSplitTouch) attribute inside ViewGroup where buttons are located'.

But I don't think it will work in cocos2d-x, because there is only a GLSurfaceView for android, while these two options only prevent other view to receive touch event to make just only one touch take effect. ( I didn't run and check it. if I'm wrong, point it out)

In my changed code, ACTION_DOWN and ACTION_UP are the states for the first touch point, while ACTION_POINTER_DOWN and ACTION_POINTER_UP are the states for no-first touch points. So I break these two case if set multi-touch false.

---

After more test, I find a bug. And I will send new commit after I fixing it.

pktangyue:
I update a new commit, plz review it.

cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
```diff
@@ -56,6 +56,7 @@ of this software and associated documentation files (the "Software"), to deal
 	private Cocos2dxRenderer mCocos2dxRenderer;
 	private Cocos2dxEditText mCocos2dxEditText;
 
+	private boolean is_multi_touch_enabled = true;
```
dumganhar: Please use the variable naming style in java. `mXxxXxx`. Thanks.

PR #4660, opened 2013-12-25
Develop fix test case render texture

dabingnn:


xpol:
As commented here https://github.com/cocos2d/cocos2d-x/commit/20a8808b78563dd9d7290c6f6cd90b2399c5744b#commitcomment-14443628 .

getChar16VectorFromUTF16String would simpler if using iterator pair constructor:

``` C++
std::vector<char16_t> getChar16VectorFromUTF16String(const std::u16string& utf16)
{
    return std::vector<char16_t>(utf16.begin(), utf16.end());
}
```

PR #4666, opened 2013-12-26
change play interface name

2youyou2:
change play interface name

### DIFF
--- a/cocos/editor-support/cocostudio/CCArmatureAnimation.h
+++ b/cocos/editor-support/cocostudio/CCArmatureAnimation.h
@@ -127,7 +127,7 @@ public:
 
     /**
      * Play animation by index, the other param is the same to play.
-     * Deprecated, please use playWithIndex
+     * @deprecated, please use playWithIndex
      * @param  animationIndex  the animation index you want to play
      */
     CC_DEPRECATED_ATTRIBUTE virtual void playByIndex(int animationIndex,  int durationTo = -1, int loop = -1);


cocos/editor-support/cocostudio/CCArmatureAnimation.h
```diff
@@ -127,13 +127,14 @@ class  ArmatureAnimation : public ProcessBase
 
     /**
      * Play animation by index, the other param is the same to play.
+     * Deprecated, please use
```
minggo: Use what?

cocos/editor-support/cocostudio/CCArmatureAnimation.h
```diff
@@ -127,13 +127,14 @@ class  ArmatureAnimation : public ProcessBase
 
     /**
      * Play animation by index, the other param is the same to play.
+     * Deprecated, please use playWithIndex
```
dumganhar: Uses `@deprecated please use playWithIndex instead!`. 
Please use `doxygen` comments.

cocos/editor-support/cocostudio/CCArmatureDataManager.h
```diff
@@ -199,7 +199,7 @@ class  ArmatureDataManager : public cocos2d::Object
 
     bool _autoLoadSpriteFile;
 
-    std::map<std::string, RelativeData> _relativeDatas;
+    std::unordered_map<std::string, RelativeData> _relativeDatas;
```
dumganhar, line 202: Yep, the performance of `std::unordered_map` is higher than `std::map`. We need to use `c++11` features on develop branch.

PR #4682, opened 2013-12-26
Restricts the element type for Vector<T> and Map<K, V>, it has to be `Object*` or `ObjectSubClass*`.

dumganhar:


ricardoquesada:
I think this could be achieved with templates, and can be detected at compile time.

dumganhar:
@ricardoquesada 
Yep, you're right. I closed this PR, and sent another PR here(https://github.com/cocos2d/cocos2d-x/pull/4685/files)
Using c++11 `static_assert`. That work as you expected. :)

PR #4685, opened 2013-12-26
Restricts the element type for Vector<T> and Map<K, V>, it has to be `Object*` or `ObjectSubClass*`.

dumganhar:


cocos/base/CCMap.h
```diff
@@ -69,6 +70,7 @@ class CC_DLL Map
     explicit Map<K, V>(ssize_t capacity)
     : _data()
     {
+        static_assert(std::is_convertible<V, Object*>::value, "Invalid Type for cocos2d::Map<K, V>!");
```
ricardoquesada, line 73: nice :-)

PR #4688, opened 2013-12-26
make CC_USE_PHYSICS can ACTUALLY be switch off

walzer:
By the way, bring Scene::update(float), Scene::addChild(...) out of CC_USE_PHYSICS segment

### DIFF
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -88,26 +88,22 @@ Scene* Scene::getScene()
     return this;
 }
 
+#if CC_USE_PHYSICS
 void Scene::addChild(Node* child, int zOrder, int tag)
 {
     Node::addChild(child, zOrder, tag);
-#if CC_USE_PHYSICS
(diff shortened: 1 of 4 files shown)


cocos/2d/CCScene.cpp
```diff
@@ -88,7 +88,26 @@ Scene* Scene::getScene()
...
+void Scene::update(float delta)
+{
+    Node::update(delta);
+#if CC_USE_PHYSICS
+    if (nullptr != _physicsWorld)
+    {
+        _physicsWorld->update(delta);
+    }
+#endif
+}
```
boyu0, line 105: This is unnecessary，`update` and `addChild` are virtual functions in `Node`, I override these functions only for physics.

walzer, line 105: So your mistake is that, `Scene::update` and `addChild` haven't been wrapped by CC_USE_PHYSICS in CCScene.h

dumganhar:
Thanks. @walzer

PR #4690, opened 2013-12-26
Added missing CocoStudio Trigger code to WP8 and WinRT projects

stammen:
Added the missing CocoStudio Trigger code to WP8 and WinRT libextension projects to fix builds broken by the check in of Trigger code. Please note that the Trigger code added to the Extensions Tests crashes the Win32 and WP8 builds when MainMenu is clicked and the Test app is returning to the main menu scene. The author of the test code should probably test on Win32 and fix the bug.


minggo:
Thanks.

PR #4697, opened 2013-12-27
CCData bugfix

seobyeongky:
Hi
I fixed a Dangerous bug.

### DIFF
--- a/cocos/base/CCData.cpp
+++ b/cocos/base/CCData.cpp
@@ -39,13 +39,17 @@ _size(0)
     CCLOGINFO("In the empty constructor of Data.");
 }
 
-Data::Data(Data&& other)
+Data::Data(Data&& other) :
+_bytes(nullptr),
+_size(0)
 {
     CCLOGINFO("In the move constructor of Data.");
     move(other);
 }
 
-Data::Data(const Data& other)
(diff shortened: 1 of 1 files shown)


cocos/base/CCData.cpp
```diff
@@ -39,13 +39,17 @@
...
     move(other);
 }
 
-Data::Data(const Data& other)
+Data::Data(const Data& other) :
+_bytes(nullptr),
+_size(0)
 {
     CCLOGINFO("In the copy constructor of Data.");
     copy(other._bytes, other._size);
```
dumganhar, line 55: Why is it dangerous? 
In `Data::move` and `Data::copy`, the member variables will be assigned.

dumganhar, line 55: But I think your fix is all right.

seobyeongky, line 55: Ah, Because 'copy' calls 'clear'. and 'clear' calls 'free(_bytes)'. _byte has unexpected value. so dangerous.

dumganhar, line 55: Ouch, it's my fault. :) Thanks

seobyeongky, line 55: Welcome

PR #3509, opened 2013-08-30
the Cocos2dxETCLoader was removed?

alenl2:

PR #4699, opened 2013-12-27
closed #3509: Sprites with PhysicsBody moves to wrong position when swit...

boyu0:
...ch foreground/background.

skip one fps to avoid incorrect time calculation when resume.


dumganhar:
Thanks.

PR #4454, opened 2013-12-09
AngelCode BMFont binary format

SBKarr:
Implemented binary file format (http://www.angelcode.com/products/bmfont/doc/file_format.html)

Binary files is small and faster to parse, then text format. Implementation based on text fomat code (same sanity checks) and valid for version 3 of this format.

PR #4715, opened 2013-12-27
AngelCode binary file format support for LabelBMFont

SBKarr:
Format specification: http://www.angelcode.com/products/bmfont/doc/file_format.html
Pull request for master branch: https://github.com/cocos2d/cocos2d-x/pull/4454


dumganhar:
@SBKarr , Thanks a lot.
Could you also add a test in `LabelTest`?

cocos/2d/CCLabelBMFont.cpp
```diff
@@ -183,10 +183,16 @@ void CCBMFontConfiguration::purgeFontDefDictionary()
...
+
+	Data data = FileUtils::getInstance()->getDataFromFile(fullpath);
+    CCASSERT((!data.isNull() && data.getSize() > 0), "CCBMFontConfiguration::parseConfigFile | Open file error.");
+
+    if (memcmp("BMF", data.getBytes(), 3) == 0) {
+        std::set<unsigned int>* ret = parseBinaryConfigFile(data.getBytes(), data.getSize(), controlFile);
+        return ret;
+    }
+
+    std::string contents((const char*)data.getBytes());
```
dumganhar: I think here should be 

```
std::string contents((const char*)data.getBytes(), data.getSize());
```

Since the data is not \0 terminal string.

dumganhar:
Merged!. Thanks. 
BTW, master branch is only for updating `CocoStudio Runtime` and `Critical Bug fix`. So do you mind to close that PR?

PR #4720, opened 2013-12-30
openal context not destroy correctly on mac and ios

zarelaky:


dumganhar:
How did you find this issue? Does it cause some problems after exit application?

zarelaky:
this can cause others module in the same process failed when create openal context.
on another hand, in cocostudio parser, has so many memory not clean up when cocos2d system exited

dumganhar:
Thank you. Issue http://www.cocos2d-x.org/issues/3556 was created.

PR #4731, opened 2013-12-30
Develop fix raw stencil test

dabingnn:


samples/Cpp/TestCpp/Classes/ClippingNodeTest/ClippingNodeTest.h
```diff
@@ -164,8 +165,8 @@ class RawStencilBufferTest : public BaseClippingNodeTest
     void onDisableStencil();
     void onBeforeDrawClip(int planeIndex, const Point& pt);
     void onBeforeDrawSprite(int planeIndex, const Point& pt);
-protected:
-    Sprite* _sprite;
+private:
+    std::vector<Sprite*> _sprites;
```
dumganhar: Why not to use `Vector<Sprite*>` which resolves reference counting

PR #4748, opened 2013-12-31
animation part

super626:


minggo:
Please don't use Chinese words in log.
Thanks.

cocos/3d/C3DAABB.h
```diff
@@ -0,0 +1,85 @@
...
+	~C3DAABB();
+
+	/**
+	* Gets the center point of the bounding box.
+	*/
+	Vector3 getCenter();
+
+	Vector3 getSize();
+
+	float getLength();
```
minggo, line 48: Add const for `getXXX` functions?
Such as `float getLength() const`.

cocos/3d/C3DAABB.h
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	* Tests whether this bounding box intersects the specified bounding object.
+	*/
+	bool intersects(const C3DAABB& box) const;
+
+	bool containPoint( const Vector3& point) const;
+	bool containSphere( const Vector3 &center,float radius ) const;
+
+	float distance( const Vector3& vPoint);
```
minggo, line 63: How about rename to `getDistance`?

cocos/3d/C3DAABB.h
```diff
@@ -0,0 +1,85 @@
...
+/**
+* Defines a class describe 3-dimensional axis-aligned bounding box.
+*/
+class C3DAABB
+{
+public:
+
+	Vector3 _min;
+
+	Vector3 _max;
```
minggo, line 19: Why declare member variable as `public`?

cocos/3d/C3DAABB.h
```diff
@@ -0,0 +1,85 @@
...
+#include "math/Vector3.h"
+#include "math/Matrix.h"
+
+namespace cocos2d
+{
+class C3DPlane;
+/**
+* Defines a class describe 3-dimensional axis-aligned bounding box.
+*/
+class C3DAABB
```
minggo, line 13: Why not remove `C3D` prefix?
Is there any conflict?

cocos/3d/C3DOBB.cpp
```diff
@@ -0,0 +1,325 @@
...
+	Cov.m[9] = (float)(S2[1][2] - S1[1]*S1[2] / n) / n;
+	Cov.m[8] = (float)(S2[0][2] - S1[0]*S1[2] / n) / n;
+	Cov.m[1] = Cov.m[4];
+	Cov.m[2] = Cov.m[8];
+	Cov.m[6] = Cov.m[9];
+
+	return Cov;
+}
+
+static float& _getElement( Vector3& point, int index)
```
minggo, line 53: `float` type doesn't have to use reference. Did you want to modify the original value?

cocos/3d/C3DOBB.cpp
```diff
@@ -0,0 +1,325 @@
...
+{
+	if (index == 0)
+		return point.x;
+	if (index == 1)
+		return point.y;
+	if (index == 2)
+		return point.z;
+
+	CC_ASSERT(0);
```
minggo, line 62: I think it is better to add assert at the beginning of a function.

cocos/3d/C3DOBB.cpp
```diff
@@ -0,0 +1,325 @@
...
+			_getElement(b, ip) += _getElement(z, ip);
+			_getElement(d, ip) = _getElement(b, ip); // d.m[ip] = b.m[ip];
+			_getElement(z, ip) = 0.0f; //z.m[ip] = 0.0;
+		}
+	}
+
+	v.transpose();
+	*vout = v;
+	*dout = d;
+	return;
```
minggo, line 151: Don't need `return` here.

cocos/3d/C3DOBB.cpp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+//	return an OBB extracing from the vertices;
+static Matrix _GetOBBOrientation(const Vector3* aVertPos, int nVertCount)
+{
+	Matrix Cov;
```
minggo, line 157: Variable name starts with lower case.

cocos/3d/C3DPlane.h
```diff
@@ -0,0 +1,89 @@
...
+    float getDist() const  { return m_fDist; }
+
+	/**
+     * Return positon relatively to the plane.
+     */
+    POINT_CLASSIFICATION pointClassify(const Vector3& point) const;
+
+protected:
+    Vector3 m_vNormal;
+    float m_fDist;
```
minggo, line 83: `m_fDist` -> `_dist`.

cocos/3d/C3DSkeleton.h
```diff
@@ -0,0 +1,93 @@
...
+
+private:
+    C3DBone* _rootBone;
+    /// Lookup by bone name
+    std::map<std::string, C3DBone*> _bones;
+    
+    /// Storage of animations, lookup by name
+    std::map<std::string, C3DAnimation*> _animations;
+
+
```
minggo, line 85: Unneeded spaces.

cocos/3d/C3DActionListener.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+C3DActionListener::~C3DActionListener()
+{
+    _object = nullptr;
+    _function = nullptr;
+}    
+   
+
+C3DActionListener * C3DActionListener::create(ListenerObject* object, ListenerFunction function)
```
dumganhar, line 25: Why not to use `c++11` `std::function` instead of `target selector` callback mode?

cocos/3d/C3DAnimationCurve.h
```diff
@@ -0,0 +1,190 @@
...
+};
+
+    /**
+    * animation curve manager
+    */
+    class C3DAnimationCurveMgr
+    {
+    public:
+        typedef std::map<std::string, C3DAnimationCurve*> CurveMap;//boneId, animation curve for this bone
+        typedef std::map<std::string, CurveMap* > CurvesMap;//animation path, animation curve collection
```
dumganhar, line 132: Uses `c++11 std::unordered_map` will be better.

cocos/3d/C3DSkeleton.h
```diff
@@ -0,0 +1,93 @@
...
+     */
+    C3DSkeleton* clone();
+
+private:
+    C3DBone* _rootBone;
+    /// Lookup by bone name
+    std::map<std::string, C3DBone*> _bones;
+    
+    /// Storage of animations, lookup by name
+    std::map<std::string, C3DAnimation*> _animations;
```
dumganhar, line 83: uses c++11 unoredered_map will better.

PR #4755, opened 2013-12-31
Animation

houguohua:


cocos/3d/C3DActionListener.h
```diff
@@ -0,0 +1,103 @@
...
+      
+    /**
+     * Destructor.
+     */
+    virtual ~C3DActionListener();
+      
+    /**
+    * The method for creating C3DActionListener.
+    */
+    static C3DActionListener * create(ListenerObject* object, ListenerFunction function);
```
minggo, line 47: Use std::function<> instead of ListenerFunction?

cocos/3d/C3DAnimation.h
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * Adds an animation clip to current animation.
+     *
+     * @param name The name of the target animation clip.
+     * @param startFrame The start frame.
+     * @param endFrame The end frame.
+     * @param repeatCount repeat count.
+     * @param speed The speed of the animation clip play.
+     */
+    C3DAnimationClip* addClip(const char* name,unsigned int startFrame,unsigned int endFrame,float repeatCount,float speed);
```
minggo, line 54: It is better to use `int` instead of `unsigned int`. Refer to [this document](https://github.com/cocos2d/cocos2d-x/blob/develop/docs/CODING_STYLE.md#on-unsigned-integers) for detail information.

cocos/3d/C3DAnimation.h
```diff
@@ -0,0 +1,265 @@
...
+
+    /**
+     * Gets the animation's ID.
+     */
+    const char* getId() const;
+    
+    /**
+     * Gets the animation's duration.
+     */
+    unsigned long getDuration() const;
```
minggo, line 43: It is better to use `int` instead of `unsigned long`.

cocos/3d/C3DAnimationClip.h
```diff
@@ -0,0 +1,235 @@
...
+    unsigned short update(unsigned long elapsedTime);
+
+    unsigned short update_loop(unsigned long elapsedTime);
+
+    unsigned short update_once(unsigned long elapsedTime);
+
+    /**
+    * Is have target state
+    */
+    bool isState(unsigned short bit) const;
```
minggo, line 192: isState -> hasState?

cocos/3d/C3DAnimationClip.h
```diff
@@ -0,0 +1,235 @@
...
+     * Destructor.
+     */
+    ~C3DAnimationClip();
+
+    /**
+    * Update
+    */
+    unsigned short update(unsigned long elapsedTime);
+
+    unsigned short update_loop(unsigned long elapsedTime);
```
minggo, line 185: update_loop -> updateLoop.

cocos/3d/C3DActionListener.h
super626, line 47: OK

cocos/3d/C3DAnimationClip.h
super626, line 192: It means that it is in this state, running or stopping

minggo, line 192: Got it.

If so, i think this function is needless. Can use the result of `getState()` to compare.

cocos/3d/C3DAnimation.h
super626, line 43: change it to long, ok?

minggo, line 43: It is ok if it is needed. Will it cast to int in other API, because most of our API use int.

cocos/3d/C3DAnimationClip.h
super626, line 192: In fact, the state is complex, it may be a combined state, can not be compared using ==.

Your first suggestion is right, hasState is better.

cocos/3d/C3DAnimation.h
super626, line 43: We use unsigned in many place. And most of them will be used internally.

We will  unify our api to cocos2d::Action using adapter parten, can we change it at that time. I am not sure which will be needed by the adapter.

cocos/3d/C3DAnimationClip.h
super626, line 185: Ok

cocos/3d/C3DAnimation.h
minggo, line 43: Yep, we can modify it later.

PR #4758, opened 2013-12-31
Material

houguohua:


cocos/3d/C3DEffect.cpp
```diff
@@ -0,0 +1,519 @@
...
+void C3DEffect::bind()
+{
+	GL_ASSERT( glUseProgram(_program) );
+
+	__currentEffect = this;
+}
+
+C3DEffect* C3DEffect::getCurrentEffect()
+{
+	return __currentEffect;
```
minggo, line 486: __currentEffect -> _currentEffect

cocos/3d/C3DEffect.cpp
```diff
@@ -0,0 +1,519 @@
...
+}
+
+C3DEffect* Uniform::getEffect() const
+{
+	return _effect;
+}
+
+const char* Uniform::getName() const
+{
+	return _name.c_str();
```
minggo, line 511: Why not just return std::string?

cocos/3d/C3DEffect.cpp
```diff
@@ -0,0 +1,519 @@
...
+	}
+
+	SAFE_DELETE_ARRAY(vshSource);
+	SAFE_DELETE_ARRAY(fshSource);
+
+	return true;
+}
+
+
+bool C3DEffect::load(const char* vshSource, const char* fshSource, const char* defines)
```
minggo, line 156: How about break this function into more small functions?

cocos/3d/C3DEffect.h
```diff
@@ -0,0 +1,272 @@
...
+	void bindValue(Uniform* uniform, const C3DMatrix& value);
+
+	/**
+	* Sets a matrix array uniform value.
+	*
+	* @param uniform The uniform to set.
+	* @param values The array to set.
+	* @param count The number of elements in the array.
+	*/
+	void bindValue(Uniform* uniform, const C3DMatrix* values, unsigned int count = 1);
```
minggo, line 123: I think `GLProgram` also has these functions. Can reuse them?

cocos/3d/C3DEffect.cpp
houguohua, line 486: this is global var.

cocos/3d/C3DEffect.cpp
houguohua, line 156: good idea.split in into vs,fs

cocos/3d/C3DEffect.cpp
houguohua, line 511: ok

cocos/3d/C3DEffect.h
houguohua, line 123: program is one member of effect

minggo, line 123: Yep, can use `GLProgram` inside. Then don't have re-write the codes for loading vs and fs.

houguohua, line 123: GLProgram can not satisfy our need.

cocos/3d/C3DEffect.cpp
lvlonggame, line 486: all global var need prefix "__"  ?

minggo, line 486: About global variable, i suggest to use `g_xxx`. But it is ok to use `__xxx`.

cocos/3d/C3DEffect.cpp
lvlonggame, line 511: I will change this function.

PR #4759, opened 2013-12-31
fix crash related to not support the z length modifier for size_t on vs

WenhaiLin:


minggo:
Please modify other places that use `zd`.

PR #4761, opened 2013-12-31
Develop nutty modify framework modify label class name

nutty898:


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/4766 . Thanks.

PR #4810, opened 2014-01-02
Update SceneReaderTest Samples

chengstory:


cocos/2d/CCComponent.h
```diff
@@ -60,6 +60,9 @@ class CC_DLL Component : public Object
     virtual void setEnabled(bool b);
     static Component* create(void);
     
+    virtual Node* getNode();
+	virtual void setNode(Node *node);
```
dumganhar: What's the difference between `(s|g)Node` and `(s|g)etOwner`?

cocos/editor-support/cocostudio/CCComAttribute.cpp
```diff
@@ -116,5 +116,13 @@ ComAttribute* ComAttribute::create(void)
 	return pRet;
 }
 
+void ComAttribute::setJsonFile(const std::string &jsonName)
+{
+	_jsonName = jsonName;
+}
+std::string ComAttribute::getJsonFile()
```
dumganhar: Returns `const std::string&` will be better.

cocos/editor-support/cocostudio/CCSSceneReader.cpp
```diff
@@ -200,9 +200,9 @@
                     }
                     
                     gb->addComponent(pRender);
-					if (_pListener && _pfnSelector)
+					if (_listener && _fnSelector)
 					{
-						(_pListener->*_pfnSelector)(pSprite, (void*)(&subDict));
+						(_listener->*_fnSelector)(pSprite, (void*)(&subDict));
```
dumganhar: It's better to use `c++11 std::function` for callbacks on develop branch.
I just mention it to let you know that. We should not use `target selector` mode if `std::function` could accomplish.

PR #4813, opened 2014-01-03
fix compile error of Map's getRandomObject

zilongshanren:


zilongshanren:
unordered_map doesn't support iterator arithmetic operation, so I change it to std::advance

zilongshanren:
when I call the getRandomObject, I must call `srand(time(NULL))` method first. I think we should call it for our developers.

PR #4821, opened 2014-01-03
optimize button

nutty898:


cocos/gui/UIButton.cpp
```diff
@@ -29,10 +29,10 @@
...
 namespace gui {
 
-#define NORMALRENDERERZ (0)
-#define PRESSEDRENDERERZ (0)
-#define DISABLEDRENDERERZ (0)
-#define TITLERENDERERZ (1)
+#define NORMALRENDERERZ (-2)
+#define PRESSEDRENDERERZ (-2)
+#define DISABLEDRENDERERZ (-2)
+#define TITLERENDERERZ (-1)
```
dumganhar: I think using `static const int XXX = 111;` will be better. Right?

cocos/gui/UIButton.h
```diff
@@ -132,16 +132,16 @@ class Button : public Widget
     virtual void setScale9Enabled(bool able);
     
     //override "setFlipX" of widget.
-    virtual void setFlipX(bool flipX) override;
+    virtual void setFlipX(bool flipX);
     
     //override "setFlipY" of widget.
-    virtual void setFlipY(bool flipY) override;
+    virtual void setFlipY(bool flipY);
```
dumganhar: Why to remove `override` keyword?

cocos/gui/UIButton.cpp
minggo: Yep.
And it is better to use `_` to connect multiple words with constants, such as

```
NORMALRENDERERZ -> NORMAL_RENDERER_Z
```

dumganhar: +1 :)

nutty898: OK

PR #4439, opened 2013-12-05
Cocos2dx master hotfix (small bug)

gelldur:
If in .plist file we don't have those lines
    <key>textureFileName</key>
    <string>mainscreen.png</string>

We should have CCAssert now. Before changes application crush with SIGSEGV

PR #4846, opened 2014-01-03
Cocos2dx develop hotfix

gelldur:
The same as here https://github.com/cocos2d/cocos2d-x/pull/4439

If in .plist file we don't have those lines
textureFileName
mainscreen.png

We should have CCAssert now. Before changes application crush with SIGSEGV + small fix in documentation


dumganhar:
@gelldur , your changes is correct.
But I could reproduce what you said `SIGSEGV` issue.
It did make an assert in 

``` c++
std::string textureData = dictionary["textureImageData"].asString();
CCASSERT(!textureData.empty(), "");
```

If in .plist file we don't have those lines
textureFileName
mainscreen.png

dumganhar:
Merged. Thanks.

PR #4856, opened 2014-01-04
add function for skip binding

WenhaiLin:


dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/4858/files

PR #4885, opened 2014-01-04
Fixes the font rendering on Windows

seagullua:
When font was rendered the black pixels used to appear around the letters. This patch fixes this issue


seagullua:
Labels zoomed 300%
cocos2d-bug

sergey-shambir:
After this patch, is blurred area white or semi-transparrent? In other words, would it look normally at black surface?

seagullua:
Yes it is semi-tranparrent.

seagullua:
untitled-3

minggo:
Thanks.

PR #4891, opened 2014-01-05
make TMXLayer::getProperties() return reference

minggo:
`TMXMapInfo::startElement()` use the following codes

``` cpp
TMXLayerInfo* layer = tmxMapInfo->getLayers().back();
 Value value = attributeDict["value"];
std::string key = attributeDict["name"].asString();
// Add the property to the layer
layer->getProperties().insert(std::make_pair(key, value));
```

If `TMXLayer::getProperties()` returns an object, then it can not change the value of `TMXLayer::_properties`. I think it is reasonable to return reference of it, because it can speed up, and TMXLayer is used internally.

### DIFF
--- a/cocos/2d/CCTMXXMLParser.cpp
+++ b/cocos/2d/CCTMXXMLParser.cpp
@@ -57,7 +57,7 @@ TMXLayerInfo::~TMXLayerInfo()
     }
 }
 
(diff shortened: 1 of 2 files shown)


dumganhar:
YEP,YOU'RE RIGHT.THANKS FOR THIS PATCH.

PR #4893, opened 2014-01-05
open external URL in internal default device's browser

sortris:
Patch that allows to open external URLs in device's internal default browser. Also with JS-bindings.
Usage:
cc.Application.getInstance().openURL("externalURL");

Tested on:
Samsung Galaxy S4 Mini
Samsung Galaxy S4
Samsung Galaxy Note2
Google Nexus 10
Nexus One

Apple iPhone 5 with iOS7
Apple iPhone 4S with iOS7
Apple iPad Retina with iOS7

### DIFF
--- a/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
+++ b/cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
@@ -151,7 +151,13 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
         // Set framelayout as the content view
 		setContentView(framelayout);
 	}
-	
+
+	public static void openURL(String url) {
+       		Intent i = new Intent(Intent.ACTION_VIEW);
+        	i.setData(Uri.parse(url));
+	        me.startActivity(i);
+	}
+
     public Cocos2dxGLSurfaceView onCreateView() {
     	return new Cocos2dxGLSurfaceView(this);
     }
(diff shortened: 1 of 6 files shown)


cocos2dx/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
```diff
@@ -151,7 +151,13 @@ public void init() {
         // Set framelayout as the content view
 		setContentView(framelayout);
 	}
-	
+
+	public static void openURL(String url) {
+       		Intent i = new Intent(Intent.ACTION_VIEW);
+        	i.setData(Uri.parse(url));
```
gelldur, line 157: probably here you should add Intent.FLAG_ACTIVITY_NEW_TASK. Because without this one sometimes it crashes on my Nexus4

my code looks like:

```
// Opens URL as the new activity
public static void openURL(String url) {
    Intent i = new Intent(Intent.ACTION_VIEW);
    i.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
    i.setData(Uri.parse(url));
    sContext.startActivity(i);
}
```

gelldur, line 157: also exceptions for invalud parsing uri should be handled ?

gelldur:
Please look for simlar commit on forum because someone adds also Linux,Windows support :)

dumganhar:
@gelldur 
Could you tell me the URL of the similar patch on our forum? Thanks.

dumganhar:
It's better to send this patch to `develop` branch too. Thanks.

gelldur:
http://cocos2d-x.org/forums/6/topics/11290
https://github.com/cocos2d/cocos2d-x/pull/1940

I can;t find one more that i'm using

sachingarg05:
It will be great if this is merged. Looking forward.

PR #4924, opened 2014-01-06
Fix CCBAnimationManager inconsistency between cocos2d-iphone and cocos2d-x

x1024:
Running animations with CCBAnimationManager no longer cancels all actions currently running on a node.
It now only cancels actions that have been started by the animation.

This is a mirror of: https://github.com/cocos2d/CCBReader/commit/6598e89625fa495269bc19ad85a5d2f918d244cb


dumganhar:
Could you send this patch to develop branch too?
Master branch is only for critical bug fixes now.
Thanks.

PR #4925, opened 2014-01-06
fix crash in CCAnimation.h on vs.

WenhaiLin:


dumganhar:
Fixed this issue at https://github.com/cocos2d/cocos2d-x/pull/4927

PR #4933, opened 2014-01-06
fix cc.pDistanceSQ

samael65535:


samuele3hu:
This API refers to the Point::getLengthSq,the "SQ" suffix means square not sqrt.

samael65535:
oringin source:

function cc.pLengthSQ(pt)
    return cc.pDot(pt)
end

It has only one argument. How can I use it?

samuele3hu:
Yeah,It is a bug,We should do like this:
function cc.pLengthSQ(pt)
    return cc.pDot(pt,pt)
end

Is it right?

samuele3hu:
pt is a table like this pt = { x = valueX, y = valueY}

samael65535:
Yeah

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/4985. Thanks.

PR #4990, opened 2014-01-07
const correctness: use cbegin/cend, const std::string

ghost:


dumganhar:
Hi @daltomi ,
Did our codes cause some warning on your platform if don't use `cbegin`, `cend` and `const std::string`?

kis2009dsh:
Hi @daltomi 

There exist two methods of std::map::find  which listing as follow 

iterator find( const Key& key );
const_iterator find( const Key& key ) const;

You shall only use const_iterator when the object is const qualified . 

Thanks.

ghost:
Hi @dumganhar 
No, no warning at all.

dumganhar:
Thanks. merged.

PR #5003, opened 2014-01-09
fix WebSocket cannot send/receive more than 4096 bytes of data

edwardzhou:
I have setup a temporary websocket server for yr testing.
And, the write buffer size reduced to 2048. I re-tested 4096 is okay in intranet, but failed on internet. 2048 is the safe size tested.

here is the test case in LUA:

  ws = WebSocket:create("ws://118.26.229.45:8000")

  ws:registerScriptHandler(function(strData)
    print('Websocket Open: ', strData)
    local data = ''
    for i=1, 2560 do
      data = data .. '0123456789abcdef'
    end
    -- send out 40k string
    ws:sendString(data)
  end,cc.WEBSOCKET_OPEN)

  ws:registerScriptHandler(function(strData)
    print('Websocket Message: ', #strData, strData)  
  end,cc.WEBSOCKET_MESSAGE)

### DIFF
--- a/cocos/network/WebSocket.cpp
+++ b/cocos/network/WebSocket.cpp
@@ -37,6 +37,8 @@
 
 #include "libwebsockets.h"
 
+#define WS_WRITE_BUFFER_SIZE 2048
+
 NS_CC_BEGIN
 
 namespace network {
@@ -221,6 +223,9 @@ WebSocket::WebSocket()
 , _delegate(nullptr)
 , _SSLConnection(0)
 , _wsProtocols(nullptr)
+, _pending_frame_data_len(0)
+, _current_data_len(0)
+, _current_data(NULL)
 {
 }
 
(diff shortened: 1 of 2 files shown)


edwardzhou:
the temp websocket server is on ws://118.26.229.45:8000
it will send back the received data.

dumganhar:
OK, I will test it by using the server. Thanks.

dumganhar:
Two questions:
1) Could we test it by `ws://echo.websocket.org` ?
2) I sent text, but your test server reply me with `binary data`.

edwardzhou:
1) ws://echo.websocket.org   won't handle large incoming data.

2) the test server is a simple faye-websocket-node server, I found it only send back the first 64 chars of the string (echo.websocket.org does the same). so send back binary data instead.

and now, the test server was refactor to send back in text-format already. please try again.

PR #5011, opened 2014-01-09
fix compiling error in CCControlSlider.cpp on vs.

WenhaiLin:


dumganhar:
Oh, it's my fault that I didn't compile it by VS. Thanks for your fix.

PR #5033, opened 2014-01-10
Remove spammy debug logging on resizing a CCArray

BodbDearg:
The logging here is a bit too verbose and ends up filling up a lot of the console output when Cocos2dx debugging is enabled, hiding other potentially useful information. If we want to log these sort of events perhaps we should have an extra (extremely verbose) level of logging or tracing that developers can turn on to help diagnose serious issues. At the minute there is none, so suggesting we remove this.


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/ccCArray.cpp#L76
Thanks.

PR #5035, opened 2014-01-11
Console fixes

ricardoquesada:
- Console no longer inherits from `Object`
- The way to get the console is by doing `director->getConsole()`
- It is possible to redirect all the `cocos2d::log()` messages to the remote console
- `cocos2d::log` functions were moved from `2d/platform/XXX/CCCommon.cpp` to `CCConsole.cpp`

### DIFF
--- a/cocos/base/CCObject.h
+++ b/cocos/base/CCObject.h
@@ -107,7 +107,6 @@ public:
      */
     inline void release()
     {
-        cocos2d::log("carlos");
         CCASSERT(_reference > 0, "reference count should greater than 0");
         --_reference;
 


ricardoquesada:
And also `Director:setDelegate` was removed since the Director now emits an event when the projection is changed

cocos/base/CCConsole.cpp
```diff
@@ -108,43 +109,94 @@ static const char* inet_ntop(int af, const void* src, char* dst, int cnt)
...
+
+void log(const char * format, ...)
+{
+    va_list args;
+    va_start(args, format);
+    log(format, args);
+    va_end(args);
+}
+
+void log(const char *format, va_list args)
```
samuele3hu, line 134: When calling CCLOG("%s","test string"), it will call this function,then lead to crash.

ricardoquesada, line 134: I wasn't able to reproduce the bug, but I changed the API a bit in order to prevent overload resolution issues:
https://github.com/cocos2d/cocos2d-x/pull/5073

PR #5039, opened 2014-01-11
Added x64 support for Windows - CCExtensions

martell:
Just like before where I added audio support for windows x64 this should now add support for Extensions.

### DIFF
--- a/extensions/proj.win32/Win32InputBox.cpp
+++ b/extensions/proj.win32/Win32InputBox.cpp
@@ -280,14 +280,14 @@ void CWin32InputBox::InitDialog()
 // Message handler for about box.
 LRESULT CALLBACK CWin32InputBox::DlgProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
 {
-  CWin32InputBox *_this = (CWin32InputBox *) ::GetWindowLong(hDlg, GWL_USERDATA);
+  CWin32InputBox *_this = (CWin32InputBox *) ::GetWindowLongPtr(hDlg, GWLP_USERDATA);
   WIN32INPUTBOX_PARAM *param = _this ? _this->GetParam() : 0;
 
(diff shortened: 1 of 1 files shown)


extensions/proj.win32/Win32InputBox.cpp
```diff
@@ -280,14 +280,14 @@ void CWin32InputBox::InitDialog()
 // Message handler for about box.
 LRESULT CALLBACK CWin32InputBox::DlgProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
 {
-  CWin32InputBox *_this = (CWin32InputBox *) ::GetWindowLong(hDlg, GWL_USERDATA);
+  CWin32InputBox *_this = (CWin32InputBox *) ::GetWindowLongPtr(hDlg, GWLP_USERDATA);
```
dumganhar, line 283: Hi @martell ,
What's the problem of using `G/SetWindowLong` in 64bit windows?
Could you describe more about it? Why do we have to use `GetWindowLongPtr`?
Thanks.

martell, line 283: Hi @dumganhar,
Please look at the first note here
http://msdn.microsoft.com/en-us/library/windows/desktop/ms633584%28v=vs.85%29.aspx
and here
http://msdn.microsoft.com/en-us/library/windows/desktop/ms633585%28v=vs.85%29.aspx

I done this to the audio engine a few weeks ago and it was merged.
We get an error on compile on windows x64 with the non ptr version version.

Basically the return type is changed to LONG_PTR which is 32bit on x86 and 64bit x64
The non pointer version is always a LONG which is 32bit and can't be cast to (CWin32InputBox *_this) on x64 becuase _this is 64bits long.

Hope that explains the change :)

dumganhar:
Thanks. Merged.

PR #5040, opened 2014-01-11
Added clang support for Windows

martell:
I've added a script to build cocos2d-x with clang on windows through mingw64.

I had to enable thread local storage in clang for a successful build.
Alexey will be putting up an updated clang later today with tls enabled.

Also clang on windows is bleeding edge and cmake from git is required thanks to this bug
http://www.gccxml.org/Bug/view.php?id=13035

when cmake 3.0 is out in a few days we will switch and this should work like a charm for everyone

### DIFF
--- a/build/build-mingw32-clang-make.sh
+++ b/build/build-mingw32-clang-make.sh
@@ -3,7 +3,7 @@
 # msys2 Pacman Manager for cocos2d-x
 
 #/****************************************************************************
-# Copyright (c) 2012-2013 Martell Malone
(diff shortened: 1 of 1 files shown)


martell:
Hey @dumganhar 

Can you see that this is merged.
here is a how to guide for the Dev's also
https://gist.github.com/martell/8362294 

I have also used this build of clang with emscripten on windows which I will add to the 2.2 branch at some point

Thanks
Martell

dumganhar:
Thanks. Merged.

PR #5041, opened 2014-01-11
closed #3630:Uses a struct of TTF configuration for Label::createWithTTF  to reduce parameters and make this interface more easily to use.

WenhaiLin:
1. Uses a struct for TTF configuration for Label::createWithTTF to reduce parameters
2. Add create/setTTFConfig/setBMFontFilePath for make this interface more easily to use.

### DIFF
--- a/cocos/scripting/javascript/bindings/cocos2d_specifics.cpp
+++ b/cocos/scripting/javascript/bindings/cocos2d_specifics.cpp
@@ -3936,7 +3936,7 @@ JSBool jsval_to_TTFConfig(JSContext *cx, jsval v, TTFConfig* ret) {
 
     ret->fontSize = (int)fontSize;
     ret->glyphs = GlyphCollection((int)glyphs);
-    ret->distanceFieldEnable = distanceFieldEnable;
+    ret->distanceFieldEnabled = distanceFieldEnable;
     if(ret->glyphs == GlyphCollection::CUSTOM && customGlyphs.length() > 0)
         ret->customGlyphs = customGlyphs.c_str();
     else


cocos/2d/CCLabel.h
```diff
@@ -54,23 +54,45 @@ enum class LabelEffect {
...
 class FontAtlas;
 
-
+typedef struct _ttfConfig
+{
+    std::string fontFilePath;
+    int fontSize;
+    GlyphCollection glyphs;
+    const char *customGlyphs;
+    bool distanceFieldEnable;
```
minggo: distanceFieldEnable -> distanceFieldEnabled

cocos/2d/CCLabel.h
```diff
@@ -54,23 +54,45 @@ enum class LabelEffect {
 struct FontLetterDefinition;
 class FontAtlas;
 
-
+typedef struct _ttfConfig
+{
+    std::string fontFilePath;
+    int fontSize;
+    GlyphCollection glyphs;
+    const char *customGlyphs;
```
minggo, line 62: Why use const char*?

cocos/2d/CCLabel.cpp
```diff
@@ -118,6 +109,7 @@ Label* Label::createWithAtlas(FontAtlas *atlas, TextHAlignment alignment, int li
 , _fontSize(0)
 , _uniformEffectColor(0)
 {
+    _cascadeColorEnabled = true;
```
dumganhar, line 111: Why not to add it after the line `, _uniformEffectColor(0)`?

WenhaiLin, line 111: _cascadeColorEnabled is not label's member variable

cocos/2d/CCLabel.h
WenhaiLin, line 62: It mean create  two string when the formal parameter is std::string& and the  actual parameter is char*.

WenhaiLin, line 62: In most cases,the actual parameter is char*.

dumganhar:
Thanks.

PR #5046, opened 2014-01-13
Fix for compiling error for windows / new label bug fixes

hanjukim:
- Conflict with 'far', 'near' defined in windef.h
- undefined variable 'szBuf' in Console (windows)
- Fix for LabelTextFormatter bug
- TTF position bug for new label
- Fix for crash bug when using unknown character

### DIFF
--- a/samples/Cpp/TestCpp/Classes/LabelTest/LabelTestNew.cpp
+++ b/samples/Cpp/TestCpp/Classes/LabelTest/LabelTestNew.cpp
@@ -1319,17 +1319,49 @@ LabelTTFOldNew::LabelTTFOldNew()
     auto s = Director::getInstance()->getWinSize();
 	float delta = s.height/4;
 
-	auto label1 = CCLabelTTF::create("Cocos2d-x Old", "arial", 24);
+	auto label1 = CCLabelTTF::create("Cocos2d-x Label Test", "arial", 24);
 	addChild(label1, 0, kTagBitmapAtlas1);
     label1->setAnchorPoint(CCPoint(0.5f, 0.5f));
     label1->setPosition(CCPoint(s.width/2, delta * 2));
 	label1->setColor(Color3B::RED);
 
     TTFConfig ttfConfig("fonts/arial.ttf", 48);
-	auto label2 = Label::createWithTTF(ttfConfig, "Cocos2d-x Old");
+	auto label2 = Label::createWithTTF(ttfConfig, "Cocos2d-x Label Test");
     addChild(label2, 0, kTagBitmapAtlas2);
     label2->setAnchorPoint(CCPoint(0.5f, 0.5f));
     label2->setPosition(CCPoint(s.width/2, delta * 2));
+}
+
+void LabelTTFOldNew::onDraw()
+{
+    kmMat4 oldMat;
+    kmGLGetMatrix(KM_GL_MODELVIEW, &oldMat);
+    kmGLLoadMatrix(&_modelViewTransform);
+    
+    auto label1 = (Label*)getChildByTag(kTagBitmapAtlas2);
+    auto labelSize = label1->getContentSize();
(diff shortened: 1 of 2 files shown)


cocos/2d/renderer/CCFrustum.cpp
```diff
@@ -27,6 +27,9 @@
 
 #include <stdlib.h>
 
+#undef far
+#undef near
```
dumganhar: I think it isn't a good idea to `undefine` `far` and `near`.
We should not use `far` and `near` in our codes. Instead, it should be `farPlane` and `nearPlane`.

cocos/2d/CCLabel.cpp
```diff
@@ -454,6 +458,8 @@ bool Label::recordPlaceholderInfo(int spriteIndex)
     if (static_cast<std::size_t>(spriteIndex) >= _lettersInfo.size())
     {
         LetterInfo tmpInfo;
+
+        memset(&tmpInfo, 0, sizeof(tmpInfo));
```
WenhaiLin: Why there is need for call the memset method? It will set validDefinition to false below.

cocos/2d/CCLabel.cpp
```diff
@@ -438,6 +440,8 @@ bool Label::recordLetterInfo(const cocos2d::Point& point,unsigned short int theC
...
```
WenhaiLin: It will reassignment below.Could you describe the bug when not initialize the tmpInfo?

cocos/2d/CCLabel.cpp
hanjukim:
```
_fontAtlas->getLetterDefinitionForChar(theChar, _lettersInfo[spriteIndex].def);
```

if above code fails, _lettersInfo[spriteIndex].def.validDefinition will be uninitialized value.
I found that below code also can fix the problem

```
    if (!_fontAtlas->getLetterDefinitionForChar(theChar, _lettersInfo[spriteIndex].def))
        _lettersInfo[spriteIndex].def.validDefinition = false;
```

WenhaiLin: @hanjukim Thanks for your comment. I think set [validDefinition] to [false] in FontAtlas::getLetterDefinitionForChar method is more reasonable.

hanjukim: @Dhilan007 Since getLetterDefinitionForChar can be used in other situations, I agree with you

WenhaiLin:
@hanjukim I submit  PQ for fix crash when using unknown character.https://github.com/cocos2d/cocos2d-x/pull/5082. Thanks for your feedback again.

cocos/2d/CCFontAtlas.cpp
```diff
@@ -114,6 +114,8 @@ bool FontAtlas::prepareLetterDefinitions(unsigned short *utf16String)
     std::unordered_map<unsigned short, FontLetterDefinition> fontDefs;
     int length = cc_wcslen(utf16String);
 
+    auto yDelta = _currentPageLineHeight * 0.7;
```
WenhaiLin: How the proportionality coefficient calculated?

hanjukim: I think we need more research on this. I found that it's not working well on every platform (mac)
Actually, what I was trying to do is reversely porting new label system into Cocos2d-x 2.x and use it for CocosBuilder extension to make my games cooler.
I finished this work and then found several problems positioning bug with TTF label.
Positioning problem with new label system can be easily examined if we create old/new TTF label together with same position, anchor and bounding box.
However everything's not working well as used to be :(

WenhaiLin: Hi,I think i find out the reason of the bug.I fixed it at https://github.com/cocos2d/cocos2d-x/pull/5164.

PR #5047, opened 2014-01-13
Fix long string will be cut off by function “cc_utf8_to_utf16”

btspoony:
“cc_utf8_strlen” returns long

### DIFF
--- a/cocos/2d/ccUTF8.cpp
+++ b/cocos/2d/ccUTF8.cpp
@@ -279,7 +279,7 @@ cc_utf8_get_char (const char * p)
 
 unsigned short* cc_utf8_to_utf16(const char* str_old, int length/* = -1 */, int* rUtf16Size/* = nullptr */)
 {
-    unsigned short len = cc_utf8_strlen(str_old, length);
+    long len = cc_utf8_strlen(str_old, length);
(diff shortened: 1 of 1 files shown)


cocos/2d/ccUTF8.cpp
```diff
@@ -279,7 +279,7 @@ void cc_utf8_trim_ws(std::vector<unsigned short>* str)
 
 unsigned short* cc_utf8_to_utf16(const char* str_old, int length/* = -1 */, int* rUtf16Size/* = nullptr */)
 {
-    unsigned short len = cc_utf8_strlen(str_old, length);
+    long len = cc_utf8_strlen(str_old, length);
     if (rUtf16Size != nullptr) {
         *rUtf16Size = len;
```
dumganhar, line 284: Thanks. Yep, your fix is correct.
rUtf16Size is `int*`, we assign `long` to `int` will also lose accuracy

btspoony, line 284: Hi, will you merge this pull request? or has another solution?(like change long to int)

dumganhar, line 284: I will merge this PR soon. Thanks.

PR #5053, opened 2014-01-14
fix compiling error cause by clash with the existing macro definition on window platform.

WenhaiLin:
1.fix compiling error cause by clash with the existing macro definition on window platform.
2.fix undefined variable 'szBuf' in Console (windows)

### DIFF
--- a/cocos/2d/CCTMXTiledMap.cpp
+++ b/cocos/2d/CCTMXTiledMap.cpp
@@ -30,8 +30,6 @@ THE SOFTWARE.
 #include "CCSprite.h"
 #include <algorithm>
 
-#undef min
-#undef max
 NS_CC_BEGIN
 
 // implementation TMXTiledMap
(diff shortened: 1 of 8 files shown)


cocos/2d/CCTMXTiledMap.cpp
```diff
@@ -30,6 +30,8 @@ of this software and associated documentation files (the "Software"), to deal
 #include "CCSprite.h"
 #include <algorithm>
 
+#undef min
+#undef max
```
dumganhar: I noticed there are lots of `#undef` in `cpp` files. In fact, only `windows` define these macros.
So I think we need to un-define them in `platform/win32/CCStdC.h`.

WenhaiLin: You are right.Thanks.

cocos/2d/platform/win32/CCStdC.h
```diff
@@ -148,5 +148,8 @@ inline errno_t strcpy_s(char *strDestination, size_t numberOfElements,
 #undef DELETE
 #endif
 
+#undef min
+#undef max
```
dumganhar, line 152: Please check whether the macro was defined. Thanks.

PR #5057, opened 2014-01-14
refactor PhysicsContact inherits from EventCustom; simplify emitting col...

edwardzhou:
...lision events

a bit addition, for LUA binding, I added PhysicsContact::cast to cast correct type from EventCustom. Because, tolua.cast won't return the right data, and it will crash the app.
# eg.

```
local function collisionCallback(event)
    print("Custom event received, ", event)
    print('getEventName => ', event:getEventName())
    local contact = tolua.cast(event, 'PhysicsContact')
    print('shapeA => ', contact:getShapeA())  --- app will crash here.
    print('A tag => ', contact:getShapeA():getBody():getNode():getTag())
    print('B tag => ', contact:getShapeB():getBody():getNode():getTag())        
end

local listener1 = cc.EventListenerCustom:create("PhysicsContactEvent", collisionCallback)
self._listener1 = listener1
local eventDispatcher = self:getEventDispatcher()
eventDispatcher:addEventListenerWithFixedPriority(listener1, 1)
```
# 
# but in this way is ok

```
local function collisionCallback(event)
    print("Custom event received, ", event)
    print('getEventName => ', event:getEventName())
    local contact = cc.PhysicsContact:cast(event) -- let PhysicsContact handle the type
    print('shapeA => ', contact:getShapeA())
    print('A tag => ', contact:getShapeA():getBody():getNode():getTag())
    print('B tag => ', contact:getShapeB():getBody():getNode():getTag())
end

local listener1 = cc.EventListenerCustom:create("PhysicsContactEvent", collisionCallback)
self._listener1 = listener1
local eventDispatcher = self:getEventDispatcher()
eventDispatcher:addEventListenerWithFixedPriority(listener1, 1)
```
#

### DIFF
--- a/cocos/physics/CCPhysicsContact.h
+++ b/cocos/physics/CCPhysicsContact.h
@@ -86,8 +86,6 @@ public:
     inline void setData(void* data) { _data = data; }
     /** get the event code */
     EventCode getEventCode() const { return _eventCode; };
-    
-    static PhysicsContact* cast(EventCustom* event);
 
 private:
     static PhysicsContact* construct(PhysicsShape* a, PhysicsShape* b);
--- a/cocos/physics/CCPhysicsContact.cpp
+++ b/cocos/physics/CCPhysicsContact.cpp
@@ -71,11 +71,6 @@ PhysicsContact* PhysicsContact::construct(PhysicsShape* a, PhysicsShape* b)
     return nullptr;
 }
 
-PhysicsContact* PhysicsContact::cast(EventCustom* event) {
-	PhysicsContact* contact = static_cast<PhysicsContact*>(event);
-	return contact;
-}
-
 bool PhysicsContact::init(PhysicsShape* a, PhysicsShape* b)
 {
     do


cocos/physics/CCPhysicsContact.h
```diff
@@ -86,6 +87,8 @@ class PhysicsContact : Event
     /** get the event code */
     EventCode getEventCode() const { return _eventCode; };
     
+    static PhysicsContact* cast(EventCustom* event);
```
boyu0: I don't think this is necessary. I can do this in `lua_cocos2dx_physics_manual.cpp`.
And, you don't need to use `EventListenerCustom` to receive events, you can use `EventListenerPhysicsContact` or it's subclasses instead. Lua banding for these listeners haven't finished yet, I will commit a PR for this recently.
And `event` param doesn't needed in `EventListenerPhysicsContact` callback functions, I will fix this after this PR is merged.
BTW, great job for this PR, thanks!

edwardzhou: I agree on it. It is just a workaround for this moment for me :)

boyu0:
I created a issue for this:
http://www.cocos2d-x.org/issues/3716
I will continue to finish this work and commit another PR, thanks.

PR #5058, opened 2014-01-14
ScrollView should implement ActionTweenDelegate

zhiqiangxu:
ScrollView uses ActionTween, so it should implement CCActionTweenDelegate::updateTweenAction.

(the same as ControlSwitchSprite)

If CCScrollView::setZoomScaleInDuration is ever executed, it will crash.


litefeel:
if you want implement  CCActionTweenDelegate, please use Node, not ScrollView.

dumganhar:
Thanks.

PR #5060, opened 2014-01-14
setScissorInPoints should also be called even if two ScissorRect doesn't intersect

zhiqiangxu:
If the parent scissor rect doesn't intersect with the current one, we should call setScissorInPoints(x,y,0,0), not skip.


dumganhar:
@zhiqiangxu , thanks.
Could you please add a test case for this issue? The same as https://github.com/cocos2d/cocos2d-x/pull/5058

zhiqiangxu:
@dumganhar 

The version I forked from doesn't compile(5ec17a), so I don't have the environment to test it.

But I think it's correct for obvious reasons.

Besides, no idea how to write the test case.

cocosQA:
Dear zhiqiangxu:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6466

PR #5061, opened 2014-01-14
UIWidget::getChildByName should recursive search

zhujj2008:
Hi,  
     In v3.0 beta, the UIWidget::getChildByNam does not recursively search the children widgets when parsing the cocosstudio exported json file,  which cause our code cannot find some child widget when loading cocossdudio generated UI.  
    Hope this patch can fix all the problem.

```
Thanks.
```

### DIFF
--- a/cocos/gui/UIWidget.cpp
+++ b/cocos/gui/UIWidget.cpp
@@ -248,6 +248,14 @@ Widget* Widget::getChildByName(const char *name)
             {
                 return widgetChild;
             }
+            else
+            {
+                widgetChild = widgetChild->getChildByName(name);
+                if (widgetChild)
+                {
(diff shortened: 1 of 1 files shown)


nutty898:
Hello.

Thanks for your pull request.

If you want to recursively search a widget, you can use Helper::seekWidgetByName(Widget\* root, const char\* name);.

We hope that "getChildByName" and "getChildByTag" are the same "level".

Node::getChildByTag does not recursively search, right?

zhujj2008:
Understand,  thanks for give then right usage method.

dumganhar:
@zhujj2008 
So could this PR be closed?

zhujj2008:
Closed

PR #5062, opened 2014-01-15
Fix vbo full

darkdukey:
Fix a bug that renderer will have render artifact when exceeding VBO size, Also adding a test for it.


samples/Cpp/TestCpp/Classes/NewRendererTest/NewRendererTest.cpp
```diff
@@ -485,3 +486,32 @@ void NewClippingNodeTest::onTouchesEnded(const std::vector<Touch *> &touches, Ev
...
+
+std::string VBOFullTest::title() const
+{
+    return "New Renderer";
+}
+
+std::string VBOFullTest::subtitle() const
+{
+    return "VBO full Test, everthing should render normally";
+}
```
ricardoquesada: github says that a new line is missing :)

PR #5063, opened 2014-01-15
Performance Test fixes

ricardoquesada:
- SpritePerfTest: 3 new tests added
- SpritePerfTest: random() always use the same seed
- SpritePerfTest: Fixed tests 8 and 9 on retina display machines
- Console: Added 3 new commands: 'config', 'fileutils dump', 'textures'

### DIFF
--- a/cocos/2d/CCTextureCache.cpp
+++ b/cocos/2d/CCTextureCache.cpp
@@ -97,7 +97,7 @@ void TextureCache::addImageAsync(const std::string &path, std::function<void(Tex
 {
     Texture2D *texture = nullptr;
 
-    std::string fullpath = FileUtils::getInstance()->fullPathForFilename(path.c_str());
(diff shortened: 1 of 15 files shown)


cocos/2d/CCTextureCache.cpp
```diff
@@ -645,7 +671,7 @@ void VolatileTextureMgr::reloadAllTextures()
             break;
         case VolatileTexture::kString:
             {
-                vt->_texture->initWithString(vt->_text.c_str(), vt->_fontDefinition);
+                vt->_texture->initWithString(vt->_text, vt->_fontDefinition);
```
dumganhar, line 674: It's here cause the compilation error since `bool Texture2D::initWithString(const char *text, const char *fontName` access `const char*`.

ricardoquesada, line 674: yep. thanks. fixed.

PR #5068, opened 2014-01-15
add performance test for New renderer: LargeTileMap test for quadCommand

dabingnn:


minggo:
The corresponding issue in redmine is http://cocos2d-x.org/issues/3640.
Please use `issue #xxx` or `closed #xxx` to link the pull request to corresponding issue in future.
Thanks.

minggo:
And please modify windows project too. Or ask somebody to send you a pull request for it.

ricardoquesada:
sounds good. Currently Github says that it cannot be merged. If you fix those issues, I'll merge it. thanks.

dabingnn:
@ricardoquesada I have fixed merge conflicts, could you review it later?

ricardoquesada:
done, thanks.

minggo:
I had asked @dabingnn to add this performance test based on v3.0-alpha1 in his repo.

@dabingnn 
Could you please notify it when it is done.
Thanks.

ricardoquesada:
Thanks.
Yes, please, let me know if this test runs >= 5% slower than v3.0-alpha1.
If so, I'll fix the issue. 
thanks.

dabingnn:
@minggo @ricardoquesada I have tested this patch on my android phone. a good news is that, the performance degradation is caused by our bug in the new renderer. the bug is fixed by Nite, link is https://github.com/cocos2d/cocos2d-x/pull/5062,
before this fix, our app will crash on ios but still work on android. However, I guess there may be some android driver error happened when we run this test case. The device will need to recover from error, which has a strong affect on FPS.

ricardoquesada:
great!

PR #5070, opened 2014-01-15
add check data valid in getStringFromFile

hbbalfred:
so far no check would be crashed when getData failed.
maybe return an empty string is better.


cocos/2d/platform/CCFileUtils.cpp
```diff
@@ -547,6 +547,8 @@ static Data getData(const std::string& filename, bool forString)
 std::string FileUtils::getStringFromFile(const std::string& filename)
 {
     Data data = getData(filename, true);
+    if (! data.getBytes())
```
dumganhar: Thanks. But using `data.isNull()` will better.

PR #3401, opened 2013-08-08
Take orientation into account when reporting accelerometer sensor readout

folecr:

PR #5071, opened 2014-01-15
issue #3401 physical lua binding

boyu0:
1. add more banding code for lua.
2. add lua physics slide, action and pump test.
3. fix some bugs.

### DIFF
--- a/cocos/physics/CCPhysicsContact.h
+++ b/cocos/physics/CCPhysicsContact.h
@@ -76,7 +76,7 @@ public:
     /** get contact shape B. */
     inline PhysicsShape* getShapeB() const { return _shapeB; }
     /** get contact data */
-    inline const PhysicsContactData& getContactData() const { return *_contactData; }
+    inline const PhysicsContactData* getContactData() const { return _contactData; }
     /** get data. */
     inline void* getData() const { return _data; }
     /**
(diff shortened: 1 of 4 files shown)


cocos/physics/CCPhysicsContact.h
```diff
@@ -76,7 +76,7 @@ class PhysicsContact : Event
     /** get contact shape B. */
     inline PhysicsShape* getShapeB() const { return _shapeB; }
     /** get contact data */
-    inline const PhysicsContactData* getContactData() const { return _contactData; }
+    inline const PhysicsContactData& getContactData() const { return *_contactData; }
```
dumganhar: Could `_contractData` be an invalid pointer?

dumganhar: Why to change the return value from pointer to reference?

edwardzhou:
I found PhysicsBody::setRotation is declared as protected. Why do not make it as public?

eg. In my game, I just only use Physics for collision checking. 
I need rotate the sprite by cc.RotateBy, but failed due to Node::setRotation do not and unable to call PhysicsBody::setRotation.

cocos/physics/CCPhysicsContact.h
boyu0: This data has generated when user can use it.
Change this to reference because lua auto generate mechanism only support struct reference return value, and because of this data won't be null when use can use it, change it to reference is a better idea.

PR #5075, opened 2014-01-16
Renderer perf fixes

ricardoquesada:
- Adds new command `BatchCommand`: 10% faster in `SpriteBatchNode`
- `QuadCommand` does not copy quads. It uses them as reference. 10% faster


dabingnn:
I add a large tilemap to test the performance of our quadcommand, it can only achieve 15fps on samsung I9100.
the link is https://github.com/cocos2d/cocos2d-x/pull/5068. the tilemap is a subclass of AtlasNode, so I think we need to optimize it. 

Besides this, there is another feature we need to support in batchCommand, draw part of texture Atlas instead of hole texture Atlas. The label may need this.

Thanks.

PR #5084, opened 2014-01-16
fixed #3683. fixed convertToWorldSpaceAR of CCSkin returning error coord...

zilongshanren:
This bug can be shown again in TestColliderDetector,the  convertToWorldSpaceAR method of CCSkin is return the wrong position.


dumganhar:
Thanks.

PR #5089, opened 2014-01-16
genbinding python detection

luisparravicini:
if PYTHON_BIN is not set, the default value used is /usr/bin
I've added an intermediate step and if PYTHON_BIN is not set, it tries to find python2.7 via "which" and it that fails it defaults to /usr/bin

Issue: http://www.cocos2d-x.org/issues/3732

### DIFF
--- a/tools/tolua/genbindings.sh
+++ b/tools/tolua/genbindings.sh
@@ -25,8 +25,13 @@ if [ -z "${CLANG_ROOT+aaa}" ]; then
 fi
 
 if [ -z "${PYTHON_BIN+aaa}" ]; then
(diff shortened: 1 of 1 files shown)


dumganhar:
Thanks. Could you also update it for `tools/tolua/genbindings.sh`?

luisparravicini:
Done

dumganhar:
Thanks. I will merge it soon.

dumganhar:
We're now using python to generate bindings.
Please refer to https://github.com/cocos2d/cocos2d-x/blob/develop/tools/tojs/genbindings.py#L35
I closed this PR.

PR #5090, opened 2014-01-16
Add emscripten support for Windows Host

martell:
inet.h include is not needed and breaks emscripten on windows.

Please merge asap :)


dumganhar:
Thanks.

PR #5091, opened 2014-01-16
warnings in CCLabelBMFont.cpp when compiling in release mode

luisparravicini:
There are two "unused variable" warnings when compiling in release mode as those variables are used only in a CCASSERT.

Issue: http://www.cocos2d-x.org/issues/3734


ricardoquesada:
thanks.

PR #5101, opened 2014-01-17
Header dependency fixes

ricardoquesada:
removes `#include "cocos2d.h"` from cocos2d files.

It **FORBIDDEN** to include `cocos2d.h` in cocos2d files. **FORBIDDEN**


cocos/2d/CCFont.h
```diff
@@ -28,7 +28,6 @@
 
 #include <string>
 
-#include "cocos2d.h"
```
dumganhar, line 31: Oops...  @Dhilan007

PR #5105, opened 2014-01-17
close #3678:fix incorrect spacing between characters

WenhaiLin:


dumganhar:
Which test case could reproduce this bug?

dumganhar:
Compilation fails.

PR #5123, opened 2014-01-18
GlobalZOrder is used for render priority...

ricardoquesada:
...and not vertexZ

Node::setGlobalZOrder() is used to change that.

Node::setZOrder() -> Node::setLocalZOrder();


dumganhar:
Hi @ricardoquesada ,
Since we have modified the API, getZOrder --> getLocalZOrder, and added s/getGlobalZOrder, could you please update release note(https://github.com/cocos2d/cocos2d-x/blob/develop/docs/RELEASE_NOTES.md) ?

ricardoquesada:
yes, sure, I will update them.

ricardoquesada:
done: https://github.com/cocos2d/cocos2d-x/pull/5170

although more info is still needed.

dumganhar:
Thanks :)

PR #5125, opened 2014-01-20
close #3765:Simplify label.Remove FontAtlasFactory,FontDefinition and TextImage.

WenhaiLin:


cocos/2d/CCFontAtlas.cpp
```diff
@@ -42,7 +41,7 @@
     _makeDistanceMap = _font->isDistanceFieldEnabled();
 
     FontFreeType* fontTTf = dynamic_cast<FontFreeType*>(_font);
-    if (fontTTf && fontTTf->isDynamicGlyphCollection())
```
dumganhar, line 45: `isDynamicGlyphCollection` isn't needed here?

WenhaiLin, line 45: Yes,I think  deal the specify glyph collection as same as dynamic glyph collection is much simpler.All glyph collection will deal with one way.

PR #5131, opened 2014-01-20
Modify Lua DeprecatedClass

Pisces000221:
It's very difficult to give a message when you access a deprecated variable (e.g. print 'CCHide is deprecated' when using CCHide), so I changed this file so that it won't print many messages on startup.

### DIFF
--- a/cocos/scripting/lua/script/DeprecatedClass.lua
+++ b/cocos/scripting/lua/script/DeprecatedClass.lua
@@ -1,2173 +1,268 @@
 -- This is the DeprecatedClass
-
-DeprecatedClass = {} or DeprecatedClass
-
---tip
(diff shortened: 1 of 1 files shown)


samuele3hu:
If I don't give a lot of messages on starup, some developers will feel uncertain when they use a deprecated variable.
Can you give me some suggestions?

Pisces000221:
If you just want to make them know that those classes are deprecated, it's okay but I think we do not need so much stars (*********). But after these messages appear, we can't see debug information like OpenGL details, etc.

If we want to give a message when the deprecated classes are used, maybe we have to use libs like LuaMacro to do that. There're no ways in pure Lua to do that.

PR #5132, opened 2014-01-20
remove Object::_autoReleaseCount

minggo:
- remove `Object::_autoReleaseCount`
- can create `AutoReleasePool` in stack

Don't merge it, i will add test case for it if the API looks good to you.


cocos/2d/CCDirector.cpp
```diff
@@ -193,9 +190,8 @@ bool Director::init(void)
     delete _renderer;
     delete _console;
 
-    // pop the autorelease pool
-    PoolManager::sharedPoolManager()->pop();
-    PoolManager::purgePoolManager();
+    // clean auto release pool
+    PoolManager::getInstance()->destroyInstance();
```
dumganhar: Use `PoolManager::destroyInstance()` to destroy singleton.

minggo: Good catch.

PR #5138, opened 2014-01-21
closed #2572, Amend the arithmetic of child's position in ParallaxNode::addChild().

natural-law:


dumganhar:
Thanks.

ricardoquesada:
Interesting. We need a test case for this bug.
We need to know which bug it is fixing, and also make sure that it is not breaking any other code.

natural-law:
The bug was feedback from developer: http://www.cocos2d-x.org/issues/2572

The appearance of the bug is：
When the ParallaxNode is moving, add a child into the node. Output the position of the child immediately (Mentioned it as "pos1").
A frame later, output the position of the child again (Mentioned it as "pos2").
The pos1 is very different with pos2. The developer and I think it's abnormal.

I found the reason is:
1. The position of the child is modified by ParallaxNode::addChild() like this:

<pre>
    Point pos = _position;
    pos.x = pos.x * ratio.x + offset.x;
    pos.y = pos.y * ratio.y + offset.y;
    child->setPosition(pos);
</pre>

2. Because of the ParallaxNode is moving, the position of the child is modified again by ParallaxNode::visit() like this:

<pre>
    PointObject *point = (PointObject*)_parallaxArray->arr[i];
    float x = -pos.x + pos.x * point->getRatio().x + point->getOffset().x;
    float y = -pos.y + pos.y * point->getRatio().y + point->getOffset().y;            
    point->getChild()->setPosition(Point(x,y));
</pre>

The two arithmetics are very different. I tried to respectively modify the arithmetic in method addChild() & visit().
I found if the arithmetic in method visit() is modified, the appearance of ParallaxNode Test in TestCpp is incorrect.
But it's OK when I modified the method addChild().

So I modified the arithmetic in method addChild().

ricardoquesada:
Ok. Thanks.
Could you add a test case for TestCpp ?
For the moment, that is the only way that we have to verify that a bug is fixed.
Thanks!

natural-law:
Ok, I will add it later.

ricardoquesada:
ok. thanks.

PR #5148, opened 2014-01-22
Mac platform simplified + fixes in EventDispatcher

ricardoquesada:
- Mac platform simplified
- and EventDispatcher uses float for global Z order

### DIFF
--- a/cocos/2d/CCEventDispatcher.cpp
+++ b/cocos/2d/CCEventDispatcher.cpp
@@ -234,7 +234,7 @@ void EventDispatcher::visitTarget(Node* node, bool isRootNode)
     
     if (isRootNode)
     {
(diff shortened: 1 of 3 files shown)


cocos/2d/CCEventDispatcher.cpp
```diff
@@ -234,7 +234,7 @@ void EventDispatcher::visitTarget(Node* node, bool isRootNode)
     
     if (isRootNode)
     {
-        std::vector<int> globalZOrders;
+        std::vector<float> globalZOrders;
```
dumganhar, line 237: Thank you. :)

ricardoquesada, line 237: no problem :-)

PR #5149, opened 2014-01-22
Label::getStringNumLines returns 0 instead of -1 when containing an empty string

Pisces000221:
It used to be -1, now it's 0 when containing an empty string. It only returns -1 when string is nullptr.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -772,12 +772,12 @@ int Label::getStringNumLines() const
 {
     int quantityOfLines = 1;
     
-    unsigned int stringLen = _currentUTF16String ? cc_wcslen(_currentUTF16String) : -1;
+    int stringLen = _currentUTF16String ? cc_wcslen(_currentUTF16String) : -1;
     if (stringLen == -1)
         return -1;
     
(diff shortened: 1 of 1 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -772,9 +772,9 @@ int Label::getStringNumLines() const
 {
     int quantityOfLines = 1;
     
-    unsigned int stringLen = _currentUTF16String ? cc_wcslen(_currentUTF16String) : 0;
-    if (stringLen == 0)
-        return (-1);
+    unsigned int stringLen = _currentUTF16String ? cc_wcslen(_currentUTF16String) : -1;
+    if (stringLen == -1)
```
WenhaiLin, line 776: The return value will be 1 of your modification when stringLen is 0.It's not expected results.

Pisces000221:
Sorry, I made a mistake. Unsigned 0 minus 1 equals 4294967295, not -1. It should work correctly now.

dumganhar:
@Dhilan007 
This issue was fixed in your Pull Request, right?

WenhaiLin:
@dumganhar right.

PR #5151, opened 2014-01-22
fix memory leak of autorelease pool

minggo:


cocos/base/CCAutoreleasePool.cpp
```diff
@@ -51,17 +49,17 @@ of this software and associated documentation files (the "Software"), to deal
...
-
-    CCASSERT(object->_reference > 1, "reference count should be greater than 1");
+    // only have one copy of Object*
+    _managedObjectArray.insert(object);
 }
 
 void AutoreleasePool::clear()
 {
-    if (!_managedObjectArray.empty())
+    for (const auto obj : _managedObjectArray)
```
dumganhar: const auto& obj

PR #5156, opened 2014-01-22
deprecated some functions in Object

minggo:
- Object::retainCount() -> Object::getReferenceCount()
- deprecated Object::isSingleReference(), it is needless, because we can use the result of Object::getReferenceCount() to compare with 1 to determine if it is single reference or not.


minggo:
I will mention it in release note.

PR #5159, opened 2014-01-22
Add test case for issue 2572

natural-law:


samples/Cpp/TestCpp/Classes/ParallaxTest/ParallaxTest.cpp
```diff
@@ -150,13 +150,92 @@ void Parallax2::onTouchesMoved(const std::vector<Touch*>& touches, Event  *event
...
+
+    // After a child added, output the position of the children 3 times.
+    // Bug : The first output is much different with the second one & the third one.
+    if (_childList.size() != _preListSize) {
+        switch (_printCount) {
+            case 0:
+            case 1:
+            case 2:
+                log( "--child count-- %zd", _childList.size());
+                for (auto obj : _childList)
```
dumganhar: Using `for (const auto& obj : _childList)` will be better.

PR #5161, opened 2014-01-22
[Win32] fix localStorageGetItem crash when column text is NULL.

pyrasis:
- sqlite3_column_text win32 version returns NULL pointer if not exist column.
- fix localStorageGetItem crash when column text is NULL.


dumganhar:
Thanks.

PR #5164, opened 2014-01-22
close #3642:fix bounding box of new label is incorrect.

WenhaiLin:


dumganhar:
Thanks.

minggo:
Great.
And please update the progress in corresponding thread in forum.
Thanks.

hanjukim:
@Dhilan007 Please see my new pull request
https://github.com/cocos2d/cocos2d-x/pull/5185
I think y position for new label is not fixed yet...

huangzhentao:
I agree with @hanjukim

WenhaiLin:
@hanjukim  I agree with you.I will fix it ASAP.We will merge your PQ.

PR #5167, opened 2014-01-22
Fixed x64 builds of testcpp with long to size_t changes

martell:
Same as before long is breaking x64 stuff so size_t is more suitable :)

Please Merge

Thanks
Martell


ricardoquesada:
thanks.

PR #5173, opened 2014-01-23
Don't set a default PYTHON_ROOT & NDK_ROOT for user on Windows

lihex:
Tips user if environment variables are note set.


dumganhar:
Thank you.

PR #5219, opened 2014-01-24
3.0 Beta2 HotFix: fix shader PositionTextureColorAlphaTest

dabingnn:


dumganhar:
The issue could be reproduced in TileMapTest. I merged it since it resolved the problem. Thanks.

PR #5223, opened 2014-01-26
closed #3545 fix chiniese character shown in LabelTest.cpp (WP8) 

squirrelChen:


samples/Cpp/TestCpp/Classes/LabelTest/LabelTest.cpp
```diff
@@ -1127,6 +1127,8 @@ string LabelTTFMultiline::subtitle()
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_WINRT)
     // Marker Felt is missing Chinese characters
     CCLabelTTF *pLable = CCLabelTTF::create("中国", "PMingLiU", 30);
+#elif (CC_TARGET_PLATFORM == CC_PLATFORM_WP8)
+	CCLabelTTF *pLable = CCLabelTTF::create("中国", "fonts/arialuni.ttf", 30);
```
minggo, line 1131: We don't want to test if ttf font can be displayed for this test case.

squirrelChen:
the commit 4b0f204 is to show Chinese using fonts in Resources\fonts at TextInputTest.cpp.

PR #5232, opened 2014-01-26
Update WebSocket.cpp

floatinghotpot:
bufix: when connecting fail, libwebsocket_client_connect() will return NULL, and onSocketCallback() will not be called, so we need trigger WS_MSG_TO_UITHREAD_ERROR here.


dumganhar:
issue http://cocos2d-x.org/issues/4101 was created. Thanks.

dumganhar:
Compilation error. please fix it. you could refer it to https://travis-ci.org/cocos2d/cocos2d-x/jobs/17635750

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/5415. Thanks.

PR #5250, opened 2014-01-27
closed #3867: "Testing empty labels" in LabelTest crashes on Linux platform.

dumganhar:


cocos/2d/platform/linux/CCImage.cpp
```diff
@@ -450,23 +450,21 @@ class BitmapDC
...
+		TextAlign alignMask/* = kAlignCenter*/,
+		const char * fontName/* = nil*/,
+		int size/* = 0*/)
 {
-	bool bRet = false;
+	bool ret = false;
 	do
     {
-		CC_BREAK_IF(! pText);
+		CC_BREAK_IF(!text || 0 == strlen(text));
```
dumganhar, line 463: If it's an empty string, just returns false.

PR #5257, opened 2014-01-27
closed #3871: event->stopPropagation can't work for EventListenerTouchAllAtOnce.  And add relevant test case.

dumganhar:


cocos/2d/CCEventDispatcher.cpp
```diff
@@ -815,7 +815,7 @@ void EventDispatcher::dispatchTouchEvent(EventTouch* event)
             if (event->isStopped())
             {
                 updateListeners(event);
-                return false;
+                return true;
```
dumganhar, line 818: This line fixes the issue.

PR #5258, opened 2014-01-27
[Jenkins] add create-template-project.py

lmskater:


tools/jenkins-scripts/create-template-project.py
```diff
@@ -0,0 +1,19 @@
+import os
+
+language = ['cpp','lua','javascript']
```
dumganhar: `language` --> `languages`?

dumganhar:
Thanks. merged.

PR #5263, opened 2014-01-27
closed #2920: fix bug: voice recognition input crash on ios7

boyu0:
1. remove temporary fix
2. add context change in doCaller.
3. add refreshInactiveText function and delete siri placeholder text at getText function.


cocos/2d/platform/ios/CCDirectorCaller.mm
```diff
@@ -92,6 +94,7 @@ -(void) setAnimationInterval:(double)intervalNew
                       
 -(void) doCaller: (id) sender
 {
+    [EAGLContext setCurrentContext: [[CCEAGLView sharedEGLView] context]];
```
minggo: @ricardoquesada 
This pull request wants to resolve a problem that, when using `EditBox` and touch voice input button. It will cause crash. Full information please refer to [this thread](http://www.cocos2d-x.org/forums/6/topics/36111). 

But i am not sure if invoke this function in every frame will cause problem or not. Could you please help to review it. Thanks.

ricardoquesada: no problem with that line.
It sets the correct opengl context before starting the main loop.

extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
```diff
@@ -414,25 +416,13 @@ - (void) textChanged
...
-    }
-    
     switch (inputMode)
     {
         case EditBox::InputMode::EMAIL_ADDRESS:
             _systemControl.textField.keyboardType = UIKeyboardTypeEmailAddress;
             break;
         case EditBox::InputMode::NUMERIC:
-            _systemControl.textField.keyboardType = (above7 ? UIKeyboardTypeDecimalPad : UIKeyboardTypeNumberPad);
+            _systemControl.textField.keyboardType = UIKeyboardTypeNumbersAndPunctuation;
```
dumganhar: I think `UIKeyboardTypeDecimalPad` is more suitable.

minggo:
@DanielBocksteger 
I am not sure. But the key point is that, should get GL context before drawing a new scene every frame. I think it is because since iOS7, UITextField uses GL to draw effects too. So we should get GL context for engine every frame.

Hope it helps.

PR #5264, opened 2014-01-27
Updated CCImage.cpp to resolve failure to load custom fonts from ttf file on windows

odedsh:
The bug is described here:
http://www.cocos2d-x.org/forums/6/topcis/32189

Basically the windows only singelton BitmapDC is used by all image objects
when drawing text. The call to BitmapDC.setfont is expected to load the
font from a ttf file and then use that font.

However the call to AddFontResource was incorrectly inside a block that is
only called when the previously loaded font is being removed. As a result
the ttf file is not added and the next call to 'CreateFontIndirect' would
fail to find the font.

The change I made was to move the code that actually handles clearing the
existing font to its own private method: BitmapDC::removeCustomFont. This
function is called by the BitmapDC destructor and by setFont before
loading a new font.

Now the actual font creation will always occur after the old font is
removed.

NOTE:
I'm very new to cocos2d-x and I'm currently working on a project with
v2.2.1 when I encountered this bug. I did some minimal testing of my fix
but I am not set up to do extensive regression tests on v3.

Additional Issues with the code I can see but did not address:
1) The windows API states that the WM_FONTCHANGE should be broadcast to all
windows and not sent to the current window as is done in the actual code.
I did not touch since the code currently seems working, and I am not sure
if this is an oversight or by design.

2) Failures to load a font from TTF file are not reported. So passing an
incorrect path to a ttf file will silently fail although
AddFontResource will create a valide error message.

3) Windows loads fonts in two stages. first AddFontResource accepts a path
to a font file and loads it. Then CreateFontIndirect will pass a font name
and load it from the first matching resource.

With the current code if the user uses a path to a ttf file for instance
`fonts/XXXX.ttf` The code assumes that XXXX is the name of the font inside
the file. But there is no gurantee that this is the case. In the samples
supplied there is a font called `Marker Felt.ttf` however the font in the
file is called `Felt` as a result it is impossible to load this file in
windows. The code will load the ttf file but proceed to try and load a
resource called `Marker Felt` when it should be loading `Felt`.
(Renaming the file to Felt.ttf solves this issue)

4) The current code will always keep only a single font resource. So using
a second font will cause the system to destroy the old font first. The
expectation of the API is that fonts will be loaded once and then reused.
However I'm not aware of any actual performance impact currently.

### DIFF
--- a/cocos/2d/platform/win32/CCImage.cpp
+++ b/cocos/2d/platform/win32/CCImage.cpp
@@ -55,24 +55,7 @@ public:
         {
             DeleteDC(_DC);
         }
-        HFONT hDefFont = (HFONT)GetStockObject(DEFAULT_GUI_FONT);
-        if (hDefFont != _font)
-        {
-            DeleteObject(_font);
-            _font = hDefFont;
-        }
-		// release temp font resource	
-		if (_curFontPath.size() > 0)
-		{
-			wchar_t * pwszBuffer = utf8ToUtf16(_curFontPath);
-			if (pwszBuffer)
-			{
-				RemoveFontResource(pwszBuffer);
-				SendMessage( _wnd, WM_FONTCHANGE, 0, 0);
-				delete [] pwszBuffer;
-				pwszBuffer = NULL;
-			}
-		}
+		removeCustomFont();
     }
 
 	wchar_t * utf8ToUtf16(std::string nString)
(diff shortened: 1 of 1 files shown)


dumganhar:
The page(http://www.cocos2d-x.org/forums/6/topcis/32189) you link to can't be opened.

dumganhar:
@odedsh ,thanks for your contribution.
Could you please add a test case for testing this issue?

odedsh:
This is the proper link: http://www.cocos2d-x.org/forums/6/topics/32189 

I need to think about how to test the issue since there is no failure just drawing with the wrong font.. I'll give it a try

odedsh:
Ok. I did some investigating. And unfortunately I can't add a test case to display this issue.

In fact this issue will only occur with the first LabelTTF object created with a custom font. After the first one all additional objects will be created correctly. 

The current TestBed in TestCpp creates lots of LabelTTF with "Arial" font, A test case for this particular issue will have to be a standalone executable..

cocosQA:
Dear odedsh:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6464

PR #5265, opened 2014-01-28
Fix ControlExtension touch when parent is not visiable

Yehsam23:
Fix ControlExtension touch when parent is not visiable

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -164,7 +164,13 @@ Node::~Node()
     CC_SAFE_DELETE(_componentContainer);
     
 #if CC_USE_PHYSICS
-    CC_SAFE_RELEASE(_physicsBody);
+    if (_physicsBody != nullptr)
+    {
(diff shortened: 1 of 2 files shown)


tools/bindings-generator
```diff
@@ -1 +0,0 @@
-Subproject commit 186badf9c635f20e7c2ecd3bd9815c21140b4b0b
```
dumganhar, line 1: Why to delete the submodule?

Yehsam23, line 1: Sorry~~i don't know why!!!
It delete self when i commit!!

I close this

cocos/2d/CCNode.cpp
```diff
@@ -1325,6 +1336,22 @@ bool Node::updatePhysicsTransform()
     
     return false;
 }
+
+void Node::updatePhysicsPosition()
```
boyu0, line 1340: What this function for? I didn't find any place to use it in your PR.

PR #5266, opened 2014-01-28
lua binding to wp8 and winrt

squirrelChen:
@Stammen I add lua to wp8 and winrt(vs2012). And I fix some bugs in some test cases. But I cannot fix it now, such as TableViewTest and ScrollViewTest, may be fix at next pr.


dumganhar:
@Stammen, @martell,
What do you think of this Pull Request? Could you please review it? Thanks.

stammen:
Why were the files CCAccelerometer.cpp, CCApplication.cpp, etc copied from the platform/winrt folder into the platform/WP8 folder? The same files work on both WinRT and WP8.

squirrelChen:
@stammen 
 I had removed duplicated files in the folder of wp8 and add reference to winrt. Any opnions? Thanks.

stammen:
Thanks for the update. I will continue with the review.

dumganhar:
@stammen 
Any comments?

codelegger:
I am getting a random crash on Itable.c
gkey(mp)->value = key->value; gkey(mp)->tt = key->tt;

Since it is a random Access violation exception is this something related that i am testing it on windows phone emulator ?

PR #5269, opened 2014-01-28
Better handling of invalid shaders.

AIGRIND-LLC:
If shader is invalid the programm will be crashed.
The better way is to rise an assert and write message to the error log.
Platform: Windows
Explain: Functions in glew32.dll have non-standard call conversion. Typecast (GLInfoFunction) is invalid.

### DIFF
--- a/cocos/2d/CCGLProgram.cpp
+++ b/cocos/2d/CCGLProgram.cpp
@@ -298,17 +298,17 @@ void GLProgram::use()
     GL::useProgram(_program);
 }
 
-std::string GLProgram::logForOpenGLObject(GLuint object, GLInfoFunction infoFunc, GLLogFunction logFunc) const
+std::string GLProgram::logForOpenGLShader(GLuint object) const
 {
     std::string ret;
     GLint logLength = 0, charsWritten = 0;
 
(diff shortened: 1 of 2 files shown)


dumganhar:
> Functions in glew32.dll have non-standard call conversion. Typecast (GLInfoFunction) is invalid.
> But it was built successfully on Win32 without this patch.

How to reproduce the issue?

AIGRIND-LLC:
1. First you need to make some invalid shader. For example you may corrupt existed shader (%cocos2dx%\samples\Cpp\TestCpp\Resources\Shaders\example_Monjori.fsh). 
2. Next run TestCpp sample and open "ShaderTest" menu.
3. If there is no our fix the application will crash without any log message.
   If you apply our fix the shader content and error message will be pushed into the log.

dumganhar:
OK, I will check it .Thanks.

cocosQA:
Dear AIGRIND-LLC:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6463

PR #5271, opened 2014-01-31
GlView improvements

ricardoquesada:
- `EGLView` no longer is a singleton
-  supports fullscreen mode as a parameter
- `EGLView` is renamed to `GLView`


jonseg:
Hi Ricardo,

Is it possible you have to update the test game? 

Exactly these files: 
- https://github.com/cocos2d/cocos2d-x/blob/develop/template/multi-platform-cpp/proj.ios_mac/ios/AppController.mm 
- https://github.com/cocos2d/cocos2d-x/blob/develop/template/multi-platform-cpp/proj.ios_mac/ios/RootViewController.mm 

Instead of importing "EAGLView.h", should be "CCEAGLView.h"? 

Now returns this error: "EAGLView.h file not found" on these lines.

Thank you very much.

ricardoquesada:
@jonseg 
fixed: https://github.com/cocos2d/cocos2d-x/pull/5288
thanks.

PR #5279, opened 2014-01-31
Added getPropertiesForGID that assings value instead of returning it.

edba:
If you are looping through all the tiles in a tilemap to get their properties. The original getPropertiesForGID  function will create an empty Value object if the tile has no properties at all. This can be really inefficient if you are looping through a map that mostly has tiles with no properties.

The new function will allow you to skip the empty object creation (this saves a lot of time) giving you the option to send as an argument the variable you want the value to be assing to. An example of how the new function would be use:

``` cpp
Value* tilePropertiesValue = NULL;
if (tileMap->getPropertiesForGID(gid, &tilePropertiesValue )) {
    ValueMap tileProperties = tilePropertiesValue->asValueMap();
    //do what you need with the tile properties
} else {
    //handle no properties on tile
}
```

### DIFF
--- a/cocos/2d/CCTMXTiledMap.cpp
+++ b/cocos/2d/CCTMXTiledMap.cpp
@@ -246,6 +246,16 @@ Value TMXTiledMap::getPropertiesForGID(int GID) const
     return Value();
 }
 
+bool TMXTiledMap::getPropertiesForGID(int GID, Value** value)
+{
+    if (_tileProperties.find(GID) != _tileProperties.end()) {
+        *value = &_tileProperties.at(GID);
+        return true;
+    } else {
+        return false;
+    }
+}
+
 std::string TMXTiledMap::getDescription() const
 {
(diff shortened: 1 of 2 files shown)


dumganhar:
@edba 
We could return `Value::Null` instead.
And `getPropertiesForGID` returns `const Value&` instead of `Value`.
Then, no extra `Value` will be created.

dumganhar:
So the following will be able to work.

``` c++
const Value& tilePropertiesValue = tileMap->getPropertiesForGID(gid);
if (!tilePropertiesValue.isNull()) {
    ValueMap& tileProperties = tilePropertiesValue->asValueMap();
    //do what you need with the tile properties
} else {
    //handle no properties on tile
}
```

dumganhar:
The problem is that this method uses the same name as `Value getPropertiesXXX`.
JS and Lua bindings will not be able to know how to deal with `Value**`.

dumganhar:
New PR to fix efficient problem: https://github.com/cocos2d/cocos2d-x/pull/5390

@ricardoquesada , what's your opinion?

PR #5282, opened 2014-02-01
issue #3828: Icon resources should be packaged. 

dumganhar:
issue #3828: Icon resources should be packaged. Only XXX_info.plist should not be packaged. If it was packaged, it would cause problem of running on simulator.


ricardoquesada:
Thanks!

PR #5283, opened 2014-02-01
Fixed crashing when calling CCDirector's end().

seemk:
In the win32 application, calling CCDirector's end (e.g. the close button in the startup project) method caused a segfault. The director outside of Application::run's while loop was referenced after it had deleted itself, trying to do another cleanup. The second cleanup is now avoided, but a cleanup is still done if the window is closed manually.


dumganhar:
Thanks. I will check it.

dumganhar:
Issue http://cocos2d-x.org/issues/4099 was created. Thanks.

dumganhar:
Thanks.

PR #5284, opened 2014-02-02
Fixed bug with custom property for CCNodeLoader

zifter:
NodeLoaderLibrary holds a unique instance of the loader, we working with this instance. Accordingly, the custom parameters that are stored in the dictionary (m_pCustomProperties ) are stored in this instance. When continue to work on the loading of another node, the NodeLoader will be have the parameters of the other nodes. It's just a fix for this situation. The library returns a clean NodeLoader.


dumganhar:
Hi @zifter 
This issue was fixed at https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/editor-support/cocosbuilder/CCNodeLoader.cpp#L34 .
Did you pull the latest codes?

zifter:
I'm using old source code, sorry.
Thank you for reply!

PR #5286, opened 2014-02-02
fix timeout getting status of sockets in Console::loop()

maksqwe:
'016000' - is octal constants, 016000 == 7168
http://msdn.microsoft.com/en-us/library/00a1awxf.aspx


ricardoquesada:
thanks.

PR #5289, opened 2014-02-03
fix string size check and assert condition

maksqwe:
1. std::string::size() return size_t (is never < 0)
2. "event must be larger than 0" and 'event' has unsigned int type


dumganhar:
Merged , thanks.

PR #5300, opened 2014-02-08
closed #3900: Upgrades spidermonkey to Firefox v27.

dumganhar:


azmeuk:
What options do you pass to `configure` to compile spidermonkey ? With defaults parameter the generated `libjs_static.a` is about 500Mb.

azmeuk:
Nevermind, `strip` command reduces size to 20Mb.

PR #5302, opened 2014-02-08
Improve new label's formatting when using both CJK and ASCII

Pisces000221:
When you use the new label with both Chinese/Japanese/Korean(CJK) characters and non-CJK characters like ASCII characters, you'll find them in the wrong place when line-breaking (e.g. Chinese characters won't go to the next line). I also tried setLineBreakWithoutSpace but it doesn't work correctly.
So I did this and made it work well.

screenshot


dumganhar:
Please add test case for this issue. Thanks.

dumganhar:
Thanks.

PR #4333, opened 2013-11-19
bug of ccbreader load file in jsb

hbbalfred:
owner can equal null when ccbreader load a file with parentSize as follow

cc.BuilderReader.load( path, null, parentSize );

PR #5305, opened 2014-02-08
remove the redundant check when load a file in ccbreader

hbbalfred:
https://github.com/cocos2d/cocos2d-x/pull/4333


dumganhar:
Thanks.

PR #5306, opened 2014-02-08
Ignore files generated by cmake

azmeuk:
After running `cmake . && make`, repository is polluted with some files generated by cmake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,8 @@ tags
 /lib
 /build/linux-build
 
+# Cmake files
+CMakeCache.txt
+CMakeFiles
+Makefile
+cmake_install.cmake


.gitignore
```diff
@@ -105,3 +105,8 @@ tags
 /lib
 /build/linux-build
 
+# Cmake files
+CMakeCache.txt
+CMakeFiles
+Makefile
```
dumganhar, line 111: Hi, Is it reasonable to add `Makefile` to ignore list?

azmeuk, line 111: Well, now those files are generated by cmake, so I suppose you don't want them to be versionned. Actually there is no more Makefile in the repository, is it ? Projects that use autotools generally ignore Makefile too.
If Makefile were versionned, then a new user that would clone and build the repository may have an unnecessary conflict on Makefiles.

dumganhar:
Thanks.

PR #5309, opened 2014-02-10
use python script instead for auto generating binding codes

lihex:


tools/tolua/genbindings.py
```diff
@@ -0,0 +1,148 @@
...
+
+    conf_ini_file = os.path.abspath(os.path.join(os.path.dirname(__file__), 'userconf.ini')) 
+
+    print 'generating userconf.ini...'
+    with open(conf_ini_file, 'w') as configfile:
+      config.write(configfile)
+
+
+    # set proper environment variables
+    if 'linux' in platform or platform == 'darwin':
```
dumganhar, line 116: Why not just use `platform == 'linux'`?

lihex, line 116: since sys.platform returns 'linux2' not 'linux' on linux

dumganhar, line 116: ok

tools/tojs/genbindings.py
```diff
@@ -0,0 +1,148 @@
...
+    ndk_root = re.sub(r"\"", "", ndk_root)
+
+    platform = sys.platform
+    if platform == 'win32':
+        cur_platform = 'windows'
+    elif platform == 'darwin':
+        cur_platform = platform
+    elif 'linux' in platform:
+        cur_platform = 'linux'
+    else:
```
litefeel, line 74: no cygwin?

dumganhar, line 74: Yep, it isn't needed.

PR #5316, opened 2014-02-10
Add performance test case on game scenario.

natural-law:


natural-law:
@darkdukey
I add a test case named "Scenario Performance Test" in PerformanceTest. 
The related issue is http://www.cocos2d-x.org/issues/3738.
Can you take a look at the test case? If the test case is not very fit the requirement of the issue, please feedback to me in this PR or the issue. Thanks!

darkdukey:
Good job, but i would suggest doing the following things
1. make the tilemap not moving by default
2. add buttons to add/remote particle system as well
3. adding 5 sprites a time are not likely to causing any performance hit, maybe add 500 sprite a time?
4. have a default starting scenario(2000 sprites, 20 particle system, 2 layers of tilemaps, running at 60 on iphone5), the goal for this test is to make sure the engine performance in a real game scenario.

natural-law:
@darkdukey 
I agree with you. I will improve the test case.

dumganhar:
@darkdukey 
@natural-law has updated this PR.Does it meet your demand? If no problem, I will merge it. Thanks.

darkdukey:
Good job, It's really good.

PR #5320, opened 2014-02-10
Android Flickering Bug Fix

seagullua:
When the new Activity is launched the screen starts flickering (Two nearest frames are replacing each other all the time). This fix stops flickering. 

More people with this issue: http://www.cocos2d-x.org/forums/6/topics/34686

This fix uses Android `GLSurfaceView` method `setRenderMode` to pause rendering while the application is in the background.


dumganhar:
Thanks.

basconje:
thx its work ))

PR #5326, opened 2014-02-11
move labelttf's platform-dependent code from CCImage to CCDevice

WenhaiLin:


dumganhar:
@Dhilan007 ,Great work.
I will review soon. :)

cocos/2d/CCTexture2D.cpp
```diff
@@ -1039,28 +1040,30 @@ bool Texture2D::initWithString(const char *text, const char *fontName, float fon
...
+    if(!text || 0 == strlen(text))
+        return false;
 #if CC_ENABLE_CACHE_TEXTURE_DATA
     // cache the texture data
     VolatileTextureMgr::addStringTexture(this, text, textDefinition);
 #endif
 
     bool ret = false;
-    Image::TextAlign align;
+    Device::TextAlign align;
```
dumganhar, line 1051: Why should we have Device::TextAlign and TextHAlignment::XXX? Could we just use which in FontDefinition?

cocos/2d/platform/CCDevice.h
```diff
@@ -49,6 +64,8 @@ class CC_DLL Device
      */
     static void setAccelerometerInterval(float interval);
 
+    static Data getTextureDataForText(const char * text,const FontDefinition& textDefinition,TextAlign align,int &widht,int &height);
```
dumganhar, line 67: How about using `const std::string&` for `text` argument?

cocos/2d/platform/linux/CCDevice.cpp
```diff
@@ -30,6 +30,51 @@ of this software and associated documentation files (the "Software"), to deal
...
+#include <fontconfig/fontconfig.h>
+#include "platform/CCFileUtils.h"
+
+#include "ft2build.h"
+#include FT_FREETYPE_H
+
+using namespace std;
+
+// as FcFontMatch is quite an expensive call, cache the results of getFontFile
+static std::map<std::string, std::string> fontCache;
```
dumganhar, line 47: Please use `std::unordered_map` instead of `std::map`. It will be faster.

cocos/2d/platform/linux/CCDevice.cpp
```diff
@@ -67,6 +112,392 @@ void Device::setAccelerometerInterval(float interval)
...
+		}
+
+		// top alignment
+		return baseLinePos;
+	}
+
+    std::string getFontFile(const char* family_name) {
+    	std::string fontPath = family_name;
+
+    	std::map<std::string, std::string>::iterator it = fontCache.find(family_name);
```
dumganhar, line 328: Using `auto` for iterator.

dumganhar:
Thanks.

cocos/2d/platform/CCDevice.h
WenhaiLin, line 67: Texture2D::initWithString  and implementations in Device.cpp using "const char\* ".It will cause type conversion more than once if using const std::string& for text argument here.

dumganhar, line 67: So why doesn't `Texture2D::initWithString` access `const std::string&`?

PR #5328, opened 2014-02-11
closed #3990: ParticleSystemQuad::setTotalParticles() can't set a value larger than the initialized value.

dumganhar:


cocos/2d/CCParticleSystemQuad.cpp
```diff
@@ -506,6 +505,10 @@ void ParticleSystemQuad::setTotalParticles(int tp)
         {
             setupVBO();
         }
+        
+        // fixed http://www.cocos2d-x.org/issues/3990
+        // Updates texture coords.
+        setTexture(_texture);
```
dumganhar: @ricardoquesada 
What do you think of this patch?
`setTexture(_texture)` is needed when `setTotalParticles` a larger number.
It will invoke `initTexCoordsWithRect` to reset `texture cords`, otherwise the texture cords element in `qual`will be null. And nothing will be shown.

ricardoquesada: Thanks James.

I still don't understand why it is not working, and why it works if you call `setTexture`.

The way I see it is that allocating more particles shouldn't break the texture. So, I would fix that bug (fix in the place that the texture breaks), instead of adding this workaround.

Thanks.

dumganhar: Hi, Riq,
Notice that `memset(_quads, 0, quadsSize);` will clear the `texture cords` in `_qual`. 

``` cpp
void ParticleSystemQuad::setTextureWithRect(Texture2D *texture, const Rect& rect)
{
    // Only update the texture if is different from the current one
    if( !_texture || texture->getName() != _texture->getName() )
    {
        ParticleSystem::setTexture(texture);
    }

    this->initTexCoordsWithRect(rect); // Here reset the tex cords.
}

void ParticleSystemQuad::setTexture(Texture2D* texture)
{
    const Size& s = texture->getContentSize();
    this->setTextureWithRect(texture, Rect(0, 0, s.width, s.height));
}
```

Maybe add a method called `updateTexCoords()` will be better.

ricardoquesada: yes, sounds good.

dumganhar: OK, I will update this PR by adding `ParticleSystemQuad::updateTexCoords`. :)

dumganhar:
@ricardoquesada ,
I updated this PR. Please review again. Thanks.

ricardoquesada:
looks good!

willwl:
it's drawing, but look like the particles still the same effect before, no more particles show after set, just like from 40 to 400, there is nothing effect

willwl:
it's ok, this line lost in 2d-x 2.2
m_fEmissionRate = m_uTotalParticles / m_fLife;

dumganhar:
@willwl 
Yes, we only fix this bug in develop branch.

PR #5329, opened 2014-02-11
GLFW include path for Fedora

azmeuk:
Hi,
This line allows Fedora users to not build GLFW themselves and use the glfw-devel package, and should not affect other distributions.


dumganhar:
Thanks.

PR #5335, opened 2014-02-12
change difference designresolution as user-setting

squirrelChen:
@stammen I fixed to use existing resolution policy, not maintained the kResolutionShowAll. Is it need to keep kResolutionShowAll, any reason? Or is it just a bug? I modified it and test ok. Wait for your reply, Thanks.


stammen:
This change should be fine. Go ahead and merge the change.

PR #5336, opened 2014-02-12
Add member function of removing component By component pointer in CCNode class

chengstory:


cocos2dx/support/component/CCComponentContainer.cpp
```diff
@@ -104,6 +104,30 @@ bool CCComponentContainer::remove(const char *pName)
...
+        CCDictElement *tmp = NULL;
+        HASH_ITER(hh, m_pComponents->m_pElements, pElement, tmp)
+        {
+            if (pElement->getObject() == pCom)
+            {
+                pCom->onExit();
+                pCom->setOwner(NULL);
+                HASH_DEL(m_pComponents->m_pElements, pElement);
+                pElement->getObject()->release();
+                CC_SAFE_DELETE(pElement);
```
dumganhar, line 123: Should break here if object was found?

PR #5343, opened 2014-02-13
add spine manual binding;

ucchen:
add spine manual binding, which support setAnimationListener, findBone, findSlot and so on. but i have no idea how to bind void\* to jsval.


dumganhar:
Hi, it's impossible to bind `void*`, where did spine use `void*`? Is the interface needed to be exposed to JS?

ucchen:
Spine run times have three struct use void\* which spAnimationStateData, spTimeline and spAttachment. But those struct aren't often use. even though haven't bind void*, spine manual bind work well.

ucchen:
Actually, void\* it's not necessary bind to JS

PR #5350, opened 2014-02-13
-fixed formating and small optimizations

bagobor:


bagobor:
Fixed formating and small optimizations.
autorelease replaced with release after placing action into container.

PR #5351, opened 2014-02-13
share xcode build schemes to support xcodebuild command line build

heliclei:


dumganhar:
@heliclei 
The ios-build is still failed.

heliclei:
@dumganhar 
The scheme name has changed, thanks~

PR #5362, opened 2014-02-14
fix copy-paste error, all spxxx_to_jsval use lower case;

ucchen:
Spine manual binding; fix copy-paste error, all spxxx_to_jsval use lower case and add missing structs getAttachment;


dumganhar:
Thanks. @ucchen

PR #5365, opened 2014-02-16
Solve the bug of Image::saveImageToPNG()

natural-law:
The method Image::saveImageToPNG() can't write png files correctly on platforms without iOS.


dumganhar:
@dabingnn
What's your opinion of this PR?

dumganhar:
Thanks.

PR #5366, opened 2014-02-17
use release libs and remove debug libs

squirrelChen:
http://www.cocos2d-x.org/issues/4038


squirrelChen:
@minggo @dumganhar @stammen I reduce the size from 137.4M to 84.8M.

dumganhar:
@koowolf 
Did you build the release libraries by using the same version of source codes as before?

squirrelChen:
The release libs existed, needn't to compile again. Used release libs in debug and release mode.

stammen:
Did you check if all of the projects build for Debug/Release for Win32/ARM? WinRT and Phone?

squirrelChen:
Yes, It all tests OK.

stammen:
Excellent!

PR #5371, opened 2014-02-17
Refactor #3620 Merge cocostudio/CCActionXxx to CCAction

zilongshanren:
This PR seems like a huge monster, but it did do a few related stuff.
1. move CCActionEaseEx and CCTweenFunction from cocostudio to the core engine
2. refactor all static member function of CCTweenFunction to free function with a namespace `tweenfunc`
3. extract all of the math calculation from CCActionEase to CCTweenFunction.h and simply the code of all the ease actions
4.add full tests to all the newly added ease actions
5.changes namespace of CCActionEaseEx and CCTweenFunction from cocostudio to cocos2d
6. fix all compile problems of the refactor

### DIFF
--- a/cocos/editor-support/cocostudio/CCTween.cpp
+++ b/cocos/editor-support/cocostudio/CCTween.cpp
@@ -476,7 +476,7 @@ float Tween::updateFrameData(float currentPercent)
      *  If frame tween easing equal to TWEEN_EASING_MAX, then it will not do tween.
      */
     TweenType tweenType = (_frameTweenEasing != Linear) ? _frameTweenEasing : _tweenEasing;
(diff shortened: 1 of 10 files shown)


cocos/editor-support/cocostudio/CCTween.cpp
```diff
@@ -281,7 +282,7 @@ void Tween::updateHandler()
         percent = updateFrameData(percent);
     }
 
-    if(_frameTweenEasing != TWEEN_EASING_MAX)
+    if(_frameTweenEasing != ::cocos2d::tweenfunc::TWEEN_EASING_MAX)
```
dumganhar, line 285: Why should add `::` before `cocos2d`?

zilongshanren, line 285: already fixed

PR #5374, opened 2014-02-17
fix string concatenation

justmao945:
`set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-std=c++11")` which leads to list of strings.
The result is "-stdlib=libc++;-std=c++11"  if CMAKE_CXX_FLAGS is defined as "-stdlib=libc++"


dumganhar:
Thanks.

PR #5378, opened 2014-02-18
closed #4060:fixed Label wasn't implemented correctly when Label::_fontAtlas contains multiple texture.

WenhaiLin:


cocos/2d/CCLabel.cpp
```diff
@@ -430,9 +431,23 @@ float Label::getScaleX() const
 
 void Label::alignText()
 {
-    if(_textureAtlas)
-        _textureAtlas->removeAllQuads();  
+    for (auto batchNode:_batchNodes)
```
dumganhar: Please use `const auto& batchNode: xxx`

cocos/2d/CCLabel.cpp
```diff
@@ -661,7 +677,10 @@ void Label::onDraw()
 
     GL::blendFunc( _blendFunc.src, _blendFunc.dst );
 
-    _textureAtlas->drawQuads();
+    for (auto batchNode:_batchNodes)
```
dumganhar: The same here.

dumganhar:
Thanks.

PR #5382, opened 2014-02-18
issue #4039: fix bug: tilemap <object> x, y, width and height value parse incorrect.

boyu0:


tests/test-cpp/Classes/TileMapTest/TileMapTest.cpp
```diff
@@ -777,17 +778,17 @@ void TMXOrthoObjectsTest::onDraw()
...
         ValueMap& dict = obj.asValueMap();
         
-        float x = dict["x"].asFloat();
-        float y = dict["y"].asFloat();
-        float width = dict["width"].asFloat();
-        float height = dict["height"].asFloat();
+        float x = dict["position_x"].asFloat();
+        float y = dict["position_y"].asFloat();
+        float width = dict["size_width"].asFloat();
+        float height = dict["size_height"].asFloat();
```
dumganhar: How could developer distinguish that `x`  is original value, and `position_x` is converted value?

dumganhar: Better name?

PR #5384, opened 2014-02-18
CCImage.cpp: initWithS3TCData() - Allow non-mipmapped textures

wefiends:
Currently, s3tc compressed textures with no mipmaps fail to load due to missing the appropriate PixelFormat, since it is only set when loading mipmaps. 

This patch sets the pixel format before mipmap loading (if s3tc hardware support is present).

### DIFF
--- a/cocos/2d/platform/CCImage.cpp
+++ b/cocos/2d/platform/CCImage.cpp
@@ -1584,11 +1584,11 @@ bool Image::initWithS3TCData(const unsigned char * data, ssize_t dataLen)
     
     _width = header->ddsd.width;
     _height = header->ddsd.height;
-    _numberOfMipmaps = header->ddsd.DUMMYUNIONNAMEN2.mipMapCount;
+    _numberOfMipmaps = MAX(1, header->ddsd.DUMMYUNIONNAMEN2.mipMapCount); //if dds header reports 0 mipmaps, set to 1 to force correct software decoding (if needed).
     _dataLen = 0;
     int blockSize = (FOURCC_DXT1 == header->ddsd.DUMMYUNIONNAMEN4.ddpfPixelFormat.fourCC) ? 8 : 16;
     
-    /* caculate the dataLen */
+    /* calculate the dataLen */
     
     int width = _width;
     int height = _height;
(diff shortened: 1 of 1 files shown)


dumganhar:
@boyu0, what's your opinion about this PR?

boyu0:
I think `_numberOfMipmaps` value is equal to 1 when s3tc has no mipmaps. it won't be less than 1. But I'm not sure of this, I didn't find any explanation in internet.
But if so, I think set `_numberOfMipmaps` to 1 is a better idea.

wefiends:
DXT1/5 dds files with mip maps disabled that are created with the Squish library (used by Nvidia SDK) or Crunch have their mipmaps set to zero (which is why they failed to load).

Since the PixelFormat is per Image/Texture, is there any reason to reassign _renderFormat for every mipmap?

boyu0:
@wefiends Thanks for reply.
No, it's not. But if you want move this outside of `for` loop, I think it's better to add an `else` part and move `_renderFormat = Texture2D::PixelFormat::RGBA8888;` outside too.
Set `_numberOfMipmaps` to 1 can force program go inside `for` loop once, it can go to `else` part and decode image data for which device doesn't support s3tc format. I think `_numberOfMipmaps` is equal to 0 or 1 make no difference, and we deal with them as the same in `Texture2D`.

wefiends:
I did what you suggested. 
- Moved the `_renderFormat` assign for software decode out of the `for` loop.
- `_numberOfMipmaps` is forced to 1 if the DDS header reports 0. Needed for software decode. I think this is the correct behavior anyways.

Tested and Working (DXT5 DDS with mipmaps disabled):
- Mac OS 10.9 (gl.supports_S3TC: true)
- iPhone 4S - iOS 7 (gl.supports_S3TC: false)
- iPad 3rd Gen - iOS 6.1.3 (gl.supports_S3TC: false)

You can download the texture I used [here](https://dl.dropboxusercontent.com/u/39502/hud.dds.gz).

boyu0:
Great! Thanks very much!
@dumganhar I have tested it, I think it's ready for merge.
@wefiends I want add a test case for this bug in TestCpp, and I want use the image your provide, Is that ok for you?

dumganhar:
Thanks.

wefiends:
@boyu0 Sure, go ahead. Thanks for your help.

boyu0:
The PR add this test case is: https://github.com/cocos2d/cocos2d-x/pull/5414
@wefiends Thanks for the image :)

PR #5386, opened 2014-02-19
add more console commands: resume, pause, stopanimation and startanimation

heliclei:


ricardoquesada:
thanks. What about merging all the 4 new commands into just one. eg:

``` sh
director start
director stop
director pause
director resume
```

heliclei:
That's great! I'll merge them into one.

dumganhar:
Thanks.

PR #5388, opened 2014-02-19
Develop refresh texture

dabingnn:


cocos/2d/CCTextureCache.h
```diff
@@ -270,6 +270,32 @@ class VolatileTextureMgr
 
 #endif
 
+#if CC_ENABLE_IMAGE_FILE_TEXTURE_RELOAD
```
minggo, line 273: Why need this macro?

dabingnn, line 273: This feature is needed by editor to refresh textures on the fly. It is not needed by the real game. @dumganhar suggest that we define a macro to enable or disable this feature.

cocos/2d/CCTextureCache.cpp
```diff
@@ -690,5 +696,88 @@ void VolatileTextureMgr::reloadAllTextures()
...
+        tt->initWithImage(image);
+        Texture2D::setDefaultAlphaPixelFormat(oldPixelFormat);
+    }
+    
+    CC_SAFE_RELEASE(image);
+}
+
+void ImageFileTextureReloader::reloadAllTextures()
+{
+    for(auto iter : _imageFileTextures)
```
dumganhar, line 774: please use `const auto&` for range loop.

minggo:
I think it only needs to add one function named `TextureCache::reloadTexture(const std::string &filename)`. And the logic is that

``` cpp
bool TextureCache::reloadTexture(const string& filename)
{
if file is not loaded yet
     invoke TextureCache::addImage();
else
     get Texture2D instance by filename
     re-init Texture2D
}
```

PR #5393, opened 2014-02-19
add console autotest support

heliclei:


tests/test-cpp/Classes/controller.cpp
```diff
@@ -230,3 +231,36 @@ void TestController::onMouseScroll(Event *event)
...
+    auto _console = Director::getInstance()->getConsole();
+    
+    static struct Console::Command autotest = {
+        "autotest", 
+        "testcpp auto test command", 
+        [](int fd, const std::string& args) 
+        {
+            for(int i = 0; i < g_testCount; i++)
+            {
+                if(args.compare(g_aTestNames[i].test_name)== 0)
```
dumganhar: Why not  use `==` directly?

heliclei: thanks~, I have changed cumbersome compare() to '==' operator

PR #5395, opened 2014-02-19
issue #3851 add cceditbox to wp8

squirrelChen:
@dumganhar @stammen
I added a usercontrol called EditBox.xaml to the xaml project. It contain a textbox, a 'OK' button, a 'Cancel' button it the top of screen, it cover the screen as Modal window, while clicking the 'OK' button or press the 'enter' key, it set the new text to the label to show, and click 'Cancel' button to return directly.
I used an interface in native component and implemented at xaml to call C#, view the link:http://stackoverflow.com/questions/17304386/how-does-one-make-function-calls-or-trigger-events-from-a-native-component-into as a reference, not using the code as show IME and hide.
Please review it, Thanks.

### DIFF
--- a/samples/Cpp/TestCpp/proj.wp8-xaml/TestCpp/TestCpp/EditBox.xaml
+++ b/samples/Cpp/TestCpp/proj.wp8-xaml/TestCpp/TestCpp/EditBox.xaml
@@ -10,7 +10,7 @@
     d:DesignHeight="480" d:DesignWidth="480">
 
     <Grid x:Name="LayoutRoot" Background="{StaticResource PhoneSemitransparentBrush}">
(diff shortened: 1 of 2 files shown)


stammen:
Good work! Before I finish reviewing please add Cocos2d copyright to all new files. 

Thanks,

Dale

stammen:
Please ignore my previous comment about moving the files. I forgot they were part of extension lib. They are in the correct location. Please just add the copyright notice so I may complete the review.

dumganhar:
@koowolf 
Good work. Please add `copy right`, I will merge it. Thanks.

squirrelChen:
@stammen @dumganhar 
I added copyright OK.

PR #5402, opened 2014-02-20
CCTexture2D.cpp: Respect _PVRHaveAlphaPremultipled flag in initWithImage()

wefiends:
Currently, the `_PVRHaveAlphaPremultipled` is ignored, because the code path that PVR loading takes never reaches the required check. 

When a PVR texture is being loaded, `Image` data gets passed to `initWithData` and then `initWithMipmaps` (which forces the `_hasPremultipliedAlpha` to false).

This patch moves the check before any init methods so that the `_hasPremultipledAlpha` flag is set correctly for all texture types. For non-PVR it defaults to false if unknown (same as before).

### DIFF
--- a/cocos/2d/CCTexture2D.cpp
+++ b/cocos/2d/CCTexture2D.cpp
@@ -663,7 +663,6 @@ bool Texture2D::initWithMipmaps(MipmapInfo* mipmaps, int mipmapsNum, PixelFormat
     _maxS = 1;
     _maxT = 1;
 
-    _hasPremultipliedAlpha = false;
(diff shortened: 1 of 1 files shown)


dumganhar:
@boyu0, please review this PR. Thanks.

boyu0:
Hmm, premultiplied alpha image now is rendered incorrect in our engine, and we will remove this "premultiplied alpha" concept sooner or later. So before this, I think it's better to leave this "premultiplied alpha" thing as they are for now.

PR #5426, opened 2014-02-20
make TestCppComponent project load right

lostplesed:


squirrelChen:
Please include angleproject, libextensions etc project to the solution, Thanks.

PR #5435, opened 2014-02-22
Fix console windows crash

heliclei:


cocos/base/CCConsole.cpp
```diff
@@ -325,7 +325,12 @@ bool Console::listenOnTCP(int port)
         if (bind(listenfd, res->ai_addr, res->ai_addrlen) == 0)
             break;          /* success */
 
-        close(listenfd);    /* bind error, close and try next one */
+/* bind error, close and try next one */
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
+        closesocket(listenfd);
```
dumganhar, line 330: Should we invoke `WSACleanup` after `closesocket`?

heliclei, line 330: I think WSACleanup is to clean up the resources alloc by WSAStart(), so, we need only call WSACleanup once when CCConsole exit.

tests/test-cpp/Classes/ConsoleTest/ConsoleTest.cpp
```diff
@@ -165,14 +166,14 @@ void ConsoleTCP::onEnter()
...
-    _console = Director::getInstance()->getConsole();
-
     static struct Console::Command commands[] = {
         {"hello", "This is just a user generated command", [](int fd, const std::string& args) {
+            
             const char msg[] = "how are you?\nArguments passed: ";
-            write(fd, msg, sizeof(msg));
-            write(fd, args.c_str(), args.length());
-            write(fd, "\n",1);
+            auto _console = Director::getInstance()->getConsole();
```
dumganhar: This line isn't needed. Right?

heliclei: Yes, thanks for pointing this out.

dumganhar:
Thanks. Merged.

PR #5446, opened 2014-02-24
fix compiling error cause by using 'fmaxf'.

WenhaiLin:


cocos/2d/CCSprite.cpp
```diff
@@ -634,8 +634,13 @@ bool Sprite::culling() const
     y -= screen_half.height;
 
     // convert content size to world coordinates
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WIN32
```
ricardoquesada, line 637: thanks. I think there is no need for the `#ifdef`.

We can just use the Windows version for all the platforms.

WenhaiLin, line 637: Agree with you.

PR #5451, opened 2014-02-24
closed #4123:fixed thread deadlock if new functions are added in callback.

WenhaiLin:


huangzhentao:
Excuse me,I want to know when the  effect of label  will improve ? Because I need it in my project.

PR #5453, opened 2014-02-24
Fixed win32 build : wrong file path and missing link directories in CMakeLists.txt

theor:
The PR should be quite straight forward:
- Some file has been moved, but its path was not updated in CMakeLists
- Looks like some external link dirs were missing

TestCpp now builds out of the box, just by doing :

```
mkdir buildcmake
cd buildcmake
cmake ..
MSBuild Cocos2dx.sln
```

We use cocos in a CMake project, so we need it to build as a cmake subdirectory.
Feel free to ask any question.


dumganhar:
Hi, thanks for your contribution.
I saw your changes:
0d28ee45-bf96-4359-876d-f13dc414ab0e
This change will break `vs-build` since vs uses the link `libcurl_imp.lib`.

Another thing is that you broke `linux-build` too , you could refer to https://travis-ci.org/cocos2d/cocos2d-x/jobs/19533138 .

Could you fix those problems? Thanks.

theor:
Hi,

Thanks for the quick answer.
The linux build should be a quick fix - I'm not sure what you mean by "the bs-build" ?

Thanks,

Theo

dumganhar:
I meant by cocos2d-x by `Visual Studio`.

theor:
It works using cmake. The sln/vcxproj in the repo are not generated via cmake ?

dumganhar:
Yes, we use https://github.com/cocos2d/cocos2d-x/blob/develop/build/cocos2d-win32.vc2012.sln.
It isn't generated by `cmake`.

theor:
Can I ask why ? I admit I'm a bit surprised - our project just "add_subdirectory(${COCOS_ROOT})" and it works like a charm.

theor:
As curl is the only lib using this weird naming pattern, would that be acceptable to just fix the sln ?

PR #5461, opened 2014-02-25
Fix spine::Skeleton not updating on re-entry.

Wilhansen:
Currently, if one removes a spine::Skeleton node from the hierarchy and reinserts it, the node will no longer be updated.

This fix could be done in a tighter manner by scheduling updates in spine::SkeletonAnimation instead and listening for animation events (like scheduleUpdate when ANIMATION_START is detected..
However, this may need a more robust interface on scheduleUpdate and unscheduleUpdate to work correctly (which I could do sometime soon since the current head has Node::isScheduled).


dumganhar:
Thanks.

PR #5464, opened 2014-02-25
Add method to reload cached texture & update the nodes used the texture.

natural-law:


cocos2dx/textures/CCTextureCache.h
```diff
@@ -133,6 +133,21 @@ class CC_DLL CCTextureCache : public CCObject
...
+    /** Reload texture from the image file
+     * If the file image hasn't loaded before, load it.
+     * Otherwise the texture will be reloaded from the file image.
+     * The "filenName" parameter is the related/absolute path of the file image.
+     * Return true if the reloading is succeed, otherwise return false.
+     */
+    bool reloadTexture(const char* fileName);
+    
+    /** Update the nodes & it's children which used texture
+     * The "node" parameter is the root node will be updated
```
minggo: Now it only supports node that which is a `Sprite`. How about mention it here?

natural-law: I found objects from other classes are not necessary to invoke the method setTexture().
Just invoke method reloadTexture(), the appearance of other objects will be correct.

minggo: So change parameter type from `CCNode*` to `CCSprite*`?

natural-law: No, I think it must be a root node. If it is changed, the user must find the sprites in the scene before invoke the method.

PR #5466, opened 2014-02-26
fixed #3868. when singleton deleted, it should not dispatch callbacks.

zilongshanren:


cocos/network/HttpClient.cpp
```diff
@@ -214,7 +214,9 @@ void HttpClient::networkThread()
         s_responseQueue->pushBack(response);
         s_responseQueueMutex.unlock();
         
-        scheduler->performFunctionInCocosThread(CC_CALLBACK_0(HttpClient::dispatchResponseCallbacks, this));
+        if (nullptr != this) {
```
dumganhar: How could you confirm that after `HttpClient` instance was destroyed, the `this` pointer in `HttpClient::networkThread` will be null? I think it will be a unexpected value.

PR #5467, opened 2014-02-26
closed #4144: fix bug: Can not use Node::setPhysicsBody to reset a physics body.

boyu0:


dumganhar:
What did you mean `at the second time`? Could you describe more about that?

PR #5469, opened 2014-02-26
Add method to reload cached texture & update the nodes used the texture.

natural-law:


ricardoquesada:
My feedback:
- I like the idea of reloading texture in runtime. 
- it is a useful feature for the editor, and also for prototyping games. You can have your game running on a device (a prototype) and you can upload a new texture a just update it in runtime. Cool feature.

The implementation that you sent will work for some cases, but it won't in other cases.
eg: If won't work for Particles.
And the other thing is that `TextureCache` should know nothing about `Sprite`. It should only know about `Texture2D` objects.

One possible implementation that will work on more scenarios is:
1. `TextureCache::reloadTexture()` obtains the `Texture2D` that is using file.
2. The `Texture2D` object is recreated internally. It will keep using the same `Texture2D` object, but internally it will destroy the glTexture() and create a new one based on the new filename.

With those 2 steps it should work if the image has the same size as before...(and it will work for Sprites, Particles, MotionStreak, Menu, etc...)

What happens if the filename has different size ? It won't work... and to cover that issue,  we need an additional step:
1. Texture2D emits an event saying that the Texture has changed. `Sprite`, `ParticleSystem` and other objects should subscribe to that event.
   And when that event is received, and if the texture is the same as theirs, they should its size.

Something similar could be done with the .plist and SpriteFrameCache.

natural-law:
@ricardoquesada 
In your solution, there is also the problem mentioned before : Which rect should be used to refresh the nodes? Should we record the old texture rect? Then maybe we will have this logic code:

```
void Sprite::textureReloaded()
{
    if (this->getTextureRect() == textureOldRect)
    {
        // refresh with the new texture rect
    }
    else
    {
        // do nothing, use the texture rect before
    }
}
```

Is there any misunderstanding of the solution?
If this solution is decided to be used, can I add the method TextureCache::reloadTexture() first? And add the events & event listeners later?

ricardoquesada:
I don't think it is needed to save the state of textureRect.

But I think it is important to understand the CocoStudio Use Case:
Could you describe the Use Case in more detail?

Because, if what you want to support is to reload textures in runtime inside CocoStudio, and those textures could be modified by PhotoShop (or any other image editor), then the solution is more complex because:
- the new image could have a different size
- but also, what happens if that image is an spritesheet ?

So, if you are thinking something to be used only by CocoStudio, then probably the best is to create an ad-hoc solution for CocoStudio... something that is only used by CocoStudio, but it is not part of cocos2d-x.

For example, you could add the "reloadTexture" functions as an extensions.
Also, you might need a kind of dictionary of Used Sprites <--> textures
and Used Sprites <---> SpriteFrames.

this is just an idea... but if you could describe the Use Case in more detail, it would be great. Thanks.

natural-law:
@ricardoquesada 
Yep, we also hope there is an ad-hoc solution. As you mentioned, the solution should consider the Use Case of CocoStudio.

As I known from the communication with CocoStudio, the Use Case is the complex one you mentioned:
Once a picture/plist file is updated, the appearance of the editor should be refreshed. Though the picture size is changed.

Many classes should be modified no matter which solution is selected. So, I haven't found a simple ad-hoc solution by now.

cocos/2d/CCTextureCache.cpp
```diff
@@ -372,6 +373,69 @@ Texture2D* TextureCache::addImage(Image *image, const std::string &key)
...
+    
+    // update self
+    auto sprite = dynamic_cast<Sprite*>(node);
+    if (nullptr != sprite && sprite->getTexture() == texture) {
+        auto size = texture->getContentSize();
+        sprite->setTextureRect(Rect(0, 0, size.width, size.height));
+    }
+    
+    // update children
+    auto children = node->getChildren();
```
dumganhar: please use `auto&` instead. Otherwise, it will copy all children to a new `cocos2d::Vector<Node*>`

cocos/2d/CCTextureCache.cpp
```diff
@@ -372,6 +373,69 @@ Texture2D* TextureCache::addImage(Image *image, const std::string &key)
...
+    if (children.size() > 0)
+    {
+        for (auto obj : children)
```
dumganhar: `const auto& obj` would be better.

dumganhar:
> Texture2D emits an event saying that the Texture has changed. Sprite, ParticleSystem and other objects should subscribe to that event. And when that event is received, and if the texture is the same as theirs, they should its size.

Yep, i think this is a better way. :)

ricardoquesada:
@natural-law I discussed it with Nite, and agreed that the best solution would be create "ad-hoc" solution for Cocosbuilder and put that code in the CocosBuilder repo.

But if you want to send a patch just to recreate `Texture2D` object with a different images, it looks good to me.

Please, let me know if that works for you. Also, Nite will be at Beijing's office this Tuesday.

natural-law:
@ricardoquesada 
It's ok for me, I can only add method reloadTexture() in engine. And @xiexin36 will discuss the "ad-hoc" solution with Nite this week later.
Thank you!

darkdukey:
@ricardoquesada @natural-law 
I talked to Studio team and we came up a solution. and we break this problem into 2 problems
-  a. update texture
  for update texture `reloadTexture()` should be enough, but it would be nice to have a `Node.refresh()` method, it will make their work easier (without doing a lot of type casting) since studio's `UIWidget` will inherent from Node and they can override `Node.refresh()` it to perform special reload logic
-  b. update plist
  for update plist, Studio will just destroy the old object and reload it completely.

minggo:
@darkdukey 
For a, did you mean add `TexureCache::reloadTexture()` and  `Node::refresh()`? When `Node::refresh` invoked? And what's the logic of it?

natural-law:
@darkdukey 
If method `Node.refresh()` is added, should many classes in engine need override it?
Such as `Sprite`, if it's not override the method, how can it be refreshed?

Can you show us more details of the solution? Thank you!

ricardoquesada:
I agree with @minggo regarding `Node.refresh()`.
What is the expected behavior of `refresh` ? what should it do ?

The way I see it is:
- If we add a `refresh()` method in Node, then it should work 100% in all the supported nodes. From `Node`, to `Sprite`, to `TileMap`, to `RenderTexture`, etc... And that could be very expensive to implement, and expensive to the user as well since all the nodes will take more memory (and memory is very expensive in mobile devices).

And I'm not a fan of implementing `refresh` in Node.
But +1 for implementing `reload()` in `Texture2D`.

darkdukey:
The `Node.refresh()` will do nothing, and the child class will override it to implement different refresh logic
For example

``` cpp
Sprite::refresh()
{
    //Reload texture
    //Reset texture rect
}

Tilemap::refresh()
{
    //Reload tilemap data
    //Reload tile spreadsheet
}
```

The reason for use `refresh()` is that it will be more clean and easy to extend, the current implementation of `TextureCache::updateTexture` will only work for `Sprite` and we don't want that, because if say later we want to add support to `TileMap` we will have to add another set of type casting.

The other main reason I want to have a `refresh()` function is that currently there is really no other way for the editor to do live reloading, except maybe destroy the scene reload everything. I think we maybe can't implement all the `refresh()` at once, but if we have the interface in place, studio team can implement them one by one, overtime. we can also use documentation to tell developers how to extend it, so maybe developers can extend them and it will be useful for their own editors.

Also I don't think performance is a concern because the main user will be the editor, we're not going to call it very often. We can even #ifdef it with a `ENABLE_EDITOR` flag, so for the game developer can't use them at all.

natural-law:
@darkdukey 
The reason you described is ok to me. In the solution, it's also required engine to implement refresh the rect of nodes.
But after all, it's only the solution of where should we implement the refresh feature. We have no solution to solve : Which rect should be used to refresh nodes?

minggo:
Agree with @natural-law.

ricardoquesada:
If I understood correctly:
- We are trying to solve a Use Case for CocoStudio
- This Use Case is about updating in runtime files modified by a 3rd party editor (not integrated into CocoStudio):
  - An external editor update a file used by CocoStudio, and CocoStudio should update it in runtime.
  - And one way to solve that is by adding `refresh` to `Node`
  - but since it is time consuming to do it correctly, we will just create an interface... so Nodes will do nothing for `refresh` at the beginning
  - CocoStudio team will implement the `refresh` only in the Nodes that are needed to CocoStudio
  - The rest of the Nodes won't have a `refresh`

The way I see it is that the proposed solution:
- looks very attractive to CocoStudio users
- but doesn't look very attractive to cocos2d-x users because: 
  - It adds functionality to `Node` that is not needed (making the API more difficult to understand and more difficult to maintain)
  - and the newly added functionality doesn't work as expected since only a subset of nodes will have implemented

I'm sorry, but I don't like the proposed solution :(

darkdukey:
Yes I totally think this is a Use-Case CocoStudio, so it's probably only going to be benefit for CocoStudio. But I think eventually studio users are also cocos2d-x user so we should try our best to give them our support.

I had talked to Studio team about solutions, mainly there are two
1. cocos2d-x support it internally
   - Then cocos2d-x need to provide a way for studio to perform the refresh either explicitly calling `refresh` or send in `event`.
   - Pro: it's easy for CocoStudio to integrate with it's current architecture.
   - Con: Will introduce method not really required for game developer or the `event` will introduce more overhead.
2. cocostudio support it externally
   - Destroy the old node and recreate it, then we probably don't need to do anything. (Maybe provide `reloadTexture` to make their job a little bit easier)
   - Pro: Minimal change to cocos2d-x
   - Con: Will break some functionality for CocoStudio like Undo/Redo

I think maybe for now, let's update this one so we can at least provide `reloadTexture`

natural-law:
OK, I will remove the method `updateTexture()` in the PR as soon as possible.

edwardZhang:
I used the spriteBuilder, it is good to refresh texture. a single texture and spriteSheet are doing very well.
Maybe we can refer to it.

ricardoquesada:
+1 for refresh texture only.

PR #5471, opened 2014-02-26
Creates JSBindings for the socket.io library, allowing the same API to ...

himynameschris:
...be used for JS and JSB codes

[socketio jsbindings] initial commit of dev work done in alpha branch by chris, based on websocket jsb
[socketio jsbindings] update vs project and headings
[socketio jsbindings] modify SocketIO constructor arguments to match socket.io javascript library argument order (url first)
[socketio jsbindings] add jsb_socketio to appdelegate in js tests
[socketio jsbindings] implement event handling and callback, update SocketIO with dummy JS event fired method in delegate
[socketio jsbinding] modify tag getter and setter to be js property
[socketio jsbinding] update by removing unneeded js helpers, use new macro JS_BINDED_PROP_DEF_ACCESSOR
[socketio jsbinding] update submodule
[socketio jsbinding] update submodule
[socketio jsbinding] streamline tag property usage and connection event handling to match js style
[socketio jsbinding] update submodule for jsbindings tests

Signed-off-by: Chris Hannon himynameschris@gmail.com

### DIFF
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -443,7 +443,7 @@ void SIOClientImpl::onMessage(WebSocket* ws, const WebSocket::Data& data)
 				}
 
 				c->fireEvent(eventname, payload);
-				c->getDelegate()->eventFired(c, eventname, payload);
+				c->getDelegate()->fireEventToScript(c, eventname, payload);
 			}
 			
 			break;
--- a/cocos/network/SocketIO.h
+++ b/cocos/network/SocketIO.h
@@ -94,7 +94,7 @@ public:
         virtual void onMessage(SIOClient* client, const std::string& data) = 0;
         virtual void onClose(SIOClient* client) = 0;
         virtual void onError(SIOClient* client, const std::string& data) = 0;
-		virtual void eventFired(SIOClient* client, const std::string& eventName, const std::string& data) { CCLOG("SIODelegate event '%s' fired with data: %s", eventName.c_str(), data.c_str()); };
+		virtual void fireEventToScript(SIOClient* client, const std::string& eventName, const std::string& data) { CCLOG("SIODelegate event '%s' fired with data: %s", eventName.c_str(), data.c_str()); };
     };
 
 	/**
--- a/cocos/scripting/javascript/bindings/network/jsb_socketio.cpp
+++ b/cocos/scripting/javascript/bindings/network/jsb_socketio.cpp
@@ -84,33 +84,33 @@ public:
 	virtual void onConnect(SIOClient* client) {
 		CCLOG("JSB SocketIO::SIODelegate->onConnect method called from native");
 
-		this->eventFired(client, "connect", "");
+		this->fireEventToScript(client, "connect", "");
 
 	}
 
     virtual void onMessage(SIOClient* client, const std::string& data) {
 		CCLOG("JSB SocketIO::SIODelegate->onMessage method called from native with data: %s", data.c_str());
 		
-		this->eventFired(client, "message", data);
+		this->fireEventToScript(client, "message", data);
 
 	}
 
     virtual void onClose(SIOClient* client) {
 		CCLOG("JSB SocketIO::SIODelegate->onClose method called from native");
         
-		this->eventFired(client, "disconnect", "");
+		this->fireEventToScript(client, "disconnect", "");
 
 	}
 
     virtual void onError(SIOClient* client, const std::string& data) {
 		CCLOG("JSB SocketIO::SIODelegate->onError method called from native with data: %s", data.c_str());
 
-		this->eventFired(client, "error", data);
+		this->fireEventToScript(client, "error", data);
 
 	}
 
-	virtual void eventFired(SIOClient* client, const std::string& eventName, const std::string& data) {
-		CCLOG("JSB SocketIO::SIODelegate->eventFired method called from native with name '%s' data: %s", eventName.c_str(), data.c_str());
+	virtual void fireEventToScript(SIOClient* client, const std::string& eventName, const std::string& data) {
+		CCLOG("JSB SocketIO::SIODelegate->fireEventToScript method called from native with name '%s' data: %s", eventName.c_str(), data.c_str());
 
 		js_proxy_t * p = jsb_get_native_proxy(client);
         if (!p) return;


dumganhar:
Awesome. Thanks. I will review it soon.

cocos/network/SocketIO.cpp
```diff
@@ -603,7 +604,7 @@ void SocketIO::destroyInstance()
     CC_SAFE_DELETE(_inst);
 }
     
-SIOClient* SocketIO::connect(SocketIO::SIODelegate& delegate, const std::string& uri)
+SIOClient* SocketIO::connect(const std::string& uri, SocketIO::SIODelegate& delegate)
```
dumganhar, line 607: Why to change argument order? There will be some developers complain on this change.

cocos/network/SocketIO.h
```diff
@@ -92,6 +94,7 @@ class SocketIO
         virtual void onMessage(SIOClient* client, const std::string& data) = 0;
         virtual void onClose(SIOClient* client) = 0;
         virtual void onError(SIOClient* client, const std::string& data) = 0;
+		virtual void eventFired(SIOClient* client, const std::string& eventName, const std::string& data) { CCLOG("SIODelegate event '%s' fired with data: %s", eventName.c_str(), data.c_str()); };
```
dumganhar: What's `eventFired` used for ?
If it's needed, why not change it to `onEventFired`?

cocos/scripting/javascript/bindings/network/jsb_socketio.cpp
```diff
@@ -0,0 +1,452 @@
...
+     CCLOG("JSB SocketIO.setTag method called");
+	JSObject* jsobj = obj.get();
+	js_proxy_t *proxy = jsb_get_js_proxy(jsobj);
+	SIOClient* cobj = (SIOClient *)(proxy ? proxy->ptr : NULL);
+	JSB_PRECONDITION2( cobj, cx, false, "Invalid Native Object");
+	
+	if (cobj) {
+		const char *out = "";
+
+		jsval_to_charptr(cx, vp, &out);
```
dumganhar: Why not to use `jsval_to_std_string`?

cocos/network/SocketIO.h
himynameschris: eventFired is more of a method to translate native events to JS events,
instead of a handler

I suppose a more appropriate name would be "fireJSEvent"

cocos/scripting/javascript/bindings/network/jsb_socketio.cpp
```diff
@@ -0,0 +1,452 @@
...
+    js_cocos2dx_socketio_class->getProperty = JS_PropertyStub;
+    js_cocos2dx_socketio_class->setProperty = JS_StrictPropertyStub;
+    js_cocos2dx_socketio_class->enumerate = JS_EnumerateStub;
+    js_cocos2dx_socketio_class->resolve = JS_ResolveStub;
+    js_cocos2dx_socketio_class->convert = JS_ConvertStub;
+    js_cocos2dx_socketio_class->finalize = js_cocos2dx_SocketIO_finalize;
+    js_cocos2dx_socketio_class->flags = JSCLASS_HAS_RESERVED_SLOTS(2);
+
+	static JSPropertySpec properties[] = {
+		//{"tag", 0, JSPROP_ENUMERATE | JSPROP_PERMANENT | JSPROP_SHARED, JSOP_WRAPPER(js_cocos2dx_SocketIO_setTag), JSOP_WRAPPER(js_cocos2dx_SocketIO_getTag)},
```
dumganhar, line 416: `tag` is not needed?

cocos/scripting/javascript/bindings/network/jsb_socketio.cpp
himynameschris: No particular reason, I am still learning all of the helper methods, so
jsval_to_std_string is probably better

cocos/network/SocketIO.h
dumganhar: Got it. By why not make it as the member function of `JSB_SocketIODelegate`?
Because only that class use it. c++ don't need this function. Right?

cocos/scripting/javascript/bindings/network/jsb_socketio.cpp
dumganhar: Yes, it will be better. :) Please change to that function. I think we should deprecate `jsval_to_charptr`.

cocos/scripting/javascript/bindings/network/jsb_socketio.cpp
himynameschris, line 416: the 'tag' property is very useful and needed for the tests, these two lines
are duplicates, I saw that "JS_BINDED_PROP_DEF_ACCESSOR" is new, I
commented the first method but didn't want to lose the code if it needed to
switch back. The commented line can be removed

//{"tag", 0, JSPROP_ENUMERATE | JSPROP_PERMANENT | JSPROP_SHARED,
JSOP_WRAPPER(js_cocos2dx_SocketIO_setTag),
JSOP_WRAPPER(js_cocos2dx_SocketIO_getTag)}
JS_BINDED_PROP_DEF_ACCESSOR(SIOClient, tag),

cocos/network/SocketIO.h
himynameschris: It is called from the native codes on line 446 of cocos/network/SocketIO.h,
if it isn't declared in the parent class for JSB_SocketIODelegate then it
fails

cocos/network/SocketIO.cpp
himynameschris, line 607: this is to better match the original javascript API, which has the order (uri, args)

cocos/network/SocketIO.h
dumganhar: I think `fireJSEvent` is not suitable since `luabindings` may also need it.
How about rename it to `fireEventToScript` ?

BTW, we have split script bindings stuffs to 
https://github.com/cocos2d/cocos2d-js.git
https://github.com/cocos2d/cocos2d-lua.git
:)

himynameschris: fireEventToScript sounds good, I will update the codes. as soon as I get a chance to setup the https://github.com/cocos2d/cocos2d-js.git repo I will submit a PR there

dumganhar:
@hannon235 
Thanks for your update. I will merge your commit to https://github.com/cocos2d/cocos2d-js.git

himynameschris:
See https://github.com/cocos2d/cocos2d-js/pull/206 and https://github.com/cocos2d/cocos2d-x/pull/5923

PR #3874, opened 2013-10-06
Qt update 2

sergey-shambir:
Updated Qt5 port:
- fixed build;
- much improved support for desktop Linux (OpenGL, mouse/keyboard input, unresizable window, IME);
- modified doxygen.config to enable QtCreator help build;
- added qt5 project to template.

Forum topic: http://www.cocos2d-x.org/forums/6/topics/33997
Unfortunately, no items from TODO list in this topic are done.

PR #5472, opened 2014-02-26
closed #3874: expand textureID bit to 32bits

dabingnn:


cocos/2d/renderer/CCQuadCommand.cpp
```diff
@@ -92,13 +92,13 @@ void QuadCommand::generateMaterialID()
...
-    // +---------------------+-------------------+----------------------+
+    // +---------------------+-------------------+----------------------------------------+
+    // | Shader ID (10 bits) | Blend ID (4 bits) | empty (18bits) |  Texture ID (32 bits) |
+    // +---------------------+-------------------+----------------------------------------+
 
-    _materialID = (uint32_t)_shader->getProgram() << 22
-            | (uint32_t)blendID << 18
-            | (uint32_t)_textureID << 0;
+    _materialID = (uint64_t)_shader->getProgram() << 54
+            | (uint64_t)blendID << 40
```
robertjerovsek: Shouldn't the blendID be shifted by 50 bits?

dabingnn: Thank you, I will fix it.

dumganhar:
Thanks @dabingnn . :)

ricardoquesada:
thanks.
Was there any performance penalty after applying this patch ?

I would guess that no. But it is always good to do a performance test on an old 32-bit device to make sure that the performance it is still good.

dabingnn:
I profiled this patch, the method involved in materialID consumption are generateMaterialID and getMaterialID, they only consume 0.1-0.2% of the total CPU. Besides, there are no FPS decrease on normal tests and performance tests(tests on spritePerformance test, 3000 sprites). Given this results, I think there is no performance problem when migrating materialID to 64bits.

PR #5473, opened 2014-02-26
update releate note

minggo:
- add document reference
- fix some typo
- refactor lua section
- remove duplicated content in existing documents


minggo:
@ricardoquesada 
- Could you please review it? I merge it now. But i can modify it according your comment. 
- And could you please add the section for `auto batching`, because i don't know the difference between `auto batching` and `SpriteBatchNode`.
- And could you please add `exceptions` in `Global Z order`

PR #5478, opened 2014-02-26
WP8 backbutton fix

stammen:
This pull request fixes the following issues:
1. Back button event on device does not occur on cocos2d-x rendering thread. I added a new BackButtonEvent to be processed on the rendering thread. 
2. Added C# MessageBox to XAML so CCMessageBox will now display using native WP8 MessageBox which is not available from C++ code.
3. Refactored recently added CCEditBox code to fix broken WP8 C++ builds and to match the style of the other delegates implemented in MainPage.xaml.cs.
4. Moved CPP sample xaml and cs files to cocos2dx/platform/wp8-xaml to allow reuse and reduce bugs by not introducing changes in different copies of the same file.
5. Fixed broken wp8-xaml cpp template broken by previous checkins by community developers.
6. Added CopyTemplateProject to update wp8-xaml cpp template files if they have changed.
7. Increase FPS by about 25% by fixing size of synchronized texture. Improves performance on HD phones like the Lumia 920.

### DIFF
--- a/cocos2dx/platform/wp8/CCEGLView.cpp
+++ b/cocos2dx/platform/wp8/CCEGLView.cpp
@@ -511,28 +511,30 @@ CCPoint CCEGLView::TransformToOrientation(Point p)
 {
     CCPoint returnValue;
 
-#if 0
-    float x = getScaledDPIValue(p.X);
-    float y = getScaledDPIValue(p.Y);  
-#endif // 0
     float x = p.X;
     float y = p.Y;  
+ 
+    if(!m_isXamlWindow)
+    {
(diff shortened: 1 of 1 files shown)


dumganhar:
Thanks.

lostplesed:
I merge the pull request and the it really increase FPS by about 25% in my lumia 920. Thanks for the pr. But some label text is fuzzy. can not see it very clearly, i revert the code it just ok.

dungmv:
Hi MSOpenTech team!
will you use angle project wapper for render opengl or directx directly for render on cocos v3?

PR #5486, opened 2014-02-27
fixed #3599. add adjustTextFieldPosition function call to the visit method

zilongshanren:
add adjustTextFieldPosition function call to the visit method of CCEditBoxImplIOS to solve this issue

### DIFF
--- a/extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
+++ b/extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
@@ -614,7 +614,6 @@ void EditBoxImplIOS::setAnchorPoint(const Point& anchorPoint)
 
 void EditBoxImplIOS::visit(void)
 {
-    
 }
 
 void EditBoxImplIOS::onEnter(void)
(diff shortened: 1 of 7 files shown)


extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
```diff
@@ -614,7 +614,9 @@ static CGPoint convertDesignCoordToScreenCoord(const Point& designCoord, bool bI
 
 void EditBoxImplIOS::visit(void)
 {
-    
+    if (nullptr != _systemControl) {
+        this->adjustTextFieldPosition();
```
dumganhar: After thinking a while, I think we need to use a schedule to update the position of `UITextField`.
And the interval should be 0.5s probably.

zilongshanren: Ok, I will use the scheduler instead.

dumganhar: Thanks.

dumganhar:
Thanks

zilongshanren:
It's my honor:)

PR #5488, opened 2014-02-28
dirty in draw() and visit()

ricardoquesada:
DO NOT MERGE, 
testing github plugin

### DIFF
--- a/cocos/2d/CCAtlasNode.h
+++ b/cocos/2d/CCAtlasNode.h
@@ -69,7 +69,7 @@ public:
 
     
     // Overrides
-    virtual void draw() override;
+    virtual void draw(bool transformDirty) override;
     virtual Texture2D* getTexture() const override;
     virtual void setTexture(Texture2D *texture) override;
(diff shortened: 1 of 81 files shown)


cocos/2d/CCAtlasNode.h
```diff
@@ -69,7 +69,7 @@ class CC_DLL AtlasNode : public Node, public TextureProtocol
 
     
     // Overrides
-    virtual void draw() override;
+    virtual void draw(bool transformDirty) override;
```
minggo, line 72: Break compatibility?
How about changing to

``` cpp
virtual void draw(bool transformDirty = defaultValue) override;
```

ricardoquesada, line 72: this PR was not supposed to go.

But yes, I'm working on a new draw.
For the record, `draw` is already backwards incompatible.
You cannot use your v2.2 `draw` in v3.0. The new renderer broke it.

Hopefully, in a few days I'll have improved version of `draw` and `visit` with better performance.

PR #5491, opened 2014-02-28
visit and draw improved.

ricardoquesada:
`visit()` and `draw()` are improved.
they both receive:
- `Renderer`
- `kmMat4` for transformation
- `bool` whether or not the transform changed from the previous frame

Although `draw` and `visit` are not using the 3 new parameters,
they will start using it in the coming days (or weeks, or months).

Those parameters are needed for:
- `Renderer` shouldn't be a singleton. There should be one renderer per Director. There could be more than one renderer
- we should get rid of `kmGLXXX` functions since nodoby knows what is happening. the `kmMat4` is for that.
- whenever possible, we should try to optimize the multiplications. the `dirty` flag is for that.

We need those changes, so, it is better to change the API now, and not right after v3.0-final is released.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -492,21 +492,6 @@ void Node::setPositionZ(float positionZ)
     setGlobalZOrder(positionZ);
 }
 
-void Node::setNormalizedPosition(const cocos2d::Point &position)
-{
-    _normalizedPosition = position;
-    Size s = Director::getInstance()->getVisibleSize();
-    Point p;
-    p.x = s.width * position.x;
-    p.y = s.height * position.y;
-    setPosition(p);
-}
-
-const Point& Node::getNormalizedPosition() const
-{
-    return _normalizedPosition;
-}
-
 ssize_t Node::getChildrenCount() const
 {
     return _children.size();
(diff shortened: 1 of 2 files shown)


minggo:
It seems good to me. I will run `testcpp` and give the feedback.

Just one question: why remove `CC_TEXTURE_ATLAS_USE_TRIANGLE_STRIP` support?

ricardoquesada:
I removed `CC_TEXTURE_ATLAS_USE_TRIANGLE_STRIP` because it was not needed.
I added in cocos2d-iphone like 5 years ago when I was testing performance.
`TRIANGLE_STRIP` was never useful and I left it in the code just in case. But the reality is it is not useful, make the code less maintainable, and the rest of the code doesn't support it (Progress, QuadCommand, Effects don't support TRIANGLE_STRIP)

minggo:
Got it. Thanks.

minggo:
I have run `TestCpp` on iOS simulator, it seems good.

dabingnn:
This pr is a big improvement. I tested `testcpp`, there is a bug on RenderTextureTest->Testing Z Buffer in RenderTexture, it seems the viewport are resized use a wrong size. However, we can fix the bug when it is merged.

ricardoquesada:
I'm glad that you like it. I'll take a look at the bug. But if I can't fix it today, as you said, we can merge the PR and fix the bug later.

PR #5493, opened 2014-02-28
closed #4150: fix physics position and rotation bug.

boyu0:
1. Fix the bug if a node has a parent, it's physics body position is incorrect.
2. Set anchor point to middle when invoke Node::setPhysicsBody
3. Add PhysicsBody set/getPosition/RotationOffset method.
4. Add PhysicsPositionRotationTest.

### DIFF
--- a/cocos/physics/CCPhysicsBody.h
+++ b/cocos/physics/CCPhysicsBody.h
@@ -195,11 +195,11 @@ public:
     /** set body position offset, it's the position witch relative to node */
     void setPositionOffset(const Point& position);
     /** get body position offset. */
-    Point getPositionOffset();
+    Point getPositionOffset() const;
     /** set body rotation offset, it's the rotation witch relative to node */
     void setRotationOffset(float rotation);
     /** set the body rotation offset */
-    float getRotationOffset();
+    float getRotationOffset() const;
     
     /**
      * @brief test the body is dynamic or not.
--- a/cocos/physics/CCPhysicsBody.cpp
+++ b/cocos/physics/CCPhysicsBody.cpp
@@ -836,7 +836,7 @@ void PhysicsBody::setPositionOffset(const Point& position)
     }
 }
 
-Point PhysicsBody::getPositionOffset()
+Point PhysicsBody::getPositionOffset() const
 {
     return _positionOffset;
 }
@@ -851,7 +851,7 @@ void PhysicsBody::setRotationOffset(float rotation)
     }
 }
 
-float PhysicsBody::getRotationOffset()
+float PhysicsBody::getRotationOffset() const
 {
     return _rotationOffset;
 }


cocos/2d/CCNode.cpp
```diff
@@ -534,6 +536,14 @@ void Node::setVisible(bool var)
 
 void Node::setAnchorPoint(const Point& point)
 {
+#if CC_USE_PHYSICS
+    if (_physicsBody != nullptr && point.equals(Point::ANCHOR_MIDDLE))
```
dumganhar: Should it be `!point.equals(Point::ANCHOR_MIDDLE)`?

boyu0: Yes, sorry, I will fix this.

cocos/physics/CCPhysicsBody.h
```diff
@@ -192,6 +192,15 @@ class PhysicsBody : public Ref
     /** get the body rotation. */
     float getRotation() const;
     
+    /** set body position offset, it's the position witch relative to node */
+    void setPositionOffset(const Point& position);
+    /** get body position offset. */
+    Point getPositionOffset();
```
minggo: `Point getPositionOffset()` -> `Point getPositionOffset() const`?

cocos/physics/CCPhysicsBody.h
```diff
@@ -192,6 +192,15 @@ class PhysicsBody : public Ref
...
+    /** set body rotation offset, it's the rotation witch relative to node */
+    void setRotationOffset(float rotation);
+    /** set the body rotation offset */
+    float getRotationOffset();
```
minggo: `float getRotationOffset()` -> `float getRotationOffset() const`?

marquisd:
This fix might be incomplete, unless I'm not doing something correctly.

I've pulled the latest version from git today primarily to get this fix and found the following issues.

I am putting all of my game objects on a "gameplay" layer.  

With this push, when I pan or scale the layer to move the camera, all the physics objects are now counter translated to stay put in the world.  The parent-child system is now broken.

It would seam like the physics world would need to be assignable to a specific layer.  Then that layer can be scaled, translated, and rotated, and all it's child nodes that have physics bodies would move with it as expected.

Or have the old behavior back but just fix the "debug drawing representations" of the physics by transforming them with node heirchy.

Or am i going about building my game wrong with the new physics system?

minggo:
@marquisd 
When an object is controlled by physics engine, it should behave as real world. Fox example, when an object fall to ground, it may rotate, if the rotation is re-set by Action system, then its behavior is strange. 

So i think when an object is controlled by physical engine, it should not affect by other things.

What do you think?

boyu0:
@marquisd If I understand correctly, what you want is a world can move with parent.
Now the logic for physical in our engine is, if you add a physics body to a node, the position and rotation of this node is take care by the physics world.
There are two problem for this:
1. If the child move and rotate both effected by physics world and it's parent, it's weird, and how we deal with the parent node has a physics body? would the node move with the parent? and what if the parent doesn't have a body?
2. Technically, I don't find a way how to move a world, both with chipmunk and box2d, the only way to move the world is to move all the body under the world, if I take care of the position for the node has a parent, it will has a performance problem, because I must take attention for all the nodes that have a parent, even they don't move by the physics world.
What's your opinion of this?

marquisd:
@minggo and @boyu, I agree with everything you both said.  

I was thinking something else.  Let me explain better.

What I thought would be beneficial is if physics objects (nodes with physics bodies) could be added as children to the same node somewhere e.i. a layer.  I’ll just call this the “gameplay layer” for my explanation.

The physics world moves/rotates the physics objects while assuming they are all in the same place; within a shared local coordinate system somewhere.  The physics world still takes care of moving/rotating the objects within a coordinate system, ignoring where that is.

Then we could take advantage of both the the parent/child behavior of cocos2d and still use the physics engine to influence the nodes that have physics.  E.i. We move, rotate, and scale that “gameplay layer” and it does not effect the physics objects local transformations within that space.  They will just render differently because when rendered, they are drawn within the node hierarchy of the game.  This is the old behavior.  Even without physics.  The physics world simply moves/rotates objects within the “gameplay layer’s” local coordinate system.

I am just wondering, with the new way, how the following can be done if the physics bodies are fixed to screen coordinates.

a) Move the camera. (by translating “gameplay layer”)

b) Zoom the camera. (by scaling “gameplay layer”)

c) Add effects to the “gameplay layer” by rotating shaking e.t.c..

d) Have non physics objects move with the world.  Like decorations and non-collidable tiles.

e) Add transforms to the “gameplay layer” as effects without affecting non-parent layers like GUI’s.  E.i. to shake the game.

I might be seeing it all wrong.  I’m just trying to pan / zoom my side scroller game with the new physics and was perplexed to see the player and enemies not move with my world; “gameplay layer”. :)

What do you guys think about this?

boyu0:
@marquisd It's a good idea, but can not to implement, because a world in chipmunk doesn't have a coordinate, and all the body in it use the absolute coordinate system. so you can't put the world in some local coordinate system like "gameplay layer", what you can do is move all the body with the layer, and, you can't scale a physics body, it's a rigid body.
I know this is a very common demand, but chipmunk doesn't provide an easy way to implement it. I will do more research for it. Thanks.

boyu0:
I created an issue here: http://cocos2d-x.org/issues/4426
I don't know how to scale or rotate the physics world, I think it's ok to do the movement.

zilongshanren:
@marquisd I think we could add Camera into our 3.0 branch, so we could achieve this effect easily by moving camera.

marquisd:
Won't moving the camera make the GUI elements move off screen too?
I thought the option of moving the physics world was an interesting alternative.

marquisd:
Is there currently a way I can move the camera in 3.0?  Can't seem to find anything about this.  Do I have to use open-gl?

darkdukey:
I agree, we need the camera.

thankcreate:
To fundamentally solve the problem of moving the world, I think we need to make PhysicsWorld related to a given Layer, not a Scene. I suspect that  the camera solution may have some problems on GUI elements when we want to move the world  layer with keeping GUI elements stay where they are.
By now , synchronization mechanism use  convertToNodeSpace and  convertToWorldSpace to update the position between Node and PhysicsBody.The trouble is that both of convertToNodeSpace and  convertToWorldSpace talk about the zeropoint-based coordinate system. (in other words , the screen of the device)
In fact ,we need to call some other functions such as convertSomeRelatedLayerPointToNodeSpace or  convertToSomeRelatedLayerSpace. After that, when we move the world layer the world will move with the line.
When realising this, I began to make some inheritance on PhysicsWorld ,PhysicsBody and Node.  But unfortunately , I found that most of the related functions in PhysicsWorld  , PhysicsBody and Node are not set as virtual which make it an impossible task to solve this problem in a clean way.
I have been stuck on this problem for some days  and I don't think that cocos2d-x team will solve this in a short time. Maybe they just think that the world should be bound to the zeropoint-based coordinate system and don't want to make things complicated. But I think this is such a real demand , especially on the platformer games.  
Even in the official platformer game sample , it still has a big bug. The layer moves on but world still stay in PointZero.  
https://github.com/cocos2d/cocos-docs/blob/master/tutorial/simple-platformer-using-cocos2d-and-box2d-with-collision-detection/

bunnyhero:
no matter how it's solved, there _needs_ to be a way for a physics-controlled "world" to be moved and scaled relative to the window. even a simple physics game like angry birds has scrolling and scaling (while the GUI elements stay in place). this is such a basic requirement that i am surprised that there's no good way to do it in 3.0 yet.

even apple's spritekit can handle this because the apple SKScene is rendered in an SKView, which doesn't have to be the root of the window. on the other hand, in Cocos2d-x, the Scene is always the root of the entire view hierarchy.

@thankcreate's idea of putting the PhysicsWorld into `Layer`, instead of `Scene`, makes sense to me.

boyu0:
@bunnyhero @thankcreate @marquisd 
Is done and merged at PR: https://github.com/cocos2d/cocos2d-x/pull/6288
Thank for your advise.

devmanny:
is working now? is included into 3.0? 
or I have to replace the modified files?

boyu0:
@MannyDev 
I add a test in that PR, you can test it.
it's not in 3.0, you can merge that PR or change your code according that PR.

devmanny:
Thank you @boyu0 now works

ocruzv:
I have the same problem of @MannyDev. This update doesn't work for me :(

ricardoquesada:
Hi,

`Scene` has a `PhysicsWorld` and it makes sense to have it that way:
- There is one `Scene` per scene, and there is one `PhysicsWorld` per scene
- And `PhysicsWorld` manages objects in "physics coordiantes". By convention we are assuming that "physics coordinates" are the same as "world coordiantes".

So, if we have this relationship (`Sprite` has a `PhysicsBody`).

`Scene -> Layer ->  Sprite`

...and if `PhysicsWorld` is in world coordiantes, then we can't move the `Layer`. If we do so, unexpected things could happen.

But what if there is a way to tell `PhysicsWorld` that it is no longer in world coordiantes, but in a "node" coordinate ?

Something like this:

``` cpp
PhysicsWorld *p = scene->getPhysicsWorld();
p->setWorldCoordinates( layer );
```

So, now, the `PhysicsWorld`, when it updates the `PhysicsBody`, it should update it based on the new "world" coordiantes.

Thoughts?

BTW, I rolled back this commit since it wasn't fixing the bug, and it was breaking compatibility.
BTW2: How is SpriteKit solving this issue ? and cocos2d-iphone ?
BTW3: A Camera should solve this issue as well, planned for v3.3

ricardoquesada:
Another possibility is that Nodes that are using a PhysicsBody always use the identity matrix for the ModelViewTransform.

bunnyhero:
cocos2d-iphone has a `CCPhysicsNode` that contains the chipmunk space/world. everything that uses physics has to be a child of that node. CCNode uses the function `NodeToPhysicsTransform` to convert the node's coordinates to physics-world coordinates:

```
static inline CGAffineTransform
NodeToPhysicsTransform(CCNode *node)
{
    CGAffineTransform transform = CGAffineTransformIdentity;
    for(CCNode *n = node; n && !n.isPhysicsNode; n = n.parent){
        transform = CGAffineTransformConcat(transform, n.nodeToParentTransform);
    }

    return transform;
}
```

the `isPhysicsNode` property returns `YES` for the root physics node (the one with the chipmunk space/world).

this means the physics world can be rooted anywhere in the scene graph (not just in the `Scene` object), and can be moved around relative to the window without affecting the physics, while non-physics nodes (like UI) can be outside of that.

ricardoquesada:
@bunnyhero 
Thanks! Interesting approach.

I'm not sure how it works in SpriteKit, but if possible I would try to avoid the `PhysicsNode` in cocos2d-x.

bunnyhero:
i think in SpriteKit, the physics world is rooted to the SKScene. unlike cocos2d-x, though, an SKScene is not the root of the entire display, so you can scroll/scale the scene but still have stationary UI elements (non-SpriteKit).

twhittock:
@ricardoquesada  - you recently rolled back the change to move PhysicsWorld to Layer. This was a valid workaround for my use case. Can you let us know what users of cocos2d-x should do now?

I am writing a side-scrolling game which has 2 sibling layers - the game level, with physics bodies in it, and an overlay layer for the HUD. By moving the game level layer to the left I was simulating a camera panning to the right. This is not possible with a PhysicsWorld on the Scene, as the physics behaves badly if I move the layer manually.

As far as I can tell, currently v3 is unable to support any physics simulation alongside a moving scene. You say that a camera will solve this in 3.3, but until then, what is the way to get a moving scene working in 3.x?

ricardoquesada:
@bunnyhero 
Interesting. Actually the `SKScene` should be the root of the scene. Mixing SpriteKit  + UIKit is not the recommended way to do games in SpriteKit.
Perhaps I'm missing something, but there must be a way to do that in SpriteKit. Or perhaps the proposed solution is to parent all the physics nodes to SKScene, including the background layer.
So, if you want to scroll the scene, you just scroll background layer.

@twhittock 
I rolled it back because the proposed solution was, as you said, a workaround. And I think we need to fix it properly in v3.1.

I said that a Camera should fix it. But actually I was wrong... A camera won't fix this bug.

The solutions that I see are:

**a):**
Add a function to `PhysicsWorld` to the the physics root node. eg:

``` c++
// By default is Scene
physicsWorld->setPhysicsRootNode(layer);

// alternative name:
physicsWorld->setWorldCoordinates(layer);
```

**b):**
Tell users to parent all the physics nodes to Scene.

**c):**
Or the cocos2d-iphone approach.
Create the `PhysicsNode`... which would be similar to our `Layer + Physics code`, but from a semantic point of view, it is much stronger (and less confusing).

**d):**
Fix it automatically, if possible.
Do the conversion between world coordinates to local coordinates automatically.

But first of all , I would try to understand how SpriteKit is fixing this issue. 
@boyu0 can you work on this ?

traitree:
@boyu0 :
what is PR ?

ricardoquesada:
@traitree 
PR means Pull Request

boyu0:
@ricardoquesada 
a) It seems a little wiled for me. all the bodies will follow that layer, even the bodies don't in that layer.
b) I don't think this is a solution, it breaks the coding style between the games with physical and without physical.
c) I prefer this solution, actually, I was thinking that if we duplicate layer, we can create a `PhysicsNode`, or we can make the `PhysicsWorld` inherits the `Node`, but now, developers are accustomed to use `Layer` to make their layout, That's why I moved `_physicsWorld` to `Layer`.
d) I don't think so, we can't let the body both followed with its parent and physics world. For example, we have two objects 'a' and 'b', 'a' and 'b' both have physics body, and both be effect with gravity, and 'b' is a child of 'a',  'a' will fall to ground, but what we do about 'b'? it will fall to ground with twice speed, I don't think this is correct.

I will check how sprite kit do about this.

ricardoquesada:
Ok. Thanks @boyu0 .
Please, let me know how SpriteKit is solving this issue once you find it out. Thanks!

boyu0:
Sprite kit use two nodes to simulate camera.
https://developer.apple.com/library/ios/documentation/GraphicsAnimation/Conceptual/SpriteKit_PG/Actions/Actions.html (see Example: Centering the Scene on a Node)

Sprite kit use d) to fix this, for the example I mentioned above, they separate 'a' and 'b' for physics simulation(like our engine),  but they make `b` to follow `a` when `a` use `setPosition`.

They use the node `world` as root instead of `scene`, add all children to the `world`, use world to simulate camera move and scale.

So if we use this strategy, I think a) is better, we need developers explicit specify a layer as "root layer", so they know they need to keep all the physics body in this root layer, and we can use that layer transform to change our physics debug draw (sprite kit doesn't have physics debug draw).

But I still prefer c), it's more senseable and much easy to understand for me, and it can create multiple physics world.

Make children follow the parent position when use parent->setPosition() is a good idea, maybe we need this too, @ricardoquesada , what's your opinion?

ricardoquesada:
Thanks for the research!
I'll study it in detail next week (I'm working on v3.1-alpha1 today)
thanks

DWWilson:
"no matter how it's solved, there needs to be a way for a physics-controlled "world" to be moved and scaled relative to the window. even a simple physics game like angry birds has scrolling and scaling (while the GUI elements stay in place). this is such a basic requirement that i am surprised that there's no good way to do it in 3.0 yet." @bunnyhero

I couldn't agree more.  My game was working fine with zooming and panning using the layer based physics that was in a few days ago, now everything is broken. Can we please go back to using physicsLayer?

boyu0:
Hi @ricardoquesada 
What's the solution about this?

ricardoquesada:
this features is planned for v3.2.
we need to have a lot of tests, but I don't have final decision yet. I'll add my comments next week.

boyu0:
@ricardoquesada 
OK, thanks.

vladubogdan:
Hello @ricardoquesada 
I also encounter this bug when creating Camera functionality for LevelHelper 2 cocos2d-x support. 
You are correct in having the physics world on the scene. The physics world should always stay in the same position. 
When the scene is moved you just need to apply a translation matrix with the position of the scene. This way the drawing of physics enabled nodes will be correct. 

Setting the position of a physics enabled node should transform that position to the world coordinates (without the scene translation) and set the body to the transformed position.

Getting the position of a physics enabled node should get the body position and transform to the local coordinates (again without the scene translation)

That is all that cocos2d-x must do and thats how SpriteKit and Cocos2d-iphone does it also.

I consider this a bug and not a features so it must be fix in all versions of Cocos2d-X with physics support (3.x)

Without the ability to move the scene or a node containing physics enabled children, cocos2d-x physics system becomes very restrictive. 

In other words, I'm very impressed with the state of cocos2d-x v3. Good job and congratulation to the Cocos2d-X team!

boyu0:
Hi @ricardoquesada ,
What's the plan about this now?

DWWilson:
I would also like to know what the plan is for the physics in 3.2...

ricardoquesada:
@boyu0 Sorry for my delayed answer. 

I like Apple's way of solving the issue.
We will need to add the missing methods to `Scene`, and it will be a good idea to add the `delegate` method to `Scene` too.

boyu0:
@ricardoquesada 
I'm a little confused, how to apply a scale to physics now?

ricardoquesada:
@boyu0 
Sorry, what do you mean: "apply scale to physics?". Could you give an example ? thanks.

makzimus:
I've had much success by removing the physics scene completely and
associating the physics world with a layer. You can then apply scale and
translation to a parent without adversely affecting the physics world.

boyu0:
@ricardoquesada I mean, if I want to scale the physics world( camera zoom in/zoom out ), how can I do this with the new plan?

@makail We have done this in PR: https://github.com/cocos2d/cocos2d-x/pull/6288
and we rejected this solution and reverted the code.

ricardoquesada:
@boyu0 

scaling the layer or scene should work. Am I missing something ?
An example will help me understand the question better.

thanks.

boyu0:
Yes, if we keep all physics body in one layer, scaling the layer can work correct, but our PhysicsDebugDraw is in the scene, it doesn't know the layer is scaled, so our debug draw is wrong.

If we scale the scene, sometimes we don't want ui layer to be scaled, but we can't do this if we scale the scene.

makzimus:
Making the PhysicsDebugDraw's DrawNode a child of the layer makes it work
with the transform hierarchy just fine.

boyu0:
@makail 
Yes, but we need to keep PhysicsWorld in Scene, so PhysicsDebugDraw need kept in Scene too, so we can't predict which layer developer use to contain the physics bodies, so we can't transform PhysicsDebugDraw with that layer, that's why PhysicsDebugDraw will be incorrect.

makzimus:
Like I said earlier, I ended up moving the PhysicsWorld to a new object
(PhysicsLayer) then when you add a node as a child it searches up
the hierarchy for a PhysicsLayer instead of PhysicsScene. Works flawlessly
for me.

ricardoquesada:
@boyu0 
For me, the best way to understand the issues that we have is to have a test case that reproduces them.

boyu0:
@ricardoquesada 
OK, I will create a branch and add a test case for it.

ricardoquesada:
ok. thanks

boyu0:
@ricardoquesada 
I changed cpp-tests->PhysicsTest->Pyramid Stack, you can test it.
The branch is: https://github.com/boyu0/cocos2d-x/tree/physics_wrong_transform_sample

marquisd:
So.... whats the plan?  As this been fixed yet?

digitalhoax:
What's the status on this? Any way to see when this has been merged (or hasn't) into the main branch (as it was merged to develop initially)?

PR #5498, opened 2014-02-28
Label lines

squirrelChen:
@stammen @dumganhar
I fixed two problem:
1. CCEditBox didn't need in native c++ project. I add a macro to it.
2. Chinese Characters couldn't seperate lines while using CCLabelTTF with dimensions. I checked it  by each word while using utf8 encoding.  US characters and Chiniese characters test ok.
Please review it.

### DIFF
--- a/samples/Cpp/HelloCpp/Classes/HelloWorldScene.cpp
+++ b/samples/Cpp/HelloCpp/Classes/HelloWorldScene.cpp
@@ -18,6 +18,18 @@ CCScene* HelloWorld::scene()
     return scene;
 }
 
+BOOL WCharToMByte(LPCWSTR lpcwszStr, LPSTR lpszStr, DWORD dwSize)
+{
+	DWORD dwMinSize;
+	dwMinSize = WideCharToMultiByte(CP_UTF8,NULL,lpcwszStr,-1,NULL,0,NULL,FALSE);
+	if(dwSize < dwMinSize)
+	{
+		return false;
+	}
+	WideCharToMultiByte(CP_UTF8,NULL,lpcwszStr,-1,lpszStr,dwSize,NULL,FALSE);
+	return true;
+}
+
 // on "init" you need to initialize your instance
 bool HelloWorld::init()
 {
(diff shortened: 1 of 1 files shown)


cocos2dx/platform/winrt/CCFreeTypeFont.cpp
```diff
@@ -359,14 +393,15 @@ FT_Error CCFreeTypeFont::initGlyphs(const char* text)
...
-        while ((pos = line.find_first_of(" ", prev)) != std::string::npos)
+		while ((pos = getNextWord(line, prev)) != std::string::npos)
         {
             if (pos > prev)
             {
-                addWord(line.substr(prev, pos-prev));
+                addWord(line.substr(prev, pos - prev));
             }
-            prev = pos + 1;
+			prev = pos;
```
dumganhar, line 402: Wrong indention, please use `4 spaces` instead of `tab`.

samples/Cpp/HelloCpp/Classes/HelloWorldScene.cpp
```diff
@@ -55,8 +67,10 @@ bool HelloWorld::init()
 
     // add a label shows "Hello World"
     // create and initialize a label
-    
-    CCLabelTTF* pLabel = CCLabelTTF::create("Hello World", "Arial", TITLE_FONT_SIZE);
+	wchar_t str[100] = {L"English 中文a 中文和英文自动换行问题 english world自动换行"};
+	char strs[200] = {0};
+	WCharToMByte(str,strs,sizeof(str)/sizeof(strs[0]));
```
dumganhar, line 72: If we keep `HelloWorldScene.cpp` as `utf8` text format, we don't need this conversion.

squirrelChen, line 72: Yes, but vs2012 will compile error while the file-encoding is utf-8 and write long chinese characters  in .cpp, so I chnage it to gb2312 to avoid compiler error.

dumganhar, line 72: Please check `LabelTest` in `testcpp`. There isn't any problem when using `utf8` string.
You could add some `spaces` or `other letters` to fix this problem. Please check it.
It's the bug of `VS`.

PR #5499, opened 2014-02-28
add CCConosle [autotest run] command

heliclei:


dumganhar:
How to stop while `autotest` is running?

heliclei:
To support autotest stop, we need to implement run in a seperated thread.

PR #5500, opened 2014-03-01
Refactors Scheduler class.

dumganhar:
- Timer is an abstract class now.
- Adds three new timers ( `TimerTargetSelector`, `TimerTargetCallback`, `TimerScriptHandler`) which are inherited from Timer.
- Adds new API `scheduleCallback` / `unscheduleCallbackForKey`, and callback's target could be any classes since we're using `void*` to identify target.
- `Scheduler::scheduleCallback` use `std::function`, therefore, it supports passing `lambda`, `member_function` and `global_function`.
- The old selector API is still kept since **it's very useful** when schedule callback is a **member function**, it doesn't need a `key` to identify the callback function. 
- Adds relevant test cases.

### DIFF
--- a/cocos/2d/CCScheduler.h
+++ b/cocos/2d/CCScheduler.h
@@ -202,7 +202,7 @@ public:
     
     /** The scheduled method will be called every 'interval' seconds.
      If paused is true, then it won't be called until it is resumed.
-     If 'interval' is 0, it will be called every frame, but if so, it's recommended to use 'scheduleUpdateForTarget:' instead.
+     If 'interval' is 0, it will be called every frame, but if so, it's recommended to use 'scheduleUpdate' instead.
      If the 'callback' is already scheduled, then only the interval parameter will be updated without re-scheduling it again.
      repeat let the action be repeated repeat + 1 times, use kRepeatForever to let the action run continuously
      delay is the amount of time the action will wait before it'll start
@@ -219,7 +219,7 @@ public:
     
     /** The scheduled method will be called every 'interval' seconds.
      If paused is true, then it won't be called until it is resumed.
-     If 'interval' is 0, it will be called every frame, but if so, it's recommended to use 'scheduleUpdateForTarget:' instead.
+     If 'interval' is 0, it will be called every frame, but if so, it's recommended to use 'scheduleUpdate' instead.
      If the selector is already scheduled, then only the interval parameter will be updated without re-scheduling it again.
      repeat let the action be repeated repeat + 1 times, use kRepeatForever to let the action run continuously
      delay is the amount of time the action will wait before it'll start
@@ -265,7 +265,7 @@ public:
     void unschedule(const std::string& key, void *target);
 
     /** Unschedule a selector for a given target.
-     If you want to unschedule the "update", use unscheudleUpdateForTarget.
+     If you want to unschedule the "update", use unscheudleUpdate.
      @since v3.0
      */
     void unschedule(SEL_SCHEDULE selector, Ref *target);


dumganhar:
@ricardoquesada @minggo 
Please review this PR. Thanks.
I really like the old API (selector / target), I have mentioned the reason above. :)

tests/Classes/SchedulerTest/SchedulerTest.cpp
```diff
@@ -1112,6 +1103,51 @@ void SchedulerIssue2268::update(float dt)
...
+{
+    SchedulerTestLayer::onEnter();
+    
+    _scheduler->scheduleCallback([](float dt){
+        log("In the callback of scheduleCallback(lambda, ...), dt = %f", dt);
+    }, this, "lambda", 1.0f, false);
+    
+    _scheduler->scheduleCallback(CC_CALLBACK_1(ScheduleCallbackTest::callback, this), this, "member_function", 1.0f, false);
+    
+    _scheduler->scheduleCallback(ScheduleCallbackTest_global_callback, this, "global_function", 1.0f, false);
```
dumganhar: The API Usages are here. :)

cocos/2d/CCScheduler.h
```diff
@@ -222,36 +243,59 @@ class CC_DLL Scheduler : public Ref
...
+     The lower the priority, the earlier it is called.
+     @since v0.99.3
+     @lua NA
+     */
     template <class T>
     void scheduleUpdateForTarget(T *target, int priority, bool paused)
     {
-        target->retain();
-        this->scheduleUpdate([=](float dt){
+        this->scheduleCallbackPerFrame([target](float dt){
```
dumganhar: `scheduleCallbackPerFrame` is only used internally. It's a protected method.

cocos/2d/CCScheduler.cpp
```diff
@@ -33,28 +33,16 @@ of this software and associated documentation files (the "Software"), to deal
...
 NS_CC_BEGIN
 
-long schedule_selector_to_key(SEL_SCHEDULE selector)
-{
-    static union{
-        SEL_SCHEDULE func;
-        long key;
-    };
-    func = selector;
-    return key;
```
dumganhar, line 47: This conversion from `selector` to `key` is not safe. 
Since `sizeof(selector)` will be 8bits while long is 4bits on 32bit system.
On 64bit system, the values are  doubled.

cocos/2d/CCScheduler.h
```diff
@@ -53,46 +51,24 @@ of this software and associated documentation files (the "Software"), to deal
 //
 class CC_DLL Timer : public Ref
 {
+protected:
+    Timer();
```
minggo, line 55: Why should be `protected` since it is an abstract class?

dumganhar, line 55: Yes. Timer is an `abstract` class now. There are three types of timer in engine. `TimerTargetSelector`, `TimerTargetCallback` and `TimerScriptHandler(for lua bindings)`.

dumganhar, line 55: The codes are clearer now. Since we don't insert `luabindings` stuff in `TImer::update`.

cocos/2d/CCScheduler.h
```diff
@@ -100,13 +76,69 @@ class CC_DLL Timer : public Ref
...
+};
+
+
+class CC_DLL TimerTargetSelector : public Timer
+{
+public:
+    TimerTargetSelector();
+
+    /** Initializes a timer with a target, a selector and an interval in seconds, repeat in number of times to repeat, delay in seconds. */
+    bool initWithSelector(SEL_SCHEDULE selector, Ref* target, float seconds, unsigned int repeat, float delay);
```
minggo, line 88: Can it be `protected`, and move parameters to constructor?

dumganhar, line 88: I don't think so. It should be public since I don't want make a `create` method for `TimerXXX`. I don't want to make it an `autorelease` object while creating.

The usage of it is

```
TimerXXX* timer = new TimerXXX;
timer->initWIthXXX(...);
...
// After being used.
timer->release()

```

cocos/2d/CCScheduler.h
```diff
@@ -53,46 +51,24 @@ of this software and associated documentation files (the "Software"), to deal
...
     /** set interval in seconds */
     inline void setInterval(float interval) { _interval = interval; };
-    /**
-     * @js NA
-     * @lua NA
-     */
-    inline const ccSchedulerFunc& getCallback() const { return _callback; };
-    inline long getKey() const { return _key; };
+    
+    void setupTimerWithInterval(float seconds, unsigned int repeat, float delay);
```
minggo, line 62: Can merge this function into constructor?

dumganhar, line 62: No. It's the common member function in `Timer` class.
`TimerTargetSelector`, `TimerXXX` will use it in `initWithXXX`.

dumganhar, line 62: Did you mean put it to the constructor of `Timer`?

dumganhar, line 62: I don't think it's a good idea since `TimerScriptHandler` class will not use all of the parameters. If I do it like you said, there would be two constructors for `Timer`.

ricardoquesada:
I like it. Yes, the old "target, selector" API is easier to understand, and makes the code cleaner.

Probably you already have done performance test. But in case you haven't yet, make sure that your changes doesn't affect the performance.

Thanks!

dumganhar:
I haven't done the performance test yet. But I think the difference is to use `std::function` or `member function`. We have used `std::function` for `CallFunc` and `EventListener`. So i think the performance will be good enough. :)  I will only compare the performance between `std::function` and `member function`. Is it correct?

ricardoquesada:
Probably the perf. is going to be good. 
It is because we are so close to the final release, that we should make sure that any change that we add does not affect the performance.

dumganhar:
OK, I will do that.

PR #5501, opened 2014-03-01
Develop nutty modify framework for315

nutty898:


cocos/gui/UIButton.h
```diff
@@ -107,6 +107,8 @@ class Button : public Widget
      */
     void setCapInsetsNormalRenderer(const Rect &capInsets);
     
+    const Rect& getCapInsetsNormalRenderer();
```
minggo, line 110: `const Rect& getCapInsetsNormalRenderer()` -> `const Rect& getCapInsetsNormalRenderer() const`?

And this issue exists in many places, please modify them.

cocos/gui/UIButton.h
```diff
@@ -131,6 +137,8 @@ class Button : public Widget
      */
     virtual void setScale9Enabled(bool able);
     
+    bool isScale9Enabled();
```
minggo, line 140: Is it a virtual function? If so, please add keyword `virtual`.

PR #5509, opened 2014-03-03
add autotest.py to jenkins-scripts

shujunqiao:
add autotest.py to jenkins-scripts for jenkins build.

### DIFF
--- a/tools/jenkins-scripts/autotest.py
+++ b/tools/jenkins-scripts/autotest.py
@@ -1,0 +1,61 @@
+import os
+import sys
+import subprocess
+import socket
+import time
+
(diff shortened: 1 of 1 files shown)


tools/jenkins-scripts/autotest.py
```diff
@@ -0,0 +1,61 @@
...
+
+def main():
+	try:
+		buildAndRun()
+	except Exception, e:
+		print 'BUILD FAILED!'
+	else:
+		autotest()
+
+main()
```
dumganhar, line 61: It's better to add

``` python
if __name__ == "__main__":
```

PR #5511, opened 2014-03-03
create install.py to set environment variables needed by cocos2d-x

minggo:
- Set environment variables: `ndk_root` and `cocos_console_root`.
- Because there is not API in python to set system environment variables, so i add it in `~/.bash_profile` on linux and Mac OS X, and modify register table on Windows.  Any other elegant way?
- After setting environment variables, any way to make it take effect immediately? I want to use the environment value to use `cocos-console` after `install.py` return, like this

``` shell
./install
cocos project new ...
```

I wanted to use `os.system()` to do it, but found it would create a new process.

### DIFF
--- a/install.py
+++ b/install.py
@@ -35,6 +35,7 @@ NDK_ROOT = 'NDK_ROOT1'
 class SetEnvVar:
 
 	current_absolute_path = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


minggo:
Should add this script to lua repo and js repo too.

PR #5517, opened 2014-03-03
Feature/fix timer cancel bug

bmanGH:
Fix Timer::cancel always call Director::getInstance()->getScheduler() even in another Scheduler;


dumganhar:
Thanks for this patch. I will check the problem. :)

dumganhar:
Thanks.

PR #5519, opened 2014-03-03
CCRepeat bug fix with repeat count == 99

gelldur:
When we set repeat count to 99 inner action is not correctly finished.
Refer: http://www.cocos2d-x.org/forums/6/topics/46503


gelldur:
Ok problem isn't only with 99 i have found more bugs simply problems with float :)

PR #5525, opened 2014-03-04
fixed #3868. Add guard check to send method of HttpClient

zilongshanren:


cocos/network/HttpClient.cpp
```diff
@@ -469,12 +469,14 @@ void HttpClient::send(HttpRequest* request)
         
     request->retain();
     
-    s_requestQueueMutex.lock();
-    s_requestQueue->pushBack(request);
-    s_requestQueueMutex.unlock();
-    
-    // Notify thread start to work
-    s_SleepCondition.notify_one();
+    if (nullptr != s_requestQueue) {
```
dumganhar, line 472: `lazyInitThreadSemphore` ensure `s_requestQueue` isn't null. Why do we need this check?

PR #5527, opened 2014-03-04
Disable cceditbox

squirrelChen:


dumganhar:
Why to disable `EditBox`?

squirrelChen:
editbox just used in xaml project, I remove it from native c++ proj, native c++ do not support.

PR #5534, opened 2014-03-04
Develop render texture improvement

dabingnn:
This pull request add two features for RenderTexture
1. add `_keepMatrix`, if it is true, we will use current modelView and projection matrix instead of generate a new one.
2. add `setVirtualViewport` interface, we can only grab fullscreen RenderTexture by default, by using this interface, we can only grab part of screen to a texture.

### DIFF
--- a/cocos/2d/CCRenderTexture.h
+++ b/cocos/2d/CCRenderTexture.h
@@ -157,7 +157,13 @@ public:
     virtual void visit(Renderer *renderer, const kmMat4 &parentTransform, bool parentTransformUpdated) override;
     virtual void draw(Renderer *renderer, const kmMat4 &transform, bool transformUpdated) override;
 
+    //flag: use stack matrix computed from scene hierarchy or generate new modelView and projection matrix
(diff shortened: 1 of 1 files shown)


dabingnn:
@ricardoquesada @minggo @dumganhar please take a look at this pull request, thank you.

cocos/2d/CCRenderTexture.h
```diff
@@ -157,6 +157,9 @@ class CC_DLL RenderTexture : public Node
     virtual void visit(Renderer *renderer, const kmMat4 &parentTransform, bool parentTransformUpdated) override;
     virtual void draw(Renderer *renderer, const kmMat4 &transform, bool transformUpdated) override;
 
+    void setKeepMatrix(bool keepMatrix);
+    void setVirtualViewport(const Point& rtBegin, const Rect& fullRect, const Rect& fullViewport);
```
minggo, line 167: Please write doxygen doc for these new two functions.

Thanks.

PR #5544, opened 2014-03-04
CCNode:Add String tag, to support runtime manipulation

heliclei:


cocos/base/CCConsole.cpp
```diff
@@ -810,9 +810,23 @@ bool Console::parseCommand(int fd)
...
+                scriptEngine->executeString(buf);
+            }
+            else
+            {
+                CCLOG("No Script Engine found!");
+            }
+        });
+        
+
     }
```
ricardoquesada: The script should be a command too, and not the default option. The command name could be `run` or `exec` or `script` or `s`.
eg:
`
exec  var a =10 * 10;
`

cocos/2d/CCNode.h
```diff
@@ -1470,7 +1511,9 @@ class CC_DLL Node : public Ref
     Vector<Node*> _children;        ///< array of children nodes
     Node *_parent;                  ///< weak reference to parent node
 
-    int _tag;                       ///< a tag. Can be any number you assigned just to identify this node
+    int _tag;                         ///< a tag. Can be any number you assigned just to identify this node
+    
+    char _strTag[64];               ///<a string label, a user defined string to identify this node
```
ricardoquesada: why not an `std::string` instead ?
It takes less memory, and is less error prone.

cocos/2d/CCNode.h
```diff
@@ -823,6 +832,38 @@ class CC_DLL Node : public Ref
...
+     */
+    virtual char* getStrTag();
+    /**
+     * Changes the string TAG that is used to identify the node easily.
+     *
+     * Please refer to getLabel for the sample code.
+     *
+     * @param strTag   A C String that indentifies the node.
+     */
+    virtual void setStrTag(const char* strTag);
```
ricardoquesada: please, use `std::string` instead.
And how can I search for a node globally ?

ricardoquesada: If we are going to support string tags, then I think we should support something like this API to search nodes:
https://developer.apple.com/library/ios/documentation/GraphicsAnimation/Conceptual/SpriteKit_PG/Nodes/Nodes.html#//apple_ref/doc/uid/TP40013043-CH3-SW17

heliclei: I have added a function Node\* getChildByStrTag(const char\* label, bool recursive) to search nodes

heliclei: Maybe we can rename it to childNodeWithName(...), and change _strTag to name

cocos/2d/CCNode.h
heliclei: A pre-allocated char[] avoid memory issues, I think it is quicker and more robust.

cocos/2d/CCNode.h
minggo: I think it is better to rename
`setName`, `getName` and `getChildByName()`.

cocos/2d/CCNode.h
ricardoquesada: it is not because:
- All nodes, even if they don't use a name, will allocate 64 bytes more. Bytes are very expensive in mobile devices
- what if the names is bigger than 64 bytes ?
- It is difficult to manipulate `char*`.

That can be solved by using `std::string`.

cocos/2d/CCNode.h
ricardoquesada: My feedback:
If we are going to add names to Nodes, we should add a good API, like SpriteKit's.
Please, take a look at that API. It is great.

cocos/2d/CCNode.h
heliclei: You convinced me:) I will change char\* to std::string.

cocos/2d/CCNode.h
heliclei: Thanks, we will have another good API soon.

minggo: Yep, we need to refer to SpriteKit.
And this API is useful for debugging.

cocos/2d/CCNode.h
```diff
@@ -823,6 +832,38 @@ class CC_DLL Node : public Ref
...
+    /**
+     * Returns a label that is used to identify the node easily.
+     *
+     * You can set tags to node then identify them easily.
+     @code
+ 
+     // set tags
+     node1->setStag("player");
+     node2->setStag("monster");
+     node3->setStag("boss");
```
dumganhar: The comments should be updated too.

cocos/2d/CCNode.h
```diff
@@ -823,6 +832,38 @@ class CC_DLL Node : public Ref
...
+     *
+     * @return A C string that identifies the node.
+     */
+    virtual std::string& getName();
+    /**
+     * Changes the string TAG that is used to identify the node easily.
+     *
+     * Please refer to getLabel for the sample code.
+     *
+     * @param strTag   A C String that indentifies the node.
```
dumganhar: `C String` -> `string`

cocos/base/CCConsole.cpp
dumganhar: @heliclei , what's your opinion?

cocos/2d/CCNode.h
heliclei: forgive my mindlessness, I'll correct the comments

cocos/base/CCConsole.cpp
heliclei: I think add an 'exec' or 'run' seems a bit verbose

ricardoquesada: Please, be more verbose in your comment :-)

Yes, you need to type more... so it is more verbose. So what ?
If I type `ehlp` I expect to receive the message `command "ehlp" not found. type "help" for more info`.
If you put the scripts as the default it will say:
`variable ehlp couldn't be found...` or something like that. Very confusing for the user.

I designed the console in a way that is easy to use for our users. If you are going to change that behavior, please explain in a detailed manner why you think your solution is better. Be verbose in your description. thanks.

heliclei: A big  advantage to not add a exec command is we can use pipleline commands
such as [telnet localhost 5678<<test.lua] or [cat script.cmd | telnet localhost 5678] 
so please consider this option even if it violate the initial design consideration.

ricardoquesada: @heliclei 
Now we are talking. That's the way to propose a feature :)

The way I see it, is that the feature that you are proposing is useful for testing and/or automating things, but not for users who want to use the console.

Perhaps there should be another way to "pipe" commands.

dumganhar: > The way I see it, is that the feature that you are proposing is useful for testing and/or automating things, but not for users who want to use the console.

I agree with @ricardoquesada 's point. `console` should not contain default behavior of parsing string as lua script. What about adding a new command like `script start | end`, it will parse string as lua | js script between `start` and `end`.

heliclei: Thanks, Riq:) It's my honor to know you also think this is a useful feature. I think we can managed to meet both needs of users and automation, as we continuously improving the console.

ricardoquesada: cool :)

heliclei: @ricardoquesada , How about add a ScriptConsole to do lua/js stuff?

ricardoquesada: So, the console was created in a way that it is easy to extend.
I think that adding an `exec` command for scripts is perfect for our users.

But if you need something that can be "piped" there are two options:
- You can prepend the word `exec` before sending the script. Just don't use telnet for that... use `nc` (netcat). That should work for you.
- But if for some reason it doesn't work, just create a new object that listens to another TCP port, and you redirect everything to the console by prepending the word "exec".

Remember, netcat (nc) and the unix command line tools are your friends for automation :)

heliclei: Thanks~ for CCConsole exec command, it's great to add [exec string\n] & [exec file\n], but for the piped version, my question is, do we have to prefix an 'exec' at every line? I love nc either, It's a great tool:)

PR #5561, opened 2014-03-05
Move Box2dTestBed step method to onDraw.

boyu0:
Because Box2dTestBed draws some elements in Step, and we invoke glClear after update, so some Box2dTestBed shows nothing.
issue link:http://cocos2d-x.org/issues/4282

### DIFF
--- a/tests/Classes/Box2DTestBed/Box2dView.cpp
+++ b/tests/Classes/Box2DTestBed/Box2dView.cpp
@@ -176,9 +176,7 @@ Box2DView* Box2DView::viewWithEntryID(int entryId)
 }
 
 bool Box2DView::initWithEntryID(int entryId)
-{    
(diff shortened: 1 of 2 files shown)


tests/Classes/Box2DTestBed/Box2dView.cpp
```diff
@@ -204,7 +204,6 @@ bool Box2DView::initWithEntryID(int entryId)
 
 void Box2DView::tick(float dt)
 {
-    m_test->Step(&settings);
```
dumganhar, line 207: So `Box2DView::tick` isn't needed? Should we delete that schedule?

boyu0, line 207: OK，I will delete this.

PR #5569, opened 2014-03-05
fix bug: when bouncable is false, scroll view can't fling

stubma:
When CCScrollView set bouncable to false, it can't fling, here is the fix


stubma:
没有看到有merge, 不接受至少也给个理由吧, 就这么close算什么意思

PR #5590, opened 2014-03-06
Fix LableBMFont is not use padding

waiter:
LableBMFont have not use padding, so the word space is not correct. 
if the padding is very big, you will see the error.

I fix it in the cocos2d-x 2.2, I have not test 3.0, but I think it should be like that!

### DIFF
--- a/cocos/2d/CCFontFNT.cpp
+++ b/cocos/2d/CCFontFNT.cpp
@@ -413,7 +413,7 @@ std::set<unsigned int>* BMFontConfiguration::parseBinaryConfigFile(unsigned char
              */
 
             uint16_t lineHeight = 0; memcpy(&lineHeight, pData, 2);
-            _commonHeight = lineHeight;
+            _commonHeight = lineHeight - _padding.top - _padding.bottom;
 
             uint16_t scaleW = 0; memcpy(&scaleW, pData + 4, 2);
             uint16_t scaleH = 0; memcpy(&scaleH, pData + 6, 2);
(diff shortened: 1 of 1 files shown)


minggo:
Thanks for the patch.
We will use new Label to re-write LabelBMFont.

So how about close this pull request and polish it with new Label?
Thanks.

waiter:
@minggo 
What does the "New Label" mean?  I pull the request to the Develop branch. And I modify the "CCFontFnt.cpp" in this branch. 
And another pull request is pulled to the Master branch ( https://github.com/cocos2d/cocos2d-x/pull/5591 )

cocos/2d/CCFontFNT.cpp
```diff
@@ -413,7 +413,7 @@ void BMFontConfiguration::purgeFontDefDictionary()
              */
 
             uint16_t lineHeight = 0; memcpy(&lineHeight, pData, 2);
-            _commonHeight = lineHeight;
+            _commonHeight = lineHeight - _padding.top - _padding.bottom;
```
WenhaiLin, line 416: Is the padding  increase line height  in version 1.9【Block type 2】?The padding only increase size of character in version 1.13. I can't found old version for test it.

waiter, line 416: @Dhilan007 
I have no idea with "Block type 2". 
I use "Hiero V3.0" to make FNT file. And I test in cocos2d-x 2.2, The padding increase the line height.

cocosQA:
Dear waiter:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6454

PR #5592, opened 2014-03-06
modify autotest.py

shujunqiao:
Modify autotest.py add android part.

### DIFF
--- a/tools/jenkins-scripts/autotest.py
+++ b/tools/jenkins-scripts/autotest.py
@@ -4,12 +4,27 @@ import subprocess
 import socket
 import time
 
-HOST = 'localhost'
+HOST_MAC = 'localhost'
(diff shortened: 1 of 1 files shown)


tools/jenkins-scripts/autotest.py
```diff
@@ -4,12 +4,27 @@
 import socket
 import time
 
-HOST = 'localhost'
+HOST_MAC = 'localhost'
+HOST_ADNROID = '10.10.30.64'
+HOST_IOS = '10.10.30.61'
```
heliclei, line 9: Why not read the host IPs from command line arguments or env viariables?

shujunqiao, line 9: oh, it's a better choice. i will modify it.
It's just for test in a fixation mashine now, so i'm lazy, sorry.

PR #5601, opened 2014-03-06
Fix touch event bug.

palange:
A global object tracking touch events is now cleared when CCEGLView::end() is called, on Android.


palange:
Meant for my own fork...

PR #5614, opened 2014-03-07
Master nutty modify framework for315

nutty898:


cocos2dx/label_nodes/CCLabelBMFont.cpp
```diff
@@ -588,6 +588,7 @@ void CCLabelBMFont::createFontChars()
     unsigned int stringLen = m_sString ? cc_wcslen(m_sString) : 0;
     if (stringLen == 0)
     {
+        this->setContentSize(CC_SIZE_PIXELS_TO_POINTS(tmpSize));
```
WenhaiLin, line 591: The content size of node have an initial value[Zero].Set content size of LabelBMFont is unnecessary when length of string is zero.

nutty898, line 591: It's a bug here.
LabelBMFont dose not do any thing with its content size when we set a string value with zero length. The "_contentSize" would keep its value(width and height) last time, it couldn't be zero size.

WenhaiLin, line 591: You are right.

PR #5622, opened 2014-03-07
Issue #4294: Add unimplemented position getter/setter in MotionStreak

pandamicro:


cocos/2d/CCMotionStreak.h
```diff
@@ -74,6 +74,13 @@ class CC_DLL MotionStreak : public Node, public TextureProtocol
 
     // Overrides
     virtual void setPosition(const Point& position) override;
+    virtual void setPosition(float x, float y);
```
dumganhar: This function should also use `override` keyword.

cocos/2d/CCMotionStreak.cpp
```diff
@@ -143,6 +143,42 @@ void MotionStreak::setPosition(const Point& position)
...
+}
+
+float MotionStreak::getPositionY() const
+{
+    return  _positionR.y;
+}
+
+void MotionStreak::setPositionY(float y)
+{
+    setPosition(Point(_positionR.x, y));
```
dumganhar: Why not just invoke `_positionR.y = y`?

cocos/2d/CCMotionStreak.cpp
```diff
@@ -143,6 +143,42 @@ void MotionStreak::setPosition(const Point& position)
...
+}
+
+float MotionStreak::getPositionX() const
+{
+    return _positionR.x;
+}
+
+void MotionStreak::setPositionX(float x)
+{
+    setPosition(Point(x, _positionR.y));
```
dumganhar: `_positionR.x = x` will be better.

cocos/2d/CCMotionStreak.cpp
```diff
@@ -143,6 +143,42 @@ void MotionStreak::setPosition(const Point& position)
     _positionR = position;
 }
 
+void MotionStreak::setPosition(float x, float y)
+{
+    setPosition(Point(x, y));
```
dumganhar:
``` c++
_positionR.x = x;
_positionR.y = y;
```

will be more efficient.

PR #5641, opened 2014-03-09
setTouchPriority doesn't work due to a bug in CCTouchDispatcher

adamfischer:
The bug was introduced in this fix: http://cocos2d-x.org/issues/752
I created an issue in issue tracker too: http://www.cocos2d-x.org/issues/4317
This is my first pull request, I am not sure if the fix is correct, but the issue certainly exists. The problem is that CCLayer's setTouchPriority method sometimes has no effect. The fix was created by wondeful18.

Test case:

I created a Layer, which has a sole purpose to block ("swallow";) touches, and this feature can be turned on and off.
The class is very basic, if it receives the touch it always swallows it:

`bool BlockingLayer::init(){

// Super init.
if ( !CCLayer::init() )
{
return false;
}

setTouchEnabled(true);
setTouchMode(kCCTouchesOneByOne);
setTouchPriority(INT_MAX);

return true;
}

bool BlockingLayer::ccTouchBegan(CCTouch *pTouch, CCEvent *pEvent)
{
CCLOG("BlockingLayer swallowed touch!");
return true;
}

So by default it has a really bad priority, it receives touches if no other class claimed it. But in the scene where I am using this layer I would like to set it to a different priority when certain events occur:

bool MyScene::init(int unitNumber, CCString\* path){
// Super init.
...
blockingLayer = BlockingLayer::create();
this->addChild(blockingLayer);

return true;
}

bool MyScene::ccTouchBegan(CCTouch *pTouch, CCEvent *pEvent){
blockingLayer->setTouchPriority(INTMIN);
}`

Now the layer should have the best priority possible, so it should swallow all touches. But it does not, it's behaviour does not change.
I see its registerWithTouchDispatcher() called and the m_nTouchPriority changed correctly. BUT the layer's behaviour is unchanged.

### DIFF
--- a/cocos2dx/touch_dispatcher/CCTouchDispatcher.cpp
+++ b/cocos2dx/touch_dispatcher/CCTouchDispatcher.cpp
@@ -141,8 +141,12 @@ void CCTouchDispatcher::addStandardDelegate(CCTouchDelegate *pDelegate, int nPri
          */
         if (ccCArrayContainsValue(m_pHandlersToRemove, pDelegate))
         {
-            ccCArrayRemoveValue(m_pHandlersToRemove, pDelegate);
-            return;
+            CCTouchHandler *pOldHandler = findHandler(pDelegate);
+            if (pOldHandler && pOldHandler->getPriority() == nPriority)
+            {
+                ccCArrayRemoveValue(m_pHandlersToRemove, pDelegate);
+                return;
+            }
         }
 
         m_pHandlersToAdd->addObject(pHandler);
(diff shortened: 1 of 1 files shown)


adamfischer:
If the priority is changed we have the CCTouchHandler present both in m_pHandlersToRemove and m_pHandlersToAdd. But in CCTouchDispatcher::touches we first iterate m_pHandlersToRemove and than m_pHandlersToAdd. So the CCTouchHandler stays and there is no memory leak.

squirrelChen:
@minggo  @dumganhar  I test Ok in this issue, forexample: when you try to modify the other object's priority in the ccTouchBegan event and try to let the object respond, need to add the condition. And just as @adamfischer said, it will not cause memory leak.

PR #5663, opened 2014-03-10
Winrt build fix

stammen:
Fixed broken winrt and wp8 builds due to recent pull requests:
1. Added missing CocosStudio files
2. Added missing CocosStudio tests
3. fixed #pragma mark (only available for IOS and MAC)
4. fixed LUA build warnings
5. consistend build settings for all projects.

### DIFF
--- a/scripting/lua/proj.winrt/liblua.vcxproj
+++ b/scripting/lua/proj.winrt/liblua.vcxproj
@@ -190,6 +190,9 @@
       <IgnoreAllDefaultLibraries>false</IgnoreAllDefaultLibraries>
       <GenerateWindowsMetadata>false</GenerateWindowsMetadata>
     </Link>
+    <Lib>
+      <AdditionalOptions>/IGNORE:4264 %(AdditionalOptions)</AdditionalOptions>
+    </Lib>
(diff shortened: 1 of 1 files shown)


cocos2dx/script_support/CCScriptSupport.cpp
```diff
@@ -37,8 +37,10 @@ bool CC_DLL cc_assert_script_compatible(const char *msg)
 
 NS_CC_BEGIN
 
+ #if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
 // #pragma mark -
 // #pragma mark CCScriptHandlerEntry
+#endif
```
dumganhar, line 43: Hi, these two lines are commented.

stammen:
Not sure what you mean. Whoever checked in the #pragma mark - code broke the build. #pragma mark - is for xcode only. I just added the platform check so whoever is responsible for #pragma mark - can commented or not comment the #pragma mark - and not break other platforms.

dumganhar:
You're correct. Thanks for the patch.

PR #5671, opened 2014-03-11
Add enumerateChildrenByName, walk node tree with callback

heliclei:


cocos/2d/CCNode.h
```diff
@@ -789,18 +789,26 @@ class CC_DLL Node : public Ref
...
      *
      * @param name  A  string that indentifies the node.
      */
     virtual void setName(const std::string& name);
     /**
+     * Search the children to perform processing for nodes which share a name.
+     * @param name  the name to search for
+     * @param callback, A callback to execute on nodes that match the name parameter. The callback takes the following arguments:
+     *  node: A node that matches the name.
+     *  stop: A pointer to a Boolean variable. Your callback can set this to YES to terminate the enumeration.
```
dumganhar: `YES` -> `true`

cocos/2d/CCNode.h
```diff
@@ -789,18 +789,26 @@ class CC_DLL Node : public Ref
...
+     */
+     virtual void enumChildNodesByName(const std::string& name, std::function<void(Node* node, bool* stop)> callback);
```
dumganhar: Please use `const &` for `callback` parameter.

samples/cpp-tests/Classes/NodeTest/NodeTest.cpp
```diff
@@ -354,6 +355,70 @@ void Test6::addAndRemove(float dt)
...
+    addChild(sp11);
+    addChild(sp12);
+    addChild(sp20);
+    addChild(sp21);
+    addChild(sp22);
+
+    auto fn1 = std::bind(&NodeEnumChildByNameTest::runAction1, this, std::placeholders::_1, std::placeholders::_2);
+    auto fn2 = std::bind(&NodeEnumChildByNameTest::runAction2, this, std::placeholders::_1, std::placeholders::_2);
+    
+    enumChildNodesByName("sister1", fn1);
```
dumganhar: How about using `lambda` for this test? That will simplify our codes.

heliclei: good idea~

tests/cpp-tests/Classes/NodeTest/NodeTest.cpp
```diff
@@ -354,6 +355,65 @@ void Test6::addAndRemove(float dt)
...
+
+    addChild(sp10);
+    addChild(sp11);
+    addChild(sp12);
+    addChild(sp20);
+    addChild(sp21);
+    addChild(sp22);
+
+    enumChildNodesByName("sister1", 
+        [&](Node* node, bool* stop)
```
dumganhar: I think `&` isn't needed.

heliclei: correct~

heliclei:
[ci rebuild]

cocos/2d/CCNode.cpp
```diff
@@ -733,6 +732,25 @@ Node* Node::getChildByName(const std::string& name)
     return nullptr;
 }
 
+void Node::enumChildNodesByName(const std::string& name, const std::function<void(Node* node, bool* stop)>& callback)
```
minggo: Node::enumChildNodesByName -> Node::emulateChild ?

minggo: Node::emulateChildren?

ricardoquesada:
guys, this is a great feature.
But was it needed for v3.0 ?
I really do thing that names are better that tags, but adding names now it means that we need to deprecate tags... since names replace tags.

So, we need to deprecate the `addChild(child, z, tag)` API... deprecate `setTag()`, deprecate `getTag()`....

darkdukey:
Maybe use tag for searching a group of node?

ricardoquesada:
mmmm... confusing. why both ? which is better?  users are going to ask those questions and more.

we need to keep the API simple:
- simple to understand
- simple to maintain

From the Zen of python:
http://legacy.python.org/dev/peps/pep-0020/

```
There should be one-- and preferably only one --obvious way to do it.
```

heliclei:
Thanks, Riq.

heliclei:
I think deprecate tag is a good idea.

dumganhar:
Yep, since we don't have enough time for the coming release, we will remove this feature.
We need to think more about the pros and cons of `tag` and `name` in the version after the coming release. :)

dumganhar:
@ricardoquesada , sorry for the merge. :)

heliclei:
OK, I will revert this PR.

ricardoquesada:
@dumganhar no problem
This is going to be nice feature.... and we want to promote it once it is ready. So it makes sense to postpone, and add it v3.1. 

TODO:
- We need more tests (including performance tests)
- update all our samples.
- Update actions APi (?)
- what else ??? Does the new API conver all the search patterns supported by SpriteKit ?

PR #5680, opened 2014-03-11
Fixing memory leak in CCSpriteFrameCache

gelldur:
If we add again the same Plist file we will have ~200KB of memory leak.
Ofcourse it will be cleared in some time but the case it when we reload
Plist for example in restartGame() function.

Example:

//We want clean memory every game start

CCSpriteFrameCache::sharedSpriteFrameCache()->purgeSharedSpriteFrameCache();
    CCTextureCache::sharedTextureCache()->purgeSharedTextureCache();

///////////////////////////////////////////////////////////////////////////////////////////////

LevelConfiguration config = getCurrentLevelConfiguration();
CCTextureCache::sharedTextureCache()->addImage(config.textureFileName.c_str());

CCSpriteFrameCache::sharedSpriteFrameCache()->addSpriteFramesWithFile(config.plistFileName.c_str(),config.textureFileName);

Also when someone use another function than
void CCSpriteFrameCache::addSpriteFramesWithFile(const char *pszPlist);

we don't add to loaded file names:
m_pLoadedFileNames->insert(pszPlist);


dumganhar:
Good finding. :+1:

PR #5681, opened 2014-03-11
Fix gui to ui

zilongshanren:


tools/cocos2d-console
```diff
@@ -1 +1 @@
-Subproject commit 7e7f6d215c9bb83ac926a1436b20a685e3d5b66b
+Subproject commit 39738950219adba48ae97e853673278337df4658
```
dumganhar: Why to change the submodule reference?

dumganhar:
Thanks.

PR #5682, opened 2014-03-11
-Fixed an issue in HttpClient where calling destroyInstance and then usi...

adamfischer:
Fixed an issue in HttpClient where calling destroyInstance and then using the HttpClient class again sometimes freezes the application.

### DIFF
--- a/cocos/network/HttpClient.cpp
+++ b/cocos/network/HttpClient.cpp
@@ -445,10 +445,10 @@ bool HttpClient::lazyInitThreadSemphore()
         s_requestQueue = new Vector<HttpRequest*>();
         s_responseQueue = new Vector<HttpResponse*>();
         
+        s_need_quit = false;
+        
         auto t = std::thread(CC_CALLBACK_0(HttpClient::networkThread, this));
         t.detach();
-        
-        s_need_quit = false;
     }
     
     return true;


adamfischer:
Test case: http://www.cocos2d-x.org/issues/3986

After successfully using HttpClient and then calling destroyInstance s_need_quit becomes true, and the while(true) in networkThread breaks, static resources (s_requestQueue, s_responseQueue etc...) are freed. This is fine.

Then using HttpClient again,then lazyInitThreadSemphore is called, then background thread is detached which calls networkThread. The problem is that s_need_quit is set false _after_ dispatching the background thread. Sometimes s_need_quit is still true when networkThread first checks it, which makes it immediately free its resources. Then the application freezes, because these resources are NULL, but their methods are called later (in send for example).

If s_need_quit is set false _before_ the background thread is dispatched this never happens.

zilongshanren:
Hi @adamfischer ,
Thanks for you contribution.
Please recheck the latest version of cocos2d-x 3.0, the bug should be solved some days ago

adamfischer:
Thanks. I checked the last 3 commits here: https://github.com/cocos2d/cocos2d-x/commits/1772044fe5ed9af725a9b126d8ce13bcc2a0bfc2/cocos/network/HttpClient.cpp

1772044fe5ed9af725a9b126d8ce13bcc2a0bfc2
503edb4da40ecab5057e2967958b31632a845e7d
944d6c85ce83e9809884200294aacc95465ddeac

I think they are remedies, and those null pointer checks make the application stop freezing, but the root cause of the problems is not fixed. If s_need_quit is set properly, than the null pointers don't happen at the first place. ("//occurs when cocos thread fires but the network thread has already quited") And because of the null checks sometimes valid send messages are ignored.

This is just my 0.02$. :)

cocosQA:
Dear adamfischer:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6448

PR #5688, opened 2014-03-11
Deployment iOS >= 5.1.1 (needed for ARM64)

ricardoquesada:
Adds also Valid signature = iOS Developer


dumganhar:
Hi Riq, Thanks for this fix. I have a question:
Why doesn't my xcode have a `5.1.1` target?
Do I need to install something?

ricardoquesada:
Mmm... I don't know. Have you installed Xcode 5.1 ?

What I did was:
- remove the target platform... and then re-add it again... and Xcode told me: "Hey, if you want to create a fat binary with ARM64 then the deployment target should iOS >= 5.1.1. Should I fix it for you ?"
  and I clicked on "Yes".

PR #5689, opened 2014-03-11
Pvr support

newnon:
software pvr decompression support
small touches processing improvements
getCurrentLanguageCode() function

### DIFF
--- a/cocos/2d/CCEventListenerTouch.h
+++ b/cocos/2d/CCEventListenerTouch.h
@@ -28,6 +28,7 @@
 
 #include "CCEventListener.h"
 #include "CCTouch.h"
+#include "CCEventTouch.h"
 
 #include <vector>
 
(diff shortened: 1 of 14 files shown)


cocos/2d/CCEventListenerTouch.h
```diff
@@ -51,10 +52,10 @@ class EventListenerTouchOneByOne : public EventListener
...
 
 public:
-    std::function<bool(Touch*, Event*)> onTouchBegan;
-    std::function<void(Touch*, Event*)> onTouchMoved;
-    std::function<void(Touch*, Event*)> onTouchEnded;
-    std::function<void(Touch*, Event*)> onTouchCancelled;
+    std::function<bool(Touch*, EventTouch*)> onTouchBegan;
+    std::function<void(Touch*, EventTouch*)> onTouchMoved;
+    std::function<void(Touch*, EventTouch*)> onTouchEnded;
+    std::function<void(Touch*, EventTouch*)> onTouchCancelled;
```
dumganhar, line 58: Does it need to use `EventTouch*` as the second parameter for touch callback?
Because developer only takes care of `Touch*` or `const vector<Touch*>&`.

newnon, line 58: i added wary important param to EventTouch now you can get information on all active touches. It's very useful for multitouch support, you don't need need store all touches inside you control you just can take it from second param.
it is like UIEvent in iPhone sdk 
– allTouches
https://developer.apple.com/library/ios/documentation/uikit/reference/UIEvent_Class/Reference/Reference.html
if you don't change it to EventTouch you need additional dinamic_cast it to EventTouch

PR #5690, opened 2014-03-12
issue #3821:LabelTTF re-implemented as a wrapper of Label.

WenhaiLin:


cocos/2d/CCLabelTTF.cpp
```diff
@@ -108,449 +84,189 @@ LabelTTF * LabelTTF::createWithFontDefinition(const std::string& string, FontDef
...
-        _fontSize = fontSize;
-        
-        // Force update
-        if (_string.size() > 0)
-        {
-            this->updateTexture();
-        }
-    }
+    _renderLabel->setFontSize(fontSize);
+    _contentDirty = true;
```
dumganhar, line 167: If the font size passed in is the same as the old one, should we still set the dirty flag?

cocos/2d/CCLabelTTF.cpp
```diff
@@ -108,449 +84,189 @@ LabelTTF * LabelTTF::createWithFontDefinition(const std::string& string, FontDef
...
-    
-    // set the size in the sprite
-    Rect rect =Rect::ZERO;
-    rect.size   = _texture->getContentSize();
-    this->setTextureRect(rect);
-    
-    //ok
-    return true;
+    _renderLabel->setFontName(fontName);
+    _contentDirty = true;
```
dumganhar, line 178: The same here.

PR #5703, opened 2014-03-12
label:fixed crash and incorrect getter.

WenhaiLin:
1.fixed crash if create label by font name and text is empty.
2.fixed getter of font name and font size is incorrect.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -940,7 +940,7 @@ void Label::updateContent()
 
 void Label::visit(Renderer *renderer, const kmMat4 &parentTransform, bool parentTransformUpdated)
 {
-    if (! _visible || _originalUTF8String.length() == 0)
+    if (! _visible || _originalUTF8String.empty())
     {
         return;
     }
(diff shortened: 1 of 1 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -938,7 +940,7 @@ void Label::updateContent()
 
 void Label::visit(Renderer *renderer, const kmMat4 &parentTransform, bool parentTransformUpdated)
 {
-    if (! _visible)
+    if (! _visible || _originalUTF8String.length() == 0)
```
dumganhar: Please use `_originalUTF8String.empty()` to check whether string is empty.

cocos/2d/CCLabel.cpp
```diff
@@ -999,6 +1001,19 @@ void Label::setFontName(const std::string& fontName)
...
 }
 
+const std::string& Label::getFontName() const
+{
+    switch (_currentLabelType)
+    {
+    case cocos2d::Label::LabelType::TTF:
+        return _fontConfig.fontFilePath;
+    case cocos2d::Label::LabelType::STRING_TEXTURE:
+        return _fontDefinition._fontName;
```
dumganhar: This `case` is the same as `Default`. So it isn't needed.

dumganhar: BTW, `cocos2d::Label` prefix isn't needed too.

PR #5707, opened 2014-03-12
[Jenkins]Add watchdog job, to check & kill dead buid

linchaoJack:


tools/jenkins-scripts/watchdog.py
```diff
@@ -0,0 +1,47 @@
...
+from jenkinsapi.jenkins import Jenkins
+import sys
+import time
+import os
+
+def main():
+    
+    J = Jenkins('http://115.28.134.83:8000','redmine','cocos2dx2013')
+    job = J['cocos-2dx-pull-request-build']
+    duration = os.environ['duration']
```
heliclei, line 12: I think 'duration' is not very good, I suggest use 'jenkins-job-watchdog-threshold', it is more self-explain.

linchaoJack:
Leave a comment

PR #5732, opened 2014-03-12
Fixing bug when scheduler was blocked and we want remove and add again s...

gelldur:
...cheduled method

Example

We scheduled teget for example method update

void MyObject::update(float dt)
{
    //Here CCScheduler::m_bUpdateHashLocked is set to true
    //Also here scheduler for LeafAnimator::getInstance() <b>is
paused!</b>

CCDirector::sharedDirector()->getScheduler()->unscheduleUpdateForTarget(&LeafAnimator::getInstance()
);

CCDirector::sharedDirector()->getScheduler()->scheduleUpdateForTarget
(&LeafAnimator::getInstance(), INT_MAX, false );
//So when we want reschedule paused field isn't updated. Priority also
but this is in todo :)
}

### DIFF
--- a/cocos2dx/CCScheduler.cpp
+++ b/cocos2dx/CCScheduler.cpp
@@ -470,6 +470,7 @@ void CCScheduler::scheduleUpdateForTarget(CCObject *pTarget, int nPriority, bool
 #endif
         // TODO: check if priority has changed!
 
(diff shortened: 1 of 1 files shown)


gelldur:
Why it don't compile? My commit i correct.

PR #5760, opened 2014-03-13
Pvr support

newnon:


newnon:
How can i get information about build errors?
i tested it on 2 computers and everything works

PR #5765, opened 2014-03-13
add cocos-console-test.py to tools/jenkins-scripts

shujunqiao:
add and modify cocos-console-test.py to tools/jenkins-scripts.


shujunqiao:
for test console:[console], don't need merge.

PR #5767, opened 2014-03-13
modify invalid syntax of cocos-console-test.py.

shujunqiao:
modify invalid syntax of cocos-console-test.py.


shujunqiao:
modify cocos-console-test.py for trigger console-test job.

PR #5768, opened 2014-03-13
Refactor setter of label's font name and font size.

WenhaiLin:


cocos/2d/CCLabel.cpp
```diff
@@ -208,15 +208,15 @@ bool Label::setCharMap(const std::string& plistFile)
...
-    if (initWithFontAtlas(newAtlas))
     {
-        _currentLabelType = LabelType::CHARMAP;
-        return true;
+        reset();
+        return false;
     }
 
-    return false;
+    initWithFontAtlas(newAtlas);
```
dumganhar: Don't need to check the return value of `initXXX`?

PR #5782, opened 2014-03-13
64bit support

xhcnb:
rebuild libpng, libjpeg, libwebp, libfreetype, libtiff with arm64 support
upgrade libpng to 1.6.9
details:
iOS support:  armv7 armv7s arm64 i386 x86_64
Mac support: x86_64 i386


minggo:
@xhcnb 
Thanks.
But it also meets the problem as #5769, the modification will lead to compiling errors when using Xcode 5.1.

Thoughts?

minggo:
Sorry, it will meets compiling error other than Xcode 5.1.

xhcnb:
These libs I commited all built with xcode 5.1 and OXSDK 10.9,IOS 7.1I can complie my proj smoothly to 64bit simulator or 64bit ios device.
I notied libcurl and libwebsockets are still only support 32bit, my game don't use these two, that's why I have no error.
I will rebuild libcurl and libwebsockets, upload theme to github.

minggo:
I mean, if you don't use Xcode 5.1, suppose using Xcode 5.0, the pull request will cause compiling error.

So i think we can not merge it, because not all developers use Xcode 5.1.

minggo:
Any thoughts?

xhcnb:
Yes, you are right, now, I can't test with xcode 5.0
libcurl and libwebsockets are uploaded.
I will create an rep for these libs and there .xcodeprojs for anyone who needs.

minggo:
Thanks.
Is there any way to support 64 bits, but not break the compatibility?

minggo:
Now it is a fat lib for all architectures. We are trying to break it into small lib, each one for an architecture.
And configure the build setting, make it link different lib according to architecture the app will run on.

I am not sure if it can work for both Xcode 5.1 and previous versions.

xhcnb:
These fat libs I create with `lipo -create`, and there are difference between `ios i386, x86_84` and `mac i386, x86_84`

xhcnb:
becasue they are built with different sdks, `ios i386, x86_84` use ios simulator, `mac i386, x86_84` use OSX sdk.
I think we can rebuild theme with xcode 5.0 for 32bit and xcode 5.1 for 64bit, then make an fat lib
this will not break the compatibility.
I'll try it later.

minggo:
Yep. We want to break it, as we did for android https://github.com/cocos2d/cocos2d-x/tree/develop/external/jpeg/prebuilt/android.

Try to fix the issue that break Xcode 4.6 and Xcode 5.0.

xhcnb:
Seprate different  arch is good.

minggo:
Yep. We fixed it by doing this in #5769.
So i think we can close this pull request.

PR #5788, opened 2014-03-14
Label:fixed text display incorrect when come to foreground on android.

WenhaiLin:


cocos/2d/CCFontAtlas.cpp
```diff
@@ -27,16 +27,22 @@
...
 #include "CCDirector.h"
+#include "CCEventListenerCustom.h"
+#include "CCEventDispatcher.h"
+#include "CCEventType.h"
 
 NS_CC_BEGIN
 
 const int FontAtlas::CacheTextureWidth = 1024;
 const int FontAtlas::CacheTextureHeight = 1024;
+const char* FontAtlas::EVENT_PURGE_TEXTURES = "FontAtlasPurgeTextures";
```
dumganhar: Please add `__cc_` prefix to avoid conflicts.

PR #5793, opened 2014-03-14
Crash while clicking close button on desktop platforms

squirrelChen:
glview release in GLView::end(), but still have to use later in Application::run()


dumganhar:
So Mac and Linux will also have this problem. Please also apply this patch to those platforms. Thanks.:)

squirrelChen:
@dumganhar  ok, now

PR #5817, opened 2014-03-14
Fix potential overflow in Texture::getCachedTextureInfo and where uses MultiByteToWideChar and WideCharToMultiByte

dumganhar:


dumganhar:
[ci rebuild]

PR #5823, opened 2014-03-15
Ignore invisible widget touch events

woc2006:
Skip invisible elements when dispatching touch events

### DIFF
--- a/cocos/ui/UIWidget.cpp
+++ b/cocos/ui/UIWidget.cpp
@@ -671,6 +671,9 @@ void Widget::didNotSelectSelf()
 
 bool Widget::onTouchBegan(Touch *touch, Event *unusedEvent)
 {
+    //ignore touch event on invisible elements.
+    if(!_visible)
+        return;
     _hitted = false;
     if (isEnabled() && isTouchEnabled())
     {


cocos/ui/UIWidget.cpp
```diff
@@ -671,6 +671,9 @@ void Widget::didNotSelectSelf()
 
 bool Widget::onTouchBegan(Touch *touch, Event *unusedEvent)
 {
+    //ignore touch event on invisible elements.
+    if(!_visible)
```
edwardzhou, line 675: I tested. not work.
only check itself _visible doesn't enough. because, itself's _visible can be true, but it's parent's (event parent's  parent's ...) _visible can be false. 
for a simple workaround, check up to 5 levels parents is enough for most of UI cases.

```
int upLevel = 0;
auto widget = this;
while (widget != nullptr and upLevel < 5) {
    if (!(widget->isEnabled() && widget->isVisible()))
        return false;
    widget = widget->getWidgetParent();
    upLevel++;
}
```

edwardzhou, line 675: Menu did the similar way. maybe you can adopt it.

bool Menu::onTouchBegan(Touch\* touch, Event\* event)
{
    if (_state != Menu::State::WAITING || ! _visible || !_enabled)
    {
        return false;
    }

```
for (Node *c = this->_parent; c != nullptr; c = c->getParent())
{
    if (c->isVisible() == false)
    {
        return false;
    }
}

_selectedItem = this->getItemForTouch(touch);
if (_selectedItem)
{
    _state = Menu::State::TRACKING_TOUCH;
    _selectedItem->selected();

    return true;
}

return false;
```

}

PR #5826, opened 2014-03-15
console variable init

chuanweizhang2013:


cocos/base/CCConsole.cpp
```diff
@@ -252,6 +252,8 @@ void log(const char * format, ...)
 , _running(false)
 , _endThread(false)
 , _sendDebugStrings(false)
+,_file_uploading(false)
+,_upload_file_size(0)
```
minggo: _upload_file_size -> _uploadFileSize
_file_uploading -> _fileUploading

PR #5832, opened 2014-03-16
event listener improvements

newnon:
sotware pvr decompressions suport
event listener improvements
getAllTouches() - it is like UIEvent in iPhone sdk  – allTouches
getCurrentLanguageCode() - get iso 639-1 code
iOS template improvement(useful for game center integration or some libs that need RootViewController)


dumganhar:
Thanks. Could you please split this PR to 2 different PRs?
One for software PVR  decompressions support, 
Another for the touch change.

newnon:
ok i will make it today an make a new pull request

PR #5835, opened 2014-03-17
build support for Release|Win32 and Release|ARM

squirrelChen:


squirrelChen:
late for the fix submit..

PR #5098, opened 2014-01-17
Added libcurl, libwebsockets, openssl and updated testcpp to support these on Windows Phone 8

martell:
EDIT: WinRT will go in another PR
This should close #3870

PR #5851, opened 2014-03-18
test pr #5098 and merge confilct

squirrelChen:


.gitmodules
```diff
@@ -19,3 +19,15 @@
...
+[submodule "cocos2dx/platform/third_party/winrt/libcurl"]
+	path = cocos2dx/platform/third_party/winrt/libcurl
+	url = https://github.com/martell/libcurl-winrt.git
+[submodule "cocos2dx/platform/third_party/winrt/winrtsock"]
+	path = cocos2dx/platform/third_party/winrt/winrtsock
+	url = https://github.com/martell/winrtsock.git
+[submodule "cocos2dx/platform/third_party/winrt/libwebsockets"]
+	path = cocos2dx/platform/third_party/winrt/libwebsockets
+	url = https://github.com/martell/libwebsockets.git
+[submodule "cocos2dx/platform/third_party/winrt/openssl"]
```
dumganhar: Do you use the source codes for these third party libraries?
If so, we need to build it to static libraries.

martell: Hey dumganhar,
The third_party libraries are done the same way dale stamen has done the angle project.
We should at the end make static libs for all of them. For now I just matched his structure.

martell:
@dumganhar koowolf has removed the submodules here and added them as static libs.
I didn't notice that previously. Could you merge this PR please?
Dale said to me in another thread that he will do the same for libangle and the other specific wp8 libraries during the week.

squirrelChen:
@martell I used them as static libraries, but it takes a lot of time to wait while generaling code. I think had better to use dynamic libraries, just wait for @stammen to submit.

PR #5856, opened 2014-03-18
Elastic bounceback for ScrollView

youknowone:
Add iOS-like elastic bounceback.
If you scroll content of scroll view out of insets,
its content will be scrolled 0.35 of user scrolled distance.

### DIFF
--- a/extensions/GUI/CCScrollView/CCScrollView.cpp
+++ b/extensions/GUI/CCScrollView/CCScrollView.cpp
@@ -679,10 +679,18 @@ void ScrollView::onTouchMoved(Touch* touch, Event* event)
             if (_direction == Direction::VERTICAL)
             {
                 dis = moveDistance.y;
+                float pos = _container->getPosition().y;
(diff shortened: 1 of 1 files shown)


extensions/GUI/CCScrollView/CCScrollView.cpp
```diff
@@ -679,10 +679,18 @@ void ScrollView::onTouchMoved(Touch* touch, Event* event)
             if (_direction == Direction::VERTICAL)
             {
                 dis = moveDistance.y;
+                float pos = _container->getPosition().y;
+                if (!(minContainerOffset().y <= pos && pos <= maxContainerOffset().y)) {
+                    dis *= 0.35;
```
dumganhar, line 684: Thanks. I think here should be `moveDistance.y * = 0.35`.

dumganhar:
I modified some thing based on your PR. 
Please refer to https://github.com/cocos2d/cocos2d-x/pull/5869

youknowone:
Thank you for your help ;)

PR #5858, opened 2014-03-18
enable setup.py to use on fish

sambaiz:
setup, not only bash but also fish

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ https://github.com/cocos2d/cocos2d-console
 * ANDROID_SDK_ROOT: used to generate applicatoin on Android through commands
 * ANT_ROOT: used to generate applicatoin on Android through commands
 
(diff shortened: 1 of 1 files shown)


dumganhar:
Hi, thanks for your contribution.
As i know, fish could read the variables defined in `bash`, right?
So if we invoking `setup.py` in bash, fish will also be able to use it.

sambaiz:
> As i know, fish could read the variables defined in bash, right?

I don't think so by default settings at least.
After run `setup.py` , and  run `echo $COCOS_CONSOLE_ROOT` on fish but I get nothing.
Even after I restart fish, the result is the same.

minggo:
Why need to do special support for fish?
Should we support `Zsh`, `Ksh`?
Support as more as possible is good to have. But it brings more work, test.
And we don't have enough resource for it.
Just one choice and continue to improve it is enough.

sambaiz:
OK. I see.

PR #5861, opened 2014-03-19
fix bug: voice recognition input crash on ios7

dpull:
https://github.com/cocos2d/cocos2d-x/pull/5263
for V2.x


dumganhar:
[ci rebuild]

PR #5867, opened 2014-03-19
closed #4429, resolved Boolean value could not be written to specified plist file

zilongshanren:
http://www.cocos2d-x.org/issues/4429


zilongshanren:
[ci rebuild]

dumganhar:
Float and double written to plist will lost precision on Win32.

PR #5876, opened 2014-03-20
Fix warnings

zilongshanren:


cocos/ui/UIRichText.cpp
```diff
@@ -340,7 +340,7 @@ void RichText::formarRenderers()
         float newContentSizeHeight = 0.0f;
         float *maxHeights = new float[_elementRenders.size()];
         
-        for (size_t i=0; i<_elementRenders.size(); i++)
+        for (int i=0; i<_elementRenders.size(); i++)
```
dumganhar: This line should be reverted.

PR #5895, opened 2014-03-20
Remove NativeActivity

splhack:


splhack:
re-send the pull request as https://github.com/cocos2d/cocos2d-x/pull/5896

PR #5921, opened 2014-03-21
closed #4497: use /z7 to build without .pdb

squirrelChen:
use /z7 to build without .pdb, or will face with error c3130.


minggo:
Any side-effect to remove .pdb file?

squirrelChen:
I run appliction ok, and as the link :http://msdn.microsoft.com/zh-cn/library/958x11bc.aspx said, it just has some difference of output.

dumganhar:
But `ProgramDatabase` is used in release mode for other targets. Why should `cpp-tests` change it but others not?

squirrelChen:
only cpp-tests face with error c3130, some .cpp in cpp-tests will face with this problem, but without any document about how to solve it...

stevetranby:
Why did this get closed? Shouldn't we strive to allow for Edit and Continue when debugging?

PR #5924, opened 2014-03-22
Windows Phone 8,0 cocos2d-x 3.0  changes

stammen:
This pull request contains all of the changes to the cocos2d-x 3.0 cross platform code needed to support Windows Phone 8 (WP8). The code for WP8 will be submitted in a separate pull request. Cocos2d-x 3.0 is now running very well on Windows Phone 8,0.

The following changes were made to the cross platform code:
1. Added CC_PLATFORM_WP8 and CC_PLATFORM_WINRT
2. Disabled ccconsole code for WP8 and WINRT until networking code is added.
3. Fixed a texture leak in FontAtlasCache. The texture was being overwritten when a new char was added to the texture. The previous texture was not being destroy. See comments added to code.
   Moved fix to Texture2D::initWithMipmaps as requested by reviewer.
4. Disabled webp code for WP8 and WINRT. MSOpenTech is not able to contribute webp code.
5. Added .wav support to audio. WP8 does not support .mp3
6. Fixed use of nullptr in variable arg lists on WP8 as requested by reviewer.
7. Fixed use of compile reserved works GENERIC, ABSOLUTE and RELATIVE for WP8 compiler
8. surrounds all #pragma marks with CC_PLATFORM_IOS and MAC to avoid the breaking on builds on other platforms.
9. Added the purging of the FontAtlasCache in CCDirector.
10. Please note that some tests are leaking textures in the texture cache. For example, ActionsTest exits with 15 refcounts on grossini.png so it is not being released.

Thank you for reviewing our pull request.

### DIFF
--- a/build/cocos2d_libs.xcodeproj/project.pbxproj
+++ b/build/cocos2d_libs.xcodeproj/project.pbxproj
@@ -63,6 +63,9 @@
 		1A0DB7321823827C0025743D /* CCGL.h in Headers */ = {isa = PBXBuildFile; fileRef = 1A0DB7301823827C0025743D /* CCGL.h */; };
 		1A0DB7331823827C0025743D /* CCEAGLView.h in Headers */ = {isa = PBXBuildFile; fileRef = 1A0DB7311823827C0025743D /* CCEAGLView.h */; };
 		1A0DB7381823828F0025743D /* CCGL.h in Headers */ = {isa = PBXBuildFile; fileRef = 1A0DB7351823828F0025743D /* CCGL.h */; };
+		1A12775A18DFCC4F0005F345 /* CCTweenFunction.h in Headers */ = {isa = PBXBuildFile; fileRef = 2986667918B1B079000E39CA /* CCTweenFunction.h */; };
+		1A12775B18DFCC540005F345 /* CCTweenFunction.h in Headers */ = {isa = PBXBuildFile; fileRef = 2986667918B1B079000E39CA /* CCTweenFunction.h */; };
+		1A12775C18DFCC590005F345 /* CCTweenFunction.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 2986667818B1B079000E39CA /* CCTweenFunction.cpp */; };
 		1A570061180BC5A10088DEC7 /* CCAction.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1A570047180BC5A10088DEC7 /* CCAction.cpp */; };
 		1A570062180BC5A10088DEC7 /* CCAction.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1A570047180BC5A10088DEC7 /* CCAction.cpp */; };
 		1A570063180BC5A10088DEC7 /* CCAction.h in Headers */ = {isa = PBXBuildFile; fileRef = 1A570048180BC5A10088DEC7 /* CCAction.h */; };
@@ -1224,8 +1227,6 @@
 		2905FA8D18CF08D100240AA3 /* UIWidget.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 2905FA1318CF08D100240AA3 /* UIWidget.cpp */; };
 		2905FA8E18CF08D100240AA3 /* UIWidget.h in Headers */ = {isa = PBXBuildFile; fileRef = 2905FA1418CF08D100240AA3 /* UIWidget.h */; };
 		2905FA8F18CF08D100240AA3 /* UIWidget.h in Headers */ = {isa = PBXBuildFile; fileRef = 2905FA1418CF08D100240AA3 /* UIWidget.h */; };
-		2986667C18B1B079000E39CA /* CCTweenFunction.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 2986667818B1B079000E39CA /* CCTweenFunction.cpp */; };
-		2986667D18B1B079000E39CA /* CCTweenFunction.h in Headers */ = {isa = PBXBuildFile; fileRef = 2986667918B1B079000E39CA /* CCTweenFunction.h */; };
 		2986667F18B1B246000E39CA /* CCTweenFunction.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 2986667818B1B079000E39CA /* CCTweenFunction.cpp */; };
 		2AC795DB1862870F005EC8E1 /* SkeletonBounds.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 2AC795D918628689005EC8E1 /* SkeletonBounds.cpp */; };
 		2AC795DC1862870F005EC8E1 /* Event.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 2AC795D51862867D005EC8E1 /* Event.cpp */; };
@@ -5073,6 +5074,7 @@
 				A044DEAF18C6A58700B6CCBD /* kazmath.h in Headers */,
 				2905FA7618CF08D100240AA3 /* UIScrollView.h in Headers */,
 				1A5702C4180BCE2A0088DEC7 /* CCIMEDispatcher.h in Headers */,
+				1A12775B18DFCC540005F345 /* CCTweenFunction.h in Headers */,
 				1A5702CA180BCE370088DEC7 /* CCTextFieldTTF.h in Headers */,
 				A044DEB718C6A58700B6CCBD /* mat4.h in Headers */,
 				1A5702D5180BCE570088DEC7 /* CCTexture2D.h in Headers */,
@@ -5341,7 +5343,6 @@
 				46A160491807A5D4005B8026 /* CDXMacOSXSupport.h in Headers */,
 				46A160441807A5D4005B8026 /* CDAudioManager.h in Headers */,
 				46A1604E1807A5D4005B8026 /* SimpleAudioEngine_objc.h in Headers */,
-				2986667D18B1B079000E39CA /* CCTweenFunction.h in Headers */,
 				46A160431807A5D4005B8026 /* SimpleAudioEngine.h in Headers */,
 				46A1604B1807A5D4005B8026 /* CocosDenshion.h in Headers */,
 			);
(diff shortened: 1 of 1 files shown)


minggo:
I think you can define nullptr as NULL on WP8 and windows 8. Replace nullptr to NULL is not a good idea. Because nullptr is more safe than NULL in other platforms.

cocos/2d/ccCArray.cpp
```diff
@@ -278,8 +278,10 @@ void ccArrayFullRemoveArray(ccArray *arr, ccArray *minusArr)
 	arr->num -= back;
 }
 
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
 // #pragma mark -
 // #pragma mark ccCArray for Values (c structures)
+#endif
```
minggo, line 284: I think we can remove these codes.

stammen, line 284: Yes, but someone keeps accepting pull requests with #pragma mark uncommented which breaks all of the non mac/ios builds. I would rather have protection against this.

stammen:
I can't redefine nullptr on WP8. I get the error

#error The C++ Standard Library forbids macroizing keywords. 

So it needs to be NULL or we need to define an end of list marker, something like

#if (CC_TARGET_PLATFORM == CC_PLATFORM_WP8)
#define CC_LIST_END NULL
#else
#define CC_LIST_END nullptr
#endif

Then we can do something like

 Sequence::create(arg1, arg3, CC_LIST_END);

dungmv:
wp8 and winrt still use angelproject?

cocos/2d/ccCArray.cpp
minggo, line 284: Ok, no problem, i think we can remove them after merging. And we will not accept it any more.

minggo:
Why WP8 compiler cannot terminate variable arg list with nullptr?

stammen:
Yes. wp8 and winrt still uses angle. I will be providing precompiled dlls for angle so a git submodule is no longer required.

stammen:
It appears to be a bug in the compiler. The error is "cannot cast nullptr to ..."

dungmv:
angle has many of the issue as memory, performance... so in the future which we will use angle for cocos2d-x forever or it will change and use directx directly?
I hope it will change to use directx directly
thank you.

stammen:
How could we use DirectX directly when all of cocos2d uses OpenGL? Angle is needed to translate OpenGL to DirectX. We are working to improve memory and performance on Angle for Windows Phone 8.

cocos/2d/ccCArray.cpp
minggo, line 284: Does #pragma mark cause build problem on WRT?
Because it doesn't have any problem when building Android or linux.

dungmv:
So cocos2d-x must change base core to be used directx directly.
thank so much.

minggo:
Use nullptr and NULL together is a bit confused. Use `CC_LIST_END` is also not so elegant. And it is easy to cause problem, because developers will pass nullptr too.

So is it possible to fix it in the compiler? Because it doesn't have problem on windows.

cocos/2d/CCDirector.cpp
```diff
@@ -447,7 +459,15 @@ void Director::setProjection(Projection projection)
...
-
+            
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WP8
+            //if needed, we need to add a rotation for Landscape orientations on Windows Phone 8 since it is always in Portrait Mode
+            GLView* view = getOpenGLView();
+            if(getOpenGLView() != nullptr)
+            {
+                kmGLMultMatrix(getOpenGLView()->getOrientationMatrix());
+            }
+#endif
```
minggo, line 470: What does it mean? WP8 can not support landscape?

cocos/2d/CCDirector.cpp
```diff
@@ -485,10 +505,16 @@ void Director::setProjection(Projection projection)
...
+    FontAtlasCache::purgeCachedData();
+
     if (s_SharedDirector->getOpenGLView())
     {
         SpriteFrameCache::getInstance()->removeUnusedSpriteFrames();
         _textureCache->removeUnusedTextures();
+
+        // Note: some tests such as ActionsTest are leaking refcounted textures
+        // There should be no test textures left in the cache
+        log("%s\n", _textureCache->getCachedTextureInfo().c_str());
```
minggo, line 517: log for debugging?

stammen:
I left the log in because cpp-tests are leaking textures. For example, look at ActionsTest. It leaks grossini with a refcount of 12-15. Other tests are leaking as well. You should leave this in until all tests release all textures from the texture cache.

stammen:
This is the same code as Cocos2d-x 2.2. WP8 does not have a runtime shader compiler. WinRT under VS2012 and WIndows 8.0 does not have access to a runtime shader compiler. WinRT under VS2013 and Windows 8.1 does have access to a runtime shader compiler. Windows Phone 8.1 will also have a runtime shader compiler.

Developers have to precompile their shaders for WP8 and WinRT under Windows 8.0 I will be providing a shader compiler tool in the build/winrt directory.

cocos/2d/CCGLProgram.cpp
```diff
@@ -152,9 +170,34 @@ bool GLProgram::initWithByteArrays(const GLchar* vShaderByteArray, const GLchar*
...
+    haveProgram = CCPrecompiledShaders::getInstance()->loadProgram(_program, vShaderByteArray, fShaderByteArray);
+
+    CHECK_GL_ERROR_DEBUG();
+    _hashForUniforms = NULL;
+
+    CHECK_GL_ERROR_DEBUG();  
+
+    return haveProgram;
+}
+#endif
```
minggo, line 199: How about merge the implementation into GLProgram::initWithByteArrays().

WenhaiLin:
I think delete texture in `Texture2D::initWithMipmaps` if name is not 0 is more reasonable. Check name and delete texture before generate texture.What's your opinion?

minggo:
I think now, the biggest issue of the pull request is that, the compiler can not support terminate variable arg list with nullptr.

minggo:
Any other method to resolve the problem?

stammen:
I will look into the nullptr issue on Monday morning (Seattle time).

minggo:
Thanks.

cocos/2d/CCActionInterval.h
```diff
@@ -93,7 +93,26 @@ class CC_DLL Sequence : public ActionInterval
 {
 public:
     /** helper constructor to create an array of sequenceable actions */
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_WP8)
+    // WP8 in VS2012 does not support nullptr in variable args lists and variadic templates are also not supported
```
minggo, line 97: So it is a bug of compiler?

stammen, line 97: Yes. The compiler error is "cannot convert nullptr to ...". Also redefined nullptr to NULL breaks other code. I used a solution that  is similar to VS2012 variadic template support.

minggo, line 97: The compiler doesn't want to resolve it?

stammen, line 97: No. It can only resolve NULL or (void*)nullptr but not nullptr.

minggo, line 97: I mean the team that develop the compiler doesn't want to resolve it? Because it is a bug.

stammen, line 97: Don't know. Hopefully it will be fixed in the Windows Phone 8.1 compiler.

cocos/2d/CCActionInterval.cpp
```diff
@@ -159,6 +159,19 @@ Sequence* Sequence::createWithTwoActions(FiniteTimeAction *actionOne, FiniteTime
...
+    va_list params;
+    va_start(params, action1);
+
+    Sequence *ret = Sequence::createWithVariableList(action1, params);
+
+    va_end(params);
+    
+    return ret;
+}
+#else
```
minggo, line 174: What's the difference compare to `Sequence::create(FiniteTimeAction *action1, ...)`?

cocos/2d/ccCArray.cpp
stammen, line 284: It breaks the Win32 build and the WinRT and WP8 builds. Visual Studio does not support it.

minggo, line 284: We have it for a long time. And it doesn't have problem in win32.

cocos/2d/CCActionInterval.cpp
stammen, line 174: C++ cannot choose the correct function override by function parameters alone. Sequence::create(FiniteTimeAction *action1, ...) would be confused with the other create functions used by the WP8 version.

minggo, line 174: Oh, no. Why WP8 has so many limitation? It makes engine ugly, and how can developers use it? Use macro in his codes too?

cocos/2d/ccCArray.cpp
stammen, line 284: I just uncommented a #pragma mark in the win32 build

4>c:\users\dalestam\documents\github\cocos2d-x-2.0\cocos2dx\layers_scenes_transitions_nodes\cclayer.h(191): warning C4068: unknown pragma

I suppose I could set the build to ignore C4068:errors/warnings

minggo, line 284: No problem. We will move these codes.

cocos/2d/CCActionInterval.cpp
stammen, line 174: The  limitation is the nullptr bug in the VS2102 WP8 compiler for variable arguments lists.
No macros are needed by the developer for the create functions. Up to 10 items are supported by create without any macros needed by the developer.

If IOS or Android compilers had this problem you would change the code to use NULL. We are wasting time wishing the nullptr bug in the WP8 compiler would go away. Please feel free to suggest other options but I have spent many hours trying different things. Thank you for reviewing this.

minggo, line 174: nullptr is a standard of c++11, all other platforms support it other than WP8. Because it makes engine ugly. And as you said, developer can not use `Sequence::create()`, he should use `Sequence:: variadicCreate ()`. Because developers want to make their games run on many targets. So they should use macro to control it.

Right?

stammen, line 174: No. Look at all of the code in cpp-tests where create() is used. There is NO change in the code. NO macros. It is just sequence::create() or menu::create(). They would ONLY need to put NULL at the end of a list of arguments LONGER than 10. This all compiles on WP8, WinRT, Win32 and the other platforms.

minggo, line 174: Ok, please modify the codes when the compiler fixes the bug.

stammen, line 174: Will do. Thanks for all of your help and advice and the excellent review of this code. I will submit the WP8 projects and files in a separate pull request.

minggo, line 174: Thanks.

PR #5925, opened 2014-03-22
cocos2d_libs.xcodeproj: Add CCTweenFunc.cpp to "cocos2dx Mac" target.

wefiends:
CCTweenFunc.cpp is not included in the mac target. 

The reason the tests still build is because it is in CocosDenshion, so the symbols are present. However, if you are just using the cocos2dx-mac static lib, Ease Actions will fail to link.


dumganhar:
Thanks. I also modified it for iOS. It was fixed at https://github.com/cocos2d/cocos2d-x/pull/5927.
I closed this PR.  :)

PR #5931, opened 2014-03-24
Support to clip transparent area of Label, content size will be updated with only non-transparent area.

WenhaiLin:


cocos/2d/CCLabel.h
```diff
@@ -217,6 +217,10 @@ class CC_DLL Label : public SpriteBatchNode, public LabelProtocol
 
     virtual Sprite * getLetter(int lettetIndex);
 
+    /** clip upper and lower margin for reduce height of label.
+     */
+    void setClipMarginEnabled(bool clipEnabled) { _clipEnabled = clipEnabled; }
```
dumganhar, line 222: Shall we need a `getter` function?

PR #5933, opened 2014-03-24
port from 3.x: fix WebSocket cannot send/receive more than 4096 bytes of...

lihex:
... data

### DIFF
--- a/cocos2dx/Android.mk
+++ b/cocos2dx/Android.mk
@@ -19,6 +19,7 @@ actions/CCActionGrid.cpp \
 actions/CCActionGrid3D.cpp \
 actions/CCActionInstant.cpp \
 actions/CCActionInterval.cpp \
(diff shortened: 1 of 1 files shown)


extensions/network/WebSocket.cpp
```diff
@@ -592,32 +625,64 @@ int WebSocket::onSocketCallback(struct libwebsocket_context *ctx,
...
+
+						data->bytes = bytes;
+						data->len = _currentDataLen;
+						msg->obj = (void*)data;
+
+						CC_SAFE_DELETE_ARRAY(_currentData);
+						_currentData = NULL;
+						_currentDataLen = 0;
+
+						_wsHelper->sendMessageToUIThread(msg);
```
dumganhar: Thanks. @lihex 
Please fix the wrong indention.

PR #5935, opened 2014-03-24
reduce Win32 lib release size

squirrelChen:


minggo:
Why close it?

squirrelChen:
zhangcheng changed libextensions from lib to dll in master branch.

PR #5936, opened 2014-03-24
closed #4530: Using 'Outside stroke' instead of 'Inside Stroke' for Label generated by 'Font name'

dumganhar:
And  fix
- Stroke could not work on iOS7
- Stroke size and shadow offset don't consider ContentScaleFactor

### DIFF
--- a/tests/cpp-tests/Classes/LabelTest/LabelTest.cpp
+++ b/tests/cpp-tests/Classes/LabelTest/LabelTest.cpp
@@ -1552,6 +1552,37 @@ TTFFontShadowAndStroke::TTFFontShadowAndStroke()
     // add label to the scene
     this->addChild(fontStrokeAndShadow);
     fontStrokeAndShadow->setPosition(Point(s.width/2,s.height/4*1.1));
+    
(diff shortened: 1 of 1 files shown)


cocos/2d/platform/ios/CCDevice.mm
```diff
@@ -322,8 +322,8 @@ static bool _initWithString(const char * text, cocos2d::Device::TextAlign align,
         }
         
         // add the padding (this could be 0 if no shadow and no stroke)
-        dim.width  += shadowStrokePaddingX;
-        dim.height += shadowStrokePaddingY;
+        dim.width  += shadowStrokePaddingX*2;
+        dim.height += shadowStrokePaddingY*2;
```
minggo, line 326: Why multiply 2 here?

dumganhar, line 326: If stroke size is `1 pixel`, the rectangle area of text will extend 2 pixels.
For example, |Test| ,  `|` is 1 pixel stroke. Both its left, right, top, down edge will extend 1 pixel.
So the whole rectangle will be `width + 2` , `height + 2`.

minggo, line 326: Got it.
Thanks.

PR #5940, opened 2014-03-24
closed #4513: Particle doesn't animate when readd it.

boyu0:


tests/cpp-tests/Classes/ParticleTest/ParticleTest.cpp
```diff
@@ -1863,6 +1877,8 @@ void PremultipliedAlphaTest::onEnter()
     _emitter->setEndColorVar(Color4F(0, 0, 0, 0));
 
     this->addChild(_emitter, 10);
+    
+    schedule(schedule_selector(PremultipliedAlphaTest::readdPaticle), 1.0);
```
dumganhar: 1.0 -> 1.0f, otherwise, win32 will generate `warnings`.

PR #5941, opened 2014-03-24
Fix Android build instructions

AbandonedCart:
Resolves the "cocos compile: error: argument -p/--platform: expected one argument" error


dumganhar:
Thanks.

PR #5943, opened 2014-03-25
closed #4542: WebsocketTest crashes on win32, it may delete mutex which is still being locked

squirrelChen:
@dumganhar it just crash in win32. It delete mutex in destructor while it still lock. So after get the message from queue, unlock it.


dumganhar:
Thanks.

PR #5944, opened 2014-03-25
Fix ActionObject memory leak bug

aeonmine:


aeonmine:
This bug already fixed on 3d-extension branch , but v3rc0 did not

PR #5972, opened 2014-03-25
fix for sub pixel rendering problem of Labels (Bug #4503)

thoechsmann:
When CENTER alignment is set for a Label, characters are drawn on sub pixels resulting in unsharp rendering.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -725,7 +725,8 @@ void Label::updateQuads()
                 _reusedLetter->setBatchNode(_batchNodes[letterDef.textureID]);
             _reusedLetter->setTextureRect(_reusedRect,false,_reusedRect.size);
 
-            _reusedLetter->setPosition(_lettersInfo[ctr].position);
+            _reusedLetter->setPosition(static_cast<int>(_lettersInfo[ctr].position.x),
+                                       static_cast<int>(_lettersInfo[ctr].position.y));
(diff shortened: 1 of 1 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -725,7 +725,8 @@ void Label::updateQuads()
                 _reusedLetter->setBatchNode(_batchNodes[letterDef.textureID]);
             _reusedLetter->setTextureRect(_reusedRect,false,_reusedRect.size);
 
-            _reusedLetter->setPosition(_lettersInfo[ctr].position);
+            _reusedLetter->setPosition(static_cast<int>(_lettersInfo[ctr].position.x),
+                                       static_cast<int>(_lettersInfo[ctr].position.y));
```
WenhaiLin, line 729: I don't think this is a good solution.Maybe you can try enabling/disabling antialias for the font.[label->getFontAtlas()->setAliasTexParameters()]

thoechsmann, line 729: you might be right that there are better ways to fix that. Actually it might be fixed already by some other commit. I cannot reproduce the problem anymore.

cocosQA:
Dear thoechsmann:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6436

PR #5896, opened 2014-03-21
Remove NativeActivity

splhack:

PR #5975, opened 2014-03-25
HttpClient - add static method to enable and disable ssl support

jsermeno:
The version of libcurl included with cocos2d-x supports ssl. You can currently use curl directly to support ssl, but there doesn't seem like any reason why HttpClient shouldn't support https. Currently HttpClient disables ssl checks which creates a security vulnerability for those who are unaware. Add `?w=1` to this pull request to view without whitespace changes. These files originally included a lot of trailing spaces.

This pull request adds two functions you would use like this

```
    auto sharedFileUtils = FileUtils::getInstance();
    std::string ret = sharedFileUtils->fullPathForFilename("ca-chain.crt");

    cocos2d::network::HttpClient::enableSSLCaInfo(ret); // function 1 to enable
    cocos2d::network::HttpClient::disableSSLCaInfo(); // function 2 to disable
```

The functions are static since enable or disabling https is something you either want or do not want.

### DIFF
--- a/cocos/network/HttpClient.cpp
+++ b/cocos/network/HttpClient.cpp
@@ -417,7 +417,7 @@ void HttpClient::destroyInstance()
     CC_SAFE_DELETE(s_pHttpClient);
 }
 
-void HttpClient::enableSSLCaInfo(std::string info)
+void HttpClient::enableSSLCaInfo(const std::string& info)
 {
     s_sslCaInfo = info;
 }
--- a/cocos/network/HttpClient.h
+++ b/cocos/network/HttpClient.h
@@ -54,7 +54,7 @@ public:
     static void destroyInstance();
 
     /** Enable ssl support **/
-    static void enableSSLCaInfo(std::string info);
+    static void enableSSLCaInfo(const std::string& info);
 
     /** Disable ssl support **/
     static void disableSSLCaInfo();


jsermeno:
What is with the build system? It's not deterministic? It's failing, but it looks like it's due to a chipmunk like error. The time before a different os failed also related to a chipmunk linkage error.

```
16>PreLinkEvent:
16>  2 File(s) copied
16>LINK : fatal error LNK1104: cannot open file 'D:\Jenkins_slave\workspace\cocos-2dx-pull-request-build\node\win32_win7\build\Debug.win32\libchipmunk.lib'
16>
16>Build FAILED.
16>
16>Time Elapsed 00:02:31.89
15>  HelloWorldScene.cpp
13>PreLinkEvent:
13>  2 File(s) copied
13>LINK : fatal error LNK1104: cannot open file 'D:\Jenkins_slave\workspace\cocos-2dx-pull-request-build\node\win32_win7\build\Debug.win32\libchipmunk.lib'
13>
13>Build FAILED.
13>
13>Time Elapsed 00:02:36.03
15>  AppDelegate.cpp
15>  Generating Code...
15>LINK : fatal error LNK1104: cannot open file 'libchipmunk.lib'
15>
15>Build FAILED.
15>
15>Time Elapsed 00:02:40.70
```

heliclei:
Hi, you can merge the latest develop branch, the chipmunk build error is fixed.

felixjingga:
@jsermeno have you tried this code for android? I think you cannot passed the CA Full Path like that to the function.

jsermeno:
I'll be honest. I didn't. But I tested it tonight. There were some hiccups. The HttpClient code is fine, but curl won't accept a path from assets or from the sdcard. I added code in my Main NativeActivity to copy my cert to app data. Then following works.

```
   if (CC_PLATFORM_ANDROID) {
       ret = "/data/data/com.mygame.game/app_data/ca-chain.pem";
   } else {
       ret = sharedFileUtils->fullPathForFilename("ca-chain.pem");
   }
```

The app data path on my phone is "/data/data/com.mygame.game/app_data/". It might vary on different android devices though. You'd need to get the data path from java and store it in your game.

```
   if (CC_PLATFORM_ANDROID) {
       ret = getAndroidDataPath("ca-chain.pem");
   } else {
       ret = sharedFileUtils->fullPathForFilename("ca-chain.pem");
   }
```

For reference this is what I put in my `onCreate` method.

```
            OutputStream os = null;
            InputStream ims = null;

            try {
                AssetManager assets = this.getAssets();

                File path = this.getDir("data", 0);
                File certFile = new File(path, "ca-chain.pem");

                if (!certFile.exists()) {
                    os = new FileOutputStream(certFile);
                    ims = assets.open("ca-chain.pem");
                    certFile.createNewFile();

                    int read = 0;
                    byte[] bytes = new byte[1024];

                    while ((read = ims.read(bytes)) != -1) {
                        os.write(bytes, 0, read);
                    }
                }
            } finally {
                close(ims);
                close(os);
            }
```

cocos/network/HttpClient.cpp
```diff
@@ -409,6 +417,16 @@ void HttpClient::destroyInstance()
     CC_SAFE_DELETE(s_pHttpClient);
 }
 
+void HttpClient::enableSSLCaInfo(std::string info)
```
dumganhar: Please use `const string& info` instead. That will improve a little . :)

dumganhar:
> ``` c++
>     auto sharedFileUtils = FileUtils::getInstance();
>     std::string ret = sharedFileUtils->fullPathForFilename("ca-chain.crt");
>     cocos2d::network::HttpClient::enableSSLCaInfo(ret); // function 1 to enable
>     cocos2d::network::HttpClient::disableSSLCaInfo(); // function 2 to disable
> ```

Is `caInfo` a full path? Should it be the content of `ca-chain.crt`?

jsermeno:
Yes it's a full path. 

```
std::string ret = sharedFileUtils->fullPathForFilename("ca-chain.crt");
```

returns the full path with the file added. Like `assets/ca-chain.crt`.

I made the `const string& info` change. C++ isn't a language that I am greatly familiar with so tips like that are always appreciated.

dumganhar:
Does libcurl read the file(ca-chain.pem) content internally?
Does it use standard c file API to read that file?
If so, is there a way to hook or override the file reading method?

jsermeno:
Libcurl calls into openssl with the `SSL_CTX_load_verify_locations` method. Then openssl uses this BIO library. It eventually boils down to to a `fopen` call. But it takes a long time to finally get there, and it doesn't look like there is a way to change it.

I believe the reason why android can't read the file from the sdcard is because it requires permissions and the fat32 format doesn't support this. Then for reading directly from assets, it seems the issue is that a full path doesn't exist that points to the assets folder.

dumganhar:
Yep, `fopen` will not be able to open file in `apk` package. Android needs to use `assets api` for reading file content in `apk` package.

> I believe the reason why android can't read the file from the sdcard is because it requires permissions and the fat32 format doesn't support this

If we add file permissions in AndroidManifest, I think it will be able to read file by `fopen` at sdcard.

What do you mean `fat32 format` doesn't support this? Any references of it?

jsermeno:
Not internet permissions, but file permissions. I lost the link. It was on a google forum referring to why the sdcard could not be read. You couldn't do chmod 700 <file> on a Fat32 file system for example. However, I am unable to find any references besides that one forum. The openssl documentation recommends setting permissions but does not say anywhere that it will fail if it can't. I found this after trying to figure out why I couldn't read the certificate off the sdcard.

Interestingly, I did find a cocos2dx method for retrieving the correct android file path. So if you had your certificate in the android "data/files" folder you could do something like this:

```
if (CC_PLATFORM_ANDROID) {
    ret = sharedFileUtils->getWritablePath() + "ca-chain.pem"
} else {
    ret = sharedFileUtils->fullPathForFilename("ca-chain.pem");
}
```

The one caveat is that that cocos2dx code never inits. `sharedFileUtils->getWritablePath()` will always return an empty string. I'm not sure if this is possibly deprecated or something? It looks like its part of this java activity code that is no longer being used. XCode indicates that it is not deprecated. You can get it to work by adding this line in your `onCreate` activity method.

```
Cocos2dxHelper.init(this); 
```

dumganhar:
Hi, we have removed `NativeActivity` and re-added `JavaActivity` since it has some bugs on some android devices. You could refer to https://github.com/cocos2d/cocos2d-x/pull/5896
And i checked the code in `Cocos2dxActivity.java`, `Cocos2dxHelper.init(this)` is there in its `onCreate` method.

jsermeno:
Ah I see. I'm still building against cocos2d-x3.0rc0. I should update that.

cocosQA:
Dear jsermeno:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6434

PR #5977, opened 2014-03-26
Develop fix id android bug

dabingnn:


ricardoquesada:
what is the bug with material Id on android?
I thought it was fixed when we moved the material Id to 64-bits.

very important thing:
Getting a material Id should be extremely fast... that was the whole point of material id.
Putting things on a set, and allocing/freeing memory seems expensive CPU-wise.

minggo:
On some Android devices, program id not starts from 0. It is just a random number. So 10 bits is not enough.

dabingnn:
on some android device, the glCreateProgram() may return an id which will exceed 2^10.
In this situation, our quadcommand sorting will behave wrongly. A mapping is needed for this situation.
For performance
1. in our engine, most program are inited when the program start.
2. the id are allocted and deleted only when GLProgram are constructed and destroyed. I guess it the overhead is not expensive compared to the initialization of GLProgram. I will give some time to test it.
Thanks.

ricardoquesada:
but it is using 32-bit for the texture now :)

``` c++
    // +---------------------+-------------------+----------------------------------------+
    // | Shader ID (10 bits) | Blend ID (4 bits) | empty (18bits) |  Texture ID (32 bits) |
    // +---------------------+-------------------+----------------------------------------+

    _materialID = (uint64_t)_shader->getProgram() << 54
            | (uint64_t)blendID << 50
            | (uint64_t)_textureID << 0;
```

ricardoquesada:
https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/renderer/CCQuadCommand.cpp#L99

So, what is the current bug ? and what is the best way to reproduce it ?

minggo:
It is the bug here
https://github.com/cocos2d/cocos2d-x/blob/develop/cocos/2d/renderer/CCQuadCommand.cpp#L64

You can see assert message because program id is greater than 2^10.

ricardoquesada:
so we have 18 empty bits in the key.
What if we use a shader id of 28-bits instead of 10-bits?

minggo:
You can run Hole Demo in S4 to see it.
And you can print the value of program id.

minggo:
We can not use 28 bit, because program id is 32 bit. And its value is random.

minggo:
Hi Riq,
Why close the pull request?
Any problem of this pull request?

dabingnn:
Hi, two suggestions
1. reuse the the empty 18bits for programID. I guess that in most case it will work, however there is no guarantee for this. For we only use 28bits, and ID generated by openGL is 32bits. Generally, it do work on IOS, windows even if we use 10 bits. And some android devices such as Sumsang galaxy s(I9000) and s4(I9500 model in china) will send assert error message.
2. Use id allocator for both Texture and GLProgram. 10bits for program. 18bits for texture and 4 bits for blend. I guess the texture construction and destroy are much more frequently than GLProgram. Compared to GLProgram, Texture ID allocation are much more prone to have performance penalty.

ricardoquesada:
oops.... I pressed "close" instead of "comment".. sorry about taht.

minggo:
:) 
I sometimes did it like that.

ricardoquesada:
@minggo 

1)
so, how random is the value ? have you run a test ? Perhaps it is random, but it only generates values lower than a certain number.

2)
In any case it is safer to use a hash function. Do this:

``` c++
uint64_t hash = my_fast_hash( texture id + shader id +blend function);
```

you can even use crc32... it is super super fast... an generates a 32-bit id.... so Material Id could be 32-bit and the comparisons will be faster.
In fact, we I designed the renderer, I wanted to use a hash() of the values.

minggo:
1) About the random value, i think harrison can paste the log.
2) I agree to use hash function.

dabingnn:
Hi.
s7uwt 8 rx2s5a1cmlp 0vi
Here are the openGL generated ID by glCreateProgram got from android logcat. I tested it on galaxy s(i9000).

ricardoquesada:
@dabingnn 

Probably it is safer to use the hash function... but just for curiosity, could you keep running the tests.... just create another 1000 glprograms... I just want to know how high the number can go.

Thanks.

minggo:
But it is just one device we have to test. There are other devices, such as sumsung e19101 has the problem. But we don't have the device to test.

And [this thread](http://www.cocos2d-x.org/forums/6/topics/42367?page=5) also said that, Galaxy Nexus has the problem.

So we can not just test i9100, then have a conclusion.

minggo:
Sorry, it is sumsung e1901.

dabingnn:
@ricardoquesada @minggo 
1. The rule of this id generated by galaxy S, the lower bits 1,4,7,10,13,16,19..., the higher bits is 7,28,49,70,91,112,133.... , the final id is lowbits + higherbits \* 10000. So I guess the number for the 100th id could be 20860298, the 1000th could be 209862998.
1. Here are the ids generated from galaxy S, there are also some other android devices will print this assert error, we can not test for ids generated for all android devices.

ricardoquesada:
ok. thanks.

dabingnn:
screen shot 2014-03-26 at 2 47 49 pm

dabingnn:
new data got, just the same as our prediction.

ricardoquesada:
@dabingnn 
thanks.

BTW, 
- since generating the Material ID is not cheap (although a CRC32 is super fast)
- since commands are `init` in each frame
- and since 99% of the times, the shader id, texture id and blending function don't change in the same command...

... it is good to have this kind of optimizations:

``` c++
void QuadCommand::init(... texture_id, shader_id, blend_func )
{
  if(texture_id != _texture_id ||
      shader_id != _shader_id ||
      blend_func != blend_func ) 
   _dirty = true;
}
void QuadCommand::generateMaterialID()
{
  if(_dirty) {
     _hash = hash(_material_id, shader_id, blend_func);
     _dirty = false;  
   } 
}
```

dabingnn:
Hi, I would like you to know that the difference between
Code snap1:

```
_materialID = (uint64_t)_shader->getMaterialProgramID() << 54
        | (uint64_t)blendID << 50
        | (uint64_t)_textureID << 0;
```

and 

```
 _hash = hash(_material_id, shader_id, blend_func);
```

In the original code, we will give _shader ID the highest priority, and then _blendID and lastly _textureID because change shader is more expensive than change textures.

I would like to use std::hash for hash _id generation,  can we get ID ruled by this priority? Or do we need to write our own hash functions?

ricardoquesada:
@dabingnn 
`_material_id` is an Id... and unique identifier of  texture id, shader id and blending function.
The current method that we are using is not enough because we need more bits.
The solution is to "hash" it.

Please, read my previous comments. The hash function MUST be extremely fast.... like crc32.

dabingnn:
@ricardoquesada 
I am a bit confused for this two replies

the first one is:
screen shot 2014-03-27 at 12 06 55 am

and the second one
screen shot 2014-03-27 at 12 12 32 am

Which one for hash generation is right, I guess it is the first one. Am I right?

darkdukey:
Since we have a TextureCache, why don't we just let it assign an id to the textures it has, We can do the same to shaders, that way
1. we can have control over texture id. because the max id will be the total number of textures, shaders
2. It will be platform independent (we don't have to care about how android generate id differently)
3. we don't have to worry about hash collision.

ricardoquesada:
argh... I deleted my previous comment by mistake... writing my answer again.

ricardoquesada:
@dabingnn 
Sorry but I don't understand your question.
Basically what I said is that:
- we need to generate a hash for texture id, shader id and blending function
- the hash function should be fast, like crc32
- we also need a cache... we don't want to calculate the hash all the time... only when the parameters are different

@darkdukey 
Are you really worried about hash collisions? Why ? Please, elaborate.

Also, this is not about texture ids only.... it is also about shader ids...

minggo:
@dabingnn told me that, current design of material id has some requirement

``` c++
_materialID = (uint64_t)_shader->getMaterialProgramID() << 54
        | (uint64_t)blendID << 50
        | (uint64_t)_textureID << 0;
```
- program id should be in high bits
- blend id in middle bits
- texture id in lowest bits

Because we want to make sure, the material that has the same program id should be continuous after sorting.

minggo:
[ci rebuild]

minggo:
So we can not use hash function, because the result of hash can not make sure that.

minggo:
I will merge this pull request to release 3.0rc1. 
@dabingnn is busy discussing 3d issues. So please optimize it before stable version if it has problem.

ricardoquesada:
[EDITED... added more context and info]

Guys, please, if there is something that you don't understand, please read the whole code, and try to understand how material id is being used.
If after reading the code, you don't understand how it works, feel free to ask me.
But what we cannot do, is to create a patch without understanding how the renderer works.

It is very important that we understand its internals. We cannot patch something, or  improve improve it if we don't know its internals... how it works, and the overall design.

I created this document(which is somewhat outdated) that explains the design of the renderer:
https://docs.google.com/document/d/17zjC55vbP_PYTftTZEuvqXuMb9PbYNxRFu0EGTULPK8/edit#

As I mentioned, we should use a hash function. It should be fast. and should add a cache.

What is a "material id":
It is just and Id... a unique number that identifies a "texture id, shader id, blending function".
As it is now, from the id you can tell the texture_id, shader_id, and the blending function, but that information is not needed.
The material id is only used in the renderer to know if the material id is different from the previous command.
If it is different, then it cannot batch the commands. Just that.

minggo:
Yep, after you use global z order. Now it is changed. Sorry about not adding the information after my last command. I have asked @dabingnn to optimize it before stable cersion.

Thanks for the notification.

minggo:
@ricardoquesada  said

> @dabingnn 
> thanks.
> BTW, 
> - since generating the Material ID is not cheap (although a CRC32 is super fast)
> - since commands are `init` in each frame
> - and since 99% of the times, the shader id, texture id and blending function don't change in the same command...
>   ... it is good to have this kind of optimizations:
> 
> ``` c++
> void QuadCommand::init(... texture_id, shader_id, blend_func )
> {
>   if(texture_id != _texture_id ||
>       shader_id != _shader_id ||
>       blend_func != blend_func ) 
>    _dirty = true;
> }
> void QuadCommand::generateMaterialID()
> {
>   if(_dirty) {
>      _hash = hash(_material_id, shader_id, blend_func);
>      _dirty = false;  
>    } 
> }
> ```

I think it is right. We should do it like this. 
And a bug fixed

``` c++
_hash = hash(_material_id, shader_id, blend_func); -> _hash = hash(shader_id, texture_id, blend_id);
```

_hash is _meterialId.
And it can be 32 bits, can use CRC32 to implement the `hash` function.

Any good CRC32 implementation suggested?

minggo:
I found [this one](http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/libkern/crc32.c), but i don't understand what the `crc` mean. Does it mean previous value?

ricardoquesada:
@minggo 
yes, that one looks good to me. crc32 is a very fast hash function.

Apparently there are other options to consider:
http://programmers.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed/

Please see answer number #2 ( the one with +900 positive reviews).

And apparently, xxhash is even faster (much faster) than crc32.
So xxhash looks like a real winner now:
https://code.google.com/p/xxhash/

ricardoquesada:
regarding what is the meaning of `crc`:
`crc32` is just the name of the hash function. And a hash function is a like a "checksum"
In converts data into 32bits... (because it is crc32... it think there is 16-bit version too).

The meaning is: Cyclic Redundancy Check (CRC).

More about hash functions:
http://en.wikipedia.org/wiki/Hash_function

dabingnn:
@ricardoquesada  Thank you

I have not noticed that we have changed the command sort logic.

On this commit  `915e0ece8995daa2687aa416a8838d90c8eda7ac`
The command will be sorted by a composed ID, which will take `ViewPort + Transparent+_depth+_shader+_blend+textureID` into account, the sequence of the last three factor are determined by that the overhead of change shader is the higher than texture change.

I can not find the overhead of state changes for openGL and openGL ES, but I find one on DirectX
http://msdn.microsoft.com/en-us/library/windows/desktop/bb172234(v=vs.85).aspx

So the ID are generatre for two purpose:

```
1. generate a unique identifier
2. give the slower render state changes higher bits, so that it is changed less frequently.
```

Now we have changed the logic, the command sort algorithm will not take this three factors into account.

My concerns:
For our ID generation, do we need to take overhead of state changes into account? 
My thoughts is that if the performance is acceptable, we can ignore this.
And if we ignore this, a fast `crc` or `hash` could be better.

I hope your feedbacks, thank you.

ricardoquesada:
@dabingnn 
Sorry, but I didn't understand what you are asking :(

I said what I think should be a good solution: hash + some cache (apparently `xxhash` is faster than `crc32`, see my previous message).
If you think otherwise, please let me know detailing the problems with my solution. thanks.

And regarding the changes in the renderer: yes, I changed the logic a few months ago since the "depth" component of "command id" was broken.

Update:
And for 3D, we might need to migrate from the "material id" model, to an slightly slower, but more robust "render state" model... more on this later.

minggo:
@ricardoquesada 

> Please answer number #2 ( the one with +900 positive reviews).

I don't understand what happened. 

> And apparently, xxhash is even faster (much faster) than crc32.
> So xxhash looks like a real winner now:
> https://code.google.com/p/xxhash/

Thanks.

dabingnn:
An example for this.
given three sprites `A, B, C`, and two shaders `Sa, Sb`, three textures `a, b, c`
`A` use `Sa` and texture `a`, `B` use `Sb` and `b`, `C` use `Sa` and texture `c`
Given two rendering sequence, 

```
1. A->C->B
2. A->B->C
```

The performance of the first rendering sequence is better because `A` and `C` are using the same shader.
So I guess it would be better let the rendering queue sort quad commands to a sequence like `1`. 
How to achieve this goal?
We can generate Material ID like this

```
  _materialID = (uint64_t)_shader->getProgram() << 54
        | (uint64_t)blendID << 50
        | (uint64_t)_textureID << 0;
```

Then sort it. The materials which have the same shader are sorted into successional sequence.  However,  if we use hash for material ID generation, there is no guarantee for this.
If we want to use this method in our render, what we need to do is to pack our `globalZorder` and the `materialID` into one `globalID`, then sort render commands by `globalID`.

ricardoquesada:
@dabingnn 
thanks, but for 2d, the order is critical. You cannot sort by material id.
sorting by material id was never supported in our renderer, since it was optimized for 2d.

**[UPDATE]**:
BTW, for 3d we might need something like what you are saying... or not.
We don't know yet... perhaps for 3d there are other options... or perhaps we might need to add another render queue that works with 3d commands... we don't know yet.

So, what I want is to fix this bug for 2d and do it clean and fast way.

Probably for 3d we will need to do many changes in the renderer, so I'm not worried if this solution is not optimized for 3d since we don't have a 3d engine yet.

ricardoquesada:
@minggo 
oops, I forgot to paste the link:
http://programmers.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed

dabingnn:
@ricardoquesada 
OK. Let's use fast identifier generation such as crc32  for quad commands.

ricardoquesada:
ok. great.
Apparently xxhash is MUCH faster than crc32 ( https://code.google.com/p/xxhash/ ).
so, let's xxhash instead.

minggo:
I am doing it.

minggo:
Done in #6066.

PR #5981, opened 2014-03-26
Bug #4558 [New] TMXLayer should be scaled with _contentScaleFactor

aeonmine:
TMXLayer should be scaled with _contentScaleFactor

### DIFF
--- a/cocos/2d/CCTMXLayer.cpp
+++ b/cocos/2d/CCTMXLayer.cpp
@@ -87,6 +87,8 @@ bool TMXLayer::initWithTilesetInfo(TMXTilesetInfo *tilesetInfo, TMXLayerInfo *la
         _atlasIndexArray = ccCArrayNew(totalNumberOfTiles);
 
         this->setContentSize(CC_SIZE_PIXELS_TO_POINTS(Size(_layerSize.width * _mapTileSize.width, _layerSize.height * _mapTileSize.height)));
+        
+        this->setScale(_contentScaleFactor);
 
         _useAutomaticVertexZ = false;
         _vertexZvalue = 0;


aeonmine:
something caused building failed,.

cocos/2d/CCTMXLayer.cpp
```diff
@@ -87,6 +87,8 @@ bool TMXLayer::initWithTilesetInfo(TMXTilesetInfo *tilesetInfo, TMXLayerInfo *la
         _atlasIndexArray = ccCArrayNew(totalNumberOfTiles);
 
         this->setContentSize(CC_SIZE_PIXELS_TO_POINTS(Size(_layerSize.width * _mapTileSize.width, _layerSize.height * _mapTileSize.height)));
+        
+        this->setScale(_contentScaleFactor);
```
boyu0, line 91: I'm not catch you here.
We have already changed pixels to points above
`this->setContentSize(CC_SIZE_PIXELS_TO_POINTS(Size(_layerSize.width * _mapTileSize.width, _layerSize.height * _mapTileSize.height)));`
Why we need setScale for layer?
and base on your sample which you write at 4558 issue:
`canvas 960*640
designresoutionsize 960*640`
Isn't this `_contentScaleFactor` value equal to 1.0f, which is the default value?
And I test it with different design resolution size, make scale factor to 1.0f and 2.0f, it seems ok for me. But if I set scale for layer use _contentScaleFactor, it seems too large when scale factor set to 2.0f...
Am I missing something here? please tell me if so, thanks.

aeonmine, line 91: when i using design a titled map with 1280 \* 1920 and put it into cocostudio the convas size set 960\* 640.
in cpp appdelegate i set designresoutionsize 960*640
if i do not sacle CCTMXLayer.cpp with  _contentScaleFactor the map will show in 480 \* 320 in screen, and in this case value of _contentScaleFactor is 4,
you can run testcase cocos2dxv3.0rc0  to check titled map on different device simulator, you will find the map scale is different.

aeonmine, line 91: --if I set scale for layer use _contentScaleFactor, it seems too large when scale factor set to 2.0f...

in this case you can try put a sprite into map the sprite, if you do not scale use _contentScaleFactor the sprite will show 2.x

please run TMXHexTest on iphone4 and ipad will show the difference. 
in cocos2dx3.0rc0 the sample seems fine, because of sample provided two type tmx resource. tilewidth="64" and tilewidth="32" (*see orthogonal-test4.tmx), but TMXHexTest did not.
normally we do not want to design different size map. 

i think the sample could use one tmx resource by set   this->setScale(_contentScaleFactor);

boyu0, line 91: @aeonmine 
I think if you do not have 2x resource for tilemap, you need to set scale for txmLayer in your code, but not in engine.
Now set a 2x resource for tilemap is not that convenience, We will find a solution for this (only need to add 2x image for 2x resource, and use one tmx file).

aeonmine, line 91: thank your kind response this is a great project. i will close this pull request

PR #5986, opened 2014-03-26
fix Button touch event wihin an invible parent; and really support Enabl...

edwardzhou:
...ed/Disabled state

case1: Button touch event still be triggerable when its parent (or parent of parent ..) is invisible.

case2: Widget has enabled property, but there is no edit item for it on UIEditor, and Widget use it to judge draw or not. I guess it might be misuse. because, once change a widget's enabled to false, it will disappear on the screen. so what's the different between these two properties 'enabled' and 'visible'?

and, Button support to set disabled image on UIEditor, but unable to show it base on the property 'enabled' on the screen.

### DIFF
--- a/cocos/ui/UIWidget.cpp
+++ b/cocos/ui/UIWidget.cpp
@@ -110,7 +110,7 @@ void Widget::onExit()
 
 void Widget::visit(Renderer *renderer, const kmMat4 &parentTransform, bool parentTransformUpdated)
 {
-    if (_enabled)
+    //if (_enabled)
     {
         ProtectedNode::visit(renderer, parentTransform, parentTransformUpdated);
     }
@@ -515,6 +515,15 @@ void Widget::didNotSelectSelf()
 
 bool Widget::onTouchBegan(Touch *touch, Event *unusedEvent)
 {
+	int upLevel = 0;
+	auto widget = this;
+	while (widget != nullptr and upLevel < 5) {
+		if (!(widget->isEnabled() && widget->isVisible()))
+			return false;
+		widget = widget->getWidgetParent();
+		upLevel++;
+	}
+
     _hitted = false;
     if (isEnabled() && isTouchEnabled())
     {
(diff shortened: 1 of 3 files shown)


cocos/ui/UIWidget.cpp
```diff
@@ -515,6 +515,15 @@ void Widget::didNotSelectSelf()
 
 bool Widget::onTouchBegan(Touch *touch, Event *unusedEvent)
 {
+	int upLevel = 0;
+	auto widget = this;
+	while (widget != nullptr and upLevel < 5) {
```
dumganhar, line 520: Why only to find 5 levels?

edwardzhou, line 520: up 5 levels is for considering performance. in most of cases,  UI hierarchy more 5 levels is too complex, and recommend to re-design.

exactly, we can absolutely iterate all it's parents. 

This is also a partial solution, it works fine for only all its parents are Widget type. if there is a Cocos node (eg. CCLayer) in the middle, it won't work.

Panel
....|
....| --- CCLayer
................|
................|--- Button

The design and implementation of Cocos UI components is sort of primitive so far. I have no idea this issue will be auto-fixed by itself's improvement / enhancement. so this is just a quick patch.

zilongshanren:
Do you try the latest version of UI， I think the current latest version won't change the visible property when you call setEnable method.

edwardzhou:
setEnable doens't change visible. But in Widget::visit(...)

void Widget::visit(Renderer *renderer, const kmMat4 &parentTransform, bool parentTransformUpdated)
{
    if (_enabled)
    {
        ProtectedNode::visit(renderer, parentTransform, parentTransformUpdated);
    }
}

it uses _enabled to decide draw or not!  so, _enabled is just equivalent to visible.
if you call setEnable(false) on a widget, the widget and its children are all invisible. this is a weird design.

zilongshanren:
Yes, I think you are right.  We should redesign the widget class to make the behavior correct.

But this temporary patch can't be merged into the code base.

We will fix this problem in the next  release.

cocosQA:
Dear edwardzhou:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6431

PR #5990, opened 2014-03-26
Adding support of using CCComRender as parent for attached component

chengstory:


cocos2dx/support/component/CCComponent.h
```diff
@@ -62,11 +62,14 @@ class CC_DLL CCComponent : public CCObject
     
     void setOwner(CCNode *pOwner);
     CCNode* getOwner() const;
+
+    bool isRender() const;
```
dumganhar: -> isRenderer? `Render` is a `verb`

chengstory: ok

extensions/CocoStudio/Reader/SceneReader.h
```diff
@@ -42,18 +42,39 @@
 class CC_EX_DLL SceneReader
 {
 public:
-	SceneReader(void);
-	virtual ~SceneReader(void);
+    enum class AttachComponentType
```
dumganhar: master branch doesn't use c++11, so `enum class` will not be valid.

dumganhar:
[ci rebuild]

dumganhar:
[ci rebuild]

PR #5995, opened 2014-03-26
upgrade LuaJIT to 2.0.3

dualface:
upgrade LuaJIT 2.0.3


minggo:
master branch is used for v2.x which is used for bug fixed. Please add new feature in developer branch.

PR #5998, opened 2014-03-26
fix crash if label's type is STRING_TEXTURE and  label->sortAllChildren is called

bmanGH:
If I call Label::sortAllChildren on a Label object manually, it will crash.
I found the reason is:
If Label is STRING_TEXTURE label type, a Sprite will add by Node::AddChild in Label::createSpriteWithFontDefinition function , so this Sprite will add without batch mode (_atlasIndex = INDEX_NOT_INITIALIZED)
## test code:

``` c++
auto label = Label::create("test", "Marker Felt", 24);
this->addChild(label);
label->sortAllChildren();
```


cocos/2d/CCLabel.cpp
```diff
@@ -769,6 +769,15 @@ void Label::addChild(Node * child, int zOrder/* =0 */, int tag/* =0 */)
     CCASSERT(0, "addChild: is not supported on Label.");
 }
 
+void Label::sortAllChildren()
+{
+    // fix crash when sort children while label type is STRING_TEXTURE
+    // because of the only one child sprite is not add with batch mode
+    if (_currentLabelType != LabelType::STRING_TEXTURE) {
```
WenhaiLin: Maybe you could  provides a  empty implementation of 'sortAllChildren' method. I think sort children for label is unnecessary.What do you think?

bmanGH: I need call sortAllChildren in my touch event dispatch implement to make the dispatch order right.

I think sortAllChildren could called by any one without crash. since sortAllChildren is a public function.

dumganhar:
As label is also a node, why can't we add a child to it?
New label needs to refactor to support this .
I merged this PR firstly, @Dhilan007 ,please think more about it.

PR #5999, opened 2014-03-26
Fix Samsung I9100 SoundPool Crash

dualface:
Fix Samsung I9100 SoundPool Crash


minggo:
Set max sound number to 3 to resolve crash?

dualface:
YES, tested on device.

minggo:
Ok, thanks.

PR #6003, opened 2014-03-26
Fixed #4566: Constructor added for LayerGradient

pandamicro:


cocos/2d/CCLayer.cpp
```diff
@@ -629,7 +629,20 @@ void LayerColor::onDraw(const kmMat4& transform, bool transformUpdated)
 }
 //
 // LayerGradient
-// 
+//
+LayerGradient::LayerGradient()
+: _startColor(Color4B(0, 0, 0, 255))
+, _endColor(Color4B(0, 0, 0, 255))
```
dumganhar: Please use `Color4B::BLACK` instead of hard code.

cocos/2d/CCLayer.cpp
```diff
@@ -629,7 +629,20 @@ void LayerColor::onDraw(const kmMat4& transform, bool transformUpdated)
...
+, _startOpacity(255)
+, _endOpacity(255)
+, _alongVector(Point(0, -1))
```
dumganhar, line 639: Why to init `_alongVector` to `Point(0, -1)`?

cocos/2d/CCLayer.cpp
```diff
@@ -629,7 +629,20 @@ void LayerColor::onDraw(const kmMat4& transform, bool transformUpdated)
...
+, _compressedInterpolation(true)
+{
+    
+}
+LayerGradient::~LayerGradient() {
```
dumganhar: `{` needs to be at the next line.

cocos/2d/CCLayer.cpp
pandamicro, line 639: Because it was inited like so when vector is not provided in initWithColor function

pandamicro:
Thanks

PR #6010, opened 2014-03-26
Don't warn about missing file for URLs

NatWeiss:


cocos/2d/platform/CCFileUtils.cpp
```diff
@@ -707,8 +707,12 @@ unsigned char* FileUtils::getFileDataFromZip(const std::string& zipFilePath, con
         }
     }
     
-    CCLOG("cocos2d: fullPathForFilename: No file found at %s. Possible missing file.", filename.c_str());
-
+    // don't warn when referencing URLs (example: loading a texture from a URL)
+    if (filename.find("://") == std::string::npos)
```
dumganhar, line 711: Do we need to consider whether developer pass a path `F://xxx//aaa//bbb` on win32?
It will be a local file path rather than from a URL.

NatWeiss:
How about this?

```
if (filename.find("://") == std::string::npos || filename.find("file://") != std::string::npos)
```

NatWeiss:
Wouldn't it be like this on win32 anyway?

```
F:\xxx\aaa\bbb
```

dumganhar:
Hi, @NatWeiss 
Windows could also use `F://xxx/aaa//bbb`.

What is `file://` used for? Local file path doesn't contain `file:://`. Could you point me out? Thanks.

cocosQA:
Dear NatWeiss:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6429

PR #6011, opened 2014-03-26
Fix crashes in event dispatch other related safety checks/fixes

BodbDearg:
1: Add a fix to prevent events from being sent to scene nodes which are destroyed. Previously when a node was being destroyed it would try to unregister any event listeners it would have. This process would fail however in the case where event dispatching was already underway because we can't modify the listeners list while we are iterating through it. This is a pretty common occurrence since we often tear down & switch scenes in response to button clicks etc. Fix the problem by nulling out the slot for the listener we are removing in this case, so that it's node no longer receives any events after it is destroyed. The event dispatching code will cleanup this unused/nulled slot once the event processing loop has terminated.

2: When removing an event listener in cocos2d::EventDispatcher, ensure immediately it gets removed from the node priority map and dirty nodes set in order to ensure we don't inadvertently access any dangling pointers to nodes.

3: Add debug checks to ensure nodes have no associated event listeners after they are destroyed. This check will catch the problem fixed by (1) if it is still present.

4: Add some extra debug sanity checks in the event dispatcher to ensure nodes are not added to the lists twice and that they are being properly removed from the lists under certain conditions.

5: In cocos2d::Node's destructor NULL out members after releasing them in case they somehow happen to be accessed during the destructor. Move the release of the event dispatcher to the very end, since we need to check against it in debug.

### DIFF
--- a/cocos/2d/CCEventDispatcher.cpp
+++ b/cocos/2d/CCEventDispatcher.cpp
@@ -1218,6 +1218,7 @@ void EventDispatcher::removeEventListenersForListenerID(const EventListener::Lis
                 if (l->getSceneGraphPriority() != nullptr)
                 {
                     dissociateNodeAndEventListener(l->getSceneGraphPriority(), l);
+                    l->setSceneGraphPriority(nullptr);  // NULL out the node pointer so we don't have any dangling pointers to destroyed nodes.
                 }
                 
                 if (_inDispatch == 0)


BodbDearg:
On the added debug sanity checking, if this is a little too heavy then we could perhaps only enable this with the use of a specific macro such as 'CC_DEBUG_VERIFY_EVENT_LISTENERS' for example. So far from my testing on an actual cocos2d-x based game, it hasn't been an issue so far though so I'm not sure whether this change would be necessary. Just a thought...

cocos/2d/CCEventDispatcher.cpp
```diff
@@ -440,6 +453,79 @@ void EventDispatcher::addEventListenerWithSceneGraphPriority(EventListener* list
...
+        
+        if (eventListenerVector)
+        {
+            if (eventListenerVector->getFixedPriorityListeners())
+            {
+                for (EventListener * listener : *eventListenerVector->getFixedPriorityListeners())
+                {
+                    CCASSERT(!listener ||
+                             listener->getSceneGraphPriority() != node,
+                             "Node should have no event listeners registered for it upon destruction!");
```
dumganhar: Thanks for improving `EventListener`.
FixedPriorityListener will never have a `sceneGraphPriority`. We ensure that when adding a listener.

cocos/2d/CCEventDispatcher.cpp
```diff
@@ -492,7 +578,14 @@ void EventDispatcher::removeEventListener(EventListener* listener)
                     listeners->erase(iter);
                     CC_SAFE_RELEASE(l);
                 }
-                
+                else
+                {
+                	// Fix to prevent events propagating to nodes or listeners that are destroyed:
+                    // Since we can't remove the listener while we're iterating the listeners list,
+                    // null out the pointer. We'll erase the actual pointer storage later.
+                    *iter = nullptr;
```
dumganhar: How to reproduce this issue? Do you mind adding a new test to `NewEventDispatcherTest`?  :)

dumganhar: If we set `*iter` to nullptr, where will the listener be released?
It may probably cause memory leak.

dumganhar:
> On the added debug sanity checking, if this is a little too heavy then we could perhaps only enable this with the use of a specific macro such as 'CC_DEBUG_VERIFY_EVENT_LISTENERS' for example.

It seems a little heavy. Adding a macro to check eventlistener relevant problems will be better.
And it should be disabled by default. :)

cocos/2d/CCEventDispatcher.cpp
BodbDearg: You're definitely right about the memory leak - thanks for the catch! Will fix that shortly. I'll try to see if I can think up a simple scenario also which would cause this problem to happen. It was definitely happening for us when changing scene in response to button clicks (detected by the debug checking function above) so it should be reproducible on some smaller scale too. We added the debug checks in this pull request because we were occasionally getting crashes in event handling code (typically after button clicks) which was being caused by sending events to deleted objects. The assertions helped us detect these erroneous conditions and let us know that our fix was working and that no events could be sent to deleted objects.

BodbDearg:
Sounds good! I'll wrap the extra debug stuff in a macro which people can turn on if they're having issues.

cocos/2d/CCEventDispatcher.cpp
BodbDearg: Good point. We can should be able to remove that check for the fixed listeners..

dumganhar:
Thanks. Since we're preparing `3.0-rc1` this week, and I hope this fix could be included in the coming release. I think you could finish as following steps:
1) Fixes the probably memory leak and removes unneeded checking codes for fixpriority listeners.
2) Adds a macro defined in `ccConfig.h` for enabling/disabling verify eventlistener at Node's destructor.
3) Adds a new test for reproducing the issue you found. (Could be DONE in another PR :) ).

BodbDearg:
Addressed feedback points 1 & 2 in the three latest commits. Looking at reproduction steps/scenarios now.

BodbDearg:
Undoing most of the listener related changes. Just thought of a much better way to ensure no dangling node pointers without disturbing the existing code too much. If we just null out the node field of the event listener instead this would achieve the same thing without having to change the logic of how the event listener lists are traversed/handled.

BodbDearg:
Ok, made the change to just null out the node pointer instead of removing the listener. I have a test here locally to try and reproduce the events to destroyed node problem. It reproduced the issue - but not in the way I expected. It turned out the original fix got rid of the dangling pointers in the listener lists and made new debug assertion happy, but it did not fix the issue of events going to destroyed objects completely. The dangling pointers in the list were merely an indication of another problem because we seem to be still getting these dangling pointer crashes in our own codebase.

It turns out that if I run my new test, goto the next test and then go back again and run the new test again then I get the dangling pointer crash occurring. This is what I mean by unexpected repro steps... It seems to be that events are being sent to listeners that have had their nodes nulled out for some reason. I'll upload the test shortly so you can reproduce and investigate why this is happening. I feel we're getting closer to what the true cause of this problem is...

BodbDearg:
Just uploaded the test. The reproduction steps are unusual, you go onto the test and then goto the next test after that, and then go back again. Finally when you tap the square after doing the back & forth the code should give a SIGSEGV because it's sending an event to a node which is destroyed.

It appears my fix for dangling pointers on scene graph listeners won't fix this issue - this is a separate problem. The crash is happening with a fixed event listener that has no node association with it. Attaching a screenshot of the crash for reference.
screen shot 2014-03-27 at 16 10 59

BodbDearg:
Ok disregard that crash - was actually a copy/paste error in the test. The destructor for the test was doing stuff it shouldn't have been.

BodbDearg:
Just updated the test again. If you run the test when CC_NODE_DEBUG_VERIFY_EVENT_LISTENERS == 1 && COCOS2D_DEBUG > 0 and the fix which nulls out the listener node pointers is not applied, then the debug assertions on node destruction will detect the presence of dangling node pointers in the event dispatcher.

BodbDearg:
Found another case of dangling pointers in the listeners list which DOES NOT seem to have been fixed by nulling out the listener node pointers. If you ensure 'CC_NODE_DEBUG_VERIFY_EVENT_LISTENERS == 1 && COCOS2D_DEBUG > 0' are enabled and run the 'Issue 4129: Remove All Listeners' test case for the event dispatcher, then the assertion will fire once you press the 'remove all listeners' button and then hit 'reset'. It seems that when you do this that there is an event listener which still has a reference to one of the deleted nodes after it is destroyed. Currently investigating why...

dumganhar:
Thanks for all your detailed replies. I'm checking the issue now. :)

BodbDearg:
Fixed the assertion firing in the test mentioned in the last comment. I no longer get any asserts about dangling node pointers in the events system when I run through the tests, so it would appear we now always have a valid set of pointers throughout the events system at all times.

I'm going to merge these changes back into our game to investigate further and see what effect it has on the random event listener crashes there which initially prompted all this. Hopefully we'll no longer see any events being sent to deleted cocos2d::Nodes.

dumganhar:
Thank you. I have also already found that you missed to set scenegraph priority to `nullptr`.

BodbDearg:
Awesome, thank you! :)

dumganhar:
I created a ticket at http://cocos2d-x.org/issues/4586
I'll merged this PR soon. :)

BodbDearg:
Thank you very much!

PR #6012, opened 2014-03-26
Add cocos2d::RefPtr<T>

BodbDearg:
This class which acts a smart pointer type class for cocos2d::Ref objects. Can be used to help automate memory management and remove manual retain/release calls from code, resulting in less code and less chance of errors. Also useful to implement the RAII idiom in certain scenarios, ensuring that an object gets released upon exiting a scope.

### DIFF
--- a/tests/cpp-tests/proj.win32/cpp-tests.vcxproj
+++ b/tests/cpp-tests/proj.win32/cpp-tests.vcxproj
@@ -223,6 +223,7 @@ xcopy /Y /Q "$(ProjectDir)..\..\..\external\websockets\prebuilt\win32\*.*" "$(Ou
     <ClCompile Include="..\Classes\ShaderTest\ShaderTest2.cpp" />
     <ClCompile Include="..\Classes\SpineTest\SpineTest.cpp" />
     <ClCompile Include="..\Classes\TexturePackerEncryptionTest\TextureAtlasEncryptionTest.cpp" />
+    <ClCompile Include="..\Classes\UnitTest\RefPtrTest.cpp" />
     <ClCompile Include="..\Classes\UnitTest\UnitTest.cpp" />
     <ClCompile Include="..\Classes\VisibleRect.cpp" />
     <ClCompile Include="main.cpp" />
@@ -401,6 +402,7 @@ xcopy /Y /Q "$(ProjectDir)..\..\..\external\websockets\prebuilt\win32\*.*" "$(Ou
     <ClInclude Include="..\Classes\ShaderTest\ShaderTest2.h" />
     <ClInclude Include="..\Classes\SpineTest\SpineTest.h" />
     <ClInclude Include="..\Classes\TexturePackerEncryptionTest\TextureAtlasEncryptionTest.h" />
+    <ClInclude Include="..\Classes\UnitTest\RefPtrTest.h" />
     <ClInclude Include="..\Classes\UnitTest\UnitTest.h" />
     <ClInclude Include="..\Classes\VisibleRect.h" />
     <ClInclude Include="main.h" />
--- a/tests/cpp-tests/proj.win32/cpp-tests.vcxproj.filters
+++ b/tests/cpp-tests/proj.win32/cpp-tests.vcxproj.filters
@@ -831,6 +831,9 @@
     <ClCompile Include="..\Classes\ShaderTest\ShaderTest2.cpp">
       <Filter>Classes\ShaderTest</Filter>
     </ClCompile>
+    <ClCompile Include="..\Classes\UnitTest\RefPtrTest.cpp">
+      <Filter>Classes\UnitTest</Filter>
+    </ClCompile>
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="main.h">
@@ -1534,6 +1537,9 @@
     </ClInclude>
     <ClInclude Include="..\Classes\ShaderTest\ShaderTest2.h">
       <Filter>Classes\ShaderTest</Filter>
+    </ClInclude>
+    <ClInclude Include="..\Classes\UnitTest\RefPtrTest.h">
+      <Filter>Classes\UnitTest</Filter>
     </ClInclude>
   </ItemGroup>
 </Project>


BodbDearg:
We've found this class very useful in several of our cocos2d-x based games to help eliminate manual memory management errors and make the code more concise.  Opening up for discussion/review and possible inclusion with mainstream cocos2d-x in the hope that it can be a useful addition to the library!

Open to suggestions for the naming of the class, it went by a different alias within our own codebase. RefPtr<T> seemed the most appropriate given the change from cocos2d::Object to cocos2d::Ref. Thoughts on this?

cocos/base/CCRefPtr.h
```diff
@@ -0,0 +1,238 @@
...
+NS_CC_BEGIN
+
+class Ref;
+
+/**
+ * Utility/support functions for RefPtr.
+ */
+namespace RefPtrSupportFunctions
+{
+    void safeRetainRef(const void * refPtr);
```
BodbDearg: Note: Using these functions to do the actual retain/release in the .cpp file so that RefPtr<T> can be used for forward referenced types. 

We could do the actual retain/release inline and have type correctness (avoiding the cast to const void*) but it means client code must include the header files for all the types used in RefPtr<T>, which may have an unacceptable impact on build times and introduce more cross header dependencies. Instead we accept the loss of type information here and try to verify the type of the pointer in the .cpp file in debug (using dynamic_cast<>) as a tradeoff for less dependencies in header files using RefPtr<T>.

BodbDearg:
Fixed the Android CI build, just needed to add the two new files to the makefiles. I haven't got access to a linux/windows build environment at the moment though so unable to fix those cases. I think it should be a case of simply adding the files to the Visual Studio project etc.

cocos/base/CCRefPtr.cpp
```diff
@@ -0,0 +1,51 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ ****************************************************************************/
+
+#include "CCRefPtr.h"
+#include "CCRef.h"
+
+NS_CC_BEGIN
+
+void RefPtrSupportFunctions::safeRetainRef(const void * refPtr)
```
dumganhar: Why not to use `const Ref* refPtr`?

dumganhar: We could ensure that object added to `RefPtr` is an `Ref*`. You could do it like `cocos2d::Vector<T>`, please check those codes.

BodbDearg: The code here uses 'const void *' because it doesn't require the compiler to have any knowledge of the specific type contained in the RefPtr<T>. This allows the RefPtr<T> to be used with types that are forward declared and included in modules that have no knowledge of the type pointed to, because any pointer can be cast to a 'const void*'.

For example suppose we have the following in 'Enemy.h':

namespace cocos2d
{
    class Sprite;
}

class Enemy
{
public:
   void doStuff();

private:
   RefPtr<cocos2d::Sprite>  mSprite;
};

If another file 'AI.cpp' wanted to call doStuff() on an enemy it could do so without having to include the header for sprite:

#include "Enemy.h"

void AI::doTick(Enemy & enemy)
{
   enemy.doStuff();
   // Do other stuff...
}

If we were to use 'const Ref*' it would require the compiler to have knowledge of the type in the module where the header was included. Which means we'd have to change 'Enemy.h' to look something like this:

#include <CCSprite.h>

class Enemy
{
public:
   void doStuff();

private:
   RefPtr<cocos2d::Sprite>  mSprite;
};

In this case it's not a huge problem, but in a large codebase with many headers we could end up with a lot of headers cross including each other unknowingly and this might cause issues such as circular includes etc. Essentially it's a tradeoff between minimising dependencies between headers some loss of type information. Originally the code was working the way you suggested (like in cocos2d::Vector<T>) but we switched to this method because of problems with headers cross including each other.

Do you think it's the right trade off for cocos2d-x, or should we switch back to the Vector<T> method instead?

BodbDearg:
Thinking about this more, there may be another way to fix this without loosing type info AND keeping our forward references... What if we store the type pointed to as a Ref* instead of 'T*'? 

So instead of:

private:
    T \* _ptr = nullptr;

private:
    Ref \* _ptr = nullptr;

We could then do a static pointer cast in the constructors and again in the dereferencing functions. Because these template functions should only get instantiated by the code actually using them (which needs to know the type anyway) we could be able get away with only knowing the type for 'Ref' in all other places so that release() etc. works in the destructor.

This might work actually, I think I'll give it a go and see how it fares!

dumganhar:
Actually, it's no a big deal to make it as `Ref*` or `void*`, right?
Since the method `RefPtrSupportFunctions::safeRetainRef` is only used in `RefPtr` class internally.
I think currently implementation is ok. :) Awesome work. :+1:

thoechsmann:
It looks like the Ref Object is only required for the memory management (release/retain). Only lua adds something more. 

Why do you not use std smart pointers?

If you need more than smart pointers (I am sure there are ways to move this functionality somewhere else) I would create a subclass of smart pointers (not sure if thats possible) or a new class that has a smart pointer member. Why do you want to recode that stuff??

dumganhar:
@thoechsmann , please refer to the discussion (http://cocos2d-x.org/forums/6/topics/30253?page=1&r=33375)

It's mainly because of performance and compatible issue if we switch all api to smartpointer.

BodbDearg:
@thoechsmann As dumganhar mentioned the main reason for not using smart pointers is performance. Retain/release with std::shared_ptr needs to be thread safe, so this involves expensive synchronisation every time this is done. The cost of these retain/releases can really add up if you're doing a lot of them...

There's also the API compatibility issue mentioned. std::shared_ptr has no knowledge of the internal reference count in cocos2d::Ref so it could end up destroying the object before the cocos2d::Ref reference counter reaches zero. We could use std::shared_ptr in cocos2d-x, but it would involve a lot of changes API in cocos2d-x itself. This class is designed to work with the existing API as it currently is - changing the entire memory management model of the API is well beyond the scope of one simple pull request :)

thoechsmann:
Ok, Got it. 

Performance will be better with a self made class (due to non-threadsafty). But changing everything to RefPtr<> will not be much less work than changing to smart_ptr. Maybe it can be better done in some iterations than the with to std:smart_ptr.

But nevertheless I highly prefer your RefPtr to the current manual release/retain.

dumganhar:
@thoechsmann 
We will not change every thing to `RefPtr<>`, `RefPtr<>` is just a helper class for game codes.

thoechsmann:
So a "create" will return a raw pointer and the user will have to create RefPtr for this raw pointer. 

The user code would be cleaner when "create" returns the RefPtr. This could (should) be the second step.

We do not allow our developers to use raw pointers if there is not a very strong reason (mainly in performance sensitive parts) to use raw pointers. Cocos is the only part where we use them currently.

dumganhar:
@thoechsmann ,

> The user code would be cleaner when "create" returns the RefPtr. This could (should) be the second step.

If we return `RefPtr` for create functions, it will also break compatibilities.
Thinking of the following codes:

``` c++
RefPtr<MoveTo> moveTo = MoveTo::create();
node->runAction(moveTo); // this will not work
// we have to change it to 
node->runAction(moveTo.get());  // access a raw pointer
```

and

``` c++
node->runAction(MoveTo::create()); // also be broken.
```

thoechsmann:
Yes. The whole API must be changed so that functions accept RefPtr instead of raw pointer. I think that this is not as much work as it sounds. But it could also be that more challenges arise once you start...

BodbDearg:
"The user code would be cleaner when "create" returns the RefPtr. This could (should) be the second step."

Should be no need to change existing create() functions to return RefPtr<T> - it can work with them as they are currently because RefPtr<T> is fully constructible and assignable from the raw pointer type too. We use RefPtr<T> with standard cocos2d-x create() functions in many places and it works well. If you don't need to maintain a long term reference to a pointer and only use it in the current call stack then a standard pointer works just fine as well. RefPtr<T> is really for cases where you want to maintain a long term reference (member variable etc.) to something and don't want to worry about manual retain/release.

BodbDearg:
@dumganhar 
"RefPtr<MoveTo> moveTo = MoveTo::create();
node->runAction(moveTo); // this will not work"

This should actually work because RefPtr<T> is implicitly convertible back to it's raw pointer type. We wanted to make RefPtr<T> as transparent as possible so it defines a cast operator to cast itself back to the raw type. It would have been a slight pain to use otherwise if we had to call .get() everywhere..

For example this code will work fine:

```
// Will work
RefPtr<cocos2d::String> stringRef = cocos2d::String::create("Stirng");
RefPtr<cocos2d::Number> numberRef = cocos2d::Number::create(2);
cocos2d::Array * myArray = cocos2d::Array::create();
myArray->addObject(stringRef);
myArray->addObject(numberRef);
```

The only scenario I've found where it WON'T work is with varargs lists. We could probably fix this somehow in C++ 11 though by doing variadic templates instead of varargs (depending on compiler support):

```
// Won't work:
// ERROR 'Cannot pass object of non-trivial type 'PFCCRef<cocos2d::Number>' through variadic function; call will abort at runtime'.
RefPtr<cocos2d::String> stringRef = cocos2d::String::create("Stirng");
RefPtr<cocos2d::Number> numberRef = cocos2d::Number::create(2);
cocos2d::Array * myArray = cocos2d::Array::create(stringRef, numberRef, nullptr);
```

This case can be fixed easily though by doing a manual .get() on the smart pointer:

```
// Will work:
RefPtr<cocos2d::String> stringRef = cocos2d::String::create("Stirng");
RefPtr<cocos2d::Number> numberRef = cocos2d::Number::create(2);
cocos2d::Array * myArray = cocos2d::Array::create(stringRef.get(), numberRef.get(), nullptr);
```

In practice the varargs problem doesn't seem to be much of an issue. We've yet to encounter one instance in our codebase where we ran into that specific issue, but it is something to bear in mind.

The user shouldn't need to make any modifications to the API in order to use RefPtr<T> transparently with the cocos. It should just work like a normal pointer in most cases with the exception of the varargs case just described.

dumganhar:
Good to know this. :)

thoechsmann:
Using the RefPtr also inside of the cocos framework would get rid of a lot of code lines. (Around 1000)
It would also reduce the likelihood of introducing new memory leaks. There a way better ways to improve performance than using raw pointers everywhere. I am pretty sure in nearly every single cocos app this will not be the bottle neck.

But I am happy to use it in our code. So thanks.

BodbDearg:
You're welcome!

Agree- using standard unsynchronised retain/release() (like with RefPtr<>) is extremely fast and shouldn't ever really be a performance concern.

ricardoquesada:
@all:
For the record, we tried to replace manual `retain`, `release`, `autorelease` with `shared_ptr<>` but the performance wasn't that good.

I tried other options as well, but unfortunately the best option for was was to keep using manual `retain`, `release`.

There is very long long discussion about it here:
http://www.cocos2d-x.org/forums/6/topics/30253

Regarding `RefPtr`, I like it (although I haven't tested it, I only skimmed the Pull Request).
Probably it should be part of cocos2d-x. And users who want to use it, they can use it.

BodbDearg:
Thanks for pointing out that discussion!

Here's an alternate implementation of the class which trades the ability to use forward references in some cases in return for added type safety and compile time checking in order to ensure no non cocos2d::Ref<> derived types can be used as a template parameter:

https://github.com/DarraghCoy/cocos2d-x/blob/add_refptr_class_alt_impl/cocos/base/CCRefPtr.h

With this implementation you can still forward reference if you are just returning the raw pointer type (calling .get() etc.), comparing pointers, or any other operation which doesn't really need specific knowledge of the type pointed to. You can also default or null construct a RefPtr without knowing the class implementation of the type pointed to. This means you can have inline getters in header files while just having access to a forward referenced type.

On assignment or construction from a raw pointer type however this implementation requires you to have knowledge of the type pointed to since the compiler needs to know that the 'T' parameter is derived from cocos2d::Ref. This means that typically you'll either have to implement setter type functions in the .cpp file (if you want to forward reference) or include the required header as a dependency of the class definition if you're not too worried about using forward refs.

I think this could be a good tradeoff between type safety and being able to not include headers of the types pointed to.

BodbDearg:
Just merged in the alternate implementation mentioned (at https://github.com/DarraghCoy/cocos2d-x/blob/add_refptr_class_alt_impl/cocos/base/CCRefPtr.h) for the purposes of this pull request. I think this might be the better option to go with.

ricardoquesada:
[ci build]

ricardoquesada:
[ci rebuild]

cocos/2d/Android.mk
```diff
@@ -137,6 +137,7 @@ renderer/CCRenderMaterial.cpp \
 ../base/CCGeometry.cpp \
 ../base/CCNS.cpp \
 ../base/CCRef.cpp \
+../base/CCRefPtr.cpp \
```
dumganhar: `CCRefPtr.cpp` was removed, right? We're now only having `CCRefPtr.h`.

dumganhar:
Hi @ricardoquesada 
So you also like this feature, right?
I have tested it and it worked perfectly.
@DarraghCoy , thanks to add this feature.
But since we're using `c++ 11` which supports `move semantic`, could you also make `RefPtr<>` supports `move semantic`? 
I don't know whether you're familiar with that, feel free let us know if you need help. :)

ricardoquesada:
+1 for adding it
+1 for supporting move semantics. All our container objects (including this one) MUST support move semantics.

dumganhar:
Yes, our containers( `Vector<>`, `Map<>`, `Value` ) have already supported `move semantics`. :)

cocos/2d/Android.mk
BodbDearg: Thanks! Yeah sorry forgot to remove that line from the makefile - should now be fixed.

BodbDearg:
Good suggestion with the move semantics! I've added support in the latest commit and additions to the RefPtr test which cause the move operator and constructor to be called.

BodbDearg:
Fixed up the build for win32 & linux, should be good to go now on all platforms!

dumganhar:
@DarraghCoy , thanks.

ricardoquesada:
@minggo @dumganhar 
Any update on this?
When is it going to be merged ? thanks.

minggo:
tag of v3.0 is set.
I think we can merge it now.

@thoechsmann 
Sorry for late reply.

BodbDearg:
Awesome! :)

dumganhar:
I resolved conflicts and merge this PR at https://github.com/cocos2d/cocos2d-x/pull/6365
Thanks.

BodbDearg:
Thank you!

PR #6031, opened 2014-03-27
closed #4561:Fixed read file more than once for label created by different font size.

WenhaiLin:


cocos/2d/CCFontFreeType.cpp
```diff
@@ -39,6 +39,9 @@ of this software and associated documentation files (the "Software"), to deal
 bool       FontFreeType::_FTInitialized = false;
 const int  FontFreeType::DistanceMapSpread = 3;
 
+static std::unordered_map<std::string, Data> CacheFontData;
+static std::unordered_map<std::string, int> CacheFontDataRef;
```
dumganhar: Please use `s_xXXX` for static function in cpp.

PR #6034, opened 2014-03-27
Batching of Label rendering.

thoechsmann:
Label is now derived from Node. Every characters is a Sprite added to that node. Render will batch them like other Sprites.

Problems left:
- render problems BMFonts
- set/getBlendMode not supported yet.
- inefficient code as character sprites are added/removed every time. Add caching.
- shadow and distant field not supported

The code is way more simple as there is no special rendering involved. The Label class does have too much legacy code and is very hard to read and maintain. Many code parts are only used when a Label is created in a special way.  Also the API for the user is hard too understand. This should be cleaned up. I would drop some legacy code. Is there any need for the LabelTTF class and the LabelBMFont class anymore. They are already using the Label class and lots of code is there to handle these special cases.

This is not the most efficient way to draw labels. Each character is a Sprite and the Renderer will have to copy the quad of each sprite to a single array before it will be drawn in one batch. One idea to fix this is to allow a Sprite to have more than one quad.

This pull request is for starting the discussion.


dumganhar:
Thanks, but this PR could not be automatically merged. Could you please pull the latest codes to resolve conflicts? Thanks.

WenhaiLin:
How to rendering fonts with different shader if using sprite for each letter instead of quads.Label using different shader for rendering outline,different text color[differ from color of node] or rendering by distant field.It's inefficient for each letter is sprite, don't you think so?

thoechsmann:
Different shaders should be no problem. The Renderer will merge the quads with the same material and render them in batches. The shader is just one property of the material.

About efficiency. As noted a Sprite for a each letter is not the optimal solution.

But that can be optimised in a later step. And it should be done in a way that is not limited to the Label class. 

The current label does try to optimise a lot, but the reducing the render calls is way faster than all of them together. This was not possible (or very hard) with the old renderer. But now we should take advantage of it.

cocosQA:
Dear thoechsmann:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6428

PR #6039, opened 2014-03-28
LabelTTF and LabelBMFont tagged as deprecated

WenhaiLin:


cocos/2d/CCLabelBMFont.cpp
```diff
@@ -38,9 +38,16 @@ of this software and associated documentation files (the "Software"), to deal
...
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#elif _MSC_VER >= 1400 //vs 2005 or higher
+#pragma warning (push)
+#pragma warning (disable: 4996)
+#endif
+
 NS_CC_BEGIN
 
-LabelBMFont * LabelBMFont::create()
+    LabelBMFont * LabelBMFont::create()
```
dumganhar: Wrong indention.

thoechsmann:
I would also remove all the code from Label, that creates a single Sprite. I cannot see any scenario where this is better. The small performance improvement for static labels is neglect-able. There is a lot of code in the label class to handle this. 

There are also lots of duplicate member variables in Label due to legacy code. 
There is _maxLineWidth, _labelWidth and labelDimension. Not clear what the difference is. Looks like they are nearly identical. 

Also _fontDefinition seems to be a legacy thing. It holds the same information that is already stored elsewhere.

The different behaviour depending if you create a label with FONTNAME or FONTNAME.ttf is also very confusing. Either get rid of system font rendering completely (I think this would be no great loss and it would be less code to maintain) or make it a fallback. If font is not found try to load system font .

thoechsmann:
Hey guys,

would be great if I can get some feedback on my comment. What are your thoughts.
One more comment about the Label class. There is a friend class LabelTextFormatter. I do not think this is a one of the very few scenarios where a friend class should be used. 
Also the LabelTextFormatter class is not used by anybody else than Label.

It makes refactoring even harder as you have to look also into the LabelTextFormatter to find what members of Label are required for.

ricardoquesada:
@Dhilan007 @dumganhar @minggo:

Could you help @thoechsmann with this?
He is using cocos2d-x for very advanced things, so his feedback is very important for us. 

I would happily help him, but I'm not familiar with the internals of the new Label code.
Thanks.

minggo:
@thoechsmann 
Sorry for late reply.

@Dhilan007 
Please answer the @thoechsmann's question.

WenhaiLin:
@thoechsmann Thanks for your comments. The 'maxLineWidth' is used for force line breaks if the set value not equal zero.But that doesn't mean the actual size of label to be equal to 'maxLineWidth'.The dimensions/width/height of label mean the actual size.

WenhaiLin:
@thoechsmann The _fontDefinition be used for backwards compatibility.It's good suggestion for try to load system font if the font is not found.Agree with you.

WenhaiLin:
@thoechsmann I will try to make the API and variable names  more clearly.Looking forward to your suggestions.

thoechsmann:
Thanks for the answers Dhilan. 

What about the suggestion to remove support for old Labels completely? This would be the best time to get rid of it as 3.0 is breaking backwards compatibility anyway. It would make the Label class way easier to understand and maintain.

At least I would never create a single Sprite for a Label like you do for LabelTTF. It is just one more special case to handle (and there is quite a lot code to handle this).

minggo:
I think we can move deprecated functions at the end of class declaration. And move implementation of these functions at the end of cpp.

thoechsmann:
Why not remove them completely? There are not just functions for the old Labels, but a lot of case distinctions.

minggo:
I think there are two things
- duplicate functions
  
  I think we can mark them deprecated to keep compatibility.
- load system font in Label if font is not found

It is the feature makes a lot of case distinctions. @Dhilan007  told me that he wanted to split this feature in another class. But now we have released RC1 version. And developers complained about the renaming or other modifications that brought them more work. So i think we can not do it like this.

minggo:
> At least I would never create a single Sprite for a Label like you do for LabelTTF. It is just one more special case to handle (and there is quite a lot code to handle this).

Any suggestions about this?

thoechsmann:
My suggestions for this:
It is fine if there are still classes for bmfont and ttf. Label should have create functions for bmfont, system font, and ttf file. I would not add a fallback for the system font class. At least this should not be the default behavior. It's is nearly never wanted. 

All create functions create a font atlas. The characters are quads.

minggo:
> Label should have create functions for bmfont, system font, and ttf file. I would not add a fallback for the system font class.

@thoechsmann 
I can not get the idea. Could you please describe it in more detail?
Thanks.

PS: How to quote others' comment as you did? Did you edit it by yourself?

thoechsmann:
I replied via email. Thats maybe the reason for the quote.

Label will have create functions like:

createWithTTF
createWithSystemFont
createWithFNT (or BMFont)

This create functions will all create the texture atlas (or several) for the label. From there on they are all handled equally. That means that for example creating a font with a system font will NOT create a single sprite, like it is done at the moment, but will create like all other scenarios, quads for each character.

The other thing was about the fallback to a system font. I think this was quite clear. Never do a fallback. The user will have to implement this if he really wants that.

minggo:
@thoechsmann 
I agree with you. Every function should be clear with the function name and just do one thing. And i have discussed with @Dhilan007, the codes in Label to keep compatibility should be moved to LabelTTF which is deprecated. Label should keep clean.

thoechsmann:
Beside the create functions there will not be much code in the LabelTTF class. 
This deprecated class can be used as before, however it will do things internally different to the old implementation (like not creating a single sprite)

minggo:
@Dhilan007 
Could you please describe what you want to do to improve it?
Thanks.

WenhaiLin:
@thoechsmann @minggo I think create quads for each character with font name is not to be able.We can't load system font for freetype. It's mean we can't get enough information to align text by platform-dependent code at present . 

I think label should have create functions like:
create
createWithSystemFont
createWithTTF
createWithBMFont
createWithCharMap

And remove createWithFontDefinition/setFontDefinition/getFontDefinition.

thoechsmann:
Ok, understand.

PR #6043, opened 2014-03-28
closed #4590, fixed rewind problems

zilongshanren:


cocos/audio/ios/CDAudioManager.m
```diff
@@ -118,6 +118,8 @@ -(void) pause {
 -(void) rewind {
     self->paused = NO;
     [audioSourcePlayer setCurrentTime:0];
+    [audioSourcePlayer play];
+    stopped = NO;
```
dumganhar, line 122: Here should be `self->stopped = NO`.

PR #6052, opened 2014-03-28
Fixed bugs for Editor

pipu:
Fixed bugs for Editor

### DIFF
--- a/cocos/ui/UIButton.cpp
+++ b/cocos/ui/UIButton.cpp
@@ -392,12 +392,14 @@ void Button::onPressStateChangedToNormal()
     }
     else
     {
-        _buttonNormalRenderer->stopAllActions();
-        _buttonNormalRenderer->setScale(_normalTextureScaleXInSize, _normalTextureScaleYInSize);
         if (_scale9Enabled)
         {
-            updateFlippedX();
-            updateFlippedY();
+            updateTextureRGBA();
+        }
+        else
+        {
+            _buttonNormalRenderer->stopAllActions();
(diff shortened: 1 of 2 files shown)


cocos/ui/UITextField.cpp
```diff
@@ -433,21 +446,59 @@ Size TextField::getTouchSize()
...
+        }
     }
+    
     const char* content = strText.c_str();
     if (isPasswordEnabled())
     {
         _textFieldRenderer->setPasswordText(content);
         _textFieldRenderer->setString("");
-        _textFieldRenderer->insertText(content, static_cast<int>(strlen(content)));
+        _textFieldRenderer->insertText(content, strlen(content));
```
zilongshanren, line 495: why do you discard the static_cast?  there might be warnings here. We added the static_cast<int> to get ride of the implicit type conversion.

cocos/editor-support/cocostudio/WidgetReader/LayoutReader/LayoutReader.cpp
```diff
@@ -116,6 +116,11 @@
         }
         panel->setLayoutType((LayoutType)DICTOOL->getIntValue_json(options, "layoutType"));
         
+        int bgimgcr = DICTOOL->getIntValue_json(options, "colorR");
```
zilongshanren, line 119: WidgetReader also parse colorR/G/B, why parse it again?

I know there are "bgColorB, bgColorG and bgColorR" for our panel. 
do you have any special consideration? or do I miss something?

zilongshanren:
the bugs have been fixed, so I think we should close this PR.

PR #6055, opened 2014-03-29
Cocos2d-x 3.0 Windows Phone 8 support

stammen:
This pull requests contains all of the projects and files needed to build cocos2d-x 3.0 for Windows Phone 8.0.

Future pull requests will add support for the following items:
1. Accelerometer
2. Custom shaders (needed for shader tests)
3. Console support
4. modifications to cocos2d-console to fully support cpp template for WP8

### DIFF
--- a/build/cocos2d-wp8.vc2012.sln
+++ b/build/cocos2d-wp8.vc2012.sln
@@ -3,12 +3,26 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libcocos2d", "..\cocos\2d\cocos2d_wp8.vcxproj", "{B5AF91B3-64EA-44E1-84B0-D759E93758FB}"
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libSpine", "..\cocos\editor-support\spine\proj.wp8\libSpine.vcxproj", "{7D4FC6EB-9497-4804-98F3-3EAEDC896154}"
+	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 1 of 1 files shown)


dumganhar:
[ci rebuild]

minggo:
Does it only support WP8?
How about Windows 8?

minggo:
Why use pthread?
Does WP8 doesn't support std::thread?

minggo:
@dumganhar 
I think it is our build system error.
Because this PR only modified CCLayer.cpp. And it seems not problem of the modification.
So i think we can merge it.

@koowolf
Please test it.
Thanks.

stammen:
@minggo 
WP8 is using std:thread. Where do you see pthread?

minggo:
I found there are CCPThreadWinRT.cpp/.h.

stammen:
That is to support old pthread code from cocos2d-x 2.2.  It is not used in the new 3.0 cocos2d-x code. For example, the texture loading and networking code uses cocos2d-x 3.0 code unmodified. This version of wp8 supports curl and websockets using the same code as the other platforms. All of the networking and curl tests pass.

I can remove the CCPThreadWinRT.cpp/.h files if you want.

For example, look at httpclient.cpp or websocket.cpp and you will see WP8 uses std::thread. Your team did a great job with with the C++ 11 changes.

minggo:
Ok, no problem.
You can keep it if you think it is needed.

stammen:
@minggo 

WinRT support will be added in a separate pull request. It will take me about a week after I finish the remaining WP8 items. WinRT networking support will take some time as it does not have access to WinSock API which is available to WP8.

minggo:
@stammen

Thanks for your awesome work.

stammen:
@minggo 

Thanks! The new cocos2d-x 3.0 codebase made it a lot easier to do the WP8 work. The cocos2d-x team is writing excellent cross-platform code.

minggo:
Thanks.

PR #6060, opened 2014-03-30
CCFileUtilsApple: Change getWritablePath() to return Application Support Directory subdir

wefiends:
This is a design decision, but the Application Support directory seems more appropriate for writing UserDefaults, etc. This follows the directories used on Windows (Local Settings dir), Linux (~/.config), and Android (/data/data...).

This commit returns a directory under Application Support using the Bundle ID. This path always writable in both sandboxed and non-sandbox OS X apps and on iOS.

See Apple's [File System Programming Guide](https://developer.apple.com/library/ios/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/AccessingFilesandDirectories/AccessingFilesandDirectories.html#//apple_ref/doc/uid/TP40010672-CH3-SW3)

### DIFF
--- a/cocos/2d/platform/apple/CCFileUtilsApple.mm
+++ b/cocos/2d/platform/apple/CCFileUtilsApple.mm
@@ -329,7 +329,7 @@ std::string FileUtilsApple::getWritablePath() const
     // Save to Documents folder
     NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
     NSString *documentsDirectory = [paths objectAtIndex:0];
-    std::string strRet = [bundleDirectory UTF8String];
+    std::string strRet = [documentsDirectory UTF8String];
     strRet.append("/");
     return strRet;
 #else


dumganhar:
So, is there any problems in current writable path?

minggo:
It is ok that getWritabelPath() to return a path that can be write into. If this function changed, it will break compatibility. Many developers use it to record something. 

So please not modify it.

wefiends:
The problem is the `getWritablePath()` is being used in two different ways. Like minggo mentioned, developers might use it to write files that should be user accessible (like screenshots), but internally cocos2d-x uses it to write application settings (UserDefaults, HttpClient).

On the Windows, Linux, and Android, `getWritablePath()` returns directories which are by default hidden. On Mac and iOS it returns the user's Document folder which is not an appropriate place to save applications settings.

For example, on Mac and iOS, UserDefaults creates a file called `~/Documents/UserDefault.xml`. There is no indication what application this file belongs to or what it does.

Perhaps, the solution is adding a `getSettingsPath()` which returns the appropriate directory to write settings into for each platform. Then update UserDefaults and HttpClient to use this directory instead. Developers could also use it to write any other configuration files.

minggo:
@wefiends 
Now `UserDefault` uses `NSUserDefaults` to store data. `getWritablePath()` is used just fro compatibility.

wefiends:
I still think that it's strange that `getWritablePath()` returns the user's Documents folder on Mac/iOS but hidden folders on the other platforms. If the method is meant to write user accessible files, it should return an appropriate path on all platforms.

zj831007:
IOS Data Storage Guidelines: https://developer.apple.com/icloud/documentation/data-storage/

PR #6066, opened 2014-03-31
use xxhash to generate material id

minggo:


minggo:
@koowolf
Could you please modify it for win32 and WP8 and send me a pull request?
Thanks.

heliclei:
[ci rebuild]

PR #6097, opened 2014-04-01
Android crash fix on new devices

newnon:
It is 100% crash on new android devices

### DIFF
--- a/cocos/2d/platform/android/jni/JniHelper.cpp
+++ b/cocos/2d/platform/android/jni/JniHelper.cpp
@@ -143,7 +143,7 @@ namespace cocos2d {
             return false;
         }
 
-        JniHelper::classloader = _c;
+        JniHelper::classloader = cocos2d::JniHelper::getEnv()->NewGlobalRef(_c);
         JniHelper::loadclassMethod_methodID = _m.methodID;
 
         return true;


minggo:
What did you mean `new devices`?

newnon:
i have 4.2 and on it 100% crash it is because of new garbage collector.
http://stackoverflow.com/questions/14765776/jni-error-app-bug-accessed-stale-local-reference-0xbc00021-index-8-in-a-tabl
where written that from 4.0

cocos/2d/platform/android/jni/JniHelper.cpp
```diff
@@ -143,7 +143,7 @@ jclass _getClassID(const char *className) {
             return false;
         }
 
-        JniHelper::classloader = _c;
+        JniHelper::classloader = cocos2d::JniHelper::getEnv()->NewGlobalRef(_c);
```
dumganhar, line 146: Do we need to release the global reference some where? If not, could it cause memory leak?

minggo:
@newnon 
Thanks for your feedback. We will test it.

cocos/2d/platform/android/jni/JniHelper.cpp
newnon, line 146: If i understand right this function call only once? if not yes we can try to release old value just before assigning new

squirrelChen:
@newnon  I tested in device 4.2.2, and didn't crash.

newnon:
may be but it in not good if you application crashes on some devices

newnon:
you can read full article about dangerous of such using jni variables here http://android-developers.blogspot.cz/2011/11/jni-local-reference-changes-in-ics.html

minggo:
@newnon 
Thanks.
I will read the docs carefully tomorrow.

sennin:
I have this same issue on Android 4.4

minggo:
I think we don't have to use classloader to get class id. As we did in v2.x, we can get class id like this

``` java
static jclass getClassID_(const char *className, JNIEnv *env)
    {
        JNIEnv *pEnv = env;
        jclass ret = 0;

        do 
        {
            if (! pEnv)
            {
                if (! getEnv(&pEnv))
                {
                    break;
                }
            }

            ret = pEnv->FindClass(className);
            if (! ret)
            {
                 LOGD("Failed to find class of %s", className);
                break;
            }
        } while (0);

        return ret;
    }
```

Then we don't have to use classloader, and will not cause crash.

More detail information can refer to https://github.com/cocos2d/cocos2d-x/blob/master/cocos2dx/platform/android/jni/JniHelper.cpp#L80

sennin:
After update to RC1 I noticed many strange things after all. Buttons which are used to start JNI calls are blinking now after click. Some of my native methods are just not called (I am registering those methods on my JNI_OnLoad which I merged with yours; some methods are working, some are not).

minggo:
@sennin 
Could you please show me a demo?

sennin:
I am not sure what you mean by "demo".
This is my JNI_OnLoad method where I am loading my natives.

Lets focus on two methods:
- nativeEndApp which close app after dismiss an ad (this is working),
- nativeSinglePlayer which open some dialog after dismiss an ad (this is not working).

``` cpp
static JNINativeMethod adsMethodTable[] = {
  {"nativeEndApp", "()V", (void*) nativeEndApp},
  {"nativeEndGame", "()V", (void*) nativeEndGame},
  {"nativePushAndPlay", "()V", (void*) nativePushAndPlay},
  {"nativeSinglePlayer", "()V", (void*) nativeSinglePlayer},
  {"nativeContinueGame", "()V", (void*) nativeContinueGame},
};

jint JNI_OnLoad(JavaVM* vm, void* reserved){
    //from cocos2dx javaactivity
    cocos2d::JniHelper::setJavaVM(vm);

    //my code
    JNIEnv* env = cocos2d::JniHelper::getEnv();

    if (vm->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION_1_6) != JNI_OK) {
        return -1;
    }

    //Try to find TapForTapBridge on the device side
    jclass _adsClass = env->FindClass(adsClassPath);
    if(!_adsClass){
        LOGD("Cannot find class: %s", adsClassPath);

        //Cannot find ads because of full version of app, so we clear pending exception
        //This is necessary to find other classes
        env->ExceptionClear();
    } else {
        adsClass = env->NewGlobalRef(_adsClass);

        //Register natives
        env->RegisterNatives(_adsClass, adsMethodTable, sizeof(adsMethodTable) / sizeof(adsMethodTable[0]));
    }

    /* ... */

    return JNI_VERSION_1_6;
}
```

here are methods:

``` cpp
void nativeEndApp(JNIEnv* env, jobject thiz){
    CCLOG("TEST");
    /* ... */
}
void nativeSinglePlayer(JNIEnv* env, jobject thiz){
    CCLOG("TEST");
    /* ... */
}
```

and this is my java code:

``` java
private static void dismissAd(int whatToDo) {
        Log.i(TAG, "whatToDo: " + whatToDo);
        switch(whatToDo) {
            case END_APP:
                nativeEndApp();
                break;
            case SINGLE_PLAYER:
                Cocos2dxHelper.runOnGLThread(new Runnable() {
                    @Override
                    public void run() {
                        nativeSinglePlayer();
                    }
                });
                break;
                //...
        }
    }
```

I am just tested it again because end app is called on Java thread and single player on GLThread, so maybe this is a problem and in fact it is. runOnGLThread isn't working now. Is there other option?

minggo:
What i mean is that if there is a github link that i can download then reproduce your issue.

sennin:
No there is no github repository. I did some investigations and end with conclusion that `Cocos2dxHelper.runOnGLThread` isn't working for me.

If I call function like that:

``` java
Cocos2dxHelper.runOnGLThread(new Runnable() {
                    @Override
                    public void run() {
                        nativeMethod();
                    }
                });
```

nativeMethod isn't called.

minggo:
@sennin 
Which version did you use?
This function has problem in latest codes and RC1.

sennin:
I am merged now with
https://github.com/cocos2d/cocos2d-x/pull/6159/files

Everything is working now.

PR #6101, opened 2014-04-01
Auto culling support for Label class

thoechsmann:
I moved the culling test function from Sprite up to the Node class. Label is now using the same test as Sprite to skip Labels that are offscreen. Only complete Labels will be tested and not single characters.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -986,7 +986,7 @@ void Label::draw(Renderer *renderer, const kmMat4 &transform, bool transformUpda
     // Don't do calculate the culling if the transform was not updated
     _insideBounds = transformUpdated ? isInsideBounds() : _insideBounds;
     
-//    if(_insideBounds)
+    if(_insideBounds)
     {
         _customCommand.init(_globalZOrder);
         _customCommand.func = CC_CALLBACK_0(Label::onDraw, this, transform, transformUpdated);


ricardoquesada:
Yes, I think it should work.

It will fail with particles, because particles don't update their `contentSize`, but that is a bug in the particles.
So, in order to prevent confusion with the users I think we should rename `isInsideBounds()` to `isContentSizeInsideBounds()`  (or something like that)

thoechsmann:
Currently I only added culling to Labels in addition to the Sprites. 

I would add a parameter to the isContentSizeInBounds function for the bounds the user want to check. the draw Function will use the whole screen as bounds. 

But with the parameter the function can be used by users for other bounds checks, too.

ricardoquesada:
@thoechsmann
What is the Use Case of passing a parameter?
I mean, in which scenario would you like to use a different value than the size of the screen?
thanks.

thoechsmann:
if scissoring is activated. Maybe this can be detected automatically...

minggo:
I think it is better to culling the whole screen one time a frame. Not to do culling on every node.

ricardoquesada:
@minggo 
What Thomas said was that we could add an optional "screen size" parameter to the culling function.
By default it is the size of the view port, but you can make smaller if needed.

minggo:
@ricardoquesada 
Thanks for you explain.

What i wonder is that could we have a method just do culling one time per loop. Now Label should do its culling, Sprite should do its culling, and so on. Is there any method to do culling in a place for all of them. I know we can not achieve it in current design.

ricardoquesada:
@minggo 
As I mentioned earlier, to all the nodes have the `contentSize()` set correctly, and for some nodes it might not be efficient to check it (eg: `Scene`).
So, it make sense to check the culling at `draw` time.

minggo:
Ok, got it.

cocosQA:
Dear thoechsmann:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6425

PR #6102, opened 2014-04-01
performance fixes in QuadCommand

ricardoquesada:
performance fixes in QuadCommand

### DIFF
--- a/cocos/2d/renderer/CCQuadCommand.cpp
+++ b/cocos/2d/renderer/CCQuadCommand.cpp
@@ -29,11 +29,16 @@
 
 NS_CC_BEGIN
 
+
+static  void convertIntToByteArray(int value, int* output)
+{
+    *output = value;
+}
+
 QuadCommand::QuadCommand()
 :_textureID(0)
 ,_blendType(BlendFunc::DISABLE)
 ,_quadsCount(0)
-,_dirty(false)
 ,_shader(nullptr)
 ,_quads(nullptr)
 {
@@ -43,6 +48,7 @@ QuadCommand::QuadCommand()
 void QuadCommand::init(float globalOrder, GLuint textureID, GLProgram* shader, BlendFunc blendType, V3F_C4B_T2F_Quad* quad, ssize_t quadCount, const kmMat4 &mv)
 {
     _globalOrder = globalOrder;
+
     _textureID = textureID;
     _blendType = blendType;
     _shader = shader;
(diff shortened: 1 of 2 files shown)


minggo:
I think `generateMaterialID` should be `protected` or `private`.

ricardoquesada:
yes, you are right. But this patch didn't change the scope of `generateMaterialID()`.
This patch fixes the cache. There was a `dirty` flag but it was always `dirty = true;`. 
I think it is `public` for legacy reasons, but, yes, feel free to change it.

minggo:
Ok, I will modify it in other thread.

ricardoquesada:
ok. thanks.

minggo:
@ricardoquesada 
Done in #6174.

@thoechsmann 

> Looks like the _last members are not initialised. Do you need them at all? Just move the test if a parameter has changed a bit upwards.

Yep, `_lastShader` and `_lastBlendType` are not initialized. I will fix it in another PR. But what did you mean they were needless?

thoechsmann:
Like this. If I am not missing something, it will do the same without the need of the "last" members.

```
void QuadCommand::init(float globalOrder, GLuint textureID, GLProgram* shader, BlendFunc blendType, V3F_C4B_T2F_Quad* quad, ssize_t quadCount, const kmMat4 &mv)
{
    if( _textureID != textureID || _blendType.src != blendType.src || _blendType.dst != blendType.dst || _shader != shader) {
        generateMaterialID();
    }

    _textureID = textureID;
    _blendType = blendType;
    _shader = shader;
    _globalOrder = globalOrder;
    _quadsCount = quadCount;
    _quads = quad;
    _mv = mv;
}
```

thoechsmann:
small fix:

```
void QuadCommand::init(float globalOrder, GLuint textureID, GLProgram* shader, BlendFunc blendType, V3F_C4B_T2F_Quad* quad, ssize_t quadCount, const kmMat4 &mv)
{
    if( _textureID != textureID || _blendType.src != blendType.src || _blendType.dst != blendType.dst || _shader != shader) {
       _textureID = textureID;
       _blendType = blendType;
       _shader = shader;
        generateMaterialID();
    }

    _globalOrder = globalOrder;
    _quadsCount = quadCount;
    _quads = quad;
    _mv = mv;
}
```

minggo:
Yep, you are right.
Thanks.

minggo:
@thoechsmann 
Done in #6230.

@ricardoquesada 
Am i missing something?
Why add _lastXXX variables?

ricardoquesada:
no need for the `_last` ivars. @thoechsmann 's patch looks good.

PR #6104, opened 2014-04-02
closed #4635: ControlSwitch's marksprite is shown at a wrong place

dumganhar:


extensions/GUI/CCControlExtension/CCControlSwitch.cpp
```diff
@@ -204,27 +201,6 @@ bool ControlSwitchSprite::initWithMaskSprite(
...
-
-        program->link();
-        CHECK_GL_ERROR_DEBUG();
-
-        program->updateUniforms();
-        CHECK_GL_ERROR_DEBUG();                
-
-        _textureLocation = program->getUniformLocation("u_texture");
-        _maskLocation = program->getUniformLocation("u_mask");
-        CHECK_GL_ERROR_DEBUG();
```
dumganhar, line 227: Removed unneeded program since we have used `ClippingNode` to refactoring `ControlSwitch`.
This change fixed the issue. Other changes just removed some unused files.

PR #6106, opened 2014-04-02
Fixed #4637: Add loop parameter to Animation::initWithSpriteFrames

pandamicro:


dumganhar:
So `createWithSpriteFrames` also needs add a `loop` parameter. But i couldn't see that change.

PR #6110, opened 2014-04-02
closed #4639: FileUtils::isFileExist doesn't consider SearchPaths and ResolutionOrders

dumganhar:


cocos/2d/platform/CCFileUtils.h
```diff
@@ -354,7 +354,12 @@ class CC_DLL FileUtils
      *  @return The new filename after searching in the filename lookup dictionary.
      *          If the original filename wasn't in the dictionary, it will return the original filename.
      */
-    virtual std::string getNewFilename(const std::string &filename);
+    virtual std::string getNewFilename(const std::string &filename) const;
+    
+    /**
+     *  Checks whether file exists without considering search paths and resolution orders.
+     */
+    virtual bool isFileExistInternal(const std::string& filename) const = 0;
```
minggo, line 362: It is better to added into `protected` section. Because it is used internally.

cocos/2d/platform/CCFileUtils.h
```diff
@@ -290,7 +290,7 @@ class CC_DLL FileUtils
      *  @param strFilePath The path of the file, it could be a relative or absolute path.
      *  @return true if the file exists, otherwise it will return false.
      */
-    virtual bool isFileExist(const std::string& filePath) const = 0;
+    virtual bool isFileExist(const std::string& filename, std::string* outFullpath = nullptr) const;
```
minggo: Why add out parameter in `isFileExist`?
It is strange.

cocos/2d/platform/CCFileUtils.h
dumganhar, line 362: It was in `protected` section.

cocos/2d/platform/CCFileUtils.h
dumganhar: Fixed. I wanted to re-use some codes. But yes, the interface should not do other things. Thanks.

PR #6117, opened 2014-04-02
Fixed bone's position is wrong  when animation has only a single bone

mikeweib:
when sub bone is single frame in animation and parent bone is not single frame. the transform of sub bone is error.


2youyou2:
Thanks a lot.
I have tested it,  it's really a problem.

mikeweib:
How fix it?

dumganhar:
@2youyouo2 who takes charge of Armature said it's good for merging this PR.

dumganhar:
@mikeweib thanks.

PR #6121, opened 2014-04-02
physic object is still drawn on screen after destroyed

jinyuliao:
test case:
1.add object using PhysicsBody::createEdgeSegment(p1, p2)
2.remove the object
3.the object is still drawn on screen because _bodies is empty so the DrawNode's clear() method won't be called

### DIFF
--- a/cocos/physics/CCPhysicsWorld.cpp
+++ b/cocos/physics/CCPhysicsWorld.cpp
@@ -196,17 +196,16 @@ void PhysicsWorldCallback::queryPointFunc(cpShape *shape, cpFloat distance, cpVe
 
 void PhysicsWorld::debugDraw()
 {
-    if (_debugDraw == nullptr)
-    {
-        _debugDraw = new PhysicsDebugDraw(*this);
-    }
-
     if (_bodies.empty())
     {
         CC_SAFE_DELETE(_debugDraw);
         return ;
     }
-    
+    else if (_debugDraw == nullptr)
+    {
+        _debugDraw = new PhysicsDebugDraw(*this);
+    }
+
     if (_debugDraw)
     {
         if (_debugDraw->begin())


cocos/physics/CCPhysicsWorld.cpp
```diff
@@ -201,7 +201,7 @@ void PhysicsWorld::debugDraw()
         _debugDraw = new PhysicsDebugDraw(*this);
     }
     
-    if (_debugDraw && !_bodies.empty())
```
boyu0, line 204: Hmm, Why `!_bodies.empty()` need to be deleted?

boyu0, line 204: Sorry, I didn't see your description, so ignore this.

boyu0, line 204: I think there is another way to fix this: remove _debugDraw if _bodies is empty. I think this can improve the performance, what's your opinion?

boyu0, line 204: Like this:

```
    if(_bodies.empty())
    {
        CC_SAFE_DELETE(_debugDraw);
        return;
    }
```

jinyuliao, line 204: yeah,that's a better solution

dumganhar, line 204: @mgcL 
Could you please update your codes according to the new solution?

jinyuliao, line 204: sorry for the later reply,i will update my codes right now

cocosQA:
Dear mgcL:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6424

PR #6123, opened 2014-04-03
set _node to null in createNodeWithSceneFile

ololomax:
_node was being set to null only in constructor, so when we call method createNodeWithSceneFile with path to existing file and then call it again with path to missing file (not calling destroyInstance method), it returns a pointer to previously created node, what may lead to a crash.


dumganhar:
Thank you.

PR #6125, opened 2014-04-03
update armature

2youyou2:
1.fixed set blend func
2.fixed changedisplay does not rebind parentbone


cocos/editor-support/cocostudio/CCBone.cpp
```diff
@@ -249,7 +249,7 @@ void Bone::applyParentTransform(Bone *parent)
 
 void CCBone::setBlendFunc(const BlendFunc& blendFunc)
```
dumganhar: We have removed `CC` prefix in develop.

PR #6126, opened 2014-04-03
Fix crash when has special character

kienvn:
Fix crash when has special character


cocos2dx/label_nodes/CCLabelBMFont.cpp
```diff
@@ -1146,6 +1146,7 @@ void CCLabelBMFont::updateLabel()
                         if (index < 0) continue;
 
                         CCSprite* characterSprite = (CCSprite*)getChildByTag(index);
+                        if(characterSprite)
                         characterSprite->setPosition(ccpAdd(characterSprite->getPosition(), ccp(shift, 0.0f)));
```
dumganhar: Thanks. Please fix the wrong indention. 

``` c++
if (xxx)
    xxx  (4 spaces)
```

PR #6130, opened 2014-04-03
closed #4557: Adds `EventListener::setEnabled/isEnabled`.

dumganhar:


cocos/2d/CCEventListener.h
```diff
@@ -74,34 +74,80 @@ class EventListener : public Ref
     
     /** Clones the listener, its subclasses have to override this method. */
     virtual EventListener* clone() = 0;
-    
+
+    /** Enables or disables the listener
+     *  @note Only listeners with `enabled` state will be able to receive events.
+     *        When an listener was initialized, it's enabled by default.
+     *        For a `scene graph priority` listener, to receive an event, excepting it was `enabled`,
+     *        it also shouldn't be in `pause` state.
```
minggo: An event listener can receive events when it is `enabled` and is not `paused`. `pause` is always false when it is a fixed priority listener.

PR #6131, opened 2014-04-03
Fix crash in `PerformanceTest/Node children test`.

dumganhar:


tests/cpp-tests/Classes/PerformanceTest/PerformanceNodeChildrenTest.cpp
```diff
@@ -947,6 +947,10 @@ void VisitSceneGraph::update(float dt)
     CC_PROFILER_START( this->profilerName() );
     this->visit();
     CC_PROFILER_STOP( this->profilerName() );
+
+    // Call `Renderer::render` to prevent crash if current scene is destroyed.
+    // The draw commands associated with current scene should be cleared.
+    Director::getInstance()->getRenderer()->render();
```
dumganhar, line 953: Instead of invoking `Renderer::render`, we need to add a new method called `cleanRenderQueue` for cleaning all render commands in the queue.

PR #6138, opened 2014-04-03
not set _orderOfArrival to 0 after in visit

minggo:
There may be a problem in this situation

``` c++
auto node1 = Node::node();
auto node2 = Node::node();
addChild(node1);
addChild(node2);
```

`node1` will be visited before `node2` because it is added first. But after visit, _orderOfArrival of node1 and node2 are 0. And the implementation of `nodeComparisonLess` is

``` c++
bool nodeComparisonLess(Node* n1, Node* n2)
{
    return( n1->getLocalZOrder() < n2->getLocalZOrder() ||
           ( n1->getLocalZOrder() == n2->getLocalZOrder() && n1->getOrderOfArrival() < n2->getOrderOfArrival() )
           );
}
```

It will return false, which means node2 will be visited before node1.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -983,9 +983,6 @@ void Node::visit(Renderer* renderer, const kmMat4 &parentTransform, bool parentT
     {
         this->draw(renderer, _modelViewTransform, dirty);
     }
-
-    // reset for next frame
-    _orderOfArrival = 0;
  
     kmGLPopMatrix();
 }


minggo:
Of course, we should invoke `Node::reorderChild()` to trigger sort operation.

@ricardoquesada 
Why set `_orderOfArrival` to 0?

And i found Label::visit() and SpriteBatchNode::visit also invoke `setOrderOfArrival(0);`.
Why?

Am i missed something?

minggo:
@ricardoquesada 
What's your opinion?

dumganhar:
I also feel confused about that.

dumganhar:
@minggo , why did you close this PR?

minggo:
I wrongly deleted the corresponding branch.
No matter, can fix it in other pull request after confirm it.

ricardoquesada:
I can't remember why it is there.
But removing that code before releasing v3.0 is too risky.

minggo:
Ok, so remove it in v3.1.

PR #6145, opened 2014-04-04
CCGLView: Add createWithFullscreen overloaded method that sets video mode.

wefiends:
There is currently no way to pass the desired size for a fullscreen window. It always sets the window size to the current video mode.

This adds an overloaded method that passes a size to GLFW. If the size doesn't match an available video mode, GLFW will set it to closest available size.

### DIFF
--- a/cocos/2d/platform/desktop/CCGLView.h
+++ b/cocos/2d/platform/desktop/CCGLView.h
@@ -39,7 +39,6 @@ public:
     static GLView* create(const std::string& viewName);
     static GLView* createWithRect(const std::string& viewName, Rect size, float frameZoomFactor = 1.0f);
     static GLView* createWithFullScreen(const std::string& viewName);
-    static GLView* createWithFullScreen(const std::string& viewName, Size size);
     static GLView* createWithFullScreen(const std::string& viewName, const GLFWvidmode &videoMode, GLFWmonitor *monitor);
 
     /*


cocos/2d/platform/desktop/CCGLView.cpp
```diff
@@ -374,6 +386,15 @@ bool GLView::initWithFullScreen(const std::string& viewName)
     return initWithRect(viewName, Rect(0, 0, videoMode->width, videoMode->height), 1.0f);
 }
 
+bool GLView::initWithFullScreen(const std::string &viewName, cocos2d::Size size, float frameZoomFactor)
```
dumganhar: Why does full screen need `size`?

dumganhar: And frameZoomFactor isn't needed since you never use it.

wefiends: When creating a fullscreen window with GLFW, setting the size will cause GLFW to switch video modes.

wefiends: I'll drop the frameZoomFactor param and pass in 1.0f.

dumganhar: But how does user use this function? What `size` should it pass in?
Is it possible to add a API like:

``` c++
GLView::createWithFullScreen(const std::string& viewname, int videoMode)
```

?

wefiends: I can change the method to take a `const GLFWvidmode` instead and create a `Rect` to pass to `initWithRect`.

 Developers need to retrieve the available video modes:
`GLFWmonitor *monitor = glfwGetPrimaryMonitor();`
`int count = 0;`
`const GLFWvidmode *modes = glfwGetVideoModes(monitor, &count);`

And choose one that's appropriate for their game. I can add a method that returns that array in CCGLView.

Ultimately, GLFW expects a width and height when creating the window, and it chooses the closest available video mode. You cannot directly pass it a `GLFWvidMode`.

dumganhar: > Ultimately, GLFW expects a width and height when creating the window, and it chooses the closest available video mode. You cannot directly pass it a GLFWvidMode.

good to know this. :)

dumganhar:
When i tested this PR on mac which has two monitors, whatever the `Size` value is, the game view is always at my main monitor (MBP's screen).
PS: I used MBP 2013 retina(2880*1800) with an external screen with 1920x1080 resolution.

wefiends:
Yes, the current createWithFullscreen defaults to the primary monitor (as reported by the OS). This was the behavior even before my patch.

In the latest commit:

`createWithFullScreen(const std::string& viewName, const GLFWvidmode &videoMode, GLFWmonitor *monitor)`
Create the fullscreen window on a specific monitor, setting resolution, refresh rate, and rgb bit depth. Monitor and video mode should be retrieved via GLFW at runtime.

We're using this method in our game to allow user resolution switching. We save their settings and retrieve the video mode and monitor at startup.

ricardoquesada:
+1 for supporting this feature (although I haven't read the patch).

Linux, Mac and Windows games are important for us, so +1 for adding features that are needed for this.
One feature that we need is a way to switch from fullscreen to window mode in runtime. I think GLFW 3.1 will support it.

cocos/2d/platform/desktop/CCGLView.h
```diff
@@ -39,6 +39,7 @@ class CC_DLL GLView : public GLViewProtocol, public Ref
     static GLView* create(const std::string& viewName);
     static GLView* createWithRect(const std::string& viewName, Rect size, float frameZoomFactor = 1.0f);
     static GLView* createWithFullScreen(const std::string& viewName);
+    static GLView* createWithFullScreen(const std::string& viewName, Size size);
```
dumganhar: It seems that you forgot to update `CCGLView.h`.

dumganhar: This `createWithFullScreen` has been removed, right?

wefiends: Yes, sorry. It's removed in the latest commit.

dumganhar: Thank you. I will check it some hours later. :)

dumganhar:
Thanks.

PR #6148, opened 2014-04-05
fix group collide bug

dabing1022:
In Chipmunk2D,  objects in the same non-zero group won't collide.

### DIFF
--- a/cocos/physics/CCPhysicsWorld.cpp
+++ b/cocos/physics/CCPhysicsWorld.cpp
@@ -275,7 +275,7 @@ int PhysicsWorld::collisionBeginCallback(PhysicsContact& contact)
     
     if (shapeA->getGroup() != 0 && shapeA->getGroup() == shapeB->getGroup())
     {
-        ret = shapeA->getGroup() > 0;
+        ret = false;
     }
     else
     {


cocos/physics/CCPhysicsWorld.cpp
```diff
@@ -275,7 +275,7 @@ int PhysicsWorld::collisionBeginCallback(PhysicsContact& contact)
     
     if (shapeA->getGroup() != 0 && shapeA->getGroup() == shapeB->getGroup())
     {
-        ret = shapeA->getGroup() > 0;
+        ret = false;
```
boyu0, line 278: Yes, But this value is designed the same as box2d, that's why we didn't use cpSetGroup to set this value.

dabing1022, line 278: It seems no sense. PhysicsWorld is constructed by Chipmunk2D essentially, we should keep the same mechanism as Chipmunk2D, not box2D.

发件人: "boyu0"; 
发送时间: 2014年4月6日(星期天) 上午9:05
收件人: "cocos2d/cocos2d-x"; 
抄送: "ChildhoodAndy"; 
主题: Re: [cocos2d-x] fix group collide bug (#6148)

In cocos/physics/CCPhysicsWorld.cpp:

boyu0, line 278: @dabing1022 
I don't think so, why does someone use our physics integration have to care about which physics engine they use? We only care about API friendly, performance and functionality, but not which physics engine they use, and which mechanism has to be.

boyu0, line 278: We use this mechanism is to make friendly with someone has SpriteKit experience, and this reason make more sense to me.

dabing1022:
Well, OK. Thanks for your explain, @boyu0 . We can close this issue now:)

PR #6150, opened 2014-04-05
fix Value's memory leak

jinyuliao:
Value's default constructor has memory leak

e.g:
Value leak;
Value something = std::move(leak);

variable something won't take the ownship of leak._vectorData, leak._mapData, leak._intKeyMapData because leak's type is Type::NONE


dumganhar:
Thanks. Merged.

PR #6151, opened 2014-04-05
add destroyInstance for ScriptHandlerMgr

jinyuliao:
prevent vld complain memory leak

### DIFF
--- a/cocos/scripting/lua-bindings/manual/LuaScriptHandlerMgr.cpp
+++ b/cocos/scripting/lua-bindings/manual/LuaScriptHandlerMgr.cpp
@@ -127,7 +127,6 @@ ScriptHandlerMgr::ScriptHandlerMgr()
 
 ScriptHandlerMgr::~ScriptHandlerMgr()
 {
-    CC_SAFE_DELETE(_scriptHandlerMgr);
 }
 
 ScriptHandlerMgr* ScriptHandlerMgr::getInstance()


dumganhar:
Thanks for your contribution.
And I don't know where should `ScriptHandlerMgr::destroyInstance` be invoked?
In the destructor of `AppDelegate`, or `Director::purgeDirector`?
IMO, it needs to be invoked at `Director::purgeDirector`. :)

jinyuliao:
now, i invoke the method in my demo, like this:
class CocosGC
{
public:
    ~CocosGC()
    {
        Configuration::destroyInstance();
        cocostudio::ObjectFactory::destroyInstance();
        ScriptHandlerMgr::destroyInstance();
        ScriptEngineManager::destroyInstance();
        PoolManager::destroyInstance();
    }
};
static CocosGC g_cocosgc;

invoke these method in Director::purgeDirector not always make sense
e.g:
1.in Director::purgeDirector, it will call Configuration::destroyInstance();
2.Director's destructor has been called, it will delete _renderer
3.in the Renderer's destructor it will call Configuration::getInstance(), Configuration now has been alloced again,then vld will give me the message about memory leak

dumganhar:
Correct my mistake:

> IMO, it needs to be invoked at Director::purgeDirector.

Yep, you're right, it should not be at Director::purgeDirector since it's lua relative codes.

dumganhar:
Merged. thanks.

PR #6152, opened 2014-04-05
Fixed a crash when tmx-map not have any tiles.

syrokomskyi:
- Fixed a crash when tmx-map not have any tiles.
- Fixed a batch-mode draw when sprite not have a textureAtlas.

### DIFF
--- a/cocos/2d/CCSpriteBatchNode.cpp
+++ b/cocos/2d/CCSpriteBatchNode.cpp
@@ -81,6 +81,7 @@ SpriteBatchNode* SpriteBatchNode::create(const std::string& fileImage, ssize_t c
 */
 bool SpriteBatchNode::initWithTexture(Texture2D *tex, ssize_t capacity)
 {
+    CCASSERT(tex, "Texture must be");
     CCASSERT(capacity>=0, "Capacity must be >= 0");
     
     _blendFunc = BlendFunc::ALPHA_PREMULTIPLIED;
@@ -356,7 +357,7 @@ void SpriteBatchNode::reorderBatch(bool reorder)
 void SpriteBatchNode::draw(Renderer *renderer, const kmMat4 &transform, bool transformUpdated)
 {
     // Optimization: Fast Dispatch
-    if( !_textureAtlas || (_textureAtlas->getTotalQuads() == 0) )
+    if( _textureAtlas->getTotalQuads() == 0 )
     {
         return;
     }
--- a/cocos/2d/CCTMXLayer.cpp
+++ b/cocos/2d/CCTMXLayer.cpp
@@ -62,7 +62,10 @@ bool TMXLayer::initWithTilesetInfo(TMXTilesetInfo *tilesetInfo, TMXLayerInfo *la
         texture = Director::getInstance()->getTextureCache()->addImage(tilesetInfo->_sourceImage.c_str());
     }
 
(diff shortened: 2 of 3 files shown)


cocos/2d/CCSpriteBatchNode.cpp
```diff
@@ -356,7 +356,7 @@ void SpriteBatchNode::reorderBatch(bool reorder)
 void SpriteBatchNode::draw(Renderer *renderer, const kmMat4 &transform, bool transformUpdated)
 {
     // Optimization: Fast Dispatch
-    if( _textureAtlas->getTotalQuads() == 0 )
+    if( !_textureAtlas || (_textureAtlas->getTotalQuads() == 0) )
```
dumganhar: Why could `_textureAtlas` be a `nullptr` ? It's initialized at `SpriteBatchNode::initWithTexture` .

syrokomskyi: Look at CCTMXLayer.cpp, line 65.

dumganhar: Got it.

dumganhar: Thanks. Currently, in this patch, if the texture is nullptr, you will not invoke `SpriteBatchNode::initWithTexture`.
But i think we should support `SpriteBatchNode::initWithTexture(nullptr, 10)` since all `initWith` will be protected methods. Developer will not be able to `re-initialize` it.  Invoking `SpriteBatchNode::setTexture` will cause some problems.

cocos/2d/CCTMXLayer.cpp
```diff
@@ -143,6 +146,8 @@ void TMXLayer::releaseMap()
 // TMXLayer - setup Tiles
 void TMXLayer::setupTiles()
 {    
+    if ( !_textureAtlas ) { return; }
+
     // Optimization: quick hack that sets the image size on the tileset
     _tileSet->_imageSize = _textureAtlas->getTexture()->getContentSizeInPixels();
```
boyu0, line 152: _textureAtlas is still created at `SpriteBatchNode::init()`, so in my machine it crashed at here, which `_tileSet` equal to `nullptr`.

syrokomskyi, line 152: _textureAtlas created at `SpriteBatchNode::initWithTexture()` and I control it by `CCASSERT(tex, "Texture must be")`. In the Release-mode `CCASSERT()` is disabled...

boyu0, line 152: But in SpriteBatchNode::init(), it will create a new texture and pass through it
`bool SpriteBatchNode::init()
  {
      Texture2D * texture = new Texture2D();
      texture->autorelease();
      return this->initWithTexture(texture, 0);
  }`
so tex won't be nullptr, and _textureAtlas is created, so `if ( !_textureAtlas ) { return; }` won't work, and the `_tileSet` is nullptr, it crashes because of the bad access.

syrokomskyi, line 152: Do change `if ( !_textureAtlas ) { return; }`
to `if ( !_tileSet || !_textureAtlas ) { return; }`
?

cocosQA:
Dear signmotion:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6421

PR #6157, opened 2014-04-07
Event dispatcher remove listener fix

dbaack:
Fixed a bug where an event listener removed from the event dispatcher while it is still in the _toAddedListeners vector would never have _isRegistered set to false on removal. This causes it to fail an assert if it is attempted to be added to the event dispatcher again later.


dumganhar:
Thanks.

PR #6159, opened 2014-04-07
dispatchPendingRunnables is not called from any where

matsuhiro:
In case of Java activity, dispatchPendingRunnables is not called from any where.
So, runOnGLThread is not work well.


newnon:
Yes looks line it help to solve problem with runOnGLThread

dumganhar:
We fixed this issue at https://github.com/cocos2d/cocos2d-x/pull/6182 in another way.
Please pull the latest codes to check whether it works. Thanks.

matsuhiro:
Thaks comments

PR #6160, opened 2014-04-07
CCEventDispatcher crash fix

BodbDearg:
Fix possible crashes which could be caused by the EventDispatcher having listeners associated with nodes are destroyed. Catch the case where a node registers a listener while we are dispatching events and gets destroyed while that event is still being dispatched. Check the list of nodes to be added into the event dispatcher fully when we are cleaning listeners for a target node.

This issue was found using the extra debug verification in this pull request:
https://github.com/cocos2d/cocos2d-x/pull/6011

### DIFF
--- a/tests/cpp-tests/Classes/NewEventDispatcherTest/NewEventDispatcherTest.cpp
+++ b/tests/cpp-tests/Classes/NewEventDispatcherTest/NewEventDispatcherTest.cpp
@@ -8,6 +8,7 @@
 
 #include "NewEventDispatcherTest.h"
 #include "testResource.h"
+#include "CCAutoreleasePool.h"
 
 namespace {
     
@@ -27,7 +28,8 @@ std::function<Layer*()> createFunctions[] =
     CL(PauseResumeTargetTest),
     CL(Issue4129),
     CL(Issue4160),
-    CL(DanglingNodePointersTest)
+    CL(DanglingNodePointersTest),
+    CL(RegisterAndUnregisterWhileEventHanldingTest)
 };
 
 unsigned int TEST_CASE_COUNT = sizeof(createFunctions) / sizeof(createFunctions[0]);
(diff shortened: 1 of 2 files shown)


BodbDearg:
Just added a test which can be used to reproduce the crashing issue and verify the fix.

dumganhar:
Thank you. I'm checking this issue. :)

cocos/2d/CCEventDispatcher.cpp
```diff
@@ -340,6 +340,26 @@ void EventDispatcher::removeEventListenersForTarget(Node* target, bool recursive
...
+    // This is to catch the scenario where the node gets destroyed before it's listener
+    // is added into the event dispatcher fully. This could happen if a node registers a listener
+    // and gets destroyed while we are dispatching an event (touch etc.)
+    for (auto iter = _toAddedListeners.begin(); iter != _toAddedListeners.end(); )
+    {
+        EventListener * listener = *iter;
+            
+        if (listener->getSceneGraphPriority() == target)
+        {
+            listener->release();
```
dumganhar, line 354: It's better to invoke

``` c++
listener->setRegistered(false);
```

before

``` c++
listener->release();
```

dumganhar, line 354: Since the listener may be reused after it was removed.

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/6162 . Thanks.:)

ricardoquesada:
A few days ago I decided to try cpp-test on Win32 and it crashes all the time on the `EventDispatcher`

When I saw this patch, I thought it was refering to the win32 bug but now... `cpp-test` is still crashing with `cpp-test`.

any idea? this bug is a show-stopper for v3.0.

How to reproduce it:
- run cpp-test on windows
- compile it using Visual Studio 2013 Update 2

cpp-empty-test works ok. But cpp-test crashes when I launch the app.

dumganhar:
@ricardoquesada , thanks for your feedback. I will check it ASAP. :)

cocos/2d/CCEventDispatcher.cpp
BodbDearg, line 354: Great point, thanks for catching this!

BodbDearg:
That's very concerning, thanks for letting us know! I think I'll check this out too.

dumganhar:
@ricardoquesada  @DarraghCoy 
I fixed it at https://github.com/cocos2d/cocos2d-x/pull/6169

**WHY DOES THE ISSUE ONLY EXIST IN cpp-tests on win32:**

cpp-tests registers `Mouse listener`, and if we launch application with mouse moving frequently, 
the mouse listener will receive mouse event before the first scene is ran.
So it will crash at `sortSceneGraphPriorityListeners` since `getRunningScene` will return nullptr.

Why don't other desktop platforms (like Mac or Linux) have this issue?

I think it's because different implementation in glfw.

ricardoquesada:
@dumganhar 
good finding! thanks.

BodbDearg:
Awesome, thanks for fixing up that issue!

dumganhar:
My pleasure. :)

PR #6161, opened 2014-04-08
Adding extra CCNode verification in Node's destructor

BodbDearg:
Add a check to the CCNode destructor to catch a common cocos2d-x programming mistake - forgetting to call base class onExit() in derived class onExit() implementations. We can check for this invalid API usage by checking if the node is still marked as running on destruction. If the node still thinks it's running then it's likely the cleanup logic performed by Node::onExit() was not done.


dumganhar:
Thanks.

BodbDearg:
You're welcome!

PR #6162, opened 2014-04-08
Fixed a potential crash in EventDispatcher, based on PR #6160

dumganhar:
This PR is based on #6160


cocos/2d/CCEventDispatcher.cpp
```diff
@@ -340,6 +340,27 @@ void EventDispatcher::removeEventListenersForTarget(Node* target, bool recursive
         }
     }
     
+    // Bug fix: ensure there are no references to the node in the list of listeners to be added.
+    // If we find any listeners associated with the destroyed node in this list then remove them.
+    // This is to catch the scenario where the node gets destroyed before it's listener
```
minggo, line 345: it's -> its

PR #6166, opened 2014-04-08
Develop fix renderer bug

dabingnn:


dabingnn:
profile result comparison:
old
screen shot 2014-04-08 at 3 07 33 pm
new
screen shot 2014-04-08 at 3 06 35 pm

minggo:
This pull request is to resolve [the issue](http://cocos2d-x.org/issues/4629).

@dabingnn 
Could you please describe why use recursive calls for group command.

ricardoquesada:
yes, I agree  with @minggo .
Could you describe what does the patch do ? But the new code seems to be cleaner than before.

Also, you should know that measuring the GPU (`drawBatchedNodes`) using a CPU profiler is not very reliable.

dabingnn:
The original code are mixing three jobs into one function `render()`.
1. traverse _renderGroups using manually stack.
2. do quad batching
3. do rendering.

In fact, the code has a [bug](http://cocos2d-x.org/issues/4629) which was found by accident.
Using recursive call instead of manually stack make code cleaner and less error-prone.

For performance, I tested both the fps for random selected about 25 samples for comparison.
The result is
screen shot 2014-04-08 at 4 44 03 pm
The above CPU profiling result are based on A9, 3000 sprites.
Though it is not very reliable, my opinion is that all these two data suggests there is no performance penalty if we use this new method.

dabingnn:
one supplement
if the cell has only one data, it is the old data.
if there are two data seperated by `|`, the latter is the new data.

minggo:
Great, i will merge it because the performance is good.

PR #6177, opened 2014-04-08
Small dangling ptr fix to CCEventDispatcher

BodbDearg:
If cleaning a target node that has a listener in the 'to be added' listeners list then ensure that listener does not have a stale reference to the node. This is a minor tweak in addition to pull request 6160 (https://github.com/cocos2d/cocos2d-x/pull/6160).


BodbDearg:
Something I forgot to add to PR 6160... (https://github.com/cocos2d/cocos2d-x/pull/6160)

dumganhar:
Thanks.

BodbDearg:
Thank you!

PR #6178, opened 2014-04-08
Skip a frame if there is no action updated or event dispatched

heliclei:
To reduce power consumption

### DIFF
--- a/cocos/2d/CCDirector.cpp
+++ b/cocos/2d/CCDirector.cpp
@@ -163,6 +163,7 @@ bool Director::init(void)
 #endif
     
     _dirty = true;
+    _idleFrames = 0;
     return true;
 }
 
@@ -278,6 +279,10 @@ void Director::drawScene()
     }
 
     if(!_dirty) 
+    {
+        _idleFrames++;
+    }
+    if(_idleFrames > 3)
     {
         return;
     }
@@ -1066,6 +1071,10 @@ void Director::setEventDispatcher(EventDispatcher* dispatcher)
 void Director::setDirty(bool dirty)
 {
     _dirty = dirty;
+    if(dirty == true)
+    {
+        _idleFrames = 0;
+    }
 }
 
 /***************************************************
--- a/cocos/2d/CCDirector.h
+++ b/cocos/2d/CCDirector.h
@@ -491,9 +491,10 @@ protected:
     /* Console for the director */
     Console *_console;
 #endif
-    /* skip draw frame if _dirty is false */
+    
     bool _dirty;
-
+    /* skip draw frame if idle frames is larger then 3 */
+    int _idleFrames;
     // GLViewProtocol will recreate stats labels to fit visible rect
     friend class GLViewProtocol;
 };


heliclei:
@ricardoquesada : I know this is a very premature and risky feature, but if we can do on-demand frame update eventually, I think it could reduce cocos2d-x engine's power consumption substantially.

thoechsmann:
Great feature. We implemented it for cocos2d-x v2.

ricardoquesada:
I like the idea, but I think the current patch is not covering all the possible updates. 
Perhaps I didn't understand the patch correctly, but are you covering the "updates" ?
Also, if an event is dispatched (no matter which event), then the directory should be dirty, so the code on the `EventDispatcher` could be simplified.

Anyway, I think reducing the power consumption is a great feature. What other things should we do in order to say that cocos2d is eco-friendly ? :)

As an example: How about networking ? can we improve our network library to be eco-friendly ?

heliclei:
Thanks! @ricardoquesada, I will run more tests to verify if it works on all cases.

heliclei:
@thoechsmann ,That's great, Can you join to improve this PR?

dumganhar:
@ricardoquesada  is right. This patch didn't cover all `updates`.
`Scheduler::update` also needs to set the dirty flag.

cocos/2d/CCDirector.cpp
```diff
@@ -274,6 +278,11 @@ void Director::drawScene()
         _eventDispatcher->dispatchEvent(_eventAfterUpdate);
```
heliclei, line 278: @dumganhar : if we set dirty bit in scheduler->update(), The dirty bit will be set every frame. So, We have to set it in scheduler target update() function, such as ActionManager, or other targets. This approach is more precise, but application code have to set the dirty bit if it has custom scheduler update(). Another choice is set dirty bit for all custom update(), it's less precise, but simpler.

thoechsmann:
We introduced a global variable that was set to true if a property of a CCNode changed.

You only seem to cover scheduled actions. What happens if you set e.g. the position of a sprite. It will not trigger a redraw.

heliclei:
@thoechsmann : In such case, Does Application code need set the dirty flag manually, or engine can detect the property change?

thoechsmann:
The Application should never be responsible for this. Otherwise there will be tons of problems. It must be handled inside of cocos. Maybe Ric has a better idea, but I would add code to CCNode. flag must be set to dirty if you add/remove a child or set properties like position, color, … In most (but not all) cases it can be set when _transformUpdated is set. 

Other things to keep in mind is that you cannot stop rendering after one frame when there are no more changes. This will not work on all android devices. We rendered always 3 frames after the last change. I think two frames should be enough, but we got problems with that. I can merge in our changes in V3 and make a pull request so you can take a look at the changes.

heliclei:
@thoechsmann: Thank you very much! I haven't test android device yet.

ricardoquesada:
yes, the best way to do it is in visit. and visit should return a boolean indicating whether or not the scene is dirty.

We are already passing the `dirty` flag as an argument. what we have to do it, is to return it.
We are planning to modify the API for `visit` in v3.1... so, we should introduce this change there.

heliclei:
That's great! @ricardoquesada :+1:

cocosQA:
Dear heliclei:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6418

PR #6180, opened 2014-04-08
add a switch to open/close the physics simulate

dabing1022:
add a switch to open/close the physics simulate

### DIFF
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -37,6 +37,7 @@ NS_CC_BEGIN
 Scene::Scene()
 #if CC_USE_PHYSICS
 : _physicsWorld(nullptr)
+, _physicsSimulateEnabled(true)
 #endif
 {
     _ignoreAnchorPointForPosition = true;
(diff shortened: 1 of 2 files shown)


dabing1022:
It seems better. Just do it :)PS: Notice that we disable the physics world, but the debugDraw can still exsits.
I mean 

```
if (_debugDrawMask != DEBUGDRAW_NONE)

{

    debugDraw();

}
```

These code will not be affected the physics simulation switch. What do you think?

发件人: "boyu0"; 
发送时间: 2014年4月9日(星期三) 下午4:18
收件人: "cocos2d/cocos2d-x"; 
抄送: "ChildhoodAndy"; 
主题: Re: [cocos2d-x] add a switch to open/close the physics simulate(#6180)

We already have a PhysicsWorld::setSpeed() API, set speed to zero can stop the simulation.
 I think add 
if(_speed == 0)
 {
     return;
 }
 to PhysicsWorld::update() is better(after dirty check).

boyu0:
@dabing1022 
Yes, you are right.
Just skip step and body update part.
Thanks :)

PR #6188, opened 2014-04-09
[FIX]: HTTPClient should not report 2xx status codes as errors!

Mazyod:
Current implementation of HttpClient reports HTTP responses with status code other than 200 as failures, but when consuming a RESTful API, status codes can range across the 2xx realm, and still be considered a success.

### DIFF
--- a/cocos/network/HttpClient.cpp
+++ b/cocos/network/HttpClient.cpp
@@ -339,7 +339,7 @@ public:
         if (CURLE_OK != curl_easy_perform(_curl))
             return false;
         CURLcode code = curl_easy_getinfo(_curl, CURLINFO_RESPONSE_CODE, responseCode);
-        if (code != CURLE_OK || *responseCode != 200) {
+        if (code != CURLE_OK || !(*responseCode >= 200 && *responseCode < 300)) {
             CCLOGERROR("Curl curl_easy_getinfo failed: %s", curl_easy_strerror(code));
             return false;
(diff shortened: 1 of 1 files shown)


dumganhar:
Thanks. Could you provide any references about that?

Mazyod:
@dumganhar Yeah, [here](http://en.wikipedia.org/wiki/HTTP_status_code#2xx_Success).

dumganhar:
Thanks. we will check that.  :+1:

dumganhar:
Merged. :)

Mazyod:
awesome :smile: I am just getting started with HttpClient, will try to improve it as much as possible. (e.g. Thinking about fixing the callbacks from target/selector, to c++11 `std::function`)

dumganhar:
Right now, we're using `std::function` instead of `target/selector` in develop branch.
Is there still `target/selector` stuff in HttpClient?

Mazyod:
Yeah, I created an [issue #4702](http://cocos2d-x.org/issues/4702) for that. I could've sworn I marked it as refactor, but it's somehow under bugs.

dumganhar:
I see, yep, it will be better if using `std::function`, thanks for point it out. :)

Mazyod:
Np :) I already fixed this on my branch, but didn't send a pull request because I need to update the test cases and docs. Hopefully, this weekend.

PR #6194, opened 2014-04-09
2026:72: Result of 'malloc' is incompatible with type 'unsigned char *'

iSevenDays:
2026:72: Result of 'malloc' is converted to a pointer of type 'unsigned char', which is incompatible with sizeof operand type 'unsigned char *'

### DIFF
--- a/cocos/2d/platform/CCImage.cpp
+++ b/cocos/2d/platform/CCImage.cpp
@@ -2023,7 +2023,7 @@ bool Image::saveImageToPNG(const std::string& filePath, bool isToRGB)
         {
             if (isToRGB)
             {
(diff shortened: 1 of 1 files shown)


dumganhar:
OH, shit, who makes this mistake.
Thanks for this patch. :)

dumganhar:
Thanks. Merged.

dumganhar:
How did you find this error?

iSevenDays:
You are welcome.
Xcode static analyzer. There are some possible errors still exists.

dumganhar:
Got it. Thank you. I will also use that tool to check our codes. :)

iSevenDays:
Please use it. There are some possible errors that are hard to reproduce so I fixed only obvious errors..

PR #6195, opened 2014-04-09
Incompatible pointer type?

iSevenDays:
Why sizeof(void*_) and not sizeof(void_)?


dumganhar:
Thanks. could you also send the patch to https://github.com/slembcke/Chipmunk2D ?

iSevenDays:
Yes, thanks, I will send the patch now.

iSevenDays:
The patch is already there.

dumganhar:
Thanks.

PR #6197, opened 2014-04-09
Potential leak of an object stored into "self"

iSevenDays:
1. By executing 314: _systemControl = [[CCEditBoxImplIOS_objc alloc] initWithFrame:rect editBox:this];
2. We entered to the function 67: -(id) initWithFrame: (CGRect) frameRect editBox: (void*) editBox
3. Then 69: "self = [super init];" returns object with +1 retain count
4. 93: "return nil" Potential leak of an object stored into "self".

### DIFF
--- a/extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
+++ b/extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
@@ -67,13 +67,11 @@ static const int CC_EDIT_BOX_PADDING = 5;
 -(id) initWithFrame: (CGRect) frameRect editBox: (void*) editBox
 {
     self = [super init];
-    
-    do
-    {
-        if (self == nil) break;
+    if(self){
         editState_ = NO;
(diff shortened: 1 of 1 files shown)


extensions/GUI/CCEditBox/CCEditBoxImplIOS.mm
```diff
@@ -67,13 +67,11 @@ - (void)dealloc
...
-    
-    do
-    {
-        if (self == nil) break;
+    if(self){
         editState_ = NO;
         self.textField = [[[CCCustomUITextField alloc] initWithFrame: frameRect] autorelease];
-        if (!textField_) break;
+        //TODO: what is the line below doing?
+        //if (!textField_) break;
```
dumganhar, line 74: It's used for checking whether the return value is valid.
Yep, I think it isn't needed. Feel free to remove it. :)

dumganhar:
Merged at https://github.com/cocos2d/cocos2d-x/pull/6206 , Thanks.

PR #6199, opened 2014-04-09
Label:Makes the create functions of Label clear.

WenhaiLin:


thoechsmann:
I vote against a fallback that creates a system font label when no font file was found. The fallback results in a Label, that has a totally different behaviour and it will not be easy to notice.

dumganhar:
@Dhilan007 
I agree with @thoechsmann . We should not change behavior if font is not found.

PR #6201, opened 2014-04-09
Fix memory leak

iSevenDays:
By executing this
void SimpleAudioEngine::preloadEffect(const char\* pszFilePath)
{
        soundData  *data = new soundData;
...
        file.file = fopen(fullPath.c_str(), "rb");
        if (!file.file) {
            fprintf(stderr, "Cannot read file: '%s'\n", fullPath.data());
            return;
        }
...
Dynamically allocated variable "data" is leaked if "!file.file".
And data is never freed.


dumganhar:
Thanks.

PR #6223, opened 2014-04-10
Optimization and get width and height

chengstory:
1. change map<std::string, TInfo> on unordered_map<unsigned int, TInfo> for speeding json parse.
2. fixes crash when reload texture with alpha.
3. change  reloadTexture(const char\* fileName) to reload Texture(const char\* fileName, unsigned int &imageWidth, unsigned int &imageHeight) for getting texture width and height.
4. add set CCDirector winsize just for cocostudio.

### DIFF
--- a/cocos2dx/textures/CCTextureCache.h
+++ b/cocos2dx/textures/CCTextureCache.h
@@ -140,7 +140,7 @@ public:
      * The "filenName" parameter is the related/absolute path of the file image.
      * Return true and image width and Height if the reloading is succeed, otherwise return false.
      */
-    bool reloadTexture(const char* fileName, unsigned int &imageWidth, unsigned int &imageHeight);
+    bool reloadTexture(const char* fileName, unsigned short &imageWidth, unsigned short &imageHeight);
 
     /** Purges the dictionary of loaded textures.
     * Call this method if you receive the "Memory Warning"
--- a/cocos2dx/textures/CCTextureCache.cpp
+++ b/cocos2dx/textures/CCTextureCache.cpp
@@ -592,7 +592,7 @@ CCTexture2D* CCTextureCache::addUIImage(CCImage *image, const char *key)
     return texture;
 }
 
-bool CCTextureCache::reloadTexture(const char* fileName, unsigned int &imageWidth, unsigned int &imageHeight)
+bool CCTextureCache::reloadTexture(const char* fileName, unsigned short &imageWidth, unsigned short &imageHeight)
 {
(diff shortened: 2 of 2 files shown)


extensions/CocoStudio/Trigger/ObjectFactory.h
```diff
@@ -58,8 +58,9 @@ class CC_EX_DLL ObjectFactory
         TInfo& operator= (const TInfo &t);
         std::string _class;
         Instance _fun;
+        unsigned int _classid;
     };
-    typedef std::map<std::string, TInfo>  FactoryMap;
+    typedef std::unordered_map<unsigned int, TInfo>  FactoryMap;
```
dumganhar: `master` branch doesn't use `c++11`, and `unordered_map` is a stuff in c++11.

cocos2dx/CCDirector.h
```diff
@@ -191,6 +191,11 @@ class CC_DLL CCDirector : public CCObject, public TypeInfo
     */
     CCSize getWinSize(void);
 
+    // window size
+    /** set the size of the OpenGL view in points.
+    */
+    void setWinSize(const CCSize& wsize);
```
dumganhar: Why do we need this setter?

extensions/CocoStudio/Trigger/ObjectFactory.h
```diff
@@ -28,7 +28,7 @@ THE SOFTWARE.
 #include "cocos2d.h"
 #include "ExtensionMacros.h"
 #include <string>
-#include <map>
+#include <unordered_map>
```
dumganhar: should revert this

cocos2dx/textures/CCTextureCache.h
```diff
@@ -138,9 +138,9 @@ class CC_DLL CCTextureCache : public CCObject
      * If the file image hasn't loaded before, load it.
      * Otherwise the texture will be reloaded from the file image.
      * The "filenName" parameter is the related/absolute path of the file image.
-     * Return true if the reloading is succeed, otherwise return false.
+     * Return true and image width and Height if the reloading is succeed, otherwise return false.
      */
-    bool reloadTexture(const char* fileName);
+    bool reloadTexture(const char* fileName, unsigned int &imageWidth, unsigned int &imageHeight);
```
dumganhar: I think the output parameter should use `pointer`.
So if user doesn't care about the `width` and `height`, just don't pass it.

``` c++
bool reloadTexture(const char* fileName, unsigned int* outImageWidth = NULL, unsigned int* outImageHeight = NULL);
```

This could also keep compatibility for the old usage.

PR #6227, opened 2014-04-10
add testEnd infomation in controller.cpp, and add send email in autotest...

shujunqiao:
add test end information in controller.cpp, and add send email in autotest.py.

### DIFF
--- a/tests/cpp-tests/Classes/controller.cpp
+++ b/tests/cpp-tests/Classes/controller.cpp
@@ -410,8 +410,8 @@ void TestController::addConsoleAutoTest()
                     //send end information.
                     if ( i == g_testCount-1 )
                     {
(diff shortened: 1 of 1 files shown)


tests/cpp-tests/Classes/controller.cpp
```diff
@@ -406,6 +406,13 @@ void TestController::addConsoleAutoTest()
                             break;
                         }
                     }
+
+                    //send end information.
+                    if ( i == g_testCount-1 )
+                    {
+                        std::string strEnd = "TestEnd" 
+                        send(fd, strEnd.c_str(), strlen(strEnd.c_str()), 0)
```
dumganhar: std::string has a method called `length()`, so plz use it instead of `strlen`

PR #6233, opened 2014-04-11
Fix CC_SAFE_DELETE to be more safe

iSevenDays:
CC_SAFE_DELETE is more secure now, isn't it?


dumganhar:
`delete` itself will take care of `null` pointer.

iSevenDays:
Yes, you are right, sorry.

PR #6240, opened 2014-04-11
Develop add renderer assert

dabingnn:


cocos/2d/renderer/CCRenderer.h
```diff
@@ -148,6 +148,8 @@ class Renderer
     // stats
     ssize_t _drawnBatches;
     ssize_t _drawnVertices;
+    //flags to indicate rendering
```
dumganhar: `flags to indicate` -> `the flag for indicating`.

PR #6243, opened 2014-04-11
closed #4762: fix some problems in cocosdension in wp8

squirrelChen:
1.fix stopalleffect lead to stop background music 
2.fix play effect may lead to memory leak
@stammen in fix 2, I think StopAllSoundEffects should call UnloadSoundEffect to free memory or may lead to memory leak when you dont UnloadSoundEffect by yourself.  Please review.


stammen:
This looks good. Please make a pull request to the master branch as well as it uses the same audio code.

squirrelChen:
the pr in master: https://github.com/cocos2d/cocos2d-x/pull/6246

PR #6247, opened 2014-04-11
check if the socket has been closed by peer, apply this to all platforms

heliclei:


dumganhar:
fixed at https://github.com/cocos2d/cocos2d-x/pull/6249

PR #6254, opened 2014-04-12
Memory leak fix for ClippingNode

dumganhar:


cocos/2d/CCClippingNode.cpp
```diff
@@ -76,7 +76,11 @@ ClippingNode::ClippingNode()
 
 ClippingNode::~ClippingNode()
 {
-    CC_SAFE_RELEASE(_stencil);
+    if (_stencil)
+    {
+        _stencil->stopAllActions();
```
dumganhar, line 81: If a stencil node is not added into scene graph, and it also runs an action, memory leak will appear since it isn't cleanup.

cocos/2d/CCClippingNode.cpp
```diff
@@ -291,9 +295,9 @@ Node* ClippingNode::getStencil() const
 
 void ClippingNode::setStencil(Node *stencil)
 {
+    CC_SAFE_RETAIN(stencil);
     CC_SAFE_RELEASE(_stencil);
     _stencil = stencil;
-    CC_SAFE_RETAIN(_stencil);
```
dumganhar, line 296: Just modify the order, safe for passing a pointer which is equal to itself.

PR #6257, opened 2014-04-12
GroupCommandManager should not be a singleton, it should be managed by Renderer.

dumganhar:


cocos/2d/renderer/CCGroupCommand.cpp
```diff
@@ -50,7 +36,7 @@ GroupCommandManager::GroupCommandManager()
 
 GroupCommandManager::~GroupCommandManager()
 {
-    CC_SAFE_RELEASE_NULL(s_instance);
```
dumganhar, line 53: If `delete` was invoked for GroupCommandManger, it will cause it is released in forever loop.
Please take care of the usage of `singleton`.

dumganhar:
@ricardoquesada 
Please review this PR. Thanks in advance. :)

cocos/2d/renderer/CCGroupCommand.cpp
minggo, line 53: Yep, it is a bug here.

minggo:
I agree with @dumganhar , we should use as less singleton as possible.

ricardoquesada:
agreed, the proposed API is better.

PR #6258, opened 2014-04-12
'BaseData helpParentNode' should not be a global variable, define it in the relevant functions instead.

dumganhar:


minggo:
I think it is better to define it as a static member variable.

dumganhar:
Every time the `helpParentNode` will copy contents from its `parentNode`.
Also BaseData is inherited from `Ref`, all `Ref`'s subclasses should be created by `new`.
But it just define it here, it will be released only when application exits.
I used a memory tracker to find out memory issues in our engine.
When Appdelegate::~AppDelegate() was invoked , it would print all leak `Ref` objects. 
Since `BaseData` is global variable, and it will be released after the destructor of `AppDelegate`, so my memory `tracker` would print that `there is a leak for BaseData`.
Yep, to define it as `static variable` is ok. But is it reasonable to do that?

minggo:
It depends how frequently these functions invoked.

2youyou2:
I think this pr is ok since these functions did not invoke frequently

cocosQA:
Dear dumganhar:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6414

PR #6264, opened 2014-04-12
Fix loosing precision

iSevenDays:
I had a warning on building for iPhone 64 bit.
As you can see above variable "rc" has a type of size_t too( void Console::commandUpload(int fd) );


dumganhar:
Thank you. It's because on windows `recv` returns `int` while on other unix platforms, it returns `size_t`.

dumganhar:
I will fix the issue I mentioned above on windows.

dumganhar:
Merged. Thanks.

PR #6267, opened 2014-04-14
Add interface setCullingEnabled(), culling is disabled by default

dabingnn:


cocos/2d/CCSprite.cpp
```diff
@@ -621,9 +622,15 @@ void Sprite::drawDebugData()
...
 
+void Sprite::setCullingEnabled(bool enabled)
+{
+    _cullingEnabled = enabled;
+}
+
 // Culling function from cocos2d-iphone CCSprite.m file
 bool Sprite::isInsideBounds() const
 {
+    if(!_cullingEnabled) return true;
```
minggo, line 633: It is better to add a space line bellow.

dabingnn, line 633: added
Thanks

cocos/2d/CCSprite.h
```diff
@@ -426,6 +426,8 @@ class CC_DLL Sprite : public Node, public TextureProtocol
     virtual void setOpacityModifyRGB(bool modify) override;
     virtual bool isOpacityModifyRGB(void) const override;
     /// @}
+     //set culling enabled
+    void setCullingEnabled(bool enabled);
```
minggo, line 430: Please describe why default value is false. What's the limitation of current culling.

ricardoquesada:
why was this PR merged?

disabling auto-culling is not good, specially since we did many performance tests and we promoted the how fast v3.0 is.

we have to understand why auto-culling is not working for this case, but disabling it without further discussion is not a good idea.

ricardoquesada:
BTW, usually I need 2 to 3 days to review and understand pull request. If the PR is urgent, please say so, I'll try to review it in the same day.

ricardoquesada:
fixed here:
https://github.com/cocos2d/cocos2d-x/pull/6281

minggo:
Sorry about that, @dabingnn said he had discussed it with you in email.

dabingnn:
@ricardoquesada 
Sorry for my misunderstanding of your opinion.

ricardoquesada:
Ok. no problem.

PR #6281, opened 2014-04-14
Culling working again

ricardoquesada:
Culling working again.
Code moved to Renderer

### DIFF
--- a/cocos/2d/renderer/CCRenderer.cpp
+++ b/cocos/2d/renderer/CCRenderer.cpp
@@ -518,8 +518,8 @@ bool Renderer::checkVisibility(const kmMat4 &transform, const Size &size)
     float hSizeX = size.width/2;
     float hSizeY = size.height/2;
 
-    kmVec4 v4world;
-    kmVec4 v4local = (kmVec4) {hSizeX, hSizeY, 0, 1};
+    kmVec4 v4world, v4local;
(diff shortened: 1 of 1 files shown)


cocos/2d/renderer/CCRenderer.cpp
```diff
@@ -497,4 +506,36 @@ void Renderer::flush()
...
+    // half size of the screen
+    Size screen_half = Director::getInstance()->getWinSize();
+    screen_half.width /= 2;
+    screen_half.height /= 2;
+
+    float hSizeX = size.width/2;
+    float hSizeY = size.height/2;
+
+    kmVec4 v4world;
+    kmVec4 v4local = (kmVec4) {hSizeX, hSizeY, 0, 1};
```
dumganhar: This line will have problems on Win32.The error is

```
2>renderer\CCRenderer.cpp(522): error C2059: syntax error : '{'
2>renderer\CCRenderer.cpp(522): error C2143: syntax error : missing ';' before '{'
2>renderer\CCRenderer.cpp(522): error C2143: syntax error : missing ';' before '}'
```

Since windows doesn't support `initialization list`, it will cause this error.
To fix it,  use `kmVec4Fill` to set the value instead.

``` c++
kmVec4 v4local;
kmVec4Fill(&v4local, hSizeX, hSizeY, 0, 1);
```

ricardoquesada:
builder is green.
I tested it and it works OK... so, I'll merge it.

PR #6285, opened 2014-04-15
closed #4607: Fixed project created by template crashes on iOS5.1 device...

dumganhar:
....


kicktheken:
this needs to be fixed in the lua template as well

PR #6298, opened 2014-04-16
RenderTexture output ‘stretched’ if render texture size is not same as screen size.

sachingarg05:
Fix for bug where RenderTexture output is ‘stretched’ if render texture size is not same as screen size. 

It seems that _fullRect and _fullviewPort are switched in the CCRenderTexture.cpp. Discussed in more detail here: http://www.cocos2d-x.org/forums/6/topics/45860

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -194,9 +194,7 @@ bool RenderTexture::initWithWidthAndHeight(int w, int h, Texture2D::PixelFormat 
     void *data = nullptr;
     do 
     {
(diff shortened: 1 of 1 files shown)


sachingarg05:
Purpose of RenderTexture is to render to a texture of any size. It is wrong to assume that RenderTexture will always render to a viewport of screen size.

sachingarg05:
@dabingnn 

I have created a new pull request for this bug. Instead of changing behavior of current init functions, it adds a new init function which takes height/width parameters in pixels instead of points.

https://github.com/cocos2d/cocos2d-x/pull/6374

sachingarg05:
Closing this as the other pull request #6374 seems a better fix.

PR #6306, opened 2014-04-17
Xcodeproj bugfix

yuumi3:
Fixed compile error on Mac OS which has Case-Sensitive volume.


minggo:
I think you should send pull request to `master` branch.
Right?

yuumi3:
Yes.

PR #6318, opened 2014-04-17
Fix in Renderer to support PingPong (multi step) RenderTexture usage.

sachingarg05:
The new renderer implementation with multiple queues gets VERY confused when trying to use two RenderTextures to do multi-step processing using shaders (PingPong RenderTextures).

The problem description is big, but the code changes needed to fix are simple and 'safe' :-)

Solution description:
1) Renderer::visitRenderQueue is now public (was protected).
2) Renderer::visitRenderQueue takes int renderQueueID as parameter (parameter was RenderQueue& queue).
3) RenderTexture::end() now calls Renderer::visitRenderQueue to execute its own queue.

Problem description:

Lets say we want to execute 4 shaders one by one in this order, output of shader1 is input of second2, and so on:
  Shader1, Shader2, Shader3, Shader4.

When trying to do multipass shader processing using two RenderTexture objects (known as PingPong processing), the expected order of execution is:

Correct Order:
RenderTexture1 > Shader1 > RenderTexture2 (Queue1)
RenderTexture2 > Shader2 > RenderTexture1 (Queue2)
RenderTexture1 > Shader3 > RenderTexture2 (Queue1)
RenderTexture2 > Shader4 > RenderTexture1 (Queue2)
... and so on.

The new cocos2dx renderer creates a queue for each RenderTexture, commands are pushed to the queues, and then queues are executed one by one. 

There are two problems here:
1) When same RenderTexture object is used twice, the commands go in same queue. But the CommandGroup is pushed into renderer twice.
2) Order of execution is different from expected order.

Result is this order of execution which is wrong: (The 'Repeated' commands are executed twice because commandGroup object was pushed each time, this is wrong)

Queue1 (wrong order):
RenderTexture1 > Shader1 > RenderTexture2
RenderTexture1 > Shader3 > RenderTexture2
RenderTexture1 > Shader1 > RenderTexture2 (Repeated)
RenderTexture1 > Shader3 > RenderTexture2 (Repeated)

Queue2 (wrong order):
RenderTexture2 > Shader2 > RenderTexture1 
RenderTexture2 > Shader4 > RenderTexture1 
RenderTexture2 > Shader2 > RenderTexture1 (Repeated)
RenderTexture2 > Shader4 > RenderTexture1 (Repeated)

Solution:
RenderTexture should execute its queue in RenderTexture::end().

Reasons:
1) Each RenderTexture step is independent of other render commands and its queue can be executed independently without any performance loss. 
2) Each RenderTexture step must finish executing before second RenderTexture step can be executed. We can't wait till the end.
3) After each RenderTexture step, its queue must be emptied so that queue can be used in next RenderTexture step.

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -722,6 +722,7 @@ void RenderTexture::end()
     Renderer *renderer = Director::getInstance()->getRenderer();
     renderer->addCommand(&_endCommand);
     renderer->popGroup();
+    renderer->visitRenderQueue(_groupCommand.getRenderQueueID());
     
     kmGLMatrixMode(KM_GL_PROJECTION);
     kmGLPopMatrix();
--- a/cocos/2d/renderer/CCRenderer.cpp
+++ b/cocos/2d/renderer/CCRenderer.cpp
@@ -274,8 +274,9 @@ int Renderer::createRenderQueue()
     return (int)_renderGroups.size() - 1;
 }
 
-void Renderer::visitRenderQueue(const RenderQueue& queue)
+void Renderer::visitRenderQueue(int renderQueueID)
 {
+    RenderQueue& queue = _renderGroups[renderQueueID];
     ssize_t size = queue.size();
     
     for (ssize_t index = 0; index < size; ++index)
@@ -306,7 +307,7 @@ void Renderer::visitRenderQueue(const RenderQueue& queue)
         {
             flush();
             int renderQueueID = ((GroupCommand*) command)->getRenderQueueID();
-            visitRenderQueue(_renderGroups[renderQueueID]);
+            visitRenderQueue(renderQueueID);
         }
         else if(RenderCommand::Type::CUSTOM_COMMAND == commandType)
         {
@@ -325,6 +326,8 @@ void Renderer::visitRenderQueue(const RenderQueue& queue)
             CCLOGERROR("Unknown commands in renderQueue");
         }
     }
+    flush();
+    queue.clear();
 }
 
 void Renderer::render()
@@ -346,7 +349,7 @@ void Renderer::render()
         {
             renderqueue.sort();
         }
-        visitRenderQueue(_renderGroups[0]);
+        visitRenderQueue(0);
         flush();
     }
     clean();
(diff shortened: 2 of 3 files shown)


dabingnn:
@ricardoquesada 
Please give a review for this pull request.

dabingnn:
@sachingarg05 @ricardoquesada 
If I am right, you have reported two problems:
1. RenderTexture reuse: as you pointed out, if we reuse two rendertexture for pingpong, the behavior is uncorrect. The final reason for this is that we use a member variable. If we use a rendercommand pool, when user call RenderTexture::begin() and RenderTexture::end(),  new command are generated from pool, this problem will disappear.
2. Lack of instant rendering for RenderTexture: call visitRenderQueue() in RenderTexture::end().

For the first question, my suggestion is to test the performance for RenderCommandPool, if it has no performance penalty, we can add it.

For the second question, this pull request can fix it. However, my concern is that if we add multiple threading support, can we find a solution for this? Currently, I have no idea about this.

sachingarg05:
Yes, currently problem is that when we have two RenderTexture for pingpong, and then use same RenderTexture second time. 

1) The commands go into same member variable _groupCommand which sends commands to same render queue. This changes order of commands execution. AND 2) The _groupCommand is pushed again on queue 0. This also changes order of execution, and also executes each command twice (or 3 times if same RenderTexture is used 3 times in PingPong).

I used instant rendering to fix BOTH problems. 
1) Fix ordering to Ping1 Pong1 Ping2 Pong2 instead of: Queue1 (Ping1 Ping1) Queue2 (Pong2 Pong2) 
2) Remove the queued commands to avoid double execution: Queue1(Ping1Ping2 Ping1Ping2) Queue2 (Pong1Pong2 Pong1Pong2).

We don't need instant rendering if RenderCommandPool will make sure order of command executing in both RenderTexture will remain correct, and if it will fix commandGroup getting pushed twice on queue. I don't know how RenderCommandPool works.

sachingarg05:
I have updated my comment above to make it more clear. If there is confusion, we can also talk/chat on Skype or Google Hangout. My ID on both is sachingarg@rawzor.com

cocosQA:
Dear sachingarg05:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6410

PR #6322, opened 2014-04-18
closed #4904: delete ios Image::saveToImage method and combine it to ...

boyu0:
CCImage file.
Already tested.


boyu0:
Crashes in ios i386 simulator, still can't find a solution.

PR #6325, opened 2014-04-19
Wp8 3.0 update

stammen:
This pull request contains the following updates to WP8
1. Background texture is now correctly oriented in Transition Effects
2. Accelerometer now works for WP8. Refactor of InputEvents was required to support adding Accelerometer event to rendering queue.
3. Console support for WP8.
4. Fixed cpp template for WP8. Added missing projects. Fixed property sheets.
5. Updates precompiled shaders


dumganhar:
Thanks.

PR #6344, opened 2014-04-22
closed #4850: Add png pre multi alpha back

boyu0:


zifter:
For this Pull Request, I would like for one and get an answer to my question from developer:
http://www.cocos2d-x.org/forums/6/topics/49705
And explain why this fix is very necessary:
http://www.cocos2d-x.org/forums/6/topics/47767

PR #6363, opened 2014-04-24
Texture async loading leak fix

BodbDearg:
Fix a texture memory leak which occurred if textures which were loading asynchronously in the background were also loaded in the foreground (render/game thread) before the asynchronous loading had finished. The issue was that once asynchronous loading had finished, we weren't checking to see if the texture had already been loaded on a different thread. Fix this issue by checking for this situation and add a test to reproduce the old issue and test the fix.


cocosQA:
Dear DarraghCoy:
  We have created a new branch 'v3' to replace branch 'develop', and this PR has been moved from 'develop' branch to 'v3' branch, the new PR is https://github.com/cocos2d/cocos2d-x/pull/6406

PR #6371, opened 2014-04-25
Develop migrate math lib

dabingnn:


minggo:
I think it is better to rename `Vector2` to `vec2`. Because we already have `Vector`. But there are different meanings.

If we use `Vector2`, then it will be confused.

@ricardoquesada 
What's your opinion?

dumganhar:
Since they're Class, I remember that we need to rename them to 
`Vec2`, `Mat4`

ricardoquesada:
Hi, I already gave my opinion about this on this thread:
https://groups.google.com/d/msg/cocos2d-js-devel/BWDi7oHEtck/5PH04BnZhBwJ

Basically `Vec2`, `Vec3`, etc... are Ok.. but I prefer the longer names a bit better

ricardoquesada:
@dabingnn It can't be merged now, could you update the PR?

thanks.

ricardoquesada:
Apparently it fails for the Windows project. Not a major issues.

minggo:
Oh.
I planned to merge it after resolving compiling error. Or it will block other pull request.
Sorry about not mentioned it here.

PR #6372, opened 2014-04-25
Use correct return codes from mac CCApplication

twhittock:
The main.cpp file in the template returns these values directly to the os, and the application is currently marking itself as failing when it succeeds, and succeeding when it fails.


dumganhar:
Thanks.

PR #6376, opened 2014-04-25
Lua template should fail to launch on lua error

twhittock:
Without this change, the app continues to run, attempts to access the
badly-initialised scene and issues more error messages. This causes
further errors every frame, covering up the original error in the output


twhittock:
The CI failure appears to be a Jenkins issue rather than a code issue.

twhittock:
something really went wrong with this. closing and retrying.

PR #6381, opened 2014-04-27
Fix search path for lua scripts in subdirectories

kicktheken:
To demonstrate, see below sample main.lua code:

```
cc.FileUtils:getInstance():addSearchResolutionsOrder("src")
require("lib.script")
```

Before, the code would search for "lib/src/script.lua" instead of the correct path "src/lib/script.lua". This pull request fixes this issue.

### DIFF
--- a/cocos/2d/platform/CCFileUtils.cpp
+++ b/cocos/2d/platform/CCFileUtils.cpp
@@ -666,10 +666,10 @@ std::string FileUtils::getPathForFilename(const std::string& filename, const std
         file = filename.substr(pos+1);
     }
     
-    // searchPath + file_path + resourceDirectory
+    // searchPath + resourceDirectory + file_path
     std::string path = searchPath;
-    path += file_path;
     path += resolutionDirectory;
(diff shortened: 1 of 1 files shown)


dumganhar:
I think you have miss-understanding `search path`  and `resolution order`.
The full path consists of `search path`, `relative path`, `resolution order`, `filename`.
For example, you have a file at `hd/Images/order1/a.png`,
Your code: 

``` c++
auto sprite = Sprite::create("Images/a.png");
```

To make this work, you have to set `search path` to `hd`,
`resolution order` to  `order1`

So to achieve your destination, you could just set a search path of `src` rather than `resolution order`.
Thanks.

kicktheken:
Ok, I understand now. I should use:

```
cc.FileUtils:getInstance():addSearchPath("src")
```

instead of:

```
cc.FileUtils:getInstance():addSearchResolutionsOrder("src")
```

The lua template project should use `addSearchPath("src")` instead of `addSearchResolutionsOrder("src")` then.

PR #6382, opened 2014-04-27
Fix studioactionobject

xhcnb:
two bugs.


dumganhar:
Thanks. We'll check this issue.

PR #6387, opened 2014-04-28
issue #4923: Adds UVMoveAnimation and relevant test case

dumganhar:


cocos/2d/CCActionInterval.h
```diff
@@ -1080,6 +1080,33 @@ class CC_DLL TargetedAction : public ActionInterval
...
+CC_CONSTRUCTOR_ACCESS:
+    UVMoveAnimation() {}
+    virtual ~UVMoveAnimation() {}
+
+    /** initializes the action */
+    bool initWithDuration(float duration, const Vector2& UVOffset);
+
+protected:
+    Vector2 _UVOffset;
+    Vector2 _UVOffset_Last;
```
minggo, line 1105: _UVOffset_Last -> _UVOffsetLast

cocos/2d/CCSprite.h
```diff
@@ -225,6 +225,13 @@ class CC_DLL Sprite : public Node, public TextureProtocol
      */
     virtual void setVertexRect(const Rect& rect);
 
+    /*
+     * Sets the texture Coords Offset.
+     * Useful if you want to Create UV Animation.
+     */
+    virtual void setTextureUOffset(float u);
+    virtual void setTextureVOffset(float v);
```
minggo, line 233: Please add comment to describe that, `texture atlas` is not supported.

cocos/2d/CCSprite.cpp
```diff
@@ -411,6 +411,28 @@ void Sprite::setVertexRect(const Rect& rect)
...
+void Sprite::setTextureVOffset(float v)
+{
+    if (nullptr == _textureAtlas)
+    {
+        _quad.bl.texCoords.v += v;
+        _quad.br.texCoords.v += v;
+        _quad.tl.texCoords.v += v;
+        _quad.tr.texCoords.v += v;
+    }
+}
```
minggo, line 434: Add assert is better to skip it.

minggo:
As discussed with @dabingnn, we think it is not a good idea to change Sprite::_quad directly. And we think it is better
- add Sprite::setUVTransform()
- quad command will include the uv information
- renderer will use the uv information to change UV value

Will modify it in other pull request.

PR #6479, opened 2014-04-28
[ci skip]Fix NSNotificationCenter and NSTimer leaks.

cocosQA:
- Remove keyboard notifications from NSNotificationCenter.
- Invalidate NSTimer instance of the main loop.

Because, I'm creating a hybrid(UIWebView and Cocos2d-x) iOS application. So, we should fix the memory leaks.


minggo:
@uqtimes 
Thanks.
I send #6918 to merge your modification.

bugttle:
Thanks for the merge. I'm sorry I'm late.

PR #6498, opened 2014-04-28
[ci skip]Support for play sound effect with pitch , pan, gain

cocosQA:
Port code from object c to  play sound effect with pitch, pan, gain


minggo:
It is already supported.

PR #6500, opened 2014-04-28
[ci skip]Memcpy change submit to develop branch

cocosQA:
Update the usage of call of CCTextureAtlas.updateQuad in CCLabelAtlas.updateAtlasValues and in CCTileMapAtlas::updateAtlasValueAt.

Update the reference directly instead of calling the updateQuad method.


minggo:
@liuyq
Thanks.

liuyq:
Hi, minggo

Sorry, since I have not used git hub for a long time, could you please help to describe the instructions for "re-send the pull request", and because I am not in ThunderSoft now, I do not have the related source locally. I am not sure if this is still OK for your above request, just for your information.

Thanks,
Yongqin Liu

minggo:
Ok.
Thanks for the information.
I think we will try to fix it by ourself.

Thanks.

PR #6501, opened 2014-04-28
[ci skip][FIX] CCSprite and CCSpriteFrameCache work with trimmed textures

cocosQA:
If you have sprite sheet and use trim option for textures. Then they will draw incorrectly (with parts of another textures).
This should fix it.


minggo:
@nmakoto
Thanks.

PR #6506, opened 2014-04-28
Update Android.mk

Alex-G:
With this change you don't need add new cpp files manually if they are in the root of the Classes folder.


minggo:
@Alex-G 
Thanks.

PR #6509, opened 2014-04-29
Add reload plist png

chengstory:


cocos2dx/CCDirector.h
```diff
@@ -186,11 +186,14 @@ class CC_DLL CCDirector : public CCObject, public TypeInfo
...
 
     // window size
-
     /** returns the size of the OpenGL view in points.
     */
     CCSize getWinSize(void);
 
+     /** set the window size
+     */
+    void setWinSize(CCSize size);
```
minggo: On mobile platform, we can not set window size.

cocos2dx/textures/CCTextureCache.h
```diff
@@ -138,9 +138,9 @@ class CC_DLL CCTextureCache : public CCObject
      * If the file image hasn't loaded before, load it.
      * Otherwise the texture will be reloaded from the file image.
      * The "filenName" parameter is the related/absolute path of the file image.
-     * Return true if the reloading is succeed, otherwise return false.
+     * Return true and image width and Height if the reloading is succeed, otherwise return false.
      */
-    bool reloadTexture(const char* fileName);
+    bool reloadTexture(const char* fileName, unsigned short &imageWidth, unsigned short &imageHeight);
```
minggo, line 143: You can get the width and height information by getting the texture. Don't return the value in this function.

PR #6515, opened 2014-04-29
Skip UTF8 BOM if it exists before parsing by rapidjson

hanjukim:


dumganhar:
If BOM was not skipped, what problems will it trigger?

hanjukim:
if BOM was not skipped, json.HasParseError returns true because

```
        switch (stream.Peek()) {
            case '{': ParseObject<parseFlags>(stream, handler); break;
            case '[': ParseArray<parseFlags>(stream, handler); break;
            default: RAPIDJSON_PARSE_ERROR("Expect either an object or array at root", stream.Tell());
        }
```

Above switch case goes to default because of rapidjson can't skip bom (rapidjson supports UTF8 with BOM on their trunk repository though, but it needs extra work. So I decided just skip BOM and pass it to rapidjson. see: https://code.google.com/p/rapidjson/)

hanjukim:
019f3f5 also needs to fixed because error will occur from rapidjson with not properly terminated string

dumganhar:
@hanjukim ,thank you. We will check it soon.

minggo:
@hanjukim
Thanks.

hanjukim:
Ok, I will do it soon.

PR #6516, opened 2014-04-29
Wp8 2.2 update

stammen:
This pull requests adds the following to cocos2d-x 2,2 on WP8 and WinRT

1, Merge code with cocos2d-x v2 branch.
2. Precompiled headers are enabled for much faster builds
3. Angle libs are now precompiled. Git submodule also updated to point to latest angle code.
4. CURL and libWebSockets are enabled for WP8. Prebuilt libs for curl and libwebsocket were added
5. WP8-XAML uses DrawingSurfaceBackgroundGrid for better performance and to fix Device memory leak with DrawingSurface.
6. Removed debug prebuilt libs from winrt and wp8.
7. Added purgeSharedSpriteFrameCache() to purge in CCDirector to fix dangling refcounts on textures.

This pull requests adds the following to cocos2d-x 2,2 on Win32

TestCpp libExtensions and libCocos2d-x now have precompiled headers enabled which greatly speeds up the build,


minggo:
[ci rebuild]

stammen:
The win32 build is broken on the first build pass in V2 branch without any of my changes.Building again causes AssestManager to build and then the build is complete. Whoever is responsible for Win32 version needs to look at this problem.

minggo:
Thanks @stammen, we have checked that, it is the problem of ci.

PR #6517, opened 2014-04-30
Improved directory structure

ricardoquesada:
Improved directory structure
Includes missing license files


tests/lua-tests/project/proj.android/project.properties
```diff
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-10
+target=Amazon.com:Amazon Fire TV SDK Addon:17
```
dumganhar: The target version was changed to Amazon TV, it should be standard `android-10`.

tests/lua-empty-test/project/proj.android/project.properties
```diff
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-10
+target=Amazon.com:Amazon Fire TV SDK Addon:17
```
dumganhar: The target version was changed to Amazon TV, it should be standard android-10.

tests/lua-tests/project/proj.android/project.properties
ricardoquesada: thanks. I'll fix it :)

ricardoquesada: BTW, I think we need to add `project.properties` into `.gitignore`
I think a lot of PR includes modified `project.properties` and those PR are rejected because of that (which makes sense).
So, in order to help our developers to create PR, I think it is safe to add the `project.properties` into `.gitignore`.

thoughts?

minggo: Agreed.

minggo:
@ricardoquesada 
Do you need help for other platforms to handle building problem?

ricardoquesada:
Thanks minggo.
I think I will be able to fix them by myself. If I can't I will let you know.

minggo:
Ok.

PR #6535, opened 2014-05-01
closed #5099: Optimization for Value

dumganhar:


cocos/base/CCValue.h
```diff
@@ -133,12 +134,12 @@ class Value
...
-    std::string _strData;
-    ValueVector* _vectorData;
-    ValueMap* _mapData;
-    ValueMapIntKey* _intKeyMapData;
+
+        std::string* strVal;
+        ValueVector* vectorVal;
+        ValueMap* mapVal;
+        ValueMapIntKey* intKeyMapVal;
+    }_field;
```
dumganhar, line 142: All types were defined in `_field union` now. It will save some memory.

dumganhar:
@minggo , Please review this optimization.

minggo:
It seems good to me.

PR #6536, opened 2014-05-01
class Ref   retain&release is not ThreadSafe

YukiMiyatake:
use lock free


minggo:
Yep.
All engine codes are not thread safe.
We don't plan to make it thread safe.
All it will bring performance issue.

PR #6537, opened 2014-05-01
Add cocos/audio/include to the include directories of lua-binding

ZhangJingwang:
The commit 56f737db1ec934aba8be20493c66779150217b83 broke the build, due
to a missing header file.


ZhangJingwang:
Good. You guys have fixed this problem.

PR #6545, opened 2014-05-02
Physics rollback

ricardoquesada:
Physics rollback


nickaversano:
Why was this rolled back? The physics world cannot be at the scene level as per: https://github.com/cocos2d/cocos2d-x/pull/5493

When a game layer starts to resize, the physics world remains fixed on the scene. How else can a dynamic zooming effect be achieved on a physics world?

PR #6551, opened 2014-05-02
*fix when get ccnode, return a string

x44yz:
if add a CCNode in cocosbuilder, set type: "Doc root var",  i cant get this CCNode, it's a string


minggo:
@jojo2012 
Thanks.

PR #6555, opened 2014-05-03
issue #4859:Add a widget to play video

WenhaiLin:


ricardoquesada:
Thanks. nice addition.
Which platforms does it support? Only iOS and Android ?
If it doesn't support all the cocos2d-x supported platforms (iOS, Android, Windows, Mac, Linux and Windows Phone 8), and I wouldn't put this feature in the "core"... but perhaps in "extensions" or in "experimental", or under a new namespace. eg:

`cocos2d::experimental::ui::VideoWidget`

minggo:
If we put it in other place, then it is hard to find and use this feature.
We can merge it then implement it in other platforms later.

minggo:
It is great.
Thanks.

I have some suggestions:
- move the logical codes of playing a video out of `Cocos2dxActivity.java`, and put it into a new java file, such as `Cocos2dxVideoPlayer.java`
- move the logical codes of `UIVideoViewWrapperIos.h` into `UIVideoViewWrapperIos.mm`, since it is only used in `UIVideoViewWrapperIos.mm`
- move the logical codes of `UIVideoViewWrapperAndroid.cpp` into `UIVideoWidgetAndroid.cpp`
- correct licence contents

dumganhar:
Why do we need `Cocos2dx` prefix for our java file? Java's package management could resolve conflicts, right?

cocos/ui/UIVideoWidgetAndroid.cpp
```diff
@@ -0,0 +1,349 @@
...
+        _keepAspectRatioEnabled = enable;
+        setVideoKeepRatioEnabled(_videoWidgetIndex,enable);
+    }
+}
+
+#if CC_VIDEOWIDGET_DEBUG_DRAW
+void VideoWidget::drawDebugData(const Matrix& transform, bool transformUpdated)
+{
+    kmGLPushMatrix();
+    kmGLLoadMatrix(&transform);
```
dumganhar: I think kazmath was deprecated from now. We should use new math function for doing this.

PR #6556, opened 2014-05-04
add monkey-test in emptytest.py.[ci skip]

shujunqiao:
add monkey-test in emptytest.py.[ci skip]


shujunqiao:
not need merge now.

PR #6557, opened 2014-05-04
[ci skip], not add close menuItem in test empty-test.

shujunqiao:
[ci skip], not add close menuItem in test empty-test.


shujunqiao:
not need merge now.

PR #6575, opened 2014-05-05
Lua template should fail to launch on lua error

twhittock:
Without this change, the app continues to run, attempts to access the
badly-initialised scene and issues more error messages. This causes
further errors every frame, covering up the original error in the output


minggo:
Thanks.

PR #6582, opened 2014-05-06
math fixes

ricardoquesada:
- Moves deprecated Point to CCDeprecated.h
- Adds Point2 and Point3
- removes `math` from `cocos2d::math` namespace


dumganhar:
Hi Riq, this pr breaks the ci build, it seems that cocos2d::math namespace was gone, but there are some places where still use it

ricardoquesada:
ouch. who uses it ?
I'll fix it.

UPDATE: I found the bug... fixing it. thanks.

ricardoquesada:
fixed.
https://github.com/cocos2d/cocos2d-x/pull/6584

PR #6587, opened 2014-05-06
Jsb auto play sequence

HoGarfield:
fix bug that it can not auto play sequence by jscontrolled ccb


samuele3hu:
This modification would lead to the callbcack of the animation for Lua no effective. Can you check in js.

PR #6592, opened 2014-05-06
Fix bug that the position of CCMenuItem was not the same as setting in CocosBuilder

HoGarfield:
the position of CCMenuItem is depend on the content size of CCMenu.The default content size of CCMenu in cocosBuilder is (0,0).But not in cocos2d-x.


WenhaiLin:
[ci rebuild]

minggo:
Will not fix bugs about CocosBuilder.

PR #6593, opened 2014-05-06
Improves GLProgram

super626:
Adds vetex attrib bind class


super626:
Please, don't merge this yet :)

tools/cocos2d-console
```diff
@@ -1 +1 @@
-Subproject commit 7876808788c0666d5f4510ea214e97e10eeaa9a8
+Subproject commit 4dbfa174debb2e98422510d7e0490b1ddb2a0b3d
```
dumganhar, line 1: This PR changes this submodule. Is it correct?

dabingnn, line 1: Thanks, James. 
Maybe someone just change it by accident and commit it.
We need to roll back the changes.

dumganhar, line 1: Should revert it to 7876808788c0666d5f4510ea214e97e10eeaa9a8

tests/cpp-tests/proj.android/project.properties
```diff
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-10
+target=android-19
```
dabingnn, line 11: This changes also need to be rolled back

cocos/scripting/lua-bindings/script/Cocos2dConstants.lua
```diff
@@ -131,7 +131,7 @@ cc.VERTEX_ATTRIB_FLAG_TEX_COORDS    = 0x4
 cc.VERTEX_ATTRIB_COLOR  = 0x1
 cc.VERTEX_ATTRIB_MAX    = 0x3
 cc.VERTEX_ATTRIB_POSITION   = 0x0
-cc.VERTEX_ATTRIB_TEX_COORDS = 0x2
+cc.VERTEX_ATTRIB_TEX_COORD = 0x2
```
dabingnn, line 134: need to change it in lua to make it compatible

cocos/2d/platform/android/java/project.properties
```diff
@@ -12,4 +12,4 @@
 
 android.library=true
 # Project target.
-target=android-10
+target=android-19
```
dabingnn, line 15: roll back

cocos/2d/platform/android/java/src/org/cocos2dx/lib/Cocos2dxVideoView.java
```diff
@@ -682,5 +682,10 @@ public boolean canSeekForward() {
 		// TODO Auto-generated method stub
 		return true;
 	}
-	
+
+    @Override
+    public int getAudioSessionId () {
+       // TODO Auto-generated method stub
+       return 0;
+    }
```
minggo, line 690: These codes will cause compiling error on Android lower version, such as 2.3. Because on lower version, this function is not exist in its parent class, but `@Override` means it is a parent's function.

cocos/2d/platform/android/java/project.properties
Trace0429, line 15: fixed

cocos/scripting/lua-bindings/script/Cocos2dConstants.lua
Trace0429, line 134: fixed

tests/cpp-tests/proj.android/project.properties
Trace0429, line 11: fixed

tools/cocos2d-console
Trace0429, line 1: fixed

cocos/renderer/CCGLProgramState.cpp
```diff
@@ -0,0 +1,432 @@
...
+}
+
+void UniformValue::setCallback(const std::function<void(Uniform*)> &callback)
+{
+	// delete previously set callback
+	// XXX TODO: memory will leak if the user does:
+	//    value->setCallback();
+	//    value->setFloat();
+	if (_useCallback)
+		delete _value.callback;
```
minggo, line 110: Why not declare `_value.callback` as `std::function<void(Uniform*)>`? Then you can just use it like this

``` cpp
_value.callback = callbak;
```

And you don't have to worry about memory leak.

ricardoquesada:
Ideally we should be able to define _value.callback as `std::function<void(Uniform*)>` but `_value` is a union. Unfortunately VisualStudio doesn't support all the C++11 features and we can't do that.

minggo:
Good to know.
Thanks.

ricardoquesada:
yes, in particular the problem is that VS doesn't support non-POD in unions :(

PR #6602, opened 2014-05-07
remove unneeded codes

minggo:


minggo:
Please don't merge it.
Just merge it after discussing.

ricardoquesada:
Yes, I know why they are there:

in cocos2d-iphone I added those 2 functions in order to support creating PVR textures in the PVRTexture.m file.

Probable cocos2d-x is using another code to create the PVRTextures, so it might not be needed in cocos2d-x.

minggo:
No. 
cocos2d-x doesn't support creating PVR textures from a file.

But if even we want to support this feature, the logical codes is wrong. Because `Texture2D::keepData` just return passed in parameter.

minggo:
I will merge it if there is not problem.

PR #6609, opened 2014-05-07
[Jenkins][ci skip] Fix copy logic of current_3rd_libs

lmskater:


zawasp:
@dumganhar, I don't see 3rd party libs as a subproject of cocos2d-x. How is this working?
Thanks.

EDIT: do we have to run download-deps.py before or after setup.py?

dumganhar:
> do we have to run download-deps.py before or after setup.py?

Both are ok. :)

> I don't see 3rd party libs as a subproject of cocos2d-x. How is this working?

Please wait some days. :) I will let you know when it's ready.

PR #6610, opened 2014-05-07
upgrade win32 projects for Visual Studio 2013

zawasp:
solves Platform Toolset v100 not found.
remove C4005 warnings after upgrading projects.

### DIFF
--- a/cocos/2d/cocos2d.vcxproj
+++ b/cocos/2d/cocos2d.vcxproj
@@ -23,6 +23,8 @@
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '10.0'">v100</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '11.0'">v110</PlatformToolset>
     <PlatformToolset Condition="'$(VisualStudioVersion)' == '11.0' and exists('$(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v7.1A')">v110_xp</PlatformToolset>
+	<PlatformToolset Condition="'$(VisualStudioVersion)' == '12.0'">v120</PlatformToolset>
+	<PlatformToolset Condition="'$(VisualStudioVersion)' == '12.0' and exists('$(MSBuildProgramFiles32)\Microsoft SDKs\Windows\v7.1A')">v120_xp</PlatformToolset>	
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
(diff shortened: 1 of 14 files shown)


minggo:
Thanks @zawasp.

But we don't want to support too many VS versions. Which will bring more work to maintain it. And we are limited in resources.
Sorry about that.

zawasp:
Hey @minggo, Visual Studio 2013 support is necessary, if you want to target Windows Phone 8.1 or Windows 8.1 - you can't target those platforms with older VS.
The changes made are only related to the project files, not the code - as it builds perfectly with VS 2013.
And you should always support the latest tools, not the oldest.
Please reconsider.

minggo:
But we already supports Windows Phone 8 and Windows 8 without the modification. Because many developers using windows 7 and windows XP for developing. So we should choose one tools that most developers use. And VS 2012 can be installed in Windows 8, right?

G0561:
compatibility  is  most important  ，so  come  on 

发件人: "minggo"; 
发送时间: 2014年5月8日(星期四) 下午3:04
收件人: "cocos2d/cocos2d-x"; 
主题: Re: [cocos2d-x] upgrade win32 projects for Visual Studio 2013 (#6610)

But we already supports Windows Phone 8 and Windows 8 without the modification. Because many developers using windows 7 and windows XP for developing. So we should choose one tools that most developers use. And VS 2012 can be installed in Windows 8, right?

zawasp:
WP8 yes, WP8.1 no. The changes made to the project DO NOT affect the support for VS 2010/2012. Not to mention you can target both Windows 8.1 apps and Windows Phone 8.1 apps with ONE project.

minggo:
I know it doesn't modify the codes.
But we should maintain it, which means:
- modify it when add/remove files
- install windows 8.1 to install VS2013
- do testing before releasing
- CI to test it in every pull request

It means much work, not just add the modification.

zawasp:
No you don't have to modify it when you add files - it's the same project file for VS 2010/2012/2013. So you don't need to install VS2013.
As the code is the same, you don't have to test it also. Have a look at the changes, please. It just adds some lines, similar to the ones that add support for VS 2012.

Support for VS 2012:
    `<PlatformToolset Condition="'$(VisualStudioVersion)' == '11.0'">v110</PlatformToolset>`

Support for VS 2013:
    `<PlatformToolset Condition="'$(VisualStudioVersion)' == '12.0'">v120</PlatformToolset>`

It's not rocket science.

minggo:
Thanks for explanation.

zawasp:
Thank you.

PR #6619, opened 2014-05-08
update cpp-default-template for android *Eclipse*

myourys:
update cpp-default-template for android _Eclipse_,you can use it for other template.

and the new **README.md** file has updated, it's easy to import projects to eclipse,thanks!

Major changes：
1. add a CDT config file .cproject that eclipse can realize this is a cdt project.
2. fix .project
3. add eclipse config (.setting)


WenhaiLin:
We can using python script for copy resource,compile code and pack android application,it does not need to rely to 'Cygwin'. It's more convenient.What do you think?

myourys:
@Dhilan007 we can use eclipse develop jni code ,add java library and debug（filter log ,debug java code ） more convenience. this request is major for eclipse.

WenhaiLin:
@minggo It works fine.Please review and merged.

minggo:
Ok.
Thanks all.

PR #6639, opened 2014-05-08
Added missing files to WP8 projects

stammen:
This pull request adds the missing files to cocos2d and cpp-tests projects to fix the broken WP8 build caused by recent changes to the V3 branch.


dumganhar:
Thanks.

PR #6640, opened 2014-05-08
Added missing Assets files to WP8 projects

stammen:
This pull requests adds the missing Assets folders to the WinRT, WP8 and WP8-XAML projects. These files are required in order to build the projects.


dumganhar:
Thanks.

PR #6642, opened 2014-05-09
Update README.md

chygmm:
in v2.x,  is the file  "cocos2dx/platform/third_party/android/prebuilt" delete?


minggo:
I had posted a news for it http://cocos2d-x.org/news/226

minggo:
And in the forum http://cocos2d-x.org/forums/6/topics/53206

PR #6635, opened 2014-05-08
issue #4660: ccUTF8 uses utf8-utf16 conversion in llvm3.4

dumganhar:

PR #6643, opened 2014-05-09
[v2] issue #4660: Uses utf8/16 conversion in llvm3.4

dumganhar:


Seasons7:
@dumganhar 
Thank you for your implementation for UTF8/16 conversion.
I think your suggestion and implementation is good for all developers about license( avoid LGPL ) and stability.

I appreciate your review and point to code issue.
We are happy to use our test code for cocos2d-x.

dumganhar:
@Seasons7 , thank you for sharing your `test` codes, I will submit another PR soon.
So do you also think the `llvm utf conversion functions` is ok for you?

Seasons7:
Our desire is to prevent reverse engineering by users. LGPL license allow to reverse engineering. 

I think that new solution will solve it. 
What do you think?

dumganhar:
What do you mean `new solution`? My solution in this PR? If yes, thanks for you like it too.
Yep, we should not add any codes that's under LGPL license although the old ccutf8 implementation just use 3-4 functions. :)

Seasons7:
Yes, "new solution" is your this PR. Sorry, it was need further explanation.
Now we are happy to be able to solve this LGPL license issue :+1:

dumganhar:
Thank you. BTW, my PR for v3 branch (https://github.com/cocos2d/cocos2d-x/pull/6635) also improved the utf8/16 conversion functions. 

``` c++
namespace StringUtils {

/**
 *  @brief Converts utf8 string to utf16 string
 *  @param utf8 The utf8 string to be converted
 *  @param outUtf16 The output utf16 string
 *  @return true if succeed, otherwise false
 *  @note Please check the return value before using \p outUtf16
 *  e.g.
 *  @code
 *    std::u16string utf16;
 *    bool ret = StringUtils::UTF8ToUTF16("你好hello", utf16);
 *    if (ret) {
 *        do_some_thing_with_utf16(utf16);
 *    }
 *  @endcode
 */
CC_DLL bool UTF8ToUTF16(const std::string& utf8, std::u16string& outUtf16);

/**
 *  @brief Converts utf16 string to utf8 string
 *  @param utf16 The utf16 string to be converted
 *  @param outUtf8 The output utf8 string
 *  @return true if succeed, otherwise false
 *  @note Please check the return value before using \p outUtf8
 *  e.g.
 *  @code
 *    std::string utf8;
 *    bool ret = StringUtils::UTF16ToUTF8(u"\u4f60\u597d", utf16);
 *    if (ret) {
 *        do_some_thing_with_utf8(utf8);
 *    }
 *  @endcode
 */
CC_DLL bool UTF16ToUTF8(const std::u16string& utf16, std::string& outUtf8);

/**
 *  @brief Trims the unicode spaces at the end of char16_t vector
 */
CC_DLL void trimUTF16Vector(std::vector<char16_t>& str);

/**
 *  @brief Whether the character is a whitespace character.
 *
 *  @param ch    the unicode character
 *  @returns     whether the character is a white space character.
 *
 *  @see http://en.wikipedia.org/wiki/Whitespace_character#Unicode
 *
 */
CC_DLL bool isUnicodeSpace(char16_t ch);

/**
 *  @brief Whether the character is a Chinese/Japanese/Korean character.
 *
 *  @param ch    the unicode character
 *  @returns     whether the character is a Chinese character.
 *
 *  @see http://www.searchtb.com/2012/04/chinese_encode.html
 *  @see http://tieba.baidu.com/p/748765987
 *
 */
CC_DLL bool isCJKUnicode(char16_t ch);

/**
 *  @brief Returns the length of the string in characters.
 *
 *  @param utf8 an UTF-8 encoded string.
 *  @returns the length of the string in characters
 */
CC_DLL long getCharacterCountInUTF8String(const std::string& utf8);

/**
 *  @brief Gets the index of the last character that is not equal to the character given.
 *
 *  @param str   the string to be searched.
 *  @param c     the character to be searched for.
 *
 *  @returns the index of the last character that is not \p c.
 *
 */
CC_DLL unsigned int getIndexOfLastNotChar16(const std::vector<char16_t>& str, char16_t c);

/**
 *  @brief Gets char16_t vector from a given utf16 string
 */
CC_DLL std::vector<char16_t> getChar16VectorFromUTF16String(const std::u16string& utf16);

} // namespace StringUtils {
```

You could refer to that for the new api. And the old APIs was deprecated.
I didn't do it for v2 branch.

Seasons7:
@dumganhar 
Great :) It is good news for cocos2d-x v3.x developers.

PR #6647, opened 2014-05-09
Migrate spritebuilder

toughrogrammer:
I works about migrating cocosbuilder. Some features don't works very well, but don't occure runtime error.


dumganhar:
Since cocos2d-x 3.0 doesn't follow cocos2d-iphone. There're some differences between `-x` and `-iphone`. For example, `-x` has its own `gui` which is different with which in `-iphone`.
How to deal with these differences?

toughrogrammer:
Yeah I know cocos2dx is really different with cocos2d-iphone. But SpriteBuilder is very useful tool for making layout. So many developer use CocosBuilder to do it, but there isn't CocosBuilder now. It's changed to SpriteBuilder.

hanjukim:
@growingdever This is an issue for me too. I am using CocosBuilder but it's very out dated and I wish to migrate CocosBuilder project to SpriteBuilder. Great work. However, It would be very nice since both has almost same codes everywhere which is not very nice.. If it is possible, how about adding support for SpriteBuilder on existing code?

toughrogrammer:
I agree too. I know this pull request is not perfect for supporting SpriteBuilder completely, but I thought I need another developers help. My commit is working about read binary of ccbi file. It's very stupid work but someone must did it. Now SpriteBuilder::CCBReader is read ccbi file successful, so another developer simply need to modify Loader class.

minggo:
@growingdever 
@hanjukim 
You can use CocoStudio, which can be [downloaded here](http://cocos2d-x.org/download). It is our official editor.

hanjukim:
@minggo I did try to use CocosStudio but IMHO it's not ready for production yet. It has some annoying bugs.. and It's not open source project. so I cannot fix it. It's very frustrating..

minggo:
@hanjukim 
Could you describe what's the bugs? I will ask the developers to give you support.
Thanks.

newnon:
If you interesting for sprite builder i work on its integration some time and my work already finished at 90% you can try it 
SpriteBuilderX - special fork for cocos2d-x
https://github.com/newnon/SpriteBuilder/tree/v1.0  - branch `v1.0`
and cocos with its support(based on v3.0 release)
https://github.com/newnon/cocos2d-x
if you fave some questions feel free to contact me with skype

PR #6648, opened 2014-05-09
Parse websocket url

takeru:
fixed for parse like this "ws://example.com/?name=value"

### DIFF
--- a/extensions/network/WebSocket.cpp
+++ b/extensions/network/WebSocket.cpp
@@ -275,24 +275,26 @@ bool WebSocket::init(const Delegate& delegate,
         host.erase(0,6);
         useSSL = true;
     }
-    
+
+    std::string path = "/";
+    pos = host.find("/");
+    if(pos >= 0){
+        path = host.substr(pos, host.size());
+        host.erase(pos, host.size());
+    }else{
+        pos = host.find("?");
+        if(pos >= 0){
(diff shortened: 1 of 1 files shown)


extensions/network/WebSocket.cpp
```diff
@@ -275,24 +275,26 @@ bool WebSocket::init(const Delegate& delegate,
...
-    pos = host.find("/", pos);
+
     std::string path = "/";
+    pos = host.find("/");
     if(pos >= 0){
-        path += host.substr(pos + 1, host.size());
+        path = host.substr(pos, host.size());
+        host.erase(pos, host.size());
+    }else{
+        pos = host.find("?");
```
dumganhar, line 285: If the url is `"ws://example.com/?name=value"`, would this `else` be executed?

takeru, line 285: else will be execute with:
"ws://example.com?name=value"
"ws://example.com"
"ws://example.com:8080"
"ws://example.com:8080?name=value"

Url without "/" may be invalid. Then it should be failed to connect.

zilongshanren:
@takeru  Currently our merge strategy on V2 branch is "only critical bug fix" are accepted. So perhaps you'd better send this PR to v3 branch, only if there is also the same issue there.

Thanks.

PR #3244, opened 2013-07-24
Adding popSceneBack() function to Director

mdkcore0:
This function deletes the running scene and returns the actual scene
back to the caller. This way, we can simply use it to perform a
transition animation, like this:

Director::sharedDirector()->replaceScene(
        TransitionSlideInL::create(0.5,
        Director::sharedDirector()->popSceneBack()));

PR #6654, opened 2014-05-10
Add new functionality for managing scenes. Fix bugs, added example of use.

zifter:
Now, next methods returning _nextScene.
Scene\* popScene();
Scene\* popToRootScene();
Scene\* popToSceneStackLevel(int level);
This makes it possible to creating transition for popScene (or other).
Example of use:
Director::sharedDirector()->replaceScene(TransitionSlideInL::create(0.5,Director::sharedDirector()->popSceneBack()));
Similar pull request https://github.com/cocos2d/cocos2d-x/pull/3244

But this pull request is wrong, because Replace scene cleaning up _nextScene.
In method  "replaceScene" now checks this situation.
Also, fixed bug with double calling onExitTransitionDidStart() and onExit() in popToSceneStackLevel for running scene.
Example of use for this functionality added to TestScene. Аdded more visual presentation for using this functionality.

### DIFF
--- a/tests/cpp-tests/Classes/SceneTest/SceneTest.cpp
+++ b/tests/cpp-tests/Classes/SceneTest/SceneTest.cpp
@@ -30,10 +30,18 @@ SceneTestLayer1::SceneTestLayer1()
     auto s = Director::getInstance()->getWinSize();
     auto sprite = Sprite::create(s_pathGrossini);
     addChild(sprite);
-    sprite->setPosition( Vector2(s.width-40, s.height/2) );
     auto rotate = RotateBy::create(2, 360);
     auto repeat = RepeatForever::create(rotate);
     sprite->runAction(repeat);
+
+    auto move = Sequence::create(
+        MoveTo::create(2, Vector2(0.0f, s.height)),
+        MoveTo::create(2, Vector2(s.width, s.height)),
+        MoveTo::create(2, Vector2(s.width, 0.0f)),
+        MoveTo::create(2, Vector2(0.0f, 0.0f)),
+        nullptr);
+    auto repeatMove = RepeatForever::create(move);
+    sprite->runAction(repeatMove);
 
     schedule( schedule_selector(SceneTestLayer1::testDealloc) );
 }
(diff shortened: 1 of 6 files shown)


zifter:
This pull request contains fix for http://www.cocos2d-x.org/issues/5091

tests/cpp-tests/Classes/SceneTest/SceneTest.cpp
```diff
@@ -184,10 +203,19 @@ bool SceneTestLayer3::init()
...
         sprite->runAction(repeat);
+
+        auto move = Sequence::create(
+            MoveTo::create(2, Vector2(0.0f, s.height)),
+            MoveTo::create(2, Vector2(s.width, s.height)),
+            MoveTo::create(2, Vector2(s.width, 0.0f)),
+            MoveTo::create(2, Vector2(0.0f, 0.0f)),
+            nullptr);
+        auto repeatMove = RepeatForever::create(move);
+        sprite->runAction(repeatMove);
```
minggo, line 217: Why add these codes?

minggo:
And why `onExitTransitionDidStart()` and `onExit()` are invoked twice? 
Could you describe it in more detail?
Thanks.

minggo:
I think `popToSceneStackLevel` should not trigger `onExitTransitionDidStart` event.

tests/cpp-tests/Classes/SceneTest/SceneTest.cpp
zifter, line 217: With this activity, we can see that the transition to the scene is performing correctly: the scene was paused and resumed again. Only need to visualize this process.

zifter:
Bug with `onExitTransitionDidStart()` and `onExit()` I described in http://www.cocos2d-x.org/issues/5091.
Detailed description:
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/CCDirector.cpp

```
// pop stack until reaching desired level
while (c > level)
{
    auto current = _scenesStack.back();

    if (current->isRunning())
    {
        current->onExitTransitionDidStart();
        current->onExit();
    }

    current->cleanup();
    _scenesStack.popBack();
    --c;
}

_nextScene = _scenesStack.back();
_sendCleanupToScene = false;
```

In old code. For current scene calling `onExit()` and `onExitTransitionDidStart()`, because performing condition `current->isRunning()`.
BUT, line `_nextScene = _scenesStack.back();` changing `_nextScene`.
In drawScene method:

  if (_nextScene)
    {
        setNextScene();
    }
In this method performed normal transition from current scene to destination. which calling `onExit()` and `onExitTransitionDidStart()` for current scene.
Ask me if something is not clearly explained. Sorry for the my English.

minggo:
I sent #6681 for it.

minggo:
Thanks for description.
But i think `_nextScene` will not be the same as `current`.

zifter:
You have misunderstood me.
Of course, `_nextScene` will not be the same as `current`. current it's a `_runningScene`.

You added this line only in `replaceScene`. 
`if (scene == _nextScene)
     return;`
But after Replace, I can calling `popToStackLevel()` and get invalid state.

You fix working only for pop scene without transition, but my pull requset fixed this.
Until now, there is no opportunity to make the transition at popScene.

zifter:
And please, can you rebuild build for my pull request?
I see you have some problem with builder and because of this my pull request in the wrong state "Failed".

PR #6554, opened 2014-05-03
Android studio gradle cpp-template

asmodehn:
This is a new cpp-template that can be built out of the box with android studio.
Tested on Linux and Windows with Android-Studio 0.5.7

PR #6656, opened 2014-05-10
Android studio gradle cpp-template

asmodehn:
This is a new cpp-template that can be built out of the box with android studio.
Tested on Linux and Windows with Android-Studio 0.5.7

This is a rebase of the original pull request #6554 after cocos2d-x repository compression.


minggo:
@asmodehn 
Thanks for the pull request.

About Android Studio, my opinion is
- how stable of Android Studio now?
- does most Android developers use Android Studio now?
- if we support Android Studio, then we will not support Eclipse any more, because we don't want to maintain two ways to do the same thing

asmodehn:
@minggo 
I am not sure how stable is Android studio on the long term. I have started using it recently for a new project and currently I am setting up build tools for a game using it.

I started using it because :
- i wanted to stop using ant and move to gradle for java builds. Google has moved already (android -19) and ant is a legacy system already.
- I personally dont like Eclipse, so i was using Netbeans with NBAndroid. but the setup is time consuming and prone to errors, and I am doing it on linux and windows.

Android studio provided for me a working debug setup out of the box, with only one setting : the SDK path. On my first try everything worked out of the box after setting the SDK path, including debugging the app (java code only).
So that was convincing enough for me.

I am not sure about most Android developers, as my company is very small. but I believe developers will eventually switch to gradle anyway.

I know the trouble it is to maintain multiple toolsets for a software to build on multiple platform, so about switching from Eclipse to Android studio, I guess there could be a poll on the forum to see who is interested ?
But I also think that most people tend to stick with what they know, even if there are tons of reasons to change it, learning a new tool always takes some time...

My personal opinion is that the switch ant -> gradle should be made sometime soon, based on technical reason only.
But whether using eclipse or android studio or some other IDE should be left to the user.
Android studio can "import projects" based on gradle, and most java IDE as well I believe...

minggo:
@asmodehn 
Thanks for the description.
I will try Android Studio a bit and i will also create a thread to ask the forum.
I will paste the thread url when i finish.

minggo:
Android studio is supported.

PR #6657, opened 2014-05-10
Fixing cmake project for win32

asmodehn:
The cmakelists that I tested worked on linux but not on win32.
I made small fixes to get it to work on win32 platform.

This probably needs more thorough testing, to make sure nothing is broken on other platforms where it is supposed to work but that I couldnt test on.


cocos/audio/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ if(WIN32)
   set(AUDIO_SRC
     win32/SimpleAudioEngine.cpp
     win32/MciPlayer.cpp
+	win32/MciPlayer.h
```
dumganhar, line 12: Hi, do we need to add `.h` header file to `AUDIO_SRC`?

asmodehn, line 12: This is only useful for IDE projects. the .h file will be listed and visible as part of the project.
The compiler will ignore it.

dumganhar, line 12: OK, got it.

dumganhar:
Linux ci build was broken, could you check it?
the error information is:

```
CMake Error at cocos/network/CMakeLists.txt:37 (ENDIF):
  endif An ENDIF command was found outside of a proper IF ENDIF structure.
  Or its arguments did not match the opening IF command.
  cocos/CMakeLists.txt:17 (include)


make: *** No targets specified and no makefile found.  Stop.
build finished and return 512
```

asmodehn:
I had this problem on my version after merging a recent v3...
The auto-merge on network/CMakeLists.txt got confused.
I ll push a fix in the next few hours.

PR #6670, opened 2014-05-10
Fix the variable "isTween" value error.

leafvmaple:
In the function "FrameData::copy", the variable "isTween" had not
assigned correctly.


dumganhar:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/6734
Thanks.

PR #6676, opened 2014-05-12
Fix the RenderTexture bounding box issue.

visiblelight:
1. the default anchor point of RenderTexture is center now; 
2. the default anchor point of the bound sprite is left-bottom now which insures its BB is consistent with the RenderTexture; 
3. the transforms of the bound children of a RenderTexture should ONLY based on the their parent. 
4. update RenderTextureTest.


minggo:
I think we can not modify the usage of `Sprite`.

PR #6678, opened 2014-05-12
fixed compile fail on android.

WenhaiLin:


ricardoquesada:
Tony needed the `@override` because it was not compiling without it.
strange.

minggo:
It doesn't need `@override`. We will help @Tony to resolve it.

ricardoquesada:
ok. thanks.

PR #6681, opened 2014-05-12
don't trigger onExitTransitionDidStart

minggo:


zifter:
It does not fix the bug with twice calling onExit().
If you want, I may create pull request where this bug are reproduced.

minggo:
@zifter 
Yep. This PR is just for fix the bug of `trigger onExitTransitionDidStart`.
I don't want to mix many things in a PR.

About `onExit` issue.
I think you are right.
I will fix it in other PR.

minggo:
@zifter 
It is better to send me a PR to reproduce it.
Thanks.

zifter:
@minggo 
I create PR to reproduce it.
https://github.com/cocos2d/cocos2d-x/pull/6757
And fix:
https://github.com/cocos2d/cocos2d-x/pull/6756

minggo:
Thanks.

cocos/base/CCDirector.cpp
```diff
@@ -925,7 +926,6 @@ void Director::purgeDirector()
 
     if (_runningScene)
     {
-        _runningScene->onExitTransitionDidStart();
```
miguel12345, line 928: I am trying to understand why this line was removed. It doesn't make much sense to me because onExit will be called without onExistTransitionDidStart being called before it.

I do understand that the other changes were made because onExitTransitionDidStart was called twice, but when purging the director it seems impossible for it to happen, am I right?

zifter, line 928: Methods onExitTransitionDidStart and onEnterTransitionDidFinish must be called only of 
Calls to methods onExitTransitionDidStart and onEnterTransitionDidFinish should be only if the transition occurs with the TransitionScene, is't it?

miguel12345, line 928: I can understand what you are saying, but let's imagine my game is made in a way that most cleanup is made in onExitTransitionDidStart and not on onExit, because I replace scenes using a TransitionScene and I don't want to generate any conflicts between the two scenes. (imagine my nodes use platform-specific services like Text-to-speech, audio recording, etc). 

This way, if I purge the director, my onExitTransitionDidStart methods will not be called and I will end up with resources that were not properly cleaned.

Of course I could move all that logic to onEnter/onExit instead of onExitTransitionDidStart/onEnterTransitionDidFinish but it just doesn't seem natural.

PR #6693, opened 2014-05-12
Fixes for WP8 V3 build recently broken by recent commits

stammen:
This pull request contains the fixed for the broken WP8 build. The build was broken by recent commits by the community. The precompiled shaders were updated to use the latest changes to the shader code.


ricardoquesada:
Thanks!

PR #6707, opened 2014-05-13
fix widget enable issue

zilongshanren:


cocos/ui/UIWidget.cpp
```diff
@@ -539,11 +516,56 @@ void Widget::didNotSelectSelf()
...
+    
+    return parentWidget->isAncestorsEnable();
+}
 
 bool Widget::onTouchBegan(Touch *touch, Event *unusedEvent)
 {
     _hitted = false;
-    if (isEnabled() && isTouchEnabled())
+    //check parent
+    if (isAncestorsEnable() && isAncestorsVisible(this) && isVisible() && isEnabled())
```
minggo: I think it is better to checkout self visible and enabled first. Because it is more quick. And if self visible or enabled is false, then we don't have to check its ancestors' state.

zilongshanren: good point!

cocos/ui/UIWidget.cpp
```diff
@@ -539,11 +516,64 @@ void Widget::didNotSelectSelf()
...
+    Node* parent = node->getParent();
+    
+    if (parent && !parent->isVisible())
+    {
+        return false;
+    }
+    return this->isAncestorsVisible(parent);
+}
+    
+bool Widget::isAncestorsEnable()
```
minggo: isAncestorsEnable -> isAncestorsEnabled

PR #6714, opened 2014-05-13
Fix a bug that if a node changes parent, parent's matrix will not update to reflect this change.

WenhaiLin:


minggo:
[ci rebuild]

PR #6716, opened 2014-05-13
closed #4900, add TTF font to UIText

zilongshanren:


cocos/ui/UIText.h
```diff
@@ -95,12 +100,16 @@ class Text : public Widget
 
     /**
      * Sets the font name of label.
+     *  If you are trying to use a system font, you could just pass a font name, for example: Arial
+     * If you are trying to use a TTF, you should pass a file path to the TTF file, for example: resource/XXX.ttf
```
minggo: May be add some pseudocode is better.

PR #6725, opened 2014-05-14
Adds ref memory tracker to easily find out memory leak for instance of Ref or its subclasses

dumganhar:


cocos/base/CCRef.cpp
```diff
@@ -121,4 +144,51 @@ unsigned int Ref::getReferenceCount() const
...
+        for (const auto& ref : __refAllocationList)
+        {
+            CC_ASSERT(ref);
+            const char* type = typeid(*ref).name();
+            log("[memory] LEAK: Ref object '%s' still active with reference count %d.\n", (type ? type : ""), ref->getReferenceCount());
+        }
+    }
+}
+
+void trackRef(Ref* ref)
```
minggo: Define with static here?

dumganhar: Good catch, I will fix it. Thanks.

dumganhar:
[ci rebuild]

PR #6726, opened 2014-05-14
Add a new utility function to snap the screen

visiblelight:
This function can't be invoked synchronous at present.
The users need to specify a callback handler to process further work after the screen is captured.


cocos/renderer/CCRenderer.cpp
```diff
@@ -535,4 +538,81 @@ bool Renderer::checkVisibility(const Matrix &transform, const Size &size)
     return ret;
 }
 
+void Renderer::captureScreen(std::function<void(bool, const std::string&)> afterCaptured, const std::string& filename, const Rect& rect)
```
dumganhar: Please use `const std::function<>&` for `afterCaptured` parameter.

cocos/renderer/CCRenderer.cpp
```diff
@@ -535,4 +538,81 @@ bool Renderer::checkVisibility(const Matrix &transform, const Size &size)
     return ret;
 }
 
+void Renderer::captureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename, const Rect& rect)
```
dabingnn: Another solution is to add the interface in the director. What about use this interface?
instead of using `Director::getInstance()->getRenderer()->captureScreen()`, we only need to call `Director::getInstance()->captureScreen()`

cocos/renderer/CCRenderer.cpp
```diff
@@ -535,4 +538,81 @@ bool Renderer::checkVisibility(const Matrix &transform, const Size &size)
...
+            image->initWithRawData(buffer, width * height * 4, width, height, 8);
+            CC_SAFE_DELETE_ARRAY(buffer);
+            outputFile = FileUtils::getInstance()->getWritablePath() + filename;
+            image->saveToFile(outputFile);
+            succeed = true;
+        }
+        CC_SAFE_DELETE(image);
+    }while(0);
+    	
+    afterCaptured(succeed, outputFile);
```
dabingnn: Can we support a null callback here?
If true, we need to add a `if(XXX)`.

cocos/renderer/CCRenderer.cpp
minggo: Does it need to be added in Renderer?
I think this function can be put in every where.
So why not put it in `ccUtils.h/.cpp`?

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
```diff
@@ -559,3 +560,76 @@ std::string VBOFullTest::subtitle() const
...
+        auto sp = Sprite::create(outputFile);
+        addChild(sp, 0, ChildTag);
+        Size s = Director::getInstance()->getWinSize();
+        sp->setPosition(s.width / 2, s.height / 2);
+        sp->setScale(0.25);
+        _filename = outputFile;
+    }
+    else
+    {
+	    log("Something went wrong");
```
dabingnn: How about add a log `capture screen Failed` here?

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
```diff
@@ -559,3 +560,76 @@ std::string VBOFullTest::subtitle() const
...
+    }
+}
```
dabingnn: I guess we can add a blank newline in the end of the file.

cocos/renderer/CCRenderer.cpp
```diff
@@ -535,4 +538,81 @@ bool Renderer::checkVisibility(const Matrix &transform, const Size &size)
...
 }
 
+void Renderer::captureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename, const Rect& rect)
+{
+    _captureScreen.init(std::numeric_limits<float>::max());
+    _captureScreen.func = CC_CALLBACK_0(Renderer::onCaptureScreen, this, afterCaptured, filename, true, rect);
+    addCommand(&_captureScreen);
+}
+
+void Renderer::onCaptureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename, bool flipped, const Rect& rect)
```
minggo: Why `flipped` is needed?

cocos/renderer/CCRenderer.cpp
```diff
@@ -535,4 +538,81 @@ bool Renderer::checkVisibility(const Matrix &transform, const Size &size)
...
+        width = clip(width, 0, frameSize.width - originx);
+        height = clip(height, 0, frameSize.height - originy);
+    }
+
+    do
+    {
+        GLubyte* buffer = new GLubyte[width * height * 4];
+        if (!buffer)
+        {
+            CC_SAFE_DELETE_ARRAY(buffer);
```
minggo: Why need this line? 
`buffer` is null.

cocos/renderer/CCRenderer.cpp
dabingnn: If we want to support multiple Director, leaving the interface in the Director could be better.

cocos/renderer/CCRenderer.cpp
```diff
@@ -535,4 +538,81 @@ bool Renderer::checkVisibility(const Matrix &transform, const Size &size)
...
+		
+        glPixelStorei(GL_PACK_ALIGNMENT, 1);
+        glReadPixels(originx, originy, width, height, GL_RGBA, GL_UNSIGNED_BYTE, buffer);
+
+        if (flipped)
+        {
+            GLubyte* flippedBuffer = new GLubyte[width * height * 4];
+            if (!flippedBuffer)
+            {
+                CC_SAFE_DELETE(flippedBuffer);
```
minggo: Should delete `buffer`.

cocos/renderer/CCRenderer.cpp
```diff
@@ -535,4 +538,81 @@ bool Renderer::checkVisibility(const Matrix &transform, const Size &size)
...
+            memcpy(buffer, flippedBuffer, width * height * 4);
+            CC_SAFE_DELETE_ARRAY(flippedBuffer);
+        }
+
+        Image* image = new Image();
+        if (image)
+        {
+            image->initWithRawData(buffer, width * height * 4, width, height, 8);
+            CC_SAFE_DELETE_ARRAY(buffer);
+            outputFile = FileUtils::getInstance()->getWritablePath() + filename;
```
minggo: If the `filename` is an absolute path, such as `/sd/xxx`, then what will happen?

cocos/renderer/CCRenderer.cpp
visiblelight: Although a null callback makes no sense, yes, it's better to make add a guard. I am updating.

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
visiblelight: accepted.

cocos/renderer/CCRenderer.cpp
visiblelight: The flipped param is removed from the public interface, I am keeping it in the implement method just in case that we would have further modification.

cocos/renderer/CCRenderer.cpp
visiblelight: just in case that the new operator failed when allocating memory.

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
```diff
@@ -559,3 +560,71 @@ std::string VBOFullTest::subtitle() const
...
+        auto sp = Sprite::create(outputFile);
+        addChild(sp, 0, ChildTag);
+        Size s = Director::getInstance()->getWinSize();
+        sp->setPosition(s.width / 2, s.height / 2);
+        sp->setScale(0.25);
+        _filename = outputFile;
+    }
+    else
+    {
+        log("Captrue screen failed.");
```
dabingnn: Typo

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.h
```diff
@@ -146,4 +146,22 @@ class VBOFullTest : public MultiSceneTest
     virtual ~VBOFullTest();
 };
 
+class CaptureScreenTest : public MultiSceneTest
+{
+    static const int ChildTag = 119;
```
dabingnn: Please use `childTag`, `ChildTag` is used for class name.

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
```diff
@@ -559,3 +560,71 @@ std::string VBOFullTest::subtitle() const
...
+}
+
+std::string CaptureScreenTest::title() const
+{
+    return "New Renderer";
+}
+
+std::string CaptureScreenTest::subtitle() const
+{
+    return "Capture screen test, press the munu items to capture the screen";
```
dabingnn: typo for `menu`

cocos/base/CCDirector.cpp
```diff
@@ -1216,6 +1216,71 @@ void Director::setEventDispatcher(EventDispatcher* dispatcher)
...
+{
+    _captureScreen.init(std::numeric_limits<float>::max());
+    _captureScreen.func = CC_CALLBACK_0(Director::onCaptureScreen, this, afterCaptured, filename);
+    Director::getInstance()->getRenderer()->addCommand(&_captureScreen);
+}
+
+void Director::onCaptureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename)
+{
+    Size frameSize = _openGLView->getFrameSize();
+#if CC_TARGET_PLATFORM == CC_PLATFORM_MAC
```
dabingnn: If the `getRetinaFactor` applies to all desktop platform, It is better use this code snap in mac/linux/windows, not just for mac platform

cocos/base/CCDirector.cpp
```diff
@@ -1216,6 +1216,71 @@ void Director::setEventDispatcher(EventDispatcher* dispatcher)
     }
 }
 
+void Director::captureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename)
+{
+    _captureScreen.init(std::numeric_limits<float>::max());
```
minggo: `_captureScreen` -> `_captureScreenCommand`

cocos/base/ccUtils.cpp
```diff
@@ -38,4 +44,74 @@ int ccNextPOT(int x)
     return x + 1;
 }
 
-}
\ No newline at end of file
+namespace Utilities
```
dabingnn, line 47: Why do we need sub namespace? Is it necessary? Does it make sense if we just use cocos2d namespace?

minggo, line 47: I think it makes sense. 
Because we will add more functions into it.

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
```diff
@@ -559,3 +560,71 @@ std::string VBOFullTest::subtitle() const
...
+}
+
+CaptureScreenTest::~CaptureScreenTest()
+{
+    Director::getInstance()->getTextureCache()->removeTextureForKey(_filename);
+}
+
+std::string CaptureScreenTest::title() const
+{
+    return "New Renderer";
```
minggo, line 599: Please change the title

minggo, line 599: Or, forget me.

cocos/base/ccUtils.h
```diff
@@ -46,6 +47,19 @@ ccNextPOT function is licensed under the same license that is used in Texture2D.
 
 int ccNextPOT(int value);
 
+namespace Utilities
+{
+    /**
+     * Capture screen interface
+     */
+    void captureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename);
```
minggo, line 55: Please add doxygen doc for this function.

cocos/base/ccUtils.cpp
winipcfg, line 47: Isn't that namespace should be short and in lower case?

minggo, line 47: Yep.
How about `utils`?

minggo, line 47: Or how about use class instead of namespace.

PR #6737, opened 2014-05-14
EventDispatcher bug fix

seobyeongky:
Hi
This is EventDispatcher bug fix branch.
The issue occurs when...
1. Push the button. (So, EventDispatcher's _inDispatch > 0)
2. Create a layer.
3. Call EventDispatcher's addListenerSceneGraph to listen touch events. The node is the layer.
4. Not associated immediately. Because _inDispatch is bigger then 0.
5. Call Director's pushScene a scene that has the layer.
5. So when the layer is entering -> layer resumed -> calls EventDispatcher's resumeEventListenersForTarget -> cannot find the layer. -> added listener is not resumed. -> bug

### DIFF
--- a/cocos/base/CCEventDispatcher.cpp
+++ b/cocos/base/CCEventDispatcher.cpp
@@ -293,11 +293,11 @@ void EventDispatcher::pauseEventListenersForTarget(Node* target, bool recursive/
         }
     }
 
-    for (auto addedListenerIter : _toAddedListeners)
+    for (auto& listener : _toAddedListeners)
     {
-        if (addedListenerIter->getAssociatedNode() == target)
+        if (listener->getAssociatedNode() == target)
         {
-            addedListenerIter->setPaused(true);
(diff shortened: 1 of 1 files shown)


cocos/base/CCEventDispatcher.cpp
```diff
@@ -292,6 +292,14 @@ void EventDispatcher::pauseEventListenersForTarget(Node* target, bool recursive/
             l->setPaused(true);
         }
     }
+
+    for (auto addedListenerIter : _toAddedListeners)
```
dumganhar: Thanks. The  variable `addedListenerIter` in `range loop`  is not a `iterator`, so how about just rename it to 
`for (auto& listener : _toAddedListeners)`?

seobyeongky: Yes I am not practiced at new c++  haha

dumganhar:
[ci rebuild]

PR #6740, opened 2014-05-14
Fix Progress Action. Not working with charging from 100 to 0.

zifter:
Do not change the value in the deep methods.
RepeatForever::create(ProgressTo::create(2, 100));
This means that you want to repeat to progress 100. But if progress is equal to 100 - nothing should happen.
If you want to repeat charging from 0 to 100, you must drop down value to 0.
RepeatForever::create(Sequence::createWithTwoActions(ProgressTo::create(2, 100), ProgressTo::create(0, 0)));

### DIFF
--- a/cocos/2d/CCActionProgressTimer.cpp
+++ b/cocos/2d/CCActionProgressTimer.cpp
@@ -72,13 +72,6 @@ void ProgressTo::startWithTarget(Node *target)
 {
     ActionInterval::startWithTarget(target);
     _from = ((kProgressTimerCast)(target))->getPercentage();
-
-    // XXX: Is this correct ?
-    // Adding it to support Repeat
-    if (_from == 100)
-    {
-        _from = 0;
-    }
 }
 
 void ProgressTo::update(float time)
--- a/tests/cpp-tests/Classes/ActionsProgressTest/ActionsProgressTest.cpp
+++ b/tests/cpp-tests/Classes/ActionsProgressTest/ActionsProgressTest.cpp
@@ -157,8 +157,8 @@ void SpriteProgressToRadial::onEnter()
     
     auto s = Director::getInstance()->getWinSize();
 
-    auto to1 = ProgressTo::create(2, 100);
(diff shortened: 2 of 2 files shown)


minggo:
I am not sure if we should change current behavior, because:
- it will break compatibility
- it is not convenient to use together with Repeat

Is there any problem with current behavior?

@ricardoquesada 
What's your opinion?

zifter:
Anyway, we can not correctly use ProgressTo with Repeat:
Compare:
RepeatForever::create(ProgressTo::create(2, 100)); - work well
BUT
RepeatForever::create(ProgressTo::create(2, 70)); - not working how do you want.

minggo:
Yep. If you want to repeat from 0 to 70, it can not work correctly.
But does it a common requirement?

I am not sure if we should fix it like this or not, i have to ask @ricardoquesada  why he did it like this.

minggo:
@ricardoquesada
What's your opinion?

ricardoquesada:
Zifter is right regarding that we there is a bug in the code.
But Minggo is right about backwards compatibility.

But in v3.1 we already fixed on bug that breaks backward compatibility (with the particles). We fixed the code and also added a tool to fix "broken" .plist files.

So, yes, our code should be correct, and zifter is right about it.

So, what is the best way to tell our users that we are making this change?
Release Notes? 
What else?

minggo:
If we want to fix it, i think we can mentioned it in
- CHNAGELOG
- release note
- post news for it
- cocos2d-js group

And because it will just affect the situation that use `Repeat` and `Progress from 0 to 100` together. So i think it is not a big deal.

ricardoquesada:
Ok. sounds good to me.

minggo:
I mentioned it [here](https://github.com/cocos2d/cocos2d-x/blob/v3/CHANGELOG#L2). Please correct me if it is not correct.

zifter:
Great!
Thanks for the answer!

PR #6744, opened 2014-05-14
Fix compilation error using arch i386

NatWeiss:
Adding a member variable to hold the `placeholderAttributes` property fixes a compilation error regarding synthesis when building with ARCH=i386.


minggo:
[ci rebuild]

PR #6748, opened 2014-05-15
use RGBA8 as default pixel format

minggo:


ricardoquesada:
Any particular reason for changing this? thanks

minggo:
Use png as resource is common. Now it use RGB565, not alpha channel.
And developers, such as Hao, said they can not get alpha channel value.

Or other pixel format you suggest to set as default?

ricardoquesada:
RGBA8 is slower and takes more memory, but higher definition.
RGA565 is faster, less memory, but lower definition.

It is important to understand the implications of this change.
I'm not for or against this change.

What I want is that each change that we make in cocos2d-x (even if it is very small) is documented and tested.

A few years ago, this little change was making the games ~20% slower.
Perhaps now, with faster devices, it is not anymore an issue. But we have to test it before making this change.

minggo:
Yep.
And all other templates: lua, js are all use RGBA8. And i remembered, cpp template used RGBA8 too.

ricardoquesada:
so, have you done any performance tests? thanks!

minggo:
No.
Because developers can change it if they don't want it. It is not part of core code.
It is the same as template, should we include all modules or just include 2d?

So i think our template should consider most developers, especially average developers need. 

What's your opinion?

ricardoquesada:
- 99% of the developers won't change the default template configuration (that's why you have made this change)
- we must do performance tests for everything, even for changes in the templates

Wu-Hao:
99% of the developer doesn't change texture format, that is they load all their images with RGBA8888, taking 32bit memory, then they draw that 32bit on to the screen, and ended up with 16bit color. 

This is just wrong, it is wasting memory, as you could load your images with similar color bit such as 565, 5551, 4444

If 565 is 20% better performing than 8888, then it feels like cheated, as you compromise the quality of the final product by 25500%; (RGB565 colors = 16bit =  65536) over (RGB888 = 24bit = 16777216)

565 also produces significant color bandings, it is visible on EarthWarrior3D when run on iOS, it is destroying the images on my current project
yl vh wnuz_dsudwz1fu x

Sure 16bit images performs better, and that is why we should be encouraging our developers to use 16bit colors for png "Where Possible", it works great until you have color gradient on your images.
Then we should consider change the API regarding loading texture, currently we load all textures as 8888 doesn't even care if the image is 565, or 8bit greyscale, it converts everything to 8888.
the only API to change it is to "Texture2D::setDefaultAlphaPixelFormat(Texture2D::PixelFormat::RGB5A1);"
it is both horrendously long and its purpose is to set the "Default" so you have to set it back other wise all images loaded afterwards are stored using that format. which is not friendly at all

ricardoquesada:
@Wu-Hao 

I haven't read cocos2d-x Texture2d.cpp, but it seems that you are right.
We are taking weird decisions... as an example, why are we generating Labels as RGBA8888 instead of IA88 :-(
(BTW, A8 textures are as expensive as RGBA8888... at least on iOS).

I have just added a task in the roadmap to review/fix Texture2D.cpp

I don't disagree with you regarding the pros and cons of having RGBA8/RGB565 color buffer.
What I want is that we test each feature that we add into the engine... even if the feature only affects the templates.

So, here goes my question again :-) Have we done performance test before making this change? Thanks!

ricardoquesada:
so, any update on this? thanks.

minggo:
Sorry. I haven't done the performance test. I will do it before releasing 3.1 final.

ricardoquesada:
ok. thanks.

ricardoquesada:
Any update on this?
If we can't do performance test today we should rollback this change.

minggo:
Ok.
Let rollback it.
I have not time to do performance test before 3.1

minggo:
But lua template uses RGBA8 too.

PR #6756, opened 2014-05-15
Fix twice calling on exit

zifter:
http://www.cocos2d-x.org/issues/5091
onExit() method calling twice for running scene.
Now, for running scene onExit() calling only in setNextScene method.


minggo:
Thanks.

PR #6760, opened 2014-05-15
[ci skip], modify build-job termination time.

shujunqiao:
[ci skip], modify build-job termination time.


heliclei:
@dumganhar : review OK.

PR #6773, opened 2014-05-16
Fix CCPhysicsSprite::setPosition

NatWeiss:
As it is currently, if you call `CCPhysicsSprite::setPosition` it will set the physics body's position without setting the `CCSprite`'s position. This pull request fixes the issue.


zilongshanren:
@NatWeiss  AFAK, the sprite's position will be set when physic body update their position. So this  might won't be a big problem. And once more, the CCPhysicSprite is deprecated now we provide the physics integration option.
So I will choose to close this PR. 
Thanks.

PR #6774, opened 2014-05-16
V3 mac secure input

Mazyod:
Hi!

The main feature of this pull request is the implementation of `NSSecureTextField` on Mac, so now CCEditBox shows a password field when `InputFlag::PASSWORD` is set:

Another important change is the removal of `CCCustomNSTextField` class. It was apparently copied from the iOS implementation, but `UITextField` and `NSTextField` are **vastly** different.

Two other minor changes exist, fixing the `init` method to be according to the Objective-C convention, and removing the call `self.textField = NULL`, since it's dangerous to call setters in dealloc. Instead, the ivar should be released manually.

Best,
Maz

### DIFF
--- a/cocos/base/CCPlatformMacros.h
+++ b/cocos/base/CCPlatformMacros.h
@@ -164,6 +164,10 @@ public: virtual varType get##funName(void) const { return varName; }
 protected: varType varName;\
 public: virtual const varType& get##funName(void) const { return varName; }
 
+#define CC_SYNTHESIZE_READONLY_PURE_REF(varType, varName, funName)\
+protected: varType varName;\
+public: virtual varType& get##funName(void) { return varName; }
+
 /** CC_SYNTHESIZE is used to declare a protected variable.
  We can use getter to read the variable, and use the setter to change the variable.
  @param varType     the type of variable.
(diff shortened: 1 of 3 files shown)


Mazyod:
Oh, and I added a `CCLOGWARN` to the other `InputFlag` options, so the developer can see in the console why setting the flags doesn't work.

minggo:
It looks great.
Thanks.

Mazyod:
Oh crap, one of my other changes slipped in somehow. Sorry about that! I removed it.

minggo:
When i test it, i found

Extensions -> EditBoxTest 
No response when i click edit box.

Mazyod:
@minggo Weird, it's working fine for me:



If you mean the placeholder doesn't go away, that was already how it behaves on Mac. I can look into it later, since it's doesn't hurt my game as much.

I just switched back to the v3 branch and confirmed this behavior.

Mazyod:
@minggo I did a little more research, and the textfields I interacted with in apps like Evernote, GitHub client, .. etc, all have this behavior. The placeholder doesn't clear when you click inside the box.

minggo:
Thanks @Mazyod, i will try again.

minggo:
Yep. It can work.
Sorry about it.

Mazyod:
Awesome, thanks @minggo. You are controlling what all cocos2d-x users see. Better safe than sorry!

PR #6780, opened 2014-05-16
Bugfix autorelease pool

kezhuw:
First commit is a bugfix, others are code improvements.

### DIFF
--- a/cocos/base/CCAutoreleasePool.cpp
+++ b/cocos/base/CCAutoreleasePool.cpp
@@ -110,10 +110,8 @@ PoolManager* PoolManager::getInstance()
     if (s_singleInstance == nullptr)
     {
         s_singleInstance = new PoolManager();
-        // Add the first auto release pool
-        s_singleInstance->_curReleasePool = new AutoreleasePool("cocos2d autorelease pool");
-        s_singleInstance->_releasePoolStack.push_back(s_singleInstance->_curReleasePool);
     }
+    CC_ASSERT(s_singleInstance->getCurrentPool() != nullptr);
     return s_singleInstance;
 }
 
@@ -125,6 +123,8 @@ void PoolManager::destroyInstance()
 
 PoolManager::PoolManager()
 {
+    // Add the first auto release pool
+    push(new AutoreleasePool("cocos2d autorelease pool"));
 }
 
 PoolManager::~PoolManager()
--- a/cocos/base/CCAutoreleasePool.cpp
+++ b/cocos/base/CCAutoreleasePool.cpp
@@ -111,7 +111,6 @@ PoolManager* PoolManager::getInstance()
     {
         s_singleInstance = new PoolManager();
     }
-    CC_ASSERT(s_singleInstance->getCurrentPool() != nullptr);
     return s_singleInstance;
 }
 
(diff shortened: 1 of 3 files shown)


minggo:
Thanks. 
It seems great.

cocos/base/CCAutoreleasePool.cpp
```diff
@@ -125,55 +125,42 @@ void PoolManager::destroyInstance()
...
-    
-    while (!_releasePoolStack.empty())
-    {
-        AutoreleasePool* pool = _releasePoolStack.back();
-        _releasePoolStack.pop_back();
-        
-        delete pool;
-    }
+
+    std::for_each(_releasePoolStack.begin(), _releasePoolStack.end(), std::default_delete<AutoreleasePool>());
```
minggo, line 142: Why use `std::default_delete` instead of `delete`?

kezhuw, line 142: It is a side effect of replacement while block with std::for_each. Last parameter of std::for_each needs a function object, std::default_delete just fits it.

minggo, line 142: Got it.
Thanks.

I will merge it after releasing v3.1. Since this PR modify the core codes.

minggo, line 142: And we have released 3.1rc0, 3.1 final will be released end of this week.

minggo:
Will crash at begin when i test `cpp-tests` with `iPhone Retina (3.5-inch)` simulator.

kezhuw:
Fix it.

Two bugs exist in origin code:
1.  top level autorelease pool(heap-allocated) can be popped;
2.  duplication top level autorelease pool pushed(first in its constructor, second in PoolManager::getInstance()), double-delete occurs when destructing PoolManager.

I only see first early.

Commit ea4bf84 will abort origin code.

cocos/base/CCAutoreleasePool.cpp
```diff
@@ -110,9 +113,14 @@ PoolManager* PoolManager::getInstance()
...
-        s_singleInstance->_curReleasePool = new AutoreleasePool("cocos2d autorelease pool");
-        s_singleInstance->_releasePoolStack.push_back(s_singleInstance->_curReleasePool);
+
+        // Create top level autorelease pool.
+        //
+        // AutoreleasePool's constructor will push itself to PoolManager::getInstance().
+        auto topPool = new AutoreleasePool("cocos2d autorelease pool");
+        CC_ASSERT(s_singleInstance->_releasePoolStack.size() == 1);
+        CC_ASSERT(topPool == s_singleInstance->getCurrentPool());
+        (void)topPool;
```
minggo, line 123: What does this code mean?

cocos/base/CCAutoreleasePool.cpp
```diff
@@ -125,55 +133,39 @@ void PoolManager::destroyInstance()
...
     _releasePoolStack.push_back(pool);
-    _curReleasePool = pool;
 }
 
 void PoolManager::pop()
 {
     // Can not pop the pool that created by engine
-    CC_ASSERT(_releasePoolStack.size() >= 1);
-    
+    CC_ASSERT(_releasePoolStack.size() > 1);
```
minggo, line 166: May be we should add a log here, and return if there is only one pool.
Or `PoolManager::destroyInstance()` will cause error here.

minggo, line 166: May be should just output a log message here.
Because `PoolManage::getInstance` will create a new AutoReleasePool after `PoolManager::destroyInstance()` is invoked.

kezhuw:
I messed up. Try to solve it in new pull request.

Close this pull request.

PR #6804, opened 2014-05-19
fix UIText's fontSize

hyqhyq3:
if set fontsize before setting font family, fontsize was wrong


zilongshanren:
@minggo This PR could be merged into UI framework.

PR #6808, opened 2014-05-19
Cleanup 2d/platform library references

Mazyod:
Cleanup project file.
[REM]: old library references. New libraries are under prebuilt.


minggo:
Thanks.

PR #6816, opened 2014-05-19
[ci skip], modify request proxies in post-build.py and so on.

shujunqiao:
[ci skip], modify request proxies in post-build.py / job-comment-trigger.py / job-trigger.py / pull-request-builder.py and so on.


heliclei:
@minggo: review OK

PR #6820, opened 2014-05-19
Android project improvements

newnon:
release/debug builds from eclipse
native debug support

### DIFF
--- a/templates/cpp-template-default/proj.android/jni/Application.mk
+++ b/templates/cpp-template-default/proj.android/jni/Application.mk
@@ -1,2 +1,11 @@
 APP_STL := gnustl_static
-APP_CPPFLAGS := -frtti -DCC_ENABLE_CHIPMUNK_INTEGRATION=1 -DCOCOS2D_DEBUG=1 -std=c++11 -fsigned-char
+APP_CPPFLAGS := -frtti -DCC_ENABLE_CHIPMUNK_INTEGRATION=1 -std=c++11 -fsigned-char
+
+APP_DEBUG := $(strip $(NDK_DEBUG))
+ifeq ($(APP_DEBUG),1)
+  APP_CPPFLAGS += -DCOCOS2D_DEBUG=1
+  APP_OPTIM := debug
+else
+  APP_CPPFLAGS += -DNDEBUG
+  APP_OPTIM := release
+endif
--- a/build/android-build.py
+++ b/build/android-build.py
@@ -108,9 +108,9 @@ def do_build(cocos_root, ndk_root, app_android_root, ndk_build_param,sdk_root,an
 
     num_of_cpu = get_num_of_cpu()
     if ndk_build_param == None:
-        command = '%s -j%d -C %s %s' % (ndk_path, num_of_cpu, app_android_root, ndk_module_path)
+        command = '%s -j%d -C %s NDK_DEBUG=%d %s' % (ndk_path, num_of_cpu, app_android_root, build_mode=='debug', ndk_module_path)
     else:
-        command = '%s -j%d -C %s %s %s' % (ndk_path, num_of_cpu, app_android_root, ndk_build_param, ndk_module_path)
+        command = '%s -j%d -C %s NDK_DEBUG=%d %s %s' % (ndk_path, num_of_cpu, app_android_root, build_mode=='debug', ndk_build_param, ndk_module_path)
     print command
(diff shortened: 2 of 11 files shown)


templates/cpp-template-default/proj.android/jni/Application.mk
```diff
@@ -1,2 +1,11 @@
...
-APP_CPPFLAGS := -frtti -DCC_ENABLE_CHIPMUNK_INTEGRATION=1 -DCOCOS2D_DEBUG=1 -std=c++11 -fsigned-char
+APP_CPPFLAGS := -frtti -DCC_ENABLE_CHIPMUNK_INTEGRATION=1 -std=c++11 -fsigned-char
+
+APP_DEBUG := $(strip $(NDK_DEBUG))
+ifeq ($(APP_DEBUG),1)
+  APP_CPPFLAGS += -DCOCOS2D_DEBUG=1
+  APP_OPTIM := debug
+else
+  APP_CPPFLAGS += -DNDEBUG
+  APP_OPTIM := release
```
minggo, line 10: It seems great.
Now, how to build with debug mode, and how to debug with release mode?

Thanks.

newnon:
to build debug release just right click right mouse button from eclipse and switch `Build Configurations` - > `Release` / `Debug`
to debug c++ code switch to `Debug` configuration and in eclipse choose `Debug as` `Android Native Application`
to build Debug build run python build_native.py  -n NDK_DEBUG=1 
by default build_native.py make release build

```
 +$(call import-add-path,$(LOCAL_PATH)/../../cocos2d)
 +$(call import-add-path,$(LOCAL_PATH)/../../cocos2d/external)
 +$(call import-add-path,$(LOCAL_PATH)/../../cocos2d/cocos)
```

this 3 lines needed to debug application without it debugger cant find submodules

only 1 problem you cant debug C++ & Java code in one moment 
if you run `Debug as` `Android Native Application` you can debug C++ code
if you run `Debug as` `Android Application` you can debug Java code

minggo:
I debug mode by default is better.
Because
- it will not break compatibility
- we need release version when publishing

What's your opinion?

newnon:
Before it was not debug and not release mode, because without flag `NDK_DEBUG=1` android build in release mode and flag `-DCOCOS2D_DEBUG=1` just turn cocos debug.
I thing release node is better by default, and this param applies only to new project(i have just edited template only) old project keep old behavior. And if you need real debug for example for test, you need to add `NDK_DEBUG=1` - this will turn all native debug mechanisms like safe iterators, asserts and some other useful stuff.

And one more, even with debug ability Eclipse is not so friendly IDE most people write all code in XCode/AppCode/Visual Strudio and just compile their final build on Eclipse/Command line what why most people need Release not Debug by default.

minggo:
I think debug mode is required when developing. And we use debug mode as default for Xcode and VS project too. We want to developers to get assert error as earlier as possible. And we need the log message when developing. We will use `CCLOG` in the codes, which will trigger message in debug mode.

And if building with debug mode, can we use `ndk-stack` to translate error stack into readable stack?

newnon:
I don't know about 'ndk-stack' i haven't use it
but if you want debug be default i thing it good idea to fix python build script to use this logic
to use `build_native.py ndk -b release` it will be better way

newnon:
i will try to make this but we need also fix console

minggo:
I just don't understand the benefit of using release mode as default? What's the default behavior of NDK?

newnon:
Default behavior of NDK is release whats why we need to add NDK_DEBUG=1 to command line to use real debug mode. to make debug as default we need to make some changes to python script. I'll try to make this and make one more commit.

minggo:
Ok.
Then that's not problem.

newnon:
Done but best way also add this python changes to cocos console also

minggo:
Default mode of console is debug. I have asked @natural-law, who maintained console module.
And could you please also modify lua template and `build/android-build.py` too?
Thanks.

newnon:
ok i'll do it

minggo:
Thanks.

newnon:
do we need also fix test Application.mk for test?

minggo:
Thanks.

Yep. I think we need.
You are appreciated if you can help to do it too.

newnon:
Done

minggo:
Thanks.

PR #6842, opened 2014-05-20
don't have to compare in step since it is done in init function

minggo:
In `ActionInterval::initWithDuration()`, there is codes 

``` cpp
// prevent division by 0
    // This comparison could be in step:, but it might decrease the performance
    // by 3% in heavy based action games.
    if (_duration == 0)
    {
        _duration = FLT_EPSILON;
    }
```

But it also compares in `ActionInterval::step()` too.

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -120,9 +120,7 @@ void ActionInterval::step(float dt)
     }
     
     this->update(MAX (0,                                  // needed for rewind. elapsed could be negative
-                      MIN(1, _elapsed /
-                          MAX(_duration, FLT_EPSILON)   // division by 0
-                          )
+                      MIN(1, _elapsed /_duration)
                       )
                  );
 }


minggo:
@ricardoquesada 
Please review it.
It is performance improvement.

minggo:
@ricardoquesada 
Any problem of this PR?

ricardoquesada:
It is in my todo list. 
My todo-list is pretty big, and remember that last week I was 2 days away, and this week I'll be 2 days away too, so I might be able to review this PR in one or two weeks.

ricardoquesada:
As the comments on the code says, the `MIN/MAX` things are needed because actions could be reversed, and if so, it could have a negative value.

Also, each performance improvement should have a test that compares the speed.
For example, with this patch, how much performance do we gain ?

The code is not duplicated, since once test it at init time, and the other at update time.

minggo:
Reverse will not reverse `_duration`.
As the comment said, it do compare in init just not want to do in step to save performance.
But now it did in step too.

ricardoquesada:
questions:
- how much performance are we gaining with this patch ?
- have you tested the reverse actions (there is a test that reverses actions and particles) with your patch ? Did they work Ok?

minggo:
I have tested all actions test, particle test, and schedule test.
There is not problem.

Is there a requirement of performance improvement to remove duplicate codes?

minggo:
It will improve a bit, but i haven't tested. 
This pr is to remove duplicated codes, remove unneeded operation in step function. 
It is not just for performance improvement.

PR #6852, opened 2014-05-20
fix: Subclass of Scene can't call this->scheduleUpdate() with physics feature.

xhcnb:
Refactor Scene with physics feature.
If use 'update(float)' and 'scheduleUpdate' will cause subclass of Scene can't call 'this->scheduleUpdate()', because it has already called in super class. That's confused API design.
So, I think there should be an  separate update function for Physics Scene.

### DIFF
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -97,9 +97,8 @@ void Scene::addChild(Node* child, int zOrder, int tag)
     addChildToPhysicsWorld(child);
 }
 
-void Scene::update(float delta)
+void Scene::updatePhysicsWorld(float delta)
 {
-    Node::update(delta);
     if (nullptr != _physicsWorld)
     {
         _physicsWorld->update(delta);
@@ -131,7 +130,7 @@ bool Scene::initWithPhysics()
         this->setContentSize(director->getWinSize());
         CC_BREAK_IF(! (_physicsWorld = PhysicsWorld::construct(*this)));
         
-        this->scheduleUpdate();
+        this->schedule(schedule_selector(Scene::updatePhysicsWorld));
         // success
         ret = true;
     } while (0);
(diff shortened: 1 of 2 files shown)


minggo:
@xhcnb
Why subclass can't invoke `this->scheduleUpdate()`? What's the error behavior?

xhcnb:
@minggo 

Because `this->scheduleUpdate()` has already called in `Scene::initWithPhysics()`

Check following code

```
class Test: public Scene {
public:
    virtual bool init()
    {
        Scene::initWithPhysics();
        return true;
    }

    virtual void update(float dt)
    {
        //Scene::update(dt);
    }

    virtual void onEnter()
    {
        Scene::onEnter();
        this->scheduleUpdate();
    }

    CREATE_FUNC(Test);
};
```

It'll failed with `this->scheduleUpdate()`, so I renamed the `Scene::update` for physics feature.

minggo:
So the issue is that, invoking `scheduleUpdate` twice has problem.
Right?

I think we should fix the issue of `scheduleUpdate`.

xhcnb:
@minggo 
Yes, that's better.

But,

```
class Test: public Scene {
public:
    virtual bool init()
    {
        Scene::initWithPhysics();
        return true;
    }

    virtual void update(float dt)
    {
        Scene::update(dt);
        CCLOG("update %f", dt);
    }

    CREATE_FUNC(Test);
};
```

As `update` is a virtual function, I just want to enable physics and want to schedule `update` later. But, `Scene::initWithPhysics()` has scheduled it , so the behaviour is not what I expect.

So, I think rename `Scene::update` to `Scene::updatePhysicsWorld` for physics is more clear.

minggo:
Yep.
I think it is reasonable.

zilongshanren:
@xhcnb This PR is conflict with the upstream.
Could you fix the conflicts and end another PR?
Thanks.

zilongshanren:
Since a new physics 2d integration will be merged in v3.9 and this pull request still has conflicts, so I just close it here.

PR #6853, opened 2014-05-20
fix keycode_back  from  KEY_ESCAPE to  KEY_BACKSPACE

zhaijialong:
Keypad test in cpptest and luatest can't work , because of the KEYCODE_BACK defined incorrectly.

### DIFF
--- a/cocos/platform/android/jni/TouchesJni.cpp
+++ b/cocos/platform/android/jni/TouchesJni.cpp
@@ -77,7 +77,7 @@ extern "C" {
     
     
     static std::unordered_map<int, cocos2d::EventKeyboard::KeyCode> g_keyCodeMap = {
-        { KEYCODE_BACK , cocos2d::EventKeyboard::KeyCode::KEY_ESCAPE},
+        { KEYCODE_BACK , cocos2d::EventKeyboard::KeyCode::KEY_BACKSPACE},
         { KEYCODE_MENU , cocos2d::EventKeyboard::KeyCode::KEY_MENU},
         { KEYCODE_DPAD_UP  , cocos2d::EventKeyboard::KeyCode::KEY_DPAD_UP },
         { KEYCODE_DPAD_DOWN , cocos2d::EventKeyboard::KeyCode::KEY_DPAD_DOWN },


minggo:
In previous version, `KEY_BACK` is mapped to error key code.
So we should modify corresponding test.

zhaijialong:
That's OK. Although i think neither escape nor backspace is a good name for back key on android.

minggo:
Because desktop and mobile phones share the same key code, and `escape` is used as back menu on desktop.

minggo:
It is fixed in #6855.

zhaijialong:
ok, don't forget luatest. :)

iforce2d:
I wonder how many apps will be broken by this change? I agree with Joshua, that neither of these is a good name for the back key on Android... escape is used as back menu on desktop? Says who? But more troubling is how something as fundamental and commonly used as this can seemingly be changed on a whim, leaving us floundering around searching on Google for the solution to a new problem that suddenly showed up. There is no mention of this in the changelog. Is there somewhere I can find a list of all changes like this that break my app?

minggo:
@iforce2d 
I mentioned it [here](https://github.com/cocos2d/cocos2d-x/blob/v3/CHANGELOG#L17). 

Sorry about it. What's suitable name for it? We can add one new enum value, such as

``` cpp
KEY_BACK = KEY_ESCAPE
```

iforce2d:
I see. Looks like your comment was not included in the changelog file in the download I got from here:   http://cdn.cocos2d-x.org/cocos2d-x-3.1.1.zip
As long as it is mentioned somewhere so that people can keep up-to-date, that is good.

About the name for the enum value, KEY_BACK is not bad, but I think it could be made even more specific, eg. KEY_BACK_SOFTKEY_ANDROID which should completely dispel any ambiguities. I don't think KEY_ESCAPE is such a good idea, because even on Android there are third-party keyboards that can produce a real Esc keypress (like this one: https://play.google.com/store/apps/details?id=org.pocketworkstation.pckeyboard), and the app developer would not be able to tell the difference.

In the big picture, I don't see why the existing keyBackClicked() functionality needed to be removed. This is my main complaint actually, because if that functionality was kept, I would not have to search to discover the new method in the first place.

minggo:
@iforce2d 
Thanks for your suggestion.

Because we use EventDispatcher to dispatch all events, including keyboard event and Android key back event, so we remove keyBackClicked().

PR #6855, opened 2014-05-21
fix keypad test

zilongshanren:


minggo:
Please checkout lua test too.

PR #6862, opened 2014-05-21
Update UILoadingBar.h doxygen comments

favorcode:


favorcode:
UILoadingBar and UIPageView

PR #6865, opened 2014-05-21
WinRT V2 vs2103 cpp template fix

stammen:
This pull request fixes the VS2013 winrt cpp template by adding a missing force include pch.h setting.


dumganhar:
Thanks.

PR #6871, opened 2014-05-22
Fix Repeat will run one more over in rare situations

asuuma:
When times specified at `Repeat::create` is big, `_nextDt` in `Repeat::update` will may have some error that cause `Repeat` to run one more over than you specified.

Bug can be appeared in below code.

``` C++
    auto action = Sequence::createWithTwoActions(CallFunc::create([&]() {
        // some count up
    }),
                                                 DelayTime::create(1.0f / 50));
    this->runAction(Repeat::create(action, 50));
    // count up will be executed 51 times!
```

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -439,7 +439,7 @@ void Repeat::update(float dt)
 
             _innerAction->stop();
             _innerAction->startWithTarget(_target);
-            _nextDt += _innerAction->getDuration()/_duration;
+            _nextDt = _innerAction->getDuration()/_duration * (_total+1);
         }
 
         // fix for issue #1288, incorrect end value of repeat
--- a/tests/cpp-tests/Classes/ActionsTest/ActionsTest.cpp
+++ b/tests/cpp-tests/Classes/ActionsTest/ActionsTest.cpp
@@ -81,7 +81,8 @@ static std::function<Layer*()> createFunctions[] = {
     CL(Issue1288),
     CL(Issue1288_2),
     CL(Issue1327),
-    CL(Issue1398)
+    CL(Issue1398),
+    CL(Issue2599)
 };
 
 static int sceneIdx=-1;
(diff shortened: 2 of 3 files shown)


minggo:
Thanks.
I think this PR resolves the same issue as #6562.

minggo:
And i think #6562 is more easier to understand.

asuuma:
@minggo Thanks for reply.
But I think this problem is not same as #6562 because this issue is that sometime `_nextDt` will be less than `1.0f` when it should be.
For example, when `_innerAction->getDuration()/_duration` is just slightly less than `0.02f` because of error, with old code `_nextDt` will be like `0.9999995828` after 49 run, which will cause to run action 51 times.
But with new code, `_nextDt` will be `1.0f` in this situation, so action will run 50 times as desire.
This issue is made by error of adding float many times.

minggo:
Will `_nextDt` bigger than 1.0f?

ricardoquesada:
@minggo 
if are going to merge this PR, please include a test case for it. thanks!

minggo:
Sure.

asuuma:
@minggo 
I hope `_nextDt` will not be bigger than `1.0f` because `_duration = _innerAction->getDuration() * _times`, but I can't promise since I'm not much familiar with float error.
I tested and verifyed with some cases, but I think there are needs to test with many cases.
Thanks!

minggo:
@asuuma 
Could you add a test case for it to reproduce the issue?
Thanks.

asuuma:
@minggo 
That is ok.

asuuma:
@minggo 
I added test case to reproduce the issue.
With old code, it will show 51 even 50 times to repeat.

minggo:
Thanks.

gelldur:
Im not shure that this bug exist now but on branch 2.0 i reported issue with CCRepeat.
For example i have 2 actions in CCRepeat and one of them was sometimes executed few times more then second. Problem was in the floating value  i don;t remember now but i think value of repeat times was 80? 

Here it is: https://github.com/cocos2d/cocos2d-x/pull/5519 but i don't fix it in 100%

minggo:
@gelldur
`Repeat` can only receive one action. So your issue may caused by `Sequence` or `Spawn`. 
Could you please give some example?
Thanks.

@asuuma 
Yep, it fixed the issue.
Thanks.

gelldur:
Checkout forum thread http://www.cocos2d-x.org/forums/6/topics/46503

PR #6872, opened 2014-05-22
RootNode possible before calling the destructor has been deleted

HoGarfield:


zifter:
Good PR, thanks!

cocos/editor-support/cocosbuilder/CCBAnimationManager.cpp
```diff
@@ -99,7 +99,11 @@ Node* CCBAnimationManager::getRootNode()
 
 void CCBAnimationManager::setRootNode(Node *pRootNode)
 {
+    CC_SAFE_RELEASE_NULL(_rootNode);
+    
     _rootNode = pRootNode;
+    
+    CC_SAFE_RETAIN(_rootNode);
```
zilongshanren, line 106: You retain the rootNode here, I think it should be release somewhere, like the destructor method, right?

HoGarfield, line 106: Yes, you're right

zilongshanren, line 106: @HoGarfield  So could you add it in this PR?
 After it CI is ready, we could merge this patch.
Thanks.

HoGarfield, line 106: @andyque OK,but I can not add it now. I need some time to download the project and I need to wait until after work to do

zilongshanren, line 106: @HoGarfield  I think you should also add `CC_SAFE_RELEASE_NULL(_rootNode);` in CCBAnimationManager's destructor.

minggo:
Any update of this PR?

PR #5331, opened 2014-02-12
Develop 2.2.2

MatriQ:
modify to support CCInteger,CCFloat,CCDouble,CCBool in CCFileUtils::generateElementForObject

PR #6875, opened 2014-05-22
closed #5331: _particleIdx may not equal to _particleCount if developer ...

boyu0:
...invoke Particle::addParticle() at schedule.

If user use addParticle() at schedule, and if Particle::update priority is higher than the schedule, the added particle won't be initialed, and _particleIdx won't be updated, so _particleIdx may not equal to _particleCount at ParticleSystemQuad::draw().

### DIFF
--- a/cocos/2d/CCParticleSystemQuad.cpp
+++ b/cocos/2d/CCParticleSystemQuad.cpp
@@ -370,7 +370,6 @@ void ParticleSystemQuad::postStep()
 // overriding draw method
 void ParticleSystemQuad::draw(Renderer *renderer, const Mat4 &transform, bool transformUpdated)
 {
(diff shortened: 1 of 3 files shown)


minggo:
The problem is that:
- `ParticleSystem::addParticle()` will be invoked after `ParticleSystem::update`
- `_particleCount` is changed in `ParticleSystem::addParticle()` which leads to that `_particleCount` is bigger than `_particleIdx`

I think it is better to make `ParticleSystem::addParticle()` to protected.
What's your opinion @ricardoquesada ?

ricardoquesada:
We need a test case for this.
We can't add bug fixes without test cases.

And that's my opinion, because without a test case I don't know if the patch is correct or not.

boyu0:
@ricardoquesada test case added.

ricardoquesada:
@boyu0 thanks! your patch looks good to me.

PR #6876, opened 2014-05-22
Added 'optional' paramater to GLProgramState::setUniformTexture to manually specify textureUnitIndex.

sachingarg05:
Added 'optional' paramater to GLProgramState::setUniformTexture to manually specify textureUnitIndex.

Needed to specify textures in a different order, and when changing the texture to process using a shader.

### DIFF
--- a/cocos/renderer/CCGLProgramState.cpp
+++ b/cocos/renderer/CCGLProgramState.cpp
@@ -455,17 +455,20 @@ void GLProgramState::setUniformMat4(const std::string &uniformName, const Mat4& 
 
 // Textures
 
-void GLProgramState::setUniformTexture(const std::string &uniformName, Texture2D *texture)
+void GLProgramState::setUniformTexture(const std::string &uniformName, Texture2D *texture, int textureUnitIndex)
 {
     CCASSERT(texture, "Invalid texture");
-    setUniformTexture(uniformName, texture->getName());
+    setUniformTexture(uniformName, texture->getName(), textureUnitIndex);
 }
 
-void GLProgramState::setUniformTexture(const std::string &uniformName, GLuint textureId)
+void GLProgramState::setUniformTexture(const std::string &uniformName, GLuint textureId, int textureUnitIndex)
 {
     auto v = getUniformValue(uniformName);
-    if (v)
+    if (v) {
+        if (textureUnitIndex != -1)
+            _textureUnitIndex = textureUnitIndex;
         v->setTexture(textureId, _textureUnitIndex++);
+    }
     else
         CCLOG("cocos2d: warning: Uniform not found: %s", uniformName.c_str());
 }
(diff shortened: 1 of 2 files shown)


dabingnn:
Thanks for your pull request. But I am a little confused about you test case. Could you please give a more detailed explanation?
Generally, the purpose of GLProgramState is to hide the implementation details to our developer. User only need to know that what is the sampler in `shader` and what is the texture handle. GLProgramState will bind them automatically.
Given that you have used `CC_Texture0` and `CC_Texture1` in the shader, the following function call `setUniformTexture("CC_Texture0", texture1)` `setUniformTexture("CC_Texture0", texture2)` can be done in different orders. And GLProgramState deals with low level implementation and bind the `texture1` to `CC_Texture0` and `texture2` to `CC_Texture1` respectively.

sachingarg05:
I am generating dynamic textures and also applying shader effects using renderTextures. My use case is that I want to change the texture attached to a shader to use same shader on different textures.

The current GLProgramState code only increments the value of _textureUnitIndex and on each call to setUniformTexture it will attach the texture to next index - 1, 2, 3, 4, 5... There is no way to change the texture attached to index 1.

sachingarg05:
Every time I call setUniformTexture to change the attached texture, it is attached to next index due to _textureUnitIndex++.

(Then eventually the app crashes when _textureUnitIndex reaches 8 because my device allows max 8 texture units, even though I need to use only 1 texture in my shader).

dabingnn:
If I understand right, did you mean that, you have several sampler in `shader`, `CC_Texture0`, `CC_Texture1`, `CC_Texture2`, `CC_Texture3`....
At first, you set them all by calling setUniformTexture(...), it works. 
Then after a while, you have changed CC_Texture3, when you call `setUniformTexture('CC_Texture3', newTexture)`, it does not work. Right?

sachingarg05:
Yes. Setting the textures first time works. Problem is when changing the textures later.

If it is not good option to ask developer to specify index, another option can be to add a function:

GLProgramState::resetTextureUnitIndex() { 
  _textureUnitIndex = 0; 
}

Or GLProgramState will need to remember which uniform is attached to which textureUnitIndex and then reuse that value.

dabingnn:
Thanks very much, it is a bug in GLProgramState. I will add a ticket in our issue system.
For the solution, I prefer that GLProgramState remember the textureUnitIndex and reuse them. Because if we do it like this, user do not need to know the low level implementation details. What is your opinion about this?

sachingarg05:
Yes, that will be a good solution (but will need more work) :-)

dabingnn:
Don't worry! We will work on this. Thanks again for your good findings.
Here is the issue link in our website
http://www.cocos2d-x.org/issues/5393

minggo:
It is fixed in #7008.

sachingarg05:
Thanks

PR #6880, opened 2014-05-22
Added support for the recursive setting cascade mode for color and opacity.

zifter:
New functionality.
Can change cascade opacity and color recursively.
Fixed Layer test according with this change.

### DIFF
--- a/cocos/2d/CCLayer.h
+++ b/cocos/2d/CCLayer.h
@@ -216,14 +216,16 @@ public:
     virtual void setOpacity(GLubyte opacity) override { return Layer::setOpacity(opacity); }
     virtual void updateDisplayedOpacity(GLubyte parentOpacity) override { return Layer::updateDisplayedOpacity(parentOpacity); }
     virtual bool isCascadeOpacityEnabled() const override { return Layer::isCascadeOpacityEnabled(); }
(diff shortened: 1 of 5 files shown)


zifter:
PR need to rebuild to set the correct status.

zilongshanren:
@ricardoquesada  Could you review this PR?

ricardoquesada:
@andyque 
Sure. The code looks ok but:

What I need is that you review it first and post your comments here.
Things that you can do:
- make sure it conforms to the cocos2d-x C++ guidelines: https://github.com/cocos2d/cocos2d-x/wiki/Contributing-to-Cocos2d-x
- make sure that it has the needed tests

You have to be the first filter... and the things that you can't review for whatever reason, let me know.

you will notice that this PR doesn't have a test and adds optional argument that is banned according to our C++ guide:

two possible things to do:
- you can fix them yourself in your local repo before merging it (it will take you no more than 10 minutes)
- or ask the author to re-submit it.

ricardoquesada:
info about the default arguments here: https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#default-arguments

zilongshanren:
Thanks @ricardoquesada 
In the future, I will be the first filter...

zilongshanren:
Thanks @zifter 
Please view the comments from ricardoquesada and change your patch...

minggo:
@zifter Could you please fix the conflict? Sorry for late response of this PR, please add a comment said you changed the codes in future because there is not any notice when you add a new commit.

PR #6888, opened 2014-05-23
Feature/camera

yuri-ash:
Added Camera node.

How to:
Scene has a ptr directly on current camera.
When scene is visited, if camera ptr is present, then we push a custom command to the renderer to update the projection matrix.
After the scene is done parsing all its children, a custom command is created to pop the projection matrix.
CameraTest has been added in cppTest project. You can move and zoom the camera with your mouse .

Known issue:
- Renderer::checkVisibility doesn't work properly anymore, as the perspective view is directly modified.
- Zoom will zoom over the whole render, as only  one camera is active at a time. Even if the UI can stay still by adding it to the scene graph as a camera children, there is no such trick for a node to not be affected by zoom. - FIXED
- Input are not yet transform in camera view, so picking doesnt work properly. - FIXED (more patched than fixed, but it's working)

Improvement:
- add a node ViewCamera that will handle camera operation, allowing to have multiple camera targetting specific branch of the scene graph (ie: to one ViewCamera, we add a game scene, to an other view camera, we ad dour UI, both ViewCamera have different camera and both are children of scene. While moving a camera, the other render should stay unchanged) - DONE
- I haven't code a camera in a while, so computing of projection matrix might be a bit mixed with the view matrix.


minggo:
@yuri-ash 
We plan to add Camera in v3.3.
@ricardoquesada, does this PR fit our needs?

yuri-ash:
@minggo 
I saw that you were planning on adding it but i needed it right away for my project. I just submit this pull request if it can help you go a bit faster as I don't know your status on this feature.

mindryu:
@yuri-ash 
I've find this. 
thank you.

PR #6903, opened 2014-05-27
add constness to ui framework

zilongshanren:


cocos/ui/UILayout.h
```diff
@@ -40,8 +40,8 @@ class LayoutProtocol
     virtual ~LayoutProtocol(){}
     
     virtual LayoutExecutant* createLayoutExecutant() = 0;
-    virtual const Size getLayoutContentSize() = 0;
-    virtual const Vector<Node*>& getLayoutElements() = 0;
+    virtual const Size getLayoutContentSize()const = 0;
```
minggo, line 43: It seems meaningless to return `const Size`.

minggo, line 43: Do you want to return `const Size&`?

PR #6910, opened 2014-05-27
Bug fix cc scheduler blocked element

gelldur:
Previous PR https://github.com/cocos2d/cocos2d-x/pull/5732

---

Example

We scheduled teget for example method update

void MyObject::update(float dt)
{
//Here CCScheduler::m_bUpdateHashLocked is set to true
//Also here scheduler for LeafAnimator::getInstance() is
paused!

CCDirector::sharedDirector()->getScheduler()->unscheduleUpdateForTarget(&LeafAnimator::getInstance()
);

CCDirector::sharedDirector()->getScheduler()->scheduleUpdateForTarget
(&LeafAnimator::getInstance(), INT_MAX, false );
//So when we want reschedule paused field isn't updated. Priority also
but this is in todo :)
}

### DIFF
--- a/cocos2dx/CCScheduler.cpp
+++ b/cocos2dx/CCScheduler.cpp
@@ -469,7 +469,7 @@ void CCScheduler::scheduleUpdateForTarget(CCObject *pTarget, int nPriority, bool
         CCAssert(pHashElement->entry->markedForDeletion,"");
 #endif
         // TODO: check if priority has changed!
-
+		pHashElement->entry->paused = bPaused;
         pHashElement->entry->markedForDeletion = false;
         return;
     }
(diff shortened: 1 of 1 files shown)


minggo:
@gelldur 
I can not understand the issue clearly. Could you please describe it in more detail? Thanks.

gelldur:
Sorry i write it to fast:)

Simply bug is with that if i schedule and unschedule something state of the 
    pHashElement->entry->paused = bPaused;

isn't set propertly. I mean when you delete scheduler it is paused and 
    pHashElement->entry->markedForDeletion
is set to true but when you want add it again you will set only 
    pHashElement->entry->markedForDeletion 

to false but you don't unpause it i mean set the new value as you pass it.

PR #6916, opened 2014-05-28
Fix mingw64 build: convert from LPVOID to DWORD losses precision on 64-b...

John-He-928:
Fix mingw64 build: convert from LPVOID to DWORD losses precision on 64-bit platforms.
Should be DWORD_PTR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,7 +153,7 @@ include_directories(
   ${CMAKE_CURRENT_SOURCE_DIR}/external/xxhash
 )
 
-if(WIN32)
+if(WIN32 AND NOT MINGW)
   include_directories(
     ${CMAKE_CURRENT_SOURCE_DIR}/external/${PLATFORM_FOLDER}-specific/gles/include/OGLES
     ${CMAKE_CURRENT_SOURCE_DIR}/external/${PLATFORM_FOLDER}-specific/icon/include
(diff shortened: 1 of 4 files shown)


minggo:
@Dhilan007
Could you please checkout if it has problem on win32.
Thanks.

John-He-928:
Tested on mingw64, cpp-test finished normally. Building on mingw32 still failed due to `__stdcall` mangling on libglew32.a (undefined reference to 'glewInit'), which I'm trying to fix. (I don't have Visual Studio).

> A DWORD_PTR is an unsigned long type used for pointer precision. It is used when casting a pointer to an unsigned long type to perform pointer arithmetic. DWORD_PTR is also commonly used for general 32-bit parameters that have been extended to 64 bits in 64-bit Windows. 

From what MSDN said, DWORD_PTR is 32-bit on Win32, which is the same as DWORD, so Win32 should be fine. When casting LPVOID to DWORD on Win64, gcc failed to compile and reported an error. I don't know whether Visual Studio will report an error or warning, or just fix it for us.

John-He-928:
**Fix mingw32/mingw64 build:** 
Prebuilt glew library provides header files that don't match the system libglew32.a (installed via pacman). Prebuilt headers have function prototypes without `__stdcall` directives but system libs have `__stdcall`. This makes linking to libglew32.a fails. Such issue doesn't break mingw64 builds, I don't know why.

Also, we have to link libws2_32 manually on mingw platform. GCC doesn't do this for us.

By now, mingw32 can make a successful build and run cpp-empty-test and cpp-tests.

PR #6920, opened 2014-05-28
don't have to set _orderOfArrival to 0 after visit

minggo:
The discuss is in #6138.
@ricardoquesada 
Would you please review it?

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1002,9 +1002,6 @@ void Node::visit(Renderer* renderer, const Mat4 &parentTransform, bool parentTra
         this->draw(renderer, _modelViewTransform, dirty);
     }
 
-    // reset for next frame
-    _orderOfArrival = 0;
- 
     director->popMatrix(MATRIX_STACK_TYPE::MATRIX_STACK_MODELVIEW);
 }
 


minggo:
Any update?

ricardoquesada:
thanks. I forgot about this one.
do you have a test case to reproduce it ? that is the best way to verify that the patch is correct,

minggo:
It is easy to understand this issue. And there is discussion about this issue in #6138. Have you read it?

Yehsam23:
should remove all "setOrderOfArrival(0);" in other files?

like Label.cpp
visit function have
setOrderOfArrival(0);

it will cause incorrect zorder when CCNode's _orderOfArrival not reset

minggo:
@Yehsam23
Did you mean this PR has problem?
If so, then could you please describe it more clearly or could you add a test case?
Thanks.

Yehsam23:
@minggo 

You can put a Sprite and Label at same Position...then addchild again!!
The label will reorder under Sprite

Sprite *testSprite = Sprite::create(xxxx);
testSprite->setPosition(Vec2(200, 200));
this->addChild(testSprite);

Label *testLabel = Label::createWithSystemFont("123", "Arial", 30);
testLabel->setPosition(Vec2(200, 200));
this->addChild(testLabel);

Then create a menu~~ when menu clicked

void menuclicked(xxx xx)
{
    Node *anyNode = Node::create();
    this->addChild(anyNode);
}

Then you can't see label....because the label is under testSprite.

Try it...

And setOrderOfArrival(0); also in CCSpriteBatchNode

Thanks

minggo:
I have tested in `cpp-empty-test`, Label doesn't disappear.
And could you explain why `_orderOfArrival` need to be set to 0 after `visit`?

Yehsam23:
我的意思是CCNode _orderOfArrival 沒有設成0

可是Label _orderOfArrival會在visit後 設成0

會有問題

是不是Label.cpp 也要拿掉setOrderOfArrival(0);

不拿掉的話 會有上面的情況發生

Yehsam23:
@minggo 

you can try

https://dl.dropboxusercontent.com/u/20253231/HelloWorldScene.cpp

replace with your cpp-empty-test

and click menu....Label will disappear

you should remove all "setOrderOfArrival(0);" in lib project

minggo:
@Yehsam23 
Yep. setOrderOfArrival(0) should be removed in Label too.
Thanks.

minggo:
@Yehsam23 
I fixed it in #7135.

reddevillg:
ProtectedNode, Armature and some other class still has:

```
 _orderOfArrival = 0
```

minggo:
Thanks.
We will modify it.

toddor:
There are 4 more lines with "sprite->setOrderOfArrival(0);" in CCSpriteBatchNode.cpp. After removing them all my problems with wrong zorder reordering disappeared.

minggo:
@toddor 
What's your problems?
How to reproduce them?

toddor:
@minggo , When I have more than 29 children in the batchnode and I add or remove a new child, the first child ( i.e. child.at(0) ) becomes the last child and renders in front of the others.
- I'm adding them without setting zOrder to addchild().
- I tried to increase the capacity (because of that number 29) but it didn't help.
  Only when removing sprite->setOrderOfArrival(0); it stopped reordering the batchnode children

PR #6922, opened 2014-05-28
prevent flickering when opening another activity

zawasp:
taken from cocos2d v2 branch.
prevents flickering when opening anothe activity from cocos2d game.


minggo:
Thanks.

PR #6925, opened 2014-05-29
[hot fixed]issue#4999:fixed label is unsharp if it's created by smaller font.

WenhaiLin:


ricardoquesada:
Hey guys, it is important to create a test case for this.
In fact for each fixed bug, we need a test case. Otherwise we don't know if the bug was fixed correctly, and we don't know if the bug will appear again in the future.

WenhaiLin:
Agree.I will create a test case for that ASAP.

PR #6927, opened 2014-05-29
fixed bug of cc_utf8_to_utf16() when length!=-1

TimothyZhang:
cc_utf8_to_utf16 doesn't work correctly if length!=-1 and the given char\* is not zero-terminated or contains zero.

### DIFF
--- a/cocos/base/ccUTF8.cpp
+++ b/cocos/base/ccUTF8.cpp
@@ -248,7 +248,8 @@ unsigned short* cc_utf8_to_utf16(const char* str_old, int length/* = -1*/, int* 
     unsigned short* ret = nullptr;
     
     std::u16string outUtf16;
-    bool succeed = StringUtils::UTF8ToUTF16(str_old, outUtf16);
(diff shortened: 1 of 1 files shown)


cocos/base/ccUTF8.cpp
```diff
@@ -248,7 +248,8 @@ unsigned short* cc_utf8_to_utf16(const char* str_old, int length/* = -1*/, int*
     unsigned short* ret = nullptr;
     
     std::u16string outUtf16;
-    bool succeed = StringUtils::UTF8ToUTF16(str_old, outUtf16);
+    std::string inUtf8 = length == -1 ? std::string(str_old) : std::string(str_old, length);
```
minggo, line 251: But the [comment](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/ccUTF8.h#L191) said `length` is not used any more.

TimothyZhang, line 251: Why not use `length` any more? That will cause a lot of memory copies, since strings are not 0-terminated in many cases.

minggo:
[ci rebuild]

PR #6937, opened 2014-05-29
CCUserDefault.cpp: Should not compile on Mac platform

wefiends:
On the Mac, the CCUserDefault.mm implementation should be used (it already checks for the Mac platform). That way NSUserDefaults can be used.

On a related note, why is KEEP_COMPATABILITY defined at the top of CCUserDefault.mm? Will this be removed in future releases? Otherwise, the only way to use NSUserDefaults is to remove the define. Should this setting be moved to Configuration or ccConfig?


minggo:
Yep, i think we can remove it now.

minggo:
KEEP_COMPATABILITY is used for the purpose
- in previous we use xml to save data for all platforms
- then we use system API to record data on iOS and on Android, because using xml may lost data

minggo:
So we use this macro just for the users that update from older version, may be v2.1.x.

minggo:
I think it is reasonable to move the macro to `ccConfig.h`.

minggo:
What's suitable macro name for it?

wefiends:
Maybe CC_ENABLE_NATIVE_USERDEFAULT and set it enabled by default. You want me to send you a PR?

PR #6938, opened 2014-05-29
Simplify BlendFunc operator< judgement

CaryZheng:
Simplify BlendFunc operator< judgement, and the older method of comparation is superfluous.


dabingnn:
Thanks, this is a little confusing.
My suggestion is  to change it to `src < a.src || (src == a.src && dst < a.dst)`. 
How do you think about it?

CaryZheng:
Yeah, you're right and I've modified it.

dabingnn:
Thanks

PR #6941, opened 2014-05-30
Add a callback interface in FileUtils/Image for decrypt data.

WenhaiLin:
1.Add a callback interface in FileUtils/Image for decrypt data.
2.Add a test case for decrypt data.


WenhaiLin:
@minggo Please review it.

PR #6946, opened 2014-05-30
Skip UTF8 BOM if it exists and bug fix for unterminated string

hanjukim:
- Skip UTF8 BOM if it exists
- Fix when string isn't terminated if fread() reads less size than paramter at FileUtils getData


minggo:
Thanks.

PR #6948, opened 2014-05-31
Add immediate send request

litefeel:


Mazyod:
@lite3 This is a great feature! I never realized how `HTTPClient` sends requests..

btw, I think there is a lot of duplicate code and it should be consolidated.

minggo:
@lite3 

@Mazyod 
What's the duplicate code?

litefeel:
@minggo ok, i will add test case at this weekend.

the duplicate code may be `switch (request->getRequestType())` section.

minggo:
@lite3 
Thanks.

And do you have any idea to consolidate it?

@Mazyod, any idea?

litefeel:
at this weekend, i will send on new pr.

minggo:
Thanks.

litefeel:
@minggo  please review.

Mazyod:
I'll throw in my 2 cents:
1. Naming the function `immediateSend`, from experience, will make it harder to discover. Most developers use autocomplete, and it would be more visible if it had the same `send` prefix. **Solution:** Call it `sendImmediately(...)` instead.
2. Duplicate code can easily be merged in a private method, and I personally think duplicate code is the source of all evil.

Of course, otherwise, as I said this is looking great!

litefeel:
@Mazyod ok, i changed.

litefeel:
@Mazyod @minggo i wanna change
 `typedef std::function<void(HttpClient* client, HttpResponse* response)> ccHttpRequestCallback;`
 to 
`typedef std::function<void(HttpResponse* response)> ccHttpRequestCallback;`,
 remove the HttpClient\* parameter, because HttpClient is singleton, and it  become wild pointer  in sendImmediate callback.

Mazyod:
@lite3 I agree, I haven't found much use of the `HTTPClient` pointer myself, and as you said it is a singleton, and can be accessed if necessary.

However, since this is a public API change, I don't know how the Cocos2d-x developers would approach it.

litefeel:
@Mazyod so, I did not change it immediately.  I want to ask everyone.

minggo:
@lite3 
@Mazyod 
About modifying API, we always  deprecated it and add new one to keep compatibility.

cocos/network/HttpClient.cpp
```diff
@@ -234,15 +172,44 @@ void HttpClient::networkThread()
...
+void HttpClient::networkThreadAlone(HttpRequest* request)
+{
+    // Create a HttpResponse object, the default setting is http access failed
+    HttpResponse *response = new HttpResponse(request);
+    char errorBuffer[CURL_ERROR_SIZE] = { 0 };
+    processResponse(response, errorBuffer);
+
+    auto scheduler = Director::getInstance()->getScheduler();
+    scheduler->performFunctionInCocosThread([response]{
+        auto request = response->getHttpRequest();
```
minggo: Could you please use another name? Because `request` is the name of passed in parameter.

litefeel: @minggo But they are in different scopes.

Mazyod:
@lite3 Yeah, they are, but it may confuse some developers in the future. It's common practice to avoid variable shadowing as much as possible.

litefeel:
[ci rebuild]

minggo:
It is not your fault.
Thanks.

I have tested it. I worked as expected.
I will merge it.

@Mazyod @lite3 
Is there any problem?

Mazyod:
I think it's great!

minggo:
Thanks.

PR #6951, opened 2014-06-02
3.1 hotfix

colin3dmax:


minggo:
@colin3dmax 
It is fixed in #6939.

PR #6957, opened 2014-06-03
fix scale9Sprite can't culling problem

zilongshanren:


minggo:
Ok.
Would you also please modify Hungary style too?

PR #6959, opened 2014-06-03
Bugfix autorelease pool manager

kezhuw:
Continue with pull request #6780.

Commit dbca937 will break AutoreleasePool test case.
Other commits fix it.

It may not be a bug, but a 'feature'. If that, explicit code/documentation should be added.

TestCase AutoreleasePool passed.


minggo:
@kezhuw 
Thanks.
But this PR doesn't including the modification of #6780.

PR #6960, opened 2014-06-03
fix layout constant syntax error

kicktheken:


samuele3hu:
This syntax  error can lead to load DeprecatedEnum.lua file error,Thanks.

minggo:
Oh, could you please send it to v3.
v3.1-hotfix just used to releasing v3.1.1 which is released.

PR #6965, opened 2014-06-04
Support Fast App Resume and fixed black screen may arise when using third sdk.

WenhaiLin:


WenhaiLin:
@stammen Please review it.

stammen:
This is an excellent technique that I did not know about. Go ahead and pull the request.

PR #6967, opened 2014-06-04
fix lua-binding release build

ntotani:
when release mode, compile error like this

```
[armeabi] Compile++ thumb: cocos_lua_static <= lua_cocos2dx_manual.cpp
/Users/n_totani/test/frameworks/runtime-src/proj.android/../../cocos2d-x/cocos//scripting/lua-bindings/manual/lua_cocos2dx_manual.cpp: In function 'int lua_cocos2dx_GLProgramState_setVertexAttribPointer(lua_State*)':
/Users/n_totani/test/frameworks/runtime-src/proj.android/../../cocos2d-x/cocos//scripting/lua-bindings/manual/lua_cocos2dx_manual.cpp:5387:49: error: 'tolua_err' was not declared in this scope
             if (tolua_isnumber(tolua_S, -1, 0, &tolua_err))
                                                 ^
```

To fix it, check number only debug mode.


minggo:
Thanks.

PR #6976, opened 2014-06-05
Behaviour of ParallaxNode is wrong

JochenHeizmann:
Behaviour of ParallaxNode is wrong when glFrameSize is not identical to DesignResolutionSize. 

Don't know if this should be fixed using multiplying with a matrix - but at least at my tests it works with this fix. But would be great if someone who knows more about the internal rendering will have a look at this problem.

To test the problem:
- Create a cocos2d-x app
- In AppDelegate create a GLView with a different size to designResolution, for example:

[...]
        glview = GLView::createWithRect("testApp", {0,0, 960, 512});
        director->setOpenGLView(glview);
[...]
    glview->setDesignResolutionSize(480, 512, ResolutionPolicy::FIXED_WIDTH);
[...]
- Then create some ParallaxNodes and add it to your scene and scroll it around in the update-method

### DIFF
--- a/cocos/2d/CCParallaxNode.cpp
+++ b/cocos/2d/CCParallaxNode.cpp
@@ -121,7 +121,7 @@ The positions are updated at visit because:
 - using a timer is not guaranteed that it will called after all the positions were updated
 - overriding "draw" will only precise if the children have a z > 0
 */
-void ParallaxNode::visit(Renderer *renderer, const Mat4 &parentTransform, bool parentTransformUpdated)
+void ParallaxNode::visit(Renderer *renderer, const Mat4 &parentTransform, uint32_t parentFlags)
 {
     Vec2 pos = parallaxPosition;
 
@@ -134,7 +134,7 @@ void ParallaxNode::visit(Renderer *renderer, const Mat4 &parentTransform, bool p
         }
         _lastPosition = pos;
     }
-    Node::visit(renderer, parentTransform, parentTransformUpdated);
+    Node::visit(renderer, parentTransform, parentFlags);
 }
 
 NS_CC_END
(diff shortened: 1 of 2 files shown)


JochenHeizmann:
Hmmm... seems like my fix only works under certain circumstances. Will have a deeper look into it.

JochenHeizmann:
Ok, I have updated this pull request and made some changes:

I've added a separate vector for the parallax offset. This way the user can position the ParallaxNode (that acts like a container for the sub nodes) in normal cocos style for example:

myParallaxNode->setPosition(x, y);

This won't affect the parallax offset itself. It acts like any other cocos node. To create the parallax effect the user has to change the parallax offset/position:

myParallaxNode->setParallaxPosition(x, y);

If he updates this for example every frame he will get the typical parallax effect. The user is able to scale the parent layers or the parallax node layer itself and everything works like a charm.

Another advantage is that the absolute position is required anymore so this should also improve performance a little bit (not tested).

Let me know what you think.

boyu0:
Can you describe this bug with more detail?  I use the code you wrote in description and run it in mac, I didn't find any wrong behaviors.
Can you provide more code about this bug? and what platform you use? thanks.

JochenHeizmann:
Here is a project that reproduces the wrong behaviour.

I'm adding two nodes to my ParallaxNode. One has a ParallaxFactor of {0.0, 0.0}, so shouldn't move at all. But when I scale the ParallaxNode it does.

https://dl.dropboxusercontent.com/u/5061704/ParallaxTest.zip

minggo:
@boyu0 
Any update of this pull request?

boyu0:
I think it's a good idea to use setParallaxPosition for parallax, use setPosition for the real position like other nodes, but it broke the compatibility, so I think is not that appropriate to merge it, I think right now we only fix the parallax wrong behaviour bug.
@minggo What's your opinion?

minggo:
@dabingnn 
Could you please take a took?
Thanks.

minggo:
close this PR since it has conflict and no other developers met and report the same issue for so a long time

PR #6979, opened 2014-06-05
Texture Optimize

visiblelight:
1. remove unused codes;
2. use the internal image format instead of the global format when loading an image into texture;


cocos/renderer/CCTexture2D.cpp
```diff
@@ -535,11 +535,6 @@ bool Texture2D::initWithData(const void *data, ssize_t dataLen, Texture2D::Pixel
     mipmap.address = (unsigned char*)data;
     mipmap.len = static_cast<int>(dataLen);
     return initWithMipmaps(&mipmap, 1, pixelFormat, pixelsWide, pixelsHigh);
-
-    //update information
-    _contentSize = contentSize;
-    _maxS = contentSize.width / (float)(pixelsWide);
-    _maxT = contentSize.height / (float)(pixelsHigh);
```
dabingnn, line 542: Why do we need to delete these code?

visiblelight, line 542: I think we can safely remove those codes since they are never invoked and the relevant variables are all actually set at otherwhere.

minggo:
[ci rebuild]

minggo:
Please reuse existing pictures.

PR #6986, opened 2014-06-05
Fix CCImage to be compatible with UNIX2003

iSevenDays:
Without this wrapper I get errors below.
Mac Os X 10.9.3 Xcode 6 beta(the same problem also exists in Xcode 5)
I found a solution at http://stackoverflow.com/questions/8732393/code-coverage-with-xcode-4-2-missing-files and added one more wrapping function.

Undefined symbols for architecture i386:
  "_fwrite$UNIX2003", referenced from:
      _unixErrorHandler in libcocos2dx iOS.a(tif_unix.o)
      _unixWarningHandler in libcocos2dx iOS.a(tif_unix.o)
      _empty_output_buffer in libcocos2dx iOS.a(jdatadst.o)
      _term_destination in libcocos2dx iOS.a(jdatadst.o)
      _Fax3PrintDir in libcocos2dx iOS.a(tif_fax3.o)
      _PredictorPrintDir in libcocos2dx iOS.a(tif_predict.o)
  "_strerror$UNIX2003", referenced from:
      _TIFFOpen in libcocos2dx iOS.a(tif_unix.o)
ld: symbol(s) not found for architecture i386
clang: error: linker command failed with exit code 1 (use -v to see invocation)

### DIFF
--- a/cocos/platform/CCImage.cpp
+++ b/cocos/platform/CCImage.cpp
@@ -39,6 +39,7 @@ THE SOFTWARE.
 
 extern "C"
 {
+#if defined(__APPLE__) || defined(__unix) 
 #ifndef __ENABLE_COMPATIBILITY_WITH_UNIX_2003__
 #define __ENABLE_COMPATIBILITY_WITH_UNIX_2003__
 #include <stdio.h>
@@ -54,6 +55,7 @@ extern "C"
     {
         return strerror(errnum);
     }
+#endif
 #endif
 #include "png.h"
 #include "tiffio.h"
--- a/cocos/platform/CCImage.cpp
+++ b/cocos/platform/CCImage.cpp
@@ -39,6 +39,22 @@ THE SOFTWARE.
 
 extern "C"
 {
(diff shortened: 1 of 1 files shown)


minggo:
I don't have any problem with Xcode 5 and Xcode 6.
And i haven't heard other developer have this issue.

minggo:
Oh, i met the problem on Xcode 6 today. But it is not problem yesterday.
I am not sure if we should fix it like this.
May be some configuration is changed that cause the problem.
Because it can work yesterday.

minggo:
I found some doc about this issue [here](http://www.macosxguru.net/article.php?story=2009010813350633).

But i think this PR is better. Because we don't want to recompile lib, because it may cause problem when Xcode update again.

minggo:
Yesterday, i built for iPhone5S, which is 64 bit, so it didn't have problem as http://stackoverflow.com/questions/8732393/code-coverage-with-xcode-4-2-missing-files said.

minggo:
And i added some comment in #6988.

iSevenDays:
Thank you for accepting this commit.
I use iPhone4s.

PR #6994, opened 2014-06-06
faster tiled map implementation.

boyu0:
DO NOT MERGE IT!
For review. Still has some bugs.
Classes need rename.
Doesn't support vertex z yet.


minggo:
@ricardoquesada 
Could you please help to review it too?
Thanks.

ricardoquesada:
sure. I already reviewed it.
The code looks good, although I haven't tested it.
So, once the tests pass, it is safe to merge it.

boyu0:
Unfinished work now:
1. Vertex Z support.
2. Multi platform project configuration.
3. Classes rename.
4. Maybe some bugs I haven't found yet.

Now the deference of API between the new tmx and old tmx are:
1. releaseMap(), new tmx doesn't need this.
2. API in SpriteBatchNode, new tmx inherit from Node, so all the API in SpriteBatchNode no longer available.
3. some protect functions.

If we rename new tmx to TMXLayer and TMXTiledMap, the developers will meet some compatibility problem. besides, I doesn't add vertex Z support yet..

But we need a name, I think TMXLayer2 and TMXTiledMap2 is not a good name....

@ricardoquesada What's your opinion of this?

boyu0:
@ricardoquesada 
For the vertex Z support, I discussed with @dabingnn, he said it's hard to support different z for one command call, and he think we can use sprite for each tile, it will automatic culling, and our engine will do auto batch, it will make not too much difference for performance with do culling and quad draw by ourselves, and it can support vertex Z very well. He suggest us do some performance test for this.

I will do performance test for this.

What's your opinion?

boyu0:
@ricardoquesada @dabingnn
I loaded iso-test-zorder.tmx multiple times( 10 for android, 50 for ios ), test it both with an action( for transform updated) and without an action, with all these circumstances, the performance are almost the same( same GL verts, same GL calls, and almost the same fps ), so I think we can use sprite instead of quad, it's easier to use and control.

@ricardoquesada What's your opinion?

dabingnn:
This is a good news. Could you please attach a result of your performance test?

boyu0:
Sorry, I made a mistake in the performance test, I will retest it, and commit the code to my repo.

boyu0:
Sorry, The test results what I wrote above is wrong.

I wrote the performance test code here:
https://github.com/boyu0/cocos2d-x/tree/fast_tmx_test ( use quad for render )
https://github.com/boyu0/cocos2d-x/tree/fast_tmx_sprite_test ( use sprite for render )

The data for performance test:

-| test NO. | GL verts | GL calls | fps
-- | -- | -- | -- | --
ios( sprite ) |  1 | 132318 | 10 | 8
ios( quad ) |  1 | 135396 | 327 | 14~26
ios( sprite ) |  2 | variation | 7~10 | 6~12
ios( quad ) |  2 | variation | variation | 14~36
ios( sprite ) |  3 | 12954 | 8 | 21
ios( quad ) |  3 | 77184 | 327 | 60
ios( sprite ) |  4 | 486 | 7 | 24
ios( quad ) |  4 | 486 | 7 | 60
android( sprite ) |  1 | 17370 | 8 | 6
android( quad ) |  1 | 17376 | 47 | 11
android( sprite ) |  2 | variation | 8 | 6~28
android( quad ) |  2 | variation | 47 | 10~38
android( sprite ) |  3 | 5466 | 8 | 30
android( quad ) |  3 | 24570 | 87 | 35
android( sprite ) |  4 | 486 | 7 | 47
android( quad ) |  4 | 486 | 7 | 55

Each of them load `TileMaps/iso-test-zorder.tmx` multiple times.
test NO. 1 draws the most part of the tile maps.
test NO. 2 draws the most part of the tile maps and with an action ( update the transform ).
test NO. 3 draws a little part of the tile maps.
test NO. 4 load the tile maps but not shown at the screen.

boyu0:
So base on the results, use quads for render is about 30%~40% faster than use Sprite.
@dabingnn thoughts maybe our engine has some performance bottleneck in renderer...
I will continue this PR, find a way to support vertex Z with quads...

stevetranby:
I'm about to start in on our own renderer, what do you think the time frame will be? I can share any code or tests I write as well. I don't think a performance test should include the loading of the map as most people won't care how long it takes to load, but rather only how long it takes to render.

I don't fully understand the new render pipeline, but couldn't each quad just have it's z coordinate set to _positionZ instead of 0? That'll be my first test unless you guys finish it up first. 

Either way I'll definitely help test your updated code if you want.  

Thanks for all the work you guys have done!

darkdukey:
@stevetranby we thought about that using z, but the if we do that we'll have [z fighting issue|](http://en.wikipedia.org/wiki/Z-fighting) when sprite have the same z which is unfortunately, most common case in 2d game. But anyway I think current renderer still have a lot of room for improvement.

stevetranby:
Appreciate the commits @boyu0 as it is now mostly working for me. Still have some hidden layer/tiles along with incorrect horizontal flips, but those are likely all issues with my code that uses the tilemap.

stevetranby:
I updated `FastTMXLayer::updateTiles()` for our needs to force all tiles into Sprite instances to allow for other sprites to draw correctly at z positions in-between tiles in the tilemap. 

```
// HACK: this is hack to force all tiles into sprite instances
getTileAt(Vec2(x,y));
int tileGID = _tiles[x + y * (int)_layerSize.width];
```

Prior to this I tried a couple failed attempts to add glEnable(GL_DEPTH_TEST) and a couple other small changes, but nothing seemed to allow other sprites in between isometric tiles other than forcing sprite instances for all non-empty tiles.

minggo:
#7211 based on this PR, and improve a bit.

PR #6995, opened 2014-06-07
Multitouch support improvements

newnon:
getAllTouches like in cocoa

### DIFF
--- a/cocos/base/CCEventTouch.h
+++ b/cocos/base/CCEventTouch.h
@@ -50,6 +50,7 @@ public:
 
     inline EventCode getEventCode() const { return _eventCode; };
     inline const std::vector<Touch*>& getTouches() const { return _touches; };
+    inline const std::vector<Touch*>& getAllTouches() const { return _allTouches; };
 
 #if TOUCH_PERF_DEBUG
     void setEventCode(EventCode eventCode) { _eventCode = eventCode; };
@@ -59,6 +60,7 @@ public:
 private:
     EventCode _eventCode;
     std::vector<Touch*> _touches;
+    std::vector<Touch*> _allTouches;
 
     friend class GLViewProtocol;
 };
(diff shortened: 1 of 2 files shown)


minggo:
Any doc for `getAllTouches` of cocoa?
Thanks.

newnon:
https://developer.apple.com/library/ios/documentation/uikit/reference/UIEvent_Class/Reference/Reference.html#//apple_ref/occ/instm/UIEvent/allTouches

minggo:
Thanks.

minggo:
I think touches information dispatched by EventDispatcher::dispatchEvent() already includes all touches information.

Because when a touch is released, it will be removed in `g_touches`. 
So i don't know what's requirement of `getAllTouches`.

newnon:
No EventDispatcher get information only on changed touches not all active touches. And if you make system of touch recognizing you need to now about all touches on screen. I think it not so expensive function, because it copy only pointers to new vector but it adds useful functionality missing from cocoa.(and also from cocos2d)

minggo:
It is not expensive in `touchBegan` and `touchEnded`, but it is expensive in `touchMove`.
Because these informations are not needed for everyone. 
So i think it is better just add `getAllTouches`, but not invoked in all touch event handler. 
Developers invoke it if needed.

newnon:
It is not expensive even in touchMove because maximum touches count can be 5 and copy 5 pointers is not so expensive operation.

heliclei:
[ci rebuild]

newnon:
[ci rebuild]

minggo:
It will create object, do computation.
And we know it is not needed each time.
Though it is not expensive, i think we should make engine as fast as possible.

newnon:
i have made new PR https://github.com/cocos2d/cocos2d-x/pull/7225
just stand alone getAllTouches()

PR #6996, opened 2014-06-07
move node from one parent to another fix

newnon:


minggo:
Thanks.
Could you add comment for added codes?
Thanks.

newnon:
i don't understand where to place comment?

minggo:
Ignore me.
No problem.

PR #6998, opened 2014-06-07
3.1 hotfix

bontey:


minggo:
[ci rebuild]

minggo:
It is fixed in latest codes.

PR #7019, opened 2014-06-09
fix issue on Xcode 6

minggo:


minggo:
It is to fix warnings.

PR #7037, opened 2014-06-11
closed #5537: add setAutoStep, isAutoStep and step method for PhysicsWorld

boyu0:
fixed-timestep is useful for physics, so we need provide step method that developers can call it by themselves...


boyu0:
test case added.

karikuvaja:
@boyu0 

The idea of having callback from physics engine before it's step, is to have physics engine's behaviour independent from game's framerate. It provides more stable physics simulation.

Consider this scenario, where game physics runs at fixed timestep:

gameUpdate
gameUpdate
gameUpdate
physicsUpdate
gameUpdate
physicsUpdate
gameUpdate

If users will add forces to their bodies at game update, they need to have delta time taken into account in their calculations. Sometimes there might be multiple game updates before a physics update, or sometimes other way around, there might be couple of physics updates before a game update (there might then be a physics update without any forces added by the user at game update, which might be a problem), depending on the device's capabilities. 

Best way to get stable physics is to add forces before physics engine ticks, which is the motivation for physics callback. Unity handles physics this way, I also personally think it's a good practice. There might be engines that handle this differently, I am not aware of them, it might be worth a bit of research.

boyu0:
@karikuvaja 
If I understand correctly, our physics step should work like this:

```
PhysicsWorld::update(float delta)
{
     _elapsed += delta;
     while( _elapsed - _fixedStepTime >= 0)
    {
          preStepCallback(_fixedStepTime);
          physics->step(_fixedStepTime); 
          postStepCallback(_fixedStepTime); 
          _elapsed -= fixedUpdateTime;
    }
}
```

Hmm, It's a good idea, I will consider about this...

karikuvaja:
@boyu0 
Yes, that's pretty much what I mean. I can't think of use for postStepCallback from the top of my head, but it might have some purpose.

Just a small side note about testing the fixed timestep feature. If it's implemented correctly, you should get pretty much deterministic simulation on your test cases independent of framerate. You should be able to simulate ahead for example 30 seconds on a single update  (which might be actually really handy for game developers for some cases), or just let simulation run for 30 seconds normally, and get same result. 

Cheers,
Kari

PR #7038, opened 2014-06-11
Fix GL state cache inconsistency on texture delete with multi-texturing

BodbDearg:
When deleting a texture check if it is bound against any of the available texture units rather than just assuming it is bound to the first unit, which might not be the case if we are using multi-texturing. Failure to do this can result in inconsistencies between the cocos2d-x GL state cache and the internal GL driver state, resulting in rendering issues. OpenGL will unbind the deleted texture from any texture units it is bound to on the call to glDeleteTextures(), therefore we should perform the same behaviour in the state cache.

### DIFF
--- a/cocos/renderer/ccGLStateCache.cpp
+++ b/cocos/renderer/ccGLStateCache.cpp
@@ -171,10 +171,17 @@ void deleteTexture(GLuint textureId)
 void deleteTextureN(GLuint textureUnit, GLuint textureId)
 {
 #if CC_ENABLE_GL_STATE_CACHE
-	if (s_currentBoundTexture[textureUnit] == textureId)
+    
+    // Need to check if the texture is still bound against any of the available texture units and mark it as unbound.
+    // Internally GL will do this if the texture is deleted so our state cache needs to be kept consitent with the internal GL state.
+    for (size_t i = 0; i < MAX_ACTIVE_TEXTURE; ++i)
     {
-		s_currentBoundTexture[textureUnit] = -1;
+        if (s_currentBoundTexture[i] == textureId)
+        {
+            s_currentBoundTexture[i] = -1;
+        }
     }
+
 #endif // CC_ENABLE_GL_STATE_CACHE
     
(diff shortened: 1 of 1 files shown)


BodbDearg:
This issue caused problems for us in our game when we were using multi-texturing and resulted in invalid/black textures; cocos2d-x would not bind a texture to the 2nd texture unit because it incorrectly assumed it was already bound. It happened following scene switches when textures were destroyed and recreated. Could be a pretty complex problem to try and reproduce with a test case but I can confirm this fix works in the case of our game; at worst the only side effect of the change would be to induce an extra redundant GL state change very occasionally.

BodbDearg:
Cocos2d-iphone also appears to have the same issue, the 2d-x bug appears to be inherited from the original cocos2d implementation:

https://github.com/cocos2d/cocos2d-iphone/blob/v3.0/cocos2d/ccGLStateCache.m

dabingnn:
I like your ideas.
Current implementation is a little confusing. Because if you want to delete a texture, it does not matter whether it is bound to a texture unite or not. If it is bound to a texture unit, we need to change the current bound textureID to 0 or -1.
How about if we implement it like this?

```
void deleteTextue(GLuint textureId)
{
    for (size_t i = 0; i < MAX_ACTIVE_TEXTURE; ++i)
  {
       if (s_currentBoundTexture[i] == textureId)
       {
           s_currentBoundTexture[i] = -1;
        }
  }
}
```

and deprecate function `void deleteTextureN(GLuint textureUnit, GLuint textureId)`, and let it call `deleteTexture` internally?

dabingnn:
Hi, this is the pull request link. Could you please review it? 
https://github.com/cocos2d/cocos2d-x/pull/7049
Thanks

BodbDearg:
Hi dabingnn, thanks for taking a look at this change!

I agree - the idea of deprecating deleteTextureN() definitely makes sense since the texture unit parameter no longer has any purpose. Reviewed #7049 - the change looks good!

dabingnn:
Thanks

PR #7039, opened 2014-06-11
issue #4693: add getName(), setName() and getChildByName()

minggo:


minggo:
I will deprecate `getTag()`, `setTag()` and `getChildByTag()` in other PR.
@ricardoquesada 
Would you please review it?

ricardoquesada:
The code is Ok, but it is not complete.
We need an API more flexible, like SpriteKit's

https://developer.apple.com/library/ios/documentation/SpriteKit/Reference/SKNode_Ref/Reference/Reference.html#//apple_ref/doc/uid/TP40013023-CH1-SW36

This PR has part of the code: https://github.com/cocos2d/cocos2d-x/pull/5671
But what's missing is support for regular expressions in the search.

Please, don't merge this PR until the feature is complete. Thanks.

minggo:
I will not merge it.

I don't see the implementation of SpriteKit will search its children recursively in [enumerateChildNodesWithName:usingBlock](https://developer.apple.com/library/ios/documentation/SpriteKit/Reference/SKNode_Ref/Reference/Reference.html#//apple_ref/doc/uid/TP40013023-CH1-SW36). 

Should we search it recursively as #5671 did?

minggo:
@ricardoquesada 
`enumerateChildren` is added.
Please review again.
Thanks.

ricardoquesada:
could you please add some tests ? thanks.

minggo:
Sure. I will add test cases if the API seems ok.

ricardoquesada:
The API seems Ok, but the best way to test they API is by looking at examples... real use cases.
Also, a PR cannot be merged (even if the API is OK) without test cases.

minggo:
I agree.
I will add test case in this PR.

ricardoquesada:
ok. thanks.

minggo:
Test case was added.

minggo:
[ci rebuild]

ricardoquesada:
Thanks for adding more test cases Minggo.
But we need to add more like:

``` c++
// enumerate all the children from scene
enumerateChildren("/scene/*", ....);

// enumerate all children from "layer" that start with sprite
enumerateChildren("/scene/layer/sprite*", ....);

// enumerate all some grandchildren
enumerateChildren("child/*", ....);
```

Please, take a look at this as an example:
https://developer.apple.com/library/ios/documentation/GraphicsAnimation/Conceptual/SpriteKit_PG/Nodes/Nodes.html#//apple_ref/doc/uid/TP40013043-CH3-SW17

minggo:
Now `enumerateChildren` doesn't support advanced research.
I will read the doc.

ricardoquesada:
Yes, I noticed that :)
Regular expressions is the killer feature here. The good news is that regexp is part of C++11.

minggo:
Yep. I know it.
But the implementation of SpriteKit is more than just match the name with regular expression. It is more about the tree structure.

minggo:
I mean the regular expression is something about tree structure.
I will do the same thing as it did.
It seems great.

ricardoquesada:
ok. great!

minggo:
Regular expression is supported:)
And i have added corresponding tests.
Now you can use it like this
- enumerateChildren("name")
- enumerateChildren("[[:alnum:]]")
- enumerateChildren("/name")
- enumerateChildren("//name")
- enumerateChildren("name/name2")
  ...

ricardoquesada:
great!

minggo:
[ci rebuild]

pandamicro:
The api name `enumerateChildren` seems to be a little bit complicated to me, why not `findChildren` ?
I really like the idea, it's like jQuery's selector

dingpinglv:
This feature is cool. 
I agree with @pandamicro , `findChildren` is better than `enumerateChildren`.

minggo:
I will merge it if not problem.
Then deprecated getChildByTag() in another pull request.

minggo:
@pandamicro
@dingpinglv
The name comes from SpriteKit.
And it is to find a child.
It will enumerate children, then invoke a call back when a name matched.

minggo:
It is not to find a child.

pandamicro:
@minggo why not just return all children matches the condition? Isn't callbacks over designed here?

minggo:
You can return false to match all nodes.

minggo:
If call back function returns false, it will continue to match next one.
It is a observer design pattern.

darkdukey:
I agree with @pandamicro 
We should have a function `Vector<Node*> findChildren(name)` which will return all the children with the same name, I can see it been really useful in game development and more intuitive to use in some case, we can keep `enumerateChildren` they can play different role

minggo:
Ok.
Let's add it in another pull request. It is easy to implement based on `enumerateChildren`.

minggo:
#7218 was added to add findChildren().
I don't think we should add it in `Node`, it is a helper function, so i added in `utils`.

darkdukey:
Why not, the children belongs to the `Node`. It's natural to have it there.

PR #7040, opened 2014-06-11
updated angle to fix memory leak when closing angle

stammen:
This pull request updates cocos2d-x for WP8 to fix a memory leak when the phone switches between apps.

The updated Angle libs were submitted in the following pull request

https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pulls


minggo:
@stammen 
Thanks.

Are you still focus on v2?
What's the progress of v3?

PR #7052, opened 2014-06-12
Fix video scale issue in iOS

gin0606:


heliclei:
[ci rebuild]

minggo:
@gin0606 
How to reproduce the issue?

gin0606:
@minggo oops, Sorry.

Please check this code. https://gist.github.com/gin0606/5f254961555587b5122e
I want to show video that fill of screen.
But this code show me double scale video.

↓↓↓I attach my iPhone5 screen shots.↓↓↓

# Before

Before

# After

After

gin0606:
@minggo Hi! 
Something to do?

minggo:
@gin0606
Sorry about it.
I will test it ASAP.

minggo:
I tested test case in tests on iPhone 5S. Not problem found.
The test case is `Node::UI -> GUI Dynamic Create Test -> VideoPlayerTest`.

gin0606:
@minggo Sorry for my late reply.
I take long time to find this problem the cause, and I found it.

`cocos2d_tests->cpp-tests` app's `director->getContentScaleFactor()` is setting to `1` that it is SD Resolution. And set to `2` if HD Resolution.
this is same as `[[UIScreen mainScreen] scale]`.
When this state, it will not see this problem.

But my cocos2d-x app's `director->getContentScaleFactor()` is setting to `1` that it is HD Resolution.
My cocos2d-x app does not suppot SD Resolution.

So I think VideoPlayer's frame size calculation have to use CCEAGLView's `contentScaleFactor`. It is not director's contentScaleFactor.

Can you look this PullRequest again?

minggo:
content scale factor is different from glview content scale factor in cocos2d-x. 
contentScaleFactor = resource resolution / design resolution

gin0606:
@minggo srry, I know it.
I want to say that VideoPlayer's frame size calculation have to use glview's contentScaleFactor.

minggo:
I just want to know how to reproduce the issue?
Could you please describe it step by step?

gin0606:
https://github.com/gin0606/VideoPlayerIssue

@minggo I create app to reproduce the issue.
Please see it.

I can not reproduce the issue from "cocos2d_tests.xcodeproj".
This issue do not reproduce in  "cocos2d_tests.xcodeproj".

# Steps
1. Create cocos2d project.(exec `cocos new VideoPlayerIssue -p me.gin0606.videoPlayerIssue -l cpp -d ./`)
2. Add video file from `cocos2d-x/tests/cpp-tests/Resources/cocosvideo.mp4`
3. Implemant to show videoPlayer code.
4. Bulid&Run **Retina** iPhone or Simulator, and tap `mp4 play` label.

minggo:
Thanks.
We will check it.

gin0606:
@minggo :bow:

WenhaiLin:
Thanks @gin0606
@minggo I think @gin0606 is right,we should merge this PR.

PR #7054, opened 2014-06-12
add TimelineAction

2youyou2:
TimelineAction is used for decode exported data from CocoStudio mac version.
It's a better solution for Armature.

TimelineAction inherits from cocos2d:;Action, so it's more closed to cocos2d.
TimelineAction's struct is more flexible and clear than Armature.
Frame -> Timeline -> TimelineAction
Frame is easy to extend its type, you just need to inherit from Frame, and implement Frame::onEnter, Frame::apply.

TimelineActionCache will load data from json resource and cache it.
When creates a new TimelineAction, TimelineActionCache will clone one from cache.

NodeReader will create a node tree from json resourece.

To run a TimelineAction, you need to do like this: 

Node\* node = NodeReader::getInstance()->createNode("TimelineAction/boy_1.ExportJson");
TimelineAction\* action = TimelineActionCache::getInstance()->createAction("TimelineAction/boy_1.ExportJson");

node->runAction(action);
action->gotoFrameAndPlay(0, 60, true);


minggo:
Two questions:
- Is it possible merge timeline action into engine Action system?
- Is it possible to reuse existing images?

2youyou2:
1.TimelineAction contains both NodeReader and TimelineAction, not a single Action. So if we want to merge timeline action into Action system, do we need to separate them.
1. It's possible to reuse existing images. 
   Each Timeline have an action tag, and NodeReader will create an TimelineActionData for each loaded Node as it's UserObject. 
   When node run a timeline action, timeline action will search node's all descendants.
   If descendants have a TimelineActionData, and could find Timelines have the same action tag with this TimelineActionData's action tag, then these Timelines will do action to the nodes.
   So if existing images have TimelineActionData as it's UserObject with right action tag, it's possible to reuse them

minggo:
Separate action from reader is a better design. Then this action can be reused.

2youyou2:
The reader is already seperated now.
We have two readers, NodeReader for node, TimelineActionCache for TimelineAction.
Because TimelineActionCache have the capacity to cache TimelineAction, so it named cache instead of   reader

minggo:
So is it possible to separate action from reader?
If it is possible, please modify the codes.

Is it possible to reuse images?
If it is possible, please modify it.

Thanks.

cocos/2d/CCTimelineAction.h
```diff
@@ -0,0 +1,159 @@
...
+#include "2d/CCNode.h"
+#include "2d/CCAction.h"
+#include "2d/CCTimeLine.h"
+#include "base/CCRef.h"
+#include "renderer/CCRenderer.h"
+
+NS_CC_BEGIN
+
+typedef void (Ref::*SEL_FrameEventCallFunc)(Frame *);
+#define frameEvent_selector(_SELECTOR) (SEL_FrameEventCallFunc)(&_SELECTOR)
```
minggo: Please use std::function<> instead.

cocos/2d/CCTimelineAction.h
```diff
@@ -0,0 +1,159 @@
...
+    virtual bool isPlaying();
+
+    /** Set the animation speed, this will speed up or slow down the speed. */
+    virtual void  setTimeSpeed(float speed);
+    /** Get current animation speed. */
+    virtual float getTimeSpeed();
+
+    /** duration of the whole action*/
+    virtual void setDuration(int duration) { _duration = duration; }
+    virtual int  getDuration() { return _duration; }
```
minggo: virtual int  getDuration() -> virtual int  getDuration() const.

cocos/2d/CCFrame.h
```diff
@@ -0,0 +1,334 @@
...
+    static TextureFrame* create();
+
+    TextureFrame();
+
+    virtual void setNode(Node* node);
+
+    virtual void onEnter(Frame *nextFrame) override;
+    virtual Frame* clone() override;
+
+    inline void setTexture(std::string texture) { _texture = texture;}
```
minggo: setTexture  -> setTextureName?

cocos/editor-support/cocostudio/ActionTimeline/CCActionTimelineCache.cpp
```diff
@@ -0,0 +1,374 @@
...
+    ZOrderFrame* frame = ZOrderFrame::create();
+
+    int zorder = DICTOOL->getIntValue_json(json, Value);
+    frame->setZOrder(zorder);
+
+    return frame;
+}
+
+}
+}
```
minggo, line 374: Please add a new line at the end of file.

tests/cpp-tests/Classes/ExtensionsTest/CocoStudioActionTimelineTest/ActionTimelineTestScene.h
```diff
@@ -0,0 +1,88 @@
...
+};
+
+class TestTimelinePerformance : public ActionTimelineTestLayer
+{
+public:
+    virtual void onEnter();
+    virtual std::string title() const override;
+};
+
+#endif  // __ANIMATION_SCENE_H__
```
minggo, line 88: Please add a new line at the end of file.

PR #7056, opened 2014-06-13
Use RGBA when saving a file from render texture

visiblelight:


cocos/2d/CCRenderTexture.cpp
```diff
@@ -409,29 +409,47 @@ void RenderTexture::visit(Renderer *renderer, const Mat4 &parentTransform, uint3
     _orderOfArrival = 0;
 }
 
-bool RenderTexture::saveToFile(const std::string& filename)
+bool RenderTexture::saveToFile(const std::string& filename, bool isRGBA)
 {
-    return saveToFile(filename,Image::Format::JPG);
+    std::string basename(filename);
+    std::transform(basename.begin(), basename.end(), basename.begin(), ::tolower);
```
minggo, line 415: Why change to lower case?
It will be different from user input.

visiblelight, line 415: For the case "test.PNG" should be regarded as a PNG file.
The saved filename will be kept as the original one.

PR #7061, opened 2014-06-13
add autoscale for director

archetipo:
Autoscale method useful if you work with ,fixed size or position.
If is set true in AppDelegate before startup app, setup the scale factor of main director and extend it for all object .
Tried in Android 3.0 ~ 4.4 , landscape view.

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -65,7 +65,7 @@ THE SOFTWARE.
 
 /**
  Position of the FPS
- 
+
  Default: 0,0 (bottom-left corner)
  */
 #ifndef CC_DIRECTOR_STATS_POSITION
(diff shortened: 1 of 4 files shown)


minggo:
Could you please explain this PR?
I don't understand what it wants to do.
Thanks.

archetipo:
hi ,
i think that is my mistake ,for not having read in-depth documentation , because i have found this http://www.cocos2d-x.org/wiki/Multi_resolution_support#Exact-fit , that is exactly what I wanted to do.. 
so my PR,is useless, thanks a lot for cocos2d-x is a wonderful engine!!

minggo:
Thanks.
So can i close this PR?

archetipo:
yes thanks you too

PR #7062, opened 2014-06-14
Update _bPlaying on CCActionObject.

maldewar:
The method CCActionObject::isPlaying() returns the state of the member var _bPlaying,
which is initially declared false and never gets updated. The change updates this variable
across the play, pause, stop and end states of the action.


chengstory:
I think it can be merged!  Thanks @maldewar

minggo:
@maldewar 

maldewar:
I've fixed the conflicts @chengstory @minggo, thanks

minggo:
Thanks @maldewar.

PR #7064, opened 2014-06-14
Fix bug for TextReader #setFontName.

kyokomi:


cocos/editor-support/cocostudio/WidgetReader/TextReader/TextReader.cpp
```diff
@@ -51,7 +51,10 @@ namespace cocostudio
         bool fn = DICTOOL->checkObjectExist_json(options, "fontName");
         if (fn)
         {
-            label->setFontName(DICTOOL->getStringValue_json(options, "fontName"));
+            std::string tp_c = jsonPath;
```
zilongshanren: please don't use raw `char*` to represent string, use std::string instead.

And the label->setString method also requires a std::string as argument.

because std::string is better than raw `char*`, it is the best practice used in cocos2d-x v3.x.

thank you.

kyokomi: @andyque merged, thanks.

zilongshanren:
@minggo This PR could be merged now. Thanks.

PR #7068, opened 2014-06-15
Fix UIButton.cpp added TTFFont.

kyokomi:


cocos/ui/UIButton.h
```diff
@@ -43,6 +43,12 @@ class Button : public Widget
     DECLARE_CLASS_GUI_INFO
     
 public:
+    enum class Type
```
zilongshanren: Suppose one day I want to support new types of Button, the `Type` enum would be perfect for this purpose.

please move this enum class to private part and change `Type` to `FontType`.

Thank you.

kyokomi: @andyque merged, thanks.

zilongshanren:
@minggo  This PR could be merged now.

PR #7078, opened 2014-06-16
Optimize FPS control algorithm under Android.

reckhou:
The algorithm should work fine under Android now.

### DIFF
--- a/cocos/platform/android/CCApplication.cpp
+++ b/cocos/platform/android/CCApplication.cpp
@@ -68,7 +68,16 @@ int Application::run()
 
 void Application::setAnimationInterval(double interval)
 {
-    // NYI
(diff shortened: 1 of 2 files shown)


minggo:
I tested it on Galaxy Note 2, but not effect.
I added `director->setAnimationInterval(1.0 / 20);` in AppDelegate.cpp.
But the fps is also about 60.

reckhou:
I'm using a Galaxy S3 and it works fine. I test these code on v2, unless v3 is not using Cocos2dxRenderer.java anymore, it must be working.

reckhou:
I've solved this issue, since Application::setAnimationInterval() is not implemented in v3 so it's not working.

minggo:
Thanks.
I will try it tomorrow.

PR #7086, opened 2014-06-16
using fmodf for repeatForever

dabingnn:


dabingnn:
I wonder I can find one, but it is hard. 
The bug is triggered if there is a lag on our game, which will use a large `dt` for update. If the `dt` is much larger than `_innerAction->getDuration()`. 
`float diff = _innerAction->getElapsed() - _innerAction->getDuration();`
The diff could also be larger than `_innerAction->getDuration()`.

Or shall we use a `sleep` function, enlarging the `dt` to trigger it?

ricardoquesada:
yes, don't know how to reproduce it, but for each bug fix there must be a test that verifies that the bug fix is correct.
Perhaps by calling the action manually and faking the dt could work.

PR #7087, opened 2014-06-16
Wp8 v3 template and missing texture after app switch fixes

stammen:
The pull request fixes the following issues:
1.  added releaseGLTexture to CCTexture2D.h and .cpp. Cached textures need to make sure ALL of the cached textures on the CPU are released before attempting to create new textures on the GPU. Since all of the exisiting textures on the GPU have been wiped out by the device due to an app context switch, the existing gl Texture IDs are no longer valid.
2. Updated the cpp-templates.
3. Updated angle libs. https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/9


cocos/renderer/CCTextureCache.cpp
```diff
@@ -706,6 +706,12 @@ void VolatileTextureMgr::reloadAllTextures()
 {
     _isReloading = true;
 
+    // we need to release all of the glTextures to avoid collisions of texture id's when reloading the textures onto the GPU
+    for(auto iter = _textures.begin(); iter != _textures.end(); ++iter)
+    {
+	    (*iter)->_texture->releaseGLTexture();
```
WenhaiLin, line 712: Should we do the same on V2 branch?

PR #7108, opened 2014-06-18
fix compiling error

minggo:


zq5599966:
Using example c3t files can be normal to play 3d animation, but used 3dmax export FBX convert to c3t file cannot be read properly.
Could you tell me how to convert right c3t file can be read properly?
sorry for my engilsh
http://www.cocoachina.com/bbs/read.php?tid=209637

minggo:
@zq5599966 
Thanks for your feedback.
I will ask the 3d guys help to check it.

super626:
Thanks.

I will have someone check it.

PR #7124, opened 2014-06-18
wp8 support webp

dpull:
wp8 support webp


minggo:
Where is libwebp?

dpull:
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/11

minggo:
We will not add new feature on v2.
Thanks.

PR #7127, opened 2014-06-19
Update CCEditBoxImplWp8.cpp

wujianv5:


extensions/GUI/CCEditBox/CCEditBoxImplWp8.cpp
```diff
@@ -191,7 +191,7 @@ void CCEditBoxImplWp8::setText( const char* pText )
 
 			if (EditBox::InputFlag::PASSWORD == m_eEditBoxInputFlag)
 			{
-				long length = cc_utf8_strlen(m_strText.c_str(), -1);
+				long length = StringUtils::getCharacterCountInUTF8String(m_strText);
```
zilongshanren, line 194: @wuasiam 
Here is the `cc_utf8_strlen` implementation:

```
long cc_utf8_strlen (const char * p, int max)
{
    CC_UNUSED_PARAM(max);
    if (p == nullptr)
        return -1;
    return StringUtils::getCharacterCountInUTF8String(p);
}
```

And the CCEditBoxImp on other platforms also apply the function`cc_utf8_strlen`, so to keep consistency, we don't need to change it now.
Thanks.

zilongshanren:
@minggo we could close this PR now.

PR #7137, opened 2014-06-19
add chipmunk sensor

zawasp:
Add support for Chipmunk sensors. From Chipmunk manual:

"A boolean value if this shape is a sensor or not. Sensors only call collision callbacks, and never generate real collisions"

Sensors are integrated in cocos2d-iphone v3.

### DIFF
--- a/cocos/physics/CCPhysicsBody.cpp
+++ b/cocos/physics/CCPhysicsBody.cpp
@@ -68,6 +68,7 @@ PhysicsBody::PhysicsBody()
 , _isDamping(false)
 , _linearDamping(0.0f)
 , _angularDamping(0.0f)
+, _sensor(false)
 , _tag(0)
 , _categoryBitmask(UINT_MAX)
 , _collisionBitmask(0)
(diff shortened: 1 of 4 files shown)


zawasp:
retest this please.

it seems the ios_mac build failed for some reason.

boyu0:
Hmm, We already have a way to implement this( see setCollisionBitmask, setContactTestBitmask and setCategoryBitmask ), I don't think provide two ways to implement one function is a good idea...

zawasp:
Ok, so to set the equivalent of a sensor body in chipmunk, I need to set:
 _collisionBitmask to 0x00000000

and to get notifications of collisions to set the appropriate _contactTestBitmask + _categoryBitmask?

UPDATE: no, it doesn't work. If I set _collisionBitmask to 0 I don't get collision notifications at all, even though _contactTestBitmask and _categoryBitmask are set accordingly.

So the sensor type in chipmunk acts differently:
- the objects are not affected by collision (they don't react to it), BUT you get notifications when they touch.

This feature is VERY useful when you want to use the physics engine to get collision notifications, but at the same time to let cocos2d control the movement of objects.

boyu0:
you can set _collisionBitmask to 0x00000000, set _contactTestBitmask to 0xFFFFFFFF, set _categoryBitmask to 0xFFFFFFFF( default value ), I think this can work, you can test it.

zawasp:
Ok, I tested that (works with other bitmasks as well), only onContactBegin is called, which should be OK. Well it's up to you - if you want to extend cocos2d with additional chipmunk functionality.

Thanks for your support.

minggo:
I agree with @boyu0.
It is better to have one way to do a thing.

@zawasp 
Is there any problem of current implementation.

PR #7161, opened 2014-06-20
Wp8 v3 fix

stammen:
This pull request fixes the following:
1. Adds missing files and tests that broke the WP8 build
2. Fixes the WP8 shader compiler tool in build/winrt/Cocos2dShaderCompiler
3. Updates the wp8 precompiled shaders file
4. Adds instructions on how to use the shader compiler tool.


minggo:
Thanks.

PR #7165, opened 2014-06-22
- Due to the fact that map::insert doesn't allow to insert an element by...

timur-losev:
Due to the fact that map::insert doesn't allow to insert an element by existing key, I suggest to use operator []. It allows to change the cached value, very helpful on Android platform: when context is lost, the application will reset - rebuild - recache all it's shaders.


minggo:
Yep, you are right.

> very helpful on Android platform: when context is lost, the application will reset - rebuild - recache all it's shaders

Did you mean you invoke `addGLProgram` again when context lost?

timur-losev:
> Did you mean you invoke addGLProgram again when context lost?

Yes

PR #7171, opened 2014-06-23
issue #4693: deprecated some functions

minggo:


minggo:
Deprecated functions
- addChild(Node *child, int localZOrder, int tag)
- setTag
- getChildByTag
- removeChildByTag

PR #7179, opened 2014-06-23
-Cocos2d-x used a deprecated method to get the content size of a CCLabel...

adamfischer:
Cocos2d-x used a deprecated method to get the content size of a CCLabelTTF on iOS 7. 

It did not work correctly on iOS7,  If a CCLabelTTF was created with 0 height, it resized itself to contain its contents, but the calculated size was smaller than the real size of what the label required. This caused letters like 'J' to be cropped.

 The problem was the deprecated sizeWithFont method in CCImage calculateStringSize. This worked correctly until iOS6, but gives bad results on iOS7.

In this fix on iOS6+ boundingRectWithSize is used instead of sizeWithFont in CCImage.mm. CCLabelTTF height is now calculated correctly.

See: http://stackoverflow.com/questions/22537998/cclabelttf-contentsize-different-in-ios-6-and-ios-7-cocos2d-x/24364262#24364262

### DIFF
--- a/cocos2dx/platform/ios/CCImage.mm
+++ b/cocos2dx/platform/ios/CCImage.mm
@@ -174,14 +174,22 @@ static CGSize _calculateStringSize(NSString *str, id font, CGSize *constrainSize
     textRect.height = constrainSize->height > 0 ? constrainSize->height
                                               : 0x7fffffff;
     
-    
+
     for (NSString *s in listItems)
     {
-        CGSize tmp = [s sizeWithFont:font constrainedToSize:textRect];
+        CGSize tmp;
+        
+        // Method only exists on iOS6+.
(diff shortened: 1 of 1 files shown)


minggo:
So it also has problem in v3.
The implementation of v3 in CCDevice.mm.

Could you please send it to v3 too? 
Thanks.

adamfischer:
Okay, I created this pull request: 
https://github.com/cocos2d/cocos2d-x/pull/7206

minggo:
Thanks.

minggo:
@Dhilan007
Could you please test it and apply the patch to v3 too if it resolved the issue.
Thanks.

WenhaiLin:
@minggo I will test it next weak.

minggo:
Thanks for the PR. V2 is in maintain mode, only merge very critical bugs, so i will close this PR.

PR #7188, opened 2014-06-24
Refactor #5580: Add new file&directory apis into FileUtils and test cases

pandamicro:


cocos/platform/CCFileUtils.h
```diff
@@ -284,6 +284,30 @@ class CC_DLL FileUtils
...
+     *  Write a ValueMap to a plist file.
+     *  @note This method is used internally.
+     */
+    virtual bool writeToFile(ValueMap& dict, const std::string& fullPath);
+    
+    /**
+     *  Converts the contents of a file to a ValueVector.
+     *  @note This method is used internally.
+     */
+    virtual ValueVector getValueVectorFromFile(const std::string& filename);
```
minggo, line 308: Please input internally functions into `private` or `protected`.

minggo:
I think new functions should not only support absolute path. It should support the path relative to the root of an application.

pandamicro:
@minggo I think isDirectoryExist should support relative path and this have been done. But not other manipulation functions, because user need to know exactly which directory / file to remove / add, a relative path didn't identify exactly one directory or file.

cocos/platform/CCFileUtils.cpp
```diff
@@ -916,6 +936,255 @@ bool FileUtils::isAbsolutePath(const std::string& path) const
...
+    }
+    return true;
+#else
+    if ((GetFileAttributesA(path.c_str())) == INVALID_FILE_ATTRIBUTES)
+    {
+		subpath = "";
+		for(int i = 0 ; i < dirs.size() ; ++i)
+		{
+			subpath += dirs[i];
+			BOOL ret = CreateDirectoryA(subpath.c_str(), NULL);
```
minggo, line 1052: Some issue:
- BOOL -> bool
- NULL -> nullptr
- why not implement this function based on `FileUtils::createDirectory`?

minggo, line 1052: Another one, please replace tab with 4 spaces.

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
```diff
@@ -307,6 +309,168 @@ std::string TestIsFileExist::subtitle() const
...
+    std::string msg;
+    
+    FILE *out = fopen(filepath.c_str(), "w");
+    fputs(content.c_str(), out);
+    fclose(out);
+    
+    // Check whether file can be created
+    if (sharedFileUtils->isFileExist(filepath))
+    {
+        pTTF = Label::createWithSystemFont("Test file '__test.test' created", "", 20);
```
minggo: pTTF -> TTF.
Don't use Hungary style.

pandamicro:
@minggo Thanks for the suggestions, it's done now, please refer to the last commit

minggo:
Thanks.
Could we merge `createDirectory` and `createDirectories`?
It is not convenient to has two functions.

pandamicro:
@minggo I agree with you, and it has been done in the new commit

cocos/platform/CCFileUtils.h
```diff
@@ -384,6 +439,9 @@ class CC_DLL FileUtils
     virtual std::string getFullPathForDirectoryAndFilename(const std::string& directory, const std::string& filename);
     
     
+    virtual std::string searchFullPathForFilename(const std::string& filename) const;
```
minggo, line 442: What's this function mean?

cocos/platform/CCFileUtils.h
```diff
@@ -362,6 +412,11 @@ class CC_DLL FileUtils
     virtual bool isFileExistInternal(const std::string& filename) const = 0;
     
     /**
+     *  Checks whether file exists without considering search paths and resolution orders.
+     */
+    virtual bool isDirectoryExistInternal(const std::string& dirPath) const;
```
minggo, line 417: I don't think we need this function.

cocos/platform/CCFileUtils.h
pandamicro, line 442: To detect the full path of a given filename regarding all the search path and resolution

cocos/platform/CCFileUtils.h
pandamicro, line 417: This is for keeping the `isDirectoryExist` function clean

PR #7192, opened 2014-06-24
bugfix: condition variable sleep on unrelated mutex

kezhuw:
In HttpClient::networkThread 's main loop:

```
        // step 1: send http request if the requestQueue isn't empty
        request = nullptr;

        s_requestQueueMutex.lock();

        //Get request task from queue

        if (!s_requestQueue->empty())
        {
            request = s_requestQueue->at(0);
            s_requestQueue->erase(0);
        }

        s_requestQueueMutex.unlock();

        if (nullptr == request)
        {
            // Wait for http request tasks from main thread
            std::unique_lock<std::mutex> lk(s_SleepMutex); 
            s_SleepCondition.wait(lk);
            continue;
        }
```

After `s_requestQueueMutex.unlock();` and before `s_SleepCondition.wait(lk);`, there are chances that requests send through `HttpClient::send`. Then network thread will block on `s_SleepCondition` while there are requests in `s_requestQueue`.

This pull request just fix it.

There are other race conditions in this module for now, eg:
-   Between call sequences  `HttpClient::destroyInstance()` and `HttpClient::getInstance()->send()`, there are chances `s_requestQueue` is not `nullptr` in `HttpClient::lazyInitThreadSemphore`. So that new network thread is not started, while old network thread is about to terminated.
-   After `scheduler->performFunctionInCocosThread(CC_CALLBACK_0(HttpClient::dispatchResponseCallbacks, this));`, there is no guarantee that `this` is valid in `HttpClient::dispatchResponseCallbacks`.

The easiest way to suppress those is deleting public interface `HttpClient::destroyInstance`.

Also `s_cookieFilename` may not be thread-safe, it is depend on how `HttpClient::enableCookies` supposed to be called.

### DIFF
--- a/cocos/network/HttpClient.cpp
+++ b/cocos/network/HttpClient.cpp
@@ -47,15 +47,12 @@ namespace network {
 static std::mutex       s_requestQueueMutex;
 static std::mutex       s_responseQueueMutex;
 
-static std::mutex       s_SleepMutex;
-static std::condition_variable      s_SleepCondition;
+static std::condition_variable_any s_SleepCondition;
 
 
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
 typedef int int32_t;
 #endif
-
-static bool s_need_quit = false;
 
 static Vector<HttpRequest*>*  s_requestQueue = nullptr;
 static Vector<HttpResponse*>* s_responseQueue = nullptr;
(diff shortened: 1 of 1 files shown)


cmzx3444:
why not merge this pr?

minggo:
I will review and test it ASAP.

cocos/network/HttpClient.cpp
```diff
@@ -101,44 +98,31 @@ static int processPutTask(HttpRequest *request, write_callback callback, void *s
...
-        
-        s_requestQueueMutex.lock();
-        
-        //Get request task from queue
-        
-        if (!s_requestQueue->empty())
         {
+            std::lock_guard<std::mutex> lock(s_requestQueueMutex);
+            while (s_requestQueue->empty()) {
+                s_SleepCondition.wait(s_requestQueueMutex);
```
WenhaiLin, line 117: After  s_SleepCondition.wait(s_requestQueueMutex), there can't chances that requests send through HttpClient::send. Because the s_requestQueueMutex still lock when s_requestQueue is empty.

minggo, line 117: @kezhuw 
How do you think of @Dhilan007 's comment?

kezhuw, line 117: http://en.cppreference.com/w/cpp/thread/condition_variable_any/wait

template< class Lock >
void wait( Lock& lock );

Atomically releases lock, blocks the current executing thread, and adds it to the list of threads waiting on *this. The thread will be unblocked when notify_all() or notify_one() is executed. It may also be unblocked spuriously. When unblocked, regardless of the reason, lock is reacquired and wait exits.

minggo, line 117: Yep, it seems good.
Will you please fix the conflict?
I will merge it immediately.
Thanks.

kezhuw:
I will do it after work.
It is ok to rebase to branch/v3 ?

minggo:
It is ok.
Thanks.

minggo:
[ci rebuild]

minggo:
[ci rebuild]

PR #7202, opened 2014-06-25
[AUTO]: updating luabinding automatically

CocosRobot:


litefeel:
only space in line end?

minggo:
Yep, it is strange.
We will check it.

PR #7206, opened 2014-06-25
-String size should be calculated correctly on iOS7+ now.

adamfischer:
It is the same issue with the same solution, but for cocos2d-x 3.

https://github.com/cocos2d/cocos2d-x/pull/7179

### DIFF
--- a/cocos/platform/ios/CCDevice.mm
+++ b/cocos/platform/ios/CCDevice.mm
@@ -215,14 +215,34 @@ static inline void lazyCheckIOS7()
 
 static CGSize _calculateStringSize(NSString *str, id font, CGSize *constrainSize)
 {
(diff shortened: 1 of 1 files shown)


cocos/platform/ios/CCDevice.mm
```diff
@@ -215,14 +215,34 @@ static inline void lazyCheckIOS7()
...
+            tmp = [s sizeWithFont:font constrainedToSize:textRect];
+        }
+        
+        if (tmp.width > dim.width)
+        {
+            dim.width = tmp.width;
+        }
+        
+        dim.height += tmp.height;
+    }
```
minggo, line 244: `NSString sizeWithFont: constrainedToSize` can work for multiline strings. So we don't have to break it into lines, then calculate width and height.

PR #7214, opened 2014-06-25
add operator == != for Value

litefeel:


minggo:
Yep. Float comparison is a little different.
So i think we should treat it specially.

litefeel:
`win32_win7 completed with result FAILURE`?
why ?

minggo:
It is not your fault.
The machine has problem in connecting to github.

PR #7228, opened 2014-06-27
deprecated CCControlExtensions and add UIScale9Sprite

zilongshanren:


newnon:
i thing it is not so good idea i use control extensions more than other

minggo:
@newnon 
UI system has needed widgets. You can use them.
And it is confused to have more UI system. So i think we should merge it into UI system.

newnon:
I use my own fork of cocos with support of SpriteBuilderX(it is my special fork of sprite builder with full cocos2d-x support) https://github.com/newnon/SpriteBuilderX
after this system will be table i will make pull request with it. (Somewhere SpriteBuilder much much better then cocos studio for example it support physics)

newnon:
If you want to speak about it contact me in skype newnon
or you can take SpriteBuilderX here http://newnon.github.io/SpriteBuilderX/

zilongshanren:
@newnon 
I think you can still use ControlExtension, but  other developers will get confused if there exists more than one GUI system.
BTW, thanks for your mention of SpriteBuilderX, It will try it later.

zilongshanren:
[ci rebuild]

Mazyod:
@newnon I took a look at your work, freakin awesome, great job!! .. and open source FTW!!

PR #7230, opened 2014-06-27
CMake improvement:allow linux users to use OpenAL instead of FMOD

ghost:
I modified CMakeLists.txt so that Linux users can use OpenAL as backend of SimpleAudioEngine.
This feature is disabled by default.(see my CMakeLists.txt)

### DIFF
--- a/cocos/CMakeLists.txt
+++ b/cocos/CMakeLists.txt
@@ -25,6 +25,9 @@
 
 # this CMakeLists is to generate libcocos2d.a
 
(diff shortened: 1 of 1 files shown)


minggo:
Will you fix the conflict?
Now there are only one c++ lib in `cocos/CMakefileLists.txt`.

ghost:
updated for new CMakeList.txt

minggo:
Cmake changed lot in latest codes. Please send a new PR if needed.
Thanks.

PR #7237, opened 2014-06-29
Cmake update and remove old scripts from build folder

martell:
CMake: be more robust and also tell us what we are building for

@walzer This should be a good start to making a first installable package for cocos2d-x
We have /lib/libcocos2d.a
Now we just need includes in /include/


minggo:
Thanks.

But to be honest i am not interested of running shell on windows. It requires developers install mingw on windows. We do a lot work to remove the dependence of cygwin/mingw to create/build/run a new project on windows. We don't want to make it as simple as possible for developers.

martell:
Hi @minggo,
I think msys2 needs some explanation because everyone confuses it with cygwin.
MSYS2 is not cygwin it's a a git bash with a package manager.
Could you take some time to read this please :)

Here is how it works:

For example "qt-creator"
You can either download it from the website or 
Just run "pacman -S mingw-w64-x86_64-qt-creator"
This installs a prebuilt qt-creator.exe into /mingw64/bin

After install you can either type ./qt-creator which will then run qtcreator or you just navigate to the exe in windows explorer and run it just like you would normally.

It is essentially identical to the installer on the qt website just that the package manager handles the install. In-fact our installs run faster usually and are tested better.

Here is the current guide for msys2 and cocos2d-x.
http://discuss.cocos2d-x.org/t/setup-fastest-setup-for-cocos2d-x-on-windows-now-with-64bit-builds/15177?u=martell

Hopefully with the help of @walzer I should be able to make cocos2d-x into a package soon which would make it even simpler.
This means I will be able to remove build scripts like "build-mingw32-gcc-make.sh" I have in the cocos2d-x build folder from the cocos2d-x project

For a game developer
This means that a user will be able to do just type the command
"pacman -S mingw-w64-x86_64-cocos2d-x" that will install 

/mingw64/bin/cocos.py
/mingw64/lib/libcocos2d.a
/mingw64/include/cocos2d.h
etc

Then the user will be able to do use cocos.py
cocos new MyGame -p com.your_company.mygame -l cpp -d NEW_PROJECTS_DIR
from anywhere to generate a project.

This is the best setup for everyone as
They wont have to compile cocos2d-x because it is provided already.
They just worry about making their game code this way.

martell:
@minggo I managed to build a package for cocos2d-x
https://github.com/Alexpux/MINGW-packages/commit/04b8a5cce63064f57759ee76c2ea3d55b5fc8662
I will now remove all the mingw and clang shells from the build folder :)

Congrats cocos2d-x has now made it into a package manager.
I will use that PKGBUILD to get someone to put cocos2d-x into archlinux also :)

minggo:
Thanks.
It sounds good.
Could you please let me know when it is ready and write a usage document?
I will try it and give some feed back.

martell:
Hi @minggo 

As you asked I have made a usage doc.
Have a look at the new guide here.
http://discuss.cocos2d-x.org/t/setup-fastest-setup-for-cocos2d-x-on-windows-with-qt-creator-debugging/15177?u=martell

I have tested it myself and it works quite well. 
The pacman package won't go live on to the pacman server until tomorrow so that's when you will be able to follow the tutorial.
If you want to test it right away I can send you a copy of the package.
You would just run pacman -U on the tar i send you and it will work too.

I will mirror this across platforms soon so that linux and mac users can join the fun :)

This PR should be good to merge as long as it doesn't break linux.
I think Travis CI agrees :)

minggo:
Thanks.

I will merge it after v3.2 is released. We are do testing these days.

martell:
Thats sounds fine. :)
I can the patch apply it manually on the msys2 pkgbuild until then.

martell:
Hi @minggo 
I think this is good to go.
Could do with someone to test however
Cheers
Martell

minggo:
@martell 
Sorry about this PR. I forgot to merge it.
We are make modules for lua these days, and i think we should do it for cpp. As you know, now cocos2d module includes ui, cocostudio and so on. They should be  separate modules.

According the modification, now there is a conflict in this PR. Will you fix it?
I am learning more about CMake these days, and i think we can continue improving it.

martell:
Hi @minggo,
No Problem :)
I have updated the PR to take the new changes on lua into account.

Cmake should be a lot more robust now and tell us what we are building for :)

cocos/audio/CMakeLists.txt
```diff
@@ -1,29 +1,26 @@
...
-if ( CMAKE_SIZEOF_VOID_P EQUAL 8 )
-set(ARCH_DIR "64-bit")
-else()
-set(ARCH_DIR "32-bit")
-endif()
+#if ( CMAKE_SIZEOF_VOID_P EQUAL 8 )
+#set(ARCH_DIR "64-bit")
+#else()
+#set(ARCH_DIR "32-bit")
+#endif()
```
minggo, line 6: Why comment these codes?
`ARCH_DIR` is used later.

martell, line 6: It is not needed for audio/CMakeLists.txt
I have also set it up in a manor to inherit so that would be declaring it twice

minggo, line 6: Got it.
Thanks.

PR #7247, opened 2014-06-30
fix focus functions

zilongshanren:


minggo:
Please deprecated old one.

zilongshanren:
@minggo 
I think it's no need to deprecated the old one. Because they are all private helper methods.

PR #7252, opened 2014-06-30
Fixing compile errors on WP8

visiblelight:
add missing files for the projects


minggo:
@ricardoquesada 
It is not blocked.

PR #7260, opened 2014-06-30
fix that FastTileMapTest crashes on WP8

visiblelight:


cocos/2d/CCFastTMXLayer.cpp
```diff
@@ -180,7 +180,6 @@ void FastTMXLayer::onDraw(int offset, int count)
     GL::bindTexture2D(_texture->getName());
     getGLProgramState()->apply(_modelViewTransform);
     
-    glBindVertexArray(0);
```
dabingnn, line 183: it is safer to bind VAO to zero if VAO is supported.
Please call `GL::bindVAO(0)` instead of this.

visiblelight, line 183: done.

PR #7270, opened 2014-07-01
Fix a potential crash in Scheduler

Pisces000221:
Sometimes the program crashes if we use `ccSchedulerFunc` together with `SEL_SCHEDULE` in one scheduler.
The crash is caused by invalid `static_cast`, and now it's fixed by using `dynamic_cast`.

I'm not sure whether to add a test case for this. Please tell me. Thanks!

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -311,9 +311,9 @@ void Scheduler::schedule(const ccSchedulerFunc& callback, void *target, float in
     {
         for (int i = 0; i < element->timers->num; ++i)
         {
-            TimerTargetCallback *timer = static_cast<TimerTargetCallback*>(element->timers->arr[i]);
+            TimerTargetCallback *timer = dynamic_cast<TimerTargetCallback*>(element->timers->arr[i]);
 
-            if (key == timer->getKey())
+            if (timer != nullptr && key == timer->getKey())
             {
                 CCLOG("CCScheduler#scheduleSelector. Selector already scheduled. Updating interval from: %.4f to %.4f", timer->getInterval(), interval);
                 timer->setInterval(interval);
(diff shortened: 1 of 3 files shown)


minggo:
Yep, it is better to add a test case in tests/cpp-tests.

Pisces000221:
Okay, I'll add one soon.
What should I name it? Seems like this doesn't have an issue number and I can't think of a short description.

minggo:
There is a `SchedulerTest.cpp`, i think you can add it there. The subtile can be `should not crash` or something else.

Pisces000221:
Thanks, but I can't think of a short title like `SchedulerCallbackTest`. Will SchedulerCallbackWithTwoTypes work?

Pisces000221:
`SchedulerCallbackTest` is a test case name in SchedulerTest.cpp.

Pisces000221:
Sorry for disturbing, I got a shorter name. I'll create the test case soon. Thank you anyway :)

Pisces000221:
The test case is added. Please let me know if there are any mistakes.

(PS. Don't be confused by the long time (2 hrs) between the comment and the commit. I left for a while after the comment and this only took me half an hour, not 2 hours :-D)

minggo:
Thanks.

minggo:
I run the test case and revert the modification of `CCScheduler.cpp`, no crash happened.

minggo:
@Pisces000221 
Any thought?

Pisces000221:
Are you using a Mac? I'm developing under Linux (Ubuntu 13.10 32-bit) and I think maybe it only crashes on some platforms. I'll revert the modification and re-run the test on my machine now.

Pisces000221:
It doesn't crash after reverting... But it _did_ crash in one of my projects before making this change. I think it's a mandelbug. I don't know what made it crash but after I changed this file in the project, it worked very well instead of halting immediately. That's really strange.

Pisces000221:
May I remove the test case and only make a change in Scheduler? I'm pretty sure it may cause problems in some cases.

minggo:
@Pisces000221 
Because `dynamic_cast` is slower than `static_cast`, so i think it is better to reproduce it before merging this PR.

PR #7277, opened 2014-07-01
Fix setRotation3D

darkdukey:
The `Node::setRotation3D` was not based on AnchorPoint.
Since `Node::setRotation` was based on anchor point, I think `Node::setRotation3D` should behave the same

`Sprite3DRotationTest` was created under `SpriteTest.cpp`


minggo:
We have finished testing, and i will set tag for beta0.
So i will merge it after releasing beta0.

PR #7299, opened 2014-07-03
Fast tmx

dabingnn:


dabingnn:
This pr includes
1. new implementation for hex tile map
2. improvement for tile map rendering (30-40% up when the tile map on the screen is moving.)

PR #7309, opened 2014-07-03
changes needed for swift bindings to work

mannewalis:
Pretty minor changes
The callback changes in Node are a pattern I think we should adopt everywhere where you would normally call a class method. This is script friendly and even c++ friendly, and will allow you to handle events without having to subclass the class.
The Acceleration class  in ccTypes has to derive from Ref, so perhaps it should be moved to somewhere else.


minggo:
Is it a better way not to add new functions in an existing object to enable call back?
For example, if i want a call back in `Sprite`, then i have to add call back in it. It is not convenient.

mannewalis:
I'm sorry I don't quite understand, can you explain what you mean in more
detail?

By the way, this doesn't change the default behavior, just allows you to
specify a callback outside of the class.

Thanks
Justin

minggo:
Sorry for confused.
What i mean is that, if i want to send a call back event to script, i have to add functions for existing classes. It is not convenient.

Is it possible to do it like this

``` c++
void Node::onEnter()
{
    ...

    // add this line to send a call back event to script
    sendCallbackToScript();
}
```

Then i don't have to add functions in existing classes.

mannewalis:
Do you mean the setters and getters for _onEnterCallback / _onExitCallback
etc?

Thanks
Justin

minggo:
Yep.

mannewalis:
So if we just had a callScriptMethod inside the onEnter/onExit etc. How
would I tell it to call my script without having subclass the Sprite class?

The reason for the setter/getters are so that I can make the Node call any
method or lambda method that is not inside Node or Sprite, and so I don't
have to subclass in order to get an onEnter / onExit callback.

minggo:
> So if we just had a callScriptMethod inside the onEnter/onExit etc. How
> would I tell it to call my script without having subclass the Sprite class?

As you can see, we don't add functions to send event for lua-binding and js-binding. We record a map, the map looks like
[ c++ object : (scriptFunction1, scriptFunction2, ...) ]

> The reason for the setter/getters are so that I can make the Node call any
> method or lambda method that is not inside Node or Sprite, and so I don't
> have to subclass in order to get an onEnter / onExit callback.

Yep. But how to handle classes that don't inherit from Node?

mannewalis:
I think we are talking about slightly different use cases. These setters
and getters can be used by C++ code as well, a class could create a scene
and hook up the callbacks to itself and know when the scene is entered or
its transition is done. I think this pattern where you use std::function to
callback instead of using virtual methods that you need to override by
deriving a new class, is a good one. It just also happens to be useful for
scripting.

On Jul 4, 2014, at 8:32, minggo notifications@github.com wrote:

So if we just had a callScriptMethod inside the onEnter/onExit etc. How
would I tell it to call my script without having subclass the Sprite class?

As you can see, we don't add functions to send event for lua-binding and
js-binding. We record a map, the map looks like
[ c++ object : (scriptFunction1, scriptFunction2, ...) ]

The reason for the setter/getters are so that I can make the Node call any
method or lambda method that is not inside Node or Sprite, and so I don't
have to subclass in order to get an onEnter / onExit callback.

Yep. But how to handle classes that don't inherit from Node?

minggo:
Ok.
Got it.
I just worry about that it is used for script only.

minggo:
[ci rebuild]

mannewalis:
Cool, and yeah swift is not really a scripting language, so you derive from
obj-c classes directly, so being to set delegates for callbacks from C++ to
obj-c via lambda wrapped blocks is incredibly useful.

On Jul 4, 2014, at 10:05, minggo notifications@github.com wrote:

Ok.
Got it.
I just worry about that it is used for script only.

minggo:
This PR blocked lua-binding. I will ask someone to check it.
I will merge it when it is fixed.

samuele3hu:
Can you update the newest node of cocos2d-x?Thanks.

mannewalis:
Hi

Pull request has been submitted, awaiting approval and release.

Thanks
Justin

On Jul 5, 2014, at 8:48, samuele3hu notifications@github.com wrote:

Can you update the newest node of cocos2d-x?Thanks.

minggo:
[ci rebuild]

samuele3hu:
[ci rebuild]

minggo:
[ci rebuild]

PR #7313, opened 2014-07-04
issue #4689: Refactoring game controller API.

WenhaiLin:


cocos/base/CCController.h
```diff
@@ -27,50 +27,112 @@
...
+    static Controller* getControllerByTag(int tag);
 
     static void startDiscoveryController();
     static void stopDiscoveryController();
 
-    const std::string& getVendorName();
-    bool isConnected() const;
-
-    static const int PLAYER_INDEX_UNSET = -1;
+    const std::string& getDeviceName();
```
minggo: const std::string& getDeviceName() -> const std::string& getDeviceName() const

cocos/base/CCController.h
```diff
@@ -27,50 +27,112 @@
...
+    int getDeviceId() const { return _deviceId;}
 
-    int getPlayerIndex() const;
-    void setPlayerIndex(int playerIndex);
+    bool isConnected() const;
 
-    Gamepad* getGamepad() const;
+    const KeyStatus& getKeyStatus(int keyCode);
```
minggo, line 100: const KeyStatus& getKeyStatus(int keyCode) -> const KeyStatus& getKeyStatus(int keyCode) const

PR #7324, opened 2014-07-06
drawCubicBezier no longer fills in bezier curves.

tobyflint:
Test case: "Node: Draw > Test DrawNode"

I made this change so I could draw bezier curves in a Cocos2d-js project. It's not a perfect change, but I thought I'd submit a pull request in case you guys find it useful.

Known issues:
As you can see, I'm calling drawSegment() with a constant 0.5f line width. This line width should probably be passed in as an argument to drawCubicBezier, but since 1) that would change the interface of the function and 2) I wasn't for sure how to update the js-bindings, I was thinking it would be best if someone more knowledgable makes that change.

drawQuadBezier could probably benefit from a similar fix.

Unfortunately I don't quite have the time to perfect this change, but hopefully it's a useful start :D

### DIFF
--- a/cocos/2d/CCDrawNode.cpp
+++ b/cocos/2d/CCDrawNode.cpp
@@ -480,25 +480,19 @@ void DrawNode::drawCubicBezier(const Vec2& from, const Vec2& control1, const Vec
     Tex2F texCoord = Tex2F(0.0, 0.0);
     Color4B col = Color4B(color);
     Vec2 vertex;
-    Vec2 firstVertex = Vec2(from.x, from.y);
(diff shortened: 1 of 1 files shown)


minggo:
@TobyFlint 
What's the issue?

tobyflint:
Ah, here's the forum thread with the original problem: http://discuss.cocos2d-x.org/t/drawcubicbezier-in-js-tests-project-not-rendering-correctly/15225

minggo:
@TobyFlint 

ghost:
@minggo This problem had already fixed, the pr can be closed.

PR #7334, opened 2014-07-08
change the logic of texture loading

dabingnn:


dabingnn:
This pr changed the way of texture creating
1. using defaultPixelFormat if specified one
2. using _renderFormat of image if defaultPixelFormat is None or auto

PR #7337, opened 2014-07-08
Make Setup.py Support for zsh.

chareice:
When i use zsh for my default shell, `python setup.py` don't take the correct information for me.This change make it works.


minggo:
Thanks.

PR #7339, opened 2014-07-08
fix android, add 3d files to cocos2d.h and remove unused resources

super626:
fix android may crash when change scene after switch background and foreground.

add 3d files to cocos2d.h

resolve conflict cocos2d::Bone and cocostudio::Bone

some adjustment of Bundle3D

remove unused resource


minggo:
[ci rebuild]

PR #7343, opened 2014-07-08
Add getChildByName method for get a node that can be cast to Type T

zhouxiaoxiaoxujian:


minggo:
Why need this function?

zhouxiaoxiaoxujian:
Thank you for you reply.

Because , when use UIWidget Components, I need always cast to subType of UIWidget. like this:

```
::cocos2d::ui::Button btn = static_cast<::cocos2d::ui::Button*>(this->getChildByName("btn"));

::cocos2d::ui::Button btn = this->getChildByName<::cocos2d::ui::Button*>("btn");
```

I think below could be much more  readable  and usable than above.

darkdukey:
I like this, definitely made it easier

Mazyod:
Unity style, love it.

darkdukey:
Can we merge this, this is much easier to use.

PR #7363, opened 2014-07-09
Dispatch EVENT_RENDERER_RECREATED event when renderer was recreated on android/wp8.

WenhaiLin:
1.Dispatch EVENT_RENDERER_RECREATED event when renderer was recreated on android/wp8
2:Make clear EVENT_COME_TO_FOREGROUND event.
3.fixed label displaying incorrectly when ‘Don't keep activities’ is enabled.


minggo:
[ci rebuild]

PR #7386, opened 2014-07-10
1. fixes TextReader load default font error.

chengstory:


cocos/editor-support/cocostudio/WidgetReader/TextReader/TextReader.cpp
```diff
@@ -102,8 +102,23 @@ namespace cocostudio
         label->setFontSize(DICTOOL->getIntValue_json(options, P_FontSize,20));
        
         std::string fontName = DICTOOL->getStringValue_json(options, P_FontName, "微软雅黑");
-        std::string fontFilePath = jsonPath.append(fontName);
-        label->setFontName(fontFilePath);
+        std::string file_extension = “”;
```
minggo, line 105: "" is Chinese words.

PR #7407, opened 2014-07-11
Fix compile errors on WP8

visiblelight:


minggo:
Please fix conflict.

tests/cpp-tests/Classes/MenuTest/MenuTest.cpp
```diff
@@ -445,7 +445,7 @@ MenuLayer4::MenuLayer4()
                   item3, item4,
                   back, nullptr ); // 9 items.
     
-    menu->alignItemsInColumns(2, 2, 2, 2, 1, nullptr);
+    menu->alignItemsInColumns(2, 2, 2, 2, 1, NULL);
```
minggo, line 448: Please add comment here that we need NULL here.

PR #7426, opened 2014-07-13
CCEditBox SecureInput Mac Fixes

Mazyod:
[FIX]: When getting the string value, check which textfield is in use
[FIX]: when setting the string value, change both textfields.


minggo:
@Mazyod 
What does this pull request fix?

Mazyod:
@minggo Sorry I didn't make that clear. It fixes two things for `EditBox` on Mac:
1. When you do `editbox->getText()` and the edit box type is `InputFlag::PASSWORD`, you get an empty string.
2. When you do `editbox->setText("whatever")` and the edit box type is `InputFlag::PASSWORD`, the textfield doesn't update.

PR #7439, opened 2014-07-14
merge iOS arm64 libs to fat libs, change the project setting.

boyu0:
Fix compile error on  xcode 6 belta3 iphone5s.
The corresponding third party PR is: https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/12


minggo:
Yep. It can work for `cpp-tests`, but lua test cases can not work.

minggo:
[ci rebuild]

PR #7440, opened 2014-07-14
Add utils::atof(), fix bug: particle not shown on android.

boyu0:


litefeel:
why do not use `sprintf`?

minggo:
@lite3 
How to use `spintf` instead?
Thanks.

litefeel:
i am sorry. I was wrong.

minggo:
No problem.

felixjingga:
@minggo why don't use sscanf instead?

boyu0:
@felixjingga I tested it, it has the same problem as atof.

boyu0:
[ci rebuild]

litefeel:
strtod ?

minggo:
It is strange that, if we use atof in samples shift by NDK r9d, then no problem.

Environment:
- NDK r9d
- Clang + libc++

Code:

``` c++
float f = atof("90.099998474121094");
LOGI("Test %f",f);
```

Result:

Test 90.099998  -> in shift sample by NDK r9d
Test -inf -> used in cocos2d-x

minggo:
@lite3 
`strtod`, `std::atof`, `atof` all can not work.
And we found, the result is correct if running with `armeabi-v7s`, which will use fpu.

PR #7442, opened 2014-07-15
Fixed #5724: Fix armature bone's transform issue

pandamicro:


cocos/editor-support/cocostudio/CCSkin.cpp
```diff
@@ -160,6 +160,7 @@ void Skin::updateTransform()
         //
         // calculate the Quad based on the Affine Matrix
         //
+        Mat4 transform = getNodeToParentTransform();
```
pandamicro, line 163: We can't use _transform because potential changes (rotation, translation, etc) haven't been applied to the _transform, getNodeToParentTransform will update it.

PR #7462, opened 2014-07-16
1、 fixes lua TriggerTest bug.

chengstory:


samuele3hu:
All the binary data to json for lua is ok now,Thanks.

PR #7467, opened 2014-07-16
fix enabled to interval

ntotani:


samuele3hu:
This bug was caused by carelessness,Thanks.And the most functions of Layer  were deprecated,you should use NewEventDispatcher.

PR #7503, opened 2014-07-17
修正Linux下中文显示问题

CaiCQ:
修正Linux下中文显示问题


minggo:
@CaiCQ
How to reproduce the issue?
Because there is a test case named `Node:Label-New API`. And it can show Chinese characters correctly. 
It can't show some Chinese characters in `Node:Label-Old API`, but this PR doesn't fix it.

CaiCQ:
这是解决中文显示不全问题，中文变方块是因为系统没对应字体，不是该问题

minggo:
How to reproduce it?
Could you paste some demo codes?

minggo:
PS: Please use English in future.
Because other developers will join the discussion.

Thanks.

CaiCQ:
## example：

``` c++
    auto font = LabelTTF::create();
    font->setFontName("文泉驿微米黑");
    font->setFontSize(32);
    font->setString("中文显示不全问题例子");
    this->addChild(font);
```

minggo:
Thanks.

PR #7515, opened 2014-07-18
fixed warn and replace 'printf' with 'log';

WenhaiLin:


minggo:
Menu create may need NULL on WP8, because WP8 doesn't support nullptr in variable parameters.
Please check it.

WenhaiLin:
@minggo Menu class use variadicCreate with NULL as the last argument  for create menu on wp8

PR #7519, opened 2014-07-18
Fix load Widget anchor point error.

hanrea:


hanrea:
Please use this pull：https://github.com/cocos2d/cocos2d-x/pull/7546

PR #7522, opened 2014-07-18
Update download-deps.py for download fbx-conv

WenhaiLin:
1.Update download-deps.py for download fbx-conv
2.Update git-archive-all for pack fbx-conv into cocos2d-x engine archive


WenhaiLin:
[ci rebuild]

PR #7523, opened 2014-07-18
Update CCBundleReader.h

teopeurt:
Why is tell() declared as 'long int' but then subsequently defined as 'ssize_t'?

This breaks my build, hence this 'fix'


minggo:
It is fixed in latest codes. 
Thanks.

PR #7524, opened 2014-07-18
Wp8 v3 rendertexture fix

stammen:
I looked at the CCRenderTexture.cpp code and found this around line 249:

```
    if (Configuration::getInstance()->checkForGLExtension("GL_QCOM"))
```

What is the GL_QCOM extension? Since this is not supported on WP8, I changed the line to:
# if CC_TARGET_PLATFORM != CC_PLATFORM_WP8

```
    if (Configuration::getInstance()->checkForGLExtension("GL_QCOM"))
```
# endif

This will allow the texture to be copied.

I also made changes to Cocos2dRenderer.cpp at line 72 to send the EVENT_COME_TO_FOREGROUND event.

There are still problems with the RenderTextureTests. Several tests are not listening for the EVENT_RENDERER_RECREATED event and are not recreating the sprites. Also, some of the rendertextures are not getting the EVENT_COME_TO_BACKGROUND/FOREGROUND events which seems to be a bug/bugs in the tests.

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -99,8 +99,7 @@ void RenderTexture::listenToBackground(EventCustom *event)
     int count = 0;
     count++;
 
-#if 1
-#if CC_ENABLE_CACHE_TEXTURE_DATA
+#if CC_ENABLE_CACHE_TEXTURE_DATA && (CC_TARGET_PLATFORM == CC_PLATFORM_WP8)
     CC_SAFE_DELETE(_UITextureImage);
     
     // to get the rendered texture data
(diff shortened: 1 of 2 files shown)


visiblelight:
This PR doesn't fix the issue. It can be reproduced by RenderTextureSave in RenderTextrueTest.cpp
I think we still need to find a way to send the EVENT_COME_TO_BACKGROUND message prior to the textures are destroyed.

minggo:
@stammen 
What do you think of @visiblelight 's comment?

minggo:
Hi @stammen ,
Are you still working on v3 WP8?
What's the plan?

stammen:
We just verified that Android has the same issue. This seems to be a bug in the cocos2d-x rendertexture code.

PR #7525, opened 2014-07-19
2 bugs fix on v3

pShota:
-fix start app crash in android aff00ea
-fix keyboard "done" button bug in android

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
@@ -61,16 +61,21 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
 	
 	protected void onLoadNativeLibraries() {
 		try {
-			ApplicationInfo ai = getPackageManager().getApplicationInfo(getPackageName(), PackageManager.GET_META_DATA);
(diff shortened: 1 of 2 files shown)


minggo:
How to reproduce it.
Thanks.

pShota:
my device: samsung galaxy s2, os 4.1.2
The first one just run the app on device and it will crash on startup, Bundle is null in ApplicationInfo.metaData

the 2nd one is when using TextFieldTTF, in my s2 it has a "done" button with it, when you press done, it will close the keyboard directly so that IMEDelegate::canDetachWithIME is ignore

minggo:
Thanks.
- could you please fix compiling error?
- could you please fix indent?

minggo:
I have tested on Glaxy S2 4.1.2.
No crash happened.

> the 2nd one is when using TextFieldTTF, in my s2 it has a "done" button with it, when you press done, it will close the keyboard directly so that IMEDelegate::canDetachWithIME is ignore

Is it a problem to close keyboard when press done?

pShota:
-could you please fix compiling error?
-could you please fix indent?
ok~

-No crash happened.
thought so, but it keep crashing on my s2 until i add PackageManager.GET_ACTIVITIES flag, let me test more device

-Is it a problem to close keyboard when press done?
yes

PR #7528, opened 2014-07-20
handle controller events (Ouya, FireTV)

zawasp:
Added forwarding of all keypad events (to handle DPAD, A,B,X,Y etc. keys in cocos2d-x)
See here for details:
http://discuss.cocos2d-x.org/t/ouya-support/3544/27

Tested with FireTV. Solution is for Android only.

### DIFF
--- a/cocos2dx/keypad_dispatcher/CCKeypadDelegate.h
+++ b/cocos2dx/keypad_dispatcher/CCKeypadDelegate.h
@@ -45,6 +45,8 @@ public:
 
     // The menu key clicked. only available on wophone & android
     virtual void keyMenuClicked() {};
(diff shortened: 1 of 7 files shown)


minggo:
@zawasp 
Please don't add new features in v2.
Thanks.

You are appreciated if you can improve on v3.

zawasp:
@minggo We still have to maintain existing games for v2 - it's too much effort to port them to v3 right now.
v3 already has controller support with EventKeyboard.

minggo:
@zawasp 
As we have announced before, v2 is only fixed critical bugs.
We won't add new feature support in v2.

PR #7529, opened 2014-07-21
More details for conversion error msg

edwardzhou:
1. show class and function name when luaval_to_XXX failed.
   original error msg: 
   argument #2 is 'nil'; 'number' expected.

new error msg:
    cc.DelayTime:create argument #2 is 'nil'; 'number' expected.
1. sometimes, WebSocket.close will hang app due to join thread call. so I comment it out.
   disable sent & recv log for less annoying msg.


samuele3hu:
I will check out this branch,and update the manual bindings,thanks.

PR #7532, opened 2014-07-21
replace dynamic_cast to std::is_base_of

litefeel:


litefeel:
[rebuild]

litefeel:
[ci rebuild]

litefeel:
[ci rebuild]

samuele3hu:
I also want to use this method,it can avoid not polymorphic and it is compile time.

PR #7545, opened 2014-07-21
fix a bug where LabelTTF::setFontFillColor() won't work as expected

billtt:
In CCDevice.mm, Device::getTextureDataForText() function, info.strokeColorX and info.tintColorX are float type (0-1), which are assigned the value of GLubyte (0-255). The values need to be converted.
LabelTTF::setFontFillColor() will work properly once the fixes are made.


minggo:
It is fixed in latest codes: https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/ios/CCDevice.mm#L465.
It is strange that this PR still show the difference.

PR #7549, opened 2014-07-22
add editbox ended event

yongkangchen:


zilongshanren:
Thanks @yongkangchen 
  When would `controlTextDidEndEditing` method will be called? Could you please add some tests to stress this fix?

yongkangchen:
`controlTextDidEndEditing` method will be called when press return/enter

usage:

```
input:registerScriptEditBoxHandler(function(eventName, sender)
        if eventName == "ended" then
            send()
        end
end)
```

minggo:
@andyque 
How about this PR?

zilongshanren:
@yongkangchen 
Hi, I think we should fix like this : https://github.com/cocos2d/cocos2d-x/pull/7645

zilongshanren:
@minggo 
This PR could be closed now.

PR #7556, opened 2014-07-22
fixed FontAtlasCache::getFontAtlasCharMap may returns the same instance when using different arguments.

WenhaiLin:


cocos/2d/CCFontAtlasCache.cpp
```diff
@@ -178,7 +178,10 @@ FontAtlas * FontAtlasCache::getFontAtlasCharMap(Texture2D* texture, int itemWidt
 
 FontAtlas * FontAtlasCache::getFontAtlasCharMap(const std::string& charMapFile, int itemWidth, int itemHeight, int startCharMap)
 {
-    std::string atlasName = generateFontName(charMapFile, 0, GlyphCollection::CUSTOM,false);
+    char tmp[255];
+    sprintf(tmp,"name:%s_%d_%d_%d",charMapFile.c_str(),itemWidth,itemHeight,startCharMap);
```
minggo: What will happen if string length is greater than 255?

PR #7557, opened 2014-07-22
Mac os custom font loading fix

newnon:
Mac os user font from bundle loading fix
font load from path you don't even need to add it to plist file
second version of this patch https://github.com/cocos2d/cocos2d-x/pull/7578 - just truncate name internally as in iPhone to load font with path but you still need to add all fonts to .plist file
merge only one


Mazyod:
Nice one!

I developed a hybrid iOS/Mac library in the past, and I use this to load all the fonts dynamically, so I don't need to add them to the plist and all that setup crap. They can easily be loaded using `fontWithName:`. Just throwing it out there:

``` objc
- (void)_loadAllFontsDynamically
{
    for (NSString *fontType in @[@"ttf", @"otf"])
    {
        NSArray *fonts = [[NSBundle lpBundle] recursivePathsForResourcesOfType:fontType];
        for (NSString *fontPath in fonts)
        {
            NSData *inData = [NSData dataWithContentsOfFile:fontPath];

            CFErrorRef error;
            CGDataProviderRef provider = CGDataProviderCreateWithCFData((CFDataRef)inData);
            CGFontRef font = CGFontCreateWithDataProvider(provider);

            if (!CTFontManagerRegisterGraphicsFont(font, &error))
            {
                CFStringRef errorDescription = CFErrorCopyDescription(error);
                NSLog(@"Failed to load font: %@", errorDescription);
                CFRelease(errorDescription);
            }

            DLog(@"Loaded font: %@", font);

            CFRelease(font);
            CFRelease(provider);
        }
    }
}
```

newnon:
Nice, but cocos load font with full name on most platforms, thats why my method do exactly what we need.

Mazyod:
Ah, well the more platform agnostic the better :)

zilongshanren:
Thanks, @newnon  
Could you please add a test case in cpp-tests to stress this fix?

newnon:
ok i'll try

newnon:
i made second version of this patch https://github.com/cocos2d/cocos2d-x/pull/7578

minggo:
So i can close this one?

newnon:
you must decide which version of this path if more suitable for cocos. Both fix the same problem but with different solutions.

cocos/platform/ios/CCDevice.mm
```diff
@@ -260,6 +260,8 @@ static bool _initWithString(const char * text, cocos2d::Device::TextAlign align,
         // the '.ttf' extensions when referring to custom fonts.
         fntName = [[fntName lastPathComponent] stringByDeletingPathExtension];
         
+        CTFontManagerCreateFontDescriptorsFromURL
```
Mazyod: Isn't this causing a compilation error?

minggo:
@newnon 
If i add fonts into plist, and merged this PR.
Then what will happen?

newnon:
you already merged another one. after that you don't need any ifdef to load fonts just full path to it

PR #7560, opened 2014-07-22
FIX: make sure TableView parents are visible before handling touch

Mazyod:
Context:

After having issues with a `TableView` instance receiving touches even though the parent is not visible, I found some code in the `Menu` class that checks if parent is visible before handling touches, so I ported that code to the `TableView` class.


zilongshanren:
Thanks @Mazyod  We did the same decision when handling the ui widget's touch dispatching.

@minggo  I think we could merge this PR.

PR #7565, opened 2014-07-23
fixed name of getter/setter method

chenguangqi:


minggo:
Thanks.
Would you please deprecated old functions, not just delete them.

minggo:
It is fixed.

PR #7567, opened 2014-07-23
fix remove self in Widget touch callback issue

zilongshanren:


tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIButtonTest/UIButtonTest.cpp
```diff
@@ -344,3 +344,80 @@ void UIButtonTest_Title::touchEvent(Ref *pSender, Widget::TouchEventType type)
             break;
     }
 }
+
+
+// UIButtonTest_RemoveSelf
+UIButtonTest_RemoveSelf::UIButtonTest_RemoveSelf()
```
minggo: Don't use `_` in class name.

minggo:
[ci rebuild]

PR #7575, opened 2014-07-24
-fix keyboard "done" button bug in android

pShota:
this is the fix for
https://github.com/cocos2d/cocos2d-x/pull/7525
(sorry for creating new request, this is my first time to use pull request ^^")

-fix a bug when pressing "done" button of the keyboard in android, it closes the keyboard directly so that IMEDelegate have no chance to verify if the input text is correct

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxTextInputWraper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxTextInputWraper.java
@@ -26,6 +26,7 @@ package org.cocos2dx.lib;
 import android.content.Context;
 import android.text.Editable;
 import android.text.TextWatcher;
+import android.util.Log;
 import android.view.KeyEvent;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputMethodManager;
@@ -75,6 +76,7 @@ public class Cocos2dxTextInputWraper implements TextWatcher, OnEditorActionListe
 
 	@Override
 	public void afterTextChanged(final Editable s) {
+		
 		if (this.isFullScreenEdit()) {
 			return;
 		}
@@ -148,6 +150,7 @@ public class Cocos2dxTextInputWraper implements TextWatcher, OnEditorActionListe
 			}
 			
 			final String insertText = text;
+
 			this.mCocos2dxGLSurfaceView.insertText(insertText);
 			/*
 			if (BuildConfig.DEBUG) {
@@ -157,7 +160,11 @@ public class Cocos2dxTextInputWraper implements TextWatcher, OnEditorActionListe
 		}
 		
 		if (pActionID == EditorInfo.IME_ACTION_DONE) {
-			this.mCocos2dxGLSurfaceView.requestFocus();
+//			this.mCocos2dxGLSurfaceView.requestFocus();
+			// we need to consume the done action here, so that IMEDelegate have chance to 
+			// verify if the input text is correct. if it is correct, it will eventually
+			// call didDetachWithIME(); to close the keyboard
+			return true;
 		}
 		return false;
 	}


minggo:
No problem.
Thanks.

So you think click `done` to close keyboard is not expected behavior?
And how about the crash, did you have crash now on Galaxy S2?

pShota:
-- click done to close keyboard is not expected behavior?
it should not close immediately, it should allow IMEDelegate class to run
bool canDetachWithIME() and verify the input text

-- did you have crash now on Galaxy S2?
I figured out in the AndroidManifest.xml,
if the meta-data block is inside activity block, it will cause the bug,
it should be outside of the activity block , and inside application block.
the AndroidManifest.xml in latest project template already fixed this.

```
<activity android:name="org.cocos2dx.cpp.AppActivity"
                  android:label="@string/app_name"
                  android:screenOrientation="landscape"
                  android:theme="@android:style/Theme.NoTitleBar.Fullscreen"
                  android:configChanges="orientation">

            <!-- this will cause null pointer exception-->
            <meta-data android:name="android.app.lib_name" android:value="cocos2dcpp" />

            <intent-filter>
                <action android:name="android.intent.action.MAIN" />
                <category android:name="android.intent.category.LAUNCHER" />
            </intent-filter>
 </activity>
```

minggo:
If applied this PR, then click `done` can not close keyboard?

minggo:
And thanks for sharing about the crash reason.

pShota:
-- click done can not close keyboard?
no, the keyboard will still close after clicking "done" if 
IMEDelegate::canDetachWithIME() return true.

when "done" is pressed...
1. Cocos2dxTextInputWraper:: onEditorAction     // receive key event, call insertText
2. Cocos2dxGLSurfaceView.insertText(insertText);    // call nativeInsertText
3. Cocos2dxRenderer.nativeInsertText(pText);            // jni to c++
4. IMEDispatcher::dispatchInsertText(pszText, strlen(pszText));      // dispatch to textFieldTTF
5. TextFieldTTF::insertText  
6. TextFieldTTF::detachWithIME()                                  // attempt to close keyboard at the end
7. IMEDispatcher::detachDelegateWithIME                   // ask delegate to verify input
8. TextFieldTTF::canDetachWithIME()                           // return true
9. GLView::setIMEKeyboardState                                  // finally call jni again to java
10. setKeyboardStateJNI ,closeKeyboardJNI  
11. Cocos2dxGLSurfaceView::closeIMEKeyboard        // close keyboard

so the bug is the keyboard is already closed in step 1 without going through the rest

darkdukey:
[ci rebuild]

zilongshanren:
@pShota 
Hi pShota, thanks.
I have made a simple test. No matter I return true or false from `canDetachWithIME`, the IME dialog will always close.  

Another thing, if I return false from `canDetachWithIME` and use your patch. When I press _done_ button, the IME disappeared and it can't be reattached again.

Am I doing something wrong?

pShota:
Hi andyque,

thanks for testing the patch,
if IMEDelegate::canDetachWithIME()  is returning false,
then when  "TextFieldTTF::detachWithIME()" method is called (step 6 on my previous comment)
"IMEDelegate::detachWithIME()" should return false because of that, and it should skip "glView->setIMEKeyboardState(false);" 

is this happened in your case?

`
bool TextFieldTTF::detachWithIME()
{ 
    bool ret = IMEDelegate::detachWithIME();
    if (ret)
    {
        // close keyboard
        GLView * glView = Director::getInstance()->getOpenGLView();
        if (glView)
        {
            glView->setIMEKeyboardState(false);
        }
    }
    return ret;
}
`

pShota:
by the time Im writing this, I found another keyboard bug,
in some android devices(SONY Xperia S (LT26i)), 
the keyboard has a "hide" button similar to the one in iPad,
when the button is pressed, it seems like no callback from it

minggo:
@andyque 
What's your opinion?

@pShota 
About the `hide` button on some android devices. Is it a standard key button?

pShota:
@minggo 
Im not very sure, it looks like this
screenshot_2014-08-19-10-20-31

pShota:
@minggo
done button
screenshot_2014-08-19-10-42-15

zilongshanren:
@pShota 
Thanks , I will take a deep look at the execution flow. 
After that I will post my research here.
@minggo 
I'm not sure about the  "hide" button, I will test the normal "hide" button on my Android phone to see whether it has callback or not.

zilongshanren:
@pShota 
Hi, I have done some research, here is my result:
1. After applying your patch, when I return false from canDetachWithIME method, when the done button is clicked, the IME won't disappear. It works as expected. But if I press the done button multiple times, it will insert too many copies of string into the TextField control which is not the desired behavior. Because the same logic works fine on IOS platform.

Have you encountered this issue?

pShota:
@andyque 
yes, i can reproduce it now, it is very weird that it only happen when the textfield was opened at the first time , the 2nd time it will not have this bug

pShota:
@andyque 
i think i cant come up with a "perfect solution" because I was limited by the current logic flow, I would implement a "doneWithText" method instead of using "insertText" for done button and hardware back button
for now i can write the verify text code when the text is actually use e.g. I verify the text when player press the login button of my login popup instead of pressing done button of keyboard. (also no clue for hide button, look like we need hell a lot of modification : http://felhr85.net/2014/05/04/catch-soft-keyboard-showhidden-events-in-android/)
what do you guys think?

zilongshanren:
@pShota 
If the fix will cause the wrong behavior, I think it's safe to leave it here.
After we propose a perfect solution, we can merge it. 
Thanks.

zilongshanren:
@minggo  I think we should close this PR now.

PR #7577, opened 2014-07-24
fix mac compile error

dabingnn:


dabingnn:
[ci rebuild]

minggo:
It is improved in #7581.

PR #7578, opened 2014-07-24
mac custom font fix v2

newnon:
just truncate name internally as in iPhone to load font with path but you still need to add all fonts to .plist file
another version https://github.com/cocos2d/cocos2d-x/pull/7557 that load font directly by path you don't even need to add it to plist file
merge only one


WenhaiLin:
[ci rebuild]

heliclei:
[ci rebuild]

WenhaiLin:
@minggo  I think we could merge this PR.

PR #7582, opened 2014-07-24
Allow setting bundle to use in file utils

uorbe001:
Our cocos app has an apparently uncommon use case where we need to be able to download non-essential assets for our app.  The best solution (for ios) I've been able to come up with is to download a bundle and load it for the cocos environment, but it doesn't currently support using any bundle other than the main one from what I've seen. 

This PR adds the ability to do just that, allowing the client code to call something like this to set it:

```
((cocos2d::FileUtilsApple*)(cocos2d::FileUtils::getInstance()))->setBundle(bundle);
```

As it is something specific to apple, I added the change to the FileUtilsApple, which forces you to cast the FileUtils to the subclass for apple. I'm not a big fan of this, so I'd welcome ideas to avoid having to do it.

This is an uncommon use case, but I've seen other people trying to do the same thing online, so I thought it'd be interesting to get it on the main repository. Let me know if there is any changes/improvements needed to get this  merged.

### DIFF
--- a/cocos/platform/apple/CCFileUtilsApple.h
+++ b/cocos/platform/apple/CCFileUtilsApple.h
@@ -2,19 +2,19 @@
  Copyright (c) 2010-2012 cocos2d-x.org
  Copyright (c) 2011      Zynga Inc.
  Copyright (c) 2013-2014 Chukong Technologies Inc.
- 
+
  http://www.cocos2d-x.org
- 
+
  Permission is hereby granted, free of charge, to any person obtaining a copy
  of this software and associated documentation files (the "Software"), to deal
  in the Software without restriction, including without limitation the rights
  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


minggo:
Thanks.
So the codes can only be invoked in .m or .mm file.
How will you use it?

uorbe001:
@mingo I'm setting the bundle from the ios view that is creating the cocos2dx context. In my case, I have a view based on [`RootViewController`](https://github.com/cocos2d/cocos2d-x/blob/v3/templates/cpp-template-default/proj.ios_mac/ios/RootViewController.mm) which creates the cocos context and sets this bundle, allowing me to use different bundles and start the cocos app with different scenes.

It'd probably be better to be able to set the bundle from the `AppDelegate` in cpp, but that'd mean coming up with a bundle concept for all the supported platforms afaik (we are only supporting ios for now, I haven't even looked at how to solve this problem for other platforms yet).

minggo:
Got it.
Thanks.

PR #7585, opened 2014-07-25
Update CCParticleSystem.cpp

wagulu:
diff is base on WorldSpace, when you scale scene,you will get the wrong value.

### DIFF
--- a/cocos/2d/CCParticleSystem.cpp
+++ b/cocos/2d/CCParticleSystem.cpp
@@ -771,7 +771,12 @@ void ParticleSystem::update(float dt)
 
                 Vec2    newPos;
 
-                if (_positionType == PositionType::FREE || _positionType == PositionType::RELATIVE)
+                if (_positionType == PositionType::FREE)
+                {
+                    Vec2 diff = converToNodeSpace(currentPosition) - convertToNodeSpace(p-> startPos);
+                    newPos = p->pos - diff;
+                }
+                else if(_positionType == PositionType::RELATIVE)
                 {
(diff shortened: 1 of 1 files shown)


minggo:
[ci rebuild]

cocos/2d/CCParticleSystem.cpp
```diff
@@ -771,7 +771,12 @@ void ParticleSystem::update(float dt)
 
                 Vec2    newPos;
 
-                if (_positionType == PositionType::FREE || _positionType == PositionType::RELATIVE)
+                if (_positionType == PositionType::FREE)
+                {
+                    Vec2 diff = converToNodeSpace(currentPosition) - convertToNodeSpace(p-> startPos);
```
boyu0: Why not use
`Vec2 diff = convertToNodeSpace(currentPosition - p-> startPos);`
it can save one convert computation.
By the way, you fist `converToNodeSpace` is missing a `t`.

wagulu: This is not the same.You can't replace with one convert,when the scene(or particlesystem's parentnode) has be scaled.
Please ignore my English grammar mistakes.‘(_>﹏<_)′

wagulu:
Detail :http://blog.csdn.net/cllz334/article/details/38071507

cocos/2d/CCParticleSystem.cpp
boyu0: Yes, u r right, it's not the same...
@minggo I think this is ok for merge.

PR #7587, opened 2014-07-25
Add getter and setter for TextColor to Text Widget

dzl-ian:
Add a function to set text color property for label inside Text widgets.
The "Color Tint" option in CocoStudio sets the color for the entire widget (setColor method); setting text color this way will cause outline colors for Text widget to be incorrect.

### DIFF
--- a/cocos/ui/UIText.cpp
+++ b/cocos/ui/UIText.cpp
@@ -208,7 +208,7 @@ TextVAlignment Text::getTextVerticalAlignment()const
     return _labelRenderer->getVerticalAlignment();
 }
     
-void Text::setTextColor(Color4B color)
+void Text::setTextColor(const Color4B color)
 {
     _labelRenderer->setTextColor(color);
 }
--- a/cocos/ui/UIText.cpp
+++ b/cocos/ui/UIText.cpp
@@ -213,7 +213,7 @@ void Text::setTextColor(Color4B color)
     _labelRenderer->setTextColor(color);
 }
     
-const Color4B& Text::getTextColor()const
+const Color4B& Text::getTextColor() const
 {
     return _labelRenderer->getTextColor();
 }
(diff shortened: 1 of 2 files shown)


cocos/ui/UIText.cpp
```diff
@@ -207,6 +207,18 @@ TextVAlignment Text::getTextVerticalAlignment()const
 {
     return _labelRenderer->getVerticalAlignment();
 }
+    
+void Text::setTextColor(Color4B color)
+{
+    _labelRenderer->setTextColor(color);
+    updateContentSizeWithTextureSize(_labelRenderer->getContentSize());
```
zilongshanren: Does setTextColor method change the label's contentSize? I think we could safely omit the calling of `updateContentSizeWithTextureSize` method.

dzl-ian: You're definitely right - will adjust that.

cocos/ui/UIText.cpp
```diff
@@ -207,6 +207,16 @@ TextVAlignment Text::getTextVerticalAlignment()const
 {
     return _labelRenderer->getVerticalAlignment();
 }
+    
+void Text::setTextColor(Color4B color)
+{
+    _labelRenderer->setTextColor(color);
+}
+    
+const Color4B& Text::getTextColor()const
```
minggo: How about add a space between `()` and `const`?

cocos/ui/UIText.cpp
```diff
@@ -207,6 +207,16 @@ TextVAlignment Text::getTextVerticalAlignment()const
 {
     return _labelRenderer->getVerticalAlignment();
 }
+    
+void Text::setTextColor(Color4B color)
```
minggo: Color4B color -> const Color4B& color

cocos/ui/UIText.h
```diff
@@ -161,6 +161,10 @@ class CC_GUI_DLL Text : public Widget
 
     TextVAlignment getTextVerticalAlignment()const;
     
+    void setTextColor(Color4B color);
+    
+    const Color4B& getTextColor()const;
```
minggo: Please add a space between `()` and `const`

zilongshanren:
Thanks @dzl-ian 
@minggo  I think we could merge this PR now.

PR #7590, opened 2014-07-25
Simplify Widget ContentSize cascade change algorithm

zilongshanren:
This PR removed all the redundant calling of  `onSizeChanged` method,
Since we have already called it at the end of `setContentSize` method in Widget class.
And also I added some tests to verify this refactor code.


minggo:
[ci rebuild]

zilongshanren:
[ci rebuild]

PR #7591, opened 2014-07-25
Improve Widget setEnabled method

dzl-ian:
UIButton class has onPressStateChanged methods that would never actually be called.
Fixed the setEnabled function in the UI Widget base so that these methods are called.  
No other Widget classes override setEnabled so this shouldn't introduce any other issues.


dzl-ian:
Sorry - made a little mistake branching, will re-do the PR correctly.

PR #7592, opened 2014-07-25
Add code to change press state of widgets

dzl-ian:
UIButton class has onPressStateChanged methods that would never actually be called.
Fixed the setEnabled function in the UI Widget base so that these methods are called.
No other Widget classes override setEnabled so this shouldn't introduce any other issues.

### DIFF
--- a/cocos/ui/UIWidget.cpp
+++ b/cocos/ui/UIWidget.cpp
@@ -246,6 +246,11 @@ Widget* Widget::getWidgetParent()
 void Widget::setEnabled(bool enabled)
 {
     _enabled = enabled;
+    if (_enabled){
+        onPressStateChangedToNormal();
+    } else {
+        onPressStateChangedToDisabled();
+    }
 }
     
 void Widget::initRenderer()


cocos/ui/UIWidget.cpp
```diff
@@ -246,6 +246,11 @@ Widget* Widget::getWidgetParent()
 void Widget::setEnabled(bool enabled)
 {
     _enabled = enabled;
+    if (_enabled){
```
zilongshanren, line 249: Thanks, @dzl-ian 
If we call `setEnable` method in TouchBegin method, the button's state will be wrong. 

```
{
    _hitted = false;
    if (isVisible() && isEnabled() && isAncestorsEnabled() && isAncestorsVisible(this) )
    {
        _touchBeganPosition = touch->getLocation();
        if(hitTest(_touchBeganPosition) && isClippingParentContainsPoint(_touchBeganPosition))
        {
            _hitted = true;
        }
    }
    if (!_hitted)
    {
        return false;
    }
    setHighlighted(true);
    Widget* widgetParent = getWidgetParent();
    if (widgetParent)
    {
        widgetParent->interceptTouchEvent(TouchEventType::BEGAN, this, touch);
    }
    pushDownEvent();
    return true;
}
```

We call setHighlighted(true) in onTouchBegan method which will cause button's state changing from normal  to pressed, then the callback will be called in pushDownEvent. If we call setEnabled(true) in the pushDownEvent method, the button's state will change back to normal according to your modification.  That might bring new bugs.

zilongshanren:
When the Button is disabled, we should change its state to disabled. But we might need to consider another way to fix this issue.

minggo:
@dzl-ian 
How do you think about @andyque 's suggestion?

dzl-ian:
Sorry for the delay in response - I will take another look at this in the next couple days.

zilongshanren:
@dzl-ian 
We couldn't call `widget->setEnabled(false)` to change the widget to  disabled state right now.

I think its a historical design flaw, we call `widget->setEnabled(false)` to disable touch event,
and we must call `widget->setBright(false)` to make the appearance to disabled state.

*\* We have divided the disabled state to two stage(touch and appearance)**

And in Apple's documentation, the `setEnabled` method will disable touch and change appearance on default, but it has options to turn off the default behavior of changing appearance to disabled state.

So, to keep back-compatibility, it's better to keep the current state and tell the difference in our document.

dzl-ian:
Ahhhh well now I feel a little bit foolish, had no idea about the `setBright` method!
Will go ahead and close this pull request then as there was no issue in the first place.

zilongshanren:
@dzl-ian  All right, we might need to improve the API design in the future.

PR #7596, opened 2014-07-26
Fix the default value for project creator's script

timotei:


timotei:
Any idea how to see where exactly the build fails? It shouldn't be related to my modification.

minggo:
Thanks.

PR #7602, opened 2014-07-28
[Don't merge]New Audio Engine API

WenhaiLin:


cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,227 @@
...
+/**
+  @class          AudioEngine
+  @brief          
+  @note           
+ */
+
+class EXPORT_DLL AudioEngine
+{
+public:
+    enum SoundState
```
minggo: Why not use `enum class`?

cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,227 @@
...
+     * Release the shared Audio Engine object
+     @warning It must be called before the application exit
+     */
+    static void destroyInstance();
+
+    /** Play 2d sound
+     * @param filePath The path of a audio file
+     * @param loop Whether audio instance loop or not
+     * @param volume volume value (range from 0.0 to 1.0)
+     * @return a audio ID. It allows you to dynamically change the behavior of a audio instance on the fly.
```
minggo: a -> an

cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,227 @@
...
+     @param audioID a audioID returned by the play2d/play3d function
+     @param loop Whether audio instance loop or not
+     */
+    void setLoop(int audioID, bool loop);
+
+    /** Checks whether a audio instance is loop.
+     * @param audioID a audioID returned by the play2d/play3d function
+     * @return 
+     */
+    bool isLoop(int audioID);
```
minggo: Please add const.

cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,227 @@
...
+     * @param audioID a audioID returned by the play2d/play3d function
+     * @param volume volume value (range from 0.0 to 1.0)
+     */
+    void setVolume(int audioID, float volume);
+
+    /** Gets the volume value of a audio instance.
+     * @param audioID a audioID returned by the play2d/play3d function
+     * @return volume value (range from 0.0 to 1.0)
+     */
+    float getVolume(int audioID);
```
minggo: add const please

cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,227 @@
...
+     * @param audioID a audioID returned by the play2d/play3d function
+     * @return 
+     */
+    void setCurrentTime(int audioID, float time);
+
+    /** Gets the current playback position of a audio instance.
+     * @param audioID a audioID returned by the play2d/play3d function
+     * @return the current playback position of a audio instance
+     */
+    float getCurrentTime(int audioID);
```
minggo, line 131: add const please

cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,227 @@
...
+     * @param audioID a audioID returned by the play2d/play3d function
+     * @return the status of a audio instance
+     */
+    SoundState getState(int audioID);
+
+    /** Register a callback to be invoked when a audio instance has completed playing.
+     * @param audioID a audioID returned by the play2d/play3d function
+     * @param callback 
+     */
+    void setFinishCallback(int audioID, const std::function<void(int,const std::string&)>& callback);
```
minggo, line 149: What's the parameter meaning inside callback?

cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,227 @@
...
+
+    /** Place listener
+     * In case a Vec2 position is passed in, the z value will always be 0.
+     * @param position a Vec2 position
+     */
+    void setEar(Vec2 position);
```
minggo: Please add const, and use reference.

cocos/audio/include/AudioEngine.h
```diff
@@ -0,0 +1,229 @@
...
+
+#include <string.h>
+
+#include "2d/CCNode.h"
+
+NS_CC_BEGIN
+namespace experimental {
+
+/**
+  @class          AudioEngine
```
gelldur, line 36: Small proposition. We use soundID to control sounds  maybe we should use classes. I mean something like this.

```
class Sound
{
    bool isLooping()...
    int getSoundType() //for example return is this sound 2D or 3D

    int _volume = 1;

//other stuff
}
```

It would be easier to compose such sound than raw int ID also management would be easier . In future we could add more stuff to class and now we can't all the time change play\* method. Simply use composition for Sound

gelldur, line 36: Also it would be easier for example

```
Sound moneySound;

moneySound.play()
moneySound.stop() 
```

etc

minggo, line 36: It sounds good.

minggo, line 36: But the problem may be that, if you want to play a sound twice, you should create two instances for each sound. And there may be many sound instances.

minggo, line 36: And should we stop sound in destructor? If not, then you should keep the instance if you only want to play a sound.

So i think use a Class for a sound is not a good idea.

gelldur, line 36: I was thinking about Sound class as a holder for ID. Idon't know use case for multiple sounds effects. For example when i want play sound of collecting coin i play sound with ID of coin sound when i want stop this sound i want stop all instances of this sound.

If you need to stop explicit one sound and other keep still playing than yes my idea is a bad idea :)

gelldur, line 36: I mean instance of Sound is only configuration for specyfic sound.

darkdukey:
1. Can we not merge new features into `v3` directly? Instead having a `v3develop` branch?
2. I think gelldur's suggestion is a good one, we should think about it. Like sprite, we just let user create sprite with image and don't have to deal with texture id, maybe we should do that, It's much easier to manage. 

how to play multiple sound, it's up to developer to decide how they want to use that.

``` cpp
auto img = sprite::create("image.png");
addChild(img)

auto sound1 = sound::create("sound1.mp3");
sound1->play()

auto sound2 = sound::create("sound2.mp3";
sound2->play()
```

Compare to

``` cpp
int sound1 = audio::play2d("sound1.mp3");
audio::stop(sound1)

int sound2 = audio::play2d("sound2.mp3";
audio::stop(sound2)
```

In fact libGDX, flash, Unity, you can all create sound and use something like `sound->play()`

minggo:
libGDX doesn't work as you said.
I have replied you in the email.

About v3deveoper branch. I think we can work as you suggested after releasing v3.3. Because there are many PR merged after v3.2. And we will merge a new feature when it is ready.

minggo:
@darkdukey 

If using class, when i want to play a sound, i have to

``` c++
auto sound = Sound::create("sound.mp3");
sound.play();
```

ricardoquesada:
I think we can have both APIs.
eg:

``` c++

namespace sound {

int create(const std::string& file);
void play(int handle);
void stop(int handle);
void setVolume(int handle);

// helper for newbies
int play(const std::string& file)
{
   int handle = create(file);
   play(handle);
   return handle;
}

}

```

WenhaiLin:
@ricardoquesada I think we don't need  'create' function. We provide one API for play audio, developers don't need cares about the implementation details[create,load,streaming,decode etc.].

PR #7606, opened 2014-07-28
Reuse file utils when resolving video path (ios)

uorbe001:
`UIVideoPlayer` is repeating code that is already implemented on FileUtils, it seems to make more sense to reuse that one rather than having almost the same code in two different places.


minggo:
[ci rebuild]

darkdukey:
[ci rebuild]

PR #7616, opened 2014-07-29
Fix iOS VideoPlayer memory leak.

gin0606:
# Before
- `MPMoviePlayerController alloc] init~~` -> retain count equal 1.
- set to `strong` property  -> retain count to 2.
- set nullptr to `self.moviePlayer` -> retain count to 1.
- **Leak!**
# After
- `MPMoviePlayerController alloc] init~~` -> retain count equal 1.
- call `autorelease` -> retain count equal 0 when out of autoReleasePool.
- set to `strong` property  -> retain count to 1.
- set nullptr to `self.moviePlayer` -> retain count to 0.
- **MemFree**

### DIFF
--- a/cocos/ui/UIVideoPlayerIOS.mm
+++ b/cocos/ui/UIVideoPlayerIOS.mm
@@ -133,11 +133,11 @@ using namespace cocos2d::experimental::ui;
     }
     
     if (videoSource == 1) {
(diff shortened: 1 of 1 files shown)


WenhaiLin:
@gin0606 You are right.Thank you very much.

WenhaiLin:

PR #7622, opened 2014-07-30
fix LabelTTF::getBoundingBox() position error

rny:
this is because getContentSize() is required to refresh the position first.


minggo:
[ci rebuild]

darkdukey:
[ci rebuild]

minggo:
@rny 
We are all agree to use #7652.

minggo:
Sorry, they are different issues.

PR #7627, opened 2014-07-30
refactor ui::widgets to use ui::Scale9Sprite instead of extension::Scale9Sprite

zilongshanren:
1. Since many Widget classes use Sprite & Scale9Sprite at the same time, when the user change Scale9Enabled properties, we removed the previous added sprites and re-added them again. Now the ui::Scale9Sprite supports `setScale9Enabled` method, so I changed the old implementation.
2. In this PR, I also had fixed some minor bugs of ui::Scale9Sprite and added the tests


minggo:
[ci rebuild]

PR #7631, opened 2014-07-30
added 3D rotation support to the RotateTo action

maltium:
Currently the _RotateTo_ action does not work for 3D rotations. This commit adds a new static _create()_ method that accepts a _Vec3()_ as input to enable animated rotations in 3D.

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -733,6 +733,29 @@ RotateTo* RotateTo::create(float duration, float deltaAngle)
     return rotateTo;
 }
 
+RotateTo* RotateTo::create(float duration, float deltaAngleX, float deltaAngleY)
+{
+    RotateTo* rotateTo = new RotateTo();
+    rotateTo->initWithDuration(duration, deltaAngleX, deltaAngleY);
+    rotateTo->autorelease();
+    
+    return rotateTo;
+}
+
+RotateTo* RotateTo::create(float duration, const Vec3& deltaAngle3D)
+{
+    RotateTo* rotateTo = new RotateTo();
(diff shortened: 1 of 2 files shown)


cocos/2d/CCActionInterval.cpp
```diff
@@ -733,6 +733,29 @@ RotateTo* RotateTo::create(float duration, float deltaAngle)
...
+{
+    RotateTo* rotateTo = new RotateTo();
+    rotateTo->initWithDuration(duration, deltaAngle3D);
+    rotateTo->autorelease();
+    
+    return rotateTo;
+}
+
+RotateTo::RotateTo()
+: _is3D(false)
```
minggo, line 755: I think it is better to initialize all parameters.

maltium:
@minggo, my initial idea was to use a Vec3() for both 2D and 3D since 2D is really a subset of 3D. However I looked at the code for _RotateBy_ and for consistency, I implemented it the same way. I will change both if you prefer.

minggo:
Yep. I think it is better to initialize all parameters in constructor though they will be changed in initialization functions. You are appreciated if you can change both of them and add corresponding test case.

maltium:
@minggo, I'm now using Vec3() exclusively so there are no issues with uninitialized variables for _RotateTo_ and _RotateBy_

minggo:
[ci rebuild]

maltium:
@minggo, why is the build failing on Jenkins?

minggo:
[ci rebuild]

minggo:
@maltium 
It is not your fault.
Because the building server lies in China. And it sometimes has problem of connecting to github.
Sorry about that.

minggo:
[ci rebuild]

PR #7643, opened 2014-08-01
modify for loading bar render error

liamcindy:


zilongshanren:
Thanks @liamcindy 
The tests of loading bar are not rendering wrong? Is your modification just a refactoring?

liamcindy:
if you change the direction of loading bar , the position of loading bar
will be changed;

zilongshanren:
Thanks @liamcindy 
You'd better split the loadingBar pull request and slider pull request into different branches.

zilongshanren:
@liamcindy 
And I think we shouldn't change the anchor point of y. It should be at 0.5, otherwise the behavior is not the same as before. I will make another pull request to fix this issue. Once it is finished, I will paste the link here.

Thanks again for your reporting.

zilongshanren:
The loading bar issue is fixed at : https://github.com/cocos2d/cocos2d-x/pull/7712

minggo:
Can close this PR?

zilongshanren:
@minggo  I think so. We could close this PR.
@liamcindy  We will reconsider the Slider Ball size change issue with the studio team. Before that, it is very risky to merge your PR.

PR #7652, opened 2014-08-02
bug fix for LabelBMFont::getBoundingBox()

zhiqiangxu:
`LabelBMFont::getBoundingBox()` is the bounding box of LabelBMFont relative to its parent.

while `_label->getBoundingBox()` is the bounding box of _label relative to LabelBMFont.

Although the size of the above 2 boxes are the same, their offset are not necessarily the same.

(`_label` is always positioned at `Point::ZERO` of `LabelBMFont`, which is not necessarily the case for `LabelBMFont`)

This bug can cause that we can't correctly detect whether a `LabelBMFont` is clicked or not:http://www.cocoachina.com/bbs/read.php?tid=209406

### DIFF
--- a/cocos/2d/CCLabelBMFont.cpp
+++ b/cocos/2d/CCLabelBMFont.cpp
@@ -204,7 +204,7 @@ const Size& LabelBMFont::getContentSize() const
 
 Rect LabelBMFont::getBoundingBox() const
 {
-    return _label->getBoundingBox();
+    return Node::getBoundingBox();
 }
 #if CC_LABELBMFONT_DEBUG_DRAW
 void LabelBMFont::draw(Renderer *renderer, const Mat4 &transform, uint32_t flags)


cocos/2d/CCLabelBMFont.cpp
```diff
@@ -204,7 +204,7 @@ const Size& LabelBMFont::getContentSize() const
 
 Rect LabelBMFont::getBoundingBox() const
 {
-    return _label->getBoundingBox();
```
WenhaiLin, line 207: Thank you for report this isue.
Please consult https://github.com/cocos2d/cocos2d-x/pull/7622/files for fix this issue.

rny:
i think it is a similar issue #7622

cocos/2d/CCLabelBMFont.cpp
zhiqiangxu, line 207: yes,it's a similar issue, except that `const_cast<LabelTTF*>(this)->setContentSize(_renderLabel->getContentSize());` is not needed, as it's already called in various methods like `LabelBMFont::initWithString`, `LabelBMFont::setString` etc.

darkdukey:
I agree with @zhiqiangxu reset contentSize is not needed

minggo:
@Dhilan007 
I think reset content size is not needed too.
What's your opinion?

WenhaiLin:
@minggo I agree with @zhiqiangxu .

PR #7653, opened 2014-08-03
Update : added function for setting min/max scale

takaken1994:
- Added function for setting min/max scale.
- Fix to remove unused params.

### DIFF
--- a/extensions/GUI/CCScrollView/CCScrollView.cpp
+++ b/extensions/GUI/CCScrollView/CCScrollView.cpp
@@ -51,10 +51,7 @@ static float convertDistanceFromPointToInch(float pointDis)
 
 
 ScrollView::ScrollView()
-: _zoomScale(0.0f)
-, _minZoomScale(0.0f)
-, _maxZoomScale(0.0f)
-, _delegate(nullptr)
+: _delegate(nullptr)
 , _direction(Direction::BOTH)
 , _dragging(false)
 , _container(nullptr)
@@ -65,6 +62,7 @@ ScrollView::ScrollView()
 , _minScale(0.0f)
 , _maxScale(0.0f)
 , _touchListener(nullptr)
+, _scissorRestored(false)
 {
 
 }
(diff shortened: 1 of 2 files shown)


extensions/GUI/CCScrollView/CCScrollView.cpp
```diff
@@ -51,10 +51,7 @@ static float convertDistanceFromPointToInch(float pointDis)
 
 
 ScrollView::ScrollView()
-: _zoomScale(0.0f)
```
zilongshanren, line 54: I found that these deleted code are not used by any code in CCScrollView, should we also delete the declaration in the header file?  If we just leave the member variable uninitialized which I think is not a good idea.

extensions/GUI/CCScrollView/CCScrollView.h
```diff
@@ -146,6 +146,34 @@ class CC_EX_DLL  ScrollView : public Layer, public ActionTweenDelegate
...
      */
     void setZoomScaleInDuration(float s, float dt);
+
+    /**
+     * Set min/max scale
+     *
+     * @param minScale min scale
+     * @param maxScale max scale
+     */
+    void setMinMaxScale(float minScale, float maxScale) {
```
zilongshanren: Regards to this API, it seems a little verbose if we have had the `setMinScale` and `setMaxScale` APIs.

takaken1994: I see. I will remove this function.

extensions/GUI/CCScrollView/CCScrollView.cpp
takaken1994, line 54: I had forgotten to remove them from the header file. I'll do it. Thank you!

minggo:
[ci rebuild]

zilongshanren:
Thanks @takaken1994 
@minggo  I think we could merge this PR now.

PR #7658, opened 2014-08-05
fix scale9Sprite anchorPoint bug when switch anchor point and scale9Enabled

zilongshanren:
1. fixed issue when we change ui::Scale9Sprite's anchor point and scale9Enabled at the same time, it will not behavior correctly.
2. add tests to prove this fix
3. refactor: improve indent to meet our coding style and fixes some typo.


zilongshanren:
[ci rebuild]

PR #7666, opened 2014-08-05
Template fix

zawasp:
- fix wrong libcocos2d project reference
- fix wrong libcocosdenshion project reference
- add VS 2013 support for template project (backwards compatible with VS2010, VS2012)


minggo:
[ci rebuild]

PR #7668, opened 2014-08-05
New Particle System

boyu0:
DO NOT MERGE.
for review.

Defined some classes that can be used for particle:
`Graph<T>`
`LineGraph<T>`
`BezierGraph<T>`
`FloatBezierGraph`
`Gradient`

Add some particle classes:
`Particle`
`ParticleEffector`
`ParticleEmitter`
`ParticleRendererProtocol`
`ParticleTextureAtlasRenderer`


darkdukey:
Are there some kind of design doc? Or some examples of how to use it?

boyu0:
@darkdukey 
Design doc is:
https://docs.google.com/a/cocos2d-x.org/document/d/1vrBwdPumDY6o21vawNmOxQJVTp9ZqYbCL1xCJxRBuBI/edit?usp=sharing

boyu0:
Sorry @darkdukey , google doc link is unstable in China, I couldn't get in, so I put my answer here.
1. Isn't Line curve just a type of Bezier Curve? Called LinearCurve
   Yes, but the control point structure for bezier curve need more information(two more T space), so I think sometimes we might be not need the bezier curve( like color gradient ), so I add this.
   LinearCurve is better, I will change the name, thanks.
2. Need a use case to understand how it works
   I will add tests after finish this implementation.
3. (vector<T> _compiled;) What does this mean?
   We can use some sample x to compute the _func result and store them in _compiled, we can get the result later without run _func each time when we need it ( if x is between x[i] to x[i+1], it will return _compiled[i] for result, so it not accurate, but save the times).

4.(T y) Why use T?
I think T is more useful, we can use it with  float, Vec2, Color, etc...

darkdukey:
Thanks Boyu

For No.2 can you add some example in the documentation? Even pseudo code is fine.

-Nite

boyu0:
@darkdukey 
I can't access the google doc link, I will update the doc and write example when I can.
I will let you know after I update it.
If I still can't access the google doc, tomorrow I will write example here...

boyu0:
@darkdukey 
I updated the google doc, and add a Gradient example at the end of doc.

zii:
@boyu0 
I bought a Linode server, we can use it to jump the wall.
1. Download Stunnel client. https://www.stunnel.org/index.html
2. Edit config file and add this at the end:
[https]
client = yes
accept  = 8888
connect = 23.239.7.245:443
3. Use chrome plugin SwitchySharp, add a proxy on localhost:8888

PS. best not watch video.

boyu0:
@zii 
Great, I will try it, thank you.

PR #7669, opened 2014-08-05
Enable easy usage of alpha test builtin shader.

codesuki:
Added setAlphaTestValue() to Director and setup built-in uniform in CCShader.
Can be used together with DepthTest to achieve better batching.

### DIFF
--- a/cocos/renderer/CCGLProgramState.h
+++ b/cocos/renderer/CCGLProgramState.h
@@ -206,7 +206,7 @@ public:
     void setAlphaTestValue(float alpha);
 
     /** gets the alpha value used in the alpha test shader */
-    float getAlphaTestValue();
+    float getAlphaTestValue() const;
 
 protected:
     GLProgramState();
--- a/cocos/renderer/CCGLProgramState.h
+++ b/cocos/renderer/CCGLProgramState.h
@@ -146,6 +146,8 @@ protected:
 class CC_DLL GLProgramState : public Ref
 {
     friend class GLProgramStateCache;
+
+    float _alphaTestValue;
 public:
 
     /** returns a new instance of GLProgramState for a given GLProgram */
@@ -198,6 +200,13 @@ public:
     void setUniformCallback(GLint uniformLocation, const std::function<void(GLProgram*, Uniform*)> &callback);
     void setUniformTexture(GLint uniformLocation, Texture2D *texture);
     void setUniformTexture(GLint uniformLocation, GLuint textureId);
+
+    // setting an alpha test value only is effective if the alpha test shader is used
+    /** sets the alpha value used in the alpha test shader */
+    void setAlphaTestValue(float alpha);
+
+    /** gets the alpha value used in the alpha test shader */
+    float getAlphaTestValue();
 
 protected:
     GLProgramState();
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -710,6 +710,16 @@ void Director::setDepthTest(bool on)
     CHECK_GL_ERROR_DEBUG();
 }
 
+void Director::setAlphaTestValue(float alpha)
+{
+    _alphaTestValue = alpha;
+}
+
+float Director::getAlphaTestValue()
+{
+    return _alphaTestValue;
+}
+
 static void GLToClipTransform(Mat4 *transformOut)
 {
     if(nullptr == transformOut) return;
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -710,16 +710,6 @@ void Director::setDepthTest(bool on)
     CHECK_GL_ERROR_DEBUG();
 }
 
-void Director::setAlphaTestValue(float alpha)
-{
-    _alphaTestValue = alpha;
-}
-
-float Director::getAlphaTestValue()
-{
-    return _alphaTestValue;
-}
-
 static void GLToClipTransform(Mat4 *transformOut)
 {
     if(nullptr == transformOut) return;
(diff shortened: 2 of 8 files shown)


darkdukey:
[ci rebuild]

minggo:
@codesuki 
Thanks, it is a good idea to make set alpha value more convenient. But i am not sure if Director:: setAlphaTestValue() is a good idea. I think we should make Director less relation with render issues. That's why we have a new renderer. Our final target is that, all render related things are put into renderer.

codesuki:
@minggo 
Thank you for your comment. I thought so, too. I just added it there since I found setDepthTest there.
Anyhow, where would you put the setAlphaTestValue method?  Maybe ProgramState?

minggo:
@codesuki 
Yep, i think ProgramState is a good place, and it is designed for this purpose. We have used it for other uniforms.

codesuki:
Thanks for your feedback, I will amend the code as necessary.

minggo:
Any update?

codesuki:
Hey, I added the setAlphaTestValue() method to GLProgramState, but I am not completely happy with the way it is now. I think eventually the alpha test value should be a Uniform, but until now only user defined uniforms are converted to uniforms and I did not want to change the code too much.

minggo:
@codesuki 
Sorry, i can not catch your idea.

codesuki:
@minggo 
Sorry after re-reading my comment above is confusing even for me. 
Of course it is an uniform inside the shader but I meant the Uniform struct inside GLProgram.h. 
Built-in shader uniforms are not parsed into GLProgram Uniforms. 
If the CC_alphaValue uniform (in the shader) would be reflected as a Uniform inside GLProgram it could be accessed naturally via GLProgramState->setUniform().

Did that get my idea across?

minggo:
Yep, built-in shaders use uniforms directly for performance.

codesuki:
Makes sense.
The only thing I didn't like now was the private float _alphaTestValue;
But somewhere the information needs to be held.

cocos/renderer/CCGLProgramState.h
```diff
@@ -199,6 +201,13 @@ class CC_DLL GLProgramState : public Ref
     void setUniformTexture(GLint uniformLocation, Texture2D *texture);
     void setUniformTexture(GLint uniformLocation, GLuint textureId);
 
+    // setting an alpha test value only is effective if the alpha test shader is used
+    /** sets the alpha value used in the alpha test shader */
+    void setAlphaTestValue(float alpha);
+
+    /** gets the alpha value used in the alpha test shader */
+    float getAlphaTestValue();
```
minggo: It is better to add `const`.

minggo:
> The only thing I didn't like now was the private float _alphaTestValue;
> But somewhere the information needs to be held.

Why it is a problem to declare as private?

codesuki:
@minggo 
The private is not the problem, just that there is the variable itself.
It's only needed for one built-in shader so i hesitated to add a member variable for that.

Thanks, I'll add the const.

minggo:
I have ask @dabingnn to take a look, and think about if there is a better way to do it.

dabingnn:
@codesuki 
Thanks for your pull request. I have several questions about this.
1. How to use it both in desktop(open GL) and mobile (GL es)?
2. In which cases, we need it?

codesuki:
@dabingnn 
Thanks for your comment.
1. It should work across both platforms since it's just doing "discard" in the shader. (Can be slow on some GPUs)
2. Imagine you have an isometric map with items (i.e. boxes) on it and a character. 
One box is behind and one box in front of the player.
Suppose the character and the item have separate textures. 
To be able to use 2 (player->boxes) draw calls instead of 3 (box->player->box) I first needed to enable depth testing.
Without alpha testing the translucent border of the player sprite will write to the depth buffer and create a black square instead of showing the box behind the player. (The box will not draw since the depth value of the player is in the depth buffer)

minggo:
@dabingnn 
Any question or good suggestion?

dabingnn:
Here is my considerations:
If we add `Alpha Test`, who owns the state, `GLProgram` or `GL Render context`? 
How to get/set the current state of `Alpha Test`(including the on/off and the ref value)?

In this pull request, it seems that `GLProgram`/`GLProgramState` will own the state, so the on/off state is controlled by the `shader`, depending on whether `discard` is used or not.
In the original openGL, `Alpha test` state can be set or get. This implementation means we can only get the on/off state, we can not set it.
But if user call `setAlphaTestValue`, they don't know whether the alpha test is on or off in the `GLProgram`. An interface to get the on/off state is needed.

minggo:
@codesuki 
Any thought?

codesuki:
Sorry super busy at the moment. I will come back later to this.

codesuki:
@minggo 
I didn't follow recent developments. If the feature is still viable then I will work on improving the implementation, let me know.

codesuki:
@minggo 
I guess you are busy? :)

minggo:
@codesuki 
Sorry for late reply.
Yep, we need any feature that will improve the engine.
@dabingnn is familiar with renderer codes, and he is core developer of cocos2d-x, so i asked him to help to review it.

codesuki:
Cool, I will wait for a reply and then continue with the patch/pullreq :)

minggo:
@codesuki 
I think @dabingnn had added his opinion, and how do you think about it?

codesuki:
Oh yeah, let me read again and think. Will comment here later today!

minggo:
@codesuki  Thanks for the PR. I think there is enough time to think about it. Now it has conflict, so please send a new PR if you it is still needed.

PR #7675, opened 2014-08-06
Chore/reuse file utils for video path

darkdukey:
Try to fix the build for #7606

Credits goes to @uorbe001


uorbe001:
Thanks @darkdukey, I didn't realize there was a problem with the build. I didn't have problems building it in my project, but I probably have different header search paths!

PR #7680, opened 2014-08-06
fix normalizedPosition bug

zilongshanren:


zilongshanren:
@darkdukey could you please take a look at this PR?
Does it fix the normalizedPosition bug?

minggo:
@ricardoquesada 
Is it reasonable to fix the bug like this?

ricardoquesada:
This patch breaks the `NodeNormalizedPositionTest2` test.

ricardoquesada:
Patch looks good to me. merging it.

PR #7683, opened 2014-08-06
Let Widget class don't support cascaded opacity and cascaded color on default.

zilongshanren:
Make Widget class disable cascaded color and opacity on default.


darkdukey:
Can you give a use case for it?

liamcindy:
some one want to set different color for an widget and its parent;

darkdukey:
Just to see if I understand correctly, correct me if I'm wrong
1. If I add a `slider` as a child of a `button`, if I change the color of `button`, slider's color won't change
2. If I change the color of a `slider`, the color of the `slider ball` will change.

zilongshanren:
@darkdukey 
This PR will guarantee the above two points.

darkdukey:
@andyque Sounds good thanks~

PR #7696, opened 2014-08-07
Fixed: when body mass infinity, setGravityEnable(false) does not work

varvelworld:
Fixed: when body mass infinity, function PhysicsBody::setGravityEnable(false) does not work

### DIFF
--- a/cocos/physics/CCPhysicsBody.cpp
+++ b/cocos/physics/CCPhysicsBody.cpp
@@ -324,7 +324,7 @@ void PhysicsBody::setGravityEnable(bool enable)
     if (_gravityEnabled != enable)
     {
         _gravityEnabled = enable;
-        _info->getBody()->gravity_enable = enable;
+        _info->getBody()->gravity_enable = enable ? cpTrue : cpFalse;
     }
 }
 
--- a/cocos/physics/CCPhysicsBody.cpp
+++ b/cocos/physics/CCPhysicsBody.cpp
@@ -289,13 +289,6 @@ void PhysicsBody::setDynamic(bool dynamic)
             
             if (_world != nullptr)
             {
-                // reset the gravity enable
-                if (isGravityEnabled())
-                {
-                    _gravityEnabled = false;
-                    setGravityEnable(true);
-                }
-                
                 cpSpaceAddBody(_world->_info->getSpace(), _info->getBody());
             }
         }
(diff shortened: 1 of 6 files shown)


minggo:
@varvelworld 
It is a bug of chipmunk?

varvelworld:
@minggo 
Chipmunk does not provide function to set gravity enable. In PhysicsBody, when the GravityEnable is false, give the body a anti-gravity force, I think that this implementation is not good. So I change the chipmunk code directly.

And why is win32_win7 building in jenkins success? but the others are failed.

minggo:
About jenkins, i think it is not your problem. It is about network issue.
@boyu0
Could you take a look about this PR?

minggo:
chipmunk is written in c, so you can not use bool, true.

boyu0:
I think change chipmunk code is not an option. we need to keep chipmunk easy to update, besides, we are going to build chipmunk to a ".a" library...
I think set body mass to infinity is not necessary, there are no "infinity mass object" in real world, I think only the static bodies need to set mass to infinity.
So I think we can keep this bug unfixed unless we find a way to fix it without change the chipmunk code...

minggo:
I agree with @boyu0, change chipmunk codes will make it hard to update.

varvelworld:
I agree with @boyu0 ,too. It's not eassy to update. But I think set body mass to infinity is necessary sometimes.
Now, the static body ( setDynamic(false) ) can't set velocity.I remember that in box2d the static body can has velocity.If I can set body mass to infinity, it is same with "static body" in box2d.

boyu0:
@varvelworld 
ok, so I think which we need is a movable static body, not a dynamic body with infinity mass, I will check is there a possible way to implement this...

varvelworld:
@boyu0 
Look this pr.Thank you!
https://github.com/cocos2d/cocos2d-x/pull/7707

minggo:
@varvelworld 
Can close this PR?

varvelworld:
@minggo  ok.
@boyu0  But I still think that it's a briefer way with modify chipmunk code directly, if we do not care update.

minggo:
@varvelworld 
Update 3rd party is important. We should make it easy to update.
Or will you send a patch to chipmunk?

varvelworld:
@minggo ok.I agree with you. I wiil try.

PR #7697, opened 2014-08-07
Camera and Camera test

super626:
1. add Camera and Camera Test
   support perspective camera and orthographic camera.
   it will create default camera for each scene. 
   camera usage:
   auto camera = Camera::createPerspective(60, (GLfloat)s.width/s.height, 1, 1000);
   camera->setCameraFlag(CameraFlag::CAMERA_USER1);
   layer->addChild(camera); //add this camera to where you want, it will take effect no the scene contain it
   layer->setCameraMask((unsigned short)CameraFlag::CAMERA_USER1);  //which camera can see the layer
2. add DrawNode3D to draw 3D primitives which is similar to DrawNode


cocos/base/CCCamera.h
```diff
@@ -0,0 +1,164 @@
...
+{
+    CAMERA_DEFAULT = 1,
+    CAMERA_USER1 = 1 << 1,
+    CAMERA_USER2 = 1 << 2,
+    CAMERA_USER3 = 1 << 3,
+    CAMERA_USER4 = 1 << 4,
+    CAMERA_USER5 = 1 << 5,
+    CAMERA_USER6 = 1 << 6,
+    CAMERA_USER7 = 1 << 7,
+    CAMERA_USER8 = 1 << 8,
```
minggo: I think we can remove the prefix `CAMERA`, because it is in a enum class, which is a strong type.

cocos/base/CCCamera.h
```diff
@@ -0,0 +1,164 @@
...
+
+    /** create default camera, the camera type depends on Director::getProjection */
+    static Camera* create();
+    
+    /**
+    * Gets the type of camera.
+    *
+    * @return The camera type.
+    */
+    Camera::Type  getCameraType() const;
```
minggo: How about change to getType(), because it is in class `Camera`.

cocos/base/CCCamera.h
```diff
@@ -0,0 +1,164 @@
...
+
+    /**get & set Camera flag*/
+    CameraFlag getCameraFlag() const { return (CameraFlag)_cameraFlag; }
+    void setCameraFlag(CameraFlag mask) { _cameraFlag = (unsigned short)mask; }
```
minggo: set flag or set mask?

cocos/base/CCCamera.h
```diff
@@ -0,0 +1,164 @@
...
+    const Mat4& getProjectionMatrix();
+    /**
+    * Gets the camera's view matrix.
+    *
+    * @return The camera view matrix.
+    */
+    const Mat4& getViewMatrix();
+
+    /**get view projection matrix*/
+    const Mat4& getViewProjectionMatrix();
```
minggo, line 124: Please add const for get functions.

cocos/base/CCDirector.cpp
```diff
@@ -300,9 +339,9 @@ void Director::drawScene()
     if (_displayStats)
     {
         showStats();
+        _renderer->render();
     }
 
-    _renderer->render();
```
minggo, line 305: Why move to `if (_displayStatas)`?

cocos/base/CCCamera.h
super626: It is flag to camera, and mask to Node.

cocos/base/CCCamera.h
super626: agree

cocos/base/CCCamera.h
super626: agree

cocos/base/CCCamera.h
super626, line 124: We can not add const to some of them. Because it uses lazy update sometimes

cocos/base/CCCamera.h
super626: Will we add getType to the Node? There are so many classes inherit from Node. It is may useful to know the exact class. 
We should use getCameraType in that way.

cocos/base/CCCamera.h
minggo, line 124: Ok

super626:
[ci rebuild]

cocos/base/CCCamera.h
dabingnn, line 124: using `mutable` can allow const added in this situation

cocos/base/CCDirector.cpp
```diff
@@ -283,11 +285,48 @@ void Director::drawScene()
...
+            {
+                defaultCamera = _currentCamera;
+                continue;
+            }
+            
+            pushMatrix(MATRIX_STACK_TYPE::MATRIX_STACK_PROJECTION);
+            loadMatrix(MATRIX_STACK_TYPE::MATRIX_STACK_PROJECTION, _currentCamera->getViewProjectionMatrix());
+            
+            //visit the scene
+            _runningScene->visit(_renderer, Mat4::IDENTITY, false);
```
minggo, line 306: false -> 0.

cocos/base/CCCamera.cpp
```diff
@@ -0,0 +1,274 @@
...
+        {
+            auto scene = dynamic_cast<Scene*>(parent);
+            if (scene)
+            {
+                setScene(scene);
+                break;
+            }
+            parent = parent->getParent();
+        }
+    }
```
minggo, line 238: There is Node::getScene()

PR #7700, opened 2014-08-07
Update CCSpriteBatchNode.cpp

GavinThornton:
Setting opacity of a sprite (via setOpacity or FadeOut/In action) has the sprite turning black, it should simply be opaque.  

Cause is in bool SpriteBatchNode::initWithTexture, third line, "if(tex->hasPremultipliedAlpha())" 
Fix is making it "!if(!tex->hasPremultipliedAlpha())"

Discussed here on the forums: http://discuss.cocos2d-x.org/t/in-cocos2dx-3-2final-sprite-become-black-when-the-opacity-is-less-than-255/16096


minggo:
Thanks.
I have joined the discussion.

PR #7702, opened 2014-08-07
Camera and Camera test

super626:
1. add Camera and Camera Test
   support perspective camera and orthographic camera.
   it will create default camera for each scene. 
   camera usage:
   auto camera = Camera::createPerspective(60, (GLfloat)s.width/s.height, 1, 1000);
   camera->setCameraFlag(CameraFlag::CAMERA_USER1);
   layer->addChild(camera); //add this camera to where you want, it will take effect no the scene contain it
   layer->setCameraMask((unsigned short)CameraFlag::CAMERA_USER1); //which camera can see the layer
2. add DrawNode3D to draw 3D primitives which is similar to DrawNode


super626:
[ci rebuild]

PR #7706, opened 2014-08-07
We can use C++11 random library

gin0606:
`rand()` is poor random number generators than C++11 random library.
And `rand()%N` is not good.
`rand()%N` is used in various classes.
Please look this URL about it. http://c-faq.com/lib/randrange.html

I think, We need to use C++11 random library.


minggo:
[ci rebuild]

minggo:
Thanks.
I will check if c++ 11 random is well supported in all supported platforms.

felixjingga:
if it's well supported on all supported platforms, you should add another macro that receive two parameters  to make the random engine produce result for random integer & float between min and max.

gin0606:
@felixjingga I implemented `cocos2d::random`. That receive two parameters to make the random value. `cocos2d::random<float>(0, 1)` return random value between 0 and 1.
C++ has `template` and `inline`. I think, we do not need to use C style macro.

gin0606:
@minggo I have no way to add ccRandom.cpp to winProject.
Because I don't have Windows PC.
Can you teach me to add ccRandom to winProject?

minggo:
Ok.
I will continue your work a few days later.
Sorry about for late reply.
I am busy these days.

gin0606:
No problem :)
Thank you your review.

zilongshanren:
Thanks @gin0606 
I will continue your work and add WinProject back.

zilongshanren:
@gin0606 
I have added windows project files.
Here is the PR: https://github.com/cocos2d/cocos2d-x/pull/7827

gin0606:
@andyque Thanks your work.
I will close this pull requst.

PR #7707, opened 2014-08-07
Let a static body can be moveable, without modify chipmunk code.

varvelworld:
Let a static body can be moveable, without modify chipmunk code.

### DIFF
--- a/cocos/physics/CCPhysicsBody.cpp
+++ b/cocos/physics/CCPhysicsBody.cpp
@@ -597,12 +597,6 @@ void PhysicsBody::addMoment(float moment)
 
 void PhysicsBody::setVelocity(const Vec2& velocity)
 {
-    if (!_dynamic)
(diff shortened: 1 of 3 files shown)


minggo:
[ci rebuild]

minggo:
@boyu0
Could you please review this PR?

boyu0:
Thank you for improve our physics intergration.
I ran your test, I found that the yellow box is walk through the wall without collision, and even doesn't invoke a contact callback function. I think this might be a problem.
And, I think damping should be work either...

varvelworld:
@boyu0 
The yellow box is walk through the wall, because both of them are static.So I think it's not a problem.The problem is that the collision callback function is not invoked.
I think the key is that a static body is not in "space".
I will have a try to fix it.

minggo:
Any update of this PR?

PR #7718, opened 2014-08-08
 Fixed multitouch not enabled for iOS in Cpp template

idimiter:
It seems that the C++ template is missing setMultipleTouchEnabled for the eaglView in AppController.mm. So the multi-touches does not work for iOS when I create new project with cocos-console.


minggo:
Yep. You should enable it if you need it.
It is not an error, it is just a selection.
If we change it, it will break compatibility.

PR #7719, opened 2014-08-08
fix PhysicsBody bitmask bug.

boyu0:
Delete bitmask in PhysicsBody, use PhysicsShape bitmask instead.


slug404:
fecth 2014-08-09的代码之后. 现在createWithPhysics出来的Scene全是黑的. 不能加sprite, 而且官方demo中Node:Physics 也显示不出来

boyu0:
@slug404 Thanks for report, I will fix this...

minggo:
It is fixed in #7723.

PR #7722, opened 2014-08-09
Support DOT syntax for getChildByName

zii:
When I load GUI from Cocos Studio directly, it is not very easy to get deeply widget.

So, how about this:
layer->getChildByName("root.panel.button")
layer->removeChildByName("root.panel.button")

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -822,24 +822,24 @@ Node* Node::getChildByName(const std::string& name) const
 {
     CCASSERT(name.length() != 0, "Invalid name");
 
-    std::string first_name = name;
-    std::string suffix = "";
+    std::string head = name;
+    std::string tail = "";
 
     int dotpos = name.find('.');
     if (dotpos > 0)
     {
-        first_name = name.substr(0, dotpos);
-        suffix = name.substr(dotpos+1);
+        head = name.substr(0, dotpos);
+        tail = name.substr(dotpos+1);
     }
     
     std::hash<std::string> h;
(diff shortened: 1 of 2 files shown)


cmzx3444:
Nice feature!

cocos/2d/CCNode.cpp
```diff
@@ -821,6 +821,16 @@ Node* Node::getChildByTag(int tag) const
...
+
+    int dotpos = name.find('.');
+    if (dotpos > 0)
+    {
+        first_name = name.substr(0, dotpos);
+        suffix = name.substr(dotpos+1);
+    }
     
     std::hash<std::string> h;
     size_t hash = h(name);
```
cmzx3444: size_t hash = h(name); 
the name should be first_name

cocos/2d/CCNode.cpp
```diff
@@ -829,7 +839,7 @@ Node* Node::getChildByName(const std::string& name) const
     {
         // Different strings may have the same hash code, but can use it to compare first for speed
         if(child->_hashOfName == hash && child->_name.compare(name) == 0)
```
cmzx3444: child->_name.compare(name) 
the name should be first_name

zii:
Thank you very much! I nearly make a big bug.

zii:
Is there faster implemention?

minggo:
@zii 
Thanks. But it will break the compatibility.
I think it is better to add a helper function for it.

zii:
You are right sir, it is a little complicated. Suite for script level.

PR #7723, opened 2014-08-09
Update CCScene.cpp

jimmystar:
in line 157, here need to create  and add camera, if not ,in the function createWithPhysics, will show black screen.


jimmystar:
fixed a bug about Scene::createWithPhysics(); function.

minggo:
Thanks.

PR #7732, opened 2014-08-11
add virtual destructor

zilongshanren:
If we don't add the virtual destructor, XCode will complains and generate many warnings.


zilongshanren:
[ci rebuild]

darkdukey:
[ci rebuild]

PR #7743, opened 2014-08-12
Update CCNode.h

zii:
Exchange the two declarations of "addChild". Let lua-bind priority generate for "string name" method.


minggo:
What's the benefit?

samuele3hu:
In the `lua_cocos2dx_Node_addChild`,we will judge the type of params, so the overload function with name param would be called,if the third param is string type.I will check it soon.

samuele3hu:
In the `lua_cocos2dx_Node_addChild`,we will judge the type of params, so the overload function with name param would be called,if the third param is string type.I will check it soon.

zii:
@samuele3hu 
Thank you!

PR #7749, opened 2014-08-12
virtual Node::pause() / Node::resume()

zawasp:
Add virtual modifier for Node::pause() / Node::resume()

This allows more versatility for derived classes to implement their own resume / pause mechanism. For example, one can implement a **Pause / Resume Scene** where all its children are paused / resumed.
Scene's children nodes would then also override behavior according to specific needs (eg. only some nodes will be paused, other may continue running etc.)


minggo:
[ci rebuild]

PR #7750, opened 2014-08-12
Add FadeBy as an action

hawkwood:


cocos/2d/CCActionInterval.h
```diff
@@ -834,6 +834,40 @@ class CC_DLL FadeTo : public ActionInterval
...
+	
+    /** initializes the action with duration and opacity */
+    bool initWithDuration(float duration, GLubyte opacity);
+	
+protected:
+    GLubyte _byOpacity;
+    GLubyte _toOpacity;
+    GLubyte _fromOpacity;
+    friend class FadeOut;
+    friend class FadeIn;
```
minggo: Why should declare these classes as friend?

minggo:
Could you add a test case?

hawkwood:
They do not need to be, that was a carry over from copying FadeTo.

minggo:
So could you please modify it and add a test case?
Thanks.

hawkwood:
I made the change, but am not sure what/how to add a test case.  Do you want me to put a code snippet here showing how to use it, or something else?

minggo:
You can add a test case in `tests/cpp-tests/Classes/ActionsTest`.

minggo:
@hawkwood 
Any update? Do you have any question about test case?

hawkwood:
Once I started to add a test case, I realized that my code had some significant issues.  I have fixed them and added the test case as requested.

minggo:
Thanks, could you please fix indent? 
Replace tab with 4 spaces.

minggo:
[ci rebuild]

minggo:
[ci rebuild]

minggo:
[ci rebuild]

cocos/2d/CCActionInterval.cpp
```diff
@@ -1953,6 +1953,85 @@ void FadeTo::update(float time)
...
+    if (_target)
+    {
+        float dt = (time - _lastTime);
+        _lastTime = time;
+
+        GLubyte fromOpacity = _target->getOpacity();
+        GLubyte nextOpacity = fromOpacity;
+        if (_additive) {
+            float val = _byOpacity * dt + _lastFrac;
+            float num = roundf( val );
```
minggo, line 2016: It will cause compiling error on windows.

minggo, line 2016: `roundf` is not defined in vs2012.

hawkwood:
Huh, looks like it wasn't added till vs2013.

minggo:
Yep.
So i think we should work around it.

minggo:
Any update?

PR #7751, opened 2014-08-12
add minimize / restore for desktop (win32, linux, mac)

zawasp:
implement glfwSetWindowIconifyCallback for win32 / mac / linux - add support for applicationDidEnterBackground / applicationWillEnterForeground on desktop

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -974,9 +974,9 @@ Developers:
         
     DavidPartouche
         Fixed the bug that video player not showing on iOS if it's not in FullScreen mode
-		
-	zaWasp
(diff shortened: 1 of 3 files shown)


minggo:
Please replace tab with 4 spaces.

minggo:
And i think Mac/Linux can did it like this, these platforms are all using glfw.

zawasp:
Ok, I will try to add mac/Linux as well.

minggo:
Thanks.

zawasp:
Ok, I changed the tabs - sorry about that. Indeed, the solution does work on mac / linux as well.

PR #7724, opened 2014-08-09
Fix a correct onGLFWCharCallback

cpascal:

PR #7762, opened 2014-08-13
add UITextField color & place hold color, fix double delete bug

zilongshanren:
1. add UITextField color & placeholder color support
2. fix twice delete event dispatch on desktop platform
3. add Tests to test NO.1
4. fix CCTextField titleColor & placeholder color issue


zilongshanren:
The No.2 bug was introduced by [this PR](https://github.com/cocos2d/cocos2d-x/pull/7724)

PR #7763, opened 2014-08-13
add camera mask check to visit

super626:
Others override visit should check camera mask by calling isVisitableByVisitingCamera.

use renderer passed instead of get from global for Sprite3D::draw function


cocos/2d/CCNode.h
```diff
@@ -1521,6 +1521,9 @@ class CC_DLL Node : public Ref
     bool doEnumerate(std::string name, std::function<bool (Node *)> callback) const;
     bool doEnumerateRecursive(const Node* node, const std::string &name, std::function<bool (Node *)> callback) const;
     
+    //check whether this camera mask is visible by the current visiting camera
+    bool checkCameraMask() const;
```
minggo: How about rename to `isVisibleByVisitingCamera`?

super626: isVisibleByVisitingCamera seems that it is going to do frustum culling.

tools/cocos2d-console
```diff
@@ -1 +1 @@
-Subproject commit 5db5a4283aa22e362901e2924146ad66c2e1484f
+Subproject commit 4b020e4900e601a2d5845095f2e6970c13c59613
```
minggo: Please not change it.

super626:
Have changed

PR #7782, opened 2014-08-14
mLastTickInNanoSeconds is the renderStart

pangliang:
because:
this.mLastTickInNanoSeconds = renderEnd - renderInterval;
renderInterval = renderEnd - renderStart
so:
this.mLastTickInNanoSeconds = renderEnd - ( renderEnd - renderStart );
so:
this.mLastTickInNanoSeconds = renderStart ;

so:
no need nanoTime() twice

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxRenderer.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxRenderer.java
@@ -100,11 +100,8 @@ public class Cocos2dxRenderer implements GLSurfaceView.Renderer {
 			/*
 			 * Render time MUST be counted in, or the FPS will slower than appointed.
 			*/
-			final long renderStart = System.nanoTime();
+			this.mLastTickInNanoSeconds = System.nanoTime();
 			Cocos2dxRenderer.nativeRender();
-			final long renderEnd = System.nanoTime();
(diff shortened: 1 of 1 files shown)


zilongshanren:
@liang8305 I don't think the twice call of nanoTime() equals.
Suppose we call the nanoTime the first time, it returns 0.01.
If we call it another time , it returns 0.1.
We can't simply do algebra here.

minggo:
Yep.
The value is different in two `nanoTime()`.

pangliang:
yes, The value is different in two nanoTime().
and is different between renderStart and renderEnd
but, the time , you want save,  "this.mLastTickInNanoSeconds" equals renderStart

final long renderStart = System.nanoTime();                 //0.01
Cocos2dxRenderer.nativeRender();
final long renderEnd = System.nanoTime();                  // 0.1
final long renderInterval = renderEnd - renderStart;     // 0.09
this.mLastTickInNanoSeconds = renderEnd - renderInterval;  // =  0.1 - 0.09 = 0.01

minggo:
@liang8305 
Yep, you are right.

zii:
oh, it is funny~

pandamicro:
23333333

zhaijialong:
23333333333333

VisualSJ:
233333333333333333

zhaijialong:
good job,胖梁 23333333333333

PR #7784, opened 2014-08-14
Fixed WP8 compilation errors

Teivaz:
- Fixed WP8 compilation errors (added path to CCGLView.h)
- Added CCCamera.h/cpp to the project


minggo:
Thanks.

Teivaz:
@minggo, there's a problem with custom shaders in WP8. I found no way to easily precompile a shader and then insert it to the cocos rendering engine. Maybe I missed something? If there is nothing like that I can come up with a solution.

minggo:
I hope [this document](http://cocos2d-x.org/wiki/How_to_update_wp8_shader) can give you some help. And there is a limitation of using shaders on WP8, since it doesn't support complicated shaders.

Teivaz:
The thing is our team is working on library for cocos2dx which requires a set of custom shaders. I've precompiled them but there is no way to attach them in cocos template generated project. Including our shaders in cocos's one would be totally wrong as they are used only by our library. I'll send another pull request with a solution for review on monday.

PR #7785, opened 2014-08-14
use fileName for _actionDic key

hijikix:
./cocos/editor-support/cocostudio/CCActionManagerEx.cpp

cant use jsonName for  _actionDic key


liamcindy:
please merge this pr,thanks

PR #7788, opened 2014-08-14
Fixes compiler warnings

NatWeiss:
Fixes compiler warnings due to shadowing a member variable.


minggo:
In cocos2d-x, we will turn off warnings for 3rd libs. Or it will be hard to update Box2D.

PR #7790, opened 2014-08-14
b2World::DrawDebugData shows contacts in red

NatWeiss:
This pull request adds the ability for static bodies to be drawn in red when they have active contacts.

It was so helpful in my latest project I decided to share.

### DIFF
--- a/external/Box2D/Dynamics/b2World.cpp
+++ b/external/Box2D/Dynamics/b2World.cpp
@@ -1146,6 +1146,12 @@ void b2World::DrawDebugData()
 		for (b2Body* b = m_bodyList; b; b = b->GetNext())
 		{
 			const b2Transform& xf = b->GetTransform();
+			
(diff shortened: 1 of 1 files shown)


minggo:
Thanks. Will you send a pull request to Box2D, or it may be hard for cocos2d-x to update Box2D.

NatWeiss:
Ok, I will. The reason I submitted it here is that cocos2d-x includes it's own copy of Box2d (it's not a project reference / link).

minggo:
It is a separate target on Xcode, has a separate project on VS, android, linux and so on.

minggo:
And we decide to use prebuilt version in near future.

NatWeiss:
I see. Well, I emailed Erin Catto with the updates. I couldn't see any other way to contribute to Box2d.

minggo:
Thanks.

PR #7791, opened 2014-08-15
fix replaceAll()

xiangxw:
This function will never return if new_value contains old_value.
For example, old_value = "'", new_value = "''".


minggo:
Thanks.

PR #7797, opened 2014-08-15
add swallow touch and propagation to parent widgets customization

zilongshanren:
1. Deprecated setSelectedState/getSelectedState of UICheckBox to isSelected/setSelected
2. Add swallowing touch options to UIWidgets, now we could call `widget->setSwallowTouches(false)` to change the default behavior.  Swallowing touch won't be that mandatory.
3. Add  touch events propagation options to widget.
   
   Suppose we have a Button on top of a PageView, if we click the Button and drag it with a distance, it will cause the PageView to scroll and cancels Buttons touch event handling.  If we don't want want this behavior, we could call `widget->setPropagateTouchEvents(false)`, then the touch event won't be passed to PageView.
   Another use case is :  If we place a ScrollView on top of another ScrollView, if we scroll the inner ScrollView, we don't want its parent moving. So we must call `scrollView->setPropagateTouchEvents(false)` to achieve this function.
   
    Why do we need this feature? Because some developer ask me to add this options. By default, it will always pass the touch events to its parent widgets.
4. Add tests and fix styling issues.

Note: Now the Widgets swallow touch and propagate touch events to its parent widgets on default. If you set both of these two options to false, when you press and drag a button on a PageView, the Button and PageView both will handing the touch events. 
So do it at your own risk. We might need to add the limitation to our document later.

Any suggestion will be welcome.


cocos/ui/UICheckBox.h
```diff
@@ -166,14 +166,17 @@ class CC_GUI_DLL CheckBox : public Widget
...
     /**
      * Gets selcted state of checkbox.
      *
      * @return selected    true that checkbox is selected, false otherwise.
      */
-    bool getSelectedState()const;
+    CC_DEPRECATED_ATTRIBUTE bool getSelectedState()const{return this->isSelected();}
+    
+    bool isSelected()const;
+    void setSelected(bool selected);
```
minggo, line 179: Please modify comment to.

PR #7798, opened 2014-08-15
move to the latest rapidjson, and change the include path to rapidjson/

rny:
as stated https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/issues/18

### DIFF
--- a/cocos/3d/CCBundle3D.cpp
+++ b/cocos/3d/CCBundle3D.cpp
@@ -30,7 +30,7 @@ THE SOFTWARE.
 #include "renderer/CCGLProgram.h"
 #include "CCBundleReader.h"
 #include "base/CCData.h"
-#include "json/document.h"
+#include <rapidjson/document.h>
 
 #define BUNDLE_TYPE_SCENE               1
 #define BUNDLE_TYPE_NODE                2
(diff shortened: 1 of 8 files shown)


minggo:
Sorry, where is new rapidjson?

rny:
as stated https://code.google.com/p/rapidjson/

This repository has been moved to https://github.com/miloyip/rapidjson

chengstory:
Thanks @rny.
@minggo  I think it may be merged!

minggo:
@rny 
cocos2d-x doesn't have rapidjson libs.

rny:
cocos2d-x does use rapidjson, it is header only, but yours(in the dependency) are an early version and it conflict with a later version (in the case that the project using the latest raidjson like us).  i will try to solve the conflicts tomorrow.

rny:
@chengstory @minggo  I rebuild the patch to the latest version, and tested with the test build.

```
the headers is put into <rapidjson/*>, the same way the latest rapidjson library does.
```

cocos/3d/CCBundle3D.cpp
```diff
@@ -30,7 +30,7 @@ THE SOFTWARE.
 #include "renderer/CCGLProgram.h"
 #include "CCBundleReader.h"
 #include "base/CCData.h"
-#include "json/document.h"
+#include <rapidjson/document.h>
```
minggo, line 33: We don't have rapidjson folder.

rny, line 33: @chengstory @minggo you can pull it from https://github.com/miloyip/rapidjson

and removed the old /json

zilongshanren:
@minggo  This PR could be closed now, since upgrade rapidjson is in our v3.7 milestone.

PR #7804, opened 2014-08-16
 Support for password encrypted zip files

idimiter:
Added optional parameter to FileUtils:: getFileDataFromZip to support password protected zip archives. From the discussion in: http://discuss.cocos2d-x.org/t/cocos2dx-load-from-password-protected-zip/16280

### DIFF
--- a/build/cocos2d_libs.xcodeproj/project.pbxproj
+++ b/build/cocos2d_libs.xcodeproj/project.pbxproj
@@ -1256,6 +1256,8 @@
 		50ABC0671926664800A911A9 /* CCPlatformDefine.h in Headers */ = {isa = PBXBuildFile; fileRef = 50ABBF541926664700A911A9 /* CCPlatformDefine.h */; };
 		50ABC0691926664800A911A9 /* CCStdC.h in Headers */ = {isa = PBXBuildFile; fileRef = 50ABBF551926664700A911A9 /* CCStdC.h */; };
 		A07A4CAF1783777C0073F6A7 /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 1551A342158F2AB200E66CFE /* Foundation.framework */; };
+		A47366DB199F4EBE0014DA06 /* crypt.h in Headers */ = {isa = PBXBuildFile; fileRef = A47366DA199F4EBE0014DA06 /* crypt.h */; };
+		A47366DC199F4EBE0014DA06 /* crypt.h in Headers */ = {isa = PBXBuildFile; fileRef = A47366DA199F4EBE0014DA06 /* crypt.h */; };
 		B2165EEA19921124000BE3E6 /* CCPrimitiveCommand.cpp in Sources */ = {isa = PBXBuildFile; fileRef = B257B45E198A353E00D9A687 /* CCPrimitiveCommand.cpp */; };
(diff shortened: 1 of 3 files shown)


build/cocos2d_libs.xcodeproj/project.pbxproj
```diff
@@ -1256,6 +1256,8 @@
 		50ABC0671926664800A911A9 /* CCPlatformDefine.h in Headers */ = {isa = PBXBuildFile; fileRef = 50ABBF541926664700A911A9 /* CCPlatformDefine.h */; };
 		50ABC0691926664800A911A9 /* CCStdC.h in Headers */ = {isa = PBXBuildFile; fileRef = 50ABBF551926664700A911A9 /* CCStdC.h */; };
 		A07A4CAF1783777C0073F6A7 /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 1551A342158F2AB200E66CFE /* Foundation.framework */; };
+		A47366DB199F4EBE0014DA06 /* crypt.h in Headers */ = {isa = PBXBuildFile; fileRef = A47366DA199F4EBE0014DA06 /* crypt.h */; };
+		A47366DC199F4EBE0014DA06 /* crypt.h in Headers */ = {isa = PBXBuildFile; fileRef = A47366DA199F4EBE0014DA06 /* crypt.h */; };
```
minggo, line 1260: Where is crypt.h?

idimiter, line 1260: I've committed it to the depts repo:
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/20

minggo:
Thanks.
But i think it is better to add a call back function, then user can do decoding by himself. Different developers will use different encryption algorithm.

idimiter:
I'm not sure I understand, where the cb will be called? The crypt.h In minizip currently only supports PKWARE encryption.

minggo:
I mean it is better to modify the function to `getFileDataFromZip(const std::string& zipFilePath, const std::string& filename, ssize_t *size, std::function<void(unsigned char* data)> callback)`. The callback function will do anything it wants including decoding, then it can support any type of encryption.

idimiter:
I'm sorry I may seem a little dumb, but I still cannot understand what the developer will provide in the callback function. Should I close the PR?

minggo:
Sorry, we can not just modify `getFileDataFromZip`. We should add `FileUtils::setCallbackForGetFileDataFromZip(callback function)`, then `getFileDataFromZip` will invoke the call back function if it exists.

idimiter:
I think, that if the developer wants encryption other than password he can always implement it after they take the data from the file. But as for a regular usage I think a password should be fine, since it's supported by common zip tools...

 Im sorry for the "noob" question, but I'm not sure if I should close this pull request or leave it for you to close it?

minggo:
Does `PKWARE encryption` is common used?

idimiter:
AFAIK, PKWARE and AES are the two most commonly used encryptions for zip files. Unfortunately the official minizip package does not support AES. There are forks of minizip which support it, but I'm not sure if we can use them for cocos. I think all zip utilities support PKWARE encryption, so for a basic security PKWARE should be fine.

 Offtopic: There's a repository which supports AES encryption for minizip: https://github.com/nmoinvaz/minizip (I'm not sure if this is a fork, or the official minizip repository. If its not a fork, please excuse my ignorance).

minggo:
I think it is better to implement as #15537 said.

PR #7806, opened 2014-08-18
[CCConsole]Add auto connection and remote control from another device

heliclei:


cocos/base/CCConsole.cpp
```diff
@@ -314,6 +328,54 @@ Console::~Console()
...
 }
 
+#if ((CC_TARGET_PLATFORM != CC_PLATFORM_WIN32) && (CC_TARGET_PLATFORM != CC_PLATFORM_WP8))
+int Console::broadcastLocalIP()
+{
+    
+    int sock;
+    if( (sock = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP)) == -1)
+    {
+        perror("broadcastLocalIP: create socket error!");
```
minggo, line 338: Can not print log on Android, i think you should use `log`.

cocos/base/CCConsole.cpp
```diff
@@ -314,6 +328,54 @@ Console::~Console()
     stop();
 }
 
+#if ((CC_TARGET_PLATFORM != CC_PLATFORM_WIN32) && (CC_TARGET_PLATFORM != CC_PLATFORM_WP8))
```
minggo, line 331: And why can not work on win32?

PR #7823, opened 2014-08-19
Label performance problem

newnon:
if _textColor.a != 255 this function call updateContent every time


WenhaiLin:

PR #7827, opened 2014-08-20
add c++11 random library support

zilongshanren:
1. add C++11 random library.
2. fix wp8 cpp-tests compile error


gin0606:
Thanks your work :+1:

minggo:
Should add `ccRandom.h` into `cocos2d.h`.

PR #7831, opened 2014-08-20
add android as target platform

ntotani:
platform android need original path too

### DIFF
--- a/templates/lua-template-runtime/frameworks/runtime-src/Classes/runtime/Runtime.cpp
+++ b/templates/lua-template-runtime/frameworks/runtime-src/Classes/runtime/Runtime.cpp
@@ -986,7 +986,7 @@ int lua_cocos2dx_runtime_addSearchPath(lua_State* tolua_S)
         if (!FileUtils::getInstance()->isAbsolutePath(arg0))
             arg0 = g_resourcePath+arg0;
         cobj->addSearchPath(arg0);
-#if(CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
+#if(CC_TARGET_PLATFORM == CC_PLATFORM_IOS || CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
     if (!FileUtils::getInstance()->isAbsolutePath(argtmp))
         cobj->addSearchPath(argtmp);
 #endif


chuanweizhang2013:
Why android platform  needs original path too?At debug mode which only reads resources in writable path.

ntotani:
When all resources deployed under `debugruntime` dir, it's ok (like using Cocos Code IDE) . But if resources not deployed, it cant find.

Steps to reproduce
1. create new project using `cocos new` command.
2. deploy to device like `cocos deploy -p android --ap 19`
3. tap PLAY button
got resource not found error

chuanweizhang2013:
You can deploy to device like :  cocos deploy -p android --ap 19 -m release
The debug mode is for  Cocos Code IDE.

ntotani:
hmm... I think it useful that using debug build without Cocos Code IDE.

Then, why only `CC_PLATFORM_IOS` need original path?

chuanweizhang2013:
It's for xcode in debug module without Cocos Code IDE

ntotani:
My question mean that why "only" iOS support debug mode without Cocos Code IDE. Building debug mode without using Cocos Code IDE on both iOS and Android is but natural.

minggo:
@chuanweizhang2013 
Any idea?

chuanweizhang2013:
@minggo Close this request,please.

minggo:
@ntotani 
Can i close this pull request?

ntotani:
@chuanweizhang2013 Can i ask reason?

chuanweizhang2013:
@ntotani   You don't need to add this for android because Cocos Code IDE doesn't include resources folders like "res", "src" for android package while Xcode does package all resources folders into its generated package. So if you choose to compile with resources using the command line tool, you can add these codes yourself

ntotani:
@chuanweizhang2013 Thanks for your explain.

I think it can compatible because writable path searched at first. For example, running `cc.FileUtils:getInstance():addSearchPath("res");` in Lua script, then search order is
1. debugruntime/res
2. res

When use Cocos Code IDE, resource file exists under `debugruntime/res`, so resource can found if not exit under `res`.

chuanweizhang2013:
@ntotani We don't hope have resources in android package,
because if no resources under the "debugruntime" folder,but it still can run,
we can't find error reasons when debug by cocos code ide.

ntotani:
@chuanweizhang2013 you said that

> Cocos Code IDE doesn't include resources folders like "res", "src" for android package

so when no resources under the `debugruntime` folder, it also no resources under original folder, then you can find "file not found" error.

chuanweizhang2013:
yes

minggo:
Can i close it now?

ntotani:
@minggo
I answered that the problem

> we can't find error reasons when debug by cocos code ide

is not caused by this pull request.
Can i ask why this pull request cant merge?

pandamicro:
@ntotani In fact, the iOS platform walk around is only for the compatibility with Xcode, if we don't add this search path, then user will fail to run it with Xcode.
However, this walk around have a bad side effect that we don't want, if you add this search path, Cocos Code IDE will fail to find some potential errors. This is what we don't want, so we haven't activate this search path for Android.
The reason that we make it compatible with Xcode is only because it's massively used, we don't want our user to complain.

That's why we can't merge your pull request, but I'm sure this solution is useful for you and you now fully understand its disadvantage, so I'm sure you are able to avoid problems. I don't think all users want to dig into the implementation like this, so we encourage you to keep this solution to yourself. However we can't merge it. Thanks

ntotani:
@pandamicro Ok, now I understood reason. Thanks for your polite explain.

PR #7832, opened 2014-08-20
Fixed WinRT project

Teivaz:
Inserted newly added files to cocos WIN RT project


minggo:
Thanks.
But it is better not add new pictures if possible. Because it will make repo bigger.

Teivaz:
OK, @minggo, I removed last commit with pictures

minggo:
And please don't let it exist in commit history too.

minggo:
Thanks.

Teivaz:
Sure, it's no longer in the repo in any way.

PR #7873, opened 2014-08-25
Add webview

gin0606:
Hi @andyque
I try to add My WebView to cocos2d-x after lookng this issue. https://github.com/gin0606/cocos2dx-WebView/issues/2

But I have compile error from `libui iOS`.
I try to fix that. But I'm not sure if this is correct.
Would you please check this?


zilongshanren:
Thanks a lot @gin0606 
Sure, I will take over this PR and fix the compile errors.
BTW, in order to merge your code into cocos2d-x repository, I have to change the license and namespace part. Since the WebView only works on Android and IOS platform, so we might need to put it under cocos2d::ui::experimental namespace.

zilongshanren:
@gin0606 
The new PR is opened here: https://github.com/cocos2d/cocos2d-x/pull/7882 , please check it out.

gin0606:
@andyque I will check it! Thank you!

PR #7876, opened 2014-08-25
CMake support for Windows, Mac, Linux, and initial support for IOS and Android

martell:
done: initial IOS cmake import
done: initial Android support

done: fix linux win32
done: update final description
done: win mingw-w64 and mac support


martell:
@minggo feel free to ask me any questions on this PR.

To generate a visual studio project open cmdline with cmake in the system path

cd cocos2d-x/build
mkdir wintest
cd wintest
cmake -G"Visual Studio 12 2013" -DDEBUG_MODE=OFF ../..

This should create all the visual studio projects for you.
here are the supported visual studio arguments
"Visual Studio 10 2010"
"Visual Studio 11 2012"
"Visual Studio 12 2013"

add -DBUILD_STATIC=OFF from above if you want a shared build
As you can see cmake makes it easy to switch back and forth
I think some more work has to go into a shared builds.

This PR wont break anything as static is the default so its good to merge as is

minggo:
Thanks @martell.
As i mentioned in forum, i had modified `MakeLists.txt` to have just one static library for c++ and another for lua-bindings. 

Will you please fix conflict?
Thanks.

minggo:
And i plan to use CMake for all platforms.
Then we will not provide vs projects, xcodes projects in github, just provide CMakeLists.txt.
What's your opinion?

martell:
Hi @minggo,

I would look to merge the having just one library after the next release like you said in the forum
This PR should be good for this release however.
I will fix the conflicts between our patches after the next version is released :)

The build failed error seems more like a jenkins issue and not related to my code.
I can't make much sense of the output

Well yes the ideal scenario would be to remove all project specific code such as vcproj and xcodeproj.
I would only do this for building cocos2d.lib , I would not use cmake in an enduser project because every time cmake is run all customizations to the visual studio or xcode project are lost.

Using CMake we could package prebuilt versions of cocos2d-x for people who dont want to have to build cocos2d-x from source and just want to start on their game creations.

That's the route i'd take with cmake

minggo:
Ok, let's merge this PR after releasing v3.3alpha0.

About CMake usage, i just want to use in github when developing cocos2d-x. As you know now, it is easy to cause conflict if add/remove files. And it is hard for us to modify all platforms project files.

When releasing a version, i think we have to shift the project files as you said.

minggo:
@martell 
v3.3alpha0 was released.
Will you please continue your work?
Thanks.

minggo:
@martell 
Are you doing it?

martell:
Hey @minggo 
Yes I will get a chance to finish this off at the weekend.
You want all the libraries to be just 1 now correct? cococs2d.lib ?

minggo:
Yep, it is only one lib on linux now.
I don't mind if there are more libs or not, i just want to make sure that, developers only have to link libcocos2d, then it will linker other needed libs automatically. Because it is more easier for developers to use.

martell:
Hi @minggo,

I was a bit busy with some other opensource projects but I finally got around to it ;)
I've done a good bit of work on this since the last time.
I noticed that someone partially added mac support but completely broken windows.
This is currently completely untested and I still have to tidy up a few things in this before it becomes merge-able.

I do want to ask however make a few notes before I finish it
1. Do you plan to make prebuilt box2d library like you now have for chipmunk?
   I think this would be best and I've added support for this even though its not done yet.
   I can revert it but I would like to go this route
2. Some of the external libs have a cocod2d namespace.
   I think unzip does I must look again to be sure
   This is a very bad idea because it the system provides these libraries we can't use them :(
3. I liked having multiple CMakeLists.txt because it made it more readable and if you lets say added a new source in 3d its easy to find where you add it to the 3d/CMakeLists.txt but having all the code in one file makes it quite messy and confusing.
   It will all link into one static cocos2d.lib I have not changed that.

Can you give me some feedback on these points before I finish the work

Cheers
Martell

minggo:
@martell 
Thanks.

How about continue discussion in this thread: http://discuss.cocos2d-x.org/t/cmake-for-mac-qtcreator-and-clion/16889/16?
It is Ricardo makes the modification. And there are some discussion about CMake in the thread.

martell:
@minggo Damn I missed the 3.3 beta window.
When is the 3.3 final release?
I've just got Mac working again.
I'm moving to IOS and making linux adjustments atm

minggo:
We will release v3.3 final in middle October. 
I think you can make them work then.
But i think we can remove IDE project in v3.4, but have CMake support all platforms in v3.3.

martell:
Yes I would not remove IDE project in 3.3
I forseen some issues. 
So extra time for bug fixes between 3.3 and 3.4 is a good idea :)

martell:
@minggo @ricardoquesada I have tested this on a range of systems including android and it works.
Could you merge this now. I have signed off on all the commits.
The sooner this is merged the better as we can have a few days of testing and fixing with the other devs before the 3.3 final release.

vovkasm:
I just test it, and get this error on Mac OS with Makefile generator:

```
Linking CXX executable ../../bin/cpp-empty-test/cpp-empty-test
ld: warning: directory not found for option '-L/Users/vovkasm/work/cocos2d-x/external/tinyxml2/prebuilt/mac'
ld: warning: directory not found for option '-L/Users/vovkasm/work/cocos2d-x/external/box2d/prebuilt/mac’
ld: warning: directory not found for option '-L/Users/vovkasm/work/cocos2d-x/external/curl/prebuilt/mac'
ld: library not found for -lbox2d
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/cpp-empty-test/cpp-empty-test] Error 1
make[1]: *** [tests/cpp-empty-test/CMakeFiles/cpp-empty-test.dir/all] Error 2
make: *** [all] Error 2
```

I can principally build box2d, tinyxml2 and curl for mac myself if any instructions exists. I have some questions:
- What versions of libraries used to build? Where this documented?
- Is there any building scripts for them?
- Is it really intent to prebuilt libraries or we can use system provided versions?

minggo:
@vovkasm 
You should run `download-deps.python` to get prebuilt libraries.

vovkasm:
yea, I do that, and see `external/jpeg/prebuilt/mac/libjpeg.a` for example, but there is no `external/curl/prebuilt/mac` directory...

martell:
@vovkasm The cocos2d-x team will fix the prebuilt depends over the coming days.
We just needed to get this merged so we had some time before the 3.3 final release

There is new source files added like CCLight that need to be added to the cmakelists.
This can be considered as unstable over the next week until prebuilts are ready

vovkasm:
@martell, thank you. I will wait ;-) But I have a mac and can build needed deps, simple not sure about versions and any automatizations for this.

ricardoquesada:
This is great news, unfortunately WP8 is not linking any more :-(
Could @martell or @minggo take a look at it ?
It basically says that it can't find the CocosDenshion symbols... I guess it is related to the "static" macro... just a guess.

ricardoquesada:
The other bug is that when on Mac I do

```
$ cmake .. && make -j4 cpp-tests
```

it doesn't compile, and it used to work a few days ago.

martell:
@ricardoquesada The reason cmake doesn't work is because the prebuild libs have to be added.
As for the WP8 issue, I'll fix that now.

minggo:
@martell 
What did you mean the prebuilt libraries?
Is cmake different from Xcode project, because it can work with Xcode.

martell:
@minggo I mean the prebuilt version of box2d, tinyxml2 and minizip for mac.
You guys need to add them to the prebuilt libs repo.
cmake expects the folder layout the same as the third party deps so it should be easy for you to add.

So cmake is the same for mac as it is for linux etc

@ricardoquesada  I've fixed the WP8 issue. I'll do a PR in the morning
I was expecting some teething issues so I'm glad we added it early enough :D

minggo:
@martell
There is problem using prebuilt version of box2d, so all platforms use source code instead. And on linux, we use `apt-get install` to install some needed libraries. But now you change it to using our own prebuilt version.

Using prebuilt version for all platforms seems good, but there is much work to do it. The most important in current stage is fixing bugs then release v3.3. So we will change it to make it work for linux as before.

minggo:
> > There is problem using prebuilt version of box2d

I mean on windows.

martell:
@minggo what I would suggest to do here is build the prebuilt for linux and
mac and insert them into the premade depends package. You can still leave
the source code in the repo for windows.

This way we can move forward with unifying cmake without the box2d issue
holding us up.

PR #7877, opened 2014-08-26
Update CCArmatureAnimation.cpp

function-adlof:
it's for animation "broken body" problem.when ths animation completed , but the update of animation is still running,because of the single frame special case.we should set a condition that the animation has completed couldn't run again.

### DIFF
--- a/cocos/editor-support/cocostudio/CCArmatureAnimation.cpp
+++ b/cocos/editor-support/cocostudio/CCArmatureAnimation.cpp
@@ -337,11 +337,14 @@ void ArmatureAnimation::update(float dt)
 {
     ProcessBase::update(dt);
     
-    for (const auto &tween : _tweenList)
+    if (!_isComplete && !_isPause)
     {
-        tween->update(dt);
+        for (const auto &tween : _tweenList)
+        {
+            tween->update(dt);
+        }
     }
-
+    
     if(_frameEventQueue.size() > 0 || _movementEventQueue.size() > 0)
     {
         _armature->retain();


function-adlof:
Armature "broken body" problem

cocos/editor-support/cocostudio/CCArmatureAnimation.cpp
```diff
@@ -337,11 +337,14 @@ void ArmatureAnimation::update(float dt)
 {
     ProcessBase::update(dt);
     
-    for (const auto &tween : _tweenList)
+    if (!_isComplete && !_isPause)
```
2youyou2, line 340: Thanks for your pr!
But this condition may cause Tween can not update last frame and the frame event in the last frame can not be emitted.

minggo, line 340: @function-adlof 
What's your opinion?

function-adlof, line 340: my cocos2dx version is 2.2.4.
because there was a problem of Armature , it's Corpse disintegration on that Armature which is single frame.
and when I debug it , I found that the frame has been update completely ,but the update function was still running.
it makes the armature flying anywhere which should stay still.
so I add this Control.And It goes well.‍

发送时间: 2014年8月29日(星期五) 下午4:11
抄送: "当代美髯公。"349653683@qq.com; 
主题: Re: [cocos2d-x] Update CCArmatureAnimation.cpp (#7877)

In cocos/editor-support/cocostudio/CCArmatureAnimation.cpp:

minggo, line 340: So, you should send a pull request to v2, not v3?

2youyou2, line 340: I remember i had already fixed this problem before, could you test it in latest version.
If still has problem, could you send your test to me by email?

kaile.yin@chukong-inc.com

Thanks!

PR #7880, opened 2014-08-26
Optimize compiling

minggo:
- add cocos2d.h to pre-complied header on iOS, will added to other platforms in future
- remove unneeded include header files


minggo:
[ci rebuild]

PR #7883, opened 2014-08-26
Add the feature that CCSSceneReader can load name properties as node names.

giginet:
When name property of objects on CocosStudio Scene are set.
They will be loaded and set as Node::name property.

Thanks to this commit. We are enable to access to each nodes on scene via its names.

``` cpp
auto scene = cocos2d::cocostudio::SceneReader::getInstance()->createNodeWithSceneFile("Scene.json");
auto object = scene->getChildByName<cocos2d::Sprite *>("objectName");
```


minggo:
Thanks.

PR #7896, opened 2014-08-27
Adds `std::nothrow` to the `new` statements

ricardoquesada:
Since cocos2d-x doesn't support exceptions, it is important to add
`std::nothrow` to all the `new` statements


minggo:
[ci rebuild]

PR #7910, opened 2014-08-29
Modify EditBox for win32

summerinsects:
1. Do render when the system DialogBox popup
2. Star('*') wll be shown in win32 edit control if the EditBox's input flag is
   password
3. Function editBoxTextChanged will be called when the text in win32
   edit control is changed
4. The temporary variable (here is std::string) may destruct
   when use expression like AnsiToUtf8(...).c_str() or Utf8ToAnsi(...).c_str()


extensions/GUI/CCEditBox/CCEditBoxImplWin.cpp
```diff
@@ -149,30 +151,39 @@ void EditBoxImplWin::setText(const char* pText)
 {
     if (pText != nullptr)
     {
+        if (_text.compare(pText))
```
zilongshanren: Do you mean _text equals to pText, then we should return?
It think here should be (0 == _text.compare(pText)).

summerinsects: I'm so sorry
Yes. It should return if _text equals to pText
Btw, bool EditBoxImplWin::isEditing() should return _isEditing
Thx!

summerinsects: I'm sorry than in function char\* CWin32InputBox::AnsiToUtf8(const char\* strAnsi) the local variable ret has not been initialized, and NULL should be assigned

PR #7920, opened 2014-09-01
win32 can not stop looping sound effect

rny:


minggo:
@Dhilan007
Could you please review it?
Thanks.

minggo:
I do it like this:
- run cpp-tests
- enter `Audio - CocosDenshion`
- touch `play in loop`
- after playing the effect several times, touch `stop`
- effect was stopped

So it can stop loop sound effect.
How to reproduce the issue you said?

minggo:
@rny 
Any thought?

rny:
we have two audio effect playing simultaneously with different files, 

```
1. play one time effect A
2. play loop effect B
3. A finished playing
4. call stop for B
```

B stopped this time, but playing again from the beginning soon.

PR #7921, opened 2014-09-01
V3 triangles command

dabingnn:
This pull request add Triangles command to support rendering of triangles.
quad command now is a subclass of trianglesCommand


dabingnn:
[ci rebuild]

cocos/renderer/CCRenderer.h
```diff
@@ -75,7 +76,9 @@ Whenever possible prefer to use `QuadCommand` objects since the renderer will au
 class CC_DLL Renderer
 {
 public:
-    static const int VBO_SIZE = 65536 / 6;
+    static const int VBO_SIZE = 8192;
```
ricardoquesada, line 79: Why `8192` and not `65536 / 6` ?

dabingnn, line 79: Now VBO_SIZE is the number of vertices, which is the number of quad before. And the size of Vertex array is 65536, which is very big, we need to allocate a big array, but in most of situations, this is not needed.
Now the VBO_SIZE is 8192, which means that we can batch up to 2048 quads, which means 2048 sprites on the screen, I think it is enough.

We can change this back to 65536 if it is needed, what is your suggestions about this?

cocos/renderer/CCQuadCommand.cpp
```diff
@@ -28,15 +28,12 @@
...
 
 NS_CC_BEGIN
 
-
 QuadCommand::QuadCommand()
-:_materialID(0)
-,_textureID(0)
-,_glProgramState(nullptr)
-,_blendType(BlendFunc::DISABLE)
+:TrianglesCommand()
```
minggo: Why need to invoke `TrianglesCommand()`?
It is a subclass of TrianglesCommand, it will invoke TrianglesCommand's constructor automatically.

cocos/renderer/CCRenderer.h
dabingnn, line 79: It seems there are some crashes if user send a large amount of quads to render(bigger than VBO_SIZE).
Given that we are using unsigned short as element index.
I will roll VBO_SIZE back to 65536.

PR #7925, opened 2014-09-02
Animation3D

super626:
Add transition between Animate3D.


super626:
[ci rebuild]

minggo:
About `Sequence3D`, please discuss it in [this thread](http://discuss.cocos2d-x.org/t/does-3d-node-need-to-support-actions/16691).

super626:
I have removed Sequence3D according to Riq's suggestion.

minggo:
[ci rebuild]

PR #7927, opened 2014-09-02
fix transition scene stencil buffer

zilongshanren:
When a scene transition contains ScrollView or ListView, the effect will be wrong. We should enable the stencil buffer when scene transition happens.

### DIFF
--- a/cocos/2d/CCTransition.cpp
+++ b/cocos/2d/CCTransition.cpp
@@ -1278,7 +1278,7 @@ void TransitionCrossFade::onEnter()
     LayerColor* layer = LayerColor::create(color);
 
     // create the first render texture for inScene
-    RenderTexture* inTexture = RenderTexture::create((int)size.width, (int)size.height);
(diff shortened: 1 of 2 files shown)


minggo:
@dabingnn 
Could you please review it too.
Is there any problem to change it like this?

dabingnn:
The effect of this PR seems to use extra GPU memory to fix this problem.
It is safe to merge it.
Just for optimisation and memory saving, if we can find a way to test whether depth and stencil buffer is really needed, it could be much better.

owen200008:
I use the 3.9 and 3.10 version
when i transition the scene have have the clip contain, the clip panel background will be green
how can i solve the problem

owen200008:
@zilongshanren 
the code in the 3.9 and 3.10 is contain the #7927, how can i solve the problem

owen200008:
@zilongshanren  I write the class NodeGridDeepCopy overwrite the settarget and visit to create the reandertexture to solve the problem

owen200008:
i find the bug. if you use the this code your can not use the fastTMXLayer, because when the visit

owen200008:
const Camera\* pVisitCamera = Camera::getVisitingCamera(); is nil

PR #7934, opened 2014-09-03
fix: win32 cpp-tests sometimes cannot display font correctly

summerinsects:
Pass FILE_SHARE_READ to the 3rd parameter of win32 function CreateFileW. Otherwise, it
may fail to get date from the font file.

### DIFF
--- a/cocos/platform/win32/CCFileUtilsWin32.cpp
+++ b/cocos/platform/win32/CCFileUtilsWin32.cpp
@@ -147,7 +147,7 @@ static Data getData(const std::string& filename, bool forString)
         WCHAR wszBuf[CC_MAX_PATH] = {0};
         MultiByteToWideChar(CP_UTF8, 0, fullPath.c_str(), -1, wszBuf, sizeof(wszBuf)/sizeof(wszBuf[0]));
 
-        HANDLE fileHandle = ::CreateFileW(wszBuf, GENERIC_READ, 0, NULL, OPEN_EXISTING, NULL, nullptr);
+        HANDLE fileHandle = ::CreateFileW(wszBuf, GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, NULL, nullptr);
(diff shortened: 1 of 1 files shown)


minggo:
@wohaaitinciu 
How to reproduce the issue?

summerinsects:
Run cpp-tests, then close it. And run again, close again...Repeat several times.
qq 20140903145852

minggo:
I run cpp-tests more than 10 times, can not reproduce it.

cmzx3444:
I found this issue too,the font can not be read. reboot the cumpter will be ok.

minggo:
Can this PR fix your issue?

summerinsects:
Yes.It seems to be correctly

minggo:
@cmzx3444 
Can this PR fix your issue?

cmzx3444:
Sorry in my company's computer I can't  reproduce the issue, but at home, i occur frequently，I only click some testcase and it crash by can not load font.
I will try tonight

minggo:
@cmzx3444 
Thanks.

summerinsects:
I found the reason why the font cannot be read is some other process has opend the font already,
thus passing 0 to the 3rd parameter(share mode) of CreateFileW is going to be failed

minggo:
But what process will open it?

cmzx3444:
I  reproduce the issue in my home computer, and this pr works.
The error is Get data from file(fonts/arial.ttf) failed, error code is 32
as @wohaaitinciu said  the font  has opened in some other process.
refer to http://msdn.microsoft.com/en-us/library/bb202810.aspx
When i try to delete the fonts/arial.ttf in windows,I occur error
qq 20140903233902

summerinsects:
Does cocos create any background process?

cmzx3444:
I think the issue is case by Label::createWithSystemFont,when i replace it with Label::createWithTTF,it do not happen again

cmzx3444:
But I can't understand the implementation of Label , I'm sorry.

minggo:
@wohaaitinciu 
No, there is not background process.

PR #7939, opened 2014-09-03
remove Node::_setLocalZOrder()

minggo:
I don't think `Node::_setLocalZOrder()` is needed.

@ricardoquesada 
Do you think there is any problem?


minggo:
[ci rebuild]

PR #7947, opened 2014-09-04
add simple implement of WebView for win32

summerinsects:
Is it being implemented by official developing team? If yes, ignore it

This implement contains some simple functions such as loadUrl, loadFile,
evaluateJS
Other more functions have not been implemented yet
My environment is win8.1 vs2013, and it works well


minggo:
Thanks.
No, we are not doing right now.

Will you continue to implement other functions?

summerinsects:
@minggo I will try my best.

minggo:
Thanks.

minggo:
You should exclude `UIWebViewImpl_win32.cpp` to be compiled on linux.
You can do it in two ways:
- modify `cocos/CMakeLists.txt` to exclude `UIWebViewImpl_win32.cpp`
- make `UIWebViewImpl_win32.cpp` can be included in all platforms, just add macros at begin and end like this

``` c++
// UIWebViewImpl_win32.cpp
#if CC_TARGET_PLATFROM == CC_PLATFORM_WINDOWS
...
#endif
```

summerinsects:
OK

cocos/ui/UIWebViewImpl_win32.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+    class Win32WebView
+    {
+    public:
+        Win32WebView()
+            : m_pWebBrowser2(NULL)
+            , m_bCanGoBack(true)
+            , m_bCanGoForward(true)
+            , m_dwCookie(0)
+            , m_ulRef(0)
```
minggo: Please use _xxx as member variables to obey cocos2d-x programming guide.

cocos/ui/UIWebViewImpl_win32.cpp
```diff
@@ -0,0 +1,417 @@
...
+        {
+            lazyInit();
+        }
+
+        bool createWebView()
+        {
+            bool bRet = false;
```
minggo: Please don't use Hungary style.
bRet -> ret

summerinsects:
These codes is form an old win32 project.
OK, I'll modify them later

minggo:
Thanks.

summerinsects:
@minggo 
Cound I pass nullptr to C-style win32 api functions, such as GetModuleHandle,
or initialize some handle with nullptr, such as HWND,
or compare handle with nullptr to check whether some function call has no error?

minggo:
I am not sure. I haven't noticed that it is win32 API, so use NULL is ok.

summerinsects:
Then, I keep using NULL when calling win32 APIs or doing some operation with handle?

minggo:
Yep, using NULL for win32 API.

summerinsects:
I have added more implementation.
Since I have no IOS environment, I built an apk, and studied the 3 callbacks.
However, due to platform differences, there should be many differences when callback is called

minggo:
> Since I have no IOS environment, I built an apk, and studied the 3 callbacks.
> However, due to platform differences, there should be many differences when callback is called

You are implementing for windows, is there any problem related to iOS?

summerinsects:
I want to make the platform differences as small as possible

minggo:
Got it.
Is this PR ready to be merged?

summerinsects:
Yes

summerinsects:
@minggo 
It seems that my version is too old. Should I recheckout and PR again?

minggo:
You are appreciated if you recheckout and PR again.
And please add files into VS project and make the test case work.

PR #7950, opened 2014-09-04
3D BillBoard

super626:
Add BillBoard and BillBoardTest.
BillBoard are mostly transparency so add a transparency queue in the render.

BillBoard usage:
auto billboard = BillBoard::create("Images/Icon.png");
addChild(billboard);
billboard->setCameraMask(2);// which camera do you want this billboard be seen.


cocos/3d/CCBillBoard.h
```diff
@@ -0,0 +1,121 @@
...
+ * Inherit from Sprite, achieve BillBoard.
+ */
+class CC_3D_DLL BillBoard : public Sprite
+{
+public:
+
+    enum class Mode
+    {
+        View_Point_Oriented,
+        View_Plane_Oriented
```
minggo: Please use capital characters.

minggo: And could you please write some comment of different modes?

cocos/3d/CCBillBoard.cpp
```diff
@@ -0,0 +1,165 @@
...
+        }
+        _modeDirty = false;
+
+        if (camDir.length() < MATH_TOLERANCE)
+        {
+            camDir.set(camWorldMat.m[8], camWorldMat.m[9], camWorldMat.m[10]);
+        }
+        camDir.normalize();
+
+        static Vec3 upAxis(0.0f, 1.0f, 0.0f);
```
minggo, line 127: Why use static variable here?
Could it be a static class variables?

cocos/base/CCDirector.cpp
```diff
@@ -285,6 +285,9 @@ void Director::drawScene()
     
     if (_runningScene)
     {
+        //clear draw stats
+        _renderer->clearDrawStats();
+
```
minggo, line 289: Why need to invoke `_renderer->clearDrawStats()` here?

cocos/renderer/CCRenderer.h
```diff
@@ -92,6 +108,9 @@ class CC_DLL Renderer
 
     /** Adds a `RenderComamnd` into the renderer specifying a particular render queue ID */
     void addCommand(RenderCommand* command, int renderQueue);
+    
+    /** add transprent command */
+    void addTransparentCommand(RenderCommand* command);
```
minggo: addTransparentCommand -> addCommandToTransparentQueue?

cocos/renderer/CCRenderer.cpp
```diff
@@ -337,6 +368,58 @@ void Renderer::visitRenderQueue(const RenderQueue& queue)
...
+        if(RenderCommand::Type::QUAD_COMMAND == commandType || RenderCommand::Type::TRIANGLES_COMMAND == commandType)
+        {
+            auto cmd = static_cast<TrianglesCommand*>(command);
+            _batchedCommands.push_back(cmd);
+            fillVerticesAndIndices(cmd);
+            drawBatchedQuads();
+        }
+        else if(RenderCommand::Type::GROUP_COMMAND == commandType)
+        {
+            int renderQueueID = ((GroupCommand*) command)->getRenderQueueID();
```
minggo: Use static_cast<> instead?

cocos/renderer/CCRenderer.cpp
```diff
@@ -358,6 +438,16 @@ void Renderer::render()
...
         flush();
+        
+        //Process render commands
+        //draw transparent objects here, do not batch for transparent objects
+        if (0 < _transparentRenderGroups.size())
+        {
+            _transparentRenderGroups.sort();
+            glEnable(GL_DEPTH_TEST);
+            visitTransparentRenderQueue(_transparentRenderGroups);
+            glDisable(GL_DEPTH_TEST);
```
minggo, line 449: Should cache previous stat then re-install it?
For example, depth test may already enabled before visiting transparent render queue.

cocos/base/CCDirector.cpp
super626, line 289: We may call renderer->render() multiple times.
We should reset the stats to zero before call render(); and make the addition in the render routine.

What we did before is that reset stats to zero in the render(), and then increase it there. It will get wrong stats when calling render multiple times.

cocos/3d/CCBillBoard.h
super626: Ok.

cocos/3d/CCBillBoard.cpp
super626, line 127: static class variable can work. 
But I think this variable has nothing to do with the class, It is only used in this function, and I do not want to construct it each call. Maybe you can use another algorithm without this variable in future.
We can make it as class variable if static variable here is not allowed.

cocos/renderer/CCRenderer.h
super626: OK

cocos/renderer/CCRenderer.cpp
super626: Ok

cocos/renderer/CCRenderer.cpp
super626, line 449: It is better that GLStateCache has the method enable depth then we can cache it there.

most of the time we disable the depth test during draw 2d objects. To make sure it works we should re-install it, because previous state may be invalid.

minggo:
[ci rebuild]

super626:
[ci rebuild]

super626:
[ci rebuild]

minggo:
[ci rebuild]

PR #7952, opened 2014-09-04
Added math SSE2 implementation on x86 platforms

sunzhuoshi:
NOTE: double is used as scalar, it may conflict with float assumption of existing code.


minggo:
Why need to use double? 
The modification make codes complex.

PR #7956, opened 2014-09-05
fix ScrollView bug

denpen:
I Fixed a bug that scroll view hidden picks up the touch events.


minggo:
@andyque
Could you please review it.

zilongshanren:
@minggo  Sure, I will review it later.

PR #7957, opened 2014-09-05
bugfix for UI animation playing crash if GUI JSON file is loaded again.

hongliang-goudou:
_actionDic is a std::unordered_map, the insert() function will not work if the key is already existed. If a UI JSON file is loaded by GUIReader more than once, the animation information should update in the internal _actionDic map, as the rootWidget pointer is different.

...Detail see: http://blog.segmentfault.com/hongliang/1190000000659937

### DIFF
--- a/cocos/3d/CCBillBoard.cpp
+++ b/cocos/3d/CCBillBoard.cpp
@@ -148,7 +148,8 @@ void BillBoard::draw(Renderer *renderer, const Mat4 &transform, uint32_t flags)
     //FIXME: frustum culling here
     {
         _quadCommand.init(_zDepthInView, _texture->getName(), getGLProgramState(), _blendFunc, &_quad, 1, _billboardTransform);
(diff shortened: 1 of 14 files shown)


zilongshanren:
@hongliang-goudou 
Thanks, I will test your code on v3.3.

minggo:
@hongliang-goudou 
I think this PR include many modification that is nothing about this PR.

hongliang-goudou:
@minggo sorry about that, not very good at git and github. :(

minggo:
No matter, would you please send a new PR.
And you can do it like this(suppose you have v3 too):

```
git checkout v3
git pull --rebase https://github.com/cocos2d/cocos2d-x.git
git checkout -b new_branch
```

Than continue your work in new_branch.

hongliang-goudou:
I'll send a new PR. This one close.

minggo:
Thanks @hongliang-goudou .

PR #7959, opened 2014-09-05
Adds missing xcshared data for missing projects

ricardoquesada:
Adds missing xcshared data for missing projects


minggo:
Why need it?

PR #7963, opened 2014-09-08
Added Math SSE implementation on x86

sunzhuoshi:


minggo:
[ci rebuild]

cocos/math/Mat4.cpp
```diff
@@ -414,8 +414,11 @@ void Mat4::add(float scalar)
 void Mat4::add(float scalar, Mat4* dst)
 {
     GP_ASSERT(dst);
-
+#ifdef __SSE__
```
minggo, line 417: It is dangerous only use `__SSE__` to control the code.
Because this macro may be defined on windows too.
We will fix it in another pr.

sunzhuoshi, line 417: SSE is supported on all x86 platforms, including windows.
It SHOULD be OK to use SSE implementation when <code>**SSE**</code> defined

minggo, line 417: But you include `<xmmintrin.h>` which is not supported on windows.

sunzhuoshi, line 417: Got it. Let me check.

minggo, line 417: At least we met problem in lua-bindings.
As you know, we will search all header files included recursively in lua-binding.
We are fixing it.

sunzhuoshi, line 417: All right, sorry for inadequate tests.

sunzhuoshi, line 417: about <code>&lt;xmmintrin.h></code>  not supported on windows.
what is your visual studio version and windows version?
I've tested in VS2008, VS2010 on windows 7, and VS2012 on windows 8. It works.

minggo, line 417: We use clang to generate lua-binding. And it will use clang's `<xmmintrin.h>` which is a little different from vs version. Then it will cause problem in lua-binding.

PR #8001, opened 2014-09-11
Update CCApplicationProtocol.h, CCDeprecated.h

halx99:
solve conflict, OS_WINDOWS already defined on shlwapi.h


minggo:
It will break compatibility.

PR #8002, opened 2014-09-11
CCDeprecated.h CCApplicationProtocol.h

halx99:
solve conflict, OS_WINDOWS defined on shlwapi.h

### DIFF
--- a/cocos/deprecated/CCDeprecated.h
+++ b/cocos/deprecated/CCDeprecated.h
@@ -956,7 +956,7 @@ CC_DEPRECATED_ATTRIBUTE const LanguageType kLanguagePolish      = LanguageType::
 CC_DEPRECATED_ATTRIBUTE typedef LanguageType ccLanguageType;
 
 
-CC_DEPRECATED_ATTRIBUTE const Application::Platform kTargetWindows      = Application::Platform::OS_WINDOWS;
+CC_DEPRECATED_ATTRIBUTE const Application::Platform kTargetWindows      = Application::Platform::OS_WINDOWS_COCOS_SPEC;
 CC_DEPRECATED_ATTRIBUTE const Application::Platform kTargetLinux        = Application::Platform::OS_LINUX;
 CC_DEPRECATED_ATTRIBUTE const Application::Platform kTargetMacOS        = Application::Platform::OS_MAC;
 CC_DEPRECATED_ATTRIBUTE const Application::Platform kTargetAndroid      = Application::Platform::OS_ANDROID;


minggo:
Two questions:
- what's shlwapi.h
- there is not Application:: OS_WINDOWS_COCOS_SPEC

halx99:
windows system header

发自我的 Windows Phone

---

发件人: minggomailto:notifications@github.com
发送时间: ‎2014/‎9/‎11 14:51
收件人: cocos2d/cocos2d-xmailto:cocos2d-x@noreply.github.com
抄送: xseekerjmailto:xseekerj@live.com
主题: Re: [cocos2d-x] CCDeprecated.h CCApplicationProtocol.h (#8002)

Two questions:
- what's shlwapi.h
- there is not Application:: OS_WINDOWS_COCOS_SPEC

---

halx99:
"Therefore, 'OS_' prefix is added to avoid conflicts with the definitions of system macros"

But OS_WINDOWS already define as macro at shlwapi.h.

minggo:
How did you meet conflict?

halx99:
Well, when I develop tools based on cocos2d-x， I need include Windows SDK header shlwapi.h, the  follow code is copy from it:

// Returns TRUE/FALSE depending on question
#define OS_WINDOWS                  0           // Windows 9x vs. NT
#define OS_NT                       1           // Windows 9x vs. NT
#define OS_WIN95ORGREATER           2           // Win95 or greater
#define OS_NT4ORGREATER             3           // NT4 or greater
#define OS_WIN98ORGREATER           5           // Win98 or greater
#define OS_WIN98_GOLD               6           // Win98 Gold (Version 4.10 build 1998)
#define OS_WIN2000ORGREATER         7           // Some derivative of Win2000

minggo:
Can you resolve the issue if include cocos2d.h before shlwapi.h?

halx99:
Cannot, there are still have compile error

minggo:
Ok, so i think we should modify it like this
- OS_WINDOWS -> CC_OS_WINDOWS
- modify all platforms
- deprecated old ones, we can not just modify them to break compatibility

halx99:
Ok,Great

发自我的 Windows Phone

---

发件人: minggomailto:notifications@github.com
发送时间: ‎2014/‎9/‎11 21:00
收件人: cocos2d/cocos2d-xmailto:cocos2d-x@noreply.github.com
抄送: xseekerjmailto:xseekerj@live.com
主题: Re: [cocos2d-x] CCDeprecated.h CCApplicationProtocol.h (#8002)

Ok, so i think we should modify it like this
- OS_WINDOWS -> CC_OS_WINDOWS
- modify all platforms
- deprecated old ones, we can not just modify them to break compatibility

---

PR #8025, opened 2014-09-12
fix Editbox lua binding

zilongshanren:
1. add ui::EditBox Lua binding, deprecated cc.EditBox, please use ccui.EditBox instead
2. remove deprecated warnings
3. There is one caution about Lua module register order : 

```
 int lua_module_register(lua_State* L)
 {
      register_ui_moudle(L);
     register_extension_module(L);
    // other Lua modules
}
```

The `register_ui_module` must comes before the `resigter_extension_module`.


zilongshanren:
[ci rebuild]

PR #8029, opened 2014-09-13
HttpClient SSL verification

kobakei:
This patch adds a method HttpClient::enableSSLVerification(const std::string&) to turn on SSL verification, which is useful for security purpose.

### DIFF
--- a/cocos/network/HttpClient.h
+++ b/cocos/network/HttpClient.h
@@ -56,8 +56,12 @@ public:
     /** Enable cookie support. **/
     void enableCookies(const char* cookieFile);
     
-    /** Enable SSL verification. */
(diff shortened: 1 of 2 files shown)


cocos/network/HttpClient.h
```diff
@@ -55,6 +55,9 @@ class CC_DLL HttpClient
 
     /** Enable cookie support. **/
     void enableCookies(const char* cookieFile);
+    
+    /** Enable SSL verification. */
+    void enableSSLVerification(const std::string& caFile);
```
minggo: Is it better to rename it to setSSLVerification?
Generally, we pass an boolean value to enableXXX function.

kobakei:
Thank you for your review.
I have renamed the method and add more comment.

PR #8032, opened 2014-09-14
fix memory leak when use menu_selector.

joewan:
1.MenuItemFont::create("reset", this, menu_selector(UpdateLayer::reset ));
2.MenuItemFont::create("reset", CC_CALLBACK_1(UpdateLayer::reset, this));

The first old style will cause memory leak:
the menuitem(child) owns the parent("this")'s reference,
the parent(this) owns the menuitem's reference.
At last they both can not release.

we can not release the target's reference in the menuitem's deconstructor.
or there is no need to retain the target when initialize the menuitem with menu_selector.


joewan:
it will cause serious leak when the target has many children.

PR #8046, opened 2014-09-15
FileUtils isPopupNotify not work

svzdev:
fixed: ParticleSystem "set not pop-up message box when load image failed" doesn't work


minggo:
Will you please fix conflict?
Thanks.

PR #8050, opened 2014-09-15
fix: ParticleSystem "set not pop-up message box when load image failed" ...

svzdev:
...doesn't work

isPopupNotify() doesn't work


svzdev:
ParticleSystem setPopupNotify doesn't work

PR #8062, opened 2014-09-15
optimize decompress jpg data

codetypess:
it may be unnecessary copy jpg bytes one by one when decompress jpg data

### DIFF
--- a/cocos/platform/CCImage.cpp
+++ b/cocos/platform/CCImage.cpp
@@ -885,8 +885,8 @@ bool Image::initWithJpgData(const unsigned char * data, ssize_t dataLen)
         _width  = cinfo.output_width;
         _height = cinfo.output_height;
         _hasPremultipliedAlpha = false;
-        row_pointer[0] = static_cast<unsigned char*>(malloc(cinfo.output_width*cinfo.output_components * sizeof(unsigned char)));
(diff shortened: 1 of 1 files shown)


cocos/platform/CCImage.cpp
```diff
@@ -896,12 +896,12 @@ bool Image::initWithJpgData(const unsigned char * data, ssize_t dataLen)
...
-            for (i=0; i<cinfo.output_width*cinfo.output_components; i++)
-            {
-                _data[location++] = row_pointer[0][i];
-            }
+            row_pointer[0] = _data + location;
+            location += cinfo.output_width*cinfo.output_components;
+            jpeg_read_scanlines(&cinfo, row_pointer, 1);
         }
+        
+        row_pointer[0] = nullptr;
```
minggo: Why need to add this line?
To prevent been destroyed in 

``` c++
if (row_pointer[0] != nullptr)
    {
        free(row_pointer[0]);
    };
```

?

If so, i think we can remove the codes.

And please fix warnings. After this modification, `i` is not used.

Thanks.

codetypess: yes, add that line to prevent been destroyed in follow codes

PR #8092, opened 2014-09-17
Added font fallback for WP8 & WinRT

guykogus:
Added the ability to add fallback fonts so that characters from different languages can be displayed on WP8/WinRT.

Implemented loading system fonts for WP8/WinRT.

Windows Phone 8 has a specific list of officially supported system fonts, and an `enum` has been created for each one so that they can be loaded.

### DIFF
--- a/cocos2dx/platform/winrt/CCFreeTypeFont.cpp
+++ b/cocos2dx/platform/winrt/CCFreeTypeFont.cpp
@@ -40,7 +40,7 @@ using namespace std;
 NS_CC_BEGIN
 
 static map<std::string, FontBufferInfo> s_fontsNames;
-static vector<FT_Face> s_faces;
+static vector<pair<FT_Face, unsigned char*>> s_faces;
 static FT_Library s_FreeTypeLibrary = nullptr;
 
 CCFreeTypeFont::CCFreeTypeFont() 
@@ -138,7 +138,7 @@ bool CCFreeTypeFont::initWithString(
 
     for (auto i = s_faces.begin(); !error && i != s_faces.end(); ++i)
     {
-        error = FT_Set_Char_Size(*i, nSize << 6, nSize << 6, 72, 72);
+        error = FT_Set_Char_Size(i->first, nSize << 6, nSize << 6, 72, 72);
     }
 
 	if(!error && !m_face)
@@ -340,7 +340,8 @@ void CCFreeTypeFont::setFallbackFonts(const std::vector<const char*>& fonts)
     // Clear out the system fonts first
     for (auto i = s_faces.begin(); i != s_faces.end(); ++i)
     {
-        FT_Done_Face(*i);
+        FT_Done_Face(i->first);
+        delete [] i->second;
     }
     s_faces.clear();
     s_faces.reserve(fonts.size());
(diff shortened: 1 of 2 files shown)


guykogus:
I was unable to test these using samples/Cpp/TestCpp as it fails to build for Visual Studio, but it has been used for Ready Steady Play and works very well.

minggo:
@stammen
Will you review it?

stammen:
I will review this tomorrow.

stammen:
Why do we need a fallback font? Why not just try and load the system font first on WP8 if it is available and then use the System font? Also, font caching was re-enabled which will speed up performance but use a lot of memory if several backup fonts are used. 

We need to refactor the winrt/wp8 CCFreeTypeFont class to remove FreeType headers from the CCFreeTypeFont.h. It will be difficult for people to have to deal with include paths and other errors introduced by the FreeType headers. This is work I am going to have to do for the Universal App version since font access is greatly improved.It would be great if Guy is willing to work with me on this.

guykogus:
I created this system to suit a game that we recently released. The game is localised into 12 different languages (including Japanese, Korean, Simplified Chinese and Traditional Chinese) and we use a custom font that only contains latin and cyrillic characters. On iOS and Android, when the game needs to draw Asian characters the OSes use backup fonts automatically. WP8 doesn't do this, it just prints squares where there are missing glyphs, so I wanted to create a fallback system and share it with others.

Font caching was turned on because we only use one font, so why continually reload it? We can turn it back off, but on average how many different fonts do games use?

I'm very happy to continue working with you on this. Let me know what you need and I'll help you out.

stammen:
Thank you for sharing your experience. I will review the code again using this context.

guykogus:
Pleasure. If I know of a bug in my code (i.e. leaking the font data when setting a new vector of fallback fonts) should I fix it myself or do you need to do it?

stammen:
Go ahead and fix any bugs you find and push to your branch. It will update the pull request.

guykogus:
@minggo Why is Jenkins failing to build? It seems to be having some trouble with Box2D, even though I didn't touch it.

minggo:
@guykogus It is not your fault. We are fixing it.

zilongshanren:
[ci rebuild]

zilongshanren:
[ci rebuild]

zilongshanren:
[ci rebuild]

zilongshanren:
[ci rebuild]

guykogus:
[ci rebuild]

guykogus:
[ci rebuild]

guykogus:
[ci rebuild]

guykogus:
[ci rebuild]

guykogus:
[ci rebuild]

guykogus:
Maybe Jenkins is struggling because this is linking back to v2, not v3?

minggo:
Oh, you send pull request to v2, not v3.

guykogus:
Yeah, I haven't really used v3 yet and the `CCFreeTypeFont` class was changed for it, so I'd need to refactor my code.

minggo:
Yep, please work on v3.
v2 will only accept critical bug fix.

guykogus:
OK, I'll delete this branch and do a new one for v3.

PR #8095, opened 2014-09-17
Update CCDevice-ios.mm

stevekanter:
On iOS, Device::setKeepScreenOn(bool value) was ignoring the value and simply always disabling the timer.


stevekanter:
This fixes the functionality originally added in #7824

PR #8101, opened 2014-09-18
added utils::getCascadeBoundingBox()

SunLightJuly:
This boundingBox will calculate all children's boundingBox every time

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -803,6 +803,49 @@ Rect Node::getBoundingBox() const
     return RectApplyAffineTransform(rect, getNodeToParentAffineTransform());
 }
 
+Rect Node::getCascadeBoundingBox(void)
+{
+    Rect cbb;
+
+    // check all childrens bounding box, get maximize box
+    Node* child = NULL;
+    bool merge = false;
+    for(auto object : _children)
+    {
+        child = dynamic_cast<Node*>(object);
+        if (!child->isVisible()) continue;
+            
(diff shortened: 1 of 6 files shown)


cocos/2d/CCNode.h
```diff
@@ -1079,6 +1079,11 @@ class CC_DLL Node : public Ref
     /** @deprecated Use getBoundingBox instead */
     CC_DEPRECATED_ATTRIBUTE inline virtual Rect boundingBox() const { return getBoundingBox(); }
 
+    /**
+     * This boundingBox will calculate all children's boundingBox every time
+     */
+    virtual Rect getCascadeBoundingBox(void);
```
minggo: I think it is better to be a helper function into `utils`.

cocos/2d/CCNode.cpp
```diff
@@ -803,6 +803,49 @@ Rect Node::getBoundingBox() const
     return RectApplyAffineTransform(rect, getNodeToParentAffineTransform());
 }
 
+Rect Node::getCascadeBoundingBox(void)
+{
+    Rect cbb;
+
+    // check all childrens bounding box, get maximize box
+    Node* child = NULL;
```
minggo: NULL -> nullptr

SunLightJuly: fixed

cocos/2d/CCNode.h
SunLightJuly: moved

PR #8103, opened 2014-09-18
Fixed compile error when I turned off physics definitions.

cpascal:
Fixed the following compile error.
cocos2d\extensions\physics-nodes\ccphysicssprite.cpp(305): error C4716:
'cocos2d::extension::PhysicsSprite::getRotation' : must return a value
and more..

Conditions:
CC_USE_PHYSICS = 0
CC_ENABLE_CHIPMUNK_INTEGRATION = 0
CC_ENABLE_BOX2D_INTEGRATION = 0


minggo:
But `CC_USE_PHYSICS` is used for physics integration in engine. It is not used for using external physical engine.

cpascal:
@minggo Is it okay to use "#if (!CC_ENABLE_CHIPMUNK_INTEGRATION) and (!CC_ENABLE_BOX2D_INTEGRATION)" instead of "#if CC_USE_PHYSICS"?

minggo:
@cpascal 
I think it is ok.
The modification seems good.

cpascal:
I've checked the build server log at http://115.28.134.83:8000/job/cocos-2dx-pull-request-build/5250/node=linux_centos/console

It says "fatal: Not a git repository: external/../.git/modules/external"..
It seems that this error is not occured by this code.

darkdukey:
[ci rebuild]

PR #8111, opened 2014-09-21
fix compile on ubuntu.

joewan:


joewan:
you can reproduce this problem on ubuntu with these steps:
cocos new -p com.xxx.test33 -l lua test33
cocos compile -p linux

Distributor ID: Ubuntu
Description:    Ubuntu 14.04.1 LTS
Release:    14.04
Codename:   trusty

PR #8139, opened 2014-09-24
[CCLabel] Apply updateContent() for getStringNumLines() when Label is dirty

tttreal:
getStringNumLines() does only returning _currNumLines, so, It will return incorrect value right after Label is changed.
To avoid above, if Label is dirty, call updateContent() before return _currNumLines.

Test:
This is tested on Mac


minggo:
@Dhilan007 
Could you please review it.

WenhaiLin:
@minggo It's OK.We should merge it.

minggo:
Thanks guys.

Issue #8133, opened 2014-09-24
EditBox has wrong behavior on iOS 8


minggo:
When you click EditBox, soft keyboard pops up wrongly.

PR #8149, opened 2014-09-25
fix ios8 EditBox orientation problem

zilongshanren:


zilongshanren:
fixed #8133

kevinvan44:
Thank you for this solution, however, I find the whole view stays still as the keyboard showing up. How can I fix that?

zilongshanren:
@kevinvan44  Which device and which iOS version do you use? How can I reproduce this issue?

PR #8162, opened 2014-09-25
3D Light

super626:
3D light support for 3D sprite
There are 4 types of lights, Ambient light, Directional light, Point light, Spot light

Usage,
auto light = AmbientLight::create(Color3B(200, 200, 200));
light->setLightFlag(LightFlag::LIGHT0);
addChild(light);

auto sprite = Sprite3D::create("xxx.c3b");
addChild(sprite);
sprite->setLightMask(LightFlag::LIGHT0); // this 3d sprite only receive light from LIGHT0

NOTE:
1. There are shaders with light and without light
2. Use shader with light when there are lights whose flag AND sprite's mask is true, use shader without light other wise.
3. Directional light, Point light, Spot light will not take effect when there is not normal on the model.  Ambient light affect on models both with normal and without normal.


minggo:
There is conflict and failed to build.

cocos/3d/CCLight3D.h
```diff
@@ -0,0 +1,272 @@
...
+    LIGHT14 = 1 << 14,
+    LIGHT15 = 1 << 15,
+};
+
+class CC_3D_DLL BaseLight3D : public Node
+{
+public:
+    
+    //get light type
+    virtual LightType getLightType() = 0;
```
minggo: Add const?

cocos/3d/CCLight3D.h
```diff
@@ -0,0 +1,272 @@
...
+    //get light type
+    virtual LightType getLightType() { return LightType::DIRECTIONAL; }
+    /**
+     * Creates a direction light.
+     * @param direction The light's direction
+     * @param color The light's color.
+     *
+     * @return The new direction light.
+     */
+    static DirectionLight3D* create(const Vec3 &direction, const Color3B &color);
```
minggo: Please put static method declaration at first.

cocos/3d/CCSprite3D.h
```diff
@@ -112,6 +112,10 @@ class CC_3D_DLL Sprite3D : public Node, public BlendProtocol
     void setCullFace(GLenum cullFace);
     // set cull face enable or not
     void setCullFaceEnabled(bool enable);
+    
+    /** light mask getter & setter, light works only when _lightmask & light's flag is true, default value of _lightmask is 0xffff */
+    void setLightMask(unsigned short mask) { _lightMask = mask; }
+    unsigned short getLightMask() const { return _lightMask; }
```
minggo: Why not use unsigned int? unsigned int is faster than unsigned short.

cocos/renderer/CCMeshCommand.cpp
```diff
@@ -46,6 +48,33 @@ static GLenum s_cullFace = 0;
 static bool   s_depthTestEnabled = false;
 static bool   s_depthWriteEnabled = false;
 
+struct DirectionLightUnifromNames
```
minggo: DirectionLightUnifromNames -> DirectionLightUniformNames

cocos/renderer/CCGLProgramCache.h
```diff
@@ -92,6 +92,8 @@ class CC_DLL GLProgramCache : public Ref
 private:
     bool init();
     void loadDefaultGLProgram(GLProgram *program, int type);
+    
+    std::string getShaderMacrosForLight() const;
```
minggo, line 96: What's shader macro?

tests/cpp-tests/Android.mk
```diff
@@ -0,0 +1,200 @@
+LOCAL_PATH := $(call my-dir)
```
minggo: This file is removed, please add new test cases in `tests/cpp-tests/proj.android/jni/Android.mk`.

cocos/3d/CCSprite3D.h
super626: Ok, let us use unsigned int.
I thought that unsigned short is enough and it can take less memory.

minggo: Yep, unsigned short will use less memory. 
But it will not save too much because few variables use it.

cocos/renderer/CCGLProgramCache.h
super626, line 96: We need to define the number of max support directional light, point light, and spot light in shader.
So we generate the macro using the max support light from configuration file.

cocos/3d/CCLight3D.h
super626: Done

super626:
[ci rebuild]

super626:
[ci rebuild]

PR #8164, opened 2014-09-26
move logic about camera from Director to scene

super626:
Discussion is here,
http://discuss.cocos2d-x.org/t/3d-light-discussion/16928/34

Please do not merge before we confirm it on the forum.


super626:
Riq agrees

PR #8167, opened 2014-09-26
bugfix for UI animation playing crash if GUI JSON file is loaded again.

hongliang-goudou:
_actionDic is a std::unordered_map, the insert() function will not work if the key is already existed. If a UI JSON file is loaded by GUIReader more than once, the animation information should update in the internal _actionDic map, as the rootWidget pointer is different.

Detail see: http://blog.segmentfault.com/hongliang/1190000000659937


hongliang-goudou:
@minggo Here's the new PR. Sorry about the previous one.

minggo:
Thanks @hongliang-goudou .

chengstory:
@minggo  I think it may be merged and  thanks @hongliang-goudou

minggo:
Thanks, guys.

PR #8168, opened 2014-09-26
export lua function "cc.Device::getDPI()"

hongliang-goudou:
export lua function cc.Device:getDPI()

For those developers who need to get the current device DPI in Lua environment. and also `setAccelerometerInterval()` and `setAccelerometerEnabled()` functions in the CCDevice.cpp class.


minggo:
@samuele3hu 
Could you please review it?

samuele3hu:
I have already checked it,it's ok.Thanks @hongliang-goudou.

minggo:
Thanks.

PR #8175, opened 2014-09-26
CCImage::saveToFile is implemented for Mac.

denisdolgikh:
Used the code from IOS and adapted it for Mac. RenderTextureTest tests it.


minggo:
Thanks @denisdolgikh.
But we will not add new features in v2.

PR #8184, opened 2014-09-27
Fix WP8 app exit with back button

timotei:
CEGLView end() is called from the CCDirector's
end() method, so we should terminate the app then.

### DIFF
--- a/cocos2dx/platform/wp8-xaml/Cocos2dRenderer.cpp
+++ b/cocos2dx/platform/wp8-xaml/Cocos2dRenderer.cpp
@@ -97,17 +97,18 @@ IAsyncAction^ Cocos2dRenderer::OnSuspending()
 // user pressed the Back Key on the phone
 void Cocos2dRenderer::OnBackKeyPress()
 {
-    // handle the backkey in your app here.
-    // call Cocos2dEvent::TerminateApp if it is time to exit your app.
-    // ie. the user is on your first page and wishes to exit your app.
-    // uncomment next line and comment TerminateApp to respond with keyBackClicked() in layers, remember to add init layer use setKeypadEnabled(true)
-    // CCDirector::sharedDirector()->getKeypadDispatcher()->dispatchKeypadMSG( ccKeypadMSGType::kTypeBackClicked );
-    m_delegate->Invoke(Cocos2dEvent::TerminateApp);
+    CCDirector::sharedDirector()->getKeypadDispatcher()->dispatchKeypadMSG(ccKeypadMSGType::kTypeBackClicked);
 }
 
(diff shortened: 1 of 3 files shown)


cocos2dx/platform/wp8-xaml/Cocos2dRenderer.cpp
```diff
@@ -97,17 +97,18 @@ IAsyncAction^ Cocos2dRenderer::OnSuspending()
 // user pressed the Back Key on the phone
 void Cocos2dRenderer::OnBackKeyPress()
 {
-    // handle the backkey in your app here.
-    // call Cocos2dEvent::TerminateApp if it is time to exit your app.
-    // ie. the user is on your first page and wishes to exit your app.
-    // uncomment next line and comment TerminateApp to respond with keyBackClicked() in layers, remember to add init layer use setKeypadEnabled(true)
-    // CCDirector::sharedDirector()->getKeypadDispatcher()->dispatchKeypadMSG( ccKeypadMSGType::kTypeBackClicked );
-    m_delegate->Invoke(Cocos2dEvent::TerminateApp);
+    CCDirector::sharedDirector()->getKeypadDispatcher()->dispatchKeypadMSG(ccKeypadMSGType::kTypeBackClicked);
```
timotei, line 100: The reason for this change is that, the ending of the app can/and probably should be done, at least from the cocos C++ game impl, by calling `end` on the `CCDirector`. Otherwise every app will quit instantly on back key, even if it's not on the main screen :)

The `TerminateApp` is sent when the view has been terminated.

timotei:
Any news on this? @minggo  Thanks!

minggo:
@timotei 
Sorry about late reply.
We are in holiday these days.
We will come back on Oct.8.

timotei:
Any updates on this? :)

minggo:
@stammen
Could you please take a look?

stammen:
I will also look at this for the Universal App.

timotei:
Any feedback? Thanks!

stammen:
I am looking at this now. The problem is that I need to know if the app can exit BEFORE I issue 

CCDirector::sharedDirector()->getKeypadDispatcher()>dispatchKeypadMSG(ccKeypadMSGType::kTypeBackClicked);

since it is no longer allowed to terminate the Windows Phone 8.1 app. I need to respond correctly to the
backbutton event received from the OS. I am working on a proposal to coco2d-x to fix this problem.

The current pull request will not work as the call to CCDirector::sharedDirector()->getKeypadDispatcher() is on the wrong thread.

zilongshanren:
@timotei 
As mentioned by @stammen , this PR couldn't accept. 
Thanks.

PR #8185, opened 2014-09-27
Scale9Sprite improvements

newnon:
small bugs fix
small optimizations
support for cropped sprite frames
Useful for buttons. On old behavior, if you have images for different states with glow effect for example, and use sprite sheets with crop, on state change you can see visual move or size change.

### DIFF
--- a/cocos/ui/UIScale9Sprite.cpp
+++ b/cocos/ui/UIScale9Sprite.cpp
@@ -314,7 +314,8 @@ y+=ytranslate;         \
         
         // center center
         TRANSLATE_X(x, y, left_w);
-        Rect centerbounds = Rect(x, y, center_w, center_h);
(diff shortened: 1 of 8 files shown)


minggo:
@andyque 
Could you please review it?
Thanks.

zilongshanren:
@newnon 
Hi, thanks for you patch.
But it seems this patch break some tests of Slider and LoadingBar.
Please check out them.

zilongshanren:
[ci rebuild]

zilongshanren:
@newnon 
Any update?

newnon:
Yes

zilongshanren:
@newnon 
Thanks for you hard work.
@minggo 
I think we could merge it now.

PR #8187, opened 2014-09-28
fixed crack on android when schedule mixed with "TimerTargetCallback" and other “Timer”

martin-magang:
when a node was scheduled with same “Timer” that is not "TimerTargetCallback", call "schedule" method with parameter "key", it will crack on same android devices at line "if (key == timer->getKey())".  
"static_cast" is an unsafe conversion from a pointer of a base class to a pointer of a derived class, timer->getKey() will access over range , so i think it should replace with "dynamic_cast".

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -311,8 +311,11 @@ void Scheduler::schedule(const ccSchedulerFunc& callback, void *target, float in
     {
         for (int i = 0; i < element->timers->num; ++i)
         {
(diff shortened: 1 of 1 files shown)


minggo:
How to reproduce it?

joewan:
If it is single-inheritance, there is no difference between "static_cast" and "dynamic_cast".
But I suggest we'd better use dynamic_cast when down-casting. @minggo

minggo:
> If it is single-inheritance, there is no difference between "static_cast" and "dynamic_cast".

Could you show me any docs about this?

And i will not merge it now since we are releasing a new version. And we don't have enough time to test it fully.

minggo:

PR #8188, opened 2014-09-28
V3 pipu protocol buffers

pipu:
1. Add google protocol buffers files into external
2. Create that parse files for published files of Cocos Studio Editor
3. Modify Node UI Test
4. Modify CocosStudioActionTimelineSceneTest


minggo:
Please use prebuilt version for protocol buffer.

PR #8196, opened 2014-09-29
Added LongPressed touch event to ui::Widget.

cpascal:
Many games(mostly on Japanese - Puzzle & Dragons, ..) are using
LongPressed touch action. (to show information?)

It only works if Widget::setLongPressTime(time) is set.


zilongshanren:
@ciaranj 
Thanks for providing this patch.
But why it has so many commit logs in your PR?
Could you please sync your v3 branch at first and then checkout a new branch, applying the path and send the PR again?
The easy way should be like this:

```
git checkout v3
git pull upstream v3
git checkout -b addLongTimeTouch
git cherry-pick a4dbc5c  1c134bd  c948c98
git push origin addLongTimeTouch

```

cpascal:
Okay, I'll make new PR.
Thank you very much.

PR #8204, opened 2014-09-29
normalize path  on android.

litefeel:
because AAssetManager_open do not support '../' in path.
for example:

can not open `test/../img.png` in android.
open `img.png` is ok.

this patch will change `test/../img.png` to `img.png`.


minggo:
Change `test/../img.png` to `img.png` is strange.

litefeel:
@minggo this patch is remove `..` in path.
for example:
old: `test/dir/../xx.png`
now: `test/xx.png`

old: `test/.../xx.png`
now: `xx.png`

old: `test/dir/dir/../../xx.png`
now:`test/xx.png`

exception:
old: `../xx.png`
now: `../xx.png`

old: `../test/../xx.png`
now: `../xx.png`

minggo:
Yep, i know what is PR do.
I just curious  why should do it like this?
Developer should pass correct path, or it should get error information.
It is not a good idea  to translate it silently internal.

joewan:
we have also met this problem,
I think we need to solve this bug,
or provide functions to normalize the filepath.

First,  Engine need to provide utils check whether the filepath is correct or not!
Second, Engine need to provide platform consistence for game developers, some platfoms support "..",  but some not. 

Just my own opinion above.

发送时间: 2014年9月30日(星期二) 中午11:09

主题: Re: [cocos2d-x] remove '../' at path on android. (#8204)

Yep, i know what is PR do.
 I just curious  why should do it like this?
 Developer should pass correct path, or it should get error information.
 It is not a good idea  to translate it silently internal.

litefeel:
@minggo 
i use cocostudio ui, for share plist, i put it to common dir.
and correct path to common dir in  .ExportJson.
so the path has `../`.

minggo:
@joewan 
Yep, you are right.
I am considering offer a new file system that will handle common file operation.
Now `FileUtils` is simple.

litefeel:
win and ios support `../`.
android not support `../`

minggo:
Yep, Android is special. Its resources are in .apk after installing.

minggo:
I have created a ticket of new file system here: http://discuss.cocos2d-x.org/t/about-file-system/17357.

cocos/platform/android/CCFileUtils-android.h
```diff
@@ -61,6 +61,8 @@ class CC_DLL FileUtilsAndroid : public FileUtils
     /* override funtions */
     bool init();
 
+    std::string getNewFilename(const std::string &filename) const override;
```
WenhaiLin: please add 'virtual' keyword

litefeel:
@Dhilan007  added it.

minggo:
Now there are two PRs for this issue.
I think we should close one.

@lite3 
@Dhilan007 
Any idea?

litefeel:
@minggo @Dhilan007 
i think will close @Dhilan007 .

because the path can not cache when normalize path on `AAssetManager_open`.

litefeel:
[ci rebuild]

litefeel:
[ci rebuild]

minggo:
> because the path can not cache when normalize path on AAssetManager_open

What did you mean?
Could you describe it in more detail?

minggo:
Ok, i got your meaning.
But there is a problem of this PR.
Because getNewFileName is not used for this purpose, it is used to get a alias of a file name.

litefeel:
@minggo 
i have a path `script.lua`, the real path is `script/script.lua`.
first times, search path to `script/script.lua`, and put _fullPathCache.
second times, get path from _fullPathCache.

litefeel:
@minggo yes.
getNewFileName 
1. get a alias of a file name
2. normalize path

i think is ok.

minggo:
Yep, you are right.

@Dhilan007 
Could you please test it?
And i will close your pull request.

WenhaiLin:
@minggo It's work as expected.We can merge this pr.
@lite3 Thanks very much.

PR #8205, opened 2014-09-29
Added LongPressed touch event to ui::Widget.

cpascal:
Many games(mostly on Japanese - Puzzle & Dragons, ..) are using
LongPressed touch action. (to show information?)

It only works if Widget::setLongPressTime(time) is set.

### DIFF
--- a/cocos/ui/UIWidget.cpp
+++ b/cocos/ui/UIWidget.cpp
@@ -143,6 +143,7 @@ _brightStyle(BrightStyle::NONE),
 _touchBeganPosition(Vec2::ZERO),
 _touchMovePosition(Vec2::ZERO),
 _touchEndPosition(Vec2::ZERO),
+_longPressTime(0.0f),
 _touchEventListener(nullptr),
 _touchEventSelector(nullptr),
 _actionTag(0),
@@ -680,6 +681,16 @@ bool Widget::isSwallowTouches()const
     return false;
 }
 
+void Widget::setLongPressTime(float time)
+{
+    if (time <= 0.0f )
+    {
+        unschedule(schedule_selector(Widget::checkLongPress));
+    }
+
+    _longPressTime = time;
+}
+
 bool Widget::onTouchBegan(Touch *touch, Event *unusedEvent)
 {
     _hitted = false;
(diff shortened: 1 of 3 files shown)


zilongshanren:
@cpascal 
Hi, thanks very much for providing this patch.
I have only one question here?
If we call `Widget::setLongPressTime(time)` with a parameter greater than 0,  does the original click event  handler and touch ended event handler still be called?

In general, when a long pressed event occurs, the original click event shouldn't be execute any more.

cpascal:
@andyque Generally, the touch ended event is used together if `Widget::setLongPressTime(time)` is called.
For example, in the party member changing scene..
'LongPressed' => `pushScene()` to show the current party member(character) information
'Touch Ended' => `pushScene()` to change party member

And I also think 'Touch Began' can be used if someone wants to highlight other widgets when touch began..

zilongshanren:
@cpascal 
Suppose we have a weapon item here, if we click it(means click it in a short time), it will show the details info of the weapon.  If we long click the item,  it will trigger the long click event handler thus will enable drag and drop feature.

What if we drag and drop it on the hero's equipment slot and then when we release our finger, the original click event will also be issued. The detail info of the weapon shows unexpected. 

What I mean is when we trigger the long press event, the original click event shouldn't be trigger any more which would make more sense.

zilongshanren:
@cpascal 
I'm confused now. 
The 'Long press' event is conflict with 'Short click'.

If we register the two listeners:

```
addTouchEventListener(const ccWidgetTouchCallback& callback);
 addClickEventListener(const ccWidgetClickCallback& callback);
```

And so we call this:

```
setLongPressTime(1.0)
```

If we click the button for 1.0+ secs, then when we release our finger, there will be three handlers get executed. Am I right? Does this behavior is expected?

If I miss something, please point out here, thanks

zilongshanren:
@cpascal 
I discuss this issue with my colleague now and he think "touch event handler" should always be called.

And the "quick tap"(we haven't implement it yet), "click" and "long click" event handler should be exclusive. Only one at a time will be called.

What's your opinion?

cpascal:
@andyque 

Because `addClickEventListner()` is fully same as `addTouchEventListner()'s TouchEnded(releaseUp) event`,
I think `addClickEventListner()` isn't needed on any other cases(even without longpressed case) if we register `addTouchEventListener()`.

## 

My opinions:

Method 1. No further touch/click event is called If a flag variable is set. (In next touch, the flag must be cleared like Node::_hitted.) So user's long-pressed callback function should set the flag if long-pressed event is processed well.
Method 2. `Node::isTouchLongPressed() (I'm sorry I'm not good in English naming)` method, so touch/click-event can recognize the touch is long-pressed or short-click.

What do you think which method is better?

## 

Games that is using 'LongPressed' doesn't use 'Drag & Drop' feature both at a time. As you said, they can be conflicted.
So the developers must choose just one of the following.

'Short-click' + 'Drag & drop'
'Short-click' + 'Long press'

I didn't consider 'Drag & Drop' feature in this code.. As you said, I think it can be used at easy drag & drop feature if we consider more..

My english is not good, please forgive me if I make you misunderstanding.

cpascal:
@andyque I've added `ui::Widget::isLongPressedTouch()` method and test code.

cpascal:
Win7 Build server says 'fatal: Not a git repository (or any of the parent directories): .git'.
Maybe build agent problem?

zilongshanren:
@cpascal  I would like to introduce gesture recognizer to resolve this issue.
http://stackoverflow.com/questions/6179347/uibutton-long-press-event

PR #8209, opened 2014-09-30
fix bug: physics crashes when remove/add children too quick.

boyu0:
Fix PhysicsTest->Contact Test crashes when click buttons very fast.

@Dhilan007 Please check this.


WenhaiLin:
@minggo It's OK.I think we should merge this pr.

PR #8542, opened 2014-10-01
fix crash

WenhaiLin:


minggo:
It is fixed in another PR by @ricardoquesada.

PR #8554, opened 2014-10-03
Scale9Sprite capInsets set fix

newnon:


minggo:
@andyque
Could you please take a look?

minggo:
[ci rebuild]

newnon:
It is just misprint i'v got , const Rect& capInsets but use Rect::ZERO instead

zilongshanren:
Thanks @newnon 
@minggo Yes, we should merge this PR.

PR #8573, opened 2014-10-08
fix #8569: LabelAtlas opacity do not change when setting parent node's opacity

### LINKED ISSUE #8569: LabelAtlas opacity do not change when setting parent node's opacity
cocos2dx version: 3.2
Add LabelAtlas to a node, and set node->setCascadeOpacityEnabled(true),  then node->setOpacity(128) do not affect labelAtlas

WenhaiLin:


WenhaiLin:
[ci rebuild]

PR #8575, opened 2014-10-08
[AUTO]: updating luabinding automatically

CocosRobot:


joewan:
@minggo this commit contains some wrong English words mis-spelling.
retunr->return.

samuele3hu:
This error was triggered by the overload function template file,i will fix it soon.

PR #8577, opened 2014-10-08
call detach method on thread exit

ntotani:
When call JniHelper::cacheEnv from native thread (not main thread), thread must call JavaVM:: DetachCurrentThread before exit.

http://developer.android.com/training/articles/perf-jni.html#threads


minggo:
@Dhilan007 
Could you please take a look?

WenhaiLin:
@ntotani Thanks.
@minggo We could merge this PR.

PR #8586, opened 2014-10-09
Fix warning of Bundle3D and modify light

super626:
1. remove unused variable from Bundle3D
2. modify return type of getDirection and getDirectionInWorld for Light, because it is not safe to return reference.


super626:
[ci rebuild]

super626:
[ci rebuild]

super626:
Fix spot light error.

PR #8590, opened 2014-10-09
Linux: fix glfw3 searching

martell:
As I don't have a linux box at hand I'm borrowing the build bot for testing.
Don't merge until the build bot succeeds :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,14 @@ include_directories(
 
 )
 
+if(LINUX)
+
+  find_package(GLFW3 REQUIRED)
+  message( STATUS "GLFW3 dirs: ${GLFW3_INCLUDE_DIRS}")
+  include_directories(${GLFW3_INCLUDE_DIRS})
+
+endif(LINUX)
+
 
 if(NOT MINGW)
 
--- a/cmake/Modules/FindChipmunk.cmake
+++ b/cmake/Modules/FindChipmunk.cmake
@@ -2,7 +2,7 @@
 # FindChipmunk
 # ------------
 #
-# Locate FindChipmunk library
+# Locate Chipmunk library
 #
 # This module defines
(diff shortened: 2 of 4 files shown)


martell:
okay so now the build finds GLFW and gets to this stage
Linking CXX static library ../lib/libcocos2d.a

martell:
okay so this fully builds and gets to the end of tests.cpp

If fails because of this

[100%] Building CXX object tests/cpp-tests/CMakeFiles/cpp-tests.dir/proj.linux/main.cpp.o
Linking CXX executable ../../bin/cpp-tests/cpp-tests
/usr/bin/ld: cannot find -lbox2d

All we need is prebuilt box2d tinyxml and minizip and this will link on the build bot.

@minggo @ricardoquesada 

Can you merge this please

minggo:
@martell
Thanks for fixing it.

But it is strange we use prebuilt and resource codes in the same time.

minggo:
I need to make jenkins work as soon as possible, or it will break all pull requests.

martell:
Yes the build bot is important.

I know it sounds strange.

It should be fine to drop in the prebuild libs if only just for linux into here
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin

This is where to add the library
cocos2d-x-3rd-party-libs-bin/Box2D/prebuilt/linux/xx-bit/libbox2d.a
and the headers
cocos2d-x-3rd-party-libs-bin/Box2D/include

even if you leave source code in the folder for all the other build systems it will still work.
Cmake will ignore the source code and just grab the include dir and the prebuilt dir

Hope you have the buildbot up soon with this :)

minggo:
Thanks @martell.

We will use source code at the moment because it is more quickly. 
We can use prebuilt version step by step that will not break CI.

Sorry about that.

PR #8592, opened 2014-10-09
wp8: use CC_STATIC because libs are static

martell:
@ricardoquesada 
As promised a fix for windows phone.
In fact it wasn't me who broke it.

It was broken since someone re-enabled dll support on win32 by uncommenting the dllexport and dllimport defines. 

CC_STATIC was introduced in case we had a scenario like this where we are using static libs.
Looks like i was right and we needed it :D

We need someone to convert all the different components in the wp8 branch into libcocos2d.

@minggo can you merge this also please :)


ricardoquesada:
Thanks!

PR #8595, opened 2014-10-09
Fix linux compile error

zilongshanren:


zilongshanren:
@martell
Hi, I have modified the linux cmake part to make Jenkins happy. If you are working on the prebuilt version of box2d, tinyxml, unzip and xx libraries. please remove the `add_subdirectory` section in the CMakeLists.txt file.

martell:
@andyque 
It looks good :)
You haven't broken anything from a prebuilt point of view also
I just need to add more find packages so that you don't have to hard code /usr/include anymore
good job

PR #8603, opened 2014-10-09
Fix UIButton Scale 9 Highlight

ahlwong:
Currently, when you press a UIButton with Scale 9 enabled that doesn't have a pressed texture, the button is tinted gray. When you let go of the button, the button remains gray. This pull request reverts the tint.

### DIFF
--- a/cocos/ui/UIButton.cpp
+++ b/cocos/ui/UIButton.cpp
@@ -374,11 +374,19 @@ void Button::onPressStateChangedToNormal()
     }
     else
     {
-        _buttonNormalRenderer->stopAllActions();
-        _buttonNormalRenderer->setScale(_normalTextureScaleXInSize, _normalTextureScaleYInSize);
-        _titleRenderer->stopAllActions();
-        _titleRenderer->setScaleX(_normalTextureScaleXInSize);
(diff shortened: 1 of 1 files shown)


cocos/ui/UIButton.cpp
```diff
@@ -374,11 +374,19 @@ void Button::onPressStateChangedToNormal()
     }
     else
     {
-        _buttonNormalRenderer->stopAllActions();
-        _buttonNormalRenderer->setScale(_normalTextureScaleXInSize, _normalTextureScaleYInSize);
-        _titleRenderer->stopAllActions();
-        _titleRenderer->setScaleX(_normalTextureScaleXInSize);
-        _titleRenderer->setScaleY(_normalTextureScaleYInSize);
+        if (_scale9Enabled)
```
zilongshanren, line 377: Why should we set the color of _buttonNormalRenderer to `Color3B::WHITE` when it is scale9 enabled?
Could you write a test to reproduce this issue?

zilongshanren:
@ahlwong 
Thanks, I have reproduce the issue. I will check why this issue happens.

zilongshanren:
@ahlwong 
I got it. Thanks:)

Issue #8568, opened 2014-10-06
FileUtils::removeFile doesn't work iOS


ahlwong:
In Cocos2d-x 3.3beta0

The FileUtils::removeFile doesn't seem to work. It generates the correct command but it doesn't seem to do anything. Perhaps it has to do with this Stack Overflow post? http://stackoverflow.com/questions/11291164/system-command-not-working-when-enabling-app-sandboxing

"rm -f \"/Users/my_user/Library/Developer/CoreSimulator/Devices/2821C4FC-9316-4183-A3A9-62E19BAE4A65/data/Containers/Data/Application/FEDDBD50-526B-4C88-A58B-816190CE6A47/Documents/db.sqlite\""

minggo:
@Dhilan007 
Could you please take a look?

WenhaiLin:
@ahlwong I test it on IOS7.1 Simulator,it's work normal.I think you passes the wrong path.["...db.sqlite\"]

ahlwong:
@Dhilan007 I just tested it again and it's not working in the simulator or on my devices. It is working on Android. Here are some more details:

iOS 8 - iPhone 6 simulator - Xcode 6.0.1
rm -f "/Users/my_user/Library/Developer/CoreSimulator/Devices/2821C4FC-9316-4183-A3A9-62E19BAE4A65/data/Containers/Data/Application/173CEBC8-04B3-47EA-BD90-A8E646220CD4/Documents/db.sqlite"

iOS 7.1.2 - iPhone 5 device - Xcode 6.0.1
cocos2d: rm -f "/var/mobile/Applications/348EC781-C8C3-49B0-BF12-523B0EA7FA63/Documents/db.sqlite"

Android 4.4.2 Device
10-09 11:15:10.538: D/cocos2d-x debug info(5910): rm -f "/data/data/com.my.app/files/db.sqlite"
10-09 11:15:10.568: D/CCFileUtils-android.cpp(5910): relative path = database/db.sqlite

I'm creating the path like so:
FileUtils::getInstance()->getWritablePath() + "db.sqlite"

My copyFile code and reading the file works so the path should be correct. Also, the removeFile is still returning true even when the file sin't deleted.

ahlwong:
@Dhilan007 The exit code I'm seeing on iOS is 5 whereas in Android it is 0. The code that checks the exit code is:

if (system(command.c_str()) >= 0) // an exit code > 0 means an error occurred

it should be:

[...]

PR #8608, opened 2014-10-10
issue #8568: Fixed FileUtils::removeFile doesn't work on iOS

WenhaiLin:


cocos/platform/CCFileUtils.cpp
```diff
@@ -1103,6 +1107,18 @@ bool FileUtils::createDirectory(const std::string& path)
 #endif
 }
 
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
+int unlink_cb(const char *fpath, const struct stat *sb, int typeflag, struct FTW *ftwbuf)
```
minggo: Please use static function.

Issue #8567, opened 2014-10-06
ui::Button Title Only Content Size Not Available Immediately


ahlwong:
In Cocos2d-x 3.3beta0

When creating a button with a title only, the buttons content size is not immediately updated.

Button\* button = Button::create();
button->setTitle("My Title");
button->getContentSize(); // == Size::ZERO, not available yet

I fixed this by adding an updateSizeAndPosition to setTitleText

void efcc::Button::setTitleText(const std::string& text)
{
    ui::Button::setTitleText(text);
    updateSizeAndPosition(_parent ? _parent->getContentSize() : Size::ZERO);
}

minggo:
@andyque 
Could you please take a look?

zilongshanren:
@ahlwong 
Thanks, I have confirmed the issue.
Consider your fix, I think the following way would be better:

```
void Button::setTitleText(const std::string& text)
{
    _titleRenderer->setString(text);
    if (_ignoreSize)
    {
        Size s = getVirtualRendererSize();
        this->setContentSize(s);
    }
}
```

Because `updateSizeAndPosition` is used for updating children node's size and position.

What's your opinion?

zilongshanren:
@ahlwong 
I have created a patch here: https://github.com/cocos2d/cocos2d-x/pull/8606  
If you find any issues, please let me know.

PR #8635, opened 2014-10-10
Fix UIButton setFontSize sync ContentSize

ahlwong:
Similar to this issue: https://github.com/cocos2d/cocos2d-x/issues/8567#issuecomment-58606780

Note: I didn't add a test case to this commit, but here is a simple one to reproduce:

ui::Button *button = ui::Button::create();
button->setTitleText("some text");
button->setTitleFontSize(30.0f); // set font size
Size c = button->getContentSize();
Size r = button->getVirtualRendererSize();
CCLOG("content size: %f x %f | %f x %f", c.width, c.height, r.width, r.height); // c != r

Also, it seems that since we are adding this setContentSize check in multiple locations that UIButton might need to be refactored in the future.


zilongshanren:
@ahlwong 
Thanks, I think we should encapsulate the following code snippet:

```
  if (_ignoreSize) {
       Size s = getVirtualRendererSize();        
       this->setContentSize(s);    
}
```

into a single protected member function, let's give git a name `updateContentSize`?
 And now we could call it in both `setTitleText` and `setTitleFontSize` function.
Could you please improve your PR like this?
Thanks:)

zilongshanren:
@ahlwong  Thanks 
When the build system is green, we will merge your PR.

zilongshanren:
[ci rebuild]

PR #8637, opened 2014-10-10
Migrate Deprecated Lua API from Luabinding of modules to folder src/cocos

dualface:
Changes:
- Don't load the deprecated API from the luabinding of modules.
- Copy all Lua script files from `cocos/scripting/lua-bindings/script/` to `<NEW_PROJECT>/src/cocos` when creating new project with the `cocos new` command.

Status:
- `templates/lua-template-default/` updated.
- tested on iOS/Android/Mac, no testing on Win32, but I think it's work fine.


cocos/scripting/lua-bindings/script/cocosstudio/CocosStudio.lua
```diff
@@ -1,11 +1,15 @@
-require "json"
-require "extern"
+
+if not json then
+    require "cocos.cocos2d.json"
+end
 
 ccs = ccs or {}
 
+require "cocos.cocosstudio.StudioConstants"
```
samuele3hu, line 8: Does it affect the lua-tests and lua-empty-test.

PR #8640, opened 2014-10-11
issue 8523:fixed relative path issue on Android

WenhaiLin:


litefeel:
@Dhilan007 fixed at #8204

PR #8642, opened 2014-10-11
mingw64: mirgrate to upstream minizip

martell:
This is to add support for upstream minizip located here
https://github.com/nmoinvaz/minizip
I added a FIXME note for other platforms.
So that this must be added to cocos3 3rd party sources and bins for them

I have submitted a PR to this repo to add support for a function only in the cocos
https://github.com/nmoinvaz/minizip/pull/17

Hopefully when that is merged we can use this as our upstream :)


ricardoquesada:
thanks!

PR #8647, opened 2014-10-11
Update build_native.py

flashjay:
copy_resources use build-cfg.json config


minggo:
Why change it like this?

This script will be removed in future, because we have cocos commands.

flashjay:
eclipse project use this script

python ${ProjDirPath}/build_native.py -b debug

minggo:
Eclipse project can use cocos command too.

flashjay:
So, eclipse project template should be changed.

minggo:
I send a pull request to use cocos console in #9708.

PR #8650, opened 2014-10-11
V3 add staggered tile map

dabingnn:


minggo:
[ci rebuild]

minggo:
[ci rebuild]

PR #8659, opened 2014-10-13
Fix one memory leak of removeFromParentAndCleanup in cocos2dx 2.x

neil-wu:
//neilwu (wuyupeng2007(AT)gmail.com) modify
Because Widget only supports Widgets as children, so if child is not a widget, 
when you call child->removeFromParentAndCleanup(), this will lead to memory leak!!! 

The following code show detail info about this bug:
    //....(TODO:load armature res)
    UIImageView\* pbg = UIImageView::create();
    pbg->loadTexture("CloseNormal.png");
    pbg->setPosition(ccp(200, 100));
    addChild(pbg);
    CCArmature\* armature = CCArmature::create("demo_armature"); //TODO:you need preload som demo res
    pbg->addNode(armature, 1, 111);

```
//check this:
armature->removeFromParentAndCleanup(true); //This will lead the bug.
```

### DIFF
--- a/extensions/CocoStudio/GUI/BaseClasses/UIWidget.cpp
+++ b/extensions/CocoStudio/GUI/BaseClasses/UIWidget.cpp
@@ -228,8 +228,21 @@ void Widget::removeChild(CCNode *child)
 
 void Widget::removeChild(CCNode *child, bool cleanup)
 {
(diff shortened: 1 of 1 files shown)


neil-wu:
Fix one memory leak of removeFromParentAndCleanup

neil-wu:
cocos2dx 2.x

minggo:
Will not accept PR of 2.x except very critical bugs fix.

PR #8667, opened 2014-10-14
Modify return type of Mesh::getAABB,  add Animate3D::create from start frame and end frame

super626:
1. Modify static AABB& Mesh::getAABB to AABB Mesh::getAABB
2. Add new interface to create Animate3D from start frame and end frame
3. Draw Sprite3D Using red color texture when texture is missing.
4. fix warning from MeshCommand.
5. Fix transparent support.


minggo:
[ci rebuild]

minggo:
@super626 
This PR fixes which issue?
Please connect issue number like `fixed #xxx` if there is a corresponding issue.

minggo:
[ci rebuild]

minggo:
The error message is 
`4>..\auto\lua_cocos2dx_3d_auto.cpp(1699): error C2668: 'cocos2d::Animate3D::create' : ambiguous call to overloaded function`.

It seems there is a problem of lua-binding.
@samuele3hu 
Could you please help to fix it?

minggo:
Oh, it is nothing about lua-binding.
It is because the two overload functions of Animate3D::create() is ambiguous.

PR #8682, opened 2014-10-14
Fixbillboard

super626:
Fix billboard computation error.
Make sure BillBoard computation base on anchor point.


super626:
[ci rebuild]

PR #8706, opened 2014-10-15
Add clipping rectangle node

SunLightJuly:
added ClippingRectangleNode and test


minggo:
Please don't modify wp8 projects since MS guys are refactoring it.

SunLightJuly:
ok

PR #8714, opened 2014-10-15
fix loading bar display bug when sprite frame is trimmed

codetypess:
there is a display bug when sprite frame texture rect size not equal to
the origion size.


minggo:
@anyque
Please take a look.

PR #8715, opened 2014-10-15
CMakes support for Mac Os X & iOS

iSevenDays:
It was not easy to make cmakes work. Brew can help with 3rd party libraries.
You can test this by executing "cocos new..." and opening project in CLion.

brew tap homebrew/versions
brew install glfw3 freetype chipmunk protobuf webp

Please notice that default demo "HelloWorldScene" searches resources within project executable dir. But resources are copied into "Resources" directory. 

p.s. iOS not tested due to certificates problem. But on Mac everything works.


minggo:
I think we should build our own libs.
Because we may use the version different from the version installed by brew.

PR #8738, opened 2014-10-16
Modify Reader:

pipu:
1. Parsing "FlipX", "FlipY", "AnchorPoint".
2. Parsing "VisibleForFrame" instead of "Visible"

### DIFF
--- a/cocos/editor-support/cocostudio/ActionTimeline/CSLoader.cpp
+++ b/cocos/editor-support/cocostudio/ActionTimeline/CSLoader.cpp
@@ -1555,17 +1555,9 @@ void CSLoader::setPropsForNodeFromXML(cocos2d::Node *node, const tinyxml2::XMLEl
         {
             node->setRotationSkewY(atof(value.c_str()));
         }
-        else if (name == "Rotation")
-        {
-
-        }
         else if (name == "ZOrder")
         {
             node->setZOrder(atoi(value.c_str()));
-        }
-        else if (name == "Visible")
-        {
-            
         }
         else if (name == "VisibleForFrame")
         {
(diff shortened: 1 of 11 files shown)


cocos/editor-support/cocostudio/ActionTimeline/CSLoader.cpp
```diff
@@ -1595,11 +1595,11 @@ void CSLoader::setPropsForNodeFromXML(cocos2d::Node *node, const tinyxml2::XMLEl
         }
         else if (name == "Visible")
         {
-            node->setVisible((value == "True") ? true : false);
+//            node->setVisible((value == "True") ? true : false);
```
minggo: If `Visible` is not supported, please remove all the codes.
And is there compatibility issue?

pipu:
I will modify it immediately.

cocos/editor-support/cocostudio/ActionTimeline/CSLoader.cpp
```diff
@@ -1587,7 +1587,7 @@ void CSLoader::setPropsForNodeFromXML(cocos2d::Node *node, const tinyxml2::XMLEl
         }
         else if (name == "Rotation")
         {
-//            node->setRotation(atoi(value.c_str()));
+
         }
```
minggo: The same issue.
And please modify all of them.

I will not mention it one by one.

pipu:
Similar code as this line:
"//            node->setRotation(atoi(value.c_str()));"
I've deleted.
at above commit "Deleting original code that parsing "Visible""

minggo:
[ci rebuild]

PR #8804, opened 2014-10-17
Modify reader in order to parse attribute "Plist"

pipu:


cocos/editor-support/cocostudio/ActionTimeline/CSLoader.cpp
```diff
@@ -1005,7 +1005,6 @@ void CSLoader::setPropsForSpriteFromProtocolBuffers(cocos2d::Node *node,
             
         case 1:
         {
-			SpriteFrameCache::getInstance()->addSpriteFramesWithFile(_protocolBuffersPath + fileNameData.plistfile());
```
VisualSJ, line 1008: Can be delete this code?
after delete,  js-tests cannot run~~~~

Issue #8806, opened 2014-10-17
[Mac] CPP template project crash


samuele3hu:

WenhaiLin:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/8807

PR #8822, opened 2014-10-17
change ios and mac template's resource struct

ghost:
more to #8806


minggo:
Did you test it?
The `font` folder is not a blue folder, and new c++ project still crash.

minggo:
Sorry, my fault.
I didn't succeed to synchronize codes.

PR #8826, opened 2014-10-17
add libprotobuf-dev to build/install-linux-dep.sh

lowstz:
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/editor-support/cocostudio/CSParseBinary.pb.h 
This file include Protobuf Buffers headers file, project will build failure when missing libprotobuf-dev, add  libprotobuf-dev to fix it.


minggo:
Now we use source code, don't need it.

PR #8827, opened 2014-10-17
don't show runtime on top by default

cocoscodeide:


minggo:
@honghui
Why did you modify cocos studio reader?

honghui:
I have fixed it by pulling the newest code again.

PR #8842, opened 2014-10-18
add needed link library for linux

zhaijialong:
when i try to link with libcocos2d.a in linux , got this error

```
/usr/bin/ld: /home/joshua/github/cocos2d-js/frameworks/js-bindings/cocos2d-x/external/websockets/prebuilt/linux/64-bit/libwebsockets.a(client.c.o): undefined reference to symbol 'BIO_ctrl@@OPENSSL_1.0.0'
//lib/x86_64-linux-gnu/libcrypto.so.1.0.0: error adding symbols: DSO missing from command line
```

after adding ssl and crypto , everything turns ok

### DIFF
--- a/cocos/CMakeLists.txt
+++ b/cocos/CMakeLists.txt
@@ -145,7 +145,7 @@ if(MINGW)
 elseif(WINDOWS)
   set(PLATFORM_SPECIFIC_LIBS libjpeg libpng libwebp libtiff libcurl_imp libwebsockets freetype250 glfw3 glew32 opengl32 libiconv libzlib)
 elseif(LINUX)
-  set(PLATFORM_SPECIFIC_LIBS jpeg webp tiff freetype curl websockets
+  set(PLATFORM_SPECIFIC_LIBS jpeg webp tiff freetype curl websockets ssl crypto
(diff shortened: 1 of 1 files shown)


minggo:
What's your linux system?
We didn't meet the issue.

zhaijialong:
It's Ubuntu 14.04 x64.  Seems the prebuilt libwebsockets depends on libssl and libcrypto.
I will have a try again tomorrow, and close this pr if no problem.

zhaijialong:
@minggo 
still facing this issue. Did you test websockets? 
I run `nm -u libwebsockets.a`, and can find lots of symbols which have name beginning with `SSL_`.

minggo:
Sure.
There is a test case of WebSocket in cpp-tests.

zhaijialong:
@minggo
Yes, there is , but it seems not compiled into ccp-tests on linux.

zhaijialong:
https://github.com/cocos2d/cocos2d-x/blob/v3/tests/cpp-tests/Classes/ExtensionsTest/ExtensionsTest.cpp#L56

no linux here

PR #8845, opened 2014-10-19
Decoupling the calculation of FPS from displaying FPS

dsternfeld:
Prior to this change the functions Director::getSecondsPerFrame and Director::getFrameRate returned garbage data unless _displayStats was set to true. This makes it so getSecondsPerFrame and getFrameRate are usable even when you aren't visually displaying them on the screen by decoupling the calculation from the debug rendering.


minggo:
[ci rebuild]

minggo:
@dsternfeld 
Could you pleas fix conflict?
I will merge it as soon as you fix the conflict.

Thanks.

PR #8846, opened 2014-10-19
* change default opacity value

thenewvu:
If the default opacity value is 0 and you don't change it, when you want to set another color by calling setColor(const Color3B&), you must call setOpacity(GLubyte) also to set another non-zero opacity value, otherwise, you still have a black LayerColor. Sometime, it can make someone confused.


minggo:
It will break compatibility.
I think we can not modify it like this.

thenewvu:
@minggo : Could you explain more about 'compatibility' ? Such as an example. Actually I've just decided to dig into cocos2d-x. Great to learn something new. Thanks.

minggo:
The default value is changed that break the compatibility.
Many developers use `Layer`, it will  change Layer's default behavior.

Issue #8816, opened 2014-10-17
new AssetManager will change the resources in Resources folder on linux


dabingnn:
3.3 rc

PR #8855, opened 2014-10-20
change asset manager to use a new resource, which will not affect other ...

dabingnn:
Fix asset manager EX test case, which will affect the result of other test cases before.


samuele3hu:
I think you should do the same change for related lua test case.

minggo:
Please write the corresponding issue number if exist.

dabingnn:
fix #8816

PR #8857, opened 2014-10-20
fix memory leak when add listener to ui object

codetypess:


samuele3hu:
It's right for releasing the reference of lua function,thanks @zhongfq

codetypess:
@samuele3hu it is unnecessary to add those functions?

PR #8885, opened 2014-10-21
get_num_of_cpu() returns 1 after update to OS X Yosemite

ZhangZiqi:


minggo:
We should test if it has problem that not using OS X Yosemite.

minggo:
@zii 
Thanks for testing.

minggo:
@ZhangZiqi 
So we can not change it like this.
Should only change it when using OS X Yosemite?

ZhangZiqi:
@minggo 
Why？According to @zii，it is usable on the other platforms.

minggo:
@ZhangZiqi 
Oh, i misunderstood.
Sorry.

pyrosphere:
The exact same code and issue is also present in [templates/cpp-template-default/proj.android/build_native.py](https://github.com/cocos2d/cocos2d-x/blob/3e124222d95a6944ba228c1b0608c00f7e8e4d4a/templates/cpp-template-default/proj.android/build_native.py#L22-23)

PR #8892, opened 2014-10-21
Removes uthash in CCGLProgram,uses std::unordered_map instead

dumganhar:


dumganhar:
[ci rebuild]

dumganhar:
@Minggo, how about this PR?

PR #8905, opened 2014-10-22
rewrite http module while preserve compatibility

kezhuw:
In #7192, I mentioned that several problems in HttpClient.

Poor quality of origin code makes me difficult to fix it, so I rewrite it.
Curl related part is untouched.

The core difference between two versions of HttpClient is that:
Now HttpClient{s} are instances to hold states instead of a static object to export interfaces.
So, it is possible to write an interface `static void HttpClient* newInstance();` in HttpClient, to get different HttpClients for customization.

PS:
After rebase to tip of v3, cocos2d-x is failed to build. 
Some errors:
clang: error: no such file or directory: '/Users/kezhuw/Sources/cocos2d-x/external/Box2D/Dynamics/Joints/b2WeldJoint.cpp'

'google/protobuf/stubs/common.h' file not found.

I am not a cocos2d-x guy, even toy project. How to fix it?

### DIFF
--- a/cocos/network/HttpClient.cpp
+++ b/cocos/network/HttpClient.cpp
@@ -381,25 +381,15 @@ static void performTasks(Scheduler *scheduler, HttpTaskQueue *tq) {
     }
 }
 
-class HttpBackend {
-public:
-
-    HttpTaskQueue* getTaskQueue() {
-        if (!_thread.joinable()) {
-            Scheduler *scheduler = Director::getInstance()->getScheduler();
-            _thread = std::thread(performTasks, scheduler, &_tqueue);
-        }
-        return &_tqueue;
-    }
-
-private:
-    HttpTaskQueue _tqueue;
-    std::thread _thread;
-};
-
 static HttpTaskQueue* getBackendTaskQueue() {
-    static HttpBackend httpBackend;
-    return httpBackend.getTaskQueue();
+    static HttpTaskQueue *tqueue = nullptr;
+    if (tqueue == nullptr) {
+        tqueue = new HttpTaskQueue;
+        Scheduler *scheduler = Director::getInstance()->getScheduler();
+        std::thread taskThread(performTasks, scheduler, tqueue);
+        taskThread.detach();
+    }
+    return tqueue;
 }
 
 static HttpClient *_pHttpClient = nullptr;
--- a/cocos/network/HttpClient.cpp
+++ b/cocos/network/HttpClient.cpp
@@ -26,46 +26,24 @@
 
 #include "HttpClient.h"
 
-#include <thread>
-#include <queue>
-#include <condition_variable>
-
-#include <errno.h>
-
 #include "base/CCVector.h"
 #include "base/CCDirector.h"
 #include "base/CCScheduler.h"
+#include "platform/CCFileUtils.h"
 
 #include "curl/curl.h"
 
-#include "platform/CCFileUtils.h"
+#include <errno.h>
+
+#include <thread>
+#include <queue>
+#include <condition_variable>
 
 NS_CC_BEGIN
 
 namespace network {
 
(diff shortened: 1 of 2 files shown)


minggo:
@kezhuw 
Thanks.

You should run `download-deps.py` to get 3rd party libs.

minggo:
@dumganhar 
Could you please take a look?

dumganhar:
@minggo, I'll review it soon. :)

cocos/network/HttpClient.cpp
```diff
@@ -80,148 +58,25 @@ static size_t writeData(void *ptr, size_t size, size_t nmemb, void *stream)
...
-        curl_easy_setopt(handle, CURLOPT_SSL_VERIFYHOST, 2L);
-        curl_easy_setopt(handle, CURLOPT_CAINFO, s_sslCaFilename.c_str());
+    HttpTask(HttpClient *client_, HttpRequest* request_, const std::string &cookiesFilename_, const std::string &sslCaFilename_)
+        : client(client_)
+        , response(new HttpResponse(request_))
+        , timeoutForConnect(client_->getTimeoutForConnect())
+        , timeoutForRead(client_->getTimeoutForRead())
+        , cookiesFilename(cookiesFilename_)
+        , sslCaFilename(sslCaFilename_) {
+        client->retain();
```
dumganhar, line 70: The client instance is retained here, where to release it?

cocos/network/HttpClient.cpp
```diff
@@ -423,33 +316,120 @@ static void processResponse(HttpResponse* response, char* errorBuffer)
...
+        _notEmpty.notify_one();
+    }
+
+    HttpTask* takeTask() {
+        std::lock_guard<std::mutex> lock(_mutex);
+        while (_immediateTasks.empty() && _tasks.empty()) {
+            _notEmpty.wait(_mutex);
+        }
+        if (!_immediateTasks.empty()) {
+            HttpTask *task = _tasks.front();
```
dumganhar: Here should be `_immediateTasks.front()`, right?

cocos/network/HttpClient.cpp
```diff
@@ -423,33 +316,120 @@ static void processResponse(HttpResponse* response, char* errorBuffer)
...
+    HttpTaskQueue _tqueue;
+    std::thread _thread;
+};
+
+static HttpTaskQueue* getBackendTaskQueue() {
+    static HttpBackend httpBackend;
+    return httpBackend.getTaskQueue();
+}
+
+static HttpClient *_pHttpClient;
```
dumganhar: Please initialize the variable to nullptr. THX. :)

cocos/network/HttpClient.cpp
dumganhar: Is the semantic of immediate task different from the original one?
The original semantic is to create another separate thread to process the http request.

cocos/network/HttpClient.cpp
kezhuw, line 70: In callback of `Scheduler::performFunctionInCocosThread`.

cocos/network/HttpClient.cpp
kezhuw: You are right. Fix it in 42b54ba .

About semantics of immediate task, we need to agree on what semantics does immediate task provide.
Under my opinion, the semantics of immediate task is that it take precedence over normal http task. It is an implementation detail that how this is guaranteed.

cocos/network/HttpClient.cpp
kezhuw: Initialize it explicitly in 828e122.

minggo:
@kezhuw  Thanks for the PR.
The codes changed a lot, so please send a new one if you think it is needed.

PR #8919, opened 2014-10-24
Fix Label Kerning on Single Characters Cocos2d-X3.3Beta0

ahlwong:
When applying additionalKerning to a Label that displays a string with only 1 character, the character is shifted.

error

### DIFF
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -390,15 +390,13 @@ bool LabelTextFormatter::createStringSprites(Label *theLabel)
         
         nextFontPositionX += charAdvance + kernings[i];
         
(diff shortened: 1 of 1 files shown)


WenhaiLin:
@ahlwong Thanks.

ahlwong:
Hang on there seems to be more of an issue here. I'll report back.

ahlwong:
Ok, this is now fixed for all strings. It seems that the problem was affecting all strings and it was because the additional kerning was being added to the total width of the label an extra time.

In other words, on the i'th iteration, the kerning was being applied for the (i+1) case preemptively. The total width was then calculated as total width = actualWidth(i) + kerning(i+1). Additional kerning was always being added 1 more time than necessary at the end of a line.

PR #8922, opened 2014-10-24
avoid locking when message queue is empty in websocket

TimothyZhang:


minggo:
@TimothyZhang 
Could you please replace tab with 4 spaces?

@dumganhar 
Could you please help to review it?

PR #8928, opened 2014-10-24
Add greater than operator to Vec2

Svenito:
Oddly this was missing in the V3 branch

### DIFF
--- a/cocos/math/Vec2.h
+++ b/cocos/math/Vec2.h
@@ -418,6 +418,15 @@ public:
      * @return True if this vector is less than the given vector, false otherwise.
      */
     inline bool operator<(const Vec2& v) const;
+    
+    /**
+     * Determines if this vector is greater than the given vector.
+     *
+     * @param v The vector to compare against.
+     *
+     * @return True if this vector is greater than the given vector, false otherwise.
+     */
+    inline bool operator>(const Vec2& v) const;
 
     /**
      * Determines if this vector is equal to the given vector.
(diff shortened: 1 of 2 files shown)


litefeel:
why need `operator >` on vector(Vec2)

Svenito:
Do you mean to use the less than operator with the args reversed instead?

litefeel:
what is mean?
`vectorA > vectorB`

minggo:
@lite3 
Yep, the usage is 
`vectorA > vectorB`

litefeel:
@minggo @Svenito

litefeel:
@minggo @Svenito

两个矢量比较大小没有意义，为什么要比较两个矢量的大小？

litefeel:
@minggo @Svenito
Please revert this patch.

minggo:
We already has a `<` operation.

litefeel:
why need `<` operation at vector?

we can remove them.

litefeel:
@minggo @Svenito

我不知道为什么要比较两个矢量的大小，但是我在DirectX3D中也看到有矢量的比较，但是它的逻辑跟这边的不一样。
http://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.comparison.xmvector3less(v=vs.85).aspx

```
return ( V1.x < V2.x && V1.y < V2.y &&  V1.z < V2.z );
```

并不是 x相等再比较y。

minggo:
@super626 
Please take a look.

Svenito:
I added the patch because the "greater than" was missing, but the "less than" operator was there.
I agree that it doesn't make sense to talk about vector size in terms of its components, the operators should compare the magnitude of the vectors instead.

However, should getPosition and setPosition then not use the Point2 type instead of Vec2?

litefeel:
@minggo @super626  
How to treat this patch.

PR #8938, opened 2014-10-27
add port to host string in websocket

TimothyZhang:
if the port is no default(80), the host string must contain the port. eg: www.host.com:8888


minggo:
Thanks for the PR. The codes changed, so i will close the PR.

PR #8941, opened 2014-10-27
Update ButtonReader.cpp

G17hao:
check titlefont in json file's and binary file's path.


minggo:
@chengstory
Could you please review it?

zilongshanren:
@G17hao Now the Json reader is deprecated and Flatbuffer is a better option.
If this is still a issue in `ButtonReader`, please sync with cocos2d upstream and send another PR to resolve this issue. 
Thanks.

PR #8945, opened 2014-10-28
added Get Only one from findChildren

kyokomi:


minggo:
New features are frozen because rc0 version is released.
I think we can merge it in next version.

Thanks.

kyokomi:
@minggo All right. Review Thanks.

cocos/base/ccUtils.h
```diff
@@ -74,6 +74,8 @@ namespace utils
      */
     std::vector<Node*> CC_DLL findChildren(const Node &node, const std::string &name);
     
+    Node* findChildByName(const Node &node, const std::string &name);
```
minggo: Please add CC_DLL, or it will cause problem on Windows.

PR #8949, opened 2014-10-28
do not intercept touch events if touch disabled

TimothyZhang:


zilongshanren:
Hi, @TimothyZhang 
Why should we add this guard at the beginning of `interceptTouchEvent`?
If we don't add this, does it cause any issues?
Thanks.

TimothyZhang:
Sorry I forgot to write a description.

The problem is, if we for some reason disable touch for UIScrollView, it will still respond to touch event if a descendant widiget is clicked and dragged.

cocos/ui/UIListView.cpp
```diff
@@ -504,6 +504,12 @@ void ListView::selectedItemEvent(TouchEventType event)
 void ListView::interceptTouchEvent(TouchEventType event, Widget *sender, Touch* touch)
 {
     ScrollView::interceptTouchEvent(event, sender, touch);
+
```
zilongshanren, line 507: @TimothyZhang 
Why scrollView doesn't call Layout::interceptTouch to propagate the touch event to its parent?

zilongshanren:
@TimothyZhang 
I think we shouldn't propagate the touch event to its parents when touch is disabled.

And the test logic is also strange:

```
if(!_touchEnabled && !_isInterceptTouch) 
```

The _isInterceptTouch is always true so we should remove it, we only need to determine the _touchEnabled field.

Thoughts?

zilongshanren:
it is fixed at https://github.com/cocos2d/cocos2d-x/pull/13253

PR #8954, opened 2014-10-29
modify changelog

zilongshanren:


minggo:
Items should be listed according character sequence.

PR #8957, opened 2014-10-29
XCode:4.6 to 5.1

napjon:


minggo:
Why?

napjon:
Readme.md still specify minimum requirements is xcode 4.6+(i'm using 5.0), which causing me pretty headache few days earlier, can't build any project at all, causing generic error which i can't figure out.The solution was so simple, had i seen the RELEASE_NOTES.md, which specify xcode 5.1+, which then i upgrade the xcode, i may not have going into so much trouble.

Refer to the other guy:
http://discuss.cocos2d-x.org/t/solved-cocos2d-x-3-2-ios-build-errors/16226

Issue #8933, opened 2014-10-25
[WP8] The project created by Cocos2dx-3.x cannot be launched on Windows Phone 8.1 devices


tealcwu:
Test Environment: 
Windows 8.1 Enterprise English, Visual Studio 2012 Update 4, Windows Phone 8.0 SDK
Test Device: 
HuaWei W1, Windows Phone 8.1, build 8.10.14176.243, English

Repro Steps:
1. Download Cocos2d-x v3.x (we use v3.3rc0 here) from http://cn.cocos2d-x.org/download/.
2. Unzip it when the download is completed, e.g. unzip it to folder D:\cocos2d-x-3.3rc0.
3. In Command Promote, run "D:\cocos2d-x-3.3rc0\tools\cocos2d-console\bin\cocos new -l cpp TestGame" to create a new Cocos2d-x project.
4. Open ..\TestGame\proj.wp8-xaml\TestGame.sln with Visual Studio 2012, select "Device" as target platform and build the solution.
5. When build is finished, deploy it to Windows Phone 8.1 device, run TestGame on test phone.

Expected Result: TestGame can be launched successfully.
Actual Result: Failed to launch TestGame - the app launches but closes immediately. After testing many times, I found TestGame can be launched successfully once in five to ten times. 

More Information:
1. This issue repros to test app cpp-test in ..\cocos2d-x-3.3rc0\build\cocos2d-wp8.sln.
2. This issue repros on all Cocos2d-x 3.x versions.
3. This issue doesn't repro on Cocos2d-x 2.2.5.
4. This issue doesn't repro on Windows Phone Emulators 8.0 and 8.1.
5. This issue is not tested on Windows Phone 8.0 device, because I don't have Windows Phone 8.0 device on hand.

edudemy:
+1

minggo:
@stammen
Did you know the issue?

stammen:
I will investigate

minggo:
Thanks.

stammen:
It seems there are two shaders that can crash some models of Windows phones

kShaderType_LabelDistanceFieldNormal
kShaderType_LabelDistanceFieldGlow

Please comment out lines 375 to 383 in cocos/renderer/ccglprogramcache.cpp and let me know if it solves your problem.

Also, can you check with the debugger if it is crashing on these lines?

stammen:
Did the change I suggested solve the problem?

tealcwu:
[...]

PR #8958, opened 2014-10-29
WP8 Shader fix for cocos2d-x issue 8933

stammen:
This pull request fixes cocos2d-x issue https://github.com/cocos2d/cocos2d-x/issues/8933 

It appears that the ccShader_Label_df.frag and ccShader_Label_df_glow.frag shaders will crash some older Windows Phones such as the Huawei W1, I disabled these shaders (and the use of these shaders) and now apps  launch and run fine on the older WP8 phones.

Minor compiler warnings in WP8 code were fixed.

Added use of lock free concurrent_queue to improve input event performance


cocos/platform/CCFileUtils.cpp
```diff
@@ -1146,7 +1146,7 @@ bool FileUtils::removeDirectory(const std::string& path)
 	bool ret=true;   
 	if (search!=INVALID_HANDLE_VALUE)   
 	{   
-		bool find=true;   
+		BOOL find=true;
```
minggo, line 1149: Why use `BOOL`?

stammen, line 1149: FindNextFile(search, &wfd) returns BOOL not bool

http://msdn.microsoft.com/en-us/library/windows/desktop/aa364428(v=vs.85).aspx

minggo, line 1149: Got it.

PR #8961, opened 2014-10-29
Winrt link fix

appgenmiguelferreira:
Fixed liking problems with winrt project if using extern consts inside user project

See discussion https://github.com/MSOpenTech/cocos2d-x/issues/98


stammen:
I approve of these changes. Thanks for fixing.

PR #8968, opened 2014-10-30
Update CSLoader.cpp

G17hao:
add assert for createNode.
Because this is a very common problem that programmer giving a wrong path, but Considered as the  framework's problem.


chengstory:
I ask @pipu to take a look

pipu:
If the actual parameter is an absolute path, the assert can be added for method createNode.

minggo:
@pipu 
So, can merge it or not?

pipu:
No, we can't merge it because we can't ensure the actual parameter value is absolute path.

minggo:
Got it.
Thanks.

minggo:
@G17hao 
Any update?

PR #8970, opened 2014-10-30
refact 3d bundle code and fix sprite

super626:
1. refact 3d bundle code
2. sprite create from cache code, fix the bug that it will not create the attached sprite.
3. fix obb overlap code bug.


cocos/3d/CCBundle3D.h
```diff
@@ -72,12 +66,6 @@ class CC_DLL Bundle3D
     
     /**
      * load material data from bundle
-     * @param id The ID of the material, load the first Material in the bundle if it is empty
-     */
-    virtual bool loadMaterialData(const std::string& id, MaterialData* materialdata);
-    
-    /**
-     * load material data from bundle
```
minggo, line 80: Will it break compatibility?

super626:
It is backward compatible.
It will not break the compatibility.

PR #8972, opened 2014-10-30
fixed bug of binding of class "Spawn"

SunLightJuly:


samuele3hu:
This binding wrote like this was to support v2.x.

SunLightJuly:
But in the original codes, the function "extendSpawnDeprecated()" is never be called, why? Some users maybe want to use "cc.Spawn:createWithTwoActions()".

samuele3hu:
Yes,"extendSpawnDeprecated()" was called is a bug.From 3.x, we should call "cc.Spawn:create(action1,action2)" instead of "cc.Spawn:createWithTwoActions()"

SunLightJuly:
OK，got it.

PR #8974, opened 2014-10-30
V3 patch

SunLightJuly:
fixed bug of LuaStack::executeScriptFile():  some script files can't be found to exec.
added FileUtils::setSearchRootPath():  a very useful function.

### DIFF
--- a/cocos/scripting/lua-bindings/manual/CCLuaStack.cpp
+++ b/cocos/scripting/lua-bindings/manual/CCLuaStack.cpp
@@ -274,10 +274,19 @@ int LuaStack::executeString(const char *codes)
 
 int LuaStack::executeScriptFile(const char* filename)
 {
-    std::string code("require \"");
-    code.append(filename);
-    code.append("\"");
-    return executeString(code.c_str());
+    CCAssert(filename, "CCLuaStack::executeScriptFile() - invalid filename");
+
+    std::string fullPath = FileUtils::getInstance()->fullPathForFilename(filename);
+    ssize_t chunkSize = 0;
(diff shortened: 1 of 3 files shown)


cocos/scripting/lua-bindings/manual/CCLuaStack.cpp
```diff
@@ -274,10 +274,19 @@ int LuaStack::executeString(const char *codes)
...
 {
-    std::string code("require \"");
-    code.append(filename);
-    code.append("\"");
-    return executeString(code.c_str());
+    CCAssert(filename, "CCLuaStack::executeScriptFile() - invalid filename");
+
+    std::string fullPath = FileUtils::getInstance()->fullPathForFilename(filename);
+    ssize_t chunkSize = 0;
+    unsigned char *chunk = FileUtils::getInstance()->getFileData(fullPath.c_str(), "rb", &chunkSize);
```
dumganhar: Please use `FileUtils::getInstance()->getDataFromFile();`
getFileData was deprecated.

minggo:
Where does `setSearchRootPath` been used?

cocos/scripting/lua-bindings/manual/CCLuaStack.cpp
SunLightJuly: fixed

SunLightJuly:
for example: App on phone can download new script files and resource files to writable path or SD card, reset root path of search to the download path, and run the new script files with the new resources well.

minggo:
You can do it by set search path.

PR #8977, opened 2014-10-30
Modify that parsing attribute "fontName" of TextReader

pipu:


minggo:
[ci rebuild]

joewan:
"微软雅黑".

we'd better define default font macro, so that users can replace default font conveniently.
@minggo

darkdukey:
@pipu Can you change the default font to some universal font instead of "微软雅黑"

pipu:
Okay.
I could change the default font to "Helvetica" for TextReader.
But, I gotta discuss it with designer of cocosstudio first.

minggo:
Yep, please don't use magic string as @joewan  said.

minggo:
@pipu
Any update?

pipu:
I still wait for design of designer of CocosStudio editor.

minggo:
I think there are two things:
- use macro in codes
- use another font as default font

I think you can use macro first.
Then i will merge this pull request.

pipu:
@minggo 
I think you could merge this pull request.
Because, this pull request fixes that parse ".ttf" file to set font name for Text widget of ui framework in TextReader.cpp.
And then, I think that I could use another font as default font refer to your advise at next pull request.

Thanks a lot.

PR #8979, opened 2014-10-30
eaglview release leak in CCEAGLViewImpl-ios

uorbe001:
This is leaking the eaglview when the Director is purged.

In our case, we are leaving the cocos environment on some views which
are implemented with cocoa, and because this eaglview is not being
released, it is getting notifications for keyboard events, and trying to
access stuff on the NULL eaglview, crashing our app.

### DIFF
--- a/cocos/platform/ios/CCGLViewImpl-ios.mm
+++ b/cocos/platform/ios/CCGLViewImpl-ios.mm
@@ -187,7 +187,8 @@ void GLViewImpl::end()
     CCEAGLView *eaglview = (CCEAGLView*) _eaglview;
 
     [eaglview removeFromSuperview];
-    [eaglview release];
+    //[eaglview release];
+    release();
 }
 
 
--- a/cocos/platform/ios/CCGLViewImpl-ios.mm
+++ b/cocos/platform/ios/CCGLViewImpl-ios.mm
@@ -128,7 +128,7 @@ bool GLViewImpl::initWithRect(const std::string& viewName, Rect rect, float fram
                                         sharegroup: nil
                                      multiSampling: NO
                                    numberOfSamples: 0];
-    
+
     [eaglview setMultipleTouchEnabled:YES];
 
     _screenSize.width = _designResolutionSize.width = [eaglview getWidth];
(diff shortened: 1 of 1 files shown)


cocos/platform/ios/CCGLViewImpl-ios.mm
```diff
@@ -182,12 +182,12 @@ of this software and associated documentation files (the "Software"), to deal
...
 {
     [CCDirectorCaller destroy];
-    
+
     // destroy EAGLView
     CCEAGLView *eaglview = (CCEAGLView*) _eaglview;
 
     [eaglview removeFromSuperview];
-    //[eaglview release];
+    [eaglview release];
```
miguel12345: eaglview is never retained in this file so it is not CCGLViewImpl-ios's job to release it

uorbe001: whose job would that be then?

miguel12345: CCEAGLView is autoreleased so no one should retain nor release it

miguel12345: Add

```
release();
```

to 

```
void GLViewImpl::end()
```

The problem is that CCDirector is never releasing its GLView

uorbe001: @miguel12345 you convinced me, I've updated it with the release you mentioned.

minggo:
@uorbe001 
How to reproduce it?

uorbe001:
@minggo replicating our crash is not very complicated, you just need to create a `CCEAGLView`on ios, add it to the parent view and then switch to another cocoa `UIViewController`.  If the `CCEAGLView` is not released, tapping the screen crashes the application because it is trying to access some attributes on the [`CCEAGLView`](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/ios/CCEAGLView-ios.mm#L777) (which is null on the CCDirector, because it has already been [purged](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/ios/CCEAGLView-ios.mm#L777)). 

This doesn't fix that crash though, I had to fix our crash by changing our `RootViewController`, but this is very different from the ones on the cocos samples, as we are mixing cocoa and cocos stuff. This does however fix the leak of the `GLView` on ios, as the director is [retaining](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/CCDirector.cpp#L375) it when it is set, but isn't releasing it when it is purged.

PR #8981, opened 2014-10-30
Cleanup order of fields initialization.

vovkasm:
Make initialization order to match declarations.
This change twice reduces number of warnings when compiling on MacOS X with Xcode 6.1
Warnings was like "Field 'XXXX' will be initialized after field 'YYYY'"


vovkasm:
About failed build. It seems that buildbots on linux and win32 failed itself...

minggo:
[ci rebuild]

vovkasm:
Thank you!

PR #8993, opened 2014-11-03
add aabb to 3d model file and better support to multi-animations

super626:
Given an animation ID access the animation by offset instead of loading the animation one by one


minggo:
[ci rebuild]

minggo:
It is a new feature, right?
If so, i think we should not merge it.
Because we have released v3.3rc0.

super626:
@minggo 
It is not a new feature.
However, we modified the c3b/c3t file. Add animation ref table and aabb to it to improve performance.
In the past, we read all the animations to find specific animation and calculate the aabb after loading the 3d model. In fact, the aabb can be calculated offline in the fbx-conv.
This is all the information. You can decide merge it or not.

PR #8997, opened 2014-11-03
fix Cocos2dxSound.java  thread safety problem and unable stop some sounds  bug in coco2dX 2.x

zhong317:
call "preloadEffect" will release the mSemaphore also, so if you call "preloadEffect" first, and then call "playEffect" to play some unloaded sound, the mSemaphore's acquire will return immediately,  and gain a wrong streamID, so you can't stop the sound later. also, it will be possible to  cause the "ConcurrentModificationException".


minggo:
There is not `mSemaphore` any more, the PR is outdated. So i will close it. Thanks for the PR.

PR #9005, opened 2014-11-04
Update CCFileUtils-linux.cpp

### LINKED ISSUE #8939: hard coded _defaultResRootPath in FileUtils
The `_defaultResRootPath` in FileUtils is hard coded , for example on Linux it is "/xxx/xxx/xxx/Resources/" :https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/linux/CCFileUtils-linux.cpp#L74

But we use a different name of resource folder, we use `res`, and seems no way to configure it. This is a bug?

zhaijialong:
fixed #8939 
this is needed for we use a different resource folder in cocos2d-js https://github.com/cocos2d/cocos2d-js/pull/1090

### DIFF
--- a/cocos/platform/linux/CCFileUtils-linux.cpp
+++ b/cocos/platform/linux/CCFileUtils-linux.cpp
@@ -36,6 +36,10 @@ THE SOFTWARE.
 #include <stdio.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


zhaijialong:
[ci rebuild]

PR #9022, opened 2014-11-05
Download jobs should all follow http redirect

hongliang-goudou:
`groupBatchDownload()` has this curl option on already, so as do`downloadToBuffer()` and `download()`. It's important to support a redirect download sometimes.


jgod:
Good idea :+1:

minggo:
close this PR since it is outdated

PR #9026, opened 2014-11-05
Fixed compile error with CC_ENABLE_SCRIPT_BINDING=0

cpascal:
There are too much compile error if I set EE_ENABLE_SCRIPT_BINDING=0.

CCEventTouch, CCTouch, std::list related things...
This is temporary fix... Many source/header files must be patched for the right way..


minggo:
Thanks.

@samuele3hu 
Please fix other errors caused by the macro.

minggo:
@samuele3hu 
I assigned to Wenhai.

PR #9028, opened 2014-11-06
Fix widget flip issue

zilongshanren:
### Problem description

When setting Widget or Widget descendant class to flip, then get one of its inner renderer, change the scale of the renderer, it will behavior wrong.
### After apply this patch.

The problems described above is fixed.


darkdukey:
[ci rebuild]

PR #9034, opened 2014-11-06
Ron

RonTang:
performTaskInCocosThread and performFunctionInCocosThread difference is that we can use performTaskInCocosThread return-value to determine , whether submitted to CocosThread function is finished, and get submitted to CocosThread function's return-value .
When we need to perform some operations in other threads , these thread need wait for cocos thread update over, this method provides convenience.
why our function return 'bool',because of Microsoft bug https://connect.microsoft.com/VisualStudio/feedback/details/791185/std-packaged-task-t-where-t-is-void-or-a-reference-class-are-not-movable,at current,we must need function returning a value like 'bool'.
It is ok in ubuntu 12.04 and window 7,other platform not test.

performTaskInCocosThread与performFunctionInCocosThread区别在于，我们可以通过performTaskInCocosThread返回值确定，提交给CocosThread的函数是否执行完毕，并获取提交给CocosThread函数的返回值。为什么提交函数要返回bool，因为Microsoft的bug，https://connect.microsoft.com/VisualStudio/feedback/details/791185/std-packaged-task-t-where-t-is-void-or-a-reference-class-are-not-movable。暂时我们需要返回值
当我们需要在其他线程执行一些操作，并且要等待cocosThread更新完成时，此方法提供便利。在ubuntu和window下测试正常，其他平台未测试。

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -261,6 +261,7 @@ Scheduler::Scheduler(void)
 {
     // I don't expect to have more than 30 functions to all per frame
     _functionsToPerform.reserve(30);
+    _tasksToPerform.reserve(30);
 }
 
 Scheduler::~Scheduler(void)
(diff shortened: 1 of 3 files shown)


RonTang:
this function  are not compatible with lua.

RonTang:
和自动生成的lua方法不兼容

samuele3hu:
@RonTang,you should skip binding this function by modify tools/tolua/cocos2dx.ini, as follows:

Add `performTaskInCocosThread` at the end of `
Scheduler::[pause resume unschedule schedule update isTargetPaused isScheduled performFunctionInCocosThread]`

RonTang:
Thank you very much！I will try again.This is my first time to contribute my code. Thank you for your guidance！~谢谢@samuele3hu

minggo:
i am not sure if we need this new function, but i don't find any other developer ask for this feature, so close this PR

PR #9037, opened 2014-11-06
Update CCBone.cpp

G17hao:
fix the bone disorderly flying
http://www.cocoachina.com/bbs/read.php?tid-220321-keyword-%C2%D2%B7%C9.html


2youyou2:
Thanks, it works well.

minggo:
Thanks.

PR #9039, opened 2014-11-06
#9061: Polish AssetsManagerEx, bug fixes and feature added

pandamicro:


minggo:
We will not add new features since we have released rc0 version.
Could you please separate bug fixing and new feature into different PR?
Thanks.

pandamicro:
Done, this pull request contains only bug fixes.

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -138,22 +138,40 @@ void AssetsManagerEx::prepareLocalManifest()
...
-    // Prefer to use the cached manifest file, if not found use user configured manifest file
+    Manifest *cachedManifest = nullptr;
+    // Find the cached manifest file
     if (_fileUtils->isFileExist(_cacheManifestPath))
     {
-        _localManifest->parse(_cacheManifestPath);
-        if (_localManifest->isLoaded())
-            prepareLocalManifest();
-        else
+        cachedManifest = new Manifest();
```
minggo: cachedManifest = new Manifest() -> cachedManifest = new (std::nothrow) Manifest()

And should determine the return value of cachedManifest because it may be null.

pandamicro: I'm not sure how to solve when Manifest failed to be constructed, because, first of all, the constructor is very simple, secondly, the logic rely on the creation of empty Manifest objects, if they don't exist, nothing will be succeed

Can you give me some example ?

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -138,22 +138,40 @@ void AssetsManagerEx::prepareLocalManifest()
...
+    Manifest *cachedManifest = nullptr;
+    // Find the cached manifest file
     if (_fileUtils->isFileExist(_cacheManifestPath))
     {
-        _localManifest->parse(_cacheManifestPath);
-        if (_localManifest->isLoaded())
-            prepareLocalManifest();
-        else
+        cachedManifest = new (std::nothrow) Manifest();
+        cachedManifest->parse(_cacheManifestPath);
```
minggo: Should determine the value of `cacheManifest`.

minggo:
[ci rebuild]

gyf19:
热更新实战的问题：
1. 下载zip包，有没有断点续传？
2. zip包压缩失败，会不会重新下载？
3. 重新安装完整包，会不会把热更新目录资源删除？
4.如何设置热更新支持最小版本？
5.如何多个低版本都能更新到最新版本？
新版本可以支持吗？

pandamicro:
@gyf19
1. Zip file download resuming is supported if the server support
2. All failed file won't be saved, and will be re-downloaded
3. We won't delete downloaded assets while a new package is downloaded
4. 5. I'm not sure what you mean

gyf19:
1. 设置热更新支持最小版本。 
   例如：用户手机App版本是1.4.0 ，这时候我们发布2.0版本，1.4.0 版本到2.0版本不能通过热更新，需要安装完整包。  
   是否增加一个热更新的最小版本号，如果比最小版本号要小的话，直接提示安装完整包。

gyf19:
1. 如何多个低版本都能更新到最新版本。  例如： A用户手机App版本是1.1.0，B用户手机App版本是1.2.0，  现在发布最新版本是1.3.0。  A用户怎么如何更新到1.3.0？

gyf19:
1. 重新安装完整包，会不会把热更新目录资源删除？
   我觉得重新安装完整包，是需要把热更新目录资源删除的。 1.App会使用老资源 ;  2. 浪费用户手机空间

pandamicro:
@gyf19 这个PR中已经做到比较包内与本地更新的版本号，使用最新的本地版本来和服务器比较，也就是说，如果包内版本是2.0，本地缓存版本是1.4，那么会使用2.0版本和服务器比较
你说的最小版本号校验确实也是一个需求，建议你在这个帖子中回复：http://discuss.cocos2d-x.org/t/assetsmanagerex-bugs-and-new-features-requests/18211
目前可以在自己的脚本中校验

pandamicro:
关于多个低版本更新到最新，已经支持，因为我们是差量更新

gyf19:
官方是否写一个文档说一下： 关于多个低版本更新到最新的配置

gyf19:
如果包内版本是2.0 大于 本地缓存版本是1.4 的话，是否会删除本地缓存资源？ 好像看代码会删除
if (strcmp(_localManifest->getVersion().c_str(), cachedManifest->getVersion().c_str()))
{
_fileUtils->removeDirectory(_storagePath);
CC_SAFE_RELEASE(cachedManifest);
}

pandamicro:
包内版本更大的话会删除的

不同低版本更新到同一个新版本的配置是一样的，没有区别，我们会比较本地版本和服务端版本产生一个动态的差量更新列表

gyf19:
谢谢 @pandamicro 大神解答！！

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -138,22 +170,42 @@ void AssetsManagerEx::prepareLocalManifest()
...
+    // Load local manifest in app package
+    _localManifest->parse(_manifestUrl);
+    if (_localManifest->isLoaded())
     {
-        _localManifest->parse(_manifestUrl);
-        if (_localManifest->isLoaded())
-            prepareLocalManifest();
+        // Compare with cached manifest to determine which one to use
+        if (cachedManifest) {
+            if (strcmp(_localManifest->getVersion().c_str(), cachedManifest->getVersion().c_str()) > 0)
```
wasedaigo, line 195: What is the expected version format?
With this implementation following format won't work as expected:
strcmp("101", "99") = -8

pandamicro, line 195: Oh, we are imaging the normal version is something like: x.x.x
But you are right, 10 and 9 can lead to mistaken compare result too. I will create an issue

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -138,22 +170,42 @@ void AssetsManagerEx::prepareLocalManifest()
...
+            {
+                // Recreate storage, to empty the content
+                _fileUtils->removeDirectory(_storagePath);
+                _fileUtils->createDirectory(_storagePath);
```
wasedaigo, line 199: I set my storage path to "./" With the combination of what I just commented above, it ended up removing entire userdata after manifest update. (update from 1.0.9 to 1.0.10)

Ideally this should check the diff between localManifest and cachedManifest to selectively remove unnecessary assets, what do you think?

pandamicro, line 199: In fact, some user needs to avoid newer version from being downgraded. They may published their game's 2.0 version in the App Store.
The remote manifest version may still be 1.x, because they don't have a 2.1 yet. Then the hot update will cause issues.

Though I'm not sure whether I can find a reliable way to compare versions of whichever format.

edwardzhou, line 199: I recommend that AssetManager(Ex?) implement a default compare with common version format x.y.z, and also provide a callback interface (like std::function) to allow user implement their own comparision.
so the pseudo flow will like
bool needToUpdate = false;
if (this->_versionCompareCallback != null_ptr) {
    needToUpdate = this->_versionCompareCallback(this, remoteVersion, localVersion);
} else {
    needToUpdate = this->defaultVersionCompare(remoteVersion, localVersion);
}
.....

for most simple cases, the default version compare is good enough. for complicated case, user can provide their own comparing.

as well, if it is a good option you think, please also bind the callback to LUA :)

wasedaigo, line 199: For the default behaviour, could you please avoid using strcmp?
If you compare 1.1.10 and 1.1.9, the former should be larger, right?

Maybe something like:
If we have version x.y.z and x'.y'.z'
int newVersion = x \* 1000000 + y \* 1000 + z
int oldVersion = x' \* 1000000 + y' \* 1000 + z'
if (newVersion > oldVersion) {
// This is certainly newer version
}

wasedaigo, line 199: Btw, personally I do not think it makes much sense to use x.x.x as the format for version number of manifest update. It is hard to define minor version and major version for asset updates. Anyways, make it overwritable is a great idea, just do not forget to support JS-binding as well! :)

PR #9050, opened 2014-11-07
fix sprite frame cache, and remove unnecessary headers dependencies.

joewan:
we should add private headers for ccs, do not expose privae headers to users.

### DIFF
--- a/cocos/editor-support/cocostudio/CCSpriteFrameCacheHelper.h
+++ b/cocos/editor-support/cocostudio/CCSpriteFrameCacheHelper.h
@@ -28,6 +28,11 @@ THE SOFTWARE.
 #include "cocostudio/CCArmatureDefine.h"
 #include "cocostudio/CocosStudioExport.h"
 #include <string>
+#include <vector>
+
+namespace cocos2d {
+    class SpriteFrame;
+}
 
 namespace cocostudio {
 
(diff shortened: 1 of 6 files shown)


joewan:
at present, it will cause memory leak. I am working for it, pls do not merge.

joewan:
11.07
memory leak:If I create two armatures, and then delete them, unload the resources, we can't release the texture any more.

11.18
the message of 11.07 is wrong, now there is no above problems.

joewan:
@minggo @chengstory please help review this pr.
1.when we load and create armature, both keep the reference to the used texture and sprite frames.

step1:load and create an armature;
step2:director purge cache data
step3:create an armature again, we can not show this armature because the sprite frame has been purged.

2.fix memory leak
3.remove unnecessary headers dependencies.

minggo:
@chengstory 
Please take a look.

chengstory:
@2youyouo2 Please take a look

cocos/editor-support/cocostudio/CCSpriteFrameCacheHelper.h
```diff
@@ -50,11 +56,16 @@ class CC_STUDIO_DLL SpriteFrameCacheHelper
      *	@brief	Add sprite frame to CCSpriteFrameCache, it will save display name and it's relative image name
      */
     void addSpriteFrameFromFile(const std::string& plistPath, const std::string& imagePath);
+    void removeSpriteFrameFromFile(const std::string& plistPath);
 
 private:
+    void _retainSpriteFrames(const std::string& plistPath);
```
2youyou2: Thanks for your work!
But function should not be named with "_" even it's a private function.
You can find the cocos coding style here: 
https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#function-parameter-ordering

joewan:
@2youyouo2 pls review again, thanks.

2youyou2:
I think it's ok, thanks.

PR #9059, opened 2014-11-07
can't build IPA using developer certificate in Mac OS X 10.10

cocoscodeide:
fix this error:Warning: usage of --preserve-metadata with option "resource-rules" (deprecated in Mac OS X >= 10.10)!

I have **only** tested this modification on 10.10, please make sure it works fine on older versions


minggo:
I don't have the issue.

PR #9066, opened 2014-11-08
Fixed "warning: bundle reader out of range"

cpascal:
This warning was printed even if it's valid c3b file..
Changed "validLength <= needLength" to "validLength < needLength".


minggo:
@super626 
Could you please take a look?

super626:
@cpascal 
Thanks

@minggo 
cpascal is right. It is our fault.
You can merge it.

Issue #9067, opened 2014-11-09
bug cocosstudio2.2 fill background color to none


aeonmine:
set color to type none is not working

LayoutReader
ListViewReader
PageViewReader
ScrollViewReader

chengstory:
I ask @pipu to take a look.

pipu:
We've fixed this bug.
Wait for next version of CocosStudio editor, please.
Thanks a lot.

PR #9068, opened 2014-11-09
bug cocosstudio2.2 fill background color to none #9067

aeonmine:
the default value of  background color should be 0, when set fill is none in cocosstudio 2.2


minggo:
@chengstory 
Please take a look.

minggo:

PR #9073, opened 2014-11-10
fix lua http download binary data's bug

dantezhu:
官方代码里面，将二进制数据当作字符串处理了，使用strcpy函数，这个会导致遇到0截断，而之后的数据都是错乱的。

### DIFF
--- a/cocos/scripting/lua-bindings/manual/network/lua_xml_http_request.cpp
+++ b/cocos/scripting/lua-bindings/manual/network/lua_xml_http_request.cpp
@@ -244,16 +244,13 @@ void LuaMinXmlHttpRequest::_sendRequest()
         
         /** get the response data **/
         std::vector<char> *buffer = response->getResponseData();
-        char* concatenated = (char*) malloc(buffer->size() + 1);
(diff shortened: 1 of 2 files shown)


samuele3hu:
@dantezhu ,Thanks. This is a bug, I have been verified it.it is right.And,I think 

```
        char* concatHeader = (char*) malloc(headers->size() + 1);
        std::string header(headers->begin(), headers->end());
        strcpy(concatHeader, header.c_str());

        std::istringstream stream(concatHeader);
        std::string line;
        while(std::getline(stream, line)) {
            _gotHeader(line);
        }
```

also can be modified to 

```
std::string header(headers->begin(), headers->end());

std::istringstream stream(header);
std::string line;
while(std::getline(stream, line)) {
       _gotHeader(line);
}
```

samuele3hu:
I think  `lua_pushstring` of `lua_get_XMLHttpRequest_response` and `lua_get_XMLHttpRequest_responseText` should be replaced with `lua_pushlstring`.

minggo:
@dantezhu 
@samuele3hu refactor the PR in #9149.

PR #9074, opened 2014-11-10
set default priority to avoid warning

codetypess:
when call 'scheduleUpdate' and 'unscheuleUpdate' several times in one frame, it will get warning "warning: you CANNOT change update priority in scheduled function"


cocos/base/CCScheduler.cpp
```diff
@@ -447,6 +447,7 @@ void Scheduler::appendIn(_listEntry **list, const ccSchedulerFunc& callback, voi
     listElement->callback = callback;
     listElement->target = target;
     listElement->paused = paused;
+    listElemetn->priority = 0;
```
minggo: listElemetn -> listElement

PR #9080, opened 2014-11-11
Cmake refactorings and fixes

vovkasm:
This should fix building cocos2d-x on Mac, next I plan Windows, and then big refactoring to state when other cmake projects and applications can find cocos2d libs from their CMakeLists.txt files.
I do it because it seems that work on cmake framework somewhere stalled [forum](http://discuss.cocos2d-x.org/t/cmake-for-mac-qtcreator-and-clion/16889/57).

This patch includes these changes:
- Fix build tests for Mac (for this connect to build into cocos2d some missing parts: UIEditBox, protobuf, audio)
- Convert test applications for Mac to bundles to easy start and properly find Resources folders
- Move `cmake/BuildHelpers.CMakeLists.txt` to `cmake/Modules/CocosBuildHelpers.cmake` so it can be easy included from another projects in future.

I test build and samples on Mac (with ninja and Xcode generators) and Linux (with make generator). Next I plan to fix Windows MSVC generated projects.


minggo:
Thanks.

jwriteclub:
I've noticed that one needs to download the chipmunk sources to get a successful build using the current cmake files. Perhaps it would be worth failing with a note about where to get the chipmunk source?

vovkasm:
@jwriteclub, you do not need chipmunk sources on supported platforms. Prebuilt version in external/chipmunk (But yes, you need to run download-deps.py to obtain external content).

vovkasm:
@minggo, thanks for feedback ;-) But I don't understand. Is it will be merged or need more discussion, or may be some more work?
I shared some thoughts on the [forum](http://discuss.cocos2d-x.org/t/cmake-for-mac-qtcreator-and-clion/16889/59?u=vovkasm) about the future direction of my work. But I would like to move smoothly and steadily, so waiting for this request will be accepted.

martell:
@vovkasm I am not a member of the cocos2d-x team so I can't merge it,
but I am doing most of the cmake work and it looks good to me.
I have to rebase my work on cmake if it is merged but you have done some very good work here that fixes a few important things ;)

PR #9082, opened 2014-11-11
- Optimization in CCClippingNode.

Teivaz:
Disable visiting when no objects has to be drawn under stencil.

### DIFF
--- a/cocos/2d/CCClippingNode.h
+++ b/cocos/2d/CCClippingNode.h
@@ -64,7 +64,7 @@ public:
     then this method should return true every time you wish stencil to be visited.
     By default returns true if has any children attached.
     */
-    virtual bool hasContent();
+    virtual bool hasContent() const;
 
     /** The alpha threshold.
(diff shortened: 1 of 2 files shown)


Teivaz:
Doesn't seem the build failure is caused by my changes.

cocos/2d/CCClippingNode.h
```diff
@@ -58,7 +58,14 @@ class CC_DLL ClippingNode : public Node
...
     Node* getStencil() const;
     void setStencil(Node *stencil);
-    
+
+    /** If stencil has no childre it will not be drawn.
+    If you have custom stencil-based node with stencil drawing mechanics other then children-based,
+    then this method should return true every time you wish stencil to be visited.
+    By default returns true if has any children attached.
+    */
+    virtual bool hasContent();
```
minggo: virtual bool hasContent() -> virtual bool hasContent() const

Teivaz:
@minggo I've thought about this. Here's updated code.

minggo:
Thanks.

PR #9085, opened 2014-11-11
fix bug - potential memory leak.

jicheol-ryu:
const std::string text = std::string((char*)ch,0,len); is using next
string constructor.
string (const string& str, size_t pos, size_t len = npos);
memory leak can occur while type converting char to string.

So we have to use next string constructor.
string (const char\* s, size_t n);

http://www.cplusplus.com/reference/string/string/string/


ricardoquesada:
thanks.

PR #9087, opened 2014-11-12
Fix Z-Order Bug SpriteBatchNode

ahlwong:
Fixes Z-ORder bug for SpriteBatchNode. Previously, setting the zOrder on childs of SpriteBatchNode would not reorder the children. The reason is that Node::setLocalZOrder would set _localZOrder = z before the parent would check zOrder == child->getLocalZOrder() causing the reorderChild methods to return without reordering the children.

Please refer to https://github.com/cocos2d/cocos2d-swift/issues/598 for additional information.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -286,10 +286,13 @@ void Node::setLocalZOrder(int z)
     if (_localZOrder == z)
         return;
     
-    _localZOrder = z;
     if (_parent)
     {
         _parent->reorderChild(this, z);
+    }
+    else
+    {
+      _localZOrder = z;
     }
 
     _eventDispatcher->setDirtyForNode(this);
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -1010,6 +1010,8 @@ void Sprite::setBatchNode(SpriteBatchNode *spriteBatchNode)
         // using batch
         _transformToBatch = Mat4::IDENTITY;
         setTextureAtlas(_batchNode->getTextureAtlas()); // weak ref
+        
+        _reorderChildDirty = true; // force reorder
     }
 }
 


ahlwong:
Added another commit to fix the case when Sprite add children with custom zOrder and then Sprite is added to SpriteBatchNode before children are sorted.

void Sprite::sortAllChildren() only sorts its immediate children. Grandchildren are sorted in their own visit() calls. The problem happens when you create a Sprite, add children to it with custom zOrders, and then add it to a SpriteBatchNode. The grandchildren are never sorted. This commit sorts all the grandchildren when initially added to the SpriteBatchNode to fix this issue.

ricardoquesada:
thanks. Could you also a test case that reproduces the bug ?
thanks.

zilongshanren:
[ci rebuild]

ahlwong:
@ricardoquesada What is the best way to add a test case? I'll post it here below, but if you give me further instructions I can try to create another commit. Thanks.

To reproduce the bug:

Sprite\* containerSprite = Sprite::create();

Sprite\* childSprite1 = Sprite::create();
childSprite1->setLocalZOrder(99);
containerSprite->addChild(childSprite1); // childSprite1 added out of order

Sprite\* childSprite2 = Sprite::create();
containerSprite->addChild(childSprite2); // childSprite2 will eventually have to be sorted below childSprite1

// Sprite::Node::visit is never called because containerSprite is never added to the scene, and so sortAllChildren() is never called

SpriteBatchNode\* spriteBatchNode = SpriteBatchNode::create(); // using same texture as previous sprites
spriteBatchNode->addChild(containerSprite); // childSprite1 and childSprite2 still out of order and stay that way
scene->addChild(spriteBatchNode);

minggo:
@ahlwong 
You can add a test case in `cpp-tests`.

minggo:
close this PR since it is unresponable

PR #9089, opened 2014-11-12
Update MathUtilNeon64.inl

wangqm0513:
add arm64bit  neon support

### DIFF
--- a/cocos/math/MathUtilNeon64.inl
+++ b/cocos/math/MathUtilNeon64.inl
@@ -46,173 +46,209 @@ public:
 
 inline void MathUtilNeon64::addMatrix(const float* m, float scalar, float* dst)
 {
-    dst[0]  = m[0]  + scalar;
-    dst[1]  = m[1]  + scalar;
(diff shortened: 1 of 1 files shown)


ricardoquesada:
thanks. unfortunately the build failed with this patch.

anyway, do you think it would be better to migrate to SIMD intrinsics ? thanks.

ricardoquesada:
[rebuild]

ricardoquesada:
[ci rebuild]

dumganhar:
@ricardoquesada, we got a juno board, which is the first arm64 android device, from ARM.
I will check this improvement. 
 @wangqm0513 Thanks for your contribution. :)

dumganhar:
[ci rebuild]

ricardoquesada:
thanks. looks good.
For future patches, perhaps might want to add support for NEON intrinsics. I think, if we do that, we will get ARM64 Neon for free (but I could be wrong).

http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0205j/BABGHIFH.html

PR #9115, opened 2014-11-14
Fix compile errors with recent mingw.

vovkasm:
Constify arguments to copy constuctors. Without this gcc fail to compile if classes used inside containers.
Error message is rather cryptic: 

```
error: 'constexpr std::pair<_T1, _T2>::pair(const std::pair<_T1, _T2>&) [with _T1 = const int; _T2 = cocos2d::experimental::AudioPlayer]' declared to take const reference, but implicit declaration would take non-const
```

I test this:
1. With mingw by successfully create package from MSYS package source [PKGBUILD](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-cocos2dx-git/PKGBUILD)
2. With recent Microsoft Visual Studio Community 2013 (build success, test applications runnable)

### DIFF
--- a/cocos/audio/win32/AudioCache.cpp
+++ b/cocos/audio/win32/AudioCache.cpp
@@ -51,7 +51,7 @@ AudioCache::AudioCache()
     
 }
 
-AudioCache::AudioCache(AudioCache& cache)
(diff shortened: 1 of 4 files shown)


ricardoquesada:
thanks.

Issue #9117, opened 2014-11-14
UIWebView doesn't initialize properly (fails to load first time) when using loadHTMLString


jgod:
**Expected behavior**: Webview loads page just like it does with loadURL and loadFile.
**Actual behavior**: self.uiWebView isn't initialized until the second loadHTMLString request, the first request brings up a blank page and a NSURLErrorDomain -999 error if it even fires at all.

How to reproduce:

```
// Example asset locations
const std::string TEMPLATE_LOCATION = "assets/www/description.html";
baseUrl = FileUtils::getInstance()->fullPathForFilename(TEMPLATE_LOCATION);
// Form the baseURL properly by using directory one level up, and replacing spaces with %20.
myUtils::replaceInPlace(baseUrl, "/description.html", "");
myUtils::replaceInPlace(baseUrl, " ", "%20");
baseUrl += "/";

// Put valid html into html template (string)...

// Relevant part
webView = experimental::ui::WebView::create();
webView->loadHTMLString(htmlTemplate, baseUrl);
myLayer->addChild(webView, LayerOrder::UI);
```

Here's how to fix:
[`UIWebViewImpl-ios.mm` line 142](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/ui/UIWebViewImpl-ios.mm#L142-L144):

```
- (void)loadHTMLString:(const std::string &)string baseURL:(const std::string &)baseURL {
+   if (!self.uiWebView) {[self setupWebView];}
    [self.uiWebView loadHTMLString:@(string.c_str()) baseURL:[NSURL URLWithString:@(baseURL.c_str())]];
}
```

PS: **UIWebViewImpl::loadData() probably needs this too, haven't used that method yet.**

walzer:
Fixed in #10909 and #10200

PR #9118, opened 2014-11-14
Fix UIWebViewImpl::loadHTMLString()

jgod:
As described [here](https://github.com/cocos2d/cocos2d-x/issues/9117).


heliclei:
[ci rebuild]

jgod:
[ci rebuild]

minggo:
Thanks.

jgod:
@minggo, what kind of test would you like? 
If there's any already existing test code for this class it should be transparent since it's just the implementation that changed (functionally the same).
This is my first pr in this repo so still figuring things out.

minggo:
Hi @jgod ,
Just add a test case that can reproduce this issue.
There is already a test case in `tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIWebViewTest/UIWebViewTest.cpp`, you can add a test case in it.

jgod:
@minggo, Ah I think I understand.

Create a test case as part of the PR that shows how it is broken without this fix? Like a unit test almost

minggo:
Yep.
Thanks.

jgod:
I added a test. It kind of covers [https://github.com/cocos2d/cocos2d-x/pull/9157] too because they both are dealing with the same thing, just different implementation. Let me know what you think

zilongshanren:
[ci rebuild]

minggo:
Fixed in #10909

PR #9121, opened 2014-11-14
This should fix and unify finding GLFW3 for all desktop archs.

ricardoquesada:
Instead of using glfw3 on each required configuration, use it centralized way.
Add our prebuilt location to search paths, so FindGLFW3.cmake can find it also.


ricardoquesada:
it works ok. thanks!

zilongshanren:
@ricardoquesada 
It break the Linux compile bot. Please please ... don't merge PR when CI is not pass.
Thanks.

PR #9131, opened 2014-11-16
Update CCTransformHelp.cpp

halx99:
fix bug: nodeToMatrix(const BaseData&, Mat4&)

### DIFF
--- a/cocos/editor-support/cocostudio/CCTransformHelp.cpp
+++ b/cocos/editor-support/cocostudio/CCTransformHelp.cpp
@@ -124,8 +124,8 @@ void TransformHelp::nodeToMatrix(const BaseData &node, Mat4 &matrix)
 
     if (node.skewX == -node.skewY)
     {
-        double sine   = sin(node.skewX);
-        double cosine = cos(node.skewX);
+        double sine   = sin(CC_DEGREES_TO_RADIANS(node.skewX));
+        double cosine = cos(CC_DEGREES_TO_RADIANS(node.skewX));
 
         matrix.m[0] = node.scaleX * cosine;
         matrix.m[1] = node.scaleX * -sine;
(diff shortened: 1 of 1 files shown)


halx99:
sin,cos,etc accept radius，but degrees gives.

minggo:
@chengstory 
Please take a look.

zilongshanren:
[ci rebuild]

dabingnn:
@xseekerj The newest codes of `nodeToMatrix` and `matrixToNode` are a pair function.
It is radian, I think you can convert you data to radian before using the function.
Thanks.

halx99:
@dabingnn I'm not call the function directly; that's the truth:when I rotate a armture throgth touch enent,the behavior isn't correctly.

dabingnn:
I think there is no bug in `nodeToMatrix`. If the behaviour of `Armature` is  not correct, there may be some bugs in armature. Could you please give a description of the problem? And if there is a test case to reproduce the bug, it would be more clearer.

halx99:
In Version:3.2: This function is called at CCBone.cpp:219, and _skewX,_skewY,_rotationZ_X,_rotationZ_Y are used directly which are the members of cocos2d::Node. 

I  realized follow diff: 
version 3.2: 
_worldInfo->skewX = _worldInfo->skewX + _skewX + rotationZ_X; 
_worldInfo->skewY = _worldInfo->skewY + _skewY - rotationZ_Y; 
newest version:
_worldInfo->skewX = _worldInfo->skewX + _skewX + CC_DEGREES_TO_RADIANS(_rotationZ_X); 
_worldInfo->skewY = _worldInfo->skewY + _skewY - CC_DEGREES_TO_RADIANS(_rotationZ_Y);

halx99:
So I think convert the data to radian in the pair function maybe a better choice.

minggo:

PR #9139, opened 2014-11-17
Unignore libs dir and fix Travis script to descend into included directories. (follows #9108)

nilium:
Both commits follow #9108, which un-ignored jar files but unfortunately only
worked locally due to these changes, so here's a couple additional patches
to make the "gitingore" file not ignore the libs directories in Android plugins
and to get the Travis template file list script to actually record the contents
of included directories, since it previously would not (it'd record the directory
but ignore the contents, so you'd just end up with empty stub directories,
which isn't very useful).

Sorry that the first pull request didn't nail this immediately, but it was at least
a building block.
#### 4102746 Descend into included directories to check for file inclusion.

Only applies to directories that are marked as unignored, since the
travis script has the odd problem of not descending into those
directories and, as a result, not recording the files in them.

I'm assuming the failure to descend into included directories is
unintended.
#### 590975d Also un-ignore libs dir under plugins.

The configure script isn't smart enough to check if there are
un-ignored files in a directory, so this fixes that.

### DIFF
--- a/tools/travis-scripts/config.gitingore
+++ b/tools/travis-scripts/config.gitingore
@@ -115,6 +115,7 @@ tags
 !/cocos/2d/platform/android/java/res/
 # Permit plugins to ship third-party libraries.
 !/plugin/plugins/*/proj.android/libs/*.jar
+!/plugin/plugins/*/proj.android/libs
 
 v*-deps-*.zip
 v*-lua-runtime-*.zip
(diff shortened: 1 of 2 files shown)


nilium:
Build not passing on CentOS doesn't look like it's related to this particular change (could be wrong, obviously). Anyone see anything that suggests otherwise?

darkdukey:
@dumganhar can you take a look at this one? It's related to the plugins
@nilium I think we're using jenkins now, since we have more control over it.

dumganhar:
> Build not passing on CentOS doesn't look like it's related to this particular change (could be wrong, obviously).

Yep, the build failure is not triggered by your changes. We will check it.
@darkdukey , this patch looks good. I'll merge it. :)

dumganhar:
@andyque, could you look at the jenkins problem?

dumganhar:
Merged. Thanks.

Issue #9155, opened 2014-11-18
Android UIWebView loadHTMLString uses baseURL as Mimetype


irisdominguez:
I'm working with the UIWebView on both android and iOS, providing some dynamic HTML code with the loadHTMLString method. I was having problems with the Android implementation, as the HTML code was being rendered as plain text.

I just realized that on the Android implementation, the baseURL argument falls to the mimetype of the text on the actual WebView. Thus, something like:

<pre>webView->loadHTMLString("<html><body style=\"background-color:#00FF00;\">HELLO WORLD!</body></html>", "text/html");
</pre>

is the only way to make the code work on both Android and iOS, while the standard way:

<pre>webView->loadHTMLString("<html><body style=\"background-color:#00FF00;\">HELLO WORLD!</body></html>", "");
</pre>

will only work on iOS. On Android, all the code will be rendered as plain text.

For my problem, I don't need the baseURL (I use HTML only for formatting text), and this workaround is enough, but I think this could be a problem for someone in the future, if not fixed. Nothing urgent, just wanted to let you know.

The real solution, I think, would be changing the loadHTMLStringJNI method of UIWebViewImpl-android.cpp , and making:

<pre>t.env->CallStaticVoidMethod(t.classID, t.methodID, index, jString, jBaseURL,nullptr);</pre>

become

<pre>t.env->CallStaticVoidMethod(t.classID, t.methodID, index, jString, "text/html",nullptr);</pre>

But I'm quite new to all this and not really sure of the possible problems of this fix. Also, this way the baseURL is never used, which doesn't seem right :/.

jgod:
:+1: 

 I noticed this too, same use case as you, except I actually did need the baseURL (using javascript, html, css, linked resources).

This is what I ended up with:

`Cocos2dxWebViewHelper.java`:

```
    @SuppressWarnings("unused")
    public static void loadHTMLString(final int index, final String htmlString, final String baseURL) {
        cocos2dxActivity.runOnUiThread(new Runnable() {
            @Override
            public void run() {
                Cocos2dxWebView webView = webViews.get(index);
                if (webView != null) {
                    webView.loadDataWithBaseURL(baseURL, htmlString, null, null, null);
                }
            }
        });
    }
```

[...]

PR #9157, opened 2014-11-18
Fix loadHTMLString on Android

jgod:
- Fix implementation to conform to API
- Use html string as html string, not mimetype!
- Correctly handle baseURL (like iOS) instead of just ignoring it (why completely ignore param?)

As discussed [here](https://github.com/cocos2d/cocos2d-x/issues/9155)


minggo:
Thanks.

zilongshanren:
[ci rebuild]

WenhaiLin:
@jgod Thank you very much.I try to fixed the base URL of Webview on IOS and Android,please look at https://github.com/cocos2d/cocos2d-x/pull/10200

PR #9159, opened 2014-11-18
In Android, when EditBox is released, callback should not be called from...

matsuhiro:
... Java.

When EditBoxDialog is closed, editBoxCallbackFunc is called.
However, if the instance of EditBoxImplAndroid has been removed at that time, thiz is already released.


darkdukey:
[ci rebuild]

zilongshanren:
@matsuhiro This PR is dead for a long time, I want to close it here.
If the issue still exists in the latest v3 branch, feel free to resend another PR.
Thanks.

PR #9169, opened 2014-11-19
fix : flickering when use fade animation with easing

keidroid:
invalid opacity value (when use fade animation with elastic or back easing)
ex. EaseBackIn::create(FadeTo::create(1.0f, 255.0f));


zilongshanren:
[ci rebuild]

minggo:

keidroid:
@minggo sorry, I added a test case to EaseAction test.

minggo:
It is ok to add there.
Thanks.

keidroid:
thanks for reviewing.

PR #9170, opened 2014-11-19
Scissor & Stencil commands

mannewalis:
- This change implements renderer commands for scissor and stencil.
- All uses of scissor and stencil that were previously using custom commands now use the scissor and/or stencil commands.
- commands can be nested.
- refactored the renderer visit to use a switch instead of gigantic if else statement, which should be faster and avoid branch mis-prediction.

I have tested all instances using cpp-tests and it appears to function correctly, however I think additional testing is required to ensure complete backward compatibility.

Thanks
Justin


ricardoquesada:
[ci rebuild]

ricardoquesada:
[ci build]

dumganhar:
[ci rebuild]

darkdukey:
[ci rebuild]

mannewalis:
Closing this and reissuing against v4-develop where it's supposed to be.

PR #9171, opened 2014-11-20
Updated angle to commit 4a980d2612

stammen:
This pull request updates the Windows 8.1 versions of ANGLE for Windows Store and Windows Phone. This version of ANGLE uses the new interface approved by the ANGLE Project. Please note that the Universal App on the cocos2d-x v3 branch will be broken by this new version of ANGLE until this pull request becomes part of the download-deps. This pull request depends on the following pull request containing the new ANGLE libs and header files:

https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/56

### DIFF
--- a/cocos/platform/win8.1-universal/OpenGLES.cpp
+++ b/cocos/platform/win8.1-universal/OpenGLES.cpp
@@ -48,51 +48,107 @@ void OpenGLES::Initialize()
         EGL_NONE
     };
 
-    const EGLint displayAttributes[] =
-    {
-        // This can be used to configure D3D11. For example, EGL_PLATFORM_ANGLE_TYPE_D3D11_FL9_3_ANGLE could be used.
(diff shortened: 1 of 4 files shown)


minggo:
Why update Angle?
Does it fix some bugs or performance improved or others?

stammen:
This version now uses the approved interface by the ANGLE Project. Also it has slightly better frame rate and fixes a bug with negative viewport values. Mipmap bugs were also fixed by the angle team.

minggo:
Thanks.
What did you mean `approved interface`?

stammen:
Google AngleProject has accepted the interface changes needed to create the GL Context from the app to angle via EGL. This mean that the Angle interface should now be stable. The code is currently being merged into AngleProject. In a few months, we will probably be able get Angle from the main AngleProject repo and not the MS Open Tech repo.

PR #9176, opened 2014-11-20
Fixed accelerometer for windows store apps device orientation

stammen:
Fixed accelerometer for windows store apps device orientation. Windows Phone 8.1 and Windows Store 8.1 require different handling of x and y values from the accelerometer based on the device orientation.


ricardoquesada:
thanks dale.

Issue #9109, opened 2014-11-13
CSParseBinary.proto


martell:
Where is this file?
If someone uses a different version of google's protobuf
They get this error
CSParseBinary.pb.h:18:2: error: 
# error This file was generated by an older version of protoc which is
# error incompatible with your Protocol Buffer headers.  Please
# error regenerate this file with a newer version of protoc.

I have no idea what protobuf does but we need the source not a generated binary blob

Can someone give instructions on how to generate the header and source file from this also

@ricardoquesada This may interest you :)

minggo:
We will use flat buffer instead of protocol buffer before releasing v3.3.

martell:
Thanks for the reply minggo :)
I am in the middle of releasing a cocos.py for msys2.
and the end user cant use CCB or CCS because of this.

The sooner the better on this so I can get in time for testing before 3.3 :)

minggo:
already use flatbuffer for studio reader, so close this issue

PR #9177, opened 2014-11-20
Cmake unify usage of prebuilt libraries

vovkasm:
Cmake build system: unify usage of prebuilt libraries

This patch unify usage of external libs while building process. And should simplify future development. Also some small fixes that I was need to include because can’t test without them.
1. By default use our prebuilt libs. (exception is mingw for witch we don’t have prebuilt libs, so try to find system installed pkgs)
2. Allow to set `-DUSE_PREBUILT_LIBS=NO` to force finding external deps in system (as other soft with cmake Find\* modules)
3. Add some missing deps and link errors.
4. Build luasocket on Linux and Mingw
5. Build test executables on Mingw
6. Situation with unzip. For now we have modified minizip sources from [official zlib contrib/minizip folder](https://github.com/madler/zlib/tree/master/contrib/minizip), most systems package unmodified version from this sources. But MSYS2 package more modern «version» from [github.com/nmoinvaz/minizip repo](https://github.com/nmoinvaz/minizip) (in quotes, because repo contains no branches or tags, so no real released versions). So… we use own sources mostly with one exception - mingw, where MSYS2 package fresh version.
7. protobuf. We use own sources, because version packaged with other systems can be incompatible with generated protobuf interfaces (witch itself has not committed sources). (I mean file: `cocos/editor-support/cocostudio/CSParseBinary.pb.cc`)
8. Connect new audio engine to build.
9. On Win, test executable not search Resources folder, so install resources in same folder, as executable.

This can be tested on:
- MSYS2. mingw32: build success, tests runnable.
- MSYS2. mingw64: build success, tests runnable.
- Arch Linux 64. USE_PREBUILT_LIBS=YES. success, runnable (also not all deps was prebuilt, zlib not prebuilded, fontconfig not prebuilded)
- Arch Linux 64. USE_PREBUILT_LIBS=NO. success, runnable.
- Mac. USE_PREBUILT_LIBS=YES. success, runnable.
- Mac. USE_PREBUILT_LIBS=NO. (with all deps installed from Homebrew). success, runnable.

Lua tests currently build, but runs for me only on some configs (but this can be fixed in next patches).


martell:
Regarding number 7.
https://github.com/cocos2d/cocos2d-x/issues/9109

vovkasm:
Thank you @martell! So when new library will be committed, we will simply remove protobuf from deps.

ricardoquesada:
Thanks, great work!

zilongshanren:
@vovkasm 
Hi, This patch break the linux CI server. Could you please take a look at the error message and send another patch to fix it?
Thanks.

vovkasm:
@andyque, yes... Sorry for that (
I will send fixes today!

zilongshanren:
@vovkasm 
Thanks, our Linux CI machine is CentOS 6.5 and the CMake version is 2.8.8.

And here is the error message link: http://115.28.72.132:8000/job/cocos-2dx-pull-request-build/6043/node=linux_centos/console

vovkasm:
Thanks, for information. @andyque, can you please in addition say, how dependencies installed to Linux CI? Are they preinstalled or some script exists?

vovkasm:
Done: https://github.com/cocos2d/cocos2d-x/pull/9218
I've test with centos, but some libraries (glew and glfw) was installed from source because I can't find any reasonable versions of it in system repos. gcc also was used from devtoolset-2.

If needed, I have shell script, that install all deps and requirements on centos 6.5. I used it to configure virtualbox vm from vagrant.

PR #9189, opened 2014-11-21
make clippingRectangleNode support nest invoke

hugohuang1111:


ricardoquesada:
Thanks. Excuse my ignorance, what is the goal of this patch ? To call `clippingRectangleNode` recursively?

hugohuang1111:
yes

PR #9194, opened 2014-11-21
changed few loops for C++11 range-based for loop

ghost:
As the title says - I changed few loops for C++11 range-based for loops.


ricardoquesada:
thanks. looks good to me.

PR #9198, opened 2014-11-23
refactor

ghost:
I still try to learn this project so I keep doing some refactoring.


minggo:
I think it is already refactored.

PR #9201, opened 2014-11-23
V3 win8.1 universal app pause and resume support

stammen:
This pull request adds support for when the app state changes between visible and not visible. It also adds support for when the directx device is lost. These changes are necessary to pass windows app store certification.

Several of the files had support for Android but not WP8 or WinRT.

Fixed a bug in CCImage.cpp when reloading a tga texture from the volatile texture cache. Since there is no filename (ie. _filePath is "") , the filename should not be used to test if the data is tga.

Some shader tests were disabled for WP8/WinRT since the shaders will not compile on the device.

11/25/14 merged with most recent v3. Updated cpp template file

11/25/14 Added missing OpenURL support to CCApplication.cpp for Universal App

11/25/14 Fixed Application::getCurrentLanguage() for WP8 and Universal App


ricardoquesada:
thanks dale.

PR #9213, opened 2014-11-24
read data from file may contain '\0'

hcqmaker:
read data from file to string need set size


hcqmaker:
read file data to string need set size

PR #9236, opened 2014-11-26
Add applicationWillTerminate to CCApplicationProtocol

programmerdave:
I added the applicationWillTerminate method to CCApplicationProtocol for completeness. I use it in my games to upload any final analytics before the apps close so it would be great to add this to the cocos2dx source.


minggo:
No all platforms support the feature, so close it.

PR #9237, opened 2014-11-26
add exception guard to db operation

zilongshanren:
add exception guard to db operation, this will prevent CI server from reporting positive failure when db operations error occur.


zilongshanren:
[ci rebuild]

PR #8902, opened 2014-10-22
Cmake fix modules include path

vovkasm:
This change allow to include cocos project directory from another CMakeFiles.txt (with add_subdirectory). And allow to us connect cocos as submodule in our own projects and build it with projects itself.

PR #9250, opened 2014-11-26
Fix modules include path to be relative project directory.

vovkasm:
This change allow to include cocos project directory from another CMakeFiles.txt (with add_subdirectory). And will allow anyone to connect cocos as submodule to their project easily.

I can't easily fix config agains current 'v3' in #8902 , sorry. And just created this pull request instead.
Old #8902 can be closed now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ project (Cocos2d-X)
 # The version number
 set(COCOS2D_X_VERSION 3.3.0-rc1)
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${PROJECT_SOURCE_DIR}/cmake/Modules/")
 include(CocosBuildHelpers)
 
 message(${BUILDING_STRING})


vovkasm:
Thank you!

dabingnn:
@vovkasm I found that we are using the new mechanism for linux build, but the template is not using the new mechanism. Could you help us to change the code in template?

vovkasm:
@dabingnn 
Hi!

> Could you help us to change the code in template?

Yes, of course. Moreover it is in my to-do list, but before I need another set of patches to restructure cmake build system. Basically project need easy way to determine various build requirements, such as include dirs, libraries to link, compiler flags and definitions. In example, that I provided in [this post](http://discuss.cocos2d-x.org/t/how-to-add-dragonbones-library-to-my-cocos2d-x-v3-3-project/18395/3) some hacks are used to stole this info from cocos project:

``` cmake
# get definitions required for cocos
get_directory_property(COCOS_DEFINITIONS DIRECTORY cocos2d-x COMPILE_DEFINITIONS)
foreach(def ${COCOS_DEFINITIONS})
    add_definitions(-D${def})
endforeach()

# get link directories required for cocos
get_directory_property(COCOS_LINK_DIRECTORIES DIRECTORY cocos2d-x LINK_DIRECTORIES)
foreach(link_dir ${COCOS_LINK_DIRECTORIES})
    link_directories(${link_dir})
endforeach()

# add cocos include dirs to our include dirs
include_directories($<TARGET_PROPERTY:cocos2d,INCLUDE_DIRECTORIES>)
```

This hacks requires very latest version of cmake. Right solutions should be mach simpler.

So I need some prepare work before templates can be made simple. I plan to do it in near 3-5 days.

dabingnn:
@vovkasm Thanks very much, we will try this.

PR #9257, opened 2014-11-26
fix UIScale9Sprite setFlipped bug

yszheda:
Currently a Scale9Sprite in 3.3RC0 will be flipped even if `_flippedX` or `_flippedY` is set to false, which is a bug.


zilongshanren:
Thanks, it has been solved in the latest code, please check it out

zilongshanren:
@minggo 
This issue was solved at https://github.com/cocos2d/cocos2d-x/pull/9057
so I think we could close this PR.

yszheda:
Sorry for the duplicate patch, I will close the PR.

PR #9260, opened 2014-11-27
Sprite Performance Test automation works OK

ricardoquesada:
Sprite Performance Test automation works OK

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -67,7 +67,7 @@ THE SOFTWARE.
  */
 #ifndef CC_DIRECTOR_STATS_POSITION
 #define CC_DIRECTOR_STATS_POSITION Director::getInstance()->getVisibleOrigin()
-#endif
+#endif // CC_DIRECTOR_STATS_POSITION
 
 using namespace std;
 
@@ -1076,7 +1076,7 @@ void Director::showStats()
     static unsigned long prevCalls = 0;
     static unsigned long prevVerts = 0;
     static float prevDeltaTime  = 0.016; // 60FPS
-    static const float FPS_FILTER = 0.05;
+    static const float FPS_FILTER = 0.10;
 
     _accumDt += _deltaTime;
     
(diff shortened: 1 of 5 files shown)


ricardoquesada:
[ci rebuild]

minggo:
[ci rebuild]

minggo:
Error on windows, the error message: `5>..\Classes\PerformanceTest\PerformanceSpriteTest.cpp(578): error C3861: 'roundf': identifier not found`.

ricardoquesada:
@minggo  thanks. trying again...

 Unfortunately I don't have a vs2012 express. I'm testing everything on vs2013 community edition and it works OK.
I think for v3.4 we should set VS2013 Community Edition as the base compiler for windows.

stammen:
VS2012 does not support roundf

http://blogs.msdn.com/b/vcblog/archive/2013/07/19/c99-library-support-in-visual-studio-2013.aspx

ricardoquesada:
@stammen 
thanks Dale. good to know.
what `#ifdef` should I use to detect VS2012?
I'm using `roundf()` only in the one of the performance tests. But I want to say if using VS2012 then don't use `roundf()`.

Thanks!

stammen:
Use

#if _MSC_VER < 1800

to detect versions of VS before VS 2013

ricardoquesada:
@stammen 
thanks. good to know.

ricardoquesada:
yeah!
(and finally I installed VS2012 Express to prevent these issues with my PR)

cocos/base/CCDirector.cpp
```diff
@@ -1076,7 +1076,7 @@ void Director::showStats()
     static unsigned long prevCalls = 0;
     static unsigned long prevVerts = 0;
     static float prevDeltaTime  = 0.016; // 60FPS
-    static const float FPS_FILTER = 0.05;
+    static const float FPS_FILTER = 0.10;
```
dumganhar, line 1079: @ricardoquesada, do you know why you added `FPS_FILTER` and what does it used for?

dumganhar: I guess it just used for making FPS display more stable, right?
Ccurrently, I wanna get the real FPS (frameRate = 1/dt), but `_frameRate` will return a value after the calculation of `FPS_FILTER`.

Issue #8844, opened 2014-10-18
DrawNode::drawPoints(), cannot set pointSize


ledenis:
```
void drawPoint(const Vec2& point, const float pointSize, const Color4F &color);

void drawPoints(const Vec2 *position, unsigned int numberOfPoints, const Color4F &color);
```

`drawPoints` has no `pointSize` argument, whereas `drawPoint` has it. How to set the size?

dabingnn:
Point Size can be set, but the effect is wrong, we need to fix it later.

dabingnn:
`PointSize` problem postponed

PR #9263, opened 2014-11-27
DrawNode::drawPoint add pointSize parameter

ghost:
fix issue #8844


cocos/2d/CCDrawNode.cpp
```diff
@@ -441,7 +441,7 @@ void DrawNode::drawPoint(const Vec2& position, const float pointSize, const Colo
     _dirtyGLPoint = true;
 }
 
-void DrawNode::drawPoints(const Vec2 *position, unsigned int numberOfPoints, const Color4F &color)
+void DrawNode::drawPoints(const Vec2 *position, unsigned int numberOfPoints, const float pointSize, const Color4F &color)
```
minggo, line 444: It breaks compatibility.

Issue #9255, opened 2014-11-26
label crashed in android.


shrimpz:
this issue will fix by me.
i need an issue code for write testcase.

:{

shrimpz:
```
Size s = Director::getInstance()->getWinSize();
auto parent = Node::create();
parent->setPosition(s.width/2, s.height/2);
parent->setVisible(false);
this->addChild(parent);

auto label =  Label::createWithTTF("Crashed!!!", "fonts/HKYuanMini.ttf", 24);
label->setPosition(VisibleRect::center());
parent->addChild(label);
```

switch home & switch back, crashed

akk0rd87:
Just look at this thread please http://discuss.cocos2d-x.org/t/test-for-issue-9255-crashed/39439.

shrimpz:
I can't logon on discuss.cocos2d-x.org
run the testcase, no crashed, is normal
if the issue is not fixed, run this testcase will be crashed.

PR #9265, opened 2014-11-27
Fix label crashed on android #9255

shrimpz:
fixed issue #9255 

when a label instance add to a invisibility parent node, app switch to home/desktop,  switch back.
CRASHED!!!


minggo:
@Dhilan007 
Could you please take a look?

WenhaiLin:
@shrimpz Thank you very much.

PR #9305, opened 2014-11-28
possible a bug, all these methods should be public

MadWeedFall:
方法应该声明成public


minggo:
Why need to use public?

MadWeedFall:
I thought all these convenient math utils should be open to end users，for users who use this engine， mostly not all objects are inherited from cocos2d::Ref, so all these math utils for handling raw data should be accessible. The Engine should provide math utils that makes matrix computing easy, when comes code that mixed with c(or OC)& c++, I can see no reason why these code not set to public.Why force user to wrap their data to cocos data structure when it's much easier to use methods handles raw data that already exists in the Engine?

PR #9316, opened 2014-11-29
Integrate Spritebuilder loader

Super-Genius:
This pull request integrates spritebuilder loading into cocos2dx.
The library compiles and doesn't affect the building of apps, but the loading of spritebuilder files has not been functionally tested.


minggo:
@klhurley 
Thanks.
There is not plan to support Spritebuider.

PR #9318, opened 2014-11-30
switch input flag

LingJiJian:
if editbox’s flag is passwd, and then i want to change it be normal flag, the _systemControl.textField need reset in setInputFlag.

### DIFF
--- a/cocos/ui/UIEditBox/UIEditBoxImpl-ios.mm
+++ b/cocos/ui/UIEditBox/UIEditBoxImpl-ios.mm
@@ -456,6 +456,7 @@ int EditBoxImplIOS::getMaxLength()
 
 void EditBoxImplIOS::setInputFlag(EditBox::InputFlag inputFlag)
 {
+    _systemControl.textField.secureTextEntry = NO;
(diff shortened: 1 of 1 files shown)


cocos/ui/UIEditBox/UIEditBoxImpl-ios.mm
```diff
@@ -583,7 +584,7 @@ static CGPoint convertDesignCoordToScreenCoord(const Vec2& designCoord, bool bIn
 
 void EditBoxImplIOS::setVisible(bool visible)
 {
-//    _systemControl.textField.hidden = !visible;
```
zilongshanren, line 586: why should we uncomment this line?

zilongshanren:
Thanks, @LingJiJian 

Could you add a bug tests to verify this patch?

cocos/ui/UIEditBox/UIEditBoxImpl-ios.mm
LingJiJian, line 586: you're right, it is no use

PR #9324, opened 2014-12-01
fix : incorrect point when ignore anchor point for position

litefeel:


litefeel:
其实我觉得`convert..AR` 这3个函数一点用处没有，不知道什么情况下使用，因为转换出来的point没有意义。

dabingnn:
@ricardoquesada 
Our developer do not know how to use the function `convertToWorldSpaceAR` and `convertToNodeSpaceAR`. Our test case has covered the functions, but I can not understand it, too.
Could you help us to make it clearer?
Thanks.

dabingnn:
@lite3 I think these `ConvertXXXXAR` functions could be deprecated if no one knows how to use it correctly.

litefeel:
@dabingnn yes, I think so too. because it makes no sense.

minggo:
Please send a new PR since there is conflict. And the issue may be fixed.

PR #9340, opened 2014-12-02
Fixed fail to play background music again after invoking stopBackgroundMusic

WenhaiLin:


minggo:
Please connect to corresponding issue.

PR #9366, opened 2014-12-04
Fixed label may show nothing when using system font on Mac

WenhaiLin:


minggo:
Is there a corresponding issue?

WenhaiLin:
This issue is feedback from QQ.I will create corresponding issue late.

PR #9379, opened 2014-12-04
fix crash when shader error on Win32

starwing:
现在在Windows上，一旦Shader编译错误，程序就会直接崩溃。因为在Windows上，诸如glGetShaderiv这种函数其实并不是函数，而是函数指针，因此取其指针得到的其实是指针的指针。而且它的调用方式并不是__cdecl，而是__stdcall，所以这里应该直接写，不能通过函数指针来统一处理。

不过，还是希望有一个两全其美的方法……

### DIFF
--- a/cocos/renderer/CCGLProgram.cpp
+++ b/cocos/renderer/CCGLProgram.cpp
@@ -602,8 +602,9 @@ std::string GLProgram::logForOpenGLObject(GLuint object, GLInfoFunction infoFunc
     if (logLength < 1)
         return "";
 
-    char *logBytes = (char*)malloc(logLength);
+    char *logBytes = (char*)malloc(logLength+1);
     logFunc(object, logLength, &charsWritten, logBytes);
+    logBytes[logLength] = '\0';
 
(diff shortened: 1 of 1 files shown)


dabingnn:
@starwing Thanks, could we redefine `GLInfoFunction` and `GLLogFunction` like this?

```
typedef void (_stdcall *GLInfoFunction)(GLuint program, GLenum pname, GLint* params);
typedef void (_stdcall *GLLogFunction) (GLuint program, GLsizei bufsize, GLsizei* length, GLchar* infolog);
```

starwing:
@dabingnn Okay, that will good on Windows, but you should not use __stdcall on phones.

maybe use GLAPIENTRY or use GLEW's definitions directly?

dabingnn:
It seems that we need to Macro in `CCGLProgram.h` and `CCGLProgram.cpp`, which is not good for maintenance. So I will post another pull request based on yours. Thanks.

dabingnn:
@starwing 
Here is the pull request link #9786.

PR #9393, opened 2014-12-05
Texture created in RenderTexture object has improperly set property that...

jotel:
... alpha is not multiplied

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -224,7 +224,7 @@ bool RenderTexture::initWithWidthAndHeight(int w, int h, Texture2D::PixelFormat 
         _texture = new (std::nothrow) Texture2D();
         if (_texture)
         {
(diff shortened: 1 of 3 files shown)


cocos/2d/CCRenderTexture.cpp
```diff
@@ -238,7 +238,7 @@ bool RenderTexture::initWithWidthAndHeight(int w, int h, Texture2D::PixelFormat
             _textureCopy = new (std::nothrow) Texture2D();
             if (_textureCopy)
             {
-                _textureCopy->initWithData(data, dataLen, (Texture2D::PixelFormat)_pixelFormat, powW, powH, Size((float)w, (float)h));
+                _textureCopy->initWithData(data, dataLen, (Texture2D::PixelFormat)_pixelFormat, powW, powH, Size((float)w, (float)h), true);
```
dabingnn, line 241: Why should we treat Texture2D created from RenderTexture to be premultiplied one?

darkdukey:
[ci rebuild]

minggo:

PR #9394, opened 2014-12-06
Fixed CCFileUtils::renameFile() for wp8 and winrt to pass FileUtils and AssetMangers tests

stammen:
This pull request adds support for renaming a file on WP8 and Windows 8.1 Universal Apps. The FileUtils:: file control functions rename and remove file tests were failing due to incorrect file renaming code for WP8 and WinRT.

Please note these tests fail for the Win32 version in the V3 branch. The Win32 version probably should use the same code as WP8.


minggo:
Thanks @stammen .

Issue #9405, opened 2014-12-08
[v2.2.6 issue]EditBox iOS8 orientation issue  + iphone6 display logic error


zilongshanren:
@minggo  please move this issue to 2.2.6

PR #9060, opened 2014-11-07
Fixed retina/contentScale position errors for iPhone 6+ devices in EditBox

starloop-xavierarias:
Problem is that factor was 1.0 or 2.0 until iOS 8 introduced 3.0.

I've replaced everything that depended on dividing by 2 for contentScaleFactor.

PR #9409, opened 2014-12-08
fix EditBox on iOS8 and iPhone6 Plus issues

zilongshanren:
this PR fix issue https://github.com/cocos2d/cocos2d-x/issues/9405


zilongshanren:
@minggo 
This PR fix EditBox orientation error  on iOS8+.
And it also fixes the issue with this PR https://github.com/cocos2d/cocos2d-x/pull/9060

zilongshanren:
[ci skip]

PR #9412, opened 2014-12-08
fix rotationx and rotationy

super626:
Do not hurry to merge it, discussion is here,
http://discuss.cocos2d-x.org/t/possible-bug-on-rotation-and-scale-in-the-node/18601


darkdukey:
Can you create a test case for it?

ricardoquesada:
let's add test case and merge it for v3.4

minggo:
@super626 

super626:
I see

minggo:
What's the relationship with #9611?

super626:
@minggo 
#9611 can fix this issue, I have tested.
So this pull request can be abandoned.

PR #9417, opened 2014-12-08
retain the sprite frame in sprite

litefeel:
do not removed by SpriteFrameCache::removeUnusedSpriteFrames


litefeel:
@minggo please review.

minggo:
What's the issue of current implementation?

litefeel:
@minggo  
1. add test.plist, contain a.png
2. create sprite with sprite frame a.png, and retain(use).
3. spriteFrameCache::removeUnusedSpriteFrames()
4. create sprite width sprite frame a.png, (error can not find sprite frame a.png)

minggo:
Yep, it will has problem in this situation.
But there is another problem.
After you create Sprites from sprite frames, then you don't need them, so you want to invoke `spriteFrameCache::removeUnusedSpriteFrames` to reduce memory. Then all sprite fames that are used to create Sprite can not be released. But indeed they are not needed.

litefeel:
@minggo 
你的思路是尽量减少内存，但是主要占用内存的是纹理而不是spriteFrame，所以即便是所有的spriteFrame都移除掉了，但是texture还在被使用，所以内存不会降。
这样就出现了texture还在，但是spriteFrame不能找到了。

PR #9418, opened 2014-12-08
add FileUtils::getNormalizePath and use in getNewFilename

litefeel:


minggo:
What's `normalizePath`?
And it is better to use `getNormalizePath` instead if we accept it.

litefeel:
@minggo  it's OK.

litefeel:
[ci rebuild]

minggo:
@lite3 
Why can not use `getNewFileName`?

litefeel:
`getNormalizePath`  call by `getNewFileName`.

why add `getNormalizePath`?
1. 因为标准化路径是一个很独立的事情。
2. 单独作为一个方法可以方便其他地方调用。

cocos/platform/CCFileUtils.cpp
```diff
@@ -1300,6 +1300,63 @@ long FileUtils::getFileSize(const std::string &filepath)
     }
 }
 
+std::string FileUtils::getNormalizePath(const std::string& path) const
+{
+    auto pos = path.find("../");
+    if (pos == std::string::npos || pos == 0)
+    {
+        return path;
+    }
```
minggo, line 1360: You said `so that A//B, A/B/, A/./B and A/foo/../B all become A/B`, but it seems that it is not true, because it will return if it can not find `../`.

litefeel:
[ci rebuild]

cocos/platform/CCFileUtils.cpp
```diff
@@ -1300,6 +1300,65 @@ long FileUtils::getFileSize(const std::string &filepath)
...
 
+std::string FileUtils::getNormalizePath(const std::string& path) const
+{
+    auto pos = path.find("../");
+    if (pos == std::string::npos || pos == 0)
+    {
+    	pos = path.find("//");
+    	if (pos == std::string::npos)
+            return path;
+    }
```
minggo, line 1360: How about A/./B?

Could you please add a test case for all these situations?

litefeel, line 1360: ok, i will test and fix it.

litefeel:
[ci rebuild]

litefeel:
[ci rebuild]

litefeel:
@minggo why not success. please see it.

minggo:
http://115.28.72.132:8000/job/cocos-2dx-pull-request-build/6560/node=linux_centos/console

minggo:
You can search `error` and get the error information.

litefeel:
> error: branch 'pull9418' not found.

why?

minggo:
The real error is

/usr/local/include/c++/4.8.2/bits/vector.tcc:146:5: note:   no known conversion for argument 1 from ‘__gnu_cxx::__normal_iterator<const long unsigned int*, std::vector<long unsigned int> >’ to ‘std::vector<long unsigned int>::iterator {aka __gnu_cxx::__normal_iterator<long unsigned int*, std::vector<long unsigned int> >}’
/data/jenkins/workspace/cocos-2dx-pull-request-build/node/linux_centos/cocos/platform/CCFileUtils.cpp:1408:63: error: no matching function for call to ‘std::vector<long unsigned int>::erase(__gnu_cxx::__normal_iterator<const long unsigned int*, std::vector<long unsigned int> >, __gnu_cxx::__normal_iterator<const long unsigned int*, std::vector<long unsigned int> >)’
                 v.erase(v.cbegin() + i - 2, v.cbegin() + i + 2);

litefeel:
[ci rebuild]

litefeel:
[ci rebuild]

litefeel:
[ci rebuild]

cocos/platform/CCFileUtils.h
```diff
@@ -411,6 +411,15 @@ class CC_DLL FileUtils
     /** Returns the full path cache */
     const std::unordered_map<std::string, std::string>& getFullPathCache() const { return _fullPathCache; }
 
+    /** 
+     * Normalize a pathname by collapsing redundant separators and up-level references 
+     * so that A//B, A/B/, A/./B and A/foo/../B all become A/B.
```
minggo, line 416: Please add test case for these types path.

litefeel:
OK, but why failed by default?
qq 20150107113312

minggo:
We don't know the default, we just care about Jenkins CI.

Current implementation is complex, following is my suggestion:

``` c++
// path: original path

if (last character is . or /)
   remove it

i = path.find(0, '.');
while (i != string::npos)
{
    // copy from 0 to i-1
    newPath = path.copy(0, i-1);
    // skip following . or /
    while (path[i] == '.' or path[i] == '/')
       i++;

    i = path.find(i, '.');
}
```

minggo:
Sorry, should be

``` c++
// path: original path

if (last character is . or /)
   remove it

previous = 0;
next = path.find('.');
while (next != string::npos)
{
    // copy from 0 to i-1
    newPath += path.copy(previous, next - 1);
    // skip following . or /
    while (path[next+1] == '.' or path[next+1] == '/')
       next++;

    previous = next;
    next = path.find(previous, '.');
}
```

newPath += path.copy(previous, end-1);

litefeel:
`/.a/b` is a file path. can not normalize.

minggo:
I think we don't have to handle `/.a/b`, it is a wrong path.

minggo:
`.a` will be treated as a folder name.

litefeel:
yes. so `.a` can appear.

minggo:
I mean it is a wrong path, we don't have to correct it.

litefeel:
why wrong path for `/.a/b`?

litefeel:
for example `.git/config`

minggo:
> /.a/b is a file path. can not normalize.

What did you mean?

litefeel:
`/.a/b` will become to `/.a/b`

minggo:
What's the difference?

litefeel:
no change.

minggo:
So, do you have any question about my suggestion?
Current implementation is complex.

litefeel:
it is incorrect.
`.a`
`a/../b`

minggo:
Why it is incorrect, it will skip  `./`, `../`

minggo:
I just describe the thought, not a detail implementation.
How do you think the thought?

litefeel:
I use vector<int> store position and length, use less memory and memory copy.

minggo:
You can use vector too. The codes are pseudocode.

minggo:
But the algorithm is simpler.

litefeel:
can not search `.`, delimiter is `/` only. so the logic is incorrect. (`/.a/b`)
in my code, the last two loop can merge to one. but logic will can not clean.

litefeel:
OK, I add test.

litefeel:
[ci rebuild]

litefeel:
[ci rebuild]

litefeel:
please review it. @minggo

minggo:
@lite3 
Thanks.

We have released rc0 version and will release v3.4 final on Jan.15.
This pull request modifies the basic and important part of engine, so i can not merge it now.

And could you please simplify the algorithm?

minggo:

PR #9432, opened 2014-12-09
Refactor build_native.py

methane:


minggo:
Thanks.
We will use `cocos command` internally in this script file.

minggo:
I send a pull request to use cocos console #9708 .

PR #9458, opened 2014-12-11
Solve the environment conflict with Cocos.

natural-law:
@ricardoquesada, @minggo 
Please review this PR. I removed the logic of write environment `COCOS_X_ROOT` in `setup.py`.

Because `Cocos` depends on the environment to finding the prebuilt libs. Developers will encountered problems when they following these steps:
1. Download & Install Cocos.
2. Create a project from Cocos.
3. Download & Install cocos2d-x.
4. Invoke `setup.py` in cocos2d-x.
5. The project created from Cocos CAN NOT build on platforms: Android & Win32.

So I think the `setup.py` of cocos2d-x should not write the environment `COCOS_X_ROOT`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -483,28 +483,6 @@ class SetEnvVar(object):
 
             self._force_update_env(COCOS_CONSOLE_ROOT, cocos_consle_root)
 
-    def set_cocos_x_root(self):
-        print("->Check environment variable %s" % COCOS_X_ROOT)
-        cocos_x_root = self.current_absolute_path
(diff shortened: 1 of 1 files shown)


minggo:
Is there any problem without this environment for cocos2d-x developers?

natural-law:
As far as I know, there isn't any problem without this environment for cocos2d-x developers.

ricardoquesada:
I'm OK with this PR. But please double-check that it works with cocos2d-x users. thanks.

natural-law:
@ricardoquesada 
OK. Thanks for your review.

PR #9473, opened 2014-12-11
add method setEnabledEventListenersForTarget

unlimitedtas:
AddEventListenerしたあとのリスナーに対して個別にsetEnabledできる機能がなかったため、作りました。

Because I couldn't "setEnabled" of EventListener instance  that executed "addEventListenerWithSceneGraphPriority", I wrote this method.

### DIFF
--- a/cocos/base/CCEventDispatcher.h
+++ b/cocos/base/CCEventDispatcher.h
@@ -104,6 +104,7 @@ public:
     /////////////////////////////////////////////
     
     // Pauses / Resumes event listener
+    void setEnabledEventListenersForTarget(Node* target, bool isEnabled, bool recursive/* = false */);
     
     /** Pauses all listeners which are associated the specified target. */
     void pauseEventListenersForTarget(Node* target, bool recursive = false);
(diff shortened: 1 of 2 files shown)


cocos/base/CCEventDispatcher.h
```diff
@@ -104,6 +104,7 @@ class CC_DLL EventDispatcher : public Ref
     /////////////////////////////////////////////
     
     // Pauses / Resumes event listener
+    void setEnabledEventListenersForTarget(Node* target, bool isEnabled, bool recursive/* = false */);
```
minggo, line 107: How about rename to `enableEventListenersForTarget`?

darkdukey:
[ci rebuild]

cocos/base/CCEventDispatcher.h
minggo, line 107: @unlimitedtas 
How about the opinion?

minggo, line 107: Oh, it will set a boolean value.
So i think the name is good.

minggo:
@unlimitedtas 

minggo:
close this pr since it is unresponsible

PR #9477, opened 2014-12-11
Update that serialize plist file name and load plist file

pipu:


darkdukey:
Does plist only works with png? what if there is different file format?

pipu:
@darkdukey 
Plist also could work with jpeg I find out.

minggo:
[ci rebuild]

PR #9489, opened 2014-12-13
Clang static analyzer crash fix

newnon:
Clang static analyzer crash on this two lines
just ignore it in analyzer

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -971,6 +971,7 @@ void Director::purgeDirector()
 #pragma warning (push)
 #pragma warning (disable: 4996)
 #endif
+//it will crash clang static analyzer so hide it if __clang_analyzer__ defined
 #ifndef __clang_analyzer__
     DrawPrimitives::free();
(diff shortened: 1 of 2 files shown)


minggo:
@newnon 
Why it will crash?

minggo:
And how did you use clang to do static analyzation?
Is there any document for it?

newnon:
Simplest way to run it just run XCode - Analyze it will run CLang static analyzer for whole project. but unfortunately it will crash on this 2 files, after some research if found that it crashes on this 2 lines of code

newnon:
[ci rebuild]

minggo:
Thanks @newnon , i will try it.

minggo:
Yep, i have tried, and met the problem you said.
Do you know why it will cause crash?

newnon:
No i have absolutely no idea why, but i think it's a clang bug. But the simplest way to fix it just add this few lines.

minggo:
Ok.
Could you please add some comment for these added codes?

PR #9491, opened 2014-12-13
iOS OpenGL refactoring

newnon:
in apple documentation written that OpenGL buffers must cleared on layoutSubview but created only on next frame layoutSubview can call several time on one frame and sometimes [context_ renderbufferStorage:GL_RENDERBUFFER fromDrawable:layer] return false and picture freeze.
For example if you game is landscape, and you use banner or interstitials user click on it, StoreWindows opens(on iPhone it is only portrait) after go back picture freeze on last frame.

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -258,6 +258,7 @@ void Director::drawScene()
 
     if (_openGLView)
     {
+        _openGLView->setBuffers();
         _openGLView->pollEvents();
     }
 
--- a/cocos/platform/CCGLView.cpp
+++ b/cocos/platform/CCGLView.cpp
@@ -119,6 +119,10 @@ void GLView::pollEvents()
 {
 }
 
+void GLView::setBuffers()
+{
+}
+
 void GLView::updateDesignResolutionSize()
 {
     if (_screenSize.width > 0 && _screenSize.height > 0
--- a/cocos/platform/CCGLView.h
+++ b/cocos/platform/CCGLView.h
@@ -124,6 +124,12 @@ public:
     CC_DEPRECATED_ATTRIBUTE virtual void pollInputEvents();
 
     virtual void pollEvents();
+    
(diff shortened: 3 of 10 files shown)


dabingnn:
@huangshiwu Could you please review this PR?

dabingnn:
@newnon 
Could you please give us the link of the documentation?
If you can provide a small test case to show the freeze bug, it would be great.
Thanks.

newnon:
Sorry at this moment a cant find this in documentation but problem still exists.
it was here 
https://developer.apple.com/opengl-es/
but link GLES2Sample is broken
You can find some info here http://stackoverflow.com/questions/4093434/opengl-es-iphone-and-intermittent-error-gl-framebuffer-incomplete-attachment-o
The problem is that layoutSubviews may call several times in one frame, in most cases it is ok but sometimes it is not possible to create framebuffer immediately.
I have no 100% method to reproduce this, but one case in which you can get this error.
App must be landscape but portrait mode need to be supported on app level
i have got this problem on Iphone 6+ if it iphone is in landscape mode on moment than app starts.
Second you can get similar problem after you create some portrait only window in landscape app(StoreViewController) after it closes, sometimes picture freeze.
It is really hard to reproduce this bug, i spend about one week to understand problem.

dabingnn:
Thanks. I have read the info in the link.
From the GLEssentials demo, I saw the code is similar as cocos2d-x one.
If `layoutSubviews()` has some problem, does this problem exist in the sample?

newnon:
Yes i have seen this sample, the problem must be exist. The easiest way to reproduce it to use SKStoreProductViewController in landscape mode, it will force portrait mode and after that layoutSubview will be called 2-3 times, sometime with error, and it can cause problem.

dabingnn:
Sorry, I am not very clear.
Could you help to reproduce the bug? It could be based on cocos2d-x or on apple sample.
Thanks.

PR #9494, opened 2014-12-14
RichText can handle ImageRenderer created with frame name

samael65535:
RichText can handle ImageRenderer created with frame name


darkdukey:
[ci rebuild]

PR #9495, opened 2014-12-14
Fix usage of tinyxml2 in cocostudio code

vovkasm:
Replace `#include "tinyxml2/tinyxml2.h"` with `#include "tinyxml2.h"` in new cocostudio code.
This change fix build with cmake when system installed libraries are used (`-DUSE_PREBUILT_LIBS=NO`).

Some systems may install `tinyxml2` includes without special prefix directory, so `FindTinyXML2.cmake` was written to find include directory that directly contains `tinyxml2.h` header (without prefix dir). Moreover, official install script in [tinyxml2 project](https://github.com/leethomason/tinyxml2/blob/master/CMakeLists.txt#L67) install header to `<prefix>/include/tinyxml2.h`

Also fix some warnings in code. And remove unneeded inclusion of `tinyxml2.h` from user visible header file (`cocos/editor-support/cocostudio/FlatBuffersSerialize.h`), as a result usage of tinyxml in cocos may be "private" for user project.

### DIFF
--- a/cocos/editor-support/cocostudio/ActionTimeline/CCActionTimelineCache.cpp
+++ b/cocos/editor-support/cocostudio/ActionTimeline/CCActionTimelineCache.cpp
@@ -33,7 +33,7 @@ THE SOFTWARE.
 
 #include "cocostudio/CSParseBinary_generated.h"
 
-#include "tinyxml2/tinyxml2.h"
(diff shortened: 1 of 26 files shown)


vovkasm:
[ci rebuild]

PR #9498, opened 2014-12-15
add cocos2d::Vector constructor with initializer_list

mezum:
cocos2d::Vector is now able to create with initializer_list like this:

``` cpp
  auto *cache = cocos2d::SpriteFrameCache::getInstance();

  cocos2d::Vector<cocos2d::SpriteFrame*> frame{
      cache->getSpriteFrameByName("foo"),
      cache->getSpriteFrameByName("bar"),
      cache->getSpriteFrameByName("buz"),
      // ...
  };

  mySprite->setSpriteFrame(frame.at(0));
```

### DIFF
--- a/cocos/base/CCVector.h
+++ b/cocos/base/CCVector.h
@@ -32,7 +32,7 @@ THE SOFTWARE.
 #include <functional>
 #include <algorithm> // for std::find
 
-#if _MSC_VER >= 1800
+#if !define(_MSC_VER) || _MSC_VER >= 1800
 #include <initializer_list>
 #endif
 
@@ -86,7 +86,7 @@ public:
     }
     
     /** Constructor with a initializer_list */
-#if _MSC_VER >= 1800
+#if !define(_MSC_VER) || _MSC_VER >= 1800
     Vector<T>(std::initializer_list<T> list)
       : _data(list)
     {
--- a/cocos/base/CCVector.h
+++ b/cocos/base/CCVector.h
@@ -31,7 +31,10 @@ THE SOFTWARE.
 #include <vector>
 #include <functional>
 #include <algorithm> // for std::find
+
+#if _MSC_VER >= 1800
 #include <initializer_list>
+#endif
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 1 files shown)


cocos/base/CCVector.h
```diff
@@ -32,6 +32,20 @@ THE SOFTWARE.
...
+#include <initializer_list>
+#define __CCVECTOR_ENABLE_INITIALIZER_LIST
+#endif
+
+#else
+
+#include <initializer_list>
+#define __CCVECTOR_ENABLE_INITIALIZER_LIST
+
+#endif
```
minggo, line 47: What's the difference between `#if _MSC_VER >= 1800` and `#else`?

mezum:
In MSVC 2012 (_MSC_VER == 1700), initializer_list is not implemented.
so I want to write like this:

``` cpp
#if !defined(_MSC_VER) || _MSC_VER >= 1800
#include <initializer_list>
#define __CCVECTOR_ENABLE_INITIALIZER_LIST
#endif
```

but bindings generator cannot parse its condition.

minggo:
But your implementation is that, no matter `_MSC_VER >= 1800`, all include the same codes

``` c++
#include <initializer_list>
#define __CCVECTOR_ENABLE_INITIALIZER_LIST
```

minggo:
And if VS2012 doesn't support this implementation, then how developers write the codes?
His codes should write like this

``` c++
#if !defined(_MSC_VER) || _MSC_VER >= 1800
cocos2d::Vector<cocos2d::SpriteFrame*> frame{
      cache->getSpriteFrameByName("foo"),
      cache->getSpriteFrameByName("bar"),
      cache->getSpriteFrameByName("buz"),
      // ...
  };
#else
xxx
```

??

mezum:
I think, we should not use this constructor in our projects which target includes windows and windows phone like that WebView can be used on just Android and iOS, or should have to use MSVC 2013+.
Maybe, this PR was too early to create...

(To be honest, I forgot that we cannot use initializer_list in MSVC 2012:sweat_smile:)

minggo:
@mezum 
WebView can be used on Android and iOS now, so it is put into `cocos2d::experimental` namespace. And we will implement it on all supported platforms

minggo:
@mezum 
Thanks for you contribution.
I think i can not merge this pull request now.
May be we can add it after we remove supporting VS2012.

zilongshanren:
[ci rebuild]

zilongshanren:
[ci rebuild]

PR #9501, opened 2014-12-15
V3 fix uninitialized members. 

joewan:
V3 fix uninitialized members.


cocos/platform/win32/CCApplication-win32.cpp
```diff
@@ -63,11 +63,9 @@ int Application::run()
     PVRFrameEnableControlWindow(false);
 
     // Main message loop:
-    LARGE_INTEGER nFreq;
     LARGE_INTEGER nLast;
     LARGE_INTEGER nNow;
 
-    QueryPerformanceFrequency(&nFreq);
```
minggo, line 70: Why remove this line?

joewan, line 70: it is unused.
do not need to query the frequency.

PR #9530, opened 2014-12-16
Added an option to create resizable windows on desktop.

Neurochrom:
You can also add callbacks that inform the application that the window was resized using GLView::addResizeCalback.
A dummy parameter was added to createWithRect in mobile mobile versions to keep them in sync.
Tested on Windows and iOS.
This is my first official contribution to this project - hope I did not break to many guidelines ;)

### DIFF
--- a/cocos/platform/CCGLView.cpp
+++ b/cocos/platform/CCGLView.cpp
@@ -197,14 +197,10 @@ void GLView::setFrameSize(float width, float height)
     {
        director->_winSizeInPoints = _designResolutionSize;
 
-       //auto ds = isDisplayStats();
-       //director->setDefaultValues();
        director->setGLDefaultValues();
        director->setProjection(director->_projection);
-       //director->createStatsLabel();
-       //director->setDisplayStats(ds);
 
-       // It seems the scene content size doews not matter in most cases but just to be on the safe side.
+       // It seems that the scene content size does not matter in most cases but just to be on the safe side.
        auto scene = director->getRunningScene();
        if (scene)
           scene->setContentSize(_designResolutionSize);
@@ -476,7 +472,7 @@ float GLView::getScaleY() const
     return _scaleY;
 }
 
-void GLViewProtocol::addResizeCalback(const std::function<void(const Size&)>& callback)
+void GLView::addResizeCalback(const std::function<void(const Size&)>& callback)
 {
    _resizeCallbacks.push_back(callback);
 }
--- a/cocos/platform/CCGLView.cpp
+++ b/cocos/platform/CCGLView.cpp
@@ -29,6 +29,8 @@ THE SOFTWARE.
 #include "base/CCDirector.h"
 #include "base/CCEventDispatcher.h"
 
+#include "2d/CCScene.h"
+
 NS_CC_BEGIN
 
 namespace {
@@ -189,6 +191,24 @@ const Size& GLView::getFrameSize() const
 void GLView::setFrameSize(float width, float height)
 {
     _designResolutionSize = _screenSize = Size(width, height);
+
+    auto director = Director::getInstance();
+    if (director->getOpenGLView() != NULL)
+    {
+       director->_winSizeInPoints = _designResolutionSize;
+
+       //auto ds = isDisplayStats();
+       //director->setDefaultValues();
+       director->setGLDefaultValues();
+       director->setProjection(director->_projection);
+       //director->createStatsLabel();
+       //director->setDisplayStats(ds);
+
+       // It seems the scene content size doews not matter in most cases but just to be on the safe side.
+       auto scene = director->getRunningScene();
+       if (scene)
+          scene->setContentSize(_designResolutionSize);
+    }
 }
 
 Rect GLView::getVisibleRect() const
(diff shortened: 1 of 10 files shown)


zilongshanren:
@Neurochrom 
Thanks for the patch, but it seems broke the CI server. Your patch made the Lua binding-generator fails.
Please check it out, thanks.

zilongshanren:
[ci rebuild]

cocos/platform/android/CCGLViewImpl-android.cpp
```diff
@@ -51,7 +51,7 @@ void initExtensions() {
 
 NS_CC_BEGIN
 
-GLViewImpl* GLViewImpl::createWithRect(const std::string& viewName, Rect rect, float frameZoomFactor)
+GLViewImpl* GLViewImpl::createWithRect(const std::string& viewName, Rect rect, float frameZoomFactor, bool _ignored_resizable)
```
minggo, line 54: _ignored_resizable is not used.

cocos/platform/ios/CCGLViewImpl-ios.mm
```diff
@@ -61,7 +61,8 @@ of this software and associated documentation files (the "Software"), to deal
     return nullptr;
 }
 
-GLViewImpl* GLViewImpl::createWithRect(const std::string& viewName, Rect rect, float frameZoomFactor)
+GLViewImpl* GLViewImpl::createWithRect(const std::string& viewName, Rect rect, float frameZoomFactor,
+                                       bool _ignored_resizable))
```
minggo: _ignored_resizable is not used. And it is better to use `ignoredResizable` for parameters.
Thanks.

Neurochrom: Yes this parameter is used only in the desktop version, as there are no resizable windows on mobile devices. It is there just to keep everything compatible.

Since it is not really used in the mobile versions I used an odd name like _ignored_resizable instead of ignoredResizable - but I can change that if you think it would be better.

Neurochrom:
@andyque I fixed the Lua problems, CI build seems successful - check it out ;)

zilongshanren:
@Neurochrom 
No, you don't need to change the lua_cocos2dx_auto.cpp, we use binding-generator to generate this file automatically. Any changes apply to this file will be override by the scripts.

So please remove the modification and do CI rebuild again. The CI slave would call binding-generator to generate the binding code at first before doing build job.

Thanks.

Neurochrom:
@andyque
I suspected that it might by auto generated.
But then maybe it should not really be under source control?
I suspect the auto generated version will be the same as the one I've prepared.

zilongshanren:
@Neurochrom 
Yes, it should be the same as yours. If we don't put the auto generated files under version control, before running the Lua tests project, developers must run `python gen-bindings.py` at first.
So I think it exists mainly for easy usage.

Because the Robot will send another PR after merge your PR. thanks.

@minggo How do you think about this?

minggo:
We need auto generated files automatically.

cocos/platform/android/CCGLViewImpl-android.cpp
minggo, line 54: _ignored_resizable -> _ignoredResizable
And this variable is not used, and has not effective, why need this parameter?

minggo:
And how to create a resizable windows on desktop as the tile said?

Neurochrom:
@andyque Ok, I reverted the change to the autogenerated file. I understand we need it for convenience, yet still I think it is a very bad idea to keep autogenerated files in the same repository as the sources. Maybe as a workaround we could have it in a different repository which somebody can make a submodule if they for some reason do not want to run the python script? The repository with the auto generated files would also be a part of the .zip distribution.

@minggo To create a resizable window on desktop (windows) pass true as the last argument to GLViewImpl::createWithRect
For example:
<code>
glview = GLViewImpl::createWithRect(title, Rect(0, 0, 1280, 720), 1.0f, true);
</code>
The _ignored_ argument it to keep source code level compatibility between all versions (so the above example will compile on iOS - even though iOS is not a windowed system). Since the argument has a default value old code such as
<code>
glview = GLViewImpl::createWithRect(title, Rect(0, 0, 1280, 720), 1.0f);
</code>
still works.

Neurochrom:
@andyque After I removed the changes in the auto generated file the CI reports a fail (yet I cannot find any errors in Jenkins logs). Could you please check?

darkdukey:
[ci rebuild]

Neurochrom:
@andyque @minggo ping

cocos/platform/CCGLView.cpp
```diff
@@ -29,6 +29,8 @@ THE SOFTWARE.
 #include "base/CCDirector.h"
 #include "base/CCEventDispatcher.h"
 
+#include "2d/CCScene.h"
```
zilongshanren, line 32: GLView is a low level concepts and Scene is much high concepts.  The low level class shouldn't rely on the top.

cocos/platform/android/CCGLViewImpl-android.cpp
zilongshanren, line 54: The default parameter are banned in `Cocos Coding Style`: https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#default-arguments

zilongshanren:
Thanks, @Neurochrom  I would rather add a member variable `_isResizeWindow` to the GLView and provide public member functions to toggle the state.
Thoughts?

PR #9543, opened 2014-12-18
comment line 83

tjunxin:
As known,  _glProgramStates is a cocos2d::Map container,it will call object->retain() when insert, and call position->second->release() when erase, so the statement of line 83 is redundant.
If i don't comment line 83, the application will crash when GLProgramStateCache::removeUnusedGLProgramState() is called.


minggo:
Good catch.

PR #9566, opened 2014-12-20
Fixing variable name lenght -> length.

austinkelleher:
Updating variable name lenght to length in CCFont.cpp


austinkelleher:
I am confused as to why the build failed. Anyone? It appears the linux_centos build aborted.

minggo:
Good catch.

PR #9567, opened 2014-12-20
fix serious performance problem of Animate.

joewan:
this is a serious bug, it cause massive cpu consume.
if we have 1000 animate in active scene, this pr can raise 5% performance globally, raise about 70% performance locally of Animate!!!!!!!!!!!! .
because it have massive calls cocos2d::Vector constructor and de-constructor. 
Vector's constructor will also call "addRefForAllObjects"(for loop)!!!!!

This problem is very, very serious,  and still has many similar errors in cocos2d-x.

we must use "auto&" sometimes, especially in update per frame.

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -2346,7 +2346,7 @@ void Animate::update(float t)
 
 Animate* Animate::reverse() const
 {
-    auto oldArray = _animation->getFrames();
+    auto& oldArray = _animation->getFrames();
     Vector<AnimationFrame*> newArray(oldArray.size());
    
     if (oldArray.size() > 0)
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -2237,7 +2237,7 @@ bool Animate::initWithAnimation(Animation* animation)
         float accumUnitsOfTime = 0;
         float newUnitOfTimeValue = singleDuration / animation->getTotalDelayUnits();
 
-        auto frames = animation->getFrames();
+        auto& frames = animation->getFrames();
 
         for (auto& frame : frames)
         {
@@ -2312,7 +2312,7 @@ void Animate::update(float t)
         t = fmodf(t, 1.0f);
     }
 
-    auto frames = _animation->getFrames();
+    auto& frames = _animation->getFrames();
     auto numberOfFrames = frames.size();
     SpriteFrame *frameToDisplay = nullptr;
 
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -591,7 +591,7 @@ void Label::alignText()
         batchNode->getTextureAtlas()->removeAllQuads();
     }
     _fontAtlas->prepareLetterDefinitions(_currentUTF16String);
-    auto textures = _fontAtlas->getTextures();
+    auto& textures = _fontAtlas->getTextures();
     if (textures.size() > _batchNodes.size())
     {
         for (auto index = _batchNodes.size(); index < textures.size(); ++index)
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -591,12 +591,12 @@ void Label::alignText()
         batchNode->getTextureAtlas()->removeAllQuads();
     }
     _fontAtlas->prepareLetterDefinitions(_currentUTF16String);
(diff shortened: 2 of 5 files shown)


joewan:
@minggo, please help review it.

joewan:
i'm not sure about modifications of foreachNodeDescendant in CCActionTimeline.cpp,
But it is necessary to optimize this function, it is recursively!!!!!!!!!!!!

samuele3hu:
[ci rebuild]

minggo:
Thanks @joewan .

cocos/2d/CCActionInterval.cpp
```diff
@@ -2237,7 +2237,7 @@ bool Animate::initWithAnimation(Animation* animation)
         float accumUnitsOfTime = 0;
         float newUnitOfTimeValue = singleDuration / animation->getTotalDelayUnits();
 
-        auto frames = animation->getFrames();
+        auto& frames = animation->getFrames();
```
minggo, line 2240: The return type of `animation->getFrames()` is a reference, should we use `auto&` here?

joewan, line 2240: it is necessary, I will commit it.

joewan, line 2240: yeah. you have finished it.

minggo, line 2240: I wrote a test case like this

``` c++
class Test
{
public:
    Test()
    {
        printf("Test\n");
    }

    void func()
    {
        printf("func");
    }
};

class ReturnReference
{
public:
    Test& getTest()
    {
        test = new Test;
        return *test;
    }

    Test *test;
};

int main(int argc, const char * argv[]) {

    ReturnReference returnReference;

    auto test = returnReference.getTest();
    test.func();

    // insert code here...
    std::cout << "Hello, World!\n";
    return 0;
}
```

Test() was only invoked once. It means `auto test = returnReference.getTest();` doesn't create a new Test. So why need to use `auto&`?

cocos/2d/CCLabel.cpp
```diff
@@ -591,12 +591,12 @@ void Label::alignText()
...
     }
     _fontAtlas->prepareLetterDefinitions(_currentUTF16String);
-    auto textures = _fontAtlas->getTextures();
+    auto& textures = _fontAtlas->getTextures();
     if (textures.size() > _batchNodes.size())
     {
         for (auto index = _batchNodes.size(); index < textures.size(); ++index)
         {
-            auto batchNode = SpriteBatchNode::createWithTexture(textures[index]);
+            auto batchNode = SpriteBatchNode::createWithTexture(textures.at(index));
```
minggo, line 599: We don't use exception in engine, is there need to use at instead of `[]`?

cocos/2d/CCActionInterval.cpp
joewan, line 2240: it calls the copy constructor Test(const Test&).

minggo, line 2240: Yep, you are right.
Thanks.

cocos/2d/CCLabel.cpp
joewan, line 599: though we do not use exception, I suggest we use operator::at, it is much more intuitive especially when the element does not exist with the key.

Many programmers don't know that operator[] will create a new element when it does not exist.

And operator[] let problem become much more hidden and harder to solve.

If we use map::at, it will cause app crash directly when element does not exist, and it is convenient for developers.

So In Cocos2d-x Engine, we should use map::at instead of map::operator[].

map::at
If k does not match the key of any element in the container, the function throws an out_of_range exception.

map::operator[]
If k does not match the key of any element in the container, the function inserts a new element with that key and returns a reference to its mapped value. Notice that this always increases the container size by one, even if no mapped value is assigned to the element (the element is constructed using its default constructor).

minggo, line 599: Thanks for your detail explain.
I agree with you we should use insert instead of operator[].

> Many programmers don't know that operator[] will create a new element when it does not exist.

This is true for map but not true for vector. Right?

joewan, line 599: use insert "at" instead of operator[].
I agree with that.

"This is true for map but not true for vector. Right?"  yes.
 but we'd better use "at" everywhere, reduce the "technical difference", such as use "at" with map, use "[]" with vector .

minggo, line 599: Yep, i agree with you. 
Just to make clear.

PR #9569, opened 2014-12-20
cmake: cleanups

vovkasm:
Remove unused code from Find*.cmake modules. All logic with `USE_PREBUILT_LIBS` are handled in `cmake/Modules/CocosUsePrebuiltLibs.cmake`


zilongshanren:
@vovkasm 
Thanks.

zilongshanren:
@minggo 
After CI is ready, I think we could merge it.

PR #9575, opened 2014-12-22
#9555 MoveTo and MoveBy support Vec3

minggo:


minggo:
#9555

stevetranby:
If this is to be included in 3.x then this breaks when a node is previously set to a different positionZ other than zero (0). MoveTo should set a flag (use node's Z position) when passed in a Vec2 instead of just assuming it should be set to 0.

OR it should be noted that this is a breaking change, possibly by not accepting a 2D position?

minggo:
@stevetranby 
Thanks.
A flag is needed. 
I will send a PR to fix it.

minggo:
@stevetranby 
I have looked the codes again.
0 is just set in create, i think it is safe.

stevetranby:
MoveBy is fine because it's using a delta, so delta.z == 0 works fine as you said. Plus the _use3D flag seems unecessary since it's being set both with MoveBy::create(float,Vec2) and MoveBy::create(float,Vec3)

I could be mistaken, but with my code as is I use setPositionZ() on basically every game object (remnant from old setVertexZ, otherwise I'd use setPosition3D instead) for one of our isometric games. When the MoveTo is executed it moves to the end position which is set to have _endPosition.z = 0, and thus the deltaPosition it passes to MoveBy internally has a delta.z == 0 - node.z. 

I know I should use MoveTo with Vec3 going forward, but then I'd have to update all my MoveTo::create call statements and would rather do that on a per need basis.

PR #9578, opened 2014-12-22
update for object lay out system with samples

liamcindy:


cocos/ui/UILayoutComponent.h
```diff
@@ -29,68 +29,140 @@ THE SOFTWARE.
...
+        virtual float getRightMargin();
+        virtual void setRightMargin(float margin);
+
+        virtual float getTopMargin();
+        virtual void setTopMargin(float margin);
+
+        virtual float getButtomMargin();
+        virtual void setButtomMargin(float margin);
+
+#pragma endregion
```
minggo: It is not compatible in all compiles, and it will cause warnings in unsupported compilers.

liamcindy: Ok, i will remove that.

cocos/ui/UILayoutComponent.cpp
```diff
@@ -54,253 +84,530 @@ namespace ui {
...
+    void LayoutComponent::refreshHorizontalMargin()
+    {
+        Node* parent = this->getOwnerParent();
+        if (parent == nullptr)
+            return;
+
+        Point ownerPoint = _owner->getPosition();
+        Point ownerAnchor = _owner->getAnchorPoint();
+        Size ownerSize = _owner->getContentSize();
+        Size parentSize = parent->getContentSize();
```
minggo: Please use reference instead, for example
Point &ownerPoint = _owner->getPosition();

cocos/ui/UILayoutComponent.h
```diff
@@ -29,68 +29,140 @@ THE SOFTWARE.
...
+        virtual void setSizeHeight(float height);
+
+        virtual float getPercentHeight();
+        virtual void setPercentHeight(float percentHeight);
+
+        virtual bool isUsingStretchWidth();
+        virtual void setStretchWidthEnabled(bool isUsed);
+
+        virtual bool isUsingStretchHeight();
+        virtual void setStretchHeightEnabled(bool isUsed);
```
minggo, line 128: Can add const for getXXX functions?

cocos/ui/UILayoutComponent.h
```diff
@@ -29,68 +29,140 @@ THE SOFTWARE.
...
+        virtual void setButtomMargin(float margin);
+
+#pragma endregion
+
+#pragma region Size & Percent
+        virtual Size getSize();
+        virtual void setSize(Size _size);
+
+        virtual bool isUsingPercentWidth();
+        virtual void setPercentWidthEnabled(bool isUsed);
```
minggo, line 107: These two functions are mismatch.

cocos/ui/UILayoutComponent.cpp
liamcindy: OK

cocos/ui/UILayoutComponent.h
liamcindy, line 128: OK

cocos/ui/UILayoutComponent.h
liamcindy, line 107: virtual bool isUsingPercentWidth(); 
virtual void setPercentWidthEnabled(bool isUsed); 

change to set , get ?

cocos/ui/UILayoutComponent.h
```diff
@@ -29,68 +29,140 @@ THE SOFTWARE.
...
+
+        virtual float getSizeWidth();
+        virtual void setSizeWidth(float width);
+
+        virtual float getPercentWidth();
+        virtual void setPercentWidth(float percentWidth);
```
minggo, line 113: What's the meaning of percent width?

liamcindy, line 113: Is that width percent better?

cocos/ui/UILayoutComponent.h
liamcindy, line 107: @minggo 
virtual bool isUsingPercentWidth(); 
 virtual void setPercentWidthEnabled(bool isUsed); 

change to set , get ?

PR #9596, opened 2014-12-23
performance improve

minggo:
Do these things
- use `auto&` instead of `auto` if possible
- correct some coding style
- initialize an object in constructor to remove unneeded assign operation
- remove unneeded temple object

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -457,6 +457,7 @@ void Node::setScale(float scaleX,float scaleY)
 #if CC_USE_PHYSICS
     if(g_physicsSceneCount == 0)
         return;
+    
     auto scene = getScene();
     if (!scene || scene->getPhysicsWorld())
     {
@@ -477,6 +478,7 @@ void Node::setScaleX(float scaleX)
 #if CC_USE_PHYSICS
     if(g_physicsSceneCount == 0)
         return;
+    
     auto scene = getScene();
     if (!scene || scene->getPhysicsWorld())
     {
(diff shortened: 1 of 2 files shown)


cocos/2d/CCNode.cpp
```diff
@@ -580,11 +583,7 @@ void Node::setPosition3D(const Vec3& position)
 
 Vec3 Node::getPosition3D() const
 {
-    Vec3 ret;
-    ret.x = _position.x;
-    ret.y = _position.y;
-    ret.z = _positionZ;
-    return ret;
+    return Vec3(_position.x, _position.y, _positionZ);
```
joewan, line 586: i think here is not elegant enough, but I do not know how to improve.
maybe we need to keep a 3d position.

cocos/2d/CCNode.cpp
```diff
@@ -848,7 +849,7 @@ Node* Node::getChildByTag(int tag) const
 {
     CCASSERT( tag != Node::INVALID_TAG, "Invalid tag");
 
-    for (auto& child : _children)
+    for (const auto& child : _children)
```
joewan, line 852: in my opinion, "child" is a pointer, a raw data, there is no need to use &.

minggo, line 852: Yep, you are right.

cocos/2d/CCNode.cpp
minggo, line 586: In fact, it is better to keep a 3d position.
But as you know, cocos2d-x originally is a 2d game engine, so there is _position(which is a Vec2), and _position is used in many places which means we can not remove it.

If we add a _position3D, then there will be redundant information in engine.

PR #9613, opened 2014-12-24
Adding assets windows phone

PhamDinhTri:
Fix error assets when build windows phone


minggo:
How to reproduce the issue?

PR #9614, opened 2014-12-24
Update HelloCpp.csproj

PhamDinhTri:
fix adding assets


minggo:
What's the difference compared to #9613.

PhamDinhTri:
sorry no difference

minggo:
Ok.
Could you please close one.

And what's the issue?
There is compiling error of new WP8 project?

minggo:
@phamdinhtribk 
Why open it again?

PR #9617, opened 2014-12-25
set search root path & writable path

ghost:


cocos/platform/CCFileUtils.h
```diff
@@ -262,6 +262,11 @@ class CC_DLL FileUtils
     virtual void setSearchPaths(const std::vector<std::string>& searchPaths);
     
     /**
+     * Set search root path.
+     */
+    void setSearchRootPath(const char* path);
```
minggo: What's the meaning of this function?

cocos/platform/CCFileUtils.h
```diff
@@ -496,6 +506,11 @@ class CC_DLL FileUtils
     std::unordered_map<std::string, std::string> _fullPathCache;
     
     /**
+     * Writable path (for debug)
+     */
+    std::string _writablePath;
```
minggo, line 511: Where initialize this member variable?

minggo, line 511: `for debug`?
What does it mean?

cocos/platform/CCFileUtils.h
```diff
@@ -284,6 +289,11 @@ class CC_DLL FileUtils
     virtual std::string getWritablePath() const = 0;
     
     /**
+     *  Set writable/cache path (for debug).
+     */
+    virtual void setWritablePath(const char *writablePath);
```
minggo: Please use std::string for parameter type as other functions do.

cocos/platform/CCFileUtils.h
minggo: Rename to `setDefaultResourceRootPath`?

PR #9619, opened 2014-12-25
Track log for windows

ghost:


cocos/base/CCConsole.h
```diff
@@ -51,6 +51,7 @@ typedef SSIZE_T ssize_t;
 #include "base/ccMacros.h"
 #include "platform/CCPlatformMacros.h"
 
+#define CCLOG_STRING 1
```
minggo: Why not define it in .cpp or use const member variable?

minggo: And i think the name is not suitable.
`1` is not a string, and the name is `CCLOG_STRING`.

PR #9637, opened 2014-12-25
V3 actiontimeline animationlist

geron-cn:
add animationlist into actiontimeline, support play with animation name.


minggo:
[ci rebuild]

PR #9662, opened 2014-12-27
disable checking '.luac' file when load Lua module

dualface:


samuele3hu:
Doest this pr mean that all the lua file should use the `.lua` as the suffix whether original lua file or encrypt file?

PR #9663, opened 2014-12-27
update lua corelib

dualface:
Highlights:
-   improved `class()` function, add multiple inheritance support
-   add `cc.load()`, `cc.bind()`, `cc.unbind()` for Entity-Component system support
-   add some usable functions

### DIFF
--- a/cocos/scripting/lua-bindings/script/cocos2d/functions.lua
+++ b/cocos/scripting/lua-bindings/script/cocos2d/functions.lua
@@ -22,6 +22,26 @@ THE SOFTWARE.
 
 ]]
 
+function printLog(tag, fmt, ...)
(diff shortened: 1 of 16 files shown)


cocos/scripting/lua-bindings/script/cocos2d/functions.lua
```diff
@@ -0,0 +1,634 @@
...
+    end
+end
+
+function math.newrandomseed()
+    local ok, socket = pcall(function()
+        return require("socket")
+    end)
+
+    if ok then
+        -- 如果集成了 socket 模块，则使用 socket.gettime() 获取随机数种子
```
minggo, line 321: Pleas don't use Chinese.

PR #9667, opened 2014-12-27
fixed bug which caused crash when display FPS

SunLightJuly:
fixed bug which caused crash when display FPS: the value of "_secondsPerFrame" of Director was not be Initialized


zilongshanren:
@SunLightJuly 
How to reproduce this bug?

SunLightJuly:
see: https://github.com/dualface/v3quick/issues/122

PR #9669, opened 2014-12-28
fix runtime err remove self or brother when onExit

litefeel:


litefeel:
[ci rebuild]

minggo:
how to reproduce it?

litefeel:
1. create layer.
2. add sprite1, sprite2 to layer.
3. setOnExitCallback for sprite1, and remove sprite1 in the callback.
4. add layer to scene.
5. delay
6. remove the layer.
7. crash

minggo:
It is not safe to remove/add a child in loop.
onExit is good for release some memory.

In theory, it is better to make all function safe to add/remove a child, such as `Node::enumerateChildren`, `Node::onEnter` and `onExit`, but it will make codes to complex to manage.

litefeel:
in this case.

i have an function popMsg(), 
the function to do:
1. show an msg or icon or sh. on running scene
2. delay 3s
3. remove self.

when change scren, it must was removed, becase old scene will be show after sometime.

so. I must be remove it on `onExit`.

so. i want an safe `onExit`.

litefeel:
I think more people need to remove the node in `onExit`.

minggo:
A child will be removed by its parent when a scene is replaced.
Why you need to remove it in `onExit()`?

litefeel:
1. runing scene is scene1.
2. show msg node in scene1.
3. pushScene scene2
4. popScene
5. running scene is scene1.

so, msg node still in scene1.

minggo:
You remove msg node in popScene's onExit()?

litefeel:
In fact, i will remove msg node in pushScene's onExit.

minggo:
Yep, you remove it in scene2's onExit?

litefeel:
remove msgNode in msgNode's onExit when call pushScene.

minggo:
As i said, in theory it is right that every function can be invoked in every situation. But it will make engine too complex and slow performance. Or, we should modify everywhere when looping _children, and the codes seems tricky.

litefeel:
add or remove child in onExit or onEnter is simple. 
so i think will be allow.

litefeel:
you see my code. it is so simple and ease.

minggo:
It is problem if invoking `removeFromParent` in `onExit`, `onExit` or `enumerateChildren`.

litefeel:
now. can not call removeFromParent in enumerateChildren.
but call removeFromParent is safe in onEnter or onExit.

litefeel:
i see. in c++.  first invoke callback . my test code in lua.

why invoke order in onExit:
1. c++ callback
2, js callback
3. other code
4. lua callback

minggo:
But how to resolve the problem: remove more than one node in onExit()?

litefeel:
后面我又想了想，对于`onEnter` `onExit` `enumerateChildren`的作用。
1. `onEnter`: 初始化或者为启动做准备，允许添加自身/兄弟/孩子节点，允许删除孩子节点，不允许删除自身/兄弟节点
2. `onExit`: 释放资源，允许删除自身/兄弟/孩子节点，不允许添加自身/兄弟/孩子节点
3. ·enumerateChildren`: 遍历节点，不允许添加删除自身/兄弟/孩子节点。

那么基于上面的思想。
onEnter函数体内的执行顺序为：
1. c++ js lua 的 callback
2. 其他代码

onExit函数体内的执行顺序为：
1. 其他代码
2. c++ js lua 的 callback

以上还有一个假设，一个Node仅有一个callback(c++ js lua)，或者最多仅有一个修改了节点树结构。

这次的代码还没有修改onEnter onExit的执行顺序，如果你同意我的看法，那么我就修改下。

litefeel:
| op | `onEnter` | `onExit` | `enumerateChildren` |
| --- | --- | --- | --- |
| addChild | [o] | [o] | [o] |
| removeChild | [o] | [o] | [o] |
| addBrother | [o] | [x] | [x] |
| removeBrother | [x] | [o] | [x] |
| removeSelf | [x] | [o] | [x] |
- [x]: can not allow
- [o]: allow

now. i make the table.

minggo:
Please paste english version by google translate if you are not confident with you English. Then all developers can discuss with it.
Thanks.

It sounds a good idea to do it like this. But it is very hard to remember in which situation can do what.

minggo:
If we don't have a good solution about how to fix it more elegant, then let's keep as it is. Because it is simple, can not do remove/add operation in these functions.

litefeel:
please see my table. It can achieve.

minggo:
Yep, i have seen it.
How can developers know these limitation?

minggo:
It is hard to remember, so many things should remember.

litefeel:
- `onEnter`: do init sth, can add node to tree.
- `onExi`t: clear and free resource, can remove node to tree
- `enumerateChildren`:  for loop, can not change node tree.

by this mind.  this rules is so simple and ease.

minggo:
But you table says addChild and removeChild can be used in enumerateChildren/onEnter.

litefeel:
yes. it is allow by code that addChild/removeChild in enumerateChildren, but this rule is complex.
so remove this rule.

minggo:
Got it.
So what's the final table?
Could you share it?

litefeel:
| op | `onEnter` | `onExit` | `enumerateChildren` |
| --- | --- | --- | --- |
| addChild | [o] | [o] | [o] |
| removeChild | [o] | [o] | [o] |
| addBrother | [o] | [x] | [x] |
| removeBrother | [x] | [o] | [x] |
| removeSelf | [x] | [o] | [x] |
- [x]: can not allow
- [o]: allow

if have not this table. Use the following text description , it is simple but not accurate.
- `onEnter`: do init sth, can add node to tree.
- `onExit`: clear and free resource, can remove node from tree
- `enumerateChildren`:  for loop, can not change node tree.

minggo:
The table is the same as before. It can invoke `addChild` and `removeChild` in `enumerateChildren` in the table.

litefeel:
yes.

minggo:
But you said

> enumerateChildren: for loop, can not change node tree.

They are conflict.

litefeel:
我的意思是，这张表里的是最精确的，但是如果没有这张表，用简单的文字来简单描述的话，就不要在`enumerateChildren`中修改节点树，这是为了方便记忆，虽然这个文字描述是不精确的。

又或者，作为一个写程序的思想， `enumerateChildren`是为了遍历，查找，而不是为了修改节点树，
虽然代码上运行在`enumerateChildren`中添加子节点。

cocos/2d/CCNode.cpp
```diff
@@ -1364,9 +1367,19 @@ void Node::onEnter()
...
+
+    _childrenDirty = false;
+    for (ssize_t i = 0; i < _children.size(); i++)
+    {
+        auto child = _children.at(i);
+        if (!child->isRunning())
+            child->onEnter();
+        if (_childrenDirty)
+        {
+            i = -1;
```
minggo, line 1379: It will loop from begin again?
Some children's `onEnter` will be invoked more than once.

litefeel, line 1379: if invoke `onEnter`, the child is running. so, it invoked `onEnter` once.

minggo:
@lite3 
After more thinking about it i think i can not fix this PR, because
- it will slow down the performance
- it is not a right way to remove self or bother in `onExit()`

litefeel:
this pr is not finish.

PR #9682, opened 2014-12-29
Fixes missing libbox2d.lib and libSpine.lib errors

NatWeiss:
libbox2d.lib and libSpine.lib are already built by the main Visual C++ solution and are therefore automatically included as linker input. Removing them from this file ensures that no "missing libbox2d.lib" or "missing libSpine.lib" errors are encountered.

### DIFF
--- a/cocos/2d/cocos2dx.props
+++ b/cocos/2d/cocos2dx.props
@@ -7,7 +7,7 @@
   </PropertyGroup>
   <ItemDefinitionGroup>
     <Link>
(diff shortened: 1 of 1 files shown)


minggo:
I think we need to link them. Box2d and spine are separates projects, they will build corresponding libs. After you remove them, the error happened: http://115.28.72.132:8000/job/cocos-2dx-pull-request-build/6554/node=win32_win7/consoleFull.

PR #9699, opened 2014-12-30
fix background/foreground switch

### LINKED ISSUE #9689: cpp-Test case crash or behave wrong on android device after a background/foreground switch
These test Case behave wrongly or crash:
1. Node:Sprite3D -> Testing UV Animation
2. Node:Sprite3D -> fake shadow effect
3. Node:Sprite3D -> basic toon shader
4. Node:Sprite3D-> light map test

The code can be reproduced by enable `Developer options->Don't keep activities` on android devices.

super626:
fix #9689


super626:
[ci rebuild]

minggo:
[ci rebuild]

PR #9713, opened 2014-12-30
Update CCAnimation3D.cpp

hcqmaker:
FileUtils use lookup file this will be fail

### DIFF
--- a/cocos/3d/CCAnimation3D.cpp
+++ b/cocos/3d/CCAnimation3D.cpp
@@ -30,7 +30,8 @@ NS_CC_BEGIN
 
 Animation3D* Animation3D::create(const std::string& fileName, const std::string& animationName)
 {
-    std::string fullPath = FileUtils::getInstance()->fullPathForFilename(fileName);
+    std::string fullPath = fileName;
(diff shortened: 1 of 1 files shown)


hcqmaker:
FileUtils use lookup file this will be fail

minggo:
@super626 
Could you please take a look?

dabingnn:
@hcqmaker Could you please give a more detailed description of the bug? A test case is preferred if you can provide one. Thanks.

cocos/3d/CCAnimation3D.cpp
```diff
@@ -30,7 +30,8 @@ NS_CC_BEGIN
 
 Animation3D* Animation3D::create(const std::string& fileName, const std::string& animationName)
 {
-    std::string fullPath = FileUtils::getInstance()->fullPathForFilename(fileName);
+    std::string fullPath = fileName;
```
dabingnn, line 33: It seems this code will fail if a relative path is used.

super626:
@hcqmaker  
Agree with @dabingnn 
Can you describe it in detail?

dabingnn:
This pr will be closed, if you have any feedbacks, please add a detailed description of bug and reopen it.
Thanks.

PR #9714, opened 2014-12-30
Update CCSprite3D.cpp

hcqmaker:
FileUtils use lookup will be fail


hcqmaker:
FileUtils use lookup will be fail

minggo:
@super626 
Could you please take a look?

dabingnn:
Similar to #9713

dabingnn:
This pr will be closed, if you have any feedbacks, please add a detailed description of bug and reopen it.
Thanks.

PR #9718, opened 2014-12-30
Disable write to depth buffer for transparent objects

darkdukey:
- Disable writing to depth buffer for transparent and 2D objects
- Depth test will be added later with some performance test
- A better solution is still under development
- [Fantasy Warrior](https://github.com/darkdukey/FantasyWarrior3D/tree/originalCocos) should be rendering correctly with unmodified engine

### DIFF
--- a/cocos/renderer/CCRenderer.cpp
+++ b/cocos/renderer/CCRenderer.cpp
@@ -321,10 +321,8 @@ void Renderer::addCommand(RenderCommand* command, int renderQueue)
     CCASSERT(!_isRendering, "Cannot add command while rendering");
     CCASSERT(renderQueue >=0, "Invalid render queue");
     CCASSERT(command->getType() != RenderCommand::Type::UNKNOWN_COMMAND, "Invalid Command Type");
-    if (command->isTransparent())
-        _transparentRenderGroups.push_back(command);
-    else
-        _renderGroups[renderQueue].push_back(command);
+    
+    _renderGroups[renderQueue].push_back(command);
 }
 
 void Renderer::pushGroup(int renderQueueID)
@@ -823,10 +821,23 @@ void Renderer::flush()
 
 void Renderer::flush2D()
 {
+    //Check depth write
+    GLboolean depthWirte;
+    glGetBooleanv(GL_DEPTH_WRITEMASK, &depthWirte);
+    //Turn depth write off if necessary
+    if(depthWirte)
+    {
+        glDepthMask(false);
+    }
     drawBatchedQuads();
     _lastMaterialID = 0;
     drawBatchedTriangles();
     _lastMaterialID = 0;
+    //Turn depth write on if necessary
+    if(depthWirte)
+    {
+        glDepthMask(true);
+    }
 }
 
 void Renderer::flush3D()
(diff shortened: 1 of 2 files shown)


cocos/renderer/CCTextureAtlas.cpp
```diff
@@ -605,7 +605,16 @@ void TextureAtlas::drawNumberOfQuads(ssize_t numberOfQuads, ssize_t start)
 
     if(!numberOfQuads)
         return;
-
+    
+    //Check depth write
+    GLboolean depthWirte;
```
minggo, line 610: depthWirte -> depthWrite

cocos/renderer/CCRenderer.cpp
```diff
@@ -321,10 +321,8 @@ void Renderer::addCommand(RenderCommand* command, int renderQueue)
     CCASSERT(!_isRendering, "Cannot add command while rendering");
     CCASSERT(renderQueue >=0, "Invalid render queue");
     CCASSERT(command->getType() != RenderCommand::Type::UNKNOWN_COMMAND, "Invalid Command Type");
-    if (command->isTransparent())
-        _transparentRenderGroups.push_back(command);
-    else
-        _renderGroups[renderQueue].push_back(command);
+    
+    _renderGroups[renderQueue].push_back(command);
```
super626, line 325: @darkdukey 
@minggo 
The global z in transparent objects is different with that in opaque objects.
The global z in transparent objects is the distance to the camera. You will get wrong result sort the transparent and opaque objects together.

darkdukey, line 325: I agree, trying to think of a way to organize the command better, because sprites should also considered as transparent, but currently transparent queue only handles 3d object, correct?

minggo, line 325: But how to determine if a sprite is transparent or not?

darkdukey, line 325: In real world use case 90% of the sprites are transparent, note even if there is only 1 pixel of a sprite is transparent, we have to consider it as transparent in rendering.

Also the graphics processor will handle opaque objects internally, so sorting on opaque objects are not necessary anymore. 
[read this](https://developer.apple.com/library/ios/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/Performance/Performance.html#//apple_ref/doc/uid/TP40008793-CH105-SW22)

minggo, line 325: Thanks for the link.
The document says

> A TBDR graphics processor automatically uses the depth buffer to perform hidden surface removal for the entire scene, ensuring that only one fragment shader is run for each pixel. Traditional techniques for reducing fragment processing are not necessary. For example, sorting objects or primitives by depth from front to back effectively duplicates the work done by the GPU, wasting CPU time.

Does it means it take effect for all objects, including transparent objects and opaque objects?

And it says that

> The GPU cannot perform hidden surface removal when blending or alpha testing is enabled, or if a fragment shader uses the discard instruction or writes to the gl_FragDepth output variable.

But we can not make sure blending or alpha testing is not enabled.

PR #9725, opened 2014-12-31
iOS HttpClient replace libcurl

ghost:


cocos/network/HttpClientIOS.mm
```diff
@@ -0,0 +1,518 @@
...
+
+static std::mutex       s_requestQueueMutex;
+static std::mutex       s_responseQueueMutex;
+
+static std::condition_variable_any s_SleepCondition;
+
+static Vector<HttpRequest*>*  s_requestQueue = nullptr;
+static Vector<HttpResponse*>* s_responseQueue = nullptr;
+
+static HttpClient *s_pHttpClient = nullptr; // pointer to singleton
```
minggo: Don't use Hungary style.

cocos/network/HttpCookie.h
```diff
@@ -0,0 +1,56 @@
...
+
+class HttpCookie
+{
+public:
+    void readFile();
+    
+    void writeFile();
+    void setCookieFileName(const std::string fileName);
+    
+    std::vector<CookiesInfo>* getCookies();
```
minggo: std::vector<CookiesInfo>\* getCookies() -> std::vector<CookiesInfo>\* getCookies() const

cocos/network/HttpClientIOS.mm
```diff
@@ -0,0 +1,518 @@
...
+
+#include "base/CCVector.h"
+#include "base/CCDirector.h"
+#include "base/CCScheduler.h"
+
+#include "platform/CCFileUtils.h"
+
+NS_CC_BEGIN
+
+#define ERROR_SIZE 256
```
dabingnn: The macro `ERROR_SIZE` is easy to conflict with others.
If `static const int ERROR_SIZE ＝ 256;` works well, do not use macro.

PR #9731, opened 2014-12-31
Refactor Physics to improve performance[Don't merge]

WenhaiLin:
1.Move the logic of updating bodies' position and rotation to Node::updatePhysicsBodyTransform
2.Move the logic of coordinate conversion to Node::processParentFlags from PhysicsBody::update
3.Change the existing implementation of gravity.The existing implementation of gravity is not written to be maintainable and may cause unexpected behavior.
4.Remove XXXXInfoXXX classes.

I'm sorry for do much things on one commit and my English. I will try to split it later.


cocos/2d/CCNode.cpp
```diff
@@ -1299,8 +1240,14 @@ void Node::visit()
     visit(renderer, parentTransform, true);
 }
 
-uint32_t Node::processParentFlags(const Mat4& parentTransform, uint32_t parentFlags)
+uint32_t Node::processParentFlags(const Mat4& parentTransform, uint32_t parentFlags, bool updateTransform)
```
minggo, line 1243: `updateTransform` is nothing about parent's flags. It just uses existing mechanism to update Node's transform caused by physics engine. It is strange.

PR #9770, opened 2015-01-03
add portrait setting to cpp-default-template

toriumi0118:
``` sh
cocos new --portrait Foo
```

The _--portrait_ setting didn't work for cpp-default-template.

I added option settings using current cocos-console method.


minggo:
@natural-law
Could you please take a look?

natural-law:
@toriumi0118 
Now our templates are not support creating portrait projects. Because there aren't any resources for portrait. In fact, your PR modified the config files for Android, iOS. But platforms Mac, Win32 & Linux are not ready.
So, I think it's not ready for merged. We will optimize the templates once the portrait template is ready.
After all, thanks for your contribution!

toriumi0118:
@natural-law 
I see. You are right, I could not find out how to create portrait on linux and others.
Thank you!!

Issue #9655, opened 2014-12-26
Can't get data from file when file was using by other application on windows


ffskyfan:
In my project,I use csv file to store my game config data. 
When I tried to get data from the csv file While I was using open-office open this file, I got an error code 32.
I solved this problem.
Change the 3th par of function CreateFileW  from NULL to FILE_SHARE_READ|FILE_SHARE_WRITE in function FileUtilsWin32::getFileData .It will make getFileData function read the file properly when other app using it.

minggo:
Thanks.
Why not send a pull request for it?

minggo:
@Dhilan007 
Could you please send a pull request for it?

PR #9777, opened 2015-01-04
fix the problem that engine can't get data from file on windows

ffskyfan:
fix the problem that engine can't get data from file while other app using the file on windows platform


minggo:
Thanks.
Fix #9655.

PR #9780, opened 2015-01-04
Fixed physics body not still after disable gravitational force by PhysicsBody::setGravityEnable.

WenhaiLin:
Node:Physics-->Set Gravity Enable Test:The white ball should still unless you drag it


minggo:
[ci rebuild]

PR #9786, opened 2015-01-05
fix GLProgram log crash on windows

dabingnn:


minggo:
[ci rebuild]

dabingnn:
issue link #9790

Issue #9789, opened 2015-01-05
read file of size 0 cause memory leak on win32


herojia2021:
i use vld to detect memory leak, in file CCFileUtilsWin32.cpp there is a function getData()

there must be a 

if(buffer)
    free(buffer);

in if(buffer == nullptr || size==0) block, i meet this bug when i use a spine file with no texture,
because i only need the skeleton info, but spine still export a empty atlas for read. thats the file with size 0.

minggo:
Thanks.

PR #9792, opened 2015-01-05
fix memory leak

minggo:
issue #9789

### DIFF
--- a/cocos/platform/win32/CCFileUtils-win32.cpp
+++ b/cocos/platform/win32/CCFileUtils-win32.cpp
@@ -168,8 +168,12 @@ static Data getData(const std::string& filename, bool forString)
 
         if (!successed)
         {
-            free(buffer);
-            buffer = nullptr;
+            // should determine buffer value, or it will cause memory leak
+            if (buffer)
+            {
+                free(buffer);
+                buffer = nullptr;
+            }    
         }
     } while (0);
     


herojia2021:
hi minggo, i am not familiar in github, so i don't know how to make a PR now.
but if i am right. the code you change is not what i mean. for make sure, i test the code on my case.
its still leak.
here is the code what i mean
    if (buffer == nullptr || size == 0)
    {
        std::string msg = "Get data from file(";
        // Gets error code.
        DWORD errorCode = ::GetLastError();
        char errorCodeBuffer[20] = {0};
        snprintf(errorCodeBuffer, sizeof(errorCodeBuffer), "%d", errorCode);
        msg = msg + filename + ") failed, error code is " + errorCodeBuffer;
        CCLOG("%s", msg.c_str());
        //fix memory leak if file size == 0 -Hack by Japarazzi
        if (buffer)
            free(buffer);
    }
    else
    {
        ret.fastSet(buffer, size);
    }

cocos/platform/win32/CCFileUtils-win32.cpp
```diff
@@ -168,8 +168,12 @@ static Data getData(const std::string& filename, bool forString)
 
         if (!successed)
         {
-            free(buffer);
-            buffer = nullptr;
+            // should determine buffer value, or it will cause memory leak
```
joewan, line 171: @minggo
there is no difference between them,
why we need to merge this pr?

thanks.

minggo, line 171: @joewan 
You are right. If the pointer is null, it will do nothing.
But change it like this make it more elegant. 

@Japarazzi 
Sorry, i misunderstood it.
I will fix it now.

minggo, line 171: Done: #9837.

minggo:
Issue #9789.

Issue #9779, opened 2015-01-04
[Win32 CurlTest] Error Code 23


zilongshanren:
Since version 3.4beta

When upgrade libCURL from 7.26 to 7.39.0, the CurlTest displayed "Error Code 23".

PR #9795, opened 2015-01-05
Fix curl error code23

zilongshanren:
This PR fix issue https://github.com/cocos2d/cocos2d-x/issues/9779


zilongshanren:
[ci rebuild]

PR #9797, opened 2015-01-05
Bug fix:When parsing a plist file as image mergin, if in fact the plist ...

doomsdayer:
Bug fix:When parsing a plist file as image mergin, if in fact the plist file is a particle, a crush will occur.

Now it's fixed.


chengstory:
Thanks @doomsdayer.  @minggo It can be merged please.

PR #9801, opened 2015-01-05
add node quaternion to changelog

super626:


CHANGELOG
```diff
@@ -1,6 +1,8 @@
 cocos2d-x-3.4rc0  Jan.9 2015
+    [NEW]           Node: rotation representation using quaternion
```
minggo, line 2: It is merged in beta0.
So please add it into beta0.

PR #9807, opened 2015-01-06
update release notes

mannewalis:
My markdown editor modified some of the spacing, lmk if you want me to add my changes in another program to avoid this problem.

Added allocator release notes.


minggo:
I use [mou](http://25.io/mou/) to edit markdown file. It works well.

mannewalis:
I use LightPaper. The issue turned out to be Xcode "helping" format when I pasted back into the document, so I just saved it out from LightPaper instead which fixed the issue.

PR #9811, opened 2015-01-06
move class cocos2dEGLConfigChooser to Cocos2dxActivity

dabingnn:


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
```diff
@@ -62,6 +62,127 @@ of this software and associated documentation files (the "Software"), to deal
     private static Cocos2dxActivity sContext = null;
     private Cocos2dxVideoHelper mVideoHelper = null;
     private Cocos2dxWebViewHelper mWebViewHelper = null;
+
+    public class cocos2dEGLConfigChooser implements GLSurfaceView.EGLConfigChooser
```
minggo: First character should be capital for class name, and prefix of other classes name is `Cocos2dx`, so
cocos2dEGLConfigChooser -> Cocos2dxEGLConfigChooser

PR #9819, opened 2015-01-06
Update project.pbxproj

zhenelse:
fixed the libcurl.a path.


minggo:
How to reproduce it?
We run tests before releasing v2.2.6, but didn't meet any link issue.

Issue #9821, opened 2015-01-06
CCGLView's Viewport rectangle not initialized on desktops


ericmux:
This is based upon the current tip of the `v3` branch. What I noticed is that unless the user explicitly calls the method `glview->setDesignResolutionSize(...);`  the variable `_viewPortRect` on CCGLView will not be set on desktops, since `GLViewImpl::initWithRect` never writes to it. This causes the touch events' coordinates from GLFW to not be converted correctly, and this is where I came across this issue. I noticed the `getLocationInview` method was not changing the position at all and then I saw this variable was not initialized. 

I suspect this was not the intended behavior (forcing the user to call `glview->setDesignResolutionSize(...)`) so I submitted a small PR with this fix, thanks!

minggo:
@ericmux 
How to reproduce it?

PR #9822, opened 2015-01-06
Setting _viewPortRect for GLViewImpl::initWithRect on desktops.

ericmux:
Please read #9821.


minggo:
@huangshiwu 
Please take a look?

ghost:
@ericmux Thank you for your PR. Yes the problem is exist indeed, but the engine still work well for the touch event handle have already consider this situation.Does some thing wrong caused by this? If not I think we don't need to merge the PR for we don't sure if it will cause some other side-effects.

darkdukey:
[ci rebuild]

minggo:
Yep, i agree with @ghost, so close this PR.

Issue #9493, opened 2014-12-14
3.3rc0 DrawNode drawRect 


GreatToad:
I find DrawNode drawRect , what it show not a rect , it is a circle! 
code: 
this->drawRect(Vec2(0,0),Vec2(0,100),Vec2(100,100),Vec2(100,0),Color4F::RED);

ghost:
@GreatToad You should use another function drawRect(drawRect(const Vec2 &origin, const Vec2 &destination, const Color4F &color))
this->drawRect(Vec2(0,0),Vec2(100,100),Color4F::RED);

Issue #8963, opened 2014-10-29
DrawNode::drawPoint uses the last color


ledenis:
When using DrawNode to draw points, it uses the color of the last point.

The bug can be reproduced by putting this code in cpp-tests, DrawPrimitivesTest.cpp in DrawNodeTest():

```
draw->drawPoint(Vec2(200 + 0, 200), 4, Color4F(1, 0, 0, 1));
draw->drawPoint(Vec2(200 + 10, 200), 4, Color4F(0, 1, 0, 1));
draw->drawPoint(Vec2(200 + 20, 200), 4, Color4F(0, 0, 1, 1));
```

The three points are blue, instead of red / green / blue.

Tested as a win32 application.

ledenis:
I also found out that the **pointSize** is affected as well.

dabingnn:
`Color` problem has been fixed,`PointSize` problem is duplicated with #8844

PR #9830, opened 2015-01-07
DrawNode override drawPoints function to set pointSize, fix draw points ...

### LINKED ISSUE #8844: DrawNode::drawPoints(), cannot set pointSize

ghost:
...use last color bug and drawRect behavior bug

fix #8844 #9493 #8963

### DIFF
--- a/cocos/2d/CCDrawNode.cpp
+++ b/cocos/2d/CCDrawNode.cpp
@@ -663,9 +663,12 @@ void DrawNode::drawDot(const Vec2 &pos, float radius, const Color4F &color)
 	_dirty = true;
 }
 
-void DrawNode::drawRect(const Vec2 &lb, const Vec2 &lt, const Vec2 &rt, const Vec2& rb, const Color4F &color)
+void DrawNode::drawRect(const Vec2 &p1, const Vec2 &p2, const Vec2 &p3, const Vec2& p4, const Color4F &color)
 {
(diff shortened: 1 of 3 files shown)


cocos/2d/CCDrawNode.cpp
```diff
@@ -662,23 +665,7 @@ void DrawNode::drawDot(const Vec2 &pos, float radius, const Color4F &color)
...
-    V2F_C4B_T2F_Triangle *triangles = (V2F_C4B_T2F_Triangle *)(_buffer + _bufferCount);
-    V2F_C4B_T2F_Triangle triangle0 = {a, b, c};
-    V2F_C4B_T2F_Triangle triangle1 = {a, c, d};
-    triangles[0] = triangle0;
-    triangles[1] = triangle1;
-	
-    _bufferCount += vertex_count;
-
-    _dirty = true;
+    drawRect(lb, rt, color);
```
dabingnn: If the rect is rotated, the effect would be wrong.

minggo:
Please add corresponding test cases.

PR #9832, opened 2015-01-07
Fix warning log when loading sphere.c3b

super626:
Fix warning log when load sphere.c3b which is a static model with skin


minggo:
It just fix warning?

PR #9838, opened 2015-01-07
modify trans lua value to fontdefinition error

hugohuang1111:


cocos/scripting/lua-bindings/manual/LuaBasicConversions.cpp
```diff
@@ -846,7 +846,7 @@ bool luaval_to_fontdefinition(lua_State* L, int lo, FontDefinition* outValue , c
         lua_gettable(L,lo);
         if (!lua_isnil(L,-1))
         {
-            luaval_to_color3b(L, -1, &outValue->_fontFillColor);
+            luaval_to_color3b(L, lua_gettop(L), &outValue->_fontFillColor);
```
samuele3hu, line 864: Why using lua_gettop(L) replace -1?They are the same.

samuele3hu, line 864: I'm sorry.This is really a bug.The -1 index would be represent the different top index.For those table parameter conversion,we should use the top index, Thanks to @htlxyz.

samuele3hu: I'm sorry.This is really a bug.The -1 index would be represent the different top index.For those table parameter conversion,we should use the top index, Thanks to @htlxyz.

samuele3hu:
[ci rebuild]

PR #9840, opened 2015-01-07
Improve performance of Physics

WenhaiLin:
1.Move the logic of updating bodies' position and rotation to Node::updatePhysicsBodyTransform
2.Move the logic of coordinate conversion to Node::processParentFlags from PhysicsBody::update
3.Try to avoid the recursion

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -60,7 +60,6 @@ THE SOFTWARE.
 #define RENDER_IN_SUBPIXEL(__ARGS__) (ceil(__ARGS__))
 #endif
 
-extern int g_physicsSceneCount;
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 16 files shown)


Dhilan007:
[ci rebuild]

WenhaiLin:
[ci rebuild]

cocos/2d/CCNode.cpp
```diff
@@ -2128,52 +2004,65 @@ void Node::setPhysicsBody(PhysicsBody* body)
...
+    {
+        node->updatePhysicsBodyTransform(scene, _modelViewTransform, flags, scaleX, scaleY);
     }
 }
 
-PhysicsBody* Node::getPhysicsBody() const
+void Node::updateTransformFromPhysics(const Mat4& parentTransform, uint32_t parentFlags)
 {
-    return _physicsBody;
+    auto newPos = _physicsBody->getPosition();
```
minggo: Use auto&?

minggo:
[ci rebuild]

cocos/2d/CCNode.cpp
WenhaiLin, line 2052: It involved a compiler optimization technique. For more information: http://en.wikipedia.org/wiki/Return_value_optimization

PR #9842, opened 2015-01-07
improve ui::Text and ui::TextField creation performance

zilongshanren:
When call setTTFConfig function with long strings, the creation of ui::Text and ui::TextField is very flow.

Both the `setFontName` and `setFontSize` method call `setTTFConfig` method of Label which is very time consuming when the label string is long.

Try to delay the string assignment could improve the performance significant.

This PR fix issue https://github.com/cocos2d/cocos2d-x/issues/9849


minggo:
Please create an issue for it and link this PR to the issue.
Thanks.

PR #9855, opened 2015-01-08
Fix XmlHttpRequestTest parsing error of response header on iOS platform

### LINKED ISSUE #9788: XmlHttpRequest for Lua can't get the correct _statusText on the iOS and android platform

ghost:
fix #9788

### DIFF
--- a/cocos/network/HttpAsynConnection.m
+++ b/cocos/network/HttpAsynConnection.m
@@ -77,7 +77,17 @@
     statusString = [[NSHTTPURLResponse localizedStringForStatusCode:responseCode] copy];
     if(responseCode == 200)
         statusString = @"OK";
-    if (responseCode/100 != 2)
+ 
+    /*The individual values of the numeric status codes defined for HTTP/1.1
+    | “200”  ; OK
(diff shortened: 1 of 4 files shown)


cocos/network/HttpAsynConnection.m
```diff
@@ -73,7 +74,10 @@ - (void) connection:(NSURLConnection *)connection
     responseHeader = [[httpResponse allHeaderFields] copy];
 
     responseCode = httpResponse.statusCode;
-    if (responseCode != 200)
+    statusString = [[NSHTTPURLResponse localizedStringForStatusCode:responseCode] copy];
+    if(responseCode == 200)
+        statusString = @"OK";
+    if (responseCode/100 != 2)
```
dabingnn: 1.Why using magic code `200`?
2. Is `if (responseCode/100 != 2)` different from `if(responseCode != 200)`?

ghost: That is because when responseCode is 200 the statusString of iOS is different with other platform.

PR #9862, opened 2015-01-08
[Physics]Reduce temporary object creation and save mathematical labor 

WenhaiLin:


cocos/physics/CCPhysicsBody.h
```diff
@@ -189,18 +189,18 @@ class CC_DLL PhysicsBody : public Ref
...
+    const Vec2& getPosition();
     /** get the body rotation. */
-    float getRotation() const;
+    float getRotation();
     
     /** set body position offset, it's the position witch relative to node */
     void setPositionOffset(const Vec2& position);
     /** get body position offset. */
-    Vec2 getPositionOffset() const;
+    Vec2 getPositionOffset() const { return _positionOffset; }
```
dabingnn: Why not using `const Vec2&` here?

PR #9887, opened 2015-01-09
fixed getParentToNodeTransform. it returned a wrong transform if _transf...

takaokato:
...ormDirty was true.


dabingnn:
Thanks.

Issue #9898, opened 2015-01-09
EventDispatcher: crash if adding/removing event listeners and dispatching event in event callback function


fusijie:
`version:` v3.3

`Issue:` if dispatch the same event on the listener's event callback, the program will crash.

`reproduce:`

download the 2 classes(HelloWorldScene & BugUnit) to v3.3 HelloWorld Project.

[https://github.com/fusijie/Cocos2dx-Issue/tree/master/dispatchCustomEvent-crash](https://github.com/fusijie/Cocos2dx-Issue/tree/master/dispatchCustomEvent-crash)

WenhaiLin:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/9963

PR #9901, opened 2015-01-09
add issue9898

fusijie:
refer to : [https://github.com/cocos2d/cocos2d-x/issues/9898](https://github.com/cocos2d/cocos2d-x/issues/9898)


minggo:
Thanks.
But i think it is better not use is id as class name.
Please use a class name that can describe the issue.

fusijie:
@minggo 
ok, I will correct it and resend the pr.

zilongshanren:
[ci rebuild]

minggo:
Duplicated with #9963.

PR #9904, opened 2015-01-09
Fixed dead loop when the width of label[system font] smaller then one word of the string on android

WenhaiLin:


minggo:

PR #9915, opened 2015-01-10
FileUtils::fullPathForFilename should return "" if resource is not found.

AknEp:
I Think this method should return empty string when the resource is not found.
Because, I cannot check if the file found or not with current implementation.


AknEp:
This problem was already commented as `FIXME`.

dabingnn:
Thanks.

PR #9917, opened 2015-01-11
add clear pressed & disabled texture to UIButton

edwardzhou:
UIButton can be created with only normal image, and it will auto scale for pressed state, auto grey for disabled state. This handy feature is saving the effort for graphics-making.

But CocoStudio Editor force you provide 3 images (normal, pressed, disabled) for a button widget. I don't think it is a good design. I have raised this problem in Chinese Forum in several times since 2.0beta. unfortunately, no any improve so far.

Adding this two new workaround functions clearPressedTexture and clear DisabledTexture to UIButton, It is very useful when you have only one normal image using CocoStudio.

Usage (in lua, cpp is the same): 
-- load the scene
local ui = cc.CSLoader:createNode('SampleScene.csb')
-- get the root panel (with default name 'Panel_1' in CocoStudio)
local rootPanel = ui:getChildByName('Panel_1')
-- get the button (suppose button_1)
local button = tolua.cast(ccui.Helper:seekWidgetByName(rootPanel, 'button_1'), 'ccui.Button')
button:clearPressedTexture()
button:clearDisabledTexture()

-- another egg-pain thing for UIButton is, if u want the button disabled (grey and without touch response), you have to 
button:setEnabled(false) -- disable touch ...
button:setBright(false) -- show in grey

### DIFF
--- a/cocos/ui/UIButton.cpp
+++ b/cocos/ui/UIButton.cpp
@@ -854,6 +854,25 @@ Size Button::getNormalSize() const
     return Size(width,height);
 }
 
+void Button::clearPressedTexture()
+{
+    CCLOG("[Button::clearPressedTexture] disable pressed texture");
+    this->removeProtectedChild(_buttonClickedRenderer);
+    _buttonClickedRenderer = Scale9Sprite::create();
+    _buttonClickedRenderer->setScale9Enabled(_scale9Enabled);
+    addProtectedChild(_buttonClickedRenderer, PRESSED_RENDERER_Z, -1);
+    _pressedTextureLoaded = false;
+}
+
+void Button::clearDisabledTexture()
+{
+    this->removeProtectedChild(_buttonDisableRenderer);
+    _buttonDisableRenderer = Scale9Sprite::create();
+    _buttonDisableRenderer->setScale9Enabled(_scale9Enabled);
+    addProtectedChild(_buttonDisableRenderer, PRESSED_RENDERER_Z, -1);
+    _disabledTextureLoaded = false;
+}
+
 }
 
 NS_CC_END
--- a/cocos/ui/UIButton.h
+++ b/cocos/ui/UIButton.h
@@ -204,6 +204,9 @@ public:
      * @since v3.3
      */
     float getZoomScale()const;
+
+    void clearPressedTexture();
+    void clearDisabledTexture();
     
 CC_CONSTRUCTOR_ACCESS:
     virtual bool init() override;


zilongshanren:
@edwardzhou 
Thanks for pointing out the egg-pain issue. But AFAIK, the Studio team is working on a version to support the feature you wanted-- only normal texture is needed when creating a ui::Button.

So I think this PR is not needed.

zilongshanren:
@edwardzhou 
for 

```
-- another egg-pain thing for UIButton is, if u want the button disabled (grey and without touch response), you have to 
button:setEnabled(false) -- disable touch ...
button:setBright(false) -- show in grey
```

It is designed like this. I'm sorry this exists only for back compatible reasons. I think we should consider to change it in v4.x.

edwardzhou:
this PR is useful until CocoStudio 2.1 stable (at least on Mac). especially, those guys like me are still using 2.0.6.

in other hand. I think it also necessary for UIButton to give a chance to clear loaded pressed and disabled texture. In current design, it is one-way ticket, user unable to unload the texture once pressed or disabled texture loaded. this cause less reusability.

darkdukey:
[ci rebuild]

edwardzhou:
will this PR be merged or implementing an alternative?

zilongshanren:
@edwardzhou 
Why do you recreated the pressedTextures and disabledTextures after removing?
I guess you might want to remove the pressedTexture and the disabledTextures in ui::Button, right?

edwardzhou:
yes, but according to the current UIButton design. it requires all those 3 renderers to be initialized.

=========quote from UIButton.cpp=========

```
void Button::initRenderer()
{
    _buttonNormalRenderer = Scale9Sprite::create();
    _buttonClickedRenderer = Scale9Sprite::create();
    _buttonDisableRenderer = Scale9Sprite::create();
    _buttonClickedRenderer->setScale9Enabled(false);
    _buttonNormalRenderer->setScale9Enabled(false);
    _buttonDisableRenderer->setScale9Enabled(false);

    _titleRenderer = Label::create();
    _titleRenderer->setAnchorPoint(Vec2::ANCHOR_MIDDLE);

    addProtectedChild(_buttonNormalRenderer, NORMAL_RENDERER_Z, -1);
    addProtectedChild(_buttonClickedRenderer, PRESSED_RENDERER_Z, -1);
    addProtectedChild(_buttonDisableRenderer, DISABLED_RENDERER_Z, -1);
    addProtectedChild(_titleRenderer, TITLE_RENDERER_Z, -1);
}

void Button::setScale9Enabled(bool able)
{
    if (_scale9Enabled == able)
    {
        return;
    }

    _scale9Enabled = able;

    _buttonNormalRenderer->setScale9Enabled(_scale9Enabled);
    _buttonClickedRenderer->setScale9Enabled(_scale9Enabled);
    _buttonDisableRenderer->setScale9Enabled(_scale9Enabled);

    if (_scale9Enabled)
    {
        bool ignoreBefore = _ignoreSize;
        ignoreContentAdaptWithSize(false);
        _prevIgnoreSize = ignoreBefore;
    }
    else
    {
        ignoreContentAdaptWithSize(_prevIgnoreSize);
    }

    setCapInsetsNormalRenderer(_capInsetsNormal);
    setCapInsetsPressedRenderer(_capInsetsPressed);
    setCapInsetsDisabledRenderer(_capInsetsDisabled);

    _brightStyle = BrightStyle::NONE;
    setBright(_bright);
}    
```

======end quote=======================

So I have to create a new renderer and add it back to UIButton. or definitely nullptr access error will be raised.

zilongshanren:
@edwardzhou 
If so , I think we should add nullptr guard to these pointer access. 
But, you know, Studio will allow us to create a ui::Button only with one normal state image.

So I think your implementation is not mandatory.

N.B.  As a reminder, I think we should also modify ui::Button's implementation, we don't have to create the _buttonClickedRenderer and the _buttonDisableRenderer. We should create them as needed.

edwardzhou:
@andyque For button reusing purpose, I still prefer that UIButton can providing reset methods for pressed & disabled textures. It is also more flexible.

zilongshanren:
@edwardzhou 
But we will left some empty Scale9Sprite instances in the button class, they would be a memory waste, right?

edwardzhou:
@andyque 
first of all, base on current UIButton implementation. there some empty Scale9Sprite instances were created when the button initialized already. this patch won't increase memory usage.

secondly, according to yr previous comment, you will refactor UIButton to create Scale9Sprite instance when need. we can also refactor these 2 functions too. So, no empty Scale9Sprite instances left any more. this is the better way.

zilongshanren:
@edwardzhou 
I know your patch is good at the moment but it's not perfect.  I don't want to refactor your patch later. 
What we need to do is do it once and then everything is done.

For adding two extra API is not acceptable when the final release is nearing. More APIs means more memory, it will push the developers to remember all these dirty things. What we need is an simple and elegant way to solve problems. No extra overhead to the users and ourselves.

I hope you could understand.  
Thanks.

PR #9919, opened 2015-01-11
Lua load chunks from zip supports folder structured lua

edwardzhou:
suppose the lua files was in following structure:

folderA/sample.lua

and package with zip command as normal zip file.
Current luaLoadChunksFromZIP will load it into _G['preload'] as 'folderA/sample.lua', it will fail to be require with require('folderA.sample').

this patch will remove the extension name (.lua & .luac), and replace folder seperator ('/' '\') with '.' .
So 'folderA/sample.lua' is changed to 'folderA.sample'.


samuele3hu:
[ci rebuild]

edwardzhou:
how about this PR?

minggo:
@dualface 
Could you please take a look?

minggo:
[ci rebuild]

minggo:
[ci rebuild]

PR #9922, opened 2015-01-12
Change the order of detach and attach of TextField::upadate

zhaijialong:
I clicked a textfield, make it attached. if I click it again, it will attach, then detach. Is it right behaviour? Or the right way should be detach first, then attach. @andyque


zilongshanren:
Thanks, @joshuastray 

PR #9936, opened 2015-01-12
Integrate CocosPlayClient

WenhaiLin:


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxMusic.java
```diff
@@ -245,6 +247,10 @@ private MediaPlayer createMediaplayer(final String path) {
         MediaPlayer mediaPlayer = new MediaPlayer();
 
         try {
+        	CocosPlayClient.notifyFileLoaded(path);
+        	if (CocosPlayClient.isEnabled() && !CocosPlayClient.isDemo()) {
+				CocosPlayClient.updateAssets(path);
+			}
```
dumganhar: Why notify file loaded before updateAssets?

PR #9939, opened 2015-01-12
Remove cancel connection when http response 20x.

yalab:
Hi.

Almost of http success (code 20x) response has response body(JSON or XML).
The connection cancel is wrong  especially HTTP 200 I think.

Thanks.


ghost:
@yalab Yes you are right, sorry for my fault.Thank you for your pr,but we need the codes to keep consistent with other platforms so i propose another pr to fix the problem #9981

dabingnn:
@yalab could you please test the pull request #9981?

PR #9953, opened 2015-01-13
Render fix

darkdukey:
1. Merge 3D and 2D render queue into one
2. Optimize rendering by grouping opaque and transparent objects
3. Billboard, Sprite3D now supports add 2D object as children
4. Refactor Camera, make it easier to use


cocos/3d/CCSprite3D.h
```diff
@@ -148,6 +148,12 @@ class CC_DLL Sprite3D : public Node, public BlendProtocol
     /** load file and set it to meshedatas, nodedatas and materialdatas, obj file .mtl file should be at the same directory if exist */
     bool loadFromFile(const std::string& path, NodeDatas* nodedatas, MeshDatas* meshdatas,  MaterialDatas* materialdatas);
 
+    /**
+     * Visits this Sprite3D's children and draw them recursively.
+     * Note: all its children will rendered as 3D objects
+     */
+    virtual void visit(Renderer *renderer, const Mat4& parentTransform, uint32_t parentFlags);
```
minggo: add override?

cocos/renderer/CCRenderCommand.h
```diff
@@ -65,6 +65,14 @@ class CC_DLL RenderCommand
     /** set transparent flag */
     inline void setTransparent(bool isTransparent) { _isTransparent = isTransparent; }
 
+    inline bool isSkipBatching() const { return _skipBatching; }
+
+    inline void setSkipBatching(bool value) { _skipBatching = value; }
+    
+    inline bool is3D() { return _is3D; }
```
minggo: adding const?

cocos/base/CCDirector.cpp
```diff
@@ -260,7 +254,7 @@ void Director::setGLDefaultValues()
     setProjection(_projection);
 
     // set other opengl default values
-    glClearColor(_clearColor.r, _clearColor.g, _clearColor.b, _clearColor.a);
+    setClearColor(Color4F::BLACK);
```
minggo: If not cache value of clear color, then setGLDefaultValues() will override it.
And setGLDefaultValues() will be invoked every frame.

darkdukey: If that's the case, it's probably safe to remove it since it just re-apply cached color, and renderer already set default color as black.

minggo: I agree to remove it.

minggo:
[ci rebuild]

PR #9982, opened 2015-01-14
enable disable culling using macro

super626:


minggo:
Please fix conflict.

plugin
```diff
@@ -1 +1 @@
-Subproject commit 3609894790678a7a3a9e66cefe881f75534f98b0
+Subproject commit d62bf5184903f3f8b442190d477f7c8571682b7e
```
minggo: Please don't modify submodule.

tools/cocos2d-console
```diff
@@ -1 +1 @@
-Subproject commit 53130d33970b5d9b15a8f0e4c2632e352a6ade9f
+Subproject commit 9663fc53f3f93311ecef798dca5ea38a52137bb6
```
minggo: Please don't modify submodule.

PR #10033, opened 2015-01-15
Made SocketIO client compatible with SocketIO version 1.x

roed:
For our indie gamedev studio Toothpaste & Bubblegum we wanted to use the newest socket.io version, but we discovered the 1.x version has a different protocol and didn't work with the client shipped with cocos2d-x. That's why we have rewritten the client for socketio version 1.x. We have tested the client on socket.io version 1.2.1.

### DIFF
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -606,8 +606,10 @@ void SIOClientImpl::onClose(WebSocket* ws)
         {
             iter->second->receivedDisconnect();
         }
+        _clients.clear();
     }
-
+    _connected = false;
+    
     this->release();
 }
 
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -135,6 +135,10 @@ public:
 
     std::string getUri();
     std::string getSid();
+    
+    inline bool isConnected() {
+        return _connected;
+    }
 };
 
 
@@ -858,6 +862,10 @@ SIOClient* SocketIO::connect(const std::string& uri, SocketIO::SIODelegate& dele
             socket->addClient(path, c);
 
             socket->connectToEndpoint(path);
+        }
+        
+        if (!socket->isConnected()) {
+            socket->connect();
         }
     }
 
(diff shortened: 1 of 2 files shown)


rauchg:
Great stuff!

samuele3hu:
It seems that it wasn't compatible with socketio 0.9x,does it?

roed:
This client doesnt work with 0.9, that's right, they are so different, 0.9 and 1.x

roed:
What do you want @samuele3hu, that the client should support both? Perhaps it's possible to switch implementation with a define statement? The protocol and logic for version 1.x is very different. Each endpoint connection uses the same session id now. So I think it's not possible to support both within the same class.

darkdukey:
[ci rebuild]

zilongshanren:
[ci rebuild]

darkdukey:
[ci rebuild]

himynameschris:
Ive merged this feature with backwards compatibility into #11441

roed:
was it difficult?

himynameschris:
most of the v1.x protocol handling was contributed by @IVBakker to another implementation I started using the Poco c++ framework, he created a class to handle the difference in packet format between .9 and 1, which i think was the hardest part. I just ported it over to what I original wrote in cocos2d-x for 0.9

here is the original 1.0 client using the Poco libraries: https://github.com/hannon235/socket.io-poco

zilongshanren:
@roed  Another PR to resolve this issue? https://github.com/cocos2d/cocos2d-x/pull/11583

roed:
@andyque it seems so. The other PR is backwards compatible with socketio 0.9. Seems like we both reinvented the wheel. I dont know how the other PR handles reconnection logic etc, but I think the cocos team will go for backwards compatibility

himynameschris:
I did not add any logic to automatically reconnect, so in its current state it is up to the user to catch the socket closing then reconnect on their own. That would be something worth porting over once its merged-

pandamicro:
As https://github.com/cocos2d/cocos2d-x/pull/11583 is merged into v3, I close this PR, Thanks for all the contributions

PR #10045, opened 2015-01-16
fix websocket dead lock

zilongshanren:


zilongshanren:
fix https://github.com/cocos2d/cocos2d-x/issues/10013

PR #10059, opened 2015-01-19
Fixed inconsistent time-scale in Httpclient.

cpascal:
If I use "cocos2d::network::HttpClient::getInstance()->setTimeoutForRead( 60 )" (or setTimeoutForConnect),
it means 60 milli-seconds on Android platforms, while other platforms says 60 seconds.

### DIFF
--- a/cocos/network/HttpClient-android.cpp
+++ b/cocos/network/HttpClient-android.cpp
@@ -577,7 +577,7 @@ private:
         if(nullptr == instnace)
             return false;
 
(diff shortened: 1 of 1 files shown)


samuele3hu:
I think it is right.I forgot the cross-platform property because i didn't use the `setTimeoutForRead`,Thanks to @cpascal.

samuele3hu:
[ci rebuild]

zilongshanren:
[ci rebuild]

PR #10073, opened 2015-01-19
the ui widget need to refresh size by ignore state

liamcindy:


zilongshanren:
Hi, @liamcindy 
Why remove the if statement here?
I think it exists only for performance issue. Could you give me a example why we need this fix?

liamcindy:
@andyque We need to refresh widget's size while refresh ignoreSize state

zilongshanren:
ok, thanks @liamcindy 
@minggo We could merge this PR now.

minggo:
@andyque 
Please create a corresponding issue.
Thanks.

zilongshanren:
@minggo 
This PR is not perfect. @liamcindy  will send another PR to fix this problem.

PR #10077, opened 2015-01-19
Dynamic change the root node's size when changing the resolusion of the simulator

doomsdayer:
When reading csd/csb and the simulator's resolution is changed, change the root node's size to fit the simulator windows size. It will make sure the layout component feature works and simulator will show its effect.


dualface:
Please merge it, thanks.

PR #10097, opened 2015-01-20
Flatbuffer generation updated

doomsdayer:
Move the version to the 1st of its map, ensure the version in csb can be read correctly from every version of reader.


doomsdayer:
[ci rebuild]

PR #10103, opened 2015-01-20
Update CCRef.cpp

uxdxdev:
Updated the comment explaining the Ref::retain and Ref::release functions for readability.


minggo:
Thanks @damorton .

PR #10110, opened 2015-01-20
Fix Animate3D speed bug and add parameter to Ray::intersects

### LINKED ISSUE #10108: Animate3D::setSpeed has not effect when it is used in Sequence

super626:
1. fix #10108
2. add distance parameter to Ray::intersects to know how far is the hit position


dabingnn:
Could you add a ticket for `add distance parameter to Ray::intersects to know how far is the hit position`?

super626:
@dabingnn 
This parameter is very useful especially in the editor.
It let you know which object you are picking up when there is many objects. We pick up the object that near us when the ray intersect multiple objects.

super626:
fix #10198

PR #10156, opened 2015-01-21
Fix warnings

vovkasm:
Fix some warnings in Xcode:
- Fix "field 'XXX' will be initialized after 'YYY'" warning in Xcode by reorder initialization sentences.
- Fix warning: CCLOG_STRING_TAG not used by enclosing it in platform ifdef. Provide empty SendLogToWindow implementation by default.

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -202,10 +202,10 @@ static const char* inet_ntop(int af, const void* src, char* dst, int cnt)
 }
 #endif
 
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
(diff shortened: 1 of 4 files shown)


cocos/base/CCConsole.cpp
```diff
@@ -219,8 +219,12 @@ void SendLogToWindow(const char *log)
             (WPARAM)(HWND)hwnd,
             (LPARAM)(LPVOID)&myCDS);
     }
-#endif
 }
+#elseif
```
minggo: #elseif -> #else?

vovkasm: ouch... of course `#else`, sorry...

minggo:
[ci rebuild]

PR #10175, opened 2015-01-22
fix slider scale9 render

zilongshanren:
this PR fixed issue https://github.com/cocos2d/cocos2d-x/issues/10174


minggo:
Please update CHANGELOG.

Issue #8414, opened 2014-09-30
Mouse button defines in CCEventMouse.h


minggo:
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/4854). It was created at 2014/04/15 11:11:01 +0000
# Description

CCEventMouse.h contains defines for the mouse buttons:

```
#define MOUSE_BUTTON_LEFT       0
#define MOUSE_BUTTON_RIGHT      1
#define MOUSE_BUTTON_MIDDLE     2
#define MOUSE_BUTTON_4          3
#define MOUSE_BUTTON_5          4
#define MOUSE_BUTTON_6          5
#define MOUSE_BUTTON_7          6
#define MOUSE_BUTTON_8          7
```

These defines are outside of the cocos2d namespace, which can lead to conflicts.

They should be moved into the namespace (by moving NS_CC_BEGIN to the line above the defines).

I also wonder about the use of defines, wouldn't they be better of in a MouseButtons enum for example?

Kind regards,
Michaël
- MFHSchoonbrood added comment:
  Here's my proposed solution:

```
#include "CCEvent.h"

NS_CC_BEGIN

enum MouseButtons
{
    MOUSE_BUTTON_LEFT   = 0,
    MOUSE_BUTTON_RIGHT,
    MOUSE_BUTTON_MIDDLE,
    MOUSE_BUTTON_4,
    MOUSE_BUTTON_5,
    MOUSE_BUTTON_6,
    MOUSE_BUTTON_7,
    MOUSE_BUTTON_8
};

class EventMouse : public Event
```

I can send the file to if you prefer...

Kind regards,
Michaël
- dumganhar added comment:
  It seems better. But i think we could make it even better by using `enum class`

``` c++
enum class MouseButton
{
    LEFT,
    RIGHT,
    MIDDLE,
    _4,
    _5,
    _6,
...
};

// When using it
MouseButton::LEFT
MouseButton::_4

```
- MFHSchoonbrood added comment:
  That would make it even more in style with the v3 release, I like it.

kompjoefriek:
Would you consider either of these suggestions for v4?

razermirtz:
[...]

PR #10177, opened 2015-01-22
Fixes cocos2d/cocos2d-x/#8414

kompjoefriek:
Changes as suggested by @dumganhar


minggo:
It will break compatibility.

kompjoefriek:
Ok, would you consider the original solution proposed by MFHSchoonbrood in #8414?
I'm aware this could also break compatibility when placed inside the cocos2d namespace.

PR #10187, opened 2015-01-23
Fix win32 editbox content not clipped correctly

zhaijialong:


zilongshanren:
@minggo 
This PR could be merged now.

PR #10189, opened 2015-01-23
disbale cache for s_attributeFlags

dabingnn:


dabingnn:
This is a temp fix for crash with no VAO support

PR #10195, opened 2015-01-23
Fix the bug in ui::LoadingBar->setProgress()

doomsdayer:
Think about this code:

```
auto bar = ui::LoadingBar::create("1.png");  //create the bar
bar->setPercent(80); //set to 80%
bar->loadTexture("2.jpg"); //change the bar's background to another image, and it is reverted to 100%
bar->setPercent(80); //BUG: This will not work, because the same percent, it is returnd in setPercent()
```

### DIFF
--- a/cocos/ui/UILoadingBar.cpp
+++ b/cocos/ui/UILoadingBar.cpp
@@ -194,6 +194,7 @@ void LoadingBar::loadTexture(const std::string& texture,TextureResType texType)
     barRendererScaleChangedWithSize();
     updateContentSizeWithTextureSize(_barRendererTextureSize);
     _barRendererAdaptDirty = true;
+    _percent = 100;
 }
 
 void LoadingBar::setScale9Enabled(bool enabled)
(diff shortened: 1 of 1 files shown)


minggo:
@andyque 
Please take a look.

zilongshanren:
@doomsdayer 
Thanks.
@minggo  I think we could merge it now.

doomsdayer:
Maybe it's better do something in loadTexture().

The CORE problem is: 
After loadTexture(), the loadingBar shows in 100%, but the _percent in it is 80%.
The return sentence in setPercent is for a better performance, it's not wrong itself.

I think it's better to add setPercent(_percent); in the end of loadTexture().
But still it will be returned.....

@andyque how do you think?

zilongshanren:
@doomsdayer 
yeah, it would be nicer. Could you modify your code based on this PR?

doomsdayer:
ok, I'll do it.

PR #10209, opened 2015-01-26
Fix loading bar reload texture issue

zilongshanren:


doomsdayer:
@andyque  I think it's ok.:thumbsup:

tests/cpp-tests/Resources/ccs-res
```diff
@@ -1 +1 @@
-Subproject commit 615f52a6c5fc1f60c6f01832d110c78a65be3a11
+Subproject commit ce5606d4e520d2671a678b0ba3d5a0b84fdb5ab9
```
minggo, line 1: Should update it?

zilongshanren, line 1: @minggo 
Yes, I added two new images.

minggo, line 1: Ok.

minggo:
[ci rebuild]

PR #10218, opened 2015-01-26
fixed "call ui::TextField getContentSize in insert or delete event callback ,the result is wron"

zilongshanren:
this PR fixed issue #10183


zilongshanren:
@minggo 
This PR is ready to be merged.

Issue #10199, opened 2015-01-23
[FantasyWarrior3D] Billboard bug.


fusijie:
The billboard sometimes does not face to current camera.

darkdukey:
@super626 can you take a look at this one? I created a test case for it.

PR #10220, opened 2015-01-26
Billboard test-case

darkdukey:
Create test-case for billboard bug #10199 
The billboard should always facing the camera.


darkdukey:
[ci rebuild]

darkdukey:
Reviewed with @ricardoquesada

PR #10225, opened 2015-01-27
save/restore render state

darkdukey:
Fix Group command leaking OpenGL state changes.
Save and restore render state


darkdukey:
[ci rebuild]

PR #10227, opened 2015-01-27
Added override of ui::EditBox::getDescription()

cpascal:
Most other Widgets in cocos2d::ui namespace have getDescription() overrided method. (returning its class name)
But cocos2d::ui::EditBox doesn't have getDescription() overrided method  yet..

P.S. In my project, I'm using getDescription() like as RTTI. How about adding "getClassName()" method to every objects that is inherited from cocos2d::Node (or cocos2d::Ref)? I think it's useful for script projects sometimes, too.


zilongshanren:
Thanks, @cpascal  

For adding a "getClassName" method in cocos2d::Node( or cocos2d::Ref), I think it's ok. @minggo @ricardoquesada  How do you guys think about it?

minggo:
But you can use [type info](http://www.cplusplus.com/reference/typeinfo/type_info/) to get class name.
So i don't think we need `getClassName`.

PR #10228, opened 2015-01-27
templates/cocos2dx_files.json does not have "tools/bindings-generator/tools/win32/dos2unix.exe",

flamefox:
when run bindings-generator on windows，it need this


flamefox:
why build failed? This file is only templates files, if not add this line, when you gen lua custom binding, it will show you that you don't have dos2unix

PR #10239, opened 2015-01-27
fix bug on uniform cache

### LINKED ISSUE #10207: Bug on GLProgram::setUniformLocationWith3fv
It will crash when we modify the config file to set the max directional/point/spot light to 3. It fails to pass the light array to the shader.

However, it works if we modify the MeshCommand::setLightUniforms like this,
the function call setUniformLocationWith3f to glUniform3fv

There may be bugs on the hash cache of GLProgram::setUniformLocationWith3f.

super626:
fix #10207


darkdukey:
[ci rebuild]

minggo:

PR #10241, opened 2015-01-27
Fixed may access wrong memory address in Scheduler::schedule

WenhaiLin:


minggo:

PR #10246, opened 2015-01-27
remove warning "declaration shadows a local variable" on Xcode IDE

jeli23:


minggo:
@jeli23 
We can not modify spine runtime. 
Its license doesn't allow to be modified.
Thanks.

Issue #10237, opened 2015-01-27
3d + 2d use case (ui - 3d - ui) does not work correctly


super626:
ui - 3d - ui is a common use case for 3d + 2d.

![tmp23393bea](https://cloud.githubusercontent.com/assets/4159090/5914951/019b8746-a631-11e4-965a-d1d7c87516c1.png)

When click Message menu to pop up a message layer. The message layer should be top most. But it is something like this currently,

![tmp734e482b](https://cloud.githubusercontent.com/assets/4159090/5914971/49c44e86-a631-11e4-9e58-65c34bd6c7a2.png)

super626:
Relative PR is #10236

Issue #10212, opened 2015-01-26
Transparent 3d sprite + 2d sprite can not work correctly


super626:
There are transparent 3d girl and 2d circle on the scene. The girl or circle stay on top alternatively. Look the girl's legs. 

![tmp736df607](https://cloud.githubusercontent.com/assets/4159090/5896278/d01b7518-a564-11e4-9e1e-4ebfbd14bc38.png)
![tmp0f1fd2c7](https://cloud.githubusercontent.com/assets/4159090/5896283/ed87a82e-a564-11e4-94df-d4a842c2bac6.png)

minggo:
Related test case is #10211

Issue #10190, opened 2015-01-23
3d objects and 2d objects overlap


super626:
The cpp-tests->Node: Sprite3D is something like this. The Menu and 3d objects overlap each other. It looks bad.

![tmp7262cc9a](https://cloud.githubusercontent.com/assets/4159090/5870921/4f85e752-a308-11e4-8f3e-7098abe1bc65.png)

Issue #10176, opened 2015-01-22
[linux ios etc] outline is wrong for sprite3D


dabingnn:
![screen shot 2015-01-22 at 3 55 49 pm](https://cloud.githubusercontent.com/assets/4794489/5852304/60f01daa-a24f-11e4-9fa1-111cf64c812f.png)

dabingnn:
@darkdukey @ricardoquesada @super626 
This issue does not exist in 3.3final and 3.4beta0, the outline error plane is a transparent one. I think it is a bug of renderer in current v3.4 branch.

darkdukey:
I briefly investigated this. It's because depth write was off when rendering the spaceship
It only happens to the fade effect, because the fade effect change the alpha which set transparency flag of the RenderCommand which turn off the depth write as a result.

Fade effect -> Sprite3D->setTransparent -> RenderCommand -> setTransparent -> DepthWriteOff

Haven't thought of a good solution.

PR #10248, opened 2015-01-28
V3 fix renderer

dabingnn:
Please do not merge it now, review is needed. There may be several more use case to be fixed.


dabingnn:
issue #10237 #10212 #10190 #10176 are fixed.

minggo:
[ci rebuild]

Issue #9691, opened 2014-12-29
Paused game awaked by the Clock on Android


NewCai:
We can reproduce this bug by following steps:
1. Start a game normally which plays a BGM (this could be a indicator later)
2. Press the power button to lock the android devices.
3. Wait for the Clock to ring(assuming we have settled before)
4. Click the "take a nap" label on Screen(most android phones have this function). The device locked again but the game resumed!

I found the reason is that onResume method was invoked in this situation.But I have not found a good way to fix it.

WenhaiLin:
Fixed at https://github.com/cocos2d/cocos2d-x/pull/10255

PR #10255, opened 2015-01-28
Fixed game awaked by the clock on Android

WenhaiLin:


darkdukey:
[ci rebuild]

walzer:
Bug report from developer #9691

minggo:
It looks good to me, but i am not sure if it will break compatibility especially when integrating with 3d party libraries, such as Facebook.

Need more test on it.

PR #10272, opened 2015-01-29
WP8 libcocos2d build fix: increased build heap memory option to /Zm384

stammen:
Recent commits have caused the WP8 ARM build to fail by running out of heap memory. I increased the build heap size on WP8 and WinRT libcocos2d projects with the /Zm384 option.


stammen:
Thanks!

PR #10274, opened 2015-01-29
Force depth test

darkdukey:
1. Fix an error that forceDepthWrite was not initialized correctly
2. Create test case for this bug, it will happen about 30% of the chance when switch to this test.


minggo:
[ci rebuild]

PR #10279, opened 2015-01-29
MotionStreak MoveTo/MoveBy action OK

seobyeongky:
bug fixed

### DIFF
--- a/cocos/2d/CCMotionStreak.cpp
+++ b/cocos/2d/CCMotionStreak.cpp
@@ -169,6 +169,11 @@ float MotionStreak::getPositionX() const
     return _positionR.x;
 }
 
+Vec3 MotionStreak::getPosition3D() const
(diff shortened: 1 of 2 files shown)


minggo:
What's the bug?

seobyeongky:
MoveTo & MoveBy action doesn't work on MotionStreak node. because MoveTo & MoveBy action use getPosition3D method and MotionStreak has its own coordinate variable(_positionR).

minggo:
Did you mean MoveTo::create(vec3)?
Could you please give some sample codes?

seobyeongky:
MoveTo::create(vec2) and MoveTo::create(vec3) both don't work ok on MotionStreak node.  Because MoveTo uses MoveBy and MoveBy uses MotionStreak's getPosition3D.

PR #10301, opened 2015-01-30
Fix ListView::setDirection issue

pandamicro:


pandamicro:
Before modification, the setDirection with Direction::NONE won't take any effect, because it's directly returned

minggo:
Could you please send to v4-developer?
I have just set default branch to v4-developer.

pandamicro:
Done: https://github.com/cocos2d/cocos2d-x/pull/10303

PR #10303, opened 2015-01-30
Fix ListView::setDirection issue

pandamicro:


darkdukey:
[ci rebuild]

darkdukey:
send to the wrong repo?

pandamicro:
As we switch the development repo to v4_develop, it's supposed to send my PR here, @andyque please review it.

cocos/ui/UIListView.cpp
```diff
@@ -385,14 +385,15 @@ void ListView::setDirection(Direction dir)
 {
     switch (dir)
     {
+        case Direction::NONE:
```
zilongshanren, line 388: why move it up?

pandamicro, line 388: In the old code, suppose we do the following:

```
listView->setDirection(Direction::NONE);
```

This will return directly at line 395. And `ScrollView::setDirection(dir);` will never be invoked. Then this function does nothing when the parameter equals to `Direction::NONE`. Though it suppose to make the listView fixed (un-scrollable)

For `Direction::BOTH`, imagine the listView was initially vertically scrollable, then passing `Direction::BOTH` won't make it scrollable in both direction, this function won't have any effect.

zilongshanren, line 388: thanks.

zilongshanren:
@minggo @dabingnn  We could merge this PR now.

zilongshanren:
@minggo  This PR could be merged.

PR #10307, opened 2015-01-30
Spine release dangling pointer when timeline has any not supported element

yalab:
Hi.

I hound a problem that Spine release dangling pointer when it use not supported timeline element (ex. flipX).

Thanks.


minggo:
@yalab
Because Spine runtime license, we can not modify it.
Could you please send a pull request to https://github.com/EsotericSoftware/spine-runtimes/tree/master/spine-cocos2dx?

yalab:
@minggo Thank you for your infomation. I'll try it.

PR #10325, opened 2015-02-02
HttpResponse reference counting bug fixed

seobyeongky:
Hi.
This is a bug fix branch.
Recently I merged cocos2d-x engine, And I got assertion error from about 1% of HttpRequest (Using XmlHttpRequest)
The assertion failing occurs on the HttpRequest's last release phase in the HttpClient.
The reason of this bug is that, HttpResponse retain/release the HttpRequest and this job is done in worker thread.
The Ref object's are not thread-safe so, should be retain/released in the one thread (which is the main thread)

### DIFF
--- a/cocos/network/HttpClient-android.cpp
+++ b/cocos/network/HttpClient-android.cpp
@@ -795,10 +795,8 @@ void HttpClient::networkThread()
 }
 
 // Worker thread
-void HttpClient::networkThreadAlone(HttpRequest* request)
+void HttpClient::networkThreadAlone(HttpRequest* request, HttpResponse* response)
 {
-    // Create a HttpResponse object, the default setting is http access failed
-    HttpResponse *response = new (std::nothrow) HttpResponse(request);
     std::string responseMessage = "";
     processResponse(response, responseMessage);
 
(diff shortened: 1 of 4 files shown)


samuele3hu:
I think HttpResponse new in the worker thread but release in the CocosThread(main thread) may cause the assert under a certain environment.Can you give me a demo case to reproduct it. And Why the related implement in the `HttpClient::networkThread()` weren't modified.

seobyeongky:
Oh, This bug only occurs when using HttpClient's sendImmediate method. And sendImmediate doesn't use any queue and networkThread function, It just creates new thread. So, networkThread function is not related in this issue.
This is sample code that can reproduce the bug.

1  HttpRequest * req = new HttpReqeust();
2  HttpClient::getInstance()->sendImmediate(req);
3  req->release();

[main thread]                      [worker thread]
ref = 1 (creation)
-------------------------------------------------------
ref--   (line 3)                        ref++   (HttpResponse's constructor)
-------------------------------------------------------
ref-- (HttpResponse's destructor)
ref-- (last release)
(ref means reference count)

As you see there could be a confliction. And the confliction can ignore another operation. When it ignores "ref++" operation then it will fail the assertion, ignores "ref--" operation then it will never released.

seobyeongky:
Oh, This bug only occurs when using HttpClient's sendImmediate method. And sendImmediate doesn't use any queue and networkThread function, It just creates new thread. So, networkThread function is not related in this issue.
This is sample code that can reproduce the bug.

```
1  HttpRequest * req = new HttpReqeust();
2  HttpClient::getInstance()->sendImmediate(req);
3  req->release();
```

| main thread | worker thread |
| --- | --- |
| ref = 1 (creation) |  |
| ref++ (sendImmediate) |  |
| ref-- (line 3) | ref++ (HttpResponse's constructor) |
| ref-- (HttpResponse's destructor) |  |
| ref-- (last release) |  |

As you see there could be a confliction. And the confliction can ignore another operation. When it ignores "ref++" operation then it will fail the assertion, ignores "ref--" operation then it will never released.

PR #10352, opened 2015-02-03
Set cocos version to 3.5

Teivaz:
As cocos version 3.4 has already released.


minggo:
Now we are targeting on v4.
We will keep v3.x as bug fixed, and current there is not plan for next v3 version.

PR #10358, opened 2015-02-04
Vertex buffer draw node

mannewalis:


cocos/renderer/CCTextureAtlas.h
```diff
@@ -165,38 +164,37 @@ class CC_DLL TextureAtlas : public Ref
...
-    void drawNumberOfQuads(ssize_t n);
+    void drawNumberOfQuads(size_t n);
 
     /** draws n quads from an index (offset).
     n + start can't be greater than the capacity of the atlas
 
     @since v1.0
     */
-    void drawNumberOfQuads(ssize_t numberOfQuads, ssize_t start);
+    void drawNumberOfQuads(size_t numberOfQuads, size_t start);
```
minggo, line 186: I like to change ssize_t to size_t, i think we should do the same thing in other places.

ricardoquesada, line 186: it should be `ssize_t`... it means `signed`... by being signed it is easier to detect possible bugs. 
It is defined in the Cocos2d-x C++ guidelines.

mannewalis:
How is it easier? With ssize_t you have two checks to make instead of one.

cocos/renderer/CCTextureAtlas.h
minggo, line 186: Yep. I think we should modify the guide if it is no suitable.
The reason why it is not good are:
- most stl containers use size_t, so in the codes, we have to change ssize_t to size_t to remove warnings
- ssize_t is not included in the standard and isn't portable. For example, it is hard to be used in printf, what's the format type of ssize_t?

darkdukey:
[ci rebuild]

darkdukey:
[ci rebuild]

cocos/renderer/CCBatchCommand.cpp
```diff
@@ -69,11 +97,46 @@ void BatchCommand::execute()
...
+{
+    return _start;
+}
+
+unsigned  BatchCommand::getCount() const
+{
+    return _count;
+}
+
+void BatchCommand::setStart(unsigned start)
```
minggo, line 132: unsigned int or unsigned long?

cocos/renderer/CCVertexIndexBuffer.h
```diff
@@ -25,94 +26,326 @@
...
+    size_t appendElementsT(T* source, size_t elements = 1, bool defer = true)
+    {
+        CCASSERT(0 == sizeof(T) % getElementSize(), "elements must divide evenly into elementSize");
+        auto mult = sizeof(T) / getElementSize();
+        appendElements((const void*)source, mult * elements, defer);
+        return getSize();
+    }
+    
+    template <typename T>
+    void updateElementsT(const T* element, size_t count, size_t begin, bool defer = true)
```
minggo, line 189: Do we need the post-fix `T`? Because other template functions don't have the post-fix.

mannewalis, line 189: It's to disambiguate vs the non template versions. Let me see if I can figure out why it was ambiguous, the template versions should just look like overloads.

cocos/renderer/CCBatchCommand.cpp
mannewalis, line 132: It's unsigned int. Unsigned is short for unsigned int, no need to specify the int. The compiler picks the optimal unsigned type.

minggo, line 132: I think it is different on 32-bit and 64-bit machines, and it may different from different compilers.
Right?

mannewalis, line 132: yes it always picks the optimal type. It will be whatever the native register size is on the cpu. It's part of the C standard, so it's guaranteed cross compiler.

cocos/renderer/CCVertexIndexBuffer.h
minggo, line 189: But other template functions in cocos2d-x don't obey this rule.

darkdukey:
[ci rebuild]

cocos/renderer/CCVertexIndexBuffer.h
mannewalis, line 189: Yeah I know... I will fix it. Right now I just want to get it into v4

PR #10359, opened 2015-02-04
getAttachNode should failed when there is no bone with name

flamefox:
when use spr->getAttachNode("bone"),  if the bone is not exist, the attachnode should return nullptr that user can notice the _skeleton is not contain that bone.


darkdukey:
@super626 Can you take a look?

super626:
Thanks @flamefox 

@darkdukey , @minggo 
That is ok.

darkdukey:
[ci rebuild]

PR #10362, opened 2015-02-04
add project function to camera

super626:
project 3d space position to screen position


super626:
[ci rebuild]

cocos/2d/CCCamera.h
```diff
@@ -128,9 +128,22 @@ class CC_DLL Camera :public Node
...
-    * Convert the specified point of viewport from screenspace coordinate into the worldspace coordinate.
-    */
-    void unproject(const Size& viewport, Vec3* src, Vec3* dst) const;
+     * convert the specified point of viewport from world-space coordinates into the screen-space coordinates.
+     *
+     * @param viewport The viewport size to use.
+     * @param src The world-space position.
+     * @param dst The screen-space position.
+     */
+    void project(const Size& viewport, const Vec3* src, Vec2* dst) const;
```
darkdukey: Can you make it return the dst instead of pass it in like that?
simular to `Node::convertToWorldSpace`
`Vec2 project(const Vec3& worldCoords) const;`

Why we have to specify the viewport size, camera should know it's viewport size right?

cocos/2d/CCCamera.h
```diff
@@ -128,9 +128,22 @@ class CC_DLL Camera :public Node
...
+    
+    /**
+     * Convert the specified point of viewport from screen-space coordinate into the world-space coordinate.
+     *
+     * @param viewport The viewport size to use.
+     * @param src The screen-space position.
+     * @param dst The world-space position.
+     */
+    void unproject(const Size& viewport, const Vec3* src, Vec3* dst) const;
```
darkdukey, line 154: Can you do the same for the unproject? Probably have to deprecate the code method.

`Vec3 unproject(const Vec3& screenCoords) const`

cocos/2d/CCCamera.cpp
```diff
@@ -238,9 +238,27 @@ bool Camera::initOrthographic(float zoomX, float zoomY, float nearPlane, float f
     return true;
 }
 
-void Camera::unproject(const Size& viewport, Vec3* src, Vec3* dst) const
+void Camera::project(const Size& viewport, const Vec3* src, Vec2* dst) const
 {
-    assert(dst);
+    GP_ASSERT(src);
```
darkdukey: Also can we use CCASSERT, and provide correct warning messages, GP_ASSERT just a CCASSERT with out message, why are we using that?

darkdukey:
[ci rebuild]

cocos/2d/CCCamera.h
super626: Camera does not know the viewport size currently

darkdukey: Can we get it internally such as `Director->getWinSize()`?

darkdukey: How about having a default value for viewport?

darkdukey:
[ci rebuild]

super626:
[ci rebuild]

zilluss:
I think the projected Y-Coordinates from the ndcY should be calculated like this:
screenPos.y = (ndcY + 1.0f) \* 0.5f \* viewport.height;

Otherwise the projected coordinate will be mirrored along the Y-axis.

zilongshanren:
@super626  This PR is duplicated with https://github.com/cocos2d/cocos2d-x/pull/12083 ?

zilongshanren:
@super626  Any updates?

PR #10376, opened 2015-02-05
addSpriteFramesAsync be not defined

ts0307:
remove parameter named callback,because addSpriteFramesAsync be not defined in SpriteFrameCache


darkdukey:
@dualface Can you take a look?

darkdukey:
Can you also send this PR for v3 branch? Because v4 will not be released for a couple month.

minggo:

PR #10383, opened 2015-02-05
Replace deprecated API/Class with new API/Class in engine code and test cases.

WenhaiLin:


cocos/base/ccUTF8.cpp
```diff
@@ -31,6 +31,27 @@
...
+        
+    char* buf = (char*)malloc(CC_MAX_STRING_LENGTH);
+    if (buf != nullptr)
+    {
+        vsnprintf(buf, CC_MAX_STRING_LENGTH, format, ap);
+        ret = buf;
+        free(buf);
+    }
+    va_end(ap);
+
```
minggo: Please undefine here.

cocos/editor-support/cocosbuilder/CCNodeLoader.cpp
```diff
@@ -96,14 +100,13 @@ void NodeLoader::parseProperties(Node * pNode, Node * pParent, CCBReader * ccbRe
         }
         else if (isExtraProp && pNode == ccbReader->getAnimationManager()->getRootNode())
         {
-            __Array *extraPropsNames = static_cast<__Array*>(pNode->getUserObject());
+            auto extraPropsNames = (__StringArray*)(pNode->getUserObject());
             if (! extraPropsNames)
             {
-                extraPropsNames = Array::create();
```
zilongshanren, line 102: The new statement will cause memory leaks.

cocos/scripting/lua-bindings/manual/cocos2d/lua_cocos2dx_manual.cpp
```diff
@@ -32,6 +32,36 @@
 #include <sys/socket.h>
 #endif
 
+class __TouchListenerHandler : public Map<const char*, Ref*>, public Ref
```
zilongshanren: We should avoid multiple inheritance. It is very weird.

cocos/editor-support/cocosbuilder/CCNodeLoader.cpp
```diff
@@ -11,6 +11,12 @@ using namespace std;
 using namespace cocos2d;
 using namespace cocos2d::extension;
 
+class __StringArray : public Ref
```
minggo: __StringArray -> StringArray

minggo: And it is better to put it into a anonymous namespace.

Issue #9165, opened 2014-11-19
cocos2d-x label position is not right when upgrade from v3.1 to v3.3rc0 in iOS


richieyan:
version: cocos2d-x 3.3 rc0
issue: label position is a little upwards when upgrade from v3.1 to v3.3rc0 in iOS but android is normal.

before upgrade:
![image](https://cloud.githubusercontent.com/assets/839363/5101635/6a36e752-6ff7-11e4-9498-ccb04eec0ad4.png)
after upgrade:
![image](https://cloud.githubusercontent.com/assets/839363/5101637/76c1bb32-6ff7-11e4-8246-2d2d3de623a0.png)

Font file for the label is SourceHanSansK-Bold.ttf

minggo:
Could you please paste some codes for this issue?
Thanks.

minggo:
And could you share the .ttf file.

richieyan:
code like this:

``` cpp
    static Label* newLabel(const std::string& text, const std::string& font, float fontSize){
        return Label::createWithSystemFont(text,font,fontSize);
    }
   ...
   //somewhere
    _energyLabel = wt::newLabel(StringUtils::format("%d/30",energy),rs::s_font_name,24);
    _energyLabel->setPosition(Vec2(100,300));
    view->addChild(_energyLabel);
```

font file: http://yun.baidu.com/s/1sjkCORr

yasuharu519:
I got same issue as here. Is that already fixed? @minggo

maginemu:
I'm in the same situation too.  It occurs only on ios Device.  but ios-simulator is normal.

PR #10394, opened 2015-02-06
Fixed the bug that texture size of string has unexpected padding on iOS7 later. 

yasuharu519:
Since iOS7, `sizeWithFont:constrainedToSize:` function is deprecated, so `boundingRectWithSize:options:attributes:context` is used for iOS version higher than.
In the current code, `NSStringDrawingUsesLineFragmentOrigin` and
`NSStringDrawingUsesFontLeading` are passed as an option. However the result texture has unexpected padding on the bottom. The result size is differ from the size which actually strings are rendered.

I think the issue #9165 is caused by this bug. `initWithString` function on cocos2d::Texture2D is used on the cocos2d::Label class when system font is used. Therefore, I think this problem is serious one.

I tested same code and compared result texture on difference versions of iOS. These are tested on simulator.

This is the result on iOS6.1.
iphone retina ios6 1 ios simulator screen shot 2015 02 06 12 09 32

This is the result on iOS7.1
iphone5s 7 1 ios simulator screen shot 2015 02 06 11 38 04

And this is the result on iOS8.1
iphone5s 8 1 ios simulator screen shot 2015 02 06 11 38 35

The tested code is following,

``` cpp
#include "HelloWorldScene.h"

USING_NS_CC;

Scene* HelloWorld::createScene()
{
    auto scene = Scene::create();
    auto layer = HelloWorld::create();
    scene->addChild(layer);
    return scene;
}

bool HelloWorld::init()
{
    if ( !Layer::init() )
    {
        return false;
    }

    Size visibleSize = Director::getInstance()->getVisibleSize();
    Vec2 origin = Director::getInstance()->getVisibleOrigin();

    std::string testString =
    "11111111111111111111111111111111111111111"
    "11111111111111111111111111111111111111111"
    "11111111111111111111111111111111111111111"
    "11111111111111111111111111111111111111111"
    ;

    FontDefinition fontdef;
    fontdef._fontName = "HiraKakuProN-W6";
    fontdef._fontSize = 25;
    fontdef._alignment = TextHAlignment::LEFT;
    fontdef._vertAlignment = TextVAlignment::CENTER;
    fontdef._dimensions = Size{500, 0};

    auto texture = new (std::nothrow) Texture2D;
    texture->initWithString(testString.c_str(), fontdef);

    auto textSprite = cocos2d::Sprite::createWithTexture(texture);
    textSprite->setAnchorPoint(cocos2d::Vec2::ANCHOR_BOTTOM_LEFT);
    texture->release();

    auto texSize = textSprite->getContentSize();
    auto background = cocos2d::Sprite::create();
    background->setColor(cocos2d::Color3B::RED);
    background->setTextureRect(cocos2d::Rect(0, 0, texSize.width, texSize.height));
    background->setPosition(visibleSize.width / 2, visibleSize.height / 2);

    background->addChild(textSprite);
    this->addChild(background);

    return true;
}
```

After changed to the code I requested, the result will be as following,
iphone5s 8 1 fixed
This is the result on iOS8.1

### DIFF
--- a/cocos/platform/ios/CCDevice-ios.mm
+++ b/cocos/platform/ios/CCDevice-ios.mm
@@ -226,7 +226,7 @@ static CGSize _calculateStringSize(NSString *str, id font, CGSize *constrainSize
     CGSize dim;
     if(s_isIOS7OrHigher){
         NSDictionary *attibutes = @{NSFontAttributeName:font};
-        dim = [str boundingRectWithSize:textRect options:(NSStringDrawingOptions)(NSStringDrawingUsesLineFragmentOrigin|NSStringDrawingUsesFontLeading) attributes:attibutes context:nil].size;
+        dim = [str boundingRectWithSize:textRect options:(NSStringDrawingOptions)(NSStringDrawingUsesLineFragmentOrigin) attributes:attibutes context:nil].size;
     }
     else {
         dim = [str sizeWithFont:font constrainedToSize:textRect];


WenhaiLin:
@yasuharu519 Thanks very much.

minggo:
Thanks.

yasuharu519:
Thanks for merging, is that available on v4.0?

minggo:
@yasuharu519 
Yep, it is available in first v4.0 version, may be v4.0alpha0.

yasuharu519:
@minggo Ok, I got that. I will wait for that comes out. Thanks for working.

PR #10395, opened 2015-02-06
updating for tizen

sukwonsuh:
updating for tizen based on cocos2d-x.3.4


minggo:
@sukwonsuh 
Please send to v3.
Thanks.

PR #10403, opened 2015-02-06
a better CSLoader::bindCallback implementing

ts0307:
This is the method using binding callback at http://www.cocoachina.com/bbs/read.php?tid=274210
It has five steps,

but using my implementing,It can be reduced to a 2 steps,

in other words,Only the second and the fifth step is necessary.

Of course you must  implement this interface cocostudio::WidgetCallBackHandlerProtocol.

I updated the test CustomWidgetCallbackBindTest

notice you must clear the class name of "CustomRootNode" in cocosui/CustomTest/CustomWidgetCallbackBindTest/CustomWidgetCallbackBindTest.csb


ts0307:
in this way,you don't need to create class of CustomRootNode and CustomRootNodeReader, and call registReaderObject,you only need to create a class which inherit from WidgetCallBackHandlerProtocol and implements it's interface.the last,you need to add a parameter of cocos2d::Node\* deleget to CSLoader::createNode as this CSLoader::createNode(const std::string& filename, cocos2d::Node\* deleget).now you can call createNode with parameters csb's path and point to object of class which implementing WidgetCallBackHandlerProtocol above,
this all.

darkdukey:
@chengstory I think this solution is better than the one we currently have.

darkdukey:
[ci rebuild]

darkdukey:
[ci rebuild]

minggo:

PR #10431, opened 2015-02-10
Fix win32 resource search path to be relative to the executable.

vovkasm:
See forum topic: [Is set the default resource search path to current directory on win32 is desirable behavior?](http://discuss.cocos2d-x.org/t/is-set-the-default-resource-search-path-to-current-directory-on-win32-is-desirable-behavior/19835)

In short that patch do:
1. Set the default resource search path on win32 to directory where the executable is.
2. `build/cocos2d-win32.vc2012.sln` now builds executables in subfolders of the main output path, for ex. `build/Debug.win32/cpp-tests` and copy all needed files to that subfolders. These executables now self-contained and can be run from any place.

It is a risk that copying dlls and resource files to subfolders will slowdown the build process, but I use parameter `/D` for `xcopy` command in the projects `PreBuild` event. And the actual copy will be happened only for modified files, so slowdown will be minimal.

This patch against `v3` branch, I will be glad to make distinct pull request against `v4-develop` if this will be needed.

### DIFF
--- a/cocos/platform/win32/CCFileUtils-win32.cpp
+++ b/cocos/platform/win32/CCFileUtils-win32.cpp
@@ -29,6 +29,7 @@ THE SOFTWARE.
 #include "CCFileUtils-win32.h"
 #include "platform/CCCommon.h"
 #include <Shlobj.h>
+#include <cstdlib>
 
 using namespace std;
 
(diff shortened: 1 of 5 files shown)


darkdukey:
[ci rebuild]

darkdukey:
[ci rebuild]

darkdukey:
I tested this, before the patch, run cpp-test.exe from other dir will crash on win32
After this patch it runs fine.

minggo:
@vovkasm 
What's the progress of fixing template issue?

vovkasm:
Oh, I didn't have enough free time... Still in progress. I plan to send pull request tomorrow.

vovkasm:
Done. Pull request #10784

vovkasm:
I would be very glad if this changes will be landed in `v4` also.

PR #10440, opened 2015-02-11
Audio can not resume

WenhaiLin:


minggo:
Merged into v3 #10467.

PR #10452, opened 2015-02-11
fix SpriteEx

ntotani:
`onComplete` should be in args


darkdukey:
Thanks, can you also send this for v3 branch?

minggo:
@dualface 
Please take a look.

dualface:
Thanks @ntotani

PR #10461, opened 2015-02-12
fix SpriteEx

ntotani:
`onComplete` should be in args

see alos
https://github.com/cocos2d/cocos2d-x/pull/10452


dualface:
thanks

PR #10483, opened 2015-02-12
apply custom request header for every request type (ios)

redism:
I think custom request header should be set not only for POST or PUT, but also every http methods.


zilongshanren:
[ci rebuild]

samuele3hu:
@redism , I agree with you that every http methods should set the request header,Thanks.

zilongshanren:
[ci rebuild]

Mazyod:
This is great, I was thinking of submitting a PR for that myself.

PR #10500, opened 2015-02-13
Removed CC_PLATFORM_WP8 from v4

stammen:
This pull request does the following:
1. Removes all code for CC_PLATFORM_WP8
2. Removes WP8 Shader Compiler code and support for precompiled shaders
3. Removes cpp template for WP8
4. Fixes broken audio build for winrt Universal App caused by pull request https://github.com/cocos2d/cocos2d-x/pull/10426
5. renames libcocos2d from 3.4 to 4.0


minggo:
Thanks @stammen .

PR #10502, opened 2015-02-13
FEATURE 8766

Sheado:
Added the ability to hide or show the cursor on desktop builds.

### DIFF
--- a/cocos/base/CCController.cpp
+++ b/cocos/base/CCController.cpp
@@ -25,7 +25,7 @@
 
 #include "base/CCController.h"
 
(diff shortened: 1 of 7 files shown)


darkdukey:
[ci rebuild]

darkdukey:
[ci rebuild]

darkdukey:
[ci rebuild]

minggo:
But we can not add a function that only takes effect on some platforms.
@ricardoquesada 
What do you think?

darkdukey:
I think it's okey, for example desktop don't have accelerometer support. But mobile has them.

we probably need to remove controller code from this PR, since it's been working on #10503

darkdukey:
I clean up the code and create test case at #10720
Close this PR to avoid confusion. Credit goes to @Sheado

zilongshanren:
[ci rebuild]

PR #10520, opened 2015-02-17
adding CC_DEPRECATE(version) macro

mannewalis:


mannewalis:
This also passes except for the windows build which fails with a script error.

FATAL: Unable to delete script file C:\Users****\AppData\Local\Temp\hudson1414863680372348985.py

PR #10521, opened 2015-02-18
Fixed HttpClient-ios request to work when the data is started by 0x00.

ryule:
Fixed a bug in HttpClient-ios.mm.
If the request data is started by a null character, it does not fill http body because of using strlen that leads to a wrong condition check.


darkdukey:
[ci rebuild]

samuele3hu:
It is really a bug, thanks to @ryule. @minggo , this pr can be merged.

PR #10531, opened 2015-02-19
Ewells/disable cocos2dx warnings

Number10Ox:


darkdukey:
Is this a patch for cocos2d-x 2.x?

PR #10535, opened 2015-02-20
V4 label cleanup

mannewalis:
Pare down LabelTTF and LabelBMFont which are deprecated now, and rather than contain the new Label class, inherit directly and fix any API issues. 
Remove the .cpp implementations.
Now a lot less code to maintain.


mannewalis:
Despite the long list of commits, this change has been reduced to just the Label changes.

darkdukey:
Seem like it's breaking the cmake build, can you try it?

mannewalis:
I will revisit this. The cmake fixes are easy, but the bindings break in cmake and this change is thus a breaking change. Hopefully we can remove CCLabelBMFont and CCLabelTTF in 4.0

PR #10539, opened 2015-02-21
Source refactor. Abstracted common functionalities from 4 onTouches meth...

krishnanm86:
...ods in Camera3dTest

### DIFF
--- a/tests/cpp-tests/Classes/Camera3DTest/Camera3DTest.cpp
+++ b/tests/cpp-tests/Classes/Camera3DTest/Camera3DTest.cpp
@@ -709,11 +709,8 @@ void Camera3DTestDemo::updateCamera(float fDelta)
         }
     }
 }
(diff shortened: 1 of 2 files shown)


krishnanm86:
Passes all tests and removes redundant code.  Cam this be accepted?

mannewalis:
Hi, was looking at this and I think you do not need a template type here, you can just make the method onTouchesCommon take a bool, since all the places you call it you are passing a bool. 

Also, you just just return the true/false... setting touchProperty to true does nothing since it was passed in by value. :)

krishnanm86:
Hi,
   So how do you propose I resolve this? 

Would it help if I just changed the argument to bool pointer and actually passed in a pointer to the corresponding touch property that is being set?

mannewalis:
Yeah that looks better :)

krishnanm86:
Nice. So can this be merged now?

minggo:
We will merge it when jenkins passes.

PR #10542, opened 2015-02-22
CocosStudio Sprite3D Reader

sun-ray:
CocosStuidio Sprite3DReader pull request


minggo:
Please add lua-binding and add lua test cases.

PR #10543, opened 2015-02-22
Fix an issue with iOS HTTP Request code

Mazyod:
Background:
When the server returns a response code greater than 300, it may return a JSON or some data to explain the error.
I have made sure this is a standard by asking on StackOverflow: http://stackoverflow.com/q/28662167/456434

Thanks,
Maz

### DIFF
--- a/cocos/network/HttpAsynConnection.m
+++ b/cocos/network/HttpAsynConnection.m
@@ -88,11 +88,10 @@
     | “206”  ; Partial Content
     */
     if (responseCode < 200 || responseCode >= 300)
-    {// something went wrong, abort the whole thing
-        
-        [connection cancel];
-        finish = true;
-        return;
+    {// something went wrong, record a response error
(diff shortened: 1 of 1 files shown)


minggo:
But the `if` sentence includes response code that is less than 200.

Mazyod:
@minggo thanks, yes you're right.
Don't you think this is another problem that can be solved in a different PR? Even 3xx response isn't really an error, it's a redirection, and is usually handled transparently.

Mazyod:
What I mean to say, the code was terminating the connection for that if statement. This PR is just making sure we don't terminate at all, but try to see the server response.

So, 1xx response was terminating, but not anymore with this PR.

samuele3hu:
@Mazyod , I think you are right, we can't cancel the connect when responseCode < 200 or responseCode >= 300.@minggo, this pr could be merged.

PR #10545, opened 2015-02-23
Fix major memory leaks in iOS http request

Mazyod:
... and other small fixes
### Brief:
1. In iOS, whenever you have a new spawned thread with its own run loop (i.e. while true), you must wrap it with an autorelease pool, so it cleans up the objects at the end of each loop cycle.
2. There were a few "alloc" and "new" objects that weren't released.
3. The HttpAsync class itself was missing dealloc, also property assignment should use `self.property` so the previous value gets released as appropriate.
4. Resolved a small warning

**NOTE**: I have a cocos2d-x game that heavily relies on the HTTP requests, with cookies, but without SSL, so I couldn't test the SSL.
I haven't tested the cookies and SSL yet, but for starters, these improvements are meant to fix the memory issues only, not the logic.

I would also like to improve this code more, make is iOS and Mac friendly, and follow Objective-C conventions.

### DIFF
--- a/cocos/network/HttpAsynConnection.h
+++ b/cocos/network/HttpAsynConnection.h
@@ -36,7 +36,7 @@
 
 @property (strong) NSString *sslFile;
 
-@property (strong) NSDictionary *responseHeader;
+@property (copy) NSDictionary *responseHeader;
 
 @property (strong) NSMutableData *responseData;
 
--- a/cocos/network/HttpAsynConnection.m
+++ b/cocos/network/HttpAsynConnection.m
@@ -63,14 +63,14 @@
     NSLog(@"Starting to load %@", srcURL);
     finish = false;
 
-    self.responseData = [NSMutableData new];
+    self.responseData = [NSMutableData data];
     getDataTime = 0;
     self.responseError = nil;
     
     // create the connection with the target request and this class as the delegate
-    self.conn = [[NSURLConnection alloc] initWithRequest:request
-                                                delegate:self
-                                        startImmediately:NO];
+    self.conn = [[[NSURLConnection alloc] initWithRequest:request
+                                                 delegate:self
+                                         startImmediately:NO] autorelease];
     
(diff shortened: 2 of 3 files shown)


Mazyod:
Please note, I added `@autorelease` right after the `while (true)` in order to minimize the diff size.

minggo:
Please fix conflict.

Mazyod:
Ok, will do that soon.

Mazyod:
@minggo Done

Mazyod:
I went over the code one more time, and made sure the retain/release balances out. Do you guys find problems working with ARC for Objective-C?

darkdukey:
[ci rebuild]

minggo:
We don't use ARC in our code. I think it is better not use it if possible.

Mazyod:
@minggo Actually, in terms of maintainability **and** performance, ARC is better.

**Maintainability:** Best code is no code. ARC obvioulsy remove tons of code.
**Performance:** ARC inlines retain/release as needed, instead of programmers relying too much on autorelease objects, which have a slightly higher overhead. Example:

``` objc
// non-arc
self.property = [NSString string]; // autoreleased
// arc
self.property = [NSString new]; // retain/release, more effecient
```

There are edge cases where ARC is not the way to go, and you can flag those files as non-ARC, as needed. Also, the way I see it is, in cocos2d-x specifically, non-ARC code will probably be better written as C++ code in the first place, and leave the ARC/ObjC code for native features, which play well with ARC.

minggo:
Thanks @Mazyod .
Could you please send a pull request to v4-develop too?

Mazyod:
Thanks @minggo, will do.

Mazyod:
@minggo Also the other PR or not? https://github.com/cocos2d/cocos2d-x/pull/10543

minggo:
We should include it if we decide to merge it.
Thanks.

minggo:
I will ask your help when we decide to merge #10543.

PR #10550, opened 2015-02-23
merge cocos2d/v3 -> cocos2d/v4-develop

mannewalis:


minggo:
@mannewalis 
We can not just synchronize v3 to v4 because there are some codes we don't need in v4, such as cocos studio reader related.

We have to synchronize codes by commits, and i have marked all pull requests that need to synchronize to v4.

mannewalis:
ok got it. Are you going to be handling sync between branches going forward? If so I will just work against v4-develop and leave v3 alone.

minggo:
Yep, i will sync between branches.

PR #10566, opened 2015-02-24
Removed no_gl_PointSize shaders for Windows Phone 8.1

stammen:
This pull request removes the special case shaders for Window Phone 8.1. The latest version of Angle now supports gl_PointSize correctly.


stammen:
@minggo please update download-deps.py so the new Angle dll's get downloaded

Thanks!

minggo:
@stammen 
We will update it in these two days as we will modify it soon.
As you know, the speed is slow here, so we want to do less updating as possible.

stammen:
Sure. But with the removed WP8 and old WinRT DLL's the download will be a lot smaller.

minggo:
Yep, the package size will be smaller.
As we are using prebuilt version of all 3rd party libraries, it will be larger again.

stammen:
@minggo We are working to remove png, jpeg, and tiff libs from Windows 8.1 Store and Phone external dependencies by using the Windows Imaging Component (WIC). We are also planning on implementing native networking code to remove curl and openssl. This will really help reduce the size of the download, especially when we add Windows 10 (Visual Studio 2015) support in the coming months.

minggo:
Sounds great.

PR #10580, opened 2015-02-25
CocosStudio 3DReader

sun-ray:
The UserCameraReader and fix commit


minggo:
Please add studio related resources in `ccs-res` submodule.
Please add lua test case.

sun-ray:
fix it, please check it again

PR #10629, opened 2015-02-27
V4 winrt fixes

stammen:
This pull request removes the Link Time Code Generation from winrt 8.1 libspline project to speed up release builds.

Added external/fbx-conv to .gitignore


minggo:
@stammen 
Does win32 project can use this method to optimize release builds?

zilongshanren:
@minggo 
I think Win32 project has already taken this approach.

stammen:
It appears that the Win32 projects already have this fix.

minggo:
Ok, thanks.

PR #10636, opened 2015-02-27
Fixed the position of physics body is wrong when the position of parent node has been change.

WenhaiLin:


minggo:
Please send a pull request to v3 too.

PR #10671, opened 2015-02-28
enable setWritablePath on platform win32 and mac

kpkhxlgy0:


cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -337,6 +337,11 @@ static void addObjectToNSDict(const std::string& key, const Value& value, NSMuta
 
 std::string FileUtilsApple::getWritablePath() const
 {
+    if (_writablePath.length())
```
WenhaiLin, line 340: I suggest you determines whether or not the string is not empty like this:

``` cpp
if ( !_writablePath.empty())
```

zilongshanren:
[ci rebuild]

minggo:
@kpkhxlgy0 
I don't find where to enable setWritablePath.

kpkhxlgy0:
@minggo 
I call setWritablePath, but there's no effect on mac and win32, the getWritablePath() always return the same path.

minggo:
@kpkhxlgy0 
Got it thanks.

minggo:
[ci rebuild]

PR #10681, opened 2015-03-01
Let Mac project use ObjC HTTP client

Mazyod:
Removed HTTPClient.cpp from the cocos2d_libs.xcodeproj, and made the Mac library compile the ObjC HTTP implementation, to be consistent with iOS.

**NOTE**:
This is a breaking change! 
The Mac app will not compile properly until you link against the Security.framework, but I think it's necessary, since iOS should be consistent with Mac as much as possible.


minggo:
Some questions:
- i am not sure if it can be used on Mac or not, have you tried it?
- we should rename `HttpClient-ios.mm` to `HttpClient-apple.mm` if it can work
- should change template too

Mazyod:
@minggo Thanks for the reply.
1. Yes, it only uses the `Foundation` and `Security` frameworks, so it can be used with Mac OS. We can always run the tests to make sure, as well.
2. I thought about renaming the files, but wasn't sure about the convention. I will rename it to `apple`.
3. I will change the template to include these files, and add the `Security` framework for Mac.

I hope that works for you

zilongshanren:
[ci rebuild]

minggo:
Thanks @Mazyod.

minggo:
[ci rebuild]

zilongshanren:
[ci rebuild]

minggo:
@Mazyod 
Will you fix conflict and modify template too?

Mazyod:
@minggo yes, sorry I am just not finding time to do that.
Will hopefully finish it up today.

Mazyod:
Sorry about the delay guys, but it's not worth fixing merge conflict. I created a new branch, and will submit a new PR

PR #10683, opened 2015-03-01
Fix editbox on Mac

Mazyod:
**Brief**:

Editbox on iOS removes the native textfield in `closeKeyboard`, and the adds the textfield again in `openKeyboard`:
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/ui/UIEditBox/UIEditBoxImpl-ios.mm#L121

But, in Mac OS, the editbox is removed in `closeKeyboard`, and never added back in `openKeyboard`. This causes the textfield to disappear once it's removed from the scene, even if you try to add it back in:
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/ui/UIEditBox/UIEditBoxImpl-mac.mm#L184

This PR fixes this issue, by tracking which textfield is being used, then adds the textfield back to the view hierarchy as appropriate.

Tested on my game.

Thanks,
Maz

### DIFF
--- a/cocos/ui/UIEditBox/UIEditBoxImpl-mac.h
+++ b/cocos/ui/UIEditBox/UIEditBoxImpl-mac.h
@@ -50,6 +50,7 @@
 @property(nonatomic, retain) NSMutableDictionary* placeholderAttributes;
 @property(nonatomic, readonly, getter = isEditState) BOOL editState;
 @property(nonatomic, assign) void* editBox;
+@property(nonatomic, assign, getter = isSecure) BOOL secure;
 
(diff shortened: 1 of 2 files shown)


zilongshanren:
[ci rebuild]

Mazyod:
rebuild?

WenhaiLin:
[ci rebuild]

minggo:
@Mazyod 
How to reproduce it issue?
I will try it.
Thanks.

Mazyod:
@minggo in my game, I add the edit box as a child of a node, then remove that node from the hierarchy and add it back in.

I think the same issue happens if you set the parent node visibility to false then true again.

minggo:
Thanks.

PR #10684, opened 2015-03-01
Template compiles in OSX using cmake. Template adds "Resource" to search...

sean-sh:
cpp-template-default now works in OS X


zilongshanren:
[ci rebuild]

darkdukey:
[ci rebuild]

darkdukey:
[ci rebuild]

minggo:
cmake is not ready for all platforms, and it has conflict with Xcode projects(if you add/remove a file through Xcode project, cmake doesn't know it). So i close this PR

PR #10692, opened 2015-03-02
If mesh has normals, use them

babcca:
Fix issue when create mesh without texture but with normals vectors and add light. Currently you only see black mesh without shades, it's because light need normals.


babcca:
[ci rebuild]

zilongshanren:
[ci rebuild]

minggo:
@super626 
Could you please take a look?

super626:
@babcca 
Thanks
Do you mean a user defined shader?

babcca:
Not custom shader but cocos SHADER_3D_POSITION_NORMAL. When you create scene with Light and add Sprite3D where model has normals, then you currently see only black model. But when you have normals and use SHADER_3D_POSITION_NORMAL shader then you can see light.

**With GLProgram::SHADER_3D_POSITION_NORMAL**
With GLProgram::SHADER_3D_POSITION_NORMAL

**With GLProgram::SHADER_3D_POSITION**
image

super626:
@babcca Thanks.

Can you show me the code snip of adding 3d sprite and lights to the scene?

There may be also some problems on the judgement of using light. The variable `usingLight` should be true if we find there are lights in the scene.

babcca:
Hi, I added small test for this issue. Withou this patch (and right shader) you can see this warning `cocos2d: warning: Attribute not found: a_normal`

super626:
@babcca Thanks.
What you said is correct, I have tested it.

There is some conflicts. I sent another PR.
https://github.com/cocos2d/cocos2d-x/pull/12777

pandamicro:
Since it will be merged to v3.7, @super626 have fix the conflict and improved the solution in https://github.com/cocos2d/cocos2d-x/pull/12777. I have migrated the test case in https://github.com/cocos2d/cocos2d-x/pull/12790. So I will close this Pull Request.

Thank you @babcca , you will be added to the AUTHORs

PR #10695, opened 2015-03-02
Fix Sprite3DCache crash/cleanup

uorbe001:
`Sprite3DCache` is not being cleaned up properly, so it is possible to create
an instance of it and if you then destroy the Director and try to create
a new one, the new one will use the same `Sprite3DCache` which has
dangling references to the old Director, causing a crash as soon as the
application starts (with the second Director) in this case.

This fixes it by destroying the Sprite3DCache when the director is
purged.

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -60,6 +60,7 @@ THE SOFTWARE.
 #include "base/CCAsyncTaskPool.h"
 #include "platform/CCApplication.h"
 //#include "platform/CCGLViewImpl.h"
+#include "3d/CCSprite3D.h"
 
 #if CC_ENABLE_SCRIPT_BINDING
 #include "CCScriptSupport.h"
@@ -996,12 +997,13 @@ void Director::reset()
     GLProgramStateCache::destroyInstance();
     FileUtils::destroyInstance();
     AsyncTaskPool::destoryInstance();
-    
+    Sprite3DCache::destroyInstance();
+
     // cocos2d-x specific data structures
     UserDefault::destroyInstance();
-    
+
     GL::invalidateStateCache();
-    
+
     destroyTextureCache();
 }
 


super626:
Great
Thanks

super626:
@uorbe001 
`Sprite3DCache` does not be cleaned properly.

But I still do not understand why it will cause crash.  `Sprite3DCache` does not refer to Director. It only contains `SpriteData` which has the data ref using to create `Sprite3D`.

So can you show me a test case for it.

uorbe001:
@super626 It's not the Sprite3DCache itself that causes the crash. Some of the data cached in the Sprite3dCache keeps references to a `GLProgramState`, which in turn has references a `GLProgram`, which is saving a reference to `Director` as `_director`. I suppose another way of getting rid of the crash would be to remove that `_director` variable in the `GLProgram` and simply get the singleton instance every time that variable is referenced, but I haven't tested that.

The reason we are hitting this is because we are loading cocos2dx as part of a bigger application, so cocos2dx is only loaded in one of our views and when we exit that view, the Director gets destroyed. So a way of replicating this would be to create an empty scene with a Sprite3D, destroying the Director without actually killing the process/application and trying to recreate it.

super626:
@uorbe001  
I got it.
Thanks very much.

minggo:
[ci rebuild]

minggo:
[ci rebuild]

zilongshanren:
[ci rebuild]

minggo:
Please fix compiling error on Android

> /android_slave/workspace/ccs-pr/node/android/tests/cpp-empty-test/proj.android/../../../cocos/./base/CCDirector.cpp:1000: error: undefined reference to 'cocos2d::Sprite3DCache::destroyInstance()'
> collect2: error: ld returned 1 exit status
> make: **\* [obj/local/armeabi/libcpp_empty_test.so] Error 1

minggo:
@uorbe001 
Could you please fix the compiling error?

uorbe001:
@minggo I'm sorry, I've never built cocos on android. I'm trying to find the issue, but I can't see it, can you please give me any pointers on how to debug this issue? The error sounds like a missing source file to me, but I can't make sense of it, I didn't need to change the sources on the ios project, I'm surprised android would require me to.

minggo:
I have looked into the codes. It is strange why there is a problem.
I will rebuild it again.

minggo:
[ci rebuild]

uorbe001:
@minggo I can confirm I am getting that same error compiling (android) on my machine, I just don't understand why. It sounds like the android project isn't including the Sprite3D sources to me, but then I'd expect it to stop compiles on master too.

zilongshanren:
error: undefined reference to 'cocos2d::Sprite3DCache::destroyInstance()'
make: **\* [obj/local/armeabi/libcpp_empty_test.so] Error 1  [ci rebuild]

super626:
@uorbe001 

Please include Sprite3D.h in the Director.cpp

minggo:
I think @uorbe001 has already included it.

uorbe001:
@super626 yeah, it is included (cocos/base/CCDirector.cpp#63). There must be something really silly I'm missing, just can't see it.

zilongshanren:
@uorbe001 
Any  updates?  This PR is conflicts with the upstream.

uorbe001:
@andyque I haven't been able to figure out why it's not compiling on android, it might be safe to close it if it isn't affecting anyone else

zilongshanren:
@uorbe001  ok, thanks.

PR #10712, opened 2015-03-03
V3 WP8/Universal App fixes

stammen:
This pull request included the following
1. Removes old .h files that no longer exist from the projects to avoid unnecessary rebuilds
2. Updated cocos2d dll name from 3.4 to 3.5
3. Disabled Check SDL setting to enable CC_DEPRECATED as warning and not error
4. Disabled LTCG to speed up release builds


stammen:
@minggo This pull request also needs 

https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/104

and download-deps version updated.

PR #10735, opened 2015-03-05
Ability to change default AudioEngine mode(actual for iOS)

newnon:
SimpleAudioEngine::configure(AudioManagerMode mode)


minggo:
It is a special feature for iOS, we can not add it in common header file. Because other platforms will have their own special feature, if we add all these features like this, then the class will be too complex.

newnon:
Yes but this function is very important. Some games have music as a part of game play and they need MediaPlayback or SoloAmbient mode, some haven't music or they gameplay not depends on music and sound, they can set Ambient mode. Default mode in cocos is very strange it use solo ambient mode on start(but not on activate) second path also fix it and play ability checked every activation.
P.S Imho this is must have function for many iOS game.

WenhaiLin:
[ci rebuild]

minggo:
I have not idea how to fix elegant.
But current implementation is not good, because
- it only works on iOS
- the enum variable name is even the same as iOS

newnon:
Enum is good enough to convey the essence.  
For example Android have setAudioStreamType() and setStreamSolo() methods for same functionality

newnon:
As for me i think its good idea to change default mode to Ambient it is suitable for 95% games
Many users turn music off and listen their own music in background

minggo:
@WenhaiLin 
Could you please take a look of default value of the audio mode on iOS?

dumganhar:
> As for me i think its good idea to change default mode to Ambient it is suitable for 95% games
> Many users turn music off and listen their own music in background

@newnon , good catch. 
It's better to implement configure method for all platforms, at least Android should implement this since Android market is a big shark.

BTW, we will deprecate `SimpleAudioEngine` recently, it will be re-implemented by `AudioEngine`.
I totally agree that we need add a `configure` method in `AudioEngine` too.

Anyway, I will keep on working on the issues of Audio. I will let u know when I gonna get something update.

newnon:
unfortunately i haven't implemented this for other platforms

minggo:
I create an issue https://github.com/cocos2d/cocos2d-x/issues/16826 for it. So close this one since it is not completely.

PR #10737, opened 2015-03-05
blending function for Scale9Sprite

newnon:
BlendProtocol for Scale9Sprite


zilongshanren:
Thanks, @newnon 

PR #10739, opened 2015-03-05
WebP loading improvements

newnon:
WebP loaded as premultiplied alpha if it have alpha


newnon:
If we don't have alpha load image without alpha, if we have load image with premultiplied alpha because of unify with png that loads with premultiplied alpha.

newnon:
[ci rebuild[

newnon:
[ci rebuild]

PR #10740, opened 2015-03-05
Windows network path fix

newnon:
fix that allow cocos2d-x work with windows network paths ("//host/path")


minggo:
How can it be used?

newnon:
If you will run game written in cocos2d-x on windows from network, it can't find its resources because it newer will get full path, this patch fix full patch checking with network paths.

minggo:
Could you please give some codes to reproduce it?
Thanks.

newnon:
it is not code
copy any cocos application with its resources on any windows shared folder like \some_host\some_folder and run it you will get crash because application can't find it resources.

newnon:
[ci rebuild]

minggo:
Thanks.

PR #10744, opened 2015-03-05
Fixed stats labels are recreated 3 times during launch

WenhaiLin:


cocos/platform/CCGLView.cpp
```diff
@@ -156,7 +156,7 @@ void GLView::updateDesignResolutionSize()
         // reset director's member variables to fit visible rect
         auto director = Director::getInstance();
         director->_winSizeInPoints = getDesignResolutionSize();
-        director->createStatsLabel();
+        director->_recreateStatsLabel = true;
```
minggo: How about rename it to `_isStatusLabelUpdated`?

PR #10752, opened 2015-03-06
Fixes for mips support on Android

ricardoquesada:
Fixes for mips support on Android, in particular for Android on CI20


ricardoquesada:
[ci rebuild]

PR #10780, opened 2015-03-09
Fix CCDevice-ios.mm to have the right alpha color for a stroke effect.

ryule:
There is a typo in CCDevice-ios.mm while handling a stroke alpha color.


WenhaiLin:

minggo:
@ryule 
Could you please send a pull request to v4-develop too?
Because i can not use `git cherry-pick` to sync it easily.
Thanks.

ryule:
@minggo 
It seems that the same patch was already applied to just v4-develop branch by other pr #10441.

minggo:
Thanks.

PR #10784, opened 2015-03-09
Fix templates after change default resource search path on win32

vovkasm:
Fix templates after change default resource search path on win32.

Fix win32 projects to properly copy resources to the folder with executable in `cpp-template-default` and `lua-template-default` templates.

I don't fully understand how to deal with `lua-template-runtime` so don't fix it. If anyone can suggest how to build and run project generated with it, then I will fix it also...

This pull request is the answer for #10431 discussion.


minggo:
@samuele3hu 
Could you please take a look?

samuele3hu:
lua-template-runtime use the same project configure like lua-template-default. I have checked the project created by the lua-template-runtime,the results were ok both debug and release.

vovkasm:
Thank you!

Issue #10589, opened 2015-02-25
MoveTo::reverse() have no effect


fusijie:
`version:` 3.4Final
`issue:` MoveTo::reverse() have no effect.
`How to reproduce:` add following code to HelloWorld Project.

``` cpp
    MoveTo* x = MoveTo::create(1, Vec2(100,100));
    MoveBy* y = x->reverse();
    sprite->runAction(Sequence::create(x, y, nullptr));
```

`How to fix?`

``` cpp
    MoveTo* x = MoveTo::create(1, Vec2(100,100));
    x->startWithTarget(sprite);
    MoveBy* y = x->reverse();
    sprite->runAction(Sequence::create(x, y, nullptr));
```

startWithTarget(Node *target) initial the `_positionDelta` property. if you don't call the function, `y action` will be wrong.

minggo:
Because MoveTo is just to move to a absolute position, there is not meaning for MoveTo::reverse().

minggo:
After thinking more about it, i think it is a bug.

fusijie:
I also confused it a bug or not. Because if `MoveTo`'s reverse action is a `MoveTo` action, nobody know the d

fusijie:
I also confused it is a bug or not. Because if `MoveTo`'s reverse action is a `MoveTo` action too, nobody know the destination of the target(It has no target for now), such as,

``` cpp
    MoveTo* x = MoveTo::create(1, Vec2(100,100));
    MoveTo* y = x->reverse();//what's the destination?
```

But now, the reverse action of `MoveTo` is a `MoveBy` action( from base class). the implemention is 

``` cpp
MoveBy* MoveBy::reverse() const
{
    return MoveBy::create(_duration, -_positionDelta);
}
```

So the destination issue should be resolved by `_positionDelta`.  `startWithTarget()` is a internal API, should we need some other public API to bind target to Action for implementing the reverse action?

minggo:
I think MoveTo::reverse() should move to its original position.
What's your opinion?

fusijie:
Yes, MoveTo::reverse() should move to its original position.
But no matter to use MoveBy or MoveTo, we need a target so that we can get the original position from the target by `target->getPosition()`.

minggo:
Yep, what's your suggestion?

fusijie:
＝。＝

A.enable MoveTo::reverse()
add a new API to bind target and record the original position, and note the limits. It's not a good idea.  

[...]

PR #10789, opened 2015-03-10
disable MoveTo::reverse()

fusijie:


fusijie:
For the issue: https://github.com/cocos2d/cocos2d-x/issues/10589

cocos/2d/CCActionInterval.h
```diff
@@ -549,6 +549,7 @@ class CC_DLL MoveTo : public MoveBy
     // Overrides
     //
     virtual MoveTo* clone() const override;
+    virtual MoveTo* reverse(void) const  override;
```
minggo: Don't need void.

minggo:
[ci rebuild]

fusijie:
OK, I have removed the `void`.

PR #10792, opened 2015-03-10
fix touch event handling in LayerEx

ntotani:
Callback for registerScriptTouchHandler should return value to decide handling "began" event.


WenhaiLin:
[ci rebuild]

PR #10796, opened 2015-03-10
Http client mac

Mazyod:
This is another attempt for a previous PR #10681 
### Purpose:

Consistency between Mac and iOS in using the same HTTP client. Mac client now uses the ObjC client.
### Changes:
- Removed `HttpClient.cpp` from Mac target
- renamed `HttpClient-ios.mm` to `HttpClient-apple.mm`
- Added `HttpCookie`, `HttpAsyncConnection.m`, `HttpClient-apple.mm` to Mac target
- Mac target now needs to be linked against Security.framework, so I updated tests and templates for that
### Notes
- The HTTP tests are missing SSL and Cookie tests, I think, but the available tests all pass.
- The developers need to be aware of this change, and add the Security.framework to their mac target.


minggo:
[ci rebuild]

minggo:
[ci rebuild]

PR #10817, opened 2015-03-11
fix issue autorelease will not release trigger by autorelease.

garfieldkwong:


garfieldkwong:
I have found the problem in autoreleasepool. If i call an object's (A) autorelease in another's(B) destructor triggered by B's autorelease. The A will nerver relase.
The reason is the B is destructed by calling release in AutoReleasePool::clear(). Autorelease function called in for loop.
It mean that the A will insert into _managedObjectArray in for loop of _managedObjectArray. So A will not be reached.
And then _managedObjectArray will be cleared after the for loop. So A object's release will not call.

minggo:
Yep. It is a bug in your situation, some ideas:
- i think it is better to invoke `release` instead of `autorelease` in this situation
- How much cost by invoking `swap`? Because it is invoked every frame.

garfieldkwong:
1. There is no choice in some case.
2. In c++ spec, swap is the CONSTANT time operation. It suppose is the fast operation. I believe that the implementation of it is swap pointer only.

minggo:
Could you share the link that saying spec that saying it is a constant time operation?
Thanks.

garfieldkwong:
FYI
http://en.cppreference.com/w/cpp/container/vector/swap

minggo:
Thanks.

PR #10832, opened 2015-03-12
fix bugs

### LINKED ISSUE #10355: using removeAllChildren on Scene causes Camera to be destroyed
Like the title says. The camera is added to the scene during creation, but if you call remove all children it destroys the camera. Why is the camera a child? Can't it be a protected member? The scene or the renderer should just visit its list of cameras directly, and not rely on the scene graph IMO.

super626:
fix #10355
Do not use MoveTo::reverse


dabingnn:
[ci rebuild]

PR #10843, opened 2015-03-12
Get rid of compiler warnings: field '_xxx' will be initialized after field '_yyy'...

konopka:
... [-Wreorder]

Initialization follows the order of declaration, not the order of the initialization list.


darkdukey:
Thanks! there are so much this warning it's actually breaking the mac build.

PR #10884, opened 2015-03-16
Fix strange zoom issue of ScrollView

matsujun:
If we add ScrollView as child of any other view which scale is not 1.0, offset is strange when we do pinch zoom.
It's bug of setZoomScale.
- '_touchPoint' is coordinates of ScrollView's node space, but handled as world space coordinates in setZoomScale
- 'offset' is offset on world space, but handled as offset on node space of ScrollView,

### DIFF
--- a/extensions/GUI/CCScrollView/CCScrollView.cpp
+++ b/extensions/GUI/CCScrollView/CCScrollView.cpp
@@ -259,7 +259,7 @@ void ScrollView::setZoomScale(float s)
         }
         else
         {
-            center = _touchPoint;
+            center = this->convertToWorldSpace(_touchPoint);
         }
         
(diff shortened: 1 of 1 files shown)


zilongshanren:
hi @matsujun  Sorry for the late reply. Could you provide a test case for your bug fix? Like we did in the `cpp-tests` project.
Thanks.

matsujun:
Where is test code of CCScrollView ?

zilongshanren:
@matsujun  There is no CCScrollView, only CCTableView has a test case. I think it's the right time to add an extra CCScrollView in Extensions.

minggo:
close it since it is unresponsible

PR #10909, opened 2015-03-16
Webview ios load data fix

babcca:
Fix loadHTMLString on iOS. Your test works bacause you load website (call `loadUrl(...)` ) at the first time and init `self.uiWebView` using `setupWebView`.
But when you call `loadHTMLString` first you didn't initialize webView and call method on `nil`.
Same issue is when calling `setScalesPageToFit(...)` befor first draw.

Example:

```
auto webView = WebView::create();
webView->loadHTMLString("<body>Hello World</body>", "");
addChild(webView);
```


WenhaiLin:

PR #10910, opened 2015-03-16
GlStateCache no correct reset Director matrix

perminovVS:
Use other OpenGL code in render frame cocos2d.
Need reset cocos 2d GL state, and continue render use cocos2d.
render Cocos -> render Other GL -> render Cocos.


minggo:
@dabingnn 
Please take a look.

PR #10913, opened 2015-03-16
Fix crash when play streamed mp4 file.

babcca:
Fix playing MP4 video from stream on iOS. More info at [http://stackoverflow.com/a/14836580/2175312](http://stackoverflow.com/a/14836580/2175312)
If you want audio on iOS, call

```
#import <AVFoundation/AVFoundation.h>
...
AVAudioSession *session = [AVAudioSession sharedInstance];
[session setCategory:AVAudioSessionCategoryPlayback error:nil];
```

in `proj.ios_mac/ios/AppController.mm:applicationDidBecomeActive(...)`
More at [Apple docs](https://developer.apple.com/library/ios/documentation/Audio/Conceptual/AudioSessionProgrammingGuide/ConfiguringanAudioSession/ConfiguringanAudioSession.html#//apple_ref/doc/uid/TP40007875-CH2-SW1)

### DIFF
--- a/cocos/ui/UIVideoPlayer-ios.mm
+++ b/cocos/ui/UIVideoPlayer-ios.mm
@@ -134,8 +134,9 @@ using namespace cocos2d::experimental::ui;
     }
     
     if (videoSource == 1) {
-        self.moviePlayer = [[[MPMoviePlayerController alloc] initWithContentURL:[NSURL URLWithString:@(videoUrl.c_str())]] autorelease];
+        self.moviePlayer = [[[MPMoviePlayerController alloc] init] autorelease];
         self.moviePlayer.movieSourceType = MPMovieSourceTypeStreaming;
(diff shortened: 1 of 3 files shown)


WenhaiLin:
@babcca I test this pr on iPhone 5s iOS 8.1,the strange "the playback of video will automatically pause "  question arises.

babcca:
When the video is paused? Original MP4 is really short because of licence (only first 1 minute is under CC). It's only for stream testing.

WenhaiLin:
It will automatically pause at regular intervals.

WenhaiLin:
So,i think you can set new video URL in VideoPlayerTest::menuOnlineVideoCallback, because the old URL is invalid.

WenhaiLin:
@minggo This pr should be close. @matsuokah I sync your fix on another one pr, update AUTHORS:https://github.com/cocos2d/cocos2d-x/commit/d24247a833f7c3c01f2fbf0d981598a79da22b0e

PR #10931, opened 2015-03-17
Added Bulgarian language.

lyubomirv:
Added a new value in the language enum and handled it in all needed places.

### DIFF
--- a/cocos2dx/platform/CCCommon.h
+++ b/cocos2dx/platform/CCCommon.h
@@ -69,7 +69,8 @@ typedef enum LanguageType
     kLanguageJapanese,
     kLanguageHungarian,
     kLanguagePortuguese,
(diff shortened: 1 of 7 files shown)


lyubomirv:
Jenkins fails the build of my changes, even though I have compiled them on linux, windows, ios and android myself. Jenkins doesn't even try to compile them. Can someone comment on this?

zilongshanren:
@lyubomirv  We don't use Jenkins for v2 branch. 
So you could just ignore it.

BTW, we only accept critical bug fix in v2 branch.

lyubomirv:
@andyque I will make a similar pull request for v3 in less than an hour.

I didn't know that you accept only but fixes in v2. Still, if you decide to include other things as well, please include my pull request.

Issue #10429, opened 2015-02-10
CCLabel cant be visited by custom camera


huangqiaobo:
first,I creat a perspective camera and setCameraFlag(CameraFlag::USER1);
add the camera to scene;

then I creat a label and add it to a billboard ;

auto label = Label::create();
label->setString("label");
auto billboard = BillBoard::create();
billboard->addChild(label);
billboard->setCameraMask((unsigned short)CameraFlag::USER1, true);

then add the billboard to the scene;
but i cant find the label in scene;

then i find the bug in CCLabel.cpp

void Label::createSpriteWithFontDefinition()
{
    _currentLabelType = LabelType::STRING_TEXTURE;

```
auto texture = new (std::nothrow) Texture2D;
texture->initWithString(_originalUTF8String.c_str(),_fontDefinition);

_textSprite = Sprite::createWithTexture(texture);
_textSprite->setCameraMask(this->getCameraMask());//we should add the code here,so that we can see the label;
_textSprite->setAnchorPoint(Vec2::ANCHOR_BOTTOM_LEFT);
this->setContentSize(_textSprite->getContentSize());
texture->release();
if (_blendFuncDirty)
{
    _textSprite->setBlendFunc(_blendFunc);
}

Node::addChild(_textSprite,0,Node::INVALID_TAG);

_textSprite->updateDisplayedColor(_displayedColor);
_textSprite->updateDisplayedOpacity(_displayedOpacity);
```

}

_textSprite->setCameraMask(this->getCameraMask());
we should add the code above after creat _textSprite ,so that we can see the label;   

（＝ ＝，sorry for my elementary English；）

WenhaiLin:
I think this issue cause by the camera mask no be cascaded when invoking Node::addChild.

super626:
#10936

PR #10935, opened 2015-03-18
V4fix

super626:
1. #10429 and clipping node
2. do not use Director::getInstance()->getRunningScene()->
3. shader not work well on some old windows platform


minggo:
Please modify CHANGELOG too.

PR #10936, opened 2015-03-18
Camera mask for label

super626:
To make sure label's member has correct camera mask

set camera mask when member created

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -932,6 +932,7 @@ void Label::createSpriteWithFontDefinition()
     texture->initWithString(_originalUTF8String.c_str(),_fontDefinition);
 
     _textSprite = Sprite::createWithTexture(texture);
+    //set camera mask using label's camera mask, because _textSprite may be null when setting camera mask to label
     _textSprite->setCameraMask(getCameraMask());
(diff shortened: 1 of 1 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -932,6 +932,8 @@ void Label::createSpriteWithFontDefinition()
     texture->initWithString(_originalUTF8String.c_str(),_fontDefinition);
 
     _textSprite = Sprite::createWithTexture(texture);
+    //set camera mask using label's camera mask, because _textSprite may be null when setting camera mask to label
```
minggo, line 935: `_textSprite may be null`: it is not correct, it is camera mask of _textSprite is null.

cocos/2d/CCLabel.cpp
```diff
@@ -1073,6 +1075,8 @@ void Label::drawTextSprite(Renderer *renderer, uint32_t parentFlags)
             {
                 _shadowNode->setBlendFunc(_blendFunc);
             }
+            //set camera mask using label's mask. Because _shadowNode may be null when setting the label's camera mask
```
minggo, line 1078: set -> Set,
and please add `.` at the end of a sentence.

PR #10980, opened 2015-03-19
3d particle performance test

super626:
1. particle3d performance test
2. fix crash on camera test because of using MoveTo::inverse


tests/cpp-tests/Classes/PerformanceTest/PerformanceParticle3DTest.cpp
```diff
@@ -0,0 +1,211 @@
...
+    kTagLabelAtlas = 3,
+    kTagMenuLayer = 1000,
+    kTagParticleSystem = 1001,
+    TEST_COUNT = 2,
+};
+
+enum {
+    kMaxParticles = 14000,
+    kNodesIncrease = 1,
+};
```
minggo: Please use enum class, even it is test codes.

PR #11007, opened 2015-03-19
Added drawRect function to CCDrawNode

Tiger-Coding:
New drawRect that takes a Rect as a parameter.  I use this a lot for debug drawing bounding boxes and thought it would be useful in v4.


minggo:
close it since v4 is freezed and can just use current drawRect function

PR #11034, opened 2015-03-20
update linux install deps script

zilongshanren:


zilongshanren:
@dabingnn  This PR should be merged before release alpha1. thanks.

Issue #11020, opened 2015-03-20
Linker failed to add libx11 on linux with cmake 3.2.1


samuelion:
Hello,
Just to inform, to warm, to let you aware of the following behavior I got after updating my system.
Long story short, I think the behavior I got is caused by the update of cmake 3.1.3 to 3.2.1 and this new deprecation : 

> The FindOpenGL module no longer explicitly searches for any dependency on X11 libraries with the FindX11 module. Such dependencies should not need to be explicit. Applications using X11 APIs themselves should find and link to X11 libraries explicitly.

http://www.cmake.org/cmake/help/v3.2/release/3.2.html#id12

```
cocos compile -p linux
```

> /usr/bin/ld: lib/libcocos2d.a(CCDevice-linux.cpp.o): undefined reference to symbol 'XOpenDisplay'
> /usr/lib/libX11.so.6: error adding symbols: DSO missing from command line
> collect2: error: ld returned 1 exit status
> CMakeFiles/MyGame.dir/build.make:155: recipe for target 'bin/MyGame' failed
> make[2]: **\* [bin/MyGame] Error 1
> CMakeFiles/Makefile2:64: recipe for target 'CMakeFiles/MyGame.dir/all' failed
> make[1]: **\* [CMakeFiles/MyGame.dir/all] Error 2
> Makefile:75: recipe for target 'all' failed
> make: **\* [all] Error 2

Manually adding the -lx11 to the linker command line resolves issues.
Don't know at all cmake and git to propose a PR.
Just wanted to trace the info as I didn't manage to find anything useful by myself expect by searching the code.

Hope this could help someone.

zilongshanren:
@samuelion 
Which branch do you use? v4-develop?

samuelion:
Your question make me think I didn't post the bug at the right place =).
I'm using the release version v3.4 directly taken from cocos2d website.
I can test with a git branch to confirm or not.

zilongshanren:
@samuele3hu 
Thanks. So I think this is a bug in v3.x.  We might need to specify the linker flag manually and don't rely on FindOpenGL.

samuelion:
Other way possible could be to explicitally use FindX11. Didn't manage to modify CMakeList to solve this (im a cmake newbie) but findx11 seems to give decent logs.

samuelion:
Manage to get something working, easier than I thought.

cocos2D/CMakelist : 
 cocos_find_package(OpenGL OPENGL REQUIRED)
[...]

PR #11041, opened 2015-03-20
#11020: Cmake 3.2.1 deprecation, no more auto link X11 with OpenGL

samuelion:
@andyque @samuele3hu 
Regarding the cmake 3.2.1 issue I got
To check, comment, ignore, whatever !
No change on Win, change working for me on Linux, should be changed on MACOSX, I try but would be nice to be tested or ignored !


worblehat:
Works for me on Linux, but shouldn't `cocos_find_package(X11 X11 REQUIRED)` be placed inside the `if(LINUX)`-branch in CMakeLists.txt?

samuelion:
Yes indeed, first I put the findX11 for all platform, second commit was supposed to remove it on Windows, but looks like I have done only half the work ! Will update tomorrow.
Dont know for Mac, but as an Unix I believed it might be needed also ?

worblehat:
I don't know if X11 is used on Mac Os X. If in doubt I would not include it...

minggo:

PR #11049, opened 2015-03-21
  Edit V3 updated precompiled shaders for WP8

stammen:
This pull request updates the WP8 version to the latest shaders and fixes a build problem due to the change in shaders. It also fixes the broken WP8 build by removing ShaderPositionColor_noMVP_GrayScale for WP8.


zilongshanren:
[ci rebuild]

PR #10911, opened 2015-03-16
SpriteFrameCache support async load

perminovVS:
Async based on Sprite3D and TextureCache.
Load texture and plist file in AsyncTaskPool TASK_IO thread.
Add TextureCache image and generate Sprite Frames in main thread.

PR #11063, opened 2015-03-22
UserDefaultData based on ValueMap store 

perminovVS:
UserDefault very slow, each call need read / write and parse xml file.
Delegate makes it possible to use other variants.

UserDefaultData stores user variables on ValueMap.
Get and Set is fast.
Loading data once, save on demand(use standart metod UserDefault::flush)
For call UserDefaultData::makeCurrentUserDefault() use on standart
UserDefault(Delegate UserDefault to new UserDefaultData )

Supports different ways to store( make you store based on
DataStoreDelegate, and set it UserDefaultData::setDataStoreDelegate)

Default data store based on standart UserDefault xml file, see
DataStoreXml in CCUserDefaultData.cpp

### DIFF
--- a/cocos/2d/libcocos2d.vcxproj
+++ b/cocos/2d/libcocos2d.vcxproj
@@ -435,6 +435,7 @@ xcopy /Y /Q "$(ProjectDir)..\..\external\chipmunk\prebuilt\win32\release-lib\*.*
     <ClCompile Include="..\base\CCTouch.cpp" />
     <ClCompile Include="..\base\ccTypes.cpp" />
     <ClCompile Include="..\base\CCUserDefault.cpp" />
+    <ClCompile Include="..\base\CCUserDefaultData.cpp" />
     <ClCompile Include="..\base\ccUTF8.cpp" />
     <ClCompile Include="..\base\ccUtils.cpp" />
     <ClCompile Include="..\base\CCValue.cpp" />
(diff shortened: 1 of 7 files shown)


minggo:
@perminovVS 
The design seems good.

I agree that make UserDefault extendable, but i think it is better to let developers to add UserDefaultData if they need.

perminovVS:
@minggo 
Old version is very slow, and need other code for android(CCUserDefault-android.cpp) and ios(CCUserDefault-apple.mm).
New version(UserDefaultData) support delegate store(UserDefaultData::setDataStoreDelegate), and need implementation DataStoreAndroid and DataStoreIos if incorrect work DataStoreXml.

I propose to leave both options, testing, implement Android and Ios and in the future to make UserDefaultData default.

please see other request: #10983 #10911 #10910

minggo:
UserDefault is not a small memory-based database. It will record all settings permanently. Because some it will lost data sometimes on iOS, so we replace xml with NSUserDefault on iOS and SharedPreference on Android.

perminovVS:
well understand.
Recreate request only "UserDefault support delegate" ?

minggo:
Yep, may be.
But the developer should implement all functions, so what's the difference between `delegate` and `create your own database` :)

perminovVS:
new UserDefaultData based on ValueMap delayed up to support the implementation of android and ios. 
UserDefault only setDelegate see #11499

PR #11091, opened 2015-03-24
fix memory leak

super626:


cocos/3d/CCSprite3D.cpp
```diff
@@ -290,6 +291,7 @@ bool Sprite3D::initWithFile(const std::string &path)
             }
             
             Sprite3DCache::getInstance()->addSprite3DData(path, data);
+            delete meshdatas;
```
minggo: Why some place use `CC_SAFE_DELETE`, and some use `delete`.
Is it any problem?

super626:
Done

PR #11119, opened 2015-03-24
Accelerometer polling for iOS

BodbDearg:
Add the ability to poll the accelerometer on iOS. Implement accelerometer polling to allow the acceleration data to be more easily used in update loops (rather than event driven input) and to workaround issues with lag on accelerometer events described here: http://discuss.cocos2d-x.org/t/accelerometer-data-lagging/16105

Note: Polling can also be implemented for Android and Windows phone, but currently stubbed out for those platforms.

### DIFF
--- a/cocos/platform/CCDevice.h
+++ b/cocos/platform/CCDevice.h
@@ -32,6 +32,7 @@ THE SOFTWARE.
 
 NS_CC_BEGIN
 
+class Acceleration;
 struct FontDefinition;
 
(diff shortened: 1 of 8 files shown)


BodbDearg:
Had to use this polling method to work around issues in a project I'm working on. Was occasionally experiencing severe lag and delays with accelerometer events when using the event based method...

minggo:

PR #11128, opened 2015-03-24
V4 renderer buffer refactor

mannewalis:
This replaces the native opengl code in Renderer with VertexData/VertexBuffer/IndexBuffer so that it will use the abstracted version in the Metal dev branch.

This also removes TriangleCommand which is not used or referenced anywhere.


mannewalis:
This builds fine, but it has not been thoroughly tested, so DO NOT MERGE!!!!

PR #11155, opened 2015-03-25
fix scale9sprite fro some cropped & rotated spriteframes

newnon:


newnon:
not sure about it is ready

PR #11164, opened 2015-03-26
Texturecube

super626:
1. TextureCube
2. SkyBox


minggo:
Please add doxygen comments for header files.

minggo:
And please fix compiling error on WP8.

PR #11173, opened 2015-03-26
Fixed every node of UI animation(json) is starting at frame 0.

Giseop:
I'm using both CocosStudio 1.6 and 2.1.5.
( Because CocosStudio 2.1.5 doesn't support opacity and visible
animation, we're using CocosStudio 1.6 for some UI animation. )

With 1.6 json file, we have some action nodes with non-zero startframe.
But non-zero startframe node run from zero-frame.

missing_start_frame


patriciog:
Hi @Giseop:

I think our pull request is related with yours.

https://github.com/cocos2d/cocos2d-x/pull/15366

We're using Cocos Studio 1.6 too.

Thanks and best!

PR #11195, opened 2015-03-27
refresh the search paths

kpkhxlgy0:
When I set the workdir in the command line on mac platform, it not affect the search paths, because the search paths set first, _defaultResRootPath is null string, so refresh the search paths can solve it.

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -846,6 +846,7 @@ void FileUtils::setWritablePath(const std::string& writablePath)
 void FileUtils::setDefaultResourceRootPath(const std::string& path)
 {
     _defaultResRootPath = path;
+    setSearchPaths(_searchPathArray);
 }
 
 void FileUtils::setSearchPaths(const std::vector<std::string>& searchPaths)


cocos/platform/CCFileUtils.cpp
```diff
@@ -846,6 +846,7 @@ void FileUtils::setWritablePath(const std::string& writablePath)
 void FileUtils::setDefaultResourceRootPath(const std::string& path)
 {
     _defaultResRootPath = path;
+    setSearchPaths(_searchPathArray);
```
dumganhar, line 849: Yep, good catch.

But i think this line
https://github.com/kpkhxlgy0/cocos2d-x/blob/e87207462912eef8235dfcab568eb4405a599002/cocos/platform/CCFileUtils.cpp#L857

will make the `searchPaths` parameter cleared.

We need to make a copy of `searchPaths` parameter before `clear`.

dumganhar:
I submitted a new Pull Request (https://github.com/cocos2d/cocos2d-x/pull/17435) to fix this issue, it also resolves some other issues I found in FileUtils.

Therefore, I closed this one.

PR #11196, opened 2015-03-27
3d fix doxygen add to group

tangziwen:


minggo:
Please fix conflict.

cocos/3d/CCBundle3DData.h
```diff
@@ -71,29 +89,16 @@ struct NodeData
...
-    {
-        resetData();
-    }
-    virtual void resetData()
-    {
-        bones.clear();
-        invBindPose.clear();
-    }
-};
-
```
minggo, line 96: Why modify codes?

super626, line 96: In fact, he did not modify codes here.

There may be some problems on git.

PR #11216, opened 2015-03-30
CCClippingRectangleNode add to a scale parent bug

lostplesed:
don't need to scale the point since it is a world coordinate


zilongshanren:
@lostplesed  How to reproduce this issue? Could you add a bug fix test case? 
Thanks.

minggo:
close it since don't know what's the issue

PR #11224, opened 2015-03-30
Remove use of autorelease NSString and NSURL

hawkwood:
Since there is no autorelease pool, these were leaking.


minggo:
Duplicated with #11223.

hawkwood:
This is not a duplicate, this is on a different file, AudioPlayer.mm instead of AudioCache.mm

minggo:
Sorry for my careless.

PR #11248, opened 2015-04-01
add new labelTTF enableShadow mehtod support to set shadow color

jianglong0156:
add new labelTTF enableShadow mehtod support to set shadow color 
the issues reference 
http://discuss.cocos2d-x.org/t/cc-labelttf-enableshadow-doesnt-work-in-native-builds/20747


minggo:
@WenhaiLin 
Please take a look.

minggo:
LabelTTF is deprecated, please use Label instead.

PR #11260, opened 2015-04-02
Making the context current before rendering

elnormous:
Context can be changed when using GUI components on OSX. This pull request makes the glfw make it's context current.
The following code is used, because of a bug in GLFW that does not change the current context if the internal pointer to window is the same (although the actual context has changed), so I set it first to nullptr and then to the main window pointer:

```
glfwMakeContextCurrent(nullptr);
glfwMakeContextCurrent(_mainWindow);
```


darkdukey:
[ci rebuild]

darkdukey:
[ci rebuild]

PR #11261, opened 2015-04-02
Added Application::applicationScreenSizeChanged call on winrt

musikov:


minggo:

PR #11274, opened 2015-04-03
Fix Desktop app crash upon exit when NotificationNode exists

vlad-tkachenko:
**Problem:**
When CCDirector contains NotificationNode and user tries to close the desktop app - it crashes.

**Solution:**
If notification node exists - destroy it inside _cc::Director->reset_ function.

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -957,6 +957,15 @@ void Director::reset()
     _runningScene = nullptr;
     _nextScene = nullptr;
     
+    if(_notificationNode)
+    {
+        _notificationNode->onExit();
+        _notificationNode->cleanup();
+        _notificationNode->release();
+    }
+    
+    _notificationNode = nullptr;
+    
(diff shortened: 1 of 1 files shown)


minggo:
How can i reproduce it?
I run cpp-empty-test on Mac, can not find it.

vlad-tkachenko:
In my game I'm adding a layer as a notification node. I'm also calling "onEnter" function on that layer after setting it to the director. I need that to make sure menu items will work.

When I do native Mac build and click on close (red) button - app crashes without this fix.

So, inside HelloWorldScene::init function just add the following:

``` cpp
auto notification = Layer::create();
Director::getInstance()->setNotificationNode(notification);
notification->onEnter();
```

Run Mac app and try to close it.

dotsquid:
I can't figure out why this issue is still not fixed. Please apply Vlad's pull request.

minggo:
Sorry for late reply. I tested and it fixed the issue.
Thanks @vlad-tkachenko

PR #11275, opened 2015-04-03
Fix NotificationNode event listeners priority

vlad-tkachenko:
**Problem:**
When notification node is used event listeners are not getting sorted for it and its child nodes.

**Solution:**
If notification node is available consider it as a root and calculate scene graph priority for it first.
If notification node is not available consider scene as a root.

**cocos2d-html5 related pull request:**
https://github.com/cocos2d/cocos2d-html5/pull/2809

### DIFF
--- a/cocos/base/CCEventDispatcher.cpp
+++ b/cocos/base/CCEventDispatcher.cpp
@@ -1185,15 +1185,34 @@ void EventDispatcher::sortEventListenersOfSceneGraphPriority(const EventListener
     
     if (sceneGraphListeners == nullptr)
         return;
-
     // Reset priority index
     _nodePriorityIndex = 0;
     _nodePriorityMap.clear();
-
-    Node *notificationNode = Director::getInstance()->getNotificationNode();
+    
+    Node* notificationNode = Director::getInstance()->getNotificationNode();
     if(notificationNode) {
         visitTarget(notificationNode, true);
-        visitTarget(rootNode, false);
+        
+        // save result in temp variable for later use
+        std::unordered_map<Node*, int> notificationNodePriorityMap; // = _nodePriorityMap
+        for(auto kv : _nodePriorityMap) {
+            notificationNodePriorityMap[kv.first] = kv.second;
+        }
+        int notificationNodePriorityIndex = _nodePriorityIndex;
+        
+        // reset
+        _nodePriorityIndex = 0;
+        _nodePriorityMap.clear();
+        
+        // visit rootNode
+        visitTarget(rootNode, true);
+        
+        // update priority map
+        for(auto kv : notificationNodePriorityMap) {
+            _nodePriorityMap[kv.first] = notificationNodePriorityMap[kv.first] + _nodePriorityIndex;
+        }
+        // update index
+        _nodePriorityIndex += notificationNodePriorityIndex;
     } else {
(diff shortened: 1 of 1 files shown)


cocos/base/CCEventDispatcher.cpp
```diff
@@ -1190,7 +1190,13 @@ void EventDispatcher::sortEventListenersOfSceneGraphPriority(const EventListener
     _nodePriorityIndex = 0;
     _nodePriorityMap.clear();
 
-    visitTarget(rootNode, true);
+    Node *notificationNode = Director::getInstance()->getNotificationNode();
+    if(notificationNode) {
+        visitTarget(notificationNode, true);
+        visitTarget(rootNode, false);
```
minggo: Why invoke notification node first?

minggo: In Director::drawScene, notification node is triggered after drawing a scene.

vlad-tkachenko: This fix is all about event handling. Logically notification node is always "on top" of scene, so it should receive events (e.g. touches) first. 

If you use CCMenu on both notification node and scene without this fix there is no guarantee that notification node's one will be triggered (if both are visually on the same location).

Shulepov:
I'm also using NotificationNode for some touchable alerts that should survive scene replacement. As many UI classes use sceneGraphPriority finally i also came to this solution.

footprint:
//footprint: fix notificationNode toch event
    Node\* notificationNode = Director::getInstance()->getNotificationNode();
    if(notificationNode) {
        // visit rootNode
        visitTarget(rootNode, false);
        visitTarget(notificationNode, true);
    } else {
        visitTarget(rootNode, true);
    }

cocos/base/CCEventDispatcher.cpp
```diff
@@ -1185,12 +1185,37 @@ void EventDispatcher::sortEventListenersOfSceneGraphPriority(const EventListener
...
+        // save result in temp variable for later use
+        std::unordered_map<Node*, int> notificationNodePriorityMap; // = _nodePriorityMap
+        for(auto kv : _nodePriorityMap) {
+            notificationNodePriorityMap[kv.first] = kv.second;
+        }
+        int notificationNodePriorityIndex = _nodePriorityIndex;
+        
+        // reset
+        _nodePriorityIndex = 0;
+        _nodePriorityMap.clear();
```
dumganhar, line 1205: Why not using 

``` c++
auto notificationNodePriorityMap = std::move(_nodePriorityMap);
```

Then _nodePriorityMap will be automatically cleared and avoid data copying.

dumganhar:
Observing a touch event by `addEventListenerWithSceneGraphPriority` for NotificationNode is strange since NotificationNode isn't in the scene graph.
How will you guys use NotificationNode? Did you add `Menu`, `MenuItem` to NotificationNode?
Or just make a customize clickable control by addEventListenerWithSceneGraphPriority?

vlad-tkachenko:
@dumganhar We were using NotificationNode for both MenuItems and custom event listeners. Mainly it was used for pop-ups that are not scene related and we want user to be informed even upon scene transitions. We also want user to dismiss the popups with some button and have a screen overlay (semi transparent mask), so that user have no way to interact with scene behind it.

PR #11276, opened 2015-04-03
Global variable fixes for HttpClient, note that this fix is only for Android, other platforms also need to be modified.

dumganhar:
WHY NEEDS THIS FIX?

Because in normal case, a game exits with its process shut down. But there are some cases like in `Cocos Runtime` which should not kill process while exiting game, and when entering game next time, global variables will stay with old values. There will be tons of  bugs in current strategy.

So **BE CARE OF** using global variables in Engine. Thanks.


dumganhar:
@minggo, how about this PR?

minggo:
Sorry, i missed it.
I will take a look now.

minggo:
[ci rebuild]

minggo:
It seems good.
Could you please fix compiling error?

dumganhar:
@minggo , I only refactored codes for Android since `Cocos Runtime` only supports Android now.
And I don't have enough time to modify codes for other platforms.
Could you please ask someone in your team to finish this patch? Thanks.

minggo:
Fixed in #11966.

dumganhar:
Thanks for the merge :)

PR #11281, opened 2015-04-03
Update UIRichText.cpp

LingJiJian:
support underline cache label/sprite


minggo:
Could you please use one PR for .h and .cpp files?

zilongshanren:
Thanks @LingJiJian 

I agree with @minggo , you'd better  merge your two PRs into one.
And it's better to provide another test case for the fix.

thanks :+1:

PR #11282, opened 2015-04-03
Fix iOS 5.1.1 crash caused by interruptionHandler

hawkwood:
checks OS version and sets handling based on what is available.


minggo:
@WenhaiLin 
Please help to check if it can work on newer iOS versions, such as iOS 8 and iOS 7.

minggo:
@hawkwood 
Do you know why it will cause crash on iOS 5.1.1?

WenhaiLin:
@minggo It's work fine on iOS8.1.

hawkwood:
It was crashing because the notification **AVAudioSessionInterruptionNotification** was added in iOS 6.

Issue #10086, opened 2015-01-20
Question about Sprite::setTexture


pandamicro:
Here is the implementation of `Sprite::setTexture`

```
void Sprite::setTexture(Texture2D *texture)
{
    // If batchnode, then texture id should be the same
    CCASSERT(! _batchNode || texture->getName() == _batchNode->getTexture()->getName(), "CCSprite: Batched sprites should use the same texture as the batchnode");
    // accept texture==nil as argument
    CCASSERT( !texture || dynamic_cast<Texture2D*>(texture), "setTexture expects a Texture2D. Invalid argument");

    if (texture == nullptr)
    {
        // Gets the texture by key firstly.
        texture = Director::getInstance()->getTextureCache()->getTextureForKey(CC_2x2_WHITE_IMAGE_KEY);

        // If texture wasn't in cache, create it from RAW data.
        if (texture == nullptr)
        {
            Image* image = new (std::nothrow) Image();
            bool isOK = image->initWithRawData(cc_2x2_white_image, sizeof(cc_2x2_white_image), 2, 2, 8);
            CC_UNUSED_PARAM(isOK);
            CCASSERT(isOK, "The 2x2 empty texture was created unsuccessfully.");

            texture = Director::getInstance()->getTextureCache()->addImage(image, CC_2x2_WHITE_IMAGE_KEY);
            CC_SAFE_RELEASE(image);
        }
    }

    if (!_batchNode && _texture != texture)
    {
        CC_SAFE_RETAIN(texture);
        CC_SAFE_RELEASE(_texture);
        _texture = texture;
        updateBlendFunc();
    }
}
```

Two questions:
1. The second ASSERT indicate that we accept `nullptr` as texture parameter, but in the first ASSERT we already used the texture which will cause crash, so `nullptr` cannot be passed, please fix the problem
2. It's a question, why instead of making `_texture` null, we make it a `2x2` white texture ? I don't think it should be the expected behavior.

fusijie:
https://github.com/cocos2d/cocos2d-x/pull/12748 is to fix first issue.

@dabingnn please help to check.

fusijie:
[This branch ](https://github.com/cocos2d/cocos2d-x/compare/v3...fusijie:remove_cc_2x2_white_image)is to fix second issue. But @dabingnn & I think it not necessary to remove 2x2 white image, because it used to give a warning to developers that there are some null sprite exist.

fusijie:
@pandamicro Please close this issue.

Issue #8328, opened 2014-09-30
cc_2x2_white_image  is missing , giving warning at runtime


minggo:
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/3886). It was created at 2014/01/30 12:55:16 +0000
# Description

In CCSprite.cpp
there is #define CC_2x2_WHITE_IMAGE_KEY  "/cc_2x2_white_image" definnition 
but there is no file by this name 
in runtime there is warning that looks like this :

```
cocos2d: fullPathForFilename: No file found at /cc_2x2_white_image. Possible missing file.
```
- zhangxm added comment:
  The data is in the byte array named `cc_2x2_white_image` in CCSprite.cpp. You don't have to care about it. It is used when there is not a texture for a `Sprite`.

pandamicro:
It's not a bug

PR #11283, opened 2015-04-04
TextureCache::getTextureForKey add param checkFullPath

perminovVS:
Is not always necessary to check FullPath, if key not found. example
wrong message: "cocos2d: fullPathForFilename: No file found at
/cc_2x2_white_image. Possible missing file." #8328

Sprite::setTexture
CCASSERT add check texture not nullptr
#10086 First questions

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -318,14 +318,14 @@ void Sprite::setTexture(const std::string &filename)
 void Sprite::setTexture(Texture2D *texture)
 {
     // If batchnode, then texture id should be the same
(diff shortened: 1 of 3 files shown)


darkdukey:
[ci rebuild]

WarlockD:
I agree.  Or if its just that white image key, putting this gets rid of that annoying message:

```
 FileUtils::getInstance()->setPopupNotify(false); 
 texture = _director->getTextureCache()->getTextureForKey(CC_2x2_WHITE_IMAGE_KEY);
 FileUtils::getInstance()->setPopupNotify(true);
```

minggo:
`cc_2x2_white_image` is a special case, it is used by engine internally for fps shown.

PR #11290, opened 2015-04-05
Add macros CC_MAKE_SUITABLE_F_OPEN

perminovVS:
win32 fopen only ansi encoding.
if filePath contains utf8 characters fopen fail.
AudioCache FileFormat::OGG bag, can't open file, if filePath contains
utf8.

Need make all filename for fopen correct encoding.
Only Win32 CC_MAKE_SUITABLE_F_OPEN return
StringUtf8ToAnsi(filename).c_str(),
other platform return filename.c_str()

all fopen use CC_MAKE_SUITABLE_F_OPEN


WenhaiLin:
[ci rebuild]

perminovVS:
Sorry, forgot to add in the commit StringUtf8ToAnsi function, corrected.

minggo:
Why only win32 needs to convert utf8 to ansi?

perminovVS:
@minggo for everyone win platform need convert to local ansi, but I do not have the possibility to check on the others.

minggo:
There are some helper functions in ccUTF8.h/.cpp.
I think we should add the function there and implement it in all platforms.

perminovVS:
@minggo Add macro and support functions there?

minggo:
Can't it be a function instead of macro?

perminovVS:
@minggo Better to add std::string makeSuitableFOpen(const std::string& fullPath) in FileUtils.
Platform that needs implements its functionality(No add in ccUTF8.h, this specific platform)

perminovVS:
see #11496

PR #11293, opened 2015-04-06
GLViewImpl initWithRect bug size

perminovVS:
glfwCreateWindow not guarantees create windows need size, watch real
create size and change setFrameSize( see declaration glfwCreateWindow)

Example Windows 7, if request height size > Desktop height make window
size: Desktop height - height Caption.


darkdukey:
[ci rebuild]

dabingnn:
Thanks, It is a limitation on Windows.
I found it in [GLFW documentation](http://www.glfw.org/docs/latest/group__window.html)
`Window systems put limits on window sizes. Very large or very small window dimensions may be overridden by the window system on creation. Check the actual size after creation.`

@pandamicro Please merge it.

PR #11295, opened 2015-04-06
RenderQueue command buffer optimisation

Almax27:
Made _commands an array of vectors that is no longer reallocated every frame
Provided method to reallocate queues manually and reserve memory in order to minimise std::vector dynamic reallocation

### DIFF
--- a/cocos/renderer/CCRenderer.h
+++ b/cocos/renderer/CCRenderer.h
@@ -73,10 +73,7 @@ public:
 
 public:
     /**Constructor.*/
-    RenderQueue()
-    {
-        clear();
-    }
+    RenderQueue();
     /**Push a renderCommand into current renderqueue.*/
     void push_back(RenderCommand* command);
     /**Return the number of render commands.*/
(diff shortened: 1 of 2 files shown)


minggo:
[ci rebuild]

minggo:
@Almax27 
What's the benefit of it?
Do you have any testing about the benefit?

Almax27:
I came across it while profiling:

In this example I was rendering 10,000 60x60px sprites and taking a profile sample of ~10 seconds. The memory allocation overhead from destroying and rebuilding the RenderQueue vectors was taking up 1.5+% of the total CPU time. After the proposed change, the cost to clear the RenderQueue becomes negligible.

Before
before
After
after

darkdukey:
[ci rebuild]

minggo:
@Almax27 
It seems great.

cocos/renderer/CCRenderer.h
```diff
@@ -87,6 +84,8 @@ class RenderQueue {
     RenderCommand* operator[](ssize_t index) const;
     /**Clear all rendered commands.*/
     void clear();
+    /**Realloc command queues and reserve with given size. Note: this clears any existing commands.*/
+    void realloc(size_t reserveSize);
```
minggo, line 88: When will invoke this function?

Almax27:
That's for developers who want more control, and for the sake of code completion :P. Say you were on a tight memory budget and transitioned from a scene with 10,000 sprites to one with 100, realloc(size) allows you to clear the memory allocated by the queue and prepare for what will be rendered next.

minggo:
Thanks.

minggo:
I have a question. 
When QUEUE_COUNT is not enough, developers should invoke realloc manually?

minggo:
Sorry, ignore my question.
I misunderstood.

darkdukey:
[ci rebuild]

ricardoquesada:
looks good to me.

Issue #10293, opened 2015-01-30
More mobile features requested: Album access, Camera access etc


pandamicro:
There are some common used features in mobile games, like Album for save screen captions or upload user images. Camera access for taking photos or reality augmented games. If Cocos2d-x can provide a common API across iOS / Android / Windows Phone, it should be very useful.

Here is a what I can image:
- Album
- Camera
- Microphone
- Geolocalisation

You can add your own ideas to this list

zilongshanren:
+1 for Vibration

gyf19:
+1

yagjak:
Yes, it will be very useful for us, thanks cocos team.

0xJimmyBeh:
Any milestone for this feature to be implemented? Need to bind them to use in Cocos2d-JS for cross platform project.

Thanks!

ricardoquesada:
sounds good to me.

What other useful features are missing? Let's identify them and then prioritize them.

0xJimmyBeh:
From the post on Cocos2d-x forum : 
http://discuss.cocos2d-x.org/t/more-mobile-features-requested-album-access-camera-access-etc/19694/3

Gyroscope
Text to speech
Local Notifications
Gesture Support
Social
Payments
Advertising

ricardoquesada:
What do you mean by Text-to-Speech ?

I think these features should be divided in two:

a)
those features that are offered by the device itself. eg: camera, gyroscope, etc.

b)
those features that are services. eg: social, payments, advertising

this issue should only cover a)... a different issue should be created for b).

regarding a)... this is the list so far, any other feature ?
- Gyroscope
- Camera
- Gestures
- Microphone
- Photo Albums
- Geolocalization

0xJimmyBeh:
Those features are copy from the comments of forum post as summary.

From what i understand, Text-to-Speech is more to feature provided by the app itself, so should not related to this issue.

I'm not sure what mobile features does Cocos2d-X provided, so i just listed out what may fall on category a)
- Gyroscope
- Camera
- Gestures
- Microphone
- Photo Albums
- Geolocation (GPS)
- Magnetic Sensor
- Light Sensor
- Proximity Sensor
- Vibration
- Accelerometer (Implemented)
- NFC

Does NFC consider a mobile feature?

ricardoquesada:
thanks.... probably NFC could be considered a mobile feature.

0xJimmyBeh:
[...]

PR #11296, opened 2015-04-06
ImagePicker utility with support for Android/iOS(iPhone+iPad)/Win/OSX

0xJimmyBeh:


minggo:
@Zinitter 
Thanks.
Why add image picker support?

lalaloha:
Finally! I needed this feature before to make apps that can use custom pictures from their phones. Maybe this can be added to extension if it's not making sense to add to core engine :)

0xJimmyBeh:
@minggo 
Image picker is part of the basic mobile features as issue here : 
https://github.com/cocos2d/cocos2d-x/issues/10293

We can use image picker to upload image and there are many usages with the upload image features.

For example :
Customise our profile picture
Share screenshot (game or report bug)
Exchange photo

Nowadays, we can't provide only playing feature in a game. 
We have to provide more features to attract and keep users in the game like a community which can share and discuss.

From a post on Cocos2d-x forum, @ricardoquesada said
"Yes, we want cocos2d-x to be much more than just a renderer."

http://discuss.cocos2d-x.org/t/development-impression/20862/14

I believe the basic mobile features support are a must to improve gaming experience but not limited to game usage since Cocos2d-x can be used to develop utility app as what i doing now.

Thanks.

0xJimmyBeh:
Please refer to the pull request of author
https://github.com/cocos2d/cocos2d-x/pull/11344

PR #11314, opened 2015-04-07
Android: Video player stability

stari4ek:
fixes:
- if video has same aspect ratio as android device's screen - it won't be displayed properly
- sometime video isn't visible

Second issue is race condition, if `onMeasure` is called before video is loaded or rect is sent view's size will be 0x0

Related discussion: http://discuss.cocos2d-x.org/t/videoplayer-bug-with-android/17072/6


stari4ek:
sorry. 1 more commit.

checked with devices:
- Samsung GT-I9300I (S3 Duos)
- Samsung GT-I9500 (S4)
- LG-D855 (G3)
- Samsung GT-P5100
- HTC One M8

WenhaiLin:
@stari4ek Many thanks for your work.I will test it tomorrow.

stari4ek:
one more commit, thanx to @Daryl from http://discuss.cocos2d-x.org/t/videoplayer-bug-with-android/17072/8

WenhaiLin:
@stari4ek Let VideoPlayer play video in full screen isn't good idea. Please try this solution: https://github.com/cocos2d/cocos2d-x/pull/11651

PR #11344, opened 2015-04-08
Image Picker with tests, tested in iOS/Android/OSX/Win already

qiankanglai:
The origin PR is here: https://github.com/cocos2d/cocos2d-x/pull/11296 by @Zinitter .

I'm the guy who wrote it and took some time tidying up my codes. Additionally, I put my codes in extensions and add the tests needed. For other platforms like linux/WP8, I only add an empty function which can compile but not work(Maybe I'll implement WP8 later, after google it...)


darkdukey:
Wow, this is great

qiankanglai:
I'm quite sorry for so many commits. After digging into build farm console logs, I've added these files into WP8/W8/Linux

extensions/ImagePicker/ImagePickerIOS.mm
```diff
@@ -0,0 +1,61 @@
+#import <QuartzCore/QuartzCore.h>
```
minggo: ImagePicker-ios.mm

extensions/ImagePicker/ImagePickerIOS.h
```diff
@@ -0,0 +1,14 @@
+#ifndef __ImagePicker__ImagePickerIos_
```
minggo: ImagePicker-ios.h

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
```diff
@@ -23,6 +23,8 @@ of this software and associated documentation files (the "Software"), to deal
  ****************************************************************************/
 package org.cocos2dx.lib;
 
+import java.io.ByteArrayOutputStream;
+
```
minggo: I think it is better to add a separate java file for it, or Cocos2dxActivity.java will be complex.

qiankanglai: @minggo I see. Maybe I can use ActivityResultListener

minggo: I am not sure. 
But it is better to put it into a separate file.

qiankanglai: Got it. I'll do it later. iOS files have been renamed.

minggo: Thanks.

WenhaiLin:
[ci rebuild]

minggo:
We are developing cocos package manager, which treat extension features as a plugin, so i think it is better to use cocos package manager to add image picker. Or cocos2d-x will more complex because all extensions will be integrated into cocos2d-x.

@ricardoquesada 
What's your opinion?

qiankanglai:
@minggo I do agree to add as a plugin. I google around and only find some discussions. Is there any documentation or developer's guide?
Also, I'm not sure whether js/lua bindings can be done within package manager?

minggo:
@dualface 
What's the status of package manager?
Is there any documentation of it?

0xJimmyBeh:
@minggo 
Does Cocos2d-JS support package manager?

I would like to use this Image Picker with Cocos2d-JS ASAP as my project can't complete without this feature. I'm waiting this feature to be implemented since January.

I believe there are a lot of Cocos2d developer who want to use this basic feature too and happy to see more basic features bring to Cocos2d.

Is there any temporarily solution to make this Image Picker work with Cocos2d-JS v3.5?

Thanks.

lalaloha:
I also need to have this feature for my upcoming project.. Can we integrate this first, and after cocos package manager ready, move this to package manager. Also, we don't know whether package manager will be available for developers outside china?

minggo:
Hi guys.
I am not sure if i should merge it or not.
I created [this topic](http://discuss.cocos2d-x.org/t/how-to-handle-pull-requests-that-adding-features-to-extend-cocos2d-x/20961) to ask the community in forum.

qiankanglai:
I've added support for win8.1-universal. The code for phone version is a bit ugly since I cannot find a way to separate codes from `App.xaml`.

However, when I try to compile wp8, the visual studio solution has some problems related to freetype2. Also, its `cpp-tests` shows strange images like this. 

qq 20150410211058

This indeed shows that maintaining is a hard work :smile:

PR #11365, opened 2015-04-09
removes warning

ricardoquesada:
warnings are treated as errors on Mac.
Why this was even merged? Do we have a Mac builder? The Mac builder
should fail.


minggo:
@andyque 
I think we use cmake to build project on Mac, right?
And as i know, it is the requirement by you Riq.

PR #11366, opened 2015-04-10
Fix PURibbonTrail bug

super626:


minggo:
What's the bug?

minggo:
Please modify CHANGELOG.

Issue #11371, opened 2015-04-10
Add disableOutline and disableShadow API for ui::Text


zilongshanren:
I only want to add one API for it. 
For more information, please refer to http://wiki.qt.io/API-Design-Principles

PR #11375, opened 2015-04-10
add disableEffect API for ui::Text

zilongshanren:
In this PR, it also deletes all the the empty space in source code.
In some  languages, like Python, the extra white space would consider to be a warning according to PEP8.


zilongshanren:
This PR fixed issue https://github.com/cocos2d/cocos2d-x/issues/11371

Issue #11370, opened 2015-04-10
Fix ui::Text content size bug when applying outline effect


zilongshanren:
If apply other label effect, the content size should also be updated.

PR #11378, opened 2015-04-10
fix ui::Text outline don't change content size

zilongshanren:
This PR also does some comments format. I know it's not a good practice to do two things in one PR.
But I think formatting stuff is secure. Thanks.


zilongshanren:
This PR fixed issue https://github.com/cocos2d/cocos2d-x/issues/11370

PR #11384, opened 2015-04-10
Fix PU bug

super626:
make particle local does not work
particle direction calculateion bug


extensions/Particle3D/PU/CCPUScriptTranslator.h
```diff
@@ -159,7 +159,7 @@ enum eDefineStaticToken
     TOKEN_EMITTER_PARTICLE_HEIGHT,
     TOKEN_EMITTER_PARTICLE_DEPTH,
     TOKEN_EMITTER_AUTO_DIRECTION,
-    TOKEN_EMITTER_FORCE_EMISISON,
+    TOKEN_EMITTER_FORCE_EMISSION,
```
minggo, line 162: I think it is better to add 
TOKEN_EMITTER_FORCE_EMISSION = TOKEN_EMITTER_FORCE_EMISISON,

Or it will break compatibility.

super626, line 162: Thanks.

In fact, it is used internally.

super626:
[ci rebuild]

PR #11441, opened 2015-04-14
Updates socket.io network extension to correctly detect and manage connection to socket.io server versions 0.9.x - 1.x

himynameschris:
This commit is a combination of the following commits:
-update handshake and handshakeResponse methods to work with both 0.9.x and 1.x socket.io versions, successfully parsing session id, heartbeat and timeout
-adds SocketIOPacket class and updates socket send, disconnect and emit methods
-update the receive / onMessage method, seperate for v0.9 and 1.0
-implements simple string packets, disconnect call, event name detection
-correctly parse endpoint and event names from the payload for socket.io 1.x
-implement json writing for socket.io 1.x message and event sending
-fixes endpoint handling between .9 and 1, cleans up some codes
-safeguard against non-existing clients
    -this occurs when you connect to an endpoint without connecting to the default namespace "" or "/" first
-onMessage and onConnect are no longer required to be overridden and should be deprecated in the future
    -onError and onClose will still be used to communicate network level issues, but seperate socket.io event handlers should be created
-socket.io connect, errors and disconnects are now handled as socket.io events, while network layer error and close are still implemented by the delegate
-update comments and license header
-fix event name and payload parsing for 0.9 method
-update tests with separate tests for socket.io 0.9.x and 1.x using the same test layer and a version initializer
-add missing register call for json message handler
-add constructor to take version as argument, update host used for tests to dev.channon.us

Signed-off-by: Chris Hannon himynameschris@gmail.com

### DIFF
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -1,6 +1,6 @@
 /****************************************************************************
- Copyright (c) 2013      Chris Hannon
- Copyright (c) 2013-2014 Chukong Technologies Inc.
+ Copyright (c) 2013-2015 Chris Hannon http://www.channon.us
+ Copyright (c) 2013-2015 Chukong Technologies Inc.
 
  http://www.cocos2d-x.org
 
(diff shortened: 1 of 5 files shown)


himynameschris:
@minggo @ricardoquesada Is socket.io support something that is being kept in v4?

Also, should this update be backported to v3.x and included in cocos2d-js as well? (the existing bindings should work without modification as the updates are backwards compatible)

minggo:
@hannon235 
Now we focus on v3.x, so i think this pull request should be sent to v3.x.
Thanks.

himynameschris:
Ok, makes sense. I will port this over to the v3 branch, if I should be using a different branch such as v3.6 let me know

cocos/network/SocketIO.cpp
```diff
@@ -35,12 +35,319 @@
 #include <algorithm>
 #include <sstream>
 
+#include "rapidjson/rapidjson.h"
+#include "rapidjson/document.h"
+#include "rapidjson/stringbuffer.h"
+#include "rapidjson/writer.h"
+
```
eltonjothi, line 42: the cocos2d 3rd party libraries has the name json instead of rapidjson...
i think it should be #include "json/xxx.h"

cocos/network/SocketIO.cpp
```diff
@@ -35,12 +35,319 @@
...
+if (_type == "message") {
+	outS = _args[0];
+}
+else {
+
+	rapidjson::StringBuffer s;
+	rapidjson::Writer<rapidjson::StringBuffer> writer(s);
+
+	writer.StartObject();
+	writer.Key("name");
```
eltonjothi, line 217: No member Named Key in the default 3rd party cocos2d json library

himynameschris, line 217: I closed this PR and opened another on the v3 branch, it includes the
changes you mention above

himynameschris:
See #11538

himynameschris:
Sorry meant #11583

eltonjothi:
awesome.. Thanks ... will check on it..

PR #11172, opened 2015-03-26
Fixed jni illegal start byte error.

matsuokah:
In JNI,  `methodInfo.env->NewStringUTF(text);` method happens `JNI DETECTED ERROR IN APPLICATION: input is not valid Modified UTF-8: illegal start byte` when we input Emoji such as "🍣".
Then, I Fixed that using the byte array instead of String.
And creating the text String in Java runtime.

And, I searched the depending to [createTextBitmap](https://github.com/cocos2d/cocos2d-x/blob/2b77006d1c5505865d3d1f87690ddb1ec8612fcb/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxBitmap.java#L92) method,  but it is not used anywhere. But I fixed too for complie.

Thank you for your working! :smile:

PR #11451, opened 2015-04-15
Fixed JNI illegal start byte error

WenhaiLin:


WenhaiLin:
sync https://github.com/cocos2d/cocos2d-x/pull/11172

PR #11453, opened 2015-04-15
added getLayer() method for Node

SungChang:
Added a method "getLayer()" in Sprite class.
A Sprite can get the layer that contains it. If the sprite is not contained in a Layer, then returns a nullptr.

*modified

### DIFF
--- a/cocos/2d/CCLayer.cpp
+++ b/cocos/2d/CCLayer.cpp
@@ -424,6 +424,11 @@ std::string Layer::getDescription() const
     return StringUtils::format("<Layer | Tag = %d>", _tag);
 }
 
+Layer* Layer::getLayer()
+{
+    return this;
+}
+
 __LayerRGBA::__LayerRGBA()
 {
     CCLOG("LayerRGBA deprecated.");
--- a/cocos/2d/CCLayer.cpp
+++ b/cocos/2d/CCLayer.cpp
@@ -1,1001 +1,5536 @@
-/****************************************************************************
-Copyright (c) 2008-2010 Ricardo Quesada
-Copyright (c) 2010-2012 cocos2d-x.org
-Copyright (c) 2011      Zynga Inc.
-Copyright (c) 2013-2014 Chukong Technologies Inc.
-
-http://www.cocos2d-x.org
-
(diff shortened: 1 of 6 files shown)


minggo:
It is strange that add this function in Node. Because Layer is a sub-class of Node.

SungChang:
Hmm getScene() method is also in Node class.
Should we change getScene Too?

Do you think getLayer fit better to Sprite?

PeqNP:
Have you considered using the `getParent()` method on the `Node` class? This should satisfy 99.9% of the use cases as most sprites are added to a `Layer`; which is generally the parent.

I'm not sold on the idea of `getScene()` on the `Node` class either. That should be removed for the same reasons that `getLayer()` shouldn't be on the `Node`; as a `Scene` is a `Node`.

That being said, you could potentially walk up the hierarchy tree to determine that too. For instance, if there was a method that asked for a type of object that a `Node` belongs to, and returns the first instance of that type, you could resolve both of these issues. Something like:

```
local layer = cc.Layer:create()
local node = cc.Node:create()
layer:addChild(node)
local layerRef = node:getFirstParentWithType(cc.Layer)
```

You incur the cost of looking up the tree (which _may_ not be too costly), but I feel you wouldn't be calling a method like this often in the first place _ducks from people throwing fruits and vegetables at him_. If you really need access to a `Node`'s layer often, then it might be better to create a model/controller that has a reference to the layer and all of its children. Such that you have a 'LevelController' object that contains a reference to the `Layer` and also contains references to all of the actors within that `Layer`.

I'm curious to know what the use cases of both of these functions would be, as I feel I can get the information I need without either of these functions.

SungChang:
I agree with you. But I thought this method would be convenient. There are a lot of methods that we do not have to have them. Many of them are written and being used for convenience. Some Sprites, such as HP bar and equipments , have other Sprites as parent, not Layer. For the reason, I thought it is inconvenient to look up and think of the layer design of game when I implement methods in some of the Sprites nodes, like HP bar and equipments I mentioned.

This is why I thought getLayer() and getScene() are useful. I dont think it's right to remove getScene() because some users are definitely using it.

SungChang:
I will change it to be in Sprite instead of Node.

PR #11458, opened 2015-04-16
fix side effect of ui::button clone

zilongshanren:
In the previous PR https://github.com/cocos2d/cocos2d-x/pull/11429, 
I have optimized the memory consumption. 

Now the title of button will be created on demand. But in the previous PR, the clone method couldn't clone a exactly copy of the original one. 

When the original button doesn't have title renderer, we shouldn't create the title renderer for the cloned copy.

So this PR fix this issue and a test case also be added.

### DIFF
--- a/cocos/ui/UIButton.cpp
+++ b/cocos/ui/UIButton.cpp
@@ -899,10 +899,13 @@ void Button::copySpecialProperties(Widget *widget)
         setCapInsetsNormalRenderer(button->_capInsetsNormal);
         setCapInsetsPressedRenderer(button->_capInsetsPressed);
         setCapInsetsDisabledRenderer(button->_capInsetsDisabled);
-        setTitleText(button->getTitleText());
-        setTitleFontName(button->getTitleFontName());
-        setTitleFontSize(button->getTitleFontSize());
-        setTitleColor(button->getTitleColor());
+        if(nullptr != button->getTitleRenderer())
+        {
+            setTitleText(button->getTitleText());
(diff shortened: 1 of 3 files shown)


edwardzhou:
There is another similar isssue, ui::Button does not clone its children.
eg. in lua:

local buttonSize = button:getContentSize()
local tittle = cc.Sprite:create('buttonTitle.png')
title:setAnchorPoint(0.5, 0.5)
title:setPosition(buttonSize.width / 2, buttonSize.height /2)
button:addChild(title)

local newButton = button:clone()

// the newButton was created without 'buttonTitle.png' child sprite.

zilongshanren:
@edwardzhou 
Button's children will be cloned in Widget's implementation.

```
Widget* Widget::clone()
{
    Widget* clonedWidget = createCloneInstance();
    clonedWidget->copyProperties(this);
    clonedWidget->copyClonedWidgetChildren(this);
    return clonedWidget;
}
```

The widget do clone all its children.  Could you verify it?

zilongshanren:
@edwardzhou 
Oh, I know. Only children with `widget` type will be cloned.

I think it should be a bug.

zilongshanren:
@edwardzhou 
And I think the clone shouldn't be recursive...

zilongshanren:
@edwardzhou 
Because non-widget class doesn't support deep copy. 
And one Node can't has two parents. 
So we could only implement deep copy and recursive copy for all the Widgets.
All other Node types are not supported.
Thanks.

PR #11465, opened 2015-04-16
Fix crash when play streamed MP4 file.

WenhaiLin:


WenhaiLin:
sync https://github.com/cocos2d/cocos2d-x/pull/10913

PR #11472, opened 2015-04-16
Adding Lerp for Vec3

milos1290:
Missing lerp for Vec3


milos1290:
I've also added Lerp method for float values in MathUtils. This is something that is really used.

PR #11475, opened 2015-04-17
fix error when ListView is empty

enhhh:
size_t is an unsigned int ,when ListView has no item,total height will be a max value.and cause jumpToTop and other scroll functions error


cocos/ui/UIListView.cpp
```diff
@@ -93,7 +93,7 @@ void ListView::updateInnerContainerSize()
         case Direction::VERTICAL:
         {
             size_t length = _items.size();
-            float totalHeight = (length - 1) * _itemsMargin;
+            float totalHeight = ((float)length - 1) * _itemsMargin;
```
zilongshanren: _itemsMargin is float, so the (length-1) \* _itemsMargin would also be float, right?
Why should we add a extra cast here?

enhhh: length is unsigned int , if length is 0, (length - 1) will be a max value, so if _itemsMargin is not 0,totalHeight will be a large value

.

zilongshanren: Thanks @enhhh 
size_t is evil here. we should change it with `int` or `ssize_t`.
Thoughts?

enhhh: change `size_t` with `int` or `ssize_t` is a good idea. :)

PR #11488, opened 2015-04-18
added getLayer() to Node

SungChang:
This is a helper function to get the Layer that containing the Node.

This has similar concept to the getScene method in Node.
https://github.com/cocos2d/cocos2d-x/pull/4589


zilongshanren:
hi, @scalated  The get layer API is very trivial and think developers could implement it very easily.
I think it's no need to provide it.
@ricardoquesada  Thoughts?

ricardoquesada:
we are going to remove `Layer` in the future. Layer will be replaced by `Node` and its events API will be replaced by using directly the Events API.

So, I don't think we need to keep adding functionality to `Layer`

PR #11493, opened 2015-04-18
Fix ownership bug in CCFontAtlasCache.

AknEp:
I sometimes met EXC_BAD_ACCESS in FontAtlasCache::purgeCachedData().
I think CCFontAtlasCache retain/release logic is reverse, causes BAD ACCESS problem above.

this pull-req fixes them.

Best regards.


AknEp:
I'm sorry, this is bad pull-req; makes your project crashes more frequently.

I'll make pull-req again & add test code covers this problem.

PR #11494, opened 2015-04-18
Simplified the way to bind callbacks from Cocostudio

TheCodez:
This change simplifies the way callbacks can be handled from Cocostudio. Instead of registering a new node reader you just call CSLoader::createNodeWithRoot with the desired target rootnode. Additionally, I've added three macros to simplify the process of binding your custom events.

this :

``` c++
if (callBackName == "onClick")
{
        return CC_CALLBACK_1(CustomRootNode::onClick, this);
}
```

becomes this :

``` c++
CCS_CLICK_CALLBACK_GLUE(this, "onClick", CustomRootNode::onClick)
```


chengstory:
Thanks @minggo  @TheCodez ,  I ask @pipu to take a look.

TheCodez:
This is pretty much the same as in this pr : https://github.com/cocos2d/cocos2d-x/pull/10403
But a bit simpler and with extra macros.

TheCodez:
@chengstory @pipu any news?

minggo:

PR #11288, opened 2015-04-05
Optimize Vec3

perminovVS:
small function Vec3 move to Vec3.inl
Added:
add(float xx, float yy, float zz);
setZero();

Change all code:
_vec3 = Vec3(x, y, z);   ->   _vec3.set(x, y, z);
Vec3 vec3 = Vec3(x, y, z);   ->  Vec3 vec3(x, y, z);
_vec3 += Vec3(x, y, z);   ->   _vec3.add(x, y, z);
_vec3 = Vec3::ZERO;   ->   _vec3.setZero();

performance issue #11170

PR #11497, opened 2015-04-19
Optimize Vec3 use constructor default.

perminovVS:
Make Vec3 vec3(Vec3::ZERO);   to   Vec3 vec3;

extansion #11288


getsuitablefopen.patch
```diff
@@ -0,0 +1,792 @@
+﻿From 19780399aae4e77419dbd4e89f135784bfba7c6c Mon Sep 17 00:00:00 2001
```
minggo: Why uploaded this patch file?

minggo:
Could you please fix conflict and fix compiling error.

minggo:

perminovVS:
@minggo yes of course

perminovVS:
fixed all in #11518

Issue #11511, opened 2015-04-20
Button->setBright(bool) makes memory go up.


fusijie:
`version:` 3.5Final
`issue:` Every time creates a button and setBright(false), GLProgramStateCache will cache the program state makes memory go up.
`reproduce:` add following code to HelloWorld project, then pay attention to memory.

``` cpp
    this->schedule([](float){
        ui::Button* btn = ui::Button::create("CloseNormal.png","","",ui::Widget::TextureResType::LOCAL);
        btn->setBright(false);
    }, 0.05, "lambda");
```

zilongshanren:
@fusijie 
Please check whether this issue is resolved or not in the upstream of cocos2d-x.
Thanks.

fusijie:
@andyque Thanks, it's fixed.

PR #11517, opened 2015-04-20
Fix widget gray shader issue

zilongshanren:
This PR fix issue https://github.com/cocos2d/cocos2d-x/issues/11511


zilongshanren:
I delete the shader config in UI folder and move grayscale shader as a builtin shader of cocos2d.
Now the grayscale shader would be cached and the memory consumption issue also be resolved.

One side effect of this solution:
The engine upgrade process would be a little hard, developers need to remove the old ui shader files from their project.

But I think it is more reasonable to move the ui shader part of the core of cocos2d.

BTW, sorry for the removing of extra whitespace in the git diff, I figure it out in my editor and it won't happen again...  It is annoying for code review.

PR #11523, opened 2015-04-21
V3 Windows Phone 8.1 back button support

stammen:
This pull request adds proper support for the back button event on Windows Phone. It follows the current cococs2d-x pattern of sending the EventKeyboard::KeyCode::KEY_BACKSPACE event. The game developer is responsible for listening for this event if they want to handle the back button.

The default behavior is to exit the app if the game does not handle the event. The developer MUST call    event->stopPropogation() on the event if they handled the event and don't want their app to quit.


kidproquo:
What is the easy way to update an existing v3.5 based project to get the back-button fix incorporated?

tests/cpp-tests/Classes/AppDelegate.cpp
```diff
@@ -165,12 +165,7 @@ bool AppDelegate::applicationDidFinishLaunching()
     
     fileUtils->setSearchPaths(searchPaths);
 
-#if (CC_TARGET_PLATFORM == CC_PLATFORM_WP8)
-    // a bug in DirectX 11 level9-x on the device prevents ResolutionPolicy::NO_BORDER from working correctly
     glview->setDesignResolutionSize(designSize.width, designSize.height, ResolutionPolicy::SHOW_ALL);
```
dumganhar, line 168: @stammen , do you know why it doesn't work correctly in `NO_BORDER` mode for DirectX 11 level9-x?
Doesn't it support that view port size is larger than device screen size?

stammen: There was a bug in the old WP8 GPU drivers that returned incorrect values when using negative positions for the viewport as calculated by cocos2d-x.

PR #3860, opened 2013-09-28
TMXXMLParser support for object 'rotation' property

andresmakesgames:
Added TMXXMLParser support for object 'rotation' property, available in latest daily builds of Tiled Map Editor. -A. Ortiz

PR #11540, opened 2015-04-21
TMX: Added support for reading the object.rotation attribute

### LINKED ISSUE #921: How can i get the "Rotation" information of one object
I set the value of the object's rotation, but i can't get the value of "Rotation" in code, so i want to know how can i get the "Rotation" information of the object.

bjorn:
Introduced with Tiled 0.10, this attribute stores the rotation of an object in degrees clockwise.

Closes bjorn/tiled#921

### DIFF
--- a/cocos/2d/CCTMXXMLParser.cpp
+++ b/cocos/2d/CCTMXXMLParser.cpp
@@ -466,6 +466,8 @@ void TMXMapInfo::startElement(void *ctx, const char *name, const char **atts)
         dict["width"] = Value(s.width);
         dict["height"] = Value(s.height);
 
+        dict["rotation"] = attributeDict["rotation"].asDouble();
+
         // Add the object to the objectGroup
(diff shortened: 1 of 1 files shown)


bjorn:
I've done it a little bit different as in #3860, to make sure the value has the appropriate type rather than string. I hope the PR does not get lost this time.

bjorn:
I've rebased my patch, which is still relevant. Any chance of it getting merged?

minggo:
What will happen if there is not rotation attribute, for example, old Tiled version doesn't have this attribute.

bjorn:
> What will happen if there is not rotation attribute, for example, old Tiled version doesn't have this attribute.

In case the attribute is missing, which is not only the case for old versions of Tiled but also when the object wasn't rotated, then the returned `Value` from `attributeDict["rotation"]` will have type ``NONE`` and `asDouble` will return 0.

minggo:
Got it, thanks.

Issue #8306, opened 2014-09-30
TestCpp error in RenderTextureTest, Testing depthStencil attachment on Chipset Tegra 3


minggo:
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/3672). It was created at 2014/01/11 23:33:45 +0000
# Description

Hi, this issue happens on Tegra 3 (ASUS TF300), Cocos2d-x 3.0 alpha1 (12 jan 2014)
Not happens on Nexus 7, Amazon Kindle HD, Galaxy s plus, Galaxy nexus, note 2.
Same issue happens with cocos2d-x 2.2.1 and 2.1.5 (but openGL errors are in different files)

TestCpp ~~> RenderTextureTest~~> Testing depthStencil attachment

No Circle appears, and LogCat shows this error

01-12 00:26:36.770: D/cocos2d-x debug info(8523): Assert failed: Could not attach texture to framebuffer
01-12 00:26:36.770: E/cocos2d-x assert(8523): /path/../cocos/2d/CCRenderTexture.cpp function:initWithWidthAndHeight line:275
01-12 00:26:36.800: D/cocos2d-x debug info(8523): OpenGL error 0x0500 in /path/../cocos/2d/CCTexture2D.cpp initWithMipmaps 622

I have also noticed, another error when testCpp goes in background, in this page:

01-12 00:29:50.550: D/cocos2d-x debug info(8523): Assert failed: only RGBA8888 can be saved as image
01-12 00:29:50.560: E/cocos2d-x assert(8523): /path/../cocos/2d/CCRenderTexture.cpp function:newImage line:428

thanks
Don
- Donmizzi added comment:
  I solved using this method:

https://github.com/blackberry/GamePlay/commit/a8656a693d4efade367f1f13076144638e18d2df

use GL_DEPTH_COMPONENT16 + GL_STENCIL_INDEX8 separately, in the init() of the CCRenderTexture (Cocos2d-x 2.1.5)

tigerwng:
我在3.6版本，当使用RenderTexture渲染了ClippingNode中的数据时也遇到了这个问题。看了描述中提到的解决方法，不过没大看懂。有什么解决方案吗？

pandamicro:
It will be checked in v3.9 test phase

PR #11553, opened 2015-04-22
Use packed depth and stencil buffer only when available

babcca:
Check `packed_depth_stencil` extension when use `RenderTexture`. On the most of the Nvidia tegra devices don't have this extension. When extension isn't available then use custom stencil buffer.

Currently it use only 16-bit depth buffer, because Nvidia Tegra doesn't support `GL_OES_depth24` too, but can be checked (need change `platform/{Android, iOS)/CCGL-{0}.h`)

Tested on Nexus7, Acer A500.


babcca:
Maybe Issue #8306

minggo:

PR #11571, opened 2015-04-23
Fix memory leak

super626:


minggo:
Please send it to v3.6 branch.

PR #11573, opened 2015-04-23
Sync Cocos2d-JS modifications to Cocos2d-x

pandamicro:


plugin
```diff
@@ -1 +1 @@
-Subproject commit 3a71e881f1b39b5b23ffcf32322e4139bf9350f7
+Subproject commit fbd9d50b94b1d472141d307e69f0e2c5f6215ec8
```
minggo, line 1: Is it modified by accident?

cocos/storage/local-storage/LocalStorage.h
```diff
@@ -46,7 +46,7 @@ void CC_DLL localStorageFree();
 void CC_DLL localStorageSetItem( const std::string& key, const std::string& value);
 
 /** Gets an item from the JS. */
-std::string CC_DLL localStorageGetItem( const std::string& key );
+bool CC_DLL localStorageGetItem( const std::string& key, std::string *outItem );
```
minggo, line 49: Why modify like this?
It will break compatibility.
Why not add a new function if it is needed.

plugin
pandamicro, line 1: It's for updating FacebookAgent versions

cocos/storage/local-storage/LocalStorage.h
pandamicro, line 49: The old API and implementation is wrong. It can't differentiate the situation when item doesn't exist and when item's value is empty string.
And the result is also confirmed different from the web localStorage result.

PR #11589, opened 2015-04-24
Fixed mainloop paused when slide webview on the iOS.

### LINKED ISSUE #10847: When slide the WebView, the main loop of engine is paused.
`version:`3.4Final
`platform:` iOS
`issue:`When slide the WebView, the main loop of engine is paused.
`reproduce:`Add following code to mainloop, then open the TestCpp - webview test, slided the webview and found that nothing output.

``` cpp
void DisplayLinkDirector::mainLoop()
{
    CCLOG("mainloop");
...
}
```

fusijie:
To fix #10847 .

Add `UITrackingRunLoopMode` mode, so that when the user to drag the UITableView/UIWebView/UIScrollView in UITrackingRunLoopMode mode, the mainloop of Cocos2d-x still goes on.


minggo:
@andyque 
Please take a look.

zilongshanren:
@fusijie  I'm not sure the impact of changing RunLoop in cocos2d-x. I would rather keep this PR here and leave the option for developers who are using UITableView/UIWebView and UIScrollView.
Thanks.

PR #11590, opened 2015-04-24
Animate3D Quality control

super626:
Add new quality type none which means that will not update animation data to bone.
It is useful when the Sprite3D is out of the screen.
If you are sure the Sprite3D moves out you can set its animation quality to none, so that it can save cpu time.


super626:
[ci rebuild]

PR #11593, opened 2015-04-24
return true when path is remote url

qsjia:
fix: TextureCache::addImage doesn't work when path is remote url


minggo:
@qsjia 
Currently there is not way to add an image from remote url.

PR #11598, opened 2015-04-26
Fix android http client memory leak

hueypark:


minggo:
@samuele3hu 
Please take a look.

samuele3hu:
[ci rebuild]

cocos/network/HttpClient-android.cpp
```diff
@@ -422,6 +422,7 @@ class HttpURLConnection
             jobject jObj = methodInfo.env->CallStaticObjectMethod(methodInfo.classID, methodInfo.methodID, jurl);
             _httpURLConnection = methodInfo.env->NewGlobalRef(jObj);
             methodInfo.env->DeleteLocalRef(jurl);
+            methodInfo.env->DeleteLocalRef(jObj);
```
samuele3hu, line 425: Why use DeleteLocalRef to delete a GlobalRef?

hueypark:
I think _httpURLConnection is GlobalRef and jObj is LocalRef.
And my project was shut down when call http client. (512 times)

WenhaiLin:
I agree with @HueyPark .

samuele3hu:
Ok,I know.I think it is no problem.

PR #10042, opened 2015-01-16
added check filename case in win32 FileUtils

qiutaoleo:
make win32 check filename case like linux

PR #11611, opened 2015-04-27
Check filename case on windows

minggo:
This pull request fix indent issue and translate Chinese comment into English based on #10042.


walzer:
Thanks @minggo, high efficiency! :+1: 
But should we add doxygen comment in its header file or somewhere?

minggo:
I don't think it needs doxygen comment because it is implemented in .cpp.

PR #11622, opened 2015-04-28
Use TextBMFont when text is updated from slider for faster performance

stammen:
This pull request replaces the use of ui::Text with ui::TextBMFont for the text field updated by the slider. ui::Text is too slow to handle all of the update events from the Slider.

ui::Text should not be used for text fields that need to be updated in real time.

Note: We need a good Helvetica or Marker Felt bm font in the fonts folder in resources.


minggo:
[ci rebuild]

minggo:
> Note: We need a good Helvetica or Marker Felt bm font in the fonts folder in resources.

Could you please explain why?

stammen:
I had to search all of the fonts in Resources to find a clean white version of MarkerFelt. The closest I could find was ccb/markerfelt24shadow.fnt.

minggo:
Can not use other font on WP?

PR #11626, opened 2015-04-28
Use TextBMFont when text is updated from slider for faster performance

stammen:
This pull request replaces the use of ui::Text with ui::TextBMFont for the text field updated by the slider. ui::Text is too slow to handle all of the update events from the Slider.

ui::Text should not be used for text fields that need to be updated in real time.


minggo:
@stammen 
Thanks.
We will merge v3.6 back to v3 after v3.6 is released.
So i think we can close this PR.

PR #11639, opened 2015-04-28
Close  Warning and fix bug on reading cocostudio file as percent mode

lbfamous:
1. CloseWarning C4018
2. 1. fix bug on reading cocostudio file as Percent Mode

### DIFF
--- a/cocos/platform/CCPlatformConfig.h
+++ b/cocos/platform/CCPlatformConfig.h
@@ -154,7 +154,6 @@ THE SOFTWARE.
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
 #ifndef __MINGW32__
 #pragma warning (disable:4127) 
-#pragma warning (disable:4018)
 #endif 
 #endif  // CC_PLATFORM_WIN32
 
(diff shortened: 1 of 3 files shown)


cocos/platform/CCPlatformConfig.h
```diff
@@ -154,6 +154,7 @@ THE SOFTWARE.
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
 #ifndef __MINGW32__
 #pragma warning (disable:4127) 
+#pragma warning (disable:4018)
```
minggo, line 157: What's the type of warning?
Why disable it?

lbfamous, line 157: warning C4018: '<' : signed/unsigned mismatch.

zilongshanren, line 157: @lbfamous  

 signed/unsigned mismatch would make our code base error-prone, I think we should fix the issue rather than disable the warning.
@minggo  Thoughts?

lbfamous, line 157: just on windows.and I see it from cocos2dx-0.x to now.never have seen anybody fixed it.Are you sure it will be fixed.And I think if you want to fix it ,then you can open it that day.. @andyque

zilongshanren:
Thanks @lbfamous  for your contribution.

But this PR contains many unneeded commits.

qq20150429-1 2x

lbfamous:
how to delete it.

zilongshanren:
@lbfamous 
Please google "git merge squash" or "git rebase" to reorganize your commits.

lbfamous:
thanks.i will

Issue #10688, opened 2015-03-02
Label::enableGlow() bug.


fusijie:
`version:` 3.4Final
`issue:` While create a menuItem with MenuItemLabel, then enable the Label's glow feature, the position of the menuItem will shift. and the touch position is shift too.
![399a4b2d-be9f-47ea-94d7-43038167324c](https://cloud.githubusercontent.com/assets/5085300/6435079/bb2a9ca6-c0d0-11e4-877d-5fcdbe0dc84f.jpg)
`reproduce:` add following code to HelloWorld project.

``` cpp
    auto label1 = Label::createWithTTF("Play", "fonts/Marker Felt.ttf", 100);
    auto menuItem1 = MenuItemLabel::create(label1, [](Ref*){});
    auto label2 = Label::createWithTTF("Play", "fonts/Marker Felt.ttf", 100);
    label2->enableGlow(Color4B(255, 255, 0, 255));
    auto menuItem2 = MenuItemLabel::create(label2, [](Ref*){});

    auto mymenu = Menu::create(menuItem1,menuItem2,NULL);
    this->addChild(mymenu);
    mymenu->setPosition(Vec2(300, 450));
    mymenu->alignItemsVerticallyWithPadding(50);
```

kdudnyk:
It`s seems that issue is due to :
const int Label::DistanceFieldFontSize = 50;
being setup in CCLabel.cpp 
while creating FontAtlas for Label with efffect in :
CCFontAtlasCache.cpp :
    if (useDistanceField)
    {
        fontSize = Label::DistanceFieldFontSize / contentScaleFactor;
    }
the font size is being downscaled by 2 ( 50%) ...
I didnt find any sens in DistanceFieldFontSize  being set to 50...
but if you change it to 100 - the effect seems to be applied right :)
sent pull request https://github.com/cocos2d/cocos2d-x/pull/11640

qiululu:
also see in CCLabel.cpp 

        if(ttfConfig.distanceFieldEnabled)
        {
            this->setFontScale(1.0f * ttfConfig.fontSize / DistanceFieldFontSize);  
        }

to set the scale size of distantFieldFont ,we do get a fixed size bitmap from freetype to align the text,and compute the content size, and then scale it  in opengl render pipline, fragment shader, in this situation ,it seem that compute the size of the fixed one, not the real size of the finall effect.  and a unfinished thought ,I tried this:  overide the setContentSize in Label:

[...]

PR #11640, opened 2015-04-28
fixing label being missaligned after glow effect applied (bug 10688)

kdudnyk:
Due to DistanceFieldFontSize being set to 50 during creating font atlas for label with glow effect 
label size is wrongly calculated. After applying 100 value all effects on label seems to work fine,
so do the correct allignment, please reffer to bug  #10688 for details 
(https://github.com/cocos2d/cocos2d-x/issues/10688)

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -40,7 +40,7 @@
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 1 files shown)


minggo:
@WenhaiLin 
Please take a look.

pandamicro:
This have been fixed at https://github.com/cocos2d/cocos2d-x/pull/13084

PR #11644, opened 2015-04-29
Fix Jenkins Xcode build return value

zilongshanren:
xcpretty command line tool has a bug... So I add another extra xcodebuild . Since Xcode will cache the previous build result, so the second xcodebuild command just to verify the build result.


zilongshanren:
@minggo This PR fixed the xcodebuild sometimes returns wrong value when use xcpretty.

PR #11651, opened 2015-04-29
VideoPlayer:Fixed can't play video and may place video in the wrong place on Android 2.3.x 

WenhaiLin:


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxVideoView.java
```diff
@@ -343,7 +344,13 @@ public void fixSize() {
     }
     
     public void fixSize(int left, int top, int width, int height) {
-        if (width != 0 && height != 0) {
+    	if (mVideoWidth == 0 || mVideoHeight == 0) {
+    		mVisibleLeft = left;
+            mVisibleTop = top;
+            mVisibleWidth = width;
+            mVisibleHeight = height;
+		}
```
minggo: wrong indent

Issue #11607, opened 2015-04-27
[v3.6-universal-wp8.1-cpptests]Node: Draw:Crash


WenhaiLin:
Device:Nokia Lumia 625
OS: Window Phone 8.1

WenhaiLin:
Console output:

D3D11: Removing Device.
线程 0xbd0 已退出，返回值为 0 (0x0)。
线程 0xdf8 已退出，返回值为 0 (0x0)。
reload all texture
Dirty Uniform and Attributes of GLProgramState
Dirty Uniform and Attributes of GLProgramState
Dirty Uniform and Attributes of GLProgramState
D3D11: Removing Device.
线程 0x7e8 已退出，返回值为 0 (0x0)。
线程 0x3c0 已退出，返回值为 0 (0x0)。
0x773835D7 处(位于 cpp-tests.WindowsPhone.exe 中)最可能的异常: Microsoft C++ 异常: _com_error，位于内存位置 0x0222CBD8 处。
0x773835D7 (KERNELBASE.DLL) (cpp-tests.WindowsPhone.exe 中)处的第一机会异常:  0x40080201: WinRT originate error (参数:  0x80004005, 0x00000018, 0x0222F0D8)。
0x773835D7 处(位于 cpp-tests.WindowsPhone.exe 中)最可能的异常: Microsoft C++ 异常: Platform::FailureException ^，位于内存位置 0x0222F754 处。 HRESULT:0x80004005 Unspecified error
WinRT 信息: Failed to initialize EGL
堆栈跟踪:

> [外部代码]
> cpp-tests.WindowsPhone.exe!OpenGLES::Initialize() 行 152
> cpp-tests.WindowsPhone.exe!OpenGLES::Reset() 行 189
> cpp-tests.WindowsPhone.exe!cocos2d::OpenGLESPage::RecoverFromLostDevice() 行 254
> cpp-tests.WindowsPhone.exe!cocos2d::OpenGLESPage::StartRenderLoop::__l31::void <lambda>(Windows::Foundation::IAsyncAction ^)::__l55::<lambda>() 行 346
> [外部代码]

WenhaiLin:
https://github.com/cocos2d/cocos2d-x/issues/11606 cause by the same reason.

stammen:
@WenhaiLin @minggo I am not able to reproduce this issue on my phones or in the emulator. I do not have a Lumia 625. Would you please send me more information.

minggo:
[...]

Issue #11606, opened 2015-04-27
[v3.6-universal-wp8.1-cpptests]Node: Clipping-->Scroll View Demo:Crash


WenhaiLin:
Device:Nokia Lumia 625
OS: Window Phone 8.1

minggo:
@andyque 
@stammen
Could you please take a look?

stammen:
@WenhaiLin  @minggo I am not able to reproduce this issue on my phones or in the emulator. I do not have a Lumia 625. Would you please send me more information.

stammen:
Fixed in pull request https://github.com/cocos2d/cocos2d-x/pull/11652

PR #11652, opened 2015-04-29
V3.6 Windows Phone 8.1 fix for issues 11606 and 11607

stammen:
This pull request fixes issues 11606 and 11607 on Windows Phone 8.1

https://github.com/cocos2d/cocos2d-x/issues/11607
https://github.com/cocos2d/cocos2d-x/issues/11606

Angle for Windows Phone 8.1 is not able to correctly compile shaders with gl_Pointsize on all devices. I have confirmed this issue with the Microsoft Angle team. This pull request disables gl_Pointsize in a manner similar to v3.4 and v3.5. 

The Angle team will be working to fix this problem, hopefully in time for v3.7. https://github.com/MSOpenTech/angle/issues/37

Note: An warning message will be printed by Angle to the debug output window if the user tries to use gl_PointSize in their shader.


minggo:
Thanks @stammen .
But i think it is better not add wp special codes other than in platform folder anymore. 
Cann't it be resolved in Angle?

minggo:
Ok, i found that you created an issue for Angle.
Let's move the codes when Angle supports it.

PR #11659, opened 2015-04-30
Label:Fixed crash.

WenhaiLin:
When you invoking `getLetter(letterIndex)` and `setString(text)` in sequence, and `letterIndex` greater than the length of `text`, it might crash your program.


minggo:

minggo:
[ci rebuild]

PR #11673, opened 2015-05-01
Fix warnings in v3.

vovkasm:
Fix most warnings that reported by Xcode. Most of them are incorrect field initialization order. But some of them is about unused entities.

After that changes, a project generated by cmake has these warnings count:
cocos2d - 16 (before was more 50. Remain warnings are hard to fix quickly.)
luacocos2d - 411 (i did not fix these warnings because there are autogenerated files)
cpp-empty-test - 0
cpp-tests - 0
lua-empty-test - 0
lua-tests - 1


minggo:
Thanks @vovkasm .

PR #11676, opened 2015-05-03
Set the texture format on the specified image file

xinjuncoding:
We need to set the texture format  when the asynchronous loading resource, and change back to default in callbacks, so,  if when the asynchronous loading and synchronous loading in the same time, the texture format settings will not get the result we wanted, so the increase or the set of texture format of the specified file is necessary.

当前在异步加载的时候设置纹理需要在回调的时候才能改回默认值，如果在异步加载的时候又去同步加载，那么纹理格式的设置将得不到我们想要的结果，所以增加这种对指定文件进行设置纹理格式是很有必要的，否则，异步加载将形同虚设。

### DIFF
--- a/cocos/renderer/CCTextureCache.cpp
+++ b/cocos/renderer/CCTextureCache.cpp
@@ -219,7 +219,17 @@ void TextureCache::loadImage()
             const std::string& filename = asyncStruct->filename;
             // generate image      
             image = new (std::nothrow) Image();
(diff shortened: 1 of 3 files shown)


zilongshanren:
@xinjuncoding  Could you change the title of this PR to English?  And you'd better provide a English version of PR description. You could use Google Translate to do this.

BTW: You can also keep the Chinese description here.

Thanks.

xinjuncoding:
I has updated the title and PR description.

PR #11677, opened 2015-05-03
Fix setSprite never updating progress on ProgressTimer

tankorsmash:
Basically whenever you set a new sprite on a ProgressTimer, the new sprite is hidden.

 Calling setPercentage once with the same value doesn't work because the new percentage is the same as the old one. I've got a workaround setting the percentage to 0 and then back to the old value, but this change seems to solve that issue entirely.


WenhaiLin:
@tankorsmash You are right.Thank you!

tankorsmash:
Awesome

minggo:
Thanks @tankorsmash .

PR #11700, opened 2015-05-05
Material system

ricardoquesada:
Adds MaterialSystem to cocos2d-x


ricardoquesada:
canceling... adding a new one with just one commit

PR #11704, opened 2015-05-05
V3: WinRT updated output name to libcocos2d_v3.7

stammen:
This pull request updates the Windows 8.1 Universal App DLL names to libcocos2d_v3.7


minggo:
Why use version info in target name?
Then it should be changed every version.

stammen:
It is a request of the Windows Store for cocos2d-x version statistics.

minggo:
Got it.

PR #11728, opened 2015-05-07
Label: Fixed position is wrong if label content is changed after invoking `getLetter(letterIndex)`

WenhaiLin:


WenhaiLin:
[ci rebuild]

PR #11731, opened 2015-05-07
Fix a bug: to handle big frame packet

cnsoft:
Ensure trigger message read event after got whole frame packet.


samuele3hu:
Sorry, i didn't understand the result for this pr.Can you give the test cases about this pr?thanks.

cnsoft:
@samuele3hu  in some case, the websocket packet of one frame will trigger the data received event more than once time, but, current implement in cocos , it will cause one whole packet is splitted into two message data. this make high level logic confused.

samuele3hu:
Sorry for replying late.I have known your meaning,but how can I reproduce this bug?

zilongshanren:
@cnsoft  @samuele3hu  Any updates?

PR #11762, opened 2015-05-08
AudioEngine:Fixed program may freeze if `AudioEngine::stop` or `AudioEngine::stopAll()` is invoked frequently on Android.

WenhaiLin:
This solution is not perfect, by this method the probability of `freeze` is greatly reduced.

### DIFF
--- a/cocos/audio/android/AudioEngine-inl.cpp
+++ b/cocos/audio/android/AudioEngine-inl.cpp
@@ -365,7 +365,11 @@ void AudioEngineImpl::stop(int audioID)
         log("%s error:%u",__func__, result);
     }
 
-    //If destroy openSL object immediately,it may cause dead lock.
+    /*If destroy openSL object immediately,it may cause dead lock.
+     *It's a system issue.For more information:
+     *    https://github.com/cocos2d/cocos2d-x/issues/11697
+     *    https://groups.google.com/forum/#!msg/android-ndk/zANdS2n2cQI/AT6q1F3nNGIJ
+     */
     player._delayTimeToRemove = DELAY_TIME_TO_REMOVE;
     //_audioPlayers.erase(audioID);
 }
--- a/cocos/audio/android/AudioEngine-inl.cpp
+++ b/cocos/audio/android/AudioEngine-inl.cpp
@@ -43,6 +43,8 @@
 using namespace cocos2d;
 using namespace cocos2d::experimental;
 
+#define DELAY_TIME_TO_REMOVE 0.5f
+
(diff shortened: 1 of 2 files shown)


WenhaiLin:
[ci rebuild]

WenhaiLin:
[ci rebuild]

kakamatong:
Still happen

BlackStar0313:
still happen , test device  nexus6 ,Android 6.0     ,

my describe: http://www.cocoachina.com/bbs/read.php?tid=1679095&page=1&toread=1#tpc

dumganhar:
@BlackStar0313 , thank you for your report, I'll check it.

Jayatubi:
Still happen on GiNOEE F103/amigo3.0.7/Android 5.0

dumganhar:
@Jayatubi, could you try the Pull Request here (https://github.com/cocos2d/cocos2d-x/pull/15875)? I refactored something of AudioEngine for Android.

Jayatubi:
@dumganhar Thanks for your reply but I'm still using the cocos2dx 2.2.6 so I can't merge you pull.

dumganhar:
@Jayatubi , sorry for that we aren't maintaining the v2 branch.
But `AudioEngine` is a new class added from v3.0, if it's a bug of `CocosDension`/ `SimpleAudioEngine`, please just create another issue.

Jayatubi:
@dumganhar Could you provide any hints about how you fix this issue on 3.x? Maybe I can try the same solution on my 2.2.6.

BTW I'm currently using the 2.2.6 with the refactoring from https://github.com/WenhaiLin/cocos2d-x/commit/be0ac68a5e8512c665956acfab90ba5c4e18686f to avoid Android 5 AwesomePlayer issue.

dumganhar:
My PR(#15875) uses a different approach to play effects.
The problem may be resolved and I will do more test on it. :)
What're `AwesomePlayer issue`?

Jayatubi:
@dumganhar Oops it seems I messed up the name should be `Nuplayer`.

In the early releases of the Android 5 the `Nuplayer`, which was experimental and buggy, has been set as the default audio engine for the `SoundPool`. It had a very terrible issue that it will cause the app freeze for a few seconds when load short audio clips. To walk around the audio engine could be changed back to the old `AwesomePlayer` in the `Developer options`.

However, I couldn't tell all my APP users to do such things because they might not understand what was going to do so I have to change all implementation from `SoundPool` to the `OpenSLES`.

dumganhar:
Mmm, but we don't have enough time to continue maintaining v2 branch...
You could refer to my PR for what I changed.

PR #11763, opened 2015-05-08
Fix web view leak on iOS

Dimon4eg:


WenhaiLin:
@Dimon4eg Thank you.

PR #11764, opened 2015-05-08
Fixed crash - ogg filepath has unicode in Win32.

cpascal:
In Win32, you can't play ogg file if the path contains unicode characters like CJK.

mp3(mpg123) library accepts utf8 filepath, but ogg(Vorbisfile) library
accepts ACP(ANSI CodePage) in Win32 environment.


WenhaiLin:
@cpascal We have fixed the crash problem: https://github.com/cocos2d/cocos2d-x/pull/11679.
@minggo This pr should be close.

PR #11767, opened 2015-05-08
Change widget::addEventListener() to support multiple event listeners

enhhh:
change Widget::addEventListener() from only one event listener to multiple event listeners

We may need to add an Event Listener anytime( for example: newbie guide) and remove an Listener anytime.
so I change `_touchEventCallback` to an `std::map`, key is an `uint32_t`,return key after add an event listener, and use it to remove event listener.

### DIFF
--- a/build/cocos2d-js-win32.vc2013.sln
+++ b/build/cocos2d-js-win32.vc2013.sln
@@ -14,6 +14,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libjscocos2d", "..\cocos\sc
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "js-tests", "..\tests\js-tests\project\proj.win32\js-tests.vcxproj", "{D0F06A44-A245-4D13-A498-0120C203B539}"
 EndProject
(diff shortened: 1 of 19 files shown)


zilongshanren:
Thanks @enhhh 
Could you provide a test case for your addition? In this way we could see some practical use case.

cocos/ui/UIWidget.cpp
```diff
@@ -832,10 +833,10 @@ void Widget::onTouchCancelled(Touch *touch, Event *unusedEvent)
 void Widget::pushDownEvent()
 {
     this->retain();
-    if (_touchEventCallback)
-    {
-        _touchEventCallback(this, TouchEventType::BEGAN);
-    }
+	for (auto &iter : _touchEventCallbacks)
```
zilongshanren: the indentation is wrong here.

cocos/ui/UIWidget.cpp
```diff
@@ -899,9 +900,38 @@ void Widget::addTouchEventListener(Ref *target, SEL_TouchEvent selector)
     _touchEventSelector = selector;
 }
 
-void Widget::addTouchEventListener(const ccWidgetTouchCallback& callback)
+uint32_t Widget::addTouchEventListener(const ccWidgetTouchCallback &callback)
+{
+	uint32_t key = 0;
+	if (!_touchEventCallbacks.empty())
```
zilongshanren: please use 4 spaces as indentation, it seems that all of your code use 8 spaces as indentation.

enhhh:
I am so sorry about the indentation. Do I need to add a new comment ? 

I had written a  blog about the use of my addition(in Chinese).
http://enhhh.github.io/2015/05/12/about-newbie-guide/

PR #11820, opened 2015-05-13
Fix crash in web view if illegal character in url (Android only)

Dimon4eg:
URI.create can throw exceptions, in my case was IllegalArgumentException. We should handle it.


minggo:
Thanks @Dimon4eg .

PR #11836, opened 2015-05-14
V3 camera mask for2 d and3 d rendering

dabingnn:
This pull request is opened for mixing 2D and 3D rendering by using camera mask.
Work to do before merge
1. make code runs on android, windows and linux

**Update**: runs on all platform


dabingnn:
@ricardoquesada @minggo 
This pull request solves the 2D and 3D mixing rendering problem by using multiple camera and camera masks.
Could you please review it? Thanks.

tools/cocos2d-console
```diff
@@ -1 +1 @@
-Subproject commit ec36c46e0a7e82d8e6c441d478f01dcd46377fbb
+Subproject commit fe7ccc496d507c66dbfef2907fec9ac39d737300
```
minggo: Please don't modify submodule.

minggo:
@super626 
Could you please help to review it too?
Thanks.

cocos/renderer/CCGLProgramCache.cpp
```diff
@@ -483,6 +495,14 @@ void GLProgramCache::loadDefaultGLProgram(GLProgram *p, int type)
         case kShaderType_3DTerrain:
             p->initWithByteArrays(cc3D_Terrain_vert, cc3D_Terrain_frag);
             break;
+        case kShaderType_CameraClear:
+            p->initWithByteArrays(ccCameraClearVert, ccCameraClearFrag);
+            break;
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WP8 || defined(WP8_SHADER_COMPILER)
```
minggo: CC_PLATFORM_WP8 is not supported now, and corresponding macros are removed too.
So please don't use this macro.

cocos/2d/CCCamera.cpp
```diff
@@ -361,4 +365,76 @@ void Camera::setScene(Scene* scene)
...
+    
+    quad.bl.colors = quad.br.colors = quad.tl.colors = quad.tr.colors = Color4B(0,0,0,1);
+    
+    quad.bl.texCoords = Tex2F(0,0);
+    quad.br.texCoords = Tex2F(1,0);
+    quad.tl.texCoords = Tex2F(0,1);
+    quad.tr.texCoords = Tex2F(1,1);
+    
+    auto shader = GLProgramCache::getInstance()->getGLProgram(GLProgram::SHADER_CAMERA_CLEAR);
+    auto programState = GLProgramState::create(shader);
```
super626: Does the glClear work for your purpose? It should run fast than this.

Also GLProgramState::getOrCreateWithGLProgram(shader) can avoid to create GLProgramState every frame in your case.

super626:
@minggo @dabingnn 
I like this work. This is the multiple camera idea we want to solve the problem mixing of 2d and 3d objects. The render queues are complex and it is incomplete. 

To improve:
Camera contain a view port, and then it can clear the view port only.

However, users may complain that it is difficult to use. I have an idea like this,
Layer::visit()
{
      if (shouldClearBackGround)
      renderer->addClearCommand();
      visit();//visit the layer and its child
}

The render queue,
ClearCommand layer1_command1 layer1_command2 .... ClearCommand layer2_command1 layer layer2_command2 ...

We should sort the command between ClearCommand. So that we can make sure that layer2 is on top of layer1.
We can also merge the clear command to the draw function of layer, for example ColorLayer can do drawing together with clear depth.

cocos/2d/CCCamera.cpp
dabingnn: glClear are based on Render Target, but we need to do is to clear a viewport, though we do not have this then. So `glClear` is not the thing we need.
`GLProgramState::getOrCreateWithGLProgram(shader)` would be better, thanks.

PR #11841, opened 2015-05-14
shorten csb file name in ccs-res

xiaofeng11:


minggo:
@samuele3hu 
Please help to check the modification, and please check corresponding lua test cases.
Thanks.

samuele3hu:
I have checked the modification, all are ok except the csb file name used in the `UILayoutComponentTest_Editor`  test weren't modified, i would update it. The related Lua test cases pr is https://github.com/cocos2d/cocos2d-x/pull/11858

Issue #11569, opened 2015-04-23
Support .9.png format from android


walzer:
- support .9.png format, get its source code from android
- offer its API from `Scale9Sprite`.

PR #11850, opened 2015-05-15
 Add Android 9-patch image format support.

zilongshanren:
The 9-patch format is used for creating Scale9Sprite.

This PR fixed issue https://github.com/cocos2d/cocos2d-x/issues/11569


zilongshanren:
I will send another PR with squashed commit.

PR #11856, opened 2015-05-15
Add Android Studio support

WenhaiLin:


templates/js-template-default/frameworks/runtime-src/proj.android-studio/HelloJavascript/src/org/cocos2dx/javascript/AppActivity.java
```diff
@@ -0,0 +1,42 @@
+/****************************************************************************
+Copyright (c) 2008-2010 Ricardo Quesada
+Copyright (c) 2010-2012 cocos2d-x.org
+Copyright (c) 2011      Zynga Inc.
+Copyright (c) 2013-2014 Chukong Technologies Inc.
```
minggo: Please modify the license file.

PR #11860, opened 2015-05-15
Added a low-pass-filter to the PhysicsWorld::update method.

jimrange:
Added a low-pass-filter to the PhysicsWorld::update method to make the physics simulation run more smoothly.

The filter is by default turned off, due to the default value of _updateFilterFactor being 1.0 in the PhysicsWorld  constructor.

The filter can be turned on by setting the filter value like this:

```
float filterFactor = 0.1;
currentScene->getPhysicsWorld()->setUpdateFilterFactor(filterFactor);
```

This has been tested on Android 4.4.3 Nexus 7, iOS 8.1.3 Ipad Mini Retina, iOS 8.3 iPhone 5.
Without the low-pass-filter, a fast paced 2D side scrolling racing game physics simulation using the built-in physics might see a jerky or jittery sprite movement of sprites controlled by physics bodies.

The jittery simulation is a result of the physics update loop having variable dt values.

More info about this proposed fix: http://discuss.cocos2d-x.org/t/why-jitter-jerk-when-following-sprite-with-layer-video-included/16799/15

More info about a discussion about alternative fixes should this patch not be applied: http://discuss.cocos2d-x.org/t/does-the-built-in-physics-engine-physicsworld-update-need-a-low-pass-filter-for-dt/21534

### DIFF
--- a/cocos/physics/CCPhysicsWorld.h
+++ b/cocos/physics/CCPhysicsWorld.h
@@ -286,6 +286,14 @@ public:
     inline int getSubsteps() const { return _substeps; }
 
     /**
+     * Set the update filter factor.
+     *
+     * Value must be between FLT_EPSILON and 1.0, where 1.0 applies no filter to dt and FLT_EPSILON applies maximum filter to dt. The filterFactor is a parameter for a low-pass-filter that is used to smooth out the dt of the physics update loop. Setting this valuie to 1.0 will turn the filter off. Setting the value to less than 1.0 will turn the filter on. e.g. Setting filterFactor to 0.15 will apply a significant low-pass-filter to smooth out the delta time step dt and as a result smooth out the phyiscs simulation.
+     * @param filterFactor A float number, default value is 1.0. Valid values are between FLT_EPSILON and 1.0.
+     */
+    void setUpdateFilterFactor(float filterFactor);
+    
+    /**
     * Set the debug draw mask of this physics world.
     * 
     * This physics world will draw shapes and joints by DrawNode acoording to mask.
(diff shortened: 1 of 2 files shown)


cocos/physics/CCPhysicsWorld.h
```diff
@@ -358,6 +366,8 @@ class CC_DLL PhysicsWorld
     float _updateTime;
     int _substeps;
     cpSpace* _cpSpace;
+    float _updateFilteredDelta;
```
andyque: Could we remove the member variable `_updateFilteredDelta` and leave it a local variable in the `world::step` method?

zilongshanren:
@WenhaiLin  please take a look at this PR, thanks.

jimrange:
This pull request might not be necessary since the manual step interface of PhysicsWorld enables a user of the API to implement their own custom step filtering or fixed time step. 

However, it might be a good idea to consider adding this feature and/or a fixed time step to the PhysicsWorld class. These features would give users of the built-in physics some tools to more easily tune their physics simulation. 

I believe that adding such features to the physics world would make using the built in physics engine and cocos2d-x in general more appealing to a larger user base.

ricardoquesada:
@jimrange  yes. thanks for the suggestion. we need more stable dt. When people create the PhysicsWorld it could pass what kind of stepper they want to use.

TheCodez:
@WenhaiLin @ricardoquesada Any news on this? Is this considered to be taken?

ricardoquesada:
@thecodez: it is not my call whether to include it or not, but a low pass filter is useful.

PR #11863, opened 2015-05-16
Added new audio engine with wav support

anniruddh:


stammen:
The Linux build seems to have failed for an unrelated reason. This code is from MSOpenTech and is good to pull once the Linux build passes.

minggo:
Sorry.
There is a conflict.
Could you please fix it?
Thanks.

minggo:
Why use so many wav files to test?

anniruddh:
Currently this code supports only wav files hence added those files so that the tests can work. Support for MP3 and Ogg is in progress. Once that is in place these wav files can be dropped.

anniruddh:
Closing this PR for now. Will make a new PR with all the issues resolved.

PR #11878, opened 2015-05-18
Assertions for unexpected behavior and Memory Leak fixes

hobogenized:
This patch is meant to fix several memory leaks present in the current Cocos2d-x framework as well as provide assertions for when we get unexpected behavior. Possible unexpected behaviors were found through Xcode's static code analysis.

### DIFF
--- a/cocos/editor-support/cocostudio/CCSGUIReader.cpp
+++ b/cocos/editor-support/cocostudio/CCSGUIReader.cpp
@@ -1512,6 +1512,10 @@ Widget* WidgetPropertiesReader0300::widgetFromJsonDictionary(const rapidjson::Va
                         }
                         child->setPosition(Vec2(child->getPositionX() + widget->getAnchorPointInPoints().x, child->getPositionY() + widget->getAnchorPointInPoints().y));
                     }
+                    if (widget == NULL) {
+                        CCLOG("widget unexpectedly null!\n");
+                        assert(false);
+                    }
                     widget->addChild(child);
                 }
             }
--- a/cocos/editor-support/cocostudio/CCSSceneReader.cpp
+++ b/cocos/editor-support/cocostudio/CCSSceneReader.cpp
@@ -364,6 +364,10 @@ cocos2d::Node* SceneReader::createObject(CocoLoader *cocoLoader, stExpCocoNode *
     {
         className = pNodeArray[1].GetValue(cocoLoader);
     }
+    if (className == nullptr) {
+        CCLOG("className unexpectedly a nullptr!\n");
+        assert(false);
+    }
     if(strcmp(className, "CCNode") == 0)
     {
         Node* gb = nullptr;
@@ -443,6 +447,10 @@ cocos2d::Node* SceneReader::createObject(CocoLoader *cocoLoader, stExpCocoNode *
         setPropertyFromJsonDict(cocoLoader, cocoNode, gb);
         for (std::vector<Component*>::iterator iter = _vecComs.begin(); iter != _vecComs.end(); ++iter)
         {
+            if (gb == NULL) {
+                CCLOG("gb unexpectedly null!\n");
+                assert(false);
+            }
             gb->addComponent(*iter);
         }
         
--- a/cocos/network/HttpAsynConnection.m
+++ b/cocos/network/HttpAsynConnection.m
@@ -51,7 +51,7 @@
     [sslFile release];
     [responseHeader release];
     [responseData release];
-    [statusString release];
+    //[statusString release];
     [responseError release];
     [conn release];
     [runLoop release];
@@ -180,7 +180,12 @@
         SecTrustSetExceptions(serverTrust, errDataRef);
         
         SecTrustEvaluate(serverTrust, &trustResult);
+        [(id)errDataRef release];
+        
     }
+    
+    [certData release];
+    [(id)certArrayRef release];
     
     //Did our custom trust chain evaluate successfully?
     return trustResult = kSecTrustResultUnspecified || trustResult == kSecTrustResultProceed;    
--- a/cocos/ui/UIWidget.cpp
+++ b/cocos/ui/UIWidget.cpp
@@ -558,6 +558,10 @@ void Widget::setTouchEnabled(bool enable)
     if (_touchEnabled)
     {
         _touchListener = EventListenerTouchOneByOne::create();
+        if (!_touchListener) {
+            CCLOG("Unable to create touch listener!\n");
+            assert(false);
+        }
         CC_SAFE_RETAIN(_touchListener);
         _touchListener->setSwallowTouches(true);
         _touchListener->onTouchBegan = CC_CALLBACK_2(Widget::onTouchBegan, this);
--- a/extensions/assets-manager/Manifest.cpp
+++ b/extensions/assets-manager/Manifest.cpp
@@ -499,11 +499,12 @@ void Manifest::loadManifest(const rapidjson::Document &json)
 
 void Manifest::saveToFile(const std::string &filepath)
 {
-    rapidjson::StringBuffer buffer;
+    rapidjson::StringBuffer buffer = NULL;
     rapidjson::PrettyWriter<rapidjson::StringBuffer> writer(buffer);
     _json.Accept(writer);
     
     std::ofstream output(filepath, std::ofstream::out);
+    
     if(!output.bad())
         output << buffer.GetString() << std::endl;
 }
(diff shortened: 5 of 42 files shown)


felixjingga:
hopefully the studio team can add feature to support this too.. :)

cocos/editor-support/cocostudio/CCSGUIReader.cpp
```diff
@@ -1512,6 +1512,10 @@ Widget* WidgetPropertiesReader0300::widgetFromJsonDictionary(const rapidjson::Va
                         }
                         child->setPosition(Vec2(child->getPositionX() + widget->getAnchorPointInPoints().x, child->getPositionY() + widget->getAnchorPointInPoints().y));
                     }
+                    if (widget == NULL) {
```
zilongshanren, line 1515: This line is useless since we only enter this line when widget is not null.

cocos/editor-support/cocostudio/CCSSceneReader.cpp
```diff
@@ -364,6 +364,10 @@ cocos2d::Node* SceneReader::createObject(CocoLoader *cocoLoader, stExpCocoNode *
     {
         className = pNodeArray[1].GetValue(cocoLoader);
     }
+    if (className == nullptr) {
```
zilongshanren, line 367: This is also the same. className == nullptr is ok here.

tests/cpp-tests/Resources/ccs-res
```diff
@@ -1 +1 @@
-Subproject commit 0f5e0dff1a7545eef7bbe8f23d4b232d0d2d6ae3
```
zilongshanren, line 1: Why do you change the submodule?

cocos/ui/UIWidget.cpp
```diff
@@ -558,6 +558,10 @@ void Widget::setTouchEnabled(bool enable)
     if (_touchEnabled)
     {
         _touchListener = EventListenerTouchOneByOne::create();
+        if (!_touchListener) {
```
zilongshanren, line 561: When will the `create` method return nullptr? I think it's rare case. Could you provide a test case to verify it?

extensions/assets-manager/Manifest.cpp
```diff
@@ -499,11 +499,12 @@ void Manifest::loadManifest(const rapidjson::Document &json)
 
 void Manifest::saveToFile(const std::string &filepath)
 {
-    rapidjson::StringBuffer buffer;
+    rapidjson::StringBuffer buffer = NULL;
```
zilongshanren, line 502: `rapidjson::StringBuffer` is a generic class, why do you assign `NULL` to it?

cocos/network/HttpAsynConnection.m
```diff
@@ -180,8 +180,13 @@ - (BOOL) shouldTrustProtectionSpace:(NSURLProtectionSpace*)protectionSpace
         SecTrustSetExceptions(serverTrust, errDataRef);
         
         SecTrustEvaluate(serverTrust, &trustResult);
+        [(id)errDataRef release];
+        
     }
     
+    [certData release];
+    [(id)certArrayRef release];
```
zilongshanren, line 188: I think there is no need to release `certArrayRef`. Refer https://developer.apple.com/library/mac/documentation/CoreFoundation/Reference/CFArrayRef/index.html#//apple_ref/c/func/CFArrayCreate  The `CFArrayCreate` follows the `Create Rule ownership`. It is returned as a autorelease object. 
So I think you don't need to call release explicit. And you event call release twice.

zilongshanren:
I have opened a new PR to resolve the memory issue in `AudioEngine`. 
Thanks for contributing.
https://github.com/cocos2d/cocos2d-x/pull/12076

cocos/editor-support/cocostudio/CCSGUIReader.cpp
hobogenized, line 1515: That is not necessarily true. In the code fragment: 

```
    for (int i = 0; i < elementCount; ++i) {
        std::string key = stChildArray[i].GetName(cocoLoader);
        std::string value = stChildArray[i].GetValue(cocoLoader);

        if (key == "classname" )
        {
            if (!value.empty())
            {
                classname = value;
                widget = this->createGUI(classname);
            }
            else
            {
                CCLOG("Warning!!! classname not found!");
            }
        }else if(key == "children"){
            childrenNode = &stChildArray[i];
        }else if(key == "options"){
            optionsNode = &stChildArray[i];
        }
```

There is the distinct possibility of widget not being assigned to because all keys are either equal to "children" or "options." From there, if reader is null, we will of course go to the following code:

```
else{
            CCLOG("Widget or WidgetReader doesn't exists!!!  Please check your csb file.");
        }
```

However, there is no assert statement here, meaning that we can continue on down to the following code fragment:

```
   //parse children
    if (nullptr != childrenNode) {
        rapidjson::Type tType22  = childrenNode->GetType(cocoLoader);
        if (tType22 == rapidjson::kArrayType) {

            int childrenCount = childrenNode->GetChildNum();
            stExpCocoNode* innerChildArray = childrenNode->GetChildArray(cocoLoader);
            for (int i=0; i < childrenCount; ++i) {
                rapidjson::Type tType  = innerChildArray[i].GetType(cocoLoader);

                if (tType == rapidjson::kObjectType) {

                    Widget *child = widgetFromBinary(cocoLoader, &innerChildArray[i]);

                    if (child)
                    {
                        PageView* pageView = dynamic_cast<PageView*>(widget);
                        if (pageView)
                        {
                            pageView->addPage(static_cast<Layout*>(child));
                        }
                        else
                        {
                            ListView* listView = dynamic_cast<ListView*>(widget);
                            if (listView)
                            {
                                listView->pushBackCustomItem(child);
                            }
                            else
                            {
                                if (dynamic_cast<Layout*>(widget))
                                {
                                    if (child->getPositionType() == ui::Widget::PositionType::PERCENT)
                                    {
                                        child->setPositionPercent(Vec2(child->getPositionPercent().x + widget->getAnchorPoint().x,
                                                                       child->getPositionPercent().y + widget->getAnchorPoint().y));
                                    }
                                    child->setPosition(Vec2(child->getPositionX() + widget->getAnchorPointInPoints().x,
                                                            child->getPositionY() + widget->getAnchorPointInPoints().y));
                                }
                                widget->addChild(child);
                            }
                        }
                    }

                }
            }
        }
    }
```

In that case, then the widget will be null. Of course, that means then that I should have put the assert statement in the area I posted out above! Thanks for catching that; will you take care of that?

cocos/editor-support/cocostudio/CCSSceneReader.cpp
hobogenized, line 367: The assertion here is to catch before the strcmp, as the following code fragment contains:

```
    if(strcmp(className, "CCNode") == 0)
```

As far as I know, strcmp on a null value can potentially be dangerous, as can be seen in the following StackOverflow link:
http://stackoverflow.com/questions/4932848/strcmp-behaviour

tests/cpp-tests/Resources/ccs-res
hobogenized, line 1: This was an accident, thanks for catching it.

cocos/ui/UIWidget.cpp
hobogenized, line 561: From the code for create:

```
EventListenerTouchOneByOne* EventListenerTouchOneByOne::create()
{
    auto ret = new (std::nothrow) EventListenerTouchOneByOne();
    if (ret && ret->init())
    {
        ret->autorelease();
    }
    else
    {
        CC_SAFE_DELETE(ret);
    }
    return ret;
}
```

If I understand this correctly, then the method could return a null if we do not have enough memory to create an EventListener, which I believe an assert should be in place for just in case.

hobogenized:
I've added a few comments on some of your code comments; I still need to take a look at the rest of your comments later, but would it be possible to take a look at what I've written in the meantime?

zilongshanren:
Thanks @hobogenized 
Yes, I will definitely look all of your comments.

cocos/network/HttpAsynConnection.m
samuele3hu, line 188: @hobogenized ,I have create a new pr to resolve the memory leak for AudioEngine.
Thanks for contributing and could you help to review these codes.
#12103.

cocos/ui/UIWidget.cpp
zilongshanren, line 561: In modern operating system, we will never come across "run out of memory exception"? 
When the memory is not enough, os will use external storage and paging occurs.

Am I right? If not, please correct me.

And in mobile platform, every apps are running inside of a sandbox, if the apps consumes many memory, the os will kill it. So we don't need to use assert here.
If we do want to check "run out of memory issue", maybe this kind of checking code will scatter everywhere in the code base.

cocos/editor-support/cocostudio/CCSSceneReader.cpp
zilongshanren, line 367: Maybe it's true here. But AFAIK nobody report this issue. I think modern compiler will take care of this issue.

zilongshanren:
Hi, @hobogenized 
The Audio issue and memory leaks in HttpClient has been solved by us.
For the assert part, I suggest you to start another PR which only contains the assertion patch.
We could continue the talking in that PR.

Thanks.

Issue #9633, opened 2014-12-25
[Feature] Add Title Alignment for ui::Button


zilongshanren:
Wish to have changes in UIButton.h and UIButton.cpp for titleAligment.

void setTitleAlignment(TextHAlignment hAlignment); 
void setTitleAlignment(TextHAlignment hAlignment, TextVAlignment vAlignment);
void Button::setTitleAlignment(TextHAlignment hAlignment)
{
_titleRenderer->setAlignment(hAlignment);
}

void Button::setTitleAlignment(TextHAlignment hAlignment, TextVAlignment vAlignment)
{
_titleRenderer->setAlignment(hAlignment, vAlignment);
}

And default title align setup in void Button::initRenderer()

_titleRenderer->setAlignment(TextHAlignment::CENTER, TextVAlignment::CENTER);

Dimon4eg:
I implemented the feature but PR can't be merged automatically.
Could you please merge?

PR #11883, opened 2015-05-18
Fix crash on Windows if pass string more than 16kb to cocos2d::log.

Dimon4eg:
- crash on Windows

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -123,12 +123,12 @@ static bool isFloat( std::string myString ) {
 static ssize_t mydprintf(int sock, const char *format, ...)
 {
     va_list args;
-	char buf[16386];
+    char buf[16386];
 
-	va_start(args, format);
-	vsnprintf(buf, sizeof(buf), format, args);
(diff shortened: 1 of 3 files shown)


zilongshanren:
Thanks, @Dimon4eg 
I found that you put two fixes in the same PR which is not good.
I have opened a issue for adding setTitleAlignment for ui::Button here https://github.com/cocos2d/cocos2d-x/issues/9633
I suggest you to open another PR and link your PR with the issue I listed above.

Thanks.

Dimon4eg:
Sorry, I should create branch for each task.

zilongshanren:
@Dimon4eg You should create another topic branch and isolate your changes only in that branch.

Dimon4eg:
Thanks, but what do now?

zilongshanren:
@Dimon4eg 
Try `git reset --hard HEAD~1` and `git push origin v3 --force`.

Or you could just close this PR and create another two topic branch and cherry-pick the commits in this branch.

The later option is safer, the first option might lose data. Do it carefully...

PR #11892, opened 2015-05-19
V3 frame buffer object

dabingnn:


dabingnn:
@minggo @ricardoquesada @masterhou @super626 
Could you please review this pull request?
It is ready.

ricardoquesada:
How do you add RenderTargets to the FrameBuffer ?

We need more examples as well, like how to grab the texture from the FrameBuffer so that we can generate a shadow map, or post-processing effects.

dabingnn:
FrameBufferObject is the low level infrastructure for shadow map and post processing. 
Taking shadow map for example, what we need to do is rendering the scene two times
- Rendering to shadow map

What we need is to generate a virtual camera at the light point of view, and then rendering the scene by this virtual camera. What is a shadow map, it is Frame Buffer Object. So in this rendering phrase, we do not rendering to the game screen, instead we rendering to the Frame Buffer Object.
The pseudocode would be:

``` cpp
//could has some slight changes if we need a specific texture format.
auto fbo = FrameBufferObject::create(order, with, height, format); 
auto camera = light->generateVirtualCamera(...);
camera->setFrameBufferObject(fbo);
camera->renderSceneWithShadowMapPass(...);
```
- Rendering to game screen

This is the normal rendering, by rendering the scene at the player point of view. we take the shadow map (A texture) from the first rendering phrase, and draw the shadow with the shadow map.

``` cpp
auto texture = fbo->getColorTexture(...);
auto camera = getPlayerCamera();
camera->setFrameBufferObject(gameScene);
//set texture to game objects that receive shadows
for (auto object : _shadowReceives)
{
object->setShadowMapTexture(texture);
}
camera->renderSceneWithShadowShadingPass(...);
```

cocos/2d/CCScene.cpp
```diff
@@ -186,6 +187,7 @@ void Scene::render(Renderer* renderer)
 #endif
 
     Camera::_visitingCamera = nullptr;
+    FrameBufferObject::applyDefaultFBO();
```
super626: Switch frame buffer multiple times.
It is better to cache current frame buffer.

super626:
As you know that frame buffer is time consuming.
Maybe we can add a refresh rate to the FrameBuffer. Then you can decrease the refresh rate for the shadows and mirrors.

super626:
And where is the viewport?

ricardoquesada:
@dabingnn 
Yes, thanks. I think we need to add an example that shows how to do that.

Also, how can I add different render targets to the frame buffer object ?

Please, take a look at the `FrameBuffer` / `RenderTarget` objects from GamePlay in order to understand what I mean by Render Target.
- https://github.com/gameplay3d/GamePlay/blob/master/gameplay/src/FrameBuffer.h
- https://github.com/gameplay3d/GamePlay/blob/master/gameplay/src/RenderTarget.h

dabingnn:
@super626 
- For Performances issue

Yes, you are correct, I think we should test the performance to see if the optimisation is needed.
- For Viewport

I did not add it because I would like to add it later in another pull request to keep the merge smaller to be reviewed.

dabingnn:
@ricardoquesada 
Is there any suggestions of the API and feature requirement for FrameBufferObject?
Generally, the idea of this pull request is the same as frame buffer object, but with slightly difference that we use one class (`FrameBufferObject`) instead of two (`FrameBuffer` and `RenderTarget`). Off course, we do not have some features such as `Multiple Render Targets`, `Render to Texture cube` etc. Do we need to add it now? If so, let us discuss the requirements and refactor it. Currently, this frame buffer object only has a bacis logic, that is, be able to render to texture, and use the rendered texture as the texture source of the next step rendering.
Thanks.

ricardoquesada:
@dabingnn 
I think it would good to add RenderTarget support in FrameBufferObject before merging the pull request.
@super626 and Yingtao are the best persons to ask what are the requirements, but I remember that Yingtao said that he needed multiple Render Target support.

dabingnn:
Thanks for your suggestion. @ricardoquesada 
So, Let us start to discuss the requirement and API based on this pull request on the forum to make it clearer, then we can continue.

ricardoquesada:
@dabingnn sure. Could you contact Yingtao and @super626  and discuss the requirements directly with them? Thanks.

cnsoft:
i have some  common requirement: RenderTexture?  e.g:  mini map.   3D Avatar target ui. (preview the model in ui )

ricardoquesada:
@dabingnn were you able to discuss it with Tony and Yingtao?

If not I think it is better to add RenderTarget class, like in GamePlay3D, and then start adding many examples and tests.
Let me know if you need help with it. Thanks.

dabingnn:
@ricardoquesada What is your suggestions of this pull request? If it works well, I will solve the compile errors on windows and winrt, platform, then we can merge it.

ricardoquesada:
@dabingnn 
questions: will I be able to create Post Processing effects with this API ?
If yes, could you post a sample code ? If not, what are the needed steps to make it support Post Processing effects ?

What is the difference between "render order" and "z order" for the cameras... I missed that part? And what is "fid" ? 

minor things:
I would remove the word `Object` from `FrameBufferObject`... I would just call it `FrameBuffer` since everything in the end is an `Object`. Just personal preference.

Thanks.

dabingnn:
1. I think it is not very easy to create a sample for `post processing` and `Shadow mapping`, because we need a class such as `Shadow mapping` and `Post processing` for their rendering work flow. However, these feature would rely on the `Frame Buffer Object`.
2. `fid`, the order of frame buffer object, which means that if we want to do `shadow mapping`, the shadow map should be generated before the final rendering. so the `fid` of the `FBO` for `Shadow mapping` is smaller than the one for `Final rendering.`
3. Yes, We can use `FrameBuffer` instead of `FrameBufferObject`. But I suggest to make it clear on the comments(which could be added after this pull request). Because it could be a little confusing that `Frame Buffer` is more likely to refer to the final game screen created by the platform such as `IOS` and `Android`. 

Generally speaking, I think there are several works left after this feature.
1. Viewport feature, which allows us to rendering to a portion of the `FBO`.
2. some use case for this, such as `Post processing` and `mirror reflection`.

ricardoquesada:
I don't think we need a `PostProcessing` class in order to create a post processing effect.

What we need to do is to get the texture from the `Camera`, apply a hard-coded effect, and the render it. If we cannot do that, I think we should improve the API.

What I mean, is basically something like this pseudo code:

``` c++
// render one frame 
camera->render();

// get the color buffer from the camera in the form of Texture
auto texture = camera->getFrameBuffer()->getRenderTarget()->getTexture();

// apply hard coded post-processing effect
apply_blur(texture);

// assign texture to full screen sprite
sprite->setTexture(texture);
```

dabingnn:
Sorry, I do not understand your code.
If I am right, these lines demonstrate the scene creation

``` cpp
auto texture = camera->getFrameBuffer()->getRenderTarget()->getTexture();
sprite->setTexture(texture);
```

these two lines are the rendering workflow.

``` cpp
camera->render();
apply_blur(texture);
```

I do not think we can mixing them.

If a blur is needed, we can do it like this.

``` cpp
auto texture = camera->getFrameBuffer()->getRenderTarget()->getTexture();
sprite->setTexture(texture);
auto shader = Shader::create("blur. vsh", "blur.fsh");
sprite->setShader(sprite);
```

But it does not have much relationship with `post effect` and `post processing` system.

`Post processing` effect could be much more complex, 
eg 1.
- . we can blur a texture from FBO
- . we can monochromatization from FBO
  we may need both blur and monochromatization the result from FBO. which means we may need process multiple post effects one by one.

eg 2.
`Depth of field`, the step could be:
- . we need to render the part of scene which is farther than the focus to a texture A, blur it to the screen
- . we need to render the focus area to a screen
- . we need to render the part of scene which is nearer than the focus to a texture B, and blur it to the screen.

ricardoquesada:
Thanks. Let me try again... I think there are some misunderstandings...
I just want to know if the pseudo code that I posted could be implemented. I don't care if we need more code... but if the code that I posted can't be done, I want to know what is missing. 
Let me explain again the pseudo code.

``` c++
// forces the camera to render what it sees.
// it will render everything into a texture in case a FrameBuffer is attached to it.
// if no FrameBuffer is associated with it, it will render everything into the main Color Buffer
camera->setFrameBuffer(fbo);
camera->render();

// now I get the texture from the FrameBuffer
auto texture = camera->getFrameBuffer()->getRenderTarget()->getTexture();

// and now I reuse the texture on a sprite that is already on the scene
// I can in fact use this texture in any other place... for the sake of simplicity, I'm just using it in an existing sprite.
sprite->setTexture( texture );
```

If I cannot do that, then for sure I won't be able to do post-processing effects.
Basically my question is: Can I do something like that with your patch ? If not, what is missing ? Thanks.

dabingnn:
The blocker is that `camera->render();` is called in `Scene::render()` function, but other functions are in `SomeClass::init()`中。

ricardoquesada:
and can we add that function in `camera`? and does it make sense to have it?

dabingnn:
I have no idea about supporting `Post processing` in this way.
If we need `Post processing`, here is the wikipedia page for `post processing`[http://en.wikipedia.org/wiki/Video_post-processing], we need to research it and  have a discussion of implement it in our engine.

ricardoquesada:
come one... help me here.
You added `FrameBuffer`... so, for post-processing effects we need to use `FrameBuffer` as well. correct? Or would you create another new object?
If we are going to use the same `FrameBuffer`, then how are we going to use it?

dabingnn:
Yes, `Post processing` need `FrameBuffer`, but it does not mean that we support `Post processing` immediately after we have `FrameBuffer`. `FrameBuffer` has several use case besides `Post processing`. In my opinion, Cocos need a rendering pipeline, `Post processing` is a section of this pipe line, that it the thing missed in our engine. 
I remember that `Post processing` is not in this milestone, shall we add it, or in the next milestone?

ricardoquesada:
yes, Post Processing will be in the next milestone. I thought it was for this one, but I discussed it yesterday with @walzer and he told me is not for this one.

Taking that into account, the only thing that we should pay attention to, is to make sure that we will be able to extend `FrameBuffer` without breaking compatibility.

dabingnn:
@ricardoquesada @walzer 
Hi, Riq and Walzer, given that the frame buffer may be not stable because there are new features based on it, I have a proposal for it:
*. Place it in `experimental` namespace.
*. when we adding `FrameBuffer` and features base on `FrameBuffer`, we only keep the compatibility that, if user do not use them (it may be used internally, but users do not call any API of them), it is compatible.
*. the `FrameBuffer` API and features base on it is not that stable
*. after we iterate several milestones, the features and API could be statble, then we move it to stable namespace.

Compatibility means more constrains, which will slow down our progress.
Does it make sense?

ricardoquesada:
yes, makes sense. Let's put in the `cocos2d::experimental` namespace

dabingnn:
@ricardoquesada @masterhou  @super626 
Pull request updated. Could you please review it?
####Changes
1. class rename: `FrameBufferObject->FrameBuffer` as you suggested in comments
2. move FrameBuffer related classes to `cocos2d::experimental`
3. add `Viewport`, it is also in namespace `cocos2d::experimental`

####Compatibility
1. The behaviour of our engine keep the same if user do not use these `FrameBuffer` related API in `cocos2d::experimental` .
2. These API may be used internally.
3. These API is not very stable given that we need to add several features based on it.

If you agree, I will add project file for win32/winRT platform and merge it.

ricardoquesada:
Thanks. good for me.

PR #11893, opened 2015-05-19
New audio engine support in winrt win8.1-universal

anniruddh:
-- Added new audio engine support to win8.1-universal.
-- Currently supports only WAV file formats.
-- Support for MP3 and Ogg in progress and will be following soon.


minggo:
@anniruddh 
Why use so many wav files in test case?
Is it needed?

anniruddh:
Yes, since current implementation only supports WAV files. For the new audio engine tests (esp. 'play-simultaneously') to run successfully, effect files in WAV format are necessary. These can be later dropped when MP3 support is implemented. I will drop them in the next PR with MP3 support. For now they are necessary for new audio tests to run properly for WinRT. I have created very small size wav files. (just few kbs)

minggo:
Got it, thanks.

PR #11895, opened 2015-05-19
fix #10767

### LINKED ISSUE #10767: Transition breake when there is a Sprite3D in the scene
you can try it with the TransitionTest.cpp, add to `TestLayer1::TestLayer1`:

```
auto sp = Sprite3D::create();
 addChild(sp);
```

then it will break when transition occurs, callstack:

super626:
fix #10767

Transition breaks when there is a Sprite3D in the scene

Reason is that there is no visiting camera when transition occurs.


super626:
[ci rebuild]

super626:
[ci rebuild]

PR #11897, opened 2015-05-19
Convert websocket received binary message as string for Lua

xpol:
Websocket Lua API should not convert data to `LuaValueArray`.

Currently `void LuaWebSocket::onMessage(WebSocket* ws, const WebSocket::Data& data)` calls `SendBinaryMessageToLua` to convert a binary data (byte array) to `LuaValueArray` (table in lua side). But `LuaValueArray` is not actually a binary data.

Please note that Lua does not distinguish string and binary.

Lua string:
1. are binary
2. contains length information
3. does not contains tailing `\0`
4. It is safe even string data has `\0` in any position of string.

So, my suggestion is just pass string to Lua even received message is binary.

### DIFF
--- a/cocos/scripting/lua-bindings/manual/network/Lua_web_socket.cpp
+++ b/cocos/scripting/lua-bindings/manual/network/Lua_web_socket.cpp
@@ -35,40 +35,6 @@
 
 using namespace cocos2d;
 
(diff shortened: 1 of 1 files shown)


samuele3hu:
@xpol, you're right. But considering the compatibility,it has been not changed.

minggo:
close it since it will break compatibility

PR #11920, opened 2015-05-20
Fix crash on Windows if pass string more than 16kb to cocos2d::log

Dimon4eg:
- crash on Windows


zilongshanren:
Hi, @Dimon4eg 
This PR seems conflict with main repository. Could you resolve it?
Thanks.

Dimon4eg:
How can I resolve?
I've latest v3 branch and no conflicts.

zilongshanren:
@Dimon4eg 
Sorry for the late reply.
I think you should fetch cocos2d-x upstream and merge the v3 branch to your local branch.
It may have conflicts, you need to resolve it.

Dimon4eg:
I've created new PR #12369.

PR #11921, opened 2015-05-20
Uibutton support bmfont

felixjingga:
Support BMFont in UIButton

### DIFF
--- a/cocos/ui/UIButton.cpp
+++ b/cocos/ui/UIButton.cpp
@@ -781,22 +781,27 @@ Color3B Button::getTitleColor() const
 
 void Button::setTitleFontSize(float size)
 {
-    if(nullptr == _titleRenderer)
+    if (nullptr == _titleRenderer)
     {
         this->createTitleRenderer();
     }
+
+    _fontSize = size;
     if (_type == FontType::SYSTEM)
     {
-        _titleRenderer->setSystemFontSize(size);
+        _titleRenderer->setSystemFontSize(_fontSize);
     }
     else if (_type == FontType::TTF)
     {
         TTFConfig config = _titleRenderer->getTTFConfig();
-        config.fontSize = size;
+        config.fontSize = _fontSize;
         _titleRenderer->setTTFConfig(config);
     }
+    else
+    {
+        return;
+    }
     updateContentSize();
-    _fontSize = size;
 }
 
 float Button::getTitleFontSize() const
--- a/cocos/ui/UIButton.cpp
+++ b/cocos/ui/UIButton.cpp
@@ -795,11 +795,6 @@ void Button::setTitleFontSize(float size)
         config.fontSize = size;
         _titleRenderer->setTTFConfig(config);
     }
-    else
-    {
-        // BMFont can't set its font size
-        size = _fontSize;
-    }
     updateContentSize();
     _fontSize = size;
 }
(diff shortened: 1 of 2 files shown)


cocos/ui/UIButton.cpp
```diff
@@ -821,11 +824,19 @@ void Button::setTitleFontName(const std::string& fontName)
     }
     if(FileUtils::getInstance()->isFileExist(fontName))
     {
-        TTFConfig config = _titleRenderer->getTTFConfig();
-        config.fontFilePath = fontName;
-        config.fontSize = _fontSize;
-        _titleRenderer->setTTFConfig(config);
-        _type = FontType::TTF;
+        if (fontName.find(".fnt") != std::string::npos || fontName.find(".FNT") != std::string::npos)
```
Dimon4eg: Will not work in case of .Fnt extension. Convert fontName to lower case before search.

cocos/ui/UIButton.cpp
```diff
@@ -788,11 +788,14 @@ void Button::setTitleFontSize(float size)
     {
         _titleRenderer->setSystemFontSize(size);
     }
-    else
+    else if (_type == FontType::TTF)
     {
         TTFConfig config = _titleRenderer->getTTFConfig();
         config.fontSize = size;
         _titleRenderer->setTTFConfig(config);
+    } else {
```
zilongshanren: The indentation is wrong here.

felixjingga: I fix it in the commit below

cocos/ui/UIButton.cpp
felixjingga: Yes, and I add the code to support this in newer commit.

felixjingga:
hopefully the studio team can support this feature in the cocos studio too.. :)

minggo:
@andyque 
How about this pull request?

cocos/ui/UIButton.cpp
```diff
@@ -788,12 +789,17 @@ void Button::setTitleFontSize(float size)
...
+    else if (_type == FontType::TTF)
     {
         TTFConfig config = _titleRenderer->getTTFConfig();
         config.fontSize = size;
         _titleRenderer->setTTFConfig(config);
     }
+    else
+    {
+        // BMFont can't set its font size
+        size = _fontSize;
```
zilongshanren: why change the argument here? If BMFont can't set it's font size, it should just omit it.

felixjingga: @andyque so what your suggestion this function should do when user invoke the setTitleFontSize and use BMFont? Should do nothing at all? or we can use assert there?

Thanks,

zilongshanren: @felixjingga  I think just do nothing is OK.  You could simply remove the else clause.

felixjingga: is that okay to let the UIButton class to "updateContentSize", and set the member variable "_fontSize" even though nothing change?

zilongshanren: @felixjingga  We shouldn't call `updateContentSize` if the content size is not changed. But we do need to record `_fontSize`, so if we change the Font type to TTF or system font, the new font size would take effect.

felixjingga:
@andyque I already change the code to keep record the _fontSize variable, and call the updateContentSize only when necessary. Thanks

zilongshanren:
@felixjingga Thanks.

zilongshanren:
@minggo  This PR is ready now.

PR #11950, opened 2015-05-21
Console:Fixed crash

WenhaiLin:


cocos/base/CCConsole.cpp
```diff
@@ -1056,6 +1055,11 @@ void Console::addClient()
         _maxfd = std::max(_maxfd,fd);
 
         sendPrompt(fd);
+
+#if CC_TARGET_PLATFORM == CC_PLATFORM_IOS
+        int set = 1;
+        setsockopt(fd, SOL_SOCKET, SO_NOSIGPIPE, (void*)&set, sizeof(int));
+#endif
```
minggo, line 1062: Please add comment about these codes.
Why need to do special operation on iOS?

PR #11951, opened 2015-05-21
update spine runtime to commit 27270a5781

jtianling:
add spine blend mode support
add spine blend mode test


jtianling:
In the lastest spine tools, add the additive, multiply, screen blend mode, but the old runtime did not support it. Update to lastest commit and fix it.

minggo:
Thanks @jtianling .
I have two questions:
- Is it a stable release version?
- Did you modify any source code of Spine runtime?

jtianling:
1. I didn't know which version is stable release version of spine runtime, because as a personal project(but earn much money), It is not so strict project managed. As I know, in the official website about runtime, it also only have a link to github and had not a download link to stable release version.(http://zh.esotericsoftware.com/spine-runtimes) 
2. I didn't not modify any source code of Spine runtime and it looks like violating the license of spine runtime(heard from one of cocos2d-x team), but only modified source code of cocos2d-x in the part of blend mode. It pass the original spine test and the test of blend mode I added.

jtianling:
I have watched the spine-runtime code again, and found the files(SkeletonRenderer.h .cpp SkeletonAnimation.h .cpp) also a part of spine-runtime -_-! I though these files were a part of cocos2d-x and only the C  language part runtime is from spine. Maybe cocos2d-x just need update all the files from spine runtime and should be ok, because I found the lastest spine runtime had fix the problem a month ago.

Issue #10759, opened 2015-03-06
Clipping node doesn't work on Android 5.0:


piotrros:
See this topic:

http://stackoverflow.com/questions/28842039/cocos2d-x-android-5-0-clipping-node-not-working

hsienwei:
i got it too.
It occurred to Nexus 7 with android 5.1.
Some info below.
https://dl.dropboxusercontent.com/u/10581994/cocos2dx_issue/nexus.txt
https://dl.dropboxusercontent.com/u/10581994/cocos2dx_issue/Screenshot_2015-04-16-13-43-24.png
https://dl.dropboxusercontent.com/u/10581994/cocos2dx_issue/Screenshot_2015-04-16-13-41-14.png

if modify Cocos2dxActivity.java like code below, problem  will be fix.
// Switch to supported OpenGL (ARGB888) mode on emulator
if (isAndroidEmulator())
     this.mGLSurfaceView.setEGLConfigChooser(8, 8, 8, 8, 16, 0);
else 
     this.mGLSurfaceView.setEGLConfigChooser(8, 8, 8, 8, 16, 8);

p.s.  using cocos2d-x 3.5

hussain-sohaib:
to fix this issue, in appcontroller.mm change pixelformat and depthFormat to this.

pixelFormat: kEAGLColorFormatRGBA8
depthFormat:  GL_DEPTH24_STENCIL8_OES

thanks!

tomhmp:
I'm also experiencing this issue although neither of the above suggestions worked for me.

My clipping nodes seem to work fine on our Samsung S3 (4.1.2) but not on our S4 (4.2.2) or our Nexus 7 (4.4.3). 

I've also tried changing the #define GL_DEPTH24_STENCIL8 in CCGL-android.h to GL_DEPTH24_STENCIL8_OES but that didn't seem to make a difference. No luck with creating the ClippingNodes in a CallFunc either (which I read as a suggestion elsewhere).

I'm currently using cocos2dx v3.4 if anyone else has any thoughts?

lalaloha:
Hi,

I got the problem described in "Screenshot_2015-04-16-13-41-14.png" too with cocos2d-x 3.5, and using UIScrollView, and it happens in nexus 7.. I tested all the code in many devices, and all is okay.. but in nexus 7 (the first model), it became like the screenshot.. Is there any quick and permanent fix for this? thanks

yangws:
[...]

PR #11952, opened 2015-05-21
Enhance egl select config

felixjingga:
I notice that when we are using "glContextAttrs = {8, 8, 8, 8, 24, 8}" in AppDelegate, and when using configchooser to choose the config, it will search whether it has the config "8,8,8,8,24,8", if it's not exists then it will select the first config available. But in ASUS Nexus 7, one of the many configs available is "8,8,8,8,16,8", but instead the old codes chooses "8,8,8,8,0,0", and it makes UIScrollView implementation in Nexus 7 become all green and white..

This changes need to be addressed further by testing it onto many devices.


felixjingga:
@walzer @ricardoquesada  I think this PR is related to #10759

minggo:
@dabingnn 
Could you please take a look of it?

felixjingga:
Hi, how about this pull request @minggo @dabingnn ? is it okay? Thanks!

dabingnn:
Let us define the behaviour of the code `EGLConfig selectConfig` function, suppose we want a `glConfigAttribs[6] = {8,8,8,8,24,8}`, but there are only two available configs `{8,8,8,8,0,0}` and `{8,8,8,8,16,8}`. If I understood right, it would be better to select the second one, but I do not think the new function will return that. So, what is the proposal of this function? Thanks.

felixjingga:
First, I'll try to explain what happened in Nex7 device,

when cocos2dx use the Cocos2dxEGLConfigChooser class, the method "chooseConfig(EGL10 egl, EGLDisplay display)" will be called. Following the logic inside, it will enter the if because "egl.eglGetConfigs(display, null, 0, numConfigs)" return true.

inside the if, when invoking "egl.eglChooseConfig(display, EGLattribs, configs, numConfigs[0], choosedConfigNum)", the "choosedConfigNum[0]>0" will return false, because it can't find a config that matches or better than "8,8,8,8,24,8".. so it will enter the else condition

inside there, it define two attribs, the "defaultEGLattribs" (without alpha) and "defaultEGLattribsAlpha" (with alpha).. and it again checks whether the user want alpha/not by testing "this.configAttribs[3] == 0", and because we want alpha, it will enter the else condition..

inside the else, it run "egl.eglChooseConfig(display, defaultEGLattribsAlpha, configs, numConfigs[0], choosedConfigNum)", and it find config matches or better than "4,4,4,4,0,0"..

and run the "selectConfig(egl, display, configs, attribs);" with attribs is "4,4,4,4,0,0".. and inside the selectConfig, this function should really return the best config among the available config.. in the case of this, the attribs is "4,4,4,4,0,0", the "egl.eglChooseConfig" return {8,8,8,8,0,0}, {8,8,8,8,16,8}, etc (the point is the function return more than one avail config decent to choose from)..

the old code, only return the first "decent" config (in this case, {8,8,8,8,0,0}).. and never consider better config to choose, in this case "{8,8,8,8,16,8}"..

So, my take is the "selectConfig" proposal will be like this.. given current available config to choose from (EGLConfig[] configs that are returned by eglChooseConfig), and given our preference of config (int[] attribs).. the function has to choose the most best config available from all available config ( {8,8,8,8,0,0}, ... {8,8,8,8,16,8}, etc..) to choose from that matches our preferred config, or better than our preferred config (in this case {4,4,4,4,0,0})

What do you think?

When we say that we want "{8,8,8,8,24,8}", the code cocos2d-x use is already right that it has to query the device whether the device support the config wanted, and if it is, cocos2dx will use the "selectConfig" function. But, selectConfig function never take into account that the device will return more than one config that has a match with the "8,8,8,8,24,8" and better than that.. there should be no instances in the function that the attribs[] variable will be "8,8,8,8,24,8" and the only available configs are "{8,8,8,8,0,0}" and "{8,8,8,8,16,8}".. CMIIW

ricardoquesada:
I'm not familiar with this part of the code, so I at the moment I don't have the needed information to review this patch.

But what I can say is that we need a unified way to create GL surfaces in cocos2d-x.
iOS is using its own code, Android is using its own code, Windows Phone too.
We need something like GLFW3 for mobile.
In fact, the best thing would be to have Android and iOS support on GLFW3.

felixjingga:
@ricardoquesada I agree with you.. This PR is more like a "quick fix" to the existing solution cocos2d-x offer.

@dabingnn What do you think about this PR?

when cocos2d-x query the device about available configuration, it passes our preferred configuration and the device will then return lists of configuration that are either matches or better than the preferred configuration. Say, that the device cannot find such configurations, cocos2d-x will pass the necessary minimum condition (either {4,4,4,4,0,0} or {5,6,5,0,0,0} depends whether user want to have alpha/not) and again query the device to this minimum preferred configuration (which pre-set by cocos2d-x).. with this "minimum condition", the device can give us more than one configuration that we can choose. With the modified code, it'll chooses the best configuration from all the configuration the device suggested.

zilongshanren:
@dabingnn  Any updates?

dabingnn:
Thanks, it will solves the problems as you suggested.
If we specify `glContextAttrs = {4, 4, 4, 4, 0, 0}`, the final result is that it will also choose the `{8,8,8,8,16,8}` for the final rendering, am I right? 
This result may be not the one we want because `{4,4,4,4,0,0}` is enough.

felixjingga:
@dabingnn Yes! Is there any use case that someone might prefer "{4,4,4,4,0,0}" without the depth and stencil? and if using {4,4,4,4,0,0} is enough for someone to use, what is the advantage to choose "{4,4,4,4,0,0}" instead of "{8,8,8,8,24,8}" if the "{8,8,8,8,24,8}" is available too for the device?

dabingnn:
If I remembered correctly, I heard riq was saying that `{5,6,5,0,0,0}` is faster than `{8,8,8,8,0,0}`. Is there 
@ricardoquesada Is there any more info about this?

The purpose of providing `glContextAttrs` is to let the developer to select a prefered framebuffer bits. If we return the best quality that is supported on current platform, we do not need `glContextAttrs` at all.
My thoughts of `glContextAttrs` would be like this:
1. if the device has the exact format, just select it.
2. select the best matched one, which is not done in current version

felixjingga:
@dabingnn can you elaborate more on "the best matched one"? what is the criteria in your mind about this? we need more info about this, regardless this code might suit for my purpose, but we need to discuss this further to suit other use case / purpose too..

ricardoquesada:
@dabingnn 
One consumes more memory (I guess is slower) than the other. RGBA8888 uses 32-bit and the RGB565 uses 16-bit.
However, the user must be able to configure that parameter since RGBA8888 has more definition.

dabingnn:
@felixjingga 
I have a general idea about this. 
Let me demonstrate it by an example. if we have these supported attributes on my device.
A: {4,4,4,4,0,0}
B: {4,4,4,4,16,0}
C: {4,4,4,4,16,8}
D: {5,6,5,0,0,0}
E: {5,6,5,0,16,0}
F: {5,6,5,0,16,8}
G: {8,8,8,8,0,0}
H: {8,8,8,8,16,0}
I: {8,8,8,8,16,8}

TestCase1: if we prefer a {4,4,4,4,8,0}, the B is selected.
TestCase2: if we prefer a {4,4,4,4,24,0}, the B is selected.
TestCase3: if we prefer a {8,8,8,0,0,0}, the G is selected.

I know these example would be not appropriate, but I would think that there are several conditions for best match
1. For a preferred attributes, if we have the exact one, select it.
2. For a preferred attributes, we do not have it, but we have several attributes whose quality is better than it, select the worst quality one., like `TestCase1` and `TestCase3`
3. For a preferred attributes, we do not have it, and some attribute bits have exceeded the capacity of the device, use the capacity of this device, like `TestCase2`

dabingnn:
@felixjingga Is there any updates for this pull request? Thanks.

felixjingga:
@dabingnn I haven't had time to refine this PR. Sorry. But I understand what you meant, and will try to refine it like you suggested.

dabingnn:
@felixjingga This pull request could be closed given that it is fixed in #12762 .

felixjingga:
let's close this.. thx for enhanced code by @yangws

PR #11972, opened 2015-05-22
Some small performance change

ggggamer:
Reduce some construct copy.
Avoid unnessesary OpenGL draw call in CCParticleSystemQuad.
Quick compare observer.


minggo:
[ci rebuild]

PR #11983, opened 2015-05-22
use trianglescommand for sprite

dabingnn:


cocos/2d/CCSprite.cpp
```diff
@@ -614,6 +614,17 @@ void Sprite::draw(Renderer *renderer, const Mat4 &transform, uint32_t flags)
     }
 }
 
+TrianglesCommand::Triangles Sprite::getRenderedTriangles() const
+{
+    static unsigned short indices[6] = {0, 1, 2, 3, 2, 1};
```
milos1290, line 619: I've noticed that using a static in a method doesn't work on Android. 
For example:

``` cpp
for (int i = 0; i < 10; i++)
{
    static int a = 0;
    a++;
}
```

Result will always be 1.

I might be wrong, but i think i had problem with that.

dabingnn, line 619: Thanks, could you please tell us which type of devices have this problem?

milos1290, line 619: If i remember well, i tried it on Nexus 5/7 using ndk r9d.

dabingnn, line 619: Hi,  I tested the this branch in several android device, including Nexus 5, it works well.
And for you proposed code, I tested it again, and it also works.

milos1290, line 619: Ok, great news then :)

dabingnn, line 619: Anyway!
Thanks for your information, there could be some bugs like this on android devices.

PR #11980, opened 2015-05-22
Some small performance change 2

ggggamer:

PR #11986, opened 2015-05-22
Some small performance change 2

ggggamer:
ReOpen pullrequest #11980 

Reduce some construct copy.
Optimize kmGLPushMatrix, avoid duplicate memcpy.

### DIFF
--- a/cocos2dx/actions/CCActionEase.h
+++ b/cocos2dx/actions/CCActionEase.h
@@ -407,7 +407,7 @@ public:
 class CC_DLL CCEaseBounce : public CCActionEase
 {
 public:
-    float bounceTime(float time);
+    static float bounceTime(float time);
     /**
      *  @js NA
      *  @lua NA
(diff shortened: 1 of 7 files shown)


cocos2dx/actions/CCActionEase.h
```diff
@@ -407,7 +407,7 @@ class CC_DLL CCEaseElasticInOut : public CCEaseElastic
 class CC_DLL CCEaseBounce : public CCActionEase
 {
 public:
-    float bounceTime(float time);
+    static float bounceTime(float time);
```
andyque: When change a member function to static function. The original object can't access the member function any more. So I think it will break back compatability.

cocos2dx/kazmath/include/kazmath/GL/mat4stack.h
```diff
@@ -44,6 +44,8 @@ void km_mat4_stack_push(km_mat4_stack* stack, const kmMat4* item);
 void km_mat4_stack_pop(km_mat4_stack* stack, kmMat4* pOut);
 void km_mat4_stack_release(km_mat4_stack* stack);
 
+void km_mat4_stack_duplicate_top(km_mat4_stack* stack);
```
zilongshanren, line 47: kazmath is a third party library, it's bad to change it directly. Maybe you should talk to the original author about your changes and then update kazmath in cocos2dx.
Thanks.

ggggamer, line 47: I found Kazade/kazmath but ours is a modified version, not all the same.

ggggamer:
CCEaseBounce is sub class of CCAction, so I think no one will use it via -> .

zilongshanren:
@ggggamer  Currently out merge strategy on V2 is "only accept critical bug fix" and leave the freedom to the community. And you even modify some 3rd party libs which is not acceptable.
Thanks.

PR #11995, opened 2015-05-24
Permitted setting flags to action

MAUSMAUSgames:
I would love if I could use Action::_tag as a flag field.
It's a very small change to the code but a very powerful tool...

Use case:

```
    enum Flags {
        FRAME = 1,
        COLOR = 2,
        OPACITY = 4,
        POSITION = 8,
        SCALE = 16,
        ROTATION = 32,
        SKEW = 64
    };

    auto action1 = RepeatForever::create(Sequence::create(FadeTo::create( 0.1f, 155), 
                                                          TintTo::create(0.1f, 255,0,0),
                                                          nullptr));
    action1->setTag(Flags::OPACITY|Flags::COLOR);
    sprite->runAction(action1);

    auto action2 = RepeatForever::create(Sequence::create(Animate::create(idleAnimation),
                                                          MoveBy::create(0.3f, Vec2(5, 0)),
                                                          nullptr));
    action2->setTag(Flags::FRAME|Flags::POSITION);
    sprite->runAction(action2);

    // stop all actions that change position or opacity
    sprite->stopAllActionsByTag(Flags::OPACITY|Flags::POSITION, true); 
```

### DIFF
--- a/cocos/2d/CCActionManager.cpp
+++ b/cocos/2d/CCActionManager.cpp
@@ -287,7 +287,7 @@ void ActionManager::removeActionByTag(int tag, Node *target)
     }
 }
 
-void ActionManager::removeAllActionsByTag(int tag, Node *target, bool bitwiseFilter)
+void ActionManager::removeAllActionsByTag(int tag, Node *target)
 {
     CCASSERT(tag != Action::INVALID_TAG, "");
     CCASSERT(target != nullptr, "");
@@ -302,7 +302,38 @@ void ActionManager::removeAllActionsByTag(int tag, Node *target, bool bitwiseFil
         {
             Action *action = (Action*)element->actions->arr[i];
             
-            if ((action->getTag() == (int)tag || (bitwiseFilter && action->getTag() != Action::INVALID_TAG && (action->getTag() & tag) != 0)) && action->getOriginalTarget() == target)
+            if (action->getTag() == (int)tag && action->getOriginalTarget() == target)
+            {
+                removeActionAtIndex(i, element);
+                --limit;
+            }
+            else
+            {
+                ++i;
+            }
+        }
+    }
+}
+
+void ActionManager::removeActionsByFlags(unsigned int flags, Node *target)
+{
+    if (flags == 0)
+    {
+        return;
+    }
+    CCASSERT(target != nullptr, "");
+
+    tHashElement *element = nullptr;
+    HASH_FIND_PTR(_targets, &target, element);
+
+    if (element)
+    {
+        auto limit = element->actions->num;
+        for (int i = 0; i < limit;)
+        {
+            Action *action = (Action*)element->actions->arr[i];
+
+            if ((action->getFlags() & flags) != 0 && action->getOriginalTarget() == target)
             {
                 removeActionAtIndex(i, element);
                 --limit;
--- a/cocos/2d/CCActionManager.cpp
+++ b/cocos/2d/CCActionManager.cpp
@@ -287,7 +287,7 @@ void ActionManager::removeActionByTag(int tag, Node *target)
     }
 }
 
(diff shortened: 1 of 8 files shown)


minggo:
What will happen if passing an unknown flag, such as 

sprite->stopAllActionsByTag(20, true);

MAUSMAUSgames:
Well, flags are single bits. So 20 would be 16 | 4. 
In my use case this would mean all Actions with Flags::OPACITY and Flags::SCALE would be stopped.

MAUSMAUSgames:
Just to be clear, it's up to the user how to use the flags. 

All I want is to add an option to `stopAllActionsByTag(int tag)` to stop all actions where `(Actions::_tag & tag) != 0` (bitwise AND)

zilongshanren:
@MAUSMAUSgames 

Sounds nice to me.

@ricardoquesada  Thoughts?

ricardoquesada:
the patch seems to be clean in the sense that by modifying only a few lines of code it as a new behavior to the code without breaking backwards compatibility.

But I prefer to add a new API instead of overriding `stopAllActionsByTag()`..

Instead I would add another one, like: `stopAllActionsByBitwiseTag()` or something like that to avoid confusion.

We also need some test cases in the patch.

zilongshanren:
Thanks, @ricardoquesada 
It makes sense and the suggested API is more cleaner.

Hi @MAUSMAUSgames  
Could you modify your PR according to @ricardoquesada  's comments?
Thanks.

MAUSMAUSgames:
I introduced another member variable to Action: `unsigned int _flags` 
It defaults to to `0` (as opposed to `_tag` which is `int` and defaults to `-1 (Action::INVALID_TAG)`) which is more appropriate for a flag field

This is less invasive. Since I found that some classes set and use tags to actions internally  (`MenuItemLabel` and `ControlButton` use `kZoomActionTag`, and cocostudio seems to use tags on actions see `InnerActionFrame::onEnter()` in `CCFrame.cpp`).

I used the following APIs: 

```
void Action::setFlags(unsigned int flags);
unsigned int Action::getFlags();

void Node::stopActionsByFlags(unsigned int flags);
void ActionManager::removeActionsByFlags(unsigned int flags, Node *target);
```

Added a test case to `ActionManagerTest.cpp` but since I'm quite new to this, I'm not sure I did it the way it's supposed to be done.

milos1290:
Great addition, exactly what i needed :)

cocos/2d/CCActionManager.cpp
```diff
@@ -315,6 +315,37 @@ void ActionManager::removeAllActionsByTag(int tag, Node *target)
     }
 }
 
+void ActionManager::removeActionsByFlags(unsigned int flags, Node *target)
+{
+    if (flags == 0)
+    {
+        return;
+    }
+    CCASSERT(target != nullptr, "");
```
zilongshanren, line 324: We should say something in the CCASSERT macro, like "Target can't be nullptr".

ricardoquesada, line 324: yeah.

MAUSMAUSgames, line 324: I totally agree. 

Actually this is c&p-code from `ActionManager::removeAllActionsByTag()`. 
As a matter of fact, there are a several "empty" asserts of this kind in CCActionManager.cpp (same in CCActionInterval.cpp, CCActionEase.cpp, CCScheduler.cpp...)

Maybe someone should have a look at all of them and change them in one commit...

zilongshanren, line 324: @MAUSMAUSgames  Thanks. I will change all the empty asserts.

zilongshanren, line 324: @MAUSMAUSgames  Thanks. I will change all the empty asserts.

Issue #11581, opened 2015-04-24
Add JSB projects to jenkins-ci


walzer:
In v3.7, we merged JSB and is using separated project files on all platforms.
We need to add these JSB projects into jenkins-ci, and track the status in readme.md

minggo:
Why move to v3.8?

PR #12004, opened 2015-05-25
Add JSB to CI

zilongshanren:


zilongshanren:
This PR fixed issue https://github.com/cocos2d/cocos2d-x/issues/11581

PR #12015, opened 2015-05-25
update 3rd part config

hugohuang1111:


zilongshanren:
@hugohuang1111 you should upgrade 3rd party version number?

PR #12019, opened 2015-05-26
getIntersectionPoint failed when there is no intersection on the terrain

super626:


cocos/3d/CCTerrain.cpp
```diff
@@ -327,6 +327,10 @@ float Terrain::getHeight(float x, float z, Vec3 * normal)
     if(image_x>=_imageWidth-1 || image_y >=_imageHeight-1 || image_x<0 || image_y<0)
     {
         return 0;
+        if (normal)
```
zilongshanren, line 329: This line will never be executed... You put a clause after `return 0`

cocos/3d/CCTerrain.h
```diff
@@ -343,6 +343,8 @@ class CC_DLL Terrain :public Node
      */
     Vec3 getIntersectionPoint(const Ray & ray);
 
+    bool getIntersectionPoint(const Ray & ray, Vec3 & intersectionPoint);
```
zilongshanren: please add `const` at the end of `getIntersectonPoint`

cocos/3d/CCTerrain.h
```diff
@@ -341,7 +341,9 @@ class CC_DLL Terrain :public Node
      * Ray-Terrain intersection.
      * @return the intersection point
      */
-    Vec3 getIntersectionPoint(const Ray & ray);
+    Vec3 getIntersectionPoint(const Ray & ray) const;
+
+    bool getIntersectionPoint(const Ray & ray, Vec3 & intersectionPoint) const;
```
zilongshanren, line 352: @super626  please add comment to this method

zilongshanren:
@super626  This PR does two thing. 
1. Fix getIntersectionPoint failed when there is no intersection on the terrain
2. add two APIs for Physic3DWorld?

I think it's better to do one thing in a PR.  You don't need to change this PR. 
But please pay attention to the future's PR.
Thanks.

super626:
@andyque 
Ok. Thanks

PR #12025, opened 2015-05-26
Fix bug: skybox can't move to other position except origin point in w…

yangws:
…orld space.

### DIFF
--- a/cocos/2d/CCCamera.h
+++ b/cocos/2d/CCCamera.h
@@ -172,7 +172,14 @@ public:
      */
     int getDepth() const { return _depth; }
     
+    /**
+     * Get the frustum's far plane.
+     */
     float getFarPlane() const { return _farPlane; }
+
+    /**
+     * Get the frustum's near plane.
+     */
     float getNearPlane() const { return _nearPlane; }
     
     //override
--- a/cocos/2d/CCCamera.h
+++ b/cocos/2d/CCCamera.h
@@ -172,6 +172,9 @@ public:
      */
     int getDepth() const { return _depth; }
     
+    float getFarPlane() const { return _farPlane; }
+    float getNearPlane() const { return _nearPlane; }
+    
     //override
     virtual void onEnter() override;
     virtual void onExit() override;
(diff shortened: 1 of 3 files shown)


cocos/3d/CCSkybox.cpp
```diff
@@ -152,18 +152,32 @@ void Skybox::draw(Renderer* renderer, const Mat4& transform, uint32_t flags)
...
+     To render the skybox, we should keep camera at the center of skybox.
+     So we need a translate matrix, which can transform origin point to camera pos.
+     Camera's node to word transform matrix don't match our requement,
+        because it maybe contain the scale, rotate, reflact... effects, which isn't need.
+     First, we transform origin point to camera position by camera's node to world matrix.
+     Second, we create a translate matrix with the origin point's world position.
+     */
+    Vec3 cameraPosInNode(0, 0, 0);
+    Vec3 cameraPosInWorld;
+    Mat4 cameraModelMat = camera->getNodeToWorldTransform();
```
andyque: ` Mat4 cameraModelMat = camera->getNodeToWorldTransform();` try to use `auto&` to avoid extra copy?

cocos/2d/CCCamera.h
```diff
@@ -172,6 +172,9 @@ class CC_DLL Camera :public Node
      */
     int getDepth() const { return _depth; }
     
+    float getFarPlane() const { return _farPlane; }
```
zilongshanren, line 178: Add any code comments? I know this function is clear enough, but we should give every public API a comment. 
Use comment to talk about the indent not the implementation.
Thanks.

cocos/3d/CCSkybox.cpp
```diff
@@ -152,18 +152,32 @@ void Skybox::draw(Renderer* renderer, const Mat4& transform, uint32_t flags)
 
 void Skybox::onDraw(const Mat4& transform, uint32_t flags)
 {
-    Mat4 trans(transform);
-    const cocos2d::Vec3 pos(Camera::getVisitingCamera()->getPosition3D());
-    trans.m[12] = pos.x;
-    trans.m[13] = pos.y;
-    trans.m[14] = pos.z;
-
+    auto camera = Camera::getVisitingCamera();
```
andyque: And here use `auto&` instead.

PR #12034, opened 2015-05-26
added CCGridView to extension/Android.mk

ryule:
CCGridView was missing from extension/Android.mk.


zilongshanren:
@ryule  
There is no CCGridView in our repository, right?

ryule:
@andyque

Yes, that's right, I'm sorry.
There are missing commits before this pr.
I'll clean up and rewrite pr next time.

PR #12036, opened 2015-05-26
fixed border width in cocos2d::DrawNode::drawPolygon

musikov:
Fixed a tiny bug when cocos2d::DrawNode::drawPolygon drew a polygon with border width two times larger than you told it;


minggo:
Why the width is two times?

minggo:
Close it since it is not response.

PR #12071, opened 2015-05-28
Fix a bug in Scheduler which may case Force Close

waiter:
Fix a bug in Scheduler which may case Force Close

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -347,9 +347,9 @@ void Scheduler::unschedule(const std::string &key, void *target)
     {
         for (int i = 0; i < element->timers->num; ++i)
         {
-            TimerTargetCallback *timer = static_cast<TimerTargetCallback*>(element->timers->arr[i]);
+            TimerTargetCallback *timer = dynamic_cast<TimerTargetCallback*>(element->timers->arr[i]);
 
-            if (key == timer->getKey())
+            if (timer && key == timer->getKey())
             {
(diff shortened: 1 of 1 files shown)


pandamicro:
[ci rebuild]

dabingnn:
@waiter Thanks.
In my opinion, the new code seems to be safer, but `dynamic_cast` is slower than `static_cast`. Would you give a small test case to reproduce the bug?

waiter:
@dabingnn 
There are two type of selectors: `TimerTargetCallBack` and `TimerTargetSelector`
SO, If one Node scheduled both of the two type selectors, then , call the function `isScheduled(xx,xx)`
Because all selectors will `static_cast` to `TimerTargetCallBack` or `TimerTargetSelector`, and one of the selectors can't cast successful.
And when cast failed, `static_cast` will cast **FORCE CLOSE**! But `dynamic_cast` will just return null!

WenhaiLin:
@dabingnn I think this PR is OK for merge.`Scheduler::unschedule` is not the key point of the overall performance.

dabingnn:
@waiter Thanks, it seems that `TimerTargetCallBack` and `TimerTargetSelector` are subclass of `Timer`, but we need the concrete information on the subclass. My thoughts is that we can refine it.
Although this pull request can be merged, but we may need a better design for it.
@WenhaiLin What is your thoughts?

waiter:
@dabingnn 
You are right, just fix it for now.

PR #12072, opened 2015-05-28
Add enhance service  API

iris-wy:
Add enhance service  API


zilongshanren:
hi @iris-wy 
You don't need to modify `templates/cocos2dx_files.json` file, our robot will do the housework for us.
Like this one: 
https://github.com/cocos2d/cocos2d-x/pull/12064

Issue #12078, opened 2015-05-28
Regroup the modules of cocos2d-x


linshun:
Regroup the modules of cocos2d-x

PR #12080, opened 2015-05-28
issue #12078 added audio, network, physics, platform and storage group.

linshun:
issue #12078 added audio, network, physics, platform and storage group.

### DIFF
--- a/cocos/audio/include/AudioEngine.h
+++ b/cocos/audio/include/AudioEngine.h
@@ -41,7 +41,7 @@
 #endif // ERROR
 
 /**
- * @addtogroup core
+ * @addtogroup audio
  * @{
  */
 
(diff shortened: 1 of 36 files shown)


minggo:
Following is my feedback:
- `audio`, `network`, `renderer` and `storage` looks good to me
- About `platform`: though these files are under `platform` folder, but i think it is not good to name the module to `platform`. The implementation of these class are platform dependent, but it will be confused if the module is named `platform`.
- Now there are 2d physics and 3d physics, so i think it is better to distinguish them. How about `physics2d` and `physics3d`?

linshun:
- As to `platform`, how about name it as `support`?
- As to `physics`, it could be:

`physics`
         - `physics 2d`
         - `physics 3d`
What do you think?

minggo:
Though i think `support` is not a good name, but i can't find a better one.
`physics 2d` and `physics 3d` looks good.

pandamicro:
Maybe we don't need a `platform` module, they can remain in `core`, what do you think?

PR #12106, opened 2015-05-29
AudioEngine:Fixed audio can not resume if it is interrupted cause by an incoming phone call.

WenhaiLin:


darkdukey:
[ci rebuild]

WenhaiLin:
[ci rebuild]

PR #12139, opened 2015-06-02
webview and videoplayer test case bug fix

jianglong0156:
webview and videoplayer test case bug fix, change the mp4 files.


jianglong0156:
[ci rebuild]

Issue #12148, opened 2015-06-02
cocos2d-js: ccui.WebView breaks Win32 and Windows 8.1 Universal Apps


stammen:
The recent addition of ccui.WebView to cocos2d-js reaks Win32 and Windows 8.1 Universal Apps. This feature should only be enabled for iOS and Android. ccui.VideoPlayer will also break the Win32 and Windows 8.1 Universal Apps versions.

How to reproduce:

Build cocos2d-js-win32.sln
run js-tests

Here is the error in the output window

JS: D:/GitHub/cocos2d-x-v3 - Copy/build/Debug.win32/js-tests/script/ccui/jsb_cocos2d_ui.js:384:TypeError: ccui.WebView is undefined

(evaluatedOK == JS_FALSE)
(evaluatedOK == JS_FALSE)

stammen:
Fixed in PR https://github.com/cocos2d/cocos2d-x/pull/12149

stammen:
Fixed by PR https://github.com/cocos2d/cocos2d-x/pull/12139

PR #12149, opened 2015-06-02
Issue: 12148 ccui.WebView and ccui.VideoPlayer only on iOS and Android

stammen:
Fix for issue https://github.com/cocos2d/cocos2d-x/issues/12148

ccui.WebView and ccui.VideoPlayer are only available on iOS and Android. These will be null on the other platforms and will throw an NULL exception.


zilongshanren:
@pandamicro  please review this PR.

pandamicro:
Hi, Dale, Thank you for noticing, though we have already a fix for it, 

https://github.com/cocos2d/cocos2d-x/pull/12139

As the game may run under ios/android without WebView or VideoPlayer, it's more proper to apply another PR, but thank you very much for the fix

stammen:
@pandamicro  Ok. Thanks for the update. Please complete and pull the other request soon so Win32 and Windows 8.1 Universal Apps will work. They currently do not run.

PR #12160, opened 2015-06-03
Blink::stop() access violation

namezero111111:
Blink::stop() is called twice when inside a Repeat action. This results in an access violation

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -1814,7 +1814,8 @@ bool Blink::initWithDuration(float duration, int blinks)
 
 void Blink::stop()
 {
-    _target->setVisible(_originalState);
+    if(NULL != _target)
+        _target->setVisible(_originalState);
     ActionInterval::stop();
 }
 


pandamicro:
@WenhaiLin Please review this PR

WenhaiLin:
@namezero111111 I think "Blink::stop() is called twice" is incorrect,we should fix it in `Repeat::update(float dt)`.

``` cpp
if (_total == _times)
{
    _innerAction->update(1);
    _innerAction->stop();  // I think we should delete this line.
}
```

namezero111111:
Possibly, I'm really new to cocos2d-x.
Looking at `void Animate::stop()`, however, it also checks for _target being valid:
`if (_animation->getRestoreOriginalFrame() && _target)`

So maybe this was also committed due to the stop() action being called incorrectly in `Repeat::update(float dt)`.

WenhaiLin:
`Animate::stop` have this same situation.We should fix it in `Repeat::update(float dt)`.

namezero111111:
Ok, thanks for clearing that up! Either way, invoking a public function (or in the wrong order, i.e. before `startWithTarget()` should never result in a segfault. I think the `NULL` checks should stay in there nonetheless; the penalty is negligible imho.

WenhaiLin:
@namezero111111 I agree with you. @andyque It's OK for merge.

cocos/2d/CCActionInterval.cpp
```diff
@@ -1814,7 +1814,8 @@ bool Blink::initWithDuration(float duration, int blinks)
 
 void Blink::stop()
 {
-    _target->setVisible(_originalState);
+    if(NULL != _target)
```
pandamicro, line 1817: Shouldn't we use nullptr instead of NULL ?

namezero111111, line 1817: Yes, that would would in line with the coding style for cocos2d-x

PR #12168, opened 2015-06-03
V3: Fixes for Windows 8.1 Universal App JavaScript

stammen:
This pull request updates the application framework to keep the cocos2d-x rendering thread alive when the app is not visible. The rendering thread now waits for a std::condition_variable when it is no longer visible. SpiderMonkey MUST remain on the same thread for the life of the app as it is a single threaded implemenation. This pull request now uses the same threading model for both the cpp and js versions of the Windows 8.1 Universal App.

cpp and js templates for the Windows 8.1 Universal App.have also been updated with these changes.

Support for Windows Phone Back Button is now part of cocos2d-js.


pandamicro:
Thank you very much, Dale.
Can you also apply the changes in js-template-runtime ?

stammen:
@pandamicro What is js-template-runtime used for?

pandamicro:
It's used for creating Code IDE project, you can create a project with it using cocos console

```
 cocos new -l js -t runtime
```

stammen:
@pandamicro  I will add the template  files in a separate pull request.

pandamicro:
OK, great! Thank you!

PR #12181, opened 2015-06-04
TintBy::clone() incorrect cast

namezero111111:
TintBy::clone() incorrectly casts the GLshort members to GLubyte, resulting in an object that isn't a clone and an incorrect tint.


WenhaiLin:
[ci rebuild]

WenhaiLin:
@pandamicro @andyque  It's OK for merge.

PR #12132, opened 2015-06-02
cocos2d/cocos2d-js#1658: Fix js_load_remote_image issue with multithread

pandamicro:

PR #12184, opened 2015-06-04
Fix issue that __JSDownloaderDelegator::onError get called twice

pandamicro:


pandamicro:
A completion for this PR: https://github.com/cocos2d/cocos2d-x/pull/12132#issuecomment-108794808

pandamicro:
About CURLOPT_NOSIGNAL, it's need based on the following research:

http://curl.haxx.se/libcurl/c/libcurl-tutorial.html#Multi-threading
http://curl.haxx.se/mail/lib-2014-02/0081.html

Otherwise, when timeout, libcurl will crash on curl_resolv_timeout, because the asynchronous call is executed in another thread

pandamicro:
[ci rebuild]

PR #12188, opened 2015-06-04

CocosRobot:


pandamicro:
@samuele3hu Should lua skip these functions of Component

see https://github.com/cocos2d/cocos2d-x/pull/12194

PR #12205, opened 2015-06-05
Update extension.h, amend a really  dangerous macros that may cause memory allocation error

ffskyfan:
# define MALLOC(TYPE,COUNT) ((TYPE*)_malloc(sizeof(TYPE) \* COUNT, __FILE__, **LINE**))

It is a really dangerous macros here.
If use it like this "MALLOC(Type, n+2)", it will alloc sizeof(TYPE) \* COUNT + 2 byte memory.but I want it alloc sizeof(TYPE) \* (COUNT + 2) byte. It is a typical macro unwind error.

### DIFF
--- a/cocos/editor-support/spine/extension.h
+++ b/cocos/editor-support/spine/extension.h
@@ -32,7 +32,7 @@
 #define SPINE_EXTENSION_H_
 
 /* All allocation uses these. */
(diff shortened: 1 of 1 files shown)


pandamicro:
I see the same fix have been applied to Spine runtime repo, thank you I will merge it
https://github.com/EsotericSoftware/spine-runtimes/blob/master/spine-c/include/spine/extension.h

PR #12212, opened 2015-06-05
call setTransparent

super626:
setTransparent is called.
Because the default value from RenderCommand is true.

### DIFF
--- a/cocos/3d/CCMesh.cpp
+++ b/cocos/3d/CCMesh.cpp
@@ -350,6 +350,7 @@ void Mesh::draw(Renderer* renderer, float globalZOrder, const Mat4& transform, u
 
 
     _meshCommand.setSkipBatching(isTransparent);
+    _meshCommand.setTransparent(isTransparent);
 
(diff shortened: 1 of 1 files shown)


dabingnn:
@super626 Stand at the maintenance point of view, could you please add a small piece of code to show the bug? The description does not show enough useful message.

dabingnn:
@super626 Stand at the maintenance point of view, could you please add a small piece of code to show the bug? The description does not show enough useful message.

super626:
@dabingnn 

I found that Riq moved the draw function  from Sprite3D to Mesh, and he missed this line when he added material support. We do not have this problem in the past version.

We do not need a test for it.

dabingnn:
@super626 Sounds good, could you please add the commit ID or pull request which introduced the bug? And let us merge it.
Thanks.

super626:
commit id: ce95bfe05945ba1bbce36ab3e4f03e2eaddbd7e7

PR #12218, opened 2015-06-05
samplers are treated as uniforms

ricardoquesada:
samplers are treated as uniforms

### DIFF
--- a/cocos/3d/CCMesh.cpp
+++ b/cocos/3d/CCMesh.cpp
@@ -283,6 +283,9 @@ void Mesh::setTexture(Texture2D* tex)
         auto technique = _material->_currentTechnique;
         for(auto& pass: technique->_passes)
         {
+            // FIXME: Ideally it should use glProgramState->setUniformTexture()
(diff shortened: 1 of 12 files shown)


cocos/3d/CCMesh.cpp
```diff
@@ -283,6 +283,9 @@ void Mesh::setTexture(Texture2D* tex)
         auto technique = _material->_currentTechnique;
         for(auto& pass: technique->_passes)
         {
+            // FIXME: Ideally it should use glProgramState->setUniformTexture()
+            // and set CC_Texture0 that way. But trying to it, will trigger
+            // another bug
```
dumganhar, line 288: @ricardoquesada, do you remember what's `another bug` here?

ricardoquesada: not really. what the issue with the code?

dumganhar: No yet, I just reviewed the code and found this `FIXME:` comment.
And I was confused with the comments '`will trigger another bug`'.

PR #12233, opened 2015-06-07
Fix C++ template CMakeLists.txt file to prevent include related errors on Windows

Furkanzmc:
My favorite IDE is Qt Creator. So I use Cmake for my projects. But the default CMakeLists.txt file has issues with builds on Windows and I had similar problems on OS X too but I only tested the changes I made on Windows. The commits pretty much explain themselves.


zilongshanren:
@Furkanzmc  Thanks.

PR #12238, opened 2015-06-08
remove old path When added existing new searchPath

litefeel:


litefeel:
@dabingnn please review this pr.

dabingnn:
@lite3 Thanks.
Did you mean that , if we add a search path which is existed in `searchPaths`, we should remove the old one?

litefeel:
@dabingnn yes

dabingnn:
@pandamicro @andyque 
What is your suggestions?
My suggestion is to merge it. Thanks.

zilongshanren:
@lite3  @dabingnn  I think we should keep the API as the old behavior. `addSearchPath` means doesn't consider duplication. If you want to reorder the search path, I think it's better to provide another two  APIs:

```
removeSearchPath
insertSearchPathAtIndex
```

I don't like one API takes care of too much details and make so many side effects. It's generally a bad smell when maintaining code.

litefeel:
@andyque 
I don't want to reorder it,  just to improve the efficiency of the search.

old behavior:
insert existed path to front is valid, append is invalid.

zilongshanren:
@lite3  I see. So the old behavior is "insert existing path to front"?

litefeel:
@andyque Now. it's keep old behavior.

zilongshanren:
@lite3 
I mean it's better to provide another remove path API to the `FileUtils` class:
Just put the following code in a `removeSearchPath` API.  

```
  auto iter = std::find(_searchPathArray.begin(), _searchPathArray.end(), searchpath);
+    if (iter != _searchPathArray.cend())
+    {
+        if (front)
+            _searchPathArray.erase(iter);
+        else
+            return;
+    }
```

litefeel:
@dabingnn I know. I just remove duplicate path.

litefeel:
[ci rebuild]

pandamicro:
@andyque I think it could be merged too, it doesn't change the behavior, and do improved the search efficiency .

zilongshanren:
@pandamicro  ok, maybe we could merge it in the next release.
Thanks @lite3

pandamicro:
Done

litefeel:
[ci rebuild]

PR #12239, opened 2015-06-08
Fix transform issue of PhysicsSprite itself and its children

pandamicro:


pandamicro:
I'm not sure the solution is a proper one.
I use scheduled `update` to synchronize the physics world to cocos world.

@minggo @andyque @WenhaiLin If you think there is a better solution, please let me know

pandamicro:
Originally reported : 
http://discuss.cocos2d-x.org/t/parent-child-physics-body-position-in-cocos-3-4/20090
http://discuss.cocos2d-x.org/t/physics-sprite-child-not-following-parent-until-window-is-resized/21792

pandamicro:
@WenhaiLin Please review this PR

pandamicro:
[ci rebuild]

extensions/physics-nodes/CCPhysicsSprite.cpp
```diff
@@ -393,22 +393,20 @@ void PhysicsSprite::syncPhysicsTransform() const
...
 
-// returns the transform matrix according the Chipmunk Body values
-const Mat4& PhysicsSprite::getNodeToParentTransform() const
+void PhysicsSprite::onEnter()
 {
-    syncPhysicsTransform();
-    
-	return _transform;
+    Node::onEnter();
+    scheduleUpdate();
```
WenhaiLin: It is possible that this `onEnter` method  is called more than once.Schedules the `update` method here may not be a good idea.

WenhaiLin: The `syncPhysicsTransform` method should be executed after the other scheduled callbacks is executed. Scheduler does not ensure it.

pandamicro:
@WenhaiLin  I have improved the logic with Director::EVENT_AFTER_UPDATE to ensure the transform is synced after scheduler's update.

TheCodez:
This issue still exists for normal Sprites and Nodes.

pandamicro:
@TheCodez  It has nothing todo with normal Sprites and Nodes, the issue is attached with Physics. Can you be more specific ?

TheCodez:
@pandamicro I have one Sprite and a PhysicsBody set to it, this Sprite has a Node attached, which also has a its own PhysicsBody set. Now the problem is, that moving the parent Sprite won't move the child Node's PhysicsBody.

This is also what is described in the forum, because he wasn't talking about PhysicsSprites but normal Sprites:
 http://discuss.cocos2d-x.org/t/parent-child-physics-body-position-in-cocos-3-4/20090

PR #12247, opened 2015-06-08
improve rendering of letter's inner shapes when outline is used

stari4ek:
Fixes "dirty" holes in letters, when outline is used 
outline_frag_fix


WenhaiLin:
@pandamicro @andyque It's OK for merge. @stari4ek Thank you!

PR #12248, opened 2015-06-08
MotionStreak opacity support

eschriek:
SetOpacity support for MotionStreak, allowing use of FadeTo and such
actions to work with MotionStreak

### DIFF
--- a/cocos/2d/CCMotionStreak.cpp
+++ b/cocos/2d/CCMotionStreak.cpp
@@ -51,6 +51,7 @@ MotionStreak::MotionStreak()
 , _vertices(nullptr)
 , _colorPointer(nullptr)
 , _texCoords(nullptr)
+, _opacity(0xFF)
 {
 }
 
@@ -126,7 +127,7 @@ bool MotionStreak::initWithFade(float fade, float minSeg, float stroke, const Co
     _blendFunc = BlendFunc::ALPHA_NON_PREMULTIPLIED;
 
     // shader state
-    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR));
+    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_A8_COLOR));
 
     setTexture(texture);
     setColor(color);
@@ -232,13 +233,12 @@ const BlendFunc& MotionStreak::getBlendFunc(void) const
 
 void MotionStreak::setOpacity(GLubyte opacity)
 {
-    CCASSERT(false, "Set opacity no supported");
+	_opacity = opacity;
 }
 
 GLubyte MotionStreak::getOpacity(void) const
 {
-    CCASSERT(false, "Opacity no supported");
-    return 0;
+	return _opacity;
 }
 
 void MotionStreak::setOpacityModifyRGB(bool bValue)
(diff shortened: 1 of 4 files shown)


zilongshanren:
Thanks, @eschriek 
Could you provide a test case for your patch? Just as we did in cpp-tests project.

zilongshanren:
[ci rebuild]

eschriek:
Yep, i'll push one later today

cocos/2d/CCMotionStreak.cpp
```diff
@@ -232,13 +233,12 @@ const BlendFunc& MotionStreak::getBlendFunc(void) const
 
 void MotionStreak::setOpacity(GLubyte opacity)
 {
-    CCASSERT(false, "Set opacity no supported");
+	_opacity = opacity;
```
pandamicro, line 236: Please use four space instead of a tab, check for all added lines

cocos/2d/CCMotionStreak.cpp
```diff
@@ -126,7 +127,7 @@ bool MotionStreak::initWithFade(float fade, float minSeg, float stroke, const Co
     _blendFunc = BlendFunc::ALPHA_NON_PREMULTIPLIED;
 
     // shader state
-    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR));
+    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_A8_COLOR));
```
zilongshanren, line 130: We can't change the shader here. The `SHADER_NAME_POSITION_TEXTURE_A8_COLOR` shader will ignore the texture color and only keep the texture opacity.
I

eschriek, line 130: I don't fully understand your problem here, the texture color might be ignored but is that necassarily a bad thing? For me personally I can't even notice it since the texture is just a white square.

Besided that, introducing a new shader which solves this is an easy fix.

zilongshanren, line 130: @eschriek  If you take a look at the tests in cpp-tests, there is test which do rendering the texture. 
After applying your patch, the test is break.

eschriek, line 130: You are right, but what i tried to say was that it is an easy fix.
The SHADER_NAME_POSITION_TEXTURE_A8_COLOR shader uses v_fragColor.rgb, if you modify that to use the texture2D rgb values it _should_ be fixed

alfogrillo:
@zilongshanren @eschriek 
It's this feature completed or has issues yet?

cocos/2d/CCMotionStreak.cpp
minggo: Why need to change the shader? And could you please fix the conflict?

eschriek: @minggo 

It has been over a year since I filed this pull request. I don't use cocos2d-x at the moment nor do I even have the code at my computer. In short, I can't help you further.

If I remember correctly there was some issue with the shader, even though I never fully understood that issue. 
The easiest solution is to keep the current MotionStreak class, and write some documentation on how to change it to add opacity support, or derive some MotionStreakOpacity class from it which provides opacity support, perhaps at the cost of some other feature.

minggo: @eschriek thanks for the information.

PR #12250, opened 2015-06-09
Fails if file does not exist

hawkwood:
This adds a callback with nullptr if the file does not exists.  Checking this before checking cache means that if a file is cached then deleted, the async call will fail.  To allow a cached but deleted file to return the texture, move the isFileExist check to ALTERNATIVE.

If leaving check where it is, on failure should it also check if the file was cached and remove it?

### DIFF
--- a/cocos/renderer/CCTextureCache.cpp
+++ b/cocos/renderer/CCTextureCache.cpp
@@ -102,6 +102,12 @@ void TextureCache::addImageAsync(const std::string &path, const std::function<vo
 
     std::string fullpath = FileUtils::getInstance()->fullPathForFilename(path);
 
+    // check if file exists
+    if (! FileUtils::getInstance()->isFileExist( fullpath )) {
+        callback(nullptr);
+        return;
+    }
+    
     auto it = _textures.find(fullpath);
     if( it != _textures.end() )
         texture = it->second;
@@ -111,6 +117,8 @@ void TextureCache::addImageAsync(const std::string &path, const std::function<vo
         callback(texture);
         return;
     }
+
+    // ALTERNATIVE: insert file exists check here to allow loading cached but deleted file
 
     // lazy init
     if (_asyncStructQueue == nullptr)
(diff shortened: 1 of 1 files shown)


pandamicro:
@jianglong0156 Please review this PR

cocos/renderer/CCTextureCache.cpp
```diff
@@ -102,6 +102,12 @@ void TextureCache::addImageAsync(const std::string &path, const std::function<vo
 
     std::string fullpath = FileUtils::getInstance()->fullPathForFilename(path);
 
+    // check if file exists
+    if (! FileUtils::getInstance()->isFileExist( fullpath )) {
```
WenhaiLin: `fullpath` is empty if the file wasn't found

sbrednikhin: And what will happen if file is failed to load (for example if it exists but, for example, in wrong format). Looks like additional check should be somewhere inside TextureCache::loadImage.

hawkwood: `fullpath` may contain data if `path` was absolute even if it doesn't exist, so line 106 should be

`if (fullpath.empty() || ! FileUtils::getInstance()->isFileExist( fullpath )) {`

A later fail by returning nullptr already happens.  During `addImageAsyncCallBack()` if `image` fails to load then `texture` stays `nullptr` and is returned regardless at `asyncStruct->callback(texture);`

cocos/renderer/CCTextureCache.cpp
```diff
@@ -102,6 +102,12 @@ void TextureCache::addImageAsync(const std::string &path, const std::function<vo
 
     std::string fullpath = FileUtils::getInstance()->fullPathForFilename(path);
 
+    // check if file exists
+    if ( fullpath.empty() || ! FileUtils::getInstance()->isFileExist( fullpath ) ) {
+        callback(nullptr);
+        return;
+    }
+
```
jianglong0156: If the image already loaded in texture, it's allow to loaded. So the code need to move the comment  after the textures.

zilongshanren:
@jianglong0156  What's the status of this PR?

jianglong0156:
@andyque @hawkwood I think the PR have a problem need to move the code to line 121

hawkwood:
Moved and added check if callback set

jianglong0156:
@andyque @pandamicro  review done. it's ok. can be merge

pandamicro:
[ci rebuild]

WenhaiLin:
@pandamicro It's OK for merge.

PR #12267, opened 2015-06-09
Capitalize default option in download-deps.py

oscr:
A minor improvement.
- By capitalizing "Yes" it's more clear that this is the default option.
- By having [] instead of () it's more clear that the input is optional.


zilongshanren:
@oscr 
Thanks

PR #12277, opened 2015-06-10
Change UI element(Menu and uiWidgets) hit tests from 2D space to 3D s…

yangws:
…pace.

And scene 3D test change to composite 2D/3D scene by moving UI and camera far away.

### DIFF
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -139,18 +139,23 @@ static bool camera_cmp(const Camera* a, const Camera* b)
     return a->getDepth() < b->getDepth();
 }
 
-void Scene::render(Renderer* renderer)
+const std::vector<Camera*>& Scene::getCameras()
 {
-    auto director = Director::getInstance();
-    Camera* defaultCamera = nullptr;
-    const auto& transform = getNodeToParentTransform();
     if (_cameraOrderDirty)
     {
         stable_sort(_cameras.begin(), _cameras.end(), camera_cmp);
         _cameraOrderDirty = false;
     }
-    
-    for (const auto& camera : _cameras)
+    return _cameras;
+}
+
+void Scene::render(Renderer* renderer)
+{
+    auto director = Director::getInstance();
+    Camera* defaultCamera = nullptr;
+    const auto& transform = getNodeToParentTransform();
+
+    for (const auto& camera : getCameras())
     {
         if (!camera->isVisible())
             continue;
@@ -167,7 +172,7 @@ void Scene::render(Renderer* renderer)
         //clear background with max depth
         camera->clearBackground(1.0);
         //visit the scene
-        visit(renderer, transform, 0);
+        visit(renderer, transform, Node::FLAGS_TRANSFORM_DIRTY);
         
         renderer->render();
         
--- a/cocos/2d/CCMenu.cpp
+++ b/cocos/2d/CCMenu.cpp
@@ -24,6 +24,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 ****************************************************************************/
 #include "2d/CCMenu.h"
+#include "2d/CCCamera.h"
 #include "base/CCDirector.h"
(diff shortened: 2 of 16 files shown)


yangws:
Forum thread: http://discuss.cocos2d-x.org/t/found-bugs-when-composite-3d-2d-mixed-scene-by-moving-ui-and-the-camera-which-look-at-far-away/21907

yangws:
[ci rebuild]

yangws:
Please review this PR @ricardoquesada  @masterhou

cocos/2d/CCScene.cpp
```diff
@@ -167,7 +172,7 @@ void Scene::render(Renderer* renderer)
         //clear background with max depth
         camera->clearBackground(1.0);
         //visit the scene
-        visit(renderer, transform, 0);
+        visit(renderer, transform, Node::FLAGS_TRANSFORM_DIRTY);
```
ricardoquesada, line 175: by passing dirty here, none of the optimizations will work. why this is needed?

ricardoquesada:
What is the purpose of this PR?
Perhaps it is needed. But my question we should avoid:
- adding new methods to node. In this case `hitTest` was added. Is this really needed in node? can it be moved to another part?
- it disables one of the most important optimizations of cocos2d-x: the dirty transform. This is serious. Nobody should make that change without really understanding the internals.

cocos/2d/CCScene.cpp
yangws, line 175: - hitTest logic is like a 3D version "Node.contertToNodeSpce + Rect.containPoint", for convenience add it to Node, or any ideas?
- If this flag don't set to Node::FLAGS_TRANSFORM_DIRTY, the node's transform matrix can't update correctly. Like this:
- if let playerDlg, which in Scene3DTest.cpp, do rotage action(code below), the result is:
- `_playerDlg->runAction(RotateBy::create(5000, Vec3(360000, 360000, 0)));`
  cpp_tests

ricardoquesada:
_hitTest logic is like a 3D version "Node.contertToNodeSpce + Rect.containPoint", for convenience add it to Node_

The more things that are added to Node, the more bloated it will be. Node must be thin... only the "must have" functions must be added into it. Everything inherits from Node, so Node must be as small as possible.
`hitTest` looks like a helper function. I guess it could be added outside Node.

_If this flag don't set to Node::FLAGS_TRANSFORM_DIRTY, the node's transform matrix can't update correctly_

The problem with this is that everything under `Scene` (which is the whole scene graph) will need to recalculate the matrix. What is the point of having that flag then?
So, the dirty flag must only be set when it is needed. If it is always needed, then there is a bug in the Scene/Camera code. But forcing to use the dirty_flag in Scene is worse.

cocos/2d/CCScene.cpp
yangws, line 175: HitTest is vritual function in UI::Widget class before, i reused this interface and move it to Node class.  
If Change it to a Helper function like HitTestInNodeContent(...),  i'll move hittest back to UI::Widget.
How about this solution?
- About Node::FLAGS_TRANSFORM_DIRTY problem, i'll try to find out other solution later.

ricardoquesada, line 175: _HitTest is vritual function in UI::Widget class before, i reused this
interface and move it to Node class.
If Change it to a Helper function like HitTestInNodeContent(...), i'll move
hittest back to UI::Widget._

If hitTest is only used in UI components, then, yes, I think it is better
to move it back to ui::Widget.

Thanks

dabingnn, line 175: @ricardoquesada Hi, Riq, there are some bugs when visiting the scene by multiple camera, because of the cached matrix does not work correctly. Passing the dirty flag in this function will make the cache invalid, but with performance degeneration. This bugs is missed, so let us record it, find the real bug and fix it, then we can remove this flag.

ricardoquesada, line 175: yes, let's find the real bug... but please, don't merge this PR with the DIRTY_FLAG on. Thanks.

dabingnn, line 175: Agreed. Thanks.

PR #12286, opened 2015-06-10
Remove incorrect README.md instruction

oscr:
This step doesn't work (and isn't needed) since you're already in the cocos2d-x folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,11 @@ Git user attention
 2. After cloning the repo, please execute `download-deps.py` to download and install dependencies.
 
          $ cd cocos2d-x
-         $ python download-deps.py
+         cocos2d-x $ python download-deps.py
 
 3. After running `download-deps.py`.
 
-         $ cd cocos2d-x
-         $ git submodule update --init
+         cocos2d-x $ git submodule update --init
 
 How to start a new game
 -----------------------


README.md
```diff
@@ -36,7 +36,6 @@ Git user attention
 
 3. After running `download-deps.py`.
 
-         $ cd cocos2d-x
```
pandamicro, line 39: It's more like a hint to let user know where the next command should be executed

zilongshanren, line 39: @oscr  I agree with @pandamicro

oscr, line 39: @pandamicro @andyque Thank you for your quick feedback!

If you just copy-paste the instructions from the readme you will get an error:

```
oscr@oscr-VirtualBox:~/temp/cocos2d-x$ python download-deps.py
(...)
oscr@oscr-VirtualBox:~/temp/cocos2d-x$ cd cocos2d-x
bash: cd: cocos2d-x: No such file or directory
```

This actually caused me some confusion when I got started. If we want the user to know where (s)he should be when a command is executed then maybe we could hint this by letting the command prompt show the current catalog? For example:

```
$ git clone https://github.com/cocos2d/cocos2d-x.git
$ cd cocos2d-x
cocos2d-x $ python download-deps.py
cocos2d-x $ git submodule update --init
```

pandamicro, line 39: Oh, I see, there is another `cd cocos2d-x` in the previous instruction.

I think it's ok to delete this line, but can you append `cocos2d-x` in the front of `$`, that would be helpful, thank you

PR #12290, opened 2015-06-11
[ci skip]add pr#2928 test case, load same remote img. can callback the function

jianglong0156:
relate pr:https://github.com/cocos2d/cocos2d-html5/pull/2930

this is the testcase

load same remote img. can callback the function


pandamicro:
Next time, if there is only script changes, please add [ci skip] to the PR title

PR #12299, opened 2015-06-11
Label:the texture of character have not cropped if character cross the axis-aligned bounding-box.

WenhaiLin:


pandamicro:
@qiululu As you have some experience in label rendering, if you like, you can take a look at our label system, helping review our issues. @WenhaiLin is the main maintainer of Cocos2d's `Label` class.

qiululu:
ok  I'll try, just make changes to some parameters to freetype or some computation.

PR #12336, opened 2015-06-15
modify js special resource path

jianglong0156:
modify js special resource path
modify all platform
android, ios, mac, win32, winphone. expect linux

I test it in android, ios, win32, winphone.


pandamicro:
[ci rebuild]

PR #12359, opened 2015-06-16
remove unused code, and use nullptr for pointer

super626:


dabingnn:

PR #12363, opened 2015-06-16
Fix terrain lod computing bugs 

super626:
1. lod calculate bug
2. bug when light mask = 0.
3. atanf->atanf2

### DIFF
--- a/cocos/3d/CCMesh.cpp
+++ b/cocos/3d/CCMesh.cpp
@@ -613,6 +613,7 @@ void Mesh::setLightUniforms(Pass* pass, Scene* scene, const Vec4& color, unsigne
         if (hasAmbient)
         {
             ambient.x /= 255.f; ambient.y /= 255.f; ambient.z /= 255.f;
+            //override the uniform value of u_color using the calculated color 
             glProgramState->setUniformVec4("u_color", Vec4(color.x * ambient.x, color.y * ambient.y, color.z * ambient.z, color.w));
         }
     }
--- a/cocos/3d/CCMesh.cpp
+++ b/cocos/3d/CCMesh.cpp
@@ -613,8 +613,8 @@ void Mesh::setLightUniforms(Pass* pass, Scene* scene, const Vec4& color, unsigne
         if (hasAmbient)
         {
             ambient.x /= 255.f; ambient.y /= 255.f; ambient.z /= 255.f;
+            glProgramState->setUniformVec4("u_color", Vec4(color.x * ambient.x, color.y * ambient.y, color.z * ambient.z, color.w));
         }
-        glProgramState->setUniformVec4("u_color", Vec4(color.x * ambient.x, color.y * ambient.y, color.z * ambient.z, color.w));
     }
 }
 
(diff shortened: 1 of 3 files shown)


pandamicro:
@dabingnn Please review
@super626 Please also update the status on redmine issue

super626:
@pandamicro 
I have modified the status of redmine.

cocos/3d/CCTerrain.cpp
```diff
@@ -159,11 +159,9 @@ void Terrain::onDraw(const Mat4 &transform, uint32_t flags)
 
     if(_isCameraViewChanged )
     {
-        auto camPos = camera->getPosition3D();
-        auto camModelMat = camera->getNodeToWorldTransform();
-        camModelMat.transformPoint(&camPos);
+        auto m = camera->getWorldToNodeTransform();
```
dabingnn: Why use `WorldToNodeTransform()` instead of `NodeToWorldTransform()` to get world position for Camera?

cocos/3d/CCMesh.cpp
```diff
@@ -613,8 +613,8 @@ void Mesh::setLightUniforms(Pass* pass, Scene* scene, const Vec4& color, unsigne
         if (hasAmbient)
         {
             ambient.x /= 255.f; ambient.y /= 255.f; ambient.z /= 255.f;
+            glProgramState->setUniformVec4("u_color", Vec4(color.x * ambient.x, color.y * ambient.y, color.z * ambient.z, color.w));
```
dabingnn, line 617: This function is called in Mesh::Draw function, and the function `setUniformVec4("u_color"` will be called in two separate functions, it is better to refactor it into one function.
Thanks.

zilongshanren:
@super626  Any updates with this PR?

cocos/3d/CCMesh.cpp
super626, line 617: setLightUniforms take effect only then there is light. We should also set the `u_color` when there is no light exist.

cocos/3d/CCTerrain.cpp
super626: I will have @tangziwen correct it.

cocos/3d/CCMesh.cpp
dabingnn, line 617: Yeah, No matter we have ambient light or not, we must set the uniform `u_color`.
 I would suggest we do it like this.

``` cpp
if(hasAmbients)
{
 glProgramState->setUniformVec4("u_color", a, b, c);
}
else
{
 glProgramState->setUniformVec4("u_color", d, e, f);
}
```

In this code fragment, it is easy to confuse others, because we do not setUniform here if there is no ambient, but set it in another function.

cocos/3d/CCTerrain.cpp
tangziwen: Sorry about that. In the test case, the negative camera position didn't make big difference about distance caculating, so i didn't notice it.

dabingnn:
@pandamicro @andyque It could be merged. Thanks.

PR #12365, opened 2015-06-16
更新头文件文件格式 从GBK到UTF8。

chengstory:


pandamicro:
@chengstory This modification is also needed in the v3 branch, please update them

PR #12369, opened 2015-06-16
Fix crash on Windows if pass string more than 16kb to cocos2d::log

Dimon4eg:
- crash on Windows

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -123,12 +123,12 @@ static bool isFloat( std::string myString ) {
 static ssize_t mydprintf(int sock, const char *format, ...)
 {
     va_list args;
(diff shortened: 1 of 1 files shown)


zilongshanren:
@Dimon4eg  Thanks. I have only one question with this PR. How could I reproduce the crash with a 16kb log?

Dimon4eg:
Hi @andyque.
Need to pass string more than 16kb because cocos uses 16kb buffer.
This reproduces the issue (on Windows only):

``` C++
string str;
for (int i = 0; i < 17000; ++i)
{
    str += 'A';
}
CCLOG("%s", str.c_str());
```

zilongshanren:
@Dimon4eg   Thanks.

PR #12372, opened 2015-06-16
using auto instead of auto& when the correct places

ricardoquesada:


zilongshanren:
why use auto instead of auto&?  The auto keyword would result extra copy in the loop.   auto& is fast?  correct me if I am wrong here.

ricardoquesada:
@andyque 
http://discuss.cocos2d-x.org/t/on-auto-vs-auto/22134/11

ricardoquesada:
@andyque @minggo 
if there are no objections, I'll merge this PR.

ricardoquesada:
Don't merge this PR

ricardoquesada:
invalid.

PR #12382, opened 2015-06-17
add librecast to cpp template

tangziwen:


super626:
@andyque 
@pandamicro

pandamicro:
Can you also add recast to JS and Lua template ?

pandamicro:
Thank you

PR #12418, opened 2015-06-18
fix compile error for linux

dabingnn:


dabingnn:
@pandamicro Please merge it, the linux compile passed.

PR #12428, opened 2015-06-18
修复 默认资源搜索路径

zhaolihang:


zilongshanren:
@zhaolihang  What  issue does this PR try to solve?

zhaolihang:
@andyque 设置默认的资源路径的时候 ，应该将这个路径加入到搜索路径中去啊，否则这个默认的资源路径是没有意义的

zilongshanren:
Hi, @zhaolihang
We can't accept this patch.
1.   You commit log is in Chinese.
2. We don't need to add the default path to search path. We get a file path, the root path will be used as the prefix.

PR #12434, opened 2015-06-18
add jsb navmesh test

jianglong0156:
add jsb navmesh test all platform


pandamicro:
Merged in: https://github.com/cocos2d/cocos2d-x/pull/12438

PR #12438, opened 2015-06-18
Bind Navmesh in JSB and add test case

pandamicro:


stammen:
@pandamicro Where is jsb_cocos2dx_navmesh_auto.hpp? The file appears to be missing. jsb_cocos2dx_navmesh_auto.cpp is also missing from the repo.

pandamicro:
The auto-bindings are always generated by CocosRobot automatically, so it's not included in this PR
The CocosRobot PR is here: https://github.com/cocos2d/cocos2d-x/pull/12453

PR #12442, opened 2015-06-19
Update the VS projects of templates.

natural-law:
1. Update the VS projects of templates. Add custom build step for copy resource files.
2. Solve the error usage of print in framework-compile scripts.


pandamicro:
[ci rebuild]

PR #12463, opened 2015-06-21
Set DEBUG_MODE on or off based on CMAKE_BUILD_TYPE

Furkanzmc:
When setting the CMAKE_BUILD_TYPE with Cmake parameter the setting gets overwritten by the option in cocos2d/CMakeLists.txt. With this, If the developer specifies that it should be a release build with `-DCMAKE_BUILD_TYPE=RelWithDebInfo` or `-DCMAKE_BUILD_TYPE=Release` DEBUG_MODE is set to OFF, other wise it stays on because it's the default value in cocos2d/CMakeLists.txt.

### DIFF
--- a/templates/cpp-template-default/CMakeLists.txt
+++ b/templates/cpp-template-default/CMakeLists.txt
@@ -37,6 +37,11 @@ include(CocosBuildHelpers)
 set(BUILD_CPP_TESTS OFF CACHE BOOL "turn off build cpp-tests")
 set(BUILD_LUA_LIBS OFF CACHE BOOL "turn off build lua related targets")
 set(BUILD_JS_LIBS OFF CACHE BOOL "turn off build js related targets")
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(DEBUG_MODE ON CACHE BOOL "turn on debug mode")
+else()
+  set(DEBUG_MODE OFF CACHE BOOL "turn off debug mode")
+endif()
 add_subdirectory(${COCOS2D_ROOT})
 
 # Some macro definitions


dabingnn:
@yangws Could you please review it?

templates/cpp-template-default/CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ include(CocosBuildHelpers)
 set(BUILD_CPP_TESTS OFF CACHE BOOL "turn off build cpp-tests")
 set(BUILD_LUA_LIBS OFF CACHE BOOL "turn off build lua related targets")
 set(BUILD_JS_LIBS OFF CACHE BOOL "turn off build js related targets")
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(DEBUG_MODE ON CACHE BOOL "turn on debug mode")
+else()
+  set(DEBUG_MODE OFF CACHE BOOL "turn off debug mode")
+endif()
```
yangws, line 44: Typically, cocos2d use option: DEBUG_MODE to control CMAKE_BUILD_TYPE, because cocos2d only test on Debug and Release mode.

Can you resolve your problem by this way?
In templates/cpp-template-default/CMakeLists.txt

```
if(DEBUG_MODE)
  set(CMAKE_BUILD_TYPE DEBUG)
else(DEBUG_MODE)
  set(CMAKE_BUILD_TYPE RELEASE)
endif(DEBUG_MODE)
```

then replace CMAKE_BUILD_TYPE with DEBUG_MODE in cmake command line.

Furkanzmc, line 44: I'm searching for occurrences of "DEBUG_MODE" in Cocos2dx v3.4 (That's the one I'm using currently.) and as you said cocos2d only uses "DEBUG_MODE" to control "CMAKE_BUILD_TYPE". Is that really necessary? CMake also has `RelWithDebInfo` and `MinSizeRel` options for `CMAKE_BUILD_TYPE`. Wouldn't it be better to just use `CMAKE_BUILD_TYPE`? Is there an advantage using `DEBUG_MODE` instead of `CMAKE_BUILD_TYPE`?
**EDIT:** OK. You mentioned that "... because cocos2d only test on Debug and Release mode". Sorry.

PR #12465, opened 2015-06-22
Allow optional custom logging function.

trojanfoe:
Allowing the setting of an optional custom logging function means all cocos2d-x logging messages can be processed by the game, thereby allowing low-level errors to be visible to crash reporting frameworks, for example.


minggo:
close it since there is conflict

PR #12467, opened 2015-06-22
fix a problem where WebSocket messages may pile up

billtt:
In the original implementation, only one message (frame) will be processed per tick, so the WebSocket messages will pile up if server sends them faster than client's frame rate.
Fixed by processing all messages in the queue in one update.

### DIFF
--- a/cocos/network/WebSocket.cpp
+++ b/cocos/network/WebSocket.cpp
@@ -185,8 +185,6 @@ void WsThreadHelper::joinSubThread()
 
 void WsThreadHelper::update(float dt)
 {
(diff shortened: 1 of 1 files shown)


pandamicro:
@yangws Can you help to review this PR ?

pandamicro:
[ci rebuild]

pandamicro:
Logic OK, I think it can be merged, @dumganhar Please confirm

pandamicro:
[ci rebuild]

dumganhar:
@pandamicro, it's ok for me for merging this patch. :)

PR #12474, opened 2015-06-23
Update 3d reader for compatible with cocosstudio 3.1.2 version csd file

xiaofeng11:
Update easing default speed


pandamicro:
Please also modify the JS parser

PR #12482, opened 2015-06-23
fixbug: cpp-test -> NewRenderer （drag the layer to test the result of…

yangws:
… culling）, some object will be culling before out of screen.


WenhaiLin:
[ci rebuild]

PR #12495, opened 2015-06-23
Fix terrain getIntersectionPoint

super626:


pandamicro:
Please fix the compilation issues

dabingnn:
What is the purpose of this pull request? Is it a bug fix or a refactor?

super626:
@pandamicro 
@dabingnn 
The old getIntersectionPoint is an approximate calculation of intersection point.

We improve it by calculate the intersection with triangles.

cocos/3d/CCTerrain.cpp
```diff
@@ -1255,6 +1257,37 @@ void Terrain::Chunk::calculateSlope()
...
+            float dist = ray._origin.distance(p);
+            if (dist<minDist)
+            {
+            interscetPoint = p;
+            minDist = dist;
+            }
+            isFind =true;
+        }
+    }
+    if (isFind)
```
dabingnn: we can use `return isFind` here

cocos/3d/CCTerrain.cpp
```diff
@@ -490,80 +491,62 @@ cocos2d::Vec3 Terrain::getNormal(int pixel_x, int pixel_y) const
...
-    Vec3 collisionPoint = (startPosition + endPosition) * 0.5f; 
-    return collisionPoint;
+    Vec3 collisionPoint;
+    if (getIntersectionPoint(ray, collisionPoint))
+    {
+        return collisionPoint;
+    }
+    else
+    {
+        return Vec3(0,0,0);
```
dabingnn, line 501: It is not very safe to return Vec3(0,0,0) here, because in some cases, the intersect point could be very near from the ZERO point.

cocos/3d/CCTerrain.cpp
```diff
@@ -490,80 +491,62 @@ cocos2d::Vec3 Terrain::getNormal(int pixel_x, int pixel_y) const
...
+        int y1 = floorf(start.y);
+        int y2 = ceilf(start.y);
+        for (int x = x1; x <= x2; x++) {
+            for (int y = y1; y <= y2; y++) {
+                auto chunk = getChunkByIndex(x, y);
+                if (chunk)
+                {
+                    if (closeList.find(chunk) == closeList.end())
+                    {
+                        if (chunk->getInsterctPointWithRay(ray, intersectionPoint))
```
dabingnn: What will happen if the ray has more than two intersection points?

cocos/3d/CCTerrain.cpp
```diff
@@ -1539,4 +1577,77 @@ Terrain::DetailMap::DetailMap()
...
+
+void Terrain::Triangle::transform(cocos2d::Mat4 matrix)
+{
+    matrix.transformPoint(&_p1);
+    matrix.transformPoint(&_p2);
+    matrix.transformPoint(&_p3);
+}
+
+bool Terrain::Triangle::getInsterctPoint(const Ray &ray, Vec3& interScetPoint)const
+{
```
dabingnn, line 1599: Could you please give more info about the algorithm? I can not understand it very well.

zilongshanren:
[ci rebuild]

cocos/3d/CCTerrain.cpp
super626, line 1599: 3D Math Primer for Graphics and Game Development

cocos/3d/CCTerrain.cpp
super626, line 501: cocos2d::Vec3 getIntersectionPoint(const Ray & ray) const should be deprecated.
Use bool getIntersectionPoint(const Ray & ray, Vec3 & intersectionPoint) const instead.

There is no correct value to return when there is no intersection point.

pandamicro:
[ci rebuild]

cocos/3d/CCTerrain.cpp
dabingnn, line 1599: After several hours, I found the original link for the algorithm:
https://en.wikipedia.org/wiki/M%C3%B6ller%E2%80%93Trumbore_intersection_algorithm

dabingnn:

PR #12501, opened 2015-06-23
Circle shape debug draw bug

namezero111111:
Circle shape debug draw didn't draw properly when rotated


WenhaiLin:
@namezero111111 Thank you! @zilongshanren It's OK for merge.

PR #12534, opened 2015-06-25
Skybox improvement

super626:
Make the skybox in projection space.
So that it doesn't affect by the camera position, near clipping plane and far clipping plane. It always the biggest and farthest in the screen.

### DIFF
--- a/cocos/renderer/ccShader_3D_Skybox.vert
+++ b/cocos/renderer/ccShader_3D_Skybox.vert
@@ -1,6 +1,5 @@
 const char* cc3D_Skybox_vert = STRINGIFY(
 
-uniform float u_scalef;
 uniform mat4  u_cameraRot;
 attribute vec3 a_position;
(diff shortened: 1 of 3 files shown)


pandamicro:
[ci rebuild]

cocos/renderer/ccShader_3D_Skybox.vert
```diff
@@ -1,12 +1,13 @@
 const char* cc3D_Skybox_vert = STRINGIFY(
 
-uniform float u_scalef;
+uniform mat4  u_cameraRot;
 attribute vec3 a_position;
 varying vec3 v_reflect;
```
dabingnn, line 6: It seems that Model,View,Project matrix are not considered in this shader. How to specify vertex data for Sky box?
Besides of this, it seems the uniform `u_cameraRot` will take scale into consideration if there is scale in `Camera's transform`.

PR #12541, opened 2015-06-25
[ci skip] Encoded into AVC

VisualSJ:


pandamicro:
Please repost it to v3.7-release, otherwise it won't be included in v3.7

PR #12568, opened 2015-06-26
Added keypad support

Cybertwip:
Keypad support for onKeyPressed and onKeyReleased events when using lua bindings.

### DIFF
--- a/cocos/2d/CCLayer.cpp
+++ b/cocos/2d/CCLayer.cpp
@@ -266,10 +266,21 @@ void Layer::onAcceleration(Acceleration* acc, Event* unused_event)
 #endif
 }
 
+/////////////////////////////////
+// greentwip fix for key codes //
+/////////////////////////////////
+
 void Layer::onKeyPressed(EventKeyboard::KeyCode keyCode, Event* unused_event)
 {
-    CC_UNUSED_PARAM(keyCode);
     CC_UNUSED_PARAM(unused_event);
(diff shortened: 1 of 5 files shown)


pandamicro:
@dualface Please review this PR

Cybertwip:
@pandamicro @dualface With the last commit the keypad support compiles for all platforms. It is being used a LuaValueDict to send the key and a boolean with the state of the key.

dualface:
i agree

cocos/2d/CCLayer.cpp
```diff
@@ -266,10 +266,21 @@ void Layer::onAcceleration(Acceleration* acc, Event* unused_event)
 #endif
 }
 
+/////////////////////////////////
+// greentwip fix for key codes //
```
pandamicro, line 270: What does this comment mean ?

Cybertwip, line 270: That's our organization. I did share the fork with my team and used that comment to give them proper indications about the changes done to the code.

pandamicro:
@VictorJL Sorry that I just got time to take a look at your PR. I found that your PR provide new functionalities for deprecated functions. We don't want to encourage our user to use `setKeyboardEnabled` anymore, so I don't think this PR can be merged to our current develop branch, it may be a good addon for v2 branch though.

Instead of using `setKeyboardEnabled`, you can use `EventDispatcher::addEventListenerWithXXX` which should support keypad.

PR #12575, opened 2015-06-28
fix onTouch begin don't return value

ywxzm:
fix only receive "begin" envent,can't receive "moved" "ended"


pandamicro:
@dualface Please review this PR

dualface:
i agree

pandamicro:
@ywxzm Thanks

Issue #10487, opened 2015-02-12
Camera fails to show sprites outside visible size


Tiger-Coding:
From the forum:
http://discuss.cocos2d-x.org/t/camera-bug-when-moving/19904/2

Using 3.4 final, I create a bunch of sprites, then position them one by one to the right.  I move the camera to the right, and everything off the screen is hidden.  However, if I move the layer to the left, I can see the sprites.

Code from forum post:

The init of HelloWorld:

```
auto spritebatch = SpriteBatchNode::create("MySprites.png");
addChild(spritebatch);
auto spriteCache = SpriteFrameCache::getInstance();
spriteCache->addSpriteFramesWithFile("MySprites.plist");
```

Then I create the sprites and add them to the layer, each one positioned to the right of the previous sprite.

```
for (int i = 0; i < 100; i++) {
    auto sprite = Sprite::createWithSpriteFrameName("MySprite.png");
    sprite->setPosition(Vec2(visibleSize.width/2 + origin.x + (visibleSize.width * i), visibleSize.height/2 + origin.y));
    this->addChild(sprite);
}
```

In the onEnter:

```
void HelloWorld::onEnter()
{
Layer::onEnter();
Size visibleSize = Director::getInstance()->getVisibleSize();
Camera *camera = Director::getInstance()->getRunningScene()->getCameras().front();
auto move = MoveBy::create(35.0, Vec3(10 * visibleSize.width, 0, 0));
camera->runAction(move);
}
```

Related?
https://github.com/cocos2d/cocos2d-x/issues/10275

marianopixowl:
I've experienced this issue too in cocos2d-x 3.4. It seems as if culling is not taking into account the camera movement, but only the original screen bounds. (Looking at _Renderer::checkVisibility()_ seems to confirm this. I can't see the camera being used in the visibility calculations).

For the time being, I'm setting _#define CC_USE_CULLING 0_ as a workaround. (Which by the way, creates a compilation error in Label::draw() since the code references a local variable only defined within the _#if CC_USE_CULLING ... #endif_ block).

zeyangl:
confirm issue exists in 3.6...

fusijie:
I have added a test case to reproduce this bug.

Renderer->Auto-culling is affected by transform and view matrix. Now the implementation ignore the view matrix dirty so lead to this bug.

[...]

PR #12586, opened 2015-06-29
add auto culling test case.

fusijie:


fusijie:
Relate to https://github.com/cocos2d/cocos2d-x/issues/10487

PR #12593, opened 2015-06-29
V3: WinRT unsupported cocos2d-x image formats removed or updated

stammen:
This pull request removes unsupported cocos2d-x image formats from the WinRT WIC converter (we use this native code to decode png files).


stammen:
Mac Slave went offline during the build
ERROR: Connection was broken: java.io.IOException: Unexpected termination of the channel

zilongshanren:
@stammen  Mac slave is not stable, I'm trying to resolve this issue.

PR #12598, opened 2015-06-30
Fixed iOS/JS-Bind callStaticMethod() with bool arg

shinhirota:
### Problem

Boolean arguments from **JavaScript** via `jsb.reflection.callStaticMethod()` cannot be received by C++ static methods like:

``` objc
+ (void) setAdVisible:(BOOL) isVisible {
   if (isVisible) { // <== This "isVisible" is always NO due to the problem.
    // do something
   } 
}
```

Here is the workaround we are using currently.

``` objc
// Workaround
+ (void) setAdVisible:(NSNumber*) isVisible {
   if ([isVisible booleanValue) { //<== Use NSNumber for boolean arguments.
    // do something
   } 
}
```

After this commit **"Fixed iOS/JS-Bind callStaticMethod() with bool arg."**, you can call the function of the upper example.


pandamicro:
@jianglong0156 please review this PR

jianglong0156:
the pr is right, can be merge.@pandamicro @shinhirota

jianglong0156:
i will add the testcase in the js-test. thank you for contribution.

shinhirota:
Great. Thank you for your support!

Issue #10171, opened 2015-01-22
[android]CppTests-->Renderer-->SpriteBatchTest:Touch screen might cause rendering order disorder when the screen have numerous sprites


WenhaiLin:

dabingnn:
linux has this bug too.

fusijie:
The test case is not exist now . @WenhaiLin

WenhaiLin:
Test case:`Node:Sprite-->Testing SpriteBatchNode` has this bug too.

fusijie:
@WenhaiLin Please help to check. Thanks.

PR #12600, opened 2015-06-30
Fix SpriteBatchNode sort error.

fusijie:


fusijie:
To fix https://github.com/cocos2d/cocos2d-x/issues/10171

Issue #8305, opened 2014-09-30
CCLabelTTF - sometimes rendering as black boxes on Android


minggo:
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/3669). It was created at 2014/01/10 20:00:25 +0000
# Description

From time to time, if I exit my game on Android while text is being displayed with CCLabelTTF, then when I re-enter the game, the text is replaced with black boxes. I’m guessing the texture is getting lost. All other textures are fine though.

Another way to reproduce the issue:
# start game
# launch some activity (such as an interstitial ad)
# close that activity and return to game
# at this point, have CCLabelTTF display some text ==> here the text is just black boxes

Also, here’s a related forum post in which more than one developer have experienced this issue:
http://cocos2d-x.org/forums/6/topics/42500
# Comments
- Sheado added comment:
  Ok,

After tinkering I found the cause of the problem.
I was preparing my CCLabelTTF in applicationWillEnterForeground(). Simply moving the code to a runAction(…callfunc…) fixes the problem.

I don’t know if this is a bug or if the documentation just needs to be updated to warn people that they should schedule the work to be done on resuming the application.
- wollf added comment:
  Problem exists with new Label::createFromTTF. If i create a label label like so:
  `Label::createWithTTF(fontBig,"Example ttf text",TextHAlignment::RIGHT);`
  I hide app to background (pressing HOME button) and when i re-enter it i get "white boxes".
  normal label:
  ![before](http://i.imgur.com/Wia5I54.png)
  after re-entering:
  ![after](http://i.imgur.com/VudTjOd.png)

I think this is caused by not reloading ttfConfig created resources in reloadAllTextures.
This is bad and must be fixed, though i can't deal with this myself :(
Tested Device: Google Nexus 7(2013)
- yaxinhoo added comment:
  @Sheado i got the same , does it fixed?
- WenHai added comment:
  Hi guys,which version of Cocos2d-x are you using?

WenhaiLin:
@minggo Confirmed fixed in v3 branch.We should close this issue.

MelonS:
Confirmed fixed in v3 branch?
@minggo @Dhilan007 How to fix? Let me know commit number.

ahmed-belal:
@minggo @Dhilan007 Can you please let me know the commit number? I need to merge this fix into my application running @ v3.2

WenhaiLin:
[...]

PR #12609, opened 2015-06-30
Ensure CCLabel atlases are purged before resetting

alexpeak:
This is a fix for Github issue #8305

The current purge->reset behavior occurs per label according to scene
graph priority.  This has two disadvanages; firstly, any labels that are
initialized but not part of the scene graph will not be reset.
Secondly, any labels which share atlases in the FontAtlasCache may not
be reset since each label will call releaseFontAtlas individually before
resetting its fontConfig.  This means that any shared atlases may not be
released all the way to a reference count of 0 and the invalid atlas
will not be deleted and reset from scratch.

By putting a fixed priority on both the purge and reset steps we can
ensure that firstly: all labels will be reset regardless of whether they
have been added to the scene-graph.  Secondly, ALL labels with the same
atlas must call releaseFontAtlas before any others are allowed to
rebuild their textures.  This will prevent any old atlases from sticking
around after a new GL context is created.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -262,7 +262,7 @@ Label::Label(FontAtlas *atlas /* = nullptr */, TextHAlignment hAlignment /* = Te
     setAnchorPoint(Vec2::ANCHOR_MIDDLE);
     reset();
 
-    auto purgeTextureListener = EventListenerCustom::create(FontAtlas::CMD_PURGE_FONTATLAS, [this](EventCustom* event){
+    _purgeTextureListener = EventListenerCustom::create(FontAtlas::CMD_PURGE_FONTATLAS, [this](EventCustom* event){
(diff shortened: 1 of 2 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -294,6 +294,9 @@ Label::~Label()
     if (_fontAtlas)
     {
         FontAtlasCache::releaseFontAtlas(_fontAtlas);
+
+        _eventDispatcher->removeEventListener(_purgeTextureListener);
+        _eventDispatcher->removeEventListener(_resetTextureListener);
```
WenhaiLin: These two lines should not place into _if_  block.

alexpeak: Of course, fixed in new commit.  Thanks for the quick response.

WenhaiLin:
@pandamicro @andyque It's OK for merge. @alexpeak Thank you!

PR #12613, opened 2015-07-01
ignore file name case sensitive under windows system to avoid checkFile report file not exist

xiaofeng11:


WenhaiLin:
Bad idea.I think we should check file name case at all platform.

PR #12614, opened 2015-07-01
Remove unused link libs config in cocos2dx.props.

natural-law:


bfloch:
Any good reasons why this was rejected?

It basically means that the CC_USE_PHYSICS=0 defines won't work as VS will still try to link box2d, also libSpine if you decide to remove that dependency.

IMO the optional dependencies should go to the vcproj.

bfloch:
I think for the pullrequest to be accepted the removed libraries should go to the vs project generation instead, right?

PR #12618, opened 2015-07-01
Sprite3D Clipping

super626:
1. Remove stencil from Renderstate.
2. Add clippingnode test for sprite3d.


pandamicro:
This PR have been reviewed in previous PR and discussed here: http://discuss.cocos2d-x.org/t/how-to-make-clipping-node-work-for-3d/22252

I merged this PR in Cocos2d-x v3.7 RC0, but in case it have some side effect, let's collect more user feed backs in RC0, please stay tuned

PR #12419, opened 2015-06-18
Modify Scene3DTest to composite 2D/3D scene by moving UI and camera f…

yangws:
…ar away.

PR #12620, opened 2015-07-01
Fix bug: wrong usage of function isVisiableByVisitingCamera, which should only effect draw call.

yangws:
Related PR: https://github.com/cocos2d/cocos2d-x/pull/12419


pandamicro:
@yangws Can you create an issue and link all related PRs in the issue, so that I can schedule the issue into v3.8

dabingnn:
@pandamicro This pull request should be merged, given that `isVisitableByVisitingCamera()` should only affect the draw function of the node. Now it affect the `draw` and `Transform update` of the node, which is not correct.

pandamicro:
Thanks for the review

PR #12635, opened 2015-07-02
Fix bug: spine animition can't show in Scene3DTest.

yangws:


pandamicro:
@dabingnn Please review

dabingnn:
@yangws Could you please describe the purpose of this pull request?

yangws:
In Scene3DTest draw call, Terrian enable GL_CULLING, but Spine Animation need this state disabled, but in Spine draw function don't call glDisable(GL_CULLING), so in this test case , need wrapper spine animation's draw call to disable culling manually, thus spine animaiton show correctly.
@dabingnn

PR #12636, opened 2015-07-02
CheckBox bug fix

xiaofeng11:
Fix check box front cross texture expand to normal size when change status between normal and disable frequency


zilongshanren:
@pandamicro  This bug fix is mainly for CocoStudio.

Issue #9878, opened 2015-01-08
Be able to remove key pairs from UserDefault


smitpatel88:
First create any key field,
UserDefault::getInstance()->setIntegerForKey("keyName", 1);
after that we cannot delete "keyName" from database.
If we can remove unwanted fields from database, less memory will occupied.

walzer:
Good idea! We should add the `remove` feature in `UserDefault`.
Added in v3.8.

PR #12643, opened 2015-07-02
Add userdefaut UserDefault::deleteValueForKey

jianglong0156:
Add userdefaut UserDefault::deleteValueForKey

already test in android, win32, ios, mac, winrt

### DIFF
--- a/cocos/base/CCUserDefault-android.cpp
+++ b/cocos/base/CCUserDefault-android.cpp
@@ -521,6 +521,12 @@ void UserDefault::flush()
 
 bool UserDefault::deleteValueForKey(const char* key)
 {
+    // check the params
+    if (!key)
+    {
+        return false;
+    }
+    
     deleteValueForKeyJNI(key);
     return true;
 }
--- a/cocos/base/CCUserDefault-android.cpp
+++ b/cocos/base/CCUserDefault-android.cpp
@@ -60,24 +60,24 @@ static tinyxml2::XMLElement* getXMLNodeForKey(const char* pKey, tinyxml2::XMLDoc
 {
     tinyxml2::XMLElement* curNode = nullptr;
     tinyxml2::XMLElement* rootNode = nullptr;
-    
+
     if (! UserDefault::isXMLFileExist())
     {
         return nullptr;
     }
-    
+
     // check the key value
     if (! pKey)
     {
         return nullptr;
     }
-    
+
     do
     {
         tinyxml2::XMLDocument* xmlDoc = new tinyxml2::XMLDocument();
         *doc = xmlDoc;
         ssize_t size;
-        
+
         std::string xmlBuffer = FileUtils::getInstance()->getStringFromFile(UserDefault::getInstance()->getXMLFilePath().c_str());
 
         if (xmlBuffer.empty())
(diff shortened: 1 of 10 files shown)


pandamicro:
@andyque Please review this PR

@jianglong0156 Please link this PR to related issue

jianglong0156:
sorry for my misstake. the related issue is https://github.com/cocos2d/cocos2d-x/issues/9878

cocos/base/CCUserDefault-android.cpp
```diff
@@ -519,6 +519,17 @@ void UserDefault::flush()
 {
 }
 
+bool UserDefault::deleteValueForKey(const char* key)
+{
+    // check the params
+    if (!key)
```
zilongshanren, line 525: It's better to output some debug info to tell why it fails.
Like `CCLOG("key can't be nullptr!")`.

cocos/base/CCUserDefault-apple.mm
```diff
@@ -542,6 +542,20 @@ static void deleteNodeByKey(const char *pKey)
     [[NSUserDefaults standardUserDefaults] synchronize];
 }
 
+bool UserDefault::deleteValueForKey(const char* key)
+{
+    // check the params
+    if (!key)
```
zilongshanren, line 548: The same as above comments.

cocos/base/CCUserDefault-android.cpp
```diff
@@ -519,6 +519,17 @@ void UserDefault::flush()
...
+    {
+        return false;
+    }
+    
+    deleteValueForKeyJNI(key);
```
zilongshanren, line 530: missing a `flush` function here?

cocos/base/CCUserDefault-winrt.cpp
```diff
@@ -329,6 +329,21 @@ void UserDefault::flush()
...
+{
+    // check the params
+    if (!key)
+    {
+        return false;
+    }
+
+    ApplicationDataContainer^ localSettings = ApplicationData::Current->LocalSettings;
+    auto values = localSettings->Values;
+    values->Remove(PlatformStringFromString(key));
```
zilongshanren, line 342: do we need call `flush` here?

cocos/base/CCUserDefault.h
```diff
@@ -202,6 +202,14 @@ class CC_DLL UserDefault
      */
     virtual void flush();
 
+    /**
+    * delete any value by key, if the key doesn't exist, will return true.
```
zilongshanren: What does it mean "if the key doesn't exist, will return true"?
If the parameter is nullptr, you return false. If the key doesn't exist, you return false.
I think we can't tell these two cases apart, right?

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
```diff
@@ -526,6 +526,13 @@ public static void setStringForKey(String key, String value) {
         editor.commit();
     }
     
+    public static void deleteValueForKey(String key) {
```
zilongshanren, line 529: This function should return bool? So we can tell whether it delete value successfully or not.

jianglong0156:
arrange new pr:https://github.com/cocos2d/cocos2d-x/pull/12697

PR #12647, opened 2015-07-02
UISlider update

xiaofeng11:
Fix when slider set to disable mode then clean slider ball disable texture, normal texture won't show


zilongshanren:
@pandamicro  This fix is also needed for CocoStudio.

Issue #8262, opened 2014-09-30
CCTargetedAction executes callback twice


minggo:
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/3264). It was created at 2013/11/22 07:19:03 +0000
# Description

```
CCLayer::init() {
…
CCSprite* r = CCSprite::create(“xxx”);
this->addChild;
CCSprite* e = CCSprite::create;
this->runAction(CCTargetedAction::create(e, CCCallFuncN::create(r, callfuncN_selector(CCSprite::addChild)))); // CCSprite::addChild will be called twice
e->runAction(CCCallFuncN::create(r, callfuncN_selector(CCSprite::addChild))); // ok
}
```

CCTargetedAction will make “r->addChild(e)” called twice, then CCAssert will fail.

refer: http://cocos2d-x.org/forums/6/topics/39546

PR #12670, opened 2015-07-03
Fixbug: CCTargetedAction executes callback twice.

yangws:
The reason is TargetedAction don't override the parent(ActionInterval)'s function isDone().
The TargetedAction's isDone logic is different of ActionInterval.

https://github.com/cocos2d/cocos2d-x/issues/8262


pandamicro:
Thanks for the fix

Issue #9156, opened 2014-11-18
I think the types to handle time are not consistent.


sappho192:
I found DisplayLinkDirector::setAnimationInterval gets argument "interval" which is double, and schedule methods in Node class get argument "delay" which is float.
AFAIK, schedule methods are belong to setAnimationInterval but they have different types to control time. I think this makes data loss and dulls time precision, right?

It needs to be same type(especially double) to save data and sharpen the precision.

minggo:
Yep, you are right.
I think we should use float in `DisplayLinkDirector::setAnimationInterval()` too.

sappho192:
How about unify to double?

minggo:
Because most of other API uses float.

PR #12689, opened 2015-07-06
Refine setAnimationInterval from double to float.

fusijie:
@andyque please help to check.


fusijie:
To fix https://github.com/cocos2d/cocos2d-x/issues/9156

Issue #9621, opened 2014-12-25
Really needed a set of file access functions


pandamicro:
Cocos2d-x have three option for data storage: 
- FileUtils::writeToFile
- localStorageSetItem
- UserDefault

I want to talk about `FileUtils` which have a whole set of functions to read a file, and one function to write a map object to a file. The balance is obviously not quite well here. The reason we need a set equivalent functions to write content into a file are: 
1. To balance the API design.
2. In JS or Lua, if there are no such API bindings, then no way to write even a simple string into a file, user must write C++ code to achieve this and bind them manually.

So please consider it

minggo:
What functions you think are needed?
Could you write more detail of requirement?

pandamicro:
What we have now for reading:
- FileUtils::getStringFromFile
- FileUtils::getDataFromFile
- FileUtils::getFileDataFromZip
- FileUtils::getValueMapFromFile
- FileUtils::getValueVectorFromFile

For write access:
- FileUtils::writeToFile

The API `writeToFile` doesn't represent what we understand from the name, it actually write a value map to a file.

What I suggest:
- FileUtils::writeStringToFile
- FileUtils::writeDataToFile
- FileUtils::writeValueMapToFile // which is the current writeToFile
- FileUtils::writeValueVectorToFile

minggo:
Thanks.
We will consider the requirements.

pandamicro:
Here is a related user request: https://github.com/cocos2d/cocos2d-js/issues/1315

pandamicro:
Another user requested similar features for FileUtils: https://github.com/cocos2d/cocos2d-js/issues/1323
1. Save content to file
2. Uncompress zip file
3. Calculate MD5 value for verification

dumganhar:
> Another user requested similar features for FileUtils: cocos2d/cocos2d-js#1323
> 
> Save content to file
> Uncompress zip file
> Calculate MD5 value for verification

It will be awesome if these features could be supported. :)
It's really a strong requirement for games which use `Cocos2d-JS`. So please support it.

walzer:
Do it in v3.8. Marked as highlight.

TimothyZhang:
Unfortunately, `writeDataToFile` is not supported yet by JSB in cocos2d-x v3.8 

I created a pull request #13960 for this.

jminht:
how about Lua?
we also need support like FileUtils::getDataFromFile.
@pandamicro @minggo

PR #12690, opened 2015-07-06
add fileUtils write functions(add four api in FileUtils)

jianglong0156:
add fileUtils write functions

```
writeStringToFile(std::string dataStr, const std::string& fullPath);
writeDataToFile(Data retData, const std::string& fullPath);
writeValueMapToFile(ValueMap& dict, const std::string& fullPath);
writeValueVectorToFile(ValueVector vecData, const std::string& fullPath);
```

and add the testcase


jianglong0156:
related issue:https://github.com/cocos2d/cocos2d-x/issues/9621

PR #12697, opened 2015-07-06
add UserDefault api deleteValueForKey

jianglong0156:
add UserDefault api deleteValueForKey


jianglong0156:
the related issue is #9878

PR #12709, opened 2015-07-07
Fixed crash cause by the case of file name on WIN32.

WenhaiLin:
We will rollback the changes in the future.


pandamicro:
This modification is for avoiding crash in Cocos Simulator for Cocos v2.3.1. in the next version we will roll back the changes, and verify the file name while publishing in Cocos Studio, so that it won't have the file name issue.

Issue #12626, opened 2015-07-01
Engine support composite 2D/3D scene by moving UI and camera far away.


yangws:
Forum thread: http://discuss.cocos2d-x.org/t/found-bugs-when-composite-3d-2d-mixed-scene-by-moving-ui-and-the-camera-which-look-at-far-away/21907

@pandamicro 

This function seperated to several PRs, list below:
- [x] https://github.com/cocos2d/cocos2d-x/pull/12419
- [x] https://github.com/cocos2d/cocos2d-x/pull/12602 :bugfix https://github.com/cocos2d/cocos2d-x/pull/12711 , https://github.com/cocos2d/cocos2d-x/pull/13000
- [x] https://github.com/cocos2d/cocos2d-x/pull/12620
- [x] https://github.com/cocos2d/cocos2d-x/pull/12635

yangws:
All PRs are done. 
@pandamicro

pandamicro:
Thanks

PR #12602, opened 2015-06-30
Fixed bug: UI component can't click correctly by moving UI and camera far away of origin.

yangws:
Related PR: https://github.com/cocos2d/cocos2d-x/pull/12419

PR #12711, opened 2015-07-07
Fix bug taken from pr12602: scroll view can't scroll if pressed the b…

yangws:
…utton on it.

the related issue and pr:
https://github.com/cocos2d/cocos2d-x/issues/12626
https://github.com/cocos2d/cocos2d-x/pull/12602

### DIFF
--- a/cocos/ui/UIWidget.cpp
+++ b/cocos/ui/UIWidget.cpp
@@ -783,7 +783,9 @@ void Widget::propagateTouchEvent(cocos2d::ui::Widget::TouchEventType event, coco
     Widget* widgetParent = getWidgetParent();
     if (widgetParent)
     {
+        widgetParent->_hittedByCamera = _hittedByCamera;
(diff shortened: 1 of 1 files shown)


cocos/ui/UIWidget.cpp
```diff
@@ -783,7 +783,9 @@ void Widget::propagateTouchEvent(cocos2d::ui::Widget::TouchEventType event, coco
     Widget* widgetParent = getWidgetParent();
     if (widgetParent)
     {
+        widgetParent->_hittedByCamera = _hittedByCamera;
```
pandamicro, line 786: Shouldn't the previous value of `widgetParent->_hittedByCamera` be saved ? It may not be nullptr

yangws, line 786: _hittedByCamera is a temp variable in touch begin->move->end event period, don't need save at all.

PR #12723, opened 2015-07-07
Add a setter for line width in DrawNode

neokim:
It was unable to change DrawNode line width which was set to 2. Add a setter for line width.

### DIFF
--- a/cocos/2d/CCDrawNode.cpp
+++ b/cocos/2d/CCDrawNode.cpp
@@ -104,6 +104,8 @@ static inline Tex2F __t(const Vec2 &v)
 
 // implementation of DrawNode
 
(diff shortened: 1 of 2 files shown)


zilongshanren:
@neokim  Is the line width always be 2 in all OpenGL & OpenGLES implementation?

neokim:
@andyque I have no idea about that. But in current implementation of DrawNode, the line width is hard-coded as 2 and not adjustable.

zilongshanren:
@neokim  I see. If it  passes the CI, I will merge this PR.
Thanks.

zilongshanren:
@neokim Thanks :+1:

PR #12726, opened 2015-07-07
Fixes some warnings and a related bug in CCSprite.

xpol:
This story tells us that we should not just ignore warnings.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -223,7 +223,7 @@ bool Sprite::initWithPolygon(const cocos2d::PolygonInfo &info)
 {
     Texture2D *texture = Director::getInstance()->getTextureCache()->addImage(info.filename);
     bool res = false;
-    if(initWithTexture(texture));
+    if(initWithTexture(texture))
     {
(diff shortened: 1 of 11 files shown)


pandamicro:
Thanks @xpol 
@VirtualSJ Please review the modifications in readers. Do we need to change the JSON parser too?

cocos/2d/CCSprite.cpp
```diff
@@ -223,7 +223,7 @@ bool Sprite::initWithPolygon(const cocos2d::PolygonInfo &info)
 {
     Texture2D *texture = Director::getInstance()->getTextureCache()->addImage(info.filename);
     bool res = false;
-    if(initWithTexture(texture));
```
zilongshanren, line 226: good catch!

zilongshanren:
@xpol  Thanks :+1:

zilongshanren:
@xpol  I also hate warnings, but now there are tons of warnings and many of them are in 3rd party libs..

xpol:
@zilongshanren thanks. Lets get the warnings fixed.

Issue #11887, opened 2015-05-18
Need to replace thread unsafe code from TextureCache::addImageAsync


sbrednikhin:
TextureCache::loadImage will be runnedin parallel thread by TextureCache::addImageAsync, but _textures.find() and _textures.end() used inside aren't thread-safe and there is no lock.

http://discuss.cocos2d-x.org/t/looks-like-texturecache-loadimage-is-not-thread-safe/21414

walzer:
Yes, this is a critical bug!

0xJimmyBeh:
@walzer 
There is a long existing issue with TextureCache.addImageAsync in Cocos2d-JS too.

https://github.com/cocos2d/cocos2d-js/issues/1658

Please solve them together since all engine going to merge soon.

Thanks.

PR #12735, opened 2015-07-08
Fix bug: Need to replace thread unsafe code from TextureCache::addImageAsync

yangws:
relative issue:
https://github.com/cocos2d/cocos2d-x/issues/11887


yangws:
@pandamicro please review & merge, thanks.

pandamicro:
I'm not really qualified for reviewing thread relative code, @dabingnn please take a look, thanks

cocos/renderer/CCTextureCache.cpp
```diff
@@ -139,213 +170,151 @@ void TextureCache::addImageAsync(const std::string &path, const std::function<vo
...
-        
-        if(image)
-        {
-            image->release();
-        }       
+
+        // release the asyncStruct
         delete asyncStruct;
-        delete imageInfo;
+        --asyncStruct;
```
dabingnn: Why using `--asyncStruct`?

PR #12736, opened 2015-07-08
Fixed inertial scrolling for `CCScrollView`

marzapower:
The inertial scrolling in `CCScrollView` usually works only if you're using the `Direction::BOTH` direction, and if there's enough scrolling space **both** on the X and the Y axis. But when you use the `Direction::HORIZONTAL` or `Direction::VERTICAL` directions you usually have 0px of available space on the Y or X axis respectively, so it won't work.
This change enables you to scroll the view just on the X or on the Y axis using `Direction::HORIZONTAL` or `Direction::VERTICAL` and to use the inertial scroll accordingly.


zilongshanren:
@marzapower  thanks. merged.

PR #12737, opened 2015-07-08
Fix cocos reader bug.

chengstory:
Fixes child nodes can't be rendered when particle and TiledMap as parent and their resource have been removed from disk.


pandamicro:
@VisualSJ Please review

VisualSJ:
@pandamicro it's ok for merge...

PR #12740, opened 2015-07-08
improve coding style of Socket.IO

zilongshanren:


cocos/network/SocketIO.cpp
```diff
@@ -57,32 +52,32 @@ class SocketIOPacketV10x;
...
-    void initWithType(std::string packetType);
+    void initWithType(const std::string& packetType);
     void initWithTypeIndex(int index);
 
-    std::string toString();
-    virtual int typeAsNumber();
-    std::string typeForIndex(int index);
+    std::string toString()const;
+    virtual int typeAsNumber()const;
+    const std::string& typeForIndex(int index)const;
```
ZhangMenghe, line 68: Only if the returned string doesn't need to be changed,or else change to 
std::string& typeForIndex(int index)const;

cocos/network/SocketIO.cpp
```diff
@@ -57,32 +52,32 @@ class SocketIOPacketV10x;
...
 
-    void setEndpoint(std::string endpoint){ _endpoint = endpoint; };
-    std::string getEndpoint(){ return _endpoint; };
+    void setEndpoint(const std::string& endpoint){ _endpoint = endpoint; };
+    const std::string& getEndpoint()const{ return _endpoint; };
     void setEvent(std::string event){ _name = event; };
```
ZhangMenghe: change to 
void setEvent(const std::string& event)

cocos/network/SocketIO.cpp
```diff
@@ -57,32 +52,32 @@ class SocketIOPacketV10x;
...
-    void addData(std::string data);
-    std::vector<std::string> getData(){ return _args; };
-    virtual std::string stringify();
+    void addData(const std::string& data);
+    std::vector<std::string> getData()const{ return _args; };
+    virtual std::string stringify()const;
 
-    static SocketIOPacket * createPacketWithType(std::string type, SocketIOPacket::SocketIOVersion version);
-    static SocketIOPacket * createPacketWithTypeIndex(int type, SocketIOPacket::SocketIOVersion version);
+    static SocketIOPacket * createPacketWithType(std::string type, SocketIOVersion version);
```
ZhangMenghe, line 79: Add const to the end of this function. The same to L80
static SocketIOPacket \* createPacketWithType(std::string type, SocketIOPacket::SocketIOVersion version)const;

ZhangMenghe:
And other two queries are as following:
1.In L36.37,Shall we include the file with absolute path?
2.For functions like L362, when the parameters are of pointer type , whether  'const' is needed?
@zilongshanren

cocos/network/SocketIO.cpp
zilongshanren, line 79: We can't add `const` here since the function `createPacketWithType` will modify the state of the object.

cocos/network/SocketIO.cpp
zilongshanren: good catch

cocos/network/SocketIO.cpp
zilongshanren, line 68: Why return `std::string&`? I think we don't want the developer to modify the return value of `typeForIndex`.

ZhangMenghe, line 68: Just make sure because sometimes we need to modify the returned Object ;)

zilongshanren:
@ZhangMenghe  
`1.In L36.37,Shall we include the file with absolute path?`  No, there are in the same directory. I think it's OK.

`2.For functions like L362, when the parameters are of pointer type , whether 'const' is needed?`
pointers doesn't need to add `const`.

ZhangMenghe:
@zilongshanren 
Thanks for corrections.
But I consider that pointer also represents an address like a reference,which we do not want them to be changed. 
And here is  a reference of pointer parameter (whether const or not )
http://www.functionx.com/cppcli/functions/Lesson02c.htm

zilongshanren:
[ci rebuild]

zilongshanren:
@ZhangMenghe  That's true. But in cocos2d-x, we don't add `const` to pointer args. 
https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Reference_Arguments
You could open a ticket in the forum about using `const` in pointer args.
Thanks.

zilongshanren:
[ci rebuild]

himynameschris:
awesome, glad to see this made it. thanks all!

PR #12742, opened 2015-07-08
AudioEngine:support preload for audio

WenhaiLin:
Pre-load feature not support for Android.
We need to be responsible for decode audio if support pre-load on Android.
It's mean we need decode audio by a third-party library or give up Android below version 4.0.

### DIFF
--- a/cocos/audio/AudioEngine.cpp
+++ b/cocos/audio/AudioEngine.cpp
@@ -415,4 +415,18 @@ AudioProfile* AudioEngine::getProfile(const std::string &name)
     }
 }
 
+void AudioEngine::preload(const std::string& filePath)
+{
+    lazyInit();
+
+    if (_audioEngineImpl)
+    {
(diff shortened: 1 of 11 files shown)


zilongshanren:
@WenhaiLin  Where is the issue ID?

cocos/audio/AudioEngine.cpp
```diff
@@ -82,6 +185,18 @@ bool AudioEngine::lazyInit()
         }
     }
 
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WIN32
+    if (_audioEngineImpl && s_threadPool == nullptr)
+    {
+        s_threadPool = new (std::nothrow) AudioEngineThreadPool(true);
```
dumganhar, line 191: @WenhaiLin, @zilongshanren 
Why should we `detach` threads only for `Windows` platform?

PR #12766, opened 2015-07-09
Set default 3d animation quality to low

super626:
Set default 3d animation quality to low, which means do not interpolate between key frame, select the nearest instead.

It gets better performance.


super626:
[ci rebuild]

pandamicro:
Review done, tested

PR #12768, opened 2015-07-09
Maging the OpenGL context current before rendering

elnormous:


zilongshanren:
@elvman  What problem does this patch want to resolve? Could you provide more information?
Thanks.

elnormous:
On OS X, one process can have multiple OpenGL contexts. I am using cocos2d-x together with OS X native GUI. Everything is fine until I open NSOpenPanel, which uses OpenGL. The Open Panel immediately changes the context of current process to its context and cocos2d-x crashes.
1. GLFW was not aware of that, and I was not able to switch back to cocos2d-x’s context, so I created a pull request for them to fix it. Now the pull request is merged into GLFW 3.1.2 and glfwMakeContextCurrent now works correctly.
2. cocos2d-x must call glfwMakeContextCurrent every frame before rendering, to be sure, that current context is the right one. My pull request contains _openGLView->makeContextCurrent(); that is being called every frame before rendering.

PS. Is it possible to get write access to cocos2d-x repository? I am constantly making fixes to cocos2d-x, but it takes too much time to get them in the repository. Maybe I can send you an e-mail, whenever I come up with a new fix or improvement.

PR #12777, opened 2015-07-10
if has normal use the shader with normal

super626:
#10692


pandamicro:
Test case migrated in #12790

Issue #9586, opened 2014-12-22
FileUtils::isDirectoryExist not correct on Android when using relative path in 'assets'


tankery:
`FileUtils::isDirectoryExist` use sys/stat to check whether directory exist, witch is not work when check relative directory in 'assets', like 'assets/res-hdpi'.

On Android, we can use `AAssetManager_openDir` & `AAssetDir_getNextFileName` to check directory in 'assets'.

Here is the code for my fix (override isDirectoryExistInternal in platform android):

``` cpp
bool FileUtilsAndroid::isDirectoryExistInternal(const std::string& dirPath) const
{
    if (dirPath.empty())
    {
        return false;
    }

    bool bFound = false;

    // Check whether file exists in apk.
    if (dirPath[0] != '/')
    {
        const char* s = dirPath.c_str();

        // Found "assets/" at the beginning of the path and we don't want it
        if (dirPath.find(_defaultResRootPath) == 0) s += strlen("assets/");

        if (FileUtilsAndroid::assetmanager) {
            AAssetDir* aa = AAssetManager_openDir(FileUtilsAndroid::assetmanager, s);
            if (aa && AAssetDir_getNextFileName(aa))
            {
                bFound = true;
                AAssetDir_close(aa);
            } else {
                // CCLOG("[AssetManager] ... in APK %s, found = false!", dirPath.c_str());
            }
        }
    }
    else
    {
        struct stat st;
        if (stat(dirPath.c_str(), &st) == 0)
        {
            bFound = S_ISDIR(st.st_mode);
        }
    }
    return bFound;
}
```

minggo:
@tankery 
Why not send a pull request?

PR #12778, opened 2015-07-10
Fix bug: FileUtils::isDirectoryExist not correct on Android when using relative path in 'assets'

yangws:
related issue:
https://github.com/cocos2d/cocos2d-x/issues/9586


pandamicro:
Please use 4 spaces instead of tab

PR #12780, opened 2015-07-10
Updated AudioEngine to enable setting pitch of a playing sound file.

jimrange:
Updated AudioEngine to enable setting pitch of a playing effect. Updated applied to iOS, Mac, Android, WinRT and Win32.

Valid pitch range is [0.5,2.0]. 

Setting pitch on all platforms affects the playback speed and effective pitch of the audio file.

Demo Video: http://youtu.be/4iX9gct5kvM

This is a great feature to simulate a motor sound, or many other effects.

The pitch/rate changing behavior, as designed and tested, seems to be identical across Mac, iOS, Android, WinRT and Win32. 

Tested on Macbook Pro, iOS iPhone 5 device, Android 4.4.4 Nexus 7 2012, WinRT Nokia Lumina 928W and Win32 simulator.

The test project that I used is located here. It is mostly the same code from the New Audio Engine test case. https://github.com/jimrange/AudioEnginePitchTest

### DIFF
--- a/cocos/audio/AudioEngine.cpp
+++ b/cocos/audio/AudioEngine.cpp
@@ -181,6 +181,24 @@ void AudioEngine::setVolume(int audioID, float volume)
     }
 }
 
+void AudioEngine::setPitch(int audioID, float pitch)
+{
+    auto it = _audioIDInfoMap.find(audioID);
+    if (it != _audioIDInfoMap.end()){
+        if (pitch < 0.5f) {
+            pitch = 0.5f;
+        }
+        else if (pitch > 2.0f){
+            pitch = 2.0f;
+        }
+        
+        if (it->second.pitch != pitch){
+            _audioEngineImpl->setPitch(audioID, pitch);
+            it->second.pitch = pitch;
+        }
+    }
+}
+
 void AudioEngine::pause(int audioID)
 {
     auto it = _audioIDInfoMap.find(audioID);
--- a/cocos/audio/android/AudioEngine-inl.cpp
+++ b/cocos/audio/android/AudioEngine-inl.cpp
@@ -161,7 +161,7 @@ bool AudioPlayer::init(SLEngineItf engineEngine, SLObjectItf outputMixObject,con
         SLuint32 numSupportedInterfaces = 0;
         result = slQueryNumSupportedEngineInterfaces(&numSupportedInterfaces);
         if (numSupportedInterfaces < OPEN_SL_REQUIRED_INTERFACE_COUNT) {
-            ERRORLOG("Warning: AudioEngine OpenSL numSupportedInterfaces:%d is less than OPEN_SL_REQUIRED_INTERFACE_COUNT:%d",
+            log("Warning: AudioEngine OpenSL numSupportedInterfaces:%d is less than OPEN_SL_REQUIRED_INTERFACE_COUNT:%d",
                 numSupportedInterfaces, OPEN_SL_REQUIRED_INTERFACE_COUNT);
         }
         
@@ -349,12 +349,12 @@ void AudioEngineImpl::setPitch(int audioID,float pitch)
     
     if (playbackRate < player._minRate)
     {
-        ERRORLOG("Warning: AudioEngine attempting to set rate:%d which is lower than minRate=%d.", playbackRate, player._minRate);
+        log("Warning: AudioEngine attempting to set rate:%d which is lower than minRate=%d.", playbackRate, player._minRate);
         playbackRate = player._minRate;
     }
     else if (playbackRate > player._maxRate)
     {
-        ERRORLOG("Warning: AudioEngine attempting to set rate:%d which is higher than maxRate=%d.", playbackRate, player._maxRate);
+        log("Warning: AudioEngine attempting to set rate:%d which is higher than maxRate=%d.", playbackRate, player._maxRate);
         playbackRate = player._maxRate;
     }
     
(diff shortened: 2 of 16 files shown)


WenhaiLin:
@jimrange Great! Thank you very much!

jimrange:
I am using NDK r10e.

On Android 5.0 and 5.1.1 on a Nexus 7 2013 and on Android 5.0 on a Nexus 9, the setup of OpenSLES for setting the pitch completes without errors. There are enough OpenSLES interfaces available. The call to setRate returns success. The call to get the pitch range available to set returns [500,2000] as expected. Everything seems like it should work.

So there does not seem to be any setup or errors with calling setRate.
But for some reason the calls to setRate on Android 5 do not have any effect. The playback rate does not change.

On Android 4.4.4 the setRate works as expected.

I read the OpenSLES documentation located here, but not sure why this is not working on Android 5. $NDK/docs/Additional_library_docs/opensles/OpenSL_ES_Specification_1.0.1.pdf and index.html.

jimrange:
Also tried using NDK r10d. The Android 5 issue described above still exists when using NDK r10d.

WenhaiLin:
Hi @jimrange, you can try to set the current rate property constraints to SL_RATEPROP_PITCHCORAUDIO.

``` cpp
auto result = (*fdPlayerRate)->SetPropertyConstraints(fdPlayerRate, SL_RATEPROP_PITCHCORAUDIO);
if (SL_RESULT_PARAMETER_INVALID == result) {
    LOGD("Parameter Invalid");
}
if (SL_RESULT_FEATURE_UNSUPPORTED == result) {
    LOGD("Feature Unsupported");
}
if (SL_RESULT_SUCCESS == result) {
    LOGD("Success");
}
```

jimrange:
Yes, I have tried that and SL_RATEPROP_PITCHCORAUDIO is not available Android 5.  However, the default for SL_RATEPROP in Android 4.4.4 and Android 5.x is SL_RATEPROP_NOPITCHCORAUDIO, and the code in this pull request for setting pitch works in 4.4.4.

When pitch setting works ok on Android 4.4.4, the SL_RATEPROP is set to SL_RATEPROP_NOPITCHCORAUDIO. You can inspect the GetRateRange capabilities to see that it is set to 0x00000400 by debug logging with a %X (see page 357 of OpenSL_ES_Specification_1.0.1.pdf).

The SL_RATEPROP_NOPITCHCORAUDIO is what is needed to match the behavior of iOS, Windows, and Mac. 

**From the OpenSL_ES_Specification_1.0.1.pdf:**
SL_RATEPROP_NOPITCHCORAUDIO - Plays audio at the current rate, but without pitch correction.
SL_RATEPROP_PITCHCORAUDIO - Plays audio at the current rate, but with pitch correction.

If SL_RATEPROP_PITCHCORAUDIO were enabled, then there would be pitch **correction** for the faster or slower playback rate. We do not want pitch correction, because iOS, Mac, and Windows do not have pitch correction enabled in the implementation I am submitting.

Unfortunately, I have spent about 6 hours reading everything available that I can find about this. It seems there might be a but in 5.x where setRate is just not implemented. I have also tried doing dynamic initialization of the SLInterfaceID SL_IID_PLAYBACKRATE interface. That too did not work.

Calling the setRate method returns success in Android 4.4.4 and Android 5.x...

9.2.38
SL_RATEPROP
#define SL_RATEPROP_RESERVED1 ((SLuint32) 0x00000001)
#define SL_RATEPROP_RESERVED2 ((SLuint32) 0x00000002)
#define SL_RATEPROP_SILENTAUDIO ((SLuint32) 0x00000100)
#define SL_RATEPROP_STAGGEREDAUDIO ((SLuint32) 0x00000200)
#define SL_RATEPROP_NOPITCHCORAUDIO ((SLuint32) 0x00000400)
#define SL_RATEPROP_PITCHCORAUDIO ((SLuint32) 0x00000800)

jimrange:
Here is an open question I have regarding finding a fix for the Android 5.x set pitch: http://stackoverflow.com/questions/31527584/opensles-slplaybackrateitf-setrate-not-working-on-android-5-x-but-works-on-andr

WenhaiLin:
Hi @jimrange , I have squashed your commits and resolve conflicts.
Here is the new PR: https://github.com/cocos2d/cocos2d-x/pull/13662

WenhaiLin:
@jimrange `setRate` on Android 5 do not have any effect is likely the min/max rate is locked at 1000. I found a discussion about this problem at http://grokbase.com/t/gg/android-ndk/121naq22vj/opensl-in-ics-does-not-support-playbackrate-for-uriplayers

jimrange:
Looks like a similar problem existed before on Ice Cream Sandwich.
But setRate works fine on the newer versions of Jelly Bean and KitKat.
And now it doesn't work on Lollipop.

This problem seems like it might be different though, because the OpenSL API getters in my code return the value that the rate range is 500 to 2000, implying that there is no lock at 1000.

I also get SL_RESULT_SUCCESS returned from setRate, indicating that maybe something else is wrong.

Unfortunately, I did not get any new information from the discussion at grokbase.com.

It seems like the only possibility is to not use OpenSL for setRate, and instead modify the rate with some other library. But I have not found an open source option yet.

I also opened this question on Stack Overflow.
http://stackoverflow.com/questions/31527584/opensles-slplaybackrateitf-setrate-not-working-on-android-5-x-but-works-on-andr

pandamicro:
As the Android 5 issue, we can't merge it for v3.9, push it to future versions

jimrange:
I added a bug report to the Android project for OpenSLES SLPlaybackRateItf SetRate not working properly: https://code.google.com/p/android/issues/detail?id=186773&thanks=186773&ts=1442506284

zilongshanren:
@jimrange 
Since @WenhaiLin  has opened a new PR in #13662
let's move the discuss there.

PR #12782, opened 2015-07-10
UISliderReader and component update

xiaofeng11:
Fix percentage setting won't effect when UISlider's background resource set to null


pandamicro:
@VisualSJ Please review

VisualSJ:
@pandamicro This pr can be merged

PR #12786, opened 2015-07-10
Add Animate:getCurrentFrameIndex

WenhaiLin:


pandamicro:
I think instead of giving the SpriteFrame reference, we should give user the frame id which will be more useful to identify the progression of the Animation, you can refer to the discussion in the original issue

WenhaiLin:
@pandamicro I agree with you.

WenhaiLin:
@pandamicro I agree with you.

pandamicro:
@WenhaiLin Are you interested in adding the same functionality into the web engine ?

PR #12790, opened 2015-07-12
Use shader with normal while creating mesh with normals

pandamicro:
The solution was proposed originally in https://github.com/cocos2d/cocos2d-x/pull/10692 by @babcca


pandamicro:
[ci rebuild]

Issue #12145, opened 2015-06-02
formarRenderers in RichText doesn’t update container size


sbrednikhin:
need to update container size in RichText

http://discuss.cocos2d-x.org/t/formarrenderers-in-richtext-doesnt-update-container-size/21710

pandamicro:
Related PR: https://github.com/cocos2d/cocos2d-x/pull/12806

pekayatt:
This has been solved for JS as well? Having the same problem here in the HTML5...

PR #12806, opened 2015-07-13
Automatically customize ContentSize for RichText. 

ArcherPeng:
- Added four methods 
  void setAutoReSize(bool,bool = false)
  bool isAutoReSizeBothWidthAndHeight() const
  bool isAutoReSizeWidth()
  bool isAutoReSizeHeight()
  (Please help me will this some function change a name! thx!)
- Automatically adjust row height of ContextSize; automatically adjust row height and column width of a single row. 
- Satisfy developer’s needs. Automatically size adjustment is a convenient function for easily aligning text.


pandamicro:
This have been fixed in #12145 
Thanks for your contribution !

Issue #11447, opened 2015-04-15
Cocos2D-X 3.5 utils::captureScreen gets black screenshot when using Camera.


bryanwong:
I use 3 cameras in my game, and all children of the scene have their own camera set.

I get black screenshot when using utils::captureScreen to catch screen and share.

Please fix.

Thanks,
Bryan

walzer:
Duplicated with issue #11310

fusijie:
It fixed with https://github.com/cocos2d/cocos2d-x/pull/12810.

PR #12810, opened 2015-07-13
fix multiple camera capture screen bug.

### LINKED ISSUE #11310: multiple camera case utils::captureScreen bug.
If i have two or more cameras, the first camera draws nothing(use camera flag control this), but it is visited first in the scene, under this condition use utils::captureScreen function will get nothing, just a black image.

fusijie:
To fix https://github.com/cocos2d/cocos2d-x/issues/11310 and https://github.com/cocos2d/cocos2d-x/issues/11447


pandamicro:
@dabingnn Please review

fusijie:
Please notice that I add a new API for capture screen:

``` cpp
void Scene::setCaptureScreenCommand(CustomCommand* captureScreenCommand)
```

yangws:
Fix this issue don't need add new interface in Scene, just modify captureScreen implement: do capture screen command when Director::EVENT_AFTER_DRAW.
I'll post a request for this issue later.
The new API label could be remove.

PR #12812, opened 2015-07-13
Refactor scroll view bar

neokim:
Refactor scroll view and scroll bar
- Added methods to manipulate scroll bar's opacity
- Refined scroll event dispatching.
  - Previously, scroll events were dispatched BEFORE the container moved. It caused that event listener has no way to know how much scrolled and where the inner container is now. It is changed to be dispatched AFTER the container moved.
  - Added an event which informs the inner container is moved. It is dispatched whenever the inner container's position is changed even without scroll like jump or direct changing.


zilongshanren:
@neokim 
Thanks.  :+1:

neokim:
@zilongshanren Corrected the typo.

PR #12832, opened 2015-07-14
Fix jsb loader

pandamicro:
Image loader is asynchronous while other loaders are all synchronous. This fix is to ensure the loading process reliable


VisualSJ:
this pr is ok for merge

PR #12841, opened 2015-07-14
fix Sprite::setColor() not working with AutoPolygon

zawasp:
When setting the color on a Sprite initialized with a PolygonInfo, the internal Sprite _polyInfo is not updated from the _quad. Therefore, it has no effect when the sprite is drawn.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -948,6 +948,8 @@ void Sprite::updateColor(void)
     _quad.tl.colors = color4;
     _quad.tr.colors = color4;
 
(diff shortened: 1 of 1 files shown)


fusijie:
@zawasp  Maybe it's not the right solution. because it will cause polygon sprite turn to normal sprite. I'll solve it. Thanks.

fusijie:
I have fixed it in https://github.com/cocos2d/cocos2d-x/pull/12863.

@pandamicro @zawasp please close this pr. Thank you.

zawasp:
Thank you.

PR #12849, opened 2015-07-14
EditBox now calls EditBoxDelegate::editBoxReturn only when return button is pressed on iOS and OS X

musikov:
Right now, the `editBoxEditingDidEnd` and `editBoxReturn` methods are called simultaneously.
This pull request makes `editBoxReturn` be called only after `return` button is pressed on iOS and OS X.


zilongshanren:
@musikov 
Now Android and iOS share the same implementation, they both inherit from `Editbox-common`, so we can't accept this PR now. 

If anyone who want to implement this feature, if should work for all platforms.

PR #12852, opened 2015-07-15
fix #10242, reload shader when light number changed

### LINKED ISSUE #10242: loadConfigFile() should be called earlier than the default shader's initialization
For example, when we modify the light' number from 1 to 3 in the file of "config-example.plist", we also just have one light, because "loadConfigFile()" is called later than "GLProgramCache::loadDefaultGLProgram()" , so the light's number in the shader is still one.

super626:
fix #10242

### DIFF
--- a/cocos/base/CCConfiguration.cpp
+++ b/cocos/base/CCConfiguration.cpp
@@ -26,6 +26,7 @@ THE SOFTWARE.
 
 #include "base/CCConfiguration.h"
 #include "platform/CCFileUtils.h"
+#include "renderer/CCGLProgramCache.h"
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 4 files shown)


yangws:
@pandamicro @super626 
This pr is OK for this issue.
But i have another idea about this issue below:
How about fix this bug using event mechanism?
Add a custom event like "config reloaded", subscribe this event in glstatuscache.
Thus, the new API "GLProgramCache::isInstanceCreated" is not necessary.
In the other hand, this will meet other requirements which reaction of config file loaded.

pandamicro:
@super626 What do you think, I think event system is a good idea

super626:
@yangws 
Thanks. It is great idea.

super626:
@yangws 
@pandamicro 
Done

PR #12856, opened 2015-07-15
fixed renderTexture switch foreground to background issue.

### LINKED ISSUE #10445: RenderTexture switch foreground to background issue
How to reproduce:

Use  RenderTexture in Android and WP,  draw something and switch to background and switch back to foreground, the drawn texture will be missing. It won't rebuild.

The following code is grabbed from V4.0 latest version.

```
    // We have not found a way to dispatch the enter background message before the texture data are destroyed.
    // So we disable this pair of message handler at present.
#if 0
    CC_SAFE_DELETE(_UITextureImage);

    // to get the rendered texture data
    _UITextureImage = newImage(false);

    if (_UITextureImage)
    {
        const Size& s = _texture->getContentSizeInPixels();
        VolatileTextureMgr::addDataTexture(_texture, _UITextureImage->getData(), s.width * s.height * 4, Texture2D::PixelFormat::RGBA8888, s);

        if ( _textureCopy )
        {
            VolatileTextureMgr::addDataTexture(_textureCopy, _UITextureImage->getData(), s.width * s.height * 4, Texture2D::PixelFormat::RGBA8888, s);
        }
    }
    else
    {
        CCLOG("Cache rendertexture failed!");
    }

    glDeleteFramebuffers(1, &_FBO);
    _FBO = 0;
#endif
#endif
```

fusijie:
To fix https://github.com/cocos2d/cocos2d-x/issues/10445.

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -35,6 +35,7 @@ THE SOFTWARE.
 #include "base/CCEventDispatcher.h"
 #include "renderer/CCRenderer.h"
 #include "2d/CCCamera.h"
+#include "renderer/CCTextureCache.h"
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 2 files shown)


zilongshanren:
@fusijie  :+1:  good

PR #12860, opened 2015-07-15
fixed Scale9Sprite gray state issue.

zilongshanren:
When call `setCapInsets`, the Scale9Sprite's inner sprite will be
rebuild and it's state will be lose.

This commit reset the shader state.


pandamicro:
Review ok, thanks

PR #12864, opened 2015-07-15
add setCurPageIndex API to PageView.

zilongshanren:
1. This commit also fix the default font size issue in ui::Button.


zilongshanren:
@pandamicro  @WenhaiLin  please review this PR.

WenhaiLin:
@zilongshanren I think `setCurPageIndex` is better than `jumpToPage`,since PageView has `getCurPageIndex` API.

PR #12865, opened 2015-07-15
Fixed #11073 : mutiTouch trigger touches always only one size.

### LINKED ISSUE #11073: mutiTouch trigger touches always only one size
cpp-tests mutitouchTest sample. 
I add the test code 
image

But the touches' size always one.

engine version:Cocos2d-x 3.4
platform:android4.0 
device:xt885

the test codes in onTouchesBegan:

```
char strTemp[1024] = "";
sprintf(strTemp, "%s\ntouch length:%d", this->_label->getString().c_str(), touches.size());
this->_label->setString(strTemp);
```

yangws:
related issue: #11073

Caution: this pr will cause the touch event response a little delay on android.
Please discuss this issue, ensure to fix it is necessary.
@pandamicro

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
@@ -33,6 +33,8 @@ import android.view.KeyEvent;
 import android.view.MotionEvent;
 import android.view.inputmethod.InputMethodManager;
 
+import java.util.ArrayList;
+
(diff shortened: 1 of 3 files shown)


pandamicro:
@dumganhar @WenhaiLin Please review this PR.
It makes Android multi touch behavior same as iOS, but as @yangws said, it caused a little delay on Android, and the delay may be device related.

WenhaiLin:
@yangws I think you should flush touch event at each draw frame too. Please try to reuse `TouchInfo` object, and consider reducing the number of new object if possible.

PR #12872, opened 2015-07-16
remove EMSCRIPTEN related code

zilongshanren:


pandamicro:
Thanks, I think it's ok to remove EMSCRIPTEN support for v3, as it's not maintained since the first version of v3. We also merged web engine into Cocos2d-x, so for web platform, developers can use the web engine and our JS API

PR #12878, opened 2015-07-16
Fixed #11073 : mutiTouch trigger touches always only one size.

### LINKED ISSUE #11073: mutiTouch trigger touches always only one size

yangws:
related issue: #11073

Caution: this pr will cause the touch down event response a little delay(16ms more or less) on android.
Please discuss this issue, ensure to fix it is necessary.
@pandamicro @dumganhar @WenhaiLin

The previous pr #12865 has been closed, this pr used new buffer logic.

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
@@ -33,6 +33,10 @@ import android.view.KeyEvent;
 import android.view.MotionEvent;
 import android.view.inputmethod.InputMethodManager;
 
+import java.util.ArrayList;
+import java.util.LinkedList;
+
+@SuppressWarnings("unused")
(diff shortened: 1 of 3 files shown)


zilongshanren:
@yangws  If the touch down event response will be slowdown, I think we can't merge this PR. 
In some action games, the touch response time is very important.

@ricardoquesada  @pandamicro  Thoughts?

ricardoquesada:
Yes, I agree with @zilongshanren

pandamicro:
Ok, I will close this issue, in case user need it, they can find it in the issue's discussion

PR #12908, opened 2015-07-19
Fixing crash in Allocator if there are no allocated pages

milos1290:
This issue happens when _pages are null, trying to get item from it will cause a crash.


milos1290:
@mannewalis Can you check this out ?

PR #12910, opened 2015-07-19
Adding xxtea to ios/mac project and Android.mk file

milos1290:
This commit adds xxtea to be built in ios and mac project, and also in Android.mk file.

### DIFF
--- a/build/cocos2d_libs.xcodeproj/project.pbxproj
+++ b/build/cocos2d_libs.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


zilongshanren:
@milos1290  
Hi, thanks for the patch. But xxtea is used for Lua script encryption and it has already been added in the lua-bindings/proj.android/Android.mk.

milos1290:
Yeah, but it is not added for C++ for ios/mac and android project. If i want to use it i have to manually add it to xcode and to android.mk file.

zilongshanren:
@milos1290  xxtea is an optional component, not everyone need it, right?

minggo:
Yep, developers should add it themselves if they need it since xxtea is added for lua projects.

PR #12915, opened 2015-07-20
Fixed #12887 : [Android] Cocos2dxHelper will not be initialized after…

### LINKED ISSUE #12887: [Android] Cocos2dxHelper will not be initialized after activity recreate
Sometimes the devices will recreate the activity after application enter foreground, like too heavy memory load or on specified process policy.
You can use the development tools in Android settings to simulate it. In Settings > Developer options > Apps > Don't keep activities[turn on]

In this case, since the application go in background, the main activity will be destroyed, meanwhile the Cocos2dxHelper.sInited is still marked as "initialized" what will make Cocos2dxHelper not be initialized when application enter foreground.  This will lead to at least following issue:
Cocos2dxHelper.showDialog/Cocos2dxHelper.showEditTextDialog/Cocos2dxHelper.setEditTextDialogResult will not work.

After some tracking for Cocos2dxHelper source code, I found the "only initialize once" behavior is added for fixing following issue:http://cocos2d-x.org/issues/3723, but I cannot reproduce #3723 issue after removing the "only initialize once" logic. 

Could you check it?

yangws:
… activity recreate.

related issue : #12887

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
@@ -91,10 +91,10 @@ public class Cocos2dxHelper {
 
     private static boolean sInited = false;
     public static void init(final Activity activity) {
+        sActivity = activity;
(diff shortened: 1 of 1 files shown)


WenhaiLin:
@pandamicro It's OK for merge. @yangws Thank you!

PR #12921, opened 2015-07-20
Fixed #9382 : ui::ListView->getCurSelectedIndex() is out of range.

### LINKED ISSUE #9382: ui::ListView->getCurSelectedIndex() is out of range
1. create ListView
2. add 10 items
3. touch 10th item
4. remove all items
5. add 5 items
6. listView->getCurSelectedIndex() // output 9 which is out of range

yangws:
#9382

### DIFF
--- a/cocos/ui/UIListView.cpp
+++ b/cocos/ui/UIListView.cpp
@@ -35,7 +35,7 @@ ListView::ListView():
 _model(nullptr),
 _gravity(Gravity::CENTER_VERTICAL),
 _itemsMargin(0.0f),
-_curSelectedIndex(0),
+_curSelectedIndex(-1),
 _refreshViewDirty(true),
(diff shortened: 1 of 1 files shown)


zilongshanren:
@yangws  
There are still three functions `addChild(node*,int,int)` and `addChild(node,int,string)` and `insertDefaultItem` which would cause the `curSelectedIndex` to be invalid. 

So please also change them.

BTW: could you provide a test case for this fix? Thanks.

yangws:
@zilongshanren it's impossible to add curSelectedIndex logic in addChildXXX, because item index don't pass to these functions. For add item must call insertXXXItem functions.
insertDefaultItem has been modified yet.
The modification logic is simple, review is sufficent.

zilongshanren:
@yangws  
ok . thanks.

PR #12889, opened 2015-07-17
remove CCClippingRectangleNode transform error support.

fusijie:
To fix https://github.com/cocos2d/cocos2d-x/issues/10229.

PR #12928, opened 2015-07-20
Fix ClippingRectangleNode scissor positioning

alexpeak:
Currently, `onBeforeVisitScissor` function scales BOTH the position and
size by the nodes world scale.  The position, however, is already in
world coordinates since it is passed through `convertToWorldSpace`. This
change ensures that a rectangle in the local space of the node which
is passed into `setClippingRegion` is correctly converted to GL coords
before it is passed to `glScissor`.

### DIFF
--- a/cocos/2d/CCClippingRectangleNode.cpp
+++ b/cocos/2d/CCClippingRectangleNode.cpp
@@ -42,11 +42,21 @@ void ClippingRectangleNode::onBeforeVisitScissor()
     if (_clippingEnabled) {
         glEnable(GL_SCISSOR_TEST);
 
+        float scaleX = _scaleX;
+        float scaleY = _scaleY;
+        Node *parent = this->getParent();
+        while (parent) {
+            scaleX *= parent->getScaleX();
+            scaleY *= parent->getScaleY();
+            parent = parent->getParent();
+        }
+
+        const Point pos = convertToWorldSpace(Point(_clippingRegion.origin.x, _clippingRegion.origin.y));
         GLView* glView = Director::getInstance()->getOpenGLView();
-        glView->setScissorInPoints(_clippingRegion.origin.x,
-                                   _clippingRegion.origin.y,
-                                   _clippingRegion.size.width,
-                                   _clippingRegion.size.height);
+        glView->setScissorInPoints(pos.x,
+                                   pos.y,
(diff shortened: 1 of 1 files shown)


fusijie:
@alexpeak, I have sent the pr https://github.com/cocos2d/cocos2d-x/pull/12889. Clipping region is base on world coordinate, so I don't think it's necessary to scale rect size. How do you think?

alexpeak:
@fusijie Indeed the region passed to `glScissor` is in world (aka GL) coordinates, but the size of the rect passed to the `ClippingRectangleNode` with `setClippingRegion` is in the local space of that node, so it does need to be scaled such that it is converted to its corresponding world coordinate scale.

alexpeak:
@fusijie I see that your pull request #12889 has been merged, but I don't think that it is the best choice to use raw screen coordinates for the node.  The original intention for the `ClippingRectangleNode` seems to be for the node's `clippingRegion` to be specified in its own, local coordinate space.  This has the major benefit that the clipping box will be moved and scaled with the nodes parent as it moves across the screen.  If the clipping region is a static screen coordinate, it then is forced to stay in the position and will ignore any updates to its parent's position and scale.  

I've updated my pull request to apply cleanly to v3 after #12889

fusijie:
@alexpeak Yep, I know what your mean. I have discussed it with my colleage, and consult from the author(@dualface) of  ClippingRectangleNode, He said that ClippingRectangleNode does not support any transform.

I will not close this issue/pr, if someone wanna to scale the clippingRegion that can refer here.

alexpeak:
Right, I see now that rotations, skews, etc. wouldn't be possible since the clipping region that is finally passed to `glscissor` must always be a simple rectangle.  Perhaps if a simple, transformable polygon interface is needed for clipping, we could make a new class which draws a CCLayer then uses `glstencil` under the hood.  

Thanks for the quick response!

fusijie:
@alexpeak Yes, It's designed for the purpose of more quickly and easily, so remove most of the functions. Thanks for your pr.

fusijie:
@alexpeak  Sorry for my thoughtless. Node shouldn't use world coordinate.
I have reverted https://github.com/cocos2d/cocos2d-x/pull/12889.

PR #12949, opened 2015-07-22
Fixed #12906 : Can't add custom member to cocos2d::Scene subclass.

### LINKED ISSUE #12906: [Mac] Can't add custom member to cocos2d::Scene subclass.
On Mac, new project crashes as soon as I add a member to my custom cocos2d::Scene subclass.
# Steps to reproduce
1. Create a new project with `cocos new rnj -p org.myself.rnj -l cpp -d .`
2. Go `cd rnj/ && cocos run -p mac`. Everything goes as expected.
3. Create a new subclass of Scene, and add a pointer member to it. E.g.

```
class Verona: public Scene {
private:
  Sprite* mercutio;
public:
  /* Init the member with anything but nullptr. */
  Verona(): mercutio((Sprite *)0xdeadbeef) {}; // 
  CREATE_FUNC(Verona);
}
```
1. In `AppDelegate.cpp` instead of `HelloWorld::createScene()`, do `Verona::create()`.
2. `cocos run -p mac` once again and the game crashes.
# Cause

The project that generates cocos2d static lib (`cocos2d_libs.xcodeproj`) contains `CC_ENABLE_BULLET_INTEGRATION=1` preprocessor macro defined. Whereas the main generated project (`rnj.xcodeproj` in the example above) does not. Thus, the version of cocos2d::Scene in the static lib has two more actual members (namely, `_physics3DWorld` and `_physics3dDebugCamera`), that compiler doesn't expect while building the main project. A pointer member declared in a cocos2d::Scene subclass overlaps `_navMesh` pointer and in `Scene::stepPhysicsAndNavigation(...)` conditional statement

```
if (_navMesh)
{
    _navMesh->update(deltaTime);
}
```

evaluates to true and calls `update()`, which eventually leads to the crash.

I'm not sure, which of the projects contains correct settings (is the bullet integration supposed to be on or off by default), but it's certain that the settings must match.

yangws:
#12906

The reason is when compile CCScene.cpp, the CC_ENABLE_BULLIET_INTERATION=1 effect the CCScene.h, but when compile the custom scene derived from scene in new game project, the CC_ENABLE_BULLIET_INTERATION not defined. This cause the Scene class layout in cocos2dlib and in derived class is different.


yangws:
[ci rebuild]

pandamicro:
@super626 Please review this PR

super626:
@yangws 
You can search CC_ENABLE_BULLIET_INTERATION in the projects. There are many codes based on this macros. The 3d physics demo maybe can not work if deleted.
Are you sure it is not defined in the new game project, I find that you delete  CC_ENABLE_BULLIET_INTERATION=1 in the template.

yangws:
@super626  If defined CC_ENABLE_BULLIET_INTERATION in project files, it seems defined to many places. I think defined it in one place is better.

super626:
@yangws, @pandamicro  
Thanks. 
No question for me.

Issue #12847, opened 2015-07-14
SpriteFrame rect offset changing on device, with textures exported from TexturePacker.


UKDeveloper99:
I encountered this problem when I started cutting the level in my game in to smaller peices. I am using a sprite atlas exported from TexturePacker. If I move my Cocos2d::Camera (orthographic) in the Y axis the rect for each sprite frame appears to be changing its offset 1 by pixel. It doesn't happen consistently every frame, but maybe every 4-5 frames. 
There is a visible offset of the spriteframe and I know the rect is changing because I can see the first line of pixels from the frame above that particular frame in the sprite atlas is getting drawn. I've tried every option in TexturePacker and I'm convinced that's not the problem. I'm 99% it's something going wrong in the Cocos2D-x engine.

It should be fairly simple to reproduce, export a texture pack (sprite atlas) with a few images. Run it on a device, I used iPhone 4S. Then move the camera every frame on the y-axis, I moved it up (+0.1f). I have also tested it on an Android device (Samsung Galaxy Tab 4 10.1) and it still happens.

It's worth mentioning the problem appears to go away if running on the iOS simulator or in Win32, hopefully that may help to fix it. 

It's also worth mentioning that the problem appears to happen rather sporadically, i.e. it has no set pattern. My first thought was floating point innacuracies in the shader or something.

UKDeveloper99:
Oh I forgot to mention, it's much easier to see visually if you set the background clear colour to red or white. It's also easier to see the problem if you disabled anti aliasing. 
sprite->getTexture()->setAliasTexParameters();

Don't misunderstand the cause however, the problem still exists with anti aliasing and no glclearcolour enabled.

UKDeveloper99:
![levelone_pack_complete](https://cloud.githubusercontent.com/assets/11443031/8672248/8721dfba-2a24-11e5-84f0-9e058f075c98.png)

Here is a sprite atlas that I made with texture packer that you can use to test, just load 2 copies of "test_50x50.png" in to a spriteframe and position them on top of each other in the y-axis, one above the other.

UKDeveloper99:
[...]

PR #12950, opened 2015-07-22
add sprite frame flash line bug test case.

fusijie:
Er...Actually it no matter SpriteFrame, normal sprite setTextureRect can cause this problem too.


fusijie:
Refer to https://github.com/cocos2d/cocos2d-x/issues/12847

PR #12972, opened 2015-07-23
Fixed #12145 : formarRenderers in RichText doesn’t update container s…

### LINKED ISSUE #12145: formarRenderers in RichText doesn’t update container size

yangws:
…ize.

related issue: #12145

### DIFF
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -115,8 +115,7 @@ bool RichElementCustomNode::init(int tag, const Color3B &color, GLubyte opacity,
 RichText::RichText():
 _formatTextDirty(true),
 _leftSpaceWidth(0.0f),
-_verticalSpace(0.0f),
-_elementRenderersContainer(nullptr)
+_verticalSpace(0.0f)
 {
     
 }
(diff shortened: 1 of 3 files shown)


pandamicro:
For fixing #12145

fusijie:
`_elementRenderersContainer` could not be removed. It's will cause error effect. e.g.

``` cpp
richtext->removeAllChildren();
```

fusijie:
That's ok, @pandamicro It's can be merged now.

springrider:
@pandamicro  sorry for bring this up after months, but I found it's still not right,   in the method "formarRenderers", and the _ignoreSize==false branch, it didn't setContentSize after calculated newContentSizeHeight. makes the height not updated. 

shouldn't `this->setContentSize(Size(newContentSizeWidth, newContentSizeHeight));` been added to that branch too?

sbrednikhin:
@pandamicro I agree with @springrider 
I need to know final height of RichText element with limited width.

alfogrillo:
As @sbrednikhin I need the final height of `RichText` with limited width containing multiple elements (to set the parent inner scroll view size). 
@pandamicro  Is this possible?

PR #12988, opened 2015-07-23
Fixed #11563: Implemented AudioEngine for Linux

### LINKED ISSUE #11563: AudioEngine should support the rest of our target platforms
So finally, `AudioEngine` can support all our target platforms.
My idea is that, after v3.7, we can refactor `SimpleAudioEngine` as a wrapper of `AudioEngine` for backward compatibility.

From users feedback, most users are very satisfied with `AudioEngine` with its performance and useful callbacks.

cesarpachon:
Issue #11563: implements the new AudioEngine Interface for Linux OS, using FMOD Audio library. adds newfiles and also include linux-platform in the existing test for AudioEngine.


ricardoquesada:
Thanks for the patch.
Quick question: what is the best way to test it?  Thanks!

WenhaiLin:
@cesarpachon Great!

cesarpachon:
@ricardoquesada the best way to test is using the existing AudioEngine test from cpp-tests.. I added linux-platform to the preprocessor directives in the required places, so now it is being included in the compilation for that platform.

WenhaiLin:
@cesarpachon Thank you very much! I squash your commits into one commit. Replace tab with spaces and remove unnecessary code.
https://github.com/cocos2d/cocos2d-x/pull/12991

zilongshanren:
@cesarpachon  Since a new PR is opened, so let's move to that PR.

PR #12991, opened 2015-07-24
AudioEngine: support Linux platform

WenhaiLin:
@cesarpachoh Thank you very much!  I squash your commits into one commit. Replace tab with spaces and remove unnecessary code.


cesarpachon:
@WenhaiLin  you are welcome! happy for contribute and promote linux platform!

ricardoquesada:
@cesarpachon 
Thanks. one more question. Does it require any special version of fmod? what is the best way to install on ubuntu ? thanks.

cesarpachon:
@ricardoquesada I used the same version of FMOD that is already included in the cocos dependencies.. the SimpleAudioEngine for linux used that, so I took that as reference to build the new implementation.

WenhaiLin:
[ci rebuild]

WenhaiLin:
@cesarpachon 
Test report:
OS:Ubuntu 14.04 LTS
Test cases: AudioEngineTests from CppTests.

1.Fail to stop all audio by invoke `AudioEngine::stopAll()`. (All audio should be stop when switch to other test case.)
2.Fail to play the same audio file many times at the same time.(AudioEngineTests::LargeAudioFileTest)
3.Perform some operations may cause crash when returns from other test case.

cesarpachon:
@WenhaiLin I am going to check that. but not sure about the steps to follow: should I fix over my current fork, updated to v3 branch,  or merge first from the fork where you mixed the original commits?

WenhaiLin:
@cesarpachon I suggest you fix issues on a new branch,for example:

```
git fetch https://github.com/WenhaiLin/cocos2d-x.git v3.8-linux-audio:v3.8-linux-audio
git checkout v3.8-linux-audio
git submodule update
```

PR #12995, opened 2015-07-24
fix blur shader compliant on win8 universal.

### LINKED ISSUE #12505: Resources/Shaders/example_Blur.fsh is not OpenGL ES 2.0 compliant
The for loop in Resources/Shaders/example_Blur.fsh is not OpenGL ES 2.0 compliant and will not compile on some platforms that enforce the requirement that control flow be limited to forward branching and to loops where the maximum number of iterations can easily be determined at compile time. 

The use of the calculated value r from a uniform value means the maximum number of loop iterations cannot be known by the shader compiler. Angle for Windows 8.1 and 10 enforces this requirement. Other platforms may skip the for loop entirely without the compiler reporting an error.

The blur shader does the following:

uniform float blurRadius;
...
float r = blurRadius;
...
for(float x = -r; x < r; x += sampleStep) // error! compiler cannot unroll this loop 

The blur shader is not a correct GLSL shader.

See Appendix A section 4: Flow Control of the GLSL spec.

https://www.khronos.org/registry/gles/specs/2.0/GLSL_ES_Specification_1.0.17.pdf

fusijie:
To fix #12505

### DIFF
--- a/tests/cpp-tests/Classes/ShaderTest/ShaderTest.cpp
+++ b/tests/cpp-tests/Classes/ShaderTest/ShaderTest.cpp
@@ -401,17 +401,25 @@ bool SpriteBlur::initWithTexture(Texture2D* texture, const Rect& rect)
 
 void SpriteBlur::initGLProgram()
 {
-    GLchar * fragSource = (GLchar*) FileUtils::getInstance()->getStringFromFile(
-                                FileUtils::getInstance()->fullPathForFilename("Shaders/example_Blur.fsh")).c_str();
-    auto program = GLProgram::createWithByteArrays(ccPositionTextureColor_noMVP_vert, fragSource);
+	GLchar * fragSource = nullptr;
+#if (CC_TARGET_PLATFORM != CC_PLATFORM_WINRT)
+    fragSource = (GLchar*) String::createWithContentsOfFile(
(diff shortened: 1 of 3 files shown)


fusijie:
@yangws please help to review.

yangws:
@pandamicro 
review ok, all seem's good.

tests/cpp-tests/Classes/ShaderTest/ShaderTest.cpp
```diff
@@ -401,17 +401,25 @@ bool SpriteBlur::initWithTexture(Texture2D* texture, const Rect& rect)
...
 {
-    GLchar * fragSource = (GLchar*) FileUtils::getInstance()->getStringFromFile(
-                                FileUtils::getInstance()->fullPathForFilename("Shaders/example_Blur.fsh")).c_str();
-    auto program = GLProgram::createWithByteArrays(ccPositionTextureColor_noMVP_vert, fragSource);
+	GLchar * fragSource = nullptr;
+#if (CC_TARGET_PLATFORM != CC_PLATFORM_WINRT)
+    fragSource = (GLchar*) String::createWithContentsOfFile(
+                                FileUtils::getInstance()->fullPathForFilename("Shaders/example_Blur.fsh").c_str())->getCString();  
+#else
+	fragSource = (GLchar*)String::createWithContentsOfFile(
```
pandamicro: Please use 4 spaces instead of tab, check all new lines please

Issue #12997, opened 2015-07-24
Bind BoneNode/SkeletonNode/SkinNode to JS and Lua and Add test cases


pandamicro:
- [x] Add JS Bindings (@pandamicro)
- [x] Add Lua Bindings (@dualface)
- [x] Add JS test case (@VisualSJ)
- [x] Add Lua test case (@dualface)

PR #12998, opened 2015-07-24
Add BoneNode/SkeletonNode/SkinNode JS bindings

pandamicro:


pandamicro:
Adding jsb for https://github.com/cocos2d/cocos2d-x/issues/12997

PR #13005, opened 2015-07-24
fix bug:use wrong _modelviewTransform in director to draw bone

geron-cn:
fix bug:use wrong _modelviewTransform in director to draw bone


geron-cn:
@zilongshanren @pandamicro   fix a bug: bone draw at the other edge when move to outside right edge.

@VisualSJ   could you please sync this to js.

pandamicro:
We should enhance our test case to ensure this problem is solved and future stability

pandamicro:
[ci rebuild] ios not pass

pandamicro:
Bug fix for https://github.com/cocos2d/cocos2d-x/issues/12982

PR #13019, opened 2015-07-25
Implement CCDirector#getScreenCenter

giginet:
We often use center coordinates  of screen. For example drawing the background image.

This PR provides the way to get screen center easily.

These methods are inspired by on Kobold2D. (It is an extension of cocos2d-iPhone.)

https://github.com/kobold2d/kobold2d/blob/master/Kobold2D/__Kobold2D__/kobold2d/cocos2d-ext/CCDirectorExtensions.m#L16


pandamicro:
We have `VisibleRect::center`

PR #13020, opened 2015-07-25
Add CREATE_SCENE_FUNC Macro

giginet:
In common way, we implement `createScene` method for `Layer` classes.

This Macro implements the method to get the specific Node as wrapped Scene automatically.

``` cpp
class YourLayer :public cocos2d::Scene
{
  public:
    bool init() override;
    CREATE_SCENE_FUNC(YourLayer);
}
```

``` cpp
auto scene = YourLayer::createScene();
Director::getIntance()->replaceScene(scene);
```

### DIFF
--- a/cocos/platform/CCPlatformMacros.h
+++ b/cocos/platform/CCPlatformMacros.h
@@ -78,6 +78,30 @@ CC_DEPRECATED_ATTRIBUTE static __TYPE__* node() \
     } \
 }
 
+/** @def CREATE_SCENE_FUNC(__TYPE__)
+ * Define a createScene function for a specific node, such as Layer.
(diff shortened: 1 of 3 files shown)


pandamicro:
I think this macro is not very appropriate in the engine, we don't want to predict user's behavior because it's not possible, what do you think ? @zilongshanren

ricardoquesada:
agreed. I don't think it is needed in the engine.

zilongshanren:
agreed + 1. We should drop Macros and use template instead.

pandamicro:
Thanks @ricardoquesada @zilongshanren 
@giginet If you think differently, please comment

PR #13021, opened 2015-07-26
Add Luabinding conversions for std::map<std::string, std::string>

dualface:


pandamicro:
[ci rebuild]

pandamicro:
For https://github.com/cocos2d/cocos2d-x/issues/12997

PR #13024, opened 2015-07-27
[LUA] Gen bindings for BoneNode and SkeletonNode

pandamicro:


pandamicro:
For https://github.com/cocos2d/cocos2d-x/issues/12997

pandamicro:
Mac fail due to slave down...

PR #13032, opened 2015-07-27
Fix issue https://github.com/cocos2d/cocos2d-x/issues/12968

xiaofeng11:


pandamicro:
@xiaofeng11  New Added API can't be bound in Lua because conversion for `Data` doesn't exist

fusijie:
@xiaofeng11 please fix the compile error. Thanks.

tests/cpp-tests/Classes/ExtensionsTest/CocoStudioActionTimelineTest/ActionTimelineTestScene.cpp
```diff
@@ -66,11 +66,11 @@ void TestActionTimeline::onEnter()
 {
     ActionTimelineBaseTest::onEnter();
 
-    Node* node = CSLoader::createNode("ActionTimeline/DemoPlayer.csb");
-    ActionTimeline* action = CSLoader::createTimeline("ActionTimeline/DemoPlayer.csb");
+    Data data = FileUtils::getInstance()->getDataFromFile("ActionTimeline/DemoPlayer.csb");
+    Node* node = CSLoader::createNodeWithDataBuffer(data);
+    ActionTimeline* action = CSLoader::createTimelineWithDataBuffer(data, "ActionTimeline/DemoPlayer.csb");
```
fusijie: The better way to create node and timeline is:  

``` cpp
Data data = FileUtils::getInstance()->getDataFromFile("ActionTimeline/DemoPlayer.csb");
Node* node = CSLoader::createNode(data);//overload.
ActionTimeline* action = CSLoader::createTimeline(data);//overload, and remove filename string.
```

So some internal methods need to be split.

xiaofeng11: @fusijie 
ActionTimelineCache use file name as key to cache ActionTimeline, so the file name must be send and can not be remove.

fusijie: IIRC, we just only support one ActionTimeline for a csb, so why not get it from data?

xiaofeng11: Let me description how CSLoader::createTimelineWithDataBuffer work first: When been called, it will check if ActionTimeline already been loaded, if it was loaded and in cache, return cached item, if it won't loaded, it will be load from data, and after loaded, the new ActionTimeline will be store in cache. The KEY to distinguish ActionTimeline in cache is the csb file name, we send file name here is provide a key for cache, but not use it to read file.

fusijie: I mean that you can get the filename from the `data`, such as: 
P-CODE

``` cpp
CSLoader::createTimelineWithDataBuffer(Data data)
{
    auto filename = getFileNameFromData(data);
    //query in cache.
    if(isInActionTimelineCache(filename))
    {
         return getFromActionTimelineCache(filename);
    }
    else
    {
        return addToActionTimelineCache(filename,another_data?);
    }
}
```

er...or is the key can be any strings no just the csb filename?

xiaofeng11: If add filename to data, we need define a new data structure and construct it before calling the function, it will increase the cost for user start using it, not that friendly. Key can be any strings, but use full path of file name is a simple, clearly and cost less way to distinguish every item. And it has been used with old version for a long time, change it may cause error for old projects.

fusijie: OK, I see. Keep the `key`  string arg.
But `createBy`, `createWith` is not recommended, replace with `create` should be better, as well as `createTimelineWithDataBuffer` may be get timeline from cache, so `WithDataBuffer` is not a good idea.

xiaofeng11: OK, function name changed, removed unclear part

pandamicro:
@fusijie Is this pr passed review ?

fusijie:
@pandamicro, It can be merged. Thanks.

fusijie:
@pandamicro, It can be merged. Thanks.

PR #13036, opened 2015-07-27
fix shader-basic testcase 'center' uniform error.

### LINKED ISSUE #12505: Resources/Shaders/example_Blur.fsh is not OpenGL ES 2.0 compliant

### LINKED ISSUE #13004: [Test Case] Shader position offset.
`reproduce:` TestCase-Shader-All shaders
`issue:` position of shaders have offsets on retina screen, such as iPhone6+, Mac retina screen.
ps: JS also have this issue.

fusijie:
To fix https://github.com/cocos2d/cocos2d-x/issues/13004.

### DIFF
--- a/tests/cpp-tests/Classes/ShaderTest/ShaderTest.cpp
+++ b/tests/cpp-tests/Classes/ShaderTest/ShaderTest.cpp
@@ -111,7 +111,10 @@ void ShaderNode::setPosition(const Vec2 &newPosition)
 {
     Node::setPosition(newPosition);
     auto position = getPosition();
-    _center = Vec2(position.x * CC_CONTENT_SCALE_FACTOR(), position.y * CC_CONTENT_SCALE_FACTOR());
+    auto frameSize = Director::getInstance()->getOpenGLView()->getFrameSize();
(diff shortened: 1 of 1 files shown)


fusijie:
@yangws please take a look :)

yangws:
All seem's OK,  please merge it.
@pandamicro

PR #13039, opened 2015-07-27
AudioEngine:support callback for preload feature.

WenhaiLin:
2.AudioEngineThreadPool:Make the number of thread be a fixed number.

### DIFF
--- a/cocos/audio/AudioEngine.cpp
+++ b/cocos/audio/AudioEngine.cpp
@@ -23,11 +23,12 @@
  ****************************************************************************/
 
 #include "platform/CCPlatformConfig.h"
-#include <condition_variable>
 
 #if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT || CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID || CC_TARGET_PLATFORM == CC_PLATFORM_IOS || CC_TARGET_PLATFORM == CC_PLATFORM_MAC || CC_TARGET_PLATFORM == CC_PLATFORM_WIN32
 
 #include "audio/include/AudioEngine.h"
+#include <condition_variable>
+#include <queue>
 #include "platform/CCFileUtils.h"
 #include "base/ccUtils.h"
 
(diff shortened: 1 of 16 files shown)


WenhaiLin:

cocos/audio/include/AudioEngine.h
```diff
@@ -284,11 +289,12 @@ class EXPORT_DLL AudioEngine
     /**
     * Preload audio file.
     * @param filePath The file path of an audio.
+    * @param callback A callback which will be called after loading is finished.
     */
-    static void preload(const std::string& filePath);
+    static void preload(const std::string& filePath, AudioPreloadCallback callback = nullptr);
```
zilongshanren: please use overload instead of default arguments, add another `static void preload(const std::string& filePath, AudioPreloadCallback callback );` is better.
For more discussion, please refer to https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#default-arguments

cocos/audio/apple/AudioEngine-inl.mm
```diff
@@ -200,10 +200,14 @@ -(void) dealloc
...
         audioCache->_fileFullPath = FileUtils::getInstance()->fullPathForFilename(filePath);
         
-        AudioEngine::addTask(std::bind(&AudioCache::readDataTask, audioCache));
+        AudioEngine::addTask(std::bind(&AudioCache::readDataTask, audioCache), callback);
     }
     else {
         audioCache = &it->second;
+        if(callback && audioCache->_alBufferReady)
+        {
+            callback(true);
```
zilongshanren: When `callback(false)` is called in iOS & Mac platform?

cocos/audio/AudioEngine.cpp
```diff
@@ -67,101 +70,98 @@ AudioEngine::AudioEngineThreadPool* AudioEngine::s_threadPool = nullptr;
...
-        _threads.reserve(_numThread);
-        _tasks.reserve(_numThread);
+        _scheduler = Director::getInstance()->getScheduler();
 
-        for (int index = 0; index < _numThread; ++index) {
-            _tasks.push_back(nullptr);
-            _threads.push_back(std::thread(std::bind(&AudioEngineThreadPool::threadFunc, this, index)));
+        for (int index = 0; index < threads; ++index)
+        {
+            _workers.push_back(std::thread(std::bind(&AudioEngineThreadPool::threadFunc, this)));
```
zilongshanren: please use `emplace_back` instead of `push_back`: http://www.cplusplus.com/reference/vector/vector/emplace_back/

cocos/audio/AudioEngine.cpp
```diff
@@ -67,101 +70,98 @@ AudioEngine::AudioEngineThreadPool* AudioEngine::s_threadPool = nullptr;
...
-            std::function<void()> task = nullptr;
-            _taskMutex.lock();
-            task = _tasks[index];
-            _taskMutex.unlock();
+        while (true) {
+            std::function<bool()> task = nullptr;
+            AudioPreloadCallback callback = nullptr;
+            {
+                std::unique_lock<std::mutex> lk(_queueMutex);
+                if (_stop)
```
zilongshanren, line 114: should be `if(stop || || _taskQueue.empty()) return;`

cocos/audio/AudioEngine.cpp
```diff
@@ -67,101 +70,98 @@ AudioEngine::AudioEngineThreadPool* AudioEngine::s_threadPool = nullptr;
...
-            {
-                _threads[_numThread].detach();
-            }
-            _numThread++;
-        }
-        _taskMutex.unlock();
-        _sleepCondition.notify_all();
+    void addTask(const std::function<bool()> &task, AudioPreloadCallback callback){
+        std::unique_lock<std::mutex> lk(_queueMutex);
+        _taskQueue.push(task);
```
zilongshanren: please use `emplace` instead of `push_back`: http://www.cplusplus.com/reference/queue/queue/emplace/

cocos/audio/apple/AudioEngine-inl.mm
WenhaiLin: `callback(false)` is called in `AudioEngineThreadPool::threadFunc` when `task()` returns false.

cocos/audio/AudioEngine.cpp
```diff
@@ -67,101 +70,98 @@ AudioEngine::AudioEngineThreadPool* AudioEngine::s_threadPool = nullptr;
...
+                {
+                    task = std::move(_taskQueue.front());
+                    _taskQueue.pop();
+
+                    callback = std::move(_callbackQueue.front());
+                    _callbackQueue.pop();
+                }
+            }
 
             if (nullptr == task)
```
zilongshanren: This if statement is redundant

cocos/audio/AudioEngine.cpp
```diff
@@ -67,101 +70,98 @@ AudioEngine::AudioEngineThreadPool* AudioEngine::s_threadPool = nullptr;
...
-    void threadFunc(int index)
+    void threadFunc()
     {
-        while (s_threadPool == this) {
-            std::function<void()> task = nullptr;
-            _taskMutex.lock();
-            task = _tasks[index];
-            _taskMutex.unlock();
+        while (true) {
+            std::function<bool()> task = nullptr;
```
zilongshanren: std::function doesn't need to assign a nullptr.

cocos/audio/AudioEngine.cpp
WenhaiLin, line 114: I don't agree with you. The thread should be `wait` when the queue of task is empty

PR #13041, opened 2015-07-27
Fixed #10386 : ui::TextField hitTest not working as expected.

### LINKED ISSUE #10386: ui::TextField hitTest not working as expected
If you do: setTouchSize and setTouchAreaEnabled(true) and have setAnchorPoint(Vec2::ANCHOR_MIDDLE_LEFT)
Then it will do the hit collision wrong. Plus why isn't the Rect function containsPoint used?

My suggested solution is replace the function code with (my code might be wrong, but it seems to work okay for me now):
bool TextField::hitTest(const Vec2 &pt)
{
    if (_useTouchArea)
    {
        Vec2 nsp = convertToNodeSpace(pt);
        Rect bb = Rect(-_touchWidth \* _anchorPoint.x, -_touchHeight \* _anchorPoint.y, _touchWidth, _touchHeight + _touchHeight \* _anchorPoint.y \* 0.5f);
        return bb.containsPoint(nsp);
    }
    else
        return Widget::hitTest(pt);
}

yangws:
related issue: #10386

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -2344,8 +2344,8 @@ bool isScreenPointInRect(const Vec2 &pt, const Camera* camera, const Mat4& w2l, 
     // second, get three points which define content plane
     //  these points define a plane P(u, w) = A + uB + wC
     Vec3 A = Vec3(rect.origin.x, rect.origin.y, 0);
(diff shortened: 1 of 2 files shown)


fusijie:
@pandamicro ok, please merge it.

PR #13048, opened 2015-07-28
Lua XMLHttpRequest getResponseHeader

galigalikun:
getResponseHeader("Server")


dumganhar:
I think this Pull Request could be merged!

minggo:
It is to improve the lua test case, right? Would you mind to fix conflict?

cocos/scripting/lua-bindings/manual/network/lua_xml_http_request.cpp
```diff
@@ -1039,8 +1039,9 @@ static int lua_cocos2dx_XMLHttpRequest_getResponseHeader(lua_State* L)
         string value = streamData.str();
         
         
-        auto iter = self->getHttpHeader().find(value);
-        if (iter != self->getHttpHeader().end())
+        map<string, string> httpHeader = self->getHttpHeader();
```
dumganhar, line 1042: `_httpHeader` member variable should be `std::unordered_map<std::string, std::string>`
and `getHttpHeader` should return `const std::unordered_map<std::string, std::string>&`.
and this line should be 

``` c++
const auto& httpHeader = self->getHttpHeader();
```

PR #13051, opened 2015-07-28
Fixed #9400 : Widget::setHighlighted does not work after setBright

### LINKED ISSUE #9400: Widget::setHighlighted does not work after setBright
# how to reproduce

``` cpp
auto btn = ui::Button::create("CloseNormal.png", "CloseSelected.png");
btn->setPosition(Vec2(origin.x + visibleSize.width/2, origin.y + visibleSize.height/5));
addChild(btn);

btn->setHighlighted(true); // this shows highlighted
btn->setBright(true); // this turns to normal
btn->setHighlighted(true); // this does nothing
```

I'm using 3.3rc1.

yangws:
related issue #9400

### DIFF
--- a/cocos/ui/UIWidget.cpp
+++ b/cocos/ui/UIWidget.cpp
@@ -587,10 +587,6 @@ bool Widget::isHighlighted() const
 
 void Widget::setHighlighted(bool hilight)
 {
-    if (hilight == _highlight)
-    {
-        return;
-    }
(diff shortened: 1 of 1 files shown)


pandamicro:
Review done, logic ok.
@yangws Can you add a small test to UI Tests to verify the effect, it will also be needed for the test

PR #13055, opened 2015-07-28
Fixed #10082 : Label and Text have different types for fontSize

### LINKED ISSUE #10082: Label and Text have different types for fontSize
Label's fontSize parameter is float
Text fontSize parameter is int

We should make these the same type. Text should probably accept float?

yangws:
related issue: #10082


pandamicro:
Although it changes the API, but I think it's ok to merge, it's backward compatible, what do you think @zilongshanren

PR #13059, opened 2015-07-28
fix bug: BoneNode 's debugdraw can not be controlled by ancestor's vi…

geron-cn:
fix bug: BoneNode 's debugdraw can not be controlled by ancestor's visible

### DIFF
--- a/cocos/editor-support/cocostudio/ActionTimeline/CCBoneNode.cpp
+++ b/cocos/editor-support/cocostudio/ActionTimeline/CCBoneNode.cpp
@@ -112,6 +112,10 @@ void BoneNode::removeChild(Node* child, bool cleanup /* = true */)
 void BoneNode::removeFromBoneList(BoneNode* bone)
 {
     _childBones.eraseObject(bone);
+    auto skeletonNode = dynamic_cast<SkeletonNode*>(bone);
+    if (bone != nullptr) // nested skeleton
+        return;
+
     bone->_rootSkeleton = nullptr;
     auto subBones = bone->getAllSubBones();
     subBones.pushBack(bone);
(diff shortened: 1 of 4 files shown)


geron-cn:
Bug fix for #13060

pandamicro:
@VisualSJ Please review it and check whether need to migrate to the web engine

geron-cn:
@VisualSJ  
1.  BoneNode.cpp 　line:116  need sync to JS。
2.  TestCase need sync.
3.  Others are batch draw bone in c++, you do not nedd sync if bone's draw has no batch.

geron-cn:
[cr rebuld]

geron-cn:
[ci rebuild]

geron-cn:
@pandamicro  could you review this: fixed #13060 and #13005, add test case for them and bone nest skeleton.

VisualSJ:
@pandamicro this pr is ok for merge

cocos/editor-support/cocostudio/ActionTimeline/CCBoneNode.cpp
```diff
@@ -112,6 +112,10 @@ void BoneNode::removeChild(Node* child, bool cleanup /* = true */)
 void BoneNode::removeFromBoneList(BoneNode* bone)
 {
     _childBones.eraseObject(bone);
+    auto skeletonNode = dynamic_cast<SkeletonNode*>(bone);
+    if (bone != nullptr) // nested skeleton
```
pandamicro: Why this check ? It can't ensure bone is nested skeleton
If bone is an autorelease object, then eraseObject only reduce its reference by 1, if the reference count reaches 1, it will only be deleted before the next loop.
If bone is retained by user, then eraseObject won't delete it neither.

So please find another way to detect whether it's a nested skeleton or not

geron-cn: than you,   fixed.

PR #13062, opened 2015-07-28
[ci skip]Add framework architecture figure to README

ZhangMenghe:
A new figure of framework architecture for Cocos2d-x v3.7


pandamicro:
@minggo @ricardoquesada What do you think about this architecture graph ?
It's from our new Cocos2d-x team member: @ZhangMenghe

ricardoquesada:
+1

ricardoquesada:
sounds good. but there is one issue with the names:
- The top label says "cocos2d-x games"
- and the two bottom labels say: "cocos2d-html5 engine" and "cocos2d-x engine".

Perhaps I would use these two names instead:
- "Cocos2d  JavaScript engine" (instead of cocos2d-html5 engine)
- "Cocos2d C++ engine" (instead of cocos2d-x engine)

minggo:
Yep, it looks good.
And i have the same confusion as Riq said.
`cocos2d-x` in the picture seems to have more meanings.

pandamicro:
"Cocos2d JavaScript engine" (instead of cocos2d-html5 engine)

I think we can use Web Engine and Native Engine to differentiate

pandamicro:
I agree with Riq's idea, thanks for the feed backs

@ZhangMenghe Can you update the graph?

PR #13063, opened 2015-07-28
Fix parser compatible for layer project file

xiaofeng11:
The new reader update for bone animation cause a layer csb file parse issue, scene embed layer will cause simulator crash


pandamicro:
@VisualSJ Please check for JS parser

pandamicro:
BTW, Why the field name have been changed again ???

xiaofeng11:
The field name hasn't changed, bone animation add some new field name and optimized parser to process these names, but forget to special GameLayer should be processed as Node, so GameLayer can not be process correctly in last version.

Issue #13003, opened 2015-07-24
JSB ccui.Widget onEnter cause infinite recursion


pandamicro:
Migrate from: https://github.com/cocos2d/cocos2d-js/issues/1789

Reproduce by:

```
    var scene = new cc.Scene();
    var layout = new ccui.Layout();
    layout.onEnter = function(){
        cc.log("--------------begin");
        ccui.Widget.prototype.onEnter.call(this);
        cc.log("--------------end");
    };
    layout.addTouchEventListener(function(){},layout);
    scene.addChild(layout);
    cc.director.runScene(scene);
```

> layout.addTouchEventListener(function(){},layout) will add a CallBackComponent;
>     After _componentContainer->onEnter(); invoked, isCalledFromScript() will become true, which will cause ScriptEngineManager::sendNodeEventToJS(this, kNodeOnEnter) invoke JS level layout.onEnter 

This may be another problem

PR #13065, opened 2015-07-28
Ensure C++ classes onEnter, onExit can be safely override in JS

pandamicro:
It may cause too much recursion issue


pandamicro:
Fix for https://github.com/cocos2d/cocos2d-x/issues/13003

jianglong0156:
review done. it's ok.

PR #13067, opened 2015-07-28
Bind UIScrollViewBar and RadioButton

pandamicro:


pandamicro:
@zilongshanren Please review

PR #13069, opened 2015-07-28
Fixed a compile error when CC_SPRITE_DEBUG_DRAW is on

dmurtagh:
There was a missing semi-colon after debugDraw(true)


pandamicro:
Good catch, thanks !

dmurtagh:
You're welcome!

PR #13073, opened 2015-07-29
Bug Fix

super626:
1. clamp asin value to [-1,1]
2. sync RenderState::StateBlock

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -402,7 +402,9 @@ void Node::updateRotation3D()
     //convert quaternion to Euler angle
     float x = _rotationQuat.x, y = _rotationQuat.y, z = _rotationQuat.z, w = _rotationQuat.w;
     _rotationX = atan2f(2.f * (w * x + y * z), 1.f - 2.f * (x * x + y * y));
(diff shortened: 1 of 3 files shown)


pandamicro:
What problem does this PR solve ?

super626:
Euler angle is NaN when update Euler angle from quaternion and asin value is not between -1 and 1 by accident.

pandamicro:
@dabingnn Please review

dabingnn:
@super626 Could you please give a case for the bug?
EG,if a quaternion q = XXXX, the Euler angle would be NaN.

super626:
(cocos2d::Quaternion) $1 = (x = 0, y = -0.707106829, z = -0, w = 0.707106829)
2.f \* (w \* y - z \* x) = $3 = -1.00000012

0.707106829 equals sqrtf(2) / 2.

super626:
@dabingnn @pandamicro

dabingnn:
Thanks, it seems that the Nan happens because of the float precision problem. We need to fix it.
Besides of this, I suggest we normalise quaternion when user call `Node::setRotationQuat()` with a non-normalised quaternion.

PR #13088, opened 2015-07-29
add ActionTime lua test

liuzhi1989gbc:


pandamicro:
Lua test case for skeleton animation https://github.com/cocos2d/cocos2d-x/issues/12997

Issue #12952, opened 2015-07-22
Support generate prebuilt libs of engine with debug mode.


natural-law:
The script in framework-compile should support generating prebuilt libs with debug mode.

PR #13118, opened 2015-07-30
[ci skip] Support generate prebuilt libs of engine with debug mode.

natural-law:
Add parameter `-m, --mode` in the script to support the function.


natural-law:
Fixed #12952

PR #13092, opened 2015-07-29
Add AudioEngine js bindings

pandamicro:

PR #13125, opened 2015-07-30
modify bindings-generator index, fix the binding error.

jianglong0156:
modify bindings-generator index, fix the binding error.


jianglong0156:
other pr already update the index in https://github.com/cocos2d/cocos2d-x/pull/13092 if the pr merge, not need to merge this pr

PR #13126, opened 2015-07-30
remove "using namespace cocos2d" from CCFrame.h

takaken1994:
Avoid error the following on xcode
- Reference to “Point” is ambiguous in MacTypes.h
- Reference to “Rect" is ambiguous in MacTypes.h


pandamicro:
[ci rebuild]

pandamicro:
Thanks for the correction !

PR #13151, opened 2015-08-01
Update AppController.h

flashjay:


zilongshanren:
@flashjay Good catch.

Issue #10778, opened 2015-03-09
PageView should support Vertical scroll


zilongshanren:
Since many developers are asking me for this feature, we might need to implemented in v3.5 and v4.x.

anjandaffo:
For Now I had modified the PageView.h and PageView.cpp file so the PageView allowing for both direction. if any one need it you can use it with replacing the cocos2d/cocos/ui/PageView.h and cocos2d/cocos/ui/PageView.cpp file with attached file in below link.

http://discuss.cocos2d-x.org/t/will-pageview-support-vertical-direction/13307/3?u=anjanjis

Thanks
Anjan

Stephen-Sachson:
After I rotated a page view by 180°, it turned into a piece of SH**. The scroll direction was totally messed up. What a time-wasting worthless volatile stuff.

Stephen-Sachson:
It's going nuts if I rotate it by 180°. Really a shoddy & cheap piece of work ehhh.

anjandaffo:
@Stephen-Sachson  If you check original PageView with horizontal direction also not worked after rotate . After rotate page change reverse. shame is happen with vertical direction. I did't worked for rotate I just make modification to support Vertical direction. I required Vertical PageView in my one of our project. I made these changes and working fine. Here I just share If any one need he can take it. If it not fulfill your requirement you can escape it.

cxfshelter:
this should be done early

PR #13152, opened 2015-08-01
add PageView vertical scroll support

zilongshanren:


pandamicro:
Fixed https://github.com/cocos2d/cocos2d-x/issues/10778

cocos/ui/UIPageView.h
```diff
@@ -76,7 +76,15 @@ class CC_GUI_DLL PageView : public Layout
     enum class TouchDirection
     {
         LEFT,
-        RIGHT
+        RIGHT,
+        UP,
```
pandamicro: Need to add these constants in JS and Lua

cocos/ui/UIPageView.h
```diff
@@ -76,7 +76,15 @@ class CC_GUI_DLL PageView : public Layout
...
+        DOWN
+    };
+    
+    enum class Direction
```
pandamicro: Need to add Direction constants in JS and Lua

cocos/ui/UIPageView.h
pandamicro, line 80: Need to add these constants in JS & Lua

cocos/ui/UIPageView.cpp
```diff
@@ -230,11 +231,23 @@ float PageView::getPositionXByIndex(ssize_t idx)const
...
 {
     Layout::onSizeChanged();
-    _rightBoundary = getContentSize().width;
+    if (_direction == Direction::HORIZONTAL)
+    {
+        _rightBoundary = getContentSize().width;
+    }
+    else
+    {
+        _rightBoundary = getContentSize().height;
```
pandamicro, line 249: It may should be renamed to _boundary for a general usage

cocos/ui/UIPageView.cpp
```diff
@@ -409,16 +469,16 @@ void PageView::doLayout()
     _doLayoutDirty = false;
 }
 
-void PageView::movePages(float offset)
+void PageView::movePages(Vec2 offset)
```
pandamicro, line 472: If the direction is Vertical, shouldn't the page view prevent Horizontal moving action ? Otherwise, it may cause some unwanted effect.

cocos/ui/UIPageView.cpp
```diff
@@ -409,16 +469,16 @@ void PageView::doLayout()
...
     {
-        page->setPosition(Vec2(page->getPosition().x + offset,
-                               page->getPosition().y));
+        Vec2 oldPosition = page->getPosition();
+        page->setPosition(oldPosition + offset);
     }
 }
 
-bool PageView::scrollPages(float touchOffset)
+bool PageView::scrollPages(Vec2 touchOffset)
```
pandamicro, line 481: Same remark as movePages

cocos/ui/UIPageView.cpp
```diff
@@ -513,28 +610,46 @@ void PageView::handleReleaseLogic(Touch *touch)
...
-        float curPageLocation = curPagePos.x;
-        float pageWidth = getContentSize().width;
-        if (!_usingCustomScrollThreshold) {
-            _customScrollThreshold = pageWidth / 2.0;
+        
+        auto contentSize = getContentSize();
+       
+        float moveBoundray = 0.0f;
+        float scrollDistance;
+        if (_direction == Direction::HORIZONTAL)
```
pandamicro, line 619: This will cause unwanted effect when user touch move pages in two direction. Because it adjust only one direction, not the other

cocos/ui/UIPageView.cpp
zilongshanren, line 249: thanks.

cocos/ui/UIPageView.cpp
zilongshanren, line 481: I guess you might need to grab the code and play with it.

cocos/ui/UIPageView.cpp
zilongshanren, line 472: I have restricted the touch to one direction at a time. In the touch moved method, if the direction is horizontal, then the the movement in y direction is ignored.
So it's ok here.

cocos/ui/UIPageView.cpp
zilongshanren, line 619: It won't, please download the code and play with it.

cocos/ui/UIPageView.h
zilongshanren, line 80: These constants are used internally, so I need to move them into private section.

zilongshanren, line 80: my mistake, I will add them. But actually developer don't need to care about it. It's a historical issue.

cocos/ui/UIPageView.cpp
zilongshanren, line 249: There are two boundaries here: leftBoundary and rightBoundary, what's your suggestion about the name for left/bottom boundary and right/up boundary?

I can't find a good name.

pandamicro, line 249: just _boundary maybe

pandamicro:
I didn't have time to really test it this morning, will give it a try in a while

pandamicro:
I saw the effect, but, the positive scroll direction in vertical is from bottom up, which may be very confusing. We read from left to right, and from top to bottom, not the inversion. I think the positive vertical scroll direction should inverted. What do you think ?

zilongshanren:
@pandamicro
You could jump to the last page and then the initial scroll direction is upside down.

pandamicro:
please notice the page numbers, they are inverted

zilongshanren:
@pandamicro 
Page numbers are assigned by developers, we could also change it.

pandamicro:
Great, review done

PR #13157, opened 2015-08-02
Sync Cocos2d-x v3.7.1

pandamicro:


pandamicro:
@WenhaiLin Please take a look at the modification of Lables

PR #13160, opened 2015-08-03
Assetsmanager bug fix

jianglong0156:
Assetsmanager bug fix，avoid assetsmanager release the downloader crash


pandamicro:
This fix is not very proper, it just avoid the problem with try catch, it didn't solve the issue

jianglong0156:
yes,you are right. Maybe we should use a more elegant way to solve it

PR #13161, opened 2015-08-03
Fix 3d particle can not show in simulator

xiaofeng11:


pandamicro:
Does this need to be synced in JS parser ?

xiaofeng11:
js code is node.setCameraMask(camMask); which use the default true value, so it won't need change.

pandamicro:
OK, thanks

PR #13162, opened 2015-08-03
Terrain: Add light map support

super626:
And add interface Terrain::setLightDirection


cocos/3d/CCTerrain.h
```diff
@@ -476,11 +479,13 @@ class CC_DLL Terrain : public Node
...
     Texture2D * _detailMapTextures[4];
     Texture2D * _alphaMap;
+    Texture2D * _lightMap;
+    Vec3 _lightDir;
     CustomCommand _customCommand;
     QuadTree * _quadRoot;
     Chunk * _chunkesArray[MAX_CHUNKES][MAX_CHUNKES];
     std::vector<TerrainVertexData> _vertices;
-    std::vector<GLushort> _indices;
+    std::vector<unsigned int> _indices;
```
dabingnn, line 493: It seems that `unsigned int index` could not work correctly on some old android device?
Could we keep using GLushort?

cocos/3d/CCTerrain.h
```diff
@@ -304,6 +304,9 @@ class CC_DLL Terrain : public Node
     friend QuadTree;
     friend Chunk;
 public:
+    void setLightMap(const std::string& fileName);
+
+    void setLightDir(const Vec3& lightDir);
```
dabingnn, line 314: What space should be used when specify Light direction? World space, view space, or any other one?

cocos/3d/CCTerrain.h
super626, line 493: This is used to compute normal, and short can be overflow when the number of trunk is huge. `_indices` is not going to be used during render.

cocos/3d/CCTerrain.h
super626, line 314: It is in the terrain's local space. Terrain usually placed at (0, 0, 0) and without rotation, so most of the time we can say it is in world space.

cocos/3d/CCTerrain.h
dabingnn, line 493: Thanks, please ignore it.

cocos/3d/CCTerrain.h
dabingnn, line 314: Could you please add some comments to make it clearer?

PR #13163, opened 2015-08-03
add IF-CASE to fix debug draw under SpriteBatchNode.

fusijie:


cocos/2d/CCSprite.cpp
```diff
@@ -430,6 +430,10 @@ void Sprite::setTextureRect(const Rect& rect, bool rotated, const Size& untrimme
 
 void Sprite::debugDraw(bool on)
 {
+    if (_batchNode) {
+        log("Sprite doesn't support denbug draw when using SpriteBatchNode");
```
PeqNP, line 434: I think you meant "debug" and not "denbug", correct?

fusijie, line 434: Sorry for my mistake! It has been fixed.

PR #13169, opened 2015-08-03
Fixes mess code while passing special emoticon from Java to C++

dumganhar:


cocos/platform/android/jni/JniHelper.cpp
```diff
@@ -262,17 +262,26 @@ namespace cocos2d {
...
         }
-        
+
         JNIEnv *env = JniHelper::getEnv();
         if (!env) {
-            return nullptr;
+            return "";
         }
 
-        const char* chars = env->GetStringUTFChars(jstr, nullptr);
```
dumganhar, line 271: `GetStringUTFChars` will not deal with `emoticon`. I think it maybe a bug of Android System.

dumganhar:
@jianglong0156 has a better implementation. Close this PR.

PR #13172, opened 2015-08-03
add emotion bug test current nothing will show

jianglong0156:
add emotion bug test  current nothing will show, because the emotion can't be translated in jsb.


dumganhar:
This test case could not test `emotion` from Java to c++.
My fix at https://github.com/cocos2d/cocos2d-x/pull/13169 resolve the problem of mess code while `jstring2string` is invoked.
Please help to add another test case for testing this issue.

pandamicro:
Android build passed

PR #13173, opened 2015-08-03
fix bug: crash when use bone after erase it form Vector.

geron-cn:
fix bug: crash when use bone after erase it form Vector. move the operation before erase, make sure bone ref\* is not realeaed


geron-cn:
[ci rebuild]

geron-cn:
@VisualSJ  Can you check this fix,  make sure whether need Sync it to JS

pandamicro:
[ci rebuild]

PR #12258, opened 2015-06-09
fix issue #8384

nocrim:
fix bug "ParticleSystem::PositionType::FREE mode fail on rotation".

PR #13178, opened 2015-08-03
fix particle system emitter rotation bug in FREE mode.

### LINKED ISSUE #8384: With ParticleSystem::PositionType::FREE mode, The emitted particle still follow the emitter position when set a rotation for particle.
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/4577). It was created at 2014/03/27 06:13:58 +0000
# Description

Use this code in DemoFlower::onEnter():
`
    ParticleDemo::onEnter();

```
_emitter = ParticleFlower::create();
_emitter->setEmissionRate(50);
_emitter->setRotation(-30);
_emitter->setAngleVar(0);
_emitter->setAngle(90);
_emitter->setRadialAccel(0);
_emitter->setTangentialAccel(0);
_emitter->retain();
_background->addChild(_emitter, 10);
_emitter->setTexture( Director::getInstance()->getTextureCache()->addImage(s_stars1) );

_background->stopAllActions();

setEmitterPosition();`
```

Move the particle by mouse/finger, the emitted particle will follow the emitter.
- boyu0 added comment:
  Since we have `ParticleSystem::setAngleVar();` and `ParticleSystem::setAngleVar();` method, the demand of rotate a  ParticleSystem is not that common.
  Fix this bug has lots of "transform" work to do, without an inclusive test,  add this fix to 3.0 is not safety, and it also has some performance lost.
  So I changed this issue to "Candidate for next 3.0 release".
  And here is the PR link for someone who urgent need this fix:
  https://github.com/cocos2d/cocos2d-x/pull/6127

fusijie:
To fix #8384.

Relative pr: https://github.com/cocos2d/cocos2d-x/pull/12258.

### DIFF
--- a/cocos/2d/CCParticleSystem.cpp
+++ b/cocos/2d/CCParticleSystem.cpp
@@ -508,7 +508,7 @@ void ParticleSystem::initParticle(tParticle* particle)
 
     particle->pos.y = _sourcePosition.y + _posVar.y * CCRANDOM_MINUS1_1();
 
-
+    particle->affineTransform = getNodeToWorldAffineTransform();
     // Color
     Color4F start;
     start.r = clampf(_startColor.r + _startColorVar.r * CCRANDOM_MINUS1_1(), 0, 1);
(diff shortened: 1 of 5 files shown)


fusijie:
@dabingnn Please take a look. Thanks.

To fix the bug, I added a member variable `AffineTransform` to sParticle, which will cause the memory of particle rising. **About 2.4K for 100 particles.**

cocos/2d/CCParticleSystem.cpp
```diff
@@ -794,7 +786,7 @@ void ParticleSystem::update(float dt)
 
                 // translate newPos to correct position, since matrix transform isn't performed in batchnode
                 // don't update the particle with the new position information, it will interfere with the radius and tangential calculations
-                if (_batchNode)
+                if (_batchNode && PositionType::FREE != _positionType)
```
dabingnn, line 789: What happened if `_batchNode` is `nullptr`?

fusijie, line 789: Because matrix transform isn't performed in batchnode, so if `_batchnode` is not nullptr, here needs to add _position. But now the particle system support auto-batch, so It's no necessary to refactory ParticleBatchNode but just only to compatible with it.

PR #13187, opened 2015-08-04
Remove file error notice label from TextBMFontReader

xiaofeng11:
Keep consistency with other components


pandamicro:
@VisualSJ Please check with JSON parser

Issue #9061, opened 2014-11-07
Polish AssetsManagerEx


pandamicro:
Based on feedbacks from developers who published their games using AssetsManagerEx in Cocos2d-JS and pushed hot updates to several hundreds K users.

Issue : 
- The met this issue : https://github.com/cocos2d/cocos2d-x/issues/9038

Important features they require : 
- Delay the timing for modify downloaded file name from .temp to avoid resources being used before real success of an update
- Need to provide a function to clear JS file cache to support JS file reload
- Bind Manifest::getAssets for assets preprocess after update finished
- When the assets manager is released, the download callbacks should be canceled

Requirements to be discussed : 
- Provide connection timeout getter setter to AssetsManagerEx
- Provide access to file size during download
- Provide a checking callback for download process, if fail, mark the download as failed
- Permit user to forbid download resume feature

pandamicro:
Update 1

Issues : 
- When user updated the app package, then local manifest may be more recent than the cached manifest, but we always use cached manifest when it exists. This behavior should be modified
- When Downloader::batchDownload receives zero units to download, it will return directly without any notifications. When this happens, AssetsManager won't know everything have been updated already, and  end up with a endless wait.

pandamicro:
Update 2

Related PR: https://github.com/cocos2d/cocos2d-x/pull/9039
- When user updated the app package, then local manifest may be more recent than the cached manifest, but we always use cached manifest when it exists. This behavior should be modified
- When Downloader::batchDownload receives zero units to download, it will return directly without any notifications. When this happens, AssetsManager won't know everything have been updated already, and end up with a endless wait.
- Bind Manifest::getAssets for assets preprocess after update finished

pandamicro:
A discussion thread have been created on the forum: http://discuss.cocos2d-x.org/t/assetsmanagerex-bugs-and-new-features-requests/18211

pandamicro:
Partially scheduled to v3.8
@jianglong0156 Please set the possible tasks for v3.8

jianglong0156:
the assetsmanager will improve these feather in v3.8
- [x] When Downloader::batchDownload receives zero units to download, it will return directly without any notifications. When this happens, AssetsManager won't know everything have been updated already, and end up with a endless wait.
- [x] When the assets manager is released, the download callbacks should be canceled
- [ ] add the option. when all files downloaded, the old file will be replaced.

pandamicro:
Future improvement will be rescheduled.

PR #13189, opened 2015-08-04
Fix AssetsManager crash by protecting Downloader with shared_ptr

pandamicro:
Fix for https://github.com/cocos2d/cocos2d-x/issues/9061


jianglong0156:
review done. it's ok. not crash happend
but when the assetsmanager exit, the curl still downloading 
image

jianglong0156:
and if start downloading again, the network response slowly

pandamicro:
Yes the downloader will act like this, otherwise it can't be finished safely and may cause crash

Issue #12292, opened 2015-06-11
UIText clone not clone the outline and shadow


flamefox:
as the title, and the label's '_outlineSize' have should be 0 in the reset function. some time it will cause  the problom #9499

zilongshanren:
thanks @flamefox 
We will test the issue.

minggo:
It is fixed in latest codes.

Issue #12943, opened 2015-07-22
Physics debugDraw does not match with nodes when Scene is transformed


hawkwood:
If a transformation is applied to the Scene, such as MoveBy/To, ScaleBy/To, or RotateBy/To,  the debugDraw layer no longer matches up with the visual nodes.

This is easiest to see by adding a runAction to PhysicsTests.cpp of cpp-tests:

> void PhysicsDemo::toggleDebug()
> {
>    runAction( RotateBy::create( 2.0, 360 ));
>    ....

You can see that the debug frames spin as expected with the non-physics elements on the scene, whereas the actual nodes (such as the sprites of grossini in Pyramid Stack test) spin as a group the wrong direction, as well as individually spinning, which they shouldn't.

qiululu:
add codes  like this:
void PhysicsDemo::toggleDebug()
{
    #if CC_USE_PHYSICS
    _debugDraw = !_debugDraw;
    getPhysicsWorld()->setDebugDrawMask(_debugDraw ? PhysicsWorld::DEBUGDRAW_ALL :   PhysicsWorld::DEBUGDRAW_NONE);
**runAction( ScaleTo::create(2020, 0.5, 0.5));**
#endif
}
to your physicsTest demo,  run the action after you set the drawMask, 
and then, at **CCScene.cpp** ,
at **void Scene::render(Renderer\* renderer)** function, after  
**director->pushMatrix(MATRIX_STACK_TYPE::MATRIX_STACK_PROJECTION);**
 add these codes:

```
Camera* visitingCamera = Camera::_visitingCamera;
        Mat4 scalProject;
        Mat4::createScale(Vec3(_scaleX,_scaleY,_scaleZ), &scalProject);
        visitingCamera->setAdditionalProjection(scalProject);
```

then you'll see
<img width="885" alt="tranphysics" src="https://cloud.githubusercontent.com/assets/5564318/9024989/732cdb2a-3921-11e5-86ab-682305140796.png">
<img width="915" alt="ps2" src="https://cloud.githubusercontent.com/assets/5564318/9025017/c3bc0830-3922-11e5-85f6-c05fca661e29.png">

I think consider adding this function to allow developer's to change the scene's shape and run action. but it has some other problems such as the touch event location ect.

qiululu:
or override the scene node's setScaleX, scaleY, scaleZ  etc. works properly ,  but still have some problem, i'll ask if it is take into consideration since it will have some other effection. @pandamicro

[...]

PR #13190, opened 2015-08-04
fix bug #12292 at CCLabel   and bug #12943 at CCScene

qiululu:
fix two bugs and at CCLabel and CCScene, test case can be seen at physicsTest demo.


pandamicro:
Thank you, here are some suggestions and questions:
1. You should not mix two bug fixes into one PR.
2. Why do we need to add APIs for fixing bug ?
3. getFontConfigInfo is a getter function, it should not modify anything, it's also a serious mistake to modify another object's content.
4. You shouldn't use public APIs `setAdditionalProjection` internally, it will prevent user from using it, and it will cause issues when user use it

qiululu:
1， for some reason , but it still have can be read easily.
2,  because it if I don't do so , it will be some modify to base class CCNode, it may be bad in design.
3, I will fix itl
4, for the concepts of the tranlation to a scene, this doing if better I think.

pandamicro:
Your solution is not very proper, and it cannot be accepted.

qiululu:
ok , wait for your suggestion. fixed 3

qiululu:
ok, thanks,  I'll think it twice

PR #13201, opened 2015-08-04
Label:support auto batch with bitmap font or char map.

WenhaiLin:


zilongshanren:
:+1:  Good job

sandfresh:
If the BM label is added as child,the auto batch will not work.
I must change the label global z order ,or the batch doesn't work.
Other people also encounter the same issue.
http://discuss.cocos2d-x.org/t/mixing-sprites-and-bitmap-fonts-on-a-single-texture-to-limit-draw-calls/26088

sandfresh:
If the sprite add child label,the label batch will only work in the sprite.
for example, if 1 sprite adds n BitMap label, the draw call will be 1.
but the the n sprites add n bm label, the draw call will be n.
Even the all bitmap using the same texture,but the batch will be merged.
I must change the label global z order ,or the batch doesn't work.

PR #13210, opened 2015-08-05
Fixed FontAtlasCache::purgeCachedData might cause crash.

WenhaiLin:


cocos/2d/CCFontAtlasCache.cpp
```diff
@@ -39,7 +39,8 @@ std::unordered_map<std::string, FontAtlas *> FontAtlasCache::_atlasMap;
 
 void FontAtlasCache::purgeCachedData()
 {
-    for (auto & atlas:_atlasMap)
+    auto atlasMapCopy = _atlasMap;
+    for (auto&& atlas : atlasMapCopy)
```
zilongshanren, line 43: Why we need a copy here?

WenhaiLin, line 43: `_atlasMap`  could be modified during iteration.

joewan, line 43: why we need two "&" here?

PR #13217, opened 2015-08-05
fix textfield and editbox can not input emoticon

jianglong0156:
fix textfield and editbox can not input emoticon and improve the api comments


jianglong0156:
relate issue:https://github.com/cocos2d/cocos2d-x/issues/13221

PR #13220, opened 2015-08-05
Improved EditBox implement for platform win32

zilongshanren:


summerinsects:
Thanks for accepting

zilongshanren:
https://github.com/cocos2d/cocos2d-x/issues/13431

WenhaiLin:
@zilongshanren +1 I think this pr is OK for merge.

WenhaiLin:
@wohaaitinciu I think you should revert to the original state when user cancel input.

zilongshanren:
@wohaaitinciu  
When user press the cancel button, all the inputed text should be removed.
Before your patch, the text in the popup window only sync to EditBox when user pressed done.

Now the sync behavior happens when user starts to type.

The patch break the back compatibility.

WenhaiLin:
@wohaaitinciu Any updates?

summerinsects:
I see

summerinsects:
So, cancel button pressed means it will give up all the edit operations in this time?

summerinsects:
For example, original text was `hello', then I do some edit operations, such as deleting, appending or modifying, now the displaying text may no longer be`hello' (see article 6), and then I press cancel button, should it revert to `hello'?

Very sorry for keeping you waiting so long
@WenhaiLin @zilongshanren

WenhaiLin:
@wohaaitinciu Yes,give up all the edit operations in this time that is the expected result. @zilongshanren  What do you think?

wohaaitinciu:
see https://github.com/zilongshanren/cocos2d-x/pull/25

WenhaiLin:
@zilongshanren I think this PR is OK for merge now. @wohaaitinciu Thank you very much!

PR #13230, opened 2015-08-05
Spaces should not be lost if label created with Fingerpop.ttf

WenhaiLin:


WenhaiLin:

PR #13232, opened 2015-08-05
Add getFileExtension to FileUtils

perminovVS:
Gets filename extension is a suffix (separated from the base filename by
a dot) in lower case.

More code need get filename extension, everyone does it differently.
use check UPPER and lower case, use . and no(example ".csb", ".CSB",
"csb" )

And bag in AudioEngineImpl: find point from left( ext =
strchr(filePath.c_str(), '.'); )
If file path contains point. always unsupported media type


WenhaiLin:
:+1:

WenhaiLin:
@zilongshanren Please merge this PR.

PR #13233, opened 2015-08-05
Use GL_GENERATE_MIPMAP for OpenGL version less 3.0

perminovVS:
glGenerateMipmap supports:
OpenGL ES 2.0 or greater
https://www.khronos.org/opengles/sdk/docs/man3/html/glGenerateMipmap.xhtml
OpenGL 3.0 or greater
https://www.opengl.org/sdk/docs/man/html/glGenerateMipmap.xhtml

Add Configuration::supportsGenerateMipmap() for check it
Add Texture2D::setAutoGenerateMipmap for use old GL_GENERATE_MIPMAP (
need set before glTexImage2D ), and use it any initWithImage

GL_GENERATE_MIPMAP  -
https://www.opengl.org/wiki/Common_Mistakes#Automatic_mipmap_generation


perminovVS:
new version #15196

PR #13251, opened 2015-08-06
fix crash when fore call remove child bone from a bone which already …

geron-cn:
fix crash when fore call remove child bone from a bone which already removed from a skeleton.


geron-cn:
@pandamicro  could you please review at this, thx.

pandamicro:
Next time, please make sure the first line of commit message less than 50 letters, otherwise it will be wrapped

geron-cn:
Yes, I will watch out this .

pandamicro:
Rearranged to https://github.com/cocos2d/cocos2d-x/pull/13256

PR #13268, opened 2015-08-07
fix engine crash because of wrong initialisation on some android device

dabingnn:
It seems that `egl.eglChooseConfig` will return true even there is no available configs.


felixjingga:
What devices can reproduce this issue @dabingnn ? and if it return null, what will happens for those devices?

pandamicro:
[ci rebuild]

PR #13270, opened 2015-08-07
[ci skip]fix the node color property can not use

jianglong0156:
code example:

```
node.color = xxxx // fix it
```


pandamicro:
Review ok

Issue #13147, opened 2015-08-01
Reimplement EditBox on Android platform


zilongshanren:
Discussion move to here: http://discuss.cocos2d-x.org/t/editbox-on-android-has-been-improved-dramatically/22977

Issue #9572, opened 2014-12-22
problem of UIEditbox::setFontSize & setPlaceholderFontSize


zhaijialong:
```
void EditBox::setFontSize(int fontSize)
{
    _fontSize = fontSize;
    if (_editBoxImpl != nullptr && _fontName.length() > 0)
    {
        _editBoxImpl->setFont(_fontName.c_str(), _fontSize);
    }
}
```

As you can see, `setFontSize` need `_fontName.length()>0` to work. So if I just want to use default font, and don't invoke `setFontName`, I can't change the font size. 
@andyque

PR #13273, opened 2015-08-07
Re-implement Editbox on Android platform.

zilongshanren:
This PR fixed issue: 
https://github.com/cocos2d/cocos2d-x/issues/13147

This PR also fixed issue:

https://github.com/cocos2d/cocos2d-x/issues/9572

For the new EditBox:
1. Added a ResizeLayout for updating layout when soft keyboard is shown.
   This is because some Android devices won't refresh the view and when
   the view is panned, the native EditText will turn black. This is really
   a hack, but it works for many Android devices from 2.3 to 5.0.
   When keyboard is beginning to show, the ResizeLayout's doLayout
   method will be called 24 frames per second. Once the keyboard is closed,
   or the user begins to input, the doLayout method will be disabled.
2. EditText is working in a different UI thread, so all the data
   are passed to GL thread via callbacks in UI thead.
3. It also overrides onKeyDown method in Cocos2dxEditBox.
   When the EditText got the focus, users could press back key to exit
   the game. This behavior is not allowed. We give developer the choice to
   handle back key event manually.
4. It also add `static Rect convertBoundingBoxToScreen(Node* node);` in ui::Helper class to remove redundant code. And this function is very useful for synchronizing native control transform with OpenGL transform.


zilongshanren:
@pandamicro  @fusijie  @WenhaiLin  
please help to review this PR, thanks.

pandamicro:
[ci rebuild]

zilongshanren:
@pandamicro  Could this PR be merged now?

pandamicro:
Yes, I think it should be merged so that QA team can test it in the next week

pandamicro:
Besides, it should be link to an issue

zilongshanren:
It fixes issue https://github.com/cocos2d/cocos2d-x/issues/13147

zilongshanren:
@pandamicro  please merge...

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    public static void setVisible(final int index, final boolean visible) {
+        mCocos2dxActivity.runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+                Cocos2dxEditBox editBox = mEditBoxArray.get(index);
+                if (editBox != null) {
+                    editBox.setVisibility(visible ? View.VISIBLE : View.GONE);
+                    if (visible) {
```
fusijie, line 274: why not?

``` java
if(_oldVisible != visible)
{
    ...do sth.
}
```

The same to other setBool.

PR #13274, opened 2015-08-07
remove unused code

super626:


pandamicro:
Clicked by mistake... There is a usage of `info` variable just after

Issue #8137, opened 2014-09-24
UIButton's APIs confusion


pandamicro:
Here are the APIs which makes developers really confused:

setEnabled
setTouchEnabled
setBright
setHighlighted

I will start with `setBright` and `setHighlighted`, `setBright` true will make the button shows normal state or highlighted state which is decided by `setHighlighted`, false will make the button shows disabled state.

So far, complicated but logically acceptable.

But the problem is that bright or not is not related to the button's enable state.
When a button is not enabled or its touch is not enabled, it's still shown in normal state unless developer set its bright state... Same case for touch enable state. Reversely, it's also very strange, I can set the bright state to false, and though it can still be pressed because it's enabled.

The concern is why all these states and why they are not related when they should be ?

As the API can't be modified, I suggest we make it more clear and more logical for developers.

zilongshanren:
@pandamicro 
You got it!  I have also thought about these weird API definitions, but failed to find a good solution here.

darkdukey:
It's always a good idea to start with a use case, what kind of developer use case does this API trying to address?

Is this the use case?

```
Developer wants to make a button highlighted. So it can attract user's attention.
```

pandamicro:
One use case is

```
Developer want the button to be shown as disabled when it's disabled
```

darkdukey:
Trying to summarize the problems here
1. `setEnabled(false)` should set button to disabled state, which I also agree, it's very natural for developer to understand.
2. `setEnabled` should also take care of enable touch, because it's frustrating to have a button shows as normal, but can't response to tap/click, so maybe we should have `setEnabled` and `setTouchEnabled` behaved the same.
3. maybe we should deprecate `setBright`, because it seems not really useful base on @pandamicro 's description, `setHighlighted` has the final say of the things, but we don't really have a highlighted state

# 

Not related
Moving forward, I think button should just have enabled and disabled state as default, but we should make button states extendable, so developers can create and maintain their own button state.

So they can do things like

``` cpp
button->addState(Action::create("popup"));
button->setState("popup");
```

minggo:
About confused API, i think we can refer to iOS/Android or other GUI framework first.

zilongshanren:
@darkdukey 
I totally agree with you. API should be clean and concise. 
And it might be good for developers adding new button state. But I don't find any use case in my mind.

The worst things are that every minor change might break the back compatibility.  

Thoughts?

PR #13289, opened 2015-08-09
Synchronize enable state and bright state for Widget

pandamicro:
Fix issue: https://github.com/cocos2d/cocos2d-x/issues/8137


pandamicro:
@zilongshanren Please review

zilongshanren:
@pandamicro 
It's OK for me, but it need to be documented in the release note.

PR #13291, opened 2015-08-10
fix shader test crash on android device

dabingnn:
The crash happens because some uniforms and shader are not restored correctly on android.


fusijie:
@pandamicro It's can be merged.

PR #13293, opened 2015-08-10
fix js-tests can't use cocos run -p web -m release

jianglong0156:
fix js-tests can't use cocos run -p web -m release


tests/js-tests/main.js
```diff
@@ -117,7 +117,15 @@ cc.game.onStart = function(){
     }
     else
     {
-        cc.loader.resPath = '../cpp-tests/Resources'
+        // js-test use cpptest resource in debug mode , and in the release mode, console will copy the resource into the res dir
+        // so the respath will modify to res,
+        if (cc.game.config[cc.game.CONFIG_KEY.engineDir] !== "frameworks/cocos2d-html5") {
```
pandamicro, line 122: You should use `cc.game.config[cc.game.CONFIG_KEY.engineDir] === undefined` or `=== ""` if it's initialized
I don't know which case it's in, but don't hard code the url, it may be changed.

tests/js-tests/src/tests_resources.js
```diff
@@ -2,9 +2,18 @@
 var s_resprefix = "";
 
 var ccbjs = "";
+// js-test use cpptest resource in debug mode , and in the release mode, console will copy the resource into the res dir
+// so the respath will modify to res,
 if (!cc.sys.isNative)
 {
-    ccbjs = "../../js-tests/resjs/";
+    if (cc.game.config[cc.game.CONFIG_KEY.engineDir] !== "frameworks/cocos2d-html5") {
```
pandamicro, line 9: Same as above

tests/js-tests/main.js
jianglong0156, line 122: default set "frameworks/cocos2d-html5" in engine init

pandamicro, line 122: I know it's this value today, but who can say next month we won't change it, so please modify it to something more general

PR #13298, opened 2015-08-10
fix vibrate resource error, and hide the vibrateTest in web

jianglong0156:
fix vigrate resource error, and hide the vibrateTest in web


pandamicro:
Review done

PR #13300, opened 2015-08-10
modify js-test mac device target 10.10 to 10.7

jianglong0156:
modify js-test mac device target 10.10 to 10.7


pandamicro:
Review ok, thanks

PR #13301, opened 2015-08-10
fix bug: #21906 Pause/Resume 's result not correct.

yangws:


WenhaiLin:
@zilongshanren It's OK for merge.

PR #13303, opened 2015-08-10
Fix issue described in #13284 by reorder state setting and event dispatching

pandamicro:
The real reason of #13284 is that the `_selected` state is set after touch events dispatching.
So I decided to override `onTouchEnded` function to set the `_selected` state.

Also fix http://punchbox.info:3000/issues/21988


pandamicro:
@zilongshanren Please review this PR, it's first priority

PR #13310, opened 2015-08-11
Solve the error when build cpp-tests for Mac/iOS with release mode.

natural-law:


zilongshanren:
@natural-law  we need to use cocos2d::log instead.

PR #13314, opened 2015-08-11
fix bug: #21904 : cpp-tests Renderer Test -> New Render Drag the laye…

yangws:
…r to test the result of culling test ckick regin error after drag.

### DIFF
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -177,7 +177,6 @@ void Scene::render(Renderer* renderer)
 {
     auto director = Director::getInstance();
     Camera* defaultCamera = nullptr;
-    const auto& transform = getNodeToParentTransform();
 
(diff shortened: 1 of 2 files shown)


dabingnn:
@pandamicro This pull request can fix the touch bug, but it break compatibility.
`Scene` could be moved by calling `SetPostion` functions Before. And if this pull request is merged, `Scene` could be moved, but it takes no effect.
My proposal is
1. Try to find a solution which has no side effects.
2. If we can not find one, we must choose one in this options: `touch bug fixed` and `Scene can be moved`.

pandamicro:
I think backward compatibility is important, move a scene seems like a behavior many user may have, so I also suggest to find another way to fix this problem

dabingnn:
The bug does not exist before Version 3.5.

zilongshanren:
@pandamicro @dabingnn Could we accept this patch?

PR #13315, opened 2015-08-11
[ci skip]Add testcase to issue3053

ZhangMenghe:
Add testcase in UIListView,UIPageView and UIScrollView, which belong to issue3053(https://github.com/cocos2d/cocos2d-html5/pull/3053)


VisualSJ:
@zilongshanren this pr is ok for merge

PR #13321, opened 2015-08-11
V2

Still4:
Fixing typos in create_project.py


zilongshanren:
v2 is not maintained now and this PR is bad.

PR #13326, opened 2015-08-12
fix bug: #22060 cpp-tests Scheduler -> Pause/Resume All User test cas…

yangws:


minggo:
I think we can merge this PR when jenkins passes.

PR #13335, opened 2015-08-12
Upgrade Spine runtime

pandamicro:
from https://github.com/EsotericSoftware/spine-runtimes/tree/56bee5ccd2aeba8897e12f4a526bb04fc6f58cf3


minggo:
ios slave was offline, so it failed.
But all other platforms passed, so i will merge this PR.

PR #13339, opened 2015-08-12
updating AudioEngine implementation for linux

cesarpachon:
updating AudioEngine linux implementation with latests changes in AudioEngine interface (a new callback was added) and implementing pending stopAll methods. also, refactor internal data structure to support requirement of same sound being played many times simultaneously


zilongshanren:
@WenhaiLin  please take a look at this PR.

PR #13349, opened 2015-08-13
Fix bug: timer dt time error when interval = 0.

yangws:


WenhaiLin:
@zilongshanren It's OK for merge.

PR #13364, opened 2015-08-14
Fixed `Scheduler::update(float dt)` can lead to an infinite loop when the value of `dt` is zero.

WenhaiLin:
Test Case:Cpp Tests-->Actions - Basic


WenhaiLin:
@yangws Please review this PR.

PR #13380, opened 2015-08-14
[ci skip]fixed js shaderNode center in retina mode.

fusijie:
@ZhangMenghe Please take a look. Thanks.


fusijie:
@ZhangMenghe Please take this issue to fix both native and web.

PR #13393, opened 2015-08-15
Gyroscope

ku-kueihsi:
I enabled gyroscope and magnetic field sensor on android by forking the existing accelerometer sensor. 

There is no test as I did not find any unit test for accelerometer. 

Some devices may no have magnetic field sensor at all. I have handled this error in this pull request.


pandamicro:
@dabingnn This may be helpful for your research ?

ku-kueihsi:
I left academic world for a while. This is for my hobby. I plan to use phone to measure distance. That is why I need gyroscope. This also benefits cocos2d-x, right? Using full motion as an input is a unique feature on mobile devices.

cocos2d-x already has accelerometer pipeline. Why not support other sensors for free?

pandamicro:
Can you clean up your branch to be based on the latest v3 ? Then we can see a clean commit history

ku-kueihsi:
Sorry for the delay. Branch cleaned up, rebased. Please check it out.

dabingnn:
@sftrytry I would suggest that we can merge all the sensor value into one struct `EventSensor`. which would be easy to extend. For compatibility, I think Acceleration sensor event could be a subclass of `EventSensor`.
What is your opinion?

ku-kueihsi:
Sounds good. I am working on that.

sftrytry:
@dabingnn 
I checked my change. That is actually what I did. Maybe the branch name "gyroscope" is confusing. I have a general DeviceMotion event and keep a device sensor type list. If we need more types, just extend the list in ccTypes.h and Cocos2dxMotionSensor.java (android only so far).

Do you prefer EventSenor to DeviceMotion? I can rename it.

DimGun:
Hi guys! I'm also interested in this feature. While I was integrating it into my project I noticed that there are a lot of obsolete whitespace diffs like that
whitespace_diffs
Caused by autoformat I guess.

So for convenience I created "clean" patch, you can find it in my [fork](https://github.com/DimGun/cocos2d-x/commit/c7d98e89e9a35ca60561c6e2e14d294cc0115180). Hope this will help reviewing.

@sftrytry thank you for your work, you've saved me a lot of time.

ku-kueihsi:
@DimGun 
Np. I am so happy my work helps. 

I also noticed this whitespace problem. Is there any reason why whitespace is preferred in cocos2dx?

My change breaks buids anyway, I have to fix that first. Whitespace changes could be excluded easily.

ku-kueihsi:
@dabingnn 
I checked my change. That is actually what I did. Maybe the branch name "gyroscope" is confusing. I have a general DeviceMotion event and keep a device sensor type list. If we need more types, just extend the list in ccTypes.h and Cocos2dxMotionSensor.java (android only so far).

Do you prefer EventSenor to DeviceMotion? I can rename it.

tests/cpp-empty-test/proj.android/project.properties
```diff
@@ -8,6 +8,6 @@
 # project structure.
 
 # Project target.
-target=android-10
+target=android-22
```
zilongshanren: why change the target to 22?

ku-kueihsi: It is automatically generated because I have only api 22 installed on my machine. I will revert that back.

ku-kueihsi:
@dabingnn 
I have two questions.
1. Should I strip out all whitespace changes? Why keep whitespace in the code?
2. I am confused. Jenkins builds are broken. But local tests are fine, e.g., lua test and cpp test. My change only affects cpp/android and there should not be any existing tests calling the new events. Should I fixed Jenkins build before I create the pull request?

dabingnn:
@sftrytry Sorry for my late reply, I am busy preparing for our developer conference in the previous several days.

Here are my suggestions, I wish it would give you some help.
1. `Whitespace` is preferred in the code because the indent of `Tabs` varies on platforms. However, this is not a big issue, I think we can do it later.
2. Please do not hesitate to create pull request even if jenkins compiling test does not passed. The jenkins include many platforms, and it would fail if any of those platforms compile fails.
3. Generally, `EventSensor` and `Device motion` is acceptable, I would prefer `EventSensor` because we are using `EventXXXX` in Cocos2d.

Finally, thanks for your hard work. If you have any questions and concerns, let's discuss it and work it out.

ku-kueihsi:
@dabingnn 
Thanks for your explanation.
I have updated my change. Rename it to EventSensor and clean up whitelist.

ku-kueihsi:
Hi,

Any update for this pull request? I am sorry for the confusing because I updated my id.

IgorMats:
Hi guys, I think it's very useful PR.
Should be tested and merged ASAP.

minggo:
close it since there are many conflict

PR #13398, opened 2015-08-16
More FileUtils::getFileExtension

perminovVS:
check .material in CCPU
check .zip in AssetsManager


WenhaiLin:
@perminovVS Thank you! I think `FileUtils::getFileExtension` should be define as a static function.

perminovVS:
@WenhaiLin Probably yes, i could not think of another implementation heirs.

zilongshanren:
@perminovVS  
Thanks, merged.

PR #13406, opened 2015-08-17
Add BlendFrame supporting to Skeleton Animation

xiaofeng11:
Add BlendFrame supporting to Skeleton Animation
Update json parser to prepare compatible for both current version output json description file and future optimized version json description file.


cocos/editor-support/cocostudio/ActionTimeline/CCActionTimelineCache.cpp
```diff
@@ -882,6 +889,34 @@ Frame* ActionTimelineCache::loadInnerActionFrameWithFlatBuffers(const flatbuffer
...
+    }
+    frame->setBlendFunc(blend);
+
+    int frameIndex = flatbuffers->frameIndex();
+    frame->setFrameIndex(frameIndex);
+
+    bool tween = flatbuffers->tween() != 0;
+    frame->setTween(tween);
+
+    auto easingData = flatbuffers->easingData();
```
pandamicro: What does easing data do with blend frame ?
AFAIK, the blend frame cannot be animated, how does the easing function work ?

Issue #13404, opened 2015-08-17
Add physics component


minggo:
Now physics integration added too much code in `Node` and other engine codes. We should move physics related codes into a component.

pandamicro:
Fixed by https://github.com/cocos2d/cocos2d-x/pull/13923

PR #13408, opened 2015-08-17
add physics component

minggo:
The corresponding issue is #13404.


minggo:
@WenhaiLin 
Could you please review it too?

WenhaiLin:
@minggo Please fix conflicts.

minggo:
Ok, i fixed the conflict.

minggo:
[ci rebuild]

minggo:
@WenhaiLin 
Do you have any comments?
I will merge it if not.

minggo:
> I suggest to use C++ 11 smart pointer rather than the raw delete operators.

What did you mean?

> Does every scene should own a PhysicsManager or just a PhysicWorld?

PhyiscsWorld is managed by PhysicsManager.

zilongshanren:
@minggo  
I mean prefer unique_ptr to raw pointer new & delete.

minggo:
@zilongshanren 
The pointer is used internally, so there is not need to use smart pointer.

minggo:
I will send another squashed PR after it passes compiling.

cocos/physics/CCComponentPhysics2d.cpp
```diff
@@ -0,0 +1,249 @@
...
+        ret->autorelease();
+    
+    return ret;
+}
+
+ComponentPhysics2d::ComponentPhysics2d()
+: _physicsBody(nullptr)
+, _physicsRotation(0.0f)
+, _ownerOriginRotation(0.0f)
+{}
```
WenhaiLin: You should initialize the name of component.

cocos/physics/CCComponentPhysics2d.cpp
```diff
@@ -0,0 +1,249 @@
...
+        _physicsBody->_componentBelongsTo = nullptr;
+        _physicsBody->release();
+        _physicsBody = nullptr;
+    }
+}
+
+void ComponentPhysics2d::addToPhysicsManager()
+{
+    if (_owner && _owner->getScene())
+        _owner->getScene()->getPhysicsManager()->addPhysicsComponent(this);
```
WenhaiLin: Please avoid calling a costly method(`getScene()`) multiple times.

cocos/physics/CCComponentPhysics2d.cpp
minggo: Yep, good catch.

cocos/physics/CCComponentPhysics2d.cpp
minggo: Ok.

minggo:
I sent another squashed PR in #13709.

PR #13427, opened 2015-08-18
performance optimization and bug fixing

songchengjiang:


super626:
@pandamicro  It is ok.

PR #13440, opened 2015-08-18
Fixed `Label::getLetter(index)->setVisible(true)` cause draws duplicate letters.

WenhaiLin:


WenhaiLin:

cocos/2d/CCLabel.cpp
```diff
@@ -144,6 +144,11 @@ class LabelLetter : public Sprite
 
         _textureAtlas->updateQuad(&_quad, _atlasIndex);
     }
+
+    //LabelLetter is not draw self directly.
```
zilongshanren: Do you mean LabelLetter doesn't need to draw themselves? The comments have grammar error here.

WenhaiLin: Letters will be draw by `TextureAtlas::drawQuads()`

Issue #13405, opened 2015-08-17
AutoPolygon's copy contructor


lucbloom:
In PolygonInfo::PolygonInfo(const PolygonInfo& other),
isVertsOwner = other.isVertsOwner;

Shouldn't that be
isVertsOwner = true;

Because it allocates new memory, which should be released in
PolygonInfo::releaseVertsAndIndices

zilongshanren:
@fusijie please take a look at this issue.

fusijie:
@lucbloom Thanks, I have fixed it.

PR #13442, opened 2015-08-18
fix copy construct & assignment operator memory leakage.

fusijie:


fusijie:
To fix #13405.

@zilongshanren Please take a look. Thanks.

zilongshanren:
@fusijie  I will merge the code after CI passed.

PR #13444, opened 2015-08-18
Fix when picture file used by TMX map file missing or broken, create TMX object will cause program crashed

xiaofeng11:
@jun-bing This pr fix bug #22273 on redmine


pandamicro:
Please send your PR to v3.8 branch

PR #13449, opened 2015-08-18
fixed JSB_glGetUniformfv uniform location is different from index.

fusijie:

PR #13465, opened 2015-08-19
fix opengl test error, add the condition

jianglong0156:
fix opengl test error, add the condition, when the data type is samewith long, it will reset data
relate pr:#13449
fix open gl show error

### DIFF
--- a/cocos/scripting/js-bindings/manual/js_manual_conversions.cpp
+++ b/cocos/scripting/js-bindings/manual/js_manual_conversions.cpp
@@ -187,8 +187,13 @@ bool jsval_to_int( JSContext *cx, JS::HandleValue vp, int *ret )
     // Since this is called to cast uint64 to uint32,
     // it is needed to initialize the value to 0 first
 #ifdef __LP64__
-    long *tmp = (long*)ret;
-    *tmp = 0;
+    // only unsigned int size is 8Bit, same with the long, it's need to reset the data.
+    // the ret will convert to int32_t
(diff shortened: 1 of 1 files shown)


cocos/scripting/js-bindings/manual/js_manual_conversions.cpp
```diff
@@ -187,8 +187,13 @@ bool jsval_to_int( JSContext *cx, JS::HandleValue vp, int *ret )
     // Since this is called to cast uint64 to uint32,
     // it is needed to initialize the value to 0 first
 #ifdef __LP64__
-    long *tmp = (long*)ret;
-    *tmp = 0;
+    // only unsigned int size is 8Bit, same with the long, it's need to reset the data.
```
pandamicro: When `int` size is 8 Bit (same as `long`), the following operation is needed

cocos/scripting/js-bindings/manual/js_manual_conversions.cpp
```diff
@@ -243,8 +248,13 @@ bool jsval_to_uint( JSContext *cx, JS::HandleValue vp, unsigned int *ret )
...
```
pandamicro: When `unsigned int` size is 8 Bit (same as `long`), the following operation is needed

pandamicro:
Please squash all commits into one

jianglong0156:
other squash pr:https://github.com/cocos2d/cocos2d-x/pull/13487
close this pr

PR #13471, opened 2015-08-19
skybox should not be transparent

super626:
Default transparent value of rendercommand is true.
It should be opaque. So that it can be drawn before transparent objects.


super626:
[ci rebuild]

Issue #10205, opened 2015-01-25
Crash when loading cocostudio json files with null or empty fontName


ghost:
Call this function:
cocostudio::GUIReader::getInstance()->widgetFromJsonFile(filename);

If the specified json file contains a "TextField" with the property "fontName" set to either null or "", it prints a could not load file error and then crashes.

walzer:
Well, it should print an error log + return false, instead of crash immediately.

Issue #9488, opened 2014-12-12
setColor and TintTo/By work on a whole armature, but not on an individual bone


hawkwood:
Trying to setColor or run TintTo on just one bone in an armature does not work.  Verify by adding following code to the end of TestCSWithSkeleton::onEnter() of ArmatureScene.cpp

```
armature->setColor( Color3B(128, 0, 0));
Bone *hat = armature->getBone( "Layer136" );
hat->setColor( Color3B(0, 128, 128));
```

With this, hat should be black.  If line 1 is commented out, hat should be teal. If lines 2 & 3 are commented out, hat should be red.

The problem seems to be line 260 of CCBone.cpp, where _realColor is reset to WHITE, removing the individual bone's color.  Commenting out this line solves the issue, but I do not know what other ramifications this might cause.

Related, since line 266 of CCBone.cpp resets _realOpacity to 255, this means that opacity can not be set on a specific bone.

chengstory:
@pipu   please take a look

pipu:
@2youyouo2

Take a look, please.

2youyou2:
hi, hawkwood.
I have test it. It seems it's ok to remove those two lines in CCBone.cpp.
I forgot what problem to solve with those two lines.

xiaofeng11:
Maybe line 260 & line 266 is used for Cocostudio force bone to been draw on screen, In user project these two line can be comment to avoid problem

PR #13478, opened 2015-08-20
Fix Cocos Studio Bugs

xiaofeng11:
Fixed: #10205 and #9488


cocos/editor-support/cocostudio/CCBone.cpp
```diff
@@ -257,13 +257,15 @@ void Bone::setBlendFunc(const BlendFunc& blendFunc)
 
 void Bone::updateDisplayedColor(const Color3B &parentColor)
 {
-    _realColor = Color3B(255, 255, 255);
+    // remove comment mark for next line when use in studio
```
pandamicro: You mean cocos studio need this ?
Why not wrap it in the macro ?

pandamicro:
Thanks

PR #13480, opened 2015-08-20
add blendfunc Frame test case

geron-cn:
add blendfunc Frame test case


geron-cn:
@jun-bing   @pandamicro @VisualSJ  can you check this out.

PR #13489, opened 2015-08-20
fix texturecache didn't show correctly in win32

jianglong0156:
fix texturecache didn't show correctly in win32


jianglong0156:
all platform test over, it's fine as expected. @pandamicro

PR #13490, opened 2015-08-20
Camera Background Clear brush

super626:
Add camera background clear brush
And test demo

Usage
camera->setBackGroundBrush(CameraBackGroundBrush::createSkyboxBrush(...));
camera->setBackGroundBrush(CameraBackGroundBrush::createDepthBrush(...));
And so on.

PR #13497, opened 2015-08-21
Add CCCameraBackgroundBrush

pandamicro:


pandamicro:
Squashed from https://github.com/cocos2d/cocos2d-x/pull/13490

PR #13504, opened 2015-08-21
Renderer::checkVisibility crashes

ryule:
Added nullptr check to fix a crash of Render::checkVisibility().
It may be happened when visit() called manually for a reason such as to perform render-to-texture.


fusijie:
Thanks for your pr.
If you use a multiple scene, the bug is still exist because `Camera::_visitingCamera` is a static variable and not clear by replace scene.

ryule:
Yes, you are right. This pr can be closed.

PR #13535, opened 2015-08-24
Add necessary methods into RadioButton

neokim:
Some useful methods are added into RadioButton.


zilongshanren:
@neokim :+1:  Merged.

Issue #13564, opened 2015-08-25
Scale9Sprite draw extra 1 pixel when creating from spritesheet


zilongshanren:
This issue was reported by developers from Chinese forum.
He found that when creating Scale9Sprite from spritesheet and the resulting Scale9Sprite will draw extra 1 pixel from other spriteframe.  

He also proposed a solution:

```
Vec2 offsetPosition(ceil(_offset.x + (_originalSize.width - _spriteRect.size.width) / 2),
                            ceil(_offset.y + (_originalSize.height - _spriteRect.size.height) / 2));
```

should be changed to 

```
Vec2 offsetPosition(floor(_offset.x + (_originalSize.width - _spriteRect.size.width) / 2),
                            floor(_offset.y + (_originalSize.height - _spriteRect.size.height) / 2));
```

Chinese forum:

http://www.cocoachina.com/bbs/read.php?tid=320932&page=1&toread=1#1376926

and 

his blog:

http://www.cnblogs.com/songcf/p/4755193.html

zilongshanren:
@newnon  
I remember that the `ceil` function was added by you.
Thoughts?  How about changing ceil to floor?

Thanks.

newnon:
I'm not sure it was long time ago. may be it is correct fix may be it will give some other artifacts

zilongshanren:
@newnon
I agree with you. The offset and originalSize read from plist files will be divided by ContentScaleFactor which will cause some precision lose.

I still can't make sure the fix is correct but at least it resolve the developer's issue and all the tests are good now.

I will keep a an eye with it.

Thanks for reply.

PR #13572, opened 2015-08-26
fix Scale9Sprite draw extra pixels in spritesheet

zilongshanren:
Fixed https://github.com/cocos2d/cocos2d-x/issues/13564

### DIFF
--- a/cocos/ui/UIScale9Sprite.cpp
+++ b/cocos/ui/UIScale9Sprite.cpp
@@ -580,8 +580,8 @@ namespace ui {
         float width = _originalSize.width;
         float height = _originalSize.height;
 
-        Vec2 offsetPosition(ceilf(_offset.x + (_originalSize.width - _spriteRect.size.width) / 2),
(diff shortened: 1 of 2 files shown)


cocos/ui/UIScale9Sprite.cpp
```diff
@@ -794,7 +794,7 @@ namespace ui {
         //shrink the image size when it is 9-patch
         if(_isPatch9)
         {
-            float offset = 1.4f;
```
pandamicro, line 797: why ?

WenhaiLin: @zilongshanren I think it mean one pixel, right?

zilongshanren, line 797: @pandamicro  Android 9-patch image has extra 1 pixel should be removed.
The previous 1.4 is a magic number.

zilongshanren:
[ci rebuild]

PR #13573, opened 2015-08-26
add annotations

songchengjiang:


pandamicro:
Please add documentation for Physics3DColliderDes.
All parameters and return value should be documented too.

You can refer to other header files

PR #13577, opened 2015-08-26
Fixed sprite debug draw.

### LINKED ISSUE #13454: CC_SPRITE_DEBUG_DRAW does nothing in Cocos2d-x 3.7
I set CC_SPRITE_DEBUG_DRAW to 1 and to 2 and it does nothing. Not sure if there is something else I have to do but this looks like a bug.

From what I can tell, in 3.6 the DrawNode used to draw the outline of sprites is refreshed during each draw call, so it always uses the latest vertex data when drawing the outline. In 3.7, the DrawNode is created and set once. Unfortunately this is done in the constructor of the Sprite before texture data is set, so the triangle count is always 0 and no outline is drawn.

Calling `debugDraw(true);` immediately after `_polyInfo.setQuad(&_quad);` works for me, surrounded by `#if CC_SPRITE_DEBUG_DRAW` of course.

fusijie:
To fix #13454.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -289,7 +289,8 @@ Sprite::Sprite(void)
 , _insideBounds(true)
 {
 #if CC_SPRITE_DEBUG_DRAW
-    debugDraw(true);
+    _debugDrawNode = DrawNode::create();
+    addChild(_debugDrawNode);
 #endif //CC_SPRITE_DEBUG_DRAW
 }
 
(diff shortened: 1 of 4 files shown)


fusijie:
@dabingnn @pandamicro @zilongshanren.

The pr will **remove Sprite::debugdraw() API** to restore the status of v3.6. Please take a look, thanks.

fusijie:
[ci rebuild]

WenhaiLin:
@zilongshanren I think this PR is OK for merge.

PR #13578, opened 2015-08-26
bone's color and opacity cannot cascade to bone

geron-cn:
bone's color and opacity cannot cascade to bone, also bone can not cascade from its parent.


geron-cn:
@minggo @VisualSJ  @jun-bing can you check this out, thx.

pandamicro:
Review ok, @VisualSJ, you may need to update the web engine

PR #13587, opened 2015-08-26
implementation of new AudioEngine interface for Linux

cesarpachon:
new attempt of PR this feature.. I finally was able to squash all in a single commit and also it is updated to the latest changes of the v3 branch. hope it pass!


WenhaiLin:
:+1: Great!  I will test it ASAP.

cesarpachon:
Hello @WenhaiLin  I noticed issues with FMOD in ubuntu 15 (not in cocos code, but in the fmod library itself), I made a question in the FMOD forums regarding the current version used in cocos2d-x (4.38.00)  and they said that we are using a really very old version.. I would like to test with the latest one, but that involves changes in the dependencies project and also will impact the SimpleAudioEngine. what do you think? http://www.fmod.org/questions/question/creepy-sound-on-ubuntu-15-app-not-listed-in-pulseaudio

WenhaiLin:
@cesarpachon I approve of you test with an up to date version. Then we can wrap AudioEngine to SimpleAudioEngine interfaces if there is no compatibility problem.

Dimon4eg:
Any update?

minggo:
This PR seems to have many unneeded commits.

PR #13592, opened 2015-08-27
Fix #13591

### LINKED ISSUE #13591: Android app freezes sometimes because Cocos2dxRenderer.nativeOnResume() is not called when the activity is resumed.
`Cocos2dxRenderer.nativeOnResume()` is supposed to be called when the activity is resumed. It is called through by `Cocos2dxActivity.onWindowFocusChanged()`.

I found a case that `onWindowFocusChanged()` is not being called after `onResume()`. It causes the app to freeze.

Normally `onWindowFocusChanged(false)` is called right after `onPause()` and `onWindowFocusChanged(true)` right after `onResume()`.
But if the device's display is turned off by itself because of screen timeout, a time gap is generated between `onPause()` and `onWindowFocusChanged(false)` like following. In this case, It took 7 seconds. Test device was Samsung Galaxy S4(SHV-E300S).

```
08-27 16:39:42.561: D/Cocos2dxActivity(26545): onPause()
08-27 16:39:49.646: D/Cocos2dxActivity(26545): onWindowFocusChanged() hasFocus=false
08-27 16:39:52.761: D/Cocos2dxActivity(26545): onResume()
08-27 16:39:53.166: D/Cocos2dxActivity(26545): onWindowFocusChanged() hasFocus=true
```

If user turns on the display in the time gap right after `onPause()`, `onResume()` is called right then. It leads `onWindowFocusChanged(true)` not to be called because the activity hasn't lose focus.

```
08-27 16:40:09.651: D/Cocos2dxActivity(26545): onPause()
08-27 16:40:10.586: D/Cocos2dxActivity(26545): onResume()
```

In this case, `Cocos2dxRenderer.nativeOnResume()` is not being called and the app freezes.

neokim:
Fix #13591

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
@@ -108,6 +108,7 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
     private Cocos2dxVideoHelper mVideoHelper = null;
     private Cocos2dxWebViewHelper mWebViewHelper = null;
     private Cocos2dxEditBoxHelper mEditBoxHelper = null;
+    private boolean hasFocus = false;
 
     public Cocos2dxGLSurfaceView getGLSurfaceView(){
         return  mGLSurfaceView;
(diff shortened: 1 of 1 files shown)


zilongshanren:
@WenhaiLin  @fusijie   please review this PR.

WenhaiLin:
@zilongshanren I think this PR is OK for merge. @neokim Thank you!

Issue #13575, opened 2015-08-26
cocos2d::ui::Slider does not tak TextureResType into consideration


Birkemose:
In UISlider.cpp, line 99

widget->loadBarTexture(barTextureName);
widget->loadSlidBallTextureNormal(normalBallTextureName);

should be changed to

widget->loadBarTexture(barTextureName, resType);
widget->loadSlidBallTextureNormal(normalBallTextureName, resType);

zilongshanren:
@Birkemose  Good catch, thanks.

PR #13598, opened 2015-08-27
fix Slider create missing res type.

zilongshanren:
fixed https://github.com/cocos2d/cocos2d-x/issues/13575


pandamicro:
Review ok

PR #13622, opened 2015-08-29
Introduce touch down event support for CCMenu

giginet:
We can't activate callback functions when `MenuItem`s are pressed down.

This PR make it to enable to handle touch down event on `Menu`.

``` cpp
auto menuItem = cocos2d::MenuItemImage::create("button.png", "button.png", [](Ref *button) {
    cocos2d::log("touch down event");
});
auto menu = cocos2d::Menu::create(menuItem, nullptr);
menu->setEventType(cocos2d::Menu::EventType::TOUCH_DOWN);
```

### DIFF
--- a/cocos/2d/CCMenu.cpp
+++ b/cocos/2d/CCMenu.cpp
@@ -134,6 +134,7 @@ bool Menu::initWithArray(const Vector<MenuItem*>& arrayOfItems)
     if (Layer::init())
     {
         _enabled = true;
+        _eventType = EventType::TOUCH_UP;
         // menu in the center of the screen
         Size s = Director::getInstance()->getWinSize();
 
@@ -273,7 +274,11 @@ bool Menu::onTouchBegan(Touch* touch, Event* event)
         _state = Menu::State::TRACKING_TOUCH;
         _selectedWithCamera = camera;
         _selectedItem->selected();
-        
+        if (_eventType == EventType::TOUCH_DOWN) 
+        {
+            _selectedItem->activate();
+        }
+
         return true;
     }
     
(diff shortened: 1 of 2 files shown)


pandamicro:
I think it's a good add-up to the current Menu system, but it adds new APIs, so I need your opinion @minggo @zilongshanren

giginet:
Hi, any update on this issue? 
do you think about this feature? Please check it.
@minggo @zilongshanren 

Thank you :smiley:

cocos/2d/CCMenu.cpp
```diff
@@ -273,7 +274,11 @@ bool Menu::onTouchBegan(Touch* touch, Event* event)
         _state = Menu::State::TRACKING_TOUCH;
         _selectedWithCamera = camera;
         _selectedItem->selected();
-        
+        if (_eventType == EventType::TOUCH_DOWN) 
+        {
+            _selectedItem->activate();
+        }
+
```
minggo, line 281: Why need these codes?

cocos/2d/CCMenu.cpp
```diff
@@ -287,7 +292,10 @@ void Menu::onTouchEnded(Touch* touch, Event* event)
     if (_selectedItem)
     {
         _selectedItem->unselected();
-        _selectedItem->activate();
+        if (_eventType == EventType::TOUCH_UP) 
+        {
+            _selectedItem->activate();
+        }
```
minggo, line 298: Why need to determine like `_eventType == EventType::TOUCH_UP`?

cocos/2d/CCMenu.cpp
giginet, line 281: Because if event_type was `TOUCH_UP`,  `activate` should not be called when touch is began.

cocos/2d/CCMenu.cpp
giginet, line 298: Same as above

minggo:
There is already `onTouchBegan`, `onTouchMove` events, why need these events? And it seems they are duplicated.

PR #13625, opened 2015-08-30
Update CCImage.cpp

frranck:
When you're creating a sprite (e.g. with Sprite\* Sprite::create(filename))
There's no way to test the file before, without getting "a unsupport image format!" thrown.

### DIFF
--- a/cocos/platform/CCImage.cpp
+++ b/cocos/platform/CCImage.cpp
@@ -576,7 +576,7 @@ bool Image::initWithImageData(const unsigned char * data, ssize_t dataLen)
                 }
                 else
                 {
-                    CCAssert(false, "unsupport image format!");
+                    CCLOG("cocos2d: unsupported image format!");
                 }
                 
                 free(tgaData);


pandamicro:
I think this should be merged, @WenhaiLin @fusijie What do you think ?

fusijie:
I think `assert` will give the devers more explicit instructions...

frranck:
@fusijie the problem is: how do you test the data if you fetch it online ? You can only crash your app right now. It used to be different in v2.

pandamicro:
In development phase, of course a more explicit assert or warning can avoid user from some mistaken implementations. But when the game is published, developers will want to avoid crashes as much as they can.

So it's about guarantee experience to the developer or the end user. Personally, I think experience to the end user is more important.

frranck:
@pandamicro @fusijie Development or published phase is not important for me in this situation:
The problem is I'm getting the images from the Facebook API, so I need a function that can tell me if cocos2d-x can handle a given image or not. It's totally fine if cocos-2d can't handle a certain image type but then I need to be able to know it in a proper way (e.g. by checking if the Sprite is null for example).
Right now the only way to know an image format is handled or not is by crashing my program, it's totally not OK... 
How can I programmatically know if cocos2d-x can handle an image format or not  ?

WenhaiLin:
@frranck @pandamicro @fusijie You should publish your game in release mode.`CCASSERT` used to help developers find problems in debug mode.

WenhaiLin:
And, of course, we should avoid some specific error cause crash in engine code in release mode. For example,file does not exist,unsupported image format,etc.

pandamicro:
@frranck Thanks for the clarification. In this particular case, I agree with you, a log and return null is sufficient to notify developers, and the return value can be useful to make game logic more flexible.

PR #13626, opened 2015-08-30
cancel request from queue in HttpClient

dakabang:
cancel non-immediate request from queue in HttpClient

### DIFF
--- a/cocos/network/HttpClient.cpp
+++ b/cocos/network/HttpClient.cpp
@@ -472,7 +472,7 @@ void HttpClient::cancelSend(const char* tag){
 
 	_requestQueueMutex.lock();
 	for (int i = 0; i < _requestQueue.size(); i++){
-		if (_requestQueue.at[i]->getTag() == tag){
+		if (_requestQueue.at(i)->getTag() == tag){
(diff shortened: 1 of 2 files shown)


cocos/network/HttpClient.cpp
```diff
@@ -455,6 +455,30 @@ void HttpClient::send(HttpRequest* request)
...
+//Cancel the request based on tag
+void HttpClient::cancelSend(const char* tag){
+	if (_requestQueue.empty()){
+		return;
+	}
+
+	_requestQueueMutex.lock();
+	for (int i = 0; i < _requestQueue.size(); i++){
+		if (_requestQueue.at(i)->getTag() == tag){
+			_requestQueue.erase(i);
```
Dimon4eg, line 476: no break here?
If you want to remove all requests with specified tag then you need to decrement 'i' variable after removing request.

xdeop:
Hi, don't you need to modify also HttpClient-android.cpp, HttpClient-apple.mm, HttpClient-winrt.cpp ?? 

Also, have you been able to figure out how to cancel a current ongoing request interrupting the network thread?

Thanks.

PR #13627, opened 2015-08-30
 trace() does not default to cocos2d::Rect::Zero

seivan:
The comments/docs state this   

```
@param   rect    a texture rect for specify an area of the image, use Rect::Zero for the size of the image, default to Rect::Zero
```

Which wasn't accurate. Now it is :+1:


seivan:
Squashed commits into one.

fusijie:
trace()'s rect should not be Zero which will cause a assert, I will modify the comments.

PR #13631, opened 2015-08-31
fix bug: bone can be see by other cameras

geron-cn:
fix bug: bone can be see by other cameras


geron-cn:
@jun-bing @VisualSJ  @pandamicro   could you check this out, thx.

pandamicro:
[ci rebuild]

PR #13632, opened 2015-08-31
Fix missing config.json notice when simulator start to run a cocostudio project

xiaofeng11:
@jun-bing Fix bug http://punchbox.info:3000/issues/22077
@pandamicro Please review it


tools/simulator/frameworks/runtime-src/proj.ios_mac/mac/SimulatorApp.mm
```diff
@@ -205,10 +205,29 @@ - (void) parseCocosProjectConfig:(ProjectConfig*)config
     
     // set project directory as search root path
     FileUtils::getInstance()->setDefaultResourceRootPath(tmpConfig.getProjectDir());
-    
+    string solutionDir = tmpConfig.getProjectDir();
```
pandamicro, line 207: It should be put before the previous line and setDefaultResourceRootPath with solutionDir

PR #13635, opened 2015-08-31
Enable depth write for transparent object

super626:


pandamicro:
@dabingnn Can I merge this PR ?

pandamicro:
It's not a definitive solution, but a temporary and incomplete fix for 3D transparent object overlapping, @super626 will continue to find a correct solution for this issue

super626:
@pandamicro 
I am going to try to find a better solution, Weighted, Blended Order-Independent Transparency (OIT). There are some articles discussing it. It may comes after the version of v3.9

PR #13634, opened 2015-08-31
This is a small optimization about TextReader.

WenhaiLin:

PR #13640, opened 2015-08-31
Fix/Optimisation of TextReader

Almax27:
If a long piece of text is defined in cocosStudio with a custom font, TextReader will unnecessarily (and implicitly) build the label with the default system font first. It will also trigger repeat recalculations of the label's size, atlas and system font texture via the setFontName() and setFontSize() calls.

By setting the text content last we can drastically minimise this overhead.

Furthermore if the text content is large the system font renderer will potentially produce a texture wider than the maximum texture size of the device. This is avoided with the given change for custom fonts but will still pose an issue for large labels relying on the default system font.


zilongshanren:
duplicated with https://github.com/cocos2d/cocos2d-x/pull/13634

PR #13646, opened 2015-09-01
Refactor scroll view's inertia scroll

neokim:
Scroll view used to have two different logics for scroll by inertia and auto scroll. The logics are merged into one. Inertia scroll now uses auto scroll feature.

This is base change to implement scrolling to specific item and magnetic scroll in list view.

### DIFF
--- a/cocos/ui/UIScrollView.cpp
+++ b/cocos/ui/UIScrollView.cpp
@@ -31,8 +31,6 @@ THE SOFTWARE.
 #include "2d/CCCamera.h"
 NS_CC_BEGIN
 
-namespace ui {
-
 static const float INERTIA_DEACCELERATION = 700.0f;
 static const float INERTIA_VELOCITY_MAX = 2500;
 static const float BOUNCE_BACK_DURATION = 1.0f;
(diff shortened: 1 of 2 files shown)


WenhaiLin:
:+1:

zilongshanren:
[ci rebuild]

zilongshanren:
[ci rebuild]

WenhaiLin:
[ci rebuild]

pandamicro:
So guys, who is reviewing this PR and can we merge it ?

zilongshanren:
@pandamicro  I will review this PR later.

zilongshanren:
@neokim  :+1:  I have tested your code and it works great.  Keep up with your improvement over the ScrollView and ListView.

neokim:
@zilongshanren Thanks for reviewing.

I'm adding more features to ListView, and refactoring, too.
Because ScrollView is the very base component, this pull request was necessary for those changes.

Also planning to change PageView to derive from ListView. I saw an issue about vertical PageView. It will be resolved if it shares ListView's implementation. And ScrollView, ListView and PageView will have same user experience about scrolling. Actually PageView is one type of ListView which has magnetic scroll option as center.

zilongshanren:
@neokim 
What features would you plan to add for ListView?  Because currently I'm also trying to add cell reuse to ListView and PageView.

I like the idea which PageView could share some implementations of ListView and ScrollView.
But we should do it carefully, we can't break back compatibility in v3.x versions.

neokim:
@zilongshanren I agree with that we have to take care for back compatibility. I'll try to keep every existing API of PageView.

I'm adding magnetic scroll feature to ListView now. I think that it has no dependency with your reusable cell feature.

zilongshanren:
@neokim  Great :+1:

PR #13659, opened 2015-09-01
EventListenerMouse will dispatch EventMouse events

seivan:
If you use an EventListenerMouse you want to know the coordinates for onMouseMove.
You don't get that with normal Event object, but through EventMouse.

The header was included, I suspect someone forgot to use it in the callbacks for the new model on handling touch/mouse.

### DIFF
--- a/cocos/base/CCEventListenerMouse.cpp
+++ b/cocos/base/CCEventListenerMouse.cpp
@@ -82,19 +82,19 @@ bool EventListenerMouse::init()
         {
             case EventMouse::MouseEventType::MOUSE_DOWN:
                 if(onMouseDown != nullptr)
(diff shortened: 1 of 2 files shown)


pandamicro:
I think it's a good refine, @WenhaiLin please take a look.

@seivan Can you squash your commits into one ?

seivan:
@pandamicro Done! 
While we're on the subject of touch, if you get some time could you take a look at https://github.com/cocos2d/cocos2d-x/issues/13671 - thanks!

WenhaiLin:
@pandamicro I think this PR is OK for merge.

pandamicro:
ok, thanks guys

PR #13661, opened 2015-09-01
Fix compile error in __LayerRGBA

TheCodez:


pandamicro:
Good catch, thanks

pandamicro:
Your commit have been merged in #13707

PR #13673, opened 2015-09-03
Removed ScrollView::_innerContainer pointer copy

Almax27:
Copied pointer becomes invalid when model widget destructs. Instead we simply copy the position.

### DIFF
--- a/cocos/ui/UIScrollView.cpp
+++ b/cocos/ui/UIScrollView.cpp
@@ -1342,8 +1342,8 @@ void ScrollView::copySpecialProperties(Widget *widget)
     if (scrollView)
     {
         Layout::copySpecialProperties(widget);
-        _innerContainer = scrollView->_innerContainer;
         setDirection(scrollView->_direction);
+        setInnerContainerPosition(scrollView->getInnerContainerPosition());
         setInnerContainerSize(scrollView->getInnerContainerSize());
         _topBoundary = scrollView->_topBoundary;
         _bottomBoundary = scrollView->_bottomBoundary;


cocos/ui/UIScrollView.cpp
```diff
@@ -1342,8 +1342,8 @@ void ScrollView::copySpecialProperties(Widget *widget)
     if (scrollView)
     {
         Layout::copySpecialProperties(widget);
-        _innerContainer = scrollView->_innerContainer;
         setDirection(scrollView->_direction);
+        setInnerContainerPosition(scrollView->getInnerContainerPosition());
```
pandamicro, line 1346: I think it's more reasonable to do the following:

_innerContainer->copySpecialProperties(widget->_innerContainer);

But I'm not very familiar with UI, @zilongshanren please take a look

Almax27:
@pandamicro That's a much more complete solution indeed. If @zilongshanren is cool with it too I'll update the pull request.

cocos/ui/UIScrollView.cpp
Almax27, line 1346: @pandamicro That's a much more complete solution indeed. If @zilongshanren is cool with it too I'll update the pull request.

pandamicro:
@zilongshanren @ZhangMenghe please help review

cocos/ui/UIScrollView.cpp
zilongshanren, line 1346: thanks @Almax27 
@pandamicro  The `copySpecialProperties` method is a protected method and the `_innerContainer` is initialized with the default settings, there is no need to copy the properties.
Since when ScrollView is created, the `_innerContainer` already has the default settings and we should only clone the children which is done by `ScrollView::copyClonedWidgetChildren`.
So your PR is good to merge.

zilongshanren:
@Almax27  Thanks for the patch :+1:

PR #13684, opened 2015-09-04
fix #13285: AppController - synthesize window

### LINKED ISSUE #13285: iOS AppController: synthesize window
I found that AppController doesn't syntesize window property, or provide getter/setter for it, it just declare local variable:

```
@interface AppController : NSObject <UIApplicationDelegate> {
    UIWindow *window;
}
```

But UIApplicationDelegate has required property:

```
@property (nonatomic, retain) UIWindow *window NS_AVAILABLE_IOS(5_0);
```

**Why it is important:**
Standard way to get root view controller is to use:

```
[UIApplication sharedApplication].delegate.window.rootViewController;
```

But in our case app will crash - window selector doesn't exist for our implementation of app delegate.

You can say that can be used just:

```
[UIApplication sharedApplication].keyWindow.rootViewController
```

But application may have multiple windows, and keyWindow may be not our window in some cases (really rare cases, ofc). But more important that some libraries may use previous approach.

**How to fix:**
Add next row inside implementation file:

```
@synthesize window;
```

Shulepov:


WenhaiLin:
please submit it to v3 branch

Issue #12518, opened 2015-06-24
iOS: ARC support


Shulepov:
Is it possible to add ARC support for generated project? I'm not asking for adding ARC support for cocos2d-x itself, just for game project. Many libraries require ARC support and i don't see any reason why it can't be added, at least as an additional option when starting new project.

It won't bring any performance issues as usually there are no any performance sensitive iOS code in the generated projects, just some bindings to libraries.

Shulepov:
How would be better to implement this feature? 
- add new project folder, something like proj.ios_mac_arc
- use preprocessor definitions to check if ARC enabled:
  #if __has_feature(objc_arc)
  ...
  #else
  ...
  #endif

But with second approach it belongs to user to enable/disable ARC in project settings.

minggo:
It is fixed.

PR #13686, opened 2015-09-04
arc support #12518

Shulepov:
Arc still needs to be enabled/disabled manually in project settings. But that the only thing that should be done.

### DIFF
--- a/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm
+++ b/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm
@@ -49,7 +49,7 @@ static AppDelegate s_sharedApplication;
 
     // Init the CCEAGLView
     CCEAGLView *eaglView = [CCEAGLView viewWithFrame: [window bounds]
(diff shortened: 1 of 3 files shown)


pandamicro:
Thanks for your contribution, please submit it to v3 branch, because we are closing v3.8

zilongshanren:
@Shulepov @pandamicro 
I think cocos2d-x can't support ARC. There are many `autorelease` call scattered in the code base.
And the patch only modify the template project...

Shulepov:
@zilongshanren
Actually, autorelease used in subproject. And if main project support ARC, there is not requirement for any sub-project or library to also support ARC. So it's safe to do this.

I modified only template project, because it won't give any performance penalties for cocos2d-x itself, also it will simplify integration of different libraries and sdks, and writing Obj-C code itself.

 Ofc, it's easy to do this migration manually each time, but i found that slightly annoying...

zilongshanren:
@Shulepov  
I see. Feel free to reopen this PR. Thanks.

zilongshanren:
@Shulepov  Maybe you should send a PR in v3 branch.

PR #13687, opened 2015-09-04
Adjustments to isDirectoryExistInternal

hawkwood:


pandamicro:
@jianglong0156 Please review

cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -394,6 +394,24 @@ static void addObjectToNSDict(const std::string& key, const Value& value, NSMuta
     return ret;
 }
 
+bool FileUtilsApple::isDirectoryExistInternal(const std::string& dirPath) const
```
minggo, line 397: Why need the implementation? What's the issue of current implementation?

luyaor:
Dear Justin Hawkwood(hawkwood),
We are working on identifying redundant development and duplicated pull requests. We have found there is a pull request: https://github.com/cocos2d/cocos2d-x/pull/14883 which might be similar or duplicate to this one. We would really appreciate if you could help us to validate and give us feedback.
Thank you very much for your time!

Issue #13656, opened 2015-09-01
cocos2d::Map may cause Dangling Pointers when inserting Ref Object which already exist in the Map


c3076802:
cocos2d::Map may cause Dangling Pointers when inserting Ref Object which already exist
CCMap  line 274

``` C++
    void insert(const K& key, V object)
    {
        CCASSERT(object != nullptr, "Object is nullptr!");
        erase(key);
        _data.insert(std::make_pair(key, object));
        object->retain();
    }
```

For example
Object A already exist in Map B, and Map B has the very last reference of Object A. The code above will release Object A and Object A will be deleted, then Map B holds a Dangling Pointer.

How about retain the input object first, or check if the input object exist?

seivan:
But from what I gather, the actual deletion happens in an autoreleasepool and might not trigger during the execution of insert().

You are actually accurate that the retain happens _after_ a release and that's bad.
Always retain what you work with, and release once you're done. So it should def be fixed :)

c3076802:
@seivan 
I think you mis-understand the problem. For a newly created object surely the code above won't crash, because autorelease pool holds the +1 ref count that increased by the constructor. That Is not the case I mentioned ,  "Object A already exist in Map B, and Map B has the very last reference of Object A". 
Of course I could retain the object  then insert then release or check if the object already inserted, but shouldn't be the container at least generates a assert in DEBUG MODE for that ? 
if we insert a already existed item into std::map it will fail, which might cause memory leak but it certainly won't crash, cocos2d::Map is a wrapper of std::map but it behave differently at some point, It is a bit confusing for me, overall cocos2d::Map works fine.

c3076802:
void insert(const K& key, V object)
	{
		CCASSERT(object != nullptr, "Object is nullptr!");
		object->retain();		// retain first might fix the problem
		erase(key);
		_data.insert(std::make_pair(key, object));
		//object->retain();
	}

seivan:
No dude, I understood you correctly. 
You erase it before you retain it, it might trigger a delete.

c3076802:
@seivan 
What I am trying to state is How the problem should be handled, not what the problem is. 
I think this kind of error should be and can be handled by the lib. If we need to retain every object before we pass to any function we don't we just use smart-pointer ? This problem similar to self-assign. It should be and can be handled by the container, like what I commented before at least should raise assert in DEBUG MODE.

seivan:
@c3076802 
[...]

PR #13701, opened 2015-09-06
Don't call release() before retain() #13656

seivan:
My c++ is pretty poor, but it might help with https://github.com/cocos2d/cocos2d-x/issues/13656

### DIFF
--- a/cocos/base/CCMap.h
+++ b/cocos/base/CCMap.h
@@ -274,9 +274,9 @@ public:
     void insert(const K& key, V object)
     {
         CCASSERT(object != nullptr, "Object is nullptr!");
+        object->retain();
         erase(key);
         _data.insert(std::make_pair(key, object));
-        object->retain();
     }
     
     /** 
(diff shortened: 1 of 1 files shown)


WenhaiLin:
@seivan This is not correct solution for issue #13656.  please review suggestions from @c3076802 .

seivan:
Squashed and updated with the fix from @c3076802

WenhaiLin:
[ci rebuild]

cocos/base/CCMap.h
```diff
@@ -274,9 +274,9 @@ class Map
     void insert(const K& key, V object)
     {
         CCASSERT(object != nullptr, "Object is nullptr!");
-        erase(key);
-        _data.insert(std::make_pair(key, object));
         object->retain();
+        erase(key)
```
WenhaiLin: You should add a ";" (semicolon) at the end.

WenhaiLin:
@pandamicro I think this PR is OK for merge.

PR #13718, opened 2015-09-08
 Disable iOS8 Quicktype Keyboard

Javanaise:
Disable iOS8 Quicktype Keyboard


seivan:
This might also disable auto-correct, might not be that obvious. I like that my phone knows my name and other things.

pandamicro:
I think for general purpose, we'd better keep the default behavior

Javanaise:
Well find it quite confusing, people not on iOS8 or android (?) probably don't expect the textfield content to change from more than 1 letter, and it adds an extra space at the end...

Javanaise:
@pandamicro @seivan  Something else: Quicktype Keyboard collides with the setMaxLength attribute; if the word you select with the Quicktype suggestion is bigger than your setMaxLength, it just get ignored... That looks very buggy.

PR #13725, opened 2015-09-09
Refine ComponentContainer to  improve performance

WenhaiLin:


WenhaiLin:
@minggo Please review this PR.

minggo:
It is good to me. 

The potential problem is that `std::type_info::name` can not make sure to be unique in different classes. The specification says:

> The particular representation pointed by the returned value is implementation-defined, and may or may not be different for different types.

Can refer to [here](http://www.cplusplus.com/reference/typeinfo/type_info/name/) for detail information. 

`std::type_info::name` is used in script binding codes to identify a class type and we don't find problems in current supported platforms. So i think it is ok to use it here too.

PR #13754, opened 2015-09-09
Fix CMake for Android Target

WenhaiLin:


ricardoquesada:
deprecated. use this PR instead: https://github.com/cocos2d/cocos2d-x/pull/15146

PR #13760, opened 2015-09-10
fix material clone

super626:
Material::clone doesn't set the parent correctly.

### DIFF
--- a/cocos/renderer/CCMaterial.cpp
+++ b/cocos/renderer/CCMaterial.cpp
@@ -442,6 +442,10 @@ Material* Material::clone() const
         for (const auto& technique: _techniques)
         {
             auto t = technique->clone();
+            t->_parent = material;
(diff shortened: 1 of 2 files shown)


cocos/renderer/CCMaterial.cpp
```diff
@@ -442,6 +442,10 @@ Material* Material::clone() const
         for (const auto& technique: _techniques)
         {
             auto t = technique->clone();
+            t->_parent = material;
+            for (ssize_t i = 0; i < t->getPassCount(); i++) {
+                t->getPassByIndex(i)->_parent = t;
```
dabingnn: I would prefer to move this `for loop` code into Technique::clone(), it seems to be better.

dabingnn:
@pandamicro This pull request could be merged if Jenkins passed/

PR #13765, opened 2015-09-10
Add performance tests project.

natural-law:
1. Add a  project named performance-tests for automatically test performance.
2. Remove the performance test cases in cpp-tests project.

Now performance tests project only support platforms : iOS, Mac, Android & Win32.


pandamicro:
Related issue: https://github.com/cocos2d/cocos2d-x/issues/13481

PR #13776, opened 2015-09-11
Remove performance tests from lua-tests & js-tests.

natural-law:


natural-law:
Related issue : #13481

PR #13778, opened 2015-09-11
user dynamic_cast instead of static_cast in Scheduler

ggshily:
A node could has selector & callback scheduler at same time, so should use dynamic_cast instead of static_cast when get element from timers in Scheduler.

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -345,9 +345,9 @@ void Scheduler::unschedule(const std::string &key, void *target)
     {
         for (int i = 0; i < element->timers->num; ++i)
         {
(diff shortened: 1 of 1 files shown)


WenhaiLin:
[ci rebuild]

WenhaiLin:
[ci rebuild]

pandamicro:
@WenhaiLin is this pr ok for merge?

WenhaiLin:
@pandamicro Yes, it's OK for merge.

WenhaiLin:
The same fix: https://github.com/cocos2d/cocos2d-x/pull/12071

pandamicro:
duplicated with https://github.com/cocos2d/cocos2d-x/pull/12071
We have already merged this one, so closing it, but thanks anyway

Issue #13771, opened 2015-09-10
Chipmunk's addCollisionHandler doesn't work using JSB


vitkovskii:
1. Download cocos2d-x v3.8 final
2. `cd tests/js-tests/`
3. `cocos run -p mac` (I'm using El Capitano beta)
4. Go to "Chipmunk test" in launched app
5. Go to "Chipmunk Collision test"
6. There is no "Collision detected" text in test: http://take.ms/uKbPp

Other thoughts: 
1. It doesn't work on all native platforms because of issue in chipmunk's manual binding cpp code
2. setDefaultCollisionHandler works fine

pandamicro:
Hi, it's confirmed, we will try to fix it as soon as possible.
Temporarily, please use `setDefaultCollisionHandler` instead.

pandamicro:
It's fixed, in the PR I sent, what matters for this issue is lines like the following:

```
cpCollisionType xxx = 0;
```

And the modification of `__jsb_cpSpace_addCollisionHandler`

alankevin2:
Hi, I tested addCollisionHandler with cocos2d-x 3.8.1 engine, it still doesn't work in native app but html5 WebGL.

Is there anything should I do before I use it? 

---

Sorry for another question, I tested our game in many platforms, they all worked great as expected, but it lost so many frames in "chrome app" whether in iOS or Android OS. Especially in Android OS, the game stucked and cannot be play normally.(but the bgm still running!)

Thank you, cocos2d is awesome!

PR #13779, opened 2015-09-11
Fix Chipmunk setCollisionType issue by initializing variables

pandamicro:
setCollisionType will create a cpCollisionType(long) type variable to store a JS unsigned int value, as it's not initialized, jsval_to_uint will only modify its low bit values, then the final result will be wrong.

There are many other variables not initialized in JSB codes, I modified all that matters in chipmunk binding codes.


pandamicro:
To fix https://github.com/cocos2d/cocos2d-x/issues/13771

pandamicro:
@WenhaiLin please review

WenhaiLin:
@pandamicro It's OK  for merge.

pandamicro:
Thanks for reviewing

PR #13785, opened 2015-09-12
Add support for Unsigned with CCValue

stevetranby:
This is meant to support getting and storing unsigned values into the Value class. The instigator of adding this is to fix a bug in TMX Tiled parser where flipped/rotated GIDs are not parsed correctly as they're larger than MAX_INT.

### DIFF
--- a/cocos/base/CCValue.cpp
+++ b/cocos/base/CCValue.cpp
@@ -53,7 +53,7 @@ Value::Value(int v)
     _field.intVal = v;
 }
 
-Value::Value(unsigned v)
+Value::Value(unsigned int v)
 : _type(Type::UNSIGNED)
 {
     _field.unsignedVal = v;
@@ -275,7 +275,7 @@ Value& Value::operator= (int v)
     return *this;
 }
 
-Value& Value::operator= (unsigned v)
+Value& Value::operator= (unsigned int v)
 {
     reset(Type::UNSIGNED);
     _field.unsignedVal = v;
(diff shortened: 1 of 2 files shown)


cocos/base/CCValue.cpp
```diff
@@ -491,6 +522,50 @@ int Value::asInt() const
...
+
+    if (_type == Type::STRING)
+    {
+        // NOTE: strtoul is required (need to augment on unsupported platforms)
+        return static_cast<unsigned int>(strtoul(_field.strVal->c_str(), NULL, 10));
+    }
+
+    if (_type == Type::FLOAT)
+    {
+        return static_cast<unsigned int>(_field.floatVal);
```
minggo, line 553: Here uses `static_cast<unsigned int>`, but it uses `static_cast<unsigned>` in other places. I think they both the same thing, so it is better to use only one style.

cocos/base/CCValue.cpp
```diff
@@ -491,6 +522,50 @@ int Value::asInt() const
...
+
+    if (_type == Type::BYTE)
+    {
+        return static_cast<unsigned>(_field.byteVal);
+    }
+
+    if (_type == Type::STRING)
+    {
+        // NOTE: strtoul is required (need to augment on unsupported platforms)
+        return static_cast<unsigned int>(strtoul(_field.strVal->c_str(), NULL, 10));
```
minggo: NULL -> nullptr?

cocos/base/CCValue.cpp
stevetranby, line 553: yep, thanks.

cocos/base/CCValue.cpp
stevetranby: yep, thanks.

stevetranby:
Fixed based on notes.

Do you prefer me doing a rebase into a single commit and then force pushing, or do you recommend doing the merge to get the feature up-to-date?

Or maybe you prefer creating another pull request that includes only a single commit with all changes?

minggo:
@stevetranby Thanks. Now github is easy to squash when merging a PR.

PR #13790, opened 2015-09-13
individual kerning is wrong

MAUSMAUSgames:
`_horizontalKernings[index]` contains the amount of kerning between the letter at index and the next letter, so adding 1 to index is wrong


MAUSMAUSgames:
This bug is still in 3.10, but it's now just one line which has to change:
line 189: in CCLabelTextFormatter.cpp in `bool Label::multilineTextWrap(std::function<int(const std::u16string&, int, int)> nextTokenLen)`
from

```
nextLetterX += _horizontalKernings[letterIndex + 1];
```

to

```
nextLetterX += _horizontalKernings[letterIndex];
```

minggo:
The codes is changed so much. Please send a new PR if it already has a problem.
Thanks.

PR #13802, opened 2015-09-14
fix crash recall a cached timeline after exit scene

geron-cn:
fix crash recall a cached timeline after exit scene


geron-cn:
@pandamicro @chengstory  @jun-bing  could you check this out, thx.
[redmine issue](http://punchbox.info:3000/issues/22712)

pandamicro:
Review ok

PR #13806, opened 2015-09-14
Fix 1.6 version bone animation crash when perform getBoneAtPoint(0,0)

xiaofeng11:
Bug description at : http://punchbox.info:3000/issues/22830


jianglong0156:
@pandamicro it's ok for merge

PR #13832, opened 2015-09-15
Restructure CCDownloader implement on iOS and Android don't use libcurl.

yangws:


pandamicro:
No problem with APIs it won't break backward compatibility. We will submit it to QA team for a complete test, @yangws please give specific instructions to QA team for what, how and under which conditions to test

yangws:
related task : http://punchbox.info:3000/issues/20579
build OK, please merge it. @pandamicro

PR #13835, opened 2015-09-15
Add a feature of jumping to a specific item in list tview

neokim:
- Add `jumpToItem()` to list view.
- Refactor the layout related methods of list view to comply with `Layout`'s practice like `Layout::doLayout()`.
  - `requestDoLayout()` and `doLayout()` are overridden and have to use another dirty flag for list view layout other than `Layout::_doLayoutDirty` because `Layout::onEnter()` forces `_doLayoutDirty` to be true always. But, list view should have a way not to do layout in `onEnter()`.

### DIFF
--- a/cocos/ui/UILayout.h
+++ b/cocos/ui/UILayout.h
@@ -401,7 +401,7 @@ public:
     /**
      * request to refresh widget layout
      */
-    void requestDoLayout();
+    virtual void requestDoLayout();
     
     /**
      * @lua NA
--- a/cocos/ui/UIListView.cpp
+++ b/cocos/ui/UIListView.cpp
@@ -469,6 +469,7 @@ void ListView::refreshView()
         item->setLocalZOrder(i);
         remedyLayoutParameter(item);
     }
+    _innerContainer->forceDoLayout();
     updateInnerContainerSize();
     _refreshViewDirty = false;
 }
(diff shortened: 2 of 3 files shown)


neokim:
@zilongshanren Is it possible to remove `_doLayoutDirty = true` from `Layout::onEnter()`? Not all `Layout` should re-layout when it enters. Actually, some must not, like `ListView`.

neokim:
[ci rebuild]

cocos/ui/UIListView.h
```diff
@@ -320,6 +320,13 @@ class CC_GUI_DLL ListView : public ScrollView
     virtual void jumpToPercentHorizontal(float percent) override;
     virtual void jumpToPercentBothDirection(const Vec2& percent) override;
 
+    /**
+     * @brief Jump to specific item
```
zilongshanren, line 324: not all the parameters are documented.. `itemIndex` is missing here.

zilongshanren:
@neokim  I can't contact with the original author of ui::Layout. The logic is wired here...

I agree with you that the dirty flag is unnecessary. But in case introducing some bugs, we should be careful..

cocos/ui/UILayout.h
```diff
@@ -401,7 +401,7 @@ class CC_GUI_DLL Layout : public Widget, public LayoutProtocol
     /**
      * request to refresh widget layout
      */
-    void requestDoLayout();
+    virtual void requestDoLayout();
```
zilongshanren, line 404: please add `override` keyword here. XCode 7.0 will report warnings.

cocos/ui/UIListView.cpp
```diff
@@ -455,42 +455,33 @@ void ListView::setDirection(Direction dir)
...
+    forceDoLayout();
 }
 
-void ListView::refreshView()
+void ListView::requestDoLayout()
+{
+    _innerContainerDoLayoutDirty = true;
+}
+
+void ListView::doLayout()
```
zilongshanren, line 468: `ListView::doLayout` should call `Layout::doLayout`, right?

neokim, line 468: No. `Layout::doLayout()` is doing layout of its children. But `ListView` has only one child `innerContainer`. Actually `ListView` is not a `Layout`, but the `innerContainer` is. So `innerContainer->forceDoLayout()` is called below.

cocos/ui/UIListView.h
neokim, line 324: I'll add it.

cocos/ui/UILayout.h
neokim, line 404: You mistook it. It is not overriding. `requestDoLayout()` is defined here in `Layout` at first.

zilongshanren: @neokim  sorry.

zilongshanren, line 404: @neokim  sorry.

cocos/ui/UIListView.cpp
zilongshanren, line 468: @neokim 
The original code is 

```
void ListView::doLayout()       
{       
    Layout::doLayout();     
   refreshViewIfNecessary();        
 }
```

It does call `Layout::doLayout`, I think it's wrong.

Thanks for clarification.

cocos/ui/UILayout.h
neokim, line 404: @zilongshanren :+1:

PR #13843, opened 2015-09-16
Fixed #13551 : MenuItemImage Camera Issue.

### LINKED ISSUE #13551: MenuItemImage Camera Issue (Cocos2d-x 3.8 beta0)
Youtube Video of Issue: https://youtu.be/t7EJp3Al_mc
Github repo to reproduce issue: https://github.com/jimrange/cocos2d-x-bugs/tree/cameraTouch

The Issue:
Touching the screen on a location nowhere near a UIMenuItemImage can trigger the MenuItemImage, causing strange UI behavior that is difficult to debug. But I know know what is causing it...

There does not seem to be a way to make UI HUD elements such as the MenuItemImage only touchable by coordinates touched in the default camera. Or for specific alternative cameras.

This issue seems to be a result of the EventDispatcher::dispatchTouchEventToListeners cycling through all cameras on line 834 of CCEventDispatcher.cpp of Cocos2d-x 3.8 beta 0.

 e.g. touch the same coordinates of the MenuItemImage in the view of camera 1 and the button will be pressed. This creates a problem if world node has the same coordiantes as the hud node.

Maybe a workaround is to not let my game world coordinates overlap with my HUD coordinates? e.g. position the hud at Rect(0,0,maxHudWidth,maxHudHeigth) and the game world at Rect(maxHudWidth,maxHudHeigth,worldWidth,worldHeight)? 

Or maybe for efficiency reasons and to fix the above issue, a filter could be applied somehow to only forward events from cameras that the API user selects.

Another way of describing the issue:
The issue is that that MenuItemImage by default (all Nodes do) has the CameraFlag::DEFAULT mask. This may stop the button from being visible in camera 1. However, if the rect that is defined for the button in the default camera is touched in camera 1, an event is dispatched as if the button were pressed in the defualt camera's view.

This issue can be demonstrated by this code by touching the HelloWorld.png sprite and observe that the button is being pressed and released.

yangws:
related issue: #13551

### DIFF
--- a/cocos/base/CCEventDispatcher.cpp
+++ b/cocos/base/CCEventDispatcher.cpp
@@ -841,8 +841,13 @@ void EventDispatcher::dispatchTouchEventToListeners(EventListenerVector* listene
                 }
                 
                 Camera::_visitingCamera = camera;
+                auto cameraFlag = (unsigned short)camera->getCameraFlag();
                 for (auto& l : sceneListeners)
                 {
+                    if (0 == (l->getAssociatedNode()->getCameraMask() & cameraFlag))
+                    {
+                        continue;
+                    }
                     if (onEvent(l))
                     {
                         shouldStopPropagation = true;


pandamicro:
Thanks, review ok

PR #13850, opened 2015-09-16
调整对触摸点击的控制，增加使用中的灵活性

happyzhaole:


jianglong0156:
relate pr: https://github.com/cocos2d/cocos2d-x/pull/13886

minggo:

PR #13853, opened 2015-09-16
修复: Callback没有删除导致崩溃的问题

happyzhaole:


jianglong0156:
it's can not be merged, you can retain it in js code

minggo:

PR #13857, opened 2015-09-16
修复:帧动画无法透明渐变的问题

happyzhaole:


minggo:

PR #13858, opened 2015-09-16
动作加载优化处理用到的接口。

happyzhaole:


jianglong0156:
don't need modify c++ code, you can modify it in js code.  like:
''''
ccs.CCArmatureDataManager.prototype.isDataLoaded = function(id) {
  xxxx&&xxxx
}
''''

minggo:

Issue #13844, opened 2015-09-16
ActionTimeline triggers multiple lastFrameEvent when its FPS is lower then application FPS


c3076802:
1. create an animation in cocos-studio with fps **30**
2. run cocos-application in fps **60**
3. set **CCActionTimeline::setLastFrameCallFunc** with a print 
4. receive multiple lastFrameEvents

``` C++
void ActionTimeline::step(float delta)
{
    if (!_playing || _timelineMap.size() == 0 || _duration == 0)
    {
        return;
    }

    _time += delta * _timeSpeed;
    const float endtoffset = _time - _endFrame * _frameInternal;

    if (endtoffset < _frameInternal)
    {
        _currentFrame = (int)(_time / _frameInternal);
        stepToFrame(_currentFrame);
        if (endtoffset >= 0 && _lastFrameListener != nullptr) // last frame 
            _lastFrameListener();
    }
    else
    {
        _playing = _loop;
        if (!_playing)
        {
            _time = _endFrame * _frameInternal;
            if (_currentFrame != _endFrame)
            {
                _currentFrame = _endFrame;
                stepToFrame(_currentFrame);
                if (_lastFrameListener != nullptr)  // last frame 
                    _lastFrameListener();
            }
        }
        else
            gotoFrameAndPlay(_startFrame, _endFrame, _loop);
    }
}
```

Say we set the **ActionTimeLine's** FPS to **30**, then **_timeSpeed** will be **0.5**

``` C++
    _time += delta * _timeSpeed;
    const float endtoffset = _time - _endFrame * _frameInternal;

    if (endtoffset < _frameInternal)
    {
        _currentFrame = (int)(_time / _frameInternal);
        stepToFrame(_currentFrame);
        if (endtoffset >= 0 && _lastFrameListener != nullptr) // last frame 
            _lastFrameListener();
    }
```

So that_time increase slower than real time. 
After **the first "last frame event"** is triggered,  since the member **_playing** is not set to false, **ActionTimeline::step** is able to call again; And because **_time** increase with **0.5**  **_timeSpeed**, then **endtoffset** will be able to stay **> 0**, then **the second  "last frame event"** is triggered.

note: **_frameInternal** is set to **1/60** by default.

I think the member _playing should be set right after.

``` C++
        if (endtoffset >= 0 && _lastFrameListener != nullptr) // last frame 
            _lastFrameListener();
```

pandamicro:
[...]

PR #13859, opened 2015-09-16
fix bug: ReEntry a same frameindex after scaled action's speed.

geron-cn:
fix bug:  fix ReEntry a same frameindex when scale action's speed.


geron-cn:
reference to issue: https://github.com/cocos2d/cocos2d-x/issues/13844
@chengstory @pandamicro

PR #13861, opened 2015-09-16
for texture A8

happyzhaole:


minggo:
How can i reproduce the issue?

PR #13863, opened 2015-09-16
导出：CCArmatureDataManager::[.*AnimationData .*ArmatureData]接口到js

happyzhaole:
修复: android异步文件读取引发崩溃的问题.

### DIFF
--- a/extensions/CocoStudio/Armature/utils/CCDataReaderHelper.cpp
+++ b/extensions/CocoStudio/Armature/utils/CCDataReaderHelper.cpp
@@ -36,6 +36,9 @@ THE SOFTWARE.
 #include <cctype>
 #include <queue>
 #include <list>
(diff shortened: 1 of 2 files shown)


extensions/CocoStudio/Armature/utils/CCDataReaderHelper.cpp
```diff
@@ -150,7 +153,7 @@ NS_CC_EXT_BEGIN
 
 float s_PositionReadScale = 1;
 
-std::vector<std::string> CCDataReaderHelper::s_arrConfigFileList;
+std::vector<LoadCfgNode> CCDataReaderHelper::s_arrConfigFileList;
```
dumganhar: `LoadCfgNode` is confused, how about change it to `FileInfo`?

minggo:

PR #13865, opened 2015-09-16
新增功能:自定义多边形蒙板裁切

happyzhaole:


minggo:

PR #13868, opened 2015-09-16
修复: 按钮放大处理时显示有误的问题.

happyzhaole:


dumganhar:
@zilongshanren , could you help to review this PR? Thanks.

extensions/CocoStudio/GUI/UIWidgets/UIButton.h
```diff
@@ -171,6 +171,10 @@ class CC_EX_DLL Button : public Widget
     void setTitleFontName(const char* fontName);
     const char* getTitleFontName() const;
 
+	const std::string& getNormalFileName() const { return _normalFileName; }
```
zilongshanren, line 174: The indentation is wrong here.

extensions/CocoStudio/GUI/UIWidgets/UIButton.cpp
```diff
@@ -391,15 +391,8 @@ void Button::onPressStateChangedToNormal()
     }
     else
     {
-        if (_scale9Enabled)
```
zilongshanren, line 394: This fix will break back compatibility. We can't just delete the `updateTextureRGBA` here.

extensions/CocoStudio/GUI/UIWidgets/UIButton.cpp
```diff
@@ -424,15 +417,8 @@ void Button::onPressStateChangedToPressed()
         _buttonNormalRenderer->setVisible(true);
         _buttonClickedRenderer->setVisible(true);
         _buttonDisableRenderer->setVisible(false);
-        if (_scale9Enabled)
-        {
-            dynamic_cast<CCRGBAProtocol*>(_buttonNormalRenderer)->setColor(ccGRAY);
```
zilongshanren, line 429: Why remove this line? It will break back compatibility.

minggo:

PR #13869, opened 2015-09-16
新增一个接口,通过工具自动导出到js层,给业务层用.

happyzhaole:


minggo:

PR #13874, opened 2015-09-16
Fix shader light parameter bug

super626:
The bug is because that all the mesh share light parameter.


dabingnn:
@pandamicro This pull request could be merged.

PR #13886, opened 2015-09-17
fix the layer touch recover,  using the Scene:

jianglong0156:
add layer XX into the scene. and setTouchEnabled false, then, remove XX from the scene,but not release. at last , add XX into scene again, it will setTouchEnable true. So, this is a bug, the layer add into scene again, the touch state need to same with  it onExit


jianglong0156:
某个layer之前是设置为不响应鼠标点击的，然后把他从场景上摘除，但并不释放，然后需要的时候在重新加到场景上，如果不改的话，这个layer又可以点击了

PR #13890, opened 2015-09-17
add CCGLProgram function binding setUniformLocationWithMatrix

jianglong0156:
add CCGLProgram function binding setUniformLocationWithMatrix


jianglong0156:
relate issue:https://github.com/cocos2d/cocos2d-x/issues/13891

PR #13897, opened 2015-09-17
SkyboxBrush : Make the shader parameter take effect at once

super626:


dabingnn:

PR #13899, opened 2015-09-17
ActionObject的脚本回调处理.

happyzhaole:


minggo:

PR #13905, opened 2015-09-17
FileUtilsWin32 removeDirectory correct remove file begins with point

perminovVS:
Need check fileName . and .. for delete folders and files begin name character point.

### DIFF
--- a/cocos/platform/win32/CCFileUtils-win32.cpp
+++ b/cocos/platform/win32/CCFileUtils-win32.cpp
@@ -592,8 +592,9 @@ bool FileUtilsWin32::removeDirectory(const std::string& dirPath)
         BOOL find = true;
         while (find)
         {
-            //. ..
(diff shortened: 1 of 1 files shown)


jianglong0156:
don't need change, there are two siduation:
1、wfd.cFileName  is L"."  the judgement wfd.cFileName[0] != '.'  is false
2、 wfd.cFileName  is L".." the judgement wfd.cFileName[0] != '.'  also false

perminovVS:
@jianglong0156 have file first symbol equals poin, example .git .vs
Now spread use first point for name hide and settings folders.

Real now not remove folder and file first character point. It is bag

jianglong0156:
@perminovVS  yes. you are right, when the dir name start with ".", it can't remove. @pandamicro  the pr can be merged

pandamicro:
@perminovVS Thanks !

PR #13907, opened 2015-09-17
AssetsManagerEx unzipping freeze app. Use async.

perminovVS:
Unzipping in main thread bad idea.
Need  good solution. My code fix freezing, but need similar batch
Download.


pandamicro:
@yangws please take a look

yangws:
Reviewd OK, merge it please @pandamicro

PR #13909, opened 2015-09-18
[ci skip] testcase for s9sprite action 

ZhangMenghe:
Add this testcase after modification to s9sprite in #3103 (https://github.com/cocos2d/cocos2d-html5/pull/3103)


pandamicro:
Does it run correctly in both WebGL and Canvas mode ?

PR #13923, opened 2015-09-18
Refine components to improve physics performance

WenhaiLin:


WenhaiLin:
@minggo Please review this PR

cocos/2d/CCNode.h
```diff
@@ -1862,6 +1847,19 @@ class CC_DLL Node : public Ref
...
+    PhysicsBody* _physicsBody;
+public:
+    void setPhysicsBody(Component* physicsBody) 
+    { 
+        addComponent(physicsBody);
+    }
+    PhysicsBody* getPhysicsBody() const { return _physicsBody; }
+
+    friend class PhysicsBody;
+#endif
```
minggo, line 1861: Please don't keep these functions. It is strange that Node can get physicsbody from component and from `getPhysicsBody()`.

cocos/editor-support/cocostudio/CCComAudio.cpp
```diff
@@ -29,14 +29,17 @@ THE SOFTWARE.
 namespace cocostudio {
 
 IMPLEMENT_CLASS_COMPONENT_INFO(ComAudio)
-ComAudio::ComAudio(void)
+
+const std::string ComAudio::COMPONENT_NAME = "CCComAudio";
```
minggo: Why not add it in `Component`?

cocos/2d/CCNode.h
minggo, line 1861: Ok, it can improve performance and keep compatibility.

minggo:
This PR looks good to me.

cocos/2d/CCNode.h
WenhaiLin, line 1861: I think backward compatibility is important unless we redesigned for better physical.

PR #13935, opened 2015-09-18
use reference.

joewan:


joewan:
@zilongshanren  pls review，thks.

joewan:
[ci rebuild]

minggo:
@joewan This PR looks good to me. But i think there is not difference if the return type is not reference.

And could you please fix conflict?

PR #13936, opened 2015-09-18
AppController: synthesize window #13285

Shulepov:


Shulepov:
While it looks like a small change, but it fixes UIApplicationDelegate conformance and gives ability to get root ViewController in most correct way: _[[[[UIApplication sharedApplication] delegate] window] rootViewController]_. Especially from other libraries.

When used _[UIApplication sharedApplication].keyWindow.rootViewController_ - keyWindow can change and, for example, in-case of an active UIAlert or UIActionSheet, you get the Alert or Action Sheet as your key window.

minggo:
Thanks @Shulepov .

PR #13937, opened 2015-09-18
iOS: ARC support #12518

Shulepov:
Ability to easily turn on/off ARC support in 'Build Settings' for template project (and only for it). Cocos2d-x engine itself is not affected.

It's safe to have outer project with arc support and libraries/frameworks/depended projects with or without arc support.

### DIFF
--- a/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm
+++ b/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm
@@ -49,7 +49,7 @@ static AppDelegate s_sharedApplication;
 
     // Init the CCEAGLView
     CCEAGLView *eaglView = [CCEAGLView viewWithFrame: [window bounds]
-                                         pixelFormat: (NSString*)cocos2d::GLViewImpl::_pixelFormat
+                                         pixelFormat: (__bridge NSString *)cocos2d::GLViewImpl::_pixelFormat
                                          depthFormat: cocos2d::GLViewImpl::_depthFormat
                                   preserveBackbuffer: NO
                                           sharegroup: nil
@@ -81,7 +81,7 @@ static AppDelegate s_sharedApplication;
     [[UIApplication sharedApplication] setStatusBarHidden:true];
 
     // IMPORTANT: Setting the GLView should be done after creating the RootViewController
-    cocos2d::GLView *glview = cocos2d::GLViewImpl::createWithEAGLView(eaglView);
+    cocos2d::GLView *glview = cocos2d::GLViewImpl::createWithEAGLView((__bridge void *)eaglView);
     cocos2d::Director::getInstance()->setOpenGLView(glview);
 
     app->run();
@@ -140,10 +140,13 @@ static AppDelegate s_sharedApplication;
 }
 
 
+#if __has_feature(objc_arc)
+#else
 - (void)dealloc {
     [window release];
     [super dealloc];
 }
+#endif
 
 
 @end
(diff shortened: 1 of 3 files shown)


Shulepov:
It doesn't enable ARC in cocos2dx_lib project, it just gives ability to easily enable/disable ARC support for main project - for easier adding of different libraries and writing support code like notifications / social networks / analytics. So while it enables ARC only for outer project it will not bring any performance penalties - as core cocos2dx_lib project will still have manual memory management.

So, it's just about **ability to easily switch** from non-arc to arc, nothing more. All stuff are compile time so non-arc users will not see any changes.

minggo:
@Shulepov I think it will affect cocos2d-x because cocos2d-x itself also use some objective-c codes, and it doesn't use ARC.

Shulepov:
I mean that it enables ARC only for template project, not for inner cocos2dx_lib project (it still will use manual memory management. It's okay to have dependencies with another memory management style.)

Template project only contains next files:
- AppController {.h, .mm}
- RootViewController {.h, .mm}
- main.m

And i have updated them to work correctly with and without ARC. 
Not sure how SDKBOX connected, if just by copying files - then yes - that will not work.

Anyway, it not enables ARC, it just gives ability to enable ARC just by changing **no** to **yes** in BUILD_SETTINGS (for default iOS template project).

But if i have missed something, i can close that PR :) Actually, it's not very important, just for convenience in some cases.

stevetranby:
I believe @autoreleasepool {} block is mutually exclusive from ARC. I believe that it is supported for non-ARC projects as well. 

See: https://developer.apple.com/library/ios/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html

minggo:
@stevetranby thanks for the link.
@Shulepov is this PR still need?

stevetranby:
@minggo I think it's smart to allow this and merge, except could get rid of the NSAutoReleasePool #else in main.cpp based on my previous comment.

I also think you should enable ARC going forward. The iOS/OSX code is small and thus isn't really necessary, but it is a best practice going forward and might as well allow people to enable it in their version of the engine? Just a thought

minggo:
@stevetranby If i understand correctly, if ARC is enabled, then `@autoreleasepool{}` and this PR is not needed, right?

stevetranby:
This PR does a little more than #ifdef the NSAutoReleasePool by adding a few qualifiers for ARC allowing for both ARC and Non-ARC builds. I haven't tested it, however.

templates/cpp-template-default/proj.ios_mac/ios/AppController.mm
```diff
@@ -49,7 +49,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     // Init the CCEAGLView
     CCEAGLView *eaglView = [CCEAGLView viewWithFrame: [window bounds]
-                                         pixelFormat: (NSString*)cocos2d::GLViewImpl::_pixelFormat
+                                         pixelFormat: (__bridge NSString *)cocos2d::GLViewImpl::_pixelFormat
```
minggo, line 52: What's `__bridge`?

templates/cpp-template-default/proj.ios_mac/ios/main.m
```diff
@@ -1,9 +1,15 @@
 #import <UIKit/UIKit.h>
 
 int main(int argc, char *argv[]) {
-    
+    NSString *appControllerClassName = @"AppController";
+#if __has_feature(objc_arc)
+    @autoreleasepool {
```
minggo, line 4: If ARC is enabled, does `@autoreleasepool` needed?

Shulepov:
Hi, will check today `@autoreleasepool` in non arc project, looks like it should be okay. 
`__bridge` keywords are required for ARC when casting from `void *` to Objective-C types. Simple `__bridge` used when there is no need to transfer ownership to ARC. It has no effect when not using ARC.

@minggo, yep, `@autoreleasepool` needed in both cases.

PR #13950, opened 2015-09-19
Added CC_PROPERTY const qualifier

Almax27:
CC_PROPERTY macro is inconsistent with others, get methods should be const.


minggo:
Thanks.

PR #13957, opened 2015-09-20
Make Ref _referenceCounter thread-safe

TheCodez:
This will add a performance overhead, as atomic operations are more expensive then just normal operations


TheCodez:
Apparently it only works on windows

PR #13960, opened 2015-09-21
Add jsb.fileUtils.writeDataToFile()

TimothyZhang:
and also with a testcase.


pandamicro:
Thanks for the completion, we will test it

PR #13969, opened 2015-09-21
fix the bug:when the frames play over, the effect in  frame is not disappear

jianglong0156:
fix the bug:when the frames play over, the effect in  frame is not disappear


jianglong0156:
relate pr:https://github.com/cocos2d/cocos2d-x/pull/13854

PR #13976, opened 2015-09-21
Update CCPhysics3DComponent.cpp

xifengxi:
there should save old bool.


lsybblll:
The PhysicsObject can't be removed.

xifengxi:
@zilongshanren The PhysicsObject can't be removed.

pandamicro:
Good catch, thanks for the fix. @super626

PR #13982, opened 2015-09-21
Update CCMenu.cpp

xifengxi:
I think if the menu in 2D view, it only be triggered by 2D camera. 3D is also the same.


pandamicro:
Already been fixed at https://github.com/cocos2d/cocos2d-x/pull/13843, Thanks anyway

PR #13999, opened 2015-09-22
Fix crash (integer overflow) in Device::getTextureDataForText (ios)

igorzel:
## Explanation

There was the problem when _height_ and _width_ were typecast to **short**
integer. So for values bigger than **sizeof(short)** we had integer overflow. Values may became negative.
## Steps to reproduce:

Create simple application with **cocos2d::ui::EditBox**. Start it. Add 1000 (just
example) characters at a time to the edit box. Method
**Device::getTextureDataForText** will be called and there will be integer
overflow for _height_ and/or _width_:

screen shot 2015-09-22 at 16 00 34
As a result application will crash later.
## Notes

This fix is related to platform (IOS) dependent code! So it is in _CCDevice-ios.mm_.

### DIFF
--- a/cocos/platform/ios/CCDevice-ios.mm
+++ b/cocos/platform/ios/CCDevice-ios.mm
@@ -485,8 +485,8 @@ Data Device::getTextureDataForText(const char * text, const FontDefinition& text
         {
             break;
         }
-        height = (short)info.height;
-        width = (short)info.width;
+        height = info.height;
+        width = info.width;
         ret.fastSet(info.data,width * height * 4);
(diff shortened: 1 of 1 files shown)


igorzel:
Probably mac/ios problem is caused by this:
http://stackoverflow.com/questions/26197347/agreeing-to-the-xcode-ios-license-requires-admin-privileges-please-re-run-as-r

igorzel:
[ci rebuild]

pandamicro:
Thanks for your contribution

Issue #13919, opened 2015-09-18
[Web Engine] Fix setColor and Opacity in uiButton


ZhangMenghe:
- Should not setOpacity while setColor.
- There exists a mechanism in the Engine that when visit the protected nodes attached to button, `_synStatus` will be called, in which color and opacity will be changed as button did. 
- Avoid set repetitive attributes.

PR #14007, opened 2015-09-24
[ci skip] button Opacity setting testcase

ZhangMenghe:
TestCase for #13919


pandamicro:
Thanks, please add a redmine task and assign to Shanshan

PR #14008, opened 2015-09-24
Fix loading pluginx lib when compile Android with --compile-script flag

thuydx55:
Fix bug if compile Android build with `--compile-script` flag then it cannot load file `jsb_pluginx.js` because it renamed to `jsb_pluginx.jsc`


pandamicro:
Thanks for the fix

PR #14024, opened 2015-09-26
Fix : RefPtr test wasn't executed even in debug build 

TheCodez:


TheCodez:
[ci rebuild]

pandamicro:
Thanks for the fix

PR #14026, opened 2015-09-27
Make Ref _referenceCounter thread-safe

TheCodez:
This will add a performance overhead, as atomic operations are more expensive then just normal operations

### DIFF
--- a/cocos/base/CCRef.cpp
+++ b/cocos/base/CCRef.cpp
@@ -40,9 +40,8 @@ static void untrackRef(Ref* ref);
 #endif
 
 Ref::Ref()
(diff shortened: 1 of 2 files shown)


pandamicro:
Do we need Ref to be thread safe ?

I mean, Cocos2d-x doesn't support multithread usage for now, all cocos APIs must be used in the GL thread, otherwise we can't guarantee it work correctly. 

We have plan to support multi thread, but when we do this kind of work, it should be done globally. @minggo @yangws What do you guys think ?

TheCodez:
If this is considered for 4.0, i think making Ref thread-safe is the first step towards multithreading.

PR #14032, opened 2015-09-27
Remove scale factor for label shadow

danutchereches:
Scale factor is not needed here.
It just creates differences between devices.


danutchereches:
I messed up something.

PR #14049, opened 2015-09-29
Extending SpriteFrameCache to read PolygonInfo from atlas plist file

CodeAndWeb:
Extended SpriteFrameCache to read polygon info from atlas .plist file.
The polygon information is added to a sprite instead of quads when creating a sprite from a named sprite frame.
Polygon creation can be done offline using TexturePacker 4.0.0.
Works with auto batching.


ricardoquesada:
PR looks good to me.

cocos/2d/CCSpriteFrameCache.h
```diff
@@ -223,6 +249,16 @@ class CC_DLL SpriteFrameCache : public Ref
     */
     void removeSpriteFramesFromDictionary(ValueMap& dictionary);
 
+    /** Parses list of space-separated integers */
+    void parseIntegerList(const std::string string, std::vector<int> &res);
```
zilongshanren: `const std::string string` should be `const std::string& string`?

cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -79,6 +80,63 @@ SpriteFrameCache::~SpriteFrameCache()
...
+                                             const std::vector<int> &verticesUV,
+                                             const std::vector<int> &triangleIndices,
+                                             PolygonInfo &info)
+{
+    size_t vertexCount = vertices.size();
+    size_t indexCount = triangleIndices.size();
+    
+    float scaleFactor = CC_CONTENT_SCALE_FACTOR();
+
+    V3F_C4B_T2F *vertexData = new V3F_C4B_T2F[vertexCount];
```
zilongshanren: when does the structure memory of `vertexData` got released? I think it's the polygonInfo's responsibility to do the house clean work. So we should set `_isOwner` to true inside polygon info, thoughts?

zilongshanren:
@pandamicro  
I'm OK with this PR now.

zilongshanren:
[ci rebuild]

pandamicro:
@CodeAndWeb  Thanks for supporting polygon sprite in TexturePacker and thanks for bringing this to cocos2d too.

CodeAndWeb:
@pandamicro Thanks for the merge.

PR #14051, opened 2015-09-30
filter 3D Touch force event at touchesMoved.

Naruto:
The force of the touch event invokes touchesMoved in iPhone 6s/6s+.
So, single tap invokes touchesBegan and touchesMoved. 

This PR filters events that doesn't move x and y coordinates since prev event. Because of filtering force  touch event.


Naruto:
https://github.com/cocos2d/cocos2d-x/commit/bb832f8d60d8e4759c22679ba30d7851e0794313

v3.10 is considering 3D touch. this PR is finished role.

PR #14052, opened 2015-09-30
Fix an issue that list view's Magnetic::CENTER is not working well when non-bounceable.

neokim:
Centric magnetic list view is not working well around boundaries if it is non-bounceable. This is a fix for it.


pandamicro:
@liamcindy Please review this PR

zilongshanren:
@pandamicro  review done.
@neokim  thanks.

PR #14054, opened 2015-09-30
fix: web view crash in ios 6.x

Dimon4eg:
Web view crashes on iOS 6.x devices because frame is zero.


pandamicro:
@liamcindy Please help review this PR
@zilongshanren

zilongshanren:
@Dimon4eg How to reproduce this issue?

Dimon4eg:
@zilongshanren 
It's always crash on iPhone 4.
My fix is workaround. Maybe you will find more accurate fix.

minggo:
Close it since iOS 6 is not supported now.

PR #14055, opened 2015-09-30
fix crash in LoadingBar

Dimon4eg:
fix crash in LoadingBar if no sprite.


pandamicro:
In which circumstances the innerSprite will be absent ?

Dimon4eg:
If no image

pandamicro:
Thanks, review done

PR #14056, opened 2015-09-30
fix crash if csb file is broken

Dimon4eg:


pandamicro:
@xiaofeng11 Please review this PR

xiaofeng11:
@Dimon4eg I think this pr is fix program crash when csb file is broken, but not csb file missing, please fix the title of this pr, thanks.

pandamicro:
Done, @xiaofeng11 is this a proper fix ?

xiaofeng11:
@pandamicro Yes, If there's nothing have been readout from the csb file, We won't need parse a blank data buffer, It will cause access error.

PR #14058, opened 2015-09-30
Multi-window implementation

DerekFromAmaya:
WindowKey is used as a multiplatform identifier for windows.  For now it's a hash of the GLFWwindow pointer, but that can be changed.

DirectorContext encapsulates a "mini director" instance for each window.  There is still a central director which is responsible for updating each context.  Each context also has its own event manager, action manager, and scheduler.

To reduce the amount of code changes necessary for mobile, by default the first window created is the main window.  The main window is special; if it gets closed, it terminates the application.  There are special accessors to grab the "main"scheduler/eventmanager/actionmanager.  For certain tasks, especially async tasks, they need to schedule their callbacks on the main window, because the existence of the other windows is not guaranteed, unlike the main one.

The contexts work in a similar manner as opengl.  A context is set to be active (setContextAndLock) and then non-actve (unlock).  Doing so changes what gets returned by the getCurrentWindow call and the getCurrentScheduler/EventManager/ActionManager calls.  This allows nodes and virtually everything to not care which window it was created in.  Unless the node is transplanted into a new window/context, its getCurrentManagers functions will always return the same.  It also hides any window dependancy.  Almost everything outside of the director doens't know or care which window it was created in.

CCGLViewImpl-desktop.cpp required substantial changes.  The poll events function grabs all the events for all windows at once.  What we want to do is sort them by window and defer the execution of them until the correct context is active.  Otherwise, we'll be switching contexts rapidly while processing the polled events.  Thus, the callbacks all push std::functions onto a map, keyed to the correct window.

There are also a number of order of operation fixes and smaller miscelaneous changes that were required as a result of this feature.

cpp-empty-test has been extended to create 3 windows at launch to demonstrate the multi window functionality.


Mazyod:
The objc code needs some cleanups .. The properties are defined with `_` prefix, then there is a backing variable with a single `_`, but it should be double. Then again, the whole convention is a bad idea for objc. We should limit manual variable declarations, whether in the class scope, or using `@synthesize`.

e.g.: Below, the class will have `__variable` and `_variable` backing stores.

``` objc
@interface XYZ ()

@property (nonatomic, retain) id _variable;

@end

@implementation XYZ {
    id _variable;
}

@end
```

@zilongshanren

DerekFromAmaya:
I'll notify our Mac developer to clean up the objc portion.  There's also a lot of build errors that did not occur in my build that I'll need to fix.

Mazyod:
@DerekFromAmaya Thanks for your contributions! I can also help with the objc code, so don't worry if it's not perfect. The current implementation needs work anyway. That's why I want to hear @zilongshanren's opinion

DerekFromAmaya:
There's a bunch of test cpp files that I never built (most of my testing was with our own custom client and cpp-tests/cpp-empty-test where mainly there as sanity checks to make sure i/o and rendering still worked :)

I'm fixing up the compile errors for the test files and I'll resubmit another pull request later on.  

The most important changes are in Director.cpp and GLViewImpl-desktop.cpp.  If you are willing to compile the pull request locally on a windows machine, please give cpp-empty-test a run to see the multi-windows.

Mazyod:
I would love to, except I only run Mac OS, lol.

DerekFromAmaya:
Continued in https://github.com/cocos2d/cocos2d-x/pull/14060

DerekFromAmaya:
Rebased to master and deleted the remote branch instead of force pushing (d'oh).  This is why this pull request got closed and it's continued in 14060

DerekFromAmaya:
I'm having troubles reproducing the build error that jenkins is spitting out.

```
   "f:\win32_slave\workspace\ccs-pr\node\win32_bak\build\cocos2d-win32.sln" (Build target) (1) ->
   "f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\js-bindings\proj.win32\libjscocos2d.vcxproj" (default target) (11) ->
   (ClCompile target) -> 
     ..\auto\jsb_cocos2dx_auto.cpp(7): fatal error C1083: Cannot open include file: 'CCComponentJS.h': No such file or directory [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\js-bindings\proj.win32\libjscocos2d.vcxproj]


   "f:\win32_slave\workspace\ccs-pr\node\win32_bak\build\cocos2d-win32.sln" (Build target) (1) ->
   "f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\lua-bindings\proj.win32\libluacocos2d.vcxproj" (default target) (5) ->
     ..\auto\lua_cocos2dx_auto.cpp(12456): error C2664: 'void object_to_luaval<cocos2d::WindowKey>(lua_State *,const char *,T *)' : cannot convert argument 3 from 'cocos2d::WindowKey' to 'cocos2d::WindowKey *' [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\lua-bindings\proj.win32\libluacocos2d.vcxproj]
     ..\auto\lua_cocos2dx_auto.cpp(16620): error C2065: 'Projection' : undeclared identifier [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\lua-bindings\proj.win32\libluacocos2d.vcxproj]
     ..\auto\lua_cocos2dx_auto.cpp(16620): error C2146: syntax error : missing ';' before identifier 'arg0' [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\lua-bindings\proj.win32\libluacocos2d.vcxproj]
     ..\auto\lua_cocos2dx_auto.cpp(16620): error C2065: 'arg0' : undeclared identifier [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\lua-bindings\proj.win32\libluacocos2d.vcxproj]
     ..\auto\lua_cocos2dx_auto.cpp(16622): error C2065: 'arg0' : undeclared identifier [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\lua-bindings\proj.win32\libluacocos2d.vcxproj]
     ..\auto\lua_cocos2dx_auto.cpp(16628): error C2065: 'arg0' : undeclared identifier [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\scripting\lua-bindings\proj.win32\libluacocos2d.vcxproj]
```

These files compile just fine on my machine, as does the cocos2d-win32 solution.  Are they not getting cleaned properly?

Would you be able to get one of your engineers that uses windows to take a look at this problem?

(Also, could we continue discussion in the new pull-request; https://github.com/cocos2d/cocos2d-x/pull/14060 )

DerekFromAmaya:
It looks like those might be auto generated files build by test.bat in the bindings-generator directory.  If that's the case, it requires an android NDK to also be installed, something I don't currently have.  

Apart from the autogenerated js/lua stuff, cpp-empty-test should work.  cpp-test works too, but since that test is basically a giant singleton, it's not really compatible with multi-windows anyways.

It'd be great if I could get some feedback on this.

PR #14061, opened 2015-10-01
Added ability to set default min/magnification filtering 

Super-Genius:
for 2D Textures and Font atlas.

The default min/magnification settings for rendering textures was GL_LINEAR, which blurred textures.  This allows the user to set the default filtering when creating textures.


pandamicro:
@xiaofeng11 @liamcindy Please review this PR, @WenhaiLin please also confirm

xiaofeng11:
Logic of these code is OK, but I'm not familiar with OpenGL programming, @liamcindy @WenhaiLin please confirm does these setting change will change the render effect of follow-up operations

cocos/2d/CCFontFreeType.cpp
```diff
@@ -38,6 +39,8 @@ FT_Library FontFreeType::_FTlibrary;
 bool       FontFreeType::_FTInitialized = false;
 const int  FontFreeType::DistanceMapSpread = 3;
 
+int FontFreeType::_defaultFontDPI;
```
WenhaiLin: Please specify the initial value of `FontFreeType::_defaultFontDPI`.

minggo:
This PR includes so many unneeded commits. So i will close it.

minggo:
@klhurley Could you please send a new PR?

PR #14062, opened 2015-10-02
Add missing override keyword

mogemimi:
This fixes the following clang warnings when building using Xcode 7.

```
UIEditBoxImpl-common.h:53:18: 'initWithSize' overrides a member function but is not marked 'override'
UIEditBoxImpl-common.h:55:18: 'setFont' overrides a member function but is not marked 'override'
...
```


pandamicro:
Thanks for your contribution !

Issue #13983, opened 2015-09-21
Should Node::getName() return the string by reference?


AIGRIND-LLC:
At this moment `Node::getName()` returns node's name by the value.
It would be more effecient to declare and implement this method in the folllowing way:

``` cpp
// declaration
virtual const std::string & getName() const;

// implementation
const std::string & Node::getName() const
{
    return _name;
}
```

PR #14075, opened 2015-10-05
Node::getName() now returns string by reference

TheCodez:
https://github.com/cocos2d/cocos2d-x/issues/13983


pandamicro:
Thanks for the improvement

Issue #14050, opened 2015-09-30
ActionManager memory leak


SongCF:
when I create a node with an action, and not add to any other node, such as:

```
var sp = cc.Sprite.create("text.png");
sp.runAction(cc.fadeOut(10));
//has not parent
```

This will lead to memory leak, in correctly, 'sp' will be released in next frame, but when it run an action, 'sp' will not be released forever. 

//CCActionManager.cpp/addAction

```
void ActionManager::addAction(Action *action, Node *target, bool paused)
{
    //...
    HASH_FIND_PTR(_targets, &tmp, element);
    if (! element)
    {
        element = (tHashElement*)calloc(sizeof(*element), 1);
        element->paused = paused;

 // this code retain the 'sp', but released when action remove. if 'sp' has not parent, it not be running, so it's action has not been run
        target->retain();
        element->target = target;
        HASH_ADD_PTR(_targets, target, element);
    }
```

we solved by this code:

```
void ActionManager::update(float dt)
{
        // ...
        if (_currentTargetSalvaged && _currentTarget->actions->num == 0)
        {
            deleteHashElement(_currentTarget);
       }
//in next frame, if 'sp' refrence count == 1, remove it
//if some node refrence 'sp', it's refrence count >= 2
       else if (_currentTarget->target->getReferenceCount() == 1)
       {
        deleteHashElement(_currentTarget);
       }
}
```

zilongshanren:
@SongCF 
Thanks, could you open a PR?

ricardoquesada:
closed, since PR #14090 was already merged

PR #14090, opened 2015-10-09
issues #14050 ActionManager memory leak

SongCF:
fixed ActionManager memory leak
when I create a node with an action, and not add to any other node, such as:

```
var sp = cc.Sprite.create("text.png");
sp.runAction(cc.fadeOut(10));
//has not parent
```

This will lead to memory leak, in correctly, 'sp' will be released in next frame, but when it run an action, 'sp' will not be released forever.

### DIFF
--- a/cocos/2d/CCActionManager.cpp
+++ b/cocos/2d/CCActionManager.cpp
@@ -450,6 +450,11 @@ void ActionManager::update(float dt)
         {
             deleteHashElement(_currentTarget);
         }
+        //if some node refrence 'target', it's refrence count >= 2 (issues #14050)
+        else if (_currentTarget->target->getReferenceCount() == 1)
+        {
+            deleteHashElement(_currentTarget);
+        }
     }
 
     // issue #635


pandamicro:
The real problem here is that Reference Counting is not perfect, it can't solve circular referencing problem. For the given issue, it may be easy to figure out a solution, but in general the problem persist and it's hard to find a good way for solving it. 

For now, we rely on developers to do the right thing, I'm not sure if it's good to merge this PR, @minggo thoughts ?

minggo:
If Sprite is not added into any node, then it can not be rendered, which means it is not visible. Then why create such a Sprite and run an action?

SongCF:
@minggo 
dynamic get a layout content size, such as get table cell size from ccbi(cocosbuilder)/csd(cocostudio), but not added into any node.
This will lead to memory leak when ccbi/csd has timeline.

minggo:
@SongCF 
Sorry, i don't understand why you need to run an action to get cell size. Could you please describe it in more detail?

Thanks.

SongCF:
@minggo 
when I create a tableview, I don't know the tablecell size in my code, I get cell size form ccbi(cocosbuilder layout file):

```
var tempItem, cellSize;
tempItem = cc.BuilderReader.load("ccbiFile.ccbi");
cellSize = tempItem.rootNode.getContentSize();
```

but the ccbi had TimeLine(some actions), and I just only get the size,not added into any node.

minggo:
So the `tempItem` is the sprite, and it is created and run an action, but is not added into any node?

SongCF:
@minggo 
yeah

minggo:
Thanks @SongCF

PR #14093, opened 2015-10-09
Fixed AudioPlayer crash in Win32

cpascal:
Environment: Win32

Because cocos2d::experimental::AudioPlayer copy constructor doesn't initialize _readForRemove property, _readForRemove value is not false. So AudioEngineImpl::update() erases this AudioPlayer immediate during the audio is playing. So that makes crash at the AudioPlayer::rotateBufferThread().

※ AudioPlayer is inserted into AudioEngineImpl::_audioPlayers by the copy constructor at AudioEngineImple::play2d().


pandamicro:
@WenhaiLin Please review

WenhaiLin:
@cpascal Thank you. @pandamicro It's OK for merge.

PR #14097, opened 2015-10-09
Add missing getChildByTag<>() api

TheCodez:


pandamicro:
Thanks for noticing the API inconsistency

Issue #14070, opened 2015-10-05
CMake does not check for sqlite3 dependency


firefly2442:
It seems that it's possible to run CMake and generate a solution without having the sqlite dependency (libsqlite3-dev on Linux) installed.  This results in the code not being able to compile because it's missing the appropriate include headers.

tested version hash: b38eec4d470d3ed7cf3062c4cad5f821ebc352d5
OS: Linux

zilongshanren:
@firefly2442  
Could you send a PR to resolve it?
Thanks.

firefly2442:
Please test, I haven't worked with CMake in some time.

zilongshanren:
@firefly2442 
How could I test your code?

What kind of Linux distribution should I have?
I have Ubuntu, but it doesn't complain about the sqlite3 dependency.

firefly2442:
@zilongshanren try this:

On Ubuntu, uninstall the `libsqlite3-dev` package.  This package contains the header file for sqlite.  Delete the CMake cache to clear out any old settings (or just delete your build folder contents).  By not having this sqlite dependency, CMake should fail and give an error but it doesn't.  CMake completes successfully and generates a solution.  Then, when you try to build Cocos2d-x, it will fail saying it's missing the sqlite header file.

I believe CMake should recognize if sqlite is missing and provide an appropriate warning/error message to prevent the solution from being generated.

ricardoquesada:
thanks. fixed.

PR #14098, opened 2015-10-10
add FindSQLite3.cmake under Linux, fixes: #14070

firefly2442:


firefly2442:
Also, I'm not sure what influence this would have on the `cocos` console helper script.  I ended up having to set the `CMAKE_MODULE_PATH` for the generated project in order for it to have access to the Find module.

ricardoquesada:
thanks. merged in #15984

PR #14106, opened 2015-10-10
TMXLayer cast bug

danutchereches:
When using float values (for example the actual position of the character) to get the current tile, the wrong tile is sometimes/usually returned, because the index is not calculated properly.


pandamicro:
Thanks for the contribution !

PR #14108, opened 2015-10-10
Changed label font size type to float

danutchereches:
Very useful for small (design) resolutions.


WenhaiLin:
@pandamicro I think this PR is OK for merge. @loadrunner Thank you!

PR #14109, opened 2015-10-10
Added line spacing/leading feature to Label

danutchereches:
Not sure is this is something generally useful or if it can be done in any other way, but I needed the feature and I couldn't find anything already implemented for it.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -446,6 +446,7 @@ void Label::reset()
     }
     _additionalKerning = 0.f;
     _lineHeight = 0.f;
+    _lineSpacing = 0.f;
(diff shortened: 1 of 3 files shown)


pandamicro:
I think it should have been supported by line height, @WenhaiLin does line height do the same trick ?

danutchereches:
Not exactly the same thing. (see preview)
Also, this is probably a bug, but line height seems to ignore the content scale factor.
(I use a 120x160 design resolution with scale factor 6 in this sample, so 80px line height is obviously wrong)

screenshot from 2015-10-11 14 13 08

WenhaiLin:
@loadrunner Great, thank you! @pandamicro Line spacing not as thick as line height, i think this PR is OK  for merge.

Issue #14122, opened 2015-10-13
BAD_EXCEPTION crash in CCSpriteBatchNode::removeAllChildrenWithCleanup


jgod:
I've always seen `removeAllChildren()` as a very safe api to call, even if the node had already been processed as destroyed by cocos (like when leaving a scene) and you had pointers to it hanging around for some reason.  Usually you can call it with no negative consequences.

But `CCSpriteBatchNode` doesn't behave the same as the other Node types here.
You get a bad_exception on [this line](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/2d/CCSpriteBatchNode.cpp#L243) because `_textureAtlas` is null. 

I think this at least deserves an assert, but would checking for null be so bad? The check shouldn't effect performance and it makes it a lot safer.

pandamicro:
I agree with you, but I want to know more specifically about how I can trigger the issue

PR #14123, opened 2015-10-13
Check that textureAtlas isn't null before removing quads in SpriteBatchNode

jgod:
See issue #14122.


pandamicro:
I'm ok with the modification, but I also want to know in what circumstances that you met `_textureAtlas` will be null while removing ?

minggo:
The code looks good to me. As SpriteBatchNode:: removeAllChildrenWithCleanup() is a public function, it  may be invoked at anytime.

PR #14129, opened 2015-10-13
Fix UIPageViewVerticalTest in cpp-test can not scroll back to page 1

xiaofeng11:


pandamicro:
Can you squash to one commit ? Three commits for one line is a bit wasted

xiaofeng11:
OK, I'll submit a new pr

PR #14134, opened 2015-10-14
fix the #23183 bug

songchengjiang:


pandamicro:
Why this can solve the problem ? Is there any other node have been set a global z order and masked the red line ?

PR #14136, opened 2015-10-14
fix lua Scene3DTest "back" button not working issue

songchengjiang:


pandamicro:
Please describe the bug more specifically, otherwise we can't review it

songchengjiang:
In lua Scene3DTest, the "back" button does not work and redundancy, we can use default "MainMenu" button instead.

PR #14140, opened 2015-10-14
Bug fix

xiaofeng11:
Fix when image size of variable button status texture are different, Helper::restrictCapInsetRect result may cause pressed & disabled status picture won't display.


pandamicro:
Why the texture size can be negative?

xiaofeng11:
No texture size hasn't be negative, the bug is wrong scale9 splice cause picture won't show

pandamicro:
Sorry I mistakenly seen width as 0

PR #14144, opened 2015-10-15
Remove screen direction check in simulator to avoid render error

xiaofeng11:
Simulator set to landscape screen direction in default, when create a portrait scene in CocoStudio, the simulator try to set default scene size as landscape, it will make scene display in stretch mode can not fill the window.

### DIFF
--- a/tools/simulator/config.json
+++ b/tools/simulator/config.json
@@ -1,6 +1,5 @@
 {
     "init_cfg":{
-       "isLandscape": true,
        "isWindowTop": false,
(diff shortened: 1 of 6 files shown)


pandamicro:
So after removing these code, how the windows change to right orientation?

xiaofeng11:
After remove these code, there's won't have orientation any more, the windows size is determine by user param from command line or default(960x640). CocoStudio will always send windows size to simulator in command line param, simulator won't need check window size.

PR #14146, opened 2015-10-15
Fix terrain render bug

### LINKED ISSUE #14145: Terrain render bug on some devices
Terrain can not be rendered correctly on MeiZu MX3 and IPOD5.

super626:
correct wrong precision, which leads to failed on some devices
fix #14145


super626:
@pandamicro

pandamicro:
Complete for https://github.com/cocos2d/cocos2d-x/issues/14145

PR #14153, opened 2015-10-15
PMA Handling issue

Fraggle:
Hi,

I compared to cocos2D-x v2.x and it seems that there is an issue with the _hasPremultipliedAlpha flag not beeing set correctly for mipmaps and compressed textures.
I guess it's a mistake due to the reorganization of the code between the old CCTexture2D and the new one.
In our tests, this fixes the issue.

Best,

Sebastien


pandamicro:
@Fraggle Can you add a test case to reproduce the wrong effect ?
@xiaofeng11 please review this PR

xiaofeng11:
@super626 please review this PR

super626:
@Fraggle 
Thanks.

@pandamicro @xiaofeng11 It could be merged.

PR #14154, opened 2015-10-15
Fix crash in web view on android

Dimon4eg:
Reason of crash - shouldStartLoading notification is executed from Java's thread.

I redirected calls to cocos thread. Now all looks fine ;)


pandamicro:
@liamcindy Please review this PR

zilongshanren:
@Dimon4eg  I think only `shouldStartLoading` should be in cocos thread.
Other methods are already run in cocos thread.

zilongshanren:
@Dimon4eg 

```
 activity.runOnGLThread(new Runnable() {
                @Override
                public void run() {
                    Cocos2dxWebViewHelper._didFinishLoading(mViewTag, url);
                }
            });
```

As we see from the above code snippets the `_didFinishLoading` is wrapped with `runOnGLThread` which is also known as the cocos thread.

cocos/ui/UIWebViewImpl-android.cpp
```diff
@@ -401,45 +404,73 @@ namespace cocos2d {
...
-                return true;
+
+                    m.lock();
+                    finished = true;
+                    m.unlock();
+                    cv.notify_one();
+                });
+
+                // wait for result from cocos thread
+                std::unique_lock<std::mutex> lk(m);
```
zilongshanren: This may block Android UI thread if the developers do some heavy things in Cocos thread.
And I have added some comments to the `shouldStartLoading` declaration. Don't do any OpenGL related operations here. So if you improve this method, I think it's better to also improve the comments. Thanks.

Dimon4eg:
@zilongshanren 
Yes, you're right.
In this case maybe better to fix the issue in java code only for didFinishLoading?
Because it's not clear in cpp in which thread a function is executed. Currently we have mix here.
Ok, I will add comments.

Dimon4eg:
Note: in cocos 3.8.1 all notifications are executed in java's thread.

zilongshanren:
@Dimon4eg yeah, v3.8.1 has bugs.
But you are sending a patch to v3 which is v3.9 version.
So you might need to remove all the fix except shouldStartLoading.

Dimon4eg:
@zilongshanren I think it's better to fix in java code. No?

zilongshanren:
@Dimon4eg 
We can't fix it in Java code, because the the method in Runnable class can't return value, only void is allowed.
That's why I leave the method running still in Android UI thread. (Maybe I'm wrong here?)

And we could avoid this crash by not doing any OpenGL related operations in this callback.

Dimon4eg:
@zilongshanren 
I think not only OpenGL related operations lead to crash.
My code is also single-threaded as cocos and it's dangerous to call it from java's thread.

Dimon4eg:
@zilongshanren I fixed it on java side. Please review.

zilongshanren:
@Dimon4eg  :+1: 
Thanks.  It's great.  @pandamicro  I'm OK with this PR now.

pandamicro:
Thanks guys

Issue #14113, opened 2015-10-12
Memory leak from CCSpriteFrameCacheHelper.cpp


fuyifan:
In SpriteFrameCacheHelper's retainSpariteFrames function, it will retain every SpriteFrame defined in .plist file. However, when I call the purge function, the destructor of SpriteFrameCacheHelper does not call release for each SpriteFrame. I find that Cocos defines SpriteFrameCacheHelper::releaseSpriteFrames(const std::string &plistPath), however, this function dosen't been called for each file in _usingSpriteFrames when destroyed.

PR #14159, opened 2015-10-16
fix the issue #14113

fuyifan:


pandamicro:
@xiaofeng11 Please review this PR
@fuyifan Can you squash you commits into one ?

xiaofeng11:
@fuyifan I think use a "for statement" here can make the code more clearly, current code may confuse some newbie. Please change to use for statement to implement the function.

fuyifan:
The original 'while statement' is replaced by 'for statement'

fuyifan:
@pandamicro  It's my first time to use github. Sorry for commit so many times and I know it could make some confusion.
@xiaofeng11 "while statement" is revised.

xiaofeng11:
@fuyifan why do you simplify the code as follow

```
for (auto i = _usingSpriteFrames.begin(); i != _usingSpriteFrames.end(); i++;)
{
    removeSpriteFrameFromFile(i->first);
}
```

it's use few statement and won't need temporary variable. 

And commits can be squash into one as pandamicro mentioned. You may reference the document in github wiki : "https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit" or "http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html"

fuyifan:
@xiaofeng11 Function removeSpriteFrameFromFile will erase the iterator for that file and make the iterator in my code invalid. So in my opinion, 'while statement' is more readable.

xiaofeng11:
@fuyifan sorry I forget releaseSpriteFrames will remove the iterator, I think we need a descending enum here to avoid overflow. Positive while will have issue either. How do you think about the follow code

auto i = _usingSpriteFrames.end();
if(i != _usingSpriteFrames.begin())
{
    for(; i != _usingSpriteFrames.begin(); i--)
    {
        removeSpriteFrameFromFile(i->first);
    }
    removeSpriteFrameFromFile(i->first);
}

fuyifan:
@xiaofeng11 First, I don't think its readable. Second, there is critical bug in this code. when i = xx.end(), i->first is actually invalid when you first call removeSpriteFrameFromFile.  So how about i revert the state to the while statement and add some comment?

xiaofeng11:
@fuyifan OK, please give a descending enum solution, because erase won't in this function, we can not use "i = _usingSpriteFrames.erase(i)" to update iterator, a positive enum may cause error.

fuyifan:
@xiaofeng @pandamicro At present, what i can do is only adding some comment and reverting the code to 'while statement'. Any suggestion?

xiaofeng11:
@fuyifan OK
@pandamicro This pr can be merged.

pandamicro:
I have squashed your commits and resend another PR, so closing this one

PR #14174, opened 2015-10-16
CPPCheck fixes, correct deallocation, array size.

noshbar:
General Fixes:
- changed some delete operations to be deletions of arrays where applicable
- changed some "free" operations to "delete" where memory was allocated with "new"

CCPUScriptCompiler.cpp:
- added checks to see if temp1 and temp2 are non-null before dereferencing them

FileServer.cpp:
- changed the size of startFlag to be 13, seeing as the strcpy into it copies a string, "RuntimeSend:",
  is already 12 characters long, but strcpy() attempts to copy the NULL terminator too.

### DIFF
--- a/cocos/platform/winrt/CCApplication.cpp
+++ b/cocos/platform/winrt/CCApplication.cpp
@@ -122,7 +122,7 @@ const char * Application::getCurrentLanguageCode()
 
         if (pwszLanguagesBuffer)
         {
-            delete pwszLanguagesBuffer;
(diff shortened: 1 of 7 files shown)


pandamicro:
@xiaofeng11 Please review

xiaofeng11:
Reviewed, it can be merge

pandamicro:
Thanks @noshbar @xiaofeng11

PR #14177, opened 2015-10-17
Fix EventListenerCustom registered by UILabel never has been released bug

xiaofeng11:
Revert "Remove screen direction check in simulator" commit. Can not confirm if it will cause framework display status bug.

### DIFF
--- a/cocos/base/CCEventDispatcher.cpp
+++ b/cocos/base/CCEventDispatcher.cpp
@@ -619,7 +619,7 @@ void EventDispatcher::removeEventListener(EventListener* listener)
                 
                 if (_inDispatch == 0)
                 {
-                    listeners->erase(iter);
+                    iter = listeners->erase(iter);
                     CC_SAFE_RELEASE(l);
                 }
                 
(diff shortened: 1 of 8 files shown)


pandamicro:
Hi，@xiaofeng11 I think this solution to #14160 is too rough. Here is my thoughts, the problem was causing by removing nodes in the event dispatching process, so as [the original issue](http://www.cocoachina.com/bbs/read.php?tid=202831#1036521) described, it will prevent the custom event listener from being removed. Here is the code to process the listener which should be removed:

```
if (_inDispatch == 0)
{
    listeners->erase(iter);
    CC_SAFE_RELEASE(l);
}
```

There is no `else` logic, which means, if user remove node in the event dispatching process. The node have other listeners than the one being dispatching. Then those listeners will be leaked.

I think we should put the listeners into a `listenersToRemove` array and process them after the event dispatching process.

Besides, The revert for simulator modifications have been merged here: https://github.com/cocos2d/cocos2d-x/pull/14158

cocos/base/CCEventDispatcher.cpp
```diff
@@ -619,7 +619,7 @@ void EventDispatcher::removeEventListener(EventListener* listener)
                 
                 if (_inDispatch == 0)
                 {
-                    listeners->erase(iter);
+                    iter = listeners->erase(iter);
```
pandamicro, line 622: why do you need the returned value to be stored in iter here ?
Normally, the loop will be broken, then the iter will have no use any more

xiaofeng11, line 622: It's just for meet the C++11 standards, many 2dx user won't that good at C++ skills, I think give a sample in 2dx engine may help them decrease bug in coding.

xiaofeng11:
@pandamicro I've think about use a remove list, but we won't have a class map of 2dx, so I can not sure if there will be same problem in other process. "Touch Event" is a low rate event, and most lister won't response current event but needs to release it occur by user input in this event, I think enum listeners in this event won't reduce performance, and it will remove all unused listeners ensure.
If you confirm there's won't other operate cause sam problem, I'll change to use remove list to correct this problem.

pandamicro:
`removeEventListener` is the fundamental function to erase a listener, so I think it's safe to add the event listener to the to remove list (while dispatching) in this function. There shouldn't be other way to remove the event listener, otherwise we should rethink the design.

Besides, in EventDispatcher there is a `_toAddedListeners`, so it's reasonable to have a `_toRemovedListeners`, it's more consistent. So please change the logic by using remove list. Thank you

cocos/base/CCEventDispatcher.cpp
pandamicro, line 622: Ok

PR #14184, opened 2015-10-19
Fix EventListenerCustom registered by UILabel never has been released bug

xiaofeng11:


pandamicro:
[ci rebuild]

cocos/base/CCEventDispatcher.cpp
```diff
@@ -1525,4 +1543,61 @@ void EventDispatcher::setDirty(const EventListener::ListenerID& listenerID, Dirt
     }
 }
 
+void EventDispatcher::cleanToRemovedListeners(Event* event)
```
pandamicro: `event` haven't been used in this function, we should remove the parameter

pandamicro:
Fixed #14160

PR #14185, opened 2015-10-19
Fix some memory leaks

mogemimi:
This fixes some potential memory leaks when using create() functions.


pandamicro:
It's indeed more reliable, thanks

PR #14197, opened 2015-10-20
Fix CCDownloader-curl will crash in js-test & Fix broken image link in TextureCacheTest.js

xiaofeng11:


tests/js-tests/src/TextureCacheTest/TextureCacheTest.js
```diff
@@ -79,9 +79,9 @@ var TextureLoadImgTest = TextureCacheTestBase.extend({
         });
         this.addChild(this._labelSecond, 1);
 
-        var url = "http://cn.cocos2d-x.org/image/logo.png";
+        var url = "http://www.cocos2d-x.org/images/logo.png";
         cc.textureCache.addImageAsync(url, this.texFirstLoaded, this);
-        cc.textureCache.addImageAsync(url, this.texSecondLoaded, this);
```
pandamicro, line 84: Why remove the second one ?

xiaofeng11, line 84: Sorry, I comment the code for debug and forget to roll back

cocos/network/CCDownloader-curl.cpp
```diff
@@ -845,7 +847,7 @@ namespace cocos2d { namespace network {
             DLLOG("    DownloaderCURL: finish Task: Id(%d)", coTask.serialId);
         }
         
-        if (_impl->stoped())
+        if (!_breakScheduleFunc && _impl->stoped())
```
pandamicro: When the current object became wild pointer, we can't assume its property value remains the same, so there may still be problems

xiaofeng11: Current object destruct is raised by a function inside this object, so I think before the function exist, the object won't be really released.

pandamicro: But its destructor have been called, are you sure ?

xiaofeng11: I use this way to process class with sub-thread inside it on pc before. It works.
On mobile platform need tester do more test on variable device to confirm .

pandamicro:
Great !

PR #14207, opened 2015-10-22
fix shader on pc

super626:


super626:
Should be together with #14146
@pandamicro

pandamicro:
Complete for #14145

Issue #14128, opened 2015-10-13
[bug][jsb_boot.js] getRes crash if url is null


pixmaster:
cocos\scripting\js-bindings\script\jsb_boot.js
"cc.path.extname(url)" can return null and variable "type" is not checking it and immediately calling "type.toLowerCase()".
And in function "extname", can checking pathStr on null.

pandamicro:
Thanks for report this, I will submit a PR soon

PR #14209, opened 2015-10-22
Disable bitcode for all targets

pandamicro:
And #14128 Fix cc.loader.getRes crash caused by null type


pandamicro:
Cocos2d-x's third party library don't support bitcode, so it can't be enabled

PR #14211, opened 2015-10-22
Optimize scale9sprite

zilongshanren:
1. Add Sliced sprite support for Sprite class and rewrite the original scale9sprite. The new Scale9Sprite inherit from Sprite with type default to `Sliced`.
   The memory consumption is lower than previous implementation. And the computation is also more efficient.
4. A few minor change to use the new slice sprite in UI widgets. No behavior changed!  If you notice one, please report.

Note: **SpriteBatchNode**  can't support the new sliced sprite at the moment. In the past, it also the same. The original Scale9Sprite is a node, so it can't be added into the `SpriteBtachNode`.
But now the Sliced sprite is a Sprite which means it should be OK when adding it to the `SpriteBatchNode`.  

I suggest we mention this restrict in the release note and fix it in the future release. Thoughts?

Thanks.

### DIFF
--- a/cocos/ui/UISlider.cpp
+++ b/cocos/ui/UISlider.cpp
@@ -699,15 +699,13 @@ void Slider::copySpecialProperties(Widget *widget)
     {
         _prevIgnoreSize = slider->_prevIgnoreSize;
         setScale9Enabled(slider->_scale9Enabled);
-        auto barSprite = slider->_barRenderer->getSprite();
-        if (nullptr != barSprite)
+        if (!slider->_barRenderer->isUsingDefaultTexture())
         {
-            loadBarTexture(barSprite->getSpriteFrame());
+            loadBarTexture(slider->_barRenderer->getSpriteFrame());
         }
-        auto progressSprite = slider->_progressBarRenderer->getSprite();
-        if (nullptr != progressSprite)
+        if (!slider->_progressBarRenderer->isUsingDefaultTexture())
         {
-            loadProgressBarTexture(progressSprite->getSpriteFrame());
+            loadProgressBarTexture(slider->_progressBarRenderer->getSpriteFrame());
         }
         loadSlidBallTextureNormal(slider->_slidBallNormalRenderer->getSpriteFrame());
         loadSlidBallTexturePressed(slider->_slidBallPressedRenderer->getSpriteFrame());
--- a/cocos/ui/UISlider.cpp
+++ b/cocos/ui/UISlider.cpp
@@ -176,8 +176,8 @@ void Slider::setupBarTexture()
     this->updateChildrenDisplayedRGBA();
     _barRendererAdaptDirty = true;
     _progressBarRendererDirty = true;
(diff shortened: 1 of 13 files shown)


zilongshanren:
# API changes:

## Add two types to Sprite:

```
    enum class Type
    {
        Simple,
        Sliced
    };

    enum class State
    {
        NORMAL,
        GRAY
    };

```

## New added API for Sprite:

These APIs are all the original APIs for ui::Scale9Sprite except for `isUsingDefaultTexture`.
I will talk about it in the implementation details section.

```
    void setType(Type type);

    Type getType()const;

    bool isUsingDefaultTexture()const;

    void setCapInsets(const Rect& rect);

    Rect getCapInsets()const;

    void setState(State state);

    State getState()const;

    Size getOriginalSize() const;

    void setInsetLeft(float leftInset);


    float getInsetLeft()const;

    void setInsetTop(float topInset);

    float getInsetTop()const;

    void setInsetRight(float rightInset);

    float getInsetRight()const;

    void setInsetBottom(float bottomInset);

    float getInsetBottom()const;

```

# Implementation details:

## The  _isPatch9 member variable

The main logic to calculate the vertices and indices are in `caculateSlicedVertices`.
The  `_isPatch9` is used for Android `.9.png` format.

## isUsingDefaultTexture

It is used to query whether the sprite is using a default texture.

## Slice Sprite info calculate

All the vertices and indices info are computed in the `caculateSlicedVertices` method. It could be split into two.  One for computing the uv coordinate and another for computing the vertices. But for keep things simple, I just keep them into one.

## About the extra memory consumption for none Slice sprite

Since I added some extra member variables into the Sprite class, the memory consumption will be a little more than the previous Sprite. I could add a pointer here and dynamically alloc/dealloc the memory for Sliced sprite. But it may make code a little bit complex. Thoughts?

zilongshanren:
# Another approach to handle Scale9Sprite

We implement the Slice feature totally in Scale9Sprite and the Scale9Sprite is inherit from Sprite class for  reusing purpose.

The Scale9Sprite is mimic to Unity Image class which is sliced enabled on default.
The original Sprite class is mimic to Unity 2D:Sprite.

In this way, the original Sprite won't be touched and all the optimization work for the ui::Scale9Sprite.

Even though the Scale9Sprite is a subclass of Sprite and it could be added children of SpriteBatchNode.

But we could claim that we don't support adding Scale9Sprite to SpriteBatchNode.   It should make sense.

Thoughts?

@jwu  @nantas  @pandamicro  @2youyouo2  @jareguo

cocos/ui/UISlider.cpp
```diff
@@ -533,7 +533,7 @@ void Slider::adaptRenderers()
 
 Size Slider::getVirtualRendererSize() const
 {
-    return _barRenderer->getContentSize();
+    return _barRenderer->getOriginalSize();
```
zilongshanren, line 536: Maybe I should change all the `getVirtualRendererSize` to return the original size instead of the content size. Because of the content size may equals to the preferred size.

PR #14216, opened 2015-10-23
add Scale9Sprite optimization

zilongshanren:
1. Completely rewrite the slice-9 feature of ui::Scale9Sprite. The new Scale9Sprite inherits from Sprite with type default to `Sliced`.
4. A few minor change to use the new slice sprite in UI widgets. No behavior changed!  If you notice one, please report.

Note: **SpriteBatchNode**  can't support the new sliced sprite at the moment. In the past, it also the same. The original Scale9Sprite is a node, so it can't be added into the `SpriteBtachNode`.
But now the Sliced sprite is a Sprite which means it should be OK when adding it to the `SpriteBatchNode`.  

I suggest we mention this restrict in the release note and fix it in the future release. Thoughts?

Thanks.

### DIFF
--- a/cocos/2d/CCAutoPolygon.cpp
+++ b/cocos/2d/CCAutoPolygon.cpp
@@ -87,6 +87,12 @@ void PolygonInfo::setQuad(V3F_C4B_T2F_Quad *quad)
     triangles.verts = (V3F_C4B_T2F*)quad;
 }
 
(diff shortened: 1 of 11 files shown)


zilongshanren:
[ci rebuild]

zilongshanren:
[ci rebuild]

PR #14219, opened 2015-10-23
Add pageSize param to PageView

adamkotkowski:
Previously size of the pages inside pageView were set to fill all pageView, which didn't allow to have ex. more than one page visible on the screen.

Now you can do:

```
pageView->setContentSize(Size(1200,500));
pageView->setPageSize(600);
```

to have 2 pages visible, or `pageView->setPageSize(300)` to have `4`. It works for both HORIZONTAL and VERTICAL.

**If you don't specify pageSize it works as before**.

### DIFF
--- a/cocos/ui/UIPageView.cpp
+++ b/cocos/ui/UIPageView.cpp
@@ -41,6 +41,7 @@ _touchMoveDirection(TouchDirection::LEFT),
 _leftBoundaryChild(nullptr),
 _rightBoundaryChild(nullptr),
 _leftBoundary(0.0f),
+_pageSize(0.0f),
 _rightBoundary(0.0f),
 _customScrollThreshold(0.0),
 _usingCustomScrollThreshold(false),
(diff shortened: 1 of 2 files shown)


pandamicro:
This will change fundamentally the behavior of PageView, we haven't design such feature for PageView, so the page turn behavior, current page parameter, and other APIs may become really confusing while using `pageSize`.

I don't think it's a good idea to add this feature to PageView, it's more like another widget inherit from PageView. What do you think ?

adamkotkowski:
Right, when using page size, the current page is always the first page visible. Other parameters like custom paging threshold should work the same way.

In Android's `ViewPager` there is a `PagerAdapter` which provides pages and its size (default size is 100% as well). Also in Android there is much more functionality in ViewPager like transition animations, loading pages asynchronously, etc.

I think page size can be very useful especially when you support horizontal and portrait mode. Then, very often in horizontal there is 2x more width so there is a room for two pages at once. Also you can create something like a carousel/slider (http://owlgraphic.com/owlcarousel/demos/itemsCustom.html)  using pageSize param.

I needed pageSize in my game anyway so feel free to use it or not :)

pandamicro:
I understand, and thanks for your contribution.

So would you like to make this feature an extended class of PageView ?

adamkotkowski:
Sure, will do it when have a while.

pandamicro:
Thanks ! I'm closing this one, feel free when you have time to resend another PR

PR #14224, opened 2015-10-25
Research : solving memory model of JSB

pandamicro:
A proposed solution to use script object life cycle control native object reference counting.

To finally remove retain / release usage in JSB


pandamicro:
Based on https://github.com/cocos2d/bindings-generator/pull/198

pandamicro:
[ci rebuild]

pandamicro:
Commit https://github.com/pandamicro/cocos2d-x/commit/348509165c935b2c5fef61853b7cf8cd82f2bf4f is a new way to control the native object life cycle using FinalizeHook objects

based on : https://github.com/cocos2d/bindings-generator/pull/200

zilongshanren:
[ci rebuild]

pandamicro:
Two way reflecting solution have been implemented:

Branch: https://github.com/pandamicro/cocos2d-x/tree/memory_TwoWayReflect

Related commits:

https://github.com/pandamicro/cocos2d-x/commit/9a1465a5e24a46a8412ca0ab7bcfd7f6b2e43e00

https://github.com/pandamicro/bindings-generator/commit/89eb832311dc6b2c1e1ee1e40a290b9fd04c1e51

Originally proposed in http://discuss.cocos2d-x.org/t/a-possible-solution-for-jsb-memory-model-for-solving-invalid-native-object/17359

pandamicro:
## About the current PR

The current PR doesn't contain two way reflecting solution. It's a simpler solution which only reference native object by script object, so that native object won't be released as long as script object is alive.

## Problem beneath

But it's not perfect, some connection between objects are created in native scope, like `node.addChild(childNode);`, the script scope don't know `node` have a reference to `childNode`, so it's possible that `childNode` is collected by GC while its native object is still alive. It will cause further problems while user retrieve the object with native APIs and try to use it, because it's no longer the same one.

More precisely, I have encountered a case that the script object is an instance of a Class `TestLayer` extended from `LayerGradient`. It have been released by GC while its native object is still alive because it's in the scene graph. Then I retrieve the object in a event callback, instead of retrieving `TestLayer` object, I only got `LayerGradient` object, so all APIs of `TestLayer` is invalid now.

So there are references between object created in script scope, and also references created in native scope, neither the GC nor the reference counting system can be fully aware of the reference state of an object.

## Solutions
1. The two way reflecting
   
   The two way reflecting solution means: 
   - One way: Script object hold a reference to native object by `retain` / `release`
   - Another way: Native object hold a reference to script object too, by `AddNamedRootObject` and `RemoveNamedRootObject`
   
   This is a possible solution to solve the previous problem, well, theoretically. In real test, it cause unknown GC crash all the time, and the object releasing is always delayed, so not friendly to memory management.
2. Full GC relied
   
   This solution haven't been implemented yet. The essential idea is to pass the knowledge of connections in native to script scope. Ex:
   
   ```
   node.addChild(childNode);
   ```
   
   While the previous code is executing, the JSB code will also add a script reference from node to childNode like the following code:
   
   ```
   node.__nativeReference.push(childNode);
   ```
   
   In this way, `childNode` will keep alive in script scope. While node is collected by GC, then `__nativeReference` will be released too, `childNode` will also be garbage collected.
   
   But this means a very heavy work in JSB, we need to pass all connections to script scope, so all APIs which will create connections should be refactored in JSB level.

pandamicro:
@ricardoquesada Here is the PR for JSB memory model

ricardoquesada:
Thanks @pandamicro 

I couldn't understand the proposed solutions, both the "two way reflecting" and the "Full GC relied".

But I think what we should do is to hook/modify the way the VM handles reference counting.
ATM I can't remember how Spidermonkey handles them... 

Is it by `RootedObject` ? If so, then my suggestion is that we modify that class and test if the object that is going get rooted is a native object. If so, then we should call `retain`...and when the `RootedObject` gets destroy, then we should call `release`.

Does it make sense ? Thanks.

pandamicro:
> Is it by RootedObject ? If so, then my suggestion is that we modify that class and test if the object that is going get rooted is a native object. If so, then we should call retain...and when the RootedObject gets destroy, then we should call release.

RootedObject is the new Generational GC friendly object, briefly, as long as the RootedObject is alive, it will prevent the GC from collecting the object it represent. You can find more information about GC rooting in [this document](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/GC_Rooting_Guide)

In fact, the current solution can retain any native object that is bound to script, then release them while the script object get garbage collected, it's achieved via the finalize hook function. This have been already tested and proved.

But it's not the only problem, it's possible that a script object is destroyed while its native object is still alive. Then user can still get a new script object while accessing it via native APIs, but it's no longer the same object, take a look at the code snippet:

``` javascript
var tag = 999;
var TestLayer = cc.Layer.extend({
    layerId: 1
});

function addLayer(parent) {
    // `layer` is a local object and could be destroyed any moment after the function call
    var layer = new TestLayer();
    parent.addChild(layer, 0, tag);
    // This will print "[object object] : 1"
    cc.log(layer + " : " + layer.layerId);
}

addLayer(node);
// Force GC to destroy `layer`
cc.sys.garbageCollect();

// Retrieve the layer that have been added previously
// As the script object have been destroyed, JSB will create a new object to return
var theLayer = node.getChildByTag(tag);

// This will print "[object Layer] : undefined"
// Here `theLayer` is not any more the same object as `layer`, it's created by native code, but the native code only knows it's bound to a `Layer` object, it don't know the original class `TestLayer`
cc.log(theLayer + " : " + theLayer.layerId);
```

This is a serious problem because user don't expect their object to be destroyed, they think it's still hold by our system because it have been added to the scene graph. That's the reason we need to reference the script object to prevent them from being destroyed.

pandamicro:
Now to the two way reflecting solution:

The two way reflecting solution have a huge disadvantage, the memory will not be released in time. Image we have the following scene:

```
native scope: scene > layer > node
                |       |       |
script scope: scriptS scriptL scriptN
```

Each of them hold a script object, the script object will also retain the native object. 

During scene transition, the `scene`'s reference count drop to 1 (means only script object need it), then it will remove `scriptS` from root. At this time `scene` is still alive because the reference count haven't drop to 0, so layer, node still have 2 reference count (one from the parent, one from the script object).

Then during GC, `scriptS` get destroyed because it's no longer in root, its finalize hook will release `scene`. At this moment, `layer`'s reference count reduce to 1, it will remove `scriptL` from root. So it can only be destroyed during next GC, and the native `layer` can be released. 

This process will go through all the scene graph, means we need three GC to fully release the memory used by `scene`.

This is just catastrophic for memory management.

pandamicro:
The Full GC relied solution is simpler in theory: we made GC fully aware of all reference for an object.

If the object is a pure script object, nothing to do.

If the object is bound to a native object, we will pass the native reference to script scope. If the native API is being executed (in JSB) and create a new reference in native, for example, `addChild`, the native API will also push the child script into parent script object's `__nativeReference` array. In this way, the script scope will know parent holds a reference to child, then GC won't collect child as long as parent is alive.

I haven't implemented and tested this solution yet

ricardoquesada:
@pandamicro So, if I understood correctly you wan to do something like this:

``` c++
// CCRef.cpp file
Ref::retain()
{
   ...
   if (this->js_obj)
     JS_INC_REF(this->_jsObj);
}

Ref::release()
{
   if (this->js_obj)
     JS_DEC_REF(this->_jsObj);
}
```

Is that correct ? I think it should work.

The case that we should care about is this one then:

``` JavaScript
some_pure_js_object.sprite = cc.Sprite.create();
cc.sys.garbageCollect();
```

And the newly created sprite should not be deleted since `some_pure_js_object` is holding it.
Perhaps we need to support also what I said about hooking the `Root` thing and call `retain()`, `release()` from there... but if we do so, we should avoid the infinite recursion: eg: `retain()` calling `RootObject()` and `RootObject()` calling `retain().

pandamicro:
Thanks riq, I didn't have a very good solution to pass the native reference to script, what you suggested might work and should be easy to maintain.

> the newly created sprite should not be deleted since some_pure_js_object is holding it.

The case you mentioned should be ok, script GC can manage it correctly.

ricardoquesada:
ok. good to know. let me know if you have further questions!

pandamicro:
https://github.com/cocos2d/cocos2d-x/commit/b3f955c68ab07788c244f4b5a97a57c70449b3c3 An implementation for passing native references to script scope have been done.

But it crashes in GC, need more time to debug and figure out how to stabilize this version

ricardoquesada:
@pandamicro how can I help you in this?

pandamicro:
Hi, Riq, that would be great if you can checkout my `memory` branch and take a look at the crash, I have no clue for now. (You will need to regenerate the auto bindings)

The way to reproduce the crash:
1. run js-tests
2. entering one test case
3. return to the test case list view
4. click on the close button at top right corner (I modified the button callback to call garbage collection)

Then the crash should happen while releasing things.

Here is a brief information about what have been done the last days:

I have modified most places where create native references, which can not be known by GC, and pass the reference to script scope. The implementation to achieve it is in `ScriptingCore::retainScriptObject`, `ScriptingCore::releaseScriptObject`, which actually invoke `jsb.registerNativeRef` and `jsb.unregisterNativeRef` in `jsb_prepare.js`

In the current implementation, I have taken care things like `Node::addChild`, not all native refs haven been covered, but it's enough for now. What we need is to verify the viability of this method, those which haven't been covered won't trigger serious problems.

pandamicro:
I have drawn a graph to demonstrate different memory models



The original graph: https://docs.google.com/drawings/d/14byh2-vcealLBkDH0QB_LwxqhFzmpu_BEPuhogZjGW4/edit

The current implementation is shown in Model A, the Model B is the two way reflecting model, the Model C is @ricardoquesada 's Model (https://github.com/cocos2d/cocos2d-x/pull/14224#issuecomment-152263818)

In these graphs, the dotted line means native object reference count hold by the script object. The red line means how the native reference is constructed in script scope.

### About Model C:

As the implementation to pass native reference is in the Ref::retain, we may not be able to track the owner of the native reference. For example, `Node::addChild` actually retain the child by `_children` vector, so in `Ref::retain`, `this` point to the vector instead of the parent. So we can only create a global script object to host the child script object to make sure it won't be garbage collected.

The problem of Model C is similar to Model B, while `root` is garbage collected, the children won't be unless it receive notification from its native object. So finally, native object need to tell `globalScriptObj` to remove a reference, when its reference count reduce to 1(only script object still hold its reference). Actually the native object life cycle still control the script life cycle and the GC delay problem on Model B also exist in Model C.

### The Model A:

The Model A 's essential difference with other two is the script object life cycle is completely controlled by GC. While its root get garbage collected, all the children will also be collected instantly, then they will all release their reference to the native object. So the GC won't have delay.

ricardoquesada:
@pandamicro 
Regarding **Model C**, I didn't understand your comment.
What I proposed originally is to have a pointer of the JS object  in `Ref`.
eg:

``` c++
class Ref
{
    void retain();
    void release();

    JSObj* _jsObj;
    int _referenceCount;
}

Ref::retain()
{
     ++_referenceCount;
     JS_ADD_REF(_jsObj); 
}


Ref::release()
{
     JS_DEL_REF(_jsObj);
     --_referenceCount;
     if (_referenceCount==0)
        ...
}
```

with a code like that, I don't think there will be any problem with children and other stuff.
Thoughts?

ricardoquesada:
BTW, I tried your `memory` branch, regenerated the jsbindings and it crashes at the very beginning, when I tried to run the `cpp-tests`... before even selecting any tests.
I was able to reproduce the crash when running `js-tests`

pandamicro:
@ricardoquesada Thanks for testing

> with a code like that, I don't think there will be any problem with children and other stuff.

The thing is there is no reference counting system in JS, so the Model C is based on your idea, but not exactly the same. What I understand in your idea is that we can track the native reference in Ref::retain and Ref::release. In this way it's abstract and we don't need to know who is retaining who, all we need is to pass such relationship to script scope.

As there is no reference counting, I simulate the relationship with a global script object to hold the native references, so that whatever being retained won't get garbage collected.

About cpp-test, I haven't tried to compile it yet, it's possible my modifications are breaking things, but it won't be hard to fix them, you can just ignore it.

ricardoquesada:
@pandamicro yeah... I haven't read the internals yet (actually I read them long time ago but I forgot)...

how about this then?

``` c++
// cocos2d-x code
class Ref
{
     void retain();
     void release();

     JSObj* _jsObj;
}

void Ref::retain()
{
   ++_refCount;

   if (! _alreadyRooted)
   {
      JS::RootedObject(_jsObj);  
      _alreadyRooted = true;
   } 
}

void Ref::release()
{
   --_refCount;
   if (_refCount == 0)
   {
       if (_alreadyRooted)
          JS::RemoveObjectRoot(_jsObj);
      ....
   }
}
```

``` C++
// SpiderMonkey code
RemoveObjectRoot()
{
     // add something like this
     if (_wrappingCococs2dObject)
         // calls _release() instead of release() to avoid infinite recursion
         getCocos2dXObject()->_release();
}

AddObjectRoot()
{
     // add something like this
     if (_wrappingCococs2dObject)
         // calls _retain() instead of retain() to avoid infinite recursion
         getCocos2dXObject()->_retain();
}
```

pandamicro:
```
void Ref::release()
{
   --_refCount;
   if (_refCount == 0)
   {
       if (_alreadyRooted)
          JS::RemoveObjectRoot(_jsObj);
      ....
   }
}
```

The problem here is the _refCount will never be 0 before the script object get released, because it was retained by the script object. What we can do is to remove script object from root while the _refCount drop to 1, which means there is only the script object hold it. Then it's safe to garbage collect the script object.

In this way, it's the same with Two Way Reflecting model, it will have the same problem : https://github.com/cocos2d/cocos2d-x/pull/14224#issuecomment-152055063

The memory can't get released in time because the release of each object rely on the garbage collection of script object, and garbage collection is not instant, it's only called when spidermonkey want.

pandamicro:
Observed two interesting behavior:
1. During finalize function call, you can access the object being finalized with C++ SpiderMonkey APIs but you can't access the object in JavaScript. More specifically, in javascript, the code can be executed during finalize and have the correct results. But it will cause crash in GC (very similar crash with the one I mentioned https://github.com/cocos2d/cocos2d-x/pull/14224#issuecomment-155636777)
2. Trying to make object from not being garbage collected with `__nativeRefs` failed, in my test `cc.director.__nativeRefs[0].__nativeRefs[0]` is getting garbage collected.

pandamicro:
Great news, the crash have been fixed ! It was a mistake in my code, I created `__hook` one more time in the extended Class constructor...

See commit: https://github.com/pandamicro/cocos2d-x/commit/343655a643f1d35bd798360f36b4893e0162db03

But now the memory usage keep increasing, looks like old scene and its contents haven't been released, I will check out why.

pandamicro:
In https://github.com/cocos2d/cocos2d-x/commit/503c139357cb028891380dffa41ec495d4380358, the following changes have been done:
1. Remove code for passing native ref to script scope in ActionManager, EventDispatcher, to avoid affected by them and easier to stabilize.
2. Node::cleanup will now cleanup event listeners, and remove child nodes (which will also delete the native ref in script scope), otherwise objects won't get released, maybe a bug, need to investigate.
3. Fix a bug in MenuItem cleanup.
4. Remove native-script object hash while native object get deleted in Ref::~Ref
5. Added more log messages to help debugging

PR #14228, opened 2015-10-26
Fix ActionManagerEx some times crash when reload action frames.

xiaofeng11:
@pandamicro This pr fix the ActionManagerEx crash when switch between scenes, please review it.


pandamicro:
Review ok, iOS/Mac jenkins slave is temporary down, so ignore them for this pr

PR #14232, opened 2015-10-27
Fix FadeOutDownTilesTest won't show last tile

xiaofeng11:


cocos/2d/CCActionTiledGrid.cpp
```diff
@@ -572,12 +572,12 @@ FadeOutDownTiles* FadeOutDownTiles::clone() const
...
-    if (pos.height == 0)
-    {
-        return 1.0f;
-    }
+    //if (pos.height == 0)
+    //{
+    //    return 1.0f;
+    //}
 
-    return powf(n.y / pos.height, 6);
```
pandamicro, line 580: Could pos.height be negative here ?

xiaofeng11, line 580: time value will be always <= 1.0, so pos.height won't be negative

pandamicro:
Review ok

cocos/2d/CCActionTiledGrid.cpp
```diff
@@ -572,12 +572,12 @@ FadeOutDownTiles* FadeOutDownTiles::clone() const
...
 {
     Vec2 n = Vec2((float)_gridSize.width, (float)_gridSize.height) * (1.0f - time);
-    if (pos.height == 0)
-    {
-        return 1.0f;
-    }
+    //if (pos.height == 0)
+    //{
+    //    return 1.0f;
+    //}
```
pandamicro: Please remove these lines

xiaofeng11: removed

PR #14240, opened 2015-10-27
Change the android test app to be landscape reversible.

neokim:


pandamicro:
But the cpp test haven't been designed for portrait orientation, so what's the meaning of this ?

neokim:
Yes, but currently it supports only one-side landscape (home button on right). There is one more orientation for landscape (home button on left). This is for supporting both of them.

pandamicro:
Oh, I see, thanks

PR #14252, opened 2015-10-28
Change PageView to derive from ListView

neokim:
`PageView` was derived from `Layout` and it implemented the features of scrolling and item arrangement from scratch. But the features are already there in `ListView`. So remove those duplicated implementations from `PageView` and make it subclass from `ListView`.

By this, `PageView` becomes more simplified and maintainable because it considers only paging implementation. And the user experiences regarding scrolling are processed in one place - `ScrollView`.

One more, page view now has a page indicator which is represented as a list of small circles.

### DIFF
--- a/cocos/ui/UIPageView.cpp
+++ b/cocos/ui/UIPageView.cpp
@@ -34,7 +34,7 @@ IMPLEMENT_CLASS_GUI_INFO(PageView)
 PageView::PageView():
 _indicatorEnabled(false),
 _indicator(nullptr),
-_indicatorPositionAsAnchorPoint(Vec2::ANCHOR_MIDDLE),
+_indicatorPositionAsAnchorPoint(Vec2(0.5f, 0.1f)),
 _currentPageIndex(-1),
 _customScrollThreshold(0.0),
 _usingCustomScrollThreshold(false),
@@ -65,11 +65,11 @@ PageView* PageView::create()
     
 bool PageView::init()
 {
+    _indicator = PageViewIndicator::create();
+    addProtectedChild(_indicator, 10000);
+
     if (ListView::init())
     {
-        _indicator = PageViewIndicator::create();
-        addProtectedChild(_indicator, 10000);
-        
         setDirection(Direction::HORIZONTAL);
         setMagneticType(MagneticType::CENTER);
 		setScrollBarEnabled(false);
@@ -200,9 +200,8 @@ void PageView::handleReleaseLogic(Touch *touch)
 
     Vec2 touchMoveVelocity = flattenVectorByDirection(calculateTouchMoveVelocity());
 
-    static const float THRESHOLD = 500;
-    CCLOG("handleReleaseLogic() touchMoveVelocity.length()=%.2f", touchMoveVelocity.length());
-    if(touchMoveVelocity.length() < THRESHOLD)
+    static const float DEFAULT_THRESHOLD = 500;
+    if(touchMoveVelocity.length() < DEFAULT_THRESHOLD)
     {
         startMagneticScroll();
     }
--- a/cocos/ui/UIPageView.cpp
+++ b/cocos/ui/UIPageView.cpp
@@ -96,17 +96,17 @@ void PageView::addWidgetToPage(Widget *widget, ssize_t pageIdx, bool forceCreate
     insertCustomItem(widget, pageIdx);
 }
 
-void PageView::addPage(Layout* page)
+void PageView::addPage(Widget* page)
 {
     pushBackCustomItem(page);
 }
 
-void PageView::insertPage(Layout* page, int idx)
+void PageView::insertPage(Widget* page, int idx)
 {
     insertCustomItem(page, idx);
 }
 
-void PageView::removePage(Layout* page)
+void PageView::removePage(Widget* page)
 {
     removeItem(getIndex(page));
 }
--- a/cocos/ui/UIPageView.cpp
+++ b/cocos/ui/UIPageView.cpp
@@ -137,6 +137,13 @@ bool PageView::isUsingCustomScrollThreshold()const
     return _usingCustomScrollThreshold;
 }
 
+void PageView::onItemListChanged()
+{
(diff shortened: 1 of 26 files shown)


zilongshanren:
:+1: :+1: :+1: :+1: :+1: :+1:  You guys always shock me with amazing PR.

pandamicro:
Really great !
Thanks for your efforts @neokim 

We may not be able to merge it in v3.9, but it will definitely on the roadmap for the next version

zilongshanren:
[ci rebuild]

neokim:
Jenkins complains about a line I have added in `ccUtils.h` at 111. It seems that `CC_DLL` is causing a syntax error. I have no idea why it is. @zilongshanren Can you help me out about this?

cocos/ui/UIPageView.h
```diff
@@ -111,81 +107,108 @@ class CC_GUI_DLL PageView : public Layout
...
-    void scrollToPage(ssize_t idx);
+    void scrollToItem(ssize_t itemIndex);
 
+    /**
+     * Gets current displayed page index.
+     * @return current page index.
+     *
+     * Since v3.9, this is deprecated. Use `getCurrentPageIndex()` instead.
+     */
+    CC_DEPRECATED_ATTRIBUTE ssize_t getCurPageIndex() const;
```
zilongshanren, line 183: If we mark these method as deprecated, then they won't be exported in JS and Lua binding. So we have to manually write some wrappers in Deprecated.js or Deprecated.lua file.

zilongshanren:
@neokim 
It's wired...
The first method also contains the CC_DLL macro.

How about simply remove the CC_DLL?

zilongshanren:
@neokim 
The `setCustomScrollThreshold` API in PageView now have no effect due to your implementation.

This API controls when the PageView will start to auto scroll, on default, only when user swiper more than half of the PageView's width.

If we want to turn the page after swipe a short distance, then we should call this API.

I guess you might miss this API.

neokim:
I understand what `setCustomScrollThreshold()` was doing. But it is not necessary any more because the feature is done by `ListView`'s magnetic scroll, which is utilizing auto-scroll of `ScrollView`. You will see that the page scrolling by magnetic is much more smoother than before.

The paging scroll threshold will be adjusted by changing magnetic strength which is not done yet.

zilongshanren:
@neokim 
So you will include the magnetic strength customization in this PR, right?

neokim:
I think it would be right to be another PR because it is an additional feature for `ListView`.

zilongshanren:
@neokim  But it also means this PR is incomplete. Anyway, if you could open another PR after this one is merged. I think it's ok for me.

cocos/ui/UIPageView.cpp
```diff
@@ -70,683 +62,192 @@ PageView* PageView::create()
...
     {
-        Node * page = _pages.at(pageIdx);
-        page->addChild(widget);
+        _indicator->setDirection(direction);
+        refreshIndicatorPosition();
     }
 }
 
-Layout* PageView::createPage()
+void PageView::addWidgetToPage(Widget *widget, ssize_t pageIdx, bool forceCreate)
```
zilongshanren, line 110: Here is a TAB, please change it to 4 spaces. And it happens many places in the header file and the implementation file.

neokim:
@zilongshanren Ok. I thought the paging scroll threshold can be removed from this PR. I might need to mark it as deprecated APIs.

zilongshanren:
@neokim 
qq20151029-0 2x
In the UIPageViewVerticalTest:7 , the PageView

 Indicator's initial position is wrong.

zilongshanren:
@neokim 
Regards the original PageView API, it seems you have deprecated almost all of them.

Due to the implementation, I think it make sense to do this change.

But the original PageView API's names indent better than the ListView's API.

Like `pushBackCsutomItem` doesn't make sense compare to `addPage` when using PageView.

So I suggest we just keep the old API.

Thoughts?

neokim:
@zilongshanren I'll see the issue about wrong initial position of indicators.

Regarding original `PageView` APIs, the previous APIs accept and return `Layout` as its page item. But I think that the type of page item doesn't need to be limited to `Layout`, it can be any `Widget`. Overloading the APIs to accept `Widget` can be an option. But the methods returning `Layout` like `getPages()`, `getPage()` cannot be overloaded to return `Widget` because overloading is not permitted when only return type is different.

I agree with you about the naming, but I just wanted to be consistent.

I will overload some methods accepting `Widget` like `addPage()`, but `getPages()` still needs to be changed.

zilongshanren:
@neokim  
OK, add more overload methods is a possible solution.  I agree with you that the `getPages` API should be deprecated.
Thanks.

liamcindy:
@zilongshanren Taking into account the features of the current editor, these changes cloud be merged.

zilongshanren:
@neokim 
Hi,  do you have  time now? Since we have released v3.9,  I'm ready to merge this PR now.
But as we discussed here, you owe me another PR  :smile:

neokim:
@zilongshanren I haven't had much time to look into the issues in this PR. I will wrap it up soon like in early next week.

Regarding to adding magnetic scrolling strength (the PR you saying I owe :smiley:), I think it should be considered as an independent one. So after I resolve the issues you mentioned in this PR and you merge it, there will be no remaining things in here.

zilongshanren:
[ci rebuild]

neokim:
@zilongshanren The issues you mentioned are all resolved. Please review PR again.

neokim:
And removed the `CC_DLL` in front of `createSpriteFromBase64()` in `ccUtils.h` to avoid the compilation errors on windows platform.

zilongshanren:
[ci rebuild]

zilongshanren:
@neokim  
:+1:  Good work. 
I suggest to add some adapt APIs for `addPage`, `removePage` and `setCustomScrollThreshold`.

Look forward to your next PR. :smile:

neokim:
@zilongshanren Thanks.

PR #14257, opened 2015-10-28
Improve scale9sprite v3

zilongshanren:
1. Completely rewrite the slice-9 feature of ui::Scale9Sprite. 

Thanks.


ricardoquesada:
could you explain how does it work? Is the API the same?
Does it use one texture instead of 9 ? does it support auto-batching ?
Does it work like SpriteKit's `centerRect` ? If not, can we implement something similar to `centerRect` ?
https://developer.apple.com/library/mac/documentation/SpriteKit/Reference/SKSpriteNode_Ref/index.html#//apple_ref/occ/instp/SKSpriteNode/centerRect

And also, does it pass all the tests ?  and will it require more tests ?

zilongshanren:
@ricardoquesada 
Sure, I only used one texture and a capInsets(something like centerRect in SpriteKit, keep this just for back compatibility) to calculate the UV. The main logic is laid down in `calculateUV` function.

```
 // (0,0)  O = capInsets.origin
    // v0----------------------
    // |        |      |      |
    // |        |      |      |
    // v1-------O------+------|
    // |        |      |      |
    // |        |      |      |
    // v2-------+------+------|
    // |        |      |      |
    // |        |      |      |
    // v3-------------------- (1,1)  (texture coordinate is flipped)
    // u0       u1     u2     u3
```

 Then I used a preferredSize and capInsets to calculate the vertices which is finally composed to the 16 vertices. (Code in `calculateVertices` function)

```
 //
    // y3----------------------(preferedSize.width, preferedSize.height)
    // |        |      |      |
    // |        |      |      |
    // y2-------O------+------|
    // |        |      |      |
    // |        |      |      |
    // y1-------+------+------|
    // |        |      |      |
    // |        |      |      |
    //x0,y0--------------------
    //         x1     x2     x3
```

A Scale9Sprite is simply 18 triangles with some vertices sharing. So we need 16 vertices and 54 indices to compute the 18 triangles.

Since they all share the same material id, so it supports batch drawing.

All the APIs are remaining unchanged, so there will be no back compatibility issues here.

And it does pass all the tests and I will do more test including memory leaks tests and performance test.

Thanks for your reply.

zilongshanren:
@ricardoquesada  
Squashed new PR here https://github.com/cocos2d/cocos2d-x/pull/14263

ricardoquesada:
great!. This is a very good feature since it allow us to batch sprites.

Could you create a test that shows that sprite9 + regular sprites  can be batched ? (of course the should share the the same texture id + shader)

zilongshanren:
@ricardoquesada  Thanks, I will add the sprite9 + regular sprites batch tests.

PR #14260, opened 2015-10-28
Added Application::getVersion() to get the app version.

pyrosphere:
Added a method to the Application class to get the version of the app on iOS/Mac/Android/Win8, on the remaining platforms that don't have app versions it should return an empty string.

The feature has been requested in this discussion thread: http://discuss.cocos2d-x.org/t/check-app-version/6294/5

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
@@ -212,6 +212,15 @@ public class Cocos2dxHelper {
         sVibrateService.vibrate((long)(duration * 1000));
     }
 
+ 	public static String getVersion() {
+ 		try {
+ 			String version = Cocos2dxActivity.getContext().getPackageManager().getPackageInfo(Cocos2dxActivity.getContext().getPackageName(), 0).versionName;
+ 			return version;
+ 		} catch(Exception e) {
+ 			return "-";
+ 		}
+ 	}
+
     public static boolean openURL(String url) { 
         boolean ret = false;
         try {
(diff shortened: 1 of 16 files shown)


TheCodez:
@pyrosphere Wouldn't it be better to just return "1.0" if no version is available?

pyrosphere:
Personally I would prefer an empty string as it would let you check for `.empty()` and not display that information rather than an incorrect number. If you'd want to have "1.0" you could always do something like:

```
std::string version = cocos2d::Application::getVersion();
if (version.empty()) { version = "1.0"; }
```

TheCodez:
Well, this also makes a lot of sense. So let the cocos developer decide

pandamicro:
@minggo you was in the original discussion, can you review this PR?

minggo:
It looks good to me.
But it is better to add `override` for all platforms.

@pandamicro 
I think we should add a test case for it.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
```diff
@@ -212,6 +212,15 @@ public static void vibrate(float duration) {
         sVibrateService.vibrate((long)(duration * 1000));
     }
 
+ 	public static String getVersion() {
+ 		try {
+ 			String version = Cocos2dxActivity.getContext().getPackageManager().getPackageInfo(Cocos2dxActivity.getContext().getPackageName(), 0).versionName;
+ 			return version;
+ 		} catch(Exception e) {
+ 			return "-";
```
TheCodez: You should change this to "" , to prevent inconsistencies

pyrosphere: Fixed, nice catch. :+1:

pandamicro:
@pyrosphere Please fix jenkins compilation issues, try to compile your code in Xcode 7

pyrosphere:
It seems Jenkins fails on iOS/Mac because the override warnings of unrelated methods are being treated as errors:

```
/jenkins/workspace/ccs-pr/node/mac/build/../cocos/platform/mac/CCApplication-mac.h:55:18: error: 'setAnimationInterval' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
```

PR #14263, opened 2015-10-29
improve scale9sprite

zilongshanren:
1. Completely rewrite the slice-9 feature of ui::Scale9Sprite. 
4. Add RenderingType to slice sprite. When Scale9Enable is true, then we have two rendering type here. 
   
   In Simple mode, the 4 borders are all 0 and the whole sprite will scale horizontally and vertically. In this mode only 1 quad is used to rendering.
   
   In Slice mode, it will use 18 triangles to rendering the slice 9 sprite. If the 4 borders are 0,  there still be 18 triangles computed. So choose your RenderingType wisely.

### DIFF
--- a/cocos/2d/CCAutoPolygon.cpp
+++ b/cocos/2d/CCAutoPolygon.cpp
@@ -87,6 +87,17 @@ void PolygonInfo::setQuad(V3F_C4B_T2F_Quad *quad)
     triangles.verts = (V3F_C4B_T2F*)quad;
 }
 
+void PolygonInfo::setTriangles(TrianglesCommand::Triangles other)
+{
+    this->releaseVertsAndIndices();
+    isVertsOwner = false;
+    
+    this->triangles.vertCount = other.vertCount;
+    this->triangles.indexCount = other.indexCount;
+    this->triangles.verts = other.verts;
+    this->triangles.indices = other.indices;
+}
+
 void PolygonInfo::releaseVertsAndIndices()
 {
     if(isVertsOwner)
--- a/cocos/ui/UIScale9Sprite.cpp
+++ b/cocos/ui/UIScale9Sprite.cpp
@@ -477,6 +477,14 @@ namespace ui {
 
         return true;
     }
+    
+    void Scale9Sprite::configureSimpleModeRendering()
+    {
+        this->setInsetTop(0);
+        this->setInsetBottom(0);
+        this->setInsetLeft(0);
+        this->setInsetRight(0);
+    }
 
     void Scale9Sprite::createSlicedSprites()
     {
@@ -488,6 +496,11 @@ namespace ui {
             if (tex == nullptr)
             {
                 return;
+            }
+            
+            if (_renderingType == RenderingType::SIMPLE)
+            {
+                this->configureSimpleModeRendering();
             }
 
             auto capInsets = CC_RECT_POINTS_TO_PIXELS(_capInsetsInternal);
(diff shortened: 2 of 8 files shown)


zilongshanren:
[ci rebuild]

zilongshanren:
@ricardoquesada 
The tested are added at the end of Scale9SpriteTest in Node:UI Test.

zilongshanren:
@dabingnn 
Please help to review this PR, thanks.

cocos/2d/CCAutoPolygon.cpp
```diff
@@ -87,6 +87,17 @@ void PolygonInfo::setQuad(V3F_C4B_T2F_Quad *quad)
...
 }
 
+void PolygonInfo::setTriangles(TrianglesCommand::Triangles other)
+{
+    this->releaseVertsAndIndices();
+    isVertsOwner = false;
+    
+    this->triangles.vertCount = other.vertCount;
+    this->triangles.indexCount = other.indexCount;
+    this->triangles.verts = other.verts;
```
dabingnn, line 97: This is a pointer reference, please its safety.

cocos/ui/UIScale9Sprite.cpp
```diff
@@ -515,18 +426,9 @@ namespace ui {
...
-
-        if (!spriteFrame)
-        {
-            return false;
-        }
-
         Rect rect(textureRect);
         Size size(originalSize);
         
-        _capInsets = capInsets;
```
dabingnn, line 528: Why do you delete this line?

cocos/2d/CCAutoPolygon.cpp
zilongshanren, line 97: What does it mean?

cocos/ui/UIScale9Sprite.cpp
zilongshanren, line 528: It shouldn't be deleted, I will check it later.

cocos/2d/CCAutoPolygon.cpp
dabingnn, line 97: `verts` is a pointer, It is not safe to do it like this if `other`'s `verts` is deleted.

cocos/ui/UIScale9Sprite.cpp
zilongshanren, line 528: @dabingnn  I found it, because the _capInsets are not needed any more. Only the _capInsetsInternal is used ATM.

cocos/ui/UIScale9Sprite.h
```diff
@@ -679,59 +654,50 @@ namespace ui {
...
         /** Preferred sprite's size. By default the preferred size is the original size. */
         
         //if the preferredSize component is given as -1, it is ignored
         Size _preferredSize;
-        /**
-         * The end-cap insets.
-         * On a non-resizeable sprite, this property is set to CGRect::ZERO; the sprite
-         * does not use end caps and the entire sprite is subject to stretching.
-         */
-        Rect _capInsets;
```
zilongshanren, line 734: @dabingnn  Here I have already deleted _capInsets.

cocos/ui/UIScale9Sprite.cpp
dabingnn, line 528: Got it.

zilongshanren:
@pandamicro  
My PR is ready for merging now.

PR #14264, opened 2015-10-29
Fix: Mesh::setGLProgramState may modify state block

super626:


super626:
@pandamicro

pandamicro:
Review ok

PR #14265, opened 2015-10-29
Fixed AssetsManagerEx bugs

cpascal:
I've tested this code with cpp-tests's AssetsManagerExTest.cpp code.

BUGS 1. After download error occurs, AssetsManagerEx never calls extension::EventAssetsManagerEx::EventCode::UPDATE_FAILED nor extension::EventAssetsManagerEx::EventCode::UPDATE_FINISHED because AssetsManagerEx::_totalWaitToDownload can't be 0.
( Download-retrying code is only at the UPDATE_FAILED event trigger code in the example.. )

BUGS 2. AssetsManagerEx::batchDownload() creates CCDownloader-curl task, but it opens temporary files at the time, not in the task thread. :( so.. if I pass assets lists above 512 files, that will makes open file limit error. (Windows's fopen() default file-open-limit is 512)

### DIFF
--- a/cocos/network/CCDownloader-curl.cpp
+++ b/cocos/network/CCDownloader-curl.cpp
@@ -132,15 +132,6 @@ namespace cocos2d { namespace network {
                     }
                 }
 
-                // open file
-                _fp = fopen(util->getSuitableFOpen(_tempFileName).c_str(), "ab");
-                if (nullptr == _fp)
-                {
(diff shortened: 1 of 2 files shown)


cpascal:
BUGS 3. If last download-asset file meets error, AssetsManagerEx never calls EventAssetsManagerEx::EventCode::UPDATE_FAILED.

PR #13716, opened 2015-09-08
Refactory particle system update.

fusijie:
To improve #11566.

The comparison of particle update.

Env:  before = v3.8, after = v3.8 + this pr. iOS(50 ParticleFireworks). Android (30 ParticleFireworks).
eb75da33-520e-47aa-bb62-d6031afb86bb

PR #14277, opened 2015-10-30
Simplified ParticleSystem

TheCodez:
There were a lot of unnecessary for loops which I merged together. This should also improve performance a little bit.


stevetranby:
If I had to guess as why there were so many loops it would be the possibility of data locality for each component makes the multiple for loops more efficient. ??

TheCodez:
In my understanding using only a single for loop is faster. Not sure?
@fusijie any reason for using so many for loops instead of just one?

fusijie:
@TheCodez, to reduce cache miss.And also it's unwisely to use if/else inside loop.
you can refer to https://github.com/cocos2d/cocos2d-x/pull/13716 and https://github.com/cocos2d/cocos2d-x/issues/11566

PR #14280, opened 2015-10-31
CREATE_FUNC macro NULL -> nullptr

jwvg0425:


zilongshanren:
@jwvg0425 
Good catch :+1:

PR #14282, opened 2015-11-01
Add size config in manifest for AssetsManagerEx

ookcode:
We know AssetsManagerEx cannot get the total size before update and cannot get the percent(not percent by file) timely.
In most cases,we should get the total size and let users to commit download especially in 3G network.
I add a member “size” in Manifest file,and cal the total size after get diff assets map.If you config the size in manifest file,you can get the total size and it will not continue download unless you call batchDownload().
I think I have considered the compatibility
Thank you for reviewing my code! Wish you have a nice day!


ookcode:
Sorry for my mistake,I will recreate when I master these git operations.

PR #14303, opened 2015-11-03
Fix ComponentTest in la-test can not load lua file under iOS & Android

xiaofeng11:


minggo:
It looks good to me.

PR #14310, opened 2015-11-03
Added the ability to set physics debug color of Shapes & Joints

TheCodez:
This pr adds the ability to change the physics debug color of the Shapes fill and outline. 
You can also change this on the PhysicsBody itself to change the color of all its shapes. 
Additionally, you can also set the line and point color of Joints

### DIFF
--- a/cocos/physics/CCPhysicsBody.cpp
+++ b/cocos/physics/CCPhysicsBody.cpp
@@ -855,6 +855,46 @@ int PhysicsBody::getGroup() const
     }
 }
 
+void PhysicsBody::setDebugOutlineColor(const Color4F& outlineColor)
+{ 
+    for (auto& shape : _shapes)
+    {
+        shape->setDebugOutlineColor(outlineColor);
+    }
+}
+
+Color4F PhysicsBody::getDebugOutlineColor() const
+{
+    if (!_shapes.empty())
+    {
+        return _shapes.front()->getDebugOutlineColor();
+    }
+    else
+    {
+        return Color4F(1.0f, 0.0f, 0.0f, 1.0f);
+    }
+}
+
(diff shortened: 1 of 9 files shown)


TheCodez:
@pandamicro @WenhaiLin any chance to get this into 3.10?

WenhaiLin:
@pandamicro This PR seems good to me.

pandamicro:
@TheCodez, v3.10 can't add any more features, so after merge, it could be published with v3.11

pandamicro:
@xiaofeng11 Please confirm whether we can merge this improvement to v3.11.

xiaofeng11:
@TheCodez Please add a test case for this feature, thanks!
@pandamicro Yes, it can be merge to v3.11 after add a test case for testers can verify the function.

TheCodez:
It has already some examples in the Physicstest, but i can add an extra test.

xiaofeng11:
Hi TheCodez,
We have a testing group verify all new features or functions, when add new features or functions we should provide a testing document to testing group to explain how to confirm it's works fine. So please add an extra test, that will be much helpful for explain how to verify this new feature. Thank you very much.

TheCodez:
@xiaofeng11 @pandamicro OK, added test case.

minggo:
@TheCodez Could you please fix conflict?

TheCodez:
I don't think it's possible anymore to change color on individual shapes as this is now handled by chipmunk.
@newnon you changed to the new chipmunk debug system. Is the above still possible?

newnon:
I'm not sure chipmunk just call primitive call method with Paramus there is some method with which you can set color of this primitives, but it needs some additional research...

minggo:
@TheCodez  i think i have to close the PR, and please send a new one to fix it if needed. thanks.

PR #14321, opened 2015-11-05
Fix SocketIO crash on reconnect

### LINKED ISSUE #14287: Socket.IO crash due to over-release
If a Socket.IO connection is dropped a crash occurs when attempting to reconnect. This has been observed on iOS but likely impacting other platforms.

Steps to reproduce:
- connect to a Socket.IO server (websocket)
- break the connection
- reconnect to the same url

Then the ref count assertion in cocos2d::Ref fires:
`CCASSERT(_referenceCount > 0, "reference count should be greater than 0");`

Here's what I think is happening in SocketIO.cpp on v3 branch:

(Disconnect)
Line 989:  `SIOClientImpl::onClose(WebSocket* ws)`  is called when the websocket connection is dropped.
Line 997: `void SIOClient::socketClosed()` is called
Line 1089: _SIOClient instance released_ (ref count ->1 because still in `SIOClientImpl::_clients` map)

(Reconnect)
Line 1211: SIOClient instance returned (from `SIOClientImpl::_clients` map)
Line 1223: `SIOClient::disconnect` called
Line 1077: `SIOClientImpl::disconnectFromEndpoint` called
Line 659: SIOClient erased from client map (ref count -> 0)
Line 1079: SIOClient instance released (ref count == 0 !!!)

The lifecycle management in this implementation is very hard to follow though.

pandemosth:
Cleans up SIOClientImpl instance when websocket is unexpectedly closed. Fixes #14287

### DIFF
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -986,6 +986,10 @@ void SIOClientImpl::onClose(WebSocket* ws)
         {
             iter->second->socketClosed();
         }
+        // discard this client
+        _connected = false;
+        Director::getInstance()->getScheduler()->unscheduleAllForTarget(this);
+        SocketIO::getInstance()->removeSocket(_uri);
     }
 
     this->release();


ricardoquesada:
looks safe. Although we need a test case that reproduces the crash.

pandamicro:
@pandemosth Can you add a test case to reproduce the problem so that we won't break it again in future versions

pandemosth:
To test this the websocket connection needs to be dropped by the server or a network error. Any ideas on how to do this in an automated test?

pandamicro:
Can we test the current SocketIO test case and cut the network off manually to verify ? I think we can ask our QA team to do it

minggo:
@pandemosth Thanks.

The corresponding issue is added a milestone, so not add milestone for this PR.

PR #13206, opened 2015-08-05
Improved EditBox implement for platform win32

summerinsects:
1. Associated button caption with 'ReturnType'
2. Re-implement the maximum input length
3. Associated win32 edit control style with some InputFlag and InputMode, such as InputFlag::PASSWORD, InputMode::NUMERIC
4. No longer block the rendering when the win32 dialog has not yet been closed
5. Invoke editBoxTextChanged when text in win32 edit control has been changed
6. Keep the displaying text of Label in EditBox synced between win32 edit control

PR #14327, opened 2015-11-05
Fix a fatal bug in EditBox implement for platform win32

summerinsects:
@zilongshanren @WenhaiLin  
Very sorry to say that, I found my PR (https://github.com/cocos2d/cocos2d-x/pull/13206) has a fatal bug.
In the earlier version, the Win32InputBox is a modal dialog box, and it would block rendering, then I use a timer to do rendering in that PR. However, there is a fatal error, i.e., main loop of Director was re-entered.
For example, when the dialog box is not yet closed, and try to remove the EditBox in a schedule, it will crash.

This PR uses modaless dialog box, which avoids re-entry, and won't crash any more.
I've done a series of tests, such as removing an EditBox in a schedule, in a Callback action, etc.

Please do some more test before merging.
Sorry trouble you.


ricardoquesada:
Thanks. Could you add a test case as well?

summerinsects:
In cpp-tests?

zilongshanren:
@wohaaitinciu 
Yes, cpp-tests is OK.

summerinsects:
@zilongshanren test case is added

zilongshanren:
@wohaaitinciu :+1:

PR #14328, opened 2015-11-06
Fix "TextureCache Test" in js-test crash under Android

xiaofeng11:
@pandamicro Please review this pr

### DIFF
--- a/cocos/network/CCDownloader-android.cpp
+++ b/cocos/network/CCDownloader-android.cpp
@@ -65,9 +65,6 @@ namespace cocos2d { namespace network {
         : _id(++sDownloaderCounter)
         , _impl(nullptr)
         {
-            // use local static variable make sure native methods registered once
-            if(!_registered)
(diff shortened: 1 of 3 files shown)


cocos/network/CCDownloader-android.cpp
```diff
@@ -66,7 +66,8 @@ namespace cocos2d { namespace network {
         , _impl(nullptr)
         {
             // use local static variable make sure native methods registered once
-            static bool _registered = _registerNativeMethods(JniHelper::getEnv());
+            if(!_registered)
+                _registered = _registerNativeMethods(JniHelper::getEnv());
```
pandamicro: I don't think we still need this, what we need is to guarantee the invocation of _preloadJavaDownloaderClass in JNI init

xiaofeng11: removed

pandamicro:
Review ok, thanks @xiaofeng11

PR #14333, opened 2015-11-06
Fix cannot add view to mFrameLayout when extends Cocos2dxActivity

nhuanvd:
Cause class org.cocos2dx.lib.ResizeLayout is private. So activity class extends from Cocos2dxActivity will not know mFrameLayout variable, can't do anything with it. So make ResizeLayout public and move class into separate file. 

I meet the issue with Android Studio project.


zilongshanren:
@nhuanvd

This change is reasonable, thanks.

ricardoquesada:
yes. looks a like a safe patch

PR #14334, opened 2015-11-06
Fix nullptr when Android app enter background

zilongshanren:


zilongshanren:
@pandamicro  Don't Merged ATM. it needs more test.

zilongshanren:
@pandamicro 
This PR is for v2 branch.

ricardoquesada:
looks like a safe patch. a test case that reproduces the crash would be great.

PR #14338, opened 2015-11-06
adds port to host str while making websocket connection

TimothyZhang:
If the port is not specified, some server implementations might deny the connection.

### DIFF
--- a/cocos/network/WebSocket.cpp
+++ b/cocos/network/WebSocket.cpp
@@ -38,6 +38,9 @@
 #include <signal.h>
 #include <errno.h>
 
+#include <iostream>
+#include <sstream>
+
 #include "libwebsockets.h"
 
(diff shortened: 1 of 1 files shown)


ricardoquesada:
Thanks @TimothyZhang What is the best way to test it ?
patch looks good to me.

TimothyZhang:
I'm afraid there's no public websocket server for testing which listens on non-standard ports. 
Maybe we could write a simple server with python.

dumganhar:
@TimothyZhang , I tried to use `socket.io` (nodejs) to reproduce this issue but I couldn't.
Is it just an issue of service side implementation of `socket.io`?

dumganhar:
@TimothyZhang 
Since Websocket code has been updated a lot, please send another PR for this issue. Thanks.

PR #14340, opened 2015-11-07
easy jni

fnz:
This commit adds variadic template wrappers for JNI calls. Full description is available in the following forum post: http://discuss.cocos2d-x.org/t/easy-jni-with-variadic-templates/24740

Tests are added as well.


pandamicro:
I think it would be great to change the current jni usage in engine to adapt the new API

pandamicro:
@ricardoquesada Can you review this PR

cocos/platform/android/jni/JniHelper.cpp
```diff
@@ -275,4 +275,61 @@ namespace cocos2d {
...
+    template <>
+    std::string JniHelper::getJNISignature(std::string) {
+        return "Ljava/lang/String;";
+    }
+
+    jstring JniHelper::convert(cocos2d::JniMethodInfo& t, const char* x) {
+        return t.env->NewStringUTF(x ? x : "");
+    }
+
+    jstring JniHelper::convert(cocos2d::JniMethodInfo& t, std::string x) {
```
Dimon4eg: Better to pass string by const reference

ricardoquesada:
looks good to me.
is it possible to call system libraries with the JNIHelper. If so, I would add the in the tests... like creating a windows or creating a button or something like that.

fnz:
It's seems like there's some room for improvement with constexpr and garbage collection, I'll update the pr with these features and more tests in two days or so

cocos/platform/android/jni/JniHelper.cpp
fnz: The problem here is with template specialization, there should be an exact type match, otherwise general template will be instantiated. I'll fix this soon with non-template functions, thanks.

fnz:
Improvements:
1. Signatures are now constructed during the compile time thanks to constexpr. It means there is no overhead compared to writing the signatures manually.
2.  Deletion of local refs created for String arguments no longer relies on Java garbage collector, which can cause local reference table overflow when a Java method is called from C++ loop for instance. Corresponding test is added.
3. If the user tries to call Java method with unsupported argument, say CCSprite*, it will cause compile time error with a clear message.

I think the next step is to refactor existing JNI calls.

fnz:
I'm in progress

fnz:
I refactored existing C++ to Java calls with template methods, however all cases that didn't fit new approach are left intact. A clear run-time message in case of incorrect class name/method name/signature is added. Unfortunately I had to remove constexpr signature inference due to some cryptic memory issues.

cezheng:
I think this PR is awesome. Will this be merged in version 3.10? I am maintaining a private SDK which is dependent on the old JNIHelper, and I definitely think this can also be a great improvement on my code base @pandamicro @ricardoquesada

ricardoquesada:
yes, I think this can be merged for v3.10. @linshun ?

linshun:
Hi @fnz, It is a great improvement, thanks. 
Hi @ricardoquesada, it could be merged in version 3.10 if this code could be merged today. Or we should ship it in next version.

ricardoquesada:
@linshun Ok. can we merge it today? Who is in charge of merging this? thanks.

zilongshanren:
@fnz 
Let's merge this PR in the next release cycle. :smile:

fnz:
@zilongshanren
Sure, I'll merge 3.10 changes after current release

zilongshanren:
@fnz  :100:  :+1:  Awesome work! :+1:  :heart:

cocos/audio/android/jni/cddandroidAndroidJavaEngine.cpp
```diff
@@ -78,271 +69,53 @@ AndroidJavaEngine::~AndroidJavaEngine()
...
-
-    if (!getJNIStaticMethodInfo(methodInfo, "preloadBackgroundMusic", "(Ljava/lang/String;)V")) {
-        return;
-    }
-
-    jstring stringArg = methodInfo.env->NewStringUTF(fullPath.c_str());
-    methodInfo.env->CallStaticVoidMethod(methodInfo.classID, methodInfo.methodID, stringArg);
-    methodInfo.env->DeleteLocalRef(stringArg);
-    methodInfo.env->DeleteLocalRef(methodInfo.classID);
+    JniHelper::callStaticVoidMethod(helperClassName, "preloadBackgroundMusic", filePath);
```
minggo, line 78: Why use `filePath` instead of `fullPath`?

cocos/audio/android/jni/cddandroidAndroidJavaEngine.cpp
```diff
@@ -78,271 +69,53 @@ AndroidJavaEngine::~AndroidJavaEngine()
...
-
-    if (!getJNIStaticMethodInfo(methodInfo, "playBackgroundMusic", "(Ljava/lang/String;Z)V")) {
-        return;
-    }
-
-    jstring stringArg = methodInfo.env->NewStringUTF(fullPath.c_str());
-    methodInfo.env->CallStaticVoidMethod(methodInfo.classID, methodInfo.methodID, stringArg, loop);
-    methodInfo.env->DeleteLocalRef(stringArg);
-    methodInfo.env->DeleteLocalRef(methodInfo.classID);
+    JniHelper::callStaticVoidMethod(helperClassName, "playBackgroundMusic", filePath, loop);
```
minggo, line 83: Why use `filePath` instead of `fullPath`?

zilongshanren, line 83: @minggo  It's a bug...

minggo, line 83: Ok. It is fixed in #15512

fnz:
Oops. I've checked everything using tests before committing, background music was fine 😬

cocos/audio/android/jni/cddandroidAndroidJavaEngine.cpp
fnz, line 83: Oops. I've checked everything using tests before committing, background music was fine 😬

minggo, line 83: @fnz It is ok. It only has problem when search path is set then fullpath is different from filepath.

Issue #8634, opened 2014-10-10
Pixelated TransitionCrossFade when used with setDesignResolutionSize


pyrosphere:
Hi,

I've noticed that TransitionCrossFade makes the image pixelated when using setDesignResolutionSize with a resolution different than the window. You can see this effect in the following images:

![normal](https://cloud.githubusercontent.com/assets/779064/4596727/78ed7308-50a6-11e4-87ea-87084b93ca10.png)

![transition](https://cloud.githubusercontent.com/assets/779064/4596730/7e7c0d20-50a6-11e4-886a-bda589561b7b.png)

This seems to be caused by the RenderTexture using the design resolution (which is usually lower) instead of the actual window resolution.

As a workaround I made some minor changes in TransitionCrossFade::onEnter method, marked here with comments // CHANGE:

```
void TransitionCrossFade::onEnter() {
    TransitionScene::onEnter();

    // create a transparent color layer
    // in which we are going to add our rendertextures
    Color4B color(0,0,0,0);
    Size size = Director::getInstance()->getWinSize();
    Size screenSize = Director::getInstance()->getOpenGLView()->getFrameSize(); //CHANGE
    LayerColor* layer = LayerColor::create(color);

    // create the first render texture for inScene
    RenderTexture* inTexture = RenderTexture::create(round(screenSize.width/CC_CONTENT_SCALE_FACTOR()), round(screenSize.height/CC_CONTENT_SCALE_FACTOR()),Texture2D::PixelFormat::RGBA8888, GL_DEPTH24_STENCIL8); // CHANGE
    if (nullptr == inTexture) {
        return;
    }
    inTexture->getSprite()->setAnchorPoint(Vec2(0.5f,0.5f));
    inTexture->setPosition(size.width/2, size.height/2);
    inTexture->setAnchorPoint( Vec2(0.5f,0.5f) );
    inTexture->setKeepMatrix(true); // CHANGE

    // render inScene to its texturebuffer
    inTexture->beginWithClear(0, 0, 255, 255);
    _inScene->visit();
    inTexture->end();

    // create the second render texture for outScene
    RenderTexture* outTexture = RenderTexture::create(round(screenSize.width/CC_CONTENT_SCALE_FACTOR()), round(screenSize.height/CC_CONTENT_SCALE_FACTOR()),Texture2D::PixelFormat::RGBA8888, GL_DEPTH24_STENCIL8); // CHANGE
[...]

PR #14342, opened 2015-11-08
Fix resolution in TransitionCrossFade

pyrosphere:
TransitionCrossFade has some resolution issues that can, in some situations, result in a pixelated image during the transition.

I described the issue some time ago in https://github.com/cocos2d/cocos2d-x/issues/8634 and posted a possible workaround - the code in this PR is cleaner and should work on all platforms.

### DIFF
--- a/cocos/2d/CCTransition.cpp
+++ b/cocos/2d/CCTransition.cpp
@@ -1299,10 +1299,12 @@ void TransitionCrossFade::onEnter()
     // in which we are going to add our rendertextures
     Color4B  color(0,0,0,0);
     Size size = Director::getInstance()->getWinSize();
+    Size screenSize = Director::getInstance()->getOpenGLView()->getFrameSize();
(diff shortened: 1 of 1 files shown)


ricardoquesada:
Thanks. 
@pandamicro I haven't tested this patch, but the code looks good to me.

cristian1980:
Hi,

There is still a issue with this code. The scale of the textures seems to be slightly bigger then the final scene and you may notice a small jump in scale when this transition completes.

I added this code to the end of TransitionCrossFade::onEnter() which seems to fix the issue but I couldn't figure out what is the root cause of this scaling issue:

```
void TransitionCrossFade::onEnter()
{
................................
    Vec2 r1=layer->convertToWorldSpace(Vec2(0,0));
    r1 = Director::getInstance()->convertToUI(r1);
    Vec2 r2=layer->convertToWorldSpace(size);
    r2 = Director::getInstance()->convertToUI(r2);
    layer->setScaleX(size.width/(r2.x-r1.x));
    layer->setScaleY(size.height/(r1.y-r2.y));
}
```

In my case size is (640, 960), size.width/(r2.x-r1.x) and size.height/(r1.y-r2.y) both return 0.998357594 instead of 1.

Best regards

PR #14343, opened 2015-11-09
fix cull face for render queue

super626:
2D render queue should disable cull face.


pandamicro:
[ci rebuild] only iOS failed

pandamicro:
Logic ok, @super626 Please create an issue which describes this problem more in detail, and describe the current solution too. Thanks

super626:
fix #14345

PR #14354, opened 2015-11-10
[js-tests] Scale9Sprite fade actions with cascade opacity enabled

1scaR1:
Test case for https://github.com/cocos2d/cocos2d-html5/pull/3181


pandamicro:
Thanks !

PR #14372, opened 2015-11-12
Combine Studio change for 2d components

xiaofeng11:
@pandamicro @minggo Here's the cocostudio added / modified for 2d components that hope to be merge into 2dx engine, please review this PR.


cocos/2d/CCDrawNode.cpp
```diff
@@ -398,6 +405,14 @@ void DrawNode::onDrawGLLine(const Mat4 &transform, uint32_t flags)
         // texcood
         glVertexAttribPointer(GLProgram::VERTEX_ATTRIB_TEX_COORD, 2, GL_FLOAT, GL_FALSE, sizeof(V2F_C4B_T2F), (GLvoid *)offsetof(V2F_C4B_T2F, texCoords));
     }
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    //������߿����,��ο��߻���, ���ȹرն��ز���
```
minggo: Please don't use Chinese comments.

cocos/2d/CCCameraBackgroundBrush.h
```diff
@@ -236,8 +240,15 @@ class CC_DLL CameraBackgroundSkyBoxBrush : public CameraBackgroundBrush
      * Draw background
      */
     virtual void drawBackground(Camera* camera) override;
-    
-CC_CONSTRUCTOR_ACCESS:
+
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    bool isActived();
```
minggo: Please add `const` if possible.

cocos/2d/CCDrawNode.h
```diff
@@ -312,7 +312,15 @@ class CC_DLL DrawNode : public Node
...
     
     void setLineWidth(int lineWidth);
-    
+
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    void csSetLineSmooth(bool enable);
+
+    bool csIsLineSmooth();
+
+    float csGetLineWidth();
```
minggo: I think it is not a good idea to start with `cs` even it is only be used for cocos studio.

cocos/2d/CCNode.h
```diff
@@ -38,6 +38,10 @@
 #include "2d/CCComponentContainer.h"
 #include "2d/CCComponent.h"
 
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+#include "CocosStudioExtension.h"
+#endif
+
```
minggo: Why need it?

cocos/2d/CCDrawNode.h
xiaofeng11: cs is the cocos studio name rule, it means function will be map to C# layer

cocos/2d/CCNode.h
xiaofeng11: Cocos Studio define a data structure of resource description named "ResouceData" in this header file.

minggo:
I think most added codes are not special to cocos studio, they can be part to engine. So i think this PR is for a common review, it is better to divide it into more small PR, each PR only do a thing. For example, there can be a PR for camera modification.

And in the small PR, it is better to add comments for the reason for the PR, for example, why you need to modify camera, what issues it resolves.

cocos/2d/CCDrawNode.h
minggo: I think it is better to add comment for the purpose, not using function names for the target.

xiaofeng11: OK, these function name changed to 2dx standard mode.

xiaofeng11:
@minggo Most code added by studio is get resource file name & resource status, or force reload resource file(because the file may overwrited by updating). These function is useful for editors, but for most game, all resource is fixed, these function can not give much help.

xiaofeng11:
@TheCodez Yes, my mistake. Thanks a lot for notify me.

minggo:
@xiaofeng11 
Yep, but i think it is not problem to public these functions. Is there any limitation usage of these functions?

xiaofeng11:
@minggo
Most modify in this pr is OK to be release to public, but the modify of Sprite::initWithFile function entrance, we have remove ASSERT, because editor can not been stopped when pass a wrong param, it is not meet engine design rules, I think it should be set to editor only. And follow values add to CameraFlag in CCCamera.h have never use in Engine code 

+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
-    SkyBox = 1 << 10,
-    DEFAULTCAMERA = DEFAULT | SkyBox,
-    PIXEL = 1 << 13,
-    FRONT = 1 << 14,
  +#endif

Shall we need release them to public?

cocos/2d/CCNode.h
minggo: But there is not need to include it here because it is not used in CCNode.

cocos/2d/CCScene.h
```diff
@@ -109,7 +109,11 @@ class CC_DLL Scene : public Node
      * @param renderer The renderer use to render the scene.
      * @js NA
      */
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    virtual void render(Renderer* renderer);
+#else
     void render(Renderer* renderer);
+#endif
```
minggo: I think can just add `virtual`, not need to add a new function.

cocos/2d/CCNode.h
xiaofeng11: Many component inherit from node use "ResouceData" define, add it to here so we won't need add this header file to every component use it.

minggo: I don't think it is a good idea to do it like this because other files that don't need this header file will also include it too. It will slow down the compiling speed. We should include as less header file as possible.

minggo:
About `Sprite::initWithFile()`, i think it is better that it will not crash at any situation.

About camera flags, please ask @super626 to take a look. I think they are needed if cocos studio use them because cocos studio is a use case of cocos2d-x.

cocos/2d/CCNode.h
xiaofeng11: OK, I'll remove it from header file and add to every cpp need it

minggo: Thanks.

cocos/2d/CCSprite.cpp
```diff
@@ -164,7 +164,15 @@ bool Sprite::initWithTexture(Texture2D *texture, const Rect& rect)
 
 bool Sprite::initWithFile(const std::string& filename)
 {
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    if (filename.empty())
+        return false;
```
minggo, line 171: It is better to print some log saying that the filename is invalid.

xiaofeng11:
@minggo
I've notice there has many ASSERT in cocos2dx engine, they can help developers quickly check their params pass to engine function, because in a game, developer never need transfer a invalid param to engine function. Please confirm we will remove this ASSERT, for some developer used to ASSERT checking, this may increase their debugging work.

minggo:
I agree with you. It is valuable for debugging.

xiaofeng11:
@minggo all new functions remove cocos studio only limitation now, except ASSERT modification and default value that use different value in cocos studio.

minggo:
Thanks.
I think this common review PR can be closed. Then we can splits it into more small PR. Then we can view them in more detail. @ricardoquesada will join the review too, so i think it is important to make more PR for reviewing.

ricardoquesada:
- Some classes have new functions: for those new functions, we need test cases. No new functionality should be added without test cases. For example, there are new functions to reload stuff... so, test cases should be added that prove that reload works ok.
- +1 for having setter/getters: useful for the editor and users.
- -1 for having `#ifdef  COCOS_STUDIO`... how can we get rid of those `#ifdef`... will subclasses work?

super626:
@xiaofeng11 @minggo 
The added camera flags are studio only, it confuses the users.

But I do not know how to deal with this except the macros COCOS_STUDIO.

xiaofeng11:
@super626 @minggo  There's a macro define is ASSERT related, it was discus in forum. And other macros we have just discus and confirm we can remove them.

super626:
@ricardoquesada 
It is helpful to having macros for editor I think.
1. It is struggle to keep the balance between engine and editor. Engine wants smaller package and editor wants more features.
2. Many developers have their own editors. What should do when they need to access to protected  members? Subclass or modify? Most of them could modify the engine code directly. So have an editor version may be helpful. It is better that smaller change is needed when editor is made using cocos2d-x.

ricardoquesada:
@super626 

For 1). Engine should be smaller and simpler. If Studio needs more features, then it should subclass, or use components... I guess it depends case by case.

For 2). I guess it depends case by case. Having more public methods means more maintenance work for the engine team.

But I think we should avoid having the `#ifdef COCOS_STUDIO` macro.

xiaofeng11:
[ci rebuild]

xiaofeng11:
Add wrong change to this commit, I'll close this PR and move it to : https://github.com/cocos2d/cocos2d-x/pull/14502

PR #14373, opened 2015-11-12
modi AutoreleasePool class to avoid wild pointer

Y-way:
moidfied AutoreleasePool  class`s clear mothed to avoid wild pointer


minggo:
@Y-way Why it has wild pointer?

PR #14377, opened 2015-11-13
Correctly calculate milliseconds on OSX

elnormous:
You were mixing milliseconds with microseconds when calculating frame interval, so the animations were really glitchy on OSX. This pull request fixes that and uses microseconds everywhere correctly which results in a smooth animation.


minggo:
> You were mixing milliseconds with microseconds when calculating frame interval, so the animations were really glitchy on OSX.

@elvman It looks good. But i tested without this PR and can not find the issue You said.

minggo:
Could you tell me how to reproduce it?

PR #14379, opened 2015-11-13
Fix deprecated warnings

mogemimi:
This fixes the following warnings when compiling in Xcode 7:

```
CCArmatureDataManager.cpp:253:27: 'CCArmatureDataManager' is deprecated
CCArmatureDataManager.cpp:261:36: 'CCArmatureDataManager' is deprecated
ArmatureNodeReader.cpp:168:24: 'CCArmature' is deprecated
```


zilongshanren:
:+1:

PR #14385, opened 2015-11-14
ios9 cceditbox keyboard

jasonz1987:
fix ios9 cceditbox keyboard bug

### DIFF
--- a/cocos/platform/ios/CCEAGLView-ios.mm
+++ b/cocos/platform/ios/CCEAGLView-ios.mm
@@ -718,6 +718,15 @@ Copyright (C) 2008 Apple Inc. All Rights Reserved.
 
 - (void)onUIKeyboardNotification:(NSNotification *)notif;
 {
+    if ([[[UIDevice currentDevice] systemVersion] floatValue] >= 9.0)
+    {
+        if(originalRect_.size.height > originalRect_.size.width) {
+            CGFloat tempValue = originalRect_.size.height;
+            originalRect_.size.height = originalRect_.size.width;
+            originalRect_.size.width = tempValue;
+        }
+    }
+
     NSString * type = notif.name;
     
     NSDictionary* info = [notif userInfo];


zilongshanren:
Why we need this patch?  The EditBox keyboard orientation is wrong after iOS 9?

zilongshanren:
@xiaofeng11  please check this issue, thanks.

jasonz1987:
@zilongshanren  sorry,i've forgotten giving my solution: i use cocos2dx version: 2.2.x.and it's okay for ios8,but wrong keyboard orientation on ios9,plz check.

xiaofeng11:
@jasonz1987 Can you provide us a simple project to test this issue? I've check this problem with cpp-test, but it can not reappear.

jasonz1987:
@xiaofeng11 configure your xcode project support "portrait" device orientation,you'll found the problem.

xiaofeng11:
@jasonz1987 Cocos2d won't support create a app support both "landscape" and "portrait" orientation. If you want your app display as "portrait" orientation, please use "cocos new" command create a new project with "--portrait" param. Force change screen orientation of a exist project will cause many issue.

jasonz1987:
@xiaofeng11 though we configuer xcode support both landscape and portrait,but we write some code in rootvc forcely make landscape orientation, it's okay for ios6~ios8 ,but not work on ios9.

xiaofeng11:
@jasonz1987 I see. Maybe this is a iOS9 compatible issue, but the reason cause this issue is conflict cocos2dx design. @zilongshanren Please determine whether should us merge this PR in cocos2d-x or for special user project only.

zilongshanren:
@jasonz1987 let's just keep your solution here and wait for other developer's response.
@xiaofeng11  I think it's safer to not merge it at the moment.

jasonz1987:
@zilongshanren @xiaofeng11 okay,maybe it's only used for my solution.

linguofeng:
Thanks，fix my problem.

PR #14391, opened 2015-11-16
Combine Studio change for configuration

xiaofeng11:
These value in cocos studio use max but not 0 as in cocos-2dx engine
@pandamicro @minggo Please review this pr.


minggo:
These values will be calculated in `Configuration::getInfo()`.  0 is just initialization value.

xiaofeng11:
I've just confirm with XieXin, this change is for old WPF version cocos studio and won't need any more. I'll close this pr.

PR #14393, opened 2015-11-16
Combine Studio change for 3d components

xiaofeng11:
Some ASSERT limited to use in cocos-2dx but not cocos studio.
And Texture2D::setPath may cause resource management error, so it and function related have been limited use in cocos studio only

### DIFF
--- a/cocos/3d/CCMesh.h
+++ b/cocos/3d/CCMesh.h
@@ -194,7 +194,7 @@ public:
      */
     void setForce2DQueue(bool force2D) { _force2DQueue = force2D; }
 
-    GLuint getTextureName();
+    void checkTexture();
 
 CC_CONSTRUCTOR_ACCESS:
 
--- a/cocos/3d/CCMesh.h
+++ b/cocos/3d/CCMesh.h
@@ -194,6 +194,8 @@ public:
      */
     void setForce2DQueue(bool force2D) { _force2DQueue = force2D; }
 
+    GLuint checkTextureName();
+
 CC_CONSTRUCTOR_ACCESS:
 
     Mesh();
@@ -234,6 +236,8 @@ protected:
     std::vector<float> _spotLightUniformInnerAngleCosValues;
     std::vector<float> _spotLightUniformOuterAngleCosValues;
     std::vector<float> _spotLightUniformRangeInverseValues;
+
+    std::string _texFile;
 };
 
 // end of 3d group
--- a/cocos/3d/CCMesh.h
+++ b/cocos/3d/CCMesh.h
@@ -194,7 +194,7 @@ public:
      */
     void setForce2DQueue(bool force2D) { _force2DQueue = force2D; }
 
-    GLuint checkTextureName();
+    GLuint getTextureName();
 
 CC_CONSTRUCTOR_ACCESS:
 
--- a/cocos/3d/CCPlane.h
+++ b/cocos/3d/CCPlane.h
@@ -100,6 +100,11 @@ public:
     */
     PointSide getSide(const Vec3& point) const;
 
+    /**
+    * Return the side where the point is.
+    */
+    PointSide getSide(const Vec3& point, const Vec3& halfSize) const;
+
 protected:
     Vec3 _normal; // the normal line of the plane
     float _dist; // original displacement of the normal
--- a/cocos/3d/CCPlane.h
+++ b/cocos/3d/CCPlane.h
@@ -103,7 +103,7 @@ public:
     /**
     * Return the side where the point is.
     */
-    PointSide getSide(const Vec3& point, const Vec3& halfSize) const;
+    PointSide getSide(const Vec3& point, const Vec3& tolerance) const;
 
 protected:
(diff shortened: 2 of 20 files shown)


pandamicro:
[ci rebuild]

xiaofeng11:
Condition macro removed, and update implement logic.

cocos/3d/CCSprite3D.h
```diff
@@ -200,6 +200,9 @@ class CC_DLL Sprite3D : public Node, public BlendProtocol
     */
     void setForce2DQueue(bool force2D);
 
+    Vector<Mesh*>& getMeshes() { return _meshes; }
```
minggo: Vector<Mesh*>& getMeshes() -> Vector<Mesh*>& getMeshes() const

cocos/renderer/CCTextureCache.h
```diff
@@ -204,6 +204,12 @@ class CC_DLL TextureCache : public Ref
      */
     const std::string getTextureFilePath(Texture2D* texture)const;
 
+    void setDirty(bool dirty) { _dirty = dirty; }
+    bool isDirty() const { return _dirty; }
+
+    void renameTextureWithKey(std::string srcName, std::string dstName);
```
minggo: `std::string srcName` -> `const std::string& srcName`
`std::string dstName` -> `const std::string& dstName`

And please add document for new functions.

minggo:
I think it is better to add test case for new features if possible. I am not familiar with 3d, so i think @super626 should review the modification is suitable or not.

cocos/3d/CCSprite3D.h
xiaofeng11: Here the return value is Vector& and _meshes is not const, so we can not add const declare. We can only change to Vector getMeshes() const, but in this way, it will return a copy _meshes, use more memory, please confirm how can we do.

minggo: getMeshes() is expected to return a value to be modified?

xiaofeng11: No, we don't modify it. I'll change this function to Vector getMeshes() const

minggo: If it is not modified, why not change to `const Vector<Mesh*>& getMeshes() const { return _meshes; }`?

xiaofeng11:
@minggo functions modify as suggestion, and comment has been added

minggo:
Thanks.

ricardoquesada:
as @minggo said, please add a test case for these functions. thanks.

cocos/3d/CCSprite3D.h
```diff
@@ -200,6 +200,21 @@ class CC_DLL Sprite3D : public Node, public BlendProtocol
...
+
+    /*
+    * Get AABB
+    * If the sprite has animation, it can't be calculated accurately,
+    * because bone can drive the vertices, we just use the origin vertices
+    * to calculate the AABB.
+    *
+    * @param world Use world transform or not.
+    */
+    const AABB& getAABB(bool world) const;
```
super626: The old one is const AABB& getAABB() const. I do not think this modification is necessary.

cocos/renderer/CCTexture2D.h
```diff
@@ -406,6 +406,9 @@ class CC_DLL Texture2D : public Ref
     /** Get a shader program from the texture.*/
     GLProgram* getGLProgram() const;
 
+    void setValid(bool valid) { _valid = valid; }
```
super626, line 410: What happens if setValid(false)

cocos/3d/CCPlane.h
```diff
@@ -100,6 +100,11 @@ class CC_DLL Plane
     */
     PointSide getSide(const Vec3& point) const;
 
+    /**
+    * Return the side where the point is.
+    */
+    PointSide getSide(const Vec3& point, const Vec3& halfSize) const;
```
super626: halfSize is confusing. tolerance is better.
And update the comment. Distance between point and plane is greater than tolerance returns PointSide::FRONT_PLANE, distance is less than -tolerance returns return PointSide::BEHIND_PLANE, returns PointSide::IN_PLANE otherwise.

cocos/3d/CCMesh.h
```diff
@@ -194,6 +194,8 @@ class CC_DLL Mesh : public Ref
      */
     void setForce2DQueue(bool force2D) { _force2DQueue = force2D; }
 
+    GLuint checkTextureName();
```
super626: getTextureName may be better.

xiaofeng11:
@super626 Codes update as suggestion, please reivew again, Thanks

xiaofeng11:
@super626 Add some missing change for particle 3d, please review

xiaofeng11:
@ricardoquesada Add a new change to Director::drawScene for cocos studio need to draw 3d scene, please review, Thanks.

xiaofeng11:
@super626 @ricardoquesada liamcindy has add test to this PR please reivew

super626:
ok

PR #14394, opened 2015-11-16
Terrain: char* to string

super626:


super626:
@pandamicro review please

PR #14395, opened 2015-11-16
Combine Studio change for ActionTimeline

xiaofeng11:
Use exception in cocos studio but not ASSERT to avoid application crash

### DIFF
--- a/cocos/editor-support/cocostudio/ActionTimeline/CCFrame.cpp
+++ b/cocos/editor-support/cocostudio/ActionTimeline/CCFrame.cpp
@@ -618,20 +618,44 @@ void InnerActionFrame::onEnter(Frame *nextFrame, int currentFrameIndex)
 
 void InnerActionFrame::setStartFrameIndex(int frameIndex)
 {
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    if (_enterWithName)
+    {
+        CCLOG(" cannot set start when enter frame with name. setEnterWithName false firstly!");
+        throw std::exception();
+    }
+#else
     CCASSERT(!_enterWithName, " cannot setStartFrameIndex when enterWithName is set");
+#endif
     _startFrameIndex = frameIndex;
 }
 
 
 void InnerActionFrame::setEndFrameIndex(int frameIndex)
 {
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    if (_enterWithName)
+    {
+        CCLOG(" cannot set end when enter frame with name. setEnterWithName false firstly!");
+        throw std::exception();
+    }
+#else
     CCASSERT(!_enterWithName, " cannot setEndFrameIndex when enterWithName is set");
+#endif
     _endFrameIndex = frameIndex;
 }
 
 void InnerActionFrame::setAnimationName(const std::string& animationName)
 {
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    if (!_enterWithName)
+    {
+        CCLOG(" cannot set aniamtioname when enter frame with index. setEnterWithName true firstly!");
+        throw std::exception();
+    }
+#else
     CCASSERT(_enterWithName, " cannot set aniamtioname when enter frame with index. setEnterWithName true firstly!");
+#endif
     _animationName = animationName;
    
 }


xiaofeng11:
@pandamicro @minggo Please review this pr.

minggo:
What will happen in cocos studio when the exception happens? Can cocostudio continue to work?

xiaofeng11:
@minggo Cocos Studio can catch the exception give user a notice and continue running.

minggo:
Ok, thanks.
It looks good to me.

pandamicro:
@minggo If you review ok, please merge it directly, thanks ~

minggo:
Riq will join the review too.
So i think it is better to listen his opinion before merging.

cocos/editor-support/cocostudio/ActionTimeline/CCFrame.cpp
```diff
@@ -618,20 +618,44 @@ void InnerActionFrame::onEnter(Frame *nextFrame, int currentFrameIndex)
...
     _endFrameIndex = frameIndex;
 }
 
 void InnerActionFrame::setAnimationName(const std::string& animationName)
 {
+#ifdef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
+    if (!_enterWithName)
+    {
+        CCLOG(" cannot set aniamtioname when enter frame with index. setEnterWithName true firstly!");
+        throw std::exception();
```
linshun, line 654: Please follow the coding style: https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#exceptions

ricardoquesada:
why does it throw exceptions? Please read the cocos2d-x cpp coding guidelines:
https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#exceptions

[IGNORE THIS COMMENT... See my comment below]

ricardoquesada:
So, I guess the exceptions were added because that code is compiled only when used by cocos studio, which makes sense.
So, my suggestion is that instead of throwing `std::exception`, a different exception is thrown... like: `FatalException` or something like that.

Even better... perhaps cocos2d-x should have a "safe" mode where it doesn't crash at all.

xiaofeng11:
@ricardoquesada @linshun  Ok, I'll add a self defined exception to replace std::exception.  I have talk with minggo about ASSERT, and our conclusion is ASSERT can give more help for game developers, ithelp game developer quickly find param error.  "safe mode" is friendly to editor developers, but there's no other cocos editor developer other than ourself.

ricardoquesada:
@xiaofeng11 I think cocos2d-x should keep using `asserts` because it is working like that now... so mixing asserts and exceptions is not good. Either asserts or exceptions, but not both.
But I think we should stop using asserts (unless there is a precondition) and try to fail in a safe way.
For example, if we try to create an sprite with an invalid filename and an invalid shader file, I think cocos2d-x will crash. It shouldn't crash... it should return a default texture and default shader and log that in console or something.

xiaofeng11:
@ricardoquesada So, how about overload these function with a new implement that through exception and won't use ASSERT?

Create sprite with default texture when we get a invalid filename is excellent, but we have to mix cocos2dx and cocostudio in next version, there's won't enough time to make these change done.

Edit : Sorry, I make a mistake in above talk, we hope editor and game engine use the same implementation, so the effect in editor will match the effect in game best. If we overload the functions, in future update, these function may have difference, and cause editor can not match game engine. Maybe split these function into param check function and implement function is a choice, we can add a new param check function for editor.

ricardoquesada:
@xiaofeng11 whatever we do, it must be consistent with the whole engine.

+1 for Engine and editor using the same codebase.
-1 for having inconsistent code
-1 for adding hacks because we don't have time. I've been hearing that I joined Chukong (2 years ago).

I think what you (and the rest of the editor team should do) is discuss these changes before sending a PR. For example opening a bug saying:
  "cocos2d-x crashes when X, Y, Z and that is big problem because that makes Cocos Studio crash. We want to replace asserts with exceptions... what is the best practice here?..."

And we can help you with that. We also want our users to have a good UX in Cocos Studio.

xiaofeng11:
@ricardoquesada We have discus in editor team, and we think the exception can be removed. The reason we modify these code is we hope in release version, ASSERT can return from the function, but not crash our program. Refer the discus in http://stackoverflow.com/questions/37473/how-can-i-assert-without-using-abort, GTK use macro RETURN_IF_FAIL & RETURN_VAL_IF_FAIL to avoid crash, can we use similar resolution? We have start a discus on cocos2dx forum: http://discuss.cocos2d-x.org/t/make-assert-safe-in-release-mode/25115, Please help us. Thank you very much!

PR #14396, opened 2015-11-16
Combine Studio change for spine

xiaofeng11:
Add export statement to function and class to make them can be compile as dynamic library in cocostudio project


minggo:
As i know, we can not modify spine codes because of its license.

xiaofeng11:
@minggo Just talk with XieXin we will use static lib of spine, so we won't need modify it any more. But the license of spine is conflict with cocos2dx license, why does it have been embed?

minggo:
Because spine is well, many developers want it.
If developers modify the spine code, it is nothing about cocos2d-x, but we can not modify it.

xiaofeng11:
I known, Thanks a lot. I'll close this PR.

PR #14399, opened 2015-11-16
Combine Studio change for compatible with 32 bit Mac System

xiaofeng11:
Cocos studio is compile as 32 bit executable file, so the code must be compatible with 32 bit compiler.

### DIFF
--- a/cocos/network/HttpAsynConnection-apple.h
+++ b/cocos/network/HttpAsynConnection-apple.h
@@ -34,6 +34,18 @@
 /// @cond
 @interface HttpAsynConnection : NSObject <NSURLConnectionDelegate, NSURLConnectionDataDelegate>
 {
+    NSString *srcURL;
+    NSString *sslFile;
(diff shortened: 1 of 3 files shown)


minggo:
It looks good to me.
Should coco studio continue to support 32 bit?

xiaofeng11:
@minggo Yes, on both Windows and MacOSX, Cocos Studio executable file is compiled as 32bit executable. And we won't have schedule to upgrade it to 64bit now, because some 3rd lib use in Cocos Studio is only support 32bit.

minggo:
Good to know, thanks.
So we will remove 32bit if all 3rd lib support 64bit?

xiaofeng11:
On MacOSX platform 64bit is enough for now, but on Windows platform, there's still many user use 32bit system, so I think we should continue 32bit support on Windows, unless Microsoft give up 32bit  Windows all user upgrade to 64bit system.

PR #14409, opened 2015-11-17
Combine Studio change for openGL related operation

xiaofeng11:
Cocos Studio will init openGL itself, so we add a param to CCGLViewImpl-desktop constructor to let user choose whether cocos2dx will init openGL or not.

In 3d scene, cocos studio will use _vertShader & _fragShader in CCGLProgram to detect click on 3d item, but after glLinkProgram called it have been deleted and cleaned, we changed code to save them, and clean them when reload shader or destruct class.

### DIFF
--- a/cocos/renderer/CCGLProgram.cpp
+++ b/cocos/renderer/CCGLProgram.cpp
@@ -181,7 +181,6 @@ GLProgram::GLProgram()
 , _vertShader(0)
 , _fragShader(0)
 , _flags()
-, _isShaderClearable(true)
 {
     _director = Director::getInstance();
     CCASSERT(nullptr != _director, "Director is null when init a GLProgram");
@@ -221,11 +220,7 @@ bool GLProgram::initWithByteArrays(const GLchar* vShaderByteArray, const GLchar*
     std::string replacedDefines = "";
     replaceDefines(compileTimeDefines, replacedDefines);
 
-    if (_isShaderClearable)
-        // When _isShaderClearable = true, vertShader & _fragShader has been already released, won't need release again.
-        _vertShader = _fragShader = 0;
-    else
-        clearShader();
+    _vertShader = _fragShader = 0;
 
     if (vShaderByteArray)
     {
@@ -548,12 +543,7 @@ bool GLProgram::link()
     parseVertexAttribs();
     parseUniforms();
 
-    // Keep _vertShader & _fragShader for editor by set _isShaderClearable to false,
-    //  because them are needed for 3d object click detection in 3d scene editing.
-    if (_isShaderClearable)
-    {
-        clearShader();
-    }
+    clearShader();
 
 #if DEBUG || (CC_TARGET_PLATFORM == CC_PLATFORM_WINRT)
     glGetProgramiv(_program, GL_LINK_STATUS, &status);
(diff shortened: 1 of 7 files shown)


minggo:
> Cocos Studio will init openGL itself, so we add a param to CCGLViewImpl-desktop constructor to let user choose whether cocos2dx will init openGL or not.

It looks good to me.

cocos/renderer/CCGLProgram.cpp
```diff
@@ -935,5 +913,20 @@ void GLProgram::reset()
...
+        glDeleteShader(_vertShader);
+    }
+
+    if (_fragShader)
+    {
+        glDeleteShader(_fragShader);
+    }
+
+    _vertShader = _fragShader = 0;
+}
```
minggo, line 931: I think it is better to remove them in mobile platform, because it will consume some memory.

cocos/renderer/CCGLProgram.cpp
```diff
@@ -230,7 +220,7 @@ bool GLProgram::initWithByteArrays(const GLchar* vShaderByteArray, const GLchar*
     std::string replacedDefines = "";
     replaceDefines(compileTimeDefines, replacedDefines);
 
-    _vertShader = _fragShader = 0;
+    clearShader();
```
minggo: Why replace it with `clearShader()` here?

cocos/renderer/CCGLProgram.cpp
minggo, line 931: And it is better to add some comment for why not delete them in PC platform.

cocos/renderer/CCGLProgram.cpp
xiaofeng11: Because in this change we won't delete these shaders after linked, if only clear these two mark, the shader loaded will never been deleted, it will cause memory leak.

xiaofeng11:
@minggo I've change the code, now save _vertShader & _fragShader only on PC platform. On mobile platform, it will works as before.

minggo:
Thanks.

xiaofeng11:
@ricardoquesada Please review this PR.

super626:
Ok for me.

cocos/renderer/CCGLProgram.cpp
```diff
@@ -553,17 +547,11 @@ bool GLProgram::link()
...
-    {
-        glDeleteShader(_fragShader);
-    }
-
-    _vertShader = _fragShader = 0;
+#if CC_TARGET_PLATFORM != CC_PLATFORM_MAC && CC_TARGET_PLATFORM != CC_PLATFORM_WIN32
+    // Keep _vertShader & _fragShader on PC for cocos studio, because them 
+    //  are needed for 3d object click detection in 3d scene editing.
+    clearShader();
+#endif
```
ricardoquesada: This logic is not correct. 
If Cocos Studio needs a feature, then it should say either: `#if COCOS_STUDIO...` or add a methods that prevents clearing the shaders like `void dontClearShaders()`.

xiaofeng11:
@ricardoquesada @minggo The clearShader logic have been changed, now add a clearable mark with a setter to control if shader will be cleared after link. Please review changes, Thanks.

ricardoquesada:
looks good. Please add a test for this. thanks.

xiaofeng11:
@minggo  @ricardoquesada liamcindy has find a new way to detect 3d object click, so we reverse modify for disable shader clear setting, and export MeshVertexData. Please review, thanks.

ricardoquesada:
looks better now

xiaofeng11:
@ricardoquesada Could this PR be merged? Now most change in it is add export declaration

PR #14419, opened 2015-11-18
Combine Studio change for UI components

xiaofeng11:
This PR can not pass auto build, because it need 2d/CocosStudioExtension.h file in 2d component combine pr.

This is a big PR. But most modify is to save resource file name, and add function that can return these resource file name back. And some other function modify is to clean / replace resource already loaded. Because in editor, resource may change often, not like in game, after resource loaded, it never change.

UILayoutComponent.cpp is changed for make sub component do the right layout operation by Liam.

### DIFF
--- a/cocos/ui/UIAbstractCheckButton.cpp
+++ b/cocos/ui/UIAbstractCheckButton.cpp
@@ -136,12 +136,6 @@ void AbstractCheckButton::loadTextures(const std::string& backGround,
 void AbstractCheckButton::loadTextureBackGround(const std::string& backGround,TextureResType texType)
 {
     _backGroundFileName = backGround;
-#ifndef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
-    if (backGround.empty())
-    {
-        return;
-    }
-#endif
 
     _backGroundTexType = texType;
     switch (_backGroundTexType)
@@ -178,12 +172,6 @@ void AbstractCheckButton::loadTextureBackGroundSelected(const std::string& backG
 {
     _backGroundSelectedFileName = backGroundSelected;
     _isBackgroundSelectedTextureLoaded = !backGroundSelected.empty();
-#ifndef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
-    if (backGroundSelected.empty())
-    {
-        return;
-    }
-#endif
 
     _backGroundSelectedTexType = texType;
     switch (_backGroundSelectedTexType)
@@ -215,12 +203,6 @@ void AbstractCheckButton::setupBackgroundSelectedTexture()
 void AbstractCheckButton::loadTextureFrontCross(const std::string& cross,TextureResType texType)
 {
     _frontCrossFileName = cross;
-#ifndef CC_STUDIO_ENABLED_VIEW   // for cocostudio only
-    if (cross.empty())
-    {
-        return;
-    }
-#endif
 
     _frontCrossTexType = texType;
     switch (_frontCrossTexType)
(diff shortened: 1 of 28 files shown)


xiaofeng11:

minggo:
@zilongshanren 
Will you please help to review it?

cocos/ui/UIAbstractCheckButton.cpp
```diff
@@ -53,6 +54,11 @@ _frontCrossDisabledTexType(TextureResType::LOCAL),
 _zoomScale(0.1f),
 _backgroundTextureScaleX(1.0),
 _backgroundTextureScaleY(1.0),
+_backGroundFileName(""),
```
zilongshanren, line 57: why add these files names? I think it is only useful for Editor. Why not persist it in the Editor side?

cocos/ui/UISlider.cpp
```diff
@@ -723,6 +744,42 @@ void Slider::copySpecialProperties(Widget *widget)
     }
 }
 
+ResouceData Slider::getBackFile()
```
zilongshanren, line 717: These APIs  `getBackFile`, `getProgressBarFile` are only used by CocoStudio?

cocos/ui/UIAbstractCheckButton.cpp
xiaofeng11, line 57: Because lua extension for editor set these resource name by api call, it not set in editor. So we should save these name in ui component and when editor export project, get these file name from ui components.

cocos/ui/UISlider.cpp
xiaofeng11, line 717: @zilongshanren Yes, most modify by studio is editor only, but @minggo think these function can be part of engine. Please refer comments in https://github.com/cocos2d/cocos2d-x/pull/14372

zilongshanren, line 717: @xiaofeng11  
I see, I have no question now.

zilongshanren, line 717: @xiaofeng11  
I see, I have no question now.

zilongshanren:
@minggo  I'm OK with this PR now.

minggo:
@zilongshanren 
Thanks. I think we can merge it after passing jenkins.

xiaofeng11:
@minggo This pr can not pass jenkins, because 2d/CocosStudioExtension.h already add in 2d component pr, so I haven't add it to this PR, but this PR need the header file to compile. Shall I add the header file to this PR?

minggo:
@xiaofeng11
Yep, i think you should add needed codes to pass the jenkins.

xiaofeng11:
[ci rebuild]

xiaofeng11:
Condition macro removed, and update implement logic.
Now if  set resource of ui component with a blank file name, it won't return immediately, it will continue run follow codes and clean resource already loaded.

minggo:
@zilongshanren 
Please review it, and merge it if you think it is ok.
Thanks.

tools/bindings-generator
```diff
@@ -1 +1 @@
-Subproject commit 2318a1593d2d490a3860179acaf81a1bf867a36b
+Subproject commit a0fe326ea3d685d342ba2a61c0be6f9dee83f800
```
zilongshanren, line 1: @xiaofeng11 Do you commit this accidentally?

ricardoquesada:
My comments are:
- every time that we fix or add a new functionality to the engine, we need to add a test that proves that the fix and/or new functionality works correctly.

PR #14420, opened 2015-11-18
Combine Studio change for particle 3d

xiaofeng11:
Add dll_export statement to particle 3d related class to make it can be use when cocos2d-x compiled as dynamic library.
Move initSystem function in PUParticleSystem3D class to public, so editor can change particle resource of a particle 3d object.

### DIFF
--- a/extensions/ExtensionExport.h
+++ b/extensions/ExtensionExport.h
@@ -24,6 +24,8 @@
             #define NULL    ((void *)0)
         #endif
     #endif
+#elif defined(_SHARED_)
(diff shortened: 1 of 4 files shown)


minggo:
It looks good to me.

@super626 
Could you help to take a look?

super626:
@minggo 
It is ok.

minggo:
I think @ricardoquesada takes charge of these codes.
@ricardoquesada 
Will you please take a look too?

ricardoquesada:
looks good to me. I'll merge it (I think @super626 is the owner of Particle3D since it belongs to 3d).

Issue #12335, opened 2015-06-15
HttpRequest.h inline function uses deprecated API.


xpol:
In HttpRequest.h:
function: [void HttpRequest::setResponseCallback(Ref\* pTarget, SEL_CallFuncND pSelector)](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/network/HttpRequest.h#L245) uses deprecated function [void HttpRequest::setResponseCallback(Ref\* pTarget, SEL_HttpResponse pSelector)](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/network/HttpRequest.h#L2550). This cause any cpp file includes "HttpRequest.h" will raise a deprecation warning.

To avoid this:
- Move `void HttpRequest::setResponseCallback(Ref* pTarget, SEL_CallFuncND pSelector)` implementation in to cpp file, 
- Or implements `void HttpRequest::setResponseCallback(Ref* pTarget, SEL_CallFuncND pSelector)` inline with non-deprecated API.

BTW, I suggest to check other headers.

xpol:
@pandamicro 

This is very boring, just move these function body to cpp would fixes this.
Do it please!

Orz

zilongshanren:
@xpol 
I think it's the best time for you to contribute :smile:

xpol:
Here we go #14423

minggo:
close it since #14423 is merged.

PR #14423, opened 2015-11-18
Fixes boring deprecated warning.

xpol:
I know I duplicated the code but there is no cpp for me to move body into.

### DIFF
--- a/cocos/network/HttpRequest.h
+++ b/cocos/network/HttpRequest.h
@@ -242,7 +242,15 @@ public:
      */
     CC_DEPRECATED_ATTRIBUTE inline void setResponseCallback(Ref* pTarget, SEL_CallFuncND pSelector)
     {
-        setResponseCallback(pTarget, (SEL_HttpResponse) pSelector);
+        // This function have duplicated body with setResponseCallback(Ref* , SEL_HttpResponse )
+        // You must repeat youself in that function when modified.
+        _pTarget = pTarget;
+        _pSelector = (SEL_HttpResponse)pSelector;
+
+        if (_pTarget)
+        {
+            _pTarget->retain();
+        }
     }
     
     /**
@@ -254,6 +262,8 @@ public:
      */
     CC_DEPRECATED_ATTRIBUTE inline void setResponseCallback(Ref* pTarget, SEL_HttpResponse pSelector)
     {
+        // This function have duplicated body with setResponseCallback(Ref* , SEL_CallFuncND )
+        // You must repeat youself in that function.
         _pTarget = pTarget;
         _pSelector = pSelector;
         
--- a/cocos/network/HttpRequest.h
+++ b/cocos/network/HttpRequest.h
@@ -242,15 +242,7 @@ public:
      */
     CC_DEPRECATED_ATTRIBUTE inline void setResponseCallback(Ref* pTarget, SEL_CallFuncND pSelector)
     {
-        // This function have duplicated body with setResponseCallback(Ref* , SEL_HttpResponse )
-        // You must repeat youself in that function when modified.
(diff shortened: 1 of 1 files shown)


xpol:
ref #12335

cocos/network/HttpRequest.h
```diff
@@ -254,6 +262,8 @@ class CC_DLL HttpRequest : public Ref
      */
     CC_DEPRECATED_ATTRIBUTE inline void setResponseCallback(Ref* pTarget, SEL_HttpResponse pSelector)
     {
+        // This function have duplicated body with setResponseCallback(Ref* , SEL_CallFuncND )
```
zilongshanren: I think deprecated functions should call non-deprecated functions.

zilongshanren: Here is a bad code smell here, DRY :smile:

zilongshanren:
@xpol  :+1:  thanks.

cocos/network/HttpRequest.h
```diff
@@ -342,6 +336,18 @@ class CC_DLL HttpRequest : public Ref
    		return _headers;
    	}
     
+private:
+    inline void doSetResponseCallback(Ref* pTarget, SEL_HttpResponse pSelector)
```
zilongshanren, line 340: why use another private function here... it's a bit wired...
What do you think?

zilongshanren, line 340: It will cause the function `setResponseCallback` hard to read. I have to dig into the `doSetResponseCallback` function.

I think it's ok to put the implementation of `doSetResponseCallback` directly in `setResponseCallback`

xpol:
Just realized that we needs introduce a private helper function do the stuff.

zilongshanren:
@xpol 
Why we need a helper method here, it will cause another indirectly function call.

xpol:
Which to choose:
1. Duplicated code in `void setResponseCallback(Ref* pTarget, SEL_CallFuncND pSelector)` and  `void setResponseCallback(Ref* pTarget, SEL_HttpResponse pSelector)` as I have done in 1st commit.
2. Introduce private helper method to remove duplicate logic in 1, as I have done in 2nd commit.
3. Add a cpp file, move `void setResponseCallback(Ref* pTarget, SEL_CallFuncND pSelector)` body in to that cpp, and keep it call to `void setResponseCallback(Ref* pTarget, SEL_HttpResponse pSelector)` as it was. Let report warnings in cocos2d-x lib. User of cocos2d-x will not got warnings.

zilongshanren:
@xpol 

Make the original deprecated function implementation into the new recommended one is a best practice here.

When we read code of the deprecated function, we know exactly it's redirect to another function.

I hope you could change it to confirm this best practice.

Thanks.

zilongshanren:
@xpol 
Don't need such complexity.  You might want to give `CCLayer.h` a look and you will get what I mean here.

xpol:
@zilongshanren 

The only valid API is `void setResponseCallback(const ccHttpRequestCallback& callback)`.

You mean std::bind ?

xpol:
Or I can use lambda.

zilongshanren:
@xpol 
Sorry... I think your second commit is better...
Could you rollback to the 2nd commit and force push your branch...
Sorry again for the mistake.

I thought only one function is deprecated, but both `setResponseCallback`  are deprecated..
My bad.. Hope you don't mind.

xpol:
OK! but wait, I'm not under the operation system where I worked with the cocos2d-x repo.

xpol:
Reverted. @zilongshanren

zilongshanren:
@xpol  :+1:  Thanks.

PR #14426, opened 2015-11-18
Fix crash when using cocos2d::Controller on iOS6

pyrosphere:
Using Controller on iOS6 crashes the app because GameController.framework was only added on iOS7. 

This PR fixes that crash by doing nothing in Controller::startDiscoveryController if the class GCController is not available.


minggo:
Thanks.

PR #14429, opened 2015-11-19
Combine Studio change for get program file full path.

xiaofeng11:
_get_wpgmptr function will cause ASSERT in system library when program is start by C#, so it can not use in Cocos Studio.

@pandamicro @minggo Please reivew this PR.

### DIFF
--- a/cocos/platform/win32/CCFileUtils-win32.cpp
+++ b/cocos/platform/win32/CCFileUtils-win32.cpp
@@ -136,14 +136,9 @@ static void _checkPath()
 {
     if (s_resourcePath.empty())
     {
-        WCHAR *pUtf16ExePath = nullptr;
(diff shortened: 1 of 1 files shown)


xiaofeng11:
Remove condition macro, use GetModuleFileNameW in all case, it is more safe.

minggo:
It looks good to me.
Just need to run cocos2d-x tests on windows to check if it can work or not.

xiaofeng11:
[ci rebuild]

xiaofeng11:
@minggo I have run cpp-test, js-test and lua-test under windows system, all works fine.

xiaofeng11:
@minggo I've test and confirm this change is OK for all test under windows system, and before "FileUtils" start use _get_wpgmptr get current program path, old version is use GetModuleFileName to get program path. Could this PR been merged?

PR #14435, opened 2015-11-19
Add Overflow to Label

zilongshanren:
This PR add many good Overflow features to new label.
### Add three different type of Overflow to Label.

```
 enum class Overflow
 {
        //for keep back compatibility
        NORMAL,

        /**
         *In CLAMP mode, when label content goes out of the bounding box, it will be clipped.
         */
        CLAMP,
      /**
         * In SHRINK mode, the font size will change dynamically to adapt the content size.
         */
        SHRINK,
        /**
         *In RESIZE_HEIGHT mode, you can only change the width of label and the height is changed automatically.
         */
        RESIZE_HEIGHT
 };

```
### Added APIs:

```
    /**
     * Change font size of label type BMFONT
     * Note: This function only scale the BMFONT letter to mimic the font size change effect.
     *
     * @param fontSize The desired font size in float.
     */
    void setBMFontSize(float fontSize);

    /**
     * Return the user define BMFont size.
     *
     * @return The BMFont size in float value.
     */
    float getBMFontSize()const;

    /**
     * Toggle wrap option of the label.
     * Note: System font doesn't support manually toggle wrap.
     *
     * @param enable Set true to enable wrap and false to disable wrap.
     */
    void enableWrap(bool enable);

    /**
     * Query the wrap is enabled or not.
     * Note: System font will always return true.
     */
    bool isWrapEnabled()const;

    /**
     * Change the label's Overflow type, 
     * currently only TTF and BMFont support all the valid Overflow type.
     *  Char Map font also supports all the Overflow type except for SHRINK.
     *
     * @param overflow   see `Overflow`
     */
    void setOverflow(Overflow overflow);

    /**
     * Query the label's Overflow type.
     *
     * @return see `Overflow`
     */
    Overflow getOverflow()const;

```
### Limitations:

Currently the System font doesn't support these Overflow feature and Char Map font doesn't support `Shrink` feature.

@WenhaiLin  @dabingnn 

please help to review this PR, thanks.


cocos/2d/CCFontAtlas.cpp
```diff
@@ -157,6 +157,19 @@ void FontAtlas::addLetterDefinition(char16_t utf16Char, const FontLetterDefiniti
     _letterDefinitions[utf16Char] = letterDefinition;
 }
 
+void FontAtlas::scaleFontLetterDefinition(float scaleFactor)
+{
+    for (auto&& fontDefinition : _letterDefinitions) {
+        FontLetterDefinition letterDefinition = fontDefinition.second;
```
WenhaiLin:
``` cpp
auto& letterDefinition = fontDefinition.second;
...
// _letterDefinitions[fontDefinition.first] = letterDefinition; //this line should be removed
```

cocos/2d/CCLabel.cpp
```diff
@@ -548,46 +555,37 @@ void Label::setFontAtlas(FontAtlas* atlas,bool distanceFieldEnabled /* = false *
...
-
-bool Label::setBMFontFilePath(const std::string& bmfontFilePath, const Vec2& imageOffset /* = Vec2::ZERO */)
-{
-    FontAtlas *newAtlas = FontAtlasCache::getFontAtlasFNT(bmfontFilePath,imageOffset);
-
-    if (!newAtlas)
-    {
-        reset();
-        return false;
+    if(fontSize > 0){
```
WenhaiLin: I think this line should be:

``` cpp
else if(fontSize > 0.0f){
```

PR #14438, opened 2015-11-19
Add windowActionModeOverlay style to main activity and fixes #14401.

### LINKED ISSUE #14401: UIEditBox not aligned when the text been selected.
As the follow picture shows, before select (left), it works fine.
Then I select the text, the  "select cut copy paste" toolbar shows, and it pushed the UIEditBox down (right).

uieditbox

I have tested under Android 5.1.1 (Mi note with original Android 5.1.1 ROM.)

xpol:
Finally the solution is set `windowActionModeOverlay` to be `true`.


zilongshanren:
@xpol 
:+1: 
Sorry for the late reply, I will test this PR later.

I'm very busy with cocos creator these day, sorry for this.

PR #14441, opened 2015-11-20
Fix #14440

### LINKED ISSUE #14440: A serious bug in new UIPageView
In new UIPageView, getPages function has been deprecated, and add a ASSERT in entrance of the function, when it been called, the program will crash. But in LayoutComponent::refreshLayout() function, getPages() is still been using, this make program can never run.

@neokim @zilongshanren Please check & fix this issue

neokim:
Fix for https://github.com/cocos2d/cocos2d-x/issues/14440


zilongshanren:
:+1:  Thanks for the fix.

PR #14446, opened 2015-11-20
Fixed AssetsManager bug.

cpascal:
1. Incorrect argument to Downloader::createDownloadFileTask()
2. In Visual C++ environment, string formatter doesn't support "%zd",
   now using PRIdPTR definition of inttypes.h


extensions/assets-manager/AssetsManager.cpp
```diff
@@ -176,7 +187,7 @@ void AssetsManager::checkStoragePath()
 static std::string keyWithHash( const char* prefix, const std::string& url )
 {
     char buf[256];
-    sprintf(buf,"%s%zd",prefix,std::hash<std::string>()(url));
+    sprintf(buf,"%s%" PRIdPTR,prefix,std::hash<std::string>()(url));
```
minggo, line 190: Sorry, i don't quite understand the usage.

PR #14447, opened 2015-11-20
Fix CCFileUtils-win32.cpp for CMake (mingw)

dariodepaolis:
This is needed for fixing compiling with CMake (mingw) under CLion.
for this error: undefined reference to `getwpgmptr'

And then the Resources path is not correctly working, even if it works under Visual Studio, but I don't know if it's related


minggo:
I will close this PR since it is fixed.

PR #14448, opened 2015-11-20
Fix UIEditBoxImpl-win32.cpp for CMake

dariodepaolis:
Compiling with cmake I have this error: "invalid user-defined conversion"

In this way it works, I hope this can be useful


minggo:
I will close this PR since the code changed, and it doesn't use `TIMERPROC` any more.

PR #14451, opened 2015-11-21
Improve unicode support

### LINKED ISSUE #14444: Font system not follow Unicode Standard 2.0 or later.
As listed [here](http://unicode.org/faq/utf_bom.html) and referenced [here](https://en.wikipedia.org/wiki/Unicode).

> Q: Is Unicode a 16-bit encoding?
> 
> A: No. The first version of Unicode was a 16-bit encoding, from 1991 to 1995, but starting with Unicode 2.0 (July, 1996), it has not been a 16-bit encoding. The Unicode Standard encodes characters in the range U+0000..U+10FFFF, which amounts to a 21-bit code space. Depending on the encoding form you choose (UTF-8, UTF-16, or UTF-32), each character will then be represented either as a sequence of one to four 8-bit bytes, one or two 16-bit code units, or a single 32-bit code unit.

The follow code is not corrent:

``` C++
int * FontFreeType::getHorizontalKerningForTextUTF16(const std::u16string& text, int &outNumLetters) const
{
    if (!_fontRef)
        return nullptr;

    outNumLetters = static_cast<int>(text.length());

    if (!outNumLetters)
        return nullptr;

    int *sizes = new int[outNumLetters];
    if (!sizes)
        return nullptr;
    memset(sizes,0,outNumLetters * sizeof(int));

    bool hasKerning = FT_HAS_KERNING( _fontRef ) != 0;
    if (hasKerning)
    {
        for (int c = 1; c < outNumLetters; ++c)
        {
            sizes[c] = getHorizontalKerningForChars(text[c-1], text[c]);
        }
    }

    return sizes;
}
```

so is the function `unsigned char* FontFreeType::getGlyphBitmap(unsigned short theChar, long &outWidth, long &outHeight, Rect &outRect,int &xAdvance)`.

The problem is:
1. number of code points is not length of std::u16string.
2. can't just get code point via `text[c]`.
3. getGlyphBitmap should use `uint32_t` for code point (1st arg), as `sizeof(unsigned short) == 2` is at least true under iOS.

xpol:
Try to fixes #14444
1. refactor and add full set UTF convert functions to StringUtils.
2. replace all usage to UTF16 with UTF32 in Font/Label system.
3. add deprecated inline implement to old public APIs.

### DIFF
--- a/cocos/2d/CCFont.h
+++ b/cocos/2d/CCFont.h
@@ -30,6 +30,7 @@
 
 #include <string>
 #include "base/ccTypes.h"
+#include "base/ccUTF8.h"
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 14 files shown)


xpol:
@zilongshanren 
Seems you are active here.
Would please have a look at this PR?

zilongshanren:
@xpol 
Why use u32 to replace u16?

xpol:
@zilongshanren 
see #14444, Unicode is not 16 bit encoding.

PR #14458, opened 2015-11-23
add checking for malloc result

yszheda:
malloc fails to allocate memory will cause crash when calling fread


minggo:
@yszheda Thanks, and i think we should initialize `fileSize` too.

PR #14459, opened 2015-11-23
New API in ui::EditBox(Android and IOS) setTextAlignment and setPadding.

anjandaffo:
Above API is work for Android and IOS.

From these API User can set InnerPadding in ui editbox.
Now editbox support RTL. 
User can set text alignment in ui editbox.

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
@@ -102,16 +102,6 @@ public class Cocos2dxEditBoxHelper {
                 editBox.setTextColor(Color.WHITE);
                 editBox.setSingleLine();
                 editBox.setOpenGLViewScaleX(scaleX);
-                Resources r = mCocos2dxActivity.getResources();
-                float density =  r.getDisplayMetrics().density;
-                int paddingBottom = (int)(height * 0.33f / density);
-                paddingBottom = convertToSP(paddingBottom  - 5 * scaleX / density);
-                paddingBottom = paddingBottom / 2;
-                int paddingTop = paddingBottom;
-                int paddingLeft = (int)(5 * scaleX / density);
-                paddingLeft = convertToSP(paddingLeft);
-
-                editBox.setPadding(paddingLeft,paddingTop, 0, paddingBottom);
 
 
                 FrameLayout.LayoutParams lParams = new FrameLayout.LayoutParams(
@@ -382,7 +372,42 @@ public class Cocos2dxEditBoxHelper {
         });
     }
 
+    public static void setInnerPadding(final int index, final float left, final float top, final float right, final float bottom){
+        mCocos2dxActivity.runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+                Cocos2dxEditBox editBox = mEditBoxArray.get(index);
+                if (editBox != null) {
+                    float mScaleX = editBox.getOpenGLViewScaleX();
+                    Resources r = mCocos2dxActivity.getResources();
+                    float density =  r.getDisplayMetrics().density;
+                    int paddingLeft = (int)(left * mScaleX / density);
+                    paddingLeft = convertToSP(paddingLeft);
+                    int paddingTop = (int)(top * mScaleX / density);
+                    paddingTop = convertToSP(paddingTop);
+                    int paddingRight = (int)(right * mScaleX / density);
+                    paddingRight = convertToSP(paddingRight);
+                    int paddingBottom = (int)(bottom * mScaleX / density);
+                    paddingBottom = convertToSP(paddingBottom);
 
+                    editBox.setPadding(paddingLeft,paddingTop, paddingRight, paddingBottom);
+
+                }
+            }
+        });
+    }
+
+    public static void setTextAlignment(final int index, final int hAlign, final int vAlign){
+        mCocos2dxActivity.runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+                Cocos2dxEditBox editBox = mEditBoxArray.get(index);
+                if (editBox != null) {
+                    editBox.setInputTextAlignment(hAlign, vAlign);
+                }
+            }
+        });
+    }
 
     public static void openKeyboard(int index) {
         final InputMethodManager imm = (InputMethodManager) mCocos2dxActivity.getContext().getSystemService(Context.INPUT_METHOD_SERVICE);
--- a/cocos/ui/UIEditBox/UIEditBox.cpp
+++ b/cocos/ui/UIEditBox/UIEditBox.cpp
@@ -353,6 +353,22 @@ void EditBox::setReturnType(EditBox::KeyboardReturnType returnType)
     }
 }
 
+void EditBox::setPadding(float left, float top, float right, float bottom)
+{
+    if (_editBoxImpl != nullptr )
+    {
+        _editBoxImpl->setPadding(left, top, right, bottom);
+    }
+}
+
+void EditBox::setTextAlignment(TextHAlignment hAlign, TextVAlignment vAlign)
+{
+    if (_editBoxImpl != nullptr)
+    {
+        _editBoxImpl->setTextAlignment(hAlign, vAlign);
+    }
+}
+    
 /* override function */
 void EditBox::setPosition(const Vec2& pos)
 {
--- a/cocos/ui/UIEditBox/UIEditBox.cpp
+++ b/cocos/ui/UIEditBox/UIEditBox.cpp
@@ -45,6 +45,12 @@ EditBox::EditBox(void)
 , _colPlaceHolder(Color3B::GRAY)
 , _maxLength(0)
 , _adjustHeight(0.0f)
+, _leftPadding(5.0f)
+, _rightPadding(5.0f)
+, _topPadding(5.0f)
+, _bottomPadding(5.0f)
+, _textHorizontalAlignMent(TextHAlignment::LEFT)
+, _textVerticalAlignMent(TextVAlignment::CENTER)
 #if CC_ENABLE_SCRIPT_BINDING
 , _scriptEditBoxHandler(0)
 #endif
@@ -123,6 +129,10 @@ bool EditBox::initWithSizeAndBackgroundSprite(const cocos2d::Size &size, cocos2d
         
         this->addTouchEventListener(CC_CALLBACK_2(EditBox::touchDownAction, this));
         
+        this->setPadding(_leftPadding, _topPadding, _rightPadding, _bottomPadding);
+        
+        this->setTextAlignment(_textHorizontalAlignMent, _textVerticalAlignMent);
+        
         return true;
     }
     return false;
@@ -157,6 +167,10 @@ bool EditBox::initWithSizeAndBackgroundSprite(const Size& size,
         this->setTouchEnabled(true);
         
         this->addTouchEventListener(CC_CALLBACK_2(EditBox::touchDownAction, this));
+        
+        this->setPadding(_leftPadding, _topPadding, _rightPadding, _bottomPadding);
+        
+        this->setTextAlignment(_textHorizontalAlignMent, _textVerticalAlignMent);
         
         return true;
     }
@@ -355,18 +369,68 @@ void EditBox::setReturnType(EditBox::KeyboardReturnType returnType)
 
 void EditBox::setPadding(float left, float top, float right, float bottom)
 {
+    _leftPadding = left;
+    _topPadding = top;
+    _rightPadding = right;
+    _bottomPadding = bottom;
+    
     if (_editBoxImpl != nullptr )
     {
         _editBoxImpl->setPadding(left, top, right, bottom);
     }
 }
 
+float EditBox::getLeftPadding()
+{
+    return _leftPadding;
+}
+
+float EditBox::getTopPadding()
+{
+    return _topPadding;
+}
+
+float EditBox::getRightPadding()
+{
+    return _rightPadding;
+}
+
+float EditBox::getBottomPadding()
+{
+    return _bottomPadding;
+}
+    
 void EditBox::setTextAlignment(TextHAlignment hAlign, TextVAlignment vAlign)
 {
+    _textHorizontalAlignMent = hAlign;
+    _textVerticalAlignMent = vAlign;
+    
     if (_editBoxImpl != nullptr)
     {
         _editBoxImpl->setTextAlignment(hAlign, vAlign);
     }
+}
+
+void EditBox::setHorizontalTextAlignment(TextHAlignment hAlign)
+{
+    _textHorizontalAlignMent = hAlign;
+    this->setTextAlignment(_textHorizontalAlignMent, _textVerticalAlignMent);
+}
+
+TextHAlignment EditBox::getHorizontalTextAlignment()
+{
+    return _textHorizontalAlignMent;
+}
+
+void EditBox::setVerticalTextAlignment(TextVAlignment vAlign)
+{
+    _textVerticalAlignMent = vAlign;
+    this->setTextAlignment(_textHorizontalAlignMent, _textVerticalAlignMent);
+}
+
+TextVAlignment EditBox::getVerticalTextAlignment()
+{
+    return _textVerticalAlignMent;
 }
     
 /* override function */
--- a/cocos/ui/UIEditBox/UIEditBox.cpp
+++ b/cocos/ui/UIEditBox/UIEditBox.cpp
@@ -49,8 +49,8 @@ EditBox::EditBox(void)
 , _rightPadding(5.0f)
 , _topPadding(5.0f)
 , _bottomPadding(5.0f)
-, _textHorizontalAlignMent(TextHAlignment::LEFT)
-, _textVerticalAlignMent(TextVAlignment::CENTER)
+, _textHorizontalAlignment(TextHAlignment::LEFT)
+, _textVerticalAlignment(TextVAlignment::CENTER)
 #if CC_ENABLE_SCRIPT_BINDING
 , _scriptEditBoxHandler(0)
 #endif
@@ -131,7 +131,7 @@ bool EditBox::initWithSizeAndBackgroundSprite(const cocos2d::Size &size, cocos2d
         
         this->setPadding(_leftPadding, _topPadding, _rightPadding, _bottomPadding);
         
-        this->setTextAlignment(_textHorizontalAlignMent, _textVerticalAlignMent);
+        this->setTextAlignment(_textHorizontalAlignment, _textVerticalAlignment);
         
         return true;
     }
@@ -170,7 +170,7 @@ bool EditBox::initWithSizeAndBackgroundSprite(const Size& size,
         
         this->setPadding(_leftPadding, _topPadding, _rightPadding, _bottomPadding);
         
-        this->setTextAlignment(_textHorizontalAlignMent, _textVerticalAlignMent);
+        this->setTextAlignment(_textHorizontalAlignment, _textVerticalAlignment);
         
         return true;
     }
@@ -380,30 +380,30 @@ void EditBox::setPadding(float left, float top, float right, float bottom)
     }
 }
 
-float EditBox::getLeftPadding()
+float EditBox::getLeftPadding() const
 {
     return _leftPadding;
 }
 
-float EditBox::getTopPadding()
+float EditBox::getTopPadding() const
 {
     return _topPadding;
 }
 
-float EditBox::getRightPadding()
+float EditBox::getRightPadding() const
 {
     return _rightPadding;
 }
 
-float EditBox::getBottomPadding()
+float EditBox::getBottomPadding() const
 {
     return _bottomPadding;
 }
     
 void EditBox::setTextAlignment(TextHAlignment hAlign, TextVAlignment vAlign)
 {
-    _textHorizontalAlignMent = hAlign;
-    _textVerticalAlignMent = vAlign;
+    _textHorizontalAlignment = hAlign;
+    _textVerticalAlignment = vAlign;
     
     if (_editBoxImpl != nullptr)
     {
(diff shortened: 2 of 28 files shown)


cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
```diff
@@ -45,4 +44,47 @@ - (CGRect)editingRectForBounds:(CGRect)bounds
...
+    float factor = glview->getScaleX() / glview->getContentScaleFactor();
+    
+    _leftPadding = left * factor;
+    _topPadding = top * factor;
+    _rightPadding = right * factor;
+    _bottomPadding = bottom * factor;
+    
+}
+
+- (void)setTextAlignMent:(cocos2d::TextHAlignment)hAlign :(cocos2d::TextVAlignment)vAlign
```
zilongshanren: `setAlignMent` should be `setAlignment` to keep consistence with the system API `setTextAlignment`.
So I think `M` shouldn't be upper case letter.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
```diff
@@ -102,16 +102,6 @@ public void run() {
                 editBox.setTextColor(Color.WHITE);
                 editBox.setSingleLine();
                 editBox.setOpenGLViewScaleX(scaleX);
-                Resources r = mCocos2dxActivity.getResources();
```
zilongshanren, line 105: why remove this defaults? I calculate it based on the default font size and default padding.

anjandaffo, line 105: Now I am not using any default padding, since if Api is available for padding then User can set padding what he needed. So we don't need to set default padding.

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
anjandaffo: Sorry This is typo mistake I will fix it. Can I need to send a new PR for this.Please explain me because this is my first PR. Can I make change and update it in this PR?

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
```diff
@@ -27,17 +27,16 @@ of this software and associated documentation files (the "Software"), to deal
...
 
-#include "base/CCDirector.h"
-
-
 @implementation CCUISingleLineTextField
 
 - (CGRect)textRectForBounds:(CGRect)bounds
 {
-    auto glview = cocos2d::Director::getInstance()->getOpenGLView();
+    if (!_leftPadding) {
```
zilongshanren: `!_leftPadding` is wired... I think compare float point should be something like `if(fabs(xxx) < FLT_EPSLON)`.

zilongshanren:
@Mazyod  Could you please review this PR? 
Especially for the iOS and Mac part.
Thanks.

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
zilongshanren: You can make your changes and commit, then push to the remote branch.
It's no need to open another PR.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
zilongshanren, line 105: @anjandaffo 
Yeah, but does it break some back compatibility?

anjandaffo, line 105: @zilongshanren  If I revert it, then It's not match with position of cocos label. Because I also removed the default padding from cocos label. So you think I revert it? please suggest what I do? I revert or keep it.

Mazyod:
Sure @zilongshanren

cocos/ui/UIEditBox/UIEditBox.h
```diff
@@ -407,7 +407,20 @@ namespace ui {
...
+         * @param padding left, top, right, bottom.
+         */
+        void setPadding(float left, float top, float right, float bottom);
+        
+        /**
+         * Set the input text horizontal alignment and vertical alignment.
+         * @param hAlign the alignment for horizontally.
+         * @param vAlign the alignment for vertically.
+         */
+        void setTextAlignment(TextHAlignment hAlign, TextVAlignment vAlign);
```
Mazyod, line 449: @zilongshanren @anjandaffo Don't you guys think separating the alignment into two functions is more consistent with cocos APIs? It is troublesome if the developer wants to only change one alignment with this approach.

cocos/ui/UIEditBox/iOS/CCUIEditBoxIOS.h
```diff
@@ -59,4 +59,7 @@
 - (void)openKeyboard;
 - (void)closeKeyboard;
 
+- (void)setInnerPadding:(float)left :(float)top :(float)right :(float)bottom;
+- (void)setTextAlignMent:(cocos2d::TextHAlignment)hAlign :(cocos2d::TextVAlignment)vAlign;
```
Mazyod: **[Minor]:** it would be appreciated if you add names to each parameter, since it follows apple's code guidelines:

``` objc
- (void)setInnerPaddingLeft:(float)left top:(float)top right:(float)right bottom:(float)bottom;
- (void)setTextHorizontalAlignment:(cocos2d::TextHAlignment)hAlign verticalAlignment:(cocos2d::TextVAlignment)vAlign;
```

cocos/ui/UIEditBox/UIEditBox.h
anjandaffo, line 449: @Mazyod  @zilongshanren  I also think to separate the API, But there are one issue in Android. There are no any API is there for alignment. You set the gravity to achieve the alignment feature. There are not separate method to set gravity horizontal or vertical. Either you set single or with together. If you set only Horizontal or vertical Gravity at a time then you missed the previous set gravity.

Mazyod, line 449: @anjandaffo I see .. There is no way to query for the existing value? As I see, you've added setters, but no getters. I don't know if this is a priority or not, though.

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
```diff
@@ -45,4 +41,47 @@ - (CGRect)editingRectForBounds:(CGRect)bounds
...
+    _rightPadding = right * factor;
+    _bottomPadding = bottom * factor;
+    
+}
+
+- (void)setTextAlignment:(cocos2d::TextHAlignment)hAlign :(cocos2d::TextVAlignment)vAlign
+{
+    switch (hAlign) {
+        case cocos2d::TextHAlignment::LEFT :
+            [self setTextAlignment:NSTextAlignmentLeft];
```
Mazyod, line 60: `NSTextAlignment` APIs are only available in iOS 6+, but cocos2d-x says it supports iOS 5+.
Personally, I think iOS 5 is just too old, and should be dropped, but that's not my call.

Reference:
https://developer.apple.com/library/prerelease/ios/documentation/UIKit/Reference/NSString_UIKit_Additions/index.html#//apple_ref/c/tdef/NSTextAlignment

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
```diff
@@ -27,17 +27,13 @@ of this software and associated documentation files (the "Software"), to deal
...
-
 @implementation CCUISingleLineTextField
 
 - (CGRect)textRectForBounds:(CGRect)bounds
 {
-    auto glview = cocos2d::Director::getInstance()->getOpenGLView();
     
-    float padding = CC_EDIT_BOX_PADDING * glview->getScaleX() / glview->getContentScaleFactor();
-    return CGRectInset(bounds, padding, padding);
+    return CGRectMake(bounds.origin.x + _leftPadding, bounds.origin.y + _topPadding,
```
Mazyod, line 35: Great work! I really like the fact that this function became much simpler :+1:

cocos/ui/UIEditBox/UIEditBox.h
zilongshanren, line 449: @anjandaffo 
I'm with @Mazyod , it's better to have getters along with setters.
And for the textAlignment API, the label class give us a example.

```
    /** Sets the Label's text horizontal alignment.*/
    void setAlignment(TextHAlignment hAlignment) { setAlignment(hAlignment,_vAlignment);}

    /** Returns the Label's text horizontal alignment.*/
    TextHAlignment getTextAlignment() const { return _hAlignment;}

    /** Sets the Label's text alignment.*/
    void setAlignment(TextHAlignment hAlignment,TextVAlignment vAlignment);

    /** Sets the Label's text horizontal alignment.*/
    void setHorizontalAlignment(TextHAlignment hAlignment) { setAlignment(hAlignment,_vAlignment); }

    /** Returns the Label's text horizontal alignment.*/
    TextHAlignment getHorizontalAlignment() const { return _hAlignment; }

    /** Sets the Label's text vertical alignment.*/
    void setVerticalAlignment(TextVAlignment vAlignment) { setAlignment(_hAlignment,vAlignment); }

    /** Returns the Label's text vertical alignment.*/
    TextVAlignment getVerticalAlignment() const { return _vAlignment; }

```

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
zilongshanren, line 105: @anjandaffo  
If the new default you changed is better than the old, I think it's OK.
The old way use some magic number and I really hate magic number. 
But if the new way break some back compatibility, especially for the visually looks, it will make our developers unhappy.

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
anjandaffo, line 60: @Mazyod  In cocos2d-x 3.9 I see deployment target is 6.0.

https://github.com/cocos2d/cocos2d-x/commit/4633a9363168bd4647932263b85ee799fad7f083

cocos/ui/UIEditBox/UIEditBox.h
anjandaffo, line 449: @Mazyod  @zilongshanren  I will separate the API and add the getter for textAlignment. Is need to add getter for padding also?

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
Mazyod, line 60: Ah, sweet! Should've updated the readme as well, I guess.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
anjandaffo, line 105: @zilongshanren  Then Can I call the padding method when editbox is created with default value 5. So it's support backward compatibility.

cocos/ui/UIEditBox/iOS/CCUIMultilineTextField.mm
```diff
@@ -130,6 +127,42 @@ - (void)drawRect:(CGRect)rect
...
+            break;
+        case cocos2d::TextHAlignment::RIGHT :
+            [self setTextAlignment:NSTextAlignmentRight];
+            [_placeHolderLabel setTextAlignment:NSTextAlignmentRight];
+            break;
+        default:
+            break;
+    }
+    
+    _verticalAlign = vAlign;
```
Mazyod, line 163: @anjandaffo Shouldn't you refactor the code to update the vertical alignment by moving it outside `observeValueForKeyPath:...` into a separate method, then call it from here to make sure that it gets updated?

An easy fix would be to move the code from `observeValueForKeyPath:` into this method, then within `observeValueForKeyPath:...` call something like:

``` objc
-(void)observeValueForKeyPath:(NSString *)keyPath   ofObject:(id)object   change:(NSDictionary *)change   context:(void *)context {
    // update vertical insets
    [self setTextAlignment:_horizontalAlignment :_verticalAlignment];
}
```

A much better way is to have a separate method, call it `updateVerticalAlignment`, and it gets called from here and `observeValueForKeyPath:...`

cocos/ui/UIEditBox/UIEditBox.h
zilongshanren, line 449: @anjandaffo  I vote for adding getters for padding.

cocos/ui/UIEditBox/iOS/CCUIEditBoxIOS.mm
```diff
@@ -304,6 +303,26 @@ - (void)closeKeyboard
     [self.textInput removeFromSuperview];
 }
 
+- (void)setInnerPadding:(float)left :(float)top :(float)right :(float)bottom
+{
+    if ([self.textInput isKindOfClass:[UITextField class]]) {
```
Mazyod, line 308: @anjandaffo Can you please refactor the `setInnerPadding:` and `setTextAlignment` methods into the `CCUITextInput` protocol? This will get rid of all these if statements, and consolidate the code better.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
zilongshanren, line 105: yeah, it's better to do this. And I think we shouldn't change the anchor point of the label and the placeholder label.

cocos/ui/UIEditBox/iOS/CCUISingleLineTextField.mm
zilongshanren, line 60: @anjandaffo @Mazyod  It's time for me to change the README now. :smile:

cocos/ui/UIEditBox/iOS/CCUIEditBoxIOS.mm
```diff
@@ -109,9 +109,8 @@ - (void)setTextInput:(UIView<UITextInput,CCUITextInput> *)textInput
 - (void)createSingleLineTextField
 {
     CCUISingleLineTextField *textField = [[[CCUISingleLineTextField alloc] initWithFrame:self.frameRect] autorelease];
-    textField.contentVerticalAlignment = UIControlContentVerticalAlignmentCenter;
```
Mazyod, line 112: @anjandaffo Is there a reason to remove this default value? it goes with the discussion about defaults on android.

cocos/ui/UIEditBox/UIEditBox.h
anjandaffo, line 449: @zilongshanren  We need to add 4 method for getting the padding. I am going to add 4 separate method to get the padding i.e getleftPadding, getTopPadding, getRightPadding, getBottomPadding.

Mazyod:
overall, I think it looks pretty solid. I didn't run the code, just doing code review, of course.

anjandaffo:
@zilongshanren  @Mazyod  I made the modification and added getter for padding and text alignment. New separate Api for horizontal and vertical alignment. Please review it and revert your suggestion.

cocos/ui/UIEditBox/UIEditBox.cpp
```diff
@@ -353,6 +367,72 @@ void EditBox::setReturnType(EditBox::KeyboardReturnType returnType)
...
+    _rightPadding = right;
+    _bottomPadding = bottom;
+    
+    if (_editBoxImpl != nullptr )
+    {
+        _editBoxImpl->setPadding(left, top, right, bottom);
+    }
+}
+
+float EditBox::getLeftPadding()
```
zilongshanren: Let's add `const` at the end of these getters. :smile:

cocos/ui/UIEditBox/UIEditBox.cpp
```diff
@@ -123,6 +129,10 @@ bool EditBox::initWithSizeAndBackgroundSprite(const cocos2d::Size &size, cocos2d
         
         this->addTouchEventListener(CC_CALLBACK_2(EditBox::touchDownAction, this));
         
+        this->setPadding(_leftPadding, _topPadding, _rightPadding, _bottomPadding);
+        
+        this->setTextAlignment(_textHorizontalAlignMent, _textVerticalAlignMent);
```
zilongshanren: `_textHorizontalAlignMent` should be `_textHorizontalAlignment` and `_textVerticalAlignMent` should be `_textVerticalAlignment`

cocos/ui/UIEditBox/UIEditBox.h
```diff
@@ -407,7 +407,68 @@ namespace ui {
          * @param returnType One of the EditBox::KeyboardReturnType constants.
          */
         void setReturnType(EditBox::KeyboardReturnType returnType);
-            
+        
+        /**
+         * Set the inner padding in edit box.
+         * @param padding left, top, right, bottom.
```
zilongshanren: The doc string here is not confirm to the rule of Doxygen.
It should be 

```
@param left   The left padding.
@param top The top padding.
@param right The right padding.
@param bottom The bottom padding.
```

Only with this format we could generate the correct developer reference.

cocos/ui/UIEditBox/UIEditBox.cpp
zilongshanren: `_textHorizontalAlignMent` should be `_textHorizontalAlignment` and `_textVerticalAlignMent` should be `_textVerticalAlignment`

cocos/ui/UIEditBox/UIEditBoxImpl-common.h
```diff
@@ -133,6 +136,14 @@ class EditBoxImplCommon : public EditBoxImpl
     
     int   _maxLength;
     Size _contentSize;
+
```
zilongshanren, line 139: why duplicate these values both in `EditBox.h` and `EditBox-common.h`?
I think `EditBox-common.h` is the implementation details, it could use the data in `EditBox.h`, right?

zilongshanren:
@anjandaffo 
:+1:  Great PR. I have finished my review.

cocos/ui/UIEditBox/UIEditBoxImpl-common.cpp
```diff
@@ -95,8 +99,39 @@ void EditBoxImplCommon::initInactiveLabels(const Size& size)
 
 void EditBoxImplCommon::placeInactiveLabels()
 {
-    _label->setPosition(CC_EDIT_BOX_PADDING, _contentSize.height / 2.0f);
-    _labelPlaceHolder->setPosition(CC_EDIT_BOX_PADDING, _contentSize.height / 2.0f);
+    switch (_textHorizontalAlignMent) {
+        case cocos2d::TextHAlignment::LEFT :
```
zilongshanren, line 103: @anjandaffo 
I think here we should call `label->setTextAlignment` API, otherwise the alignment is wrong after finishing typing.
A quick way to reproduce this issue:

```
 _editName->setVerticalTextAlignment(TextVAlignment::TOP);
        _editName->setHorizontalTextAlignment(TextHAlignment::RIGHT);

```

cocos/ui/UIEditBox/UIEditBoxImpl-common.h
anjandaffo, line 139: To use EditBox.h data in EditBox-common.h we need to make it public. Can I make it public?

zilongshanren, line 139: @anjandaffo 
We could simply remove the data in `EditBox.h`, right?

anjandaffo, line 139: @zilongshanren  If I remove data from EditBox.h then How to access the private data of EditBox-common.h in EditBox.cpp.

For Ex I see
InputMode    _editBoxInputMode;
InputFlag    _editBoxInputFlag;
EditBox::KeyboardReturnType  _keyboardReturnType;
Color4B _colText;
Color4B _colPlaceHolder;

These value are also duplicate in EditBox.h and EditBox-common.h

zilongshanren, line 139: @anjandaffo 
I see, let's keep them here now.

anjandaffo:
@zilongshanren  According to your suggestion I done changes.Please review it.

cocos/ui/UIEditBox/UIEditBoxImpl-common.cpp
zilongshanren, line 103: @anjandaffo  
In the method `placeInactiveLabels`, we should also call `label->setTextAlignment` related API.

zilongshanren:
@anjandaffo 
Great :+1:   But there is still one tiny bug. See the comments in `placeInactiveLabels`.

cocos/ui/UIEditBox/UIEditBoxImpl-common.cpp
anjandaffo, line 103: @zilongshanren  In previous commit you check I called 
_label->setAlignment(_textHorizontalAlignment, _textVerticalAlignment);
    _labelPlaceHolder->setAlignment(_textHorizontalAlignment, _textVerticalAlignment);

method from placeInactiveLabels.

anjandaffo:
@zilongshanren  I called the _label->setAlignment(_textHorizontalAlignment, _textVerticalAlignment);
_labelPlaceHolder->setAlignment(_textHorizontalAlignment, _textVerticalAlignment); from placeInactiveLabels. Please verify it right?

zilongshanren:
@anjandaffo  
Oh.. my bad. It looks good to me now.
Thanks very much.

zilongshanren:
@pandamicro  @ricardoquesada  please take a look at this PR. 
Thanks.

zilongshanren:
@xiaofeng11  please help to review this PR, thanks.

xiaofeng11:
@liamcindy Please take a look of this PR.

ricardoquesada:
Hey, features that are not supported in all the platforms should be tagged as experimental.

zilongshanren:
@ricardoquesada  
How could we tag a few APIs as experimental?

ricardoquesada:
@zilongshanren you mean `experimental` ?
- If it is a new class, then the class must be under the `experimental` namespace.
- it is a new function, then the name in the function must some tell the user that is only available for certain platforms... eg: `setTexture_iOS` ... or perhaps `setTexture_experimental`...

Search for "experimental" in the cocos2d-x source code to see more examples

zilongshanren:
@ricardoquesada 
Yes, I mean `experimental`.
But if the function works for both Android and iOS, should be name it like `setTextAlignment_iOS_Android`?  We can't call it `setTextAlignment_mobile`, because it is not supported for WP8.

ricardoquesada:
yes...I think calling it something like "_experimental" or "_preview" or "_test" is better.

Who requested this feature? Because if this feature is only needed for CocosRuntime, then it is safe to add it only in Cocos Runtime.

zilongshanren:
Our developer @anjandaffo  is requested for this feature :smile: 

Hi, @anjandaffo  What do you think? Would you have any plan to support this feature across all the platforms?

anjandaffo:
@zilongshanren Sorry, I am not familiar in Windows and Linux. Even I don't have development setup for these platform so I can test it. 

Please suggest Please suggest what should I keep the method name.

If any one want to integrate these feature for remaining platform he can do it. I already passed the data in implemented class of Linux and windows.

ricardoquesada:
+1 for implementing features that are needed for our users.
-1 for implementing incomplete features. We already have too many `experimental` features. Instead of adding new unfinished features, we should finish the ones that are already unfinished.

If a users needs a certain feature, but it is not a high priority feature, or we can finish the feature (support it in all the platforms) we should not do it. In any case, we should tell our users how to solve their problem and they can fix them themselves.

zilongshanren:
@anjandaffo 
I think it's better to keep the PR not merged now.
If developers do need this feature, we could schedule it in the next release cycle and put more energy to implement the missing features on other platforms.

Thanks.

minggo:
Since it is not a complete feature, so close this PR. PR is not a good way to share codes.

waqas19921:
How to align text right through edit box in recent release?

waqas19921:
Text alignment is the basic functionality of edit text why it is not added in the project?

minggo:
@waqas19921 i think you can refer to this PR: https://github.com/cocos2d/cocos2d-x/pull/16772

minggo:
Please create an issue if there is some feature still missed.

waqas19921:
Thanks but it is only horizontal alignment please add vertical alignment also.

minggo:
Will you please create an issue for it?

PR #14485, opened 2015-11-25
Fix app-project generates .lib file in Win32

cpascal:
I'm sorry I'm not good at express in English well.

When I link my game with cocos2d-x v3.x in Visual C++, they also generate AppName.lib file.
So I analyzed the lib and obj files, it contains the following 4 lines..

/EXPORT:?clone@__Bool@cocos2d@@QBEPAVClonable@2@XZ
/EXPORT:?clone@__Float@cocos2d@@QBEPAVClonable@2@XZ
/EXPORT:?clone@__Double@cocos2d@@QBEPAVClonable@2@XZ
/EXPORT:?clone@__Integer@cocos2d@@QBEPAVClonable@2@XZ

the Demangled name is.. "cocos2d::Clonable\* cocos2d::__Bool::clone() const "
Then I looked the __Bool source code..

"cocos2d::Clonable\* cocos2d::Clonable::clone() const" is pure virtual function..
But the Clonable-derived __Bool class only defined "cocos2d::__Bool\* cocos2d::__Bool::clone() const" doesn't override the Clonable::clone() because return-value is different and virtual" keyword is not specified.
So I just added "virtual" keyword to them, the problems are solved. and then now I added "override" keywords for the coding convention.

### DIFF
--- a/cocos/deprecated/CCArray.h
+++ b/cocos/deprecated/CCArray.h
@@ -502,7 +502,7 @@ public:
      * @js NA
      * @lua NA
      */
-    virtual __Array* clone() const;
+    virtual __Array* clone() const override;
 
     // ------------------------------------------
     // Iterators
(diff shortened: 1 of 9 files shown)


minggo:
Thanks.

PR #14489, opened 2015-11-25
Fixed build for mingw and android with CMake.

adesentenac:
Added option in CMake to be able build Box2D without using it as physics library in cocos2d-x (similar to what is done in the Visual Studio solution).
Updated android.toolchain.cmake from https://github.com/taka-no-me/android-cmake/commit/556cc14296c226f753a3778d99d8b60778b7df4f and added c++_shared support.

Tested with CMake builds for MinGW and Android; and a Visual Studio build to check that no regression occurred.


minggo:
@adesentenac Android cmake is updated by @ricardoquesada and the makefiles are refactored. So could you please make a new pull request?

adesentenac:
Hi @minggo OK, I'll do that when we'll update to v3.11!

minggo:
Thanks @adesentenac .

Issue #14477, opened 2015-11-25
Keyboard fails to hide when the outside area of keyboard is touched while using EditBox on ios 7.1


wayne5ning:
### Context

cocos2d-js version: v3.9
ios verion: 7.1
xcode Version 7.1.1 (7B1005)
### Problem

The keyboard fails to hide when the outside area of keyboard was touched while using EditBox on ios 7.1. It worked in cocos2d-js version v3.8.1.
In terms of code, the `UIEditBoxImplIOS_objc::textFieldShouldEndEditing:` is not called, causing `editBoxReturn` which hides the keyboard not being called.
### How to reproduce

Run the js-test on ios 7.1 and navigate to Extensions Test->EditBoxTest, then touch an edit box, and then touch outside of the keyboard. The keyboard fails to hide.

wayne5ning:
Suggestion:
in file cocos/platform/ios/CCEAGLView-ios.mm, change selector `handleTouchesAfterKeyboardShow` to the following:

``` objective-c
-(void) handleTouchesAfterKeyboardShow
{
    NSArray *subviews = self.subviews;

    for(UIView* view in subviews)
    {
        if([view isKindOfClass:NSClassFromString(@"CCCustomUITextField")] ||
           [view isKindOfClass:NSClassFromString(@"UISingleLineTextField")] ||
           [view isKindOfClass:NSClassFromString(@"UIMultilineTextField")] ||
           //+++++ hide keyboard when touched outside of text view or hit return
           [view isKindOfClass:NSClassFromString(@"UITextView")]||
           [view isKindOfClass:NSClassFromString(@"UITextField")])
            //-----
        {
            if ([view isFirstResponder])
            {
                [view resignFirstResponder];
                return;
            }
        }
    }
}

```

zilongshanren:
@wayne5ning 
:+1:  good point. Could you send a PR?

next2016:
@wayne5ning 的方法，好像不行，我试过后，还是不能隐藏

cocos2d-js version: v3.9
ios verion: 9.2

PR #14491, opened 2015-11-26
bugfix:#14477 Keyboard fails to hide when the outside area of keyboard is touched while using EditBox on ios 7.1

wayne5ning:


zilongshanren:
@wayne5ning 
This PR is wrong...

PR #14499, opened 2015-11-26
Change ASSERT to safety mode

xiaofeng11:
Remove condition MACRO and change ASERT into safety mode. As the discuss result at http://discuss.cocos2d-x.org/t/make-assert-safe-in-release-mode/25115/6. The change of CCBundle3D has been confirmed by @super626 already.

@ricardoquesada @minggo Please review this PR.


minggo:
@xiaofeng11 
If the modification of 3d part is confirmed, then you can merge it after passing jenkins.

ricardoquesada:
looks good, but a test case should be added. A test case that proves that it doesn't crash if the condition is not met

PR #14502, opened 2015-11-27
Combine Studio change for 2d components

xiaofeng11:
Move from https://github.com/cocos2d/cocos2d-x/pull/14372


xiaofeng11:
@ricardoquesada @minggo Combine for studio changed 2d components move from https://github.com/cocos2d/cocos2d-x/pull/14372 to this PR. I have removed studio only camera define from this PR. Please review

zilongshanren:
@xiaofeng11 This PR have merge conflicts, please resolve it.

xiaofeng11:
@zilongshanren Fixed.

xiaofeng11:
[ci rebuild]

tools/bindings-generator
```diff
@@ -1 +1 @@
-Subproject commit 250673608e23c9644b96a8e9ef886ab4db96aaf7
```
pandamicro: Do not change sub modules if not necessary

xiaofeng11:
Add new api test in cpp-test project
Remove Sprite::setOffsetPosFromCenter function added by cocos studio, and reset _unflippedOffsetPositionFromCenter to Vec2::Zero when call Sprite::setTexture(const std::string &filename) function, Because when user load the sprite from a plist file, then load it from a picture file again, the _unflippedOffsetPositionFromCenter is still use plist setting, cause sprite render at wrong position.

@ricardoquesada @pandamicro Please review the change, Thanks.

plugin
```diff
@@ -1 +1 @@
-Subproject commit ad539b56c5354c1083f479d4c8d275ae8c7fdfa7
```
pandamicro: This is still unchanged

ricardoquesada:
`Sprite` and `Particles` should not include a CocosStudio file.  They are core components and they should not depend on external components.

xiaofeng11:
@ricardoquesada CocosStudio header removed from Sprite and Particles, Please review, Thanks.

cocos/2d/CCDrawNode.cpp
```diff
@@ -407,8 +417,16 @@ void DrawNode::onDrawGLLine(const Mat4 &transform, uint32_t flags)
     }
     
     glBindBuffer(GL_ARRAY_BUFFER, 0);
-    
     CC_INCREMENT_GL_DRAWN_BATCHES_AND_VERTICES(1,_bufferCountGLLine);
+
+#if CC_TARGET_PLATFORM == CC_PLATFORM_MAC || CC_TARGET_PLATFORM == CC_PLATFORM_WIN32
```
ricardoquesada: why this? If this is a cocos studio thing, then the `#ifdef` should be `#if cocosstudio`

cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -543,4 +548,177 @@ SpriteFrame* SpriteFrameCache::getSpriteFrameByName(const std::string& name)
...
+        texturePath = plist;
+
+        // remove .xxx
+        size_t startPos = texturePath.find_last_of(".");
+        texturePath = texturePath.erase(startPos);
+
+        // append .png
+        texturePath = texturePath.append(".png");
+
+        CCLOG("cocos2d: SpriteFrameCache: Trying to use file %s as texture", texturePath.c_str());
```
ricardoquesada: please remove the CCLOG debug lines.

cocos/2d/CCTMXTiledMap.cpp
```diff
@@ -29,6 +29,7 @@ THE SOFTWARE.
 #include "2d/CCTMXLayer.h"
 #include "2d/CCSprite.h"
 #include "deprecated/CCString.h" // For StringUtils::format
+#include "2d/CocosStudioExtension.h"
```
ricardoquesada: please remove this dependency as well

cocos/2d/CocosStudioExtension.cpp
```diff
@@ -0,0 +1,13 @@
+#include "CocosStudioExtension.h"
```
ricardoquesada: why this file is created under the `2d` folder?  CocosStudio has its own folder.

plugin
ricardoquesada, line 1: same comment as @pandamicro , why the plugin submodule was updated?

cocos/2d/CCDrawNode.cpp
xiaofeng11: GL_MULTISAMPLE and GL_LINE_SMOOTH only works on desktop system, GLES won't support these features, so it been limited to Windows and MacOS

plugin
xiaofeng11, line 1: last plugin cocos2dx version is still 3.7, this commit is change version number to 3.9 in plugin.

xiaofeng11:
@ricardoquesada Update code as comment, please review. Thanks.

cocos/2d/CCDrawNode.cpp
```diff
@@ -398,6 +398,21 @@ void DrawNode::onDrawGLLine(const Mat4 &transform, uint32_t flags)
         // texcood
         glVertexAttribPointer(GLProgram::VERTEX_ATTRIB_TEX_COORD, 2, GL_FLOAT, GL_FALSE, sizeof(V2F_C4B_T2F), (GLvoid *)offsetof(V2F_C4B_T2F, texCoords));
     }
+
+#if CC_TARGET_PLATFORM == CC_PLATFORM_MAC || CC_TARGET_PLATFORM == CC_PLATFORM_WIN32
```
ricardoquesada: the issue with this code is that we are adding functionality that only works on certain platforms.
for example, if I do `setLineSmooth(true)` on an iOS device, what will happen?
It won't work, but the user won't receive any kind of error.

xiaofeng11: @ricardoquesada Could you give me some suggestion on this issue? Make setLineSmooth cocos studio only, or when user call it on a mobile platform give a error message, or other? Thank you very much!

ricardoquesada: We shouldn't add features that don't work in all platforms.
My suggestion is that we temporary remove this feature from this PR until we find a better solution.

xiaofeng11:
@ricardoquesada Desktop only function removed from this PR, and we set onDraw... functions in CCDrawNode override able, Please review, Thanks.

ricardoquesada:
looks good to me.

cocos/2d/CCDrawNode.h
```diff
@@ -298,23 +300,26 @@ class CC_DLL DrawNode : public Node
     /**
      * @js NA
      */
-	void onDraw(const Mat4 &transform, uint32_t flags);
+	virtual void onDraw(const Mat4 &transform, uint32_t flags);
```
pandamicro: Indent issue, please use 4 spaces

cocos/2d/CCDrawNode.h
```diff
@@ -298,23 +300,26 @@ class CC_DLL DrawNode : public Node
...
     /**
      * @js NA
      */
-	void onDrawGLLine(const Mat4 &transform, uint32_t flags);
+	virtual void onDrawGLLine(const Mat4 &transform, uint32_t flags);
```
pandamicro: Indent issue, please use 4 spaces

xiaofeng11:
@pandamicro indent fixed.

PR #14514, opened 2015-11-28
fix Keyboard fails to hide on iOS.

zilongshanren:
ticket: https://github.com/cocos2d/cocos2d-x/issues/14477


zilongshanren:
@xiaofeng11  This PR is better than https://github.com/cocos2d/cocos2d-x/pull/14505

xiaofeng11:
Yes, this PR is much more better.

PR #14518, opened 2015-11-29
Fix memory leak when initWithImage() failed

mogemimi:
There is a memory leak in `TextureCache::addImage()` when `texture->initWithImage(image)` failed.


pandamicro:
Thanks for the fix

PR #14525, opened 2015-12-01
Add enum value for no key

elnormous:
Add enum value for "none" key just like KEY_NONE in EventKeyboard::KeyCode enum so that you don't have to cast 0 to Controller::Key.


minggo:
Thanks.

PR #14526, opened 2015-12-01
3D Normal Map Support

super626:
Normal map and test case


cocos/3d/CCMesh.h
```diff
@@ -94,10 +95,39 @@ class CC_DLL Mesh : public Ref
...
+    /**
+     * set diffuse texture
+     * @param texPath texture path
+     */
     void setTexture(const std::string& texPath);
+    /**
+     * set diffuse texture
+     * @param tex texture to be set
+     */
     void setTexture(Texture2D* tex);
```
minggo, line 107: Why not named `setDiffuseTexture` to make more clear?

minggo:
And missing js test?

cocos/3d/CCMesh.h
super626, line 107: keep the old interface.
This interface for person who does not understand material is enough.

super626:
We are not familiar with js. Any one who maintains js should add js test later.

cocos/3d/CCMesh.h
minggo, line 107: But if developers don't understand, then how do they how to use this function?

super626, line 107: Junior user uses setTexture as before.
Senior user uses setTexture(Texture\* tex, NTexture::Usage usage);

minggo, line 107: Got it.
I think it is better to add more clear comment to describe why keeping this function to set diffuse texture, and use setTexture(tex, usage) to set other type of textures.

minggo:
So who takes charge of js test?
I think we should not merge without js test.

cocos/renderer/CCGLProgram.h
```diff
@@ -110,6 +110,10 @@ class CC_DLL GLProgram : public Ref
         VERTEX_ATTRIB_BLEND_WEIGHT,
         /**Index 8 will be used as Blend index.*/
         VERTEX_ATTRIB_BLEND_INDEX,
+        /**Index 9 will be used as tangent.*/
+        VERTEX_ATTRIB_TANGENT,
+        /**Index 10 will be used as Binormal.*/
+        VERTEX_ATTRIB_BINORMAL,
```
dabingnn, line 116: Vertex attribute must be handled carefully, because in some devices, only 8 vertex attribute is supported, we may need map `VERTEX_ATTRIB_BLEND_INDEX`, `VERTEX_ATTRIB_POSITION` dynamically.

super626:
[ci rebuild]

cocos/renderer/CCGLProgram.h
super626, line 116: Sure. tangent and binormal are dynamically bind

super626:
I am adding js test with help of zhangcheng

super626:
[ci rebuild]

super626:
normal mapping for js test added

cocos/renderer/CCGLProgram.h
dabingnn, line 116: OK, It is good to hear that.

PR #14528, opened 2015-12-01
Add 3d touch support to iOS 9.0 or higher version on iPhone6s or later device

xiaofeng11:
Test case:
cpp - 64:Touches->2:3D Touch Test : see information lable
js  - Event Test -> Touch all at once : see log in console
lua - NewEventDispatcherTest -> RemoveAndRetainNodeTest : When presser more than 80% of max, will reset sprite position and log info.

@ricardoquesada @pandamicro Please review this PR.

### DIFF
--- a/cocos/base/CCTouch.h
+++ b/cocos/base/CCTouch.h
@@ -57,7 +57,9 @@ public:
      */
     Touch() 
         : _id(0),
-        _startPointCaptured(false)
+        _startPointCaptured(false),
+        _curForce(0.f),
+        _maxForce(0.f)
     {}
 
     /** Returns the current touch location in OpenGL coordinates.
@@ -101,13 +103,17 @@ public:
      * @param id A given id
      * @param x A given x coordinate.
      * @param y A given y coordinate.
+     * @param force Current force for 3d touch.
+     * @param maxForce Maxium possible force for 3d touch.
      */
-    void setTouchInfo(int id, float x, float y)
+    void setTouchInfo(int id, float x, float y, float force = 0.0f, float maxForce = 0.0f)
     {
         _id = id;
         _prevPoint = _point;
         _point.x   = x;
         _point.y   = y;
+        _curForce = force;
+        _maxForce = maxForce;
         if (!_startPointCaptured)
         {
             _startPoint = _point;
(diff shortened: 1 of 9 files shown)


cocos/platform/ios/CCEAGLView-ios.mm
```diff
@@ -420,17 +420,27 @@ - (void)touchesMoved:(NSSet *)touches withEvent:(UIEvent *)event
...
+    float fs[IOS_MAX_TOUCHES_COUNT] = {0.0f};
+    float ms[IOS_MAX_TOUCHES_COUNT] = {0.0f};
     
     int i = 0;
     for (UITouch *touch in touches) {
         ids[i] = touch;
         xs[i] = [touch locationInView: [touch view]].x * self.contentScaleFactor;;
         ys[i] = [touch locationInView: [touch view]].y * self.contentScaleFactor;;
+
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_9_0 // User 9.0 or higher SDK compiled
```
ricardoquesada: is the `#if __iPHONE...` needed ? We are doing the check in runtime, so not sure if we have to do it at compile time too.

cocos/base/CCTouch.h
```diff
@@ -101,13 +103,17 @@ class CC_DLL Touch : public Ref
      * @param id A given id
      * @param x A given x coordinate.
      * @param y A given y coordinate.
+     * @param force Current force for 3d touch.
+     * @param maxForce Maxium possible force for 3d touch.
      */
-    void setTouchInfo(int id, float x, float y)
+    void setTouchInfo(int id, float x, float y, float force = 0.0f, float maxForce = 0.0f)
```
ricardoquesada: Please, avoid using default arguments.
The "cocos2d c++ guidelines" explains why we should avoid it:
https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#default-arguments

cocos/platform/ios/CCEAGLView-ios.mm
xiaofeng11: @ricardoquesada Yes, we need it. Because 3d touch is introduced in iOS SDK version 9.0, touch.force define as "@property(nonatomic,readonly) CGFloat force NS_AVAILABLE_IOS(9_0);" in UITouch.h, same as touch.maxinumPossibleForce, if it compiled with SDK under version 9.0, will get a compile error.

xiaofeng11:
@ricardoquesada default arguments removed, please review.

linshun:
@ricardoquesada This PR should be merged in V3.10 if it pass the review. Thanks.
We would like to ship v3.10 with this feature.

ricardoquesada:
looks good, although I haven't tested it.

PR #14540, opened 2015-12-02
Improvements for JavaScript bindings

ricardoquesada:
Moves rooting to one function so it will be easier to migrate to the GC model.

### DIFF
--- a/cocos/base/CCRef.cpp
+++ b/cocos/base/CCRef.cpp
@@ -98,7 +98,7 @@ void Ref::retain()
         auto scriptMgr = ScriptEngineManager::getInstance()->getScriptEngine();
         if (scriptMgr && scriptMgr->getScriptType() == kScriptTypeJavascript)
         {
-            scriptMgr->rootRef(this);
+            scriptMgr->rootObject(this);
             _rooted = true;
         }
     }
@@ -116,7 +116,7 @@ void Ref::release()
         auto scriptMgr = ScriptEngineManager::getInstance()->getScriptEngine();
         if (scriptMgr && scriptMgr->getScriptType() == kScriptTypeJavascript)
         {
-            scriptMgr->unrootRef(this);
+            scriptMgr->unrootObject(this);
             _rooted = false;
         }
     }
--- a/cocos/base/CCRef.cpp
+++ b/cocos/base/CCRef.cpp
@@ -34,6 +34,10 @@ THE SOFTWARE.
 
 NS_CC_BEGIN
 
+// EXPERIMENTAL: Enable this in order to get rid of retain/release
+// when using the Garbage Collector
+#define CC_NATIVE_CONTROL_SCRIPT 0
+
 #if CC_REF_LEAK_DETECTION
 static void trackRef(Ref* ref);
 static void untrackRef(Ref* ref);
@@ -41,12 +45,16 @@ static void untrackRef(Ref* ref);
 
 Ref::Ref()
 : _referenceCount(1) // when the Ref is created, the reference count of it is 1
+#if CC_ENABLE_SCRIPT_BINDING
+, _luaID (0)
+, _scriptObject(nullptr)
+, _rooted(false)
+, _scriptOwned(false)
+#endif
 {
 #if CC_ENABLE_SCRIPT_BINDING
     static unsigned int uObjectCount = 0;
-    _luaID = 0;
     _ID = ++uObjectCount;
-    _scriptObject = nullptr;
 #endif
     
 #if CC_REF_LEAK_DETECTION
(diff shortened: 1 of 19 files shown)


ricardoquesada:
It is failing because I haven't updated the submodules, but if the PR is good for you ( @pandamicro  ), I'll update the submodules.

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -1960,4 +1985,67 @@ void jsb_remove_proxy(js_proxy_t* nativeProxy, js_proxy_t* jsProxy)
...
+{
+    CCLOG("jsb_ref_init: JSObject address =  %p. %s", obj->get(), debug);
+#if CC_NATIVE_CONTROL_SCRIPT
+    (void)cx;
+    (void)obj;
+    ref->_scriptOwned = true;
+#else
+    ref->autorelease();
+    JS::AddNamedObjectRoot(cx, obj, debug);
+#endif
```
pandamicro, line 2012: Is the macro control inverse ? I think we only should add root for JS object when native control script, not reversely.

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -1960,4 +1985,67 @@ void jsb_remove_proxy(js_proxy_t* nativeProxy, js_proxy_t* jsProxy)
...
+    {
+        auto ref = static_cast<cocos2d::Ref*>(jsproxy->ptr);
+        nproxy = jsb_get_native_proxy(jsproxy->ptr);
+
+        if (ref)
+        {
+            CCLOG("jsb_ref_finalize: JSObject address = %p (%s)", obj, typeid(*ref).name());
 
+            jsb_remove_proxy(nproxy, jsproxy);
+            ref->release();
```
pandamicro, line 2031: If native control script, I don't think we should release the native object, am I missing something here ?

ricardoquesada, line 2031: native control is disabled by default. The code is there so that we could keep doing our tests.

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
ricardoquesada, line 2012: ha... well, I know what you mean.
"native" for me means: the native way for JS... but I'll rename it to avoid confusion.

ricardoquesada:
@pandamicro  thanks. I fixed the macro name. New name is: `CC_ENABLE_GC_FOR_NATIVE_OBJECTS`

pandamicro:
Thanks, I will try to understand the essential idea of your PR and give my feed backs

ricardoquesada:
Ok. thanks. 
My PR is pretty simple... it just remove some Root/Unroot from the bindings, and moved it to `jsb_ref_init()` and `jsb_ref_rebind()`...and also converted some `Node::create`, to `new Node;node->init()`

cocos/base/CCRef.cpp
```diff
@@ -83,6 +91,18 @@ void Ref::retain()
...
     CCASSERT(_referenceCount > 0, "reference count should be greater than 0");
     ++_referenceCount;
+
+#if CC_ENABLE_SCRIPT_BINDING && CC_ENABLE_GC_FOR_NATIVE_OBJECTS
+    if (!_rooted && _scriptOwned)
+    {
+        auto scriptMgr = ScriptEngineManager::getInstance()->getScriptEngine();
+        if (scriptMgr && scriptMgr->getScriptType() == kScriptTypeJavascript)
+        {
+            scriptMgr->rootRef(this);
```
pandamicro: I think you mean rootObject and unrootObject below

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
pandamicro, line 2012: :joy: Ok, I see

cocos/base/CCRef.cpp
ricardoquesada: nice catch. thanks.

pandamicro:
I think we should merge: 
- js_get_type_from_native
- jsb_ref_create_jsobject
- jsb_ref_init
- jsb_ref_rebind
- jsb_ref_finalize

It can greatly simplify bindings code ! It should be shipped with v3.10.

As for `CC_ENABLE_GC_FOR_NATIVE_OBJECTS`, can you remove them from this PR ? We should do more test and try to have a stable solution for GC memory model, then we can create another PR for it.

What do you think ? @ricardoquesada

ricardoquesada:
My opinion is:
- we shouldn't remove `CC_ENABLE_GC_FOR_NATIVE_OBJECTS` from the PR
- we should ship for v3.10
- it should be disabled by default.
- QA should not test it
- Only you, me, another member of the community who are interested in this feature should test it.
- we should mention on the v3.10 press release that we have an experimental GC solution for JS (and Lua?)

ricardoquesada:
Squashed everything in one commit here:
https://github.com/cocos2d/cocos2d-x/pull/14553

ricardoquesada:
@pandamicro I merged the PR (squashed) with `CC_ENABLE_GC_FOR_NATIVE_OBJECTS` but I can remove it if you want.

I think the next steps are:
- Make sure there are not manual calls to `AddRoot` /  `RemoveRoot`... instead we should replace then with `jsb_ref_init` / `jsb_ref_finalize`
- Make sure there not `XXX::create()` calls when creating new objects... instead we should replace them with `new` and `init`

Once we have that, I think we will be ready to test GC for native objects

pandamicro:
Ok, I will check for AddRoot / RemoveRoot usage after submit v3.10 to QA.

Issue #14533, opened 2015-12-01
cc.textureCache.addImageAsync does not call callback (JSB on iOS)


pandemosth:
In 3.9 cc.textureCache.addImageAsync is not working to load an image from a URL.

See TextureCache -> Remote Texture Test in JS Tests. This is failing for me - no image loaded and no error.

PR #14542, opened 2015-12-02
Fix for Issue 14539

### LINKED ISSUE #14539: CCDownloader on iOS is broken in v3.9
The `onDataTaskSuccess` callback is never called for a data download task. You can reproduce this in the Downloader Test, Top Left test in cocos2d-x v3.9 cpp-tests project.

Inside CCDownloader-apple.mm it just looks like `onTaskFinish` callback is never called for a data task.

Related issue #14533

pandemosth:
Fixes #14539 and #14533


pandamicro:
I have tested it, indeed solved the problem, thanks !

PR #14547, opened 2015-12-02
Add missing header

colajam93:
This commit fixes the following compile error on Linux clang 3.7

```
In file included from /home/user/local/cocos2d-x/cocos/base/allocator/CCAllocatorGlobal.h:31:
/home/user/local/cocos2d-x/cocos/base/allocator/CCAllocatorStrategyDefault.h:45:16: error: use of undeuserred identifier 'malloc'
        return malloc(size);
               ^
/home/user/local/cocos2d-x/cocos/base/allocator/CCAllocatorStrategyDefault.h:51:13: error: use of undeuserred identifier 'free'
            free(address);
            ^
```

### DIFF
--- a/cocos/base/allocator/CCAllocatorStrategyDefault.h
+++ b/cocos/base/allocator/CCAllocatorStrategyDefault.h
@@ -27,6 +27,7 @@
 #define CC_ALLOCATOR_STRATEGY_DEFAULT_H
 /// @cond DO_NOT_SHOW
 
+#include <stdlib.h>
 #include "base/allocator/CCAllocatorMacros.h"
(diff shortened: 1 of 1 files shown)


colajam93:
Any news on this PR?

minggo:
Does it only happen on Linux clang? Because it has not problem in all other platforms.

colajam93:
Yes. I’m using Arch Linux. I can reproduce this issue, also under Debian 8 with Clang.

minggo:
Ok. Since it has not side effect to other platforms and i don't have these linux distribution. I will merge it.

colajam93:
Thank you to @minggo for reviewing pull requests!

PR #14551, opened 2015-12-02
changed to cocos2d-x.org server

slackmoehrle:
changes URL from `tools.itharbors.com` to `cocos2d-x.org/assets/cpp-tests-resources`


ricardoquesada:
looks good to me

PR #14566, opened 2015-12-03
fix scale9sprite be culled unexpectedly

zilongshanren:
tickets: https://github.com/cocos2d/cocos2d-x/issues/14564


zilongshanren:
@dabingnn  please review this PR, thanks.

PR #14595, opened 2015-12-05
Refactored Cocos2dxDownloader

holahmeds:
- Turned DataTaskHandler and FileTaskHandler into inner classes of Cocos2dxDownloader. These classes are only supposed to be used by Cococ2dxDownloader instances and this way they do not need to be passed a reference to said instances explicitly.
- Removed DownloadTask as it seems to be unnecessary.
- Deprecated the use of static methods to create instances and doing work on instances.
- Using SparseArray instead of HashMap for tracking tasks because they provide the same functionality, without the need to use auto-boxing for the integer keys.

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
@@ -6,6 +6,8 @@ import com.loopj.android.http.BinaryHttpResponseHandler;
 import com.loopj.android.http.FileAsyncHttpResponseHandler;
 import com.loopj.android.http.RequestHandle;
 
+import android.util.SparseArray;
+
 import org.apache.http.Header;
 import org.apache.http.message.BasicHeader;
 
@@ -171,11 +173,11 @@ public class Cocos2dxDownloader {
     private AsyncHttpClient _httpClient = new AsyncHttpClient();
     private String _tempFileNameSufix;
     private int _countOfMaxProcessingTasks;
-    private HashMap _taskMap = new HashMap();
+    private SparseArray<DownloadTask> _taskMap = new SparseArray<DownloadTask>();
     private Queue<Runnable> _taskQueue = new LinkedList<Runnable>();
 
     void onProgress(final int id, final long downloadBytes, final long downloadNow, final long downloadTotal) {
-        DownloadTask task = (DownloadTask)_taskMap.get(id);
+        DownloadTask task = _taskMap.get(id);
         if (null != task) {
             task.bytesReceived = downloadBytes;
             task.totalBytesReceived = downloadNow;
@@ -190,14 +192,14 @@ public class Cocos2dxDownloader {
     }
 
     public void onStart(int id) {
-        DownloadTask task = (DownloadTask)_taskMap.get(id);
+        DownloadTask task = _taskMap.get(id);
         if (null != task) {
             task.resetStatus();
         }
     }
 
     public void onFinish(final int id, final int errCode, final String errStr, final byte[] data) {
-        DownloadTask task = (DownloadTask)_taskMap.get(id);
+        DownloadTask task = _taskMap.get(id);
         if (null == task) return;
         _taskMap.remove(id);
         Cocos2dxHelper.runOnGLThread(new Runnable() {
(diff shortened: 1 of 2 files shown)


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
```diff
@@ -1,319 +1,300 @@
...
+		    }
+		    handle = _httpClient.get(Cocos2dxHelper.getActivity(), url, headers, null, handler);
+		    //task.handle = downloader._httpClient.get(url, task.handler);
+		}
+
+		taskHandles.put(taskID, handle);
+	    }
+	};
+	if (_taskQueue.size() < _countOfMaxProcessingTasks) {
+	    Cocos2dxHelper.getActivity().runOnUiThread(taskRunnable);
```
cezheng: Litterally this line breaks my logic that fixed the out of memory problem in https://github.com/cocos2d/cocos2d-x/pull/14530, making the _taskQueue totally useless. You have to push a null into the queue if you are to execute the task immediately otherwise the _taskQueue is forever empty because a taskRunnable is never pushed into the _taskQueue since the _taskQueue is empty initially

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
```diff
@@ -1,319 +1,300 @@
...
+
+	@Override
+	public void onStart() {
+	    Cocos2dxDownloader.this.onStart(_id);
+	}
+
+	@Override
+	public void onFinish() {
+	    // onFinish called after onSuccess/onFailure
+	    Runnable taskRunnable = Cocos2dxDownloader.this._taskQueue.poll();
```
cezheng: Also this refactor breaks the original logic. Since the each of the first N tasks should push a null into the _taskQueue, once the first task is finished, it should pop out all the first N nulls in order to start the next non-null task runnable, otherwise when the first N tasks are done, only N nulls are popped and the (N+1)th task is left in the queue untouched.

holahmeds:
I'm sorry. At the time I thought it was strange to put nulls into the taskqueue and that it was something left behind from debugging.

Also, are you sure that in the original code, after the first task finished it would remove the first N nulls. It looks to me like the queue is popped at most two times when a task finishes.

Can you tell me how you tested the out of memory errors? I could not reproduce it.

holahmeds:
Hmm, after going through the code again, I realised that this isn't thread safe.
Maybe we should add synchronized methods to add and remove tasks from the queue.

holahmeds:
@cezheng, take a look at 40b899e

cezheng:
@holahmeds 

> Can you tell me how you tested the out of memory errors? I could not reproduce it.

It was described in the PR. My employer's game has 3000+ files managed by the AssetManagerEx, and after upgrading to 3.9 the Downloader seems to start all the task all at once and the game crashed shortly after the download began. The logcat reported something about creating threads and memory. After adding a buffer queue this was fixed. Same problem in iOS and fixed with a similar approach.

> Also, are you sure that in the original code, after the first task finished it would remove the first N nulls. It looks to me like the queue is popped at most two times when a task finishes.

I'm pretty sure this won't happen as it is tested with my employer's game. The first while loop keeps removing until the queue is empty or the first element is not null, then if the the queue is not empty the pop out the first element, so only 1 non-null element is popped while multiple nulls might be removed.

cezheng:
> @cezheng, take a look at 40b899e

@holahmeds LGTM with improved readability. Thanks.

holahmeds:
> I'm pretty sure this won't happen as it is tested with my employer's game. The first while loop keeps removing until the queue is empty or the first element is not null, then if the the queue is not empty the pop out the first element, so only 1 non-null element is popped while multiple nulls might be removed.

@cezheng Which while loop? The onFinish method would only call dequeue() once and it looked like this:

``` java
    public Runnable dequeue() {
        if (!_taskQueue.isEmpty() && _taskQueue.element() == null) {
            _taskQueue.remove();
        }
        if (!_taskQueue.isEmpty()) {
            return _taskQueue.remove();
        }
        return null;
    }
```

cezheng:
@holahmeds 

``` java
    public Runnable dequeue() {
        while (!_taskQueue.isEmpty() && _taskQueue.element() == null) {
            _taskQueue.remove();
        }
        if (!_taskQueue.isEmpty()) {
            return _taskQueue.remove();
        }
        return null;
    }
```

Sorry I made a mistake in android, it should look like this, the first if shoulda been while. But luckily the if version works ok since it pops more than 1 element so the queue won't be stuck. But your version looks cleaner so we don't have to care about this anymore

PR #14598, opened 2015-12-06
Make GL context current before rendering

elnormous:
If some other process or UI element (like NSOpenPanel) changes the current OpenGL context, cocos2d-x crashes, so it is impossible to create GUI apps with cocos2d-x on Mac right now. This pull request fixes that by resetting GL context every time before rendering.

I fixed the problem with GL context switching in GLFW 3.1.2 (https://github.com/glfw/glfw/pull/487), so now it is safe to add the code to cocos2d-x.

The bug has been reported before here: http://discuss.cocos2d-x.org/t/issue-with-nsopenpanel-osx/14310


Yehsam23:
This is good for me, thanks.

zilongshanren:
@elvman 
Thanks, should we upgrade the GLFW? I think we are using the 3.0.4 version of GLFW now.

stevetranby:
Isn't GLFW at version 3.1.2?
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-src/blob/v3/contrib/src/glfw/rules.mak

minggo:
@elvman Can we update GLFW without this PR to fix the issue?

stevetranby:
I submitted a PR in the 3rd-src to update GLFW to 3.2
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-src/pull/115

minggo:
@stevetranby Thanks, but i don't know if update GLFW can fix the issue. But we need to upgrade GLFW though it fixes the issue or not.

@elnormous Is this PR needed if GLFW is upgrade to 3.2?

elnormous:
Yes, it is still needed. GLFW 3.2 fixes only the internal pointer check when setting OpenGL context.

PR #14615, opened 2015-12-09
Remove default focus item from "Node:UI->1:GUI Dynamic Create Test->4:Focus Test"

xiaofeng11:
pandamicro think we should set default status same as all other test case


xiaofeng11:
[ci rebuild]

zilongshanren:
@xiaofeng11  OK， LGTM

PR #14624, opened 2015-12-09
Allow input '\n' on desktop.

xpol:
Android and iOS are allow input '\n' so should desktop.

### DIFF
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
@@ -728,10 +728,24 @@ void GLViewImpl::onGLFWKeyCallback(GLFWwindow *window, int key, int scancode, in
         auto dispatcher = Director::getInstance()->getEventDispatcher();
         dispatcher->dispatchEvent(&event);
     }
-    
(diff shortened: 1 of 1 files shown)


zilongshanren:
@xpol 
Why add this patch, what issue does it  try to solves?

xpol:
This allow custom implemented text edit node to have multiple line text be inputted on desktop.
So they behave the same as Android and iOS.

stevetranby:
Assuming this is approved, it should only occur (or be allowed) with an editbox multi-line enabled? It probably should behave more like backspace and then internally check for multiline and insert text or if single-line then it should maybe "blur" the input (make it no longer in focus, hide keyboard on mobile).

zilongshanren:
@xpol 
I agree with @stevetranby , and I think editbox could handle enter key by itself, I'm working on implementing multiline input on Mac and Win32.

I think your patch is only for TextField, right?

xpol:
Yes, This is for Cocos internal nodes like Text Field.
on mobile platforms it is already send newline to text field.but on windows, it does not send newlines to text field.

This shoud not affect editbox, because it does not use the Cocos events.

zilongshanren:
@xpol  since this PR contains merge conflicts, I choose to close it here.

Feel free to reopen a new PR and let's continue the discussion there. thanks.

PR #14626, opened 2015-12-09
Bug fix for shouldStartLoading method, should return value from JS

dumganhar:
also fix that webview callback is already in Cocos Thread, don't need to post to cocos thread for running again.


cocos/scripting/js-bindings/manual/experimental/jsb_cocos2dx_experimental_webView_manual.cpp
```diff
@@ -20,24 +20,21 @@ static bool jsb_cocos2dx_experimental_webView_setOnShouldStartLoading(JSContext
...
+            else
+                arg[0] = JSVAL_NULL;
+            arg[1] = std_string_to_jsval(cx, url);
+            JS::RootedValue rval(cx);
+
+            bool ok = func->invoke(2, arg, &rval);
+            if (!ok && JS_IsExceptionPending(cx)) {
+                JS_ReportPendingException(cx);
+            }
+            return rval.toBoolean();
```
dumganhar, line 37: This line fixes the problem i mentioned.

PR #14641, opened 2015-12-10
fix controlButton test error

liamcindy:
@zilongshanren @xiaofeng11 Please review this PR, thanks.

### DIFF
--- a/extensions/GUI/CCControlExtension/CCControlButton.cpp
+++ b/extensions/GUI/CCControlExtension/CCControlButton.cpp
@@ -120,6 +120,14 @@ bool ControlButton::initWithLabelAndBackgroundSprite(Node* node, ui::Scale9Sprit
     }
 }
 
+ControlButton* ControlButton::create(Node* label, cocos2d::ui::Scale9Sprite* backgroundSprite)
+{
+    ControlButton *pRet = new (std::nothrow) ControlButton();
+    pRet->initWithLabelAndBackgroundSprite(label, backgroundSprite, true);
+    pRet->autorelease();
+    return pRet;
+}
+
 ControlButton* ControlButton::create(Node* label, cocos2d::ui::Scale9Sprite* backgroundSprite, bool adjustBackGroundSize)
 {
     ControlButton *pRet = new (std::nothrow) ControlButton();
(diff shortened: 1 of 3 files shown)


zilongshanren:
The effect shouldn't depend on the calling order, we should fix ControlButton instead.

liamcindy:
@zilongshanren Please review the newly changes, thanks.

tests/cpp-tests/Classes/ExtensionsTest/ControlExtensionTest/CCControlButtonTest/CCControlButtonTest.cpp
```diff
@@ -280,12 +280,12 @@ ControlButton *ControlButtonTest_Styling::standardButtonWithTitle(const char *ti
...
     auto backgroundHighlightedButton = ui::Scale9Sprite::create("extensions/buttonHighlighted.png");
     backgroundHighlightedButton->setPreferredSize(Size(45, 45));  // Set the prefered size
-    
+
     auto titleButton = Label::createWithTTF(title, "fonts/Marker Felt.ttf", 30);
 
     titleButton->setColor(Color3B(159, 168, 176));
     
-    ControlButton *button = ControlButton::create(titleButton, backgroundButton);
+    ControlButton *button = ControlButton::create(titleButton, backgroundButton,false);
```
zilongshanren: If we want to pass false here, I think the default argument should be "false".
The tests can't be changed!

zilongshanren, line 288: If we want to pass false here, I think the default argument should be "false".
The tests can't be changed!

liamcindy, line 288: if the default argument is false, then other user's logic cloud be changed, because other user had use the controlButton with "adjustBackGroundSize" as true already.

zilongshanren, line 288: @liamcindy  I mean after merging your patch, whether the users need also change their code or not? 
I hope your patch won't break the back compatibility.

liamcindy, line 288: The default logic in ControlButton is used "adjustBackGroundSize" as true before, so i set the default argument as true.

extensions/GUI/CCControlExtension/CCControlButton.cpp
```diff
@@ -79,9 +79,9 @@ bool ControlButton::initWithLabelAndBackgroundSprite(Node* node, ui::Scale9Sprit
 
         _isPushed = false;
 
-        // Adjust the background image by default
-        setAdjustBackgroundImage(true);
```
zilongshanren, line 83: why delete this line?

liamcindy, line 83: I replace "setAdjustBackgroundImage(true)"  as "setAdjustBackgroundImage(adjustBackGroundSize); " In the following  code

zilongshanren:
@liamcindy  
Please use overload instead of Default argument.

Refer to: 

https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#default-arguments

PR #14644, opened 2015-12-10
Spawn/Sequence init* were not bound

ricardoquesada:
Fixed. Now JS Scheduler Test works Ok


ricardoquesada:
take a look at this PR instead: https://github.com/cocos2d/cocos2d-x/pull/14645

PR #14652, opened 2015-12-11
more js fixes

ricardoquesada:
fixes in how ref count is measured
opengl bindings converted to new API

### DIFF
--- a/cocos/base/CCRef.cpp
+++ b/cocos/base/CCRef.cpp
@@ -50,6 +50,7 @@ Ref::Ref()
 , _scriptObject(nullptr)
 , _rooted(false)
 , _scriptOwned(false)
+,_referenceCountAtRootTime(0)
 #endif
 {
 #if CC_ENABLE_SCRIPT_BINDING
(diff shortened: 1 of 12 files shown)


cocos/base/CCRef.cpp
```diff
@@ -111,7 +116,7 @@ void Ref::release()
     --_referenceCount;
 
 #if CC_ENABLE_SCRIPT_BINDING && CC_ENABLE_GC_FOR_NATIVE_OBJECTS
-    if (_scriptOwned && _referenceCount==1 && _rooted)
+    if (_scriptOwned && _rooted && _referenceCount==_referenceCountAtRootTime)
```
pandamicro, line 119: Why is this necessary ? Sorry, I don't understand

ricardoquesada, line 119: The thing is that I don't have to unroot the obj when refCount==1.
I have to unroot it when the refCount == sameRefCount when the object was rooted.

pandamicro, line 119: But we don't know the retain release order, it may be retained by A then B then C, and released in order of B A C, I can't see why it's useful. For the rest of the PR, I have no problem

ricardoquesada, line 119: You are right... We don't know the release order, but also we don't know if unrooting when rc==1 is the best solution since it can leak... in any case we need more testing.

PR #14617, opened 2015-12-09
Fix #24345

chengstory:
TMX Iso Zorder and TMX Ortho Zorder rendering error in TestCPP.

PR #14660, opened 2015-12-14
Fix #24345

chengstory:
```
    TMX Iso Zorder and TMX Ortho Zorder rendering error in TestCPP.
```


xiaofeng11:
This pr is merge https://github.com/cocos2d/cocos2d-x/pull/14617 into v3.10 branch

PR #14673, opened 2015-12-14
fix setflip for batch sprite

super626:
Since we use _quad.bl, _quad.br, _quad.tl, _quad.tr, we should modify the texture coordinates of it according to flip flag.


super626:
@fusijie 
It is maintained by you. Can you confirm this modification?

PR #14676, opened 2015-12-14
Remove duplicate allocation to fix memory leak

mogemimi:
The `action` is created at line 954 of `CCActionTimelineCache.cpp` but `ActionTimeline::create()` has already been called at line 949.

``` cpp
949: ActionTimeline* action = ActionTimeline::create();
950: 
      ...
953:
954: action = ActionTimeline::create(); // memory leak
```

There is a memory leak here, and this patch fixes it, thanks.


zilongshanren:
@chengstory 
please review this PR, thanks. @xiaofeng11

minggo:

mogemimi:
Oh I got it. :relieved:
Thank you for helping and explaining in detail.

Issue #14677, opened 2015-12-14
Downloader on iOS 9 treats 403 Forbidden response as success


cezheng:
This bug is due to a change of behaviour of NSURLSession on iOS 9 runtime.
### How to reproduce

Create a file task with a URL pointing a non-existing resource on an existing server, and the server should return a 403 status code. 
Our server also returns a response xml string like this:

```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>3B4257EFABACF8C2</RequestId><HostId>jAsg+mkGyWMO8nOuYbNY16UHBGiHQx4UervcMecHSehgMow5lnP8IsHGxj9U3NuM6EBMf/o7VbM=</HostId></Error>
```
### Result

On iOS 8, this response will pass a `You do not have permission to access the requested resource` error, while on iOS 9 it will just save the above xml data into the specified storage path, treating the task as successful.
### Have I fixed this?

Probably yes, on my app I added a few lines of code to check the response's status code, and it seems to working properly. I will clean up the code and make a PR a few days later once the change is proven to be OK.

PR #14683, opened 2015-12-15
[Issue 14677]Fix Downloader bug on iOS 9 that treats 403 Forbidden response as success

cezheng:
On iOS 9 when server returns a response with HTTP status 403 caused the
downloader to callback successfully and saving the error message to the
storage path. This commit adds a check on the HTTP status code and
do failure callback if status code is 4xx or 5xx.

> This bug is due to the different behaviour of NSURLSession on iOS 8 and iOS 9 runtime. This commit makes them behave alike.
> 
> This is a fix for https://github.com/cocos2d/cocos2d-x/issues/14677

### DIFF
--- a/cocos/network/CCDownloader-apple.mm
+++ b/cocos/network/CCDownloader-apple.mm
@@ -515,7 +515,18 @@ namespace cocos2d { namespace network {
     NSString *destPath = [NSString stringWithUTF8String:storagePath];
     NSFileManager *fileManager = [NSFileManager defaultManager];
     NSURL *destURL = nil;
-    
+
+    // On iOS 9 a response with status code 4xx(Client Error) or 5xx(Server Error)
+    // might end up calling this delegate method, saving the error message to the storage path
+    // and treating this download task as a successful one, so we need to check the status code here
+    NSInteger statusCode = ((NSHTTPURLResponse*)downloadTask.response).statusCode;
+    if (statusCode >= 400) {
+        std::vector<unsigned char> buf; // just a placeholder
+        std::string response = [[NSString stringWithContentsOfURL:location] UTF8String] ?: "";
+        _outer->onTaskFinish(*[wrapper get], cocos2d::network::DownloadTask::ERROR_IMPL_INTERNAL, noErr, response, buf);
+        return;
+    }
+
     do
     {
         if ([destPath hasPrefix:@"file://"])


cocos/network/CCDownloader-apple.mm
```diff
@@ -515,7 +515,18 @@ - (void)URLSession:(NSURLSession *)session downloadTask :(NSURLSessionDownloadTa
     NSString *destPath = [NSString stringWithUTF8String:storagePath];
     NSFileManager *fileManager = [NSFileManager defaultManager];
     NSURL *destURL = nil;
-    
+
+    // On iOS 9 a response with status code 4xx(Client Error) or 5xx(Server Error)
+    // might end up calling this delegate method, saving the error message to the storage path
```
pandamicro, line 520: What do you mean by `might`, it's not a consistent behavior ? Maybe something else triggered the problem ?

cezheng, line 520: It is definitely caused by the different behaviour of iOS runtime on iOS 8 and iOS 9. I installed the same app on 4 different devices(iOS 8.3 & iOS 8.4.1 & iOS 9.0 & iOS 9.1), iOS 8.x all returns `You do not have permission to access the requested resource` and failed the download, while iOS 9.x all saved the error response in the storage path(as stated in https://github.com/cocos2d/cocos2d-x/issues/14677 ) @pandamicro

cezheng, line 520: I used `might` because I haven't checked whether all the 4xx & 5xx HTTP status code will end up being success, so just in case any other 4xx or 5xx response calls `URLSession:downloadTask:didFinishDownloadingToURL`, they should check the HTTP status just in case, since 4xx & 5xx shall never result in a successful callback @pandamicro

cezheng, line 520: Just to clarify, this is a consistent behaviour for HTTP 403, 404 on iOS 9

cezheng:
@pandamicro just to help you check the NSURLSession behaviour difference on iOS 8 & iOS 9 you can create a `single view application` project in Xcode and replace the `ViewController.swift` file content with the following, then check the debug console log output:

In this case, on iOS 9`"https://www.google.com/no/such/resource.zip"` request will return 404, but the `URLSession(_:downloadTask:didFinishDownloadingToURL:):` is called and `URLSession(_:task:didCompleteWithError:):`'s error is nil which will make the downloader think it is successful

```
import UIKit

class ViewController: UIViewController {

  override func viewDidLoad() {
    super.viewDidLoad()
    // Do any additional setup after loading the view, typically from a nib.
    let session = NSURLSession(configuration: NSURLSessionConfiguration.defaultSessionConfiguration(), delegate: self, delegateQueue: NSOperationQueue.mainQueue())
    let request = NSURLRequest(URL: NSURL(string: "https://www.google.com/no/such/resource.zip")!, cachePolicy: NSURLRequestCachePolicy.UseProtocolCachePolicy, timeoutInterval: 45)
    let task = session.downloadTaskWithRequest(request)
    task.resume()
  }

  override func didReceiveMemoryWarning() {
    super.didReceiveMemoryWarning()
    // Dispose of any resources that can be recreated.
  }
}

extension ViewController: NSURLSessionTaskDelegate {
  func URLSession(session: NSURLSession, task: NSURLSessionTask, didCompleteWithError error: NSError?) {
    if let response = task.response as? NSHTTPURLResponse {
      print(response.statusCode)
    }
    print("\(__FUNCTION__): \(error)");
  }
}

extension ViewController: NSURLSessionDownloadDelegate {
  func URLSession(session: NSURLSession, downloadTask: NSURLSessionDownloadTask, didFinishDownloadingToURL location: NSURL) {
    print("\(__FUNCTION__):")
  }
}
```

cocos/network/CCDownloader-apple.mm
pandamicro, line 520: Thank you for the detailed information ! Very helpful

pandamicro:
This is a critical bug fix, I will create a test case and merge it into v3.10. @cezheng Thanks

pandamicro:
Migrated to https://github.com/cocos2d/cocos2d-x/pull/14702
Closing this one

PR #14686, opened 2015-12-15
CCMotionStreak add opacity and fade switch.

joelyoung:


minggo:
@joelyoung Could you please a test case or modify current test case?

PR #14687, opened 2015-12-15
fix Shader-Sprite crash when shader can not compile

super626:
Blur shader can not compile on XiaoMi3 because of the for statement in the shader.
And it will crash.


super626:
@pandamicro Please review

PR #14690, opened 2015-12-15
Add TextureCache::addTexture

pandamicro:
Related web engine PR: https://github.com/cocos2d/cocos2d-html5/pull/3214

@ricardoquesada Adding texture directly with a key into TextureCache have been frequently requested, I also think it's very useful, so took some time to create a patch. What do you think ?

### DIFF
--- a/cocos/renderer/CCTextureCache.cpp
+++ b/cocos/renderer/CCTextureCache.cpp
@@ -422,6 +422,21 @@ Texture2D* TextureCache::addImage(Image *image, const std::string &key)
     return texture;
 }
 
+void TextureCache::addTexture(Texture2D *tex, const std::string &key)
+{
+    CCASSERT(tex != nullptr, "TextureCache: Texture MUST not be nil");
+
+    auto it = _textures.find(key);
+    if( it != _textures.end() ) {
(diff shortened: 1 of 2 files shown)


ricardoquesada:
@pandamicro 
code looks Ok for me.
but what is the use case? How does the user create the texture in the first place? thanks

pandamicro:
In JSB and web, we can create texture by cc.loader.loadImg, which will return a texture from local or remote image resources.
In C++, we can also create a texture with Image object (which can be created manually with a file).

Anyway, user can not add a texture into TextureCache is kind of wired.

pandamicro:
[ci rebuild]

ricardoquesada:
This is how I designed originally in cocos2d-iphone:
- in order to create a texture, users should use `TextureCache`... invoking `Texture2D::create()` was not recommended.

So, if someone is creating a texture manually, I guess it is because he doesn't want to "cache" it.
If he wants to add it into the cache, what he should have done, is to create it using the `TextureCache` API.
If there is no way to create a certain texture with the `TextureCache` API, what we should do, is to add an API to do that... but not add an API to add an already created texture.

PR #14692, opened 2015-12-15
Load anchor points from plist files

CodeAndWeb:
This change allows loading anchor points from sprite sheet plist files. The default behaviour is not changed. 

The anchor point is set only if an **anchor** entry is found in the plist file for a specific sprite. 

```
   <key>anchor</key>
   <string>{1,0.5}</string>
```

The new anchor point feature also works with animation frames taken from the sprite sheet.

This change is required to use TexturePacker's upcoming pivot/anchor point  editor with cocos2d-x and should also work with other sprite sheet tools (if they implement such functionality).


zilongshanren:
@ricardoquesada  
How about this PR? I have put it into v3.11.

ricardoquesada:
@zilongshanren ok. sounds good to me.

ricardoquesada:
@zilongshanren just to avoid misunderstandings... have you merged it into v3.11.... or should I do it? I can merge it in master...

zilongshanren:
@ricardoquesada  yeah, you could merge it into v3 branch and it is the candidate branch for the v3.11 release.

ricardoquesada:
ok. merged.

PR #14695, opened 2015-12-16
writing empty data should be allowed in CCFileUtils

TimothyZhang:
and the original error message "Invalid parameters" was not informative.


cocos/platform/CCFileUtils.cpp
```diff
@@ -575,7 +575,7 @@ bool FileUtils::writeDataToFile(Data retData, const std::string& fullPath)
     size_t size = 0;
     const char* mode = "wb";
 
-    CCASSERT(!fullPath.empty() && retData.getSize() != 0, "Invalid parameters.");
```
zilongshanren, line 578: maybe you should add another  assert to keep back compatibility.

`CCASSERT(retData.getSize()!=0, "reData can't be size 0")`

PR #14696, opened 2015-12-16
Improved robustness in release mode

WenhaiLin:


WenhaiLin:
@pandamicro Please review this PR.

cocos/ui/UIScale9Sprite.cpp
```diff
@@ -105,15 +113,20 @@ namespace ui {
...
-                 "SpriteFrameCache::getInstance() must be non-NULL");
-
-        SpriteFrame *frame = SpriteFrameCache::getInstance()->getSpriteFrameByName(spriteFrameName);
-        CCASSERT(frame != NULL, "CCSpriteFrame must be non-NULL");
-
-        if (NULL == frame) return false;
-        bool pReturn = this->initWithSpriteFrame(frame, capInsets);
-        return pReturn;
+        bool ret = false;
+        do {
```
zilongshanren: why use do .. while(false) here? 
I see a lot of `do .. while(true)` use case, but `do ... while(false)` is seldom used. 
What's the advantage of `do ... while(false)`?

cocos/ui/UIScale9Sprite.cpp
```diff
@@ -105,15 +113,20 @@ namespace ui {
...
+            auto spriteFrameCache = SpriteFrameCache::getInstance();
+            CCASSERT(spriteFrameCache != nullptr,
+                     "SpriteFrameCache::getInstance() must be non-NULL");
+            
+            SpriteFrame *frame = spriteFrameCache->getSpriteFrameByName(spriteFrameName);
+            CCASSERT(frame != nullptr, "CCSpriteFrame must be non-NULL");
+            if (frame == nullptr) return false;
```
zilongshanren: `if (frame == nullptr) return false;` should be `if (frame == nullptr) break`?

WenhaiLin: good catch

cocos/ui/UIScale9Sprite.cpp
```diff
@@ -84,17 +84,25 @@ namespace ui {
...
-        Sprite *sprite = Sprite::createWithSpriteFrame(spriteFrame);
-        CCASSERT(sprite != NULL, "sprite must be not nil");
-        bool pReturn = this->init(sprite,
-                                  spriteFrame->getRect(),
-                                  spriteFrame->isRotated(),
-                                  spriteFrame->getOffset(),
-                                  spriteFrame->getOriginalSize(),
-                                  capInsets);
-        return pReturn;
+        bool ret = false;
```
zilongshanren, line 87: ret should be `true` and before break, we should set `ret = false`.
otherwise, the return value would always be false.

cocos/2d/CCSprite.cpp
```diff
@@ -152,9 +154,11 @@ bool Sprite::initWithTexture(Texture2D *texture)
     CCASSERT(texture != nullptr, "Invalid texture for sprite");
 
     Rect rect = Rect::ZERO;
-    rect.size = texture->getContentSize();
+    if (texture) {
+        rect.size = texture->getContentSize();
+    }
 
-    return initWithTexture(texture, rect);
+    return initWithTexture(texture, rect, false);
```
zilongshanren, line 161: why pass extra `false` argument here?

zilongshanren:
@xiaofeng11  @pandamicro  please review this PR. 
please take care, this PR is important and we can't do it wrongly.

cocos/ui/UIScale9Sprite.cpp
zilongshanren, line 87: ignore this, I found ret is set with `this->init(...)` function.

zilongshanren: ignore this, I found ret is set with `this->init(...)` function.

WenhaiLin, line 87: I don't think so. Initially,`ret` should be `false`.  `ret` will be assigned the result of `init(sprite,xx,xx,xx,xx,xx)`

cocos/2d/CCSprite.cpp
WenhaiLin, line 161: The `rotated` default to `false`. You can refer to `virtual bool initWithTexture(Texture2D *texture, const Rect& rect);`

xiaofeng11, line 161: @zilongshanren  "initWithTexture(texture, rect);" will call "initWithTexture(texture, rect, false);", call "initWithTexture(texture, rect, false);" directly here is better for performance. I think the code is OK.

xiaofeng11:
@zilongshanren It's OK for me, but I'm not familiar with scale9sprite, @liamcindy Could you help double check the change of scale9sprite?

PR #14697, opened 2015-12-16
fix setflip for batch sprite

super626:
Since we use _quad.bl, _quad.br, _quad.tl, _quad.tr, we should modify the texture coordinates of it according to flip flag.


super626:
@fusijie 
It is maintained by you. Can you confirm this modification?

super626:
move from #14673

minggo:
Why need to modify like this?
Any problem of current implementation?

super626:
111.tmx here failed to display, http://pan.baidu.com/s/1skc4wJr

It shows as following,
5 pic

super626:
@minggo , @ricardoquesada Can you help to review this PR?

ricardoquesada:
what is the best way to reproduce the bug? could you add a test case?
thanks.

super626:
@ricardoquesada 
I added a tmx with filpped spritebatch test.

ricardoquesada:
thanks. merged

ricardoquesada:
sorry, I merged but then I revert the merge because it was merged in v3.10 and I don't control v3.10.
If you send the PR against v3, I'll merge it in v3.

PR #14698, opened 2015-12-16
fix Shader-Sprite crash when shader can not compile

super626:
Blur shader can not compile on XiaoMi3 because of the for statement in the shader.
And it will crash.

### DIFF
--- a/tests/cpp-tests/Classes/ShaderTest/ShaderTest2.cpp
+++ b/tests/cpp-tests/Classes/ShaderTest/ShaderTest2.cpp
@@ -93,11 +93,14 @@ public:
 
                 if(std::get<0>(effect) >=0)
                     break;
-                QuadCommand &q = std::get<2>(effect);
-                q.init(_globalZOrder, _texture->getName(), std::get<1>(effect)->getGLProgramState(), _blendFunc, &_quad, 1, transform, flags);
-                renderer->addCommand(&q);
+                auto glProgramState = std::get<1>(effect)->getGLProgramState();
+                if (glProgramState)
+                {
+                    QuadCommand &q = std::get<2>(effect);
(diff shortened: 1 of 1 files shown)


super626:
Move from #14687

minggo:
So should we modify the shader instead?

super626:
XiaMi3 does not support "for" statements and "for" could not be unrolled. In this test case, we adjust the blur radius.

minggo:
So i think we can do it in two ways:
1. Delete this test case, because many developers will refer to test cases to write their codes, and they don't know that the test case has some limitation.
2. Create a documents that describe the issues that we found during development

I prefer method 2, because all developers will benefit from it.

zhaijialong:
那个for的终止条件是个uniform。。这应该es2.0不允许的，改成常量/宏吧，话说cocos还缺少管理shader里宏的机制。。。

super626:
@ricardoquesada Any idea about this?

ricardoquesada:
@super626  your PR looks good to me. No need to remove the test case. 

@minggo what part of this PR you don't like? thanks.

minggo:
This PR just avoid issues on some devices, and we will meet this problem again if we don't have a method to handle it. That's why i ask for some mechanism to handle it above.

ricardoquesada:
yes, but your proposals were:
- remove the sample: doesn't solve the issue.
- or document it: doesn't solve the issue... cocos2d-x will keep crashing if we don't apply @super626 patch.

What I mean... we can still add more documentation, but is there something that you don't like about the patch?

I mean, the correct solution is to detect the device, and use a shader that doesn't use `for` when running on Xioami3... that that will take a lot of time, but it is the correct solutiuon.

minggo:
This PR looks good to me.
I just mean we need to add a documentation about it, it will help us and all developers, and yep, it is another issue. I just mention it here.

ricardoquesada:
Ok. thanks. So I'm merging this PR.

PR #14701, opened 2015-12-16
Fixed label's letter can't be hide by `setVisible(false)`.

WenhaiLin:


WenhaiLin:

cocos/2d/CCLabel.cpp
```diff
@@ -130,13 +131,18 @@ class LabelLetter : public Sprite
             return;
         }
 
-        Color4B color4(_displayedColor.r, _displayedColor.g, _displayedColor.b, _displayedOpacity);
+        auto displayedOpacity = _displayedOpacity;
+        if(!_letterVisible)
```
zilongshanren, line 135: why not use the visible property in sprite?

WenhaiLin, line 135: I want LabelLetter not be rendered directly, it is desired to maintain the visible property in sprite equal to `false`.

PR #14702, opened 2015-12-16
Fix Downloader callback bug on iOS9

pandamicro:
Fixed https://github.com/cocos2d/cocos2d-x/issues/14677
Migrated from https://github.com/cocos2d/cocos2d-x/pull/14683 
Thanks to the contribution of @cezheng

### DIFF
--- a/cocos/network/CCDownloader-apple.mm
+++ b/cocos/network/CCDownloader-apple.mm
@@ -389,7 +389,8 @@ namespace cocos2d { namespace network {
     
     if(_outer)
     {
-        if(error) {
+        if(error)
+        {
             std::vector<unsigned char> buf; // just a placeholder
             _outer->onTaskFinish(*[wrapper get],
                              cocos2d::network::DownloadTask::ERROR_IMPL_INTERNAL,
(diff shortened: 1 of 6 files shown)


cezheng:
@pandamicro I just realized that this fix is not comprehensive. In iOS 8, after calling `onTaskFinish` in `URLSession(_:downloadTask:didFinishDownloadingToURL:):`, `URLSession(_:task:didCompleteWithError:):` will still be called with an error `You do not have permission to access the requested resource` so in this case the `onTaskFinish` will be called twice.

Is it ok to go with 2 callbacks on 1 task, or shall we add a flag to prevent the second one?

pandamicro:
I tested on iOS 8, only `URLSession(_:task:didCompleteWithError:):` will be called when error happened, so `onTaskFinish` shouldn't be called twice, have you produced the buggy scenario ?

cezheng:
@pandamicro  After looking through the code 1 more time this just came into my mind and I haven't tested it yet. 

Since you have tested it, that should be OK. Just ignore it.

pandamicro:
So the conclusion is when 4xx / 5xx error happened during file downloading:

iOS 9 : will only call `URLSession(_:downloadTask:didFinishDownloadingToURL:):`
iOS < 9 : will only call `URLSession(_:task:didCompleteWithError:):`

cezheng:
iOS 9 : will call both `URLSession(_:downloadTask:didFinishDownloadingToURL:):` and `URLSession(_:task:didCompleteWithError:):` , but `URLSession(_:task:didCompleteWithError:):`'s error is nil.

pandamicro:
It's more clear now, so I improved the solution.

@cezheng @yangws Please take a look

cocos/network/CCDownloader-apple.mm
```diff
@@ -396,24 +398,47 @@ - (void)URLSession:(NSURLSession *)session task :(NSURLSessionTask *)task
...
+            const char *orignalURL = [task.originalRequest.URL.absoluteString cStringUsingEncoding:NSUTF8StringEncoding];
+            std::string errorMessage = "";
+            
+            // Check for error status code
+            if (statusCode >= 400)
+            {
+                errorMessage = cocos2d::StringUtils::format("Downloader: Failed to download %s with status code (%d)", orignalURL, (int)statusCode);
+            }
+            // Unknown issue
+            else
```
cezheng: This is not necessary because if `statusCode < 400` it should not be treated as failure

cocos/network/CCDownloader-apple.mm
```diff
@@ -396,24 +398,47 @@ - (void)URLSession:(NSURLSession *)session task :(NSURLSessionTask *)task
...
+            {
+                errorMessage = cocos2d::StringUtils::format("Downloader: Failed to download %s with unknown error", orignalURL);
+            }
+            
+            _outer->onTaskFinish(*[wrapper get],
```
cezheng: This should only be called when `statusCode >= 400`

cocos/network/CCDownloader-apple.mm
pandamicro: It's in `URLSession(_:task:didCompleteWithError:):`, so I assume there must be something wrong, we just don't know what.

cezheng: This method name `didCompleteWithError` is kinda confusing, but all NSURLSessionTasks calls this method when finished. Success or failure is determined by whether error is nil or not

pandamicro: :joy: 

Ok... Thanks for the clarification

cezheng:
@pandamicro Thanks. And regarding my previous pull request in https://github.com/cocos2d/cocos2d-x/pull/14530 , my manager says it is causing the main thread UI to stuck for a bit when too many tasks are created by the AssetManagerEx, I'm thinking about delaying the NSURLSessionTask allocations. One more PR when I'm done

PR #14710, opened 2015-12-17
no more uthash

ricardoquesada:
Instead of using two `uthash` entires, it uses two `unordered_map<>` entries.
Benefits are:
- easier to debug. I can inspect the contents of the `unorderd_map` with the
  debugger
- less memory. I don't need to allocate two `proxies`. I only allocate one `proxy` and I added it into both `maps`

If also fixes an important bug when unrooting objects. Unroot was
cleaning the heap object. So now proxy adds a `_jsobj`, which is a raw
pointer. Needed under certain cases.

Adds a new `jsb_remove_proxy()` function that only takes one `proxy` as argument. Old one is still supported.

### DIFF
--- a/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
+++ b/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
@@ -93,9 +93,12 @@ static uint32_t s_nestedLoopLevel = 0;
 // server entry point for the bg thread
 static void serverEntryPoint(unsigned int port);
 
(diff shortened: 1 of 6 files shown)


pandamicro:
:+1:  +10086

ricardoquesada:
@pandamicro 
Not sure why the build failed... actually "jenkings" is green, but "default" (what is "default"?) is red.
Any idea why?

If you like it, feel free to merge it. thanks.

PR #14716, opened 2015-12-17
fix font for EditBox in iOS & Mac

newnon:
iOS & Mac load font only by name

### DIFF
--- a/cocos/ui/UIEditBox/UIEditBoxImpl-ios.mm
+++ b/cocos/ui/UIEditBox/UIEditBoxImpl-ios.mm
@@ -207,6 +207,8 @@ UIFont* EditBoxImplIOS::constructFont(const char *fontName, int fontSize)
     float retinaFactor = eaglview.contentScaleFactor;
     NSString * fntName = [NSString stringWithUTF8String:fontName];
     
+    fntName = [[fntName lastPathComponent] stringByDeletingPathExtension];
+    
     auto glview = cocos2d::Director::getInstance()->getOpenGLView();
     float scaleFactor = glview->getScaleX();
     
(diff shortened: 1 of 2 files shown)


zilongshanren:
@xiaofeng11  please review

xiaofeng11:
@newnon @zilongshanren I'm not sure what type of font we would load here, if the code want to load custom TTF font, we need do more work to make it work proper. Please refer the follow links:
http://stackoverflow.com/questions/15447558/can-not-include-ttf-font-into-project
http://stackoverflow.com/questions/16788330/how-do-i-get-the-font-name-from-an-otf-or-ttf-file
http://www.waterworld.com.hk/en/node/157

And if we only want to load system font, under Mac & iOS, we need give font family name, this may differ from file name. Second link at above demo how to get the family name.

WenhaiLin:
@newnon This PR is unnecessary, because we already handled that in `platform/ios/CCDevice-ios.mm:_initWithString`/`platform/mac/CCDevice-mac.mm::_initWithString`.

@xiaofeng11 Game developers should use font with font family name for EditBox on iOS/Mac, either system font or custom font.

newnon:
@xiaofeng11  @WenhaiLin iOS & Mac use not font family it use font full name and it is good practice to use the same font name & file name. This pull request help to avoid unnecessary #ifdefs in code
user place font in his Resources, write it in info.plist and use everywhere the same name.
And yes you can find similar code in Label implementation.

zilongshanren:
@WenhaiLin 
Thoughts?

WenhaiLin:
@zilongshanren I think this PR is OK for merge.

PR #14719, opened 2015-12-17
fix include in cocos network module

newnon:


zilongshanren:
@newnon 
:+1:  Thanks.

PR #14720, opened 2015-12-17
getDevideUID() in Device class

newnon:
getDevideUID() for some platforms

### DIFF
--- a/cocos/platform/win32/CCDevice-win32.cpp
+++ b/cocos/platform/win32/CCDevice-win32.cpp
@@ -501,6 +501,12 @@ void Device::vibrate(float duration)
     CC_UNUSED_PARAM(duration);
 }
 
+std::string Device::getDevideUID()
+{
+	std::string ret;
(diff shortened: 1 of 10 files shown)


zilongshanren:
@newnon  
Thanks, could you send this PR to v3 branch?
We have almost finished our tests, this feature should be in the next release cycle.

cocos/platform/win32/CCDevice-win32.cpp
```diff
@@ -501,6 +501,12 @@ void Device::vibrate(float duration)
     CC_UNUSED_PARAM(duration);
 }
 
+std::string Device::getDevideUID()
+{
+	std::string ret;
```
zilongshanren, line 506: Maybe we could return  the mac address or somethings else on the PC?

zilongshanren: Maybe we could return  the mac address or somethings else on the PC?

PR #14721, opened 2015-12-17
slider misprint fix

newnon:


zilongshanren:
@xiaofeng11  @liamcindy 
This PR seems good to me. But is may break compatibility.
What do you guys think about it?

xiaofeng11:
@zilongshanren I think it won't break compatibility of SliderReader and CocosStudio. @liamcindy How about your opinion?

zilongshanren:
@xiaofeng11  
Thanks.

@liamcindy  
If it's OK to you, please ping me and I will merge this PR.

PR #14722, opened 2015-12-17
WebP loading improvements WebP loaded as premultiplied alpha if it has

newnon:
This PR solve problem with blending function and also WebP need less memory if it has no alpha


cocos/platform/CCImage.cpp
```diff
@@ -2121,19 +2121,19 @@ bool Image::initWithWebpData(const unsigned char * data, ssize_t dataLen)
         if (WebPGetFeatures(static_cast<const uint8_t*>(data), dataLen, &config.input) != VP8_STATUS_OK) break;
         if (config.input.width == 0 || config.input.height == 0) break;
         
-        config.output.colorspace = MODE_RGBA;
-        _renderFormat = Texture2D::PixelFormat::RGBA8888;
+        config.output.colorspace = config.input.has_alpha?MODE_rgbA:MODE_RGB;
```
zilongshanren, line 2124: `MODE_rgbA` should be `MODE_RGBA`?

zilongshanren:
@newnon 
I remembered the WebP format doesn't have premultiplied  alpha,  where can I find the info about WebP premultiplied  alpha?

Thanks.

newnon:
MODE_rgbA mean that WebP library give you image as premultiplied alpha
it automatically convert any colorspace to premultiplied alpha RGBA

``` cpp
// Colorspaces
// Note: the naming describes the byte-ordering of packed samples in memory.
// For instance, MODE_BGRA relates to samples ordered as B,G,R,A,B,G,R,A,...
// Non-capital names (e.g.:MODE_Argb) relates to pre-multiplied RGB channels.
// RGB-565 and RGBA-4444 are also endian-agnostic and byte-oriented.
typedef enum { MODE_RGB = 0, MODE_RGBA = 1,
               MODE_BGR = 2, MODE_BGRA = 3,
               MODE_ARGB = 4, MODE_RGBA_4444 = 5,
               MODE_RGB_565 = 6,
               // RGB-premultiplied transparent modes (alpha value is preserved)
               MODE_rgbA = 7,
               MODE_bgrA = 8,
               MODE_Argb = 9,
               MODE_rgbA_4444 = 10,
               // YUV modes must come after RGB ones.
               MODE_YUV = 11, MODE_YUVA = 12,  // yuv 4:2:0
               MODE_LAST = 13
             } WEBP_CSP_MODE;
```

zilongshanren:
@newnon
Could you add a test case?

newnon:
Done you can see that WebP with alpha now is premultiplied
And WebP without alpha now correctly loads as RGB888

newnon:
And also i recommend to update WebP library, new version is much faster

cocos/platform/CCImage.cpp
```diff
@@ -2121,19 +2121,19 @@ bool Image::initWithWebpData(const unsigned char * data, ssize_t dataLen)
...
+        _renderFormat = config.input.has_alpha?Texture2D::PixelFormat::RGBA8888:Texture2D::PixelFormat::RGB888;
         _width    = config.input.width;
         _height   = config.input.height;
         
-        //webp doesn't have premultipliedAlpha
-        _hasPremultipliedAlpha = false;
+        //we ask webp to give data with premultiplied alpha
+        _hasPremultipliedAlpha = config.input.has_alpha;
```
zilongshanren, line 2130: Could you adjust the statement `_hasPremultipliedAlpha = config.input.has_alpha;` to the top and use `_hasPremultipliedAlpha` instead of `config.input.has_alpha`?

In this way, the code looks more concise.

Thanks.

newnon, line 2130: But it give some misunderstanding of our principles - we do all the stuff not because we have premultiplied alpha, but because we have alpha. 
If you wish a can add bool hasAlpha on top of function. but i think that compiler is smart enough to optimize this code.

zilongshanren, line 2130: @newnon 
Thanks.

PR #14723, opened 2015-12-17
fix web socket crash

newnon:
crash in destructor if websocket was not initialized


zilongshanren:
@newnon 
:+1:  thanks

PR #14725, opened 2015-12-17
getCachePath() implementation for most platforms

newnon:
getCachePath() for most platform except linux(i just don't now what does it mean on linux it return writable path)
for example if you download something from internet to documents dir(writable path) you app can be rejected and also changed writable path for mac for more suitable(its bad practice to write something in root of user documents folder)


zilongshanren:
@newnon 
Thanks, this feature can't be merged into v3.10. Could you send it to v3 branch?

PR #14739, opened 2015-12-19
3 10 stand alone test

ricardoquesada:
- adds a stand along test to test GC. "js-tests" is so big that it is impossible to test it.
- fixes memory leak when "subclassing" (extending) native objects

Status of the bindings:
- native objects and "extended" objects work as expected. They are freed when the GC collects them
- every one and then there are a few crashes when using the new memory model. I guess it is related to events... more debugging is needed
- GC doesn't collect the objects immediately... not sure why... I think @pandamicro knows why


zilongshanren:
@ricardoquesada  
Is this PR ready for merge?

ricardoquesada:
@zilongshanren it was already merged. closing it.

PR #14740, opened 2015-12-19
Bug Fix & h5 engine ref update

xiaofeng11:
Fix pageViewTest Horizontal scroll won't work in Lua-test
PageView default direction is horizontal, we won't set it again after create it. And after refact, now ccui.PageViewDirection won't exist any more, all scroll direction should use ccui.ScrollViewDir define.

synchronize h5 engine parser match modifications and update h5 engine ref

@pandamicro Please review this PR


pandamicro:
Pure script changes, no need to build in jenkins

PR #14746, opened 2015-12-19
fix lua focus test

zilongshanren:


xiaofeng11:
I have test this modification in lua-test, it solved after focus test all button will turn red when been clicked issue. It can be merged.

PR #14748, opened 2015-12-19
Temporary disable forceGC on win32

pandamicro:


cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -925,8 +925,10 @@ bool ScriptingCore::executeScript(JSContext *cx, uint32_t argc, jsval *vp)
 
 bool ScriptingCore::forceGC(JSContext *cx, uint32_t argc, jsval *vp)
 {
+#if CC_TARGET_PLATFORM != CC_PLATFORM_WIN32
```
dumganhar, line 928: `Temporary disable forceGC on win32`?
If so, why not add some comments for this?

pandamicro, line 928: I'm gonna fix it in this version

Envoyé de mon iPhone

PR #14755, opened 2015-12-20
Fix OpenGL macro redefinition warning in JS bindings/Mac

mogemimi:
When I build `libjscocos2d Mac` on Xcode 7, I get the following warning:

```
cocos/platform/mac/CCGL-mac.h:44:9: 'glReleaseShaderCompiler' macro redefined
cocos/scripting/js-bindings/manual/jsb_opengl_manual.h:42:13: Previous definition is here
```

I think JS bindings can use `#include "platform/CCGL.h"` instead of the macro redefinition in `jsb_opengl_manual.h`. This PR improves and fixes the warning, thanks.


zilongshanren:
@mogemimi 
:+1:

PR #14764, opened 2015-12-21
Update V3 document branch with V3.10

xiaofeng11:
@pandamicro @zilongshanren Please review this PR.


cocos/2d/CCLabel.h
```diff
@@ -108,6 +108,23 @@ class EventListenerCustom;
 class CC_DLL Label : public Node, public LabelProtocol, public BlendProtocol
 {
 public:
+    enum class Overflow
```
zilongshanren, line 111: It seems you don't translate the new APIs in label.

xiaofeng11, line 111: I'll notice the owner of every module to update document after it merged

PR #14777, opened 2015-12-22
prevent recompile when run Android app on Android Studio to device or Emulator.

raksa:
to prevent recompile when run Android app on Android Studio to device or Emulator.


zilongshanren:
@WenhaiLin  please review

WenhaiLin:
@zilongshanren This PR is OK for me. @raksa Thank you!

PR #14779, opened 2015-12-22
Bug fix & update

xiaofeng11:
Roll back disable force GC change on windows platform, this cause lua binding can not respond mouse click

Remove 3d touch demo iOS sdk checking, it SDK defines won't work in cpp compile


xiaofeng11:
@pandamicro @zilongshanren Please review this pr

PR #14781, opened 2015-12-22
Rollback slider and loadingbar in non scale9 mode

zilongshanren:


zilongshanren:
@liamcindy 
please review, thanks.

PR #14788, opened 2015-12-23
Fix method name spelling in AutoPolygon

mogemimi:
I found a small mistake in the `AutoPolygon` class and renamed one:
- `getTriaglesCount` => `getTrianglesCount`

Also I've marked the old functions as deprecated, so there's no breaking changes in the API.


zilongshanren:
@mogemimi 
:+1:  thanks.

PR #14793, opened 2015-12-24
fix setflip for batch sprite

super626:


super626:
Discussion is here, #14697

PR #14802, opened 2015-12-25
fix layout backimage size error with newest scale9sprite

liamcindy:
@zilongshanren This PR just compatible with old size logic, Please review this PR, thanks.


zilongshanren:
@liamcindy 
Looks good to me

PR #14803, opened 2015-12-25
Fix NavMeshDebugDraw bug on iphone5s

super626:


cocos/navmesh/CCNavMeshDebugDraw.cpp
```diff
@@ -148,6 +148,9 @@ void NavMeshDebugDraw::drawImplement(const cocos2d::Mat4& transform, uint32_t fl
         _dirtyBuffer = false;
     }
     for (auto &iter : _primitiveList){
+        if (iter->type == GL_POINTS)
+            continue;
+        
         _stateBlock->setDepthWrite(iter->depthMask);
         if (iter->type == GL_POINTS){
```
pandamicro: This should be removed

super626:
@pandamicro Done.

PR #14806, opened 2015-12-26
Allow same components to be added multiple times

TheCodez:
With this change if a component with the same name is added multiple times e.g. PhysicsBody, it won't assert, instead it will replace the old component with the new.

Different approach to https://github.com/cocos2d/cocos2d-x/pull/14371


minggo:
@TheCodez I think Component is designed in this way though i don't know why. As 14371 is merged, can i close this PR?

PR #14810, opened 2015-12-29
fix compile error in visual studio 2015

GuoLunHao:
There's no include file "afxres.h" in visual studio 2015 Enterprise version, so that change it to "winresrc.h"


GuoLunHao:
@xiaofeng11 @pandamicro

xiaofeng11:
Seems afxres.h only contain in vs2013 and lower version, winres.h and winresrc.h contain in all version, but winres.h is MFC based include file, cocos2dx is not dependecy MFC lib, so we use winresrc.h to make it compatible with all version of visual studio.

xiaofeng11:
[ci rebuild]

xiaofeng11:
[ci rebuild]

PR #14814, opened 2015-12-29
Fix js project with pre-build frameworks crash when exit program

xiaofeng11:
@pandamicro Please review it


zilongshanren:
[ci rebuild]

PR #14843, opened 2016-01-06
fix iOS downloader crash when storage path contains spaces

### LINKED ISSUE #14842: [BUG]iOS downloader crashes when storagePath contains spaces
since downloader is using [NSURL urlWithString:] to convert absolute path NSString to NSURL, and when the NSString path contains characters like ' ' which needs to be escaped, the method will return nil thus crashing the downloader.

An example path containing space is iOS's "Application Support" directory.

cezheng:
This PR fixes https://github.com/cocos2d/cocos2d-x/issues/14842
which causes iOS downloader to crash when storagePath contains spaces (e.g. the app's Application Support directory)

This problem can either be fixed by changing

```
[NSURL URLWithString:[prefix stringByAppendingString: destPath]];
```

to 

```
[NSURL URLWithString:[[prefix stringByAppendingString: destPath] stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding]];
```

or

```
destURL = [NSURL fileURLWithPath:destPath];
```

which is more concise.

### DIFF
--- a/cocos/network/CCDownloader-apple.mm
+++ b/cocos/network/CCDownloader-apple.mm
@@ -553,9 +553,7 @@ namespace cocos2d { namespace network {
         
         if ('/' == [destPath characterAtIndex:0])
         {
-            // absolute path, need add prefix
-            NSString *prefix = @"file://";
-            destURL = [NSURL URLWithString:[prefix stringByAppendingString: destPath]];
+            destURL = [NSURL fileURLWithPath:destPath];
             break;
         }
         


cezheng:
Seems the jenkins is not working. I only modified 1 source file on iOS platform but mac and windows build failed.

cezheng:
@pandamicro Would you review & merge this PR please? It is pretty short. And our team still has a lot of other fixes relating to the iOS & Android network code that starts from 3.9.

minggo:
Thanks.

cezheng:
@minggo Thanks. When do you plan to release 3.11? Our team still have a dozen of bug fixes related to the downloader & AssetManagerEx, that are yet put in a PR. Not sure if we can make it in time for 3.11

minggo:
@cezheng Please send to v3.11 if it is a critical bugs. We will release v3.11 via that branch. We may release v3.11 about in mid-May.

cezheng:
@minggo 
Thanks. I checked the v3.11 branch and realized that this fix is not yet in the v3.11. I guess I will put all fixes including this in the upcoming PR.

minggo:
Thanks @cezheng

Issue #14378, opened 2015-11-13
Scale9Sprite opacity cant change when I setState gray


semgilo:
cocos2dx 3.6
for example:
local img = ccui.ImageView:create()  
img:loadTexture("test.png")
img:setState(1)
img:setOpacity(50)

the img opacity is 255 
50 expected

zilongshanren:
@semgilo 
Does this issue still exists in the latest v3.9 version?

semgilo:
@zilongshanren i dont try,i  had used shader instead

semgilo:
@zilongshanren i try3.9 the issue exist

stevetranby:
SHADER_NAME_POSITION_GRAYSCALE
Update `ccPositionTexture_GrayScale_frag` to multiply v_fragmentColor.a possibly not multiplying by 0.5 and setting CCScale9Sprite's opacity by default to 128 (0.5) instead.

```
    gl_FragColor.xyz = vec3(0.2126*c.r + 0.7152*c.g + 0.0722*c.b);
    gl_FragColor.a = v_fragmentColor.a * 0.5; 
```

zilongshanren:
@stevetranby  @semgilo

Fixed at #14854

semgilo:
@zilongshanren 3q

semgilo:
@stevetranby 3q

PR #14852, opened 2016-01-07
fix scale9sprite gray state opacity issue.

zilongshanren:
Ticket: https://github.com/cocos2d/cocos2d-x/issues/14378


liamcindy:
@xiaofeng11 @zilongshanren It looks good to be merged.

PR #14854, opened 2016-01-07
fix scale9sprite gray state opacity issue.

zilongshanren:
Ticket: https://github.com/cocos2d/cocos2d-x/issues/14378


stevetranby:
Thanks! 

Might need to make a note in RELEASE_NOTES of the change since it technically breaks an old behavior?

zilongshanren:
@stevetranby 
yeah, I will make a note in RELEASE_NOTES after this PR is merged.

PR #14860, opened 2016-01-08
Update CCLabel.cpp

animaseed:
if not return, "batchNode->getTextureAtlas()->drawQuads();" will run another time, it's a waste

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -1405,7 +1405,7 @@ void Label::onDraw(const Mat4& transform, bool transformUpdated)
             //draw text without outline
             glprogram->setUniformLocationWith4f(_uniformEffectColor,
                 _effectColorF.r, _effectColorF.g, _effectColorF.b, 0.f);
-            break;
+            return;
         case LabelEffect::GLOW:
             glprogram->setUniformLocationWith4f(_uniformEffectColor,
                 _effectColorF.r, _effectColorF.g, _effectColorF.b, _effectColorF.a);


cocos/2d/CCLabel.cpp
```diff
@@ -1405,7 +1405,7 @@ void Label::onDraw(const Mat4& transform, bool transformUpdated)
             //draw text without outline
             glprogram->setUniformLocationWith4f(_uniformEffectColor,
                 _effectColorF.r, _effectColorF.g, _effectColorF.b, 0.f);
-            break;
```
zilongshanren, line 1408: @animaseed  if you return here, the text without outline won't be drawn... @WenhaiLin Thoughts?

animaseed, line 1408: @zilongshanren @WenhaiLin 

``` cpp
 glprogram->setUniformLocationWith4f(_uniformTextColor,
     _textColorF.r, _textColorF.g, _textColorF.b, _textColorF.a);

 glprogram->setUniformLocationWith4f(_uniformEffectColor,
     _effectColorF.r, _effectColorF.g, _effectColorF.b, _effectColorF.a);

 for (auto&& batchNode : _batchNodes)
 {
     batchNode->getTextureAtlas()->drawQuads();
 }

//for the text with outline, the code above will draw the text part and the outline part

case LabelEffect::NORMAL:
    glprogram->setUniformLocationWith4f(_uniformTextColor,
        _textColorF.r, _textColorF.g, _textColorF.b, _textColorF.a);
    break;

//the text without outline will enter case LabelEffect::NORMAL, not case LabelEffect::OUTLINE, and will draw the text part using the code below

for (auto&& batchNode : _batchNodes)
{
    batchNode->getTextureAtlas()->drawQuads();
}

//so add return, both work OK, I have test it
```

zilongshanren:
@WenhaiLin thoughts?

PR #14880, opened 2016-01-12
Fix typos in comments and documentation

mogemimi:
This commit fixes some typos and doxygen annotations (`@parame` => `@param`).


zilongshanren:
@mogemimi  good catch :+1:

PR #14883, opened 2016-01-13
Analyzing directory exists function for mac & ios

ookcode:
The 'isDirectoryExist(const std::string& dirPath)' function has some problems that can not return the correct result in iOS and mac when analyzing the bundle directory.
Also,it may not work in win32 and android for the same reason.
I fixed it for iOS and mac, continue seeking solution for android and win32.

### DIFF
--- a/cocos/platform/apple/CCFileUtils-apple.h
+++ b/cocos/platform/apple/CCFileUtils-apple.h
@@ -57,6 +57,7 @@ public:
     void setBundle(NSBundle* bundle);
 private:
     virtual bool isFileExistInternal(const std::string& filePath) const override;
+    virtual bool isDirectoryExistInternal(const std::string& dirPath) const override;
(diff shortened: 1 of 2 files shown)


WenhaiLin:

minggo:
@ookcode How can i reproduce the issue of current implementation?

luyaor:
Dear Vincent Yao(ookcode),
We are working on identifying redundant development and duplicated pull requests. We have found there is a pull request: https://github.com/cocos2d/cocos2d-x/pull/13687 which might be similar or duplicate to this one. We would really appreciate if you could help us to validate and give us feedback.
Thank you very much for your time!

drelaptop:
@FancyCoder0 

did you find __redundant development and duplicated pull requests__ by some automated algorithm? that's cool and it will save much time.

PR #14889, opened 2016-01-14
Some Important Fixes for File Reading, may cause stack overflow

dumganhar:
- The fix for BMFontConfiguration::parseConfigFile reads string file.
  
  'strchr' finds a char until it gets a '\0' or the char to find, if 'contents' self doesn't end with '\0',
  'strchr' will search '\n' out of 'contents' 's buffer size, it will trigger **potential and random crashes** since lineLength may bigger than 512 and 'memcpy(line, contents + parseCount, lineLength);' will cause stack buffer overflow.
- const Data -> const Data& for function argument, performance improvement.
- Removes unused code in Material::initWithFile.
- Refactor CCBundle3D, remove wrong use of getDataFromFile. _jsonBuffer is std::string, _binaryBuffer is Data instance now.
- SpriteTest fix
  - Image is Ref class, please new & release it. Don't allocate it on stack.
  - Fixes complicated logic of getting string from file.

### DIFF
--- a/cocos/2d/CCFontFNT.cpp
+++ b/cocos/2d/CCFontFNT.cpp
@@ -279,21 +279,37 @@ std::set<unsigned int>* BMFontConfiguration::parseConfigFile(const std::string& 
     CCASSERT((!data.isNull()), "BMFontConfiguration::parseConfigFile | Open file error.");
 
     if (memcmp("BMF", data.getBytes(), 3) == 0) {
+        // Handle fnt file of binary format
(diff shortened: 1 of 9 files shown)


pandamicro:
@ricardoquesada This is an important bug fix for bug found in Runtime, can you review it ?

ricardoquesada:
@pandamicro @dumganhar 

looks good to me.
merged.

dumganhar:
@ricardoquesada Thanks.

PR #14890, opened 2016-01-15
can use new firefox version to debug cocos2d-x jsb projects

minggo:
Use Firefox to remote debug cocos2d-x jsb projects is invalid since version 39. This pull request is to fix the problem.


pandamicro:
:+1:  Tested with Firefox v36 and Firefox v43, works well.

PR #14891, opened 2016-01-15
Fix warning: extra tokens at end of #ifdef directive

mogemimi:
I get the following warning when building libcocos2d-iOS (and tvOS) on Xcode 7.2:

```
cocos/platform/ios/CCEAGLView-ios.mm:430:21: Extra tokens at end of #ifdef directive
```

This patch fixes it, thanks!


zilongshanren:
@mogemimi  
keep fighting with evil warnings :+1:

PR #14046, opened 2015-09-29
Homogenous shadows and strokes for system fonts.

j-jorge:
Please review and merge this commit which provides the following features.

The stroke and the shadow of a cocos2d::Label rendered using the system fonts
are now handled on cocos' side.

For Linux and Mac targets this commit enables the stroke and shadow features.

For Linux targets this commits fixes the color used to render system fonts.

For Mac targets this commit also fixes the font creation. Previously only the
font family was searched. Now we try to get the font with the provided name.

For iOS and Android targets this commit enables the stroke and the shadow on
the emojis.

Textures created with system fonts are now cached and reused if a new label
is requested with the same text and same font definition.

PR #14899, opened 2016-01-16
Homogenous shadows and strokes for system fonts.

j-jorge:
_Note: this branch has been previously submitted for the pull request #14046. Due to a renaming of the original branch for clarity, the previous request has been closed. This commit includes the merge fixes requested in the previous pull request. Sorry for the inconvenience_

Please review and merge this commit which provides the following features.

The stroke and the shadow of a cocos2d::Label rendered using the system fonts
are now handled on cocos' side.

For Linux and Mac targets this commit enables the stroke and shadow features.

For Linux targets this commits fixes the color used to render system fonts.

For Mac targets this commit also fixes the font creation. Previously only the
font family was searched. Now we try to get the font with the provided name.

For iOS and Android targets this commit enables the stroke and the shadow on
the emojis.

Textures created with system fonts are now cached and reused if a new label
is requested with the same text and same font definition.

### DIFF
--- a/build/cocos2d_libs.xcodeproj/project.pbxproj
+++ b/build/cocos2d_libs.xcodeproj/project.pbxproj
@@ -7,6 +7,12 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0AE22ECF1C4AE8D80099B83F /* CCStroke.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 0AE22ECD1C4AE8D80099B83F /* CCStroke.cpp */; };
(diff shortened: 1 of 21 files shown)


j-jorge:
@pandamicro @liamcindy @WenhaiLin you were referenced in the previous pull request. Any chance for this branch to be merged in v3? (supposedly when the conflicts will be resolved).

Fixing the conflicts as upstream changes is quite a workload so I would rather know if it has an interest or not :)

j-jorge:
@pandamicro, @liamcindy, @WenhaiLin can you please have a look at this pull request? Here is a screenshot from cpp-test on OSX :

OSX Example

The result is the same on iOS and Android. On Linux the system fonts do not contain the emojis but the border is the same.

I can't stress this enough: as far as I know there is no way to achieve a similar result in the current version of Cocos2d-x.

cc @minggo, @zilongshanren

minggo:
@zilongshanren please take a look.

PR #14900, opened 2016-01-17
fix enable box2d compile error

svzdev:


zilongshanren:
@mailgyc 
Hi, I have enabled box2d in cpp-test mac， but I can compile it successfully

svzdev:
please review the change:

in the CCColliderDetector.h file 
    std::vectorcocos2d::Vec2 _calculatedVertexList;

but in the CCColliderDetector.cpp file
    _calculatedVertexList = Array::create();

It's clear an error.

zilongshanren:
@xiaofeng11  please take a look at this issue.

xiaofeng11:
@zilongshanren Yes, the code in CCColliderDetector.cpp is a bug.

@mailgyc Why do you change CCColliderDetector.h? Chipmunk and box2d won't work together, so I think the old code won't have problem here.

svzdev:
i don't really understand the relation between  ColliderDetector and box2d,  I enable both of them :-(

xiaofeng11:
@mailgyc ColliderDetector use physical engine to detect collision.  There's two physical engine embed in cocos2d-x, chipmunk and box2d, user can only enable one of them, but not both.  Your change in  CCColliderDetector.h make both of them can be enable, this may cause error.  Please rollback .h file, the changes in cpp file already solved the bug.

svzdev:
@xiaofeng11   I revert the .h file.
can you please explain the code for me again, I can't understand the design.

In class ColliderBody

`#if ENABLE_PHYSICS_BOX2D_DETECT
    b2Fixture *_fixture;
    ColliderFilter *_filter;
#elif ENABLE_PHYSICS_CHIPMUNK_DETECT
    cpShape *_shape;
    ColliderFilter *_filter;
#elif ENABLE_PHYSICS_SAVE_CALCULATED_VERTEX
    std::vector<cocos2d::Vec2> _calculatedVertexList;
#endi`

As I think, ENABLE_PHYSICS_SAVE_CALCULATED_VERTEX is just a calculate result cache, it must enable with box2d or chipmunk, what's the if-else meaning ?

xiaofeng11:
@mailgyc ColliderFilter use box2d or chipmunk to detect collision and the calculate result only been saved when  ENABLE_PHYSICS_SAVE_CALCULATED_VERTEX set. ENABLE_PHYSICS_SAVE_CALCULATED_VERTEX is a separate setting other than box2d or chipmunk.
And yes there is a bug in code you post here. ENABLE_PHYSICS_SAVE_CALCULATED_VERTEX should be a separate #ifdef check but not a #elif.
Now CocosStudio won't use CCColliderDetector  any more, and the original designer has left maintain team, so the code has a long time hasn't update. Could you help us to improve it? Thanks a lot.

minggo:

Issue #14461, opened 2015-11-23
A lot of hardcoded constants in ScrollView


sbrednikhin:
There a lot of hardcoded constants inside of ScrollView source code.

```
static const int NUMBER_OF_GATHERED_TOUCHES_FOR_MOVE_SPEED = 5;
static const float OUT_OF_BOUNDARY_BREAKING_FACTOR = 0.05f;
static const float BOUNCE_BACK_DURATION = 1.0f;

#define MOVE_INCH            7.0f/160.0f
```

```
Vec2 ScrollView::calculateTouchMoveVelocity() const
{
    float totalTime = 0;
    for(auto &timeDelta : _touchMoveTimeDeltas)
    {
        totalTime += timeDelta;
    }
    if(totalTime == 0 || totalTime >= 0.1f)
    {
        return Vec2::ZERO;
    }
```

If you don't want to give a chance to change it, at least you could give a way to use them from the outside, by moving it's declarations to header.

zilongshanren:
@neokim 
What do you think about these hardcoded values?

sbrednikhin:
Possibility of changing this values through class interface would be ideal. As an example: In PageView scrolling parameters can be changed from outside of class. 
If you don't want to allow parameter changing - it would be useful to declare them as static constants in public area of the class, so user could use this values to configure his code.

My widgets have custom touch behavior and this behavior is critical for functionality, but breaks scrolling. So now I have 2 variants:
1. Make some workaround to fix scrolling. In this case I need to use this hardcoded values in my code, but I cant - values are declared in source file.
2. Change hardcoded values in cocos sources, and get portion of headache with merging after new version of cocos will be released.

neokim:
Those constants are internal properties for elastic scroll pattern. Any changes on them will be spoiling scrolling pattern. They are not supposed to be exposed to outside and to be modified without careful concern about user experiences for scroll UI.

If your changes on touch behavior breaks scrolling and so you need to refer to these values, it looks to me that there may be some design flaw in your implementation. Touch and scrolling pattern should be considered as independent parts. Even though scrolling is triggered and driven by touch event, it doesn't mean that its scroll pattern is dependent to touch feature.

If you have custom touch implementation, it can be separated from scrolling pattern in some ways.

sbrednikhin:
[...]

PR #14907, opened 2016-01-18
Add a setter for touch total time threshold in ScrollView

neokim:
Fix for #14461


zilongshanren:
@neokim  
Thanks.  [ci rebuild]

PR #14914, opened 2016-01-19
Add ability to get the font family

steerlink:
The ability to know the font family is helpful for implementing custom rendering.

Additional fix for compiler warning where the return value from calling `first->find(fontName)` is unsigned and thus always >= 0. The proper comparison is whether `find()` returns `std::string::npos`.


zilongshanren:
@steerlink 
thanks.

Issue #14737, opened 2015-12-18
Assert attempting to play background music for 2nd time (win8.1-universal project)


StreakyGames:
Using Cocos2d-x v3.9

Run my game...music plays smile
(quit game)
Run my game again...assert!!!!!!!!!!!!

The game will NOT play music until I rename the audio file or rename the folder it's in (eg. Resources/Audio/music.wav to Resources/WTF/music.wav), obviously I rename the filename/path within the code to reflect this change.

After renaming, it'll run once, then everytime after....assert!!!!

This is the assert that triggers:
long FileUtils::getFileSize(const std::string &filepath)
{
    CCASSERT(false, "getFileSize should be override by platform FileUtils");
    return 0;
}

Apparently occurs on an actual Windows Phone too, and the mp3 from cpp-tests works fine.

stammen:
Fixed in pull request https://github.com/cocos2d/cocos2d-x/pull/14848

minggo:
@stammen i found you closed #14848. So what's the status of this issue?

stammen:
@minggo It was fixed in pull request https://github.com/cocos2d/cocos2d-x/pull/14848

minggo:
Oh, it is fixed in #14915

PR #14915, opened 2016-01-19
v3 added missing getFileSize() method for winrt

stammen:
This pull request adds the missing getFileSize() method to winrt file implementation. This fixes issue #14737.


zilongshanren:
@stammen  thanks。

PR #14927, opened 2016-01-20
Fix warning and a bug in FontAtlasCache::unloadFontAtlasTTF()

mogemimi:
I get the following warning when compiling on Xcode 7:

```
CCFontAtlasCache.cpp:269:44: Comparison of unsigned expression >= 0 is always true
```

The warning is caused by the line `CCFontAtlasCache.cpp:269`, and there is a bug.

``` cpp
if (item->first.find(fontFileName) >= 0) // Here is a bug
```

So I think the following code is better:

``` cpp
if (item->first.find(fontFileName) != std::string::npos)
```

This patch fixes it.


zilongshanren:
@mogemimi 
:+1:  thanks.

PR #14939, opened 2016-01-21
v3.10 added missing getFileSize() method for winrt 

stammen:
This pull request adds the missing getFileSize() method to winrt file implementation. This fixes issue #14737.


zilongshanren:
@stammen 
Thanks, but I think 3.10 branch is already merged into v3 branch and 3.10 will be release soon.

So this PR should be in v3 now.

PR #14940, opened 2016-01-21
Add ability to stop auto scrolling in ui::ScrollView and extension::ScrollView

ShaneSmit:
The new methods are:

``` c++
ui::ScrollView::stopAutoScroll();
extension::ScrollView::stopAnimatedContentOffset();
```

These methods effectively do the same thing, but are named in correlation to the methods that initiated the auto scroll.

Also made all the `ui::ScrollView::scrollTo*` methods `virtual`, which mirrors the functionality already present in the `jumpTo*` methods.

@neokim @steerlink @joshbodily


zilongshanren:
@ShaneSmit 
Thanks, this PR looks good to me.

PR #14941, opened 2016-01-21
Mark equals() functions as const

mogemimi:
I think several `equals()` functions are effectively const methods, but they aren’t marked `const`.
This PR adds `const` keyword to them.


zilongshanren:
@mogemimi  
:+1:

mogemimi:
@zilongshanren Thank you for always reviewing so quickly!!! :blush:

PR #14944, opened 2016-01-21
Export WICImageLoader form DLL

joshbodily:
We have a remote image loader that base64 encodes WICPixelFormat buffers and sends them to our backend.  We found it useful to DLL export WICImageLoader so that we didn't have to use a different library and/or roll our own.

@stammen


stammen:
This PR is ready to merge.

minggo:
Thanks @joshbodily .
@stammen Thanks for reviewing.

PR #14949, opened 2016-01-22
V3 chipmunk 7.0

newnon:
It is the first version. At this moment it more for discuss.
You also need to build chipmunk 7.0 for your platform (i've already tested it on Mac & iOS only)
Whats done:
- build project for iOS & Mac 
- all test passed
- new debug draw system

Whats not done
- projects for other platforms(may be for some platforms needed some include path changes)
- binding system i am not experience enough to make bindings.

chipmunk v7.0 prebuild libraries  for Mac iOS Android & Win32 https://www.dropbox.com/s/h7e8otq6d6qnq6w/chipmunk.zip?dl=0

### DIFF
--- a/build/cocos2d_libs.xcodeproj/project.pbxproj
+++ b/build/cocos2d_libs.xcodeproj/project.pbxproj
@@ -16658,7 +16658,7 @@
 				);
 				MACH_O_TYPE = staticlib;
 				OTHER_LDFLAGS = "-llibsql3";
-				USER_HEADER_SEARCH_PATHS = "$(inherited) $(SRCROOT)/../external/png/include/mac $(SRCROOT)/../external/tiff/include/mac $(SRCROOT)/../external/webp/include/mac $(SRCROOT)/../external/jpeg/include/mac $(SRCROOT)/../external/freetype2/include/mac $(SRCROOT)/../external/glfw3/include/mac $(SRCROOT)/../external/websockets/include/mac $(SRCROOT)/../external/chipmunk/include/chipmunk $(SRCROOT)/../external/curl/include/mac $(SRCROOT)/../external/freetype2/include/mac/freetype2";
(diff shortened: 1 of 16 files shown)


zilongshanren:
@newnon 
thanks :+1:

TheCodez:
Can you add support for [cpHastySpace](https://github.com/slembcke/Chipmunk2D/blob/master/src/cpHastySpace.c) to support multithreaded solver on platforms that support pthread?

newnon:
@TheCodez may be it good idea i can try to include it in next commit

TheCodez:
You have to replace all cpSpace calls on supported platforms with cpHastySpace e.g. cpHastySpaceStep, cpHastySpaceNew, to support the new multihreading feature in chipmunk.

ricardoquesada:
nice feature!

ricardoquesada:
[ci-rebuild]

ricardoquesada:
@newnon great work. I've just tried it and works Ok with your prebuilt libs (v7.0)
however if I compiled my own libs with v7.0.1, it crashes on iOS, but works Ok on Mac.
Have you tried with chipmunk v7.0.1 ? Did it work Ok ? thanks.

**update**: I tried using your modified `chipmunk_types.h` file and recompile chipmunk again... but it still fails on iOS. Tried compilling chipmunk in debug mode. Failed.
Tried v7.0... failed.
My guess is that there is a bug in Xcode 7.2.1... @newnon do you remember which Xcode version did you use ? Thanks.

**update 2**: Everything worked fine with Xcode 6.4... I don't know if it was a bug in 7.2 or if the `-fembed-bitcode` was causing it... continue the tests tomorrow

ricardoquesada:
@TheCodez How do you use the `cpHastySpace` functions? Are they just replacements of `cpSpace`?

TheCodez:
Yes,  instead of cpSpace use cpHastySpace and then replace the cpSpaceXXX methods by cpHastySpaceXXX. 
See the header for reference: https://github.com/slembcke/Chipmunk2D/blob/master/include/chipmunk/cpHastySpace.h

ricardoquesada:
@TheCodez ok. thanks for the info.

newnon:
I will try with new xcode

ricardoquesada:
I sent a new PR that supersedes this PR: https://github.com/cocos2d/cocos2d-x/pull/15108

PR #14953, opened 2016-01-23
Fixed flipped polygon sprite

amakaseev:
There was an error: if you change the polygon information that previously defined flipX(or Y) not taken into sprite. Changes polygon information derived from the change of the frame in the animation.


TheCodez:
[ci rebuild]

wefiends:
Looks like it needs a call to `updateColor()` as well to set the vertex colors.

PR #14971, opened 2016-01-25
Replace char* with std::string in HttpRequest

mogemimi:
This replaces the usage of `const char*` with `const std::string&` in `HttpRequest::setUrl()` and `setTag()` arguments, for the following reasons:
- Both `_url` and `_tag` members are `std::string` type.
- To avoid calling `std::string::c_str()` when passing arguments to them

Also, there are no backwards breaking changes to APIs. Thanks!


zilongshanren:
@mogemimi 
thanks

PR #14977, opened 2016-01-27
Allow inherit form FileUtilsLinux, FileUtilsWin32 and FileUtilsWinRT.

xpol:
The doc for FileUtils::setDelegate() says:

```
 * You can inherit from platform dependent implementation of FileUtils, such as FileUtilsAndroid,
 * and use this function to set delegate, then FileUtils will invoke delegate's implementation.
 * For example, your resources are encrypted, so you need to decrypt it after reading data from
 * resources, then you can implement all getXXX functions, and engine will invoke your own getXX
 * functions when reading data of resources.
```

But the constructor of  FileUtilsLinux, FileUtilsWin32 and FileUtilsWinRT are private.

So make these constructors protected which allows inherit from them.

---

BTW:

Why inheritance?
Why not something like `FileUtils::setCrypt(const Crypt& crypt)`.
After set, `Crypt::encrypt()` are called before save data/string to file.
and `Crypt::decrypt()` are called after read data/string from file.
If not set a Crypt to file utils, just do normal reading and writing.

### DIFF
--- a/cocos/platform/apple/CCFileUtils-apple.h
+++ b/cocos/platform/apple/CCFileUtils-apple.h
@@ -54,13 +54,9 @@ public:
     virtual bool writeToFile(ValueMap& dict, const std::string& fullPath) override;
 
     virtual ValueVector getValueVectorFromFile(const std::string& filename) override;
-    void setBundle(NSBundle* bundle);
 private:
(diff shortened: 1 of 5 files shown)


xpol:
I have create new PR ( #15304 ) for this

PR #14978, opened 2016-01-27
Small refactor FileUtils.

xpol:
- Should be const reference for data/string args to be written in FileUtils methods.
-  Better name for data to be saved.
- Remove unnecessary memory copy when saving string to file.


zilongshanren:
@xpol  
:+1: 

@minggo  please take a look at this PR, thanks.

minggo:
It is great.
Thanks @xpol .

PR #14980, opened 2016-01-27
fix typo

catinred2:


zilongshanren:
@catinred2 
This PR looks good to me, @dualface  Does it break back compatibility?

PR #14988, opened 2016-01-29
add etc2 CPU and GPU decode. (PVR fileformat only)

yutaka-takeda-drecom:
add etc2 CPU and GPU decode. (PVR fileformat only)
original source code is KTX(Khronos).
optimize for KTX cpu decode.
etc2 GPU decode setting is add "-DCC_USE_OPENGLES=3" for compile option.
## ETC2 GPU decode

GPU decode of ETC2 requires OpenGLES3.

for example.

Xcode
Preprocessor Macros
“CC_USE_OPENGLES=3”

Android
Application.mk
“APP_CPPFLAGS := -DCC_USE_OPENGLES=3”


yutaka-takeda-drecom:
In order to respond to new renderer.cpp, this pull request will be close.

PR #14990, opened 2016-01-29
Uniformize image support macro usage.

j-jorge:
All images format implementation but WIC test the value of a CC_USE_<FORMAT>
preprocessor macro to verify if the implementation must be available or not.
WIC support is checked against the definition of the macro only,
independently of its value, thus making the implementation available even
if CC_USE_WIC = 0.

This commit makes the tests of the CC_USE_WIC macro identical to the tests
of the other similar image format macros.

### DIFF
--- a/cocos/platform/CCImage.cpp
+++ b/cocos/platform/CCImage.cpp
@@ -802,7 +802,7 @@ namespace
 #endif // CC_USE_JPEG
 }
 
-#ifdef CC_USE_WIC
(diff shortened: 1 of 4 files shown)


minggo:
@j-jorge Could you please fix conflict? Then i will merge this PR.

j-jorge:
@minggo done :)

minggo:
Thanks.

PR #14994, opened 2016-02-01
Added CCTouchAction which simulates touches as FiniteTimeAction

dolow:
This PR contains CCTouchAction which performs gestures such as Tap, Swipe, Pinch as FiniteTimeAction.
The main purpose of this module is to help writing UI test, and also it provides intuitive way to dispatch touch events as gesture.
The usage of this is exactly same as original CCActions.
I hope this feature suits cocos2d-x.


dolow:
[ci rebuild]

dolow:
[ci rebuild]

dolow:
"Slave went offline during the build"

dolow:
[ci rebuild]

minggo:
It is interesting. It is just used for testing?

minggo:
I am not sure if it is suitable to add testing related code to 2d folder. @ricardoquesada what's your opinion?

ricardoquesada:
If I understood correctly, this code is useful to simulate gestures, is that correct?

there are two things to discuss:

a)  should this be part of the core of cocos2d-x
b) is this PR ready to be merged

Personal thoughts

a) Although it is handy to have a library that simulates gestures, I don't think it should be part of the core of cocos2d-x.
In any case it could be as an extension, or part of a "test" framework (that we don't have).
We are in the process of trying to get cocos2d-x thinner an thinner, so we are somewhat picky at what new code to add.

b) In case we want to add it, we will need:
- test cases
- squash all commits into a single one

minggo:
@ricardoquesada I agree with you, i think it is not suitable to be part of cocos2d-x, that's why i ask your opinion.

PR #15002, opened 2016-02-02
Fixed TargetedAction::isDone always return false!

WenhaiLin:


zilongshanren:
@WenhaiLin 
Could you add a test case for this patch?

PR #15015, opened 2016-02-04
Remove unnecessary c_str() calls

mogemimi:
This commit removes some unnecessary `c_str()` calls to avoid creating temporary string objects.


zilongshanren:
:+1:  @mogemimi 
Thanks.

PR #15033, opened 2016-02-10
Fixes USE_BULLET option

NatWeiss:
The USE_BULLET option can affect both CC_USE_PHYSICS and CC_USE_3D_PHYSICS depending on whether the option is enabled or disabled, which will clobber the definitions that other options have already set such as USE_CHIPMUNK.

This fixes it to just affect CC_USE_3D_PHYSICS (which presumably was the intention all along).


minggo:
Compiler options are moved to `cmake/Modules/SetCompilerOptions.cmake`. Could you please send a new pull request for it, thanks.

PR #15039, opened 2016-02-11
Scale9Sprite: allow to set custom shaders (fix #12753)

### LINKED ISSUE #12753: ClippingNode does not support Scale9Sprite in some platform
Cocos2d-x v3.6

``` cpp
    auto win_size = Director::getInstance()->getWinSize();
    auto stencil = cocos2d::ui::Scale9Sprite::create("CloseNormal.png");
    stencil->ignoreAnchorPointForPosition(false);
    stencil->setAnchorPoint(Vec2::ANCHOR_MIDDLE);
    stencil->setScale(2.0f);
    stencil->setPosition(Vec2::ZERO);


    auto clipping = ClippingNode::create();
    clipping->setStencil(stencil);
    clipping->setInverted(true);
    clipping->setAlphaThreshold(0.5f);
    clipping->setPosition(win_size/2);
    addChild(clipping);

    auto content = Sprite::create("HelloWorld.png");
    content->setPosition(Vec2::ZERO);

    clipping->addChild(content);
```

Display the results on Linux is ok,  on android, ios is wrong:

img_0357

Shulepov:
As **State** also sets shaders, may be it would be better to add third state like _CUSTOM_ or _NONE_ that will be set when used custom shaders.

### DIFF
--- a/cocos/ui/UIScale9Sprite.cpp
+++ b/cocos/ui/UIScale9Sprite.cpp
@@ -483,7 +483,11 @@ namespace ui {
         }
 
         applyBlendFunc();
-        this->setState(_brightState);
+        if (getGLProgramState()) {
+            _scale9Image->setGLProgramState(getGLProgramState());
+        } else {
+            this->setState(_brightState);
+        }
         if(this->_isPatch9)
         {
             size.width = size.width - 2;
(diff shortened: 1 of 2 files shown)


zilongshanren:
@Shulepov 
Thanks :+1:

Issue #15005, opened 2016-02-02
node curve animation bug in Animate3D::update


yutaka-takeda-drecom:
in cocos2d-x 3.9.0

There is a bug in the case of nodeCurve animation of Aimate3D.

There is a transform of the node in the c3b. (This name is A)
and transform of animation are there in the c3b. (This name is B)
A and B are the same transform.

The animation data at zero frame contains the transform of the node.
In the original source code , transform of the node will be multiplied twice.
(node transform \* animation transform)

original source code

```
void Animate3d::update(float t)
{
    // ...
                for (const auto& it : _nodeCurves)
                {
                    // ...
                    node->setAdditionalTransform(&transform);
                }
}
```

The API was changed from Node::setAdditionalTransform to Node::setNodeToParentTransform.
This was without a node transform.

bug fix source code

```
void Animate3d::update(float t)
{
    // ...
                for (const auto& it : _nodeCurves)
                {
                    // ...
                    node->setNodeToParentTransform(transform);
                }
    // ...
}
```

As a result , it was able to display correctly in my test data.

However , a few cases in which the test.

yutaka-takeda-drecom:
update
I made a pull request (Pull #15045).

Node curve animation bugfix in cocos 3.10.

There is a bug in the case of nodeCurve animation of Aimate3D.

There is a transform of the node in the c3b. (This name is A)
and transform of animation are there in the c3b. (This name is B)
A and B are the same transform.

The animation data at zero frame contains the transform of the node.
In the original source code , transform of the node will be multiplied twice.
(node transform \* animation transform)

But, Sprite3D root node must do the calculation of the transform.
Reason is that the operation of the transform information in non-animated .
(For example, an application programmer change the position to use the API of Sprite3D :: setPosition.)

In conclusion , it is necessary to change the calculation in Sprite3D root node and Sprite3D children  node.

In the case of Sprite3D root node.

```
void Animate3d::update(float t)
{
    // ...
                for (const auto& it : _nodeCurves)
                {
                    // ...
                    node->setAdditionalTransform(&transform);
                }
}
```

In the case of Sprite3D children node.

```
void Animate3d::update(float t)
{
    // ...
                for (const auto& it : _nodeCurves)
                {
                    // ...
                    node->setNodeToParentTransform(transform);
                }
    // ...
}
```

PR #15045, opened 2016-02-12
Animate3D. node curve animation bugfix.

yutaka-takeda-drecom:
#15005

Node curve animation bugfix in cocos 3.10.

There is a bug in the case of nodeCurve animation of Aimate3D.

There is a transform of the node in the c3b. (This name is A)
and transform of animation are there in the c3b. (This name is B)
A and B are the same transform.

The animation data at zero frame contains the transform of the node.
In the original source code , transform of the node will be multiplied twice.
(node transform \* animation transform)

But, Sprite3D root node must do the calculation of the transform.
Reason is that the operation of the transform information in non-animated .
(For example, an application programmer change the position to use the API of Sprite3D :: setPosition.)

In conclusion , it is necessary to change the calculation in Sprite3D root node and Sprite3D children  node.

In the case of Sprite3D root node.

```
void Animate3d::update(float t)
{
    // ...
                for (const auto& it : _nodeCurves)
                {
                    // ...
                    node->setAdditionalTransform(&transform);
                }
}
```

In the case of Sprite3D children node.

```
void Animate3d::update(float t)
{
    // ...
                for (const auto& it : _nodeCurves)
                {
                    // ...
                    node->setNodeToParentTransform(transform);
                }
    // ...
}
```


yutaka-takeda-drecom:
[ci rebuild]

minggo:
Sorry, i don't quite understand why there is a difference between root node and children. How can i test?

yutaka-takeda-drecom:
[test.zip](https://github.com/cocos2d/cocos2d-x/files/251898/test.zip)

this is test data.
This is shifted animation of the position of the door.

PR #15047, opened 2016-02-12
Let ScrollView swallow touch events by default.

Zzzen:
This is more intuitive. It's used by `UIWidget` too.


zilongshanren:
@Zzzen 
Thanks.

PR #15048, opened 2016-02-12
Bitmap font sizes are not loaded from binary .fnt files.

### LINKED ISSUE #15042: CCFontFNT missing _fontSize on binary fonts.
Was trying to use some binary fonts when, for some reason, the fonts came out with no fontSize.  I noticed in BMFontConfiguration::parseConfigFile that it would read the fontSize from the text version but it would never set it from the binary version.

I just added these lines to BMFontConfiguration::parseBinaryConfigFile.

``` C
uint16_t fontSize = 0; memcpy(&fontSize, pData, 2);  // add
_fontSize = fontSize;  // add
 _padding.top = (unsigned char)pData[7];
_padding.right = (unsigned char)pData[8];
_padding.bottom = (unsigned char)pData[9];
_padding.left = (unsigned char)pData[10];
```

I don't have access right now to Git any time soon, so thought I post it to see if someone else can confrere.

aardvarkk:
This problem causes bitmap fonts based on .fnt files to not appear at all, and no error is thrown.
Fixes #15042.

### DIFF
--- a/cocos/2d/CCFontFNT.cpp
+++ b/cocos/2d/CCFontFNT.cpp
@@ -397,6 +397,7 @@ std::set<unsigned int>* BMFontConfiguration::parseBinaryConfigFile(unsigned char
              fontName       n+1 string   14 null terminated string with length n
              */
 
+            memcpy(&_fontSize, pData, 2);
             _padding.top = (unsigned char)pData[7];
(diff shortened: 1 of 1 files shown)


zilongshanren:
@aardvarkk 
Thanks :+1:

PR #15050, opened 2016-02-12
Fixes #15049

### LINKED ISSUE #15049: iOS Simulator Build Linking Issues
It appears that the libcocos2d iOS build has only the following valid architectures for Debug and Release:

`VALID_ARCHS = "arm64 armv7";`

This caused me linking headaches in XCode 7.2 as it refused to link against the Cocos libraries when doing a simulator build. This makes sense, because the simulator build is for an x86_64 architecture, not an ARM architecture like the actual device build. By removing the manually listed "valid architectures" and specifying not to only build the active architecture only the linker issues disappeared. I am submitting a pull request to correct this.

aardvarkk:


zilongshanren:
@xiaofeng11  please take a look at this PR, thank.

Issue #14658, opened 2015-12-13
Crashes on iOS 7 and iOS 9


Shulepov:
http://discuss.cocos2d-x.org/t/app-crashes-on-ios9-cocos2dx-2-2-6/25527/5

It crashes a lot on iOS 7 and less often on iOS 9 due to gpus_ReturnNotPermittedKillClient.
I'm receiving thousands оf crash reports from crashlytics for several games (cocos2dx v3.7 & v3.8).
Unfortunately i don't have iOS 7 device so can't reliably reproduce this issue, on my iOS 9 devices i have seen it only once.

```
Thread : Crashed: com.apple.main-thread
0  libGPUSupportMercury.dylib     0x348428f6 gpus_ReturnNotPermittedKillClient + 9
1  libGPUSupportMercury.dylib     0x34843391 gpusSubmitDataBuffers + 104
2  IMGSGX535GLDriver              0x2eb6350f <redacted> + 122
3  GLEngine                       0x32051a01 gliPresentViewES_Exec + 172
4  GLEngine                       0x3205190f gliPresentViewES + 134
5  OpenGLES                       0x3205c0cd -[EAGLContext presentRenderbuffer:] + 64
6  app-mobile           0xb2a95b -[CCEAGLView swapBuffers] + 3563390
7  app-mobile           0xabcc67 cocos2d::Director::drawScene() + 3113610
8  app-mobile           0xabdff7 cocos2d::DisplayLinkDirector::mainLoop() + 3118618
9  QuartzCore                     0x322a2df3 CA::Display::DisplayLinkItem::dispatch() + 98
10 QuartzCore                     0x322a2b9d CA::Display::DisplayLink::dispatch_items(unsigned long long, unsigned long long, unsigned long long) + 344
11 IOMobileFramebuffer            0x3501b75d IOMobileFramebufferVsyncNotifyFunc + 104
12 IOKit                          0x30aa5451 IODispatchCalloutFromCFMessage + 248
13 CoreFoundation                 0x2fd79ea9 __CFMachPortPerform + 136
14 CoreFoundation                 0x2fd84a67 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE1_PERFORM_FUNCTION__ + 34
15 CoreFoundation                 0x2fd84a03 __CFRunLoopDoSource1 + 346
16 CoreFoundation                 0x2fd831d7 __CFRunLoopRun + 1398
[...]

PR #15051, opened 2016-02-13
Fix drawing in incorrect app states on iOS

Shulepov:
1. Sometimes _layoutSubviews_ called from WebThread. Possibly it occurs due to some issues with banners.
2. Sometimes drawScene called too early, while app still is not Active (_applicationDidFinishLaunching_ was called, but _applicationDidBecomeActive_ wasn't).

P.S. Haven't find any reports when app crashed due to using opengl in inactive state after it successfully started. So it may be better to fix this issue only for _start_ flow, and remove _if_ condition inside _doCaller_.

### DIFF
--- a/cocos/platform/ios/CCDirectorCaller-ios.h
+++ b/cocos/platform/ios/CCDirectorCaller-ios.h
@@ -31,6 +31,7 @@
 @interface CCDirectorCaller : NSObject {
         id displayLink;
         int interval;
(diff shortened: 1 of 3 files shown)


Shulepov:
It fixes #14658

zilongshanren:
@Shulepov 
Thanks :+1:

PR #15066, opened 2016-02-15
setPhysicsBody can be called multiple times again

TheCodez:
Before the addition of Physics component, one was able to call `setPhysicsBody` multiple times. The current `setPhysicsBody` method for backwards compability doesn't allow that anymore, because same components can't be added multiple times. This pr fixes that.

This is a updated version of my previous pr https://github.com/cocos2d/cocos2d-x/pull/14371

### DIFF
--- a/cocos/2d/CCNode.h
+++ b/cocos/2d/CCNode.h
@@ -38,6 +38,10 @@
 #include "2d/CCComponentContainer.h"
 #include "2d/CCComponent.h"
 
+#if CC_USE_PHYSICS
+#include "physics/CCPhysicsBody.h"
+#endif
+
 NS_CC_BEGIN
 
 class GridBase;
@@ -1940,9 +1944,9 @@ protected:
     ComponentContainer *_componentContainer;        ///< Dictionary of components
     
     // opacity controls
-    GLubyte         _displayedOpacity;
+    GLubyte     _displayedOpacity;
     GLubyte     _realOpacity;
-    Color3B         _displayedColor;
+    Color3B     _displayedColor;
     Color3B     _realColor;
     bool        _cascadeColorEnabled;
     bool        _cascadeOpacityEnabled;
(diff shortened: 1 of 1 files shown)


TheCodez:
@pandamicro can you get this in 3.11, this has been a poblem since 3.9.
https://github.com/cocos2d/cocos2d-x/pull/14806 is also a possible fix instead of this.

cocos/2d/CCNode.h
```diff
@@ -38,6 +38,10 @@
 #include "2d/CCComponentContainer.h"
 #include "2d/CCComponent.h"
 
+#if CC_USE_PHYSICS
+#include "physics/CCPhysicsBody.h"
+#endif
```
minggo, line 43: Why need including `physics/CCPhysicsBody.h`?

cocos/2d/CCNode.h
```diff
@@ -1862,8 +1866,13 @@ class CC_DLL Node : public Ref
 #if CC_USE_PHYSICS
     PhysicsBody* _physicsBody;
 public:
-    void setPhysicsBody(Component* physicsBody) 
-    { 
+    void setPhysicsBody(PhysicsBody* physicsBody)
```
TheCodez, line 1968: @minggo because `setPhyiscsBody`should only be called with a physicsbody, and if i remember correctly, I can't call removeComponent with just `PhysicsBody`forward declared.

minggo, line 1968: But it break compatibility. And we don't want Node depends on Physics related codes if possible, that's why we use component.

TheCodez, line 1968: If you really want Node not depend on physics you have to also remove `setPhysicsBody` method which was only added for backwards compatibility

If you don't call setPhysicsBody with a PhysicsBody the code crashes when using getPhysisBody method because _physicsBody is null in that case.

In conclusion I don't think changing Component to PhysicsBody here  breaks anyones code, as it makes no sense to call setPhysicsBody with a component other than PhysicsBody.

Hope that makes more sense now.

minggo, line 1968: Ok, i will ask @WenhaiLin to take a look. Why the parameter type is Component.

WenhaiLin, line 1968: @minggo I agree with @TheCodez . The parameter type should be `PhysicsBody*`.

minggo:
@TheCodez Could you please fix the conflict? Then i will merge it.

TheCodez:
@minggo Ok fixed conflicts.

minggo:
@TheCodez Thanks. I will merge it after it passes Jenkins.

TheCodez:
@minggo please get this in 3.11, the bug exists since 3.9

minggo:
Ok, i will send a PR to v3.11.

minggo:
Done in #15576

PR #15069, opened 2016-02-16
Fix value type error when read value of bool type from .plist file.

chengstory:


chengstory:
@ricardoquesada  Please  review this pr, Thanks very much.

cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -60,7 +60,8 @@ static void addItemToArray(id item, ValueVector& array)
         const char* numType = [num objCType];
         if(num == (void*)kCFBooleanFalse || num == (void*)kCFBooleanTrue)
         {
-            array.push_back(Value([num boolValue]));
+            bool v = [num boolValue];
+            array.push_back(Value(v));
```
minggo, line 64: What's the error?
Is there any differences between the modified codes and previous.

ricardoquesada:
@chengstory please, add a test case that triggers the bug. thanks.

cocos/platform/apple/CCFileUtils-apple.mm
chengstory, line 64: `Value([num boolValue])` will call 
`Value::Value(int v)
: _type(Type::INTEGER)
{
    _field.intVal = v;
}
`
this is wrong calling, the correct calling should be
`Value::Value(bool v)
: _type(Type::BOOLEAN)
{
    _field.boolVal = v;
}`
so add `bool v = [num boolValue];`

error:
wrong

correct:
correct

chengstory:
@ricardoquesada This error occurs on `Native Test->FileUtilsTest` of  `js-tests iOS`:
`fileUtils: TestWriteValueMap to files` and `fileUtils: TestWriteValueVector to files`.
Equivalent to fix test case bug, so I think this is not necessary to add another test case that triggers the bug, what do you think? I expect your reply. Thanks so much.

error:
wrong

correct:
correct

cocos/platform/apple/CCFileUtils-apple.mm
minggo, line 64: Yep, you are right.
As mentioned [here](http://stackoverflow.com/questions/541289/objective-c-bool-vs-bool), `BOOL` is defined like this:

```
/// Type to represent a boolean value.
#if (TARGET_OS_IPHONE && __LP64__)  ||  TARGET_OS_WATCH
#define OBJC_BOOL_IS_BOOL 1
typedef bool BOOL;
#else
#define OBJC_BOOL_IS_CHAR 1
typedef signed char BOOL; 
// BOOL is explicitly signed so @encode(BOOL) == "c" rather than "C" 
// even if -funsigned-char is used.
#endif
```

Which means:

> on 64-bit iOS devices and on WatchOS `BOOL` is exactly the same thing as bool while on all other devices (OS X, 32-bit iOS) it is `signed char` and cannot even be overridden by compiler flag `-funsigned-char`

so i think we should convert it to bool explicitly.

minggo, line 64: And `[Number boolValue]` will return a `BOOL` value.

minggo:
@ricardoquesada 
I think this pull request is good to merge as the reason i mentioned above.

PR #15085, opened 2016-02-18
Fixed behavior of "cc.isObject" function for undefined and null values.

1scaR1:
On native platforms if obj is undefined or null then attempt to access **obj.__nativeObj** leads to incorrect behavior of function.

### DIFF
--- a/cocos/scripting/js-bindings/script/jsb_prepare.js
+++ b/cocos/scripting/js-bindings/script/jsb_prepare.js
@@ -129,8 +129,7 @@ cc.isUndefined = function(obj) {
  * @returns {boolean}
  */
 cc.isObject = function(obj) {
-    return obj && obj.__nativeObj !== undefined ||
-        ( typeof obj === "object" && Object.prototype.toString.call(obj) === '[object Object]' );
+    return return ( obj !== null && typeof obj === "object" );
 };
 
 /**
--- a/cocos/scripting/js-bindings/script/jsb_prepare.js
+++ b/cocos/scripting/js-bindings/script/jsb_prepare.js
@@ -129,7 +129,7 @@ cc.isUndefined = function(obj) {
  * @returns {boolean}
  */
 cc.isObject = function(obj) {
-    return obj.__nativeObj !== undefined ||
+    return obj && obj.__nativeObj !== undefined ||
         ( typeof obj === "object" && Object.prototype.toString.call(obj) === '[object Object]' );
 };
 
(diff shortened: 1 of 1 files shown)


minggo:
Thanks.

minggo:
@pandamicro Please take a look.
It looks good to me.

cocos/scripting/js-bindings/script/jsb_prepare.js
```diff
@@ -129,7 +129,7 @@ cc.isUndefined = function(obj) {
  * @returns {boolean}
  */
 cc.isObject = function(obj) {
-    return obj.__nativeObj !== undefined ||
+    return obj && obj.__nativeObj !== undefined ||
```
pandamicro: I didn't realize that it was so complicated, we can simply rewrite it as below

```
cc.isObject = function(obj) {
    return typeof obj === "object";
}
```

minggo: @1scaR1 What's your opinion?

1scaR1: I agree with @pandamicro , i can update this pr with simple code.

minggo: Thanks. I will merge it when you update the PR.

1scaR1:
Iy seems to me that's now it is right:
- `cc.isObject(null) -> false`
- `cc.isObject(1) -> false`
- `cc.isObject(new Number(1)) -> true`
- `cc.isObject({}) -> true`
- `cc.isObject(Object.create(null)) -> true`

cocos/scripting/js-bindings/script/jsb_prepare.js
```diff
@@ -129,8 +129,7 @@ cc.isUndefined = function(obj) {
  * @returns {boolean}
  */
 cc.isObject = function(obj) {
-    return obj.__nativeObj !== undefined ||
-        ( typeof obj === "object" && Object.prototype.toString.call(obj) === '[object Object]' );
+    return ( obj !== null && typeof obj === "object" );
```
minggo, line 132: need to check null?

1scaR1, line 132: Yes, without this check it will return `true` for `null`

1scaR1, line 132: http://wiki.ecmascript.org/doku.php?id=harmony:typeof_null

minggo:
@pandamicro I will merge it if you don't have different opinion.

PR #15091, opened 2016-02-19
Fix return value of VertexAttribBinding::init()

mogemimi:
According to the declaration, `VertexAttribBinding::init()` function must return a bool type value, not null pointer constant. So this pull request replaces `NULL` with `false`, to avoid implicit conversion of null to bool.


minggo:
Thanks.

PR #15092, opened 2016-02-19
Fix all capital letters on Android (Editbox)

ghost:
This patch sets capital letter at sentences start only, not on every single character.

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
@@ -93,7 +93,7 @@ public class Cocos2dxEditBoxHelper {
                 final Cocos2dxEditBox editBox = new Cocos2dxEditBox(mCocos2dxActivity);
                 editBox.setFocusable(true);
                 editBox.setFocusableInTouchMode(true);
-                editBox.setInputFlag(4); //kEditBoxInputFlagInitialCapsAllCharacters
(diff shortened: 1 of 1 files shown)


zilongshanren:
@Delape 
We use all capital letters only for compatible with other platforms, especially iOS.
I think this change will break the back compatibility.

Delape:
Hi @zilongshanren 
What do you mean? This patch is only for the Android platform. Why you say so?
Current EditBox's behaviour is not the one users expect. 

Greetings!

zilongshanren:
@Delape 
I mean on the other platforms, the initial value is also all capital letters... You could call setInputFlag to change the default value. 
I mean we should keep the initial value all the same on all the supported platforms.
thanks.

Delape:
@zilongshanren 
I didn't know about "setInputFlag" existence, that's why I suggested that change in the code.
Thanks for your help :+1: 

I understand your point and now I'm fine with it but I would like to know one last thing:
Why cocos2d-x sets all capital letters by default? Which is the reason behind that decision? It impresses me a lot because It's not the behaviour I would expect (in terms of default mode).

zilongshanren:
@Delape 
I think It's a historical mistake...  maybe @dumganhar  know better about the design decision...

zilongshanren:
@Delape 
If the default is not what all the developers want, we could change this value on all platforms.
@ricardoquesada  Thoughts?

ricardoquesada:
yep. I agree that we shouldn't break compatibility. Migrating to new versions should be smooth.
As @zilongshanren  suggested, using `setInputFlag` looks Ok to me.

zilongshanren:
@ricardoquesada  thanks, I will close this PR now.

PR #15096, opened 2016-02-22
Set music volume control as default

fnz:
Cocos2d-x uses STREAM_MUSIC for audio and video playback, however this stream is not set as a volume control stream. If any audio is playing, pressing volume keys will affect audio level, however, if app is silent at the moment, volume keys will change alarm level.

So the user will have to use volume keys exactly when something is playing to change the sound level which is a bad experience. This commit fixes the issue.

Based on http://developer.android.com/training/managing-audio/volume-playback.html


minggo:
Thanks.

PR #15099, opened 2016-02-23
Add "id" for parser object

feijing566:
Add "id" for parser object


feijing566:
qq 20160223141445

minggo:
Sorry, i am not quite familiar with TiledMap, but i have a question:

Is `id` useful since it is not parsed, and nobody complained it, could you show me the usage? Thanks.

feijing566:
@minggo 
id is the unique identification for object . sometime I need to distinguish between different objects.

minggo:
I search the resources of `cpp-tests`, and could not find `id` element. Is it a new element?

feijing566:
@minggo 
Yes,it is added in version 0.11.0
https://github.com/bjorn/tiled/blob/master/NEWS

minggo:
@feijing566 Thanks for the link.

PR #15117, opened 2016-02-26
Added command lines of building a Android project

lzackx:
It's weird to build and run a Android project with command line as followed
`$ cocos run -p android -j 4`
It seems didn't work for me.
I prefer to use Android Studio when I try to build a Android project,and I can only build it and import it to Android Studio and run successfully with those command lines I added.
So I think that those command lines should be showed directly to who need them.


README.md
```diff
@@ -69,6 +69,15 @@ You can also create a JS project or Lua project with `-l js` or `-l lua`.
 
     $ cocos run -p android -j 4
 
+### Build a new project for Android ###
+
+    $ cd NEW_PROJECTS_DIR/proj.android
+    $ python ./build_native.py
+
```
minggo, line 76: It is better to use `cocos compile -p android`

minggo:
@lzackx What's your opinion about my comment?

minggo:
build_native.sh is deleted, so i close this pr.

Issue #15119, opened 2016-02-26
Can't resume application by CCConsole on iOS.


takumi-shimomura:
Can't resume application by `director start` after executing `director stop` command on CCConsole.
`director pause` and `director resume` pair is too.
`Director::mainLoop` is never called after executing `director start`.
This probrem is recoverable by going to background(taskmgr) once.

Both  iPhone and iOS simulator have that problem. (XCode7, iOS 9)

ricardoquesada:
yep, you are right. Internally it works like this:
- Director will create the Console
- Director will delete the Console when `Director::purge` is called

So, when you call `director end` from the console, it will destroy the console as well.
the solution is to dissociate the `Console` from the `Director`... and users should start and end the console manually, and the console should be a singleton... which makes sense if you want to control everything from the console... but low priority for the moment.

Issue #15031, opened 2016-02-09
could you get PageView EventListener CallBack in v3.10 c++ ?


xiongdi9456:
I tested  PageView in v3.10 by c++, can not get EventListener when the page was turning.
I run the same code in v3.9 , it works.

zilongshanren:
https://github.com/cocos2d/cocos2d-x/pull/14964

PR #15013, opened 2016-02-03
Console refactor

m-yukio:
Refactoring of existing code.
1. Common
   - And the private functions to the methods of Utility class, made available from external code to add the command.
   - The option of command reconstituted in sub-command was set to be held in the nest.
   - From member variables "help" of Command structure, it was to be able to output the help of sub-command.
   - It was to organize the command options handling.
2. command Allocator
   - supported option "-h" and option "help".
3. command Config
   - supported option "-h" and option "help".
4. command DebugMsg
   - supported option "-h" and option "help".
   - Set the help wording options.
5. command Director
   - deleted option "help" code.
   - Set the help wording options.
6. command Exit
   - supported option "-h" and option "help".
7. command FileUtils
   - supported option "-h" and option "help".
   - Quit the error wording output when an invalid option is specified.
   - Set the help wording options.
   - Non-flush option (including the empty) was so as to print when you are specified.
8. command Fps
   - supported option "-h" and option "help".
   - Set the help wording options.
   - Non-on/off option (including the empty) was so as to print when you are specified.
9. command Help
   - supported option "-h" and option "help".
10. command Projection
    - supported option "-h" and option "help".
    - Set the help wording options.
    - Quit the error wording output when an invalid option is specified.
    - Non-2d/3d option (including the empty) was so as to print when you are specified.
11. command Resolution
    - supported option "-h" and option "help".
12. command SceneGraph
    - supported option "-h" and option "help".
13. command Texture
    - supported option "-h" and option "help".
    - Set the help wording options.
    - Quit the error wording output when an invalid option is specified.
    - Non-flush option (including the empty) was so as to print when you are specified.
14. command Touch
    - deleted option "help" code.
    - Set the help wording options.
15. command Upload
    - No change
16. command Version
    - supported option "-h" and option "help".

PR #15120, opened 2016-02-26
Add 'trap' command to the CCConsole

takumi-shimomura:
## About

Trap command monitors the application, to stop the application if it exceeds a set threshold.
I propose this command for paformance test usage.
Application is stopped by Director::stopAnimation.
## Example

```
// set threthold. fire when FPS falls below 30.
> trap fps 30

// start inspecting.
> trap start

// (play something heavy scene)
traped by fps threshold.

// show scene graph that saved just trap. 
// to use the scenegraph subcommand instead of default scenegraph command,
// because performFunctionInCocosThread dose not work when director stopped.
> trap scenegraph

// resume application
> director start
```

(`director start` is not work on iOS. See Issue #15119)

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -240,7 +240,7 @@ namespace {
     };
 
     Trap::Trap ()
-    : _command("trap", "")
+    : _command("trap", "Inspect application, to stop application if it exceeds a set threshold")
     {
         _command.addSubCommand({"start", ": start inspecting.", [](int fd, const std::string &args) {
             Director::getInstance()->getScheduler()->performFunctionInCocosThread([fd](){
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -170,7 +170,7 @@ namespace {
                 if (drawCallsThreshold > 0) {
                     auto drawnBatches = Director::getInstance()->getRenderer()->getDrawnBatches();
                     if (drawnBatches >= drawCallsThreshold) {
-                        trap ("traped by draw call threshold.");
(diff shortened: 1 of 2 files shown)


takumi-shimomura:
Thanks for reviewing!

takumi-shimomura:
[ci rebuild]

minggo:
What does this command used for? It seems it depends heavily on fps, and fps is not so accurate.

takumi-shimomura:
Sure, your opinion is right.
This command is only provides simple method for performance testing.
It can not use as argument other then application status variables (FPS, VertexCount and DrawCall).

I think, if function of getting Memory Usage was supported by cocos2dx, trap command would improve usability.

Do you think we should support interface for setting trigger variable?
Not only FPS but also other things?

minggo:
I think Console is designed to be a module that developers can add their own commands as needed, but not add all commands into cocos2d-x core.

@ricardoquesada What's your opinion? Am i right about the Console module?

ricardoquesada:
I think it is a nice feature. But I would put it first in the cpp-test as an example... later, we can put in the Console if there is demand for it.

minggo:
@ricardoquesada Thanks for you advise.
@takumi-shimomura I think this PR can be refactored after #15013, and you can take a look of that PR. You are appreciated if you can give some advise too.

minggo:
@takumi-shimomura I have merged #15031.

minggo:
@takumi-shimomura Will you please add the command into `ConcoleTest.cpp` as @ricardoquesada suggested.

minggo:
@takumi-shimomura What's your opinion?

PR #15123, opened 2016-02-26
added methods to set individual components of velocity of a PhysicsBody

justinl1996:
Sometimes I find it annoying having to set the individual components of velocity of a body. It would usually involve something that looks like this in the case where I just want to set the x-component of velocity, but leave the y-component unchanged:

x = 100;
body->setVelocity(x, body->getVelocity().y);

I propose an alternative, more nicer solution, by introducing the methods:
setVelocityX();
setVelocityY();
which allow the individual components to be set, whilst leaving the other unchanged.

For completedness I have also included the methods:
getVelocityX();
getVelocityY()

### DIFF
--- a/cocos/physics/CCPhysicsBody.h
+++ b/cocos/physics/CCPhysicsBody.h
@@ -261,10 +261,10 @@ public:
     virtual Vec2 getVelocity();
 
     /** Get the x-component of velocity of a body. */
-    float getVelocityX();
+    float getVelocityX() const;
 
      /** Get the x-component of velocity of a body. */
-    float getVelocityY();   
+    float getVelocityY() const;   
     
     /** 
      * Set the angular velocity of a body.
--- a/cocos/physics/CCPhysicsBody.h
+++ b/cocos/physics/CCPhysicsBody.h
@@ -242,9 +242,29 @@ public:
      * @param velocity The velocity is set to this body.
      */
     virtual void setVelocity(const Vec2& velocity);
-    
+
+    /**
+    * Set the x-component of velocity of a body. The y-component remains unchanged.
+    *
+    * @param velocity The x component of velocity set to this body
+    */
+    virtual void setVelocityX(float velocity);
+
+     /**
+    * Set the y-component of velocity of a body. The x-component remains unchanged.
+    *
+    * @param velocity The y component of velocity set to this body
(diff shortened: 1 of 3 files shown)


cocos/physics/CCPhysicsBody.h
```diff
@@ -242,9 +242,29 @@ class CC_DLL PhysicsBody : public Component
...
+    *
+    * @param velocity The y component of velocity set to this body
+    */
+    virtual void setVelocityY(float velocity);   
+
     /** Get the velocity of a body. */
     virtual Vec2 getVelocity();
+
+    /** Get the x-component of velocity of a body. */
+    float getVelocityX();
```
minggo: float getVelocityX() -> float getVelocityX() const ?

cocos/physics/CCPhysicsBody.h
```diff
@@ -242,9 +242,29 @@ class CC_DLL PhysicsBody : public Component
...
+
+     /** Get the x-component of velocity of a body. */
+    float getVelocityY();
```
minggo: float getVelocityY() -> float getVelocityY() const ?

justinl1996: I will fix that

cocos/physics/CCPhysicsBody.h
```diff
@@ -242,9 +242,29 @@ class CC_DLL PhysicsBody : public Component
...
+    virtual void setVelocityY(float velocity);   
+
     /** Get the velocity of a body. */
     virtual Vec2 getVelocity();
+
+    /** Get the x-component of velocity of a body. */
+    float getVelocityX() const;
+
+     /** Get the x-component of velocity of a body. */
+    float getVelocityY() const;
```
minggo, line 267: Should `getVelocityX` and `getVelocityY` be virtual functions?

minggo, line 267: @greekman99 Any idea?

justinl1996, line 267: will fix, sorry I'm new

minggo, line 267: It is ok.

minggo:
@greekman99 any update?

PR #15126, opened 2016-02-26
Add program-generated geometry test case

shadowphiar:
Add a small test case showing how Sprite3D can support displaying of geometry which has been generated by the program at run-time, instead of loading a predetermined model.


minggo:
@shadowphiar thanks. cpp-tests is to use test engine features to find potential bugs. This PR is just want to share some codes?

DelinWorks (approved):

PR #15128, opened 2016-02-27
Allow Sprite3D::setTexture to specify texture usage

shadowphiar:
Allow a texture usage to be specified for Sprite3D::setTexture (as it can for Mesh::setTexture)

### DIFF
--- a/cocos/3d/CCSprite3D.cpp
+++ b/cocos/3d/CCSprite3D.cpp
@@ -664,6 +664,20 @@ void Sprite3D::setTexture(Texture2D* texture)
         mesh->setTexture(texture);
     }
 }
+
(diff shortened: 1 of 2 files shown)


minggo:
@shadowphiar Could you please add a test case?

minggo:
[ci rebuild]

cocos/3d/CCSprite3D.cpp
```diff
@@ -664,6 +664,20 @@ void Sprite3D::setTexture(Texture2D* texture)
         mesh->setTexture(texture);
     }
 }
+
+void Sprite3D::setTexture(const std::string& texFile, NTextureData::Usage usage)
+{
+    auto tex = Director::getInstance()->getTextureCache()->addImage(texFile);
```
minggo, line 670: Should determine the value of `tex`?

PR #15145, opened 2016-03-02
three related bug fixes in FontAtlas::prepareLetterDefinitions

grwhitehead:
I found four bugs in FontAtlas::prepareLetterDefinitions in version 3.8.1

One of those bugs has already been fixed, in commit bae36b7a98dc27d68ce1aa690d9ef30516671692

This branch contains fixes for the other three bugs:
1) _currLineHeight shouldn't be updated until after we've checked for overflow
2) the check for vertical overflow should include _letterPadding and _letterEdgeExtend, which can make the line higher than _lineHeight
3) _currLineHeight should be used instead of _lineHeight when updating the texture, since it might be greater than _lineHeight

These bugs, and the one already fixed, all lead to previous letter textures being partially overwritten by new letter textures. The bugs are triggered when the letter that causes a horizontal overflow is taller than any letter that follows it before the next horizontal overflow.

Note that the order that letters are rendered in is determined by key iteration over an unordered_map, so it is hard to predict.

### DIFF
--- a/cocos/2d/CCFontAtlas.cpp
+++ b/cocos/2d/CCFontAtlas.cpp
@@ -341,17 +341,12 @@ bool FontAtlas::prepareLetterDefinitions(const std::u16string& utf16Text)
             tempDef.offsetX = tempRect.origin.x + adjustForDistanceMap + adjustForExtend;
             tempDef.offsetY = _fontAscender + tempRect.origin.y - adjustForDistanceMap - adjustForExtend;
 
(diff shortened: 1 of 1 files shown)


zilongshanren:
@WenhaiLin

WenhaiLin:
@grwhitehead :+1:  @zilongshanren This PR is OK for me.

zilongshanren:
[ci rebuild]

zilongshanren:
[ci rebuild]

zilongshanren:
@grwhitehead 
Thanks :+1:

PR #15153, opened 2016-03-02
[ci skip]add a function to get object name

minggo:
@pandamicro 
Please take a look. It only has effect in VSCode debugger.


minggo:
@pandamicro 
Do you have any question of this PR?
I will merge it if not.

PR #15159, opened 2016-03-03

CocosRobot:


zilongshanren:
@ricardoquesada  Hi, please take a look at this PR, it was made by CocosRobot, because we lack some jsb config files in _tools/tojs_ directory, so every time when the main repo has a merge commit, it will send a PR to remove the jsb bindings you added a few days ago.

please help...

ricardoquesada:
@zilongshanren well, of course it shouldn't remove those files.

sure, what is the best way to integrate this script in the builder ?https://github.com/ricardoquesada/jsbindings

The problem was the the chipmunk auto-generated files (which are generated using a different script) were placed in the "manual" directory... and people modified them manually.

correct solution: include the new script in the server? Don't know how to do that... I will need help.
incorrect solution: create a new directory called "auto2" and put the chipmunk files there.

zilongshanren:
@ricardoquesada 
The @CocosRobot  use https://github.com/cocos2d/cocos2d-x/blob/v3/tools/travis-scripts/generate-bindings.sh file to generate all the lua and js bindings automatically when upstream has a merge event. We deploy this service in the travis. That's the magic part, you could take a closer look at the code here:

```
TOLUA_ROOT="$PROJECT_ROOT/tools/tolua"
TOJS_ROOT="$PROJECT_ROOT/tools/tojs"
LUA_COMMIT_PATH="cocos/scripting/lua-bindings/auto"
JS_COMMIT_PATH="cocos/scripting/js-bindings/auto"

generate_bindings_glue_codes()
{
    echo "Create auto-generated luabinding glue codes."
    pushd "$TOLUA_ROOT"
    ./genbindings.py
    popd

    echo "Create auto-generated jsbinding glue codes."
    pushd "$TOJS_ROOT"
    ./genbindings.py
    popd
}
```

The script will run the `genbindings.py` in the TOLUA_ROOT and TOJS_ROOT.
So what we should do is to modify genbindings.py in TOLUA_ROOT or TOJS_ROOT.

The genbindings.py use https://github.com/cocos2d/bindings-generator/tree/0bec01d038c2d764d5e6244b41d149efb858b243 to generate the bindings.

So I think you should send a patch to the repo, merge back your change to chipmunk and make it work with binding generator.

Hope this helps.

ricardoquesada:
I could do that, although it won't be easy for me since I don't know how our backend system work. If someone could do that, great. If not, I could do that after finishing what I'm working on.

@linshun what should I do? Should I work on this after finishing cmake for Android?

But I think the easiest solution is to tell the server to NOT remove the chipmunk files from the "auto" directory.

zilongshanren:
@ricardoquesada  Yeah, I will try to  tell the server to NOT remove the chipmunk files from the "auto" directory.

ricardoquesada:
great. if you could do that, that's great!! thanks!

PR #15171, opened 2016-03-03
[ci skip] js-test bug fix

xiaofeng11:
@pandamicro @zilongshanren Please review this PR.
Fix playMusic test case under CocosDenshionTest won't work on android system
Fix click setNotificationNode button in SceneTest then return to main menu the animation is still play on screen

### DIFF
--- a/tests/js-tests/src/CocosDenshionTest/CocosDenshionTest.js
+++ b/tests/js-tests/src/CocosDenshionTest/CocosDenshionTest.js
@@ -231,7 +231,11 @@ var soundId = null;
 
 var playMusic = function () {
     cc.log("play background music");
-    audioEngine.playMusic(MUSIC_FILE, false);
+    var musicFile = MUSIC_FILE;
+    if (cc.sys.os == cc.sys.OS_ANDROID) {
+        musicFile = "res/"+musicFile;
+    }
+    audioEngine.playMusic(musicFile, false);
 };
 
 var stopMusic = function () {
--- a/tests/js-tests/src/SceneTest/SceneTest.js
+++ b/tests/js-tests/src/SceneTest/SceneTest.js
@@ -103,7 +103,8 @@ var SceneTestLayer1 = cc.Layer.extend({
 
         director.pushScene(new cc.TransitionSlideInT(1, scene));
     },
-    onQuit:function (sender) {
+    onExit:function (sender) {
+        cc.director.setNotificationNode(null);
     }
 
     //CREATE_NODE(SceneTestLayer1);


tests/js-tests/src/SceneTest/SceneTest.js
```diff
@@ -103,7 +103,8 @@ var SceneTestLayer1 = cc.Layer.extend({
 
         director.pushScene(new cc.TransitionSlideInT(1, scene));
     },
-    onQuit:function (sender) {
+    onExit:function (sender) {
+        cc.director.setNotificationNode(null);
```
zilongshanren, line 107: why add this line?

xiaofeng11, line 107: This line is fix another bug in comment: "Fix click setNotificationNode button in SceneTest then return to main menu the animation is still play on screen"

tests/js-tests/src/CocosDenshionTest/CocosDenshionTest.js
```diff
@@ -231,7 +231,11 @@ var soundId = null;
 
 var playMusic = function () {
     cc.log("play background music");
-    audioEngine.playMusic(MUSIC_FILE, false);
+    var musicFile = MUSIC_FILE;
+    if (cc.sys.os == cc.sys.OS_ANDROID) {
+        musicFile = "res/"+musicFile;
+    }
+    audioEngine.playMusic(musicFile, false);
```
minggo, line 238: But why other resources, such as .png file don't have to add prefix `res`? For example, all the file path in [test_resources.js](https://github.com/cocos2d/cocos2d-x/blob/v3/tests/js-tests/src/tests_resources.js) don't need to add prefix `res`?

xiaofeng11, line 238: @minggo These png file path is used in C/C++ layer, it have extra process to get right path, but in CocosDenshionTest, these param will directly send to Java layer, which not have search path or similar process to find full path of the file. Please refer to Java layer implementation at https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxMusic.java#L244.

minggo, line 238: > These png file path is used in C/C++ layer, it have extra process to get right path

Could you show me the codes about it?

xiaofeng11, line 238: @minggo Please refer https://github.com/cocos2d/cocos2d-x/blob/v3/tests/js-tests/main.js#L118, these search path will effect in C/C++ layer, but it won't effect in Java layer

minggo, line 238: Got it.

PR #15189, opened 2016-03-06
Bug fix: CCValue operator overloading of Comparison "=="

Sai628:
Bug fix: CCValue operator overloading of Comparison "==" in case Type::VECTOR


minggo:
@Sai628 This pull request do more thing than the title said.
Could you please just do one thing at a pull request?

PR #15191, opened 2016-03-07
[preview] Adds thread pool support

dumganhar:


dumganhar:
This PR is just for preview. Please leave comments if you have any advice or questions.

WenhaiLin:
:+1:

zilongshanren:
@yangws  please take a look at this PR.

dumganhar:
@minggo ,could you please review this PR? I think common threadpool class is a big improvement for Cocos2d-x.

I know this fix may need more polish, but just see whether the design match our needs. :)

minggo:
I will take a look when i have time. This PR may take more time.

dumganhar:
Thank you. :)

minggo:
I have some questions:
- what's the difference from `AsyncTaskPool`
- it seems the class mixed all types of thread pool and use flags to do different things, can use inheritance internally, such as `CacheThreadPool` inherits from ThreadPool, of course `CacheThreadPool` is internally implementation

dumganhar:
> > what's the difference from AsyncTaskPool

AsyncTaskPool just uses only only one thread for a type of task.

``` c++
    enum class TaskType
    {
        TASK_IO,
        TASK_NETWORK,
        TASK_OTHER,
        TASK_MAX_TYPE,
    };
```

It means that `IO` will use one thread, `NETWORK` also use one thread.

ThreadPool shares all threads for tasks.

> > it seems the class mixed all types of thread pool and use flags to do different things, can use inheritance internally, such as CacheThreadPool inherits from ThreadPool, of course CacheThreadPool is internally implementation

Seems a better idea, if in that case, I think the inheritance should not be internal, it should be public. Something like:

``` c++
static CachedThreadPool* newCachedThreadPool(int minThreadNum, int maxThreadNum, int shrinkInterval, int shrinkStep, int stretchStep);
static FixedThreadPool* newFixedThreadPool(int threadNum);
static SingleThreadPool* newSingleThreadPool();
```

BTW, `AsyncTaskPool` is more like `SingleThreadPool` here. :)

halx99:
I think, for client program, multi-threadpool is not necessary.

cezheng:
I'm not sure how good the performance of this implementation is, but I think having a thread pool is definitely good for many asynchronous loading scenario.

iOS has the advantage of libdispatch in its runtime, which is a highly performant thread pool implementation, since it is optimized for apple's hardware.

I tried replacing the async loading code using AsyncTaskPool with `dispatch_async`s in the project I'm currently working on, it gives my a 20%-30% shorter loading time on a dual-core iPhone 5, which is very appealing(Many Android phone have up to 8 cores today, so the performance boost would be even better on these devices).

For Android, libdispatch is ported by PSPDFKit-labs last year: https://github.com/PSPDFKit-labs/libdispatch

But later they moved away from it and switched to boost due to the fact that none of the Android debugging & instrument tools understand the block syntax. (refer to this blog post: https://pspdfkit.com/blog/2016/a-pragmatic-approach-to-cross-platform)

I'm yet to test the actual performance of `boost::asio::io_service` to decide wether to use it or not.

minggo:
I think using an mature framework is better if there is one.

halx99:
I found `CCThreadPool` already add to cocos/audio/android/ by other guy.

minggo:
It is added by @dumganhar, only used in audio engine to have good performance. We can replace it if we found a mature library.

dumganhar:
@halx99 , currently CCThreadPool is only used for AudioEngine of Android platform.
It's internal use only.

PR #15194, opened 2016-03-07
Modified the layer follow action class to include a horizontal (x) and vertical (y) offset.

AlwynC:
The layer follow action by default centers on the node it is following.I've added two offset parameters for horizontal (x) and vertical (y) offset so that the layer will follow the node based on the vertical and horizontal offset which is set while creating the follow action without centering on the node it is following.In case,centered following is required,the offset parameters can be set to zero.


aseem2625:
This would be nice to have in-built feature as there are much lesser games with character in the center than character at around 1/4 of the screen as the keeping in center wastes major portion of screen.

@AlwynC  Thanks for PR

minggo:
@AlwynC Could you please add a test case?

AlwynC:
@minggo I've added a test case.

minggo:
@AlwynC Thanks. I will try it ASAP.

AlwynC:
@minggo Sure.

cocos/2d/CCAction.h
```diff
@@ -323,8 +323,15 @@ class CC_DLL Follow : public Action
...
+     * @param xOffset The horizontal offset from the center of the screen from which the
+     *               node  is to be followed.It can be positive,negative or zero.If
+     *               set to zero the node will be horizontally centered followed.
+     *  @param yOffset The vertical offset from the center of the screen from which the
+     *                 node is to be followed.It can be positive,negative or zero.
+     *                 If set to zero the node will be vertically centered followed.
+     *   If both xOffset and yOffset are set to zero,then the node will be horizontally and vertically centered followed.
+     */
+    static Follow* create(Node *followedNode, const Rect& rect = Rect::ZERO,float xOffset=0.0,float yOffset=0.0);
     /** Return boundarySet.
```
minggo, line 347: It is better to add a new `create` action according [cocos2d-x coding style](https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#function-overloading).

minggo:
I run the test case, but can not find any difference from previous Follow action test.

AlwynC:
@minggo I will re-look the test case.

AlwynC:
@minggo I've modified the test case so as to show the offset.Since the background layer is black and the sprite moves quite fast,it is rather difficult to observe the offset.The sprite which is being followed in the test case which I've written is not in the centre of the screen when followed and is displaced to the right side and a little to the bottom because of the horizontal and vertical offsets which have been set.This is how it differs from the previous follow action test.I've tried out this approach of providing offsets for the follow action in a game project using cocos 2d-x which I'm currently working on and it works very well.Please have a look.

minggo:
@AlwynC Thanks. Could you please add a new create function as i said above.
Thanks.

AlwynC:
@minggo I've added a new create function for follow action with offsets as you have suggested above.

cocos/2d/CCAction.h
```diff
@@ -323,8 +323,27 @@ class CC_DLL Follow : public Action
...
+     *               node  is to be followed.It can be positive,negative or zero.If
+     *               set to zero the node will be horizontally centered followed.
+     *  @param yOffset The vertical offset from the center of the screen from which the
+     *                 node is to be followed.It can be positive,negative or zero.
+     *                 If set to zero the node will be vertically centered followed.
+     *   If both xOffset and yOffset are set to zero,then the node will be horizontally and vertically centered followed.
+     */
+
+    static Follow* createWithOffset(Node* followedNode,const Rect& rect = Rect::ZERO,float xOffset = 0.0,float yOffset = 0.0);
+
```
minggo, line 346: I think default value of `xOffset` and `yOffset` is not needed.

cocos/2d/CCAction.h
```diff
@@ -381,8 +402,16 @@ class CC_DLL Follow : public Action
...
+     *                node  is to be followed.It can be positive,negative or zero.If
+     *                set to zero the node will be horizontally centered followed.
+     * @param yOffset The vertical offset from the center of the screen from which the
+     *                node is to be followed.It can be positive,negative or zero.
+     *                If set to zero the node will be vertically centered followed.
+     *   If both xOffset and yOffset are set to zero,then the node will be horizontally and vertically centered followed.
+
      */
-    bool initWithTarget(Node *followedNode, const Rect& rect = Rect::ZERO);
+    bool initWithTarget(Node *followedNode, const Rect& rect = Rect::ZERO,float xOffset=0.0,float yOffset=0.0);
```
minggo: I think we need a new initWithTarget() for new create function.

minggo:
@AlwynC Except to comment above, i think the old functions can be write with

``` c++
Follow* Follow::create(Node *followedNode, const Rect& rect = Rect::ZERO)
{
    return create(followedNode,rect, 0, 0);
}

bool Follow::initWithTarget(Node *followedNode, const Rect& rect = Rect::ZERO)
{
    return initWithTarget(followedNode, rect, 0, 0);
}
```

AlwynC:
@minggo I will consider your suggestions above and refine the code further.

AlwynC:
@minggo I've refined the code as you have suggested above.Please have a look.

cocos/2d/CCAction.cpp
```diff
@@ -192,29 +192,47 @@ Follow::~Follow()
...
+    bool valid;
+    
+    if(xOffset == 0.0 && yOffset == 0.0){
+        
+        valid = follow->initWithTarget(followedNode,rect);
+    
+    }else{
+        
+        valid = follow->initWithTargetAndOffset(followedNode, xOffset, yOffset,rect);
+    }
```
minggo: Can just invoke

``` c++
valid = follow->initWithTargetAndOffset(followedNode, xOffset, yOffset,rect);
```

here?

minggo: I mean don't have to compare `xOffset` and `yOffset`, just invoke

``` c++
valid = follow->initWithTargetAndOffset(followedNode, xOffset, yOffset,rect);
```

AlwynC:
@minggo Please have a look at the latest code.Shall I remove the old initWithTarget method as it is not being used anywhere?

minggo:
@AlwynC Thanks. I don't think should remove old initWithTarget since it may be used by developers.

AlwynC:
@minggo Sure.I've not removed the old method.

AlwynC:
[ci skip]

aseem2625:
@minggo 
This is merged by can you please tell since which version will it be available?

minggo:
@aseem2625 as you can see the `Milestone` is set to `3.12`.

PR #15198, opened 2016-03-07
Wrong log message if use tga image

perminovVS:
detectFormat don't need log message.
initWithImageData call detectFormat, and if Format::UNKNOWN check tga format.
Before check tga, CCLOG("cocos2d: unsupported image format!");

### DIFF
--- a/cocos/platform/CCImage.cpp
+++ b/cocos/platform/CCImage.cpp
@@ -724,7 +724,6 @@ Image::Format Image::detectFormat(const unsigned char * data, ssize_t dataLen)
     }
     else
     {
-        CCLOG("cocos2d: can't detect image format");
         return Format::UNKNOWN;
     }
 }


perminovVS:
[ci rebuild]

perminovVS:
[ci rebuild]

perminovVS:
[ci rebuild]

minggo:
Sorry, i can't catch your idea.

perminovVS:
@minggo if use tga image format, always call CCLOG("cocos2d: unsupported image format!");

minggo:
So it means we should check the format of `tga` since it is supported?

perminovVS:
@minggo no.
Now call detectFormat, but it no check tga format, check only other format.
After called detectFormat, if no support all format(except tga), checked tga, and if no tga format, send log message CCLOG("cocos2d: unsupported image format!");

detectFormat should not send log message, it called after check tga format.

minggo:
What i mean is that, why not checking tga in this function and can work correctly. I will do more research.

perminovVS:
@minggo ok

PR #14659, opened 2015-12-13
Resizable window

elnormous:
Add "resizable" flag to desktop GLView to allow creating a resizable window

PR #15202, opened 2016-03-07
Correct work onGLFWWindowSizeFunCallback on desktop

perminovVS:
Incorrect calculate Viewport, if minimize and maximize window.
Incorrect calculate y mouse position
Before #14659


perminovVS:
[ci rebuild]

perminovVS:
Merged here #15236

PR #13599, opened 2015-08-27
WinRT support mouse

perminovVS:
Show/Hide mouse cursor.
Mouse event implemented similar Desktop version.
Left button send mouse event and touch.
Support other mouse button and scroll wheel.

PR #15203, opened 2016-03-07
Menu items react on mouse focus

perminovVS:
Item react if TouchMoved
If mouse move, item not react

### DIFF
--- a/cocos/2d/CCMenu.cpp
+++ b/cocos/2d/CCMenu.cpp
@@ -170,8 +170,8 @@ bool Menu::initWithArray(const Vector<MenuItem*>& arrayOfItems)
         
         _eventDispatcher->addEventListenerWithSceneGraphPriority(touchListener, this);
 
-
-
+        // For Desktop platform listen mouse
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC || CC_TARGET_PLATFORM == CC_PLATFORM_WIN32 || CC_TARGET_PLATFORM == CC_PLATFORM_LINUX)
         auto mouseListener = EventListenerMouse::create();
 
         mouseListener->onMouseMove = [this](EventMouse* event) {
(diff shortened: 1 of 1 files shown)


Dimon4eg:
Hi,
I think this is relevant only for desktop and no need to add mouse listener for mobile targets.

perminovVS:
@minggo @Dimon4eg add only desktop platform

perminovVS:
@minggo merge please, it convenient to use

minggo:
@perminovVS add a feature not work for all platforms is not a good idea, we should avoid it.

zilongshanren:
@minggo I think only Desktop platforms support mouse, so this PR is safe, right?

minggo:
Yep, this PR is safe, but as i said, we should avoid features that only work for some platforms.

zilongshanren:
I think this feature could be considered as a feature working for all the supported platforms.

Mouse events are only valid on Desktops? Maybe Windows universal also support mouse?

I'm not sure...

perminovVS:
@zilongshanren uwp support mouse now, i add it #13599.
Need api for check support mouse and hardware keyboard in realtime.
Example android support mouse, and not all pc support mouse.

zilongshanren:
@perminovVS  but in this PR, you only add mouse support for win32, linux and mac?

perminovVS:
@zilongshanren now yes, i add it before
Dimon4eg commented on 14 Mar
Hi,
I think this is relevant only for desktop and no need to add mouse listener for mobile targets.
 @zilongshanren zilongshanren was assigned by minggo on 6 Apr
 @minggo minggo added type:feature module:ui labels on 6 Apr

Initial it work for all platform, listen mouseListener->onMouseMove
If send, it work.

pyrosphere:
For desktop games having menus react to the mouse is an important feature, otherwise the game will feel like a lazy mobile port.
Ideally MenuItem would have a 4th state 'hover' (besides 'normal', 'selected' and 'disabled') but this PR is a good compromise.

PR #15204, opened 2016-03-07
GLProgram replaceDefines correct and other feature

perminovVS:
replaceDefines if input more ; incorrect out
Example:
compileTimeDefines = "MAX_OPTIMIZATION;USE_PARAM1;"
out:
# define MAX_OPTIMIZATION
# define USE_PARAM1
# define

some device not compile this shader

Now compileTimeDefines = ";;MAX_OPTIMIZATION;;;USE_PARAM1;;;;"
out correct:
# define MAX_OPTIMIZATION
# define USE_PARAM1

### DIFF
--- a/cocos/renderer/CCGLProgram.cpp
+++ b/cocos/renderer/CCGLProgram.cpp
@@ -44,14 +44,23 @@ THE SOFTWARE.
 static void replaceDefines(const std::string& compileTimeDefines, std::string& out)
 {
     // Replace semicolons with '#define ... \n'
-    if (compileTimeDefines.size() > 0)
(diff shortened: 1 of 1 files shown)


perminovVS:
[ci rebuild]

ricardoquesada:
Thanks!

ricardoquesada:
ouch... it is crashing right now with the Material Tests... I'll fix the patch.
The problem with this patch is that if the define doesn't end with a `;` it won't use it, and there is code that are setting defines without the final `;` and it must be 100% backward compatible

perminovVS:
@ricardoquesada fix it #15376

PR #15215, opened 2016-03-09
[ci skip]fix a bug that will miss packets

minggo:
@pandamicro Could you please take a look.
Thanks.


minggo:
Android device debug need this patch.

PR #15220, opened 2016-03-10
fix scrollview bugs

geron-cn:
##### need merge  https://github.com/dumganhar/ccs-res/pull/57

fix scrollview bugs: 
- fix bar's position not right after size changed
- fix bar setAutoHideEnabled some times not right


geron-cn:
[ci rebuild]

PR #15231, opened 2016-03-12
Fix calculation of base64Encode outLength

SteveKChiu:
The order of integer arithmetic is important, the original code fail to
calculate the length correctly, e.g. inLength 20 for base64Encode should
output 28, not 30. It is fixed and simplified with this commit.

### DIFF
--- a/cocos/base/base64.cpp
+++ b/cocos/base/base64.cpp
@@ -142,7 +142,7 @@ int base64Decode(const unsigned char *in, unsigned int inLength, unsigned char *
     unsigned int outLength = 0;
     
     //should be enough to store 6-bit buffers in 8-bit buffers
-    *out = (unsigned char*)malloc(inLength * 3.0f / 4.0f + 1);
+    *out = (unsigned char*)malloc(inLength / 4 * 3 + 1);
     if( *out ) {
         int ret = _base64Decode(in, inLength, *out, &outLength);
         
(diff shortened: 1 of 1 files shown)


minggo:
@SteveKChiu How can it test?

SteveKChiu:
The change is simple and obvious, but if you need test sample, try to call base64Encode with inLength from 1 to 24, then you should about to identify the problem:

``` c++
const char* in = "123456789012345678901234";
for (int inLen = 1; inLen <= 24; inLen++) {
    char* out;
    int outLen = base64Encode(in, inLen, &out);
    printf("base64Encode %d -> %d\n", inLen, outLen);
    free(out);
} 
```

The output should be:

```
base64Encode 1 -> 4
base64Encode 2 -> 4
base64Encode 3 -> 4
base64Encode 4 -> 8
base64Encode 5 -> 8
base64Encode 6 -> 8
base64Encode 7 -> 12
base64Encode 8 -> 12
base64Encode 9 -> 12
base64Encode 10 -> 16
base64Encode 11 -> 16
base64Encode 12 -> 16
base64Encode 13 -> 20
base64Encode 14 -> 20
base64Encode 15 -> 20
base64Encode 16 -> 24
base64Encode 17 -> 24
base64Encode 18 -> 24
base64Encode 19 -> 28
base64Encode 20 -> 28
base64Encode 21 -> 28
base64Encode 22 -> 32
base64Encode 23 -> 32
base64Encode 24 -> 32
```

minggo:
I am not familiar with base64 encoding. After reading the algorithm definition, i think you are right. Thanks.

PR #15235, opened 2016-03-14
More fix to compile libjscocos2dx as dll under win32

xiaofeng11:


cocos/scripting/js-bindings/manual/chipmunk/js_bindings_chipmunk_manual.cpp
```diff
@@ -1081,7 +1081,7 @@ bool __jsb_cpSpace_addCollisionHandler(JSContext *cx, jsval *vp, jsval *argvp, J
 
     cpCollisionHandler* cp_collision_handler = cpSpaceAddCollisionHandler(space, handler->typeA, handler->typeB);
     cp_collision_handler->userData = handler;
-#ifdef _USRDLL
+#if defined(_WINDOWS) && (defined(_WIN32) || defined(WIN32))
```
pandamicro, line 1084: I think you should keep _USRDLL here too

xiaofeng11, line 1084: _USRDLL will conflict with 3d module js-binding, so I remove it. on win32 platform use function call here won't cause big performance lost.

PR #15239, opened 2016-03-14
V3 scissors fix

newnon:
Fix fox scissors clipping type


zilongshanren:
@newnon 
How to reproduce this issue? Could you add a test please?

newnon:
Easy to reproduce set clipping type to scissors and you will see wrong clipping.
May be i will add some test but only if i will have time.

PR #15244, opened 2016-03-14
ImageView PolyInfo Support

Almax27:
Fixed loss of PolyInfo when using ImageView/Scale9Sprite

Unsure on the need/purpose of Scale9Sprite::RenderingType; SIMPLE is unnecessary; suggest removal.

### DIFF
--- a/cocos/ui/UIImageView.cpp
+++ b/cocos/ui/UIImageView.cpp
@@ -112,7 +112,7 @@ bool ImageView::init(const std::string &imageFileName, TextureResType texType)
 void ImageView::initRenderer()
 {
     _imageRenderer = Scale9Sprite::create();
-    _imageRenderer->setRenderingType(Scale9Sprite::RenderingType::SIMPLE);
+    _imageRenderer->setScale9Enabled(false);
     
     addProtectedChild(_imageRenderer, IMAGE_RENDERER_Z, -1);
 }
@@ -186,13 +186,8 @@ void ImageView::setScale9Enabled(bool able)
         return;
     }
     
-    
     _scale9Enabled = able;
-    if (_scale9Enabled) {
-        _imageRenderer->setRenderingType(Scale9Sprite::RenderingType::SLICE);
-    }else{
-        _imageRenderer->setRenderingType(Scale9Sprite::RenderingType::SIMPLE);
-    }
+    _imageRenderer->setScale9Enabled(_scale9Enabled);
     
     if (_scale9Enabled)
     {
--- a/cocos/ui/UIScale9Sprite.cpp
+++ b/cocos/ui/UIScale9Sprite.cpp
@@ -696,7 +696,7 @@ namespace ui {
     
     void Scale9Sprite::draw(Renderer *renderer, const Mat4 &transform, uint32_t flags)
     {
-        if (_scale9Image && _scale9Enabled) {
+        if (_scale9Image) {
 #if CC_USE_CULLING
             // Don't do calculate the culling if the transform was not updated
             auto visitingCamera = Camera::getVisitingCamera();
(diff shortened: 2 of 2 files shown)


zilongshanren:
@Almax27  
When set Scale9Sprite enableScale9Sprite(false), it behaviors like a Node + A Sprite as child.
If we set Scale9Sprite enableScale9Sprite(true), we have two rendering mode here. SIMPLE vs SLICE.

In SIMPLE mode, only 2 triangles are generated and when do resizing the triangles will also be resized.
In SLICE mode, we will generate 18 triangles and image will be rendered according to slice 9 rule.

I think  the removal of SIMPLE is ImageView is not acceptable and it will break some back compatibility.

Almax27:
@zilongshanren 

The problem was that it does not behave like Node + Sprite as a child because the polygon information is never used. That's what my fix achieves. I can confirm this does not affect cocos studio files and this pull request does not remove SIMPLE.

Also there are actually 3 rendering modes. SIMPLE, which is a manual quad, SLICE, which is a 9 patch and Sprite (when enableScale9Slice is false). This pull request does not change the behaviour of SIMPLE, only fixes the expected behaviour of setting enableScale9Slice to false. 

Perhaps we should remove enableScale9Slice, and have SIMPLE behave like Node+Sprite?

zilongshanren:
@Almax27  Your commit to unify the rendering of Node+Sprite rendering mode and SIMPLE+SLICe rendering mode is OK for me.

But you remove SIMPLE of ImageView which break some sort of back compatibility...

When in SIMPL mode, if we change the content size of ImageView, the ImageView will be stretched... but not in Node+Sprite rendering mode, please check..

If I'm wrong, please point me out, thanks.

zilongshanren:
@Almax27  I'm sorry... your are right here...

zilongshanren:
@Almax27 

> Perhaps we should remove enableScale9Slice, and have SIMPLE behave like Node+Sprite?

For compatibility issue, we can't do this.

zilongshanren:
@Almax27  
Could you add a test with Polygon info? so that we could prevent some regression in the future.
Thanks.

zilongshanren:
@Almax27 
I have tested your PR, the modification of Scale9Sprite is great, but ImageView not work as expected with the previous version when setScale9Sprite(false) is called...

In the previous version, ImageView use Simple rendering mode which will scale content when size is changed, but now you make it a Node + Sprite rendering mode...

image

The UIImageViewTest_Scale9_State_Change test can't pass when you click the ImageView...

zilongshanren:
@Almax27 
The scale9sprite patch also break some tests in UI:Scale9SpriteTest... please check it in the cpp-tests project.

Almax27:
@zilongshanren 

I expect the test failure is due to the fact that the behaviour of a Node+Sprite child does not resize the sprite when Node's size changes, and my patch changes the default behaviour to use Node+Sprite instead of SIMPLE. Even setting a Sprite's content size directly won't change it's size, but that's another problem. Without changing the Sprite class we could solve this by applying a scale to the sprite to have it match the Node's size. e.g. in Scale9Sprite::setContentSize(). Scaling will preserve polygon information, whereas using setTextureRect will not.

Alternatively the ImageView could start in SIMPLE rendering mode with enableScale9Sprite(true). But this will require a manual step to setScale9Enabled(false) if polygon information is desired. Not a great situation when loading a cocosStudio file.

I honestly think we should remove either enableScale9Sprite() or SIMPLE, because the API is quite confusing. Personally I would expect SIMPLE to behave like enableScale9Sprite(false) and make use of polygon information. But I understand the need for backwards compatibility. Perhaps instead we could have SIMPLE render the sprite instead of building it's own quad? That way the API would stay the same.

Thoughts?

zilongshanren:
@Almax27 
In SIMPLE rendering, Scale9Sprite not only build its own quad, but also stretch the quad when content size is changed. So the behavior is similar to Sprite->setScale. But why there are two rendering mode for SLICE mode, just because we want to unify the behavior of setContentSize API.

The original Sprite only change the node content size, leave the rendering untouched..

I think the `setScale9Enabled` API could be removed in the future, the Node + Sprite design is only for back compatibility issue...

Almax27:
@zilongshanren 

OK, aiming to remove setScale9Enabled() in the future sounds good. But the purpose of the patch was to provide Polygon support to Scale9Slice and by extension ImageView. Should SIMPLE become a Sprite render mode with scaling to match ContentSize? 

Aside from giving ImageView two renderers; a Sprite and Scale9Sprite, I'm not sure how else we could approach this.

PR #15246, opened 2016-03-15
added CCTapRecorder extension.

dolow:
# Overview

This feature allows recording and playback gestures.
Both of them can be controlled via CCConsole, or directly from an instance of CCRecorder/CCPlayer.
The main purpose of this feature is to provide an easier way to implement UI testing.
# Features
- gesture and screen capture recording/playback
- file uploading/downloading by url
- dedicated CCConsole commands
# Usage
### Recording

**start**

```
> taprecorder record start some_save_name
```

**stop**

```
> taprecorder record finish
```
### Playback

**start**

```
> taprecorder play start some_save_name
```

**stop**

```
> taprecorder play stop
```
### Recording

**initialize**

```
Recorder::getInstance()->init();
```

**start**

```
Recorder::getInstance()->record(DurationUnit::Frame);
```

**stop and save**

```
Recorder* recorder = Recorder::getInstance()
recorder->stop();
recorder->save("some_save_name");
```
### Playback

**play**

```
Player::getInstance()->play("some_save_name");
```

Notice :
Playback have to be began at the same point with recorder initialized.
Due to base time is counted up since recorder is initialized.

**stop**

```
Player::getInstance()->stop();
```
## Records

Recorded data is stored to application's writable directory in json format.
Although only json format is available currently, other format can be applied by implementing adapter.


minggo:
@dolow Thanks. But cocos2d-x only ship external that is needed  by cocos2d-x internally.

dolow:
@minggo 
Thanks for review.
Is there any appropriate place in cocos2d-x repo for all those user contributed new feature ?

minggo:
It is a pity there is not a place for this situation. `https://github.com/cocos2d/cocos2d-x-extensions` seems to be a suitable place, but it is not active.

dolow:
@minggo 
So then, I'm gonna host this.
IMO, some service like ruby gem, cocoa pods what ever, is seems to be necessary to make community more active.
I'm grad to contribute to such project if it is launched.
Thank you.

minggo:
Yep, it is good if we can share codes easily. Because cocos2d-x distribute with C/C++ source codes, and i don't know if there is a good way to share the codes. Do you have any idea?

dolow:
@minggo 
There is no right place for C/C++ codes.
CocoaPods is close to it but this is only for Apple things.
(also codes for each platform like ObjC, java make things complicated...)

One of easiest ways is to create a developer readable/writable dictionary of third party modules such as 
Node.js's npm.
Npm reduces cost by giving authors an open place with responsibility.
It has both installing and publishing functions without hosting or maintaining any modules.

I understand it costs a lot if we try to make such system.
But also it can be began from minimum functions.

minggo:
Thanks for you advise, i will deliver the idea to other guys, i can not make the decision.

PR #15254, opened 2016-03-16
fix TextureCache memory leak on android

andylua040427a:


Obg1:
It seems the fix doesn't free "texture" if the file couldn't be loaded, and there's no auto release call for it in that case.

minggo:
@andylua040427a Could you please say more about this PR? Why it causes memory leak? Thanks.

andylua040427a:
I repeat with the 3D particle system added to the layer, and then removed from the layer "on android"
windows Will not happen

minggo:
And it is fixed by this PR?

andylua040427a:
There should be，you can try

minggo:
How did you check the memory usage on Android?

andylua040427a:
adb shell procrank

minggo:
@Obg1 Yep, i agree with you though the codes looks good.

PR #15260, opened 2016-03-16
Add optional alpha parameter to Color4B and Color4F constructors

dennacerise:
The constructors for `Color4B` and `Color4F` that take a `Color3B` parameter should also take an optional alpha parameter. This is for programmers' convenience. Since the alpha parameter defaults to full opacity, this change is backward compatible.


zilongshanren:
[ci rebuild]

dennacerise:
@zilongshanren, what do you think about this change?

steerlink:
:+1: This is very convenient.

minggo:
Thanks.

PR #15273, opened 2016-03-17
Binz coocs2dx 2.2.6

Bin-z-yy:


zilongshanren:
@yihuang  we use luajit for 32bit iOS apps.

Issue #15270, opened 2016-03-17
getRandomObject() issue in Vector


wkdtjsgur100:
My develop environment is consist of eclipse, cocos2d-x 3.10, windows 10 64bit and
my android build target is Android 6.0

I want to use getRandomObject() in cocos2d::Vector class.
It works well on win32-platform, but the same code doesn't work in android-platform.
so I track my class's member methods. And figure out rand() function cause this problem.

getRandomObject function in CCVector is:

T getRandomObject() const
    {
        if (!_data.empty())
        {
            ssize_t randIdx = rand() % _data.size();
            return *(_data.begin() + randIdx);
        }
        return nullptr;
    }

In addition, it can be caused that it would be non-complete random object.

so I recommend to change a function with using a RandomHelper class in cocos2d-x

or using a uniform_real_distribution in std namespace.

like this:

T getRandomObject() const
    {
        if (!_data.empty())
        {
            ssize_t randIdx = RandomHelper:: random_int(0,_data.size());
            return *(_data.begin() + randIdx);
        }
        return nullptr;
    }

dumganhar:
Good point, could you please send a Pull Request for this? Thanks.

wkdtjsgur100:
I sent a pull request.. please check :)

gestern:
Here is better to use a more elegant solution `random(0,_data.size());` than `RandomHelper:: random_int(0,_data.size());` Because random is:

``` cpp
template<typename T>
inline T random(T min, T max) {
    return RandomHelper::random_int<T>(min, max);
}

template<>
inline float random(float min, float max) {
    return RandomHelper::random_real(min, max);
}
...
```

PR #15279, opened 2016-03-18
getRandomObject() in Vector, Map is changed

wkdtjsgur100:
https://github.com/cocos2d/cocos2d-x/issues/15270

getRandomObject issue in Vector, Map


minggo:
[ci rebuild]

PR #15286, opened 2016-03-21
Removed unnecessary search paths.

xpol:
Since we now have (almost) single search path.
The search path for editor-support is kept as spine requires it and it uses `#include <....>`.

I also refactored the cmake script for link spidermonkey library:
1. moved `code into cocos/scripting/js-bindings/CMakeLists.txt` and
2. use `find_library` instead of `link_directories`.


xpol:
Oops!
The build breaks, the follow needs to be done:
- [x] [unzip.h](https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/blob/v3/unzip/unzip.h#L46) has depends on `#include "CCPlatformDefine.h"` PR: https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/208 and new PR: https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/209
- [x] https://github.com/cocos2d/bindings-generator/pull/221
- [x] update binding generator ini files. 
- [x] Update binding generator submodule and generate bindings in BuildBot.
- [x] Release a new v3-deps-nn.zip. and update `external/config.json` in cocos2d-x project.

@zilongshanren please help with these tasks.

zilongshanren:
@xpol  Deps is updated to v3-deps-85, please modify the `external/config.json` in this PR an also the ref of  bindings PR.

xpol:
I have merge v3 into this PR and fixes include path generated by #15285 .

xpol:
@zilongshanren 
I have revert the mothod of fixing unzip.h (cocos2d/cocos2d-x-3rd-party-libs-bin#209) please help.

PR #15290, opened 2016-03-22
std::shared_ptr<bool> -> std::shared_ptr<std::atomic<bool>>

dumganhar:
The reference count field in shared_ptr is thread safe but the value field is not. So we need to wrap a std::atomic for preventing potential thread racing.

Refer to http://en.cppreference.com/w/cpp/memory/shared_ptr/atomic

If multiple threads of execution access the same std::shared_ptr object without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur unless all such access is performed through these functions, which are overloads of the corresponding atomic access functions (std::atomic_load, std::atomic_store, etc.)
Note that the control block of a shared_ptr is thread-safe: different std::shared_ptr objects can be accessed using mutable operations, such as operator= or reset, simultaneously by multiple threads, even when these instances are copies, and share the same control block internally.

### DIFF
--- a/cocos/network/WebSocket.cpp
+++ b/cocos/network/WebSocket.cpp
@@ -313,7 +313,7 @@ WebSocket::WebSocket()
 , _wsHelper(nullptr)
 , _wsInstance(nullptr)
 , _wsContext(nullptr)
-, _isDestroyed(std::make_shared<bool>(false))
+, _isDestroyed(std::make_shared<std::atomic<bool>>(false))
 , _delegate(nullptr)
 , _SSLConnection(0)
 , _wsProtocols(nullptr)
@@ -327,7 +327,7 @@ WebSocket::WebSocket()
 
     __websocketInstances->push_back(this);
     
-    std::shared_ptr<bool> isDestroyed = _isDestroyed;
+    std::shared_ptr<std::atomic<bool>> isDestroyed = _isDestroyed;
     _resetDirectorListener = Director::getInstance()->getEventDispatcher()->addCustomEventListener(Director::EVENT_RESET, [this, isDestroyed](EventCustom*){
         if (*isDestroyed)
             return;
@@ -815,7 +815,7 @@ void WebSocket::onClientReceivedData(void* in, ssize_t len)
             frameData->push_back('\0');
         }
 
-        std::shared_ptr<bool> isDestroyed = _isDestroyed;
+        std::shared_ptr<std::atomic<bool>> isDestroyed = _isDestroyed;
         _wsHelper->sendMessageToCocosThread([this, frameData, frameSize, isBinary, isDestroyed](){
             // In UI thread
             LOGD("Notify data len %d to Cocos thread.\n", (int)frameSize);
@@ -851,7 +851,7 @@ void WebSocket::onConnectionOpened()
     _readyState = State::OPEN;
     _readStateMutex.unlock();
 
-    std::shared_ptr<bool> isDestroyed = _isDestroyed;
+    std::shared_ptr<std::atomic<bool>> isDestroyed = _isDestroyed;
     _wsHelper->sendMessageToCocosThread([this, isDestroyed](){
         if (*isDestroyed)
         {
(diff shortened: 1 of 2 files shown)


dumganhar:
[ci rebuild]

cocos/network/WebSocket.h
```diff
@@ -238,7 +239,7 @@ class CC_DLL WebSocket
 
     struct lws*         _wsInstance;
     struct lws_context* _wsContext;
-    std::shared_ptr<bool> _isDestroyed;
+    std::shared_ptr<std::atomic<bool>> _isDestroyed;
```
minggo, line 242: Why bool type need to use std::shared_ptr<>? It is a raw type.

dumganhar, line 242: Since we use multi-thread, sendMessageaToCocosThread & sendMessageToWebsocketThread are asynchronous.
the Websocket instance may be destroyed while sending messages, therefore we do something in the callback, it will lead to dangerous instance invocation. 

Fortunately, C++11 supports lambda to catch a variable, in the destructor of WebSocket, we set the `_isDestroyed` to true, if callback is triggered after WebSocket destruction, we could check `isDestroyed` to decide whether to continue.

minggo, line 242: Yep, i see.
What i mean is why we need `std::shared_ptr<std::atomic<bool>>` instead of `std::atomic<bool>`, because bool is a raw type.

dumganhar, line 242: You could refer to `WebSocket::onConnectionError` method.
If it's a raw type like `bool` or `std::atomic<bool>`, lambda will just allocate a new variable to from `_isDestroyed`, the lifecycle of the new captured variable will not be the same as `WebSocket::_isDestroyed`. In this situation, although the `_isDestroyed` is set to `true`, the lambda will still catch a `false` value.

dumganhar, line 242: Since there isn't more than twice to set the _isDestroyed, to make `std::shared_ptr<bool>` to `std::shared_ptr<std::atomic<bool>>` isn't needed yep.

But for best practice, it's better to make it as `atomic` since the variable is used in different thread. :)

minggo, line 242: > You could refer to WebSocket::onConnectionError method.
> If it's a raw type like bool or std::atomic<bool>, lambda will just allocate a new variable to from _isDestroyed, the lifecycle of the new captured variable will not be the same as WebSocket::_isDestroyed. In this situation, although the _isDestroyed is set to true, the lambda will still catch a false value.

I think it is because you pass value not reference to lambda.

dumganhar, line 242: It I pass a reference to lambda, the captured value will be destroyed before the callback is invoked.
The value of `isDestroyed` will be invalid & dangerous.

cocos/network/WebSocket.cpp
```diff
@@ -900,7 +900,7 @@ void WebSocket::onConnectionClosed()
     _readStateMutex.unlock();
 
     _wsHelper->quitWebSocketThread();
-    std::shared_ptr<bool> isDestroyed = _isDestroyed;
+    std::shared_ptr<std::atomic<bool>> isDestroyed = _isDestroyed;
     _wsHelper->sendMessageToCocosThread([this, isDestroyed](){
```
dumganhar, line 904: So I create a new local variable `isDestroyed` shared_ptr to share the internal value of `WebSocket::_isDestroyed`.
Lambda will catch `isDestroyed` and get the shared value in `WebSocket::_isDestroyed` although `WebSocket::_isDestroyed` shared_ptr itself is destroyed. But we could catch its internal value.

cocos/network/WebSocket.h
minggo, line 242: Yep, you are right. Can we just pass `_isDestroyed` directly without create a new shared_ptr<> ?

dumganhar, line 242: I think we couldn't do that.
Compiler will give you some errors:

screen shot 2016-04-07 at 14 57 40

dumganhar, line 242: you could only capture by `this`, but then it's a reference capture rather than an assignment capture.

minggo, line 242: Oh, the member variable is not in the scope.

dumganhar:
Thanks for the merge.

PR #15300, opened 2016-03-22

CocosRobot:


cocos/scripting/lua-bindings/auto/lua_cocos2dx_extension_auto.cpp
```diff
@@ -1,9 +1,8 @@
 #include "scripting/lua-bindings/auto/lua_cocos2dx_extension_auto.hpp"
-#include "cocos-ext.h"
+#include "/home/travis/build/cocos2d/cocos2d-x/extensions/cocos-ext.h"
```
zilongshanren, line 2: @xpol  please take a look at this

PR #15304, opened 2016-03-22
Allow inherit form platform fileutils

xpol:
The doc for FileUtils::setDelegate() says:

```
 * You can inherit from platform dependent implementation of FileUtils, such as FileUtilsAndroid,
 * and use this function to set delegate, then FileUtils will invoke delegate's implementation.
 * For example, your resources are encrypted, so you need to decrypt it after reading data from
 * resources, then you can implement all getXXX functions, and engine will invoke your own getXX
 * functions when reading data of resources.
```

But the constructor of  FileUtilsLinux, FileUtilsWin32 and FileUtilsWinRT are private.

So make these constructors protected which allows inherit from them.

### DIFF
--- a/cocos/platform/apple/CCFileUtils-apple.h
+++ b/cocos/platform/apple/CCFileUtils-apple.h
@@ -56,7 +56,7 @@ public:
 
     virtual ValueVector getValueVectorFromFile(const std::string& filename) override;
 #if CC_FILEUTILS_APPLE_ENABLE_OBJC
-    CC_DEPRECATED_ATTRIBUTE void setBundle(NSBundle* bundle);
+    void setBundle(NSBundle* bundle);
 #endif
 private:
     virtual bool isFileExistInternal(const std::string& filePath) const override;
--- a/cocos/platform/apple/CCFileUtils-apple.h
+++ b/cocos/platform/apple/CCFileUtils-apple.h
@@ -54,13 +54,15 @@ public:
     virtual bool writeToFile(const ValueMap& dict, const std::string& fullPath) override;
 
     virtual ValueVector getValueVectorFromFile(const std::string& filename) override;
-    void setBundle(NSBundle* bundle);
+#ifdef FILEUTILS_APPLE_ENABLE_OBJC
+    CC_DEPRECATED_ATTRIBUTE void setBundle(NSBundle* bundle);
+#endif
 private:
     virtual bool isFileExistInternal(const std::string& filePath) const override;
     virtual bool removeDirectory(const std::string& dirPath) override;
 
-    NSBundle* getBundle() const;
-    NSBundle* _bundle;
+    struct IMPL;
+    IMPL* pimpl_;
 };
 
 // end of platform group
--- a/cocos/platform/apple/CCFileUtils-apple.h
+++ b/cocos/platform/apple/CCFileUtils-apple.h
@@ -26,6 +26,7 @@
 #ifndef __CC_FILEUTILS_APPLE_H__
 #define __CC_FILEUTILS_APPLE_H__
 
+#include <memory>
 #include <string>
 #include <vector>
 
@@ -62,7 +63,7 @@ private:
     virtual bool removeDirectory(const std::string& dirPath) override;
 
     struct IMPL;
-    IMPL* pimpl_;
+    std::unique_ptr<IMPL> pimpl_;
 };
 
 // end of platform group
--- a/cocos/platform/apple/CCFileUtils-apple.h
+++ b/cocos/platform/apple/CCFileUtils-apple.h
@@ -55,7 +55,7 @@ public:
     virtual bool writeToFile(const ValueMap& dict, const std::string& fullPath) override;
 
     virtual ValueVector getValueVectorFromFile(const std::string& filename) override;
-#ifdef FILEUTILS_APPLE_ENABLE_OBJC
+#if CC_FILEUTILS_APPLE_ENABLE_OBJC
     CC_DEPRECATED_ATTRIBUTE void setBundle(NSBundle* bundle);
 #endif
 private:
--- a/cocos/platform/apple/CCFileUtils-apple.mm
+++ b/cocos/platform/apple/CCFileUtils-apple.mm
@@ -320,7 +320,7 @@ static void addObjectToNSDict(const std::string& key, const Value& value, NSMuta
 FileUtilsApple::FileUtilsApple() : pimpl_(new IMPL([NSBundle mainBundle])) {
 }
 
-#ifdef FILEUTILS_APPLE_ENABLE_OBJC
+#if CC_FILEUTILS_APPLE_ENABLE_OBJC
 void FileUtilsApple::setBundle(NSBundle* bundle) {
     pimpl_->setBundle(bundle);
 }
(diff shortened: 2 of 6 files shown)


cocos/platform/apple/CCFileUtils-apple.h
```diff
@@ -54,13 +54,10 @@ class CC_DLL FileUtilsApple : public FileUtils
     virtual bool writeToFile(const ValueMap& dict, const std::string& fullPath) override;
 
     virtual ValueVector getValueVectorFromFile(const std::string& filename) override;
-    void setBundle(NSBundle* bundle);
```
zilongshanren, line 57: @xpol  why remove `setBundle` method?

xpol, line 57: So that the CCFileUtilsApple can be inherited by C++ class.

zilongshanren, line 57: @xpol  But what about the back compatibility? We lose a public API...

xpol, line 57: My C++ code is like follow, If CCFileUtilsApple has ObjectiveC class, the follow `CryptFileUtils` can compiled:

CryptFileUtils.h:

``` C++
#if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
#include "platform/android/CCFileUtils-android.h"
typedef cocos2d::FileUtilsAndroid FlatformFileUtils;

#elif (CC_TARGET_PLATFORM == CC_PLATFORM_IOS) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
#include "platform/apple/CCFileUtils-apple.h"
typedef cocos2d::FileUtilsApple FlatformFileUtils;

#elif (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32)
#include "platform/win32/CCFileUtils-win32.h"
typedef cocos2d::FileUtilsWin32 FlatformFileUtils;

#elif (CC_TARGET_PLATFORM == CC_PLATFORM_WINRT)
#include "platform/winrt/CCFileUtilsWinRT.h"
typedef cocos2d::CCFileUtilsWinRT FlatformFileUtils;
//
// ...
#endif


class CryptFileUtils: public FlatformFileUtils {
public:
    CryptFileUtils(Crypt* crypt);
// ...
```

CryptFileUtils.cpp:

```
#include "CryptFileUtils.h"

// all code here is OK.
```

If CCFileUtilsApple has objc class, I can make `CryptFileUtils` compile by have a .mm file:

CryptFileUtils.mm:

``` Objective-C++
#include "CryptFileUtils.cpp"
```

But I can't still initialize the `CryptFileUtils` in `AppDelegate.cpp`.

## Conclusion

If the cocos2d-x intended, as the comments for `FileUtils::setDelegate()` says, to let user extend the PlatformSpec FileUtils, All platform FileUtils should be pure C++ in header files.

The current state is the design of `FileUtils::setDelegate()` requires inheritance, but the implements are:
- `FileUtilsLinux`, `FileUtilsWin32` and `FileUtilsWinRT` has private constructor prevents inheritance.
- `FileUtilsApple` has ObjectiveC class in header which prevents inheritance (by C++).

So shall we remove `FileUtils::setDelegate()` or redesign `FileUtils` or remove `FileUtilsApple:setBundle()`?

I know remove of `FileUtilsApple:setBundle()` may break compatibility, but it is the fault already done by cocos2d-x history.

xpol:
@zilongshanren 

Now I use pimpl to hide the objc member var,
and added a `FILEUTILS_APPLE_ENABLE_OBJC` switch to reenable the `FileUtilsApple::setBundle`.
Is this OK?

cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -306,18 +318,14 @@ static void addObjectToNSDict(const std::string& key, const Value& value, NSMuta
 }
 
 FileUtilsApple::FileUtilsApple() {
-    _bundle = [NSBundle mainBundle];
+    pimpl_ = new IMPL([NSBundle mainBundle]);
```
zilongshanren: you might forget to release the memory owned by pimpl_..

zilongshanren:
@xpol  I'm ok with this PR, @minggo  what do you think about this PR?

cocos/platform/apple/CCFileUtils-apple.h
```diff
@@ -54,13 +55,15 @@ class CC_DLL FileUtilsApple : public FileUtils
...
+#ifdef FILEUTILS_APPLE_ENABLE_OBJC
```
zilongshanren: @xpol  I think we must declare `FILEUTILS_APPLE_ENABLE_OBJC` some where?

xpol: `ccConfig.h` ?

xpol:
Updates by last commit ffe05eb :
- Rename macro by prefix with `CC_`.
- Use `#if` rather than `#ifndef`。
- Enable by default for better compatibility。
- define that macro in `ccConfig.h`.

minggo:
@xpol `setBundle()` is needed, please refer to #7582 for detail information. About the inheritance, how about change the parameter type from `NSBundle*` to `void*` and convert the type in `CCFileUtils-apple.mm` manually.

xpol:
I have kept the `setBundle`. 
It is separated by discussion notes. You may need to uncheck the **Show Notes** checkbox.

image

and `FILEUTILS_APPLE_ENABLE_OBJC` is defined as 1 by default.

xpol:
@minggo @uorbe001
how about `fileutils->addSearchPath("bundle:path/to/bundle")` for bundle files?

xpol:
This PR is need by my cross platform hot update and encryption solution. And thus I don't use ios specific bundle at all.

minggo:
@xpol Yep, you don't remove it, but deprecated it. If `fileutils->addSearchPath("bundle:path/to/bundle")` can work, then it is better.

xpol:
Wait...

 [As apple doc says](https://developer.apple.com/library/ios/documentation/CoreFoundation/Conceptual/CFBundles/AboutBundles/AboutBundles.html):

> A bundle is a directory with a standardized hierarchical structure that holds executable code and the resources used by that code.

So bundles are works already:

``` C++
    fileutils->addSearchPath("path/to/My.bundle/language_region.lproj/"); // for localized resources
    fileutils->addSearchPath("path/to/My.bundle/"); // for nonlocalized resources
```

I think the Apple bundle API is just helping find correct localized resources.

Apple bundle is not cross platform solution, IMHO, the above basic search path support is enough and should works well.

xpol:
@minggo 
Is this acceptable without `bundle:path/to/bundle` support?
As `CC_FILEUTILS_APPLE_ENABLE_OBJC` is enabled by default, so it is 100% compatibility by default.

define `CC_FILEUTILS_APPLE_ENABLE_OBJC` to `0` is for users, like me, who wants to inherit form platform fileutils.

minggo:
Yep, it is acceptable to me. But it is better to
- remove CC_DEPRECATED_ATTRIBUTE for setBundle

xpol:
OK, removed. 😄

PR #15315, opened 2016-03-24
Fixed Renderer::checkVisibility() to support all cameras

Almax27:
Converted design resolution 2D content to 3D AABB for frustum check to better support non default cameras.

### DIFF
--- a/cocos/2d/CCCamera.cpp
+++ b/cocos/2d/CCCamera.cpp
@@ -332,6 +332,16 @@ bool Camera::isVisibleInFrustum(const AABB* aabb) const
     return !_frustum.isOutOfFrustum(*aabb);
 }
 
+bool Camera::isVisibleInFrustum(const Rect& rect) const
+{
+    if (_frustumDirty)
+    {
+        _frustum.initFrustum(this);
+        _frustumDirty = false;
+    }
+    return !_frustum.isOutOfFrustum(rect);
+}
+
 float Camera::getDepthInView(const Mat4& transform) const
 {
     Mat4 camWorldMat = getNodeToWorldTransform();
--- a/cocos/2d/CCCamera.h
+++ b/cocos/2d/CCCamera.h
@@ -194,6 +194,11 @@ public:
     bool isVisibleInFrustum(const AABB* aabb) const;
     
     /**
+     * Is this rect visible in frustum
+     */
+    bool isVisibleInFrustum(const Rect& rect) const;
+    
+    /**
      * Get object depth towards camera
      */
     float getDepthInView(const Mat4& transform) const;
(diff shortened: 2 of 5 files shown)


zilongshanren:
@Almax27  I think the `checkVisibility` API is mainly used for culling 2D rect and for performance reasons, 3D culling is not efficient for 2D rect.  @ricardoquesada  Thoughts?

Almax27:
@zilongshanren I had originally attempted implementing a projection of the transformed rect and testing against the visible rect but found this not to work when the camera's view matrix was updated (I have animations that scale and rotate the camera). I considered also projecting the visible rect but this approach seemed the most clear and concise. I haven't performed any profiling but don't belive the AABB frustum check will be much of an overhead (if any).

Almax27:
@zilongshanren Spent some time profiling and turns out I was wrong to assume there would be no overhead. Using the cocos profiling class I find the following:

VisibilityOriginal :: avg1: 0µ, avg2: 0µ, min: 0µ, max: 951µ, total: 0.10s, nr calls: 15106000
VisibilityNew :: avg1: 0µ  avg2: 0µ, min: 0µ, max: 1533µ, total: 0.52s, nr calls: 15106000

While the max is only a 1.5x increase the total cost is a 5x increase, implying the average cost of the original implementation is lower. I'll see if I can find a more efficient solution.

Almax27:
A more efficient solution can be achieved by implementing:

```
bool Frustum::isOutOfFrustum(const Rect& rect)
{
    if (_initialized)
    {
        Vec3 point;

        int plane = _clipZ ? 6 : 4;
        float minX = rect.getMinX();
        float maxX = rect.getMaxX();
        float minY = rect.getMinY();
        float maxY = rect.getMaxY();
        for (int i = 0; i < plane; i++)
        {
            const Vec3& normal = _plane[i].getNormal();
            point.x = normal.x < 0 ? maxX : minX;
            point.y = normal.y < 0 ? maxY : minY;

            if (_plane[i].getSide(point) == PointSide::FRONT_PLANE )
                return true;
        }
    }
    return false;
}
```

VisibilityOriginal :: avg1: 0µ, avg2: 0µ, min: 0µ, max: 191µ, total: 0.03s, nr calls: 3895000
VisibilityAABB :: avg1: 0µ, avg2: 0µ, min: 0µ, max: 158µ, total: 0.15s, nr calls: 3895000
VisibilityRect :: avg1: 0µ, avg2: 0µ, min: 0µ, max: 320µ,    total: 0.06s, nr calls: 3895000

I will update the pull request

ricardoquesada:
correct. `checkVisibility()` is a fast way to check "frustrum" when in 2D space.

PR #15319, opened 2016-03-25
Refine 'CCActionXXXX'

WenhaiLin:
+Improved robustness in release mode

### DIFF
--- a/cocos/2d/CCAction.cpp
+++ b/cocos/2d/CCAction.cpp
@@ -2,7 +2,7 @@
 Copyright (c) 2008-2010 Ricardo Quesada
 Copyright (c) 2010-2012 cocos2d-x.org
 Copyright (c) 2011      Zynga Inc.
-Copyright (c) 2013-2014 Chukong Technologies Inc.
+Copyright (c) 2013-2016 Chukong Technologies Inc.
 
 http://www.cocos2d-x.org
 
@@ -114,7 +114,8 @@ Speed* Speed::create(ActionInterval* action, float speed)
 bool Speed::initWithAction(ActionInterval *action, float speed)
 {
     CCASSERT(action != nullptr, "action must not be NULL");
-    if (action == nullptr) {
+    if (action == nullptr)
+    {
         log("Speed::initWithAction error: action is nullptr!");
         return false;
     }
@@ -128,16 +129,16 @@ bool Speed::initWithAction(ActionInterval *action, float speed)
 Speed *Speed::clone() const
 {
     // no copy constructor
-    if (_innerAction) {
+    if (_innerAction)
         return Speed::create(_innerAction->clone(), _speed);
-    }
     
     return nullptr;
 }
 
 void Speed::startWithTarget(Node* target)
 {
-    if (target && _innerAction) {
+    if (target && _innerAction)
+    {
         Action::startWithTarget(target);
         _innerAction->startWithTarget(target);
     }
@@ -147,9 +148,8 @@ void Speed::startWithTarget(Node* target)
 
 void Speed::stop()
 {
-    if (_innerAction) {
+    if (_innerAction)
         _innerAction->stop();
-    }
     
     Action::stop();
 }
@@ -166,9 +166,8 @@ bool Speed::isDone() const
 
 Speed *Speed::reverse() const
 {
-    if (_innerAction) {
+    if (_innerAction)
         return Speed::create(_innerAction->reverse(), _speed);
-    }
     
     return nullptr;
 }
--- a/cocos/2d/CCAction.cpp
+++ b/cocos/2d/CCAction.cpp
@@ -114,6 +114,11 @@ Speed* Speed::create(ActionInterval* action, float speed)
 bool Speed::initWithAction(ActionInterval *action, float speed)
 {
     CCASSERT(action != nullptr, "action must not be NULL");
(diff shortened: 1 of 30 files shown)


WenhaiLin:

zilongshanren:
@WenhaiLin  :+1:

cocos/2d/CCAction.cpp
```diff
@@ -123,21 +128,29 @@ bool Speed::initWithAction(ActionInterval *action, float speed)
...
     // no copy constructor
-    auto a = new (std::nothrow) Speed();
-    a->initWithAction(_innerAction->clone(), _speed);
-    a->autorelease();
-    return a;
+    if (_innerAction) {
+        return Speed::create(_innerAction->clone(), _speed);
+    }
+    
+    return nullptr;
```
minggo, line 135: It is strange to return null for clone function. And because initWithAction() will return false when inner action is null, so the `_innerAction` can not be a null.

cocos/2d/CCAction.cpp
```diff
@@ -123,21 +128,29 @@ bool Speed::initWithAction(ActionInterval *action, float speed)
...
 }
 
 void Speed::startWithTarget(Node* target)
 {
-    Action::startWithTarget(target);
-    _innerAction->startWithTarget(target);
+    if (target && _innerAction) {
```
minggo: When can a target be null?

minggo:
@WenhaiLin It is a good pull request, but i have some questions and added comments.

cocos/2d/CCAction.cpp
WenhaiLin, line 135: In some cases,only the constructor is called[JSB].

minggo, line 135: Yep, you are right.

cocos/2d/CCAction.cpp
WenhaiLin: 😢I don't remember it in any detail. I found the case from CocosRuntime.

cocos/2d/CCActionInterval.cpp
```diff
@@ -298,6 +314,18 @@ Sequence::~Sequence(void)
...
+        log("Sequence::startWithTarget error: target is nullptr!");
+        return;
+    }
+    if (_actions[0] == nullptr || _actions[1] == nullptr) {
+        log("Sequence::startWithTarget error: _actions[0] or _actions[1] is nullptr!");
+        return;
+    }
+    if (_duration <= FLT_EPSILON) {
+        return;
+    }
```
minggo: Why need this line?

cocos/2d/CCActionInterval.cpp
```diff
@@ -391,17 +422,20 @@ Sequence* Sequence::reverse() const
...
+    delete repeat;
+    return nullptr;
 }
 
 bool Repeat::initWithAction(FiniteTimeAction *action, unsigned int times)
 {
     float d = action->getDuration() * times;
 
-    if (ActionInterval::initWithDuration(d))
+    if (action && ActionInterval::initWithDuration(d))
```
minggo, line 447: `action` is used above, so it is better to check it at first of this function.

minggo:
And some small issues:
- please add `{` in a new line
- there are some places that is not use 4 spaces

cocos/2d/CCAction.cpp
minggo: It is not a big deal, it makes codes more stronger.

cocos/2d/CCActionInterval.cpp
WenhaiLin: prevent division by 0

minggo: I have two questions:
- Can `_duration` be 0? 
- `FLT_EPSILON` is a very small value, it is used to check if two floats are equal, so i think it is not a good to use it here

stevetranby:
This update causes crash with test ActionSequence2 and a couple others that create a Sequence with one of the instant actions Show, Place, etc. `If (_duration <= FLT_EPSILON)` causes this bug because _target is never set.

Maybe you can move `ActionInterval::startWithTarget(target);` above that test, or otherwise you'll have to figure out how to fix this depending on your goal with this PR.

minggo:
@WenhaiLin Please take a look.

stevetranby:
This seems to fix the issues and moves default state to constructor.

```

Sequence::Sequence()
: _split(0)
, _last(-1)
{
    _actions[0] = nullptr;
    _actions[1] = nullptr;
}

Sequence::~Sequence()
{
    CC_SAFE_RELEASE(_actions[0]);
    CC_SAFE_RELEASE(_actions[1]);
}

void Sequence::startWithTarget(Node *target)
{
    if (target == nullptr)
    {
        log("Sequence::startWithTarget error: target is nullptr!");
        return;
    }
    if (_actions[0] == nullptr || _actions[1] == nullptr)
    {
        log("Sequence::startWithTarget error: _actions[0] or _actions[1] is nullptr!");
        return;
    }

    if(_duration > FLT_EPSILON)
        _split = _actions[0]->getDuration() / _duration;;

    ActionInterval::startWithTarget(target);
    // REMOVE: if setting in constructor is fine _last = -1;
}
```

WenhaiLin:
@stevetranby Fixed crash at https://github.com/cocos2d/cocos2d-x/pull/15458

Issue #15323, opened 2016-03-27
cocos compile -p mac fails with #include "SimpleAudioEngine.h"  


anagorko:
I needed to add

```
                "$(SRCROOT)/../cocos2d/cocos/audio/include/"
```

to HEADER_SEARCH_PATHS in proj.ios_mac/*.xcodeproj/project.pbxproj to get it working.

I see that this path is included in CMakeLists.txt in the project root directory.

zilongshanren:
@anagorko  thanks. I will fix this issue later.

xpol:
@anagorko 
Can you attach full error log please?

zilongshanren:
@anagorko 

We have cleaned up our header file search path, now all the include should relative to `$engineRoot/cocos` .

~~So please change `#include "SimpleAudioEngine.h" to`#include "audio/include/SimpleAudioEngine.h" "--``~~

I will fix the back compatibility issue.

Thanks.

zilongshanren:
@anagorko  please check the latest V3 head, it should be fixed now.

PR #15328, opened 2016-03-28
[ci skip]add audio SimpleAudioEngine.h search path back

zilongshanren:


zilongshanren:
https://github.com/cocos2d/cocos2d-x/issues/15323

PR #15342, opened 2016-03-29
Allow plist and other ValueMap files be (easy) decrypted.

xpol:
The origin implementation of `FileUtilsApple::getValueMapFromFile()` reads the contents of file and parse it by calling one Apple API.
This make it harder to decrypt user encrypted plist files.

This PR use the `FileUtils::getDataFromFile()` to read file data and then use `getValueMapFromData()`
to parse it into a ValueMap. Where `FileUtils::getDataFromFile()` can be overwrite by user's decoding file utils (say `DecodingFileUtils`).

I know that user's `DecodingFileUtils` can also overwrite `getValueMapFromFile`, but this pr make the author of `DecodingFileUtils` less suffering.

### DIFF
--- a/cocos/platform/apple/CCFileUtils-apple.mm
+++ b/cocos/platform/apple/CCFileUtils-apple.mm
@@ -403,7 +403,7 @@ static int unlink_cb(const char *fpath, const struct stat *sb, int typeflag, str
     {
         log("Fail to remove: %s ",fpath);
     }
-    
+
     return ret;
 }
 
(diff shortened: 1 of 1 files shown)


minggo:
It looks great.
Thanks.

xpol:
@minggo 
Thank you!
Does next milestone means v3.11?

minggo:
No, it is the version after v3.11. Because we will release v3.11 this month, we have tested a lot. And we will not bring any potential error into v3.11. This PR looks good, but as you know, any modification can bring bugs.

xpol:
Thank you for your information!

minggo:
I have created v3.11, so merge it into master.

PR #15347, opened 2016-03-30
Add missing virtual destructor in ILocalizationManager

mogemimi:
Hello, I get the following warning when I compile with Xcode 7.3:

```
cocos/editor-support/cocostudio/LocalizationManager.h:13:11: 'cocostudio::ILocalizationManager' has virtual functions but non-virtual destructor
```

This pull request adds missing virtual destructor in ILocalizationManager, and fixes it. Thanks!


minggo:
Thanks.

Issue #15354, opened 2016-03-30
ccui::TextInput changes password style symbol


1scaR1:
If ui TextField is in focus then password style text is like '*****', but after loosing focus it changes to '● ● ● ● ●'.

Works with HEAD of v3.

zilongshanren:
@1scaR1  Got it :+1:  I will fix it later, thanks

zilongshanren:
@1scaR1  I have fixed it in v3 HEAD, please give it a try.

1scaR1:
@zilongshanren now all is ok. Thanks!

PR #15358, opened 2016-04-01
fix textField password bug

zilongshanren:


zilongshanren:
https://github.com/cocos2d/cocos2d-x/issues/15354

PR #15361, opened 2016-04-02
Iterator erase pattern

stevetranby:
Should update the iterator with return value instead of incrementing after erase. This matches the rest of the code base that already has this erase iterator inside loop pattern.


minggo:
Thanks.

PR #15364, opened 2016-04-04
proposed new template idea

slackmoehrle:
This is a proposed idea for a new template. The goal was to show the users something more interesting than the current template as well as demonstrate a better coding practice (i.e not needing to subclass Layer).

This is not ready for merge. I would still need to 
- finish for the other platforms and test.
- clean up AppDelegate.

Please review the idea and check out the way I coded it. I wanted to give good comments and demonstrate practices like using Node, Sprite, std::vector as a way to store objects, Physics, Actions, Sequences, clone(), etc.


templates/cpp-template-default/Classes/AppDelegate.cpp
```diff
@@ -48,37 +51,37 @@ bool AppDelegate::applicationDidFinishLaunching() {
...
-        director->setContentScaleFactor(MIN(smallResolutionSize.height/designResolutionSize.height, smallResolutionSize.width/designResolutionSize.width));
+        //director->setContentScaleFactor(MIN(smallResolutionSize.height/designResolutionSize.height, smallResolutionSize.width/designResolutionSize.width));
     }
 
     register_all_packages();
 
     // create a scene. it's an autorelease object
-    auto scene = HelloWorld::createScene();
-
+    Intro* intro = new Intro();
```
ricardoquesada, line 81: `Intro` is not an `autorelease` object. It is just a pure C++ object. In fact, `Intro` is being leaked.

templates/cpp-template-default/Classes/Intro.cpp
```diff
@@ -0,0 +1,278 @@
...
+    auto menuNode = Node::create();
+    menuNode->setName("menunode1");
+    
+    auto menuItem1 = MenuItemFont::create("Start Game");
+    menuItem1->setFontNameObj("fonts/Marker Felt.ttf");
+    menuItem1->setFontSizeObj(64);
+    menuItem1->setTag(99);
+    
+    menuItem1->setCallback([&](Ref *sender) {
+        Level1* level1 = new Level1();
```
ricardoquesada, line 192: same here. `Level1` is being leaked.

ricardoquesada:
+1 for improving our templates. they are outdated.

But I'm not convinced that the memory model chosen for this template is the correct one. Memory is being leaked... but even if the leaks are fixed, I wouldn't use the current memory model.

If the idea is to remove the logic from nodes, then that's Ok. But it should be done in a way that the controllers are somehow connected to the life-cycle of the "models".

templates/cpp-template-default/Classes/AppDelegate.cpp
slackmoehrle, line 81: Can you explain how you know it leaks and a proper fix?

ricardoquesada, line 81: you are `new`ing..., but not `delete`ing., and nobody is holds a reference to `intro`.
The fix is to `delete` after you use it... but if you do that, then it looks kind of weird... why are you creating an `Intro` instance if you don't use the instance?... looks strange.

ricardoquesada, line 81: another super quick fix is that instead of using `new` you just use `Intro` and `Level` as stack variables... but I don't know if you are still using that instance...

slackmoehrle, line 81: Thanks I will refine this more.

stevetranby:
I think a new template (or two) that is much more advanced would help out a lot. While keeping the barebones version as an option.

Could include similar to the ideas in this PR:
- multi-resolution example (say 1x,2x,3x or similar).
- creator/studio menu loaded scene
- code-based menu scene
- scene transitions
- using actions, scale9, layout, etc
- no mention or use of Layer :)
- could be a sample game (breakouts with over-the-top "animations" or similar)

cocos new -> advanced
cocos new --template=basic -> original (or keyword "barebones", "bare", etc)

Anyway, I think this would help reduce the "simple" questions over and over regarding some aspects of using the engine?

PR #15369, opened 2016-04-05
Make selectedSprite optional in JSB too

ntrrgc:
Requested by @pandamicro at https://github.com/cocos2d/cocos2d-html5/pull/3252.

I have not tried this as I don't work currently with JSB. Please review it.


pandamicro:
Thanks I will test it

PR #15375, opened 2016-04-06
Fix memory leak in DataReaderHelper

mogemimi:
Hello, I found a memory leak in DataReaderHelper. The following code leaks memory:

``` cpp
if (movementBoneData->frameList.size() > 0)
{
    FrameData *frameData = new (std::nothrow) FrameData();       // The first 'frameData' is never used
    frameData = movementBoneData->frameList.at(framesizemusone); // it causes a memory leak right here
    movementBoneData->addFrameData(frameData);
    frameData->release();
    frameData->frameID = movementBoneData->duration;
}
```

This pull request removes unnecessary `new` call and fixes it.
Best wishes!


minggo:
Thanks.

PR #15376, opened 2016-04-06
Fix compileTimeDefines not end ;

perminovVS:
Fix for #15204


ricardoquesada:
Thanks! Also fixed in #15377

PR #15377, opened 2016-04-06
Fix #15204 bis

ricardoquesada:
fix for github #15204


perminovVS:
@ricardoquesada review it #15376

ricardoquesada:
@perminovVS yes, thanks. both of us were working on the same fix. thanks.

PR #15383, opened 2016-04-07
Fix implicit conversion issues with Xcode 7.3

mogemimi:
Hi, I get the following errors when building `cpp-tests Mac` target on Xcode 7.3 (since commit https://github.com/cocos2d/cocos2d-x/commit/107cef06672279d6332b226ad2f4b24e132ea48e):

```
cocos/base/CCVector.h:291:75: Implicit conversion loses integer precision: 'unsigned long' to 'int'
cocos/base/CCMap.h:349:76: Implicit conversion loses integer precision: 'unsigned long' to 'int'
```

This pull request fixes the implicit conversion issues.

### DIFF
--- a/cocos/base/CCMap.h
+++ b/cocos/base/CCMap.h
@@ -346,7 +346,7 @@ public:
     {
         if (!_data.empty())
         {
-            ssize_t randIdx = RandomHelper::random_int<int>(0, _data.size()-1);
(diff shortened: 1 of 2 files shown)


minggo:
@mogemimi Thanks.

mogemimi:
@minggo Thank you for the quick review! :smiley:

Issue #15338, opened 2016-03-29
Chipmunk demo issue


minggo:
Steps to reproduce:
- run cpp-tests -> Node::Physics -> enter `3:PhysicsComponentTest`
- touch the screen to add some sprites
- continue clicking `Toggle debug` and touching screen to add sprites

result:
- sometimes sprites disappeared after clicking `Toggle debug`

test device：`iPad4 iOS 9.0.1`
PS: Mac works ok.

minggo:
@chengshan1Please add more detail information if i missed something.

ricardoquesada:
@minggo I cannot reproduce it
My environment is:
- Xcode 7.3
- iPhone 5S
- iOS 8.3
- cocos2d-x from v3 (last commit 9762a2616e75376cbaa475c6b444a28fcbc4d44f)

minggo:
@ricardoquesada Thanks. I may be hard to reproduce.

ricardoquesada:
Anyway, I had many chipmunk bugs when compiling it with Xcode 7.2. I had to compile the chipmunk libraries with Xcode 6.4. Not sure if someone re-generated the chipmunk libraries with Xcode 7.2.

minggo:
@zilongshanren Is anyone re-generated chipmunk libraries with Xcode 7.2? 
@ricardoquesada Does it mean the library is Xcode dependent?

ricardoquesada:
@minggo i had to compile it with xcode 6.4. but cocos2x can be compiled with any version.
according to this https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/commits/v3/chipmunk/prebuilt/ios i'm the last one to commit the chipmunk library so it is the good one (compiled with 6.4)

minggo:
@ricardoquesada My Xcode is 7.2 too, and i can run cpp-tests.

minggo:
Oh i upgrade it to 7.3 the day before yesterday. And can run cpp-tests.

minggo:
@ricardoquesada Could you please connect the corresponding pull request to this issue?
Thanks.

ricardoquesada:
done.

minggo:
Thanks.

PR #15394, opened 2016-04-08
feat: use new web sumodule with chipmunk v7.0 fixes

ricardoquesada:


ricardoquesada:
Github issue #15338

minggo:
Since jenkins doesn't handle web engine, so merge it.

PR #15396, opened 2016-04-08
remove useless warning log of stencil buffer.

zilongshanren:
1. disable GC for Native object (it is not mature...)  @pandamicro 
2. improve js tests of label


zilongshanren:
The log message `CCLOG("Stencil buffer is not enabled.");` is useless when using Stencil buffer in the first scene. Which means, when you use ui::Layout, ui::ScrollView, ui::ListView, ui::PageView or ClippingNode in the first scene, you will get the warning.. which is confusing..

Because the following method is only called after the first scene is initialized:

```
- (void) layoutSubviews
{
    [renderer_ resizeFromLayer:(CAEAGLLayer*)self.layer];
    size_ = [renderer_ backingSize];

    // Issue #914 #924
//     Director *director = [Director sharedDirector];
//     [director reshapeProjection:size_];
    cocos2d::Size size;
    size.width = size_.width;
    size.height = size_.height;
    //cocos2d::Director::getInstance()->reshapeProjection(size);

    // Avoid flicker. Issue #350
    //[director performSelectorOnMainThread:@selector(drawScene) withObject:nil waitUntilDone:YES];
    if ([NSThread isMainThread])
    {
        cocos2d::Director::getInstance()->drawScene();
    }
}

```

After the above method call , the color framebuffer and stencil buffer could be setup properly..

zilongshanren:
@minggo  Please review this PR, I think it should be included in v3.11...

minggo:
It is not a critical bug, so i don't think it needs to be included in v3.11.

zilongshanren:
@pandamicro  how about disabling the GC for native object in jsb?  @ricardoquesada

ricardoquesada:
If I recall correctly, the idea was to have the GC disabled in cocos2d-x, and that it will be enabled by default in Creator runtime.

zilongshanren:
@ricardoquesada  yep, but it is enabled by default on v3.11, I think it's risky and it makes a lot of js-tests not working, so I disable it by default in this PR, please check...

ricardoquesada:
looks good to me. not sure why the `boundingWitdh` of the test was also updated on this PR, but it looks safe

zilongshanren:
@ricardoquesada  the `boundingWidth` settings is no longer working, because we added new Overflow feature to the Label...  Maybe there is a back compatibility issue here, but I think the impact is small...

PR #15399, opened 2016-04-08
Make PageView page turning event time tweak configurable.

adesentenac:
This change allows user to tweak when the page turning event is raised.
Even with the existing tweak, it seemed to me that the event was raised a bit late. So instead of changing the hard-coded value, I added a way to set a different value for the tweak.


zilongshanren:
@adesentenac  Thanks, it looks good to me. 👍

PR #15401, opened 2016-04-08
Image::initWithRawData() support for more formats

ShaneSmit:
Supports RGBA8888 (as usual), plus additional support for BGRA8888, and RGB888.

### DIFF
--- a/cocos/platform/CCImage.cpp
+++ b/cocos/platform/CCImage.cpp
@@ -2155,7 +2155,7 @@ bool Image::initWithWebpData(const unsigned char * data, ssize_t dataLen)
 }
 
 
-bool Image::initWithRawData(const unsigned char * data, ssize_t dataLen, int width, int height, int bitsPerComponent, bool preMulti)
+bool Image::initWithRawData(const unsigned char * data, ssize_t dataLen, int width, int height, int bitsPerComponent, bool preMulti, Texture2D::PixelFormat renderFormat)
 {
     bool ret = false;
     do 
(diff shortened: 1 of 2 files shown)


perminovVS:
we should not use default arguments in instance methods:
https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#default-arguments

Add new version it method.
Old method called new.

ShaneSmit:
@perminovVS: Done!

cocos/platform/CCImage.cpp
```diff
@@ -2165,10 +2170,10 @@ bool Image::initWithRawData(const unsigned char * data, ssize_t dataLen, int wid
         _height   = height;
         _width    = width;
         _hasPremultipliedAlpha = preMulti;
-        _renderFormat = Texture2D::PixelFormat::RGBA8888;
+        _renderFormat = renderFormat;
 
-        // only RGBA8888 supported
-        int bytesPerComponent = 4;
+        // only RGBA8888, BGRA8888 and RGB888 supported
+        int bytesPerComponent = (_renderFormat == Texture2D::PixelFormat::RGB888) ? 3 : 4;
```
minggo, line 2176: All other formats use 4 bytes?

ShaneSmit, line 2176: This function has _never_ had any protection from people passing in the wrong data.  And I am not attempting to add any.  Only to support some of the more common "raw" formats.

minggo, line 2176: Yep, i know. What will happen if the passing format is RGB565, then `bytesPerComponent` will be 4.

ShaneSmit, line 2176: If you'd like to give me a list of all the `pixelFormats` you'd like to support in this PR, I'd be happy to add them all in.

Another option is to add in the protection that this function lacked.  We could simply throw an error if the pixel format is not supported.

Or we could do both. :smile:   Your call.

minggo, line 2176: @ShaneSmit I'd like this pull request, i am just afraid the compatibility. This function is a basic feature, it will bring big effect if it breaks compatibility. And it seems it breaks compatibility in logic.

ShaneSmit, line 2176: @minggo Since you are concerned about compatibility would you like me to make an entirely new function?  Then I can leave the old function perfectly intact.

minggo, line 2176: @ShaneSmit That may be better.

minggo:
close it since it is not responsible

PR #15344, opened 2016-03-29
Withespaces does matter

xpol:
Atom, my favorite text editor, it trims white spaces at end of line and make one and only empty ending line.
Each time I make a pull request, it will contains a lots of changes to tailing spaces and ending lines.
It is not nice for a simple PR, like [this](https://github.com/cocos2d/cocos2d-x/pull/15343/files), contains a lot of white space changes.

So I write a coding style tool `tools/coding-style/tailing-spaces.py` which clean the tailing whitespaces and ending lines once for all.

in this tool:
- it remove tailing white spaces
- and ensures one and only one empty ending line.
- it fixes '.c', '.cpp', '.h', '.hpp', '.m', '.mm', '.java' sources in
  - $engineRoot/cocos
  - $engineRoot/extensions
  - $engineRoot/templates
  - $engineRoot/tests
  - $engineRoot/tools/simulator

I think this tool is good to add to @CocosRobot when updating bindings.

PR #15435, opened 2016-04-16
Remove include of cocos2d.h

xpol:
The benefits are:
- Avoid mass rebuild when just changed one header file.
- Boost build speed.


xpol:
1st commit, let ci check if its OK.

xpol:
This PR also introduced some remove of trailing spaces.
Please consider merge  #15344 and apply that script when update bindings in CI.

xpol:
[ci rebuild]

xpol:
@zilongshanren 
Would please have a look at this PR.

cocos/editor-support/spine/SkeletonAnimation.h
```diff
@@ -34,7 +34,7 @@
 
 #include <spine/spine.h>
 #include <spine/SkeletonRenderer.h>
-#include "cocos2d.h"
```
minggo, line 37: According spine's license, we can not modify the runtime. It is better if you can send a pull request to https://github.com/EsotericSoftware/spine-runtimes.

cocos/deprecated/CCDeprecated.h
```diff
@@ -32,8 +32,59 @@
 #define __COCOS2D_CCDEPRECATED_H__
 
 #include <math.h>
-#include "math/CCGeometry.h"
+#include "deprecated/CCSet.h"
```
zilongshanren: why add so many include in CCDeprecated.h?

xpol: `CCDeprecated.h` uses lots of cocos class. Just include `CCDeprecated.h` is not enough if I don't add these includes.
Another solution is remove use of  `CCDeprecated.h` in side cocos engine (actually just 3 cpp files in lua binding use it).

cocos/editor-support/spine/SkeletonAnimation.h
xpol, line 37: OK, I'll revert the changes to spine here. and try to make a PR at https://github.com/EsotericSoftware/spine-runtimes.

cocos/deprecated/CCDeprecated.h
xpol: Oops, It can't be done.
`CCDeprecated.h` needs lots of headers to be included.
And it uses the depended classes directly in `CCDeprecated.h`.

I think it is harmless to add required headers in `CCDeprecated.h`.
So when user wants to enable deprecated features, just include `CCDeprecated.h`, rather than, as `CCDeprecated.h` was required, include the whole `cocos2d.h`.

Include of `cocos2d.h` will cause lots of rebuild when touched any of cocos headers.

cocos/editor-support/spine/SkeletonAnimation.h
minggo, line 37: Thanks. 
It is a good PR.
I will merge it after you revert it.

minggo:
Thanks.

xpol:
@minggo Thank you!

PR #15443, opened 2016-04-18
Fixes some warnings in Lua manual bindings.

xpol:
1. Use __Type for deprecated classes.
2. Refactored LuaStack::luaLoadChunksFromZIP() and use FileUtils:getDataFromFile().
3. Fixes wrongly compare NULL with char array.


minggo:
@xpol Could you please fix conflict?

xpol:
@minggo 
Fixed.

minggo:
Thanks, i will merge it when Jenkins passes.

cocos/scripting/lua-bindings/manual/CCLuaEngine.cpp
```diff
@@ -439,13 +439,13 @@ int LuaEngine::handleCommonEvent(void* data)
         return 0;
    
     CommonScriptData* commonInfo = static_cast<CommonScriptData*>(data);
-    if (NULL == commonInfo->eventName || 0 == commonInfo->handler)
+    if ('\0' == commonInfo->eventName[0] || 0 == commonInfo->handler)
```
dumganhar, line 442: @xpol , why to change this line ?
Before:
If eventName is empty string, this line will not return directly.
Now:
If eventName is empty string, this line will return directly.

It broke compatibility.

dumganhar, line 442: What's the warning of this?
XMLHttpRequest & WebSocketTest in lua test will not get response after this change.

xpol, line 442: See
https://github.com/cocos2d/cocos2d-x/blob/c9306a053f051325a03b5297be7be6d645584780/cocos/base/CCScriptSupport.h#L586

CommonScriptData::eventName is array of char not pointer to char. So NULL
== commonInfo->eventName will always eval to true.

This cimmit just do the intended check.
Its noting wrong about this fix, it just disclosure a bug hidden in our
code.

James Chen notifications@github.com于2016年6月22日周三 下午5:28写道：

xpol, line 442: Sorry its line 534:

https://github.com/cocos2d/cocos2d-x/blob/c9306a053f051325a03b5297be7be6d645584780/cocos/base/CCScriptSupport.h#L534

Xpol Wan xpolife@gmail.com于2016年6月22日周三 下午10:50写道：

dumganhar, line 442: Yep, `eventName` doesn't need to be checked since it's always non-null.
But your change affects the original logic.
The original logic allows to pass in an empty string("") while current logic doesn't allow to pass in an empty string.
Please refer my Pull Request: https://github.com/cocos2d/cocos2d-x/pull/15922/files

PR #15450, opened 2016-04-18
Replace cc_utf8_to_utf16 with UTF8ToUTF16 to avoid deprecation warning

mogemimi:
This pull request fixes the following warning when building JS Bindings with Xcode 7.3:

```
cocos/scripting/js-bindings/manual/js_manual_conversions.cpp:2079:39: 'cc_utf8_to_utf16' is deprecated
```


minggo:
@mogemimi Could you please fix the compiling error?

mogemimi:
@minggo Thank you for reviewing!

I've just pushed a fix for Platform Toolset v120 within Visual Studio. Could you please review it again?

minggo:
Thanks.

PR #15452, opened 2016-04-18
Implement createDirectory for apple platform.

### LINKED ISSUE #15073: FileUtils::createDirectory fails on OS X app with sandbox
If you want to release application to Mac App Store, you must turn on Sanbox entitlement.
But **createDirectory** function fails to work with it, because function **mkdir** returns _-1_, not directory exists flag, for folders, that are not in Sandbox.( if pass full path to directory in function createDirectory).

``` c++
for (int i = 0; i < dirs.size(); ++i)
    {
        subpath += dirs[i];
        dir = opendir(subpath.c_str());

        if (!dir)
        {
            // directory doesn't exist, should create a new one

            int ret = mkdir(subpath.c_str(), S_IRWXU | S_IRWXG | S_IRWXO);
            if (ret != 0 && (errno != EEXIST))
            {
                // current directory can not be created, sub directories can not be created too
                // should return
                return false; // <------ Here function exits on '/' folder if path if absolute
            }
        }
        else
        {
            // directory exists, should close opened dir
            closedir(dir);
        }
    }
```

1scaR1:
Fixes #15073.
I think using of file manger for creating directories is more suitable for apple platforms.

### DIFF
--- a/cocos/platform/apple/CCFileUtils-apple.mm
+++ b/cocos/platform/apple/CCFileUtils-apple.mm
@@ -534,4 +534,16 @@ ValueVector FileUtilsApple::getValueVectorFromFile(const std::string& filename)
     return ret;
 }
 
+bool FileUtilsApple::createDirectory(const std::string& path)
+{
+    CCASSERT(!path.empty(), "Invalid path");
+    
+    if (isDirectoryExist(path))
+        return true;
+    
+    NSError* error;
+    
+    return [s_fileManager createDirectoryAtPath:[NSString stringWithUTF8String:path.c_str()] withIntermediateDirectories:YES attributes:nil error:&error];
+}
+
 NS_CC_END
--- a/cocos/platform/apple/CCFileUtils-apple.mm
+++ b/cocos/platform/apple/CCFileUtils-apple.mm
@@ -543,7 +543,14 @@ bool FileUtilsApple::createDirectory(const std::string& path)
     
     NSError* error;
     
-    return [s_fileManager createDirectoryAtPath:[NSString stringWithUTF8String:path.c_str()] withIntermediateDirectories:YES attributes:nil error:&error];
+    bool result = [s_fileManager createDirectoryAtPath:[NSString stringWithUTF8String:path.c_str()] withIntermediateDirectories:YES attributes:nil error:&error];
+    
+    if(!result && error != nil)
+    {
+        CCLOGERROR("Fail to create directory \"%s\": %s", path.c_str(), [error.localizedDescription UTF8String]);
+    }
+    
+    return result;
 }
 
 NS_CC_END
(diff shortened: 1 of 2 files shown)


cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -534,4 +534,16 @@ static int unlink_cb(const char *fpath, const struct stat *sb, int typeflag, str
...
+{
+    CCASSERT(!path.empty(), "Invalid path");
+    
+    if (isDirectoryExist(path))
+        return true;
+    
+    NSError* error;
+    
+    return [s_fileManager createDirectoryAtPath:[NSString stringWithUTF8String:path.c_str()] withIntermediateDirectories:YES attributes:nil error:&error];
+}
```
minggo, line 554: I think we should remove error if it is not used, or we should log the error message.

1scaR1:
@minggo Added error logging.
Testing can be simple: turn on `Sandbox entitlement` ant try to create directory such as `/abc`

minggo:
@1scaR1 I applied your PR, and it also can not create a directory. My code is:

``` c++
bool result = FileUtils::getInstance()->createDirectory("/abc");
if (result)
    CCLOG("true");
```

1scaR1:
And it must not. In sandbox you can create dirs only in writable path. 
But if there is no sandbox you do not have rights in root dir.
This fix allows to create dir such as `getWritablePath()+"abc"` or `fullPathForXXX("dir")` with sandbox enabled.

minggo:
@1scaR1 Without this PR, i can create directory too, my codes changed to

``` c++
auto fileUtils = FileUtils::getInstance();
    bool result = fileUtils->createDirectory(fileUtils->getWritablePath() + "/abc");
    if (result)
        CCLOG("true");
```

And i already enable sandbox.

minggo:
The configuration is:

screen shot 2016-04-19 at 5 36 11 pm

minggo:
I tested in simulator, should i test in devices?

1scaR1:
@minggo I have your code(without this PR) above not logging `true`(with sandbox enabled, **directory must not exists**).
Also i printed writable path when directory not creates, ant it's like:
`/Users/_USER_/Library/Containers/_APP_ID_/Data/Documents/`

It not works only on OS X, why there:

> I tested in simulator, should i test in devices?

?

1scaR1:
XCode 7.3, OS X 10.11.4(and same effect on 10.11.5 beta 1)

minggo:
> I tested in simulator, should i test in devices?

Oh, yep. It's my fault. I tested on Mac. I will double check it.

minggo:
@1scaR1 You are right. I think i didn't remove `abc` folder when testing without this PR.
Thanks.

PS: You publish games to Mac Store?

1scaR1:
@minggo , yes I published my game)

minggo:
Congratulations.

1scaR1:
Thanks)

PR #15458, opened 2016-04-19
Fixed crash and actions may not have been executed.

WenhaiLin:


WenhaiLin:
@minggo Please review this PR.

cocos/2d/CCActionInterval.cpp
```diff
@@ -327,11 +328,10 @@ void Sequence::startWithTarget(Node *target)
         log("Sequence::startWithTarget error: _actions[0] or _actions[1] is nullptr!");
         return;
     }
-    if (_duration <= FLT_EPSILON)
-        return;
+    if (_duration > FLT_EPSILON)
+        _split = _actions[0]->getDuration() / _duration;
```
minggo, line 332: What will happen if _duration <= FLT_EPSILON?

stevetranby, line 332: Looks like you merged anyway, but to include an answer here _duration == FLT_EPSILON if it was 0 during initialization.

minggo, line 332: Yep, @WenhaiLin describe it to me in other way. If _duration == FLT_EPSILON, then we should not do `_split = _actions[0]->getDuration() / _duration` since it will be an error.

@stevetranby 
Is there any problem?

Issue #15423, opened 2016-04-14
Node:UI->GUI Dynamic Create Test->19:Rich Text Test crashed on windows


minggo:
OS: win7
compiler: VS2015

![screenshot_1_1459243006](https://cloud.githubusercontent.com/assets/488203/14516532/89fccab6-0236-11e6-9eca-b41058890694.png)

minggo:
@ricardoquesada Could you please take a look?

stammen:
@ricardoquesada  The crash also happens in Win10 and Windows 8.1 builds. The crash is at line 714 of UIRichText.cpp

static bool isWrappable(const std::string& text)
{
    for (int i=0; i<text.length(); ++i)
    {
        if (!std::isalnum(text[i])) // CRASH IS HERE
            return true;
    }
    return false;
}

text is "ä¸­å›½ä¸­å›½ä¸­å›½ä¸­å›½ä¸­å›½ä¸­å›½ä¸­å›½ä¸­å›½ä¸­å›½ä¸­å›½"

summerinsects:
`std::isalnum((int)(unsigned char)text[i])` may be ok
other `std::isXXX`s are likewise

stammen:
@wohaaitinciu @ricardoquesada  The suggestion by wohaaitinciu works but moves the crash to line 851 

```
    if (std::isspace(text[rightStart]))
```

We need a review the use of all isXXX() code in cocos2d-x for correct usage and syntax.

ricardoquesada:
sent PR both to v3.11 and v3

minggo:
@ricardoquesada 
- The PR for v3 has compiling error
- The PR for v3.11 is contains many unneeded commits

Could you please fix them?

ricardoquesada:
@minggo what is the easiest way to send a PR to v3.11 if the PR was based on the v3 branch?

minggo:
I think you can use [git cherry-pick](https://git-scm.com/docs/git-cherry-pick).

ricardoquesada:
@minggo I mean... there should be a better workflow. If each PR should go both to v3 and and v3.11, there should be a better way than doing cherry pick... just as guess.

what have you been doing? cherry-pick? rebase? something else? thanks.

minggo:
Yep, i use cherry-pick. And i only merge critical commits into v3.11 since it is release candidate. And after release v3.11, i will merge it back to v3 and remove the branch.

PR #15461, opened 2016-04-19
fix: fix crash in win for rich text

ricardoquesada:
use `std::locale()` when invoking `std::isXXX` function
to avoid crash

github issue #15423


minggo:
@ricardoquesada Could you please fix compiling errors?

ricardoquesada:
The error seems to be this one:
"error: branch 'pull15462' not found"... which means that it is a CI error

ricardoquesada:
[ci rebuild]

PR #15462, opened 2016-04-19
Issue 15423

ricardoquesada:
fix issue #15423


minggo:
@ricardoquesada I think you should fix it based on v3.11 branch. This PR includes so many commits that are not needed.

PR #15466, opened 2016-04-20
Copy polygonInfo when cloning SpriteFrame

wefiends:
Copy the `_polygonInfo` over in SpriteFrame's `clone()` function. Otherwise, for example, calling `reverse()` on an Animate action will incorrectly draw the the node that it is applied to.


minggo:
Thanks.

Issue #15457, opened 2016-04-19
[FileUtils] memory leak when read emty file.


xpol:
in `cocos/platform/CCFileUtils.cpp`:

``` C++
static Data getData(const std::string& filename, bool forString)
{
// ...
        if (forString)
        {
            buffer = (unsigned char*)malloc(sizeof(unsigned char) * (size + 1));
            buffer[size] = '\0';
        }
        else
        {
            buffer = (unsigned char*)malloc(sizeof(unsigned char) * size);
        }

        readsize = fread(buffer, sizeof(unsigned char), size, fp);

// ...
    } while (0);

    if (nullptr == buffer || 0 == readsize)
    {
        CCLOG("Get data from file %s failed", filename.c_str());
    }
    else
    {
        ret.fastSet(buffer, readsize);
    }

    return ret;
}
```

The buffer only assigned to ret when not `(nullptr == buffer || 0 == readsize)`.
That is when malloc buffer succeed, and the file has zero length, the `buffer` leaks.

Issue #15456, opened 2016-04-19
[FileUtils] Should not support text mode.


xpol:
We are current using `"rt"` to open files:
1. https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/CCFileUtils.cpp#L627
2. https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/android/CCFileUtils-android.cpp#L297

For 1, it works only on Windows.
For 2, it is already android, `"rt"` has not effect here.

And as cocos2d-x it is a cross platform game engine, what's the point of supporting `rt` on Windows?

Eg, for a user works on windows developing mobile game for android. The resource files has text with CRLF.
It may works on windows. When user release the game to android (and iOS), the code may not work with the resources any more...

PR #15479, opened 2016-04-21
Add FileUtils::getContents().

xpol:
Why another file read API?

Current APIs has some problem:
1. `FileUtils::getDataFormFile()` can't return string, convert from Data to string will make extra copy.
2. `FileUtils::getStringFormFile()` can't read binary file, and it truncate contents.
3. `FileUtils::getDataFormFile()` and `FileUtils::getStringFormFile()` does not tell different from empty file and io errors.

Advantage of `FileUtils::getContents()`
1. Support read into `cocos2d::Data` `std::basic_string` and `std::vector`
2. It has abstract of the output buffer, custom type of buffer object can be adapted by writing subclass `ResizableBuffer`.
3. It returns error code, so that we can know the different form empty file and io error.

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -610,79 +610,58 @@ void FileUtils::purgeCachedEntries()
     _fullPathCache.clear();
 }
 
-static Data getData(const std::string& filename, bool forString)
+
+// TODO: move this function to StringUtils.
+// http://stackoverflow.com/a/7724536/87021
+static inline std::string replaceAll(
+    std::string const& original,
+    std::string const& before,
+    std::string const& after
+)
 {
-    if (filename.empty())
-    {
-        return Data::Null;
+    std::string retval;
+    retval.reserve(original.size());
+    auto end = original.end();
+    auto current = original.begin();
+    auto next = std::search( current, end, before.begin(), before.end() );
+    while ( next != end ) {
+        retval.append( current, next );
+        retval.append( after );
+        current = next + before.size();
+        next = std::search( current, end, before.begin(), before.end() );
     }
-
-    Data ret;
-    unsigned char* buffer = nullptr;
-    size_t size = 0;
-    size_t readsize;
-    const char* mode = nullptr;
-
-    if (forString)
-        mode = "rt";
-    else
-        mode = "rb";
-
-    auto fileutils = FileUtils::getInstance();
-    do
-    {
-        // Read the file from hardware
-        std::string fullPath = fileutils->fullPathForFilename(filename);
-        FILE *fp = fopen(fileutils->getSuitableFOpen(fullPath).c_str(), mode);
-        CC_BREAK_IF(!fp);
-        fseek(fp,0,SEEK_END);
-        size = ftell(fp);
-        fseek(fp,0,SEEK_SET);
-
-        if (forString)
-        {
-            buffer = (unsigned char*)malloc(sizeof(unsigned char) * (size + 1));
-            buffer[size] = '\0';
-        }
-        else
-        {
-            buffer = (unsigned char*)malloc(sizeof(unsigned char) * size);
-        }
-
-        readsize = fread(buffer, sizeof(unsigned char), size, fp);
-        fclose(fp);
-
-        if (forString && readsize < size)
-        {
-            buffer[readsize] = '\0';
-        }
-    } while (0);
-
-    if (nullptr == buffer || 0 == readsize)
-    {
-        CCLOG("Get data from file %s failed", filename.c_str());
-    }
-    else
-    {
-        ret.fastSet(buffer, readsize);
-    }
-
-    return ret;
+    retval.append( current, next );
+    return retval;
 }
+
+static inline void textFormCRLF(std::string& s) {
+    static const std::string CRLF("\r\n");
+    static const std::string LF("\n");
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WIN32 // this is what getStringFromFile() does before...
+    s = replaceAll(s, CRLF, LF, result);
+#endif
+}
+
 
 std::string FileUtils::getStringFromFile(const std::string& filename)
 {
-    Data data = getData(filename, true);
-    if (data.isNull())
-        return "";
+    std::string s;
+    getContents(filename, &s);
 
-    std::string ret((const char*)data.getBytes());
-    return ret;
+    // truncated
+    s.resize(strlen(s.data()));
+
+    // simulate text mode, CRLF -> LF
+    textFormCRLF(s);
+
+    return s;
 }
 
 Data FileUtils::getDataFromFile(const std::string& filename)
 {
-    return getData(filename, false);
+    Data d;
+    getContents(filename, &d);
+    return d;
 }
 
 
@@ -719,31 +698,18 @@ FileUtils::Error FileUtils::getContents(const std::string& filename, ResizableBu
 
 unsigned char* FileUtils::getFileData(const std::string& filename, const char* mode, ssize_t *size)
 {
-    unsigned char * buffer = nullptr;
     CCASSERT(!filename.empty() && size != nullptr && mode != nullptr, "Invalid parameters.");
-    *size = 0;
-    do
-    {
-        // read the file from hardware
-        const std::string fullPath = fullPathForFilename(filename);
-        FILE *fp = fopen(getSuitableFOpen(fullPath).c_str(), mode);
-        CC_BREAK_IF(!fp);
 
-        fseek(fp,0,SEEK_END);
-        *size = ftell(fp);
-        fseek(fp,0,SEEK_SET);
-        buffer = (unsigned char*)malloc(*size);
-        *size = fread(buffer,sizeof(unsigned char), *size,fp);
-        fclose(fp);
-    } while (0);
+    std::string s;
+    getContents(filename, &s);
 
-    if (!buffer)
-    {
-        std::string msg = "Get data from file(";
-        msg.append(filename).append(") failed!");
+    bool textmode = strchr(mode, 't') != NULL;
+    if (textmode)
+        textFormCRLF(s);
 
-        CCLOG("%s", msg.c_str());
-    }
+    unsigned char * buffer = (unsigned char*)malloc(s.size());
+    memcpy(buffer, s.data(), s.size());
+
     return buffer;
 }
 
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -610,6 +610,9 @@ void FileUtils::purgeCachedEntries()
     _fullPathCache.clear();
 }
 
+
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
+// TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
 namespace
 {
     // TODO: move this function to StringUtils.
@@ -634,15 +637,14 @@ namespace
         retval.append( current, next );
         return retval;
     }
-    
+
     static inline void textFormCRLF(std::string& s) {
         static const std::string CRLF("\r\n");
         static const std::string LF("\n");
-#if CC_TARGET_PLATFORM == CC_PLATFORM_WIN32 // this is what getStringFromFile() does before...
         s = std::move(replaceAll(s, CRLF, LF));
-#endif
     }
 }
+#endif
 
 
 std::string FileUtils::getStringFromFile(const std::string& filename)
@@ -702,9 +704,12 @@ unsigned char* FileUtils::getFileData(const std::string& filename, const char* m
     std::string s;
     getContents(filename, &s);
 
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
+    // TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
     bool textmode = strchr(mode, 't') != NULL;
     if (textmode)
         textFormCRLF(s);
+#endif
 
     unsigned char * buffer = (unsigned char*)malloc(s.size());
     memcpy(buffer, s.data(), s.size());
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -610,43 +610,6 @@ void FileUtils::purgeCachedEntries()
     _fullPathCache.clear();
 }
 
-
-#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
-// TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
-namespace
-{
-    // TODO: move this function to StringUtils.
-    // http://stackoverflow.com/a/7724536/87021
-    static inline std::string replaceAll(
-                                         std::string const& original,
-                                         std::string const& before,
-                                         std::string const& after
-                                         )
-    {
-        std::string retval;
-        retval.reserve(original.size());
-        auto end = original.end();
-        auto current = original.begin();
-        auto next = std::search( current, end, before.begin(), before.end() );
-        while ( next != end ) {
-            retval.append( current, next );
-            retval.append( after );
-            current = next + before.size();
-            next = std::search( current, end, before.begin(), before.end() );
-        }
-        retval.append( current, next );
-        return retval;
-    }
-
-    inline void textFormCRLF(std::string& s) {
-        static const std::string CRLF("\r\n");
-        static const std::string LF("\n");
-        s = std::move(replaceAll(s, CRLF, LF));
-    }
-}
-#endif
-
-
 std::string FileUtils::getStringFromFile(const std::string& filename)
 {
     std::string s;
@@ -700,18 +663,12 @@ FileUtils::Status FileUtils::getContents(const std::string& filename, ResizableB
 unsigned char* FileUtils::getFileData(const std::string& filename, const char* mode, ssize_t *size)
 {
     CCASSERT(!filename.empty() && size != nullptr && mode != nullptr, "Invalid parameters.");
+    (void)(mode); // mode is unused, as we do not support text mode any more...
 
     *size = 0;
     std::string s;
     if (getContents(filename, &s) != Status::OK)
         return nullptr;
-
-#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
-    // TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
-    bool textmode = strchr(mode, 't') != NULL;
-    if (textmode)
-        textFormCRLF(s);
-#endif
 
     unsigned char * buffer = (unsigned char*)malloc(s.size());
     if (!buffer)
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -686,20 +686,20 @@ Data FileUtils::getDataFromFile(const std::string& filename)
 }
 
 
-FileError FileUtils::getContents(const std::string& filename, ResizableBuffer* buffer)
+FileUtils::Error FileUtils::getContents(const std::string& filename, ResizableBuffer* buffer)
 {
     if (filename.empty())
-        return FileError::NotExists;
+        return Error::NotExists;
 
     auto fs = FileUtils::getInstance();
 
     std::string fullPath = fs->fullPathForFilename(filename);
     if (fullPath.empty())
-        return FileError::NotExists;
+        return Error::NotExists;
 
     FILE *fp = fopen(fs->getSuitableFOpen(fullPath).c_str(), "rb");
     if (!fp)
-        return FileError::OpenFailed;
+        return Error::OpenFailed;
 
     fseek(fp, 0, SEEK_END);
     size_t size = ftell(fp);
@@ -711,10 +711,10 @@ FileError FileUtils::getContents(const std::string& filename, ResizableBuffer* b
 
     if (readsize < size) {
         buffer->resize(readsize);
-        return FileError::ReadFaild;
+        return Error::ReadFaild;
     }
 
-    return FileError::OK;
+    return Error::OK;
 }
 
 unsigned char* FileUtils::getFileData(const std::string& filename, const char* mode, ssize_t *size)
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -666,20 +666,20 @@ Data FileUtils::getDataFromFile(const std::string& filename)
 }
 
 
-FileUtils::Error FileUtils::getContents(const std::string& filename, ResizableBuffer* buffer)
+FileUtils::Status FileUtils::getContents(const std::string& filename, ResizableBuffer* buffer)
 {
     if (filename.empty())
-        return Error::NotExists;
+        return Status::NotExists;
 
     auto fs = FileUtils::getInstance();
 
     std::string fullPath = fs->fullPathForFilename(filename);
     if (fullPath.empty())
-        return Error::NotExists;
+        return Status::NotExists;
 
     FILE *fp = fopen(fs->getSuitableFOpen(fullPath).c_str(), "rb");
     if (!fp)
-        return Error::OpenFailed;
+        return Status::OpenFailed;
 
     fseek(fp, 0, SEEK_END);
     size_t size = ftell(fp);
@@ -691,10 +691,10 @@ FileUtils::Error FileUtils::getContents(const std::string& filename, ResizableBu
 
     if (readsize < size) {
         buffer->resize(readsize);
-        return Error::ReadFaild;
+        return Status::ReadFaild;
     }
 
-    return Error::OK;
+    return Status::OK;
 }
 
 unsigned char* FileUtils::getFileData(const std::string& filename, const char* mode, ssize_t *size)
@@ -703,7 +703,7 @@ unsigned char* FileUtils::getFileData(const std::string& filename, const char* m
 
     *size = 0;
     std::string s;
-    if (getContents(filename, &s) != Error::OK)
+    if (getContents(filename, &s) != Status::OK)
         return nullptr;
 
 #if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -653,9 +653,6 @@ std::string FileUtils::getStringFromFile(const std::string& filename)
     // truncated
     s.resize(strlen(s.data()));
 
-    // simulate text mode, CRLF -> LF
-    textFormCRLF(s);
-
     return s;
 }
 
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -634,8 +634,8 @@ namespace
         retval.append( current, next );
         return retval;
     }
-    
-    static inline void textFormCRLF(std::string& s) {
+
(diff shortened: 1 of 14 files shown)


cocos/platform/CCFileUtils.h
```diff
@@ -100,6 +171,49 @@ class CC_DLL FileUtils
...
+     *      - FileError::NotExists when file not exists, the buffer will not changed.
+     *      - FileError::OpenFailed when cannot open file, the buffer will not changed.
+     *      - FileError::ReadFaild when read end up before read whole, the buffer will fill with already read bytes.
+     *      - FileError::NotInitialized when FileUtils is not initializes, the buffer will not changed.
+     */
+    template <
+        typename T,
+        typename std::enable_if<
+            std::is_base_of< ResizableBuffer, ResizableBufferAdapter<T> >::value
+        >::type* = nullptr
```
minggo: I know the effect of enable_if<>, but i don't know the effect of 

``` c++
typename std::enable_if<xx>::type* = nullptr
```

What does it mean? I can not find the document for it.

xpol: http://en.cppreference.com/w/cpp/types/enable_if See Example section `#4` usage.

minggo: Yep, i have seen it, but don't understand, why can not use

``` c++
typename std::enable_if<xx>::type
```

xpol: It will cause [Substitution Failure](http://en.cppreference.com/w/cpp/language/sfinae) if `ResizableBufferAdapter<T>` is not a subcalss of `ResizableBuffer`.

I only defined `ResizableBufferAdapter< std::vector<> >` `ResizableBufferAdapter< std::string<> >` and `ResizableBufferAdapter< Data >` inherited form `ResizableBuffer`. So the template version of `getContents()` will enabled only for `std::vector<>`  `std::string<>` and `Data`.

If user defined his own version `ResizableBufferAdapter< UserType >`, that type `UserType` will also suitable for calling template version of `getContents()`.

minggo: I know the meaning, but don't understand why need to assign to nullptr. Ok, i will do more research about it.

Thanks.

minggo: I mean why can not use 

``` c++
template <
 +        typename T,
 +        typename std::enable_if<
 +            std::is_base_of< ResizableBuffer, ResizableBufferAdapter<T> >::value
 +        >
```

xpol: Did you mean:

``` C++
template <
        typename T,
        typename X = std::enable_if<
            std::is_base_of< ResizableBuffer, ResizableBufferAdapter<T> >::value
        >::type
>
```

X is optional:

``` C++
template <
        typename T,
        typename  = std::enable_if<
            std::is_base_of< ResizableBuffer, ResizableBufferAdapter<T> >::value
        >::type
>
```

minggo: Yep. I think it is easier to understand

xpol:
``` C++
template <
        typename T,
        typename X = std::enable_if<
            std::is_base_of< ResizableBuffer, ResizableBufferAdapter<T> >::value
        >::type
>
```

will not compile:

```
CCFileUtils.h:206:22: error: need 'typename' before 'std::enable_if<std::is_base_of<cocos2d::ResizableBuffer, cocos2d::ResizableBufferAdapter<T> >::value>::type' because 'std::enable_if<std::is_base_of<cocos2d::ResizableBuffer, cocos2d::ResizableBufferAdapter<T> >::value>' is a dependent scope
         typename X = std::enable_if<
                      ^

```

minggo: Got it, thanks.

minggo:
I found only Android platform implements `FileError getContents(const std::string& filename, ResizableBuffer* buffer)`, what will happen if it is invoked in other platforms?

cocos/platform/CCFileUtils.h
minggo: But the `#5` section of http://en.cppreference.com/w/cpp/types/enable_if can work.

xpol: This works:

``` C++
template <
        typename T,
        typename X = typename std::enable_if<
            std::is_base_of< ResizableBuffer, ResizableBufferAdapter<T> >::value
        >::type
>
```

xpol:
There is default: `FileError FileUtils::getContents(const std::string& filename, ResizableBuffer* buffer)` in [`CCFileUtils.cpp`](https://github.com/cocos2d/cocos2d-x/pull/15479/files#diff-cb65223cd3edf30671fefad4a1883edfR689).

cocos/platform/CCFileUtils.h
xpol: `#5` section of http://en.cppreference.com/w/cpp/types/enable_if uses `enable_if_t` (note the tailling `_t` ) is a C++14 feature.

minggo:
Sorry about my careless. I will continue viewing it tomorrow.

xpol:
@minggo 
Thank you very much.

cocos/platform/CCFileUtils.h
```diff
@@ -41,6 +42,76 @@ NS_CC_BEGIN
...
+    }
+    virtual void* buffer() const override {
+        return &buffer_->front();
+    }
+};
+
+template<typename T, typename A>
+class ResizableBufferAdapter< std::vector<T, A> > : public ResizableBuffer {
+    typedef std::vector<T, A> BufferType;
+    BufferType* buffer_;
```
minggo: buffer_ -> _buffer to obey the cocos2d-x coding style

cocos/platform/CCFileUtils.h
```diff
@@ -41,6 +42,76 @@ NS_CC_BEGIN
  * @{
  */
 
+
+class ResizableBuffer {
+public:
+    virtual ~ResizableBuffer() {}
+    virtual void resize(size_t size) = 0;
```
minggo: How about return `bool` to indicate the result?

minggo: Ok, forget it since std::vector::resize() doesn't return bool

cocos/platform/CCFileUtils.h
```diff
@@ -41,6 +42,76 @@ NS_CC_BEGIN
...
+};
+
+enum class FileError
+{
+    OK = 0,
+    NotExists = 0x01, /** File not exists */
+    OpenFailed = 0x02, /** Open file failed. */
+    ReadFaild = 0x03, /** Read failed */
+    NotInitialized = 0x04, /** FileUtils is not initializes */
+};
```
minggo: How about move it into FileUtils?

minggo:
I run the test case on Mac, it said `getContents() failed: error: sbuf != dbuf`. 

And can we use FileUtils::getContents() to implement FileUtils::getDataFromFile() and FileUtils::getStringFromFile()? Then it will remove duplicated codes.

cocos/platform/CCFileUtils.h
xpol: Should be OK.
(actually I was initially wrote the error in FileUtils, but move out for shorten the return type).

minggo: But i think it will be less conflict.

xpol:
`FileUtils::getStringFromFile()` is actually bad:
- Performance: read whole file, then truncate.
- Portable: Text mode is not portable #15456 
- Memory Usage: use twice memory and copy memory around, 1st whole to the buffer, the copy to a string.
- Memory leak: #15457

Does the new code of `FileUtils::getStringFromFile()` needs to implement the truncate and convert CRLF?
For convert CRLF, do I need to convert CRLF only for Windows (that is the current behavior) ?

minggo:
@xpol What i mean is use FileUtils::getContents() internally in FileUtils::getStringFromFile(), for example

``` c++
 FileUtils::getStringFromFile()
{
    ...
    getContents(...); // use getContents() to get needed result
}
```

xpol:
Yes.
- Does the new code of `FileUtils::getStringFromFile()` needs to implement the truncate and convert CRLF?
- For converting CRLF, do I need to convert CRLF only for Windows (that is the current behavior) ?

minggo:
@xpol I think it is better to keep compatibility and fix bugs in other pull request.

cocos/platform/CCFileUtils.h
```diff
@@ -41,6 +42,66 @@ NS_CC_BEGIN
...
+
+
+template<typename C, typename T, typename A>
+class ResizableBufferAdapter< std::basic_string<C, T, A> > : public ResizableBuffer {
+    typedef std::basic_string<C, T, A> BufferType;
+    BufferType* _buffer;
+public:
+    explicit ResizableBufferAdapter(BufferType* buffer) : _buffer(buffer) {}
+    virtual void resize(size_t size) override {
+        _buffer->resize((size + sizeof(C) - 1) / sizeof(C));
```
minggo: Why the size is `(size + sizeof(C) - 1) / sizeof(C)`

cocos/platform/CCFileUtils.h
```diff
@@ -41,6 +42,66 @@ NS_CC_BEGIN
...
+    virtual void resize(size_t size) = 0;
+    virtual void* buffer() const = 0;
+};
+
+template<typename T>
+class ResizableBufferAdapter { };
+
+
+template<typename C, typename T, typename A>
+class ResizableBufferAdapter< std::basic_string<C, T, A> > : public ResizableBuffer {
```
minggo: Should we support std::basic_string<> or only just support std::string?

xpol: std::basic_string<> will allow read UTF8 UTF16 and UTF32 files?

xpol:
@minggo 
Shall I remove follow API from `FileUtilsAndroid` as they are not necessary now (4c6e4dc):
- FileUtilsAndroid::getDataFromFile()
- FileUtilsAndroid::getStringFromFile()
- FileUtilsAndroid::getFileData()

cocos/platform/CCFileUtils.h
xpol: because that the char type for string and member type for vector may not 8 bit.

cocos/platform/CCFileUtils.h
minggo: Ok.

cocos/platform/CCFileUtils.h
minggo: But i think std::basic_string<> has already handles it.

xpol: [`std::basic_string<>::size()`](http://en.cppreference.com/w/cpp/string/basic_string/size) and [`std::basic_string<>::resize()`](http://en.cppreference.com/w/cpp/string/basic_string/resize) is talking about the number of elements (yes the name make users think it's bytes size).

minggo: So the size here is not the number of element? Why?

xpol: The `ResizableBuffer::resize()` is about file bytes.

When `C` is 8 bit. `std::basic_string<char>::size() == file bytes`
When `C` is not 8 bit: `std::basic_string<wchar_t>::size() != file bytes`

minggo: What i mean is why size in `ResizableBuffer::resize()` is bytes. If you want to read wchar_t, then developers should calculate the size of bytes itself?

xpol: `ResizableBuffer::resize()` is not for user calling, `FileUtils::getContents()` calls it internally:
1. `ResizableBuffer::resize()` is called by ``at cocos/platform/CCFileUtils.cpp` [line 687](https://github.com/cocos2d/cocos2d-x/pull/15479/files#diff-cb65223cd3edf30671fefad4a1883edfR687) in `FileUtils::getContents()` when the file size is determined, and thus, the file size is passed to `ResizableBuffer::resize()`.
2. `ResizableBufferAdapter` convert bytes to number of elements need.

Example Usage:

``` C++
    std::string<char32_t> buf;
    fs->getContents(file, &buf); // resize buf automatically inside getContents()
    std::cout << "I got " << buf.size() << " of wchar32_t" << std::endl;
```

minggo: Got it, thanks.

cocos/platform/CCFileUtils.h
minggo: I think it is better to change type names to:

```
C -> charT
T -> traits
A -> Alloc
```

It is more readable.

xpol: OK, I'll follow the name in here: http://en.cppreference.com/w/cpp/string/basic_string

minggo: Thanks.

xpol:
Updated template parameter names.

cocos/platform/CCFileUtils.cpp
```diff
@@ -610,108 +610,107 @@ void FileUtils::purgeCachedEntries()
...
-        mode = "rt";
-    else
-        mode = "rb";
+static inline void textFormCRLF(std::string& s) {
+    static const std::string CRLF("\r\n");
+    static const std::string LF("\n");
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WIN32 // this is what getStringFromFile() does before...
+    s = std::move(replaceAll(s, CRLF, LF));
+#endif
+}
```
minggo: It is better to add namespace

``` c++
namespace 
{
    inline std::string replaceAll(...)
    {
        ...
    }
}
```

minggo:
I will run tests to check the result next week.

cocos/platform/CCFileUtils.cpp
xpol: How about add `replaceAll` to `StringUtils`?

minggo: Sounds good.

xpol:
## getStringFromFile()

I checked the code for win32 and winrt, found that actually they don't use text mode (the CRLF are not converted to LF):
- https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/win32/CCFileUtils-win32.cpp#L300
- https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/winrt/CCFileUtilsWinRT.cpp#L336

## getFileData()

Accroding to this:
- https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/win32/CCFileUtils-win32.cpp#L370

The `FileUtilsWin32::getFileData()` does not handle the mode arg, thus, only binary mode are implemented.
For winrt, it use the default implementation in FileUtils, that is text mode are used.

## conclusion:
- `FileUtils::getStringFromFile()` : We don't need to simulate the text mode, just truncate at 1st zero.
- `FileUtils::getFileData()` : Only WinRT has support text mode. (I guess this is a bug, shall we fix it?)

xpol:
`CCFileUtilsWin32.cpp`

https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/win32/CCFileUtils-win32.cpp#L298

call to `checkFileName` has no effect: both arg is const ref and return value not checked.

I will remove it in `FileUtilsWin32::getContents()`.

cocos/platform/CCFileUtils.cpp
```diff
@@ -610,108 +610,111 @@ void FileUtils::purgeCachedEntries()
...
 
-        fseek(fp,0,SEEK_END);
-        *size = ftell(fp);
-        fseek(fp,0,SEEK_SET);
-        buffer = (unsigned char*)malloc(*size);
-        *size = fread(buffer,sizeof(unsigned char), *size,fp);
-        fclose(fp);
-    } while (0);
+    std::string s;
+    getContents(filename, &s);
```
minggo: Should check the result?

cocos/platform/CCFileUtils.cpp
```diff
@@ -610,108 +610,111 @@ void FileUtils::purgeCachedEntries()
...
-        std::string msg = "Get data from file(";
-        msg.append(filename).append(") failed!");
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
+    // TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
+    bool textmode = strchr(mode, 't') != NULL;
+    if (textmode)
+        textFormCRLF(s);
+#endif
+
+    unsigned char * buffer = (unsigned char*)malloc(s.size());
```
minggo, line 673: Check the result?

cocos/platform/CCFileUtils.h
```diff
@@ -99,6 +160,87 @@ class CC_DLL FileUtils
      */
     virtual Data getDataFromFile(const std::string& filename);
 
+
+    enum class Error
```
minggo: I think it is better to rename Result or something similar because the return type may be ok.

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
```diff
@@ -592,6 +593,127 @@ std::string TestWriteString::subtitle() const
...
+void TestGetContents::onEnter()
+{
+    FileUtilsDemo::onEnter();
+    auto fs = FileUtils::getInstance();
+
+    auto test_if_compiles = [fs]() {
+        fs->getContents("", (CustomBuffer*)(nullptr));
+        AlreadyExistsBuffer buf;
+        fs->getContents("", &buf);
+    };
```
minggo, line 641: Test codes? If so, please remove them.

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
```diff
@@ -592,6 +593,127 @@ std::string TestWriteString::subtitle() const
     return "";
 }
 
+class CustomBuffer : public ResizableBuffer {};
+
+struct AlreadyExistsBuffer {};
+
+NS_CC_BEGIN
```
minggo, line 600: Why need the namespace?

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
```diff
@@ -592,6 +593,127 @@ std::string TestWriteString::subtitle() const
...
+    (void)(test_if_compiles);
+
+    auto winSize = Director::getInstance()->getWinSize();
+
+    auto readResult = Label::createWithTTF("show readResult", "fonts/Thonburi.ttf", 16);
+    this->addChild(readResult);
+    readResult->setPosition(winSize.width / 2, winSize.height / 2);
+
+    std::vector<char> binary = {'\r','\n','\r','\n','\0','\0','\r','\n'};
+    generated_file = fs->getWritablePath() + "file-with-zeros-and-crlf";
```
minggo: `generated_file -> generatedFile` to obey cocos2d-x coding style.

cocos/platform/CCFileUtils.cpp
```diff
@@ -610,108 +610,111 @@ void FileUtils::purgeCachedEntries()
...
 
-    auto fileutils = FileUtils::getInstance();
-    do
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
+// TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
+namespace
+{
+    // TODO: move this function to StringUtils.
+    // http://stackoverflow.com/a/7724536/87021
+    static inline std::string replaceAll(
```
minggo: I think static is not need here.

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
xpol, line 600: `ResizableBufferAdapter` was defined in `cocs2d` namespace.
So specialized `ResizableBufferAdapter< AlreadyExistsBuffer >` sould defined in `cocs2d` namespace.

minggo, line 600: i think it can be just do as
`class ResizableBufferAdapter<AlreadyExistsBuffer> : public cocos2d::ResizableBuffer`

xpol, line 600: No you can't. [Why aren't template specializations allowed to be in different namespaces?](http://stackoverflow.com/questions/3072248/why-arent-template-specializations-allowed-to-be-in-different-namespaces)

minggo, line 600: But there is not compiling error in Xcode. I haven't tried on other platforms.

xpol, line 600: Se [in detail section in here](http://en.cppreference.com/w/cpp/language/template_specialization):

> Explicit specialization can only appear at namespace scope, in the same namespace as the primary template (or in the nearest enclosing namespace if the primary is a member template), and it has to appear after the non-specialized template declaration. It is always in the scope of that namespace:

minggo, line 600: But i think it is not explicit specialization here, it is inheritance here. Did you met compiling error? It works ok when when i
- remove NS_CC_BEGIN NS_CC_END
- and changed to `class ResizableBufferAdapter<AlreadyExistsBuffer> : public cocos2d::ResizableBuffer`

xpol:
@minggo 
Have you got any idea about the Linux build error:

```
In file included from /linux_slave/workspace/ccs-pr/node/linux/cocos/platform/linux/CCDevice-linux.cpp:30:0:
/linux_slave/workspace/ccs-pr/node/linux/cocos/platform/CCFileUtils.h:164:16: error: anonymous scoped enum is not allowed
     enum class Status
                ^
In file included from /linux_slave/workspace/ccs-pr/node/linux/cocos/platform/linux/CCDevice-linux.cpp:39:0:
/linux_slave/workspace/ccs-pr/node/linux/cocos/platform/CCFileUtils.h:164:5: warning: elaborated-type-specifier for a scoped enum must not use the ‘class’ keyword
     enum class Status
     ^
In file included from /linux_slave/workspace/ccs-pr/node/linux/cocos/platform/linux/CCDevice-linux.cpp:30:0:
/linux_slave/workspace/ccs-pr/node/linux/cocos/platform/CCFileUtils.h:164:16: error: expected identifier before ‘int’
     enum class Status
                ^
In file included from /linux_slave/workspace/ccs-pr/node/linux/cocos/platform/linux/CCDevice-linux.cpp:39:0:
/linux_slave/workspace/ccs-pr/node/linux/cocos/platform/CCFileUtils.h:165:5: error: expected unqualified-id before ‘{’ token
     {
     ^
cocos/CMakeFiles/cocos2dInternal.dir/build.make:2607: recipe for target 'cocos/CMakeFiles/cocos2dInternal.dir/platform/linux/CCDevice-linux.cpp.o' failed
```

minggo:
Remove the last `,`?

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
xpol, line 600: It is definitely explicit specialization.

C++ does not support defined a class with `< XX >` eg:

``` C++
class A<int> {
};
```

After :
- remove NS_CC_BEGIN NS_CC_END
- and changed to class ResizableBufferAdapter<AlreadyExistsBuffer> : public cocos2d::ResizableBuffer

That will got the follow error on VS2015:

```
1>  FileUtilsTest.cpp
1>..\Classes\FileUtilsTest\FileUtilsTest.cpp(601): error C2906: 'cocos2d::ResizableBufferAdapter<AlreadyExistsBuffer>': explicit specialization requires 'template <>'
```

minggo, line 600: Ok. Thanks.

minggo:
About the linux error, i think it is because some header files define a macro named `Status`, such as

```
#define Status
```

xpol:
Yes, I have moved the include of `CCFileUtils.h` up in `cocos\platform\linux\CCDevice-linux.cpp`

xpol:
ping

cocos/platform/CCFileUtils.cpp
minggo: @xpol I think static is not needed when it is in an anonymous namespace.

cocos/platform/CCFileUtils.cpp
```diff
@@ -610,108 +610,116 @@ void FileUtils::purgeCachedEntries()
...
+    std::string s;
+    if (getContents(filename, &s) != Status::OK)
+        return nullptr;
+
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
+    // TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
+    bool textmode = strchr(mode, 't') != NULL;
+    if (textmode)
+        textFormCRLF(s);
+#endif
```
minggo: I think these codes should be applied to WIN32 too. Sorry for my previous answer about it. And NULL -> nullptr

minggo:
@xpol I added comments above. I think i can merge it when it is resolved. Thanks.

cocos/platform/CCFileUtils.cpp
xpol: According to [comments above](https://github.com/cocos2d/cocos2d-x/pull/15479#issuecomment-213694429) in our **current code base** :
- For `FileUtils::getStringFromFile()` : do not convert CRLF to LF on **all platform** including [**Win32**](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/win32/CCFileUtils-win32.cpp#L300) and [**WinRT**](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/winrt/CCFileUtilsWinRT.cpp#L336).
- For `FileUtils::getFileData()` : Only **WinRT** do CRLF to LF conversion ( [`mode` arg is not used on Win32](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/win32/CCFileUtils-win32.cpp#L370) ). **And I guess this is a bug**.

Did you mean in this PR, we shall do support CRLF to LF on Win32 and WinRT for both `FileUtils::getStringFromFile()` and `FileUtils::getFileData()`?

IMHO, we shold remove all support of CRLF to LF behavior as:
- it's not portable (and cocos2d-x is a cross platform engine).
- currently only `FileUtils::getFileData()` on WinRT support CRLF to LF.
- all other case (non WinRT or not `FileUtils::getFileData()`), the current behavior is not convert CRLF to LF.

xpol:
@minggo 

I updated the line note, I just want to make it sure.

Would please inspect the the links in above line note link to current `v3` code if you needs detailed information.

cocos/platform/CCFileUtils.cpp
minggo: Sorry, i misunderstood of the logic of `textFormCRLF`. Yep, i agree with you, we should not convert CRLF to LF. So i think we can remove these codes.

minggo:
@xpol I think we can remove the codes

```
+#if CC_TARGET_PLATFORM == CC_PLATFORM_WINRT
+    // TODO: only winrt support text mode. it is possibly a bug. if true, remove the follow code for text mode.
+    bool textmode = strchr(mode, 't') != NULL;
+    if (textmode)
+        textFormCRLF(s);
+#endif
```

And static is not needed if a function is in an anonymous namespace.

xpol:
OK then all platform has the same behavior: no text mode support for file reading.

BTW, the static and NULL issues is gone as they are in the CRLF to LF support code.

minggo:
Thanks. I think it is ok to do it like this. And let's fix it if we met issues in future.

xpol:
Glad I can help.

PR #15487, opened 2016-04-21
add setIgnoreAnchorPointForPosition(bool) func, clean lines

njh0602:
and, removed completed @todo comments.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1,5 +1,3 @@
-
-
 /****************************************************************************
 Copyright (c) 2008-2010 Ricardo Quesada
 Copyright (c) 2009      Valentin Milea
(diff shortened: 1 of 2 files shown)


minggo:
@njh0602 Thanks. But i think it is better to remove the usage of `ignoreAnchorPointForPosition()`. Because this function is used so many places, so i think it's better not to modify it.

njh0602:
@minggo Thanks to opinion, But `ignoreAnchorPointForPosition()` call `setIgnoreAnchorPointForPosition()`. So I think It does not affect the existing codes.

minggo:
But i will bring many warnings.

njh0602:
@minggo i think it has some mistakes. i will check again.

njh0602:
@minggo i understand your opinion now. for example, Sprite class has ignoreAnchorPointForPosition and many other classes? if i fix all classes methods, is it ok?

minggo:
Yep, there are two cases:
- Classes that inherits from Node and override `ignoreAnchorPointForPosition()` should be deprecated too and add new implementation
- codes that invoked `ignoreAnchorPointForPosition()` should use new functions

njh0602:
@minggo ok, i will be back.

PR #15488, opened 2016-04-21
Fixed a bug that audio resources

VisualSJ:
@pandamicro 
Why does JSB need to add 'res'?


minggo:
Because it is under `res` folder.
And please send this PR to v3.11.

PR #15496, opened 2016-04-23
[fix bug] reset UIGrayScale Shader when reload shaders

treert:
reset GLProgram for UIGrayScale Shader when reload shaders


minggo:
Thanks.

joewan:
well done！

PR #15498, opened 2016-04-24
Warn if fontDef.charID exceeds 16bits.

ynaoto:
It would be better to warn and ignore if charID in *.fnt file is bigger than 65536.

### DIFF
--- a/cocos/2d/CCFontFNT.cpp
+++ b/cocos/2d/CCFontFNT.cpp
@@ -809,7 +809,11 @@ FontAtlas * FontFNT::createFontAtlas()
         tempDefinition.validDefinition = true;
         tempDefinition.xAdvance = fontDef.xAdvance;
         // add the new definition
-        tempAtlas->addLetterDefinition(fontDef.charID,tempDefinition);
+        if (65535 < fontDef.charID) {
+            CCLOGWARN("Warning: 65535 < fontDef.charID (%u), ignored", fontDef.charID);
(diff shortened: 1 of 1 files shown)


cocos/2d/CCFontFNT.cpp
```diff
@@ -809,7 +809,11 @@ FontAtlas * FontFNT::createFontAtlas()
         tempDefinition.validDefinition = true;
         tempDefinition.xAdvance = fontDef.xAdvance;
         // add the new definition
-        tempAtlas->addLetterDefinition(fontDef.charID,tempDefinition);
+        if (65535 < fontDef.charID) {
```
minggo, line 812: Should include 65535 as `65535 <= fontDef.charID` ?

ynaoto, line 812: I'm not sure.  The problem is that if .fnt file includes 'id' equals or greater than 65536, the upper bits of the value will be lost because of down casting to char16_t, the first arg of FontAtlas::addLetterDefinition method function. The value is used as a key for _letterDefinitions map.  Therefore, if 'id' equals to 65633, the value will override an entry of _letterDefinitions for character 'a'.  This made me trouble to find the problem in my .fnt file.  Just produce warning message will save our time.

minggo, line 812: I think you are right. It includes 65535.

PR #15500, opened 2016-04-25
Fix compiler warning C4804 under Visual Studio

mogemimi:
This pull request fixes the warning C4804 that was created by commit a4d67867d78bb8f95a8eb5182249850fe2014d5b with Visual Studio 2015:

```
..\3d\CCSprite3D.cpp(777): warning C4804: '>' : unsafe use of type 'bool' in operation [cocos\2d\libcocos2d.vcxproj]
```


minggo:
Thanks.

PR #15508, opened 2016-04-26
Remove last internal use of deprecated `FileUtils::getFileData()`.

xpol:
Sorry lots of whitespace changes.

### DIFF
--- a/cocos/platform/winrt/CCFreeTypeFont.cpp
+++ b/cocos/platform/winrt/CCFreeTypeFont.cpp
@@ -577,7 +577,7 @@ void  CCFreeTypeFont::compute_bbox(std::vector<TGlyph>& glyphs, FT_BBox  *abbox)
 }
 
 namespace {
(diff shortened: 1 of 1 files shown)


cocos/platform/winrt/CCFreeTypeFont.cpp
```diff
@@ -571,12 +571,22 @@ void  CCFreeTypeFont::compute_bbox(std::vector<TGlyph>& glyphs, FT_BBox  *abbox)
...
     }
-  
+
     /* return string bbox */
     *abbox = bbox;
 }
 
-unsigned char* CCFreeTypeFont::loadFont(const char *pFontName, ssize_t *size) 
+namespace {
+    inline unsigned char* takeBuffer(Data& data, ssize_t* size = nullptr) {
```
minggo: Why size need default value? As it is used below, it size is needed.

xpol: Updated.

PR #15515, opened 2016-04-26
Add OBB (zip) support

adesentenac:
This pull request adds support for OBB files.
OBB must be a zip file with audio and video files stored uncompressed.
The data must be at the root of the OBB (not in an assets folder like in the APK),

For example, in the assets folder, the OBB can be created with: 
`zip -rn .ogg:.mp3:.wav:.mp4 main.<version>.<android package name>.obb *`

Tested with CppTests, either with assets in the APK (to confirm there is no regression) or in the OBB.

Using the new AudioEngine with OBB is only supported on Android 3.1 (API level 12) or above.


cocos/platform/android/CCFileUtils-android.cpp
```diff
@@ -249,6 +268,12 @@ FileUtils::Status FileUtilsAndroid::getContents(const std::string& filename, Res
     } else {
         relativePath = fullPath;
     }
+    
+    if (obbfile)
+    {
+        if (obbfile->getFileData(relativePath, buffer))
+            return FileUtils::Status::OK;
```
minggo, line 275: I think it needs to return true here.

minggo:
@adesentenac how can i test it? Could you please describe it in more detail? 

> Using the new AudioEngine with OBB is only supported on Android 3.1 (API level 12) or above.

If i build with Android 2.3, does it cause compiling error?

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
```diff
@@ -226,6 +284,29 @@ public static boolean openURL(String url) {
...
+    
+    public static long[] getObbAssetFileDescriptor(final String path) {
+        long[] array = new long[3];
+        if (Cocos2dxHelper.sOBBFile != null) {
+            AssetFileDescriptor descriptor = Cocos2dxHelper.sOBBFile.getAssetFileDescriptor(path);
+            if (descriptor != null) {
+                try {
+                    ParcelFileDescriptor parcel = descriptor.getParcelFileDescriptor();
+                    Method method = parcel.getClass().getMethod("getFd", new Class[] {});
+                    array[0] = (Integer)method.invoke(parcel);
```
minggo, line 296: I got it, use reflection to avoid depending on API 12 because ParcelFileDescriptor.getFd() exists since API 12. But it is better to add some comment here.

minggo:
And it requires permission `<uses-permission
        android:name="android.permission.WRITE_EXTERNAL_STORAGE" />`

Should we add it into cpp-tests?
And maybe we should add it into template too.

cocos/platform/android/CCFileUtils-android.cpp
adesentenac, line 275: I'm not sure what you mean.
It will return OK if it is found in the OBB, otherwise it will try to find it in the APK (so it must not return if getFileData returns false).
Is that what you meant?
I can add a comment to make it clearer maybe?

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
adesentenac, line 296: You're right, I'll add a comment to make it clearer.

adesentenac:
Good point, I'm not sure how to handle the permission.
By default, cpp-tests still puts everything in the APK.

To be able to create an APK+OBB, you need to build the cpp-tests as usual with android-build.py, then:
- go into tests/cpp-tests/proj.android
- rename the assets folder to data (or any other name)
- run: ant debug (to generate the APK without the assets in it)
- go into the new data folder and run: 
  - `zip -rn .ogg:.mp3:.wav:.mp4 main.1.org.cocos2dx.cpp_tests.obb *`

Also what is the policy for the templates, should I add the permission and users will remove it if they don't use the OBB in their project?

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
minggo, line 296: Thanks

minggo:
@adesentenac should i upload it to google play to have a try? I am in China, it is hard for me to push it to google play, do you already have a demo?

About template, i planned to add the permission by default. I am not sure if it has side effect if user doesn't  use OBB in their project.

cocos/platform/android/CCFileUtils-android.cpp
minggo, line 275: As the return value type is `bool`, and `FileUtils::Status::OK` is an enum value. So it return incorrect type here.

adesentenac, line 275: I think you misread it, this code is in FileUtilsAndroid::getContents which returns a FileUtils::Status

minggo, line 275: Oh, yep. Sorry about it.

mars3142:
Why would you add the _write external storage_ permission? The obb file is located in **/Android/obb/packageName** which is a private folder and do not need any specific permission.

minggo:
@mars3142  please refer to [Declaring user permissions section](https://developer.android.com/google/play/expansion-files.html).

mars3142:
@minggo Oh, my fault. Sorry. I mixed it up with **/Android/data/packageName**. This folder do not need any permissions in API 19+.

PR #15517, opened 2016-04-27
Add tizen support 

zilongshanren:
@minggo  please review it


cocos/editor-support/cocostudio/CCSGUIReader.cpp
```diff
@@ -42,6 +42,11 @@ THE SOFTWARE.
 #include "editor-support/cocostudio/CocoLoader.h"
 #include "tinyxml2.h"
 
+#include "platform/CCFileUtils.h"
+#include "2d/CCSpriteFrameCache.h"
+#include "base/CCDirector.h"
+#include "base/ccUtils.h"
```
minggo, line 48: Why need to include these header files?

Issue #15111, opened 2016-02-25
When make get error：‘isnan’ was not declared in this scope


icecity96:
When I make I got the following error：

```
/opt/cocos2d-x/cocos/2d/CCActionCamera.cpp:197:22: error: ‘isnan’ was not declared in this scope
     if( isnan(_angleX) )
                      ^
/opt/cocos2d-x/cocos/2d/CCActionCamera.cpp:197:22: note: suggested alternative:
In file included from /usr/include/c++/5.3.0/random:38:0,
                 from /opt/cocos2d-x/cocos/base/ccRandom.h:29,
                 from /opt/cocos2d-x/cocos/base/ccMacros.h:73,
                 from /opt/cocos2d-x/cocos/math/CCGeometry.h:32,
                 from /opt/cocos2d-x/cocos/2d/CCAction.h:32,
                 from /opt/cocos2d-x/cocos/2d/CCActionInterval.h:33,
                 from /opt/cocos2d-x/cocos/2d/CCActionCamera.h:31,
                 from /opt/cocos2d-x/cocos/2d/CCActionCamera.cpp:28:
/usr/include/c++/5.3.0/cmath:641:5: note:   ‘std::isnan’
     isnan(_Tp __x)
cocos/CMakeFiles/cocos2d.dir/build.make:86: recipe for target 'cocos/CMakeFiles/cocos2d.dir/2d/CCActionCamera.cpp.o' failed
make[2]: *** [cocos/CMakeFiles/cocos2d.dir/2d/CCActionCamera.cpp.o] Error 1
CMakeFiles/Makefile2:419: recipe for target 'cocos/CMakeFiles/cocos2d.dir/all' failed
make[1]: *** [cocos/CMakeFiles/cocos2d.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

```

OS:Arch linux
Compiler: GNU 5.3.0

icecity96:
I have solved this problem,just using std::isnan repalce isnan...
[see here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=48891)

Lechindianer:
Same problem occurs here with the same architecture. It would be great if this fix would get merged into master. @icecity96: Could you please reopen the issue to fix this issue for all? :)

The isnan <-> std::isnan problem exists in 2 files:  cocos/2d/CCActionCamera.cpp and extensions/GUI/CCControlExtension/CCControlUtils.cpp

[...]

PR #15522, opened 2016-04-27
Fix compile error isnan -> std::isnan #15111

gelldur:
Issue: https://github.com/cocos2d/cocos2d-x/issues/15111

### DIFF
--- a/cocos/2d/CCActionCamera.cpp
+++ b/cocos/2d/CCActionCamera.cpp
@@ -194,11 +194,11 @@ void OrbitCamera::startWithTarget(Node *target)
 
     float r, zenith, azimuth;
     this->sphericalRadius(&r, &zenith, &azimuth);
-    if( isnan(_radius) )
+    if( std::isnan(_radius) )
         _radius = r;
-    if( isnan(_angleZ) )
+    if( std::isnan(_angleZ) )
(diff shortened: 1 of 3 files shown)


minggo:
> f:\jenkins\workspace\ccs-pr\node\windows-universal_bak\cocos\2d\libcocos2d_8_1\libcocos2d_8_1\libcocos2d_8_1.Shared......\CCActionCamera.cpp(197): error C2039: '_isnan' : is not a member of 'std' [f:\jenkins\workspace\ccs-pr\node\windows-universal_bak\cocos\2d\libcocos2d_8_1\libcocos2d_8_1\libcocos2d_8_1.Windows\libcocos2d_8_1.Windows.vcxproj]

It seems std::isnan is missing in VS.

stammen:
@gelldur I think you might need to #include <cmath> to get isnan into the std namespace.

stammen:
@gelldur @minggo The issue with std::isnan is in cocos2d\cocos\platform\winrt\CCStdC.h

Please add the following at line 35 
#include <cmath>

Please remove the following at line 49

#if !defined(isnan)
    #define isnan   _isnan
#endif

I suspect that the win32 version will have the same issue.

minggo:
@gelldur Could you try as @stammen suggested?

minggo:
I did as @stammen suggested in #15569

PR #15535, opened 2016-04-29
Refactor FileUtils::getFileData() and CCFreeTypeFont::loadFont().

xpol:
- Remove unnecessary memory copy in `FileUtils::getFileData()`
- Add `Data::takeBuffer()` to simplify `FileUtils::getFileData()` and `CCFreeTypeFont::loadFont()`.


minggo:
Thanks, looks good to me.

xpol:
@minggo 
Thank you!

xpol:
The FileUtils::getFileData() was implemented first read to string, then convert CRLF to LF, and then make a copy of string data and return.
And later we removed the convert CRLF.

So we don't need to read to string and copy we can just read to Data and take buffer from Data and return.

PR #15541, opened 2016-04-30
Fix doxygen format mistakes in enum class Platform

walzer:
An user reported this error in the forum http://forum.cocos.com/t/api-document/36199/2
I noticed that our doxygen was written in wrong format.


aimspike:
Haha, I found that tiny bug~

PR #15543, opened 2016-05-02
Fix shadow variable warning in GameNode3DReader

mogemimi:
This pull request fixes the following warning `-Wshadow` when compiling `libcocos2d Mac` with Xcode 7.3:

```
cocos/editor-support/cocostudio/WidgetReader/GameNode3DReader/GameNode3DReader.cpp:164:25: declaration shadows a local variable
```

Thanks!


minggo:
Thanks.

PR #15548, opened 2016-05-03
Add findChild for cocos2d::utils

halx99:
Add findChild for cocos2d::utils


minggo:
@halx99 Why there are so many unrelated modification? Did you use windows line break(`\r\n`)?

halx99:
@minggo  I convert to unix style, there are so many unrelated modification still, I don't known why?

minggo:
@halx99 I think there is not unrelated modifications now. All added codes are not added by you?

halx99:
@minggo I just upload the copy files to my branch.

cocos/base/ccUtils.h
```diff
@@ -109,7 +109,42 @@ namespace utils
      * @return Returns an instance of sprite
      */
     Sprite* createSpriteFromBase64(const char* base64String);
+    
+    /**
+     * Search a node by name
+
+     * @return  Returns searched node or nullptr
+     */
+    Node*  searchNode(Node* levelRoot, const char* name, bool recursively = true);
```
minggo: Why not just rename to findChild, it indeed just find a child.

minggo: And i think `recursively` is needed, or you can use `Node::getChildByName()` instead.

cocos/base/ccUtils.h
```diff
@@ -109,7 +109,42 @@ namespace utils
...
+    {
+        return dynamic_cast<_Ty*>(searchNode(levelRoot, name, recursively));
+    }
+
+    /**
+     * Search a node by tag
+
+     * @return  Returns searched node or nullptr with specificed type '_Ty'
+     */
+    template<typename _Ty> inline
```
minggo: The name `_Ty` is strange.

cocos/base/ccUtils.h
halx99: @minggo 
Yes it can rename to findChild
The Node::getChildByName() not support recursively currently.

cocos/base/ccUtils.h
halx99: This like:  /**
     \* Gets a child from the container with its tag that can be cast to Type T.
     *
     \* @param tag   An identifier to find the child node.
     *
     \* @return a Node with the given tag that can be cast to Type T.
    */
    template <typename T>
    inline T getChildByTag(int tag) const { return static_cast<T>(getChildByTag(tag)); }

cocos/base/ccUtils.h
minggo: Yep, i mean this function should be recursively without parameter `recursively`.

cocos/base/ccUtils.h
halx99: Regards T as a pointer is a better choice.

cocos/base/ccUtils.h
halx99: OK, I got it.

cocos/base/ccUtils.h
minggo: Yep, use pointer is better. And rename `_Ty` to `T` is better too.

minggo:
Thanks.

PR #15555, opened 2016-05-03
Reduce unnecessary c_str() calls

mogemimi:
This pull request removes several `c_str()` calls to improve readability and avoid the copying overhead of conversion to `std::string`.


minggo:
Thanks.

PR #15558, opened 2016-05-04
Add captureNode for cocos2d::utils

halx99:
Add captureNode for cocos2d::utils and fix CC_DLL syntax.

Bring all CC_DLL to front of function decl, The follow is the reason:
For MSVC: 
int __declspec(dllexport) compile_ok();
int\* __declspec(dllexport) compile_error(); // compile will failed with syntax error, when the return type is pointer
int  __declspec(dllexport) \* compile_ok2(); // compile ok, but not styled.
__declspec(dllexport) int\* compile_fine(); // so this a good syntax for DLL export.

### DIFF
--- a/cocos/base/ccUtils.h
+++ b/cocos/base/ccUtils.h
@@ -53,6 +53,7 @@ Examples:
 int ccNextPOT(int value);
 
 class Sprite;
+class Image;
 
 namespace utils
 {
@@ -64,7 +65,16 @@ namespace utils
      * base filename ("hello.png" etc.), don't use a relative path containing directory names.("mydir/hello.png" etc.).
      * @since v3.2
      */
-    void CC_DLL captureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename);
+    CC_DLL void  captureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename);
+
+    /** Capture a specific Node.
+    * @param startNode: specify the snapshot Node. It chould be cocos2d::Scene
+    * @param scale
+    * @returns: return a Image, then can call saveToFile to save the image as "xxx.png or xxx.jpg".
+    * @since v3.11
+    * !!! remark: Caller is responsible for releasing it by calling delete.
+    */
+    CC_DLL Image* captureNode(Node* startNode, float scale = 1.0f);
     
     /** Find children by name, it will return all child that has the same name.
      * It supports c++ 11 regular expression. It is  a helper function of `Node::enumerateChildren()`.
@@ -75,7 +85,7 @@ namespace utils
      * @return Array of Nodes that matches the name
      * @since v3.2
      */
-    std::vector<Node*> CC_DLL findChildren(const Node &node, const std::string &name);
+    CC_DLL std::vector<Node*>  findChildren(const Node &node, const std::string &name);
     
     /** Same to ::atof, but strip the string, remain 7 numbers after '.' before call atof.
      * Why we need this? Because in android c++_static, atof ( and std::atof ) is unsupported for numbers have long decimal part and contain
@@ -83,46 +93,46 @@ namespace utils
      * @param str The string be to converted to double.
      * @return Returns converted value of a string.
      */
-    double CC_DLL atof(const char* str);
+    CC_DLL double  atof(const char* str);
 
     /** Get current exact time, accurate to nanoseconds.
      * @return Returns the time in seconds since the Epoch.
      */
-    double CC_DLL gettime();
+    CC_DLL double  gettime();
 
     /**
      * Get current time in milliseconds, accurate to nanoseconds
      *
      * @return  Returns the time in milliseconds since the Epoch.
      */
-    long long CC_DLL getTimeInMilliseconds();
+    CC_DLL long long  getTimeInMilliseconds();
 
     /**
      * Calculate unionof bounding box of a node and its children.
      * @return Returns unionof bounding box of a node and its children.
      */
-    Rect CC_DLL getCascadeBoundingBox(Node *node);
+    CC_DLL Rect getCascadeBoundingBox(Node *node);
 
     /**
      * Create a sprite instance from base64 encoded image.
 
      * @return Returns an instance of sprite
      */
-    Sprite* createSpriteFromBase64(const char* base64String);
+    CC_DLL Sprite* createSpriteFromBase64(const char* base64String);
     
     /**
      * Find a child by name recursively
 
      * @return  Returns found node or nullptr
      */
-    Node*  findChild(Node* levelRoot, const char* name);
+    CC_DLL Node*  findChild(Node* levelRoot, const char* name);
 
     /**
      * Find a child by tag recursively
 
      * @return Returns found node or nullptr
      */
-    Node*  findChild(Node* levelRoot, int tag);
+   CC_DLL Node*  findChild(Node* levelRoot, int tag);
 
     /**
      * Find a child by name recursively
--- a/tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
+++ b/tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
@@ -35,6 +35,7 @@ NewRendererTests::NewRendererTests()
     ADD_TEST_CASE(NewCullingTest);
     ADD_TEST_CASE(VBOFullTest);
     ADD_TEST_CASE(CaptureScreenTest);
+    ADD_TEST_CASE(CaptureNodeTest);
     ADD_TEST_CASE(BugAutoCulling);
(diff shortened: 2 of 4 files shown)


cocos/base/ccUtils.h
```diff
@@ -64,7 +65,16 @@ namespace utils
...
+    CC_DLL void  captureScreen(const std::function<void(bool, const std::string&)>& afterCaptured, const std::string& filename);
+
+    /** Capture a specific Node.
+    * @param startNode: specify the snapshot Node. It chould be cocos2d::Scene
+    * @param scale
+    * @returns: return a Image, then can call saveToFile to save the image as "xxx.png or xxx.jpg".
+    * @since v3.11
+    * !!! remark: Caller is responsible for releasing it by calling delete.
+    */
+    CC_DLL Image* captureNode(Node* startNode, float scale = 1.0f);
```
minggo, line 77: Why return Image*? Why not implemented as `captureScreen`?

halx99, line 77: @minggo That's for share the Image data to wx directly, not only store to disk. So I think return Image\* is flexible.

minggo, line 77: But what should i do if i want to share a file? Return file, then developers can create an Image if needed, but if i want a file, what should i do?

halx99, line 77: @minggo  Just call Image::saveToFile.

minggo, line 77: Sounds reasonable.
And `captureScreen` can use this function internally? Will you please modify it too?

halx99, line 77: You mean, Let me modify the captureScreen implemention? The captureNode is finished snapshot at this frame. If call this function at captureScreen internally, "captureScreen" prototype will be void captureScreen(const std::string& filename), the call back "const std::function<void(bool, const std::string&)>& afterCaptured" never needed. such as follow:

void captureScreen(std::string& filename)
{
     auto image = captureNode(Director::getInstance()->getRunningScene());
    if (FileUtils::getInstance()->isAbsolutePath(filename))
            {
                outputFile = filename;
            }
            else
            {
                CCASSERT(filename.find("/") == std::string::npos, "The existence of a relative path is not guaranteed!");
                outputFile = FileUtils::getInstance()->getWritablePath() + filename;
            }
     AsyncTaskPool::getInstance()->enqueue(AsyncTaskPool::TaskType::TASK_IO, mainThread, (void*)NULL, [image, outputFile]()
            {
                image->saveToFile(outputFile);
                image->release();
            });
}

minggo, line 77: Ok, let's keep it as it is now.

halx99, line 77: OK.

minggo:
@halx99 it looks good to me, could you please add a test case?

halx99:
@minggo  Your mean is add to cpp-test project? or the usage?

minggo:
Yep, add a test case in `cpp-tests` project.

halx99:
Add to CaptureScreenTest or create a new Test case "CaptureNodeTest" ?

minggo:
It is better to add a new test case.

halx99:
OK, wait a while.

minggo:
@halx99 thanks for adding a test case. But i think it is better to show the image in screen or it is hard for testing.

halx99:
@minggo  Yes, I already show the image in screen and also store to disk:

```
// create a sprite with the captured image directly
auto sp = Sprite::createWithTexture(TextureCache::getInstance()->addImage(image, _filename));
addChild(sp, 0, childTag);
Size s = Director::getInstance()->getWinSize();
sp->setPosition(s.width / 2, s.height / 2);

// store to disk
image->saveToFile(_filename, false);
```

minggo:
Thanks. i will have a try.

halx99:
OK.

minggo:
The effect seems a little problem:

capturenodetest

And the effect of `captureSceen()` looks good

capturescreentest

halx99:
I test it at cocos2d-x-3.10 cpp-test project, the effect is ok:
qq 20160506180225

minggo:
You should check the generated file.

halx99:
Because I store file with alpha channel, use windows 10 photo app open it: the effect is also ok:

qq 20160506180749

But use mspaint open it, the effect is show by yours.

halx99:
If you want to get the file same with captureScreen, you just change `image->saveToFile(_filename, false);` to `image->saveToFile(_filename, true);`

minggo:
How about use the same logic the same as captureScreen, it will be more easier for other guys to test.

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
```diff
@@ -478,6 +479,72 @@ void CaptureScreenTest::afterCaptured(bool succeed, const std::string& outputFil
...
+    Director::getInstance()->getTextureCache()->removeTextureForKey(_filename);
+    removeChildByTag(childTag);
+    
+    _filename = FileUtils::getInstance()->getWritablePath() + "/CaptureNodeTest.png";
+
+    // capture this
+    auto image = utils::captureNode(this, 0.5);
+
+    // create a sprite with the captured image directly
+    auto sp = Sprite::createWithTexture(TextureCache::getInstance()->addImage(image, _filename));
```
minggo: `TextureCache::getInstance()` is deprecated, please use `Director::getInstance()->getTextureCache()` instead.

halx99:
captureNode just return Image\* with alpha channel, I think it's flexiable.
Developer determine reserve alpha channel or not when saveToFile: 
qq 20160506181714

tests/cpp-tests/Classes/NewRendererTest/NewRendererTest.cpp
halx99: OK.

minggo:
Yep, i is flexible. I mean modify the test codes to save a file without alpha channel.

halx99:
OK.

halx99:
@minggo I complete the modify, please check.

minggo:
Thanks. I will merge it when it passes jenkins.

halx99:
@minggo 
I check the build logs, there is no error for compile, mac platform faield with script broken: 

*\* BUILD SUCCEEDED **

Checking out files:   8% (53/630)  
...
Checking out files: 100% (630/630), done.
HEAD is now at a2c5304 Merge commit 'refs/pull/15558/head' of https://github.com/cocos2d/cocos2d-x into pull15558
Removing external/Box2D/
Removing external/ConvertUTF/
Removing external/bullet/
Removing external/chipmunk/
Removing external/clipper/
Removing external/curl/
Removing external/edtaa3func/
Removing external/flatbuffers/
Removing external/freetype2/
Removing external/glfw3/
Removing external/jpeg/
Removing external/json/
Removing external/linux-specific/
Removing external/lua/
Removing external/png/
Removing external/poly2tri/
Removing external/recast/
Removing external/spidermonkey/
Removing external/sqlite3/
Removing external/tiff/
Removing external/tinyxml2/
Removing external/unzip/
Removing external/version.json
Removing external/webp/
Removing external/websockets/
Removing external/win10-specific/
Removing external/win32-specific/
Removing external/winrt_8.1-specific/
Removing external/wp_8.1-specific/
Removing external/xxhash/
Removing external/xxtea/
Removing external/zlib/
Removing tests/js-tests/res/
Removing tests/lua-empty-test/src/cocos/
Removing tests/lua-tests/src/cocos/
Removing tools/fbx-conv/
Removing tools/tojs/userconf.ini
Removing tools/tolua/userconf.ini
Removing v3-deps-92.zip
Slave went offline during the build
ERROR: Connection was broken: java.io.IOException: Unexpected termination of the channel
    at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:50)
Caused by: java.io.EOFException
    at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2325)
    at java.io.ObjectInputStream$BlockDataInputStream.readShort(ObjectInputStream.java:2794)
    at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:801)
    at java.io.ObjectInputStream.<init>(ObjectInputStream.java:299)
    at hudson.remoting.ObjectInputStreamEx.<init>(ObjectInputStreamEx.java:48)
    at hudson.remoting.AbstractSynchronousByteArrayCommandTransport.read(AbstractSynchronousByteArrayCommandTransport.java:34)
    at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:48)

Build step 'Execute Python script' marked build as failure
An attempt to send an e-mail to empty list of recipients, ignored.
Finished: FAILURE

minggo:
@halx99 You can add `[ci rebuild]` to trigger jenkins in future.

halx99:
OK.

PR #15568, opened 2016-05-05
fix jsb 3.11 related issues

zilongshanren:


zilongshanren:
@minggo  please review

PR #15575, opened 2016-05-05
Prevent memory leak: force drain on autoreleased `data`.

stevetranby:
Based on discussion: http://discuss.cocos2d-x.org/t/capturescreen-memory-leak/29108/
`@autoreleasepool block` supposedly requires ARC, but seems to work with or without ARC on OS X 10.7+ and iOS 5.0+.
It is currently being used already in HttpClient-apple.mm.

### DIFF
--- a/cocos/platform/ios/CCImage-ios.mm
+++ b/cocos/platform/ios/CCImage-ios.mm
@@ -99,22 +99,20 @@ bool cocos2d::Image::saveToFile(const std::string& filename, bool isToRGB)
     CGImageRelease(iref);    
     CGColorSpaceRelease(colorSpaceRef);
     CGDataProviderRelease(provider);
-    
-    NSData *data;
-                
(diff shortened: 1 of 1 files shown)


minggo:
I don't why when and why `@autorelease block` is used in `HttpClient-apple.mm`, but i think it is better not use it if possible. And i found there is not need to use `@autorelease block` in `HttpClient-apple.mm`, there are all c++ codes inside the `@autorelease block`, i will send a PR to fix it.

davidejones88:
I found this on stack. Looks like you can use @autoreleasepool even in non-ARC environment?

http://stackoverflow.com/questions/7950583/autoreleasepool-without-arc

which references

http://clang.llvm.org/docs/AutomaticReferenceCounting.html#autoreleasepool

What do you think?

minggo:
@davidejones88 thanks for the link.

PR #15578, opened 2016-05-06
Suppress warning C4800 on MSVC

mogemimi:
This patch fixes MSVC warning [C4800](https://msdn.microsoft.com/en-us/library/b6801kcy.aspx) about forcing integer value to bool with Visual Studio 2015:

```
cocos\physics\CCPhysicsBody.cpp(758): warning C4800: 'cpBool' : forcing value to bool 'true' or 'false' (performance warning)
```


minggo:
Thanks.

PR #15581, opened 2016-05-06
revert correct search path

minggo:


minggo:
Jenkins will not build simulator. The jenkins error is jenkins itself.

PR #15583, opened 2016-05-06
loader test bug

VisualSJ:
Test case expression is not clear


minggo:
@VisualSJ i think you planned to send to v3.11 branch, right?

VisualSJ:
@minggo yes~~  =.=

minggo:
So please send to v3.11 branch instead.

PR #15590, opened 2016-05-07
add missing lua const values

ntotani:
Some ScrollViewEvent values are missing for lua bindings.


minggo:
@zilongshanren Would you please take a look?

zilongshanren:
@minggo  
This PR looks good to me.

PR #15591, opened 2016-05-08
Prevent variable shadowing warnings

mogemimi:
This commit prevents two warnings caused by variable shadowing on Xcode 7.3 since commit 9ff61f915ca1e9342114918a898b78655d6369a2 :

```
cocos/base/ccUtils.cpp:348:14: Declaration shadows a local variable
cocos/base/ccUtils.cpp:370:14: Declaration shadows a local variable
```


minggo:
Thanks.

PR #15594, opened 2016-05-09
CCSprite::createNode bug fix.

yutaka-takeda-drecom:
This bug is to occur in a particular model data.
When you call the "Sprite3D::createNode" to children of node generation, pointer of the parent node is null.

This is test data.
[test_create_child_node_bug.zip](https://github.com/cocos2d/cocos2d-x/files/254450/test_create_child_node_bug.zip)

### DIFF
--- a/cocos/3d/CCSprite3D.cpp
+++ b/cocos/3d/CCSprite3D.cpp
@@ -589,6 +589,7 @@ void Sprite3D::createNode(NodeData* nodedata, Node* root, const MaterialDatas& m
                     setScaleY(scale.y);
                     setScaleZ(scale.z);
                     
+                    node = this;
                 }
             }
             else


minggo:
Yep, the logic looks good to me, but i am not familiar with these codes. @ricardoquesada Could you please take a look?

ricardoquesada:
@minggo sure.

ricardoquesada:
@yutaka-takeda-drecom 

what's the code to reproduce it?
I tried creating an `Sprite3D` like this:

``` c++
auto sprite = Sprite3D::create("gate_enter_child_node_bug.c3t", "UV.jpg");
```

and it didn't crash. It works Ok. thanks.

yutaka-takeda-drecom:
@ricardoquesada 

It is the flow of easy-to-understand bug.
Make a Sprite in this code.

```
auto sprite = Sprite3D::create("gate_enter_child_node_bug.c3t", "UV.jpg");
auto animation = Animation3D::create("gate_enter_child_node_bug.c3t");
auto animate = Animate3D::create(animation);
auto repeate = RepeatForever::create(animate);
sprite->runAction(repeate);
```

Then,
1. 528 line " if(mesh) " of Sprite3D::createNode function will be true.
2. Argument node of the 635 line of createNode will be null. Therefore , node is not created.
3. "warning: no animation found for the skeleton" is output to the log. The reason for this is because there is no node.

ricardoquesada:
@yutaka-takeda-drecom  thanks. I was able to reproduce the bug. Added test case in this PR: https://github.com/cocos2d/cocos2d-x/pull/15719

@minggo this is a sprite3d bug. I think Tony should take a look at it, instead of me.

yutaka-takeda-drecom:
thank you.

ricardoquesada:
tested your PR. Works Ok. Thanks.

PR #15612, opened 2016-05-11
V3: Win10 UWP removed missing files from project

stammen:
This pull request removes the old cocos studio test files from the Window 10 UWP cpp-test project.


minggo:
Thanks.

PR #13903, opened 2015-09-17
Textfiled support cursor v1.1

perminovVS:
Add StringUtils::StringUTF8 for easy manipulate utf8 string.
If TextField active(focused input text) show text cursor( default '|',
change TextFieldTTF::setCursorChar ).
Show/Hide interval CURSOR_TIME_SHOW_HIDE
Change position text cursor control key: KEY_HOME, KEY_END,
KEY_LEFT_ARROW, KEY_RIGHT_ARROW
CC_PLATFORM_MAC, CC_PLATFORM_WIN32, CC_PLATFORM_LINUX default use
cursor( For change state TextFieldTTF::setCursorUse )
Desktop implement dispatchControlKey
WinRT support key repeat and implement dispatchControlKey
Text cursor worked only on LabelType::TTF

new version it: #13600 , for easy merge

PR #15620, opened 2016-05-11
Fix password style text in UTF-16

perminovVS:
Label used UTF-16 in internal.
Need string in UTF-8.

Fix incorrect show password char: ? vs ●.

### DIFF
--- a/cocos/2d/CCTextFieldTTF.cpp
+++ b/cocos/2d/CCTextFieldTTF.cpp
@@ -34,6 +34,7 @@ NS_CC_BEGIN
 
 #define CURSOR_TIME_SHOW_HIDE 0.5f
 #define CURSOR_DEFAULT_CHAR '|'
+#define PASSWORD_STYLE_TEXT_DEFAULT "\xe2\x97\x8f"
 
 static int _calcCharCount(const char * text)
 {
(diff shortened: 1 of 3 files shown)


minggo:
@perminovVS How can it test it?

perminovVS:
@minggo Need create TextFieldTTF and setSecureTextEntry true

minggo:
I tested on iPhone 6, not difference. And the password is not `?` or `*`, it is a rectangle.

perminovVS:
@minggo rectangle as it https://github.com/cocos2d/cocos2d-x/pull/13903 zilongshanren commented on 21 Sep 2015 ? It rectangle bad char.

Password char ●

minggo:
But what's good char? I input `a`, `b` or numbers, also show rectangles.

minggo:
I tested on iPhone6.

perminovVS:
@minggo good char http://www.fileformat.info/info/unicode/char/25CF/index.htm
But now in utf-16 char 0x25CF
Need in utf-8 char 0xE2 0x97 0x8F
This pr make it.

perminovVS:
@minggo @zilongshanren Bad commit https://github.com/cocos2d/cocos2d-x/commit/90087ae5a7db99cd00606bb44aa5a0b2910ae962

_passwordStyleText = {(char)0xe2, (char)0x80, (char)0xa2, (char)0x00}; <---  UTF-8
_passwordStyleText("\u25CF") <--- UTF-16

minggo:
@perminovVS As i comment above. No matter i applied this PR or not, the password is shown as rectangles. Which means the current codes have an issue, but your PR doesn't fix it.

I tested by running `TextInput` in cpp-tests, tested on Mac and iPhone devices.

perminovVS:
@minggo ok, i tested in win32, winRT and android.
I will test on mac and write this.

minggo:
Thanks

perminovVS:
@minggo yes, it char 0xE2 0x97 0x8F not correct show in mac.
Change to old utf-8 char 0xE2 0x80 0xA2
And add test secure text entry

perminovVS:
Win32, WinRT and Adnroid 0xE2 0x80 0xA2 also correct.

minggo:
Thanks, it works as expected.

PR #15639, opened 2016-05-14
Fixed APP_NAME for cpp, cpp-empty, js, and lua tests.

Rubonnek:
Compiling and running the cpp, cpp-empty, js, and lua tests for Linux results in a "No such file or directory" error without this fix.


minggo:
Thanks.

PR #15641, opened 2016-05-14
Add missing references to avoid unnecessary copies

mogemimi:
This PR adds several missing reference types (such as `const &`) to the function declarations for better performance.


minggo:
Thanks.

PR #15645, opened 2016-05-14
Refactor CCFileUtils-apple to remove code duplication

Shulepov:
Refactor CCFileUtils-apple to remove code duplication of conversion from CCValue to NSObject and vise versa.

### DIFF
--- a/cocos/platform/apple/CCFileUtils-apple.mm
+++ b/cocos/platform/apple/CCFileUtils-apple.mm
@@ -52,18 +52,74 @@ private:
     NSBundle* bundle_;
 };
 
+static id convertCCValueToNSObject(const cocos2d::Value &value);
+static cocos2d::Value convertNSObjectToCCValue(id object);
+
 static void addValueToDict(id nsKey, id nsValue, ValueMap& dict);
 static void addObjectToNSDict(const std::string& key, const Value& value, NSMutableDictionary *dict);
+static void addValueToArray(id item, ValueVector& array);
+static void addObjectToNSArray(const Value& value, NSMutableArray *array);
 
-static void addItemToArray(id item, ValueVector& array)
+static id convertCCValueToNSObject(const cocos2d::Value &value)
+{
+    switch (value.getType()) {
+        case Value::Type::NONE:
+            return [NSNull null];
+            
+        case Value::Type::STRING:
+            return [NSString stringWithCString:value.asString().c_str() encoding:NSUTF8StringEncoding];
+            
+        case Value::Type::BYTE:
+            return [NSNumber numberWithInt:value.asByte()];
+            
+        case Value::Type::INTEGER:
+            return [NSNumber numberWithInt:value.asInt()];
+            
+        case Value::Type::UNSIGNED:
+            return [NSNumber numberWithUnsignedInt:value.asUnsignedInt()];
+            
+        case Value::Type::FLOAT:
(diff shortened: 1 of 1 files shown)


cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -52,18 +52,74 @@ void setBundle(NSBundle* bundle) {
...
+            NSMutableDictionary *dictionary = [NSMutableDictionary dictionary];
+            const ValueMap &map = value.asValueMap();
+            for (auto iter = map.begin(); iter != map.end(); ++iter) {
+                addObjectToNSDict(iter->first, iter->second, dictionary);
+            }
+            return dictionary;
+        }
+            
+        case Value::Type::INT_KEY_MAP:
+            break;
```
minggo, line 110: Add default?

cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -99,224 +155,49 @@ static void addItemToArray(id item, ValueVector& array)
...
-        array.push_back(Value(subArray));
-        return;
+        return Value(subArray);
     }
+    
+    return Value::Null;
 }
 
-static void addObjectToNSArray(const Value& value, NSMutableArray *array)
+static void addValueToArray(id item, ValueVector& array)
```
minggo: how about rename `addValueToArray` to `addNSValueToCCVector`?

cocos/platform/apple/CCFileUtils-apple.mm
```diff
@@ -99,224 +155,49 @@ static void addItemToArray(id item, ValueVector& array)
...
-        for (auto iter = valueDict.begin(); iter != valueDict.end(); ++iter)
-        {
-            addObjectToNSDict(iter->first, iter->second, element);
-        }
+    array.push_back(convertNSObjectToCCValue(item));
+}
 
-        [array addObject:element];
-    }
+static void addObjectToNSArray(const Value& value, NSMutableArray *array)
```
minggo: How about rename `addObjectToNSArray` to `addCCValueToNSArray`?

minggo:
And it is good PR.
Thanks.

cocos/platform/apple/CCFileUtils-apple.mm
Shulepov, line 110: Adding 'default' may hide warning when new Value::Type will appear (it should tell that not all branches are checked). For that case there is default return statement at the end of the function.

minggo, line 110: But i think it will cause warnings by the compiler without default. I think we can trigger a log information in default branch.

Shulepov, line 110: No, currently there are no warnings, because all possible branches are handled. E.g. if i remove `case Value::Type::INT_KEY_MAP` then there will appear warning _Enumeration value 'INT_KEY_MAP' not handled in switch_.
 But if there would be `default` statement then that warning will not appear and we will not know that we should separately handle new `Value::Type` (if such would appear, for example something like `Value::Type::SHORT`).
I think compiler warnings is better than log, as it can be seen immediately after Value::Type update. Though currently there are quite much warnings so it may be hidden...

minggo, line 110: Yep, sounds reasonable.

Issue #15439, opened 2016-04-16
EXC_BAD_ACCESS Random crashes on iOS context presentRenderbuffer


Obg1:
I'm getting random crashes at:

```
if(![context_ presentRenderbuffer:GL_RENDERBUFFER])
File: CCEAGLView-ios.mm
```

EXC_BAD_ACCESS (code=1, address 0x1)

OK, I was finally able to identify the cause for it.
The bug is because of the new static __indices in QuadCommand:

```
void QuadCommand::reIndex(int indicesCount)
{
    if (indicesCount > __indexCapacity)
    {
        CCLOG("cocos2d: QuadCommand: resizing index size from [%d] to [%d]", __indexCapacity, indicesCount);
        __indices = (GLushort*) realloc(__indices, indicesCount * sizeof(__indices[0]));
        __indexCapacity = indicesCount;
    }
```

When calling realloc, it moves the memory that is being used by OpenGl, thus causing the BAD ACCESS error.
I was able to "Fix" it by calling realloc with null (will work like malloc):

`__indices = (GLushort*) realloc(nullptr, indicesCount * sizeof(__indices[0]));`

felixjingga:
the more important question is, how do we replicate this bug reliably..? I got this bug too, and can't ever seem to replicate it reliably.. is there any simple way to do so?

Obg1:
I think this bug is related to threading so it is not easy to reproduce it.
But by looking at the code your can notice the realloc function is deleting a pointer that is still being use by previous render commands.

minggo:
> When calling realloc, it moves the memory that is being used by OpenGl, thus causing the BAD ACCESS error.

@Obg1 Good catch. But i am not sure if it is caused by the codes. How did you make sure? But i think we should use `malloc` here since we don't care about the value of `__indices`.

@ricardoquesada Did you know why use `realloc` here?

minggo:
And i think it is a potential issue here as @Obg1 mentioned.

Obg1:
@minggo  my temp fix for it was when allocating a new __indices 
to save the previous allocated __indices without releasing it, it is only release on the next "reIndex" call.
This is not 100% fix, but it did stop the crash from happening. Crash only appears on iOS, but it may cause unknown errors on other platform, it all depends on what the system does with the free memory.

realloc = causes the memory address to be changed, thus causing the pointers to be invalid.

minggo:
[...]

PR #15646, opened 2016-05-15
Fix crash due to reallocation of shared indices memory

Rypac:
This PR provides a fix for issue #15439.

As detailed in the issue, the crash is caused by multiple references to the static `__indices[]` array and the invalidation of these references when this cache is expanded with `realloc()`.

Rather than simply expanding, new memory is allocated to `__indicies[]` and references to any previous `__indicies` are captured by the `QuadCommand` for the duration of its lifetime. The triangle indices used by the `Renderer` are now guaranteed to be valid at the point copying the GL buffer.

The proposed fix will result in a slightly higher amount of memory usage whenever the shared indices buffer grows but will eventually stabilise (assuming no references to the `QuadCommand` are kept from deallocating).

Maybe a larger refactor for dealing with the shared memory should be considered but this is a fairly low impact fix which should at least stop the crashes!

### DIFF
--- a/cocos/renderer/CCQuadCommand.cpp
+++ b/cocos/renderer/CCQuadCommand.cpp
@@ -39,13 +39,18 @@ NS_CC_BEGIN
 int QuadCommand::__indexCapacity = -1;
 GLushort* QuadCommand::__indices = nullptr;
 
-QuadCommand::QuadCommand()
-: _indexSize(-1)
+QuadCommand::QuadCommand():
+_indexSize(-1),
+_ownedIndices()
 {
 }
 
 QuadCommand::~QuadCommand()
 {
+    for (auto& indices : _ownedIndices)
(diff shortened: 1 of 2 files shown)


ricardoquesada:
Thanks!

Rypac:
No probs!

mori1234:
void QuadCommand::reIndex(int indicesCount)
{
    if (indicesCount > __indexCapacity)
    {
		CCLOG("cocos2d: QuadCommand: resizing index size from [%d] to [%d]", __indexCapacity, indicesCount);
		__indices =new (std::nothrow) GLushort[indicesCount];
		_ownedIndices.push_back(__indices);
              //  __indices = new (std::nothrow) GLushort[indicesCount]; // ( up)
		__indexCapacity = indicesCount;
    }

PR #15659, opened 2016-05-17
Use const references whenever possible

mogemimi:
Hello,

When we call a function with its non-[fundamental](https://msdn.microsoft.com/en-us/library/cc953fe1.aspx) parameter, we should pass [by reference](https://msdn.microsoft.com/en-us/library/f0a5bh3e.aspx) instead of by value if available.
This PR changes some parameters to use const reference instead, and avoids needless copying. Also it keeps backward compatibility with the original version.

Thank you for your time.


minggo:
[ci rebuild]

mogemimi:
[ci rebuild]

minggo:
Thanks.

PR #15676, opened 2016-05-18
Use ndk-build with Android Studio.

chaoren:
This feature depends on the latest canary preview of Android Studio. You might want to hold off on this pull request until the feature stabilizes.

### DIFF
--- a/templates/cpp-template-default/proj.android-studio/app/build.gradle
+++ b/templates/cpp-template-default/proj.android-studio/app/build.gradle
@@ -10,14 +10,27 @@ android {
         targetSdkVersion 22
         versionCode 1
         versionName "1.0"
+
+        externalNativeBuild {
+            ndkBuild {
+                targets 'MyGame'
+                arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external'
+                abiFilters 'x86', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
+            }
+        }
     }
 
     sourceSets.main {
         java.srcDir "src"
         res.srcDir "res"
-        jniLibs.srcDir "libs"
         manifest.srcFile "AndroidManifest.xml"
         assets.srcDir "assets"
+    }
+
+    externalNativeBuild {
+        ndkBuild {
+            path "jni/Android.mk"
+        }
     }
 
     signingConfigs {
--- a/templates/cpp-template-default/proj.android-studio/app/build.gradle
+++ b/templates/cpp-template-default/proj.android-studio/app/build.gradle
@@ -15,6 +15,7 @@ android {
             ndkBuild {
                 targets 'MyGame'
                 arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external'
+                arguments '-j' + Runtime.runtime.availableProcessors()
                 abiFilters 'x86', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
             }
         }
--- a/templates/cpp-template-default/proj.android-studio/build.gradle
+++ b/templates/cpp-template-default/proj.android-studio/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
+        classpath 'com.android.tools.build:gradle:2.2.0'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/templates/cocos2dx_files.json
+++ b/templates/cocos2dx_files.json
@@ -1062,6 +1062,7 @@
         "cocos/platform/android/libcocos2dx-with-controller/AndroidManifest.xml", 
         "cocos/platform/android/libcocos2dx-with-controller/build.gradle", 
         "cocos/platform/android/libcocos2dx-with-controller/proguard-rules.pro", 
+        "cocos/platform/android/libcocos2dx/.gitignore", 
         "cocos/platform/android/libcocos2dx/AndroidManifest.xml", 
         "cocos/platform/android/libcocos2dx/build.gradle", 
         "cocos/platform/android/libcocos2dx/proguard-rules.pro", 
--- a/templates/cpp-template-default/proj.android-studio/app/.gitignore
+++ b/templates/cpp-template-default/proj.android-studio/app/.gitignore
@@ -1,2 +1,2 @@
 /build
-/jniLibs
+/.externalNativeBuild
--- a/templates/cpp-template-default/proj.android-studio/gradle/wrapper/gradle-wrapper.properties
+++ b/templates/cpp-template-default/proj.android-studio/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
--- a/tests/cpp-empty-test/proj.android-studio/app/.gitignore
+++ b/tests/cpp-empty-test/proj.android-studio/app/.gitignore
@@ -1,2 +1,2 @@
 /build
-/jniLibs
+/.externalNativeBuild
--- a/tests/cpp-empty-test/proj.android-studio/app/build.gradle
+++ b/tests/cpp-empty-test/proj.android-studio/app/build.gradle
@@ -10,14 +10,27 @@ android {
         targetSdkVersion 22
         versionCode 1
         versionName "1.0"
+
+        externalNativeBuild {
+            ndkBuild {
+                targets 'cpp_empty_test'
+                arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external'
+                abiFilters 'x86', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
+            }
+        }
     }
 
     sourceSets.main {
         java.srcDir "src"
         res.srcDir "res"
-        jniLibs.srcDir "libs"
         manifest.srcFile "AndroidManifest.xml"
-        assets.srcDir "assets"
+        assets.srcDir "../../Resources"
+    }
+
+    externalNativeBuild {
+        ndkBuild {
+            path "jni/Android.mk"
+        }
     }
 
     signingConfigs {
--- a/tests/cpp-empty-test/proj.android-studio/app/build.gradle
+++ b/tests/cpp-empty-test/proj.android-studio/app/build.gradle
@@ -15,6 +15,7 @@ android {
             ndkBuild {
                 targets 'cpp_empty_test'
                 arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external'
+                arguments '-j' + Runtime.runtime.availableProcessors()
                 abiFilters 'x86', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
             }
         }
--- a/tests/cpp-empty-test/proj.android-studio/build.gradle
+++ b/tests/cpp-empty-test/proj.android-studio/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
+        classpath 'com.android.tools.build:gradle:2.2.0'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/tests/cpp-empty-test/proj.android-studio/gradle/wrapper/gradle-wrapper.properties
+++ b/tests/cpp-empty-test/proj.android-studio/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
--- a/tests/cpp-tests/proj.android-studio/app/.gitignore
+++ b/tests/cpp-tests/proj.android-studio/app/.gitignore
@@ -1,2 +1,2 @@
 /build
-/jniLibs
+/.externalNativeBuild
--- a/tests/cpp-tests/proj.android-studio/app/build.gradle
+++ b/tests/cpp-tests/proj.android-studio/app/build.gradle
@@ -11,14 +11,26 @@ android {
         versionCode 1
         versionName "1.0"
 
+        externalNativeBuild {
+            ndkBuild {
+                targets 'cpp_tests'
+                arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external'
+                abiFilters 'x86', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
+            }
+        }
     }
 
     sourceSets.main {
         java.srcDir "src"
         res.srcDir "res"
-        jniLibs.srcDir "libs"
         manifest.srcFile "AndroidManifest.xml"
-        assets.srcDir "assets"
+        assets.srcDir "../../Resources"
+    }
+
+    externalNativeBuild {
+        ndkBuild {
+            path "jni/Android.mk"
+        }
     }
 
     signingConfigs {
--- a/tests/cpp-tests/proj.android-studio/app/build.gradle
+++ b/tests/cpp-tests/proj.android-studio/app/build.gradle
@@ -15,6 +15,7 @@ android {
             ndkBuild {
                 targets 'cpp_tests'
                 arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external'
+                arguments '-j' + Runtime.runtime.availableProcessors()
                 abiFilters 'x86', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
             }
         }
--- a/tests/cpp-tests/proj.android-studio/build.gradle
+++ b/tests/cpp-tests/proj.android-studio/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
+        classpath 'com.android.tools.build:gradle:2.2.0'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/tests/cpp-tests/proj.android-studio/gradle/wrapper/gradle-wrapper.properties
+++ b/tests/cpp-tests/proj.android-studio/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip


minggo:
Thanks @chaoren . Can i debug C++ codes after apply this patch?
And do you know when will the stable version been released?

chaoren:
> Can i debug C++ codes after apply this patch?

Yes. Please try it out and file bugs if you run into any.

> And do you know when will the stable version been released?

I can't give you any specific dates here, but stable is usually about a month or so behind canary.

minggo:
Thanks, i will have a try.

zilongshanren:
[ci rebuild]

minggo:
@chaoren How to test it? I just downloaded Android Studio 2.2, and click `run` menu, it doesn't build native codes, just build java codes and generate the app. And the CPP files are not displayed in Android Studio, then how can i set a breakpoint if we fix compiling issue?

minggo:
Oh, i tested cpp-empty-test, and i found you didn't modify it. I will try by creating a new project.

minggo:
It seems i blocked on importing new generated project

screen shot 2016-05-20 at 6 35 44 pm

TheCodez:
@chaoren I'm getting the following error on windows:
`Error:Process 'command 'C:\NVPACK\android-sdk-windows\cmake\bin\cmake.exe'' finished with non-zero exit value 1`

Any ideas?

chaoren:
@minggo Ah, my apologies, I'll change the tests as well.

@TheCodez currently, you'll need to install this for it to work on Windows: https://www.microsoft.com/en-us/download/details.aspx?id=48145
We're aware of this limitation, and there has been a change that removes this requirement, however, that'll probably not be available until next week or so.

TheCodez:
@chaoren Thanks, but that didn't fix the problem.

[EDIT] I created a new project and now I'm getting a different error:

```
Error:Execution failed for task ':HelloCpp:externalNativeBuildDebug'.
> com.android.ide.common.process.ProcessException: Error while executing 'C:\NVPACK\android-sdk-windows\cmake\bin\cmake.exe' with arguments {--build C:\Users\Michi\Desktop\cpp-template-default\proj.android-studio\app\build\intermediates\cmake\debug\json\armeabi-v7a --target MyGame}
```

chaoren:
@TheCodez are you opening the test projects or creating a new project from template and opening that?

Please provide the steps you performed before reaching that error.

TheCodez:
@chaoren I just created a new project using cocos console, and opened the `proj.android-studio` folder in Android Studio. While doing gradle sync it's encountering the cmake error.

I've attached my log, hope it's somehow useful.
[idea.txt](https://github.com/cocos2d/cocos2d-x/files/275443/idea.txt)

[EDIT] When doing `Build/Clean Project` I get the following output:

```
Information:Gradle tasks [clean]
CMake Error at C:/NVPACK/android-sdk-windows/cmake/android.toolchain.cmake:616 (message):
  Could not find any working toolchain in the NDK.  Probably your Android NDK
  is broken.
  C:/NVPACK/android-sdk-windows/cmake/share/cmake-3.4/Modules/CMakeDetermineSystem.cmake:98 (include)
  CMakeLists.txt:30 (project)
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
CMake Error at C:/NVPACK/android-sdk-windows/cmake/android.toolchain.cmake:616 (message):
  Could not find any working toolchain in the NDK.  Probably your Android NDK
  is broken.
  C:/NVPACK/android-sdk-windows/cmake/share/cmake-3.4/Modules/CMakeDetermineSystem.cmake:98 (include)
  CMakeLists.txt:30 (project)
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
```

minggo:
I met the error when importing a new project:

```
Gradle 'proj.android-studio' project refresh failed
  Error:A problem occurred configuring project ':MyCppGame'.
```

Steps to reproduce:
- use `cocos new -l cpp` to create a new project
- use Android Studio 2.2 preview 1 to import the new project
- then the error happened

I uses latests cocos2d-x codes and apply this PR. And my OS is Mac OS X 10.11.4. The full log file is

[idea.log.zip](https://github.com/cocos2d/cocos2d-x/files/276739/idea.log.zip)

Edit: i installed Android Studio 2.2 preview 1 and Android Studio 2.0.

chaoren:
Could you please make sure that these SDK components are installed and up to date?

SDK Platforms -> Android 5.1 (Lollipop) 22
SDK Tools -> CMake
SDK Tools -> LLDB 2.2
SDK Tools -> NDK

We'll need to change that message to something more useful.

minggo:
It seems LLDB 2.2 is missing. I will try to install it.

minggo:
@chaoren My computer has Cmake, but Android Studio can not find it and ask me to install it. And i can not find a way to set Cmake path.

chaoren:
The CMake integration requires the custom CMake from the SDK, so the CMake on your computer would not work. When this feature stabilizes, we'll hopefully upstream our changes to CMake, but whether the ~20MB of space is worth the additional complexity is debatable.

chaoren:
@TheCodez could you please make sure that you're using the latest NDK?

minggo:
@chaoren Ok, i downloaded the cmake and needed components. And it can compile the codes. But i can not see the log, for example, which cpp file is compiled, and what's the compiling flags and so on as i can before using `cocos command`.

chaoren:
> But i can not see the log, for example, which cpp file is compiled, and what's the compiling flags and so on as i can before using `cocos command`.

Yes, that is the expected behavior right now. We could change the compiling to be more verbose, but we'll have to look at whether the extra output would significantly slow down compilation.

minggo:
@chaoren The app crashed when i clicked `Debug MyCppGame-native`. But it can run of i use `cocos run -p android`, which means the project itself is not problem.

chaoren:
@minggo does clicking just run (without debug) work? Are you running against a physical device or an emulator?

minggo:
No. It can not run no matter debug or not, and no matter select MyCppGame or MyCppGame-native. I tested on Android device(Sangsung S4 Zoom, Android 4.2.2).

minggo:
And i found there is a strange log 
`Connected to the target VM, address: 'localhost:8605', transport: 'socket'`

chaoren:
That's a normal message when attaching the debugger.

When you say "crashed", does it just look like it's hanging, or is it opening the app then immediately closing? Could you please provide the logcat as it crashes?

minggo:
It closed immediately, the logcat is

[logcat.txt.zip](https://github.com/cocos2d/cocos2d-x/files/276966/logcat.txt.zip)

And i found the output `MyCppGame/proj.android-studio/app/build/intermediates/cmake/debug/lib/armeabi/libMyGame.so` is 42.9M, while `MyCppGame/proj.android/libs/armeabi/libMyGame.so` is only 10.5M. I think it is because cmake full link all modules. Instead we use `LOCAL_STATIC_LIBRARIES` instead of `LOCAL_WHOLE_STATIC_LIBRARIES` for most modules in Android.mk.

cezheng:
I installed Android Studio 2.2 Preview 1 on OSX 10.11.4, checked out your branch, then created a new project. But I couldn't sync the gradle file with the project. The IDE simply got an exception.

Here's the Error [Logs](https://gist.github.com/cezheng/6fde572d1b25cab7275c92ca414709a4)
image

chaoren:
@cezheng you get that error when the NDK isn't present, please make sure `SDK Tools -> NDK` is installed and up to date.

TheCodez:
@chaoren I'm having problems getting LLDB 2.2

image

TheCodez:
@chaoren I reinstalled the latest NDK and it's gradle syncing now successful but I still can't compile.I also found out that somehow I can't install LLDB 2.2
image

chaoren:
@TheCodez could you please make sure you're running the canary release, and that `Appearance & Behavior > System Settings > Updates` is pointed to Canary Channel? LLDB 2.2 should be available in canary right now.

Please make sure you have the latest `Android SDK Build Tools` and `NDK` from canary as well, that could explain your compile issue.

chaoren:
@minggo tracking the problem here: http://b.android.com/210932, you can temporarily workaround the issue by either using a newer device (Lollipop+), or by changing the compileSdkVersion to 17 (your device's version).

minggo:
@chaoren Yep, it can work. And i can debug it, great.

minggo:
But as i mentioned above, the output .so file is more larger compared with `cocos command`. And i found Android Studio doesn't record NDK path. For example, i set NDK path with one project, then i import another project, the NDK path is missing, then it will met the error `sync xx failed` i mentioned above. I am not sure if it is because i don't use Android Studio to download NDK, i used an existing NDK.

chaoren:
> the output .so file is more larger compared with `cocos command`.

There are supposed to be two .so files when compiling with CMake, one with full debug symbols that remains on the host, and one with stripped symbols that gets packaged and installed on the device.

Are you comparing the full debug information or stripped .so files?

> And i found Android Studio doesn't record NDK path.

You should be able to choose NDK path via `File > Project Structure > Android NDK location`, but this is a project specific setting. You _should_ be able to set the default NDK path for all new projects via `File > Other Settings > Default Project Structure > Android NDK location`.

minggo:
> Are you comparing the full debug information or stripped .so files?

I am not sure. But the output MyCppGame-debug.apk is very large, it is 42.3M. And the output of `cocos command` is only 3.6M. I know cmake builds 3 architectures, but it is too large.

chaoren:
> I am not sure. But the output MyCppGame-debug.apk is very large, it is 42.3M.

Yes. It looks like we're not stripping the .so file properly before putting it in the APK. I'll verify tomorrow and open a bug to track it.

sachingarg05:
@chaoren Does the new instant run in Android Studio 2.x help with NDK projects? Only copy the new .so file to device and skip all the other build steps?

TheCodez:
@chaoren it's working now, probably due to a new build tool update I installed today. Do you have a code snippet on how to use ndk-build instead?

chaoren:
@sachingarg05 sorry, instant run is not available for external native builds at the moment.
@TheCodez just for you: https://github.com/chaoren/NDK-Build-Sample.

sachingarg05:
@chaoren What do you mean by 'external' native builds? Does this new gradle+cmake native build mean 'internal' native builds, and this supports instant run?

TheCodez:
@chaoren Thanks, I appreciate the help :+1:

chaoren:
@sachingarg05 no, sorry. CMake is a build system external to gradle (dependency management of source files, compiling, etc, is done outside of gradle in a black box), so that's why we're calling it (and ndk-build) "external" native builds right now.

cezheng:
@chaoren So by using CMake won't enable debugging an cocos2d-x application inside Android Studio? What we really wish to have, is being able to modify source code, setting break points etc. all in Android Studio though(just like what we have been doing in Xcode). If just being able to build, I would rather use `cocos compile`

minggo:
@cezheng It can set breakpoint. I succeeded to debug the C++ codes.

cezheng:
Oh, that would be great. Ignore what I have said then

sachingarg05:
@chaoren I hope you can add instant run for NDK builds by checking the datetime stamp of .so files (similar to how compilers look for modified source files for incremental builds). Thanks for the debug/breakpoint patch.

chaoren:
@sachingarg05 yes, we do have plans for that. It's just not available yet.

cezheng:
@chaoren If we setup the external native build to use ndk-build rather than CMake(just like your sample repo: https://github.com/chaoren/NDK-Build-Sample ), will we still be able to debug?

chaoren:
@cezheng yes, you will still be able to debug.

piotrros:
Will it work with precompiled libs?

cezheng:
@chaoren  Then maybe using ndk-build rather than CMake would be a better choice? Its product would remain almost the same as `cocos compile`, which also uses ndk-build.

chaoren:
> Will it work with precompiled libs?

@piotrros not sure what you mean by that.

If you mean if this would work _in addition_ to precompiled libs, then yes, you can debug these libraries built from CMake while your app also uses precompiled libs.

If you mean whether we can debug the precompiled libs _themselves_, then that depends on whether these libs are built with debug symbols and that you have the sources somewhere on your machine. Though, if they are built elsewhere, and the source location doesn't match the debug information, you'll need to inform LLDB of where the sources are located using the source map setting.

@cezheng we have discussed with the cocos team, and our conclusion is that CMake is a better primary choice, since this reduces their burden to maintain an additional build system. This feature is still in canary right now so we can still make changes to address any complaints anyone might have.

tichang99:
> Yes. It looks like we're not stripping the .so file properly before putting it in the APK. I'll verify tomorrow and open a bug to track it.

@chaoren FYI, for the raw libMyGame.so(armeabi) I made, cocos compile vs AS cmake were:
In Debug, 10.2MB vs 40.9MB
In Release,  4.2MB vs 9.8MB

chaoren:
@tichang99 yes, thank you. cocos compile is definitely stripping those shared objects, while our CMake integration does not at the moment.

For reference, here are the numbers I'm seeing right now, it's comparable to yours:

| Build Type | Current Size | Stripped Size |
| --- | --- | --- |
| Debug | 41M | 9.8M |
| Release | 9.9M | 4.5M |

This is from manually running `sdk/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-strip` on the shared libraries. Once we add this, the size shouldn't be a problem.

piotrros:
@chaoren I meant that cocos2d-x have 2 different versions of project: first with cocos2d directory (full source code) and second with precompiled libs linked from /Applications/Cocos directory. First version makes project directory very heavy (around 1.5gb or more) and compilation time is a lot longer. Second version makes project small and compiles fast. However (in xcode for example) you can't debug code outside your own classes (so you can't "go" in Sprite::create method and see what's "inside").

D3vN00b:
@chaoren How do we specify the APP_ABI in the externalNativeBuild/ndkBuild?
I tried `abiFilters 'armeabi-v7a'` but that didn't work.

D3vN00b:
@chaoren, I tried to debug an externalNativeBuild/cmake project on Nexus 7 (Android 6.0.1). The breakpoints for Java are working fine but the C++ break points are not working at all, I tried debug type **Native** and **Hybrid**. The app is running fine on the device. But there are errors in logcat:

`05-26 17:14:27.896 31039-31066/? E/art: Thread attaching while runtime is shutting down: Binder_1
05-26 17:14:39.701 190-587/? E/NetlinkEvent: NetlinkEvent::FindParam(): Parameter 'UID' not found
05-26 17:14:45.885 190-587/? E/NetlinkEvent: NetlinkEvent::FindParam(): Parameter 'UID' not found
05-26 17:15:16.310 32022-32022/? E/Diag_Lib:  Diag_LSM_Init: Failed to open handle to diag driver, error = 2
05-26 17:15:16.310 32022-32022/? E/Sensors: sns_fsa_la.c(386):fsa: fflush failed, 9
05-26 17:15:16.310 32022-32022/? E/Sensors: sns_fsa_la.c(386):fsa: fflush failed, 9
05-26 17:15:16.361 32022-32028/? E/Sensors: sns_debug_main.c(565):Debug Config File missing in EFS!
`

Edit: uploaded file with more info
[debug-errors.txt](https://github.com/cocos2d/cocos2d-x/files/285110/debug-errors.txt)

tichang99:
> @chaoren How do we specify the APP_ABI in the externalNativeBuild/ndkBuild?

@D3vN00b use 
    defaultConfig {
        cmake {
            abiFilters 'x86', 'x86_64', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
        }
   }
This worked for ndkBuild too.

chaoren:
> However (in xcode for example) you can't debug code outside your own classes (so you can't "go" in Sprite::create method and see what's "inside").

@piotrros yes, in general it requires more effort to debug libraries that was built elsewhere, provided that it had been built with debug symbols and that the source files are located on the machine. We use the same debugger as Xcode (LLDB), and the way to do this in LLDB is with the source-map setting. (See the bottom of this page: http://lldb.llvm.org/lldb-gdb.html, the debugger console is available at `Debug > Debugger > LLDB`)

@D3vN00b @tichang99 it should be

```
defaultConfig {
    ndkBuild {
        abiFilters 'x86', 'x86_64', 'armeabi', 'armeabi-v7a', 'arm64-v8a'
    }
}
```

tichang99:
@chaoren It seems it should be. But abiFilters in the ndkBuild block didn't work for me. cmake one did.

chaoren:
@D3vN00b regarding the debugging, could you please submit idea.log here: http://b.android.com/211391.

D3vN00b:
@tichang99 & @chaoren, yes that looks like a bug with the ndkBuild setting. The ndkBuild picks up the abiFilters from the cmake block.

@chaoren, when using the externalNativeBuild's ndkBuild the generated script for the argument to the `...\ndk-build.cmd NDK_PROJECT_PATH= {project path}\proj.android-studio\app\jni\Android.mk ...` is incorrect. The project path points to the Android.mk file, it should be just _{project path}\proj.android-studio\app_.

sachingarg05:
@chaoren With AS 2.1.1 (stable), with instant-run enabled, if I try to run the app again with from Android Studio after doing ndk-build (or cocos compile...) AS says 'Nothing to deploy'.

As you already mentioned, I understand that "fast" instant-run for NDK is still work in progress. I was hoping I can at least get instant-run speed for java/resource changes while NDK build will be done slower but at least work. But it looks like I need to disable instant-run completely to get my NDK changes deployed to device.

Is this the current expected behavior or a possible misconfiguration issue with my environment?

piotrros:
So what's left to do to get it fully working and merging to main cocos2d-x project? Just apk size (stripping symbols)?

minggo:
@piotrros I think there are two things:
- APK size issue, it is important
- use cmake or `ndk-build`
  Use cmake can make all platforms use the same way, now cmake can work on linux, mac, android. And i think it can work on all supported platforms. And use `ndk-build` can keep compatible, and many developers are familiar with `Android.mk` and `Application.mk`. I think we need more discussion. @ricardoquesada , what's your opinion?

ricardoquesada:
@minggo If I understood it correctly, the APK size shouldn't be an issue once `/arm-linux-androideabi-strip` is added.

Adding cmake support, specially if it allows users to debug code from Android Studio is a MUST HAVE feature.
And supporting cmake for Android is not time-consuming for us since we already support cmake.
However, I wouldn't remove Android.mk files since we need to for backward compatibility. Users migrating from v3.xx to v3.12 will need it.

chaoren:
@sachingarg05 that is the current expected behavior, however we are working on what you're asking, and that feature should be available in a future release.

tichang99:
@chaoren Since importing and building entire source tree take too long and slow down IDE, I tried using prebuilt libraries created by 'cocos gen-libs' command and use find_library to link them but cmake couldn't find the libraries. Can you help on this? Followings are cmake commands I used.

```
find_library(COCOS_LIBS
  NAMES "cocos2d"
  PATHS "/Users/tichang/Projects/cocos2d-x/prebuilt/android/${ANDROID_ABI}" )
target_link_libraries(${APP_NAME} ${COCOS_LIBS})
```

I tried libcocos2d for NAMES too, didn't work either.

minggo:
@ricardoquesada but it is not convenient if Android Studio can not do it automatically, and i think @chaoren is fixing it.

I agree that we can keep Android.mk to be compatible.

chaoren:
@tichang99 you need the put `CMAKE_FIND_ROOT_PATH_BOTH` or `NO_CMAKE_FIND_ROOT_PATH` at the end of find_library, e.g., 

```
find_library(COCOS_LIBS
  NAMES "cocos2d"
  PATHS "/Users/tichang/Projects/cocos2d-x/prebuilt/android/${ANDROID_ABI}"
  NO_CMAKE_FIND_ROOT_PATH)
```

It's a weird thing that the toolchain file that we're currently doing does, please see http://b.android.com/211927 for more details.

minggo:
> Adding cmake support, specially if it allows users to debug code from Android Studio is a MUST HAVE feature.

@ricardoquesada ndk-build allows users to debug codes too as @chaoren said though i haven't tried. So i think we should use ndk-build to keep compatible and fix APK size issue. And i looked currently CMakeLists.txt of cocos2d-x, it is complex. I don't think Android developers can use it easily. So i think we can change to use cmake after i refactored it.

TheCodez:
@minggo I aggree with you. Most users can just reuse their Android.mk they were using in Eclipse. Maybe you can add a commandline switch to cocos console which uses an alternative build system in the gradle template file, e.g `cocos new -l cpp -t cmake`

jarsj:
@chaoren Are you going to update tests too ? I tried but couldn't figure out what should be cmake.path.

chaoren:
@jarsj the tests are blocked on this issue: http://b.android.com/210929

minggo:
@chaoren I tried to use `ndk-build` for `cpp-empty-test` but failed. cocos2d-x's Android.mk is more complicated than  https://github.com/chaoren/NDK-Build-Sample. The Android.mk depends on other Android.mk. Will you try to make it work?

Thanks.

minggo:
ping @chaoren

chaoren:
@minggo sorry for the late reply. `ndk-build` integration is not as robust as the CMake integration, since we're actually just parsing the compile output from `ndk-build` behind the scenes. Currently, I don't think recursively included Android.mk files will be displayed in Android Studio, but I do believe the source files listed in them should still show up. I'll have to take another look tomorrow.

minggo:
@chaoren Thanks for the information. And can i know the progress of APK size issue?

chaoren:
@minggo you can track that issue here: http://b.android.com/211072

chaoren:
@minggo I added `ndk-build` integration to cocos in this branch here: https://github.com/chaoren/cocos2d-x/tree/ndk-build, but it depends on the alpha3 release of the gradle plugin which is not available yet. Please give it a week or so until it's available.

minggo:
@chaoren Thanks. I will try it when alpha3 is available.

cezheng:
@chaoren I'm trying out preview 3 and it looks cool. But is there a way to see the current build progress or status in Android Studio? There's no progress bar and the console is printing nothing. I can only tell that the build is progressing by the noice of the fan

chaoren:
@cezheng we're tracking that here: http://b.android.com/210930.

ivalduan:
Is it possible to use prebuilt native libraries for your app with the latest gradle version (and without gradle-experimental)? I can only find examples of all sources building with the app.
If it is so it would be great if you could you point me to an example of that packaging for several ABIs.

chaoren:
@ivalduan Yes, it is. There's some things that we have to sort out first, but you should see examples showing up [here](https://github.com/googlesamples/android-ndk) soon.

cezheng:
@chaoren Is there a ndk version requirement? With this branch merged with the latest v3, I had no problem building with r10e but the Helloworld crashed immediately on launch. Using r11c, cmake exited with 1.

chaoren:
@cezheng the cmake toolchain file that we ship right now is [this one](https://github.com/taka-no-me/android-cmake), which only really supported up to NDK 10. We're in the process of changing that, and the requirement will eventually just be the latest NDK.

cezheng:
@chaoren Thanks. It turned out to be a bug with Android Studio when changing the ndk path in project structure. I recreated the project then set it to r12, it is still building so far.

piotrros:
Hi again, when can we have stable version to finally switch to gradle? Using eclipse is so hard these days :D

minggo:
@chaoren Does `ndk-build` work now?

chaoren:
@minggo Oh, sorry. Forgot to tell you last time, you just needed to supply `NDK_MODULE_PATH`, like this: https://github.com/chaoren/cocos2d-x/commit/66fd7e4fa6c88546d0fb8f20e71016a6e7348cd3. I wonder if you could just put that into the `Android.mk`.

minggo:
@chaoren Thanks, i will try it.

minggo:
@chaoren should i download new version Android Studio? Now i use 2.2 preview 1, and met error when apply this patch.

minggo:
Now i can not use `cocos compile -p android --android-studio` after doing these things:
- checkout a new branch base on v3, and apply this patch
- use Android Studio to open `cpp-tests/prog.android-studio`, and met error
- checkout v3 without this patch, then met error when invoke `cocos compile -p android --android-studio`

The error message is

```
Parsing /Users/minggo/bin/android/adt-bundle-mac-x86_64/sdk/system-images/android-10/x86/package.xml
Observed package id 'system-images;android-10;default;x86' in inconsistent location '/Users/minggo/bin/android/adt-bundle-mac-x86_64/sdk/system-images/android-10/x86' (Expected '/Users/minggo/bin/android/adt-bundle-mac-x86_64/sdk/system-images/android-10/default/x86')
Parsing /Users/minggo/bin/android/adt-bundle-mac-x86_64/sdk/system-images/android-16/armeabi-v7a/package.xml
Observed package id 'system-images;android-16;default;armeabi-v7a' in inconsistent location '/Users/minggo/bin/android/adt-bundle-mac-x86_64/sdk/system-images/android-16/armeabi-v7a' (Expected '/Users/minggo/bin/android/adt-bundle-mac-x86_64/sdk/system-images/android-16/default/armeabi-v7a')

...

Parsing the SDK, no caching allowed
Using incremental javac compilation.
Incremental java compilation is an incubating feature.
Not using incremental javac compilation.
Using incremental javac compilation.
Not using incremental javac compilation.
Not using incremental javac compilation.

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':CppTests'.
> java.lang.NullPointerException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output.

BUILD FAILED
```

Edit: .so is generated correctly, the error happened when generating APK file. And there is not problem before i apply this PR. I doubt if this PR changed some environments that needed by Android Studio.

minggo:
Oh, its my fault, i merged this PR into v3. But this issue exists:

> @chaoren should i download new version Android Studio? Now i use 2.2 preview 1, and met error when apply this patch.

jarsj:
@chaoren @minggo I am testing this on latest Android Studio (Preview 5), build tools 2.2.0-alpha5 and it gives the following error
"Gradle DSL method not found: 'ndkBuild()'".

chaoren:
@jarsj in alpha5, you need to use `defaultConfig.externalNativeBuild.ndkBuild` instead of `defaultConfig.ndkBuild`. I'll update to reflect that shortly.

ricardoquesada:
[ci rebuild]

minggo:
[ci rebuild]

minggo:
@chaoren could you please modify it to support ndk-build? I will merge it after testing. Thanks.

minggo:
ping @chaoren

chaoren:
@minggo do you mean you want to use the ndk-build integration instead of CMake? I have that on a separate branch: https://github.com/chaoren/cocos2d-x/tree/ndk-build.

Are you sure you want to merge it before we reach a stable release?

minggo:
@chaoren yep, many developers need this feature, even use it to debug app. We will describe it in release note. Do you know any known critical issue with current alpha version?

chaoren:
@minggo there's [this issue](http://b.android.com/214626) that affects this project. The fix is not in the current alpha version.

minggo:
@chaoren thanks for the information. This issue is critical to cocos2d-x, because cocos2d-x uses heavily uses `LOCAL_MODULE_FILENAME`. So we have to wait until it is fixed.

ddovod:
Hello guys! Any updates on this issue? Seems there's 2.2 preview 7 available

chaoren:
I updated PR to use beta1. The tests now compile, but don't run on device.
Same thing with the ndk-build branch: https://github.com/chaoren/cocos2d-x/tree/ndk-build

chaoren:
Sorry for the delay. The tests now run correctly on devices for both the cmake integration and the ndk-build integration.

Do you want to try both out before deciding which one to merge?
CMake integration is here: https://github.com/chaoren/cocos2d-x/tree/cmake
ndk-build integration is here: https://github.com/chaoren/cocos2d-x/tree/ndk-build
CMake build is faster right now. We should probably add a `-j` flag to the ndk-build using the number of cores.

TheCodez:
@chaoren I'm getting this error when building a project using ndk-build:
`make (e=87): The parameter is incorrect.`

chaoren:
@TheCodez oh, geez. Looks like we hit the command line length limit on Windows.
Please try adding `'APP_SHORT_COMMANDS=true'` to `arguments` in `app/build.gradle`.
E.g.,

```
defaultConfig {
    ...
    externalNativeBuild {
        ndkBuild {
            ...
            arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external',
                      'APP_SHORT_COMMANDS=true'
        }
    }
}
```

TheCodez:
@chaoren adding this I get the following error doing Gradle sync:

```
Gradle 'proj.android-studio' project refresh failed
Error:@E:\Projects\CocosTest\proj.android-studio\app\.externalNativeBuild\ndkBuild\debug\obj/local/armeabi/objs/cocos2dx_static/archiver.list
```

chaoren:
@TheCodez Yep. That's a bug on our side. The ndk-build integration is not as robust as the CMake integration. We just parse the verbose command line output from ndk-build, so it didn't expect the response file. (Bug filed here: http://b.android.com/220758).

chaoren:
@TheCodez As a silly workaround, you can remove the `APP_SHORT_COMMANDS` and just move the entire project to a shorter path. E.g., `C:\cocos2d-x`.

minggo:
@chaoren thanks, i will merge it when ndk-build is stable. And please change the default type to ndk-build.

kbhulani:
Just tested it out with Android Studio 2.2 Beta 2 and everything works fine! Thanks a lot for making this happen!!!

kbhulani:
@chaoren @minggo  http://android-developers.blogspot.com/2016/09/android-studio-2-2.html. Android Studio 2.2 has been officially released. Any chance of getting this PR merged for the next release?

minggo:
@kbhulani thanks for the information. @chaoren is the stable version fixes previous issues? I will test and merge it if you think it is good to merge now. Thanks.

chaoren:
I discovered some problems when upgrading to stable: http://b.android.com/223083. We'll try to resolve these and get back to you. Thank you for your patience.

madebymozart:
@chaoren Also, Please remember to add a -j option. It's single threaded which will make things super slow

chaoren:
@mozartalouis Please see http://b.android.com/220746. There's no good hard-coded default for the -j flag; it should depend on the host processor. Even -j4 might melt someone's laptop. If the owners really want a specific -j value, then I'll add it.

TheCodez:
@chaoren is http://b.android.com/220758 fixed? Because I'm still experiencing the same issue in stable.

chaoren:
@TheCodez The NDK part of the issue is fixed in r14: https://github.com/android-ndk/ndk/issues/186

templates/cpp-template-default/proj.android-studio/app/build.gradle
```diff
@@ -10,16 +10,29 @@ android {
         targetSdkVersion 22
         versionCode 1
         versionName "1.0"
+
+        externalNativeBuild {
+            ndkBuild {
+                targets 'MyGame'
+                arguments 'NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external'
+                abiFilters 'x86', 'armeabi', 'armeabi-v7a'
```
dumganhar: Since we have supported `arm64-v8a` from v3.13, we need to add `arm64-v8a` to abiFilters.

templates/cpp-template-default/proj.android-studio/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
+        classpath 'com.android.tools.build:gradle:2.2.0-beta3'
```
dumganhar: Android Studio 2.2 has been released, do we need to remove `-beta3` suffix?

gmikhalevskiy:
I use Android Studio 2.2 but only cmake works for me. ndk-build compiles but debug of C++ code is not available for me. What may be wrong ? Thank.

templates/cpp-template-default/proj.android-studio/app/build.gradle
chaoren: Done.

templates/cpp-template-default/proj.android-studio/build.gradle
chaoren: Done, but there's a resource conflict now. See http://b.android.com/223083. Do you actually need both the `.pvr` and the `.pvr.gz` file of the same name?

minggo:
My latest feedback:
- can not view other modules's cpp source code, such as cocos2d-x
- it is hard to stop building, 5 minutes past after i stop building, and haven't stopped
- can not see the detail information about building, for example, i don't know which file is compiling
- it takes a long time to build .so for an architecture

chaoren:
@minggo

> - can not view other modules's cpp source code, such as cocos2d-x

This may be due to how the ndk-build integration works. I don't think the CMake integration would have this problem.

> - can not see the detail information about building, for example, i don't know which file is compiling

See http://b.android.com/210930. The best we can do right now is show which library is compiling.

> - it takes a long time to build .so for an architecture

See http://b.android.com/220746. We can add a default value for the `-j` flag if you want.
CMake integration shouldn't have this problem, FYI.

@kravindran, @jomof, could you guys please also look at this?

SaxonDruce:
Android Studio prompted me to update to NDK 13.0.3315539 today.

After I did that, I got the following error when building my project:

> CMake Error at cocos2d/cocos/CMakeLists.txt:131 (include):
>   include could not find load file:
>     /Users/aaa/Library/Android/sdk/ndk-bundle/build/cmake/AndroidNdkModules.cmake

Looking in the SDK folder, I found AndroidNdkModules.cmake in /Users/aaa/Library/Android/sdk/cmake/

Modifying line 131 from this:

```
include(${__module_path}/AndroidNdkModules.cmake)
```

To this:

```
include(${__module_path}/../../../cmake/AndroidNdkModules.cmake)
```

Seemed to fix the problem.

chaoren:
@SaxonDruce The latest CMake SDK package should have AndroidNdkModules directly in the CMake modules directory, so you could just do:

```
include(AndroidNdkModules)
```

SaxonDruce:
@chaoren Looks like I had a package called "CMake 3.4". I uninstalled that, then installed the "CMake" package, which is 3.6.3155560. Then I changed line 131 to `include(AndroidNdkModules)` as you suggested.

Now I'm getting the following error while building:

> It appears you are cross compiling for  with Clang
> You have changed variables that require your cache to be deleted.
> Configure will be re-run and you may have to reset some variables.
> The following variables have changed:
> CMAKE_C_COMPILER= /Users/aaa/Library/Android/sdk/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-gcc
> CMAKE_CXX_COMPILER= /Users/aaa/Library/Android/sdk/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-g++
> CMAKE_C_COMPILER= /Users/aaa/Library/Android/sdk/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-gcc
> CMAKE_CXX_COMPILER= /Users/aaa/Library/Android/sdk/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-g++
> 
> CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.

Any ideas what would cause this? Are there any other tools / dependencies I'm missing?

Thanks,
Saxon

jomof:
You probably need to manually remove the .externalNativeBuild folder. If
so, it's a bug that we've fixed in an upcoming release.

minggo:
@chaoren any update?

abc20899:
it works well like xcode

piotrros:
Is apk size ok? Is debugging working? Can it be used with precompiled library?

abc20899:
@piotrros
apk size is 10.2 MB，yes it used precompiled library.

chaoren:
@jomof, any progress on these issues?
http://b.android.com/227685
http://b.android.com/227686
http://b.android.com/210930
http://b.android.com/220746

jomof:
- 227685: I read the bug but I don't understand the issue yet. I think if you have a lib module in Android Studio then you can see the sources for that module under that lib?
- 227686 : I thought this was fixed but maybe not for 2.2.X. I'll track it down.
- 210930 : This issue has been resolve in the main Android Studio branch and should be available in the next release from that branch.
- 220746 : I'm hesitant to change the make default for two reasons:
  (1) it is a breaking change
  (2) gradle isn't the right place to make this decision (just like make isn't)
  Only the end user can properly tune this flag. To do this, they can hard code a flag in build.gradle under externalNativeBuild.ndkBuild.arguments or they can pass a flag from the gradle command-line that maps to a flag passed to  externalNativeBuild.ndkBuild.arguments.

jomof:
@minggo Yeah, I think I'll need some help understanding 227685.

minggo:
@jomof I think you can reproduce the issue like
- clone cocos2d-x
- merge this PR
- use Android Studio to open cocos2d-x/tests/cpp-empty-test
- then you can not see the cocos2d-x cpp files in Android Studio

screen_shot_2016-11-16_at_9_31_34_am

jomof:
@minggo, thank you, that's a legitimate bug that has to do with how relative paths are handled. I have a fix to the gradle plugin in-flight

minggo:
@jomof good news. When can i try the new version?

jomof:
@minggo, it will either be AS 2.2.3 or a 2.3 preview. I'm trying for 2.2.3
but it isn't certain. Either way, I'll let you know here.

minggo:
thanks @jomof

formatCvt:
working fine for me with Android Studio 2.3 Canary 2

piotrros:
@minggo how long does it compile? I've ran compilation and after 30mins of waiting I'm out of patience. In gradle console it's working on this:

cocos2d-x-3.13.1/tests/cpp-empty-test/proj.android-studio/app/.externalNativeBuild/ndkBuild/debug/obj/local/armeabi-v7a/libcpp_empty_test.so

chaoren:
@piotrros have you tried adding a `-j` flag (e.g., `'.../external', '-j4'`) here? https://github.com/chaoren/cocos2d-x/blob/ndk-build/tests/cpp-empty-test/proj.android-studio/app/build.gradle#L17
We're tracking that problem here: http://b.android.com/220746.

madebymozart:
@piotrros You can also do this.

```externalNativeBuild {
            ndkBuild {
                targets 'MyGame'
                arguments 'NDK_MODULE_PATH=../../../../:../../../../cocos:../../../../external'
                arguments '-j' + Runtime.runtime.availableProcessors()
                abiFilters 'armeabi-v7a', 'x86'
            }
        } 

Although this will use all of your cpu threads.

**Comment by jomof** [comment]
I'm trying to modify Chaoren's PR to add the -j. Will post back when I have
something.

**Comment by minggo** [comment]
@piotrros sorry, i forgot it.

**Comment by piotrros** [comment]
@charon thanks
I've added: arguments '-j' + Runtime.runtime.availableProcessors()
Now first compilation took 22 minutes. Next compilations are taking about 3min 30 seconds. Still a bit long. 
Anyway debugging is working. 
How can I use prebuilt libs instead? (I mean gradle part, because I know that I have to remove import-add-path's from Android.mk and modify paths in build-cfg.json.

**Comment by piotrros** [comment]
I've figured out to leave only armeabi in ndkBuild. That shortened apk build time and its size. Still final .so file (after extracting apk) is 10.3MB, while old one (console build, old-style without gradle) was 7MB. 
I've tried to build release from console:

cocos deploy -m release -p android --android-studio

Yet I'm getting strange error:

> Creating configuration testReleaseAnnotationProcessor
> Parsing the SDK, no caching allowed
> Using incremental javac compilation.
> Incremental java compilation is an incubating feature.
> Not using incremental javac compilation.
> Using incremental javac compilation.
> Not using incremental javac compilation.
> Not using incremental javac compilation.
> 
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> A problem occurred configuring project ':Music'.
> > java.lang.NullPointerException (no error message)

I've tried adding --debug to cocos command, but I guess it's the wrong place to add this :D

**Comment by jomof** [comment]
I wanted to point out that when you debug from Android Studio it will only
build the .so needed for the current emulator/device so in a typical
code/debug cycle the build should be faster than building APK by the number
of ABIs that are not built.

@mk:
 (1) I don't know why the .so would get bigger. I wonder if somehow symbols
are getting packed. If you still have the larger .so handy could you
try 'readelf
--debug-dump=decodedline xxxx.so'?
 (2) It would be nice to get the full stack trace for that error. I also
don't know how to pass flags to gradle via cocos command-line but it would
be nice to see --stack-trace output.

**Comment by piotrros** [comment]
Okay, after analyzing build_android.py from cocos console I've found out gradle command:

./gradlew --parallel --info assembleRelease

And I've added -s to see stracktrace. Here's it:

> org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':Music'.
>         at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:79)
>         at org.gradle.configuration.project.LifecycleProjectEvaluator.notifyAfterEvaluate(LifecycleProjectEvaluator.java:74)
>         at org.gradle.configuration.project.LifecycleProjectEvaluator.evaluate(LifecycleProjectEvaluator.java:61)
>         at org.gradle.api.internal.project.AbstractProject.evaluate(AbstractProject.java:540)
>         at org.gradle.api.internal.project.AbstractProject.evaluate(AbstractProject.java:93)
>         at org.gradle.execution.TaskPathProjectEvaluator.configureHierarchy(TaskPathProjectEvaluator.java:47)
>         at org.gradle.configuration.DefaultBuildConfigurer.configure(DefaultBuildConfigurer.java:35)
>         at org.gradle.initialization.DefaultGradleLauncher$2.run(DefaultGradleLauncher.java:124)
>         at org.gradle.internal.Factories$1.create(Factories.java:22)
>         at org.gradle.internal.progress.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:91)
>         at org.gradle.internal.progress.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:53)
>         at org.gradle.initialization.DefaultGradleLauncher.doBuildStages(DefaultGradleLauncher.java:121)
>         at org.gradle.initialization.DefaultGradleLauncher.access$200(DefaultGradleLauncher.java:32)
>         at org.gradle.initialization.DefaultGradleLauncher$1.create(DefaultGradleLauncher.java:98)
>         at org.gradle.initialization.DefaultGradleLauncher$1.create(DefaultGradleLauncher.java:92)
>         at org.gradle.internal.progress.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:91)
>         at org.gradle.internal.progress.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:63)
>         at org.gradle.initialization.DefaultGradleLauncher.doBuild(DefaultGradleLauncher.java:92)
>         at org.gradle.initialization.DefaultGradleLauncher.run(DefaultGradleLauncher.java:83)
>         at org.gradle.launcher.exec.InProcessBuildActionExecuter$DefaultBuildController.run(InProcessBuildActionExecuter.java:99)
>         at org.gradle.tooling.internal.provider.ExecuteBuildActionRunner.run(ExecuteBuildActionRunner.java:28)
>         at org.gradle.launcher.exec.ChainingBuildActionRunner.run(ChainingBuildActionRunner.java:35)
>         at org.gradle.launcher.exec.InProcessBuildActionExecuter.execute(InProcessBuildActionExecuter.java:48)
>         at org.gradle.launcher.exec.InProcessBuildActionExecuter.execute(InProcessBuildActionExecuter.java:30)
>         at org.gradle.launcher.exec.ContinuousBuildActionExecuter.execute(ContinuousBuildActionExecuter.java:81)
>         at org.gradle.launcher.exec.ContinuousBuildActionExecuter.execute(ContinuousBuildActionExecuter.java:46)
>         at org.gradle.launcher.exec.DaemonUsageSuggestingBuildActionExecuter.execute(DaemonUsageSuggestingBuildActionExecuter.java:51)
>         at org.gradle.launcher.exec.DaemonUsageSuggestingBuildActionExecuter.execute(DaemonUsageSuggestingBuildActionExecuter.java:28)
>         at org.gradle.launcher.cli.RunBuildAction.run(RunBuildAction.java:43)
>         at org.gradle.internal.Actions$RunnableActionAdapter.execute(Actions.java:173)
>         at org.gradle.launcher.cli.CommandLineActionFactory$ParseAndBuildAction.execute(CommandLineActionFactory.java:239)
>         at org.gradle.launcher.cli.CommandLineActionFactory$ParseAndBuildAction.execute(CommandLineActionFactory.java:212)
>         at org.gradle.launcher.cli.JavaRuntimeValidationAction.execute(JavaRuntimeValidationAction.java:35)
>         at org.gradle.launcher.cli.JavaRuntimeValidationAction.execute(JavaRuntimeValidationAction.java:24)
>         at org.gradle.launcher.cli.ExceptionReportingAction.execute(ExceptionReportingAction.java:33)
>         at org.gradle.launcher.cli.ExceptionReportingAction.execute(ExceptionReportingAction.java:22)
>         at org.gradle.launcher.cli.CommandLineActionFactory$WithLogging.execute(CommandLineActionFactory.java:205)
>         at org.gradle.launcher.cli.CommandLineActionFactory$WithLogging.execute(CommandLineActionFactory.java:169)
>         at org.gradle.launcher.Main.doAction(Main.java:33)
>         at org.gradle.launcher.bootstrap.EntryPoint.run(EntryPoint.java:45)
>         at org.gradle.launcher.bootstrap.ProcessBootstrap.runNoExit(ProcessBootstrap.java:55)
>         at org.gradle.launcher.bootstrap.ProcessBootstrap.run(ProcessBootstrap.java:36)
>         at org.gradle.launcher.GradleMain.main(GradleMain.java:23)
>         at org.gradle.wrapper.BootstrapMainStarter.start(BootstrapMainStarter.java:33)
>         at org.gradle.wrapper.WrapperExecutor.execute(WrapperExecutor.java:130)
>         at org.gradle.wrapper.GradleWrapperMain.main(GradleWrapperMain.java:48)
> Caused by: java.lang.NullPointerException
>         at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:210)
>         at com.android.build.gradle.internal.ndk.NdkHandler.getPlatformVersion(NdkHandler.java:130)
>         at com.android.build.gradle.internal.ndk.NdkHandler.supports64Bits(NdkHandler.java:234)
>         at com.android.build.gradle.internal.ndk.NdkHandler.getSupportedAbis(NdkHandler.java:297)
>         at com.android.build.gradle.tasks.ExternalNativeJsonGenerator.create(ExternalNativeJsonGenerator.java:569)
>         at com.android.build.gradle.internal.TaskManager.createExternalNativeBuildJsonGenerators(TaskManager.java:1225)
>         at com.android.build.gradle.internal.ApplicationTaskManager$11.call(ApplicationTaskManager.java:219)
>         at com.android.build.gradle.internal.ApplicationTaskManager$11.call(ApplicationTaskManager.java:216)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:156)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:120)
>         at com.android.build.gradle.internal.ApplicationTaskManager.createTasksForVariantData(ApplicationTaskManager.java:214)
>         at com.android.build.gradle.internal.VariantManager.createTasksForVariantData(VariantManager.java:485)
>         at com.android.build.gradle.internal.VariantManager$3.call(VariantManager.java:293)
>         at com.android.build.gradle.internal.VariantManager$3.call(VariantManager.java:290)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:156)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:120)
>         at com.android.build.gradle.internal.profile.SpanRecorders.record(SpanRecorders.java:44)
>         at com.android.build.gradle.internal.VariantManager.createAndroidTasks(VariantManager.java:286)
>         at com.android.build.gradle.BasePlugin$11.call(BasePlugin.java:688)
>         at com.android.build.gradle.BasePlugin$11.call(BasePlugin.java:685)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:156)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:120)
>         at com.android.build.gradle.BasePlugin.createAndroidTasks(BasePlugin.java:683)
>         at com.android.build.gradle.BasePlugin$10.call(BasePlugin.java:608)
>         at com.android.build.gradle.BasePlugin$10.call(BasePlugin.java:605)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:156)
>         at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:120)
>         at com.android.build.gradle.BasePlugin.lambda$createTasks$1(BasePlugin.java:603)
>         at org.gradle.internal.event.BroadcastDispatch$ActionInvocationHandler.dispatch(BroadcastDispatch.java:93)
>         at org.gradle.internal.event.BroadcastDispatch$ActionInvocationHandler.dispatch(BroadcastDispatch.java:82)
>         at org.gradle.internal.event.AbstractBroadcastDispatch.dispatch(AbstractBroadcastDispatch.java:44)
>         at org.gradle.internal.event.BroadcastDispatch.dispatch(BroadcastDispatch.java:79)
>         at org.gradle.internal.event.BroadcastDispatch.dispatch(BroadcastDispatch.java:30)
>         at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:93)
>         at com.sun.proxy.$Proxy11.afterEvaluate(Unknown Source)
>         at org.gradle.configuration.project.LifecycleProjectEvaluator.notifyAfterEvaluate(LifecycleProjectEvaluator.java:67)
>         ... 44 more

**Comment by piotrros** [comment]
Ok - after getting this stacktrace I've found out these issues:

https://code.google.com/p/android/issues/detail?id=212688
https://code.google.com/p/android/issues/detail?id=221642

So I've added ndk.dir to local.properties and updated NDK_ROOT and ANDROID_NDK_HOME all with the same path which is in my case: /Users/piotr/sdk/ndk-bundle
Instead of external ndk directory (which had the same version but whatever). 
Anyway bug is fixed in 2.3 as I saw, but this way it's working now.

Building release apk with console gave me the same APK size. I was wondering if maybe apk from android studio haven't symbols stripped. 

So to sum up my questions now are:

1) How to use precompiled libs in this scenario instead of compiling cocos2d directory (which increases compilation time significally)?
2) Why is .so file bigger than the old one? 
@jomof I don't think there's readelf for mac os x so I've uploded libMyGame.so here: http://www6.zippyshare.com/v/3S9z5DKX/file.html 
Can you check this out?
3) What abiFilters are required? App works fine just with armeabi. From the other hand when using just armeabi I wasn't able to debug on my Nexus 6P - I had to add arm64-v8a. Or maybe I should use all of them for debug or just armeabi for release version? How can I set gradle script to work this way?
4) Is build-cfg.json even necessary? I don't see it being used in gradle build.

**Comment by piotrros** [comment]
As for 1) I tried this:
                
> arguments 'NDK_MODULE_PATH="${System.env.COCOS_X_ROOT}/cocos2d-x-3.13.1:${System.env.COCOS_X_ROOT}/cocos2d-x-3.13.1/cocos:${System.env.COCOS_X_ROOT}/cocos2d-x-3.13.1/external:${System.env.COCOS_X_ROOT}/cocos2d-x-3.13.1/cocos/prebuilt-mk'

yet it doesn't work, it's stopping on setup-imports.mk.

**Comment by jomof** [comment]
@mk: readelf returns nothing so symbols are stripped. I'm not sure what
could cause the size difference then. Maybe compile flags changed.


Regarding (1), if I understand the problem you're trying to solve correctly
I think you need to use $(include PREBUILT_SHARED_LIBRARY)
<https://developer.android.com/ndk/guides/prebuilts.html>. NDK_MODULE_PATH
is for location Android.mk imports.

Regarding (3), the simplest scenario is that you don't have any abiFilters.
This will cause all ABIs to be built and packaged into the APK (and will
give you support for the widest set of devices). Regardless of what you set
in abiFilters only the ABI relevant to the current device or emulator will
be built by android studio when you do run or debug.

**Comment by piotrros** [comment]
Thanks for the answer. I've added $(include PREBUILT_SHARED_LIBRARY) in Android.mk file.
Here's current build.gradle:

```
        ext {
            cocospath = "/Users/piotr/Documents/pierdoly/cocos2d-x-3.13.1"
        }

        externalNativeBuild {
            ndkBuild {
                targets "MyGame"
                arguments "NDK_MODULE_PATH=$cocospath:$cocospath/cocos:$cocospath/external:$cocospath/cocos/prebuilt-mk:$cocospath/extensions"
                arguments "-j" + Runtime.runtime.availableProcessors()
                abiFilters "armeabi"
            }
        }
```

And I'm getting the following error, when trying to sync gradle:

> Error:Configuration with name 'default' not found.

**Comment by piotrros** [comment]
Anyone help? @chaoren

**Comment by jomof** [comment]
I don't recognize the error and I don't think it's directly related to
external native. I did find this stackoverflow
<http://stackoverflow.com/questions/22547364/configuration-with-name-default-not-found-android-studio>
that
seems to indicate a missing build.gradle in your setup. Could that apply in
this case?

**Comment by newnon** [comment]
Better solution for this problem https://github.com/cocos2d/cocos2d-x/pull/17100

**Comment by piotrros** [comment]
@jomof it breaks after cleaning project and removing:

$(call import-add-path,$(LOCAL_PATH)/../../../cocos2d)
$(call import-add-path,$(LOCAL_PATH)/../../../cocos2d/external)
$(call import-add-path,$(LOCAL_PATH)/../../../cocos2d/cocos)
$(call import-add-path,$(LOCAL_PATH)/../../../cocos2d/cocos/audio/include)

from Android.mk

Maybe I should add something like this to get prebuilt working? I've opened Android.mk from old android eclipse project and it doesn't contain any of these.

**Comment by piotrros** [comment]
Ok I know what's the origin of this error. In settings gradle there's libcocos2dx project relative path:

project(':libcocos2dx').projectDir = new File(settingsDir, '../cocos2d/cocos/platform/android/libcocos2dx')

I've changed it to: 
project(':libcocos2dx').projectDir = new File(settingsDir, '/Users/piotr/Documents/pierdoly/cocos2d-x-3.14/cocos/platform/android/libcocos2dx')

(yes I've upgraded to 3.14) 
error is still here :(

**Comment by piotrros** [comment]
I did it!: 

project(':libcocos2dx').projectDir = new File('/Users/piotr/Documents/pierdoly/cocos2d-x-3.14/cocos/platform/android/libcocos2dx')

I had to remove "settingsDir" from here. Then it worked.
First compilation took few minutes, but next ones are taking only about 15-45 seconds (if I changed something in cpp files).
Also I can debug :)

**Comment by piotrros** [comment]
I've found some minor issues. 
1) I can't clean my project:
```
Error:rm: fts_read: No such file or directory
make: *** [clean-audioengine_static-armeabi] Error 1
make: *** Waiting for unfinished jobs....

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':Music:externalNativeBuildCleanDebug'.
> Build command failed.
  Error while executing '/Users/piotr/sdk/ndk-bundle/ndk-build' with arguments {NDK_PROJECT_PATH=null APP_BUILD_SCRIPT=/Users/piotr/Documents/projekty/Music/proj.android-studio/app/jni/Android.mk NDK_APPLICATION_MK=/Users/piotr/Documents/projekty/Music/proj.android-studio/app/jni/Application.mk APP_ABI=armeabi NDK_ALL_ABIS=armeabi NDK_DEBUG=1 APP_PLATFORM=android-15 NDK_OUT=/Users/piotr/Documents/projekty/Music/proj.android-studio/app/build/intermediates/ndkBuild/debug/obj NDK_LIBS_OUT=/Users/piotr/Documents/projekty/Music/proj.android-studio/app/build/intermediates/ndkBuild/debug/lib NDK_MODULE_PATH=/Users/piotr/Documents/pierdoly/cocos2d-x-3.14:/Users/piotr/Documents/pierdoly/cocos2d-x-3.14/cocos:/Users/piotr/Documents/pierdoly/cocos2d-x-3.14/external:/Users/piotr/Documents/pierdoly/cocos2d-x-3.14/cocos/prebuilt-mk:/Users/piotr/Documents/pierdoly/cocos2d-x-3.14/extensions -j4 clean}
  Android NDK: WARNING: Unsupported source file extensions in /Users/piotr/Documents/projekty/Music/proj.android-studio/app/jni/Android.mk for module MyGame_shared    
  Android NDK:   ../../../Classes/Definitions.h    
  [armeabi] Clean          : MyGame_shared [armeabi]
  [armeabi] Clean          : audioengine_static [armeabi]
  [armeabi] Clean          : box2d_static [armeabi]
  [armeabi] Clean          : bullet_static [armeabi]
  rm: fts_read: No such file or directory
  make: *** [clean-audioengine_static-armeabi] Error 1
  make: *** Waiting for unfinished jobs....


* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

I have to erase build folder instead.

2) I don't see separate .h files in project explorer. I have for example Definitions.h file for general settings and it's annoying that I can't see it under "Classes".

**Comment by jarsj** [comment]
Do we know when this will be stable enough to get merged ? Is it already stable ?

**Comment by piotrros** [comment]
I sometimes have a feeling that supposed precompiled library isn't really precompiled while using method I've described above. It's just taking a bit too long (compared to the old precompiled cocos studio project) and it's making tons (in matter of size) of files in build directory (5gigs).

**Comment by jomof** [comment]
I believe, and it is unfortunate, that -j4 isn't compatible with 'clean'
target in gnu make. The race sometimes causes crashes such as this. This is
the reason we think we need to add a purpose-built multi-thread flag into
gradle so that it can be used for just the build part and not the clean
part.

**Comment by piotrros** [comment]
I figured out that running clean second time fixes the problem. We should somehow disable multithreading while doing clean.

**Comment by jomof** [comment]
I have a PR in Android Studio gradle plugin to do this. I don't know which
release it will make.

**Comment by minggo** [comment]
@jomof what's the status of Android Studio? Can it test this PR again?

**Comment by jomof** [comment]
Is the only remaining issue at this point the -j / clean problem mentioned
by @mk on January 24th?

**Comment by minggo** [comment]
Does all other issues i reported been fixed?
Such as too big apk size, can not show cpp files and so on.



On Feb 16, 2017, at 12:24 AM, Jomo Fisher <notifications@github.com<mailto:notifications@github.com>> wrote:

Is the only remaining issue at this point the -j / clean problem mentioned
by @mk on January 24th?

**Comment by jomof** [comment]
I've read through the thread again and I believe all issues have been
addressed by Android Studio 2.3 except the following:
(1) Stripping the -j flag didn't make 2.3 but it should make 2.4.
Workaround is that some projects will need to clean a second time.
(2) I don't think there is any closure on the APK size topic. Symbols are
definitely stripped (modulo unknown bugs). I think I'll need more
information to be useful. @piotrros <https://github.com/piotrros> do you
have any news on that front?
I think that's it. I'm not sure who makes the call about when to accept the
PR and what the bar is. If these last two must be resolved then we're not
there yet.

Separately, does anyone have background on why ndk-build was used for this
PR rather than CMake? CMake is generally a stronger build system and it
also happens to not suffer from the -j issue.

**Comment by chaoren** [comment]
> I'm not sure who makes the call about when to accept the
PR and what the bar is.

@jomof I believe @minggo makes the call, and the bar is when all the above issues are resolved.

> Separately, does anyone have background on why ndk-build was used for this
PR rather than CMake?

Yeah, I started the PR with CMake integration (we also agreed that we want to use CMake before the PR), I think there was a some bug with CMake integration that wasn't present in the ndk-build integration (APK size?), and @minggo [requested](https://github.com/cocos2d/cocos2d-x/pull/15676#issuecomment-234180084) that we use ndk-build instead.

**Comment by minggo** [comment]
@jomof i will test it again and merge it if it works. Thanks.

About using ndk-build is because cocos2d-x and many other 3rd party libraries use ndk-build on Android. If we change to use cmake, then we will break compatibility.

**Comment by minggo** [comment]
@jomof but Android Studio 2.3 is not released. I just find 2.2.3 is the latest version.

**Comment by minggo** [comment]
> then you can not see the cocos2d-x cpp files in Android Studio

@jomof this issue still exists in Android Studio 2.2.3. Is it fixed in Android Studio 2.3? And i can not find where to download 2.3.

**Comment by piotrros** [comment]
I'm using AS 2.2.3 and can see all .cpp files. Problem is with extra .h files. I have a file named "Definitions.h", it is in android.mk file, but studio isn't showing it.
As for apk size I think stripping is fine.
As for clean second attempt always works.

I had no other issues and moved all my cocos2d-x projects to android studio.
One thing, which can be annoying is when have something missing in android.mk file. Error is sooo long that it's very hard to find out what is missing. But it's just android studio :) 

Also: some people report having this error: "Error: Configuration with name 'default' not found.". I don't know what causes this, because it's not happening on my device.

**Comment by minggo** [comment]
@piotrros you can see cocos2d-x .cpp files?

**Comment by piotrros** [comment]
@minggo it depends on build.gradle. When I was using cocos2d-x bundled in project I could see all cocos2d-x files. Now I use precompiled library so I naturally don't see them.

**Comment by minggo** [comment]
I mean with this PR, i can not see cocos2d-x files. Does it mean this PR has some problem in build.gradle?

**Comment by piotrros** [comment]
It should be fine. I've modified build.gradle to my own needs. Using precompiled libraries makes app compile in seconds.. By the way it's not fine to use all abiFilters (abiFilters 'x86', 'armeabi', 'armeabi-v7a', 'arm64-v8a'), apk will be huge. I'm personally using armeabi-v7a and can debug c++ code on all my devices.

**Comment by minggo** [comment]
@piotrros yep, we should only build one target. I will merge this PR in 3.15 if possible because we have to modify cocos command to adapt it too. And i am not familiar with build.gradle grammar. I have to learn more about it.

**Comment by jomof** [comment]
2.3 is in beta here: http://tools.android.com/download/studio/canary/latest

**Comment by minggo** [comment]
@jomof i will merge this PR though i can not view the cocos2d-x cpp files in Android Studio. Can you view the cocos2d-x cpp files in Android Studio?

**Comment by jomof** [comment]
@minggo, I verified your steps of 11/15 show cpp files with an earlier 2.3
beta. I'll recheck with the most recent beta. Monday is a Google holiday,
but I'll check Tuesday

**Comment by minggo** [comment]
@jomof thanks. I haven't tried 2.3 since it is not stable.

**Comment by minggo** [comment]
@jomof 
```
sourceSets.main {
        java.srcDir "src"
        res.srcDir "res"
        jniLibs.srcDir "libs"
        manifest.srcFile "AndroidManifest.xml"
        assets.srcDir "../../../src"  // how can include "src" folder name in "assets"
    }
```
I am doing lua project setting for Android Studio as mentioned above. I want to include `src` folder name in `assets`. How can i do it?

**Comment by minggo** [comment]
I have to do like this to delete unneeded folders, but is there any other way to exclude them?
```
android.applicationVariants.all { variant ->
    def rmassets = task("delete${variant.name}Assets", type: Delete) {
        delete "${buildDir}/intermediates/assets/${variant.dirName}/project"
        delete "${buildDir}/intermediates/assets/${variant.dirName}/simulator"
    }
    variant.mergeAssets.finalizedBy rmassets
}
```

jomof:
I don't directly know how to solve that problem but Gradle's FileCollection
is typically how you build sets of files with exclusions.

minggo:
How to exclude resources? I tried `assets.exclude` or `assets.filter.exclude`, but can not work.

minggo:
@jomof I updated Android Studio to 2.3, can not see cocos2d-x c++ source codes.

jomof:
That's not good to hear, are you following the steps from 11/15/16 in this
thread? (Aside from merge the PR)

minggo:
I tested with latest codes which includes this PR. You can have a try. And the steps is the same as 11/15/16 in this thread.

jomof:
I'm having trouble just building the latest project with Gradle. Downstream
this would cause the files not to show up in Android Studio, but am I doing
this wrong?

$ git clone https://github.com/cocos2d/cocos2d-x.git
$ cd cocos2d-x/tests/cpp-empty-test/proj.android-studio
$ ./gradlew assemble

../../../../cocos/./Android.mk:312: *** Android NDK: Aborting.    .  Stop.
:CppEmptyTest:generateJsonModelDebug FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':CppEmptyTest:generateJsonModelDebug'.
> Build command failed.
  Error while executing process
/usr/local/google/home/jomof/Android/Sdk/ndk-bundle/ndk-build with
arguments {NDK_PROJECT_PATH=null
APP_BUILD_SCRIPT=/usr/local/google/home/jomof/projects/cocos2d-x/tests/cpp-empty-test/proj.android-studio/app/jni/Android.mk
NDK_APPLICATION_MK=/usr/local/google/home/jomof/projects/cocos2d-x/tests/cpp-empty-test/proj.android-studio/app/jni/Application.mk
APP_ABI=armeabi NDK_ALL_ABIS=armeabi NDK_DEBUG=1 APP_PLATFORM=android-9
NDK_OUT=/usr/local/google/home/jomof/projects/cocos2d-x/tests/cpp-empty-test/proj.android-studio/app/build/intermediates/ndkBuild/debug/obj
NDK_LIBS_OUT=/usr/local/google/home/jomof/projects/cocos2d-x/tests/cpp-empty-test/proj.android-studio/app/build/intermediates/ndkBuild/debug/lib
NDK_TOOLCHAIN_VERSION=4.9 APP_PLATFORM=android-13
NDK_MODULE_PATH=../../../..:../../../../cocos:../../../../external -j12
APP_SHORT_COMMANDS=false LOCAL_SHORT_COMMANDS=false -B -n}
  Android NDK:
/usr/local/google/home/jomof/Android/Sdk/ndk-bundle/sources/android/cpufeatures/Android.mk:
Cannot find module with tag 'freetype2/prebuilt/android' in import path
  Android NDK: Are you sure your NDK_MODULE_PATH variable is properly
defined ?
  Android NDK: The following directories were searched:
  Android NDK:         ../../../.
  ./cocos/./Android.mk:312: *** Android NDK: Aborting.    .  Stop.


* Try:
Run with --stacktrace option to get the stack trace. Run with --info or
--debug option to get more log output.

BUILD FAILED

minggo:
After updating to latest gradle version and build tool version. The issue is fixed. Thanks.

TheProphetOfRa:
@jomof Did the issue with the path lengths on Windows get resolved? The latest version of AS works perfectly on macOS but when trying to compile on Windows I get path length errors. When setting APP_SHORT_COMMANDS and LOCAL_SHORT_COMMANDS to true in my build.gradle if I run gradle in debug it shows that after I assign them to true they get assigned back to false.

jomof:
Hi David,
At this point I think all the known Windows path length issues are resolved
so, depending on your environment, your issue may be something I haven't
seen before. Could you open a bug at b.android.com?

Regarding APP_SHORT_COMMANDS and  LOCAL_SHORT_COMMANDS. We call ndk-build
twice:

(1) The first time is with -nB to create a build script that we analyze for
metadata about the build to give to Android Studio. This is the where you
see us setting these flags to false. Gnu make should not be accessing the
filesystem during this phase so it should be safe with respect to path
lengths. You can see the exact command we use for this by opening
ndk_build_command.txt. The metadata we produce is next to this file in
android_gradle_build.json.

(2) Later, during the build phase, we call ndk-build again without
resetting those flags. You can see the exact build command we will use by
opening android_gradle_build.json in an editor and looking at
libraries.{target}.buildCommand in the JSON. You should see only your use
of xxx_SHORT_COMMANDS there.


I'm sorry you still have path issues on Windows. I'll fix it if it is
possible to do so. Please provide as much information as you can in the bug
and CC my email (jomof@google.com).
Thanks!
Jomo

minggo:
Hi all, could you please discuss Android Studio issue in the forum? I already created [a ticket](http://discuss.cocos2d-x.org/t/android-studio-support/35364/70). And i will lock this thread since there are too many conversations to open this page.

PR #15683, opened 2016-05-19
Enable WebView's local storage on Android

rynkowsg:
Problem:
cocos2d::experimental::ui::WebView' local storage wasn't working on
Android. On iOS worked fine.

Solution:
Enable local storage of android.webkit.WebView in Android.

References:
https://developer.android.com/reference/android/webkit/WebSettings.html
http://stackoverflow.com/questions/5899087/android-webview-localstorage


minggo:
Thanks.

PR #15689, opened 2016-05-20
Suspendable and configurable Downloader and AssetsManagerEx

cezheng:
## This PR does the following things
- Added suspend/resume methods to Downloader for apple & android
- Added suspend/resume methods AssetsManagerEx
- Made AssetsManagerEx more configurable
  - configurable storage path
  - configurable downloader hints
- Optimized apple downloader, putting expensive tasks creation & I/O operations into a background queue to avoid UI blocking
- Slightly improved the performance of generating manifest diff for AssetsManagerEx(around 10% ~ 20%)
- Refactored Cocos2dxDownloader.java, correcting the coding style inconsistency 

Detailed description is written in each commit message.
## TODO
- [ ] Add tests for suspend/resume
- [ ] Add implementation for curl downloader (currently it is assert(false)) ?
- [ ] js & lua bindings ?

@minggo Would you mind discussing how we should proceed on these TODOs? Thanks:)


cezheng:
Windows platform seems to be having problems linking to this method, I'm not sure why this happens only on Windows. Do I need to change any build settings?

``` cpp
const AssetsManagerEx::Config& AssetsManagerEx::Config::defaultConfig() {
    //...
}
```

update:
Since it might be a bug of the VS's compiler, I removed the default parameter value and take another approach by overloading the constructor, and removed the defaultConfig method from the header, making it static. Let's see if this time the Windows build can pass

``` cpp
static const AssetsManagerEx::Config& defaultConfig() {
    //...
}
```

cezheng:
@minggo I'm sorry I don't have Windows environment so I can't figure out why a linker error only happens on Windows build(The related method is in AssetsManagerEx.cpp which is shared among all platforms). I would appreciate it if you could take a look on the error on Windows when you have time. Thanks.

minggo:
@cezheng Thanks for the PR. I think i need sometime to review the PR.

extensions/assets-manager/Manifest.h
```diff
@@ -140,10 +140,21 @@ class CC_EX_DLL Manifest : public Ref
...
+                   const std::function<void(const std::string&)>& onUnchanged) const;
+
     /** @brief Generate resuming download assets list
-     * @param units   The download units reference to be modified by the generation result
+     * @param units               The download units reference to be modified by the generation result
+     * @param getAssetStoragePath Function used to calculate the storage path of each asset
      */
-    void genResumeAssetsList(DownloadUnits *units) const;
+    void genResumeAssetsList(DownloadUnits *units,
+                             const std::function<std::string(const std::string& key, const Asset& asset)>& getAssetStoragePath) const;
```
minggo, line 157: break compatibility?

extensions/assets-manager/Manifest.cpp
```diff
@@ -206,18 +206,76 @@ std::unordered_map<std::string, Manifest::AssetDiff> Manifest::genDiff(const Man
...
 
-void Manifest::genResumeAssetsList(DownloadUnits *units) const
+void Manifest::yieldDiff(const Manifest *b,
+                         const std::function<void(const std::string&, const AssetDiff&)>& onDiff,
+                         const std::function<void(const std::string&)>& onUnchanged) const
+{
+    const std::unordered_map<std::string, Asset> &bAssets = b->getAssets();
+
+    std::unordered_map<std::string, Asset>::const_iterator valueIt, it;
+    for (it = _assets.begin(); it != _assets.end(); ++it)
```
minggo: How about use [range for](http://en.cppreference.com/w/cpp/language/range-for)?

extensions/assets-manager/Manifest.h
cezheng, line 157: This is a method that is only used internally by AssetsManagerEx. With the default AssetsManagerEx::Config set, it will behave exactly the same as it used to be.

https://github.com/cocos2d/cocos2d-x/pull/15689/files#diff-3d58187924dcf35ee09bd7de35f8839fR59

extensions/assets-manager/Manifest.cpp
```diff
@@ -206,18 +206,76 @@ std::unordered_map<std::string, Manifest::AssetDiff> Manifest::genDiff(const Man
...
+        const auto &valueA = it->second;
+
+        // Deleted
+        valueIt = bAssets.find(key);
+        if (valueIt == bAssets.cend())
+        {
+            AssetDiff diff;
+            diff.asset = valueA;
+            diff.type = DiffType::DELETED;
+            onDiff(key, diff);
```
minggo: Check `onDiff` since it may be nullptr as you commented?

extensions/assets-manager/AssetsManagerEx.h
```diff
@@ -67,16 +67,33 @@ class CC_EX_DLL AssetsManagerEx : public Ref
...
     
     /** @brief Create function for creating a new AssetsManagerEx
      @param manifestUrl   The url for the local manifest file
      @param storagePath   The storage path for downloaded assets
+     @param config        AssetsManagerEx configuartion
      @warning   The cached manifest in your storage path have higher priority and will be searched first,
                 only if it doesn't exist, AssetsManagerEx will use the given manifestUrl.
      */
-    static AssetsManagerEx* create(const std::string &manifestUrl, const std::string &storagePath);
+    static AssetsManagerEx* create(const std::string &manifestUrl,
```
minggo, line 98: It is better not use default parameter to obey [cocos coding style](https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#default-arguments).

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -494,7 +530,12 @@ void AssetsManagerEx::downloadManifest()
     if (_updateState != State::PREDOWNLOAD_MANIFEST)
         return;
 
-    std::string manifestUrl = _localManifest->getManifestFileUrl();
+    std::string manifestUrl;
+    if (_remoteManifest->isVersionLoaded()) {
+        manifestUrl = _remoteManifest->getManifestFileUrl();
+    } else {
+        manifestUrl = _localManifest->getManifestFileUrl();
+    }
```
minggo: Fix bug? If so, i think it is better to be done in a new PR, or the PR is too complex.

extensions/assets-manager/AssetsManagerEx.h
```diff
@@ -67,16 +67,33 @@ class CC_EX_DLL AssetsManagerEx : public Ref
...
+         */
+        std::function<std::string(const std::string& key, const Manifest::Asset& asset)> getAssetStoragePath;
+
+        /** @brief  Hints for configuring downloader
+         */
+        network::DownloaderHints downloaderHints;
+
+        /** @brief  Default configuration for AssetsManagerEx
+         */
+        static const Config& defaultConfig();
```
minggo: defaultConfig -> getDefaultConfig? because most of cocos2d-x function names starts with a verb.

extensions/assets-manager/Manifest.h
minggo, line 157: Ok.

minggo:
@cezheng Could you please remove codes that fixes bugs and send new PRs for them? Or it is hard to review the codes. Thanks.

extensions/assets-manager/AssetsManagerEx.cpp
cezheng: Um, it's hard to say if this is a bug. It is rather about how it should be.

If we have downloaded the new version file, should we use the manifestUrl of one in the new version file or the one already exists locally?

minggo: @pandamicro could you please take a look? I don't know the original design and the logic that it may affect.

cezheng: Let me explain why we are making this change.

Each member of our team has their own set of assets for development. For example, we separate them by url in this format:

> https://host/<user-name>/xxx/xxx/xxx.png

And we have a master branch of assets 

> https://host/master/xxx/xxx/xxx.png

In the development build, we can switch between assets branches by making AssetsManagerEx update via a different url. But the original implementation won't use the url in the newly downloaded version file, making this switch impossible.

cezheng: Let me explain why we made this change.

Since we will have a initial version file bundled with the app, the first time we come to the download process, the url of the remote manifest might be different from the one in the bundled version file.

For the original logic, the AssetsManagerEx will just download from the url of the old version manifest, which is incorrect.

With this change, it will download manifest from the url in the newly downloaded version file.

cezheng:
@pandamicro @minggo It is now here https://github.com/cocos2d/cocos2d-x/pull/15699

minggo:
@cezheng thanks. Are all the codes in this PR related with `Suspendable and configurable Downloader and AssetsManagerEx`. Please remove all other codes if it has.

cezheng:
@minggo Logically no, if you don't put code refactoring or optimization into account.

minggo:
@cezheng I think it is better to put refactoring and optimization in other PRs is better. It will make my life easier. Thanks.

cezheng:
@minggo But it is really hard to divide them. Since much of the refactoring is needed only because it makes the code better if we are implementing `Suspendable and configurable Downloader and AssetsManagerEx`. ( which means if we are not implementing this functionality, there's no reason to make many of those refactorings)

I have written detailed description under each commit of this PR, and each commit is really small. I would recommend you to review by commit.

cezheng:
@minggo For example this commit
https://github.com/cocos2d/cocos2d-x/pull/15689/commits/2a498329b8566ac7d57e98fbfd8e3bc3d441ee43

It is really impossible to remove the refactoring from it. And I personally believe the refactoring makes the code easier to read. I have been dividing original methods into new methods, because some of them are needed for the new functionality, and it makes the original method shorter and clearer.

I just don't think first implementing a new functionality in a ugly way and with more code, then make another PR to refactor it would be a very good idea.

minggo:
Ok.

cezheng:
@minggo 

After some more iterations of this version of modified AssetsManagerEx on our app, I guess the current assets manager is still consuming so much CPU resources, that it is hardly satisfiable without a thorough rewrite.

## The problems include

### CustomEvents are heavy
- It is using CustomEvent to notify progress & state of the manager, which are dispatched to main thread. Especially the `onProgress` callback happens really frequently(once when a chunk of data is read from the HTTP data stream), that it is making the main thread really heavy, and FPS is greatly affected. 
- However it is the developer that decide whether the app should display the progress(if they are just downloading some assets in background, they don't need to display the progress at all), or how frequent they are refreshing the progress display(if they only want it refreshed once every second).

### AssetsManagerEx/Downloader is not multi-thread friendly
- Basically, all the operations that are not related to UI should be done by a background thread. But the callbacks of the Downloader/AssetsManagerEx are all performed in main thread, and it is not configurable.

### Download rate is not adjustable

Let's say we have a RPG game like this
1. We launch the game then we download assets for Scene A
2. Download completes then we load assets and launch Scene A
3. We can go to Scene B from Scene A, in order to reduce the waiting time, we start to pre-download assets in a background thread when the player is in Scene A. But we do not want to consume too much CPU resources so we apply a limit on the downloader.
4. Before the downloads finishes, the player goes to Scene B. Then we shows the loading UI, and displays the downloading progress of the background task.
5. Since we are showing the progress at foreground, we do not want to apply the limit on the downloader any more.
6. Sadly we cannot adjust the download rate of the downloader at the moment.

## Have to rewrite

So for our scenario, we finally draw to the conclusion that the AssetsManagerEx/Downloader do not meet our needs, due to the fact that it heavily relies on CustomEvents. We have to rewrite it in this way:
1. Dispatch no CustomEvents. Use callbacks from another thread to notify only important events(error, success), and provide getters to let developer to retrieve progress info at their needs.
2. Make downloader able to adjust its download rate.
3. Making all callbacks configurable on whether they should be executed on main thread.
4. Of course, they should still be able to suspend/pause.

minggo:
@pandamicro What's your opinion?
@cezheng Thanks for the suggestion. Will you continue to improve it?

cezheng:
@minggo Since our app has some deadlines at the moment, I think I will put my code into another class to make sure it is working at least with our app. After that, I would consider cleaning it up and see if it can be added to cocos2d-x.

minggo:
@cezheng thanks

cezheng:
@minggo @pandamicro 
There's just one thing in the current Downloader that really confuses me

**_Why does it have a `createDataTask` method?**_

Why do we even need this anyway? If what we want to get from a url is not something we want to save in a file, why not use HttpClient?

extensions/assets-manager/AssetsManagerEx.h
```diff
@@ -67,17 +67,38 @@ class CC_EX_DLL AssetsManagerEx : public Ref
...
+    /** @brief Create function for creating a new AssetsManagerEx
+     @param manifestUrl   The url for the local manifest file
+     @param storagePath   The storage path for downloaded assets
+     @param config        AssetsManagerEx configuartion
+     @warning   The cached manifest in your storage path have higher priority and will be searched first,
+                only if it doesn't exist, AssetsManagerEx will use the given manifestUrl.
+     */
+    static AssetsManagerEx* create(const std::string &manifestUrl,
+                                   const std::string &storagePath,
+                                   const Config& config);
```
pandamicro, line 100: JS / Lua manual binding for this function and the constructor is needed

pandamicro:
@cezheng Thanks for the hard work, 

> 1. Dispatch no CustomEvents. Use callbacks from another thread to notify only important events(error, success), and provide getters to let developer to retrieve progress info at their needs.
> 2. Make downloader able to adjust its download rate.
> 3. Making all callbacks configurable on whether they should be executed on main thread.
> 4. Of course, they should still be able to suspend/pause.

These are indeed good improvements to AssetsManagerEx.

There would be some compatibility issues, for example, user won't be notified for progression update anymore, but it won't crash user code so I think it's acceptable if we cover it well in the documentation.

@minggo For this PR, I think suspend/resume are really important features that user would love, the configurable feature doesn't break compatibility on API level, so I think we can accept.

@cezheng I want to discuss about the API design for configurable AssetsManagerEx, now you made a new parameter on create/constructor of AssetsManagerEx, but after that, user won't be able to change any of the configuration via any APIs, what do you think about additional APIs like:
- `getDownloaderHints`
- `setDownloaderHints`
- `getAssetStoragePathGetter`
- `setAssetStoragePathGetter`

cezheng:
@pandamicro 
Thanks.

TBH even myself ain't totally happy with this PR's implementation.

I'm currently working on a version based on the original one.
- For the downloader hints: I decided to provide a getter of the downloader(will remove downloaderHints from Config struct) and let the user configure the downloader directly on that, so that we don't have to wrap those downloader APIs on AssetsManager once more.
- For the assetStoragePathGetter, I think since currently one AssetsManagerEx instance can only be used once, if we allow the user to modify this value, it may cause inconsistency problems during the process of updating.

Also, I've been using std::shared_ptr & std::unique_ptr instead of using the cocos2d-x's auto release pool. But I'm not sure if this will cause any problem when binding them to lua/js.

Currently I'm just making it for the app of my employer. 
Because the way I would like the AssetsManager to be is quite different, I'm putting it in cocos2d::experimental. Also, since the some problems lies in Downloader, and experimental downloader will also be put in cocos2d::experimental.

I will share the code once I'm done.

pandamicro:
> For the downloader hints: I decided to provide a getter of the downloader(will remove downloaderHints from Config struct) and let the user configure the downloader directly on that, so that we don't have to wrap those downloader APIs on AssetsManager once more.

That would be better

> For the assetStoragePathGetter, I think since currently one AssetsManagerEx instance can only be used once, if we allow the user to modify this value, it may cause inconsistency problems during the process of updating.

ok, I understand

@cezheng @minggo I propose we do the following:
1. I will bind Downloader in v3.12
2. We can wait the experimental version of @cezheng 
3. Then discuss the configurable AssetsManagerEx API based on that

What do you guys think

cezheng:
image
@pandamicro would you tell me why we need the dataTask method? It seems pretty redundant to me, since HttpClient can just do the same thing. Since removing it can greatly simplify the code structure, I'm removing it in the experimental version of Downloader.

pandamicro:
I'm not very familiar with network implementation, maybe @minggo or @yangws can tell you

minggo:
I don't know the original design of it. I can ask @yangws to take a look, he implemented it.

minggo:
> Also, I've been using std::shared_ptr & std::unique_ptr instead of using the cocos2d-x's auto release pool. But I'm not sure if this will cause any problem when binding them to lua/js.

If it is used in logical internally, there is not problem. If it is the object retain to lua/js, then it will have problem.

yangws:
NSURLSession APIs has NSURLSessionDataTask, createDataTask is a wrapper for it.
For downloading data to mem use HttpClient or createDataTask both OK.

cezheng:
@minggo So if a class is to be used in lua/js, it has to inherit Ref, is that correct?

cezheng:
@yangws  Thanks. Then I think it's ok to remove it. Since NSURLSession can be used for general network purposes, not only for downloading files, it is not really necessary to wrap it in Downloader.

Also, since NSURLConnection is deprecated since iOS 9, sooner or later HttpClient's implementation will also be rewritten in NSURLSession I believe. (Popular library AFNetworking has already removed its NSURLConnection APIs last year and only uses NSURLSession)

cezheng:
Just to share 2 more problems I found
1. On older versions of Android, the FileUtils::createDirectory is very expensive. And AssetsManagerEx is calling this method for every diff it finds.
   https://github.com/cocos2d/cocos2d-x/blob/v3/extensions/assets-manager/AssetsManagerEx.cpp#L607
   For a manifest of 1500 files, it took a API 14 Samsung Galaxy III 30+ second to check & create directories. Which should be avoided.
2. Android downloader is using the Range header simply using the size of the temp file. If the file on remote has changed, this will cause a 416 Range not satisfiable HTTP response exception, and downloader in this state is not recoverable from the exception. We should delete the temp file if we have 416 status code. iOS does not have this problem because NSURLSession's resume data is already taking care of it.

minggo:
@cezheng As you said you will refactor the PR. So will you send a new PR or just reuse this one? Please let me know what should i do with this PR?

Thanks.

cezheng:
@minggo I have made a simplified version of the current Downloader/AssetsManagerEx in our own project, and we are waiting to see if we have any problems. Currently, the code is not clean enough to be put in a PR yet but I will do it soon. For this PR I think it is OK to close for now.

minggo:
@cezheng Got it, thanks.

pandamicro:
Sorry for being so late, but I recently refactored our AssetsManagerEx in #17115 , a brief info about the upgrade in described in the PR, @cezheng can you take a look ? If you can re patch your improvements onto AssetsManagerEx, it would be very nice.

I think suspend/resume, configurable and the optimization are great features

cezheng:
@pandamicro 

I added these features to my previous employer's code base, adding a bunch of new features(not blocking main thread any more, suspend/resume, low-priority-mode for iOS(for power saving, etc.) & redesign some APIs(better progress tracking, more flexible task management, etc.), but I made another class based on the original code. Due to our preference of std::shared_ptr & std::unique_ptr to cocos2d::Ref, I did not use any cocos2d::Ref in my code, so there might be issues with lua/js binding.

But after all, I do not have access to the code any more. Forgive me for leaving in such a hurry. You might consider contacting my previous manager @kiyoto-suzuki to see if he can share that part of code with you.

Cheers :beers:

pandamicro:
I see, thank you very much for your effort anyway, and good luck for your new career ~
I think I will reimplement these features later based on this PR.

PR #15690, opened 2016-05-20
change if (fontName.c_str()) to if(!fontName.empty())

halx99:
fix warnning, change if (fontName.c_str()) to if(!fontName.empty()), The first expression's value always true.

### DIFF
--- a/cocos/platform/win32/CCDevice-win32.cpp
+++ b/cocos/platform/win32/CCDevice-win32.cpp
@@ -99,7 +99,7 @@ public:
 
     }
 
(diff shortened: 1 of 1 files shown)


minggo:
Thanks.

minggo:
But how about change const chart\* to const std::string&?

halx99:
I think chagne to std::string is better.

halx99:
Maybe const char\* is ok, because, there is always needs copy input parameter.

minggo:
Thanks.

PR #15699, opened 2016-05-23
use manifestUrl from remote version

cezheng:
In case if local version file is very old, and remote version has changed the manifest url ever since, this fix will make sure that AssetsManagerEx will download the correct manifest file.

This PR is moved from https://github.com/cocos2d/cocos2d-x/pull/15689#issuecomment-220927212


pandamicro:
Thanks, reasonable fix, I'm ok with this PR @minggo

PR #15712, opened 2016-05-24
Use nullptr instead of 0 or NULL

mogemimi:
In C++11 and later, we should use `nullptr` instead of `0` or `NULL` if possible.  So this PR changes some null pointer constants to `nullptr` and makes them consistent with the [coding style](https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#0-and-nullptrnull).


minggo:
Thanks @mogemimi

PR #15716, opened 2016-05-25
correct assign to equal in cert trust expression

cezheng:
This PR simply fixes an obvious bug which uses assign(=) in a boolean expression


minggo:
Yep, it is a bug.
Thanks.

PR #15718, opened 2016-05-25
Fix some small issues

pandamicro:


pandamicro:
- Update network licenses and reformat: Synced some changes from cocos2d-x-lite, so that we can merge from each other easier
- Update bindings generator: Fix bindings generator simple test
- Fix Component event issue: Component don't have schedules and actions
- Fix android compilation issue with thumb mode: editor support modules are still causing `relocation overflow in R_ARM_THM_CALL` issue

cezheng:
Code format is definitely a problem. Our team is using [clang-format](http://clang.llvm.org/docs/ClangFormat.html) to deal with it.

Just define a .clang-format file at the root of the repo.

There is also a [Xcode Plugin](https://github.com/travisjeffery/ClangFormat-Xcode), you can make it auto format the source file each time use `save` in Xcode

pandamicro:
Thanks for the suggestions, @minggo what do you think about forcing team member to use a code formatting tool ?

pandamicro:
@minggo Besides, things need to be included if we release a second v3.11 version:
1. web engine and bindings generator ref
2. Fix of android compilation issue with thumb mode
3. Fix of Component event issue

minggo:
[ci rebuild]

minggo:
@pandamicro  thanks.
@cezheng  will you share the `.clang-format file`? And i think it is better to put it in root of cocos2d-x

cezheng:
@minggo 
Since we have different code style than cocos2d-x, I don't think cocos2d-x can use it. But it looks like this:

```
---
BasedOnStyle:  Google
ColumnLimit: 110
BinPackParameters: false
SortIncludes: false
```

we have a coding style based on the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html), which has many differences such as using 2-space indent, not starting new lines on open braces, etc.

There's a website that can help you build a .clang-format file with visual effects:
http://clangformat.com/

minggo:
@cezheng thanks

PR #15720, opened 2016-05-25
Grammar corrections

billyquith:
I corrected some English grammar.
Also some function open brackets weren't consistent.


minggo:
Thanks very much.

PR #15724, opened 2016-05-26
Try catch IllegalStateException in Cocos2dxMusic.java.

dumganhar:


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxMusic.java
```diff
@@ -29,9 +29,6 @@ of this software and associated documentation files (the "Software"), to deal
 import android.content.res.AssetFileDescriptor;
 import android.media.MediaPlayer;
 import android.util.Log;
-import android.content.res.AssetFileDescriptor;
-import com.android.vending.expansion.zipfile.APKExpansionSupport;
-import com.android.vending.expansion.zipfile.ZipResourceFile;
```
minggo, line 34: Why remove these codes, it is needed for OBB. Please refer to #15515 for detail information.

dumganhar, line 34: This file doesn't use any interfaces in these packages.
And `import android.content.res.AssetFileDescriptor;` is duplicated.

minggo, line 34: Yep, you are right.

PR #15728, opened 2016-05-26
Remove duplicate header includes

mogemimi:
When reading through the implementation, I found some duplicate code such as the following:

``` cpp
#include "renderer/CCTextureAtlas.h"
#include "base/CCDirector.h"
#include "base/CCDirector.h" // Here is the duplicate header include.
#include "renderer/CCTextureCache.h"
```

This patch removes them to improve compile speeds.


minggo:
Thanks.

PR #15734, opened 2016-05-26
Support cc.view.setOrientation on web

pandamicro:
It can force your game be shown in a wished orientation on web


pandamicro:
Related https://github.com/cocos2d/cocos2d-html5/pull/3293

PR #15735, opened 2016-05-26
AudioEngine-linux fix: add map contains IDs of preloaded files

MrCapone:
Fix for `Fail to play <something> cause by limited max instance of AudioEngine` when playing same audio file multiple times. Platform Linux.

Forum topic of issue http://discuss.cocos2d-x.org/t/fail-to-play-cause-by-limited-max-instance-of-audioengine/26800


minggo:
@MrCapone Thanks.

PR #15737, opened 2016-05-27
It mute warning on android clang.

yalab:
Hi.
I'm using clang to build fo Android. It work fine. But it says  some warning. I mute it.
Thanks.


minggo:
Thanks.

PR #15740, opened 2016-05-27
Use CC_USE_3D_PHYSICS=1 for Bullet physics only

qbart:
I think definition for bullet should look like this, otherwise Chipmunk will be turn on :)


minggo:
@qbart i think you are right.

PR #15743, opened 2016-05-28
fix AudioEngine bug for iOS, UTF8 filePath

halx99:
Fix bug for apple platform, open audio file failed with chinese App Name.


minggo:
Thanks.

PR #15744, opened 2016-05-28
Fix warnings caused by comparison between RefPtr and C++11's nullptr

mogemimi:
When using `RefPtr<T>` with GCC 5.2.1 on Ubuntu 15.10, I get the following warning messages:

```
cocos/base/CCRefPtr.h: In member function ‘bool cocos2d::RefPtr<T>::operator>(std::nullptr_t) const’:
cocos/base/CCRefPtr.h:215:79: warning: ordered comparison of pointer with integer zero [-Wextra]
     inline bool operator > (const std::nullptr_t other) const { return _ptr > other; }
                                                                               ^
cocos/base/CCRefPtr.h: In member function ‘bool cocos2d::RefPtr<T>::operator<(std::nullptr_t) const’:
cocos/base/CCRefPtr.h:224:79: warning: ordered comparison of pointer with integer zero [-Wextra]
     inline bool operator < (const std::nullptr_t other) const { return _ptr < other; }
                                                                               ^
cocos/base/CCRefPtr.h: In member function ‘bool cocos2d::RefPtr<T>::operator>=(std::nullptr_t) const’:
cocos/base/CCRefPtr.h:233:81: warning: ordered comparison of pointer with integer zero [-Wextra]
     inline bool operator >= (const std::nullptr_t other) const { return _ptr >= other; }
                                                                                 ^
cocos/base/CCRefPtr.h: In member function ‘bool cocos2d::RefPtr<T>::operator<=(std::nullptr_t) const’:
cocos/base/CCRefPtr.h:242:81: warning: ordered comparison of pointer with integer zero [-Wextra]
     inline bool operator <= (const std::nullptr_t other) const { return _ptr <= other; }
                                                                                 ^
```

This problem seems to be caused by ordered comparison between pointers and null pointer constant. So we can use `std::less()` instead of relational operators such as `<`, `>`, `<=` and `>=`. This pull request fixes them.

### DIFF
--- a/cocos/base/CCRefPtr.h
+++ b/cocos/base/CCRefPtr.h
@@ -29,6 +29,7 @@
 
 #include "base/CCRef.h"
 #include "base/ccMacros.h"
+#include <functional>
 #include <type_traits>
 
 NS_CC_BEGIN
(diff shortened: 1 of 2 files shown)


minggo:
Thanks

PR #15755, opened 2016-05-30
Remove unnecessary sinf, cosf, tanf and sqrtf macros

mogemimi:
This pull request removes the user-defined C99 mathematical macros below from `CCStdC.h`:

```
#define tanf tan
#define sqrtf sqrt
#define cosf cos
#define sinf sin
```

The problem is that when I use `::sinf(float)` with my project on Linux, our `sinf` macro replaces it with `::sin(double)`. These macros are unnecessary, because they are [defined in C99](http://en.cppreference.com/w/c/numeric/math/sin), and the latest cocos2d-x [requires a C99 compiler](https://github.com/cocos2d/cocos2d-x/search?utf8=%E2%9C%93&q=c99&type=Code).

### DIFF
--- a/cocos/platform/linux/CCStdC-linux.h
+++ b/cocos/platform/linux/CCStdC-linux.h
@@ -50,12 +50,6 @@ THE SOFTWARE.
 #define MAX(x,y) (((x) < (y)) ? (y) : (x))
 #endif  // MAX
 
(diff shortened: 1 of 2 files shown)


minggo:
Thanks.

PR #15760, opened 2016-06-01
Triger callback

kermitbu:
Notify the trigger scroll page callback

### DIFF
--- a/cocos/ui/UIPageView.cpp
+++ b/cocos/ui/UIPageView.cpp
@@ -145,12 +145,24 @@ void PageView::setCurPageIndex( ssize_t index )
 
 void PageView::setCurrentPageIndex(ssize_t index)
 {
+	auto preindex = MAX(0, getCurrentPageIndex());
+		
     jumpToItem(index, Vec2::ANCHOR_MIDDLE, Vec2::ANCHOR_MIDDLE);
+
+	if (preindex != index)
+	{
+		pageTurningEvent();
+	}
 }
 
 void PageView::scrollToPage(ssize_t idx)
 {
-    scrollToItem(idx);
+	auto preindex = MAX(0, getCurrentPageIndex());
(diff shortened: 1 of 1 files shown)


cocos/ui/UIPageView.cpp
```diff
@@ -145,12 +145,24 @@ void PageView::setCurPageIndex( ssize_t index )
 
 void PageView::setCurrentPageIndex(ssize_t index)
 {
+	auto preindex = MAX(0, getCurrentPageIndex());
+		
     jumpToItem(index, Vec2::ANCHOR_MIDDLE, Vec2::ANCHOR_MIDDLE);
+
+	if (preindex != index)
+	{
+		pageTurningEvent();
```
dumganhar, line 154: Do we need to fire a turning event in `setCurrentPageIndex`?

cocos/ui/UIPageView.cpp
```diff
@@ -145,12 +145,24 @@ void PageView::setCurPageIndex( ssize_t index )
...
+	}
 }
 
 void PageView::scrollToPage(ssize_t idx)
 {
-    scrollToItem(idx);
+	auto preindex = MAX(0, getCurrentPageIndex());
```
zilongshanren, line 160: After calling scrollToItem, it ListView will scroll to the specific item, when the scroll ended , it will fire the `Page turning` event, so there is no need to fire the event manually.

cocos/ui/UIPageView.cpp
zilongshanren, line 154: I'm also curious about this change, do we really need fire  a turning event in `setCurrentPageIndex` ?

minggo:
@kermitbu any thought with the comments?

minggo:
@kermitbu i will close the PR if you don't have any reply.

PR #15764, opened 2016-06-02
ListView select item programmatically

AlexanderBykin:
later at 2015 this was rejected in case of too much merges, now it is clear


minggo:
@AlexanderShniperson Thanks, it looks good to me.
@zilongshanren please take a look.

PR #15766, opened 2016-06-02
Fix unused parameter warning

mogemimi:
This pull request fixes the following `-Wunused-parameter` warning when compiling with GCC 5.2.1 on Ubuntu 15.10.

```
cocos/2d/CCActionCatmullRom.cpp:57:43: warning: unused parameter ‘capacity’ [-Wunused-parameter]
 bool PointArray::initWithCapacity(ssize_t capacity)
                                           ^
```


minggo:
Thanks.

PR #15784, opened 2016-06-04
Fix unused parameter warning when compiling CCArray

mogemimi:
This patch fixes the following warning when compiling `__Array` class with GCC 5.2.1 on Ubuntu 15.10:

```
cocos/deprecated/CCArray.cpp:695:69: warning: unused parameter ‘releaseObject’ [-Wunused-parameter]
 void __Array::replaceObjectAtIndex(ssize_t index, Ref* object, bool releaseObject/* = true*/)
                                                                     ^
```

Thanks!


minggo:
Thanks.

PR #15792, opened 2016-06-06
Refine cpp template for iOS

AlwynC:
Refined cpp template for iOS by moving CCEAGLView set up code from AppController.mm to RootViewController.mm.Also added documentation for navigating between native iOS screen and cocos2d-x game scene.


minggo:
Thanks.

PR #15795, opened 2016-06-06
Avoid shadow warning when building with Xcode 7.3

mogemimi:
Hello, this PR avoids the following `-Wshadow` warning when compiling AssetsManagerEx with Xcode 7.3:

```
extensions/assets-manager/AssetsManagerEx.cpp:667:20: declaration shadows a local variable [-Wshadow]

        AsyncData* asyncData = (AsyncData*)param;
                      ^
```

So far this is the last patch for fixing minor compiler warnings in `libcocos2d Mac` (except those caused by third party libraries or bindings). :tada:

Thanks for the great work on cocos2d-x!

### DIFF
--- a/extensions/assets-manager/AssetsManagerEx.cpp
+++ b/extensions/assets-manager/AssetsManagerEx.cpp
@@ -664,8 +664,8 @@ void AssetsManagerEx::updateSucceed()
     _compressedFiles.clear();
 
     std::function<void(void*)> mainThread = [this](void* param) {
-        AsyncData* asyncData = (AsyncData*)param;
-        if (asyncData->errorCompressedFile.empty())
+        auto asyncDataInner = reinterpret_cast<AsyncData*>(param);
+        if (asyncDataInner->errorCompressedFile.empty())
         {
             // 5. Set update state
             _updateState = State::UP_TO_DATE;
@@ -675,10 +675,10 @@ void AssetsManagerEx::updateSucceed()
         else
         {
             _updateState = State::FAIL_TO_UPDATE;
-            dispatchUpdateEvent(EventAssetsManagerEx::EventCode::ERROR_DECOMPRESS, "", "Unable to decompress file " + asyncData->errorCompressedFile);
+            dispatchUpdateEvent(EventAssetsManagerEx::EventCode::ERROR_DECOMPRESS, "", "Unable to decompress file " + asyncDataInner->errorCompressedFile);
         }
 
-        delete asyncData;
+        delete asyncDataInner;
     };
     AsyncTaskPool::getInstance()->enqueue(AsyncTaskPool::TaskType::TASK_OTHER, mainThread, (void*)asyncData, [this, asyncData]() {
         // Decompress all compressed files


stevetranby:
I think you're technically changing what the code does here. It may be checked to be valid and good, but your code is taking a copy of asyncData at time the lambda is created whereas the previous code accepts the asyncData at time the lambda executes.

Again, it's possible this is still correct, but I believe it's different semantically.

mogemimi:
@stevetranby  Thank you for review. :smiley:

I propose three solutions for this issue:

#### Solution 1

``` cpp
AsyncData* asyncData = new AsyncData;
asyncData->compressedFiles = _compressedFiles;
_compressedFiles.clear();

std::function<void(void*)> mainThread = [this](void* param) {
    AsyncData* asyncDataInner = (AsyncData*)param;
    // ...
    delete asyncDataInner;
};
AsyncTaskPool::getInstance()->enqueue(AsyncTaskPool::TaskType::TASK_OTHER, mainThread, (void*)asyncData, [this, asyncData]() {
    // ...
});
```
- **Pro:** This change is simply renaming `asyncData` -> `asyncDataInner`:

#### Solution 2

``` cpp
std::function<void(void*)> mainThread = [this](void* param) {
    AsyncData* asyncData = (AsyncData*)param;
    // ...
    delete asyncData;
};

AsyncData* asyncData = new AsyncData;
asyncData->compressedFiles = _compressedFiles;
_compressedFiles.clear();

AsyncTaskPool::getInstance()->enqueue(AsyncTaskPool::TaskType::TASK_OTHER, mainThread, (void*)asyncData, [this, asyncData]() {
    // ...
});
```
- **Pro:** We do not have to rename `asyncData`.

#### Solution 3 [I've already submitted]

``` cpp
AsyncData* asyncData = new AsyncData;
asyncData->compressedFiles = _compressedFiles;
_compressedFiles.clear();

std::function<void(void*)> mainThread = [this, asyncData](void*) {
    // ...
    delete asyncData;
};
AsyncTaskPool::getInstance()->enqueue(AsyncTaskPool::TaskType::TASK_OTHER, mainThread, nullptr, [this, asyncData]() {
    // ...
});
```
- **Pro:** This removes unneeded C-style casting or `reinterpret_cast<AsyncData*>(param)`.
- **Cons:** This change will cause the overhead of lambda capture.

I think the solution 1 is easier to change than 2 and 3.

Any thought on that @stevetranby?

stevetranby:
Shadowing is sort of a tab vs. spaces argument. Some think it should be allowed (and thus don't allow use of -Wshadow flag) and is easier to read if the variable is essentially the same data. I currently enable this warning, but am more frequently wondering whether I should stop and allow shadowing instead of trying to work around the warning.

#1 works, obviously since only a rename. 
#3 looks to be valid with the added context. 
Both seem fine. #2 just seems like confusing things unnecessarily, even though it's cleaner syntax-wise.

I guess my preference would probably be to change as little as possible and thus going with #1. Also, since the parameter is part of the API it feels cleaner than #3 since it doesn't have an unused parameter.

mogemimi:
@stevetranby Thank you for answering this question so quickly!

I will submit the solution 1.

mogemimi:
I've just submitted a new commit. (Btw, the previous one: https://github.com/mogemimi/cocos2d-x/commit/beb26481d7e4caf613b5314aa2f1351b715bc5e9)
Could you please review it again?

minggo:
I like solution 1.

minggo:
Thanks guys.

mogemimi:
@minggo Thank you for considering this request. ☺️

PR #15810, opened 2016-06-08
Update CCTableView.cpp

GodHong:
cleanup shoule be false, otherwise we will remove the action in cell.
BTW, there are two ';' at the end.


zilongshanren:
@GodHong  
Thanks for the patch.
I have one question, In what situation, we need to keep the action when move cell out of sight?

GodHong:
@zilongshanren  issue 15756 it's weird

PR #15811, opened 2016-06-09
Fix typo in method name: 'destory' -> 'destroy'

mogemimi:
There is a typo in method name, `destory` should be `destroy` in LocalStorage for Android. And so this patch fixes it.

Thanks.


minggo:
Thanks.

PR #15816, opened 2016-06-10
Fixed to fail FileUtils::writeValueMapToFile on iOS.

yalab:
Hi.
I found a bug that FileUtils::writeValueMapToFile fail on iOS, because NSDict writeToFile cannot accept null.
So this patch remove null data key from the ValueMap.

### DIFF
--- a/tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
+++ b/tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
@@ -795,6 +795,8 @@ void TestWriteValueMap::onEnter()
     ValueMap mapInValueMap;
     mapInValueMap["string1"] = "string in dictInMap key 0";
     mapInValueMap["string2"] = "string in dictInMap key 1";
+    mapInValueMap["none"].getType();
+    
     valueMap["data0"] = Value(mapInValueMap);
 
     valueMap["data1"] = Value("string in array");
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -1268,4 +1268,12 @@ std::string FileUtils::getFileExtension(const std::string& filePath) const
     return fileExtension;
 }
 
+void FileUtils::valueMapCompact(ValueMap& valueMap)
+{
+}
+
+void FileUtils::valueVectorCompact(ValueVector& valueVector)
+{
+}
+
 NS_CC_END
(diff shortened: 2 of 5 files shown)


minggo:
@yalab How to reproduce the issue?

yalab:
@minggo Commet out "cocos/platform/apple/CCFileUtils-apple.mm Line:370". After that build cpp-tests on ios. And open cpp-tests app "20: FileUtils" -> "12: FileUtils: TestWriteValueMap to files". You can see "write failed".

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
```diff
@@ -795,6 +795,8 @@ void TestWriteValueMap::onEnter()
     ValueMap mapInValueMap;
     mapInValueMap["string1"] = "string in dictInMap key 0";
     mapInValueMap["string2"] = "string in dictInMap key 1";
+    mapInValueMap["none"].getType();
```
minggo, line 798: What does this code do? Create a `none` element? If so then it is strange to do it like this.

minggo:
@yalab If i do as you said, the will cause assert error in [this line](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/CCValue.cpp#L743)

tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
yalab, line 798: > What does this code do? Create a none element? If so then it is strange to do it like this.

Yes it create none element. I want to say that anyone create none element if they intended or not.

yalab:
@minggo I reproduced the error.  It becames fail to save the ValueMap.

minggo:
@yalab so it is something error about this PR?

yalab:
@minggo Yes it is.

Shulepov:
@minggo yep, plist can contain only next object types:
`NSData, NSString, NSNumber, NSDate, NSArray, or NSDictionary`

So, it's actually not acceptable to store `null` or `NSNull` objects. And that PR removes their occurrences.

minggo:
@Shulepov I am waiting @yalab to fix the bug.

yalab:
@minggo It have alredy fixed. If you comment out "cocos/platform/apple/CCFileUtils-apple.mm Line:370" then it assert error because the ValueMap contain none. But you uncomment "cocos/platform/apple/CCFileUtils-apple.mm Line:370" then it is fine.

minggo:
@yalab I misunderstood your meaning.

PR #15819, opened 2016-06-10
CCFontFreeType : Changed the type of parameter 'outline' from int to float.

ryule:
I found that 'outline' parameter is `int` type in CCFontFreeType.
The private value for 'outline' is `float`, so I think that there is no reason to limit the type as `int`.
Is there any reason to keep the parameter type as `int`?

p.s.) I didn't modify `int` 'outline' for codes related to ui::Widgets as I worried a side-effect.

### DIFF
--- a/cocos/2d/CCFontFreeType.cpp
+++ b/cocos/2d/CCFontFreeType.cpp
@@ -49,7 +49,7 @@ typedef struct _DataRef
 
 static std::unordered_map<std::string, DataRef> s_cacheFontData;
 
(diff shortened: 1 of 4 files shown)


minggo:
[ci rebuild]

zilongshanren:
@rouddyNeo 
good catch, but this PR is incomplete, it doesn't handle the system font outline.
I will handle it later, thanks.

dumganhar:
@zilongshanren, any update of changing outline type from int to float for System font?

Issue #8435, opened 2014-09-30
JSON for the cocos2d::Value


minggo:
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/5038). It was created at 2014/05/07 05:14:45 +0000
# Description

It will be great if Value can be parse to and from a JSON format std::string. It will be so much better then using a third party parser.

PR #15821, opened 2016-06-11
CCValue <-> JSON converter

Shulepov:
Noticed that somebody requested this feature - https://github.com/cocos2d/cocos2d-x/issues/8435
And btw i had already implemented it. So want to share it if it is still relevant.


minggo:
@Shulepov Thanks for the great PR, but i think it is better not add it into cocos2d-x. It would be great to be an extension, but we don't have a good way to share the extension.

Issue #15826, opened 2016-06-12
Label System font support Overflow: clamp and shrink


zilongshanren:
supported platforms:
- [x] iOS
- [x] Mac
- [x] Android
- [ ] Win32
- [ ] Windows Universal
- [ ] Linux
- [ ] Tizen
- [ ] Window RT

PR #15827, opened 2016-06-12
Add system label overflow

zilongshanren:
for issue: https://github.com/cocos2d/cocos2d-x/issues/15826

Note: currently only iOS, Android and Mac are supported.

This PR mainly add missing `clamp` and `shrink` feature to system label.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -560,6 +560,11 @@ void Label::updateShaderProgram()
 
 void Label::setFontAtlas(FontAtlas* atlas,bool distanceFieldEnabled /* = false */, bool useA8Shader /* = false */)
 {
+    if(atlas)
+    {
+        _systemFontDirty = false;
+    }
+
     if (atlas == _fontAtlas)
     {
         FontAtlasCache::releaseFontAtlas(atlas);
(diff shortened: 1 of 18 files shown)


minggo:
[ci rebuild]

zilongshanren:
iOS platform can't pass CI, I need to add CoreText to other tests project: js-tests/lua-tests etc

zilongshanren:
@dumganhar  please review, thanks.

cocos/2d/CCLabel.cpp
```diff
@@ -560,6 +560,11 @@ void Label::updateShaderProgram()
 
 void Label::setFontAtlas(FontAtlas* atlas,bool distanceFieldEnabled /* = false */, bool useA8Shader /* = false */)
 {
+    if(atlas)
+    {
+        _systemFontDirty = false;
```
dumganhar, line 565: Why to set `_systemFontDirty` flag to false here?

cocos/base/ccTypes.h
```diff
@@ -566,7 +568,10 @@ struct CC_DLL FontDefinition
     FontShadow            _shadow;
     /// font stroke
     FontStroke            _stroke;
-
+    /// enable text wrap
+    bool                  _enableWrap;
+    /// enable shrink font size
+    int                  _overflow;
```
dumganhar, line 577: Why it's an integer value? If it has to be integer, it's better to add comments to describe the meaning of integer value.

zilongshanren:
@dumganhar  The comments has been improved, please check.

cocos/base/ccTypes.h
```diff
@@ -570,7 +570,10 @@ struct CC_DLL FontDefinition
     FontStroke            _stroke;
     /// enable text wrap
     bool                  _enableWrap;
-    /// enable shrink font size
+    /** There are 4 overflows: none, clamp, shrink and resize_height.
+     *  The conresponding integer values are 0, 1, 2, 3 respectively
+     * For more information, please refer to Label::Overflow enum class.
+     */
     int                  _overflow;
```
minggo, line 577: why not use enum instead of int?

zilongshanren, line 577: The enum is defined inside Label class, and it need to pass around to JNI. Int is simpler.

PR #15831, opened 2016-06-13
Add a config option to enable or disable the premultiplied alpha code.

j-jorge:
Computing premultiplied alpha when loading an image is expensive and can be ignored in some situations. This commit adds a config option to enable or disable the premultiplied alpha code (default is to enable) so the developer can choose the option most suitable for him.

### DIFF
--- a/cocos/base/ccConfig.h
+++ b/cocos/base/ccConfig.h
@@ -397,4 +397,12 @@ THE SOFTWARE.
 #define CC_FILEUTILS_APPLE_ENABLE_OBJC  1
 #endif
 
+/** @def CC_ENABLE_PREMULTIPLIED_ALPHA
+ * If enabled, all textures will be preprocessed to multiply its rgb components
+ * by its alpha component.
+ */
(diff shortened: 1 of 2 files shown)


minggo:
@j-jorge Thanks for the PR. But according Spine's license, we can not modify the spine runtime codes.

stevetranby:
Should contact Spine to have them approve.

j-jorge:
@minggo Would it be an acceptable pull request without the changes in Spine? In this case Spine animations would render incorrectly if the option is disabled :/

minggo:
> Should contact Spine to have them approve.

@stevetranby We have already done like this, but failed. @ricardoquesada right?

@j-jorge Yep.

ricardoquesada:
this is a two part patch:
- part of the patch should be applied in cocos2d-x (all files except skeleton...)
- 2nd part: Skeleton part... send the patch directly to spine

j-jorge:
The changes in Spine's code have been reverted. Please review the changes in this branch.

minggo:
Thanks.

Issue #15828, opened 2016-06-12
RenderTexture: missed setOpacityModifyRGB


Shulepov:
Just noticed an issue, that changing opacity of sprite from RenderTexture doesn't work properly.

It sets:

```
_sprite->setBlendFunc( BlendFunc::ALPHA_PREMULTIPLIED );
```

But it doesn't set:

```
_sprite->setOpacityModifyRGB(true);
```

minggo:
How to reproduce the issue?

Shulepov:
If i want to render something once and use result later:

```
RenderTextur *rt = RenderTexture::create(...);
rt->begin();
//draw something here
rt->end();

// add render texture directly
rt->setCascadeOpacityEnabled(true);
rt->setOpacity(0);
addChild(rt);

// clone sprite and copy important properties
Sprite *result = Sprite::createWithTexture(rt->getSprite()->getTexture());
result->setBlendFunc(rt->getSprite()->getBlendFunc());
result->setOpacityModifyRGB(rt->getSprite()->isOpacityModifyRGB());
result->setOpacity(0);
addChild(result);
```

In both cases sprite shouldn't be visible, but it is.

PR #15834, opened 2016-06-13
RenderTexture - set opacity modify RGB

Shulepov:
Fixes https://github.com/cocos2d/cocos2d-x/issues/15828


Shulepov:
Possibly, it may be better somehow to set _texture.alphaPremultiplied = true.

PR #15835, opened 2016-06-13
Move StringUtils functions from deprecated header file to ccUTF8.h

mogemimi:
Hello,

When I use `StringUtils::format()` and `StringUtils::toString()` functions under v3 branch, I need to add `#include "deprecated/CCString.h"` into my code, then the `deprecated` directory makes me confused, because the two functions are not deprecated.

Also according to the [cocos2d.h](https://github.com/cocos2d/cocos2d-x/blob/103bdb3d378554260e8324ac633b5c1fa1d30357/cocos/cocos2d.h#L314-L315):

```
It is recommanded that you just inlcude what is needed.
eg. #include "deprecated/CCString.h" if you only need cocos2d::__String.
```

So I suggest that we should move these function declarations outside of the deprecated header, in the same way as v4-develop changes (commit: https://github.com/cocos2d/cocos2d-x/commit/497db86977921deeb159527dd91809fe75a36329).

This PR does the following:
- Move `StringUtils::format()` and `StringUtils::toString()` from `deprecated/CCString.h` to `base/ccUTF8.h`
- Add `#include "base/ccUTF8.h"` into `CCString.h` to keep the backward compatibility
- Replace inclusion of deprecated header with `base/ccUTF8.h`

Thank you for considering!

### DIFF
--- a/cocos/2d/CCAction.cpp
+++ b/cocos/2d/CCAction.cpp
@@ -29,7 +29,7 @@ THE SOFTWARE.
 #include "2d/CCActionInterval.h"
 #include "2d/CCNode.h"
 #include "base/CCDirector.h"
-#include "deprecated/CCString.h"
+#include "base/ccUTF8.h"
 
 NS_CC_BEGIN
 //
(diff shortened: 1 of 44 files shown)


minggo:
Thanks @mogemimi

PR #15837, opened 2016-06-14
Various CMake improvements

adesentenac:
- Added c++_shared (LLVM) to Android toolchain options. Fixed AndroidNdkGdb.cmake.
- Fixed includes when not using prebuilt libs. Added option to use sources in external when not using prebuilt libs.
- Added option in CMake to be able build Box2D without using it as physics library in cocos2d-x (similar to what is done in the Visual Studio solution).


tools/simulator/libsimulator/lib/runtime/RuntimeCCSImpl.cpp
```diff
@@ -3,7 +3,7 @@
 #include "ConfigParser.h"
 #include "cocostudio/CocoStudio.h"
 #include "ui/UIHelper.h"
-#include "tinyxml2/tinyxml2.h"
```
minggo, line 6: Why need to change it like this?

minggo:
> Added option to use sources in external when not using prebuilt libs.

Did you mean don't want to use prebuilt libs, but want to use source codes in external?

minggo:
> Fixed AndroidNdkGdb.cmake.

What's the bug?

tools/simulator/libsimulator/lib/runtime/RuntimeCCSImpl.cpp
adesentenac, line 6: If you don't use prebuilts, since the 'external' folder is present in the include path, it founds the tinyxml2 in the external folder before finding the system one.

adesentenac:
> > Added option to use sources in external when not using prebuilt libs.
> 
> Did you mean don't want to use prebuilt libs, but want to use source codes in external?

yes, we target platforms not supported by the prebuilts (linux arm), but that doesn't need to prevent from building the externals provided as source.
I'm guessing it's the same for people using MinGW, there is no reason to depend on the version provided in the packages when the correct source is present in the externals.

adesentenac:
> > Fixed AndroidNdkGdb.cmake.
> 
> What's the bug?

It was not working here to debug with gdb, notably the get_property(TARGET_LOCATION...) was returning an error. 
Is it working for you as-is? if so do you have an example? maybe it's because we use it differently, but I did not find any call to it in the source code to compare with how we use it.

tools/simulator/libsimulator/lib/runtime/RuntimeCCSImpl.cpp
minggo, line 6: Got it, but it is easy to change back to use `tinyxml2/tinyxml2.h` because we don't test the case you meet and current implementation can reduce the probability of naming conflict.

minggo:
> It was not working here to debug with gdb, notably the get_property(TARGET_LOCATION...) was returning an error. 
> Is it working for you as-is? if so do you have an example? maybe it's because we use it differently, but I did not find any call to it in the source code to compare with how we use it.

No, i haven't use gdb to debug. But it works great with Android Studio, which build/debug it correctly. I am  not sure if the modification will affect it. I have to try it.

minggo:
Ok, it will not affect Android Studio debugging.

minggo:
Thanks.

davidchai-6waves:
i've got a compilation error when i ran 'cocos gen-libs' command in terminal.
`cocos2d-x/tools/simulator/libsimulator/lib/runtime/RuntimeCCSImpl.cpp:6:10: fatal error: 'tinyxml2.h' file not found
#include "tinyxml2.h"
         ^
1 error generated.`

P.S. i am using the latest code from github repo, and was running the command on my Mac.

adesentenac:
Hi @davidchai-6waves, I'll try to see if I can reproduce the problem on Windows or Linux (I don't currently have a Mac available) and see if I can fix it!

PR #15838, opened 2016-06-14
lua binding : Rename end() to endToLua().

MagicXrEv:
There're still some end() functions need to rename to endToLua(), such as AudioEngine::end() and GLView::end(). And the way RenderTexture used to rename end() is out-of-date, so I changed it too.Please review this patch. Thanks.

### DIFF
--- a/cocos/2d/CCRenderTexture.h
+++ b/cocos/2d/CCRenderTexture.h
@@ -113,12 +113,9 @@ public:
      */
     virtual void beginWithClear(float r, float g, float b, float a, float depthValue, int stencilValue);
 
-    /** End is key word of lua, use other name to export to lua.
-     * @js NA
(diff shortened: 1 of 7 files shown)


minggo:
Why need to rename to `endToLua()`?

MagicXrEv:
Because "end" is a keyword in lua. And as you can see, some classes of cocos2d-x had done this work before, such as Director::end() and RenderTexture::end(). 

It should be noted that SimpleAudioEngine::end() had been renamed to destoryInstance() in lua binding before this patch. For upward compatible, I think it shouldn't be renamed, so I didn't change it.

minggo:
@MagicXrEv Yep, keep compatibility is important.

minggo:
Thanks.

PR #15839, opened 2016-06-14
Cleanup and refactor RootViewController

Shulepov:
- removed some redundant code & comments;
- make EAGLView a root view of RootViewController instead of subview;
- added missed `bridge` casts;

Also, few proposals:
- do some actions in `viewDidDisappear` & `viewWillAppear` (when new ViewController is pushed). E.g. pause/resume the game, so open GL view is not used while it's not visible;
- rename `RootViewController` to `CocosViewController`;
- extract multi touch setting into cocos side;

### DIFF
--- a/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm
+++ b/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm
@@ -38,8 +38,6 @@
 static AppDelegate s_sharedApplication;
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {    
-
-    
     // Override point for customization after application launch.
 
     // Add the view controller's view to the window and display.
--- a/templates/cpp-template-default/proj.ios_mac/ios/RootViewController.mm
+++ b/templates/cpp-template-default/proj.ios_mac/ios/RootViewController.mm
@@ -27,15 +27,7 @@
 #import "cocos2d.h"
 #import "platform/ios/CCEAGLView-ios.h"
 
-@interface RootViewController()
 
-
-@property(strong,nonatomic) CCEAGLView *eaglView;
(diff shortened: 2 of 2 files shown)


minggo:
> do some actions in viewDidDisappear & viewWillAppear (when new ViewController is pushed). E.g. pause/resume the game, so open GL view is not used while it's not visible;

May be some app want to continue to work even it is not visible?

> rename RootViewController to CocosViewController;

It looks good to me, but it may bring compatibility issue and don't have any benefit, so i think it is better to keep as it is now.

> extract multi touch setting into cocos side;

I think only iOS  supports enable/disable multi touch. So it is not a good idea to add it into cocos side.

stevetranby:
The events could propagate in the same way the "enter foreground" event works. User could then choose to listen or set callback and pause/resume themselves.

minggo:
But not all platforms have similar events.

stevetranby:
alright, no worries, we have focus/blur for desktop window events added into our fork of the engine, figured any that don't implement would just never get the callbacks? It's fine to not add and have users do their own platform-specific event handling.

AlwynC:
@Shulepov What is the benefit of making eaglview as root view of RootViewController rather than subview I'd like to know?

Shulepov:
@AlwynC Actually, there is no a big difference... But initially it was set as root view. And there will be less subviews.
The only benefits i see of adding it as a subview - ability to set custom size and position, more convenient integration with UIKit / storyboards - but that are quite rare cases. In these cases developer possibly knows what he doing and can move code from `loadView` to `viewDidLoad`.

AlwynC:
@Shulepov So finally eaglview is removed as a property and made as 'root view' of RootViewController instead of subview and put into loadView method with all 'redundant code' and 'comments' removed.There's no big difference after all isn't it.Very very good.

PR #15844, opened 2016-06-15
fix: missing properties for clone

halx99:
ui::Widget missing CascadeColorEnabled and CascadeOpacityEnabled for clone.


zilongshanren:
@minggo  This PR looks good to me.

minggo:
Thanks guys.

PR #15847, opened 2016-06-15
[ci skip]hide virtual button by default

minggo:


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
```diff
@@ -301,13 +304,15 @@ protected void onCreate(final Bundle savedInstanceState) {
...
         super.onResume();
+        this.hideVirtualButton();
        	resumeIfHasFocus();
     }
     
     @Override
     public void onWindowFocusChanged(boolean hasFocus) {
     	Log.d(TAG, "onWindowFocusChanged() hasFocus=" + hasFocus);
         super.onWindowFocusChanged(hasFocus);
+        this.hideVirtualButton();
```
dumganhar, line 315: Should we check `hasFocus` value? hideVirtualButton need to be invoked only while hasFocus is `true`.

minggo, line 315: Yep, i think it is better add it to `resumeIfHasFocus()`

minggo, line 315: fixed in #15848

PR #15852, opened 2016-06-15
V3: updated to support Windows 10 UWP x64 builds

stammen:
This pull request includes the following:
- Adds support to build x64 version of Windows 10 UWP (x64 support depends on PR https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/230) This will be needed to build Windows 10 UWP versions for Xbox.
- Updates libcocos2d name to 3.12 (needed for Windows Store version tracking)
- Adds missing OpenGL defines in winrt/CCGL.h


minggo:
Thanks.

PR #15854, opened 2016-06-15
Fix CCController-android symbols not being exported to the Java side.

adesentenac:
Now that cocos2dxandroid_static is used to export symbols to the Java side, CCController-android.cpp must be included in it for GameControllerAdapter.java to work correctly, otherwise android does not find the symbols for the JNI calls.


minggo:
I have some questions:
1. how can i reproduce the issue?
2. It is not good just move `CCController-android.cpp` to platform because it is part of base.

adesentenac:
> I have some questions:
> 
> how can i reproduce the issue?

If you use ControllerManualAdapter (i.e. you game activity inherits from GameControllerActivity), the game will crash as soon as cocos2d-x detects the controller with the following error:

```
06-16 10:56:47.625: W/dalvikvm(9606): No implementation found for native Lorg/cocos2dx/lib/GameControllerAdapter;.nativeControllerConnected:(Ljava/lang/String;I)V
06-16 10:56:47.703: W/dalvikvm(9606): threadid=13: thread exiting with uncaught exception (group=0x410cf2a0)
06-16 10:56:47.703: E/AndroidRuntime(9606): FATAL EXCEPTION: GLThread 775
06-16 10:56:47.703: E/AndroidRuntime(9606): java.lang.UnsatisfiedLinkError: Native method not found: org.cocos2dx.lib.GameControllerAdapter.nativeControllerConnected:(Ljava/lang/String;I)V
06-16 10:56:47.703: E/AndroidRuntime(9606):     at org.cocos2dx.lib.GameControllerAdapter.nativeControllerConnected(Native Method)
06-16 10:56:47.703: E/AndroidRuntime(9606):     at org.cocos2dx.lib.GameControllerAdapter.access$000(GameControllerAdapter.java:5)
06-16 10:56:47.703: E/AndroidRuntime(9606):     at org.cocos2dx.lib.GameControllerAdapter$1.run(GameControllerAdapter.java:35)
```

> It is not good just move CCController-android.cpp to platform because it is part of base.

I thought it made sense to move platform specific code to the platform folder, it is not very clear to me the difference with base other than that, but maybe there is another way to do it?
Since I only moved a .cpp file, I thought this wouldn't have too much impact on anything else.
 The goal is to have it included in cocos2dxandroid_static library, so that java can find the symbol.

minggo:
> I thought it made sense to move platform specific code to the platform folder, it is not very clear to me the difference with base other than that, but maybe there is another way to do it?
> Since I only moved a .cpp file, I thought this wouldn't have too much impact on anything else.
> The goal is to have it included in cocos2dxandroid_static library, so that java can find the symbol.

What i mean is that, if you move the implementation of android to platform specific, then you should move all other platforms implementation too.

adesentenac:
Oh, makes sense :D
I can do that tomorrow morning when I get back to work.
I don't have the setup to test it for iOS though (but the risk should be minimal).

minggo:
Thanks.

adesentenac:
Should I move CCController.h and CCController.cpp too, or just the platform dependant -iOS.mm?

minggo:
@adesentenac Yep, move `CCController.h and CCController.cpp` to platform, and dependant -iOS.mm to `platform/ios`. 

But i think i have to reproduce it first. Can i just use `game-controller-test` to reproduce it?

adesentenac:
I didn't reproduce the issue with game-controller-test using the Android .mk files, but the generated libraries are not the same as with CMake.

I'll try to create a CMake project for game-controller-test to duplicate how we do it in our game, but I didn't have time to do it today.

minggo:
@adesentenac as controller is not common used by developers, so i want to close this PR and there are some conflicts of this PR. Please create a new one if you thin it is needed. Thanks.

PR #15855, opened 2016-06-15
Fix possible memory leak when pass initialization fails

mogemimi:
This patch fixes two possible memory leaks if `pass` initialisation fails. Thanks.


mogemimi:
[ci rebuild]

minggo:
Thanks.

PR #15859, opened 2016-06-16
updated external/config.json to version v3-deps-98

stammen:
Updated externall/config.json to version v3-deps-98


minggo:
@stammen The linux error is a known issue. We will fix it and generate v3-deps-99. So i will close this PR.

PR #15869, opened 2016-06-16
Fixed android controller events propagation

adesentenac:
To test it, simply use the propagation mechanism on a controller event (call stopPropagation on it), and all future events of the same type, will be dispatched with their propagation already stopped.

Another solution would have been to create new Event objects instead of re-using Event objects created once, but maybe there is a specific reason for how it's done here, so I chose to simply add a resetPropagation method which is called before dispatching the event.

### DIFF
--- a/cocos/base/CCController.cpp
+++ b/cocos/base/CCController.cpp
@@ -77,12 +77,14 @@ const Controller::KeyStatus& Controller::getKeyStatus(int keyCode)
 
 void Controller::onConnected()
 {
+    _connectEvent->resetPropagation();
     _connectEvent->setConnectStatus(true);
     _eventDispatcher->dispatchEvent(_connectEvent);
 }
 
 void Controller::onDisconnected()
 {
+    _connectEvent->resetPropagation();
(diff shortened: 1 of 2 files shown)


minggo:
@adesentenac from the description i don't know what's the issue, and how can i test it

adesentenac:
Hi @minggo sorry for the late answer.
Here is a quick way to test it. In GameControllerTest.cpp replace onKeyDown and onKeyUp with:

```
void GameControllerTest::onKeyDown(cocos2d::Controller *controller, int keyCode, cocos2d::Event *event)
{
    if (!event->isStopped())
        showButtonState(controller, keyCode, true);
}

void GameControllerTest::onKeyUp(cocos2d::Controller *controller, int keyCode, cocos2d::Event *event)
{
    if (!event->isStopped())
        showButtonState(controller, keyCode, false);
    event->stopPropagation();
}
```

since the same instance is reused for every event, only the first press of any button will be shown without my fix.
With my fix, all button presses will correctly be shown.

PR #15870, opened 2016-06-16
Remove unused debug function in RenderCommand

mogemimi:
This pull request removes `printBits` function that is not used anywhere in the RenderCommand (since commit: https://github.com/cocos2d/cocos2d-x/commit/8931d968c0ea639e055440acd194d9397afc8e60) to speed up compile time. Thanks.


minggo:
Thanks.

PR #15878, opened 2016-06-17
fix downloader build issue for Android API 23

### LINKED ISSUE #15823: Compiling error with Android 6.0(API 23)
`Cocos2dxDownloader.java` uses [Apache HTTP Client](https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-apache-http-client) which is removed in Android 6.0, should use `HttpURLConnection` instead.

There is corresponding [ticket](http://discuss.cocos2d-x.org/t/cant-compile-to-android/26874) in the forum.

cezheng:
fixes #15823 

android-async-http 1.4.9 builds for API 23, but it introduces an external dependency, with a jar file of 1.1 MB.

Considering the fact that replacing libcurl with a java network library is to reduce the apk size, if this PR is merged, will the apk size still be smaller than using libcurl?

If not, we should consider another approach.

@minggo


cezheng:
List of java network libraries to pick from

https://github.com/snowdream/awesome-android#network

Personally I think okhttp has the best quality, but it is 300+ KB, 3 times as android-async-http 1.4.8

minggo:
Thanks @cezheng .

Issue #15776, opened 2016-06-03
cocos2d::Texture2D addImage crashes if fed a bad cocos2d::Image


jonnyijapan:
cocos2d::Texture2D addImage crashes if fed a bad cocos2d::Image

Confirmed to crash with Cocos2d-x 3.10...
Sorry I didn't test with Cocos2d-x 3.11.

How to reproduce:

```
static cocos2d::TextureCache mytexturecache;

+(cocos2d::Texture2D*)getAdTexture2dFromUIImage:(UIImage*)uiimage {
    NSData *objCImageData = NULL; // This doesn't make sense but is just to reproduce crash later. Normally I would get NSData from UIImage...
    cocos2d::Image *cocos2dxImage = new cocos2d::Image();

    const bool LOADEDCORRECTLY = cocos2dxImage->initWithImageData(objCByteData, objCImageData.length);

    // Purposelly we do not check the return value of LOADEDCORRECTLY - it will fail of course in this case, but if we don't check it the addImage below will crash.

    free(objCByteData);

    auto texture2d = mytexturecache.addImage(cocos2dxImage, "unused"); // Crashes here

    return texture2d;
}
```

ricardoquesada:
Added "fix".
The thing is that cocos2d-x has an strange policy of crashing when the data is invalid. We have to change that policy in the whole cocos2d-x.

What I did, was to add an ASSERT. But a proper solution is to return null instead... but in order to do that we have to change the "crash on fail" policy.

PR #15880, opened 2016-06-17
fix: TextureCache ASSERTS if image is invalid

ricardoquesada:
Github issue #15776

### DIFF
--- a/tests/cpp-tests/Classes/BugsTest/Bug-15776.cpp
+++ b/tests/cpp-tests/Classes/BugsTest/Bug-15776.cpp
@@ -1,0 +1,26 @@
+//
+//  Bug-15776.cpp
+//  cocos2d_tests
+//
+//  Created by Ricardo Quesada on 6/17/16.
+//
+//
+
+#include "Bug-15776.h"
+
+
+USING_NS_CC;
+
+bool Bug15776Layer::init()
+{
+    if (BugsTestBase::init())
+    {
+        cocos2d::Image *cocos2dxImage = new cocos2d::Image();
+        cocos2dxImage->initWithImageData(nullptr, 0);
+        // should not crash. invalid cocos2dImage
+        auto texture2d = Director::getInstance()->getTextureCache()->addImage(cocos2dxImage, "unused");
+        return texture2d;
+    }
+
+    return false;
+}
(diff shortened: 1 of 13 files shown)


tests/cpp-tests/Classes/BugsTest/Bug-15776.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+bool Bug15776Layer::init()
+{
+    if (BugsTestBase::init())
+    {
+        cocos2d::Image *cocos2dxImage = new cocos2d::Image();
+        cocos2dxImage->initWithImageData(nullptr, 0);
+        // should not crash. invalid cocos2dImage
+        auto texture2d = Director::getInstance()->getTextureCache()->addImage(cocos2dxImage, "unused");
+        return texture2d;
```
minggo, line 22: @ricardoquesada it will cause `assert` error as you added `CCASSERT(image->getData() != nullptr, "TextureCache: image MUST not be nil");` in `TextureCache::addImage`. So i am confused with the test case.

ricardoquesada, line 22: @minggo correct. I haven't "fixed" the bug. What I added was an assert so that it crashes when it should crash and not somewhere else.

But I think in v4 we should change the policy of cocos2d-x, and it should never crash, even when it receives invalid data. When it receives invalid data it should use a default one.

minggo, line 22: Got it, thanks.

minggo, line 22: Can add a title to describe it, or testing engineer will treat it as an error?

ricardoquesada, line 22: sure. makes sense. I'll add the comments on Thursday (tomorrow wednesday I'll be away)

minggo, line 22: thanks

ricardoquesada:
@minggo done. sent one PR to v3 and another one to v3.13

minggo:
thanks

PR #15884, opened 2016-06-18
Fix for the continuous repeat suddenly stopping.

kshoji:
With Repeat class, the continuous repeat animation stops suddenly during about 30 minutes.
It may be a rare timing-bug.
I've tested this fix and verified with Android and iOS app.

Regards, Kaoru


minggo:
@kshoji Could you describe why need to modify like this?

minggo:
@kshoji how to reproduce it?

minggo:
@kshoji i will close this PR if you can not tell me how to reproduce it.

PR #15888, opened 2016-06-20
remove port info to avoid enametoolong error

springrider:
while testing on iOS simulator and devices, the connect function in libwebsockets, returns errno 36, which is enametoolong. 
I have no idea what this means, but remove the port did fix the problem.

more details could be found here:

http://forum.cocos.com/t/libwebsockets-ipv6-no-route-to-host/37089/2

### DIFF
--- a/cocos/network/WebSocket.cpp
+++ b/cocos/network/WebSocket.cpp
@@ -623,7 +623,7 @@ void WebSocket::onSubThreadStarted()
 
         char portStr[10];
         sprintf(portStr, "%d", _port);
-        std::string ads_port = _host + ":" + portStr;
+        std::string ads_port = _host; // + ":" + portStr;
 
         _wsInstance = lws_client_connect(_wsContext, _host.c_str(), _port, _SSLConnection,
                                              _path.c_str(), ads_port.c_str(), ads_port.c_str(),


minggo:
But the fix remove the port info, it is needed.

springrider:
yep, I am using default port 80, so it's a quick and dirty fix for me.  
sorry I do not really understand the reason why this is happening, so can't fix the port problem...

by the way, the port information is already applied by "_port".  so I am not why do you still need "ads_port".  
lws_client_connect received it as "const char *origin", seems an irrelevant parameter.

minggo:
> by the way, the port information is already applied by "_port". so I am not why do you still need "ads_port".

@dumganhar i found you added the codes, could you please answer it?

> lws_client_connect received it as "const char *origin", seems an irrelevant parameter.

Which parameter?

dumganhar:
It was copied/pasted from `libwebsockets`'s sample code. I'll check the sample code again.

dumganhar:
You could refer to https://github.com/warmcat/libwebsockets/blob/1c428c3154c4b9834ff952301a4ed795cb525140/test-server/test-echo.c#L472

minggo:
@springrider any problem with @dumganhar 's comments? I will close this PR if there is not good solution.

springrider:
@dumganhar  you just copied the sample code from libwebsockets and no idea why they use that parameter?

@minggo  my guess is that the port info in ads_port is irreverent, only as http header, and can be removed.   and it's okay if you close it.  seems no one other than me has this problem....

dumganhar:
@springrider, it's better to look inside the source code of `libwebsockets` to find how it handles this parameter. But now, i don't have enough time to investigate.

minggo:
@springrider Thanks, i will close this PR, and please send a new one if you find any solution.

Issue #14730, opened 2015-12-17
SpriteBatchNode crash if CC_SPRITE_DEBUG_DRAW enabled


newnon:
enable CC_SPRITE_DEBUG_DRAW will crash test with SpriteBatchNode nodes

ricardoquesada:
at least in v3.11 it doesn't crash anymore but it doesn't render the sprites... just the frames.

**UPDATE**: it didn't crash for me... but only by pure chance.

ricardoquesada:
fixed.

PR #15892, opened 2016-06-21
fix: not crash if CC_SPRITE_DEBUG_DRAW is enabled in SpriteBatchNode

ricardoquesada:
`SpriteBatchNode` was trying to render `DrawNode` when
`CC_SPRITE_DEBUG_DRAW` was enabled.

Github issue #14730


CartmanORCamille:
When I am using cocos2dx3.16, when CC_SPRITE_DEBUG_DRAW is set not to 0, the addChild of BatchNode will have an assert, the details of the assert are "vector iterator not incrementable", this assert locates the for loop in BatchNode.appendChild It seems that the problem of #14730 still exists, what is the better solution?

Looking forward to your reply.

PR #15895, opened 2016-06-21
FileUtils: Add std::move for retun value.

halx99:
Explicit avoid unnecessary memory alloc & copy.

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -618,14 +618,14 @@ std::string FileUtils::getStringFromFile(const std::string& filename)
     // truncated
     s.resize(strlen(s.data()));
 
-    return s;
+    return std::move(s);
 }
 
 Data FileUtils::getDataFromFile(const std::string& filename)
 {
     Data d;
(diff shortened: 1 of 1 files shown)


mogemimi:
Sorry I might be wrong, but this commit will cause the following warnings when compiling code with `-Wpessimizing-move` option under Clang (I have reproduced with Xcode 7.3.1):

```
cocos/platform/CCFileUtils.cpp:621:12: Moving a local object in a return statement prevents copy elision
cocos/platform/CCFileUtils.cpp:628:12: Moving a local object in a return statement prevents copy elision
```

For this reason, please see below at the section 12.8.31 of [N3690](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf), page 274. Would we need to change this?

halx99:
OK, you are right, The compiler will call the move-constructor automatically. Thanks, let's close this PR.

mogemimi:
@halx99  Thank you for considering it. I basically agree with you. :+1: We should avoid copying in favour of using `std::move` whenever possible.

PR #15897, opened 2016-06-21
luabindings: Add script with utf-8 bom support.

halx99:


minggo:
@halx99 Thanks for the PR, but is it needed, because developers will encrypt script files.

halx99:
Yes, In our project, I also encrypt all script files.

minggo:
If so, then why you add the logic in `else`?

halx99:
I missing that, should handle encrypted script at same time.
Well, in our project, bacause I hook FileUtils to decrypt all assets and scripts, so I ignore the else logic.

minggo:
Thanks.

PR #15900, opened 2016-06-21
Fixed simulator project files by adding tinyxml2 to the include paths

adesentenac:
Fixed regression in simulator build caused by https://github.com/cocos2d/cocos2d-x/pull/15837

PR #15906, opened 2016-06-22
Fixed simulator project files to add tinyxml2 to the include paths

adesentenac:
Fixed regression in simulator build caused by https://github.com/cocos2d/cocos2d-x/pull/15837 for v3.12.
As asked by @minggo in https://github.com/cocos2d/cocos2d-x/pull/15900


minggo:
Thanks.

PR #15919, opened 2016-06-22
fix system font creation on Android platform

zilongshanren:
Since this PR https://github.com/cocos2d/cocos2d-x/pull/9915/files changed the default behavior of `fullPathForFilename`, when pass system font name isn't a valid file path, it will return empty string.


zilongshanren:
@minggo  This PR also fix js-tests UI missing UITest.json file problem.
I update the ccs-res submodule and add the missing UITest.json file.

zilongshanren:
[ci rebuild]

Issue #15909, opened 2016-06-22
cpp-tests Scene3D -> description buttons don't work


ricardoquesada:
- cd cpp-tests
- click Node::Scene3D
- click Description
- click `hair` or `glass`

The hair or glass are not changed.

In v3.11, it works as expected. I don’t remember if there is any codes except your VR PR changed the logic.

minggo:
`27:Node:Camera 3D Test->5:Camera ArcBall Moving` can not move the scene and click the `Switch Operation` and `Switch Target` doesn't have effect.

ricardoquesada:
Patch for v3.12: https://github.com/cocos2d/cocos2d-x/pull/15931

Patch for v3: https://github.com/cocos2d/cocos2d-x/pull/15930

ricardoquesada:
merged for v3.
v3.12 patch still open. @minggo I've not merged the v3.12 patch since you own that branch. let me know if you have any question about it. thanks.

minggo:
@ricardoquesada Thanks, i will have a try.

minggo:
Thanks, it is fixed.

ricardoquesada:
With this PR I found a few bugs in VR now :(... I'll send a new PR tomorrow.

minggo:
Ok.

PR #15930, opened 2016-06-22
fix: reverse iterator on events + doesn't use additionalTransform

ricardoquesada:
use reverse Camera iterator when traversing events (of course).
but somehow `setAdditionalTransform()` breaks
`setNodeToParentTransform()`. And since `setAdditionalTransform` is not
needed, we just don't use it.

Github issue #15909


ricardoquesada:
[ci rebuild]

ricardoquesada:
merged for v3.
v3.12 patch still open. @minggo I've not merged the v3.12 patch since you own that branch. let me know if you have any question about it. thanks.

PR #15936, opened 2016-06-23
Use nullptr instead of 0 or NULL whenever possible

mogemimi:
This pull request changes some null pointer constants to nullptr for the following reasons:
- When compiling with the latest GCC, they will cause `-Wzero-as-null-pointer-constant` warnings.
- If possible, we should follow [our adopted coding style](https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#0-and-nullptrnull) for better maintenance and readability.

Thanks!


minggo:
Thanks.

Issue #15937, opened 2016-06-23
Node:Particles Test will cause huge lag


zilongshanren:
@ricardoquesada   This PR https://github.com/cocos2d/cocos2d-x/pull/15646 will alloc indices memory on the fly and it will cause particles play laggy... please refer to cpp-tests -> Node: Particles for example.

Here is the logging message.

```
cocos2d: QuadCommand: resizing index size from [90] to [534]
cocos2d: QuadCommand: resizing index size from [534] to [564]
cocos2d: QuadCommand: resizing index size from [564] to [708]
cocos2d: QuadCommand: resizing index size from [708] to [852]
cocos2d: QuadCommand: resizing index size from [852] to [1002]
cocos2d: QuadCommand: resizing index size from [1002] to [1110]
cocos2d: QuadCommand: resizing index size from [1110] to [1122]
cocos2d: QuadCommand: resizing index size from [1122] to [1248]
cocos2d: QuadCommand: resizing index size from [1248] to [1272]
cocos2d: QuadCommand: resizing index size from [1272] to [1362]
cocos2d: QuadCommand: resizing index size from [1362] to [1410]
cocos2d: QuadCommand: resizing index size from [1410] to [1446]
cocos2d: QuadCommand: resizing index size from [1446] to [1464]
cocos2d: QuadCommand: resizing index size from [1464] to [1518]
cocos2d: QuadCommand: resizing index size from [1518] to [1560]
cocos2d: QuadCommand: resizing index size from [1560] to [1668]
cocos2d: QuadCommand: resizing index size from [1668] to [1776]
cocos2d: QuadCommand: resizing index size from [1776] to [1782]
cocos2d: QuadCommand: resizing index size from [1782] to [1884]
cocos2d: QuadCommand: resizing index size from [1884] to [1908]
cocos2d: QuadCommand: resizing index size from [1908] to [1956]
cocos2d: QuadCommand: resizing index size from [1956] to [1992]
cocos2d: QuadCommand: resizing index size from [1992] to [2046]
cocos2d: QuadCommand: resizing index size from [2046] to [2058]
cocos2d: QuadCommand: resizing index size from [2058] to [2124]
cocos2d: QuadCommand: resizing index size from [2124] to [2142]
cocos2d: QuadCommand: resizing index size from [2142] to [2166]
cocos2d: QuadCommand: resizing index size from [2166] to [2268]
cocos2d: QuadCommand: resizing index size from [2268] to [2292]
[...]

PR #15946, opened 2016-06-23
fix: fix delay when creating quadcommands

ricardoquesada:
triggered by particles. uses bigger buffer by default. better logic to increate size.

Github issue #15937

### DIFF
--- a/cocos/renderer/CCQuadCommand.cpp
+++ b/cocos/renderer/CCQuadCommand.cpp
@@ -72,9 +72,20 @@ void QuadCommand::init(float globalOrder, GLuint textureID, GLProgramState* glPr
 
 void QuadCommand::reIndex(int indicesCount)
 {
+    // first time init: create a decent buffer size for indices to prevent too much resizing
+    if (__indexCapacity == -1)
+    {
+        indicesCount = std::max(indicesCount, 2048);
+    }
+
     if (indicesCount > __indexCapacity)
     {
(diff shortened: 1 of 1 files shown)


ricardoquesada:
@minggo I think this patch is import for v3.12. Please, review it. Thanks.

I already applied in v3.

cocos/renderer/CCQuadCommand.cpp
```diff
@@ -72,9 +72,20 @@ void QuadCommand::init(float globalOrder, GLuint textureID, GLProgramState* glPr
...
+    // first time init: create a decent buffer size for indices to prevent too much resizing
+    if (__indexCapacity == -1)
+    {
+        indicesCount = std::max(indicesCount, 2048);
+    }
+
     if (indicesCount > __indexCapacity)
     {
+        // if resizing is needed, get needed size plus 25%, but not bigger that max size
+        indicesCount *= 1.25;
```
crazyhappygame, line 84: Usually this factor is in range [1.5;2]. Is there any particular reason why 1.25?
https://github.com/facebook/folly/blob/master/folly/docs/FBVector.md

ricardoquesada, line 84: I didn't want to be too aggressive here since this memory probably won't be freed.

minggo:
@ricardoquesada thanks, i think it needs to be v3.12 too.

PR #15948, opened 2016-06-23
fix: collection of vr fixes

ricardoquesada:
- Node::additionalTransform() and Node::setNodeToParentTransform() can be used at the same time
- Node allocs _additionalTransform in runtime taking 60 bytes lees for 99% of the nodes
- Node calculates the transform faster when an additionalTransform is being used
- VR works when the camera is being moved manually with setNodeToParentTransform
- VR has its own tests
- VR can be disabled and works as expected


web
```diff
@@ -1 +1 @@
-Subproject commit d3f467431111ef39753dfa138d4e6f4b0c70f5c7
+Subproject commit 2b3a2df2cf17552daf2e78fcb69b944412090d26
```
dumganhar: Does it need to change the reference of the `web` sub module?

ricardoquesada: nope. thanks. I'll fix it.

ricardoquesada:
@minggo this fixes many bugs in VR... it would be great if this can be part of v3.12.

minggo:
Yep, i just wait the jenkins to pass it.

ricardoquesada:
[ci rebuild]

argh... problems with the slaves..

minggo:
[ci rebuild]

minggo:
And there is conflict.

ricardoquesada:
fixed. thanks

minggo:
No, the conflict still exist.

ricardoquesada:
@minggo yep, you were right. a recent commit on v3.12 created the conflicts. fixed.

minggo:
Thanks, i will merge it when it passes jenkins.

PR #15950, opened 2016-06-24
fix part of the android restart game crash in Layer::Color to glDrawArrays

Sumxx:
restart game -> call invalidateStateCache to s_attributeFlags  = 0;
-> reset  glEnableVertexAttribArray type, Cause glDrawArrays crash.

part of the android (test mi3, huawei)

### DIFF
--- a/cocos/renderer/ccGLStateCache.cpp
+++ b/cocos/renderer/ccGLStateCache.cpp
@@ -64,7 +64,6 @@ void invalidateStateCache( void )
 {
     Director::getInstance()->resetMatrixStack();
     s_currentProjectionMatrix = -1;
-    s_attributeFlags = 0;
 
 #if CC_ENABLE_GL_STATE_CACHE
     s_currentShaderProgram = -1;


minggo:
Does it crash when i restart game?
How to restart game?

Sumxx:
ues AssetsManager update Game,  received jsb.EventAssetsManager.UPDATE_FINISHED callback notification After call cc.game.restart

minggo:
But why you think it is the codes cause the problem?

minggo:
How did you find the solution?

Sumxx:
i annotation this code, app is not crash. Will only call when initializing and restart, but restart and the initialization process is not the same. openGL not reset in restart.

minggo:
But this function can be invoked more than one time, if it has problem, i can reproduce it without restart game, right?

Sumxx:
yes, this function will only be called when restart. However, he will collapse on the part of the android machine. After the update. You can try mi3 or huawei

Sumxx:
this crash, It is inevitable

minggo:
It is a special case, i think we can not just change the codes here. `invalidateStateCache()` should invalid all as the function mean.

Sumxx:
OK, add new function of restart? I hope to hear your opinion on this BUG and amendments.

minggo:
I don't know the reason, so i can not give you advise right now. You are appreciated if you can provide a simple test case to reproduce it.

Sumxx:
OK, The problems happened in a commercial project,  I try to reproduce this problem, you can close this PR first.

minggo:
Thanks.

PR #15952, opened 2016-06-24
Avoid implicit conversion from size_t to int

mogemimi:
This patch fixes a implicit conversion warning below when compiling libjscocos2d with Xcode 7.3.1:

```
cocos/scripting/js-bindings/manual/jsb_opengl_manual.cpp:392:32: Implicit conversion loses integer precision: 'unsigned long' to 'int'
```


minggo:
Thanks

Issue #15934, opened 2016-06-23
FPS could not be shown on iOS


dumganhar:
After this Pull Request (https://github.com/cocos2d/cocos2d-x/pull/15853) was merged, FPS would not be shown on iOS while it works ok on Android.
I reverted the changes in this PR and FPS was shown correctly on iOS too.

@ricardoquesada , please help to check what's going wrong. Thanks.

ricardoquesada:
sent PR to v3.12. awaiting @minggo approval.
I'll merge the PR sent to v3.

PR #15956, opened 2016-06-24
fix: FPS appears on iOS

ricardoquesada:
`glViewport()` has a default value so that it has a valid value
after `Scene::render()` exits

github issue #15934

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -256,6 +256,14 @@ void Director::setGLDefaultValues()
     setAlphaBlending(true);
     setDepthTest(false);
     setProjection(_projection);
+
+    // Everything should be drawn within `Scene::render()`.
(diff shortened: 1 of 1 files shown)


ricardoquesada:
@minggo please, review it. thanks. this is for v3.12

minggo:
@ricardoquesada yep, the fps is shown again, but i don't know why it doesn't have problem before? Is there any modification that cause the problem?

ricardoquesada:
@minggo yes, the bug was triggered with the VR code. 

Basically VR "fixed" a bug with the viewport... basically the viewports were set by the camera, but when the camera finished drawing, it never restored the viewport.
So I fixed that bug for VR, but that bug triggered this bug.
In "desktop" it was not triggered because they were setting a default viewport.

minggo:
Got it, thanks.

PR #15961, opened 2016-06-25
Fix typo in FileUtils::Status enum

mogemimi:
This patch corrects a minor typo I found while reading v3.12 features: `ReadFaild` -> `ReadFailed`


minggo:
Thanks

PR #15962, opened 2016-06-25
Gesture Recognizer classes inside extensions

alfogrillo:
Classes for gesture recognition (iOS and Android tested).
For further information visit this page:
https://github.com/Drakon-Cocos/cocos2dx-gesture-recognizers


minggo:
Thanks for the good PR, but i think it is better not include it into cocos2d-x. It is better treated as a package when we implement package manager.

PR #15963, opened 2016-06-25
Fix PageViewIndicator opacity for 'inactive' pages

alfogrillo:


zilongshanren:
@Drakon-Cocos 
I think it's better to provide a option for developers to tweaking the opacity of inactive indicators.
No hardcoded opacity value is much better.

alfogrillo:
@zilongshanren 
I agree, it is better something like that:

`PageView::setIndicatorIndexOpacity(GLubyte opacity)`

minggo:
It works as expected, thanks.

PR #15964, opened 2016-06-26
Can set scroll to item duration in ListView and derived classes

alfogrillo:
This patch add the possibility to set the scroll duration between items.
I found this particularly useful in PageView (derived from ListView).

### DIFF
--- a/cocos/ui/UIListView.h
+++ b/cocos/ui/UIListView.h
@@ -245,14 +245,17 @@ public:
     float getItemsMargin()const;
     
     /**
-     * Set time to scroll between items (see scrollToItem)
-     * @param time
+     * Set the time in seconds to scroll between items.
+     * Subsequent calls of function 'scrollToItem', will take 'time' seconds for scrolling.
+     * @param time The seconds needed to scroll between two items. 'time' must be >= 0
+     * @see scrollToItem(ssize_t, const Vec2&, const Vec2&)
      */
-    void  setScrollDuration(const float time);
+    void  setScrollDuration(float time);
     
      /**
-     * Get time to scroll between items
-     * @param time
+     * Get the time in seconds to scroll between items.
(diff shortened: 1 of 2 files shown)


cocos/ui/UIListView.h
```diff
@@ -244,6 +244,18 @@ class CC_GUI_DLL ListView : public ScrollView
      */
     float getItemsMargin()const;
     
+    /**
+     * Set time to scroll between items (see scrollToItem)
+     * @param time
+     */
+    void  setScrollDuration(const float time);
```
minggo: I think const is not needed here since it is passed by value.

cocos/ui/UIListView.h
```diff
@@ -244,6 +244,18 @@ class CC_GUI_DLL ListView : public ScrollView
      */
     float getItemsMargin()const;
     
+    /**
+     * Set time to scroll between items (see scrollToItem)
+     * @param time
```
minggo: I think it is better to mention that, the value should big or equal 0. And please complete the comments. Thanks.

cocos/ui/UIListView.h
alfogrillo: I mark `const` read only parameters for two reasons:
- Sometime compilers can optimize `const` parameters
- Avoid programmer mistakes (as unwanted parameter modification).
  Anyway I see that cocos2d-x programmers typically does not mark `const` to read only parameter. Then I will delete the `const` for consistency

cocos/ui/UIListView.h
alfogrillo: Ok!

cocos/ui/UIListView.h
minggo: const is useful if passing reference or pointer, but it is needed by value.

minggo:
Thanks.

PR #15966, opened 2016-06-26
Improve RefPtr to avoid casting when possible

mogemimi:
This pull request improves the `RefPtr` implementation. It does the following:
- Replace the type of `RefPtr::_ptr` with `T*` instead of `Ref*` to avoid casting when possible
- Remove unneeded `reinterpret_cast` and `const_cast`
- Add `static_assert` to ensure `T` is derived from `cocos2d::Ref` at compile-time
- Add test case for `RefPtr<T const>`
- Minor code formatting

This change has no overhead at runtime. It also provides more compile-time safety.
Thanks!

### DIFF
--- a/cocos/base/CCRefPtr.h
+++ b/cocos/base/CCRefPtr.h
@@ -87,8 +87,7 @@ template <typename T> class RefPtr
 public:
     
     inline RefPtr()
-    :
-        _ptr(nullptr)
+        : _ptr(nullptr)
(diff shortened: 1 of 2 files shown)


minggo:
Thanks, it looks good.

mogemimi:
@minggo Thank you for reviewing this request so quickly! :smiley:

PR #15971, opened 2016-06-27
Avoid const_cast when calling Node::setAdditionalTransform

mogemimi:
This PR adds a const qualifier to `Node::setAdditionalTransform(Mat4*)` parameter and removes unnecessary `const_cast` for the reasons below:
- The definition of `Node::setAdditionalTransform(Mat4*)` doesn't seem to require non-const parameter which should be const. It can also use `const Mat4*` instead.
- We should avoid using `const_cast` whenever possible for better maintenance and performance.

The function is not virtual, so this change doesn't break backward compatibility.
Thanks.


minggo:
Thanks.

PR #16002, opened 2016-06-29
PageView::setCurrentPageIndex triggers PageView::EventType::TURNING

alfogrillo:
When `PageView::setCurrentPageIndex` is called the event `PageView::EventType::TURNING` is triggered

### DIFF
--- a/cocos/ui/UIPageView.cpp
+++ b/cocos/ui/UIPageView.cpp
@@ -146,6 +146,8 @@ void PageView::setCurPageIndex( ssize_t index )
 void PageView::setCurrentPageIndex(ssize_t index)
 {
     jumpToItem(index, Vec2::ANCHOR_MIDDLE, Vec2::ANCHOR_MIDDLE);
+    if (getItem(index)) // PageView::EventType::TURNING triggered only if index is valid
+        pageTurningEvent();
 }
 
 void PageView::scrollToPage(ssize_t idx)


minggo:
@zilongshanren Please take a look.

minggo:
It will break compatibility. Developers may listen the event and do some operation.

minggo:
I think it is better to trigger a new event.

minggo:
@Drakon-Cocos any idea?

alfogrillo:
@minggo 
Think to a scenario like this:
When user scroll to a page some action must be taken to update page content.
Now it makes sense to do that in `PageView::EventType::TURNING`.
Sometimes however developer can scroll pages with `PageView::setCurrentPageIndex` maybe because the `PageView` can be scrolled also with `ui::Button` added by the developer.
In the latter case the event `PageView::EventType::TURNING` was not triggered.
Now with this patch it is triggered even in this case.
If you think this can broke compatibility (make me an example to understand well the case you are thinking about) we can do something like that:

The normal scroll triggers:

```
PageView::EventType::TURNING
PageView::EventType::NEW_PAGE
```

The `PageView::setCurrentPageIndex` triggers:
`PageView::EventType::NEW_PAGE`

Where `PageView::EventType::NEW_PAGE`is a new event

minggo:
> If you think this can broke compatibility (make me an example to understand well the case you are thinking about)

There are two ways to turn a page:
- scrollToPage
- setCurrentPageIndex

developers may listen `PageView::EventType::TURNING` event and do some operation. They supposed only `scrollToPage` will trigger the event. But now `setCurrentPageIndex` can trigger the event and do the operation. 

I am not sure if `PageView::EventType::NEW_PAGE` is meaningful. Because `scrollToPage` will also switch to a new page too.

alfogrillo:
I thought about this, maybe it can be better to do that:
- `PageView::setCurrentPageIndex` doesn't trigger nothing (no modification)
- `void PageView::scrollToPage(ssize_t idx)` should became `PageView::scrollToPage(ssize_t idx, float duration)`
  In this way developer could scroll in zero seconds triggering the event!

@minggo 
What do you think?

minggo:
`PageView::scrollToPage(ssize_t idx, float duration)` will break compatibility. I think we just need a new event name, a more meaningful event name.

alfogrillo:
@minggo 
I think compatibility can be kept by making `duration` a default parameter or overloading `PageView::scrollToPage`.
Anyway there is a little adjustment that should be made here:

```
void ListView::scrollToItem(ssize_t itemIndex, const Vec2& positionRatioInView, const Vec2& itemAnchorPoint)
{
    scrollToItem(itemIndex, positionRatioInView, itemAnchorPoint, DEFAULT_TIME_IN_SEC_FOR_SCROLL_TO_ITEM);
}
```

Should use:
`_scrollTime` instead of `DEFAULT_TIME_IN_SEC_FOR_SCROLL_TO_ITEM`

Because this PR: #15964

minggo:
@Drakon-Cocos i think overloading is better.

alfogrillo:
@minggo 
Overloading here #16390

PR #16006, opened 2016-06-29
Fix getSubStringOfUTF8String bug

mogemimi:
Hello, this pull request addresses the `ui::Helper::getSubStringOfUTF8String` issue raised here:
http://discuss.cocos2d-x.org/t/problem-with-utf-strings/30461

I think we should not handle a raw utf-8 string without unicode-specific third-party library/implementation (such as LLVM's `ConvertUTF` and UTF8-CPP etc.) for maintenance or security reasons. So this changeset simply replaces with a code using `StringUtils::UTF8ToUTF32`, and adds a new test case.

Thank you for your time and consideration.

### DIFF
--- a/cocos/ui/UIHelper.cpp
+++ b/cocos/ui/UIHelper.cpp
@@ -26,6 +26,7 @@ THE SOFTWARE.
 #include "ui/UIWidget.h"
 #include "ui/UILayoutComponent.h"
 #include "base/CCDirector.h"
+#include "base/ccUTF8.h"
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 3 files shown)


minggo:
@mogemimi why need to change to utf32 than change to utf8 later? Why not just hanlde ut8 directly?

mogemimi:
You have a point there. I think using UTF32 is too over-engineered to substring. Actually, the worst-case time complexity of the original implementation is `O(n)`, and the best-case complexity is `O(start + length)`, but the worst-case time complexity of this PR is `O(2n)`.

At the same time, using `StringUtils::UTF8ToUTF32` will make it substantially more secure and more simple than the one which directly handling UTF-8. I would rather rely on `StringUtils::UTF8ToUTF32` or any third-party libraries than directly handling raw UTF-8 characters which incurs a high maintenance cost. In fact, the original one has a possibility of causing bugs such as this issue and the redundant UTF-8 sequence problem.

In addition, the `ui::Helper::getSubStringOfUTF8String` is mainly used for `TextField` and `RichText`. They need to be correctly processed UTF-8 strings because we cannot predict what string which contains invalid UTF-8 character(s) is entered by a user into a text field/rich text. Execution speed is important, but not as important as security and maintainability here.

Also, the reason why I have implemented with UTF-32 is because `ccUTF8.h` and `ConvertUTF.h` are already contained in cocos2d-x project. If we want to make it faster and more secure, we can rely on UTF-8-specific libraries such as [UTF8-CPP](https://github.com/nemtrif/utfcpp), [utf8.h](https://github.com/sheredom/utf8.h), [utf8proc](https://github.com/JuliaLang/utf8proc) or [MicroUTF-8](http://puszcza.gnu.org.ua/software/microutf8/) instead.

Thank you for reading the long description (and sorry I'm not fluent in writing English).

mogemimi:
@minggo I just came up with an alternative way of solving the problem using LLVM's `ConvertUTF.h` without converting to UTF-32. I guess it is faster than the old way (commit: https://github.com/mogemimi/cocos2d-x/commit/a7d87ea68c0d8291ea92f934449352d01ac4c6b2) .
Would you please review this change again? Thanks a lot.

minggo:
@mogemimi Thanks for your explanation. I was not online yesterday, so didn't reply in time. I think your original implementation is good after reading your explanation. The reason are:
- this function is not invoked so frequently, so the performance is not the first requirements
- the original implementation re-use existing codes and make codes more simple and maintainable

So could you please revert to original implementation, sorry about it.

mogemimi:
@minggo Thank you for reply. I have reverted back to the original commit and rebased against current master.

minggo:
@mogemimi thanks, will test it later

minggo:
It works well, thanks.

mogemimi:
@minggo Thank you for testing and confirming this! :relaxed:

Issue #15965, opened 2016-06-26
Discuss about remove cocos2d::Node _orderOfArrival?


halx99:
It's for simplify & robustness.

minggo:
Yep, i think we can remove it. It is not used any more.

minggo:
fixed in #16009

PR #16009, opened 2016-06-30
#15965: remove unused _orderOfArrival

halx99:


halx99:
[ci rebuild]

minggo:
thanks @halx99

cocos/2d/CCNode.h
```diff
@@ -668,27 +668,6 @@ class CC_DLL Node : public Ref
...
-     */
-    void setOrderOfArrival(int orderOfArrival);
-    /**
-     * Returns the arrival order, indicates which children is added previously.
-     *
-     * @see `setOrderOfArrival(unsigned int)`
-     *
-     * @return The arrival order.
-     */
-    int getOrderOfArrival() const;
```
minggo, line 689: @halx99 since we use `order of arrival` back, i think we can add back these functions. replace `updateOrderOfArrival()` with `setOrderOfArrival()` and add back getOrderOfArrival().

halx99, line 689: ok

halx99, line 689: but, 3.12 already remove the two function, does this really necessary?

minggo, line 689: Oh, i forgot which version it is removed. Ok, let's keep it as it is until developers need it. Thanks.

PR #16021, opened 2016-07-02
Fixed to use TexturePacker with cocosstudio.

yalab:
This is same as https://github.com/cocos2d/cocos2d-x/pull/16008 for UIButton and Slider.


minggo:
[ci rebuild]

minggo:
@yalab could you please add a test case?

yalab:
@minggo Which file should I write test?

minggo:
@yalab you can add a test case in `tests/cpp-tests/Classes/UITest/CocoStudioGUITest`

PR #16022, opened 2016-07-02
Fix format string warning when compiling with Android NDK

mogemimi:
This patch fixes the following warning when compiling with Clang for Android NDK r10e:

```
cocos/ui/UIHelper.cpp:125:58: warning: format specifies type 'unsigned long' but the argument has type 'std::string::size_type' (aka 'unsigned int') [-Wformat]
        cocos2d::log("'start' is out of range: %lu, %s", start, str.c_str());
                                               ~~~       ^~~~~
                                               %u
```

Thanks.

### DIFF
--- a/cocos/ui/UIHelper.cpp
+++ b/cocos/ui/UIHelper.cpp
@@ -122,7 +122,7 @@ std::string Helper::getSubStringOfUTF8String(const std::string& str, std::string
         return "";
     }
     if (utf32.size() < start) {
-        CCLOGERROR("'start' is out of range: %lu, %s", start, str.c_str());
+        CCLOGERROR("'start' is out of range: %zu, %s", static_cast<size_t>(start), str.c_str());
         return "";
     }
     std::string result;


mogemimi:
[ci rebuild]

stevetranby:
I think you should change this to better support across the board? Add a macro:

```
#ifdef __WIN32__ // or whatever
#define CC_SSIZET_FMT "%Id"
#define CC_SIZET_FMT "%Iu"
#else
#define CC_SSIZET_FMT "%zd"
#define CC_SIZET_FMT "%zu"
#endif
cocos2d::log("'start' is out of range: " CC_SIZET_FMT ", %s", start, str.c_str());
```

 or use the standard's macros: `PRId64` and `PRIu64` with

```
#define __STDC_FORMAT_MACROS 1 
#include <inttypes>`.
cocos2d::log("'start' is out of range: " PRIu64 ", %s", start, str.c_str());
```

minggo:
@stevetranby i can only find [cinttypes](http://www.cplusplus.com/reference/cinttypes/). And i can not find `PRId64` in the doc.

stevetranby:
Ah, silly haste copy paste. It's cinttypes or inttypes.h. The former is probably better (custom cc prefixed macros) since apparently compiler writers may or may not implement correctly.

You can test on VS2015 to confirm whether '%zu' works or not. I don't believe it does.

If it works, great, then never mind. Otherwise my replies are just a for your info (FYI).

We're using separate pre-compiled headers and have our own macros.

```
// windows (msvc)
#define PRINTF_ZD "%Id"
#define PRINTF_ZU "%Iu"
```

```
// ios/mac/android (probably linux since I think this supports gcc and clang)
#define PRINTF_ZD "%zd"
#define PRINTF_ZU "%zu"
```

MSVC (msdn) reference
https://msdn.microsoft.com/en-us/library/tcxf1dw6.aspx

Here's a stackoverflow reference (read with grain of salt as usual):
https://stackoverflow.com/questions/1546789/clean-code-to-printf-size-t-in-c-or-nearest-equivalent-of-c99s-z-in-c

stevetranby:
`%zu` is a C99 standard, but Microsoft only recently caught up to it. So again if there's only a single use case in the engine source you can probably ignore my comments.

https://stackoverflow.com/questions/2524611/how-can-one-print-a-size-t-variable-portably-using-the-printf-family

Issue #16003, opened 2016-06-29
Design Resolution broken after minimize window on win32


MrCapone:
After minimize window design resolution sets to window size. This can cause problems if design resolution does not equal to frame size. Sample code

`glview = GLViewImpl::createWithRect("MyGame", Rect(0, 0, 1280, 720));`
`director->setOpenGLView(glview);`
`glview->setDesignResolutionSize(640, 360, ResolutionPolicy::FIXED_WIDTH);`

Before minimize `Director::getInstance()->getVisibleSize()` will return 640x360, after 1280x720.

MrCapone:
https://github.com/cocos2d/cocos2d-x/pull/16031

PR #16031, opened 2016-07-04
GLView::setFrameSize fixed

MrCapone:
fix for https://github.com/cocos2d/cocos2d-x/issues/16003
also fixed changing size of resizable window

### DIFF
--- a/cocos/platform/CCGLView.cpp
+++ b/cocos/platform/CCGLView.cpp
@@ -194,7 +194,7 @@ const Size& GLView::getFrameSize() const
 
 void GLView::setFrameSize(float width, float height)
 {
-    _designResolutionSize = _screenSize = Size(width, height);
+    _screenSize = Size(width, height);
 }
 
 Rect GLView::getVisibleRect() const


minggo:
thanks

ricardoquesada:
This is probably is the correct fix for the design resolution, but it is breaking backward compatibility. See this issue: https://github.com/cocos2d/cocos2d-x/issues/16485

@minggo I think we should revert this PR, and find a better way to fix it.

ricardoquesada:
This PR should fix both #16003 and #16485 
https://github.com/cocos2d/cocos2d-x/pull/16533

minggo:
@ricardoquesada it may break backward compatibility, but why it will cause crash?

minggo:
Oh, it is an assert error.

ricardoquesada:
@minggo my PR won't break backward compatibility. The one that broke compatibility was this PR (#16031). My PR fixes the backward compatibility issued caused by this PR while fixing bug #16003

minggo:
@ricardoquesada yep, i got it. Thanks.

PR #16041, opened 2016-07-06
Fix various typos

mogemimi:
This patch fixes several small typos I found while reading code and API docs. Thanks!


minggo:
thanks

PR #16065, opened 2016-07-07
Fix typos in README and other markdown docs

mogemimi:
Hello, this patch fixes minor spelling mistakes I found while reading the documentation.


minggo:
thanks

Issue #16073, opened 2016-07-08
cocos2d-x v3.12 Text::create crash in iOS.


vkreal:
The code below crashes in iOS. This doesn't happen in v3.11. It's caused from the "\360" text string. Anyone know a workaround?

`Text* text_top = Text::create("mwhahaha\360", "Verdana", 40);`

Thanks!

ricardoquesada:
what exactly is `\360`? are you trying to print "\360"... or is that an escape char? thanks.

vkreal:
@ricardoquesada I am not sure what it is. The data is coming in from users entering text in TextField widget. The code above crashes consistently. I google around looking for ways to remove the slash but haven't found a good solution for c++ string. You have any ideas?

ricardoquesada:
@vkreal I'll create a test case and see what happens. thanks for reporting it.

mogemimi:
Hi. `\360` is arbitrary octal digit number according to section 2.14.3 of [n3690](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf).
- http://en.cppreference.com/w/cpp/language/escape
- https://msdn.microsoft.com/en-us/library/h21280bw.aspx

Also, `"mwhahaha\360"` contains invalid UTF-8 sequence such as `\xf0`, and so the code below always returns `nil`:

``` objc
NSString * string  = [NSString stringWithUTF8String:"mwhahaha\360"];
```

It works well when I add error handling instead:

``` objc
// _initWithString() in CCDevice-mac.mm

NSString * string  = [NSString stringWithUTF8String:text];

if (string == nil) {
    CCLOGERROR("'text' contains invalid UTF-8 sequence: %s", text);
    string = [NSString stringWithUTF8String:""];
}

...

NSAttributedString *stringWithAttributes =[[[NSAttributedString alloc] initWithString:string
                                                                           attributes:tokenAttributesDict] autorelease];
```

vkreal:
@mogemimi very nice! Will this nil check be in next release? Just noticed we should have same thing in CCDevice-ios.mm

mogemimi:
@vkreal I don't know the release plan for cocos2d-x 3.13, but hopefully it can be fixed in the next release.

ricardoquesada:
fixed. thanks.

PR #16075, opened 2016-07-09
fix: Text::create() does not crash on mac/ios

ricardoquesada:
...on invalid unicode.

fixes issue #16073


ricardoquesada:
[ci rebuild]

PR #16092, opened 2016-07-12
fix, replace std::sort by std::stable_sort

halx99:
Developer feedback: http://forum.cocos.com/t/cocos2dx-3-12/37958


halx99:
Need review

halx99:
@minggo: 
 I post some test result at http://forum.cocos.com/t/cocos2dx-3-12/37958 please take a look.

minggo:
@halx99 looks good me, thanks.

minggo:
since mac slave is offline, so merge it

PR #16093, opened 2016-07-12
make -n args works

zilongshanren:


build/android-build.py
```diff
@@ -38,21 +38,24 @@ def caculate_built_samples(args):
...
-def build_samples(target, build_mode):
+def build_samples(target, build_mode, app_abi):
 
     if build_mode is None:
         build_mode = 'debug'
     elif build_mode != 'release':
         build_mode = 'debug'
 
+    if app_abi is None:
+        app_abi = 'armeabi-v7a'
```
minggo, line 57: Why use `armeabi-v7a` by default, it will break compatibility. I think we should discuss it again in forum before changing it. I [discussed it](http://discuss.cocos2d-x.org/t/how-about-use-armeabi-v7a-as-default-abi-on-android/17247) last time, and many developers don't agree with it.

minggo:
[ci rebuild]

build/android-build.py
minggo, line 57: since it doesn't modify template, so there is not problem.

PR #16094, opened 2016-07-12
Match parent's camera mask when adding child.

stevetranby:
Let me know if you need test or anything else.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -948,6 +948,7 @@ void Node::addChildHelper(Node* child, int localZOrder, int tag, const std::stri
         child->setName(name);
     
     child->setParent(this);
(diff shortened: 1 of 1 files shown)


minggo:
@stevetranby could you please connect to the issue? Thanks.

minggo:
[ci rebuild]

ricardoquesada:
@stevetranby this patch broke the Scene3dTest. No nodes are shown on that test (cpp-tests -> Scene3dTest)

Any idea how improve this patch so that it works with the Scene3dtest ?

Perhaps this patch if fixing a real bug, but I'm worried that perhaps some games are relying on this bug... like Scene3dtest.

@minggo any idea what should we do?

Thanks.

stevetranby:
@ricardoquesada yep, will look into it (see discussion)

PR #16096, opened 2016-07-12
Fix a bug, in lua can't get the "backClicked" but a nil

coderwj:
in lua, I registerScriptKeypadHandler,when debug on Andriod, after clicking back button, I didn't get "backClicked",but a nil 
I see that in CHANGELOG:
cocos2d-x-3.1  May.24 2014
    [FIX]           EventKeyboard::KeyCode: key code for back button changed from KEY_BACKSPACE to KEY_ESCAPE
so,I fix it , and get the right "backClicked"


minggo:
thanks

PR #16099, opened 2016-07-12
Bump simulator version number to 3.12

mogemimi:
Simulator version number does not seem to be updated in v3.12. This patch bumps the current version to 3.12 (July 6, 2016). Thanks.


minggo:
thanks

PR #16101, opened 2016-07-12
fix buffer over-read in CCGLProgram.cpp in memcmp call (Address Sanitizer...

mtak-:
… spots it easily)
- previously the flow went `memcmp`, if that `!= 0`, do a bounds check, etc
- now do a bounds check, if `destSize >= srcSize`, do `memcmp`, etc

This bug manifests when you use any of the `setUniform*v` methods, with a buffer of size `x`, then on a subsequent frame, data of a larger size, `y`.

The fix was swapping the locations of two if clauses, nothing else was changed.

### DIFF
--- a/cocos/renderer/CCGLProgram.cpp
+++ b/cocos/renderer/CCGLProgram.cpp
@@ -640,17 +640,17 @@ bool GLProgram::updateUniformLocation(GLint location, const GLvoid* data, unsign
     }
     else
     {
-        if (memcmp(element->second.first, data, bytes) == 0)
+        if (element->second.second < bytes)
         {
-            updated = false;
+            GLvoid* value = realloc(element->second.first, bytes);
+            memcpy(value, data, bytes);
+            _hashForUniforms[location] = std::make_pair(value, bytes);
         }
         else
         {
-            if (element->second.second < bytes)
+            if (memcmp(element->second.first, data, bytes) == 0)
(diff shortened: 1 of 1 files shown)


minggo:
It looks good to me.
@ricardoquesada could you pleas take a look?

ricardoquesada:
[ci rebuild]

ricardoquesada:
good catch!  Did you use a tool to find it? which one? Thanks!

mtak-:
Used this tool: http://clang.llvm.org/docs/AddressSanitizer.html
Which is built into Xcode: http://stackoverflow.com/questions/32150924/how-do-you-enable-clang-address-sanitizer-in-xcode

dumganhar:
There is a crash in reported by `testin`. It's just a cpp-empty-test. The crash information is:
screen shot 2016-07-22 at 14 36 39

I think this Pull Request should fix the crash. Thanks @mtak-

dumganhar:
Oops, it's a critical bug started from `cocos2d 2.x`...

https://github.com/cocos2d/cocos2d-x/blob/cocos2d-2.0-rc0a-x-2.0/cocos2dx/shaders/CCGLProgram.cpp#L266

mtak-:
I'd think this was unlikely to fix any crashes. Just the occasional graphical glitch.

PR #16105, opened 2016-07-13
Solve the error effect when using hex map with stagger-axis is x.

natural-law:
@ricardoquesada 
I found that the render effect is wrong when using hex map with stagger-axis is x. It can be fixed by this PR. Please review it. Thanks!

PS:
Here is a sample map for reproduce the **error effect**. The error effect means it's different with the effect in **Tiled Editor**.
[hex-sample.zip](https://github.com/cocos2d/cocos2d-x/files/360654/hex-sample.zip)


ricardoquesada:
please, add a test case that shows this bug.
Actually I added the code that your patch is removing because the hex maps were totally broken, and I added 4 tests cases for that.
Please add a new test case that proves that the current code is broken, and make sure that this patch doesn't break the existing tests.
thanks!

natural-law:
@ricardoquesada 
I've just added a test case. Also I have tested the existing tests. They're all right as before.

ricardoquesada:
@natural-law your patch breaks hexagonal maps with staggered x.
With your patch, they look like this:
screen shot 2016-07-13 at 6 08 35 pm
screen shot 2016-07-13 at 6 08 42 pm

Please, run test "33" and 32" from TMX tests

natural-law:
I'm sorry that I have not noticed the details. Then could you please solve the problem I mentioned in the PR? I've no idea how to solve it. Thanks very much!

ricardoquesada:
please, update the PR with only the test case (and not the patch), and I'll take a look at it. thanks.

natural-law:
@ricardoquesada 
I've updated the PR with only the test case. Thanks!

natural-law:
The correct effect in Tiled Editor:
image

The error effect in Tests project:
image

ricardoquesada:
Added a new PR that fixes this issue and includes your tests case:
https://github.com/cocos2d/cocos2d-x/pull/16121

PR #16107, opened 2016-07-13
fix crash in EaseExponentialOut::clone()

cezheng:
This is just an obvious bug fix. It always returns nullptr.


cezheng:
[ci rebuild]

minggo:
yep, thanks

PR #16108, opened 2016-07-13
Add missing override keyword

mogemimi:
When compiling cocos2d-x v3.12 with Android NDK r11 and Clang, I get the following warning message about `-Winconsistent-missing-override`. This patch adds missing `override` keyword at the end of function declaration. Thanks.

```
cocos2d/cocos/ui/UIEditBox/UIEditBoxImpl-android.h:73:18: warning: 'setNativeMaxLength' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    virtual void setNativeMaxLength(int maxLength);
                 ^
```


minggo:
thanks

PR #16112, opened 2016-07-13
Additional checks in findChild

sbrednikhin:
Checks must help to avoid nullptr dereference or finding node witch tag or name was not initialized.

### DIFF
--- a/cocos/base/ccUtils.cpp
+++ b/cocos/base/ccUtils.cpp
@@ -370,7 +370,10 @@ Node* findChild(Node* levelRoot, const char* name)
 {
     if (levelRoot == nullptr)
         return nullptr;
-
+        
+    if ( !name || strcmp(name,"") )
+        return nullptr;
+        
     // Find this node
     {
         auto target = levelRoot->getChildByName(name);
@@ -393,6 +396,9 @@ Node* findChild(Node* levelRoot, int tag)
     if (levelRoot == nullptr)
         return nullptr;
 
+	if (tag == Node::INVALID_TAG)
+		return nullptr;
+		
     // Find this node
     {
         auto target = levelRoot->getChildByTag(tag);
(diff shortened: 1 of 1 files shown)


cocos/base/ccUtils.cpp
```diff
@@ -393,6 +396,9 @@ Node* findChild(Node* levelRoot, int tag)
     if (levelRoot == nullptr)
         return nullptr;
 
+	if (tag == Node::INVALID_TAG)
```
Dimon4eg: You use tabs indentation but in cocos spaces are used.

cocos/base/ccUtils.cpp
```diff
@@ -370,7 +370,10 @@ Node* findChild(Node* levelRoot, const char* name)
 {
     if (levelRoot == nullptr)
         return nullptr;
-
+        
+    if ( !name || strcmp(name,"") )
```
Dimon4eg: strcmp(name,"") returns 0 if strings are equal so this check is incorrect.

halx99:
[ci rebuild]

halx99:
I think the condition should be "nullptr == name || '\0' == name[0] " is better.

cocos/base/ccUtils.cpp
```diff
@@ -393,6 +396,9 @@ Node* findChild(Node* levelRoot, int tag)
     if (levelRoot == nullptr)
         return nullptr;
 
+    if (tag == Node::INVALID_TAG)
+    return nullptr;
```
minggo, line 400: please fix indent

minggo:
> I think the condition should be "nullptr == name || '\0' == name[0] " is better.

How about pass std::string reference instead of `char*`?

halx99:
Yes, It's better: https://github.com/cocos2d/cocos2d-x/pull/16117

minggo:
thanks @halx99 
@sbrednikhin how do you think about @halx99 's PR?

minggo:
I merged #16117, so close this one.
@sbrednikhin thanks for your PR.

PR #16114, opened 2016-07-13
Use fstat instead of fseek and ftell for performance and portability

mogemimi:
Hello,
I suggest we use `fstat()` instead of `fseek()` and `ftell()` when getting the size of a file, for the following reasons:
1. `FileUtils::getContents()` uses `fseek()` and `ftell()` to get the file size, but `fseek(fp, 0, SEEK_END)` has no portability for a binary stream, and the code is not secure. (for more details, please see **NOTE** below)
2. `fstat()` is faster than `fseek()` and `ftell()`— concretely, it is around 1000 times faster on my Mac — because `fseek()` is not a system call, but a library function. `fstat()` is a system call and likely to be _O(1)_.
3. `fstat()` is not the C/C++ standard, it's [part of the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fstat.html); nevertheless it works well on all target platforms which include Linux, Mac, Android, iOS and Windows. Actually, the `<sys/stat.h>` has already included at `CCFileUtils.cpp:42`, and indeed Android or Windows platforms don't use `FileUtils::getContents()` in favour of a platform-specific implementation (platform-overrides) such as `FileUtilsAndroid::getContents()` or `FileUtilsWin32::getContents()`, so there is no problem.
#### NOTE

Section 7.19.3.9 of [ISO/IEC 9899:TC2](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf) says:

> 228) Setting the file position indicator to end-of-file, as with fseek(file, 0, SEEK_END), **has undefined behavior for a binary stream** (because of possible trailing null characters) or for any stream with state-dependent encoding that does not assuredly end in the initial shift state.

Additionary, section 7.19.9.2 of [ISO/IEC 9899:TC2](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf) says:

> A binary stream need not meaningfully support fseek calls with a whence value of SEEK_END.

Also, [std::fseek reference](http://en.cppreference.com/w/cpp/io/c/fseek) says:

> Binary streams are not required to support SEEK_END, in particular if additional null bytes are output.
#### Simple performance benchmark

``` cpp
FileUtils::Status FileUtils::getContents(const std::string& filename, ResizableBuffer* buffer)
{
    // ...

    // 1. Use fseek and ftell
    auto start1 = std::chrono::high_resolution_clock::now();
    fseek(fp, 0, SEEK_END);
    size_t size1 = ftell(fp);
    fseek(fp, 0, SEEK_SET);
    auto end1 = std::chrono::high_resolution_clock::now();

    // 2. Use fstat instead
    auto start2 = std::chrono::high_resolution_clock::now();
#if defined(_MSC_VER)
    auto descriptor = _fileno(fp);
#else
    auto descriptor = fileno(fp);
#endif
    struct stat statBuf;
    if (fstat(descriptor, &statBuf) == -1) {
        return Status::ReadFailed;
    }
    size_t size = statBuf.st_size;
    auto end2 = std::chrono::high_resolution_clock::now();

    // Check backward compatibility
    CC_ASSERT(size1 == size);

    // Performance benchmark
    static double totalTime1 = 0;
    static double totalTime2 = 0;
    totalTime1 += std::chrono::duration_cast<std::chrono::duration<double>>(end1 - start1).count();
    totalTime2 += std::chrono::duration_cast<std::chrono::duration<double>>(end2 - start2).count();
    CCLOG("(1) = %lf\n(2) = %lf", totalTime1, totalTime2);

    // ...
}
```

Thank you for your time and consideration!

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -640,9 +640,16 @@ FileUtils::Status FileUtils::getContents(const std::string& filename, ResizableB
     if (!fp)
         return Status::OpenFailed;
 
-    fseek(fp, 0, SEEK_END);
-    size_t size = ftell(fp);
-    fseek(fp, 0, SEEK_SET);
+#if defined(_MSC_VER)
+    auto descriptor = _fileno(fp);
+#else
+    auto descriptor = fileno(fp);
+#endif
+    struct stat statBuf;
+    if (fstat(descriptor, &statBuf) == -1) {
+        return Status::ReadFailed;
+    }
+    size_t size = statBuf.st_size;
 
     buffer->resize(size);
     size_t readsize = fread(buffer->buffer(), 1, size, fp);


minggo:
> and indeed Android or Windows platforms don't use FileUtils::getContents() in favour of a platform-specific implementation (platform-overrides) such as FileUtilsAndroid::getContents() or FileUtilsWin32::getContents(), so there is no problem.

Current implementation works like this, but it may change. We can not depends on it.

mogemimi:
@minggo OK, I understood, thanks. 👍  I'll try to investigate further to provide more details and run the tests with real devices for Android, Win32 and UWP platforms as soon as I'm back.

ricardoquesada:
I think we can apply this PR.
If in the future we change our code (for whatever reason), we can also revert this patch.

mogemimi:
Thank you for review and comments @ricardoquesada, and I agree with you.

Also, I tried to run the test to see whether or not `fstat` is enabled on some platforms such as Android, Win32 and UWP. As far as I investigated, it works fine on the following devices that I have.
- ASUS Zenfone 5 (Android 4.4.2)
- Windows 10 (Bootcamp-ed on Mac)

I found [`fstat` reference](https://msdn.microsoft.com/en-us/library/221w8e43.aspx) in MSDN, and there is no matter while reading through the page. Android and Tizen are basically Linux kernel-based OS, so I guess it works well.

minggo:
@mogemimi thanks for testing and the information
@ricardoquesada yep, i will merge it

mogemimi:
@minggo Thank you for merge! :relaxed:

PR #16117, opened 2016-07-14
#Use std::string reference instead of char* for utils::findChild

halx99:


tests/cpp-tests/Resources/ccs-res
```diff
@@ -1 +1 @@
-Subproject commit 892e3cbdc27d24c39c62d700a81be27e593823b7
```
minggo: please don't modify the submodules

halx99: OK, I will revert its.

sbrednikhin:
Sorry for late answer.
For me  '\0' == name[0] looks less clear than strcmp. And using std::string reference can be uncomfortable to use (and possibly breaks backward capability for users, who will use this method from 3.12 revision).

halx99:
What's about 0 == strlen(name)?
Why uncomfortable, const std::string& parameter type can accept std::string or c style string.

minggo:
yep, it will not break compatibility, that's why i merge it.

sbrednikhin:
I'v checked. Looks like you absolutely right. My bad. 
Yes, this variant is much better.

PR #16118, opened 2016-07-14
Add ETC1 builtin alpha support.

halx99:


dumganhar:
Hi, @halx99 , thanks for adding this cool feature.
Could you also add a test case to `tests/cpp-tests`?

halx99:
@dumganhar I will add test case later.

cocos/2d/CCAtlasNode.cpp
```diff
@@ -102,7 +102,7 @@ bool AtlasNode::initWithTexture(Texture2D* texture, int tileWidth, int tileHeigh
     _quadsToDraw = itemsToRender;
 
     // shader stuff
-    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP));
+    setGLProgramState(GLProgramState::getPositionTextureColorGLProgramState(texture)); // setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP));
```
ricardoquesada: why was this line changed? Does it have anything to do with ETC1?

ricardoquesada:
@halx99 could you explain a little big what the patch does? 
I reviewed it and the shader API and new shaders were added. But what does it have to do with ETC1 Alpha support?
Is ETC1 a texture format? if so, why do we need the change the shaders?
Thanks!

halx99:
please see: http://malideveloper.arm.com/resources/sample-code/etcv1-texture-compression-and-alpha-channels/

halx99:
@minggo This PR is finished, and chould be merged.

minggo:
@halx99 thanks, i will review it ASAP.

cocos/renderer/ccShaders.cpp
```diff
@@ -92,4 +92,7 @@ NS_CC_BEGIN
 #include "renderer/ccShader_CameraClear.vert"
 #include "renderer/ccShader_CameraClear.frag"
 
+// x-studio365 spec
+#include "renderer/ccShader_ETC1AS_PositionTextureColor.frag.ipp"
+#include "renderer/ccShader_ETC1AS_PositionTextureGray.frag.ipp"
```
minggo: why the file ends with `.ipp`? what does it mean?

halx99: It's from c++ boost library, It's mean the file contains impl-code which included in a cpp file.
It may let people regard the file as a c++ source code  more easy than without .ipp.

cocos/renderer/CCTexture2D.cpp
```diff
@@ -453,6 +454,11 @@ Texture2D::~Texture2D()
 #if CC_ENABLE_CACHE_TEXTURE_DATA
     VolatileTextureMgr::removeTexture(this);
 #endif
+	if (_alphaTexture != nullptr)
+	{ // x-studio365 spec, ANDROID ETC1 ALPHA supports.
+		_alphaTexture->release();
+		_alphaTexture = nullptr;
+	}
```
minggo: can use `CC_SAFE_RELEASE_NULL(_alphaTexture)` to simplify codes

halx99: OK

halx99: Any others? Let me modify & commit once.

cocos/renderer/CCTexture2D.cpp
```diff
@@ -495,6 +501,11 @@ GLuint Texture2D::getName() const
     return _name;
 }
 
+GLuint Texture2D::getAlphaName() const
+{
+    return _alphaTexture == nullptr ? 0 : _alphaTexture->getName();
+}
```
minggo, line 503: getAlphaName seems strange, i don't know it returns alpha texure's name. But i don't have better suggestion.

cocos/renderer/CCTexture2D.h
```diff
@@ -410,6 +410,9 @@ class CC_DLL Texture2D : public Ref
 
     std::string getPath()const { return _filePath; }
 
+    void setAlphaTexture(Texture2D *pAlpha);
```
minggo: pAlpha -> alphaTexture?

cocos/renderer/CCTexture2D.h
```diff
@@ -410,6 +410,9 @@ class CC_DLL Texture2D : public Ref
...
+	Texture2D* getAlphaTexture(){ return _alphaTexture; }
```
minggo: add const please and fix the indent issue

cocos/renderer/CCTexture2D.h
```diff
@@ -547,6 +550,8 @@ class CC_DLL Texture2D : public Ref
 
     bool _valid;
     std::string _filePath;
+
+	Texture2D* _alphaTexture;
```
minggo: indent issue

cocos/renderer/CCTexture2D.cpp
halx99, line 503: What's about getAlphaTextureName?

cocos/renderer/CCTextureCache.cpp
```diff
@@ -47,25 +47,32 @@ using namespace std;
 
 NS_CC_BEGIN
 
+static std::string s_etc1AlphaFileSuffix = "Alpha";
```
minggo: i am not sure if using `Alpha` as suffix for alpha picture's name is good or not, maybe `@alpha` or `_alpha` is better. @ricardoquesada any suggestion?

And it is better to define it as static member variable as `TextureCache`.

halx99: I agree with you. Use lower case asset file name is better.

cocos/renderer/CCTextureCache.cpp
```diff
@@ -288,16 +305,24 @@ void TextureCache::addImageAsyncCallBack(float dt)
...
-                _textures.insert( std::make_pair(asyncStruct->filename, texture) );
+                _textures.insert(std::make_pair(asyncStruct->filename, texture));
                 texture->retain();
-                
+
                 texture->autorelease();
-            } else {
+                // x-studio365 spec, ETC1 ALPHA supports.
+                if (asyncStruct->imageAlpha.getFileType() == Image::Format::ETC) {
+                    auto alphaTexture = new Texture2D();
```
minggo: it is better to use `std::nothrow` as others and check the result.

cocos/renderer/CCTextureCache.cpp
```diff
@@ -288,16 +305,24 @@ void TextureCache::addImageAsyncCallBack(float dt)
...
+                    alphaTexture->initWithImage(&asyncStruct->imageAlpha, asyncStruct->pixelFormat);
```
minggo: check the result?

halx99: ok

cocos/renderer/CCTextureCache.cpp
```diff
@@ -345,14 +370,28 @@ Texture2D * TextureCache::addImage(const std::string &path)
...
+                _textures.insert(std::make_pair(fullpath, texture));
+
+                //-- x-studio365 spec, ANDROID ETC1 ALPHA SUPPORTS.
+                std::string alphaFullPath = path + s_etc1AlphaFileSuffix;
+                if (image->getFileType() == Image::Format::ETC && !s_etc1AlphaFileSuffix.empty() && FileUtils::getInstance()->isFileExist(alphaFullPath))
+                {
+                    Image alphaImage;
+                    if (alphaImage.initWithImageFile(alphaFullPath))
+                    {
+                        Texture2D *pAlphaTexture = new Texture2D;
```
minggo: it is better to use `std::nothrow` and check the result

cocos/renderer/CCTextureCache.cpp
```diff
@@ -345,14 +370,28 @@ Texture2D * TextureCache::addImage(const std::string &path)
...
+                        pAlphaTexture->autorelease();
```
minggo: it is better to invoke `release` at the end of this block

cocos/renderer/CCTextureCache.cpp
```diff
@@ -345,14 +370,28 @@ Texture2D * TextureCache::addImage(const std::string &path)
...
+                        pAlphaTexture->initWithImage(&alphaImage);
```
minggo: check result?

cocos/renderer/CCTrianglesCommand.h
```diff
@@ -68,7 +68,7 @@ class CC_DLL TrianglesCommand : public RenderCommand
      @param mv ModelView matrix for the command.
      @param flags to indicate that the command is using 3D rendering or not.
      */
-    void init(float globalOrder, GLuint textureID, GLProgramState* glProgramState, BlendFunc blendType, const Triangles& triangles,const Mat4& mv, uint32_t flags);
+    void init(float globalOrder, GLuint textureID, GLProgramState* glProgramState, BlendFunc blendType, const Triangles& triangles,const Mat4& mv, uint32_t flags, GLuint textureID1 = 0);
```
minggo: `textureID1` here is used for ect1 alpha texture. But the name seems strange. Can it be used for other purpose?

cocos/renderer/ccShader_ETC1AS_PositionTextureColor.frag.ipp
```diff
@@ -0,0 +1,60 @@
+/*
```
minggo: please remove `ipp` suffix as other files.

cocos/renderer/CCTrianglesCommand.h
halx99: No other purpose currently, It should modify to alphaTextureID

cocos/renderer/ccShader_ETC1AS_PositionTextureColor.frag.ipp
halx99: ok

cocos/renderer/ccShader_ETC1AS_PositionTextureColor.frag.ipp
```diff
@@ -0,0 +1,60 @@
...
+varying vec2 v_texCoord;
+
+void main() {
+    vec4 texColor = vec4(texture2D(CC_Texture0, v_texCoord).rgb, texture2D(CC_Texture1, v_texCoord).r);
+
+    texColor.rgb *= texColor.a; // Premultiply with Alpha channel
+    
+    gl_FragColor = v_fragmentColor * texColor;
+}
+);
```
minggo: is it a shader? but it uses c++ codes and cocos2d-x codes.

halx99: Yes, it's just like other builtin shaders.

cocos/renderer/CCTexture2D.cpp
minggo, line 503: it is better

cocos/renderer/CCTrianglesCommand.h
minggo: `alphaTextureID` is more meaningful.

cocos/renderer/CCTexture2D.cpp
halx99, line 503: ok

cocos/renderer/ccShader_ETC1AS_PositionTextureColor.frag.ipp
minggo: yep, you use `STRINGIFY`. Sorry for my careless.

minggo: But i think `extern CC_DLL` is not needed as other fragment shaders.

halx99: ok, remove it

halx99:
And there is anothor issue when I want run the cpp-test: it could not compile by VS2015: The error info is:
qq 20160718121714

cocos/renderer/CCGLProgramState.h
```diff
@@ -241,6 +241,9 @@ class CC_DLL GLProgramState : public Ref
 {
     friend class GLProgramStateCache;
 public:
+    /** x-studio365 spec, update node program state */
+    static GLProgramState* getPositionTextureColorGLProgramState(Texture2D* texture, bool noMVP = true);
+    static GLProgramState* getPositionTextureGrayGLProgramState(Texture2D* texture);
```
minggo: I think it is not a good idea to add functions special for `position_texture_color` shaders.

halx99: It's just wrapper for easy to update any other subclass of Node's position_texture_color shaders, do you have a good idea for do this?

minggo:
@halx99 thanks for the PR, it is a good feature. Now the PR modify the behavior of many Classes, and seems to a hot fix for current implementation:
- TrianglesCommand
  - add default parameter in `init(float globalOrder, GLuint textureID, GLProgramState* glProgramState, BlendFunc blendType, const Triangles& triangles,const Mat4& mv, uint32_t flags, GLuint textureID1)`. it seems strange, why pass two texture ids separated with other parameters? should't texture id lies near each other? can not get alpha texture id by `textureID`?
  - `init(float globalOrder, GLuint textureID, GLProgramState* glProgramState, BlendFunc blendType, const Triangles& triangles,const Mat4& mv)` this function doesn't support alpha texture?
  - in `TrianglesCommand::generateMaterialID()`, the material id generated by `{ glProgram, (int)_textureID, (int)_blendType.src, (int)_blendType.dst}`, but what's the relationship between alpha texture and material id?
- GLProgramState
  -  add `static GLProgramState* getPositionTextureColorGLProgramState(Texture2D* texture, bool noMVP = true)` and `static GLProgramState* getPositionTextureGrayGLProgramState(Texture2D* texture)` special for `position_texture_color`. If we add these two functions, then should we add functions to get other types shaders?

So i wonder if we can add a new type of Sprite that supports alpha texture or just add logical in Sprite only to support alpha texture without hotfix in `GLProgramState`, `TriangleCommand` and even `TextureCache`.

halx99:
But the are too many types needs for this, such as Label, ui::Button, AtlasNode, and etc.
My implement is aim that developer do not need to do any changes for support etc1 alpha texture.
They only need generate separated etc1 textures.

halx99:
init(float globalOrder, GLuint textureID, GLProgramState\* glProgramState, BlendFunc blendType, const Triangles& triangles,const Mat4& mv):
I realaize this function is deprecated, so I do not let it support alpha texture

minggo:
@halx99 ok, we should modify other classes, but we have to find a better way to fix these issues:
- does `textureID1` needed in `TriangleCommand`, since we can get it from `Texture2D`
- how should we generate material id after adding alpha texture
- how to implement `getPositionTextureColorGLProgramState` more elegant? It is not good to add it in `GLProgramState`.

halx99:
ok, let's think about it...

halx99:
1. does textureID1 needed in TriangleCommand, since we can get it from Texture2D:
   Yes, but currently, TrianglesCommand::init do not use Texture2D\* as parameter.
2. how should we generate material id after adding alpha texture:
   I search the whole engine source code, the material id is seems only for adjust some condition,
   I think the origin material id is  enough regarded as a major.
3. how to implement getPositionTextureColorGLProgramState more elegant? It is not good to add it in GLProgramState:
   may be move to other location for helpers functions

minggo:
> does textureID1 needed in TriangleCommand, since we can get it from Texture2D: Yes, but currently, TrianglesCommand::init do not use Texture2D\* as parameter.

Yep, it passes texture id. I think it is better to add a new `init` function is better than add default parameters. And pass two ids near each other is better.

> how to implement getPositionTextureColorGLProgramState more elegant? It is not good to add it in GLProgramState: may be move to other location for helpers functions

yep, it is better.

> how should we generate material id after adding alpha texture: I search the whole engine source code, the material id is seems only for adjust some condition, I think the origin material id is enough regarded as a major.

I am not sure. Let's ask @ricardoquesada 's help.

halx99:
What's about pass Texture2D\* for new `TriangleCommand::init` function?

minggo:
> What's about pass Texture2D\* for new TriangleCommand::init function?

It sounds good to me. @ricardoquesada is there any potential issue if using `Texture2D*`?

ricardoquesada:
@minggo @halx99 yes, adding a new `TriangleCommand::init()` with `Texture2D*` instead of `GLuint textureID` is good to me.

Regarding the original question about material Id: originally the renderer in v3 was designed as a Command Buffer renderer, where commands could be sorted by material Id (or not).
But we couldn't finished that idea, and don't have a real material Id in TriangleCommand.

However, we added a real Material System somewhere in v3.7, but uses MeshCommand I think.

minggo:
@ricardoquesada thanks, so it will not effect the material id.
@halx99 could you please add a new init function that passes `Texture2D*`?

halx99:
OK, I will do it for this, what's about follow: 
9oyifo34zacs9jn r d9f a

minggo:
it looks good to me

minggo:
@halx99 thanks. I think the only issue of this PR is 

> how to implement getPositionTextureColorGLProgramState more elegant? It is not good to add it in GLProgramState: may be move to other location for helpers functions

halx99:
@minggo  Yes, do you have a good idea for where should we move them to?
                 What's about ccUtils?

minggo:
@halx99 i have not better idea than moving to `ccUtils`.

halx99:
@minggo But if move to ccUtils, there are some file need include ccUtils.h

minggo:
@halx99 include a header file is not a big deal. Let's ask @ricardoquesada if he has better idea.

halx99:
OK, and what's about anothor two singleton classes: GLProgramCache and GLProgramStateCache?

minggo:
> OK, and what's about anothor two singleton classes: GLProgramCache and GLProgramStateCache?

Sorry, what did you mean?

halx99:
My mean is move to one of them

ricardoquesada:
@minggo @halx99 

I wouldn't move `getPositionTextureColorGLProgramState` to ccUtils. It should be in any of the `CCGL*.*` files since it is only used in the context of GL Programs / Shaders.

I have two minor questions:
- `getPositionTextureColorGLProgramState()`: its name says nothing to me. Can we improve the name? What does it do? I noticed that it replaced the old `getOrCreate...`. Does it do a "get or create" ? if so, I would add the prefix `getOrCreate`.
- `x-studio...`: I noticed that lot of the changes include the "x-studio.." signature. I guess it is because it was added by the "x-studio" company, but the credits should be added in the AUTHORS file and not in the implementation file. Imagine of each of the contributors added their names to each change... the code would be full of unneeded data.

halx99:
@ricardoquesada 
1. Yes it does a "get or create", let's add getOrCreate prefix
2.The signature `x-studio...` for easy to merge this feature to cocos2d-x latest version, because I implement this feature at early version: since 3.6, I can remove this signature. it's not a company name, it's a game develop tool which are developing

minggo:
@ricardoquesada the function is just to get/create for a special shader(textrure_positon_color supports alpha texture or not), please refer to the codes for its meaning. What i mean is that, it is a function to just get a special shader, so it should not to be part of GLProgramState. It we add this function, then should we add a function for each shader?

> My mean is move to one of them

@halx99 i think they are different, one to cache GLProgram, one to cache GLProgramState. And it will break compatibility.

halx99:
@minggo What's about change ```getPositionTextureColorGLProgramState``` --> ```getOrCreateWithGLProgramTexture(Texture2D*, const std::string& name)```

minggo:
@halx99 i think we can rename the function name like this:
- `getPositionTextureColorGLProgramState(Texture2D* texture, bool noMVP)` -> `getOrCreateWithGLProgramName(const std::string& glProgramName, const Texture2D* texture2d)`
- remove `getPositionTextureGrayGLProgramState` , can use `getOrCreateWithGLProgramName(const std::string& glProgramName, const Texture2D* texture2d)` instead

halx99:
@minggo  I also think about this, but only PositionTextureColor & Gray shaders need have correspond ETC1 shaders. then the implement source code will be:

```
GLProgramState* GLProgramState::getOrCreateWithGLProgramName(const std::string& glProgramName, Texture2D* texture)
{
    if (texture != nullptr && texture->getAlphaTextureName() != 0) {
        if (glProgramName == GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR) {
            return GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_ETC1AS_POSITION_TEXTURE_COLOR);
        }
        else if (glProgramName == GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP) {
            return GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_ETC1AS_POSITION_TEXTURE_COLOR_NO_MVP);
        }
        else if (glProgramName == GLProgram::SHADER_NAME_POSITION_GRAYSCALE) {
            return GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_ETC1AS_POSITION_TEXTURE_GRAY_NO_MVP);
        }
    }

    return GLProgramState::getOrCreateWithGLProgramName(glProgramName);
}
```

minggo:
@halx99 i think it is good. And we can add more logic for other shaders if needed.

halx99:
@minggo ok, let's do for this.

minggo:
thanks @halx99

halx99:
@minggo You are welcome, I will be very happy if you add me to AUTHORS.

minggo:
@halx99 sure, i will. the authors are not maintained for a long time. If you can send a pull request to change the file, i will merge it. Yep, you can send a PR to modify it too to add your information.

cocos/2d/CCProgressTimer.cpp
```diff
@@ -78,7 +78,7 @@ bool ProgressTimer::initWithSprite(Sprite* sp)
     setSprite(sp);
 
     // shader state
-    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR));
+    setGLProgramState(sp->getGLProgramState()/*GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR)*/);
```
minggo, line 81: @halx99 why need to modify like this. Now the effect of test case of `Actions-progress` is wrong. The position is wrong.

halx99, line 81: let's check it.

minggo, line 81: Any idea? I think we should not change it like this, ProgressTimer needs to use its own shader. I will send a PR to revert it.

minggo, line 81: i sent #16295 to fix it

minggo:
@halx99 i think we should add `getOrderOfArrival()` back, or it may break compatibility.

ricardoquesada:
@minggo are you talking about this PR? or the one that is no longer using `stable_sort()`?

halx99:
@ricardoquesada see: https://github.com/cocos2d/cocos2d-x/pull/16262

ricardoquesada:
@halx99 yes, exactly. I was referring to that PR. My guess is that @minggo wanted to add his latest comment to that PR, and not this one.

minggo:
@ricardoquesada yep, i should add comment in that PR. Anyway we discussed in another PR and found the functions are removed since v3.12, not in this branch. And we found there is not need to invoke these functions, so keep as it is now.

young40:
Great feature!

stevetranby:
@minggo This **BREAKS** sprites custom shaders on sprites because it modifies `GLProgramState` on `Sprite::setTexture` now and so causes issues with sprites that have custom shader and change their texture (or `SpriteFrame`).

cocos/2d/CCSprite.cpp
```diff
@@ -289,7 +289,7 @@ bool Sprite::initWithTexture(Texture2D *texture, const Rect& rect, bool rotated)
         _quad.tr.colors = Color4B::WHITE;
         
         // shader state
-        setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP));
+        // setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP));
```
stevetranby, line 292: Why did you change where this is set?

cocos/2d/CCSprite.cpp
```diff
@@ -365,6 +365,8 @@ void Sprite::setTexture(const std::string &filename)
 
 void Sprite::setTexture(Texture2D *texture)
 {
+    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP));
```
stevetranby, line 368: If this needs to be here it should reset its own GLProgramState or otherwise make sure to check that it's not a different shader.

halx99:
I think you can apply your custom shader after the call at external, or override the setTexture do this.

It's for update internal shader automatically.

stevetranby:
@halx99 yes I can workaround this issue easily by reverting your code, but that requires custom changes to the engine which I'd rather not have. 

If this stays in the engine as is I'll probably derive Sprite and override its setTexture, but that's also less optimal than just using Sprite.

Fundamentally this breaks compatibility with previous code (though we may be the only ones who are affected or just the only one who noticed).

It should test whether the alpha texture exists before resetting it, should it not?

stevetranby:
@halx99 yes I can workaround this issue easily by reverting your code, but that requires custom changes to the engine or subclassing Sprite which I'd rather not have to do, but I will if its determined this PR and its changes are desired as they currently exist.

The main problem is that setTexture is called within setSpriteFrame which is set inside Animate which makes re-setting the GLProgramState after every time a little cumbersome. Especially when it should just leave the shader state alone unless the Sprite is using ETC1.

If this stays in the engine as is I'll probably derive Sprite and override its setTexture, but I'd prefer to not have to change my code upgrading within a v3.x version (I do understand there have been breaking changes between minor versions).

halx99:
Actually, the function GLProgramState::getOrCreateWithGLProgramName(const std::string&, Texture2D*) will check wheter alphaTexture exist.

stevetranby:
@halx99 yes it does, but setTexture does not (which is what is required to make it work for ALL sprites, and not just ETC1)

halx99:
You can send a PR for fix this incompatible.

stevetranby:
Well I'm discussing here because I'm not sure what ramifications this has, but I will submit a new PR so it can be discussed in isolation.

halx99:
ok

halx99:
what's about:

```
if(texture->getAlphaTextureName() > 0) {
 setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
}
```

And resume original shader setup code.

stevetranby:
Yes, but then the same code in `initWithTexture` needs to be uncommented.

```
// shader state
setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP)); 
```

I'm not sure how this affects everything without testing, but that should do it for our issues.

halx99:
Yes

stevetranby:
That's probably enough since setTexture isn't called that often and a simple condition compared to applying parent transforms is probably of negligible performance cost.

However it would be smart to re-check this API to determine if cramming everything into a single API within the Sprite class is desired when the use cases tend to vary. Probably again something to not worry about until design for v4.

Thanks. I'll submit a PR soon.

cocos/2d/CCLabel.cpp
```diff
@@ -525,6 +525,25 @@ void Label::reset()
     setRotationSkewX(0);        // reverse italics
 }
 
+//  ETC1 ALPHA supports, for LabelType::BMFONT & LabelType::CHARMAP
+static Texture2D* _getTexture(Label* label)
+ {
```
zilongshanren, line 530: why make an inner class in this function and cast the base class pointer as the derived class pointer...
Why not just add a new function in FontAtlas?

halx99: I think just add a new function in FontAtlas is better.

PR #16119, opened 2016-07-14
Fix browserVersion detection

moljac024:
Due to typo, the code meant for browser versions below 6.2 would
never run.


ricardoquesada:
thanks for the patch!

Issue #16100, opened 2016-07-12
"Dirty" variables incorrectly reset with a multiple camera setup causing drawing issues on Sprites


wildrabbit:
On our project we have a scene with two cameras: the default one, with 2D nodes (mostly sprites), and a new one we create ourselves to render a 3D scene. Some of the nodes rendered using the default camera begin initially offscreen, and at a later point we scroll them to enter the visible area. However, they're not showing.

After debugging for quite a while to rule out possible problems on our client code related to clipping or incorrect positioning, we found evidence pointing to the two cameras as the likely suspect. I started digging a bit deeper and spotted a potential bug on the visiting/drawing logic.

The method Node::processParentFlags() resets the variables _transformUpdated and    _contentSizeDirty, and that happens as soon as we visit the node with the first of the scene cameras, regardless that the node's camera mask determines that it won't be rendered by that camera. Interestingly, there's a commented check on processParentFlags() that checks if the node will be visited by the current camera and returns early if it won't.

```
uint32_t Node::processParentFlags(const Mat4& parentTransform, uint32_t parentFlags)
{
    if(_usingNormalizedPosition)
    {
        CCASSERT(_parent, "setNormalizedPosition() doesn't work with orphan nodes");
        if ((parentFlags & FLAGS_CONTENT_SIZE_DIRTY) || _normalizedPositionDirty)
        {
            auto& s = _parent->getContentSize();
            _position.x = _normalizedPosition.x * s.width;
            _position.y = _normalizedPosition.y * s.height;
            _transformUpdated = _transformDirty = _inverseDirty = true;
            _normalizedPositionDirty = false;
        }
    }

    //remove this two line given that isVisitableByVisitingCamera should not affect the calculation of transform given that we are visiting scene
    //without involving view and projection matrix.

//    if (!isVisitableByVisitingCamera())
//        return parentFlags;

    uint32_t flags = parentFlags;
    flags |= (_transformUpdated ? FLAGS_TRANSFORM_DIRTY : 0);
    flags |= (_contentSizeDirty ? FLAGS_CONTENT_SIZE_DIRTY : 0);


    if(flags & FLAGS_DIRTY_MASK)
        _modelViewTransform = this->transform(parentTransform);

    _transformUpdated = false;
    _contentSizeDirty = false;

    return flags;
}
```

Then let's change to a Sprite's draw() method.  

```
    if (visitingCamera == defaultCamera) {
[...]

PR #16122, opened 2016-07-14
testcase: add test case for issue #16100

ricardoquesada:
Fixes github issue #16100

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1180,12 +1180,11 @@ uint32_t Node::processParentFlags(const Mat4& parentTransform, uint32_t parentFl
         }
     }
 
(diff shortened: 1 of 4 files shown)


ricardoquesada:
[ci rebuild]

tools/cocos2d-console
```diff
@@ -1 +1 @@
-Subproject commit 2ceba42b89945229ac22d09c0864c86b113f3f17
```
minggo, line 1: I think there is not need to change console submodule here.

ricardoquesada, line 1: oops. thanks. I'll fix it.

ricardoquesada, line 1: Apparently it was fixed here: https://github.com/cocos2d/cocos2d-x/commit/3e33b7a1cad14dab8bd896c270e319e2d12a78e3
thanks.

PR #16127, opened 2016-07-15
Add missing fclose() to fix a possible bug caused by PR #16114

mogemimi:
This patch adds a missing `fclose()` to fix a possible bug that caused by PR #16114 .
I'm sorry, it's my mistake. :sweat: Thanks.


minggo:
thanks

PR #16128, opened 2016-07-15
[ci skip]Android navigator

minggo:


minggo:
this PR fix the bug that, virtual button is not hidden after keyboard dismissed, bug report [here](http://discuss.cocos2d-x.org/t/android-hiding-keyboard-breaks-full-screen/30737/).

minggo:
corresponding forum issue: http://discuss.cocos2d-x.org/t/android-hiding-keyboard-breaks-full-screen/30737

PR #16143, opened 2016-07-18
Update the test case of hex map with axis is x.

natural-law:
Related issue : https://github.com/cocos2d/cocos2d-x/pull/16105

@ricardoquesada 
Once there are more tiles in the hex map with axis is x. The effect is error again. 
Could you please check it once more? Thanks.

BTW, I think the existing test cases `32:TMX Hex Odd X` & `33:TMX Hex Even X` are not the common usage of hex map. Because the `png` file used by the `tmx` file is for **Axis is Y**:


natural-law:
The correct effect should be:
image

But the effect in Tests project is:
image

ricardoquesada:
thanks!
Hopefully this PR fixes all the hexagonal staggered bugs:
https://github.com/cocos2d/cocos2d-x/pull/16165

ricardoquesada:
fixed in #16165

PR #16147, opened 2016-07-18
[audio] Fixes some issues while stop/uncache right after play/preload on OSX/iOS platform.

dumganhar:


bolso80:
Hi @dumganhar 
Maybe the defines of VERY_VERY_VERBOSE_LOGGING should have been commented?
Because are a lot of logs on debug...

PR #16148, opened 2016-07-18
Fix misspelled enum: POLYGEN -> POLYGON

mogemimi:
I noticed that there are minor misspellings in the `PhysicsShape::Type` enum. `POLYGEN` should be `POLYGON`, `EDGEPOLYGEN` should be `EDGEPOLYGON`.

This pull request fixes them and also keeps backward compatibility of API.
Thanks.


minggo:
thanks

PR #16151, opened 2016-07-19
Fix spelling for `getPolygonCenter` function

mogemimi:
This PR deprecates the misspelled `getPolyonCenter` function in favour of correctly spelled `getPolygonCenter` in PhysicsShape class. It does the following:
- Replaces the `getPolyonCenter` declaration and definition with `getPolygonCenter`
- Adds deprecated `getPolyonCenter` function to keep backward compatibility
- Lua binding is also the same as C++ code

Thanks.


minggo:
thanks

PR #16154, opened 2016-07-19
[BugFix]apply contentScale calculating label letter width

cezheng:
Currently the label text formater has a bug when calculating letter width of bitmap fonts, which causes the multi-line text wrapping to look weird.

Say if we have a label created by the following code:

``` cpp
  auto label = Label::createWithBMFont(bitmapFontFile, "");
  label->setAlignment(cocos2d::TextHAlignment::RIGHT);
  label->setColor(cocos2d::Color3B::WHITE);
  label->setString("000000\n111111\n010101");
  addChild(label);
```

The actually label looks like this
screen shot 2016-07-19 at 11 09 52 pm

After applying this fix, it looks like this
screen shot 2016-07-19 at 11 08 58 pm

If we take a look at how `letterPosition.x` is calculated, we will see that it is devided by `contentScaleFactor`, but the line involved does not apply `contentScaleFactor` to `letterDef.width` causing `tokenRight` to be miscalculated.

### DIFF
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -218,7 +218,7 @@ bool Label::multilineTextWrap(std::function<int(const std::u16string&, int, int)
                     nextLetterX += _horizontalKernings[letterIndex + 1];
                 nextLetterX += letterDef.xAdvance * _bmfontScale + _additionalKerning;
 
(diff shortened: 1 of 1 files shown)


cezheng:
Please do not merge this PR yet, since I found that when contentScaleFactor == 1.0, the problem still exists. I will add another commit to fix this problem once I figure it out.

cezheng:
This fix is not correct, I opened another PR for this problem: #16172

PR #16159, opened 2016-07-20
New PR based on @Jaxz PR - Rewrote all action ease classes using templates

stevetranby:
Duplicated Jakz PR to maybe get into 3.13 since it appears to be inactive since April.

I'm also thinking that maybe there could be a more generic data driven Action refactor where there's only one or a couple actual functions instead of the many functions that currently exist. This may not be as performant, however, so will have to test out an idea or two and I'll post discussion on forum first if I implement something.

Original PR Desc:

> Rewrote all action ease classes using templates to remove the redundancy of having a single class definition for every possible ease function. This has been done through some template classes which accept functions as arguments.
> 
> This makes both header and source file much shorter and much more maintanable. Readability shouldn't be a concern since now there is almost no code to read.
> 
> Performance should identical since everything is bound at compile time. The following patch should be compatible with all recent (C++11) compilers but I have no way to test it, I tested it on LLVM3.4 on a OS X machine.
> 
> This patch also gives the possiblity to have ready to go custom ease function without the need to sublass ActionEase at all (as long as the prototype of the function is supported).

### DIFF
--- a/cocos/2d/CCActionEase.cpp
+++ b/cocos/2d/CCActionEase.cpp
@@ -103,6 +103,7 @@ ActionInterval* ActionEase::getInnerAction()
 // EaseRateAction
 //
 
+/*
 EaseRateAction* EaseRateAction::create(ActionInterval* action, float rate)
 {
     EaseRateAction *easeRateAction = new (std::nothrow) EaseRateAction();
@@ -115,6 +116,7 @@ EaseRateAction* EaseRateAction::create(ActionInterval* action, float rate)
     delete easeRateAction;
     return nullptr;
 }
+*/
 
 bool EaseRateAction::initWithAction(ActionInterval *action, float rate)
 {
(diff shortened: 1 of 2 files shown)


minggo:
It seems the PR is blocked with binding codes. Currently, the binding codes can not work well with template, i will try to fix it, but i am not sure if i can fix it or not.

stevetranby:
I can give it a try. Unless it's quick for you? Are you interested in generated or are there currently any manually (or semi-auto) generated bindings that could be done in this case?

Again, it's possible a more generalized functor/lambda would work just as well or even better for the template versions? I'll give a little thought to this tonight/tomorrow.

minggo:
@stevetranby i am not quicker than you. About the binding generator you can refer to [this repo](https://github.com/cocos2d/bindings-generator). It has detail information.

If lambda can work, i think it is better. Template is more hard to debug and read.

cezheng:
Maybe keeping the original interface is better, or at least marking them as deprecated. This just PR breaks existing code.

stevetranby:
@cezheng what code does it break? It should be equivalent to the original interface, it's just that anything that needs to be updated for some reason across all easing actions now only needs one code change instead of 20+. 

That said, maybe this is best left for v4 where a better tween/animation system can be put in place using even more modern c++14/17 (and ideally some of the c++ TS)

minggo:
close it since #16335 is merged

PR #16162, opened 2016-07-20
Fixed the formarRenderers() function of UIRichText.

m-yukio:
- Fixed an issue where contentSize is not updated.
- Fixed a problem position of the child node is not correct.


moonwith:
what's next about this PR,@m-yukio,@minggo

minggo:
Since @ricardopetrere doesn't work for cocos2d-x now, so i will take a look of this PR.
@m-yukio will you please fix the conflict? Thanks.

m-yukio:
As time passes since pull request, I think that there are many differences, so please wait for a while.

Issue #16046, opened 2016-07-06
Using chrono::steady_clock() instead of gettimeofday for FPS calculation


dumganhar:
Hackers could modify system time to speed up game.
Game running should not depend on system time, it should depend on CPU clock time which could not be modified.

BTW, some articles said that the invocation of `clock_gettime` is faster and more precise than `gettimeofday`.

@minggo , any ideas?

minggo:
Yep, sounds good.

> BTW, some articles said that clock_gettime is faster and more precise than gettimeofday.

Do you have any link of the doc?

dumganhar:
I found a Chinese article here: http://www.cnblogs.com/leaven/archive/2011/03/21/1989946.html

You could google the function information to see whether we should use it. :)

minggo:
`clock_gettime()` is not cross platform, may be we can use [std::clock()](http://en.cppreference.com/w/cpp/chrono/c/clock).

Guy-kun:
To switch away from system time and time synchronization shifts in my game I changed ccUtils::getTimeInMilliseconds() to 

`long long getTimeInMilliseconds()
{
    return std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now()).time_since_epoch().count();
}`

This works on android, osx/ios and windows without any issues to me.
clock_gettime is not available on OSX

Maybe someone could review and use this in both gettime and getTimeInMilliseconds since the gettimeofday implementation shifts on android devices time to time during gameplay

dumganhar:
@Guy-kun, thanks for reminding that we can use `std::chrono::steady_clock`. I will try that.

dumganhar:
Documentation updated: 
- 1

http://en.cppreference.com/w/cpp/chrono/steady_clock

> Class std::chrono::steady_clock represents a monotonic clock. The time points of this clock cannot decrease as physical time moves forward. This clock is not related to wall clock time (for example, it can be time since last reboot), and is most suitable for measuring intervals.
> std::chrono::steady_clock meets the requirements of TrivialClock.
- 2

http://stackoverflow.com/questions/16177295/get-time-since-epoch-in-milliseconds-preferably-using-c11-chrono

> unsigned long milliseconds_since_epoch =
>     std::chrono::system_clock::now().time_since_epoch() / 
>     std::chrono::milliseconds(1);
> although, especially since you want platform independence, it might be better to replace unsigned long with a type that's more likely to be large enough:
> 
[...]

PR #16164, opened 2016-07-20
Uses std::chrono::steay_clock for calculating dt.

dumganhar:


dumganhar:
Fixed https://github.com/cocos2d/cocos2d-x/issues/16046

PR #16172, opened 2016-07-20
fix label text formatter right alignment

cezheng:
Current text formatter for label looks weird when label's alignment is set to right.

``` cpp
  auto label = Label::createWithBMFont(fontFilePath, "");
  label->setAlignment(cocos2d::TextHAlignment::RIGHT);
  label->setColor(cocos2d::Color3B::WHITE);
  label->setString("0000000000\n1111111111\n0101010101");
  addChild(label);
```

This PR changes to use the beginning of the next letter when determining the width of a line, so that right alignment does not look weird anymore.

Before:
img_0831

After:
img_0830

### DIFF
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -218,7 +218,7 @@ bool Label::multilineTextWrap(std::function<int(const std::u16string&, int, int)
                     nextLetterX += _horizontalKernings[letterIndex + 1];
                 nextLetterX += letterDef.xAdvance * _bmfontScale + _additionalKerning;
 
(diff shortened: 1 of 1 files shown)


minggo:
I tested on Mac and it looked as expected, thanks.

dumganhar:
[font_numbers.zip](https://github.com/cocos2d/cocos2d-x/files/615746/Archive.zip)
@cezheng, please use this bmfont to test it again. I found that there are still some problems.
Besides, this PR broke https://github.com/cocos2d/cocos2d-x/issues/16866

cezheng:
Um, seems there're some side effects on some cases. I'll take a look soon

dumganhar:
@cezheng , any update on this?

PR #16178, opened 2016-07-21
Fix bugs with music not resuming when iOS app is reactivated

SaxonDruce:
The audio manager's automatic resign/active behaviour aims to pause the music when the app goes to the background, and resume it again when it comes back to the foreground.

This used to work in older versions of cocos, but appears to have been broken by https://github.com/cocos2d/cocos2d-x/commit/26a04b38f2bd879d5705b38f55470f9fd9423df6

This change pauses instead of stops the music on resign.

The [audioSource stop] causes the music to be stopped, and therefore it fails to resume later on a call to [audioSource resume], due to the addition of the if (!stopped) check in resume added in https://github.com/cocos2d/cocos2d-x/commit/26a04b38f2bd879d5705b38f55470f9fd9423df6

### DIFF
--- a/cocos/audio/ios/CDAudioManager.m
+++ b/cocos/audio/ios/CDAudioManager.m
@@ -644,7 +644,7 @@ static BOOL configured = FALSE;
         case kAMRBStopPlay:
             
             for( CDLongAudioSource *audioSource in audioSourceChannels) {
-                if (!audioSystem->systemPaused) {
+                if (!audioSource->systemPaused) {
                     if (audioSource.isPlaying) {
                         audioSource->systemPaused = YES;
                         audioSource->systemPauseLocation = audioSource.audioSourcePlayer.currentTime;
--- a/cocos/audio/ios/CDAudioManager.m
+++ b/cocos/audio/ios/CDAudioManager.m
@@ -647,7 +647,7 @@ static BOOL configured = FALSE;
                 if (audioSource.isPlaying) {
                     audioSource->systemPaused = YES;
                     audioSource->systemPauseLocation = audioSource.audioSourcePlayer.currentTime;
-                    [audioSource stop];
+                    [audioSource pause];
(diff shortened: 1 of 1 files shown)


SaxonDruce:
In this situation:
1. Start game, music plays
2. Switch to Music app, game music paused, start other music
3. Switch to game, game music not resumed due to other music playing
4. Switch to Music app, stop other music
5. Switch back to game, game music should resume due to no other music playing

At step 5 the game music doesn't currently resume. This is because at step 4 when switching to the Music app, the game music gets re-paused (actually isPlaying is false, so systemPaused is set to NO, even though the music is still paused). This causes the music to not be resumed at step 5.

This change fixes this, by skipping the pause logic if systemPaused is already true.

minggo:
@dumganhar could you please help to review it?

dumganhar:
Thanks @SaxonDruce , I will try to reproduce this issue by the step you provided.

SaxonDruce:
Thanks @dumganhar.

To reproduce, put this in application:didFinishLaunchingWithOptions in AppController.mm, to set up the auto resign behaviour:

```
[[CDAudioManager sharedManager] setResignBehavior: kAMRBStopPlay autoHandle: YES];
```

And then in AppDelegate::applicationDidFinishLaunching() in AppDelegate.cpp, play some music:

```
CocosDenshion::SimpleAudioEngine::getInstance()->playBackgroundMusic(
    "path/to/music.mp3", true)
```

dumganhar:
@SaxonDruce , ok,  I have already tried to reproduce but failed.
I'll follow your reply and test it again.

dumganhar:
@minggo , this patch works as expected. It's good for being merged.

minggo:
thanks guys

SaxonDruce:
Great, thanks @minggo and @dumganhar!

PR #16183, opened 2016-07-21
Implement willPlayBackgroundMusic() on Android

SaxonDruce:
Implement willPlayBackgroundMusic(), so that we can check if music is already playing in another app.

### DIFF
--- a/cocos/audio/android/jni/cddandroidAndroidJavaEngine.cpp
+++ b/cocos/audio/android/jni/cddandroidAndroidJavaEngine.cpp
@@ -102,7 +102,7 @@ void AndroidJavaEngine::rewindBackgroundMusic() {
 }
 
 bool AndroidJavaEngine::willPlayBackgroundMusic() {
-    return true;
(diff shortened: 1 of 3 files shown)


minggo:
So developers should check the result of `willPlayBackgroundMusic` before playing music?

SaxonDruce:
Yeah, that seems to be the logic on iOS anyway, eg something like this:

```
    CocosDenshion::SimpleAudioEngine* simpleAudioEngine =
        CocosDenshion::SimpleAudioEngine::getInstance();
    if (simpleAudioEngine->willPlayBackgroundMusic())
    {
        simpleAudioEngine->setBackgroundMusicVolume(1.0f);
        simpleAudioEngine->playBackgroundMusic("path/to/music.mp3", true);
    }
```

On iOS, the willPlayBackgroundMusic() also returns true/false depending on whether other audio is already playing in another app.

minggo:
thanks

Issue #16197, opened 2016-07-22
sdkbox imports should be implemented at Cocos2dxActivity


AlexanderBykin:
I'm tired to modify Activity sdkbox to use it, so i want to do ease usage of it by adding simple flag of enable.

i will provide PR.

minggo:
@AlexanderShniperson i don't quite understand what's your need, could you please paste some codes to explain it?

AlexanderBykin:
@minggo sure i will provide Java class that initialize SDKBOX from Cocos2dxActivity

AlexanderShniperson:
#16203

PR #16203, opened 2016-07-25
issue_16197

AlexanderBykin:
#16197


minggo:
@AlexanderShniperson thanks, because SDKBOX is not open source, and these codes are just used for SDKBOX, so i can not merge it. I think you can connect SDKBOX guys, and contribute `SdkBoxInitializer.java` to it. Then you can subclass Cocos2dxActivity to use it.

PR #16214, opened 2016-07-26
fixed #16169: new android project crash in android 5.0.2 device (Nexus 7) when use 3.12

### LINKED ISSUE #16169: new android project crash in android 5.0.2 device (Nexus 7) when use 3.12
This is the process to see this bug:

```
1. cocos new test_project -p com.test.test -l cpp
2. cd test_project
3. python /Users/jun/Code/cocos2d-x-3.12/tools/cocos2d-console/bin/cocos.py  compile -p android --android-studio --ap android-15 --ndk-mode release --mode release  --output-dir .
4. adb install test_project-release-signed.apk
```

When run this test app in android 5.0.2 device (Nexus 7) , it will crash!  but in android 4.4 and android 6.0, it was no problem.

The ndk version is r11c

dumganhar:
While using clang compiler with gnustl_static on android, the copy assignment operator of `std::u16string` will affect the memory validity, it means after `newChars` is destroyed, the memory of `u16Text` holds will be a dead region. `u16text` represents the variable in `Label::_utf16Text`, when somewhere allocates memory by `malloc, realloc, new, new[]`, the generated memory address may be the same as `Label::_utf16Text` holds. If doing a `memset` or other memory operations, the orignal `Label::_utf16Text` will be in an unknown state. Meanwhile, a bunch lots of logic which depends on `Label::_utf16Text` will be broken.

### DIFF
--- a/cocos/2d/CCFontAtlas.cpp
+++ b/cocos/2d/CCFontAtlas.cpp
@@ -265,7 +265,7 @@ void FontAtlas::findNewCharacters(const std::u16string& u16Text, std::unordered_
     //find new characters
     if (_letterDefinitions.empty())
     {
-        newChars = u16Text;
+        newChars.append(u16Text);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


minggo:
i changed to use gcc, but this PR is good with using gcc or clang, so i merge it

Issue #9968, opened 2015-01-13
Crash on Android devices with arm-v7a, but without NEON


ondesly:
LOCAL_ARM_NEON := true in cocos/Android.mk and APP_ABI := armeabi-v7a in Application.mk causes crash on launch (SIGILL) on Android devices with CPU arm-v7a, but without NEON instructions.

Such CPU is Nvidia Tegra 2. It supports arm-v7a, but it doesn't support NEON instructions.
Devices: Samsung Galaxy Tab 10.1 (P7500), Asus Transformer TF101, Motorola XOOM and others.

minggo:
As you said, the device doesn't support NEON instructions, why you use `LOCAL_ARM_NEON := true`?

ondesly:
I don't. It's enabled in cocos/Android.mk by default for all arm-v7a devices.

dabingnn:
Because Neon support can not be checked when compiling the code.
So the solution for neon on android is that the neon code will always be include in the apk file, and when code is running, CPU features is checked, if neon is enabled, math computation will use NEON, or C implementation will be used.
Please refer to 
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/math/MathUtil.cpp

Could you please attach some crash logs to give us more info? Thanks

minggo:
[this document](http://www.kandroid.org/ndk/docs/CPU-ARM-NEON.html) says

> LOCAL_ARM_NEON:
> Define LOCAL_ARM_NEON to 'true' in your module definition, and the NDK will build all its source files with NEON support. This can be useful if
> you want to build a static or shared library that specifically contains NEON code paths.

So we can not use `LOCAL_ARM_NEON := true` to select neon version of math library.

minggo:
@ondesly 
Is it work correctly if you comment `LOCAL_ARM_NEON := true`?
Could you please help to check it because we don't have the devices you mentioned.

Yehsam23:
I comment LOCAL_ARM_NEON := true....then i works fine on my device.

My device is Asus Transformer TF101.

So...we can comment this?

minggo:
@Yehsam23 
Yep, it is safe to comment this line.
We use it to select neon version of math library.
As you know, neon instructions are more efficient.

dabingnn:
@Yehsam23 
I submitted a pr #10157，it works on armv7 with neon support, could you try it?
I don't know if it works on armv7-a without neon

ondesly:
[...]

PR #16233, opened 2016-07-27
Android import cpufeatures in cocos.

perminovVS:
Add import cpufeatures internal.
Now cpufeatures imported in 3rd party libs webp.
If not use webp, android not build.

### DIFF
--- a/cocos/Android.mk
+++ b/cocos/Android.mk
@@ -280,9 +280,11 @@ LOCAL_WHOLE_STATIC_LIBRARIES += cpufeatures
 # define the macro to compile through support/zip_support/ioapi.c
 LOCAL_CFLAGS   :=  -DUSE_FILE32API
 LOCAL_CFLAGS   +=  -fexceptions
-ifeq ($(TARGET_ARCH_ABI),armeabi-v7a)
-    LOCAL_CFLAGS += -DHAVE_NEON=1
-endif
+
+#ifeq ($(TARGET_ARCH_ABI),armeabi-v7a)
+#    LOCAL_CFLAGS += -DHAVE_NEON=1
+#endif
+
 LOCAL_CPPFLAGS := -Wno-deprecated-declarations
 LOCAL_EXPORT_CFLAGS   := -DUSE_FILE32API
 LOCAL_EXPORT_CPPFLAGS := -Wno-deprecated-declarations
--- a/cocos/Android.mk
+++ b/cocos/Android.mk
@@ -281,6 +281,7 @@ LOCAL_WHOLE_STATIC_LIBRARIES += cpufeatures
 LOCAL_CFLAGS   :=  -DUSE_FILE32API
 LOCAL_CFLAGS   +=  -fexceptions
 
+# Issues #9968
 #ifeq ($(TARGET_ARCH_ABI),armeabi-v7a)
 #    LOCAL_CFLAGS += -DHAVE_NEON=1
 #endif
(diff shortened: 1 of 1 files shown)


cocos/Android.mk
```diff
@@ -275,10 +275,14 @@ LOCAL_STATIC_LIBRARIES += recast_static
...
 
 LOCAL_WHOLE_STATIC_LIBRARIES := cocos2dxandroid_static
+LOCAL_WHOLE_STATIC_LIBRARIES += cpufeatures
 
 # define the macro to compile through support/zip_support/ioapi.c
 LOCAL_CFLAGS   :=  -DUSE_FILE32API
 LOCAL_CFLAGS   +=  -fexceptions
+ifeq ($(TARGET_ARCH_ABI),armeabi-v7a)
+    LOCAL_CFLAGS += -DHAVE_NEON=1
+endif
```
minggo: Should check armeabi-v8 too?

minggo:
> If not use webp, android not build.

I will check it.

cocos/Android.mk
minggo: sorry, it is `arm64-v8a`

minggo:
Yep, because `MathUtils.cpp` depends on cpu-feature, and this issue fix it.

cocos/Android.mk
minggo: And as i know, `armeabi-v7a` may doesn't support neon, can refer to #9968 for detail information. So we can not add this line.

minggo: @perminovVS what's your idea about it? And i think the modification should not be part of this PR.

perminovVS: @minggo i get it from https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/blob/v3/webp/prebuilt/android/Android.mk

If armeabi-v7a cpu not always support neon instruction, need check it runtime and change use code.

I see MathUtil.cpp, code:

``` c++
bool MathUtil::isNeon32Enabled()
{
#ifdef USE_NEON32
    return true;
#elif (defined (INCLUDE_NEON32) && (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID) )
    class AnrdoidNeonChecker
    {
    public:
        AnrdoidNeonChecker()
        {
            if (android_getCpuFamily() == ANDROID_CPU_FAMILY_ARM && (android_getCpuFeatures() & ANDROID_CPU_ARM_FEATURE_NEON) != 0)
                _isNeonEnabled = true;
            else
                _isNeonEnabled = false;
        }
        bool isNeonEnabled() const { return _isNeonEnabled; }
    private:
        bool _isNeonEnabled;
    };
    static AnrdoidNeonChecker checker;
    return checker.isNeonEnabled();
#else
    return false;
#endif
}
```

It now implemented, additionally do not need to do anything.
For arm64-v8a not need define HAVE_NEON

``` c++
#elif (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
    #if defined (__arm64__) || defined (__aarch64__)
    #define USE_NEON64
    #define INCLUDE_NEON64
    #elif defined (__ARM_NEON__)
    #define INCLUDE_NEON32
    #else
    #endif
#else
```

Need add log message, for mobile platform, support neon.

minggo: Will you please add a new PR for it, we can discuss there. I will merge the PR if you remove the codes.

perminovVS:
[ci rebuild]

PR #16234, opened 2016-07-27
Win32 move unicode utils global.

perminovVS:
It need for easy converting utf-8 string in all code.
And use it api in user code also WinRT version.


minggo:
thanks

PR #16250, opened 2016-07-28
Win32 LuaLog use unicode api

perminovVS:
LuaLog convert utf8 by common unicode api.
based on pr #16234


perminovVS:
common pr #16249

PR #16263, opened 2016-07-29
#Use Data::takeBuffer replace deep copy data.

halx99:
see: https://github.com/EsotericSoftware/spine-runtimes/pull/650

### DIFF
--- a/cocos/editor-support/spine/spine-cocos2dx.cpp
+++ b/cocos/editor-support/spine/spine-cocos2dx.cpp
@@ -75,11 +75,8 @@ void _spAtlasPage_disposeTexture (spAtlasPage* self) {
 }
 
 char* _spUtil_readFile (const char* path, int* length) {
-	Data data = FileUtils::getInstance()->getDataFromFile(
-			FileUtils::getInstance()->fullPathForFilename(path).c_str());
+    Data data = FileUtils::getInstance()->getDataFromFile(
+			FileUtils::getInstance()->fullPathForFilename(path));
     if (data.isNull()) return 0;
-	*length = static_cast<int>(data.getSize());
-	char* bytes = MALLOC(char, *length);
-	memcpy(bytes, data.getBytes(), *length);
-	return bytes;
+    return (char*)(data.takeBuffer((ssize_t*)length));
 }


minggo:
thanks

ricardoquesada:
this PR broke spine. it is crashing now :( I'll revert it until we find a solution that doesn't crash.

halx99:
It fetch latest v3 and test, the default spine test case not crash.

ricardoquesada:
@halx99 in which platform did you try it? which test did you try?
I tried it on Mac OS X 10.11 and run Scene3DTest and SpineTest and both crashed on the spine runtime.
I reverted this PR, and those tests started to work again.

halx99:
@ricardoquesada I run the win32.sln at Windows 10 14393

minggo:
@ricardoquesada i think you should revert it on v3.13 branch too. I will have a test too.

minggo:
@ricardoquesada i tested on Mac 10.11.6, and there is not crash.

minggo:
Oh, the codes in v3.13 is changed too. I will apply the PR and test again.

minggo:
Yep, it will crash on Mac. Let me check why.

halx99:
Think about follow reason:

cocos2d-x getDataFromFile use C global function malloc and pass 1 parameters to alloc memory
The spine MALLOC is defined as _malloc/_free and pass 3 parameters, see follow

```
#ifndef SPINE_EXTENSION_H_
#define SPINE_EXTENSION_H_

/* All allocation uses these. */
#define MALLOC(TYPE,COUNT) ((TYPE*)_malloc(sizeof(TYPE) * (COUNT), __FILE__, __LINE__))
#define CALLOC(TYPE,COUNT) ((TYPE*)_calloc(COUNT, sizeof(TYPE), __FILE__, __LINE__))
#define NEW(TYPE) CALLOC(TYPE,1)
...
/* Frees memory. Can be used on const types. */
#define FREE(VALUE) _free((void*)VALUE)
```

And at windows the `malloc/free` and `_malloc/_free` behavior is same, but on Mac maybe not.

ricardoquesada:
Found the issue. The culprit was this line:

``` c++
return (char*)(data.takeBuffer((ssize_t*)length));
```

`length` is `int`. You cannot convert it to `ssize_t` since `ssize_t` might be bigger in certain platforms, specially in 64-bit platforms.

PR #16274, opened 2016-07-29
Android cpu features neon armeabi v7a

perminovVS:
If armeabi-v7a cpu not always support neon instruction, need check it runtime and change use code.

I see MathUtil.cpp, code:

``` c++
bool MathUtil::isNeon32Enabled()
{
#ifdef USE_NEON32
    return true;
#elif (defined (INCLUDE_NEON32) && (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID) )
    class AnrdoidNeonChecker
    {
    public:
        AnrdoidNeonChecker()
        {
            if (android_getCpuFamily() == ANDROID_CPU_FAMILY_ARM && (android_getCpuFeatures() & ANDROID_CPU_ARM_FEATURE_NEON) != 0)
                _isNeonEnabled = true;
            else
                _isNeonEnabled = false;
        }
        bool isNeonEnabled() const { return _isNeonEnabled; }
    private:
        bool _isNeonEnabled;
    };
    static AnrdoidNeonChecker checker;
    return checker.isNeonEnabled();
#else
    return false;
#endif
}
```

It now implemented, additionally do not need to do anything.
For arm64-v8a not need define HAVE_NEON

``` c++
#elif (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
    #if defined (__arm64__) || defined (__aarch64__)
    #define USE_NEON64
    #define INCLUDE_NEON64
    #elif defined (__ARM_NEON__)
    #define INCLUDE_NEON32
    #else
    #endif
#else
```

Need add log message, for mobile platform, support neon.

In cpu-features.h check define **ARM_NEON** not HAVE_NEON ?

``` c++
/* define _ARM_HAVE_NEON for ARMv7 architecture if we support the
 * Neon SIMD instruction set extensions. This also implies
 * that VFPv3-D32 is supported.
 */
#if __ARM_ARCH__ >= 7 && defined __ARM_NEON__
#  define __ARM_HAVE_NEON
#endif
```

### DIFF
--- a/cocos/Android.mk
+++ b/cocos/Android.mk
@@ -280,9 +280,11 @@ LOCAL_WHOLE_STATIC_LIBRARIES += cpufeatures
 # define the macro to compile through support/zip_support/ioapi.c
 LOCAL_CFLAGS   :=  -DUSE_FILE32API
 LOCAL_CFLAGS   +=  -fexceptions
-ifeq ($(TARGET_ARCH_ABI),armeabi-v7a)
-    LOCAL_CFLAGS += -DHAVE_NEON=1
-endif
+
+#ifeq ($(TARGET_ARCH_ABI),armeabi-v7a)
+#    LOCAL_CFLAGS += -DHAVE_NEON=1
+#endif
+
 LOCAL_CPPFLAGS := -Wno-deprecated-declarations
 LOCAL_EXPORT_CFLAGS   := -DUSE_FILE32API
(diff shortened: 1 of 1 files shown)


minggo:
@perminovVS could you please fix conflict. It is because the PR includes some modification in #16233. So please create a new branch based on v3 for a new PR in future, then each PR will be independent on other PR.

minggo:
close it since it is not responsive.

PR #16275, opened 2016-07-29
new actions ResizeBy and ResizeTo

charlesstlaurent:
New actions to resize (mostly widget) instead of scaling.

### DIFF
--- a/cocos/2d/CCActionInterval.h
+++ b/cocos/2d/CCActionInterval.h
@@ -783,16 +783,16 @@ public:
     */
     static ResizeTo* create(float duration, const cocos2d::Size& final_size);
 
-CC_CONSTRUCTOR_ACCESS:
-    ResizeTo() {}
-    virtual ~ResizeTo() {}
-    
     //
     // Overrides
     //
     virtual ResizeTo* clone() const override;
     void startWithTarget(cocos2d::Node* target) override;
     void update(float time) override;
+
+CC_CONSTRUCTOR_ACCESS:
+    ResizeTo() {}
+    virtual ~ResizeTo() {}
     
     /**
     * initializes the action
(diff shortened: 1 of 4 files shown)


minggo:
could you please add a test case for it, thanks

charlesstlaurent:
Hi @minggo,  no problem, should I put the test with the other actions (but setContentSize does nothing on a sprite so I would have to add a Widget like an ImageView for example), or should I put it with the UI tests where there already is something that I can apply the action on?

minggo:
@charlesstlaurent so the actions only have effect on Widget? Why need these actions?

charlesstlaurent:
Hi @minggo, it works on all nodes where setContentSize is effective (which is not the case for Sprites). But it's mostly useful for nodes where 9-slice is enabled, because if gives different results than ScaleTo/ScaleBy in that case.

minggo:
@charlesstlaurent i think you can add test case with other actions but use widget instead.

> But it's mostly useful for nodes where 9-slice is enabled, because if gives different results than ScaleTo/ScaleBy in that case.

Could you please add comment about it too? Then developers may know how to use it correctly. Thanks.

cocos/2d/CCActionInterval.h
```diff
@@ -768,6 +768,92 @@ class CC_DLL SkewBy : public SkewTo
...
+CC_CONSTRUCTOR_ACCESS:
+    ResizeTo() {}
+    virtual ~ResizeTo() {}
+    
+    //
+    // Overrides
+    //
+    virtual ResizeTo* clone() const override;
+    void startWithTarget(cocos2d::Node* target) override;
+    void update(float time) override;
```
minggo, line 791: why put these functions in `CC_CONSTRUCTOR_ACCESS`?

charlesstlaurent:
Hi @minggo mistakes were made. It's now fixed.

PR #16279, opened 2016-07-30
fixed base URL for webview on android doesn't work

summerinsects:
fixed base URL for webview on android doesn't work


summerinsects:
[ci rebuild]

minggo:
@wohaaitinciu how can i reproduce the error?

summerinsects:
run cpp-tests -> UITest -> UIWebViewTest->press 'Load Data' button
`_webView->loadHTMLString("<body style=\"font-size:50px;\">Hello World  </body>","Images/")`
The Icon.png will not be shown in WebView. And after invoking getFixedBaseUrl, it will be OK.
This is a little mistake when replacing traditional JNI calls with JniHelper

minggo:
thanks

PR #16294, opened 2016-08-02
fixes issue #16263 correctly

ricardoquesada:
no stackoverflow when converting int to ssize_t

Related to Github issue #16263

### DIFF
--- a/cocos/editor-support/spine/spine-cocos2dx.cpp
+++ b/cocos/editor-support/spine/spine-cocos2dx.cpp
@@ -62,6 +62,7 @@ GLuint filter (spAtlasFilter filter) {
 
 void _spAtlasPage_createTexture (spAtlasPage* self, const char* path) {
 	Texture2D* texture = Director::getInstance()->getTextureCache()->addImage(path);
+	CCASSERT(texture != nullptr, "Invalid image");
 	texture->retain();
 
 	Texture2D::TexParams textureParams = {filter(self->minFilter), filter(self->magFilter), wrap(self->uWrap), wrap(self->vWrap)};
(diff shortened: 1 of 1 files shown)


ricardoquesada:
@minggo do you want me to send it to v3.13 as well? or can I apply it only in v3 to have it ready for v3.14?

ricardoquesada:
probably we should wait, since this PR won't be applied in spine now. more info here:
https://github.com/EsotericSoftware/spine-runtimes/pull/650

minggo:
@ricardoquesada i think we can move it into v3.14.

halx99:
may be there could use FileUtils::getFileData for compitiable

ricardoquesada:
@halx99 sorry, what do you mean? use `getFileData()` instead of `getDataFromFile()` ?
That's a good idea so that it can be applied in Spine.

I'm curious... why do we have these two functions? are they different?

halx99:
The `getFileData` is a very early compatible API since 2.x, The function just malloc & read file data, not manage the data memory free, caller need manage the memory to call `free` when it unused.

minggo:
`getFileData` is deprecated. Why need to use it? Any benefit?

halx99:
Well, Just for compitable for spine, like:

```
 badlogic commented 2 days ago
We have a couple of users on < 3.12, so we'll keep the copying version for
a little while. Thanks for the PR!
```

halx99:
The spine runtime needs keep the readed file data, for this case: the getFileData is well.

minggo:
@halx99 current implementation, Spine gets the file data. `data.takeBuffer(&tmpLen)` will change the owner of the pointer.

ricardoquesada:
what we should do, perhaps is to add certain defines in the spine patch

``` c++
// version higher than v3.10 ?? (not sure when `getDAtaFromFile()` was added
#if COCOS2D_VERSION >= 0x00031000
  ... = getDataFromFile();
#else // using older verison
  ... = getFileData()...
#endif
```

minggo:
@ricardoquesada i don't think we need it. I think Spine is cocos2d-x version related. We can not make sure Spine can work for all cocos2d-x versions.

minggo:
Of course, its my fault that i should not merge it since we should can not modify Spine runtime source code. I think we should revert it and discuss in https://github.com/EsotericSoftware/spine-runtimes/pull/650

halx99:
The takeBuffer is added at v3.12

minggo:
@halx99 i see, that's why i said we should revert it and discuss it in Spine repo.

minggo:
i reverted it in #16321

ricardoquesada:
@minggo ok. I put the question on the spine's issue tracker

ricardoquesada:
so, the patch was applied in spine. I'll send a new patch for cocos2d.

PR #16295, opened 2016-08-03
use correct shader

minggo:


halx99:
ok, it's good.

halx99:
@minggo I test your fix code, and for ETC1, the programTimer seems not support alpha

minggo:
@halx99 may be we should not just change the shader to support ETC1, should do more work. Will you please send a PR for it and add a test case? Thanks.

halx99:
Yes ,I will, I just work for it

minggo:
thanks, and please send to v3.13 branch.

halx99:
ok

halx99:
I done it, see the PR: https://github.com/cocos2d/cocos2d-x/pull/16296

PR #16298, opened 2016-08-03
Fix format string warning when compiling for iOS armv7

mogemimi:
This pull request fixes the following warning when compiling libcocos2d for iOS armv7 (32-bit) on Xcode 7.3.1:

```
cocos/audio/apple/AudioEngine-inl.mm:70:70: values of type 'OSStatus' should not be used as format arguments; add an explicit cast to 'int' instead [-Wformat]
```


minggo:
thanks

PR #16299, opened 2016-08-03
fix pageview turning event

zilongshanren:
1. Fix PageView turning event timing, only issue turning event when PageView is indeed turn pages
2. Fix deprecated addEventListenerPageView can't receive turning event issue.
3. Fix PageView clone missing cloning some member instances.


minggo:
[ci rebuild]

minggo:
> Fix PageView turning event timing, only issue turning event when PageView is indeed turn pages

What does it mean?

PR #16303, opened 2016-08-03
fixed #10482: New AudioEngine class can't play large ogg file on Win32.

### LINKED ISSUE #10482: New AudioEngine class can't play large OGG files
Hi,

I have tried to play longer OGG files by new AudioEngine::play2D() but it doesn't work at all. I tried several longer OGG files (>2MB).  I debugged the audio engine and the problem seems to be in AudioCache.cpp, if the sound PCM size is larger than PCMDATA_CACHEMAXSIZE.  The loading / queing thread is not calling this method again or something....
I detected this problem on Windows 8, in the latest Cocos2dx 3.4, and tested it in cpp-tests too (in class LargeAudioFileTest).

Thanks,
Attila

dumganhar:


dumganhar:
[ci rebuild]

halx99:
Why not use the common log function: cocos2d::log?

dumganhar:
cocos2d::log isn't threadsafe ...

halx99:
But the winlog do nothing for threadsafe...

dumganhar:
Hmm, yes, I remember that there's some where will cause thread competition problem.
Probably, `Director::getInstance()->getConsole()->log("balabala")`.
If `director` is ended, then some threads invoke `cocos2d::log` may trigger problems.
Therefore, the implementation of `winLog` comments that line.
Furthermore, I don't like to write `cocos2d::log` in our code directly. Instead, we need more powerful `log` system.

halx99:
What's about add cocos2d::logTSF, lock and just call cocos2d::log?

dumganhar:
We already have log print function with argument list format. Look at this:

``` c++
Android: int __android_log_print(int prio, const char *tag,  const char *fmt, ...);
Mac/iOS/Linux/Tizen: int printf(const char *fmt, ...);
Windows: OutputDebugStringW(WCHAR* buf);
```

Only Windows doesn't support to use just one function to output a `formatted` log.

It could be simply just use this definitions on non-Windows platform:

``` c++
#if defined(COCOS2D_DEBUG) && COCOS2D_DEBUG > 0
#define CCLOGV(fmt, ...) audioLog("V/" LOG_TAG " (" QUOTEME(__LINE__) "): " fmt "", ##__VA_ARGS__)
#else
#define CCLOGV(fmt, ...) do {} while(false)
#endif
#define CCLOGD(fmt, ...) audioLog("D/" LOG_TAG " (" QUOTEME(__LINE__) "): " fmt "", ##__VA_ARGS__)
#define CCLOGI(fmt, ...) audioLog("I/" LOG_TAG " (" QUOTEME(__LINE__) "): " fmt "", ##__VA_ARGS__)
#define CCLOGW(fmt, ...) audioLog("W/" LOG_TAG " (" QUOTEME(__LINE__) "): " fmt "", ##__VA_ARGS__)
#define CCLOGE(fmt, ...) audioLog("E/" LOG_TAG " (" QUOTEME(__LINE__) "): " fmt "", ##__VA_ARGS__)
```

Above, `audioLog` represents the platform log output function directly. But currently, we encapsulate a method `_log` to allocate / free buffer dynamically, then output to `CCConsole`.
It cause some issues:
- No logging tag support
- Complex
- Thread unsafe
- May trigger memory fragments if `_log` is invoked frequently

dumganhar:
> What's about add cocos2d::logTSF, lock and just call cocos2d::log?

It's not good to add a new function like `logTSF`.
Probably, we need a log function which is higher performance, threadsafe, and not log to `CCConsole`.

halx99:
So, I think `log to CCConsole` is unnecessary, and the cocos2d::log could be optimize like the `_winLog`

dumganhar:
Actually, `log to CCConsole` is needed now.
log output could be put into `terminal` for command control, for example, set FPS, update Sprite position, etc. CCConsole probably need this to work.  Therefore, I don't have a better idea for that.

Anyway, I will keep on thinking how to make our log system better. :)

PR #16314, opened 2016-08-04
fix the problem in ccUTF8.cpp and CCLuaJavaBridge.cpp

JaryGuo:
- ccUTF8: when srcjStr is nullptr, the call to `UTF16ToUTF8` crashes the app
- CCLuaJavaBridge: fix the return value to `nil` instead of empty string when java return `null`.


minggo:
@JaryGuo how to reproduce it?

JaryGuo:
call java method from lua and let the java method return null, the app will crash before fixing. After fixing the lua will receil nil.

minggo:
Thanks, could you please add a test case for it?

JaryGuo:
I tested it with cocos2d-x-3.11.1 on android.
java code

``` java
package org.cocos2dx.lua_empty_test;
import org.cocos2dx.lib.Cocos2dxActivity;
public class AppActivity extends Cocos2dxActivity {
    public static String testNullString(){
        //return "hello";
                return null;
    }
}
```

lua code

``` lua
local luaj = require "cocos.cocos2d.luaj"
local className = "org/cocos2dx/lua_empty_test/AppActivity"
local sigs = "()Ljava/lang/String;"
local ok,ret  = luaj.callStaticMethod(className,"testNullString",{}, sigs)
if not ok then
    print("luaj error:", ret)
else
    print("The ret is:", ret)
end
```

minggo:
Yep, i see. I mean could you please add a test case in `lua-tests`, then we will test it every version in future.

JaryGuo:
Ok.

minggo:
Thanks, you can add it in LuaBridgeTest.lua.

PR #16318, opened 2016-08-05
Fix pageview click turning event

zilongshanren:
Fix https://github.com/cocos2d/cocos2d-x/pull/16310

remove trailing whitespace...


cocos/ui/UIPageView.cpp
```diff
@@ -220,11 +220,12 @@ void PageView::refreshIndicatorPosition()
         _indicator->setPosition(Vec2(posX, posY));
     }
 }
-    
+
 void PageView::handlePressLogic(Touch *touch)
 {
     ListView::handlePressLogic(touch);
     if (!_isTouchBegin) {
+        _currentPageIndex = getIndex(getCenterItemInCurrentView());
```
zilongshanren, line 228: The currentPageIndex should be calculate here.

PR #16319, opened 2016-08-05
[Android] Nofity waitup only while preloadEffect(info, cb) is invoked in play2d.

dumganhar:
Currently, if a audio isn't preloaded, then play2d directly will wait preload operation to finish.


dumganhar:
Sorry, sent to wrong branch, close it, i will commit to v3.13 branch.

PR #16326, opened 2016-08-05
*Fix: GLThread and other threads access sDownloaderMap in a thread-un…

pwang08:
GLThread and other threads(`js_load_remote_image` creates CCDownloader in another thread) access `sDownloaderMap` in a thread-unsafe way, which will result in crash in some occasion on Android.

### DIFF
--- a/cocos/network/CCDownloader-android.cpp
+++ b/cocos/network/CCDownloader-android.cpp
@@ -27,6 +27,8 @@
 #include "network/CCDownloader.h"
 #include "platform/android/jni/JniHelper.h"
 
+#include <mutex>
+
 #define JCLS_DOWNLOADER "org/cocos2dx/lib/Cocos2dxDownloader"
 #define JCLS_TASK       "com/loopj/android/http/RequestHandle"
 #define JARG_STR        "Ljava/lang/String;"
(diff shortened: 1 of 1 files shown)


dumganhar:
Hi, pwang08,
I notice that `Downloader` instance is usually constructed in GL thread.
And the callback functions are also invoked from GL thread. Could this cause thread problems?

``` java
Cocos2dxHelper.runOnGLThread(new Runnable() {
            @Override
            public void run() {
                nativeOnProgress(_id, id, downloadBytes, downloadNow, downloadTotal);
            }
        });
```

pwang08:
`void __JSDownloaderDelegator::downloadAsync()
{
    retain();
    auto t = std::thread(&__JSDownloaderDelegator::startDownload, this);
    t.detach();
}
`

As the code above shows: `__JSDownloaderDelegator::startDownload` is called in another thread, in which a `Downloader` instance is created. As the code below shows:

`
void __JSDownloaderDelegator::startDownload()
{
    if (auto texture = Director::getInstance()->getTextureCache()->getTextureForKey(_url))
    {
        onSuccess(texture);
    }
    else
    {
        _downloader = std::make_shared<cocos2d::network::Downloader>();
    }
}
`

dumganhar:
@pwang08 , it makes sense.

@pandamicro , Could you please check it since if I remember correctly, __JSDownloaderDelegator stuff was written by u.

pandamicro:
Yes, it was originally created by me, at the time, there was a `performFunctionInCocosThread` to ensure everything is thread safe. But @yangws  have refactored the network module and removed these code because network it self will make sure callbacks are invoked in cocos thread.

minggo:
just close/reopen it to trigger ci. This PR seems good.

PR #16328, opened 2016-08-05
Fix size/position calculations for Linux system fonts.

adesentenac:
Here are screenshots from cpp-tests before and after the change on Linux, and the same screenshots on Windows for comparison.
- RichText XMLSmallBig test before than changes:
  The line height was incorrect (too big) and the horizontal spacing was too small ("normal" was stuck to the dot after "bigger").
  xmlsmallbig_before
- after the changes:
  xmlsmallbig_after
- on Windows:
  xmlsmallbig_windows
- XMLFace test before the changes:
  The line height was incorrect (too big).
  xmlface_before
- after the changes:
  xmlface_after
- on Windows:
  xmlface_windows

### DIFF
--- a/cocos/platform/linux/CCDevice-linux.cpp
+++ b/cocos/platform/linux/CCDevice-linux.cpp
@@ -71,7 +71,8 @@ struct LineBreakLine {
     void calculateWidth() {
         lineWidth = 0;
         if ( glyphs.empty() == false ) {
-            lineWidth = glyphs.at(glyphs.size() - 1).paintPosition + glyphs.at(glyphs.size() - 1).glyphWidth;
+            LineBreakGlyph& glyph = glyphs.at(glyphs.size() - 1);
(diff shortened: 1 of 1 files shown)


adesentenac:
Hi @minggo do you know if there is a way to force jenkins to restart, because from the logs it seems the iOS build failed because the slave went down (the change in this pull request should have no impact on iOS anyway).

minggo:
[ci rebuild]

@adesentenac you can use the above command to restart CI.

zilongshanren:
[ci rebuild]

minggo:
@adesentenac how can i reproduce it?

adesentenac:
@minggo if you run the rich text test in cpp-test on linux, you'll see that the texts have incorrect position and line height when using system fonts (cf screenshots 1 & 4)

minggo:
Yep, the test works as expected.

PR #16331, opened 2016-08-06
update visibleTiles rect

noprops:
I modified the size of visibleTiles in order to draw layer properly if contentScaleFactor is not 1.


minggo:
@noprops how to reproduce it?

noprops:
img_1345

When contentScaleFactor is 2, only one-quarter of the map is drawn.

https://github.com/noprops/FastTMXLayerTest

minggo:
When contentScaleFactor is 2?

PR #16333, opened 2016-08-06
Remove AppDelegate memory leak in Android projects

rynkowsg:
AppDelegate object and all its members are never released.
As a solution, I proposed to use static `unique_ptr` that could destroy it at the end of the application.

Before applying the aforementioned fix I corrected also code style and correctness of `main.cpp` files in Android projects:
- removed unused includes (reducing dependencies and compilation time)
- removed unused code
- corrected code style (whitespaces, headers' order, etc).


minggo:
looks good to me, thanks

PR #16337, opened 2016-08-08
V3.13 remove the /ZW option from all .c files in Win10 libSpline

stammen:
This PR removes the /ZW option from all .c files in Win10 libSpline project for all build configurations. Recent updates to spline broke the Win10 build.


stammen:
Wrong branch.

PR #16339, opened 2016-08-08
RichText improvements

adesentenac:
- Added BM Font support for text elements
- Added Sprite frame support for image elements
- Added means to align RichText elements on their baseline
  - Added means to obtain the ascent (used as baseline offset) for BM Font, TTF and system font (on Windows, Linux, and Android only for now; through Texture2D)

RichText::setAlignToBaseline supports 3 values:
- ALL: align all elements on the text baseline (default on Android, Windows and Linux).
- TEXT: align only text element on their baseline, the images positioned on the bottom.
- NONE: no baseline alignment, all elements are positioned on the bottom (same as before, default for all other platforms).

To work correctly on Linux with system fonts it needs the fix from: https://github.com/cocos2d/cocos2d-x/pull/16328

Here are screenshots to show the different alignments:
- before the changes:
  alignonbaseline_before
- NONE (as we can see, it gives the same result as before):
  alignonbaseline_none
- TEXT:
  alignonbaseline_text
- ALL:
  alignonbaseline_all

### DIFF
--- a/cocos/renderer/CCTexture2D.cpp
+++ b/cocos/renderer/CCTexture2D.cpp
@@ -1088,11 +1088,12 @@ bool Texture2D::initWithString(const char *text, const std::string& fontName, fl
     tempDef._fontFillColor = Color3B::WHITE;
     tempDef._enableWrap    = enableWrap;
     tempDef._overflow      = overflow;
+    float tempFontAscent = 0;
 
-    return initWithString(text, tempDef);
+    return initWithString(text, tempDef, tempFontAscent);
 }
 
-bool Texture2D::initWithString(const char *text, const FontDefinition& textDefinition)
+bool Texture2D::initWithString(const char *text, const FontDefinition& textDefinition, float& fontAscent)
 {
     if(!text || 0 == strlen(text))
     {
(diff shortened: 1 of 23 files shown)


cocos/renderer/CCTexture2D.cpp
```diff
@@ -1094,6 +1094,11 @@ bool Texture2D::initWithString(const char *text, const std::string& fontName, fl
 
 bool Texture2D::initWithString(const char *text, const FontDefinition& textDefinition)
 {
+    return initWithString(text, textDefinition, NULL);
```
minggo: NULL -> nullptr

minggo:
@adesentenac thanks for so a great PR, could you please
- add test cases for `BM Font support` and `Sprite frame support`
- implement `RichText::setAlignToBaseline` on all other platforms

adesentenac:
Hi @minggo I don't seem to be able to do it for Tizen, the relevant method for my needs seems to be [evas_object_text_ascent_get](https://docs.enlightenment.org/api/evas/doc/html/group__Evas__Object__Text.html#ga4aaf7a903228927fa0adac885304e43e) but it does not seem to be available for Tizen, it's commented out in tizen-sdk\platforms\tizen-2.4\mobile\rootstraps\mobile-2.4-device.core\usr\include\evas-1\canvas\evas_text.eo.legacy.h
I've never worked on Tizen so maybe one of you guys might know if there is another way to do it.

I'll test the change needed in Device for mac soon but I don't have any iOS device to test the iOS change (which is very similar to the mac code).

minggo:
@adesentenac what will happen on Tizen? Just wrong effect or will cause crash?

I will test on iOS.

minggo:
[ci rebuild]

PR #16343, opened 2016-08-09
Avoid unnecessary allocation if targetGrid is reusable

mogemimi:
This PR reduces unnecessary GridBase allocation in `GridAction::startWithTarget()`.

`GridAction::startWithTarget()` calls `this->getGrid()`, which results in a GridBase allocation such as `Grid3DAction::getGrid()` or `TiledGrid3DAction::getGrid`. If `targetGrid->getReuseGrid() > 0` is true, `targetGrid` is reusable, then these allocations can be avoided.

Thanks!


minggo:
thanks

PR #16352, opened 2016-08-11
Android arm64-v8a also uses Luajit.

dumganhar:


dumganhar:
This pull request needs https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/246 to be merged

dumganhar:
@zilongshanren, please review.

zilongshanren:
@dumganhar  looks good to me

minggo:
@dumganhar i think we should provide tool to generate 64-bit bytecode, currently `cocos luacompile` only generate 32-bit bytecode.

dumganhar:
It's a pity that cocos luacompile doesn't support to generate 64bit bytecode... : ( 
But how do developers generate 64bit luajit bytecode for 64bit bytecode now?

minggo:
In previous, it is not convenient to generate 64bit bytecode, you need to use real 64bit device(iPhone6 or so) to generate it. But @dualface told me that, now it can generate 64bit bytecode in PC, but i haven't tried it.

dumganhar:
So do you mean we should make `cocos luacompile` supports this feature in v3.13?

minggo:
Yep, i think it is better.

minggo:
@dumganhar i merged cocos2d/cocos2d-x-3rd-party-libs-bin#246 and release `deps-108`. Could you please update it in this PR?

dumganhar:
OK, I will.

dumganhar:
Updated.

PR #16359, opened 2016-08-12
Use the VS 2015 libraries for MSVC 14

Furkanzmc:
When compiling with MSVC 14, the linker cannot find `libpng-2015.lib`, `libtiff-2015.lib`, `libjpeg-2015.lib` and `libglfw-2015.lib`. This commit fixes that problem. I haven't contributed to the CMake files in cocos2dx before so I'm not really sure If this is in the right format or, really, the best way to do it but it works.


cmake/Modules/CocosUsePrebuiltLibs.cmake
```diff
@@ -4,7 +4,11 @@
 
 set(_chipmunk_inc chipmunk/chipmunk.h)
 set(_chipmunk_inc_paths include)
-set(_chipmunk_libs chipmunk libchipmunk)
+if (${MSVC_VERSION} STREQUAL "1900")
+    set(_chipmunk_libs chipmunk libchipmunk-2015)
```
minggo: where is libchipmunk-2015?

Furkanzmc: I don't understand what you mean. It is on line 7. Can you clarify?

minggo: Sorry, i found it. There is libchipmunk-2015 in external.

minggo:
Please fix compiling error:

```
CMake Error at cmake/Modules/CocosUsePrebuiltLibs.cmake:7 (if):
  if given arguments:

    "STREQUAL" "1900"

  Unknown arguments specified
  CMakeLists.txt:94 (include)


See also "/linux_slave/workspace/ccs-pr/node/linux/linux-build/CMakeFiles/CMakeOutput.log".
cmake generate error
```

Furkanzmc:
I'm not familiar with Windows Store and Phone applications, should I check for those platforms too? Is checking just for WINDOWS enough?

minggo:
I am not sure. Most people just use VS to build apps. And could you please tell me how to reproduce the issue?

Furkanzmc:
If you mean the initial issue, when using MSVC 14 the linker looks for libraries with `-2015` suffix but cannot find it. I'm much familiar with Cocos2dx's CMake files so I can't tell the root cause as I couldn't figure out where that `-2015` suffix is added :smile: But I found the `CocosUsePrebuiltLibs.cmake` file and just made the changes there to see if it works. And it did.
The second issue (Linux compilation issue) happens because of how `if` works in CMake. `if` treats `MSVC_VERSION` as a variable and it gives and error when it cannot find it or it is empty.

Since we are at it, is VS 2013 still supported on cocos2dx v3.12? because I get errors because the linker cannot find `version.lib` and whole lot of other errors because of the C implementation of the compiler. Apparently, VS 2015 changed its underlying C implemention so I thought maybe cocos2dx is using VS 2015 lib files for VS 2013 compiler too.

minggo:
yep, VS2013 is still supports. Did you use command line to compile for windows?

Furkanzmc:
No, I used Qt Creator with the default CMakeLists file.
And here's the command Qt Creator uses: `cmake --build --target-all`

Furkanzmc:
These are the errors I get while compiling v3.12 with VS 2013

```
libeay32.lib(gost_eng.obj) : error LNK2001: unresolved external symbol ___acrt_iob_func
libeay32.lib(cryptlib.obj) : error LNK2001: unresolved external symbol ___acrt_iob_func
libeay32.lib(pem_lib.obj) : error LNK2001: unresolved external symbol ___acrt_iob_func
libeay32.lib(ui_openssl.obj) : error LNK2001: unresolved external symbol ___acrt_iob_func
libeay32.lib(gost_eng.obj) : error LNK2001: unresolved external symbol ___stdio_common_vfprintf
libeay32.lib(pem_lib.obj) : error LNK2001: unresolved external symbol ___stdio_common_vfprintf
libeay32.lib(ui_openssl.obj) : error LNK2001: unresolved external symbol ___stdio_common_vfprintf
libeay32.lib(e_cswift.obj) : error LNK2001: unresolved external symbol ___stdio_common_vsprintf
libeay32.lib(cryptlib.obj) : error LNK2001: unresolved external symbol ___stdio_common_vsprintf
libeay32.lib(dso_win32.obj) : error LNK2001: unresolved external symbol ___stdio_common_vsprintf
libeay32.lib(v3_utl.obj) : error LNK2001: unresolved external symbol ___stdio_common_vsscanf
```

mirth:
Still can reproduce in Microsoft Visual Studio 2017. I created new project via 'cocos new' and generated msvc project with cmake.

Furkanzmc:
@mirth This commit only fixes the CMake files. So I don't think it will be of any help for  Visual Studio projects.

mirth:
@Furkanzmc but you've said that you've been compiling with  MSVC 14? explain please

Furkanzmc:
@mirth You mentioned the IDE name, not the compiler. So I just assumed you were using the Visual Studio project file. Is that not the case? Sorry.

mirth:
i generate visual studio project file with cmake though. can you give a link how to compile cocos project without ide in windows?

Furkanzmc:
In my experience, generating the Visual Studio project file from CMake with Cocos2dx is not the way to go. Cocos2dx already has Visual Studio projects. And they work. I suggest you use those. Or you could directly use CMake to compile. Personally, I prefer Qt Creator over Visual Studio so I'm using CMake. I'm not sure if Visual Studio has CMake support but a quick Google search gave me [this](https://channel9.msdn.com/Events/Visual-Studio/Visual-Studio-2017-Launch/T138).

mirth:
But how i should i maintain both win and os x build? So there will be two separate project for vs and code?

Furkanzmc:
Are you developing for mobile or desktop?

If you are only developing for desktop, then you can use CMake for both Windows and macOS. If you are developing for mobile, you still need to modify the platform specific project files I believe. You can also use the [Cocos2dx's command line tool](http://www.cocos2d-x.org/wiki/Cocos_compile#iOS-Arguments) to build.

mirth:
For mobile. But i want to have build for desktop too. What the option for CMake on Windos what target/compiler?

Furkanzmc:
I'm not sure If I understand the question correctly but I'm assuming you are asking for which compiler to use with CMake.

You can use MSVC on Windows.

Issue #8354, opened 2014-09-30
about 3d projection bug in Director::setProjection()


minggo:
# Note

 This issue is migrated from [here](http://www.cocos2d-x.org/issues/4275). It was created at 2014/03/05 04:08:21 +0000
# Description

If we use 3d projection mode, we will set the modelview matrix by kmMat4LookAt function, where we need to calculate the eye position whose coordniate-z is derived from the function of getZeye().
In this function, (_winSizeInPoints.height / 1.1566f) is simplication of (_winSizeInPoints.height / 2 / tanf(pi/6)) where there is some deviation, the exact value should be 1.1547005383793f.
Normally this deviation can be igonred, however if we use framebuffer(RenderTexture), we may use the projection more than once, then the deviation error will be amplified, and we may get an unwished result.

cwq:
@minggo I was wondering why not use the exact value, and use 1.1566f.

minggo:
@ricardoquesada do you have any idea?

ricardoquesada:
@minggo yes, we should use the exact value.
that number came from cocos2d-iphone, and IIRC, I added that harcoded number in cocos2d-iphone because it worked for me.

ricardoquesada:
fixed.

ricardoquesada:
reopened since `tanf(pi/6) * 2` doesn't work as expected. See comments here:  #16369

newnon:
@ricardoquesada @TheCodez i made pull request for this problem, 2 days my team spend to find why cocos crop some pixels on screen border. And it is because constant error i think  @TheCodez got artefact because wrong texture but before cocos always crop one pixel on every side of screen.

PR #16369, opened 2016-08-15
fix: uses correct constant for Eye in Director

ricardoquesada:
fixes github issue #8354

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -660,7 +660,7 @@ void Director::purgeCachedData(void)
 
 float Director::getZEye(void) const
 {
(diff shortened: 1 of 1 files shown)


ricardoquesada:
[ci rebuild]

TheCodez:
@ricardoquesada I get some strange artifacts of the background leaking 1 pixel after this change. The object has a y position of 0 so the background shouldn't be visible
Left side before the change, right after:

image

ricardoquesada:
@TheCodez thanks. I'll review it.

ricardoquesada:
@TheCodez thanks. I need to do more research. PR reverted.

Issue #16375, opened 2016-08-17
PageView Page Turning event issue


zilongshanren:
Steps to Reproduce:
1.  scroll PageView continuously
2. Only one page turning event will be triggered after scroll ended.

Expected behavior:
1. Each page turning should trigger the page turning event.

zilongshanren:
Now you could use `setAutoScrollStopEpsilon(0.5)` to resolve this issue.

PR #16379, opened 2016-08-17
Fix issue 16375

zilongshanren:
https://github.com/cocos2d/cocos2d-x/issues/16375


zilongshanren:
Not a robust solution..  just close it here.

Issue #16327, opened 2016-08-05
Large Memory Leaks on iOS-JS due to Spidermonkey


DavidDeSimone:
- cocos2d-x version: 3.11
- devices test on: iPad Air 2
- developing environments
  - NDK version: N/A
  - Xcode version: 7.2.1
  - VS version: N/A
  - brower type and version:

We have discovered several large and concerning memory leaks originating from spidermonkey in our Cocos2d-x project on the iOS platform. We have not confirmed if this issue is or isn't happening on other platforms due to lack of good memory leak tooling on our other development platform (Android). Our rough polling of Heap Size in my Android application seems to suggest that this issue is also happening on Android, but we don't have solid proof of this. The two principle leaks seem to originate from the following two actions:
- Calling eval (either directly or indirectly) on any script code will leak the string of the evaluated code.
- Calling a function that has undergone cc.Class's _super variable replacement will cause the "ArgumentsData" object of Spidermonkeys Internal "ArgumentsObject" to leak. ArgumentsObject here is spidermonkeys internal C++ representation of the "arguments" object in a Javascript function. 

We believe this is related to issue https://github.com/cocos2d/cocos2d-x/issues/16291, and that issue is caused by these leaks. 

Examining either of these action's in Apple's Instruments program produces the following data:

These are the stacks for the Eval case
![screen shot 2016-08-05 at 2 17 41 pm](https://cloud.githubusercontent.com/assets/7245174/17450856/9dbd03c4-5b17-11e6-8e26-5edec404f2d7.png)
![screen shot 2016-08-05 at 2 17 19 pm](https://cloud.githubusercontent.com/assets/7245174/17450868/a15ae46a-5b17-11e6-9cc6-54caf81a8df3.png)

This is the stack for the arguments case
![screen shot 2016-08-05 at 2 17 19 pm](https://cloud.githubusercontent.com/assets/7245174/17450888/c5303bce-5b17-11e6-80d6-b0126c42f0cc.png)

For the eval() case
Steps to Reproduce:
1. Examine a Cocos2d-x application with the Instruments program, examining memory leaks.
2. Repeated eval a string of arbitrary Javascript.
3. Watch as leaked memory rises.

For the _super case
Steps to Reproduce:
1. Examine a Cocos2d-x application with the Instruments program, examining memory leaks.
2. Run the following code:

```
var HelloWorld = cc.Layer.extend({
                    ctor:function () {
                        this._super();
                        this.scheduleUpdate();
[...]

PR #16380, opened 2016-08-17
Merge part of Cocos2d-x-lite

pandamicro:
and :
1. Update js auto binding settings with new ndk version
2. Remove unnecessary property from Ref
3. Fix spine track entry circle referencing issue
4. #16327 : Fix global object leak issue in JSB
5. Improve evalString implementation

### DIFF
--- a/cocos/scripting/js-bindings/manual/network/jsb_websocket.cpp
+++ b/cocos/scripting/js-bindings/manual/network/jsb_websocket.cpp
@@ -24,6 +24,7 @@
 #include "scripting/js-bindings/manual/network/jsb_websocket.h"
 
 #include "base/ccUTF8.h"
+#include "base/CCDirector.h"
 #include "network/WebSocket.h"
 #include "platform/CCPlatformMacros.h"
 #include "scripting/js-bindings/manual/ScriptingCore.h"
@@ -78,6 +79,9 @@ public:
         js_proxy_t * p = jsb_get_native_proxy(ws);
         if (!p) return;
 
+        if (cocos2d::Director::getInstance() == nullptr || cocos2d::ScriptEngineManager::getInstance() == nullptr)
+            return;
+
         JSB_AUTOCOMPARTMENT_WITH_GLOBAL_OBJCET
 
         JSContext* cx = ScriptingCore::getInstance()->getGlobalContext();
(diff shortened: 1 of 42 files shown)


pandamicro:
@dumganhar

cocos/scripting/js-bindings/manual/network/jsb_websocket.cpp
```diff
@@ -78,6 +79,9 @@ class JSB_WebSocketDelegate : public WebSocket::Delegate
         js_proxy_t * p = jsb_get_native_proxy(ws);
         if (!p) return;
 
+        if (cocos2d::Director::getInstance() == nullptr || cocos2d::ScriptEngineManager::getInstance() == nullptr)
```
dumganhar, line 82: Could Director::getInstance be nullptr? It will always return non-null pointer in cocos2d-x.
I think it's a copy paste logic from cocos2d-x-lite since I know that lite does it in this way.

pandamicro: Yes, it's synced from -x-lite, so in lite it's necessary and in -x it's not ? Or I should remove it from -lite also ?

pandamicro, line 82: Yes, it's synced from -x-lite, so in lite it's necessary and in -x it's not ? Or I should remove it from -lite too ?

dumganhar, line 82: -x-lite needs it. lite did some memory recycle jobs and getInstance may return nullptr while -x doesn't.

dumganhar, line 82: But if this could be compiled successfully, it's not a problem. Just a unused check.

pandamicro, line 82: ok, I think keep them synced is beneficial, who knows when we are gonna merge them

pandamicro:
@dumganhar I will merge some stuff from https://github.com/cocos2d/cocos2d-x/pull/16383 then this PR is ready to be merged.

minggo:
[ci rebuild]

minggo:
[ci rebuild]

minggo:
[ci rebuild]

web
```diff
@@ -1 +1 @@
-Subproject commit 48f6318cf66c56852d9cf09c8241d0ecb8d7989d
```
minggo: why update web submodule in `fix auto bindings`?

pandamicro: It was merged into one commit, if it's a problem, I can separate them into two commit

minggo: No, i just want to double check if it is a mistake because the submodule is updated in previous commit.

pandamicro: Updated

minggo:
[ci rebuild]

PR #16383, opened 2016-08-17
Fixing issue where the JS global object would not be finalized when c…

DavidDeSimone:
…alling __restartVM(), due to an issue where PersistentRootedObject were not being properly cleaned up.

This is being provided as a point of reference for issue https://github.com/cocos2d/cocos2d-x/issues/16327 . You may choose to iterate on this more, or go a different route to fix it. I figured it would be easier for your team if we shared the code we used to work around this issue. 

I created a new class definition to facilitate the removal of the mozilla::Maybe objects. A lot of generated and non-generated code was using the .ref() interface, and it was easier to create a class that shared an interface then change all of the code that used this interface.


pandamicro:
Merged in https://github.com/cocos2d/cocos2d-x/pull/16380

Issue #16365, opened 2016-08-15
use luajit 2.1.0-beta2


minggo:
now we use 2.1.0-alpha on iOS and use lua for Android 64-bit, we should upgrade to use luajit 2.1.0-beta2.

Edit: and some platforms use lua, should change to use luajit too.

minggo:
@dumganhar [added luajit 2.10-beta2 for Android arm64](https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/246) to fixed the bug caused by `ByteCodeEncryptTest` in `lua-tests`. So the things left are:
- update luajit to 2.1.0-beta2 for iOS, Mac and win32
- use luajit 2.1.0-beta2 instead of lua for other platforms

ricardoquesada:
@minggo I've just updated the makefile to use LuaJIT instead of Lua for Android64.

I tested it, and now lua-tests works Ok.

minggo:
Yep, it is fixed in v3.13 branch. And @dumganhar thought it is better to fix the issue in 3.13 and he is working on it. So i assign it to @dumganhar ?

ricardoquesada:
some updates:

I recompiled myself luajit-2.1-beta for:
- android (all version, using ndk12b)
- ios (all versions)
- mac (all versions)

and it only works on ios.
but it doesn't work on android (both 32-bit and 64-bit versions) and mac.

Not sure if we need to re-generate the bytecodes or not.
I traced the problem until this file:
https://github.com/cocos2d/cocos2d-x/blob/v3/tests/lua-tests/src/ByteCodeEncryptTest/ByteCodeEncryptTest.lua

and it has an special case for ios. perhaps we need to do the same for android and mac.

@dumganhar @minggo any idea?
@dumganhar how did you compile luajit 2.1-beta2 for android? because it doesn't work for me now. thanks!

ricardoquesada:
@minggo yes, if you want to assign it to him, Ok for me.
@dumganhar I added my some code here:
https://github.com/cocos2d/cocos2d-x/pull/16384

dumganhar:
I have updated it to 2.1.0-beta2, and updated cocos2d-console, 3rd-party-libs-bin, 3rd-party-libs-src. cocos2d-console could generate 64bit luajit bytecode either.

I didn't notice that this issue was assigned to @ricardoquesada .
Please check my work:

cocos2d-console: https://github.com/cocos2d/cocos2d-console/pull/368
[...]

PR #16384, opened 2016-08-17
fix: uses luajit for android 64

ricardoquesada:
instead of using Lua, it uses LuaJIT for android64

fixes github issue #16365


ricardoquesada:
don't merge it yet since Lua precompiled bytes are not working on android

minggo:
close this PR since it is included in #16620

PR #16390, opened 2016-08-18
PageView with custom 'scrollToPage' time

alfogrillo:
`PageView` has two new public member functions:

```
void scrollToPage(ssize_t idx, float time);
void scrollToItem(ssize_t idx, float time);
```

These can be useful when developer wants scroll to a Page with a custom time (maybe zero), still triggering the `PageView::EventType::TURNING` event.


minggo:
[ci rebuild]

zilongshanren:
This PR looks good to me 👍

alfogrillo:
@minggo fixed the build error

minggo:
thanks

PR #16392, opened 2016-08-18
#Fix use const std::string& instead a temp string

halx99:


minggo:
thanks

halx99:
This PR also fix the UITextField bug(return temp std::string by c styled string):

```
const char* TextField::getPasswordStyleText()const
{
    return _textFieldRenderer->getPasswordTextStyle().c_str();
}
```

halx99:
[ci rebuild]

Issue #14276, opened 2015-10-30
ios 9 opengl error


kakamatong:
When I cut back my game from an application，opengl take a error 
![20151030203314](https://cloud.githubusercontent.com/assets/12620172/10845910/adcdb64e-7f46-11e5-9277-f5df0c87ec0c.jpg)

kakamatong:
OpenGL error 0x0506 in -[CCEAGLView swapBuffers] 324

kakamatong:
use cocos2dx 3.3

ondesly:
Found the same issue on cocos2dx 3.8. I am just hiding the app by central button. It happens only if I use Launch Screen File.

failed to call context
cocos2d: surface size: 0x0
Failed to make complete framebuffer object 0x8CD6

OpenGL error 0x0506 in cocos/renderer/CCFrameBuffer.cpp applyFBO 444

FrameBuffer Status Error 36054
OpenGL error 0x0506 in cocos/renderer/CCFrameBuffer.cpp applyFBO 444

FrameBuffer Status Error 36054
FrameBuffer Status Error 36054
OpenGL error 0x0506 in cocos/renderer/CCRenderer.cpp saveRenderState 161

OpenGL error 0x0506 in cocos/renderer/CCRenderer.cpp restoreRenderState 192

FrameBuffer Status Error 36054
OpenGL error 0x0502 in -[CCEAGLView swapBuffers] 324

gestern:
**The issue is still exists at cocos2dx 3.10**

slackmoehrle:
any idea on how to reproduce this?

gestern:
> any idea on how to reproduce this?

You have to specify "Launch Screen File" in your cocos2d-x project in xCode

xiongdi9456:
The issue still exists at cocos2d-x 3.11

unexpect3rd:
I am on 3.10.
I called [[UIApplication sharedApplication] openURL:url] to open a website, and when returning to the app using the top left iOS quick return link, this happens.

And after some testing I noticed, for me, it happens when the above actions are performed and SDKBox::PluginAdmob::cache() is being called in AppDelegate::applicationWillEnterForeground().

minggo:
@xiongdi9456 how can i reproduce it?

davidejones88:
Hello,
I'm facing a similar issue with 3.12

[...]

PR #16394, opened 2016-08-19
fix: validates director before layout in iOS

ricardoquesada:
Adds Director::isValid()
removes DisplayLinkDirector()
removes virtual functions from Director()
Director is no longer an abstract class

github issue #14276

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -81,7 +81,7 @@ NS_CC_BEGIN
 // FIXME: it should be a Director ivar. Move it there once support for multiple directors is added
 
 // singleton stuff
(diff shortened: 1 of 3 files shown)


cwasp:
Fails on iOS 6, because layoutSubviews called only once, when cocos2d::Director::getInstance()->isValid() return false, so  [renderer_ resizeFromLayer:(CAEAGLLayer*)self.layer]; is never called.
- (void) layoutSubviews
  {
  -    if (!cocos2d::Director::getInstance()->isValid())
  -    {
  -        return;
  -    }
  -   
    [renderer_ resizeFromLayer:(CAEAGLLayer*)self.layer];
    size_ = [renderer_ backingSize];

Issue #16155, opened 2016-07-19
Sprite3D can't release its texture


sdspageprotein:
Create a sprite3d object, and then add it to one layer. After a while remove it from parent, then the texture it refred can't be released after calling the function removeUnusedTextures!
This is a very serious issue, because it can consume much memory when there are many 3d sprites on one scene.
The version of engine we are using is 3.8.1

minggo:
`removeUnusedTextures` just remove textures that the reference count is 1. Please make sure it is not used other places.

sdspageprotein:
This is definitely a bug, I test many times by only adding one Sprite3D object on one scene.

minggo:
@sdspageprotein could you paste some code snipper to show how you use it?

sdspageprotein:
This is one simple example:

```
local sp = cc.Sprite3D:create("plane3D/A-5.c3b")
self._core:addChild(sp)

self._core:runAction( cc.Sequence:create(
    cc.DelayTime:create(1),
    cc.CallFunc:create(function (  )
        sp:removeFromParent()
        cc.Director:getInstance():getTextureCache():removeUnusedTextures()
    end),
    cc.CallFunc:create(function (  )
        print(cc.Director:getInstance():getTextureCache():getCachedTextureInfo())
    end) ) )
```

And we may find the reason of this issue.
We change RenderState's finalization from:
RenderState::~RenderState()
{
    CC_SAFE_RELEASE(_state);
}
to 
RenderState::~RenderState()
{
    CC_SAFE_RELEASE(_texture);
    CC_SAFE_RELEASE(_state);
}

And then Sprite3D works fine. I do not know will this change cause some other issues.

dumganhar:
@minggo , for a glance, I think @sdspageprotein 's change is reasonable, if _texture isn't released, it may cause memory leak.

cwq:
can test in Sprite3DBasicTest

```
addChild( sprite );
removeChild(sprite);
```

the reference count of texture is 2, because RenderState did not release _texture

minggo:
@cwq i think you are right. @ricardoquesada could you please take a look? `_texture` is retained in `RenderState::setTexture()` but it is not released.

ricardoquesada:
fixed. thanks.

PR #16396, opened 2016-08-19
fix: release texture in RenderState

ricardoquesada:
memory leak fixed in RenderState

fixes github issue #16155

### DIFF
--- a/cocos/renderer/CCRenderState.cpp
+++ b/cocos/renderer/CCRenderState.cpp
@@ -50,6 +50,7 @@ RenderState::RenderState()
 
 RenderState::~RenderState()
 {
+    CC_SAFE_RELEASE(_texture);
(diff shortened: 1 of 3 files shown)


ricardoquesada:
[ci rebuild]

minggo:
@ricardoquesada i think this test case is hard to test. Can just use reference count to test it, for example

```
assert(reference_count >=2);
```

ricardoquesada:
@minggo you are right.
I could do that if needed. I was just checking it with the debugger and it was releasing it. the thing is that if that same texture was already cached it could affect the ref count.
that's why I chose the "quick & dirty" way to test it. 
I'll improve it later this afternoon.

ricardoquesada:
@minggo new PR here: https://github.com/cocos2d/cocos2d-x/pull/16418

PR #16397, opened 2016-08-19
Revert luajit

minggo:


halx99:
ci is dropped?

minggo:
yep, we are fixing it

minggo:
[ci rebuild]

Issue #16399, opened 2016-08-19
The latest Label calculate Size seems not correct with some .ttf font


halx99:
- cocos2d-x version: github latest
- devices test on: Windows  10 14393.82 win32 program
- developing environments
  - NDK version:
  - Xcode version:
  - VS version: 2013 or 2015
  - brower type and version:

Steps to Reproduce:
1. Use latest cocosstudio to edit a demo UI with Label
2. publish to .csb and load it, then the postion not correct because the contetSize of Label not be calculated correct.
3. test cocosstudio project: http://halx99.oschina.io/releases/LabelBugTest.zip

CocosStudio Effect:
![cocosstudioeffect](https://cloud.githubusercontent.com/assets/6977319/17798009/4a9498e4-6600-11e6-8b84-fb51b9055166.png)

The latest Label render effect:
![gameeffect](https://cloud.githubusercontent.com/assets/6977319/17798017/5ee0e51e-6600-11e6-99aa-278b4b4ac631.png)

halx99:
@minggo please take a look, does this is a bug?

minggo:
@halx99 sorry, cocos studio is canceled, we don't have resource to support it.

halx99:
@minggo Well, I think it's not cocosstudio problem, you also can reproduce this issue draw the border of the label, the render text will surpass right edge.

minggo:
@halx99 could you please provide a test case? Thanks.

halx99:
ok, I will.

halx99:
@minggo The test case: https://github.com/halx99/cocos2d-x/tree/labelsizebugtest

My test result:
![a _mrg6co5oa tkv 2n89 w](https://cloud.githubusercontent.com/assets/6977319/17798940/b8586f06-6608-11e6-940b-a6b435bd1589.png)

minggo:
@halx99 i can not find the error from your pasted image. What does the red rectangle mean?

halx99:
@minggo The red rectangle is border of the label, normally all letters shoud at inner of the rectangle, you can see the last letter is surpassed a little.

minggo:
Yep.

dumganhar:
@halx99, I tested your demo and I could reproduced it.

@newnon has submitted a Pull Request https://github.com/cocos2d/cocos2d-x/pull/16689

After I applied @newnon's patch, this issue was fixed.

@halx99, could you try to apply that patch and test again? Thanks.

[...]

PR #16401, opened 2016-08-19
#FIX Label size calculation bug.

halx99:
see issue: #16399

### DIFF
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -218,7 +218,7 @@ bool Label::multilineTextWrap(const std::function<int(const std::u16string&, int
                     nextLetterX += _horizontalKernings[letterIndex + 1];
                 nextLetterX += letterDef.xAdvance * _bmfontScale + _additionalKerning;
 
-                tokenRight = nextLetterX / contentScaleFactor;
+                tokenRight = (letterPosition.x + letterDef.width) / contentScaleFactor; // #FIX Label size calculation bug.
             }
             nextChangeSize = true;
 
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -218,7 +218,7 @@ bool Label::multilineTextWrap(const std::function<int(const std::u16string&, int
                     nextLetterX += _horizontalKernings[letterIndex + 1];
                 nextLetterX += letterDef.xAdvance * _bmfontScale + _additionalKerning;
 
-                tokenRight = (letterPosition.x + letterDef.width) / contentScaleFactor; // #FIX Label size calculation bug.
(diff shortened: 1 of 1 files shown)


minggo:
The codes is changed by @cezheng in #16172. @cezheng could you please take a look?

cezheng:
I don't see why this is a bug after looking at the issue. FYI,  #16172

cezheng:
Also, I think this code is wrong if contentScaleFactor or _bmFontScale does not equal to 1.0

minggo:
@cezheng you can refer to #16399 for detail information.

cezheng:
I guess my PR is moving letters from right to left as you can see in the screenshots in #16172. However the screenshots in #16399 the letters seems to be overlapping with the right of the boundingbox.

cezheng:
I can't say much yet. My advice:
1. Also test with bitmap fonts
2. Also test with different contentScaleFactor
3. Also test with different _bmfontScale
4. Also test with different Alignment configurations

My fix might not be right in all situations, but neither is this PR at the moment I guess.

halx99:
How about:

```
tokenRight = letterPosition.x + letterDef.width * _bmfontScale / contentScaleFactor; // #FIX Label size calculation bug.
```

minggo:
@halx99 @cezheng Yep, i think we should add test case for it first.

halx99:
Yes

minggo:
@halx99 will you add the test case?

halx99:
I have not add test case, but I test in our game, it's work ok:
test with bitmap fonts(work ok)
test with different contentScaleFactor(work ok)
test with different _bmfontScale(the engine implement always 1.0)
test with different Alignment configurations(work ok)
arp2nsmm505hn wk4 vs bp

minggo:
@halx99 can you add these test cases in cpp-tests?

halx99:
And where shoud I add to? bug-test?

minggo:
I think it can be added in `UITest`, now the test case is nothing about cocos studio.

halx99:
OK, I will add later.

minggo:
thanks

cezheng:
This just reverts #16172 and it will cause problems for right alignment for bitmap font.

cezheng:
Let me explain a little bit about why I made the changes in #16172 

Characters that have different `letterDef.width`, for example 0 and 1, the calculated `letterRight` for 1 will be smaller than 0. If we use this value for calculating line width, if the last character for different lines will have problems when alignment is set to right.

#16172 change it to use the starting position of the next character to calculate the line width, which makes right alignment look normal. Because this adds xAdvance & additionalKerning to it, which make all characters the same width. The original formula ends with calculating the last character of each line a little bit narrower.

halx99:
Actually the new Label have anothor incorrect behavior:

vb441wk122 nnsxa 0cw

halx99:
so still need test

PR #16402, opened 2016-08-19
Remove unnecessary ActionInterval casts

mogemimi:
This PR removes several unnecessary casts (from Sequence to ActionInterval) and changes some of the type declarations to `auto` keywords. Thanks!


minggo:
thanks

PR #16405, opened 2016-08-19
Fixed Camera::_viewProjectionUpdated state

### LINKED ISSUE #16404: Camera::_viewProjectionUpdated not initialised
cocos2d-x version: 3.12

Camera::_viewProjectionUpdated is not initialised in constructor. Should be initialised to false.
This can lead to redundant visibility checks if _viewProjectionUpdated happens to resolve true.

Furthermore as Camera::visit() is never called, _viewProjectionUpdated is never actually updated, resulting in no visibility tests when the camera transform changes.

I propose we add 
`_viewProjectionUpdated = _transformUpdated;`
to Camera::apply() as well.

Almax27:
Camera::_viewProjectionUpdated is not initialised in constructor. Should be initialised to false.
This can lead to redundant visibility checks if _viewProjectionUpdated happens to resolve true.

As Camera::visit() is never called, _viewProjectionUpdated is never actually updated, resulting in no visibility tests when the camera transform changes.

Fixes #16404

### DIFF
--- a/cocos/2d/CCCamera.cpp
+++ b/cocos/2d/CCCamera.cpp
@@ -111,6 +111,7 @@ Camera::Camera()
 , _viewProjectionDirty(true)
 , _cameraFlag(1)
 , _frustumDirty(true)
+, _viewProjectionUpdated(false)
 , _depth(-1)
 , _fbo(nullptr)
 {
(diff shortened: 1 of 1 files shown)


ricardoquesada:
[ci rebuild]

PR #16409, opened 2016-08-20
#Improve RichText::createWithXml performance.

halx99:
# Improve RichText::createWithXml performance with real XML SAX parser.


halx99:
[ci rebuild]

halx99:
- And what's about replace tinyxml2 with rapidxml?

halx99:
@minggo Take a look about this PR?

minggo:
> And what's about replace tinyxml2 with rapidxml?

It is good for speed, but it may break compatibility.

@ricardoquesada could you please take a look of this PR? I am not familiar with these codes. Thanks.

halx99:
As we known, the official rapidxml only support DOM parse, but it easy to add pure SAX parse(without any additional memory alloc) support based on rapidxml, and this PR already complete it.
For more detail and test project see: https://github.com/halx99/xmldrv3

ricardoquesada:
@minggo I haven't review the whole code, but yeah, SAX should be faster and/or consume less memory. at least in theory.

@halx99 could you add a performance test with and without your patch? thanks.

halx99:
see the new PR: https://github.com/cocos2d/cocos2d-x/pull/16419

Issue #16355, opened 2016-08-11
Replace fmod on Linux with SDL Audio or something similar


slackmoehrle:
Users have stated that we are not truly open-source on Linux because we require `fmod`. This isn't free and they have no other options. During our weekly meeting @ricardoquesada suggested that we could offer SDL Audio as an alternative and let users turn on `fmod` if they choose to. @linshun agreed. The timing of this change is TBD.

@slackmoehrle is happy to test when ready.

ricardoquesada:
yes, other alternatives are:
- use ALSA driver directly
- or use [PulseAudio](https://en.wikipedia.org/wiki/PulseAudio)

minggo:
Can refer to #16410. The PR use OpenAL for linux.

JOOpdenhoevel:
I don't really know how powerful/fast/etc. those alternatives are, but from my point of view, OpenAL has a big advantage: It's a cross-platform library. The implementation for Windows only uses OpenAL, MPEG, and Vorbis, which are available for Windows as well as for Linux. Due to this fact, I could simply copy the source files and it worked. All tests were positive, the C++ test and the Lua test. Another option would be to merge those implementations into one which would make maintaining the engine way easier. I could do that for Linux and Windows if you like this idea.

minggo:
@Janonard you are right. @dumganhar i think you can take a look too. I think it is better that we can use OpenAL as possible, then it is easy to main the codes.

minggo:
> Another option would be to merge those implementations into one which would make maintaining the engine way easier. I could do that for Linux and Windows if you like this idea.

Yep, it is better.

dumganhar:
Agree this.

We only need OpenSLES for Android & OpenAL for other platforms. Easily maintaining our engine is really important. 

BTW, we also need to improve New Audio Engine since I think the logic isn't very clear for developer to understand, we need to do more encapsulation, abstraction and documentation.

chuckhacker:
OpenAL-soft has a OpenSLES driver built-in:

https://github.com/kcat/openal-soft/blob/2f1f7f4c6aaf54f9361865d508d82cdf027368bf/Alc/backends/opensl.c

cirosantilli:
In particular, this blocks a Debian package from being created.

And creates problems that we can't solve properly because no source, e.g.:

- porting fmod: https://github.com/cocos2d/cocos2d-x/issues/18001
[...]

Issue #15419, opened 2016-04-13
AudioEngine to OpenAL sync issue on Win32


rh101:
In AudioEngine-win32.cpp, the AudioEngineImpl::update is scheduled to be called every 0.05 seconds.  This is done in AudioEngineImpl::play2d, via the following call:

```
    scheduler->schedule(schedule_selector(AudioEngineImpl::update), this, 0.05f, false);
```

The AudioEngineImpl layer uses its own audio state, which becomes out of sync with the true state of the audio source being played via the OpenAL layer.

This causes a problem with any method that is called with an the audio ID provided by the play2d method. For example, calling AudioEngine::getCurrentTime(audioID) should return the current position of the audio source being played.  This works fine until OpenAL gets to the end of the audio source, where the OpenAL state of the audio becomes AL_STOPPED, yet the AudioEngine state for that specific audio file is still set to AudioState::PLAYING, since the AudioEngineImpl::update method has not yet been called to update this state (given it is only called every 0.05 seconds).

This results in an invalid value of 0.0 being returned by AudioEngine::getCurrentTime(audioID), since the AudioEngineImpl::update code has not yet been called to update the true status of this audioID.

There are several ways to fix this, and a few that I can think of at the moment are as follows:

(1) Set the AudioEngineImpl::update function to be called called every frame instead of every 0.05 seconds.  

(2) If updating every frame causes performance issues, then modify each method that performs actions against a specific audioID in order to check for the correct OpenAL state of that audio source.  This is not the preferred way to handle the issue, since the AudioEngine state and the OpenAL state will still be out of sync, which may cause other issues when using the AudioEngine. An example of this modification would be as follows:

```
float AudioEngineImpl::getCurrentTime(int audioID)
{
    float ret = 0.0f;
    auto& player = _audioPlayers[audioID];
    if(player._ready){
        if (player._streamingSource) {
            ret = player.getTime();
        } else {
            ALint sourceState;
            alGetSourcei(player._alSource, AL_SOURCE_STATE, &sourceState);
            // Check the OpenAL state of the source, and if it has finished, return the correct value
            if (sourceState == AL_STOPPED)
            {
                ret = 1.0f;
            }
            else
            {
                alGetSourcef(player._alSource, AL_SEC_OFFSET, &ret);
            }

            auto error = alGetError();
            if (error != AL_NO_ERROR) {
                log("%s, audio id:%d,error code:%x", __FUNCTION__,audioID,error);
            }
        }
    }

[...]

PR #16410, opened 2016-08-21
Copying the audio engine implementation for Windows to Linux

JOOpdenhoevel:
The title says what this is about: I have copied the audio engine for Windows to Linux and made it work. I did this because the Windows implementation uses OpenAL instead of FMOD and
it was quite strange to me that Linux uses a proprietary audio library, while every other platform uses the open one. Also, I think that the audio engine is easier to maintain when every platform uses the same API.

I was surprised that this was such an easy job because I only needed to copy the source files, change every "Win32" to "Linux" and tell CMake to link against these libraries. I wonder why no one did this before. Anyway, why was the audio engine implemented with FMOD and not with OpenAL?

I also updated the installation script, but the deps-archive still contains FMOD. It would be really nice if someone could remove it when this get merged.

EDIT: The test failed since the dependencies for linux have changed.


minggo:
[ci rebuild]

ricardoquesada:
[ci rebuild]

ricardoquesada:
I'll take a look at this one.

ricardoquesada:
@minggo 
the PR is failing because I can't find some libraries.
what script should I edit to tell the builder to install the missing libraries?

in particular:

```
-- Vorbis include dirs: 
CMake Error at cmake/Modules/FindPackageHandleStandardArgs.cmake:136 (message):
  Could NOT find MPG123 (missing: MPG123_INCLUDE_DIR MPG123_LIBRARY)
```

minggo:
@ricardoquesada i think the codes `cocos_find_package(MPG123 MPG123 REQUIRED)` will find it. Did you invoke `build/install-linux-deps.sh` to install MPG?

ricardoquesada:
@minggo I mean, I think the builder is failing because the builder doesn't have installed those packages. So, I have to tell the builder to install the packages.

minggo:
@zilongshanren will the builder run the `install-linux-deps` for every PR?

minggo:
@ricardoquesada i asked @zilongshanren help to manually installed it.

zilongshanren:
[ci rebuild]

ricardoquesada:
@minggo thanks!

zilongshanren:
[ci rebuild]

ricardoquesada:
I've just tested this PR on my Linux machine and everything seems to be working ok except that in the "7 - Audio CocosDenshion", it fails with the "Control effects" with this error:

`Unsupported media type file: effect1.wav`

ricardoquesada:
[ci rebuild]

JOOpdenhoevel:
@ricardoquesada Themen Cocos Denshion Test uses the simple Audio engine which uses a standart Windows library which propably supports .wav. The simple a.e. for Linux uses the proper audio engine, which is why it doesn't support that. I'm going to fix that.

ricardoquesada:
@Janonard great. thanks!

JOOpdenhoevel:
I guess I need some help: I would like to use FFmpeg's avcodec, since it supports almost everything. How do I tell cmake to link against it? I'm quite new to the project and I'm not yet used to cocos' cmake. May I get some help?

JOOpdenhoevel:
As mentioned above, I need to add another library to support .wav files. When I think about it, it makes more sense to get the merging done and add the library afterward. It appears to me that adding a new dependency is quite a big thing for Cocos and I assume that it is better to discuss which library to use and which files to support later. I will do that, but I guess it's better to fulfill the cause of this pull request and add other file formats then. Is that okay for you?

JOOpdenhoevel:
I merged the audio engine for Windows and for Linux into the "Desktop" implementation. The tests for the main audio engine are positive, but the ones for the simple audio engine are broken because the S.A.E. is a wrapper for the new A.E. and it doesn't support the file formats  used in the test. It was like this on Linux all the time and the new engine does everything the old does while beeing easier to extend. Because of this I suggest do abandon the old engine. Do you subscribe my point of view?

ricardoquesada:
@Janonard 
Yes, SimpleAudioEngine will be deprecated in v4.0. But we can't deprecate it right now, not in a Point release.
Regarding the wav file, we should add support for it before merging this PR since it was working before. Adding "FFmpeg" dependency is not a problem in linux, assuming that it is a package that users can install using "apt-get" or similar.

ricardoquesada:
@Janonard should we also include Mac in the "desktop" thing? I mean, use this same for Mac? (what are we using on Mac BTW?)

JOOpdenhoevel:
@ricardoquesada

> Yes, SimpleAudioEngine will be deprecated in v4.0.

Okay, good to know.

> Regarding the wav file, we should add support for it.

OK, will do that, but this will take some more time since I need to familiarize myself in the CMake setup and in FFmpeg.

> Should we also include Mac in the "desktop" thing?

From my point of view, Mac should be included, but I can't do this since I don't own a Mac or have the money to buy one. Someone else would need to do that.

ricardoquesada:
@Janonard 

> will do that, but this will take some more time since I need to familiarize myself in the 
> CMake setup and in FFmpeg.

if you just make it work (the fixes in the code), I can fix the cmake stuff and needed dependencies.

> From my point of view, Mac should be included, but I can't do this since I don't own a Mac or have the money to buy one. Someone else would need to do that.

Ok. I can fix that I guess.

JOOpdenhoevel:
@ricardoquesada 
I'm having a problem: I trying to test FFmpeg in a new CMake project, but linking to FFmpeg simply doesn't work, the linker complains that some symbols haven't been resolved, although I linked against FFmpeg. I tried all possible options for CMake, but it even doesn't work from the command line. Can you figure that out, please? It would be nice if you would add FFmpeg to cocos and send me a pull request with it. :)

ricardoquesada:
@Janonard sure. I can figure out how to link it. Have you added the ffmpeg code? It doesn't matter if doesn't link, but in order to fix the linking issue I need to try with code that uses it.

JOOpdenhoevel:
@ricardoquesada 
The problem is because I couldn't link against FFmpeg in any way, I couldn't experiment with and therefore I couldn't learn how to use it. I have added a small and silly test that uses libavformat, libavcodec and libavutil, which are the libraries I'm probably going to use, which prints the gcd of 4 and 8, the version of avformat and the version of avcodec into the log. I'm going to remove it when you're done, but you can do that too if you like.

ricardoquesada:
@Janonard ok. got it. I'll add ffmpeg first.

cocos/audio/CMakeLists.txt
```diff
@@ -3,14 +3,12 @@ set(COCOS_AUDIO_SRC
     audio/AudioEngine.cpp
     )
 
-if(WINDOWS)
+if(LINUX OR WINDOWS)
     set(COCOS_AUDIO_PLATFORM_SRC
-        audio/win32/SimpleAudioEngine.cpp
-        audio/win32/MciPlayer.cpp
-        audio/win32/MciPlayer.h
-        audio/win32/AudioEngine-win32.cpp
```
ricardoquesada, line 11: it seems that `MciPlayer.cpp` was removed from windows. was this done on purpose? or is it just a copy&paste error?

JOOpdenhoevel:
Yes, this is intended. MciPlayer used a standart Windows library and therefore I couldn't keep it. The simple audio engine for Windows was a standalone implementation while the SAE for Linux wrapped the new one. I decided to go with the Linux way.

minggo:
@ricardoquesada any update about this PR? Can i merge it when it passes CI and fix conflict?

ricardoquesada:
@minggo there is one bug that needs to be fixed. but in order to do that I told the @Janonard that I was going to add CMake support for FFMpeg, but I didn't have the time to do it yet.

JOOpdenhoevel:
I am very sorry, but I am not interested in this topic anymore. While I have waited for the FFmpeg integration, I began to work on a new project, which is near public release right now. Since I have to focus on this, I have no more time for cocos2d-x, maybe I will return to it later.

v1993:
@drelaptop @minggo Can you finish and merge this pull request, please?

drelaptop:
@v1993  it's a same issue with https://github.com/cocos2d/cocos2d-x/issues/16355 https://github.com/cocos2d/cocos2d-x/issues/15419, I think we should finish it at 3.18 milestone, what do you think? @minggo

minggo:
Yep, i think we can do it in 3.18 because FMOD always cause issues on linux.

PR #16417, opened 2016-08-22
updated VR, per Minggo

slackmoehrle:
update VR section, per Minggo.


minggo:
thanks @slackmoehrle

minggo:
@slackmoehrle  i can't find any information about VR plugin usage. Will it be added in http://www.cocos2d-x.org/docs/programmers-guide/vr/index.html?

PR #16419, opened 2016-08-23
#Improve XML parse performance with rapidxml pure SAX parser

halx99:
This PR is will never break any compatible, and more easy to review.


halx99:
I will add performance test case later.

minggo:
Why put `rapidxml_sax3.hpp` in `platform`? Is it platform dependent?

halx99:
It's just for CCSAXParser, and CCSAXParser is platform independent.
Where put `rapidxml_sax3.hpp` is better?

halx99:
[ci rebuild]

minggo:
@halx99 i think it is better to move it to `base` folder. And i don't know why `CCSAXParser.cpp` in `platform` folder.

halx99:
@minggo Yes, the new `rapidxml_sax3.hpp` could move to base, but move the `CCSAXParser` will break compatible.

halx99:
standalone test result:
vb j025i1he 11vydr t

test project: https://github.com/halx99/xmldrv3

minggo:
@halx99 yep, just move `rapidxml_sax3.hpp`.

zilongshanren:
@minggo   There is already a `rapidxml.hpp` in external directory?

zilongshanren:
@minggo  My mistake, it's rapidjson..
But  I think rapidxml is also a 3rd party dependencies, it should be in external directory.

halx99:
So, move to external?

halx99:
[ci rebuild]

halx99:
The ios ci seems dropped:
`ERROR: Connection was broken: java.io.IOException: Unexpected termination of the channel`

minggo:
@zilongshanren Yep,  in theory it should be in external folder. But as it is just a header file, so may be should can put it in cocos2d-x repo. @ricardoquesada what's your opinion?

halx99:
[ci rebuild]

ricardoquesada:
@minggo 
As you said, in theory it should be in the "external" folder.
But this is only a header file. But if we decide to put it in the cocos/ directory structure, then, where?

Probably it is better to put it in the external folder, even if it is just one file.

halx99:
Yes, I agree, so I create a PR to `cocos2d-x-3rd-party-libs-bin` https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/249

minggo:
guys, you are right. We should put it in external. Sorry about it.

halx99:
Ok，I will changed the code after the PR: cocos2d/cocos2d-x-3rd-party-libs-bin#249 merged.

ricardoquesada:
@halx99 thanks. PR in 3rd-party-libs-bin merged.

halx99:
ok, I just changed the source code.

halx99:
@ricardoquesada It seems needs a new tag `v3-deps-111` of `cocos2d/cocos2d-x-3rd-party-libs-bin`

ricardoquesada:
@halx99 done: https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/releases/tag/v3-deps-111

halx99:
ok, but it the release notes seems wrong: 
rlvketpelk_j 2i i5n h

It should be `Includes rapidxml SAX parser` 😀

halx99:
[ci rebuild]

halx99:
[ci rebuild]

ricardoquesada:
@halx99 fixed. thanks :)

ricardoquesada:
looks good to me. merged.

PR #16423, opened 2016-08-23
Add missing class declaration and remove duplicate declaration

mogemimi:
This PR adds a missing type declaration and header include for class `Quaternion`, it also removes a duplicate declaration of `Data` in `CCProperties.h`. Thanks!

### DIFF
--- a/cocos/base/CCProperties.cpp
+++ b/cocos/base/CCProperties.cpp
@@ -30,6 +30,7 @@
 #include "math/Vec3.h"
 #include "math/Vec4.h"
 #include "math/Mat4.h"
+#include "math/Quaternion.h"
 #include "base/ccUTF8.h"
 #include "base/CCData.h"
 
--- a/cocos/base/CCProperties.h
+++ b/cocos/base/CCProperties.h
@@ -41,7 +41,7 @@ class Vec2;
 class Vec3;
 class Vec4;
 class Mat4;
-class Data;
+class Quaternion;
 class Data;
 
 


minggo:
But why it doesn't have compiling error? Does it mean we can remove some `#include`?

mogemimi:
@minggo  Thank you for quick reply.

> But why it doesn't have compiling error?

The following ascii art represents the dependency graph:

```
"math/Vec2.h"
"math/Vec3.h"
"math/Vec4.h"
"math/Mat4.h"
"math/Quaternion.h"
      ^
      +
"math/CCMath.h"
      ^     ^
      |     +--------------+
      +                    |
"math/CCGeometry.h"        |
      ^     ^              |
      |     +--------+     |
      |              +     +
      |            "base/ccTypes.h"
      |              ^     ^
      +              +     |
"renderer/CCTexture2D.h"   |
      ^                    |
      +--------------------+
      +
"base/CCProperties.h"
```

`CCProperties.h` indirectly depends on `math/Quaternion.h` as shown in the graph.

> Does it mean we can remove some #include?

Probably yes. I guess we can remove `#include "renderer/CCTexture2D.h"` and `#include "base/ccTypes.h"` from `CCProperties.h` in favour of adding them to `CCProperties.cpp` instead. So I tried to apply these changes and it worked fine with Xcode, but I don't know if these can cause some side effects. 

Btw, if we are going to improve the compile time and any redundant dependencies, we can use a automated static analysis tool such as [include-what-you-use (IWYU)](http://include-what-you-use.org/), which will help us to solve the `#include` and type declaration problems.

minggo:
@mogemimi good to know, thanks

minggo:
And yes, we need to improve compile time if possible.

mogemimi:
@minggo Sounds good! I'll try to improve the build performance using IWYU, thanks.

minggo:
thanks @mogemimi

PR #16424, opened 2016-08-23
possible memory leak when font size is 0

cinakyn:
When outWidth or outHeight is 0, new bitmap is allocated with 0 size.
And this bitmap will never be deallocated because FontAtlas::prepareLetterDefinitions just skip character which has 0 size.


ricardoquesada:
this PR looks good to me. @minggo should I merge it?

minggo:
@ricardoquesada sure

PR #16451, opened 2016-08-27
Fix memory leak caused by packetOut in SocketIO.cpp

Rubonnek:
Static code analyzer cppcheck showed the following error on SocketIO.cpp:

```
Checking SocketIO.cpp...
[SocketIO.cpp:964]: (error) Memory leak: packetOut
```

I manually inspected the code and as far as I can tell it is, indeed, a memory leak.


dumganhar:
Good catch. 👍

ricardoquesada:
I'm not familiar with this code, but yes, it seems that there is a leak there and the PR seems to be Ok.

@minggo should I merge it?

minggo:
Yep, it seems `packetOut` is not freed, but the logic is to send an event.

minggo:
@ricardoquesada yep, it seems the codes is needless. Feel free to merge it. Thanks.

PR #16482, opened 2016-08-31
Fix MenuItemSprite::unselected() on disabled items

pyrosphere:
Calling MenuItemSprite::unselected() on a disabled item makes it show _normalImage instead of _disabledImage.


minggo:
thanks

PR #16487, opened 2016-09-01
Fix memory leak when loading Cocos Studio file

Rubonnek:
Another leak found by cppcheck.

In [line 64](https://github.com/cocos2d/cocos2d-x/blob/v3/tools/simulator/libsimulator/lib/runtime/RuntimeCCSImpl.cpp#L64) of RuntimeCCSImpl.cpp a new tinyxml2::XMLDocument is allocated when loading a Cocos Studio file, but the allocated memory is never deleted.


minggo:
thanks

Issue #16471, opened 2016-08-30
Text color lose in  cocos2d-x 3.13


Dauch:
- cocos2d-x version: 3.13
- devices test on: win32
- developing environments
  - NDK version:
  - Xcode version:
  - VS version: 2015
  - brower type and version:
- cocostudio 3.10

Steps to Reproduce:

open csb, is Text type's color is lose, it won't set
origin:
![image](https://cloud.githubusercontent.com/assets/5201160/18089332/07eac9c8-6ef2-11e6-8f64-8c3464093945.png)
output:
![image](https://cloud.githubusercontent.com/assets/5201160/18089395/5f7f7882-6ef2-11e6-921c-e20bde7dee72.png)

i didn't do anything in code change it's color, it should be expected like in cocostudio(is ok with cocos2d-x 3.12)

zhaozhantao:
Me too.

Dauch:
pls give some help @minggo

minggo:
sorry, cocostudio is canceled, we don't have resource to handle corresponding issues.

Dauch:
@minggo  any suggestion for this bugifx?

minggo:
I am sorry, i are not familiar with cocostudio and we don't make sure v3.13 can work with cocos studio.

zhaozhantao:
this is min code:

```
local node = cc.Node:create()
self:addChild(node, 100)
-- 此处设置为false则testLabel:setTextColor有效
-- 设置为true则testLabel:setTextColor无效
-- if set false then  testLabel:setTextColor is useful
node:setCascadeColorEnabled(true) 
local testLabel = ccui.Text:create()
testLabel:setFontSize(72)
testLabel:setString("test setCascadeColorEnabled")
testLabel:setTextColor({r = 0, g = 0, b = 0})
node:addChild(testLabel)
```

Dauch:
In CCLabel.cpp line 1893, I comment line 1893&1894 it work fine. changging those has any bad influence?
@minggo
`
void Label::updateDisplayedColor(const Color3B& parentColor)
{
    Node::updateDisplayedColor(parentColor);

```
//if (_currentLabelType == LabelType::TTF || _currentLabelType == LabelType::STRING_TEXTURE)
//    setTextColor(Color4B(_displayedColor));
```

`

zhaozhantao:
@Dauch I delete the two line code. It work fine.
[...]

Issue #15214, opened 2016-03-09
The color of the underline is different from the text color.


m-yukio:
If use TrueType font, the color of the underline is different from the text color.

``` c++
Label* label = Label::createWithTTF("CHECK!", "ipaexm.ttf", 48.0f);
label->enableUnderline();
label->setColor(cocos2d::Color3B::BLUE);
label->setPosition(200.0, 400.0);
this->addChild(label);
label = Label::createWithSystemFont("CHECK!", "Verdana", 48.0f);
label->enableUnderline();
label->setColor(cocos2d::Color3B::BLUE);
label->setPosition(200.0, 300.0);
this->addChild(label);
```

![uilabel](https://cloud.githubusercontent.com/assets/9098173/13622886/8d41cf72-e5e5-11e5-87ba-ac3f82c4a790.jpg)

ricardoquesada:
either wait until v3.13 is released, or use `setTextColor()` instead. thanks.

ricardoquesada:
re-fixed in v3.14.

`setColor()` changes the color of the text and underline.
`setTextColor()` changes the color of only the text. This is the same behavior as system fonts.

PR #16489, opened 2016-09-01
Issue 16471

ricardoquesada:
fix: Label updateColor() and underline color work as expected
fixes Github issue #15214 correctly. underline uses _displayColor
and not _textColorF. _textColorF is only for the text.
This also emulates the SystemFont underline behavior

fixes Github issue #16471


ricardoquesada:
[ci rebuild]

PR #16503, opened 2016-09-02
fixed #16502: SPINE color bug (temporary storage of color was wrong in SkeletonRenderer).

### LINKED ISSUE #16502: v3.13 -> SPINE color bug.
SPINE will display objects as black, when color is at all modified in the editor. Comparing color code with previous SPINE versions fixed it. Since I am not sure if this is a correct fix, I will not make a PR.

screen shot 2016-09-02 at 08 14 11

dumganhar:


minggo:
@dumganhar what's the version of the spine supported?

dumganhar:
@minggo , it's synchronized from the master branch of https://github.com/EsotericSoftware/spine-runtimes. Except that I didn't include the new feature of SkeletonBinary (https://github.com/EsotericSoftware/spine-runtimes/blob/master/spine-c/src/spine/SkeletonBinary.c
) since it was added recently and may not be stable enough.
If you think that we need to include `SkeletonBinrary`, just let me know.

dumganhar:
@minggo , this PR only contains `color` bug fix now.

dumganhar:
This patch is from https://github.com/EsotericSoftware/spine-runtimes/commit/c8b72e382f10d555187ef74e38306893972bd489

PR #16516, opened 2016-09-05
3.13.1 text color

minggo:


ricardoquesada:
@minggo looks good to me

PR #16522, opened 2016-09-05
misspelling check on cocos directory

pineoc:
Misspelling fix on some comments,
cocos/editor-support/cocostudio/CSParseBinary_generated.h
fix misspelling postion -> position


minggo:
thanks

Issue #16527, opened 2016-09-06
can't compile libcocos2dx-with-controller with "--android-studio"


mars3142:
- cocos2d-x version: 3.13
- developing environments
  - NDK version: r12b (NDK in Android SDK)

Steps to Reproduce:
1. create new project
2. edit <project>/proj.android-studio/settings.gradle and change the path to libcocos2dx-with-controller
3. call _compile compile -p android --android-studio_
   --> error: package com.android.vending.expansion.zipfile does not exist 

PS: A change to the proj.android works fine.

mars3142:
https://github.com/cocos2d/cocos2d-x/pull/16529

PR #16529, opened 2016-09-06
fixed compile error for android-studio build (with-controller)

mars3142:
fixed issue https://github.com/cocos2d/cocos2d-x/issues/16527


minggo:
thanks

PR #16530, opened 2016-09-06
Support for compiling more OpenGL ES shaders on desktop platforms also.

sachingarg05:
Support compiling OpenGL ES shaders that use lowp, mediump or highp on desktop platforms also. These symbols are not defined on desktop platforms so the code defines these symbols.

### DIFF
--- a/cocos/renderer/CCGLProgram.cpp
+++ b/cocos/renderer/CCGLProgram.cpp
@@ -461,6 +461,8 @@ bool GLProgram::compileShader(GLuint* shader, GLenum type, const GLchar* source,
         (type == GL_VERTEX_SHADER ? "precision mediump float;\n precision mediump int;\n" : "precision mediump float;\n precision mediump int;\n"),
 #elif (CC_TARGET_PLATFORM != CC_PLATFORM_WIN32 && CC_TARGET_PLATFORM != CC_PLATFORM_LINUX && CC_TARGET_PLATFORM != CC_PLATFORM_MAC)
         (type == GL_VERTEX_SHADER ? "precision highp float;\n precision highp int;\n" : "precision mediump float;\n precision mediump int;\n"),
+#else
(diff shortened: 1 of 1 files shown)


sachingarg05:
These changes build fine for me on Mac. It looks like CI build failed building for mac due to a network issue. How do I check more details about why it failed?

minggo:
@sachingarg05 how to reproduce the issue? And it just defined the symbols, but what's the value of this symbols?

minggo:
[ci rebuild]

sachingarg05:
OpenGL ES allows specifying precision for individual variables, but desktop OpenGL does not support this. 

For example:
  varying highp vec2 xy;

This code will compile on mobile platforms but fail on desktop platforms. By adding empty definitions of lowp, mediump, highp on desktop platforms we can use same shaders there. The precision identifier just gets ignored on desktops as it wouldn't mean anything there.

minggo:
It sounds reasonable, which means we can remove 

```
#ifdef GL_ES
```

in shaders. @ricardoquesada what's your opinion?

ricardoquesada:
what I like:
we stop using the `#ifdef`

what I don't like:
is that it is not obvious that `lowp` (and friends) has not effect on desktop GL.

I'm good if we add a comment on the shaders saying that `lowp...` has not effect on desktop.

minggo:
> I'm good if we add a comment on the shaders saying that lowp... has not effect on desktop.

If we should add comment for each shader, then it is not convenient compared to add `#ifdef`. So i have no idea about this PR, it seems it haven't bring much benefit.

ricardoquesada:
it is just replacing the `#ifdef` with  `// lowp... has no effect on Desktops`

sachingarg05:
OK, I will add this check for #ifdef GL_ES to my shaders.

I am closing this pull request.

PR #16532, opened 2016-09-06
getSlidBallRenderer methods to allow more customization of UISlider

sachingarg05:
getSlidBallRenderer methods to allow more customization of UISlider look and feel.

```
Sprite* getSlidBallNormalRenderer();
Sprite* getSlidBallPressedRenderer();
Sprite* getSlidBallDisabledRenderer();
Node* getSlidBallRenderer();
```

We also need jsb bindings for these but I think those are to be generated automatically by scripts, so I am not including them in the pull request. I can add them if needed.

### DIFF
--- a/cocos/ui/UISlider.h
+++ b/cocos/ui/UISlider.h
@@ -265,6 +265,11 @@ public:
      */
     float getZoomScale()const;
 
+    Sprite* getSlidBallNormalRenderer();
+    Sprite* getSlidBallPressedRenderer();
+    Sprite* getSlidBallDisabledRenderer();
+    Node* getSlidBallRenderer();
+
     ResourceData getBackFile();
     ResourceData getProgressBarFile();
     ResourceData getBallNormalFile();
--- a/cocos/ui/UISlider.h
+++ b/cocos/ui/UISlider.h
@@ -265,10 +265,10 @@ public:
      */
     float getZoomScale()const;
 
-    Sprite* getSlidBallNormalRenderer();
-    Sprite* getSlidBallPressedRenderer();
-    Sprite* getSlidBallDisabledRenderer();
-    Node* getSlidBallRenderer();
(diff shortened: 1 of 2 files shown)


sachingarg05:
I have pushed another pull request for same API changes with the HTML5 web engine. I think both these pull requests should be merged together to maintain compatibility between HTML5 and JSB APIs.

https://github.com/cocos2d/cocos2d-html5/pull/3371

sachingarg05:
For example, we need these functions to be able to change scale of the sliderball to adjust the sliderball size independent of image texture used. And to apply a color tint to sliderball based on how far the sliderball has been moved.

cocos/ui/UISlider.h
```diff
@@ -265,6 +265,11 @@ class CC_GUI_DLL Slider : public Widget
      */
     float getZoomScale()const;
 
+    Sprite* getSlidBallNormalRenderer();
+    Sprite* getSlidBallPressedRenderer();
+    Sprite* getSlidBallDisabledRenderer();
+    Node* getSlidBallRenderer();
```
minggo: I think it is better to add const like

``` c++
Sprite* getSlidBallNormalRenderer() const;
```

And why `getSlidBallRenderer ()` return `Node*`, other functions return `Sprite*`?

sachingarg05: I have updated the pull request to declare these functions as const.

The SlidBallNormalRenderer, SlidBallPressedRenderer and SlidBallDisabledRenderer are declared as Sprites but SlidBallRenderer is declared as a Node (it is only used as a parent for the other Sprites).

minggo:
it looks good to me, @zilongshanren please take a look

zilongshanren:
@minggo  yeah, it looks good.

minggo:
Thanks @sachingarg05

PR #16542, opened 2016-09-07
v3.13.1 updated winrt libcocos2d dll version to 3.13.1

stammen:
This pull request updates the dll version of the Windows 10 UWP and Windows 8.1 libcocos2d dlls to v3.13.1 as required by the Windows Store version tracking.


minggo:
thanks.

Issue #16512, opened 2016-09-03
Hexagonal map fails on TMXTiledMap::getTileAt


wduminy:
- cocos2d-x version: 3.13
- devices test on: Windows 8.1 (win32)
- developing environments
  - NDK version: n/a
  - Xcode version: n/a
  - VS version: 2015 Community
  - brower type and version: n/1

I really like the idea that _tiles are sprites :)_ -- but I could not get it to work. The problem I found can be reproduced with the code snippet below.  You may need the provided map to reproduce the problem.  It can be downloaded from drop-box: [here](https://www.dropbox.com/s/vixpjo4ulm6reez/hexamap.zip?dl=0). 

```
        auto map = TMXTiledMap::create("Map.tmx");
        auto floor = map->getLayer("Floor");
        for (auto x = 0; x < map->getMapSize().width; x++)
            for (auto y = 0; y < map->getMapSize().height; y++) {
                Vec2 p(x, y);
                auto t = floor->getTileGIDAt(p);
                cocos2d::log("tile at (%d,%d) = %d", x, y, t);
                auto sprite = floor->getTileAt(p);
            }
```

The snippet produces this output:

```
tile at (0,0) = 6
tile at (0,1) = 6
tile at (0,2) = 4
tile at (0,3) = 4
tile at (0,4) = 4
tile at (0,5) = 4
tile at (0,6) = 4
tile at (0,7) = 6
tile at (0,8) = 6
tile at (1,0) = 6
Assert failed: TMX atlas index not found. Shall not happen
Assertion failed!
```

ricardoquesada:
I working on this. And I was able to reproduce it.

ricardoquesada:
fixed.

PR #16552, opened 2016-09-08
fix: staggered hex maps returns correct tile

ricardoquesada:
order was fixed in the array that contains the GIDs.

fixes Github issue #16512


ricardoquesada:
[ci rebuild]

PR #16558, opened 2016-09-09
Remove redundant semicolons

mogemimi:
This pull request removes extra semicolons which generate empty statements.


minggo:
good finding

PR #16562, opened 2016-09-11
Remove unnecessary inline keywords

mogemimi:
This pull request removes unnecessary `inline` keywords from member functions defined inside a class definition for the following reasons:
- In C++, any function defined inside the body of a class definition becomes implicitly inline function without `inline` keyword. (For more information, please see this FAQ entry: https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more.)
- It makes no difference whether you put explicit `inline` or not. [MSDN](https://msdn.microsoft.com/en-us/library/bw1hbe6y.aspx) says: _"In the class declaration, the functions were declared without the inline keyword. The inline keyword can be specified in the class declaration; **the result is the same.**"_
- According to [LLVM Coding Standards](http://llvm.org/docs/CodingStandards.html#don-t-use-inline-when-defining-a-function-in-a-class-definition), LLVM does not recommend the use of `inline` keyword when defining a function in a class definition.

So we can get rid of some `inline` keywords. This will simplify the code for better readability and also keeps backward compatibility.

Thank you for your time and consideration.

### DIFF
--- a/cocos/2d/CCAction.h
+++ b/cocos/2d/CCAction.h
@@ -119,17 +119,17 @@ public:
      *
      * @return A certain target.
      */
-    inline Node* getTarget() const { return _target; }
+    Node* getTarget() const { return _target; }
     /** The action will modify the target properties. 
      *
      * @param target A certain target.
      */
-    inline void setTarget(Node *target) { _target = target; }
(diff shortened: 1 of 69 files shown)


minggo:
thanks

PR #16569, opened 2016-09-13
[DON'T MERGE] Initial update so that Sprites work as before (with one exception)

stevetranby:
There's an issue still when you change texture from ETC1Alpha to another texture type.

The fundamental issue I see with the PR [16118](https://github.com/cocos2d/cocos2d-x/pull/16118)  is that ETC1 is treated as a normal use case when I think that it should almost have its own SpriteETC1Alpha class derived from Sprite overriding a couple of the methods.

However, people probably would like to be able to create a sprite and then set it's texture to some texture type (e.g. ETC1Alpha) and then later change it to a different texture type and vice versa.

This PR will attempt to find the best way to support that as well as allow for setting custom shader commands. If this is not straightforward then documentation may be necessary to note that if one wants to change a Sprite from one texture type to another may be required to update the GLProgramState manually after every setTexture call.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -289,7 +289,7 @@ bool Sprite::initWithTexture(Texture2D *texture, const Rect& rect, bool rotated)
         _quad.tr.colors = Color4B::WHITE;
         
         // shader state
-        // setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP)); 
+        setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP));
 
         // update texture (calls updateBlendFunc)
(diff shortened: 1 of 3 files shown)


minggo:
@stevetranby yep, i think you are right. But i don't know the meaning of the tile, what's the `one exception`? Could you please describe it in detail? Thanks.

stevetranby:
@minggo yeah the one exception is that it's currently unsupported to set a Sprite's texture to ETC1+Alpha (or at create/init time) and then later set it back to a regular non-ETC1+Alpha (PNG, 8888, etc) texture.

If keeping the ETC1+Alpha stuff integrated with Sprite as it is now then I think the solution is  probably to just **reset** the shader to the default shader when setting a new regular texture where the current (old) texture is ETC1+Alpha, but not resetting it if the old texture was also a regular texture (to support custom shaders correctly).

I'll try that, test with our game(s), and finally improve the current test to confirm each type of texture change.

minggo:
@stevetranby thanks, i think this PR resolved the `one exception`, right?

stevetranby:
I'll submit a new PR only after it's fully working.

PR #16578, opened 2016-09-14
Features: add events for window resize, focus, and unfocus

stevetranby:
Allow game to receive event for easy notification of resize/focus/unfocus window events for GLFW platforms.

### DIFF
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
@@ -423,7 +423,7 @@ bool GLViewImpl::initWithRect(const std::string& viewName, Rect rect, float fram
     glfwSetFramebufferSizeCallback(_mainWindow, GLFWEventHandler::onGLFWframebuffersize);
     glfwSetWindowSizeCallback(_mainWindow, GLFWEventHandler::onGLFWWindowSizeFunCallback);
     glfwSetWindowIconifyCallback(_mainWindow, GLFWEventHandler::onGLFWWindowIconifyCallback);
-    cglfwSetWindowFocusCallback(_mainWindow, GLFWEventHandler::onGLFWWindowFocusCallback);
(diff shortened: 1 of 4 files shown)


ricardoquesada:
[ci rebuild]

minggo:
Looks good to me. 
@ricardoquesada this PR passes the CI, can i merge it?

ricardoquesada:
@minggo I've just merged it. thanks.

sachingarg05:
Instead of dispatching custom events for these, we should use existing AppDelegate (cocos2d::Application) callbacks when these events occur.

```
virtual void applicationDidEnterBackground();
virtual void applicationWillEnterForeground();
virtual void applicationScreenSizeChanged(int newWidth, int newHeight);
```

Almost everyone already implements the Background/Foreground callbacks for iOS and Android. And the screenSizeChanged callback is called on orientation change on iOS and Android and is used by apps which support both landscape and portrait orientations.

If we use the same callbacks for desktop mode also, apps will have one cross platform way to support these events.

stevetranby:
I thought about adding them to the Application protocol, but decided that the custom event made more sense since it's a less rigid structure or interface. It's also less discoverable so I'm willing to change it if others agree with you.

Also, I determined that Background/Foreground are different on iOS/Android (where the app should fully pause) than Desktop (where the game could continue, continue at reduced fps, continue with muted sound, pause, etc).

If @minggo, @ricardoquesada, or others have a strong opinion or direction they think best here, I can submit a new PR that reverts this and adds app protocol methods.

sachingarg05:
I think that the framework should provide for one unified cross platform API. If instead of sending custom events, we call the already existing functions we can achieve that goal. Events would have been fine too, but we already have the callback routines so it seems to be fragmenting the API between platforms.

If an app requires different behavior for different platforms on background or resize, that can be done with #if checks within an app's implementation of appdelegate functions.

PR #16579, opened 2016-09-14
XCode sometimes gets confused with multiple 'Rect' types

stevetranby:
Doesn't have an issue with Point/Size/Rect normally, except in some instances with these MACROs probably since they're more "global" during compilation. I think MACROs should always have fully qualified names?


ricardoquesada:
good catch, thanks!

PR #16580, opened 2016-09-14
Fix -Wpessimizing-move warning when using Xcode 8

mogemimi:
This PR removes a superfluous `std::move` causing the following warning when building with `-Wpessimizing-move` compiler option on Xcode 8.0 (probably it can be reproduced on Xcode 7.3+):

```
cocos/editor-support/cocostudio/CCDataReaderHelper.cpp:150:13: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
        s = std::move(fs->getStringFromFile(filename));
            ^
cocos/editor-support/cocostudio/CCDataReaderHelper.cpp:150:13: note: remove std::move call here
        s = std::move(fs->getStringFromFile(filename));
            ^~~~~~~~~~                               ~
```


ricardoquesada:
sounds good to me. thanks!

PR #16585, opened 2016-09-15
Move skipBOM function outside to prevent shadowing warning

mogemimi:
I'm getting the following warning message when compiling libluacocos2d with Xcode 8.0:

```
cocos/scripting/lua-bindings/manual/CCLuaStack.cpp:851:36: declaration shadows a local variable [-Wshadow]

    auto skipBOM = [](const char*& chunk, int& chunkSize){
                  ^

cocos/scripting/lua-bindings/manual/CCLuaStack.cpp:851:48: declaration shadows a local variable [-Wshadow]

    auto skipBOM = [](const char*& chunk, int& chunkSize){
                                                      ^
```

This patch moves `skipBOM` function definition outside of `luaLoadBuffer` function to prevent those warnings.
Thanks!

### DIFF
--- a/cocos/scripting/lua-bindings/manual/CCLuaStack.cpp
+++ b/cocos/scripting/lua-bindings/manual/CCLuaStack.cpp
@@ -844,20 +844,25 @@ int LuaStack::luaLoadChunksFromZIP(lua_State *L)
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


minggo:
thanks

PR #16588, opened 2016-09-16
tidy up TextureAtlas::resizeCapacity

tankorsmash:
My intent is to reduce clutter and improve readability, so that going forward it's a little easier to edit this.

No behaviour change.


ricardoquesada (approved):
thanks.

PR #16591, opened 2016-09-18
Fix -Wreturn-type warning on Linux clang

colajam93:
This commit fixes the following warning on Linux clang 3.8.1

```
/home/user/local/cocos2d-x/cocos/audio/linux/AudioEngine-linux.cpp:221:1: warning: control may reach end of non-void function [-Wreturn-type]
}
^
```


minggo:
thanks

Issue #16541, opened 2016-09-07
ScriptingCore::evalString not work


flybird:
- cocos2d-x version: v3.13
- devices test on:iPhone stimulate
- developing environments
  - NDK version:android-ndk-r12b
  - Xcode version:7.3.1
  - VS version:
  - brower type and version:

Steps to Reproduce:
I use latest version 3.13. I found I call ScriptingCore::evalString but It don't fire. the code always return false, the "script" is nullptr,

> JSAutoCompartment ac(cx, global);
> JS::PersistentRootedScript script(cx);
> if (script == nullptr) {
> return false;
> }

I think it should be compare with point "&script"

>  if (&script == nullptr)

thuydx55:
I'm also getting this issue

DavidDeSimone:
I was able to trivially reproduce this by running the line 
`CCASSERT(ScriptingCore::getInstance()->evalString("(function() { cc.log ('World Hello!'); })();"), "Testing EvalString");` in AppDelegate.cpp after ScriptingCore initialization. The reasoning seems to be that we are initializing the JS::PersistentRootedScript without an initialization value, which will mean it's operator== against nullptr will return true. Like the reporter suggests we should use the overridden operator& of the script, comparing that against nullptr. 

Also on a slightly related note, I am unsure on why we are using a JS::PersistentRootedScript here over a JS::RootedScript. The script in this use case has a definite lifetime, and is not cached or stored globally, so I see no point for using a PersistentRooted<T> here. 

I will submit a PR that fixes this issue. @pandamicro 

PR is here https://github.com/cocos2d/cocos2d-x/pull/16597

pandamicro:
Yes, you are right, this was due to code merging, I used to new a PersistentRootedScript, and delete it at the end, but totally it's useless, let's use JS::RootedScript.

Sorry for the trouble.

PR #16597, opened 2016-09-19
Fixing issue where ScriptingCore::evalString would always return fals…

DavidDeSimone:
…e, and not evaluate the string, due to an improper null check.

### DIFF
--- a/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
+++ b/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
@@ -520,8 +520,8 @@ void ScriptingCore::string_report(JS::HandleValue val) {
 bool ScriptingCore::evalString(const char *string, JS::MutableHandleValue outVal, const char *filename, JSContext* cx, JS::HandleObject global)
 {
     JSAutoCompartment ac(cx, global);
-    JS::PersistentRootedScript script(cx);
-    if (script == nullptr) {
+    JS::RootedScript script(cx);
+    if (&script == nullptr) {
         return false;
     }
     
--- a/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
+++ b/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
@@ -521,10 +521,6 @@ bool ScriptingCore::evalString(const char *string, JS::MutableHandleValue outVal
 {
     JSAutoCompartment ac(cx, global);
     JS::RootedScript script(cx);
-    if (&script == nullptr) {
-        return false;
-    }
-    
     JS::CompileOptions op(cx);
     op.setUTF8(true);
     


DavidDeSimone:
Further background can be found in this issue: https://github.com/cocos2d/cocos2d-x/issues/16541

minggo:
@pandamicro could you please take a look?

pandamicro (approved):

minggo:
@DavidDeSimone could you please add a test case for this issue?

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -520,8 +520,8 @@ void ScriptingCore::string_report(JS::HandleValue val) {
 bool ScriptingCore::evalString(const char *string, JS::MutableHandleValue outVal, const char *filename, JSContext* cx, JS::HandleObject global)
 {
     JSAutoCompartment ac(cx, global);
-    JS::PersistentRootedScript script(cx);
-    if (script == nullptr) {
+    JS::RootedScript script(cx);
+    if (&script == nullptr) {
```
ricardoquesada: I think that `script` can't be `nullptr` with the new change as well.
And why did you change `PersistentRootedScript` to `RootedScript`? thanks.

DavidDeSimone: @ricardoquesada I believe you're correct. I don't believe that if check is necessary. It seems likely that if check is a hold over from when we created a PersistentRootedObject on the heap here. It makes sense to do that check if we are doing a dynamic allocation, but since we are creating a Rooted<T> on the stack, I don't believe we need to check the value of operator& on the Rooted<T>. I don't believe we if check other Rooted<T>s we create on the stack in other places in code, so it makes sense to not have that if check here. 

w.r.t the PersistentRootedScript, In the issue I wrote 

> Also on a slightly related note, I am unsure on why we are using a JS::PersistentRootedScript here over a JS::RootedScript. The script in this use case has a definite lifetime, and is not cached or stored globally, so I see no point for using a PersistentRooted here.

The spidermonkey API suggests that PersistentRooted<T>s should be used for objects with dynamic lifetime that could change based on runtime parameters. Since the lifetime of the script object here is known at compile time, it seems more proper to use a JS::Rooted<T> instead of a PersistentRooted<T>.

DavidDeSimone:
@minggo I will write up some test cases either tonight or tomorrow and add them to the PR.

DavidDeSimone:
Updated the PR based on feedback from @ricardoquesada

As for the tests, it seems that there isn't a binding for evalString, and because of that I'm not sure of the best way to add a test to the js testing framework for it. As far as I can see, the tests in tests/js-tests only test functionality available directly from the JS, while this is a c++ only function. 

I could add a binding for evalString in scripting core, and write a test that uses it, but I'm curious what other people think.

minggo:
@ricardoquesada any update about this PR since @DavidDeSimone updated the codes according to your comment.

DavidDeSimone:
@minggo @pandamicro It looks like @pandamicro fixed this issue in https://github.com/cocos2d/cocos2d-x/pull/16751, should we just close this PR now as duplicate?

minggo:
@DavidDeSimone thanks for the information

PR #16598, opened 2016-09-20
Fix implicit conversion warning

mogemimi:
Hello, I now get two warnings when building libcocos2d for Mac using Xcode 7.3.1. This fix tries to address the warnings.

```
cocos/platform/CCSAXParser.cpp:114:73: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
        SAXParser::textHandler(_ccsaxParserImp, (const CC_XML_CHAR *)s, len);
        ~~~~~~~~~                                                       ^~~
cocos/platform/CCSAXParser.cpp:166:33: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
        parser.parse<>(xmlData, dataLength);
        ~~~~~~                  ^~~~~~~~~~
```

### DIFF
--- a/cocos/platform/CCSAXParser.cpp
+++ b/cocos/platform/CCSAXParser.cpp
@@ -163,7 +163,7 @@ bool SAXParser::parseIntrusive(char* xmlData, size_t dataLength)
 
     rapidxml::xml_sax3_parser<> parser(&printer);
     try {
-        parser.parse<>(xmlData, dataLength);
+        parser.parse<>(xmlData, static_cast<int>(dataLength));
         return true;
     }
     catch (rapidxml::parse_error& e)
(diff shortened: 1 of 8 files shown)


cocos/platform/CCSAXParser.cpp
```diff
@@ -111,7 +111,7 @@ class RapidXmlSaxHander : public rapidxml::xml_sax2_handler
 
     void xmlSAX2Text(const char *s, size_t len) override
     {
-        SAXParser::textHandler(_ccsaxParserImp, (const CC_XML_CHAR *)s, len);
+        SAXParser::textHandler(_ccsaxParserImp, (const CC_XML_CHAR *)s, static_cast<int>(len));
```
minggo: may be we should change to use `size_t` in textHandler.

mogemimi: Thank you for reviewing! I think so, too. I'll work on it and submit soon.

mogemimi:
I've submitted a new commit 2918d4eea74e7be7a8203172d63d3f6276289095 and changed `len` type from `int` to `size_t`. Could you please review this again? @minggo

minggo:
thanks

PR #16599, opened 2016-09-20
Fix bug ：expoEaseInOut

csy1988525:
When the parameter in function expoEaseInOut is 1.0 , it return 0.99995

### DIFF
--- a/cocos/2d/CCTweenFunction.cpp
+++ b/cocos/2d/CCTweenFunction.cpp
@@ -307,9 +307,12 @@ float expoEaseInOut(float time)
     {
         time = 0.5f * powf(2, 10 * (time - 1));
     }
-    else
+    else if (time < 2)
     {
         time = 0.5f * (-powf(2, -10 * (time - 1)) + 2);
+    } else
+    {
+        time = 1.0f;
     }
 
     return time;


minggo:
@csy1988525 how to reproduce it? can you add a test case? And please use english, especially in commit message.

csy1988525:
void TestLayer::test()
{
    auto\* updateFunc = EaseExponentialInOut::create(CCActionTween::create(0.7f, "test", 0, 10000));
    updateFunc->setTarget(this);
    this->runAction(updateFunc);
}

void TestLayer::updateTweenAction(float value, const std::string & key)
{
     CCLOG("%d"m,value)
}

when we use such case, we can get log from 0 - 9995, witch doesn't contain 10000。the reason is what I list in the question conment. when the parameter in function "expoEaseInOut" is 1.0f,  it's return a float 0.9995.

stevetranby:
I feel like this would be better?

``` cpp
float expoEaseInOut(float time)
{
    if(time == 0 || time == 1) return time;
    if (time < 0.5f) {
        return 0.5f * powf(2, 10 * (time / 0.5f - 1));
    }
    return 0.5f * (-powf(2, -10 * (time / 0.5f - 1)) + 2);
}
```

minggo:
Yep, i think @stevetranby 's implementation is clear. And i don't know why it should do 

``` c++
time /= 0.5f;
```

at fist.

@ricardoquesada do you know?

ricardoquesada:
@minggo I guess it was done to make the math easier, like in the other expo functions.

`time /= 0.5f` is the same as doing `time *= 2`.  But `time *= 2` looks cleaner.

@stevetranby patch looks better. And we can replace `time / 0.5f` by `time * 2` in his patch.

minggo:
@csy1988525 will you please update the PR according to the suggestion?

minggo:
send a new PR https://github.com/cocos2d/cocos2d-x/pull/17340

PR #16600, opened 2016-09-20
add sprite used in creator

dabingnn:


ricardoquesada:
[ci rebuild]

ricardoquesada:
So, I finally reached a point where I need "scale 9 sprite" inside sprite.
I don't know all the feature needed by Creator, but I think we can add scale-9-support inside Sprite without creating a new class.

dabingnn:
Generally speaking, sprite used by creator is not a scale9 sprite at all. Anything needed to be drawn with a picture is sprite, simple(like sprite in cocos2d-x), scale9, tiled(tiling the picture), filled(just like progress bar) is supported in this class.

ricardoquesada:
@dabingnn ok. let's add those features into `Sprite` then... or what is the limitation/issue that those features can't be added into Sprite?

it will be very confusing to have 3 classes: Sprite, Spritev2 and Scale9Sprite.

I can help you with that if you want. Let's have a powerful Sprite API

dabingnn:
@ricardoquesada Thanks. Personally, I prefer to use only one class `sprite` to fulfill the requirements in creator, and it should be our final target. And I think there are two ways to achieve this:
1. Refine sprite in cocos2d-x, reimplement the feature in sprite v2.(scale9 sprite is not used in creator 'sprite component' at all).
2. Keep the three version in cocos2d-x first, and refactor all the three into a sprite in next big version, maybe cocos2d-x 4.0.

And our current requirements:
1. cocos creator lua version need it ASAP, confirmed by @dualface 
2. cocos creator cpp version need these features if we develop it.
3. keep compatibility, (I am not sure).

The first solution is the optimal solution. However, in my opinion, it is impossible to add these features and keep compatibility, and it will be time consuming. 
The second solution is not elegant, but it will meet these three requirements.

Any thoughts?

ricardoquesada:
@dabingnn 

>  it is impossible to add these features and keep compatibility

could you explain why? If it is impossible, then we shouldn't do it. but I would like to understand why it is impossible to do them without breaking compatibility.
thanks!

Issue #16603, opened 2016-09-23
 ImageView used  loadtexture() with etc, the alpha is wrong!


huangzhentao:
- cocos2d-x version:3.13.1
- devices test on:all
- developing environments
  - NDK version:
  - Xcode version:
  - VS version:
  - brower type and version:

Steps to Reproduce:
1. ImageView used png file first
2. then loadtexture() with etc support alpha

huangzhentao:
I think  after loadTexture() ,it must call "_imageRenderer->setState(_imageRenderer->getState());"
to update GLProgramState

huangzhentao:
![image](https://cloud.githubusercontent.com/assets/5028674/18779058/e34c9432-81a9-11e6-8147-ae7f0c6ef17b.png)
in UIScale9Sprite.cpp
maybe use "this->setState(_brightState);"   more better.  And fix all Scale9Sprite problem about alpha.
Thinks @halx99

halx99:
@huangzhentao  Thanks for your feedback, I post a PR to fix this issue.

minggo:
Thanks @halx99

PR #16604, opened 2016-09-23
Fix issue: 16603

halx99:
see: https://github.com/cocos2d/cocos2d-x/issues/16603


minggo:
@zilongshanren could you please take a look.

zilongshanren:
@minggo 
it looks good to me.

PR #16609, opened 2016-09-24
lua: generate bindings for Sprite3DMaterial

Glidos:
Add Sprite3DMaterial to the list of classes for autogeneration of
lua bindings.

Add a new test case to the end of Sprite3DTest.lua showing the
use of outline.material

### DIFF
--- a/tests/lua-tests/src/Sprite3DTest/Sprite3DTest.lua
+++ b/tests/lua-tests/src/Sprite3DTest/Sprite3DTest.lua
@@ -1295,6 +1295,63 @@ end
 function Sprite3DNormalMappingTest:onExit()
 end
 
+----------------------------------------
+----Sprite3DMaterialTest
+----------------------------------------
+local Sprite3DMaterialTest = class("Sprite3DMaterialTest", function ()
+    local layer = cc.Layer:create()
+    Helper.initWithLayer(layer)
+    return layer
+end)
+
+function Sprite3DMaterialTest:ctor()
+    -- body
+    self:init()
+end
+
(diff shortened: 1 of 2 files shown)


minggo:
Error happened when running the test case on Mac:

```
cocos2d: ERROR: 0:15: 'precision' : syntax error: syntax error

cocos2d: ERROR: Failed to compile vertex shader
[LUA-print] ASSERT FAILED ON LUA EXECUTE: invalid shader
stack traceback:
    [string "Sprite3DTest/Sprite3DTest.lua"]:1342: in function 'onEnter'
    [string "Sprite3DTest/Sprite3DTest.lua"]:1318: in function <[string "Sprite3DTest/Sprite3DTest.lua"]:1316>
```

And i think you can reuse resources in `/tests/cpp-tests/Resources/Materials`.

Glidos:
Okay, I've changed the test to use outline.material and simplified the test in other ways.

minggo:
@Glidos thanks, you don't have to update files in `cocos/scripting/lua-bindings/auto`. These files will be auto-generated.

Glidos:
@minggo okay, autogenerated files removed. So how does that work? I notice the autogenerated files do end up committed to the repo. Is it part of the commit hook?

Glidos:
So without the autogenerated changes, the test build now fails. I'm confused.

Glidos:
Oh, and now it's passed. Okay, looks ready to go now.

minggo:
@Glidos the CI will generate binding codes automatically when checking a PR.

Glidos:
@minggo thank you

PR #16637, opened 2016-10-02
Adding 'addImageAsyncImmediate' to CCTextureCache. 

DavidDeSimone:
This will allow engine users to load multiple textures async across multiple threads, or load a texture before other textures in the texture loading queue.

In several of our cocos2d-x projects, we found that increasing the number of texture loading threads would reduce our loading times by a measurable amount. We found that the cleanest way to accomplish this was adding a "loadImageAsyncImmediate" function, which will spin up a new thread dedicated to loading that texture. This is designed to mimic the architecture of HttpClient::sendImmediate. 

This feature should be almost entirely additive, and have minimal impact to users who do not opt-in to this new functionality. 

Benefits: 
- On systems with a high core count, using this may yield solid performance gains in texture load times.
- Engine users can prioritize texture loads over others.

### DIFF
--- a/cocos/renderer/CCTextureCache.cpp
+++ b/cocos/renderer/CCTextureCache.cpp
@@ -31,6 +31,7 @@ THE SOFTWARE.
 #include <stack>
 #include <cctype>
 #include <list>
+#include <future>
 
 #include "renderer/CCTexture2D.h"
 #include "base/ccMacros.h"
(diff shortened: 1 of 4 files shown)


cocos/renderer/CCTextureCache.cpp
```diff
@@ -31,6 +31,7 @@ THE SOFTWARE.
 #include <stack>
 #include <cctype>
 #include <list>
+#include <future>
```
dumganhar, line 34: Why should include `future` header file?
Is there anywhere we using `future` functionalities in this Pull Request?

dumganhar:
I think the best way to load textures in multi-threads is to implement a common `ThreadPool` for cocos2d-x. Then any modules in cocos2d-x could use it and it's also easier for maintaining.

cocos/renderer/CCTextureCache.cpp
DavidDeSimone: This was an accidental inclusion, I will remove it.

DavidDeSimone:
@dumganhar I think either using a thread pool, or the standard libraries std::async, makes sense to me. I used the approach given in this PR due to the fact that it was additive and opt-in, which was useful for my use case. However I think that a better long term solution would be to just use a thread pool that can be utilized by different facilities in cocos2d-x. I see that there is CCThreadPool in the experimental namespace used for audio, do you know anything about it? If there is anything holding it back from being used in texture loading/networking?

dumganhar:
>  I see that there is CCThreadPool in the experimental namespace used for audio, do you know anything about it? 

@DavidDeSimone , yes, CCThreadPool was added by me for AudioEngine engine on Android two month ago. It could be a candidate for our common thread pool. The reason why we just put it in `cocos/audio/android` is that 
- Keep it for internal use (private), we could replace it if we could find another better implementation.
- Make sure it's stable enough before make it visible for other modules or make it public.

> If there is anything holding it back from being used in texture loading/networking?

Yes, it could be used for texture loading / networking if you wish, you could just modify no much code for that purpose.

minggo:
How about overload `TextureCache::addImageAsync()`, and re-use codes as more as possible?

DavidDeSimone:
@dumganhar 
What is your opinion on the stability of CCThreadPool, now that it has seen a few months in production? Do you think that it needs more time marked as experimental before other modules in the engine start using it?

For experimentation, I created another branch that has texture loading use CCThreadPool. The texture loading on that branch has better performance then this PR, with less code changed. If CCThreadPool is stable, I would prefer to cancel this PR and open another for the Thread Pool solution.

My preliminary findings show that these multi-threaded optimizations greatly decrease texture load times on my windows machine (50% reduction in texture load times for 30 concurrently loaded textures of various sizes). I have yet to test mobile platforms, but I will report my findings when I do.

dumganhar:
@DavidDeSimone , yep, CCThreadPool has implemented a thread pool. Probably, we need to discuss the user experience of thread pool API usage before merging it into `core`. I definitely agree that current implementation of CCThreadPool (https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/audio/android/CCThreadPool.h) still has a lot that could be improved.

@ricardoquesada , @minggo , please visit that file and see whether the public methods of `CCThreadPool` meet our requirement.

@DavidDeSimone , what do you think of current API of CCThreadPool since you have already used it?

DavidDeSimone:
@dumganhar I think the public API of CCThreadPool is solid and straight forward, however I feel that it would make sense for `ThreadPool *ThreadPool::newCachedThreadPool` to have default values for the shrink interval, shrink step, and stretch step. Users of the API may have an idea of how many threads minimum and maximum they would like, but not know proper values for the shrink interval, step, etc. It makes sense for the implementation to provide a sane default in that case.

dumganhar:
@DavidDeSimone , yep, it will make users feel more comfortable if those arguments  have a default value. Thanks for your feedback.

ricardoquesada:
@dumganhar Could you post a few use cases? for example, how would you use it load load a some textures and sound files? thanks.

dumganhar:
@ricardoquesada , We are discussing the API of Thread Pool, asynchronous texture loading and file loading could use a thread pool rather than current implementation that is a single thread.
The main API of Thread Pool is:

``` c++
class ThreadPool
{
public:

    enum class TaskType
    {
        DEFAULT = 0,
        NETWORK,
        IO,
        AUDIO,
        USER = 1000
    };

    /*
     * Gets the default thread pool which is a cached thread pool with default parameters.
     */
    static ThreadPool *getDefaultThreadPool();

    /*
     * Destroys the default thread pool
     */
    static void destroyDefaultThreadPool();

    /*
     * Creates a cached thread pool
     * @note The return value has to be delete while it doesn't needed
     */
    static ThreadPool *newCachedThreadPool(int minThreadNum, int maxThreadNum, int shrinkInterval, int shrinkStep, int stretchStep);

    /*
     * Creates a thread pool with fixed thread count
     * @note The return value has to be delete while it doesn't needed
     */
    static ThreadPool *newFixedThreadPool(int threadNum);

    /*
     * Creates a thread pool with only one thread in the pool, it could be used to execute multiply tasks serially in just one thread.
     * @note The return value has to be delete while it doesn't needed
     */
    static ThreadPool *newSingleThreadPool();

    // the destructor waits for all the functions in the queue to be finished
    ~ThreadPool();

    /* Pushs a task to thread pool
     *  @param runnable The callback of the task executed in sub thread
     *  @param type The task type, it's TASK_TYPE_DEFAULT if this argument isn't assigned
     *  @note This function has to be invoked in cocos thread
     */
    void pushTask(const std::function<void(int /*threadId*/)>& runnable, TaskType type = TaskType::DEFAULT);
```

 Basically, you just need to create a kind of thread pool and push your task to it. The task will be executed by one of threads in the pool.

Usage:

``` c++
// New a thread pool

auto pool = ThreadPool::newCachedThreadPool(2, 10); //minThreadNum: 2, maxThreadNum: 10

// Push tasks
pool->pushTask(task1);
pool->pushTask(task2);
pool->pushTask(task3);
pool->pushTask(task4);

// Task 1~4 will be executed by 4 threads at the same time.

// Destroy thread pool
delete pool;
```

dumganhar:
@ricardoquesada , the API design is inspired by  Java Thread Pool (http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html)

DavidDeSimone:
@ricardoquesada @dumganhar An example of using CCThreadPool to manage texture loading

``` c++
void TextureCache::addImageAsync(const std::string &path, const std::function<void(Texture2D*)>& callback)
{
    Texture2D *texture = nullptr;
    // ... Current Implementation ...

    if (_threadPool == nullptr)
    {
        int numCores = std::thread::hardware_concurrency(); // May return 0 if unable to detect.
       _threadPool = ThreadPool::newCachedThreadPool(0, std::max(numCores, 1),
                                                    5.0f, 2, 2);
    }

    _threadPool->pushTask([this, data] (int taskid) 
    {
        data->loadSuccess = data->image.initWithImageFileThreadSafe(data->filename);
        // ETC1 ALPHA supports.
        if (data->loadSuccess && data->image.getFileType() == Image::Format::ETC && !s_etc1AlphaFileSuffix.empty())
        { // check whether alpha texture exists & load it
            auto alphaFile = data->filename + s_etc1AlphaFileSuffix;
            if (FileUtils::getInstance()->isFileExist(alphaFile))
                data->imageAlpha.initWithImageFileThreadSafe(alphaFile);
        }

        _responseMutex.lock();
        _responseQueue.push_back(data);
        _responseMutex.unlock();
    });
}
```

DavidDeSimone:
@dumganhar Another option that we have is to use std::async. I think that std::async may be  the cleanest in that it's 
- standard, and available since c++11
- fairly performant (though many aspects of it are implementation dependent, such as pooling behavior and thread reuse)

Leveraging the standard library can be a good way of avoiding the development overhead of maintaining a custom thread pool solution used in multiple game systems.

dumganhar:
@DavidDeSimone , I don't know whether `std::async` uses  a thread pool internally.
But `std::async` is strictly depended on the compiler supports c++11 well. I remember that `std::async` could not be supported well on android with `armeabi` architecture.

Yes, if all platforms cocos2d-x supports could use `std::async` and if it could be configured, for example, setting the min & max thread count `std::async` will use, that will be great to use this standard function.

DavidDeSimone:
@dumganhar Unfortunately, std::async does not give the user much control over the underlying behavior of the multi-threading, it's entirely implementation defined. I have also not tested it on armeabi recently (though I know it works properly on armeabi-v7a).

DavidDeSimone:
After talking to community members and engine maintainers, I don't believe adding 'getTextureAsyncImmediate' right thing for cocos2d-x. I think for now it will be best for me to just close this PR.

I do think that this PR led to solid discussion about multi-threading opportunities in cocos. I believe that there is a big opportunity to increase performance in the engine by using tools like CCThreadPool for jobs like texture loading and networking. 

@dumganhar if you want any help on testing/iterating on CCThreadPool, or other potential threading solutions, I'm more than willing to help.

@minggo

minggo:
Thanks @DavidDeSimone .

PR #16640, opened 2016-10-03
Remove unused functions

mogemimi:
Hello, I get the following warnings during compilation with libcocos2d under Xcode 8.0.

```
cocos/2d/CCDrawNode.cpp:79:20: warning: unused function 'v2fforangle' [-Wunused-function]
extensions/physics-nodes/CCPhysicsDebugNode.cpp:71:14: warning: unused function 'cpVertArray2ccpArrayN' [-Wunused-function]
```

These functions don't seem to be used anywhere, and I think we can remove them because they are `static` function that has internal linkage. So this pull request removes them to prevent the compiler warnings.

Thanks.

### DIFF
--- a/cocos/2d/CCDrawNode.cpp
+++ b/cocos/2d/CCDrawNode.cpp
@@ -76,11 +76,6 @@ static inline float v2fdot(const Vec2 &p0, const Vec2 &p1)
     return  p0.x * p1.x + p0.y * p1.y;
 }
 
-static inline Vec2 v2fforangle(float _a_)
-{
-    return v2f(cosf(_a_), sinf(_a_));
-}
-
 static inline Vec2 v2fnormalize(const Vec2 &p)
 {
     Vec2 r(p.x, p.y);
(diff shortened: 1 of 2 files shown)


stevetranby:
I'm curious what the purpose of the duplicate methods? I would be surprised if the assembly using the Vec2 class inline methods wasn't identical?

Maybe all of them should be removed in favor of the Vec2 class instance versions? or add them as additional static methods. If there's one or a few specific to DrawNode only then those can stay.

E.g.

``` cpp
Vec2 n = (b - a).getPerp().getNormalized();
// instead of the current
Vec2 n = v2fnormalize(v2fperp(v2fsub(b, a)));
```

stevetranby:
I mean is this note still true and relevant?

```
// Vec2 == CGPoint in 32-bits, but not in 64-bits (OS X)
// that's why the "v2f" functions are needed
```

mogemimi:
Cocos2d-x v2 (2012) didn't have the `Vec2` class, it was added to master since cocos2d-x v3 (commit: 0e2969f4955d7d5a094270efb5b59a8a1c1ee221). Instead, `Vertex2F` was originally used at DrawNode until commit 2e315fb2159c9dfa25555c9fb7e048d255ae42d7 (2014) from d2a57239739d362dadf02f67335bde02bc69a20e (2012).

Due to the historical background, DrawNode now use duplicate functions which are equivalent of `Vec2`. I guess we can completely replace them with `Vec2`'s function.

mogemimi:
[ci rebuild]

ricardoquesada:
thanks.

PR #16642, opened 2016-10-04
Performance tweak: use range-based for-loops instead of iterator-based for-loops.

Rubonnek:
The following error messages related to the ++ and -- operators are shown by `cppcheck`:

```
[2d/CCAutoPolygon.cpp:517]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[2d/CCAutoPolygon.cpp:556]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[2d/CCAutoPolygon.cpp:572]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[2d/CCAutoPolygon.cpp:588]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[audio/win32/SimpleAudioEngine.cpp:86]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[audio/win32/SimpleAudioEngine.cpp:208]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[audio/win32/SimpleAudioEngine.cpp:226]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[audio/win32/SimpleAudioEngine.cpp:235]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[editor-support/cocosbuilder/CCNodeLoaderLibrary.cpp:73]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[editor-support/cocostudio/CCActionManagerEx.cpp:170]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[editor-support/cocostudio/CCDataReaderHelper.cpp:508]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[network/HttpClient-android.cpp:470]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[network/HttpClient-android.cpp:508]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[platform/linux/CCDevice-linux.cpp:265]: (performance) Prefer prefix ++/-- operators for non-primitive types.
[scripting/js-bindings/manual/ScriptingCore.cpp:669]: (performance) Prefer prefix ++/-- operators for non-primitive types.
```

This pull request is rather self-explanatory, but prefixing ++ for non-primitive data types will yield a performance boost.

### DIFF
--- a/cocos/2d/CCAutoPolygon.cpp
+++ b/cocos/2d/CCAutoPolygon.cpp
@@ -514,9 +514,9 @@ std::vector<Vec2> AutoPolygon::expand(const std::vector<Vec2>& points, const coc
     ClipperLib::Path subj;
     ClipperLib::PolyTree solution;
     ClipperLib::PolyTree out;
-    for(const auto& it : points)
+    for(const auto& pt : points)
     {
-        subj << ClipperLib::IntPoint(it.x* PRECISION, it.y * PRECISION);
+        subj << ClipperLib::IntPoint(pt.x* PRECISION, pt.y * PRECISION);
     }
     ClipperLib::ClipperOffset co;
     co.AddPath(subj, ClipperLib::jtMiter, ClipperLib::etClosedPolygon);
@@ -569,9 +569,9 @@ TrianglesCommand::Triangles AutoPolygon::triangulate(const std::vector<Vec2>& po
         return TrianglesCommand::Triangles();
     }
     std::vector<p2t::Point*> p2points;
-    for(const auto& it : points)
+    for(const auto& pt : points)
     {
-        p2t::Point * p = new (std::nothrow) p2t::Point(it.x, it.y);
+        p2t::Point * p = new (std::nothrow) p2t::Point(pt.x, pt.y);
         p2points.push_back(p);
     }
     p2t::CDT cdt(p2points);
@@ -704,6 +704,7 @@ PolygonInfo AutoPolygon::generateTriangles(const Rect& rect, const float& epsilo
     ret.rect = realRect;
     return ret;
 }
+
 PolygonInfo AutoPolygon::generatePolygon(const std::string& filename, const Rect& rect, const float epsilon, const float threshold)
 {
     AutoPolygon ap(filename);
--- a/cocos/2d/CCAutoPolygon.cpp
+++ b/cocos/2d/CCAutoPolygon.cpp
@@ -514,9 +514,9 @@ std::vector<Vec2> AutoPolygon::expand(const std::vector<Vec2>& points, const coc
     ClipperLib::Path subj;
     ClipperLib::PolyTree solution;
     ClipperLib::PolyTree out;
-    for(std::vector<Vec2>::const_iterator it = points.begin(); it<points.end(); ++it)
+    for(const auto& it : points)
     {
-        subj << ClipperLib::IntPoint(it-> x* PRECISION, it->y * PRECISION);
+        subj << ClipperLib::IntPoint(it.x* PRECISION, it.y * PRECISION);
     }
     ClipperLib::ClipperOffset co;
     co.AddPath(subj, ClipperLib::jtMiter, ClipperLib::etClosedPolygon);
@@ -553,9 +553,9 @@ std::vector<Vec2> AutoPolygon::expand(const std::vector<Vec2>& points, const coc
         p2 = p2->GetNext();
     }
     auto end = p2->Contour.end();
-    for(std::vector<ClipperLib::IntPoint>::const_iterator pt = p2->Contour.begin(); pt < end; ++pt)
+    for(const auto& pt : p2->Contour)
     {
-        outPoints.push_back(Vec2(pt->X/PRECISION, pt->Y/PRECISION));
+        outPoints.push_back(Vec2(pt.X/PRECISION, pt.Y/PRECISION));
     }
     return outPoints;
 }
@@ -569,9 +569,9 @@ TrianglesCommand::Triangles AutoPolygon::triangulate(const std::vector<Vec2>& po
         return TrianglesCommand::Triangles();
     }
     std::vector<p2t::Point*> p2points;
-    for(std::vector<Vec2>::const_iterator it = points.begin(); it<points.end(); ++it)
+    for(const auto& it : points)
     {
-        p2t::Point * p = new (std::nothrow) p2t::Point(it->x, it->y);
+        p2t::Point * p = new (std::nothrow) p2t::Point(it.x, it.y);
         p2points.push_back(p);
     }
     p2t::CDT cdt(p2points);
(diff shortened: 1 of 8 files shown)


stevetranby:
I wonder if converting them to the range-for version would be better. If supported I believe it is as fast or better.

cocos/audio/win32/SimpleAudioEngine.cpp
```diff
@@ -83,7 +83,7 @@ void SimpleAudioEngine::end()
     {
         delete p->second;
         p->second = nullptr;
-        p++;
```
ricardoquesada, line 86: I think there is no need for this in this particular case.

ricardoquesada:
as @stevetranby  said, or just add `auto` in some of them to make it more readable

Rubonnek:
@stevetranby You're right. I performed two simple tests and range-based for-loops are way faster at least in Linux:

Here's the code for the iterator-based for-loop test:

```
#include <vector>
using namespace std;

int main ()
{
        std::vector<int> v (500000000, 0);
        for( std::vector<int>::const_iterator it = v.begin(); it < v.end(); it++ )
        {
                ;
        }
        return 0;
}
```

Result:

```
$ time ./iterator_based_for_loop 
./iterator_based_for_loop  12.90s user 0.54s system 99% cpu 13.470 total
```

And here's the code for the range-based for-loop:

```
#include <vector>
using namespace std;

int main ()
{
        std::vector<int> v (500000000, 0);
        for( auto& i : v )
        {
                ;
        }
        return 0;
}
```

Result:

```
$ time ./range_based_for_loop
./range_based_for_loop  7.51s user 0.57s system 99% cpu 8.106 total
```

cocos/audio/win32/SimpleAudioEngine.cpp
Rubonnek: You're right. I just fixed that in the latest commit.

stevetranby:
Since you have your benchmark handy try saving the end iterator. The excess time prob mostly due to recalc end iterator value. 
'''
Auto it_end = it.end; it < it_end;
'''

stevetranby:
Still think range for is best going forward unless specific reasons not to.

Rubonnek:
Yeah -- range-based for-loops are still better. I did as you said:

```
#include <vector>
using namespace std;

int main ()
{
        std::vector<int> v (500000000, 0);
        auto it_end = v.end();
        for( std::vector<int>::const_iterator it = v.begin(); it < it_end; it++ )
        {
                ;
        }
        return 0;
}
```

And did not expect the huge time save at all:

```
$ time ./iterator_based_for_loop
./iterator_based_for_loop  8.28s user 0.37s system 99% cpu 8.676 total
```

but range-based for-loops are still better for about .77 seconds, give or take, in this case.

Rubonnek:
Just now I prefixed the iterator with ++ in the for loop to get the following:

```
#include <vector>
using namespace std;

int main ()
{
        std::vector<int> v (500000000, 0);
        auto it_end = v.end();
        for( std::vector<int>::const_iterator it = v.begin(); it < it_end; ++it )
        {
                ;
        }
        return 0;
}
```

And surprisingly it is now faster than the range-based for-loop:

```
$ time ./iterator_based_for_loop 
./iterator_based_for_loop  6.64s user 0.38s system 99% cpu 7.041 total
```

But enabling the `-O2` optimization flag makes these two loops virtually the same time-wise. The range-based for-loop seems to be consistently and slightly faster with this flag enabled:

```
g++ -O2 range_based_for_loop.cpp -o range_based_for_loop
g++ -O2 iterator_based_for_loop.cpp -o iterator_based_for_loop

$ time ./range_based_for_loop 
./range_based_for_loop  0.33s user 0.40s system 99% cpu 0.735 total

$ time ./iterator_based_for_loop
./iterator_based_for_loop  0.35s user 0.38s system 99% cpu 0.734 total
```

And, of course, I dropped the Linux caches before running each of these tests:

```
# echo 3 > /proc/sys/vm/drop_caches
```

cocos/2d/CCAutoPolygon.cpp
```diff
@@ -514,9 +514,9 @@ std::vector<Vec2> AutoPolygon::expand(const std::vector<Vec2>& points, const coc
     ClipperLib::Path subj;
     ClipperLib::PolyTree solution;
     ClipperLib::PolyTree out;
-    for(std::vector<Vec2>::const_iterator it = points.begin(); it<points.end(); it++)
+    for(const auto& it : points)
```
Dimon4eg: better to name "pt" instead of "it"

cocos/2d/CCAutoPolygon.cpp
```diff
@@ -569,9 +569,9 @@ TrianglesCommand::Triangles AutoPolygon::triangulate(const std::vector<Vec2>& po
         return TrianglesCommand::Triangles();
     }
     std::vector<p2t::Point*> p2points;
-    for(std::vector<Vec2>::const_iterator it = points.begin(); it<points.end(); it++)
+    for(const auto& it : points)
```
Dimon4eg: better to name "pt" instead of "it"

cocos/editor-support/cocostudio/CCDataReaderHelper.cpp
```diff
@@ -504,18 +504,18 @@ void DataReaderHelper::addDataAsyncCallBack(float dt)
...
-        if (*i == configFile)
-        {
-            it = i;
-        }
-    }
-
-    if (it != _configFileList.end())
-    {
-        _configFileList.erase(it);
+		if (i == configFile)
```
Dimon4eg: Broken code formatting - mix spaces and tabs

cocos/network/HttpClient-android.cpp
```diff
@@ -133,18 +133,16 @@ class HttpURLConnection
         if(!headers.empty())
         {
             /* append custom headers one by one */
-            for (HttpRequestHeadersIter it = headers.begin(); it != headers.end(); ++it)
+            for (auto& it : headers)
```
Dimon4eg: better to name "header" instead of "it"

cocos/network/HttpClient-android.cpp
```diff
@@ -502,16 +497,15 @@ class HttpURLConnection
             cookiesInfoVec.push_back(co);
         }
 
-        std::vector<CookiesInfo>::iterator cookiesIter = cookiesInfoVec.begin();
         std::string sendCookiesInfo = "";
         int cookiesCount = 0;
-        for (; cookiesIter != cookiesInfoVec.end(); cookiesIter++)
+        for (auto& cookiesIter : cookiesInfoVec)
```
Dimon4eg: better to name "cookieInfo" instead of "cookiesIter"

cocos/platform/linux/CCDevice-linux.cpp
```diff
@@ -262,14 +262,14 @@ class BitmapDC
                         textLines.push_back(currentLine);
                         currentLine.reset();
                         currentPaintPosition = 0;
-                        for ( it = tempGlyphs.begin(); it != tempGlyphs.end(); it++ ) {
+                        for ( auto& it : tempGlyphs ) {
```
Dimon4eg: better to name "glyph" instead of "it"

cocos/editor-support/cocostudio/CCDataReaderHelper.cpp
```diff
@@ -504,18 +504,18 @@ void DataReaderHelper::addDataAsyncCallBack(float dt)
 
 void DataReaderHelper::removeConfigFile(const std::string& configFile)
 {
-    std::vector<std::string>::iterator it = _configFileList.end();
-    for (std::vector<std::string>::iterator i = _configFileList.begin(); i != _configFileList.end(); i++)
+    std::vector<std::string>::iterator it = _configFileList.begin();
+    for (auto& i : _configFileList)
```
Dimon4eg: No reason to use range based for here because iterator in any case needed for removing.

Rubonnek:
@Dimon4eg Thanks.

ricardoquesada:
looks good to me... merging.
Thanks!

PR #16655, opened 2016-10-06
Correct the Skybox fov

Glidos:
The Skybox implementation didn't correctly account for the camera's field of view and aspect ratio. This
showed particularly badly if the camera was pointed down and rotated, which produced a deformation
of the image as it rotated. On of these commits corrects that.

Also, while analysing this, it was noticed that the skybox implementation was using an unnecessarily
complicated mesh with faces that were never rendered. That is corrected in one of the other commits.

The one other commit removes all setScale applied to skyboxes from the test code. The skybox
implementation makes no use of the MVP transform and hence is uneffected by scaling operations.


minggo:
@songchengjiang could you please take a look?

songchengjiang:
@minggo Correct!
@Glidos Thanks for your solution!

Glidos:
Thank you @minggo and @songchengjiang. Is there anything else I need to do to get this merged in?

PR #16658, opened 2016-10-07
Remove unnecessary arguments from some lua cc.mat4 methods

Glidos:
cc.mat4 methods transformVector, createTranslation and createRotation all take an argument called dst
that is unused, either for passing a value in or out. This commit removes the unnecessary argument and
updates the test files correspondingly.


cocos/scripting/lua-bindings/script/cocos2d/Cocos2d.lua
```diff
@@ -511,54 +511,6 @@ function cc.mat4.createIdentity()
...
-    dst[12] = 0.0
-
-    dst[13] = 0.0
-    dst[14] = 0.0
-    dst[15] = 0.0
-    dst[16] = 1.0
-
-    return dst
-end
-
```
minggo, line 561: remove these two functions may break compatibility.

Glidos: It's not clear from the diff, but those two functions were defined a second time in that same file, hence the ones I've deleted were overwritten.

minggo: I think it is a little different. `function cc.mat4.createRotation(q, dst)` will modify `dst`, not just return a table.

Glidos: True, but it is currently impossible to call because of the second assignment in the same file. The current code is in effect:
cc.mat4.createRotation = x
cc.mat4.createRotation = y

minggo: yep, you are right

Issue #16635, opened 2016-10-01
System font enhancements on OSX and macOS


slackmoehrle:
this was reported on the forums: http://discuss.cocos2d-x.org/t/labels-with-system-fonts-on-macos/32394

My understanding is `SystemFont` is a label type that uses the default `system font` and `font size`. This is a font that is meant not to have its properties changed. You should think of it as **system font, system rules**.

User is asking if this behavior can be changed and provides some thoughts in the thread on how to change it.

kmaker:
The font name given is used as a "font family name" on OS X and not as a "font name". That for me is the bug. It should be used as a font name, which is actually the semantic given in iOS' implementation (and also is what the parameter name suggests). I do not follow why you would say that system fonts need to be used with default font name and size. What would be the default size? Why should we not use italic or bold fonts? 
Creating the font using NSFont rather than NSFontManager is analogous to what is done in iOS... so if iOS is right, macOS is wrong.

dumganhar:
@zilongshanren ,I found that `_createSystemFont` function in `CCDevice-ios.mm` & `CCDevice-mac.mm` was modified by you in https://github.com/cocos2d/cocos2d-x/pull/15827 .
Could you tell us why should  we use `[NSFontManager sharedFontManager]` for macOS while using `[UIFont fontWithName`  for iOS? Thanks.

zilongshanren:
@dumganhar I just do some refactoring in that PR, but I think the implementation is wrong here.
I have made a new PR to resolve this issue.

@kmaker  please test the PR to see whether it works or not, thanks.

kmaker:
works great! thanks!

zilongshanren:
@kmaker  😄   thanks for reporting this issue.

PR #16659, opened 2016-10-08
fix mac system font creation with NSFont

zilongshanren:
https://github.com/cocos2d/cocos2d-x/issues/16635


cocos/platform/ios/CCDevice-ios.mm
```diff
@@ -93,6 +93,11 @@ static CGSize _calculateShrinkedSizeForString(NSAttributedString **str, id font,
                actualSize.size.height > constrainSize.height) {
             fontSize = fontSize - 1;
 
+            if(fontSize < 0) {
```
zilongshanren, line 96: make code more robust in shrink mode.

kmaker (approved):
I've tested this PR and it works fine. Also like the fact that you are not hardcoding "Arial" as the default font, but letting the system to do so instead.

PR #15132, opened 2016-02-29
Remove unnecessary const from return types

mogemimi:
When compiling libcocos2d with GCC on Linux, it will always show warning messages below:

```
cocos/2d/CCAutoPolygon.h:104:39: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
cocos/2d/CCAutoPolygon.h:110:44: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
cocos/2d/CCAutoPolygon.h:113:67: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
cocos/2d/CCAutoPolygon.h:119:27: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
cocos/2d/CCSprite.h:477:33: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
cocos/renderer/CCGLProgram.h:492:38: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
...
```

In C/C++, const qualifiers on function return types are indeed superfluous. It will also trigger the warning when compiling with option `[-Wignored-qualifiers]`.
So this pull request removes unnecessary `const` from return types and fixes them.

Thank you for your time and attention.

PR #16671, opened 2016-10-10
Remove unnecessary const return qualifiers

mogemimi:
This PR gets rid of `const` qualifiers from return types which are copy and not reference, for the same reason PR #15132. Thanks.


minggo:
thanks

PR #16677, opened 2016-10-11
Lua: have cc.Ray:intersects additionally return the distance

Glidos:
cc.Ray:intersects now returns two results, an indication of hit or
not, plus the distance along the ray. The function can still be called
expecting a single argument and hence is backward compatible.

Also correct the debug error message

### DIFF
--- a/cocos/scripting/lua-bindings/manual/3d/lua_cocos2dx_3d_manual.cpp
+++ b/cocos/scripting/lua-bindings/manual/3d/lua_cocos2dx_3d_manual.cpp
@@ -2044,15 +2044,17 @@ int lua_cocos2dx_3d_Ray_intersects(lua_State* L)
         if (!ok)
             return 0;
 
-        bool ret = self->intersects(*arg0);
+        float distance;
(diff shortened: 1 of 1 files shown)


Glidos:
I couldn't find a sensible way to add a test for this. Only one of the current lua tests makes use of cc.Ray:intersects and in that case all the objects are at the same distance. I could add a whole new test, but it seems over the top for this little change.

minggo:
please fix the compiling error

```
/home/travis/build/cocos2d/cocos2d-x/cocos/scripting/lua-bindings/manual/3d/lua_cocos2dx_3d_manual.cpp:2048:38: error: ‘obb’ was not declared in this scope
         bool ret = self->intersects(*obb, &distance);
```

Glidos:
Oops! Sorry, slip made during a rebase. Fixed now.

Glidos:
Is this okay now, or is there more I need to do?

minggo:
thanks

Glidos:
Thanks @minggo

PR #16679, opened 2016-10-12
Fix bug when app paused the music is still resumed to play after interruption completed.

QingliWang:
Dear Mingoo,
please review this change for Tizen audio bug.
Thanks a lot.


QingliWang:
Hi, Mingoo, why it built failed? What's the error?

dumganhar:
It seems that Jenkins CI machines are all offline. :(

minggo:
@QingliWang yep, the jenkins is offline.

PR #16686, opened 2016-10-12
fixed cpp-tests ClickAndMoveTestCase bug

AbnerZhou:
The sprite (Grossini) need't move when touch position is the same as sprite's (Grossini) position.
And when "a" equal 0,  "atanf(o / a)" will have problem.

### DIFF
--- a/tests/cpp-tests/Classes/ClickAndMoveTest/ClickAndMoveTest.cpp
+++ b/tests/cpp-tests/Classes/ClickAndMoveTest/ClickAndMoveTest.cpp
@@ -48,19 +48,23 @@ void ClickAndMoveTestCase::onTouchEnded(Touch* touch, Event  *event)
     auto location = touch->getLocation();
 
     auto s = getChildByTag(kTagSprite);
-    s->stopAllActions();
-    s->runAction( MoveTo::create(1, Vec2(location.x, location.y) ) );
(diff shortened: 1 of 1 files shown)


tests/cpp-tests/Classes/ClickAndMoveTest/ClickAndMoveTest.cpp
```diff
@@ -48,19 +48,23 @@ void ClickAndMoveTestCase::onTouchEnded(Touch* touch, Event  *event)
...
     {
-        if(  o < 0 )
-            at = 180 + fabs(at);
-        else
-            at = 180 - fabs(at);    
+        s->stopAllActions();
+        s->runAction(MoveTo::create(1, Vec2(location.x, location.y)));
+        float o = location.x - s->getPosition().x;
+        float a = location.y - s->getPosition().y;
+        float at = (0.f == a) ? 90.f : (float)CC_RADIANS_TO_DEGREES(atanf(o / a));
```
minggo, line 58: Why it is 90 if `y` is not changed? I think it should be 0, because it will run along x axis.

minggo: any update?

PR #16688, opened 2016-10-12
bug fix label outline(int to float)

newnon:
fix label outline

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -1056,11 +1056,11 @@ void Label::enableGlow(const Color4B& glowColor)
     }
 }
 
-void Label::enableOutline(const Color4B& outlineColor,int outlineSize /* = -1 */)
+void Label::enableOutline(const Color4B& outlineColor,float outlineSize /* = -1 */)
 {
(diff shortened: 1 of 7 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -1056,11 +1056,11 @@ void Label::enableGlow(const Color4B& glowColor)
     }
 }
 
-void Label::enableOutline(const Color4B& outlineColor,int outlineSize /* = -1 */)
+void Label::enableOutline(const Color4B& outlineColor,float outlineSize /* = -1 */)
```
dumganhar, line 1059: Hi @newnon, did you get any reason that outlineSize should be `float` instead of `int`?

newnon:
because in FontFreeType field  float _outlineSize and it works well sometime you need float outline it's normal

```
FT_Stroker_Set(_stroker,
            (int)(_outlineSize * 64),
            FT_STROKER_LINECAP_ROUND,
            FT_STROKER_LINEJOIN_ROUND,
            0);
```

minggo:
How to reproduce the issue? Could you please add a test case and paste expected effect and current wrong effect?

minggo:
there is conflict in this PR, please send a new one, thanks

PR #16690, opened 2016-10-12
hide soft keys on keyboard hide

newnon:
fix android soft keys hide on keyboard hide

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
@@ -28,6 +28,7 @@ import android.content.Context;
 import android.content.res.Resources;
 import android.graphics.Color;
 import android.graphics.Typeface;
+import android.os.Handler;
 import android.os.Looper;
 import android.text.Editable;
 import android.text.InputType;
@@ -157,6 +158,15 @@ public class Cocos2dxEditBoxHelper {
 
                 editBox.setOnFocusChangeListener(new View.OnFocusChangeListener() {
 
+                	 private Handler mHandler = new Handler();
+                	 private Runnable decor_view_settings = new Runnable()
+                     {
+                         public void run() 
+                         {
+                         	mCocos2dxActivity.hideVirtualButton();
+                         }
+                     };
+                	
                     @Override
                     public void onFocusChange(View v, boolean hasFocus) {
                         if (hasFocus) {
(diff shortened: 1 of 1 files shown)


minggo:
what's the issue?

newnon:
On some devices soft keys does not hide after keyboard hide

newnon:
it some android bug http://stackoverflow.com/questions/21164836/immersive-mode-navigation-becomes-sticky-after-volume-press-or-minimise-restore/21278040#21278040

minggo:
How did you hide keyboard? Could you please describe 
- what did you do
- what's the wrong behavior
- what's the expected behavior

thanks

newnon:
On Android without hard keys. You need editbox click -> keyboard & soft keys will show click back keyboard will hide but soft keys is still here, not all devices have this bug

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
```diff
@@ -157,6 +158,15 @@ public void afterTextChanged(Editable s) {
 
                 editBox.setOnFocusChangeListener(new View.OnFocusChangeListener() {
 
+                	 private Handler mHandler = new Handler();
+                	 private Runnable decor_view_settings = new Runnable()
```
zilongshanren, line 162: the indentation is wrong here and the variable name `decor_view_settings` doesn't confirm to the coding style.
CameCase naming convention is used here.

zilongshanren:
@newnon  I found that `mCocos2dxActivity.hideVirtualButton();` is called right after the EditBox is invisible.      Why call it again in a delayed message handler?  How could I reproduce this issue?
I test on my Huawei P8 device without your patch and the virtual key is hidden correctly.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
```diff
@@ -172,6 +182,7 @@ public void run() {
                             Log.d(TAG, "edit box get focus");
                         } else {
                             editBox.setVisibility(View.GONE);
+                            mHandler.postDelayed(decor_view_settings, 500);
```
zilongshanren, line 185: there is another `mCocos2dxActivity.hideVirtualButton();` call in the following statements. Should we remove that one and make it called delayed across all the Android devices?

newnon:
I've got this bug only on nexus 7 with android 5.1.1 but may be it possible to get it on some other tablets.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
newnon: i'm not sure should you remove immediate hide, because it works on most android's and this fix is needed only on some devices. But it safe to call hideVirtualButton again

minggo:
i will close this PR since i can not reproduce it on many devices and it is a tricky to delay sometime to do something.

PR #16691, opened 2016-10-12
fix memory corruption

newnon:
fix possible memory corruption on cache clean

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -1068,6 +1068,8 @@ void Sprite::setSpriteFrame(SpriteFrame *spriteFrame)
     // update texture before updating texture rect
     if (texture != _texture)
     {
+        //fix possible memory corruption
+        _polyInfo.setTriangles(TrianglesCommand::Triangles{nullptr, nullptr, 0, 0});
         setTexture(texture);
     }
 


dumganhar:
Could you describe more about `cache clean`?
Some code snippets for reproducing `memory corruption` will be better. :)

newnon:
It is really hard to reproduce but set texture call set color that try to set memory that was deallocated with sprite frame (if you call clean sprite frame cache). 

```
if(spriteFrame->hasPolygonInfo())
{
    _polyInfo = spriteFrame->getPolygonInfo();
}
```

but if haven't we does not clean this field
may be it is better to fix it another way 

```
if(spriteFrame->hasPolygonInfo())
{
     _polyInfo = spriteFrame->getPolygonInfo();
}
else
{
    _polyInfo.setTriangles(TrianglesCommand::Triangles{nullptr, nullptr, 0, 0});
}
```

newnon:
no memory corruption is inside setTexture so we need clear triangles before

minggo:
I don't understand the issue too. Did you mean `_polyInfo` should have a default value or it will cause crash? If so, then 
- how to reproduce it
- it is better to initialize it in constructor

newnon:
no if you change spriteFrame on already initialized sprite and new one does not have poly info some time you can get memory corruption because inside setTexture method it writes some data inside polyInfo that can be already deallocated, it is hard to reproduce but it can heppen some time if you create Scale9Sprite with spriteframe from texture atlas and later change it to separate texture. Example default user avatar and then load real avatar from facebook ore something similar

minggo:
> no if you change spriteFrame on already initialized sprite and new one does not have poly info some time you can get memory corruption

If `_polyInfo` is initialized in in Sprite's construction,  then it will not happen.

minggo:
so i think it is better to add the codes into Sprite's constructor

``` c++
 _polyInfo.setTriangles(TrianglesCommand::Triangles{nullptr, nullptr, 0, 0});
```

newnon:
no bug only happens if you change spriteFrame after it construction
how to reproduce
1 create Scale9Sprite from SpriteFrame 
2 call clean sprite frame cache end texture cache (removeUnusedSpriteFrames & removeUnusedTextures)
3 set new spriteframe with full testure
4 repeat it many time and you'll get memory corruption
the problem is that inside setTexture it try to set some fields of _polyInfo but _polyInfo can be already deleted

minggo:
> the problem is that inside setTexture it try to set some fields of _polyInfo but _polyInfo can be already deleted

Sorry, i don't find the codes that `setTexture` modify _polyInfo.

newnon:
but it modify
setTexture()->updateBlendFunc()->setOpacityModifyRGB()->updateColor()
    for (ssize_t i = 0; i < _polyInfo.triangles.vertCount; i++) {
        _polyInfo.triangles.verts[i].colors = color4;
    }

minggo:
@newnon yep, i can see setTexture will modify _polyInfo, but i still don't understand why it will cause problem. 

When create a Sprite from a SpriteFrame, it will copy polyInfo from SpriteFrame like this

``` c++
void Sprite::setSpriteFrame(SpriteFrame *spriteFrame)
{
    ...
    if (spriteFrame->hasPolygonInfo())
    {
        _polyInfo = spriteFrame->getPolygonInfo();
    }
    ...
}
```

But PolygonInfo has an assign constructor like this

``` c++
PolygonInfo& PolygonInfo::operator= (const PolygonInfo& other)
{
    if(this != &other)
    {
        releaseVertsAndIndices();
        _filename = other._filename;
        _isVertsOwner = true;
        _rect = other._rect;
        triangles.verts = new (std::nothrow) V3F_C4B_T2F[other.triangles.vertCount];
        triangles.indices = new (std::nothrow) unsigned short[other.triangles.indexCount];
        CCASSERT(triangles.verts && triangles.indices, "not enough memory");
        triangles.vertCount = other.triangles.vertCount;
        triangles.indexCount = other.triangles.indexCount;
        memcpy(triangles.verts, other.triangles.verts, other.triangles.vertCount * sizeof(other.triangles.verts[0]));
        memcpy(triangles.indices, other.triangles.indices, other.triangles.indexCount * sizeof(other.triangles.indices[0]));
    }
    return *this;
}
```

So Sprite's polyInfo has its own memory, not shared some memory with other PolygonInfo Object, so i think there is not problem.

newnon:
Yes but 

```
void PolygonInfo::setTriangles(const TrianglesCommand::Triangles& other)
{
    this->releaseVertsAndIndices();
    isVertsOwner = false;

    this->triangles.vertCount = other.vertCount;
    this->triangles.indexCount = other.indexCount;
    this->triangles.verts = other.verts;
    this->triangles.indices = other.indices;
}
```

it's just store pointers and Scale9Sprite call this inside createSlicedSprites. May be it can cause some problems i can't reproduce this bug in one place, but i have real game inside which this bug reproduce 100%.
it happens than i have scale9sprite with sprite frame -> clean unused spriteFrames & textures - > set one texture as new image for this sprite

minggo:
I think the problem is caused by `PolygonInfo::setTriangles()`, it should keep its own memory, should not share the memory with others. 

This PR seems a little hack, and can not fix all the problem because this issue may happen in all places that access PolygonInfo's `verts`.

PR #16692, opened 2016-10-12
V3 scroll fixes

newnon:
Fixed scrolls behavior


minggo:
what's the issue?

newnon:
2 different problems
1. if you enable magnetic scroll and set scroll for example horizontally but height of element is bigger then height of container scroll will not work.
2. If you enable both scroll direction and inertial scroll, inertial scrolling will stop on any bound 
for example you have big width and small height of container after short vertical scroll it stop but it can continue scroll horizontally

minggo:
Could you please provide a test case?

PR #16694, opened 2016-10-12
Android pixel fixes

ricardoquesada:
Android Pixel fixes

### DIFF
--- a/cocos/navmesh/CCNavMesh.cpp
+++ b/cocos/navmesh/CCNavMesh.cpp
@@ -32,11 +32,12 @@
 
 NS_CC_BEGIN
 
+#pragma pack(push,1)
 struct TileCacheSetHeader
 {
-    int magic;
-    int version;
-    int numTiles;
+    int32_t magic;
+    int32_t version;
+    int32_t numTiles;
     dtNavMeshParams meshParams;
     dtTileCacheParams cacheParams;
 };
@@ -44,8 +45,9 @@ struct TileCacheSetHeader
 struct TileCacheTileHeader
 {
     dtCompressedTileRef tileRef;
-    int dataSize;
+    int32_t dataSize;
 };
+#pragma pack(pop)
 
 static unsigned char* parseRow(unsigned char* buf, unsigned char* bufEnd, char* row, int len)
 {
--- a/tests/cpp-tests/Resources/Shaders/example_Heart.fsh
+++ b/tests/cpp-tests/Resources/Shaders/example_Heart.fsh
@@ -9,7 +9,7 @@ uniform vec2 resolution;
 
 void main(void)
 {
-		float time = CC_Time[1];
+    float time = CC_Time[1];
     vec2 p = 2.0 * (gl_FragCoord.xy - center.xy) / resolution.xy;
 
     // animate
--- a/tests/cpp-tests/Resources/Shaders/example_Julia.fsh
+++ b/tests/cpp-tests/Resources/Shaders/example_Julia.fsh
@@ -9,7 +9,7 @@ uniform vec2 resolution;
 
 void main(void)
 {
-		float time = CC_Time[1];
+    float time = CC_Time[1];
     vec2 p = 2.0 * (gl_FragCoord.xy - center.xy) / resolution.xy;
     vec2 cc = vec2( cos(.25*time), sin(.25*time*1.423) );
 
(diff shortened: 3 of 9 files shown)


cocos/navmesh/CCNavMesh.cpp
```diff
@@ -32,20 +32,22 @@
...
 
+#pragma pack(push,1)
 struct TileCacheSetHeader
 {
-    int magic;
-    int version;
-    int numTiles;
+    int32_t magic;
+    int32_t version;
+    int32_t numTiles;
```
minggo, line 40: What's the issue?

ricardoquesada: `TileCacheSetHeader` is used to read the navMesh file. So its size must be exact.
two things I fixed:
- added `pragma pack(1)`
- replaced `int` with `int32_t`, since the size of `int` varies depending on the architeture

ricardoquesada: Just try to run the "NavMeshTes" on any android device, and it will crash while parsing the navmesh file.

minggo: I run `Navigation Mesh`, no crash. Did i run wrong test case?

ricardoquesada: nope, that was the test case. Well, perhaps it runs Ok in your devices. But it crashed in all my devices: Pixel 5, Pixel 6, Nexus 5x and Galaxy S4.
I used GCC 4.9, NDK 13 and build tools 25.

minggo: I have Nexus 5x, i will have a try.

ricardoquesada: ok. thanks! And if it works for you... then I don't know what was causing the crash... but I know that I fixed with that. In any case, we should always use `#pragma pack(push,0)` for file/network headers.

minggo: I used `cocos run -p android --ap android-24` to build apk with NDK 13, can run on Nexus 5x. I don't know the build tools version. How to set build tools version?

ricardoquesada: the build tools are in the SDK directory, but I can't remember if I really used build-tools 25, or the compiler chose the "default" one.
In any case, is there any reason for not applying this patch? do you think it could introduce bugs?

minggo: No, i just want to confirm the issue, because the codes used here is different from others.

minggo: And i think there are many places that read a file and parse it, such as tilemap xml file, cocos builder output file and so on. Should we check them too?

ricardoquesada: all binary file formats should be double-checked. TMX is a text format (xml), so there is no need to check it.
Could you run the Spine binary tests? They work for me on iOS and Mac, but there is one test that doesn't work Ok on Android. Could you test that one? I just want to confirm that it is not my toolchain that is generating errors on android. Thanks!

minggo: Sure, i will test it.

minggo: All binary test cases worked ok on my nexus 5x. I used NDK 13 with API level 13(the default API level of the test case).

ricardoquesada: Thanks for testing it. interesting... very interesting. One of the spine binary tests (the guy riding the dragon) is super accelerated. Something is different on my setup. Not sure what.

minggo: I will ask QA team to test it in more android phones to take a look.

minggo: The QA team tested on these Android devices, and worked ok too:
- xiaomi4 (Android 6.0)
- Oppo X9007(Android 4.3)
- HTC 710S(Android 2.3.7)
- Nexus 6(Android 5.1)
- sungsang GT N7100(Android 5.0.2)
- xiaomi2(Android 5.0.2)
- huawei honor6(Android 4.4.2)

All devices used the same APK built with NDK 13 with API level 13.

ricardoquesada: Thanks Minggo! At this point I don't know what I have different, but I do have something different.

PR #16695, opened 2016-10-13
[Don't merge]support Mutiview Matrix

songchengjiang:
In order to support GLES Mutiview extension

### DIFF
--- a/cocos/2d/CCScene.h
+++ b/cocos/2d/CCScene.h
@@ -109,7 +109,7 @@ public:
      * @param renderer The renderer use to render the scene.
      * @js NA
      */
-    virtual void render(Renderer* renderer, const Mat4& eyeTransform, const Mat4* eyeProjection = nullptr);
+    virtual void render(Renderer* renderer, const Mat4* eyeTransforms = nullptr, const Mat4* eyeProjections = nullptr, unsigned int mutiViewCount = 1);
     
     /** override function */
     virtual void removeAllChildren() override;
--- a/cocos/base/CCDirector.h
+++ b/cocos/base/CCDirector.h
@@ -455,15 +455,15 @@ public:
      * Clones a specified type matrix and put it to the top of specified type of matrix stack.
      * @js NA
      */
-    void pushMatrix(MATRIX_STACK_TYPE type);
+    void pushMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0);
     /** Pops the top matrix of the specified type of matrix stack.
      * @js NA
      */
-    void popMatrix(MATRIX_STACK_TYPE type);
+    void popMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0);
     /** Adds an identity matrix to the top of specified type of matrix stack.
      * @js NA
      */
-    void loadIdentityMatrix(MATRIX_STACK_TYPE type);
+    void loadIdentityMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0);
     /**
      * Adds a matrix to the top of specified type of matrix stack.
      * 
@@ -471,7 +471,7 @@ public:
      * @param mat The matrix that to be added.
      * @js NA
      */
-    void loadMatrix(MATRIX_STACK_TYPE type, const Mat4& mat);
+    void loadMatrix(MATRIX_STACK_TYPE type, const Mat4& mat, unsigned int index = 0);
     /**
      * Multiplies a matrix to the top of specified type of matrix stack.
      *
@@ -479,17 +479,18 @@ public:
      * @param mat The matrix that to be multiplied.
      * @js NA
      */
-    void multiplyMatrix(MATRIX_STACK_TYPE type, const Mat4& mat);
+    void multiplyMatrix(MATRIX_STACK_TYPE type, const Mat4& mat, unsigned int index = 0);
     /**
      * Gets the top matrix of specified type of matrix stack.
      * @js NA
      */
-    const Mat4& getMatrix(MATRIX_STACK_TYPE type) const;
+    const Mat4& getMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0) const;
     /**
      * Clear all types of matrix stack, and add identity matrix to these matrix stacks.
      * @js NA
      */
-    void resetMatrixStack();
+    void resetMatrixStack(unsigned int stackCount = 1);
+    unsigned int getMatrixStackCount(MATRIX_STACK_TYPE type) const;
 
     /**
      * returns the cocos2d thread id.
@@ -525,10 +526,10 @@ protected:
     void initTextureCache();
     void destroyTextureCache();
 
-    void initMatrixStack();
+    void initMatrixStack(unsigned int stackCount);
 
     std::stack<Mat4> _modelViewMatrixStack;
-    std::stack<Mat4> _projectionMatrixStack;
+    std::vector< std::stack<Mat4> > _projectionMatrixStackList;
     std::stack<Mat4> _textureMatrixStack;
 
     /** Scheduler associated with this director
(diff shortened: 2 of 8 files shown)


cocos/2d/CCScene.h
```diff
@@ -109,7 +109,7 @@ class CC_DLL Scene : public Node
      * @param renderer The renderer use to render the scene.
      * @js NA
      */
-    virtual void render(Renderer* renderer, const Mat4& eyeTransform, const Mat4* eyeProjection = nullptr);
+    virtual void render(Renderer* renderer, const Mat4* eyeTransforms = nullptr, const Mat4* eyeProjections = nullptr, unsigned int mutiViewCount = 1);
```
minggo, line 112: It is better to add a new function 

``` c++
virtual void render(Renderer* renderer, const Mat4* eyeTransforms = nullptr, const Mat4* eyeProjections = nullptr, unsigned int multiViewCount)
```

And there is an typo, `mutiViewCount` -> `multiViewCount`

cocos/base/CCDirector.h
```diff
@@ -455,41 +455,42 @@ class CC_DLL Director : public Ref
      * Clones a specified type matrix and put it to the top of specified type of matrix stack.
      * @js NA
      */
-    void pushMatrix(MATRIX_STACK_TYPE type);
+    void pushMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0);
```
minggo, line 458: It is better to add a new function.

cocos/base/CCDirector.h
```diff
@@ -455,41 +455,42 @@ class CC_DLL Director : public Ref
...
     /** Pops the top matrix of the specified type of matrix stack.
      * @js NA
      */
-    void popMatrix(MATRIX_STACK_TYPE type);
+    void popMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0);
```
minggo, line 462: It is better to add a new function.

cocos/base/CCDirector.h
```diff
@@ -455,41 +455,42 @@ class CC_DLL Director : public Ref
...
     /** Adds an identity matrix to the top of specified type of matrix stack.
      * @js NA
      */
-    void loadIdentityMatrix(MATRIX_STACK_TYPE type);
+    void loadIdentityMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0);
```
minggo, line 466: It is better to add a new function.

cocos/base/CCDirector.h
```diff
@@ -455,41 +455,42 @@ class CC_DLL Director : public Ref
...
     /**
      * Adds a matrix to the top of specified type of matrix stack.
      * 
      * @param type Matrix type.
      * @param mat The matrix that to be added.
      * @js NA
      */
-    void loadMatrix(MATRIX_STACK_TYPE type, const Mat4& mat);
+    void loadMatrix(MATRIX_STACK_TYPE type, const Mat4& mat, unsigned int index = 0);
```
minggo, line 474: It is better to add a new function.

cocos/base/CCDirector.h
```diff
@@ -455,41 +455,42 @@ class CC_DLL Director : public Ref
...
     /**
      * Multiplies a matrix to the top of specified type of matrix stack.
      *
      * @param type Matrix type.
      * @param mat The matrix that to be multiplied.
      * @js NA
      */
-    void multiplyMatrix(MATRIX_STACK_TYPE type, const Mat4& mat);
+    void multiplyMatrix(MATRIX_STACK_TYPE type, const Mat4& mat, unsigned int index = 0);
```
minggo, line 482: It is better to add a new function.

cocos/base/CCDirector.h
```diff
@@ -455,41 +455,42 @@ class CC_DLL Director : public Ref
...
     /**
      * Gets the top matrix of specified type of matrix stack.
      * @js NA
      */
-    const Mat4& getMatrix(MATRIX_STACK_TYPE type) const;
+    const Mat4& getMatrix(MATRIX_STACK_TYPE type, unsigned int index = 0) const;
```
minggo, line 487: It is better to add a new function.

cocos/base/CCDirector.h
```diff
@@ -455,41 +455,42 @@ class CC_DLL Director : public Ref
...
     /**
      * Clear all types of matrix stack, and add identity matrix to these matrix stacks.
      * @js NA
      */
-    void resetMatrixStack();
+    void resetMatrixStack(unsigned int stackCount = 1);
```
minggo, line 492: It is better to add a new function.

cocos/base/CCDirector.h
```diff
@@ -525,10 +526,10 @@ class CC_DLL Director : public Ref
     void initTextureCache();
     void destroyTextureCache();
 
-    void initMatrixStack();
+    void initMatrixStack(unsigned int stackCount);
```
minggo, line 529: It is better to add a new function.

cocos/renderer/CCGLProgram.cpp
```diff
@@ -116,8 +116,10 @@ const char* GLProgram::SHADER_CAMERA_CLEAR = "ShaderCameraClear";
 // uniform names
 const char* GLProgram::UNIFORM_NAME_AMBIENT_COLOR = "CC_AmbientColor";
 const char* GLProgram::UNIFORM_NAME_P_MATRIX = "CC_PMatrix";
+const char* GLProgram::UNIFORM_NAME_MUTIVIEW_P_MATRIX = "CC_MutiViewPMatrix";
```
minggo, line 119: a typo
CC_MutiViewPMatrix -> CC_MultiViewPMatrix

minggo: UNIFORM_NAME_MUTIVIEW_P_MATRIX -> UNIFORM_NAME_MULTIVIEW_P_MATRIX

cocos/renderer/CCGLProgram.cpp
```diff
@@ -116,8 +116,10 @@ const char* GLProgram::SHADER_CAMERA_CLEAR = "ShaderCameraClear";
...
 const char* GLProgram::UNIFORM_NAME_MV_MATRIX = "CC_MVMatrix";
 const char* GLProgram::UNIFORM_NAME_MVP_MATRIX  = "CC_MVPMatrix";
+const char* GLProgram::UNIFORM_NAME_MUTIVIEW_MVP_MATRIX  = "CC_MutiViewMVPMatrix";
```
minggo, line 122: CC_MutiViewMVPMatrix -> CC_MultiViewMVPMatrix
UNIFORM_NAME_MUTIVIEW_MVP_MATRIX -> UNIFORM_NAME_MULTIVIEW_MVP_MATRIX

minggo:
@ricardoquesada this PR is added to support ARM Mali VR, because ARM Mali VR depends on multiple views. Could you please take a look if the modification is suitable or not without detail inspect. @songchengjiang will continue to polish the PR(such as fix typo, add new functions instead of use default parameter) if the modification is ok.
Thanks.

ricardoquesada:
The PR good looks to me. My only concern is the "multi views" parameter. What's the maximum number of views ? 
If the answer is two, like "left eye" and right "eye", I would make it more explicit , like "left eye" stack, and "right eye" stack.

songchengjiang:
@ricardoquesada As ARM said, The scene is rendered twice for each eye position, once with a wide field of view, and once with a narrower field of view in order to render the midpoint of the screen with a higher resolution than the rest.

ricardoquesada:
@songchengjiang ok. adding "explicit" names for the stacks, or better documentation is needed.

Otherwise developers (including myself) won't understand why we have multiple Transform stacks.

But this code won't work correctly in legacy code (like the Box2d debug code) that calls `GLProgram::setUniformsForBuiltins();`
It will "work", but not correctly, since it will always use the stack number 0.

If that is not important, then probably we shouldn't have multiple stacks.

minggo:
@ricardoquesada i am not sure if it is possible to add `explicit` names for stacks. Yep, the meaning of ARM VR is clear, but is there any other VR that needs multiple views too but have different meanings with each view? So i think have good documents is better.

PR #16689, opened 2016-10-12
fix font rendering position

newnon:
fix font always padded on 2 pixels left

PR #16703, opened 2016-10-14
Adds test case for PR #16689 (fix font rendering position)

dumganhar:
Adds test case for PR #16689 (fix font rendering position)


dumganhar:
@minggo, I tested this patch and added test case. It's good for merge.

PR #16707, opened 2016-10-16
C API

colemancda:
This is still WIP, but I wanted to bring it to the devs attention so I know my work won't be in vain. Have a C API is very important for interoperability with other languages (like Swift). You can checkout my Swift (no ObjC) layer for Cocos2D-X.

[Swift Repo](https://github.com/PureSwift/Cocos2D)
[Forum Discussion](http://discuss.cocos2d-x.org/t/discussion-adding-open-source-swift-to-cocos2d-x/21979/10)


minggo:
thanks, but we don't have plan to provide C API

PR #16711, opened 2016-10-17
Mesh* Sprite3D::getMesh() should return nullptr when _meshes.empty() is true 

zloop1982:
So we check _meshes.empty() before return.


minggo:
Yep, you are right.
Thanks.

PR #16712, opened 2016-10-17
Fix typos in local variables

mogemimi:
This pull request fixes some minor typos in local variables. It also doesn't break backward compatibility. Thanks!


minggo:
thanks

PR #16716, opened 2016-10-18
Performance tweak: Use range-based for-loops and allocate std::vector size() and *end() on the stack where favorable.

Rubonnek:
Basically this is the same as #16642, but on a bigger scale. All the files changed in this pull request were identified by using the following commands in Linux:

```
$ cd "${COCOS_GIT_ROOT}"/cocos
$ find . -type f -exec grep -Iq . {} \; -and -print0 | xargs -0 grep -P "^\s*for\s*\(.*end\s*\(|^\s*for\s*\(.*size\s*\(|^\s*while\s*\(.*end\s*\(|^\s*while\s*\(.*size\s*\("
```

The only files that I didn't take a look at were those under the `cocos/editor-support`, `cocos/deprecated` and `cocos/scripting` directories, but they can be identified by using the above commands.

### DIFF
--- a/cocos/2d/CCActionCatmullRom.cpp
+++ b/cocos/2d/CCActionCatmullRom.cpp
@@ -67,10 +67,9 @@ bool PointArray::initWithCapacity(ssize_t capacity)
 PointArray* PointArray::clone() const
 {
     vector<Vec2*> *newArray = new (std::nothrow) vector<Vec2*>();
-    vector<Vec2*>::iterator iter;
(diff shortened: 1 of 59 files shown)


Rubonnek:
[ci rebuild]

minggo:
Looks good to me, thanks. It is a good PR, thanks.

Rubonnek:
Thanks a lot. I appreciate it.

PR #16717, opened 2016-10-18
Fixes label wrong blending and improves performance while enabling outline.

dumganhar:
The old outline shader will cause wrong blending while enabling outline for label.
You could refer to the following screenshot to compare.

The old label display:

old

The label after this fix:
after-fix

Please notice the blue color and its alpha depth.

The Reproduced Code:

``` c++
    auto label = Label::createWithTTF("Hello World", "fonts/arial.ttf", 70);

    // position the label on the center of the screen
    label->setPosition(origin.x + visibleSize.width/2,
                            origin.y + visibleSize.height - label->getContentSize().height);

    label->enableOutline(Color4B(0, 255, 0, 100), 10); // Set 100 alpha for outline
    label->setTextColor(Color4B(0, 0, 255, 100)); // Also set 100 alpha for text
    // add the label as a child to this layer
    this->addChild(label, 1);
```

For more easily to find this blending problem, you could set both outline and text color to green, and see the difference.

The old outline shader:

screen shot 2016-10-18 at 11 57 23

The new outline shader:

screen shot 2016-10-18 at 12 00 22

### DIFF
--- a/cocos/renderer/ccShader_Label_outline.frag
+++ b/cocos/renderer/ccShader_Label_outline.frag
@@ -15,13 +15,28 @@ uniform vec4 u_textColor;
 void main()
 {
     vec4 sample = texture2D(CC_Texture0, v_texCoord);
-    float fontAlpha = sample.a; 
-    float outlineAlpha = sample.r; 
-    if ((fontAlpha + outlineAlpha) > 0.0){
-        vec4 color = u_textColor * fontAlpha + u_effectColor * (1.0 - fontAlpha);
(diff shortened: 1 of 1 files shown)


dumganhar:
Performance Test Result (Label with outline width 10):
screen shot 2016-10-18 at 15 48 44

@ricardoquesada , @minggo ,please review.

ricardoquesada:
Interesting... I thought that `discard` should be avoided because I made things slower.
But if this fixes a bug, great.

PR #16723, opened 2016-10-20
Support MultiView matrix array

songchengjiang:
Support MultiView matrix array and add new interface in GLProgram to support shader header definitions.

### DIFF
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -199,13 +199,13 @@ void Scene::render(Renderer* renderer, const Mat4* eyeTransforms, const Mat4* ey
     {
         if (!camera->isVisible())
             continue;
-        
+
         Camera::_visitingCamera = camera;
         if (Camera::_visitingCamera->getCameraFlag() == CameraFlag::DEFAULT)
         {
             defaultCamera = Camera::_visitingCamera;
         }
-   
+
         // There are two ways to modify the "default camera" with the eye Transform:
         // a) modify the "nodeToParentTransform" matrix
         // b) modify the "additional transform" matrix
@@ -242,7 +242,7 @@ void Scene::render(Renderer* renderer, const Mat4* eyeTransforms, const Mat4* ey
 
         // we shouldn't restore the transform matrix since it could be used
         // from "update" or other parts of the game to calculate culling or something else.
-        //        camera->setNodeToParentTransform(eyeCopy);
+//        camera->setNodeToParentTransform(eyeCopy);
     }
 
 #if CC_USE_3D_PHYSICS && CC_ENABLE_BULLET_INTEGRATION
@@ -267,7 +267,7 @@ void Scene::render(Renderer* renderer, const Mat4* eyeTransforms, const Mat4* ey
 #endif
 
     Camera::_visitingCamera = nullptr;
-    //    experimental::FrameBuffer::applyDefaultFBO();
+//    experimental::FrameBuffer::applyDefaultFBO();
 }
 
 void Scene::removeAllChildren()
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -186,69 +186,7 @@ const std::vector<Camera*>& Scene::getCameras()
 
 void Scene::render(Renderer* renderer, const Mat4& eyeTransform, const Mat4* eyeProjection)
 {
-    auto director = Director::getInstance();
-    Camera* defaultCamera = nullptr;
-    const auto& transform = getNodeToParentTransform();
-
-    for (const auto& camera : getCameras())
-    {
-        if (!camera->isVisible())
-            continue;
-
-        Camera::_visitingCamera = camera;
-        if (Camera::_visitingCamera->getCameraFlag() == CameraFlag::DEFAULT)
-        {
-            defaultCamera = Camera::_visitingCamera;
-        }
-
-        // There are two ways to modify the "default camera" with the eye Transform:
-        // a) modify the "nodeToParentTransform" matrix
-        // b) modify the "additional transform" matrix
-        // both alternatives are correct, if the user manually modifies the camera with a camera->setPosition()
-        // then the "nodeToParent transform" will be lost.
-        // And it is important that the change is "permanent", because the matrix might be used for calculate
-        // culling and other stuff.
-        if (eyeProjection)
-            camera->setAdditionalProjection(*eyeProjection * camera->getProjectionMatrix().getInversed());
-        camera->setAdditionalTransform(eyeTransform.getInversed());
-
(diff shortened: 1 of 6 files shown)


songchengjiang:
@minggo Please review.

cocos/2d/CCScene.h
```diff
@@ -110,6 +110,7 @@ class CC_DLL Scene : public Node
      * @js NA
      */
     virtual void render(Renderer* renderer, const Mat4& eyeTransform, const Mat4* eyeProjection = nullptr);
+    virtual void render(Renderer* renderer, const Mat4* eyeTransforms, const Mat4* eyeProjections, unsigned int multiViewCount);
```
minggo, line 123: When and where is this function be used?

minggo: Can re-use the codes of `virtual void render(Renderer* renderer, const Mat4& eyeTransform, const Mat4* eyeProjection = nullptr);`?

cocos/base/CCDirector.cpp
```diff
@@ -466,6 +465,48 @@ void Director::resetMatrixStack()
     initMatrixStack();
 }
 
+void Director::resetMatrixStack(unsigned int stackCount)
```
minggo: Why need stackCount? I think reset will reset all matrix stacks.

cocos/base/CCDirector.cpp
```diff
@@ -474,7 +515,27 @@ void Director::popMatrix(MATRIX_STACK_TYPE type)
...
+    {
+        _textureMatrixStack.pop();
+    }
+    else
+    {
+        CCASSERT(false, "unknown matrix stack type");
+    }
+}
+
+void Director::popMatrix(MATRIX_STACK_TYPE type, unsigned int index)
```
minggo: `_modelViewMatrixStack` and  `_textureMatrixStack` is not need to support this function, so i think it is better to rename it to `popProjectionMatrix(unsigned index)`. The same opinion with `loadIdentityMatrix` and other functions.

cocos/renderer/CCGLProgram.cpp
```diff
@@ -236,50 +249,55 @@ bool GLProgram::initWithByteArrays(const GLchar* vShaderByteArray, const GLchar*
...
     }
     CHECK_GL_ERROR_DEBUG();
-
+    
     if (_fragShader)
     {
         glAttachShader(_program, _fragShader);
     }
-
+
```
minggo: pease fix indent issue, remove unneeded spaces, there are many places have this issue

cocos/renderer/CCGLProgram.h
```diff
@@ -351,6 +361,8 @@ class CC_DLL GLProgram : public Ref
     bool initWithByteArrays(const GLchar* vShaderByteArray, const GLchar* fShaderByteArray);
     static GLProgram* createWithByteArrays(const GLchar* vShaderByteArray, const GLchar* fShaderByteArray, const std::string& compileTimeDefines);
     bool initWithByteArrays(const GLchar* vShaderByteArray, const GLchar* fShaderByteArray, const std::string& compileTimeDefines);
+    static GLProgram* createWithByteArrays(const GLchar* vShaderByteArray, const GLchar* fShaderByteArray, const std::string& compileTimeHeaders, const std::string& compileTimeDefines);
```
minggo: What's `compileTimeHeaders`?

cocos/base/CCDirector.cpp
songchengjiang: because we don't know the size of stack, maybe 4 or 2

cocos/2d/CCScene.h
songchengjiang: this funtion will call only in MaliVR render for now.
It's hard to re-use, because the code is hard to break up

cocos/renderer/CCGLProgram.cpp
songchengjiang: ok

cocos/base/CCDirector.cpp
songchengjiang: ok

cocos/renderer/CCGLProgram.h
songchengjiang: In order to define the header of shader.

songchengjiang: eg. For mailvr, we need define #version 300 es in **first line** of shader

songchengjiang: eg. For malivr, we need define #version 300 es in first line of shader

songchengjiang: eg. For malivr, we need define #version 300 es in **first line** of shader

cocos/base/CCDirector.cpp
minggo: the vector has the size information

cocos/renderer/CCGLProgram.h
minggo: Isn't `define #version 300` part of `compileTimeDefines`?

minggo: I mean can not use `compileTimeDefines` since it just define something.

cocos/2d/CCScene.cpp
```diff
@@ -239,11 +248,21 @@ void Scene::render(Renderer* renderer, const Mat4& eyeTransform, const Mat4* eye
...
+            director->pushProjectionMatrix(i);
+            director->loadProjectionMatrix(physics3dDebugCamera->getViewProjectionMatrix(), i);
+        }
+
         _physics3DWorld->debugDraw(renderer);
         renderer->render();
-        director->popMatrix(MATRIX_STACK_TYPE::MATRIX_STACK_PROJECTION);
+
+        for (unsigned int i = 0; i < multiViewCount; ++i)
+            director->popProjectionMatrix(i);
```
minggo, line 265: can we add `Director::resetProjectionMatrixStack()` to simplify the usage?

songchengjiang:
@minggo All is done!

cocos/base/CCDirector.cpp
```diff
@@ -534,7 +562,7 @@ void Director::multiplyMatrix(MATRIX_STACK_TYPE type, const Mat4& mat)
     }
     else if(MATRIX_STACK_TYPE::MATRIX_STACK_PROJECTION == type)
     {
-        _projectionMatrixStack.top() *= mat;
+        _projectionMatrixStackList[0].top() *= mat;
```
minggo, line 565: I think we should add comment in `Director::multiplyMatrix` that describe only `_projectionMatrixStackList[0]` is affected.

minggo: the same comment should be added in other functions, such as `Director::loadMatrix(MATRIX_STACK_TYPE type, const Mat4& mat)`.

PR #16724, opened 2016-10-20
[Don't merge]VRSDK modification list

songchengjiang:


songchengjiang:
@minggo  
Generate patch command on Mac: `$ diff -uN SOURCE_FILE_NAME MODIFIED_FILE_NAME > GENERATED_FILE_NAME`
eg. 
`$ diff -uN AppDelegate.cpp AppDelegate2.cpp > AppDelegate.cpp.js.3.13.patch`

PR #16727, opened 2016-10-20
Fix android-build.py

apoluektov:
Fixes various problems with android-build.py:
- allows to specify android API level (to propagate to '--ap' flag of cocos console tool)
- fixes help messages
- replaces deprecated OptionParser with ArgumentParser
- simplifies the code

### DIFF
--- a/build/android-build.py
+++ b/build/android-build.py
@@ -11,10 +11,12 @@ LUA_SAMPLES = ['lua-empty-test', 'lua-tests', 'lua-game-controller-test']
 JS_SAMPLES = ['js-tests']
 ALL_SAMPLES = CPP_SAMPLES + LUA_SAMPLES + JS_SAMPLES
 
-def caculate_built_samples(args):
-    ''' Compute the sampels to be built
-    'cpp' for short of all cpp tests
-    'lua' for short of all lua tests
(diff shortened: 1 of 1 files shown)


build/android-build.py
```diff
@@ -1,120 +1,100 @@
-#!/usr/bin/python
+#!/usr/bin/env python
```
minggo, line 1: Why changed like this? I found python is installed in `usr/bin/python` in my machine.

```
A03101308207:~ minggo$ which python
/usr/bin/python
```

apoluektov: Using shebang like this ensures that the `python` that is actually on PATH is used.

As an example, it is a common way for macOS users to install python using homebrew, and homebrew doesn't put files to system directories.

minggo: But if don't use homebrew as me, the script will not work?

apoluektov: It will work.

`/usr/bin/env python` will find the python interpreter that is on your `PATH`.
On your machine it will find `/usr/bin/python`, unless you have installed another version in i.e. `/usr/local/bin`.

http://stackoverflow.com/a/2429517/104451

minggo:
thanks

PR #16734, opened 2016-10-24
fix app crash when dimensions too large

ghost:
app will crash when video dimensions too large


zilongshanren:
@VindroidH 
Thanks, how could I reproduce this issue?

ghost:
@zilongshanren 
You can create a videoplayer then use **videoplayer->setScale()** to reproduce this issue.

zilongshanren:
@VindroidH 
Thanks, I will give it a test.

minggo:
I add `_videoPlayer->setScale(1.5);` in `UIVideoTest.cpp`, not crash on huawei p9.

PR #16736, opened 2016-10-24
fixed #16735: [native] The behavior of (setRotation + setSkewX) is wrong

### LINKED ISSUE #16735: [native] The behavior of (setRotation + setSkewX) is wrong
- cocos2d-x version: 3.13.1
- devices test on: all platforms
- developing environments
  - NDK version: r10e
  - Xcode version: 8.0
  - VS version: 2015

Steps to Reproduce:
1. Create a sprite
2. setRotation(30) and setSkewX(30)

Please refer to the last reply in https://github.com/cocos2d/cocos2d-x/pull/4543.

dumganhar:
Since I removed some logic, this patch may also improves a little bit performance.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1672,21 +1672,11 @@ const Mat4& Node::getNodeToParentTransform() const
         }
         
         bool needsSkewMatrix = ( _skewX || _skewY );
-        
-        
-        Vec2 anchorPoint(_anchorPointInPoints.x * _scaleX, _anchorPointInPoints.y * _scaleY);
-        
-        // calculate real position
-        if (! needsSkewMatrix && !_anchorPointInPoints.isZero())
-        {
-            x += -anchorPoint.x;
-            y += -anchorPoint.y;
-        }
-        
+
         // Build Transform Matrix = translation * rotation * scale
         Mat4 translation;
         //move to anchor point first, then rotate
-        Mat4::createTranslation(x + anchorPoint.x, y + anchorPoint.y, z, &translation);
+        Mat4::createTranslation(x, y, z, &translation);
         
         Mat4::createRotation(_rotationQuat, &_transform);
         
@@ -1707,10 +1697,7 @@ const Mat4& Node::getNodeToParentTransform() const
             _transform.m[1] = sy * m0 + cx * m1, _transform.m[5] = sy * m4 + cx * m5, _transform.m[9] = sy * m8 + cx * m9;
         }
         _transform = translation * _transform;
-        //move by (-anchorPoint.x, -anchorPoint.y, 0) after rotation
-        _transform.translate(-anchorPoint.x, -anchorPoint.y, 0);
-        
-        
+
         if (_scaleX != 1.f)
         {
             _transform.m[0] *= _scaleX, _transform.m[1] *= _scaleX, _transform.m[2] *= _scaleX;
@@ -1738,15 +1725,15 @@ const Mat4& Node::getNodeToParentTransform() const
             Mat4 skewMatrix(skewMatArray);
             
             _transform = _transform * skewMatrix;
-            
-            // adjust anchor point
-            if (!_anchorPointInPoints.isZero())
-            {
-                // FIXME:: Argh, Mat4 needs a "translate" method.
-                // FIXME:: Although this is faster than multiplying a vec4 * mat4
-                _transform.m[12] += _transform.m[0] * -_anchorPointInPoints.x + _transform.m[4] * -_anchorPointInPoints.y;
-                _transform.m[13] += _transform.m[1] * -_anchorPointInPoints.x + _transform.m[5] * -_anchorPointInPoints.y;
-            }
+        }
+
+        // adjust anchor point
+        if (!_anchorPointInPoints.isZero())
+        {
+            // FIXME:: Argh, Mat4 needs a "translate" method.
+            // FIXME:: Although this is faster than multiplying a vec4 * mat4
+            _transform.m[12] += _transform.m[0] * -_anchorPointInPoints.x + _transform.m[4] * -_anchorPointInPoints.y;
+            _transform.m[13] += _transform.m[1] * -_anchorPointInPoints.x + _transform.m[5] * -_anchorPointInPoints.y;
         }
     }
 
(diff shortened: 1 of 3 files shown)


dumganhar:
@ricardoquesada , @minggo , please review this PR. Thanks.

cocos/2d/CCNode.cpp
```diff
@@ -1707,10 +1697,7 @@ const Mat4& Node::getNodeToParentTransform() const
             _transform.m[1] = sy * m0 + cx * m1, _transform.m[5] = sy * m4 + cx * m5, _transform.m[9] = sy * m8 + cx * m9;
         }
         _transform = translation * _transform;
-        //move by (-anchorPoint.x, -anchorPoint.y, 0) after rotation
-        _transform.translate(-anchorPoint.x, -anchorPoint.y, 0);
```
dumganhar, line 1711: Doesn't need to translate here, save Mat4*Mat4 (16 multiply) operations while enabling skew.

cocos/2d/CCNode.cpp
```diff
@@ -1672,21 +1672,11 @@ const Mat4& Node::getNodeToParentTransform() const
         }
         
         bool needsSkewMatrix = ( _skewX || _skewY );
-        
-        
-        Vec2 anchorPoint(_anchorPointInPoints.x * _scaleX, _anchorPointInPoints.y * _scaleY);
```
dumganhar, line 1677: Save 2 multiply operations.

cocos/2d/CCNode.cpp
```diff
@@ -1738,15 +1725,15 @@ const Mat4& Node::getNodeToParentTransform() const
...
-            }
+        }
+
+        // adjust anchor point
+        if (!_anchorPointInPoints.isZero())
+        {
+            // FIXME:: Argh, Mat4 needs a "translate" method.
+            // FIXME:: Although this is faster than multiplying a vec4 * mat4
+            _transform.m[12] += _transform.m[0] * -_anchorPointInPoints.x + _transform.m[4] * -_anchorPointInPoints.y;
+            _transform.m[13] += _transform.m[1] * -_anchorPointInPoints.x + _transform.m[5] * -_anchorPointInPoints.y;
```
dumganhar, line 1736: Just move the code within `skew logic`.

dumganhar:
[ci rebuild]

dumganhar:
[ci rebuild]

ricardoquesada:
the code looks good to me. but I think the "real test" is to make sure that all the "sprite tests" from cpp-tests pass.

ricardoquesada:
I would like to add that sometimes a bug becomes a feature. What I mean is that if we broke the "skew" functionality in v3.0, and we fixe it in v3.14, it means that games that are using skew (with the bug), might not want to migrate to v3.14 since it has a different behavior (perhaps the correct one, but for existing users it will be a "bug").

dumganhar:
Indeed, I did a test by using `cpp-tests` and I don't find anything I break until now.

> I would like to add that sometimes a bug becomes a feature. What I mean is that if we broke the "skew" functionality in v3.0, and we fixe it in v3.14, it means that games that are using skew (with the bug), might not want to migrate to v3.14 since it has a different behavior (perhaps the correct one, but for existing users it will be a "bug").

That should be right sometimes.
But this fix is very very special. 
- Imagine that developer wanna to rotate and skew a sprite which have an anchor point (0.5, 0.5), the result in 3.x will be that the sprite isn't in the center. I guess if developer uses it like this, he/she will get an unexpected or really strange result. And the result is useless since there isn't any rules for this wrong behavior could follow.
- If JS developers use `rotate + skew` functionality, it will work perfectly on cocos2d-html(web) while the display on native will hurt them. It's different on native & web platform. Then, developers will blame that `Why are they different?` In fact, this issue was submitted by one of our developers. They were really confused why the behavior of rotate + skew was different and this wasted them **nearly two weeks** for just trying to check it out before I helped.
- If developers don't use `rotate + skew` at the same time, nothing goes wrong. Therefore, this patch is only for  `rotate + skew` usage which isn't widely used. I guess it's safe to merge this patch.
- And with this fix, the logic of `node to parent transform` is clearer than before, plus, we get a little performance improved. :)

I hope this patch could be merged into upstream. Yep, more test work could be done by our QA team before we release 3.14.

What do you think? @ricardoquesada

dumganhar:
@ricardoquesada, any update? :)

ricardoquesada:
@dumganhar sorry for the delay.

It seems a small use case (skew + rotate). So, I'm good with the PR. Thanks. feel free to merge it.

dumganhar:
@ricardoquesada, thanks.

dumganhar:
@minggo , please check it out again, if no problem, please merge. Thanks.

minggo:
I have a question about the matrix, in theory, the matrix should be

```
translate(position) * translate(anchor) * rotation * scale * translate(-anchor)
```

now we use `_ignoreAnchorPointForPosition` to determine if we need to do `translate(anchor)`. 
Am i right?

minggo:
@ricardoquesada current logical equals

``` c++
if (_ignoreAnchorPointForPosition)
   matrix = translate(position) * translate(anchor) * rotation * scale * translate(-anchor)
else
  matrix = translate(position) * rotation * scale * translate(-anchor)
```

I don't quite understand the meaning of `_ignoreAnchorPointForPosition`.

dumganhar:
My thought is that:
`_ignoreAnchorPointForPosition` just says whether to translate `anchorPoint` length at the end of node transform.
If it's false, we just move the node from (anchor point **position** (0, 0) ) to the (pos: (x, y)).
If it's true, move the node from (anchor point **position**( 0, 0) ) to `anchor point length, e.g (0.5, 0.5)`, this will make the bottom left corner to origin(0, 0), then move the node to pos (x, y).

Therefore, `ignore anchor point for position`  means that we don't position a node by an anchor point, `don't ignore` means moving a node from its initialized position (bottom left is 0, 0) to pos (x, y).

@ricardoquesada, if I'm wrong , please correct me. :)

minggo:
I know the final effect, i just wonder the parameter name. `_ignoreAnchorPointForPosition` is hard to understand its meaning, at least to me.

dumganhar:
@minggo , yes, it's a little confused.
Perhaps, it should be named like `_isPositionByAnchorPoint`, if it's true, position a node from anchor point (0, 0) to position (x, y).
If it's false, position a node from original point (bottom left) to position (x, y).

ricardoquesada:
@minggo @dumganhar 
`ignoreAnchorForPosition` was added only to make `Layer` look Ok by default.

It has anchorPoint value of 0.5 this menas that:
- it can be rotated by its center
- but its position is 0,0
- and everything looks Ok.

In hindsight, I should have never added it. I imagined that that was a valid use case. But it is not and adds unneeded complexity to the code.

In v4.0, I would remove it.

minggo:
Good to know. Thanks.

PR #16742, opened 2016-10-24
Removes unused 'return' letter in builtin shaders.

dumganhar:


dumganhar:
[ci rebuild]

dumganhar:
[ci rebuild]

stevetranby:
@Rubonnek & @dumganhar - maybe consider using the c++11 multiline string literal? 

``` cpp
const char* cc3D_Color_frag = R"(

#ifdef GL_ES
varying lowp vec4 DestinationColor;
#else
varying vec4 DestinationColor;
#endif
uniform vec4 u_color;

void main(void)
{
    gl_FragColor = u_color;
}
)";
```

Rubonnek:
@stevetranby Thanks, that works. I updated my pull request.

dumganhar:
@stevetranby, it seems to be another choice to make a multi-line string.
Currently, we use `#define STRINGIFY(A)  #A`, it's also easy to use.
What benefit would `R"()";` take compiled to `STRINGIFY` macro?

Rubonnek:
I just noticed R"()" takes more memory when there are more newlines between the text. Here's how I tested it:

``` cpp
#include <iostream>
using namespace std;

#define STRINGIFY(A) #A

int main ()
{
    const char* multilined = R"(
    #ifdef GL_ES
        varying lowp vec4 DestinationColor;
    #else
        varying vec4 DestinationColor;
    #endif
        uniform vec4 u_color;







    void main(void)
    {
        gl_FragColor = u_color;
    }
    )";

    const char* stringified = STRINGIFY(
    \n#ifdef GL_ES\n
        varying lowp vec4 DestinationColor;
    \n#else\n
        varying vec4 DestinationColor;
    \n#endif\n
        uniform vec4 u_color;







    void main(void)
    {
        gl_FragColor = u_color;
    }
    );


    // Count the characters
    int count = 0;
    while(multilined[count] != '\0')
    {
        ++count;
    }
    cout << "Multiine string character count: " << count << endl;
    cout << "========" << endl;
    cout << multilined << endl << endl;
    cout << "========" << endl << endl << endl;


    count = 0;
    while(stringified[count] != '\0')
    {
        ++count;
    }
    cout << "STRINGIFY character count: " << count << endl;
    cout << "========" << endl;
    cout << stringified << endl << endl;
    cout << "========" << endl;
    return 0;
}
```

This will output:

```
C++ multiline R"()" character count: 185
========

        #ifdef GL_ES
                varying lowp vec4 DestinationColor;
        #else
                varying vec4 DestinationColor;
        #endif
                uniform vec4 u_color;







        void main(void)
        {
            gl_FragColor = u_color;
        }

========


STRINGIFY character count: 164
========

#ifdef GL_ES
 varying lowp vec4 DestinationColor;
#else
 varying vec4 DestinationColor;
#endif
 uniform vec4 u_color; void main(void) { gl_FragColor = u_color; }
========
```

I think we should stay with `STRINGIFY` instead. The `\n` character will be a bother, but at least we can add more whitespace to the code without sacrificing memory.

stevetranby:
@Rubonnek in my mind it's about correctness. The R"" literal shader code will always compile unless there's a known issue whereas the Stringify(X) macro version obviously has unforeseen issues. I don't foresee these shaders changing very often so it's probably of no real concern for cocos2d-x 3.x.

As for your concerns I would argue two points. The first is that the shader will get compiled down to the same binary size in GPU memory and regardless the extra storage/memory is negligible compared to textures and audio. Also I wouldn't indent them, add excess empty lines, nor excess comments to the final shader code that goes into these engine shader files.

However, it does allow for comments (even if they go against your concerns) without having to remember to put in a '\n' escape which may be helpful when used sparingly.

Anyway, my 2c

Rubonnek:
@stevetranby I understand and I agree. For the reasons you mentioned it's better to keep string literal instead. I'd just opt for saving memory in this case even if it's negligible.

In the end I don't mind which implementation gets chosen as long as the issue with the Windows build gets fixed. After working in the pull request I noticed a very simple Perl script can switch the `STRINGIFY` macro and the `R"()"` string literal in these files -- I can just write that and keep the files with the `STRINGIFY` macro to myself.

minggo:
Yep, i like the R"()" expression. But we should test it on all platforms as not all platforms support c++ 11 well.

Rubonnek:
OK. I updated the pull request to use `R"()"`.

PR #16749, opened 2016-10-27
Creator C++ support: test cases

ricardoquesada:
This is WIP. DO NOT MERGE


halx99:
The `creator C++ support solution` is generate C++ source code UI-Layout?

ricardoquesada:
@halx99 sorry, I didn't understand the question. 

But right now, the python script parses the Creator ".fire" files and the output are C++ files.

halx99:
So, c++ files must be compiled, and any changes of creator scene or ui, then new c++ files is generted, and need recompile the final app/binary, It's my not  extensibility, why not write a reader to parse .fire and create scene directly?

ricardoquesada:
@halx99 writing a .fire parser in c++ is not an option. it would be very slow for games. people who use C++ (instead of Lua or JS) prefer speed.
So in any case, it should parse FlatBuffer files. But let's go one step at the time: first generate C++ files. If everything works as expected, then create a flatbfuffer parser.

ricardoquesada:
superseded by this PR:
https://github.com/cocos2d/cocos2d-x/pull/16891

Issue #16756, opened 2016-10-27
Use x flip in Tiledmap may cause the map not to be displayed


chgowolf:
Compiler:Visual Studio 2013
Cocos-version:3.13

In Tiledmap, make a new tile with horizontal filp. 
In Cocos, use TMXTiledMap to create the map.
Those layers who use horizontal flip can not to be displayed.

Will only appear in the windows, not appear in the mac.

PR #16758, opened 2016-10-27
fix issue#16756 tileGid may overflow when use horizontal flip

chgowolf:
#16756


cocos/2d/CCTMXXMLParser.cpp
```diff
@@ -746,7 +746,7 @@ void TMXMapInfo::endElement(void *ctx, const char *name)
 
             uint32_t* bufferPtr = reinterpret_cast<uint32_t*>(buffer);
             for(auto gidToken : gidTokens) {
-                auto tileGid = (uint32_t)strtol(gidToken.c_str(), nullptr, 10);
```
minggo, line 749: what's the value of `gidToken.c_str()` if using horizontal flip?

chgowolf: if gid is 1, after using horizontal flip, it will be " 1 | kTMXTileHorizontalFlag" = 0x80000001

minggo: what's the result of ``strtol(gidToken.c_str(), nullptr, 10)` and `strtoul(gidToken.c_str(), nullptr, 10)`?

chgowolf: If value is bigger then "long int" strtol returns it can represent the maximum integer.
In win32 "long int" is same as "int".Value 0x80000001 in "strtol" will return 0x7FFFFFFF and "strtoul" will return 0x80000001

stevetranby: Yep this is a good change. It's what's used to convert string in Value::asUnsignedInt().

stevetranby:
Could you also update Line 358 from

``` cpp
uint32_t gid = static_cast<uint32_t>(attributeDict["gid"].asInt());
```

to this

``` cpp
uint32_t gid = static_cast<uint32_t>(attributeDict["gid"].asUnsignedInt());
```

minggo:
thanks

PR #16762, opened 2016-10-27
Toggle between windowed/fullscreen on desktop

pyrosphere:
On desktop it's currently possible to create a window either fullscreen or windowed.

This PR adds the methods to GLViewImpl on desktop to toggle between fullscreen and windowed. It includes a test for these methods.

### DIFF
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.h
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.h
@@ -79,6 +79,14 @@ public:
     void pollEvents() override;
     GLFWwindow* getWindow() const { return _mainWindow; }
 
+    bool isFullscreen();
+    void setFullscreen();
+    void setFullscreen(int monitorIndex);
+    void setFullscreen(const GLFWvidmode &videoMode, GLFWmonitor *monitor);
+    void setWindowed(int width, int height);
+    int getMonitorCount();
+    Size getMonitorSize();
+
     /* override functions */
     virtual bool isOpenGLReady() override;
     virtual void end() override;
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.h
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.h
@@ -79,13 +79,13 @@ public:
     void pollEvents() override;
     GLFWwindow* getWindow() const { return _mainWindow; }
 
-    bool isFullscreen();
+    bool isFullscreen() const;
     void setFullscreen();
     void setFullscreen(int monitorIndex);
     void setFullscreen(const GLFWvidmode &videoMode, GLFWmonitor *monitor);
     void setWindowed(int width, int height);
-    int getMonitorCount();
-    Size getMonitorSize();
+    int getMonitorCount() const;
+    Size getMonitorSize() const;
 
     /* override functions */
     virtual bool isOpenGLReady() override;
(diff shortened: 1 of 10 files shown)


cocos/platform/desktop/CCGLViewImpl-desktop.h
```diff
@@ -79,6 +79,14 @@ class CC_DLL GLViewImpl : public GLView
     void pollEvents() override;
     GLFWwindow* getWindow() const { return _mainWindow; }
 
+    bool isFullscreen();
```
minggo: bool isFullscreen() -> bool isFullscreen() const

cocos/platform/desktop/CCGLViewImpl-desktop.h
```diff
@@ -79,6 +79,14 @@ class CC_DLL GLViewImpl : public GLView
...
+    void setFullscreen();
+    void setFullscreen(int monitorIndex);
+    void setFullscreen(const GLFWvidmode &videoMode, GLFWmonitor *monitor);
+    void setWindowed(int width, int height);
+    int getMonitorCount();
+    Size getMonitorSize();
```
minggo: add const for `getMonitorCount` and `getMonitorSize` is better

tests/cpp-tests/Classes/WindowTest/WindowTest.cpp
```diff
@@ -0,0 +1,62 @@
...
+    WindowTest::onEnter();
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32) || (CC_TARGET_PLATFORM == CC_PLATFORM_LINUX) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
+    GLViewImpl* view = (GLViewImpl*)Director::getInstance()->getOpenGLView();
+    view->setFullscreen(1);
+#endif
+}
+
+std::string WindowTestFullscreen2::subtitle() const {
+    return "Fullscreen Monitor 2 (if present)";
+}
```
minggo, line 54: how about add the macro

```
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32) || (CC_TARGET_PLATFORM == CC_PLATFORM_LINUX) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
```

like [this](https://github.com/cocos2d/cocos2d-x/blob/v3/tests/cpp-tests/Classes/controller.cpp#L39). Then the test case only appeared on these platforms, and don't have to add macro in this file.

cocos/platform/desktop/CCGLViewImpl-desktop.h
```diff
@@ -79,6 +79,14 @@ class CC_DLL GLViewImpl : public GLView
     void pollEvents() override;
     GLFWwindow* getWindow() const { return _mainWindow; }
 
+    bool isFullscreen();
+    void setFullscreen();
+    void setFullscreen(int monitorIndex);
+    void setFullscreen(const GLFWvidmode &videoMode, GLFWmonitor *monitor);
```
minggo, line 85: why need these two functions:
- void setFullscreen(int monitorIndex);
- void setFullscreen(const GLFWvidmode &videoMode, GLFWmonitor *monitor);

when will use these two functions?

pyrosphere: The method `setFullscreen(int monitorIndex)` lets the game dev add options to switch the game to another monitor, if more than one is present. I personally use it along with `getMonitorCount()` for this.

The other method `setFullscreen(const GLFWvidmode &videoMode, GLFWmonitor *monitor);` is present to give game devs the parallel of `createWithFullScreen(const std::string& viewName, const GLFWvidmode &videoMode, GLFWmonitor *monitor)` which some might be using when creating the window.

minggo:
the demo looks good, thanks

PR #16766, opened 2016-10-28
Adding performance.now() to cocos2d-js. 

DavidDeSimone:
While developing on cocos2d-x, it is useful to bench mark JS with tools like Date.now(). A problem with Date.now() is that it only has millisecond precision, which is not always useful for measuring certain operations. This PR attempts to give cocos2d-js performance.now(), which is an open standard available in modern browsers. You can read the details here https://developer.mozilla.org/en-US/docs/Web/API/Performance/now, and the spec here https://www.w3.org/TR/hr-time/#dom-performance-now.

The quick summary, is that performance.now measures time in milliseconds (up to micro second precision), from the "time origin". I am defining the time origin for us to be the call of ScriptingCore::start(). In order to satisfy the requirement of a monotonic clock, I am using std::chrono::steady_clock.

### DIFF
--- a/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
+++ b/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
@@ -563,6 +563,7 @@ bool ScriptingCore::evalString(const char *string)
 
 void ScriptingCore::start()
 {
+	_engineStartTime = std::chrono::steady_clock::now();
     // for now just this
     createGlobalContext();
 }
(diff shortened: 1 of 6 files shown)


minggo:
will you please add a test case?

DavidDeSimone:
@minggo Added tests, and fix a bad conversion between microseconds to milliseconds with micro precision. Let me know if you would like me to do anything else, or if there was a problem with me setting up the tests.

minggo:
@DavidDeSimone this function only exists in JSB? What about html5? In general, we should add functions both work on web and html5.

DavidDeSimone:
@minggo Modern browsers should implement this function for us in html5. According to https://developer.mozilla.org/en-US/docs/Web/API/Performance/now it is available is all major desktop and mobile browsers, except mobile opera.

minggo:
Thanks for the info, it looks good to me.
@pandamicro what's your opinion?

pandamicro:
Hi, @minggo  I recently added `cc.sys.now` based on `Date.now` in Creator (some old browsers doesn't have Date.now, so `cc.sys.now` include a fallback implementation inside), but I agree with @DavidDeSimone, `Performance.now` is also useful because it provide higher resolution of time and it's more reliable.

I checked our JSB which is based on Spidermonkey v33, `performance` is not defined, so I believe it's implemented in Firefox Gecko and not available in Spidermonkey. So I vote yes for this PR.

minggo:
thanks guys

PR #16772, opened 2016-10-28
Add Horizontal Text Alignment to UIEditBox

joshbodily:
57174bec-9158-11e6-8747-3e1cbf207c5d


minggo:
@joshbodily is this PR still valid?

joshbodily:
Yes, it is still valid.  I need to fix the Win32 compile error in Jenkins, but after that it should be ready.

minggo:

zilongshanren:
@joshbodily  thanks, it looks good to me. 👍

joshbodily:
Thanks guys! 🙏 👍

PR #16774, opened 2016-10-28
CC_UNUSED_PARAM

pyrosphere:
Added CC_UNUSED_PARAM in methods with parameters that are unused, reducing warnings for anyone with -Wunused-parameter


minggo:
may be we should close the the type of warning to make codes more clear

pyrosphere:
That's a valid approach and would simplify the code. I would suggest removing all instances of CC_UNUSED_PARAM to make the code consistent. Currently it's used in some methods but not in others.

pyrosphere:
@minggo I've made a PR #16812 that removes this type of warning and makes the code cleaner as per your suggestion. If that one is merged this one should be closed.

PR #16781, opened 2016-10-31
Support float line width

ggggamer:
glLineWidth use float but DrawNode use int


minggo:
looks good to me, thanks

PR #16788, opened 2016-11-01
synthesize the accessor of window property

morou:
`AppController` class that implements `UIApplicationDelegate` protocol must have accessor methods for `window` property. If not, the following code will raises the exception.

```objc
UIWindow* window = [UIApplication sharedApplication].delegate.window;
```

### DIFF
--- a/templates/js-template-default/frameworks/runtime-src/proj.ios_mac/ios/AppController.h
+++ b/templates/js-template-default/frameworks/runtime-src/proj.ios_mac/ios/AppController.h
@@ -27,7 +27,6 @@
 
 @interface AppController : NSObject <UIApplicationDelegate>
 {
-    UIWindow *window;
     RootViewController    *viewController;
 }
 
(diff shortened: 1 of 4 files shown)


minggo:
how about modify cpp template too?

morou:
Thanks for your suggestion! but cpp template already has `@synthesize` directive for `window` property.

https://github.com/cocos2d/cocos2d-x/blob/v3/templates/cpp-template-default/proj.ios_mac/ios/AppController.mm#L32

The difference between cpp template and js/lua template is that cpp template implements `viewController` as property, but js/lua template implements it as just instance variable.

Can I change the implementation of `viewController` of js/lua template to property?

minggo:
@morou i am not an expert of iOS developing. What do you think is better?

morou:
I think js/lua template implements `viewController` as property, because all templates should have same functionality.

minggo:
Ok, could you please fix it in this PR too?

minggo:
There is a conflict of this PR, please send a new one.

PR #16789, opened 2016-11-01
Add new methods to get the number of actions running in a given node with a specific tag.

saturnodigital:
This is useful for cases that we want to know how many
animations with a **specific** tag is running in our target.
For example:

auto tag1Count = this->getNumberOfRunningActionsByTag(kMyTag1);
auto tag2Count = this->getNumberOfRunningActionsByTag(kMyTag2);

While this could be achieved (more or less) with callbacks
to increment the tagCount at start of action and decrement
at the end the proposed API is much more concise and less error
prone.

### DIFF
--- a/cocos/2d/CCActionManager.cpp
+++ b/cocos/2d/CCActionManager.cpp
@@ -90,7 +90,7 @@ void ActionManager::actionAllocWithHashElement(tHashElement *element)
 
 void ActionManager::removeActionAtIndex(ssize_t index, tHashElement *element)
 {
-    Action *action = static_cast<Action *>(element->actions->arr[index]);
+    Action *action = static_cast<Action*>(element->actions->arr[index]);
 
     if (action == element->currentAction && (! element->currentActionSalvaged))
     {
@@ -415,7 +415,7 @@ size_t ActionManager::getNumberOfRunningActionsInTargetByTag(const Node *target,
     auto limit = element->actions->num;
     for(int i = 0; i < limit; ++i)
     {
-        auto action = static_cast<Action *>element->actions->arr[i];
+        auto action = static_cast<Action*>(element->actions->arr[i]);
         if(action->getTag() == tag)
             ++count;
     }
--- a/cocos/2d/CCActionManager.cpp
+++ b/cocos/2d/CCActionManager.cpp
@@ -90,7 +90,7 @@ void ActionManager::actionAllocWithHashElement(tHashElement *element)
 
 void ActionManager::removeActionAtIndex(ssize_t index, tHashElement *element)
 {
-    Action *action = (Action*)element->actions->arr[index];
+    Action *action = static_cast<Action *>(element->actions->arr[index]);
 
     if (action == element->currentAction && (! element->currentActionSalvaged))
     {
@@ -275,7 +275,7 @@ void ActionManager::removeActionByTag(int tag, Node *target)
         auto limit = element->actions->num;
         for (int i = 0; i < limit; ++i)
         {
-            Action *action = (Action*)element->actions->arr[i];
+            Action *action = static_cast<Action*>(element->actions->arr[i]);
 
             if (action->getTag() == (int)tag && action->getOriginalTarget() == target)
             {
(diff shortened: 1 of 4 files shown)


cocos/2d/CCActionManager.cpp
```diff
@@ -398,6 +398,32 @@ ssize_t ActionManager::getNumberOfRunningActionsInTarget(const Node *target) con
...
+    HASH_FIND_PTR(_targets, &target, element);
+
+    if(!element || !element->actions)
+        return 0;
+
+    int count = 0;
+    auto limit = element->actions->num;
+    for(int i = 0; i < limit; ++i)
+    {
+        auto action = (Action *)element->actions->arr[i];
```
minggo: it is better to use `static_cast<Action*>` instead

cocos/2d/CCActionManager.h
```diff
@@ -145,6 +145,24 @@ class CC_DLL ActionManager : public Ref
...
+     * - If you are running 7 Sequences of 2 actions, it will return 7.
+     *
+     * @param target    A certain target.
+     * @param tag       Tag that will be searched.
+     * @return  The numbers of actions that are running in a certain target
+     *          with a specific tag.
+     * @see getNumberOfRunningActionsInTarget
+     * @js NA
+     */
+    size_t getNumberOfRunningActionsInTargetByTag(const Node *target, int tag);
```
minggo, line 163: it is better to add const like 

``` c++
size_t getNumberOfRunningActionsInTargetByTag(const Node *target, int tag) const;
```

minggo: and it is better to use reference instead like 

``` c++
size_t getNumberOfRunningActionsInTargetByTag(const Node &target, int tag) const;
```

or we should check the invalidation of target.

cocos/2d/CCActionManager.cpp
```diff
@@ -398,6 +398,32 @@ ssize_t ActionManager::getNumberOfRunningActionsInTarget(const Node *target) con
     return 0;
 }
 
+// FIXME: Passing "const O *" instead of "const O&" because HASH_FIND_IT requires the address of a pointer
+// and, it is not possible to get the address of a reference
+size_t ActionManager::getNumberOfRunningActionsInTargetByTag(const Node *target,
```
cmzx3444, line 403: Method name can be shorter? like getRunningActionsNumber?

saturnodigital: I have the same feel, but I just  followed the pattern in the code!
The other methods are "overloaded" by adding the "byTag" suffix, so I did the same.

For the other side, I think that is easier to make a mental model with this name since
we are used to the other methods that uses the "byTag" suffix.

cocos/2d/CCActionManager.cpp
saturnodigital: @minggo I just used the same pattern of the "surrounding" methods.
I agree with the static_assert, but we should change for the other methods too?

cocos/2d/CCActionManager.h
saturnodigital: @minggo I just used the same pattern of the "surrounding" methods.
Actually I didn't understand your comment "it is better to add const like ...", my PR 
is already using ``const Node *`, isn't it?

About the reference, I have no ideas what would be better, we have a lot of comments with this issue:

```
// FIXME: Passing "const O *" instead of "const O&" because HASH_FIND_IT requires the address of a pointer
// and, it is not possible to get the address of a reference
```

But this could go deeper in the engine architecture than I could make a clear statement about.
I leave to you guys decide what is better, regarding this modification.

Anyway is just let me know about that I'll happily make another PR with the changes!

cocos/2d/CCActionManager.cpp
minggo: You are appreciated if you can help to fix other methods.

cocos/2d/CCActionManager.cpp
minggo: follow the same pattern looks good to me

cocos/2d/CCActionManager.h
minggo: ok, let's keep the same pattern at this PR

PR #16797, opened 2016-11-02
Use resolved file name (through dictionary lookup) in font atlas creation

adesentenac:
If using the dictionary lookup mechanism of FileUtils, and multiple TTF files are replaced by the same one, FontAtlasCache was creating as many duplicate copies of the replacing font as there were of replaced fonts.

When using big TTF font, that can lead to a significant increase in memory used.

I've added 2 tests to FontTests. The first one is the reference test without any replacements, the second one have every font used being replaced by arial.

Without the change to FontAtlasCache, the second test takes more memory than the first one (since multiple atlases are created for arial, which is a bigger TTF file than the other), whereas with the change, the second test takes less memory (since only one atlas is created for arial).

### DIFF
--- a/build/cocos2d_tests.xcodeproj/project.pbxproj
+++ b/build/cocos2d_tests.xcodeproj/project.pbxproj
@@ -8336,7 +8336,7 @@
 				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
 				INFOPLIST_FILE = "$(SRCROOT)/../tests/cpp-tests/proj.mac/Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "@loader_path/../Frameworks";
-				USER_HEADER_SEARCH_PATHS = "$(inherited) $(SRCROOT)/../external/glfw3/include/mac";
+				USER_HEADER_SEARCH_PATHS = "$(inherited) $(SRCROOT)/../external/glfw3/include/mac $(SRCROOT)/../external/freetype2/include/mac/freetype2";
 			};
 			name = Debug;
 		};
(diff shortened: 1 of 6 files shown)


minggo:
please fix compiling error on linux, mac and iOS

```
In file included from /home/travis/build/cocos2d/cocos2d-x/cocos/2d/CCFontFreeType.h:34:0,
                 from /home/travis/build/cocos2d/cocos2d-x/tests/cpp-tests/Classes/FontTest/FontTest.cpp:4:
/usr/include/ft2build.h:56:38: fatal error: freetype/config/ftheader.h: No such file or directory
 #include <freetype/config/ftheader.h>
```

minggo:
@adesentenac could you please fix the compiling error?

adesentenac:
Hi @minggo the compiling errors are fixed, sorry for the delay.

build/cocos2d_tests.xcodeproj/project.pbxproj
```diff
@@ -8336,7 +8336,7 @@
 				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
 				INFOPLIST_FILE = "$(SRCROOT)/../tests/cpp-tests/proj.mac/Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "@loader_path/../Frameworks";
-				USER_HEADER_SEARCH_PATHS = "$(inherited) $(SRCROOT)/../external/glfw3/include/mac";
+				USER_HEADER_SEARCH_PATHS = "$(inherited) $(SRCROOT)/../external/glfw3/include/mac $(SRCROOT)/../external/freetype2/include/mac/freetype2";
```
minggo, line 8339: why need to include `$(SRCROOT)/../external/freetype2/include/mac/freetype2`?

minggo: Got it, ignore it.

PR #16801, opened 2016-11-03
fixed #16800: [AssetsManagerEx] project.manifest will be downloaded twice

### LINKED ISSUE #16800: [AssetsManagerEx] project.manifest will be downloaded twice
- cocos2d-x version: 3.13.1
- devices test on: Android, Mac
- developing environments
   - NDK version: r10e
   - Xcode version: 8.1
   - VS version: N/A
   - browser type and version: N/A

Steps to Reproduce:

```c++
    std::string updateDir = FileUtils::getInstance()->getWritablePath() + "assetsmanager-test";
    log("updateDir: %s", updateDir.c_str());
    FileUtils::getInstance()->addSearchPath(updateDir, true);

    auto manager = AssetsManagerEx::create("project.manifest", updateDir);
    manager->retain();

    auto listener = EventListenerAssetsManagerEx::create(manager, [=](EventAssetsManagerEx* event){

        static const char* eventCodeStr[] = {
            "ERROR_NO_LOCAL_MANIFEST",
            "ERROR_DOWNLOAD_MANIFEST",
            "ERROR_PARSE_MANIFEST",
            "NEW_VERSION_FOUND",
            "ALREADY_UP_TO_DATE",
            "UPDATE_PROGRESSION",
            "ASSET_UPDATED",
            "ERROR_UPDATING",
            "UPDATE_FINISHED",
            "UPDATE_FAILED",
            "ERROR_DECOMPRESS"
        };

        log("event code: %s, assetID: %s, percent: %f", eventCodeStr[(int)event->getEventCode()], event->getAssetId().c_str(), event->getPercent());
        if (event->getEventCode() == EventAssetsManagerEx::EventCode::NEW_VERSION_FOUND)
        {
            manager->update();
        }
    });

    Director::getInstance()->getEventDispatcher()->addEventListenerWithSceneGraphPriority(listener, this);

    manager->checkUpdate();
```

dumganhar:


extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -423,6 +419,32 @@ void AssetsManagerEx::decompressDownloadedZip()
...
+            break;
+        case EventAssetsManagerEx::EventCode::NEW_VERSION_FOUND:
+            if (_updateEntry == UpdateEntry::CHECK_UPDATE)
+            {
+                _updateEntry = UpdateEntry::NONE;
+            }
+            break;
+        default:
+            break;
+    }
```
pandamicro, line 446: Why put all _updateEntry setting together instead of setting it in related logic ?

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -476,11 +498,14 @@ void AssetsManagerEx::parseVersion()
...
         {
+            // Save the old update entry type since its value may be changed in event listener.
+            UpdateEntry oldUpdateEntry = _updateEntry;
+
             _updateState = State::NEED_UPDATE;
             dispatchUpdateEvent(EventAssetsManagerEx::EventCode::NEW_VERSION_FOUND);
 
             // Wait to update so continue the process
-            if (_waitToUpdate)
+            if (oldUpdateEntry == UpdateEntry::DO_UPDATE)
```
pandamicro, line 508: If the old entry was CHECK_UPDATE then it will be set to NONE, why not check 

if (_updateEntry == UpdateEntry::DO_UPDATE) here ?

extensions/assets-manager/AssetsManagerEx.cpp
dumganhar: dispatchUpdateEvent is a function will be invoked in many places.
I don't wanna insert lots of `_updateEntry = UpdateEntry:XXX` after `dispatchUpdateEvent`.
So it's a common place to reset `_updateEntry` variable.

extensions/assets-manager/AssetsManagerEx.cpp
dumganhar: If it's in `checkUpdate` logic, we don't need to do the real `update` logic since developers should take care of whether to do `update` in EventListenerAssetsManagerEx's callback.

pandamicro: Yes, I understand, I mean if it's in checkUpdate, then _updateEntry will be set to NONE, then it's ok to verify _updateEntry here, no need to use oldUpdateEntry

dumganhar: This logic is from the original code. I don't wanna change big logic of old code.
My thought is: if developers didn't invoke `checkUpdate`, he/she just invokes `update` method, then `oldUpdateEntry` is `DO_UPDATE` here, need to `continue` the following steps like downloading manifest, assets ..., otherwise, the flow may be stopped.

pandamicro (approved):

PR #16812, opened 2016-11-04
No CC_UNUSED_PARAM

pyrosphere:
In #16774 I tried to make cocos2d-x more consistent in the use of CC_UNUSED_PARAM but to keep the code more simple and cleaner it was suggested it might be better to just ignore this type of warning.

This PR adds the -Wno-unused-parameter (which is turned on in DEBUG with -Wextra) and removed all the uses of the macro CC_UNUSED_PARAM. The macro itself is kept as it might be in use by some game devs in their own code.


stevetranby:
@pyrosphere @minggo 

This may be an invalid solution, but ...

I think where possible you could just remove/comment the parameter name from most or all of these instances of code where CC_UNUSED_PARAM was used before. 

The warning could then stay enabled to allow catching some subtle error in the future? 

```
void Action::step(float dt)
{
  CC_UNUSED_PARAM(dt);
  CCLOG("[Action step]. override me");
}
```

After

```
void Action::step(float /*dt*/) // or just `step(float)`
{
  CCLOG("[Action step]. override me");
}
```
- I prefer just removing the name because then any change to isn't more complicated by having to find it in the comments. And if you ever need to know the name to use it later one can just look at the header and add the name (which, honestly, should already be known since the name likely explains the usage).

2c

stevetranby:
It would not surprise me if there were issues that occur when not using the actual name, but it's the method I usually use if there's a good reason for the parameter to exist and yet there's not a usage of it within. YMMV

Maybe documentation?
Specific compiler issues?
etc

minggo:
@stevetranby the solution looks smart. @pyrosphere what's your opinion?

stevetranby:
The main issue reading all the changes would be the header files. The parameter names are probably necessary for Doxygen (not something we care about) and the use cases inside header files mostly in virtual base classes my suggestion might make it less clear.

It might be nice to reduce the warnings in the .cpp files, however. So you could keep the warning disabled by default, but if someone enables it briefly for testing or some other reason then they wouldn't see as many warnings??

minggo:
I am not sure if the parameter name in function decalaration is needed by Doxygen. I think Doxygen just need to parse `@param` section.

pyrosphere:
@minggo The solution @stevetranby suggests is a good one. I would prefer `step(float)` over `step(float /*dt*/)`, applied to .cpp files.

pyrosphere:
I checked the coding style doc and it says

> If some parameters are unused, comment out the variable name in the function definition

so /\* var_name */ seems the best approach.

stevetranby:
One last concern would be issues with less helpful autocomplete.

Maybe all method declarations in header files and especially valid inline virtual methods that have an empty body block (which triggers the warning) should keep their actual parameter names for the purpose of autocomplete, etc? UNUSED_PARAM() to get rid of warning if desired.

Could also move non-inline methods into the .cpp files where appropriate since declarations don't trigger the warning, only definitions.

minggo:
Keep parameter name in header file is ok since it will not cause warning.

cmake/Modules/SetCompilerOptions.cmake
```diff
@@ -53,7 +53,7 @@ macro (SetCompilerOptions)
 	  set(CMAKE_C_FLAGS_DEBUG "-g -Wall -Wextra")
 	  set(CMAKE_CXX_FLAGS_DEBUG ${CMAKE_C_FLAGS_DEBUG})
 	  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -fPIC")
-	  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-deprecated-declarations -Wno-reorder -Wno-invalid-offsetof -fPIC")
+	  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-deprecated-declarations -Wno-reorder -Wno-invalid-offsetof -Wno-unused-parameter -fPIC")
```
minggo: I think we don't need it since we fix warnings.

cocos/2d/CCLayer.h
```diff
@@ -69,15 +69,15 @@ class CC_DLL Layer : public Node
...
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchCancelled(Touch *pTouch, Event *pEvent) final {}
     
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesBegan(__Set *pTouches, Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesMoved(__Set *pTouches, Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesEnded(__Set *pTouches, Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesCancelled(__Set *pTouches, Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesBegan(__Set *pTouches, Event *pEvent) final {}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesMoved(__Set *pTouches, Event *pEvent) final {}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesEnded(__Set *pTouches, Event *pEvent) final {}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesCancelled(__Set *pTouches, Event *pEvent) final {}
```
minggo: these functions are defined in header files, i think we should move the definition into .cpp file.

minggo: or we should comment the parmaters

pyrosphere: I think the best approach is to move the definition to the .cpp and avoid commenting parameter names in headers.

cocos/2d/CCNode.h
```diff
@@ -1790,7 +1790,7 @@ class CC_DLL Node : public Ref
      *  If you want the opacity affect the color property, then set to true.
      * @param value A boolean value.
      */
-    virtual void setOpacityModifyRGB(bool value) {CC_UNUSED_PARAM(value);}
+    virtual void setOpacityModifyRGB(bool value) {}
```
minggo: will it cause warning?

cocos/2d/CCMenu.h
```diff
@@ -171,7 +171,7 @@ class CC_DLL Menu : public Layer
     
     virtual void onEnter() override;
     virtual void onExit() override;
-    virtual void setOpacityModifyRGB(bool bValue) override {CC_UNUSED_PARAM(bValue);}
+    virtual void setOpacityModifyRGB(bool bValue) override {}
```
minggo: will it cause warning?

cocos/2d/CCTextFieldTTF.h
```diff
@@ -73,9 +71,6 @@ class CC_DLL TextFieldDelegate
      */
     virtual bool onTextFieldInsertText(TextFieldTTF * sender, const char * text, size_t nLen)
     {
-        CC_UNUSED_PARAM(sender);
-        CC_UNUSED_PARAM(text);
-        CC_UNUSED_PARAM(nLen);
```
minggo, line 78: will it cause warning?

minggo:
@pyrosphere i found you just remove `CC_UNUSED_PARAM` in some functions that are defined in header files, i think it will cause warnings. How about define these functions in corresponding cpp files?

minggo:
Keep parameter names in header files are needed to generate lua API files that [in this folder:](https://github.com/cocos2d/cocos2d-x/tree/v3/cocos/scripting/lua-bindings/auto/api). These files are used by code IDE for code completion. Code IDE is canceled, but i don't know if there is some developers need these files too, so i think it is better to keep as it is.

minggo:
@pyrosphere any updated?

yahont:
Please let me know when it's merged by commenting in #16838

cocos/2d/CCLayer.cpp
```diff
@@ -246,10 +255,8 @@ void Layer::setAccelerometerInterval(double interval) {
     }
 }
 
-void Layer::onAcceleration(Acceleration* acc, Event* unused_event)
+void Layer::onAcceleration(Acceleration* acc, Event* /*unused_event*/)
 {
-    CC_UNUSED_PARAM(acc);
```
yahont, line 251: acc is still unused if CC_ENABLE_SCRIPT_BINDING is false, it might be better too add the following at the end of the function:

```
#else
    CC_UNUSED_PARAM(acc);
#endif
```

cocos/2d/CCLayer.cpp
```diff
@@ -260,15 +267,12 @@ void Layer::onAcceleration(Acceleration* acc, Event* unused_event)
...
+void Layer::onKeyPressed(EventKeyboard::KeyCode /*keyCode*/, Event* /*unused_event*/)
 {
-    CC_UNUSED_PARAM(keyCode);
-    CC_UNUSED_PARAM(unused_event);
 }
 
-void Layer::onKeyReleased(EventKeyboard::KeyCode keyCode, Event* unused_event)
+void Layer::onKeyReleased(EventKeyboard::KeyCode keyCode, Event* /*unused_event*/)
 {
-    CC_UNUSED_PARAM(unused_event);
```
yahont, line 271: keyCode is unused if CC_ENABLE_SCRIPT_BINDING is false, it would be nice to have the following at the end of the function:

```
#else
    CC_UNUSED_PARAM(keyCode);
#endif
```

cocos/2d/CCLayer.cpp
```diff
@@ -320,7 +324,6 @@ bool Layer::onTouchBegan(Touch *touch, Event *event)
         return executeScriptTouchHandler(EventTouch::EventCode::BEGAN, touch, event) == 0 ? false : true;
     }
 #endif
-    CC_UNUSED_PARAM(event);
```
yahont, line 323: touch and event are unused if CC_ENABLE_SCRIPT_BINDING is false, it might be better to add the following at the end of the function:

```
#else
    CC_UNUSED_PARAM(touch);
    CC_UNUSED_PARAM(event);
#endif
```

cocos/2d/CCLayer.cpp
```diff
@@ -347,8 +348,6 @@ void Layer::onTouchEnded(Touch *touch, Event *event)
         return;
     }
 #endif
-    
-    CC_UNUSED_PARAM(event);
```
yahont, line 351:
```
#else
    CC_UNUSED_PARAM(event);
    CC_UNUSED_PARAM(touch);
#endif
```

cocos/2d/CCLayer.cpp
```diff
@@ -360,8 +359,6 @@ void Layer::onTouchCancelled(Touch *touch, Event *event)
         return;
     }
```
yahont, line 380:
```
#else
    CC_UNUSED_PARAM(event);
    CC_UNUSED_PARAM(touch);
```

cocos/2d/CCLayer.cpp
```diff
@@ -373,7 +370,6 @@ void Layer::onTouchesBegan(const std::vector<Touch*>& touches, Event *event)
...
 #endif
-    CC_UNUSED_PARAM(event);
```
yahont, line 376:
```
#else
    CC_UNUSED_PARAM(event);
    CC_UNUSED_PARAM(touches);
#endif
```

cocos/2d/CCLayer.cpp
```diff
@@ -385,8 +381,6 @@ void Layer::onTouchesMoved(const std::vector<Touch*>& touches, Event *event)
         return;
     }
 #endif
-    
-    CC_UNUSED_PARAM(event);
```
yahont, line 389:
```
#else
    CC_UNUSED_PARAM(event);
    CC_UNUSED_PARAM(touches);
#endif
```

cocos/2d/CCLayer.cpp
```diff
@@ -398,7 +392,6 @@ void Layer::onTouchesEnded(const std::vector<Touch*>& touches, Event *event)
         return;
     }
 #endif
-    CC_UNUSED_PARAM(event);
```
yahont, line 401:
```
#else
    CC_UNUSED_PARAM(event);
    CC_UNUSED_PARAM(touches);
#endif
```

cocos/2d/CCLayer.cpp
```diff
@@ -410,7 +403,6 @@ void Layer::onTouchesCancelled(const std::vector<Touch*>& touches, Event *event)
...
```
yahont, line 413:
```
#else
    CC_UNUSED_PARAM(event);
    CC_UNUSED_PARAM(touches);
#endif
```

cocos/base/CCIMEDelegate.h
```diff
@@ -116,7 +116,7 @@ class CC_DLL IMEDelegate
     * @js NA
     * @lua NA
     */
-    virtual void insertText(const char * text, size_t len) {CC_UNUSED_PARAM(text);CC_UNUSED_PARAM(len);}
+    virtual void insertText(const char * text, size_t len) {}
```
yahont: text and len should be commented out

cocos/base/CCIMEDelegate.h
```diff
@@ -146,22 +146,22 @@ class CC_DLL IMEDelegate
      * @js NA
      * @lua NA
      */
-    virtual void keyboardWillShow(IMEKeyboardNotificationInfo& info)   {CC_UNUSED_PARAM(info);}
+    virtual void keyboardWillShow(IMEKeyboardNotificationInfo& info)   {}
```
yahont:
```
/*info*/
```

cocos/base/CCIMEDelegate.h
```diff
@@ -146,22 +146,22 @@ class CC_DLL IMEDelegate
...
     /**
      * @js NA
      * @lua NA
      */
-    virtual void keyboardDidShow(IMEKeyboardNotificationInfo& info)    {CC_UNUSED_PARAM(info);}
+    virtual void keyboardDidShow(IMEKeyboardNotificationInfo& info)    {}
```
yahont:
```
/*info*/
```

cocos/base/CCIMEDelegate.h
```diff
@@ -146,22 +146,22 @@ class CC_DLL IMEDelegate
...
     /**
      * @js NA
      * @lua NA
      */
-    virtual void keyboardWillHide(IMEKeyboardNotificationInfo& info)   {CC_UNUSED_PARAM(info);}
+    virtual void keyboardWillHide(IMEKeyboardNotificationInfo& info)   {}
```
yahont:
```
/*info*/
```

cocos/base/CCIMEDelegate.h
```diff
@@ -146,22 +146,22 @@ class CC_DLL IMEDelegate
...
     /**
      * @js NA
      * @lua NA
      */
-    virtual void keyboardDidHide(IMEKeyboardNotificationInfo& info)    {CC_UNUSED_PARAM(info);}
+    virtual void keyboardDidHide(IMEKeyboardNotificationInfo& info)    {}
```
yahont:
```
/*info*/
```

cocos/editor-support/cocostudio/CCInputDelegate.h
```diff
@@ -69,40 +69,40 @@ class CC_STUDIO_DLL InputDelegate
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual void didAccelerate(cocos2d::Acceleration* accelerationValue) final { CC_UNUSED_PARAM(accelerationValue); };
+    CC_DEPRECATED_ATTRIBUTE virtual void didAccelerate(cocos2d::Acceleration* accelerationValue) final {};
```
yahont:
```
/*accelerationValue*/
```

cocos/editor-support/cocostudio/CCInputDelegate.h
```diff
@@ -69,40 +69,40 @@ class CC_STUDIO_DLL InputDelegate
...
     // Deprecated touch callbacks.
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual bool ccTouchBegan(cocos2d::Touch *pTouch, cocos2d::Event *pEvent) final {CC_UNUSED_PARAM(pTouch); CC_UNUSED_PARAM(pEvent); return false;};
+    CC_DEPRECATED_ATTRIBUTE virtual bool ccTouchBegan(cocos2d::Touch *pTouch, cocos2d::Event *pEvent) final {return false;};
```
yahont: Shouldn't pTouch and pEvent be commented out?

cocos/editor-support/cocostudio/CCInputDelegate.h
```diff
@@ -69,40 +69,40 @@ class CC_STUDIO_DLL InputDelegate
...
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchMoved(cocos2d::Touch *pTouch, cocos2d::Event *pEvent) final {CC_UNUSED_PARAM(pTouch); CC_UNUSED_PARAM(pEvent);}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchMoved(cocos2d::Touch *pTouch, cocos2d::Event *pEvent) final {}
```
yahont: Shouldn't pTouch and pEvent be commented out?

cocos/editor-support/cocostudio/CCInputDelegate.h
```diff
@@ -69,40 +69,40 @@ class CC_STUDIO_DLL InputDelegate
...
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchEnded(cocos2d::Touch *pTouch, cocos2d::Event *pEvent) final {CC_UNUSED_PARAM(pTouch); CC_UNUSED_PARAM(pEvent);}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchEnded(cocos2d::Touch *pTouch, cocos2d::Event *pEvent) final {}
```
yahont: Shouldn't pTouch and pEvent be commented out?

cocos/editor-support/cocostudio/CCInputDelegate.h
```diff
@@ -69,40 +69,40 @@ class CC_STUDIO_DLL InputDelegate
...
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesEnded(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesEnded(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {}
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesCancelled(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesCancelled(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {}
```
yahont: Shouldn't pTouches and pEvent be commented out?

cocos/editor-support/cocostudio/CCInputDelegate.h
```diff
@@ -69,40 +69,40 @@ class CC_STUDIO_DLL InputDelegate
...
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesMoved(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesMoved(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {}
     /**
      * @js NA
      */
-    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesEnded(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {CC_UNUSED_PARAM(pTouches); CC_UNUSED_PARAM(pEvent);}
+    CC_DEPRECATED_ATTRIBUTE virtual void ccTouchesEnded(cocos2d::__Set *pTouches, cocos2d::Event *pEvent) final {}
```
yahont: Shouldn't pTouches and pEvent be commented out?

yahont (changes requested):

minggo:
Please fix the compiling error on linux

```
tests.dir/Classes/ExtensionsTest/ControlExtensionTest/CCControlSceneManager.cpp.o
../../../lib/libcocos2dInternal.a(CSLoader.cpp.o): In function `cocos2d::CSLoader::bindCallback(std::string const&, std::string const&, cocos2d::ui::Widget*, cocos2d::Node*)':
CSLoader.cpp:(.text+0x18e5): undefined reference to `typeinfo for cocostudio::WidgetCallBackHandlerProtocol'
../../../lib/libcocos2dInternal.a(CSLoader.cpp.o): In function `cocos2d::CSLoader::nodeWithFlatBuffers(flatbuffers::NodeTree const*, std::function<void (cocos2d::Ref*)> const&)':
CSLoader.cpp:(.text+0x6694): undefined reference to `typeinfo for cocostudio::WidgetCallBackHandlerProtocol'
../../../lib/libcocos2dInternal.a(CSLoader.cpp.o): In function `cocos2d::CSLoader::createNode(cocos2d::Data const&, std::function<void (cocos2d::Ref*)> const&)':
CSLoader.cpp:(.text+0x6e70): undefined reference to `typeinfo for cocostudio::WidgetCallBackHandlerProtocol'
../../../lib/libcocos2dInternal.a(CSLoader.cpp.o): In function `cocos2d::CSLoader::createNodeWithFlatBuffersFile(std::string const&, std::function<void (cocos2d::Ref*)> const&)':
CSLoader.cpp:(.text+0x7309): undefined reference to `typeinfo for cocostudio::WidgetCallBackHandlerProtocol'
make[2]: *** [bin/lua-tests/lua-tests] Error 1
make[1]: *** [tests/lua-tests/project/CMakeFiles/lua-tests.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

I will merge it when it passes CI.

minggo:
thanks

PR #16815, opened 2016-11-07
[Improve]use STL emplace when possible

cezheng:
* replaced some unordered_map::insert(std::make_pair(foo, bar)) with unordered_map::emplace(foo, bar)
* replaced some vector::push_back(std::make_pair(foo, bar)) with vector::emplace_back(foo, bar)

The old way will construct a std::pair first then call move constructor when putting it into the container, while using emplace will construct the pair in-place in the container which will make it a little bit faster. Also, the emplace way is shorter & more concise.


minggo:
good to know, thanks

PR #16816, opened 2016-11-07
rapidjson::FileStream should be replace with rapidjson::FileReadStream 

deflinhec:
This issue may cause compile error when executing cocos gen-libs


tools/simulator/libsimulator/lib/runtime/FileServer.cpp
```diff
@@ -62,7 +62,8 @@ void FileServer::readResFileFinfo()
     FILE * pFile = fopen (filecfg.c_str() , "r");
     if(pFile)
     {
-        rapidjson::FileStream inputStream(pFile);
+        char buffer[65536];
+        rapidjson::FileReadStream inputStream(pFile, buffer, sizeof(buffer));
```
dumganhar, line 66: What's `char buffer[65536]` used for?

deflinhec: FileReadStream requires a buffer space in order to fread to it.

https://github.com/miloyip/rapidjson/blob/master/include/rapidjson/filereadstream.h

minggo:
why it only cause issue in `cocos gen-libs`?

dumganhar:
@falconlegend, could we get rid of hard code 65536?
Could we just allocate the buffer size which is the same as the file size?

stevetranby:
I believe that it doesn't matter what size it is for "correctness" (somewhere in rapidjson docs). So you could set it to `buffer[1]` and it should still work, technically. 

It's a number that could be changed for each platform specific to its memory paging and storage characteristics. So you can put the buffer on heap/stack/elsewhere. AFAIK

stevetranby:
@minggo issue present if you open the mac libsimulator project manually. When else is that built other than through gen-libs? At least recently. I'm sure it is/was built when sending out a new engine version that included Cocos IDE. Cocos Creator is on a whole separate runtime, no?

minggo:
@stevetranby Cocos IDE is outdated and Cocos Creator is using lite version of cocos2d-x which removes some modules such as 3d, cocos studio reader and so on.

> I believe that it doesn't matter what size it is for "correctness" (somewhere in rapidjson docs). So you could set it to buffer[1] and it should still work, technically.

@falconlegend is it true? If so, may be we can use just buffer[1] and add some comment.

stevetranby:
@minggo I'd prob use a smarter default like [4096] or larger.

dumganhar:
> I believe that it doesn't matter what size it is for "correctness" (somewhere in rapidjson docs). So you could set it to buffer[1] and it should still work, technically.

> It's a number that could be changed for each platform specific to its memory paging and storage characteristics. So you can put the buffer on heap/stack/elsewhere. AFAIK

@stevetranby,  good to know this. :)

PR #16818, opened 2016-11-08
cocos2dInternal cleaned of some warnings

yahont:
Cleaned of most unused parameters and signed-unsigned comparisons warnings found with gcc-6.2


minggo:
I think this PR includes some modification included in https://github.com/cocos2d/cocos2d-x/pull/16812

minggo:
looks good to me, thanks

minggo:
https://github.com/cocos2d/cocos2d-x/pull/16828/files it seems the parameter in header file is needed to generate lua API files though i don't know who uses these files.

PR #16822, opened 2016-11-08
required check for nullptr with  new(std::nothrow)

yahont:


cocos/2d/CCCameraBackgroundBrush.cpp
```diff
@@ -95,9 +95,12 @@ CameraBackgroundDepthBrush::~CameraBackgroundDepthBrush()
 CameraBackgroundDepthBrush* CameraBackgroundDepthBrush::create(float depth)
 {
     auto ret = new (std::nothrow) CameraBackgroundDepthBrush();
-    ret->_depth = depth;
-    ret->init();
-    ret->autorelease();
+    if (nullptr != ret)
+    {
+        ret->_depth = depth;
+        ret->init();
```
dumganhar: How about also check the return value of `init`?
For example:

``` c++
auto ret = new (std::nothrow) XXX;
if (ret != nullptr && ret->init())
{
...
...
    ret->autorelease();
}
else
{
    CC_SAFE_DELETE(ret);
}
```

cocos/2d/CCCameraBackgroundBrush.cpp
```diff
@@ -264,35 +279,64 @@ CameraBackgroundSkyBoxBrush::~CameraBackgroundSkyBoxBrush()
...
+
+        if (nullptr != ret && ret->init())
+        {
+            ret->setTexture(texture);
+            ret->autorelease();
+        }
+        else
+        {
+            CC_SAFE_DELETE(texture);
+            CC_SAFE_DELETE(ret);
```
minggo, line 320: ret should be set to null or an invalid pointer will be returned

minggo: `CC_SAFE_DELETE` will set ret to null, sorry about it

Issue #16158, opened 2016-07-19
[Cocos2d-x 3.12] extensions::ControlSwitch::create() bug


alfogrillo:
The function 
`ControlSwitch* create(Sprite *maskSprite, Sprite * onSprite, Sprite * offSprite, Sprite * thumbSprite)` makes the game to crash because calls `addChild` on `nullptr` object.
The function tries to add a null `Label` that user can specify using this other function:
`ControlSwitch* create(Sprite *maskSprite, Sprite * onSprite, Sprite * offSprite, Sprite * thumbSprite, Label* onLabel, Label* offLabel)`

dumganhar:
Does this issue appear from v3.12?

minggo:
@Drakon-Cocos how can i reproduce it?

alfogrillo:
@dumganhar : I don't know, it is the first time I use `extensions::ControlSwitch` class.
@minggo : here there is a simple test project (iOS tested).

[ControlSwitchTest.zip](https://github.com/cocos2d/cocos2d-x/files/373067/ControlSwitchTest.zip)

Anyway I think that `ControlSwitch::create` is not consistent with cocos2d-x style.
It should be something like that:
ControlSwitch\* create(const string& maskSprite, const string& onSprite, const string& offSprite, const string& thumbSprite)
Parameters of type`const string&` instead of `Sprite*`

minggo:
@Drakon-Cocos these features are added by cocosbuilder author to support cocosbuilder, and we port it in C++ to support cocosbuilder. It is not part of cocos2d-x core, it is a long time unmaintained.

minggo:
close this issue since the PR is merged, please reopen it if it still has the issue

Issue #14946, opened 2016-01-22
ControlSwitch::create(Sprite *maskSprite, Sprite * onSprite, Sprite * offSprite, Sprite * thumbSprite) crashes


ingtellect:
ControlSwitch::create(Sprite _maskSprite, Sprite \* onSprite, Sprite \* offSprite, Sprite \* thumbSprite) crashe. It calls initWithMaskSprite(maskSprite, onSprite, offSprite, thumbSprite, nullptr, nullptr) and eventually crashes when calling clipper->addChild(onLabel) and clipper->addChild(offLabel) in ControlSwitchSprite::initWithMaskSprite(
    Sprite *maskSprite, 
    Sprite *onSprite, 
    Sprite *offSprite,
    Sprite *thumbSprite,
    Label_ onLabel, 
    Label\* offLabel)
because onLabel and OffLabel is nullptr.

minggo:
close this issue since the PR is merged. And please reopen it if it is not fixed.

PR #16827, opened 2016-11-09
Fix ControlSwitch bug: added null check.

bugttle:
I found same issues.
 - https://github.com/cocos2d/cocos2d-x/issues/14946
 - https://github.com/cocos2d/cocos2d-x/issues/16158


minggo:
@zilongshanren please take a look. It looks good to me.

zilongshanren:
@minggo  yep, it looks good.

bugttle:
@minggo @zilongshanren I'm glad, thanks so much.

PR #16830, opened 2016-11-10
Android setMultipleTouchEnabled

pyrosphere:
On iOS switching between single or multitouch app is done with a single call to `[eaglView setMultipleTouchEnabled:NO];` but there is no simple equivalent on Android.

This PR adds this feature to `Cocos2dxGLSurfaceView` which can then be easily called on `AppActivity` with `this.getGLSurfaceView().setMultipleTouchEnabled(false);`.

This is a recurring situation that has been discussed [here](http://discuss.cocos2d-x.org/t/disabling-multitouch-in-android/9649/3), [here](http://discuss.cocos2d-x.org/t/single-touch-problem/17921/7) and [here](http://discuss.cocos2d-x.org/t/about-the-single-touch-by-android/17957). There was also an very similar PR #4654 that was closed but not merged.

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
@@ -68,6 +68,16 @@ public class Cocos2dxGLSurfaceView extends GLSurfaceView {
 
     private boolean mSoftKeyboardShown = false;
 
+    public boolean isMultipleTouchEnabled() {
+        return mMultipleTouchEnabled;
+    }
+
+    public void setMultipleTouchEnabled(boolean multipleTouchEnabled) {
+        this.mMultipleTouchEnabled = multipleTouchEnabled;
+    }
+
+    private boolean mMultipleTouchEnabled = true;
+
 
     // ===========================================================
     // Constructors
@@ -218,6 +228,9 @@ public class Cocos2dxGLSurfaceView extends GLSurfaceView {
         switch (pMotionEvent.getAction() & MotionEvent.ACTION_MASK) {
             case MotionEvent.ACTION_POINTER_DOWN:
                 final int indexPointerDown = pMotionEvent.getAction() >> MotionEvent.ACTION_POINTER_INDEX_SHIFT;
+                if (!mMultipleTouchEnabled && indexPointerDown != 0) {
+                    break;
+                }
                 final int idPointerDown = pMotionEvent.getPointerId(indexPointerDown);
                 final float xPointerDown = pMotionEvent.getX(indexPointerDown);
                 final float yPointerDown = pMotionEvent.getY(indexPointerDown);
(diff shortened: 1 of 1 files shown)


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
```diff
@@ -68,6 +68,16 @@ public void setSoftKeyboardShown(boolean softKeyboardShown) {
...
 
+    public boolean isMultipleTouchEnabled() {
+        return mMultipleTouchEnabled;
+    }
+
+    public void setMultipleTouchEnabled(boolean multipleTouchEnabled) {
+        this.mMultipleTouchEnabled = multipleTouchEnabled;
+    }
+
+    private boolean mMultipleTouchEnabled = true;
```
minggo: it is better to define the variable at the beginning of the class as other variables

minggo:
I think should handle `ACTION_MOVE` too, or it will also send multi touch move events.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
```diff
@@ -245,6 +256,24 @@ public void run() {
...
+                    // handle only touch with id == 0
+                    for (int i = 0; i < pointerNumber; i++) {
+                        if (ids[i] == 0) {
+                            final int[] idsMove = new int[]{0};
+                            final float[] xsMove = new float[]{xs[i]};
+                            final float[] ysMove = new float[]{ys[i]};
+                            this.queueEvent(new Runnable() {
+                                @Override
+                                public void run() {
+                                    Cocos2dxGLSurfaceView.this.mCocos2dxRenderer.handleActionMove(idsMove, xsMove, ysMove);
```
minggo, line 269: please fix the indent

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
```diff
@@ -245,6 +256,24 @@ public void run() {
...
+                                }
+                            });
+                            break;
+                        }
+                    }
+                    break;
+                }
```
minggo, line 282: to make codes more clean, it is better to use`if else` here.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
pyrosphere: I'm not sure what is wrong in the indentation here, can you help me?

minggo: Oh, it looks wrong when looking in `Files changed`. Sorry about it.

Issue #16784, opened 2016-11-01
Missing Chinese characters when use GB2312 encoding TTF file 


zilongshanren:
When using Chinese TTF to rendering text, if the text contains some special Chinese characters which aren't inside the collection of GB2312, the entire sentence will be truncated when translated from UTF16 to GB2312.

This behavior only happen in Linux, iOS and MaxOS X. Because these platform use iconv to do the converting.

PR #16832, opened 2016-11-11
fix missing chinese characters when using ttf label

zilongshanren:
https://github.com/cocos2d/cocos2d-x/issues/16784


minggo:
@ericduck40 what's the meaning?

Issue #16779, opened 2016-10-31
EditBox doesn't show text when editing in Mac version (but no problem in iOS)


amuseamuse:
- cocos2d-x version: 3.13 , 3.13.1
- devices test on: MacBook Pro 13 inch 2010 and iMac 27 inch 2010 (Mac OS Sierra)
    (Note: this issue doesn't appear before the Mac OS Sierra I upgraded from former os, and the issue appears after I completely reinstall Mac OS Sierra)
- developing environments
    Mac (but not ios) in Mac OS Sierra

Steps to Reproduce:
1 .  create a EditBox

    bool HelloWorld::init(){
    EditBox* gameIdEditBox=EditBox::create(Size(480,180),Scale9Sprite::create());
    this->addChild(gameIdEditBox);
    gameIdEditBox->setInputMode(ui::EditBox::InputMode::SINGLE_LINE);
    gameIdEditBox->setReturnType(ui::EditBox::KeyboardReturnType::SEND);
    gameIdEditBox->setAnchorPoint(Vec2::ZERO);
    gameIdEditBox->setText("abc");
    return true;
    }

![e01](https://cloud.githubusercontent.com/assets/11567495/19843023/a7fa6c24-9f59-11e6-93f5-0ab3817ba562.png)

it can show the text when it is not editing

2 . when click the edit box and start to edit, the EditBox doesn't show the editing text:

![e02](https://cloud.githubusercontent.com/assets/11567495/19843059/e5a1c586-9f59-11e6-801c-65e65902e359.png)

3 . when edit finished, the text appear again, and it seems the text can change normally:

![e03](https://cloud.githubusercontent.com/assets/11567495/19843076/0ab73f40-9f5a-11e6-9bc1-e738358de392.png)

Note 1: I tried to solve the problem by setNeedsLayer in CCApplication-mac.mm

    int Application::run()
    {
        initGLContextAttrs();
        if(!applicationDidFinishLaunching())
        {
            return 1;
        }
    
        long lastTime = 0L;
        long curTime = 0L;
    
        auto director = Director::getInstance();
        auto glview = director->getOpenGLView();
        NSWindow* window = glview->getCocoaWindow();
        [[window contentView] setWantsLayer:YES];
        .
        .
        .

This time the EditBox can show normally but it contains a lot of OpenGL errors:

Before Editing:
[...]

PR #16833, opened 2016-11-11
fix glfw move window to different monitor

zilongshanren:
https://github.com/cocos2d/cocos2d-x/issues/16779


minggo:
why modify like this?

zilongshanren:
@minggo  it's a temp hack, I'm still digging into the internals to find the solution. So don't merge this PR at the moment.

PR #16842, opened 2016-11-14
fix editbox can't input on macos 10.12

zilongshanren:
Fix issue: https://github.com/cocos2d/cocos2d-x/issues/16779

### DIFF
--- a/cocos/ui/UIEditBox/Mac/CCUIEditBoxMac.mm
+++ b/cocos/ui/UIEditBox/Mac/CCUIEditBoxMac.mm
@@ -135,6 +135,8 @@
 - (void)openKeyboard
 {
     [self.window.contentView addSubview:self.textInput];
+    //FIXME: it will cause a few OpenGL error once, should be fix when glfw version upgraded.
+    self.window.contentView.wantsLayer = YES;
     if (![self.textInput isKindOfClass:[NSTextView class]]) {
         [self.textInput becomeFirstResponder];
     }else {
--- a/external/config.json
+++ b/external/config.json
@@ -1,5 +1,5 @@
 {
-    "version":"v3-deps-113",
+    "version":"v3-deps-114",
     "zip_file_size":"135533806",
     "repo_name":"cocos2d-x-3rd-party-libs-bin",
     "repo_parent":"https://github.com/cocos2d/",


cocos/ui/UIEditBox/Mac/CCUIEditBoxMac.mm
```diff
@@ -135,6 +135,8 @@ - (NSWindow *)window
 - (void)openKeyboard
 {
     [self.window.contentView addSubview:self.textInput];
+    //FIXME: it will cause a few OpenGL error once, should be fix when glfw version upgraded.
```
dumganhar, line 138: errors? fixed? is upgraded?

zilongshanren: There is a few OpenGL errors, but it doesn't affect the rendering and these error only occurs once. I think the problem should be resolved in the glfw upstream. But I can't handle it properly at the moment. This is a workaround, but at least it works well except for the error message.

zilongshanren: @dumganhar 
In the future, when glfw fixed the EditBox input issue on mac os, we should remove this hack code.

zilongshanren: @dumganhar 
The glfw patch is here https://github.com/cocos2d/cocos2d-x-3rd-party-libs-src/blob/v3/contrib/src/glfw/fix-mac-os-10-12.patch

minggo: I think you can send a patch to GLFW to discuss with them.

zilongshanren: @minggo  
Already opened a ticket in  https://github.com/glfw/glfw/issues/892  But it seems not get active response.

minggo: Looks good to me. I will merge it, but i think it is better to create an issue in cocos2d-x that revert the codes when GLFW fixes the issue.

rockingdice: can you share a working GLFW library file?

rockingdice: well I compiled the GLFW library by myself but, it is still not working yet. I also try the combination of cocos2d-x side  changes and glfw changes.

zilongshanren: @rockingdice
You could got a GLFW library from here: https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin

PR #16844, opened 2016-11-14
Fixed Android.mk file name not match real file.

x-tinkerer:


minggo:
https://github.com/cocos2d/cocos2d-x/blob/v2/extensions/Android.mk#L106 it is `CCTimeLine.cpp`, i don't know why the PR can be generated.

PR #16852, opened 2016-11-16
std::map -> std::unordered_map for int & string key type.

dumganhar:


dumganhar:
It seems that Bindings-generator could not recognize `std::unordered_map`.

```
details = "no type named 'unordered_map' in namespace 'std'">
```

:(

ricardoquesada:
this is a nice change. could you resolve the conflicts?

vosafonov:
@dumganhar, it seems that you just forgot `#include <unordered_map>` in some of files.

PR #16856, opened 2016-11-16
unscheduleAll() does not unschedule scheduled with performFunctionInCocosThread.

DavidDeSimone:
Fixing issue where Scheduler::unscheduleAll() would not unschedule functions scheduled with Scheduler::performFunctionInCocosThread.

If I were to schedule a function via Scheduler::performFunctionInCocosThread, and then call unscheduleAll, the scheduled function will still be called. 

This causes issues for my team, as we call Director::getInstance()->restart in cocos2d-js regularly,  and we can have crashes when our custom functions scheduled with Scheduler::performFunctionInCocosThread attempt to call back into the destroyed and recreated javascript context.

In general, it seems like a reasonable expectation that scheduleAll will unschedule functions scheduled with performFunctionInCocosThread.

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -629,11 +629,6 @@ void Scheduler::unscheduleAllWithMinPriority(int minPriority)
 #if CC_ENABLE_SCRIPT_BINDING
     _scriptHandlerEntries.clear();
 #endif
-    
-    {
-        std::unique_lock<std::mutex> lock(_performMutex);
-        _functionsToPerform.clear();
-    }
 }
 
 void Scheduler::unscheduleAllForTarget(void *target)
@@ -828,6 +823,12 @@ void Scheduler::performFunctionInCocosThread(const std::function<void ()> &funct
     _functionsToPerform.push_back(function);
 
     _performMutex.unlock();
+}
+
+void Scheduler::removeAllFunctionsToBePerformedInCocosThread()
+{
+    std::unique_lock<std::mutex> lock(_performMutex);
+    _functionsToPerform.clear();
 }
 
 // main loop
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -629,6 +629,11 @@ void Scheduler::unscheduleAllWithMinPriority(int minPriority)
 #if CC_ENABLE_SCRIPT_BINDING
     _scriptHandlerEntries.clear();
 #endif
+    
+    {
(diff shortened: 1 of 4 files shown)


dumganhar:
Good catch.

minggo:
I think performFunctionInCocosThread just use schedule to do some synchronization task between rendering thread and other thread. It is not a normal scheduler as others. May be we should provide another function to stop these tasks instead of doing it in unscheduleAll.

@ricardoquesada what's your opinion?

DavidDeSimone:
@minggo that makes sense to me. Current users of cocos2d-x may be depdendent on this behavior in their custom code, and we would not want to break compatibility.

I'm also curious what @ricardoquesada thinks.

ricardoquesada:
yes, correct, `unscheduleAll()` does not unschedule `performinCocosThread`, and I think it is the correct behavior.

`performInCocosThread` is part of the scheduler only because it makes sense from an implementation point of view, but not so much from a design point of view. The scheduler is run in the Cocos thread, so adding that function there, made sense... only a few lines of code.

So, the scope `unscheduleAll` is to unschedule all the functions that were scheduled with `scheduleXXX` functions (or at least in theory)

DavidDeSimone:
@ricardoquesada That all makes sense to me. With that in mind, do you think its worthwhile to instead have a function to unschedule all pending tasks queue'd up with performInCocosThread? If not, I will just close this PR.

ricardoquesada:
@DavidDeSimone yes, it makes sense to have a function that unschedules things added with `performInCocosThread`... 

I don't know what is the "reverse" of "perform", but ideally we should not use the word "unschedule" because that is reserved for the schedule/unschedule.
Perhaps "disconnect", or "remove" ?... like "removeFromCocosThread" ?

minggo:
`removeFromCocosThread ` sounds good, but to my opinion it is not obvious that it is removing tasks added by performInCocosThread.

dumganhar:
Would `removeFunctionsPerformedInCocosThread` be a better name?

DavidDeSimone:
I think I like either `removeFunctionsPerformedInCocosThread` or `removeFunctionsToBePerformedInCocosThread`

DavidDeSimone:
@minggo @ricardoquesada 

Updated PR based on review feedback. 
Added Scheduler::removeAllFunctionsToBePerformedInCocosThread. 
Added test case to Scheduler tests to test that this functionality works properly.

minggo:
looks good to me, thanks

DavidDeSimone:
Thanks for merging

PR #16857, opened 2016-11-16
Remove unnecessary return statement

Rubonnek:
This return statement is unnecessary since the same return statement can be found above in the `#else` directive.

### DIFF
--- a/cocos/platform/winrt/CCWinRTUtils.cpp
+++ b/cocos/platform/winrt/CCWinRTUtils.cpp
@@ -59,7 +59,6 @@ bool isWindowsPhone()
 #else
     return false;
 #endif
-    return false;
 }
 
 CC_DEPRECATED_ATTRIBUTE std::wstring CC_DLL CCUtf8ToUnicode(const char * pszUtf8Str, unsigned len /*= -1*/)


cocos/platform/winrt/CCWinRTUtils.cpp
```diff
@@ -59,7 +59,6 @@ bool isWindowsPhone()
 #else
     return false;
 #endif
-    return false;
```
minggo, line 62: it is needed or no value will return in `#if _MSC_VER >= 1900` and `if` sentence returns false

Rubonnek: Right, I meant to put it in the else branch of that [if statement](https://github.com/Rubonnek/cocos2d-x/commit/00fdf48752173ee849934096dccf98cfbfe66ef7). The return statement only becomes unnecessary on some cases.

minggo: yep, so could you please modify it to send a new PR?

Rubonnek: Glad to. I used a more descriptive title [this time](https://github.com/cocos2d/cocos2d-x/pull/16874).

minggo: thanks, i merged it

Issue #16876, opened 2016-11-19
CCEAGLView::layoutSubviews not calling on iOS 7 and and earlier


MrCapone:
- cocos2d-x version: 3.13.1 (git+88f6825c4840beea4f18ac7a9c80ac403137f48c)
- devices test on: iPod 5th gen iOS 7.1, iPod 4th gen iOS 6.1.6
- developing environments
   - Xcode version: 7.3.1

Steps to Reproduce:

1. run ccp-empty-test on iOS 7 or earlier

Issue appeared after this pull request https://github.com/cocos2d/cocos2d-x/pull/16394

PR #16877, opened 2016-11-19
fix for #16876

MrCapone:
fix for https://github.com/cocos2d/cocos2d-x/issues/16876
also call `[window makeKeyAndVisible];` after app init action seems more correct.

Tested on:
iPod 4th gen iOS 6.1.6
iPod 5th gen iOS 7.1
Simulator iPhone 4s iOS 8.3
iPad mini 2 iOS 9.3.5
iPhone 6 iOS 10.1.1


MrCapone:
The problem is more complicated than I expect. I close pull request for finding better solution.

Issue #16565, opened 2016-09-12
cocos2dx 3.12 and 3.13.1 device orientation


dirtyshadowss:
Hello,
when i create New Project with version 3.12, 3.13 and 3.13.1 i have this:

Landscape:
![06993cd2707c4c6d55b6774f7e4622618716550e](https://cloud.githubusercontent.com/assets/22153658/18443537/09746840-791e-11e6-8156-1bb89908eccb.PNG)
and Portrait:
![0f4502b69ccc2d7b8dc6d9226c79ef2cb0976edb](https://cloud.githubusercontent.com/assets/22153658/18443570/1a437ef4-791e-11e6-9a5e-a54ab4177e03.PNG)
My appdelegate.cpp is standart creating, only change 
designResolutionSize = cocos2d::Size(960, 640);
and
ResolutionPolicy::EXACT_FIT;

This issue is not fix the problem

[https://github.com/cocos2d/cocos2d-x/issues/16485]
- cocos2d-x version: 3.13.1
- devices test on:Iphone 4S IOS 7.0.4
- developing environments
  - NDK version: iOS 9.2
  - Xcode version: 7.2.1

ricardoquesada:
Either it is already fixed, or something is wrong with your example. I tried with v3 branch (HEAD) and this is the result after using EXACT_FIT AND 960x640 as designed resolution.

<img width="480" alt="screen shot 2016-09-14 at 5 13 49 pm" src="https://cloud.githubusercontent.com/assets/232330/18534333/c6167304-7a9e-11e6-9d4a-506b46194886.png">

dirtyshadowss:
Hello ricardo,
i don't have problem with iphone simulator with IOS 9.2 toо, the problem is IOS 7.0.3 on real device (iphone 4S). For me is no problem to upgrade my device to IOS 9.3 but when going time to production deployment whats should be minimum IOS system requirement (7.1, 8.0, 8.1).   Is  Cocos2d-x Runtime Requirements for IOS 7.0+ is right or must be highter?

ricardoquesada:
yep. I tested also on a real device: iPhone 4S + iOS 8.3 and it worked Ok.
As of Xcode 8, the minimum deployment target is iOS 8.

dirtyshadowss:
Thanks man,
[...]

PR #16879, opened 2016-11-20
fixes for iOS 7 and earlier

MrCapone:
1. `cocos2d::Director` `_invalid` variable must be set to `false` on first init to prevent unwanted return on `CCEAGLView-ios::layoutSubviews` when it called first time.
Improves pull request: https://github.com/cocos2d/cocos2d-x/pull/16394
Fixes issue: https://github.com/cocos2d/cocos2d-x/issues/16876
Also tested that case: https://github.com/cocos2d/cocos2d-x/issues/14276 to prevert regression.

2. Creation and setting the `GLView` should be done after creating the `RootViewController` and after `CCEAGLView` was added to view hierarchy. Otherwise, it can cause a problem when using the landscape orientation, because view will return wrong bounds size.
Improves pull request: https://github.com/cocos2d/cocos2d-x/pull/15792
Fixes issue: https://github.com/cocos2d/cocos2d-x/issues/16565

Tested on:
iPod 4th gen iOS 6.1.6
iPod 5th gen iOS 7.1
iPad mini 2th gen iOS 9.3.5
iPhone 6 iOS 10.1.1


minggo:
will you please modify other templates too？

MrCapone:
@minggo 
Should i just copy iOS templates from cpp to js and lua? Because iOS templates contain a lot of changes and fixes and lua/js templates are pretty outdated.

UPD
I did so, but I didn't touch copyrights.

templates/js-template-default/frameworks/runtime-src/proj.ios_mac/ios/AppController.mm
```diff
@@ -140,5 +140,125 @@ - (void)dealloc {
 }
 
 
+@end#import "AppController.h"
```
minggo: should remove old codes then add new codes?

MrCapone:
@minggo 
I reset broken commit. Then push valid commits. Also i make some tests:

- create blank cpp project with new templates. Build passed, runs on iOS 10.1.
- create blank js project with new templates. Build passed, runs on iOS 10.1.
- create blank lua project with new templates. Build crash on `/lib/runtime/FileServer.cpp` but i applied this fix https://github.com/cocos2d/cocos2d-x/pull/16816 (for test build only) and build passed successfully, runs on iOS 10.1.

minggo:
thanks

PR #16884, opened 2016-11-21
fixed #16871: Material second shader texture will be lost when removeUnusedTextures is invoked.

### LINKED ISSUE #16871: Material second shader texture will be lost,when removeUnusedTextures
var mat = cc.Material.createWithFilename("demo.material");
sprite3D.setMaterial(mat);
cc.TextureCache.getInstance().removeUnusedTextures();

what is happen?
sprite3D will lost the u_sampler0 textures. why Material not save the texture(u_sampler0)?

demo.material file:
material demo
{
	technique
	{
		pass
		{
			shader
			{
				vertexShader = ccShader_3D_PositionTex.vert
				fragmentShader = ccShader_3D_ColorTex.frag
				sampler u_sampler0
				{
					path = daytime_lawn.png
					wrapS = REPEAT
					wrapT = REPEAT
					minFilter = NEAREST
					magFilter = NEAREST
				}
			}
		}
	}
}

dumganhar:


dumganhar:
@ricardoquesada , could you please take a look at this Pull Request? Thanks.

cocos/renderer/CCGLProgramState.cpp
```diff
@@ -165,8 +175,24 @@ void UniformValue::setTexture(GLuint textureId, GLuint textureUnit)
...
     _value.tex.textureId = textureId;
     _value.tex.textureUnit = textureUnit;
+    _value.tex.texture = nullptr;
+    _type = Type::VALUE;
+}
+
+void UniformValue::setTexture(Texture2D* texture, GLuint textureUnit)
+{
+    CCASSERT(texture != nullptr, "texture is nullptr");
+
```
ricardoquesada, line 185: a check could be added here for the texture like:
```
if (texture != _value.tex.texture) {
  RELEASE...();
  RETAIN...();
}

**Comment by ricardoquesada** [comment]
@dumganhar looks good to me. I've just added a comment.

having a test case that reproduces the issue would be great however.

File: cocos/renderer/CCGLProgramState.cpp (hunk 1 above)
Comment (dumganhar) [reply cocos/renderer/CCGLProgramState.cpp:185]: OK, I will do it. :)

**Comment by dumganhar** [comment]
@ricardoquesada , the second commit is the test case for this issue.
I reused a test case in Sprite3DTest. Just add a line `Director::getInstance()->getTextureCache()->removeUnusedTextures();` after creating a Sprite3D. 
You will find the plane in that test case is missing texture.

**Comment by dumganhar** [comment]
@ricardoquesada, I fixed the problem you commented and added more comments of releasing resource order. Please review again.

File: tests/cpp-tests/Classes/Sprite3DTest/Sprite3DTest.cpp [hunk 2]
```diff
@@ -786,6 +786,7 @@ void Sprite3DEffectTest::addNewSpriteWithCoords(Vec2 p)
     material->setTechnique("outline_noneskinned");
     sprite->setMaterial(material);
     sprite->setScale(6.f);
+    Director::getInstance()->getTextureCache()->removeUnusedTextures();
```
dumganhar, line 789: This line is the  test case for this issue.

dumganhar:
@ricardoquesada, please review again. Thanks.

cocos/renderer/CCGLProgramState.cpp
ricardoquesada: could it be possible that someone calls:

```c++
setTexture(someTexture, 0);
setTexture(someTexture, 1);
```
because, if that is the case, the 2nd call won't do anyting

tests/cpp-tests/Classes/Sprite3DTest/Sprite3DTest.cpp
ricardoquesada: ok. thanks. what about moving that test (or adding a new one) in Texture2d test file?

cocos/renderer/CCGLProgramState.cpp
dumganhar: Currently, only `GLProgramState` class uses `UniformValue`. Developers could not access anything of `UniformValue`. So `UniformValue::setTexture` is only used by `GLProgramState` now.
Please refer the logic of `GLProgramState::setUniformTexture`:

```c++
void GLProgramState::setUniformTexture(const std::string& uniformName, Texture2D *texture)
{
    CCASSERT(texture, "Invalid texture");
    auto v = getUniformValue(uniformName);
    if (v)
    {
        if (_boundTextureUnits.find(uniformName) != _boundTextureUnits.end())
        {
            v->setTexture(texture, _boundTextureUnits[uniformName]); //James:  Get the cached texture unit id here
        }
        else
        {
            v->setTexture(texture, _textureUnitIndex);
            _boundTextureUnits[uniformName] = _textureUnitIndex++;
        }
    }
    else
    {
        CCLOG("cocos2d: warning: Uniform not found: %s", uniformName.c_str());
    }
}
```

If there're several texture samplers, each texture sampler will have an unique name which is also associated with an unique `texture unit id`. 

Therefore,  I guess
```c++
UniformValue::setTexture(texture, 0);
UniformValue::setTexture(texture, 1);
```
will never happen.

tests/cpp-tests/Classes/Sprite3DTest/Sprite3DTest.cpp
dumganhar: This is an issue while using Sprite3D. IMO, moving it to Texture2DTest is probably not a good idea. 
As I know, one of the purpose of test case is for easily find out issues before releasing a new version. And our QA team will check the display of every test case. When he/she checks `Sprite3DEffectTest` and found the texture is missing, he/she will report an issue to us.
Therefore,
* If a new test case is needed indeed, would it be better to add it in Sprite3DTest or BugTests?
* New test case will contain duplicated code, probably all the same as `Sprite3DEffectTest` except adding a new line of `removeUnusedTextures`.

@ricardoquesada , anyway, I will add it if you wish to. :)

PR #16885, opened 2016-11-21
non-copyable Ref and related fixes

yahont:
- thread-safe refcounter for Ref
- copying of Ref is explicitly forbidden
- unneeded inheritance  of Acceleration from Ref removed
- copying of BaseData in cocostudio sources fixed


minggo:
cocos2d-x is not thread safe, so no need to make Ref thread safe. it will slow down the speed

yahont:
OK. Sorry, I've got confused by thread-safe leak detector in CCRef.cpp.
I'll revert atomic refcounter and resubmit only nob-copyable Ref with related fixes.

minggo:
> copying of Ref is explicitly forbidden

Why? It will break compatibility.

yahont:
I mean the default copy constructor and assignment operator are explicitly deleted.
Two main reasons:
1) default copy constructor copies ref. counter
2) it is safe taking into account cocos's memory management model (autorelease, etc.) 
Have a look at the code, you'll see a couple of places where it was abused.
One still can declare copy constructors and assignment operators in derived classes, but the default one cannot be used.

minggo:
Sounds reasonable, please fix the compiling error. The error seems to in binding codes.

yahont:
I couldn't reproduce the compiling error. And I couldn't find Jenkins' job rules to make sure I used the same setup.

yahont:
The error seems to be auto-generated binding files in lua (cocos/scripting/lua-bindings/auto/). Can you advise me how to regenerate those files, please?

yahont:
I cannot reproduce the building error either on Windows with VS2015 or on Mac with XCode 8. I've built it all using both cmake and cocos python script. No errors.
Can you tell me the exact configuration for the Jenkins' jobs, please? Then I hope I'll be able to reproduce the errors and fix them.

PR #16897, opened 2016-11-24
fix: Scale9Sprite test case 18

ricardoquesada:
Bug calculating rotated frames.
the width was being used instead of the height and vice-versa

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -529,19 +529,20 @@ void Sprite::updatePoly()
         //
         // textCoords Data: Y must be inverted.
         //
-        const float u0 = oox + osw * 0;
-        const float u1 = oox + osw * cx1;
-        const float u2 = oox + osw * cx2;
-        const float v0 = ooy + osh - (osh * cy1);
-        const float v1 = ooy + osh * (1 - cy2);
-        const float v2 = ooy + osh * 0;
-
         const float w0 = osw * cx1;
         const float w1 = osw * (cx2-cx1);
         const float w2 = osw * (1-cx2);
         const float h0 = osh * cy1;
         const float h1 = osh * (cy2-cy1);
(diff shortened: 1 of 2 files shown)


minggo:
please send to v3.14 branch too, thanks

ricardoquesada:
@minggo 
oops, yes. here it is:
https://github.com/cocos2d/cocos2d-x/pull/16898

dumganhar:
This pull request fixes `UIS9_TexturePacker`.
But `UIS9FrameNameSpriteSheetCropped` and `UIS9FrameNameSpriteSheetCroppedRotated ` is still not fixed.

ricardoquesada:
@dumganhar yes, correct. I'm still working on that. I thought it was the same bug, but it is not :(

ricardoquesada:
@dumganhar so, I almost fixed test cases 8 and 7. There is an offset of 2 pixels... I might be doing the math incorrectly somewhere. I'll fix it after holidays (this monday).

I did the math manually, and it matches what it happens on the code... but as I said, perhaps the bug is somewhere else.
screen shot 2016-11-24 at 1 03 26 am

ricardoquesada:
I'll merge it and on Monday I'll send a new PR with this "2 pixel" issue bug.

ricardoquesada:
PR for v3.14: https://github.com/cocos2d/cocos2d-x/pull/16902

minggo:
Currently, the codes is complex. I think we can use matrix to simplify it. Don't modify texture coordinate and vertex coordinate manually, just use model matrix to do it. It will make things more clear. Of course we can optimize matrix operation as we did in `Node::getNodeToParentTransform()`.

ricardoquesada:
yes, agreed. the code is complex.  good idea about rotating the texture, although I'm not sure if that would be feasible... but for sure, the code should be simplified. too many `if(_rotated)` scattered in different files.

PR #16905, opened 2016-11-26
Update AssetsManagerEx.cpp

Dauch:
Let's suppose then player updated a half file in version 0.2 and turn off the game, and then the
hot update server had already updated this file with anthor version (said version 0.3).
And the player contiune to update this file. It will skip the updated bytes and contiune to download
the rest part. but this file half is ver0.2, half is ver0.3, this file will be damage.
It must cause problem.

### DIFF
--- a/extensions/assets-manager/AssetsManagerEx.cpp
+++ b/extensions/assets-manager/AssetsManagerEx.cpp
@@ -621,6 +621,12 @@ void AssetsManagerEx::startUpdate()
             {
                 Manifest::AssetDiff diff = it->second;
 
+                // remove downloaded tmp file. it found extName by CCDownloader.cpp:71
+                // 假设玩家在版本0.2更新了一半该文件，然后关掉热更新，服务器更新了几版这个文件，
+                // 下次玩家再次更新该文件，由于已经更新了一半缓存到缓存文件，更新时会更新剩下的文件大小然后进行拼接，
+                // 完成后重命名回来，导致该文件损坏。
+                _fileUtils->removeFile(_storagePath + diff.asset.path + ".tmp");
+
                 if (diff.type == Manifest::DiffType::DELETED)
                 {
                     _fileUtils->removeFile(_storagePath + diff.asset.path);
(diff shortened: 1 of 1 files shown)


extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -621,6 +621,12 @@ void AssetsManagerEx::startUpdate()
             {
                 Manifest::AssetDiff diff = it->second;
 
+                // remove downloaded tmp file. it found extName by CCDownloader.cpp:71
+                // 假设玩家在版本0.2更新了一半该文件，然后关掉热更新，服务器更新了几版这个文件，
+                // 下次玩家再次更新该文件，由于已经更新了一半缓存到缓存文件，更新时会更新剩下的文件大小然后进行拼接，
+                // 完成后重命名回来，导致该文件损坏。
```
minggo: please don't use Chinese characters

pandamicro:
Hi, this issue will be solved by my upgrade to AssetsManagerEx in this PR

https://github.com/cocos2d/cocos2d-x/pull/17115

Here is some brief informations about the upgrade:

1. **IMPORTANT**: Separate storage folder for updating and local cached version, so that the failing in update won't damage the current local version. The updating version will be overwritten to the local cache folder only when the update is fully successful.
2. **IMPORTANT**: Bound `Downloader` to JS and support file download task, data task is not supported
3. **IMPORTANT**: Decompress process is moved before file success notification, downloading a zip file is considered successful only after the decompress process is successfully done.
4. Support `size` property in manifest file for collecting total byte size of the update before it's started.
5. Fix download resuming on Android and some edge condition bugs.
6. Support customized version compare API between manifests: setVersionCompareHandle
7. Support verification callback after file downloaded: setVerifyCallback
8. Support setting maximum concurrent tasks count: setMaxConcurrentTask
9. Ensure saving download progression to temporary manifest
10. Delay folder creation right before file download request to ensure the path is correct
11. Ensure reading the correct manifest in app package

These modifications have already been tested and published in cocos creator, I need a little more time to migrate them to cocos2d-x repo. Sorry for the delay.

In Cocos Creator's hot update sample, I also provided a manifest generation script, feel free to try out:

https://github.com/cocos-creator/tutorial-hot-update/blob/master/version_generator.js

pandamicro:
@minggo you can close this PR now, @Dauch Thanks anyway for the contribution

Dauch:
That's all right, good to know you fix those problem. @pandamicro

PR #16906, opened 2016-11-26
Update AssetsManagerEx.cpp

Dauch:
if load succeed, when it come to startUpdate(), it may execute the condition:
 "if (_tempManifest->isLoaded() && _tempManifest->versionEquals(_remoteManifest))"
 and download all the resource which we don't need to.

### DIFF
--- a/extensions/assets-manager/AssetsManagerEx.cpp
+++ b/extensions/assets-manager/AssetsManagerEx.cpp
@@ -137,7 +137,9 @@ void AssetsManagerEx::initManifests(const std::string& manifestUrl)
         if (_tempManifest)
         {
             _tempManifest->parse(_tempManifestPath);
-            if (!_tempManifest->isLoaded() && _fileUtils->isFileExist(_tempManifestPath))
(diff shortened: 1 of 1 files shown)


extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -137,7 +137,9 @@ void AssetsManagerEx::initManifests(const std::string& manifestUrl)
         if (_tempManifest)
         {
             _tempManifest->parse(_tempManifestPath);
-            if (!_tempManifest->isLoaded() && _fileUtils->isFileExist(_tempManifestPath))
+            // 如果成功加载该文件，下面进行startUpdate()更新时会导致执行 if (_tempManifest->isLoaded() && _tempManifest->versionEquals(_remoteManifest))
+            // 下载所有该文件清单的文件。
```
minggo: please don't use Chinese characters

pandamicro:
This won't happen in the new design, please refer to the explanation in https://github.com/cocos2d/cocos2d-x/pull/16905#issuecomment-285287675

PR #16907, opened 2016-11-26
No default copy for ref

yahont:
Submitting a new PR, in attempt to fix building errors


yahont:
The PR was created to reveal a possible bug in Jenkins' job building #16885, but the build is still failing. I withdraw the request as a spare one

Issue #16846, opened 2016-11-14
Add async file utils to cocos2d-x


DavidDeSimone:
Currently in cocos2d-x and cocos2d-js, we have functions like FileUtils::getStringFromFile, and jsb.fileUtils.getStringFromFile for interacting with the file system. These functions are synchronous, and performed on the main thread. This can be detrimental to performance. It can cause hitching and frame drops while loading data from the filesystem.

I suggest that we create asynchronous versions of these functions, that each take a callback. The operation will be completed off thread, and then call the callback on the main thread once complete. 

I think the best mechanism for managing the threading is to iterate on CCThreadPool, and pull that feature out of experimental, and bring it into the engine proper. 

I have been iterating on CCThreadPool, and have an iteration I would like to share for consideration. It can be found here: https://github.com/DavidDeSimone/cocos2d-x/blob/async-file-utils/cocos/platform/CCThreadPool.cpp. I have made major changes to this file that I would be happy to discuss/iterate on with other developers involved. 

I'd be curious what @dumganhar thinks of all of this.

liuyi:
good job! I indeed need this feature.

DavidDeSimone:
Update on this:

I have written async versions of each of the basic file operations in CCFileUtils, each taking a callback, and each done off thread. I am currently in the process of writing tests for all of this. I also need to make sure that the auto binding generator works properly for the new functions I've created. I hope to have a PR open next week for this feature.

DavidDeSimone:
Closing this issue as the corresponding PR was merged.

PR #16911, opened 2016-11-28
Adding Async FileUtils to cocos2d-x

DavidDeSimone:
Following through on https://github.com/cocos2d/cocos2d-x/issues/16846

This adds "async" versions of the major FileUtil functions, that will be performed off thread, and callback to the main thread when complete with their result. 

Important Notes:

- I used the AsyncTaskPool under the task type IO.
- In order to keep the details of the async mechanism self contained, I pulled the functionality to a function called performOperationOffthread. This way if we ever want to change the mechanism from AsyncTaskPool to another threading mechanism, it should be easy.
- While FileUtils is not thread safe, I was able to work around this by calling "fullPathForFilename" on the main thread before calling the delegate functions. 
- I attempted several template meta-programming solutions to this problem before settling on the submitted approach. All of the template solutions were too complex, and non readable in my opinion.
- You can see I am using a number of value types instead of const-references. This is intentional as we are moving this data between threads, and I want to leverage move semantics.


DavidDeSimone:
It looks like the auto binding generator doesn't like that I have all the callbacks using r-value references either. In fact, it looks like the auto generator doesn't handle taking non-pointer/booleans in lambda callbacks well.

DavidDeSimone:
After talking with some colleagues, it seems the best way to fix the issue of data copying would be to have the async functions take their data by value instead of by reference. This way a user who wants to avoid a copy can just std::move their data at the function call.

DavidDeSimone:
I apologize for the large number of commits I have made as part opening this PR. I have been having trouble getting the binding generator to work lately, and I'm not set up to iterate on certain compilers (like Visual Studio 2013).

So most of these commits have been in an effort to prevent copying data unnecessarily via move operations. I was very close to getting everything working, however it looks like Visual Studio 2013 does not support many of the tricks I was employing to move variables into lambdas for offthread execution. For the time being, I am going to let this PR sit in it's current state and let the maintainers examine it and comment. I still believe there is value merging this PR in it's current state, as file I/O on the main cocos thread will be detrimental to a smooth loading experience.

DavidDeSimone:
@minggo What do you think of this PR? is this something worth adding to cocos2d-x?

minggo:
Thanks, it looks good. Could you please add comments for all these new functions. Especially mentioned that they are asynchronous.

DavidDeSimone:
@minggo Going to close this PR and open a new one that is squashed to keep the history clean.

PR #16916, opened 2016-11-29
Adds missing Scale9Sprite::initXXX overrided methods after refactoring Scale9 logic. Otherwise, `setupSlice9` will not be invoked.

dumganhar:
Some fixes for Scale9Sprite:

* Adds missing Scale9Sprite::initXXX overrided methods after refactoring Scale9 logic. Otherwise, `setupSlice9` will not be invoked. Test case is: `1. Node: Scene3D -> Description -> Scale9Sprite`
* Call Sprite::initWithFile(filename) in Scale9Sprite::initWithFile(filename) rather than Scale9Sprite::initWithFile(filename, Rect::ZERO). Otherwise content size will be zero which is incorrect. Test case: `52: UI -> GUI Dynamic Create Test -> Scale9Sprite -> background display is wrong without this fix`
* setCapInsets in updateCapInset only when it’s in slice mode.
* setCapInsets should reset _insetLeft, _insetTop, _insetRight, _insetBottom member variables.
* Remove `_previousCenterRectNormalized`, just updateCapInset while switching to SLICE mode.
* Fixes JSTest -> ExtensionTest -> EditBoxTest: no background of editbox.
* Avoids an unused `setCapInsets` invocation in setupSlice9.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -235,10 +235,10 @@ bool Sprite::initWithSpriteFrame(SpriteFrame *spriteFrame)
         return false;
     }
 
-    bool bRet = initWithTexture(spriteFrame->getTexture(), spriteFrame->getRect(), spriteFrame->isRotated());
+    bool ret = initWithTexture(spriteFrame->getTexture(), spriteFrame->getRect(), spriteFrame->isRotated());
     setSpriteFrame(spriteFrame);
 
-    return bRet;
+    return ret;
 }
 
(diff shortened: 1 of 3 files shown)


minggo:
@ricardoquesada please take a look, thanks

ricardoquesada:
@dumganhar thanks!

the PR looks Ok to me. should I merge it?

minggo:
I merged it. Thanks.
Yep, you can merge it next time.

PR #16931, opened 2016-11-30
Update AssetsManagerEx.cpp

Dauch:
If we met downloading project.manifest error, we call update again, we need to ```downloadManifest()``` again.


pandamicro:
Thanks, it's indeed an issue, but your modification may cause download manifest file repeatly.

I will change _updateState to FAIL_TO_UPDATE when failed to download project.manifest file

pandamicro:
This have been included in https://github.com/cocos2d/cocos2d-x/pull/17115

Thanks ~

PR #16947, opened 2016-12-01
update CCNode.cpp

fred1024:
The _transform property of CCNode would not be restored on setAdditionalTransform(nullptr)
causing in my project a rotation run away ...

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1786,6 +1786,7 @@ void Node::setAdditionalTransform(const Mat4* additionalTransform)
 {
     if (additionalTransform == nullptr)
     {
+        if(_additionalTransform)  _transform = _additionalTransform[1];
         delete[] _additionalTransform;
         _additionalTransform = nullptr;
     }


dumganhar:
Good finding. Could you add a test case for this issue? Thanks.

fred1024:
Sorry for answering so late ... I am not sure how i could make a testCase ... been reading some articles but not sure what is the correct way of doing that. If It can help here's an exemple class that I used. 

```c++
//
//  testAdditionalT.cpp
//

#include "testAdditionalT.h"

//--------------------------
//  CTOR / DTOR
//--------------------------

testAdditionalT::testAdditionalT(): Node()
{
    setScheduler( Director::getInstance()->getScheduler());
    this->scheduleUpdate();
    
    _sprTest = DrawNode::create();
    _sprTest->setPosition( Vec2(Director::getInstance()->getWinSize().width * 0.5f,
                                Director::getInstance()->getWinSize().height * 0.5f) );
    _sprTest->drawTriangle( Vec2(-300, 140), Vec2(300, 0), Vec2(-300, -140), Color4F(1.0f, 0.5f, 0.0f, 1.0f));
    addChild( _sprTest );
    
    _clock = 0.0f;
}


testAdditionalT::~testAdditionalT()
{
    unscheduleUpdate();
}


//--------------------------
//  METHODS
//--------------------------

testAdditionalT* testAdditionalT::create()
{
	testAdditionalT* obj = new testAdditionalT();
	obj->autorelease();
	return obj;
}


//--------------------------
//  OVERRIDES
//--------------------------

void testAdditionalT::update( float p_dt)
{
    _clock = fmodf(_clock + p_dt, M_PI * 2.0f);
    
    Mat4 matrix;
    matrix.setIdentity();
    matrix.rotateZ( _clock );
    
    _sprTest->setAdditionalTransform( matrix );
    
    // do stuff like render to a texture ....
    
    // why this line ? .. for the test i need cocos to compute the transformation matrix with the additional transform to trigger the problem
    // (thing that i do for the rendering to a texture)
    Mat4 matrix2 = _sprTest->getNodeToWorldTransform();

    // end render to the texture
    
    // After the rendering to a texture i need the sprite back to its original place
    _sprTest->setAdditionalTransform( nullptr );

    // without the fix I would have the drawNode rotating faster and faster instade of being static, a king of rotation run away,
}

```

if you have some more details informations about how to make a test case I'd be happy to do it

thanks

minggo:
@fred1024 you can add a test case in `tests/cpp-tests/Classes/NodeTest/NodeTest.cpp`

fred1024:
Sorry loads of work .. I'll do that as soon as i can ... thanks for ur feed back

minggo:
After dig into the issue. You are right, it is an obvious issue. So i will merge it.

Issue #16930, opened 2016-11-30
UILoadingBarTest is broken


dumganhar:
- cocos2d-x version: v3.14
- devices test on: macOS
- developing environments
   - NDK version: N/A
   - Xcode version: 8.1
   - VS version:N/A
   - browser type and version:N/A

Steps to Reproduce:

1. cpp-tests -> 52:Node: UI -> 1: GUI Dynamic Create Test -> 9: LoadingBar Test

2. UILoadingBarTest_Left:

The right direction loading bar is missing:

Current Result (Wrong):

![screen shot 2016-11-30 at 11 59 13](https://cloud.githubusercontent.com/assets/493372/20739323/750e5e76-b6f4-11e6-8582-a45bc5a8bb14.png)

Expected result:

<img width="959" alt="screen shot 2016-11-30 at 11 59 20" src="https://cloud.githubusercontent.com/assets/493372/20739327/82b29e8e-b6f4-11e6-8a6d-976e536f1d7f.png">

3. UILoadingBarTest_Left_Scale9 & UILoadingBarTest_Right is broken

Current Result (Wrong):

![screen shot 2016-11-30 at 12 00 36](https://cloud.githubusercontent.com/assets/493372/20739347/a4190400-b6f4-11e6-88e1-6bfdb8630700.png)

Expected result:
<img width="958" alt="screen shot 2016-11-30 at 12 00 41" src="https://cloud.githubusercontent.com/assets/493372/20739353/a9b94302-b6f4-11e6-8994-eded37d6e1d4.png">

4. UILoadingBarReloadTexture, Clicking `ToggleScale9`
After clicking `ToggleScale9` several times, the display in v3.14 will be wrong. And sometimes the texture is missing after clicking toggle many times.
![screen shot 2016-11-30 at 13 42 03](https://cloud.githubusercontent.com/assets/493372/20741165/6a13f5d0-b703-11e6-9cbd-ea105ae6858b.png)

5. UILoadingBarReloadTexture, Clicking `ChangeTexture`, the animation of progress is wrong:

![screen shot 2016-11-30 at 13 47 59](https://cloud.githubusercontent.com/assets/493372/20741249/f8be87fa-b703-11e6-9fb5-cf38a7ce48c4.png)

6. UILoadingBarIssue 12249: (Left: v3.13.1, Right: v3.14)
[...]

Issue #16928, opened 2016-11-30
The display of UISliderTest_scale9_State_Change is wrong


dumganhar:
- cocos2d-x version: v3.14
- devices test on: macOS
- developing environments
   - NDK version: N/A
   - Xcode version: 8.1
   - VS version:N/A
   - browser type and version:N/A

Steps to Reproduce:

1. cpp-tests -> 52:Node: UI -> 1: GUI Dynamic Create Test -> 7: Slider Test -> UISliderTest_Scale9_State_Change

Current Result (Wrong):

![screen shot 2016-11-30 at 11 46 27](https://cloud.githubusercontent.com/assets/493372/20739133/1ec8b756-b6f3-11e6-96dc-1f26af64276d.png)

Expected Result (in v3.13.1):

<img width="958" alt="screen shot 2016-11-30 at 11 46 19" src="https://cloud.githubusercontent.com/assets/493372/20739142/29f20e34-b6f3-11e6-8005-41f0409dc1e5.png">

dumganhar:
@ricardoquesada , please take a look at this issue.  It's also relative to Scale9Sprite.

dumganhar:
7: UISliderIssue12249Test is also broken.
In 3.13.1, we get:
<img width="960" alt="screen shot 2016-11-30 at 11 54 21" src="https://cloud.githubusercontent.com/assets/493372/20739230/e0ebfe74-b6f3-11e6-8cad-57d34dc863fa.png">

But in current v3.14 branch:

![screen shot 2016-11-30 at 11 54 29](https://cloud.githubusercontent.com/assets/493372/20739234/e9de8920-b6f3-11e6-8080-21f1e24d916f.png)

dumganhar:
Probably, the `LoadingBar issue` (https://github.com/cocos2d/cocos2d-x/issues/16930) is the same as this one.

ricardoquesada:
PR is here: https://github.com/cocos2d/cocos2d-x/pull/16949

dumganhar:
@ricardoquesada ,there are still two issues. I have posted it in your Pull Request.

PR #16949, opened 2016-12-02
Scale9Sprite / Slider / LoadingBar fixes

ricardoquesada:
Fixes issues #16930 #16928 

There were different bugs:

* cloning was not supported in the new API. Actually, the previous API was cloning inside the `init` method which was extremely confusing. Added `copyTo` in `Scale9Sprite`
* fix negative streching
* fix bug when calculating centerRect after the texRect was modified


ricardoquesada:
@dumganhar this PR should fix the slider/loadingbar issues. thanks.

dumganhar:
@ricardoquesada, thanks. I will test it.

dumganhar:
uiloadingbarreloadtexturebug

@ricardoquesada , I found this test still can't work correctly.

UPDATED:

Reproduce steps:

* Go to `6, UILoadingBarReloadTexture`
* Click `ToggleScale9` once
* Click `ChangeTexture` once
* Click `ToggleScale9` once again

Current result:

The purple progress bar will move as the gif picture I uploaded.

Correct result:

The left picture in the gif.

dumganhar:
Another issue is:
different-height

ricardoquesada:
@dumganhar thanks.

I'll try to fix the height... but it is mostly black magic. The thing is that it is using invalid capInsets... like super invalid capInsets, so the texture displayed in v3.13 is incorrect. In v3.14 is the correct one. 

The capInsets used for left,right,top,bottom are: 66.6, 66.6, 66.6, 66.6. And the texture size is 50x57.  I really don't know what should be the expected behavior when using invalid capInsets.

As an example, the rendered quad has the same dimensions both in v3.13 and v3.14 but the texture coordinates are different. in v3.13 they are going beyond the texture limits.

screen shot 2016-12-02 at 8 21 20 am

ricardoquesada:
@dumganhar regarding the other bug, fixed!
thanks.

dumganhar:
@ricardoquesada 

> so the texture displayed in v3.13 is incorrect. In v3.14 is the correct one.

Yes, I also checked it, and found that it's incorrect in v3.13. I think developer will not pass in an incorrect capInset all the time. Therefore, it's probably safe to correct the mistake.

> regarding the other bug, fixed!

Thank you, I have confirmed that they were fixed in this Pull Request.

So, feel free to merge this Pull Request. :)

ricardoquesada:
Thanks for the great testing!

Issue #16814, opened 2016-11-07
CCFileUtils' thread safety improvement


cezheng:
Since FileUtils is not designed as thread safe, when multiple threads are trying to resolve some paths in the file system, the cache table inside it might cause a crash. Currently we have been getting quite a lot crash logs caused by this.

A straight forward fix for this is to add a mutex but this will require a lot of change in the code, and will cause much overhead for programs that mainly use FileUtils in main thread.

There's a much easier solution to this(only change 2 lines): 

> Add a [`thread_local`](http://en.cppreference.com/w/cpp/keyword/thread_local) specifier for the `FileUtils::s_sharedFileUtils`

This makes each thread that uses FileUtils have their own instances.

If an app does not use FileUtils in multiple threads, then `getInstance` will only be called in main thread and only main thread will hold a FileUtils instance so it will behave exactly the same as the current implementation.

Please let me know what you think, I can create a PR right after. Thanks. @minggo

toreinkun:
some async method will create instance of FileUtils, such as TextureCache::addImageAsync. I think i need a clone method for FileUtils.

cezheng:
@panzhihao So strictly saying, the TextureCache::addImageAsync is calling Image::initWithImageFileThreadSafe which calls FileUtils::fullPathForFilename which is not thread safe at the moment, making `initWithImageFileThreadSafe` not quite standing up to its name

cezheng:
Here is a sample iOS crash log we collect from our app still in development

```
Thread 0 Crashed:
[...]

PR #16953, opened 2016-12-02
make _fullPathCache of FileUtils thread-safe

cezheng:
Adopted a Read-Write SpinLock, since
* Operations on _fullPathCache don't take too long
* Performance for single-threaded apps won't be affected too much

I have made a simple test program, and have the following result

* When thread contention isn't heavy(5 threads reading & writing at the same time), a SpinLock can be 7 times faster than a std::mutex, about 5 times faster than a pthread_rwlock
* When thread contention gets heavy(30 threads reading & writing at the same time), a SpinLock gets a bit slower than pthread_rwlock, but still a little faster than a std::mutex
* If we have more than 30 threads in contention, SpinLock will become slower than a std::mutex at some point

But most of the time, there won't be many threads in contention for the FileUtils' _fullPathCache, I guess 99% of the time there will be no contention, and 99% of the 1% when there is contention, there are only 2 threads, so a SpinLock will be ideal.

I only tested on iOS devices & Mac, I will shared my test program later. @minggo @dumganhar

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -43,6 +43,44 @@ THE SOFTWARE.
 
 NS_CC_BEGIN
 
+SpinRWLock::SpinRWLock(): _state({false, 0}) {}
+
+void SpinRWLock::readLock() {
+    State newState {false, 0}, oldState = _state.load(std::memory_order_acquire);
+    do {
+        oldState.writeFlag = false; // if writing, then wait
+        newState.readCount = oldState.readCount + 1;
+    } while (!_state.compare_exchange_weak(oldState, newState, std::memory_order_acquire, std::memory_order_relaxed));
+}
+
+void SpinRWLock::readUnLock() {
+    State newState, oldState = _state.load(std::memory_order_acquire);
+    do {
+        newState = oldState;
+        --newState.readCount;
+    } while (!_state.compare_exchange_weak(oldState, newState, std::memory_order_release, std::memory_order_relaxed));
+}
+
+void SpinRWLock::writeLock() {
+    State newState {true, 0}, oldState = _state.load(std::memory_order_acquire);
+    // set writeFlag first to hold off incoming readers
+    do {
+        newState.readCount = oldState.readCount;
+        oldState.writeFlag = false; // if writing, then wait
+    } while (!_state.compare_exchange_weak(oldState, newState, std::memory_order_acquire, std::memory_order_relaxed));
+    // wait for all ongoing readers
+    do {
+        newState = _state.load(std::memory_order_acquire);
+    } while (newState.readCount);
+}
+
+void SpinRWLock::writeUnlock() {
+    State newState {false, 0}, oldState = _state.load(std::memory_order_acquire);
+    do {
+        newState.readCount = oldState.readCount;
+    } while (!_state.compare_exchange_weak(oldState, newState, std::memory_order_release, std::memory_order_relaxed));
+}
+
 // Implement DictMaker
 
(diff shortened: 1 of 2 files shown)


cezheng:
Linking the original issue: https://github.com/cocos2d/cocos2d-x/issues/16814

cezheng:
Gosh, GCC don't support <stdatomic.h>...

cezheng:
Replaced <stdatomic.h> stuff with std::atomic, see if it compiles with GCC

cocos/platform/CCFileUtils.cpp
```diff
@@ -43,6 +43,44 @@ THE SOFTWARE.
...
+    } while (newState.readCount);
+}
+
+void SpinRWLock::writeUnlock() {
+    State newState {false, 0}, oldState = _state.load(std::memory_order_acquire);
+    do {
+        newState.readCount = oldState.readCount;
+    } while (!_state.compare_exchange_weak(oldState, newState, std::memory_order_release, std::memory_order_relaxed));
+}
+
```
DavidDeSimone, line 83: This class seems generic enough that maybe we would want to pull it out into it's own file, so it can be more easily shared across the engine. I'd be curious what other people think.

cocos/platform/CCFileUtils.cpp
```diff
@@ -603,7 +641,9 @@ bool FileUtils::init()
 
 void FileUtils::purgeCachedEntries()
 {
+    _cacheLock.writeLock();
     _fullPathCache.clear();
+    _cacheLock.writeUnlock();
```
DavidDeSimone, line 646: Do you think it would be worthwhile to write a lock_guard equivalent for writeLock and readLock that uses RAII? Something like 

`WriteLockGaurd lock(_cacheLock); ReadLockGaurd lock(_cacheLock);` ?

DavidDeSimone:
Thank you for your work on this. I've left some inline comments, and have one additional comment.

So this PR makes accessing the full path cache thread safe, but over all FileUtils is still not completely thread safe. It looks like we do not guard on access to the searchPathCache or the searchResolutionOrder. This gives me concern for letting users out in the wild use FileUtils in a multithreaded manner.

cocos/platform/CCFileUtils.cpp
cezheng: Yeah, but I'm not sure which is the best place to put the file so I'd be expecting some comments from the maintainers.

cocos/platform/CCFileUtils.cpp
cezheng: Sounds good

cocos/platform/CCFileUtils.cpp
minggo: I think it is better to create a new file in cocos/base.

cocos/platform/CCFileUtils.cpp
minggo: Sounds good.

minggo:
@DavidDeSimone FileUtils is not thread safe, it will be a performance issue if make FileUtils to thread safe.

minggo:
@cezheng any update?

cezheng:
@minggo Sorry I'm recently busy with relocation & do not have free time to update this PR. Maybe a few weeks later.

minggo:
Got it, thanks.

cezheng:
BTW, this PR is merged into the fork used by one of my previous employer's project & has been working fine ever since.

minggo:
Yep, the codes is fine, just adjust the location of the codes to make it useful for other developers as commented above.

PatriceJiang:
Another [PR](
https://github.com/cocos2d/cocos2d-x/pull/19046) with similar function has been merged, this can be closed.

drelaptop:
close PR, for https://github.com/cocos2d/cocos2d-x/pull/19046 solved this issue

PR #16956, opened 2016-12-02
Ref is not copyable and related fixes

yahont:
As a replacement for #16885 which become messy after rebase with v3.
Built with gcc-6.2, xcode 8 and VC2015.
Hopefully, jenkins will build it with no errors, either.


minggo:
The compiling error happens in binding codes, i will help to fix it.

PR #16971, opened 2016-12-07
template proj.visualstudio for android

deflinhec:
This template should be able to run on android simulator with arm architecture. Running android build on visual studio emulator, which support only x86 architecture, may not work properly as expect. No error occurs when building and deploying on emulator, but GLRender is not able to render any single frame causing by glError x0501 and x0502. I'm guessing EGL.so and OpenGLES.so provide by android-ndk-r11c with x86 architecture specific did not support VS android emulator. Still working on it, and hoping for some advice.


minggo:
Thanks for your contribution, but we plan to use Android Studio, and not plan to support Android on Windows. So close the PR.

PR #16979, opened 2016-12-09
[lua-tests] Fixes PhysicsDemoBug5482 crashes.

dumganhar:
Fixed http://punchbox.info:3000/issues/26011


dumganhar:
Sync test case from cpp-tests.
cpp-tests uses addComponent to add Physics component rather than setPhysicBody.

PR #16986, opened 2016-12-10
Fix CCParticleSystemQuad::setDisplayFrame()

danut007ro:
CCParticleSystemQuad::setDisplayFrame() didn't used the rect for SpriteFrame


cocos/2d/CCParticleSystemQuad.cpp
```diff
@@ -239,11 +239,7 @@ void ParticleSystemQuad::setDisplayFrame(SpriteFrame *spriteFrame)
     CCASSERT(spriteFrame->getOffsetInPixels().isZero(), 
              "QuadParticle only supports SpriteFrames with no offsets");
 
-    // update texture before updating texture rect
-    if ( !_texture || spriteFrame->getTexture()->getName() != _texture->getName())
```
minggo, line 243: Why remove the checking?
And could you please provide a test case to reproduce the issue?

danut007ro: The same checking was implemented in setTextureWithRect(). Is ok?

minggo: Yep, it is ok.
Then could you tell me what's the issue? You are appreciated if you can provide a test case to reproduce it.

danut007ro: Without the fix, here's how particles look like (the entire texture atlas is used)
wrong

And with the fix, they look like this (ignore the sprite at bottom left)
ok

By test cases you mean tests in cpp-tests? It's my first commit to cocos2d-x so I don't know my way around... Thanks!

minggo: Yep, you can add a test case in `tests/cpp-tests/ParticleTest/ParticleTest.cpp`.

danut007ro: @minggo Just added the tests.

Here are some screenshots also. In the wrong screenshot you will see two textures used (entire atlas).

Wrong
wrong

Ok
ok

PR #16987, opened 2016-12-11
Do not wrap label on non-wrappable spaces (fixes #16939)

### LINKED ISSUE #16939: GetFirstWordLen method breaks line on "no-break" space symbol
- cocos2d-x version: 3.13
environmental and platform independent problem

Steps to Reproduce:

1.  Try to enable word wrapping on label and use symbol of "no-break" space 0xA0 
https://unicodelookup.com/# /1
For example line: "This line contains only nonbreak spaces and no spaces at all" shouldn't be broken by word wrapping. Just like in this text field: -------------------------------------------> "This line contains only nonbreak spaces and no spaces at all".
Problem is in function isUnicodeSpace which is used for word wrapping (but probably shouldn't).

fcapano:
When a label is drawn and wrapping is enabled, text should not wrap on non-wrappable spaces.
Please note that no-break spaces and non-wrappable spaces are two different sets of unicode characters. The code prevents labels' text to wrap on non-wrappable spaces.

### DIFF
--- a/cocos/base/ccUTF8.cpp
+++ b/cocos/base/ccUTF8.cpp
@@ -96,9 +96,32 @@ static void trimUTF16VectorFromIndex(std::vector<char16_t>& str, int index)
  * */
 bool isUnicodeSpace(char16_t ch)
 {
-    return  (ch >= 0x0009 && ch <= 0x000D) || ch == 0x0020 || ch == 0x0085 || ch == 0x00A0 || ch == 0x1680
-    || (ch >= 0x2000 && ch <= 0x200A) || ch == 0x2028 || ch == 0x2029 || ch == 0x202F
-    ||  ch == 0x205F || ch == 0x3000;
+    return (ch >= 0x0009 && ch <= 0x000D)   // Tabulation, Line Feed, Line Tabulation
+                                            // Form Feed, Carriage Return
+         || ch == 0x0020                    // Space
+         || ch == 0x0085                    // Next Line
+         || ch == 0x00A0                    // No Break Space
+         || ch == 0x1680                    // Ogham Space Mark
+         || (ch >= 0x2000 && ch <= 0x200A)  // EN Quad, EM Quad, EN Space, EM Space,
+                                            // Three-per-em Space, Four-per-em Space,
+                                            // Six-per-em Space, Figure Space,
+                                            // Punctuation Space, Thin Space,
+                                            // Hair Space
+         || ch == 0x2028                    // Line Separator
+         || ch == 0x2029                    // Paragraph Separator
+         || ch == 0x202F                    // Narrow no break Space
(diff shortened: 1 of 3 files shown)


cocos/base/ccUTF8.cpp
```diff
@@ -96,9 +96,32 @@ static void trimUTF16VectorFromIndex(std::vector<char16_t>& str, int index)
  * */
 bool isUnicodeSpace(char16_t ch)
 {
-    return  (ch >= 0x0009 && ch <= 0x000D) || ch == 0x0020 || ch == 0x0085 || ch == 0x00A0 || ch == 0x1680
-    || (ch >= 0x2000 && ch <= 0x200A) || ch == 0x2028 || ch == 0x2029 || ch == 0x202F
-    ||  ch == 0x205F || ch == 0x3000;
+    return (ch >= 0x0009 && ch <= 0x000D)   // Tabulation, Line Feed, Line Tabulation
+                                            // Form Feed, Carriage Return
```
minggo, line 100: What's the comment for?

cocos/base/ccUTF8.cpp
```diff
@@ -96,9 +96,32 @@ static void trimUTF16VectorFromIndex(std::vector<char16_t>& str, int index)
...
+         || ch == 0x0020                    // Space
+         || ch == 0x0085                    // Next Line
+         || ch == 0x00A0                    // No Break Space
+         || ch == 0x1680                    // Ogham Space Mark
+         || (ch >= 0x2000 && ch <= 0x200A)  // EN Quad, EM Quad, EN Space, EM Space,
+                                            // Three-per-em Space, Four-per-em Space,
+                                            // Six-per-em Space, Figure Space,
+                                            // Punctuation Space, Thin Space,
+                                            // Hair Space
```
minggo, line 109: What are the comments for?

minggo:

fcapano:
@minggo I thought this way it would be clearer what characters we're filtering

cocos/base/ccUTF8.cpp
fcapano: @minggo some lines filter more than one character, hence the extra comments

minggo: Got it, thanks.

minggo:
@fcapano could you please provide a test case to describe the issue? I don't know what's the issue. Thanks.

fcapano:
a simple example of this issue is, in the helloworld scene, to replace the space in the label "Hello World" with a non-breaking space "Hello\xA0World", and set the width to be half of what it currently is.
The label will wrap the text even though it shouldn'n

minggo:
> World" with a non-breaking space "Hello\xA0World"

The label is not shown even i didn't change the width.

minggo:
@fcapano any update?

PR #16989, opened 2016-12-12
To use some HTTP client library those use `send(null)` like `superage…

kosmosR2:
To use some HTTP client library those use `XMLHttpRequest.send(null)` like `superagent`,`XMLHTTPRequest` should send a empty request instead of abort.


dumganhar (approved):

minggo:
Sorry, what's the issue?

kosmosR2:
In JSB,when I use some HTTP libraries from npm like `superagent` and send a GET request without specify content,the default action of those libraries is calling `XMLHttpRequest.send(null)` which leads to abort the request in original code.

leeyeh:
Any progress here?

PR #16991, opened 2016-12-12
[win32] Fixes crash of getSubStringOfUTF8String.

dumganhar:
fixed http://punchbox.info:3000/issues/26021

### DIFF
--- a/cocos/ui/UIHelper.cpp
+++ b/cocos/ui/UIHelper.cpp
@@ -122,7 +122,7 @@ std::string Helper::getSubStringOfUTF8String(const std::string& str, std::string
         return "";
     }
     if (utf32.size() < start) {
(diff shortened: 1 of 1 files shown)


cocos/ui/UIHelper.cpp
```diff
@@ -122,7 +122,7 @@ std::string Helper::getSubStringOfUTF8String(const std::string& str, std::string
         return "";
     }
     if (utf32.size() < start) {
-        CCLOGERROR("'start' is out of range: %zu, %s", static_cast<size_t>(start), str.c_str());
+        CCLOGERROR("'start' is out of range: %ld, %s", static_cast<long>(start), str.c_str());
```
dumganhar, line 125: Windows doesn't support "%zu" to format string.

PR #16993, opened 2016-12-12
Fix calculation of first word length (fixes #16889)

### LINKED ISSUE #16889: Last letter in first word not transferred to new line. Label::getFirstWordLen
- cocos2d-x version: 3.13.1
- devices test on: Win32
Steps to Reproduce:
1.  Create multiline cocos2d::Label with text with one word. Set the content size that was not seen the last half of the letter. The last letter should be transferred to the next line. But its not transferred.

Problem is in function Label::getFirstWordLen. See for code:
for (int index = startIndex + 1; index < textLen; ++index)
Width of first letter not added to word width.

fcapano:


dumganhar:
@zilongshanren , please review this pull request.

minggo:
@fcapano could you please add a test case to show the issue? Thanks.

PR #17003, opened 2016-12-13
Synchronize creator engine bug fixes and upgrade web engine 

pandamicro:


cocos/scripting/js-bindings/manual/cocos2d_specifics.cpp
```diff
@@ -1328,7 +1328,7 @@ bool js_CCNode_unschedule(JSContext *cx, uint32_t argc, jsval *vp)
 
         auto targetArray = JSScheduleWrapper::getTargetForSchedule(args.get(0));
         if (targetArray) {
-            CCLOGINFO("unschedule target number: %d", targetArray->count());
+            CCLOGINFO("unschedule target number: %zd", targetArray->count());
```
dumganhar: "%zd" isn't supported on win32, please use "%ld" and `static_cast<long>(targetArray->count())`.

minggo:
Thanks, could you please also update CHANGELOG too?

PR #17011, opened 2016-12-14
[mac, ios] 'ExtAudioFileRead' may return 0 frame while it doesn’t reach the end of very short audio file.

dumganhar:
This is a bug of Apple API `ExtAudioFileRead`. An assert in the original code will be triggered while play very very small file. So we should not remove the assertion check.

Bug fix code:

```c++
-            CCASSERT(_framesRead == originalTotalFrames, "The read frame count isn't equal to orignal frame count!");
+            if (_framesRead < originalTotalFrames)
+            {
+                memset(_pcmData + _framesRead * bytesPerFrame, 0x00, (totalFrames - _framesRead) * bytesPerFrame);
+            }

```

Another refactor is that adding `readFixedFrames` method for decoder to make sure one read operation could read enough data since ExtAudioFileRead may return fewer frames than expected.
Please refer to API doc:
```c++
/*!
	@function   ExtAudioFileRead
	
	@abstract   Perform a synchronous sequential read.
	
	@param		inExtAudioFile
					The extended audio file object.
	@param		ioNumberFrames
					On entry, ioNumberFrames is the number of frames to be read from the file.
					On exit, it is the number of frames actually read. A number of factors may
					cause a fewer number of frames to be read, including the supplied buffers
					not being large enough, and internal optimizations. If 0 frames are
					returned, however, this indicates that end-of-file was reached.
	@param		ioData
					Buffer(s) into which the audio data is read.
	@result		An OSStatus error code.

	@discussion
				If the file has a client data format, then the audio data from the file is
				translated from the file data format to the client format, via the
				ExtAudioFile's internal AudioConverter.
				
				(Note that the use of sequential reads/writes means that an ExtAudioFile must
				not be read on multiple threads; clients wishing to do this should use the
				lower-level AudioFile API set).
*/
extern OSStatus
ExtAudioFileRead(			ExtAudioFileRef			inExtAudioFile,
							UInt32 *				ioNumberFrames,
							AudioBufferList *		ioData)					
																			__OSX_AVAILABLE_STARTING(__MAC_10_4,__IPHONE_2_1);
```


minggo:
So what's the error behavior? Can not play the effect or wrong sound?

cocos/audio/apple/AudioCache.mm
```diff
@@ -255,7 +252,10 @@ static ALvoid  alBufferDataStaticProc(const ALint bid, ALenum format, ALvoid* da
                 remainingFrames -= framesRead;
             }
 
-            CCASSERT(_framesRead == originalTotalFrames, "The read frame count isn't equal to orignal frame count!");
+            if (_framesRead < originalTotalFrames)
```
dumganhar, line 255: @minggo , it may trigger this assertion while playing very small audio file.
As the title said, 'ExtAudioFileRead' may return 0 frame while it doesn’t reach the end of very short audio file.
This is a bug of Apple SDK. :(

PR #17041, opened 2016-12-19
Sync issue fixes from creator

pandamicro:


tests/js-tests/src/SpineTest/SpineTest.js
```diff
@@ -216,7 +216,7 @@ var SpineTestLayerFFD = SpineTestLayer.extend({
     ctor: function(){
         this._super(cc.color(0,0,0,255), cc.color(98,99,117,255));
 
-        var skeletonNode = sp.SkeletonAnimation.createWithJsonFile("spine/goblins.json", "spine/goblins.atlas", 1.5);
```
minggo, line 219: Please don't modify it. It will cause issue in JSB.

PR #17061, opened 2016-12-24
fix type conversion warnings 

ilqvya:
I hate warnings in my warnings list!

### DIFF
--- a/cocos/2d/CCLabel.h
+++ b/cocos/2d/CCLabel.h
@@ -410,7 +410,7 @@ public:
     /**
     * Return the outline effect size value.
     */
-    float getOutlineSize() const { _outlineSize; }
+    float getOutlineSize() const { return _outlineSize; }
 
     /**
     * Return current effect type.
--- a/cocos/2d/CCLabel.h
+++ b/cocos/2d/CCLabel.h
@@ -410,7 +410,7 @@ public:
     /**
     * Return the outline effect size value.
     */
-    int getOutlineSize() const { return _outlineSize; }
+    int getOutlineSize() const { return static_cast<int>(_outlineSize); }
 
     /**
     * Return current effect type.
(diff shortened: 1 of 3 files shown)


cocos/2d/CCLabel.h
```diff
@@ -410,7 +410,7 @@ class CC_DLL Label : public Node, public LabelProtocol, public BlendProtocol
     /**
     * Return the outline effect size value.
     */
-    int getOutlineSize() const { return _outlineSize; }
+    int getOutlineSize() const { return static_cast<int>(_outlineSize); }
```
minggo: I think we should return float instead of cast internally.
@zilongshanren

cocos/3d/CCTerrain.h
```diff
@@ -424,7 +424,7 @@ class CC_DLL Terrain : public Node
     /**
      * get the terrain's size
      */
-    Size getTerrainSize() const { return Size(_imageWidth, _imageHeight); }
+    Size getTerrainSize() const { return Size(static_cast<float>(_imageWidth), static_cast<float>(_imageHeight)); }
```
minggo, line 427: I think it is more reasonable to change the type of `_imageWidth ` and `_imageHeight ` to float. 
@ricardoquesada what do you think?

cocos/2d/CCLabel.h
ricardoquesada: agreed with @minggo

cocos/3d/CCTerrain.h
ricardoquesada: @minggo the code, internally, treat it as int. in this particular case, it seems better to cast it to float in `getTerrainSize`

minggo: Agree with @ricardoquesada

cocos/2d/CCLabel.h
minggo: @effolkronium what's your opinion? Could you please change to int to float?

ilqvya: @minggo i've made the change. Float seems better.

PR #17067, opened 2016-12-26
Fix QA reported issues in v3.14

pandamicro:


pandamicro:
@minggo why merged commits is not recognized ? Are we using some new merging policy ?

minggo:
Yep, we use squash and merge, so you can not use the same branch for more PR.

PR #17072, opened 2016-12-27
vec4 operator < seems error

niu2x:
vec4 operator < seems error,


minggo:
Yep, i think you are right. But i don't know what's the meaning of comparing two vectors. The math library comes from gameplay, i will take a look at it.

minggo:
Yep, gameplay fixed the bug: https://github.com/gameplay3d/GamePlay/blob/master/gameplay/src/Vector4.inl

PR #17079, opened 2016-12-28
Fixes crash that playing long audio and then uncacheAll

dumganhar:
This issue was reported at http://punchbox.info:3000/issues/26038

Also refactor the logic of New Audio Engine for win32.
Abstract decoding class.
Syncs with the OpenAL code from `cocos/audio/apple`

### DIFF
--- a/cocos/audio/AudioEngine.cpp
+++ b/cocos/audio/AudioEngine.cpp
@@ -70,17 +70,12 @@ AudioEngine::AudioEngineThreadPool* AudioEngine::s_threadPool = nullptr;
 class AudioEngine::AudioEngineThreadPool
 {
 public:
-    AudioEngineThreadPool(bool detach, int threads = 4)
-        : _detach(detach)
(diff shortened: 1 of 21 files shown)


cocos/audio/AudioEngine.cpp
```diff
@@ -169,17 +160,10 @@ bool AudioEngine::lazyInit()
         }
     }
 
-#if CC_TARGET_PLATFORM == CC_PLATFORM_WIN32
     if (_audioEngineImpl && s_threadPool == nullptr)
     {
-        s_threadPool = new (std::nothrow) AudioEngineThreadPool(true);
+        s_threadPool = new (std::nothrow) AudioEngineThreadPool();
     }
-#elif CC_TARGET_PLATFORM != CC_PLATFORM_ANDROID
```
dumganhar, line 177: Android should not create a threadpool here. I will fix it.

PR #17098, opened 2017-01-03
make ListView::setGlobalZOrder() work

minggo:


minggo:
@zilongshanren please help to review it, thanks.

cocos/ui/UIButton.cpp
```diff
@@ -871,6 +871,19 @@ std::string Button::getDescription() const
...
+{
+    Widget::setGlobalZOrder(globalZOrder);
+    if (_buttonNormalRenderer)
+        _buttonNormalRenderer->setGlobalZOrder(globalZOrder);
+    if (_buttonClickedRenderer)
+        _buttonClickedRenderer->setGlobalZOrder(globalZOrder);
+    if (_buttonDisabledRenderer)
+        _buttonDisabledRenderer->setGlobalZOrder(globalZOrder);
+    if (_titleRenderer)
+        _titleRenderer->setGlobalZOrder(globalZOrder);
```
minggo: Oh it is not needed since `ProtectedNode::setGlobalZOder()` will change the protected children's global z order.

zilongshanren:
@minggo 
It looks good to me, but I suggest to add another three tests for `ui::Layout`, `ui::ScrollView` and `ui::PageView`.

tiagomartines11:
I believe ui::Widget needs similar treatment to propagate globalZOrder properly.

Example, assuming a tree like this:
ScrollView (z:0) -> Layout (z:0) -> TextBMFont (z:0) -> TextBMFont (z:0)

Before this fix:
`scrollView->setGlobalZOrder(200)` would make:
ScrollView (z:200) -> Layout (z:0) -> TextBMFont (z:0) -> TextBMFont (z:0)

After this fix:
`scrollView->setGlobalZOrder(200)` makes:
ScrollView (z:200) -> Layout (z:200) -> TextBMFont (z:200) -> TextBMFont (z:0)

This is highly confusing and prevents properly ordering UI elements (both for clipping and for ordering commands).

Can you please review this?

tankorsmash:
Yeah, this change means that if you add a child to a Layout, and want to have a custom global z order for it, you must wait until after you add the child to the Layout.

Not sure what the best approach is though, but it seems like there's a few different concerns that are conflicting with each other.

PR #17115, opened 2017-01-05
Bindings fix and assets manger ex fix

pandamicro:


minggo:
please fix compiling error

pandamicro:
More changes needed, please do not merge it yet

minggo:
@pandamicro what's the status of this PR?

pandamicro:
Here is some brief informations about the upgrade:

1. **IMPORTANT**: Separate storage folder for updating and local cached version, so that the failing in 1. update won't damage the current local version. The updating version will be overwritten to the local cache folder only when the update is fully successful.
1. **IMPORTANT**: Bound Downloader to JS and support file download task, data task is not supported
1. **IMPORTANT**: Decompress process is moved before file success notification, downloading a zip file is considered successful only after the decompress process is successfully done.
1. Support size property in manifest file for collecting total byte size of the update before it's started.
1. Fix download resuming on Android and some edge condition bugs.
1. Support customized version compare API between manifests: setVersionCompareHandle
1. Support verification callback after file downloaded: setVerifyCallback
1. Support setting maximum concurrent tasks count: setMaxConcurrentTask
1. Ensure saving download progression to temporary manifest
1. Delay folder creation right before file download request to ensure the path is correct
1. Ensure reading the correct manifest in app package

In Cocos Creator's hot update sample, I also provided a manifest generation script, feel free to try out:

https://github.com/cocos-creator/tutorial-hot-update/blob/master/version_generator.js

PR #17133, opened 2017-01-09
WinRT: updated location of openssl dlls

stammen:
This PR fixes the broken Windows 8.1 and Windows 10 UWP builds. The openssl libs were moved from external\curl\prebuilt\win10 to external\oepnsll\prebuilt\win10.


minggo:
thanks @stammen

PR #17137, opened 2017-01-10
add cocos/editor-support into search path

minggo:


ljwlaji:
Already Added $(SRCROOT)/../cocos2d/cocos/editor-support" to target And It Can Auto Link To The Head File  Buy When I Try To Complete, The Not Found Error appear Again

PR #17140, opened 2017-01-10
Adding async version of common CCFileUtil functions.

DavidDeSimone:
Following through on #16846. 

This adds "async" versions of the major FileUtil functions, that will be performed off thread, and callback to the main thread when complete with their result.

Important Notes:

I used the AsyncTaskPool under the task type IO.
In order to keep the details of the async mechanism self contained, I pulled the functionality to a function called performOperationOffthread. This way if we ever want to change the mechanism from AsyncTaskPool to another threading mechanism, it should be easy.
While FileUtils is not thread safe, I was able to work around this by calling "fullPathForFilename" on the main thread before calling the delegate functions.
I attempted several template meta-programming solutions to this problem before settling on the submitted approach. All of the template solutions were too complex, and non readable in my opinion.
You can see I am using a number of value types instead of const-references. This is intentional as we are moving this data between threads, and I want to leverage move semantics.

@minggo

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -569,6 +569,13 @@ bool FileUtils::writeStringToFile(const std::string& dataStr, const std::string&
     return rv;
 }
 
+void FileUtils::writeStringToFile(std::string dataStr, const std::string& fullPath, std::function<void(bool)> callback)
+{
+    performOperationOffthread([fullPath](const std::string& dataStr) -> bool {
+            return FileUtils::getInstance()->writeStringToFile(dataStr, fullPath);
+    }, std::move(callback),std::move(dataStr));
+}
+
 bool FileUtils::writeDataToFile(const Data& data, const std::string& fullPath)
 {
     size_t size = 0;
@@ -594,6 +601,13 @@ bool FileUtils::writeDataToFile(const Data& data, const std::string& fullPath)
     return false;
 }
 
+void FileUtils::writeDataToFile(Data data, const std::string& fullPath, std::function<void(bool)> callback)
+{
+    performOperationOffthread([fullPath](const Data& data) -> bool {
+            return FileUtils::getInstance()->writeDataToFile(data, fullPath);
+    }, std::move(callback), std::move(data));
+}
+
 bool FileUtils::init()
 {
     _searchPathArray.push_back(_defaultResRootPath);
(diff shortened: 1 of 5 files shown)


cocos/platform/CCFileUtils.cpp
```diff
@@ -569,6 +569,13 @@ bool FileUtils::writeStringToFile(const std::string& dataStr, const std::string&
     return rv;
 }
 
+void FileUtils::writeStringToFile(std::string dataStr, const std::string& fullPath, std::function<void(bool)> callback)
+{
+    performOperationOffthread([fullPath](const std::string& dataStr) -> bool {
+            return FileUtils::getInstance()->writeStringToFile(dataStr, fullPath);
+    }, std::move(callback),std::move(dataStr));
```
minggo, line 576: Why need std::move for std::string and std::function?

DavidDeSimone: It is leveraging move sematics to avoid data copying. If I did not use std::move here, it would invoke the copy constructor of callback and dataStr, causing a potentially costly copy operation. Instead it will invoke the move constructor, which is a less expensive move operation (usually just a pointer assignment). 

Here is an example to demonstrate. 

```
class NoCopy {
public:
    NoCopy() = default;
    NoCopy(const NoCopy&) = delete;
    NoCopy& operator=(const NoCopy&) = delete;
    
    NoCopy(NoCopy&&) { CCLOG("IM MOVING!"); }
    NoCopy& operator=(NoCopy&&) { CCLOG("OPERATOR="); return *this; }
};

void moveSemantics(NoCopy value)
{
    NoCopy value2 = std::move(value); //value2 will trigger move constructor of value
    NoCopy copy = value; // compile error, attempts to trigger copy constructor. would
    // copy if copy constructor was available. 
}

int main(int argc, char** argv)
{
    moveSemantics(NoCopy()); // Works fine, NoCopy object is not copied
    NoCopy t;
    moveSemantics(t); // Compile error, causes a copy, would call copy constructor if provided
}
```

In general, if we are assigning values like this, we should be taking them by value instead of by const reference (There are some exceptions to this). Assigning from a cost l-value reference will trigger the copy constructor, while passing by value will allow us to avoid the copy if the callsite is an r-value. Then we can call std::move to trigger the move and avoid copying within the function.

Properly leveraging move semantics gives us the ability to avoid copying, while having the safety given by RAII.

minggo: Thans for explanation

PR #17158, opened 2017-01-11
Correct spelling of function name `getAncestorWidget`

mogemimi:
This corrects spelling mistake in `getAncestorWidget` function. It also keeps the old one `getAncensterWidget` as a deprecated function for backwards compatibility.


ricardoquesada:
thanks.

PR #17159, opened 2017-01-11
fixed two scheduler bugs (fixes #17149)

### LINKED ISSUE #17149: Possible Memory Bad Access Error on cocos2d::Scheduler
- cocos2d-x version: 3.12
- devices test on: iMac
- developing environments
   - NDK version: none
   - Xcode version: 8.0
   - VS version: none
   - browser type and version: none

1. start scheduler's update loop
2. create object `one` which has `TypeA`. 
3. address of object `one` is 0x12341234
3. run Scheduler::scheduleUpdate for `one`
4. unschedule `one` immediately. Scheduler will mark it as deleted. because _updateHashLocked is true.
```
void Scheduler::unscheduleUpdate(void *target)
{
    if (target == nullptr)
    {
        return;
    }

    tHashUpdateEntry *element = nullptr;
    HASH_FIND_PTR(_hashForUpdates, &target, element);
    if (element)
    {
        if (_updateHashLocked)
        {
            element->entry->markedForDeletion = true;
        }
        else
        {
            this->removeUpdateFromHash(element->entry);
        }
    }
}
```
5. release `one`
6. create object `two` which has `TypeB`.
7. address of object `two` is 0x12341234. same with previous one.
8. run Scheduler::scheduleUpdate for `two`
9. scheduler already has hash element with 0x12341234. so it will change markedForDeletion as true. but callback isn't changed.
```
void Scheduler::schedulePerFrame(const ccSchedulerFunc& callback, void *target, int priority, bool paused)
{
    tHashUpdateEntry *hashElement = nullptr;
    if (hashElement)
    {
        // check if priority has changed
        if ((*hashElement->list)->priority != priority)
        {
            if (_updateHashLocked)
            {
                CCLOG("warning: you CANNOT change update priority in scheduled function");
                hashElement->entry->markedForDeletion = false;
                hashElement->entry->paused = paused;
                return;
            }
            else
            {
            	// will be added again outside if (hashElement).
                unscheduleUpdate(target);
            }
        }
        else
        {
            hashElement->entry->markedForDeletion = false;
            hashElement->entry->paused = paused;
            return;
        }
    }
...
```
10. end of update loop

now target object `two` has callback with wrong template type, and it will make undefined actions.
is it bug? if not, What do I have to do?

cinakyn:
1. Fixed that Scheduler checks target's priority change incorrectly.
2. Fixed Scheduler bug occured when objects that have same address are inserted and removed in same tick. (refer the #17149)

thanks.

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -465,7 +465,7 @@ void Scheduler::schedulePerFrame(const ccSchedulerFunc& callback, void *target, 
     if (hashElement)
     {
         // check if priority has changed
-        if ((*hashElement->list)->priority != priority)
+        if (hashElement->entry->priority != priority)
         {
             if (_updateHashLocked)
             {
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -472,6 +472,7 @@ void Scheduler::schedulePerFrame(const ccSchedulerFunc& callback, void *target, 
                 CCLOG("warning: you CANNOT change update priority in scheduled function");
                 hashElement->entry->markedForDeletion = false;
                 hashElement->entry->paused = paused;
+                hashElement->entry->callback = callback;
                 return;
             }
             else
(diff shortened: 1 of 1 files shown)


cinakyn:
There's still one problem.
if object that have same addresses are have different priority, and each object is swapped in same tick, refreshed object's prioirty will not be preserved.

PR #17163, opened 2017-01-11
Particle set display frame

danut007ro:
I'm opening this because https://github.com/cocos2d/cocos2d-x/pull/16986 hasn't been merged and it's still an issue


minggo:
sorry, it may be closed by mistook

PR #17164, opened 2017-01-11
Fix incorrect function name for calculateBillboardTransform

mogemimi:
There is a spelling mistake in function name, `calculateBillbaordTransform` should be `calculateBillboardTransform`. This patch fixes it, but keeps the old function for backwards compatibility. Thanks!


minggo:
thanks

PR #17168, opened 2017-01-12
Fix Scheduler Bugs.(issue #17149)

cinakyn:
1. Fix bug that `Scheduler::schedulePerFrame` compares targets priority incorrectly.
2. (refer #17149) Fix bugs occur when two objects that have same address are inserted and removed in same tick. scheduler will always remove target from `_hashForUpdates` when unschedule target.

### DIFF
--- a/cocos/base/CCScheduler.cpp
+++ b/cocos/base/CCScheduler.cpp
@@ -464,26 +464,27 @@ void Scheduler::schedulePerFrame(const ccSchedulerFunc& callback, void *target, 
     HASH_FIND_PTR(_hashForUpdates, &target, hashElement);
     if (hashElement)
     {
+        CCAssert(!hashElement->entry->markedForDeletion, "All elements in _hashForUpdate must not be marked for deletion.");
         // check if priority has changed
-        if ((*hashElement->list)->priority != priority)
+        if (hashElement->entry->priority != priority)
         {
             if (_updateHashLocked)
             {
                 CCLOG("warning: you CANNOT change update priority in scheduled function");
-                hashElement->entry->markedForDeletion = false;
                 hashElement->entry->paused = paused;
+                hashElement->entry->callback = callback;
                 return;
             }
             else
             {
-            	// will be added again outside if (hashElement).
+                // will be added again outside if (hashElement).
                 unscheduleUpdate(target);
             }
         }
         else
         {
-            hashElement->entry->markedForDeletion = false;
             hashElement->entry->paused = paused;
+            hashElement->entry->callback = callback;
             return;
         }
     }
@@ -547,10 +548,16 @@ void Scheduler::removeUpdateFromHash(struct _listEntry *entry)
     HASH_FIND_PTR(_hashForUpdates, &entry->target, element);
     if (element)
     {
-        // list entry
-        DL_DELETE(*element->list, element->entry);
-        CC_SAFE_DELETE(element->entry);
-
+        if (_updateHashLocked)
+        {
+            element->entry->markedForDeletion = true;
+        }
+        else
+        {
+            // list entry
+            DL_DELETE(*element->list, element->entry);
+            CC_SAFE_DELETE(element->entry);
+        }
         // hash entry
         HASH_DEL(_hashForUpdates, element);
         free(element);
@@ -568,14 +575,7 @@ void Scheduler::unscheduleUpdate(void *target)
     HASH_FIND_PTR(_hashForUpdates, &target, element);
     if (element)
     {
-        if (_updateHashLocked)
-        {
-            element->entry->markedForDeletion = true;
-        }
-        else
-        {
-            this->removeUpdateFromHash(element->entry);
-        }
+        this->removeUpdateFromHash(element->entry);
     }
 }
 
@@ -606,7 +606,15 @@ void Scheduler::unscheduleAllWithMinPriority(int minPriority)
         {
             if(entry->priority >= minPriority)
             {
-                unscheduleUpdate(entry->target);
+                if (entry->markedForDeletion)
+                {
+                    DL_DELETE(_updatesNegList, entry);
+                    CC_SAFE_DELETE(entry);
+                }
+                else
+                {
+                    unscheduleUpdate(entry->target);
+                }
             }
         }
     }
@@ -615,13 +623,26 @@ void Scheduler::unscheduleAllWithMinPriority(int minPriority)
     {
         DL_FOREACH_SAFE(_updates0List, entry, tmp)
         {
-            unscheduleUpdate(entry->target);
+            if (entry->markedForDeletion)
+            {
+                DL_DELETE(_updates0List, entry);
+                CC_SAFE_DELETE(entry);
+            }
+            else
+            {
+                unscheduleUpdate(entry->target);
+            }
         }
     }
 
     DL_FOREACH_SAFE(_updatesPosList, entry, tmp)
     {
-        if(entry->priority >= minPriority)
+        if (entry->markedForDeletion)
+        {
+            DL_DELETE(_updatesPosList, entry);
+            CC_SAFE_DELETE(entry);
+        }
+        else
         {
             unscheduleUpdate(entry->target);
         }
@@ -920,7 +941,8 @@ void Scheduler::update(float dt)
     {
         if (entry->markedForDeletion)
         {
-            this->removeUpdateFromHash(entry);
+            DL_DELETE(_updatesNegList, entry);
+            CC_SAFE_DELETE(entry);
         }
     }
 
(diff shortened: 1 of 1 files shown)


cocos/base/CCScheduler.cpp
```diff
@@ -547,10 +548,16 @@ void Scheduler::removeUpdateFromHash(struct _listEntry *entry)
...
+        if (_updateHashLocked)
+        {
+            element->entry->markedForDeletion = true;
+        }
+        else
+        {
+            // list entry
+            DL_DELETE(*element->list, element->entry);
+            CC_SAFE_DELETE(element->entry);
+        }
```
minggo, line 560: removeUpdateFromHash should not consider `_updateHashLocked `, it just remove from hash, or it will make things complex as you added the codes following.

cocos/base/CCScheduler.cpp
```diff
@@ -464,26 +464,27 @@ void Scheduler::schedulePerFrame(const ccSchedulerFunc& callback, void *target,
     HASH_FIND_PTR(_hashForUpdates, &target, hashElement);
     if (hashElement)
     {
+        CCAssert(!hashElement->entry->markedForDeletion, "All elements in _hashForUpdate must not be marked for deletion.");
```
minggo, line 467: Add assert can make developers notice the error, but it doesn't fix the issue. I think we should find a way to fix it.

cinakyn: Then I think deleting this line is better, because that situation will never occur.

cocos/base/CCScheduler.cpp
cinakyn: @minggo 
Then how about removing `Scheduler::removeUpdateFromHash` and insert that logic into `Scheduler::unscheduleUpdate`? and _updateHashLocked will be used in `Scheduler::unscheduleUpdate` again.

because
1. removeUpdateFromHash is appeared only in one place.
2. removeUpdateFromHash actually also remove item from some doubly linked list.

thanks. sorry for bad english skill.

minggo:
After look into detail, i found `_updateHashLocked` is set to true in `Scheduler::update` to prevent current working target is removed from list. But we use `DL_FOREACH_SAFE` to loop the list, and it is safe to delete element while iterating. So i think we can just remove `_updateHashLocked` and remove it directly.

cinakyn:
I already tried that method. but it makes our game (pretty big) be crahed when scheduler calls target's callback. and could not find the reason.  below code is diif that I used.
```diff
diff --git a/cocos2d/cocos/base/CCScheduler.cpp b/cocos2d/cocos/base/CCScheduler.cpp
index a4a0d00..004058b 100644
--- a/cocos2d/cocos/base/CCScheduler.cpp
+++ b/cocos2d/cocos/base/CCScheduler.cpp
@@ -44,7 +44,6 @@ typedef struct _listEntry
     void                *target;
     int                 priority;
     bool                paused;
-    bool                markedForDeletion; // selector will no longer be called and entry will be removed at end of the next tick
 } tListEntry;
 
 typedef struct _hashUpdateEntry
@@ -252,7 +251,6 @@ Scheduler::Scheduler(void)
 , _hashForTimers(nullptr)
 , _currentTarget(nullptr)
 , _currentTargetSalvaged(false)
-, _updateHashLocked(false)
 #if CC_ENABLE_SCRIPT_BINDING
 , _scriptHandlerEntries(20)
 #endif
@@ -390,7 +388,6 @@ void Scheduler::priorityIn(tListEntry **list, const ccSchedulerFunc& callback, v
     listElement->priority = priority;
     listElement->paused = paused;
     listElement->next = listElement->prev = nullptr;
-    listElement->markedForDeletion = false;
 
     // empty list ?
     if (! *list)
@@ -446,7 +443,6 @@ void Scheduler::appendIn(_listEntry **list, const ccSchedulerFunc& callback, voi
     listElement->target = target;
     listElement->paused = paused;
     listElement->priority = 0;
-    listElement->markedForDeletion = false;
 
     DL_APPEND(*list, listElement);
 
@@ -464,28 +460,8 @@ void Scheduler::schedulePerFrame(const ccSchedulerFunc& callback, void *target,
     HASH_FIND_PTR(_hashForUpdates, &target, hashElement);
     if (hashElement)
     {
-        // check if priority has changed
-        if ((*hashElement->list)->priority != priority)
-        {
-            if (_updateHashLocked)
-            {
-                CCLOG("warning: you CANNOT change update priority in scheduled function");
-                hashElement->entry->markedForDeletion = false;
-                hashElement->entry->paused = paused;
-                return;
-            }
-            else
-            {
-            	// will be added again outside if (hashElement).
-                unscheduleUpdate(target);
-            }
-        }
-        else
-        {
-            hashElement->entry->markedForDeletion = false;
-            hashElement->entry->paused = paused;
-            return;
-        }
+        // will be added again outside if (hashElement).
+        unscheduleUpdate(target);
     }
 
     // most of the updates are going to be 0, that's way there
@@ -550,7 +526,6 @@ void Scheduler::removeUpdateFromHash(struct _listEntry *entry)
         // list entry
         DL_DELETE(*element->list, element->entry);
         CC_SAFE_DELETE(element->entry);
-
         // hash entry
         HASH_DEL(_hashForUpdates, element);
         free(element);
@@ -568,14 +543,7 @@ void Scheduler::unscheduleUpdate(void *target)
     HASH_FIND_PTR(_hashForUpdates, &target, element);
     if (element)
     {
-        if (_updateHashLocked)
-        {
-            element->entry->markedForDeletion = true;
-        }
-        else
-        {
-            this->removeUpdateFromHash(element->entry);
-        }
+        this->removeUpdateFromHash(element->entry);
     }
 }
 
@@ -621,10 +589,7 @@ void Scheduler::unscheduleAllWithMinPriority(int minPriority)
 
     DL_FOREACH_SAFE(_updatesPosList, entry, tmp)
     {
-        if(entry->priority >= minPriority)
-        {
-            unscheduleUpdate(entry->target);
-        }
+        unscheduleUpdate(entry->target);
     }
 #if CC_ENABLE_SCRIPT_BINDING
     _scriptHandlerEntries.clear();
@@ -778,7 +743,7 @@ std::set<void*> Scheduler::pauseAllTargetsWithMinPriority(int minPriority)
     tListEntry *entry, *tmp;
     if(minPriority < 0)
     {
-        DL_FOREACH_SAFE( _updatesNegList, entry, tmp ) 
+        DL_FOREACH_SAFE( _updatesNegList, entry, tmp )
         {
             if(entry->priority >= minPriority)
             {
@@ -828,8 +793,6 @@ void Scheduler::performFunctionInCocosThread(const std::function<void ()> &funct
 // main loop
 void Scheduler::update(float dt)
 {
-    _updateHashLocked = true;
-
     if (_timeScale != 1.0f)
     {
         dt *= _timeScale;
@@ -845,7 +808,7 @@ void Scheduler::update(float dt)
     // updates with priority < 0
     DL_FOREACH_SAFE(_updatesNegList, entry, tmp)
     {
-        if ((! entry->paused) && (! entry->markedForDeletion))
+        if ((! entry->paused))
         {
             entry->callback(dt);
         }
@@ -854,7 +817,7 @@ void Scheduler::update(float dt)
     // updates with priority == 0
     DL_FOREACH_SAFE(_updates0List, entry, tmp)
     {
-        if ((! entry->paused) && (! entry->markedForDeletion))
+        if ((! entry->paused))
         {
             entry->callback(dt);
         }
@@ -863,7 +826,7 @@ void Scheduler::update(float dt)
     // updates with priority > 0
     DL_FOREACH_SAFE(_updatesPosList, entry, tmp)
     {
-        if ((! entry->paused) && (! entry->markedForDeletion))
+        if ((! entry->paused))
         {
             entry->callback(dt);
         }
@@ -907,36 +870,6 @@ void Scheduler::update(float dt)
             removeHashElement(_currentTarget);
         }
     }
-
-    // delete all updates that are marked for deletion
-    // updates with priority < 0
-    DL_FOREACH_SAFE(_updatesNegList, entry, tmp)
-    {
-        if (entry->markedForDeletion)
-        {
-            this->removeUpdateFromHash(entry);
-        }
-    }
-
-    // updates with priority == 0
-    DL_FOREACH_SAFE(_updates0List, entry, tmp)
-    {
-        if (entry->markedForDeletion)
-        {
-            this->removeUpdateFromHash(entry);
-        }
-    }
-
-    // updates with priority > 0
-    DL_FOREACH_SAFE(_updatesPosList, entry, tmp)
-    {
-        if (entry->markedForDeletion)
-        {
-            this->removeUpdateFromHash(entry);
-        }
-    }
-
-    _updateHashLocked = false;
     _currentTarget = nullptr;
 
 #if CC_ENABLE_SCRIPT_BINDING
```

minggo:
yep, it seems we can not remove `_updateHashLocked `, but can change its action scope. Currently, `_updateHashLocked` is used like
```c++
   if (_updateHashLocked)
       can not remove any element from list
       can not remove any element from hash
       can not delete any list element
       can not delete any hash element
```
we should change it to
```c++
if (_updateHashLocked)
       can not delete current list element
       can not delete current hash element
```

I will modify and send a PR.

minggo:
i sent the PR: https://github.com/cocos2d/cocos2d-x/pull/17198

PR #17175, opened 2017-01-12
fix rendering font with not null line gap

newnon:
cocos2d-x does not respect line gap in ttf font


minggo:
how to reproduce it?

newnon:
You need any font with not null line gap you can see it on second picture [http://designwithfontforge.com/en-US/Line_Spacing.html](http://designwithfontforge.com/en-US/Line_Spacing.html), you need for example 2 lines text and then draw once with system render once with own and you will see that line spacing is different

minggo:
could you please add a test case for it

zilongshanren:
@newnon 
What's the difference between `lineSpacing` and `linGap`?  I think there is already a API for changing the line spacing between lines.

newnon:
lineSpacing is our own font param and linGap is internal TTF font param without it font will be misalignment (half of them added to top space half to bottom)

zilongshanren:
@newnon 
It would be good if you could provide a test case to reproduce the issue you talked above.
And the test case could also prevent future false regression.

newnon:
some code refining & test case

newnon:
@minggo have you looked at this?

minggo:
@newnon i haven't tested it. I will test i ASAP.

minggo:
Yep, it looks good to me.

minggo:
@newnon why not use existing ttf file to avoid making the repo bigger.

newnon:
all fonts that included have line gap = 0 or bad
AmericanTypeWriter for example have different values for different systems HH Line Gap & Type Line Gap must be same because (you can see example for bad font on this screen http://www.high-logic.com/fontcreator/manual10/fontcreator_039_font_settings_metrics.png) right params (ascender descender and line gap) used only by mac & iOS and left for params for all other system include lib freetype. 
You can include this code without test i think it is not so bad

minggo:
Ok, thanks. The test case is needed.

newnon:
i can take another font or remove most letters from this and it will be very small

minggo:
@newnon thanks. And please remove the commit log too or it will in the history.

newnon:
small font for test

PR #17176, opened 2017-01-12
outline to float

newnon:
Float outline just compare enableOutline for Label with 0,5 and 1 you will see difference


minggo:
looks good to me, @zilongshanren please take a look

zilongshanren:
@minggo  It looks good to me.

PR #17177, opened 2017-01-12
fix child removing from ListView

newnon:
small fix not calling NeedLayout if on remove child


zilongshanren:
@minggo  This PR looks good to me.

PR #17178, opened 2017-01-12
getLetter function for Text

newnon:
getLette fot ui:Text

### DIFF
--- a/cocos/ui/UIText.h
+++ b/cocos/ui/UIText.h
@@ -37,6 +37,7 @@ THE SOFTWARE.
 NS_CC_BEGIN
 
 class Label;
+class Sprite;
 
 namespace ui {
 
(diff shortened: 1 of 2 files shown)


cocos/ui/UIText.h
```diff
@@ -322,6 +323,12 @@ class CC_GUI_DLL Text : public Widget
     * Return current effect color value.
     */
     Color4B getEffectColor() const;
+    
+    /**
+     * Provides a way to treat each character like a Sprite.
+     * @warning No support system font.
+     */
+    virtual Sprite * getLetter(int lettetIndex);
```
minggo, line 331: why need `virtual`?

minggo: can add const? such as 
```c++
Sprite * getLetter(int lettetIndex) const
```

newnon:
i've just copied it from Label, it is virtual and non const in Label

PR #17179, opened 2017-01-12
Full unicode support

newnon:
Full range unicode support including emoji symbols


minggo:
Awesome PR. Could you please add a test case for emoji symbols?

newnon:
sorry may be later i'have too much work on my job.

newnon:
emoji test

minggo:
@newnon thanks, i tested on Mac, it seems the emoji font loses color information
screen shot 2017-02-06 at 9 50 29 am

newnon:
No it is just black & white emoji font
To support colour emoji font you need to compile freetype with png support and some small fixes in font rendering classes. I have tried this and it is possible. May be you can ask man who responding for libraries to add png support to freetype?

minggo:
Thanks, yep, i think we should rebuild freetype with png support in future.

sbrednikhin:
What is the status of freetype rebuilding& Wright now it looks like it cant render color emoji.

PR #17190, opened 2017-01-14
fix #14936

### LINKED ISSUE #14936: CallFunc with Spawn bug
Here is the code I use:

```
CallFunc* logFunc = CallFunc::create([] ()
{
    static int i = 0;
    CCLOG("CALL %d", i);
    ++i;
});

Vector<FiniteTimeAction*> vec;
for (int i = 0; i < 5; ++i)
{
    vec.pushBack(logFunc);
}

Sequence* spAct = Sequence::create(vec);

runAction(spAct);
```

The log is as follows:

CALL 0
CALL 1
CALL 2
CALL 3
CALL 4

If I change the code to read `Sequence` as `Spawn` i.e. 

```
CallFunc* logFunc = CallFunc::create([] ()
{
    static int i = 0;
    CCLOG("CALL %d", i);
    ++i;
});

Vector<FiniteTimeAction*> vec;
for (int i = 0; i < 5; ++i)
{
    vec.pushBack(logFunc);
}

Spawn* spAct = Spawn::create(vec);

runAction(spAct);
```

the log is the following:

CALL 0
CALL 1
CALL 2
CALL 3
CALL 4
CALL 5
CALL 6

I have debugged and the fact is that first two actions in `Vector<FiniteTimeAction*> vec;` are being called twice at the end, i.e. CALL 0 and 1 are the same actions as CALL 5 and 6 correspondingly.

myxrome:


minggo:
@myxrome thanks, could you please add the test case into cpp-tests? Then we will avoid it in future. Thanks.

minggo:
thanks

PR #17192, opened 2017-01-14
Added MotionStreak::followNode for easier setup

Rubonnek:
Setting up MotionStreak to follow a node does not seem intuitive enough in my opinion.

This patch is based on @zawasp's [patch](https://github.com/zawasp/cocos2d-x/commit/41214636ba620777e6910b7c310be56fdd0088ba), although slightly modified.

Assuming we want a MotionStreak node to follow another node, coding that with this patch is reduced to the following:

````cpp
auto streak = MotionStreak::create(.3, 3, 16, Color3B::RED, "Images/streak.png" );
addChild(streak);

// to start following another node:
streak->followNode(existing_node, Vec2(10,10));

// to stop following a node:
// streak->stopFollowNode();
````

### DIFF
--- a/cocos/2d/CCMotionStreak.cpp
+++ b/cocos/2d/CCMotionStreak.cpp
@@ -112,10 +112,10 @@ bool MotionStreak::initWithFade(float fade, float minSeg, float stroke, const Co
 
     _stroke = stroke;
     _fadeDelta = 1.0f/fade;
-    
+
     double fps = 1/Director::getInstance()->getAnimationInterval();
     _maxPoints = (int)(fade*fps)+2;
-    
+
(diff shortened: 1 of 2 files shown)


Rubonnek:
This can be done using actions.

Something like the following works:
````cpp
auto streak = MotionStreak::create(.3, 3, 16, Color3B::RED, "streak.png" );
auto follow = RepeatForever::create(
		Sequence::create(
			CallFunc::create(
				[=] () {
				streak->setPosition(node_to_follow->getPosition());
				}
				)
			, nullptr
			)
		);
streak->runAction(follow);
this->addChild(streak);
````

PR #17198, opened 2017-01-16
remove scheduler in time

minggo:


cocos/base/CCScheduler.cpp
```diff
@@ -464,26 +464,15 @@ void Scheduler::schedulePerFrame(const ccSchedulerFunc& callback, void *target,
     HASH_FIND_PTR(_hashForUpdates, &target, hashElement);
     if (hashElement)
     {
-        // check if priority has changed
+        // change priority: should unschedule it first
         if ((*hashElement->list)->priority != priority)
```
cinakyn: I think this line should be like
`if (hashElement->entry->priority != priority)`

minggo: Yep, you are right.

fuyifan:
@minggo this commit can produce lots of warning: _**don't update it again message**_ when loading the cocostudio json file for js environment.

minggo:
it may have problem in cocostudio reader, it added the same scheduler again.

imtrobin:
I confirm it is causing a lot of  "don't update it again message" with cocosstudio reader, because Node::addcomponent will do a scheduleUpdate call.

However, I don't see an unscheduleUpdate in removeComponent, is that a bug? What is this scheduleUpdate for, is it really needed? I comment it out, it seems to work fine for my csb files

minggo:
The log just mean adding the same scheduler again has not effect, it doesn't have any side effect.

imtrobin:
yes, but my question is about the scheduler. The components added are LayoutComponent, ComExtensionData etc. Do they need to be added to scheduleUpdate or is it unnecessary?

minggo:
About `LayoutComponent`, did you mean `UILayoutComponent`? But i don't find `scheduleUpdate ` in it. So could you please be more specific?

imtrobin:
I'm loading csb files exported from cocostudio, when I step through, it has __ui_layout which creates ui::LayoutComponent , and is added using Node::addComponent, which calls scheduleUpdate. Then another component ComExtensionData is created and also added using Node::addComponent, calling ScheduleUpdate again and resulting in the warning. 

I don't see a way to disable in cocosstudio. I don't understand why  Node::addcomponent will need to call scheduleUpdate internally cos I commented it out, it seems to work fine.

imtrobin:
```
bool Node::addComponent(Component *component)
{
    // lazy alloc
    if (!_componentContainer)
        _componentContainer = new (std::nothrow) ComponentContainer(this);
    // WHY SCHEDULE BUT NO UNSCHEDULE WHEN removeComponent
   // Also, is it needed?
    // should enable schedule update, then all components can receive this call back
    scheduleUpdate();
    
    return _componentContainer->add(component);
}

bool Node::removeComponent(const std::string& name)
{
    if (_componentContainer)
        return _componentContainer->remove(name);
    
    return false;
}
```

PR #17203, opened 2017-01-17
Fixed Scheduler bug occured when unschedule target during update.

cinakyn:
After #17198 
Sometimes, unscheduled target's update function is called.
When one target unscheduled its next target in scheduler's doubly linked list, uscheduled next target's update function will be called.
Please refer the test case.


minggo:
yep, you are right, thanks

PR #17206, opened 2017-01-17
Allow to unbind asynchronous texture loading callback with a custom key.

j-jorge:
In order to unbind the callback passed to
`cocos2d::TextureCache::addImageAsync(path, callback)`, one has to
call `cocos2d::TextureCache::unbindImageAsync(path)`. In the cases
where the loading of the same texture is requested from several sources
simultaneously, then none of the source can unbind its own callback
unambiguously.
    
This commit adds an overload of the `addImageAsync` function taking an extra
argument identifying the callback, thus allowing to unbind it unambiguously
in cases where the loading of path is requested by several sources
simultaneously.

### DIFF
--- a/cocos/renderer/CCTextureCache.cpp
+++ b/cocos/renderer/CCTextureCache.cpp
@@ -99,10 +99,17 @@ std::string TextureCache::getDescription() const
 struct TextureCache::AsyncStruct
 {
 public:
-    AsyncStruct(const std::string& fn,const std::function<void(Texture2D*)>& f) : filename(fn), callback(f), pixelFormat(Texture2D::getDefaultAlphaPixelFormat()), loadSuccess(false) {}
+    AsyncStruct
+    ( const std::string& fn,const std::function<void(Texture2D*)>& f,
+      const std::string& key )
+      : filename(fn), callback(f),callbackKey( key ),
+        pixelFormat(Texture2D::getDefaultAlphaPixelFormat()),
+        loadSuccess(false)
+    {}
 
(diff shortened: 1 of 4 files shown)


minggo:
This PR looks good to me, but i don't quite understand what's the issue of current implementation.

j-jorge:
Say object A starts loading a large texture:
```
void A::loadTexture()
{
    Director::getInstance()->getTextureCache()->addImageAsync
      ("large-texture.png", CC_CALLBACK_1(A::textureLoaded, this));
}
```

While the texture is loading, another object B starts to load the same texture:
```
void B::loadTexture()
{
    Director::getInstance()->getTextureCache()->addImageAsync
      ("large-texture.png", CC_CALLBACK_1(B::textureLoaded, this));
}
```
Finally, before the texture is loaded, let's say that B is destroyed and unbinds its callback:
```
B::~B()
{
    Director::getInstance()->getTextureCache()->unbindImageAsync
      ("large-texture.png");
}
```
There `TextureCache::unbindImageAsync` will clear all callbacks associated with the texture's name, including the callback of A. Thus A will never know when the texture is finally loaded.

In my project there are several cases where identical texture loading may happen: when loading the players' profile picture or when loading the flags on the leaderboards. By associating a custom key with the callbacks, I'm able to unbind the callback of B and keep the one of A.

minggo:
thanks for explanation, could you please add a test case?

j-jorge:
Test case added :)

minggo:
thanks

PR #17209, opened 2017-01-17
Compilation fix

j-jorge:
Please review this pull request. These commits fix various compilation errors on Android, Linux, Mac and iOS.


minggo:
How to reproduce the compiling error?

j-jorge:
About commit b81f2bfcab, the fields initialization order warning messages occur for me when compiling on Linux without `-Wno-reorder`. This flag is passed to GCC at line 56 of `cmake/Modules/SetCompilerOptions.cmake` but if you remove it or if you compile without CMake and with `-Werror` then the build will fail. Here is a sample output from GCC:

```
./tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIListViewTest/UIListViewTest.h: In constructor ‘UIListViewTest_Vertical::UIListViewTest_Vertical()’:
./tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIListViewTest/UIListViewTest.h:59:11: warning: ‘UIListViewTest_Vertical::_updateTimer’ will be initialized after [-Wreorder]
     float _updateTimer;
           ^~~~~~~~~~~~
./tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIListViewTest/UIListViewTest.h:58:11: warning:   ‘float UIListViewTest_Vertical::_updateInterval’ [-Wreorder]
     float _updateInterval;
           ^~~~~~~~~~~~~~~
./tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIListViewTest/UIListViewTest.cpp:22:1: warning:   when initialized here [-Wreorder]
 UIListViewTest_Vertical::UIListViewTest_Vertical()
 ^~~~~~~~~~~~~~~~~~~~~~~
```

The missing files in `CMakeLists.txt` and various missing imports result in errors at compile and link time on OSX by typing `cmake . && make` in a freshly cloned repository.

About commit f54660a07, GCC 6 introduces the `-Wmisleading-indentation` warning for misleading indentation, triggerred when an instruction seems to be part of a block while it is semantically outside the block, and the `-Wnonnull-compare` trigerred when a pointer that cannot be null (typically `this`) is compared with `nullptr`. Here is the error messages from `CCArray.cpp`:

```
./cocos/deprecated/CCArray.cpp: In member function ‘virtual cocos2d::__Array* cocos2d::__Array::clone() const’:
./cocos/deprecated/CCArray.h:172:5: error: nonnull argument ‘this’ compared to NULL [-Werror=nonnull-compare]
     if ((__array__) && (__array__)->data->num > 0)                                                                     \
     ^
../../../../../../extensions/cocos/cocos/deprecated/CCArray.cpp:739:5: note: in expansion of macro ‘CCARRAY_FOREACH’
     CCARRAY_FOREACH(this, obj)
     ^~~~~~~~~~~~~~~
```

About commit f4313cab1, I cannot reproduce the error, for an unknown reason.

About commit 994f728f6, some `.jar` files are required for the Android builds (for example the content of `cocos/platform/android/java/libs/`) but they are ignored by `.gitignore`. Thus these files cannot be updated and a `git archive` export excludes them, making the export useless.

cocos/3d/CCMesh.cpp
```diff
@@ -125,13 +125,12 @@ Mesh::Mesh()
 : _skin(nullptr)
 , _visible(true)
 , _isTransparent(false)
+, _force2DQueue(false)
 , _meshIndexData(nullptr)
-, _material(nullptr)
 , _glProgramState(nullptr)
 , _blend(BlendFunc::ALPHA_NON_PREMULTIPLIED)
-, _visibleChanged(nullptr)
```
minggo, line 132: _visibleChanged is missed

cocos/deprecated/CCArray.cpp
```diff
@@ -350,7 +350,8 @@ __Array* __Array::clone() const
     Ref* obj = nullptr;
     Ref* tmpObj = nullptr;
     Clonable* clonable = nullptr;
-    CCARRAY_FOREACH(this, obj)
+    const __Array* self( this );
+    CCARRAY_FOREACH(self, obj)
```
minggo, line 354: don't quite understand it

cocos/deprecated/CCDeprecated.h
```diff
@@ -986,9 +986,6 @@ CC_DEPRECATED_ATTRIBUTE const TransitionScene::Orientation kCCTransitionOrientat
 
 CC_DEPRECATED_ATTRIBUTE typedef TransitionScene::Orientation tOrientation;
 
-CC_DEPRECATED_ATTRIBUTE const int kCCPrioritySystem = Scheduler::PRIORITY_SYSTEM;
-CC_DEPRECATED_ATTRIBUTE const int kCCPriorityNonSystemMin = Scheduler::PRIORITY_NON_SYSTEM_MIN;
-
```
minggo, line 991: why remove these codes

cocos/3d/CCMesh.cpp
j-jorge: `_visibleChanged` is an `std::function` and initializing an `std::function` with `nullptr` is equivalent to using its default constructor. Thus there is no need to keep this line.

cocos/deprecated/CCArray.cpp
j-jorge: The expansion of `CCARRAY_FOREACH` will begin with `if (this)` but since GCC 6 `this` is expected not to be null. Here is a quote from [GCC's release notes](https://gcc.gnu.org/gcc-6/changes.html):

> Value range propagation now assumes that the this pointer of C++ member functions is non-null.

Thus this test triggers the `-Wnonnull-compare` warning of GCC. Using another pointer (`self`) in the test prevents the warning.

cocos/deprecated/CCDeprecated.h
j-jorge: These lines can be kept, I will revert the commit.

cocos/deprecated/CCArray.cpp
minggo: good to know, thanks

cocos/deprecated/CCDeprecated.h
minggo: please revert it, then i will merge this pr, thanks

j-jorge: commit removed from the branch :)

PR #17217, opened 2017-01-18
Add a preprocessor test to disable the logs even in debug.

j-jorge:
Having asserts enabled in debug builds is nice to find various coding mistakes but having a lot of logs can be annoying. This commit adds a preprocessor test allowing to turn off the logs and keep the asserts in debug builds.

### DIFF
--- a/cocos/network/HttpAsynConnection-apple.m
+++ b/cocos/network/HttpAsynConnection-apple.m
@@ -64,7 +64,7 @@
 
 - (void) startRequest:(NSURLRequest *)request
 {
-#ifdef COCOS2D_DEBUG
+#if defined( COCOS2D_DEBUG ) && ( COCOS2D_LOG_LEVEL != 0 )
     NSLog(@"Starting to load %@", srcURL);
 #endif
     
@@ -96,7 +96,7 @@
  **/
 - (void) connection:(NSURLConnection *)connection 
  didReceiveResponse:(NSURLResponse *)response {
-#ifdef COCOS2D_DEBUG
+#if defined( COCOS2D_DEBUG ) && ( COCOS2D_LOG_LEVEL != 0 )
     NSLog(@"Received response from request to url %@", srcURL);
 #endif
     
--- a/cocos/platform/CCPlatformMacros.h
+++ b/cocos/platform/CCPlatformMacros.h
@@ -224,7 +224,7 @@ public: virtual void set##funName(varType var)   \
 
 /// @name Cocos2d debug
 /// @{
-#if !defined(COCOS2D_DEBUG) || COCOS2D_DEBUG == 0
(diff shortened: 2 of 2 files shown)


stevetranby:
Shouldn't this just be global? 
```
#if defined( COCOS2D_DEBUG ) && ( COCOS2D_LOG_LEVEL != 0 )
#define NSLog(..) ...
#else
#define NSLog(..) do{}while(0);
#endif
```

minggo:
assert is nothing about log, if you don't want the log, just don't define those macros.

j-jorge:
@stevetranby `CCLOG()` is indeed defined globally but `NSLog()` is a function from Apple's SDKs thus we should not redefine it. Maybe we should use `CCLOG()` where `NSLog` is currently used.

@minggo I do not see how I can disable one of those macros independently from the other:

In [`base/ccMacros.h`](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/ccMacros.h#L39), `CCASSERT()` is defined as long as `COCOS2D_DEBUG` is defined and greater than zero.

Then, in [`platform/CCPlatformMacros.h`](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/CCPlatformMacros.h#L233), `CCLOG()` is defined under the same criteria.

Thus, if I set `COCOS2D_DEBUG=0` then I will have neither `CCLOG()` nor `CCASSERT()`. The intent of this commit is to add a criterion allowing to enable `CCASSERT()` independently of `CCLOG()`.

stevetranby:
@j-jorge yeah, I realized my mistake after minggo posted re-reading my comment.

I think your goal here makes sense. 

You can both ignore my further comment when deciding to merge this PR or not.

----

I would ask, however, what asserts you're running into that you want to ignore? I would think that those asserts would be better changed to warning logs if it doesn't prevent the game from continuing execution, or your usage of the API should change to include more pre-condition checks.

Another idea is to change asserts to the less intrusive debugger asserts where you can continue past the assert, again only if it makes sense to allow that.

j-jorge:
Just to be clear: the asserts are triggered regularly during development and I want to keep them.

However, Cocos2d-x's logs tend to be noisy in the console as they make it more difficult to find my own logs. For example, `HttpClient` logs its initialization, destruction and other steps with lines like `CCLOG("In the constructor of HttpClient!");` These logs are certainly useful for the _developer_ of `HttpClient` but they are of no use for the ones _using_ `HttpClient`.

Hence I propose this PR allowing the user of Cocos2d-x to benefit from the asserts without being disturbed by the various logs.

stevetranby:
Ah, yeah again while your PR is fine by me, I would argue that those excessive log messages should be changed to CCLOGINFO so that they're ignored except on higher debug levels.

minggo:
I agree with @stevetranby.

PR #17233, opened 2017-01-19
[Android] Software decoding while preloading audio files to avoid some crashes on some specific android devices, and it’s 5x - 15x faster than the old implementation.

dumganhar:
Discussion is here: http://discuss.cocos2d-x.org/t/android-audio-decoding-issues-discussion-please-come-in/34610

### DIFF
--- a/cocos/audio/android/Android.mk
+++ b/cocos/audio/android/Android.mk
@@ -43,7 +43,7 @@ LOCAL_EXPORT_LDLIBS := -lOpenSLES
 LOCAL_C_INCLUDES := $(LOCAL_PATH)/../include \
                     $(LOCAL_PATH)/../.. \
                     $(LOCAL_PATH)/../../platform/android \
-                    $(LOCAL_PATH)/../../../external
+                    $(LOCAL_PATH)/../../../external/android-specific
 
 LOCAL_STATIC_LIBRARIES += libvorbisidec libpvmp3dec
 include $(BUILD_STATIC_LIBRARY)
@@ -68,5 +68,5 @@ LOCAL_C_INCLUDES := $(LOCAL_PATH)/../include \
 
 include $(BUILD_STATIC_LIBRARY)
 
-$(call import-module,tremolo)
-$(call import-module,pvmp3dec)
+$(call import-module,android-specific/tremolo)
+$(call import-module,android-specific/pvmp3dec)
--- a/cocos/audio/android/Android.mk
+++ b/cocos/audio/android/Android.mk
@@ -12,6 +12,11 @@ LOCAL_SRC_FILES := AudioEngine-inl.cpp \
                    CCThreadPool.cpp \
                    AssetFd.cpp \
                    AudioDecoder.cpp \
+                   AudioDecoderProvider.cpp \
+                   AudioDecoderSLES.cpp \
+                   AudioDecoderOgg.cpp \
+                   AudioDecoderMp3.cpp \
+                   AudioDecoderWav.cpp \
                    AudioPlayerProvider.cpp \
                    AudioResampler.cpp \
                    AudioResamplerCubic.cpp \
@@ -26,7 +31,9 @@ LOCAL_SRC_FILES := AudioEngine-inl.cpp \
                    audio_utils/format.c \
                    audio_utils/minifloat.cpp \
                    audio_utils/primitives.c \
-                   utils/Utils.cpp
+                   utils/Utils.cpp \
+                   mp3reader.cpp \
+                   tinysndfile.cpp
 
 
 LOCAL_EXPORT_C_INCLUDES := $(LOCAL_PATH)/../include
@@ -35,8 +42,10 @@ LOCAL_EXPORT_LDLIBS := -lOpenSLES
 
 LOCAL_C_INCLUDES := $(LOCAL_PATH)/../include \
                     $(LOCAL_PATH)/../.. \
-                    $(LOCAL_PATH)/../../platform/android
+                    $(LOCAL_PATH)/../../platform/android \
+                    $(LOCAL_PATH)/../../../external
 
+LOCAL_STATIC_LIBRARIES += libvorbisidec libpvmp3dec
 include $(BUILD_STATIC_LIBRARY)
 
 #SimpleAudioEngine
@@ -58,3 +67,6 @@ LOCAL_C_INCLUDES := $(LOCAL_PATH)/../include \
                     $(LOCAL_PATH)/../../platform/android
 
 include $(BUILD_STATIC_LIBRARY)
+
+$(call import-module,tremolo)
+$(call import-module,pvmp3dec)
--- a/cocos/audio/android/AudioDecoder.cpp
+++ b/cocos/audio/android/AudioDecoder.cpp
@@ -27,89 +27,52 @@ THE SOFTWARE.
 #include "audio/android/AudioDecoder.h"
 #include "audio/android/AudioResampler.h"
 #include "audio/android/PcmBufferProvider.h"
+#include "audio/android/AudioResampler.h"
 
 #include <thread>
 #include <chrono>
 
 namespace cocos2d { namespace experimental {
 
-/* Explicitly requesting SL_IID_ANDROIDSIMPLEBUFFERQUEUE and SL_IID_PREFETCHSTATUS
-* on the AudioPlayer object for decoding, SL_IID_METADATAEXTRACTION for retrieving the
-* format of the decoded audio */
-#define NUM_EXPLICIT_INTERFACES_FOR_PLAYER 3
-
-/* Size of the decode buffer queue */
-#define NB_BUFFERS_IN_QUEUE 4
-
-/* size of the struct to retrieve the PCM format metadata values: the values we're interested in
- * are SLuint32, but it is saved in the data field of a SLMetadataInfo, hence the larger size.
- * Nate that this size is queried and displayed at l.452 for demonstration/test purposes.
- *  */
-#define PCM_METADATA_VALUE_SIZE 32
-
-/* used to detect errors likely to have occurred when the OpenSL ES framework fails to open
- * a resource, for instance because a file URI is invalid, or an HTTP server doesn't respond.
- */
-#define PREFETCHEVENT_ERROR_CANDIDATE (SL_PREFETCHEVENT_STATUSCHANGE | SL_PREFETCHEVENT_FILLLEVELCHANGE)
-
-//-----------------------------------------------------------------
-
-// __SLPlayerMutex is a workaround for https://github.com/cocos2d/cocos2d-x/issues/16849
-static std::mutex __SLPlayerMutex;
-
-static int toBufferSizeInBytes(int bufferSizeInFrames, int sampleSize, int channelCount)
+size_t AudioDecoder::fileRead(void* ptr, size_t size, size_t nmemb, void* datasource)
 {
-    return bufferSizeInFrames * sampleSize * channelCount;
+    AudioDecoder* thiz = (AudioDecoder*)datasource;
+    ssize_t toReadBytes = std::min((ssize_t)(thiz->_fileData.getSize() - thiz->_fileCurrPos), (ssize_t)(nmemb * size));
+    if (toReadBytes > 0)
+    {
+        memcpy(ptr, (unsigned char*) thiz->_fileData.getBytes() + thiz->_fileCurrPos, toReadBytes);
+        thiz->_fileCurrPos += toReadBytes;
+    }
+    // ALOGD("File size: %d, After fileRead _fileCurrPos %d", (int)thiz->_fileData.getSize(), thiz->_fileCurrPos);
+    return toReadBytes;
 }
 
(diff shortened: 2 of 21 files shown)


mquaker:
Great! Worked fine on my Nexus 5 (7.1.1).
thanks

formatCvt:
Thanks! Great work!
No sound issues anymore on Samsung Galaxy Grand
Also this patch fix issues with sound on Google Pixel and Pixel XL (Android 7.1.1)

dumganhar:
@mquaker , @formatCvt , good to know it works for you. :)

Rubonnek:
Works great on the devices I tested the code on:
* LG G3 Stylus (Android 5.0.2)
* Nexuz 7 Nakasi (Android 4.4.4)

piotrros:
So.. will it be merged in 3.15?

dumganhar:
@piotrros , it definitely will be if this PR fixes most issues developers reported. :)

dumganhar:
[ci rebuild]

dumganhar:
UPDATE: Supports to decode WAV file. Decoding WAV file will be 20x faster. Yeah. :)

ryanxia2016:
@dumganhar
I had merged the PR to our cocos2dx version (a modified Version based on 3.3)
when complie the asm files, I got the warrning like:
clang: warning: argument unused during compilation: '-MMD'
clang: warning: argument unused during compilation: '-MP'
clang: warning: argument unused during compilation: '-MF ./obj/local/armeabi/objs/pvmp3dec/src/asm/pvmp3_polyphase_filter_window_gcc.o.d'
clang: warning: argument unused during compilation: '-fpic'
clang: warning: argument unused during compilation: '-ffunction-sections'
clang: warning: argument unused during compilation: '-funwind-tables'
clang: warning: argument unused during compilation: '-fstack-protector'
clang: warning: argument unused during compilation: '-mtune=xscale'

After remove LOCAL_CFLAGS += -Werror from android.mk, the .so file built success.
and the final APK seemed works right on my devices.

and ,is there any risk to complie the code with old ndk version and  gcc4.8? (I noticed in androd.mk LOCAL_CLANG is setted to true)

now we are still using NDK R9D/GCC4.8 , because we found some strange Runtime SigFault Errors with R10E/CLANG...

Any Suggestion?

Dauch:
hi, I merged those file in my project, **It preload mp3 take really long time(12.0 KB take 2 sec.) in x86 machine**(use simulator in the market like https://www.yeshen.com ), but it's really OK for the physics machine

ryanxia2016:
@dumganhar

and, since the software decoding is ready, the android/jni/cddandroidAndroidJavaEngine.cpp file may need to update? it maybe the best way for old projects based on SimpleAudioEngine to switch to the new implementation.

piotrros:
But we shouldn't remove SimpleAudioEngine imo. New engine cannot manage playing more than 24 sounds at once (I'm saying about the same sounds few at once), while SimpleAudioEngine will handle it no problem.

ryanxia2016:
@piotrros 
AudioEngine::setMaxAudioInstance(32) can not fix the 24 limit?

dumganhar:
@piotrros , currently, AudioEngine uses the `Android source code` (https://android.googlesource.com/platform/frameworks/av/+/master/services/audioflinger/AudioMixer.h) 

Please refer to that header file, it said:
```c++
class AudioMixer
{
public:
                            AudioMixer(size_t frameCount, uint32_t sampleRate,
                                       uint32_t maxNumTracks = MAX_NUM_TRACKS);
    /*virtual*/             ~AudioMixer();  // non-virtual saves a v-table, restore if sub-classed
    // This mixer has a hard-coded upper limit of 32 active track inputs.
    // Adding support for > 32 tracks would require more than simply changing this value.
    static const uint32_t MAX_NUM_TRACKS = 32;
    // maximum number of channels supported by the mixer
```

It means that android internal only support 32 tracks at most.
You said `SoundPool` supports more than 32 tracks? If so , I guess SoundPool uses a different strategy when play more than 32 sounds at the same time.

Now, cocos just ignores (don't play) the newest audio if the audio count is greater than 32.
I guess SoundPool will stop the oldest audio, then play the newest audio. In this way, it looks like it could play more 32 audios simultaneously. But in fact, it couldn't since Android Flinger Audio Framework only supports 32 audios.

dumganhar:
@piotrros , and you could see how we initialize SoundPool:
image

```java
  /**
     * Constructor. Constructs a SoundPool object with the following
     * characteristics:
     *
     * @param maxStreams the maximum number of simultaneous streams for this
     *                   SoundPool object
     * @param streamType the audio stream type as described in AudioManager 
     *                   For example, game applications will normally use
     *                   {@link AudioManager#STREAM_MUSIC}.
     * @param srcQuality the sample-rate converter quality. Currently has no
     *                   effect. Use 0 for the default.
     * @return a SoundPool object, or null if creation failed
     * @deprecated use {@link SoundPool.Builder} instead to create and configure a
     *     SoundPool instance
     */
    public SoundPool(int maxStreams, int streamType, int srcQuality) {
        this(maxStreams,
                new AudioAttributes.Builder().setInternalLegacyStreamType(streamType).build());
    }
```

It means that currently `Cocos2dxSound` only set 3 or 5 max streams to `SoundPool`, only 3 or 5 audios could be played in the meanwhile.

Dig into the android source code more:
```c++
SoundChannel* SoundPool::allocateChannel_l(int priority, int sampleID)
{
    List<SoundChannel*>::iterator iter;
    SoundChannel* channel = NULL;

    // check if channel for given sampleID still available
    if (!mChannels.empty()) {
        for (iter = mChannels.begin(); iter != mChannels.end(); ++iter) {
            if (sampleID == (*iter)->getPrevSampleID() && (*iter)->state() == SoundChannel::IDLE) {
                channel = *iter;
                mChannels.erase(iter);
                ALOGV("Allocated recycled channel for same sampleID");
                break;
            }
        }
    }

    // allocate any channel
    if (!channel && !mChannels.empty()) {
        iter = mChannels.begin();  // James:  Get the first channel instance
        if (priority >= (*iter)->priority()) {
            channel = *iter;
            mChannels.erase(iter);  // James: Erase the first channel instance.
            ALOGV("Allocated active channel");
        }
    }

    // update priority and put it back in the list
    if (channel) {
        channel->setPriority(priority);
        for (iter = mChannels.begin(); iter != mChannels.end(); ++iter) {
            if (priority < (*iter)->priority()) {
                break;
            }
        }
        mChannels.insert(iter, channel);  // James: Insert the new channel instance. The total instance will be less than 3 or 5 since Cocos2dxSound.java initializes SoundPool with 3 or 5. And SoundPool supports only 32 sound channels.
    }
    return channel;
}
```

Dauch:
Have you try on X86 machine @dumganhar

ryanxia2016:
@Dauch

I had tested on ATOM devices with x86 .so, it's ok.

dumganhar:
@Dauch, I tried, but I couldn't connect to the emulator by `adb` which in the latest Android SDK.
@ryanxia2016 , thanks for your testing. :)

piotrros:
Ok, I see. Then maybe AudioEngine should have the same behaviour? Or at least some setting to do so.

dumganhar:
@piotrros , yes, I think it should be implemented like which in `SoundPool`.

ryanxia2016:
@dumganhar 
how about the question I post yesterday:
======================
I had merged the PR to our cocos2dx version (a modified Version based on 3.3)
when complie the asm files, I got the warrning like:
clang: warning: argument unused during compilation: '-MMD'
clang: warning: argument unused during compilation: '-MP'
clang: warning: argument unused during compilation: '-MF ./obj/local/armeabi/objs/pvmp3dec/src/asm/pvmp3_polyphase_filter_window_gcc.o.d'
clang: warning: argument unused during compilation: '-fpic'
clang: warning: argument unused during compilation: '-ffunction-sections'
clang: warning: argument unused during compilation: '-funwind-tables'
clang: warning: argument unused during compilation: '-fstack-protector'
clang: warning: argument unused during compilation: '-mtune=xscale'

After remove LOCAL_CFLAGS += -Werror from android.mk, the .so file built success.
and the final APK seemed works right on my devices.

and ,is there any risk to complie the code with old ndk version and gcc4.8? (I noticed in androd.mk LOCAL_CLANG is setted to true)

now we are still using NDK R9D/GCC4.8 , because we found some strange Runtime SigFault Errors with R10E/CLANG...

Any Suggestion?
===============================================

dumganhar:
@ryanxia2016 , currently, we're using `gcc` to build project.
It seems that in v3.3, you're using `clang`.
We suggest developer to keep using `gcc` since `clang` isn't stable by now.

Try to modify `tools/cocos2d-console/plugins/plugin_compile/build_android.py` 

```python
    def get_toolchain_version(self, ndk_root, compile_obj):
        return '4.9'  // James: force using gcc 4.9 rather than clang.
```

ryanxia2016:
@dumganhar 
Thanks a lot.

and, which ndk version is prefred now ? :-)

dumganhar:
@ryanxia2016 , probably `android-ndk-n10e` is more stable.

ryanxia2016:
@dumganhar
Thank you.

dumganhar:
@piotrros , I created an issue there https://github.com/cocos2d/cocos2d-x/issues/17387.
If you have any suggestion, please reply in that thread.

Dauch:
It's strange, may be not the x86 machine's problem, I switch [海马玩模拟器](http://droid4x.haimawan.com) and [夜神模拟器](https://www.yeshen.com)，夜神模拟器 I test out is **preload mp3 take really long time(12.0 KB take 2 sec.)**， but 海马玩模拟器 is ok.. I don't know why.... you could have a try @dumganhar @ryanxia2016

dumganhar:
@Dauch , I will check it when I have time.

Dauch:
thanks.

Dauch:
hi, since you have already install the 夜神Emulator 

> @Dauch, I tried, but I couldn't connect to the emulator by adb which in the latest Android SDK.

you don't need to connect to the emulator, you could just 
```
local useTime = os.clock()
audio.playSound("xxx.mp3")
useTime = os.clock() - useTime

local pText = ccui.Text:create()
pText:setString("useTime :" .. useTime )
```
you can see the use time on screen.
@dumganhar

dumganhar:
@Dauch , but I have to debug to find out why it cost so many time on `Ye shen` Emulator, so I have to see the log output.

Dauch:
Since you had check out it is slow, you can check this
[adb debug](http://jingyan.baidu.com/article/359911f55f327257ff030648.html)

> 先启动夜神模拟器
然后运行cmd命令，cd到夜神安装目录，执行命令
nox_adb.exe connect 127.0.0.1:62001
连接到模拟器

dumganhar:
@Dauch, I write a test case base on cpp-empty-test and run it on YeShen emulator.
The decoding work is really fast, it only wastes 13ms.
I uses `tests/cpp-tests/Resources/audio/SoundEffectsFX009/FX082.mp3` to test.
So I couldn't reproduce this issue. Could you upload the audio file you're using for reproducing this issue?

Dauch:
Sorry for delay. I first test out that when I enter the game is normal(really fast). but I went to the center game.(It use some memory), it turn out slow preload the sound. (2+Sec.)
Then I wrote some code in the first Enter the game
```
local useTime = os.clock()
audio.playSound("xxx1.mp3")
useTime = os.clock() - useTime
print("1useTime ", useTime)

DelayCall.delay(function()
    for i = 1, 50 do
        ui:clone():addTo(self):setVisible(false)
    end
    DelayCall.delay(function()
        local useTime = os.clock()
        audio.playSound("xxx2.mp3")
        useTime = os.clock() - useTime
        print("2useTime ", useTime)
    end, 3)
end, 5)
```
then it turn out is slow(2+Sec.)
(ui is the ui resource.)
It's not reach the top memory limit, not crash.

dumganhar:
@Dauch , Since I could not reproduce the problem you said, could you make a demo which bases on cpp-template or lua-template?

Dauch:
here is memory used:


内存（专用工作集）|  内存（专用字节）|效果
-----------------------|-----------------------|-----------------------
66M|93M|这个时候播放是正常的(几毫米)
150M|180M|这个时候播放（未预加载音频文件）延迟2秒的

dumganhar:
@Dauch, what I mean is to make a simple project to reproduce the issue you said.
Then I could run the project on YeShen emulator to reproduce it.

Dauch:
@dumganhar you can use your project, just make the momory use higher then 180M, you will reproduce it on YeShen emulator.
If you can't reproduce  it, I will try make a demo.

dumganhar:
How to make the memory cost higher than 180M?
Could you provide both log output while memory cost is greater than 180 and less than 180?
I guess you could use `nox_adb.exe` to get the log.
I wanna check the difference if you could help to grab the log.

Dauch:
wait

dumganhar:
@Dauch, I created an issue at https://github.com/cocos2d/cocos2d-x/issues/17413 and will fix it soon.

Dauch:
It work, Thanks @dumganhar

Issue #8764, opened 2014-10-16
DrawNode - can't change opacity


zawasp:
I just noticed DrawNode doesn't support opacity changes. Once the item is drawn, setOpacity, FadeIn, FadeOut actions etc. have no effect on DrawNode.

Please add this feature.

minggo:
It is a new feature.
We will fix it in next version if we accept it since it is close to releasing v3.3.

metalim:
Any update on this one? It's obvious DrawNode isn't complete without opacity. From developer's point of view there's not much difference between image and DrawNode.

purepure:
I am looking for this issue too.

liangma87:
is the fix released?

slackmoehrle:
Not yet. I am working on a test case to demonstrate some DrawNode issues and I will throw this in too.

dapetcu21:
Any news of this? I kinda need it for my game

zilongshanren:
@zawasp  I have added the opacity support to DrawNode at https://github.com/cocos2d/cocos2d-x/pull/17248

please give it a try.

PR #17248, opened 2017-01-23
add opacity support to DrawNode

zilongshanren:
fixed issue: https://github.com/cocos2d/cocos2d-x/issues/8764


davidchai-6waves:
In cocos/renderer/ccShader_PositionColorLengthTexture.vert, should "uniform float u_alpha;" place outside the if else block?

minggo:
@davidchai-6waves good catch. I just tested in mac. I will send a pr to fix it.

minggo:
fixed in https://github.com/cocos2d/cocos2d-x/pull/17324

intmainreturn00:
At what version of the engine It should work? 
I tested with 3-13 and it doesen't work..

PR #17259, opened 2017-01-30
Make extensions work out of the box in Linux

Rubonnek:
`extensions/GUI/CCControlExtension/CCControlStepper.h` contains the [line](https://github.com/cocos2d/cocos2d-x/blob/v3/extensions/GUI/CCControlExtension/CCControlStepper.h#L33):
````cpp
#include "extensions/ExtensionExport.h"
````
and `extensions/assets-manager/Manifest.h` contains the [line](https://github.com/cocos2d/cocos2d-x/blob/v3/extensions/assets-manager/Manifest.h#L37):
````cpp 
#include "json/document-wrapper.h"
````

These two headers, along with other files under the `extensions` directory assume that `${COCOS2D_ROOT}` and `${COCOS2D_ROOT}/external`, respectively, are in included in the list of paths used by the compiler when it searches for headers, but this is not reflected in `CMakeLists.txt` for Linux.

I think we should include these paths by default.

### DIFF
--- a/templates/cpp-template-default/CMakeLists.txt
+++ b/templates/cpp-template-default/CMakeLists.txt
@@ -111,7 +111,9 @@ endif()
 include_directories(
   /usr/local/include/GLFW
   /usr/include/GLFW
+  ${COCOS2D_ROOT}
   ${COCOS2D_ROOT}/cocos
+  ${COCOS2D_ROOT}/external
   ${COCOS2D_ROOT}/cocos/platform
   ${COCOS2D_ROOT}/cocos/audio/include/
   Classes

### CODE REVIEW

File: templates/cpp-template-default/CMakeLists.txt [hunk 1]
```diff
@@ -111,7 +111,9 @@ endif()
 include_directories(
   /usr/local/include/GLFW
   /usr/include/GLFW
+  ${COCOS2D_ROOT}
```
dumganhar, line 114: Is `${COCOS2D_ROOT}/external` enough?
Is `${COCOS2D_ROOT}` needed actually?

Rubonnek: Both are needed as far as I can tell unless we modify the headers themselves. Here are the full details:

First create a new project with the cocos version in `HEAD`.

````shell
cocos new -l cpp extensionstest

````

Add the following block somewhere in HelloWorldScene::init():

````cpp
auto minusSprite = Sprite::create("HelloWorld.png");
auto plusSprite = Sprite::create("HelloWorld.png");
auto stepper = extension::ControlStepper::create(minusSprite, plusSprite);
stepper->setScale(.5);
stepper->setStepValue(.05);
stepper->setMinimumValue(1);
stepper->setMaximumValue(10);
stepper->setValue(1);
stepper->setPosition(visibleSize * 9.f / 10.f);
this->addChild(stepper);
````

After trying to compile (i.e. `cocos run -p linux -j$(nproc)`), you'll get:

````cpp
/path/to/extensionstest/Classes/HelloWorldScene.cpp:69:20: error: ‘extension’ has not been declared
auto stepper = extension::ControlStepper::create(minusSprite, plusSprite);
````

Looking at the directory tree from the root of the project, we get:

```shell
$ cd ${PROJECT_ROOT} && tree -d -L 3
.
├── Classes
├── cocos2d
│   ...
│   ├── extensions
│   │   ├── assets-manager
│   │   ├── GUI
│   │   ├── Particle3D
│   │   └── physics-nodes
````

which means that including the following in `HelloWorldScene.cpp`, without modifying the `CMakeLists.txt`, should work:

```cpp
#include <../extensions/cocos-ext.h>
````

You'll get the following error:

````cpp
/path/to/extensionstest/cocos2d/cocos/../extensions/GUI/CCControlExtension/CCControlUtils.h:39:40: fatal error: extensions/ExtensionExport.h: No such file or directory
 #include "extensions/ExtensionExport.h"
````

And this is why we need to include `${COCOS2D_ROOT}` to the `include_directories` list in `CMakeLists.txt`.

Fixing the `#include` in `HelloWorldScene.cpp` to the following:

```cpp
#include <extensions/cocos-ext.h>
````

Adding `${COCOS2D_ROOT}` to `CMakeLists.txt`, and recompiling, you'll get the following error:

````cpp
/path/to/extensionstest/cocos2d/cocos/../extensions/assets-manager/Manifest.h:37:35: fatal error: json/document-wrapper.h: No such file or directory
 #include "json/document-wrapper.h"
````

Looking at the directory tree again:

````shell
$ cd ${PROJECT_ROOT} && tree -d -L 3
.
├── Classes
├── cocos2d
│   ...
│   ├── external
│   │   ├...
│   │   ├── json
````

It becomes clear that we also have to add `${COCOS2D_ROOT}/external` to the `include_directories` in `CMakeLists.txt` because, again, the headers under the `external` directory assume that `${COCOS2D_ROOT}/external` has been added to the `include_directories` in `CMakeLists.txt` but that's not the case. After including `${COCOS2D_ROOT}/external` in `CMakeLists.txt` as in this pull request and recompiling once again, the scene will run.

PR #17270, opened 2017-02-02
Add setting for default value to enable antialias

giginet:
I developed a retro-styled game. So I'd like to disable antialias when I scale textures.

I added the setting whether to enable antialias for all textures.

```cpp
// Disable antialias for all textures
Texture2D::setDefaultAntialiasEnabled(false);
```

Could you review this? 🙇 

antialias

### DIFF
--- a/cocos/renderer/CCTexture2D.cpp
+++ b/cocos/renderer/CCTexture2D.cpp
@@ -120,6 +120,10 @@ const Texture2D::PixelFormatInfoMap Texture2D::_pixelFormatInfoTables(TexturePix
 // Default is: RGBA8888 (32-bit textures)
 static Texture2D::PixelFormat g_defaultAlphaPixelFormat = Texture2D::PixelFormat::DEFAULT;
 
+// The default value whether to apply antialias
+// Default is: true
+static bool g_defaultAntialiasEnabled = true;
+
 //////////////////////////////////////////////////////////////////////////
 //convertor function
 
@@ -442,7 +446,7 @@ Texture2D::Texture2D()
 , _hasPremultipliedAlpha(false)
 , _hasMipmaps(false)
 , _shaderProgram(nullptr)
-, _antialiasEnabled(true)
+, _antialiasEnabled(g_defaultAntialiasEnabled)
 , _ninePatchInfo(nullptr)
 , _valid(true)
 , _alphaTexture(nullptr)
@@ -1415,6 +1419,16 @@ void Texture2D::setDefaultAlphaPixelFormat(Texture2D::PixelFormat format)
 Texture2D::PixelFormat Texture2D::getDefaultAlphaPixelFormat()
 {
     return g_defaultAlphaPixelFormat;
+}
+
+void Texture2D::setDefaultAntialiasEnabled(bool enabled)
+{
+    g_defaultAntialiasEnabled = enabled;
+}
+
+bool Texture2D::getDefaultAntialiasEnabled()
+{
+    return g_defaultAntialiasEnabled;
 }
 
 unsigned int Texture2D::getBitsPerPixelForFormat(Texture2D::PixelFormat format) const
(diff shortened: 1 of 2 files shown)


mogemimi:
Why do you use it instead of `setAntiAliasTexParameters` and `setAliasTexParameters` ?

giginet:
> Why do you use it instead of setAntiAliasTexParameters and setAliasTexParameters ?

I think these methods seems to be for each textures.
I would like to disable antialias for all textures globally.

mogemimi:
@giginet I got it, thanks! These functions looks useful but I guess we should avoid using a global state variable if possible, such as `g_defaultAntialiasEnabled` for a better api and unit testability.

giginet:
>  I guess we should avoid using a global state variable if possible

I agree with you 😅 

I referred implementation of `defaultAlphaPixelFormat`. Unfortunately, it uses the global variable.

https://github.com/cocos2d/cocos2d-x/pull/17270/files#diff-82700ca74ac4671b96a820e09fb43248R121

Can I use the class variable instead? How do you think?

mogemimi:
My personal opinion is that `setDefaultAlphaPixelFormat` should be marked as deprecated in favour of explicitly specifying a pixel format with `initWithImage(Image*, PixelFormat)`, because that function doesn't support PVR file format, and users are confused why there are two ways.
However it has defined since cocos2d-x v0.8 and derived from cocos2d-iphone so we will unavoidably need to keep it for backwards compatibility.

If you want to specify the `_antialiasEnabled` parameter when initializing a Texture2D object, `setAliasTexParameters` can't be used before calling `initWithData` (or `initWithMipmaps`), then you can add/submit the following functions instead:

```cpp
bool initWithData(const void *data, ssize_t dataLen, PixelFormat pixelFormat,
    int pixelsWide, int pixelsHigh, const Size& /*contentSize*/, bool antialiasEnabled);

bool initWithMipmaps(MipmapInfo* mipmaps, int mipmapsNum, PixelFormat pixelFormat,
    int pixelsWide, int pixelsHigh, bool antialiasEnabled);
```

cpp:

```cpp
bool Texture2D::initWithData(const void *data, ssize_t dataLen,
    PixelFormat pixelFormat, int pixelsWide, int pixelsHigh, const Size& contentSize)
{
    return initWithData(data, dataLen, pixelFormat, pixelsWide, pixelsHigh, contentSize, true);
}

bool Texture2D::initWithData(const void *data, ssize_t dataLen,
    PixelFormat pixelFormat, int pixelsWide, int pixelsHigh, const Size& contentSize, bool antialiasEnabled)
{
    _antialiasEnabled = antialiasEnabled;

    ...
}

bool Texture2D::initWithMipmaps(MipmapInfo* mipmaps, int mipmapsNum,
    PixelFormat pixelFormat, int pixelsWide, int pixelsHigh)
{
    return initWithMipmaps(mipmaps, mipmapsNum, pixelFormat, pixelsWide, pixelsHigh, true);
}

bool Texture2D::initWithMipmaps(MipmapInfo* mipmaps, int mipmapsNum,
    PixelFormat pixelFormat, int pixelsWide, int pixelsHigh, bool antialiasEnabled)
{
    _antialiasEnabled = antialiasEnabled;

    ...
}
```

usage:

```cpp
auto texture = new (std::nothrow) Texture2D();
texure->initWithData(..., true);  // linear sampler

// or
texure->initWithData(..., false); // point sampler
```

Btw, the modern graphics API (such as OpenGL ES 3.0, Metal and Vulkan) separates sampler state from texture object. If there are any plans for adding support for modern renderers in the near future or cocos2d-x v4, I think that it would be better not to add a new/another method (which has equivalent functionality to `setAliasTexParameters`) into Texture2D for only setting sampler state so far. It's easy to make new functions, but hard to get rid of them for backwards compatibility.

Thank you for reading this. :relaxed:

minggo:
@mogemimi 's suggestion is good.

giginet:
I thought @mogemimi 's suggestion seems to be good. 
However, this solution is same as calling `setAliasTexParameters` for every texture.

I thought again, I shouldn't add new API. If we want to disable antialias for all textures, it is better to modify the implementation of `TextureCache` in each game not modify cocos2d-x.

Thanks for your opinion. 🙇

PR #17278, opened 2017-02-04
fix the bug that ect1 texture lost on android

minggo:
the commit do the following things:
- reload ect1 related shaders when comes from background
- reload alpha texture when comes from background


Canon0223:
// etc1 support check whether alpha texture exists & load it
         auto alphaFile = vt->_fileName + TextureCache::getETC1AlphaFileSuffix();
         Texture2D::PixelFormat oldPixelFormat = Texture2D::getDefaultAlphaPixelFormat();	            reloadTexture(vt->_texture->getAlphaTexture(), alphaFile, vt->_pixelFormat);

//After reload alphaTexture, you should setAlphaTexture,Like:
vt->_texture->setAlphaTexture(vt->_texture->getAlphaTexture());

//otherwise,ect1 texture will lost still

PR #17295, opened 2017-02-06
Signal the scroll once the list is updated in cocos2d::extension::TableView

j-jorge:
This commit moves the call of `TableViewDelegate::scrollViewDidScroll` such that it becomes the last step in `TableView::scrollViewDidScroll`. Since the delegate may want to read the state of the table's cells or the view's position, it seems necessary to notify it once that the view has updated its state.


minggo:
It looks good to me. But let's ask @zilongshanren to double check it.

PR #17296, opened 2017-02-06
Assert that a node cannot be inserted as a child of itself.

j-jorge:
This commits adds an assert to ensure that a node is not inserted as a descendent of itself.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -940,6 +940,21 @@ void Node::addChild(Node* child, int localZOrder, const std::string &name)
 
 void Node::addChildHelper(Node* child, int localZOrder, int tag, const std::string &name, bool setTag)
 {
+    auto assertNotSelfChild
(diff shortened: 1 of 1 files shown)


cocos/2d/CCNode.cpp
```diff
@@ -940,6 +940,21 @@ void Node::addChild(Node* child, int localZOrder, const std::string &name)
...
+        ( [ this, child ]() -> bool
+          {
+              for ( Node* parent( getParent() ); parent != nullptr;
+                    parent = parent->getParent() )
+                  if ( parent == child )
+                      return false;
+              
+              return true;
+          } );
+    (void)assertNotSelfChild;
```
minggo, line 953: this line is just to fix warning in release mode?

j-jorge: Yes it is. Without it the compiler complains about "unused variable assertNotSelfChild".

PR #17298, opened 2017-02-06
Disable subpixel rendering for label textures.

j-jorge:
Following [this discussion](http://discuss.cocos2d-x.org/t/set-nodes-position-on-integer-screen-coordinates/31071), this PR allows to disable rendering labels on non integer coordinates, thus avoiding the blurry rendering currently occurring on them.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -1263,8 +1263,11 @@ void Label::createSpriteForSystemFont(const FontDefinition& fontDef)
     _textSprite->setCameraMask(getCameraMask());
     _textSprite->setGlobalZOrder(getGlobalZOrder());
     _textSprite->setAnchorPoint(Vec2::ANCHOR_BOTTOM_LEFT);
+    _textSprite->setSubpixelRendering(false);
+
     this->setContentSize(_textSprite->getContentSize());
     texture->release();
+
     if (_blendFuncDirty)
     {
         _textSprite->setBlendFunc(_blendFunc);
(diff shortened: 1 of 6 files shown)


minggo:
I think we should reuse `CC_NODE_RENDER_SUBPIXEL`. And we can just adjust the vertex coordinate once but in renderer that will do the calculation every frame.

j-jorge:
I disagree about using `CC_NODE_RENDER_SUBPIXEL` for this because in my opinion the developer should have the choice to enable or disable it on a case by case basis. In my case, only the labels are ugly when rendered on subpixels but everything else is perfect. I would not want to affect the display of everything just to fix the labels.

For the renderer I do not see how to precompute the coordinates since the final vertex position is actually computed here, using the `modelView`. Even if I use integer coordinates upstream, the model view may shift them to non-integer coordinates :confused:

minggo:
I think all textures will meet the issue because we use float type for position. I think it is not a good way fix the issue like this though i don't have better solution. May be you should make sure label is added into a scene and the label's position is integer by yourself.

j-jorge:
I have tested your proposal of reusing `CC_NODE_RENDER_SUBPIXEL` on a Linux build and it seems indeed OK for me. I am going to test it on an Android build and adjust the PR accordingly.

minggo:
@j-jorge any update?

j-jorge:
It seems perfect on Android too. Feel free to close this PR :)

minggo:
thanks for the feedback

Issue #17249, opened 2017-01-23
Mac fullscreen works incorrectly when Retina display is scaled from Preferences


Bilalmirza:
- cocos2d-x version: v3.14.1
- devices test on: MAC
- developing environments
   MacOSx

Steps to Reproduce:

1. Go to system preferences->displays-> set resolution to scaled
2. Select more Space
https://www.dropbox.com/s/dg78y8lm6m85zkz/Screen%20Shot%202017-01-23%20at%203.04.50%20PM.png?dl=0

3. Run Mac build ***Fullscreen*** on xcode
4. Incorrect results (scaling is wrong)

MileyHollenberg:
If you select Larger Text (the first option) the issue happens as well, only with the three options in the middle does the scaling seem right

Edit:

I was just about to look into this issue myself but it appears to be working fine now, the only thing that changed (that might have some effect on this) is that I plugged in an external 1080P monitor today, other then that I haven't made any changed though I did experience this bug as well yesterday.

PR #17309, opened 2017-02-07
Workaround fix for issue #17249

MileyHollenberg:
There is an issue with Mac OSX not wanting to create a proper fullscreen GLFW context, I have read elsewhere that there may be some issues with certain width/height combinations on Mac that cause OSX to not create the proper context and create some odd scaling issues. 

This issue appears mostly, though not always, (for me) when setting the display scaling option within Mac OSX to either "Larger Text" (1024 x 640) or "More Space" (1920 x 1200) on a Retina display (2800 x 1800)

Instead of just having a createWithFullScreen initializer with a viewName and optionally the video mode and monitor (which you can't retrieve without first creating a window or initializing GLFW yourself which would cause issues with Cocos2D-X trying to initialize it again) I created a new initializer which automatically tries to create a fullscreen window with the highest possible screen size the current primary display can offer

In my AppDelegate.cpp I've had to change the applicationDidFinishLaunching a bit when creating the glview in order to properly support the Retina Display and it's scaling.

```
if(!glview) {
#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC) || (CC_TARGET_PLATFORM == CC_PLATFORM_LINUX)
	glview = GLViewImpl::createWithFullScreen("Cocos2D-X", true);
	if (((GLViewImpl *)glview)->getRetinaFactor() > 1.0)
	{
		((GLViewImpl *)glview)->enableRetina(true);
		director->setContentScaleFactor(2.0);
	}
#else
        glview = GLViewImpl::create("Cocos2D-X");
#endif
        director->setOpenGLView(glview);
}
```

I have not yet tested this code on either Windows or Linux, nor have I tested this code on Mac with the primary screen not being a retina screen.

### DIFF
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
@@ -348,6 +348,17 @@ GLViewImpl* GLViewImpl::createWithFullScreen(const std::string& viewName)
     return nullptr;
 }
 
+GLViewImpl* GLViewImpl::createWithFullScreen(const std::string& viewName, bool maxSize)
+{
+	auto ret = new (std::nothrow) GLViewImpl();
+	if(ret && ret->initWithFullScreen(viewName, maxSize)) {
+		ret->autorelease();
+		return ret;
+	}
+	CC_SAFE_DELETE(ret);
+	return nullptr;
+}
+
 GLViewImpl* GLViewImpl::createWithFullScreen(const std::string& viewName, const GLFWvidmode &videoMode, GLFWmonitor *monitor)
 {
     auto ret = new (std::nothrow) GLViewImpl();
@@ -460,6 +471,24 @@ bool GLViewImpl::initWithFullScreen(const std::string& viewName)
 
     const GLFWvidmode* videoMode = glfwGetVideoMode(_monitor);
     return initWithRect(viewName, Rect(0, 0, videoMode->width, videoMode->height), 1.0f, false);
+}
+
+bool GLViewImpl::initWithFullScreen(const std::string& viewName, bool maxSize)
+{
+	if (!maxSize)
+		return initWithFullScreen(viewName);
+	
+	//Get the monitor
+	_monitor = glfwGetPrimaryMonitor();
+	if (nullptr == _monitor)
+		return false;
+	
+	//Get all the possible video modes
+	int count;
+	const GLFWvidmode* modes = glfwGetVideoModes(_monitor, &count);
+	GLFWvidmode videoMode = *(modes + count - 1);//Select the last mode since they are in ascending order of refresh rate and then screen size
+	
+	return initWithRect(viewName, Rect(0, 0, videoMode.width, videoMode.height), 1.0f, false);
 }
 
 bool GLViewImpl::initWithFullscreen(const std::string &viewname, const GLFWvidmode &videoMode, GLFWmonitor *monitor)
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.h
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.h
@@ -60,6 +60,7 @@ public:
     static GLViewImpl* create(const std::string& viewName, bool resizable);
     static GLViewImpl* createWithRect(const std::string& viewName, Rect size, float frameZoomFactor = 1.0f, bool resizable = false);
     static GLViewImpl* createWithFullScreen(const std::string& viewName);
+	static GLViewImpl* createWithFullScreen(const std::string& viewName, bool maxSize)
     static GLViewImpl* createWithFullScreen(const std::string& viewName, const GLFWvidmode &videoMode, GLFWmonitor *monitor);
 
     /*
@@ -126,6 +127,7 @@ protected:
 
     bool initWithRect(const std::string& viewName, Rect rect, float frameZoomFactor, bool resizable);
     bool initWithFullScreen(const std::string& viewName);
+	bool initWithFullScreen(const std::string& viewName, bool maxSize);
     bool initWithFullscreen(const std::string& viewname, const GLFWvidmode &videoMode, GLFWmonitor *monitor);
 
     bool initGlew();
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.h
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.h
@@ -60,7 +60,7 @@ public:
     static GLViewImpl* create(const std::string& viewName, bool resizable);
     static GLViewImpl* createWithRect(const std::string& viewName, Rect size, float frameZoomFactor = 1.0f, bool resizable = false);
     static GLViewImpl* createWithFullScreen(const std::string& viewName);
-	static GLViewImpl* createWithFullScreen(const std::string& viewName, bool maxSize)
+	static GLViewImpl* createWithFullScreen(const std::string& viewName, bool maxSize);
     static GLViewImpl* createWithFullScreen(const std::string& viewName, const GLFWvidmode &videoMode, GLFWmonitor *monitor);
 
     /*


minggo:
Sorry, i can't catch your idea. What's the issue? How to reproduce it?

MileyHollenberg:
The issue is that on OSX it handles display scaling in a odd way for GLFW to create a fullscreen view, the largest and smallest display scaling options don't allow Cocos2D to create a OpenGL surface at that exact size (they appear smaller and only a quarter is shown due to the retina scaling which is inactive).

This workaround fix basically ignores what OSX says about the current scale/screen resolution and just creates a fullscreen windows which will take the entire resolution of the display itself.

You can reproduce it with a Macbook which has a Retina display, just select either the "Larger Text" or "More Space" options (as shown in https://github.com/cocos2d/cocos2d-x/issues/17249) and try to create a fullscreen view with a rectangle (half size of the screen so the borders are visible on all edges) in the center of it. The Larger Text and More Space will show a rectangle in the top right corner and the other 3 options wil show the rectangle in the center of the screen. My workaround fix ignores the current resolution and takes the entire display which results in the rectangle being in the center of it as expected.

minggo:
Thanks for the description, i will have a try.

minggo:
Yep, it works great with your sample codes. But the effect is wrong if i create a window like
```c++
glview = GLViewImpl::createWithFullScreen("Cocos2D-X", true);
```
Does it mean the PR should work together with retina enabled? On windows and linux, there is not `GLViewImpl::getRetinaFactor()` and `GLViewImpl::enableRetina()`.

MileyHollenberg:
Is the effect wrong on just Mac or also Windows and Linux? It should work as expected on Windows and Linux since it just takes the entire screen size and uses that to create it's context. On Mac it is required to enable retina if available to make sure the OpenGL view fits properly. (also how do you handle High DPI screens on Windows and Linux? I have a 4K screen set at 150% scaling which would be the same as 1.5 Retina scale factor imo)

It could become part of the createWithFullscreen method with a ifdef for MacOSX. I unfortunately won't be able to do that anymore since starting tomorrow I'm gone for at least a month probably longer.

Bilalmirza:
@minggo  are there any other workaround for this? I use createWithFullscreen and results are really wrong in my case.

Bilalmirza:
I've come back to the same bug after a long time... Your solution doesn't work for me. I'm on a macbook pro 2018

what works for me is just simply adding

glview = GLViewImpl::createWithFullScreen("Explottens");
#if( CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
                glview->setFrameSize(1440, 900); //base size
#endif

Bilalmirza:
okay so the previous fix had a few problems ... not all scaled sizes were producing retina output. and mouse can be taken out of screen. :/
proper fix is 
```
#if( CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
     glview->setFrameSize(1440, 900);
                ((GLViewImpl *)glview)->enableRetina(true);
                ((GLViewImpl *)glview)->setFrameZoomFactor(2.0f);
#endif
```

this will work for all macbook pro resolutions and will produce retina output. 

disadvantage is that screen adjusts resolution 2-3 times before the app is starting. so in order to fix that i went in the glviewImp::initwithRect() and where you set the framesize
```
    setFrameSize(rect.size.width, rect.size.height);
```
changed it to

```
#if CC_TARGET_PLATFORM == CC_PLATFORM_MAC
    _isRetinaEnabled = true;
    _frameZoomFactor = 2.0f;
    setFrameSize(1440,900);

#else
    setFrameSize(rect.size.width, rect.size.height);
#endif

```
This way updateFrameSize() is only called once. 
I'm not sure if its the best solution but now there are no black screen and all resolutions are supported.

Tested on macbook pro 2018 && macbook pro mid 2015. and on all scaled displays.

your thoughts? @MileyHollenberg  @minggo

Bilalmirza:
Sorry, above solution does not work for all resolution. Is there no fix to this?

Bilalmirza:
Sorry my fix doesn't work for macbooks before 2015 :(
This bug is blocker for me for the release. @minggo  can you please have a look

minggo:
@Bilalmirza you released game for mac? Could you please provide a test case or code snippet?

Bilalmirza:
@minggo  thanks for the quick response. Yes I am releasing for mac, apple rejected due to resolution issue

Steps
- new project
- go in appdelegate
```
//        glview = GLViewImpl::createWithRect("macTest", cocos2d::Rect(0, 0, designResolutionSize.width, designResolutionSize.height));
        glview = GLViewImpl::createWithFullScreen("macTest");
```
- Go to system preferences -> displays -> display tab -> pick scaled -> choose different scales



results for different scaling 
screenshots attached
https://www.dropbox.com/sh/ho87mumbywuuii5/AAB-x4XGc8tVGA-KbVHYXO-ia?dl=0

minggo:
@Mee-gu please take a look.

Mee-gu:
> Sorry my fix doesn't work for macbooks before 2015 :(
> This bug is blocker for me for the release. @minggo can you please have a look

@Bilalmirza Your fix can work well for all resolutions on my MacBook Pro(Early 2013).

Bilalmirza:
@Mee-gu but my method has a few problems if you switch app (cmd+tab). The resolution is changed for all mac.
second there is a black screen delay due to resizing.
Third this fix doesn't cover all devices 2011 to 2015 air  and 2011 to 2013 macbooks... basically the ones that don't have the resolution 1440x900

Mee-gu:
@Bilalmirza 
1. I can't switch app by cmd+tab when the game is running in full screen.
2. Does the black screen you means is the screen blinking when launching a game?

Bilalmirza:
@Mee-gu 
1. So for mac you need to be able to switch apps which in cocos2d-x is not possible yet.
and for mac app if you are in app-debug mode and in fullscreen and if there is a crash you are unable to switch to xcode to debug or even stop the app from running. you have to do cmd+alt+esc and force quit xcode which takes time. 

plus you want your app to be able to hide etc.
so you have to add
in Application::run() /// or any other place you can get the context
NSWindow *nsWindow = glview->getCocoaWindow();
nsWindow.canHide = true;
nsWindow.hidesOnDeactivate = true;


2. Yes i mean the screen blinking. But I know it has to blink due to resolution change.
But we don't simply want the resolution change. 
lets suppose if the user has 1440x900 in displays settings the screen wont blink. 
so is there a way not to get the blinking by somehow setting the gl size. I don't think setting a hardcoded value like setFrameSize(1440,900); should be the way to go.

Mee-gu:
@Bilalmirza Great, I can switch the app now. And indeed the resolution is changed. I will take into account for those 3 issues.

Mee-gu:
@Bilalmirza This [PR](https://github.com/cocos2d/cocos2d-x/pull/20043) fixed the issues. It needs to update the contents of the window by [`glfwWaitEvents`](http://fabien.benetou.fr/pub/Prototypes/vAtelier/HeadlessGL/installprocess/glfw-3.1.2/docs/html/input.html#events) when the window or framebuffer is resized(I think).

Bilalmirza:
@Mee-gu This didn't change anything for me. Do i have to add my solution as well to this or just the glfwWaitEvents?

Bilalmirza:
@Mee-gu but this does seem to work in a new project(i was testing in my game for the last comment) for all display sizes except for 1024x 640... let me do some more testing and see.

Mee-gu:
@Bilalmirza Is there any problems with 1024x640 since it can work on my mac?

Bilalmirza:
@Mee-gu thats probably because the background is black in helloworld. 
do this in helloworld
  
 sprite->setScale(20, 20);
the side borders are cut from the screen.
https://www.dropbox.com/s/qfb72ny5k0pgx4d/Screenshot%202019-08-16%20at%204.04.28%20PM.png?dl=0

Bilalmirza:
I suppose we can let this resolution slip. The more important ones work fine for me right now... I'm gonna test them on older devices soon. I'll let you know the results. 

Thanks for the quick fix man, You are the best :D

PR #17341, opened 2017-02-10
Fix tizen compilation issue

zilongshanren:
1. change appid from `org.tizen` to `org.cocos2d-x` in all the tests file and project template file.
2. bump the 3rd party deps to 122 with websockets and openssl update
3. fix compilation issues
4. enable websockets tests on Tizen


dumganhar:
[ci rebuild]

PR #17342, opened 2017-02-10
Some fixes for websocket

dumganhar:
* libwebsockets & curl are dynamic libraries now, modify project configuration for win32.
* Corrects some logic in WebSocket.cpp, each ws connection will create a virutal host now.
* Adds wss test case. WebSocketTest suggests developers to use `WebSocket::closeAsync` method which will not block current thread.
* Allows self-assign cert.


dumganhar:
@minggo, please review.

Issue #17357, opened 2017-02-13
Better encapsulate the definition of mouse buttons


radman0x:
- cocos2d-x version: 3.13.1
- devices test on: Linux
- developing environments
   - NDK version: NA
   - Xcode version: NA
   - VS version: NA
   - browser type and version: NA

This issue is to propose a design adjustment to ``EventMouse`` so that it exposes the definition of a mouse button as an ``enum class`` instead of raw ``#define``'s (for the obvious beneficial reasons).  The result of the change is that clients receiving mouse click events will need to handle a ``EventMouse::MouseButton`` type rather than a raw int. The benefit of this is that the type of the mouse button will be immediately obvious, the downside being that existing code will need to be updated to handle the type (this is trivial though e.g. static_cast<int>).

This new issue was triggered by discussion on #8414.

As an aside; Cocos2dx uses GLFW3 for its input handling and the current defined button values are a copy paste from there to the existing ``#define``'s. From a clients point of view however they shouldn't be exposed to the fact that GLFW is being used, so that if a new input system is used at some point in the future then it can be swapped out without causing unnecessary pain.

radman0x:
I've put up a PR for my suggested changes #17358.

PR #17358, opened 2017-02-13
17357: Better encapsulate the definition of mouse buttons

radman0x:
Move definition of buttons from a list of #define's to an enum class and made appropriate updates for successful compilation.

Built on linux only at this stage, work spawned from #17357, see there for more discussion for the reasons for the change.

### DIFF
--- a/cocos/base/CCEventMouse.h
+++ b/cocos/base/CCEventMouse.h
@@ -26,8 +26,6 @@
 #ifndef __cocos2d_libs__CCMouseEvent__
 #define __cocos2d_libs__CCMouseEvent__
 
-#include <ostream>
-
 #include "base/CCEvent.h"
 #include "math/CCGeometry.h"
 
@@ -195,8 +193,6 @@ private:
 
     friend class EventListenerMouse;
 };
-
-std::ostream& operator<<(std::ostream& out, const EventMouse::MouseButton& button);
 
 NS_CC_END
 
--- a/cocos/base/CCEventMouse.h
+++ b/cocos/base/CCEventMouse.h
@@ -26,17 +26,10 @@
 #ifndef __cocos2d_libs__CCMouseEvent__
 #define __cocos2d_libs__CCMouseEvent__
 
+#include <ostream>
+
 #include "base/CCEvent.h"
 #include "math/CCGeometry.h"
-
-#define MOUSE_BUTTON_LEFT       0
-#define MOUSE_BUTTON_RIGHT      1
-#define MOUSE_BUTTON_MIDDLE     2
-#define MOUSE_BUTTON_4          3
-#define MOUSE_BUTTON_5          4
-#define MOUSE_BUTTON_6          5
-#define MOUSE_BUTTON_7          6
-#define MOUSE_BUTTON_8          7
 
 /**
  * @addtogroup base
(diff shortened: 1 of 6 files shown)


cocos/base/CCEventMouse.h
```diff
@@ -64,6 +57,20 @@ class CC_DLL EventMouse : public Event
...
+      BUTTON_RIGHT   =  1,
+      BUTTON_MIDDLE  =  2,
+      BUTTON_4       =  3,
+      BUTTON_5       =  4,
+      BUTTON_6       =  5,
+      BUTTON_7       =  6,
+      BUTTON_8       =  7
+    };
+
+
```
minggo, line 71: it will break compatibility thought i like enum class

cocos/base/CCEventMouse.h
```diff
@@ -189,6 +196,8 @@ class CC_DLL EventMouse : public Event
     friend class EventListenerMouse;
 };
 
+std::ostream& operator<<(std::ostream& out, const EventMouse::MouseButton& button);
```
minggo: what does it used for?

stevetranby: I think it's just for easy "printing" to string (stringstream), cout (iostream), or when writing to file (fstream).

radman0x: Yep, easy printing is what it's there for. It means that the enum value can be written to ``std::ostream`` objects.

e.g.
```cpp
std::cout << mouseButton;
std::ostringstream << mouseButton;
```
``std::ostream`` is the standard mechanism for input/output in C++.
It's a convenience so it doesn't have to be included in the changes.

cocos/scripting/lua-bindings/auto/lua_cocos2dx_auto.cpp
```diff
@@ -36582,7 +36582,7 @@ int lua_cocos2dx_EventMouse_getMouseButton(lua_State* tolua_S)
             tolua_error(tolua_S,"invalid arguments in function 'lua_cocos2dx_EventMouse_getMouseButton'", nullptr);
             return 0;
         }
-        int ret = cobj->getMouseButton();
```
minggo: please don't modify auto generated files, the file will be auto generated in CI machine

cocos/base/CCEventMouse.h
stevetranby: I'm sure there's no good way to separate it out, but yeah it might be best left for game developers to add in their own code base instead of in the engine source.

Or maybe there's a good way to put all ostream << operator overloads into a separate CCStreamOperators.h might be better (though I can imagine that would require a lot of includes as part of the header one for each type that gets an overload ...??

radman0x: I think for the moment I'll take it out and leave it to be implemented by anyone that wants it. The implementation is pretty trivial.

cocos/scripting/lua-bindings/auto/lua_cocos2dx_auto.cpp
radman0x: My bad, didn't realise this was a generated file. Is there a template somewhere else that I need to update though?

radman0x:
Looks like  a build failed in Jenkins:
```
ccs-pr » win32_bak completed with result FAILURE
```
I can't see how to get to the actual output for that step though to see what went wrong. Any pointers?

minggo:
```
/home/travis/build/cocos2d/cocos2d-x/tests/cpp-tests/Classes/InputTest/MouseTest.cpp:5:80: error: cannot bind ‘std::basic_ostream<char>’ lvalue to ‘std::basic_ostream<char>&&’
```
I think you should modify cpp-tests too.

zilongshanren:
[ci rebuild]

radman0x:
Should we now delete the branch for this?

minggo:
Yep, you can delete the branch. Thanks.

Issue #17346, opened 2017-02-11
Why does 'EditBoxImplCommon' have the same member variable with 'EditBox'?


summerinsects:
Why does 'EditBoxImplCommon' have the same member variable with 'EditBox'?
These are

    EditBox::InputMode    _editBoxInputMode;
    EditBox::InputFlag    _editBoxInputFlag;
    EditBox::KeyboardReturnType  _keyboardReturnType;
    cocos2d::TextHAlignment _alignment;
    std::string _text;
    std::string _placeHolder;   
    Color4B _colText;
    Color4B _colPlaceHolder;
    int   _maxLength;

Why not delete them?

zilongshanren:
@summerinsects 
I think it's a design flaw, the member variables could be merged into EditBoxCommon.
If you could do this and send a PR, I'm glad to help and merge.
thanks.

summerinsects:
@zilongshanren  I will have a try

PR #17361, opened 2017-02-13
Refine editbox for issue 17346 --merge some member variables into EditBoxCommon

summerinsects:
https://github.com/cocos2d/cocos2d-x/issues/17346

### DIFF
--- a/cocos/ui/UIEditBox/UIEditBoxImpl.h
+++ b/cocos/ui/UIEditBox/UIEditBoxImpl.h
@@ -54,13 +54,28 @@ namespace cocos2d {
             virtual void setInputFlag(EditBox::InputFlag inputFlag) = 0;
             virtual void setMaxLength(int maxLength) = 0;
             virtual int  getMaxLength() = 0;
-            virtual void setTextHorizontalAlignment(cocos2d::TextHAlignment alignment) = 0;
+            virtual void setTextHorizontalAlignment(TextHAlignment alignment) = 0;
             virtual void setReturnType(EditBox::KeyboardReturnType returnType) = 0;
             virtual bool isEditing() = 0;
             
             virtual void setText(const char* pText) = 0;
             virtual const char* getText(void) = 0;
             virtual void setPlaceHolder(const char* pText) = 0;
+            virtual const char* getPlaceHolder(void) { return ""; }  // TODO: =0;
+
+            virtual const char* getFontName() { return ""; }
+            virtual int getFontSize() { return -1; }
+            virtual const Color4B& getFontColor() { return Color4B::WHITE; }
+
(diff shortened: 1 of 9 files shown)


cocos/ui/UIEditBox/UIEditBoxImpl.h
```diff
@@ -54,13 +54,28 @@ namespace cocos2d {
...
+            virtual void setTextHorizontalAlignment(TextHAlignment alignment) = 0;
             virtual void setReturnType(EditBox::KeyboardReturnType returnType) = 0;
             virtual bool isEditing() = 0;
             
             virtual void setText(const char* pText) = 0;
             virtual const char* getText(void) = 0;
             virtual void setPlaceHolder(const char* pText) = 0;
+            virtual const char* getPlaceHolder(void) = 0;
+
+            virtual const char* getFontName() { return ""; }
```
zilongshanren: I think all these method should be virtual method and needed be implemented in the platform-impl class.  Eg, in Tizen platform, the `getInputMode` method might be return the wrong input mode.  Because in UIEditboxImpl-tizen.h, the `getInputMode1 method is not properly implemented.

zilongshanren:
thanks, this PR looks good to me. But it need to be polished for the cross platform compatibility.

cocos/ui/UIEditBox/UIEditBoxImpl.h
zilongshanren: um... Tizen impl doesn't inherit from the UIEditBox-common, I think it should be refactored, this is another story.

zilongshanren:
👍  I'm OK with this PR now.  @minggo could you please fetch this PR and test it in real device.
Thanks.

zilongshanren:
@summerinsects 
I know why there are some duplicated data in EditBox and EditBoxCommon now.
Because some platform implementation inherit from EditBoXImpl and others inherit from EditBoxImplCommon.

If you merge the data, it means all the data are kept in EditBoxImplCommon, so there will break some implementation which inherit from EditBoxImpl.

cocos/ui/UIEditBox/UIEditBoxImpl.h
summerinsects: This is a lazy solution. :-)
Pure virtual method is better, but it needs to modify many files

minggo:
> @minggo could you please fetch this PR and test it in real device.
Thanks.

@zilongshanren sure. What test cases should i test?

zilongshanren:
@minggo

In the `Node::UI -> EditBox test`

minggo:
@zilongshanren the test case works ok, should i merge it?

PR #17381, opened 2017-02-17
[WebSocket] Adds support for getting url and server selected protocol

dumganhar:


cocos/network/WebSocket.h
```diff
@@ -212,6 +212,9 @@ class CC_DLL WebSocket
      */
     State getReadyState();
 
+    inline const std::string& getUrl() { return _url; }
+    inline const std::string& getProtocol() { return _selectedProtocol; }
```
minggo: please add const like
```c++
inline const std::string& getProtocol() const
```
i am not sure if inline and const can work together or not

cocos/scripting/lua-bindings/manual/network/Lua_web_socket.cpp
```diff
@@ -313,7 +349,7 @@ static int tolua_Cocos2d_WebSocket_close00(lua_State* tolua_S)
     {
         LuaWebSocket* self    = (LuaWebSocket*)  tolua_tousertype(tolua_S,1,0);
         if (NULL != self ) {
-            self->close();
+            self->closeAsync();
```
dumganhar, line 352: We suggest developers to use `closeAsync` which will not block current thread.

cocos/scripting/js-bindings/manual/network/jsb_websocket.cpp
```diff
@@ -361,11 +367,14 @@ bool js_cocos2dx_extension_WebSocket_constructor(JSContext *cx, uint32_t argc, j
...
+        JS_DefineProperty(cx, obj, "url", args.get(0), JSPROP_ENUMERATE | JSPROP_PERMANENT | JSPROP_READONLY);
+
+        // The websocket draft uses lowercase 'url', so 'URL' need to be deprecated.
         JS_DefineProperty(cx, obj, "URL", args.get(0), JSPROP_ENUMERATE | JSPROP_PERMANENT | JSPROP_READONLY);
 
-        //protocol not support yet (always return "")
+        // Initialize protocol property with an empty string, it will be assigned in onOpen delegate.
         JS::RootedValue jsprotocol(cx, c_string_to_jsval(cx, ""));
-        JS_DefineProperty(cx, obj, "protocol", jsprotocol, JSPROP_ENUMERATE | JSPROP_PERMANENT | JSPROP_READONLY);
+        JS_DefineProperty(cx, obj, "protocol", jsprotocol, JSPROP_ENUMERATE | JSPROP_PERMANENT);
```
dumganhar, line 377: `protocol` property needs to be updated in `onOpen` delegate method, therefore it should not be readonly here.

cocos/network/WebSocket.h
dumganhar: DONE

dumganhar:
[ci rebuild]

PR #17384, opened 2017-02-20
Update CCNode.cpp

dadidzf:
Retain the node before update for lua, in case the lua update remove self and cause crashes !

Example(lua):
node = cc.Node:create()
node:onUpdate(handler(self, self.nodeUpdate))
node.nodeUpdate = function(self, dt)
    self:removeFromParent()             -- Cause crashes, _componentContainer has been a wild pointer !
end

Crash Point:
void Node::update(float fDelta)
{
#if CC_ENABLE_SCRIPT_BINDING
if (0 != _updateScriptHandler)
    {
    //only lua use
    SchedulerScriptData data(_updateScriptHandler,fDelta);
    ScriptEvent event(kScheduleEvent,&data);
    ScriptEngineManager::getInstance()->getScriptEngine()->sendEvent(&event);
}
#endif
    
    if (_componentContainer && !_componentContainer->isEmpty())  -- Crash
    {
        _componentContainer->visit(fDelta);
    }
}

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1645,7 +1645,7 @@ void Node::update(float fDelta)
     {
         _componentContainer->visit(fDelta);
     }
-    this->release();
+	this->release();
 }
 
 // MARK: coordinates
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1630,6 +1630,7 @@ void Node::pauseSchedulerAndActions()
 // override me
 void Node::update(float fDelta)
 {
+    this->retain();
 #if CC_ENABLE_SCRIPT_BINDING
     if (0 != _updateScriptHandler)
     {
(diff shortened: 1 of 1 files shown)


Dauch:
you need to add to a parent to make it retain, not modify the CCNode.cpp do do the change.

cocos/2d/CCNode.cpp
```diff
@@ -1630,6 +1630,7 @@ void Node::pauseSchedulerAndActions()
 // override me
 void Node::update(float fDelta)
 {
+    this->retain();
```
dumganhar: Is it better to put retain/release things into `CC_ENABLE_SCRIPT_BINDING` area?

dadidzf: Yes, It's better ! ^_^

dadidzf:
I did not find a good way to add the retain/release pairs, so close it at present

Issue #17397, opened 2017-02-22
Win32 AudioEngine linker errors when build by CMake


MrCapone:
- cocos2d-x version: 3.14.1
- devices test on: Windows 10
- developing environments
   - VS version: 2015

Steps to Reproduce:

1. create blank cocos project and generate VS solution by cmake
2. try load mp3 or ogg by` AudioEngine::play2d("");`

error:

```
1>cocos2dInternal.lib(AudioEngine-win32.obj) : error LNK2001: unresolved external symbol ""public: static bool __cdecl cocos2d::experimental::AudioDecoderManager::init(void)" (?init@AudioDecoderManager@experimental@cocos2d@@SA_NXZ)"
1>cocos2dInternal.lib(AudioEngine-win32.obj) : error LNK2001: unresolved external symbol ""public: static void __cdecl cocos2d::experimental::AudioDecoderManager::destroy(void)" (?destroy@AudioDecoderManager@experimental@cocos2d@@SAXXZ)"
1>cocos2dInternal.lib(AudioCache.obj) : error LNK2001: unresolved external symbol ""public: static class cocos2d::experimental::AudioDecoder * __cdecl cocos2d::experimental::AudioDecoderManager::createDecoder(char const *)" (?createDecoder@AudioDecoderManager@experimental@cocos2d@@SAPAVAudioDecoder@23@PBD@Z)"
1>cocos2dInternal.lib(AudioPlayer.obj) : error LNK2001: unresolved external symbol ""public: static class cocos2d::experimental::AudioDecoder * __cdecl cocos2d::experimental::AudioDecoderManager::createDecoder(char const *)" (?createDecoder@AudioDecoderManager@experimental@cocos2d@@SAPAVAudioDecoder@23@PBD@Z)"
1>cocos2dInternal.lib(AudioCache.obj) : error LNK2001: unresolved external symbol ""public: static void __cdecl cocos2d::experimental::AudioDecoderManager::destroyDecoder(class cocos2d::experimental::AudioDecoder *)" (?destroyDecoder@AudioDecoderManager@experimental@cocos2d@@SAXPAVAudioDecoder@23@@Z)"
[...]

PR #17398, opened 2017-02-22
fix issue #17397

MrCapone:
cocos/audio/CMakeLists.txt add missing .cpp for win32 platform


dumganhar:
Oh, I didn't notice that you have sent this pull request.  :)

PR #17431, opened 2017-03-02
Prevent unexpected calls to unscheduled selector in long updates.

j-jorge:
If the delta time dt passed to cocos2d::Timer::update(float) is longer than the
interval of the timer, then the timer will be triggered several times even if
the selector is unscheduled during the call.

This commit adds a flag in cocos2d::Timer stop the update loop if the selector
has been unscheduled.


minggo:
The codes looks good to me. I will test it.

minggo:
thanks, it works good

PR #17434, opened 2017-03-02
Win10 UWP fixes for new Mouse events

stammen:
This PR update the Win10 UWP mouse handling code to handle the recent change to using EventMouse::MouseButton and EventMouse::MouseEventType.


dumganhar:
Thanks.

PR #17435, opened 2017-03-03
Some fixes for search paths:

dumganhar:
1. Adds ‘_originalSearchPaths’ variable, ’getOriginalSearchPaths’ returns the original values set by ‘setSearchPaths’  or ‘addSearchPath’.
2. Adds a getter function ‘getDefaultResourceRootPath’.
3. ‘setDefaultResourceRootPath’ should also update search paths and remove file path cache internally.
4. ‘setSearchPaths’  supports to pass self (_originalSearchPath), could be used in ‘setDefaultResourceRootPath’ to update the final ’_searchPathArray’ for searching full path.
5. ‘addSearchPath’ fix, the default resource root path should be the last element in ‘_searchPathArray’.
6. 'setSearchResolutionsOrder' supports to pass self '_searchResolutionsOrderArray' as parameter.

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -860,9 +860,11 @@ std::string FileUtils::fullPathFromRelativeFile(const std::string &filename, con
 void FileUtils::setSearchResolutionsOrder(const std::vector<std::string>& searchResolutionsOrder)
 {
     bool existDefault = false;
+    std::vector<std::string> searchResolutionOrderCopied = searchResolutionsOrder;
+
     _fullPathCache.clear();
     _searchResolutionsOrderArray.clear();
-    for(const auto& iter : searchResolutionsOrder)
+    for(const auto& iter : searchResolutionOrderCopied)
     {
         std::string resolutionDirectory = iter;
         if (!existDefault && resolutionDirectory == "")
@@ -904,7 +906,7 @@ const std::vector<std::string>& FileUtils::getSearchResolutionsOrder() const
 
 const std::vector<std::string>& FileUtils::getSearchPaths() const
 {
-    return _searchPathArray;
+    return _originalSearchPaths;
 }
 
 void FileUtils::setWritablePath(const std::string& writablePath)
@@ -912,36 +914,54 @@ void FileUtils::setWritablePath(const std::string& writablePath)
     _writablePath = writablePath;
 }
 
+const std::string& FileUtils::getDefaultResourceRootPath() const
+{
+    return _defaultResRootPath;
+}
+
 void FileUtils::setDefaultResourceRootPath(const std::string& path)
 {
-    _defaultResRootPath = path;
+    if (_defaultResRootPath != path)
+    {
+        _fullPathCache.clear();
+        _defaultResRootPath = path;
+        if (!_defaultResRootPath.empty() && _defaultResRootPath[_defaultResRootPath.length()-1] != '/')
+        {
+            _defaultResRootPath += '/';
+        }
+
+        // Updates search paths
+        setSearchPaths(_originalSearchPaths);
+    }
 }
 
 void FileUtils::setSearchPaths(const std::vector<std::string>& searchPaths)
 {
     bool existDefaultRootPath = false;
+    _originalSearchPaths = searchPaths;
 
     _fullPathCache.clear();
     _searchPathArray.clear();
-    for (const auto& iter : searchPaths)
+
+    for (const auto& path : _originalSearchPaths)
     {
         std::string prefix;
-        std::string path;
+        std::string fullPath;
 
-        if (!isAbsolutePath(iter))
+        if (!isAbsolutePath(path))
         { // Not an absolute path
             prefix = _defaultResRootPath;
         }
-        path = prefix + (iter);
+        fullPath = prefix + path;
         if (!path.empty() && path[path.length()-1] != '/')
         {
-            path += "/";
+            fullPath += "/";
         }
         if (!existDefaultRootPath && path == _defaultResRootPath)
         {
             existDefaultRootPath = true;
         }
-        _searchPathArray.push_back(path);
+        _searchPathArray.push_back(fullPath);
     }
 
     if (!existDefaultRootPath)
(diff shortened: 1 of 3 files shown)


cocos/platform/CCFileUtils.cpp
```diff
@@ -860,9 +860,11 @@ std::string FileUtils::fullPathFromRelativeFile(const std::string &filename, con
 void FileUtils::setSearchResolutionsOrder(const std::vector<std::string>& searchResolutionsOrder)
 {
     bool existDefault = false;
+    std::vector<std::string> searchResolutionOrderCopied = searchResolutionsOrder;
+
     _fullPathCache.clear();
     _searchResolutionsOrderArray.clear();
-    for(const auto& iter : searchResolutionsOrder)
+    for(const auto& iter : searchResolutionOrderCopied)
```
minggo: why need to copy `searchResolutionsOrder `?

cocos/platform/CCFileUtils.h
```diff
@@ -874,6 +879,11 @@ class CC_DLL FileUtils
     std::vector<std::string> _searchPathArray;
 
     /**
+     * The search paths which was set by 'setSearchPaths' / 'addSearchPath'.
+     */
+    std::vector<std::string> _originalSearchPaths;
```
minggo, line 893: why need _originalSearchPaths

cocos/platform/CCFileUtils.cpp
dumganhar: Suppose this invocation: 
```c++
fileUtils->setSearchResolutionsOrder(_searchResolutionsOrderArray);
```
_searchResolutionsOrderArray will be cleared before the for loop.
The same situation could happen in setSearchPaths.

cocos/platform/CCFileUtils.h
dumganhar: FileUtils uses `_searchPathArray` for searching full path internally.
`setSearchPaths` will modify `_searchPathArray` variable, it will be connected by `_defaultResourceRootPath`. 
`_originalSearchPaths` saves the original parameter `setSearchPaths` passes in, and `getSearchPaths` returns `_originalSearchPath` will be more reasonable than returning `_searchPathArray`. It's the rule of getter returns what user has set.

If `getSearchPaths` returns `_searchPathArray`, you could suppose the following code will always not work.

If current search path is `bbb/`, `ccc/`
```c++
auto oldDefaultResourceRoot = fileUtils->getDefaultResourceRootPath(); // Suppose it return "" which is an empty string.
auto oldSearchPaths = fileUtils->getSearchPaths(); // oldSearchPaths is : ["bbb/", "ccc/"]
fileUtils->setDefaultResouceRootPath("aaa"); // oldSearchPath is : ["aaa/bbb/", "aaa/ccc"]
// oldSearchPaths will be modified by adding a prefix `aaa` since `_searchPathArray` was modified. 

... ...

// reset resource root
fileUtils->setDefaultResourceRootPath(oldDefaultResourceRoot); 
auto newSearchPath = fileUtils->getSearchPaths(); // newSearchPath is incorrect: ["aaa/bbb", "aaa/ccc"] since it returns `_searchPathArray` which was added by prefix 'aaa'
```

cocos/platform/CCFileUtils.cpp
dumganhar:
```c++
fileUtils->setSearchResolutionsOrder(fileUtils->getSearchResolutionsOrder());
fileUtils->setSearchPaths(fileUtils->getSearchPaths());
```
will aways fail.

dumganhar:
@kpkhxlgy0, any thoughts about this pull request since is inspired by your PR (https://github.com/cocos2d/cocos2d-x/pull/11195).

cocos/platform/CCFileUtils.cpp
minggo: Yep, you are right. But i think it is better to check it at first like
```c++
if (_searchResolutionsOrderArray == searchResolutionsOrder)
    return;
```

dumganhar: For setSearchResolutionsOrders, we could.
But for setSearchPaths, we couldn't.
Because setSearchPaths will update `_searchPathArray` which will be affected by `_defaultResourceRootPath`.

minggo: Yep, i just mean searchResolutionsOrder.

dumganhar: @minggo , updated `setSearchResolutionsOrder`.

sbrednikhin:
Adding FileUtils::getOriginalSearchPaths is good thing, but what about people who rely on old resource path order?
Making original search paths last in the list looks strange (and I think might have some additional performance cost).

balmerdx:
This commit is not good. 
1. Break backward compatibility.
2. Strange logic. Why one path go to back?????? Confuse interface.
3. Not add new functionality. 

In older code is several variants to implement dumganhar logic without change FileUtils class. 
- addSearchPath(..., front=true)
- get/setSearchPaths and change path order.

cocos/platform/CCFileUtils.cpp
```diff
@@ -962,10 +991,21 @@ void FileUtils::addSearchPath(const std::string &searchpath,const bool front)
...
     if (front) {
+        _originalSearchPaths.insert(_originalSearchPaths.begin(), searchpath);
         _searchPathArray.insert(_searchPathArray.begin(), path);
     } else {
-        _searchPathArray.push_back(path);
+        _originalSearchPaths.push_back(searchpath);
+
+        if (!_searchPathArray.empty() && _searchPathArray[_searchPathArray.size()-1] == _defaultResRootPath)
+        {
+            _searchPathArray.insert(_searchPathArray.begin() + _searchPathArray.size() -1, path);
```
dumganhar, line 1003: @sbrednikhin, @balmerdx
So the key point is that this line push the `default resource path` to the end, right?
Yes, I know I broke the compatibility here since I think the search path of default resource should always be at the end. I don't know what kinds of situation you will do in your game. Perhaps, more code snippets and more detail are welcome.

Anyway, if you think I was wrong, feel free to submit a `new` or  `better` pull request to correct my mistake. Thanks.

sbrednikhin: I don't care where default resource path is, actually. My only point is that if you breaking regular code behavior, causing people to fix their code (which must rely on some order of paths), you should have some good reason for that. Not just thinking that it is better. 
Since  at least some people don't see any obvious reason for that change, you could point on it in comment or something.
This project is open source common code - not your own branch.

dumganhar: OK，you're totally right, thanks for reminding me. I will revert that line. I was wrong, I used it as 'my own my private branch', good luck guys, thank you guys.

sbrednikhin: Thank you, for understanding. We are very appreciate your work.

PR #17436, opened 2017-03-03
[Label] use freetype api to support bold

treert:
label enable blod use shadow to achive, so bold and outline can not use in same time.

use freetype api to support bold

### DIFF
--- a/cocos/2d/CCFontFreeType.cpp
+++ b/cocos/2d/CCFontFreeType.cpp
@@ -26,6 +26,7 @@ THE SOFTWARE.
 #include "2d/CCFontFreeType.h"
 #include FT_BBOX_H
 #include FT_BITMAP_H
+#include FT_OUTLINE_H
 #include "edtaa3func.h"
 #include "2d/CCFontAtlas.h"
 #include "base/CCDirector.h"
--- a/cocos/2d/CCFontFreeType.cpp
+++ b/cocos/2d/CCFontFreeType.cpp
@@ -25,6 +25,7 @@ THE SOFTWARE.
 
 #include "2d/CCFontFreeType.h"
 #include FT_BBOX_H
+#include FT_BITMAP_H
 #include "edtaa3func.h"
 #include "2d/CCFontAtlas.h"
 #include "base/CCDirector.h"
@@ -49,9 +50,15 @@ typedef struct _DataRef
 
 static std::unordered_map<std::string, DataRef> s_cacheFontData;
 
-FontFreeType * FontFreeType::create(const std::string &fontName, float fontSize, GlyphCollection glyphs, const char *customGlyphs,bool distanceFieldEnabled /* = false */,float outline /* = 0 */)
+FontFreeType * FontFreeType::create(const std::string &fontName, float fontSize, GlyphCollection glyphs, const char *customGlyphs,
+    bool distanceFieldEnabled /* = false */,float outline /* = 0 */,bool bold /* = false */)
 {
     FontFreeType *tempFont =  new (std::nothrow) FontFreeType(distanceFieldEnabled,outline);
+
+    if (bold)
+    {
+        tempFont->_boldSize = CC_CONTENT_SCALE_FACTOR() * 64;
+    }
 
     if (!tempFont)
         return nullptr;
@@ -105,6 +112,7 @@ FontFreeType::FontFreeType(bool distanceFieldEnabled /* = false */, float outlin
 , _fontAtlas(nullptr)
 , _encoding(FT_ENCODING_UNICODE)
 , _usedGlyphs(GlyphCollection::ASCII)
+, _boldSize(0)
 {
     if (outline > 0.0f)
     {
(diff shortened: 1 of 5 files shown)


cocos/2d/CCLabel.h
```diff
@@ -364,6 +364,11 @@ class CC_DLL Label : public Node, public LabelProtocol, public BlendProtocol
     void enableBold();
 
     /**
+     * triggle bold, use freetype to support bold only for ttf
+     */
+    void setBold(bool bold = true);
```
minggo: please don't use default parameter

minggo:

cocos/2d/CCLabel.h
treert: ok, i will modify it

cocos/2d/CCLabel.h
```diff
@@ -364,6 +364,11 @@ class CC_DLL Label : public Node, public LabelProtocol, public BlendProtocol
     void enableBold();
 
     /**
+     * set bold effect, use freetype to support bold only for ttf
+     */
+    void setBold(bool bold);
```
zilongshanren: I suggest not providing another `setBold` API, we should replace the old Bold implementation for TTF.

zilongshanren (changes requested):
We should replace the old bold implementation with the new one

cocos/2d/CCLabel.h
treert: sorry for reply so later

Dimon4eg:
[ci rebuild]

minggo:
The error description is:

```
/home/travis/build/cocos2d/cocos2d-x/cocos/2d/CCFontFreeType.cpp: In member function ‘unsigned char* cocos2d::FontFreeType::getGlyphBitmapWithOutline(uint64_t, FT_BBox&)’:

/home/travis/build/cocos2d/cocos2d-x/cocos/2d/CCFontFreeType.cpp:447:68: error: ‘FT_Outline_EmboldenXY’ was not declared in this scope

                         FT_Outline_EmboldenXY(outline, _boldSize, 0);

                                                                    ^

make[2]: *** [cocos/CMakeFiles/cocos2dInternal.dir/2d/CCFontFreeType.cpp.o] Error 1

make[1]: *** [cocos/CMakeFiles/cocos2dInternal.dir/all] Error 2
```

cocos/2d/CCLabel.cpp
```diff
@@ -997,8 +997,6 @@ bool Label::setTTFConfigInternal(const TTFConfig& ttfConfig)
 
     if (_fontConfig.italics)
         this->enableItalics();
-    if (_fontConfig.bold)
-        this->enableBold();
```
drelaptop, line 1001: This PR is another better implementation of `support bold for TTF font`, but two lines above is the logic of enable bold for label. so I think they shouldn't been removed, for those are reasonable, intuitive and easy to understand.

cocos/2d/CCLabel.cpp
```diff
@@ -1218,7 +1225,15 @@ void Label::disableEffect(LabelEffect effect)
...
+            if (_currentLabelType == LabelType::TTF) // only for ttf
+            {
+                if (_fontConfig.bold == true)
+                {
+                    _fontConfig.bold = false;
+                    setTTFConfig(_fontConfig);
+                }
+            }
+            else if (_boldEnabled) {
                 _boldEnabled = false;
```
drelaptop, line 1237: code in `Label::enableBold()` changed the meaning of __`_boldEnabled`__ ,  because `_boldEnabled = false` when a TTF label enable bold, that is strange. 

We should keep `_boldEnabled` true as long as a label enables bold.  In order to achieve this goal, we can put `if (_currentLabelType == LabelType::TTF)` in `if (!_boldEnabled)`.

cocos/2d/CCFontFreeType.cpp
```diff
@@ -49,10 +51,16 @@ typedef struct _DataRef
...
 
-FontFreeType * FontFreeType::create(const std::string &fontName, float fontSize, GlyphCollection glyphs, const char *customGlyphs,bool distanceFieldEnabled /* = false */,float outline /* = 0 */)
+FontFreeType * FontFreeType::create(const std::string &fontName, float fontSize, GlyphCollection glyphs, const char *customGlyphs,
+    bool distanceFieldEnabled /* = false */,float outline /* = 0 */,bool bold /* = false */)
 {
     FontFreeType *tempFont =  new (std::nothrow) FontFreeType(distanceFieldEnabled,outline);
 
+    if (bold)
+    {
+        tempFont->_boldSize = CC_CONTENT_SCALE_FACTOR() * 64;
```
drelaptop, line 61: __64__ is a magic number, it's difficult to understand the meaning. using this way:`#define TIME_DELAY_PRECISION 0.0001` is better.

drelaptop (changes requested):
Thank you for this good feature. it works well on macOS when I tested. meanwhile I have some suggestions to improve this PR, I found that after I read your code again and again. 

sorry for a long response. @treert

drelaptop:
one more things, we should rich test case, like below rich case existed,:
```C++
LabelOutlineAndGlowTest::LabelOutlineAndGlowTest()
{
    ...
    
    ttfConfig.outlineSize = 3;
    auto label5 = Label::createWithTTF(ttfConfig,"Outline and Bold", TextHAlignment::CENTER, size.width);
    label5->setPosition( Vec2(size.width/2, size.height*0.24) );
    label5->setTextColor( Color4B::RED );
    label5->enableOutline(Color4B::BLUE);
    label5->enableBold();
    addChild(label5);
}
```
another way is create a new case, and add it into `LabelTestNew` to test this feature.

treert:
drelaptop 完善测试后，发现只加粗不描边情况下显示不好。
https://github.com/cocos2d/cocos2d-x/pull/18598

PR #17437, opened 2017-03-03
cocos2d::Sequence::isDone() checks that the last action is actually done

j-jorge:
Due to approximations of float values, a call to `cocos2d::Sequence::isDone()` may return true even if the last action has not been executed.

This commit adds a test case reproducing the bug by creating a sequence made of one delay and one function call. The sequence is updated of the duration of the delay minus `FLT_EPSILON`. In this state, `isDone()` compares the elapsed time and the duration then returns `true`.

This commit adds a fix for this bug by testing that the last action is actually done in `cocos2d::Sequence::isDone()`.

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -422,6 +422,11 @@ Sequence* Sequence::reverse() const
         return nullptr;
 }
 
+bool Sequence::isDone() const
+{
+  return (_last == 1) && _actions[1]->isDone();
+}
+
 //
 // Repeat
 //
--- a/cocos/2d/CCActionInterval.h
+++ b/cocos/2d/CCActionInterval.h
@@ -182,6 +182,7 @@ public:
     //
     virtual Sequence* clone() const override;
     virtual Sequence* reverse() const override;
+    virtual bool isDone() const override;
     virtual void startWithTarget(Node *target) override;
     virtual void stop(void) override;
     /**
--- a/tests/cpp-tests/Classes/ActionsTest/ActionsTest.cpp
+++ b/tests/cpp-tests/Classes/ActionsTest/ActionsTest.cpp
@@ -92,6 +92,7 @@ ActionsTests::ActionsTests()
     ADD_TEST_CASE(ActionFloatTest);
     ADD_TEST_CASE(Issue14936_1);
     ADD_TEST_CASE(Issue14936_2);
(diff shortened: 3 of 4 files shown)


minggo:
@j-jorge thanks for your work, but i think we should change `ActionInterval::isDone()` to
```c++
bool ActionInterval::isDone() const
{
    // fix #14936 _duration is not 0, but _elapsed is 0.
//    return (_elapsed + FLT_EPSILON) >= _duration;
    
    return std::sqrt(_elapsed - _duration) < FLT_EPSILON;
}
```

And i think it is better to add the test case to `ActionsTest` instead.

j-jorge:
@minggo I have moved the test in `ActionsTest` as requested.

Concerning the test in `isDone()`, you do mean `std::abs` instead of `std::sqrt`, don't you? If I replace the test as you proposed with `std::abs` it breaks the test `Issue14936_1`.

minggo:
Oh, sorry, i mean `std::abs`. I will test if it breaks the test.

minggo:
Yep, it breaks the test. So i think your solution is right.
Thanks.

minggo:
@j-jorge i think the accurate issue will exist for all interval actions. And after merging this PR, Sequence::isDone() will never be true if action[1] is an interval action. You can run try [this test case](https://github.com/cocos2d/cocos2d-x/blob/v3/tests/cpp-tests/Classes/ActionsTest/ActionsTest.cpp#L505), the repeat action can not work correctly.

All interval actions will meet the accurate issue,  we should find a way to fix it in ActionInterval. I will revert this PR first.

Edit: I will revert it in v3.15 first.

minggo:
https://github.com/cocos2d/cocos2d-x/pull/17545 revert it

j-jorge:
I have submitted a fix in #17557.

Bilalmirza:
@minggo I think you should revert all this...... in v3.15 I am sometimes having a different result.

in a sequence when I call RemoveSelf::Create...sometimes this doesn't work.  and object remains on the screen. This worked perfectly fine in v3.14.1. In my opinion all the changes made to sequence should be reverted.

j-jorge:
@Bilalmirza Do you still have the issue if using a version of Cocos2d-x that includes #17557?

Bilalmirza:
@j-jorge Sorry i really can't find time to test this, I have to release my game on friday... I reverted back to v3.14.1

PR #17441, opened 2017-03-04
support delegate singleton object&add compress index&others useful function

owen200008:
1. assets-manager support compressindex
2.node object add noupdatecomponent,  
(when i write the richtext create 20000 node it use more time to do nothing just foreach the component, most component is no need update)
3.node add destroy event to lua(js i no use, so please check it is good)
(it is use to some node add or delete from father but not delete object, we should know the real destroy event to delete other resource)
4.listview add changeitemevent and support touch location
(it is useful and easy hightlight the item when choose, end event set unenable ,changeevent set enable,
location is useful to bussiness to show tip for the item)
5.ResizableBuffer need length it is useful
6.singleton object support setdelegate like file, it is useful to bussiness, ResizableBuffer length is use for file to decode filedata when the fieldata is encode
7.fast tmx support multi tileset and read resource with plist and support straggered


DavidDeSimone:
As a stylist note, it looks like you are indenting with tabs. Please indent with spaces for project consistency.

owen200008:
ok get it. i have change it and fix bug and export the interface for lua.

cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -60,6 +60,14 @@ SpriteFrameCache* SpriteFrameCache::getInstance()
     return _sharedSpriteFrameCache;
 }
 
+void SpriteFrameCache::setDelegate(SpriteFrameCache* pCache)
+{
+    swap(_sharedSpriteFrameCache, pCache);
+    if (pCache){
```
j-jorge: `CC_SAFE_RELEASE_NULL` tests the pointer for nullness so it's useless to test it here.

cocos/ui/UIListView.cpp
```diff
@@ -540,12 +540,38 @@ void ListView::selectedItemEvent(TouchEventType event)
...
                 _ccEventCallback(this, static_cast<int>(EventType::ON_SELECTED_ITEM_END));
             }
+            //end calc
+            if (event == TouchEventType::ENDED)
+            {
+                //only ended will know last choose one，begin just highlight
+                Widget* parent = sender;
+                while (parent)
+                {
+                    if (parent && (parent->getParent() == _innerContainer))
```
j-jorge: `parent` is necessarily not null here. The test should be `if (parent->getParent() == _innerContainer)`.

cocos/2d/CCSpriteFrameCache.cpp
owen200008: ok i will change

cocos/ui/UIListView.cpp
owen200008: ok i will change

minggo:
@owen200008 thanks for your contribution, but it is a huge PR, and do many different things in one PR. Could you please break it into more small ones? Just do one thing in a small PR, and it is better to add a test case for new feature or bug fixing. Thanks.

owen200008:
ok i will try. i'm not familiar with git push. so maybe it will be slow.

minggo:
thanks

minggo:
@owen200008 i close this PR since you will create more small PRS.

PR #17455, opened 2017-03-07
fixed #17427: lws_parse_url has wrong behavior, it parses ’ws://domain.com/abc/d’ to ‘path: abc/d’ rather than ‘path: /adb/d’.

### LINKED ISSUE #17427: websocket connected bug
- cocos2d-x version: 3.14
In JSB the webSocet connect was failed.

ws://domain.com/abc/d

It will parse the path is: abc/d

The correct path is: /abc/d

dumganhar:
Adds Uri class and parses Url to scheme, host, port, path in engine side.


cocos/network/WebSocket.cpp
```diff
@@ -28,6 +28,7 @@
  ****************************************************************************/
 
 #include "network/WebSocket.h"
+#include "network/Uri.h"
```
minggo, line 31: why not just add an anonymous function here?

dumganhar: I will update SocketIO code which will also depend on `Uri` class.

PR #17457, opened 2017-03-07
update compress support compressindex

owen200008:
update compress support compressindex

### DIFF
--- a/extensions/assets-manager/AssetsManagerEx.cpp
+++ b/extensions/assets-manager/AssetsManagerEx.cpp
@@ -49,6 +49,10 @@ NS_CC_EXT_BEGIN
 const std::string AssetsManagerEx::VERSION_ID = "@version";
 const std::string AssetsManagerEx::MANIFEST_ID = "@manifest";
 
+bool SortCompressFileIndexAssert(const CompressedFilesInfoAsserts& a, const CompressedFilesInfoAsserts& b)
+{
+    return a.m_nCompressIndex < b.m_nCompressIndex;
+}
 // Implementation of AssetsManagerEx
 
 AssetsManagerEx::AssetsManagerEx(const std::string& manifestUrl, const std::string& storagePath)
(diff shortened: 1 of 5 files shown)


extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -49,6 +49,10 @@ NS_CC_EXT_BEGIN
 const std::string AssetsManagerEx::VERSION_ID = "@version";
 const std::string AssetsManagerEx::MANIFEST_ID = "@manifest";
 
+bool SortCompressFileIndexAssert(const CompressedFilesInfoAsserts& a, const CompressedFilesInfoAsserts& b)
```
minggo, line 52: please put it into a namespace like
```c++
namespace {
 bool SortCompressFileIndexAssert(const CompressedFilesInfoAsserts& a, const CompressedFilesInfoAsserts& b) {
    ...
}
}
```

minggo:
will it break compatibility?

tools/tolua/cocos2dx_studio.ini
```diff
@@ -53,7 +53,7 @@ skip =  *::[^visit$ copyWith.* onEnter.* onExit.* ^description$ getObjectType .*
         ActionObject::[initWithDictionary initWithBinary],
         BaseData::[copy subtract],
         ActionTimelineCache::[getInstance loadActionTimelineFromXML loadAnimationWithDataBuffer createActionWithDataBuffer],
-        ActionTimeline::[setFrameEventCallFunc]
```
minggo, line 56: is it something related with this PR? If not, i think you should create a another branch for it.

extensions/assets-manager/AssetsManagerEx.cpp
owen200008: it is under  namespace cocos2d { namespace extension { 
is it necessary?  conflict it not possible.
other purpose?

owen200008:
it is compatibility. if no define it run the same with before

extensions/assets-manager/AssetsManagerEx.cpp
minggo: yep, to avoid conflict. It is better to put it into an anonymous namespace.

PR #17460, opened 2017-03-07
update compress support compressindex

owen200008:
update compress support compressindex

### DIFF
--- a/extensions/assets-manager/AssetsManagerEx.h
+++ b/extensions/assets-manager/AssetsManagerEx.h
@@ -46,11 +46,10 @@ NS_CC_EXT_BEGIN
 //! All files to be decompressed
 struct CompressedFilesInfoAsserts
 {
-    std::string     m_strFileName;
+    std::string     _fileName;
     int             m_nCompressIndex;
-    CompressedFilesInfoAsserts()
+    CompressedFilesInfoAsserts() : m_nCompressIndex(0)
     {
-        m_nCompressIndex = 0;
     }
 };
 
--- a/extensions/assets-manager/AssetsManagerEx.h
+++ b/extensions/assets-manager/AssetsManagerEx.h
@@ -47,9 +47,8 @@ NS_CC_EXT_BEGIN
 struct CompressedFilesInfoAsserts
 {
     std::string     _fileName;
-    int             m_nCompressIndex;
-    CompressedFilesInfoAsserts() : m_nCompressIndex(0)
-    {
+    int             _nCompressIndex;
+    CompressedFilesInfoAsserts() : _nCompressIndex(0){
     }
 };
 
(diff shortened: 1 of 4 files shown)


extensions/assets-manager/AssetsManagerEx.h
```diff
@@ -43,6 +43,17 @@
 
 NS_CC_EXT_BEGIN
 
+//! All files to be decompressed
+struct CompressedFilesInfoAsserts
+{
+    std::string     m_strFileName;
```
minggo: Please do the change `m_strFileName -> _fileName` to obey the coding style.

extensions/assets-manager/AssetsManagerEx.h
```diff
@@ -43,6 +43,17 @@
...
+    int             m_nCompressIndex;
+    CompressedFilesInfoAsserts()
```
minggo: CompressedFilesInfoAsserts() : xxx() {};

please initialize like above.

minggo:
this PR looks good to me, @pandamicro could you please take a look too.

extensions/assets-manager/Manifest.cpp
```diff
@@ -392,6 +398,7 @@ Manifest::Asset Manifest::parseAsset(const std::string &path, const rapidjson::V
     }
     else asset.downloadState = DownloadState::UNSTARTED;
     
+
```
minggo: please remove unneeded space

extensions/assets-manager/Manifest.h
```diff
@@ -72,6 +72,12 @@ class CC_EX_DLL Manifest : public Ref
         std::string path;
         bool compressed;
         DownloadState downloadState;
+        int m_nCompressIndex;
+        Asset()
+        {
+            compressed = false;
+            m_nCompressIndex = 0;
+        }
```
minggo: m_nCompressIndex -> compressIndex
and please use parameter initialization list

minggo: And should initialize `downloadState ` too?

extensions/assets-manager/AssetsManagerEx.h
```diff
@@ -43,6 +43,15 @@
 
 NS_CC_EXT_BEGIN
 
+//! All files to be decompressed
+struct CompressedFilesInfoAsserts
+{
+    std::string     _fileName;
+    int             _nCompressIndex;
```
minggo, line 50: i think it is better to rename to _compressIndex, `n` is used together with `m_`.

pandamicro:
@owen200008 Please take a look at this PR

https://github.com/cocos2d/cocos2d-x/pull/17115

I have fundamentally refactored AssetsManagerEx, including the decompress process, it's now much more secured, I think your PR won't be needed any more, can you confirm that ?

Thank you anyway for your effort

PR #17461, opened 2017-03-07
node add destroy event& add noupdatecomponent&setdelegate singleton object 

owen200008:
1.node object add noupdatecomponent
(when i write the richtext create 20000 node it use more time to do nothing just foreach the component, most component is no need update)
2.node add destroy event to lua(js i no use, so please check it is good)
(it is use to some node add or delete from father but not delete object, we should know the real destroy event to delete other resource)
3.singleton object support setdelegate like file, it is useful to bussiness, ResizableBuffer length is use for file to decode filedata when the fieldata is encode
4.export lua interface

### DIFF
--- a/cocos/2d/CCNode.h
+++ b/cocos/2d/CCNode.h
@@ -73,7 +73,8 @@ enum {
     kNodeOnExit,
     kNodeOnEnterTransitionDidFinish,
     kNodeOnExitTransitionDidStart,
-    kNodeOnCleanup
+    kNodeOnCleanup,
+    kNodeOnDestroy
 };
 
 class EventListener;
@@ -141,6 +142,15 @@ public:
 
     /// @{
     /// @name Setters & Getters for Graphic Properties
+
+    /*
+    setZLocalOrderOfArrive totally control the render order, it use for sometime node should cover the friendnode and sometime under the friendnode
+    */
+    void setZLocalOrderOfArrive(std::int64_t nZOrderArrived);
+    /*
+    getZLocalOrderOfArrive get the local real render order if same it will render with add child order like setLocalZOrder
+    */
+    std::int64_t getZLocalOrderOfArrive();
 
     /**
      LocalZOrder is the 'key' used to sort the node relative to its siblings.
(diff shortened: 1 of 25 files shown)


cocos/2d/CCNode.h
```diff
@@ -142,6 +143,15 @@ class CC_DLL Node : public Ref
     /// @{
     /// @name Setters & Getters for Graphic Properties
 
+    /*
+    setZLocalOrderOfArrive totally control the render order, it use for sometime node should cover the friendnode and sometime under the friendnode
+    */
+    void setZLocalOrderOfArrive(std::int64_t nZOrderArrived);
```
minggo, line 149: local z order of arrival is not visible outside, we take a lot of effect to make it as it is now.

cocos/2d/CCNode.h
```diff
@@ -939,15 +950,9 @@ class CC_DLL Node : public Ref
...
         static_assert(std::is_base_of<Node, _T>::value, "Node::sortNodes: Only accept derived of Node!");
-#if CC_64BITS
         std::sort(std::begin(nodes), std::end(nodes), [](_T* n1, _T* n2) {
             return (n1->_localZOrderAndArrival < n2->_localZOrderAndArrival);
         });
-#else
-        std::stable_sort(std::begin(nodes), std::end(nodes), [](_T* n1, _T* n2) {
-            return n1->_localZOrder < n2->_localZOrder;
-        });
-#endif
```
minggo, line 950: these codes are used to optimize in 64 bit architecture. please don't modify it

minggo:
Can this PR be divided into more small ones? Please do one thing in a PR.

cocos/2d/CCNode.h
owen200008: i should change the render seq. like tab view button. clk is cover the bg, no clk is under the bg.
if it can control the render seq it is the easy way to dev

owen200008: from 3.9  render seq is ++from a static val and the zorder.
i think it is good idea to control the render seq.
now is only zorder. so if you can apply change child seq to change render seq

owen200008: some ui affect is need change render seq with frends node

PR #17349, opened 2017-02-13
Update flatbuffers to 1.5

minggo:

PR #17465, opened 2017-03-07
csbload error bug

owen200008:
csbload error bug


minggo:
thanks, though cocos studio is canceled, but pull request is welcome.

minggo:
@owen200008 as https://github.com/cocos2d/cocos2d-x/issues/17869 said, the reader breaks compatibility after updating flatbuffers, so i will revert the the PR https://github.com/cocos2d/cocos2d-x/pull/17349. I am not sure if this PR should be revert too.

PR #17468, opened 2017-03-08
Prevent signed/unsigned mismatch

mogemimi:
Hello, when building Lua-bindings with Visual Studio 2015, I get the following warning message:

```
9>..\manual\CCLuaStack.cpp(814): warning C4018: '<' : signed/unsigned mismatch
```

So this PR replaces index-based for loop with range-based one to prevent the signed/unsigned mismatch warning.

Thanks!


minggo:
thanks

PR #17469, opened 2017-03-08
i have check the CSParseBinary_generated and define

owen200008:
i have check the CSParseBinary_generated and define


cocos/editor-support/cocostudio/CSParseBinary_generated.h
```diff
@@ -925,22 +925,22 @@ struct ButtonOptions FLATBUFFERS_FINAL_CLASS : private flatbuffers::Table {
     VT_DISABLEDDATA = 10,
     VT_FONTRESOURCE = 12,
     VT_TEXT = 14,
-    VT_ISLOCALIZED = 16,
```
minggo, line 928: this file is generated by flatbuffer, why modify it?

PR #17481, opened 2017-03-09
ShaderTest wrong culling calculation

sandyre:
I took the code for EffectSprite from test, and it didn't work with that culling check. So I took whole block of culling check from Sprite::draw() method, and it fixed my problem.


minggo:
i don't think we need to do the modification because the test case only has one camera.

PR #17483, opened 2017-03-10
Cpp test code for Pull request (Issue 17476)

dadidzf:
It's only happend on Windows system, because the address for two click(touch) has the same address most times!
We have two squares,  A (Left) and square B(Right), A is above of B, Both A and B have one by one touch listeners, and it will not swallow touches.
Step 1, First click at the intersection of A and B, do not move, got a touch, suppose it's address is 0x12345678
    Touch press
        A - onTouchBegan, return true, Color changed, listener->_claimedTouches is true for A
        B - onTouchBegan, return true, Color changed, listener->_claimedTouches  is true for B
    Touch release
        A - onTouchEnded, call event->stopPropagation, Color recover, listener->_claimedTouches is cleared
        B - Because A stopPropagation, Color will not be recovered, listener->_claimedTouches is not cleared

Step 2, Second click outside the two squares A and B, do not move, the address has a great probability to be 0x12345678 on Windows system, when it's 0x12345678, the situation below happened !
    Touch press
        A - onTouchBegan, return false, Color not changed, listener->_claimedTouches is false
        B - onTouchBegan, return false, Color not changed, listener->_claimedTouches is still true
    Touch release
       A - not do anything, because the listener->_claimedTouches is false, the onTouchEnd is not called
       B - onTouchEnded is called, the Color recovered

More -
1, The listener->_claimedTouches is not cleared for all system once Step 1 is done
2, We can only test the Step 2 on Windows, it's because windows has a great probaility to have the same address for one by one touch ! and the listener->_claimedTouches record the address of the touch to flaged a touch begin!

CCEventListenerTouch.h (line 94)
std::vector<Touch*> _claimedTouches;

### DIFF
--- a/tests/cpp-tests/Classes/NewEventDispatcherTest/NewEventDispatcherTest.cpp
+++ b/tests/cpp-tests/Classes/NewEventDispatcherTest/NewEventDispatcherTest.cpp
@@ -27,6 +27,7 @@ EventDispatcherTests::EventDispatcherTests()
     ADD_TEST_CASE(PauseResumeTargetTest);
     ADD_TEST_CASE(Issue4129);
     ADD_TEST_CASE(Issue4160);
+    ADD_TEST_CASE(Issue17476);
     ADD_TEST_CASE(DanglingNodePointersTest);
     ADD_TEST_CASE(RegisterAndUnregisterWhileEventHanldingTest);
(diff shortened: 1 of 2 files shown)


minggo:
Why not add the test case in 17476 too?

dumganhar:
DONE at #17527
Closed this one.

PR #17489, opened 2017-03-13
Fixes memory issues caught by Xcode Analyze Tool.

dumganhar:


cocos/base/CCProperties.cpp
```diff
@@ -480,10 +480,8 @@ char* Properties::trimWhiteSpace(char *str)
         return str;
     }
 
-    char *end;
-
     // Trim leading space.
-    while (isspace(*str))
+    while (*str != '\0' && isspace(*str))
```
dumganhar: Have to check whether is the end of string first.

dumganhar, line 484: Xcode reports this warning:
screen shot 2017-03-13 at 11 28 26

dumganhar: But `isspace('\0')` will definitely return `false`, therefore it seems a wrong report by Xcode?

cocos/ui/UIEditBox/Mac/CCUIPasswordTextField.m
```diff
@@ -99,24 +99,28 @@ - (void)editWithFrame:(NSRect)aRect
...
 
 @implementation CCUIPasswordTextField
-{
-}
+
 -(id) initWithFrame:(NSRect)frameRect
 {
-    if ([super initWithFrame:frameRect]) {
-     
+    if (self = [super initWithFrame:frameRect]) {
```
dumganhar, line 112: Should assign self from super init method.

Issue #17374, opened 2017-02-16
Android Studio support


minggo:
goole guys sent a PR about Android Studio support: https://github.com/cocos2d/cocos2d-x/pull/15676. I tested it and it works. Currently there is only one issue that, engine native codes and other 3rd party c++ codes can not be shown. But can use `step into` to refer to it when debugging.

Cocos console command should adjust for it too.

minggo:
tests and templates are fixed in https://github.com/cocos2d/cocos2d-x/pull/17389

@natural-law it's your turn to update cocos console command.

PR #17495, opened 2017-03-14
Improve the AndroidStudio configurations of tests & templates

natural-law:
Related issue : #17374


cocos/platform/android/libcocos2dx-with-controller/build.gradle
```diff
@@ -6,7 +6,7 @@ android {
 
     defaultConfig {
         minSdkVersion 10
-        targetSdkVersion 22
+        targetSdkVersion PROP_TARGET_SDK_VERSION
```
formatCvt, line 9: are you sure ` targetSdkVersion android-13` is correct value for `targetSdkVersion` key in `defaultConfig`?

PR #17511, opened 2017-03-15
Add FileUtils::getMD5Hash

pandamicro:


pandamicro:
@dumganhar Travis can't found `openssl/md5.h` on Android platform, what should I do to solve it?

minggo:
Please add this function into Utils.

PR #17512, opened 2017-03-15
VS and XCode project files have non-existent file references

rh101:
Removed and renamed several files that do not exist or have been renamed, but the project files had never been updated.  This was causing an issue building with Visual Studio 2015/2017, with a "Project is out of date" warning every time the program is run from within the IDE.


dumganhar:
@rh101 , good catch.
But we need to rename `.hpp` to `.h` for header files since most of header files are named as `***.h`.

rh101:
All good! For reference, the specific files causing the "Project is out of date" message to pop up in Visual Studio can be found by setting the MSBuild output verbosity to Detailed or Diagnostic.  This is in the menu under Tools – Options – Projects and Solutions – Build and Run.

dumganhar:
@rh101 , so could you update this Pull Request to use `.h` instead of `.hpp`? Thanks.

rh101:
All done, I think. Let me know if the changes don't show up for some reason.

dumganhar (approved):

minggo:
thanks

PR #17516, opened 2017-03-15
fix an Accelerometer second to microsecond convertion error in Android platform

blackcruiser:
fix an Accelerometer second to microsecond convertion error in Android platform
1 second = 1000000 microsecond


dumganhar (approved):
Good catch 👍

dumganhar:
@minggo, this pull request could be merged.
The docs is at:

https://developer.android.com/reference/android/hardware/SensorManager.html

samplingPeriodUs	int: The rate sensor events are delivered at. This is only a hint to the system. Events may be received faster or slower than the specified rate. Usually events are received faster. The value must be one of SENSOR_DELAY_NORMAL, SENSOR_DELAY_UI, SENSOR_DELAY_GAME, or SENSOR_DELAY_FASTEST or, the desired delay between events in microseconds. Specifying the delay in microseconds only works from Android 2.3 (API level 9) onwards. For earlier releases, you must use one of the SENSOR_DELAY_* constants.

dumganhar:
[ci rebuild]

dumganhar:
You need to send it to v3 branch...

dumganhar:
v4-develop branch is paused currently. I closed this issue.

PR #17517, opened 2017-03-15
Support for the tvOS micro gamepad as a CCController

pyrosphere:
Adds support for the tvOS [microGamepad](https://developer.apple.com/reference/gamecontroller/gcmicrogamepad). This allows some games that already support CCController to be playable on tvOS (limited to the dpad and A/X buttons).


minggo:
thanks

PR #17520, opened 2017-03-16
Add utils::getMD5Hash

pandamicro:


cocos/base/ccUtils.cpp
```diff
@@ -408,6 +409,21 @@ Node* findChild(Node* levelRoot, int tag)
     return nullptr;
 }
 
+std::string getMD5Hash(const std::string &filename)
```
dumganhar: The function name seems not to be relative to its functionality.
This function is used to get the MD5 of the file content.
But `getMD5Hash` seems to get a MD5 string from a string or data.
It's confused for me.

Probably, `getFileMD5Hash` for file, `getDataMD5Hash` for data and string?

minggo: Good idea.

pandamicro: getFileMD5Hash is more representative

PR #17523, opened 2017-03-16
Update submodule cocos2d-console.

natural-law:
Changed: Disable the command `cocos gen-templates`.


minggo:
the command is used by cocos studio, and it can not work and be not maintained for a long time

PR #17527, opened 2017-03-16
Updates PR #17476 (Remove _claimedTouches Flags for listeners)

dumganhar:
PR #17476
Uses range loop for clearing claimed touches in the next touch begin event.

Image We have two sprite squares, A (Left) and square B(Right), A is above of B, Both A and B have one by one touch listeners, and it will not swallow touches.

Click at the intersection of A and B, do not move
Touch press
A - onTouchBegan, return true, listener->_claimedTouches is true for A
B - onTouchBegan, return true, listener->_claimedTouches is true for B
Touch release
A - onTouchEnded, we manually call event->stopPropagation() in the onTouchEnd function !!! listener->_claimedTouches is cleared
B - Because A stopPropagation, listener->_claimedTouches is not cleared

In this case, Because A stop propagation for touch end, the listener->_claimedTouches for B will not be cleared forever ! because B has touch begin, which record the touch in listener->_claimedTouches, but it has not touch end !

### DIFF
--- a/cocos/base/CCEventDispatcher.cpp
+++ b/cocos/base/CCEventDispatcher.cpp
@@ -973,7 +973,30 @@ void EventDispatcher::dispatchTouchEvent(EventTouch* event)
     //
     if (oneByOneListeners)
     {
-        auto mutableTouchesIter = mutableTouches.begin();
+        if (event->getEventCode() == EventTouch::EventCode::BEGAN)
+        {
+            auto fixedPriorityListeners = oneByOneListeners->getFixedPriorityListeners();
+            auto sceneGraphPriorityListeners = oneByOneListeners->getSceneGraphPriorityListeners();
+            if (fixedPriorityListeners != nullptr)
+            {
+                for (auto lIter = fixedPriorityListeners->begin(); lIter != fixedPriorityListeners->end(); lIter++)
+                {
+                    EventListenerTouchOneByOne* listener = static_cast<EventListenerTouchOneByOne*>(*lIter);
+                    listener->_claimedTouches.clear();
+                }
+            }
+
+            if (sceneGraphPriorityListeners != nullptr)
+            {
+                for (auto lIter = sceneGraphPriorityListeners->begin(); lIter != sceneGraphPriorityListeners->end(); lIter++)
+                {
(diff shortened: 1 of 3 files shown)


dumganhar:
DON'T MERGE IT NOW.
I have to check the logic again.

cocos/base/CCEventDispatcher.cpp
```diff
@@ -973,7 +973,30 @@ void EventDispatcher::dispatchTouchEvent(EventTouch* event)
...
+        if (event->getEventCode() == EventTouch::EventCode::BEGAN)
+        {
+            auto fixedPriorityListeners = oneByOneListeners->getFixedPriorityListeners();
+            auto sceneGraphPriorityListeners = oneByOneListeners->getSceneGraphPriorityListeners();
+            if (fixedPriorityListeners != nullptr)
+            {
+                for (auto l : (*fixedPriorityListeners))
+                {
+                    EventListenerTouchOneByOne* listener = static_cast<EventListenerTouchOneByOne*>(l);
+                    listener->_claimedTouches.clear();
```
dumganhar, line 985: @dadidzf , if clear `_claimedTouches`, it may break the `multi-touch` logic of using EventListenerTouchOneByOne. I will check `TouchTest` to see whether it works after this fix.

dumganhar: I confirmed that this patch breaks `cpp-tests/63:Touches`

dadidzf: @dumganhar Thanks, Do you know why it breaks? I test the cpp-tests/63, it is ok ! it seems multi-touch do not use _claimedTouches

dumganhar: @dadidzf , just test it on iOS or Android mobile phones. Using multi-touch on the white blocks. The second time you touch the blocks, it will not be responsive.

minggo:
@dadidzf do you have any question about this PR? If not, i will merge it. Thanks.

dumganhar:
@minggo , please don't merge this PR.
There are some issues.

dadidzf:
@minggo This patch do cause some issues as @dumganhar described ,  it can not be merged! we need to find a different way to fix it ! ^_^

minggo:
Got it, thanks.

PR #17528, opened 2017-03-16
Update Cocos2dxDownloader.java

Dauch:


dumganhar:
@Dauch, good catch, it seems to be a logic error there.
BTW, what kind of problem will it cause?

Dauch:
When ```path``` is a directory, I think it will may crash

dumganhar (approved):

PR #17532, opened 2017-03-16
unify sensor timestamp units in Android / iOS / Tizen

blackcruiser:
unify sensor timestamp units in Android / iOS / Tizen
Android use nanosecond, iOS use second, Tizen use millisecond


dumganhar:
@blackcruiser , thanks for your contribution.
I found these:
iOS: https://developer.apple.com/reference/coremotion/cmlogitem/1615939-timestamp?language=objc
Android: https://developer.android.com/reference/android/hardware/SensorEvent.html
Tizen: https://developer.tizen.org/dev-guide/native/2.3.0/org.tizen.mobile.native.apireference/structsensor__event__s.html

But Tizen documentation doesn't say the unit of `timestamp`. Where could I find that information?

blackcruiser:
@dumganhar 
Here said Tizen use milliseconds:
https://developer.tizen.org/dev-guide/native/2.3.0/org.tizen.mobile.native.appprogramming/html/guide/system/sensor.htm
But I found another place said it use nanoseconds:
https://developer.tizen.org/development/api-guides/native-application/system/sensor#accelerometer

Due to the definition of stuctsensor_event_s, timestap is unsigned long long, it may use nanosecond as said above.
I can't check it, since I don't have a Tizen device.

dumganhar:
@zilongshanren , you have a tizen device, could you please check the unit of timestamp?

dumganhar:
@blackcruiser , I check the unit of timestamp on tizen 2.4 device, I confirm that its unit is milliseconds. Therefore, your patch is correct. Thanks.

dumganhar:
@blackcruiser , my mistake, its unit isn't milliseconds. It's microseconds.
```
04-01 17:40:44.192 : DEBUG / cjh ( 1386 : 1386 ) : cjh interval: 30335
04-01 17:40:44.232 : DEBUG / cjh ( 1386 : 1386 ) : cjh interval: 29693
04-01 17:40:44.242 : DEBUG / cjh ( 1386 : 1386 ) : cjh interval: 29938
04-01 17:40:44.282 : DEBUG / cjh ( 1386 : 1386 ) : cjh interval: 30029
04-01 17:40:44.312 : DEBUG / cjh ( 1386 : 1386 ) : cjh interval: 29999
```
So the code should divide `1e6` rather than `1e3` for tizen.

dumganhar:
New patch for tizen: https://github.com/cocos2d/cocos2d-x/pull/17629

PR #17533, opened 2017-03-16
fix Cocostudio runtime error

deflinhec:
Application crash while loading csb file which contain button widget.


deflinhec:
Seems to be duplicate PR, but how should I generate this file?

PR #17537, opened 2017-03-17
Gles3 support

minggo:


Bilalmirza:
@minggo will this be added to v3.15?

minggo:
Not, it will be in the version after v3.15, because we haven tested v3.15 before this PR. You can include it in your project. It is not hard.

liangzi10101:
Sorry, even I follow your step that change the code of '**cocos/platform/ios/CCES2Renderer-ios.m**', I still   can't create OpenGL ES 3.0 context on iOS devices, because I saw that the output information '**gl.version: OpenGL ES 2.0 IMGSGX543-113.3**', is there anything wrong with me?

I just debugged the program and found that after the execution of statement: 'context_ = [[EAGLContext alloc] initWithAPI:kEAGLRenderingAPIOpenGLES3];', the '**context_**' actually was '**nil**', but I don't know why? Because my cocos2d-x is 3.17

darreney:
Hi @minggo , i know this is old, but i came upon this while looking for a method to use gles3 on Android.
Have tried it, and it worked for device that supports gles3.
However for devices that only use gles2 (tested on emulator),
it does not enter into this code: Line 408 Cocos2dxActivity.java 
if (context == null)

instead, the check should be this:
**if (context.equals(EGL10.EGL_NO_CONTEXT))** 

according to https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglCreateContext.xhtml
"If eglCreateContext fails to create a rendering context, EGL_NO_CONTEXT is returned."

PR #17548, opened 2017-03-18
Added VS2017 support in project files

rh101:
Added VS2017 support in the project files, since loading a project in VS2017 doesn't automatically select the right build tools (it kept defaulting to VS2010).


minggo:
thanks

PR #17551, opened 2017-03-20
[win32] Fixes cpp-tests/FileUtilsTest crash.

dumganhar:
http://punchbox.info:3000/issues/26165


tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
```diff
@@ -139,15 +139,15 @@ void TestSearchPath::onEnter()
 
     // Save old resource root path
     std::string oldDefaultRootPath = sharedFileUtils->getDefaultResourceRootPath();
-    sharedFileUtils->setDefaultResourceRootPath("extensions");
+    sharedFileUtils->setDefaultResourceRootPath(oldDefaultRootPath + "extensions");
```
dumganhar, line 142: This line fix the issue. The new resource root path must be associated with the old one, then plus "extension".

PR #17556, opened 2017-03-20
fixed #17497: Win32 performance is bad.

### LINKED ISSUE #17497: Win32 performance is bad.
- cocos2d-x version:14.1
- devices test on:Win32
- developing environments
   - VS version: vs community 2015 14.0.25431.01

Steps to Reproduce:
print the fps.
 if (dt>0.0167f){
cclog("dt=%f",dt);
}

Our game is running with steady 0.0166667f on all devices ,but on win32 the game is lagging 
i tried to print the fps since i noticed that the update function is not steady like other platforms.
  if you can see the fps go up to 0.0200f.

capture

on the next image the game on mac is running without any issue and the fps perfect.

screen shot 2017-03-14 at 4 56 25 pm

dumganhar:
The 'Sleep(1)' in FPS controlling code isn't precise enough. Windows sets timer precision from 1ms to 15ms, we need to set the precision to the highest one (1ms). Although it's still not enough, we need to make some tricks in the FPS controlling code, that is, don't sleep if the interval is equal or less than 1 millisecond.

### DIFF
--- a/cocos/platform/win32/CCApplication-win32.cpp
+++ b/cocos/platform/win32/CCApplication-win32.cpp
@@ -63,6 +63,18 @@ int Application::run()
 {
     PVRFrameEnableControlWindow(false);
 
+    ///////////////////////////////////////////////////////////////////////////
+    /////////////// changing timer resolution
+    ///////////////////////////////////////////////////////////////////////////
+	UINT TARGET_RESOLUTION = 1; // 1 millisecond target resolution
+	TIMECAPS tc;
+	UINT wTimerRes = 0;
+	if (TIMERR_NOERROR == timeGetDevCaps(&tc, sizeof(TIMECAPS)))
+	{
+		wTimerRes = std::min(std::max(tc.wPeriodMin, TARGET_RESOLUTION), tc.wPeriodMax);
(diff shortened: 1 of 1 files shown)


cocos/platform/win32/CCApplication-win32.cpp
```diff
@@ -83,19 +95,32 @@ int Application::run()
...
     while(!glview->windowShouldClose())
     {
         QueryPerformanceCounter(&nNow);
-        if (nNow.QuadPart - nLast.QuadPart > _animationInterval.QuadPart)
+        interval = nNow.QuadPart - nLast.QuadPart;
+        if (interval >= _animationInterval.QuadPart)
         {
-            nLast.QuadPart = nNow.QuadPart - (nNow.QuadPart % _animationInterval.QuadPart);
-            
+            nLast.QuadPart = nNow.QuadPart;
```
dumganhar, line 110: I don't know why should subtract `nNow.QuadPart % _animationInterval.QuadPart` here.
I think the correct logic should be `nLast.QuadPart = nNow.QuadPart`.

minggo (approved):

PR #17557, opened 2017-03-20
Flag the end of ActionInstant and ActionInterval.

j-jorge:
Following #17437 and #17545.

This commit adds a flag to `ActionInstant` and `ActionInterval` to mark the end of the action.

`ActionInstant` sets the flag to true only when the action has actually been updated.

`ActionInterval` sets the flag to true only when the elapsed time becomes equal or greater than the action's duration.

`ActionInterval` does not overrides zero durations with `FLT_EPSILON` anymore. Thus a `Spawn` or a `Sequence` of several `ActionInstant` has a duration equal to zero, as expected.

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -89,17 +89,10 @@ bool ActionInterval::initWithDuration(float d)
 {
     _duration = d;
 
-    // prevent division by 0
-    // This comparison could be in step:, but it might decrease the performance
-    // by 3% in heavy based action games.
-    if (_duration <= FLT_EPSILON)
-    {
-        _duration = FLT_EPSILON;
-    }
-
     _elapsed = 0;
     _firstTick = true;
-
+    _done = false;
+    
     return true;
 }
 
@@ -117,8 +110,7 @@ bool ActionInterval::sendUpdateEventToScript(float dt, Action *actionObject)
 
 bool ActionInterval::isDone() const
 {
-    // fix #14936 _duration is not 0, but _elapsed is 0.
-    return (_elapsed + FLT_EPSILON) >= _duration;
+    return _done;
 }
 
 void ActionInterval::step(float dt)
@@ -141,6 +133,8 @@ void ActionInterval::step(float dt)
     if (sendUpdateEventToScript(updateDt, this)) return;
     
     this->update(updateDt);
+
+    _done = _elapsed >= _duration;
 }
 
 void ActionInterval::setAmplitudeRate(float /*amp*/)
@@ -162,6 +156,7 @@ void ActionInterval::startWithTarget(Node *target)
     FiniteTimeAction::startWithTarget(target);
     _elapsed = 0.0f;
     _firstTick = true;
+    _done = false;
 }
 
 //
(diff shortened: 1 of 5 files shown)


cocos/2d/CCActionInterval.cpp
```diff
@@ -141,6 +133,8 @@ void ActionInterval::step(float dt)
     if (sendUpdateEventToScript(updateDt, this)) return;
     
     this->update(updateDt);
+
+    _done = _elapsed >= _duration;
```
minggo, line 137: why not do it in ActionInterval::isDone()

j-jorge: If I do the test in `isDone` for an action with a zero duration then the action may say it is done while it is not. Consider the following case:

```c++
CallFunc* func( CallFunc::create( [] {} ) );
Action* action( Sequence::create( func, func, nullptr ) );
// the action was not started, it should not be done
assert( !action->isDone() );
```

The sequence is made of instant actions, thus its duration is zero. In this case `_elapsed >= _duration` is verified even if the action was not started. By using a flag updated when the action is updated we ensure that the action has been executed at least once before being done.

minggo: make sense, i will test it

minggo:
@j-jorge A developer reported [an issue](http://discuss.cocos2d-x.org/t/ccsequence-goes-on-endless-in-latest-unstable-build-from-github/35947) of `Sequence`. I am busing for releasing v3.15. I think it is something related with this PR. Could you please take a look?

Thanks.

j-jorge:
@minggo Sure, I will try the issue today.

j-jorge:
@minggo Everything seems to work as expected: https://github.com/j-jorge/cocos2d-x/tree/infinite-sequence

Can you confirm the validity of the test?

minggo:
Thanks @j-jorge . I don't have time to test it. I will do after releasing v3.15.

Dauch:
please have a look at this? #17884 @j-jorge many thanks~

cocos/2d/CCActionInterval.cpp
minggo: @j-jorge This may have issue if the last action of Sequence is an instant action. For example

```
auto sequence = Sequence(..., CallFunc::create( [] {} ));
```

when _elapsed == _duration, then CallFunc will not be invoked. Can refer to https://github.com/cocos2d/cocos2d-x/issues/17884 for detail information.

minggo:
@j-jorge this PR brings many issues, such as https://github.com/cocos2d/cocos2d-x/issues/17884. In order to fix #17884 i sent #17906 to fix it but i can not fix all the issues and may break compatibility.

The reason is that, sequence, spawn and similar actions will invoke Action::update instead of Action::step, which lead to that `_done` is not set correctly.

I think we should find other method to fix the issue resolved in this PR.

j-jorge:
@minggo I have followed these issues, unfortunately I did not have the time to work on them. The changes you made in #17906 seem OK to me so far but as you said it is quite difficult not to introduce side effects. Unfortunately I do not see how to fix every `Sequence` related issue with a no-side effect guarantee :/

As a side note, IMHO there should be some unit tests in Cocos to confirm that our changes do not break anything. Today there is a test app, which is fine as a demo, but which does not help to prevent regressions. It includes several tests for various issues (e.g. `Issue1305`, `Issue1288` in `ActionsTest.cpp`) but one cannot run all the tests to ensure that a code change did not break anything. Having a collection of unit tests to be run at each compilation would prevent some mistakes.

minggo:
Yep, i am adding the test cases and fixing the issues.

PR #17586, opened 2017-03-25
Fix bug with unresolved symbols in Android Studio 2.3

ilqvya:


templates/cpp-template-default/proj.android/jni/Android.mk
```diff
@@ -11,9 +11,9 @@ LOCAL_MODULE := MyGame_shared
 
 LOCAL_MODULE_FILENAME := libMyGame
 
-LOCAL_SRC_FILES := hellocpp/main.cpp \
-                   ../../Classes/AppDelegate.cpp \
-                   ../../Classes/HelloWorldScene.cpp
```
minggo, line 16: Can not change it like this for ant project. Or it will have compile issues as:

* create a new cpp project
* cocos compile -p android

meeting compiling issue. But it seems there is not problem when compiling android studio project. So i will revert it for ant project.

ilqvya: Got it! Nice work 
aqua

PR #17634, opened 2017-04-03
Adding additional async functions to FileUtils

DavidDeSimone:
Adding async functions for FileUtils::listFiles and FileUtils::listFilesRecursively. This will allow engine users to perform these functions off thread.

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -1226,6 +1226,14 @@ std::vector<std::string> FileUtils::listFiles(const std::string& dirPath) const
     return files;
 }
 
+void FileUtils::listFiles(const std::string& dirPath, std::function<void(std::vector<std::string>)> callback) const
+{
+    auto fullPath = fullPathForFilename(dirPath);
+    performOperationOffthread([fullPath]() {
+        return FileUtils::getInstance()->listFiles(fullPath);
+    }, std::move(callback));
+}
+
 void FileUtils::listFilesRecursively(const std::string& dirPath, std::vector<std::string> *files) const
 {
     std::string fullpath = fullPathForFilename(dirPath);
@@ -1289,6 +1297,16 @@ void FileUtils::listFilesRecursively(const std::string& dirPath, std::vector<std
         }
         tinydir_close(&dir);
     }
+}
+
+void FileUtils::listFilesRecursively(const std::string& dirPath, std::function<void(std::vector<std::string>)> callback) const
+{
+    auto fullPath = fullPathForFilename(dirPath);
+    performOperationOffthread([fullPath]() {
+        std::vector<std::string> retval;
+        FileUtils::getInstance()->listFilesRecursively(fullPath, &retval);
+        return retval;
+    }, std::move(callback));
 }
 
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32) || (CC_TARGET_PLATFORM == CC_PLATFORM_WINRT)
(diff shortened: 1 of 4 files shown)


Rubonnek:
[ci rebuild]

cocos/platform/CCFileUtils.h
```diff
@@ -809,6 +809,14 @@ class CC_DLL FileUtils
      *  @return File paths in a string vector
      */
     virtual std::vector<std::string> listFiles(const std::string& dirPath) const;
+
+    /**
+     * List all files in a directory async, off of the main cocos thread.
+     *
+     * @param dirPath The path of the directory, it could be a relative or an absolute path.
+     * @param callback The callback to be called once the list operation is complete. Will be called on the main cocos thread.
+     */
```
minggo: It seems the error happens in binding codes. Can add comments to disable binding like:

@lua NA
@js NA

DavidDeSimone:
@minggo  It looks like the binding generator is still creating bindings for these functions, even with @js NA and @lua NA. Is there anything else I need to do?

minggo:
Its my fault. It seems to modify [the ini](https://github.com/cocos2d/cocos2d-x/blob/v3/tools/tolua/cocos2dx.ini) to skip the binding. And should do the same thing for js binding too.

cocos/platform/CCFileUtils.cpp
```diff
@@ -1226,6 +1226,14 @@ std::vector<std::string> FileUtils::listFiles(const std::string& dirPath) const
     return files;
 }
 
+void FileUtils::listFiles(const std::string& dirPath, std::function<void(std::vector<std::string>)> callback) const
```
dumganhar: What about adding `&` reference symbol for `callback` parameter?

DavidDeSimone: I would like to keep the callback by value, so that users can better leverage c++ 11 move semantics to avoid copies. 

If we take the callback by const reference (const &), then when we assign it into our execution queue to be executed off-thread with the AsyncTask pool, we will be making a copy (since it is const and we cannot mutate it). 

If we take the callback by l-value reference (&), then we can call std::move on it to avoid a copy. However, then our code will not accept r-value callbacks. Example:

```
Async::enque(std::function<void(void)>& callback)
{

}

void foo() 
{
auto bar = [] () { };
Async::enque(bar); // Compiles, will not copy bar.
Async::enque([] () {}); // Fails to compile, lambda is an r-value.
}
```
Likewise, taking the callback by r-value reference (&&), we will have the opposite problem (we cannot accept bar without std::move'ing into it, which we may not always want to do.)

**However**, we can have the best of both worlds by taking by value, and using std::move! Example:

```
Async::queue(std::function<void(void)> callback) 
{
    // If I took a const reference, I couldn't std::move here, and would have to copy.
    _queue.push_back(std::move(callback)); // This is safe because I own callback.
}

foo() {
auto lambda1 = [] () { };
auto lambda2 = [] () { };
Async::enque(lambda1) // This will copy, BUT I save a copy by calling std::move in enque. 
Net result is I have the same number of copies as if I took a const reference. 
Async::enque(std::move(lambda2)); // This will move, much cheaper then copy. 
Async::enque([] () {}); // r-value, avoids copy.
}
```
As you can see, even if the user forgets to call std::move at the call site, I have the same number of copies that I would if I took the callback by const reference. And if the user calls std::move at the function callsite, we can make no copies at all! We get the benefits of using c++ objects on the stack, with traceable lift times, but we have the performance of code that is only performing pointer copies, and not expensive copy constructors. I feel that cocos2d-x could better leverage move semantics as a c++11 game engine, so it's good to have examples on how to use move semantics in the code base.

DavidDeSimone:
@minggo thanks for the heads up, and thanks @dumganhar for the feedback. Let me know if you would like anything else in this PR, or if you have additional feedback on the code.

minggo:
thanks

PR #17637, opened 2017-04-03
V3.15 win10 fixed build properties for compiling md5.c

stammen:
This PR fixes the Win10 build by correcting the build properties for the newly added file md5.c.
-turned off the /ZW flag (Consume Windows Runtime Extension) which is not compatible with .c files
- disabled Precompiled headers


stammen:
Wrong branch. Will make PR to v3.15

PR #17661, opened 2017-04-06
Tiled sprite

newnon:
Tiled sprite support for cocos

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -665,6 +665,203 @@ void Sprite::updatePoly()
         // to avoid memcpy'ing stuff
         _polyInfo.setTriangles(triangles);
     }
+    else if (_renderMode == RenderMode::TILE) {
+        float u0, u1;
+        float v0, v1;
+        float uo, vo;
+        
+        Texture2D *tex = (_renderMode == RenderMode::QUAD_BATCHNODE) ? _textureAtlas->getTexture() : _texture;
+        if (tex == nullptr)
+            return;
+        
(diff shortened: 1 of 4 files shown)


minggo:
Currently, Sprite is very complex. Adding scale9 sprite is just to support creator output. Adding a new mode you have to do the full testing with all other modes, all other Sprite's APIs. It is very easy to make it not work.

So i don't want to add this feature. It will make Sprite out of control.

newnon:
@minggo may be add this to another class like scale9sprite or even in separate class?

minggo:
Yep. What place do you think is better? Put it into `ui`? Can it be part of a type of UI element? Could you please paste some picture of this feature?

newnon:
Tiling but you can use any sprite frame
In my project i place this code inside Scale9Sprite.
There is 5 variants where to place it
1. Sprite (pros it can be used for a any place, cons sprite became to complex)
2. Scale9Sprite(it can be uses from Scale9Sprite or ui::ImageView cons Scale9Sprite now is empty and it is not so good name for it)
3. Make some Advanced sprite and place it there and make ui::ImageView use it instead Scale9Sprite(pros it possible to use this code like standalone class or from ui, it is good idea to refactor ImageView a little cons little bit more work)
4. Inside ImageView(pros easy cons it will available only from ui)
5. Inside some Advanced sprite and add new TiledImage class inside UI (pros it possible to use this code like standalone class or from ui, lowest chance to broke something, cons many code will be same with imageView)

as for me and my projects suitable are 3-rd and 5-th variants

PR #17666, opened 2017-04-07
[Label] Fix distance map computation

sp-dani-garcia:
**Problem:**

Distance map was not computed correctly. Some characters were "cut" and were not shown properly. You can see a label displaying the problem in the image below (notice S, O, P and T characters for example):

captura de pantalla 2017-04-07 a las 13 24 58

Here you can see the distance map texture generated for that text:

captura de pantalla 2017-04-07 a las 11 26 52

You can see 2 problems in the generated texture:
1- The distance map spread added to each char glyph is greater below the glyph than above the glyph when it should be the same.
2- Glyphs are not fading out correctly producing hard edges instead of a smooth gradient.

**Fixes:**
- Commit ef6c5feb63d76bce3747c3b3582d8d6168a8b795 fixes the first problem. 

You can see in the generated texture how the spread now is equal below and above the glyph:

captura de pantalla 2017-04-07 a las 11 30 15

- Commit 504b26834b8c354737b3f2349312c1663b3af325 fixes the second problem. It uses the same solution applied in [freetype-gl](https://github.com/rougier/freetype-gl).

You can see in the generated texture how the hard edges disappear and now there is a smooth gradient.

captura de pantalla 2017-04-07 a las 11 35 15

**Final result:**

captura de pantalla 2017-04-07 a las 13 25 30

### DIFF
--- a/cocos/2d/CCFontFreeType.cpp
+++ b/cocos/2d/CCFontFreeType.cpp
@@ -30,6 +30,7 @@ THE SOFTWARE.
 #include "base/CCDirector.h"
 #include "base/ccUTF8.h"
 #include "platform/CCFileUtils.h"
+#include <limits>
 
 NS_CC_BEGIN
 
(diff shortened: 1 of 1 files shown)


dumganhar:
@sp-dani-garcia , sorry for our late reply. Thanks for your contribution, we will review your patch.
Is there any reason that you close this pull request?

dumganhar:
@sp-dani-garcia, could you also add a test case to cpp-tests -> `Node: Label - New API`? Thanks.

sp-dani-garcia:
Hi @dumganhar ,

about the PR:
I closed the PR because we found that with the new normalization if we change the [FontFreeType::DistanceMapSpread](https://github.com/cocos2d/cocos2d-x/blob/d07794052fed5c3edc29d4a60f99399d49271515/cocos/2d/CCFontFreeType.cpp#L39) to have more spread you only get more black pixels. 
This is the texture with spread 10 instead of the default value of 3:

captura de pantalla 2017-04-13 a las 11 12 48

As you can see there is no new information in the texture compared with spread 3 (check it in the description of the PR), just more black pixels.

We need to think a better way to do what we did in commit 504b26834b8c354737b3f2349312c1663b3af325. The other commit (ef6c5feb63d76bce3747c3b3582d8d6168a8b795) is fine.

About the test cases:
I'm not sure what would be a suitable test case for that. The "cuts" were only noticiable on some fonts. Maybe a test that shows the distance map generated texture to show that you are not getting hard edges and that the glyphs are placed correctly can be a solution. What do you think?

dumganhar:
> About the test cases:
> I'm not sure what would be a suitable test case for that. The "cuts" were only noticiable on some fonts. 

If the specific ttf font is not large, you could add the new fonts to `tests/cpp-tests/Resources/fonts` and write a test case to use that font to show the `cuts`.

> Maybe a test that shows the distance map generated texture to show that you are not getting hard edges and that the glyphs are placed correctly can be a solution. What do you think?

Just do it, if the test case could easily show the difference before and after applying your patch, it's ok.

PR #17669, opened 2017-04-09
before draw event

newnon:


zilongshanren:
The JSB and Lua bindings are not taken into consideration in this PR.

newnon:
i'm not sure about bindings

minggo:
> The JSB and Lua bindings are not taken into consideration in this PR.

@zilongshanren i think lua/jsb just need to listen the corresponding events. Or what should need to do for lua and JSB?

zilongshanren:
@minggo 
I mean if you want to use the before draw event in Lua or Js app, there is no way to do.

minggo:
I think lua/js can just register event listener and check the event type as others.

PR #17677, opened 2017-04-10
[ci skip]do not use git command

minggo:
the script may be used in a folder that is not a git repo


dumganhar (approved):

dumganhar:
Great.
Should this patch also be in v3.15 branch?

minggo:
I though it only exists in v3. I will synchronize to v3.15 too.

PR #17679, opened 2017-04-10
do not use git command

minggo:
the script may be used in a folder that is not a git repo


minggo:
synchronize https://github.com/cocos2d/cocos2d-x/pull/17677 to v3.15

Issue #17412, opened 2017-02-26
crtdbg " Detected memory leaks!"  fix patch


ka7713:
- cocos2d-x version:  3.14.1
- devices test on:  win32(windows10)
- developing environments
   - NDK version:
   - Xcode version:
   - VS version:  2015
   - browser type and version:

Steps to Reproduce:

1.  cpp-empty-test, cpp-tests

```
int WINAPI _tWinMain(HINSTANCE hInstance,
					   HINSTANCE hPrevInstance,
                       LPTSTR    lpCmdLine,
                       int       nCmdShow)
{
    _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
    
    UNREFERENCED_PARAMETER(hPrevInstance);
    UNREFERENCED_PARAMETER(lpCmdLine);

    AppDelegate app;
    Application::getInstance()->run();

    return 0;
}
```

```
Detected memory leaks!
Dumping objects ->
{211136} normal block at 0x12F2E9D8, 4 bytes long.
 Data: <    > 00 00 00 00 
.....................
9613 lines
.....................
{89196} normal block at 0x048E94C0, 44 bytes long.
 Data: <       ?p   X   > E4 EE BF 05 00 00 80 3F 70 EB 8D 08 58 9E 8E 04 
Object dump complete.
```

2. patch
```
diff -x ''\''*.h'\''' -x ''\''*.cpp'\''' -ur D:\download\cocos2d-x-3.14.1\cocos2d-x-3.14.1/cocos/base/CCDirector.cpp D:\work\cocos2d-x-3.14.1/cocos/base/CCDirector.cpp
--- D:\download\cocos2d-x-3.14.1\cocos2d-x-3.14.1/cocos/base/CCDirector.cpp	2017-01-18 10:57:42.000000000 +0900
+++ D:\work\cocos2d-x-3.14.1/cocos/base/CCDirector.cpp	2017-02-26 15:14:29.666814100 +0900
@@ -62,6 +62,12 @@
 #include "base/CCAsyncTaskPool.h"
 #include "platform/CCApplication.h"
 
+#include "cocostudio/CCTransformHelp.h"
+#include "base/ObjectFactory.h"
+#include "3d/CCSprite3DMaterial.h"
+#include "3d/CCSprite3D.h"
+#include "extensions/Particle3D/PU/CCPUMaterialManager.h"
+
[...]

PR #17684, opened 2017-04-10
Possible memory leaks fix -- patch from issue #17412

Rubonnek:
Some apparent memory leaks were found on issue #17412.

I'm not quite familiar with the code that's being modified -- I just applied the patches created by @ka7713 and created the pull request since the patches do seem to fix a memory leak. Further review is needed.

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -62,11 +62,9 @@ THE SOFTWARE.
 #include "base/CCAsyncTaskPool.h"
 #include "platform/CCApplication.h"
 
-#include "editor-support/cocostudio/CCTransformHelp.h"
 #include "base/ObjectFactory.h"
 #include "3d/CCSprite3DMaterial.h"
 #include "3d/CCSprite3D.h"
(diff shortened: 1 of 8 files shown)


cocos/base/CCDirector.cpp
```diff
@@ -210,6 +216,13 @@ Director::~Director(void)
     
     Configuration::destroyInstance();
 
+    Sprite3DMaterial::releaseBuiltInMaterial();
+
+    cocostudio::TransformHelp::release();
+
+    ObjectFactory::destroyInstance();
```
minggo, line 224: cocos studio is not necessary by the game, so it may not exist.

Rubonnek: Added a pointer check in TransformHelp::release in the latest commit.

minggo: I mean developers may remove cocos studio/particle3d from project building, so `Director` can not be so tightly with these modules.

Rubonnek: Got it, I removed them from `Director` in the latest commit. Let me know if there's anything else that needs to be done -- I'll work on it.

PR #17686, opened 2017-04-11
fixed #17685: [android] Audio in game couldn't be mute while a ring or a call is coming

### LINKED ISSUE #17685: [android] Audio in game couldn't be mute while a ring or a call is coming
Reported at http://punchbox.info:3000/issues/26198
- cocos2d-x version: v3.14.1
- devices test on: Only appears on some specific devices like XiaoMi 2S
- developing environments
   - NDK version: r10e
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:

1. cpp-tests/NewAudioEngineTest, play a audio by loop
2. Go to background , make a system `count-down` clock
3. Go to foreground, wait for the clock appear
4. Clock comes but with the background music in game playing

dumganhar:


cocos/audio/android/UrlAudioPlayer.h
```diff
@@ -68,6 +68,8 @@ class UrlAudioPlayer : public IAudioPlayer
 
     virtual float getVolume() const override;
 
+    virtual void setFocusLost(bool isFocusLost) override;
```
minggo: setFocusLost -> setFocus?

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
```diff
@@ -75,6 +78,66 @@ public Cocos2dxGLSurfaceView getGLSurfaceView(){
...
+                            public void run() {
+                                nativeOnAudioFocusChange(AUDIOFOCUS_GAIN);
+                                Cocos2dxHelper.setAudioFocusLost(false);
+                            }
+                        });
+                    }
+                }
+            };
+
+
```
minggo: Can put it into a new java file? Cocos2dxActivity is very complex now, we should put different features into different files.

dumganhar:
@minggo , DONE.

minggo (approved):

PR #17697, opened 2017-04-13
Avoid of type overflow while list size calculation.

sbrednikhin:
If _items container is empty `(length - 1) * _itemsMargin` will be large positive value. 
Using `size_t` causes type overflow and large positive sizes.
Using int will cause negative size values.
Direct check looks like the best way.


drelaptop:
you said is right, overflow occurs when the container is empty. but I can't sure we have enough reason to modify it, because this situation didn't case any issues. what do you think ? @sbrednikhin

sbrednikhin:
It might look like bad surprise while debugging.
First time I saw it, I thought this is a bug, and spent some time to read code carefully and understand it. Just want to safe some time for others.

drelaptop:
I can't decide which is better, what do you think?  @minggo

minggo:
I think @sbrednikhin makes the codes more clear.

drelaptop (changes requested):
@sbrednikhin Sorry for a long response from the time you make PR, now it has conflicts with the latest code, will you solve it? after that PR will be merged.

sbrednikhin:
It's ok. Everything has being resolved.

minggo:
i will merge it after passing travis

drelaptop:
this PR can be merged @minggo

Issue #17704, opened 2017-04-14
[Android] Issue with backgrounding while touch is in progress


saumyamukul:
- cocos2d-x version: 3.15
- devices test on: OnePlus 2, 5.1.1
- developing environments
   - NDK version: r10e
   - Xcode version: N/A
   - VS version: N/A
   - browser type and version: N/A

Steps to Reproduce:

1. Start cpp-tests and navigate to "Node Scene"->"Test pushScene"->"replaceScene"
2. Press down on "replaceScene"
3. Start task switcher without releasing the touch
4. Release touch
5. Foreground app, notice labels blacked out

![ezgif com-video-to-gif](https://cloud.githubusercontent.com/assets/11037660/25054921/18212b74-211d-11e7-986f-90eed76ec162.gif)

**_OpenGL log when you release the touch_**

![screen shot 2017-04-14 at 12 13 38 pm](https://cloud.githubusercontent.com/assets/11037660/25055077/fb21fe1c-211d-11e7-9fe3-8838aa70fd22.png)

**_Spew when you foreground_**

![screen shot 2017-04-14 at 12 13 53 pm](https://cloud.githubusercontent.com/assets/11037660/25055078/fb23d980-211d-11e7-87bc-652bc67a6f06.png)

minggo:
I tested on xiaomi4, can not reproduce it, and i tested on huawei P9, task switcher can not take effect when touching `replaceScene`.

saumyamukul:
You're right, some devices don't allow task switching when a touch is in progress.
But I was able to repro with ease on the following device (They allow task switching while touching the screen)
Galaxy S4
Nexus 4

You can also get it to show on some devices by timing your touch just right. Tap on the task switcher and then immediately press "replaceScene". This is slightly harder to repro, but I was able to see this on:
Galaxy S5
LG G5
Galaxy Note 2

Could not repro on Galaxy S8 and Nexus 6P

minggo:
@saumyamukul  It is device related, it is a very special operation, and it will make the codes complex and slow down the performance, so i will not merge the PR.

PR #17705, opened 2017-04-14
[Android] Fix issue with backgrounding while touch is in progress

saumyamukul:
https://github.com/cocos2d/cocos2d-x/issues/17704
Cancel any touches which are in progress when you background the app


minggo:
@saumyamukul thanks for your work. According to the comment of the issue, i will close it. But this PR is good for other developers who need it. Thanks.

saumyamukul:
@minggo Sounds good 👍

minggo:
@saumyamukul thanks for understanding.

PR #17706, opened 2017-04-14
Fix reschedule issue with same key

Dimon4eg:
**The issue:**
* create timer with some repeat count
Scheduler::schedule(callback, target, interval, repeat, delay, paused, key);
* wait for last repeat executed
* in timer callback reschedule it again with the same key
* as result timer is broken and will never be executed

**Reason:**
When we reschedule timer again from callback the scheduler just updates interval value:
`timer->setInterval(interval);`

After callback executed timer will be canceled:
```
if (!_runForever && _timesExecuted > _repeat)
{    //unschedule timer
    cancel();
    return;
}
```

**Fix:**
Scheduler will update interval only for not exhausted timers (still running).
In our case, it will create new internal timer object and old one will be canceled after callback.

The same fix I did for cocos2d-html5 here [https://github.com/cocos2d/cocos2d-html5/pull/3478](https://github.com/cocos2d/cocos2d-html5/pull/3478)


minggo:
It seems we should define what should do when reschedule a callback with repeat time. This PR can resolve the issue if you reschedule at the end of the last repeat, but what should happen if reschedule before the last repeat? Should we add the repeat time instead of just change the interval?

Dimon4eg:
Yes, good point.
If reschedule before then we need change repeat count also I think.
I'll do it in this PR.

Dimon4eg:
@minggo I have question about updating repeat count also.
Example:
created timer with 10 repeat count
after 3 repeat we recreate it again with 10 repeat
How many it should repeat after?
10 or 7?

I can easy set all timer parameters using:
`Timer::setupTimerWithInterval(float seconds, unsigned int repeat, float delay)`

And then after recreate timer will be like a new timer, will start again from beginning.
So, in example above it will repeat 10 times at the end.
For me it's predictable flow. I vote for this ;)

minggo:
> created timer with 10 repeat count
after 3 repeat we recreate it again with 10 repeat
How many it should repeat after?
10 or 7?

Why not 17?

Dimon4eg:
1) the behavior will be not expected (you set 10 repeat)
2) to do that we will add previous repeat counter with new one but we don't add intervals - we just update it

minggo:
If so, then i think 10 is more reasonable.

PR #17716, opened 2017-04-17
Fix AudioEngine crash on Android

pwang08:
If call AudioEngine::stop, it'll remove pair from AudioEngine::_audioIDInfoMap with the given audioId, then when use it again on setPlayEventCallback, the AudioInfo with the given audioId is missing, then *AudioEngine::_audioIDInfoMap[id].filePath will result in null pointer exception.


dumganhar:
@pwang08 , could you please add some pseudocode about how to reproducing this issue?
It's better to add a new test case to NewAudioEngineTest.

dumganhar:
I guess this issue will only appear when AudioEngine::stop and audio while it's meanwhile played over.  I review your patch, it seems that it will not fix this issue entirely. I will try to find out a better approach and let you know.

dumganhar:
@pwang08 , I submitted a new pull request for this issue:
https://github.com/cocos2d/cocos2d-x/pull/17748
Since it's hard for me to reproduce, could you please revert your patch then apply my patch to your code, and test it again? Thanks.

pwang08:
I don't think the issue will only appear when AudioEngine::stop and audio while it's meanwhile played over. 

When the player stops an audio, the code below will be invoked, which will call AudioEngine::remove():
```
void AudioEngine::stop(int audioID)
{
     auto it = _audioIDInfoMap.find(audioID);
     if (it != _audioIDInfoMap.end()){
          _audioEngineImpl->stop(audioID);

          remove(audioID);
      }
}
```

And `_audioEngineImpl->stop(audioID);` will trigger the track's state changing to STOPPED, in some occassions. if `_callerThreadUtils->performFunctionInCallerThread(func);` in `PcmAudioPlayer::prepare` is called, which will invoke the code below in next frame:
```
player->setPlayEventCallback([this, player](IAudioPlayer::State state){

                if (state != IAudioPlayer::State::OVER && state != IAudioPlayer::State::STOPPED)
                {
                    ALOGV("Ignore state: %d", static_cast<int>(state));
                    return;
                }

                int id = player->getId();
                std::string filePath = *AudioEngine::_audioIDInfoMap[id].filePath;

                ALOGV("Removing player id=%d, state:%d", id, (int)state);

                AudioEngine::remove(id);
                _audioPlayers.erase(id);

                auto iter = _callbackMap.find(id);
                if (iter != _callbackMap.end())
                {
                    if (state == IAudioPlayer::State::OVER)
                    {
                        iter->second(id, filePath);
                    }
                    _callbackMap.erase(iter);
                }
});
```

This means playEventCallBack is invoked after `remove(audioID);` in `AudioEngine::stop(int audioID)`, which has removed `AudioEngine::_audioIDInfoMap` with the given audioId.

So it's a very common case, I hope I have described this issue clearly.

dumganhar:
@pwang08 

> This means playEventCallBack is invoked after remove(audioID); in AudioEngine::stop(int audioID), which has removed AudioEngine::_audioIDInfoMap with the given audioId.
>
> So it's a very common case, I hope I have described this issue clearly.

Yes, you're right, it's because `setPlayEventCallback` is invoked twice, the first time is by `State::STOPPED` and the second time is by `State::OVER`. But in theory, if State::STOPPED is triggered, `State::OVER` should not be fired. Therefore, this issue will only be triggered in this condition. It's not a command case, otherwise, NewAudioEngineTest should reproduce this issue, but the fact is not.
Although I might be wrong, could you help to test my patch to see whether it works. Thanks.

pwang08:
Hi, the reason is not `setPlayEventCallback` is invoked twice.

The reason is that 'std::string filePath = *AudioEngine::_audioIDInfoMap[id].filePath;' is executed after `AudioEngine::remove():` in `AudioEngine::stop`.

dumganhar:
I know that, but why `setPlayEventCallback` is still be triggered after AudioEngine::stop? 
It's because `State::OVER` is triggered after `AudioEngine::stop` is invoked and the `id` was removed.
`setPlayEventCallback` with `State::OVER` event should not be triggered. That's the root cause of the crash.

pwang08:
Hi, James,

I don't know whether `void AudioMixerController::mixOneFrame()` is executed in sub-thread?

dumganhar:
@pwang08 , yes, mixOneFrame is executed in opensles audio thread which is a subthread.
```c++
bool PcmAudioService::enqueue()
{
    if (_controller->hasPlayingTacks())
    {
        if (_controller->isPaused())
        {
            SLresult r = (*_bufferQueueItf)->Enqueue(_bufferQueueItf, __silenceData.data(), __silenceData.size());
            SL_RETURN_VAL_IF_FAILED(r, false, "enqueue silent data failed!");
        }
        else
        {
            _controller->mixOneFrame();

            auto current = _controller->current();
            ALOG_ASSERT(current != nullptr, "current buffer is nullptr ...");
            SLresult r = (*_bufferQueueItf)->Enqueue(_bufferQueueItf, current->buf, current->size);
            SL_RETURN_VAL_IF_FAILED(r, false, "enqueue failed!");
        }
    }
    else
    {
        SLresult r = (*_bufferQueueItf)->Enqueue(_bufferQueueItf, __silenceData.data(), __silenceData.size());
        SL_RETURN_VAL_IF_FAILED(r, false, "enqueue silent data failed!");
    }

    return true;
}

void PcmAudioService::bqFetchBufferCallback(SLAndroidSimpleBufferQueueItf bq)
{
    // FIXME: PcmAudioService instance may be destroyed, we need to find a way to wait...
    // It's in sub thread
    enqueue();
}
```

pwang08:
If it's executed in a sub-thread. It means the code below will be executed in a sub-thread after the state of the track was changed to STOPPED. Then in some cases, `setPlayEventCallback` will be triggered in next frame.

```
else if (state == Track::State::STOPPED)
        {
            if (track->isInitialized())
            {
                _mixer->deleteTrackName(track->getName());
            }
            else
            {
                ALOGV("Track (%p) hasn't been initialized yet!", track);
            }
            tracksToRemove.push_back(track);
        }
```

dumganhar:
User's code includes the finish callback will always be in Cocos thread. We check the thread id while invoking the callback and if it's in sub-thread, we will `performFunctionInCallerThread` it to caller's thread (cocos thread). Therefore, it may make a queue like `STOPPED`, `OVER`, `DESTROYED` although `STOPPED` is triggered and `OVER` should not be called.
```c++
    _track->onStateChanged = [this, callerThreadId](Track::State state) {
        // It maybe in sub thread
        auto func = [this, state](){
            // It's in caller's thread
            if (state == Track::State::OVER && _track->getPrevState() != Track::State::STOPPED) // THIS IS THE FIX, WE NEED TO CHECK THE PREVIOUS STATE OF TRACK.
            {
                if (_playEventCallback != nullptr)
                {
                    _playEventCallback(State::OVER);
                }
            }
            else if (state == Track::State::STOPPED)
            {
                if (_playEventCallback != nullptr)
                {
                    _playEventCallback(State::STOPPED);
                }
            }
            else if (state == Track::State::DESTROYED)
            {
                delete this;
            }
        };

        if (callerThreadId == std::this_thread::get_id())
        { // onStateChanged(Track::State::STOPPED) is in caller's (Cocos's) thread.
            func();
        }
        else
        { // onStateChanged(Track::State::OVER) or onStateChanged(Track::State::DESTROYED) are in audio mixing thread.
            _callerThreadUtils->performFunctionInCallerThread(func);
        }
    };
```

pwang08:
Yeah, I know Users code includes the finish callback will always in Cocos thread. But the key is the timing, if `_callerThreadUtils->performFunctionInCallerThread(func);` is executed, it means the finish callback is invoked in **next frame**. On that time, `AudioEngine::Remove` has been executed.

dumganhar:
Yes, you're right. That's the key point.

dumganhar:
@pwang08 , did you try my solution https://github.com/cocos2d/cocos2d-x/pull/17748 ?
Did it work for you?

pwang08:
Hi, james, I didn't try currently, because we are busy developing our app. Maybe I'll try on this weekend, I'll tell you if I have tried.

Now that you have got the key point, I don't think your PR will work from the logical view.

minggo:
@pwang08  any update of your review? Thanks.

dumganhar:
Merge this PR at #17923
Closed this one. Thanks @pwang08 .

PR #17727, opened 2017-04-19
Changing AsyncTaskPool/Scheduler to better leverage move semantics

DavidDeSimone:
Changing AsyncTaskPool/Scheduler to better leverage move semantics. This will allow us to avoid several lambda copy constructor calls, and improve performance in the engine around some of our multi-threading APIs. 

This change should be backwards compatible, and not break current API*. This change should not introduce any additional net copy constructor calls, and should overall reduce the number of copy constructor calls for lambda functions in the engine. 

\* Before AsyncTaskPool::enqueue was based off a generic parameter T, however we were assuming that T implemented operator() accepting no arguments. Internally in the engine, the parameter T could be trivially switched to std::function<void()>. I believe that any users using this code directly should continue to work, unless their lambda's had unused return values for some strange reason.

### DIFF
--- a/cocos/base/CCAsyncTaskPool.h
+++ b/cocos/base/CCAsyncTaskPool.h
@@ -89,11 +89,19 @@ public:
      * @param type task type is io task, network task or others, each type of task has a thread to deal with it.
      * @param callback callback when the task is finished. The callback is called in the main thread instead of task thread.
      * @param callbackParam parameter used by the callback.
-     * @param f task can be lambda function.
+     * @param task: task can be lambda function to be performed off thread.
      * @lua NA
      */
-    template<class F>
-    inline void enqueue(TaskType type, const TaskCallBack& callback, void* callbackParam, F&& f);
+    void enqueue(TaskType type, TaskCallBack callback, void* callbackParam, std::function<void()> task);
+
+    /**
+    * Enqueue a asynchronous task.
+    *
+    * @param type task type is io task, network task or others, each type of task has a thread to deal with it.
+    * @param task: task can be lambda function to be performed off thread.
+    * @lua NA
+    */
+    void enqueue(AsyncTaskPool::TaskType type, std::function<void()> task);
     
 CC_CONSTRUCTOR_ACCESS:
     AsyncTaskPool();
@@ -132,7 +140,7 @@ protected:
                                           }
                                           
                                           task();
-                                          Director::getInstance()->getScheduler()->performFunctionInCocosThread([&, callback]{ callback.callback(callback.callbackParam); });
+                                          Director::getInstance()->getScheduler()->performFunctionInCocosThread(std::bind(callback.callback, callback.callbackParam));
                                       }
                                   }
                                   );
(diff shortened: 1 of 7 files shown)


DavidDeSimone:
If this PR is merged, I will submit a PR to https://github.com/cocos2d/bindings-generator, to have the code auto-generated by the binding generator better leverage move semantics.

cocos/base/CCAsyncTaskPool.h
```diff
@@ -89,11 +89,19 @@ class CC_DLL AsyncTaskPool
      * @param type task type is io task, network task or others, each type of task has a thread to deal with it.
      * @param callback callback when the task is finished. The callback is called in the main thread instead of task thread.
      * @param callbackParam parameter used by the callback.
-     * @param f task can be lambda function.
+     * @param task: task can be lambda function to be performed off thread.
      * @lua NA
      */
-    template<class F>
-    inline void enqueue(TaskType type, const TaskCallBack& callback, void* callbackParam, F&& f);
+    void enqueue(TaskType type, TaskCallBack callback, void* callbackParam, std::function<void()> task);
```
minggo, line 95: Should remove `template<class F>`, since it is not used?

minggo: And why not use reference here, such as
```c++
void enqueue(TaskType type, TaskCallBack callback, void* callbackParam, std::function<void()> &task);
```

cocos/base/CCAsyncTaskPool.h
```diff
@@ -208,14 +215,17 @@ inline void AsyncTaskPool::stopTasks(TaskType type)
     threadTask.clear();
 }
 
-template<class F>
-inline void AsyncTaskPool::enqueue(AsyncTaskPool::TaskType type, const TaskCallBack& callback, void* callbackParam, F&& f)
+inline void AsyncTaskPool::enqueue(AsyncTaskPool::TaskType type, TaskCallBack callback, void* callbackParam, std::function<void()> task)
```
minggo, line 218: Why not use reference here, such as
```c++
inline void AsyncTaskPool::enqueue(AsyncTaskPool::TaskType type, TaskCallBack callback, void* callbackParam, std::function<void()> &task)
```

cocos/base/CCScheduler.cpp
```diff
@@ -803,13 +803,10 @@ void Scheduler::resumeTargets(const std::set<void*>& targetsToResume)
     }
 }
 
-void Scheduler::performFunctionInCocosThread(const std::function<void ()> &function)
+void Scheduler::performFunctionInCocosThread(std::function<void ()> function)
```
minggo, line 806: can use reference here?

cocos/base/CCAsyncTaskPool.h
DavidDeSimone: We should remove the template<class F>.

If we make if an l-value reference, then doing something like

```
AsyncTaskPool::getInstance()->enqueue(0, myFunction, myParams, [] () {

});
```

Won't compile, since the lambda will be a r-value. And we really don't get anything making it an l-value reference over taking it by value.

cocos/base/CCAsyncTaskPool.h
DavidDeSimone: See above comment

cocos/base/CCScheduler.cpp
DavidDeSimone: see above

cocos/base/CCAsyncTaskPool.h
minggo: Thanks for the explanation.
Could you please remove the template, i will merge it after that.

DavidDeSimone:
@minggo Thanks for the feedback. The template<T> should be removed now.

DavidDeSimone:
also as promised, here is my PR to the binding generator to make it better leverage move semantics. https://github.com/cocos2d/bindings-generator/pull/238

minggo:
@DavidDeSimone thanks.

PR #17735, opened 2017-04-20
Avoid crash on large file download.

sbrednikhin:
Downloading large file to memory could cause stack overflow and crash, because old code created buffer `char buf[buflen];` on stack.
New code creates buffer in heap.


dumganhar (approved):

minggo:
[ci rebuild]

PR #17736, opened 2017-04-21
JSB SocketIO.connect supports to pass a CA cert file

dumganhar:


minggo (approved):

pandamicro:
I don't quite understand the official API documentation, it declares there is no parameter for socket.connect which is not as our current API.

https://socket.io/docs/client-api/#socket-connect

dumganhar:
Just add a optional parameter (caFilePath) for SocketIO.connect method，it will not conflict with the  the API of socket io official.

minggo:
Hi guys, please use english. We ask developers using english, we should obey it ourselves.

pandamicro:
I'm ok now if the original C++ SocketIO::connect support such API form, it's just the difference between our implementation and SocketIO official documentation makes me confused, I think it should be fixed when we upgrade our C++ SocketIO implementation to the latest version

pandamicro (approved):

PR #17756, opened 2017-04-26
Memory leak in setFontAtlas, if set equal font in Label.

perminovVS:
setCharMap, setBMFontFilePath and setTTFConfigInternal use getFontAtlas.
getFontAtlas retain font atlas always.
setFontAtlas not release, if atlas equal current.

For example call Label::setTTFConfig for equal ttfConfig.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -584,6 +584,10 @@ void Label::setFontAtlas(FontAtlas* atlas,bool distanceFieldEnabled /* = false *
 
     if (atlas == _fontAtlas)
     {
+        if (_fontAtlas)
+        {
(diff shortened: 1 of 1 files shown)


zilongshanren (approved):

zilongshanren:
thanks, merged.

minggo:
the logic is strange, if you run `cpp-tests -> LabelNewTest`, it will crash in the test case `LabelTTFCJKWrappingTest`.

I think `FontAtlasCache::getFontAtlasXXX` should not invoke retain, the invoker should do retain if needed.

minggo:
I sent https://github.com/cocos2d/cocos2d-x/pull/18117 to fix it and make the logic look more correctly.

PR #17762, opened 2017-04-27
Added function for calculating md5 hash from Data.

perminovVS:
getFileMD5Hash now use getDataMD5Hash.

For example, AssetsManagerEx VerifyCallback.
Get Data from file, check equal size, if equal to get md5 hash.

### DIFF
--- a/cocos/base/ccUtils.cpp
+++ b/cocos/base/ccUtils.cpp
@@ -411,17 +411,27 @@ Node* findChild(Node* levelRoot, int tag)
 
 std::string getFileMD5Hash(const std::string &filename)
 {
+    Data data;
+    FileUtils::getInstance()->getContents(filename, &data);
+
+    return getDataMD5Hash(data);
+}
+
+std::string getDataMD5Hash(const Data &data)
+{
     static const unsigned int MD5_DIGEST_LENGTH = 16;
 
-    Data d;
-    FileUtils::getInstance()->getContents(filename, &d);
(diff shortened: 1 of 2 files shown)


cocos/base/ccUtils.cpp
```diff
@@ -411,17 +411,27 @@ Node* findChild(Node* levelRoot, int tag)
...
+    return getDataMD5Hash(data);
+}
+
+std::string getDataMD5Hash(const Data &data)
+{
     static const unsigned int MD5_DIGEST_LENGTH = 16;
 
-    Data d;
-    FileUtils::getInstance()->getContents(filename, &d);
+    if (data.isNull() || data.getSize() == 0)
```
dumganhar: `data.getSize()` isn't necessary here since `data.isNull()` will check the size.
```c++
bool Data::isNull() const
{
    return (_bytes == nullptr || _size == 0);
}
```

perminovVS:
[ci rebuild]

minggo:
Please fix the compiling error

```
In file included from /home/travis/build/cocos2d/cocos2d-x/cocos/2d/CCGrid.cpp:29:0:
/home/travis/build/cocos2d/cocos2d-x/cocos/base/ccUtils.h:180:45: error: ‘Data’ does not name a type
     CC_DLL std::string getDataMD5Hash(const Data &data);
```

perminovVS:
@minggo it done.

PR #17768, opened 2017-05-01
Fix space wrapping in CCLabelTextFormatter

ahlwong:
Previously, if you insert spaces at the end of a long string, the line would not wrap. The fix introduces correct wrapping for spaces.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -384,6 +384,7 @@ Label::Label(TextHAlignment hAlignment /* = TextHAlignment::LEFT */,
 , _boldEnabled(false)
 , _underlineNode(nullptr)
 , _strikethroughEnabled(false)
+, _trimLineSpaces(true)
 {
     setAnchorPoint(Vec2::ANCHOR_MIDDLE);
     reset();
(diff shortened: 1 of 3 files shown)


zilongshanren:
@ahlwong 
thanks for the patch.

Could you provide a test to reproduce this issue?

Just add it to the `cpp-tests/ Node : Label Test New` section.

The cpp file name is CCLabelTestNew.cpp.

ahlwong:
I noticed there is also incorrect alignment when a line break occurs on a space when the label is right aligned. I will add to the pull request and add tests in the next few days when i port the fixed code form my project.

zilongshanren:
@ahlwong 
Thanks, after you finishing the test, please let us know

cocos/2d/CCLabel.cpp
```diff
@@ -1847,6 +1850,20 @@ float Label::getAdditionalKerning() const
...
 }
 
+void Label::setTrimLineSpacesEnabled(bool enabled)
+{
+    if (_trimLineSpaces != enabled) {
+        _trimLineSpaces = enabled;
+
+        _contentDirty = true;
+    }
+}
```
minggo, line 1860: Any effect when `_trimLineSpaces ` is true? What's this variable used for?

drelaptop:
apply the patch, I find __Right Alignment__ have a different behavior with __Left Alignment__ when a line have spaces on sides. both have different way to deal with __Spaces__, so this PR isn't perfect, please review and test in detail @ahlwong

ahlwong:
Apologies for the lapse in response. This change is not important for the Cocos2d-x framework and its users. It mainly came out of my work on building a custom textfield where I noticed that the label would collapse adjacent spaces.

drelaptop:
Anyway, thanks for your warm contribution

PR #17771, opened 2017-05-03
Fixes TMX map parsing problems

MartinLutherSu:
Now, TMX map can parse rotation of object correctly.

### DIFF
--- a/docs/CODING_STYLE.md
+++ b/docs/CODING_STYLE.md
@@ -2453,7 +2453,7 @@ if (condition) {
 }
 ```
 
-## Loops and Switch Statements
+## Loops and Switch Statement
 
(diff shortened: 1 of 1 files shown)


minggo:
could you please add a test case?

xbpiao:
@MartinLutherSu  Great job!

v1993:
@minggo Why still isn't merged?

minggo:
Because i don't know how to reproduce it.
@drelaptop could you please take a look too?

drelaptop:
@v1993 not test case or test descriptions to tell us this is a right bug-fix

I will try to test it later

drelaptop:
link https://github.com/cocos2d/cocos2d-x/pull/11540 , which added a simple `rotation` property parse

kangzixiang (approved):
Nice

Issue #17767, opened 2017-05-01
Commit #b017934 for OpenGLES3 breaks multisampling on iOS


ahlwong:
- cocos2d-x version: 3.15
- devices test on: iPhone 7 plus

Steps to Reproduce:
Commit #b017934 made changes to cocos/platform/ios/CCES2Renderer-ios.m.
Now running this following code creates a black screen with swap buffer errors.
```
CCEAGLView *eaglView = [CCEAGLView viewWithFrame: [self.window bounds]
                                     pixelFormat: (__bridge NSString *)cocos2d::GLViewImpl::_pixelFormat
                                     depthFormat: cocos2d::GLViewImpl::_depthFormat
                              preserveBackbuffer: NO
                                      sharegroup: nil
                                   multiSampling: YES // Setting multisampling to YES
                                 numberOfSamples: 4]; // Setting numberOfSamples to 4
```
More info here:
https://github.com/cocos2d/cocos2d-x/commit/b0179340be4a2dc956c8ada5c1d0ee9ec34a2d0e

minggo:
Refer to [the doc](https://developer.apple.com/library/content/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/WorkingwithEAGLContexts/WorkingwithEAGLContexts.html#//apple_ref/doc/uid/TP40008793-CH103-SW4), currently cocos2d-x doesn't support multiple sample correctly. cocos2d- x doesn't

* clear sample buffer at the beginning of a frame
* doesn't resolve the contents from the multisampling buffer into the resolve buffer after submitting your drawing commands

And these operations are iOS depended, it will make ugly if we want to support it. 

@ahlwong did you even use multi sample and find it really take effect? I mean the effect is better.

minggo:
Oh, it seems `CCEAGLView-ios.mm` do the things mentioned above.

minggo:
Can fix it by replacing `glResolveMultisampleFramebufferAPPLE ` with `glBlitFramebuffer ` as [the doc](https://developer.apple.com/library/content/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/AdoptingOpenGLES3/AdoptingOpenGLES3.html) says. But `glBlitFramebuffer` is an GLES3 function. Can not link it with GLES2.0.

minggo:
Fixed in https://github.com/cocos2d/cocos2d-x/pull/17778. Just not use GLES3 context if using multisampling.

PR #17778, opened 2017-05-04
do not use ES3 context if using multi sample

minggo:
multi sample in ES3 needs to use GLES3 function


minggo:
related issue: https://github.com/cocos2d/cocos2d-x/issues/17767

PR #17787, opened 2017-05-05
Allow extensions to work out of the box by default on Win32 C++ projects

Rubonnek:
Adding `#include "cocos-ext.h"` in the `HelloWorldScene.cpp` of a new C++ project and compiling the Win32 project will yield: 

````
Cannot open include file: 'extensions/ExtensionExport.h': No such file or directory
````

This patch fixes that.


Rubonnek:
@minggo Thanks.

Issue #17795, opened 2017-05-08
[Android] Edit box text is larger and mis-aligned


saumyamukul:
- cocos2d-x version: 3.15
- devices test on: LG G5, Galaxy S8, Nexus 6P
- developing environments
   - NDK version: r10e
   - Xcode version: N/A
   - VS version: N/A
   - browser type and version: N/A

Steps to Reproduce:

1. Start cpp-tests and navigate to "Node:UI"->"GUI Dynamic Create Test"->"EditBox Test"
2. Tap the first edit box. Notice how the font size increases when the edit box is focused.
3. Tap the "Single Line" button next to the first edit box. The text will center align. Now tap the edit box. Notice how the edit box text is now top aligned.
4. Other edit boxes display similar behavior

![ezgif com-video-to-gif 6](https://cloud.githubusercontent.com/assets/11037660/25829073/4c372ec0-3411-11e7-91d9-2ca88d7da08b.gif)

PR #17796, opened 2017-05-08
[Android] Fix edit box font size and alignment

saumyamukul:
https://github.com/cocos2d/cocos2d-x/issues/17795

Edit box alignment wasn't resetting the previous value before setting the new one.
Also fixed font size.

Tested on:
Galaxy S8
Nexus 6P
LG G5

ezgif com-video-to-gif 7


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
```diff
@@ -268,8 +253,8 @@ public void run() {
                     if (fontSize >= 0){
                         float density =  mCocos2dxActivity.getResources().getDisplayMetrics().density;
```
zilongshanren, line 254: The `density` variable is not needed now.

zilongshanren:
@saumyamukul 
thanks, this PR works well except for one little issue, I have commented it above.

zilongshanren (changes requested):

PR #17807, opened 2017-05-11
[Windows]Fix edit box spacing and alignment

saumyamukul:
Also prevent a crash inside setText due to invalid cast


minggo:
@saumyamukul  `100644 → 100755` please don't change the file attributes.

minggo:
@saumyamukul please don't change the file attribute from `100644` to `100755`. You can click the `FileChanges` to check that, the file attribute is changed. And source code files do need execution permission.

saumyamukul:
@minggo Thats for the feedback. I got caught up with other work. I'll fix that soon

minggo:
thanks

saumyamukul:
File attributes have been reset!

PR #17812, opened 2017-05-13
[WinRT] Fix removeDirectory unicode.

perminovVS:
CCFileUtilsWinRT::removeDirectory need ut8 string.
temp is wstring(unicode).


stammen:
Looks good. Thanks!

PR #17813, opened 2017-05-13
[Win32,WinRT] Fix listFiles and listFilesRecursively use unicode.

perminovVS:
1. Incorrect convertation to unicode
if (length != fullpath.size()) correct only ASCII char.
2. For convert unicode exist api, to avoid errors.
3. Equal first char on '.' - error
Need check full name file on "." and "..", otherwise exclude file and
folder begin '.'

Now not use macro UNICODE, for easy use unicode convert api(Otherwise,
need to connect different versions WinRT and Win32 include).

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -34,7 +34,6 @@ THE SOFTWARE.
 //#include "base/ccUtils.h"
 
 #include "tinyxml2/tinyxml2.h"
-#include "tinydir/tinydir.h"
 #ifdef MINIZIP_FROM_SYSTEM
 #include <minizip/unzip.h>
 #else // from our embedded sources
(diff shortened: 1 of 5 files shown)


cocos/platform/CCFileUtils.cpp
```diff
@@ -1169,22 +1169,17 @@ void FileUtils::getFileSize(const std::string &filepath, std::function<void(long
 
 std::vector<std::string> FileUtils::listFiles(const std::string& dirPath) const
 {
-    std::string fullpath = fullPathForFilename(dirPath);
     std::vector<std::string> files;
+#ifdef UNICODE
```
minggo: it seems only windows serials system define `UNICODE`?

cocos/platform/CCFileUtils.cpp
```diff
@@ -1261,21 +1241,12 @@ void FileUtils::listFilesRecursively(const std::string& dirPath, std::vector<std
...
-                {
-                    filepath.resize(length);
-                    WideCharToMultiByte(CP_UTF8, 0, &path[0], (int)path.size(), &filepath[0], length, NULL, NULL);
-                }
-#else
-                std::string filepath = file.path;
-#endif
-                if (file.name[0] != '.')
+                // Need check full name file on "." and "..", otherwise exclude file and folder begin '.'
+                if (fileName != "." && fileName != "..")
```
minggo: it seems should check the first character instead of `fileName`.

cocos/platform/CCFileUtils.cpp
perminovVS: I see tinydir, it checked _MSC_VER define for use WINAPI.
But this function check UNICODE for use WINAPI , before my change.

cocos/platform/CCFileUtils.cpp
perminovVS: see it commit #13905

cocos/platform/CCFileUtils.cpp
minggo: Yep. You are right.

cocos/platform/CCFileUtils.cpp
minggo: I see your codes here. But i don't quite understand the comment `otherwise exclude file and folder begin '.'`. And should we list hidden folder here?

perminovVS: Folder and Files begins point '.' not hidden, it normal files and folder, it сan not exclude.
.git and .vs examples, exist more other folder and files users.

It function should return all folder and files.
Check '.' should exclude folder '.' - current folder and folder '..' - up folder.

minggo: Yep, i know the meaning. How about just remove the comment, because the codes is clear and i am confused when check the comment.

`.git` and `.vs` is hidden folder in Mac and linux like system.

perminovVS:
ok, I will delete it comments.

minggo:
thanks

perminovVS:
@minggo Done it. And replaced UNICODE to _MSC_VER as it use in tinydir.

minggo:
@perminovVS thanks. I thin we can remove `_MSC_VER` section since windows overrides these functions. What's your opinion?

perminovVS:
@minggo Below there is a section:
`
#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32) || (CC_TARGET_PLATFORM == CC_PLATFORM_WINRT)
// windows os implement should override in platform specific FileUtiles class
`
Сan transfer there.

PR #17814, opened 2017-05-13
[Win32,WinRT] Fix old bag, Manifest not save.

perminovVS:
Manifest::saveToFile not save, if path contains not asii chars.
std::ofstream not use unicode api.

Now use FileUtils api for write file.


minggo:
looks good to me
@pandamicro could you please take a look too? thanks.

PR #17815, opened 2017-05-13
[Win32] Fix AudioDecoder fopen unicode.

perminovVS:
If use fopen to need convert utf8 string FileUtils::getSuitableFOpen.
ov_fopen and mpg123_open use fopen.

### DIFF
--- a/cocos/audio/win32/AudioDecoderMp3.cpp
+++ b/cocos/audio/win32/AudioDecoderMp3.cpp
@@ -91,7 +91,7 @@ namespace cocos2d { namespace experimental {
                 break;
             }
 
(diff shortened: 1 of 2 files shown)


minggo:
Thanks, looks good to me.
@dumganhar could you please take a look too?

dumganhar:
This PR is good for me.
But what's the meaning of `getSuitableFOpen`?? I looked inside of it and found that it just to convert UTF8 string to Multi-bytes(probably GBK) string on windows platform while it return the string passed in on other platforms. I don't know when this interface was added. It's really confused to me.

minggo:
Yep, the function name is bad.

PR #17824, opened 2017-05-16
Allow visit Sprite without camera for render to texture

madrazo:
A fix to allow render to texture like this 
```
        Director::getInstance()->setProjection(Director::Projection::_2D);
        m_renderTarget->beginWithClear(1, 0, 0, 0.2, m_depthClear, 0);
        m_sprite->visit(Director::getInstance()->getRenderer(), Mat4::IDENTITY, 0);
        m_renderTarget->end();
        Director::getInstance()->setProjection(Director::Projection::_3D);
```
otherwise, crashes

PS: I have disabled the default camera in my Scene with

```
    _defaultCamera->setScene(NULL);
    removeChild(_defaultCamera);
    _defaultCamera = NULL;
```

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -1054,7 +1054,10 @@ void Sprite::draw(Renderer *renderer, const Mat4 &transform, uint32_t flags)
     // Don't calculate the culling if the transform was not updated
     auto visitingCamera = Camera::getVisitingCamera();
     auto defaultCamera = Camera::getDefaultCamera();
-    if (visitingCamera == defaultCamera) {
+    if (visitingCamera == NULL) {
+        _insideBounds = true;
+    }
+    else if (visitingCamera == defaultCamera) {
         _insideBounds = ((flags & FLAGS_TRANSFORM_DIRTY) || visitingCamera->isViewProjectionUpdated()) ? renderer->checkVisibility(transform, _contentSize) : _insideBounds;
     }
     else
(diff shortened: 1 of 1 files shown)


cocos/2d/CCSprite.cpp
```diff
@@ -1054,7 +1054,10 @@ void Sprite::draw(Renderer *renderer, const Mat4 &transform, uint32_t flags)
     // Don't calculate the culling if the transform was not updated
     auto visitingCamera = Camera::getVisitingCamera();
     auto defaultCamera = Camera::getDefaultCamera();
-    if (visitingCamera == defaultCamera) {
+    if (visitingCamera == NULL) {
```
minggo: NULL -> nullptr

minggo:
Interesting, why you set `_defaultCamera` to null?

madrazo:
I started using the USER cameras and then  I saw that I didn't need the default camera. 

Also, I think that when rendering to texture, the sprite was showing on the default camera on screen. Maybe I could add an extra USER camera for rendering to texture but this was more straightforward to me.  

Furthermore, I'm not sure about this, but I see that some objects do some computation if there is a default camera on the "visit" or "render" functions when enter this `if (visitingCamera == defaultCamera)`, so I  try to avoid extra computations. 

Thanks!

madrazo:
> Interesting, why you set _defaultCamera to null?

Oh, you mean why I use this line? ` _defaultCamera = NULL; ` 
Just to make sure that it is not used. But yeah, it may work if I don't do that line, or set the default camera to a different value

Issue #17690, opened 2017-04-11
CalculateBillboardTransform always updates


madrazo:
On the following line there is an "if" with lots of conditions (make sure to scroll to the right), but at the end there is a "|| true" which forces to enter the if block anyway.

https://github.com/cocos2d/cocos2d-x/blob/d07794052fed5c3edc29d4a60f99399d49271515/cocos/3d/CCBillBoard.cpp#L160

- cocos2d-x version: 2.14 git
- devices test on: windows
- developing environments
   - NDK version:
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:

1. 
2. 

@super626  seems from this commit https://github.com/cocos2d/cocos2d-x/commit/fc1dbf49fb0e09e2897cb39e7b9857f96b5b24f0

thanks

Edit:
Should "calculateBillboardTransform" be only called when "visibleByCamera" is true?

https://github.com/cocos2d/cocos2d-x/blob/d07794052fed5c3edc29d4a60f99399d49271515/cocos/3d/CCBillBoard.cpp#L113

Edit2:
rotationQuaternion seems to be not used anymore (dead code)
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/3d/CCBillBoard.cpp#L190

PR #17825, opened 2017-05-16
small BillBoard optimizations

madrazo:
-Quick return in visit if it is not visible and has no children.  
-Comment out getNodeToWorldTransform because it is dead code (doesn't do anything)
Reference issue: https://github.com/cocos2d/cocos2d-x/issues/17690


cocos/3d/CCBillBoard.cpp
```diff
@@ -186,8 +191,8 @@ bool BillBoard::calculateBillboardTransform()
         }
         camDir.normalize();
         
-        Quaternion rotationQuaternion;
-        this->getNodeToWorldTransform().getRotation(&rotationQuaternion);
```
minggo: why comment the codes?

madrazo: better remove comments?

PR #17831, opened 2017-05-18
Webview opacity and background transparency

Undume:
1.- Added the possibility of being able to change the opacity of the webviews in android and ios.
2.- Also added the possibilito of being able to set the background fully transparent.

I have done the tests, so u can try it. You can find it in Node:UI->webviewUI. There are 2 new buttons.

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebViewHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebViewHelper.java
@@ -1,10 +1,13 @@
 package org.cocos2dx.lib;
 
 import android.annotation.TargetApi;
+import android.graphics.Color;
 import android.os.Handler;
 import android.os.Looper;
+import android.util.Log;
 import android.util.SparseArray;
 import android.view.View;
+import android.webkit.WebView;
 import android.widget.FrameLayout;
 import android.webkit.WebSettings;
 
@@ -97,9 +100,23 @@ public class Cocos2dxWebViewHelper {
     }
 
     @TargetApi(11)
+    public static void setBackgroundTransparent(final int index) {
+
+        sCocos2dxActivity.runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+                Cocos2dxWebView webView = webViews.get(index);
+                if (webView != null) {
+                    webView.setBackgroundColor(Color.TRANSPARENT);
+                    webView.setLayerType(WebView.LAYER_TYPE_SOFTWARE, null);
+                }
+            }
+        });
+    }
+
+    @TargetApi(11)
     public static void setOpacityWebView(final int index, final float opacity) {
-        
-        
+
         sCocos2dxActivity.runOnUiThread(new Runnable() {
             @Override
             public void run() {
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebViewHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebViewHelper.java
@@ -113,11 +113,7 @@ public class Cocos2dxWebViewHelper {
                         try {
                             Method method = webView.getClass().getMethod("setLayerType",int.class,Paint.class);
                             method.invoke(webView,WebView.LAYER_TYPE_SOFTWARE,null);
-                        } catch (NoSuchMethodException e) {
-                            e.printStackTrace();
-                        } catch (IllegalAccessException e) {
-                            e.printStackTrace();
-                        } catch (InvocationTargetException e) {
+                        } catch (Exception e) {
                             e.printStackTrace();
                         }
                     }
@@ -136,11 +132,7 @@ public class Cocos2dxWebViewHelper {
                         try {
                             Method method = webView.getClass().getMethod("setAlpha",float.class);
                             method.invoke(webView,opacity);
-                        } catch (NoSuchMethodException e) {
-                            e.printStackTrace();
-                        } catch (IllegalAccessException e) {
-                            e.printStackTrace();
-                        } catch (InvocationTargetException e) {
+                        } catch (Exception e) {
                             e.printStackTrace();
                         }
                     }
@@ -162,11 +154,7 @@ public class Cocos2dxWebViewHelper {
                     try {
                         Method method = webView.getClass().getMethod("getAlpha");
                         valueToReturn = method.invoke(webView);
-                    } catch (NoSuchMethodException e) {
-                        e.printStackTrace();
-                    } catch (IllegalAccessException e) {
-                        e.printStackTrace();
-                    } catch (InvocationTargetException e) {
+                    } catch (Exception e) {
                         e.printStackTrace();
                     }
                 }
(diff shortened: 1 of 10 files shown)


zilongshanren (approved):
This PR looks good to me.

cocos/ui/UIWebView.h
```diff
@@ -216,6 +216,20 @@ class CC_GUI_DLL WebView : public cocos2d::ui::Widget {
...
     virtual void setVisible(bool visible) override;
+    /**
+     * SetOpacity of webview.
+     */
+    virtual void setOpacityWebView(float opacity);
+    
+    /**
+     * getOpacity of webview.
+     */
+    virtual float getOpacityWebView();
```
minggo: i think it is better to add const if possible, such as 
`virtual float getOpacityWebView() const`

Undume: True! I forgot it.

minggo: Ok, i will merge it when it is fixed.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebViewHelper.java
```diff
@@ -95,6 +99,60 @@ public void run() {
         });
     }
 
+    @TargetApi(11)
```
minggo: I checked the [doc](https://developer.android.com/reference/android/annotation/TargetApi.html), but i don't quite understand what's the effect of the annotation. Any explanation is appreciated.

minggo: @Undume could you please share it? What's the effect of it?
Thanks.

Undume: It is for suppress the lint error. Libcocos2d has the minsdk at 10. These methods are available from 11 sdk. I could compile it with the minsdk at 10 and I figured out that meanwhile no one use these methods in a device with 10 api wont be error. 

I cannot try it because I cant simulate it and dont have a device with 10 api. 
So I need to know if the cocos2d library would work on a 10-device if dont use these methods. And in case of YES warn the developer that if he want to use these, he must put his game with a minimum sdk of 11. And in case of NO, know how to discriminate it in old devices.

Im learning yet, and I wanted to contribute with Cocos2d-x. So sorry for the lack of skills. Im looking for a solution for this problem yet. If someone can enlighten me I will be grateful.

minggo: But all cocos2d-x should not use API > 10. If you want to do it like that, you should use reflections, such as

```java
if (api > 10)
   use reflection to invoke the API > 10
```

That's mean, there is not effect if running on devices with API 10. So could you please do it like that? Thanks.

Undume: I had done a new push. Im not sure if its what u mean.

minggo: @Undume which API is not exist in level 10? I think you should use java reflection here, or it can not compile if compiling with API 10+.

Undume (commented):
I think that now yes. I finally understood that reflection thing :P learnt something new.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebViewHelper.java
```diff
@@ -95,6 +102,89 @@ public void run() {
...
+                    if (webView != null) {
+                        webView.setBackgroundColor(Color.TRANSPARENT);
+                        try {
+                            Method method = webView.getClass().getMethod("setLayerType",int.class,Paint.class);
+                            method.invoke(webView,WebView.LAYER_TYPE_SOFTWARE,null);
+                        } catch (NoSuchMethodException e) {
+                            e.printStackTrace();
+                        } catch (IllegalAccessException e) {
+                            e.printStackTrace();
+                        } catch (InvocationTargetException e) {
```
minggo: Can we just catch exception in one place if there is not specific operation for exceptions, such as 
```java
catch (Exception e) {
    e.printStackTrace();
}
```

And since which API level has `setLayerType `? I can not find the information [here](https://developer.android.com/reference/android/webkit/WebView.html).

Undume: Unified Exceptions.
screen shot 2017-06-07 at 9 27 46 am
Thats the reason.

minggo: thanks

tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIWebViewTest/UIWebViewTest.cpp
```diff
@@ -144,6 +148,40 @@ bool WebViewTest::init()
...
+        this->addChild(opacityBtn);
+        
+        
+        Button *transparentBgBtn = Button::create("cocosui/animationbuttonnormal.png",
+                                            "cocosui/animationbuttonpressed.png");
+        transparentBgBtn->setTitleText("Transparent BG");
+        transparentBgBtn->setPosition(Vec2(winSize/2) + Vec2( _webView->getContentSize().width/2 +
+                                                             transparentBgBtn->getContentSize().width/2 + 10,-100 ));
+        transparentBgBtn->addClickEventListener([=](Ref*){
+            _webView->setBackgroundTransparent();
```
minggo, line 180: It seems there is not difference on Android, or may be the test case can not show it?

Undume: I think that it depends of the page. I found websites that doesnt make the transparency also. I dont really know. 
`webView.setBackgroundColor(Color.TRANSPARENT);
 webView.setLayerType(WebView.LAYER_TYPE_SOFTWARE, null);`
These are the used methods from Android.

minggo: Got it, thanks. Have you used it in your game?

cocos/ui/UIWebViewImpl-ios.mm
```diff
@@ -152,6 +158,19 @@ - (void)setBounces:(bool)bounces {
...
+    [self.uiWebView setOpaque:NO];
+}
+
+-(float) getOpacityWebView{
+    return self.uiWebView.alpha;
+}
+
+-(void) setBackgroundTransparent{
+    [self.uiWebView setBackgroundColor:[UIColor clearColor]];
+}
```
summerinsects, line 172: Hi there
there is a bug here.
When calling `setBackgroundTransparent` **immediately** after creating `ui::Webview`, this function does not work. because `self.uiWebView` is a delayed creation view, That is to say, `self.uiWebView==nil`.
function `setOpacityWebView` is similar.
On the other hand, `setOpaque:NO` is needed here.
so it should be
```obj-c
-(void) setBackgroundTransparent{
    if (!self.uiWebView) {[self setupWebView];}
    [self.uiWebView setOpaque:NO];
    [self.uiWebView setBackgroundColor:[UIColor clearColor]];
}
```

PR #17833, opened 2017-05-18
Add const keyward to Node::isScheduled and Scheduler::isScheduled

Riyaaaaa:
Methods with "is" or "get" prefixes should ensure that the method has no side effects.


dumganhar (approved):

minggo:
thanks

PR #17844, opened 2017-05-21
Fixing serious XMLHTTP leak when CC_ENABLE_GC_FOR_NATIVE_OBJECTS enabled

DavidDeSimone:
Fixing issue where XMLHttpRequests would not have their destruction fire if CC_ENABLE_GC_FOR_NATIVE_OBJECTS was set. This will cause the entire XMLHttpRequest to leak, including the **entire data payload** on the XMLHttpRequest. If your game makes a lot of these requests, it can add up very quickly. 

This is being caused by an improper rooting strategy. Basically, we are rooting the callbacks of the XMLHttpRequest, and unrooting them when ~MinXMLHttpRequest() fires. _HOWEVER_, ~MinXMLHttpRequest won't fire, because the GC will never clean up the XMLHttpRequest, because it's callback's are rooted, and these callbacks usually reference the XMLHttpRequest, creating a reference cycle. 

Since the callbacks are rooted, and usually reference the XMLHttpRequest, the GC won't see the XMLHttpRequest as garbage and won't clean it up.

The way I've chosen to fix this is to unroot ALL callbacks once certain event's have happened. For example, once the request is complete, I will unroot all pending callbacks.

This required some changes to the root and unroot functions in cocos2d_specifics. The unroot function didn't work, because 

- The root function will add a reference to jsb._root, but the unroot function attempts to remove references from jsb.jsb._root, which fails
- The unroot function also won't unroot anything that don't have a proxy from jsb_get_js_proxy, even though you can root things without a js proxy in the add root function. 

I can reproduce this issue in cocos2d-x-lite, and was also to reproduce it as in issue in a cocos creator made app. I view this as critical for cocos2d-x-lite and cocos creator, as CC_ENABLE_GC_FOR_NATIVE_OBJECTS is enabled by default in that version of the engine. I was able to reproduce this on iOS and win32. 

**REPRODUCTION STEPS**
With CC_ENABLE_GC_FOR_NATIVE_OBJECTS set to 1 in ccConfig.h, In XHRTest.js, add a call to cc.sys.garbageCollect(); to onEnter;
Run the test multiple times, notice that ~MinXMLHttpRequest never fires
apply this patch
Repeat the above action, notice that the destructor will fire. 

@pandamicro @minggo

### DIFF
--- a/cocos/scripting/js-bindings/manual/cocos2d_specifics.cpp
+++ b/cocos/scripting/js-bindings/manual/cocos2d_specifics.cpp
@@ -688,8 +688,6 @@ void js_remove_object_root(JS::HandleValue target)
     JSContext *cx = engine->getGlobalContext();
     JS::RootedObject global(cx, engine->getGlobalObject());
     JSAutoCompartment(cx, global);
-    JS::RootedObject targetObj(cx, target.toObjectOrNull());
-    js_proxy_t *pTarget = jsb_get_js_proxy(targetObj);
 
     JS::RootedObject jsbObj(cx);
     get_or_create_js_obj(cx, global, "jsb", &jsbObj);
--- a/cocos/scripting/js-bindings/manual/cocos2d_specifics.cpp
+++ b/cocos/scripting/js-bindings/manual/cocos2d_specifics.cpp
@@ -667,7 +667,7 @@ void js_add_object_root(JS::HandleValue target)
     }
 
     JS::RootedObject root(cx);
-    get_or_create_js_obj(cx, jsbObj, "jsb._root", &root);
+    get_or_create_js_obj(cx, jsbObj, "_root", &root);
     JS::RootedValue valRoot(cx, OBJECT_TO_JSVAL(root));
 
     JS::RootedValue retval(cx);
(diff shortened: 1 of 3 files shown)


cocos/scripting/js-bindings/manual/cocos2d_specifics.cpp
```diff
@@ -690,10 +690,6 @@ void js_remove_object_root(JS::HandleValue target)
     JSAutoCompartment(cx, global);
     JS::RootedObject targetObj(cx, target.toObjectOrNull());
     js_proxy_t *pTarget = jsb_get_js_proxy(targetObj);
```
pandamicro: You can also remove the above two lines

pandamicro (approved):
Thanks for the fix ! I can confirm the leak, it's surely very important.

> This required some changes to the root and unroot functions in cocos2d_specifics. The unroot function didn't work, because

These are also bugs, thanks for fixing them

DavidDeSimone:
Thanks for the feedback @pandamicro, I've updated the PR.

pandamicro:
I‘m syncing this PR to cocos2d-x-lite

PR #17875, opened 2017-05-28
Calling PhysicsWorld::setDebugDrawMask(false) will no longer crash wh…

### LINKED ISSUE #17874: Calling PhysicsWorld::setDebugDrawMask(false) right after Scene::initWithPhysics causes a crash
- cocos2d-x version:
- devices test on:
- developing environments
   - Arch Linux GCC version: 6.3.1

Steps to Reproduce:

1. Create a new project and change the first lines of `HelloWorld::init` to the following:

```cpp
    //////////////////////////////
    // 1. super init first
    if ( !Scene::initWithPhysics() )
    {
        return false;
    }

	this->getPhysicsWorld()->setDebugDrawMask(false);
```

2. Run the application and it will crash.

Rubonnek:
Fixes #17874

### DIFF
--- a/cocos/physics/CCPhysicsWorld.cpp
+++ b/cocos/physics/CCPhysicsWorld.cpp
@@ -811,7 +811,7 @@ void PhysicsWorld::removeAllBodies()
 
 void PhysicsWorld::setDebugDrawMask(int mask)
 {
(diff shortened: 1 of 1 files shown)


minggo:
thanks

PR #17889, opened 2017-05-31
RefPtr: Renable ability to use foward declares with RefPtr

millerlyte87:
The addition of this static_assert removed the ability to use forward declared types with a RefPtr, so I've removed it.

```
class Dog;
RefPtr<Dog> m_dog; // Fails at compile time in current v3

class Man;
std::shared_ptr<Man> m_man; // Compiles, RefPtr should follow same paradigm.
```

### DIFF
--- a/cocos/base/CCRefPtr.h
+++ b/cocos/base/CCRefPtr.h
@@ -44,7 +44,7 @@ NS_CC_BEGIN
     {\
         if (ptr)\
         {\
-            const_cast<Ref*>(static_cast<const Ref*>(ptr))->retain();\
+            ptr->retain();\
         }\
     \
     }   while (0);
@@ -55,7 +55,7 @@ NS_CC_BEGIN
     {\
         if (ptr)\
         {\
-            const_cast<Ref*>(static_cast<const Ref*>(ptr))->release();\
+            ptr->release();\
         }\
     \
     }   while (0);
@@ -66,7 +66,7 @@ NS_CC_BEGIN
     {\
         if (ptr)\
         {\
-            const_cast<Ref*>(static_cast<const Ref*>(ptr))->release();\
+            ptr->release();\
             ptr = nullptr;\
         }\
     \
--- a/cocos/base/CCRefPtr.h
+++ b/cocos/base/CCRefPtr.h
@@ -265,9 +265,6 @@ public:
     
 private:
     T * _ptr;
-
-    // NOTE: We can ensure T is derived from cocos2d::Ref at compile time here.
-    static_assert(std::is_base_of<Ref, typename std::remove_const<T>::type>::value, "T must be derived from Ref");
 };
 
 template <class T> inline


minggo:
But we need to make sure that, the class should inherits from `Ref`.

millerlyte87:
I seem to be missing something... Is there a fundamental reason why it has to check the type inherits from Ref? 

RefPtr is a templated class, and will work with any types that implement the retain and release methods. I see there are also macro's defined in the file, e.g. CC_REF_PTR_SAFE_RETAIN. But these macro's will also work fine without the static_assert restriction. Admittedly, if the static_assert is removed, the compilation errors won't be as pretty on an invalid type. But on the other hand, you can use forward declared types with RefPtr<>, just like it use to work before the static_assert was added.

So you can leave the static_assert in, however, the only gain is cleaner compilation errors in the rare case when the programmer uses the RefPtr with a type that doesn't implement retain() and release(). Or, we can remove the static_assert, and re-enable the use of forward declared types with RefPtr<>, just as it once was, and just like the behavior of std::shared_ptr.

millerlyte87:
I seem to be missing something... Why?

If the use of the macro's are removed in the RefPtr<> class, then there is no reason why, except getting cleaner compilation errors when an invalid type is used with RefPtr<>. However, that's the only change needed for it to work, because RefPtr is a templated type, and could work with any types that implement retain() and release() (once the macros are not used in it). Another plus, is removing the static_assert re-enables the use of forward declared types with RefPtr<>. This is a big deal, because forward declares use to work with RefPtr before the static_assert was added, so really, when it was added it broke the API. 

If this is something you can see desired, than I can push another commit removing the use the of the macros in the RefPtr class. Thanks

minggo:
@millerlyte87 yep, the only thing i can see is it uses macros, and will invoke `retain`, `release` there.

cocos/base/CCRefPtr.h
```diff
@@ -44,7 +44,7 @@ NS_CC_BEGIN
     {\
         if (ptr)\
         {\
-            const_cast<Ref*>(static_cast<const Ref*>(ptr))->retain();\
```
minggo, line 47: The compiling error log:
```
/jenkins/workspace/ccs-pr/node/mac/build/../cocos/base/CCRefPtr.h:104:32: error: member function 'retain' not viable: 'this' argument has type 'const cocos2d::__String', but function is not marked const
        CC_REF_PTR_SAFE_RETAIN(_ptr);
                               ^~~~
/jenkins/workspace/ccs-pr/node/mac/build/../cocos/base/CCRefPtr.h:47:13: note: expanded from macro 'CC_REF_PTR_SAFE_RETAIN'
            ptr->retain();\
            ^
/jenkins/workspace/ccs-pr/node/mac/tests/cpp-tests/Classes/UnitTest/RefPtrTest.cpp:121:39: note: in instantiation of member function 'cocos2d::RefPtr<const cocos2d::__String>::RefPtr' requested here
        RefPtr<__String const> ref2 = __String::create("Hello");
                                      ^
```

So it seems should change to `const_cast<Ref*>(ptr)->retain()`

minggo: @millerlyte87 any update?

PR #17892, opened 2017-06-02
Fix particle aspect for non-sqaure tile

madrazo:
Currently, if you use a non square tile (Texture from SpriteFrame) the
particle is stretched to look as a square. This fixes it so the particle
looks as in the original image. Thanks

### DIFF
--- a/cocos/2d/CCParticleSystemQuad.cpp
+++ b/cocos/2d/CCParticleSystemQuad.cpp
@@ -49,6 +49,7 @@ ParticleSystemQuad::ParticleSystemQuad()
 :_quads(nullptr)
 ,_indices(nullptr)
 ,_VAOname(0)
+, _yAspect(1.0)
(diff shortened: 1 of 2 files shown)


cocos/2d/CCParticleSystemQuad.cpp
```diff
@@ -49,6 +49,7 @@ ParticleSystemQuad::ParticleSystemQuad()
 :_quads(nullptr)
 ,_indices(nullptr)
 ,_VAOname(0)
+, _yAspect(1.0)
```
minggo, line 52: 1.0 -> 1.0f

minggo:
@madrazo could you please add a test case? 
Thanks.

madrazo:
@minggo ok

minggo:
@madrazo any update?

Issue #17902, opened 2017-06-07
CCLabel line height is reset when call FontAtlasCache::purgeCachedData


paosidufygthrj:
- cocos2d-x version: 3.15

Steps to Reproduce:

1. create label and set line height.
2. FontAtlasCache::purgeCachedData invoke setTTFConfig,  it resets line height.

PR #17903, opened 2017-06-07
Fix issue #17902

paosidufygthrj:


cocos/2d/CCLabel.cpp
```diff
@@ -416,7 +416,12 @@ Label::Label(TextHAlignment hAlignment /* = TextHAlignment::LEFT */,
         if (_fontAtlas && _currentLabelType == LabelType::TTF && event->getUserData() == _fontAtlas)
         {
             _fontAtlas = nullptr;
+            auto lineHeight = _lineHeight;
```
zilongshanren, line 419: why declare a temp variable here? I think we could just use `_lineHeight` here.

tests/cpp-tests/Classes/LabelTest/LabelTestNew.cpp
```diff
@@ -3397,4 +3398,34 @@ std::string LabelIssueLineGap::subtitle() const
...
+}
+
+void LabelIssue17902::purgeCachedData()
+{
+    FontAtlasCache::purgeCachedData();
+}
+
+std::string LabelIssue17902::title() const
+{
+    return "Github Issue XXX";
```
zilongshanren: please change `XXX` to `17902`, thanks.

zilongshanren:
@paosidufygthrj 
Thanks, I have commented on this PR.

zilongshanren (approved):
This PR looks good to me.

tests/cpp-tests/Classes/LabelTest/LabelTestNew.cpp
paosidufygthrj: Fixed. I will be more careful in the future.

cocos/2d/CCLabel.cpp
paosidufygthrj: Because call setTTFConfig reset _lineHeight .

PR #17908, opened 2017-06-08
fix precision issue

minggo:


Anywhere-Music-Player:
Tested(iOS device only). Fixed. Thanks. Will test on Mac soon.

Anywhere-Music-Player:
@minggo 
On mac got some problem. Just compile tests without any changes.

screen shot 2017-06-08 at 1 45 01 pm

minggo:
fixed, `highp` is not a keyword on mac, it uses opengl instead of opgnel es. It seems it supports shader language 1.1.

Anywhere-Music-Player:
Tested, iOS devices & Mac. Fixed. Thanks.

PR #17925, opened 2017-06-13
Processing Lua file that is encrypted with UTF8+BOM will crash,when free the buffer pointer that is changed pointer address.

Y-way:
Processing Lua file that is encrypted with UTF8+BOM will crash,when free the buffer pointer that is changed  
pointer address.


minggo:
Please use english.
Thanks.

minggo:
@Y-way please send a new PR with english comment message. Or i will do it myself, by you will lose the commit history. Thanks.

minggo:
@Y-way thanks, but commit message is still Chinese.

PR #17929, opened 2017-06-14
On XCode projects, removed specific SDK versions from framework paths.

pyrosphere:
Now instead of, for example: `Developer/SDKs/iPhoneOS7.1.sdk/` which is no longer present we have `Developer/SDKs/iPhoneOS.sdk/` so frameworks are no longer missing and marked as red on XCode.


minggo:
Thanks, how did you change Xcode to fix it like this PR? I opened Xcode, and can not find where to modify.

pyrosphere:
I opened the files `*.xcodeproj/project.pbxproj` directly in a text editor and searched for `Developer/SDKs/iPhoneOS\d+\.\d+\.sdk` (and MacOSX, AppleTVOS).

I did this in my own projects some time ago to fix those red frameworks with every SDK update once and for all.

PR #17930, opened 2017-06-15
revert updating flatbuffers

minggo:
This PR revert https://github.com/cocos2d/cocos2d-x/pull/17349 except reverting *.fbs files since i think they are more simple and they are not used since we can not update flatbuffers.


netunlimited:
When can we expect this to be released e.g. as 3.15.2?

minggo:
It is a good candidate of v3.15.2. But i can not give you the time since i am busy with c++/lua support for creator.

PR #17931, opened 2017-06-15
Setting new texture on sprite will reset the program state that is changed

Y-way:
Setting new texture on sprite will reset the ProgramState that is seted user`s program state

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -372,8 +372,10 @@ void Sprite::setTexture(const std::string &filename)
 
 void Sprite::setTexture(Texture2D *texture)
 {
-    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
-
+    if(_glProgramState == nullptr)
+    {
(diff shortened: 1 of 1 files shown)


cocos/2d/CCSprite.cpp
```diff
@@ -372,8 +372,10 @@ void Sprite::setTexture(const std::string &filename)
 
 void Sprite::setTexture(Texture2D *texture)
 {
-    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
-
+    if(_glProgramState == nullptr)
+    {
+        setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
```
minggo, line 377: GLProgramState::getOrCreateWithGLProgramName will return the same glprograme state, why it will be changed?

Y-way: In my project:
1.set default texture.
2.set mine shader program
3.set new texture.

then, the sprite will reset my shader program.

minggo: sounds reasonable

minggo:
Please change the commit message to english or send a new PR. Thanks.

PR #17938, opened 2017-06-16
Setting new texture for Sprite will reset the program state that is c…

Y-way:
…hanged

In my project:
1.set default texture.
2.set mine shader program
3.set new texture.

then, the sprite will reset my shader program.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -372,8 +372,10 @@ void Sprite::setTexture(const std::string &filename)
 
 void Sprite::setTexture(Texture2D *texture)
 {
-    setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
-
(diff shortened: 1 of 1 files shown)


dumganhar:
This Pull Request breaks `cpp-tests/44:Node: Sprite/2:Testing Sprite ETC1 Alpha support`.
default

merlintt:
@minggo This Pull Request breaks  ETC Alpha support glProgramState which used in animation. 
And I do not think should check _glProgramState before function "setGLProgramState". In function "setGLProgramState", we checked whether value of the gl program state is same. And all gl state values are get from the caches.
I think we should revert the code changes from that cpp file.

image

minggo:
The original design of ETC alpha support has bugs, we should do refactor in v3.17. This PR's logic is correctly, we should not change shader in `setTexture()`, it is weird.

merlintt:
sure, i agree with u, i will fix it temporary in local and expect ur solution in v3.17.

sbrednikhin:
Ok. Fixing it in local and waiting too.

PR #17947, opened 2017-06-19
fixed #17800: [iOS] Streaming audio (normally a background music) maybe cut and may not be looped.

### LINKED ISSUE #17800: AudioEngine loops get eventually cut
- cocos2d-x version: 3.15
- devices test on: iPad
- developing environment: Xcode version:

Steps to Reproduce:
1. Using AudioEgine, play a sound bigger than PCMDATA_CACHEMAXSIZE (1048576) with loop
2. Let the file loop a few times and eventually it will stop looping

The problem seems to be that AudioPlayer::rotateBufferThread will feed the buffer with 100ms of data and sleep for 75ms. If the thread gets delayed a few ms, the decoder will consume all the buffer and stop the sound. Then AudioEngineImpl::update will detect that the sound is stopped and delete the AudioPlayer, ending the loop.

It looks like AudioPlayer::rotateBufferThread is failing to manage all 3 buffers, but I'm not very sure of that (please note that AudioPlayer has 'hardcoded' the buffer count in 3, but there is a define QUEUEBUFFER_NUM that I guess it should be used instead).

A quick fix is to double the data on the buffer to 200ms (#define QUEUEBUFFER_TIME_STEP (0.2f)), or to lower the sleep time on AudioPlayer::rotateBufferThread, although I prefer the first one.

dumganhar:
FIXED: https://github.com/cocos2d/cocos2d-x/issues/17800

### DIFF
--- a/cocos/audio/apple/AudioEngine-inl.h
+++ b/cocos/audio/apple/AudioEngine-inl.h
@@ -66,6 +66,7 @@ public:
 
 private:
     void _play2d(AudioCache *cache, int audioID);
+    static ALvoid myAlSourceNotificationCallback(ALuint sid, ALuint notificationID, ALvoid* userData);
 
     ALuint _alSources[MAX_AUDIOINSTANCES];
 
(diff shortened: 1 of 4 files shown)


kmaker:
I've been testing this PR and I didn't get any audio cut so far. Thanks!

dumganhar:
@kmaker , thanks for testing. :)

PR #17952, opened 2017-06-20
Fixed Director memory leak

Rubonnek:
This leak was introduced in #17669, since_eventBeforeDraw was never released.


minggo:
Oh, it is cause by the https://github.com/cocos2d/cocos2d-x/pull/17897, not https://github.com/cocos2d/cocos2d-x/pull/17669. Thanks.

Rubonnek:
Ah, that's right. Thanks.

PR #17956, opened 2017-06-21
luaoc static function return type BOOL

sgamerw:
Call the luaoc bridge function , when the oc function return type is BOOL, I get the returnType from 

> const char *returnType = [methodSig methodReturnType];

is "B". 
Platform : iOS 10.3.

### DIFF
--- a/cocos/scripting/lua-bindings/manual/platform/ios/CCLuaObjcBridge.mm
+++ b/cocos/scripting/lua-bindings/manual/platform/ios/CCLuaObjcBridge.mm
@@ -122,25 +122,25 @@ int LuaObjcBridge::callObjcStaticMethod(lua_State *L)
         lua_pushboolean(L, 1);
         if (returnLength > 0)
         {
-            if (strcmp(returnType, "@") == 0)
+            if (strcmp(returnType, @encode(id)) == 0)
             {
                 id ret;
                 [invocation getReturnValue:&ret];
                 pushValue(L, ret);
             }
-            else if (strcmp(returnType, "B") == 0) // BOOL
+            else if (strcmp(returnType, @encode(BOOL)) == 0) // BOOL
             {
                 char ret;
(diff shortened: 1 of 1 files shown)


minggo:
Yep, i can reproduce it. Is the return type value changed since iOS 10.3?

minggo:
I tested on MacOS(10.12.2), the return type of `BOOL` is `c`.

sgamerw:
I got some information from the document:
[https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html#//apple_ref/doc/uid/TP40008048-CH100](url) 

And I test the code below again on my iPhone(iOS 10.3)

> if
  (strcmp(returnType, @encode(id)) == 0)
            {
                id ret;
                [invocation getReturnValue:&ret];
                pushValue(L, ret);
            }
            else if (strcmp(returnType, @encode(BOOL)) == 0) // BOOL
            {
                char ret;
                [invocation getReturnValue:&ret];
                lua_pushboolean(L, ret);
            }
            else if (strcmp(returnType, @encode(int)) == 0) // int
            {
                int ret;
                [invocation getReturnValue:&ret];
                lua_pushinteger(L, ret);
            }
            else if (strcmp(returnType, @encode(float)) == 0) // float
            {
                float ret;
                [invocation getReturnValue:&ret];
                lua_pushnumber(L, ret);
            }
            else
            {
                NSLog(@"not support return type = %s", returnType);
                lua_pushnil(L);
            }

Bool type works well, but I have't test other types neither macOS.

sgamerw:
I have tested the code beyond on my Mac(10.12.5) and iPhone(10.3)
```@encode(int), @encode(float), @encode(BOOL)``` all work well :)
This way we can support more return types.
Should I pull a new request?

minggo:
Yep, i think you can modify the codes based on this PR.

PR #17974, opened 2017-06-23
Mark warning as error

minggo:
This PR treat warning as error on iOS debug mode for engine module. Will treat warning as error in more platforms in future. 

It is better to do it in release mode too, but if do so then should modify debug mode to `release` as we want jenkins to make sure all PR will not import warnings. But i don't think use `release` mode by default is a good idea. It is better if can change to use release mode in command line, then jenkins can check both `debug` and `release` modes.


minggo:
It is strange that my local Xcode8.3 don't catch as many warnings as jenkins :(. It is not convenient to modify the codes, especially the codes are in 3rd party libraries.

Issue #17700, opened 2017-04-14
Update GLFW to 3.2.1


dumganhar:
Refer to https://github.com/cocos2d/cocos2d-x/issues/16481

Bilalmirza:
@minggo please update this... people have been asking for this for over a year.

minggo:
@Bilalmirza sorry about it, i will update it after releasing v3.16. I am busy with creator cpp support for a long time.

Bilalmirza:
@Bilalmirza , the upgrade will not be in v3.15 since it's in RC now.I create an issue #17700 here.It will probably be in v3.16.
--
:( :( :( :( :(

minggo:
fixed in https://github.com/cocos2d/cocos2d-x/pull/18434

Issue #16481, opened 2016-08-31
How to use joystick in cocos2d-x?


ZeroObject:
cocos2d-x used GLFW and GLFW support joystick function. I tried to use it. it does not work on windows, but osx and linux is no problem.

```
cocos2d-x 3.12 and 3.13 (GLFW 3.2)
windows 10
osx 10.11
ubuntu 16.04
```

```
static int joysticks[GLFW_JOYSTICK_LAST + 1];
static int joystick_count = 0;

static void joystick_callback(int joy, int event)
{
    if (event == GLFW_CONNECTED)
    {
        int axis_count, button_count;

        glfwGetJoystickAxes(joy, &axis_count);
        glfwGetJoystickButtons(joy, &button_count);

        printf("Found joystick %i named \'%s\' with %i axes, %i buttons\n",
               joy + 1,
               glfwGetJoystickName(joy),
               axis_count,
               button_count);

        joysticks[joystick_count++] = joy;
    }
    else if (event == GLFW_DISCONNECTED)
    {
        int i;

        for (i = 0;  i < joystick_count;  i++)
        {
            if (joysticks[i] == joy)
                break;
        }

        for (i = i + 1;  i < joystick_count;  i++)
            joysticks[i - 1] = joysticks[i];

        printf("Lost joystick %i\n", joy + 1);
        joystick_count--;
    }
}

static void find_joysticks(void)
{
    int joy;

    for (joy = GLFW_JOYSTICK_1;  joy <= GLFW_JOYSTICK_LAST;  joy++)
    {
        if (glfwJoystickPresent(joy))
            joystick_callback(joy, GLFW_CONNECTED);
    }
}

bool GamepadTestScene::init() {
    if (!Layer::init())
    {
        return false;
    }

    memset(joysticks, 0, sizeof(joysticks));
    find_joysticks();
    glfwSetJoystickCallback(joystick_callback);

    return true;
}
```

ZeroObject:
http://gamedev.stackexchange.com/questions/129225/how-to-use-joystick-in-cocos2d-x

dumganhar:
> it does not work on windows

What was the error?

ZeroObject:
It was not any error. GLFWEventHandler::onGLFWError was not called.

>  vs2015 output. xbox 360 Controller was connection on wire.

```
Ready for GLSL
Ready for OpenGL 2.0

{
    gl.version: 4.5.0 NVIDIA 358.91
[...]

PR #17990, opened 2017-06-26
Add controller support for Linux and Win32

Rubonnek:
I think this should be merged after GLFW has been updated to 3.2.1 (Issue #17700), but for some strange reason this pull request works on my Windows machine without the GLFW update. Back on issue #16481 the joysticks didn't work on that same Windows machine I tested this code on.

There is a small trivial GLFW issue in 3.2.1 due to some hat buttons, or the D-pad on some controllers, being mapped as joystick axes. The workaround is trivial -- it's just a matter of identifying these buttons in the callbacks and mapping them to the respective functions these buttons should call. On GLFW 3.3, which is under development, the joystick API has been updated to provide a separate call to get the state of those hat buttons and as far as I can tell this code will have to be updated once GLFW 3.3 is stable and released with Cocos2D-X.

As far as I can tell the code in this pull request should also work on Mac, but it remains untested. I noticed there was already a game controller implementation for Mac and decided to leave it as is, just to not break any current projects that upgrade if this pull request gets merged.

One last issue is that the button key codes returned by GLFW differ greatly from the keys found in [Controller.h](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/CCController.h#L57-L93). There is no standard between the two controllers I tested the code with, and as a result I decided to leave it as is and not add any other key codes in there.

The code I used for testing this PR is currently here: https://github.com/Rubonnek/cocos2d-x-controller-test/tree/master/Classes

Just replace `HelloWorldScene.cpp` and `HelloWorldScene.h` in a new project and build. 

I'll add a controller test to cpp-tests later if needed.


Rubonnek:
[ci rebuild]

Rubonnek:
I'm aware this pull request still needs to be reviewed, but don't merge this pull request yet. I noticed we can use the [joypad configuration files](https://github.com/libretro/retroarch-joypad-autoconfig/tree/master/udev) from the [RetroArch](https://github.com/libretro/RetroArch) project to map the different controller inputs to the `Controller::Key` key codes. We can store those configurations on a `std::map` and fire off events with the right key code.

I will be writing Perl scripts to parse those configuration files soon.

templates/cocos2dx_files.json
```diff
@@ -419,6 +419,7 @@
         "cocos/base/CCConsole.cpp", 
         "cocos/base/CCConsole.h", 
         "cocos/base/CCController-android.cpp", 
+        "cocos/base/CCController-linux-win32.cpp",
```
minggo: Don't have to update the file, it is auto updated by travis building.

Rubonnek: Thanks, I'll remove that change in the next commit.

Rubonnek:
I can't guarantee that the input mappings of the approximately 170 controllers towards each of the `Controller::Key` keys are correct since there seems to be some sort of udev input tweaking within the RetroArch project to easier identify certain inputs and I had to manually patch that, but for now I think this should be a good start.

Rubonnek:
I'm done working on the changes, It looks good to me now, but feedback is welcome.

minggo:
> I think this should be merged after GLFW has been updated to 3.2.1 (Issue #17700),

What happens if this PR is merged without updating GLFW? This PR is easy to cause conflict.

And it is strange that this PR will not trigger travis building. I will close/open this PR to trigger it.

minggo:
Can not trigger travis building event i close/reopen the PR. It is strange. The travis will compile linux/android platform, i can not merge this PR without travis building.

Rubonnek:
> What happens if this PR is merged without updating GLFW? This PR is easy to cause conflict.

Nothing serious happens. In the worst case, the controllers just will not be detected (more details on #16481), but that will not cause any crashes. Backwards compatibility is maintained and updating cocos2d-x after this pull request is merged should not cause any issues.

> Can not trigger travis building event i close/reopen the PR. It is strange. The travis will compile linux/android platform, i can not merge this PR without travis building.

Travis seems to be [blocking my pull requests](https://travis-ci.org/cocos2d/cocos2d-x/requests). I opened two more pull requests to see if I could trigger the building process with it, but I noticed it says: `abuse detected: known offender (request looked fishy)"`. Maybe unblocking me will trigger the building process with Travis.

minggo:
I don't know how to unblocking it. According to [the ticket](https://github.com/travis-ci/travis-ci/issues/6942), it seems you should mail to support@travis-ci.com to  fix the unblocking issue.

Rubonnek:
I'll email them then. I don't know why I got blocked in the first place, and that would be good to know.

Rubonnek:
[ci rebuild]

minggo:
It break iOS/mac build, could you please fix it?

Rubonnek:
I'm working on it. I'm currently setting up the bindings generator in my Mac to replicate the issue.

Rubonnek:
The iOS cpp-tests got built successfully on my Mac, but the job is getting canceled by Travis because it's taking too long.

Rubonnek:
Travis finally built it successfully. :D

PR #17998, opened 2017-06-27
fix issue: 17987

dingwenjie:
RenderTexture's parent scale issue


dingwenjie:
So change is not correct, I‘ll change it  again

PR #18002, opened 2017-06-28
Fix sourcePosition bug for ParticleSystem.

halx99:


minggo:
Sounds reasonable, is there any test case to reproduce it?

halx99:
You can see, actually, the old code will be missing sourcePosition value from .plist file, and always be ZERO.

Dimon4eg:
Details here at the end
[http://discuss.cocos2d-x.org/t/x-studio365-a-more-powerful-game-develop-ide-support-particlesystem-editing-ui-editing-adapting-preview-spine-animation-preview-support-lua-debug-edit-syntax-highlight-outlinng/29660/72](http://discuss.cocos2d-x.org/t/x-studio365-a-more-powerful-game-develop-ide-support-particlesystem-editing-ui-editing-adapting-preview-spine-animation-preview-support-lua-debug-edit-syntax-highlight-outlinng/29660/72)

dumganhar:
I don't know whether this change is correct. But this PR breaks the compatibility.
The result of cpp-tests->37:Node: Particles ->  1、2、3、21、22、24、25、27、29、30、31、32、35、36、40、44、46、47、48、49、52` are totally wrong now.

dumganhar:
In 3.15:
image 1
In current v3 branch:
image 2

minggo:
@halx99 i think we have to revert the codes.

halx99:
Does the test case plists specific sourcePosition?

halx99:
@minggo  @dumganhar I have checked, all the test particles has sourcePosition, so the behavior is right. so I think  remove the sourcePosition from test case's particle plists is better.

halx99:
And at ParticleBatchHybrid::onEnter() should call setEmitterPosition to center the particle.

minggo:
@halx99 sorry, did you mean we should just`setEmitterPosition in ParticleBatchHybrid::onEnter()`? Or we should also modify remove `sourcePosition ` in plist files too?

If we should remove `sourcePosition ` to make it work correctly, then it will break compatibility.

halx99:
Yes.

halx99:
because, the particle are not design in any scene/ui editor, so the testcase particle postion  shoud not dependent the plist sourcePosition.
As the particlesystem design, I think the sourcePosition, Node position is different.

halx99:
In the old engine, It seems (+bug) + (-bug) = 0(looks no bug).

minggo:
particle is created by [PartcileDesigner](https://71squared.com/particledesigner).

halx99:
but the sourcePosition is never applied in the old engine.

halx99:
It just a Particle designer, not UI designer.

minggo:
But other developers will meet the same problem as cpp-tests, we can not ask all developers do it as your suggestion. Old engine may not work correctly, but developers use it like that way. Is there any way to fix it and not break compatibility?

halx99:
I think, there are three choice: 
  1. reserved the bug, not break compatibility
  2. fix the engine bug, break compatibility
  3. fix the bug, Add a function for ParticleSystem, enableCompatibleMode(), and the default the compatible mode is enabled in several engine versions in the future, but should mark the function as depecrated.

minggo:
I am not sure if change to
```c++
this->setPosition(Vec2(x,y));
this->setSourcePosition(Vec2(x,y));
```
will work or not.

halx99:
Let'us try it.

halx99:
But I think add a member _compatibleMode to control it is better:
`
if(!_compatibleMode)
{
   this->setSourcePosition(Vec2(x,y));
}
else 
{
   this->setPosition(Vec2(x,y));
}
`

halx99:
the function should be:
 void setCompatibleMode(bool compatibleMode);
 bool isCompatibleMode() const;

minggo:
Sounds reasonable. But don't make it as deprecated, because it will bring warnings.
`setCompatibleMode ` is not clear, may be can change to `setSourcePositionCompatible`

halx99:
ok

halx99:
agree with you

minggo:
could you please send a PR to v3.16, thanks

halx99:
ok, see here: https://github.com/cocos2d/cocos2d-x/pull/18230

PR #18017, opened 2017-07-01
Use CurrentDirectory as defaultSearchRoot on Win32

halx99:
Default, CurrentDirectory equals to ExeDir, but it's more  flexible, and will not break compatible.

### DIFF
--- a/cocos/platform/win32/CCFileUtils-win32.cpp
+++ b/cocos/platform/win32/CCFileUtils-win32.cpp
@@ -65,16 +65,15 @@ static void _checkPath()
     if (s_resourcePath.empty())
     {
         WCHAR utf16Path[CC_MAX_PATH] = { 0 };
-        GetModuleFileNameW(NULL, utf16Path, CC_MAX_PATH - 1);
-        WCHAR *pUtf16ExePath = &(utf16Path[0]);
+        int nNum = GetCurrentDirectoryW(CC_MAX_PATH - 2, utf16Path);
 
-        // We need only directory part without exe
-        WCHAR *pUtf16DirEnd = wcsrchr(pUtf16ExePath, L'\\');
-
-        char utf8ExeDir[CC_MAX_PATH] = { 0 };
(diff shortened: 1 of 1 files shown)


minggo:
When CurrentDirectory != ExeDir ?

halx99:
When start a program with specific working directiry, such as you can specific it at VS project debug option.

minggo:
@halx99 it break compatibility. You can run js-tests or lua-tests, then you can find the issue. Because some resources are only copied to the folder that contains the `.exe` , such as `COCOS2DX_ROOT/build/Debug.win32/js-tests`, and `GetCurrentDirectory` will retain `COCOS2DX_ROOT/tests/js-tests/project/proj.win32` when running in VS.

So i will revert this PR.

halx99:
@minggo I think, maybe, the working dir can be modified in project properties.

minggo:
Yep, but it may break projects that use older engine.

halx99:
It's so easy, just modify ‘Working Directory' from '$(ProjectDir)' to '$(OutDir)'
default

then, the lua project work fine with VS:
default

minggo:
It is not easy if they don't know the problem, the project just crash or got a black screen. And i will not import any codes that will break compatibility.

minggo:
It doesn't fix any issue, just be more convenient.

halx99:
Because some lua tools use the working directory to locate source file, so there is anothor solution: add to defaultSearchPath,  [working director] and [path of .exe file], what do you think about this?

minggo:
Then it may have problem, which should be searched first?

halx99:
search working directory as first.

minggo:
Ok, could you please send a PR to v3.16 branch?

halx99:
because working dir is configurable, but exe dir isn't.

halx99:
ok.

minggo:
@halx99 any update? We are testing v3.16 branch these days.

halx99:
@minggo I have checked, and the engine does't support multi defaultSearchRoots currently, so Let's revert this PR currently.

minggo:
Yep, the developers can add search path if needed.

PR #18027, opened 2017-07-03
horizontal alignment support in RichText

kepler-5:
RichText supports left, center, and right alignment. Added a button to the top of every test for changing the alignment.

### DIFF
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1800,8 +1800,6 @@ void RichText::formarRenderers()
                 iter->setAnchorPoint(Vec2::ZERO);
                 iter->setPosition(nextPosX, nextPosY);
                 this->addProtectedChild(iter, 1);
-                if ( dynamic_cast<Sprite*>(iter) )
-                    do{}while(0);
                 nextPosX += iter->getContentSize().width;
             }
             
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1826,21 +1826,25 @@ void RichText::formarRenderers()
     updateContentSizeWithTextureSize(_contentSize);
 }
 
+namespace {
+    float getPaddingAmount(const RichText::HorizontalAlignment alignment, const float leftOver) {
+        switch ( alignment ) {
+            case RichText::HorizontalAlignment::CENTER:
+                return leftOver / 2.f;
+            case RichText::HorizontalAlignment::RIGHT:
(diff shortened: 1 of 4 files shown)


cocos/ui/UIRichText.cpp
```diff
@@ -1800,12 +1825,34 @@ void RichText::formarRenderers()
...
+            switch ( alignment ) {
+                case HorizontalAlignment::CENTER:
+                    return leftOver / 2.f;
+                case HorizontalAlignment::RIGHT:
+                    return leftOver;
+                default:
+                    CCASSERT(false, "invalid horizontal alignment!");
+                    return 0.f;
+            }
+        }();
```
minggo: It seems there is not need to use lambda here;

minggo:
I found two issues when running the test case:

* UIRichTextXMLShadow
It seems the alignment changed from like right -> center -> left, while all other test cases changes from left -> center -> right
* the effect of UIRichTextXMLImg is not correct. When use right alignment, texts is not align to the image 
screen shot 2017-07-04 at 2 06 17 pm

kepler-5:
Thanks for the response.  I'll look into those issues and update the PR.

cocos/ui/UIRichText.cpp
kepler-5: You're right, it's not strictly necessary.  The immediately-invoked lambda pattern is useful for complex initialization of variables, which, after initialization, should then be `const`. http://www.bfilipek.com/2016/11/iife-for-complex-initialization.html
(Also note the lambda is very likely inlined by the compiler.)

Without a lambda, it would be difficult to make the `leftPadding` variable `const`, which it should be, since it is not modified later in the function after it is initialized.  Maybe that's not a big deal; if it's too weird, I can change it.

kepler-5:
The shadow test issue was due to the shadow test's label contentSize being too small. Seeing the bounding box of the label, the alignment actually worked correctly. (If a RichText label's contentSize is smaller than the elements therein, shifting from [left, center, right] would actually make the elements appear to move from [right, center, left].) I increased it so that the test looks as expected.

The wrong appearance of the images was due to a bug in how the images were scaled.  Setting a 15x15 image's size to 30x30 in XML actually amounted to quadrupling the size, rather than doubling it.  The updated PR includes a fix for this bug.

cocos/ui/UIRichText.cpp
minggo: It is ok, thanks.

PR #18028, opened 2017-07-04
arithmetic operators for Color4F

kepler-5:
It is very intuitive and useful to be able to do component-wise and scalar arithmetic on float vectors that represent color.  This is already present and very useful in other color representations, such as GLSL's `vec4`.  This pull request allows for the following types of expressions:
```
auto blendedColor = colorA + colorB; // additively blend colorA and colorB
colorC /= 2.f; // halve the brightness and alpha of colorC, useful for fading
```
This expressive power makes more complex color calculations, like blending or interpolation, much easier to read and write:
```
Color4F interpolateTwoColors(const Color4F& a, const Color4F& b, float time) {
    return (b - a)*time + a; // same lerp formula as for simple data types
}
```

Similar operators could be useful for Color3B and Color4B, but they might end up being less intuitive and more unwieldy due to the high probability for integer overflow/underflow.

### DIFF
--- a/cocos/base/ccTypes.h
+++ b/cocos/base/ccTypes.h
@@ -167,21 +167,20 @@ struct CC_DLL Color4F
 
 Color4F& operator+=(Color4F& lhs, const Color4F& rhs);
 Color4F operator+(Color4F lhs, const Color4F& rhs);
+
 Color4F& operator-=(Color4F& lhs, const Color4F& rhs);
 Color4F operator-(Color4F lhs, const Color4F& rhs);
 
 Color4F& operator*=(Color4F& lhs, const Color4F& rhs);
+Color4F operator*(Color4F lhs, const Color4F& rhs);
 Color4F& operator*=(Color4F& lhs, float rhs);
-template<typename T>
-Color4F operator*(Color4F lhs, const T& rhs) {
-    return lhs *= rhs;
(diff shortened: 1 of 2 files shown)


cocos/base/ccTypes.h
```diff
@@ -165,6 +165,24 @@ struct CC_DLL Color4F
...
+Color4F operator+(Color4F lhs, const Color4F& rhs);
+Color4F& operator-=(Color4F& lhs, const Color4F& rhs);
+Color4F operator-(Color4F lhs, const Color4F& rhs);
+
+Color4F& operator*=(Color4F& lhs, const Color4F& rhs);
+Color4F& operator*=(Color4F& lhs, float rhs);
+template<typename T>
+Color4F operator*(Color4F lhs, const T& rhs) {
+    return lhs *= rhs;
+}
```
minggo: Why need this function? Used to convert other types to `Color4F` first then invoke `Color4F& operator*=(Color4F& lhs, const Color4F& rhs)`? If so, need to add similar template function for `-` and `+`?

kepler-5: Fair enough. Those template functions saved a few lines of code in the cpp file, but they're probably more confusing than they're worth.  I've updated the PR.

PR #18036, opened 2017-07-05
Do not define VERY_VERY_VERBOSE_LOGGING by default.

tiagomartines11:
Remove VERY_VERY_VERBOSE_LOGGING from AudioPlayer/Cache so it can be enabled/disabled with proper preprocessor flag.

### DIFF
--- a/cocos/audio/apple/AudioCache.mm
+++ b/cocos/audio/apple/AudioCache.mm
@@ -37,7 +37,6 @@
 
 #include "audio/apple/AudioDecoder.h"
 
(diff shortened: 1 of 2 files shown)


minggo:
looks good to me, but after this modification, the log don't show now, @dumganhar what's your opinion?

dumganhar:
It's ok to disable the log output now. I enabled it only because some developers may get some issues and it will be easier for us to investigate issues via sending the log to us.

dumganhar (approved):

PR #18037, opened 2017-07-05
Check lastDisplayTime before calculating deltaTime.

tiagomartines11:
Fix issue which could generate invalid `dt` during the very first frame (after #17852 was merged).

This issue only happens on Release mode, given that on Debug mode we detect `big delta time` and set it to 1/60.

### DIFF
--- a/cocos/platform/ios/CCDirectorCaller-ios.mm
+++ b/cocos/platform/ios/CCDirectorCaller-ios.mm
@@ -75,7 +75,6 @@ static id s_sharedDirectorCaller;
     if (self)
     {
         isAppActive = [UIApplication sharedApplication].applicationState == UIApplicationStateActive;
-        lastDisplayTime = 0;
         NSNotificationCenter *nc = [NSNotificationCenter defaultCenter];
         [nc addObserver:self selector:@selector(appDidBecomeActive) name:UIApplicationDidBecomeActiveNotification object:nil];
         [nc addObserver:self selector:@selector(appDidBecomeInactive) name:UIApplicationWillResignActiveNotification object:nil];
@@ -108,6 +107,7 @@ static id s_sharedDirectorCaller;
     displayLink = [NSClassFromString(@"CADisplayLink") displayLinkWithTarget:self selector:@selector(doCaller:)];
     [displayLink setFrameInterval: self.interval];
     [displayLink addToRunLoop:[NSRunLoop currentRunLoop] forMode:NSDefaultRunLoopMode];
+    lastDisplayTime = ((CADisplayLink*)displayLink).timestamp;
 }
 
 -(void) stopMainLoop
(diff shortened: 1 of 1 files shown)


minggo:
Yep, you are right. But i think it is better to initialize it in `init` function.

cocos/platform/ios/CCDirectorCaller-ios.mm
```diff
@@ -75,6 +75,7 @@ - (instancetype)init
     if (self)
     {
         isAppActive = [UIApplication sharedApplication].applicationState == UIApplicationStateActive;
+        lastDisplayTime = 0;
```
minggo: may be should change the code to invoke `init` function like this:
`s_sharedDirectorCaller = [[CCDirectorCaller new] init];` ?

tiagomartines11:
Indeed, it looks better now :)

cocos/platform/ios/CCDirectorCaller-ios.mm
tiagomartines11: True, I don't think it was being used atm.

minggo: so will you please modify it too, thanks

minggo: Oh, it seems `init` will be invoked by new method.

minggo: But i think it is better to invoke as `[[CCDirectorCaller alloc] init]`.

tiagomartines11: Should be like you suggested now :) Thanks!

cocos/platform/ios/CCDirectorCaller-ios.mm
```diff
@@ -137,7 +138,10 @@ -(void) doCaller: (id) sender
         
         [EAGLContext setCurrentContext: cocos2dxContext];
 
-        CFTimeInterval dt = ((CADisplayLink*)displayLink).timestamp - lastDisplayTime;
+        CFTimeInterval dt = 0;
+        if (lastDisplayTime != 0) {
+            dt = ((CADisplayLink*)displayLink).timestamp - lastDisplayTime;
+        }
```
minggo: can we initialize `lastDisplayTime ` to 1/60? Then, we don't have to check it every frame, it is good for performance.

minggo: Oh, can not be `1/60`, but may be initialize with a timestamp. What i think it is better to avoid the `if` codes in every frame.

tiagomartines11: I understand that, but in this case lastDisplayTime shouldn't be initialized with 1/60, but rather with some real timestamp, right?

minggo:
@tiagomartines11 thanks for pointing out it, i think i have to refactor the class, so i will send a new PR. I will let you know when it is done.

tiagomartines11:
Ok, no problems! I avoided the if with the last push, but it doesn't look great... looking forward to the refactor!

minggo:
@tiagomartines11 i sent https://github.com/cocos2d/cocos2d-x/pull/18041 for it.

tiagomartines11:
Thank you :)

Issue #17654, opened 2017-04-06
ui::Button 和ui::ImageView设置九宫格区域不正常的问题


xiayangqun:
- cocos2d-x version:3.14.1
- devices test on:iphone6
重现场景，在创建一个ImageView并加载到屏幕上之后，点击一个按钮，更换这个ImageView的纹理，比如说从彩色的纹理换成黑白的纹理。会发现这个ImageView的九宫格拉伸出现了问题，是按照默认的九宫格区域拉伸的，而不是更换纹理之前的九宫格设定区域拉伸。
使用的是
ui::Button::loadTextures
ui::ImageVIew::loadTexture
这个两个api的进行纹理的更换。

tiagomartines11:
That issue applies to several UI components which accept 9-slice sprites.

Whenever `loadTexture()` is called on a ImageView with 9-slice data, its `virtualRenderer` loses `_capInsets` data and the `_renderMode` variable is reset to `QUAD` instead of `SLICE9`.

This is an annoying behaviour introduced with the Sprite re-work on v3.14.

I also tried an alternative, using `getVirtualRenderer()->setSpriteFrame()` instead of `loadTexture()`, but `_contentSize`/`_stretchFactor` is reset in this case.

So since 3.14 it is impossible to replace a 9-slice texture in a Widget with 9-slice support while keeping its size/stretch/slices.

I'd like to see an official solution to this problem at some point :+1:

minggo:
@xiayangqun could you please provide a test case to save our time? Thanks.

tiagomartines11:
Here is a quick example:

```cpp
ImageView* sprite = ImageView::create("Assets/yellow.png");
sprite->setScale9Enabled(true);
sprite->setCapInsets(Rect(10, 10, 44, 44));
sprite->setContentSize(Size(400, 150));
sprite->setNormalizedPosition(Vec2(0.5, 0.2));
scene->addChild(sprite);
    
Button* button = Button::create("Assets/yellow.png");
button->setScale9Enabled(true);
button->setCapInsetsNormalRenderer(Rect(10, 10, 44, 44));
button->setContentSize(Size(400, 150));
button->setNormalizedPosition(Vec2(0.5, 0.5));
button->setTitleText("Press me");
button->addClickEventListener([sprite](Ref* button){
        ((Button*)button)->loadTextureNormal("Assets/red.png");
        sprite->loadTexture("Assets/red.png");
    });
scene->addChild(button);
```

[...]

PR #18038, opened 2017-07-05
Fix issue #17654 so that slicing information is not lost when loading textures.

tiagomartines11:
This is a possible solution for #17654. 

It works for my requirements and mimics the behavior from versions before v3.14.

### DIFF
--- a/cocos/ui/UIButton.cpp
+++ b/cocos/ui/UIButton.cpp
@@ -300,6 +300,12 @@ void Button::setupNormalTexture(bool textureLoaded)
     }
     _normalTextureLoaded = textureLoaded;
     _normalTextureAdaptDirty = true;
+    
+    //FIX issue https://github.com/cocos2d/cocos2d-x/issues/17654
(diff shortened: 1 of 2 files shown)


zilongshanren:
@tiagomartines11  This PR is just a workaround, we'd better provide a fix on the Sprite.
I will consider to fix it, thanks .

tiagomartines11:
I agree! A proper fix is ultimately needed - I can see other issues that exist ever since the Sprite/Scale9 classes were reworked. As long as we keep backwards compatibility, it is okay to refactor and add new functionality, but the changes from 3.14 (and some before that) broke a lot of past features - it should be reviewed as soon as possible :)

Thank you!

zilongshanren:
@tiagomartines11 
I have made a new PR:  https://github.com/cocos2d/cocos2d-x/pull/18065
Could you help to test it?
 thanks.

zilongshanren (changes requested):
This PR is hacking, so I'd rather close it ATM.

PR #18040, opened 2017-07-06
Fixed vsnprintf compatibility with vc12 and vc14

halx99:
see reference: https://github.com/halx99/cocos2d-x/pull/30

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -134,6 +134,12 @@ namespace {
             buf = new (std::nothrow) char[bufferSize];
             if (buf == nullptr)
                 return;
+            /*
+	    pitfall: The behavior of vsnprintf between VS2013 and VS2015/2017 is different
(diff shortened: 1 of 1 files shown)


minggo:
thanks, but i think the comment in halx99#30 is good, could you please also add it?

halx99:
ok

halx99:
pitfall: The behavior of vsnprintf between VS2013 and VS2015/2017 is different
VS2013 or Unix-Like System will return -1 when buffer not enough, but VS2015/2017 will return the actural needed length for buffer at this station
The _vsnprintf behavior is compatible API which always return -1 when buffer isn't enough at VS2013/2015/2017
Yes, The vsnprintf is more efficient implemented by MSVC 19.0 or later, AND it's also standard-compliant, see reference: http://www.cplusplus.com/reference/cstdio/vsnprintf/

minggo:
Oh, i mean add into the source code of `cocos/base/CCConsole.cpp`

halx99:
ok, done it.

PR #18041, opened 2017-07-06
fix invalid dt at start on iOS release mode

minggo:


tiagomartines11:
Thank you :) Seems to be working just fine now!

minggo:
thanks for your feedback

PR #18042, opened 2017-07-06
Node small optimization

madrazo:
This is the same optimization as in  https://github.com/cocos2d/cocos2d-x/commit/5a5218982147bff4e571a4e75d34e76221abff43
Quick return in visit if it is not visible and has no children. Avoids calling processParentFlags, and director's pushMatrix and popMatrix
Thanks

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1244,7 +1244,12 @@ void Node::visit(Renderer* renderer, const Mat4 &parentTransform, uint32_t paren
     {
         return;
     }
-
+    bool visibleByCamera = isVisitableByVisitingCamera();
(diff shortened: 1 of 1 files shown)


Rubonnek:
[ci rebuild]

minggo:
@Rubonnek travis doesn't know `ci rebuild`. I will try to use travis and remove jenkins at the end. Currently, iOS/Mac building on travis may exceed the building time limitation. I will use more prebuilt version of 3rd party library, such as box2d.

minggo:
The logic looks good. But most time, there is not more than on camera, but the codes will slow down the speed.

Rubonnek:
> @Rubonnek travis doesn't know ci rebuild.

@minggo I see. Thanks for the information.

PR #18049, opened 2017-07-09
Error checking for missing files in Hello World example

olehermanse:


olehermanse:
@minggo What is the "default" Jenkins job that failed?

minggo:
I don't know it too, but can ignore it.

minggo:
Thanks, but i think it is better no exit the application, just print the error log and don't operated with error result.

olehermanse:
@minggo What do you mean by

> don't operated with error result.

Return instead of exit? or do you think it is better that the example segfaults if files are not present?

minggo:
for example: `auto label = Label::createWithTTF("Hello World", "fonts/Marker Felt.ttf", 24);` returns null, then don't use `label` variable.

olehermanse:
@minggo I see, thanks. I updated the PR. Running a test now.

templates/cpp-template-default/Classes/HelloWorldScene.cpp
```diff
@@ -44,25 +60,36 @@ bool HelloWorld::init()
...
     // add a label shows "Hello World"
     // create and initialize a label
-    
+
     auto label = Label::createWithTTF("Hello World", "fonts/Marker Felt.ttf", 24);
-    
-    // position the label on the center of the screen
-    label->setPosition(Vec2(origin.x + visibleSize.width/2,
-                            origin.y + visibleSize.height - label->getContentSize().height));
+    if (label == NULL)
```
minggo: NULL -> nullptr

olehermanse: Thanks, fixed.

templates/cpp-template-default/Classes/HelloWorldScene.cpp
```diff
@@ -30,9 +37,18 @@ bool HelloWorld::init()
                                            "CloseNormal.png",
                                            "CloseSelected.png",
                                            CC_CALLBACK_1(HelloWorld::menuCloseCallback, this));
-    
-    closeItem->setPosition(Vec2(origin.x + visibleSize.width - closeItem->getContentSize().width/2 ,
-                                origin.y + closeItem->getContentSize().height/2));
+    auto w = closeItem->getContentSize().width;
+    auto h = closeItem->getContentSize().height;
+    if (w <= 0 || h <= 0)
```
TheCodez, line 42: Checking the size, doesn't make much sense to me as it will already crash when  accessing `getContentSize ` in case `closeItem `is null.

olehermanse: @TheCodez I agree that this is ugly, however in my test it did return a closeItem, with size 0x0, when the files were not there, so it works as intended.

Could improve it by:
```
if (closeItem == nullptr || 
    closeItem->getContentSize().width <= 0 || 
    closeItem->getContentSize().height <= 0)
```
I'm fairly new to cocos2d-x, is there a better way to check if `MenuItemImage::create` loaded the `.png`s correctly?

minggo: No, there is not problem.

@TheCodez good catch, i think it should check `closeItem` too.

olehermanse: @minggo I added a new PR here: https://github.com/cocos2d/cocos2d-x/pull/18057

minggo: thanks

PR #18059, opened 2017-07-12
Allow customizing opacity of PageView indicator nodes

kepler-5:
Previously, the selected index opacity couldn't be changed, and the "deselected" indices were hardcoded to 30% opacity. This PR allows getting/setting the opacity for both.

Also, the "selected" index used to be simply placed on top of the deselected ones. This PR changes that, to hide the index that the "selected" index is currently overlapping. This helps avoid some visual artefacts and allows full control over the appearance of both selected and deselected states.

### DIFF
--- a/cocos/ui/UIPageViewIndicator.cpp
+++ b/cocos/ui/UIPageViewIndicator.cpp
@@ -33,7 +33,7 @@ NS_CC_BEGIN
 
 namespace {
     static const float SPACE_BETWEEN_INDEX_NODES_DEFAULT = 23;
-    constexpr GLubyte INDEX_NODES_OPACITY_DEFAULT = 0.3*255;
+    static const GLubyte INDEX_NODES_OPACITY_DEFAULT = 0.3*255;
 }
 
 namespace ui {
--- a/cocos/ui/UIPageViewIndicator.cpp
+++ b/cocos/ui/UIPageViewIndicator.cpp
@@ -54,6 +54,7 @@ PageViewIndicator* PageViewIndicator::create()
 PageViewIndicator::PageViewIndicator()
 : _direction(PageView::Direction::HORIZONTAL)
 , _currentIndexNode(nullptr)
+, _currentOverlappingIndexNode(nullptr)
 , _spaceBetweenIndexNodes(SPACE_BETWEEN_INDEX_NODES_DEFAULT)
 , _indexNodesScale(1.0f)
 , _indexNodesColor(Color3B::WHITE)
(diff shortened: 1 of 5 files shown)


cocos/ui/UIPageViewIndicator.h
```diff
@@ -61,6 +61,10 @@ class PageViewIndicator : public ProtectedNode
     const Color3B& getIndexNodesColor() const { return _indexNodesColor; }
     void setIndexNodesScale(float indexNodesScale);
     float getIndexNodesScale() const { return _indexNodesScale; }
+	void setSelectedIndexOpacity(GLubyte opacity) { _currentIndexNode->setOpacity(opacity); }
+	GLubyte getSelectedIndexOpacity() const { return _currentIndexNode->getOpacity(); }
+	void setIndexNodesOpacity(GLubyte opacity);
+	GLubyte getIndexNodesOpacity() const { return _indexNodesOpacity; }
```
minggo: please fix indent

cocos/ui/UIPageViewIndicator.h
```diff
@@ -79,9 +83,11 @@ class PageViewIndicator : public ProtectedNode
     PageView::Direction _direction;
     Vector<Sprite*> _indexNodes;
     Sprite* _currentIndexNode;
+	Sprite* _currentOverlappingIndexNode;
```
minggo: what's it used for?

kepler-5: In the current implementation of the class, the "selected" index node simply places itself on top of the corresponding "unselected" index node.  However, both nodes are still visible, which can lead to ugly artifacts around the edges of the "selected" circle, since the "unselected" one is still visible behind it.  Also, if the user sets the "selected" node's opacity to e.g. 50%, the result will look wrong because there will still be the "unselected" node showing through behind it.

My code changes this, and sets the "unselected" node invisible when the selected one is on top of it; hence, we need a pointer to it, so that when the selection changes, the previous overlapped node can be set visible again.

minggo:
Please fix the compiling error of windows

```
 C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xutility(1866): warning C4503: 'std::allocator<_Other>::destroy' : decorated name length exceeded, name was truncated [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\2d\libcocos2d.vcxproj]


       "f:\win32_slave\workspace\ccs-pr\node\win32_bak\build\cocos2d-win32.sln" (Build target) (1) ->
       "f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\2d\libcocos2d.vcxproj" (default target) (5) ->
       (ClCompile target) -> 
         ..\ui\UIPageViewIndicator.cpp(36): error C2146: syntax error : missing ';' before identifier 'GLubyte' [f:\win32_slave\workspace\ccs-pr\node\win32_bak\cocos\2d\libcocos2d.vcxproj]
```

cocos/ui/UIPageViewIndicator.cpp
```diff
@@ -31,7 +31,10 @@ static const char* CIRCLE_IMAGE_KEY = "/__circleImage";
 
 NS_CC_BEGIN
 
-static const float SPACE_BETWEEN_INDEX_NODES_DEFAULT = 23;
+namespace {
+    static const float SPACE_BETWEEN_INDEX_NODES_DEFAULT = 23;
+    constexpr GLubyte INDEX_NODES_OPACITY_DEFAULT = 0.3*255;
```
minggo: i think VS2013 doesn't support `constexpr`

kepler-5: Ah, thanks. I couldn't figure out why the build was failing.

PR #18063, opened 2017-07-13
Moved GLFWEventHandler to its corresponding header

Rubonnek:
There are no explicit changes in this pull request other than exposing `GLFWEventHandler`, but if this pull request gets accepted I think it will have an important implication to Cocos2D-X developers for at least one use case that I have in mind: attaching an alternative UI library for providing an in-game editor.

Without this change there's no easy way to pass raw glfw events to an alternative UI libraries such as [dear imgui](https://github.com/ocornut/imgui), while also passing events to `EventDistpatcher`. One way could be creating a whole new class based on `GLViewImpl` that passes those events to both: `dear imgui` and the `EventDispatcher` but doing that will prompt many unnecessary changes, unnecessary copy-paste that will affect maintainability and that can be avoided with just this pull request.

With this change it's easy to rewire the glfw events [like this](https://github.com/Rubonnek/cocos2d-x-imgui/blob/master/Classes/CCImGui/ImGuiManager.cpp#L216-L223) to [pass raw glfw events to both](https://github.com/Rubonnek/cocos2d-x-imgui/blob/master/Classes/CCImGui/ImGuiManager.cpp#L225-L247) `EventDispatcher` and `dear imgui`.

In the future, should cocos2d-x support webassembly, it will also allow developers to use their same in-game editor based on `dear imgui` (or any other UI library like [nuklear](https://github.com/vurtun/nuklear)) on the web too since there are currently examples of running `dear imgui` alone on the web. See the following live demos:

* [Demo1](https://rawgit.com/Flix01/imgui/2015-10-Addons/examples/addons_examples/html/main.html)
* [Demo2](https://rawgit.com/Flix01/imgui/2015-10-Addons/examples/addons_examples/html/main2.html)
* [Demo3](https://rawgit.com/Flix01/imgui/2015-10-Addons/examples/addons_examples/html/main3.html)


Rubonnek:
@minggo Thanks.

PR #18064, opened 2017-07-13
Fixed typo. Added comments about GLFW 3.3 controller support.

Rubonnek:
No code changes in this pull request. Only comments.

GLFW recently added [gamepad](https://github.com/glfw/glfw/commit/953106e74ddaf925eb2d100ea67bb36bf89167d4) support. Once it's bundled with cocos2d-x, we just have to map those gamepad keys from GFLW to `Controller::Key` key codes. I just added reminders about the changes that need to be made.

GLFW will provide support for more controllers than the ones currently added to `CCController-linux-win32.cpp` by using the [SDL game controller database](https://github.com/gabomdq/SDL_GameControllerDB) and providing a unified input key codes similar to `Controller::Key`.

### DIFF
--- a/cocos/base/CCController-linux-win32.cpp
+++ b/cocos/base/CCController-linux-win32.cpp
@@ -42,6 +42,15 @@ class CC_DLL ControllerImpl
 
 		ControllerImpl()
 		{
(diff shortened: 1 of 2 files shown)


Rubonnek:
@minggo Thanks.

PR #18065, opened 2017-07-13
Fix ui slice issue

zilongshanren:
fix issue: https://github.com/cocos2d/cocos2d-x/issues/17654

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -1654,7 +1654,6 @@ void Sprite::setBatchNode(SpriteBatchNode *spriteBatchNode)
 
     // self render
     if( ! _batchNode ) {
-        _renderMode = RenderMode::QUAD;
         _atlasIndex = INDEX_NOT_INITIALIZED;
         setTextureAtlas(nullptr);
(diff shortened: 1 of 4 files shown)


cocos/2d/CCSprite.cpp
```diff
@@ -1654,7 +1654,6 @@ void Sprite::setBatchNode(SpriteBatchNode *spriteBatchNode)
 
     // self render
     if( ! _batchNode ) {
-        _renderMode = RenderMode::QUAD;
```
minggo, line 1657: why modify the render mode?

tiagomartines11:
Thank you for looking into this.

After apply this PR, I am getting a crash on `TMXTiledMap::create` 

image

This is the line at fault, probably because `_renderMode` has invalid data, so it tries to access a null `_textureAtlas`:

`Texture2D *tex = (_renderMode == RenderMode::QUAD_BATCHNODE) ? _textureAtlas->getTexture() : _texture;`

zilongshanren:
@tiagomartines11 
thanks, I will fix it.

tiagomartines11:
It seems to be working properly now :) Thanks!

PR #18066, opened 2017-07-13
use VAO/VBO to improve performance

minggo:


tiagomartines11:
@minggo I believe this is broken if `CC_TEXTURE_ATLAS_USE_VAO` is disabled on `CCConfig.h` (i.e. if `supportsShareableVAO()` returns false).

minggo:
Yep, i will fix it.

minggo:
fixed in https://github.com/cocos2d/cocos2d-x/pull/18108

PR #18083, opened 2017-07-18
Fix AudioEngine thread concurrency crash

felippeduran:
Fix `AudioEngine` thread concurrency crash between `myAlSourceNotificationCallback(ALuint sid, ALuint notificationID, ALvoid* userData)` and `_play2d(AudioCache *cache, int audioID)` when iterating over `_audioPlayers` instance variable.

This crash happens when you have multiple audio tracks running at the same time and frequently call `_play2d(AudioCache *cache, int audioID)`.


dumganhar (approved):

dumganhar:
Thanks. You're right, `myAlSourceNotificationCallback` is invoked from sub-thread, we should lock the mutex before iterating the map.

Issue #15696, opened 2016-05-22
Animation reset color and opacity of Sprite


noprops:
When I change color or opacity of Sprite and run Animate, color and opacity will be back to default in every setSpriteFrame.
This happens only if TexturePacker algorithm is polygon.

http://discuss.cocos2d-x.org/t/animation-reset-color-and-opacity-of-sprite/29452

felippeduran:
Any progress in this issue? I have the same problem here (cocos2dx 3.12). Besides, it seems that Animate/Animation resets `flippedX` and `flippedY` properties too (or just ignores them).

felippeduran:
I've found that this issue is related to `setSpriteFrame` method in `CCSprite` and seems to be indeed a bug:

``` c++
void Sprite::setSpriteFrame(SpriteFrame *spriteFrame)
{
    // retain the sprite frame
    // do not removed by SpriteFrameCache::removeUnusedSpriteFrames
    if (_spriteFrame != spriteFrame)
    {
        CC_SAFE_RELEASE(_spriteFrame);
        _spriteFrame = spriteFrame;
        spriteFrame->retain();
    }
    _unflippedOffsetPositionFromCenter = spriteFrame->getOffset();

    Texture2D *texture = spriteFrame->getTexture();
    // update texture before updating texture rect
    if (texture != _texture)
    {
        setTexture(texture);
    }

    // update rect
    _rectRotated = spriteFrame->isRotated();
    setTextureRect(spriteFrame->getRect(), _rectRotated, spriteFrame->getOriginalSize());

    if(spriteFrame->hasPolygonInfo())
    {
        _polyInfo = spriteFrame->getPolygonInfo();
    }
    if (spriteFrame->hasAnchorPoint())
    {
        setAnchorPoint(spriteFrame->getAnchorPoint());
    }
}
```

Every frame, `_polyInfo` is reset and flipped vertices and color info are lost.

I've managed to fix the issue by making the following changes to `setSpriteFrame` method:

``` c++
    if(spriteFrame->hasPolygonInfo())
    {
        _polyInfo = spriteFrame->getPolygonInfo();
        if (_flippedX) flipX();
        if (_flippedY) flipY();
        updateColor();
    }
```

And refactored `setFlippedX` and `setFlippedY` to the following:

``` c++
void Sprite::setFlippedX(bool flippedX)
{
    if (_flippedX != flippedX)
    {
        _flippedX = flippedX;
        flipX();
    }
}

[...]

PR #18085, opened 2017-07-18
Fix polygon sprite bug of resetting flip and color information

felippeduran:
Fix polygon sprite (`CCSprite`) bug of resetting flip and color information when `setSpriteFrame(SpriteFrame *spriteFrame)` is called.

This pull request fixes issue #15696.


minggo:
looks goo, thanks

PR #18086, opened 2017-07-19
Avoid unnecessary file reads and parsing if plist file isn't loaded in SpriteFrameCache.

felippeduran:
This is just a small performance improvement I found when cleaning up several spritesheets during context switching (from gameplay to menus and vice-versa).


minggo:
I think can not do it like this because sprite frames may exist but corresponding plist files are cleared. `SpriteFrameCache::removeSpriteFrameByName` will clear loaded plist files, but not all its sprite frames are removed.

PR #18091, opened 2017-07-21
Fix pageview iterator invalidation

kepler-5:
This PR fixes a regression introduced by my pull request #18059. When dynamically adding or removing pages from the PageView, the indicator nodes had an issue with iterator invalidation. That's my mistake.


minggo:
What's the difference between is PR and #18059? I found the modification are the same.

kepler-5:
The relevant changes are in `ffb4d3f`.

PR #18098, opened 2017-07-24
FontAtlas::prepareLetterDefinitions's memory leak.

ggggamer:
when FontFreeType::getGlyphBitmap return a zero-size char array.
to reproduce:
```
auto pLabel = Label::create("{1} + {2}", "TTF/xxxx.TTF", 24);
pLabel->enableOutline(Color4B(0xFF, 0xFF, 0xFF, 255), 1);
addChild(pLabel);
```

### DIFF
--- a/cocos/2d/CCFontAtlas.cpp
+++ b/cocos/2d/CCFontAtlas.cpp
@@ -415,6 +415,8 @@ bool FontAtlas::prepareLetterDefinitions(const std::u32string& utf32Text)
             tempDef.V = tempDef.V / scaleFactor;
         }
         else{
+            if(bitmap)
+             delete[] bitmap
             if (tempDef.xAdvance)
                 tempDef.validDefinition = true;
             else
--- a/cocos/2d/CCFontAtlas.cpp
+++ b/cocos/2d/CCFontAtlas.cpp
@@ -416,7 +416,7 @@ bool FontAtlas::prepareLetterDefinitions(const std::u32string& utf32Text)
         }
         else{
             if(bitmap)
-             delete[] bitmap
+             delete[] bitmap;
             if (tempDef.xAdvance)
                 tempDef.validDefinition = true;
             else
--- a/cocos/2d/CCFontAtlas.cpp
+++ b/cocos/2d/CCFontAtlas.cpp
@@ -416,7 +416,7 @@ bool FontAtlas::prepareLetterDefinitions(const std::u32string& utf32Text)
         }
         else{
             if(bitmap)
-             delete[] bitmap;
+                delete[] bitmap;
             if (tempDef.xAdvance)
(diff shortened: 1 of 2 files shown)


cocos/2d/CCFontAtlas.cpp
```diff
@@ -415,6 +415,8 @@ bool FontAtlas::prepareLetterDefinitions(const std::u32string& utf32Text)
             tempDef.V = tempDef.V / scaleFactor;
         }
         else{
+            if(bitmap)
+             delete[] bitmap
```
minggo: missing `;`

ggggamer: Sorry for typo and your time.

minggo:
And i found `bitmap` is not deleted correctly in `_fontFreeType->renderCharAt(xxx)`.

cocos/2d/CCFontAtlas.cpp
```diff
@@ -415,6 +415,8 @@ bool FontAtlas::prepareLetterDefinitions(const std::u32string& utf32Text)
             tempDef.V = tempDef.V / scaleFactor;
         }
         else{
+            if(bitmap)
+             delete[] bitmap;
```
minggo: could you please fix indent?

ggggamer: My local file's indent is ok, should i push all indent changes?

minggo: May be you use `tab`, and its effect is different in different editors. Please use white space instead.

ggggamer:
Did you mean `FontFreeType::renderCharAt` have memory leak ?

cocos/2d/CCFontAtlas.cpp
ggggamer: OK

minggo:
when `if (bitmap && bitmapWidth > 0 && bitmapHeight > 0)` is ture, bitmap is only used in `FontFreeType::renderCharAt`. And bitmap is deleted only in `else if(_outlineSize > 0)`

ggggamer:
You are right. Maybe we can free memory before `FontFreeType::renderCharAt` method return.

minggo:
I am not sure if should delete bitmap or not. It depends on the internal logic. I am not familiar with this part.

ggggamer:
Em.  My opinion is we can remove free memory logic in `FontFreeType::renderCharAt` and add free logic after `_fontFreeType->renderCharAt` when we using it.

minggo:
I don't quite understand what's your meaning.

ggggamer:
I will open a new pr, only change the first leak.

ggggamer:
new pr is here.  #18114

PR #18108, opened 2017-07-27
fix crash because of error offset

minggo:


tiagomartines11:
It is working properly now. Thanks!

minggo:
thanks for your feedback

PR #18115, opened 2017-07-28
fix PhysicsBox::getSize()

paosidufygthrj:
- cocos2d-x version: 3.15

Order of vertices in chipmunk was changed.


minggo:
did you mean chipmunk changes the oder of vertices? Is there any link about the information? Thanks.

PR #18135, opened 2017-08-08
Improve the efficiency of loadImage

ghost:
Although adding an image multi-times is abnormal, I met this situation in my project. So I think it's better to avoid loading a same image multi-times, that is very slow.


minggo:
@kaishiqi it is better that you send a PR to spine runtime repo, or it will be lost when updating spine runtime. And we can not modify the codes according to its license.

PR #18141, opened 2017-08-10
fixed GLViewImpl::getMonitorSize() for desktop platforms

MrCapone:
When application was created windowed `glfwGetWindowMonitor` may return NULL. So return of `getMonitorSize()` will be ZERO. glfwGetPrimaryMonitor was added on glfw 3.0 so you can use it.

### DIFF
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
@@ -546,6 +546,9 @@ Size GLViewImpl::getMonitorSize() const {
         GLFWwindow* window = this->getWindow();
         monitor = glfwGetWindowMonitor(window);
     }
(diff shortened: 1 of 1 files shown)


dumganhar:
Seems good. I found this:
http://www.glfw.org/docs/latest/group__window.html#gaeac25e64789974ccbe0811766bd91a16

MrCapone:
@dumganhar 
Yes, also http://www.glfw.org/docs/latest/group__monitor.html#ga721867d84c6d18d6790d64d2847ca0b1

dumganhar (approved):

minggo:
> When application was created windowed glfwGetWindowMonitor may return NULL

What does `application was created windowed` mean? Did you mean not full screen mode?

MrCapone:
I mean `GLViewImpl::createWithRect`

PR #18143, opened 2017-08-10
fix android edit box crash on some devices

newnon:
fix crash on some devices 
additional info
https://stackoverflow.com/questions/20221823/android-strange-crash-in-charat-setselection

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
@@ -336,7 +336,12 @@ public class Cocos2dxEditBoxHelper {
                     editBox.setChangedTextProgrammatically(true);
                     editBox.setText(text);
                     int position = text.length();
-                    editBox.setSelection(position);
+                    try {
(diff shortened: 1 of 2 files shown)


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxEditBoxHelper.java
```diff
@@ -336,7 +336,12 @@ public void run() {
                     editBox.setChangedTextProgrammatically(true);
                     editBox.setText(text);
                     int position = text.length();
-                    editBox.setSelection(position);
+                    try {
+                        editBox.setSelection(editBox.getMaxLength() > position ? position : editBox.getMaxLength());
```
dumganhar, line 340: Could `getMaxLength()` probably return `0`?

newnon:
Looks like here is 2 different problems setSelection must be >=0 and <= editBox.getMaxLength() 
but on some devices like some Samsungs it can throw exception even on good argument

minggo:
@newnon could you please add some comment in the codes, that `some devices like some Samsungs it can throw exception even on good argument`. And i think `https://stackoverflow.com/questions/20221823/android-strange-crash-in-charat-setselection` is good to add to comment too.

Thanks.

PR #18145, opened 2017-08-10
fix FontAtlas crash on some Android devices

### LINKED ISSUE #17390: Crash on Android after restore from background in Texture2D::updateWithData
- cocos2d-x version: 3.14.1
- devices test on: Samsung Galaxy Tab A
- developing environments
   - NDK version: r10e (64-bit)

Description:
Game crashes if it was restored from background on Android, and game state was restored.
Callstack:
1. Texture2D::updateWithData()
2. FontAtlas::prepareLetterDefinitions()

Steps to Reproduce:

1. Game should use TTF font for labels. 
2. Send game to background, force system to free game's resources to force resources reloading on restore from background. (Best solution with 100% success: button with an "open Application's page on Google Play" action, and use "launch" button on that page to return to game).
3. Restore game from background
4. Game should crash on first attempt to use TTF Fonts.

Key step to reproduce crash - force game reload resources.

newnon:
fix #17390


cocos/2d/CCFontAtlas.h
```diff
@@ -111,6 +111,8 @@ class CC_DLL FontAtlas : public Ref
 protected:
     void reset();
     
+    void reinit();
```
dumganhar, line 114: Why to set this file to 100755?

newnon:
i think it was mistake

cocos/2d/CCFontAtlas.cpp
```diff
@@ -132,6 +146,8 @@ void FontAtlas::reset()
     _currentPageOrigX = 0;
     _currentPageOrigY = 0;
     _letterDefinitions.clear();
+    
+    reinit();
```
minggo, line 150: Why need to reinit when it is destroyed.

newnon: may be its wrong and this code is not needed

newnon (commented):
it is just adapted code form @SBJoker
i have tested this code in my game and it fixes crash

cocos/2d/CCFontAtlas.cpp
minggo: if it is not needed, then there is not difference from previous codes.

minggo: Oh, it is my fault, it is added in `reset`. The github diff doesn't look good.

PR #18146, opened 2017-08-10
set AVAudioSessionCategoryAmbient for ios

newnon:


dumganhar:
This PR looks good to me.
I found this reference: https://developer.apple.com/library/content/documentation/Audio/Conceptual/AudioSessionProgrammingGuide/AudioSessionCategoriesandModes/AudioSessionCategoriesandModes.html

And `AVAudioSessionCategoryAmbient` is only set after coming back to foreground. That's weird, I found that the logic was come from early version (https://github.com/cocos2d/cocos2d-x/blob/cocos2d-x-3.9/cocos/audio/apple/AudioEngine-inl.mm#L112)

It's probably a bug.

PR #18147, opened 2017-08-10
fix nbsp

newnon:
it is not permitted to break line on nbsp

### DIFF
--- a/cocos/base/ccUTF8.cpp
+++ b/cocos/base/ccUTF8.cpp
@@ -113,7 +113,7 @@ static void trimUTF32VectorFromIndex(std::vector<char32_t>& str, int index)
  * */
 bool isUnicodeSpace(char32_t ch)
 {
-    return  (ch >= 0x0009 && ch <= 0x000D) || ch == 0x0020 || ch == 0x0085 || ch == 0x00A0 || ch == 0x1680
+    return  (ch >= 0x0009 && ch <= 0x000D) || ch == 0x0020 || ch == 0x0085 || ch == 0x1680
     || (ch >= 0x2000 && ch <= 0x200A) || ch == 0x2028 || ch == 0x2029 || ch == 0x202F
     ||  ch == 0x205F || ch == 0x3000;
 }


dumganhar:
I found this:
screen shot 2017-08-14 at 10 10 12
Could you describe the issue you met? Some pseudo code is better.

newnon:
cocos use function isUnicodeSpace to find place for line break, but it bad idea to break line on Non-breaking space it is special symbol thats looks like space but does not allow to break line here. For example apple use it between price and currency sign "1 $". But cocos some time break line on it.

dumganhar:
I noticed that `isUnicodeSpace` is  only used in `Label::getFirstWordLen` of `CCLabelTextFormatter.cpp`.
```c++
int Label::getFirstWordLen(const std::u32string& utf32Text, int startIndex, int textLen)
{
    auto character = utf32Text[startIndex];
    if (StringUtils::isCJKUnicode(character) || StringUtils::isUnicodeSpace(character) || character == (char32_t)TextFormatter::NewLine)
    {
        return 1;
    }

    int len = 1;
    FontLetterDefinition letterDef;
    if(!_fontAtlas->getLetterDefinitionForChar(character, letterDef)) {
        return len;
    }
    auto nextLetterX = letterDef.xAdvance * _bmfontScale + _additionalKerning;

    auto contentScaleFactor = CC_CONTENT_SCALE_FACTOR();
    for (int index = startIndex + 1; index < textLen; ++index)
    {
        character = utf32Text[index];
        if (!_fontAtlas->getLetterDefinitionForChar(character, letterDef))
        {
            break;
        }

        auto letterX = (nextLetterX + letterDef.offsetX * _bmfontScale) / contentScaleFactor;
        if (_maxLineWidth > 0.f && letterX + letterDef.width * _bmfontScale > _maxLineWidth
            && !StringUtils::isUnicodeSpace(character))  // FIRST
        {
            return len;
        }

        nextLetterX += letterDef.xAdvance * _bmfontScale + _additionalKerning;

        if (character == (char16_t)TextFormatter::NewLine
            || StringUtils::isUnicodeSpace(character) // SECOND
            || StringUtils::isCJKUnicode(character))
        {
            break;
        }
        len++;
    }

    return len;
}
```
I'm not familiar with the code in `Label`, so I guess if there is a `nbsp` at the end of text,
`//FIRST` `isUnicodeSpace` I commented at the above code will return `true`, therefore it will not go into `return len`, and `//SECOND` line will go to `break`. 
The `len` will not take into account for `nbsp`.
But this patch will ignore `nbsp` which means `StringUtils::isUnicodeSpace(nbsp)` will return false, it may make a label `line break`, right?

I may misunderstand it. Therefore, some code snippets for describing this issue will help me find out the problem. 

BTW, we need the code for reproducing this issue and add it to test case. Thanks.

newnon:
@minggo @dumganhar nbsp is special symbol thats look like space but on that is not allowed to make new line. This pr just remove it from place where line break can happen.

dumganhar:
@minggo 
What @newnon said sounds reasonable. Could you consider to merge this PR?

PR #18153, opened 2017-08-11
fix android webview keyback bug after opened a page

BlackCatz94:
bug happen when:
1.implement keyback in a scene
2.implement webview in that scene and load a default URL
3.click any link and wait for it to load
4.press key back, the implemented keyback will not be called

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebView.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebView.java
@@ -50,7 +50,7 @@ public class Cocos2dxWebView extends WebView {
         this.mJSScheme = "";
 
         this.setFocusable(true);
-        this.setFocusableInTouchMode(true);
+        this.setFocusableInTouchMode(false);
 
         this.getSettings().setSupportZoom(false);
 


minggo:
could you please add a test case for it, thanks.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxWebView.java
```diff
@@ -50,7 +50,7 @@ public Cocos2dxWebView(Context context, int viewTag) {
         this.mJSScheme = "";
 
         this.setFocusable(true);
-        this.setFocusableInTouchMode(true);
+        this.setFocusableInTouchMode(false);
```
drelaptop, line 53: The issue you described is existed, and your changes fix the issue, but it has a bad effect on others. 
when you set `this.setFocusableInTouchMode(false);`, some item in WebView can't get focus. For example load url `https://www.baidu.com`, the search bar can't  get focus when you touch on it.

drelaptop (changes requested):

drelaptop:
this PR solved a issue, but caused other issues. Meanwhile author didn't response for a long time, I suggest to close this PR. @minggo

Issue #15926, opened 2016-06-22
UITextField crash when text exceeds content with enabled cursor


sbrednikhin:
Repeating:
In UITextFieldTest_LineWrap::init

add 
'''textField->setCursorEnabled( true );'''
after
''' TextField\* textField = TextField::create("input words here","fonts/Marker Felt.ttf",30);'''

then run test and in test tape some letters till test will crash.

minggo:
Sorry, i don't quite understand the issue? Could you please describe it in more detail?

sbrednikhin:
Sorry for misunderstanding.

If you will enter symbols (a lots of symbols) into text field that have some content size and enabled cursor - it will crash when cursor walks outside of content area.

minggo:
Will you provide a test case? @zilongshanren could you please take a look?

AIGRIND-LLC:
Confirmed that bug in cocos2dx v3.12.
To reproduce the crash a TextField must have fixed content size that is not enough to display all text content.
**Cursor must be enabled** also for this TextField.

![text_fielad_1](https://cloud.githubusercontent.com/assets/6168693/19072447/8350f52a-8a35-11e6-8ae8-8a4f04f2ed7a.png)

If you click on the TextField the app will crash on assert.

![text_field_2](https://cloud.githubusercontent.com/assets/6168693/19072586/54ddd0b8-8a36-11e6-9cf1-85e23ec02423.png)

AIGRIND-LLC:
There is pull request containing fix for this issue:
https://github.com/cocos2d/cocos2d-x/pull/17786

zilongshanren:
@AIGRIND-LLC 
I have tested this issue on IOS, and it didn't crash.

 I have tested this issue with the latest v3 branch code.

Does this issue happen only on Win32 or cocos2d-x v3.12?

thanks.

sbrednikhin:
All versions of cocos between 3.12 and 3.14. On Win32 and Android.
More: It's not fully stable crush. It may crush after 3 symbols or after 20. It depends on cursor "pahse".
Please look into pull request.

AIGRIND-LLC:
@zilongshanren
This bug can be reproduced in the latest version of framework (incl. 3.15)
Did you **enable cursor** for the text field? Also a **fixed content size** must be set for it.

zilongshanren:
[...]

PR #18159, opened 2017-08-13
Fix crash TextFieldTTF::setCursorFromPoint

perminovVS:
A similar problem of
https://github.com/cocos2d/cocos2d-x/issues/15926
And fix it in https://github.com/cocos2d/cocos2d-x/commit/afaa9be8262e64fbc090cf2c202871efe779b94e


minggo:
how to reproduce it?

perminovVS:
@minggo Now it api need used in user code.
Add call it, in touch event to TextFieldTTF or UITextField.

minggo:

perminovVS:
@minggo Well, if possible, I will add.

minggo:
any update?

perminovVS:
There is no time. I'll try to do it on the weekend.

perminovVS:
@minggo test it in https://github.com/cocos2d/cocos2d-x/pull/18336

Issue #18163, opened 2017-08-14
Missing game-controller-test for Windows msvc (and Linux)


madrazo:
- cocos2d-x version: git 
- devices test on: Windows
- developing environments
   - VS version: 2015 update 3

Hi. I am eager to test CController for Windows, as it is something I wished for a long time. I mean this commit https://github.com/cocos2d/cocos2d-x/commit/5dd81e758a874f42da9f42a216824edc566dd4b0

However in the tests\game-controller-test directory, there are only tests for android, ios and mac. I kindly request you to provide it if you have it at hand. Thanks!

madrazo:
OK, I made it work and seems to be good. Even with 2 controllers at the same time. The only issue I had was that there are different mappings in the 3 controllers I had tested. These are a Windows controller, the Ps4 and de Xbox One controllers.  But this was expected. Thanks

PR #18164, opened 2017-08-14
Add game-controller-test for win32

madrazo:
The project files I used to build Controller test for windows https://github.com/cocos2d/cocos2d-x/issues/18163


minggo:
thanks

PR #18166, opened 2017-08-15
Fix android compile fail bug

jie-meng:
Newest version of cocos2dx run "cocos compile -p android --android-studio" failed on reason:

```
{PROJECT_NAME}/frameworks/runtime-src/proj.android-studio/app/src/org/cocos2dx/javascript/AppActivity.java:31: error: cannot find symbol
    protected void onCreate(Bundle savedInstanceState) {
                            ^
  symbol:   class Bundle
  location: class AppActivity
```
because of the method ```protected void onCreate(Bundle savedInstanceState)``` was override but Bundle class not imported.

Add "import android.os.Bundle" would be ok.


minggo:
thanks

Issue #16900, opened 2016-11-24
outdated CMakeLists.txt on templates


MrCapone:
`CMakeLists.txt` in `templates` not been updated for over a year only received minor fixes. But `CMakeLists.txt` from `cpp-empty-test` much better and has more features. I created gist with sample `CMakeLists.txt` based on  `cpp-empty-test/CMakeLists.txt` but modified for use in `templates` 
https://gist.github.com/MrCapone/ccdc1826cb7362086107e8096bfe6f90
It's only for cpp, but it can be used with small changes in js and lua. 
It can:
build and run project on linux/Cmake+gcc
build and run valid .app on macOS/Cmake+clang
generate valid Xcode project for macOS
generate valid Visual Studio project for win32

not yet tested for android studio

also i tried build win32 project on win32/MinGW, but it failed because of precompiled libraries.

minggo:
Don't have to consider Android Studio, Android Studio doesn't support normal cmake files.

MrCapone:
Generated Visual Studio 2015 project crashes on linking because of prebuilt libs. I will try to fix i and then i can make pull request.

MrCapone:
@minggo should i push my [CmakeLists.txt](https://gist.github.com/MrCapone/ccdc1826cb7362086107e8096bfe6f90) as PR? It works good on cpp projects with `linux/gcc`, `macOS+clang`, `Windows+nmake`, also it generates valid projects for `Xcode (only macOS target)` and `Visual Studio 2013/2015/2017 (only win32 target)`. Also i changed macOS output target for generate valid macOS .app bundle.

minggo:
Yep, thanks.

MrCapone:
done https://github.com/cocos2d/cocos2d-x/pull/18167

PR #18154, opened 2017-08-11
V3 android cmake support

newnon:
CMake build improvements & android CMake support 
add to graddle.properties
PROP_NDK_MODE=cmake 
replace files in folder external/android-specific 
[android-specific.zip](https://github.com/cocos2d/cocos2d-x/files/1217221/android-specific.zip)

last PR https://github.com/cocos2d/cocos2d-x/pull/17100

pull request for external
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/300

PR #18167, opened 2017-08-15
add new CMakeLists.txt to cpp-template-default

MrCapone:
New `CmakeLists.txt` for `cpp-template-default` based on `cpp-empty-test/CMakeLists.txt`. 

It can:
build and run binary on `linux+gcc`
build and run valid .app (bundle) on `macOS+clang`
build and run valid .exe on `VisualStudio+nmake` (`cmake -G "NMake Makefiles"`)
generate valid `Xcode` project for `macOS` (`cmake -G Xcode`)
generate valid `Visual Studio` project for `win32`  (`cmake -G "Visual Studio 14"`)

Full backward compatibility.  `ANDROID` target not changed.

Full history on https://gist.github.com/MrCapone/ccdc1826cb7362086107e8096bfe6f90

fixes issue https://github.com/cocos2d/cocos2d-x/issues/16900

### DIFF
--- a/templates/cpp-template-default/CMakeLists.txt
+++ b/templates/cpp-template-default/CMakeLists.txt
@@ -63,16 +63,16 @@ endif()
 # Compiler options
 if(MSVC)
   if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:msvcrt /NODEFAULTLIB:libcmt")
+  	set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:msvcrt /NODEFAULTLIB:libcmt")
   else()
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:libcmt")
+	set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:libcmt")
   endif()
   ADD_DEFINITIONS(-D_CRT_SECURE_NO_WARNINGS -D_SCL_SECURE_NO_WARNINGS
                   -wd4251 -wd4244 -wd4334 -wd4005 -wd4820 -wd4710
                   -wd4514 -wd4056 -wd4996 -wd4099)
 else()
-  if(CMAKE_BUILD_TYPE MATCHES Debug)
-    ADD_DEFINITIONS(-DCOCOS2D_DEBUG=1)
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+	ADD_DEFINITIONS(-DCOCOS2D_DEBUG=1)
   endif()
   set(CMAKE_C_FLAGS_DEBUG "-g -Wall")
   set(CMAKE_CXX_FLAGS_DEBUG ${CMAKE_C_FLAGS_DEBUG})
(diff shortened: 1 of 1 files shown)


MrCapone:
Also i compare with this pull request https://github.com/cocos2d/cocos2d-x/pull/18154

Only one string different `templates/cpp-template-default/CMakeLists.txt:81`
[this PL](https://github.com/cocos2d/cocos2d-x/pull/18167/files#diff-db25b971276415a936f4d3bf7600b488L81) vs [V3 android cmake support](https://github.com/cocos2d/cocos2d-x/pull/18154/files#diff-db25b971276415a936f4d3bf7600b488L81)

So my PL can be merged with #18154

minggo:
Did you mean i should merge #18154 first, then this PR?

MrCapone:
@minggo 
I've changed my commit, so you able merge my PR first, then https://github.com/cocos2d/cocos2d-x/pull/18154 without any conflicts. Line 81 and 83 now equal.

minggo:
thanks, i will test in on ubuntu.

minggo:
it works good, thanks

minggo:
@MrCapone will you please take a look of https://github.com/cocos2d/cocos2d-x/issues/18187?

templates/cpp-template-default/CMakeLists.txt
```diff
@@ -83,100 +83,157 @@ else()
...
+            RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin/$<CONFIG>"
+            )
 
 else()
-  pre_build(${APP_NAME}
-    COMMAND ${CMAKE_COMMAND} -E remove_directory ${APP_BIN_DIR}/Resources
-    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_SOURCE_DIR}/Resources ${APP_BIN_DIR}/Resources
-    )
+
+    set(APP_BIN_DIR "${CMAKE_BINARY_DIR}/bin/$<CONFIG>")
```
minggo, line 231: what's `$(CONFIG)`? now the bin will be in `bin/debug/linux/Debug/MyGame` instead of `bin/debug/linux/MyGame`,  it breaks cocos command.

minggo: i correct it in https://github.com/cocos2d/cocos2d-x/pull/18258, i think `bin/debug/linux/MyGame` is more reasonable.

MrCapone: If you build by
```
cmake ..
make
```
bin dir will be just `bin/`. So all targets (debug, release etc) will be in same folder.

minggo: it is strange, i think cocos command also use cmake and make to build the app. May be cocos command copy the output to needed path then run it. If so, then i have to modify cocos command. But this will not be included in v3.16.

minggo: yep, cocos command has a deploy stage, it will copy bin and resources generated by cmake into `project_root/bin/debug|release/linux` folder. If i change the codes in cocos, then we should modify all templates and all test cases. I will leave it in next version.

MrCapone: In my CMakeLists.txt win32 and macos targets also use separate path fo debug|release. cocos console doesn't use CMake for win32 and macos, but i think its good idea use same path for all targets in CMakeLists.txt. So maybe you will change APP_BIN_DIR in those targets too?

MrCapone: Also i think good idea change project name on `set(APP_NAME MyGame)` after execution `cocos new`command (MyGame -> %project_name%).
I would have done it myself on this PR, but I did not understand very well cocos console logic.

minggo: The copy logic of linux is [here](https://github.com/cocos2d/cocos2d-console/blob/v3/plugins/plugin_compile/project_compile.py#L1468), we can just modify the `result_dir `. Other platforms will have similar logic. You can try it yourself. I don't use cmake for other platforms except linux. But i think it is good if we can use cmake to generate VS|Xcode projects dynamically. That's what i want to have, but i don't have time to do it.

The `cocos new` command logic is [here](https://github.com/cocos2d/cocos2d-console/blob/v3/plugins/plugin_new/project_new.py).

PR #18171, opened 2017-08-17
Improved Cocos2dxLuaLoader.cpp

v1993:
Cocos2dxLuaLoader.cpp has been rewritten. Now it implements all default loader and cocos-only features.

### DIFF
--- a/cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
+++ b/cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
@@ -41,17 +41,13 @@ extern "C"
 
         std::string filename(luaL_checkstring(L, 1));
         size_t pos = filename.rfind(BYTECODE_FILE_EXT);
-        if (pos != std::string::npos)
-        {
+        if (pos != std::string::npos && pos == filename.length() - BYTECODE_FILE_EXT.length())
             filename = filename.substr(0, pos);
-        }
         else
         {
             pos = filename.rfind(NOT_BYTECODE_FILE_EXT);
-            if (pos == filename.length() - NOT_BYTECODE_FILE_EXT.length())
-            {
+            if (pos != std::string::npos && pos == filename.length() - NOT_BYTECODE_FILE_EXT.length())
                 filename = filename.substr(0, pos);
-            }
         }
 
         pos = filename.find_first_of(".");
@@ -77,33 +73,55 @@ extern "C"
         {
             if (next == std::string::npos)
                 next = searchpath.length();
-            std::string prefix = searchpath.substr(begin, next);
+            std::string prefix = searchpath.substr(begin, next-begin);
             if (prefix[0] == '.' && prefix[1] == '/')
-            {
                 prefix = prefix.substr(2);
-            }
 
-            pos = prefix.find("?.lua");
-            chunkName = prefix.substr(0, pos) + filename + BYTECODE_FILE_EXT;
-            if (utils->isFileExist(chunkName))
+            pos = prefix.rfind(BYTECODE_FILE_EXT);
+            if (pos != std::string::npos && pos == prefix.length() - BYTECODE_FILE_EXT.length())
+            {
+                prefix = prefix.substr(0, pos);
+            }
+            else
+            {
+                pos = prefix.rfind(NOT_BYTECODE_FILE_EXT);
+                if (pos != std::string::npos && pos == prefix.length() - NOT_BYTECODE_FILE_EXT.length())
+                    prefix = prefix.substr(0, pos);
+            }
+            pos = prefix.find_first_of("?", 0);
+            while (pos != std::string::npos)
+            {
+                prefix.replace(pos, 1, filename);
+                pos = prefix.find_first_of("?", pos + filename.length() + 1);
+            }
+            chunkName = prefix + BYTECODE_FILE_EXT;
+            if (utils->isFileExist(chunkName) && !utils->isDirectoryExist(chunkName))
             {
                 chunk = utils->getDataFromFile(chunkName);
                 break;
             }
             else
             {
-                chunkName = prefix.substr(0, pos) + filename + NOT_BYTECODE_FILE_EXT;
-                if (utils->isFileExist(chunkName))
+                chunkName = prefix + NOT_BYTECODE_FILE_EXT;
+                if (utils->isFileExist(chunkName) && !utils->isDirectoryExist(chunkName))
                 {
                     chunk = utils->getDataFromFile(chunkName);
                     break;
+                }
+                else
+                {
+                    chunkName = prefix;
+                    if (utils->isFileExist(chunkName) && !utils->isDirectoryExist(chunkName))
+                    {
+                        chunk = utils->getDataFromFile(chunkName);
+                        break;
+                    }
                 }
             }
 
             begin = next + 1;
             next = searchpath.find_first_of(";", begin);
-        } while (begin < (int)searchpath.length());
-
+        } while (begin < searchpath.length());
         if (chunk.getSize() > 0)
         {
             LuaStack* stack = LuaEngine::getInstance()->getLuaStack();
--- a/tests/lua-tests/src/LuaLoaderTest/LuaLoaderTest.lua
+++ b/tests/lua-tests/src/LuaLoaderTest/LuaLoaderTest.lua
@@ -1,0 +1,52 @@
+local testPatches = {
+{'LuaLoaderTest/?.lua', 'simple'},
+{'LuaLoaderTest/?/init.lua', 'in_dir'},
+{'LuaLoaderTest/?.lc', 'strange_ext'},
+{'./LuaLoaderTest/?.lua', 'simple'}, -- test removing leading `./`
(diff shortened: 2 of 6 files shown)


v1993:
Travis-ci just reached time limit on one job.

v1993:
@minggo Can you merge it, please? It's critical for my library and full tested.

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
```diff
@@ -41,17 +41,13 @@ extern "C"
 
         std::string filename(luaL_checkstring(L, 1));
         size_t pos = filename.rfind(BYTECODE_FILE_EXT);
-        if (pos != std::string::npos)
-        {
+        if (pos != std::string::npos && pos == filename.length() - BYTECODE_FILE_EXT.length())
```
minggo, line 44: why need this line `pos == filename.length() - BYTECODE_FILE_EXT.length()`?

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
```diff
@@ -41,17 +41,13 @@ extern "C"
...
             filename = filename.substr(0, pos);
-        }
         else
         {
             pos = filename.rfind(NOT_BYTECODE_FILE_EXT);
-            if (pos == filename.length() - NOT_BYTECODE_FILE_EXT.length())
-            {
+            if (pos != std::string::npos && pos == filename.length() - NOT_BYTECODE_FILE_EXT.length())
```
minggo, line 49: why need this line: `pos == filename.length() - NOT_BYTECODE_FILE_EXT.length()?`

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
```diff
@@ -77,33 +73,55 @@ extern "C"
...
+            else
+            {
+                pos = prefix.rfind(NOT_BYTECODE_FILE_EXT);
+                if (pos != std::string::npos && pos == prefix.length() - NOT_BYTECODE_FILE_EXT.length())
+                    prefix = prefix.substr(0, pos);
+            }
+            pos = prefix.find_first_of("?", 0);
+            while (pos != std::string::npos)
+            {
+                prefix.replace(pos, 1, filename);
```
minggo, line 94: Why replace `?` with `filename`? What's the result if search path include `?` in the middle of the path, such as `./lib/?/init.lua'`?

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
v1993: Path can include `.luac` in the middle, but we only should cut it at the end.

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
v1993: Path can include `.lua` in the middle, but we only should cut it at the end.

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
v1993: It is lua's standart, refer to https://www.lua.org/manual/5.1/manual.html#pdf-package.loaders, in case of `./lib/?/init.lua` we should try `./lib/filename/init.lua`.

minggo: thanks.

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
minggo: sounds reasonable

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
minggo: sounds reasonable

cocos/scripting/lua-bindings/manual/Cocos2dxLuaLoader.cpp
```diff
@@ -77,33 +73,55 @@ extern "C"
...
-                if (utils->isFileExist(chunkName))
+                chunkName = prefix + NOT_BYTECODE_FILE_EXT;
+                if (utils->isFileExist(chunkName) && !utils->isDirectoryExist(chunkName))
                 {
                     chunk = utils->getDataFromFile(chunkName);
                     break;
                 }
+                else
+                {
+                    chunkName = prefix;
```
minggo, line 113: Can not quite understand this. Could you please give an example? Thanks.

v1993: Someone may create no-extension file or file with non-standart extension and try to `require` it.

https://www.lua.org/manual/5.1/manual.html#pdf-package.loaders provides an example: `package.path` contain `./?.lc`.

Maybe, somone want to prevent compilation into bytecode. Anyway, it acts like reference says.

minggo:
This PR looks good, could you please add test case for it?
Thanks.

v1993:
But what should this test case test? Loading modules from various patches?

v1993:
There is a problem: you should test both on non-luacompiled version and on luacompiled version. @minggo, what do you think about it?

Of couse, I can create non-compiled .luac file just for testing, but is it good?

minggo:
.luac will only be generated in release mode, so i think there is not problem. We can test debug and release mode.

minggo:
I think we can just test

- filename extension is .lc or other
- `?` in middle of search path, such as `a/?/b`
- `.lua` in middle of filename, such as `a/.lua/c.lua`

These are all the questions i asked above.
What do you think? I may lost something.

v1993:
Thank you very much, should I store every test files into test's directory? In other words, should every path be `LuaLoaderTest/...`? Also, `.lua` in the middle will be decoded as `/lua`. I'm not sure about cutting off suffix at all.

minggo:
Yep, i think `LuaLoaderTest/` is a good place to store files for lua loader test. 

> Also, .lua in the middle will be decoded as /lua. I'm not sure about cutting off suffix at all.

I think it is good to test every condition we may consider, that's the purpose of tests.

v1993:
>>Also, .lua in the middle will be decoded as /lua. I'm not sure about cutting off suffix at all.
>
>I think it is good to test every condition we may consider, that's the purpose of tests.

I'm talking about cutting suffix at all - in my opinion, it isn't very good idea. At all, by lua reference, dot should be replaced with path separator (`/` in cocos), so `filename` (as argument to `require`) souldn't include extension.

minggo:
Sorry, i don't quite understand. Did you mean cocos replace `.` with `/` and add file extension?

v1993:
We cut off `.lua` or `.luac` extensions from filename, right?

minggo:
Yep, just to load `.luac` or `.lua` automatically. What's the problem?

v1993:
It is good for path, but not for filename, because by standarts `.` in `.lua` or `.luac` should be replaced with file separator (`/`). Refer to same docs.

minggo:
But the `filename` may include part of path, such as `require a/b`.

v1993:
`filename` **may** include part on the path, it should be like this: `a` or `a.b` (load submodule `b` of `a`), which means `a/b` instead of every `?` in path. So, `a.lua` in **`filename`** means `a/lua`, but in path means just `a.lua`.

minggo:
So what's your suggestion?

v1993:
I think that we just should stop removing `.lua` and `.luac` at the end of filename. But it may break compatiblity. Maybe, I will do some research later, if it will break template code or `lua-tests`.

v1993:
My opinion that is should be somehow marked as *deperaced and will be removed later*. So I will not make test for this one.

minggo:
cocos will load `.luac` first then `.lua` if they both exist.

v1993:
It will be still aviable, because removing `.luac`/`.lua` and then trying them will be saved for `path`, but not `filename`.

v1993:
@minggo What do you think?

minggo:
looks good, thanks

drelaptop:
@v1993 some developers make feedbacks that 3.17 slow the lua loadind speed, the load speed is quicker in 3.16 release.

only this PR change the lua loading logic between 3.16 and 3.17. could you review those changes for performance issue.

v1993:
@drelaptop It makes loading a bit more complicated, but correct. Can they use profiler to detect slow parts, please?

drelaptop:
this is discussed at Chinese forum, at http://forum.cocos.com/t/cocos2d-x-3-17-android-lua/61285/12

one developer think it's because the while loop, like below

lua-loader

v1993:
@drelaptop First `whihe` almost all times execute only once, most pathes contain only one `?`.

Second one always was here: it iterates through `;` separated sections.

P.S.: It is probably better to rewrite them using `<regex>` header. What do you think?

PatriceJiang:
performance issue is fixed at https://github.com/cocos2d/cocos2d-x/pull/19234

PR #18175, opened 2017-08-17
Fix alpha blending during the animation of TransitionCrossFade.

pyrosphere:
This PR fixes the incorrect blending in TransitionCrossFade that causes the transition to be darker than it should.


minggo:
thanks

PR #18192, opened 2017-08-25
Fix bug of returning UNKNOWN event for return type key SEARCH/SEND in EditBox.

felippeduran:
If you set and `EditBox` return type to `SEARCH` (Android and iOS) or `SEND` (Android) and click on the corresponding keyboard button, you get `UNKNOWN` event.

This pull request fixes this issue.


dumganhar (approved):

minggo:
@felippeduran  thanks, could you please add a test case for it?

felippeduran:
Yes! Sorry for the late reply, @minggo. I'll try to add a test case for it asap and add it here!

minggo:
@felippeduran the codes looks good to me, will you add a test case for it? Of course i can merge it without the test case.

minggo (approved):

PR #18204, opened 2017-08-30
Stop removing .lua and .luac at the end of filename in LuaLoader

v1993:
I've preformed some research and testing, so now you can safely stop this. Previous discussion and reasons you can found here: #18171. Note, that it MAY break some cocos-lua based projects, but there is trivial fix: do not use `.lua` and `.luac` directly in `require`.


v1993:
@minggo take a look at this.

minggo:
Sorry, i am busy releasing v3.16. I will take a look ASAP.

minggo:
@v1993 this PR looks good, but i think we should not break compatibility in v3.

v1993:
And what now?

minggo:
What did you mean `what now`?

v1993:
Sorry for my engish, I meant what is your future actions with it?

minggo:
We may change it in v4, not for v3.

PR #18124, opened 2017-08-03
*fix android AudioEngine Crash

zoxin:
switch to background when sound effect is about to end, then switch to foreground again, there is a crash may be occurred

PR #18209, opened 2017-09-01
fix android AudioEngine Crash (#18124)

dumganhar:
Sync https://github.com/cocos2d/cocos2d-x/pull/18124

* *fix android AudioEngine Crash

switch to background when sound effect is about to end, it's maybe crashed when switch to foreground again

* use "emplace" instead of "insert"


minggo:
[ci rebuild]

PR #18221, opened 2017-09-06
Fixes crash of lua particle test while building lua-tests in release mode.

dumganhar:
fixed http://punchbox.info:3000/issues/26317


minggo:
@dumganhar  i think we should send it to v3.16 branch too.

dumganhar:
Yes, I forgot.

dumganhar:
@minggo ,it's here: https://github.com/cocos2d/cocos2d-x/pull/18222

minggo:
thanks

PR #18233, opened 2017-09-07
Make the sort behavior is same on 32bit and 64bit.

halx99:


cocos/2d/CCNode.h
```diff
@@ -950,8 +955,8 @@ class CC_DLL Node : public Ref
             return (n1->_localZOrderAndArrival < n2->_localZOrderAndArrival);
         });
 #else
-        std::stable_sort(std::begin(nodes), std::end(nodes), [](_T* n1, _T* n2) {
-            return n1->_localZOrder < n2->_localZOrder;
+        std::sort(std::begin(nodes), std::end(nodes), [](_T* n1, _T* n2) {
+            return n1->_localZOrder < n2->_localZOrder || (n1->_localZOrder == n2->_localZOrder && n1->_getOrderOfArrival() < n2->_getOrderOfArrival() );
```
minggo: The codes looks good, but i am not sure the performance. May be we should keep `_orderArrival` on 32bit.

halx99: Yes, direct access the variable more better.

minggo: Yep, the bug is imported to improve performance for 64bit devices, but we can not slow down 32bit devices.

halx99: Let's change something.

halx99:
@minggo please review the changes.

minggo:
Thanks, but please send to v3 branch since Node affect all codes, i think it is better to have more time to test it. We will release v3.16 soon.

halx99:
OK, do you known how to change this PR to v3 branch directly? I find it.

minggo:
Sorry i don't know what you mean `directly`, but you can go to https://github.com/halx99/cocos2d-x/tree/patch-1 and click `Pull request`.

halx99:
I change base of this PR to v3branch

halx99:
You also can see there is a common storage for 32bit & 64bit, ONLY sort code is different:
https://github.com/halx99/cocos2d-x/pull/35

minggo:
@halx99 i think this PR includes my other codes that are not related with this issue. I think you should checkout a branch based on v3 to do it. You can use `git cherry-pick` to pick commits your need.

halx99:
ok

minggo:
@halx99 this PR contains much unneeded codes, i will close it. Please send a new one, thanks.

PR #17786, opened 2017-05-05
Fix issue #15926

AIGRIND-LLC:

PR #18237, opened 2017-09-08
Fixes ShaderRetroEffect crashes.

dumganhar:
Fixed http://punchbox.info:3000/issues/26289


dumganhar:
@minggo, PR https://github.com/cocos2d/cocos2d-x/pull/17786 broke the compatibility.
Currently, Label::getLetter(i) will return nullptr if the letter is a space while it's non-nullptr before.

PR #18239, opened 2017-09-08
fix terrible code issue in getResponseHeader

zoxin:
called from lua (cc.XMLHttpRequest)

1037: auto iter = self->getHttpHeader().find(value); **//return iterator of temporary map**
1038: if (iter != self->getHttpHeader().end())            **// compare to another temporary map.**

it will crash on windows at least


dumganhar:
Good finding.
But better approach is probably returning `const reference` instead of `value`.
I mean:
```c++
std::map<std::string, std::string> getHttpHeader() { return _httpHeader ;}
```
->
```c++
const std::map<std::string, std::string>& getHttpHeader() { return _httpHeader ;}
```

dumganhar:
I sent another Pull Request to fix this issue:
https://github.com/cocos2d/cocos2d-x/pull/18252
Close this one. Thanks.

PR #18253, opened 2017-09-12
Fix RenderTexture DepthAndStencil issue on Android.

DavidDeSimone:
Fixing issue where Stencil and Depths buffers would not function for a render texture if you backgrounded an application, and then resumed.

When we background an application on Android, we destroy our FBO, and when we come back to the foreground we re-create it. However, we _only_ recreate the FBO. If our FBO had any depth or stencil renderbuffer attachments, we do not recreate/reattach them. This PR will fix this.

To recreate:

On Android, create a RenderTexture, set AutoDraw to true, add as a child a CCClippingNode that is clipping something.
Verify that the clipping node works.
Background the app.
Without this change, the stencil buffer will not be in the correct state, and you will see something incorrect. With this change, you will see the correct result.

This was tested on a galaxy s7, though I believe it should happen on all Android.

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -60,6 +60,7 @@ RenderTexture::RenderTexture()
 , _autoDraw(false)
 , _sprite(nullptr)
 , _saveFileCallback(nullptr)
+, _depthAndStencilFormat(0)
 {
 #if CC_ENABLE_CACHE_TEXTURE_DATA
     // Listen this event to save render texture before come to background.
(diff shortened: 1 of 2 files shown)


DavidDeSimone:
@minggo Do you think you'll have a chance to look at this post 3.16?

minggo:
We have tasted v3.16 for 3 rounds, so i don't think we have time to include it in v3.16.

DavidDeSimone:
That's reasonable, feel free to take a look whenever you have time. 👍

minggo:
> On Android, create a RenderTexture, set AutoDraw to true, add as a child a CCClippingNode that is clipping something.

why need to add as a child of ClippingNode?

DavidDeSimone:
Sorry, poor english on my part. What I ment was something like

``` c++
auto renderTexture = RenderTexture::create(...); // Have this set up with a stencil attachment
auto clippingNode = ClippingNode::create(...); // Have this set up to clip a sprite or something
renderTexture->addChild(clippingNode);
renderTexture->setAutoDraw(true); 
// Let it render, view it, and then background the app. 
```

minggo:
Could you please add a test case for it, then we can test it in future to prevent it.

minggo:
@DavidDeSimone will you please fix the conflict? I will merge it after resolving the conflict.

minggo (approved):

minggo:
I fixed the conflict and send a new PR in https://github.com/cocos2d/cocos2d-x/pull/18648

Issue #17096, opened 2017-01-02
Android 6.0 : TextFieldTTF (Text Input) Soft Keyboard in full screen


siarsky:
- cocos2d-x version: cocos2d-x-3.13.1
- devices test on: Phone Wiko3 (**Android 6.0**) and Tablet Nexus 7 (**Android 5.1.1**)
- developing environments
   - NDK version: NDK 4.9, android-19

Steps to Reproduce:
1. compile cppTests
2. run them on Android 5.* and 6.* device
3. run test 51 : Node : Text Input
4.  On **Android 5.1.1** the soft keyboard is **shown correctly** - does not cover the entire screen
5. On **Android 6.0** the soft keyboard appears **incorrectly** - in a full screen mode and covers entire screen

Could you please look into this and provide a workaround or a solution?

siarsky:
Dear cocos2dx team,

I found a similar problem in the forum:

[http://discuss.cocos2d-x.org/t/feature-request-disable-fullscreen-keyboard-on-android/19497](http://discuss.cocos2d-x.org/t/feature-request-disable-fullscreen-keyboard-on-android/19497)

any news to this issue? Could you reproduce it?

siarsky:
**HISTORY:**
An immersive mode was introduced in Android 4.4 (Kitkat), version code 19. This mode allows games to run in a full screen mode - which was then integrated into cocos2dx in a commit "hide virtual button by default" [8634773](https://github.com/cocos2d/cocos2d-x/commit/8634773be30d44184973341b5f91ab7bbdb41930#diff-248c9b86bb00b0acd8d41eb1aabbcf9c) on Jun 15, 2016 by minggo

This mode has an disadvantage - it changes the way a soft keyboard behaves. In a versions before KitKat the softkeyboard was in a non full screen mode, for a immersive mode, the default behavior is a full screen mode.

**SOLUTION:**
In Cocos2dxActivity.java two lines (43, 390) need to be added:
` 43     import android.view.inputmethod.EditorInfo;`
...
`389    Cocos2dxEditBox edittext = new Cocos2dxEditBox(this);`
`390    edittext.setImeOptions(edittext.getImeOptions() | EditorInfo.IME_FLAG_NO_EXTRACT_UI);`
`391     edittext.setLayoutParams(edittext_layout_params);`

[...]

PR #18270, opened 2017-09-14
Fix TextFieldTTF Soft Keyboard in full screen

siarsky:
TextFieldTTF Soft Keyboard in full screen #17096


minggo:
this PR contains much unneeded codes, so close it, please send a new one, thanks.

PR #18247, opened 2017-09-11
[android] Fixes Cocos2dxActivity is re-created which causes crash if launching app first time and re-opening from icon

dumganhar:
fixed https://github.com/cocos2d/cocos2d-x/issues/17052
and https://github.com/cocos2d/cocos2d-x/issues/16687
It's a bug of Android system, this is just a workaround.

PR #18271, opened 2017-09-15
[android] Fixes AppAcitivity (Sub class of Cocos2dxActivity) is re-created which causes crash if launching app first time and re-opening from icon.

dumganhar:
Continue to fix after https://github.com/cocos2d/cocos2d-x/pull/18247


minggo:
please fix compiling error: https://travis-ci.org/cocos2d/cocos2d-x/jobs/275721937

PR #18275, opened 2017-09-15
EventListenerTouchOneByOne can not bind any touch event.

jakitliang:
Such code on Xcode:
```
auto my_on_click_listener = cocos2d::EventListenerTouchOneByOne::create();
my_on_click_listener->onTouchBegan = [](cocos2d::Touch* touch, cocos2d::Event* event) {};
```
qq20170915-151845


minggo:
I think this PR contains much unneeded codes, so close this one. May be you should send to v3 branch instead of v3.16.

PR #18279, opened 2017-09-18
Fix VAO in CameraBackgroundBrush (#18278)

minggo:


minggo:
sync from https://github.com/cocos2d/cocos2d-x/pull/18278

PR #18281, opened 2017-09-18
Fix CameraBackgroundColorBrush cannot use alpha

sidepelican:
`static CameraBackgroundColorBrush* createColorBrush(const Color4F& color, float depth)` 
this accepts alpha contained color value, but the alpha value dose not apply, because of thats shader.

### DIFF
--- a/cocos/2d/CCCameraBackgroundBrush.cpp
+++ b/cocos/2d/CCCameraBackgroundBrush.cpp
@@ -279,9 +279,25 @@ bool CameraBackgroundColorBrush::init()
     return true;
 }
 
+void CameraBackgroundColorBrush::drawBackground(Camera* camera)
+{
+    GL::blendFunc(_blendFunc.src, _blendFunc.dst);
+
+    CameraBackgroundDepthBrush::drawBackground(camera);
+}
+
 void CameraBackgroundColorBrush::setColor(const Color4F& color)
 {
     _quad.bl.colors = _quad.br.colors = _quad.tl.colors = _quad.tr.colors = Color4B(color);
+    
+    if (color.a != 1.f)
+    {
+        _blendFunc = BlendFunc::ALPHA_NON_PREMULTIPLIED;
+    }
+    else
+    {
+        _blendFunc = BlendFunc::ALPHA_PREMULTIPLIED;
+    }
     
     if (_vertexBuffer)
     {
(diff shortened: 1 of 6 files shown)


minggo:
looks good, could you please add a test case?

sidepelican:
I found CameraBackgroundColorBrush is not working on the latest version.
(on iPhone6s iOS11.0 GM)
I add test but my test not work without restoring `CameraBackgroundDepthBrush::drawBackground` to commit https://github.com/cocos2d/cocos2d-x/commit/d07794052fed5c3edc29d4a60f99399d49271515

minggo:
CameraBackgroundColorBrush issue is fixed in https://github.com/cocos2d/cocos2d-x/pull/18283, i will sync v3.16 to v3 later.

sidepelican:
It works with  18283. Thank you!

and, I forgot to add BlendFunc setting for alpha used color. I fixed it.

sidepelican:
rebased.

sidepelican:
please review.

minggo:
thanks, i will review it ASAP

cocos/2d/CCCameraBackgroundBrush.h
```diff
@@ -197,6 +202,7 @@ class CC_DLL CameraBackgroundColorBrush : public CameraBackgroundDepthBrush
     
 protected:
     Color4F _color;
+    BlendFunc _blendFunc;
```
minggo: it is not initialized in constructor

cocos/2d/CCCameraBackgroundBrush.cpp
```diff
@@ -279,10 +279,26 @@ bool CameraBackgroundColorBrush::init()
...
     _quad.bl.colors = _quad.br.colors = _quad.tl.colors = _quad.tr.colors = Color4B(color);
     
+    if (color.a != 1.f)
+    {
+        _blendFunc = BlendFunc::ALPHA_NON_PREMULTIPLIED;
+    }
+    else
+    {
+        _blendFunc = BlendFunc::ALPHA_PREMULTIPLIED;
+    }
```
minggo: When alpha = 1.f, then `BlendFunc::ALPHA_PREMULTIPLIED` == `BlendFunc::ALPHA_NON_PREMULTIPLIED`.  So i think there is not need to set blend function here.

minggo: just initialize it in constructor.

sidepelican:
`_blendFunc` was not needed. removed.

PR #18297, opened 2017-09-20
Fix screen resolution issue on iOS

pandamicro:


dumganhar:
A lot of `ios_mac/ios/Default-portrait-568h@2x.png 100644 → 100755`
100644 -> 100755 ??? png files should not be executable.

pandamicro:
Sorry I don't have time to solve this PR, @dumganhar if you have time to refine it, thanks.
If not, maybe it can't be released in v3.16, then resolution issues on iPhone 6/7/plus will persist.

pandamicro:
The modifications of png files was due to the solution here

https://stackoverflow.com/questions/39652867/code-sign-error-in-macos-sierra-xcode-8-3-3-resource-fork-finder-information

before applying `find . -type f -name '*.png' -exec xattr -c {} \;` to templates folder, the compilation will fail on code sign error

pandamicro:
oops... this branch have been overwrote...

PR #18301, opened 2017-09-20
Make the sort behavior is same on 32bit and 64bit

halx99:


cocos/2d/CCNode.h
```diff
@@ -1940,8 +1940,23 @@ class CC_DLL Node : public Ref
     mutable bool _additionalTransformDirty; ///< transform dirty ?
     bool _transformUpdated;         ///< Whether or not the Transform object was updated since the last frame
 
-    std::int64_t _localZOrderAndArrival; /// cache, for 64bits compress optimize.
-    int _localZOrder; /// < Local order (relative to its siblings) used to sort the node
+#if CC_LITTLE_ENDIAN
```
minggo, line 1943: I am not confident with `CC_LITTLE_ENDIAN ` computation. And i am not confident with union usage here. Does `int` and `unsigned int` should be 32 bit on all platforms? I think it is better to use two variables here.

halx99: Actual use union structure, the implement source code is more clearly and easy, and the access code is same as original code(before optimize 64bit sort performance PR), As far as I know, all platforms: int is 32 bit. Or can use std::int32_t to replace it.

minggo: Yep, use `std::int32_t` is more safe.

cocos/base/ccMacros.h
```diff
@@ -281,6 +281,27 @@ CC_ASSERT(__gl_error_code == GL_NO_ERROR, "Error"); \
...
+  #endif
+#else
+  #include <sys/param.h>
+  #if defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN
+    #define CC_LITTLE_ENDIAN 1
+  #else
+    #define CC_LITTLE_ENDIAN 0
+  #endif
+#endif
+
```
minggo, line 305: I am not confident with these codes. Do you refer to any implementation?

halx99: It's from google protobuf

minggo: Got it, could you please add a link in codes?

minggo: add it as a comment

halx99: OK

cocos/2d/CCNode.cpp
```diff
@@ -56,7 +56,7 @@ THE SOFTWARE.
 NS_CC_BEGIN
 
 // FIXME:: Yes, nodes might have a sort problem once every 30 days if the game runs at 60 FPS and each frame sprites are reordered.
-unsigned int Node::s_globalOrderOfArrival = 0;
+std::uint32_t Node::s_globalOrderOfArrival = 0;
```
dumganhar, line 59: Is `std::` needed?

halx99: include ```<cstdint>``` then use std::int32_t std::int64_t
include ```<stdint.h>``` then use int32_t int64_t

minggo (approved):

minggo:
This PR looks good to me. @dumganhar do you have any question? If not, i will merge it.

stevetranby:
Seems to cause issues. May or may not be my usage pattern.
Nodes drawn in wrong order (iOS/Mac 64-bit builds).
I've reverted this change in my fork, but you may want to test/confirm that this works as intended.

mori1234:
ios 11 nodes drawn in worng order

minggo:
Thanks for feedback, we will invest into it.

halx99:
Let's test it

stevetranby:
I'll also give it another test. Thinking about it again it's possible I tweaked it due to signed/unsigned mismatch and maybe my slight change caused the ordering issues.

halx99:
Thanks for feedback

stevetranby:
Alright, so I haven't tested any further, but I did find that this code causes my draw order to get screwed up. I'm not sure if it's just reversed Z or what, but unless I'm mistaken it's due to CC_LITTLE_ENDIAN being defined wrongly for macOS and iOS (it should be enabled == 1, but it's 0).

So, I believe it's the detection of endianness that is wrong (or incomplete).

stevetranby:
When I force it then it appears to work as before applying this PR.
`#define CC_LITTLE_ENDIAN 1`

stevetranby:
Oh, to clarify my build: XCode9, clang (4.x, unsure exact ver), libc++ (llvm), std=c++14

Which hardware that cocos2d runs on is big-endian?
x86 is little endian: should cover majority of desktop win/mac/linux builds.
RaspberryPi is likely big-endian (most embeddeds seem to use this), but I'm not sure. 
I know Arm is supposed to support both endianness depending on some flag/setting, but the default appears to be little endian.
Maybe your endian check could still be used if the default is 1 and not 0?

If you want to improve the check to cover more hardware ...

There's always BOOST if one can figure out how to combine only needed parts into a endian only single header file.
http://www.boost.org/doc/libs/1_65_1/boost/detail/endian.hpp
http://www.boost.org/doc/libs/1_65_1/boost/predef/other/endian.h

Nothing's stb_ libs or its reference page to other single header libraries might have a more complete (and accurate) compile-time endian check. 
https://github.com/nothings/stb

Sourceforge has some info (I'm testing)
https://sourceforge.net/p/predef/wiki/Endianness/

stevetranby:
Ah, I didn't even register you pulled this from protobuf ... I wonder if it defaulting to big endian is due to its main use case being over the network (which is big-endian as standardized).

Here's my updated macro, I'm not sure I'll keep this PR even if it works, since our game was working just fine using the code before this PR is applied, but maybe I'll just add in #pragma error/warning into this endian check macro to advise me further.

```
#ifdef _MSC_VER
  #if defined(_M_IX86)
    #define CC_LITTLE_ENDIAN 1
    #warning CC_LITTLE_ENDIAN enabled
  #else
    #define CC_LITTLE_ENDIAN 0
    #warning CC_LITTLE_ENDIAN DISABLED!!!!
  #endif
  #if _MSC_VER >= 1300 && !defined(__INTEL_COMPILER)
    #pragma runtime_checks("c", off)
  #endif
#else
  #include <sys/param.h>
// STEVE: TODO: #include <endian.h>


  #if defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN
    #define CC_LITTLE_ENDIAN 1
    #warning CC_LITTLE_ENDIAN enabled
  #else

// STEVE: using info from - https://sourceforge.net/p/predef/wiki/Endianness/

    #if (defined(__BIG_ENDIAN__) || \
        defined(__ARMEB__) || \
        defined(__THUMBEB__) || \
        defined(__AARCH64EB__) || \
        defined(_MIPSEB) || \
        defined(__MIPSEB) || \
        defined(__MIPSEB__) || \
        (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__) || \
        (__FLOAT_WORD_ORDER__ == __ORDER_BIG_ENDIAN__))

                #define CC_LITTLE_ENDIAN 0
                #warning CC_LITTLE_ENDIAN DISABLED!!!!

    #elif defined(__LITTLE_ENDIAN__) || \
          defined(__ARMEL__) || \
          defined(__THUMBEL__) || \
          defined(__AARCH64EL__) || \
          defined(_MIPSEL) || \
          defined(__MIPSEL) || \
          defined(__MIPSEL__) || \
          (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__) || \
          (__FLOAT_WORD_ORDER__ == __ORDER_LITTLE_ENDIAN__)

                #define CC_LITTLE_ENDIAN 1
                #warning CC_LITTLE_ENDIAN enabled

#endif // if detect little endian
  #endif // __BYTE_ORDER defined
#endif // MSC_VER
```

stevetranby:
Also, not sure why they've added this ... 
```
#pragma runtime_checks("c", off)
```
I'm guessing it allows for silent endian swapping of some sort? 
I'll assume it's only for the code that follows, but if this header is included in all translation units then this is a somewhat bluntly disabling these checks?

**Anyway, I'll leave where you go from here up to you y'all.** 

I'm probably going to just revert back again (as I mentioned before) since the original version/code was working fine for us. I'll leave a note/warning for myself when testing windows and esoteric or older android devices.

stevetranby:
I've manually expanded this out if you want to compare which defines exist.
It actually looks like only `__LITTLE_ENDIAN__` (and corresponding for bigE) as an addition is probably needed to cover ios/mac.

Sorry if I've written too much here :D

```

#if defined(__BIG_ENDIAN__)
#warning defined(__BIG_ENDIAN__)
#endif //
#if defined(__ARMEB__)
#warning defined(__ARMEB__)
#endif //
#if defined(__THUMBEB__)
#warning defined(__THUMBEB__)
#endif //
#if defined(__AARCH64EB__)
#warning defined(__AARCH64EB__)
#endif //
#if defined(_MIPSEB)
#warning defined(_MIPSEB)
#endif //
#if defined(__MIPSEB)
#warning defined(__MIPSEB)
#endif //
#if defined(__MIPSEB__)
#warning defined(__MIPSEB__)
#endif //

#if defined(__LITTLE_ENDIAN__)
#warning defined(__LITTLE_ENDIAN__)
#endif //
#if defined(__ARMEL__)
#warning defined(__ARMEL__)
#endif //
#if defined(__THUMBEL__)
#warning defined(__THUMBEL__)
#endif //
#if defined(__AARCH64EL__)
#warning defined(__AARCH64EL__)
#endif //
#if defined(_MIPSEL)
#warning defined(_MIPSEL)
#endif //
#if defined(__MIPSEL)
#warning defined(__MIPSEL)
#endif //
#if defined(__MIPSEL__)
#warning defined(__MIPSEL__)
#endif //
#if defined(__BYTE_ORDER)
#warning defined(__BYTE_ORDER)
#endif //
#if defined(__FLOAT_WORD_ORDER__)
#warning defined(__FLOAT_WORD_ORDER__)
#endif //
```

halx99:
so, the latest version of google's protobuf little endian detect macro does not work on mac/ios?

stevetranby:
Correct. It sets `CC_LITTLE_ENDIAN` to 0 as it.

What's weird though is that after further testing shows that apparently it's that it doesn't detected `#if defined(__BYTE_ORDER)`, but it does detect ` == __LITTLE_ENDIAN`.

Could very well be my setup.

```
#if defined(__BYTE_ORDER) && (__BYTE_ORDER == __LITTLE_ENDIAN)
  // this is not compiled
  #warning defined(__BYTE_ORDER) && (__BYTE_ORDER == __LITTLE_ENDIAN)
#endif
#if defined(__BYTE_ORDER)
  // this is not compiled
  #warning defined(__BYTE_ORDER)
#endif
#if (__BYTE_ORDER == __LITTLE_ENDIAN)
  // this compiles, anything set here is compiled on ios/mac
  #warning (__BYTE_ORDER == __LITTLE_ENDIAN)  <=== shows warning at compile
#endif
```

stevetranby:
Well ... Doh! I should've totally thought to look at the protobuf github project.

This is a documented issue. Looks like you pulled from an older copy (or maybe from the cocos2d-x libs version).
https://github.com/google/protobuf/issues/137
https://github.com/google/protobuf/pull/139

Updated to:
```
//...
  #include <sys/param.h>   // __BYTE_ORDER
  #if ((defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)) || \
         (defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN)) 
    #define PROTOBUF_LITTLE_ENDIAN 1
  #endif
//...
```

cocos/2d/CCNode.h
stevetranby: Anonymous union appears to be GNU extension, though Clang appears to support it just fine. Not sure about MSVC.

halx99:
Could you send a PR to fix this macro detection?

stevetranby:
I believe you can just update your ccMacro.h to the following:
```
 /******************************************************************************************/
 /** LittleEndian Sense Macro, from google protobuf see:                                  **/
 /** https://github.com/google/protobuf/blob/master/src/google/protobuf/io/coded_stream.h **/
 /******************************************************************************************/
#ifdef _MSC_VER
  #if defined(_M_IX86)
    #define CC_LITTLE_ENDIAN 1
  #else
    #define CC_LITTLE_ENDIAN 0
  #endif
  #if _MSC_VER >= 1300 && !defined(__INTEL_COMPILER)
    #pragma runtime_checks("c", off)
  #endif
#else
  #include <sys/param.h>
  #if ((defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)) || \
         (defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN)) 
    #define CC_LITTLE_ENDIAN 1
  #else
    #define CC_LITTLE_ENDIAN 0
  #endif
#endif
```

halx99:
Complete.

halx99:
I have update my ccMacro.h.

minggo:
@halx99 could you please send a new PR to fix it.

halx99:
ok, see:  https://github.com/cocos2d/cocos2d-x/pull/18354

PR #18317, opened 2017-09-23
Fix accelerometer on Android on reverse landscape/portrait.

pyrosphere:
The accelerometer should adjust when the screen is on reverse landscape or reverse portrait (upside down). This is already done on iOS:

https://github.com/cocos2d/cocos2d-x/blob/b633260020cba97a4160eb2a6f81b270b424657b/cocos/platform/ios/CCDevice-ios.mm#L250-L266

This PR makes the accelerometer behave in the same way on Android too.


minggo:
can we handle it in orientation mode?

pyrosphere:
Orientation doesn't distinguish between landscape left and landscape right.

minggo:
good to know, thanks

PR #18322, opened 2017-09-25
Fix performance problems with RenderTexture on Non-iOS devices

### LINKED ISSUE #18321: Performance issues with RenderTexture on non-iOS devices (Android, Windows)
An application using RenderTexture, runs smoothly on iOS - rendering of 34 sprites takes ca. 200ms. The same operation takes 3420ms on Android.

A similar problem was observed on Windows Phone [http://discuss.cocos2d-x.org/t/solved-rendertexture-is-very-slow-on-3-7-versions/28055](http://discuss.cocos2d-x.org/t/solved-rendertexture-is-very-slow-on-3-7-versions/28055), where Yusel described an workaround : deactivation of the group render command in RenderTexture.

The group command was introduced in 3.7 - as an implementation of "Group command" from [Roadmap of Render Pipeline](http://www.cocos2d-x.org/wiki/Cocos2d_v30_renderer_pipeline_roadmap), which speed up a rendering performance.

I implemented Yusels workaround - and saw my application needs just 200 ms on Android, exactly as on iOS. I was interesting in finding out, where is the problem and so Yusels workaround was a base for my problem analyses, which discovered the following facts:
1. rendering of "X" sprites on RenderTexture will lead to inserting of "Y" commands into one render queue assigned to the group - which is a goal of group command. Those "Y" commands render all "X" objects.
2. the group command itself is added "X" times into render queue - which is a problem, it should be added just once
3. result: rendering of RenderTexture writes "X"*"Y" times into GPU instead of just "Y" times
4. iOS has apparently kind of "smart filter" over the GPU, which filters/ignores exactly same openGL operations, Android and Windows phone does not have it and therefore the rendering is very slow.

This problem might occur in all classes using Group Command for rendering:
CCClippingNode.cpp
CCNodeGrid.cpp
CCRenderTexture.cpp
UILayout.cpp.cpp
(I just searched for "addCommand(&_groupCommand")

Due to fact, that I am using just RenderTexture, I am not sure.

SOLUTION:
Make sure group render command will be added just one time and not "X" times for "X" sprites.

I've implemented a workaround, which is added into this case as a possible solution. I wrote it in general way and therefore I needed a function from Render as well.

Maybe a better/smarter solution exists...

CONCLUSION:
Any application using RenderTexture on a non iOS operational system is lagging. I believe a quick fix should be introduced to avoid bugs with title "Android performance issue".

siarsky:
Fix #18321 : performance problems with RenderTexture on Android

### DIFF
--- a/cocos/renderer/CCRenderer.h
+++ b/cocos/renderer/CCRenderer.h
@@ -90,6 +90,8 @@ public:
 public:
     /**Constructor.*/
     RenderQueue();
+    /**Get a queue type*/
+    RenderQueue::QUEUE_GROUP getQueueGroup(RenderCommand* command);
     /**Push a renderCommand into current renderqueue.*/
     void push_back(RenderCommand* command);
     /**Return the number of render commands.*/
@@ -182,6 +184,8 @@ public:
 
     /** set color for clear screen */
     void setClearColor(const Color4F& clearColor);
+    /* returns the Render Queue */
+    std::vector<RenderQueue>& getRenderGroups() { return _renderGroups; }
     /* returns the number of drawn batches in the last frame */
     ssize_t getDrawnBatches() const { return _drawnBatches; }
     /* RenderCommands (except) TrianglesCommand should update this value */
--- a/cocos/renderer/CCRenderer.h
+++ b/cocos/renderer/CCRenderer.h
@@ -31,6 +31,7 @@
 
 #include "platform/CCPlatformMacros.h"
 #include "renderer/CCRenderCommand.h"
+#include "renderer/CCGroupCommand.h"
 #include "renderer/CCGLProgram.h"
 #include "platform/CCGL.h"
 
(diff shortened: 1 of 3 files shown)


cocos/renderer/CCRenderer.h
```diff
@@ -182,6 +184,8 @@ class CC_DLL Renderer
 
     /** set color for clear screen */
     void setClearColor(const Color4F& clearColor);
+    /* returns the Render Queue */
+    std::vector<RenderQueue>& getRenderGroups() { return _renderGroups; }
```
DavidDeSimone: I don't like the idea of just handing out a mutable reference of the render groups to the world, I think it's poor encapsulation. Can we move the logic you've added to CCRenderTexture into a helper function in CCRenderer? I think that will be better encapsulated.

siarsky:
Thanks David for the review - a good point.

PR #18324, opened 2017-09-26
Fix for Non-TTF Label Shadow, it makes it follow label opacity

SBJoker:
Bug: If FadeTo action applied to Label with non-TTF font with shadow, shadow stay opaque and doesn't fade as label.

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -1505,7 +1505,7 @@ void Label::onDrawShadow(GLProgram* glProgram, const Color4F& shadowColor)
     {
         Color3B oldColor = _realColor;
         GLubyte oldOPacity = _displayedOpacity;
-        _displayedOpacity = shadowColor.a * 255;
+        _displayedOpacity = shadowColor.a * (oldOPacity / 255.0f) * 255;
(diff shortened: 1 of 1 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -1505,7 +1505,7 @@ void Label::onDrawShadow(GLProgram* glProgram, const Color4F& shadowColor)
     {
         Color3B oldColor = _realColor;
         GLubyte oldOPacity = _displayedOpacity;
-        _displayedOpacity = shadowColor.a * 255;
```
minggo, line 1508: Why do it like this? `shadowColor.a` is already a float

SBJoker: I don't understand a question. shadowColor.a is a float, definitely.

minggo: Sorry about it. I misunderstood it. Could you please provide a test case to reproduce it?

SBJoker: Just create a Label using Label::createWithBMFont, enable shadow for it. Run action FadeOut for it.
You can see label fadeout, but shadow doesn't fade.

PR #18339, opened 2017-10-02
Support for using CCRef as a std::shared_ptr

jordo:
Often it is desirable to use a cocos2d::Ref in c++ lambda functions.  However the c++ shared memory model std::shared_ptr works fundamentally different than the cocos2d one (retain, release, etc).  Often I find I want to use lambda functions with a cocos2d::Ref in a capture list, but the memory model falls apart at this point.  This addition adds support to cocos2d::Ref that allows the object to be maintained by both the cocos2d-x memory model and the c++ std::shared_ptr model.  

The following code works great when copying this request object around as copy semantics preserve the memory of the cocos2d::Ref in the capture list:

```
auto async_request = AsyncHttpRequest.MakeRequest("http://www.example.com", [shared_label = this->label->shared_ptr<cocos2d::Label>()](bool did_succeed) {

	if(did_succeed)
		shared_label.get()->setString("success");
	else
		shared_label.get()->setString("failed");
});
```

In this example, the cocos2d::label in the capture list will only get deallocated when BOTH it's cocos2d-x retain count reaches zero, AND when the shared_ptr reference count reaches zero as well.

### DIFF
--- a/cocos/base/CCRef.h
+++ b/cocos/base/CCRef.h
@@ -129,6 +129,7 @@ public:
      *
      * @returns A shared_ptr of this as a Ref.
      * @js NA
+     * @lua NA
      */
     std::shared_ptr<cocos2d::Ref> shared_ptr() {
         if(!_self_weak_ref.expired()) {
@@ -150,6 +151,7 @@ public:
      *
      * @returns A shared_ptr of this type template.
      * @js NA
+     * @lua NA
      */
     template <class T> std::shared_ptr<T> shared_ptr() {
         if(!_self_weak_ref.expired()) {
--- a/cocos/base/CCRef.h
+++ b/cocos/base/CCRef.h
@@ -28,6 +28,7 @@ THE SOFTWARE.
 
 #include "platform/CCPlatformMacros.h"
 #include "base/ccConfig.h"
+#include <memory>
 
 #define CC_REF_LEAK_DETECTION 0
 
--- a/cocos/base/CCRef.h
+++ b/cocos/base/CCRef.h
@@ -123,6 +123,49 @@ public:
      * @js NA
      */
     unsigned int getReferenceCount() const;
+    
+    /**
+     * Returns a shared_ptr of this Ref.
+     *
+     * @returns A shared_ptr of this as a Ref.
+     * @js NA
+     */
+    std::shared_ptr<cocos2d::Ref> shared_ptr() {
+        if(!_self_weak_ref.expired()) {
+            return _self_weak_ref.lock();
+        }
+        
+        auto shared_ptr = std::shared_ptr<cocos2d::Ref>(this, [=](cocos2d::Ref* ref)
+        {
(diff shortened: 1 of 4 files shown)


cocos/base/CCRef.h
```diff
@@ -123,6 +123,51 @@ class CC_DLL Ref
...
+     * @returns A shared_ptr of this type template.
+     * @js NA
+     * @lua NA
+     */
+    template <class T> std::shared_ptr<T> shared_ptr() {
+        if(!_self_weak_ref.expired()) {
+            return std::static_pointer_cast<T>(_self_weak_ref.lock());
+        }
+        
+        auto shared_ptr = std::shared_ptr<cocos2d::Ref>(this, [=](cocos2d::Ref* ref)
```
DavidDeSimone, line 162: Do we need capture on this lambda? Can't it just be [] (cocos2d::Ref* ref) ...

DavidDeSimone:
This seems like a solid addition to me. It would be nice if we could have some tests added to test the cases that

- We delete `this` from the `is_expired` case
- We delete `this` from the destructor of the shared ptr.

cocos/base/CCRef.h
jordo: Yes, it's superfluous.  I just did for readability.

jordo:
Yeah, it should be pretty trivial to add tests for both deallocation cases...  I haven't gotten around to it, but we've been using it in a couple games and I've seen both cases handled.

summerinsects:
Maybe you can use `cocos2d::RefPtr<T>` instead, and it is a similar to `std::shared_ptr`
It is easier to use function `makeRef`

In the case of your code, use it like this:
```c++
auto captrued_label = makeRef(this->label);
auto async_request = AsyncHttpRequest.MakeRequest("http://www.example.com",
    [captrued_label](bool did_succeed) {
        if(did_succeed)
            captrued_label->setString("success");
        else
            captrued_label->setString("failed");
});
````
or
```c++
auto thiz = makeRef(this);
auto async_request = AsyncHttpRequest.MakeRequest("http://www.example.com",
    [thiz](bool did_succeed) {
        if (thiz->getReferenceCount() < 2) return;  // not in the render tree
        if(did_succeed)
            thiz->label->setString("success");
        else
            thiz->label->setString("failed");
});
````

File: cocos/base/ZipUtils.cpp [hunk 2]
```diff
@@ -23,6 +23,9 @@
  THE SOFTWARE.
  ****************************************************************************/
 
+// FIXME: hack, this needs to be before minizip/unzip...
+#include <string>
```
DavidDeSimone, line 27: Why do we need to be touching header files in ZipUtils? I don't understand why this PR would need to introduce this change.

jordo: Sorry you can close this pull request...  cocos2d::RefPtr seems a sufficient wrapper class to providing similar functionality.

PR #18356, opened 2017-10-09
improvements for StringUtils::StringUTF8

summerinsects:
1. add 'const'
2. add range support for 'getAsCharSequence'

### DIFF
--- a/cocos/base/ccUTF8.cpp
+++ b/cocos/base/ccUTF8.cpp
@@ -27,6 +27,7 @@
 #include "platform/CCCommon.h"
 #include "base/CCConsole.h"
 #include "ConvertUTF.h"
+#include <limits>
 
 NS_CC_BEGIN
 
@@ -378,11 +379,21 @@ void StringUTF8::replace(const std::string& newStr)
     }
 }
 
-std::string StringUTF8::getAsCharSequence(std::size_t pos/* = 0*/, std::size_t len/* = (std::size_t)-1*/) const
+std::string StringUTF8::getAsCharSequence() const
+{
+    return getAsCharSequence(0, std::numeric_limits<std::size_t>::max());
+}
+
+std::string StringUTF8::getAsCharSequence(std::size_t pos) const
+{
+    return getAsCharSequence(pos, std::numeric_limits<std::size_t>::max());
+}
+
+std::string StringUTF8::getAsCharSequence(std::size_t pos, std::size_t len) const
 {
     std::string charSequence;
     size_t maxLen = _str.size() - pos;
-    if (len == (size_t)-1 || len > maxLen)
+    if (len > maxLen)
     {
         len = maxLen;
     }
--- a/cocos/base/ccUTF8.cpp
+++ b/cocos/base/ccUTF8.cpp
@@ -378,13 +378,19 @@ void StringUTF8::replace(const std::string& newStr)
     }
 }
 
-std::string StringUTF8::getAsCharSequence() const
+std::string StringUTF8::getAsCharSequence(std::size_t pos/* = 0*/, std::size_t len/* = (std::size_t)-1*/) const
 {
     std::string charSequence;
-
-    for (auto& charUtf8 : _str)
+    size_t maxLen = _str.size() - pos;
+    if (len == (size_t)-1 || len > maxLen)
     {
-        charSequence.append(charUtf8._char);
+        len = maxLen;
+    }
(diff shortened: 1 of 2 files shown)


cocos/base/ccUTF8.h
```diff
@@ -207,13 +207,14 @@ class CC_DLL StringUTF8
     std::size_t length() const;
     void replace(const std::string& newStr);
 
-    std::string getAsCharSequence() const;
+    std::string getAsCharSequence(std::size_t pos = 0, std::size_t len = (std::size_t)-1) const;
```
perminovVS: Please, use C++ coding style for cocos2d-x
https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#default-arguments
Not use default arguments, add new function.

perminovVS (changes requested):

cocos/base/ccUTF8.h
summerinsects: OK

cocos/base/ccUTF8.h
```diff
@@ -196,7 +196,7 @@ class CC_DLL StringUTF8
     struct CharUTF8
     {
         std::string _char;
-        bool isAnsi() { return _char.size() == 1; }
+        bool isAnsi() const { return _char.size() == 1; }
```
minggo: utf8 chars' size may be 1 too, right?

summerinsects: chars' size is 1 for ansi char, maybe great than 1 other cases

minggo: Yep, what i mean is that

* size > 1, not ansi
* size = 1, may be ansi or not

so can not use it to determine if it is ansi or not

summerinsects: so the function name is not accurate. em, this is the old code,
could you give some advice about it?

minggo: I didn't notice that it is old code. I have not idea too.

cocos/base/ccUTF8.cpp
```diff
@@ -380,11 +381,27 @@ void StringUTF8::replace(const std::string& newStr)
...
+{
     std::string charSequence;
+    size_t maxLen = _str.size() - pos;
+    if (len > maxLen)
+    {
+        len = maxLen;
+    }
 
-    for (auto& charUtf8 : _str)
+    len += pos;
```
minggo: I think it is better to use another variable to make it more clear, such as 
```c++
auto endPos = pos + len;
```

summerinsects: maybe `endPos = _str.size();` is better. I modify it

cocos/base/ccUTF8.cpp
```diff
@@ -380,11 +381,27 @@ void StringUTF8::replace(const std::string& newStr)
...
+{
     std::string charSequence;
+    std::size_t maxLen = _str.size() - pos;
+    if (len > maxLen)
+    {
+        len = maxLen;
+    }
 
-    for (auto& charUtf8 : _str)
+    std::size_t endPos = _str.size();
```
minggo: endPos is related to len

summerinsects: I am sorry. I fixed it

minggo (approved):

cocos/base/ccUTF8.h
perminovVS: StringUTF8 it my api, for ease work cursor in TextField and other.
See specification UTF8 https://tools.ietf.org/html/rfc3629
UTF-8, the object of this memo, has a one-octet encoding unit.  It
   uses all bits of an octet, but has the quality of preserving the full
   US-ASCII [US-ASCII] range: US-ASCII characters are encoded in one
   octet having the normal US-ASCII value, and any octet with such a
   value can only stand for a US-ASCIIUTF-8, the object of this memo, has a one-octet encoding unit.  It
   uses all bits of an octet, but has the quality of preserving the full
   US-ASCII [US-ASCII] range: US-ASCII characters are encoded in one
   octet having the normal US-ASCII value, and any octet with such a
   value can only stand for a US-ASCII character, and nothing else. character, and nothing else.

Right name it function - isASCII.
In Russian ASCII encoding for Window named alias - ansi.

Sorry for the confusion.
Possible name:
isASCII
isOneByteChar
isOneByte

PR #18357, opened 2017-10-10
log, use correct reserved bytes for append char.

halx99:


minggo:
I don't quite understand why modify like this. Could you please describe it in more detail?

halx99:
see the api description: http://www.cplusplus.com/reference/cstdio/vsnprintf/
I think 2 is enough.

halx99:
Such as: 
bufferSize is 18, then vsnprintf n=18-2=16
1. The charactor count of source string is 15, then vsnprintf will write 16 bytes with null-terminated charactor, then the final required buffer bytes is 17 bytes
2. The charactor count of source string is 16, then vsnprintf also will write 16bytes without null-terminated charactor, then the final required buffer bytes is 18 bytes.

minggo:
I think just need to use `bufferSize`, because 

```
n
Maximum number of bytes to be used in the buffer.
The generated string has a length of at most n-1, leaving space for the additional terminating null character.
size_t is an unsigned integral type.
```

halx99:
Yes, but we need reserved 2 bytes for additional characters '\n'

minggo:
good finding, but why `\n` takes 2 bytes?

halx99:
Because we need append '\0' manually, it's need 2 bytes at follow situation:
image

minggo:
yep, then i think `bufferSize - 1` is enough, because `vsnprintf` will `The generated string has a length of at most n-1, leaving space for the additional terminating null character.
size_t is an unsigned integral type.`. It means `vsnprintf` already leaves space for `\0`.

halx99:
I have test it at win32 & ubuntu linux: actual when the n is equals as source string count(without '\0'), it's also write n bytes to buffer.
see the return value descrption: ```The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character.```

halx99:
Test code:
```
#include <stdio.h>
#include <string.h>

int main(int, char**)
{
    char* buffer = new char[6];
    memset(buffer, 'f', 6);
    int n = snprintf(buffer,5, "hello");
    printf("n=%d\n", n);
    printf("last=%02x\n", buffer[5]);
    n = snprintf(buffer,6, "hello");
    printf("n=%d\n", n);
    printf("last=%02x\n", buffer[5]);

    return 0;
}
```

test result:
n=5
last=66
n=5
last=00

minggo:
I misunderstood your description and the test code. The test code seems it wrote 6 bytes and the last byte is `\0`, that's the expected behavior.

halx99:
please see again, there are two testcase:
when I set n = 5, it's wrote 5 bytes.

minggo:
`int n = snprintf(buffer, 4, "hello");`, n is `5` too. Tested on macOS, it is strange.

halx99:
really? Let me test on ubuntu

minggo:
The doc says

```
The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character.
If an encoding error occurs, a negative number is returned.
Notice that only when this returned value is non-negative and less than n, the string has been completely written.
```

so the return value is always 5, but only when `5 < n` means it succeed.

halx99:
Yes, so the implement of snprintf is same on VS2017, macOS, ubuntu g++

halx99:
I have test _snprintf on vs2017, if n isn't enough, the return value will be -1

halx99:
https://github.com/cocos2d/cocos2d-x/pull/18040

minggo:
As it uses `MAX_LOG_LENGTH` here, so i think should discard log that exceed `MAX_LOG_LENGTH`. Or we should use a loop to print all log, but each time should not exceed `MAX_LOG_LENGTH`.

summerinsects:
Hey here
There is a BUG in function `void _log(const char *format, va_list args)` ,
line `nret = vsnprintf(buf, bufferSize - 3, format, args);` which is about re-use a `va_list`
Because `MAX_LOG_LENGTH ` is large enough, the buf does not need to be reallocated usually, so the BUG is not easy to reproduce.

We can use the test code below:
```c++
#include <stdio.h>
#include <stdarg.h>

void foo1(const char *fmt, va_list args) {
    vprintf(fmt, args);
    puts("");
    vprintf(fmt, args);
    puts("");
}

void foo(const char *fmt, ...) {
    va_list args;
    va_start(args, fmt);
    foo1(fmt, args);
    va_end(args);
}


int main() {
    foo("%d %d", 1, 2);  // my output (xcode): 1 2 0 1606416462, it reads the junk data when re-use va_list
    return 0;
}
```

halx99:
Actual, some developer need print large than MAX_LOG_LENGTH, so a common useable log is ok.

halx99:
This macro should renamed to be INIT_LOG_LENGTH, maybe 95% logs is little length, so the length should be 512 or 1K is ok.

halx99:
@summerinsects  which platform do you test, I paste your test code at Visual Studio 2017, the output data is seems ok.

halx99:
@summerinsects Yes, you are right, I have test your code on ubuntu linux system, and the result is junk.

summerinsects:
@halx99  I think the right way is:
```c++
va_list ap;
va_copy(ap, args);
nret = vsnprintf(buf, bufferSize - 3, format, ap);
va_end(ap);
```

halx99:
Yes, you are right, great report, thanks.

minggo:
close it since https://github.com/cocos2d/cocos2d-x/pull/18426 is merged

PR #18382, opened 2017-10-19
Updated spine runtimes

piotrros:
Please take a look if I've correctly fixed this file:

cocos/scripting/js-bindings/manual/spine/jsb_cocos2dx_spine_manual.cpp

also important note: tests will crash, when trying to load binary skeleton. You have to save it with the latest Spine! They've broken binary skeletons compatibility.


minggo:
what's the spine runtime version does this PR upgrade to?

piotrros:
It's 3.6.

minggo:
There is conflict, and i think i should update the runtime myself.

PR #18389, opened 2017-10-20
Don't use fixed gcc version on install-deps-linux

cirosantilli:
- replacing the user's main GCC with update-alternatives is too invasive
- the default GCC 6.3 of Ubuntu 17.04 also works
- we should support whichever GCC is present in the supported Ubuntu version by default


cirosantilli:
Ah OK, the Travis uses the ancient GCC 4.8 which does not support some required c++ feature, and this upgrades to... 4.9.

So instead, this should be moved to https://github.com/cocos2d/cocos2d-x/tree/79b97e99920eb0ab9bb8671a84beb22077136126/tools/travis-scripts until travis finally moves to 16.04 https://github.com/travis-ci/travis-ci/issues/5821

Issue #18392, opened 2017-10-21
Android.mk is not self-contained


rynkowsg:
VERSION
---------
v3

PROBLEM
----------

From the perspective of a developer that link his target with
`cocos2dx_static` library, it should specify in his `Android.mk` only
a path to a makefile that defines that module, by

a) using `import-add-path` and `import-module`

    $(call import-add-path, <path-to-cocos-root>)
    $(call import-module, cocos)

  or ...

b) using directly an `include` command

    include <path-to-cocos-root>/cocos/Android.mk

Unfortunately the current configuration of cocos' Android.mk doesn't
specify (using `import-add-path`) locations of modules that it uses.
In result, every dev that uses cocos to create Android app has to
populate NDK_MODULE_PATH at his Android.mk files by adding two lines:

    $(call import-add-path, <path-to-cocos-root>/cocos)
    $(call import-add-path, <path-to-cocos-root>/external)

Those lines have to be placed at cocos/Android.mk file.

The current configuration of build script breaks a good practice of
hiding internal details from clients of library or component. In fact
the script exposes internal dependencies and require from a dev to make
them out.

---

Steps to Reproduce:

1. Generate cocos2d-x app.
2. Use in Android app the `cocos2dx_static` library by importing it's definition in one of described above ways.
3. You will get a build script issue like:

        Android NDK: jni/../../cocos2d/cocos/Android.mk: Cannot find module with tag 'freetype2/prebuilt/android' in import path
        Android NDK: Are you sure your NDK_MODULE_PATH variable is properly defined ?

minggo:
`path-to-cocos-root` is added into NDK_MODULE_PATH by default in cocos command. And i still don't understand how to reproduce the issue.

rynkowsg:
@minggo yep, that proofs what I said. `cocos` command sets `NDK_MODULE_PATH` to provide paths for cocos modules. Problem is that everyone who doesn't use `cocos` command, had to provide the missing paths for modules imported inside of `cocos/Android.mk` in his mk files, by adding lines like:

    $(call import-add-path, <path-to-cocos-root>/cocos)
    $(call import-add-path, <path-to-cocos-root>/external)

_(actually in cocos templates, as I checked just now, you also add the above commands, probably to support Android Studio integration)_

Very sorry if my previous description how to reproduce was not clear.

1. Create an example cocos project.

2. Go to Android.mk file of your project and check:
 
[...]

PR #18405, opened 2017-10-25
Android project update

minggo:
According to the description of https://github.com/cocos2d/cocos2d-x/issues/18392, don't have to set NDK_MODULE_PATH in cocos and project settings.


dumganhar:
Awesome.

PR #18416, opened 2017-10-28
Fix/Optimize clean and sort of render groups.

fulvioabrahao:
This fix optimize how group commands works inside cocos2d render pipeline. Before this fix, all instances of a group command create a render queue, add him in `std::vector<RenderQueue> _renderGroups` of `Renderer`, and it never will be removed from it. Because of that, i removed the insertion from render group constructor, and it will be add only if this command will be used to render scene. Another optimization, is only iterate over render queues those in use.

### DIFF
--- a/cocos/renderer/CCGroupCommand.cpp
+++ b/cocos/renderer/CCGroupCommand.cpp
@@ -31,7 +31,7 @@ NS_CC_BEGIN
 
 GroupCommandManager::GroupCommandManager()
 {
-
+    init();
 }
 
 GroupCommandManager::~GroupCommandManager()
@@ -43,6 +43,7 @@ bool GroupCommandManager::init()
 {
     //0 is the default render group
     _groupMapping[0] = true;
+    _usedIDs.insert(0);
     return true;
 }
 
@@ -54,6 +55,7 @@ int GroupCommandManager::getGroupID()
         int groupID = *_unusedIDs.rbegin();
         _unusedIDs.pop_back();
         _groupMapping[groupID] = true;
+        _usedIDs.insert(groupID);
         return groupID;
     }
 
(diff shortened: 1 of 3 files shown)


Dimon4eg:
Hi, how much performance is increased now? Did you run some tests?

cocos/renderer/CCGroupCommand.cpp
```diff
@@ -54,27 +55,31 @@ int GroupCommandManager::getGroupID()
...
+        _groupMapping[groupID] = false;
+        _usedIDs.erase(groupID);
+        _unusedIDs.push_back(groupID);
+    }
 }
 
 GroupCommand::GroupCommand()
 {
     _type = RenderCommand::Type::GROUP_COMMAND;
-    _renderQueueID = Director::getInstance()->getRenderer()->getGroupCommandManager()->getGroupID();
```
minggo, line 77: what happened if GroupCommand::init() is not invoked?

cocos/renderer/CCRenderer.cpp
```diff
@@ -640,9 +640,8 @@ void Renderer::render()
     {
         //Process render commands
         //1. Sort render commands based on ID
-        for (auto &renderqueue : _renderGroups)
-        {
-            renderqueue.sort();
+        for (auto renderQueueID : _groupCommandManager->getUsedIDS()) {
+            _renderGroups[renderQueueID].sort();
```
minggo, line 644: Yep, it reduces some sort() invoking for empty vector, but it may bring some performance issue to invoke `_groupCommandManager->getUsedIDS()`.

So i don't know what's the performance improved. And it make things complex.

cocos/renderer/CCGroupCommand.cpp
fulvioabrahao: If GroupCommand::init() is not invoked implies that command not receive a _renderQueueID and game crashes :/ ... But i think that is not correct to a GroupCommand receive a _renderQueueID in GroupCommand::GroupCommand(). Suppose a scene with 250 ui::Layout with clipping disabled, ui::Layout has a instance of a GroupCommand, but it is used only when Layout clipping is enabled, the old code will create 250 RenderQueue and it will never be deleted! In ever frame the code calls .sort() and .clear() over all RenderGroups forever...

minggo: May be should add assert in `GroupCommand::getRenderQueueID()` to ask to invoke `GroupCommand::init()` if render queue id is not valid.

minggo: If not, then it is hard to detect the crash error.

PR #18417, opened 2017-10-28
Add skeleton animation cache.

fulvioabrahao:


cocos/editor-support/spine/SkeletonAnimation.h
```diff
@@ -50,11 +50,13 @@ class SkeletonAnimation: public SkeletonRenderer {
 public:
 	CREATE_FUNC(SkeletonAnimation);
 	static SkeletonAnimation* createWithData (spSkeletonData* skeletonData, bool ownsSkeletonData = false);
+    static SkeletonAnimation* createWithJsonFileCached (const std::string& skeletonDataFile, const std::string& atlasFile, float scale);
```
Dimon4eg, line 53: Please fix code formatting. Use the same whitespaces, in this case tabs.

Dimon4eg (changes requested):
Please fix formatting

cocos/editor-support/spine/SkeletonAnimation.cpp
```diff
@@ -69,6 +72,27 @@ static _TrackEntryListeners* getListeners (spTrackEntry* entry) {
     
 //
 
+SkeletonAnimation* SkeletonAnimation::createWithJsonFileCached (const std::string& skeletonDataFile, const std::string& atlasFile, float scale) {
+    spSkeletonData* skeletonData = NULL;
+    auto iterator = skeletonDataMap.find(skeletonDataFile);
+    if (iterator == skeletonDataMap.end()) {
+        spAtlas* _atlas = spAtlas_createFromFile(atlasFile.c_str(), 0);
```
Dimon4eg, line 79: Better to check _atlas for null if no file for example.

cocos/editor-support/spine/SkeletonAnimation.cpp
```diff
@@ -40,6 +40,9 @@ using std::vector;
 
 namespace spine {
 
+static std::unordered_map<std::string, spSkeletonData*> skeletonDataMap;
+static std::vector<spAtlas*> atlases;
```
Dimon4eg, line 44: I think better to create a skeleton cache file structure to hold both spSkeletonData* and spAtlas*.
In this case only one map collection will be.
Pros:
- easier deleting (in only one for)
- easier to delete only one element if in future it will be needed

Dimon4eg (commented):
Better to use cache structure

minggo:
please send the PR to Spine runtime too, according to its license, cocos2d-x can not modify the code.

halx99:
@fulvioabrahao I have send a seem PR to spine runtime, but they don't think it shoud added to "Spine runtime", see here: https://github.com/EsotericSoftware/spine-runtimes/pull/1020

So I added it to cocos2d-x engine: https://github.com/halx99/x-studio365/blob/master/editor-support/cocostudio/ext/SpineSkeletonDataCache.h

minggo:
@halx99 thanks for your feedback, so i have to close this PR.

PR #18418, opened 2017-10-30
Remove side-effects from CCB value parser

sp-jordi-vilalta:
The parsing of position and scale have some side effects: instead of just parsing the value and returning it so it can be applied later (like the rest of parse methods do), these also actually set the values.
It seems to be redundant because just after the parsing is done, the actual functions that apply the values are called. In addition, it has the undesired side effect that if a custom loader wants to ignore the parsed values, it can't be achieved because the values are already applied when the loader gets them.
This PR just removes these extra applications.

### DIFF
--- a/cocos/editor-support/cocosbuilder/CCNodeLoader.cpp
+++ b/cocos/editor-support/cocosbuilder/CCNodeLoader.cpp
@@ -382,7 +382,6 @@ Vec2 NodeLoader::parsePropTypePosition(Node * pNode, Node * pParent, CCBReader *
     Size containerSize = ccbReader->getAnimationManager()->getContainerSize(pParent);
     
     Vec2 pt = getAbsolutePosition(Vec2(x,y), type, containerSize, pPropertyName);
-    pNode->setPosition(pt);
     
     if (ccbReader->getAnimatedProperties()->find(pPropertyName) != ccbReader->getAnimatedProperties()->end())
     {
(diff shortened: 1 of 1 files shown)


minggo:
I think i will break compatibility.

sp-jordi-vilalta:
Hi, @minggo: could you please elaborate on what can it break?
I'm planning to use this change locally and I'd like to know what are the possible implications. In case it breaks something, shouldn't that be fixed instead? It feels wrong the way it is right now.
Thanks.

minggo:
Sorry, i am not familiar with the codes, and i close it because you don't response to my comment. So will it break something? Since Cocos Studio is cancelled, so i don't want to put energy on it.

sp-jordi-vilalta:
I'm not very familiar with the code, but from what I've seen in NodeLoader::parseProperties(), for each property type, the code is structured like this:
```
value = parsePropType();
onHandlePropType(value);
```
All the "parse*" functions just parse the value and return it, and let the "onHandlePropType*" functions apply it. The value can be applied differently depending on the concrete loader, and even depending on the property name.
In the particular cases of the position and scale, the value is applied in the parse functions (and later it gets applied again in the onHandlePropType* function), without even taking into account the property name. For instance, if someone creates a custom property of type "position", that value would be applied automatically in the parse function without letting the loader choose what to do with it, which feels just wrong.
I don't know if there's any code that relies on these values getting applied in the parse functions, but I think that should be fixed instead.

sp-jordi-vilalta:
And in case it really has to be kept like this for some compatibility reason, a comment in the code would be really appreciated ;)

PR #18425, opened 2017-11-01
Improve StringUtils::format implementation.

halx99:


summerinsects:
I suggest calling `shrink_to_fit` before return

halx99:
I don't think so, the shrink_to_fit maybe trigger memory reallcoate & copy.

drelaptop:
@halx99 https://github.com/cocos2d/cocos2d-x/issues/18724 is a `StringUtils::format` issue, would you like to solve that? I think this PR and that issue is related.

halx99:
@drelaptop ok, let me see.

PR #18348, opened 2017-10-06
Refine jni helper (final version)

summerinsects:
1. refactor all `JniHelper::callStaticXXXMethod` in one template `callStaticMethod`
2. deduct JNI signature during **compiling time**
3. add  support for `std::vector` as parameters and return value

PR #18433, opened 2017-11-02
Refine jni helper

summerinsects:
#18348 
1. deduct JNI signature during **compiling time**
2. refactor all JniHelper::callStaticXXXMethod in one template callStaticMethod, but not break compatibility
3. add support for `std::vector` as parameters and return value

### DIFF
--- a/cocos/platform/android/jni/JniHelper-inl.h
+++ b/cocos/platform/android/jni/JniHelper-inl.h
@@ -1,0 +1,494 @@
+﻿/****************************************************************************
+ Copyright (c) 2010-2012 cocos2d-x.org
+ Copyright (c) 2015 Victor Komarov
+ Copyright (c) 2017 Jeff Wang
+
+ http://www.cocos2d-x.org
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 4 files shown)


halx99:
Great job.

cocos/platform/android/jni/JniHelper.h
```diff
@@ -72,159 +69,75 @@ class CC_DLL JniHelper
...
+        return callStaticMethod<std::string>(className.c_str(), methodName.c_str(), xs...); 
+    }
+
+    template <typename Ret, typename... Args>
+    static Ret callStaticMethod(const char* className, const char* methodName, const Args& ...args) {
+        typedef typename JniHelperDetail::SignatureParser<Ret (Args...)>::Result SignatureSequence;
+        const char* signature = SignatureSequence::value;
+        JniMethodInfo t;
+        if (getStaticMethodInfo(t, className, methodName, signature)) {
+            LocalRefWrapper clazz(t.env, t.classID);
```
minggo, line 117: clazz is not used, why need it?

minggo: Got it, it is used to delete local ref, but it is a strange usage.

cocos/platform/android/jni/JniHelper-inl.h
```diff
@@ -0,0 +1,494 @@
...
+};
+
+namespace JniHelperDetail {
+
+    //
+    // CharSequence
+    //
+    template <char... Chars>
+    struct CharSequence {
+        static const char value[sizeof...(Chars) + 1];
```
minggo, line 71: Why need static variable?

minggo:
It looks good, but it is hard to understand the codes. I am not a big fan of template, it looks cool, but it is hard to understand and debug.

cocos/platform/android/jni/JniHelper-inl.h
summerinsects: CharSequence::value is used to get a **compiling time string** for deduct JNI signature

cocos/platform/android/jni/JniHelper.h
summerinsects: LocalRefWrapper is used to delete local ref.
If we do not use it, we should declare a variable to hold the return value, then delete `t.classID`, and then returns.
But for case **return type==void**, it is not allowed to declare a void variable.
So I use `LocalRefWrapper`'s **destructor** to delete `t.classID`

minggo: May be should rename it similar to `lock_guard` to make it more clear: http://en.cppreference.com/w/cpp/thread/lock_guard

cocos/platform/android/jni/JniHelper-inl.h
minggo: What i mean is that, can change to non static variable, use member variable instead, or these memory can not be freed.

summerinsects: as a rule of thumb, template metaprogram uses static member.I will have a try to modify it as a non static variable

summerinsects:
I will add more comments, and offer my test codes

minggo:
> refactor all JniHelper::callStaticXXXMethod in one template callStaticMethod, but not break compatibility
 
In order to achieve it, it adds more codes to do it. And the PR heavily depends on template meta feature, make the codes hard to understand and maintain. So i think i will not merge this PR.

Sorry about it.

summerinsects:
That's all right.
I think of it as taking take practice on template metaprogramming

PR #18438, opened 2017-11-03
Spine update

minggo:


DonMizzi:
what version of spine do you update?
So, What version of spine editor we have to use?
thank you

minggo:
the version is 3.6.39

nicolechen819:
Does js will update too? thanks!!

minggo:
Yep, sure.

PR #18445, opened 2017-11-05
Fix 60 fps for android

Dimon4eg:
Fixed issue with not stable 60 fps on android.

Reason: float and double calculation.

here is float calculation:

    public static void setAnimationInterval(float interval) {
        sAnimationInterval = (long) (interval * Cocos2dxRenderer.NANOSECONDSPERSECOND);
    }

here was double calculation:

    if (Cocos2dxRenderer.sAnimationInterval <= 1.0 / 60 * Cocos2dxRenderer.NANOSECONDSPERSECOND) {

double calculation is more precise and above condition is false if 60 fps is set.


minggo:
Oh, i haven't noticed it.

Issue #18323, opened 2017-09-26
The first character is disappeared at the head of a new line in UIRichTxt


zhangjiequan:
In some cases， the first character is disappeared at the head of a new line in UIRichTxt.

- cocos2d-x version:3.15
- devices test on:Windows7 64
- developing environments
   - NDK version:NA
   - Xcode version:NA
   - VS version:Visual Studio Community 2013
   - browser type and version:NA

Steps to Reproduce:

1. push two RichElementText， "캐릭터헤라클래스,로키,나폴레옹“，”획득시 물리공격+20%", into a RichText, and adjust the width of the richText to make "20%" in a new line.
2. We will see the following
"캐릭터헤라클래스,로키,나폴레옹획득시 물리공격+
0%"
Note that “2” is disappeared.

By checking the code，I found a problem here:
//UIRichText.cpp
//RichText::handleTextRenderer
int rightStart = leftLength;
if (std::isspace(text[rightStart], std::locale()))
            rightStart++;

I think it should be written like this:
if (std::isspace(text[leftWords.length()], std::locale()))
            rightStart++;

In this case, rightStart is the num of character in UTF8(), which is 9, and leftWords.length() is 23.

Issue #14188, opened 2015-10-19
UIRichText does not handle well "\n"


pekayatt:
Most of it from:
http://discuss.cocos2d-x.org/t/does-richtext-support-multiple-line-of-text/21105/9

It only handles a element "\n", when it is in the middle of the text all element goes crazy.

pekayatt:
On JS this code has show some efficiency:

at line 400 on UIRichText.js

if(text.indexOf("\n") > -1) { //String contains \n
            var leftLength = text.indexOf("\n")+2;
            var leftWords = text.substr(0, leftLength -2);
            var cutWords = text.substr(leftLength, text.length - 1);

```
        this._handleTextRenderer(leftWords.substr(0, leftLength), fontNameOrFontDef, fontSize, color);

        this._addNewLine();
        this._handleTextRenderer(cutWords, fontNameOrFontDef, fontSize, color);
        return;
    }
```

minggo:
@pekayatt what's the error behavior? how can i reproduce it?

PR #18447, opened 2017-11-06
RichText improvements

summerinsects:
This is a PR to improve RichText.

- Automatically customize ContentSize. Related: #12806 #12145 #15054 

There is a feedback that can not get the **real height** of a richtext from many forums.
I modify `_customSize.height` in function `RichText::formatRenderers` to make **content size** equals to **real size**.

- refactor split algorithm with `StringUtils::StringUTF8`.
  - It is a waste of performance to invokes `Helper::getSubStringOfUTF8String` several times with same string. `StringUtils::StringUTF8` convers string only once.
  - on the other hand, splits `StringUtils::StringUTF8` is easier than `Helper::getSubStringOfUTF8String`. Related: #18323 
  - add a rough estimate in function `getNextWordPos` to make it runs faster.

- handle ‘\n’ for new line, and support continuous ‘\n’. Related: #14188

------------------

改进RichText

- 自动修正ContentSize，以获得真实高度 #12806 #12145 #15054 

许多论坛反馈RichText无法获取真实高度。我在函数`RichText::formatRenderers`里面修改了`_customSize.height`，使得content size等于真实尺寸。

- 用StringUtils::StringUTF8重构了字符串切割算法
  - 对同一段文本重复调用Helper::getSubStringOfUTF8String比较浪费性能。而StringUtils::StringUTF只转换一次。
  - 另一方面，切割StringUtils::StringUTF8比用Helper::getSubStringOfUTF8String切割要容易得多，且不容易出错。#18323 
  - 在函数getNextWordPos里面增加了估算长度，使之执行得更快。

- 对文本中的\n处理为换行。并支持连续的\n #14188

### DIFF
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1355,6 +1355,7 @@ void RichText::formatText()
     {
         this->removeAllProtectedChildren();
         _elementRenders.clear();
+        _defaultHeights.clear();
         if (_ignoreSize)
         {
             addNewLine();
@@ -1714,8 +1715,15 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
             str.erase(str.begin(), str.begin() + rightStart);
             currentText = utf8Text.getAsCharSequence();
 
-            addNewLine();
+            if (!currentText.empty())
+            {
+                addNewLine();
+                _defaultHeights.back() = fontSize;
+            }
         }
+
+        addNewLine();
+        _defaultHeights.back() = fontSize;
     }
 }
 
@@ -1759,11 +1767,13 @@ void RichText::addNewLine()
 {
     _leftSpaceWidth = _customSize.width;
     _elementRenders.emplace_back();
+    _defaultHeights.emplace_back();
 }
     
 void RichText::formatRenderers()
 {
     float verticalSpace = _defaults[KEY_VERTICAL_SPACE].asFloat();
+    float fontSize = _defaults[KEY_FONT_SIZE].asFloat();
 
     if (_ignoreSize)
     {
@@ -1783,7 +1793,7 @@ void RichText::formatRenderers()
                 Size iSize = iter->getContentSize();
                 newContentSizeWidth += iSize.width;
                 nextPosX += iSize.width;
-                maxY = MAX(maxY, iSize.height);
+                maxY = std::max(maxY, iSize.height);
             }
             nextPosY -= maxY;
             rowWidthPairs.emplace_back(&element, nextPosX);
@@ -1794,6 +1804,13 @@ void RichText::formatRenderers()
     }
     else
     {
+        // erase empty rears
+        while (_elementRenders.back().empty())
+        {
+            _elementRenders.pop_back();
+        }
+
+        // calculate real height
         float newContentSizeHeight = 0.0f;
         std::vector<float> maxHeights(_elementRenders.size());
         
@@ -1803,13 +1820,18 @@ void RichText::formatRenderers()
             float maxHeight = 0.0f;
             for (auto& iter : row)
             {
-                maxHeight = MAX(iter->getContentSize().height, maxHeight);
+                maxHeight = std::max(iter->getContentSize().height, maxHeight);
+            }
+            if (row.empty())
+            {
+                maxHeight = std::max(fontSize, _defaultHeights[i]);
             }
             maxHeights[i] = maxHeight;
             newContentSizeHeight += maxHeights[i] + (i != 0 ? verticalSpace : 0.0f);
         }
         _customSize.height = newContentSizeHeight;
 
+        // align renders
         float nextPosY = _customSize.height;
         for (size_t i=0, size = _elementRenders.size(); i<size; i++)
         {
@@ -1829,6 +1851,7 @@ void RichText::formatRenderers()
     }
     
     _elementRenders.clear();
+    _defaultHeights.clear();
     
     if (_ignoreSize)
     {
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1512,6 +1512,8 @@ namespace {
     int getNextWordPos(const StringUtils::StringUTF8& text, int idx)
     {
         const StringUtils::StringUTF8::CharUTF8Store& str = text.getString();
+        if (idx + 1 >= static_cast<int>(str.size()))
+            return static_cast<int>(str.size());
         auto it = std::find_if(str.begin() + idx + 1, str.end(), isUTF8CharWrappable);
         return static_cast<int>(it - str.begin());
     }
@@ -1579,7 +1581,8 @@ namespace {
                     idx = newidx;
                     continue;
                 }
-                return idx;
+                // protruded ? undo add, or quite fit
+                return (textRendererWidth > originalLeftSpaceWidth ? idx : newidx);
             }
         }
 
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1652,11 +1652,27 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
     // split text by \n
     std::stringstream ss(text);
     std::string currentText;
+    size_t realLines = 0;
     while (std::getline(ss, currentText, '\n'))
     {
+        if (realLines > 0)
+        {
+            addNewLine();
+            _defaultHeights.back() = fontSize;
+        }
+        ++realLines;
+
+        size_t splitParts = 0;
         StringUtils::StringUTF8 utf8Text(currentText);
         while (!currentText.empty())
         {
+            if (splitParts > 0)
+            {
+                addNewLine();
+                _defaultHeights.back() = fontSize;
+            }
+            ++splitParts;
+
             Label* textRenderer = fileExist ? Label::createWithTTF(currentText, fontName, fontSize)
                 : Label::createWithSystemFont(currentText, fontName, fontSize);
 
@@ -1714,16 +1730,7 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
             // erase the chars which are processed
             str.erase(str.begin(), str.begin() + rightStart);
             currentText = utf8Text.getAsCharSequence();
-
-            if (!currentText.empty())
-            {
-                addNewLine();
-                _defaultHeights.back() = fontSize;
-            }
         }
-
-        addNewLine();
-        _defaultHeights.back() = fontSize;
     }
 }
 
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1501,6 +1501,9 @@ namespace {
 
     int getPrevWordPos(const StringUtils::StringUTF8& text, int idx)
     {
+        if (idx <= 0)
+            return -1;
+
         // start from idx-1
         const StringUtils::StringUTF8::CharUTF8Store& str = text.getString();
         auto it = std::find_if(str.rbegin() + (str.size() - idx + 1), str.rend(), isUTF8CharWrappable);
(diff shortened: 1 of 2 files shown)


cocos/ui/UIRichText.cpp
```diff
@@ -1492,115 +1493,151 @@ void RichText::formatText()
...
-    return false;
-}
 
-int RichText::findSplitPositionForWord(cocos2d::Label* label, const std::string& text)
-{
-    auto originalLeftSpaceWidth = _leftSpaceWidth + label->getContentSize().width;
+    int getNextWordPos(const StringUtils::StringUTF8& text, int idx)
+    {
+        const StringUtils::StringUTF8::CharUTF8Store& str = text.getString();
+        auto it = std::find_if(str.begin() + idx + 1, str.end(), isUTF8CharWrappable);
```
minggo: if `idx` == str.length(), then it may have problem.

cocos/ui/UIRichText.cpp
```diff
@@ -1492,115 +1493,151 @@ void RichText::formatText()
...
         {
-            if (startingNewLine)
+            while (1)
+            {
+                // try to append a word
+                int newidx = getNextWordPos(text, idx);
+                leftStr = text.getAsCharSequence(0, newidx);
+                label->setString(leftStr);
+                textRendererWidth = label->getContentSize().width;
+                if (textRendererWidth < originalLeftSpaceWidth)
```
minggo, line 1580: if textRendererWidth == originalLeftSpaceWidth, then should return `newidx` instead of `idx`

cocos/ui/UIRichText.cpp
```diff
@@ -1609,110 +1646,94 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
...
-    }
-    else
-    {
-        textRenderer->setColor(color);
-        textRenderer->setOpacity(opacity);
-        pushToContainer(textRenderer);
+            // skip spaces
+            StringUtils::StringUTF8::CharUTF8Store& str = utf8Text.getString();
+            int rightStart = leftLength;
+            while (rightStart < (int)str.size() && str[rightStart].isASCII() && std::isspace(str[rightStart]._char[0], std::locale()))
```
minggo, line 1738: why need to skip spaces?

cocos/ui/UIRichText.cpp
```diff
@@ -1795,19 +1827,24 @@ void RichText::formatRenderers()
             float maxHeight = 0.0f;
             for (auto& iter : row)
             {
-                maxHeight = MAX(iter->getContentSize().height, maxHeight);
+                maxHeight = std::max(iter->getContentSize().height, maxHeight);
+            }
+            if (row.empty())
+            {
+                maxHeight = std::max(fontSize, _defaultHeights[i]);
```
minggo: why height compare with font size?

cocos/ui/UIRichText.cpp
```diff
@@ -1609,110 +1646,94 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
...
-        if (std::isspace(text[rightStart], std::locale()))
-            rightStart++;
-        std::string cutWords = Helper::getSubStringOfUTF8String(text, rightStart, text.length() - leftLength);
-        if (leftLength > 0)
+        if (realLines > 0)
         {
-            Label* leftRenderer = nullptr;
-            if (fileExist)
+            addNewLine();
+            _defaultHeights.back() = fontSize;
```
minggo: why height vector record font size?

summerinsects: This is also for continuous new lines.
eg："abc\n\n123" should display like this:
abc

123

cocos/ui/UIRichText.cpp
summerinsects: This is for continuous new lines.
eg："abc\n\n123" should display like this:
abc

123

cocos/ui/UIRichText.cpp
minggo: may be should use another name for the member variable, the name is `_defaultHeights `, but the element value is font size, it is hard to understand.

cocos/ui/UIRichText.cpp
summerinsects: In original code, it skips one space.
But I think skipping all spaces may be better.

eg(_ instead of space): "abcd________________123"
suppose it is splitted after letter 'd'
Should it desplay like

abcd
123

or

abcd
________________123
? please give me some advice

cocos/ui/UIRichText.cpp
summerinsects: Yes. I will add an if condition

cocos/ui/UIRichText.cpp
summerinsects: Yes, it will be out of bounds. I will add an if condition

cocos/ui/UIRichText.cpp
minggo: I have not idea. How it been displayed on Android/iOS UI element? I think can have the same behavior as them.

cocos/ui/UIRichText.cpp
summerinsects: How about `_emptyLineHeights`

minggo: I don't quite understand why height is equal to font size. When i reviewed the codes, it seems font size is used as height.

cocos/ui/UIRichText.cpp
summerinsects: I test it with iOS's webview. It trims the spaces. see below

cocos/ui/UIRichText.cpp
summerinsects: - there is a `RichText` with fontSize 20 contains `RichElementText`(string="abc\n\n123", fontSize=10).
Then it will display like this:
```
abc

123
```
Now,  the gap between two lines should be equal to `RichElementText`'s fontSize 10

- In another case, there is a `RichText` with fontSize 20 contains `RichElementText`("abc"), `RichElementNewLine`, `RichElementNewLine`,  `RichElementText`("123").
Then it will display like this:
```
abc

123
```
Now,  the gap between two lines should be equal to `RichText`'s fontSize 20

summerinsects:
I test it with iOS's webview. It trims the spaces.
wx20171107-181012

img_20171107_181412

cocos/ui/UIRichText.cpp
minggo: K, thanks.

cocos/ui/UIRichText.cpp
```diff
@@ -1492,115 +1493,158 @@ void RichText::formatText()
...
+        {
+            while (leftLength < stringLength)
+            {
+                // try to append a char
+                auto& ch = text.getString().at(leftLength);
+                ++leftLength;
+                leftStr.append(ch._char);
+                label->setString(leftStr);
+                textRendererWidth = label->getContentSize().width;
+                if (originalLeftSpaceWidth < textRendererWidth)  // protruded, undo add
```
minggo, line 1636: if originalLeftSpaceWidth == textRendererWidth, then it should return leftLength

summerinsects: OK

cocos/ui/UIRichText.cpp
```diff
@@ -1795,19 +1834,28 @@ void RichText::formatRenderers()
...
             for (auto& iter : row)
             {
-                maxHeight = MAX(iter->getContentSize().height, maxHeight);
+                maxHeight = std::max(iter->getContentSize().height, maxHeight);
+            }
+
+            // gap for empty line, if _lineHeights[i] == 0, use current RichText's fontSize
+            if (row.empty())
+            {
+                maxHeight = (_lineHeights[i] != 0.0f ? _lineHeights[i] : fontSize);
```
minggo, line 1841: Why use font size for height? And when row will be empty?

cocos/ui/UIRichText.cpp
```diff
@@ -1786,6 +1818,13 @@ void RichText::formatRenderers()
     }
     else
     {
+        // erase empty rears
+        while (_elementRenders.back().empty())
+        {
+            _elementRenders.pop_back();
+        }
```
minggo: When element render will be empty?

cocos/ui/UIRichText.cpp
summerinsects: Consider the following two cases：
- a `RichText` with fontSize 20 contains `RichElementText`(string="abc\n\n123", fontSize=10).
Then it will display like this:
```
abc

123
```
Now,  the gap between two lines should be equal to `RichElementText`'s fontSize 10

- a `RichText` with fontSize 20 contains `RichElementText`("abc"), `RichElementNewLine`, `RichElementNewLine`,  `RichElementText`("123").
Then it will display like this:
```
abc

123
```
Now,  the gap between two lines should be equal to `RichText`'s fontSize 20

In both of two cases, row.empty() == true. because of `RichText::addNewLine()`, the code `_elementRenders.emplace_back();`

minggo: Why not just use vertical space instead of font size? Use font size as line height is strange.

cocos/ui/UIRichText.cpp
summerinsects: This is the case of end with some new lines, for example, the **last** element of a `RichText` is `RichElementNewLine`, or `RichElementText` with some `\n`'s.
`RichText::addNewLine()` will be invoked, but no renders in the vector.

minggo: Why remove them? May be it is the expected effect, use some new line at the end to make some space at the end of content.

summerinsects:
In my opinion
wx20171108-101726

cocos/ui/UIRichText.cpp
summerinsects: vertical space is not same as font size.please see below

minggo: sounds reasonable

cocos/ui/UIRichText.cpp
summerinsects: At first, I think new lines at end of RichText is meaningless.
Maybe we can add a method to decide whether trim or not, how about it?

minggo: I think it may make things complex. If they don't want the effect, then why add these new lines explicitly?

summerinsects: OK. I see. I will delete these codes

summerinsects:
@minggo 
I get inspiration from `class Label`s source code, which is
```c++
const Size& Label::getContentSize() const
{
    if (_systemFontDirty || _contentDirty)
    {
        const_cast<Label*>(this)->updateContent();
    }
    return _contentSize;
}

Rect Label::getBoundingBox() const
{
    const_cast<Label*>(this)->getContentSize();

    return Node::getBoundingBox();
}
```
It means that when invoke `getContentSize` with a dirty content, it should update first.

Are these codes necessary similarly?

```c++
const Size& RichText::getContentSize() const
{
    if (_formatTextDirty)
    {
        const_cast<RichText*>(this)->formatText();
    }
    return Widget::getContentSize();
}

Rect RichText::getBoundingBox() const
{
    if (_formatTextDirty)
    {
        const_cast<RichText*>(this)->formatText();
    }
    return Widget::getBoundingBox();
}
```

minggo:
From the logic it seems it is needed, but i am not sure. Do you meet any problem?

summerinsects:
I did not notice this problem.
In my code, I always invoked `formatText` explicitly before `getContentSize`
I am not sure whether overriding these functions will cause side effects

minggo:
I am not sure too, since some call back is invoked from UIWidget. So let's keep as it is and fix them when meet issue.

summerinsects:
OK

rh101:
Guys, introducing the trimming spaces on the right side of a text element makes no sense. It should not trim any space, since a RichText is made out of several elements, and spaces on the right side of text elements are used to literally put space between those elements.  

For example, say you have this UIRichText made up of 3 elements, text, image, and another text:

text[space]Image[space]text

which should be rendered as:

text image text

yet it appears as:

textimagetext

If for some reason you need it to trim right side spaces, then at least give us the option to turn that functionality on or off by some kind of flag bit (such as RichElementText::TRIM_RIGHT_SPACE etc).

PR #18457, opened 2017-11-09
Allow nodes to be relative to world.

sp-jordi-vilalta:
This adds the possibility to set individual nodes to be relative to the world instead of relative to the parent node.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -62,7 +62,8 @@ int Node::__attachedNodeCount = 0;
 // MARK: Constructor, Destructor, Init
 
 Node::Node()
-: _rotationX(0.0f)
+: _isRelativeToWorld(false)
+, _rotationX(0.0f)
 , _rotationY(0.0f)
 , _rotationZ_X(0.0f)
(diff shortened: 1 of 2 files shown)


cocos/2d/CCNode.cpp
```diff
@@ -1714,7 +1725,17 @@ const Mat4& Node::getNodeToParentTransform() const
...
+
+        if (_isRelativeToWorld && _parent)
+        {
+            _transform = _transform * _parent->getWorldToNodeTransform();
+        }
+
+        Mat4 rotation;
+        Mat4::createRotation(_rotationQuat, &rotation);
+
+        _transform = _transform * rotation;
```
minggo, line 1738: The function name is `getNodeToParentTransform `, so i think the logic is not good here.

sp-jordi-vilalta: Ok. Other than this, do you think the feature could be acceptable into Node?

minggo: I think the game logic should handle it manually, not add the feature in Node. Node is now very heavy, it is better no add new features there.

PR #18460, opened 2017-11-09
fix #18458

### LINKED ISSUE #18458: RichText issue, cause by Label when its string is very very long
- cocos2d-x version:git/v3
- devices test on:iOS
- developing environments
   - NDK version:
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:
create a `RichText` which contains `RichElementText` with a very very long text.

see the codes in `RichText::handleTextRenderer`
```c++
            float textRendererWidth = textRenderer->getContentSize().width;
            _leftSpaceWidth -= textRendererWidth;

            // no splitting
            if (_leftSpaceWidth >= 0.0f)
            {
                pushToContainer(textRenderer);
                break;
            }
```
When the string is very very long, `Label:: getContentSize` returns Zero Size, and does not contains any texture. So it seems that lost a part of text in `RichText` after `pushToContainer(textRenderer)`.

Reason:
`Label:: getContentSize` calls `Label::updateContent`.
It meets a **glError** in `Label::updateContent`. 
Console ouput is:
```
cocos2d: Texture2D: Error uploading compressed texture level: 0 . glError: 0x0501
```

I will try to fix it

summerinsects:
fix #18458

### DIFF
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1528,26 +1528,17 @@ namespace {
         return std::any_of(str.begin(), str.end(), isUTF8CharWrappable);
     }
 
-    int findSplitPositionForWord(Label* label, const StringUtils::StringUTF8& text, float leftSpaceWidth, float newLineWidth)
+    int findSplitPositionForWord(Label* label, const StringUtils::StringUTF8& text, int estimatedIdx, float originalLeftSpaceWidth, float newLineWidth)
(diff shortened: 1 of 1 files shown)


summerinsects:
eg the long text is
longtext
it displays like this
display1
and the console log is
consolelog


with this pr, it will be
display2

minggo (approved):

PR #18464, opened 2017-11-13
--force-yes => --allow, -y

wnddpd01:
apt-get --force-yes is deprecated

exchange --allow and -y


minggo:
Do you know in which version is `--allow-change-held-packages` added?

wnddpd01:
In debian manpage it released APT 1.1 

https://manpages.debian.org/stretch/apt/apt-get.8.en.html

minggo:
what i am afraid if it will break compatibility for Ubuntu 14.01.

PR #16262, opened 2016-07-29
#Improve node sort performance for 64-bit by std::sort + long long (_localZOrder)

halx99:
- And bring _orderOfArrival back for 32bits program,
- From vs2015 test result, the performance of lambda expression is better than the function: nodeComparisonLess

test program:

```
#include <iostream>
#include <stdio.h>
#include <stdlib.h>
#include <algorithm>
#include <vector>
#include <time.h>
#include <assert.h>

#if 0 // for android
#define  LOG_TAG    "cpp-empty-test"
#define  LOGD(...)  __android_log_print(ANDROID_LOG_DEBUG,LOG_TAG,__VA_ARGS__)
#define logfunc(format,...) __android_log_print(ANDROID_LOG_DEBUG,LOG_TAG,(format "\n"),##__VA_ARGS__)
#else
#define logfunc(format,...) printf((format "\n"),##__VA_ARGS__)
#endif

class SimpleNode
{
public:
    SimpleNode(int z, unsigned int a)
    {
        _localZOrder = 0;
        setLocalZOrder(z);
        setOrderOfArrival(a);
    }

    int getLocalZOrder() const
    {
        return static_cast<int>(((_localZOrder >> 32) & 0xffffffff));
    }

    unsigned int getOrderOfArrival() const
    {
        return static_cast<unsigned int>((_localZOrder) & 0xffffffff);
    }

    void setLocalZOrder(int z)
    {
        _localZOrder = ((0ULL + z) << 32) | (_localZOrder & 0xffffffff);
        _o_localZOrder = z;
    }

    void setOrderOfArrival(unsigned int orderOfArrival)
    {
        _localZOrder = (_localZOrder & 0xffffffff00000000) | orderOfArrival;
        _o_orderOfArrival = orderOfArrival;
    }

    int _o_localZOrder;
    unsigned int _o_orderOfArrival;

    std::int64_t _localZOrder;
};

inline bool nodeComparisonLess64(SimpleNode* n1, SimpleNode* n2)
{
// #if defined(_M_X64) || defined(_WIN64) || defined(__LP64__) || defined(_LP64) || defined(__x86_64)
    return (n1->_localZOrder < n2->_localZOrder);
// #endif
}

void runPerfromanceFunc()
{
    SimpleNode n1(0, 1);
    SimpleNode n2(1, 0);
    SimpleNode n3(-3, 500);


    assert(n3.getLocalZOrder() == -3);
    assert(n3.getOrderOfArrival() == 500);
[...]

PR #18480, opened 2017-11-20
Fix endian detection on Android (fixes node render order)

jonathonwalz:
When compiling for Android, the preprocessor defines for byte order are
are not included in the sys/param.h chain. By including sys/endian.h we
we can ensure that the necessary macros are included.

It is especially important that endian-ness is correctly detected as the
node render order uses this information to sort #16262. Without this
fix, nodes will be rendered in the incorrect order on 64-bit Android
devices (on 32-bit, a different algorithm is used that is not
sensitive to byte order).

I'm not sure if this is the best fix or if there are additional platforms
that are affected. I've only tested this on iOS and Android.

### DIFF
--- a/cocos/base/ccMacros.h
+++ b/cocos/base/ccMacros.h
@@ -296,6 +296,9 @@ CC_ASSERT(__gl_error_code == GL_NO_ERROR, "Error"); \
   #endif
 #else
   #include <sys/param.h>
+  #if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
+    #include <sys/endian.h>
+  #endif // CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID
   #if ((defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)) || \
          (defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN)) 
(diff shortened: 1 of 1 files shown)


cocos/base/ccMacros.h
```diff
@@ -296,6 +296,9 @@ CC_ASSERT(__gl_error_code == GL_NO_ERROR, "Error"); \
   #endif
 #else
   #include <sys/param.h>
+  #if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID)
+    #include <sys/endian.h>
+  #endif // CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID
```
minggo, line 301: Should need to include it for iOS too?

jonathonwalz: I don't know enough about how the low level library's work across platforms to be able to give a very good answer. However, `<sys/endian.h>` does not exist on iOS (at least, when compiling I get `'sys/endian.h' file not found`). 

There may be a better header to include across all platforms but I don't know what it is.

minggo: The codes from https://github.com/google/protobuf/blob/master/src/google/protobuf/io/coded_stream.h#L131, may be you can send a PR to it too.

Dimon4eg:
Cool, just met bug with node order. After this fix it works fine. Thank you!

PR #18496, opened 2017-11-23
Edit install_deps_linux

wnddpd01:
APT Version check

Add start comment


build/install-deps-linux.sh
```diff
@@ -1,6 +1,16 @@
...
 # Change directory to the location of this script
+echo "This Shell Script will install dependencies for cocos2d-x" 
+echo "if you execute this shell more than once it will get errors when building libGLFW.so"
+echo -n "Are you continue? (y/n) "
+read answer
+if echo "$answer" | grep -iq "^y" ;then
+    echo "It will take few minutes"
+else
+    exit
+fi
```
minggo, line 12: it will break travis and other automatically tools

minggo: can refer to [this url](https://travis-ci.org/cocos2d/cocos2d-x/jobs/306191033) for example

PR #18503, opened 2017-11-24
fix memory leak in FontFreeType::create

dhb118:


drelaptop (approved):

drelaptop (approved):
`FontFreeType` is the subclass of `Ref`, the return object of the create method  is been `retain()` in `FontAtlas`, and `release()` in `FontAtlasCache`. So this is correct to add the object to __AutoreleasePool__, and my testing process also verified this. please Merge @minggo

minggo:
Thanks guys.

PR #18505, opened 2017-11-25
Fixed memory leak when using TTF labels

Bidju7777:
This patch removes a memory leak that happened in Labels created using createWithTTF() method.

Tested on iOS (simulator & device (iPhone 6S, iOS 11.1.2).

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -438,7 +438,6 @@ Label::~Label()
     if (_fontAtlas)
     {
         Node::removeAllChildrenWithCleanup(true);
-        CC_SAFE_RELEASE_NULL(_reusedLetter);
         _batchNodes.clear();
         FontAtlasCache::releaseFontAtlas(_fontAtlas);
     }
@@ -447,6 +446,7 @@ Label::~Label()
 
     CC_SAFE_RELEASE_NULL(_textSprite);
     CC_SAFE_RELEASE_NULL(_shadowNode);
+    CC_SAFE_RELEASE_NULL(_reusedLetter);
 }
 
 void Label::reset()
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -438,6 +438,7 @@ Label::~Label()
     if (_fontAtlas)
     {
         Node::removeAllChildrenWithCleanup(true);
+        CC_SAFE_RELEASE_NULL(_reusedLetter);
         _batchNodes.clear();
         FontAtlasCache::releaseFontAtlas(_fontAtlas);
     }
@@ -446,7 +447,6 @@ Label::~Label()
 
     CC_SAFE_RELEASE_NULL(_textSprite);
     CC_SAFE_RELEASE_NULL(_shadowNode);
-    CC_SAFE_RELEASE_NULL(_reusedLetter);
 }
 
 void Label::reset()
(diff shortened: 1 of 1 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -447,6 +446,7 @@ Label::~Label()
 
     CC_SAFE_RELEASE_NULL(_textSprite);
     CC_SAFE_RELEASE_NULL(_shadowNode);
+    CC_SAFE_RELEASE_NULL(_reusedLetter);
```
drelaptop: what is the reason? if really need I think we should move ```        Node::removeAllChildrenWithCleanup(true);
        CC_SAFE_RELEASE_NULL(_reusedLetter);``` both line, just like the method `Label::reset()` does

Bidju7777: The problem seems to come from line 605, where `_reusedLetter` is manually retained and is never released if `_fontAtlas` is nullptr when the Label is deleted.

Calling `removeAllChildrenWithCleanup(true);` might be unnecessary as the profiler (Instuments.app) told that there was no leak after the modification.

drelaptop: when I test using cpp-empty-tests, I find at method `setFontAtlas()` member:`_fontAtlas` and `_reusedLetter` init together, and at method `reset()` set nullptr together, in fact I check this program using _instruments->leaks_, didn't find memory leaks. I didn't find the state: __fontAtlas is nullptr, \_reusedLetter not nullptr__` can you describe how you used, It's better to show some code.

Bidju7777: Here are screenshots of Instruments:

Before patch:
a_beforepatch

After patch:
b_afterpatch

drelaptop: You are right, leaks occur in this method

The state __fontAtlas is nullptr, \_reusedLetter not nullptr__ appear, when I make a condition breakpoint `(_fontAtlas==nullptr)&&(_reusedLetter!=nullptr)` and run `57:System Font Test` of cpp-tests. 
After some analysis, I found it cased by `Label::updateContent`, in this method `_fontAtlas` is clear, `_reusedLetter` not.

and the test code is
```C++
    auto label = Label::createWithTTF("This is a very long sentence", "fonts/arial.ttf", 20);
    label->setSystemFontSize(40); 
    label->setPosition(Vec2(visibleSize / 2) + origin);
    this->addChild(label, 1);
```
only `createWithTTF` leaks doesn't occur

drelaptop: so we also can add `CC_SAFE_RELEASE_NULL(_reusedLetter)` to `Label::updateContent()`, keep the both `_fontAtla`  and `_reusedLetter` clear at same time. This leaks will be fixed also.
```C++
void Label::updateContent()
{
    if (_systemFontDirty)
    {
        if (_fontAtlas)
        {
            _batchNodes.clear();
            CC_SAFE_RELEASE_NULL(_reusedLetter); // this is new line added
            FontAtlasCache::releaseFontAtlas(_fontAtlas);
            _fontAtlas = nullptr;
    ...
}
```
will you test this leaks using this method? I think it will also works. @UsualYann

Bidju7777: Ok this worked, I've pushed the change. Thanks :)

drelaptop:
thanks for your patience.

I think it's a good patch, author and I both have tested it, you could merge it. @minggo

PR #18517, opened 2017-11-27
Upgrade appveyor to Visual Studio 2017

crazyhappygame:


minggo:
Now, cocos2d-x uses VS2015 as default VS version, so will not change it.

crazyhappygame:
Why not switch to new version? Is there anything what stop us to do it?
1. Visual Studio 2017 is binary compatible with Visual Studio 2015. We do not have recompile libraries
2. We will have much butter C++ 11/14/17 support

minggo:
Developers doesn't agree with it, i started a survey to remove VS2013 before, and even many developers doesn't agree with it.

crazyhappygame:
I found one survey from may 2016. 
Can we create another one survey for vs 2017 now?
Switching from vs 2015 to 2017 is much easier than from other versions.

minggo:
yep, we may, but i don't have time to update 3rd party libs with VS2017

crazyhappygame:
I checked it  and there are only two libraries which we have to rebuild: chipmunk and Box2D. Those libraries were compiled using flag /GL or /LTCG. It is enough to recompile chipmunk and Box2D with VS2015 without /GL or /LTCG flags.

Other libraries are fine. Visual Studio 2017 is binary compatible with Visual Studio 2015. (Except when are compiling with flags /GL /LTCG)

I compiled using VS2017 and run successfully cpptest from build\cocos2d-win32.sln

Issue #18403, opened 2017-10-24
TextField::initRenderer() missing font


MrCapone:
When i trying to create `TextField` by 
```
TextField::create("some_text", "some_font_from_resources", 18);
```
I got log warning
> cocos2d: fullPathForFilename: No file found at Thonburi. Possible missing file

breakpoint brings me to this [line](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/ui/UITextField.cpp#L335). What is **Thonburi** and why it missing?

drelaptop:
__Thonburi__ is a font name, you can find `Thonburi.ttf` in the engine. but I think it's not a good method to init `TextField ` with a determinate font, I will try to find a method to solve it.

drelaptop:
this Issue has been fixed, pleased see PR #18518

PR #18518, opened 2017-11-28
add an empty create() method for UICCTextField

drelaptop:
A create method with parameters is used In an empty init method, It will do something useless, and cause issues like https://github.com/cocos2d/cocos2d-x/issues/18403

### DIFF
--- a/cocos/ui/UITextField.cpp
+++ b/cocos/ui/UITextField.cpp
@@ -46,6 +46,20 @@ UICCTextField::~UICCTextField()
 {
 }
 
+UICCTextField * UICCTextField::create()
+{
+    UICCTextField *pRet = new (std::nothrow) UICCTextField();
+    
+    if(pRet)
+    {
+        pRet->autorelease();
+        return pRet;
+    }
+    CC_SAFE_DELETE(pRet);
+    
+    return nullptr;
+}
+
 UICCTextField * UICCTextField::create(const std::string& placeholder, const std::string& fontName, float fontSize)
 {
     UICCTextField *pRet = new (std::nothrow) UICCTextField();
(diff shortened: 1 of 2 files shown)


cocos/ui/UITextField.cpp
```diff
@@ -46,6 +46,20 @@ UICCTextField::~UICCTextField()
 {
 }
 
+UICCTextField * UICCTextField::create()
+{
+    UICCTextField *pRet = new (std::nothrow) UICCTextField();
```
minggo: pRet -> ret
we don't use Hungarian Notation now. Old codes need to be modified when meet it.

cocos/ui/UITextField.h
```diff
@@ -58,6 +58,13 @@ class CC_GUI_DLL UICCTextField: public TextFieldTTF, public TextFieldDelegate
     
     virtual void onEnter() override;
     
+    /**
+     *@brief Create an empty UICCTextField.
+     
+     *@return A UICCTextField instance.
+     */
+    static UICCTextField* create();
```
minggo: static function should be declared before all other functions. Please refer to [coding style][https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md].

drelaptop: OK, I will look at code style, and change it

cocos/ui/UITextField.cpp
```diff
@@ -32,6 +32,20 @@ NS_CC_BEGIN
...
+UICCTextField * UICCTextField::create()
+{
+    UICCTextField *ret = new (std::nothrow) UICCTextField();
+
+    if(ret)
+    {
+        ret->autorelease();
+        return ret;
+    }
+    CC_SAFE_DELETE(ret);
```
minggo: Is it necessary? ret is nullptr here, may change the code to 
```c++
if (ret)
    ret->autorelease();

return ret;
```

PR #18525, opened 2017-11-29
Update Cocos2dxHelper.java

mschoi17:
1. Add comments for Cocos2dxHelper.java
2. Convert indention to tap


minggo:
all cpp codes uses 4 spaces instead of tab, so java codes should use the same style.

PR #18526, opened 2017-11-29
Add new types support to CCValue

Bidju7777:
Now supports short, unsigned short, long, unsigned long, long long, unsigned long long and long double.
Especially useful if you need to parse huge 64 bits numbers.

Tested on iOS (simulator & device (iPhone 6S, iOS 11.1.2).


minggo:
* please don't change file property from 100644 to 100755
* i think may be it is better to use template to refactor this class

PR #18536, opened 2017-12-05
Uses NDK r16 to generate js/lua binding code

dumganhar:


.gitmodules
```diff
@@ -3,7 +3,7 @@
         url = git://github.com/cocos2d/cocos2d-console.git
 [submodule "tools/bindings-generator"]
         path = tools/bindings-generator
-        url = git://github.com/cocos2d/bindings-generator.git
+        url = git://github.com/dumganhar/bindings-generator.git
```
dumganhar: Temp to modify my repo. 
Should revert before being merged.

Issue #18554, opened 2017-12-12
TextField with system font cursor support


sbrednikhin:
Current implementation of ui:TextField doesn't support cursor if initialized with system font. 
It can be easily enabled without cursor "blinking" and cursor with zero width. Anyway it's be better than nothing.

PR #18552, opened 2017-12-11
ScrollView API to stop overall scroll

AIGRIND-LLC:
Fix for issue #18551

PR #18555, opened 2017-12-12
Partial cursor support for text field with system font

sbrednikhin:
Enables cursor support without cursor blinking an zero with of cursor symbol. In case of system font usage magical "/b" symbol won't be used, also update won't be scheduled.
Issue #18554

### DIFF
--- a/cocos/2d/CCTextFieldTTF.cpp
+++ b/cocos/2d/CCTextFieldTTF.cpp
@@ -567,7 +567,7 @@ void TextFieldTTF::makeStringSupportCursor(std::string& displayText)
         {
             // \b - Next char not change x position
             if (_currentLabelType == LabelType::TTF)
-			    displayText.push_back((char) TextFormatter::NextCharNoChangeX);
+                displayText.push_back((char) TextFormatter::NextCharNoChangeX);
             displayText.push_back(_cursorChar);
         }
         else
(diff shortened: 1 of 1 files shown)


cocos/2d/CCTextFieldTTF.cpp
```diff
@@ -679,29 +681,24 @@ const std::string& TextFieldTTF::getPlaceHolder() const
...
-        CCLOG("TextFieldTTF cursor worked only LabelType::TTF");
-    }
+	if (_cursorEnabled != enabled)
+	{
+	    _cursorEnabled = enabled;
+	    if (_cursorEnabled)
+	    {
+	        _cursorPosition = _charCount;
+	
+			if (_currentLabelType == LabelType::TTF)
```
drelaptop: format error, a `Tab` should be replaced by 4 `Space`, and can you add a test case to show the fix? I noticed that the old test didn't test cursor. 

> the content needed to add test case, you can refer to [pull/18552/files](https://github.com/cocos2d/cocos2d-x/pull/18552/files)

drelaptop (changes requested):

cocos/2d/CCTextFieldTTF.cpp
sbrednikhin: Ok, thank you, I missed that. Test implementations will take some time.

drelaptop:
1. This patch is a little feature, which is support cursor __without blinking__ when create TextField using __System Font__. 
2. using this feature like:
```
        TextField* textField = TextField::create("input words here","Arial",30);
        textField->setCursorEnabled(true);
```
3. no test case for this little feature, but it works well when I test on macOS and iOS. Suggest merge @minggo

Issue #18566, opened 2017-12-15
The libSpine is crashed.


LSilent:
- cocos2d-x version: 3.16
- devices test on: Windows 7
- developing environments
   - VS version: vs2017
Steps to Reproduce:
1. run cpp-tests
2. click the "44:Node:Spine" test.
The thrid example,"GoblinsExample", is crashed in SkeletonAnimation::createWithJsonFile

drelaptop:
out-of-date issue ,fixed in https://github.com/cocos2d/cocos2d-x/pull/18582, please close @minggo

PR #18567, opened 2017-12-15
#18566 bug fixed:the spine is crashed when being created.

LSilent:
#18566 bug fixed:the spine is crashed when being created.


drelaptop:
bad and out-of-date PR ,corresponding issue fixed in https://github.com/cocos2d/cocos2d-x/pull/18582, please __close__ directly @minggo

PR #18572, opened 2017-12-17
Fix cocos2d::EditBox padding

kaplad:
**Problem:**

`cocos2d::EditBox` has padding: `static const int CC_EDIT_BOX_PADDING = 5;` and it does not count it when applying width size for labels
**Steps to reproduce:** simply create edit box and enter more characters then you can see on screen
**Solution:** track all changes of content size and apply padding:


drelaptop (approved):
Thanks for your contributions, for your quite clear description and works well patch.

review and test pass, please merge @minggo

kaplad:
@drelaptop Happy to help! :)

PR #18577, opened 2017-12-19
fix JSB Memory leaks and crashes

dumganhar:
Fixed:

wechatimg80

### DIFF
--- a/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
+++ b/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
@@ -121,6 +121,22 @@ static void ReportException(JSContext *cx)
     }
 }
 
+static void onGarbageCollect(JSRuntime* rt, JSGCStatus status, void* data)
+{
+    /* We finalize any pending toggle refs before doing any garbage collection,
+     * so that we can collect the JS wrapper objects, and in order to minimize
+     * the chances of objects having a pending toggle up queued when they are
+     * garbage collected. */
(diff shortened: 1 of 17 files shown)


cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -74,6 +74,10 @@
 #define TRACE_DEBUGGER_SERVER(...)
 #endif // #if COCOS2D_DEBUG
 
+#if CC_ENABLE_GC_FOR_NATIVE_OBJECTS
+#error "The functionality of JS controls C++ object's lifecycle isn't stable enough, please don't enable it now."
```
dumganhar, line 78: JS controls cpp isn't stable enough, make an error if user enables it.

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -1053,7 +1070,6 @@ void ScriptingCore::releaseScriptObject(cocos2d::Ref* owner, cocos2d::Ref* targe
 void ScriptingCore::unrootScriptObject(cocos2d::Ref* target)
 {
     JS::RootedObject global(_cx, _global->get());
-    JSAutoCompartment ac(_cx, global);
```
dumganhar, line 1056: No need to invoke `JSAutoCompartment` every where since we use JS_EnterCompartment/JS_LeaveCompartment

PR #18579, opened 2017-12-19
improvements for EditBox-win32

summerinsects:
1. add support for **setNativeTextHorizontalAlignment**
2. add support for **EditBoxEndAction**
3. merge `createSingleLineEditCtrl` and `createMultilineEditCtrl` into one function
4. add :: for Windows api calls and use wide-character versions explicitly
5. replace member variable `hwndEdit` with `_hwndEdit` to match the global coding style
6. replace `const_cast<char16_t*>(wstrResult.c_str())` with `(LPWSTR)&wstrResult[0]`

### DIFF
--- a/cocos/ui/UIEditBox/UIEditBoxImpl-win32.cpp
+++ b/cocos/ui/UIEditBox/UIEditBoxImpl-win32.cpp
@@ -53,16 +53,14 @@ namespace ui {
     void EditBoxImplWin::lazyInit()
     {
         s_hwndCocos = cocos2d::Director::getInstance()->getOpenGLView()->getWin32Window();
-        LONG style = GetWindowLong(s_hwndCocos, GWL_STYLE);
-        SetWindowLong(s_hwndCocos, GWL_STYLE, style | WS_CLIPCHILDREN);
+        LONG style = ::GetWindowLongW(s_hwndCocos, GWL_STYLE);
+        ::SetWindowLongW(s_hwndCocos, GWL_STYLE, style | WS_CLIPCHILDREN);
         s_isInitialized = true;
         s_previousFocusWnd = s_hwndCocos;
 
-        s_hInstance = GetModuleHandle(NULL);
+        s_hInstance = ::GetModuleHandleW(nullptr);
 
-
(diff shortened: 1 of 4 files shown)


drelaptop:
Would you like to modify __the exist test case `UIEditBoxTest`__  or _create a new case_ to test it? thanks for this feature.

summerinsects:
@drelaptop  **EditBoxEndAction** is already in test case. I add a test case for **HorizontalAlignment** only

cocos/ui/UIEditBox/UIEditBoxImpl-win32.cpp
```diff
@@ -203,119 +170,146 @@ namespace ui {
...
+
     void EditBoxImplWin::setNativeReturnType(EditBox::KeyboardReturnType returnType)
     {
         //not implemented yet
     }
+
+    void EditBoxImplWin::setNativeTextHorizontalAlignment(TextHAlignment alignment)
+    {
+        LONG style = ::GetWindowLongW(_hwndEdit, GWL_STYLE);
+        switch (alignment)
```
drelaptop, line 222: how about this code style, it's more simple and more easy to understand:
```C++
        switch (alignment)
        {
        case TextHAlignment::LEFT:
            style = (style & ~ES_CENTER & ~ES_RIGHT) | ES_LEFT;
            break;
        case TextHAlignment::CENTER:
            style = (style & ~ES_LEFT & ~ES_RIGHT) | ES_CENTER;
            break;
        case TextHAlignment::RIGHT:
            style = (style & ~ES_LEFT & ~ES_CENTER) | ES_RIGHT;
            break;
        }
```

drelaptop (approved):
The test case works well on `win32/macOS/iOS` platform. 

when I test on Android, I find `setTextHorizontalAlignment(TextHAlignment::CENTER)` didn't works well, thanks for your test so as I can find the issue on Android.

cocos/ui/UIEditBox/UIEditBoxImpl-win32.cpp
summerinsects: That is a good idea

drelaptop:
This is a good patch, and works well on `win32/macOS/iOS` platform, please merge @minggo . 

Fortunately another issue on `Android` platform was found #18581, I will keep focus on it.

PR #18582, opened 2017-12-22
update [fix warnings] ,correct old warning fix for spine

drelaptop:
fix warnings #18499 have some defect,  at that commit, variable name is modified, but some places where the variable was quoted were not modified at the same time.


drelaptop:
I have tested this on macOS and Android, All spine test cases work well, please review and merge @minggo

PR #18586, opened 2017-12-25
spine res name update, because files name changed before

drelaptop:
spine resource name update, because files name changed in old commits, however the name in code __didn't changed completely__.

without this update, `1.Node: Scene3D` in `cpp-tests`  project will crash when loading spine resource: __spine/goblins_mesh.json__.


drelaptop:
this is a simple change, and I have tested on macOS. please review and merge. @minggo

PR #18587, opened 2017-12-25
Added support for BMFont in TextField

roed:
# What

The cocos2d::ui::TextField class did not support BMFont. This PR adds that functionality and a test case as well.

# How to test
- [ ] Get this pr
- [ ] Build the test cases
- [ ] View the new textfield test case: 51:Node: UI -> 1. UI Dynamic Create Test -> 13. TextField test -> UITextFieldTest_BMFont
- [ ] See that the textfield uses a BMFont (marker felt) and renders fine

# Notes
I also changed the class TextFieldTTF, which already supported TTF and System fonts. The name of that class is ambiguous, I assumed it only supported TTF fonts. I think it should be renamed, as it clearly supports more font types. Did not to that in this PR though, as the change would be too big.

### DIFF
--- a/cocos/ui/UITextField.cpp
+++ b/cocos/ui/UITextField.cpp
@@ -446,6 +446,9 @@ void TextField::setFontSize(int size)
     {
         _textFieldRenderer->setSystemFontSize(size);
     }
+    else if (_fontType == FontType::BMFONT) {
+        _textFieldRenderer->setBMFontSize(size);
+    }
     else
     {
         TTFConfig config = _textFieldRenderer->getTTFConfig();
@@ -466,11 +469,17 @@ void TextField::setFontName(const std::string& name)
 {
     if(FileUtils::getInstance()->isFileExist(name))
     {
-        TTFConfig config = _textFieldRenderer->getTTFConfig();
-        config.fontFilePath = name;
-        config.fontSize = _fontSize;
-        _textFieldRenderer->setTTFConfig(config);
-        _fontType = FontType::TTF;
+        if (name.find(".fnt") != std::string::npos) {
+            _textFieldRenderer->setBMFontFilePath(name);
+            _fontType = FontType::BMFONT;
+        }
+        else {
+            TTFConfig config = _textFieldRenderer->getTTFConfig();
+            config.fontFilePath = name;
+            config.fontSize = _fontSize;
+            _textFieldRenderer->setTTFConfig(config);
+            _fontType = FontType::TTF;
+        }
     }
     else
     {
--- a/cocos/ui/UITextField.cpp
+++ b/cocos/ui/UITextField.cpp
@@ -469,7 +469,9 @@ void TextField::setFontName(const std::string& name)
 {
     if(FileUtils::getInstance()->isFileExist(name))
     {
-        if (name.find(".fnt") != std::string::npos) {
+        std::string lcName = name;
+        std::transform(lcName.begin(), lcName.end(), lcName.begin(), ::tolower);
+        if(lcName.substr(lcName.length() - 4) == ".fnt") {
             _textFieldRenderer->setBMFontFilePath(name);
             _fontType = FontType::BMFONT;
         }
--- a/cocos/ui/UITextField.cpp
+++ b/cocos/ui/UITextField.cpp
@@ -316,6 +316,20 @@ TextField* TextField::create(const std::string &placeholder, const std::string &
     return nullptr;
 }
     
+TextField* TextField::createWithBMFont(const std::string &placeholder, const std::string &fontName)
+{
+    TextField* widget = new (std::nothrow) TextField();
+    if (widget && widget->init())
+    {
+        widget->setFontName(fontName);
+        widget->setPlaceHolder(placeholder);
+        widget->autorelease();
+        return widget;
+    }
+    CC_SAFE_DELETE(widget);
+    return nullptr;
+}
+    
 bool TextField::init()
 {
     if (Widget::init())
--- a/cocos/ui/UITextField.cpp
+++ b/cocos/ui/UITextField.cpp
@@ -316,20 +316,6 @@ TextField* TextField::create(const std::string &placeholder, const std::string &
     return nullptr;
 }
     
-TextField* TextField::createWithBMFont(const std::string &placeholder, const std::string &fontName)
-{
-    TextField* widget = new (std::nothrow) TextField();
-    if (widget && widget->init())
-    {
-        widget->setFontName(fontName);
-        widget->setPlaceHolder(placeholder);
-        widget->autorelease();
-        return widget;
-    }
-    CC_SAFE_DELETE(widget);
-    return nullptr;
-}
-    
 bool TextField::init()
 {
     if (Widget::init())
--- a/tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UITextFieldTest/UITextFieldTest.cpp
+++ b/tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UITextFieldTest/UITextFieldTest.cpp
@@ -446,8 +446,8 @@ bool UITextFieldTest_BMFont::init()
         _uiLayer->addChild(alert);
         
         // Create the textfield
-        TextField* textField = TextField::create("input words here","fonts/markerFelt.fnt",25);
-        
+        TextField* textField = TextField::create("input words here","fonts/bitmapFontTest3.fnt",14);
+        textField->setCursorEnabled(true);
         textField->setPosition(Vec2(widgetSize.width / 2.0f, widgetSize.height / 2.0f));
         textField->addEventListener(CC_CALLBACK_2(UITextFieldTest_BMFont::textFieldEvent, this));
         _uiLayer->addChild(textField);
--- a/tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UITextFieldTest/UITextFieldTest.cpp
+++ b/tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UITextFieldTest/UITextFieldTest.cpp
@@ -446,7 +446,7 @@ bool UITextFieldTest_BMFont::init()
         _uiLayer->addChild(alert);
         
         // Create the textfield
-        TextField* textField = TextField::create("input words here","fonts/bitmapFontTest3.fnt",14);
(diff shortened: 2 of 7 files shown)


drelaptop:
Your notes is reasonable, but when considering compatibility, we shouldn't change class name, if we rename it, developers will find that old code doesn't work when they update engine from 3.16 to 3.17 or later.

I think a better way is updating comments, so developers will find this new feature when they check the new API docs. you can modify the class comments and other comments to make known the new feature.

such as comments below should been updated:
```C++
/**
 *@brief    A simple text input field with TTF font.
 */
class CC_DLL TextFieldTTF : public Label, public IMEDelegate
```

tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UITextFieldTest/UITextFieldTest.cpp
```diff
@@ -416,6 +417,82 @@ void UITextFieldTest_TrueTypeFont::textFieldEvent(Ref *pSender, TextField::Event
...
+        _displayValueLabel->setPosition(Vec2(widgetSize.width / 2.0f, widgetSize.height / 2.0f + _displayValueLabel->getContentSize().height * 1.5f));
+        _uiLayer->addChild(_displayValueLabel);
+        
+        // Add the alert
+        Text* alert = Text::create("TextField","fonts/Marker Felt.ttf",30);
+        alert->setPosition(Vec2(widgetSize.width / 2.0f, widgetSize.height / 2.0f - alert->getContentSize().height * 3.075f));
+        _uiLayer->addChild(alert);
+        
+        // Create the textfield
+        TextField* textField = TextField::create("input words here","fonts/markerFelt.fnt",25);
```
drelaptop: suggestion 1st:
```
        // Create the textfield
        TextField* textField = TextField::create("input words here","fonts/bitmapFontTest3.fnt",25);
        textField->setCursorEnabled(true);
```
1. `fonts/bitmapFontTest3.fnt` is more colorful, it's better to show this feature, right?
1.  add `setCursorEnabled(true);` to let developers know BMFont supports cursor too. I have tested it, still with a defect: __the cursor doesn't blink using BMFont, but cursor blink when using TTF font__. 
> in fact cursor doesn't blink using System Font(#18555 ), I have no idea with it

cocos/ui/UITextField.cpp
```diff
@@ -466,11 +469,17 @@ void TextField::setFontName(const std::string& name)
...
     {
-        TTFConfig config = _textFieldRenderer->getTTFConfig();
-        config.fontFilePath = name;
-        config.fontSize = _fontSize;
-        _textFieldRenderer->setTTFConfig(config);
-        _fontType = FontType::TTF;
+        if (name.find(".fnt") != std::string::npos) {
+            _textFieldRenderer->setBMFontFilePath(name);
+            _fontType = FontType::BMFONT;
+        }
```
drelaptop: consider the letters: `.fnt` case is more perfect, like this:
```C++
            std::string lowerCasedFontName = fontName;
            std::transform(lowerCasedFontName.begin(), lowerCasedFontName.end(), lowerCasedFontName.begin(), ::tolower);
            if (lowerCasedFontName.find(".fnt") != std::string::npos)
            {
                _tabLabelRender->setBMFontFilePath(fontName);
                _fontType = FontType::BMFONT;
            }
```

drelaptop (changes requested):
another word,  try to add proper comments to show this good feature.

tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UITextFieldTest/UITextFieldTest.cpp
roed: Will have a look at this. I think it has something to do with the position in the update method.

cocos/ui/UITextField.cpp
roed: Was thinking about this, but checking with npos is perhaps not the way to go. For instance, what if a font has a name like "some.fnt.ttf". Not likely to happen, but it could. Perhaps I could do:

```
std::string lcName = name;
std::transform(lcName.begin(), lcName.end(), lcName.begin(), ::tolower);
if(lcName.substr(lcName.length() - 4) == ".fnt") {
    //etc
```
What do you think?

roed:
@drelaptop perhaps renaming is the way to go and just typedef the old class to the new one and marking it as deprecated. There are other places in the codebase that use the same approach. This way the class is properly named and it doesn't bc-break anything, making upgrades still possible.

roed:
Also, I changed the unit test to use the font you suggested, but somehow I had to use half the font size defined in the fnt file for the cursor to work properly. Otherwise the cursor was half the size of the rest of the characters. Do you now why this is?

cocos/ui/UITextField.cpp
drelaptop: agree, this is more strict, `lowerCasedFontName.find(".fnt") != std::string::npos` is old logic to judge the type by font name, you can find it in other source files.

tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UITextFieldTest/UITextFieldTest.cpp
```diff
@@ -446,8 +446,8 @@ bool UITextFieldTest_BMFont::init()
         _uiLayer->addChild(alert);
         
         // Create the textfield
-        TextField* textField = TextField::create("input words here","fonts/markerFelt.fnt",25);
-        
+        TextField* textField = TextField::create("input words here","fonts/bitmapFontTest3.fnt",14);
+        textField->setCursorEnabled(true);
```
drelaptop: I found it too, the size is different between `BMFont Letter` and `Cursor`. 

another info is if we didn't keep the  `BMFont Cursor`  blink like `system font Cursor`, both size is same. it's so strange, I think a place existed the `FTF Cursor` is scaled, but `BMFont Cursor`  not. I didn't find this place for now.

roed: Yeah, that was my issue. Don't know how to solve it, as it seems to behave differently on different fonts. When you open a .fnt file, you can see the font size at the top. Somehow it works for markerFelt but not for the other font. As if it does some internal scaling (or perhaps because of something with retina?)

drelaptop: I hold the same idea with you, some internal scaling must exist. 

__typedef the old class to the new one__ sounds good, it's a way to solve compatibility. but I think we shouldn't do it until we solve the blink issue on `BMFont` and `system font`.

drelaptop: `void Label::updateLetterSpriteScale(Sprite* sprite)` this method can keep the _BMFont Cursor Sprite_ with other letters the same size, but when I try to add this into cursor update method, position isn't perfect. maybe we can find another place to put the scale method.

roed: I wil have a look later today. The cursor is blinking though, I fixed that.

roed: I found it! By default, a label will use the font size of the bmfont divided by `CC_CONTENT_SCALE_FACTOR()`. If you set a font size manually, which the widget does in this case (see TextField::create), the scale factor is not used. Perhaps we should add a different create method, just like Label has a Label::createWithBMFont method, which does not set the font size (as it will read it from the fnt file).
When I used the size 28/CC_CONTENT_SCALE_FACTOR(), it displayed correctly. So if you set a size manually and do a 'getLetter', it does not take the scale factor into account. Perhaps this is more a Label issue than a TextField issue.

roed:
@drelaptop I added the method createWithBMFont and the cursor now works in case of a BMFont. I will now look into the system font cursor support.

roed:
@drelaptop it is not possible to make the cursor blink with a system font, as it does not create individual sprites per character but one sprite for the whole text. It does this internally:

```
auto texture = new (std::nothrow) Texture2D;
texture->initWithString(_utf8Text.c_str(), fontDef);

_textSprite = Sprite::createWithTexture(texture);
```

I do not think it will be possible with this setup. (see getLetter, it return a nullptr when _textSprite is used). Perhaps if a systemfont would render internally to multiple sprites, but it does not. I think we should stop looking to fix that in this pr. What do you think?

drelaptop:
@roed  Looking into the code again and again, I found the best way to solve BMFont cursor blinking, it will scale properly with the font size changing. And I have submit a PR for you [HERE](https://github.com/toothpaste-and-bubblegum/cocos2d-x/pull/1), you can merge it and have a test. 

I agree with your opinion on the SystemFont cursor, we should stop that in this PR.

roed:
@drelaptop I merged your pr but have not tested it yet. Are you sure your changes are not going to break anything? For instance the other bmfont tests. Will have a look at those later today.

drelaptop:
It didn't break BMFont tests in `cpp-tests` project when I tested on macOS, I think making `Label::getLetter()` method support BMFont scale, it's good. after you review, and give your opinion. I will do further test on other platforms before merge.

roed:
@drelaptop I also checked your improvements on a macbook and they are good. Will do some testing on my iphone though

roed:
Works on my iPhone as well. The only issue I have is that people have to provide a font size for the bmfont, which you do not need to do when using the `Label::createWithBMFont` method (it will pick the font size defined in the fnt file). They have to look in the font to know what size they have to provide. If you have a font with really small lines (like pixel wide, Roboto for example) the rendering becomes very ugly if you scale it down on non-retina screens.

roed:
By the way, do you want me to rename (and deprecated typedef) the TextFieldTTF in this pr? Or shall I create another pr for that?

drelaptop:
I will test this PR on other platforms, and ask to merge this if it all works well. Your issue and your thinking of rename `TextFieldTTF ` is reasonable, I think creating a new PR for that is better.

drelaptop:
Have tested on macOS, iPhone, Android 7.0, win32 except Linux, this feature works well, please merge @minggo

roed:
Nice, thanks

PR #18591, opened 2017-12-26
Fix crash - wrong memory is zeroed after previous commit changed var …

eligt:
…name 'deform' to 'deform2'


eligt:
Was caused by https://github.com/cocos2d/cocos2d-x/commit/289668015d8d53056c6a67c393373c4bfe89e3e9

drelaptop:
sorry for this issue, we have a existed PR #18582 , which isn't been merged because this [spine-runtimes PR](https://github.com/EsotericSoftware/spine-runtimes/pull/1053) isn't merged. 

even though, I strongly suggest to merge #18582, it's quite better for developers. another reason we have submit a PR for EsotericSoftware/spine-runtimes, but no response for 5 days. @minggo

PR #18598, opened 2017-12-29
[Label] use freetype api to support bold - improvement && tests case

drelaptop:
> label enable blod use shadow to achive, so bold and outline can not use in same time.
> use freetype api to support bold

This feature is submitted at #17436 months ago, this is a good feature, but the author didn't response for a long time. So I decide to submit it again, improve it, and add a test case.

### DIFF
--- a/cocos/2d/CCFontAtlasCache.cpp
+++ b/cocos/2d/CCFontAtlasCache.cpp
@@ -58,11 +58,11 @@ FontAtlas* FontAtlasCache::getFontAtlasTTF(const _ttfConfig* config)
 
     char tmp[ATLAS_MAP_KEY_BUFFER];
     if (useDistanceField) {
-        snprintf(tmp, ATLAS_MAP_KEY_BUFFER, "df %.2f %d %s", config->fontSize, config->outlineSize,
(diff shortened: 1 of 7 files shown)


drelaptop:
PR #17436  have some defect, when I added test case, and do further test, I found it works well when a Label enable Outline and Bold both, but looks ugly when a Label enable Bold only. look this:

1. enable Outline and Bold both

bold and outline

2. enable Bold only

bold only

this is my test case [LabelTestNew.cpp#L3007](https://github.com/drelaptop/cocos2d-x/blob/4fc89dc5cceb4f00bb8c4e3a4620a66fe7b55fb4/tests/cpp-tests/Classes/LabelTest/LabelTestNew.cpp#L3007)

drelaptop:
__Not merge__, because it works well when a Label enable Outline and Bold both, but looks ugly when a Label enable Bold only. might a internal defect existed on the using of FreeType

drelaptop:
__Not merge__, because it works well when a Label enable Outline and Bold both, but looks ugly when a Label enable Bold only. might a internal defect existed on the using of FreeType

Issue #17962, opened 2017-06-21
App was stuck in black all the time when running on android studio emulator


guoyingtao:
- cocos2d-x version:
- devices test on:
- developing environments
   -NDK version: 14.1
   -Android Studio 2.3.3

Steps to Reproduce:

1. Create a new project and run it on Simulator

After the app started, it was stuck in black all the time. I noticed this topic [#17850](https://github.com/cocos2d/cocos2d-x/issues/17850) also met the similar thing, but my logcat gave different information like below.

06-21 03:25:04.710 16078-16078/com.MyCompany.MyGame W/art: Failed execv(/system/bin/dex2oat --runtime-arg -classpath --runtime-arg  --debuggable --instruction-set=x86 --instruction-set-features=smp,ssse3,-sse4.1,-sse4.2,-avx,-avx2 --runtime-arg -Xrelocate --boot-image=/system/framework/boot.art --runtime-arg -Xms64m --runtime-arg -Xmx512m --instruction-set-variant=x86 --instruction-set-features=default --dex-file=/data/app/com.MyCompany.MyGame-2/split_lib_dependencies_apk.apk --oat-file=/data/dalvik-cache/x86/data@app@com.MyCompany.MyGame-2@split_lib_dependencies_apk.apk@classes.dex) because non-0 exit status

guoyingtao:
**Debug info**

06-22 12:02:40.725 3050-3097/com.MyCompany.MyGame D/cocos2d-x debug info: {
                                                                          	gl.supports_OES_packed_depth_stencil: true
                                                                          	gl.supports_vertex_array_object: false
                                                                          	gl.supports_BGRA8888: false
                                                                          	gl.supports_ATITC: false
                                                                          	gl.supports_S3TC: false
                                                                          	cocos2d.x.version: cocos2d-x-3.15.1
                                                                          	gl.supports_discard_framebuffer: false
                                                                          	cocos2d.x.compiled_with_profiler: false
                                                                          	gl.supports_PVRTC: false
                                                                          	cocos2d.x.build_type: DEBUG
                                                                          	gl.renderer: Android Emulator OpenGL ES Translator (AMD Radeon R9 M370X OpenGL Engine)
                                                                          	gl.supports_OES_depth24: true
                                                                          	gl.supports_ETC1: true
                                                                          	gl.supports_OES_map_buffer: false
                                                                          	cocos2d.x.compiled_with_gl_state_cache: true
[...]

PR #18606, opened 2018-01-02
Android Emulator Blank Screen 0x501 and 0x502 Problem Bugfix Update

gamehelphere:
I added some code to make the Android Emulator read GLSL files correctly by making the data types constant for uniform variables. If the Android Emulator cannot read the data types of the variables, it will say OpenGL Error 0x501 or 0x502 glUniform unable to find uniform variable depth. Other variables will be affected after depth. With this bugfix, the variables will be read correctly by the Android Emulator and render your game.

I removed too much lines earlier including the original lines for WINRT and have placed those back with the bugfix. I needed to remove the earlier pull request to avoid confusion.

### DIFF
--- a/cocos/renderer/CCGLProgram.cpp
+++ b/cocos/renderer/CCGLProgram.cpp
@@ -487,7 +487,7 @@ bool GLProgram::compileShader(GLuint * shader, GLenum type, const GLchar* source
         headersDef = (type == GL_VERTEX_SHADER ? "precision mediump float;\n precision mediump int;\n" : "precision mediump float;\n precision mediump int;\n");
 // Bugfix to make shader variables types constant to be understood by the current Android Virtual Devices or Emulators. This will also elimiate the 0x501 and 0x502 OpenGL Errors during emulation.
 #elif CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID
-              headersDef = "#version 100\n precision mediump float;\n precision mediump int;\n";
+        headersDef = "#version 100\n precision mediump float;\n precision mediump int;\n";
 #elif (CC_TARGET_PLATFORM != CC_PLATFORM_WIN32 && CC_TARGET_PLATFORM != CC_PLATFORM_LINUX && CC_TARGET_PLATFORM != CC_PLATFORM_MAC)
         headersDef = (type == GL_VERTEX_SHADER ? "precision highp float;\n precision highp int;\n" : "precision mediump float;\n precision mediump int;\n");
 #endif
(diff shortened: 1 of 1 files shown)


drelaptop:
Thank you for this patch, which have a long discussion on #17962. it seems like a important patch about OpenGL on Android Emulator. please review @minggo , and merge if pass.

minggo:
@gamehelphere will you bring side-effect to Android devices?

gamehelphere:
Hi @drelaptop ,

You are welcome. Thank you also for checking my patch.

gamehelphere:
Hi @minggo ,

Do you mean about side effects like the patched source code would work on Android Emulators and that same code would not work on some actual Android devices? I think the patched source code will work on both. I have tested the patch with my local setup both in some emulators and my cellphone that is using Android 6.0.1. My apps ran using SwiftShader on the Android Emulator, and ran accelerated on my cellphone. There were some who tested the patch on their setups with different operating systems and Android cellphones, and the apps worked. More details can be viewed on the issue.

minggo:
thanks

gamehelphere:
Hi @minggo ,

You are welcome. Thank you also for merging my patch to the 3.17 milestone sources.

minggo:
@gamehelphere some developer find sprite joggle caused by this PR, so i think we should change `mediump` to `highp` for float.

gamehelphere:
@minggo I downloaded the current 3.17 main branch sources with the current patch and built it with Android Studio 3.4. I made a Pixel XL emulator and there was no sprite joggle in my basic Hello World app. I made the changes on the shader data types in both float from mediump to highp, yet there is still no sprite joggle. I do not know how this problem happened. Maybe some later emulator and hardware models of the phones have this problem.

I will test the highp changes in my own Android Marshmallow phone and let you know if there are any sprite joggles in the basic Hello World app. If there are no problems, I will let you know and send the new shader changes as a patch.

minggo:
@gamehelphere thanks. It is hard to find the joggle. You can put a sprite drawn in a background, then drag the background back and forth by a little distance. You will find the sprite's position joggle. And the developer said, it only happened on huawei serial devices.

gamehelphere:
You are welcome, @minggo . Thank you also for letting me know about this. I changed the data types as you suggested, and tested it both in my Pixel XL emulator and Android Marshmallow Samsung phone. There was no sprite joggling when I made the background move behind a sprite in the scene. Maybe the change will help remove the problem in the affected devices like the Huawei serial devices.

I have uploaded the patch with the new data types.

minggo:
thanks

Issue #18609, opened 2018-01-04
Render error when call CCSprite::initWithSpriteFrameName 2nd time


peerobo:
- cocos2d-x version: 3.17
- devices test on: HTC One (m7)
- developing environments
   - NDK version: r15c
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:

The following code call to CCSprite::initWithSpriteFrameName two times, cause an error in rendering.

`ui::Button *bt = ui::Button::create();`
`bt->loadTextureNormal("SpriteFrameA", ui::Widget::TextureResType::PLIST);`
`bt->loadTextureNormal("SpriteFrameB", ui::Widget::TextureResType::PLIST);`

![screenshot_2018-01-04-18-24-14](https://user-images.githubusercontent.com/4023681/34561545-3203ea14-f17d-11e7-95c7-3e408efdbe7c.png)

Should look like this: 

![screenshot_2018-01-04-18-32-24](https://user-images.githubusercontent.com/4023681/34561681-bda83ab6-f17d-11e7-912f-531008e3f277.png)

Achieve by modify UIButton.cpp:

`void Button::loadTextureNormal(const std::string& normal,TextureResType texType)`
`{`
`  ...`
`      case TextureResType::PLIST:`
`          // _buttonNormalRenderer->initWithSpriteFrameName(normal);`
`          frm = SpriteFrameCache::getInstance()->getSpriteFrameByName(normal);`
`          _buttonNormalRenderer->setSpriteFrame(frm);`
`          break;`
` ...`
`  }`

drelaptop:
Could you submit a PR for it? to add a test case like your image shows into `cpp-tests` project -> `45:Node: Sprite` would be better.

peerobo:
@drelaptop please check out the PR. Thanks!

PR #18611, opened 2018-01-05
Add Test Case: UIButton::loadTextureNormal 2nd time error

peerobo:
Test case for this issue
#18609


drelaptop:
You said fixing this issue need to modify `UIButton.cpp` in #18609, however I didn't see the change about that on __Files Changed__, is it a mistake?

peerobo:
I thought you wanted a test case to see what was the error.  I will commit my temporary fix later.

drelaptop:
please see CI log to solve the checks failed @peerobo

peerobo:
@drelaptop please check my fix if you have time. Thanks.

cocos/ui/UIButton.cpp
```diff
@@ -369,7 +377,9 @@ void Button::loadTextureDisabled(const std::string& disabled,TextureResType texT
             _buttonDisabledRenderer->initWithFile(disabled);
             break;
         case TextureResType::PLIST:
-            _buttonDisabledRenderer->initWithSpriteFrameName(disabled);
+            // _buttonDisabledRenderer->initWithSpriteFrameName(disabled);
+            frm = SpriteFrameCache::getInstance()->getSpriteFrameByName(disabled);
+            _buttonDisabledRenderer->setSpriteFrame(frm);
```
drelaptop, line 383: Confirm the issue existed, we need fix it. do you know why it happened? I am not sure this way fixing the bug is best.

one reason is in `initWithSpriteFrameName ` called `SpriteFrameCache::getInstance()->getSpriteFrameByName(disabled);`, `setSpriteFrame ` and the other method, so I think the bug is caused due to internal of `initWithSpriteFrameName`.

another reason is `initWithSpriteFrameName` do many checks, calling `getSpriteFrameByName ` and `setSpriteFrame ` directly do no checks, if I code:
```C++
spr->loadTextureNormal("Not Existed Texture", ui::Widget::TextureResType::PLIST);
```
the app will crash.

peerobo: I know the issue is within `initWithSpriteFrameName()` but I do not have time right now to debug. This is a temporary fix to make my project work.

drelaptop: Understand, and I am doing something about cmake firstly, so let it stay here for now.

PR #18613, opened 2018-01-05
Add option to set keyboard appearance for edit boxes

dennacerise:
We have found it useful to change the theme of the soft keyboard. iOS has several keyboard appearances available. Here, I've added support for DEFAULT and DARK keyboard appearance types. `setKeyboardAppearance()` is not implemented for the other platforms because it doesn't make sense to update the keyboard theme on the other platforms.

### DIFF
--- a/cocos/ui/UIEditBox/UIEditBox.h
+++ b/cocos/ui/UIEditBox/UIEditBox.h
@@ -210,6 +210,15 @@ namespace ui {
              */
             LOWERCASE_ALL_CHARACTERS
         };
+        
+        /**
+         * @brief The EditBox::KeyboardAppearance defines the keyboard's appearance
+         * while the edit box is active.
+         */
+        enum class KeyboardAppearance {
+            DEFAULT,
+            DARK
+        };
             
         /**
          * create a edit box with size.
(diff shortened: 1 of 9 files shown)


drelaptop:
@dennacerise thanks for this feature, please check the CI error, and fix it

dennacerise:
Looks like a couple of the builds timed out. Should I restart?..

drelaptop:
not need, I will review and test this PR in recent days

cocos/ui/UIEditBox/UIEditBox.h
```diff
@@ -483,7 +492,19 @@ namespace ui {
          * @return One of the EditBox::KeyboardReturnType constants.
          */
         KeyboardReturnType getReturnType() const;
-
+        
+        /**
+         * Set the keyboard appearance that is to be applied to the edit box.
+         * @param keyboardAppearance One of the EditBox::KeyboardAppearance constants.
+         */
+        void setKeyboardAppearance(KeyboardAppearance keyboardAppearance);
```
drelaptop, line 500: Suggest to comment this method doesn't support others platforms except iOS, so as to avoid conflict when android developer try to use it

cocos/ui/UIEditBox/UIEditBoxImpl-common.h
```diff
@@ -118,6 +120,7 @@ class CC_GUI_DLL EditBoxImplCommon : public EditBoxImpl
     virtual void setNativeInputMode(EditBox::InputMode inputMode) = 0;
     virtual void setNativeInputFlag(EditBox::InputFlag inputFlag) = 0;
     virtual void setNativeReturnType(EditBox::KeyboardReturnType returnType) = 0;
+    virtual void setNativeKeyboardAppearance(EditBox::KeyboardAppearance) {}
```
drelaptop, line 123: shouldn't empty realisation, suggest keep pure virtual function, add empty realize function for other platforms, and comment didn't finish. I think it's better to add assertion so that  assertion error occurs when a android developer using this method and try to compile.

drelaptop (changes requested):
Need to add a test case when a new feature is added, please refer to other EditBox case of `cpp-tests` project.
> Path: Node::UI - GUI Test - EditBox Test

PR #18619, opened 2018-01-11
fix error with xcode7: The document "LaunchScreen.storyboard" require…

summerinsects:
fix error with xcode7: The document "LaunchScreen.storyboard" requires Xcode 8.0 or later

——if not merge this PR, maybe we should modify README.md

Build Requirements
Mac OS X 10.7+, Xcode 7+ -> Xcode 8+


minggo:
Yep, that's change the read me, because LaunchScreen.storyboard is needed now.

PR #18626, opened 2018-01-16
Fix a crash bug when the UserDefault.xml file is empty.

monkey256:
When the UserDefault.xml file in local disk has no content, the program will crash with using UserDefault::setStringForKey function.

### DIFF
--- a/cocos/base/CCUserDefault.cpp
+++ b/cocos/base/CCUserDefault.cpp
@@ -57,25 +57,40 @@ static tinyxml2::XMLElement* getXMLNodeForKey(const char* pKey, tinyxml2::XMLEle
 
     do 
     {
-         tinyxml2::XMLDocument* xmlDoc = new (std::nothrow) tinyxml2::XMLDocument();
+        tinyxml2::XMLDocument* xmlDoc = new (std::nothrow) tinyxml2::XMLDocument();
         *doc = xmlDoc;
+        *rootNode = nullptr;
 
         std::string xmlBuffer = FileUtils::getInstance()->getStringFromFile(UserDefault::getInstance()->getXMLFilePath());
 
-        if (xmlBuffer.empty())
+        if (!xmlBuffer.empty())
         {
-            CCLOG("can not read xml file");
-            break;
-        }
-        xmlDoc->Parse(xmlBuffer.c_str(), xmlBuffer.size());
+            xmlDoc->Parse(xmlBuffer.c_str(), xmlBuffer.size());
 
-        // get root node
-        *rootNode = xmlDoc->RootElement();
+            // get root node
+            *rootNode = xmlDoc->RootElement();
+        }
+
         if (nullptr == *rootNode)
         {
-            CCLOG("read root node error");
-            break;
+            if (!xmlDoc->FirstChild())
+            {
+                tinyxml2::XMLDeclaration *pDeclaration = xmlDoc->NewDeclaration(nullptr);
+                if (nullptr != pDeclaration)
+                {
+                    xmlDoc->LinkEndChild(pDeclaration);
+                }
+            }
+
+            tinyxml2::XMLElement *pRootEle = xmlDoc->NewElement(USERDEFAULT_ROOT_NAME);
+            if (nullptr != pRootEle)
+            {
+                xmlDoc->LinkEndChild(pRootEle);
+            }
+
+            *rootNode = pRootEle;
         }
+
         // find the node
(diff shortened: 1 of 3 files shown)


cocos/base/CCUserDefault.cpp
```diff
@@ -57,25 +57,40 @@ static tinyxml2::XMLElement* getXMLNodeForKey(const char* pKey, tinyxml2::XMLEle
...
+                if (nullptr != pDeclaration)
+                {
+                    xmlDoc->LinkEndChild(pDeclaration);
+                }
+            }
+
+            tinyxml2::XMLElement *pRootEle = xmlDoc->NewElement(USERDEFAULT_ROOT_NAME);
+            if (nullptr != pRootEle)
+            {
+                xmlDoc->LinkEndChild(pRootEle);
```
minggo: pRootEle -> rootEle, cocos2d-x doesn't use Hungarian style now, and please modify other places.

cocos/editor-support/cocostudio/ActionTimeline/CCFrame.cpp
```diff
@@ -149,7 +149,8 @@ TextureFrame* TextureFrame::create()
 }
 
 TextureFrame::TextureFrame()
-    : _textureName("")
+    : _sprite(nullptr)
+    , _textureName("")
```
minggo, line 153: Is the modification related to the issue?

minggo:
Could you please tell me how to reproduce the issue?

cocos/editor-support/cocostudio/ActionTimeline/CCFrame.cpp
monkey256: The following two modifications is not related to this issue, it crashed accidentally. I found it because of the problems caused by the constructor, some variables have not been initialized.

monkey256:
To reproduce this issue as follows: (win32 environment)
Make the UserDefault.xml file empty(file exists but has no content), and then consider that:
``` c++
UserDefault* UserDefault::getInstance()
{
    if (!_userDefault)
    {
        initXMLFilePath();

        // only create xml file one time
        // the file exists after the program exit
        if ((!isXMLFileExist()) && (!createXMLFile()))
        {
            return nullptr;
        }

        _userDefault = new (std::nothrow) UserDefault();
    }

    return _userDefault;
}
```
The xml file will not been recreated, then look at the following codes:
``` c++
static void setValueForKey(const char* pKey, const char* pValue)
{
    tinyxml2::XMLElement* rootNode;
    tinyxml2::XMLDocument* doc;
    tinyxml2::XMLElement* node;
    // check the params
    if (! pKey || ! pValue)
    {
        return;
    }
    // find the node
    node = getXMLNodeForKey(pKey, &rootNode, &doc);
    ...
}


static tinyxml2::XMLElement* getXMLNodeForKey(const char* pKey, tinyxml2::XMLElement** rootNode, tinyxml2::XMLDocument **doc)
{
    tinyxml2::XMLElement* curNode = nullptr;

    // check the key value
    if (! pKey)
    {
        return nullptr;
    }

    do 
    {
         tinyxml2::XMLDocument* xmlDoc = new (std::nothrow) tinyxml2::XMLDocument();
        *doc = xmlDoc;

        std::string xmlBuffer = FileUtils::getInstance()->getStringFromFile(UserDefault::getInstance()->getXMLFilePath());

        if (xmlBuffer.empty())
        {
            The file is empty, break here! without reset rootNode.
            CCLOG("can not read xml file");
            break;
        }

        ...
    } while (0);

    return curNode;
}
```

So the function 'setValueForKey' will use an uninitialized variable 'rootNode', and then crashed.

cocos/base/CCUserDefault.cpp
monkey256: Got it! I'll check out later.

minggo:
> Make the UserDefault.xml file empty(file exists but has no content)

Yep, it has issue in the situation. But in what situations that will have an empty UserDefault.xml file?

monkey256:
I don't know yet, but one player met this situation in win32 platform. I'm curious too.

Canon0223:
> > Make the UserDefault.xml file empty(file exists but has no content)
> 
> Yep, it has issue in the situation. But in what situations that will have an empty UserDefault.xml file?

This problem often occurs in my project, but it has not been reproduced stably. I guess it is caused by simultaneous read and write operations on CCUserDefault.cpp.Only in Win32 platform

PR #18633, opened 2018-01-17
V3 websocket dead lock fix

newnon:
WebSocket deadlock fix on close

### DIFF
--- a/cocos/network/WebSocket.cpp
+++ b/cocos/network/WebSocket.cpp
@@ -75,6 +75,8 @@ static void _winLog(const char *format, va_list args)
             bufferSize *= 2;
 
             delete[] buf;
+            
+            return;
         }
         else
             break;
@@ -391,7 +393,7 @@ void WsThreadHelper::onSubThreadStarted()
     int log_level = LLL_ERR | LLL_WARN | LLL_NOTICE | LLL_INFO/* | LLL_DEBUG | LLL_PARSER | LLL_HEADER*/ | LLL_EXT | LLL_CLIENT | LLL_LATENCY;
     lws_set_log_level(log_level, printWebSocketLog);
 
-    memset(__defaultProtocols, 0, sizeof(2 * sizeof(struct lws_protocols)));
+    memset(__defaultProtocols, 0, 2 * sizeof(struct lws_protocols));
 
     __defaultProtocols[0].name = "";
     __defaultProtocols[0].callback = WebSocketCallbackWrapper::onSocketCallback;
@@ -646,6 +648,7 @@ void WebSocket::send(const std::string& message)
         // In main thread
         Data* data = new (std::nothrow) Data();
         data->bytes = (char*)malloc(message.length() + 1);
+        assert(data->bytes);
         // Make sure the last byte is '\0'
         data->bytes[message.length()] = '\0';
         strcpy(data->bytes, message.c_str());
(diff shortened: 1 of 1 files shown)


cocos/network/WebSocket.cpp
```diff
@@ -391,7 +393,7 @@ void WsThreadHelper::onSubThreadStarted()
     int log_level = LLL_ERR | LLL_WARN | LLL_NOTICE | LLL_INFO/* | LLL_DEBUG | LLL_PARSER | LLL_HEADER*/ | LLL_EXT | LLL_CLIENT | LLL_LATENCY;
     lws_set_log_level(log_level, printWebSocketLog);
 
-    memset(__defaultProtocols, 0, sizeof(2 * sizeof(struct lws_protocols)));
+    memset(__defaultProtocols, 0, 2 * sizeof(struct lws_protocols));
```
dumganhar, line 396: Good catch.

cocos/network/WebSocket.cpp
```diff
@@ -75,6 +75,8 @@ static void _winLog(const char *format, va_list args)
             bufferSize *= 2;
 
             delete[] buf;
+            
+            return;
```
dumganhar, line 79: Why to return directly here. 
The while loop will go to next step to use double size buffer for formatting the string.

dumganhar: Why to return directly here. 
The while loop will go to next step to use double size buffer for formatting the string.

cocos/network/WebSocket.cpp
```diff
@@ -646,6 +648,7 @@ void WebSocket::send(const std::string& message)
         // In main thread
         Data* data = new (std::nothrow) Data();
         data->bytes = (char*)malloc(message.length() + 1);
+        assert(data->bytes);
```
dumganhar, line 651: Why to assert here? Does it possible return nullptr?

cocos/network/WebSocket.cpp
```diff
@@ -722,8 +726,8 @@ void WebSocket::close()
 
     {
         std::unique_lock<std::mutex> lkClose(_closeMutex);
-        _closeCondition.wait(lkClose);
         _closeState = CloseState::SYNC_CLOSED;
+        _closeCondition.notify_one();
```
dumganhar, line 730: if `onConnectionClosed` lasts for more than 5ms, it may cause crash issue. Since user's onClose will delete websocket instance, then onConnectionClosed logic may still be executed which will cause crash.
```c++
    // Wait 5 milliseconds for onConnectionClosed to exit!
    std::this_thread::sleep_for(std::chrono::milliseconds(5));
    _delegate->onClose(this);
```
In fact, it's not dead lock in `close`, it's probably stop main thread for several seconds since `close` is a synchronous method.
If developer wanna use asynchronous method, he/she has to use `WebSocket::closeAsync` interface.

cocos/network/WebSocket.cpp
newnon: yes it is my mistake i will remove this

cocos/network/WebSocket.cpp
newnon: yes if not enough memory and it is more to make static analyser happy.

newnon:
ok i will close it in favor of https://github.com/cocos2d/cocos2d-x/pull/18643

PR #18635, opened 2018-01-17
Label & Text gradient support

newnon:
Horizontal & Vertical gradient support for Labels


minggo:
Could you please add a test case for them?

PR #18637, opened 2018-01-17
android emulator stensil fix

newnon:
there is no more need to use this Android Simulator support all modes


minggo:
Why there is not need to do it like this?

newnon:
1. it disables stencil buffer(it can break logic dependence on it)
2. it not needed now new Android simulator can handle most pixel modes

minggo:
I am not sure, but please comment it instead of delete it. Then we can enable it again easily if needed.

newnon:
done

PR #18638, opened 2018-01-17
V3 static analize fixes

newnon:


minggo (approved):

cocos/base/CCScheduler.cpp
```diff
@@ -457,10 +458,11 @@ void Scheduler::appendIn(_listEntry **list, const ccSchedulerFunc& callback, voi
     DL_APPEND(*list, listElement);
 
     // update hash entry for quicker access
-    tHashUpdateEntry *hashElement = (tHashUpdateEntry *)calloc(sizeof(*hashElement), 1);
+    tHashUpdateEntry *hashElement = new (std::nothrow) tHashUpdateEntry;
```
dumganhar: ??
tHashUpdateEntry is freed by `free`.
You could refer to https://github.com/newnon/cocos2d-x/blob/f903842c7c374b4a70bd5bd656c83d4d23efc58f/cocos/base/CCScheduler.cpp#L555

dumganhar: Using `new` will cause memory problems.

newnon:
@dumganhar fixed

Issue #18465, opened 2017-11-13
Game stops music at opening - Android regression compare to cocos2d-x.3.14.1


crazyhappygame:
- cocos2d-x version: 3.16
- devices test on: Android/all
- developing environments
   - NDK version:  android-ndk-r14b 
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:

1. Play music on Android  device
2. Run any cocos2d-x app 

Bug:
cocos2d-x app stops music at opening:
This regression compare to cocos2d-x.3.14.1

crazyhappygame:
@dumganhar Is there any chance for some quick workaround?

dumganhar:
Perhaps you should modify code in Cocos2dxActivity.java
```java
    @Override
    protected void onResume() {
    	Log.d(TAG, "onResume()");
        super.onResume();
        AudioManager manager =
            (AudioManager)this.getSystemService(Context.AUDIO_SERVICE);
       if (!manager.isMusicActive()) {
             Cocos2dxAudioFocusManager.registerAudioFocusListener(this);
       }
        this.hideVirtualButton();
       	resumeIfHasFocus();
```

AhmetAbak:
@dumganhar this solves the problem.

crazyhappygame:
@dumganhar, AhmetAbak this change should be done in cocos2dx code or game code?

crazyhappygame:
@dumganhar could you also explain issue and how this change fix it? It is "proper fix" or some workaround?

AhmetAbak:
@crazyhappygame i modified the Cocos2dxActivity class in cocos2dx library.

dumganhar:
@AhmetAbak , good to know it works for you.

I think most games will take the ownership of audio, so it's a specific requirement.
Probably, we need to add a new interface for setting this.

dumganhar:
>  could you also explain issue and how this change fix it? It is "proper fix" or some workaround?

It need to be tested. I think it's a fix not just a workaround.
But first of all, we need to define a new method for setting this.
Could you guys test this patch and give me the feedback? Thanks.

AhmetAbak:
@dumganhar 
[crash log](https://github.com/cocos2d/cocos2d-x/files/1477474/com.sngict.sngsolitaire_issue_135_crash_5A0D2270019500016317E2E7A9E3E25D_4b3a8580ca8f11e79ca756847afe9799_0_v2.txt)
[...]

PR #18640, opened 2018-01-17
fix external player muting. add audio focus flag

newnon:
Add ability to not stop sound & music on game start in android

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
@@ -68,6 +68,8 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
     private Cocos2dxEditBoxHelper mEditBoxHelper = null;
     private boolean hasFocus = false;
     private boolean showVirtualButton = false;
+    private boolean gainAudioFocus = false;
+    private boolean paused = true;
 
     public Cocos2dxGLSurfaceView getGLSurfaceView(){
         return  mGLSurfaceView;
@@ -89,6 +91,18 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
 
     public void setEnableVirtualButton(boolean value) {
         this.showVirtualButton = value;
+    }
+
+    public void setEnableAudioFocusGain(boolean value) {
+        if(gainAudioFocus != value) {
+            if(!paused) {
+                if (value)
+                    Cocos2dxAudioFocusManager.registerAudioFocusListener(this);
+                else
+                    Cocos2dxAudioFocusManager.unregisterAudioFocusListener(this);
+            }
+            gainAudioFocus = value;
+        }
     }
 
     protected void onLoadNativeLibraries() {
@@ -167,8 +181,10 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
     @Override
     protected void onResume() {
     	Log.d(TAG, "onResume()");
+        paused = false;
         super.onResume();
-        Cocos2dxAudioFocusManager.registerAudioFocusListener(this);
+        if(gainAudioFocus)
+            Cocos2dxAudioFocusManager.registerAudioFocusListener(this);
         this.hideVirtualButton();
        	resumeIfHasFocus();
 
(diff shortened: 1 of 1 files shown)


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
```diff
@@ -91,6 +91,10 @@ public void setEnableVirtualButton(boolean value) {
         this.showVirtualButton = value;
     }
 
+    public void setEnableAudioFocusGain(boolean value) {
+        Cocos2dxAudioFocusManager.gainAudioFocus = value;
```
dumganhar: I think we should not define a variable `gainAudioFocus` in Cocos2dxAudioFocusManager which will modify the behavior of `registerAudioFocusListener`.
Instead, Cocos2dxActivity should have a flag for determine whether to invoke `registerAudioFocusListener`.

Besides, I prefer the function name `setAudioFocusGainEnabled`

newnon:
@dumganhar ok i'll rewrite it

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
mars3142: And it should be saved to the preferences, so a killed app can load the last setting. And we should register/unregister it also in this set `setAudioFocusGainEnabled`. 

I also would add a function for the wanted AudioFocus value `AUDIOFOCUS_LOST`, `AUDIOFOCUS_LOST_TRANSIENT` and `AUDIOFOCUS_LOST_TRANSIENT_CAN_DUCK`. Maybe a game doesn't want to gain the full focus.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxAudioFocusManager.java
```diff
@@ -36,6 +36,7 @@
     private final static int AUDIOFOCUS_LOST = 1;
     private final static int AUDIOFOCUS_LOST_TRANSIENT = 2;
     private final static int AUDIOFOCUS_LOST_TRANSIENT_CAN_DUCK = 3;
+    public static boolean gainAudioFocus = true;
```
crazyhappygame: Please change default to false.
@mars3142 Could you support me :)

newnon: ok but default cocos behaviour was false

crazyhappygame: For me that was regression introduced in 3.14 https://github.com/cocos2d/cocos2d-x/issues/18465
I believe that be default user should be able to play music while playing

crazyhappygame: @newnon Could you apply requested changes? 
My users complains all the time about that they can listen music while playing ...

crazyhappygame: @newnon could you push this PR to master

newnon:
Refactored

crazyhappygame:
@dumganhar Could you merge this fix to cocos2d-x.3.17?

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
dumganhar: @mars3142, sorry for my late reply.
`Saved to the preferences` is a good way on user code, but I don't think it should be put into Engine. Engine just provide the ability for enabling / disabling Audio Focus Gain.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxAudioFocusManager.java
dumganhar: > ok but default cocos behaviour was false

@newnon , the default behavior is true as I know. 
@crazyhappygame , if we change it to false, it will break the compatibility. :(  Any thoughts?

By the way, the member variable naming in Java code should follow `m` as prefix.

newnon:
As for me until we doesn't write any music game it is better not to gain audio focus, also it is default behaviour for ios.

crazyhappygame:
@dumganhar @minggo could you merge it before cocos2d-x.3.17 ?

crazyhappygame:
@newnon could rebase to latest version to trigger CI build?

minggo:
v3.17 will be released soon, so i will not merge it into v3.17. I will merge it in v3.

PR #18645, opened 2018-01-18
Disable 2 appveyor android build configurations

crazyhappygame:
Currently appveyor needs ~2:20h.
After this change it should be 1:20h (40 minutes windows32, 40 minutes android cpp_tests)


crazyhappygame:
@minggo Could you also set "Rolling builds" in appveyor?
https://www.appveyor.com/docs/build-configuration/#rolling-builds
Check 'Settings > General > Rolling builds'

crazyhappygame:
We can enable this for travis:
https://blog.travis-ci.com/2017-03-22-introducing-auto-cancellation

minggo:
Android is tested in travis, so i think we don't have to test Android on win32 again.

Ok, i will modify the appveyor and travis settings.

minggo:
Does auto-cancellation mean to cancel old PR and only build latest PR? If so, i don't think it is what we need.

minggo:
I found auto-cancel is enabled of branch and pull request, i think only pull request is needed. That's the reason why i found many pull requests are cancelled automatically. I don't remember i enabled them.

crazyhappygame:
In appveyor "rolling builds" keep running only latest build for branch and PR. It helps. Only 3 builds are running right now( branch v3 and 2 PR)
In travis "auto cancel" cancel only not started builds. Because in travis we are waiting for Mac and Linux build are started/ completed there is 21 Travis builds in progress

crazyhappygame:
Auto cancel/rolling builds  is per PR/branch.  All PR are build. In one PR only latest build is running. Older builds inside one PR are cancelled.

PR #18646, opened 2018-01-19
V3 android cmake support, add mac/ios support

drelaptop:
Based on #18154 V3 android cmake support author: @newnon 

1. add Lua & JS test project support
2. add Lua & JS template support
3. change cmake struct of organizing tests and template project
4. fix cmake build error on mac, all project run success now
5. start to support `cmake ../ -G "Xcode"`, need to do some improvements


drelaptop:
test project and template project all build success by Cradle and CMake, please review @minggo

minggo:
@drelaptop great job. May be you can write something about how to test it, then more developers can help to test it.

CMakeLists.txt
```diff
@@ -42,9 +42,10 @@ project (Cocos2d-X)
 set(COCOS2D_X_VERSION 3.13)
 
 # define some variables
-
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${PROJECT_SOURCE_DIR}/cmake/Modules/")
 set(COCOS_EXTERNAL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/external)
+set(COCOS_STATIC_LIBS_PATH ${PROJECT_SOURCE_DIR}/prebuilt)
```
minggo: COCOS_STATIC_LIBS_PATH -> COCOS_PREBUILT_LIBS_PATH

It makes it more meaningful.

CMakeLists.txt
```diff
@@ -97,32 +98,77 @@ endif()
...
-  add_subdirectory(tests/cpp-empty-test)
-endif(BUILD_CPP_EMPTY_TEST)
-
-# build cpp-tests
-if(BUILD_CPP_TESTS)
-  add_subdirectory(tests/cpp-tests)
-endif(BUILD_CPP_TESTS)
-
-## Scripting
+if(USE_COCOS_STATIC_LIBS)
```
minggo: USE_COCOS_STATIC_LIBS -> USE_COCOS_PREBUILT_LIBS

CMakeLists.txt
```diff
@@ -97,32 +98,77 @@ endif()
...
+  # set the staic lib name needed to build
+  if(BUILD_CPP_LIBS)
+    SET(NEED_LIB_NAME cocos2d)
```
minggo: It is better to use `SET` or `set` for all codes.

crazyhappygame:
Are you sure that compilation passed?
I just checked travis and appveyor CI and Android compilation fails.

drelaptop:
This is the error
```sh
      Cir:/s: projects/cocos2d-x/exC:/proternal/android-specific/pvmp3dec
    does not contain a CMakeLists.tjxtects file/co.cos2d-x/tests/cpp-
  
     C  C:/:/pprojerojcts/cectocos2d-x/external/as/cndroiocod-specific/tremolos2d-x/
    doetess not containts/ a CMakecppLists-te.txt stsfile./../../ext
```
it is caused by the related PR in third library not merged, see: https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/300, I will ask minggo to merge it after the cmake work having a milestone

another word, you need  https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/300 when you try to use the cmake in your PC for now

crazyhappygame:
Could you try to enable pch (precompiled headers) for android build (or even for all cmake platforms)? On windows it is 0:22:46 vs 0:01:51 for libcocos2d 12x speedup! 
Please check https://github.com/cocos2d/cocos2d-x/pull/18662

drelaptop:
@crazyhappygame yes, I will do that pch for all cmake paltforms finally. this is the plan http://discuss.cocos2d-x.org/t/in-progress-cmake-support-all-platform-for-cocos2d-x/41164

meanwhile I am not a very expert cmake developer, so it need some time. and you can join in the great work if you like.

crazyhappygame:
@drelaptop. I do not see pch on plan.  But if it would be possible to build with pch android in 2 minutes vs 20 minutes (like it happen for windows build #18662) prebuilds are not so important any more.

Just for clarification pch and prebuilds are two different thinks. Both can be enabled with cmake.
1. precompiled header - build everything as before but with pch compiler flag, please see #18662. clang -emit-pch -include-pch, VS /Yc, /Yu
2. prebuilds - cocos2d-x build separately and linked with game

drelaptop:
@crazyhappygame sorry for confused with both, got you now, I will try to add this feature into cmake. speed up 12 times is unbelievable

another word: existed some cmake modules to do that, such as https://github.com/sakra/cotire, firstly I should add cmake support all platforms, after that consider to reduce build time

crazyhappygame:
I would like try it. Could you merge it?

drelaptop:
that's great, just have a try, and do some test, then I will merge it into this PR.

drelaptop:
Lua template project run failed on mac, others works well. all project need to be improved about mac bundle.

the group of resources and source code have been improved when generate Xcode project by CMake, headers group need to do further later.

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -66,28 +66,22 @@ function(cocos_mark_resources)
 endfunction()
 
 # mark the files in the sub dir of CMAKE_CURRENT_DIR
-function(cocos_mark_source_files cocos_target)
+function(cocos_mark_code_files cocos_target)
```
crazyhappygame: Is this code is for creating groups in xcode and Visual Studio  like i.e.: https://stackoverflow.com/a/33813154/4514853  ?
If yes I think that we should remove it.  Why we needs this groups at all? Both editor can display files using current folders structure. It will make only cmake file complicated and harder to maintain. If we think that folders structure is not as good as it should be we should change folder structure instead of creating filters/groups for xcode/visual studio editor

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -33,14 +57,42 @@ function(cocos_mark_resources)
...
+
+        if(XCODE OR VS)
+            string(REPLACE "/" "\\" ide_source_group "${opt_RESOURCEBASE}/${RES_LOC}")
+            source_group("${ide_source_group}" FILES ${RES_FILE})
+        endif()
     endforeach()
 endfunction()
 
+# mark the files in the sub dir of CMAKE_CURRENT_DIR
+function(cocos_mark_code_files cocos_target)
```
crazyhappygame: Is this code is for creating groups in xcode and Visual Studio like i.e.: https://stackoverflow.com/a/33813154/4514853 ?
If yes I think that we should remove it. Why we needs this groups at all? Both editor can display files using current folders structure. It will make only cmake file complicated and harder to maintain. If we think that folders structure is not as good as it should be we should change folder structure instead of creating filters/groups for xcode/visual studio editor.

With cmake you can use any editor not every editor supports filters/groups

crazyhappygame:
@drelaptop 
> that's great, just have a try, and do some test, then I will merge it into this PR.

My question was: when you will merge it?
This is already big update, there is 59 file changed... I prefer few smaller PR than one giant....
Do we plan to do migration to cmake in one PR?

drelaptop:
understood, I prefer smaller PR too, but this PR is unstable now, and it have much changes on cmake files, so we can't merge it now. 

I plan to merge it when cmake support all platforms works finished. after that we can do improvements,  prebuilt library, even refactor in another PR

you can submit small PR into my branch `drelaptop:v3_android_cmake_support_cpp_lua_js` for now @crazyhappygame

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: yes, this code is for creating groups in Xcode now, VS not test yet.

we can't remove it, for you will see all source files in one folder named "Source Files", not current folders structure when you remove it. that Xcode project is generated by cmake, cmake not group source files using folders structure default. I suggest you do a test using this branch.

We don't need consider all editors, just consider Xcode and VS for now, another reason we only do group when generating Xcode and VS project, it have no bad effect on other IDE

crazyhappygame:
@drelaptop could you estimate when it will be merged (days, weeks, months)?

drelaptop:
This PR would be merged in 2 weeks if every things goes as plan, and you can keep focused on this once a week @crazyhappygame

drelaptop:
Now, Most of cocos2d-x could build and run well on macOS and iOS, and can generate Xcode project.
this is the part manual

#### using CMake in macOS terminal

```sh
cmake -H$COCOS_TEMPLATES_ROOT/.. -Bbuild_cocos2dx
cd build_cocos2dx
make all -j 4
```

#### Generate Xcode Project for macOS platform

```sh
cmake -H$COCOS_TEMPLATES_ROOT/.. -Bbuild_mac -GXcode
cd build_mac
open Cocos2d-x.xcodeproj
```
`cmake --build <dir>                = Build a CMake-generated project binary tree.`
#### Generate Xcode Project for iOS platform

```sh
cd any_dir_you_want
cmake -DCMAKE_TOOLCHAIN_FILE=$COCOS_TEMPLATES_ROOT/../cmake/ios.toolchain.cmake -DIOS_PLATFORM=SIMULATOR64 -H$COCOS_TEMPLATES_ROOT/.. -Bbuild_ios_sim64 -GXcode
cd build_ios_sim64
open Cocos2d-x.xcodeproj
```
`cmake --build <dir>                = Build a CMake-generated project binary tree.`
#### Attention

1. Lua template project run failed on macOS, for it need simulator and the simulator not have any CMake files, I think it's not hurry, so I plan to put it after others works done
2. the `Images.xcassets` and `LaunchScreen.storyboard` didn't work for iOS project generated by CMake, it need to solve later

drelaptop:
Suggest to merge this PR @minggo, for this reasons:

1. the main content of CMake support iOS, macOS compile&package finished
1. CMake support Android, Linux, tested again. works well on Android7.0, Ubuntu16.04
1. **CI checks have passed**, this PR have no bad effect on old build system
1. this PR is big enough, I should finish next works in another PR

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -1,5 +1,29 @@
 include(CMakeParseArguments)
 
+# copy libs to static libs folder
+# error function, have bug
```
minggo, line 4: What does it mean? Is this function used?

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -33,14 +57,104 @@ function(cocos_mark_resources)
...
+
+# if cc_variable not set, then set it cc_value
+macro(cocos_fake_set cc_variable cc_value)
+  if(NOT DEFINED ${cc_variable})
+    set(${cc_variable} ${cc_value})
+  endif()
+endmacro()
+
+# macos package, need review
+macro(cocos_pak_xcode cocos_target)
```
minggo, line 105: There is a info.plist file, why need to set it here?

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: it's a tool function, the purpose is support the unfinished feature: prebuilt, it's not been used for now

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: it's a info.plist file template, it will become the macOS bundle info.plist

drelaptop: this macro set it as default value, if you don't want the default value, just use like:
```cmake
    if(MACOSX)
        cocos_pak_xcode(${APP_NAME} BUNDLE_NAME "Cpp Empty Test")
    elseif(IOS)
        cocos_pak_xcode(${APP_NAME} INFO_PLIST "iOSBundleInfo.plist.in")
    endif()
```

templates/cpp-template-default/proj.android/gradle.properties
```diff
@@ -18,23 +18,28 @@
...
 # Android SDK version that will be used as the earliest version of android this application can run on
 PROP_MIN_SDK_VERSION=14
 
 # Android SDK version that will be used as the latest version of android this application has been tested on
 PROP_TARGET_SDK_VERSION=27
 
-# Android SDK version that NDK will compile to
-PROP_APP_PLATFORM=14
+# Android Build Tools version that will be used as the compile project
+PROP_BUILD_TOOLS_VERSION=27.0.1
```
minggo, line 30: why modify it?

tests/js-tests/project/proj.android/app/build.gradle
```diff
@@ -32,12 +42,17 @@ android {
     }
 
     externalNativeBuild {
-        ndkBuild {
-            if (!project.hasProperty("PROP_NDK_MODE") || PROP_NDK_MODE.compareTo('none') != 0) {
+        if (!project.hasProperty("PROP_NDK_MODE") || PROP_NDK_MODE == 'ndk') {
```
minggo, line 45: `PROP_NDK_MODE == 'ndk'` where set PROP_NDK_MODE to ndk?

templates/cpp-template-default/proj.android/gradle.properties
drelaptop: actually, it is modified by @newnon , i think add `PROP_BUILD_TOOLS_VERSION ` just set a variable to avoid set number `27.0.1` everywhere. As for delete `PROP_APP_PLATFORM `, I don't known the reason

cmake/Modules/CocosBuildHelpers.cmake
minggo: I mean, why not use exist info.plist file? Developers uses to add a plist.info in project.

templates/cpp-template-default/proj.android/gradle.properties
minggo: PROP_APP_PLATFORM is now used in gradle.properties, so can just add a new property, not delete PROP_APP_PLATFORM. They are different properties.

tests/js-tests/project/proj.android/app/build.gradle
drelaptop: in file __gradle.properties__, it set `PROP_NDK_MODE `, control the native compile method by comment it and uncomment it, another way you can set value to `ndk` or `cmake` to control
```
# android cmake support
# uncomment it, native code will build by cmake
# keep comment, native code will build by ndkBuild
# PROP_NDK_MODE=cmake
```

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: understood, use the plist.info in project is a better choice, I will modify it

templates/cpp-template-default/proj.android/gradle.properties
drelaptop: thanks for explain that, I will make `PROP_APP_PLATFORM ` back, and then do another test

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: https://github.com/cocos2d/cocos2d-x/blob/f1dbc310ba6e7743b33e70354739270138bad4be/tests/cpp-empty-test/proj.mac/Info.plist#L8

variables  like `EXECUTABLE_NAME ` depend on the Xcode package config, CMake is hard to support this, the same meaning variable of `EXECUTABLE_NAME` in CMake is `MACOSX_BUNDLE_EXECUTABLE_NAME` if we use info.plist template. 

So if we use plist.info in project resource, I am afraid that we have to change those variables into value to avoid depend Xcode config.

drelaptop: another information is that when set `info.plist` in project into Xcode resource, the variables which cmake didn't support disappeared, meanwhile the open source project like `Urho3D` use info.plist template

drelaptop: understood that, I will modify the variables in info.plist to values, so that cocos-console need modify to adapt this, when using `cocos new`, after all cmake works finished.

drelaptop:
make `PROP_APP_PLATFORM` back finished, please review and `squash and merge` @minggo

newnon:
It is really BAD to set up APP_PLATFORM manually it MUST be same as MIN_SDK_VERISON and android SDK will set it on build. This variable it is one more way to shoot yourself in the leg.

drelaptop:
@newnon Sounds reasonable, but `android SDK will set it on build` automatic? how to prove  that, or is there some official docs said that?

newnon:
sure https://developer.android.com/ndk/guides/application_mk.html

APP_PLATFORM
This variable contains the minimum Android platform version you want to support. For example, a value of android-15 specifies that your library uses APIs that are not available below Android 4.0.3 (API level 15) and can't be used on devices running a lower platform version. For a complete list of platform names and corresponding Android system images, see Android NDK Native APIs.

Instead of changing this flag directly, you should set the minSdkVersion property in the defaultConfig or productFlavors blocks of your module-level build.gradle file. This makes sure your library is used only by apps installed on devices running an adequate version of Android. The ndk-build toolchain uses the following logic to choose the minimum platform version for your library based the ABI you're building and the minSdkVersion you specify:

If there exists a platform version for the ABI equal to minSdkVersion, ndk-build uses that version.
Otherwise, if there exists platform versions lower than minSdkVersion for the ABI, ndk-build uses the highest of those platform versions. This is a reasonable choice because a missing platform version typically means that there were no changes to the native platform APIs since the previous available version.
Otherwise, ndk-build uses the next available platform version higher than minSdkVersion.

drelaptop:
@newnon thanks for mention this

minggo:
@drelaptop  great job

minggo (approved):

Issue #18616, opened 2018-01-09
[BUG] PageView - scroll jumps glitch from the end


Anywhere-Music-Player:
- cocos2d-x version: latest commit from github
- devices test on: Mac, iOS
- developing environments
   - Xcode version: 9.2
Steps to Reproduce:

1. Use this code for UIPageViewTest: 
```
// Create the page view
    Size pageViewSize(209.1, 70.0);
    Size pageSize(26, 26);
    
    PageView* pageView = PageView::create();
    pageView->setDirection(PageView::Direction::HORIZONTAL);
    pageView->setContentSize(pageViewSize);        
    pageView->setBounceEnabled(true);
    Size backgroundSize = background->getContentSize();
    pageView->setPosition((widgetSize - pageView->getContentSize()) / 2.0f);
    pageView->setItemsMargin(35);
    pageView->removeAllItems();
    pageView->setGlobalZOrder(200);

    for (int i = 1; i < 10; ++i)
    {
        Layout* layout = Layout::create();
        layout->setContentSize(pageSize);
        ImageView* imageView = ImageView::create("cocosui/scrollviewbg.png");
        imageView->setScale9Enabled(true);
        imageView->setContentSize(pageSize);
        imageView->setPosition(Vec2(layout->getContentSize().width / 2.0f, layout->getContentSize().height / 2.0f));
        layout->addChild(imageView);
        Text* label = Text::create(StringUtils::format("page %d",(i)), "fonts/Marker Felt.ttf", 15);
        label->setColor(Color3B(192, 192, 192));
        label->setPosition(Vec2(layout->getContentSize().width / 2.0f, layout->getContentSize().height / 2.0f));
        layout->addChild(label);
        pageView->addPage(layout);
    }
    pageView->scrollToItem(4);
```
2. Scroll till the end of the list and it jumps! Somewhere in the middle of the list of items: 

![giphy](https://user-images.githubusercontent.com/3391456/34724754-b4e92042-f557-11e7-8a88-38be203f491e.gif)

Anywhere-Music-Player:
@drelaptop maybe you can help and push it to fix :) ? 
First time I've reported about this here http://discuss.cocos2d-x.org/t/pageview-how-to-use/40474/28?u=kamikaze but no attention to this.

drelaptop:
got this bug, I will test it using the code you give. and look into it, make a feedback for you in days

drelaptop:
[...]

PR #18650, opened 2018-01-19
Fix scroll teleport bug when bouncing in the right side of view.

tiagomartines11:
We have been using this fix since v3.12. 
Although it is not fully tested against the test app, it probably addresses issue #18616.

I'm putting this PR up for further testing.


Anywhere-Music-Player:
@tiagomartines11 any news :) ? This still has problem with slow movement, so not fully fix this.

Issue #18673, opened 2018-02-01
latest v3 -> cocos new compiler errors


slackmoehrle:
- cocos2d-x version: v3.17
- devices test on: ios

Steps to Reproduce:

1. pull latest from github v3 branch 
2. `cocos new ...` open the Xcode project.
3. try and run and you will get compiler errors like

<img width="845" alt="screen shot 2018-02-01 at 6 22 06 pm" src="https://user-images.githubusercontent.com/7491758/35708654-1d9dcc8c-077d-11e8-9ea9-da142f172225.png">

```sh
$ cocos -v
Python 2.7.10
cocos2d-x-3.17
Cocos Console 2.3
```

```sh
$ git remote -v
origin	git@github.com:cocos2d/cocos2d-x.git (fetch)
origin	git@github.com:cocos2d/cocos2d-x.git (push)
$ git branch -v
  0v 18709f38d1 Update config.json
* v3 8c1c3476f6 Fix win32 cpp-template, speedup appveyor, add cocos new CI configuration (#18663)

```

crazyhappygame:
I propose to extend Travis ci and add "cocos new  " build config for iOS/Mac together with fixing this issue.
I added ci test for "cocos new" for android and win32(there was compilation error as well :) ). Please check appveyor ci for win32 and android "cocos new"

crazyhappygame:
I propose to extend Travis ci and add "cocos new  " build config for iOS/Mac together with fixing this issue.
I added ci test for "cocos new" for android and win32(there was compilation error as well :) ). Please check appveyor ci for win32 and android "cocos new"

slackmoehrle:
Here is a PR for this: https://github.com/cocos2d/cocos2d-x/pull/18674

drelaptop:
@slackmoehrle I think this issue should been closed, for the corresponding PR have been merged

slackmoehrle:
@drelaptop agreed.Closing.

PR #18632, opened 2018-01-17
V3 multisampling support

newnon:
support for multisampling on most platforms

PR #18674, opened 2018-02-02
Fix Issues: latest v3 -> cocos new compiler errors #18673

drelaptop:
it's mistake of "V3 multisampling support #18632" @slackmoehrle


drelaptop:
@minggo please merge

slackmoehrle:
Yes, this works for me.

PR #18684, opened 2018-02-09
Get ValueMap by reference

AIGRIND-LLC:


cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -543,7 +543,7 @@ void SpriteFrameCache::removeSpriteFramesFromDictionary(ValueMap& dictionary)
     if (dictionary["frames"].getType() != cocos2d::Value::Type::MAP)
         return;
 
-    ValueMap framesDict = dictionary["frames"].asValueMap();
+    ValueMap& framesDict = dictionary["frames"].asValueMap();
```
minggo: Good catch, may be it is better to change to
```c++
const ValueMap& framesDict = dictionary["frames"].asValueMap();
```
because we don't want to modify the content here.

PR #18686, opened 2018-02-10
Adds support for executing sequential commands separated by a delimiter

gestern:
Support for the commands series, such as ``"command | command | command"``. The previous functionality saved unchanged.

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -74,6 +74,7 @@ NS_CC_BEGIN
 extern const char* cocos2dVersion(void);
 
 #define PROMPT  "> "
+#define DEFAULT_COMMAND_SEPARATOR '|'
 
 static const size_t SEND_BUFSIZ = 512;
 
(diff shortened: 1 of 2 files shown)


cocos/base/CCConsole.cpp
```diff
@@ -937,17 +939,26 @@ bool Console::parseCommand(int fd)
...
+    
+    Console::Utility::sendPrompt(fd);
+    
+    return true;
+}
+
+void Console::performCommand(int fd, const std::string& command) {
+    std::vector<std::string> args = Console::Utility::split(command, ' ');
+    if(args.empty()) {
+        throw std::runtime_error("Unknown command. Type 'help' for options\n");
```
minggo, line 961: Can we use  return value instead of exception?

gestern: Yes, if this conflicts with principles to use of exceptions in the code. Be that as it may, I plan to rewrite all the code in Console-subsystem. As practice has shown, this functionality is the most important for identifying problems in "Cocos2D-X based projects". The description of problems (Steps To Reproduce) can be completely replaced with a list of class commands. Especially if you implement the command "`sleep <ms>`". For example `<do something> | sleep 5ms | <do something>`. Difficult identifiable problems associated with (for example: Actions or Scene Transitions) can now easily be reproduced without fear that the network problems will affect.

minggo: Yep, reasonable.

Issue #18717, opened 2018-03-08
CMake need to improve, record ideas


drelaptop:
- cocos2d-x version: latest in github

- [x] 1. `info.plist` is always created when `-GXcode`, it's bad friendly for custom `info.plist`

    - can't find a method to solve this for now, most cmake use template info.plist

- [x] 2. iOS target and macOS can't be in one project, some complex for switch

    - cmake didn't support, see: https://gitlab.kitware.com/cmake/cmake/issues/17805 

- [x] 3. write cmake docs, how to use step by step, add link in main readme.md

    - finished at https://github.com/cocos2d/cocos2d-x/blob/v3/cmake/README.md

- [x] 4. sub-dir tree of header files isn't include in Xcode or VS project

    - finished at https://github.com/cocos2d/cocos2d-x/pull/18722

- [x] 5. supply simple python script to execute pre-defined shell command, let cmake easy to use

    - plan to do it, in the future, it's better finish this before 3.17 next milestone released.

- [x] 6. delete the old project build files or not if cmake can do build well

    - the cmake build isn't so good as the older way for iOS platform, need to find a solution 
    - might have to wait 4.0 version to do this

- [x] 7. cmake support pre-compiled headers, to increase the compile speed

   - no plan to do it recently, hope cocos2d-x forum help

- [x] 8. What about cmake and cocos2d-x-3rd-party-libs?

    - ~~no plan to do it recently, might do it in future~~
    - plan do it on 3.18, now new a repo at https://github.com/cocos2d/cocos2d-x-external

- [x] 9. Lua Template project on win32, exist different behavior between run with VS and  double-click run on Explorer

refer to: https://stackoverflow.com/questions/23950887/does-cmake-offer-a-method-to-set-the-working-directory-for-a-given-build-system

- [x] 10. using prebuilt libs on Android Gradle Build, we must set `COCOS_PREBUILT_ROOT` to finger the prebuilt libs root library.

    - might we can get system environment from Gradle, to get the `cocos2d-x/prebuilt` path, refer to [proper-way-to-use-system-environment-variables-in-gradle-using-android-studio](https://stackoverflow.com/questions/25589052/proper-way-to-use-system-environment-variables-in-gradle-using-android-studio)
[...]

PR #18719, opened 2018-03-08
add cmake guide docs

drelaptop:
1. improve cmake docs, add using the prebuilt libs docs
2. change the default iOS build target from i386 to 64, make default value alway build success

### DIFF
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -34,8 +34,6 @@ cmake .. -G"Visual Studio 15 2017" -Tv141
 
 Execute `cmake --build .` to compile, or open `Cocos2d-x.sln` in Explorer to use the generated project. 
 
-Execute `cmake --build` to see the build menu.
-
 ### Generate macOS Project
 
 ```sh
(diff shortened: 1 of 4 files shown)


drelaptop:

cmake/README.md
```diff
@@ -1,23 +1,101 @@
...
-* Documentation: [cmake.org/documentation](https://cmake.org/documentation/)
+2. You should use out-of-source build, that is said you need to create a different directory from `cocos2d-x` to execute cmake command.
+
+## Step by Step
+
+### Build in Linux
 
-* CMake FAQ: [Wiki/CMake_FAQ](https://cmake.org/Wiki/CMake_FAQ)
+```sh
+cd cocos2d-x
```
crazyhappygame, line 17: why we need for each platform Linux/Win32/Mac different instruction?
It should be everywhere the same:
$ cd cocos2d-x
$ mkdir build
$ cd build
$ cmake ..
$ cmake --build .

cmake/README.md
```diff
@@ -1,23 +1,101 @@
...
+cocos new -l cpp -p my.pack.app2 test_app2
+mkdir app2_build && cd app2_build
+cmake ../test_app2 -DUSE_COCOS_PREBUILT=ON
+make TemplateCpp
+open bin/TemplateCpp.app
+```
+> Any other cpp project can use prebuilt in this way
 
-2. You should use out-of-source build, that is said you need to create a different directory from `$cocos2dx_root` to execute cmake command.
+Using this feature on Android exists a little difference, for CMake can't find system environment when build in Gradle. So you need to [supply a path](https://github.com/cocos2d/cocos2d-x/blob/c087be314c2c56a757bf66163b173746b5d6ad34/tests/cpp-empty-test/proj.android/app/build.gradle#L34) as the location of prebuilt libs.
```
crazyhappygame, line 98: It seems like there is access to env from gradle System.getenv("KEYSTORE")
https://stackoverflow.com/questions/25589052/proper-way-to-use-system-environment-variables-in-gradle-using-android-studio

cmake/README.md
drelaptop: Right, we can do this in Linux/Win32/Mac native build, not include iOS and Android, different instruction tell users other informations, for example how to generate Xcode project.

Specially, cmake build only support MSVC, didn't support MinGw for win32 platform, so I think adding visible option will avoid conflicts.

>  `cocos2d-x/build` have been used for the older build files, so we can't write `cd cocos2d-x && mkdir build` into the instruction.

cmake/README.md
drelaptop: thanks, I will record this idea into https://github.com/cocos2d/cocos2d-x/issues/18717, might improve it in the future.

PR #18722, opened 2018-03-09
cmake add collect headers list, unify the logic of source add

drelaptop:
1. When cmake compile sources, we always add headers search by `include_directories` to supply the header file search path, so that when generated Xcode or VS project, not include headers files. Of course we can add headers files in `CMakeLists.txt` to achieve this goal. But it's a little complex and need much time.

This PR finish this job, add header list for cocos2d-x libs and app, except external CMakeLists.txt

auto-select-headers

2. Unify the logic of source add for app, like this:
```
list(APPEND GAME_HEADER
    Classes/HelloWorldScene.h
)
list(APPEND GAME_SOURCE
    Classes/HelloWorldScene.cpp
)
```
I think it's more friendly than the old way, easy to use


tools/simulator/libsimulator/CMakeLists.txt
```diff
@@ -114,6 +114,9 @@ include_directories(
     ${COCOS2DX_ROOT_PATH}/cocos/editor-support
 )
 
+get_headers_by_sources(cc_headers CC_SOURCES ${SIMELATOR_SRC})
```
crazyhappygame: s/SIMELATOR/SIMULATOR/g ?

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -169,27 +169,57 @@ function(cocos_mark_code_files cocos_target)
...
-    # set source_group, consider sub source group 
-    string(REPLACE "/" "\\" ide_file_group "${group_base}/${relative_path}")
-    source_group("${ide_file_group}" FILES ${single_file})
+    source_group_single_file(${single_file} GROUP_TO "Source Files" BASE_PATH "${root_dir}")
+  endforeach()
+
+endfunction()
+
+# get headers from `CC_SOURCES` save result in `headers_out`
+function(get_headers_by_sources headers_out)
```
crazyhappygame: I think that we have already too many function... It is too complicated. 
Consider use GLOB_RECURSE for headers
https://cmake.org/cmake/help/latest/command/file.html
It is not recommended for src files but for header files I think it is fine.

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -109,7 +109,7 @@ function(cocos_copy_target_dll cocos_target)
   foreach(single_target_dll ${all_depend_dlls} )
```
crazyhappygame, line 110: I do not like function get_target_depends_ext_dlls.
I think that better would be to have and maintain list of dlls.

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: I considered it before. but when we `GLOB_RECURSE` headers, we will get all platform headers, useless headers will be collected too for a specific build.

yes, too many `function`  is complicated, but `function` is more clear than `macro` for cmake, I will have to write much duplicate code if I don't use `function` or `macro`.

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: I think `get_target_depends_ext_dlls` is the most complicated function in those cmake files of cocos2d-x, it's a little difficult to understand the logic of collect dlls.

maintain a list of dlls is simple and easy to realize, and then how many dlls list should we supply? 3 dlls list, one dlls list for per language? but it should have different dlls when a win32 app use different libs, for as we all know one dll for one lib. 

I catch the logic of one dll for one lib, the app target need a dll if it use the corresponding lib. the current collect dlls logic is:
 
1. pick corresponding dll for a specific lib in function `cocos_find_dlls_for_lib`, finilly save it in `${_prefix}_DLLS`
 1. add a target property `DEPEND_DLLS` record a target depend dlls, collect dlls in function `cocos_use_pkg`
1. `get_target_depends_ext_dlls` depend on the target property `LINK_LIBRARIES` find a target depend libs, then collect dlls through the the target property `DEPEND_DLLS`. you know lib1 link lib2, lib2 may link lib3, lib3 may link lib4...., it's a tree struct, so I have to traverse through recursion.

it's so complicated, but I think it's quite closer the truth of a app use dlls.

tools/simulator/libsimulator/CMakeLists.txt
drelaptop: thanks for finger it out, it's a typo.

cmake/Modules/CocosBuildHelpers.cmake
crazyhappygame: But how many dll list do we have to maintain? Android and Linux do not use dll (only static libs). Do we use dlls for Mac/iOS?

cmake/Modules/CocosBuildHelpers.cmake
crazyhappygame: Than maybe it will be simpler to just have list of headers?
Any way this function may not find averything...

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: dlls only need on windows. should we copy all the dlls for a application? I don't think so. 

For example when we build a cpp application, we don't need copy `lua51.dll`. it's not necessary. then we should have different dll list when different language. so 3 dlls list at least. 

How about for `websockets.dll` if a app don't use the websocket lib? so I think we need copy a dll only we the app use the corresponding lib.

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: You are right, we should write the header list for headers. it's the best choice.

write header list will spend me more than 2 days and it's a little boring... Ok, I will do it in next week when I'm free.

crazyhappygame: It should not be so boring with i.e. "find -name *.h"

cmake/Modules/CocosBuildHelpers.cmake
crazyhappygame: OK. I do not insist.

mariusz102102: As I said I do not insist but if it is only for win32 I think that it is fine to copy everything :)
In future I think that would be nice to setup for 3rdparty libraries CMake and use static libraries everywhere (even for win32)

crazyhappygame: How dll copying works right now (with Visual Studio solution) for win32?

drelaptop: it works before, and I said this commit  https://github.com/cocos2d/cocos2d-x/pull/18722/commits/a31ef36f42c0ae3371d10a13bf4edea908a834ae in this PR fixed https://github.com/cocos2d/cocos2d-x/issues/18723

drelaptop:
CI passed, test finished, please review and merge @minggo

PR #18734, opened 2018-03-15
Simplify CI scripts

crazyhappygame:


crazyhappygame:
@minggo @drelaptop could you merge it?

PR #18741, opened 2018-03-19
Fix issue #18724

halx99:
see issue: https://github.com/cocos2d/cocos2d-x/issues/18724


drelaptop:
So what happened? the reason is ...

halx99:
Just view the diff, and see: http://www.cplusplus.com/reference/cstdio/vsnprintf

drelaptop (approved):

drelaptop:
Please merge @minggo , it fix a issue caused by #18425 ,so we need merge it before 3.17 released.

Issue #18571, opened 2017-12-16
Issue with label right alignment and max width


Dimon4eg:
Label with max width (for word wraping) and right alignment works with issue: some words have additional space from right side.

![image](https://user-images.githubusercontent.com/1271541/34075133-cf5ea46e-e2c5-11e7-948d-a1b073a0de6b.png)

- cocos2d-x version: 3.16 
- devices test on: windows
- developing environments
   - NDK version:
   - Xcode version:
   - VS version: 2017
   - browser type and version:

Steps to Reproduce:

1. run cpp-tests
2. open "30: Node: Label - New API"
3. go to 8 test
4. select right at bottom

drelaptop:
since solved by #18751 , issue should be closed

PR #18751, opened 2018-03-27
Don't count spaces at the end of line as part of the line width

sp-jordi-vilalta:
Currently, the label text formatting counts the space used by the spaces at the end of a line as part of the line width. This makes the center and right alignment not to work properly.

The screenshots below show the Cocos test "29:Node: Label - New API" -> "52:Clamp content Test: Word Wrap" with different alignments, before and after this change.

-|Left|Center|Right
-|-|-|-
Original|original-left|original-center|original-right
Fixed|fixed-left|fixed-center|fixed-right

### DIFF
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -235,7 +235,10 @@ bool Label::multilineTextWrap(const std::function<int(const std::u32string&, int
                     nextLetterX += _horizontalKernings[letterIndex + 1];
                 nextLetterX += letterDef.xAdvance * _bmfontScale + _additionalKerning;
 
-                tokenRight = nextLetterX / contentScaleFactor;
+                if (tokenLen != 1 || !StringUtils::isUnicodeSpace(character))
+                {
+                    tokenRight = nextLetterX / contentScaleFactor;
+                }
             }
             nextChangeSize = true;
 


Dimon4eg:
Cool! Some time ago I have opened issue about it https://github.com/cocos2d/cocos2d-x/issues/18571

drelaptop (approved):

drelaptop:
thanks for this good patch, please merge @minggo

sp-andres-barrera (approved):

tiagomartines11:
@minggo This PR breaks the behavior of RichText labels.
This will now display both words next to each other without the extra space between them:
`<font ...>HELLO </font><font ...>WORLD</font>`
However, this still works:
`<font ...>HELLO</font><font ...> WORLD</font>`
Previously, both combinations would display the expected results.

minggo:
@drelaptop please take a look, and add a new test case for it in cpp tests.

rh101:
This fix isn't quite right, and I'll explain why.

The initial problem posted by sp-jordi-vilalta involves multi-line splitting, and the fact that the end line spaces not being trimmed when the lines are split.  The implemented fix trims ALL line spaces regardless of whether there are line splits or not. 

The way it should work is that no trimming should occur if there are NO line splits, because any spaces placed at the end of the line would be intentional.

PR #18752, opened 2018-03-27
Fix the lost of sound effect issue

xiejiajian:
Fix the issue that the sound effect doesn't work after the application return from background .


xiejiajian:
Why fails?

PR #18753, opened 2018-03-27
Support for non-breaking characters in label

sp-andres-barrera:
## Description:
Add support for some Unicode characters used to prevent line breaks at their position.

A previous PR added support for non-breaking space (https://github.com/cocos2d/cocos2d-x/pull/18147), but that fix removed the character as a whitespace and could mess with other logic relaying on `StringUtils::isUnicodeSpace`, including:
- `Label::multilineTextWrap`
- `trimUTF16Vector` from `ccUTF8` 
- The  test "65: Unit Test Nº5" was broken.

## Additional Info:
- https://en.wikipedia.org/wiki/Space_(punctuation)#Spaces_in_Unicode
- https://en.wikipedia.org/wiki/Non-breaking_space
- https://en.wikipedia.org/wiki/Figure_space
- https://en.wikipedia.org/wiki/Word_joiner

### DIFF
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -90,8 +90,9 @@ int Label::getFirstWordLen(const std::u32string& utf32Text, int startIndex, int 
         char32_t character = utf32Text[index];
 
         if (character == StringUtils::UnicodeCharacters::NewLine
-            || StringUtils::isUnicodeSpace(character)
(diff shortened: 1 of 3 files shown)


sp-jordi-vilalta (approved):

drelaptop (approved):

drelaptop:
thanks, this patch fix a bug about non-breaking characters support, this feature is added in 3.17, so we should merge it before 3.17 released. please @minggo

PR #18774, opened 2018-04-16
fix NDK compile error for js-tests/lua-tests on windows

drelaptop:
add bullet include dir to cocos mk


drelaptop:
@minggo please merge

PR #18777, opened 2018-04-17
a commit about render texture destroy test cases, revert it

drelaptop:
… with global z order (#18629)"

This reverts commit 38864456bf14ac88cb3b69e8715f6d0f505efecd.

Seriously, it causes awful crash issues. when running the __Node:RenderTexture -> 1:Touch the screen__ test case. on all the tests project, include cpp-tests, js-tests, lua-tests.

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -796,12 +796,6 @@ void RenderTexture::draw(Renderer *renderer, const Mat4 &transform, uint32_t fla
     }
 }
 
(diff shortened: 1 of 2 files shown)


drelaptop:

minggo:
Should find the reason why it will cause crash instead of just reverting it. Because that PR fixed an issue.

drelaptop:
Research the core feature of render texture will spend much time, I have spend 2 days on this issue, and don't want to spend more before the 3.17 release, so select to revert it, it's a temp solution for the 3.17 release.

minggo:
K, sounds reasonable.

PR #18782, opened 2018-04-19
CMake: avoid unsafe header path when cross-compiling

fperrad:


drelaptop:
thanks for this patch, could you describe why need this, when compile android project in Linux?

fperrad:
When cross-compiling, for example, from a x86_64 build/host environment to an ARM target,
the cross-compiler must not use the includes from the host environment (for example in `/usr/include`) which are suitable only for native compilation.

drelaptop (approved):

drelaptop:
thanks, this PR let cmake-build more stable. please merge @minggo

PR #18786, opened 2018-04-23
check test cases and fix small bugs before 3.17 release

drelaptop:


tests/lua-tests/src/ParticleTest/ParticleTest.lua
```diff
@@ -230,7 +230,7 @@ end
 
 local function ParticleReorder_onEnterOrExit(tag)
 	local scheduler = cc.Director:getInstance():getScheduler()
-	if tag == "enter" then
+	if tag == "enterTransitionFinish" then
```
drelaptop, line 233: PR https://github.com/cocos2d/cocos2d-x/pull/18272, add `onEnterCallback` function, move the test case init block into `onEnterCallback`, and then call `onEnterCallback` when __enter__ event comes.

`*_onEnterOrExit` functions were registered in `onEnterCallback`, so that the condition `tag == "enter"` always be false.

drelaptop:

PR #18808, opened 2018-05-08
correct iOS project config when using CMake `-GXcode` generate (3.17 round 2 test)

drelaptop:
1. fix cmake generate iOS project, default arch config run fail on iOS11 device

    - cmake cmd: `cmake .. -GXcode -DCMAKE_TOOLCHAIN_FILE=../cmake/ios.toolchain.cmake`
    - http://punchbox.info:3000/issues/26412


drelaptop:
commit https://github.com/cocos2d/cocos2d-x/pull/18808/commits/fe52bbcc972f776e2bf82e7087aca225e893698d: [fix bugs-editbox-win32 crash](https://github.com/cocos2d/cocos2d-x/pull/18808/commits/fe52bbcc972f776e2bf82e7087aca225e893698d) 

Fix the 15th case of Bugs test in win32 - cpp-tests project, editbox crash on win32 after it's removed from parent.

http://punchbox.info:3000/issues/26419

drelaptop:
@minggo please review

PR #18816, opened 2018-05-11
refactor cmake copy res, fix #18813

### LINKED ISSUE #18813: Problem with resource copying on linux
- cocos2d-x version:git
- devices test on: linux mint 18.3
- developing environments
   - NDK version:
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:

1. Build game on linux.
2. Change  some resource files.
3. Build again, and resources will NOT be updated.

What could be a problem?

drelaptop:
Using `configure_file(<input> <output> COPYONLY)` to copy res files, instead of custom command.
fix #18813 

### Refer to:
https://stackoverflow.com/questions/34799916/cmake-copy-file-from-source-directory-to-binary-directory
> Unlike file(COPY ...) it creates a file-level dependency between input and output, that is:
> If the input file is modified the build system will re-run CMake to re-configure the file and generate the build system again.


v1993:
Now it updates changed files (thanks for that!), but removed files isn't removed.

It isn't critical, but still is a problem.

drelaptop:
What means __removed files isn't removed.__ could you describe it in detail. or make a example @v1993

v1993:
Oh my god, my poor logic…
I meant that if I remove file from resources (project) it will not be removed from target directory (build).

drelaptop:
I don't know if CMake existed any policy support this feature you want.

As you said, it isn't critical. we can try to improve this later.

crazyhappygame:
I think that we should not copy resources at all. Runtime should just load resources from proper location.

v1993:
@crazyhappygame It makes harder distributibg game to testers and move to another location/computer. Also, it requires separated commands for deploy/compile and run. It doesn't make any sence for packages (android/iOS/macOS).

So I think that small results requires a lot of work.

crazyhappygame:
@v1993 what do you think about following
Resources - original source location
bin - new folder with libs, binaries

And for deploy you have one command: "copy bin, Resource" folders

Resources are always loaded from relative path ../Resources/.....

crazyhappygame:
@v1993 there is also symlink option:
http://qrikko.blogspot.com/2016/05/cmake-and-how-to-copy-resources-during.html?m=1

v1993:
@crazyhappygame Symlink sounds interesting… So we can use this pull request for deploy and symlinks for different builds. APK still sounds like a problem: we can't pack symlink.

v1993:
@minggo Can you merge this into main branch, please? It is important for my project.

drelaptop:
the PR improve the way of  cmake copy res, please merge into v3 @minggo

PR #18819, opened 2018-05-13
fix reorder warning

mars3142:
It's just a fix for the reorder warnings, so the CLion build (on macOS) will be more warnings free.


drelaptop (approved):
only reorder the init list, suggest merge @minggo

PR #18849, opened 2018-05-24
optimize the breakpoint resume for iOS

toreinkun:
After downloading fail, it will be removed from the task list. Within the destroy method of downloader, it only save the downloading tasks.
So save the resume data after downloading fail, it can breakpoint resume next time.

### DIFF
--- a/cocos/network/CCDownloader-apple.mm
+++ b/cocos/network/CCDownloader-apple.mm
@@ -63,6 +63,7 @@
 -(const cocos2d::network::DownloaderHints&)getHints;
 -(NSURLSessionDataTask *)createDataTask:(std::shared_ptr<const cocos2d::network::DownloadTask>&) task;
 -(NSURLSessionDownloadTask *)createFileTask:(std::shared_ptr<const cocos2d::network::DownloadTask>&) task;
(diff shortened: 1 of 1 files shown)


drelaptop:
could you add a test case into cpp-tests project? @panzhihao

toreinkun:
you can use the DownloadTest's "download big file" to test it. While it’s downloading, you can offline. After  the error message show, you can press the download button again, you will see the breakpoint resume.

cocos/network/CCDownloader-apple.mm
```diff
@@ -444,6 +451,19 @@ - (void)URLSession:(NSURLSession *)session task :(NSURLSessionTask *)task
             }
         }
     }
+    
+    if (error)
+    {
+        NSData *resumeData = [error.userInfo objectForKey:NSURLSessionDownloadTaskResumeData];
+        if(resumeData) {
```
PatriceJiang, line 458: Since the error occurs, the `resumeData` **may be broken** and should not be persisted. So it is better to ignore these data.

PatriceJiang (commented):
When an error occurs, data should be ignored rather than be written to file.

PR #18850, opened 2018-05-24
be safe to handle operator=

toreinkun:
1、CCData will crash if this equals to other, because CCData::_bytes has been free before memcpy. 
2、CCGLProgramState::_value.tex.texture will leak if this equals to other.

### DIFF
--- a/cocos/base/CCData.cpp
+++ b/cocos/base/CCData.cpp
@@ -62,21 +62,21 @@ Data::~Data()
 
 Data& Data::operator= (const Data& other)
 {
-	if (this != &other)
-	{
-		CCLOGINFO("In the copy assignment of Data.");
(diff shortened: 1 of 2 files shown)


cocos/base/CCData.cpp
```diff
@@ -62,15 +62,21 @@ Data::~Data()
 
 Data& Data::operator= (const Data& other)
 {
-    CCLOGINFO("In the copy assignment of Data.");
-    copy(other._bytes, other._size);
+	if (this != &other)
```
drelaptop: TAB should be replaced with 4 SPACE

toreinkun: ok, i replaced it later

PR #18857, opened 2018-05-27
Fix `ZipUtils` include headers order

pyrosphere:
An old hack that changed the default order of included headers in `ZipUtils` causes issues when compiling cocos2d for Android 64-bit, throwing errors of missing `fgetpos` and `fsetpos` functions.


drelaptop (approved):
tested by lua-empty-test, set property `PROP_APP_ABI=armeabi-v7a:arm64-v8a:x86`

PR #18858, opened 2018-05-27
Update UIAbstractCheckButton.cpp

theSlyest:
This factory method to create a checkbox instance was not working (UICheckBox.cpp:79):
CheckBox* CheckBox::create(const std::string& backGround,
                           const std::string& cross,
                           TextureResType texType)
Because the program tries to load missing textures. I just added some exit conditions to avoid it.


drelaptop (approved):

drelaptop:
> This factory method to create a checkbox instance was not working

what's the situation this method not working, could you give some example code to show this?

v1993 (approved):

drelaptop:
stability improvement

PR #18869, opened 2018-05-30
add test case for rich text space, and correct #18751

drelaptop:
related discuss, #18751

### DIFF
--- a/cocos/2d/CCLabelTextFormatter.cpp
+++ b/cocos/2d/CCLabelTextFormatter.cpp
@@ -236,7 +236,8 @@ bool Label::multilineTextWrap(const std::function<int(const std::u32string&, int
                     nextLetterX += _horizontalKernings[letterIndex + 1];
                 nextLetterX += letterDef.xAdvance * _bmfontScale + _additionalKerning;
 
-                if (tokenLen != 1 || !StringUtils::isUnicodeSpace(character))
+                if (tokenLen != 1 || !StringUtils::isUnicodeSpace(character)
+                    || _hAlignment == cocos2d::TextHAlignment::LEFT)
                 {
                     tokenRight = nextLetterX / contentScaleFactor;
                 }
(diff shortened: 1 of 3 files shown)


drelaptop:
@tiagomartines11 could you check this test case to show wrong space behavior? I didn't found anything wrong with the space. 

this is the create function
```
RichText::createWithXML("words should be divided with space.<br /><br /><font color='#ffff00'>HELLO </font><font color='#ffff00'>WORLD</font><br /><br /><font color='#ff00ff'>HELLO</font><font color='#ff00ff'> WORLD</font>");
```

look this

case2

case1

tiagomartines11:
Thank you for this test case!

Please try using:
`RichText::createWithXML("words should be divided with space.<br /><br /><font face='fonts/tahoma.ttf' color='#ffff00'>HELLO </font><font color='#ffff00'>WORLD</font><br /><br /><font color='#ff00ff'>HELLO</font><font color='#ff00ff'> WORLD</font>");`

Apparently the bug is only triggered when we specify font faces 😱

tiagomartines11:
Reverting #18751 makes it functional again, so they are definitely related.

drelaptop:
@tiagomartines11 test case improved and bug fixed, please verify.

I have checked this test case and `29:Node: Label - New API" -> "52:Clamp content Test: Word Wrap"`, both works well.

### Reasons
> Apparently the bug is only triggered when we specify font faces 😱

if not specify font faces, RichElementText will be rendered by SystemFont Label, not call `multilineTextWrap ` 

> why case this bug

after #18751 changes,  right spaces didn't been counted, no matter what horizontal alignment. another side, RichText treats text item `<font face='fonts/tahoma.ttf' color='#ffff00'>HELLO </font>` as a left alignment TTF label, so the spaces be cut too.

### Solved

add a condition to avoid cut spaces when a TTF label was left alignment. This still make __the center and right alignment to work properly__

rh101:
Please don't merge this in, since it's an incomplete work-around that is not catering for other issues introduced by #18751  

Please refer to #18874 for more information.

drelaptop:
@rh101 thanks for this, I will check that

drelaptop:
https://github.com/cocos2d/cocos2d-x/pull/18876 is a better solution, test case should be added at that PR too, so close this.

Issue #18874, opened 2018-06-04
UIRichText should not be trimming spaces in RichElementText


rh101:
- cocos2d-x version: 3.17
- devices test on: N/A
- developing environments 
   - NDK version: N/A
   - Xcode version: N/A
   - VS version: VS2017
   - browser type and version: N/A

Steps to Reproduce:

A change made in 3.17, one of which is #18447, the other being #18751, introduced right side text trimming for a RichElementText even if it's not splitting a line, which is incorrect behavior.  

A UIRichText is made up of several elements, and spaces in RichElementText are inserted on purpose to make sure there is space between a text element and the next element.  If you trim the right side spaces, you're purposely breaking the look of the UIRichText.  It is understandable if you want to trim end spaces when line-splitting, but for single line RichElementText no spaces should be trimmed.

One of the problem is in UIRichText.cpp method handleTextRenderer():
```
            // skip spaces
            StringUtils::StringUTF8::CharUTF8Store& str = utf8Text.getString();
            int rightStart = leftLength;
            while (rightStart < (int)str.size() && str[rightStart].isASCII() && std::isspace(str[rightStart]._char[0], std::locale()))
                ++rightStart; 
```

The other is in CCLabelTextFormatter.cpp, method multilineTextWrap():
```
                if (tokenLen != 1 || !StringUtils::isUnicodeSpace(character))
                {
                    tokenRight = nextLetterX / contentScaleFactor;
                }
```

If there is a valid reason to want to trim the right side spaces in a RichElementText, then give the user an option to either turn it on or off.  This can be easily done via the flag bits, such as RichElementText::TRIM_RIGHT_SPACE added to the enum:

    enum {
        ITALICS_FLAG = 1 << 0,          /*!< italic text */
        BOLD_FLAG = 1 << 1,             /*!< bold text */
        UNDERLINE_FLAG = 1 << 2,        /*!< underline */
        STRIKETHROUGH_FLAG = 1 << 3,    /*!< strikethrough */
        URL_FLAG = 1 << 4,              /*!< url of anchor */
        OUTLINE_FLAG = 1 << 5,          /*!< outline effect */
        SHADOW_FLAG = 1 << 6,           /*!< shadow effect */
        GLOW_FLAG = 1 << 7,              /*!< glow effect */
        TRIM_RIGHT_SPACE  1 << 8     /*!< trim right side spaces */
    };

[...]

Issue #18828, opened 2018-05-17
UIRichText ignoring X and Y offsets on elements


rh101:
- cocos2d-x version: 3.15, 13.16
- devices test on: N/A
- developing environments
   - NDK version: N/A
   - Xcode version: N/A
   - VS version: N/A
   - browser type and version: N/A

Steps to Reproduce:

If a non-zero centered sprite is to be inserted as an element in a RichText parent, then it needs to be shifted to be aligned correctly with the text.  This can only happen if position X and Y of the sprite are used when rendering, but at the moment those fields of the sprite node are ignored.

Example:
```
auto richText= ui::RichText::create();
richText->ignoreContentAdaptWithSize(true);

auto reText = ui::RichElementText::create(1, color, 255, "Testing image ", "fonts/Arimo-Regular.ttf", 30, ui::RichElementText::SHADOW_FLAG);
richText->pushBackElement(reText);

auto customSprite = Sprite::createWithSpriteFrameName("sample.png");
customSprite->setPosition(-40, 50);  // I want to offset the sprite in the RichText by x = -40, and y = 50.
auto reCustom = ui::RichElementCustomNode::create(2, color, 255, customSprite);
richText->pushBackElement(reCustom);
```

You will notice that the sprite in the text is not offset by -40, 50.

The reason is this section of code in UIRichText.cpp:

```
void RichText::formarRenderers()
{
    if (_ignoreSize)
    {
        float newContentSizeWidth = 0.0f;
        float nextPosY = 0.0f;
        std::vector<std::pair<Vector<Node*>*, float> > rowWidthPairs;
        rowWidthPairs.reserve(_elementRenders.size());
        for (auto& element: _elementRenders)
        {
            float nextPosX = 0.0f;
            float maxY = 0.0f;
            for (auto& iter : element)
            {
                iter->setAnchorPoint(Vec2::ZERO);
                iter->setPosition(nextPosX, nextPosY);  // <=== This is the cause
                this->addProtectedChild(iter, 1);
                Size iSize = iter->getContentSize();
                newContentSizeWidth += iSize.width;
                nextPosX += iSize.width;
                maxY = MAX(maxY, iSize.height);
            }
            nextPosY -= maxY;
            rowWidthPairs.emplace_back(&element, nextPosX);
        }
        this->setContentSize(Size(newContentSizeWidth, -nextPosY));
        for ( auto& row : rowWidthPairs )
            doHorizontalAlignment(*row.first, row.second);
[...]

PR #18875, opened 2018-06-04
Fix for trailing whitespace trimming in Labels and UIRichText

rh101:
[UIRichText.cpp] Allow user to select if they want to trim trailing spaces for a RichElementText. Also fixed crash if estimatedIdx is less than 0 in certain conditions.
[UIRichText.h] New flag added for enabling trailing space trimming.
[CCLabelTextFormatter.cpp] Only trim trailing whitespace if lines are being split. This will trim all whitespace up to the next non-whitespace token in a line.

Please refer to #18874 #18751 

Also fix for #18828 which is to use X and Y positions on custom nodes being drawn in the UIRichText.


rh101:
Reverted the change made to X and Y elements since it has some side effects.

PR #18876, opened 2018-06-04
Fix for Label and UIRichEdit text trailing space trimming

rh101:
[UIRichText.cpp]  Allow user to select if they want to trim trailing spaces for a  RichElementText. Also fixed crash if estimatedIdx is less than 0 in  certain conditions. [UIRichText.h] New flag added for enabling trailing space trimming. [CCLabelTextFormatter.cpp] Only trim trailing whitespace if lines are  being split. This will trim all whitespace up to the next non-whitespace  token in a line. 

Please refer to #18874

### DIFF
--- a/cocos/ui/UIRichText.cpp
+++ b/cocos/ui/UIRichText.cpp
@@ -1716,12 +1716,6 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
             else
                 estimatedIdx = static_cast<int>(_leftSpaceWidth / fontSize);
 
-            // estimatedIdx should never be less than 0
-            if (estimatedIdx < 0)
-            {
-                estimatedIdx = 0;
-            }
-            
             int leftLength = 0;
             if (wrapMode == WRAP_PER_WORD)
                 leftLength = findSplitPositionForWord(textRenderer, utf8Text, estimatedIdx, _leftSpaceWidth, _customSize.width);
@@ -1735,15 +1729,11 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
                 pushToContainer(textRenderer);
             }
 
+            // skip spaces
             StringUtils::StringUTF8::CharUTF8Store& str = utf8Text.getString();
             int rightStart = leftLength;
-
-            if (flags & RichElementText::TRIM_TRAILING_WHITESPACE)
-            {
-                // skip spaces
-                while (rightStart < (int)str.size() && str[rightStart].isASCII() && std::isspace(str[rightStart]._char[0], std::locale()))
-                    ++rightStart;
-            }
+            while (rightStart < (int)str.size() && str[rightStart].isASCII() && std::isspace(str[rightStart]._char[0], std::locale()))
+                ++rightStart;
 
             // erase the chars which are processed
             str.erase(str.begin(), str.begin() + rightStart);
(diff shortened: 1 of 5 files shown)


cocos/ui/UIRichText.h
```diff
@@ -116,7 +116,8 @@ class CC_GUI_DLL RichElementText : public RichElement
         URL_FLAG = 1 << 4,              /*!< url of anchor */
         OUTLINE_FLAG = 1 << 5,          /*!< outline effect */
         SHADOW_FLAG = 1 << 6,           /*!< shadow effect */
-        GLOW_FLAG = 1 << 7              /*!< glow effect */
+        GLOW_FLAG = 1 << 7,             /*!< glow effect */
+        TRIM_TRAILING_WHITESPACE = 1 << 8       /*!< trim right side spaces */
```
drelaptop: As the descriptions  said
> Allow user to select if they want to trim trailing spaces for a RichElementText

how to select? to set the value of `TRIM_TRAILING_WHITESPACE `, so I think you should add corresponding methods, like those do for `GLOW_FLAG `:
```
static const std::string KEY_ANCHOR_TEXT_GLOW_COLOR;            /*!< key of glow color of anchor tag */
defaults[RichText::KEY_ANCHOR_TEXT_GLOW_COLOR] = "#AFEEEE";
RichText::setAnchorTextGlow()
RichText::isAnchorTextGlowEnabled()
```

drelaptop:
Could you add a test case into `UIRichText.h/cpp` for this feature, that's simple and meaningful. 
you can refer to older test case or #18869 @rh101

cocos/ui/UIRichText.h
rh101: Yep, I'll add what you requested.  I've never actually used the string keys, since I create the UIRichText in C++ code only, so for example:

    auto re1 = ui::RichElementText::create(1, color, 255, str, "fonts/Arimo-Regular.ttf", fontSize, ui::RichElementText::SHADOW_FLAG | ui::RichElementText::TRIM_TRAILING_SPACE);

Anyhow, I'll update the pull request with the changes you mentioned.

rh101:
@drelaptop Will do, and I'll use the same or similar test cases as what is in #18869

drelaptop:
thanks, this is a good patch.

rh101:
@drelaptop I'm going to revert most of the additional changes made to UIRichText.cpp/.h, since they are no longer required.  The code I added to CCLabelTextFormatter.cpp actually fixed all the issues, including those in #18751 #18869 

I'll merge the test cases from #18869 into the this pull request

cocos/ui/UIRichText.cpp
```diff
@@ -1729,14 +1729,10 @@ void RichText::handleTextRenderer(const std::string& text, const std::string& fo
...
 
-            // skip spaces
             StringUtils::StringUTF8::CharUTF8Store& str = utf8Text.getString();
-            int rightStart = leftLength;
-            while (rightStart < (int)str.size() && str[rightStart].isASCII() && std::isspace(str[rightStart]._char[0], std::locale()))
-                ++rightStart;
 
             // erase the chars which are processed
-            str.erase(str.begin(), str.begin() + rightStart);
+            str.erase(str.begin(), str.begin() + leftLength);
```
drelaptop, line 1735: I think it's the best solution to solve RichText trailing space trimming issue

drelaptop (approved):

drelaptop:
tested with Label and RichText cases, works. this is a really beautiful patch, thanks @rh101

rh101:
You're welcome, and thank you for merging it.  I'll close the outstanding issue #18874 since this has been merged.

PR #18895, opened 2018-06-13
Fix Rewrite MathUtilNeon64::crossVec3 to work around compiler bug

sp-stefano-lanza:
Our release builds, compiled with LLVM 9.x and Neon64 instructions, behave incorrectly under certain circumstances. After inspecting the assembly code and its differences with the debug build, we discovered a bug in the code generation of  Mat4::createLookAt related to MathUtilNeon64::crossVec3. Currently, this function uses assembly code which assigns the register %1 to v1, %2 to (v1 + 1) and performs two loads to store and shuffle v1.xyz in a Neon register. When inlined in the assembly of  Mat4::createLookAt, we noticed that the compiler incorrectly computes the address (v1 + 1), by using an OR instead of ADD. This leads to a wrong result.
We worked around this issue by partially rewriting MathUtilNeon64::crossVec3. Now it uses a single load of v1 instead of two, and mov instructions to shuffle the register elements. Also, we added comments to the assembly code and, as minor optimization, used the ext instruction to perform vector rotations instead of repeated mov.
In future, we would recommend using compiler instrinsics for all SIMD code or, alternatively, replacing the math library with an industry standard one.

For reference, the faulty assembly code in Mat4::createLookAt is
```
  orr	        x12, x11, #0x4   // WRONG! should be add 
  add      x13, sp,   #32
  orr	        x9, x13,  #0x4    // WRONG! should be add. Also, unused in the assembly code
  add      x10, sp,   #16
  ld1.2s  { v0 }, [x12]
  ld1.s    { v0 }[2], [x11]
```

### DIFF
--- a/cocos/math/MathUtilNeon64.inl
+++ b/cocos/math/MathUtilNeon64.inl
@@ -24,23 +24,23 @@ class MathUtilNeon64
 {
 public:
     inline static void addMatrix(const float* m, float scalar, float* dst);
-    
+
     inline static void addMatrix(const float* m1, const float* m2, float* dst);
-    
+
     inline static void subtractMatrix(const float* m1, const float* m2, float* dst);
-    
+
     inline static void multiplyMatrix(const float* m, float scalar, float* dst);
-    
+
     inline static void multiplyMatrix(const float* m1, const float* m2, float* dst);
-    
+
     inline static void negateMatrix(const float* m, float* dst);
-    
+
     inline static void transposeMatrix(const float* m, float* dst);
-    
+
     inline static void transformVec4(const float* m, float x, float y, float z, float w, float* dst);
-    
+
     inline static void transformVec4(const float* m, const float* v, float* dst);
-    
+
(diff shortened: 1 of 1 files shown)


crazyhappygame:
Why do we have assembly at all? Should not compiler generate proper and efficient code with neon flag?

sp-stefano-lanza:
Hand tuned SIMD code is likely to be more efficient than the compiler
generated one, which motivates the existence of the many third-party math
libraries. With regards to the cocos2d one, I don't know what the history
behind it is, I simply fixed a bug related to it which was a blocker in our
project. That said, if I were to write a new version of cocos2d from
scratch, I'd replace the whole math code as its quiet outdated and
inefficient. The DirectX math is a good example of a modern library, with
its use of SIMD instrincs and designed for maximum efficiency.

v1993 (changes requested):
You've changed formatting (removed spaces in empty strings).

Issue #17806, opened 2017-05-11
Not stable delta time in default update() method, CADisplayLink duration or timestamp not used?


Anywhere-Music-Player:
- Cocos2d-x 3.15 and other
 - Devices test on: iPhone 6 and any other

Steps to Reproduce:

1. Create a simple HelloWorld project and add only `scheduleUpdate(); `
2. Check `delta` from 
```
void HelloWorld::update(float delta) {
    CCLOG("%f",delta);
 }
```

Notice output like this: 

```
0.018591
0.015011
0.017311
0.015939
0.017793
0.015640
0.016709
0.016957
0.017268
0.015884
0.016776
0.016478
0.016699
0.018520
```

For compare, same setup but using **cocos2d-iphone v3**, so Objective C, gives me rock and solid output:

```
0.016720
0.016674
0.016676
0.016675
0.016675
0.016675
0.016707
0.016687
0.016672
0.016677
0.016632
0.016644
```

As result, using cocos2d-x for my game(I ported it from cocos2d-iphone) gives me a different result in terms of image quality. Main character moves quickly and as result image is not so.. clear, nice and smooth as was before in cocos2d-iphone. I'm not talking about jitter or other very noticeable things, but this problem I can clearly identify. And it's really caused by that not stable numbers comes from `update()`.

Anywhere-Music-Player:
@minggo I've found some info: http://www.learn-cocos2d.com/2013/10/game-engine-multiply-delta-time-or-not/ 

> This is a particular issue with cocos2d because it does not use the CADisplayLink duration property. Instead it relies on system timers to calculate the time difference. As a result, cocos2d’s delta time is merely an approximation and ever-so-slightly disconnected from the actual refresh rate. Other game engines may or may not have the same issue.

So, as I see, this was fixed since that time for cocos2d-iphone v3, but currently a problem for cocos2d-x.

Anywhere-Music-Player:
@minggo Here is how it works for cocos2d-iphone 
https://github.com/cocos2d/cocos2d-objc/blob/v3.5.0/cocos2d/Platforms/iOS/CCDirectorIOS.m#L379

I believe this should be same for cocos2d-x for iOS. More stable delta - more smooth and more quality from game. Thats what I'm seeing when playing my game compiled under cocos2d-iphone v3..

Anywhere-Music-Player:
[...]

PR #18898, opened 2018-06-15
Correct last update time

AIGRIND-LLC:
If next delta time was zeroed (via setNextDeltaTimeZero) the last update time was not corrected.

### DIFF
--- a/cocos/base/CCDirector.cpp
+++ b/cocos/base/CCDirector.cpp
@@ -376,6 +376,7 @@ void Director::calculateDeltaTime()
     {
         _deltaTime = 0;
         _nextDeltaTimeZero = false;
+        _lastUpdate = std::chrono::steady_clock::now();
     }
     else
     {


AIGRIND-LLC:
@minggo please take a look.
It seems that you made this bug in [this commit](https://github.com/cocos2d/cocos2d-x/pull/17852/files) by removing [these lines](https://github.com/cocos2d/cocos2d-x/pull/17852/files#diff-65b4e134e8b4b0fe0169525eb7f2e9c4L360).

Moreover there is probably one another bug added in that commit.
If we call `mainLopp(dt)` then `_deltaTimePassedByCaller` variable becomes `true`.
But it won't be reset to `false` any more. So we can't use `mainLopp(dt)` to pass some custom **dt** only once. If I'm right then I can fix it also in this PR.

AIGRIND-LLC:
@KAMIKAZEUA, let me mention you because all this stuff was added by [your request](https://github.com/cocos2d/cocos2d-x/issues/17806) and probably you have some details about it.

Anywhere-Music-Player:
@AIGRIND-LLC well, it was fixed, so now cocos2d-x uses `VSYNC` from iOS displaylink.  And so `update` gives more stable times. I can't add anything more..

stevetranby:
Does this change cause any issues? Perhaps it could be merged just as a paranoia prevention. Unless it causes an issue, of which I don't currently foresee.

AIGRIND-LLC:
@stevetranby I think this bug may cause some issues and it actually has already caused.

What does `setNextDeltaTimeZero` do? It prevents the app from the big values of delta time in some cases. For example when the app becomes resumed from background. Some libraries can't handle big `dt` properly. You can just search cocos2dx code where `setNextDeltaTimeZero` is used. For example [this commit](https://github.com/cocos2d/cocos2d-x/pull/5915/commits/004374bffb3d7c6f1e45566305a09007c1f2bdb3) relies on it. 
In our project we got `dt` more than 2 seconds and the library we use for FX-rendering crashed also.

This is the expected sequence of `dt` values if `setNextDeltaTimeZero` would work correct :
`0.1 | 0.1 | app::pause for 2 sec ... app::resume | 0.0 | 0.1 | 0.1`

But after the bug introduced in [this commit](https://github.com/cocos2d/cocos2d-x/pull/17852/files) the sequence is:
`0.1 | 0.1 | app::pause for 2 sec ... app::resume | 0.0 | 2.0 | 0.1`
As you can see the app will get big `dt` in any case. And this is issue.

v1993 (approved):

stevetranby:
@AIGRIND-LLC yeah, sorry I wasn't clear enough. 

I was trying to say that I agree with merging this PR as it likely fixes the rare occurrence you describe. But then KAMIKAZEUA's note seemed to suggest VSYNC makes it unnecessary, but I was trying to say might as well still merge it even if unnecessary.

minggo:
@AIGRIND-LLC Yep, you are right, and i don't know that the game codes will also invoke `mainLoop()`.

minggo (approved):

minggo:
> Moreover there is probably one another bug added in that commit.
If we call mainLopp(dt) then _deltaTimePassedByCaller variable becomes true.
But it won't be reset to false any more. So we can't use mainLopp(dt) to pass some custom dt only once. If I'm right then I can fix it also in this PR.

Yes, please.

PR #18926, opened 2018-07-09
Fix OSX linker error with missing GameController.framework

boondaburrah:
Currently project builds on OSX in the latest version available from the website fail because cmake doesn't include gamecontroller.framework. Building cocos2d completes fine, but when you try to link your game code to libcocos2d.a, it complains about missing GCGameController* symbols.

### DIFF
--- a/cmake/Modules/CocosUseLibs.cmake
+++ b/cmake/Modules/CocosUseLibs.cmake
@@ -166,8 +166,6 @@ macro(jscocos2d_depend_libs)
 
     list(APPEND PREBUILT_SPECIFIC_LIBS SPIDERMONKEY)
     if(APPLE)
-        find_library(GAME_CONTROLLER GameController)
-        list(APPEND PLATFORM_SPECIFIC_LIBS ${GAME_CONTROLLER})
         list(APPEND PREBUILT_SPECIFIC_LIBS SQLITE3)
     endif()
 endmacro()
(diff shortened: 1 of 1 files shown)


drelaptop (approved):

drelaptop:
When I test the js-tests project before 3.17 release, I found it needed `GameController.framework`, so I added this

https://github.com/cocos2d/cocos2d-x/blob/d1597a0e3e08d295fbf2a8a407660b089eb7f08d/cmake/Modules/CocosUseLibs.cmake#L168

for I think it's specific needed for js project. this is needed for cpp project too when you test? @boondaburrah

boondaburrah:
Yes my project is C++ only and it still needs GameController.framework. I think maybe if you call Controller::startDiscoveryController(); it needs GCControllerDidConnectNotification among other things to feed the cocos2d EventListenerController.

It's possible if you never call these functions from Cocos2D that the linker doesn't complain because code referencing GameController.framework is stripped out? The default C++ template builds fine without it, but my project (which uses game controllers) won't build unless I link the framework in.

Also I noticed the (hand built?) XCode project for tests includes the framework. (The red copy is the iOS version, which my system is missing since I don't have the iOS SDK installed.)
screen shot 2018-07-10 at 7 46 43

drelaptop:
@boondaburrah thanks for explain, it's reasonable. 

could you remove the `GameController.framework` link at [js specific location](https://github.com/cocos2d/cocos2d-x/blob/d1597a0e3e08d295fbf2a8a407660b089eb7f08d/cmake/Modules/CocosUseLibs.cmake#L168)? since this library is added at the common location, it will be no need to add this at there.

boondaburrah:
Yeah no problem. I didn't notice it was there since I'm not building a JS project, but I ran js tests afterwards and it still builds with it removed from jscocos2d_depend_libs so it should be fine.

drelaptop:
thanks for this patch

boondaburrah:
Thanks!

PR #18930, opened 2018-07-11
update console sub-module for license changes, and check appveyor CI

drelaptop:
appveyor script
```
    & git submodule update --recursive --depth=1 // only checkout latest one commit
```
so appveyor CI failed, when the submodule existed newer commit.


drelaptop:
once you update the sub-module into latest commit, the appveyor CI goes well with `--depth=1` parameter. 

As for the `NativeCommandError` appear in the log, I think it's a mistake for powershell, refer to: https://stackoverflow.com/questions/1394084/ignoring-an-errorlevel-0-in-windows-powershell

PR #18931, opened 2018-07-11
dispatch websocket callbacks to ui thread 

PatriceJiang:
improve thread safety


PatriceJiang:
reform in https://github.com/cocos2d/cocos2d-x/pull/19008

PR #18934, opened 2018-07-12
Added initialisation checks in pcm audio service

umair6:


vkreal:
Sorry, I might be missing but how is `_playItf` not initialized when calling `pause` or `resume`.

drelaptop (approved):

drelaptop:
this patch make the audio more stable. crash won't occur when `pause` and `resume` called, even with init failed.

meanwhile we don't know why `PcmAudioService::init` may failed in some special android device. because we can't reproduce it on our own android device, all the reported crash info come from third party crash collect.

drelaptop:
thanks for this patch @umair6

cocos/audio/android/PcmAudioService.h
```diff
@@ -71,6 +71,7 @@ class PcmAudioService
     int _numChannels;
     int _sampleRate;
     int _bufferSizeInBytes;
+    bool _isInitialised;
```
dumganhar, line 74: No need to add a new member variable for this purpose. It wastes 4 bytes because struct alignment.
I think checking `_playItf` could still resolve this issue.
```c++
void PcmAudioService::pause()
{
    if (_playItf != nullptr && (*_playItf) != nullptr)
    {
        SLresult r = (*_playItf)->SetPlayState(_playItf, SL_PLAYSTATE_PAUSED);
        SL_RETURN_IF_FAILED(r, "PcmAudioService::pause failed");
    }
}

void PcmAudioService::resume()
{
    if (_playItf != nullptr && (*_playItf) != nullptr)
    {
        SLresult r = (*_playItf)->SetPlayState(_playItf, SL_PLAYSTATE_PLAYING);
        SL_RETURN_IF_FAILED(r, "PcmAudioService::resume failed");
    }
}
```

PR #18935, opened 2018-07-12
https downloader timeout on win/mac

PatriceJiang:
timeout error occurs when trying to download HTTPS content on windows.
the original polling `sleep`(delay) time is too long and should be separated from `select` timeout.

### DIFF
--- a/cocos/network/CCDownloader-curl.cpp
+++ b/cocos/network/CCDownloader-curl.cpp
@@ -39,6 +39,8 @@
 // member function with suffix "Proc" designed called in DownloaderCURL::_threadProc
 // member function without suffix designed called in main thread
 
+#define CC_DOWNLOADER_WAIT_DNS 50 //wait until DNS query done
+
 namespace cocos2d { namespace network {
     using namespace std;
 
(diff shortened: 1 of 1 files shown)


v1993:
It is hard to refer to 404 page :(.

cocos/network/CCDownloader-curl.cpp
```diff
@@ -518,7 +521,7 @@ namespace cocos2d { namespace network {
                         timeout.tv_sec = timeoutMS / 1000;
                         timeout.tv_usec = (timeoutMS % 1000) * 1000;
 
-                        rc = select(maxfd+1, &fdread, &fdwrite, &fdexcep, &timeout);
+                        rc = select(maxfd+1, &fdread, &fdwrite, &fdexcep, 50);
```
drelaptop: what's the meaning of number __50__ ？

drelaptop:
I think it's better to add some description to explain what happened and what solved. @PatriceJiang

cocos/network/CCDownloader-curl.cpp
PatriceJiang: my mistake  :(

PatriceJiang:
timeout error occurs when trying to download the contents of HTTPS. 
the original polling `sleep`(delay) time is too long and should be independent from `select` timeout.

PatriceJiang:
it is a similar PR as https://github.com/cocos-creator/cocos2d-x-lite/pull/1382 . @minggo

PR #18936, opened 2018-07-12
socket io detect error

PatriceJiang:
bug: nodejs socket.io server return `Bad Request` during handshake. 
fix socket.io version detect method


PatriceJiang:
bug: nodejs socket.io server return `Bad Request` during handshake.

minggo:
This PR is included in https://github.com/cocos2d/cocos2d-x/pull/18944?

Issue #17304, opened 2017-02-07
socket.io emit with multiple arguments


milani:
In socket.io protocol, It is possible to emit an event with multiple arguments i.e. `client.emit("event","arg1","arg2")` but it is not possible in cocos2d implementation.

Also the underlying structure of `SocketIO.cpp` allows this (with multiple call to `packet->addData`), the `emit` signature is limited to a single string argument. 

- cocos2d-x version: 3.14.1

PR #18944, opened 2018-07-17
SocketIO#emit/send support multiple parameters

PatriceJiang:
refer https://github.com/cocos2d/cocos2d-x/issues/17304, support `c++` & `js`

### DIFF
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -1129,7 +1129,7 @@ void SIOClient::socketClosed()
     this->release();
 }
 
-bool SIOClient::isConnected() 
+bool SIOClient::isConnected() const
 {
     return _socket && _socket->_connected && _connected;
 }
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -1235,7 +1235,7 @@ SIOClient* SocketIO::connect(const std::string& uri, SocketIO::SIODelegate& dele
             return newC;
         }
     }
-
+ 
     return c;
 }
 
--- a/cocos/network/SocketIO.cpp
+++ b/cocos/network/SocketIO.cpp
@@ -384,6 +384,7 @@ public:
     void disconnectFromEndpoint(const std::string& endpoint);
 
     void send(const std::string& endpoint, const std::string& s);
+    void send(const std::string& endpoint, const std::list<std::string>& s);
     void send(SocketIOPacket *packet);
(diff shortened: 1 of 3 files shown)


cocos/network/SocketIO.h
```diff
@@ -223,6 +224,9 @@ class CC_DLL SIOClient
 
     friend class SIOClientImpl;
 
+	bool isConnected();
```
minggo: bool isConnected() const ?

minggo: wrong indent

cocos/network/SocketIO.h
```diff
@@ -255,12 +258,39 @@ class CC_DLL SIOClient
...
      * @param eventname
      * @param args
      */
     void emit(const std::string& eventname, const std::string& args);
+    void emit(const std::string& eventname, const std::list<std::string> &args);
+
+    template<typename ...Arg> 
+    void emit(const std::string& eventname, const std::string& arg1, Arg ...args)
+    {
+        std::list<std::string> arg_list;
```
minggo: `arg_list ` -> argList, we don't use `_` to connect words.

cocos/network/SocketIO.cpp
```diff
@@ -732,6 +743,20 @@ void SIOClientImpl::emit(const std::string& endpoint, const std::string& eventna
...
+    packet->setEndpoint(endpoint == "/" ? "" : endpoint);
+    packet->setEvent(eventname);
+    for (auto &arg : args) {
+        packet->addData(arg);
+    }
+    this->send(packet);
+}
+
+
+
```
minggo: too much spaces

cocos/network/SocketIO.h
```diff
@@ -255,12 +258,32 @@ class CC_DLL SIOClient
      * @param s message.
      */
     void send(const std::string& s);
+    void send(const std::list<std::string>& s);
```
minggo: `void send(const std::list<std::string>& s)` -> `void send(const std::list<std::vector>& s)`

minggo: I think vector is better here since it is more efficient for loop

cocos/network/SocketIO.h
```diff
@@ -255,12 +258,32 @@ class CC_DLL SIOClient
...
+
+    template<typename ...Args> 
+    void send(const std::string& s1, Args ...args)
+    {
+        std::list<std::string> list{s1};
+        send(list, args...);
+    }
+
```
minggo: i think there is not need to use template here

PR #18780, opened 2018-04-19
adjust android / linux build config for cocos CLI

drelaptop:

PR #18953, opened 2018-07-21
change PROP_BUILD_TYPE from ndk-build to CMake

crazyhappygame:


crazyhappygame:

drelaptop:
we have discussed `ndk-build` and ` CMake` in detail at https://github.com/cocos2d/cocos2d-x/pull/18780 before 3.17 released. 

It's good to switch the default value now. If we find some thing bad when using CMake , we should fix them, not to avoid.

In the feature, we hope to maintain only one build system - CMake, do that keep stable and step by  step.

PR #18956, opened 2018-07-22
switch one win32 build to cmake on appveyor CI

crazyhappygame:


crazyhappygame:

drelaptop:
as far as I am confirmed, I can't certain this is proper to replace `.sln` with `cmake`, if some developer use the Git latest code, and want to open `sln` directly.

it may case errors, for we didn't check the `sln` build files. we might need to ask for advice in forum. @crazyhappygame

crazyhappygame:
OK. I see your point we have to support sln (at least for some time). I propose following: we have two win32 builds I will switch one to cmake and second one will use sln.

drelaptop:
thanks, since cpp-tests on android have been tested in Travis, I suggest to comment target __android_cpp_tests__ in Appveyor, to add another target __win32_cpp_tests_cmake__ to test the cmake build process on win32. @crazyhappygame

crazyhappygame:
Many people (including me :) ) do android development on windows. I would like to see that Android building works on windows. There were problem in past that Android does not build on Windows ...

I propose to switch one of current win32 builds to cmake or create new win32_cmake configuration.

drelaptop:
got it, please replace one of the win32 build. @crazyhappygame

crazyhappygame:
OK. Thx

crazyhappygame:
Just for your information in next PR I will create new travis configuration: mac_cmake, ios_cmake :)

PR #18958, opened 2018-07-23
Refactor audio engine module, replace backend with OpenAL for android platform.

halx99:


drelaptop:
many checks failed, even Mac and iOS, strictly it should only have effect on android build.

halx99:
let's check.

crazyhappygame:
@halx99 what exactly is changed here? It seems like you touch all platforms. Is it possible to split to few smaller pull requests?

halx99:
You are right, I will split to few smaller pull requests.

crazyhappygame:
And what will be in those few smaller pull requests? :)

halx99:
Let's do android only.

crazyhappygame:
Why Android? What is wrong, what you would like to fix it? 
We have problem with Linux fmod - is not open source. Could you start with Linux?

halx99:
@drelaptop see the new PR: https://github.com/cocos2d/cocos2d-x/pull/18961,  BEFORE the 3rd party libs update, ci check can't sucess.
@crazyhappygame The original implentation of android does not support streaming mode, and have some issues, such as deadlock when play/stop music frequently. After finish android, may continue work for Linux.

PR #18961, opened 2018-07-24
Refactor audio engine module, replace backend with OpenAL for android

halx99:
… platform.


crazyhappygame:
This is also big pull request. It seems like you move lots of files around...
Is it possible to split this PR to smaller? 
For example in one move files in second use openal.

Please check Travis build cocos_new fails because new files were added and Json file was not updated.

halx99:
Well, tody, I have test OpenAL + mpg123 + ogg on ubuntu 18.04, use the same source with win32, it's works fine.
Acording to win32, android, linux source too much similar, so I think, merge sources is a better choice to refactor audio module.

crazyhappygame:
I think that there is one problem with mpg123: "licensed under LGPL 2.1"

halx99:
mpg123 just a decoder, and ogg is free.OpenAL is MIT.

crazyhappygame:
@halx99 Please check: https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md

> The license must be compatible for the different App Stores, so GPL and LGPL code cannot be used in cocos2d-x.

crazyhappygame:
@drelaptop @minggo Please check carefully this PR and https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/317/
What I see below two libraries have incorrect license:
https://github.com/kcat/openal-soft - LGPL
http://www.mpg123.de/ - LGPL

crazyhappygame:
One more thing: for Android game size critical!
Currently we have one libcocos2dcpp.so for everything. For empty project (cocos2d-x\tests\cpp-empty-test) libcocos2dcpp.so is ~6MB. It is possible to have less than 3MB by turning off unneeded features. 

This PR:
1. add ~3MB more!
mpg123.so = 832 KB
openal.so = 2.1 MB
2. Add 2 LGPL libraries = problems ...
Currently Android has no LGPL deps. Please do not change it,
3. add two more .so files
We will have 3 *.so files instead 1 *.so . Current compilers are great with optimization and in-lining one binary. With 3 separate *.so files you can do nothing .....

crazyhappygame:
On how many Android devices that was tested? What is performance?

halx99:
Well, due to the license limition, this PR may need close currently before we can find a non LGPL solution for both win32,android,linux platforms

halx99:
However, LGPL can be used at commercial witout midification and linked as shared library. I will continue working for this. If you are interested and don't care about android apk size increase, you can track here: https://github.com/halx99/x-studio365/tree/master/cocos2d-x-patch/cocos/audio

crazyhappygame:
@halx99 Could you create pull request for Linux only? 
Currently for Linux commercial library fmod is used. LGPL is much better than commercial license .... and for desktop size does not matter.

mpg123 and openal is already used for win32.

halx99:
Since, our project is published at china,  the OpenAL backend run at thousands of android devices. From our bugly statistics, it's very stable.

PR #18965, opened 2018-07-27
refine EditBox: add support for pressed and disabled image

summerinsects:


cocos/ui/UIEditBox/UIEditBox.cpp
```diff
@@ -26,17 +26,42 @@
 
 #include "ui/UIEditBox/UIEditBox.h"
```
drelaptop, line 27:
```
cocos/ui/UIEditBox/UIEditBox.cpp 100644 → 100755
```
most files permissions changed, a mistake?

cocos/ui/UIEditBox/UIEditBox.cpp
```diff
@@ -52,110 +77,381 @@ EditBox::~EditBox()
...
+
+
+EditBox* EditBox::create(const cocos2d::Size &size, cocos2d::ui::Scale9Sprite *normalSprite, ui::Scale9Sprite* pressedSprite, Scale9Sprite* disabledSprite)
 {
+#if defined(__GNUC__) && ((__GNUC__ >= 4) || ((__GNUC__ == 3) && (__GNUC_MINOR__ >= 1)))
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#elif _MSC_VER >= 1400 //vs 2005 or higher
+#pragma warning (push)
+#pragma warning (disable: 4996)
+#endif
```
drelaptop: why put this macro block inside `EditBox::create` function

cocos/ui/UIEditBox/UIEditBox.h
```diff
@@ -254,16 +262,124 @@ namespace ui {
...
-            
-        
+
         /**
          * Init edit box with specified size. This method should be invoked right after constructor.
          * @param size The size of edit box.
          * @param normal9SpriteBg  background image of edit box.
          * @return Whether initialization is successfully or not.
          */
+        CC_DEPRECATED_ATTRIBUTE
```
drelaptop: what's the reason to mark this function as __DEPRECATED__, I don't think it's out-of-date

tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIEditBoxTest.cpp
```diff
@@ -340,3 +341,47 @@ bool UIEditBoxTestTextHorizontalAlignment::init() {
   
     return true;
 }
+
+// UIEditBoxTestPressedAndDisabled
+bool UIEditBoxTestPressedAndDisabled::init() {
```
drelaptop, line 346: added a test case for this feature, that's wonderful

tests/cpp-tests/Classes/UITest/CocoStudioGUITest/UIEditBoxTest.cpp
```diff
@@ -340,3 +341,47 @@ bool UIEditBoxTestTextHorizontalAlignment::init() {
...
+    editbox->setPlaceholderFontColor(Color3B::WHITE);
+    editbox->setPlaceholderFontSize(editBoxSize.height/2);
+    editbox->setFontSize(editBoxSize.height/2);
+    editbox->setReturnType(ui::EditBox::KeyboardReturnType::DONE);
+    addChild(editbox);
+
+    editbox = ui::EditBox::create(editBoxSize + Size(0,40), "extensions/yellow_edit.png", "extensions/orange_edit.png", "extensions/green_edit.png");
+    editbox->setPosition(Vec2(visibleOrigin.x+visibleSize.width/2, visibleOrigin.y+visibleSize.height/2));
+    editbox->setFontColor(Color3B::GREEN);
+    editbox->setPlaceHolder("Test Disabled Image - green");
```
drelaptop, line 369: should added
```
editbox->setEnabled(false);
```
in this test case and next, since you want to test __disable__ situation.

cocos/ui/UIEditBox/UIEditBox.h
summerinsects: All of other controls which has three-state texture, such as ui::Button, ui::CheckBox, ui::RadioButton, etc, don't have a create or init function with Scale9Sprite parameters, just use std::string to specify the texture soures.

for unifying, I mark these as DEPRECATED

cocos/ui/UIEditBox/UIEditBox.cpp
summerinsects: in order to shut down some compile warnings-uses about DEPRECATED functions

cocos/ui/UIEditBox/UIEditBox.h
drelaptop: I think EditBox is a special one. when developers use EditBox, most of them don't care its three-state texture, but for Button most care. 

suggest to remove those DEPRECATED mark, and then `macro block` inside EditBox::create() function.

drelaptop (approved):

PR #18968, opened 2018-07-28
ios and mac cmake travis configs

crazyhappygame:


crazyhappygame:
@drelaptop How ios should be compiled?
I copy/paste below lines from cmake/README but it does not work
```
mkdir -p ios_cmake_build
cd ios_cmake_build
cmake .. -DCMAKE_TOOLCHAIN_FILE=$COCOS2DX_ROOT/cmake/ios.toolchain.cmake
cmake --build .
```

drelaptop:
iOS build cmd from cmake/README existed `-GXcode` option, you removed it. 

another point:

> The default build is for running on actual iOS hardware, if you want to run in the simulator, please add -DIOS_PLATFORM=SIMULATOR for architecture i386 or -DIOS_PLATFORM=SIMULATOR64 for x86_64.

```
cmake .. -DCMAKE_TOOLCHAIN_FILE=$COCOS2DX_ROOT/cmake/ios.toolchain.cmake -GXcode -DIOS_PLATFORM=SIMULATOR64
cmake --build .
```
cmd above will works

---
1. without `-GXcode`, same error occur on my macOS as the travis CI shows, maybe we can't build iOS project without Xcode, just like build Android without Android Studio.
2. without `-DIOS_PLATFORM`, sign error will occurs, because Apple ask a valid sign info when generate a app published to a iOS device. we didn't deal with the sign-process on cmake script, and I think we don't need to deal with it, for the sign-process need a specific Apple ID, it should be done by iOS developer, not cocos2d-x developer.

drelaptop:
this is a good patch, thanks @crazyhappygame . I will merge it after the travis CI passed.

crazyhappygame:
Thank you help and explanation.
Could you merge it?

PR #18970, opened 2018-07-30
Node::draw and Node::visit use wrong arg type.

ggggamer:
Should we use `FLAGS_TRANSFORM_DIRTY` here?


drelaptop (approved):
```
enum {
        FLAGS_TRANSFORM_DIRTY = (1 << 0),
        FLAGS_CONTENT_SIZE_DIRTY = (1 << 1),
```

Indeed, more clear

Issue #18140, opened 2017-08-10
QuadCommand::reIndex performance degradation.


perminovVS:
Each create QuadCommand and init, call reIndex((int)quadCount*6);
QuadCommand::reIndex fill always static buffer __indices.
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/renderer/CCQuadCommand.cpp

Why is that? It error ?
Need fill it buffer only recreate __indices?

summerinsects:
And the static buffer __indices causes memory leak

PR #18992, opened 2018-08-18
try to fix memory leak in QuadCommand

summerinsects:
#18140 

I'm not sure whether it will break compatibility

### DIFF
--- a/cocos/renderer/CCQuadCommand.cpp
+++ b/cocos/renderer/CCQuadCommand.cpp
@@ -37,7 +37,7 @@
 NS_CC_BEGIN
 
 int QuadCommand::__indexCapacity = -1;
-GLushort* QuadCommand::__indices = nullptr;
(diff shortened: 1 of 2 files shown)


minggo:
why it has memory leak?

summerinsects:
For `static GLushort* QuadCommand::__indices`, which was newed in method `QuadCommand::reIndex`, nowhere deleted it

minggo:
When resizing index, old indices are pushed into `_ownedIndices`, which is freed in destructor.

summerinsects:
but the the last time allocation was not freed.
I debug it on vs with crt library functions _CrtDumpMemoryLeaks and _CrtSetBreakAlloc, and it breaks here

minggo:
We uses it as a cache, don't freed the last one or we will malloc/free memories all the time. vs debugger doesn't treat global memory well.

PR #18993, opened 2018-08-19
add line spacing support for system font on iOS and Android

summerinsects:


drelaptop:
Looks good!

drelaptop (approved):

summerinsects:
sorry for late reply
I saw in the **testcpp** file **LabelTestNew.cpp** `LabelSystemFontTest::LabelSystemFontTest` has called `setLineSpacing`, so it may has different effect

drelaptop:
thanks for share this, I should check the Label test cases firstly.

PR #19005, opened 2018-08-28
[Android] The background music resumes even though the application is on the background

gibertoni:
**When** the game is active but there is another application on foreground
**If** the device is locked or sleeps
**Then** the application that was running on foreground will be dismissed and trigger a onWindowsFocusChanged(true) event while the game is already on background

Steps to reproduce:

- Open the game and check that background music is playing
- Open another application that takes focus (e.g. Facebook Messenger's chat heads)
- Lock the device
- Verify that the music pauses (since it enters onPause), but then the music resumes (since onWindowsFocusChanged is called)


drelaptop (commented):
It's hard to reproduce, but this patch looks good, and make the `resumeIfHasFocus` more safely.

PR #19006, opened 2018-08-29
fix crash

minggo:


minggo:
we are going to remove ccGLStateCache, so this PR is not needed

PR #19008, opened 2018-08-31
Reimpl websocket

PatriceJiang:
libwebsockets use libuv & reimplement multi-threading
- reduce data processing latency
- improve thread safety


crazyhappygame:
Could you provide some background why we need those changes and what problems will be solved?

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+
+            void pushFront(T &ele) { _TMP_CC_LOOP_TS_LOCK; _data.push_front(ele); }
+            void pushFront(const T &ele) { _TMP_CC_LOOP_TS_LOCK; _data.push_front(ele); }
+            void pushFront(T &&ele) { _TMP_CC_LOOP_TS_LOCK; _data.push_front(ele); }
+
+            void popBack() { _TMP_CC_LOOP_TS_LOCK; _data.pop_back(); }
+            T popFront() { _TMP_CC_LOOP_TS_LOCK; assert(_data.size() > 0); T x = _data.front(); _data.pop_front(); return x; }
+
+            T &front() { _TMP_CC_LOOP_TS_LOCK; return _data.front(); }
+            T &back() { _TMP_CC_LOOP_TS_LOCK; return _data.back(); }
```
crazyhappygame: Methods with T& (and others returning &) are not thread safe.
```
thread1:
auto& myValue= ThreadSafeQueue<Value>.back()
thread2:
ThreadSafeQueue<Value>.popBack();
```
Race condition: myValue point to invalid object

cocos/network/WebSocketImpl.cpp
```diff
@@ -0,0 +1,940 @@
...
+                CCASSERT(d <= _remain, "data not enought!");
+                _payload += d;
+                _remain -= d;
+                _consumed += d;
+            }
+
+            size_t consumed() { return _consumed; }
+            bool isBinary() { return _isBinary; }
+        private:
+            uint8_t *_buffer = nullptr;
```
crazyhappygame: why not std::vector<uint8_t> buffer; ?

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
+/****************************************************************************
+Copyright (c) 2008-2010 Ricardo Quesada
+Copyright (c) 2009      Valentin Milea
+Copyright (c) 2010-2012 cocos2d-x.org
+Copyright (c) 2011      Zynga Inc.
+Copyright (c) 2013-2016 Chukong Technologies Inc.
+Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
```
minggo: The Copyright is `Copyright (c) 2018 Xiamen Yaji Software Co., Ltd.`, it is nothing about other person or companies.

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+#include <mutex>
+
+#include <cassert>
+
+#ifdef _TMP_CC_LOOP_TS_LOCK
+#error "_TMP_CC_LOOP_TS_LOCK already defined!"
+#endif // _TMP_CC_LOOP_TS_LOCK
+
+
+#define _TMP_CC_LOOP_TS_LOCK std::lock_guard<std::recursive_mutex> guard(_mtx)
```
minggo: It is better use `typedef`

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+            void pushFront(const T &ele) { _TMP_CC_LOOP_TS_LOCK; _data.push_front(ele); }
+            void pushFront(T &&ele) { _TMP_CC_LOOP_TS_LOCK; _data.push_front(ele); }
+
+            void popBack() { _TMP_CC_LOOP_TS_LOCK; _data.pop_back(); }
+            T popFront() { _TMP_CC_LOOP_TS_LOCK; assert(_data.size() > 0); T x = _data.front(); _data.pop_front(); return x; }
+
+            T &front() { _TMP_CC_LOOP_TS_LOCK; return _data.front(); }
+            T &back() { _TMP_CC_LOOP_TS_LOCK; return _data.back(); }
+
+            void clear() { _TMP_CC_LOOP_TS_LOCK; return _data.clear(); }
```
minggo: void function why has return?

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+
+            size_t size() { return _data.size(); }
```
minggo: please add const

cocos/loop/Collections.h
minggo: good catch

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+                {
+                    iterFn(*m);
+                }
+            }
+            std::recursive_mutex& getMutex() { return _mtx; }
+        private:
+            std::unordered_map<K, std::vector<V> > _data;
+            std::recursive_mutex _mtx;
+        };
+
```
minggo: please remove needless spaces

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+            std::recursive_mutex _mtx;
+            std::list<T> _data;
+        };
+
+
+
+        template<typename K, typename V>
+        class ThreadSafeMap {
+        public:
+            V & operator[](const K &key) { _TMP_CC_LOOP_TS_LOCK; return _data[key]; }
```
minggo: no thread safe as mentioned above

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+            void insert(const K &key, V &value) { _TMP_CC_LOOP_TS_LOCK; _data.insert(std::make_pair(key, value)); }
```
minggo: V& can be const too

cocos/loop/Collections.h
PatriceJiang: `ThreadSafeQueue` is designed to protect inner list only, not every element inside.

cocos/loop/Collections.h
```diff
@@ -0,0 +1,139 @@
...
+                if (_data.find(key) == _data.end()) { //not found
+                    _data.insert(std::make_pair(key, crtFn()));
+                }
+                return _data[key];
+            }
+            V &onceInit(const K &key, std::function<void(V&)> initFn)
+            {
+                _TMP_CC_LOOP_TS_LOCK;
+                if (_data.find(key) == _data.end()) { //do init once
+                    initFn(_data[key]);
```
minggo: _data[key] is not exist, why use it in initFn?

cocos/loop/LoopEvent.h
```diff
@@ -0,0 +1,58 @@
...
+        public:
+
+            LoopEvent(const std::string &msg) :message(msg) {}
+            LoopEvent(const std::string &msg, void *data) :message(msg), data(data) {}
+            LoopEvent(const LoopEvent &e) :eventName(e.eventName), message(e.message), data(e.data) {}
+            LoopEvent(LoopEvent && e) : eventName(e.eventName), message(e.message), data(e.data) {
+                e.data = nullptr;
+            }
+
+            void setName(const char *name) { eventName = name; }
```
minggo: please use std::string

cocos/loop/LoopEvent.h
```diff
@@ -0,0 +1,58 @@
...
+            const std::string &getName() const { return eventName; }
+        private:
+            std::string eventName{ "[noname]" };
+            std::string message;
+            void *data{ nullptr };
```
minggo: Need to add `_` for class members, and it is better use `std::string _eventName = "[noname]"` as other codes.

cocos/network/WebSocketImpl.cpp
PatriceJiang: `_buffer` is a plain linear buffer, no resize operation during the lifetime. I think the pointer is simpler.

cocos/loop/LoopRunable.cpp
```diff
@@ -0,0 +1,94 @@
...
+            _uvLoop(loop), _task(tsk), _intervalMS(duration_cast<milliseconds>(interval).count())
+        {
+            uv_timer_init(loop, &_uvTimer);
+            _uvTimer.data = this;
+        }
+
+        void LoopRunable::beforeRun()
+        {
+            if (_task)
+                _task->before();
```
minggo: It is better add a space line after `if` statement

cocos/loop/ThreadLoop.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <mutex>
+
+namespace cocos2d
+{
+    namespace loop
+    {
+#if 0
+        //thread_local is not supported on iOS
+        thread_local uv_loop_t *__uvLoop = nullptr;
+        thread_local bool __uvLoopInited = false;
```
minggo: It is better put global variables into anonymous namespace

cocos/network/WebSocketImpl.h
```diff
@@ -0,0 +1,120 @@
...
+    namespace network
+    {
+        class NetDataPack;
+        class Helper;
+        
+        class WebSocketImpl : public std::enable_shared_from_this<WebSocketImpl>
+        {
+        private:
+            static int _protocolCounter;
+            static int64_t _wsIdCounter;
```
minggo: please put parameters together

minggo: and should initialize them

cocos/network/WebSocketImpl.h
```diff
@@ -0,0 +1,120 @@
...
+            WebSocket::State _state = WebSocket::State::UNINITIALIZED;
+            std::shared_ptr<Helper> _helper;
+
+        private:
+            std::string _url;
+            Uri _uri;
+            std::string _caFile;
+            std::string _caFullPath;
+            std::vector<std::string> _protocols;
+            std::string _joinedProtocols = "";
```
minggo: i think it is no need to initialize with "".

cocos/network/WebSocketImpl.h
```diff
@@ -0,0 +1,120 @@
...
+            std::vector<uint8_t> _receiveBuffer;
+            //libwebsocket fiels
+            lws *_wsi = nullptr;
+            lws_vhost *_lwsHost = nullptr;
+            lws_protocols *_lwsProtocols = nullptr;
+            int64_t _wsId;
```
minggo: should initialize the value

cocos/network/WebSocketImpl.h
```diff
@@ -0,0 +1,120 @@
...
+            int netOnConnected();
+            int netOnClosed();
+            int netOnReadable(void *, size_t len);
+            int netOnWritable();
+
+        public:
+            WebSocket::Delegate* _delegate;
+            WebSocket *_ws = nullptr;
+            WebSocket::State _state = WebSocket::State::UNINITIALIZED;
+            std::shared_ptr<Helper> _helper;
```
minggo: why has public class members?

cocos/loop/Collections.h
minggo: can consider [std::call_once](https://en.cppreference.com/w/cpp/thread/call_once)

cocos/loop/SeqItem.h
```diff
@@ -0,0 +1,47 @@
...
+        template<typename T>
+        struct SeqItem {
+            SeqItem(uint64_t id, const std::string &name, T &data) : data(data), id(id), name(name) {}
+            SeqItem(uint64_t id, const std::string &name, T &&data) : data(data), id(id), name(name) {}
+            SeqItem(uint64_t id, T &data) : data(data), id(id) {}
+            SeqItem(uint64_t id, T &&data) : data(data), id(id) {}
+            SeqItem(const SeqItem &d) :data(d.data), id(d.id), name(d.name) {}
+            T data;
+            uint64_t id;
+            std::string name = "";
```
minggo: id is not initialized, and i think string is not needed to be initialized, it will be empty if not intialized

cocos/loop/SeqItem.h
```diff
@@ -0,0 +1,47 @@
...
+        };
+
```
minggo: please remove unneeded space

cocos/network/WebSocketImpl.h
```diff
@@ -0,0 +1,115 @@
...
+            static void closeAll();
+        public:
+            WebSocketImpl(WebSocket *);
+            virtual ~WebSocketImpl();
+
+        private:
+
+            bool init(const WebSocket::Delegate& delegate, const std::string &uri, const std::vector<std::string> *protocols = nullptr, const std::string &caFile = "");
+            void sigCloseAsync();
+            void sigCloseSync(int timeoutMS = 3000);
```
minggo: it is better to avoid default value parameter

cocos/loop/ThreadLoop.cpp
```diff
@@ -0,0 +1,100 @@
...
+#define CC_LOOP_USE_THREAD_LOCAL 0
+
+namespace cocos2d
+{
+    namespace loop
+    {
+#if CC_LOOP_USE_THREAD_LOCAL
+        //thread_local is not supported on iOS
+        thread_local uv_loop_t *__uvLoop = nullptr;
+        thread_local bool __uvLoopInited = false;
```
minggo: it is better use anonymous namespace instead of global variable

cocos/loop/Looper.h
```diff
@@ -0,0 +1,497 @@
...
+            uv_async_t _uvAsync;
+            std::condition_variable _waitThreadCV;
+            std::mutex _waitThreadMtx;
+            friend class LoopMgr;
+            friend void async_handle<LoopEvent>(uv_async_t * data);
+        };
+
+
+
+        using namespace std::chrono;
```
minggo: it is better not using the codes like this in .h files.

crazyhappygame:
@minggo @PatriceJiang Could you provide some background why we need those changes? It looks like big change (with new library) but benefits are not clear for me. 

I see two points in description but there are generic and without details. 
1. reduce data processing latency - could you provide some tests and result what and how much was improved?
2. improve thread safety - could you provide list of defects fixed by this change (or create some tests when we have problem)?

vkreal:
I agree, these are huge changes without specifics to why its needed.

cocos/network/WebSocketImpl.cpp
crazyhappygame: I think that in C++ (especially in 2018 year :)  ) there is almost no place for alloc, malloc or raw new call.
Please consider user some smart pointer or container. 
with e.g. std::vector<uint8_t> _buffer

```
NetDataPack(const char *f, size_t l, bool isBinary) 
, buffer(l+LWS_PRE) // initiate vector to right size
{
    std::copy(f, f+l, &buffer[0] + LWS_PRE); // copy data
}
//No destructor needed
```

cocos/network/WebSocketImpl.cpp
```diff
@@ -0,0 +1,940 @@
...
+
+            NetDataPack(const NetDataPack &) = delete; //non-copyable
+            NetDataPack(NetDataPack&&) = delete;       //non-movable
+
+            ~NetDataPack() {
+                if (_buffer) {
+                    free(_buffer);
+                    _buffer = nullptr;
+                }
+                _size = 0;
```
crazyhappygame: No need to zero _size in destructor

cocos/network/WebSocketImpl.cpp
```diff
@@ -0,0 +1,940 @@
...
+                _isBinary = isBinary;
+            }
+
+            NetDataPack(const NetDataPack &) = delete; //non-copyable
+            NetDataPack(NetDataPack&&) = delete;       //non-movable
+
+            ~NetDataPack() {
+                if (_buffer) {
+                    free(_buffer);
+                    _buffer = nullptr;
```
crazyhappygame: Not need to set nullptr in destructor

crazyhappygame (changes requested):

external/config.json
```diff
@@ -1,6 +1,6 @@
 {
-    "version": "v3-deps-142",
-    "zip_file_size": "136974507",
+    "version": "v3-deps-143",
```
crazyhappygame, line 2: In this code review we update some libraries to latest version.
I propose to split this code review to at least 2 parts:
1. Update only new libraries - check if build passed and it works properly 
2. Rewrite code if needed

minggo:
@crazyhappygame @vkreal sounds reasonable, @PatriceJiang please describe it in more detail, thanks.

external/config.json
minggo: @PatriceJiang i think @crazyhappygame 's suggestion is good.

PatriceJiang: OK, I will do it later.

crazyhappygame: It does not make sense do it later. 
My point is that as first we should createPR 1 for new library only. Check if it works. 
Than create PR 2 with other changes

PatriceJiang: I will close this PR and create 2 separate PR as you say, thx for your advice.

PatriceJiang:
This PR will be split into three parts
1. 3rd party library update https://github.com/cocos2d/cocos2d-x/pull/19015
2. add a websocket latency test case https://github.com/cocos2d/cocos2d-x/pull/19018
3. code rewrite https://github.com/cocos2d/cocos2d-x/pull/19025

and more details will be described in these PRs.

PR #19010, opened 2018-09-04
mark cocos2dx activity as "singleTask" on android

drelaptop:
this PR prevent the side effect from https://github.com/cocos2d/cocos2d-x/pull/18247

the issue is game will exit with exception, when cocos2dx activity isn't the first Activity 

### How to reproduce

1. add a start activity for cocos2dx game
2. jump to cocos2dx activity by `Intent`
3. cocos2dx activity finished directly, because `!isTaskRoot()` is triggered

### DIFF
--- a/templates/cpp-template-default/proj.android/app/AndroidManifest.xml
+++ b/templates/cpp-template-default/proj.android/app/AndroidManifest.xml
@@ -21,7 +21,9 @@
             android:screenOrientation="landscape"
             android:configChanges="orientation|keyboardHidden|screenSize"
             android:label="@string/app_name"
-            android:theme="@android:style/Theme.NoTitleBar.Fullscreen" >
+            android:theme="@android:style/Theme.NoTitleBar.Fullscreen"
+            android:launchMode="singleTask"
+            android:taskAffinity=""  >
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
--- a/templates/js-template-default/frameworks/runtime-src/proj.android/app/AndroidManifest.xml
+++ b/templates/js-template-default/frameworks/runtime-src/proj.android/app/AndroidManifest.xml
@@ -19,7 +19,9 @@
             android:screenOrientation="landscape"
             android:configChanges="orientation|keyboardHidden|screenSize"
             android:label="@string/app_name"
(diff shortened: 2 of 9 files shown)


crazyhappygame:
Do we still need below code after this PR:
protected void onCreate(Bundle savedInstanceState) {
        super.setEnableVirtualButton(false);
        super.onCreate(savedInstanceState);
        // Workaround in https://stackoverflow.com/questions/16283079/re-launch-of-activity-on-home-button-but-only-the-first-time/16447508
        if (!isTaskRoot()) {
            // Android launched another instance of the root activity into an existing task
            //  so just quietly finish and go away, dropping the user back into the activity
            //  at the top of the stack (ie: the last state of this task)
            // Don't need to finish it again since it's finished in super.onCreate .
            return;
        }
        // DO OTHER INITIALIZATION BELOW
        
    }

minggo:
@crazyhappygame yep, it is needed, they resolve different issues.

drelaptop:
from the NDK issue tracker, I known the android issue (#18247 fixed) still occurs on 2017, even 2018.

refer to https://issuetracker.google.com/issues/36907463 
```
vs...@gmail.com  <vs...@gmail.com>  #54 Dec 28, 2017 03:25PM 
Good Afternoon,

My issues is
1.When i open and and went to some activity.
2.then i click home button,then it go to background task.
3.when it click lauch icon .app is not resumed activity. it starts from Splash activity.
This is only occurs in released App not debug app.debug app working well,

I tried launchmode="singleInstance" or "SingleTop" what else..i tired all posibilities..other than any reply to me.
Regards,
Srinivas

in...@bigredzebra.com  <in...@bigredzebra.com>  #57 Jun 1, 2018 11:32PM 
I already opened a new issue for this problem (which has still not been resolved, but is still open) here: https://issuetracker.google.com/issues/64108432
```


So, when the cocos2dx activity is the second activity, the issue maybe occur on the start activity, then the hack code block should be move to start activity that developer custom.

drelaptop:
I think the main reason is `android:launchMode` isn't always works, for some other developers tried, and failed. so keep the hack code block is more stable. @crazyhappygame

crazyhappygame:
Thanks for explaining

abilguun:
@drelaptop 
This solution still not working on v3.17 c++ project.

PR #19011, opened 2018-09-04
ClassLoader.loadClass use 'Binary names'.

ggggamer:
JniHelper::getStaticMethodInfo's implement using Cocos2dxActivity's classloader to load Class.
In Android [doc](https://developer.android.com/reference/java/lang/ClassLoader#name), loadClass method's param is 'Binary names' style('a.b.c').

All place using `JniHelper::getStaticMethodInfo'` or `JniHelper::callStaticXXXXMethod`  need change.

This will fix following warning:
> W dalvikvm: dvmFindClassByName rejecting 'org/cocos2dx/lib/Cocos2dxHttpURLConnection'

PS:   JNIEnv::FindClass 's param style is 'a/b/c'.

### DIFF
--- a/cocos/audio/android/jni/cddandroidAndroidJavaEngine.cpp
+++ b/cocos/audio/android/jni/cddandroidAndroidJavaEngine.cpp
@@ -39,7 +39,7 @@ THE SOFTWARE.
 #define  LOGD(...)  __android_log_print(ANDROID_LOG_DEBUG,LOG_TAG,__VA_ARGS__)
 
 // Java class
-static const std::string helperClassName = "org/cocos2dx/lib/Cocos2dxHelper";
+static const std::string helperClassName = "org.cocos2dx.lib.Cocos2dxHelper";
 
 using namespace cocos2d;
 using namespace cocos2d::experimental;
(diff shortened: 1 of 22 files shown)


drelaptop:
thanks for this contribution, I have confirmed `loadClass method's param should be 'Binary names' style`. this is corrected. 

---

another side about the __warning__, I wonder how exactly it occurs. when I run cpp-tests on Android, I didn't found any logs like `dvmFindClassByName rejecting`, and on Forum I found some thread

https://discuss.cocos2d-x.org/t/solved-httpurlconnection-fail-in-android/22789

it was solved without param style modified. any reproduce for this warning?

ggggamer:
For quick test, you can run cpp-tests and enter "22: JNIHelper". 
First time you enter, you will see:
> W dalvikvm: dvmFindClassByName rejecting 'org/cocos2dx/cpp_tests/JNITest'

If we change tests/cpp-tests/Classes/JNITest/JNITest.cpp line-51 from:
`const std::string classPath = "org/cocos2dx/cpp_tests/JNITest";`
to
`const std::string classPath = "org.cocos2dx.cpp_tests.JNITest";`
warning  is removed.

drelaptop:
following the steps, I tested on 2 devices, reproduce this issue on __MI 4W__ (Android 4.4), no reproduce on __S7 Edge__ (Android 8.0). 

maybe both style be compatible on higher Android versions. Totally, this is a good patch.

drelaptop (approved):

Issue #19012, opened 2018-09-04
remove ccGLStateCache


minggo:
Forum discussion thread: https://discuss.cocos2d-x.org/t/remove-ccglstatecache/43517

minggo:
Fixed in https://github.com/cocos2d/cocos2d-x/issues/19013.
Should add release note about how to modify the codes to adapt it.

sonyps5201314:
Still crash,after build with the latest cocos git version:
android after share information to weixin, after back to my game, cocos crash...
![image](https://user-images.githubusercontent.com/20564099/45457178-60764400-b720-11e8-9b52-7ba4fbbfeb51.png)

sonyps5201314:
if modify like this ,it will not crash,but the callback text will not show.
```c++		
		if (_buffersVBO[1])
		{
			glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, _buffersVBO[1]);

			glDrawElements(GL_TRIANGLES, (GLsizei)numberOfQuads * 6, GL_UNSIGNED_SHORT, (GLvoid*)(start * 6 * sizeof(_indices[0])));
		}
		else
		{
			int i = 0;
			i++;
		}
```

sonyps5201314:
Can you provide me with a stable cocos version, which doesn't require any new features, as long as each function is working normally. Cocos has too many problems when switching between front and back processes.
Thanks for your works.

minggo:
I think it is another issue. As you see, now cocos2d-x just fixes bugs, not adding new features. The PR of this issue doesn't fix all bugs, just fix some potential bugs. Could you please create a new issue and provide a test case to reproduce it?

sonyps5201314:
I create a new issue on this problem [#19039](https://github.com/cocos2d/cocos2d-x/issues/19039). Thanks.
@minggo

drelaptop:
issue related PR merged, so close this.

PR #19013, opened 2018-09-04
Remove ccglstatecache

minggo:


minggo:
Fixed in https://github.com/cocos2d/cocos2d-x/issues/19012.
Should add release note about how to modify the codes to adapt it.

PR #19014, opened 2018-09-05
CameraBackgroundSkyBoxBrush didn't need release vao vbo.

ggggamer:
When Android opengl context recreated, our `EVENT_RENDERER_RECREATED` event listener don't need to release old VAO VBO buffer.

For example:
when `CameraBackgroundSkyBoxBrush` init, it call `glGenBuffers` and gets [5, 6]
other nodes using VBO call `glGenBuffers` may get [7, 8]
...
Then app recreate, `EVENT_RENDERER_RECREATED` emitted, event listener's order may be:
other nodes call `glGenBuffers`, get [5, 6]
`CameraBackgroundSkyBoxBrush::initBuffer`, release [5, 6], call `glGenBuffers` gets [7, 8].

So the other node's vbo buffer is released by `CameraBackgroundSkyBoxBrush`.

### DIFF
--- a/cocos/2d/CCCameraBackgroundBrush.cpp
+++ b/cocos/2d/CCCameraBackgroundBrush.cpp
@@ -330,6 +330,9 @@ CameraBackgroundSkyBoxBrush::CameraBackgroundSkyBoxBrush()
     _backToForegroundListener = EventListenerCustom::create(EVENT_RENDERER_RECREATED,
                                                             [this](EventCustom*)
                                                             {
+                                                                _vao = 0;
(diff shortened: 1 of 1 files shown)


ggggamer:
Or maybe we can just remove these lines:
https://github.com/cocos2d/cocos2d-x/blob/80279cbf214a6a0760b92a5fcc2c4bc6be971c5f/cocos/2d/CCCameraBackgroundBrush.cpp#L496-L506

ggggamer:
I found some other similar problem, PR: https://github.com/cocos2d/cocos2d-x/pull/19023

Issue #19022, opened 2018-09-07
UI clipping problem when reload Android openGL context


pbs0512:
- cocos2d-x version: 3.15
- devices test on: Android 8.0
- developing environments
   - NDK version: r13b (64-bit)

Steps to Reproduce:

1. The test has clippingListView in the previous scene when multiple scenes are nested.
2. Continuing to click on the Android power button will recreate the opengl context.
3. Then move to the previous scene (there is a clipping Listview).
4. There is a bug where Listview childrens are not visible.

error log
> OpenGL error 0x0502 in D:/test/frameworks/runtime-src/proj.android/app/jni/…/…/…/…/cocos2d-x/cocos/2d/CCDrawNode.cpp onDraw 362

ggggamer:
What's your device?   I want to fix a crash that need monitor opengl context recreate situation, but it's too difficult to reproduce.

pbs0512:
@ggggamer thank you for reply. my device is samsung galaxy 9. but it also happens on other devices.

---Cause of error
1. Create a scene and add clipping ui.
2. Push another scene.
3. At this time, an onExit occurs on the drawnode.
4. This will cause the EVENT_RENDERER_RECREATED event (https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/2d/CCDrawNode.cpp#L286) to pause.
5. android reload gl context, the EVENT_RENDERER_RECREATED event of drawNode will not happen.

ggggamer:
I can't reproduce "opengl context reload", how did you monitor this?

For your crash, Did you use `TransitionScene`?

pbs0512:
Turn the screen off, then on, and repeat with the device power button.

ggggamer:
`TransitionScene::onEnter` will disable EventDispatcher so EVENT_RENDERER_RECREATED event can't emit.
You can change cocos\platform\android\javaactivity-android.cpp
https://github.com/cocos2d/cocos2d-x/blob/c99611589c11c09edce4b1266bf211e42af6ffe4/cocos/platform/android/javaactivity-android.cpp#L106-L107
to
```
cocos2d::EventCustom recreatedEvent(EVENT_RENDERER_RECREATED);
if (director->getEventDispatcher()->isEnabled())
{
    director->getEventDispatcher()->dispatchEvent(&recreatedEvent);
}
else
{
    director->getEventDispatcher()->setEnabled(true);
[...]

PR #19023, opened 2018-09-07
Fix some EVENT_RENDERER_RECREATED problem.

ggggamer:
1. Use CC_ENABLE_CACHE_TEXTURE_DATA macro to enable/disable EVENT_RENDERER_RECREATED event listener.
2. Terrain use a local variable, it should be member variable, cause listener didn't remove.
3. CameraBackgroundSkyBoxBrush released cached buffer objects, but these buffer objects may be invalid.
4. DrawNode's listener is `init`, every time event emitted, it add another new listener. This will realloc unneed memory in the same time.

### DIFF
--- a/cocos/2d/CCCameraBackgroundBrush.cpp
+++ b/cocos/2d/CCCameraBackgroundBrush.cpp
@@ -325,11 +325,17 @@ CameraBackgroundSkyBoxBrush::CameraBackgroundSkyBoxBrush()
 , _texture(nullptr)
 , _actived(true)
 , _textureValid(true)
+#if CC_ENABLE_CACHE_TEXTURE_DATA
+, _backToForegroundListener(nullptr)
+#endif
 {
-#if (CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID || CC_TARGET_PLATFORM == CC_PLATFORM_WINRT)
+#if CC_ENABLE_CACHE_TEXTURE_DATA
     _backToForegroundListener = EventListenerCustom::create(EVENT_RENDERER_RECREATED,
                                                             [this](EventCustom*)
                                                             {
+                                                                _vao = 0;
+                                                                _vertexBuffer = 0;
+                                                                _indexBuffer = 0;
                                                                 initBuffer();
                                                             }
                                                             );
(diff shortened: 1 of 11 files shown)


Dimon4eg:
Hi @ggggamer
Thanks for this PR!
Which problems have been fixed? Any crash stacks?
Is it well tested? On which devices/platforms?

ggggamer:
> 1.Use CC_ENABLE_CACHE_TEXTURE_DATA macro to enable/disable EVENT_RENDERER_RECREATED event listener.

This is just code style, we defined a global macro, why not use it.

> 2.Terrain use a local variable, it should be member variable, cause listener didn't remove.

This is a typo,  because Terrain::_backToForegroundListener is not in construcor's initializer list,  so destructor can't find and remove it, when opengl context recreate,  a pointer to Terrain::reload will be called but terrain node is released, it may crash.

To reproduce, just:
1. Install cpp-tests demo.
2. Enter '47:Node: Terrain' test.
3. Click 'back' to list. 
4. Click Android power button several times.
opengl context may be lost, when you back game, crash may happen.

ggggamer:
> 3.CameraBackgroundSkyBoxBrush released cached buffer objects, but these buffer objects may be invalid.

This is hard to reproduce, I tested several times but don't catch it. But we really don't need to delete those invalid buffer objects.

ggggamer:
> 4.DrawNode's listener is init, every time event emitted, it add another new listener. This will realloc unneed memory in the same time.

```
auto pDrawNode = DrawNode::create();
pDrawNode->drawDot(Vec2::ZERO, 1.0f, Color4F::BLACK);

auto pScene = Scene::create();
pScene->addChild(pDrawNode);
Director::getInstance()->runWithScene(pScene);
```

Wait opengl context recreate, DrawNode will realloc more memory which is not needed.
https://github.com/cocos2d/cocos2d-x/blob/6a442192d456456545a7ffdd6ae47da9d9a8f447/cocos/2d/CCDrawNode.cpp#L208-L210

Have a look at https://github.com/cocos2d/cocos2d-x/issues/19022.

ggggamer:
@Dimon4eg  @drelaptop

drelaptop:
thanks for so detail explanation , for renderer issue I will invite @minggo to review.

minggo:
Looks good to me, thanks.

PR #19025, opened 2018-09-07
Rewrite websocket

PatriceJiang:
Rewrite WebSocket with `message channel`. 
- lock
The original implementation use `mutex` a lot. most fields in `WebSocket` class are guarded by mutexes.
The problem is that it's hard to understand and difficult to maintain. like deadlock issues are hard to track. 
New implementation use `message channel` to communicate between threads, few mutexes is used, and it is even safer. 
- memory
Data class `WsMessage`  and `WebSocketFrame` are poorly managed and they are merged into`NetDataPack`.
smart pointers are used to improve reliability. 
- thread safety
`WebSocket` APIs are safe not limited to the main thread. 
- latency
`thread::sleep(3ms)` between `lws_service` is omitted by using libuv


cocos/base/loop/Looper.h
```diff
@@ -0,0 +1,493 @@
...
+                }
+                else {
+                    auto &&front = _pendingFns.popFront();
+                    handleFn(front._data);
+                }
+            }
+
+            while (_pendingEvents.size() > 0)
+            {
+                SeqItem<LoopEvent> item = _pendingEvents.popFront();
```
crazyhappygame, line 449: Why there is no lock here and there is lock in while loop above?

cocos/base/loop/Collections.h
```diff
@@ -0,0 +1,129 @@
...
+
+            void popBack() { _CC_LOOP_DEFINE_LOCK_GUARD_; _data.pop_back(); }
+            T popFront() { _CC_LOOP_DEFINE_LOCK_GUARD_; assert(_data.size() > 0); T x = _data.front(); _data.pop_front(); return x; }
+
+            const T &front() { _CC_LOOP_DEFINE_LOCK_GUARD_; return _data.front(); }
+            const T &back() { _CC_LOOP_DEFINE_LOCK_GUARD_; return _data.back(); }
+
+            void clear() { _CC_LOOP_DEFINE_LOCK_GUARD_; _data.clear(); }
+
+            size_t size() const { return _data.size(); }
```
crazyhappygame: why size without lock?

cocos/base/loop/Looper.h
```diff
@@ -0,0 +1,493 @@
...
+        void Looper<LoopEvent>::wait(const std::function<void()> &fn, int timeoutMS)
+        {
+            if (isCurrentThread())
+            {
+                _pendingFns.pushBack(SeqItem<DispatchF>(genSeq(), fn));
+                onNotify();
+            }
+            else
+            {
+                std::condition_variable cv;
```
crazyhappygame: Could you explain this part o code?
What is a point tocreate local mutex conditional variable lock and than pass reference and than wait?
It seems like when wait_for will expire we may have rice condition

crazyhappygame (changes requested):

cocos/base/loop/Looper.h
PatriceJiang: `popFront` will be invoked only in `onNotify()`, one threaded. `item` will not be modified.

cocos/base/loop/Collections.h
PatriceJiang: `size()` itself is threadsafe, lock make no point here.

cocos/base/loop/Looper.h
PatriceJiang: `lock2` is not used, I will drop it.

PatriceJiang: >It seems like when wait_for will expire we may have rice condition.

 yes. main thread should not be frozen forever even if it happens, and timeout value should be chosen properly.

cocos/base/loop/Collections.h
crazyhappygame: No, size of std::list (and any other and container) is not thread safe.
If in thread1 size is changed and thread2 check size this is race condition.

crazyhappygame (changes requested):

cocos/base/loop/Looper.h
crazyhappygame: I am so sorry I do not understand it....
Why we need mutex mtx as local function variable?

cocos/base/loop/Looper.h
crazyhappygame: problem is with size.
Size is not thread safe

crazyhappygame (changes requested):

cocos/base/loop/Looper.h
PatriceJiang: `size()` returns `size_t` value, it is 32-bits wide on 32-bits platform and 64-bits wide on 64-bits platform. 
single read (or write) is an atomic operation.

cocos/base/loop/Collections.h
PatriceJiang: add lock inside `size()` dose not help.

cocos/base/loop/Looper.h
PatriceJiang: the mutex is used in `cv.wait` and it will only be invoked once locally.

cocos/base/loop/Looper.h
crazyhappygame: That was (maybe) true in past for single core processors. Currently with mulitocore processors with multiple level or caches it is not true any more. E.g. without mutex second thread(on second core) may not see that size was changed

cocos/base/loop/LoopRunable.h
```diff
@@ -0,0 +1,64 @@
...
+            void stop();
+            time_point<high_resolution_clock> expectTime()
+            {
+                return _startTime + milliseconds(_intervalMS * _updateTimes);
+            }
+        private:
+            Loop * _task = nullptr;
+            uv_loop_t *_uvLoop = nullptr;
+            uv_timer_t _uvTimer;
+            int64_t _intervalMS = 3000LL;
```
crazyhappygame, line 58: Pleas use std::chrono here and in any other places when time/duration is used

cocos/base/loop/Looper.h
```diff
@@ -0,0 +1,501 @@
...
+
+            uv_loop_t *getUVLoop() { assert(_initialized); return _uvLoop; };
+
+        private:
+            void notify();
+            void onNotify();
+            void onStop();
+            void onRun();
+
+#if defined(__GNUC__)
```
crazyhappygame, line 122: Why this is needed

cocos/base/loop/Looper.h
crazyhappygame: 1. But why for timeoutMS==0 add something to _pendingFns in thread1 and than wait inside thread1 until thread2 will finish?
2. For timeoutMS>0 is when cv.wait_for(lock, milliseconds(timeoutMS)); will expire than std::condition_variable cv; but cv is passed as reference to lambda [&cv  this is crash/undefined behaviur....

cocos/base/loop/Looper.h
crazyhappygame: 1. OK. Size is fixed. Thank you! By the way e.g. here is https://baptiste-wicht.com/posts/2012/03/cp11-concurrency-tutorial-part-2-protect-shared-data.html there is simple example which you can run multiple times and see that int is really not atomic 
2. But I see another problem 
```
while(_pendingEvents.size() > 0)
//Between those two lines another thread may add remove item from  _pendingEvents and we have undefined behavior
SeqItem<LoopEvent> item = _pendingEvents.popFront();
```

cocos/base/loop/Looper.h
```diff
@@ -0,0 +1,501 @@
...
+            ThreadSafeMapArray<std::string, EventCF> _callbackMap;
+            ThreadSafeQueue<SeqItem<LoopEvent>> _pendingEvents;
+            ThreadSafeQueue<SeqItem<DispatchF>> _pendingFns;
+            
+#if defined(__GNUC__)
+            volatile uint64_t _eventFnSeq = 0;
+#else
+            std::atomic_uint64_t _eventFnSeq{ 0 };
+#endif 
+            bool _forceStoped = false;
```
crazyhappygame, line 145: I think all those members should be changed to atomic protected by mutex

crazyhappygame (changes requested):
@minggo @PatriceJiang I did not spend much time on review but current implementation have many threads problems. Such problems are hard to reproduce fix and debug ....

I think that even if we fix all those problems we will have code which is much more complicated than current implementation with more or less the same number of locks and atomics and performance problems...

Please double check current implementation. All variables (even bools and integers) must be protected by mutex (or change to atomic) if there is at least one write/update in multi-threaded code.

I checked current WebSocket.cpp and I think is much simpler to understand what's going on than this PR. 

Instead of rewriting everything I propose to fix/simplify current WebSocket.cpp implementation. 
uv is very nice library but I doubt if it fix all our current poblems.

minggo:
@crazyhappygame ok, i will take more time on this PR. The biggest purpose of this PR is simplify the codes and make it more safe. If you think it is more complex, then it may have problem.

cocos/base/loop/Looper.h
PatriceJiang: my fault, I will use shared_ptr to protect `cv`

PatriceJiang:
It is too complicated and risky, so closed.

PR #19031, opened 2018-09-09
small Sprite::setTexture optimization

tankorsmash:
since `GLProgramState::getOrCreateWithGLProgramName` takes a `const std::string&` instead of `const char*`, this saves generating one each time this is called.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -373,9 +373,12 @@ void Sprite::setTexture(const std::string &filename)
 
 void Sprite::setTexture(Texture2D *texture)
 {
+    //onetime conversion to std::string since GLProgramState::getOrCreateWithGLProgramName takes a std::string instead of const char*
+    static std::string SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP = GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP;
+
     if(_glProgramState == nullptr)
     {
-        setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
+        setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
     }
     // If batchnode, then texture id should be the same
     CCASSERT(! _batchNode || (texture &&  texture->getName() == _batchNode->getTexture()->getName()), "CCSprite: Batched sprites should use the same texture as the batchnode");


cocos/2d/CCSprite.cpp
```diff
@@ -373,9 +373,12 @@ void Sprite::setTexture(const std::string &filename)
 
 void Sprite::setTexture(Texture2D *texture)
 {
+    //onetime conversion to std::string since GLProgramState::getOrCreateWithGLProgramName takes a std::string instead of const char*
+    static std::string SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP = GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP;
+
     if(_glProgramState == nullptr)
     {
-        setGLProgramState(GLProgramState::getOrCreateWithGLProgramName(GLProgram::SHADER_NAME_POSITION_TEXTURE_COLOR_NO_MVP, texture));
```
drelaptop, line 378: many codes use the `GLProgram::SHADER_NAM...` like this, and then cause another translation from `const char*` to `std::string`.

To correct this issue, I think it's more proper to modify the `GLProgram::SHADER...` definition on `CCGLProgram.cpp`

tankorsmash: Do you mean that they should all be made `std::string` instead? I haven't looked for the conversions in the opposite direction, but I'll take a look and see what's going on.

vkreal: @drelaptop does it matter that much translation from `const char* to std::string`. Just wondering?

drelaptop: yes, I mean all `std::string` instead, for those are same issue with this.

> the main aim is keep uniform for same usage on similar things. change all is better than change none, than change one of them.

halx99: May be, in the future(C++17 or later): we should use std::string_view to instead ```const st::string&``` and ```const char*``` as parameter, see https://github.com/halx99/mini-asio/blob/master/src/string_view.hpp

minggo:
This function is not invoked frequently, so i think it is ok. Adding a static variable here seems weird.

Dimon4eg:
Another option is to add `GLProgramState::getOrCreateWithGLProgramName` with `cost char *`

tankorsmash:
>Adding a static variable here seems weird.

I have a [custom Action ](https://github.com/tankorsmash/Cocos2dx-ShatterSprite) that shatters a sprite into subsprites, and this change made enough of a difference for my own use (creating dozens of subsprites in a frame, potentially several times a second). 

Others have suggested a different solution than the static, so maybe those options are better!

minggo:
> I have a [custom Action ](https://github.com/tankorsmash/Cocos2dx-ShatterSprite) that shatters a sprite into subsprites, and this change made enough of a difference for my own use (creating dozens of subsprites in a frame, potentially several times a second).

It is a special usage, and how it affect the performance?

Another option is adding a static member variable.

tankorsmash:
>It is a special usage, and how it affect the performance?

Yeah, my action (well the action I ported from someone else) was called often enough and I needed to make sure it it would keep FPS high. 

I think the performance benefit in the PR is purely saving cocos from std::string instances each time `setTexture` is called. 

I'm no optimization pro, so I may be misunderstanding the effect here, but I saw a boost after making this change. I haven't tried using a profiler to make sure (beyond messing around with VerySleepyCS when I had made this change)

halx99:
I think just change all constants  shader names type to std::string is better.

minggo:
Yep, it is better.

tankorsmash:
Sounds good. I'll make those changes

PR #19034, opened 2018-09-11
cocos2d::Data: prevent use after free.

PatriceJiang:
prevent free `_bytes` when src and dest are identical.

### DIFF
--- a/cocos/base/CCData.cpp
+++ b/cocos/base/CCData.cpp
@@ -134,7 +134,7 @@ void Data::fastSet(unsigned char* bytes, const ssize_t size)
 
 void Data::clear()
 {
-    if(!_bytes) free(_bytes);
+    if(_bytes) free(_bytes);
     _bytes = nullptr;
     _size = 0;
 }
--- a/cocos/base/CCData.cpp
+++ b/cocos/base/CCData.cpp
@@ -82,7 +82,7 @@ Data& Data::operator= (Data&& other)
 
 void Data::move(Data& other)
 {
-    clear();
+    if( !isNull() && _bytes != other._bytes) clear();
     
     _bytes = other._bytes;
     _size = other._size;
@@ -108,14 +108,22 @@ ssize_t Data::getSize() const
 
 void Data::copy(const unsigned char* bytes, const ssize_t size)
 {
-    clear();
+    if (!isNull() && bytes != _bytes)
+    {
+        clear();
+    }
+    else
+    {
+        //do not free if bytes == _bytes
+        _bytes = nullptr;
+    }
 
     if (size > 0)
     {
-        _size = size;
         _bytes = (unsigned char*)malloc(sizeof(unsigned char) * _size);
         memcpy(_bytes, bytes, _size);
     }
+    _size = size;
 }
 
 void Data::fastSet(unsigned char* bytes, const ssize_t size)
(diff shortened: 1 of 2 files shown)


cocos/base/CCData.cpp
```diff
@@ -108,14 +108,22 @@ ssize_t Data::getSize() const
 
 void Data::copy(const unsigned char* bytes, const ssize_t size)
 {
-    clear();
+    if (!isNull() && bytes != _bytes)
```
Dimon4eg: Maybe just?
 if (bytes == _bytes)
    return;

minggo:
@PatriceJiang what problem does this PR fix? Could you please add a test case for it?

PatriceJiang:
it is not really a bug. 
just protect rare cases like this
```cpp
Data d;
d.copy(d.getBytes(), len); //expect to be no side effect
```

minggo:
I think it is more reasonable to protect in `clear()`, like this
```c++
void Data::clear()
{
    if (_bytes)
        free(_bytes);

    _bytes = nullptr;
    _size = 0;
}
```

cocos/base/CCData.cpp
```diff
@@ -108,14 +108,22 @@ ssize_t Data::getSize() const
...
-    clear();
+    if (bytes != _bytes)
+    {
+        clear();
+    }
+    else
+    {
+        //do not free if bytes == _bytes
+        _bytes = nullptr;
+    }
```
minggo: Why need to set to nullptr here? It seems it has not effect.

PatriceJiang: when `size ==  0` and `bytes == _bytes`, `_bytes` should be pointed to nullptr

cocos/base/CCData.cpp
```diff
@@ -126,7 +134,7 @@ void Data::fastSet(unsigned char* bytes, const ssize_t size)
 
 void Data::clear()
 {
-    free(_bytes);
+    if(!_bytes) free(_bytes);
```
Dimon4eg: Here is wrong check.
Should be if(_bytes)

Dimon4eg (changes requested):

cocos/base/CCData.cpp
PatriceJiang: my fault

Dimon4eg:
Now in copy memory leak possible:

Data d;
int buff[3] = {1, 2, 3};
d.copy(buff, 3); // memory allocated
d.copy(d.getBytes(), 0); //leak here

cocos/base/CCData.cpp
minggo: But if `size == 0`, then it  means doesn't copy anything.

PatriceJiang: yes, this is a better way

PatriceJiang:
1. prevent copy when `size` is negative
2. no copy when `bytes == _bytes`
```cpp
ssize_t Data::copy(const unsigned char* bytes, const ssize_t size)
{
    CCASSERT(size >= 0, "copy size should be non-negative");

    if (size <= 0) return 0;

    if (bytes != _bytes)
    {
        clear();
    }
    else
    {
        _size = size;
        return _size;
    }
    
    if (size > 0)
    {
        _bytes = (unsigned char*)malloc(sizeof(unsigned char) * _size);
        memcpy(_bytes, bytes, _size);
    }
    _size = size;
    return _size;
}
```
@Dimon4eg what do you think?

Dimon4eg:
Much better. Just don't like return in the middle of the function.

Dimon4eg (approved):
Perfect!

minggo (approved):

PR #19035, opened 2018-09-11
CCFileUtils: improve thread safety

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/16814


cocos/base/CCAtomic.h
```diff
@@ -0,0 +1,75 @@
...
+
+namespace cocos2d {
+    namespace atomic
+    {
+        template<typename T>
+        class Guard {
+        public:
+            Guard(std::shared_ptr<T> d, std::shared_ptr<std::recursive_mutex> mtx) :_mtx(mtx), _data(d) { _mtx->lock(); }
+            Guard(const Guard &o) :_mtx(o._mtx), _data(o._data) { _mtx->lock(); }
+            Guard(Guard &&o) :_mtx(o._mtx), _data(o._data) { o._data = nullptr;  _mtx->lock(); }
```
crazyhappygame: Why move constructor do lock?

cocos/platform/CCFileUtils.cpp
```diff
@@ -552,15 +552,20 @@ void FileUtils::setDelegate(FileUtils *delegate)
 }
 
 FileUtils::FileUtils()
-    : _writablePath("")
+    : _writablePath(new std::string(""))
+    , _searchResolutionsOrderArray(new std::vector<std::string>())
+    , _searchPathArray(new std::vector<std::string>())
+    , _originalSearchPaths(new std::vector<std::string>())
+    , _defaultResRootPath(new std::string(""))
+    , _fullPathCache(new std::unordered_map<std::string, std::string>())
```
crazyhappygame, line 560: Why we need Atomic access to all those members?
I think that only fullPathCache can be accessed from many threads. Other members should be initialized once

cocos/platform/CCFileUtils.cpp
```diff
@@ -817,10 +824,14 @@ std::string FileUtils::fullPathForFilename(const std::string &filename) const
     {
         return filename;
     }
+    //dead lock check
+    auto fullPathCache = _fullPathCache.load();
```
crazyhappygame, line 828: 1. Yes. Currently it is very easy to lock mutexes in wrong order and we have dead lock ...
2. I think that instead of 6 mutexes (Atomic mutexes) we should create one mutex inside FileUtils and lock it when needed (on beginning of each function where there is access to member data? )

crazyhappygame (changes requested):
Please rollback all changes and create one std::recursive_mutex on the begging of each method.
Current implementation is very complicated and error prone

cocos/base/CCAtomic.h
PatriceJiang: it is dead code and removed

cocos/platform/CCFileUtils.cpp
PatriceJiang: most fields can be modified through `set**`/`add**` methods, and they need to be protected.

crazyhappygame: Than I think that It will be simpler to have one mutex and lock it when needed than. In each place we have to access to more than one member variable.

minggo: I think `_fullPathCache` and `_searchPathArray` need mutex, others are initialized once.

cocos/platform/CCFileUtils.cpp
minggo: I think @PatriceJiang 's design is good, it makes things easy.

cocos/base/CCAtomic.h
```diff
@@ -0,0 +1,74 @@
...
+                std::lock_guard<std::recursive_mutex> guard(*_mtx);
+                return Guard<T>(_data, _mtx);
+            }
+            const Guard<T> load() const {
+                std::lock_guard<std::recursive_mutex> guard(*_mtx);
+                return Guard<T>(_data, _mtx);
+            }
+        private:
+            std::shared_ptr<T> _data = nullptr;
+            std::shared_ptr<std::recursive_mutex> _mtx = nullptr;
```
minggo, line 150: Since it is put in `base`, and it doesn't support `Ref` and its sub classes. So should make it support `Ref` or explicitly forbid it.

cocos/platform/CCFileUtils.h
```diff
@@ -918,18 +923,18 @@ class CC_DLL FileUtils
      *  The vector contains resolution folders.
```
crazyhappygame, line 923: What about: ValueMap _filenameLookupDict;
Should be mutex protected  or not?

cocos/platform/CCFileUtils.cpp
crazyhappygame: Please check following link: https://wandbox.org/nojs/gcc-8.2.0/permlink/ASbHxrzF3j6EQJdD
And compare implementation of struct Counter and AtomicCounter. I think that class/struct with lock in each public method is much better in this case.
1. struct Counter is much simpler than AtomicCounter
2. struct Counter is 20x faster than AtomicCounter
3. I think that usage of std::recursive_mutex is bad design here
Please check qoute from: "Cplusplus Concurrency In Action"

>  Most of the time, if you think you want a recursive mutex, you probably need to
> change your design instead. A common use of recursive mutexes is where a class is
> designed to be accessible from multiple threads concurrently, so it has a mutex protecting
> the member data. Each public member function locks the mutex, does the
> work, and then unlocks the mutex. However, sometimes it’s desirable for one public
> member function to call another as part of its operation. In this case, the second
> member function will also try to lock the mutex, thus leading to undefined behavior.
> The quick-and-dirty solution is to change the mutex to a recursive mutex. This
> will allow the mutex lock in the second member function to succeed and the function
> to proceed.
>  However, such usage is not recommended, because it can lead to sloppy thinking
> and bad design. In particular, the class invariants are typically broken while the lock
> is held, which means that the second member function needs to work even when
> called with the invariants broken. It’s usually better to extract a new private member
> function that’s called from both member functions, which does not lock the mutex
> (it expects it to already be locked). You can then think carefully about the circumstances
> under which that new function can be called and the state of the data under
> those circumstances

cocos/platform/CCFileUtils.h
PatriceJiang: I missed it, and it's added.

cocos/platform/CCFileUtils.cpp
minggo: I don't have much experience of recursive mutex, @PatriceJiang please consider more carefully about it. Is recursive mutex indeed needed?

@crazyhappygame but here we don't have to modify all datas at the same time, why we need to protect all datas in the same time? For example, there are two variables a1 and a2, and we use it like:

```
// thread 1 access a1

protect a1 and a2

// thread 2 access a2 at the same time when thread 1 access a1, then thread 2 have to wait thread 1, but it is not needed if we don't protect a1 and a2 at the same time.
```

PatriceJiang: `recursive_mutex` is used to support 
```cpp
{
  Atomic a;
  auto x = a.load();
  {
 //...
      auto y = a.load(); // or invoked in other function 
  //...
  }
}
```

crazyhappygame: 1. "we don't have to modify all datas at the same time"
In almost each function we protect _fullPathCache
from the first a.load() all data are protected
2. More than 2 mutex locked in different order == dead lock
In this code 6 mutex were introduced. For example below code has dead lock (a1, a2 order  different)
```
void f1()
{
    a1.load();
    a2.load();
}  
void f2()
{
    a2.load();
    a1.load();
}
```
Are you sure that **always** all atomics variable are loaded (mutex is locked) in the same order?
a1.load()
a2.load()
a3.load()
....
Please remember  that currently in each function with load ( mutex lock) may call another function with load ( mutex lock) and so on.... Are you 100% sure that all loads are always in the same order for **all possible execution path**? (I checked and I am not sure.... )
Even if we check that in current code for all possible path all load are in the same order I think that sooner or later someone will write another member function when order will be broken (e.g. a2 load() id before a1.load() and we have dead lock.......
p.s. It is really hard to reproduce dead lock :(

PatriceJiang: current locks are arranged simply in alphabet order of field names.

crazyhappygame: Please check below two functions. "locks are arranged simply in alphabet order of field names" inside each function. But if you have functions in different files (android/win32) and basically from any function you can call any other function you can very easy create dead lock.

void f1()
{
    a2.load();
    // many lines of code
    f2();
} 

// different file
void f2()
{
    a1.load();
}

minggo: @crazyhappygame please narrow down the discussion scope. We just discuss the file CCFileUtils.cpp, it doesn't across files.

crazyhappygame: In this code review there is CCFileUtils-win32.cpp, CCFileUtils-android.cpp. :)

crazyhappygame: But let me rephrase. Are we 100% sure that in current code we do not reorder locks like below?
void f1()
{
a2.load();
f2();
}

void f2()
{
a1.load();
}

minggo: Oh, yep, if across functions, it is very difficult to make sure the order.

PatriceJiang:
@crazyhappygame thank you for your generous advices. I will rewrite a PR with a single `std::recursive_mutex`.

crazyhappygame:
Thank you!

PatriceJiang:
new pr location https://github.com/cocos2d/cocos2d-x/pull/19046

PR #19041, opened 2018-09-13
Android: getSafeAreaRect() implementation, resolve #18810

### LINKED ISSUE #18810: Add Android P cutout (aka notch) support
I just want to add a reminder, that Android P will support natively cutouts and every game, which target is below P will have reduced screen sizes.

So the framework should be Android P ready, if the final SDK will be released and someone can upload new and updated apps to the PlayStore.

gestern:
For many reasons, the implementation can not resolve all Android devices with edge screens or rounded corners on the screens (primarily because there is no API for which all conflicts have been fixed), but my implementation supports most modern devices.


cocos/base/CCConsole.cpp
```diff
@@ -492,6 +492,7 @@ Console::Console()
 , _isIpv6Server(false)
 , _sendDebugStrings(false)
 , _bindAddress("")
+, _commandSeparator(DEFAULT_COMMAND_SEPARATOR)
```
drelaptop, line 495: this is a mistake, double init same value

gestern: It's not my code here is my code: https://github.com/gestern/cocos2d-x/blob/1e780cb1fe9962c24d9aff9acbd5480052d7ca47/cocos/base/CCConsole.cpp#L488

drelaptop: we should search that codes on the branch `gestern:android-getsafearearect`, refer to:

https://github.com/gestern/cocos2d-x/blob/android-getsafearearect/cocos/base/CCConsole.cpp#L495

gestern:
So wait another year

drelaptop:
thanks anyway. 

add a link for this, https://github.com/cocos2d/cocos2d-x/issues/18810

PR #19042, opened 2018-09-14
 delete ANT check on setup.py

drelaptop:
Since 3.17, we build android only by Android Studio project (Gradle Build Script).

So we don't need ANT anymore, keeping it in `setup.py` may mistake newer developers .


VersusRain:
@drelaptop  hello,, as  you  say ,,   so Since 3.17,      we can not build android by  eclipse??   
 because without ANT ?

drelaptop:
@VersusRain mainly for google force push Android Studio to developers, so please use Android Studio

VersusRain:
@drelaptop Thanks !    : )

PR #19043, opened 2018-09-14
Android: Adds supporting edge screens

gestern:
For many reasons, the implementation can not resolve all Android devices with edge screens or rounded corners on the screens (primarily because there is no API for which all conflicts have been fixed), but my implementation supports most modern devices.

### DIFF
--- a/cocos/platform/android/CCGLViewImpl-android.cpp
+++ b/cocos/platform/android/CCGLViewImpl-android.cpp
@@ -42,6 +42,9 @@ PFNGLGENVERTEXARRAYSOESPROC glGenVertexArraysOESEXT = 0;
 PFNGLBINDVERTEXARRAYOESPROC glBindVertexArrayOESEXT = 0;
 PFNGLDELETEVERTEXARRAYSOESPROC glDeleteVertexArraysOESEXT = 0;
 
+#define DEFAULT_MARGIN_ANDROID				30.0f
+#define WIDE_SCREEN_ASPECT_RATIO_ANDROID	2.0f
+
 void initExtensions() {
      glGenVertexArraysOESEXT = (PFNGLGENVERTEXARRAYSOESPROC)eglGetProcAddress("glGenVertexArraysOES");
      glBindVertexArrayOESEXT = (PFNGLBINDVERTEXARRAYOESPROC)eglGetProcAddress("glBindVertexArrayOES");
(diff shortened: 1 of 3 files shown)


gestern:
Just another link added, #18810

gestern:
Epic tests within CI, just read and cry

drelaptop:
congratulations, most of the CI tests have passed.

Sometimes, `Appveyor` is not executed so smoothly like `Travis`, it's not related with the PR, let me trigger.

cocos/platform/android/CCGLViewImpl-android.cpp
```diff
@@ -127,6 +130,62 @@ void GLViewImpl::setIMEKeyboardState(bool bOpen)
...
+    }
+
+    float margin = DEFAULT_MARGIN_ANDROID / _scaleX;
+
+    bool isScreenRound = JniHelper::callStaticBooleanMethod("org/cocos2dx/lib/Cocos2dxHelper", "isScreenRound");
+    bool hasSoftKeys = JniHelper::callStaticBooleanMethod("org/cocos2dx/lib/Cocos2dxHelper", "hasSoftKeys");
+    if(isScreenRound) {
+        // edge screen (ex. Samsung Galaxy s7, s9, s9+, Note 9, Nokia 8 Sirocco, Sony Xperia XZ3, Oppo Find X...)
+        if(safeAreaRect.size.width < safeAreaRect.size.height) {
+            safeAreaRect.origin.y += margin * 2.f;
```
drelaptop: when the screen is round, those 4 corners is round. so I think this line `origin.y += margin * 2.f;` should be changed to `origin.y += margin;`, just like landscape situation. is right?

cocos/platform/android/CCGLViewImpl-android.h
```diff
@@ -49,6 +49,7 @@ class CC_DLL GLViewImpl : public GLView
     void end() override;
     void swapBuffers() override;
     void setIMEKeyboardState(bool bOpen) override;
+	virtual Rect getSafeAreaRect() const override;
```
drelaptop: micro typo, 4 space instead.

drelaptop (commented):
this is a great patch, could you review this micro conflict @gestern

cocos/platform/android/CCGLViewImpl-android.cpp
gestern: Unfortunately but no, this pice of code looks different type of screen ("edge screen" - https://9to5google.com/2016/05/06/how-to-edge-screen-galaxy-s7-edge/) - it is straight on one side and phisicly rounded around the vertical edges. I've tested on several devices with physical round corners but variable  `isScreenRound` doesn't true and it is true only if "edge" screen.

gestern: I admit that this code will be changed and it will be necessary to introduce exceptions, for example there are already several devices that are rounded and at the edges and at the corners.

drelaptop (approved):

cocos/platform/android/CCGLViewImpl-android.cpp
drelaptop: thanks for explaining

ijm700:
@gestern will this fix work on all android devices with edge screens since you wrote "For many reasons, the implementation can not resolve all Android devices with edge screens or rounded corners on the screens" how will our game show on devices that are not supported by this fix ?? thanks if any one could give more info on this

gestern:
> will this fix work on all android devices 

The short answer is "No". Long answer: there are now more than 24,000 different Android (with different Android OS) devices in the world and now there is no generally accepted way (except to using a database with 24,000 devices with information about them screens) to find out if they have a) rounded corners, b) rounded edges or c) a physically rounded screen d)the screen with clipping areas. The method of identification that I used it has a number of assumptions (I will not repeat it described above in the code) and does not work for all 24000 devices, but logically it works for the absolute majority of them - I personally rate it somewhere around 97.93% :) 

> how will our game show on devices that are not supported by this fix

Similar to the Apple version of this function - It will return the entire area of the screen (but this will happen only in rare cases, for example, I assume that there are modern devices that have physical buttons and rounded corners, and which do not have a aspect ratio of the screen less than 2). But I do not know about such devices.

ijm700:
@gestern does this work for seeing if there is "notch" at the top of the screen or only edge to edge display ? or it works for both "notch" and "edge to edge" display? also just to clarify when you say "Similar to the Apple version of this function" does this mean there is 2 versions of this function in Cocos2d-x ? I was just kind of confused by that statement, thanks again for helping

gestern:
> or it works for both "notch" and "edge to edge" display? 

@Izzyjm This feature works with all four factors that I described in the message above (from a... to d). But with some probabilistic assumptions, what does this mean? 
The Android device have to support Android OS no lower than API 23 - please see the following code:
https://github.com/cocos2d/cocos2d-x/blob/87070d35b452c13fca2e953e4224b50cc58c5425/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java#L765-L769
which used the system `isScreenRound()` coroutine  https://developer.android.com/reference/android/content/res/Configuration.html#isScreenRound()

> Return whether the screen has a round shape. Apps may choose to change styling based on this property, such as the alignment or layout of text or informational icons.

if your device has API lower than API 23, the function `getSafeAreaRect()` still continues to try to determine without additional permissions  whether a device with a non-rectangular screen via another factors like detecting software keys (menu) or via `android.max_aspect` parameters from `AndroidManifest.xml`.

> also just to clarify when you say "Similar to the Apple version of this function" does this mean there is 2 versions of this function in Cocos2d-x ? 

I will explain it well with a concrete example: Apple has implemented the interface of `safeAreaInsets` only starting with the 11.0+ version iOS, right? https://developer.apple.com/documentation/uikit/uiview/2891103-safeareainsets and for devices with iOS < 11.0, `getSafeAreaRect()` will return the same area as the call `getVisibleRect()`. This is what I meant.

Issue #18725, opened 2018-03-12
FileUtils::listFiles does not work on Android


sbelloz:
- cocos2d-x version: 3.16
- devices test on: Nexus 5, OnePlus 5
- developing environments
   - NDK version: any

Steps to Reproduce:

1. Calling `FileUtils::getInstance()->listFiles("path/folder)` returns always an empty vector

Files exist. I've tried with relative and absolute paths but with no success.

Similar issue described here: [http://discuss.cocos2d-x.org/t/full-path-to-directory-not-to-a-file/8919](http://discuss.cocos2d-x.org/t/full-path-to-directory-not-to-a-file/8919)

imtrobin:
Is this fixed? I'm still seeing the same problem on 3.17.1

PR #19044, opened 2018-09-14
Android: CCFileUtils listFiles impl

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/18725


cocos/platform/android/CCFileUtils-android.cpp
```diff
@@ -280,6 +280,45 @@ long FileUtilsAndroid::getFileSize(const std::string& filepath)
...
+    if(nullptr == dir) {
+        LOGD("... FileUtilsAndroid::failed to open dir %s", relativePath.c_str());
+        AAssetDir_close(dir);
+        return fileList;
+    }
+    const char *tmpDir = nullptr;
+    while((tmpDir = AAssetDir_getNextFileName(dir))!= nullptr)
+    {
+        string filepath(tmpDir);
+        if(isDirectoryExistInternal(filepath)) filepath += "/";
```
minggo, line 327: Why using `isDirectoryExistInternal`? This function is to find a file, not a directory.

cocos/platform/android/CCFileUtils-android.cpp
```diff
@@ -280,6 +280,45 @@ long FileUtilsAndroid::getFileSize(const std::string& filepath)
...
+    } else {
+        relativePath = fullPath;
+    }
+
+    if (nullptr == assetmanager) {
+        LOGD("... FileUtilsAndroid::assetmanager is nullptr");
+        return fileList;
+    }
+
+    auto *dir = AAssetManager_openDir(assetmanager, relativePath.c_str());
```
minggo, line 317: Should consider OBB files?

cocos/platform/android/CCFileUtils-android.cpp
PatriceJiang: default implementation includes directories in result 
https://github.com/cocos2d/cocos2d-x/blob/3cad2a85d49fbd787c0c68e960fe0678b28bbc09/cocos/platform/CCFileUtils.cpp#L1453-L1459

cocos/base/ZipUtils.h
```diff
@@ -265,6 +265,16 @@ typedef struct unz_file_info_s unz_file_info;
...
         bool fileExists(const std::string &fileName) const;
 
+
+        /**
+         * Get files and folders in pathname
+         *
+         * @param dirname
+         * @return
+         */
+        const std::vector<std::string> listFiles(const std::string &pathname) const;
```
minggo: I think `const` is not needed if it doesn't return a reference.

cocos/base/ZipUtils.cpp
```diff
@@ -609,6 +636,37 @@ bool ZipFile::fileExists(const std::string &fileName) const
     return ret;
 }
 
+const std::vector<std::string> ZipFile::listFiles(const std::string &pathname) const
+{
+
+    // filter files which `filename.startsWith(pathname)`
+    // then make each path unique
+
+    std::set<std::string> fileSet;
```
minggo, line 619: Why use set? Does it have duplicate elements?

cocos/base/ZipUtils.cpp
```diff
@@ -609,6 +636,37 @@ bool ZipFile::fileExists(const std::string &fileName) const
...
+        if(filename.substr(0, dirname.length()) == dirname)
+        {
+            std::string suffix = filename.substr(dirname.length());
+            const std::vector<std::string>& parts = splitFilePath(suffix);
+            if(parts.size() == 1)
+            {
+                fileSet.insert(suffix);
+            }else if(parts.size() > 1){
+                fileSet.insert(parts[0] + "/");
+            }
```
minggo: `splitPath` is too complex here, we don't need to know every part content, just need to check if it has `/` or not?

cocos/base/ZipUtils.cpp
PatriceJiang: yes, `_data->fileList` caches all files in the zip file. 
```
a/b/c/
a/b/c/d
a/b/c/e
a/b/c/f/g
```
these paths will share same component `a/b/c/`

cocos/base/ZipUtils.cpp
PatriceJiang: ok.

minggo (approved):

minggo:
@jianglong you can close and open the PR the trigger CI in future.

PatriceJiang:
okay

PR #19045, opened 2018-09-14
shorten the android build path, reduce long path error

drelaptop:
1. set ndk_module_path absolutely in Gradle, not relatively in Android.mk

    `long path error` appear many times when users switch to 3.17 release, but seldom for 3.16. I think `ndk_module_path` is one the reason.

2. shorten library name
    this will make tmp obj path shorter

3. little build bug fix on `uv_static` https://github.com/cocos2d/cocos2d-x/pull/19045/commits/f8aebbb4a490cb71bd90deeb1ab595bfd2270fc7

related with:

https://discuss.cocos2d-x.org/t/cocos2d-x-v3-17-released/42614/31?u=drelaptop
http://forum.cocos.com/t/cocos-lua-3-17-ndk-bug/65222


tools/travis-scripts/run-script.sh
```diff
@@ -75,8 +75,8 @@ function build_android()
     # popd
 
     # build cpp-tests
-    pushd $COCOS2DX_ROOT/tests/cpp-tests/proj.android
-    ./gradlew assembleRelease
```
crazyhappygame: I think that we should use ./gradlew assembleRelease (at least for some configurations)

drelaptop: a good advice, I will change 2/4 tasks into release mode

drelaptop:
https://github.com/cocos2d/cocos2d-x/pull/19045/commits/551eba649708a906e8de3a5b41013bc1cbf4ecf3 need external update at https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/325

PR #19046, opened 2018-09-14
CCFileUtils: improve thread safety (use single recursive_mutex)

PatriceJiang:
ref #16814

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -42,7 +42,7 @@ THE SOFTWARE.
 #endif
 #include <sys/stat.h>
 
-#define DECLARE_GUARD std::lock_guard<std::recursive_mutex> __guard__(_mutex)
+#define DECLARE_GUARD std::lock_guard<std::recursive_mutex> mutexGuard(_mutex)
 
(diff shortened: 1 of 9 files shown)


cocos/platform/CCFileUtils.cpp
```diff
@@ -42,6 +42,8 @@ THE SOFTWARE.
 #endif
 #include <sys/stat.h>
 
+#define DECLARE_GUARD std::lock_guard<std::recursive_mutex> __guard__(_mutex)
```
crazyhappygame: Identifiers should begen from _ and __ .
17.4.3.1.2 Global names [lib.global.names]
Certain sets of names and function signatures are always reserved to the implementation:

Each name that contains a double underscore (__) or begins with an underscore followed by an uppercase letter (2.11) is reserved to the implementation for any use.
Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.165

crazyhappygame (approved):
OK. Once again thank you!

cocos/platform/CCFileUtils.cpp
PatriceJiang: fixed

PR #19054, opened 2018-09-18
refactor cmake on use libs

drelaptop:
1. use modern cmake on prebuilt external libs, https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/326
2. use it elegant

---
this will let using pure source code external (without prebuilt) more easier.

pure source code external: https://github.com/cocos2d/cocos2d-x-external

---
Test

- tested 5 tests project on all supported platforms


crazyhappygame:
@drelaptop this great idea to integrate https://github.com/cocos2d/cocos2d-x-external as soon as possible!

Can we do integration in smaller steps? For example:
1. Add cocos2d-x-external as submodule to cocos2d-x
2. Add library one by one (only one library not everything) e.g. box2d - check if build passed, check if it work
'''
target_link_libraries(cocos2d box2d)
'''

3. Add next library e.g chipmunk
....... and so on....

drelaptop:
thanks to the test cases on [cocos2d-x-external/tests/engine](https://github.com/cocos2d/cocos2d-x-external/tree/master/tests/engine) I run it by link the `external` target built by cocos2d-x-3rd-party-libs-bin, __engine__ test case passed. 

so I think it's no need to add external libs one by one. I plan to make `cpp-empty-tests` works on mac firstly, and then other test, other platforms.

crazyhappygame:
I am not sure how that may work if there are not all libraries migrated in cocos2d-x-external.
In this PR you will not use https://github.com/cocos2d/cocos2d-x-external but 
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/326 ?

drelaptop:
right, In this PR I will use `cocos2d-x-3rd-party-libs-bin`, not `cocos2d-x-external` .

with CocosPickLibs.cmake removed, use modern style link external(no matter prebuilt and pure source)

crazyhappygame:
Thanks. Great idea!

drelaptop:
could you help to review this PR, give me some suggestion if any ? @crazyhappygame.

- `CocosPickLibs.cmake` and `CocosBuildModules.cmake` deleted
- modern cmake to link libs, include directories
- this PR have passed Travis CI on Mac & Android & iOS targets. 
- the logic of re-use cocos itself libs failed (re-add it later?)

crazyhappygame:
@drelaptop great job!
This PR is big enough I think that you should fix other build  failures (Linux, cocos_new, ) and merge it. We may improve it in next iteration.

"the logic of re-use cocos itself libs failed (re-add it later?)" - what do you mean?

drelaptop:
means the feature of using prebuilt cocos2dx libraries out of work, which is aim to save second compile time by re-use libs built before.

https://github.com/cocos2d/cocos2d-x/blob/v3/cmake/README.md#prebuilt-libraries-feature

crazyhappygame:
@drelaptop Please remove everything related to prebuilt from cmake. With cmake to compile cocos2dx once and use it many games it is enough to have file structure similar to https://github.com/cocos2d/cocos2d-x-external/tree/master/tests. There are two games (game1, game2) and all external libs and "engine" compile only once for all platforms except of Android. For Android we will have to update gradle script. We can write instruction for Android later...

drelaptop:
a good suggestion, I will consider to remove it in later PR. @crazyhappygame 

another reason I think is that the logic of using prebuilt cocos2dx in cmake disturb the normal logic of use sources. I have to add macro to wrap this, so as to complex more.

.appveyor.yml
```diff
@@ -8,9 +8,10 @@ environment:
   PYTHON_VERSION: "2.7.13"
   PYTHON_ARCH: "32"
   matrix:
-    - build_type: windows32_cocos_new_test
+#   - build_type: windows32_cocos_new_test
```
crazyhappygame: Please check that cmake works for windows

drelaptop: the `cocos new` test task depend on `cocos2dx-files.json`, and this file will been updated auto after PR merged.

sure, we can do same work  like https://github.com/cocos2d/cocos2d-x/pull/19054/commits/3cb168bf12339f50e43d1a3e9a0443a1d6c17d28 at AppVeyor, but I think it's not a necessary thing. When `cpp-tests` works on windows and `cocos new` test works on linux & android, I bet 99% the `cocos new` test works on win32.

crazyhappygame: @drelaptop My point is that there is no cmake build for win32 at all. We should have some normal (msbuild) and cmake build for win32 to be sure that everything work.
 
Environment: build_type=windows32
` & msbuild $env:APPVEYOR_BUILD_FOLDER\build\cocos2d-win32.sln /t:Build /p:Platform="Win32" /p:Configuration="Release" /m /consoleloggerparameters:"PerformanceSummary;NoSummary"`
Environment: build_type=android_cpp_tests
`& ./gradlew assembleDebug`
Environment: build_type=android_lua_tests
`& ./gradlew assembleDebug`

cmake/Modules/CocosConfigDepend.cmake
```diff
@@ -1,26 +1,26 @@
...
-        endif()
+        list(APPEND PLATFORM_SPECIFIC_LIBS ws2_32 userenv psapi winmm Version Iphlpapi opengl32)
     elseif(LINUX)
         # need review those libs: X11 Xi Xrandr Xxf86vm Xinerama Xcursor rt m
         list(APPEND PLATFORM_SPECIFIC_LIBS dl X11 Xi Xrandr Xxf86vm Xinerama Xcursor rt m)
-        foreach(_pkg OPENGL GLEW GLFW3 FMOD FONTCONFIG THREADS GTK3 SQLITE3)
-            list(APPEND PREBUILT_SPECIFIC_LIBS ${_pkg})
-        endforeach()
+        # use older cmake style on below linux libs
+        cocos_find_package(Fontconfig FONTCONFIG REQUIRED)
```
crazyhappygame, line 11: 1. I think that find_package/target_link_libraries should work.
2. All those libraries should be in https://github.com/cocos2d/cocos2d-x-external/

cocos/platform/CMakeLists.txt
```diff
@@ -22,7 +22,16 @@
 # THE SOFTWARE.
 # ****************************************************************************/
 
-if(WINDOWS)
+if(ANDROID)
```
crazyhappygame, line 25: In this file should be only:
```
add_subdirectory(android)
add_subdirectory(apple)
```
sources should be added inside android/apple

crazyhappygame (approved):
Great job.
Some thinks to do in next PRs:
1.  replace all old cmake add_definitions, include_directories... to target_*
2. Minimize number of variables. For example 
instead of 
```
set(my_sources source1.cpp source.cpp)
add_library(target
main.cpp
${my_sources}
)
```
It is recommended:
```
add_library(target main)
target_sources(target source1.cpp source.cpp)
````

File: .appveyor.yml (hunk 1 above)
Comment (drelaptop) [reply .appveyor.yml]: thanks for explantion, I will add a cmake target on win32

**Comment by drelaptop** [comment]
> replace all old cmake add_definitions, include_directories... to target_*

1. most of `include_directories` have been replaced with modern style 
```
target_include_directories(target INTERFACE Dir) // for build target
set_target_properties(${target_name} PROPERTIES INTERFACE_INCLUDE_DIRECTORIES Dir)//for imported target
```
so that those include dir will be iincluded auto

2. the modern style of `add_definitions` is `add_compile_definitions`, not target_*. seems not big difference, and sometimes we need Global definations.

> Minimize number of variables. For example

do reduce variable if necessary, for example to add source with different condition, we need variable
```
if(NOT LINUX)
    list(APPEND GAME_SOURCE
         Classes/UITest/CocoStudioGUITest/UIEditBoxTest.cpp
         )
endif()
```

File: cmake/Modules/CocosConfigDepend.cmake (hunk 2 above)
Comment (drelaptop) [reply cmake/Modules/CocosConfigDepend.cmake:11]: > find_package/target_link_libraries should work.

I tried, but failed. for those module didn't use modern style, like `target_include_directories`. so when you do this, you will get "header not found error" at least

File: cocos/platform/CMakeLists.txt (hunk 3 above)
Comment (drelaptop) [reply cocos/platform/CMakeLists.txt:25]: the reasone to add `cocos/platform/android/CMakeLists.txt` file is we need to `whole-archive` part sources, or symble might be striped when linking.

some developer make feedbacks that those games works well on android when build by `ndk-build`, but crash sometimes when build by `ndk-build`. espicailly execute some `lua` method related with `CCLuaJavaBridge.cpp` on __cocos2d-x3.17__. By compare with `ndk-build`, I found  `whole-archive` is the point.

**Comment by crazyhappygame** [comment]
This code
```
if(NOT LINUX)
  list(APPEND GAME_SOURCE
         Classes/UITest/CocoStudioGUITest/UIEditBoxTest.cpp
         )
endif()
```
 in modern cmake without variables my looks like below:
```
add_executable(target main.cpp)
.......
if(NOT LINUX)
    target_sources(target Classes/UITest/CocoStudioGUITest/UIEditBoxTest.cpp)
endif()
```

**Comment by drelaptop** [comment]
indeed, more simple and clear. this is a good choice

File: cocos/platform/CMakeLists.txt (hunk 3 above)
Comment (crazyhappygame) [reply cocos/platform/CMakeLists.txt:25]: Please add comment with link to issues explaining why whole-archive is needed

Comment (drelaptop) [reply cocos/platform/CMakeLists.txt:25]: https://github.com/cocos2d/cocos2d-x/blob/5c381d8ee60078b1ff8380b794836ea61d4ed2ee/cocos/Android.mk#L281

https://github.com/cocos2d/cocos2d-x/blob/5c381d8ee60078b1ff8380b794836ea61d4ed2ee/cocos/platform/android/Android.mk#L5

see above, ndk-build use the `LOCAL_WHOLE_STATIC_LIBRARIES ` for years, use `whole-archive` in cmake to aim same logic.

> no repo issues about it, I think we should keep the logic no change

File: tools/appveyor-scripts/build.ps1 [hunk 4]
```diff
@@ -19,15 +19,24 @@ function PushAndroidArtifacts
 If ($env:build_type -eq "android_cpp_tests") {
     Write-Host "Build tests\cpp-tests"
     Push-Location $env:APPVEYOR_BUILD_FOLDER\tests\cpp-tests\proj.android\
-    & ./gradlew assembleDebug
+    & ./gradlew assembleDebug -PPROP_BUILD_TYPE=ndk-build --parallel --info
```
crazyhappygame: Could you do cmake for android cpp test and ndk build for Android Lia?

tools/appveyor-scripts/build.ps1
```diff
@@ -19,15 +19,24 @@ function PushAndroidArtifacts
...
+    & ./gradlew assembleDebug -PPROP_BUILD_TYPE=cmake --parallel --info
     if ($lastexitcode -ne 0) {throw}
     PushAndroidArtifacts
     Pop-Location
 
 } elseif ($env:build_type -eq "android_cpp_empty_test") {
     Write-Host "Build tests\cpp-empty-test"
     Push-Location $env:APPVEYOR_BUILD_FOLDER\tests\cpp-empty-test\proj.android\
-    & ./gradlew assembleRelease
+    & ./gradlew assembleRelease -PPROP_BUILD_TYPE=cmake --parallel --info
```
crazyhappygame: This build is disabled

tools/appveyor-scripts/build.ps1
drelaptop: sure, little thing

Issue #10511, opened 2015-02-15
PhysicsBody and Follow action bug in 3.4


poowww:
In 3.4final, Follow Action doesn't work when follow Node added PhysicsBody.
In 3.3, it works.

Please fix this bug.

WenhaiLin:
@Poowww Can you provided a test case that can easily reproduce this?

poowww:
@WenhaiLin Thanks for reply.

Please rewrite "Default project's HelloWorldScene.cpp" as follows.
It's a test case which this bug can reproduce.

``` cpp:HelloWorldScene.cpp
Scene* HelloWorld::createScene()
{
    auto scene = Scene::createWithPhysics();
    auto world = scene->getPhysicsWorld();
    world->setDebugDrawMask(PhysicsWorld::DEBUGDRAW_ALL);
    auto layer = HelloWorld::create();
    scene->addChild(layer);

    return scene;
}

bool HelloWorld::init()
{
    if ( !Layer::init() )
    {
        return false;
    }

    auto groundBody = PhysicsBody::createBox(Size(2000, 100));
    groundBody->setDynamic(false);
    auto ground = Node::create();
    ground->setAnchorPoint(Vec2::ANCHOR_MIDDLE);
    ground->setPhysicsBody(groundBody);
    this->addChild(ground);
    ground->setPosition(Vec2(-1000,0));

    auto followedMaterial = PhysicsMaterial();
    followedMaterial.friction = 0;
    followedMaterial.density = 0;
    followedMaterial.restitution = 0;
    auto followedBody = PhysicsBody::createBox(Size(100,100),followedMaterial);
    followedBody->setDynamic(true);
    followedBody->setRotationEnable(false);
    auto followed = Node::create();
    followed->setAnchorPoint(Vec2::ANCHOR_MIDDLE);
    followed->setPhysicsBody(followedBody);
    this->addChild(followed);
    followed->setPosition(Vec2(-2000,300));

    auto followAction = Follow::create(followed);
    this->runAction(followAction);

    const auto acceleration = Vec2(5, 0);
    auto push = [followed, acceleration](float dt){followed->getPhysicsBody()->applyImpulse(acceleration);};
    this->schedule(push, 0.1f, "push");

    return true;
}
```

PR #19056, opened 2018-09-19
DrawNode add isolate flag

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/10511

`_debugDraw` is a child node of `Scene`. When `scene` is updated by `Follow` action, the matrix of `_debugDraw` is also affected.

### DIFF
--- a/cocos/2d/CCDrawNode.cpp
+++ b/cocos/2d/CCDrawNode.cpp
@@ -101,26 +101,7 @@ static inline Tex2F __t(const Vec2 &v)
 // implementation of DrawNode
 
 DrawNode::DrawNode(GLfloat lineWidth)
-: _vao(0)
-, _vbo(0)
-, _vaoGLPoint(0)
-, _vboGLPoint(0)
-, _vaoGLLine(0)
-, _vboGLLine(0)
-, _bufferCapacity(0)
-, _bufferCount(0)
-, _buffer(nullptr)
-, _bufferCapacityGLPoint(0)
-, _bufferCountGLPoint(0)
-, _bufferGLPoint(nullptr)
-, _bufferCapacityGLLine(0)
-, _bufferCountGLLine(0)
-, _bufferGLLine(nullptr)
-, _dirty(false)
-, _dirtyGLPoint(false)
-, _dirtyGLLine(false)
-, _isolated(false)
-, _lineWidth(lineWidth)
+: _lineWidth(lineWidth)
 , _defaultLineWidth(lineWidth)
 {
     _blendFunc = BlendFunc::ALPHA_PREMULTIPLIED;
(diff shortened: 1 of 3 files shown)


cocos/2d/CCDrawNode.h
```diff
@@ -313,12 +313,22 @@ class CC_DLL DrawNode : public Node
     
     // Overrides
     virtual void draw(Renderer *renderer, const Mat4 &transform, uint32_t flags) override;
+
+    void visit(Renderer* renderer, const Mat4 &parentTransform, uint32_t parentFlags) override;
```
gestern: You can not just override virtual functions without using a keyword `virtual`. This is incorrect from the point of view of the original architecture.

cocos/2d/CCDrawNode.h
```diff
@@ -313,12 +313,22 @@ class CC_DLL DrawNode : public Node
...
     // Get CocosStudio guide lines width.
     GLfloat getLineWidth();
 
+    /**
+    * When isolated is set, `visit()` will always get and identity as parent transform matrix.
+    * It's needed when DrawNode is used as absolute node without any effects from parent node.
+    */
+    void setIsolated(bool isolated) { _isolated = isolated; }
+
+    bool getIsolated() { return _isolated; }
```
gestern: You must use the keyword `const` if the function is immutable. In this case, the function is immutable. Another good advantage would be when you reuse a macro `CC_SYNTHESIZE(bool, _isolated, Isolated)` or change the name of the getter from `bool getIsolated() const` to `bool isIsolated() const`

cocos/2d/CCDrawNode.h
PatriceJiang: my fault, thx

cocos/2d/CCDrawNode.cpp
```diff
@@ -957,5 +958,19 @@ GLfloat DrawNode::getLineWidth()
...
+    if (_isolated)
+    {
+        //in isolate mode, draw use world coordinate directly
+        Node::visit(renderer, Mat4::IDENTITY, parentFlags);
+    }
+    else
+    {
+
+
+        Node::visit(renderer, parentTransform, parentFlags);
```
minggo: please remove unneeded spaces

cocos/2d/CCDrawNode.cpp
```diff
@@ -119,6 +119,7 @@ DrawNode::DrawNode(GLfloat lineWidth)
 , _dirty(false)
 , _dirtyGLPoint(false)
 , _dirtyGLLine(false)
+, _isolated(false)
 , _lineWidth(lineWidth)
 , _defaultLineWidth(lineWidth)
```
minggo: I think can change to initialize parameters in class declaration.  We should change to use this style when modify some codes.

Issue #19058, opened 2018-09-19
CCFileUtils-win32::isFileExistInternal fails on a directory, while CCFileUtils-linux's doesn't


tankorsmash:
Windows: https://github.com/cocos2d/cocos2d-x/blob/b26e1bb08648c11a9482d0736a236aebc345008a/cocos/platform/win32/CCFileUtils-win32.cpp#L151

Linux: https://github.com/cocos2d/cocos2d-x/blob/b26e1bb08648c11a9482d0736a236aebc345008a/cocos/platform/linux/CCFileUtils-linux.cpp#L122

The C++ file to test the result of `stat` on linux, `result` is 0 on a dir that exists:

```
#include <unistd.h>
#include <sys/stat.h>
#include <stdio.h>
#include <iostream>

int main()
{
	struct stat sts;
	auto result = stat("a dir that exists on your machine/", &sts);
	std::cout << "result is: " << result;
	return 0;
};
```

whereas `attr = GetFileAttributesW("a dir that exists")` does result in `attr & FILE_ATTRIBUTE_DIRECTORY` being non-zero.

This seems to mean `CCFileUtilsWin32::listFiles` can't work, because https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/win32/CCFileUtils-win32.cpp#L276 will always return `""`, since it ends up going down to `CCFileUtils-win32::isFileExistInternal` and failing.

PatriceJiang:
>  This seems to mean CCFileUtilsWin32::listFiles can't work

This bug was fixed in a recent PR. 
https://github.com/cocos2d/cocos2d-x/pull/19044/files#diff-cb65223cd3edf30671fefad4a1883edf

I think the Linux version of `isFileExistInternal` should not accept a folder, cause `isDirecotryExistInternal` exists at the same level.

tankorsmash:
That does seem to be the case, thanks! So this is fixed in https://github.com/cocos2d/cocos2d-x/pull/19044

PatriceJiang:
similar issue https://github.com/cocos2d/cocos2d-x/issues/18835

PR #19062, opened 2018-09-20
[linux] CCFileUtils::isFileExistInternal should reject a folder

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19058

`FileUtilsLinux::isFileExistInternal(path_of_a_directroy)` should return `false`


PatriceJiang:
@minggo please review when you are available

Issue #18620, opened 2018-01-12
[BUG] Console::loop() consumes 100% of CPU usage


gestern:
- cocos2d-x version: **any**
- devices test on: **any**

Steps to Reproduce:
1. Run socket client based of any (for example: `nc`, `telnet` )
```
nc 192.168.1.xxx <YOUR CONSOLE PORT>
> fps on
> ^C
```
2. Close connection with server via closing a socket `^C` or simulate network disconnection (without `exit` console command).
3. Console::loop starts consuming 100% of CPU, it happening because of one or more of the sockets in`Console` told `select()` to watch are ready-for-read (and/or ready-for-write) so that `select()` returns right away instead of blocking... but then the code neglects to actually recv() (or send()) any data on that socket. After failing to read/write anything, your event loop would try to go back to sleep by calling `select()` again, but of course the socket's data (or buffer space, in the ready-for-write case) is still there waiting to be handled, so `select()` returns immediately again, the buggy code neglects to do the read (or `write()`) again, and around and around we go at top speed:
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/CCConsole.cpp#L744

To understand all the horror of what happens, add the output to the console before and after the command like so:
```c
CCLOG("---before select()---");
int nready = select(_maxfd+1, &copy_set, nullptr, nullptr, &timeout_copy);
CCLOG("---after select()---");
```

It will be surprised but the interval for the appearance of these messages will be **much higher** than "60PFS" (or 16000 in tv_usec) this is what I described in the 3rd paragraph. This modal loop without any delays as not previously assumed.

Another thing that is not clear why is done is https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/CCConsole.cpp#L734-L737:
```c
    timeout.tv_sec = 0;
    
    /* 0.016 seconds. Wake up once per frame at 60PFS */
    timeout.tv_usec = 16000;
```
60FPS through what? Through network? Like argument in `select()`? This argument is a **maximum interval to wait** for the selection to complete (according to the POSIX man) not for "wake up".
This parameter must be configurable via user like:
```c
void Console::setTimeout(const timeval &timeout)
{
    _timeout = timeout;
}
```

And later `_timeout` can be used for `timeout_copy` variable. Given that the server is written very specific, and any client connection blocks it, you can safely select the value in 30 seconds by default in constructor of `Console`.

minggo:
[...]

PR #19066, opened 2018-09-20
CCConsole: remove fd from _read_set when it is closed by client.

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/18620

1. increase `select` timeout to 1 seconds
2. remove `fd` on close

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -743,10 +743,10 @@ void Console::loop()
     FD_SET(_listenfd, &_read_set);
     _maxfd = _listenfd;
     
-    timeout.tv_sec = 0;
+    timeout.tv_sec = 1;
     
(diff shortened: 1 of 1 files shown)


PatriceJiang:
please review when you are available @gestern  @minggo

cocos/base/CCConsole.cpp
```diff
@@ -743,10 +743,10 @@ void Console::loop()
     FD_SET(_listenfd, &_read_set);
     _maxfd = _listenfd;
     
-    timeout.tv_sec = 0;
+    timeout.tv_sec = 1;
     
     /* 0.016 seconds. Wake up once per frame at 60PFS */
-    timeout.tv_usec = 16000;
+    timeout.tv_usec = 0;
```
minggo, line 747: should change the comment too

PR #19070, opened 2018-09-22
Update CCSpriteFrame.cpp

WangZhechao:


cocos/2d/CCSpriteFrame.cpp
```diff
@@ -38,17 +38,21 @@ NS_CC_BEGIN
 SpriteFrame* SpriteFrame::create(const std::string& filename, const Rect& rect)
 {
     SpriteFrame *spriteFrame = new (std::nothrow) SpriteFrame();
-    spriteFrame->initWithTextureFilename(filename, rect);
-    spriteFrame->autorelease();
-
+    if(spriteFrame) {
+	spriteFrame->initWithTextureFilename(filename, rect);
```
drelaptop, line 42: only check the return value of `new` on a simple class doesn't make sense. 

it's better to do like this if you want to more strict.
```C++
    if (spriteFrame && spriteFrame->init...) {
        spriteFrame->autorelease();
        return spriteFrame;
    }

    delete spriteFrame;
    return nullptr;
```

PR #19073, opened 2018-09-24
Fix several bugs with button's title

gestern:
1. when using `setTitleLabel()` the button didn't resolve the corresponded parameters;
2. bug with measurement accuracy (ignores the fractional part) of the system font and ttf-fonts sizes;
3. bug with the size of the button title (by default) did not match the value size of the label's typeface;
4. removes high coupling the typeface parameters of button because it already contains the used label.

Tested on existing cpp-tests

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -484,7 +484,7 @@ void Label::reset()
     _bmFontPath = "";
     _systemFontDirty = false;
     _systemFont = "Helvetica";
-    _systemFontSize = 12;
+    _systemFontSize = CC_DEFAULT_FONT_LABEL_SIZE;
 
     if (_horizontalKernings)
     {
--- a/cocos/2d/CCLabel.h
+++ b/cocos/2d/CCLabel.h
@@ -40,6 +40,7 @@ NS_CC_BEGIN
  * @{
  */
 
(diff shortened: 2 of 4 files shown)


drelaptop (approved):

PatriceJiang:
There is an issue that the font size attribute can be lost when the `titleLabelType` is changed later.  Since the field is removed.

souleid:
Hey! I have Q, 
I'm Bug found during upgrade from 3.16 to 3.17.1.

I was using this code,
        auto button = ui::Button::create();
        **button->setTitleFontSize(defaultFontSize);**
        **button->setTitleFontSize(defaultFontSize);**

in 3.17.1 font size return to SYSTEM_DEFAULT_SIZE(12).

so I should changed code,
        auto button = ui::Button::create();
        **button->setTitleFontName(defaultFont);**
        **button->setTitleFontSize(defaultFontSize);**

I think you drop out title size in **setTitleFontName()**
previous code, there is exist
 TTFConfig config = _titleRenderer->getTTFConfig();
                config.fontFilePath = fontName;
                config.fontSize = _fontSize;
                _titleRenderer->setTTFConfig(config);

What is right?

bolino:
We experienced this same issue, some of our Button labels had a fixed size after updating.
Right now we did what is mentioned on the comment above, calling setTitleFontName() before setTitleFontSize().

PR #19075, opened 2018-09-25
win32: fileutils removeDirectory() failure

PatriceJiang:
when trying to remove dirs like `test0`, without postfix `/`, the function will build a pattern
like `"test0*.*"` rather than `"test0/*.*"`, which leads to `removeDirectory("test0")` not working.


minggo:
Is there any issue for it?

minggo (approved):

PatriceJiang:
I met this when profiling #19058, the directories got not deleted on windows.

Issue #18997, opened 2018-08-21
Downloader class has JS binding, no Lua binding, will it be added later?


thinkingingame:
- cocos2d-x version:3.17
Downloader类有js binding，没有lua binding，后面会加上吗？

drelaptop:
could you check this in your free time? @PatriceJiang

PatriceJiang:
okay

PR #19078, opened 2018-09-28
LUA: cc.downloader binding

PatriceJiang:
add LUA binding for `cocos2d::Downloader`


PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/18997

PR #19079, opened 2018-09-28
[android] Fix crash caused by JNIEnv null pointer in getStringUTFCharsJNI function

EricDDK:
tmp_crash_android_getstring

### DIFF
--- a/cocos/base/ccUTF8.cpp
+++ b/cocos/base/ccUTF8.cpp
@@ -309,7 +309,7 @@ bool UTF32ToUTF16(const std::u32string& utf32, std::u16string& outUtf16)
 std::string getStringUTFCharsJNI(JNIEnv* env, jstring srcjStr, bool* ret)
 {
     std::string utf8Str;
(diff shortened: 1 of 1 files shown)


EricDDK:
The probability of crash on the model in the figure(lenovo ipad) is 100 percent.

drelaptop:
did you reproduce this crash locally? 

I think the game have been out-of-work when the `JNIEnv` equals nullptr, even with your crash fixed.

EricDDK:
> did you reproduce this crash locally?
> 
> I think the game have been out-of-work when the `JNIEnv` equals nullptr, even with your crash fixed.

No,this crash only occurs on some special models, the probability is 100%, after the bug is fixed, the game is running normally. I tested it locally, Lenovo's Yoga model, Android 5.0.

I guess Lenovo's tablet Android system has experienced some secondary developments that are not in place.

drelaptop (approved):
add `nullptr` check to make more stable

Issue #18785, opened 2018-04-23
drawBatchedTriangles crash on Android


shaniamjad:
- cocos2d-x version: 3.15.1
- developing environments
   - NDK version: r13b
 -platform: IOS, Android

I am getting lots of crashes on Crashlytics in Renderer::drawBatchedTriangles() 

`glBufferData(GL_ARRAY_BUFFER, sizeof(_verts[0]) * _filledVertex, nullptr, GL_STATIC_DRAW);`
`void *buf = glMapBuffer(GL_ARRAY_BUFFER, GL_WRITE_ONLY);`
`memcpy(buf, _verts, sizeof(_verts[0]) * _filledVertex);`
`glUnmapBuffer(GL_ARRAY_BUFFER);`

Crash is happening at line `memcpy(buf, _verts, sizeof(_verts[0]) * _filledVertex);`

I am not able to reproduce this crash on my side but i have seen many instances of this crash on Crashlytics as well as on Google Play Dashboard. Crash is happening mostly on non popular devices Example: "Kobo Arc 10HD" and "Rockchip Tablet PC" the most common devices with crash. As i see it looks to me that @dumganhar you have been trying to fix some issues related to this part of the code in https://github.com/cocos2d/cocos2d-x/issues/15805.

Do you think that this issue is related? Any ideas about how it can be solved?

shaniamjad:
@minggo @dumganhar Guys any update on this?

shaniamjad:
Update:: 
If i set VBO_SIZE to a less number then the crash can be reproduced easily. on debug mode it asserts but on release mode i get the similar crash stack as well as line number which i am getting on crashlytics.
I set VBO_SIZE  to 655 in my test scenario.
and INDEX_VBO_SIZE is set using formula VBO_SIZE * 6 / 4;
With above configuration when fillVerticesAndIndices is called then `_indices[_filledIndex + i] = _filledVertex + indices[i];` sometimes _indices index increased from `INDEX_VBO_SIZE` which leads to memory corruption. and hence it leads to crash at random places.
I checked that after this `_filledVertex` `_buffersVAO` `_triBatchesToDraw` gets garbaged.

Please help, I desperately wants to fix this crash.

PaulDiakov:
In CCRenderer.cpp `drawBatchedTriangles` trigger condition in `processRenderCommand` looks like 
[...]

PR #19085, opened 2018-10-01
Boby - Possible fix for the Draw Batch Triangles Crashes

BobyIlea:
@minggo, @slackmoehrle Please check this pull request and ping other developers that you think might need to have a look at it.

Background: We are facing a lot of crashes related to the Renderer::drawBatchedTriangles function. Example issue: https://github.com/cocos2d/cocos2d-x/issues/18785 

In order to fix it I've created a sample code that seems to fix the crashes. Using this fix, the index doesn't get out of bounds, but when the bounds are exceeded some elements are not drawn any more. Though I think that not drawing elements is a better solution then a crash(or maybe leave this to the developer to decide).

Note that this is just a small proof of concept on how the crash might be solved so please leave any comments you have regarding the fix and what should be changed in order to be able to merge in the main branch. Please also express any concerns that you have related to the current implementation: did I overlook something? can this cause any major issues that slipped my mind?

Let me know what you think. Thank you!

### DIFF
--- a/cocos/renderer/CCRenderer.cpp
+++ b/cocos/renderer/CCRenderer.cpp
@@ -38,14 +38,17 @@
 #include "renderer/CCTechnique.h"
 #include "renderer/CCPass.h"
 #include "renderer/CCRenderState.h"
+#include "renderer/ccGLStateCache.h"
 
 #include "base/CCConfiguration.h"
 #include "base/CCDirector.h"
 #include "base/CCEventDispatcher.h"
 #include "base/CCEventListenerCustom.h"
 #include "base/CCEventType.h"
+#include "base/CCUserDefault.h"
 #include "2d/CCCamera.h"
 #include "2d/CCScene.h"
+#include "base/CCEventCustom.h"
 
 NS_CC_BEGIN
 
@@ -202,6 +205,8 @@ Renderer::Renderer()
 ,_triBatchesToDraw(nullptr)
 ,_filledVertex(0)
 ,_filledIndex(0)
+,_filledVertex2(0)
+,_filledIndex2(0)
 ,_glViewAssigned(false)
 ,_isRendering(false)
 ,_isDepthTestFor2D(false)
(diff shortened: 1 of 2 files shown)


cocos/renderer/CCRenderer.cpp
```diff
@@ -809,6 +829,11 @@ void Renderer::drawBatchedTriangles()
         // so most probably we won't have any benefit of using it
         glBufferData(GL_ARRAY_BUFFER, sizeof(_verts[0]) * _filledVertex, nullptr, GL_STATIC_DRAW);
         void *buf = glMapBuffer(GL_ARRAY_BUFFER, GL_WRITE_ONLY);
+        GLenum err = glGetError();
+        if (err != GL_NO_ERROR)
+        {
+            UserDefault::getInstance()->setBoolForKey("IsExitInDrawBatchedTriangle", true);
```
gestern, line 835: Use expensive IO operations (even without a "modern" version with expensive mutexes) during GPU rendering loop is unexceptable.

cocos/renderer/CCRenderer.cpp
```diff
@@ -845,11 +870,18 @@ void Renderer::drawBatchedTriangles()
     /************** 3: Draw *************/
     for (int i=0; i<batchesTotal; ++i)
     {
-        CC_ASSERT(_triBatchesToDraw[i].cmd && "Invalid batch");
-        _triBatchesToDraw[i].cmd->useMaterial();
-        glDrawElements(GL_TRIANGLES, (GLsizei) _triBatchesToDraw[i].indicesToDraw, GL_UNSIGNED_SHORT, (GLvoid*) (_triBatchesToDraw[i].offset*sizeof(_indices[0])) );
-        _drawnBatches++;
-        _drawnVertices += _triBatchesToDraw[i].indicesToDraw;
+        //CC_ASSERT(_triBatchesToDraw[i].cmd && "Invalid batch");
```
gestern, line 873: In this case, the use `CC_ASSERT()` is more right (even if it's involves the signal, so this is right behaviour for `assert()` ) than through the use of if..else constructions, since CC_ASSERT identifies the existence of an architectural issue, and second condition only hide its presence.

cocos/renderer/CCRenderer.h
```diff
@@ -258,8 +258,12 @@ class CC_DLL Renderer
     // the TriBatches
     TriBatchToDraw* _triBatchesToDraw;
 
+    // used as offset in in the process of concatenating triangles for batched drawing
     int _filledVertex;
     int _filledIndex;
+    // used for the flushing logic if vertex number is high and flushing is needed
+    int _filledVertex2;
+    int _filledIndex2;
```
gestern, line 266: According to [General Naming Rules within the Engine](https://github.com/cocos2d/cocos2d-x/wiki/Cpp--Coding-Style#general-naming-rules)

> Function names, variable names, and filenames should be descriptive; eschew abbreviation.
> Give as descriptive a name as possible, within reason. Do not worry about saving horizontal space as it is far more important to make your code immediately understandable by a new reader. Do not use abbreviations that are ambiguous or unfamiliar to readers outside your project, and do not abbreviate by deleting letters within a word.

`_filledVertex2` and `_filledIndex2` implies that there are / there will be other variations (ex: `_filledVertex3`, `_filledVertex4` ... etc), that these are either elements of an array or a group

gestern (commented):
Unacceptable for several reasons.

Dimon4eg:
/home/travis/build/cocos2d/cocos2d-x/cocos/renderer/CCRenderer.cpp:41:37: fatal error: renderer/ccGLStateCache.h: No such file or directory
 #include "renderer/ccGLStateCache.h"
                                     ^
The ccGLStateCache.h was removed here https://github.com/cocos2d/cocos2d-x/pull/19013

cocos/renderer/CCRenderer.cpp
BobyIlea: @gestern I know, this will be removed. It was just added for some tests I did, it obviously should not be merged.

cocos/renderer/CCRenderer.h
BobyIlea: @gestern Yes, will work on renaming variables after. For now I am only interested in you opinion on the overall solution. This solution does fix the crashes but I'm not sure if it can introduce other issues that I missed.

BobyIlea:
> The ccGLStateCache.h was removed here #19013

@Dimon4eg Ok, thanks for the info.

Issue #19087, opened 2018-10-03
ListView create a huge number of draw calls for overlapped objects


gestern:
- cocos2d-x version: **ALL**
- devices test on: **ALL**

**Steps to Reproduce:**

1. It is enough to create a more or less expensive content (it is what will later be inserted as a function parameter of `ui::ListView::pushBackCustomItem()` or `ui::ListView::pushBackDefaultItem()`) in the `ui::ListView`.
2. `Render` spends GPU time on `ListView` content objects (its childs) that are outside the render view window and should not be drawn logically.

The existence of this problem has been known for quite some time. For example: 

1. [How to optimize ListView](https://discuss.cocos2d-x.org/t/how-to-optimize-listview/29777)
2. [ccui.ListView terrible preformence on web](https://discuss.cocos2d-x.org/t/ccui-listview-terrible-preformence-on-web/37042)
3. [Android rendering performance issue](https://github.com/cocos2d/cocos2d-x/issues/15652#issuecomment-280248792)

> **Important note**
> The problem is that absolutely all the classes inherited from `ScrollView` suffer from this same “disease”:
> 1. `ScrollView`;
> 2. `ListView`;
> 3. `PageView`.

**The good news**
The good news is that I found a solution to this performance problem. The main idea is to proactively hide the child nodes that currently cannot be displayed to the user (all those who currently clipped out by `ListView` viewport ). The basic idea of applying a function when displaying components each time when the `ListView` content was changed, the common example here:
```cpp
void ListView::paintItems() {
    
    ssize_t lowerBound = _items.size() - 1;
    ssize_t upperBound = 0;
    
    if(_direction != ScrollView::Direction::VERTICAL) {
        // all directions possible except VERTICAL
        lowerBound = getIndex(getRightmostItemInCurrentView());
        upperBound = getIndex(getLeftmostItemInCurrentView());
    }
    
    if (_direction != ScrollView::Direction::HORIZONTAL) {
        // all directions possible except HORIZONTAL
        lowerBound = std::min(getIndex(getTopmostItemInCurrentView()), lowerBound);
        upperBound = std::max(getIndex(getBottommostItemInCurrentView()), upperBound);
    }
    
    for(size_t i = 0 ; i < _items.size() ; i++) {
        _items.at(i)->setVisible(lowerBound <= i && i <= upperBound);
    }
}
```
This function reduces the number of  GPU draw calls by using within `ListView` somewhere in one and a half to two times and can increase productivity by several times

[...]

PR #19088, opened 2018-10-04
Fix issues with ListView low performance 

gestern:
This function reduces the number of GPU draw calls by using within `ListView` somewhere in one and a half to two times and can increase productivity by several times. See here #19087 for more details.


gestern:
@minggo Could you please review the bugfix and express an opinion?

drelaptop:
### ListViewTest case1 failed

- when I click the space between item6 and item7 and drap to left to right times, item7 disappear. 
- works after revert commit [Fixes performance issue within ListView class](https://github.com/cocos2d/cocos2d-x/pull/19088/commits/38e5826f575aa91fdc73c217f1767a895c8a74c0)

so I surmise bug existed in this PR,  please test and check @gestern 

listview-error

gestern:
The bug is that the test itself is incorrect, it is specifically wrong to run the `update` (with possible racing) function and manually check the status of protected child controls outbound of `ListView`. But even for this completely wrong case, a solution was found in the commit that occurred later:
https://github.com/gestern/cocos2d-x/blob/8f42f1bea00b8f5c184e9174e318a156c74c4b66/cocos/ui/UIListView.cpp#L917-L920
But does it really make any sense if the code review happens once a week and a lot of changes have been added? This fix really reduced the number of primitives on complex components of `ListView` (for example on long leaderboards) by 200-250 draw calls. But does it matter on the wrong test with one button and one label within it?

Issue #19080, opened 2018-09-29
Stuttering movement and black screen on macOS 10.14 & Xcode 10


drelaptop:
- cocos2d-x version: 3.17

https://discuss.cocos2d-x.org/t/stuttering-movement-and-black-screen-on-macos-10-14-mojave-and-cocos2d-x-3-17/43860

https://discuss.cocos2d-x.org/t/xcode-10-and-clion-black-screen/43780

I think newer issues on Xcode 10 or macOS 10.14 should be solved before next release.

Anywhere-Music-Player:
Interesting, same issue happened for me for cocos2d-iphone, because I'm using SpriteBuilderX.
When launching it - no OpenGL view seen, nothing drawn.

I have no idea why it's happening. Tried to change OpenGL view settings, but nothing helps..

Anywhere-Music-Player:
@drelaptop @minggo I think I found whats the problem https://github.com/glfw/glfw/issues/1334 
also https://github.com/hajimehoshi/ebiten/issues/691 looking this cause it. However, I'm not sure.

stevetranby:
I think I was able to hack around this issue following other's advice that the context needs to be updated. If you haven't already got it working, let me know if this works for you and if you have any thoughts.
_[MacMini, Mojave 10.14.0, XCode 10]_

CCApplication-mac.mm
```
int Application::run() 
{
//...
    unsigned int ctx_updated_count = 0;
//...
    while (!glview->windowShouldClose())
    {
        //...
        // STEVE: look into doing this outside loop to get rid of condition test per frame
        if(ctx_updated_count < 2) {
            ctx_updated_count++;
            NSOpenGLContext* ctx = (NSOpenGLContext*)glview->getNSGLContext();
            [ctx update];
        }

        director->mainLoop();
        glview->pollEvents();
        //...
    }
```

CCGLViewImpl-Desktop.h
```
//...approx. Line 128
#if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
    id getCocoaWindow() override { return glfwGetCocoaWindow(_mainWindow); }
    id getNSGLContext() override { return glfwGetNSGLContext(_mainWindow); } // STEVE: added
#endif // #if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
//...
```

CCGLView.h
```
//...approx. Line 428
#if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
[...]

PR #19090, opened 2018-10-07
a workaround to fix black screen on macOS 10.14

drelaptop:
as those discuss at https://github.com/cocos2d/cocos2d-x/issues/19080 , it seems like a bug on macOS 10.14 & Xcode 10.

this PR is a hack solution supplied by @stevetranby , thanks. maybe we don't need to merge PR if Apple fix this bug soon.

### DIFF
--- a/cocos/platform/mac/CCApplication-mac.mm
+++ b/cocos/platform/mac/CCApplication-mac.mm
@@ -79,11 +79,21 @@ int Application::run()
     
     // Retain glview to avoid glview being released in the while loop
     glview->retain();
-    
+
+    unsigned int ctx_updated_count = 0;
+
     while (!glview->windowShouldClose())
     {
         lastTime = getCurrentMillSecond();
-        
+
+        // hack to fix issue #19080, black screen on macOS 10.14
+        // stevetranby: look into doing this outside loop to get rid of condition test per frame
+        if(ctx_updated_count < 2) {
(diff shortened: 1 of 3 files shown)


cocos/platform/mac/CCApplication-mac.mm
```diff
@@ -79,11 +79,21 @@ static long getCurrentMillSecond()
...
-        
+
+        // hack to fix issue #19080, black screen on macOS 10.14
+        // stevetranby: look into doing this outside loop to get rid of condition test per frame
+        if(ctx_updated_count < 2) {
+            ctx_updated_count++;
+            NSOpenGLContext* ctx = (NSOpenGLContext*)glview->getNSGLContext();
+            [ctx update];
+        }
+
```
minggo, line 96: It is better that these codes only run on `macOS 10.14 + Xcode 10`.

drelaptop:
Not merge, because https://github.com/cocos2d/cocos2d-x/issues/19080#issuecomment-435314956

cocos/platform/mac/CCApplication-mac.mm
drelaptop: this code block be executed only in the first time of main loop, for a fixed little time, not executed every frame. and add version check will make the main loop code looks more complex , so I think it's ok without version check.

PR #19091, opened 2018-10-08
ListvView virtual representation bugfix

gestern:
Sometimes there is a need to create a virtual list view. Suppose you need to create a list of 1000 objects, but at the same time there will be only 10. This fix adds the ability to support such lists.


gestern:
One of the tests was removed, since the energy that was spent to fix the issues in it is commensurate with the energy that needs to spend by Elon Musk in order to regain its reputation.

Issue #18698, opened 2018-02-26
calling label->getFontDefinition()._fontName does not return correct font name


xerosugar:
- cocos2d-x version: 3.16
- devices test on: windows
- developing environments
MS Visual Studio Community 2013

Steps to Reproduce:
Label* txt = Label::createWithTTF("some text", "fonts/KomikaSlick.ttf", 12, Size::ZERO, TextHAlignment::LEFT);
auto name = txt->getFontDefinition()._fontName;

// name will always be 'Helvetica'

PR #19095, opened 2018-10-08
CCLabel: fix fontName attribute 

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/18698

Accquire font name through `FontAtlas::getFontName()`

### DIFF
--- a/cocos/2d/CCFontAtlas.cpp
+++ b/cocos/2d/CCFontAtlas.cpp
@@ -485,7 +485,7 @@ void FontAtlas::setLineHeight(float newHeight)
 std::string FontAtlas::getFontName() const
 {
     std::string fontName = _fontFreeType ? _fontFreeType->getFontName() : "";
-    if(fontName.empty()) return "";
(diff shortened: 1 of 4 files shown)


cocos/2d/CCFontAtlas.cpp
```diff
@@ -477,11 +477,21 @@ Texture2D* FontAtlas::getTexture(int slot)
...
     _lineHeight = newHeight;
 }
 
+std::string FontAtlas::getFontName() const
+{
+    std::string fontName = _fontFreeType ? _fontFreeType->getFontName() : "";
+    auto idx = fontName.rfind("/");
+    if (idx != std::string::npos) { return fontName.substr(idx + 1); }
+    idx = fontName.rfind("\\");
+    if (idx != std::string::npos) { return fontName.substr(idx + 1); }
```
minggo, line 492: if fontName is empty, then these codes are not needed.

minggo (approved):

PR #19107, opened 2018-10-11
Fix bug: Not support sprite frame when ignoreContentAdaptWithSize is false

chongchaoyu:


cocos/ui/UIRichText.h
```diff
@@ -565,7 +565,7 @@ class CC_GUI_DLL RichText : public Widget
                             const Color3B& outlineColor = Color3B::WHITE, int outlineSize = -1,
                             const Color3B& shadowColor = Color3B::BLACK, const Size& shadowOffset = Size(2.0, -2.0), int shadowBlurRadius = 0,
                             const Color3B& glowColor = Color3B::WHITE);
-    void handleImageRenderer(const std::string& filePath, const Color3B& color, GLubyte opacity, int width, int height, const std::string& url);
+    void handleImageRenderer(const std::string& filePath, Widget::TextureResType textureType, const Color3B& color, GLubyte opacity, int width, int height, const std::string& url);
```
PatriceJiang, line 568: It is better to add a new API, rather than change the current one.

chongchaoyu: Thanks, do you mean to provide a new api for using sprite frames?

PatriceJiang: yes.

chongchaoyu: @PatriceJiang
And the current one is a little weird, developer can't use sprite frames when they want to set the content size, they have to use a local sprite which is the same as the one in the plist. In a project, using the same sprite with different size at different places is very normal, do you think so? :)

PatriceJiang: > same sprite with different size 

These sprites do share same `Texture`  underlying, but I think they should not be "the same sprite". 
A `Sprite` holds partial render info of a `Texture` instance. Since the position/size differs, they should be separated.

chongchaoyu: The background is: language is javescript, using createWithXML to create richtext.
The case is that:
There is a file, "res/star.png". I want to create two richtexts using the "star" loading from the file, one richtext's _ignoreSize is true, and the other has to be false, then I can't put "res/star.png" into plist, because when _ignoreSize is true, RichText can't load image from plist. 
In my project, almost all of the pngs has been put into plist due to resource management and performance.

Am I clear?

PatriceJiang: Please open a new issue and provide a description there

chongchaoyu: OK, the new issue is #19114

PatriceJiang (approved):

Issue #11240, opened 2015-03-31
3.4 PhysicsJointMotor setRate and getRate Bad Access Crash


jimrange:
If setRate(50) or getRate are called on a PhysicsJointMotor after it is created (in the same code block that it was created in), then a bad access crash occurs on the iPhone 6 simulator and iPhone 5 device running iOS 8.2. 

If getRate or setRate are called later in an update / tick method, then no crash occurs.

This code below can be added to the case 10 in the PhysicsTest.cpp class to 

// This is existing code in the PhysicsTest.cpp test class.
` PhysicsJointMotor\* joint = PhysicsJointMotor::construct(sp1->getPhysicsBody(), sp2->getPhysicsBody(), (float)M_PI_2 \* 5); 
 _scene->getPhysicsWorld()->addJoint(joint); 

// This is new code to reproduce the crash, added below the above code in PhysicsTest.cpp
 auto rate = joint->getRate(); // getRate() will cause a bad access crash
 rate = rate + 1;
 joint->setRate(rate); // setRate(rate) will cause a bad access crash
`

jimrange:
Please consider adding tests for the setters and getters of the physics constraints.
This issue exists in the PhysicsJointDistance, PhysicsJointLimit, PhysicsJointSpring, PhysicsJointGroove and might exist in other joints too.
Maybe the same pattern that causes the issue exists in other joints?

jimrange:
PhysicsTest.cpp Version 3.6 

Line 697:
    PhysicsJointDistance\* joint = PhysicsJointDistance::construct(sp1->getPhysicsBody(), sp2->getPhysicsBody(), Point::ZERO, Point::ZERO);
    getPhysicsWorld()->addJoint(joint);
    **joint->setDistance(100); // This will cause a crash.**

Line 710:
    PhysicsJointLimit\* joint = PhysicsJointLimit::construct(sp1->getPhysicsBody(), sp2->getPhysicsBody(), Point::ZERO, Point::ZERO, 30.0f, 60.0f);
    getPhysicsWorld()->addJoint(joint);
    **joint->setMax(100); // This will cause a crash.**

Line 724: 
    PhysicsJointSpring\* joint = PhysicsJointSpring::construct(sp1->getPhysicsBody(), sp2->getPhysicsBody(), Point::ZERO, Point::ZERO, 500.0f, 0.3f);
    getPhysicsWorld()->addJoint(joint);
[...]

PR #19108, opened 2018-10-11
cache PhysicsJoint* attributes before PhysicsWorld update

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/11240

when invoking `PhysicsJoint*::set*(...)` before `world->updateJoints()`, thread abort.
to fix this:
1. store value in `_writeCache`
2. apply the change in `PhysicsWorld::updateJoints`

### DIFF
--- a/cocos/physics/CCPhysicsJoint.cpp
+++ b/cocos/physics/CCPhysicsJoint.cpp
@@ -34,6 +34,70 @@
 
 NS_CC_BEGIN
 
+template<typename T>
+class Optional {
+public:
+    Optional() {}
+    Optional(T d) :_isSet(true), _data(d) {}
+    Optional(const Optional &t) : _isSet(t._isSet), _data(t._data) {}
+    bool isNull() const { return !_isSet; }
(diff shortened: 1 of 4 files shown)


cocos/physics/CCPhysicsJoint.cpp
```diff
@@ -34,6 +34,83 @@
...
+#define UNLIKELY(x) (__builtin_expect((x), 0))
+#else
+#define LIKELY(x)   (x)
+#define UNLIKELY(x) (x)
+#endif
+
+
+#define CC_PJOINT_CACHE_READ(field)         \
+    do {                                    \
+    if(UNLIKELY(_initDirty))                \
```
minggo, line 88: _initDirty is initialized to `true` at constructor, why it is unlikely?

PatriceJiang: It will remain false after `initJoint()` is invoked. And I think it should be predicted to be `false` for the most time.

minggo (approved):

PR #19111, opened 2018-10-12
[travis] curl add retry options

PatriceJiang:
changes list

1. `curl` with `--retry`
2. retry download zip file in `download_deps.py`
3. use `git_retry.py` from chromium instead of `git`
4. add retry function in bash & powershell script

### DIFF
--- a/download-deps.py
+++ b/download-deps.py
@@ -252,7 +252,7 @@ class CocosZipInstaller(object):
                     print("==> Download file retry " + str(times_count))
                 self.download_file()
                 return
-            except urllib2.URLError as err:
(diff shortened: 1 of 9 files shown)


crazyhappygame:
@PatriceJiang most of failed builds are related to downloading dependencies by grade. I propose to switch latest Android build tools 28 and Android grade plugin 3.2  and grade 4.6

PatriceJiang:
yes, it is true. I have met most of these issues before, and I only improve explicit downloading code here.

minggo (approved):

PR #19112, opened 2018-10-12
Fix bug: PremultipliedAlpha is not working for png

chongchaoyu:


cocos/platform/CCImage.cpp
```diff
@@ -1149,6 +1149,11 @@ bool Image::initWithPngData(const unsigned char * data, ssize_t dataLen)
         {
             premultipliedAlpha();
         }
+        else
+        {
+            _hasPremultipliedAlpha = true;
```
drelaptop, line 1154: Why enable `premultipliedAlpha` even png format isn't RGBA8888, any errors? maybe you use `CCImage` in wrong way.

drelaptop (changes requested):

cocos/platform/CCImage.cpp
chongchaoyu: @drelaptop 
oh, yes, if png format is not RGBA8888, premultipliedAlpha should not be true.

But, there is a case, I want to use PMA pngs to improve performance and prevent blend artifacts, so I use `Image::setPNGPremultipliedAlphaEnabled(false)`, but it doesn't work, because `CCImage.hasPremultipliedAlpha()` will return false at `Texture2D::initWithImage` if you don't set premultipliedAlpha to true when png format is RGBA8888.

I will resubmit a PR, the code is like the followings:

chongchaoyu: @drelaptop The new PR is #19138, you can check it.

chongchaoyu:
New PR is #19138.

cocos/platform/CCImage.cpp
drelaptop: sorry, you hope `CCImage.hasPremultipliedAlpha()` return `true`, but you did `setPNGPremultipliedAlphaEnabled(false)` before. that is confusing.

> no need to new a PR for same bug fix, you can commit to same branch, then PR will auto update.

chongchaoyu: @drelaptop `setPNGPremultipliedAlphaEnabled(false)` changes `PNG_PREMULTIPLIED_ALPHA_ENABLED` to false, it tells cocos2dx that pngs have been PMAed by developers, so cocos2dx needn't to do `premultipliedAlpha`, `premultipliedAlpha` takes about 40ms for 2048*2048 pngs.

chongchaoyu: @drelaptop I read `CCImage.cpp`, and `setPNGPremultipliedAlphaEnabled(false)` maybe the only way to use PMAed pngs. 
Doesn't cocos2dx support  PMAed pngs?

PR #19113, opened 2018-10-12
Update gradle 4.6, sdk28, android gradle  plugin 3.1.0

crazyhappygame:


drelaptop:
I think it's time to update mini SDK version from 14 to 16, for latest published NDK 18 have removed the supported on android-14, 15.

```
PROP_MIN_SDK_VERSION=14
```

crazyhappygame:
@drelaptop could you help me to fix CI build? There is problem with proguard.

drelaptop:
ok, I will see that, that's duplicate jar issue.

```
The same input jar [C:\projects\android-sdk\platforms\android-28\android.jar] is specified twice.
```

Edit:

I can't reproduce `proguard ` issues when build on my mac

drelaptop:
when use gradle 4.6 & gradle android plugin 3.2, __assets__ didn't been merged into apk, but plugin 3.1 works. it's possiable related a Android Studio know-issues below

https://developer.android.com/studio/known-issues

> Modifying variant outputs at build time may not work

> That's because variant-specific tasks are no longer created during the configuration stage. This results in the plugin not knowing all of its outputs up front, but it also means faster configuration times.

crazyhappygame:
@drelaptop Could you merge it?

PR #19117, opened 2018-10-15
Replace binary_function for C++17 Support

JasonBaby87:
I've replace binary_function in CCString.h with the function equivalent for C++17 support.

I have only tested with Visual Studio with C++14 and C++17, though...


minggo:
It is deprecated for a long time, i'd suggest not using and updating it.

JasonBaby87:
We're not using it though... It's just that it won't compile in C++17 if I don't change it.

minggo:
Got it, thanks.

PR #19126, opened 2018-10-18
Fix iOS project warnings

pyrosphere:
**Changes:**
- Update `cocos2d_libs` Xcode project and turn on recommended warnings
- Fix `int` string formatting using "%zd" instead of "%d"
- Fix loop over reference to avoid object copy


drelaptop (approved):

minggo:
thanks

PR #19129, opened 2018-10-20
Add `GLView::setCursor` for desktop platforms

pyrosphere:


drelaptop:
could you add a test case into `cpp-tests` for this feature? @pyrosphere

pyrosphere:
> could you add a test case into `cpp-tests` for this feature? @pyrosphere

Sure, I've added a test in MouseTest that toggles between a custom pointer, a custom crosshair and the default cursor.

drelaptop (approved):

PR #19132, opened 2018-10-24
iPhone X home indicator handling

mars3142:
Don’t hide the home indicator, instead „disable“ it. So the user has to double tap it to use it. It’s tinted to be nearly invisible over the content.


stevetranby:
Curious why this wasn't included as part of iPhoneX support commit.

mars3142:
Maybe **preferredScreenEdgesDeferringSystemGestures** didn't sounds like it's useful for the iPhone X, so they used **prefersHomeIndicatorAutoHidden** but this isn't working well for games.

drelaptop (approved):

PR #19139, opened 2018-11-02
modern cmake, compile definitions improvement

### LINKED ISSUE #18914: RenderTexture depend on Chipmunk
- cocos2d-x version: git
- devices test on: XUbuntu 18.04, Android 7.1 (LineageOS)
- developing environments
   - NDK version: r17
   - Xcode version: ---
   - VS version: ---
   - browser type and version: ---

Steps to Reproduce:

1. Build with `USE_CHIPMUNK=0`
2. Try to use `RenderTexture::getSprite()`, it will return invalid pointer (not `nullptr`, just invalid).
3. Access that pointer and get happy SEGFAULT.

drelaptop:
fix https://github.com/cocos2d/cocos2d-x/issues/18914


crazyhappygame:
@drelaptop I propose to remove all USE_* variables and option from cmake files. 
Configuration is in below file:
https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/base/ccConfig.h
I think that the simplest will be when user will directly update/set c++ defines. Below some options:
1. Update ccConfig.h directly
2. Before include ccConfig.h (or any other cocos file) include file with set #defines
3. Directly set c++ define through cmake
4. Create root CMakeLists.txt where before add_subirectory(cocos) all defines are set

cmake/Modules/CocosSelectModule.cmake
```diff
@@ -1,21 +1,9 @@
...
-option(USE_CHIPMUNK "Use chipmunk for physics library" ON)
-option(USE_BOX2D "Use box2d for physics library" OFF)
-option(USE_BULLET "Use bullet for physics3d library" ON)
-option(USE_RECAST "Use Recast for navigation mesh" ON)
-option(USE_WEBP "Use WebP codec" ON)
-option(USE_PNG "Use PNG codec" ON)
-option(USE_TIFF "Use TIFF codec" ON)
-option(USE_JPEG "Use JPEG codec" ON)
-option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 option(DEBUG_MODE "Debug or Release?" ON)
```
crazyhappygame: We can do that in next PR but.... I propose to remove all options  from this file. E.g. to build only some project (with dependencies) user my run:
cmake --build . --target my_project

cmake/Modules/CocosConfigDefine.cmake
```diff
@@ -8,45 +7,77 @@
...
  endif()
 
- if(CMAKE_GENERATOR STREQUAL Xcode)
-     set(XCODE TRUE)
- elseif(CMAKE_GENERATOR MATCHES Visual)
-     set(VS TRUE)
- endif()
- #Debug is default value
+ # build mode, Debug is default value
  if(NOT CMAKE_BUILD_TYPE)
```
crazyhappygame, line 40: That will not work for xcode and visual studio generators.
cmake-generator-expressions should be used everywhere when if(CMAKE_BUILD_TYPE) is used

https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

crazyhappygame (approved):
OK. 
Please create new issue for cmake-generator-expressions and build configuration.

cmake/Modules/CocosConfigDefine.cmake
```diff
@@ -0,0 +1,185 @@
...
+                PUBLIC _USE3DDLL
+            )
+        else()
+            target_compile_definitions(${target} PUBLIC CC_STATIC)
+        endif()
+    endif()
+endfunction()
+
+ # Set compiler options
+ function(use_cocos2dx_compile_options target)
```
crazyhappygame, line 165: I think that this function should be removed. Proper default values are set by cmake (e.g. -g for Debug, RelWithDebInfo). 
With those settings user can not change it.

drelaptop: After the CI passed, I could try to disable this function to see whether default values works

cmake/Modules/CocosConfigDefine.cmake
drelaptop: Right, generator expressions should be used everywhere related with build-type. I can improve this point in another PR. 

For now, https://github.com/cocos2d/cocos2d-x/pull/19139/commits/538a77787adcbc9cd80ea3052955c19f75dd180c force set single build type when generate Xcode or VS project, so user can't switch build type in IDE. no error will occurs.

cmake/Modules/CocosConfigDefine.cmake
drelaptop: compile options have been simplified, I don't think we can remove this function, the default values possible works, but may not proper for cocos2d-x project. 

when I remove this function, and then generate VS project to test, I found `/MP` isn't been set, so that the compile process will works in single thread, spend a long time to compile. @crazyhappygame

PR #19143, opened 2018-11-06
removed extra semicolon after function definition

tarptaeya:


templates/cpp-template-default/Classes/HelloWorldScene.h
```diff
@@ -38,7 +38,7 @@ class HelloWorld : public cocos2d::Scene
     void menuCloseCallback(cocos2d::Ref* pSender);
     
     // implement the "static create()" method manually
-    CREATE_FUNC(HelloWorld);
```
jcs090218, line 41: I thought the semicolon is for the better readability?

jcs090218 (commented):
Double checking the PR purpose.

templates/cpp-template-default/Classes/HelloWorldScene.h
tarptaeya: But it gives warning in editor (QtCreator) and thus have to manually remove it.

jcs090218: Well, good to hear that. Then now this is the case I would say.

stevetranby:
Interesting, I've never had any warnings on other IDE/compilers. 
It is a valid semicolon (for >= c++11).

It makes sense, pedantically, to remove it as it's unnecessary.

For some reason with c++ I do prefer the excess `;` though definitely mostly out of consistency/habit since `CREATE_FUNC` feels like a declaration and not a definition. I don't otherwise put a `;` after function bodies, unless they're single MACRO statements where it just feels weird without, even if correct. Heh.

jcs090218:
@Tarptaeya I think you might need to provide a bit more info in order to reproduce the warning?

tarptaeya:
@jcs090218 Just open the newly created project in the QtCreator, there it shows
```
Parser Issue
Warning: Extra ';' after member function definition
(-Wextra-semi)
```

minggo:
extra `;` make it more readable, i think you should modify your IDE to remove the warning.

Dimon4eg:
@minggo maybe just delete semicolon from macro?

minggo:
The macro doesn't have `;`, it ends with `}`.

Dimon4eg:
Oh sorry

PR #19145, opened 2018-11-06
modern cmake, improve cmake_compiler_flags

crazyhappygame:


crazyhappygame:

cmake/Modules/CocosConfigDefine.cmake
```diff
@@ -166,19 +166,6 @@ endfunction()
     if(MSVC)
         target_compile_options(${target}
             PUBLIC /MP
-            PUBLIC /Z7
```
drelaptop, line 169: the reason to add `/Z7` is `too large PDB file` occurs to me when compile `cpp-tests` project

cmake/Modules/CocosConfigDefine.cmake
```diff
@@ -166,19 +166,6 @@ endfunction()
...
-            PUBLIC /MD$<$<CONFIG:Debug>:d>
         )
-    else()
-        target_compile_options(${target} 
-            PUBLIC -Wall
```
drelaptop, line 174: `-Wall` is one of the the defaults values, too?

cocos/Android.mk
```diff
@@ -284,7 +284,7 @@ LOCAL_WHOLE_STATIC_LIBRARIES += cpufeatures
 
 # define the macro to compile through support/zip_support/ioapi.c
 LOCAL_CFLAGS   :=  -DUSE_FILE32API
-LOCAL_CFLAGS   +=  -fexceptions
+LOCAL_CFLAGS   +=  -fexceptions -fsigned-char
```
drelaptop: why add `-fsigned-char` into `Android.mk`, it already been added into `Application.mk`

https://github.com/cocos2d/cocos2d-x/blob/e5b77ad9bfe3356362ff310f33b707dddb00ea8f/templates/cpp-template-default/proj.android/app/jni/Application.mk#L3

cmake/Modules/CocosConfigDefine.cmake
crazyhappygame: /Z7 it is old debug format.
All CI tests passed

cmake/Modules/CocosConfigDefine.cmake
crazyhappygame: In modern cmake flags should not be set inside library. Please see for example:
https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#dont-abuse-usage-requirements

We may add this to root folder but it should not be set here (and definitely not as PUBLIC).

cocos/Android.mk
crazyhappygame: Thank I will remove it

cmake/Modules/CocosConfigDefine.cmake
stevetranby: Yeah this has been discussed where compiling would fail without `/Z7`, but it may not necessary for VS2017+ due to the improvements in the debugging database system(s). I'm not sure the best course of action here though, but maybe if you can add that option only to cpp-tests then that would be good enough since others can choose to modify their own project target options?

crazyhappygame: @stevetranby idea of this change is that common cmake configuration does not force settings on user. If user really want to change something it possible from his code without changing cocos code. It is bad cmake practice to set compiler flags https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#get-your-hands-off-cmake_cxx_flags 

About visual studio debug database (pdb files): from my experience it does not matter with or without /Z7 pdb become broken from time to time ;)

stevetranby: Sounds good. Makes sense to give control to user. 

Maybe best to just document in cpp-tests that one can try no debug (??option??) or light debug (/Z7) if they have issues building.

I am not sure if the new templated CCActions helped or hurt in that regard, but I remember the "PDB too large" issue for me was related to the large amount of code and derived classes for all the actions. 

(It would be great to see the action system refactored into a "few lines of code" generic approach by passing in easing function instead of having the large hierarchy of derived classes - even if now templated).

I digress :-]

PR #19146, opened 2018-11-07
modern cmake, cmake_cleanup_remove_unused_variables

crazyhappygame:


cmake/Modules/CocosBuildSet.cmake
```diff
@@ -1,11 +1,3 @@
-# Disable in-source builds to prevent source tree corruption.
-if(" ${CMAKE_SOURCE_DIR}" STREQUAL " ${CMAKE_BINARY_DIR}")
```
drelaptop, line 2: suggest to keep this check, it will prevent In-source builds

or we can select to add a options `set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)`

> https://stackoverflow.com/questions/1208681/with-cmake-how-would-you-disable-in-source-builds

crazyhappygame: I think that we should not force on user that. 
I propose to keep cmake files as simple as possible.

crazyhappygame:

cmake/Modules/CocosBuildSet.cmake
drelaptop: all right, I will add `set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)` later into only `<cocos2d-x-root>/CMakeLists.txt`, not include tests project and template project.

drelaptop:
thanks for this simplify

PR #19149, opened 2018-11-07
 correct lua template cmake build

drelaptop:
__Lua template__ is so special, it depend on simulator library. so that user can switch windows size by View menu

- add `cmake .. -GXcode` to build, this project works well
- build by generated `makefile` don't works, for `*.xib` issues

lua-template


drelaptop:
but on win32 even with depend on simulator library, can't switch windows size, and need set `workdir`. 

need to research this issue later.

lua-template-win

Issue #18744, opened 2018-03-22
Repeat CallFunc for some times, but never stop


drelaptop:
- cocos2d-x version: latest in GitHub

```C++
runAction(Repeat::create(CallFunc::create([this]() {
    CCLOG("hey");
}), 10));
```
isn’t this supposed to write out “hey” 10 times? it never stops…

Forum thread: http://discuss.cocos2d-x.org/t/repeat-action-what-am-i-doing-wrong/41990

drelaptop:
Adding `_total++;` after the `if` block will solve this question, it's a bug for `Repeat` `ActionInstant `

https://github.com/cocos2d/cocos2d-x/blob/6317e35f01383943202517d639456299e9eb1bec/cocos/2d/CCActionInterval.cpp#L541

final code block
```
    else
    {
        if (!(sendUpdateEventToScript(fmodf(dt * _times,1.0f), _innerAction)))
            _innerAction->update(fmodf(dt * _times,1.0f));

        _total++;
    }
```
 what's the meaning to repeat a Instant Action with no interval?

PR #19158, opened 2018-11-09
Action::Repeat fix NaN error 

PatriceJiang:
https://github.com/cocos2d/cocos2d-x/blob/bb32eee94438f92b61799781211ddd3f5afd7736/cocos/2d/CCActionInterval.cpp#L481-L487

`_duration == 0.0f`  will lead `_nextDt` to be `NaN`


PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/18744

drelaptop (approved):

Issue #18728, opened 2018-03-13
Outdated version of uthash


slackmoehrle:
- cocos2d-x version: master +

Cocos2d-x is using an out-dated version of [uthash](https://github.com/troydhanson/uthash). Some users are wondering if this is causing issues with `Scheduler` in 3.16+

http://discuss.cocos2d-x.org/t/potential-problem-in-cocos2d-scheduler/41839

Dimon4eg:
Probably we don't need uthash at all as we plan to switch to std::unordered_map https://github.com/cocos2d/cocos2d-x/issues/16661

Issue #16661, opened 2016-10-08
Replace c style uthash with std::unordered_map<K, V>


dumganhar:
Now we're using c++11 and have replaced some part of uthash code since v3.0,
but not all code which uses uthash was modified.
The performance of `std::unordered_map` is good enough, therefore we could safely replace all uthash code with `std::unodrdered_map<K, V>`, it will be more c++ friendly and readable.

Currently, there are serval modules are using uthash:

![uthash-modules](https://cloud.githubusercontent.com/assets/493372/19211645/07960848-8d74-11e6-884b-c7c6c8cb828e.png)

TODO LIST:
- [X] FontFNT: Fixed at https://github.com/cocos2d/cocos2d-x/pull/16662
- [ ] ActionManager
- [x] GLProgram: Just include `base/uthash.h`, but there isn't anywhere uses it. Fixed at https://github.com/cocos2d/cocos2d-x/pull/16662
- [ ] Scheduler
- [ ] cocos2d_specifics

stevetranby:
Did you write a performance test (or is there an existing one) that confirms relatively similar performance? I don't doubt it, but confirmation would be nice.

dumganhar:
@stevetranby , we tested it some years ago. Please refer to

https://docs.google.com/spreadsheets/d/1XolpgYfoWszA2rxnVRCAVS7ILAGiV049o5mpL29cwLs/edit#gid=6

cocos2d::Map<K, V> is using `std::unordered_map` while `cocos2d::Dictionary` is using UT_HASH.

Test code could be found at `Performance Test` project.

PR #19159, opened 2018-11-09
[uthash] upgrade to 2.0.2

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/18728


Dimon4eg:
Probably we don't need uthash at all as we plan to switch to std::unordered_map #16661

PatriceJiang:
@Dimon4eg you are right, I will close this PR.

Issue #18755, opened 2018-03-29
LabelTTF limit is 20,152, can we get more?


slackmoehrle:
- cocos2d-x version: `v3`

`LabelTTF` has a limit of 20,152. Some users want up to 30,000.

```cpp
void I18n_Historys::loadHistoryOnline(unsigned int h_id)
{
    _curr_history->text = "";
    for(int i=0; i<30000; i++)
    {
        _curr_history->text = _curr_history->text + "0";
    }
    if(_onHistoryCallback)
        _onHistoryCallback(true);
}
```

stevetranby:
I question whether this is available for all devices. I think the solution may need to be restrict text length to 2^16/4bytes (e.g. 16384). Not sure where the limit 20152 comes from, but maybe spaces don't affect the vert count and thus allow larger length limit?

stevetranby:
Cross-discussion:
https://github.com/cocos2d/cocos2d-x/pull/19160#issuecomment-437641290

I think that to support/fix this Issue cocos2d should handle creating additional vertex buffers in order to render beyond 10k letters. As a practical matter the only situation I see a need for this is in scrolling?

drelaptop:
similar same issue  https://github.com/cocos2d/cocos2d-x/issues/17946

drelaptop:
fix by strip the extra text, refer to https://github.com/cocos2d/cocos2d-x/pull/19174

PR #19160, opened 2018-11-09
CCTextureAtlas `_indices` overflow

PatriceJiang:
ref: https://github.com/cocos2d/cocos2d-x/issues/18755

`_indices` can exceed `2^16-1` which leads to overflow.

fix:
change type to  `GLuint`

### DIFF
--- a/cocos/renderer/CCTextureAtlas.cpp
+++ b/cocos/renderer/CCTextureAtlas.cpp
@@ -172,7 +172,7 @@ bool TextureAtlas::initWithTexture(Texture2D *texture, ssize_t capacity)
     CCASSERT(_quads == nullptr && _indices == nullptr, "_quads and _indices should be nullptr.");
 
     _quads = (V3F_C4B_T2F_Quad*)malloc( _capacity * sizeof(V3F_C4B_T2F_Quad) );
-    _indices = (GLushort *)malloc( _capacity * 6 * sizeof(GLushort) );
+    _indices = (GLuint *)malloc( _capacity * 6 * sizeof(GLuint) );
     
     if( ! ( _quads && _indices) && _capacity > 0) 
     {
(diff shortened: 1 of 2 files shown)


PatriceJiang:
when `i*4+3 > 2^16 - 1` _indices go wrong.

https://github.com/cocos2d/cocos2d-x/blob/bb32eee94438f92b61799781211ddd3f5afd7736/cocos/renderer/CCTextureAtlas.cpp#L240-L251

`_capacity` was limited to `16383`

stevetranby:
I may be wrong ... but I don't believe this PR is acceptable unless it were refactored to only apply conditionally based on capability support by extension? 

 I think to use `UINT` as index type would require availability and support of the extension `'GL_OES_element_index_uint'` and I don't believe this is required with OpenGL ES 2 (WebGL 1)??

I believe OpenGL ES 3 (and thus WebGL 2) requires supporting this extension.

https://www.khronos.org/registry/OpenGL/specs/es/2.0/es_cm_spec_2.0.pdf
> "Indexing support with ubyte and ushort indices is supported. Support for uint indices is not required
> by OpenGL ES 2.0. If an implementation supports uint indices, it will export the OES element index -
> uint extension."

Let me know if I'm mistaken.

stevetranby:
The full spec doesn't even mention the extension.
https://www.khronos.org/registry/OpenGL/specs/es/2.0/es_full_spec_2.0.pdf
> type must be one of UNSIGNED_BYTE or UNSIGNED_SHORT, indicating that
> the values in indices are indices of GL type ubyte or ushort, respectively. mode
> specifies what kind of primitives are constructed;

PatriceJiang:
@stevetranby you are correct.  I would try to fix this issue in another way.

Issue #19154, opened 2018-11-09
sth about spine anim.


zcgit:
As I read the latest version code, I found the cocos always read and decode the skeletonJsonFile & atlasFile when load the spine animation. so if an animation is very complex , it will take a lot of IO and decoding time.
 In fact, it always do 3 things when using a spine anim: 1.read skeletonJsonFile; 2.read atlasFile; 3.read the imageFile. 
 I wonder if we can separate the 3 things?  just loading the config file(skeletonJsonFile & atlasFile) when we need, and then loading a png when we really want to use the spine anim.
 In my real working situations I could load 30-50 and more spines' config files before we really use the spine anim. so it will make a good experience when you really want to use the spine which config file was loaded, because you just need to load a png file any more, and then you can also release the png and config files separately.

zcgit:
anyone check it out ?  https://github.com/cocos2d/cocos2d-x/pull/19163

gestern:
If you keep all animations in a **single atlas**, then this is not a problem at all.

PR #19163, opened 2018-11-12
add spine cache.

zcgit:
https://github.com/cocos2d/cocos2d-x/issues/19154.


zcgit:
#19163 I've implemented this issue .
### the concept:
- I've made a cache for spine when using.
- when you removed a spine node, its skeletonData was still retained in the cache.
- when you need create the spine again, it will just be created from the cache.
- maybe you really want to release the spine cache, you can just dispatch an event with name "REMOVE_UNUSED_SPINES" , it will search the ref count of spine, if the spine's count < 1, then release it from the cached spine-data.
- sometimes you can dispatch an event with name "REMOVE_UNUSED_SPINE_CONFIG_DATA" to release the json&allas cache-file after you hot updated the spine resources.

### the usage is very simple:
- use the spine like yesterday before, you don't need to do anything more.
- dispatch an event with name "REMOVE_UNUSED_SPINES" if you want to release more memory(the cached spine data in the engine).
- (maybe you need to dispatch an event with name "REMOVE_UNUSED_SPINE_CONFIG_DATA" after you hot updated the spine resources).

alternatively, you can read the Chinese  description from my repo:[https://github.com/zcgit/cocos2dx-share](https://github.com/zcgit/cocos2dx-share)

AIGRIND-LLC:
What about the games that relies on the current behaviour?
As I can see in your description the spine data won't be released when spine node is destroyed. The developer must emit some special events to release spine data instead. So if someone "_uses the spine like yesterday before_" it will cause a memory leaks of spine data. 
Am I right?

zcgit:
> What about the games that relies on the current behaviour?
> As I can see in your description the spine data won't be released when spine node is destroyed. The developer must emit some special events to release spine data instead. So if someone "_uses the spine like yesterday before_" it will cause a memory leaks of spine data.
> Am I right?

yes, you are right, it‘s just like the textureCache, I removed a sprite node, but its texture-data was still in the textureCache.
maybe I should make a SpineDataCache separately :)

gestern:
@zcgit Do I understand correctly that you are proposing to use a single-threaded non-extensible caching mechanism based on static variables (like here: https://github.com/cocos2d/cocos2d-x/pull/19163/files#diff-ab8deb1cce5280f679f505ac0bb2cfdeR94) and you do this in the core render class `SkeletonRenderer`. And after this update, any game with proper multi-threading caching, which was based on the mechanism of inheritance (ex: `CachedSkeletonAnimation` class) from the class `SkeletonAnimation` will not be worked any more?

zcgit:
> @zcgit Do I understand correctly that you are proposing to use a single-threaded non-extensible caching mechanism based on static variables (like here: https://github.com/cocos2d/cocos2d-x/pull/19163/files#diff-ab8deb1cce5280f679f505ac0bb2cfdeR94) and you do this in the core render class `SkeletonRenderer`. And after this update, any game with proper multi-threading caching, which was based on the mechanism of inheritance (ex: `CachedSkeletonAnimation` class) from the class `SkeletonAnimation` will not be worked any more?

I think it could be worked well, you can check the class LuaSkeletonAnimation, it inherits from spine::SkeletonAnimation. but honestly, I haven't used the spine in multi-threading situation ever.

halx99:
Hi @zcgit , The cocos2d-x can't modify spine runtime due to it's license limit. you should post PR to spine runtime repo: https://github.com/EsotericSoftware/spine-runtimes

zcgit:
> Hi @zcgit , The cocos2d-x can't modify spine runtime due to it's license limit. you should post PR to spine runtime repo: https://github.com/EsotericSoftware/spine-runtimes

Thanks for your reminding.

PR #19164, opened 2018-11-12
Fix static token array in CCPUScriptTranslator.h

iDings:
The token array shouldn't be static in CCPUScriptTranslator.h, otherwise including it will generate duplicate token array, then will increase final program's size.


minggo:
it is synchronized to metal-support branch in https://github.com/cocos2d/cocos2d-x/pull/19585

PR #19165, opened 2018-11-12
[CCLabel] CCTextureAtlas `_indices` overflow, use addtional VAO

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/pull/19160

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -27,7 +27,6 @@
 #include "2d/CCLabel.h"
 
 #include <algorithm>
-#include <unordered_map>
 
 #include "2d/CCFont.h"
 #include "2d/CCFontAtlasCache.h"
@@ -47,9 +46,6 @@
 #include "2d/CCFontFNT.h"
 
 NS_CC_BEGIN
-
-using std::shared_ptr;
-using std::unordered_map;
 
 /**
  * LabelLetter used to update the quad in texture atlas without SpriteBatchNode.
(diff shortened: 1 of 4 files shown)


PatriceJiang:
@stevetranby please review if you are available.

PatriceJiang:
ref issue https://github.com/cocos2d/cocos2d-x/issues/18755

stevetranby:
I can't test this out for a few days, but at first glance it doesn't look like much, if anything was added? It seems like the "old" batch node set was replaced with a "newer" batch node set?

1. For cocos2d-x v3.x I would recommend using the suggested containers: Vector<>, Map<>, etc. While unique/shared_ptr are probably better in the long run this is not kosher cocos2d v3 code. 
2. This PR could be applied to the v4 branch, or whatever future branch/repo the Cocos team is working with.
3. Similarly I wouldn't create a custom forEach unless the standard `range for` doesn't work in a specific instance.
4. At a glance it appears that _batchNodes is there to provide for the use case where the number of unique character glyphs is larger in count and/or texture size (if the size of automatically generated texture atlases for glyphs used - unique based on font character & font size). 

Again, only at a glance (I haven't tested this), it appears that there's one real change? 
https://github.com/cocos2d/cocos2d-x/pull/19165/files#diff-8da8f4bd221ed5f05315c6d786686a51L1622

This could probably be changed to loop on the batch nodes (`for(auto& batch : batchNodes)`), but I'm not sure, and I am unfamiliar with the TTF rendering. My suggestion is to use Bitmap font if possible, then if not render long text by breaking up into a small number of lines and add them into a ListView or similar, finally consider showing a WebView for really long text (if the goal or purpose is showing a privacy policy, FAQ, or other). But I do understand if people need this to ask for it to be added into the engine supporting any length text.

As an aside, I would want to caution any changes to this class/code directly both to prevent breaking anything by accident until this is heavily tested. I'd almost rather see a LabelTTFLarge (or whatever).

For v4.x I could envision having minimum target graphics API that supports larger index values and very large vertex/index buffers. I could also envision managing the rendering at a lower-level where it'd use PrimitiveCommand directly instead of SpriteBatchNodes.

Lastly, someone else should really look at this on the actual dev team (Cocos team) who has previously worked on this code. 

2c

stevetranby:
I think my main point that may have been lost in my reply, or TL;DR, is that it appears the batch nodes are only for supporting "infinite" character glyphs, but don't have anything to do with rendering too many letter quads. I think there would need to be an additional "batch" of QuadCommands where once 2^16/4 (or whatever the max actually is) letters are added to the label an additional QuadCommand is added for the next 2^16/4 letters, and so forth.

stevetranby:
Hey @PatriceJiang - I'm not trying to give what I fear may seem like negative comments.

I'm also only one voice here and I am not part of the actual development team, so don't take anything I say as gospel (final decision). My only goal with staying active in this community is to try and keep cocos2d-x v3.x stable and **consistent**. 

Just trying to give my advice with that goal in mind.

PatriceJiang:
@stevetranby Your advices are quite helpful. I do follow your advices to update this PR.

PatriceJiang:
I closed this PR temporarily to reduce notification emails which triggered by my frequent commits. And I am sorry for my late response.

minggo:
What i think is that, LabelAtlas don't have to support infinite length. When it exceeds the length, then just cut it instead of crash, and may be print some warning message.

And i agree with @stevetranby , we should keep v3 stable, this PR may bring bugs.

PR #19153, opened 2018-11-08
Copy resources to "Resources/" on win32 like in linux configuration

crazyhappygame:
Copy resources to "Resources/" like in linux configuration

PR #19166, opened 2018-11-12
go on, Copy resources to "Resources/" on win32

drelaptop:
improve https://github.com/cocos2d/cocos2d-x/pull/19153

with `_defaultResRootPath ` changed, you will have to adapt js/lua resources copy too, or you will get file not found error. so I don't think it's elegent to do this work "Copy resources to "Resources/" on win32".

please give your options @crazyhappygame

### DIFF
--- a/cocos/2d/libcocos2d.vcxproj
+++ b/cocos/2d/libcocos2d.vcxproj
@@ -15,6 +15,7 @@
     <ProjectGuid>{98A51BA8-FC3A-415B-AC8F-8C7BD464E93E}</ProjectGuid>
     <RootNamespace>cocos2d-x.win32</RootNamespace>
     <Keyword>Win32Proj</Keyword>
+    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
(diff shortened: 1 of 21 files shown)


crazyhappygame:
You are right  js/lua *.vcxproj should be updated. Thank you!
Could you fix it in this PR? If no I can do this in 1-2 days.

I think that we should keep the same folder structure for linux and win32.

I have a plan to replace function cocos_copy_res with following implementation:
```
set (source "${CMAKE_SOURCE_DIR}/resources")
set (destination "${CMAKE_CURRENT_BINARY_DIR}/resources")
add_custom_command(
 TARGET ${PROJECT_NAME} POST_BUILD
 COMMAND ${CMAKE_COMMAND} -E create_symlink ${source} ${destination} 
 DEPENDS ${destination}
 COMMENT "symbolic link resources folder from ${source} => ${destination}"
)
```

drelaptop:
yeah, I can fix it in this PR.

> we plan to release a 3.17.1 test version to our test team in this week.

crazyhappygame:
@drelaptop thank you!

PR #18908, opened 2018-06-26
Fix iCCP profile using mogrify

v1993:
I get a lot of `libpng warning: iCCP: known incorrect sRGB profile` on tests/templates.

This PR should fix it. For more information, refer to https://stackoverflow.com/a/22747902/5697743.

PR #19083, opened 2018-09-29
SpriteFrameCache: keep plist filenames when trying to remove sprite

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/17949

- delete field `_loadedFileNames`
- connect `plist name` , `frame name`  and `SpriteFrame` together in `PlistFramesCache`.

`plist names` is kept when trying to remove any sprite.

PR #19170, opened 2018-11-14
fix cpp-tests SpritePolygon case

drelaptop:
- convert png format from "Gray" to "sRGB", or assert will trigger

```
    CCASSERT(_image->getRenderFormat()==Texture2D::PixelFormat::RGBA8888, "unsupported format, currently only supports rgba8888");
```
it was cause by #18908 wrong format conversion

- Not Finished, SpritePolygon cases will crash in others


cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -762,8 +762,12 @@ bool SpriteFrameCache::PlistFramesCache::eraseFrame(const std::string &frame)
     auto itFrame = _indexFrame2plist.find(frame);
     if (itFrame != _indexFrame2plist.end())
     {
-        _indexPlist2Frames[itFrame->second].erase(frame);   //update index plist->[frameNames]
+        auto plist = itFrame->second;
+        _indexPlist2Frames[plist].erase(frame);             //update index plist->[frameNames]
         _indexFrame2plist.erase(itFrame);                   //update index frame->plist
+        // erase plist index if all frames was erased
+        if (_indexFrame2plist.empty())
+            _indexPlist2Frames.erase(plist);
```
drelaptop, line 770: if a plist file was added, and then all frames was erased, the plist index may still existed, like
```
(lldb) print this->_spriteFramesCache
(cocos2d::SpriteFrameCache::PlistFramesCache) $4 = {
  _spriteFrames = {
    _data = size=0 {}
  }
  _indexPlist2Frames = size=1 {
    [0] = {
      __cc = {
        first = "Images/test_polygon.plist"
        second = size=0 {}
      }
      __nc = {
        first = "Images/test_polygon.plist"
        second = size=0 {}
      }
    }
  }
  _indexFrame2plist = size=0 {}
}
``` 
so that sprite frames will be prevent added when load same name plist  again

drelaptop: it should be a bug caused by https://github.com/cocos2d/cocos2d-x/pull/19083 ,please review @PatriceJiang

PatriceJiang: ok

PatriceJiang (approved):

PR #19177, opened 2018-11-16
fix crash of cclayercolor and fix some other error logic

### LINKED ISSUE #19168: record cpp-tests crash issues on latest v3
__should be solved before 3.17.1 released__ , tested use latest commit https://github.com/cocos2d/cocos2d-x/commit/ae61f8fbc6e0ca6eed57a795f6a3890dce8e3760

- crash 7th cases of __3:Action-Basic__ on macOS
1. run Node:Scene3D
2. run Action-Basic cases one by one, crash will occur on 7th cases

revert v3 to https://github.com/cocos2d/cocos2d-x/pull/19013 crash still occurs, but the commit before this PR this crash gone.

minggo:
This PR fix https://github.com/cocos2d/cocos2d-x/issues/19168 and fix some logic error in https://github.com/cocos2d/cocos2d-x/pull/19013:

* blend function setting error in script binding
* bind texture should consider alpha texture to support etc1+alpha

### DIFF
--- a/cocos/2d/CCLayer.cpp
+++ b/cocos/2d/CCLayer.cpp
@@ -620,13 +620,13 @@ void LayerColor::onDraw(const Mat4& transform, uint32_t /*flags*/)
     getGLProgram()->use();
     getGLProgram()->setUniformsForBuiltins(transform);
     
-    glEnableVertexAttribArray(GLProgram::VERTEX_ATTRIB_POSITION);
-    glEnableVertexAttribArray(GLProgram::VERTEX_ATTRIB_COLOR);
-    
     //
     // Attributes
     //
     glBindBuffer(GL_ARRAY_BUFFER, 0);
+    glEnableVertexAttribArray(GLProgram::VERTEX_ATTRIB_POSITION);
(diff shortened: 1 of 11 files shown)


cocos/2d/CCLayer.cpp
```diff
@@ -620,13 +620,13 @@ void LayerColor::onDraw(const Mat4& transform, uint32_t /*flags*/)
...
     
-    glEnableVertexAttribArray(GLProgram::VERTEX_ATTRIB_POSITION);
-    glEnableVertexAttribArray(GLProgram::VERTEX_ATTRIB_COLOR);
-    
     //
     // Attributes
     //
     glBindBuffer(GL_ARRAY_BUFFER, 0);
+    glEnableVertexAttribArray(GLProgram::VERTEX_ATTRIB_POSITION);
+    glDisableVertexAttribArray(GLProgram::VERTEX_ATTRIB_TEX_COORD);
```
minggo, line 628: Should disable VERTEX_ATTRIB_TEX_COORD since it is not used in the program, and it may be enabled in other places.

minggo: It is used in vertex shader, but have no effect. If this channel is enabled, then will cause crash.

PR #19191, opened 2018-11-20
Fix app crash in cocos2dx-JS if garbage collection runs between TouchStart and TouchEnd event.

vaibhavgohel:
Issue:
If Garbage collection of SpiderMonkey run between TouchStart and TouchEnd event application was crashing.

RCA:
If GC Runs between two touchMove Event. Touch event JSProxy were getting GarbageCollect because they were created as weak object and on next touchMove event application was crashing inside JS-Engine.

I Have added touchEvents to strong object map and have verified that it is getting deleted after TouchEnd/TouchCancel event.

### DIFF
--- a/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
+++ b/cocos/scripting/js-bindings/manual/ScriptingCore.cpp
@@ -1427,7 +1427,7 @@ bool ScriptingCore::handleTouchesEvent(void* nativeObj, cocos2d::EventTouch::Eve
 
     for (const auto& touch : touches)
     {
-        JS::RootedValue jsret(_cx, OBJECT_TO_JSVAL(jsb_get_or_create_weak_jsobject(_cx, touch, typeClassTouch)));
+        JS::RootedValue jsret(_cx, OBJECT_TO_JSVAL(jsb_ref_get_or_create_jsobject(_cx, touch, typeClassTouch, "cocos2d::Touch")));
         if (!JS_SetElement(_cx, jsretArr, count, jsret))
         {
             break;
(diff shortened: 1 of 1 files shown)


drelaptop:
LGTM, thanks for this contribution.

liuyi:
GOOD JOB! I really need this fix!!

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -1467,8 +1467,8 @@ bool ScriptingCore::handleTouchEvent(void* nativeObj, cocos2d::EventTouch::Event
         js_type_class_t *typeClassEvent = js_get_type_from_native<cocos2d::EventTouch>((cocos2d::EventTouch*)event);
 
         jsval dataVal[2];
-        dataVal[0] = OBJECT_TO_JSVAL(jsb_get_or_create_weak_jsobject(_cx, touch, typeClassTouch));
-        dataVal[1] = OBJECT_TO_JSVAL(jsb_get_or_create_weak_jsobject(_cx, event, typeClassEvent));
+        dataVal[0] = OBJECT_TO_JSVAL(jsb_ref_get_or_create_jsobject(_cx, touch, typeClassTouch, "cocos2d::Touch"));
+        dataVal[1] = OBJECT_TO_JSVAL(jsb_ref_get_or_create_jsobject(_cx, event, typeClassEvent, "cocos2d::Touch"));
```
fleon, line 1471: This should be "cocos2d::EventTouch"

drelaptop: thanks, correct it at https://github.com/cocos2d/cocos2d-x/pull/19244

PR #19200, opened 2018-11-22
[android tests] setup project name in AndroidStudio welcome page

PatriceJiang:
fix project name in the list
image


drelaptop:
as I concerned, it's better to do same on 3 template projects, you can refer to https://github.com/cocos2d/cocos2d-x/pull/19038 @PatriceJiang

PR #19204, opened 2018-11-23
Node:Particle3D abnormal texture effects

PatriceJiang:
revert change in https://github.com/cocos2d/cocos2d-x/commit/d1597a0e3e08d295fbf2a8a407660b089eb7f08d

This commit causes come crash and unexpected texture modification. 
We may try to use another tool to fix png file format later.


drelaptop:
this PR revert https://github.com/cocos2d/cocos2d-x/pull/18908 , see above for detail @v1993

v1993:
Oh, ok.

Developing SDL game now, so not going to find another way to fix it any time soon…

PR #19208, opened 2018-11-26
[cmake] fix iOS xcode property setting failed

drelaptop:
take out `search_depend_libs_recursive` to find depend libs.


drelaptop:
review if available @crazyhappygame

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -231,26 +241,18 @@ macro(cocos_pak_xcode cocos_target)
...
-            # for example, engine_lib: cocos2d link external_lib: flatbuffers
-            get_target_property(external_libs ${engine_lib} LINK_LIBRARIES)
-            foreach(external_lib ${external_libs})
-                if(TARGET ${external_lib})
-                    cocos_config_target_xcode_property(${external_lib})
-                endif()
-            endforeach()
+    set(depend_libs)
+    search_depend_libs_recursive(${cocos_app} depend_libs)
+    foreach(depend_lib ${depend_libs})
```
crazyhappygame, line 253: Do we have any dll for ios/xcode? 
I thought that all libraries are static (like for Android)

crazyhappygame:
Could you describe fixed problem?

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -75,6 +68,23 @@ function(get_target_depends_ext_dlls cocos_target all_depend_dlls_out)
             break()
         endif()
     endwhile(true)
+    set(${all_depends_out} ${all_depends_inner} PARENT_SCOPE)
+endfunction()
+
+# get `cocos_target` depend all dlls, save the result in `all_depend_dlls_out`
+function(get_target_depends_ext_dlls cocos_target all_depend_dlls_out)
```
crazyhappygame, line 75: I think that instead of all those functions and property CC_DEPEND_DLLS set for each target in external libs we should just have list of targets (curl zlib ... ) and used this list where needed

crazyhappygame: Please check
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/issues/330#issuecomment-436728995

I think that we should call function with list of targets
copy_imported_targets(app curl zlib ..

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: it's a function to set Xcode property, not dlls.

what we need is a function to find all depend target, and then set property one by one.

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: > property CC_DEPEND_DLLS set for each target in external libs 

yes, it's proper than this one.

> we should call function with list of targets

I don't think so, since the target lists have existed in the relationship of dependence, we can get them by `LINK_LIBRARIES` property recursively.

The defect is CMake didn't supply a functions to get all linked libraries, so we need `search_depend_libs_recursive `, it's a little complex. About the feature CMake should supply, existed thread at https://gitlab.kitware.com/cmake/cmake/issues/12435

drelaptop:
It fix an issue, the external libraries Xcode property didn't been set, so that compile may failed when the build target is a iOS device. @crazyhappygame 

> we set valid build arch `armv7` and `arm64` in iOS toolchain file, but didn't enable `ONLY_ACTIVE_ARCH`, so that we publish to a iOS device, for example: iPhone X. XCode will report a error we can't compile `armv7` into this device.

PR #19219, opened 2018-11-29
Revert "Remove ccglstatecache (#19013)"

drelaptop:
- To keep forward compatibility, revert this renderer logic changes

this modify will comes again with multi renderer background, for at that time we didn't need to consider compatibility strictly.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,5 @@
 cocos2d-x-3.17.1 Nov.19 2018
 
-[REFINE]        Renderer: remove ccGLStateCache
 [REFINE]        Renderer: properly reduce clear buffer times to save energy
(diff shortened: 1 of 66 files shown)


vkreal:
@drelaptop can you please elaborate more on this drastic revert?

minggo:
Because it breaks compatibility, and it may bring some bugs that are not easy to found and fix, such as this one https://github.com/cocos2d/cocos2d-x/pull/19177.

When supporting metal(may be next version), these OpenGL ES related codes will be deleted too, and that's a better time to do it.

minggo (approved):

minggo:
And we found wrong effect in 3D modules too.

vkreal:
Make sense. Thank you!

PR #19221, opened 2018-11-29
Fix animation duration calculation

zeeker999:
The duration of action will be miscalculated when float abs(float)
is unavailble, replacing it with std::abs fixes the issue.

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -89,7 +89,7 @@ void ExtraAction::step(float /*dt*/)
 bool ActionInterval::initWithDuration(float d)
 {
 
-    _duration = abs(d) <= MATH_EPSILON ? MATH_EPSILON : d;
+    _duration = std::abs(d) <= MATH_EPSILON ? MATH_EPSILON : d;
     _elapsed = 0;
     _firstTick = true;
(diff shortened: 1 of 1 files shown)


PatriceJiang:
> when float abs(float) is unavailable

Could you please provide more info about how would this occur? @LuoJinghua

LuoJinghua:
the only abs method defined int abs(int)


[1]http://man7.org/linux/man-pages/man3/abs.3.html

zeeker999:
I'm sorry, I accidently closed the pull request.

> > when float abs(float) is unavailable
> 
> Could you please provide more info about how would this occur? @LuoJinghua

assume |d| < 1.0f and float abs(float) is absent,
> _duration = abs(d) <= MATH_EPSILON ? MATH_EPSILON : d 

will evaluate to
> MATH_EPSILON

which is wrong,  it should be d instead.

There is no float abs(float) in stdlib.h[1] at least on linux.
>       #include <stdlib.h>
>
>       int abs(int j);
>

[1]: http://man7.org/linux/man-pages/man3/abs.3.html

PatriceJiang (approved):

PatriceJiang:
It dose trigger a compile warning message on Linux with `clang++`.

PR #19222, opened 2018-11-30
[cpp test]fix SequenceAction count error

PatriceJiang:


cocos/2d/CCActionInterval.cpp
```diff
@@ -119,7 +119,7 @@ void ActionInterval::step(float dt)
     if (_firstTick)
     {
         _firstTick = false;
-        _elapsed = 0;
+        _elapsed = MATH_EPSILON;
```
crazyhappygame, line 122: Could you explain this fix? What is wrong with 0?

PatriceJiang: PR https://github.com/cocos2d/cocos2d-x/pull/19158 result 
cpp-tests/3.Action tests (54,55)



test cause failure.

Issue #19187, opened 2018-11-19
Android crash on "Start AutoTest" latest 3.17.1


crazyhappygame:
- cocos2d-x version: latest  
https://ci.appveyor.com/project/minggo/cocos2d-x/builds/20399029/job/88b3dvcf49r83qh2/artifacts  (signed apk version here: https://dropfile.nl/f/jT8j)
- devices test on: HTC U11
- developing environments
   - NDK version: ndk16
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:

1. Run Start "Start AutoTest"
Bug: crash:

```
11-19 18:44:21.790  1455  1490 D cocos2d-x debug info: Wrong: a1[2]=19.112455, a2[2]=0.000000
11-19 18:44:21.790  1455  1490 D cocos2d-x debug info: Assert failed: The optimized instruction is implemented in a wrong way, please check it!
11-19 18:44:21.790  1455  1490 E cocos2d-x assert: C:\projects\cocos2d-x\tests\cpp-tests\Classes\UnitTest\UnitTest.cpp function:__checkMathUtilResult line:1448
11-19 18:44:21.790  1455  1490 D cocos2d-x debug info: Correct: a1[3]=0.000000, a2[3]=0.000000
11-19 18:44:21.791  1455  1490 F libc    : stack corruption detected (-fstack-protector)
11-19 18:44:21.857  1455  1490 W google-breakpad: Failed to generate minidump.
11-19 18:44:21.923  3852  1490 F google-breakpad: Microdump skipped (uninteresting)
11-19 18:44:21.936  1455  1490 W google-breakpad: ### ### ### ### ### ### ### ### ### ### ### ### ###
11-19 18:44:21.936  1455  1490 W google-breakpad: Chrome build fingerprint:
11-19 18:44:21.936  1455  1490 W google-breakpad: 70.0.3538.80
11-19 18:44:21.936  1455  1490 W google-breakpad: 353808052
11-19 18:44:21.936  1455  1490 W google-breakpad: ### ### ### ### ### ### ### ### ### ### ### ### ###
11-19 18:44:21.936  1455  1490 D cocos2d-x debug info: [TestController]Catch an crash event
11-19 18:44:21.979  2564  2968 D AudioFlinger: track session ID 1537 first underrun on thread ID 13
11-19 18:44:22.000  2564  2968 V AudFlgTr: underrun,  framesReady(0) < framesDesired(956), state: 6
[...]

PR #19223, opened 2018-11-30
[cpp test]Unit Tests/9 neon math crash

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19187#issuecomment-442970218

### DIFF
--- a/tests/cpp-tests/Classes/UnitTest/UnitTest.cpp
+++ b/tests/cpp-tests/Classes/UnitTest/UnitTest.cpp
@@ -1452,24 +1452,24 @@ static void __checkMathUtilResult(const char* description, const float* a1, cons
 void MathUtilTest::onEnter()
 {
     UnitTestDemo::onEnter();
-    
+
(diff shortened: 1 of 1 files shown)


Dimon4eg:
What reason of crash?
How replacing raw arrays by vector fixes the crash?

crazyhappygame:
I join the question. Could you explain what and how it was fixed?

PatriceJiang:
Sorry, I don't get the real reason for now. It may be a issue about `-O3` & `inline asm code`.

PatriceJiang:
Here is another neat solution https://github.com/cocos2d/cocos2d-x/pull/19224, I will close this PR later.

crazyhappygame:
Sorry this is not a solution. You have workaround for test but what about production code?

Could you check https://github.com/cocos2d/cocos2d-x/pull/18895 ? It could be related.

PatriceJiang:
Yes, it is just a workaround to make the test case pass, and I will try to fix this issue early next week.

PR #19224, opened 2018-11-30
[android/cpp test]Unit Test/9 neon32 release mode crash

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19187#issuecomment-442970218

### DIFF
--- a/cocos/math/MathUtilNeon.inl
+++ b/cocos/math/MathUtilNeon.inl
@@ -43,7 +43,7 @@ public:
     inline static void crossVec3(const float* v1, const float* v2, float* dst);
 };
 
-inline void MathUtilNeon::addMatrix(const float* m, float scalar, float* dst)
+inline void MathUtilNeon::addMatrix(const float* m, float scalar, float* dst) __attribute__((optnone))
 {
     asm volatile(
                  "vld1.32 {q0, q1}, [%1]!    \n\t" // M[m0-m7]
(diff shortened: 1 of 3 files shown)


crazyhappygame:
@PatriceJiang could you check if below helps:
https://github.com/cocos2d/cocos2d-x/pull/18895

PatriceJiang:
I think `asm` code is considered manually optimized, and disable compiler's optimize action is reasonable. 
It's also difficult to predict how will compiler 'optimize' our asm code.

crazyhappygame:
@sp-stefano-lanza could you check this fix and issue? Is it ok to disable optimization for asm?

crazyhappygame:
@PatriceJiang does #18895 fix this crash?

PatriceJiang:
The test case dose not crash on 64bit mode. @crazyhappygame

PatriceJiang:
According to the [arm document](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.100748_0606_00_en/ddx1471430827125.html)
> Use the `volatile` qualifier for assembler instructions that have processor side-effects, which the compiler might be unaware of. The volatile qualifier disables certain compiler optimizations. The `volatile` qualifier is optional.

`asm volatile` is expected to disable compiler optimization, but it takes no effect. That is perhaps `inline` function's side effect.

Issue #18948, opened 2018-07-19
CrashIfClientProvidedBogusAudioBufferList: reference #17494


guilhemvors:
- cocos2d-x version: 3.17
- devices test on: iphone 8
- Xcode version: 9.4.1

Steps to Reproduce:

1. Just create a cpp-empty-test with this code:

```
auto startButton = ui::Button::create();
startButton->setTitleText("START");
startButton->setPosition(visibleSize * 0.5f + Size(-100, 0));
this->addChild(startButton);
startButton->addTouchEventListener([=](Ref* sender, ui::Widget::TouchEventType type)
                                  {
                                      if (type == ui::Widget::TouchEventType::ENDED)
                                      {
                                          auto testBogusAudio = [this](float dt) {
                                              AudioEngine::play2d("iphone/combo_1.mp3", false, 1.0f);
                                              AudioEngine::play2d("iphone/combo_2.mp3", false, 1.0f);
                                          };
                                          cocos2d::Director::getInstance()->getScheduler()->schedule(testBogusAudio, this, 0.f, false, "AudioSwitchStateTest");
                                      }
                                  });
    
auto stopButton = ui::Button::create();
stopButton->setTitleText("STOP");
stopButton->setPosition(visibleSize * 0.5f + Size(100, 0));
this->addChild(stopButton);
stopButton->addTouchEventListener([=](Ref* sender, ui::Widget::TouchEventType type)
                                       {
                                           if (type == ui::Widget::TouchEventType::ENDED)
                                           {
                                               AudioEngine::uncacheAll();
                                               cocos2d::Director::getInstance()->getScheduler()->unschedule("AudioSwitchStateTest", this);
                                           }
                                       });
```

2. Launch the test and click on the start/stop button several times: the crash is kind of elusive and doesn't occurs everytime, it may take a few minutes to have it.

Log: https://textuploader.com/dz77q
Stack of the crash: https://textuploader.com/dz77l
Link to the combo_1.mp3 & combo_2.mp3: 
https://www.dropbox.com/s/jp60anbxhgjyi17/combo_1.mp3?dl=0
https://www.dropbox.com/s/2n9uvr39nvx12hp/combo_2.mp3?dl=0

Primary analysis:
The commit that *may fix the bug by delaying the destruction of the _pcmData in the audio crash doesn't solve the problem (see this issue: https://github.com/cocos2d/cocos2d-x/issues/17494).

From what I have gathered:

[...]

PR #19227, opened 2018-12-03
alBufferData instead of alBufferDataStatic for small audio file on Apple

drelaptop:
- Based on https://github.com/ironhidegames/cocos2d-x/pull/2

- Tested on iPhone X (iOS 11.2.2), `AudioSwitchStateTest ` case, not crash for 30 minutes

--- 

refer to https://github.com/cocos2d/cocos2d-x/issues/18948#issuecomment-419071695

> don't use alBufferDataStaticProc, so we let openAL perform a copy of the audio data, and let him handle the memory of this datas, while we clean our pcmData immediately


guilhemvors:
I have merged the pull. Will keep you up updated if it comes back.

PR #19230, opened 2018-12-04
[cmake] add find_host_library into iOS toolchain file

drelaptop:
- you can use it like

```CMake
find_host_library(Xamarin_LIB # var to save result
  NAMES Xamarin.iOS # Xamarin.iOS.framework
  PATHS ${CMAKE_CURRENT_SOURCE_DIR}/tmp) # libs dir

message(STATUS "Xamarin_LIB: ${Xamarin_LIB}") # print find libs result
```

- the reason to add this is because iOS toolchain file changes the `find_library` default behavior by:

```
    set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
```

### DIFF
--- a/cmake/ios.toolchain.cmake
+++ b/cmake/ios.toolchain.cmake
@@ -204,3 +204,12 @@ macro(find_host_package)
     set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
     set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 endmacro(find_host_package)
+
+# This macro lets you find library on the host system
+macro(find_host_library)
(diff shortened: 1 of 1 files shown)


crazyhappygame:
I think that we have to change (and simplify) this one iOS.toolchain to few iOS toolchains one per each configuration.

 It should be similar to what was done for Android here https://github.com/cocos2d/cocos2d-x-external/blob/master/tools/android_arm64_v8a_toolchain.cmake

(Probably not in this code review, and we can do that as first in cocos2d-x-external)

drelaptop:
agree, we can try to simply iOS toolchain in cocos2d-x-external firstly.

Issue #14092, opened 2015-10-09
crash when 'ScriptingCore::cleanup'


SongCF:
platform: ios
engine: v3.7 js-binding

when I exit the game, call the following code (same as the demo of cpp-helloworld click close menu item)

```
    Director::getInstance()->end();
#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
    exit(0);
#endif
```

allways crash at ScriptingCore::cleanup:

```
    for (auto iter = _js_global_type_map.begin(); iter != _js_global_type_map.end(); ++iter)
    {
        CCLOG("typeName = %s", iter->first.c_str());
        free(iter->second->jsclass);  //  EXC_BAD_ACCESS
        free(iter->second);
    }
```

I try to find where freed `iter->second->jsclass` before `ScriptingCore::cleanup`, but not found.
why crash at this code, please tell me if you known, thanks.

pandemosth:
+1

This has been an issue for a while and causes lots of unnecessary crash reports.

ANUBIS-V:
+1
ios9 cocos2d-x 3.9

minggo:
Could somebody provide a test case for it to reproduce it?

minggo:
oh, i got it, i will test it

minggo:
> when I exit the game, call the following code (same as the demo of cpp-helloworld click close menu item)
> 
> ``` c++
>     Director::getInstance()->end();
> #if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
>     exit(0);
> #endif
> ```

How did you do it like this? Did you invoke c++ codes from javascript? Could you please provide a test case?

wijanarko-sukma:
+1
I was wondering if this issue is fixed or not because it happened to me every time I restarted the game using restart method from the Director
`cocos2d::Director::getInstance()->restart();`
FYI, I mainly used C++ on the game with JSB so I frequently invoke C++ codes from javascript and vice versa

SongCF:
I'm sorry I forgot how to solved it. i no longer develop games

Dimon4eg:
Why the issue is closed?
It still reproduces on cocos2d-x 3.17 with js binding.

js code to reproduce:

`close();`

Dimon4eg:
I did following fix (commented destroys):

```C++
AppDelegate::~AppDelegate()
{
#if USE_AUDIO_ENGINE
    AudioEngine::end();
[...]

PR #7347, opened 2014-07-08
fixed memory leak cause by singleton.

WenhaiLin:

PR #19236, opened 2018-12-10
fix double free issues when js/lua-tests exit on iOS

drelaptop:
- `exit(0)` in `JSB_closeWindow` case ScriptEngine static memory freed, so when the `ScriptEngineManager::destroyInstance` called later in AppDelegate destructor, double free occurs

### DIFF
--- a/cocos/platform/CCApplicationProtocol.h
+++ b/cocos/platform/CCApplicationProtocol.h
@@ -28,7 +28,6 @@ THE SOFTWARE.
 #define __CC_APPLICATION_PROTOCOL_H__
 
 #include "platform/CCPlatformMacros.h"
-#include "base/CCScriptSupport.h"
 #include "base/CCAutoreleasePool.h"
 #include "base/ccTypes.h"
 
@@ -67,9 +66,6 @@ public:
      * @lua NA
      */
     virtual ~ApplicationProtocol(){
-#if CC_ENABLE_SCRIPT_BINDING
-        ScriptEngineManager::destroyInstance();
-#endif
         /** clean auto release pool. */
         PoolManager::destroyInstance();
     }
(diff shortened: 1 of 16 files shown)


drelaptop:
this should fix https://github.com/cocos2d/cocos2d-x/issues/14092 , welcome to review @Dimon4eg

cocos/scripting/js-bindings/manual/ScriptingCore.cpp
```diff
@@ -396,17 +396,7 @@ bool JSB_core_restartVM(JSContext *cx, uint32_t argc, jsval *vp)
 
 bool JSB_closeWindow(JSContext *cx, uint32_t argc, jsval *vp)
 {
-    EventListenerCustom* _event = Director::getInstance()->getEventDispatcher()->addCustomEventListener(Director::EVENT_AFTER_DRAW, [&](EventCustom *event) {
-        Director::getInstance()->getEventDispatcher()->removeEventListener(_event);
-        CC_SAFE_RELEASE(_event);
-        
-        ScriptingCore::getInstance()->cleanup();
```
drelaptop, line 403: no meaning to do `cleanup` in `JSB_closeWindow `, `Director::getInstance()->end()` will do it latterly. another reason is `_event ` have been freed when the lambda callback trigger

PatriceJiang (approved):

drelaptop:
no reasons https://github.com/cocos2d/cocos2d-x/commit/a05d6a55e25238d091debff194f79b4ab300158f disable close menu in lua-tests only on iOS platforms, I enable it and tested with this PR, works

cocos/platform/CCApplicationProtocol.h
```diff
@@ -67,9 +66,6 @@ class CC_DLL ApplicationProtocol
      * @lua NA
      */
     virtual ~ApplicationProtocol(){
-#if CC_ENABLE_SCRIPT_BINDING
-        ScriptEngineManager::destroyInstance();
```
drelaptop, line 71: director destructor have done it, so no need to do it again. 

this is added at https://github.com/cocos2d/cocos2d-x/pull/7347 to fix memory leak

Dimon4eg:
@drelaptop I've tested this PR and it works nice.
Thank you!

PR #19237, opened 2018-12-10
[android] all test projects allow cleartext HTTP traffic

PatriceJiang:
[cpp test] 19/2 httpclient test failed due to `java.io.IOException: Cleartext HTTP traffic to * not permitted`  on MI MIX 2S

fix by enabling `cleartext HTTP traffic` by default 

doc https://android-developers.googleblog.com/2016/04/protecting-against-unintentional.html


crazyhappygame:
Maybe instead of android:usesCleartextTraffic="true" we may change test to use https instead of http?

PatriceJiang:
HTTP is still widely used and we should keep it in our tests. Developer can add [security-config](https://developer.android.com/training/articles/security-config) in their apps.

Issue #17924, opened 2017-06-13
what's wrong with TextureAtlas::drawNumberOfQuads(int, int)


toreinkun:
on android, sometimes will cause these error, the crash stack is reported by bugly.
i dont't know how to reproduce it.

# GLThread 1985(1985)
SIGSEGV(SEGV_MAPERR)
1 #00 pc 00010620 /system/lib/libc.so (__memcpy_base+88) [armeabi-v7a]
2 #01 pc 000fd4e7 /system/vendor/lib/egl/libGLESv2_adreno.so (EsxVertexArrayObject::UpdateInternalVbos(EsxDrawDescriptor const*, unsigned int, EsxAttributeDesc const*)+1158) [armeabi-v7a]
3 #02 pc 00127ee9 /system/vendor/lib/egl/libGLESv2_adreno.so (A4xVertexArrayObject::CalcVfdRegs(EsxDrawDescriptor const*, A4xVfdRegs*, int)+932) [armeabi-v7a]
4 #03 pc 0012bf0b /system/vendor/lib/egl/libGLESv2_adreno.so (A4xContext::ValidateVfdState(EsxDrawDescriptor const*, int)+38) [armeabi-v7a]
5 #04 pc 00133e3f /system/vendor/lib/egl/libGLESv2_adreno.so (A4xContext::ValidateState(EsxDrawDescriptor const*)+1426) [armeabi-v7a]
6 #05 pc 001340dd /system/vendor/lib/egl/libGLESv2_adreno.so (A4xContext::HwValidateGfxState(EsxDrawDescriptor const*)+4) [armeabi-v7a]
7 #06 pc 000bd0c3 /system/vendor/lib/egl/libGLESv2_adreno.so (EsxContext::ValidateGfxState(EsxDrawDescriptor const*)+486) [armeabi-v7a]
8 #07 pc 000bae75 /system/vendor/lib/egl/libGLESv2_adreno.so (EsxContext::DrawElementsInstanced(EsxPrimType, unsigned int, EsxPixType, void const*, unsigned int)+284) [armeabi-v7a]
9 #08 pc 000a93fd /system/vendor/lib/egl/libGLESv2_adreno.so (EsxContext::GlDrawElements(unsigned int, int, unsigned int, void const*)+48) [armeabi-v7a]
10 #09 pc 000a0fbf /system/vendor/lib/egl/libGLESv2_adreno.so (glDrawElements+38) [armeabi-v7a]
[...]

PR #19256, opened 2018-12-19
standardize pause & resume logic on GLSurfaceView

drelaptop:
refer to

   https://stackoverflow.com/questions/2112768/prevent-onpause-from-trashing-opengl-context

This will reduce below issues when Activity onResume

```
call to OpenGL ES API with no current context (logged once per thread)
```


minggo (approved):

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
```diff
@@ -199,7 +199,7 @@ public void run() {
             }
         });
         this.setRenderMode(RENDERMODE_WHEN_DIRTY);
-        //super.onPause();
```
drelaptop, line 202: The meaning of this PR is finger out clearly that we want to preserve GLContext onPause, the older workaround is this comments.

insama:
#17924 
我也遇到了一样的问题，在google后台看到很多类似的堆栈信息的闪退。

Issue #19258, opened 2018-12-20
PR: Fix several bugs with button's title (#19073)，breaks Buttons custom fonts style in cocos studio


sonyps5201314:
- cocos2d-x version:git v3 latest
- devices test on:win32
- developing environments
   - VS version:vs2017 15.9.4
Steps to Reproduce:
cocos studio中的样子：
![image](https://user-images.githubusercontent.com/20564099/50287881-319a0780-049f-11e9-94d3-14e78efe0495.png)
[test_bug.zip](https://github.com/cocos2d/cocos2d-x/files/2699089/test_bug.zip)

1. 下载附件2
[cpp-empty-test.zip](https://github.com/cocos2d/cocos2d-x/files/2699087/cpp-empty-test.zip)
，直接编译运行看到结果，看到出问题的样子（大小缩小了）：
![image](https://user-images.githubusercontent.com/20564099/50287924-5ee6b580-049f-11e9-961d-61ed6404ea98.png)

2.撤销该提交的更改(UIButton.h,UIButton.cpp),重新编译cocos及程序，就看到正确的样子了。
![image](https://user-images.githubusercontent.com/20564099/50287830-01526900-049f-11e9-998a-a0e365068512.png)
 
官方你们说该怎样整？官方升级cocos studio，还是开放cocos studio的源代码,还是修改cocos2dx源代码以修复本问题？

English:
What cocos studio looks like:
1. Download attachment 2, compile and run directly to see the result, and see what the problem looks like (the size is reduced) :
2. Undo the submitted changes(UIButton.h,UIButton.cpp) and recompile cocos and the program, and you will see the correct appearance.
What do you say the authorities do?Will you officially upgrade cocos studio, open the source code of cocos studio, or modify the source code of cocos2dx to fix this problem?

reference linking：
https://github.com/cocos2d/cocos2d-x/issues/19253

PR: http://github.com/cocos2d/cocos2d-x/pull/19073

PatriceJiang:
@sonyps5201314 please check if this issue fixed with https://github.com/cocos2d/cocos2d-x/pull/19260

sonyps5201314:
测试了，的确是修复了，感谢！

PR #19260, opened 2018-12-21
[uibutton] fix label size init order

PatriceJiang:
size of title label font should be set after `titleLabelType`

ref https://github.com/cocos2d/cocos2d-x/issues/19258


cocos/editor-support/cocostudio/WidgetReader/ButtonReader/ButtonReader.cpp
```diff
@@ -878,9 +878,7 @@ namespace cocostudio
         Color3B titleColor(textColor->r(), textColor->g(), textColor->b());
         button->setTitleColor(titleColor);
         
-        int titleFontSize = options->fontSize();
-        button->setTitleFontSize(titleFontSize);
-        
+
         std::string titleFontName = options->fontName()->c_str();
         button->setTitleFontName(titleFontName);
```
PatriceJiang, line 883: titleLabelType can be changed here, `getTitleFontSize` will no longer return the right size.

PR #19263, opened 2018-12-22
ios_signing

crazyhappygame:


crazyhappygame:
@drelaptop Could you check below error:
https://travis-ci.org/cocos2d/cocos2d-x/jobs/471296792

> Code Signing Warning: "cpp-empty-test" isn't code signed but requires entitlements. It is not possible to add entitlements to a binary without signing it.
> Code Signing Error: Code signing is required for product type 'Application' in SDK 'iOS 12.1'

How can I setup code signing for iOS from CMake?

drelaptop:
https://stackoverflow.com/questions/40664125/cmake-and-code-signing-in-xcode-8-for-ios-projects

but I don't think it's necessary to sign it in Travis, and we even didn't have a __Team ID__ just for tests.

crazyhappygame:
@drelaptop Thank you for help! That worked for me.

I have no experience with iOS/mac I would like to port my games to iOS/Mac.
Could you help me once again?
Could you check below screenshot: https://ibb.co/VtcTQnc ?
Why for cpp-empty-test on iPad  (both device and simulator) game is not in full screen mode?
What is the best way to fix it?

PR #19273, opened 2018-12-30
Precompiled header

crazyhappygame:


crazyhappygame:
Appveyor CMake build time reduced from ~24min to ~12min.

drelaptop:
I doubt why using `if(WIN32)` to limit the precompiled header, can we use it for other platforms ? @crazyhappygame

crazyhappygame:
There is no native cmake support for precompiled headers... So if/else is needed anyway.

I knew how to enable precompiled headers for win - so I did it.
I do not know how to do it properly for clang/gcc. But I would love to see it for all platforms.

PR #19274, opened 2019-01-03
Added md5 message digest function to encrypt strings.

EricDDK:
This is a function of encrypting strings, generally used for message digests and checksum data.


cocos/base/ccUtils.h
```diff
@@ -173,6 +173,12 @@ namespace utils
      */
     CC_DLL std::string getFileMD5Hash(const std::string &filename);
 
+	/**
+	*  Gets the md5 hash for the given string.
+	*  @param the string to calculate md5 hash.
+	*  @return The md5 hash for the string
+	*/
+	CC_DLL std::string getStringMD5Hash(const std::string &dataStr);
```
PatriceJiang, line 181: Please fix the code alignment

PatriceJiang:
Could you please rebase the change on to the recent v3 branch since the commit log is not brief.

screen shot 2019-01-04 at 3 02 11 pm

cocos/ui/UIImageView.h
```diff
@@ -118,6 +118,12 @@ class CC_GUI_DLL ImageView : public Widget , public cocos2d::BlendProtocol
      * @see `setCapInsets(const Rect&)`
      */
     const Rect& getCapInsets()const;
+
+	/**
+	* Get ImageView's texture file.
+	* @return texture file.
+	*/
+	const std::string getTextureFile()const;
```
Dimon4eg, line 126: better to return const reference not just const value

Dimon4eg (changes requested):

cocos/ui/UIImageView.cpp
```diff
@@ -238,6 +238,11 @@ const Rect& ImageView::getCapInsets()const
     return _capInsets;
 }
 
+const std::string ImageView::getTextureFile()const
```
halx99, line 241: No need const qualifier for non-reference result type.

PR #19279, opened 2019-01-07
Added md5 message digest function to encrypt strings.

EricDDK:
#19274 
Resubmit and format comment.


minggo:
The indent is not correct.

PR #19284, opened 2019-01-09
update 3rd libs & console, fix file issues

PatriceJiang:


PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19292

PR #19302, opened 2019-01-16
Fix deadlock in FileUtils::fullPathForFilename in Windows

sp-rafael-lecina:
Fix deadlock in windows between ``FileUtils::_mutex`` and call to ``CCLOG`` (internally implemented using SendMessage which is blocking)

If ``FileUtils::fullPathForFilename`` is called from main thread and from a secondary thread, code may call to ``CCLOG``(internally ``SendMessage``) in secondary thread may be done while main thread is locked waiting to ``FileUtils::_mutex``. That way ``SendMessage`` cannot be handled nor mutex released

Fix is to reduce the scope where the mutex is locked

### DIFF
--- a/cocos/platform/CCFileUtils.cpp
+++ b/cocos/platform/CCFileUtils.cpp
@@ -814,20 +814,9 @@ std::string FileUtils::getPathForFilename(const std::string& filename, const std
     return path;
 }
 
-std::string FileUtils::fullPathForFilename(const std::string &filename) const
+std::string FileUtils::fullPathForFilenameImpl(const std::string &filename) const
 {
-    
     DECLARE_GUARD;
-
-    if (filename.empty())
-    {
-        return "";
-    }
-
-    if (isAbsolutePath(filename))
-    {
-        return filename;
-    }
 
     // Already Cached ?
     auto cacheIter = _fullPathCache.find(filename);
@@ -856,13 +845,30 @@ std::string FileUtils::fullPathForFilename(const std::string &filename) const
 
         }
     }
+    // The file wasn't found, return empty string.
+    return "";    
+}
 
-    if(isPopupNotify()){
+std::string FileUtils::fullPathForFilename(const std::string &filename) const
+{
+    if (filename.empty())
+    {
+        return "";
+    }
+
+    if (isAbsolutePath(filename))
+    {
+        return filename;
+    }
+
+    std::string fullpath = fullPathForFilenameImpl(filename);
+
+    if(fullpath.empty() && isPopupNotify())
+    {
         CCLOG("cocos2d: fullPathForFilename: No file found at %s. Possible missing file.", filename.c_str());
     }
 
-    // The file wasn't found, return empty string.
-    return "";
+    return fullpath;
 }
 
 
--- a/cocos/platform/CCFileUtils.h
+++ b/cocos/platform/CCFileUtils.h
@@ -1007,6 +1007,11 @@ protected:
 
         AsyncTaskPool::getInstance()->enqueue(AsyncTaskPool::TaskType::TASK_IO, [](void*){}, nullptr, std::move(lambda));
     }
+private:
+    /**
+     *  Minimum scope to be thread safe needed by fullPathForFilename
+     */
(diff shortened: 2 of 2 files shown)


PatriceJiang:
It sounds like a lock issue of `CCLOG`, perhaps we need a thread-safe `CCLOG` on windows?

sp-rafael-lecina:
@PatriceJiang CCLOG is thread safe in windows because SendMessage locks until message is processed in main thread
The problem is making a blocking call unrelated to what the mutex is protecting while mutex is locked (that mutex is there only to protect the filename cache)
To avoid deadlocks is good practice to lock mutexes only in the minimum scope needed, in this case only during the access to the filename cache

PatriceJiang:
So, there is no deadlock here. This PR aims to reduce lock scope, am I correct?

sp-rafael-lecina:
@PatriceJiang There is a deadlock if used this function in a multithreaded environment. It uses a _mutex, so it is supposed to support multiple threads accessing. Indeed it is happening to us in our project while downloading assets in background

Reducing lock scope to not include CCLOG call removes the deadlock

crazyhappygame:
Is this problem only for debug version or release as well?

crazyhappygame:
Is this problem only for windows32

sp-rafael-lecina:
@crazyhappygame 
It happens only for Win32 as far as I can tell
https://github.com/cocos2d/cocos2d-x/blob/e71ef8218cb58a66ec07cf9cc5ac80ac4e695ebd/cocos/base/CCConsole.cpp#L105

It happens in debug
https://github.com/cocos2d/cocos2d-x/blob/e71ef8218cb58a66ec07cf9cc5ac80ac4e695ebd/cocos/platform/CCPlatformMacros.h#L220
https://github.com/cocos2d/cocos2d-x/blob/e71ef8218cb58a66ec07cf9cc5ac80ac4e695ebd/cocos/platform/CCPlatformMacros.h#L226

PatriceJiang:
According to the description, I think it is better to make CCLOG `non-block`. Since it is commonly used in multiple thread context.

minggo:
> @PatriceJiang There is a deadlock if used this function in a multithreaded environment. It uses a _mutex, so it is supposed to support multiple threads accessing. Indeed it is happening to us in our project while downloading assets in background
> 
> Reducing lock scope to not include CCLOG call removes the deadlock

@sp-andres-barrera cocos2d-x functions are not thread safe, you can not use them in multi-thread environment.

crazyhappygame:
@sp-andres-barrera did you call it from different thread or cocos2d-x internally call it from different thread? (from downloading, sync thread)?

sp-rafael-lecina:
If they are not thread safe why are they protected by a mutex?
FileUtils seems to try to be thread safe
@crazyhappygame Yes, we call some of those FileUtils functions from background loading threads to get full asset paths

minggo:
Yep, FileUtils is special as developers ask for it. Sorry, i forgot it.

PatriceJiang:
@sp-rafael-lecina `SendMessage` was replaces  with `PostMessage` in a recent [PR](https://github.com/cocos2d/cocos2d-x/pull/19569/files),  `SendLogToWindow` is now non-blocking. The deadlock's condition should be removed, is it correct? 


edit: this pr was reverted

sp-rafael-lecina:
@PatriceJiang That won't deadlock, but you have a dangling pointer on that code
``(LPARAM)(LPVOID)&myCDS`` is the address of a local variable

sp-rafael-lecina:
@PatriceJiang Ping!

PatriceJiang (approved):

PatriceJiang (approved):

sp-rafael-lecina:
@PatriceJiang Hi
Related to this SendMessage issue we end up totally removing the SendMessage call from CCLOG in our code branch because we didn't find any handling for the WM_COPYDATA message in the code
Are you using it somehow? because sending that message maybe is legacy

PatriceJiang:
I totally agree the SendMessage is unnecessary.
And yes, we need the code.

Issue #19319, opened 2019-01-18
Remove Xcode projects


minggo:
Should just use cmake to construct projects as other platforms.

drelaptop:
with https://github.com/cocos2d/cocos2d-x/issues/18717#issuecomment-455471357 discuss, we hold same idea to remove Xcode build files. and have made cmake works on iOS/macOS too.

the reason to keep it in alpha release is that iOS generated Xcode project don't so convenient like origin build files.

1. info.plist issues
1. iOS sign config

that is said if developer add a file, and re-execute `cmake ..`, all custom config at Xcode will disappear. could we improve it, or just let developers to accept it? @crazyhappygame

minggo:
If using cmake, then developers should do custom configuration in cmake. Is there any configuration that can not do in cmake?

drelaptop:
I think no necessary feature cmake can't support, only some are complex, not friendly. 

__options:__
1. custom all __build__ configurations, for example headers include, link libs ... , cmake works well, and convenient for users. for we known cmake is great cross-platform build tools.

2. custom all __pack__ configurations isn't so convenient, compare with operation at Xcode UI. especially features special with other platform.

    - for example, signing operation https://github.com/cocos2d/cocos2d-x/pull/19263
    - for example, the ` iOS Deployment Target` for root project, needed by __Product -> archive__, cmake don't support until 3.11 release. https://github.com/cocos2d/cocos2d-x/pull/19315

crazyhappygame:
Signing is very easy
I successfully signed iOS by adding following two lines: 
```
        SET_XCODE_PROPERTY(${APP_NAME} CODE_SIGN_IDENTITY "iPhone Developer")
        SET_XCODE_PROPERTY(${APP_NAME} DEVELOPMENT_TEAM "XXXXXXXXXX")
```

drelaptop:
above comment is the way to sign iOS, but it's not obvious to find `DEVELOPMENT_TEAM` value.

- the right way to find it by the __Organizational Unit__ 

```
Applications -> Utilities -> Keychain Access.
Under the 'login' Keychain, go into the 'Certificates' category.
Scroll to find your development or distribution certificate. They will read:
iPhone Distribution: Team Name (certificate id)
or
iPhone Developer: Team Name (certificate id)
Simply double-click on the item, and the
"Organizational Unit"
```
[...]

PR #19327, opened 2019-01-21
refine cmake, add iOS sign config

drelaptop:
for discuss https://github.com/cocos2d/cocos2d-x/issues/19319#issuecomment-455931410

> improve cmake to remove Xcode config files


Dimon4eg:
I see a lot of copy-pase code.
Myabe better to move it to one `.cmake` file and use it.

drelaptop:
thanks, I don't think move it to one `.cmake` file is better. 

- for one single iOS App target, only once definition.
- put it into project cmake files will be more convenient for developer to modify. users need to fill `set_xcode_property(${APP_NAME} DEVELOPMENT_TEAM "")` when need to sign in cmake

PR #19329, opened 2019-01-21
refine cmake, allow developer switch Debug/Release in IDE

drelaptop:
for https://github.com/cocos2d/cocos2d-x/issues/19319

- allow developer switch debug/release in Xcode

image


cmake/Modules/CocosConfigDefine.cmake
```diff
@@ -30,29 +30,27 @@ else()
...
- # build mode, Debug is default value
- if(NOT CMAKE_BUILD_TYPE)
-     if(DEBUG_MODE)
-         set(CMAKE_BUILD_TYPE Debug)
-     else()
-         set(CMAKE_BUILD_TYPE Release)
-     endif()
- endif()
-
  if(CMAKE_GENERATOR)
```
crazyhappygame: I think that CMAKE_GENERATOR has always some value.
It does not make sense to do if/else on CMAKE_GENERATOR

drelaptop: thanks, the condition should check `CMAKE_CONFIGURATION_TYPES`.

now the generate logs like
```
-- CMAKE_GENERATOR: Unix Makefiles
-- CMAKE_BUILD_TYPE: Debug
```
```
-- CMAKE_GENERATOR: Xcode
-- CMAKE_CONFIGURATION_TYPES: Debug;Release
```

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -1,14 +1,17 @@
 include(CMakeParseArguments)
 
-# copy resource `FILES` and `FOLDERS` to `COPY_TO` folder
-function(cocos_copy_res)
+# copy resource `FILES` and `FOLDERS` to TARGET_FILE_DIR/Resources
+function(cocos_copy_target_res cocos_target)
```
drelaptop, line 4: below copy tasks will spend times if app target existed lots of res files, for example `js-tests`.

drelaptop:
please review if available @crazyhappygame

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -172,27 +178,19 @@ endfunction()
 function(setup_cocos_app_config app_name)
```
crazyhappygame, line 178: Consider remove this function

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -172,27 +178,19 @@ endfunction()
...
-
+    set_target_properties(${app_name} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin/${app_name}")
     # auto mark code files for IDE when mark app
     if(XCODE OR VS)
-        cocos_mark_code_files(${APP_NAME})
+        cocos_mark_code_files(${app_name})
     endif()
 
-    set(APP_RES_DIR ${APP_RES_DIR} PARENT_SCOPE)
+    set(APP_RES_DIR "$<TARGET_FILE_DIR:${app_name}>/Resources" PARENT_SCOPE)
```
crazyhappygame: Please remove this variable and use $<TARGET_FILE_DIR:${app_name}> when needed

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -172,27 +178,19 @@ endfunction()
...
-        set(APP_RES_DIR "${CMAKE_BINARY_DIR}/bin/${APP_NAME}/${CMAKE_BUILD_TYPE}/Resources")
         #Visual Studio Defaults to wrong type
         set_target_properties(${app_name} PROPERTIES LINK_FLAGS "/SUBSYSTEM:WINDOWS")
-    else(LINUX)
-        set(APP_BIN_DIR "${CMAKE_BINARY_DIR}/bin/${CMAKE_BUILD_TYPE}/${APP_NAME}")
-        set(APP_RES_DIR "${APP_BIN_DIR}/Resources")
     endif()
-    set_target_properties(${app_name} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${APP_BIN_DIR}")
-
+    set_target_properties(${app_name} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin/${app_name}")
```
crazyhappygame: Please remove this variable. 
Default should be OK.

cmake/Modules/CocosBuildHelpers.cmake
```diff
@@ -172,27 +178,19 @@ endfunction()
...
     if(APPLE)
-        set(APP_BIN_DIR "${CMAKE_BINARY_DIR}/bin")
         set_target_properties(${app_name} PROPERTIES MACOSX_BUNDLE 1
                               )
     elseif(MSVC)
-        # only Debug and Release mode was supported when using MSVC.
-        set(APP_BIN_DIR "${CMAKE_BINARY_DIR}/bin/${APP_NAME}/$<CONFIG>")
-        set(APP_RES_DIR "${CMAKE_BINARY_DIR}/bin/${APP_NAME}/${CMAKE_BUILD_TYPE}/Resources")
         #Visual Studio Defaults to wrong type
         set_target_properties(${app_name} PROPERTIES LINK_FLAGS "/SUBSYSTEM:WINDOWS")
```
crazyhappygame: I think that we should remove it.
In modern cmake compiler/linker should not be set.
User should be able to setup flags when needed in root CMakeLists.txt.

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: a good suggestion, I will do it

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: the default binary dir is more deep, user can't find it easily. especially for root CMakeLists.txt, 5  tests app existed.

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: I will check this

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: all configs in `setup_cocos_app_config ` is stable, user don't need to custom it, and moving it up to root CMakeLists.txt will cause much repeat.

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: convert it into
```
# visual studio default is Console app, but we need Windows app
target_link_options(${app_name} PRIVATE "/SUBSYSTEM:WINDOWS")
```
so it didn't disturb user set flags in root CMakeLists.txt

cmake/Modules/CocosBuildHelpers.cmake
drelaptop: https://github.com/cocos2d/cocos2d-x/pull/19329/commits/0e3242b123ac492a56a708b2a763a6dfbf05194a have add print info when copy files

PR #19335, opened 2019-01-22
sync cmake changes into metal-support

drelaptop:


drelaptop:
@minggo please review

PR #19336, opened 2019-01-22
Fixed some potential crashes when loading big or invalid skeletons

benjaprieto:
Increased size of world vertices for skeleton renderer to avoid some crashes. Added some null checks to avoid crashes and logs to determine the files that cause them.

### DIFF
--- a/cocos/editor-support/spine/SkeletonAnimation.cpp
+++ b/cocos/editor-support/spine/SkeletonAnimation.cpp
@@ -88,6 +88,10 @@ SkeletonAnimation* SkeletonAnimation::createWithJsonFile (const std::string& ske
 	spAtlas* atlas = spAtlas_createFromFile(atlasFile.c_str(), 0);
 	node->initWithJsonFile(skeletonJsonFile, atlas, scale);
 	node->autorelease();
+    	if(!node->getSkeleton() || !node->getSkeleton()->data)
+    	{
+        	return nullptr;
+    	}
 	return node;
 }
 
--- a/cocos/editor-support/spine/SkeletonAnimation.cpp
+++ b/cocos/editor-support/spine/SkeletonAnimation.cpp
@@ -88,10 +88,10 @@ SkeletonAnimation* SkeletonAnimation::createWithJsonFile (const std::string& ske
 	spAtlas* atlas = spAtlas_createFromFile(atlasFile.c_str(), 0);
 	node->initWithJsonFile(skeletonJsonFile, atlas, scale);
 	node->autorelease();
-    	if(!node->getSkeleton() || !node->getSkeleton()->data)
-    	{
-        	return nullptr;
-    	}
+    if(!node->getSkeleton() || !node->getSkeleton()->data)
+    {
+        CC_SAFE_RELEASE_NULL(node);
+    }
 	return node;
 }
 
--- a/cocos/editor-support/spine/SkeletonRenderer.cpp
+++ b/cocos/editor-support/spine/SkeletonRenderer.cpp
@@ -61,7 +61,7 @@ SkeletonRenderer* SkeletonRenderer::createWithFile (const std::string& skeletonD
 }
 
 void SkeletonRenderer::initialize () {
-	_worldVertices = new float[1000]; // Max number of vertices per mesh.
+	_worldVertices = new float[4096]; // Max number of vertices per mesh.
 	
 	_clipper = spSkeletonClipping_create();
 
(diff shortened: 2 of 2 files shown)


cocos/editor-support/spine/SkeletonAnimation.cpp
```diff
@@ -88,6 +88,10 @@ SkeletonAnimation* SkeletonAnimation::createWithJsonFile (const std::string& ske
 	spAtlas* atlas = spAtlas_createFromFile(atlasFile.c_str(), 0);
 	node->initWithJsonFile(skeletonJsonFile, atlas, scale);
 	node->autorelease();
+    	if(!node->getSkeleton() || !node->getSkeleton()->data)
+    	{
+        	return nullptr;
```
gestern: Here is a memory leak, you've lost a weak reference to the allocated memory before

cocos/editor-support/spine/SkeletonRenderer.cpp
```diff
@@ -61,7 +61,7 @@ SkeletonRenderer* SkeletonRenderer::createWithFile (const std::string& skeletonD
 }
 
 void SkeletonRenderer::initialize () {
-	_worldVertices = new float[1000]; // Max number of vertices per mesh.
```
gestern, line 64: The constant is predefined in some Spine's frameworks. You can't just change it, because you don’t have enough vertices in the model, just because the artist made such a reach and expensive skeleton, it will affect the global changes of the whole Spine (for example: in SkeletonRenderer.m had another reference for it: _worldVertices = MALLOC(float, 1000); // Max number of vertices per mesh. and SkeletonExtensions.cs have similar reference).

cocos/editor-support/spine/SkeletonRenderer.cpp
```diff
@@ -127,14 +127,23 @@ void SkeletonRenderer::initWithJsonFile (const std::string& skeletonDataFile, sp
 
 void SkeletonRenderer::initWithJsonFile (const std::string& skeletonDataFile, const std::string& atlasFile, float scale) {
 	_atlas = spAtlas_createFromFile(atlasFile.c_str(), 0);
-	CCASSERT(_atlas, "Error reading atlas file.");
-
+    
+    	if (!_atlas) {
```
gestern: Conditional expressions are the opposite of assertions, if you have an “if”, you should not use an assertion and vice versa. You changed the code in such a way that it became generally incorrect.

cocos/editor-support/spine/SkeletonRenderer.cpp
```diff
@@ -127,14 +127,23 @@ void SkeletonRenderer::initWithJsonFile (const std::string& skeletonDataFile, sp
...
 	_attachmentLoader = SUPER(Cocos2dAttachmentLoader_create(_atlas));
 
 	spSkeletonJson* json = spSkeletonJson_createWithLoader(_attachmentLoader);
 	json->scale = scale;
 	spSkeletonData* skeletonData = spSkeletonJson_readSkeletonDataFile(json, skeletonDataFile.c_str());
-	CCASSERT(skeletonData, json->error ? json->error : "Error reading skeleton data file.");
+    
+    	if (!skeletonData) {
+        	CCLOG("ERROR reading skeleton data file: %s", skeletonDataFile.c_str());
+		CCASSERT(skeletonData, json->error ? json->error : "Error reading skeleton data file.");
```
gestern: All the same

cocos/editor-support/spine/SkeletonRenderer.cpp
benjaprieto: Any suggestions? Set that limit to a value that can be overriden?

cocos/editor-support/spine/SkeletonRenderer.cpp
benjaprieto: Would be enough to separate the assert from the conditional? We want to be able to log what file is the one causing the issue before the assert is called

cocos/editor-support/spine/SkeletonAnimation.cpp
benjaprieto: Thanks,

So release node and spAtlas?

cocos/editor-support/spine/SkeletonRenderer.cpp
gestern: Only one solution - you have to submit PR to spine repo . As  talking several times before, and [last time here](https://github.com/cocos2d/cocos2d-x/pull/19163#issuecomment-451872989). Any of PR which referenced to Spine runtime would be closed due to Spine license:

> The cocos2d-x can't modify spine runtime due to it's license limit. you should post PR to spine runtime repo: https://github.com/EsotericSoftware/spine-runtimes

PR #19337, opened 2019-01-22
Get latency values for http client

benjaprieto:
Add interface and members to calculate, store and retrieve latency value between client and server


benjaprieto:
@PatriceJiang @minggo  Any chance that this PR get a review/merge? Thanks!

PR #19342, opened 2019-01-23
[BugFix] enable BlendState

Mee-gu:
fix Effects-Advanced Test


cocos/2d/CCGrid.h
```diff
@@ -132,6 +132,8 @@ class CC_DLL GridBase : public Ref
      * @return Return the effect grid rect.
      */
     const Rect& getGridRect() const { return _gridRect; }
+    
+    virtual void updateBlendState();
```
minggo: don't have to be public, and be virtual

Mee-gu:
update review

PR #19347, opened 2019-01-23
Added capacity of dynamically toggle of debug rendering for Lables and Sprites

benjaprieto:
Added calls to enable and disable debug rendering of labels and sprites (also still requiring the pre-compilation options). Also added way to change the color used.

Code will be enabled by default for COCOS2D_DEBUG configurations

Added example to testcpp

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -49,7 +49,7 @@ NS_CC_BEGIN
 
 #if CC_LABEL_DEBUG_DRAW
 
-bool Label::_debugDrawEnabled = false;
+bool Label::_debugDrawEnabled = true;
 Color4F Label::_debugDrawColor = Color4F::WHITE;
 
 #endif
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -47,6 +47,13 @@
 
 NS_CC_BEGIN
 
+#if CC_LABEL_DEBUG_DRAW
+
+bool Label::_debugDrawEnabled = false;
+Color4F Label::_debugDrawColor = Color4F::WHITE;
+
+#endif
+
 /**
  * LabelLetter used to update the quad in texture atlas without SpriteBatchNode.
  */
@@ -396,10 +403,16 @@ Label::Label(TextHAlignment hAlignment /* = TextHAlignment::LEFT */,
     _hAlignment = hAlignment;
     _vAlignment = vAlignment;
 
+    // CROWDSTAR START
 #if CC_LABEL_DEBUG_DRAW
-    _debugDrawNode = DrawNode::create();
-    addChild(_debugDrawNode);
+    _debugDrawNode = nullptr;
+    if (Label::_debugDrawEnabled)
+    {
+        _debugDrawNode = DrawNode::create();
+        addChild(_debugDrawNode);
+    }
 #endif
+    // CROWDSTAR END
 
     _purgeTextureListener = EventListenerCustom::create(FontAtlas::CMD_PURGE_FONTATLAS, [this](EventCustom* event){
         if (_fontAtlas && _currentLabelType == LabelType::TTF && event->getUserData() == _fontAtlas)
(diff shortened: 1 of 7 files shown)


gestern:
Unfortunately this is not the product solution. First of all it can't be implemented under other cococ2d-* solutions, especially under js. Secondly, the cocos2d-x has a lot of drawable entities (not only Label or Sprite) with problem areas for newcomers in cocos2d-x. And if the developer permits them then these problems have a logical basis (usually incorrect content size of node or incorrect anchor point), not a static one. Thus, in this case the problems cannot be resolved by adding a macro. Third, these problems are solved exclusively by inheritance and only for their own needs. Finally, the best solution here can be addressed by name or by tag to the problematic node and drawing a separate `DrawNode` instance over node and on top of the existing one. You can use class `Console` for it.

benjaprieto:
> Unfortunately this is not the product solution. First of all it can't be implemented under other cococ2d-* solutions, especially under js. Secondly, the cocos2d-x has a lot of drawable entities (not only Label or Sprite) with problem areas for newcomers in cocos2d-x. And if the developer permits them then these problems have a logical basis (usually incorrect content size of node or incorrect anchor point), not a static one. Thus, in this case the problems cannot be resolved by adding a macro. Third, these problems are solved exclusively by inheritance and only for their own needs. Finally, the best solution here can be addressed by name or by tag to the problematic node and drawing a separate `DrawNode` instance over node and on top of the existing one. You can use class `Console` for it.

A got a bit confused with your explanation. The added functionality doesn't include any new predefinitions/macro that where not already present in the game. The only modification has been to add the ability to disable/enable and set the colors for the helpers, which are useful tools when debugging a title that may have bounding boxes issues.

I could leave that these predefinition stay deactivated by default....

gestern:
> > Unfortunately this is not the product solution. First of all it can't be implemented under other cococ2d-* solutions, especially under js. Secondly, the cocos2d-x has a lot of drawable entities (not only Label or Sprite) with problem areas for newcomers in cocos2d-x. And if the developer permits them then these problems have a logical basis (usually incorrect content size of node or incorrect anchor point), not a static one. Thus, in this case the problems cannot be resolved by adding a macro. Third, these problems are solved exclusively by inheritance and only for their own needs. Finally, the best solution here can be addressed by name or by tag to the problematic node and drawing a separate `DrawNode` instance over node and on top of the existing one. You can use class `Console` for it.

> I could leave that these predefinition stay deactivated by default....

Your changes require a new definition in the project (in makefiles, xCode, and gradles). Because you do not check if a macro is defined (You used `#if` instead of `#if defined` or `#ifdef`) at all. It would be easy if your changes **were significant** and if today there  were not 20+ more such macros in the project. On the another hand, your changes will affect already working projects, working assembly chains and require a documentation. And finally, I have already explained above, that your changes are superfluous, because this is done by another way.

cocos/2d/CCLabel.cpp
```diff
@@ -47,6 +47,13 @@
 
 NS_CC_BEGIN
 
+#if CC_LABEL_DEBUG_DRAW
```
gestern, line 50: https://docs.microsoft.com/en-us/cpp/preprocessor/hash-ifdef-and-hash-ifndef-directives-c-cpp?view=vs-2017

benjaprieto: I added code to leave the behaviour of the debug helpers as it was originally, but still keeping the ability to enable or disable them if set a preprocessor definition to use them

No new macros have been added

And as any can review, the few #if CC_LABEL_DEBUG_DRAW  are following the way the code was using it before this branch

cocos/2d/CCSprite.h
```diff
@@ -647,6 +647,12 @@ CC_CONSTRUCTOR_ACCESS :
      * @lua     init
      */
     virtual bool initWithFile(const std::string& filename, const Rect& rect);
+
+#if CC_SPRITE_DEBUG_DRAW
+    static void SetDebugDrawLevel(int value);
+    static int GetDebugDrawLevel() { return _debugDrawLevel; }
+    static void SetDebugDrawColor(const Color4F& color);
```
gestern: The capital letter in the name of the function is contrary to the current edition of cocos2dx code convention.

benjaprieto: Updated and fixed

gestern (approved):

minggo:
static functions should be declared before all other functions as mentioned in https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#declaration-order

benjaprieto:
> static functions should be declared before all other functions as mentioned in https://github.com/cocos2d/cocos2d-x/blob/v3/docs/CODING_STYLE.md#declaration-order

Submitted the requested change (sorry for the late reaction)

benjaprieto:
@minggo I submitted the changes you requested. Can you check if this will still be merged in? Thanks!

cocos/2d/CCSprite.h
```diff
@@ -727,6 +738,12 @@ CC_CONSTRUCTOR_ACCESS :
 
 private:
     CC_DISALLOW_COPY_AND_ASSIGN(Sprite);
+    
+#if CC_SPRITE_DEBUG_DRAW
+    static int _debugDrawLevel;	       /// Determine to draw or not debug sprite info (0 disabled, 1 bbox, 2 texture box)
```
minggo: It seems `_debugDrawLevel ` is not used as the description. It is used as a bool to enable/disable debug draw.

benjaprieto: Thanks, you are completely right, I changed to a bool to represent better the functionality, updated the interface, comments and the test

minggo (approved):

PR #19348, opened 2019-01-24
Enable physics & Shader Tests Basic

PatriceJiang:


cocos/2d/CCNode.h
```diff
@@ -1636,6 +1637,8 @@ class CC_DLL Node : public Ref
     void setAdditionalTransform(const Mat4& additionalTransform);
     void setAdditionalTransform(const AffineTransform& additionalTransform);
 
+    virtual cocos2d::RenderCommand* getRenderCommand() { return nullptr; }
```
minggo, line 1640: Why need this function?

PR #19362, opened 2019-01-27
gradle-4.10.1 com.android.tools.build:gradle:3.3.0

crazyhappygame:


crazyhappygame:
@mars3142 Do you have any idea how to fix it?
It fails only for windows

`Task :libcocos2dx:transformClassesAndResourcesWithProguardForRelease FAILED`

drelaptop:
refer to https://github.com/cocos-creator/cocos2d-x-lite/pull/1671, it's OK to switch gradle version to 4.10 + 3.2 with the assets dir changed. that was tested. @crazyhappygame 

> upgrade to 4.10 + 3.3, unless you test it with android studio locally.

crazyhappygame:
@drelaptop I am so sorry I don't understand your question.
I updated gradle and and gradle plugin to latest version. Everything should work as before.

drelaptop:
it should be, but always not. you can't assume gradle have arrived to a perfect  forward compatibility.

EricDDK:
@crazyhappygame 
gradle 3.10+ will occur that the lua project's auto build APK didn't contain assets folder on android studio.

PR #19367, opened 2019-01-29
[Feature] Cache RenderPipeline and RenderCommandEncoder

Mee-gu:
optimize performance


Mee-gu:
revert caching render pipeline

PR #19369, opened 2019-01-30
Fix SIGABRT signal in StringUtils::format()

gestern:


crazyhappygame:
What do you think about following implementation
```
    va_list args;
    // get size
    va_start(args, format);
    int size = vsnprintf(0, 0, format, args);
    // std::cout << size << "\n";
    va_end(args);

    // fill buffer
    std::string buffer(size + 1, 0);
    va_start(args, format);
    int size2 = vsnprintf(buffer.data(), buffer.size(), format, args);
    assert(size2 == int(buffer.size()-1));
    va_end(args);
    return buffer;
```

gestern:
Your version is faster and looks shorter, but as you know, it also has major problems with portability, for example, some old implementations(MSVS until version 2005) did not have a check for the first zero parameter and the second zero. Should we be bothered? I think not, since the absolute majority of implementations `vsnprintf()` are guaranteed by checking these parameters.

crazyhappygame:
Cocos2dx supports really only msvc 2015

ilqvya:
What the f&&k? Use variadic templates instead

gestern:
> variadic templates

The version @crazyhappygame  was slightly faster than mine, your proposal with "variadic templates" is about 8-10 times slower.

ilqvya:
> about 8-10 times slower.

Do you have or even try to make some benchmarks for proof?

gestern:
> > about 8-10 times slower.
> 
> Do you have or even try to make some benchmarks for proof?

After you show your version, but if you do something like this (https://github.com/c42f/tinyformat), you will find results there

cocos/base/ccUTF8.cpp
```diff
@@ -34,49 +34,27 @@ NS_CC_BEGIN
...
-
-    va_start(args, format);
-    int nret = vsnprintf(&buffer.front(), buffer.length() + 1, format, args);
-    va_end(args);
-
-    if (nret >= 0) {
-        if ((unsigned int)nret < buffer.length()) {
-            buffer.resize(nret);
+std::string format(const char* fmt, ...) {
+    static const size_t BUFFER_LENGTH = 512;
```
minggo, line 38: I think use macro or temple variable is better than a static variable.

cocos/base/ccUTF8.cpp
```diff
@@ -34,49 +34,27 @@ NS_CC_BEGIN
...
+    size_t estimated_len = std::max(strlen(fmt), BUFFER_LENGTH);
```
minggo, line 39: According the cocos2d-x coding style, `estimated_len` -> `estimatedLen `

cocos/base/ccUTF8.cpp
```diff
@@ -34,49 +34,27 @@ NS_CC_BEGIN
...
+    std::string buffer;
+
+    va_list ap;
+    va_start(ap, fmt);
+
+    while (true) {
+        buffer.resize(estimated_len);
+        const int actual_size = vsnprintf(const_cast<char *>(buffer.data()), estimated_len, fmt, ap);
```
minggo, line 47: actual_size -> actualSize as mentioned above

minggo:
How to reproduce the issue? Could you please add a test case?

gestern:
> How to reproduce the issue? Could you please add a test case?

I do not have STR since it has been more than 2 weeks.

minggo:
Then what's the reason to cause it crash? I can not understand.

minggo:
Why mark all the comments as resolved?

gestern:
> Then what's the reason to cause it crash? I can not understand.

I don’t remember, but the call stack led me to this function, if you think that the code is absolutely safe, I will roll back my changes

PR #19383, opened 2019-02-04
If path to file is tool long crash is possible.

sbrednikhin:
If path to file is tool long crash is possible, because of chart buffer overflow.

### DIFF
--- a/cocos/2d/CCFontAtlasCache.cpp
+++ b/cocos/2d/CCFontAtlasCache.cpp
@@ -36,7 +36,7 @@
 NS_CC_BEGIN
 
 std::unordered_map<std::string, FontAtlas *> FontAtlasCache::_atlasMap;
-#define ATLAS_MAP_KEY_BUFFER 255
+#define ATLAS_MAP_KEY_PREFIX_BUFFER_SIZE 255
 
 void FontAtlasCache::purgeCachedData()
 {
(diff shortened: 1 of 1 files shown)


sbrednikhin:
Usage of stringstream instead of snprintf is possible, and maybe it would be even better, but I don't want to break something. Is this formatting is really required?

cocos/2d/CCFontAtlasCache.cpp
```diff
@@ -115,7 +112,7 @@ FontAtlas* FontAtlasCache::getFontAtlasFNT(const std::string& fontFileName, cons
         }
     }
     else
-        return it->second;
+        return _atlasMap[atlasName];
```
Dimon4eg: > return _atlasMap[atlasName];

Why you changed this?

Dimon4eg (changes requested):

Dimon4eg:
In general looks ok.
About `stringstream`, better to avoid it because it's slow.

cocos/2d/CCFontAtlasCache.cpp
sbrednikhin: Looks like I was not very attentive, thank you for pointing. I'll change that immediately.

Dimon4eg (approved):

Dimon4eg:
Thank you

cocos/2d/CCFontAtlasCache.cpp
```diff
@@ -60,15 +60,11 @@ FontAtlas* FontAtlasCache::getFontAtlasTTF(const _ttfConfig* config)
...
-    if (useDistanceField) {
-        snprintf(tmp, ATLAS_MAP_KEY_BUFFER, "df %.2f %d %s", config->fontSize, config->outlineSize,
-                 realFontFilename.c_str());
-    } else {
-        snprintf(tmp, ATLAS_MAP_KEY_BUFFER, "%.2f %d %s", config->fontSize, config->outlineSize,
-                 realFontFilename.c_str());
-    }
-    std::string atlasName = tmp;
+    std::string key;
+    char key_prefix[ATLAS_MAP_KEY_PREFIX_BUFFER_SIZE];
```
minggo: key_prefix -> keyPrefix?

minggo: As other variables uses this style.

cocos/2d/CCFontAtlasCache.cpp
```diff
@@ -147,7 +144,7 @@ FontAtlas* FontAtlasCache::getFontAtlasCharMap(const std::string& plistFile)
 
 FontAtlas* FontAtlasCache::getFontAtlasCharMap(Texture2D* texture, int itemWidth, int itemHeight, int startCharMap)
 {
-    char tmp[30];
+    char tmp[ATLAS_MAP_KEY_PREFIX_BUFFER_SIZE];
```
minggo: tmp -> keyPrefix?

minggo (approved):

minggo:
it is synchronized to metal-support branch in https://github.com/cocos2d/cocos2d-x/pull/19593

PR #19384, opened 2019-02-05
Font atlas texture initialization delayed

sbrednikhin:
If you will try to create CCLabel with True Type font, enabling it's outline and change it's font size with separate commands you will create multiple textures for font atlases, one for label second for label with outline, and third for label with another font size. 
Moving texture initialization from constructor and calling it later, right before usage will cause large memory economy, because even if atlases will be created they won't create textures of allocate memory for current page data.


minggo:
It looks good to me, but i am not sure if this function will be invoked before used. @PatriceJiang please take a look.

PatriceJiang (approved):

minggo:
it is synchronized to metal-support branch in https://github.com/cocos2d/cocos2d-x/pull/19592

PR #19385, opened 2019-02-06
feature to be able to define function to generate the sprite frame name

benjaprieto:
This feature, adds the posibility to set a function to generate the sprite frame name. For example to identify the original plist, or follow a certain naming convention

### DIFF
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -50,6 +50,8 @@ NS_CC_BEGIN
 
 static SpriteFrameCache *_sharedSpriteFrameCache = nullptr;
 
+SpriteFrameNameGenerator SpriteFrameCache::_spriteFrameNameGenerator = nullptr;
+
 SpriteFrameCache* SpriteFrameCache::getInstance()
 {
     if (! _sharedSpriteFrameCache)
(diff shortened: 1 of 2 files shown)


cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -61,6 +63,11 @@ SpriteFrameCache* SpriteFrameCache::getInstance()
     return _sharedSpriteFrameCache;
 }
 
+void SpriteFrameCache::SetSpriteFrameNameGenerator(SpriteFrameNameGenerator generator)
+{
+    SpriteFrameCache::_spriteFrameNameGenerator = generator;
```
gestern, line 68: Okey... http://www.stroustrup.com/4th.html

cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -50,6 +50,8 @@ NS_CC_BEGIN
 
 static SpriteFrameCache *_sharedSpriteFrameCache = nullptr;
 
+SpriteFrameNameGenerator SpriteFrameCache::_spriteFrameNameGenerator = nullptr;
```
gestern: See below

cocos/2d/CCSpriteFrameCache.cpp
benjaprieto: Lol you are so funny!

minggo:
It is not a good idea to do it like this, you should change the plist file.

PR #19400, opened 2019-02-16
Fix memory leak if invoking Texture2D::setAlphaTexture many times.

dumganhar:


cocos/renderer/CCTexture2D.cpp
```diff
@@ -1499,10 +1499,12 @@ void Texture2D::removeSpriteFrameCapInset(SpriteFrame* spriteFrame)
...
 void Texture2D::setAlphaTexture(Texture2D* alphaTexture)
 {
-    if (alphaTexture != nullptr) {
-        this->_alphaTexture = alphaTexture;
-        this->_alphaTexture->retain();
-        this->_hasPremultipliedAlpha = true; // PremultipliedAlpha should be true.
+    if (alphaTexture != nullptr)
+    {
+        alphaTexture->retain();
+        CC_SAFE_RELEASE(_alphaTexture);
```
dumganhar, line 1505: Old alpha texture needs to be released.

minggo:
it is synchronized to metal-support in https://github.com/cocos2d/cocos2d-x/pull/19590

PR #19411, opened 2019-02-19
Support new Mali GPUs

AIGRIND-LLC:
Some new Mali GPUs use FP16 for matrix that is not precise enough. Should be FP32. Tested on Mali-T880.

The same issue is described here:
https://discuss.cocos2d-x.org/t/critical-issue-regarding-rendering-on-devices-with-renderer-type-mali/45568


minggo:
I think we can change the default precision to `highp`.

AIGRIND-LLC:
@minggo I have tried to get rid of `#ifdef` and got an error while running on Win32:

```
cocos2d: ERROR: Failed to compile shader:...
cocos2d: 0(1) : error C0000: syntax error, unexpected type identifier, expecting '{' at token "mat4"
```
Only OpenGL ES supports such precision modifiers.
I could not find official documentation to clarify that but found [this one](https://github.com/libgdx/libgdx/wiki/Opengl-es-support#precision-modifiers).
Moreover [the same approach](https://github.com/cocos2d/cocos2d-x/blob/4db6353067a618300c8c29a800bd5f9fa83df405/cocos/editor-support/spine/SkeletonTwoColorBatch.cpp#L115) is already used in cocos2dx code.

And desktop OpenGL uses high precision by default.

minggo:
@AIGRIND-LLC what i mean is that, we change it [here](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/renderer/CCGLProgram.cpp#L495).

AIGRIND-LLC:
@minggo Now I see what you meant.

This issue has already been fixed in #19633. And now it conflicts with mine.
I think that my solution is more effective for pixel shaders.
So could you please take a look and make a proposal.

minggo:
I have a question. As you use different precision in vertex shader and fragment shader, what happen for the uniform or varying variables that are both defined in vertex shader and fragment shader.

AIGRIND-LLC:
@minggo actually this is default behaviour according the specification.
Here is the [long explanation](https://stackoverflow.com/a/6336285).
And the short one is that there is the [same approach already used](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/renderer/CCGLProgram.cpp#L494) for iOS.

minggo:
@AIGRIND-LLC thanks, could you please fix conflict?

AIGRIND-LLC:
Sure. Done!

minggo (approved):

PR #19415, opened 2019-02-20
Remove gl codes

minggo:


minggo:
The engine doesn't have cocostudio test cases, so may not work correctly.

PR #19419, opened 2019-02-20
fix the member parameter:_percent is always be 0 during the update pr…

EricDDK:
The _percent is always 0 during update process. I use event:getPercent() will get 0 always.


minggo:
it is synchronized to metal-support branch in https://github.com/cocos2d/cocos2d-x/pull/19591

PR #19422, opened 2019-02-21
Feature/save image

Mee-gu:
save image


Mee-gu:
add synchronized callback command to deal with image saving issues

cocos/platform/CCPlatformConfig.h
```diff
@@ -110,6 +110,9 @@ THE SOFTWARE.
 #endif
 
 #if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC) || (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_MAC)
+    #define CC_USE_MAC
+#endif
```
minggo, line 115: Don't have need this macro.

Mee-gu: It uses in TextureMTL.mm, since synchronizeResource only available in macos.

PR #19424, opened 2019-02-22
need to generate binding codes for all targets

drelaptop:
1. Like https://github.com/cocos2d/cocos2d-x/pull/19421 did, when interface changed, the travis task need to generate binding codes firstly, and then build.
2. Or build task will failed, for bindings issues, refer to https://travis-ci.org/cocos2d/cocos2d-x/builds/496361752?utm_source=github_status&utm_medium=notification

### DIFF
--- a/tools/travis-scripts/before-install.sh
+++ b/tools/travis-scripts/before-install.sh
@@ -103,6 +103,8 @@ fi
 if [ "$BUILD_TARGET" == "linux_cocos_new_test" ]; then
     download_deps
     install_linux_environment
+    sudo python -m pip install retry
+    # set android ndk environment by setup_android.py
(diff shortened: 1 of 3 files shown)


tools/travis-scripts/before-install.sh
```diff
@@ -103,6 +103,7 @@ fi
 if [ "$BUILD_TARGET" == "linux_cocos_new_test" ]; then
     download_deps
     install_linux_environment
+    python $COCOS2DX_ROOT/tools/appveyor-scripts/setup_android.py --ndk_only
```
minggo, line 108: Why travis invoke appveyor related script?

drelaptop: that's a pure python script to setup android environment

PR #19442, opened 2019-02-27
[3d] terrain impl & test case

PatriceJiang:


cocos/3d/CCTerrain.h
```diff
@@ -475,6 +483,9 @@ class CC_DLL Terrain : public Node
     
     Chunk * getChunkByIndex(int x,int y) const;
 
+    void onBeforeDraw();
+    void onAfterDraw();
+
```
minggo: should be public?

cocos/3d/CCTerrain.h
```diff
@@ -501,23 +511,38 @@ class CC_DLL Terrain : public Node
...
-    GLint _lightMapLocation;
-    GLint _detailMapSizeLocation[4];
-    GLint _lightDirLocation;
-    RenderState::StateBlock _stateBlock;
-
+    struct StateBlock {
+       // bool blend;
+        bool depthWrite;
+        bool depthTest;
+        backend::CullMode cullFace;
```
minggo: it is better to initialize them

PR #19444, opened 2019-02-27
Lua: addCustomEventListener refactoring

ghost:


ghost:
test this please.

minggo:
Sorry, i am not familiar with these codes. What's the test case will test the codes?

ghost:
testcase:
https://github.com/cocos2d/cocos2d-x/blob/v3/tests/lua-tests/src/NewEventDispatcherTest/NewEventDispatcherTest.lua#L618-L634

ghost:
anyone help to check this pr ?

zhongfq:
this code maybe crash in sometimes, because the variable tolua_S in callback maybe an invalid value if it is a lua coroutine after lua gc.

zhongfq:
when i check lua binding in cocos2dx, there are a lot of codes like this, it may leads to use in risk

PR #19450, opened 2019-02-28
Limited python 3 support

newnon:
support python 3 in download-deps.py and setup.py

### DIFF
--- a/download-deps.py
+++ b/download-deps.py
@@ -69,14 +69,6 @@ def delete_folder_except(folder_path, excepts):
             os.remove(full_path)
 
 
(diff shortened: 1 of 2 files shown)


minggo:
Thanks for the PR, but as cocos console doesn't support python3, so i think it is not suitable to merge this PR. And there are many other python scripts used in engine.

vperilla:
IMHO I think you can seriously consider support python 3, because python 2 is near end support
https://pythonclock.org/

Escapingbug:
Python2 has already retired. So maybe we should consider change to python3 in some way?

Issue #19456, opened 2019-03-03
Android FileUtils::listfiles 3.17.1 not working


imtrobin:
- cocos2d-x version: 3.17.1
- devices test on: Multiple
- developing environments
   - NDK version:  19

This fails to get any files

`vector <string> files = fileutil->listFiles (fileutil->getDefaultResourceRootPath ());`

imtrobin:
Fixed

PR #19457, opened 2019-03-04
[Android] fix FileUtils::listFiles with path 'assets/'

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19456

### DIFF
--- a/cocos/platform/android/CCFileUtils-android.cpp
+++ b/cocos/platform/android/CCFileUtils-android.cpp
@@ -300,7 +300,7 @@ long FileUtilsAndroid::getFileSize(const std::string& filepath) const
 std::vector<std::string> FileUtilsAndroid::listFiles(const std::string& dirPath) const
 {
 
-    if(isAbsolutePath(dirPath)) return FileUtils::listFiles(dirPath);
+    if(!dirPath.empty() && dirPath[0] == '/') return FileUtils::listFiles(dirPath);
 
     std::vector<std::string> fileList;
     string fullPath = fullPathForDirectory(dirPath);
--- a/tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
+++ b/tests/cpp-tests/Classes/FileUtilsTest/FileUtilsTest.cpp
@@ -1408,10 +1408,24 @@ void TestListFiles::onEnter()
     this->addChild(cntLabel);
     cntLabel->setPosition(winSize.width / 2, winSize.height / 3);
     // writeTest
-    auto list = FileUtils::getInstance()->listFiles("fonts");
+    std::vector<std::string> listFonts = FileUtils::getInstance()->listFiles("fonts");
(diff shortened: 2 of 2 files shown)


cocos/platform/android/CCFileUtils-android.cpp
```diff
@@ -300,7 +300,7 @@ long FileUtilsAndroid::getFileSize(const std::string& filepath) const
 std::vector<std::string> FileUtilsAndroid::listFiles(const std::string& dirPath) const
 {
 
-    if(isAbsolutePath(dirPath)) return FileUtils::listFiles(dirPath);
+    if(!dirPath.empty() && dirPath[0] == '/') return FileUtils::listFiles(dirPath);
```
PatriceJiang, line 303:
```cpp
bool FileUtilsAndroid::isAbsolutePath(const std::string& strPath) const
{
    DECLARE_GUARD;
    // On Android, there are two situations for full path.
    // 1) Files in APK, e.g. assets/path/path/file.png
    // 2) Files not in APK, e.g. /data/data/org.cocos2dx.hellocpp/cache/path/path/file.png, or /sdcard/path/path/file.png.
    // So these two situations need to be checked on Android.
    if (strPath[0] == '/' || strPath.find(_defaultResRootPath) == 0)
    {
        return true;
    }
    return false;
}
```

`_defaultResRootPath` is treated as 'absolute path'

imtrobin:
I applied this line to my code, and it crashes, I think it is infinite stack overflow. If I call it with a path like "/testfolder/", it will be endless loop.

PatriceJiang:
I added two cases 
```cpp
    auto sdcard1 = FileUtils::getInstance()->listFiles ("/sdcard/");
    auto sdcard2 = FileUtils::getInstance()->listFiles ("/sdcard/xxxshoulnotexists");
```
they returns.  
@imtrobin  do I miss something?

imtrobin:
```
if (!fileutil->isAbsolutePath (plistpath))
	plistpath = fileutil->getDefaultResourceRootPath () + "theme/" + plistpath;
vector <string> filelist = fileutil->listFiles (plistpath);

```
My plistpath is "gamecloud/modernlivingroom/animation" so it became 

assets/theme/gamecloud/modernlivingroom/animation

 when I did a debug print. When you see the code, if I pass in "/sdcard/", it should be infinite loop, no?

imtrobin:
sorry, my bad. The above fix works. It found the files, but fullPathForFilename didn't, the crash was happening there. Consider this fixed

imtrobin:
So the windows version will return you the full path, while the android version only returns you the file itself. So when I try to addSpriteFramesWithFile, it is not found on Android, while it works on win32.

imtrobin:
To add on to above, the listfiles returns only the filename for items in /assets, for things in /data folder, it returns the fullpath fine.

minggo:
it is synchronized to metal-support in https://github.com/cocos2d/cocos2d-x/pull/19588

PR #19476, opened 2019-03-07
Fix bug: can't play video in obbfile.

halx99:


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxVideoView.java
```diff
@@ -307,7 +307,11 @@ private void openVideo() {
             mDuration = -1;
             mCurrentBufferPercentage = 0;
             if (mIsAssetRouse) {
-                AssetFileDescriptor afd = mCocos2dxActivity.getAssets().openFd(mVideoFilePath);
+                AssetFileDescriptor afd = null;
+                if (Cocos2dxHelper.getObbFile() != null) {
+                    afd = Cocos2dxHelper.getObbFile() .getAssetFileDescriptor(mVideoFilePath);
+                }
+                if(afd == null) afd = mCocos2dxActivity.getAssets().openFd(mVideoFilePath);
```
minggo: Why check OBB file first? Is using OBB file more common?

halx99: It doesn't matter what order check.

minggo: But check assets may make it more faster if video files are more commonly stored in assets.

halx99: yes, you are right.

halx99: @minggo I change the check order, please review

minggo (approved):

minggo:
it is synchronized to metal-support branch in https://github.com/cocos2d/cocos2d-x/pull/19586

PR #19477, opened 2019-03-08
[bugfix] add mesh command callbacks

PatriceJiang:
Commands are reordered after `renderer->addCommand(*)`,  `CallbackCommand` does not guarantee to act as a callback. 

Add `before` and `after` hooks to `MeshCommand`

### DIFF
--- a/cocos/renderer/CCMeshCommand.h
+++ b/cocos/renderer/CCMeshCommand.h
@@ -57,6 +57,8 @@ public:
     */
     using IndexFormat = backend::IndexFormat;
 
+    typedef std::function<void()> CallBackFunc;
+
     MeshCommand();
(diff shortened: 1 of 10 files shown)


cocos/renderer/CCMeshCommand.h
```diff
@@ -66,12 +68,23 @@ class CC_DLL MeshCommand : public CustomCommand
...
+
 #if CC_ENABLE_CACHE_TEXTURE_DATA
     void listenRendererRecreated(EventCustom* event);
 #endif
 
 protected:
 
+    CallBackFunc _beforeCallback;
+    CallBackFunc _afterCallback;
+
```
minggo, line 93: Should initialize them.

PR #19478, opened 2019-03-08
make travis work

minggo:


minggo:
don't compile iOS because:

* simulator doesn't support metal
* needs developer team id to compile for iOS device

PR #19489, opened 2019-03-12
Update download-deps.py

CCbird:
Fixed some ssl authentication errors that occurred at some time


minggo (approved):

minggo:
It seems the codes breaks the CI.

Issue #19334, opened 2019-01-22
Texture2D::setSamplerDescriptor() -> Texture2D::setTexParams


minggo:
It is better not to change the function  name.

PR #19498, opened 2019-03-12
Texture2D::TexParams fields use backend enum types

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19334

### DIFF
--- a/cocos/renderer/CCTexture2D.cpp
+++ b/cocos/renderer/CCTexture2D.cpp
@@ -908,19 +908,7 @@ Texture2D* Texture2D::getAlphaTexture() const
     return _alphaTexture;
 }
 
-void Texture2D::setTexParameters(const Texture2D::TexParams &params)
-{
-    backend::SamplerDescriptor sd;
-    sd.mipmapEnabled    = true;
-    sd.mipmapFilter     = backend::SamplerFilter::DONT_CARE;
-    sd.minFilter        = params.minFilter;
-    sd.magFilter        = params.magFilter;
-    sd.sAddressMode     = params.wrapS;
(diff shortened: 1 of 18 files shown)


minggo:
But `Texture2D::setSamplerDescriptor()` is not deleted.

PatriceJiang:
alias `backend::SamplerDescriptor` as `Texture2D::TexParams`

cocos/renderer/backend/Types.h
```diff
@@ -188,21 +188,32 @@ enum class ColorWriteMask: uint32_t
 
 struct SamplerDescriptor
 {
-    bool mipmapEnabled = false;
+    bool mipmapEnabled = true;
```
minggo: Is the default value need to be true?

cocos/renderer/CCTexture2D.cpp
```diff
@@ -707,21 +707,20 @@ void Texture2D::setAliasTexParameters()
...
-
-    setSamplerDescriptor(descriptor);
+    _texture->updateSamplerDescriptor(descriptor);
 }
 
 void Texture2D::setAntiAliasTexParameters()
 {
 
-    backend::SamplerDescriptor descriptor(false,
+    backend::SamplerDescriptor descriptor(true,
```
PatriceJiang:
```suggestion
    backend::SamplerDescriptor descriptor(false,
```

cocos/renderer/backend/Types.h
PatriceJiang:
```suggestion
    bool mipmapEnabled = false;
```

minggo (approved):

PR #19514, opened 2019-03-14
[Feature] fix some TODO

Mee-gu:
fix some TODO

### DIFF
--- a/cocos/renderer/backend/metal/TextureMTL.mm
+++ b/cocos/renderer/backend/metal/TextureMTL.mm
@@ -139,7 +139,7 @@ void TextureMTL::updateSubData(unsigned int xoffset, unsigned int yoffset, unsig
     
     // metal doesn't generate mipmaps automatically, so should generate it manually.
     if (_isMipmapEnabled)
-        Utils::generateMipmaps(_mtlTexture);
+        generateMipmaps();
 }
 
 void TextureMTL::createTexture(id<MTLDevice> mtlDevice, const TextureDescriptor& descriptor)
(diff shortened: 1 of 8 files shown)


cocos/renderer/backend/metal/TextureMTL.mm
```diff
@@ -376,5 +387,13 @@ bool convertData(uint8_t* src, unsigned int length, TextureFormat format, uint8_
     [commandBuffer commit];
 }
 
+void TextureCubeMTL::generateMipmaps()
+{
+    if (TextureUsage::RENDER_TARGET == _textureUsage)
+        return;
+    
+    Utils::generateMipmaps(_mtlTexture);
+    _isMipmapEnabled = true;
```
minggo: There is an optimization: don't have to generate mipmaps if it is generated before. Of course, if the data is updated, then need to re-generate mipmap again.

cocos/renderer/backend/metal/TextureMTL.mm
```diff
@@ -317,6 +347,12 @@ bool convertData(uint8_t* src, unsigned int length, TextureFormat format, uint8_
 void TextureCubeMTL::updateSamplerDescriptor(const SamplerDescriptor &sampler)
 {
     createSampler(_mtlDevice, sampler);
+    bool needGenerateMipmap = !_isMipmapEnabled && sampler.mipmapEnabled;
+    _isMipmapEnabled = sampler.mipmapEnabled;
+    if(needGenerateMipmap)
+    {
+        generateMipmaps();
+    }
```
minggo, line 355: It seems the logic is duplicated in all Textures(TextureGL, TextureMTL, TextureCubeGL, TextureCubeMTL), can put the codes in Texture.

PR #19519, opened 2019-03-18
[Feature] fix mipmap issues

Mee-gu:
fix mipmap issues

### DIFF
--- a/cocos/renderer/backend/metal/TextureMTL.mm
+++ b/cocos/renderer/backend/metal/TextureMTL.mm
@@ -78,6 +78,24 @@ namespace
         }
         return converted;
     }
+    
+    bool isColorRenderable(TextureFormat textureFormat)
+    {
+        switch (textureFormat)
+        {
+            case TextureFormat::R8G8B8A8:
+            case TextureFormat::R8G8B8:
+            case TextureFormat::RGBA4444:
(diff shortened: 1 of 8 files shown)


cocos/renderer/backend/metal/TextureMTL.mm
```diff
@@ -148,7 +168,7 @@ bool convertData(uint8_t* src, unsigned int length, TextureFormat format, uint8_
            [MTLTextureDescriptor texture2DDescriptorWithPixelFormat:Utils::toMTLPixelFormat(descriptor.textureFormat)
                                                               width:descriptor.width
                                                              height:descriptor.height
-                                                          mipmapped:TRUE];
+                                                          mipmapped:YES];
```
minggo, line 171: If passing `YES` to mipmapped, then should we do mipmap manually too?

Mee-gu: The `mipmapped` parameter is there to tell Metal to calculate the number of mipmap levels that resulting image will have. And `generateMipmapsForTexture` method is used to generate mipmaps for a texture that already has mipmap levels.

minggo:
please fix conflict

minggo (approved):

PR #19522, opened 2019-03-18
[tolua] make genbinding.py to work with newer versions of NDK

PatriceJiang:


tools/tolua/genbindings.py
```diff
@@ -37,6 +37,52 @@ def _check_python_bin_env():
...
+    if os.path.isfile(dir):
+        if os.path.basename(dir) == fn:
+            return os.path.join(os.path.dirname(dir), fn)
+        else :
+            return None
+    elif os.path.isdir(dir):
+      for subdir in os.listdir(dir):
+          searchPath = _find_first_file_in_dir(os.path.join(dir, subdir), fn)
+          if searchPath is not None:
+              return searchPath
```
minggo, line 50: what happen if searchPath is None?

tools/tolua/genbindings.py
```diff
@@ -37,6 +37,52 @@ def _check_python_bin_env():
...
+    versionFile = _find_first_file_in_dir(_check_ndk_root_env(), "AndroidVersion.txt")
+    if versionFile is None:
+        return ""
+    versionDir = os.path.dirname(versionFile)
+    includeDir = _find_first_file_in_dir(versionDir, "stdarg.h")
+    llvmIncludePath = os.path.dirname(includeDir)
+    return "-I"+llvmIncludePath
+  
+
+def _defaultIncludePath():
```
minggo, line 88: please add comment for the function

tools/tolua/genbindings.py
PatriceJiang: the function will return None by default behavior

minggo (approved):

PR #19524, opened 2019-03-19
Ability to change title position

gestern:


minggo (approved):

minggo:
it is synchronized to metal-support branch in https://github.com/cocos2d/cocos2d-x/pull/19587

PR #19536, opened 2019-03-20
control key pressed event support for ui::TextFiled

sbrednikhin:
support of control key pressed event enabled for ui::TextField. 
New event codes added for deleting symbol by 'del' key an changing position of cursor.


sbrednikhin:
I really have no idea why it doesn't passed AppVeyor build check. For me it looks just like a problem on build machine that is not connected with fix.
Am I right?

sbrednikhin:
Ok. So it was spontaneous check fail!

PR #19538, opened 2019-03-21
Feature/remove poll events

Mee-gu:
remove pollEvents due to crashed in CaptureNode

### DIFF
--- a/cocos/platform/mac/CCApplication-mac.mm
+++ b/cocos/platform/mac/CCApplication-mac.mm
@@ -83,7 +83,6 @@ int Application::run()
         lastTime = getCurrentMillSecond();
         
         director->mainLoop();
-//        glview->pollEvents();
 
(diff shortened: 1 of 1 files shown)


cocos/platform/mac/CCApplication-mac.mm
```diff
@@ -83,7 +83,6 @@ static long getCurrentMillSecond()
         lastTime = getCurrentMillSecond();
         
         director->mainLoop();
-        glview->pollEvents();
```
minggo, line 86: Should check other platforms too.

Mee-gu: It's available only on mac

minggo: I mean should check the implementation of windows and linux.

Issue #19555, opened 2019-03-27
ui::Slider sends ON_PERCENTAGE_CHANGED event even if no change in percentage


rh101:
- cocos2d-x version: 3.17.1
- devices test on: N/A
- developing environments
   - NDK version: N/A
   - Xcode version:  N/A
   - VS version: N/A
   - browser type and version:  N/A

Steps to Reproduce:

1. Add a ui::Slider node to the scene, and register for the percentage change event:

```
    auto size = Director::getInstance()->getWinSize();
    auto slider = ui::Slider::create();
    slider->setTouchEnabled(true);
    slider->loadBarTexture("cocosui/sliderTrack.png");
    slider->loadSlidBallTextures("cocosui/sliderThumb.png", "cocosui/sliderThumb.png", "");
    slider->loadProgressBarTexture("cocosui/sliderProgress.png");
    slider->setPosition(Vec2(size.width / 2.0f, size.height / 2.0f));
    slider->setPercent(52);
    slider->addEventListener(CC_CALLBACK_2(SliderTest::sliderEvent, this));
    addChild(slider, 999);
```

```
void SliderTest::sliderEvent(Ref *pSender, ui::Slider::EventType type)
{
    if (type == Slider::EventType::ON_PERCENTAGE_CHANGED)
    {
            auto slider = dynamic_cast<ui::Slider*>(pSender);
            CCLOG(">>> LineOptionIndex: %d", slider->getPercent());
    }
}
```

2.  Run the test, and change the slider position, and you will notice in the console output that the percentage change event is sent many times, even if there is no change in the percentage value.  

In UISlider.cpp, it should be updated to only broadcast that event if and only if the percentage value has changed, which would be literally what "ON_PERCENTAGE_CHANGED" means.  

Also, it should be sent from inside the setPercent() method, and not from the onTouch*** handlers, because if setPercent() is called manually, a broadcast should still be sent out, in case there is a percentage changed handler registered to do something on any change in the value.

```
void Slider::setPercent(int percent)
{
    if (percent > _maxPercent)
    {
        percent = _maxPercent;
    }
    if (percent < 0)
    {
        percent = 0;
    }

    // Only send event if value has changed
    if (_percent != percent)
    {
        _percent = percent;
        updateVisualSlider();
        percentChangedEvent(EventType::ON_PERCENTAGE_CHANGED);
    }
}

// New method to update the visual elements
void Slider::updateVisualSlider()
{
[...]

PR #19556, opened 2019-03-27
Only send ON_PERCENTAGE_CHANGED when value has changed

rh101:
ref #19555 

This prevents the ON_PERCENTAGE_CHANGED from being sent out constantly even if the value has not been changed if the slider is being touched.


minggo (approved):

minggo:
@PatriceJiang please take a look too.

PatriceJiang (approved):

minggo:
it is synchronized to metal-branch in https://github.com/cocos2d/cocos2d-x/pull/19589

PR #19564, opened 2019-04-01
[bugfix] motionstreak & DrawNode3D on GL

PatriceJiang:
`updateBuffer` should be called before `updateSubData`

### DIFF
--- a/cocos/renderer/backend/opengl/BufferGL.cpp
+++ b/cocos/renderer/backend/opengl/BufferGL.cpp
@@ -44,7 +44,19 @@ void BufferGL::updateSubData(void* data, unsigned int offset, unsigned int size)
     if (!_bufferAllocated)
     {
         CCASSERT(offset == 0, "offset should be zero when allocate buffer");
-        updateData(data, size);
+        if (size < _size)
+        {
+            //ensure the size parameter of `updateData` is not less then `_size`
+            uint8_t *zeros = new uint8_t[_size]();
+            updateData(zeros, _size);
+            delete[] zeros;
+            updateSubData(data, offset, size);
+        }
+        else
+        {
+            updateData(data, size);
+        }
+
         return;
     }
 
--- a/cocos/renderer/backend/opengl/BufferGL.cpp
+++ b/cocos/renderer/backend/opengl/BufferGL.cpp
@@ -17,7 +17,7 @@ BufferGL::~BufferGL()
 
 void BufferGL::updateData(void* data, unsigned int size)
 {
-    assert(size);
+    assert(size && size <= _size);
     
     if (_buffer)
     {
@@ -33,7 +33,6 @@ void BufferGL::updateData(void* data, unsigned int size)
         }
         CHECK_GL_ERROR_DEBUG();
         _bufferAllocated = true;
-        _size = size;
     }
 }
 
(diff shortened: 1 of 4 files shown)


cocos/renderer/backend/opengl/BufferGL.cpp
```diff
@@ -44,7 +45,19 @@ void BufferGL::updateSubData(void* data, unsigned int offset, unsigned int size)
     if (!_bufferAllocated)
     {
         CCASSERT(offset == 0, "offset should be zero when allocate buffer");
-        updateData(data, size);
+        if (size < _size)
+        {
+            //ensure the size parameter of `updateData` is not less then `_size`
+            uint8_t *zeros = new uint8_t[_size]();
+            updateData(zeros, _size);
```
Mee-gu: since `size` is always less or less equal than `_size`, why need to change `_size` in `updateData`

minggo:
metal require to pass buffer size when creating a buffer, and sub data can not exceed the size. OpenGL should have the same behavior too. Or it will have error on metal.

cocos/renderer/backend/opengl/BufferGL.cpp
```diff
@@ -32,7 +32,7 @@ void BufferGL::updateData(void* data, unsigned int size)
             glBufferData(GL_ELEMENT_ARRAY_BUFFER, size, data, GL_STATIC_DRAW);
         }
         CHECK_GL_ERROR_DEBUG();
-        _bufferAllocated = true;
+        _bufferAllocated = size;
```
PatriceJiang, line 35: to track allocated byte count

PatriceJiang:
I have added some assertions to ensure that the size of sub data wouldn't exceed default size.

minggo (approved):

tests/cpp-tests/Classes/Sprite3DTest/DrawNode3D.cpp
```diff
@@ -58,7 +58,15 @@ void DrawNode3D::ensureCapacity(int count)
 {
     CCASSERT(count>=0, "capacity must be >= 0");
     
-    _bufferLines.reserve(_bufferLines.size() + count);
+    const int EXTENDED_SIZE = _bufferLines.size() + count;
+
+    _bufferLines.reserve(EXTENDED_SIZE);
+
+    if (!_customCommand.getVertexBuffer() || _customCommand.getVertexBuffer()->getSize() < (EXTENDED_SIZE * sizeof(_bufferLines[0])))
```
Mee-gu, line 65: Would it better to set `EXTENDED_SIZE` a litter bigger to avoid frequency new/delete buffer?

Mee-gu (approved):

tests/cpp-tests/Classes/Sprite3DTest/DrawNode3D.cpp
PatriceJiang: Yes, we can increate the size argument in `createVertexBuffer`.

tests/cpp-tests/Classes/Sprite3DTest/DrawNode3D.cpp
```diff
@@ -58,7 +58,15 @@ void DrawNode3D::ensureCapacity(int count)
...
+    {
+        _customCommand.createVertexBuffer(sizeof(V3F_C4B), EXTENDED_SIZE, CustomCommand::BufferUsage::DYNAMIC);
```
PatriceJiang:
```suggestion
        _customCommand.createVertexBuffer(sizeof(V3F_C4B), EXTENDED_SIZE + (EXTENDED_SIZE >> 1), CustomCommand::BufferUsage::DYNAMIC);
```

PR #19566, opened 2019-04-01
[Feature] colorAttachment and depth/stencil attachment width and heig…

Mee-gu:
color attachment and depth/stencil attachment width and height should be the same.


minggo (approved):

minggo:
The issue title seems different from modification.

minggo:
And is there any issue created for this PR? If so, please connect them.

Mee-gu:
In original depth/stencil attachment will always be set regardless of whether they are enabled. This may cause color attachment texture's width and height different with depth/stencil attachment texture. It will cause validation to fail when setting scissor rectangle to the encoder.

PR #19567, opened 2019-04-01
[windows] use PostMessage to replace SendMessage

PatriceJiang:
- speedup font loading 
- make CCLOG non-block

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -108,10 +108,12 @@ namespace {
         if (Director::getInstance()->getOpenGLView())
         {
             HWND hwnd = Director::getInstance()->getOpenGLView()->getWin32Window();
-            SendMessage(hwnd,
-                        WM_COPYDATA,
-                        (WPARAM)(HWND)hwnd,
-                        (LPARAM)(LPVOID)&myCDS);
+            // use non-block version of SendMessage 
+            SendNotifyMessageA(hwnd,
+                WM_COPYDATA,
+                (WPARAM)(HWND)hwnd,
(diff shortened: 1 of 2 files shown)


cocos/base/CCConsole.cpp
```diff
@@ -108,10 +108,12 @@ namespace {
         if (Director::getInstance()->getOpenGLView())
         {
             HWND hwnd = Director::getInstance()->getOpenGLView()->getWin32Window();
-            SendMessage(hwnd,
-                        WM_COPYDATA,
-                        (WPARAM)(HWND)hwnd,
-                        (LPARAM)(LPVOID)&myCDS);
+            // use non-block version of SendMessage
```
crazyhappygame, line 111: Could you double check docs:
https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-sendnotifymessagea
Are you sure about SendNotifyMessageA?
"If the window was created by the calling thread, SendNotifyMessage calls the window procedure for the window and does not return until the window procedure has processed the message"
Maybe PostMessage should be used?

PatriceJiang: `PostMessage` seems more "non-block" than `SendNotifyMessage`, thanks for your correcting .

minggo:
Could you please describe why it can speedup and not block the engine?

PatriceJiang:
Since `SendMessage` is blocking, when loading font,  `BitmapDC::setFont` sometimes takes `500ms+` to wait the invoke result. 
`PostMessage` does not wait, that makes it run faster.

minggo:
Please send a PR to v3 too.

PatriceJiang:
Okay

PR #19574, opened 2019-04-03
[bugfix] cpp-test small bugs

PatriceJiang:


cocos/renderer/backend/opengl/CommandBufferGL.cpp
```diff
@@ -522,6 +522,9 @@ void CommandBufferGL::setLineWidth(float lineWidth)
 {
     if(lineWidth > 0.0f)
         glLineWidth(lineWidth);
+    else
+        glLineWidth(1.0f);
+
```
minggo, line 527: it is strange, `1.f > 0.0f`, so will be set, why need to add this line

PatriceJiang: Default line width of some commands are `0`, that make `glLineWidth` remain unchanged if `setLineWidth` is not called, which state is wrong.

minggo: So this codes changed to
```c++
if (lineWidth >= 0.f)
    glLineWidth(lineWidth);
```
?

Mee-gu: The default width is 1:0 in OpengGL. Values less than or equal to 0:0 generate the
error INVALID_VALUE.

Mee-gu (approved):

cocos/renderer/backend/opengl/CommandBufferGL.cpp
minggo: Then it is the error of our codes to pass 0?

PR #19578, opened 2019-04-06
Update UIDeprecated.cpp

larisand:
There is no need of this last "}".


cocos/ui/UIDeprecated.cpp
```diff
@@ -31,4 +31,4 @@ namespace ui {
     
 const Margin MarginZero =  Margin::ZERO ;
     
-}}
```
minggo, line 34: NS_CC_BEGIN introduces `{`, so i think it is better to use `NS_CC_END` instead of `}` here.

PR #19579, opened 2019-04-06
Update CCArmature.cpp

larisand:
There is an unmatched ");" in the file at line 720.


minggo:
Good catch, but it is wired that it works without the modification.

PR #19596, opened 2019-04-10
fix conflict

minggo:


minggo:
sync https://github.com/cocos2d/cocos2d-x/pull/19297

PR #19598, opened 2019-04-10
Add clear request and responses method to httpclient (#19294)

minggo:
* Added functionality to clear pending responses and requests in the http client

* Responses and Requests now are processed in similar way

* Fix for clearing the http requests

* Added tests for HttpClient::clearResponseAndRequestQueue


minggo:
sync https://github.com/cocos2d/cocos2d-x/pull/19294

PR #19608, opened 2019-04-16
add build test option for cmake

deflinhec:
1. Add optional toggle for building libraries only.


cmake/Modules/CocosBuildSet.cmake
```diff
@@ -24,6 +24,7 @@ message(STATUS "ENGINE_BINARY_PATH:" ${ENGINE_BINARY_PATH})
 
 # the default behavior of build module
 option(DEBUG_MODE "Debug or Release?" ON)
+option(BUILD_TESTS "Build tests" ON)
```
minggo: CocosBuildSet.cmake is included in new project, and new project doesn't have tests and doesn't need to build tests. So it is not suitable to add this variable here.

deflinhec: Oh I did not notice that, I manage to add it in CMakeLists.txt under cocos2d-x root directory.

PR #19611, opened 2019-04-16
Make precheader files PRIVATE, otherwise they leak to the game project

Stinkfist0:
Make precheader files PRIVATE, otherwise they leak to the game project (e.g. in Visual Studio, Internal\cocos2d\precheader.cpp visible/duplicate also in MyGame project).

### DIFF
--- a/cocos/CMakeLists.txt
+++ b/cocos/CMakeLists.txt
@@ -164,7 +164,7 @@ endif()
 
 if(WINDOWS)
     # precompiled header. Compilation time speedup ~4x.
-    target_sources(cocos2d PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/precheader.cpp")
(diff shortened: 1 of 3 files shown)


Stinkfist0:
Mac build seems to time out for some reason.

Stinkfist0:
Retriggering the CI build, hopefully Mac build doesn't time out this time.

minggo:
why it can cause memory leak?

Stinkfist0:
By leak I mean that the engine's PCH file is visible and duplicate in the game project files in Visual Studio, causing it to be easily edited by accident. E.g. let say I would use the Boost library in my game and would like to add Boost to the PCH in order to speed up the compilation. Currently (without private precheader.h/cpp) I could very easily end up editing the engine's PCH and cause unnecessary overhead and engine code's recompilation as the engine doesn't use Boost.

PR #19613, opened 2019-04-17
prefer delegate constructor over new(this)

JohnCoconut:
Use delegate constructor rather than `new(this)Downloader(hints);`


minggo:
Cool, i don't know the feature before.

PR #19614, opened 2019-04-17
Optimize calls to std::string::find() and friends

JohnCoconut:
https://clang.llvm.org/extra/clang-tidy/checks/performance-faster-string-find.html

Optimize calls to std::string::find() and friends when the needle passed is a single character string literal. The character literal overload is more efficient.

Examples:

```cpp
str.find("A");

// becomes

str.find('A');
```


minggo:
thanks

PR #19619, opened 2019-04-18
rewrite parseIntegerList with better performance

JohnCoconut:
In file `cocos/2d/CCSpriteFrameCache.cpp`, the function `SpriteFrameCache::parseIntegerList` calls

```cpp
res[i++] = atoi(string.substr(start, end - start).c_str());
```

in a loop, which is bad for performance, as `substr()` function construct a new string and destruct it in every loop iteration.

The new function use `strtol`, which does not create new string. It's nearly twice as fast as the old one, according to google benchmark.

http://quick-bench.com/JjzXDobBEloaspNT0nOPC7EO9m4

### DIFF
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -79,23 +79,20 @@ SpriteFrameCache::~SpriteFrameCache()
 
 void SpriteFrameCache::parseIntegerList(const std::string &string, std::vector<int> &res)
 {
-    std::string delim(" ");
-
     size_t n = std::count(string.begin(), string.end(), ' ');
-    res.resize(n+1);
-    
-    size_t start  = 0U;
-    size_t end = string.find(delim);
-    
-    int i=0;
-    while (end != std::string::npos)
-    {
-        res[i++] = atoi(string.substr(start, end - start).c_str());
-        start = end + delim.length();
-        end = string.find(delim, start);
-    }
-    
-    res[i] = atoi(string.substr(start, end).c_str());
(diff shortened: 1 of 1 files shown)


cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -79,23 +79,20 @@ SpriteFrameCache::~SpriteFrameCache()
...
+    char *endptr;
+
+    int i = 0;
+    do {
+        long int val = strtol(cStr, &endptr, 10);
+        if (endptr == cStr)
+            return;
+        res[i++] = static_cast<int>(val);
+        cStr = endptr;
+    } while (*endptr != '\0');
```
minggo, line 95: There are two conditions to terminate the while loop:
- `endptr == cStr`
- `*endptr == '\0'`

Does it need tow conditions?

JohnCoconut: I am not sure about the precondition of the string containing list of integers. I added it because I am not sure about the string format.

Does the plist file guarantee that the string  is a list of integers separated by a single space? If so, the `endptr == cStr`  statement is redundant. It can be deleted safely.

However, if the string could be like `" "` (a string with a single space), this loop will run forever without the above condition. A plist file with wrong format may cause the program hang on. 

Or the file have been sanitized  somewhere else?

minggo: K, then i think keep it is more safe.

minggo:
Could you please send a PR to `metal-support` branch too?
Thanks.

PR #19627, opened 2019-04-19
Optimize parse integer list

JohnCoconut:
Same PR (https://github.com/cocos2d/cocos2d-x/pull/19619)  as the `v3` branch. The comment from `v3` PR is copied below.

In file cocos/2d/CCSpriteFrameCache.cpp, the function SpriteFrameCache::parseIntegerList calls
```cpp
res[i++] = atoi(string.substr(start, end - start).c_str());
```
in a loop, which is bad for performance, as `substr()` function construct a new string and destruct it in every loop iteration.

The new function use `strtol`, which does not create new string. It's nearly twice as fast as the old one, according to google benchmark.

http://quick-bench.com/JjzXDobBEloaspNT0nOPC7EO9m4


minggo:
Oh, it seems it is a bad PR. You should check out the branch based on `metal-support` branch, then modify it.

minggo:
You can do like this:
- checkout a branch based on `metal-support` branch
- use `git cherry-pick` to check the commit

The commit id may be different from your local commit id, because every PR is merged by rebase.

JohnCoconut:
@minggo 

> The commit id may be different from your local commit id, because every PR is merged by rebase.

Sorry, but I am not sure what you mean. I am unfamiliar with `git cherry-pick`. I just `cherry-picked` a local commit, pushed it to my fork, and made a PR. No conflict now. Am I doing it right?

https://github.com/cocos2d/cocos2d-x/pull/19629

Shall I close this PR if I am doing it correctly. You can close it too if it's ok.

minggo:
The new PR looks good.

PR #19629, opened 2019-04-19
rewrite parseIntegerList with better performance

JohnCoconut:


cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -79,23 +79,20 @@ SpriteFrameCache::~SpriteFrameCache()
...
-    {
-        res[i++] = atoi(string.substr(start, end - start).c_str());
-        start = end + delim.length();
-        end = string.find(delim, start);
-    }
-    
-    res[i] = atoi(string.substr(start, end).c_str());
+    res.resize(n + 1);
+
+    const char *cStr = string.c_str();
```
crazyhappygame, line 85: I like much more version from: https://en.cppreference.com/w/c/string/byte/strtol

    const char *p = string.c_str();
    char *end;
    for (long i = strtol(p, &end, 10);
         p != end;
         i = strtol(p, &end, 10))
    {
        p = end;
        if (errno == ERANGE){
            // error handling????? 
            errno = 0;
        }
        res.push_back(i);
    }

JohnCoconut: @crazyhappygame Thanks for your comment. I like the for loop. It's much cleaner. I always thought of cppreference as a pure C++ reference, so I forget to check it.

I wish the function signature could be changed from

```cpp
void SpriteFrameCache::parseIntegerList(const std::string &string, std::vector<int> &res);
```
to
```cpp
std::vector<int> SpriteFrameCache::parseIntegerList(const std::string &string);
```

The first signature seems a C code be refactored into C++.

However, I am not sure how to handle the error code in the second function signature.

Another question is, does this project provide have api stability? Is it OK to change function signatures?

crazyhappygame: Please change it to:
std::vector<int> SpriteFrameCache::parseIntegerList(const std::string &string);

This member function is protected but I think no-one should override it.

About error code I think that default min/max int is OK.

Additionally:
1. this member function should be changed to free function. It does not use internals of class
2. Consider create unit test

crazyhappygame:
@JohnCoconut I like very much your pull requests. I see that you use clang-tidy and modern C++. Please continue!

cocos/2d/CCSpriteFrameCache.cpp
JohnCoconut: > this member function should be changed to free function. It does not use internals of class

That's what I hope to see too. A free function is certainly better here. 

About the error handling, I will add a `CCLOGWARN` to log a warning message.

minggo: @JohnCoconut keep compatibility if very important in this project.

JohnCoconut: > keep compatibility if very important in this project.

Totally agree. 

@minggo Do you think is it ok to change `parseIntegerList` signature in this case? `parseIneterList` is only used in `CCSpriteFrameCache` class. And no class seems to inherit from `CCSpriteFrameCache` class.

minggo: Yep, it is better, since current signature doesn't have any performance improvement.

PR #19636, opened 2019-04-24
remove redundant user-defined copy constructor and destructor

JohnCoconut:
Remove redundant user-defined copy constructor and destructor, as compiler auto-generated ones are working fine. 

It also helps to get rid of compiler warnings when when compile it on C++11 standard. C++11 has a rule of five. If user defines any one of copy constructor, copy assignment operator, move constructor, move assignment operator, or destructor, all of them need to explicitly written.


minggo (approved):

minggo:
Could you please also send PR to metal-support too?

PR #19637, opened 2019-04-24
Optimize range-for loop variable through const reference

JohnCoconut:
With the help of clang-tidy, https://clang.llvm.org/extra/clang-tidy/checks/performance-for-range-copy.html

> Finds C++11 for ranges where the loop variable is copied in each iteration but it would suffice to obtain it by const reference.

> The check is only applied to loop variables of types that are expensive to copy which means they are not trivially copyable or have a non-trivial copy constructor or destructor.

> To ensure that it is safe to replace the copy with a const reference the following heuristic is employed:

>    The loop variable is const qualified.
>    The loop variable is not const, but only const methods or operators are invoked on it, or it is used as const reference or value argument in constructors or function calls.

A example is shown below.

```cpp
-        for (std::string str : data->armatures)
+        for (const std::string& str : data->armatures)
         {
             removeArmatureData(str);
         }
```

### DIFF
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -757,7 +757,7 @@ bool SpriteFrameCache::PlistFramesCache::erasePlistIndex(const std::string &plis
     if (it == _indexPlist2Frames.end()) return false;
 
     auto &frames = it->second;
-    for (auto f : frames)
+    for (const auto& f : frames)
     {
         // !!do not!! call `_spriteFrames.erase(f);` to erase SpriteFrame
         // only erase index here
--- a/cocos/2d/CCTMXXMLParser.cpp
+++ b/cocos/2d/CCTMXXMLParser.cpp
@@ -752,7 +752,7 @@ void TMXMapInfo::endElement(void* /*ctx*/, const char *name)
             }
 
             uint32_t* bufferPtr = reinterpret_cast<uint32_t*>(buffer);
-            for(auto gidToken : gidTokens) {
+            for(const auto& gidToken : gidTokens) {
                 auto tileGid = (uint32_t)strtoul(gidToken.c_str(), nullptr, 10);
                 *bufferPtr = tileGid;
                 bufferPtr++;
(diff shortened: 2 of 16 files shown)


minggo (approved):

crazyhappygame:
@JohnCoconut could you add in next pull request new Travis CI build configuration for clang-tidy?
In that way we will automatically detect regressions.

minggo:
@crazyhappygame good idea. And i think we can add it in CI after all codes past it.

JohnCoconut:
> could you add in next pull request new Travis CI build configuration for clang-tidy? In that way we will automatically detect regressions.

@crazyhappygame @minggo 

I will add `clang-tidy` in my fork first to see how it goes. 

`clang-tidy` have quite a number of options that reports many different types of warnings and errors. But it's too *noisy* (but really need to be fixed) for this repository at the moment. 

My plan is to add `clang-tidy` configuration options step by step. 

1. enable options that I have fixed already.
2. enable a option in my fork, fix the code that breaks it.
3. enable the option above in this repo.


This is the list of checks, https://clang.llvm.org/extra/clang-tidy/checks/list.html

I would like to enable all the `bugprone-*`, `performance-*`, `readability-*` options first. It's a lot of work to do. `clang-tidy` does have a options to fix the code automatically. While it works most of the time, sometime it just breaks things.

JohnCoconut:
I would also like to add `Coverity Scan` to travis ci sometime in future.  https://scan.coverity.com/travis_ci

`coverity` is a free static analysis tool. It has a usage limit though. 

> Up to 7 builds per week, with a maximum of 1 build per day, for projects with more than 1 million lines of code

crazyhappygame:
@JohnCoconut sounds like good plan. Please create .clang-tidy file with enabled checks fixed by you.

JohnCoconut:
This `BUILD_TARGET=mac` on travis-ci baffles me.  Why would `generator.py` sometimes fail with a strange message? 

Is there something wrong with `generator.py`?

crazyhappygame:
@JohnCoconut I am so sorry, I do not understand your question. Could you paste link to error?

About clang-tidy build please use Linux as CI host. Form my experience Travis mac hosts are slower and not always available.

minggo:
@JohnCoconut could you please send a PR to metal-support too? Thanks.

JohnCoconut:
@crazyhappygame The build on mac failed within few minutes earlier host due to a error on `generator.py`. But it succeeded after another try. 

Unfortunately, travis seems to delete earlier build history of same PR. So I couldn't find it now.

PR #19642, opened 2019-04-26
Add intial clang-tidy support

JohnCoconut:
Only two options are enabled (hard-coded in script)

* performance-faster-string-find
* performance-for-range-copy

Treat warnings as errors are enabled.

My bash script skill is super rough :)

### DIFF
--- a/tools/travis-scripts/run-clang-tidy.py
+++ b/tools/travis-scripts/run-clang-tidy.py
@@ -1,0 +1,329 @@
+#!/usr/bin/env python
+#
+#===- run-clang-tidy.py - Parallel clang-tidy runner ---------*- python -*--===#
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+#===------------------------------------------------------------------------===#
+# FIXME: Integrate with clang-tidy-diff.py
+
+"""
+Parallel clang-tidy runner
+==========================
+
+Runs clang-tidy over all files in a compilation database. Requires clang-tidy
+and clang-apply-replacements in $PATH.
+
+Example invocations.
+- Run clang-tidy on all files in the current working directory with a default
+  set of checks and show warnings in the cpp files and all project headers.
+    run-clang-tidy.py $PWD
+
+- Fix all header guards.
+    run-clang-tidy.py -fix -checks=-*,llvm-header-guard
+
+- Fix all header guards included from clang-tidy and header guards
+  for clang-tidy headers.
+    run-clang-tidy.py -fix -checks=-*,llvm-header-guard extra/clang-tidy \
+                      -header-filter=extra/clang-tidy
+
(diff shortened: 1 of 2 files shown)


JohnCoconut:
There is an warning in the `external` directory from `cocos2d-x-3rd-party-libs-bin/flatbuffers/util.h`.

```cpp
size_t i = filepath.find_last_of(".");
```

There are two ways to deal with it.

1. Ignore `external` directory in clang-tidy scan.
2. Fix `external` directory so it passes scan. But we need to create a release version of pre-built libraries. Problem is that we have to first fix the warning in external repo, create a release version of it, and test that change in cocos2d-x. This work flow seems very strange. We shouldn't create a release version of something, then test the change.

@minggo  What is your opinion?

I made a fork of `cocos2d-x-3rd-party-libs-bin` repo, and run
```bash
cmake ..  -DCMAKE_EXPORT_COMPILE_COMMANDS=on -DLINUX=on -DWINDOWS=off -DBUILD_LUA_LIBS=on -DBUILD_JS_LIBS=on
```

I run clang-tidy with the generated compilation database and noticed the warning wasn't triggered.

tools/travis-scripts/run-clang-tidy.py
```diff
@@ -0,0 +1,329 @@
+#!/usr/bin/env python
```
crazyhappygame, line 1: This script is delivered with clang-tidy. Please do not copy it to cocos2d-x source code.

tools/travis-scripts/run-script.sh
```diff
@@ -27,13 +27,23 @@ function do_retry()
...
+
 function build_linux()
 {
     echo "Building tests ..."
     cd $COCOS2DX_ROOT/build
     mkdir -p linux-build
     cd linux-build
-    cmake ../..
+    cmake ../.. -DCMAKE_EXPORT_COMPILE_COMMANDS=on
+    run_clang_tidy $PWD
```
crazyhappygame, line 46: Please create new Travis build based on Linux build. Travis executes builds in parallel

crazyhappygame:
Please ignore external directory for now. With e.g. flatbuffer I think that currently very old version is used.

tools/travis-scripts/run-clang-tidy.py
JohnCoconut: The script only shows warnings. It doesn't return a error to the shell. 

More specifically, `clang-tidy` has a option `-warnings-as-errors=<string>`, but I couldn't find it in the `run-clang-tidy.py ` script. Do you know any way to do it without modifying the script?

I just find other people doing the same, https://github.com/Sarcasm/irony-mode/blob/82ba45ec15c9011bbdf1d69cf25c8193d33c0028/server/build-aux/run-clang-tidy/run-clang-tidy.py

crazyhappygame: Please create .clang-tidy file.
E.g. with "clang-tidy -dump-config".
In .clang-tidy file there is WarningAsErrors field. Please set it to '*'

JohnCoconut: I see, thank you.

PR #19643, opened 2019-04-27
add clang-tidy support

JohnCoconut:
Only two options are enabled (hard-coded in script)

* performance-faster-string-find
* performance-for-range-copy

Treat warnings as errors are enabled.

Another try.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,7 +1,7 @@
 ---
 Checks: "-*,\
+performance-faster-string-find,\
 performance-for-range-copy,\
-readability-const-return-type,\
 "
 WarningsAsErrors: '*'
 HeaderFilterRegex: '/(?!external)/.*'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,11 @@
 ---
-Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,performance-faster-string-find,performance-for-range-copy,'
+Checks: >
+  clang-diagnostic-*,
+  clang-analyzer-*,
+  -*,
+  performance-faster-string-find,
+  performance-for-range-copy,
+
 WarningsAsErrors: '*'
 HeaderFilterRegex: '/(?!external)/.*'
 AnalyzeTemporaryDtors: false
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,8 +1,38 @@
 ---
-Checks: "-*,\
-performance-faster-string-find,\
-performance-for-range-copy,\
-"
+Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,performance-faster-string-find,performance-for-range-copy,'
 WarningsAsErrors: '*'
 HeaderFilterRegex: '/(?!external)/.*'
+AnalyzeTemporaryDtors: false
+FormatStyle:     none
+CheckOptions:    
+  - key:             cert-dcl16-c.NewSuffixes
+    value:           'L;LL;LU;LLU'
+  - key:             cppcoreguidelines-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic
+    value:           '1'
+  - key:             google-readability-braces-around-statements.ShortStatementLines
+    value:           '1'
+  - key:             google-readability-function-size.StatementThreshold
+    value:           '800'
+  - key:             google-readability-namespace-comments.ShortNamespaceLines
+    value:           '10'
+  - key:             google-readability-namespace-comments.SpacesBeforeComments
+    value:           '2'
+  - key:             modernize-loop-convert.MaxCopySize
+    value:           '16'
+  - key:             modernize-loop-convert.MinConfidence
+    value:           reasonable
+  - key:             modernize-loop-convert.NamingStyle
+    value:           CamelCase
+  - key:             modernize-pass-by-value.IncludeStyle
+    value:           llvm
+  - key:             modernize-replace-auto-ptr.IncludeStyle
+    value:           llvm
+  - key:             modernize-use-nullptr.NullMacros
+    value:           'NULL'
+  - key:             performance-faster-string-find.StringLikeClasses
+    value:           'std::basic_string'
+  - key:             performance-for-range-copy.AllowedTypes
+    value:           ''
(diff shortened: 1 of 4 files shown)


JohnCoconut:
YAML:1:1: error: not a mapping
     /home/travis/build/cocos2d/cocos2d-x/tools/travis-scripts/../../.clang-tidy

Seems to complain my `.clang-tidy` is invalid, but it runs fine in my local linux machine. Is it because I am use clang 8?

The format is copied from [CMake](https://github.com/Kitware/CMake/blob/07cfb18f9d29cfc0588ede928846a03ec5599c48/.clang-tidy) git repo.

tools/travis-scripts/run-script.sh
```diff
@@ -37,6 +37,16 @@ function build_linux()
...
 }
 
+function build_linux_clang_tidy()
+{
+    echo "Building clang tidy test on Linux ..."
+    cd $COCOS2DX_ROOT/build
+    mkdir -p clang-tidy-build
+    cd clang-tidy-build
+    cmake ../.. -DCMAKE_EXPORT_COMPILE_COMMANDS=on
+    python /usr/local/clang-7.0.0/share/clang/run-clang-tidy.py
```
minggo:
```
cmake ../.. -DCMAKE_EXPORT_COMPILE_COMMANDS=on
python /usr/local/clang-7.0.0/share/clang/run-clang-tidy.py
```

Don't quite understand these codes. And the python script path is hard coded, it may be a problem when CI changes its environment.

JohnCoconut: I `apt search clang-tidy` in travis ci and found the most recent version is 6. But the `clang-tidy` being used is 7. My guess is that `clang-tidy` is not installed by package manager. I am not sure how to find `run-clang-tidy.py`.

Or we can just copy one and store it in the repo, if you agree.

crazyhappygame: @minggo 
`cmake ../.. -DCMAKE_EXPORT_COMPILE_COMMANDS=on`
CMAKE_EXPORT_COMPILE_COMMANDS is used to generate compilation database. This is the easiest way to use clang-tidy with CMake 
https://clang.llvm.org/docs/JSONCompilationDatabase.html

About hard-coded path 
`python /usr/local/clang-7.0.0/share/clang/run-clang-tidy.py`
It can be changed to e.g.:
"python $(dirname $(dirname $(readlink -f \`which clang-tidy\`)))/share/clang/run-clang-tidy.py"

JohnCoconut: @crazyhappygame Great. It works on Fedora Linux too. I didn't know clang-tidy shares top-level directory with its python script.

.clang-tidy
```diff
@@ -0,0 +1,38 @@
+---
+Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,performance-faster-string-find,performance-for-range-copy,'
```
crazyhappygame: Please change to yaml multplie line string https://yaml-multiline.info/. In that way we can easily track changes
```
Checks: >
    -*,
    clang-analyzer-*,
    clang-diagnostic-*,
    performance-faster-string-find,
    performance-for-range-copy,
```

JohnCoconut:
```yaml
Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,performance-faster-string-find,performance-for-range-copy,'
```

The configuration above is generated by `clang-tidy -header-filter='/(?!external)/.*' -checks='-*,performance-faster-string-find,performance-for-range-copy,' -warnings-as-errors='*' -dump-config`


```yaml
Checks: >
    -*,
    clang-analyzer-*,
    clang-diagnostic-*,
    performance-faster-string-find,
    performance-for-range-copy,
```

I just found that the above two are not the same. The order matters. `-*` will disable options ahead of it.

* In the first config, `clang-diagnostic-*,clang-analyzer-*` is disabled.

* The second config, placing `-*` at the front, will not disable `clang-diagnostic-*,clang-analyzer-*`.

tools/travis-scripts/run-script.sh
```diff
@@ -291,6 +301,12 @@ function run_pull_request()
         build_linux
     fi
 
+    # linux_clang_tidy_test
+    if [ $BUILD_TARGET == 'linux_clang_tidy' ]; then
+        genernate_binding_codes
```
crazyhappygame: I think that genernate_binding_codes should be removed

.clang-tidy
```diff
@@ -0,0 +1,44 @@
+---
+Checks: >
```
crazyhappygame, line 2: Why  "-*" is needed?
I thought that plan was enable "performance-faster-string-find", "performance-for-range-copy".
```
Checks: >
  performance-faster-string-find,
  performance-for-range-copy,
```
Are you sure that all "clang-diagnostic-*" , "clang-analyzer-*" checks are already fixed?

.clang-tidy
```diff
@@ -0,0 +1,44 @@
...
+  clang-analyzer-*,
+  -*,
+  performance-faster-string-find,
+  performance-for-range-copy,
+
+WarningsAsErrors: '*'
+HeaderFilterRegex: '/(?!external)/.*'
+AnalyzeTemporaryDtors: false
+FormatStyle:     none
+CheckOptions:
```
crazyhappygame: I propose to remove "CheckOptions:".
Any way default values for "CheckOptions:" were used

tools/travis-scripts/run-script.sh
minggo: Yep, it is not needed.

tools/travis-scripts/run-script.sh
minggo: @crazyhappygame thanks for the explanation.

.clang-tidy
JohnCoconut: @crazyhappygame  `clang-tidy` enables a lot of default options. The `-*` option is to remove all options currently enabled.

crazyhappygame (approved):

crazyhappygame:
@minggo could you merge it?

minggo:
Sure, it is merged.

PR #19650, opened 2019-04-30
add methods for reloading buffer data outside the buffer when necessary

Mee-gu:
add methods for reloading buffer data outside the buffer when necessary

### DIFF
--- a/cocos/renderer/backend/opengl/BufferGL.h
+++ b/cocos/renderer/backend/opengl/BufferGL.h
@@ -16,24 +16,25 @@ public:
     
     virtual void updateData(void* data, unsigned int size) override;
     virtual void updateSubData(void* data, unsigned int offset, unsigned int size) override;
-    
+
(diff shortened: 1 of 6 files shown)


cocos/renderer/backend/opengl/BufferGL.h
```diff
@@ -16,24 +16,22 @@ class BufferGL : public Buffer
     
     virtual void updateData(void* data, unsigned int size) override;
     virtual void updateSubData(void* data, unsigned int offset, unsigned int size) override;
-    
+    virtual void usingDefaultStoredData(bool needDefaultStoredData) override ;
+
     inline GLuint getHandler() const { return _buffer; }
 
-private:
```
minggo: why remove it?

Mee-gu: Yes, private is needed and its a mistake, I will fix it.

PR #19656, opened 2019-04-30
enable more clang-tidy performance options

JohnCoconut:
Add more clang-tidy options. Fix warnings generated, 

*  performance-implicit-conversion-in-loop,
*  performance-inefficient-algorithm,
*  performance-inefficient-vector-operation,
*  performance-move-const-arg,
*  performance-type-promotion-in-math-fn,

I wanted to add `performance-inefficient-string-concatenation`, but some warnings seemed misleading. So I only fixed what is necessary.

About removal of  `std::move` below,
```cpp
_uniforms[uniform.second.location] = std::move(value);
```
We didn't implement a move constructor for `UniformValue`, so `std::move` doesn't work.


JohnCoconut:
Some thing changed in newer clang-tidy versions? This error isn't reported in my Linux workstation(version 8.0)

```cpp
/home/travis/build/cocos2d/cocos2d-x/cocos/2d/CCFastTMXLayer.cpp:212:31: error: call to 'ceil' promotes float to double [performance-type-promotion-in-math-fn,-warnings-as-errors]

visibleTiles.size.width = ceil(visibleTiles.origin.x + visibleTiles.size.width)  - floor(visibleTiles.origin.x);
```

crazyhappygame:
@JohnCoconut travis CI has clang-tidy 7.0. 
But it is not only about cland-tidy version but also about standard library. Ideally you should have exactly the same environment.

cocos/base/ccUTF8.cpp
```diff
@@ -526,7 +526,8 @@ long cc_utf8_strlen (const char * p, int /*max*/)
 unsigned int cc_utf8_find_last_not_char(const std::vector<unsigned short>& str, unsigned short c)
 {
     std::vector<char16_t> char16Vector;
-    for (const auto& e : str)
+    char16Vector.reserve(str.size());
+for (const auto& e : str)
```
minggo: error indent

minggo (approved):

JohnCoconut:
@crazyhappygame 

> But it is not only about cland-tidy version but also about standard library. 

That's a very good point. I haven't thought about that. I will spin a ubuntu xenial VM to test my commit in future.

cocos/base/ccUTF8.cpp
JohnCoconut: Fixed. How do  I even missed that?

crazyhappygame: @minggo is it possible to create .clang-format file for cocos2d-x and use automated formatting everywhere?

crazyhappygame:
@JohnCoconut I propose to create 1 PR per one clang-tidy check. In that way we will have smaller reviews and faster iteration.

crazyhappygame:
@JohnCoconut clang-tidy is great. But what do you thing about following compilation flags:
```
        -pedantic
        -pedantic-errors
        -Wall
        -Werror
        -Wextra
```

JohnCoconut:
@crazyhappygame 

> I propose to create 1 PR per one clang-tidy check. 

Agree. It's easier to make small steps.

> compilation flags

We can enable `-Wall -Wpedantic -Wextra`, fix warnings, and then enable `-Werror`.

> .clang-format

I like `clang-format`. But applying it globally in the repo can be ugly. A quote from Jakeherringbone on his [medium article](https://medium.com/@Jakeherringbone/incrementally-adopt-a-formatter-in-a-large-repo-3c253fee1d28).

> However, we want to avoid a massive disruptive re-format of the whole repository. Such a re-format is a big cost — you pollute your blame layer in version control so that it always takes an extra hop to a previous change to understand authorship. Plus, all the pending pull requests at the time of the change will need an ugly rebase. The same thing can happen later if you upgrade or re-configure the formatter.

The way Jakeherringbone deals with it is,

> To be incremental, the formatter must work together with the version control system, finding the lines changed by the commit and then formatting only regions of the file.

I am very fond of this method.

cocos/base/ccUTF8.cpp
minggo: Yep, sure. Would you please add one?

minggo:
> To be incremental, the formatter must work together with the version control system, finding the lines changed by the commit and then formatting only regions of the file.

@JohnCoconut i like the idea.

PR #19664, opened 2019-05-05
Feature/draw in rect

Mee-gu:
make Texture2D draw in point and draw in rectangle tests work


cocos/renderer/CCTexture2D.h
```diff
@@ -380,6 +382,9 @@ class CC_DLL Texture2D : public Ref
     Texture2D* getAlphaTexture() const;
 
     bool getAlphaTextureName() const;
+    
+    void setupProgram(float globalZOrder);
+
```
minggo: Does this function need to be public?

Mee-gu: I think it's fine to be public, we can call it when needed.

Mee-gu:
update review

minggo (approved):

Issue #19662, opened 2019-05-05
RenderState logic is not correct


minggo:
https://discuss.cocos2d-x.org/t/metal-support-alpha1-released/46282/28

Issue #19663, opened 2019-05-05
Pass misses getRenderState()


minggo:
https://discuss.cocos2d-x.org/t/metal-support-alpha1-released/46282/28

PR #19665, opened 2019-05-05
fix RenderState::StateBlock

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19662

- `unbindPass` is redundant and it is removed.
- `_modifiedBits` is set directly. 

- add `Pass::getStateBlock`


PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19663

PR #19670, opened 2019-05-06
add new clang-tidy option

JohnCoconut:
* readability-container-size-empty

Reference: https://releases.llvm.org/7.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-container-size-empty.html

> Checks whether a call to the `size()` method can be replaced with a call to `empty()`.

> The emptiness of a container should be checked using the `empty()` method instead of the `size()` method. It is not guaranteed that `size()` is a constant-time function, and it is generally more efficient and also shows clearer intent to use `empty()`. Furthermore some containers may implement the `empty()` method but not implement the `size()` method. Using `empty()` whenever possible makes it easier to switch to another container in the future.

### DIFF
--- a/tools/travis-scripts/run-script.sh
+++ b/tools/travis-scripts/run-script.sh
@@ -44,7 +44,8 @@ function build_linux_clang_tidy()
     mkdir -p clang-tidy-build
     cd clang-tidy-build
     cmake ../.. -DCMAKE_EXPORT_COMPILE_COMMANDS=on
-    python $(dirname $(dirname $(readlink -f `which clang-tidy`)))/share/clang/run-clang-tidy.py
+    clang_tidy_script=$(dirname $(dirname $(readlink -f `which clang-tidy`)))/share/clang/run-clang-tidy.py 
+    python $clang_tidy_script '^((?!/cocos2d-x/external/).)*$'
 }
 
 function build_mac()
--- a/tools/travis-scripts/run-script.sh
+++ b/tools/travis-scripts/run-script.sh
@@ -45,6 +45,7 @@ function build_linux_clang_tidy()
     cd clang-tidy-build
     cmake ../.. -DCMAKE_EXPORT_COMPILE_COMMANDS=on
     clang_tidy_script=$(dirname $(dirname $(readlink -f `which clang-tidy`)))/share/clang/run-clang-tidy.py 
+    # run clang-tidy on all files except those in external directory
     python $clang_tidy_script '^((?!/cocos2d-x/external/).)*$'
 }
 
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -8,6 +8,7 @@ Checks: >
   performance-inefficient-vector-operation,
   performance-move-const-arg,
   performance-type-promotion-in-math-fn,
+  readability-container-size-empty,
 
 WarningsAsErrors: '*'
 HeaderFilterRegex: '/(?!external)/.*'
--- a/cocos/2d/CCAtlasNode.cpp
+++ b/cocos/2d/CCAtlasNode.cpp
@@ -71,7 +71,7 @@ AtlasNode * AtlasNode::create(const std::string& tile, int tileWidth, int tileHe
 
 bool AtlasNode::initWithTileFile(const std::string& tile, int tileWidth, int tileHeight, int itemsToRender)
 {
-    CCASSERT(tile.size() > 0, "file size should not be empty");
+    CCASSERT(!tile.empty(), "file size should not be empty");
     Texture2D *texture = Director::getInstance()->getTextureCache()->addImage(tile);
     return initWithTexture(texture, tileWidth, tileHeight, itemsToRender);
 }
--- a/cocos/2d/CCClippingNode.cpp
+++ b/cocos/2d/CCClippingNode.cpp
@@ -332,7 +332,7 @@ void ClippingNode::setStencil(Node *stencil)
 
 bool ClippingNode::hasContent() const
 {
-    return _children.size() > 0;
+    return !_children.empty();
 }
 
 GLfloat ClippingNode::getAlphaThreshold() const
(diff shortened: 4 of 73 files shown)


minggo (approved):

minggo:
```
/home/travis/build/cocos2d/cocos2d-x/tests/cpp-tests/Classes/UnitTest/UnitTest.cpp:953:9: error: the 'empty' method should be used to check for emptiness instead of comparing to an empty object [readability-container-size-empty,-warnings-as-errors]
        EXPECT_EQ("", u.getPassword());
```

It seems need to modify more codes to use this option.

JohnCoconut:
@minggo 

```cpp
EXPECT_EQ("", u.getPassword());
```

I am changing it to, 

```cpp
EXPECT_TRUE(u.getPassword().empty());
```

minggo:
Yep, it is good.

JohnCoconut:
A error from `external/clipper/clipper.cpp:2014:7`,

```cpp
m_GhostJoins.size() > 0 && (rb->WindDelta != 0))
```

I thought I filtered `external` directory in `.clang-tidy`,  it turns out that 

    HeaderFilterRegex: '/(?!external)/.*'

only filters header file, not `cpp` file.


I found two ways to do this. One from https://stackoverflow.com/questions/49591804/clang-tidy-cmake-exclude-file-from-check. 

Another from https://github.com/Caphyon/clang-power-tools/issues/281. I will see how clang-power-tools did it.

minggo:
I think it is better that it can ignore some files for some directories. We don't want to modify 3rd libraries, or it will be hard to upgrade 3rd libraries.

JohnCoconut:
> I think it is better that it can ignore some files for some directories. We don't want to modify 3rd libraries, or it will be hard to upgrade 3rd libraries.

Yeah. I am looking for a non disruptive solution. `.clang-tidy` doesn't seem to have a option to do that directly.

@crazyhappygame Do you have any idea of how to ignore `cpp` files in `clang-tidy` config file?

crazyhappygame:
@JohnCoconut
By default run-clang-tidy checks all files from compilation database. Please change call to:
```
run-clang-tidy ".*/cocos2d-x/cocos/.*" ".*/cocos2d-x/extensions/.*" ".*/cocos2d-x/tests/.*"
```

JohnCoconut:
@minggo @crazyhappygame 

The script does have a positional argument. I can use regex to filter `external` directory.

```python
  parser.add_argument('files', nargs='*', default=['.*'],
                      help='files to be processed (regex on path)')
```

crazyhappygame:
Yes it use regexp and join with or. So below:
 ".*/cocos2d-x/cocos/.*" ".*/cocos2d-x/extensions/.*" ".*/cocos2d-x/tests/.*"
It is changed to regexp:
 ".*/cocos2d-x/cocos/.*"|".*/cocos2d-x/extensions/.*"|".*/cocos2d-x/tests/.*"

tools/travis-scripts/run-script.sh
```diff
@@ -44,7 +44,8 @@ function build_linux_clang_tidy()
     mkdir -p clang-tidy-build
     cd clang-tidy-build
     cmake ../.. -DCMAKE_EXPORT_COMPILE_COMMANDS=on
-    python $(dirname $(dirname $(readlink -f `which clang-tidy`)))/share/clang/run-clang-tidy.py
+    clang_tidy_script=$(dirname $(dirname $(readlink -f `which clang-tidy`)))/share/clang/run-clang-tidy.py 
+    python $clang_tidy_script '^((?!/cocos2d-x/external/).)*$'
```
minggo, line 49: could you please add some comment here?

JohnCoconut: `'^((?!/cocos2d-x/external/).)*$'` is a python regular expression. It's used in `run-clang-tidy.py`,

```python
file_name_re = re.compile('|'.join(args.files))
```
with `args.files = '^((?!/cocos2d-x/external/).)*$'`. 

The regular expression is used to check all files included in the compilation database. If a file path contains `/cocos2d-x/external/`, that file will be ignored for `clang-tidy`.

Reference: [Python regular expression documentation](https://docs.python.org/2.7/library/re.html)


Here is a small test case for the python re used,

```python
import re

files = [ 
"/home/john/cocos2d-x/external/openssl/include/linux/openssl/bio.h",
"/home/john/cocos2d-x/external/tiff/include/linux/tiff.h",
"/home/john/cocos2d-x/external/json/stringbuffer.h",
"/home/john/cocos2d-x/cocos/base/ccUtils.h",
"/home/john/cocos2d-x/cocos/physics/CCPhysicsBody.cpp",
"/home/john/cocos2d-x/tests/cpp-tests/Classes/ActionsEaseTest/ActionsEaseTest.cpp",
"/home/john/cocos2d-x/templates/cpp-template-default/Classes/AppDelegate.cpp"
]

pattern = '^((?!/cocos2d-x/external/).)*$'

for file in files:
    m = re.search(pattern, file)
    if m:
        print m.group(0)
```

The output shows all file includes `/cocos2d-x/external/` are filtered out.
```bash
/home/john/cocos2d-x/cocos/base/ccUtils.h
/home/john/cocos2d-x/cocos/physics/CCPhysicsBody.cpp
/home/john/cocos2d-x/tests/cpp-tests/Classes/ActionsEaseTest/ActionsEaseTest.cpp
/home/john/cocos2d-x/templates/cpp-template-default/Classes/AppDelegate.cpp
```

minggo: Thanks for the explanation, what i mean is to add comments in the script file, because regexp is not readable.

JohnCoconut: added.

minggo: thanks

PR #19671, opened 2019-05-07
modify fdGetter

ghost:
If there hasn't an OBB file, we should try to get the URL in the normal place because that that do not put audio files into the OBB file is a common strategy to publish APK.

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxMusic.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxMusic.java
@@ -280,7 +280,7 @@ public class Cocos2dxMusic {
                 mediaPlayer.setDataSource(fis.getFD());
                 fis.close();
             } else {
-                if (Cocos2dxHelper.getObbFile() != null) {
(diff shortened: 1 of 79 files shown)


minggo (approved):

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxMusic.java
```diff
@@ -280,7 +280,7 @@ private MediaPlayer createMediaPlayer(final String path) {
                 mediaPlayer.setDataSource(fis.getFD());
                 fis.close();
             } else {
-                if (Cocos2dxHelper.getObbFile() != null) {
+                if (Cocos2dxHelper.getObbFile() != null && Cocos2dxHelper.getObbFile().getAssetFileDescriptor(path) != null) {
```
minggo, line 283: I think it is better to check `assetFileDescriptor ` inside the` if` sentence.

ghost: My purpose is that checking whether the OBB file exists and whether the music file is inside the OBB file. If the music file is not inside the OBB file, we should continue to find it through `this.mContext.getAssets()`.

PR #19676, opened 2019-05-08
[cpp-tests] SpritePolygonTest invoke `UISlide::setPercent` after initialization 

PatriceJiang:
PR https://github.com/cocos2d/cocos2d-x/pull/19556  cause crash when invoke `setPercent`

For example:

https://github.com/cocos2d/cocos2d-x/blob/d6b8e994a4fd065a005fc06e52a01001d55c5e3a/tests/cpp-tests/Classes/SpritePolygonTest/SpritePolygonTest.cpp#L298-L303

- `setPercent` will trigger `changeEpsilon()`
- `changeEpsilon()` will access `_epsilonLabel` which is not initialized by then

### Fix

~~delay the trigger of event `ON_PERCENTAGE_CHANGED`~~

invoke `setPercent` after initialization completed.

### DIFF
--- a/cocos/ui/UISlider.cpp
+++ b/cocos/ui/UISlider.cpp
@@ -28,6 +28,8 @@ THE SOFTWARE.
 #include "ui/UIHelper.h"
 #include "2d/CCSprite.h"
 #include "2d/CCCamera.h"
+#include "base/CCScheduler.h"
+
 #include "editor-support/cocostudio/CocosStudioExtension.h"
 #include "renderer/ccShaders.h"
 NS_CC_BEGIN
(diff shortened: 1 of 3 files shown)


minggo:
So v3 will crash too?

PatriceJiang:
yes, v3 will also crash

minggo:
I think the modification is not reasonable, the game logic should make sure the variable is valid.

PatriceJiang:
I have revert these modifications and fix the bug in test case.

minggo (approved):

minggo:
please send a PR to 3.17.2 branch.

PR #19688, opened 2019-05-10
refactor CCIMEDispatcher

JohnCoconut:
I plan to add a new clang-tidy option here, but error message generated for these files are confusing. I ended up re-factoring them a bit.

* remove superfluous destructor
* use range-based for loop
* correct some documentation

```cpp
 if (_impl->_delegateWithIme)
```

The condition check above is not necessary, we are not dereferencing `_impl->_delegateWithIme` anyway.


minggo:
looks good to me, thanks.

minggo (approved):

PR #19690, opened 2019-05-10
android fix get_environment.sh

PatriceJiang:
pending https://github.com/cocos2d/cocos2d-console/pull/459

`gradle` can not access environment variable `COCOS_CONSOLE_ROOT` when building android on linux

### DIFF
--- a/templates/lua-template-default/frameworks/runtime-src/proj.android/app/get_environment.sh
+++ b/templates/lua-template-default/frameworks/runtime-src/proj.android/app/get_environment.sh
@@ -1,4 +1,11 @@
 #!/bin/bash
 
-source ~/.bash_profile
+if [ -f $HOME/.bashrc ]; then
+    source $HOME/.bashrc
+fi
+
+if [ -f $HOME/.bash_profile ]; then
+    source $HOME/.bash_profile
+fi
+
 echo $COCOS_CONSOLE_ROOT
--- a/tests/lua-empty-test/project/proj.android/app/build.gradle
+++ b/tests/lua-empty-test/project/proj.android/app/build.gradle
@@ -114,11 +114,19 @@ def getCocosCommandPath() {
         return 'cocos.bat'
     }
     else {
+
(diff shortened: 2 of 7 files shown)


templates/lua-template-default/frameworks/runtime-src/proj.android/app/get_environment.sh
```diff
@@ -1,4 +1,11 @@
...
 
-source ~/.bash_profile
-echo $COCOS_CONSOLE_ROOT
\ No newline at end of file
+if [ -f $HOME/.bashrc ]; then
+    source $HOME/.bashrc
+fi
+
+if [ -f $HOME/.bash_profile ]; then
+    source $HOME/.bash_profile
```
PatriceJiang, line 8: `~/.bash_profile` may not exists on linux

tests/lua-empty-test/project/proj.android/app/build.gradle
```diff
@@ -114,11 +114,19 @@ def getCocosCommandPath() {
...
+            return new File(env_console_path.trim() + '/cocos').absolutePath
+        }
+
         // on unix like system, can not get environments variables easily
         // so run a shell script to get environment variable sets by cocos2d-x setup.py
         new ByteArrayOutputStream().withStream { os ->
             def result = exec {
-                executable = project.file('get_environment.sh')
+                executable = "/bin/bash"
+                args = [ "-i", project.file('get_environment.sh')]
```
PatriceJiang, line 129: https://linux.die.net/man/1/bash

> When bash is invoked as an interactive login shell, or as a non-interactive shell with the  --login  option,
       it  first  reads  and executes commands from the file /etc/profile, if that file exists.  After reading that
       file, it looks for ~/.bash_profile, ~/.bash_login, and ~/.profile, in that order,  and  reads  and  executes
       commands  from the first one that exists and is readable.  The --noprofile option may be used when the shell
       is started to inhibit this behavior.

PR #19693, opened 2019-05-10
 FIX ios bug : You don’t have permission to save the file “Documents_temp”

yangyp-yes:
The iOS side, eg:Writable directory is `**/Documents/`,old code will get `**/Documents_temp/`, However, due to the sandbox mechanism of iOS, you do not have the permission of this directory, so instead fixed it to `**/Documents/_temp/`

### DIFF
--- a/extensions/assets-manager/AssetsManagerEx.cpp
+++ b/extensions/assets-manager/AssetsManagerEx.cpp
@@ -314,7 +314,8 @@ void AssetsManagerEx::setStoragePath(const std::string& storagePath)
     _fileUtils->createDirectory(_storagePath);
     
     _tempStoragePath = _storagePath;
-    _tempStoragePath.insert(_storagePath.size() - 1, TEMP_PACKAGE_SUFFIX);
+    _tempStoragePath.insert(_storagePath.size(), TEMP_PACKAGE_SUFFIX);
+    adjustPath(_tempStoragePath);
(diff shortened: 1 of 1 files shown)


yangyp-yes:
[forum link](https://forum.cocos.com/t/topic/47298)

yangyp-yes:
[apple link](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html)

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -314,7 +314,8 @@ void AssetsManagerEx::setStoragePath(const std::string& storagePath)
     _fileUtils->createDirectory(_storagePath);
     
     _tempStoragePath = _storagePath;
-    _tempStoragePath.insert(_storagePath.size() - 1, TEMP_PACKAGE_SUFFIX);
+    _tempStoragePath.insert(_storagePath.size(), TEMP_PACKAGE_SUFFIX);
```
minggo: Then use `string::append` is better? And change `TEMP_PACKAGE_SUFFIX ` to `temp` is better too?

yangyp-yes: I have changed!

minggo (approved):

PR #19696, opened 2019-05-11
Android Studio Gradle Warning for API 'variant.getMergeAssets()'

HTC77:
_Based on issue #19683_
- cocos2d-x version: 3.17.1
- devices test on:
- developing environments
   - NDK version: r16b
   - Xcode version:
   - VS version: 2015
   - browser type and version:
   - Android Studio version: 3.3.2
Steps to Reproduce:

1. Open proj.android in Android Studio
2. Build and run project
3. Gradle Warning will be appear:
API 'variant.getMergeAssets()' is obsolete and has been replaced with 'variant.getMergeAssetsProvider()'.
It will be removed at the end of 2019.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,9 @@
+ <!-- uncommment lines to show your screenshot -->
+<!--
+<div align="center"><img width="100%" height="100%" style="margin:auto" src="Screenshot.png"></div>
+<hr>  -->
+
+
 <img src="http://www.cocos2d-x.org/attachments/801/cocos2dx_portrait.png" width=200>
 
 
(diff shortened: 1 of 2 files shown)


minggo:
I don't think this PR fixes the corresponding issue

Dimon4eg:
PR with fix is here https://github.com/cocos2d/cocos2d-x/pull/19703

HTC77:
I think it's depends on gradle version. Gradle needs to be upgrade. I just found error was from this line:

https://github.com/HTC77/cocos2d-x/blob/fca3ddb441b26694929ecd63a9288bbf1ec26346/templates/cpp-template-default/proj.android/app/build.gradle#L113

PR #19697, opened 2019-05-12
remove duplicate function call in cmake

JohnCoconut:
* AssureOutOfSourceBuilds() is called in `PreventInSourceBuilds.cmake` module file. No need to call it in root cmake file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,6 @@ set(CMAKE_MODULE_PATH ${COCOS2DX_ROOT_PATH}/cmake/Modules/)
 
 # prevent in-source-build
 include(PreventInSourceBuilds)
-AssureOutOfSourceBuilds()
 
 # works before build libcocos2d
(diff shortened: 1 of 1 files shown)


PatriceJiang (approved):

PatriceJiang:
I think, to keep the invocation in root cmake file and remove the one in `PreventInSourceBuilds.cmake` could be better. @JohnCoconut

minggo:
@PatriceJiang i am confused. You approved the PR but you have different opinions.

PatriceJiang:
I am sorry for confusing you. I think the modification is ok, and I am not pretty sure which method is better.

JohnCoconut:
The function in preventinsourcebuild seems to be copied from a stack
overflow post. The answer in the post only calls the function in sub module
file. The module name states it's intention clearly. Putting function call
inside the module keeps main cmake file clean.

Issue #19363, opened 2019-01-28
Android Studio Gradle warning


MrCapone:
- cocos2d-x version: 3.17.1

- developing environments
   - Android Studio version: 3.3
   - Gradle plugin version: 3.2.1
 
```
WARNING: API 'variant.getAssemble()' is obsolete and has been replaced with 'variant.getAssembleProvider()'.
It will be removed at the end of 2019.
For more information, see https://d.android.com/r/tools/task-configuration-avoidance.
To determine what is calling variant.getAssemble(), use -Pandroid.debug.obsoleteApi=true on the command line to display a stack trace.
```

Problem in module that is defined in .../proj.android/build.gradle

MrCapone:
https://github.com/cocos2d/cocos2d-x/pull/19732

PR #19703, opened 2019-05-13
fix for #19363

MrCapone:
#19363

https://discuss.cocos2d-x.org/t/cocos2d-x-3-17-and-android-studio-3-3-build-issue-or-error/45503/6?u=capone

### DIFF
--- a/templates/cpp-template-default/proj.android/app/build.gradle
+++ b/templates/cpp-template-default/proj.android/app/build.gradle
@@ -117,7 +117,7 @@ android.applicationVariants.all { variant ->
     // delete previous files first
     delete dest_assets_folder
 
-    variant.mergeAssets.doFirst {
+    variant.mergeAssetsProvider.get().doFirst {
         copy {
             from project_root_folder + "/Resources"
             into dest_assets_folder
--- a/templates/js-template-default/frameworks/runtime-src/proj.android/app/build.gradle
+++ b/templates/js-template-default/frameworks/runtime-src/proj.android/app/build.gradle
@@ -152,7 +152,7 @@ android.applicationVariants.all { variant ->
     // delete previous files first
     delete dest_assets_folder
 
-    variant.mergeAssets.doFirst {
+    variant.mergeAssetsProvider.get().doFirst {
         copy {
            from project_root_folder + "/res"
            into dest_assets_folder + "/res"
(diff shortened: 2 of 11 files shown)


minggo:
It seems there is a conflict, could you please fix it, and could you please modify other templates too.

MrCapone:
Done. I also updated the tests too.

minggo:
@MrCapone have you tested it? @PatriceJiang told me it didn't take effect to modify like this. The warnings are gone, but the task was not executed.

MrCapone:
I already using this code in my projects. Assets folder has been successfully generated and copied to apk. So nothing wrong with this line. 
I'm using Android Studio version 3.4 with gradle plugin version 3.4.0.

PatriceJiang:
I tested the current v3 branch with similar changes and the files were not copied to the assets folder. @MrCapone would you please tell which version of gradle is used in your project?

MrCapone:
@PatriceJiang im using gradle plugin v3.4.0 which downloading gradle 5.1.1

PatriceJiang:
~~It seems to be a bug of AndroidStudio I am using, which is perhaps too new.~~  
~~The task is triggered correctly from the console, but it does not run in AndroidStudio.~~ 

It is caused by gradle version change.

MrCapone:
I've tested CppTest from main branch without this commit and then on my branch with it. In both cases assets folder didn't created and REsources didn't copied. So this commit has nothing to do with that issue. Anyway my project from cpp-template still working fine.

PatriceJiang:
The issue was introduced since https://github.com/cocos2d/cocos2d-x/pull/19657,  we may choose another version of gradle or modify gradle build scripts.  @MrCapone

MrCapone:
@PatriceJiang I've created empty project from my branch and tried build apk. Everything works fine. So gradle build scripts in cpp-template already working.

minggo:
Because we need to copy resources, and it doesn't depend any other thing. So i think we can copy resources not using gradle task. Then it will not be affected by gradle.

MrCapone:
not relevant

hbSumrain:
Can you tell me how to resolve this problem ,it also happened to me

MrCapone:
@hbSumrain https://discuss.cocos2d-x.org/t/cocos2d-x-3-17-and-android-studio-3-3-build-issue-or-error/45503/6?u=capone

PR #19705, opened 2019-05-13
fix toupper function call

JohnCoconut:
* Original pointer cast does nothing, as it doesn't change signature
* toupper arugment type should be unsigned char

Reference: https://en.cppreference.com/w/cpp/string/byte/toupper


minggo:
```
  C:\projects\cocos2d-x\cocos\renderer\CCRenderState.cpp(485): error C2039: 'toupper': is not a member of 'std' [C:\projects\cocos2d-x\win32-build\engine\cocos\core\cocos2d.vcxproj]
```

Miss including header files on windows.

minggo (approved):

Issue #10382, opened 2015-02-05
Remove curl dependency from AssetsManager, AssetsManagerEx, Downloader


Yehsam23:
Cocos2d-x v3.4 had a big improved feature.

Smaller Size
Libcurl has been replaced by system network library on mobile platform such as IOS and android, which make game package more lightweight.

But the AssetsManager & Downloader still use curl libs.

Would you change those curl to use HttpClient?

Thanks

minggo:
Yep, we will replace curl in AssetsManager too.
Because it needs more time, so we don't implement it in v3.4.

Yehsam23:
Thanks a lot

stammen:
The Windows 8.1 Universal App version of cocos2d-x would also like curl removed from AssetsManager. We have implemented WinRT native networking but will still need curl for the AssetsManager.

walzer:
:+1: 
Schedule in v3.7

ricardoquesada:
I'll start working on this. I'll implement the "cocos net" abstraction

ricardoquesada:
Native replacements for **iOS**:
- AFNetworking: https://github.com/AFNetworking/AFNetworking
- NSURLConnection / NSURLSession: https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/URLLoadingSystem/URLLoadingSystem.html

Native replacements for **Android**:
- Volley: https://www.youtube.com/watch/?v=yhv8l9F44qo
- OkHTTP: http://square.github.io/okhttp/

Any other ?
@minggo @pandamicro what do you recommend to use? thanks.

ricardoquesada:
@linshun @pandamicro Hey, I need some help. I noticed that we already have some networking code in cocos/network: https://github.com/cocos2d/cocos2d-x/tree/v3/cocos/network

Should AssetManager / DownloadManager use the cocos/network code ( `HttpClient`) ? And I think that was the suggestion made by @Yehsam23. 

If so, I wouldn't need to create the "network" layer... perhaps I only need to create the missing functionality.

Thoughts?

ricardoquesada:
And what is the difference between `AssetsManager` and `AssetsManagerEx` ?
And any particular reason why `AssetsManager` is a subclass of `Node` ?
Thanks.

pandamicro:
## Third party libraries

For native replacement on Android, Cocos Play & Runtime are using : 

android-async-http: https://github.com/loopj/android-async-http

## About HttpClient

[...]

PR #19706, opened 2019-05-13
use okhttp3

PatriceJiang:
use `okhttp` to replace `android-async-http` in CCDownloader android implementation. 

base on https://github.com/cocos-creator/cocos2d-x-lite/pull/1731


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
```diff
@@ -273,122 +82,210 @@ public void run() {
...
+                                        tempFile.renameTo(finalFile);
+                                    } while (false);
+
+                                    if (errStr == null) {
+                                        downloader.onFinish(id, ErrorCode.ERROR_NO_ERROR, null, null);
+                                    }
+                                    else
+                                        downloader.onFinish(id, ErrorCode.ERROR_IMPL_INTERNAL, errStr, null);
+                                } else {
+                                    // 非文件
```
minggo, line 253: comments in chinese

halx99:
why not use curl for all platforms?

minggo:
@halx99 according the issue: https://github.com/cocos2d/cocos2d-x/issues/10382, i think it is about to reduce package size. Use platform network can reduce the package size. But it is pity that it brings JAR lib on Android too, though the package size is smaller than libcurl(curl depends on ssl).

So i think it will be better if not 3rd party library is needed.

PatriceJiang:
re-commit in https://github.com/cocos2d/cocos2d-x/pull/19766

PR #19711, opened 2019-05-14
[ci skip]update release note and change log

minggo:


crazyhappygame:
@minggo what is 3.17.2 release date?
Could you recompile Android with latest ndk r19c?

minggo:
About next Tuesday. Is there any problem with ndk r19c? I will ask the QA team to use ndk 19c to take a look.

crazyhappygame:
As far as I know currently we use ndk16 which is 1.5 year old. With ndk19c we get latest compiler and possibility to compile with C++14, C++17.

PR #19715, opened 2019-05-14
fixed vertex counter for statistic in drawing of batched triangles

druzenko:
In drawing of batched triangles a count of indices (not vertices) appends to a general count for statistic. Therefore for one sprite gl verts is equal six (not four).


minggo:
A sprite uses two triangles, so there is not problem.

PR #19716, opened 2019-05-15
[Texture2D Test17/18] add methods to convert BGRA8888 format

Mee-gu:
convert BGRA8888 to RGBA8888


cocos/renderer/CCTextureUtils.cpp
```diff
@@ -938,6 +949,22 @@ namespace backend { namespace PixelFormatUtils {
         
         return format;
     }
+    
+    PixelFormat convertBGRA8888ToFormat(const unsigned char* data, ssize_t dataLen, PixelFormat format, unsigned char** outData, ssize_t* outDataLen)
+    {
```
minggo, line 954: don't use `ssize_t `, use `size_t` instead

Mee-gu:
update review

minggo (approved):

PR #19719, opened 2019-05-15
Fix removing single symbol to the right of the cursor by Backspace

kxs1024:
Fixed the problem when the string contains only one symbol and the cursor is at the beginning of the string this symbol is removed when Backspace key is pressed.


AIGRIND-LLC:
This is how it works without fix.
bug

AIGRIND-LLC:
@PatriceJiang is there any updates?

PR #16126, opened 2016-07-15
#improve node sort performance by std::sort + long long (_localZOrder) for 64-bit

halx99:

PR #19723, opened 2019-05-16
[Change] the CC_64BITS macro to 64-bit decision.

EricDDK:


cocos/base/ccMacros.h
```diff
@@ -276,7 +276,7 @@ CC_ASSERT(__gl_error_code == GL_NO_ERROR, "Error"); \
  /*********************************/
  /** 64bits Program Sense Macros **/
  /*********************************/
-#if defined(_M_X64) || defined(_WIN64) || defined(__LP64__) || defined(_LP64) || defined(__x86_64)
+#if defined(_M_X64) || defined(_WIN64) || defined(__LP64__) || defined(_LP64) || defined(__x86_64) || defined(__arm64__) || defined(__aarch64__)
 #define CC_64BITS 1
```
crazyhappygame, line 280: Could you explain why we need CC_64BITS define at all?

In code I see only
```
#if CC_64BITS
        std::sort(std::begin(nodes), std::end(nodes), [](_T* n1, _T* n2) {
            return (n1->_localZOrder$Arrival < n2->_localZOrder$Arrival);
        });
#else
        std::sort(std::begin(nodes), std::end(nodes), [](_T* n1, _T* n2) {
            return (n1->_localZOrder == n2->_localZOrder && n1->_orderOfArrival < n2->_orderOfArrival) || n1->_localZOrder < n2->_localZOrder;
        });
#endif
```
which could be simple remove.

minggo: It is for performance, can refer to https://github.com/cocos2d/cocos2d-x/pull/16126 for detail information.

minggo (approved):

cocos/base/ccMacros.h
EricDDK: And the luac performed on the 64-bit and 32-bit luajit is different.

PR #19731, opened 2019-05-17
add subTitle

Mee-gu:
add subtitle


Mee-gu:
add subtitle to prompt the test needs a touch movement

minggo (approved):

PR #19732, opened 2019-05-17
[Android] gradle copy resources to assets folder before `preDebug|ReleaseBuild`

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/pull/19703#issuecomment-493297998

### DIFF
--- a/templates/cpp-template-default/proj.android/app/build.gradle
+++ b/templates/cpp-template-default/proj.android/app/build.gradle
@@ -111,19 +111,22 @@ android {
 }
 
 android.applicationVariants.all { variant ->
-    def project_root_folder = "${buildDir}/../../.."
-    def dest_assets_folder = "${buildDir}/../assets"
+    def project_root_folder = "${projectDir}/../.."
+    def dest_assets_folder = "${projectDir}/assets"
 
     // delete previous files first
     delete dest_assets_folder
-
-    variant.mergeAssets.doFirst {
+    def targetName = variant.name.capitalize()
+    def copyTaskName = "copy${targetName}ResourcesToAssets"
+    
+    tasks.register(copyTaskName) {
         copy {
             from project_root_folder + "/Resources"
             into dest_assets_folder
             exclude "**/*.gz"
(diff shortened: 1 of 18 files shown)


templates/cpp-template-default/proj.android/app/build.gradle
```diff
@@ -111,19 +111,22 @@ android {
 }
 
 android.applicationVariants.all { variant ->
-    def project_root_folder = "${buildDir}/../../.."
-    def dest_assets_folder = "${buildDir}/../assets"
+    def project_root_folder = "${projectDir}/../.."
+    def dest_assets_folder = "${projectDir}/assets"
```
minggo, line 115: What's the value of `projectDir `?

templates/cpp-template-default/proj.android/app/build.gradle
```diff
@@ -111,19 +111,22 @@ android {
...
+    def copyTaskName = "copy${targetName}ResourcesToAssets"
+    
+    tasks.register(copyTaskName) {
         copy {
             from project_root_folder + "/Resources"
             into dest_assets_folder
             exclude "**/*.gz"
         }
     }
+    tasks.getByName("pre${targetName}Build").dependsOn copyTaskName
```
minggo, line 129: What does it mean?

templates/cpp-template-default/proj.android/app/build.gradle
PatriceJiang: According to https://docs.gradle.org/current/userguide/writing_build_scripts.html


>Name | Type | Default Value
>-- | -- | --
>projectDir | File | The directory containing the build script.

templates/cpp-template-default/proj.android/app/build.gradle
PatriceJiang: Schedule task `copyDebug(Release)ResourcesToAsset` to run before `preDebug(Release)Build`, which will always run ahead of build.

PR #19735, opened 2019-05-17
Feature/sprite etc1 alpha support test

Mee-gu:
1. no need to update shader when it had been set before; 
2. set compressed texture data length correctly.

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -403,11 +403,9 @@ void Sprite::setProgramState(backend::ProgramState *programState)
 
 void Sprite::setTexture(Texture2D *texture)
 {
-    if(_programState == nullptr)
-    {
-        auto isETC1 = texture && texture->getAlphaTextureName();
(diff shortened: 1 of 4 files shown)


cocos/2d/CCSprite.cpp
```diff
@@ -403,9 +403,11 @@ void Sprite::setProgramState(backend::ProgramState *programState)
...
 void Sprite::setTexture(Texture2D *texture)
 {
-    auto isETC1 = texture && texture->getAlphaTextureName();
-    updateShaders(positionTextureColor_vert, (isETC1) ? etc1_frag : positionTextureColor_frag);
-    
+    if(_programState == nullptr)
+    {
+        auto isETC1 = texture && texture->getAlphaTextureName();
+        updateShaders(positionTextureColor_vert, (isETC1) ? etc1_frag : positionTextureColor_frag);
+    }
```
minggo: But texture may change between ETC1 and not ETC1.

PR #19737, opened 2019-05-18
 Using auto + Removing redundant else in CREATE_FUNC macro

HTC77:
* Should use `auto` to avoid duplicating type name. auto is optimized
 * No need to else when you have `return` in single if condition


cocos/platform/CCPlatformMacros.h
```diff
@@ -42,18 +42,15 @@ Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
 #define CREATE_FUNC(__TYPE__) \
 static __TYPE__* create() \
 { \
-    __TYPE__ *pRet = new(std::nothrow) __TYPE__(); \
+    auto *pRet = new(std::nothrow) __TYPE__(); \
```
HTC77: Remove `*` after auto

HTC77:
> * You should use `auto` to avoid duplicating type name. auto is optimized
> * No need to else when you have `return` in single if condition

**I** think You should add a macro such as `CREATE_FUNC_WITH_PARAMS(__TYPE__,__PARAMS__,...)` or overload  the CREATE_FUNC() macro and passes params to `init() `method
to create a layer or node that we want to pass some parameters to it. We mostly had to
create this method manually:
`static Balloon* create(Scene* parent);`
and:
	`static Enemy* create(float speed);`

I don't know how to create this macro but I know it is same as creating menus in cocos2d-x:
  `  auto menu = Menu::create(closeItem,startSceneItem, NULL);`

minggo:
@HTC77 i think this PR includes many codes that are not related to the issue.

HTC77:
This is not issue. Just suggest

minggo:
What i mean is that the modification is not the same as the `title`. For example, it includes the modification in `README.md`.

HTC77:
> What i mean is that the modification is not the same as the `title`. For example, it includes the modification in `README.md`.

Sorry I have done one commit for this PR but all of my commits from the past has been appeared here too.
It's a08b32646836a065fa9a920d7f49c331481f957d

minggo:
@HTC77 i think you should create a new branch and use `git cherry-pick` or just modify in the new branch. The new branch should be created based on last v3, not your previous branch.

PR #19738, opened 2019-05-19
erase objects in array

HTC77:
It's same as the `removeObjectsInArray()` method of **CCArray** class


minggo:
@HTC77 please use different PR for different issues.

PR #19739, opened 2019-05-20
[win32] suppress compile warnings 

PatriceJiang:
Suppress warings in libcocos2d: 

- [LNK4099](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4099?view=vs-2019)
> PDB 'filename' was not found with 'object/library' or at 'path'; linking object as if no debug info
The linker was unable to find your .pdb file. Copy it into the directory that contains object/library.
- [C4244](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-levels-3-and-4-c4244?view=vs-2019)
>  'conversion' conversion from 'type1' to 'type2', possible loss of data 
- [C4305](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4305?view=vs-2019)
> This warning is issued when a value is converted to a smaller type in an initialization or as a constructor argument, resulting in a loss of information.

### DIFF
--- a/cmake/Modules/CocosBuildHelpers.cmake
+++ b/cmake/Modules/CocosBuildHelpers.cmake
@@ -188,7 +188,7 @@ function(setup_cocos_app_config app_name)
         set_target_properties(${app_name} PROPERTIES MACOSX_BUNDLE 1)
     elseif(MSVC)
         # visual studio default is Console app, but we need Windows app
-        set_property(TARGET ${app_name} APPEND PROPERTY LINK_FLAGS "/SUBSYSTEM:WINDOWS")
+        set_property(TARGET ${app_name} APPEND_STRING PROPERTY LINK_FLAGS "/SUBSYSTEM:WINDOWS /IGNORE:4099")
     endif()
     # auto mark code files for IDE when mark app
     if(XCODE OR VS)
--- a/cocos/CMakeLists.txt
+++ b/cocos/CMakeLists.txt
@@ -153,7 +153,7 @@ endif()
 if(WINDOWS)
     # precompiled header. Compilation time speedup ~4x.
     target_sources(cocos2d PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/precheader.cpp")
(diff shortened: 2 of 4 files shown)


crazyhappygame (commented):
Please do not commit this. In modern cmake flags shouldn't be set .

crazyhappygame:
Additionally all those errors are bugs in cocos2d-x code and should be fixed

PatriceJiang:
These three flags can reduce warning count from 1500+ to around 50 in Visual Studio, while clang does not report these warnings by default. I think treating all warnings as errors is somewhat inconvenient.

crazyhappygame:
Please read https://foonathan.net/blog/2018/10/17/cmake-warnings.html
Those flags should be set as Private. By default it is public. That means that you disabled those warning for my game as well

crazyhappygame:
About clang please set wall, extra, pedantic*  and you will see even more.

crazyhappygame:
@JohnCoconut with clang-tidy is cleaning cocos2dx code. I believe that after that most of the warning will be fixed

JohnCoconut:
>  'conversion' conversion from 'type1' to 'type2', possible loss of data
> This warning is issued when a value is converted to a smaller type in an initialization or as a constructor argument, resulting in a loss of information.

It sounds similar to this option in clang-tidy,  [cppcoreguidelines-narrowing-conversions](http://releases.llvm.org/7.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines-narrowing-conversions.html)

> PDB 'filename' was not found with 'object/library' or at 'path'; linking object as if no debug info
The linker was unable to find your .pdb file. Copy it into the directory that contains object/library.

I am not sure about this one.

 I will push a commit on `clang-tidy-cppcoreguidelines-narrowing-conversions`, maybe on tomorrow. Could you compare number of warnings after applying clang-tidy options? I don't have a Windows machine or MSVC to test.

minggo:
Yep, we should fix warnings instead of avoid them.

PR #19743, opened 2019-05-21
remove redundant semicolon

JohnCoconut:
No code logic is touched.

This is a step to enable more compiler warnings.

It removes redundant semicolons like this,

```cpp
int i;;

void foo() 
{
};
```

### DIFF
--- a/cocos/2d/CCActionCatmullRom.cpp
+++ b/cocos/2d/CCActionCatmullRom.cpp
@@ -38,7 +38,7 @@
 
 using namespace std;
 
-NS_CC_BEGIN;
+NS_CC_BEGIN
 
 /*
  *  Implementation of PointArray
@@ -526,4 +526,4 @@ CatmullRomBy* CatmullRomBy::reverse() const
     return CatmullRomBy::create(_duration, reverse);
 }
 
-NS_CC_END;
+NS_CC_END
--- a/cocos/2d/CCActionCatmullRom.h
+++ b/cocos/2d/CCActionCatmullRom.h
@@ -41,7 +41,7 @@
 #include "2d/CCActionInterval.h"
 #include "math/CCGeometry.h"
 
-NS_CC_BEGIN;
+NS_CC_BEGIN
 
 class Node;
 
@@ -343,6 +343,6 @@ extern CC_DLL Vec2 ccCardinalSplineAt(const Vec2 &p0, const Vec2 &p1, const Vec2
 // end of actions group
 /// @}
 
-NS_CC_END;
+NS_CC_END
 
 #endif // __CCACTION_CATMULLROM_H__
--- a/cocos/2d/CCActionEase.cpp
+++ b/cocos/2d/CCActionEase.cpp
@@ -159,33 +159,33 @@ ActionEase* CLASSNAME::reverse() const { \
     return REVERSE_CLASSNAME::create(_inner->reverse()); \
 }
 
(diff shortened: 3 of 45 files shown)


JohnCoconut:
I wish the macro usage could be more uniform. If you check the file changed under this commit, you can see some macro usage ends with a semicolon, while some doesn't. I hope all of them end with a semicolon. But it's hard for example like this.

```cpp
#define SOME_MACRO void foo() \
{ \
}
```

How do we call 
```cpp
SOME_MACRO;
```
without triggering a warning on  `;`?

stevetranby:
Isn't this a breaking change for users who didn't write semicolons after the macro usage? Or are these all internal macros?

Why not just enable all warnings and then disable specific ones as needed during your warnings tests? `-Wall` and other full warning checks do slow down compile times, so not advised to try enable every warning on the main new game template, at least.

JohnCoconut:
Hi, I couldn't see your reply on github, only in email. Did you post
this in the PR?

> Isn't this a breaking change for users who didn't write semicolons after the macro usage? Or are these all internal macros?

It is a breaking change. It slipped my mind when I did it. I think I
may have to revert it.

> Why not just enable all warnings and then disable specific ones as needed during your warnings tests?

Unfortunately there is no specific option to disable warnings on extra
semicolon if `-pedantic` is added.

Or we may leave macro be and change its usage code instead?

What's your opinion?

stevetranby:
Hey John, yeah sorry I deleted it because I thought it was a bit reactionary and ultimately decided that if anyone is still updating and maintaining their game projects that this one at least will throw compiler errors and thus the breaking change isn't all that bad ... ???

Anyway, not sure the best approach to take for 3.17.2+. My gut reaction is always sort of "leave the breaking changes to the new metal-support branch or v4.x". 

However, after re-reading (and partly why I deleted my reply) most of this is meant for internal use so while it may break others, they should just fix their usage.

Lastly, there seems to be an inconsistency, however, where some MARCOs usage code has semi and others don't. If you're going to make breaking changes I'd probably prefer the usage code having semicolons and leave them out of the MACRO definition. This is probably a poor coding style, but it just looks more consistent and uniform with the rest of my/cocos2d code.

Ultimately maybe just don't worry about it and wait for another to complain first. ¯\_(ツ)_/¯

JohnCoconut:
Please do feel free to review my PR. I am open to your suggestions.

I made three breaking changes in `CC_DISALLOW_COPY_AND_ASSIGN`,
`CC_PROPERTY`, `CC_PROPERTY_PASS_BY_REF` by removing semicolon after
the macro definition. I don't think people will like it if their code
suddenly stops compiling. But it's really trivial to fix the code.
@mingo, what's your opinion?

About inconsistency of semicolon, some macros are function
definitions, which aren't supposed to be followed by semicolon. Some
are function declarations, which are supposed to be followed by
semicolon. It's hard to tell the difference between them from macro
name itself. I am not sure how to deal with them.

stevetranby:
Yeah that's the reason I said "probably a poor coding style" because my preference of style is not the technically correct method, macros are fun ;)

Again, it's probably fine to keep as-is since it's the best kind of breaking change since it fails at compile-time ... at least in 99% of usage, but I guess this is c++ so the error messages might not be specific about semicolon in every case, eh.

If you want my opinion I'd revert the three you mentioned and then add them back into the metal-branch since that will require at least a few code changes as well, but I'll leave this decision to you and mingo. Sorry to waste your time on this.

Issue #16113, opened 2016-07-13
rendertexture's savetoFile problem


reasonMix:
when draw a label to the render texture and call the savetoFile to create the png file.the png file have white border on ios platform. because the label use ALPHA_NON_PREMULTIPLIED blend mode. and on ios platform the image-ios.mm savetoFile only use the kCGImageAlphaPremultipliedLast flag

ricardoquesada:
thanks. could you post the code to reproduce it? thanks.

reasonMix:
@ricardoquesada 
  local text = cc.Label:createWithTTF('hello world','font/yours.ttf',40)

  local target = cc.RenderTexture:create(winSize.width, winSize.height, 2)
  target:beginWithClear(0,0,0,0)
  text:setPosition(winSize.width / 2,winSize.height/2)
  text:visit()
  target:endToLua()
  target:saveToFile('aaa.png',1)

ricardoquesada:
Thanks. But I couldn't reproduce it. The image that I got on an iPhone is the following, and it seems to be Ok. is that right? I'm using C++ code instead of Lua though.

**UPDATE:** Image had the letters in white... generating a new one in red (see below)

reasonMix:
@ricardoquesada where is the png file,i can't seen it

ricardoquesada:
@reasonMix Uploading it again... In case it doesn't work, this is the test case that I created:
https://github.com/cocos2d/cocos2d-x/pull/16180

![issue16113](https://cloud.githubusercontent.com/assets/232330/17009687/1b2ad9d4-4eb2-11e6-9c83-fabdaf14ee19.png)

I couldn't find any white border in this image.

reasonMix:
@ricardoquesada please paste your code,do you create the label with ttf?

ricardoquesada:
@reasonMix:  the code is here, in this Pull Request: https://github.com/cocos2d/cocos2d-x/pull/16180

reasonMix:
@ricardoquesada i am wrong,in mac platform,the png file will have a black border

ricardoquesada:
@reasonMix this is the image from the Mac:
![issue16113](https://cloud.githubusercontent.com/assets/232330/17034093/fb52e60e-4f35-11e6-8e55-f1906f65940e.png)

reasonMix:
@ricardoquesada use the white color，will see the black border
[...]

PR #19748, opened 2019-05-21
allow set alpha premultiplied value for texture

drelaptop:
allow set premultiplied value manually, this will solve render texture grey border

issues like this:

image

- render a white label in white layer to render texture, but grey border appears
- set premultiplied value to `true` fixed it, after that render texture become to whole white.

### DIFF
--- a/cocos/renderer/CCTexture2D.cpp
+++ b/cocos/renderer/CCTexture2D.cpp
@@ -554,6 +554,11 @@ bool Texture2D::hasPremultipliedAlpha() const
     return _hasPremultipliedAlpha;
 }
 
+void Texture2D::setPremultipliedAlpha(bool premultiplied)
+{
+    _hasPremultipliedAlpha = premultiplied;
+}
+
 bool Texture2D::initWithData(const void *data, ssize_t dataLen, Texture2D::PixelFormat pixelFormat, int pixelsWide, int pixelsHigh, const Size& /*contentSize*/)
 {
     CCASSERT(dataLen>0 && pixelsWide>0 && pixelsHigh>0, "Invalid size");
--- a/cocos/renderer/CCTexture2D.h
+++ b/cocos/renderer/CCTexture2D.h
@@ -372,6 +372,9 @@ public:
     /** Whether or not the texture has their Alpha premultiplied. */
     bool hasPremultipliedAlpha() const;
     
+    /** Allow set alpha premultiplied value when know that from outer. */
+    void setPremultipliedAlpha(bool premultiplied);
+    
     /** Whether or not the texture has mip maps.*/
     bool hasMipmaps() const;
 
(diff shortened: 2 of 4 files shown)


drelaptop:
this PR related with https://github.com/cocos2d/cocos2d-x/issues/16113 partly, but now fix saveToFile grey border issue. @PatriceJiang

PatriceJiang (approved):

minggo:
`premultiplied` flag is used to set blend function, i don't think it is a good idea to set it directly in texture2d

drelaptop:
or manually do like:
```C++
auto texture = renderTexture->getSprite()->getTexture();
auto sprite = Sprite::createWithTexture(texture);
sprite->setBlendFunc(BlendFunc::ALPHA_PREMULTIPLIED); //need to do this!
```
 it's so internal, much hard to find it out.

minggo:
Yep, the design of premultiplied is not good, but it is hard to change it now.

minggo:
close this one as https://github.com/cocos2d/cocos2d-x/pull/19782 is merged

rh101:
@minggo This needs to be re-assessed.  There is a problem with this when using a RenderTexture, since it creates a Texture2D internally.  The problem with this Texture2D is that `_hasPremultipliedAlpha` set to false, so any call to `RenderTexture->getSprite()->getTexture()->hasPremultipliedAlpha()` will always return false.

The sprite inside the RenderTexture is creatd with the following:

```
        // retained
        setSprite(Sprite::createWithTexture(_texture));
        ...
        _sprite->setBlendFunc( BlendFunc::ALPHA_PREMULTIPLIED );
        ...
```

So, while the sprite has the blending set to `BlendFunc::ALPHA_PREMULTIPLIED`, the internal Texture2D still has premultiplied alpha set to false, and that on it's own is conflicting information.  If we store references to the Texture2D in parts of our code, and discard the sprite, then there is no way to know that it's actually premultiplied alpha (since `Texture2D::hasPremultipliedAlpha()` is false).

Either a public function is introduced to Texture2D to set the _hasPremultipliedAlpha, as done in the PR above ([](https://github.com/cocos2d/cocos2d-x/pull/19748/commits/d92638e9f778f92e85840303b8dc2fe7716acd05) 

...or set it based on the flag in ccConfig.h:

```
#ifndef CC_ENABLE_PREMULTIPLIED_ALPHA
# define CC_ENABLE_PREMULTIPLIED_ALPHA 1
#endif
```
So always assume that all textures created by default to be based off whatever CC_ENABLE_PREMULTIPLIED_ALPHA is:
```
Texture2D::Texture2D()
...
, _hasPremultipliedAlpha(CC_ENABLE_PREMULTIPLIED_ALPHA != 0)
...
{
}
```
That makes sense, since the engine internally would assume all textures are premultiplied or not based on that flag.

PR #19751, opened 2019-05-22
[Node::Light] fix no light effects on iPhone6 and iPhone6s

Mee-gu:
fix no light effects on iPhone6 and iPhone6s


minggo:
Why it will fix the issue?

PR #19759, opened 2019-05-26
fix: FastTMXLayer does not reflect opacity and anchor point

kackyt:
Fixed a bug that FastTMXLayer does not reflect opacity and anchor point.

### DIFF
--- a/cocos/2d/CCFastTMXLayer.cpp
+++ b/cocos/2d/CCFastTMXLayer.cpp
@@ -152,8 +152,9 @@ void TMXLayer::draw(Renderer *renderer, const Mat4& transform, uint32_t flags)
     if( flags != 0 || _dirty || _quadsDirty || isViewProjectionUpdated)
     {
         Size s = Director::getInstance()->getVisibleSize();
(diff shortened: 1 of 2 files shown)


cocos/2d/CCFastTMXLayer.cpp
```diff
@@ -152,8 +152,9 @@ void TMXLayer::draw(Renderer *renderer, const Mat4& transform, uint32_t flags)
     if( flags != 0 || _dirty || _quadsDirty || isViewProjectionUpdated)
     {
         Size s = Director::getInstance()->getVisibleSize();
-        auto rect = Rect(Camera::getVisitingCamera()->getPositionX() - s.width * 0.5f,
-                     Camera::getVisitingCamera()->getPositionY() - s.height * 0.5f,
+        Vec2 anchor = getAnchorPoint();
```
minggo: Use cont reference is better.

kackyt: Thank you. I fixed it.

Issue #861, opened 2012-04-23
Cannot activate C++11 with libc++


Klaim:
Hi, I'd like to use C++11 features but I can't find a way to compile Cocos2d x-code projects (the template ones).
I first generate the template project and then change the C++ dialect to either C++11 or gnu-C++11. 

```
/Users/joel-lamotte/projects/Test/test_cocos2d_chipmunk/test_cocos2d_chipmunk/libs/cocos2dx/textures/CCTexturePVR.cpp:110:69: error: constant expression evaluates to -1 which cannot be narrowed to type 'unsigned int' [-Wc++11-narrowing]
        { kPVRTextureFlagTypePVRTC_2,   GL_COMPRESSED_RGBA_PVRTC_2BPPV1_IMG, -1, -1,    2,      true, kCCTexture2DPixelFormat_PVRTC2 },
                                                                             ^~
/Users/joel-lamotte/projects/Test/test_cocos2d_chipmunk/test_cocos2d_chipmunk/libs/cocos2dx/textures/CCTexturePVR.cpp:110:69: note: override this message by inserting an explicit cast
        { kPVRTextureFlagTypePVRTC_2,   GL_COMPRESSED_RGBA_PVRTC_2BPPV1_IMG, -1, -1,    2,      true, kCCTexture2DPixelFormat_PVRTC2 },
                                                                             ^~
                                                                             st)tic_cast<unsigned         int>(
/Users/joel-lamotte/projects/Test/test_cocos2d_chipmunk/test_cocos2d_chipmunk/libs/cocos2dx/textures/CCTexturePVR.cpp:110:73: error: constant expression evaluates to -1 which cannot be narrowed to type 'unsigned int' [-Wc++11-narrowing]
        { kPVRTextureFlagTypePVRTC_2,   GL_COMPRESSED_RGBA_PVRTC_2BPPV1_IMG, -1, -1,    2,      true, kCCTexture2DPixelFormat_PVRTC2 },
                                                                                 ^~
/Users/joel-lamotte/projects/Test/test_cocos2d_chipmunk/test_cocos2d_chipmunk/libs/cocos2dx/textures/CCTexturePVR.cpp:110:73: note: override this message by inserting an explicit cast
        { kPVRTextureFlagTypePVRTC_2,   GL_COMPRESSED_RGBA_PVRTC_2BPPV1_IMG, -1, -1,    2,      true, kCCTexture2DPixelFormat_PVRTC2 },
                                                                                 ^~
[...]

PR #19760, opened 2019-05-26
remove unnecessary `tmp` parameter in `uthash.h # HASH_ITER`.

mgcation:
The `tmp` is unused macro parameter.

so, I just remove it.

and also solve redefinition problem `pTmp##__dict__##__el__` in `CCDictionary # CCDICT_FOREACH`.
(even though `CCDictionary` was deprecated codes)

---

(I've erased the wrong code and forcibly pushed it 2 times. plz ignore them)

### DIFF
--- a/cocos/base/uthash.h
+++ b/cocos/base/uthash.h
@@ -857,13 +857,11 @@ do {                                                                            
             (HASH_BLOOM_BYTELEN)))
 
 #ifdef NO_DECLTYPE
-#define HASH_ITER(hh,head,el,tmp)                                                \
-for((el)=(head), (*(char**)(&(tmp)))=(char*)((head)?(head)->hh.next:NULL);       \
-  el; (el)=(tmp),(*(char**)(&(tmp)))=(char*)((tmp)?(tmp)->hh.next:NULL)) 
+#define HASH_ITER(hh,head,el)                                                    \
+  for( (el)=(head);  el;  (*(char**)(&(el)))=(char*)((el)?(el)->hh.next:NULL) )
 #else
-#define HASH_ITER(hh,head,el,tmp)                                                \
-for((el)=(head),(tmp)=DECLTYPE(el)((head)?(head)->hh.next:NULL);                 \
-  el; (el)=(tmp),(tmp)=DECLTYPE(el)((tmp)?(tmp)->hh.next:NULL))
+#define HASH_ITER(hh,head,el)                                                    \
+  for( (el)=(head);  el;  (el)=DECLTYPE(el)((el)?(el)->hh.next:NULL) )
 #endif
 
 /* obtain a count of items in the hash */
--- a/cocos/2d/CCActionManager.cpp
+++ b/cocos/2d/CCActionManager.cpp
@@ -428,8 +428,7 @@ ssize_t ActionManager::getNumberOfRunningActions() const
 {
     ssize_t count = 0;
     struct _hashElement* element = nullptr;
-    struct _hashElement* tmp = nullptr;
-    HASH_ITER(hh, _targets, element, tmp)
+    HASH_ITER(hh, _targets, element)
     {
         count += (element->actions ? element->actions->num : 0);
     }
--- a/cocos/deprecated/CCDictionary.cpp
+++ b/cocos/deprecated/CCDictionary.cpp
@@ -102,10 +102,10 @@ __Array* __Dictionary::allKeys()
 
     __Array* array = __Array::createWithCapacity(iKeyCount);
 
-    DictElement *pElement, *tmp;
+    DictElement *pElement;
     if (_dictType == kDictStr)
     {
-        HASH_ITER(hh, _elements, pElement, tmp) 
+        HASH_ITER(hh, _elements, pElement)
         {
             __String* pOneKey = new (std::nothrow) __String(pElement->_strKey);
             array->addObject(pOneKey);
@@ -114,7 +114,7 @@ __Array* __Dictionary::allKeys()
     }
     else if (_dictType == kDictInt)
     {
-        HASH_ITER(hh, _elements, pElement, tmp) 
+        HASH_ITER(hh, _elements, pElement)
         {
             __Integer* pOneKey = new (std::nothrow) __Integer(static_cast<int>(pElement->_intKey));
             array->addObject(pOneKey);
@@ -131,11 +131,11 @@ __Array* __Dictionary::allKeysForObject(Ref* object)
     if (iKeyCount <= 0) return nullptr;
     __Array* array = __Array::create();
 
-    DictElement *pElement, *tmp;
+    DictElement *pElement;
 
     if (_dictType == kDictStr)
     {
-        HASH_ITER(hh, _elements, pElement, tmp) 
+        HASH_ITER(hh, _elements, pElement)
         {
             if (object == pElement->_object)
             {
@@ -147,7 +147,7 @@ __Array* __Dictionary::allKeysForObject(Ref* object)
     }
     else if (_dictType == kDictInt)
     {
-        HASH_ITER(hh, _elements, pElement, tmp) 
+        HASH_ITER(hh, _elements, pElement)
         {
             if (object == pElement->_object)
             {
@@ -332,8 +332,8 @@ void __Dictionary::removeObjectForElememt(DictElement* pElement)
 
 void __Dictionary::removeAllObjects()
 {
-    DictElement *pElement, *tmp;
-    HASH_ITER(hh, _elements, pElement, tmp) 
+    DictElement *pElement;
+    HASH_ITER(hh, _elements, pElement)
     {
         HASH_DEL(_elements, pElement);
         pElement->_object->release();
@@ -572,8 +572,7 @@ __Dictionary* __Dictionary::clone() const
     Clonable* obj = nullptr;
     if (_dictType == kDictInt)
     {
-        DictElement* tmp = nullptr;
-        HASH_ITER(hh, _elements, element, tmp)
+        HASH_ITER(hh, _elements, element)
         {
             obj = dynamic_cast<Clonable*>(element->getObject());
             if (obj)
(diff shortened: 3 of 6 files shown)


cocos/base/uthash.h
```diff
@@ -857,13 +857,11 @@ do {
...
-for((el)=(head), (*(char**)(&(tmp)))=(char*)((head)?(head)->hh.next:NULL);       \
-  el; (el)=(tmp),(*(char**)(&(tmp)))=(char*)((tmp)?(tmp)->hh.next:NULL)) 
+#define HASH_ITER(hh,head,el)                                                    \
+  for( (el)=(head);  el;  (*(char**)(&(el)))=(char*)((el)?(el)->hh.next:NULL) )
 #else
-#define HASH_ITER(hh,head,el,tmp)                                                \
-for((el)=(head),(tmp)=DECLTYPE(el)((head)?(head)->hh.next:NULL);                 \
-  el; (el)=(tmp),(tmp)=DECLTYPE(el)((tmp)?(tmp)->hh.next:NULL))
+#define HASH_ITER(hh,head,el)                                                    \
+  for( (el)=(head);  el;  (el)=DECLTYPE(el)((el)?(el)->hh.next:NULL) )
```
mgcation, line 864: [line #861](https://github.com/cocos2d/cocos2d-x/pull/19760/files#diff-426cb1038ce3f085862ce780d57b0c74R861) and  [line #864](https://github.com/cocos2d/cocos2d-x/pull/19760/files#diff-426cb1038ce3f085862ce780d57b0c74R864) are where the actual changes occurred.

The rest removed only the `tmp variables`.

minggo: I think it is needed. For example
```c++
HASH_ITER(hh, collision_handler_hash, current)
{
    ...
    HASH_DEL(collision_handler_hash,current);
}
```

As you see, current may be deleted in the loop, so use a tmp variable to record the next element to avoid crash of this situation.

mgcation: thanks to consider this!

I see...
agree with your example.
but.. I think `tmp` is unnecessary to be visible at outside scope.
so, what about this?

```suggestion
if( void* tmp = NULL );                                                          \
else                                                                             \
for((el)=(head),(tmp)=(head)?(head)->hh.next:NULL;                               \
el; (el)=DECLTYPE(el)(tmp),(tmp)=(tmp)?(tmp)->hh.next:NULL)
```

mgcation: thanks to consider this!

I see...
I agree with your example.
but.. I think the `tmp` variable doesn't need to be visible from outer scope.
so, what about this?

```suggestion
if( void* tmp = NULL );                                                          \
else                                                                             \
for((el)=(head),(tmp)=(head)?(head)->hh.next:NULL;                               \
el; (el)=DECLTYPE(el)(tmp),(tmp)=(tmp)?(DECLTYPE(el)(tmp))->hh.next:NULL)
```

minggo: Sorry, i don't quite understand the codes:
```c++
if( void* tmp = NULL );                                                          \
else
```

mgcation: It is equivalent to
```cpp
{ // start of local scope
    void* tmp = NULL
    if (tmp);
    else
        // for(...) will here, so tmp is visible
} // end of local scope
```

or make local variable by nested-for

```cpp
for(bool _loopOnce=true; _loopOnce;)
    for(void* tmp = NULL; _loopOnce; _loopOnce=false)
        for(/*init*/; /*cond*/; /*increment*/)
```

but so dirty..

minggo: I agree to hide tmp internally, but could you please provide a full implementation. What's the `HASH_ITER ` look like?

mgcation: ok.
to see well, I show you two version (before / after).

### before
```cpp
#ifdef NO_DECLTYPE
#define HASH_ITER(hh,head,el,tmp)                                                \
for((el)=(head), (*(char**)(&(tmp)))=(char*)((head)?(head)->hh.next:NULL);       \
  el; (el)=(tmp),(*(char**)(&(tmp)))=(char*)((tmp)?(tmp)->hh.next:NULL)) 
#else
#define HASH_ITER(hh,head,el,tmp)                                                \
for((el)=(head),(tmp)=DECLTYPE(el)((head)?(head)->hh.next:NULL);                 \
  el; (el)=(tmp),(tmp)=DECLTYPE(el)((tmp)?(tmp)->hh.next:NULL))
#endif
```

### after
```cpp
#ifdef NO_DECLTYPE
#define HASH_ITER(hh,head,el)                                                    \
if(auto tmp = el);                                                               \
else                                                                             \
  for( (el)=(head), (*(char**)(&(tmp)))=(char*)((head)?(head)->hh.next:NULL);    \
    el; (el)=(tmp), (*(char**)(&(tmp)))=(char*)((tmp )?(tmp )->hh.next:NULL) )
#else
#define HASH_ITER(hh,head,el)                                                    \
if(auto tmp = el);                                                               \
else                                                                             \
  for( (el)=(head), tmp=DECLTYPE(el)((head)?(head)->hh.next:NULL);               \
    el; (el)=(tmp), tmp=DECLTYPE(el)((tmp )?(tmp )->hh.next:NULL) )
#endif
```

it is not easy to make typecast `void*` to `decltype(el)` when `NO_DECLTYPE` condition.....

so I use auto.

can I use auto when `NO_DECLTYPE` cond?

minggo: > can I use auto when NO_DECLTYPE cond?

I am not sure. It seems the codes are designed to run in C and C++.

mgcation: by the way,
`if(auto tmp = el)` can be true. so `else` may not be executed.
it is exactly bad code....

mgcation: it is new version.

```cpp
#ifdef NO_DECLTYPE
#define HASH_ITER(hh,head,el)                                                    \
if(void* tmp = NULL);                                                            \
else                                                                             \
for((el)=(head), (*(char**)(&(tmp)))=(char*)((head)?(head)->hh.next:NULL);       \
  el; (*(char**)(&(el)))=(char*)(tmp),tmp=(el)?(el)->hh.next:NULL)
#else
#define HASH_ITER(hh,head,el)                                                    \
if(void* tmp = NULL);                                                            \
else                                                                             \
for((el)=(head), (tmp)=DECLTYPE(el)((head)?(head)->hh.next:NULL);                \
  el; (el)=DECLTYPE(el)(tmp),tmp=(el)?(el)->hh.next:NULL)
#endif
```

using the `tmp` as storage only.
and comma operator guarantees left to right evaluation.

minggo: Why use `if`, can just define it like this:
```c++
#ifdef NO_DECLTYPE
#define HASH_ITER(hh,head,el)                                                    \
void* tmp = NULL;                                                            \
for((el)=(head), (*(char**)(&(tmp)))=(char*)((head)?(head)->hh.next:NULL);       \
  el; (*(char**)(&(el)))=(char*)(tmp),tmp=(el)?(el)->hh.next:NULL)
#else
#define HASH_ITER(hh,head,el)                                                    \
void* tmp = NULL;                                                            \                                                                           
for((el)=(head), (tmp)=DECLTYPE(el)((head)?(head)->hh.next:NULL);                \
  el; (el)=DECLTYPE(el)(tmp),tmp=(el)?(el)->hh.next:NULL)
#endif
```

??

mgcation: This is because the 'tmp' variable is visible from the outside.

``` cpp
{
    int tmp; // error : we can't declare tmp here. (redundant declaration)

    HASH_ITER(...) {
        // body of hash iter
    }
}
```

because this code will precompile to

```cpp
{
    int tmp; // error : we can't declare tmp here.

    void* tmp = NULL; // because of here. (redundant declaration)
    for(...) {
        // body of hash iter
    }
}
```

minggo: The usage is tricky, so please add some comments in the codes.

mgcation: I agree. 
However, before committing the code, I thought of it.

Rather than using an `if(void* tmp = NULL); else { }`,
I would like to ask about using an `nested-for statement like below code` that is complex but does not confuse contributors.




```cpp
// nested-for statement version
// --------------------



#ifdef NO_DECLTYPE /** @see #else section which is same code but more modern */
#define HASH_ITER(hh,head,el)                                                    \
for(bool __tmp_loop_once__=true; __tmp_loop_once__; )                            \
  for(void* __tmp_next_el__ = NULL; __tmp_loop_once__; __tmp_loop_once__=false)  \
    for((el)=(head),                                                             \
        (*(char**)(&(__tmp_next_el__)))=(char*)((head)?(head)->hh.next:NULL);    \
        el;                                                                      \
        (*(char**)(&(el)))=(char*)(__tmp_next_el__),                             \
        __tmp_next_el__=(el)?(el)->hh.next:NULL)
#else
#define HASH_ITER(hh,head,el)                                                    \
for(bool __tmp_loop_once__=true; __tmp_loop_once__; )                            \
  for(void* __tmp_next_el__ = NULL; __tmp_loop_once__; __tmp_loop_once__=false)  \
    /** @note :                                                                  \
          double nested-for loop used to declare '__tmp_next_el__'               \
          in the local scope. */                                                 \
                                                                                 \
    for( (el)=(head),                                                            \
         (__tmp_next_el__)=DECLTYPE(el)((head)?(head)->hh.next:NULL);            \
                                                                                 \
         el;                                                                     \
                                                                                 \
         (el)=DECLTYPE(el)(__tmp_next_el__),                                     \
         __tmp_next_el__=(el)?(el)->hh.next:NULL                                 \
       )
#endif
```

minggo: It is hard to understand too.

mgcation: Then, should we just close this pr..?
Actually, there doesn't seem to be a big issue. (It is just about code quality.)
I just want to give a little help to a grateful engine.

I confuse how to contribute well.

minggo: I think we can close this PR though you did a great thing. You can try to fix issues in the issue system, or help to review codes, or join the forum to help to other developers.

Thanks again.

mgcation: Thanks for your kindness!

PR #19761, opened 2019-05-26
Update CREATE_FUNC macro

HTC77:
Removed unnecessary `else` and used `auto` for optimization

### DIFF
--- a/cocos/platform/CCPlatformMacros.h
+++ b/cocos/platform/CCPlatformMacros.h
@@ -42,18 +42,15 @@ Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
 #define CREATE_FUNC(__TYPE__) \
 static __TYPE__* create() \
 { \
-    __TYPE__ *pRet = new(std::nothrow) __TYPE__(); \
+    auto pRet = new(std::nothrow) __TYPE__(); \
     if (pRet && pRet->init()) \
     { \
         pRet->autorelease(); \
         return pRet; \
     } \
-    else \
-    { \
-        delete pRet; \
-        pRet = nullptr; \
-        return nullptr; \
-    } \
+    delete pRet; \
+    pRet = nullptr; \
+    return nullptr; \
 }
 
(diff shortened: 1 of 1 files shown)


cocos/platform/CCPlatformMacros.h
```diff
@@ -42,18 +42,15 @@ Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
...
         return pRet; \
     } \
-    else \
-    { \
-        delete pRet; \
-        pRet = nullptr; \
-        return nullptr; \
-    } \
+    delete pRet; \
+    pRet = nullptr; \
```
JohnCoconut, line 52:
```cpp
pRet = nullptr;
```

`pRet` is a local variable. Is it necessary to set it to `nullptr`? It cannot be reached from outside if `if (pRet && pRet->init()) ` fails.

HTC77: I think It's for releasing memory from pRet

JohnCoconut: `delete pRet;` frees memory, but it doesn't set `pRset` to null. If someone attempt to use pRet now, the program may keep running, and it is really to use.

`pRet = nullptr;` sets `pRet` to null. If someone attempt to use `pRet` now, the program just crashed immediately. However, no one can use `pRet`, because we return a nullptr.

minggo: Yep, it is not need.

And i don't know what't the problem of `else`?

HTC77: I didn't put last 3 lines at the end of this macro
I just removed `else` and used `auto`
Thouse lines was basically in cocos
I thoght `pRet = nullptr;` is additional command for releasing memory. Because C++ doesn't manage memory byself like Java

HTC77: `else` is redundant. no need to use when we have `return` in `if` condition

PR #19764, opened 2019-05-27
fix location

Mee-gu:
fix location


Mee-gu:
issue ref from [here](https://discuss.cocos2d-x.org/t/metal-support-alpha1-released/46282/39)

PR #19766, opened 2019-05-27
[android] Use okhttp3 to implement Downloader & WebSocket

PatriceJiang:
- replace `android-async-http` & `httpclient` with `okhttp3` & `okio`
- remove dependency to `libwebsockets` & `openssl`

## why `okhttp`
- current downloader implementation has issues on IPV6 network & SSL, `okhttp` is more reliable. 
- developers needed to provide a cacert.pem when using `wss://` with libwebsockets, that is confusing for many developers. after this commit, developers don't have to concern about that for most cases. (we are planning to use RocketSocket on iOS, cacert.pem will not be required for Apple devices by then)
- reduce package size

|dependencies|size|
|---|---|
|-~~libcrypto.a~~|3.89MB|
|-~~libssl.a~~|690KB|
|-~~libwebsocket.a~~|371KB|
|-~~android-async-http-1.4.9.jar~~|104 KB|
|-~~httpclient-4.4.1.1.jar~~|1.11MB|
|+okio-1.15.0.jar |86.7KB|
|+okhttp-3.12.2.jar|413KB|


cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+        void unwatchConnection(int64_t id)
+        {
+            std::lock_guard<std::mutex> guard(_mtx);
+            auto itr = _conditions.find(id);
+            if(itr == _conditions.end())
+            {
+                return;
+            }
+            _conditions.erase(itr);
```
JohnCoconut: `erase` member function has a overload `size_type erase( const key_type& key );`  [on cppreference](https://en.cppreference.com/w/cpp/container/unordered_map/erase)

It does nothing if the key doesn't exists, and delete the elements if key exists. This code can be simplified to

`_conditions.erase(id)`.

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,389 @@
...
+        std::condition_variable cond;
+    };
+
+    class ConnectionCV {
+    public:
+        ConnectionCV() = default;
+
+        virtual ~ConnectionCV() {
+            std::lock_guard<std::mutex> guard(_mtx);
+            _conditions.clear();
```
JohnCoconut, line 70:
```cpp
std::lock_guard<std::mutex> guard(_mtx);
_conditions.clear();
```

Please forgive my ignorance. I am still rough with C++ concurrency stuff. 

Do we need a lock in destructor to clear the unordered map? Are we allowed to call a object's destructor from two threads simultaneously?

cocos/network/CCWebSocket-android.cpp
PatriceJiang: Yes!

cocos/network/CCWebSocket-android.cpp
PatriceJiang: The lock is used to protect data `_conditions`, in general thinking, the field can be modified in another thread during destruction.

Here the class `ConnectionCV` has only one `static` instance, and it will only invoke one time during the process exiting procedure. So is "not required" here, but I think we should keep it to make it easy to understand.

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,384 @@
...
+                std::lock_guard<std::mutex> guard(_mtx);
+                auto itr = _conditions.find(id);
+                if(itr == _conditions.end())
+                {
+                    return;
+                }
+                p = itr->second;
+            }
+            std::unique_lock<std::mutex> guard(p->mutex);
+            p->cond.notify_all();
```
minggo: notify doesn't have to lock the mutex

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+using namespace std;
+
+
+namespace {
+
+    struct ConditionVariable {
+        std::mutex mutex;
+        std::condition_variable cond;
+    };
```
minggo, line 62: would it be better to implement `lock()` `notify()` in `ConditionVariable`?

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,384 @@
...
+            p->cond.wait_for(ul, timeout);
+        }
+    private:
+        std::mutex _mtx;
+        std::unordered_map<int64_t, std::shared_ptr<ConditionVariable>> _conditions;
+    };
+}
+
+
+namespace cocos2d{
```
minggo: i think anonymous namespace is better

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+
+
+    int64_t _callJavaConnect(const std::string &url,const std::vector<std::string> *protocals, const std::string & caFile)
```
minggo: Will this function be invoked not int gl thread? If so then, the jni functions may not be thread safe.

cocos/network/CCWebSocket-android.cpp
PatriceJiang: Yes

cocos/network/WebSocket.h
```diff
@@ -40,6 +40,10 @@
 #include "platform/CCPlatformMacros.h"
 #include "platform/CCStdC.h"
 
+#if CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID
+#include "platform/android/jni/JniHelper.h"
+#endif
+
```
minggo: Can remove platform specific codes here?

PatriceJiang: JNI types `jstring` `jlong` are referenced in friend function, these lines should be kept.

PatriceJiang:
@crazyhappygame Can I trouble you to review this PR if you are available?

crazyhappygame:
1. Could you provide why, what and how we would like to improve?
2. This code review is quite big. Could you split it to few smaller? For example I see that you did some refactoring move code, rename and did some functional changes. I propose to create one code review for refactoring (no functional changes) another PR for functional changes.

PatriceJiang:
1. I have updated the first comment to provide more info about this PR. https://github.com/cocos2d/cocos2d-x/pull/19766#issue-282442889
2. these commits have a dependency, and I can only exact commits before refactoring so that it can be a little easier to read. https://github.com/cocos2d/cocos2d-x/pull/19779 &  refactoring change page https://github.com/PatriceJiang/cocos2d-x/compare/a19198f8daf4c32157b7f2d5b48b12fba0a67b06...use-okhttp3

@crazyhappygame

crazyhappygame:
What about other platforms: win32, MacOS, Linux?

crazyhappygame:
What about following:
https://github.com/square/okhttp/blob/master/README.md

1. "OkHttp works on Android 5.0+ (API level 21+) and on Java 8+."

2. "The OkHttp 3.12.x branch supports Android 2.3+ (API level 9+) and Java 7+. These platforms lack support for TLS 1.2 and should not be used. But because upgrading is difficult we will backport critical fixes to the 3.12.x branch through December 31, 2020."

crazyhappygame:
It looks like this the recommended Android library for http:
https://developer.android.com/reference/java/net/HttpURLConnection.html

PatriceJiang:
> What about other platforms: win32, MacOS, Linux?

RocketSocket supports both iOS and MacOS.
About Linux I only have a rough idea, that is trying to access system CA store and keep libwebsocket.  I did not find a way for win32 yet. 

> 2. "The OkHttp 3.12.x branch supports Android 2.3+ (API level 9+) and Java 7+. These platforms lack support for TLS 1.2 and should not be used. But because upgrading is difficult we will backport critical fixes to the 3.12.x branch through December 31, 2020."

Should we add an option for developers to use libwebsocket, which has in-App `OpenSSL` library?

crazyhappygame:
1. We would like to use version "OkHttp 3.12.x " and on official page there is comment "should not be used"?
2. Do I understand it right that currently everywhere except of Android curl is used?

cocos/network/WebSocket-libwebsocket.h
```diff
@@ -0,0 +1,155 @@
...
+    EventListenerCustom* _resetDirectorListener;
+
+    friend class WsThreadHelper;
+    friend class WebSocketCallbackWrapper;
+
+    WebSocket *_socket = nullptr;
+
+
+    WebSocket::State       _readyState =  WebSocket::State ::CONNECTING;
+    WebSocket::Delegate*   _delegate  = nullptr;
```
JohnCoconut, line 141: `_socket`, `_readyState`, `_delegate` are initialized here and in member initializer list. The member initializer list has higher priority.

PatriceJiang:
1. To provide the best safety & compacity, `okhttp` seems does not 100 percent fit. I assumed that it has a limited impact on the security of games.
2. curl is used on win32 and Linux.  Mac, iOS and Android use platform specific implementation.

I think I should suspend this commit, and try to find a better replacement of `okhttp`. @crazyhappygame

crazyhappygame:
I think that we should use curl everywhere. If only Android does not use curl I think that it should be change that Android use curl.

This is very important for multiple platform libraries like cocos2d-x to use the same external libs. With that you have one code base. When you fix bug you fix it for all platforms.

minggo:
@crazyhappygame yep, have one code base is good, but we use different ways on Android and iOS just to reduce package size as package size is very important then. We should consider package size again now.

crazyhappygame:
@minggo thank you for explanation.
To check size I think that we should create app without http client and than:
1.https://developer.android.com/reference/java/net/HttpURLConnection.html 
2. Okhttp3
3. Curl
And check apk size.

PatriceJiang:
@crazyhappygame I did test in some different conditions, and the conclusion was that we can decrease the 1M of package size by using okhttp. 

Settings
- release mode
- use ABI `arm64-v8a`

#### empty project

`apk` sizes in different conditions

| conditions| branch v3|  this pr|
|---------------|--------------|----------|
|no ws|3,422,425 B|3,503,295 B|
|with ws|4,593,120 B|3,505,090 B|

#### cpp-tests
| branch v3|  this pr|
|--------------|----------|
|47,246,303 B|46,181,487 B|

crazyhappygame:
Ok. Thank you for providing data about size.

But I think we should cancel this PR because of the https://github.com/cocos2d/cocos2d-x/pull/19766#issuecomment-497105168

PR #19767, opened 2019-05-27
refactor CCScene

JohnCoconut:
* use member initializer list
* add const to member function

```cpp
, _event(Director::getInstance()->getEventDispatcher()->addCustomEventListener(Director::EVENT_PROJECTION_CHANGED, std::bind(&Scene::onProjectionChanged, this, std::placeholders::_1)))
```

C++ allows us to use member functions in member initializer list, right? I am not very sure of when I am allowed to do this.

### DIFF
--- a/cocos/2d/CCScene.cpp
+++ b/cocos/2d/CCScene.cpp
@@ -52,28 +52,16 @@ THE SOFTWARE.
 NS_CC_BEGIN
 
 Scene::Scene()
+: _defaultCamera(Camera::create())
+, _event(Director::getInstance()->getEventDispatcher()->addCustomEventListener(Director::EVENT_PROJECTION_CHANGED, std::bind(&Scene::onProjectionChanged, this, std::placeholders::_1)))
 {
-#if CC_USE_3D_PHYSICS && CC_ENABLE_BULLET_INTEGRATION
-    _physics3DWorld = nullptr;
-    _physics3dDebugCamera = nullptr;
-#endif
-#if CC_USE_NAVMESH
-    _navMesh = nullptr;
-    _navMeshDebugCamera = nullptr;
-#endif
-#if CC_USE_PHYSICS
-    _physicsWorld = nullptr;
-#endif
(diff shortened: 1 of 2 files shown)


minggo:
Thanks, but i think initialize it in declaration is better.

JohnCoconut:
Do you mean keeping all of the member variables in the declaration, or just the `_event` variable?

minggo:
If it is suitable in declaration, for example
```c++
class Test
{
int a = 0;
};
```

JohnCoconut:
Didn't initialize `_defaultCamera` and `_event` in-class, as they are used as incomplete type. Doing so would require us to include more headers into `CCScene.h`.

minggo:
Yep, but initialize them to `null` in declaration is a good behavior.

JohnCoconut:
> Yep, but initialize them to null in declaration is a good behaviour.

May I ask why?

I definitely prefer in-class initialization of a variable over initializer list, like the case below.

```cpp
// in Test.h
class Test {
public:
  Test();
   ....
private:
    T *t_ = nullptr;
};
```
Reading `Test.h` header file gives me the impression that 
1. `t_` is defaulted to `nullptr`
2. we can ignore it in constructor definition in cpp file.

This is all good.

But I find it confusing if `T* t_` will be initialized to another value in constructor, and we put a `nullptr` in header file, like the following,

```cpp
// in Test.h
class Test {
public:
  Test();
   ....
private:
    T *t_ = nullptr;
};
```
We are going to initialize it to `nullptr`?
```cpp
// in Test.cpp
Test() : t_(TFactory::getInstance()) {}
```

No we aren't.

minggo:
Initialize variables in declaration will avoid missing being initialized. Yep, in this case we know these variables will be initialized in constructor, but it is easy to forget doing it in constructor.

JohnCoconut:
@minggo initialize pointer to `nullptr` in class as requested.

minggo:
thanks

minggo (approved):

PR #19779, opened 2019-05-29
[android] Use okhttp3 to implement Downloader & WebSocket - p1

PatriceJiang:
Code without refactor

original PR https://github.com/cocos2d/cocos2d-x/pull/19766


cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+namespace {
+
+    using namespace cocos2d;
+
+    int64_t _callJavaConnect(const std::string &url, const std::vector<std::string> *protocals,
+                             const std::string &caFile) {
+        jlong connectionID = -1;
+        JniMethodInfo methodInfo;
+        if (JniHelper::getStaticMethodInfo(methodInfo,
+                                           J_BINARY_CLS_WEBSOCKET,
```
crazyhappygame, line 119: Please use (here and in other places)
cocos2d::JniHelper::callStaticVoidMethod
To simplify call and automate resource cleanup

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+#define JARG_STR        "Ljava/lang/String;"
+#define JARG_DOWNLOADER "L" JCLS_WEBSOCKET ";"
+
+#include <string>
+#include <mutex>
+#include <unordered_map>
+#include <condition_variable>
+#include <chrono>
+
+using namespace std;
```
crazyhappygame: Please remove this line.
https://clang.llvm.org/extra/clang-tidy/checks/google-build-using-namespace.html

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+        }
+
+    private:
+        std::mutex _mutex;
+        std::condition_variable _cond;
+    };
+
+    class ConnectionCV {
+    public:
+        ConnectionCV() = default;
```
crazyhappygame: Please remove this line

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+        virtual ~ConnectionCV() {
```
crazyhappygame: Please remove this destructor.
1. std::mutex::lock may throw
https://en.cppreference.com/w/cpp/thread/mutex/lock
2. You do not need destructor to clear  this variable

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+        void unwatchConnection(int64_t id)
+        {
+            std::lock_guard<std::mutex> guard(_mtx);
+            _conditions.erase(id);
+        }
+
+        void notifyClosed(int64_t id)
+        {
+            std::shared_ptr<ConditionVariable> p = nullptr;
```
crazyhappygame: please remove "= nullptr".
std::shared_ptr default constructor will do this

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+                    return;
+                }
+                p = itr->second;
+            }
+            p->notify_all();
+        }
+
+        void waitForClosing(int64_t id, std::chrono::duration<float> timeout)
+        {
+            std::shared_ptr<ConditionVariable> p = nullptr;
```
crazyhappygame: Please remove "= nullptr"

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+                if (itr == _conditions.end()) {
+                    return;
+                }
+                p = itr->second;
+            }
+            p->wait_for(timeout);
+        }
+    private:
+        std::mutex _mtx;
+        std::unordered_map<int64_t, std::shared_ptr<ConditionVariable>> _conditions;
```
crazyhappygame, line 111: Why std::shared_ptr ? What and how it is shared here?

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+    };
+}
+
+
+namespace {
```
crazyhappygame: Why close and open unnamed namespace

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+    using namespace cocos2d;
```
crazyhappygame: Please remove using

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+
+    namespace network {
+
+        void _WebSocketAndroidNativeTriggerEvent(JNIEnv *env, jclass *klass, jlong cid,
+                                                 jstring eventName, jstring data,
+                                                 jboolean isBinary) {
+            std::lock_guard<std::recursive_mutex> guard(__socketMapMtx);
+            auto itr = __socketMap.find(cid);
+            if (itr != __socketMap.end()) {
+                const char *ceventName = env->GetStringUTFChars(eventName, nullptr);
```
crazyhappygame: Please use cocos2d::StringUtils::getStringUTFCharsJNI for string conversion and resource management

cocos/network/CCWebSocket-android.cpp
```diff
@@ -0,0 +1,403 @@
...
+            CCLOGERROR("_registerNativeMethods: failed");
+            if (env->ExceptionCheck()) {
+                env->ExceptionClear();
+            }
+            return false;
+        }
+        return true;
+    }
+
+    static bool __registered = false;
```
crazyhappygame: 1. Please avoid to use static variable
2. Variable can not start from __

17.4.3.1.2 Global names [lib.global.names]
Certain sets of names and function signatures are always reserved to the implementation:

Each name that contains a double underscore (__) or begins with an underscore followed by an uppercase letter (2.11) is reserved to the implementation for any use.
Each name that begins with an underscore is reserved to the implementation for use as a name in the global namespace.165
165) Such names are also reserved in namespace ::std (17.4.3.1).

crazyhappygame (changes requested):

cocos/network/CCWebSocket-android.cpp
PatriceJiang: The methods of generating Java function signature and type conversions are incomplete. Only basic types are supported. Should we add more converting functions here?

cocos/network/CCWebSocket-android.cpp
PatriceJiang: `shared_ptr` is used to prevent a local reference from being deleted. 

In `waitForClosing`
```cpp
             {
               std::lock_guard<std::mutex> guard(_mtx);
                auto itr = _conditions.find(id);
                if (itr == _conditions.end()) {
                    return;
                }
                p = itr->second;
            }
            // prevent p from being delete here
            p->wait_for(timeout);
```

Issue #19781, opened 2019-05-30
Allow RenderTexture to save out non-premultiplied alpha images


rh101:
- cocos2d-x version: 3.17.1
- devices test on: Win32
- developing environments
   - NDK version: N/A
   - Xcode version: N/A
   - VS version: VS2019

Steps to Reproduce:

The current implementation of RenderTexture has several issues.

Firstly, it does not set the premultiplied alpha flag on the image created by this method:

```
Image* RenderTexture::newImage(bool flipImage)
{
...
            image->initWithRawData(buffer, savedBufferWidth * savedBufferHeight * 4, savedBufferWidth, savedBufferHeight, 8);
        }
        else
        {
            image->initWithRawData(tempData, savedBufferWidth * savedBufferHeight * 4, savedBufferWidth, savedBufferHeight, 8);
        }
...
}
```

The correct code would be to change the initWithRawData calls to this:

```
            image->initWithRawData(buffer, savedBufferWidth * savedBufferHeight * 4, savedBufferWidth, savedBufferHeight, 8, true);
        }
        else
        {
            image->initWithRawData(tempData, savedBufferWidth * savedBufferHeight * 4, savedBufferWidth, savedBufferHeight, 8, true);
```

Secondly, there is no way to choose if you want to save the file as a non-premultiplied alpha PNG.

Reversing the PMA before saving the image is possible, and a new method can be added to do this, such as: RenderTexture::saveToFileAsNonPMA().

I'll create a pull request with the changes, along with modification of the tests to cater for this new method.

This fix may work for the following reported issues:
https://github.com/cocos2d/cocos2d-x/issues/18980
https://github.com/cocos2d/cocos2d-x/issues/18193

PR #19782, opened 2019-05-30
Added support for saving images with non-premultiplied alpha

rh101:
Refer to #19781


cocos/platform/ios/CCImage-ios.mm
```diff
@@ -91,7 +91,14 @@ of this software and associated documentation files (the "Software"), to deal
...
     {
-        bitmapInfo |= kCGImageAlphaPremultipliedLast;
+		if (_hasPremultipliedAlpha)
+		{
+			bitmapInfo |= kCGImageAlphaPremultipliedLast;
+		}
+		else
+		{
+			bitmapInfo |= kCGImageAlphaLast;
+		}
```
minggo: could you please fix the indent

rh101: Sorry I didn't notice the .mm files not having correct formatting in my editor. Converted the tabs to spaces, so the indentation is fixed now.

minggo (approved):

PR #19783, opened 2019-05-30
more clang-tidy perf options

JohnCoconut:
* performance-move-constructor-init
* performance-unnecessary-copy-initialization

Disable warnings on `RefPtr` refcount test case.

Reference: 

1. https://releases.llvm.org/7.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-move-constructor-init.html

2. https://releases.llvm.org/7.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-unnecessary-copy-initialization.html


minggo:
Why disable warning? I think it is better to fix the warning.

JohnCoconut:
@minggo That RefPtr test case is to test copy constructor. We aren't supposed to replace copy with const reference in this test case. Otherwise the results would be wrong and fail to pass test.

minggo:
Is it because `ref4` is not used in release mode? Or why copy constructor is not needed? In some place it is needed.

JohnCoconut:
@minggo 

> Is it because ref4 is not used in release mode? 

No. **That's a test case. We want to test whether copy constructor behaves correctly.** 

`RefPtr` has a copy constructor and also a reference count. If `RefPtr a` has a reference count of 1. Making a copy of `a` will increase reference count to 2. 

```cpp
RefPtr<__String> s1;   // ref count = 1
RefPtr<__String> s2(s1); // ref count = 2
// const RefPtr<__String>& s2(s1); // This won't change ref count at all.
assert(s1->getReferenceCount == 2);
```

Changing the copy to const reference will totally avoid this test.

JohnCoconut:
This piece of code is in a unit test file. It shouldn't be optimized
because we want to observe the effect of copy constructor.

minggo:
@JohnCoconut yep, we don't want to optimize test codes. I just wonder what's the problem of the codes.

PR #19785, opened 2019-06-03
vertex shader does not need header define

Mee-gu:
vertex shader does not need header definition. For fragment shader, `highp` may not be defined on some systems.


Mee-gu:
According to [4.5.3 Default Precision Qualifiers](https://www.khronos.org/registry/OpenGL/specs/es/2.0/GLSL_ES_Specification_1.00.pdf):
> The vertex language has the following predeclared globally scoped default precision statements:
> precision highp float;
> precision highp int;

 **4.5.4 Available Precision Qualifiers:** The built-in macro `GL_FRAGMENT_PRECISION_HIGH` is defined to one on systems supporting highp precision in the fragment language

PR #19786, opened 2019-06-03
bind texture target

Mee-gu:
bind texture target


Mee-gu:
issue ref from [here](https://discuss.cocos2d-x.org/t/metal-support-alpha1-released/46282/47)

PR #19790, opened 2019-06-04
[Android]Override removeDirectory function in FileUtilsAndroid, fix "rm not found" error on OPPO ColorOS

wanghehacker:
Override removeDirectory function in FileUtilsAndroid, use JRE method instead of "rm -r path". Because We found that on OPPO's ColorOS, the environment variable PATH may not always correct, it could be ":/sbin" sometimes, then cause an error "rm not found", afterwards the directory could not be deleted.

### DIFF
--- a/cocos/platform/android/CCFileUtils-android.cpp
+++ b/cocos/platform/android/CCFileUtils-android.cpp
@@ -344,6 +344,15 @@ std::vector<std::string> FileUtilsAndroid::listFiles(const std::string& dirPath)
     return fileList;
 }
 
+bool FileUtilsAndroid::removeDirectory(const std::string& path) const
+{
+    if (path.empty())
+        return false;
+    
+    bool suc = removeDirectoryJNI(path.c_str());
+    return suc;
+}
+
 FileUtils::Status FileUtilsAndroid::getContents(const std::string& filename, ResizableBuffer* buffer) const
 {
     static const std::string apkprefix("assets/");
--- a/cocos/platform/android/CCFileUtils-android.cpp
+++ b/cocos/platform/android/CCFileUtils-android.cpp
@@ -349,8 +349,7 @@ bool FileUtilsAndroid::removeDirectory(const std::string& path) const
     if (path.empty())
         return false;
     
-    bool suc = removeDirectoryJNI(path.c_str());
-    return suc;
+    return removeDirectoryJNI(path.c_str());
 }
 
 FileUtils::Status FileUtilsAndroid::getContents(const std::string& filename, ResizableBuffer* buffer) const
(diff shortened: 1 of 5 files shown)


cocos/platform/android/CCFileUtils-android.cpp
```diff
@@ -344,6 +344,15 @@ std::vector<std::string> FileUtilsAndroid::listFiles(const std::string& dirPath)
     return fileList;
 }
 
+bool FileUtilsAndroid::removeDirectory(const std::string& path) const
+{
+    if (path.empty())
+        return false;
+    
+    bool suc = removeDirectoryJNI(path.c_str());
+    return suc;
```
PatriceJiang: I suggest this variable can be omitted

PatriceJiang (approved):

cocos/platform/android/CCFileUtils-android.cpp
minggo: just `return removeDirectoryJNI(path.c_str())`

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
```diff
@@ -207,6 +207,39 @@ public boolean accept(File dir, String name) {
         
         return Cocos2dxHelper.sAssetsPath;
     }
+
+    //remove directory 
+    public static boolean RemoveDirectory(final String directory)
```
minggo: `RemoveDirectory -> removeDirectory` to obey the coding style

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
```diff
@@ -207,6 +207,39 @@ public boolean accept(File dir, String name) {
...
+
+
+    private static void RecursionDeleteFile(File file)
+    {
+        if (file.isFile()) {
+            file.delete();
+            return;
+        }
+        if (file.isDirectory()) {
+            File[] childFile = file.listFiles();
```
minggo: childFile -> childrenFile ?

cocos/platform/android/CCFileUtils-android.cpp
wanghehacker: ok, I will modify it

minggo (approved):

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
wanghehacker: yeah ~ I will modify it

PR #19797, opened 2019-06-05
Feature/fix directly use open gl codes

Mee-gu:
remove opengl codes

### DIFF
--- a/cocos/3d/CCBundle3DData.h
+++ b/cocos/3d/CCBundle3DData.h
@@ -324,8 +324,8 @@ struct NTextureData
      std::string id;
      std::string filename;
      Usage type;
-     GLenum wrapS;
-     GLenum wrapT;
+     backend::SamplerAddressMode wrapS;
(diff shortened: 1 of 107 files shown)


cocos/3d/CCBundle3DData.h
```diff
@@ -324,8 +324,8 @@ struct NTextureData
      std::string id;
      std::string filename;
      Usage type;
-     GLenum wrapS;
-     GLenum wrapT;
```
minggo, line 328: Should be care about it, because the raw value type is not `backend::SamplerAddressMode`. Any code convert it from GLenum to `backend::SamplerAddressMode`?

Mee-gu: the original address mode value was stored as a string in files, the `parseSamplerAddressMode` will convert string value to `backend::SamplerAddressMode`. So it is fine to do so.

minggo (approved):

PR #19800, opened 2019-06-05
Remove android:allowBackup

crazyhappygame:
This option should be set only on application level.
Having it on library level it is impossible to set android:allowBackup="false" in application


crazyhappygame:
@minggo could you merge it?

minggo:
Sure, but please don't update submodule, is it something about submodule?

crazyhappygame:
Thank you. I will remove submodule change.

PR #19811, opened 2019-06-06
android_allowBackup2

crazyhappygame:
This option should be set only on application level.
Having it on library level it is impossible to set android:allowBackup="false" in application


crazyhappygame:
@minggo could you merge it?

PR #19822, opened 2019-06-11
refactor AudioEngine-Linux

JohnCoconut:
* use std::map::insert properly

```cpp
std::pair<iterator,bool> insert( value_type&& value );
```
From: https://en.cppreference.com/w/cpp/container/map/insert

> Inserts element(s) into the container, if the container doesn't already contain an element with an equivalent key. 

>  Returns a pair consisting of an iterator to the inserted element (or to the element that prevented the insertion) and a bool denoting whether the insertion took place.

```cpp
if (mapId.find(filePath) == mapId.end())
    mapId.insert({filePath, id});
```
This code is bad for performance. It will iterate the `mapId` twice. One in `mapId.find`, and the other in `mapId.insert`.

The new code only iterate the map once.

The performance might not matter here, but this pattern is not rare in the code. Unfortunately there is no tool to find this.

### DIFF
--- a/cocos/audio/linux/AudioEngine-linux.cpp
+++ b/cocos/audio/linux/AudioEngine-linux.cpp
@@ -72,8 +72,6 @@ AudioEngineImpl::AudioEngineImpl()
 AudioEngineImpl::~AudioEngineImpl()
 {
     FMOD_RESULT result;
-    result = pSystem->close();
-    ERRCHECKWITHEXIT(result);
     result = pSystem->release();
     ERRCHECKWITHEXIT(result);
 }
@@ -109,7 +107,7 @@ int AudioEngineImpl::play2d(const std::string &fileFullPath, bool loop, float vo
     int id = preload(fileFullPath, nullptr);
     if (id >= 0) {
         mapChannelInfo[id].loop=loop;
-        mapChannelInfo[id].channel->setPaused(true);
+        //mapChannelInfo[id].channel->setPaused(true);
         mapChannelInfo[id].volume = volume;
         AudioEngine::_audioIDInfoMap[id].state = AudioEngine::AudioState::PAUSED;
         resume(id);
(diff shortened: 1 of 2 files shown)


minggo (approved):

minggo (approved):

JohnCoconut:
@minggo I am going through `AudioEngin` code and plan to do a few minor refactoring. Could you keep the PR open and merge it when I request it in future?

minggo:
Sure.

JohnCoconut:
rebase the PR branch with a few commits.

    refactor AudioEngine and AudioEngine-linux
    
    * map::erase() can handle case if key doesn't exist.
    
    * use map::iterator when it has already obtained.
    
    * mapChannelInfo[id].channel is nullptr befor resume(). Don't
    dereference it.
    
    * FMOD::System::release() calls close, so calling close before release
    is not necessary.
    
    * use std::map::insert properly.

JohnCoconut:
remove unnecessary null check on _audioEngineImpl, as it has already been checked at the beginning of the function body.

v1993:
Won't it be better for performance to use `std::unordered_map`? It's faster in general and shouldn't increase memory usage too much.

JohnCoconut:
@v1993 
> Won't it be better for performance to use std::unordered_map?

Agree. `unordered_map` performs better than `map` when it has large number of items. 

`AudioEngine-linux.cpp` doesn't seem to make any use of ordering provided by `std::map`, so it can be replaced.

Currently I am identify the cause of cpp-tests failure. I want to keep minimize the change in this file. Maybe change `std::map` to `std::unordered_map` after I find and fix the bug?

cocos/audio/linux/AudioEngine-linux.cpp
```diff
@@ -109,7 +107,7 @@ int AudioEngineImpl::play2d(const std::string &fileFullPath, bool loop, float vo
     int id = preload(fileFullPath, nullptr);
     if (id >= 0) {
         mapChannelInfo[id].loop=loop;
-        mapChannelInfo[id].channel->setPaused(true);
+        //mapChannelInfo[id].channel->setPaused(true);
```
minggo, line 111: please add a comment here

minggo:
@JohnCoconut please let me know when  this PR is ready to be merged. And please don't make a very big PR. Thanks.

JohnCoconut:
It can be merged now. Thanks.

minggo:
Got it, thanks.

PR #19827, opened 2019-06-13
Feature/lua render texture test

Mee-gu:
fix Lua render texture test


minggo:
Don't have to commit binding codes that are automatically generated.

Issue #19707, opened 2019-05-14
CMake should just copy .plist file from template


minggo:
.plist file is a application related resource, CMake should just copy it from template.

PR #19828, opened 2019-06-13
[cmake] Update plist template for iOS projects

PatriceJiang:
Use different `Info.plist` template in each project, rather than using the same ` cmake/Modules/iOSBundleInfo.plist.in`


minggo (approved):

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19707

Issue #19836, opened 2019-06-16
VS2019 release mode compile issue


rh101:
- cocos2d-x version: 3.17.2
- devices test on: Windows 10 PC
- developing environments
   - VS version: VS2019

Steps to Reproduce:

1. Compile a release mode build for Win32
2. You will notice the compiler will be stuck on LocalStorage.cpp or CCController-linux-win32.cpp since optimizations are turned on.  The actual problem is in CCController-linux-win32.cpp, and not LocalStorage.cpp.

More info [here](https://discuss.cocos2d-x.org/t/visual-studio-2019/46250/15)

It’s only one of the optimization options that is actually causing this issue, which is global optimization (/Og, which is automatically turned on when /O1 or /O2 are enabled), so it's best to deal with that section locally in order to keep optimizations on for the rest of the cocos2d-x library.

So, in order to get the CCController-linux-win32.cpp code compiling properly, we don't need to switch off all optimizations, but just the one that's causing the problem:

```
#pragma optimize( "g", off ) // Disable global optimizations.
class CC_DLL ControllerImpl
{
	public:
...
...
};
#pragma optimize( "", on ) // resets the optimizations to those that you specified with the /O compiler option
```

Issue #19610, opened 2019-04-16
Update Glfw to 3.3


Bilalmirza:
Glfw 3.3 released today and it adds a lot of features and fixes. 
@minggo  Will it be possible to include this in the next version of cocos2d-x?

minggo:
Sure.

minggo:
I tried to update to 3.3, but met the error on mac, then could not show the application content
```
2019-04-28 17:34:02.354991+0800 cpp-empty-test Mac[39829:1293351] Unknown class 'AppController', using 'NSObject' instead. Encountered in Interface Builder file at path /Users/minggo/Library/Developer/Xcode/DerivedData/cocos2d_tests-fxkupjwuhtxygugcbsprxsripwej/Build/Products/Debug/cpp-empty-test Mac.app/Contents/Resources/en.lproj/MainMenu.nib.
2019-04-28 17:34:02.355237+0800 cpp-empty-test Mac[39829:1293351] Could not connect action, target class NSObject does not respond to -toggleFullScreen:
```

I have not idea with it. My commit is [here](https://github.com/minggo/cocos2d-x-3rd-party-libs-src/commit/963607d21845b2eaacbecdb9656e5303a2a37082).

Bilalmirza:
Sorry I didn't get a chance to use your commit.
I built glfw using cmake and replaced with the current and everything worked fine for me.

minggo:
It is wired. What's your environment?

Bilalmirza:
tested this on macOS only

rh101:
@minggo I'll try to do a test on my Mac Mini to see what happens within the next few days, and I'll post back here.  On Windows (Win32) the GLFW 3.3 worked fine with Cocos2d-x.

rh101:
@minggo I just built the template game code on Mac OS X with GLFW 3.3 successfully without issues (except some Spine SkeletonRenderer ambiguous type, just changing Rect => cocos2d::Rect fixed it).  I had to build the GLFW3.3 library, libglfw3.a, first, since one wasn't available to download.  This is using macOS Mojave 10.14.4, and Xcode 10.2.1.

The only other step is to make sure the updated GLFW3.3 headers, glfw3.h and glfw3native.h, are copied into all cocos2d/external/glfw3/include/linux, /mac and /win32 folders.

minggo:
@rh101 how did you compile GLFW 3.3?

rh101:
@minggo I did the following:

[...]

PR #19837, opened 2019-06-16
Fix for win32 release mode build in VS2019

rh101:
This is to address the issue described in #19836


crazyhappygame:
@rh101 could you explain what and how it is fixed?
If this is VS compiler error could you provide link to compiler error?

rh101:
The explanation is in the issue I linked: https://github.com/cocos2d/cocos2d-x/issues/19836

The compiler freezes when it is trying to optimize a certain section of code.  All that this fix does is disable global optimization ("g" option in the as described here: https://docs.microsoft.com/en-us/cpp/preprocessor/optimize?view=vs-2019) for the ControllerImpl class in CCControllers-linux-win32.cpp.  Other optimizations stay on (like speed optimization etc).  The global optimization is turned back on at the end of that class declaration.

At the moment /O2 flag is used for release mode optimization, and as detailed [here](https://docs.microsoft.com/en-us/cpp/build/reference/o1-o2-minimize-size-maximize-speed?view=vs-2019), it is equivalent to:

```
/O2 (Maximize Speed) turns on /Og /Oi /Ot /Oy /Ob2 /GF /Gy
```
The /Og ([global optimization](https://docs.microsoft.com/en-us/cpp/build/reference/og-global-optimizations?view=vs-2019)) is the one that is causing an issue here for the ControllerImpl class.

The compiler freezing is the same as what this post describes:
https://developercommunity.visualstudio.com/content/problem/525689/vs2019-x64-x86-cross-compiler-freeze-compiling-som.html

Quote from that site: "This is an issue with a newer optimization that can have a O(n^2) behavior in certain cases which will be fixed in a future update, most likely VS 2019 update 2."

crazyhappygame:
Thank you for explanation.

I think that we should refactor function with more than 4500 lines of code. It is bad habit to have such long functions...

One my idea is to create const table. And than loop instead of repeating:
```
deviceName = "device name";
buttonInputMap.clear();
axisInputMap.clear();
//Init
// Add the controller profile to the map
s_controllerProfiles.insert(std::make_pair(deviceName, std::make_pair(buttonInputMap, axisInputMap)));
```

rh101:
I think your idea is good, since yes, it's an excessively long function.  Refactoring it may actually fix the optimization issue, so there would be no need to apply this work-around I've submitted.

Also, I noticed this comment in the code:
```
//FIXME: Once GLFW 3.3 is bundled with cocos2d-x and no extra
//controllers have been added here, remove all the code here.
```

Now there is another issue logged #19610 which is exactly about updating to GLFW 3.3, so if that is done, then I guess the code in that function can be removed?

crazyhappygame:
Could you close this PR and refactor more than 4500 long function in new PR?

rh101:
Unfortunately, I can't simply take time away from working on my own projects at the moment, so I'm not sure I'll be able to dedicate time to figuring out what this 4500 function does and how best to refactor it.   

Also, I'm not sure that closing this PR is a good idea, since it's a valid fix that affects no other part of the code, and perhaps it's best to leave it open for now in case no-one gets around to refactoring that code.  In addition to that, refactoring the code is not a guarantee that the compile issue will be fixed.  

Also, given this is somehow related to the use of GLFW 3.3, is Cocos2d-x going to be updated to use that version?

crazyhappygame:
OK. I see your point. But please do not commit this. I have a filling that soon VS 2019 will be fixed and than we will have unneed fix for VS 2019.1 in cocos2d-x code.

If you do not have time I will try to this refactoring in the meantime ...

rh101:
I'm going to close this pull request because I need to fix some issues with my github repo. I'll recreate it later if required.

PR #19845, opened 2019-06-20
optimize the uniform buffer

Mee-gu:
optimize the uniform buffer


cocos/renderer/backend/ProgramState.cpp
```diff
@@ -262,22 +264,44 @@ ProgramState *ProgramState::clone() const
...
+        {
             _fragmentUniformInfos[uniformInfo.second.location] = uniformInfo.second;
+            totalUniformsSize += uniformInfo.second.bufferSize;
+        }
+    }
+#ifdef CC_USE_METAL
+    if (totalUniformsSize > 0)
+    {
+        _fragmentUniformBuffer.resize(totalUniformsSize);
+        std::fill(_fragmentUniformBuffer.begin(), _fragmentUniformBuffer.end(), 0);
```
minggo: Can just use _fragmentUniformBuffer.resize(totalUniformSize, 0). But i think the default value is 0 for char, so it is not need to initialize it.

PR #19849, opened 2019-06-20
speed up sortEventListenersOfSceneGraphPriority by reverse visit

huangwei1024:
speed up sortEventListenersOfSceneGraphPriority by reverse visit
its more efficient when node tree was big


cocos/base/CCEventDispatcher.cpp
```diff
@@ -1313,20 +1390,40 @@ void EventDispatcher::sortEventListeners(const EventListener::ListenerID& listen
...
+	_nodePriorityIndex = 0;
+	_nodePriorityMap.clear();
+
+	// reverse visit, from leaf to root, reduce the time of tree's travel
+	// use node userdata to mark, speed up hash find
+	for (const auto l : *sceneGraphListeners)
+	{
+		Node* node = l->getAssociatedNode();
+		if (node)
+			node->setUserData((void*)1);
```
crazyhappygame, line 1427: You can not setUserData.
It can be used by user.

cocos/base/CCEventDispatcher.cpp
```diff
@@ -222,6 +222,83 @@ EventDispatcher::~EventDispatcher()
     removeAllEventListeners();
 }
 
+
+void EventDispatcher::visitTree(Node* node)
```
crazyhappygame, line 226: Could you explain new algorithm and provide profile results. What is the speedup.

crazyhappygame (changes requested):

cocos/base/CCEventDispatcher.cpp
huangwei1024: yes, I known its not a right way which use UserData to storage some visit information for engine. and its simple to add an "SystemData" in Node for optimize

cocos/base/CCEventDispatcher.cpp
huangwei1024:
```
for (; i < childrenCount; i++)
{
	child = children.at(i);
	if (child && child->getLocalZOrder() < 0)
	{
		if (child->getUserData() == 0) // here!!!
			continue;
		visitTree(child);
	}
	else
		break;
}
```

I think its easy idea, if the tree had N nodes, and the sceneGraphListeners had M listeners, you could visit from M listener node to root, and color the path.
second, visit from root to child by the colored path. it could be avoid to visit useless path.

in my test proj, I had tree which contain 6K nodes, but listeners only about 600. I thought the case which listnener much less than nodes was in general.  O(N) -> O(c*M), c is the coefficient about the avg children and node in path. but in the worst c*M = N

huangwei1024: sorry, I could not provide the test case, it contain too many complex widgets.
but I test in win32, the old method cost 8ms, and cost 3ms when use the new one.

huangwei1024:
```
		if ((nodeData & 1) == 1)
			_globalZOrderNodeMap[node->getGlobalZOrder()].push_back(node);
```
```
if (_nodeListenersMap.find(node) != _nodeListenersMap.end())
        {
            _globalZOrderNodeMap[node->getGlobalZOrder()].push_back(node);
        }
```

and stl unordered_map also more expensive than bit mask.
when the N nodes is big, the cost need be consider

PR #19851, opened 2019-06-21
Feature/cache gl state

Mee-gu:
cache GL state

### DIFF
--- a/cocos/renderer/backend/opengl/GLStateCached.cpp
+++ b/cocos/renderer/backend/opengl/GLStateCached.cpp
@@ -1,0 +1,346 @@
+#include "GLStateCached.h"
+CC_BACKEND_BEGIN
+
+#define GL_STATE_CACHED_ENABLED
+#ifdef GL_STATE_CACHED_ENABLED
+#define RETURN() return;
+#else
+#define RETURN()
+#endif
+
+GL::DefaultGLState GL::_defaultGLState;
+
+void GL::enable(GLenum cap)
+{
+    switch(cap)
+    {
+        case GL_DEPTH_TEST:
+            if (_defaultGLState.depthEnabled)
+                RETURN();
+            glEnable(GL_DEPTH_TEST);
+            _defaultGLState.depthEnabled = GL_TRUE;
+            break;
+        case GL_STENCIL_TEST:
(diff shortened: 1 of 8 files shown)


cocos/renderer/backend/opengl/GLStateCached.h
```diff
@@ -0,0 +1,91 @@
...
+        GLenum dstRGBBlendFactor = GL_ZERO;
+        GLenum dstALphaBlendFactor = GL_ZERO;
+
+        GLenum bufferTarget = GL_ARRAY_BUFFER;
+        GLuint bufferObject = GL_ZERO;
+        GLuint framebufferObject = GL_ZERO;
+
+        GLenum textureUnit = GL_TEXTURE0;
+        GLenum textureTarget = GL_TEXTURE_2D;
+        GLuint textureObject = GL_ZERO;
```
minggo, line 85: It is better to query the state at start. The driver may not obey the spec.

cocos/renderer/backend/opengl/GLStateCached.cpp
```diff
@@ -0,0 +1,386 @@
+#include "GLStateCached.h"
+CC_BACKEND_BEGIN
+
+#define GL_STATE_CACHED_ENABLED
+#ifdef GL_STATE_CACHED_ENABLED
+#define RETURN() return;
```
minggo, line 6: return; -> return

crazyhappygame:
Is that really needed?
What is the speedup?

Mee-gu:
> Is that really needed?
> What is the speedup?

@crazyhappygame  GLStateCached will record the last-set value for mostly graphics API state, it will eliminate redundant API calls(e.g. enabling a state when it's already enabled). It's unnecessary to set these state repeatedly, although some calls are quick.

crazyhappygame:
@Mee-gu thank you for explanation.
I got an idea.
I am curious how much this code improve performance.

minggo:
I am not sure how much performance is improved with this PR too. It seems it depends on the implementation of GL driver implementation.
 
@Mee-gu i think we should do test on some android devices to make confident. The test doesn't have to use the engine, just use pure GL codes.

Issue #19852, opened 2019-06-21
memory leak with Data::fastSet


JohnCoconut:
cocos2d-x version: current v3 branch 

https://github.com/cocos2d/cocos2d-x/blob/4afda3a9abec62fa111396c19ea30cc31c11ea85/cocos/base/CCData.cpp#L127-L133

`fastSet` makes the `Data` object managing a new memory area in `[bytes, bytes + size)`, but it doesn't releasing the old data it manages. Failing to release the old data causes memory leak, as happens below,

https://github.com/cocos2d/cocos2d-x/blob/4afda3a9abec62fa111396c19ea30cc31c11ea85/cocos/base/CCUserDefault.cpp#L324-L334

Can be tested with the following example using valgrind

```cpp
int main(int argc, char **argv)
{
     UserDefault *userDefault = UserDefault::getInstance();
     Data defaultValue;
     unsigned char *someData = (unsigned char*)malloc(100);
     defaultValue.fastSet(someData, 100);
     Data newData = userDefault->getDataForKey("int_data", defaultValue);
}
```

with the following `UserDefault.xml` file,

```xml
<?xml version="1.0" encoding="UTF-8"?>
<userDefaultRoot>
    <int_data>BQAAAAQAAAADAAAAAgAAAAEAAAAAAAAA</int_data>
    <float_data>AACgQAAAgEAAAEBAAAAAQAAAgD8AAAAA</float_data>
    <double_data>AAAAAAAAFEAAAAAAAAAQQAAAAAAAAAhAAAAAAAAAAEAAAAAAAADwPwAAAAAAAAAA</double_data>
</userDefaultRoot>
```

PR #19853, opened 2019-06-21
fix memory leak in CCUserDefault

JohnCoconut:
fastSet makes the Data object managing a new memory area in
[bytes, bytes + size), but it doesn't releasing the old data
it managed. Failure to release the old data causes memory leak.

The default constructed Data manages null memory, so calling
fastSet on it is fine.

Because `Data ret = defaultValue;` malloc new memory, we might
have better performance without it.

This fix issue: #19852

### DIFF
--- a/cocos/base/CCUserDefault.cpp
+++ b/cocos/base/CCUserDefault.cpp
@@ -321,7 +321,7 @@ Data UserDefault::getDataForKey(const char* pKey, const Data& defaultValue)
         encodedData = (const char*)(node->FirstChild()->Value());
     }
     
-    Data ret = defaultValue;
+    Data ret;
     
     if (encodedData)
     {
@@ -331,6 +331,10 @@ Data UserDefault::getDataForKey(const char* pKey, const Data& defaultValue)
         if (decodedData) {
             ret.fastSet(decodedData, decodedDataLen);
         }
+    }
+    else
+    {
+        ret = defaultValue;
     }
     
     delete doc;


cocos/base/CCUserDefault.cpp
```diff
@@ -332,6 +332,10 @@ Data UserDefault::getDataForKey(const char* pKey, const Data& defaultValue)
             ret.fastSet(decodedData, decodedDataLen);
```
crazyhappygame, line 332: Could you check other places where fastSet is used?
I think that this method should be removed.

JohnCoconut: `CCData` has an unintuitive interface and incomplete documentation. I hope it behaves like smart pointers, since it manages memory ownership.

`fastSet` could be like `unique_ptr::reset(p)` by freeing previously managed memory, and obtaining ownership of `p`.

`copy` method has somewhat misleading documentation.

https://github.com/cocos2d/cocos2d-x/blob/4afda3a9abec62fa111396c19ea30cc31c11ea85/cocos/base/CCData.h#L96-L102

> Developer should free the pointer after invoking this method. 

Which pointer does it refers to in `free the pointer`? If it means the `buffer pointer`, we could call `reset` instead?

JohnCoconut:
`fastSet` is called in these locations,

```cpp
cocos/base/CCUserDefault-android.cpp
364:                ret.fastSet(decodedData, decodedDataLen);
402:        ret.fastSet(decodedData, decodedDataLen);

cocos/base/CCData.cpp
127:void Data::fastSet(unsigned char* bytes, const ssize_t size)
129:    CCASSERT(size >= 0, "fastSet size should be non-negative");
147:    fastSet(nullptr, 0);

cocos/base/CCData.h
99:     *  @see Data::fastSet
111:    void fastSet(unsigned char* bytes, const ssize_t size);

cocos/base/CCUserDefault-apple.mm
391:                ret.fastSet(decodedData, decodedDataLen);

cocos/base/CCUserDefault-winrt.cpp
187:            ret.fastSet(decodedData, decodedDataLen);

cocos/base/CCUserDefault.cpp
332:            ret.fastSet(decodedData, decodedDataLen);

cocos/platform/CCFileUtils.h
113:                _buffer->fastSet((unsigned char*)buffer, size);

cocos/platform/CCFileUtils.cpp
567:    data.fastSet((unsigned char*)dataStr.c_str(), dataStr.size());
571:    data.fastSet(nullptr, 0);

cocos/platform/android/CCDevice-android.cpp
168:        ret.fastSet(dc._data,width * height * 4);

cocos/platform/ios/CCDevice-ios.mm
592:        ret.fastSet(info.data,width * height * 4);

cocos/platform/mac/CCDevice-mac.mm
356:        ret.fastSet(info.data,width * height * 4);

cocos/platform/linux/CCDevice-linux.cpp
507:        ret.fastSet(dc._data,width * height * 4);

cocos/platform/win32/CCDevice-win32.cpp
505:        ret.fastSet(dataBuf, dataLen);
```

I suspect there is another memory leak in `cocos/platform/CCFileUtils.h:113`. I will take a look at it later. **Edit**: looking carefully, I found no mem leak in it.

minggo:
Yep, fastSet + copy is easy to trigger memory leak. I agree with @crazyhappygame , we should remove it if possible.

PR #19855, opened 2019-06-22
Support interface only libraries

rh101:
Added support for interface only libraries.  If they are linked via target_link_libraries to the application, then an error occurs in CocosBuildHelpers.cmake when certain functions try to access various properties of the interface only library, which are not allowed.


minggo:
@PatriceJiang could you please take a look? Thanks.

rh101:
If you want an example of an interface only library, then use this:

https://github.com/fmtlib/fmt

Add it to your game project as in this example:
```
# add fmt
set(fmt_INCLUDE_DIRS [your path]/fmt/include)
add_subdirectory([your path]/fmt ${PROJECT_BINARY_DIR}/fmt)

target_include_directories(${APP_NAME}
	PRIVATE Classes
	PRIVATE ${COCOS2DX_ROOT_PATH}/cocos/audio/include/
	PRIVATE ${fmt_INCLUDE_DIRS}
)

target_link_libraries(${APP_NAME} fmt-header-only)
```
You will see the errors right away, since the code in CocosBuildHelpers.cmake doesn't support it.  With the fix, it will work fine.

rh101:
Please don't forget about this PR.  It doesn't have any adverse side-effects, and allows the use of interface-only libraries.  Without the PR, CMake throws an error if an interface-only library is linked with the game, like the example in my previous post.

PatriceJiang (approved):

PR #19856, opened 2019-06-22
VS2019 win32 release mode compile fix

rh101:
Fix for the Win32 release mode compilation using VS2019, as described in another pull request #19837 and issue #19836 

The other pull request was closed due to a github branch issue.

This fix just turns off global optimizations around a specific class, but all other optimizations remain on. At the end of the class definition, it turns global optimizations back on.


cocos/base/CCController-linux-win32.cpp
```diff
@@ -37,6 +37,9 @@ THE SOFTWARE.
 
 NS_CC_BEGIN
 
+#if _MSC_VER >= 1920 && _MSC_VER <= 1921
+#pragma optimize( "g", off )
+#endif
```
minggo, line 42: It is the only VS version that cause the issue?

rh101:
It happens with all current VS2019 versions, the latest being v16.1.3, which also still has _MSC_VER = 1920.  Microsoft stated that they're looking into the optimization issue, but when they'll fix it is unknown.

It's only one optimization that has to be temporarily switched off for that class, [/Og](https://docs.microsoft.com/en-us/cpp/build/reference/og-global-optimizations?view=vs-2019), which is global optimization, but all other optimizations still work on that class.

We can wait for another release of VS2019 to see if MS fix the issue.  The only reason I put this PR In is because it was reported on the cocos forums by another user, [here](https://discuss.cocos2d-x.org/t/visual-studio-2019/46250), and it was easily reproduced.  

Also, that class has a comment regarding GLFW 3.3, and that all the code is no longer required when Cocos2d-x upgrades to GLFW 3.3, so wouldn't that be the better option?

minggo:
Yep, update to GLFW 3.3 is a better option. But i tried to update a few weeks ago and meet crash on mac and i don't know why. More detail information can refer to https://github.com/cocos2d/cocos2d-x/issues/19610.

minggo:
Close this PR since https://github.com/cocos2d/cocos2d-x/pull/19963 will update glfw to 3.3.

PR #19857, opened 2019-06-23
Alternative fix for missing android resources

rh101:
This is an alternative fix to the one in PR #19732 which copies the resources correctly.  Either one works.  The discussion regarding this is on the Cocos forums [here](https://discuss.cocos2d-x.org/t/android-cpp-tests-missing-resources/46776).

The difference between the two fixes is as follows:
My fix copies resources once during the build process, from Resources -> output.

The fix in PR 19732 copies them from Resources to app/assets, and then Gradle copies them from assets to the output folder.

@slackmoehrle requested to be informed when this was created.


templates/cpp-template-default/proj.android/app/build.gradle
```diff
@@ -111,22 +110,24 @@ android {
...
+    def projectRootFolder = "${projectDir}/../.."
+    def resourcesDir = "${projectRootFolder}/Resources"
+    def assetsOutputDir = mergeAssetsTask.outputDir.get()
+
+    mergeAssetsTask.doFirst {
+        // delete previous files first
+        delete "${assetsOutputDir}/*"
+    }
+
+    mergeAssetsTask.doLast {
```
minggo, line 123: As @PatriceJiang told me that, `doLast` may not be invoked.

rh101: I was not aware of that behavior of `doLast`, so it's best to close this PR, since the fix added by @PatriceJiang works fine as it is.

PR #19863, opened 2019-06-25
Feature/lua drawing node

Mee-gu:
fix DrawInRect in lua

### DIFF
--- a/cocos/scripting/lua-bindings/manual/cocos2d/lua_cocos2dx_manual.hpp
+++ b/cocos/scripting/lua-bindings/manual/cocos2d/lua_cocos2dx_manual.hpp
@@ -123,15 +123,15 @@ struct LuaEventMouseData
     }
 };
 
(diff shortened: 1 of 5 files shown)


cocos/scripting/lua-bindings/manual/cocos2d/lua_cocos2dx_manual.hpp
```diff
@@ -122,4 +123,15 @@ struct LuaEventMouseData
     }
 };
 
+class GLNode :public cocos2d::Node
```
minggo: `GLNode ` is not a good name.

minggo: How about using LuaRenderNode.

Mee-gu: How about LuaNode?

PR #19865, opened 2019-06-26
[bugfix] fix TextureInfo copy constructor

PatriceJiang:


cocos/renderer/backend/ProgramState.cpp
```diff
@@ -112,6 +112,9 @@ TextureInfo::TextureInfo(const std::vector<uint32_t>& _slots, const std::vector<
 TextureInfo::TextureInfo(const TextureInfo &other)
     : slot(other.slot)
     , textures(other.textures)
+#if CC_ENABLE_CACHE_TEXTURE_DATA == 1
```
minggo: `#if CC_ENABLE_CACHE_TEXTURE_DATA` is enough

PatriceJiang: yes

PatriceJiang:
```suggestion
#if CC_ENABLE_CACHE_TEXTURE_DATA
```

PR #19870, opened 2019-06-26
clang-tidy: performance-unnecessary-value-param

JohnCoconut:
The changes are made with clang-tidy option `performance-unnecessary-value-param`. However, this option makes questionable changes to the code.

For example, it changes `std::function<>` to `const std::function<>&`. `std::function` is almost always passed by value, from the code I've seen. And we're using clang-tidy 7, it doesn't offer a option to specifically allowing a type to be copied. We can enable this option, if the system supports clang-tidy 8 in the future.

So we just leave the option disabled, but keep some of the changes it made. The `-` sign in front of `-performance-unnecessary-value-param` disables this check.

Reference: https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html

### DIFF
--- a/cocos/base/ObjectFactory.cpp
+++ b/cocos/base/ObjectFactory.cpp
@@ -23,8 +23,9 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 ****************************************************************************/
 
-#include <functional>
 #include "base/ObjectFactory.h"
+#include <functional>
+#include <utility>
 
 
 NS_CC_BEGIN
@@ -47,7 +48,7 @@ ObjectFactory::TInfo::TInfo(const std::string& type, Instance ins)
 ObjectFactory::TInfo::TInfo(const std::string& type, InstanceFunc ins)
     :_class(type)
     ,_fun(nullptr)
-    ,_func(ins)
+    ,_func(std::move(ins))
 {
     ObjectFactory::getInstance()->registerType(*this);
 }
--- a/cocos/base/ObjectFactory.cpp
+++ b/cocos/base/ObjectFactory.cpp
@@ -45,10 +45,10 @@ ObjectFactory::TInfo::TInfo(const std::string& type, Instance ins)
     ObjectFactory::getInstance()->registerType(*this);
 }
 
-ObjectFactory::TInfo::TInfo(const std::string& type, InstanceFunc ins)
+ObjectFactory::TInfo::TInfo(const std::string& type, const InstanceFunc& ins)
     :_class(type)
     ,_fun(nullptr)
-    ,_func(std::move(ins))
+    ,_func(ins)
 {
     ObjectFactory::getInstance()->registerType(*this);
 }
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -10,6 +10,7 @@ Checks: >
   performance-move-constructor-init,
   performance-type-promotion-in-math-fn,
   performance-unnecessary-copy-initialization,
+  -performance-unnecessary-value-param,
   readability-container-size-empty,
   readability-delete-null-pointer,
   readability-redundant-control-flow,
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -10,7 +10,7 @@ Checks: >
   performance-move-constructor-init,
   performance-type-promotion-in-math-fn,
   performance-unnecessary-copy-initialization,
-  -performance-unnecessary-value-param,
+  performance-unnecessary-value-param,
   readability-container-size-empty,
   readability-delete-null-pointer,
   readability-redundant-control-flow,
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -2802,10 +2802,10 @@ void TargetedAction::setForcedTarget(Node* forcedTarget)
 
 // ActionFloat
 
-ActionFloat* ActionFloat::create(float duration, float from, float to, ActionFloatCallback callback)
+ActionFloat* ActionFloat::create(float duration, float from, float to, const ActionFloatCallback& callback)
 {
     auto ref = new (std::nothrow) ActionFloat();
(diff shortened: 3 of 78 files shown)


crazyhappygame:
@JohnCoconut const std::function<>& should be use almost everywhere. std::function may do dynamic allocation.

There is one exception "pass by value if you are storing it." e.g.:
```
struct A
{
    A(std::function f)
    :f(std::move(f))
    {}
    std::function f;
};
```
But even above may generate 1 additional move so it's better to use "perfect forwarding" or have overloading:

```
struct A
{
    A(std::function&& f)
    :f(std::move(f))
    {}
    A(const std::function& f)
    :f(f)
    {}
    std::function f;
};
```

crazyhappygame:
So I think we shouldn't disable this check.

JohnCoconut:
I checked two large C++ code base written by Google, `fuchisa` and `abseil`. Both teams have inconsistent usage of `std::function`. Sometimes it's passed by value, sometimes by const reference.

https://github.com/abseil/abseil-cpp/blob/72e09a54d993b192db32be14c65adf7e9bd08c31/absl/synchronization/internal/thread_pool.h#L56-L60

https://github.com/abseil/abseil-cpp/blob/72e09a54d993b192db32be14c65adf7e9bd08c31/absl/debugging/internal/symbolize.h#L44-L46

You are right about memory allocation issues. This stackoverflow thread shed some ideas when `std::function` allocates memory.

https://stackoverflow.com/questions/46163607/avoid-memory-allocation-with-stdfunction-and-member-function

> On g++, a non-trivial copy constructor on a function object, or data exceeding 16 bytes, is enough to cause it to allocate. But if your function object has no data and uses the builtin copy constructor, then std::function won't allocate. Also, if you use a function pointer or a member function pointer, it won't allocate.

It's probably better to pass it by const reference then.

crazyhappygame:
https://github.com/abseil/abseil-cpp/blob/72e09a54d993b192db32be14c65adf7e9bd08c31/absl/synchronization/internal/thread_pool.h#L56-L60
This is example of "pass by value if you are storing it." 
```
struct A
{
    A(std::function f)
    :f(std::move(f))
    {}
    std::function f;
};
```

JohnCoconut:
Enable `performance-unnecessary-value-param` to allow changes on `std::function` and callback types.

cocos/base/ObjectFactory.cpp
```diff
@@ -47,7 +48,7 @@ ObjectFactory::TInfo::TInfo(const std::string& type, Instance ins)
 ObjectFactory::TInfo::TInfo(const std::string& type, InstanceFunc ins)
     :_class(type)
     ,_fun(nullptr)
-    ,_func(ins)
+    ,_func(std::move(ins))
```
minggo: Why not use `const InstanceFunc &ins` instead of `std::move` here?

crazyhappygame: Please read e.g.:
https://abseil.io/tips/117
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#exception-9

Personally I think that const& should be used here. This is what is recommended here: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f15-prefer-simple-and-conventional-ways-of-passing-information

And later if optimization is needed then pass by && or "perfect forwarding".

So personally I do not like "pass by value and then move" but some C++ experts and clang-tidy claim that's OK.

Change to "const&" will also fix this warning but there is another check for storing temporary variables which should be fixed with: 
"pass by &&" or "perfect forwarding" or "pass by value+std::move"

minggo: > Change to "const&" will also fix this warning but there is another check for storing temporary variables which should be fixed with:
"pass by &&" or "perfect forwarding" or "pass by value+std::move"

I don't quite understand why this is a warning.

crazyhappygame: InstanceFunc is std:function. std::function could be expensive to copy (e.g. when lamda capture big object  std::function do allocation).

minggo: But it is copied when passing by value, isn't it?

crazyhappygame: No, if you call like that:
```
f(std::move(instanceFun));
```

JohnCoconut: It's copied once when passed to constructor. 
```cpp
ObjectFactory::TInfo::TInfo(const std::string& type, InstanceFunc ins)
```
It will be copied again in member intializer.

```cpp
,_func(ins)
```

I have a example here. https://wandbox.org/permlink/PXqmQmvCgdzvEFNF

JohnCoconut:
I suggest measuring the performance with this change. I doubt we will see much changes. I see that we have the following projects in `performance-tests` directory.

```bash
proj.android/  proj.ios/  proj.mac/  proj.win32/
```

There is no perf test for linux? What about `proj.android`, does it run on Linux?

cocos/base/ObjectFactory.cpp
crazyhappygame: Here the same without copy but with moves:
https://wandbox.org/permlink/dW0KTViJhS9YNdnz

minggo: @crazyhappygame yep, in your test case, it is not copied, but the invoker should pass using `std::move()`. Which means the invoker should be very careful to use it, if passing by reference, then invoker don't have to take care of it.

minggo:
@JohnCoconut i am not sure if `performance-tests` can work or not. But i think we should change it no matter how much it is improved. At least, it makes codes clean and more c++ friendly.

cocos/base/ObjectFactory.cpp
crazyhappygame: Agree. I prefer pass by const & (and than pass by && if needed)

I think that there is another clang-tidy to check if std::move is used where needed https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html

JohnCoconut:
If a object is passed by value and moved to new value. Prefer passing it by reference and copy it, if allowed.

PR #19882, opened 2019-06-28
add setOptimise function for c++

xianyinchen:
update OPPO SDK.

### DIFF
--- a/cocos/platform/CCDataManager.h
+++ b/cocos/platform/CCDataManager.h
@@ -47,6 +47,7 @@ public:
     static void onSceneLoaderEnd();
     static void onShaderLoaderBegin();
     static void onShaderLoaderEnd();
+    static void setOptimise(const char* thing, const char* value);
 };
 
 // end of platform group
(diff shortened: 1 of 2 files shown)


cocos/platform/CCDataManager.h
```diff
@@ -47,6 +47,7 @@ class CC_DLL DataManager
     static void onSceneLoaderEnd();
     static void onShaderLoaderBegin();
     static void onShaderLoaderEnd();
+    static void setOptimise(const char* thing, const char* value);
```
minggo: please use `const std::string&` instead

cocos/platform/CCDataManager.h
```diff
@@ -47,7 +48,7 @@ class CC_DLL DataManager
     static void onSceneLoaderEnd();
     static void onShaderLoaderBegin();
     static void onShaderLoaderEnd();
-    static void setOptimise(const char* thing, const char* value);
+    static void setOptimise(std::string thing, std::string value);
```
minggo: std::string -> const std::string&

minggo (approved):

cocos/platform/CCDataManager.h
JohnCoconut: @xianyinchen   You forget to add `const` in your new commit.

Use `const std::string& thing` instead of `std::string&`.

minggo: Oh, i didn't notice it.

PR #19883, opened 2019-06-28
Feature/fix tile map

Mee-gu:
fix tile map issues


tests/cpp-tests/Classes/TileMapTest/TileMapTest.cpp
```diff
@@ -88,6 +88,7 @@ TileDemo::TileDemo()
     // "test" is the default value in Director::setGLDefaultValues()
     // but TransitionTest may setDepthTest(false), we should revert it here
     Director::getInstance()->setDepthTest(true);
+    Director::getInstance()->getRenderer()->setDepthWrite(true);
```
minggo, line 91: I think it is better to delete `Director::setDepthTest()`, and uses renderer to set all the global states.

Mee-gu:
update review

minggo (approved):

minggo:
Should add to docs that, `Director::setDepth()` is removed.

PR #19891, opened 2019-07-02
fix warnings

minggo:


crazyhappygame:
Consider use below clang-tidy check for that: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html

As far as I remember there is much more places where override is missing.

Also I think that recommend is to do not put virtual when there is override.

JohnCoconut:
Are you fine with the style from `clang-tidy`. It suggests,

```cpp
virtual void A::foo();

void B::foo() override;  

// virtual void B::foo() override;  /* `override` indicates it's a virtual function. */
```

If you agree, we can add `clang-tidy`'s `modernize-use-override` option.

minggo:
I think put virtual make it more clear.

PR #19892, opened 2019-07-02
add setInsertionPointColor to EditBox

zhiqiangxu:
The default text cursor color is black, if the background of editbox is also black, the cursor will be invisible, in that occation, we may want to change the text cursor color.

This patch adds `setInsertionPointColor` to `EditBox`, only implemented for `ios/mac`, leave other platforms unimplemented.

Example that uses this api is here: https://github.com/zhiqiangxu/client/blob/master/editor/Classes/EditorScene.cpp#L22


minggo:
Please implement it on as more platforms as possible, especially on Android.

zhiqiangxu:
@minggo 我主要搞服务端IM，今天开始又要忙了。。有时间的话我会增加android实现，但不保证有时间。。可以的话先合进去，因为只是加接口，不会破坏已有功能：）

PR #19900, opened 2019-07-03
clang tidy modernize redundant void arg

JohnCoconut:
* add new clang-tidy option above
* fix warnings from performance-unnecessary-value-param

1. This commit changes `int f(void)` to `int f()`, removing redundant void arg.

2. The `performance-unnecessary-value-param` warnings are introduced from this PR, [updating luabinding & jsbinding & cocos_file.json automatically](https://github.com/cocos2d/cocos2d-x/pull/19895). Is this generated by a script, and how can we fix it? It went through because [ci skip] is used.

Reference: https://releases.llvm.org/7.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-redundant-void-arg.html

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -24,6 +24,7 @@ Checks: >
   bugprone-forwarding-reference-overload,
   bugprone-inaccurate-erase,
   bugprone-incorrect-roundings,
+  modernize-redundant-void-arg,
 
 WarningsAsErrors: '*'
 HeaderFilterRegex: '/(?!external)/.*'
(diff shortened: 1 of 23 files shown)


minggo:
About binding codes, i think we should modify [here](https://github.com/cocos2d/bindings-generator/blob/2aa9b21f11bf514ca80f243b21750e8c7c28f05e/targets/spidermonkey/templates/lambda.c#L6). But can we ignore the binding codes, as we will not update JS binding codes.

JohnCoconut:
> But can we ignore the binding codes, as we will not update JS binding codes.

@mingo Sorry for late reply. I was busy last week. Is it appropriate to just disable clang-tidy on `cocos/scripting/js-bindings` and `cocos/scripting/lua-bindings` directories?

minggo:
Sure.

PR #19902, opened 2019-07-03
refactor ObjectFactory

JohnCoconut:
* remove redundant copy constructor, copy assignment operator, and
destructor in TInfo.
* prefer in-class member initialization.
* remove <functional> in cpp file as it's already included in header file.

### DIFF
--- a/cocos/base/ObjectFactory.cpp
+++ b/cocos/base/ObjectFactory.cpp
@@ -24,69 +24,26 @@ THE SOFTWARE.
 ****************************************************************************/
 
 #include "base/ObjectFactory.h"
-#include <functional>
 #include <utility>
 
 
 NS_CC_BEGIN
 
-ObjectFactory::TInfo::TInfo()
-:_class("")
-,_fun(nullptr)
-,_func(nullptr)
-{
-}
-
 ObjectFactory::TInfo::TInfo(const std::string& type, Instance ins)
 :_class(type)
 ,_fun(ins)
-,_func(nullptr)
 {
     ObjectFactory::getInstance()->registerType(*this);
 }
 
 ObjectFactory::TInfo::TInfo(const std::string& type, const InstanceFunc& ins)
-    :_class(type)
-    ,_fun(nullptr)
-    ,_func(ins)
+:_class(type)
+,_func(ins)
 {
     ObjectFactory::getInstance()->registerType(*this);
 }
 
-ObjectFactory::TInfo::TInfo(const TInfo &t)
-{
-    _class = t._class;
-    _fun = t._fun;
-    _func = t._func;
-}
-
-ObjectFactory::TInfo::~TInfo()
-{
-   _class = "";
-   _fun = nullptr;
-   _func = nullptr;
-}
-
-ObjectFactory::TInfo& ObjectFactory::TInfo::operator= (const TInfo &t)
-{
-    _class = t._class;
-    _fun = t._fun;
-    _func = t._func;
-    return *this;
-}
-
-
 ObjectFactory* ObjectFactory::_sharedFactory = nullptr;
-
-ObjectFactory::ObjectFactory()
-{
-
-}
-
-ObjectFactory::~ObjectFactory()
-{
-    _typeMap.clear();
-}
 
 ObjectFactory* ObjectFactory::getInstance()
 {
@@ -104,20 +61,14 @@ void ObjectFactory::destroyInstance()
 
 Ref* ObjectFactory::createObject(const std::string &name)
 {
-    Ref *o = nullptr;
-    do 
-    {
-        const TInfo t = _typeMap[name];
-        if (t._fun != nullptr)
-        {
-            o = t._fun();
-        }else if (t._func != nullptr)
-        {
-            o = t._func();
-        }
-    } while (0);
-   
-    return o;
+    const TInfo t = _typeMap[name];
+    if (t._fun != nullptr) {
+        return t._fun();
+    }
+    if (t._func != nullptr) {
+        return t._func();
+    }
+    return nullptr;
 }
 
 void ObjectFactory::registerType(const TInfo &t)
--- a/cocos/base/ObjectFactory.h
+++ b/cocos/base/ObjectFactory.h
@@ -26,11 +26,11 @@ THE SOFTWARE.
 #ifndef __TRIGGERFACTORY_H__
 #define __TRIGGERFACTORY_H__
 
-#include <string>
-#include <unordered_map>
-#include <functional>
 #include "base/CCRef.h"
 #include "platform/CCPlatformMacros.h"
+#include <functional>
+#include <string>
+#include <unordered_map>
 
 NS_CC_BEGIN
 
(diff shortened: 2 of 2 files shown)


cocos/base/ObjectFactory.h
```diff
@@ -44,12 +44,9 @@ class CC_DLL ObjectFactory
...
         TInfo(const std::string& type, Instance ins = nullptr);
         TInfo(const std::string& type, const InstanceFunc& ins = nullptr);
-        TInfo(const TInfo &t);
-        ~TInfo();
-        TInfo& operator= (const TInfo &t);
         std::string _class;
-        Instance _fun;
-        InstanceFunc _func;
+        Instance _fun = nullptr;
+        InstanceFunc _func = nullptr;
```
crazyhappygame: This is initialisation is redundant. std::function default constructor will do this

cocos/base/ObjectFactory.h
```diff
@@ -44,12 +44,9 @@ class CC_DLL ObjectFactory
         TInfo();
         TInfo(const std::string& type, Instance ins = nullptr);
         TInfo(const std::string& type, const InstanceFunc& ins = nullptr);
-        TInfo(const TInfo &t);
-        ~TInfo();
-        TInfo& operator= (const TInfo &t);
         std::string _class;
-        Instance _fun;
-        InstanceFunc _func;
+        Instance _fun = nullptr;
```
crazyhappygame, line 48: Why this class has both c pointer function and std::function? std::function should be enough.

cocos/base/ObjectFactory.h
```diff
@@ -44,12 +44,9 @@ class CC_DLL ObjectFactory
         TInfo();
```
crazyhappygame: = default

cocos/base/ObjectFactory.cpp
```diff
@@ -46,36 +42,13 @@ ObjectFactory::TInfo::TInfo(const std::string& type, Instance ins)
 }
 
 ObjectFactory::TInfo::TInfo(const std::string& type, const InstanceFunc& ins)
-    :_class(type)
-    ,_fun(nullptr)
-    ,_func(ins)
+:_class(type)
+,_fun(nullptr)
```
crazyhappygame: Not needed

cocos/base/ObjectFactory.h
JohnCoconut: `_fun` is used in lua bindings. I am not sure how it works.

 https://github.com/cocos2d/cocos2d-x/blob/4107942cbf0b674883d1db325800c50d0d890240/cocos/scripting/lua-bindings/script/cocostudio/CocoStudio.lua#L30-L44

cocos/base/ObjectFactory.h
JohnCoconut: There is clang-tidy option for this, https://releases.llvm.org/7.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-redundant-member-init.html. We haven't enabled it yet.

I tried with a simple test. It failed to detect `std::function<T> = nullptr`.

minggo (approved):

cocos/base/ObjectFactory.cpp
```diff
@@ -24,65 +24,27 @@ THE SOFTWARE.
...
-
-
 ObjectFactory* ObjectFactory::_sharedFactory = nullptr;
 
-ObjectFactory::ObjectFactory()
-{
-
-}
-
 ObjectFactory::~ObjectFactory()
```
crazyhappygame: Not needed.
Default destructor do the same

cocos/base/ObjectFactory.h
```diff
@@ -41,14 +41,11 @@ class CC_DLL ObjectFactory
     typedef std::function<cocos2d::Ref* ()> InstanceFunc;
     struct CC_DLL TInfo
     {
-        TInfo();
+        TInfo() = default;
         TInfo(const std::string& type, Instance ins = nullptr);
```
crazyhappygame: I think that default =nullptr should be removed for both constructors.

cocos/base/ObjectFactory.h
crazyhappygame: I think that link cocos2d-x/cocos/scripting/lua-bindings/script/cocostudio/CocoStudio.lua is just the same implementation but in Lua.
So we can remove _fun from C++ and have only std::function

cocos/base/ObjectFactory.cpp
JohnCoconut: It's declared as virtual in header file. But we can simplify it with `virtual ~ObjectFactory() = default;`.

cocos/base/ObjectFactory.h
JohnCoconut: Yeah. It causes ambiguous overload resolution.
```cpp
TInfo(const std::string& type, Instance ins = nullptr);
TInfo(const std::string& type, const InstanceFunc& ins = nullptr);
```

The following code wouldn't compile because of it.

```cpp
int main(int argc, char **argv)
{
    ObjectFactory::TInfo t("classname");
}
```

Need to remove default parameters from them.

cocos/base/ObjectFactory.cpp
JohnCoconut: But it doesn't provide any other virtual method, so I am not sure why `virtual` is needed in destructor.

crazyhappygame: We may declare this class as final and remove virtual destructor or if that fails leave virtual destructor and add =default

JohnCoconut:
I will see how to remove `_fun` member from `TInfo` properly sometime later.

@minggo The commit message is very messy. Please let me resolve all review issues and do a rebase afterwards.

cocos/base/ObjectFactory.cpp
```diff
@@ -105,17 +62,13 @@ void ObjectFactory::destroyInstance()
...
-        {
-            o = t._fun();
-        }else if (t._func != nullptr)
-        {
-            o = t._func();
-        }
-    } while (0);
+    const TInfo t = _typeMap[name];
+    if (t._fun != nullptr) {
+        o = t._fun();
```
crazyhappygame: Consider change to return and remove "o" variable

crazyhappygame (approved):

cocos/base/ObjectFactory.cpp
JohnCoconut: Done.

I would like to change `_fun` member in TInfo in a separate PR. All changes made so far are limited to these two files. Changing `_fun` will involves more.

cocos/base/ObjectFactory.cpp
```diff
@@ -104,20 +61,14 @@ void ObjectFactory::destroyInstance()
...
-            o = t._func();
-        }
-    } while (0);
-   
-    return o;
+    const TInfo t = _typeMap[name];
+    if (t._fun != nullptr) {
+        return t._fun();
+    }
+    else if (t._func != nullptr) {
```
crazyhappygame: It doesn't make sense to have else after return.
https://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html

crazyhappygame (approved):

JohnCoconut:
more fixes to ObjectFactory
    
    * implement removeAll()
    * exit early
    * remove typedef that's only used in one place and in private scope


`FactoryMap` is only used once, and it's in private access modifier, so remove it.

`removeAll() ` is declared but has never been implemented.

JohnCoconut:
rebase

cocos/base/ObjectFactory.cpp
```diff
@@ -104,25 +61,24 @@ void ObjectFactory::destroyInstance()
...
+    }
+    return nullptr;
 }
 
 void ObjectFactory::registerType(const TInfo &t)
 {
     _typeMap.emplace(t._class, t);
 }
 
+void ObjectFactory::removeAll()
```
crazyhappygame: If this method was not implemented I propose to remove it.

PR #19904, opened 2019-07-04
V4 speedup sprite creation

PatriceJiang:
ref https://github.com/cocos2d/cocos2d-x/issues/19907

### DIFF
--- a/cocos/2d/CCSprite.cpp
+++ b/cocos/2d/CCSprite.cpp
@@ -45,6 +45,15 @@ THE SOFTWARE.
 
 NS_CC_BEGIN
 
+
+static struct {
+    backend::ProgramState *ps = nullptr;
+    backend::UniformLocation mvp;
+    backend::UniformLocation texture;
+    backend::UniformLocation alphaTexture;
+    backend::VertexLayout layout;
+} _defaulCommandInfo;
+
 // MARK: create, init, dealloc
 Sprite* Sprite::createWithTexture(Texture2D *texture)
 {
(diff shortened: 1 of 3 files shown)


crazyhappygame:
Could you explain what was performance problem and how you improve it? (benchmarks)

PatriceJiang:
Creating 40000 sprites takes 130ms in v4 and 28ms in v3.  (Tested on Linux, i5 3GHz)
This PR reduces time to 62ms when creating sprites with default shaders.

crazyhappygame:
Could you explain where is speedup for this implementation?

2x improvment is not bad but anyway it is 2x slower than v3....

cocos/2d/CCSprite.cpp
```diff
@@ -372,11 +381,48 @@ void Sprite::setVertexLayout()
...
+
+
 void Sprite::updateShaders(const char* vert, const char* frag)
 {
-    auto programState = new (std::nothrow) backend::ProgramState(vert, frag);
-    setProgramState(programState);
-    CC_SAFE_RELEASE_NULL(programState);
+    if(_defaulCommandInfo.ps != nullptr && vert == positionTextureColor_vert && frag == positionTextureColor_frag)
+    {
+        auto programState = _defaulCommandInfo.ps->clone();
```
PatriceJiang, line 390: `ProgramState::clone` is faster than `ProgramState::ProgramState(const char*, const char *)`

PatriceJiang:
There are more abstractions on ProgramState in v4, it is hard to achieve the same speed as v3, and we may improve it in the future.

cocos/2d/CCSprite.cpp
crazyhappygame: This is strange....
There is already in ProgramState::ProgramState(const char*, const char *) one cache
` _program = backend::ProgramCache::getInstance()->newProgram`
and with this PR we would like to have another cache on top of backend::ProgramCache?

PatriceJiang: We have discussed the performance issue of ProgramCache yesterday, and we will fix that later.

PatriceJiang: Since we will optimize ProgramCache, this PR is not needed.

PR #19913, opened 2019-07-11
CJK punctualion linebreaking support

sbrednikhin:
CJK line breaking punctuation rules support. Rules and symbols are mostly taken from here
https://en.wikipedia.org/wiki/Line_breaking_rules_in_East_Asian_languages


sbrednikhin:
Looks like there are some inaccuracies in Korean language section with 〈《 and 〉》 symbols in this article
https://en.wikipedia.org/wiki/Line_breaking_rules_in_East_Asian_languages
It looks like this characters are in wrong places: 〈《 shouldn't be allowed to end the line and vice versa. I've changed it, but I can't find good source in English to approve my guess. If someone knows Korean - say, please, if I'm wrong. I'll Fix it.

Using of cursor as some symbol going by '\b' symbol could cause breaking the word wrapping in text fields earlier. Now it will be fine.

PR #19914, opened 2019-07-11
Fix error with update line height

kasandko:
After movig font atlas `reinit()` (https://github.com/cocos2d/cocos2d-x/pull/19384) there was indused a problem: the first display of text with a outline in the label does not update line height. This is because the atlas `reinit()` does not occur when the outline is activated. In this case, a new atlas will be created, but the line height will not be taken from it, because line height calculate in `updateContent()`. I propose line height calculating to return to the font atlas constructor.

### DIFF
--- a/cocos/2d/CCFontAtlas.cpp
+++ b/cocos/2d/CCFontAtlas.cpp
@@ -72,6 +72,12 @@ FontAtlas::FontAtlas(Font &theFont)
             _letterPadding += 2 * FontFreeType::DistanceMapSpread;    
         }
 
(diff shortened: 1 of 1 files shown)


aardvarkk:
@kasandko I believe I'm running into this issue in my current project -- when I add an outline certain objects appear out of place on occasion. My guess is that it's due to the line height getting out of sync. Do you know if there's a workaround until this gets merged?

aardvarkk:
I can confirm that your PR fixes my alignment issues with outlined text.

PatriceJiang (approved):

PR #19917, opened 2019-07-12
[iOS EditBox] add toolbar to exit out of numeric keyboard

saumyamukul:
There is no way to exit out of the keypad. This adds a "Done" button


saumyamukul:
What it looks like

IMG_0076

minggo:
The keyboard will disappear if you click the area that outside of EditBox.

saumyamukul:
Thats a good point. We've implemented accessibility for our cocos2d-x application, and with Voiceover on, it is not possible for the user to exit the keypad. 
Feel free to close this though. Anyone who wants this functionality can refer to this MR :)

minggo:
Thanks.

PR #19921, opened 2019-07-16
remove vertex step mode

Mee-gu:
remove vertex step mode


minggo:
Should check if lua codes are affected.

Mee-gu:
Lua works fine.

PR #19926, opened 2019-07-17
Fix cocos2d-js issue: Font height was not getting calculated properly in iOS.

vaibhavgohel:
For calculating font height for iOS we were using wrong option. We were using [nsstringdrawingusesfontleading](https://developer.apple.com/documentation/uikit/nsstringdrawingoptions/nsstringdrawingusesfontleading?language=objc) which include [font leading](https://en.wikipedia.org/wiki/Leading)(Extra padding need to be added before starting next line) in total height. Due to this text was not getting vertically middled properly. We should only use NSStringDrawingUsesLineFragmentOrigin as it automatically include font leading in case we provide multiline text. 

You can use calibri font to check the issue.

### DIFF
--- a/cocos/platform/ios/CCDevice-ios.mm
+++ b/cocos/platform/ios/CCDevice-ios.mm
@@ -145,7 +145,7 @@ static CGSize _calculateShrinkedSizeForString(NSAttributedString **str,
             *str = __attributedStringWithFontSize(mutableString, fontSize);
             
             CGSize fitSize = [*str boundingRectWithSize:CGSizeMake(constrainSize.width, MAX_MEASURE_HEIGHT)
-                                    options:(NSStringDrawingUsesLineFragmentOrigin|NSStringDrawingUsesFontLeading)
+                                    options:(NSStringDrawingUsesLineFragmentOrigin)
                                     context:nil].size;
 
             if (fitSize.width == 0 || fitSize.height == 0) {
@@ -364,7 +364,7 @@ static CGSize _calculateStringSize(NSAttributedString *str, id font, CGSize *con
 
     CGSize dim;
     dim = [str boundingRectWithSize:CGSizeMake(textRect.width, textRect.height)
-                                 options:(NSStringDrawingUsesLineFragmentOrigin|NSStringDrawingUsesFontLeading)
+                                 options:(NSStringDrawingUsesLineFragmentOrigin)
                             context:nil].size;
 
     dim.width = ceilf(dim.width);


minggo:
How to reproduce it?

vaibhavgohel:
Draw two text, one with arial font and other with Calibri font. Draw rect around both label, you can see the difference. PFA screenshot and code. in screenshot label above is drawn with calibri and label below is drawn with Arial font. You can check the [mac code](https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/platform/mac/CCDevice-mac.mm) as well  that is already committed in cocos repo where we are using only NSStringDrawingUsesLineFragmentOrigin option.
[app.js.zip](https://github.com/cocos2d/cocos2d-x/files/3404923/app.js.zip)
IMG_4E90699C1B6E-1

minggo:
Is the screen the correct effect? If so, then what's the wrong effect?

vaibhavgohel:
> Is the screen the correct effect? If so, then what's the wrong effect?

No, the screen is not correct effect. As if you want to vertically centered the label with respect to some components then text will not be vertically centered. You can check the same code on the browser to check the effect. PFB correct effect after the fix.
IMG_D609974672BF-1

minggo:
Screen Shot 2019-07-18 at 4 07 05 PM
This is the effect without your patch, it seems it is correct. It runs on simulator.

vaibhavgohel:
> Screen Shot 2019-07-18 at 4 07 05 PM
> 
> This is the effect without your patch, it seems it is correct. It runs on simulator.

Hi,  I can see in the screenshot that Calibri font is not loaded in your simulator.  Download font from [here](https://www.wfonts.com/font/calibri), Add the font to your project, Add font path to your info.plist(You can refer to [this](https://kitefaster.com/2017/08/03/custom-fonts-ios/) link to add custom font in your application). 

The issue is only reproducing where the font creator has specify different font leading instead of default font leading. I was able to reproduce the issue with Calibri font but there might be other fonts for which this issue is reproducing.

minggo:
Yep, i can reproduce it. After applying this patch, it looks better, but it is not as good as default font. It seems it is still not in center, a little above center.

vaibhavgohel:
> patch

I checked this against the browser, both looks same. I also tried to get the text height in native swift as well, i'm getting the same result. so i think the issue you mention is because of look and feel of the font.

minggo:
Yep, may be. Thanks, i will merge it.

PR #19927, opened 2019-07-18
Re-implement ControllerImpl with GLFW 3.3

ayuusweetfish:
This is a new implementation of `ControllerImpl` with the gamepad API provided by GLFW 3.3 (cocos2d/cocos2d-x-3rd-party-libs-src#177). Mappings of joystick buttons are made more complete and consistent across platforms. No compatibility breakage is introduced.

The most part is based on the original `CCController-linux-win32.cpp`, and I followed this file's code style which slightly differs from the overall one.

I have tested the change on macOS 10.13.5 and Windows 10 (with manual upgrades of GLFW prebuilt libraries) on both branches `v3` and `v4` with a PS4 controller. In particular, I noticed a change of button mappings on macOS (△○╳□ = YXBA → YBAX), since the original version is based on Apple's Game Controller framework and thus uses a different (somehow incomplete?) database. Such inconsistencies may also be present on other platforms, as the mappings may differ in RetroArch's database and GLFW's SDL_GameControllerDB.

In case anything is to be changed, please kindly specify. Hope it will be of help ^ ^


ayuusweetfish:
Related to #19836 & #19856; Compilation was successful with my change applied, in the VS 2019 Win32 project generated by CMake. (Haven't tested the original version.)

rh101:
@minggo What @kawa-yoiko has done is great! In order for this PR to work GLFW has to be updated to version 3.3, as specified in #19610 

This does fix the compilation issues for VS2019, so there is no need for a work-around related to the PR #19856 

Are you sure you want to wait till after v4 is released?

ayuusweetfish:
@rh101 Thanks for the appreciation >ᴗ<

Can you try to build GLFW with [this PR](https://github.com/cocos2d/cocos2d-x-3rd-party-libs-src/pull/177) on macOS? It is put off because it doesn't work on the dev team's macOS environment, and I don't know why. It will be great if you could help by trying it out!

rh101:
Tested it out, and it worked. Full response in the other [PR](https://github.com/cocos2d/cocos2d-x-3rd-party-libs-src/pull/177)

cocos/base/CMakeLists.txt
```diff
@@ -1,18 +1,24 @@
...
     set(COCOS_BASE_SPECIFIC_SRC
         base/CCUserDefault-android.cpp
         base/CCController-android.cpp
         )
 elseif(LINUX OR WINDOWS)
     set(COCOS_BASE_SPECIFIC_SRC
-        base/CCController-linux-win32.cpp
+        base/CCController-desktop.cpp
         )
 endif()
```
minggo, line 23: v3 not only support cmake, it also has project files too. So need to modify the project files too.

ayuusweetfish: Updated, but I have not tested them with actual builds and hardware. I might be able to do so tomorrow!

ayuusweetfish:
Due to personal reasons I am unable to test the change with the project files changed recently, my apologies. I hope the Cocos team can help with that 😊

ayuusweetfish:
Bump with good news (o^ ^o)

With the updated project files for IDEs, I have succeeded in running `game-controller-test` under both macOS (10.13.5; Xcode 9.4.1) and Windows (10; VS 2019, with cocos2d/cocos2d-x-3rd-party-libs-bin@c448b70). Tested with a DS4 controller and an Xbox 360 controller (on macOS an additional driver 360Controller/360Controller@5b7b552 is necessary).

On macOS I initially ran into the same error as @minggo's in #19610.

> ```
> Unknown class 'AppController', using 'NSObject' instead. Encountered in Interface Builder file at path ...
> Could not connect action, target class NSObject does not respond to -toggleFullScreen:
> ```

Apparently there should be an Objective-C class `AppController` that responds to `-toggleFullScreen:` (presumably inheriting from `NSWindow`?) but it does not exist in the codebase (there are many `AppController`'s for iOS but none for macOS). My workaround was to simply remove `proj.mac/MainMenu.xib` (as in #19963) and the application started to work normally. I'm unaware of a better approach, but this method indeed is a viable fallback whose only downside is the loss of menu bar items.

I have added no changes for this issue due to unfamiliarity with Xcode, and I hope for help from the team to update the Xcode project in a way deemed appropriate. The rest are all ready!

PR #19934, opened 2019-07-19
fix header define for shader.

Mee-gu:
fix header define for shader.

### DIFF
--- a/cocos/renderer/backend/opengl/ProgramGL.cpp
+++ b/cocos/renderer/backend/opengl/ProgramGL.cpp
@@ -37,16 +37,16 @@ CC_BACKEND_BEGIN
 ProgramGL::ProgramGL(const std::string& vertexShader, const std::string& fragmentShader)
 : Program(vertexShader, fragmentShader)
 {
-    _vertexShaderModule = static_cast<ShaderModuleGL*>(ShaderCache::newVertexShaderModule(_vertexShader));
 #if CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID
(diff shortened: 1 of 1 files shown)


cocos/renderer/backend/opengl/ProgramGL.cpp
```diff
@@ -37,16 +37,16 @@ CC_BACKEND_BEGIN
...
-    std::string headersDef("#ifdef GL_FRAGMENT_PRECISION_HIGH\n" \
-                           "precision highp float;\n precision highp int;\n" \
-                           "#else\n" \
-                           "precision mediump float;\n precision mediump int;\n"
-                           "#endif\n");
-    headersDef.append(fragmentShader);
-    _fragmentShaderModule = static_cast<ShaderModuleGL*>(ShaderCache::newFragmentShaderModule(headersDef));
+    //some device required manually specify the precision qualifiers for vertex shader.
+    std::string vsPreDefine("#version 100\n precision highp float;\n precision highp int;\n");
+    std::string fsPreDefine("precision mediump float;\n precision mediump int;\n");
```
minggo, line 43: vsPreDefine and fsPreDefine are const strings, it is better to make them as const member variables.

Mee-gu: I think `vsPreDefine` is better as a non-const string since it will append with shader source.

minggo: But you can insert vsPreDefine at the begin of the shader.

minggo: Ok, it is not suitable, ignore it.

PR #19955, opened 2019-07-19
clang-tidy clang-tidy-modernize-use-emplace

JohnCoconut:
A example of clang-tidy fix is shown below.

Before:

```cpp
outPoints.push_back(Vec2(pt.X/PRECISION, pt.Y/PRECISION));
```

After

```cpp
outPoints.emplace_back(pt.X/PRECISION, pt.Y/PRECISION);
```

Reference: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-emplace.html

### DIFF
--- a/cocos/editor-support/cocostudio/CCColliderDetector.cpp
+++ b/cocos/editor-support/cocostudio/CCColliderDetector.cpp
@@ -194,6 +194,7 @@ void ColliderDetector::addContourData(ContourData *contourData)
     std::vector<Vec2> &calculatedVertexList = colliderBody->_calculatedVertexList;
     
     unsigned long num = contourData->vertexList.size();
+    calculatedVertexList.reserve(num);
     for (unsigned long i = 0; i < num; i++)
     {
         calculatedVertexList.emplace_back();
--- a/cocos/editor-support/cocostudio/CCColliderDetector.cpp
+++ b/cocos/editor-support/cocostudio/CCColliderDetector.cpp
@@ -196,7 +196,7 @@ void ColliderDetector::addContourData(ContourData *contourData)
     unsigned long num = contourData->vertexList.size();
     for (unsigned long i = 0; i < num; i++)
     {
-        calculatedVertexList.push_back(Vec2());
+        calculatedVertexList.emplace_back();
     }
 #endif
 }
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -25,6 +25,7 @@ Checks: >
   bugprone-inaccurate-erase,
   bugprone-incorrect-roundings,
   modernize-redundant-void-arg,
+  modernize-use-emplace,
 
 WarningsAsErrors: '*'
 HeaderFilterRegex: '/(?!external)/.*'
--- a/cocos/2d/CCAutoPolygon.cpp
+++ b/cocos/2d/CCAutoPolygon.cpp
@@ -416,7 +416,7 @@ std::vector<cocos2d::Vec2> AutoPolygon::marchSquare(const Rect& rect, const Vec2
         }
         else
         {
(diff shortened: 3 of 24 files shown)


cocos/editor-support/cocostudio/CCColliderDetector.cpp
```diff
@@ -196,7 +196,7 @@ void ColliderDetector::addContourData(ContourData *contourData)
     unsigned long num = contourData->vertexList.size();
     for (unsigned long i = 0; i < num; i++)
     {
-        calculatedVertexList.push_back(Vec2());
+        calculatedVertexList.emplace_back();
```
minggo, line 200: May be just use reserve here?

JohnCoconut: Yes. `reserve` can be used here. However, upon reading the context around this line, I found it hard to understand.

`calculatedVertexList` is a reference to a member,

```cpp
std::vector<Vec2> &calculatedVertexList = colliderBody->_calculatedVertexList;
```

It's initialized in

```cpp
#if ENABLE_PHYSICS_SAVE_CALCULATED_VERTEX
    _calculatedVertexList = Array::create();
    CC_SAFE_RETAIN(_calculatedVertexList);
#endif
```

Does the code in the macro ever run? I am not familiar with the `ColliderBody` class. The following line wouldn't compile.

```cpp
std::vector<cocos2d::Vec2> vertex_list = Array::create();
```

minggo: I am not familiar with it too. But the codes is `calculatedVertexList.emplace_back();` is not easy to understand. It seems it emplace back nothing.

JohnCoconut: `emplace_back()` means default constructor being used. For example,

```cpp
std::vector<std::string> v;
v.emplace_back();
```
appends a empty string to v.

minggo: Yep, i understand.

minggo:
@JohnCoconut will you do the same thing in v4 too? Your PRs are very good to v4 too.

JohnCoconut:
Do you want to create another clang-tidy tests build in v4 branch?

JohnCoconut:
What's the status of v4 branch? I checked out v4 branch, and it didn't compile with the following error.

```c++
../cocos/scripting/lua-bindings/auto/lua_cocos2dx_backend_auto.cpp:2014:35: error: no matching function for call to ‘cocos2d::backend::VertexLayout::setLayout(unsigned int&, cocos2d::backend::VertexStepMode&)’
 2014 |         cobj->setLayout(arg0, arg1);
      |                                   ^
In file included from ../cocos/scripting/lua-bindings/auto/lua_cocos2dx_backend_auto.cpp:5:
../cocos/renderer/backend/VertexLayout.h:82:10: note: candidate: ‘void cocos2d::backend::VertexLayout::setLayout(unsigned int)’
   82 |     void setLayout(unsigned int stride);
      |          ^~~~~~~~~
../cocos/renderer/backend/VertexLayout.h:82:10: note:   candidate expects 1 argument, 2 provided
```

minggo:
> Do you want to create another clang-tidy tests build in v4 branch?

What did you mean another clang-tidy tests?

About the error, i will check it.

JohnCoconut:
> What did you mean another clang-tidy tests?

clang-tidy tests isn't in v4 branch. It's only available in v3 branch. 

https://github.com/cocos2d/cocos2d-x/blob/21bf81848244c758438ce48d568d3acbe0812d43/tools/travis-scripts/run-script.sh#L40-L50

minggo:
Yep, we wan to add it in v4 too.

PR #19959, opened 2019-07-22
Update 3rd party lib references

rh101:
Update to v3-deps-159


minggo:
Could you please fix the compiling error?

rh101:
I updated it for both v3 and v4. Sorry about that, I can't figure out why it didn't give me an error when compiling it for v3, otherwise I would have caught the issue.

rh101:
Will you create new deps zip with the v3-deps-160 tag, or just update the current v3-deps-159?

minggo:
I just reuse v3-dpes-159.

PR #19960, opened 2019-07-23
Small performance boost in using Mat4::operator*= instead of operator*

tankorsmash:
This avoids creating a new temporary Mat4 transform in operator* https://github.com/cocos2d/cocos2d-x/blob/v3/cocos/math/Mat4.inl#L60

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -1686,7 +1686,7 @@ Mat4 Node::getNodeToParentTransform(Node* ancestor) const
 
     for (Node *p = _parent;  p != nullptr && p != ancestor ; p = p->getParent())
     {
-        t = p->getNodeToParentTransform() * t;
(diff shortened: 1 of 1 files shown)


acvrp-lab:
there's definitely some bug about this modification somewhere, although i havent done much research. maybe because *= returns a reference to itself and assign to itself

buttons cant be clicked on if it or one of it's parent nodes is scaled up. there must be some miscalculation on it's bounding box.

minggo:
Yep, it has problem.
```c++
t *= p->getNodeToParentTransform() => t = t * getNodeToParentTransform().
```
It is different from original logic.

@azurebld thanks for pointing it out.

minggo:
@takaken1994 i have to revert this PR.

minggo:
It is reverted in https://github.com/cocos2d/cocos2d-x/pull/19975

PR #19963, opened 2019-07-23
fix glfw3.3 no respond issue

Mee-gu:
fix glfw3.3 no respond issue


rh101:
For the sake of understanding what was causing the issue, what is the "MainMenu.xib" file reference that was removed?  I've never seen that file before.

Mee-gu:
@rh101 A xib file usually specifies one view controller or menu bar. Since `MainMenu.xib` was never used by the engine, it's no surprise that you are never seen that file before. I'm not digging deep into that file, so I don't know what exactly causes the issue either.

minggo (approved):

PR #19964, opened 2019-07-23
Fix travis script 

PatriceJiang:


.gitignore
```diff
@@ -175,3 +175,6 @@ metal-support*.zip
 **/org.eclipse.buildship.core.prefs
 
 desktop.ini
+
+
+tools/travis-scripts/test.sh
```
minggo: Is it a mistake modification?

PatriceJiang: yes, I will remove this

tools/travis-scripts/run-script.sh
```diff
@@ -151,6 +151,7 @@ function generate_pull_request_for_binding_codes_and_cocosfiles()
     local COCOS_BRANCH="update_lua_bindings_$ELAPSEDSECS"
     local COMMITTAG="[ci skip][AUTO]: updating luabinding & cocos_file.json automatically"
     local PULL_REQUEST_REPO="https://api.github.com/repos/cocos2d/cocos2d-x/pulls"
+    local COCOSFILE_PATH="$COCOS2DX_ROOT/templates/cocos2dx_files.json"
```
minggo: I don't quite understand what's the error. Is it because the label is used without definition?

minggo: COCOSFILE_PATH is used in `update_cocos_files()` too, so i think should declare this variable at the begin as COCOS2DX_ROOT.

PatriceJiang: yes, it is not declared in the function's scope

.gitignore
```diff
@@ -175,3 +175,4 @@ metal-support*.zip
 **/org.eclipse.buildship.core.prefs
 
 desktop.ini
+
```
PatriceJiang:
```suggestion
```

PR #19969, opened 2019-07-23
refactor-CCDrawNode

JohnCoconut:
* remove unecessary intermediate temp variables
* remove redundant pointer cast
* remove redundant memset, as the values are assigned immediately in a
for loop.
* minor formatting

### DIFF
--- a/cocos/2d/CCDrawNode.cpp
+++ b/cocos/2d/CCDrawNode.cpp
@@ -43,8 +43,7 @@ static Vec2 v2fzero(0.0f,0.0f);
 
 static inline Vec2 v2f(float x, float y)
 {
-    Vec2 ret(x, y);
-    return ret;
+    return {x, y};
 }
 
 static inline Vec2 v2fadd(const Vec2 &v0, const Vec2 &v1)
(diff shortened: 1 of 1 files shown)


JohnCoconut:
@minggo Do you know why are v2xx functions necessary, Like `v2f` and `__v2f`? It seems to me they are not doing anything particular. `__v2f` is prefixed with double underscores, which is not allowed in standard.

There is a comment saying,

> Vec2 == CGPoint in 32-bits, but not in 64-bits (OS X), that's why the "v2f" functions are needed 

Could you explain it further more? 

https://github.com/cocos2d/cocos2d-x/blob/a40444c5c1bfb7338c82a8821655f88e7ebcbb4c/cocos/2d/CCDrawNode.cpp#L40-L98

minggo:
I have not idea too. It seems they are not needed.

minggo:
May be the original type is not Vec2, it is other types that don't have neg, dot and other functions.

minggo:
Will you please refactor it?

JohnCoconut:
> Will you please refactor it?

Sure.

minggo:
Thanks.

PR #19988, opened 2019-07-25
travis script reduce log

PatriceJiang:
- remove `set -x` to disable echo command
- fix branch


minggo:
Is the modification related to the title?

PatriceJiang:
Yes, and I have updated the description.

PR #19996, opened 2019-07-30
don't invoke super.onPause() to prevent egl context lost

minggo:


twolight:
Hi @minggo 

The relative [PR](https://github.com/cocos2d/cocos2d-x/pull/19256) include two changes. 

is it right to do like this:

this.mGLSurfaceView.setPreserveEGLContextOnPause(true); // invoke
super.onPause(); // do not invoke

cocos2d-x version: 3.17.2

twolight:
And the  [PR](https://github.com/cocos2d/cocos2d-x/pull/19256) is  reverted  in cocos2d-x-4.0.

the result of diff cocos2d-x-3.17.2 with cocos2d-x-4.0:

image

image

PR #19999, opened 2019-07-31
fix crash when plist filename hasn't suffix

dzj0821:
should't call erase when the file name has't suffix

### DIFF
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -393,7 +393,6 @@ void SpriteFrameCache::addSpriteFramesWithFile(const std::string& plist)
         {
             texturePath = texturePath.erase(startPos);
         }
-        texturePath = texturePath.erase(startPos);
 
         // append .png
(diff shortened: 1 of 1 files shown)


cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -389,6 +389,10 @@ void SpriteFrameCache::addSpriteFramesWithFile(const std::string& plist)
 
         // remove .xxx
         size_t startPos = texturePath.find_last_of('.'); 
+        if(startPos != string::npos)
+        {
+            texturePath = texturePath.erase(startPos);
+        }
```
minggo, line 395: It will erase twice?

dzj0821: sorry i forgot it, it was fixed in the last commit

dzj0821: > It will erase twice?

sorry i forgot it, it was fixed in the last commit

PR #20007, opened 2019-08-01
in Node::enumerateChildren, searchFromParent shouldn't search from children

dzj0821:
scene tree:
scene -> node1 -> node2 -> node3
run code:
```c++
node1-> enumerateChildren("[[:alnum:]]+\..", [](Node* node){
    log("%s", node->getName().c_str());
    return false;
});
```
should output: node1
but output: node3
`\..` means that `The search should move up to the node's parent. Can only be placed at the end of string.` not move down to the node's children.

### DIFF
--- a/cocos/2d/CCNode.cpp
+++ b/cocos/2d/CCNode.cpp
@@ -850,22 +850,27 @@ void Node::enumerateChildren(const std::string &name, const std::function<bool (
     
     // Remove '//', '/..' if exist
     std::string newName = name.substr(subStrStartPos, subStrlength);
-
+    
+    const Node* target = this;
+    
     if (searchFromParent)
     {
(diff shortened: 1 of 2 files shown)


cocos/2d/CCNode.cpp
```diff
@@ -850,22 +850,27 @@ void Node::enumerateChildren(const std::string &name, const std::function<bool (
     
     // Remove '//', '/..' if exist
     std::string newName = name.substr(subStrStartPos, subStrlength);
-
+    
+    const Node* target = this;
```
minggo, line 854: Why set to const?

dzj0821: > Why set to const?

`this` ptr must be const

PR #20012, opened 2019-08-05
Fix bug when an action with RepeatForever should exec action multiple times in one frame

dzj0821:
If an action with RepeatForever should exec multiple times in one frame, it can only exec less than twice.
example:
```c++
auto scale = ScaleBy::create(0.2, 1.5);
auto repeat = RepeatForever::create(scale);
sprite->runAction(repeat);
sprite->runAction(Sequence::createWithTwoActions(DelayTime::create(1.0f), CallFunc::create([sprite](){
    log("%f", sprite->getScale());
})));
```
If I set fps is 60, console output 7.813323 (about 1.5^5, no problem)
But if I set fps is 1, console output 1.503364 (the action only exec once)
If I set fps is 2, console output 2.262938 (the action only exec twice)

### DIFF
--- a/cocos/2d/CCActionInterval.cpp
+++ b/cocos/2d/CCActionInterval.cpp
@@ -610,11 +610,30 @@ void RepeatForever::step(float dt)
     if (_innerAction->isDone() && _innerAction->getDuration() > 0)
     {
         float diff = _innerAction->getElapsed() - _innerAction->getDuration();
+        int time = 0;
         if (diff > _innerAction->getDuration())
+        {
+            time = (int)(diff / _innerAction->getDuration());
             diff = fmodf(diff, _innerAction->getDuration());
-        _innerAction->startWithTarget(_target);
-        // to prevent jerk. cocos2d-iphone issue #390, 1247
-        _innerAction->step(0.0f);
+        }
+        
+        while(time--)
+        {
+            if(_innerAction->isDone())
+            {
+                _innerAction->startWithTarget(_target);
+                // to prevent jerk. cocos2d-iphone issue #390, 1247
+                _innerAction->step(0.0f);
+            }
+            _innerAction->step(_innerAction->getDuration());
+        }
+        
+        if(_innerAction->isDone())
+        {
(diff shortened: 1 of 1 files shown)


minggo:
I don't know why it should execute multiple times in one frame. In the action system, an action can only be executed one time in one frame.

dzj0821:
> I don't know why it should execute multiple times in one frame. In the action system, an action can only be executed one time in one frame.

When playing a game on a low-performance device, it may take a long time for one frame. If the duration of the action using RepeatForever is less than the elapsed time of one frame, this action should be running multiple times in one frame to achieve the same display effect as the high performance device. (Actions are run by time rather than by frame)

for example:
```c++
auto scale = ScaleBy::create(10, 10);
sprite->runAction(scale);
sprite->runAction(Sequence::createWithTwoActions(DelayTime::create(1.0f), CallFunc::create([sprite](){
    log("%f", sprite->getScale());
})));
```
No matter how much fps, the code outputs the similar result in the console.

dzj0821:
> I don't know why it should execute multiple times in one frame. In the action system, an action can only be executed one time in one frame.

I mean that the action should run 5 times in one second when action's duration is 0.2 seconds. When fps is 60, it run 5 times correctly, but when fps is 1, it only run one time.

cocos/2d/CCActionInterval.cpp
```diff
@@ -610,12 +610,18 @@ void RepeatForever::step(float dt)
...
+        
+        while(fabsf(diff) > FLT_EPSILON && diff > 0.0f)
+        {
+            if(_innerAction->isDone())
+            {
+                _innerAction->startWithTarget(_target);
+                // to prevent jerk. cocos2d-iphone issue #390, 1247
+                _innerAction->step(0.0f);
+            }
+            _innerAction->step(diff);
```
PatriceJiang, line 622: If `_innerAction` takes longer than `duration`, the loop will never break.

kangzixiang (approved):

PR #20040, opened 2019-08-15
Use CMake 3.14's first-class support for iOS cross-compilation.

trojanfoe:
CMake 3.14 has built-in support for creating iOS/tvOS/watchOS Xcode projects without the need for a custom toolchain script.

See: https://cmake.org/cmake/help/v3.14/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos

To build for an iOS device:

	~ mkdir ios-build
	~ cd ios-build
	~ cmake .. -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_SYSROOT=iphoneos

To build for the iOS Simulator:

	~ mkdir ios-build
	~ cd ios-build
	~ cmake .. -GXcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_SYSROOT=iphonesimulator

To build for macOS:

	~ mkdir mac-build
	~ cd mac-build
	~ cmake .. -GXcode -DCMAKE_SYSTEM_NAME=Darwin

To run Metal code on the iOS Simulator you must set the deployment target to iOS 13.

Note: At the time of writing the cpp-tests project is building for iOS Simulator however it is crashing due to a shader compile error.
It runs without issue on an iOS 13 device.


minggo:
Great job, i will try to update ci's cmake version.

minggo:
@trojanfoe sorry, i don't know how to update ci's cmake version.

* On windows ci, the default cmake version is 3.15.1, but cmake is installed when installing Android SDK. More detail information can refer to [this issue](https://github.com/appveyor/ci/issues/3035).
* On travis ci, i tried many method, but can not update cmake's version. Can refer to [this PR](https://github.com/cocos2d/cocos2d-x/pull/20048).

trojanfoe:
@minggo Sorry I have no experience with Travis so I don't think I can help, however I think this might prove to be a serious issue given v4 is so cmake-focused.

crazyhappygame:
@trojanfoe @minggo I will help you.
I think that we can not upgrade to 3.14 for all platforms. Android use cmake 3.6  Travis cmake 3.11. It is not easy to upgrade. 
I think that we should upgrade cmake for iOS/mac build

CMakeLists.txt
```diff
@@ -26,7 +26,7 @@
 
 # build engine library and all tests
 
-cmake_minimum_required(VERSION 3.6)
+cmake_minimum_required(VERSION 3.14)
```
crazyhappygame, line 29: Please do not change cmake_minimum_required. Currently Android by default use cmake 3.6 and I think we should leave here 3.6 (at least for some time)

cmake/Modules/CocosConfigDefine.cmake
```diff
@@ -17,14 +17,13 @@
...
  elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-     if(IOS)
-         set(APPLE TRUE)
-         set(PLATFORM_FOLDER ios)
-     else()
-         set(APPLE TRUE)
-         set(MACOSX TRUE)
-         set(PLATFORM_FOLDER mac)
-     endif()
+     set(APPLE TRUE)
```
crazyhappygame, line 20: There is already set variable APPLE to 1. You do not have to set it to TRUE
https://cmake.org/cmake/help/latest/variable/APPLE.html

crazyhappygame (changes requested):

crazyhappygame:
For android to use cmake version from system not from Android SDK all files build.gradle should be updated:
```
android {
   externalNativeBuild {
       cmake {
           version "3.14+"
       }
   }
}
``` 
https://developer.android.com/studio/projects/add-native-code#use_a_custom_cmake_version

CMakeLists.txt
minggo: So we should use documentation to let developers know cmake 3.14+ is needed for iOS?

crazyhappygame: There are to options:
1. Use cmake 3.14 everywhere
That means update Travis to cmake 3.14+ and Android config
```
android {
   externalNativeBuild {
       cmake {
           version "3.14+"
       }
   }
```
2. Update cmake only for iOS 
And add apple/ios specific error
```
If(APPLE)
Iif(${CMAKE_VERSION} VERSION_LESS "3.14") 
    message(FATAL_ERROR "Please use  CMake 3.14 or newer")
endif()
endif()
```

I do not have here strong opinion what is better but... Because Android development is hard and for now cmake 3.6 is used by default in Android studio I think that it is safer to use cmake from Android SDK (option 2)

minggo: Sounds reasonable.

crazyhappygame:
@trojanfoe based on your changes I created 
https://github.com/cocos2d/cocos2d-x/pull/20058
I think that this PR can be closed.

PR #20042, opened 2019-08-16
update cmake to v3.14 for travis

minggo:


crazyhappygame:
1. Could you explain why this is needed?
2. Why 3.14 why not latest 3.15.2
3. Please update also appveyor to the same version

minggo:
Because of this PR https://github.com/cocos2d/cocos2d-x/pull/20040.

crazyhappygame:
I propose to use package manager to install latest version on mac/windows
Windows:
```choco install cmake --installargs 'ADD_CMAKE_TO_PATH=System'```
Mac
```brew install cmake```

minggo:
I am not sure if CI can do it like this as macos is simulated by linux. And we need to update cmake with specific version.

minggo:
https://www.appveyor.com/updates/ says cmake is updated to 3.15.1, so i will not update cmake here.

Issue #16561, opened 2016-09-11
Mac ,FPS drop from 60 to 30 on FullScreen .


Kotaiba:
- cocos2d-x version: 3.11.1
- devices test on: Macbook Retina 15 , MAC OSX
- developing environments
  - Xcode version: Version 7.3.1 

This Issue only on  `glview = GLViewImpl::createWithFullScreen("DCR");`
if i use this the game will run 60 fps `glview = GLViewImpl::createWithRect("DCR", Rect(0, 0, 1136 ,640));`
On some mac devices the fps will drop from 60 to 30 . and go back to 60 again 
the game drop even on the  loading that have only 1 sprite the logo.
please see the attached video . it will show the issue ,
[video-1473507408.mp4.zip](https://github.com/cocos2d/cocos2d-x/files/465879/video-1473507408.mp4.zip)

slackmoehrle:
I watched the video. Interesting. But can you clarify a bit? Which line are you using that cases the drop? Also, by Mac devices do you you mean iphones and ipads or desktop OSX?

Kotaiba:
Only on desktop OSX , the drop happen only when the game is on FullScreen 
`glview = GLViewImpl::createWithFullScreen("DCR");`

dumganhar:
If the issue only appears on macOS desktop, it's probably an issue of GLFW in fullscreen.

Kotaiba:
@dumganhar  let me do more test on windows .

dumganhar:
@Kotaiba, that will be better for us to check the issue. Thanks. :)

Kotaiba:
@dumganhar Same issue on windows but it drop to 20-30-60fps only on FullScreen.

dumganhar:
@Kotaiba, interesting. I will try to reproduce it.

dumganhar:
@Kotaiba , I tried cpp-empty-test on my MacBook Pro (Retina, 15-inch, Early 2013). And I could not reproduce it. Could you test cpp-empty-test?

Kotaiba:
@dumganhar  cpp-empty-test  worked fine . but try to run cpp-tests on fullscreen and open basic-shaders it sometimes happen see the attached image . i am using MacBook Pro (Retina, 15-inch, Mid 2014)
[...]

PR #20043, opened 2019-08-16
fix the resolution issues when create full screen app

Mee-gu:
fix the resolution issues when creating a full-screen app. 
the issue is discussed [here](https://github.com/cocos2d/cocos2d-x/pull/17309)

### DIFF
--- a/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
+++ b/cocos/platform/desktop/CCGLViewImpl-desktop.cpp
@@ -345,7 +345,7 @@ bool GLViewImpl::initWithRect(const std::string& viewName, Rect rect, float fram
     glfwSetWindowSizeCallback(_mainWindow, GLFWEventHandler::onGLFWWindowSizeFunCallback);
     glfwSetWindowIconifyCallback(_mainWindow, GLFWEventHandler::onGLFWWindowIconifyCallback);
     glfwSetWindowFocusCallback(_mainWindow, GLFWEventHandler::onGLFWWindowFocusCallback);
-
+    glfwWaitEvents();
     setFrameSize(rect.size.width, rect.size.height);
 
     // check OpenGL version at first


minggo:
glfw can register callback for resolution changing, i think it is more reasonable to do some operation there.

Mee-gu:
`glfwWaitEvents` behaves exactly like `glfwPollEvents`, `glfwWaitEvents` is a better choice here since it only needs to update the contents of the window.
Besides, `glfwWaitEvents` as well as `glfwPollEvents` can't be called from a glfw callback.

Bilalmirza:
@Mee-gu  @minggo 
Just few issues I want to point out.
On macbook 2018 15"

1024x640: borders on left and right. cant move the mouse to screen ends either.
fps: 2156

1280x800: works fine
fps: 536

1440x900: works fine
fps: 391


After this it gets weird especially with the fps

1680x1050: works fine... i think the display also loses sharpness you can see some on fonts.
fps: ~1200 

1920x1280: works fine
fps: 740

Screenshots: https://www.dropbox.com/sh/4mqxarh3bjj1usx/AADjhHhDqflSqdKvW3fI18AEa?dl=0

Mee-gu:
@Bilalmirza As for the black borders on both left and right, I try lots of methods but change nothing. So I create an [issue](https://github.com/glfw/glfw/issues/1551) in glfw GitHub hoping for a good workaround.
BTW, I can't view your screenshots, maybe you can directly paste in the edit box.

Bilalmirza:
@mee-gu any idea why fps is so much better in higher resolutions. probably because of retina being disabled in last 2 i think.

Mee-gu:
@Bilalmirza How did you  get the fps?

Bilalmirza:
@Mee-gu    Appdelegate
director->setDisplayStats(true);

    // set FPS. the default value is 1.0/60 if you don't call this
    director->setAnimationInterval(1.0f / 3500);
    glfwSwapInterval(0);

Bilalmirza:
I think this is the issue
https://github.com/cocos2d/cocos2d-x/issues/16561

Bilalmirza:
I have a terrible fix for the fps. It also links the resolution to Display resolution.
You probably wont use this but here it goes.
in info.plist
add "High Resolution Capable" and set it to "NO"
in ccApplication
```
    NSWindow *nsWindow = glview->getCocoaWindow();
    [nsWindow.contentView setWantsBestResolutionOpenGLSurface:NO];

```

in CCGLViewImpl-desktop.cpp
```
    glfwWaitEvents();

//    setFrameSize(rect.size.width , rect.size.height);
    GLView::setFrameSize(rect.size.width , rect.size.height);
```
this prevents the call on updateFrameSize() // which really messes up the resolution.

now everything is fixed. The higher the display resolution in system preference the better the resolution in game.

PR #20058, opened 2019-08-22
remove_ios.toolchain.cmake

crazyhappygame:


crazyhappygame:
@minggo could you merge it?

tools/travis-scripts/before-install.sh
```diff
@@ -41,22 +41,6 @@ function install_python_module_for_osx()
...
-    pyenv global 2.7.14
-    python -V
-}
-
-function install_cmake_osx()
-{
-    cmake --version
-    brew upgrade cmake
-    cmake --version
-}
```
minggo, line 58: Why don't have to update cmake now?

tools/travis-scripts/before-install.sh
```diff
@@ -41,22 +41,6 @@ function install_python_module_for_osx()
...
 }
 
-function install_latest_python()
-{
-    python -V
-    eval "$(pyenv init -)"
-    pyenv install 2.7.14
-    pyenv global 2.7.14
-    python -V
-}
```
minggo, line 51: Can we make sure python is 2.x, not 3.x as all scripts don't support 3.x.

tools/travis-scripts/before-install.sh
crazyhappygame: Images xcode10.3, xcode11 have cmake 3.15, 3.15.1. Image Xcode9.3 has cmake 3.11

tools/travis-scripts/before-install.sh
crazyhappygame: Please check build output. 
python -V returns 2.7.16 for images xcode10.3, xcode11.
This code was needed because image Xcode9.3 has old version of python 2.7.10 with old networking libraries - download_deps failed

tools/travis-scripts/before-install.sh
minggo: Got it, so cmake is automatically updated when updating Xcode images? Right?

tools/travis-scripts/before-install.sh
minggo: But in future, python may be updated to v3, isn't it? I think it is better to check python version to determine if need to install python 2.7?

tools/travis-scripts/before-install.sh
crazyhappygame: Yes

tools/travis-scripts/before-install.sh
crazyhappygame: On Linux/mac by default:
 ./python == 2.*  
./python3 == 3.*

Removed code was not for checking but for installing latest python 2. It is not needed any with new Xcode images

Issue #20044, opened 2019-08-16
V4 - remove tiff file format


crazyhappygame:
- cocos2d-x version: v4
Why:
1. No one use tiff for game resources
2. Tiff source code is not supported
3. It is very hard to build tiff
4. It increases cocos2d-x binary size

crazyhappygame:
@minggo could you remove tiff for v4?

minggo:
I think it is reasonable.

minggo:
> No one use tiff for game resources

I am not sure of it.

minggo:
The [topic](https://discuss.cocos2d-x.org/t/why-does-cocos2d-x-support-tiff-files/13569) shows that tiff supporting is needed by .plist file, but i don't find any .plist uses tiff in cpp-tests.

crazyhappygame:
I see date 2012-05-16. Maybe 7 years ago that was true... 
But I think that now no one use tiff format.
Maybe we can ask about that on forum?

minggo:
I don't remember any developer report issues related with tiff, it seems nobody uses it, so i think it is ok to remove it.

minggo:
fixed in https://github.com/cocos2d/cocos2d-x/pull/20047

crazyhappygame:
Thank you!

Mee-gu:
@crazyhappygame I think TIFF can't be removed because it will be used in Particles since its texture image data was specified in the corresponding plist file.

minggo:
So it seems we can not remove TIFF, as particles may use the png data in .plist files?

minggo:
We can change to external png file without using the raw data in .plist file, but it will break compatibility?

Mee-gu:
> So it seems we can not remove TIFF, as particles may use the png data in .plist files?

Yes.



> We can change to external png file without using the raw data in .plist file, but it will break compatibility?

May need some test?

crazyhappygame:
@Mee-gu I am sorry could you explain it?
TIFF image file format can be used for game resources and probably also in particles. 
But my point is that one use TIFF as image format now. Mostly png, jpeg is used

Mee-gu:
@crazyhappygame you are right, from the search result of TIFF on the [forum](https://discuss.cocos2d-x.org/search?q=Tiff) seems TIFF rarely been used. Maybe we can change to such as png instead.

minggo:
[...]

PR #20063, opened 2019-08-29
revert TIFF

Mee-gu:
revert TIFF. ref [20044](https://github.com/cocos2d/cocos2d-x/issues/20044).


crazyhappygame:
Why this is needed? I thought that we agree that tiff is not used and not needed https://github.com/cocos2d/cocos2d-x/issues/20044#issuecomment-524252423

PR #20100, opened 2019-09-04
Feature/fix alpha test

Mee-gu:
[Clipping-HoleDemo] fix alpha test


cocos/2d/CCClippingNode.cpp
```diff
@@ -343,5 +350,26 @@ void ClippingNode::setInverted(bool inverted)
     _stencilStateManager->setInverted(inverted);
 }
 
+void ClippingNode::setProgram(Node* node, backend::ProgramState* programState)
```
minggo: From the meaning, it is better to use `setProgramStateRecursive`

PR #20115, opened 2019-09-11
Avoid Using "|" with Previously Declared Enum Values

JasonBaby87:
Unfortunately, without using only literals when declaring enum values, the following error will occur when I compiled with MSVC 2019 v142:

"expression did not evaluate to a constant"

`Node::FLAGS_DIRTY_MASK`, `GL::VERTEX_ATTRIB_FLAG_POS_COLOR_TEX` and `TMXTileFlags::kTMXFlipedAll` are edited.

I do not think this is the most elegant solution but at least this let's my project compile.

I have kept the original assignment lines in comments for reference.

If there are other ways to circumvent this, please let me know. I would much prefer to do it without changing this part.

### DIFF
--- a/cocos/2d/CCNode.h
+++ b/cocos/2d/CCNode.h
@@ -117,7 +117,8 @@ public:
         FLAGS_CONTENT_SIZE_DIRTY = (1 << 1),
         FLAGS_RENDER_AS_3D = (1 << 3),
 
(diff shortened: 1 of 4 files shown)


JasonBaby87:
Hi! I figured out what happened. It turns out my operator| overload messed up things, so commenting out fixes the issue.

Sorry for bothering you guys D:

PR #20127, opened 2019-09-17
getting correctly uniform location value

Mee-gu:
getting correctly uniform location value


minggo:
I remembered it is modified by @PatriceJiang in https://github.com/cocos2d/cocos2d-x/pull/20083.

PatriceJiang:
My recent PR #20083 was a bad commit.  
https://github.com/cocos2d/cocos2d-x/blob/f3b09563f88e750a143a1beb42f4d6d090761c7c/cocos/scripting/lua-bindings/manual/cocos2d/lua_cocos2dx_manual.cpp#L7403-L7406

The second argument should not pass a relative position `-2`, which will expire after any operation in the Lua stack.  I think it is better to fix that too. @Mee-gu

Mee-gu:
yep, the second argument value should be the right position for the uniform location. But how does it expire? @PatriceJiang

PatriceJiang (approved):

PR #20141, opened 2019-09-20
Remove tinyxml2 from CCSAXParser implement.

halx99:


minggo:
Why remove it?

halx99:
> Why remove it?

Since we use rapidxml, so doesn't need tinyxml2.

Issue #20142, opened 2019-09-20
RenderTexture internal Texture2D has incorrect premultiplied alpha flag


rh101:
- cocos2d-x version: 3.17.2

There is a problem with premultiplied alpha when using a RenderTexture, since it creates a Texture2D internally.  The problem with this is that the instance of Texture2D is created with `_hasPremultipliedAlpha` set to false, so any call to `RenderTexture->getSprite()->getTexture()->hasPremultipliedAlpha()` will always return false.

The sprite inside the RenderTexture is created with the correct blending mode, as you can see here:

```
        // retained
        setSprite(Sprite::createWithTexture(_texture));
        ...
        _sprite->setBlendFunc( BlendFunc::ALPHA_PREMULTIPLIED );
        ...
```

So, while the sprite has the blending set to `BlendFunc::ALPHA_PREMULTIPLIED`, the internal Texture2D still has premultiplied alpha set to false, and that on it's own is conflicting information.  If we store references to the Texture2D in parts of our code, and discard the sprite, then there is no way to know that it's actually premultiplied alpha (since `Texture2D::hasPremultipliedAlpha()` is false).

There is no way to set the Texture2D::_hasPremultipliedAlpha variable, but the problem actually lies with these 2 methods:

```  
bool initWithData(const void *data, ssize_t dataLen, Texture2D::PixelFormat pixelFormat, int pixelsWide, int pixelsHigh, const Size& contentSize);

bool initWithMipmaps(MipmapInfo* mipmaps, int mipmapsNum, Texture2D::PixelFormat pixelFormat, int pixelsWide, int pixelsHigh);
```

When using these 2 methods, as would be the case when using RenderTexture, there is no way to indicate that the texture should have the _hasPremultipliedAlpha set to true. So, in order to fix this issue, there needs to be several changes:

CCTexture2D.h
```
bool initWithData(const void *data, ssize_t dataLen, Texture2D::PixelFormat pixelFormat, int pixelsWide, int pixelsHigh, const Size& contentSize, bool preMultipliedAlpha = false);

bool initWithMipmaps(MipmapInfo* mipmaps, int mipmapsNum, Texture2D::PixelFormat pixelFormat, int pixelsWide, int pixelsHigh, bool preMultipliedAlpha = false);
```

The `preMultipliedAlpha` parameter can be default false, so it will not change current code using these methods.

CCTexture2D.cpp:

```
[...]

PR #20143, opened 2019-09-20
Fix RenderTexture PMA setting

rh101:
This will set the PMA flag of the Texture2D instance created by RenderTexture to the correct value, as described in this issue #20142


rh101:
If this PR is accepted and merged with V3, then it should be compatible with V4 too since the changes are the same.

PR #20146, opened 2019-09-20
fix warning

Mee-gu:
add integer type for Vec2 constructor and fix warning relates to Vec2.

### DIFF
--- a/cocos/2d/CCActionTiledGrid.cpp
+++ b/cocos/2d/CCActionTiledGrid.cpp
@@ -623,7 +623,7 @@ void TurnOffTiles::update(float time)
     for (unsigned int i = 0; i < _tilesCount; i++ )
     {
         t = _tilesOrder[i];
-        Vec2 tilePos( (unsigned int)(t / _gridSize.height), t % (unsigned int)_gridSize.height );
+        Vec2 tilePos( (int)(t / _gridSize.height), t % (int)_gridSize.height );
 
         if ( i < l )
         {
@@ -795,9 +795,9 @@ void SplitRows::startWithTarget(Node *target)
 
 void SplitRows::update(float time)
 {
-    for (unsigned int j = 0; j < _gridSize.height; ++j)
+    for (int j = 0; j < _gridSize.height; ++j)
     {
-        Quad3 coords = getOriginalTile(Vec2(0, j));
+        Quad3 coords = getOriginalTile(Vec2(0.0f, (float)j));
         float    direction = 1;
 
         if ( (j % 2 ) == 0 )
@@ -810,7 +810,7 @@ void SplitRows::update(float time)
         coords.tl.x += direction * _winSize.width * time;
         coords.tr.x += direction * _winSize.width * time;
 
-        setTile(Vec2(0, j), coords);
+        setTile(Vec2(0.0f, (float)j), coords);
     }
 }
 
@@ -852,7 +852,7 @@ void SplitCols::update(float time)
 {
     for (unsigned int i = 0; i < _gridSize.width; ++i)
     {
-        Quad3 coords = getOriginalTile(Vec2(i, 0));
+        Quad3 coords = getOriginalTile(Vec2((float)i, 0.0f));
         float    direction = 1;
 
         if ( (i % 2 ) == 0 )
@@ -865,7 +865,7 @@ void SplitCols::update(float time)
         coords.tl.y += direction * _winSize.height * time;
         coords.tr.y += direction * _winSize.height * time;
 
-        setTile(Vec2(i, 0), coords);
+        setTile(Vec2((float)i, 0.0f), coords);
     }
 }
 
--- a/cocos/2d/CCActionTiledGrid.cpp
+++ b/cocos/2d/CCActionTiledGrid.cpp
@@ -623,7 +623,7 @@ void TurnOffTiles::update(float time)
     for (unsigned int i = 0; i < _tilesCount; i++ )
     {
         t = _tilesOrder[i];
-        Vec2 tilePos( (int)(t / _gridSize.height), t % (int)_gridSize.height );
+        Vec2 tilePos((t /_gridSize.height), (float)(t % (int)_gridSize.height));
 
         if ( i < l )
         {
--- a/cocos/2d/CCAutoPolygon.cpp
+++ b/cocos/2d/CCAutoPolygon.cpp
@@ -249,13 +249,13 @@ unsigned int AutoPolygon::getSquareValue(unsigned int x, unsigned int y, const R
     //NOTE: due to the way we pick points from texture, rect needs to be smaller, otherwise it goes outside 1 pixel
     auto fixedRect = Rect(rect.origin, rect.size-Size(2,2));
     
-    Vec2 tl = Vec2((float)x-1, (float)y-1);
+    Vec2 tl((float)x-1, (float)y-1);
     sv += (fixedRect.containsPoint(tl) && getAlphaByPos(tl) > threshold)? 1 : 0;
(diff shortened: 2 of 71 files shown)


cocos/2d/CCActionTiledGrid.cpp
```diff
@@ -623,7 +623,7 @@ void TurnOffTiles::update(float time)
     for (unsigned int i = 0; i < _tilesCount; i++ )
     {
         t = _tilesOrder[i];
-        Vec2 tilePos( (unsigned int)(t / _gridSize.height), t % (unsigned int)_gridSize.height );
+        Vec2 tilePos( (int)(t / _gridSize.height), t % (int)_gridSize.height );
```
minggo: It is better to convert to float.

cocos/2d/CCAutoPolygon.cpp
```diff
@@ -249,13 +249,13 @@ unsigned int AutoPolygon::getSquareValue(unsigned int x, unsigned int y, const R
...
+    Vec2 tl = Vec2((float)x-1, (float)y-1);
     sv += (fixedRect.containsPoint(tl) && getAlphaByPos(tl) > threshold)? 1 : 0;
-    Vec2 tr = Vec2(x, y-1);
+    Vec2 tr = Vec2((float)x, (float)y-1);
     sv += (fixedRect.containsPoint(tr) && getAlphaByPos(tr) > threshold)? 2 : 0;
-    Vec2 bl = Vec2(x-1, y);
+    Vec2 bl = Vec2((float)x-1, (float)y);
     sv += (fixedRect.containsPoint(bl) && getAlphaByPos(bl) > threshold)? 4 : 0;
-    Vec2 br = Vec2(x, y);
+    Vec2 br = Vec2((float)x, (float)y);
```
minggo: These converts are not needed, as Vec2 has corresponding constructors.

minggo: Using `Vec2 tl(x-1, y-1);` is better.

cocos/scripting/lua-bindings/manual/cocos2d/lua_cocos2dx_manual.cpp
```diff
@@ -2403,7 +2403,7 @@ int tolua_cocos2d_Node_setAnchorPoint(lua_State* tolua_S)
         if (!ok)
             return 0;
 
-        cobj->setAnchorPoint(cocos2d::Vec2(x,y));
+        cobj->setAnchorPoint(cocos2d::Vec2((float)x,(float)y));
```
minggo, line 2406: This convert is not needed. The reason is the same as above.

cocos/math/Vec2.h
```diff
@@ -77,6 +77,7 @@ class CC_DLL Vec2
      * @param yy The y coordinate.
      */
     Vec2(float xx, float yy);
+    Vec2(int xx, int yy) : Vec2((float)xx, (float)yy) {}
```
crazyhappygame: Do not like it ... 
What about 
```
Vec2(int xx, float yy)
Vec2(floay xx, int yy)
....
Vec2(AnyOtherNumber xx, AnyOtherNumber yy)
```

I think that better is to fix all places

crazyhappygame:
I think together with removing warnings compiler options should be enabled to catch all possible regression. Below how I did in my CMake

```
if(MSVC)
    target_compile_options(${PROJECT_NAME}
        PRIVATE
       /WX 
       /W0   # /W1 /W2 /W3 /W4 /Wall
       # /permissive-
      # /sdl
)
else()
  target_compile_options(${PROJECT_NAME}
                         PRIVATE 
                                 -Werror
                                 -Wall
                                 # -pedantic
                                 # -pedantic-errors
                                 # -Wextra
                                 # -Wconversion
                                 # -Wsign-conversion
                                 )
```

crazyhappygame:
I really like idea of removing warning but please consider following:
1. Please for each fixed warning enable compiler check. 
http://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
2. I think that good start is "-Wall -Werror". int/float conversion are tricky please leave it when all others warning will be fixed
3. Please do not use C style casts. use C++ static_cast<float>(intValue)
4. Please do not use casts everywhere consider change types if possible
instead of
```
int value1 = 2;
float value2 = float(value1);
```
maybe following is possible
```
float value1 = 2;
float value2 = value1;
```

crazyhappygame (changes requested):

crazyhappygame:
@minggo @Mee-gu Please discard this PR.
Please start from "-Wall -Werror" compilers warnings

cocos/math/Vec2.h
Mee-gu: I think this way is not well designed since the data type may be int, unsigned int, float, double, etc.

Mee-gu:
> @minggo @Mee-gu Please discard this PR.
> Please start from "-Wall -Werror" compilers warnings

@crazyhappygame  we will fix some basic warnings firstly, then make the warnings into error.

cocos/math/Vec2.h
crazyhappygame: Below constructor will accept all types with conversion to float
```
template<class T1, class T2>
Vec2(T1 xx, T2 yy)
: Vec2((float)xx, (float)yy) {}
{
}
```
But still I think that that this class should have Vec2(float xx, float yy); only

Mee-gu: yep, we can use the template but the matter is we need always specify the type in angle brackets when using it. I agree that have `Vec2(float xx, float yy)` only is reasonable.

Mee-gu:
> 3\. Please do not use C style casts. use C++ static_cast(intValue)

Is there any difference between C style cast and C++ static_cast<>() for basic types? @crazyhappygame

cocos/scripting/lua-bindings/manual/cocos2d/lua_cocos2dx_manual.cpp
Mee-gu: `x` and `y` are double, need convert to `float`.

cocos/math/Vec2.h
crazyhappygame: Thank you for removing int construction

crazyhappygame:
Please read e.g. here about static_cast
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-cstyle-cast.html

crazyhappygame:
> @crazyhappygame we will fix some basic warnings firstly, then make the warnings  into error.

Usually it is done like that:
1. Enable all warning as error
2. Enable warning 1 by 1

minggo:
@crazyhappygame if we we enable wall warnings as error, then it will block the travis and development. Treat all warnings as error is the final status, that's why we want to fix warnings.

crazyhappygame:
@minggo you can enable "warning as error" and turn on only fixed warnings.

minggo:
Good to know. 
@Mee-gu i think we can do as @crazyhappygame said. Turn on some warnings as error, then fix it.

minggo:
> Usually it is done like that:

> Enable all warning as error
> Enable warning 1 by 1

I think can nod do like this, we can do like:
- 1. turn on some type of warning as error
- 2. fix this error
- 3. loop 1 and 2 until all warnings are fixed

after all, we turn on all warnings as error

minggo:
> Please read e.g. here about static_cast
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-cstyle-cast.html

@crazyhappygame as i know, c cast and static_cast have the same effect to basic types.

PR #20152, opened 2019-09-23
set software mode will slow down performance

minggo:


drelaptop:
This PR also solved webgl samples can't run on WebView.

> cases like: https://webglsamples.org/

PR #20154, opened 2019-09-23
Sprite and RenderTexture blending mode set based off texture premultiplied alpha setting

rh101:
The blending mode of a sprite should be set based on the premultiplied alpha setting of the passed in texture when using Sprite:initWithTexture().

This also applies to RenderTexture, where the sprite opacityModifyRGB should be set based off the PMA of the texture.


cocos/2d/CCSprite.cpp
```diff
@@ -269,9 +269,16 @@ bool Sprite::initWithTexture(Texture2D *texture, const Rect& rect, bool rotated)
...
+        if (!texture || !texture->hasPremultipliedAlpha())
+        {
+            _opacityModifyRGB = false;
+            _blendFunc = BlendFunc::ALPHA_NON_PREMULTIPLIED;
+        }
+        else
+        {
+            _opacityModifyRGB = true;
+            _blendFunc = BlendFunc::ALPHA_PREMULTIPLIED;
+        }
```
minggo: `updateBlendFunc()` has the same logic.

rh101: Updated both this PR and the PR for V4 to remove that redundant code.

minggo (approved):

PR #20157, opened 2019-09-23
Make luabindings framework compat with lua52 & lua53.

halx99:


cocos/scripting/lua-bindings/manual/CCLuaEngine.cpp
```diff
@@ -37,8 +37,6 @@
 #include "base/CCDirector.h"
 #include "base/CCEventCustom.h"
 
-#pragma comment(lib,"lua51.lib")
```
minggo, line 40: As i remember, it is used to link lua lib on windows.

halx99: Yes, remove from sources, let user can specific which version of lua lib he want to link.

minggo: Where to choose? The travis is blocked.

halx99: lualib need add to the final lua test executable projects, I'll do later.

minggo: And please modify the lua template too. Thanks.

halx99: ok, I'll.

PR #20167, opened 2019-09-24
V3

iams23:
when a scrollView added into another and scroll out, the inside scrollView's container will be show in view, like this. ↓↓↓↓↓↓↓↓↓↓↓↓↓
scrollView


byWindy:
> when a scrollView added into another and scroll out, the inside scrollView's container will be show in view, like this. ↓↓↓↓↓↓↓↓↓↓↓↓↓
> scrollView



> when a scrollView added into another and scroll out, the inside scrollView's container will be show in view, like this. ↓↓↓↓↓↓↓↓↓↓↓↓↓
> scrollView

PR #20173, opened 2019-09-25
fix C4244 warning

Mee-gu:
fix C4244 warning

### DIFF
--- a/cocos/2d/CCFastTMXLayer.cpp
+++ b/cocos/2d/CCFastTMXLayer.cpp
@@ -194,8 +194,8 @@ void TMXLayer::updateTiles(const Rect& culledRect)
     float tileSizeMax = std::max(tileSize.width, tileSize.height);
     if (_layerOrientation == FAST_TMX_ORIENTATION_ORTHO)
     {
-        tilesOverX = ceil(tileSizeMax / mapTileSize.width) - 1;
-        tilesOverY = ceil(tileSizeMax / mapTileSize.height) - 1;
+        tilesOverX = (int)ceil(tileSizeMax / mapTileSize.width) - 1;
+        tilesOverY = (int)ceil(tileSizeMax / mapTileSize.height) - 1;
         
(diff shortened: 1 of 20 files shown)


cocos/2d/CCFastTMXLayer.cpp
```diff
@@ -501,11 +501,12 @@ void TMXLayer::updateTotalQuads()
                 
                 float left, right, top, bottom, z;
                 
-                z = getVertexZForPos(Vec2(x, y));
-                auto iter = _indicesVertexZOffsets.find(z);
+                int zPos = getVertexZForPos(Vec2((float)x, (float)y));
+                z = (float)zPos;
```
minggo, line 505: it seems z is not used

Mee-gu: it used here 
https://github.com/cocos2d/cocos2d-x/blob/a500e7f0ea10c0d485269162022c15ab9cde9a4f/cocos/2d/CCFastTMXLayer.cpp#L527

cocos/2d/CCFontFreeType.cpp
```diff
@@ -360,16 +360,17 @@ unsigned char* FontFreeType::getGlyphBitmap(uint64_t theChar, long &outWidth, lo
...
+                memset(blendImage, 0, imageSize);
 
                 auto px = outlineMinX - blendImageMinX;
                 auto py = blendImageMaxY - outlineMaxY;
                 for (int x = 0; x < outlineWidth; ++x)
                 {
                     for (int y = 0; y < outlineHeight; ++y)
                     {
-                        index = px + x + ((py + y) * blendWidth);
+                        index = (long)(px + x + ((py + y) * blendWidth));
```
minggo: does index need to be long?

Mee-gu: the integer type is enough, i'll fix it later.

minggo (approved):

PR #20185, opened 2019-10-09
fix triangle batch

Mee-gu:
since it would be too expensive to check the uniforms, simplify enable batching for built-in programs.

Corresponding issue post [here](https://discuss.cocos2d-x.org/t/metal-support-rc0-released/47797/3).


minggo (approved):

minggo:
Should provide a way that developers can batch themselves.

PR #20191, opened 2019-10-11
fix C4244 warning

Mee-gu:
fix C4244 warning

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -160,15 +160,16 @@ public:
         auto displayedOpacity = _displayedOpacity;
         if(!_letterVisible)
         {
-            displayedOpacity = 0.0f;
+            displayedOpacity = 0;
         }
         Color4B color4(_displayedColor.r, _displayedColor.g, _displayedColor.b, displayedOpacity);
         // special opacity for premultiplied textures
         if (_opacityModifyRGB)
         {
-            color4.r *= displayedOpacity / 255.0f;
-            color4.g *= displayedOpacity / 255.0f;
-            color4.b *= displayedOpacity / 255.0f;
+            auto opacity = displayedOpacity / 255.0f;
+            color4.r = static_cast<uint8_t>(color4.r * opacity);
+            color4.g = static_cast<uint8_t>(color4.g * opacity);
+            color4.b = static_cast<uint8_t>(color4.b * opacity);
         }
         _quad.bl.colors = color4;
         _quad.br.colors = color4;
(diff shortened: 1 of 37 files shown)


cocos/2d/CCLabel.cpp
```diff
@@ -265,7 +266,7 @@ Label* Label::createWithTTF(const std::string& text, const std::string& fontFile
     return nullptr;
 }
 
-Label* Label::createWithTTF(const TTFConfig& ttfConfig, const std::string& text, TextHAlignment hAlignment /* = TextHAlignment::CENTER */, int maxLineWidth /* = 0 */)
+Label* Label::createWithTTF(const TTFConfig& ttfConfig, const std::string& text, TextHAlignment hAlignment /* = TextHAlignment::CENTER */, float maxLineWidth /* = 0 */)
```
minggo, line 269: Is it reasonable to use float for label width?

cocos/editor-support/cocostudio/CCActionObject.cpp
```diff
@@ -192,7 +192,7 @@ bool ActionObject::valueToBool(const std::string& value)
 }
 float ActionObject::valueToFloat(const std::string& value)
 {
-    return utils::atof(value.c_str());
+    return (float)utils::atof(value.c_str());
```
minggo, line 195: Why need to convert `atof` to float? Isn't `atof` return float?

cocos/editor-support/cocostudio/CCArmatureAnimation.cpp
```diff
@@ -232,7 +232,7 @@ void ArmatureAnimation::play(const std::string& animationName, int durationTo,
         if(movementBoneData && movementBoneData->frameList.size() > 0)
         {
             _tweenList.push_back(tween);
-            movementBoneData->duration = _movementData->duration;
+            movementBoneData->duration = (float)_movementData->duration;
```
minggo, line 235: Why duration have different types?

extensions/assets-manager/AssetsManagerEx.cpp
```diff
@@ -77,7 +77,7 @@ AssetsManagerEx::AssetsManagerEx(const std::string& manifestUrl, const std::stri
                                          int64_t totalBytesReceived,
                                          int64_t totalBytesExpected)
     {
-        this->onProgress(totalBytesExpected, totalBytesReceived, task.requestURL, task.identifier);
+        this->onProgress((double)totalBytesExpected, (double)totalBytesReceived, task.requestURL, task.identifier);
```
minggo, line 80: I think it is more reasonable to modify `onProgress`. Why `onProgress` need to receive double parameters?

cocos/editor-support/cocostudio/CCActionObject.cpp
Mee-gu: the `atof` convert string to double.

extensions/assets-manager/AssetsManagerEx.cpp
Mee-gu: I'm afraid that the download size is too big to store in a float.

Mee-gu:
will split into separate PRs

PR #20196, opened 2019-10-12
[v3] add unzip test code

PatriceJiang:
unzip files with/without password


tests/cpp-tests/CMakeLists.txt
```diff
@@ -329,8 +330,7 @@ list(APPEND GAME_SOURCE
      Classes/ZwoptexTest/ZwoptexTest.cpp
      Classes/SpriteFrameCacheTest/SpriteFrameCacheTest.cpp
      Classes/controller.cpp
-     # Classes/VRTest/VRTest.cpp
-     # Classes/VRTest/VRTest.h
```
minggo, line 333: On v3, i think you should modify project files too.

PatriceJiang: ok

PR #20198, opened 2019-10-13
Don't use MTL_ pixelformat when building for iOS Simulator.

trojanfoe:
This simple fix allows metal to run under the iOS Simulator.

### DIFF
--- a/cocos/renderer/CCTexture2D.cpp
+++ b/cocos/renderer/CCTexture2D.cpp
@@ -406,7 +406,7 @@ bool Texture2D::initWithImage(Image *image, backend::PixelFormat format)
     //override renderFormat, since some render format is not supported by metal
     switch (renderFormat)
     {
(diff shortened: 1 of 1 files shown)


Mee-gu (approved):

cocos/renderer/CCTexture2D.cpp
```diff
@@ -406,7 +406,7 @@ bool Texture2D::initWithImage(Image *image, backend::PixelFormat format)
     //override renderFormat, since some render format is not supported by metal
     switch (renderFormat)
     {
-#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS)
+#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS && !TARGET_OS_SIMULATOR)
```
minggo, line 409: I think it would be better to add some logs.

PR #20199, opened 2019-10-14
PostMessage to SendMessage

KAndQ:
使用 PostMessage 无法在 windowProc 接收到消息, 导致无法在日志中输出打印信息.

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -114,7 +114,7 @@ namespace {
         if (Director::getInstance()->getOpenGLView())
         {
             HWND hwnd = Director::getInstance()->getOpenGLView()->getWin32Window();
-            PostMessage(hwnd,
+            SendMessage(hwnd,
                         WM_COPYDATA,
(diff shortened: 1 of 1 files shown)


PatriceJiang (approved):

minggo:
@KAndQ please use english in future.
@PatriceJiang as i remember, SendMessage will block the main thread, so we changed to use PostMessage. So i don't think it is a good idea to revert the modification.

PatriceJiang:
I misused `PostMessage` in a previous change.  According to the official document the referenced data should be considered read-only.  I should find another way to achieve non-block.

PatriceJiang:
@KAndQ Do you mean all messages are lost or some of them?

KAndQ:
Using multiple threads? 

I was sure that PostMessage was unusable, and that while SendMessage blocked, at least it worked.
The WM_COPYDATA message cannot be received in the windowProc function.

In: Win10, Visual Studio 2017.

minggo:
Ok, let's optimize it in future.

PR #20201, opened 2019-10-14
处理 windows 平台, 每行日志多个换行的问题

KAndQ:
统一所有平台的日志输出, 不会有多一行换行的问题出现.

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -166,7 +166,9 @@ void log(const char * format, ...)
             delete[] buf;
         }
     } while (true);
+#if CC_TARGET_PLATFORM != CC_PLATFORM_WIN32
     buf[nret] = '\n';
+#endif
     buf[++nret] = '\0';
 
 #if CC_TARGET_PLATFORM == CC_PLATFORM_ANDROID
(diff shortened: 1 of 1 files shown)


minggo:
@KAndQ thanks for your contribution, but please use english in future. Thanks.

kangzixiang (approved):

rh101:
What is the reason behind doing this? Without line breaks, it makes it very hard to scan the logs, both visually and with automated tools, because it is literally attaching unrelated log output lines together.  

EDIT: There is a bug introduced in this PR.  The fix is in PR #20357

rh101:
@minggo There are 2 issues with the code:

In `void log(const char * format, ...)`:
```
#if CC_TARGET_PLATFORM != CC_PLATFORM_WIN32
    buf[nret] = '\n';
#endif
    buf[++nret] = '\0';
```

The `nret` should not be incremented, so the correct code would be:
```
#if CC_TARGET_PLATFORM != CC_PLATFORM_WIN32
    buf[nret] = '\n';
    buf[++nret] = '\0';
#else
    buf[nret] = '\0';
#endif
```

Also, the main cause of the Windows Visual Studio debugger output problems:
```
        MultiByteToWideChar(CP_UTF8, 0, tempBuf, -1, wszBuf, sizeof(wszBuf));
        OutputDebugStringW(wszBuf);
        WideCharToMultiByte(CP_ACP, 0, wszBuf, -1, tempBuf, sizeof(tempBuf), nullptr, FALSE);
```

Should be:
```
        MultiByteToWideChar(CP_UTF8, 0, tempBuf, -1, wszBuf, sizeof(wszBuf));
        OutputDebugStringW(wszBuf);
        OutputDebugStringW(L"\n");
        WideCharToMultiByte(CP_ACP, 0, wszBuf, -1, tempBuf, sizeof(tempBuf), nullptr, FALSE);
```

The reason for this is that the `\n` is no longer added to the line (in the code before it), so it is what's causing the bad output in the Visual Studio console window.

minggo:
@PatriceJiang please take a look.

PR #20205, opened 2019-10-15
Rework CMake Xcode support.

trojanfoe:
- Dropped the use of the cocos_pak_xcode() cmake function as it was not working correctly.
- Made it easier to specify deployment targets and other Xcode settings when generating the Xcode project.
- Add documentation to demonstrate how to use the new settings.


tests/cpp-tests/proj.mac/Info.plist
```diff
@@ -2,49 +2,33 @@
...
-	</dict>
+	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
 	<key>NSHumanReadableCopyright</key>
-	<string>Copyright © 2012. All rights reserved.</string>
+	<string>Copyright © 2019. All rights reserved.</string>
 	<key>NSMainNibFile</key>
 	<string>MainMenu</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
-	<key>UIAppFonts</key>
```
PatriceJiang: The font list can be kept  https://github.com/cocos2d/cocos2d-x/pull/19692/files. 
If fonts are not specified in Info.plist, iOS will use system font instead.

tests/cpp-tests/proj.mac/Info.plist
```diff
@@ -2,49 +2,33 @@
...
-	<string>????</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
+	<key>LSApplicationCategoryType</key>
+	<string>public.app-category.action-games</string>
 	<key>LSMinimumSystemVersion</key>
-	<string>${MACOSX_DEPLOYMENT_TARGET}</string>
-	<key>NSAppTransportSecurity</key>
-	<dict>
-		<key>NSAllowsArbitraryLoads</key>
```
PatriceJiang, line 31: For test purposes cpp-tests need cleartext HTTP transportation, We should keep this option.

tests/lua-tests/project/proj.ios_mac/mac/Info.plist
```diff
@@ -2,40 +2,33 @@
...
 	<key>NSHumanReadableCopyright</key>
-	<string>Copyright © 2012. All rights reserved.</string>
+	<string>Copyright © 2019. All rights reserved.</string>
 	<key>NSMainNibFile</key>
 	<string>MainMenu</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
-	<key>NSAppTransportSecurity</key>
-	<dict>
-		<key>NSAllowsArbitraryLoads</key>
```
PatriceJiang: Same as cpp-tests

tests/cpp-tests/proj.mac/Info.plist
trojanfoe: Yeah my initial approach was to take standard Mac versions of Info.plist and overwrite those used in the templates.  That was until I realised there were special settings in them and I not do this for the iOS versions.  I will revert these changes to the Mac Info.plist files in a follow-up commit.

tests/cpp-tests/proj.mac/Info.plist
trojanfoe: As above.

tests/lua-tests/project/proj.ios_mac/mac/Info.plist
trojanfoe: As above.

cmake/Modules/CocosConfigDefine.cmake
```diff
@@ -74,8 +74,9 @@ define_property(TARGET
 # check c++ standard
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_C_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
```
PatriceJiang: C++17 is not supported by some older compilers, C++14 is better supported.

trojanfoe: This was discussed on the [forum](https://discuss.cocos2d-x.org/t/v4-would-be-a-good-time-to-move-things-forward/47962) and the consensus was that things are moved forward for the v4 branch.  If a user moved to v4 they would need to update their toolchain to the minimum required (as yet undefined).  If they cannot update their toolchain then they can stay at v3.

minggo: We haven't tested C++14, not even C++17. Metal support is already a big change, i think it is better not add other things to make it not stable. We can update it after it is stable. And developers can change it if needed, it is not difficult.

trojanfoe: That seems to be at odds with our discussion on the forum where I thought you were in favour of going to C++14/17 and moving classes out of the `experimental` namespace?

minggo: Yep, i agree to move experimental, but i don't think update to C++14/17 is a good idea.

trojanfoe: That's fine.  I will update the commit this evening.

tests/lua-tests/project/proj.ios_mac/ios/Info.plist
```diff
@@ -22,17 +22,17 @@
...
-    <string>${MACOSX_BUNDLE_GUI_IDENTIFIER}</string>
+    <string>org.cocos2dx.$(PRODUCT_NAME)</string>
     <key>CFBundleInfoDictionaryVersion</key>
     <string>6.0</string>
     <key>CFBundleName</key>
-    <string>${MACOSX_BUNDLE_BUNDLE_NAME}</string>
+    <string>$(PRODUCT_NAME)</string>
     <key>CFBundlePackageType</key>
     <string>APPL</string>
     <key>CFBundleSignature</key>
```
Mee-gu, line 32: `CFBundleShortVersionString` missing.

trojanfoe: My change didn't remove it, however I will add it this evening.

Mee-gu: Thank you.

PR #20206, opened 2019-10-15
cmake_3_16 precompiled header

crazyhappygame:
This test of cmake 3.16 and target_precompile_headers. 
Do not commit.


minggo:
I can quite catch the idea of this PR.

PR #20225, opened 2019-10-21
fix text wrong wrap issue

Mee-gu:
fix word wrap not working correctly in iOS13.


minggo (approved):

minggo:
Please synchronize to v4 too.

PR #20261, opened 2019-10-28
Fix warning when call Node::addComponent twice

dzj0821:
If i call Node::addComponent twice, scheduleUpdate will be called twice, and then console output "warning: don't update it again"(CCScheduler.cpp line484). I think scheduleUpdate only should be called in the first time.


dzj0821:
If i call Node::removeAllComponents, (CCComponentContainer.cpp line117)_owner->unscheduleUpdate() will be called. I ignored it, sorry.

PR #20266, opened 2019-10-29
Fixes so that in the web view we are able to navigate different links properly

deepshah7:
After loading a page inside a WebView we were unable to navigate the links correctly because the URL used is always the webview's url. The correct URL should be picked-up from the [navigationAction request] URL].

### DIFF
--- a/cocos/ui/UIWebViewImpl-ios.mm
+++ b/cocos/ui/UIWebViewImpl-ios.mm
@@ -250,7 +250,7 @@
 
 #pragma mark - WKNavigationDelegate
 - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler {
-    NSString *url = [webView.URL absoluteString];
+    NSString *url = [[[navigationAction request] URL] absoluteString];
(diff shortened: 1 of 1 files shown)


minggo:
Sorry, i don't quite understand what you meaning. Could you please describe it in more detail? Thanks.

deepshah7:
Yes, sorry for not describing the issue in detail earlier.

Here's what's happening, 

- When we load the WebView for the first time with a web URL say `www.google.com`, it loads the page correctly and the web page appears.
- However, inside the web page there could be many links including some which start with `mailto:` or `tel:`
- On clicking such links the callback `- (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler` is invoked. 
- This in turn calls `self.shouldStartLoading([url UTF8String]` if its available.
- The URL passed as argument to this method is by default always having the value of the `webView.URL`, which will always be the original URL i.e. `www.google.com` in our example.
- This seems incorrect, it should pass on the URL of the link that was clicked by the user. That value is available in `[[navigationAction request] URL]`. It could be `mailto:` or `tel:` or something else. And thats the change I am suggesting.

minggo:
@deepshah7 got it, thanks for the explanation.

PR #20269, opened 2019-10-30
fix crash due to iOS13 UIWindowScene

Mee-gu:
UIWindowScene available on iOS13

### DIFF
--- a/cocos/platform/ios/CCDevice-ios.mm
+++ b/cocos/platform/ios/CCDevice-ios.mm
@@ -246,24 +246,34 @@ static CCAccelerometerDispatcher* s_pAccelerometerDispatcher;
     _acceleration->timestamp = accelerometerData.timestamp;
 
     double tmp = _acceleration->x;
-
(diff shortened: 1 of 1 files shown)


cocos/platform/ios/CCDevice-ios.mm
```diff
@@ -246,24 +246,34 @@ - (void)accelerometer:(CMAccelerometerData *)accelerometerData
     _acceleration->timestamp = accelerometerData.timestamp;
 
     double tmp = _acceleration->x;
-
-    switch ([[[UIApplication sharedApplication].windows[0] windowScene] interfaceOrientation])
+    UIInterfaceOrientation orientation;
+    if (@available(iOS 13.0, *))
```
minggo, line 250: How about iOS 13.1? Does this logic also match it?

minggo (approved):

cocos/platform/ios/CCDevice-ios.mm
Mee-gu: It matches iOS 13+ and the asterisk(*) indicated the avaliablility of the declaration on all of the platforms.

PR #20272, opened 2019-10-30
comment spine codes in lua

Mee-gu:


minggo:
Please add TODO that will enable it after spine is available.

PR #20276, opened 2019-10-31
avoid symbol conflict with other libraries of base64 functions

Naruto:
`_base64Decode` and `_base64Encode` are internal functions in `base/base64.cpp`.
Sometimes, those function symbol conflict with other libraries at linking.

This PR changes those functions name to more uniquness.

### DIFF
--- a/cocos/base/base64.h
+++ b/cocos/base/base64.h
@@ -29,10 +29,6 @@ THE SOFTWARE.
 /// @cond DO_NOT_SHOW
 
 #include "platform/CCPlatformMacros.h"
-
-#ifdef __cplusplus
-extern "C" {
-#endif    
     
 namespace cocos2d {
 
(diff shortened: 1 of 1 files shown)


minggo:
They are in cocos2d namespace, so it will not be conflict with others.

Naruto:
@minggo 
I've misunderstood. You are correct. 
Internal functions are in `cocos2d` namespace.

Problem is other place.
`cocos2d::base64Encode()` and `cocos2d::base64Decode()` are declared with `extern "C"` in `base/base64.h`.
`extern "C"` ignores namespace, so those functions symbol are `_base64Encode` and `_base64Decode`.

I would like to remove `_base64Encode` and `_base64Decode` symbols from libcocos2d.a. 
Could you have any good idea?

minggo:
I think can just remove `extern "C"`.

Naruto:
@minggo 
remove `extern "C"` from `base/base64.h`.
Please review it.

minggo (approved):

Naruto:
thanks!

PR #20285, opened 2019-11-01
Fixing a crash in iOS 9.3.5 when we have a layer with WebView getting deallocated

deepshah7:
- Have a layer with a web view on it
- Try to launch it in iOS 9.3.5
- When this layer exits and memory is deallocated
- The class `UIWebViewWrapper` tries to release the property `[self.wkWebView release];`
- When the property was initialised it wasn't retained.
- Because of this the code crashes giving an error like releasing an instance that was not allocated.
- This is observed only in iOS 9.3.5 and 9.3.6. In iOS 12+ it works fine thought. 
- However I feel it should have crashed in other versions of the iOS too.
- This commit tries to fix the crash.

### DIFF
--- a/cocos/ui/UIWebViewImpl-ios.mm
+++ b/cocos/ui/UIWebViewImpl-ios.mm
@@ -124,7 +124,7 @@
 
 - (void)setupWebView {
     if (!self.wkWebView) {
-        self.wkWebView = [[[WKWebView alloc] init] autorelease];
+        self.wkWebView = [[[[WKWebView alloc] init] autorelease] retain];
         self.wkWebView.UIDelegate = self;
         self.wkWebView.navigationDelegate = self;
     }
--- a/cocos/ui/UIWebViewImpl-ios.mm
+++ b/cocos/ui/UIWebViewImpl-ios.mm
@@ -88,7 +88,7 @@
 
 
 @interface UIWebViewWrapper () <WKUIDelegate, WKNavigationDelegate>
-@property(nonatomic, retain) WKWebView *wkWebView;
+@property(nonatomic) WKWebView *wkWebView;
 
 @property(nonatomic, copy) NSString *jsScheme;
 @end
--- a/cocos/ui/UIWebViewImpl-ios.mm
+++ b/cocos/ui/UIWebViewImpl-ios.mm
@@ -124,7 +124,7 @@
 
 - (void)setupWebView {
     if (!self.wkWebView) {
-        self.wkWebView = [[[[WKWebView alloc] init] autorelease] retain];
+        self.wkWebView = [[WKWebView alloc] init];
         self.wkWebView.UIDelegate = self;
         self.wkWebView.navigationDelegate = self;
     }


minggo:
But the property is declared as `retained`.

minggo:
@deepshah7 it may cause memory leak if retaining it.

deepshah7:
I did do memory profiling with that change and it didn't seem to leak the memory. But it does confuse me that even though the property is `retained` why code crashes on deallocating the web-view on iOS 9.3.5 and 9.3.6.

minggo:
It is strange that it doesn't crash on all iOS versions.

> I did do memory profiling with that change and it didn't seem to leak the memory.

Did you do it with iOS versions that will not cause crash?

deepshah7:
Yes I did profile the other iOS versions too and adding the retain didn't introduce the leak. I will do one more round tomorrow and report back.

minggo:
@deepshah7 any update?

deepshah7:
Yes I did profile it again on iOS 13 and the change doesn't cause a memory leak.

deepshah7:
Also I can confirm that it crashed on iOS 13.1.3 on iPhone XR and iOS 12.4.2 on iPad Air retina display  without the fix while debugging with XCode for the same reason. However, if the debugging is not active it doesn't crash on these devices.

deepshah7:
It crashes while debugging with this error `-[WKWebView release]: message sent to deallocated instance`

cocos/ui/UIWebViewImpl-ios.mm
```diff
@@ -124,7 +124,7 @@ - (void)dealloc {
 
 - (void)setupWebView {
     if (!self.wkWebView) {
-        self.wkWebView = [[[WKWebView alloc] init] autorelease];
+        self.wkWebView = [[[[WKWebView alloc] init] autorelease] retain];
```
minggo: Just change to `[[WKWebView alloc] init]` is enough.

deepshah7: Yes that should do it. I will check once and circle back.

deepshah7: It works! Tested it on the iOS 13.1.3. I will test it on iOS 9.3.5 in a couple of hours and report back (don't have the device with me right now).

Have added the new commit to the pull request. The end result is, removing the `autorelease` from the `WKWebView` instantiation.

deepshah7:
Would it be ok to merge this pull request? @minggo

minggo:
How to reproduce it as our QA can not reproduce it.

minggo:
And i tested on iPhone11(iOS 13.1.2) with debugging, not crash happened. The test case is `51:Node: UI -> 1: GUI Dynamic Create Test -> 2: WebView Test ` in `cpp-tests`.

What test case did you test?

deepshah7:
It crashes in my app, we have a scene with a simple webview in it and nothing else. 

On existing the scene it crashes while release the webview instance. I will try and create a sample application with a webview in it and send it across to easily reproduce the crash.

minggo:
@deepshah7 i think we can remove the usage of `retain` in property declaration and use retain/release manually. So could you please remove `retain` in property and test it again?

deepshah7:
I will try that out as well. Ideally it should work.

minggo:
I think so, then we can remove the dependence of the property setting.

deepshah7:
It works. Pushed the commit with the changes in this pull request. Also here is the sample project which reproduces the crash on iOS 9.3.5 device: https://github.com/deepshah7/webview-test

minggo:
Indeed i can not reproduce the issue, but we can remove the usage of `retain` in property.

minggo (approved):

PR #20297, opened 2019-11-06
fix #17245, using spine before restartVM will cause memory leak quickly

### LINKED ISSUE #17245: using both restart and spine will cause memory leak and system crash
- cocos2d-x version:this bug occur on cocos2d js 3.11 to cocos2d js 3.14
- devices test on:iphone6sp and simulator 6se
- developing environments
   - NDK version:
   - Xcode version:8.2
   - VS version:
   - browser type and version:

Steps to Reproduce:
unpack sdk 3.14
1. run js test project included in sdk packet
2. press Spine Test and then press Main Menu
3. press Sys Tests and find restartGame and press it
4. press Spine Test
5. memory will increase about 0.3M per second until system crash.

chongchaoyu:
When director is reset, SkeletonBatch and SkeletonTwoColorBatch instances will be destroyed, and they can be recreated when invoking SkeletonBatch::getInstance() or SkeletonTwoColorBatch::getInstance() next time.

### DIFF
--- a/cocos/editor-support/spine/SkeletonBatch.cpp
+++ b/cocos/editor-support/spine/SkeletonBatch.cpp
@@ -34,6 +34,7 @@
 
 USING_NS_CC;
 #define EVENT_AFTER_DRAW_RESET_POSITION "director_after_draw"
+#define EVENT_ON_DIRECTOR_RESET "director_reset"
 using std::max;
 #define INITIAL_SIZE (10000)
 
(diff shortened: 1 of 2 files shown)


chongchaoyu:
relevant to issue: using both restart and spine will cause memory leak and system crash #17245

minggo:
@chongchaoyu according spine runtime's license, you should send PR to spine's runtime library.

chongchaoyu:
@minggo ok, thanks. 
I think it may be unacceptable putting some codes relevant to spine into Director.cpp, yes? 
so putting some codes into jsb-binding to destroy SkeletonBatch and SkeletonTwoColorBatch instances? any advice?

minggo:
It is better if can do it in jsb-binding codes.

mori1234:
removeCustomEventListeners ----> removeEventListener

delete _indexBuffer;  ---> delete[] _indexBuffer;

_twoColorTintShader->release();  --> _twoColorTintShaderState->release();

is this correct?

chongchaoyu:
@mori1234 after each draw, _nextFreeCommand is not reset to 0, so _commandsPool keeps growing, this leads to memory leak.

mori1234:
> @mori1234 after each draw, _nextFreeCommand is not reset to 0, so _commandsPool keeps growing, this leads to memory leak.

m_EventUpdate = Director::getInstance()->getEventDispatcher()->addCustomEventListener(EVENT_AFTER_DRAW_RESET_POSITION, [this](EventCustom* eventCustom) {
		this->update(0); // <--- here  _nextFreeCommand is reset to 0 ?
});

chongchaoyu:
@mori1234 Sorry, I didn't get your point, you can fix this issue according to this PR or put
`spine::SkeletonBatch::destroyInstance();
spine::SkeletonTwoColorBatch::destroyInstance();`
into `Director::reset()`, but I don't think it is a good solution.

PR #20300, opened 2019-11-06
Allow game code to override BM Font parsing and support rotated font atlas textures

rh101:
The changes in this PR will allow a game to override the bitmap font parsing of custom FNT files that may contain custom data.  It also adds support for indicating if font atlas textures are rotated, which may be the case if  they exist in a larger PLIST image atlas.  

While the font definition rotated flag isn't used in FontFNT, a game can extend that class to add support for rotated textures as long as a rotated flag exists.


rh101:
@minggo I would really like to add proper support for bitmap font atlas images being part of a larger image atlas file, but the current interface is extremely limited. For instance:

`FontFNT(BMFontConfiguration *theContfig, const Vec2& imageOffset = Vec2::ZERO);`

It only accepts a `Vec2`, which is just the offset into the larger texture.  It doesn't allow for supporting rotation, which would require the width and height to be passed in as well.  A better interface would be:

`FontFNT(BMFontConfiguration *theContfig, const Rect& imageRect = Rect::ZERO, bool rotated = false);`

I've done testing with this in a separate build, and it works perfectly fine.  There is absolutely no point in just supporting `Vec2 imageOffset`, and the Rect would do exactly the same thing for anyone that happens to be using the `imageOffset`.  They can just pass in:

`FontFNT(config, Rect(x, y, 0, 0), false);`

Which would be equivalent to

`FontFNT(config, Vec2(x, y));`

If it's OK to change the current interface, then I'll be more than happy to do implement full sprite sheet support with rotation for bitmap atlas fonts.

The reason for all of this is that the bitmap font labels will be batched with whatever sprites are using the textures from the same larger sprite sheet.  This reduces the draw calls significantly.

An example of a larger sprite sheet with multiple bitmap fonts (including rotated images):
MixedFontAtlas

rh101:
I've implemented the changes, but I haven't added them to this PR yet.  Please let me know if you want me to add them.

To create a BM font using the new API, you simply do any of the following:

Standard FNT support:
```
auto label = Label::createWithBMFont("fonts/arimo_regular48.fnt", "Normal BM Font Label");
```

If you want to use a PLIST atlas that contains font atlas sub-textures (like the image in my previous post):

```
auto spriteCache = SpriteFrameCache::getInstance();
spriteCache->addSpriteFramesWithFile("fonts/MixedFontAtlas.plist");
```

You can pass the sub-texture key to `createWithBMFont()`:
```
auto atlasSubTextureKey = "arimo_bold40_0.png";
auto label1 = Label::createWithBMFont("fonts/arimo_bold40.fnt", "0123456789", TextHAlignment::CENTER, 0, atlasSubTextureKey);
```

or you can just get the frame, and then pass in the sub-texture rectange and rotated flag:
```
auto frame = spriteCache->getSpriteFrameByName("arimo_bold30_0.png");
auto label = Label::createWithBMFont("fonts/arimo_bold30.fnt", "0123456789", TextHAlignment::CENTER, 0, frame->getRectInPixels(), frame->isRotated());
```

The FNT files simply need to have "file" changed to reference the large sprite sheet:

The `fonts/arimo_bold40.fnt` has this:

```
info face="Arimo" size=40 bold=1 italic=0 charset="" unicode=1 stretchH=100 smooth=1 aa=1 padding=0,0,0,0 spacing=1,1 outline=0
common lineHeight=40 base=32 scaleW=256 scaleH=256 pages=1 packed=0 alphaChnl=1 redChnl=0 greenChnl=0 blueChnl=0
page id=0 file="MixedFontAtlas.png"
chars count=97
char id=32   x=250   y=9     width=3     height=1     xoffset=-1    yoffset=39    xadvance=9     page=0  chnl=15
...
...
```

The `fonts/arimo_bold30.fnt` has this:

```
info face="Arimo" size=30 bold=1 italic=0 charset="" unicode=1 stretchH=100 smooth=1 aa=1 padding=0,0,0,0 spacing=1,1 outline=0
common lineHeight=30 base=24 scaleW=256 scaleH=256 pages=1 packed=0 alphaChnl=0 redChnl=4 greenChnl=4 blueChnl=4
page id=0 file="MixedFontAtlas.png"
chars count=95
char id=32   x=130   y=20    width=3     height=1     xoffset=-1    yoffset=29    xadvance=7     page=0  chnl=15
...
...
```

The `MixedFontAtlas.plist` contains:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
    <dict>
        <key>frames</key>
        <dict>
            <key>arimo_bold30_0.png</key>
            <dict>
                <key>aliases</key>
                <array/>
                <key>spriteOffset</key>
                <string>{0,77}</string>
                <key>spriteSize</key>
                <string>{256,102}</string>
                <key>spriteSourceSize</key>
                <string>{256,256}</string>
                <key>textureRect</key>
                <string>{{182,238},{256,102}}</string>
                <key>textureRotated</key>
                <true/>
            </dict>
            <key>arimo_bold40_0.png</key>
            <dict>
                <key>aliases</key>
                <array/>
                <key>spriteOffset</key>
                <string>{0,40}</string>
                <key>spriteSize</key>
                <string>{256,176}</string>
                <key>spriteSourceSize</key>
                <string>{256,256}</string>
                <key>textureRect</key>
                <string>{{2,238},{256,176}}</string>
                <key>textureRotated</key>
                <true/>
            </dict>
            <key>arimo_bold46_0.png</key>
            <dict>
                <key>aliases</key>
                <array/>
                <key>spriteOffset</key>
                <string>{0,12}</string>
                <key>spriteSize</key>
                <string>{256,232}</string>
                <key>spriteSourceSize</key>
                <string>{256,256}</string>
                <key>textureRect</key>
                <string>{{2,2},{256,232}}</string>
                <key>textureRotated</key>
                <false/>
            </dict>
        </dict>
        <key>metadata</key>
        <dict>
            <key>format</key>
            <integer>3</integer>
            <key>pixelFormat</key>
            <string>RGBA8888</string>
            <key>premultiplyAlpha</key>
            <false/>
            <key>realTextureFileName</key>
            <string>MixedFontAtlas.png</string>
            <key>size</key>
            <string>{286,496}</string>
            <key>smartupdate</key>
            <string>$TexturePacker:SmartUpdate:2588fbf66b689929c9b8d4c05d4f634a:e8696770b5f61fdd6f7d42249ae525c5:ff16bb0d35c5a003894281a80d77a92d$</string>
            <key>textureFileName</key>
            <string>MixedFontAtlas.png</string>
        </dict>
    </dict>
</plist>
```

minggo:
@rh101 i think send another PR after merging this one for new API is better.

> The changes in this PR will allow a game to override the bitmap font parsing of custom FNT files that may contain custom data. 

I don't quite understand, it seems just move the codes from one file to another.

rh101:
@minggo BMFontConfiguration, _BMFontPadding and _BMFontDef were all declared and defined in the CPP file, not the header file, so there was no direct access to them in case we want to sub-class BMFontConfiguration or FontFNT (which uses those other classes/structs).

BMFontConfiguration also had a lot of private members, which meant that there was no way to override the bitmap font file parsing methods:

```
std::set<unsigned int>* parseConfigFile(const std::string& controlFile);
std::set<unsigned int>* parseBinaryConfigFile(unsigned char* pData, unsigned long size, const std::string& controlFile);
```
Now those methods are both protected, and virtual, so they can be overriden.

Not all bitmap fonts have to be in BMFont FNT format, as long as they supply the expected data, so moving those members out of the private access section enables game developers to add parsing code for any other bitmap font format file.

Also, you're right, another PR would be best for the changes to the API.

minggo:
Got it, thanks.

Issue #20174, opened 2019-09-25
CCEAGLView:setMarkedText problem in iOS13


lxhhuangjin:
- cocos2d-x version: 3.12
- devices test on: iPhone
- developing environments
   - NDK version:
   - Xcode version:
   - VS version:
   - browser type and version:

Steps to Reproduce:
input Chinese in TextField just recieve one charactor in setMarketedText , input number or English both ok.

lxhhuangjin:
be the same question with https://github.com/cocos2d/cocos2d-x/issues/20050

wandeli:
- (NSString *)textInRange:(UITextRange *)range
{
    CCLOG("textInRange");
    if (nil != markedText_) {
        return markedText_;
    }
    return @"";
}

- (UITextRange *)markedTextRange
{
    CCLOG("markedTextRange");
    if (nil != markedText_) {
        return [[[UITextRange alloc] init] autorelease];
    }
    return nil; // Nil if no marked text.
}

lxhhuangjin:
> 
> 
>     * (NSString *)textInRange:(UITextRange *)range
>       {
>       CCLOG("textInRange");
>       if (nil != markedText_) {
>       return markedText_;
>       }
>       return @"";
>       }
> 
>     * (UITextRange *)markedTextRange
>       {
>       CCLOG("markedTextRange");
>       if (nil != markedText_) {
>       return [[[UITextRange alloc] init] autorelease];
>       }
>       return nil; // Nil if no marked text.
>       }

I have tried this code. The input of Chinese is ok. But when i click the screen to close IME. The IME open again. I debug this performance and find i recieved UIKeyboardWillShowNotification.

minggo:
I don't know what's the problem, i can input Chinese characters with iPhone11(iOS 13).

lxhhuangjin:
My iPhone X(iOS13) input Chinese in
 TextField has problem. The function named setMarkedText just received one charactor I input recently.

minggo:
What test case did you test?

lxhhuangjin:
> 
> 
> What test case did you test?

Any input is ok. 
eg. First i input 'w', setMarkedText  recieved 'w', then input 'd', setMarkedText  recieved 'd'. But setMarkedText  need recieve 'wd'.

minggo:
Did you use TextFieldTTF? Please show your test case and the steps to reproduce it. Thanks.

lxhhuangjin:
[...]

PR #20303, opened 2019-11-07
fix issues of textfield

minggo:


Guy-kun:
This fixes the reported issue in https://github.com/cocos2d/cocos2d-x/issues/20174#issuecomment-550347355 at first signs.
Will say if there's any further bugs after QA

minggo:
@Guy-kun any update?

Guy-kun:
We've confirmed that there seems to be no regression and it's working correctly.

There's one small issue which is:
- On iPadOS (tested on 13.3), using the 'english' mode of the Japanese keyboard defaults to starting the input with a capital letter. (this does not happen for all text fields, such as safari url)
It's a minor issue however, does cocos have an option for this behavior? If not it should likely be forced disabled.

minggo:
Sorry, i don't know how to handle the issue in codes. And it seems it can be resolved by changing iOS devices's settings: https://www.idownloadblog.com/2015/09/18/how-to-keyboard-uppercase-iphone-ipad-ios-9/.

Guy-kun:
This isn’t the same thing, the link is for changing the keyboard display case, the actual issue is auto-capitalization of the first letter which is likely a hint set on the text field itself.
Either way this isn’t a blocking issue as it’s somewhat specific.

PR #20309, opened 2019-11-08
Improve bmfont api

rh101:
This allows for embedding multiple bitmap font image atlases into a larger sprite sheet.  Related to discussion in PR #20300 

To create a BM font using the new API, you simply do any of the following:

Standard FNT support:
```
auto label = Label::createWithBMFont("fonts/arimo_regular48.fnt", "Normal BM Font Label");
```

If you want to use a PLIST atlas that contains font atlas sub-textures (like the image in my previous post):

```
auto spriteCache = SpriteFrameCache::getInstance();
spriteCache->addSpriteFramesWithFile("fonts/MixedFontAtlas.plist");
```

You can pass the sub-texture key to `createWithBMFont()`:
```
auto atlasSubTextureKey = "arimo_bold40_0.png";
auto label1 = Label::createWithBMFont("fonts/arimo_bold40.fnt", "0123456789", TextHAlignment::CENTER, 0, atlasSubTextureKey);
```

or you can just get the frame, and then pass in the sub-texture rectange and rotated flag:
```
auto frame = spriteCache->getSpriteFrameByName("arimo_bold30_0.png");
auto label = Label::createWithBMFont("fonts/arimo_bold30.fnt", "0123456789", TextHAlignment::CENTER, 0, frame->getRectInPixels(), frame->isRotated());
```

The FNT files simply need to have "file" changed to reference the large sprite sheet:

The `fonts/arimo_bold40.fnt` has this:

```
info face="Arimo" size=40 bold=1 italic=0 charset="" unicode=1 stretchH=100 smooth=1 aa=1 padding=0,0,0,0 spacing=1,1 outline=0
common lineHeight=40 base=32 scaleW=256 scaleH=256 pages=1 packed=0 alphaChnl=1 redChnl=0 greenChnl=0 blueChnl=0
page id=0 file="MixedFontAtlas.png"
chars count=97
char id=32   x=250   y=9     width=3     height=1     xoffset=-1    yoffset=39    xadvance=9     page=0  chnl=15
...
...
```

The `fonts/arimo_bold30.fnt` has this:

```
info face="Arimo" size=30 bold=1 italic=0 charset="" unicode=1 stretchH=100 smooth=1 aa=1 padding=0,0,0,0 spacing=1,1 outline=0
common lineHeight=30 base=24 scaleW=256 scaleH=256 pages=1 packed=0 alphaChnl=0 redChnl=4 greenChnl=4 blueChnl=4
page id=0 file="MixedFontAtlas.png"
chars count=95
char id=32   x=130   y=20    width=3     height=1     xoffset=-1    yoffset=29    xadvance=7     page=0  chnl=15
...
...
```

The `MixedFontAtlas.plist` contains:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
    <dict>
        <key>frames</key>
        <dict>
            <key>arimo_bold30_0.png</key>
            <dict>
                <key>aliases</key>
                <array/>
                <key>spriteOffset</key>
                <string>{0,77}</string>
                <key>spriteSize</key>
                <string>{256,102}</string>
                <key>spriteSourceSize</key>
                <string>{256,256}</string>
                <key>textureRect</key>
                <string>{{182,238},{256,102}}</string>
                <key>textureRotated</key>
                <true/>
            </dict>
            <key>arimo_bold40_0.png</key>
            <dict>
                <key>aliases</key>
                <array/>
                <key>spriteOffset</key>
                <string>{0,40}</string>
                <key>spriteSize</key>
                <string>{256,176}</string>
                <key>spriteSourceSize</key>
                <string>{256,256}</string>
                <key>textureRect</key>
                <string>{{2,238},{256,176}}</string>
                <key>textureRotated</key>
                <true/>
            </dict>
            <key>arimo_bold46_0.png</key>
            <dict>
                <key>aliases</key>
                <array/>
                <key>spriteOffset</key>
                <string>{0,12}</string>
                <key>spriteSize</key>
                <string>{256,232}</string>
                <key>spriteSourceSize</key>
                <string>{256,256}</string>
                <key>textureRect</key>
                <string>{{2,2},{256,232}}</string>
                <key>textureRotated</key>
                <false/>
            </dict>
        </dict>
        <key>metadata</key>
        <dict>
            <key>format</key>
            <integer>3</integer>
            <key>pixelFormat</key>
            <string>RGBA8888</string>
            <key>premultiplyAlpha</key>
            <false/>
            <key>realTextureFileName</key>
            <string>MixedFontAtlas.png</string>
            <key>size</key>
            <string>{286,496}</string>
            <key>smartupdate</key>
            <string>$TexturePacker:SmartUpdate:2588fbf66b689929c9b8d4c05d4f634a:e8696770b5f61fdd6f7d42249ae525c5:ff16bb0d35c5a003894281a80d77a92d$</string>
            <key>textureFileName</key>
            <string>MixedFontAtlas.png</string>
        </dict>
    </dict>
</plist>
```

MixedFontAtlas.png:
MixedFontAtlas


rh101:
I'll fix the compilation issues soon, since it's something to do with JS and LUA bindings.

EDIT: Hopefully the bindings are all sorted now.

rh101:
@minggo This is all working now, including the JS and LUA bindings.

cocos/2d/CCFontFNT.h
```diff
@@ -136,7 +140,10 @@ class CC_DLL FontFNT : public Font
     
 public:
     
-    static FontFNT * create(const std::string& fntFilePath, const Vec2& imageOffset = Vec2::ZERO);
```
minggo, line 139: It will break compatibility.

rh101: Yep, I mentioned that in the previous discussion, but I'll re-add it as a wrapper around the new functionality. I'll have it done soon.

rh101: @minggo Would it be ok if I added the CC_DEPRECATED_ATTRIBUTE to the functions with the old API?

minggo: Yep, it is better.

rh101: Ok, it's been committed.  If any automated tests fail I'll sort them out pretty quickly, otherwise it should be all good.  Let me know if there's anything else I should address for this PR.

minggo: Thanks, i need some time to review it.

cocos/2d/CCLabel.h
```diff
@@ -197,14 +197,58 @@ class CC_DLL Label : public Node, public LabelProtocol, public BlendProtocol
...
     * @param hAlignment Text horizontal alignment.
     * @param maxLineWidth The max line width.
-    * @param imageOffset
     *
     * @return An automatically released Label object.
     * @see setBMFontFilePath setMaxLineWidth
     */
     static Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
-        const TextHAlignment& hAlignment = TextHAlignment::LEFT, int maxLineWidth = 0,
-        const Vec2& imageOffset = Vec2::ZERO);
```
minggo, line 207: It will break compatibility too.

rh101: I thought I took care of all old call signatures. Please correct me if I'm wrong, but here's what I did:

The old method has this signature: 
```
Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
        const TextHAlignment& hAlignment = TextHAlignment::LEFT, int maxLineWidth = 0,
        const Vec2& imageOffset = Vec2::ZERO);
```
So, you could have called it in these ways:
```
createWithBMFont(font, text);
createWithBMFont(font, text, alignment);
createWithBMFont(font, text, alignment, maxLineWidth);
createWithBMFont(font, text, alignment, maxLineWidth, offset);
```
Now, with the introduction of the new API, the compiler wouldn't know which method to use because it would be ambiguous, since we're using default values in some parameters. So, to fix, here's the new API:
```
CC_DEPRECATED_ATTRIBUTE static Label* createWithBMFont(const std::string& bmfontPath, const std::string& text, 
const TextHAlignment& hAlignment, int maxLineWidth, const Vec2& imageOffset);

Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
        const TextHAlignment& hAlignment = TextHAlignment::LEFT, int maxLineWidth = 0);

Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
        const TextHAlignment& hAlignment, int maxLineWidth, const Rect& imageRect, bool imageRotated);

Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
        const TextHAlignment& hAlignment, int maxLineWidth, const std::string& subTextureKey);
```

So, here are the different call combinations:
```
Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
        const TextHAlignment& hAlignment = TextHAlignment::LEFT, int maxLineWidth = 0);
```
takes care of these calls:
```
createWithBMFont(font, text);
createWithBMFont(font, text, alignment);
createWithBMFont(font, text, alignment, maxLineWidth);
```
and
```
CC_DEPRECATED_ATTRIBUTE static Label* createWithBMFont(const std::string& bmfontPath, const std::string& text, 
const TextHAlignment& hAlignment, int maxLineWidth, const Vec2& imageOffset);
```
takes care of this:
`createWithBMFont(font, text, alignment, maxLineWidth, offset);`

That keeps compatibility with all the current method call signatures, so any code using them doesn't change.

The new API methods allow for these calls (no default parameters in them):
```
createWithBMFont(font, text, alignment, maxLineWidth, rect, rotation);
createWithBMFont(font, text, alignment, maxLineWidth, textureKeyString);
```

Is there something I missed?

minggo: No, it is ok. Thanks.

cocos/2d/CCFontFNT.cpp
```diff
@@ -531,13 +572,26 @@ FontFNT * FontFNT::create(const std::string& fntFilePath, const Vec2& imageOffse
     return tempFont;
 }
 
-FontFNT::FontFNT(BMFontConfiguration *theContfig, const Vec2& imageOffset /* = Vec2::ZERO */)
-:_configuration(theContfig)
-,_imageOffset(CC_POINT_PIXELS_TO_POINTS(imageOffset))
+FontFNT* FontFNT::create(const std::string& fntFilePath, const Vec2& imageOffset)
+{
+    return create(fntFilePath, Rect(imageOffset.x, imageOffset.y, 0, 0), false);
```
minggo, line 577: Why it is `Rect(imageOffset.x, imageOffset.y, 0, 0)` ? The size is 0?

cocos/2d/CCFontFNT.cpp
```diff
@@ -641,10 +701,22 @@ FontAtlas * FontFNT::createFontAtlas()
...
-        
+
+        const float left = _imageRectInPoints.origin.x;
+        const float right = (_imageRectInPoints.origin.x + rw);
+        const float top = _imageRectInPoints.origin.y;
+
+        if (_imageRotated)
+        {
+            tempDefinition.U = right - tempRect.origin.y - tempRect.size.height;
+            tempDefinition.V = tempRect.origin.x + top;
```
minggo, line 709: Don't understand the codes. Could you please explain it? Should right be changed to
`right = (_imageRectInPoints.origin.y + rw);`?

minggo: And i think the variable name is confused.

cocos/2d/CCFontAtlasCache.cpp
```diff
@@ -88,22 +88,53 @@ FontAtlas* FontAtlasCache::getFontAtlasTTF(const _ttfConfig* config)
...
-FontAtlas* FontAtlasCache::getFontAtlasFNT(const std::string& fontFileName, const Vec2& imageOffset /* = Vec2::ZERO */)
+FontAtlas* FontAtlasCache::getFontAtlasFNT(const std::string& fontFileName)
 {
-    auto realFontFilename = FileUtils::getInstance()->getNewFilename(fontFileName);  // resolves real file path, to prevent storing multiple atlases for the same file.
+    return getFontAtlasFNT(fontFileName, Rect::ZERO, false);
+}
+
+FontAtlas* FontAtlasCache::getFontAtlasFNT(const std::string& fontFileName, const std::string& subTextureKey)
+{
+    const auto realFontFilename = FileUtils::getInstance()->getNewFilename(fontFileName);  // resolves real file path, to prevent storing multiple atlases for the same file.
```
minggo, line 98: I think should use `FileUtils::getInstance()->fullPathForFilename` instead.

cocos/2d/CCFontAtlasCache.cpp
```diff
@@ -88,22 +88,53 @@ FontAtlas* FontAtlasCache::getFontAtlasTTF(const _ttfConfig* config)
...
+    }
+    else
+        return it->second;
+
+    return nullptr;
+}
+
+FontAtlas* FontAtlasCache::getFontAtlasFNT(const std::string& fontFileName, const Rect& imageRect, bool imageRotated)
+{
+    const auto realFontFilename = FileUtils::getInstance()->getNewFilename(fontFileName);  // resolves real file path, to prevent storing multiple atlases for the same file.
```
minggo, line 124: I think should use FileUtils::getInstance()->fullPathForFilename instead.

cocos/2d/CCFontFNT.cpp
```diff
@@ -508,20 +510,59 @@ void BMFontConfiguration::parseKerningEntry(const char* line)
...
     if (!newConf)
         return nullptr;
     
+    const auto tempFont = new FontFNT(newConf, imageRect, imageRotated);
+    
+    tempFont->setFontSize(newConf->_fontSize);
+    if (!tempFont)
+    {
+        return nullptr;
+    }
```
minggo, line 524: Should check tempFont before setting font size.

cocos/2d/CCFontFNT.cpp
rh101: The current API uses a Vec2, which is the X and Y offset of the bitmap font image in the sprite sheet.  In order to support rotation in a sprite sheet (90° to the right for PLIST sprite sheets), it needs x, y, width, height.  Now in adapting the old API to the new API, the passed in Vec2 imageOffset, we can just use a Rect with the image offset X,Y and width/height both 0, since it works exactly the same way.  

Any code using the Vec2 imageOffset implies that the sprite sheet does not have rotated textures.  There is a test in cpp-tests that shows this, under the "Old Label API".  So, this gets converted to Rect with width/height 0, and the code path does exactly the same thing as before (it doesn't use width/height if the texture isn't rotated).

cocos/2d/CCFontFNT.cpp
```diff
@@ -508,20 +510,59 @@ void BMFontConfiguration::parseKerningEntry(const char* line)
...
+{
+    const auto newConf = FNTConfigLoadFile(fntFilePath);
+    if (!newConf)
+        return nullptr;
+
+    const auto frame = SpriteFrameCache::getInstance()->getSpriteFrameByName(subTextureKey);
+    if (!frame)
+    {
+        return nullptr;
+    }
```
minggo, line 540: Should release newConf if frame before return.

minggo: Can check frame first.

cocos/2d/CCFontFNT.cpp
```diff
@@ -508,20 +510,59 @@ void BMFontConfiguration::parseKerningEntry(const char* line)
...
+    auto tempFont = new FontFNT(newConf, frame->getRectInPixels(), frame->isRotated());
+
+    tempFont->setFontSize(newConf->_fontSize);
+    if (!tempFont)
+    {
+        return nullptr;
+    }
```
minggo, line 546: should check tempFont before setting font size

cocos/2d/CCFontFNT.cpp
minggo: Oh, sorry, newConf is an autorelease object.

cocos/2d/CCFontFNT.cpp
rh101: That code is directly out of CCSprite.cpp.  It calculates the boundaries of an image inside the sprite sheet, as you can see in `Sprite::setTextureCoords()`:

```
    float left    = rectInPixels.origin.x / atlasWidth;
    float right   = (rectInPixels.origin.x + rw) / atlasWidth;
    float top     = rectInPixels.origin.y / atlasHeight;
    float bottom  = (rectInPixels.origin.y + rh) / atlasHeight;
```

Now, with the new code in `FontFNT::createFontAtlas()`:

```
        if (_imageRotated)
        {
            tempDefinition.U = right - tempRect.origin.y - tempRect.size.height;
            tempDefinition.V = tempRect.origin.x + top;
        }
        else
        {
            tempDefinition.U = tempRect.origin.x + left;
            tempDefinition.V = tempRect.origin.y + top;
        }
```

If the bitmap font image is rotated 90 degrees to the right in the sprite sheet, then The U and V coordinates need to point to the correct position in the sprite sheet.  

The U and V calculated there result in values that the BM font code expects to extract the BM font image.  Took me a while to figure out what the BM font code required, and that is the correct formula.

cocos/2d/CCFontAtlasCache.cpp
rh101: I don't know about that code.  That is from the current implementation, I never touched it.  Are you sure changing it would work?

cocos/2d/CCFontFNT.cpp
```diff
@@ -508,20 +510,59 @@ void BMFontConfiguration::parseKerningEntry(const char* line)
...
     {
         return nullptr;
     }
-    
-    FontFNT *tempFont =  new FontFNT(newConf,imageOffset);
+    FontFNT* tempFont = new FontFNT(newConf);
+
     tempFont->setFontSize(newConf->_fontSize);
     if (!tempFont)
     {
```
minggo, line 568: should check tempFont first

cocos/2d/CCFontAtlasCache.cpp
rh101: Same for this, I never modified this code, that's all from the current implementation. I can change it if you require, but are you certain it would work?

cocos/2d/CCFontFNT.cpp
```diff
@@ -393,9 +394,10 @@ void BMFontConfiguration::parseImageFileName(const char* line, const std::string
     int pageId;
     sscanf(line, "page id=%d", &pageId);
     CCASSERT(pageId == 0, "LabelBMFont file could not be found");
+
     // file 
     char fileName[255];
-    sscanf(strchr(line,'"') + 1, "%[^\"]", fileName);
+    sscanf(strstr(line, "file=\"") + 6, "%[^\"]", fileName);
```
minggo, line 400: Will it break compatibility?

cocos/2d/CCFontFNT.cpp
rh101: newConf is not explicitly retained.  This code was not modified, that is the current implementation.  Doesn't newConf automatically get released when it goes out of scope (since autorelease() was called on it)?

cocos/2d/CCFontFNT.cpp
rh101: Good catch, that's also the current code, since I never modified it, but I'll fix it.

cocos/2d/CCFontFNT.cpp
rh101: Before it just checked for a `"`, but it's not a good way to handle it.  It's better to explicitly set what you're searching for in the string.

For example, it looks at this line in the FNT file:

```
page id=0 file="arimo_bold30_0.png"
```

The current code searches just for a `"`, which works fine.  The modified code just searches for the whole `file="` instead, to make it clearer what the purpose is.  I can revert it if you prefer, since it really makes no difference.

cocos/2d/CCFontFNT.cpp
rh101: Yes, that's based on the current code, and wasn't modified by me, but I'll fix it now.

cocos/2d/CCFontAtlasCache.cpp
minggo: Oh, i thought it was added by you. Ok, let's keep it.

cocos/2d/CCFontFNT.cpp
minggo: Got it, thanks. Let's keep it.

cocos/2d/CCFontFNT.cpp
minggo: What i confused is that
- right is calculated with origin.x 
- right is used together with origin.y and size.height

No matter it is rotated or not, it is confused to use origin.x and origin.y together.

minggo: - float right   = (rectInPixels.origin.x + rw) / atlasWidth;
- tempDefinition.U = right - tempRect.origin.y - tempRect.size.height;

so 
- tempDefinition.U = (rectInPixels.origin.x + rw) / atlasWidth - tempRect.origin.y - tempRect.size.height

The logic is strange.

rh101: Oh that's because `U` is expected to be the top left position of the letter.  

So, if it's rotated 90° to the right in the sprite, that expression gets the correct coordinate for the top-left of the letter when it's rotated.  

The `tempRect` is that of the specific letter in the BM font texture, but it based on a texture that isn't rotated.  We have to use the values and translate them if the texture is rotated 90° to the right.  While the algorithm looks weird, it's actually correct.

minggo: So what's the coordinate of `_imageRectInPoints ` and `tempRect `:
- does _imageRectInPoints's coordinated based on rotated coordinate or not?
- does tempRect' coordinate based on rotated coordinate or not?

rh101: `_imageRectInPoints` is the rectangle of the entire bitmap font texture, ignoring rotation.  It's what comes from either a Vec2 imageOffset supplied by the user, or a rect supplied by the user, or the SpriteFrame rect.  The values do not factor in rotation, so they're just the rect of the image as if it was not rotated.

`tempRect` is the letter definition as it comes from the FNT file:
```
        BMFontDef& fontDef = e.second;
        ...
        Rect tempRect;
        
        tempRect = fontDef.rect;
        tempRect = CC_RECT_PIXELS_TO_POINTS(tempRect);
```

Example from FNT file:
```
char id=32   x=251   y=31    width=3     height=1     xoffset=-1    yoffset=45    xadvance=11    page=0  chnl=15
```

The FNT file doesn't know about the rotation, so those are non-rotated values in `tempRect`.

rh101:
```
float right = (rectInPixels.origin.x + rw) / atlasWidth;
tempDefinition.U = right - tempRect.origin.y - tempRect.size.height
```

That is to calculate the top-left position of the letter image in the bitmap texture.  So, for example:

image

```
letter "M" top left  = U = right X position of texture - tempRect.origin.y (green line) - tempRect.size.height (blue line)

and 

V = top + tempRect.origin.x (which is shown by the red line)
```

When it passes this information on to create the Sprite from these coordinates, the sprite can then render the letter out correctly, since it knows it's been rotated.

cocos/2d/CCLabel.h
```diff
@@ -197,14 +197,58 @@ class CC_DLL Label : public Node, public LabelProtocol, public BlendProtocol
...
+    * @param text The initial text.
+    * @param hAlignment Text horizontal alignment.
+    * @param maxLineWidth The max line width.
+    * @param subTextureKey Name of entry in PLIST texture atlas/sprite sheet
+    *
+    * @return An automatically released Label object.
+    * @see setBMFontFilePath setMaxLineWidth
+    */
+    static Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
+        const TextHAlignment& hAlignment, int maxLineWidth, const std::string& subTextureKey);
```
minggo, line 236: Can subTextureKey and imageRotated used together?

rh101: No, they can't, and they shouldn't be, since the method to accept a `subTextureKey` is purely for convenience.   The `subTextureKey` is used to load the SpriteFrame, which internally has the data on whether the texture is rotated or not.

I was considering an alternative implementation, which would accept the `SpriteFrame*` instead, like this:

```
static Label* createWithBMFont(const std::string& bmfontPath, const std::string& text,
        const TextHAlignment& hAlignment, int maxLineWidth, SpriteFrame* frame);
```

but that would have resulted in more typing:
```
auto frame = spriteCache->getSpriteFrameByName(subTextureKey);
auto label = Label::createWithBMFont("fontname.fnt", "text", TextHAlignment::CENTER, 0, frame);
```

That was my rational for adding the convenience method that accepts the `subTextureKey`.

cocos/2d/CCFontFNT.cpp
minggo: I found rotated information is only used here. How does it affect the frame rotation?

rh101: Where it's used was the previous PR #20300 

In CCLabel.cpp from that PR, I changed it to use the rotated flag:

Line 793:
`letterSprite->setTextureRect(uvRect, letterDef.rotated, uvRect.size);`

Line 974:
`_reusedLetter->setTextureRect(_reusedRect, letterDef.rotated, _reusedRect.size);`

Line 1802:
`letter = LabelLetter::createWithTexture(_fontAtlas->getTexture(textureID), uvRect, letterDef.rotated);`

minggo:
@rh101 could you please add a test case?

rh101:
No worries, I'll add a test case soon.

rh101:
Tests added under the "Label - New API" section.  They should be tests 16 and 17.  It tests an atlas with no rotated images (non-PLIST), and another with rotated images (and PLIST).  The non-PLIST atlas test is to validate that it still works as it did previously, and it's equivalent to the test in "Label - Old API".

minggo:
Thanks.

cocos/2d/CCFontFNT.cpp
```diff
@@ -641,10 +701,22 @@ FontAtlas * FontFNT::createFontAtlas()
...
         tempDefinition.offsetX  = fontDef.xOffset;
         tempDefinition.offsetY  = fontDef.yOffset;
-        
-        tempDefinition.U        = tempRect.origin.x + _imageOffset.x;
-        tempDefinition.V        = tempRect.origin.y + _imageOffset.y;
-        
+
+        const float left = _imageRectInPoints.origin.x;
+        const float right = (_imageRectInPoints.origin.x + rw);
+        const float top = _imageRectInPoints.origin.y;
```
minggo: Define these variables outside of `for` is better.

rh101: You're right, I missed that completely. I'll sort it out now.

minggo (approved):

PR #20313, opened 2019-11-08
[iOS]Screen Time would makes App stuck sometimes

wanghehacker:
Screen Time would prevent UIApplicationDidBecomeActiveNotification and UIApplicationWillResignActiveNotification from being fired until reboot the iPad.

OS: iOS 13.1.2
Device: iPad Pro (10.5-inch) MPF02CH/A

 Follow these steps:
1. Enter "Settings->Screen Time".
2.Turn on "Screen Time".
3. Click "Use Screen Time Passcode" and set a passcode.
4. Enter "Downtime"，turn it on, select "Every Day" and  "From 0:01 To 23:59". 
5. Enter "App Limits", turn it on, click "Add Limit", select "All Apps & Categories" and then click "Next", select 3min and turn on "Block at End of Limit" , then click "Add".
6. Back to Home, start an App that base on cocos2d-x.
7. Wait for "Time Limit Warning Layer" to apear, click "Ignore Limit->One more minute".
8. One mimute later,"Time Limit Warning Layer" appear again, click "Ask For More Time->Enter Screen Time Passcode", and enter passcode, then an menu would appear,DO NOT SELECT IT !!!  PRESS HOME BUTTON !!! Put the App enter backgroud.
9. Open the App again, click "Ask For More Time->Enter Screen Time Passcode", and enter passcode, select "Approve for 15 minutes".
10. The App would be stuck, kill process and reopen it, has no effect.
This bug affects all the Apps installed on the iPad. I think the iOS operating system should be broken at that moment. Close the Screen Time and reboot the iPad, everything returns to normal.


templates/cpp-template-default/proj.ios_mac/ios/RootViewController.mm
```diff
@@ -68,6 +69,13 @@ - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
 }
 
+//In iOS 12.0+, Screen Time's bug cause UIApplicationDidBecomeActiveNotification and UIApplicationWillResignActiveNotification do not fired
+//so we need to active CCDirectorCaller manually
+-(void)viewDidAppear:(BOOL)animated{
+    [super viewDidAppear:animated];
+    [[CCDirectorCaller sharedDirectorCaller] setActive:YES];
+}
```
minggo, line 77: `CCDirectorCaller-ios.mm` already listen `UIApplicationWillEnterForegroundNotification` event, why need these codes too?

wanghehacker: > `CCDirectorCaller-ios.mm` already listen `UIApplicationWillEnterForegroundNotification` event, why need these codes too?

When the App starts up cold,UIApplicationWillEnterForegroundNotification would not be triggered . But the viewDidAppear would be called.

[https://developer.apple.com/documentation/uikit/uiapplicationwillenterforegroundnotification?language=objc](https://developer.apple.com/documentation/uikit/uiapplicationwillenterforegroundnotification?language=objc).

minggo (approved):

Mee-gu:
@wanghehacker Since I do not have an iPad Pro, I try both on my iPhone 8(iOS 13.1.3) and iPhone 6(iOS 12.1.4) and can't reproduce it.

wanghehacker:
> @wanghehacker Since I do not have an iPad Pro, I try both on my iPhone 8(iOS 13.1.13) and iPhone 6(iOS 12.1.4) and can't reprocude it.

This bug seems reproduce on iPad only. I can't reproduce it on iPhone too.

PR #20314, opened 2019-11-08
Support multiple build configurations

crazyhappygame:
Support multiple configurations
https://cmake.org/cmake/help/v3.16/variable/CMAKE_CONFIGURATION_TYPES.html#variable:CMAKE_CONFIGURATION_TYPES
https://cmake.org/cmake/help/v3.16/prop_tgt/MAP_IMPORTED_CONFIG_CONFIG.html

For windows to enable default cmake configuration: RELWITHDEBINFO and MINSIZEREL below pull request is needed:
https://github.com/cocos2d/cocos2d-x-3rd-party-libs-bin/pull/391

PR #20317, opened 2019-11-11
Cmake mult config

crazyhappygame:
The same was done for v4
https://github.com/cocos2d/cocos2d-x/pull/20314


crazyhappygame:
@minggo could you merge it?

PR #20319, opened 2019-11-12
Sync resource automatically on Windows and Linux

PatriceJiang:
### Issues related to resources
- After editing a Lua script file, it does not trigger compilation and `POST_BUILD` commands will not fire. This causes the process started by the IDE to use non-synchronized resources. 
- Adding files in the `Resources/` directory will not be automatically synchronized to the build directory.
- Lua scripts will not be compiled to `.luac` files in Release build

### Fix 
- Trigger copy resource on every build action. 
- Use an external python script to sync resource files base on "last modification timestamp", which is much faster than `cmake -E` commands. 
- Compile Lua scripts in Release mode


crazyhappygame:
@PatriceJiang @minggo Please remove this pull request and create PR with following instead:
```
if(LINUX OR WINDOWS)
    add_custom_command(
        TARGET ${APP_NAME}
        PRE_BUILD COMMAND ${CMAKE_COMMAND} -E
        create_symlink ${CMAKE_CURRENT_SOURCE_DIR}/Resources $<TARGET_FILE_DIR:${APP_NAME}>/Resources)
endif()
```

This is what I use for my games. Above code instead coping resources create once symlink.
Advantages:
1. The fastest - create symlinks once during configurations
2. File are always in sync
3. No coping

PatriceJiang:
It is a great way for most projects.
However lua projects will generate `luac` files into Resources folder, copy is still needed.

crazyhappygame:
@PatriceJiang I do not use Lua. 
What is the problem with Lua, generated files and symlinks?
I think it still will work with symlinks.

PatriceJiang:
Lua projects will try to load "luac" scripts at first, if "lua" and "luac" files are both provided at the same time, "lua" files will never be loaded. And developers would not be able to debug project. 

In Release mode, as Android does, ".lua" files will be compiled into ".luac" files by using `luajit`.

---- 
Edit:
Once we run app in release mode, `luac` files will be generated in `Resources/` folder. If we try to debug then,  app will not load `.lua` files.

crazyhappygame:
I am so sorry but I still do not see why we can not use symlinks. Less code and simple.

Any way luac files are generated.

PatriceJiang:
Folder `Resources/` need to be different in release and debug mode. By linking to the same `Resources/` folder this can not be done.

crazyhappygame:
OK. I know nothing about Lua.
But can we do
1. symlinks for cpp projects
2. python script for Lua
?

PatriceJiang:
Good idea.

crazyhappygame:
Could you create separate *.cmKe file for Lua only? 
For example:
cmake/Modules/CocosLuaBuildHelpers.cmake

PatriceJiang:
@crazyhappygame I am sorry, could you explain why we need a separate cmake file?

minggo (approved):

PR #20322, opened 2019-11-13
add opengl header since Texture2D and Image will use opengl define

Mee-gu:


cocos/platform/CCImage.cpp
```diff
@@ -33,6 +33,10 @@ THE SOFTWARE.
 #include "base/CCData.h"
 #include "base/ccConfig.h" // CC_USE_JPEG, CC_USE_WEBP
 
+#ifndef CC_USE_METAL
+#include "platform/CCGL.h"
+#endif
```
minggo: Don't need `CC_USE_METAL `.

minggo (approved):

PR #20343, opened 2019-11-15
support RenderTexture saveToFile with WEBP format, compressionQuality for WEBP and JPG

0xJimmyBeh:
1. save image with WEBP format can save a lot space on device and cocos2dx do support WEBP.

2. current saveToFile with JPG format will save in 100% compression quality, but `UIImageJPEGRepresentation(image, 1.0f)` can easily accept any compression quality.

i'm only develop in iOS and not familiar with android, so these changes are for iOS only.


rh101:
> i'm only develop in iOS and not familiar with android, so these changes are for iOS only.

`WebPConfigPreset` requires a quality factor to be between 0.0-100.0, not 0.0-1.0 (the `compressionQuality` parameter you're passing into it).  You can leave the `compressionQuality` as 0-1 range, and just multiply by 100 when you use it.  For example:

`WebPConfigPreset(&config, WEBP_PRESET_DEFAULT, compressionQuality * 100.f)`

To support platforms other than iOS, the following changes to CCImage.cpp should work:

Add the `encode.h` include statement:
```
#if CC_USE_WEBP
#include "decode.h"
#include "encode.h"
#endif // CC_USE_WEBP
```
Update this function:
```
#if (CC_TARGET_PLATFORM != CC_PLATFORM_IOS)
bool Image::saveToFile(const std::string& filename, bool isToRGB, float compressionQuality)
{
...
    else if (fileExtension == ".webp")
    {
        saveImageToWEBP(filename, compressionQuality);
    }
...
}
#endif
```

Add this function:
```
bool Image::saveImageToWEBP(const std::string& filePath, float compressionQuality)
{
#if CC_USE_WEBP
    bool ret = false;
    do
    {
        FILE* fp = fopen(FileUtils::getInstance()->getSuitableFOpen(filePath).c_str(), "wb");
        CC_BREAK_IF(nullptr == fp);

        const auto stride = _width * (hasAlpha() ? 4 : 3);

        WebPConfig config;
        if (!WebPConfigPreset(&config, WEBP_PRESET_DEFAULT, compressionQuality * 100.f)) {
            CCLOG("WebPConfigPreset Configuration preset failed to initialize.");
            return false;
        }

        if (compressionQuality == 1.0f)
        {
            const auto compressionLevel = 4; 
            if (!WebPConfigLosslessPreset(&config, compressionLevel)) {
                CCLOG("WebPConfigLosslessPreset Lossless configuration failed.");
                return false;
            }
        }

        if (!WebPValidateConfig(&config)) {
            CCLOG("One or more configuration parameters are beyond their valid ranges.");
            return false;
        }

        WebPPicture pic;
        if (!WebPPictureInit(&pic)) {
            CCLOG("Failed to initialize structure. Version mismatch.");
            return false;
        }

        pic.width = (int)_width;
        pic.height = (int)_height;

        pic.use_argb = (compressionQuality == 1.f) ? 1 : 0;
        pic.colorspace = WEBP_YUV420;

        if (hasAlpha())
        {
            WebPPictureImportRGBA(&pic, _data, (int)stride);
        }
        else
        {
            WebPPictureImportRGB(&pic, _data, (int)stride);
        }

        if (compressionQuality < 1.0f)
        {
            WebPPictureARGBToYUVA(&pic, WEBP_YUV420);
            WebPCleanupTransparentArea(&pic);
        }

        WebPMemoryWriter writer;
        WebPMemoryWriterInit(&writer);
        pic.writer = WebPMemoryWrite;
        pic.custom_ptr = &writer;
        WebPEncode(&config, &pic);

        fwrite(writer.mem, writer.size, 1, fp);
        fclose(fp);

        free(writer.mem);
        WebPPictureFree(&pic);

        ret = true;
    } while (0);
    return ret;
#else
    CCLOG("webp is not enabled, please enable it in ccConfig.h");
    return false;
#endif // CC_USE_WEBP
}
```

There is one other thing regarding the iOS implementation that you should consider.  If the quality is set to 100% (`compressionQuality = 1.0f`), then it should use the lossless WEBP compression method, not the lossy compression.  The file size will still be smaller than PNG.  If we want 100% quality, then lossless is the way to do it.  If a developer doesn't want to use lossless, then passing a value a little less than 1.f, like `compressionQuality = 0.99f` etc., would ensure it still uses lossy compression.

The implementation in the `Image::saveImageToWEBP` shows you how this is done, and it's just a matter of updating your iOS version to cater for it.

0xJimmyBeh:
@rh101 
Thanks for your help on the WebPConfigPreset and other platform support.

I have made some changes to the compressionLevel so we can set it easily with the same compressionQuality variable.

the compressionLevel range is between 0-9, set compressionQuality to 1.0f- 10.0f to control.

rh101:
@Zinitter Awesome! Being able to set the compression level for lossless is great!

Just one thing, according to the documentation in `encode.h`, it's better to use ARGB for lossless, and YUV for lossy compression.  Check `encode.h`, approximately line 296:

```
  // It is recommended to use ARGB input (*argb, argb_stride) for lossless
  // compression, and YUV input (*y, *u, *v, etc.) for lossy compression
  // since these are the respective native colorspace for these formats.
```

So, for the IOS version of the implementation, this section:
```
            pic.colorspace = WEBP_YUV420;

            WebPPictureImportRGBA(&pic, webPImageData, (int)webPBytesPerRow);
            WebPPictureARGBToYUVA(&pic, WEBP_YUV420);
            WebPCleanupTransparentArea(&pic);
```

It's always converting to YUV, which actually a lossy process.  Perhaps it would be better to do it as follows:

```
         pic.colorspace = WEBP_YUV420;

        WebPPictureImportRGBA(&pic, webPImageData, (int)webPBytesPerRow);
        if (compressionQuality < 1.0f)
        {
            WebPPictureARGBToYUVA(&pic, WEBP_YUV420);
            WebPCleanupTransparentArea(&pic);
        }
```

So YUV is only used for the lossy compression.

Thanks for actually taking the time to add support for both the compression and the WEBP.  It's been on my TODO list for a long time, but was very low priority.  Hopefully these changes get merged in before the next release.

0xJimmyBeh:
ok, thanks for your notice as i didnt study much for those colorspace handling.
i follow your implementation about those alpha and colorspace for iOS, is there anything else i miss?

yeah, i hope the webp support can be merged too since it is very nice format for image compression and quality!

rh101:
No problem at all.  It's probably best you add a test cases for the webp and compression functionality, with several different compression values (for webp/jpeg etc), to make sure it's all working correctly.  Maybe the best place for the tests would be the RenderTexture test section of cpp-tests.

0xJimmyBeh:
@rh101 
for the test case, can you help on the cpp and lua? i can do the js.

rh101:
The most important is the CPP version.  I'm not sure if LUA and JS tests are even required, given that the point of the test is to ensure the functionality is correct, and that will be achieved with the cpp tests.  The LUA and JS tests are limited and don't mirror the cpp-tests, so there seems to be no purpose in implementing them too.

If you have issues with the CI showing errors, then it'll most likely be the JS and LUA bindings, which aren't related to the tests.  You need to regenerate the bindings and add them to the PR.  If you don't know how, then check [this ](https://discuss.cocos2d-x.org/t/pull-request-didnt-pass-ci/48231/11) forum post.

0xJimmyBeh:
i thought the test should added to all, but i saw the test case cpp version are just 2

1. _target->saveToFile(png, Image::Format::PNG, true, callback);
2. _target->saveToFileAsNonPMA(png, Image::Format::PNG, true, callback);

i using cocos2dx in js, used to success generated bindings with my own modifications.

but i not sure why this branch failed on this

```
Processing section cocos2dx_extension 

/extensions/assets-manager/Manifest.h', line 38, column 10>,
    details = "'json/document-wrapper.h' file not found">
```

rh101:
> Processing section cocos2dx_extension 
> 
> /extensions/assets-manager/Manifest.h', line 38, column 10>,
>     details = "'json/document-wrapper.h' file not found">

Sorry, don't know what would be causing that issue.  I can see the errors on the CI systems right now relate to something different:

`cannot assign to variable 'compressionLevel' with const-qualified type 'const int'`

Did you actually manage to build it on your own system before uploading the changes? If not, it's best that you do that first.

0xJimmyBeh:
i found that the js bindings is manual on cocos2d_specifics.cpp, i need sometime to check whats wrong on the saveToFile with webp, i modified to accept 5 arguments but still wrong.

png and jpeg testcase on js can be save successfully.

ya, there is error on the type of `compressionLevel` and `else if`

rh101:
> i found that the js bindings is manual on cocos2d_specifics.cpp, i need sometime to check whats wrong on the saveToFile with webp, i modified to accept 5 arguments but still wrong.

If you haven't done this already, did you check if the webp output works when you use the cpp-tests?  If it does work, then at least you would have narrowed down the issue to the JS binding code.

0xJimmyBeh:
i fixed some mistake, added test case to `RenderTextureTest.cpp` and tested working on iOS.

haven't fix the js version, still have error on `js_cocos2dx_RenderTexture_saveToFile : wrong number of arguments`

0xJimmyBeh:
@rh101 
can you please help test whats wrong on save webp with quality >= 1.0 on cpp-test?
i cant figure which part i do wrongly as the js-test working fine on all saveToFile format.

between, i only able to test on iOS with iPhone.

rh101:
The crash is happening because not enough space has been allocated for the image filename.  

```
    char image[20];
    sprintf(image, "image-600webp-%d.webp", counter);
```

It is 21 characters, not 20 (remember to add + 1 for the null terminator), but also, it's not reliable, because if it gets into double digits, it'll be 22 characters and so on.

You should use this instead:
`auto imagePath = "image-100webp-" + StringUtils::toString(counter) + ".webp";`

0xJimmyBeh:
ok thanks again for the guide and fixed, it works!

but just now the ci build is still fail on linux_clang_tidy build.

PR #20347, opened 2019-11-18
Fixed duplicate addsearchpath

imtrobin:


cocos/platform/CCFileUtils.cpp
```diff
@@ -1076,6 +1076,15 @@ void FileUtils::addSearchPath(const std::string &searchpath,const bool front)
         path += "/";
     }
 
+	auto iter = std::find(_searchPathArray.begin(), _searchPathArray.end(), searchpath);
+    if (iter != _searchPathArray.cend())
+    {
+        if (front)
```
minggo, line 1082: Is it better change to `if (front && iter != _searchPathArray.begin() )`?

imtrobin: yes, I think so. I just applied the patch from the the previous approved commit.

imtrobin: actually, no. It looks different. Actually I don't really know what front means. The check enters if a duplicate is found, and if front, or it will return.

minggo: front means needs to add the path in the front of the array. So if it is already in the front, then don't have to remove and add again.

Issue #19890, opened 2019-07-02
UiLayout is not a necessary getClippingRect correction


Rofl113:
1. Sequencing:
1.1 Create UiListView 
1.2 Set Anchor Point (0.0, 1.0)
1.3 UiLayout::setClippingType(ClippingType::SCISSOR);
1.4 Total: Anchor Point counted twice

2. Possible location of the problem:
2.1 Layout::getClippingRect
UiLayout (ClippingType::SCISSOR) call func Layout::getClippingRect()
If _clippingParent == nullptr:
`_clippingRect.origin.x = worldPos.x - (scissorWidth * _anchorPoint.x);
_clippingRect.origin.y = worldPos.y - (scissorHeight * _anchorPoint.y);
_clippingRect.size.width = scissorWidth;
_clippingRect.size.height = scissorHeight;`
2.2 But the amendment (Anchor Point) is already inside the function convertToWorldSpace:
`Vec2 worldPos = convertToWorldSpace(Vec2::ZERO);`

3. Possible Solution:
3.1 Correct function Layout::getClippingRect()
If _clippingParent == nullptr:
`_clippingRect.origin.x = worldPos.x;
_clippingRect.origin.y = worldPos.y;
_clippingRect.size.width = scissorWidth;
_clippingRect.size.height = scissorHeight;`

rh101:
@minggo I just ran into this issue as well.  It's easy to reproduce.

Expected:
![image](https://user-images.githubusercontent.com/8603230/69187774-585d3400-0b6f-11ea-85f0-60fad88c26e2.png)

Actual:
![image](https://user-images.githubusercontent.com/8603230/69187867-7dea3d80-0b6f-11ea-8bde-4cd829b7d8db.png)

The problem only occurs if we use `cocos2d::ui::Layout::ClippingType::SCISSOR` and any anchor point that is not `Vec2(0, 0)`.

This code reproduces it:
```
auto panel = cocos2d::ui::Layout::create();
panel->setBackGroundColor(Color3B::RED);
panel->setBackGroundColorType(ui::Layout::BackGroundColorType::SOLID);
panel->setClippingType(cocos2d::ui::Layout::ClippingType::SCISSOR);
panel->setPosition(visibleSize / 2);
panel->setAnchorPoint(Vec2::ANCHOR_MIDDLE); // 0.5, 0.5
panel->setClippingEnabled(true);
panel->setContentSize(visibleSize);
this->addChild(panel);

[...]

PR #20352, opened 2019-11-20
Fix ui::Layout scissor clipping

rh101:
This will fix the issue reported in #19890 

The SCISSOR clipping position was not being calculated correctly.  It affected any UI element that uses this, including `ui::Layout`, `ui::ScrollView` etc.

Also added test case to reproduce the issue and to show that the fix works.


minggo (approved):

cocos/ui/UILayout.cpp
```diff
@@ -485,29 +484,29 @@ const Rect& Layout::getClippingRect()
         
         if (_clippingParent)
         {
-            parentClippingRect = _clippingParent->getClippingRect();
-            float finalX = worldPos.x - (scissorWidth * _anchorPoint.x);
-            float finalY = worldPos.y - (scissorHeight * _anchorPoint.y);
+            const Rect parentClippingRect = _clippingParent->getClippingRect();
```
minggo: Use reference instead?

rh101: Sure I'll update that now

PR #20353, opened 2019-11-20
v4-multi-meta-textures, inprogress.

halx99:


minggo:
We will be releasing v4 recently, so i will not merge it into v4.0.

halx99:
ok, could be v4-next, and there still somethings to be done.

halx99:
@minggo  And I have a question, currently if device doesn't support, then the cocos2d-x v4 app will crash due to create device failed.

halx99:
> @minggo And I have a question, currently if device doesn't support, then the cocos2d-x v4 app will crash due to create device failed.

confirmed, iphone device is ok, iphone simulator does support metal

minggo:
Yep, v4 only supports metal. And macOS 15 + Xcode 11 supports metal in iPhone simulator.

halx99:
[ci rebuild]

